PR #133, opened 2015-05-07
automake 1.14 -> 1.15

X4fyr:


MarPiRK:
As I'm porting build system to CMake right now, I don't think this pull is necessary.
If you'd like to join our Trello and Slack group please send me your email, so you can track things we're currently working on.

PR #140, opened 2015-08-17
Mainly poppler upgrade and some minor improvements

MarPiRK:
The biggest change down there is poppler upgrade – it's using now version installed in system and downloads full sources of this version to gain access to `CairoOutputDev.h`. I don't know if it's correct way of doing this, but I took a risk and I think, that it works. Some deep testing is needed tho. The best best will be, if person who knows exactly how it's working would test it. Currently I'm on holidays, so I can't do it in person.
Unfortunately Travis persists with even more ancient poppler version, so it'll be needed to create static build system for it (as it is for boost and glibmm).

The other changes are as follow:

- finished Polish translation
- fixed glade search path to include source directory
- unified translation engine to boost::locale::translate

Btw for now I'm suspending my work on this project, because I think I'm a little fed up with high-level programming. Of course I'll surely respond to all messages, but won't write too much code if not needed.


MarPiRK:
Unfortunately Travis persists with even more ancient poppler version, so it'll be needed to create static build system for it (as it is for boost and glibmm).

wbrenna:
Thanks for working on this! I suspect that the issue is still the lack of header files in recent poppler libraries for Debian - this is why I made everything go to a static configuration. For me, I get this:
```
-- Could not find poppler include dir
CMake Error at cmake/FindPoppler.cmake:74 (file):
  file STRINGS file
  "~/etc/xournalpp-marpirk/POPPLER_INCLUDE_DIR-NOTFOUND/poppler-config.h"
  cannot be read.
  CMakeLists.txt:90 (find_package)
```
which means it's not finding poppler-config.h. This is now available in ```libpoppler-private-dev```. I don't recall whether this package was available back in Squeeze with newer poppler - I think probably not, and that's why I had to build statically. It certainly seems from the package name like it's a discouraged dependency to have...

MarPiRK:
It should be something like `libpoppler-dev` and `libpoppler-glib-dev`. On Fedora it works flawlessly.
What version do you have on your system?

wbrenna:
With only those two installed it doesn't compile. I'm pretty sure it's not the version of poppler but the distro specific packaging choices. Other people have noticed this too: https://github.com/coolwanglu/pdf2htmlEX/issues/526...

After installing that it works - do we need a distro specific dependency check for libpoppler-private?

MarPiRK:
Ok then... I have to implement poppler static building. The good thing is that poppler has CMake build system available, so it'll be less work than with boost.

Issue #162, opened 2016-02-18
Handtool gtk3 event compression


cass00:
On my gtk3.16 and gtk3.18 xournalpp installs the handtool is buggy. While dragging, the Xournal widget jumps back and forth rapidly from its original position to some new position. Problem is less severe if event_compression is turned off. From my tests it seems that the ```scrollRelativ``` function causes these back and forth jumps. 

Please do also note that in gtk>=3.16 kinetic scrolling is automatically activated for the touchscreen, i.e. overwriting the hand tool feature.

Issue #160, opened 2016-02-16
pinch-to-zoom


cass00:
One feature I'm dearly missing in xournal(pp) is pinch-to-zoom on the touchscreen. As a first step to implement this feature I've forked the gtk3 branch and followed this commit for the evince pdf viewer GNOME/evince@36c7f21f5c0f96cd5. This works fine and can be tested in this commit cass00/xournalpp@33744e262f43b771e35dd9e72ff568dfcfda73bc. However, there some problems with a centred pinch-to-zoom approach (will make a new issue for that)

wbrenna:
Thanks for the work, that is a nice feature to add to the official repository. If you can fix the centering, remove the unneeded comments, and format it in a way consistent with the original file that would be very helpful! Then just issue a pull request.

PR #165, opened 2016-02-24
Fix for hand tool jitter; pinch-to-zoom

cass00:
Use root coordinates as handtool reference instead of relative coordinates to avoid jittering. (see GNOME/evince@1adce54). Should fix issue #162.
Second commit is the (non-centered) pinch-to-zoom functionality. Issue #160


wbrenna:
Thanks!

PR #215, opened 2017-05-09
Add support for BTN_STYLUS and BTN_STYLUS2

### LINKED ISSUE #214: Add support for BTN_STYLUS and BTN_STYLUS2
Currently only BTN_TOOL_RUBBER is supported as button of digitizers. My pen has two buttons, which are mapped to BTN_STYLUS and BTN_STYLUS2. Could you please add support to bind BTN_STYLUS and BTN_STYLUS2 as tools on the gtk3 branch (I like multitouch to zoom :) )?

oberien:
Fix #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ They'll be at <a href="https://github.com/xournalpp/xournalpp/releases">Xournal+
 
 If at a loss, you can try to build with
 ```bash
-autoreconf
+autoreconf --force --install
 ./configure --enable-mathtex
 make
 ```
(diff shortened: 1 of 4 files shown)


oberien:
For some reason when using [easystroke](/thjaeger/easystroke) there is a bug where GTK3 does not report `GDK_SOURCE_PEN` but `GDK_SOURCE_MOUSE` instead. Therefore if using a digitizer button for gestures one must use the configuration of `middle` instead of the actual button.

PR #247, opened 2018-01-17
Fix up presentation mode

### LINKED ISSUE #91: Add auto-zoom in presentation mode
Automatically zoom the view so that the document becomes slides which fully fit on the page while presentation view is selected. Perhaps this should be part of the Enter Fullscreen option when in presentation mode.

Code will be similar to the auto-center code recently pushed - need to write a new getDisplayWidth() method is all, and check the two values to get a zoom ratio.

WPettersson:
This fixes #91 and also completes presentation mode, at least for me. Happy to take notes or adjust implementation, especially considering the "Things that aren't done" section in the commit message.

I remove any "padding" between or around pages, except for the padding
between dual-page layouts (I don't know how people want to use dual-page
and presentation at the same time, is padding desirable?). Then just
force the zoom to be the appropriate value when in presentation mode,
and skip to next/previous pages as buttons are pressed.

Things that aren't done:
 - Toolbars aren't hiding. I don't actually want this, since I want to
 be able to draw on presentations. Maybe as an option it could be done.
 - Scrollbar is visible. Honestly, I couldn't yet work out how to hide
 it. It should be easy, I feel, so maybe in another commit.
 - The current slide has a red border. I don't mind this, I feel it
 helps highlight the slide against the background. Easy to remove
 though.

ScrollHandler::goTo[Next|Previous]Page()
 - Don't add "top" to add padding when seeing next/previous page

Layout::layoutPages()
 - If in presentation mode, set padding = 0. Also allow zero-width
 margins.

Layout::ensureRectIsVisible()
 - If in presentation mode, don't add extra space around the page.

XournalView::getZoom()
 - If in presentation mode, force zoom to be the appropriate level.

XournalView::onKeyPressEvent()
 - If in presentation mode, left/right also go forwards/backwards in
 presentation. This is useful for things like laser pointers when
 presenting.

XournalView class
 - Add getDisplayWidth/Height functions.

### DIFF
--- a/src/control/ScrollHandler.cpp
+++ b/src/control/ScrollHandler.cpp
@@ -33,7 +33,7 @@ void ScrollHandler::goToPreviousPage()
 				double disHeight = this->control->getWindow()->getLayout()->getDisplayHeight();
 				double top = (dHeight - disHeight) / 2.0 + 7.5;
 				//the magic 7.5 is from XOURNAL_PADDING_BETWEEN/2
-				scrollToPage(this->control->getWindow()->getXournal()->getCurrentPage() - 1, top);
+				scrollToPage(this->control->getWindow()->getXournal()->getCurrentPage() - 1, 0);
 			}
 		}
 		else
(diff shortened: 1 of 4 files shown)


WPettersson:
Btw, CI checks fail due to poppler-somethings, I'm confident it's not related to my changes but feel free to yell at me if I'm wrong

wbrenna:
Thanks very much for the comprehensive patch, and clear details on your changes!

Issue #168, opened 2016-03-22
Preferences do not save


ziekke:
When I change the preferences and close xournalpp, upon reopening they are back to default.

I tried running xournalpp from command line and when I click "ok" in the preferences dialog, the following shows up:

`error : string is not in UTF-8`

I'm using dev branch

edit:

The metadata root was created (~/.xournalpp) including the settings.xml, but I'm supposing something is going on in the app that's preventing it from writing the file properly when the OK button is clicked.

edit2:

I manually edited the file and made a change, once I opened xournalpp it reverted the change back to default.

ziekke:
Without doing too much.. maybe just opening/reopening the program not it is saving!

I swear it wasn't working and I don't believe I changed anything significant.

It is worth noting that the first changes that "held" were when I enabled "Add Vertical Space" and "Add Horizontal Space" and clicked ok. I noted that the error above didn't show up. From then adding more options appeared to have worked.

I did get the app to crash after inserting a largeish LaTeX statement and resizing it.. but I'll have to post that if I get it again.

JoeKays:
This seems to still not work on Debian stretch (testing) fresh install with poppler source compile and boost on default branch.
On first startup saving the settings results in:
`I/O error: encoder error`
`error: string is not UTF-8`
`encoding error: output conversion failed due to conv error, bytes 0x8D 0x60 0x64 0x55`
0x8D is indeed no UTF-8 character, possibly 0xC3 0x8D would have been the real character.
This results in empy settings.xml file, giving a new error on second startup.

SOLUTION: Use a settings.xml file from someone else, e.g. from 3rd comment: https://github.com/xournalpp/xournalpp/issues/187
After using an already complete file, the encoder error disappears, changing settings works as expected.

xiamaz:
Can confirm this issue on the latest master. Getting a similar error message.
Strangely when I tried strace the save was successful.
```
error : string is not in UTF-8
encoding error : output conversion failed due to conv error, bytes 0xCA 0x4B 0x02 0x2C
I/O error : encoder error
```

fkjogu:
I also still have this problem on 3a86f9f86b623c1e3bd53ecd2d8f2ffa27a3ce72

Hoeze:
Same problem here.

kelmes:
Could you try the current development branch?

Hoeze:
[...]

PR #259, opened 2018-03-18
Fix preferences not saving to settings.xml file, both within the application and when restarting it

siliconninja:
Resolves issue #168 and uses pull request #223.
Platform: using Arch Linux, Linux 4.14.27, GTK3 3.22.28-1. Latest pull to the master branch, cherry-picked these fixes to my own repo.

I have tried combining both fixes (Hoeze's `UTF-8` fix from issue #168 and the official pull request to the development branch https://github.com/xournalpp/xournalpp/pull/223) and they seem to work well! Just to be sure, I've deleted the existing settings.xml file to avoid other issues. I also tried it with my existing `ISO-8859-1` encoded file, and it seems to work just fine with the new commits, maybe because the string is being read properly.

I've tried a few different settings (e.g. Xinput, sidebar on the right hand side, and zoom ruler DPI setting) and tested it in the master branch with mathtex flag in Cmake, to be sure all dependencies work. Now the error of "UTF8 string not found" when I click on the edit menu has also disappeared. The preferences are being written properly to settings.xml file in the .xournalpp folder in my home directory.

I'm opening this pull request so that there can be further discussion on this issue and if there is anything else that needs consideration before merging it into master.

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -892,7 +892,7 @@ void Settings::save()
 		saveData(root, p.first, p.second);
 	}
 
-	xmlSaveFormatFileEnc(filename.c_str(), doc, "ISO-8859-1", 1);
+	xmlSaveFormatFileEnc(filename.c_str(), doc, "UTF-8", 1);
 	xmlFreeDoc(doc);
 }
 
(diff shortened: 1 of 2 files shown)


LeoUfimtsev:
Hello,
I ran into this issue where settings are not saved.
I found your patch, checked it out and build the app with your patch applied. My settings are now saved properly.
Fedora 28, Wayland backend.

I would like to see this patch merged.

Issue #260, opened 2018-04-11
Fullscreen + presentation doesn't look right


WPettersson:
I didn't catch this earlier, but when I "fixed" presentation mode it didn't work with full screen, the pages weren't sized right. It seems the layout was updated after the whole window was resized, but before the pages and/or XournalWidget was resized.

WPettersson:
Actually, having used this for a while now, maybe there are more changes we want here. For instance:

Remove scroll bar and red borders in full-screen presentation mode.

Make toolbars auto-hide in full-screen presentation mode. I'm trying to think of a nice way of having the toolbars available when needed here. Maybe they could slide out when the cursor nears the edge - if they do slide out, do we want them to shrink the display when they slide out, or should they appear "over the top" of the display so the page doesn't change scale?

wbrenna:
Thanks, yes - having toolbars, scrollbar, and borders hide in full screen presentation is good. I'd personally rather they don't change scale in presentation mode; I like the idea of a fixed scale in that mode. I'll merge your pull after you make some more changes?

wbrenna:
I intend only to use presentation mode in full-screen.

andreasb242:
On this ticket was already some changes. For me it's not clear what to do now. I close it. For new bugs please open a new Ticket.

PR #261, opened 2018-04-11
Fix fullscreen/resize in presentation mode.

WPettersson:
Enabling full screen first resizes the layout, and then the
display/pages, which breaks presentation mode. Rather than chase this
loop directly, this fix catches the resize event on the widget holding
the pages, and calls layoutPages again, which makes presentation mode
work in full screen, and when resizing the window directly.

It currently uses a static proxy function for the gtk signal handler, I
couldn't get std::bind to work and gtk doesn't support member function
callbacks (obviously, they're C bindings). The std::bind handler
probably could be made to work, or other alternatives.

This fixes issue #260 but I am definitely happy to take feedback into consideration, especially since the proxy static function feels like a hack.


WPettersson:
Also, do we want to only layout pages when in presentation mode? Only connect signal on going into presentation mode, and disconnect when leaving?

Issue #202, opened 2017-01-27
Redo crashes when no action left to redo


niosus:
Redoing actions with <kbd>Ctrl<kbd>+<kbd>Y</kbd> crashes with the following log:
```
addRecentFileFilename: "/home/igor/Desktop/2017-01-27-Note-12-16.xoj"
Info: autosave document...
Info: autosave document...
[Crash Handler] Crashed with signal 11
[Crash Handler] Wrote crash log to: /home/igor/.xournalpp/errorlog.20170127-122430.log
Date: Fri Jan 27 12:24:30 2017
Error: signal 11
[bt]: (0) ./src/xournalpp() [0x6fc059]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x354b0) [0x7fc1ad7384b0]
[bt]: (2) ./src/xournalpp(_ZN12DocumentView9drawLayerEP6_cairoP5Layer+0x16c) [0x6e3aa6]
[bt]: (3) ./src/xournalpp(_ZN12DocumentView8drawPageE7PageRefP6_cairob+0x1c4) [0x6e4bf2]
[bt]: (4) ./src/xournalpp(_ZN9RenderJob17rerenderRectangleEPS_P9Rectangle+0x42a) [0x5ecdaa]
[bt]: (5) ./src/xournalpp(_ZN9RenderJob17rerenderRectangleEP9Rectangle+0xd1) [0x5ecff7]
[bt]: (6) ./src/xournalpp(_ZN9RenderJob3runEv+0x5a4) [0x5ed59e]
[bt]: (7) ./src/xournalpp(_ZN3Job7executeEv+0xd1) [0x5f0139]
[bt]: (8) ./src/xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0x2b9) [0x5f1487]
[bt]: (9) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x70bc5) [0x7fc1aef46bc5]
[bt]: (10) /lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba) [0x7fc1b102d6ba]
[bt]: (11) /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fc1ad80982d]


Try to get a better stracktrace...

Trying to emergency save the current open document
Segmentation fault (core dumped)
```

andreasb242:
Link this together:

https://github.com/xournalpp/xournalpp/pull/284

PR #284, opened 2018-11-07
[BUGFIX] - CTRL-Z deletion recovery segfault

morrolinux:
xournalpp segfaults upon item restore from deletion.

How to replicate the issue:

1) draw something on the page
2) delete it with CTRL+X or `canc`
3) CTRL+Z

I'm not entirely sure why do we have to sum `pos` to the element list.
Anyway the program segfaults upon element restoration due to negative array index.
I've inserted the -1 check and everything seem to work alright.

### DIFF
--- a/src/model/Layer.cpp
+++ b/src/model/Layer.cpp
@@ -82,6 +82,9 @@ void Layer::insertElement(Element* e, int pos)
 		}
 	}
 
+	if(pos == -1) pos=0; 	//otherwise it segfaults on the next step trying
+				//to access a negative index on the elements array 
+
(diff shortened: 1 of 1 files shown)


Febbe:
I don't think this is a good practice to make this check there, since it should not even possible to call 
void Layer::insertElement(Element* e, int pos) with an index out of [0,size] there must be something wrong earlier. I also appreciate to use at least an unsigned int optionally size_t.

andreasb242:
#202 This ticket is probably the same error, I link this now, and check this later.

PR #298, opened 2018-11-12
Fix the compilation of math function.

Rmano:
C++ requires #include <cmath>


andreasb242:
@Rmano Thank you for the fix!

Issue #283, opened 2018-11-06
poppler workaround and build with v 0.71.0


peetCreative:
Hey,
When compiling with poppler 0.71.0 CairoOutputDevice can not be linked anymore.

```
/usr/bin/ld: CMakeFiles/xournalpp-core.dir/pdf/popplerdirect/poppler/XojPopplerDocument.cpp.o: in function `_IntPopplerDocument::_IntPopplerDocument(PDFDoc*)':
/home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerDocument.cpp:27: undefined reference to `CairoOutputDev::CairoOutputDev()'
/usr/bin/ld: /home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerDocument.cpp:28: undefined reference to `CairoOutputDev::startDoc(PDFDoc*, CairoFontEngine*)'
/usr/bin/ld: CMakeFiles/xournalpp-core.dir/pdf/popplerdirect/poppler/XojPopplerPage.cpp.o: in function `XojPopplerPage::render(_cairo*, bool)':
/home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerPage.cpp:96: undefined reference to `CairoOutputDev::setCairo(_cairo*)'
/usr/bin/ld: /home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerPage.cpp:106: undefined reference to `CairoOutputDev::setTextPage(TextPage*)'
/usr/bin/ld: /home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerPage.cpp:125: undefined reference to `CairoOutputDev::setCairo(_cairo*)'
/usr/bin/ld: /home/peetcreative/study/openSoftHardware/xournalpp-golvok/src/pdf/popplerdirect/poppler/XojPopplerPage.cpp:126: undefined reference to `CairoOutputDev::setTextPage(TextPage*)'
```

Recently (pullrequest #265) and in the past (issue #138) it has been discussed that we use in XojPopplerPage CairoOutputDev. I'm not sure why the workaround does not work anymore.

@andreasb242 wrote on #265
[...]

PR #302, opened 2018-11-13
add possibility to build non git version of poppler

peetCreative:
Due to #283 and to play around with different versions of poppler, I added a posiblilty to provide poppler source manually.
Meanwhile I disabled building poppler-qt5, we don't need it.


peetCreative:
I kind of don't get why it does not work with travis.
It stops with the message that openjpeg is not found, which was fixed allready on the main branch....

andreasb242:
Check .travis.yml in the root of the repo, maybe you need to add libopenjpeg2?

PR #311, opened 2018-11-16
Settings fix

andreasb242:


andreasb242:
the chinese translations were merged from pull request which was not mergable, and I mixed this into this pull by accident

PR #342, opened 2018-11-20
Gtk3-merging features

morrolinux:
* CTRL+b to toggle **bold** when in text editor 
* CTRL+ALT+Plus/Minus to change font size on the fly (while in text editor)
* Changed "graph" journal style color form blue to grey (might be useful to give user customization just like background color)
* Go to next/prev. page with right/left arrows (really useful when reading slides, could be nice if customizable by user as well in the future)
* Save a timestamp for each taken stroke when recording, and audio playback feature (audio mapping)


src/control/xojfile/LoadHandler.cpp
```diff
@@ -697,6 +711,12 @@ void LoadHandler::parserEndElement(GMarkupParseContext* context, const gchar* el
 		handler->pos = PARSER_POS_IN_PAGE;
 		handler->layer = NULL;
 	}
+	else if (handler->pos == PARSER_POS_IN_LAYER && strcmp(element_name, "timestamp") == 0)
+	{
+		printf("parserEndElement: timestamp reached - LEVEL: %d\n",handler->pos);
```
andreasb242: Debug code, should be removed

src/control/xml/XmlPointNode.cpp
```diff
@@ -33,6 +55,24 @@ void XmlPointNode::writeOut(OutputStream* out)
...
 
+	/**
+	 * Write timestamp before each stroke.
+	 * This way, when reading, we only need to store 
+	 * 1 timestamp value at a time 
+	 * and assign it to the consequent stroke.
+	 * By adding it this way we don't break 
+	 * xournal's fileformat backcompatibility
+	 */
+	out->write("<timestamp ");
```
andreasb242: May there should be an option to completely disable timestamps

src/gui/MainWindow.cpp
```diff
@@ -29,6 +29,11 @@ using std::cout;
 using std::endl;
 namespace bf = boost::filesystem;
 
+//those two have to be moved somewhere else in the future
```
andreasb242: Generally: Add TODO so they are highlighted

src/gui/XournalView.cpp
```diff
@@ -278,24 +278,24 @@ bool XournalView::onKeyPressEvent(GdkEventKey* event)
 
 	if (event->keyval == GDK_KEY_Left)
 	{
-		if (control->getSettings()->isPresentationMode())
-		{
+		//if (control->getSettings()->isPresentationMode())
```
andreasb242: Why comment out?

toolbar.ini
```diff
@@ -1,69 +0,0 @@
-# Xournal++ Toolbar configuration
```
andreasb242: Why did you delete this, is this not the main toolbar config?

src/gui/XournalView.cpp
morrolinux: It was functional to me, for skipping to the next/prev slide with just the right/left arrow key, without having to be in presentation mode. I guess the best thing to to would be to make it customizable

andreasb242: I didn't ever use the presentation mode, therefore I'm asking...

src/control/xml/XmlPointNode.cpp
morrolinux: Adding a check to write/load them only if a recording has been taken

morrolinux:
It compiles!!!

PR #344, opened 2018-11-21
Formatting

andreasb242:


src/control/xml/XmlPointNode.cpp
```diff
@@ -1,11 +1,11 @@
 #include "XmlPointNode.h"
 
-XmlPointNode::XmlPointNode(const char* tag) : XmlNode(tag)
+XmlPointNode::XmlPointNode(const char* tag)
+ : XmlNode(tag),
+   points(NULL),
+   timestamp(0)
```
morrolinux: is it like setting timestamp=0 ? I've never seen this before

morrolinux:
Damn now I see... I really don't get it. I'm using vscode which (apparently) changed tabs as well. (shouldn't it be 4 spaces?)

src/control/xml/XmlPointNode.cpp
andreasb242: Yes, this is the standard C++ way, I didn't even know that setting in the header file does work at all, seems to be a new C++11 feature (or a G++ Extension)

andreasb242:
You should use Tabs (size = 4 Spaces), Not everything was from you, I also changed small things which was already there.

PR #347, opened 2018-11-21
DEST parameter ignored in make install

youduda:
When running "make DEST=<path> install" only the path set previously with CMAKE_INSTALL_PREFIX ist used.


andreasb242:
I'm not a CMake specialist, so one maybe stupid question, did you test if this is really working?

andreasb242:
Yes, working.

Files are still installed, and I checked the RPATH with:

> objdump -x xournalpp|grep RPATH

PR #352, opened 2018-11-22
Properly install workaround libs

youduda:
Install the poppler libs with native cmake functions without the need of an uninstall workaround (tested) and detect poppler libs to enable -DPOPPLER_GIT_VER without manual adjustments (tested with v0.69.0, v0.61.1).


andreasb242:
You did not remove the uninstall hack from uninstall CMake ;-)
But don't matter, we can do this later or let it be (search for install_manifest_workaround.txt)

PR #355, opened 2018-11-22
Fix settinsg

andreasb242:


src/control/xml/TextAttribute.cpp
```diff
@@ -1,29 +1,24 @@
...
 #include <StringUtils.h>
 
-TextAttribute::TextAttribute(const char* name, const char* value) : XMLAttribute(name)
+TextAttribute::TextAttribute(string name, string value)
+ : XMLAttribute(name),
+   value(value)
 {
 	XOJ_INIT_TYPE(TextAttribute);
-
-	this->value = g_strdup(value);
```
morrolinux: I'm not familiar with this class. Why isn't this needed anymore?

andreasb242: I replaced it with std::string, which is automatically freed.
There where some memory corruption, so I changed this, but I think it was not this issue

PR #386, opened 2018-11-25
timestamps are now stored only in .xopp fileformat

morrolinux:


src/control/xojfile/LoadHandler.cpp
```diff
@@ -532,10 +544,13 @@ void LoadHandler::parseStroke()
 	 * we've read just before. 
 	 * Afterwards, clean the read timestamp data.
 	 */
```
morrolinux: This is for backwards compatibility with the .xoj files still containing audio timestamps

src/control/xojfile/LoadHandler.cpp
```diff
@@ -532,10 +544,13 @@ void LoadHandler::parseStroke()
...
-	this->stroke->setTimestamp(loadedTimeStamp);
-	this->stroke->setAudioFilename(loadedFilename);
-	loadedFilename = "";
-	loadedTimeStamp = 0;
+	if(loadedFilename.length() != 0)
```
morrolinux: This is for backwards compatibility with the .xoj files still containing audio timestamps

PR #401, opened 2018-11-28
fixed build error on clang ( now using C++11 standard )

Seifar:
Small change as seen on https://stackoverflow.com/questions/29743745/c-member-array-initalisation-without-default-constructors fixes small build error on clang (mac os)


andreasb242:
Thank you! If you get it built on Mac OS, please also inform us, e.g. update the Readme or whatever, so other also know, and we may set up an automatically build or something like this.

Seifar:
Got it to build just now. I will start documenting in the following days.

PR #410, opened 2018-11-29
modified cmakelists for building on Mac OS

Seifar:
Now it is building on Mac OS if all dependencies are met.
I am not sure about the ENABLE_NLS and if it is useful on other systems. Some feedback on that would be great.


src/control/XournalMain.cpp
```diff
@@ -19,6 +19,7 @@
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/locale.hpp>
+#undef ENABLE_NLS
```
andreasb242: Does translation still work with this change?
Or do you fully disable translation?

PR #421, opened 2018-11-30
Added real-time LaTex rendering in LatexDialog

luco5826:
Hi!
I was so interested in your project that I wanted to help a little bit!
I liked the LaTex functionality a lot but I tought that a real time display of what a user was 'LaTexting' could have been really handy, so I added it!

Basically I modified only the LaTex Dialog/Controller sources in order to intercept a 'text changed' signal from the Dialog and re-render a new LaTex image while deleting the old one.
I needed to add a white background to the render displayer inside the dialog because for those who use a dark themed DE (as I do), reading the black-fonted Tex would have been really hard.

I tried to follow your code formatting rules and your 'developing pattern' as much as I could, I hope you'll like this feature!

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -11,26 +11,26 @@
 #include <Stacktrace.h>
 
 #include <boost/filesystem.hpp>
+#include "pixbuf-utils.h"
 
 #include <iostream>
 using std::cout;
 using std::endl;
 
-
-LatexController::LatexController(Control* control)
- : control(control),
-   texArea(0),
-   posx(0),
-   posy(0),
-   imgwidth(0),
-   imgheight(0),
-   doc(control->getDocument()),
-   view(NULL),
-   layer(NULL),
-   // .png will be appended automatically => tex.png
-   texImage(Util::getConfigFile("tex").string()),
-   selectedTexImage(NULL),
-   selectedText(NULL)
+LatexController::LatexController(Control *control)
+	: control(control),
+	  texArea(0),
+	  posx(0),
+	  posy(0),
+	  imgwidth(0),
+	  imgheight(0),
+	  doc(control->getDocument()),
+	  view(NULL),
+	  layer(NULL),
+	  // .png will be appended automatically => tex.png
+	  texImage(Util::getConfigFile("tex").string()),
+	  selectedTexImage(NULL),
+	  selectedText(NULL)
 {
 	XOJ_INIT_TYPE(LatexController);
 }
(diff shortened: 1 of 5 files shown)


src/control/LatexController.cpp
```diff
@@ -184,44 +184,87 @@ void LatexController::showTexEditDialog()
...
+ * Text-changed handler: when the Entry in the dialog changes,
+ * this handler updates currentTex, removes the previous existing render and creates
+ * a new one. We need to do it through 'thisContr' because signal handlers
+ * cannot directly access non-static methods and non-static fields such as
+ * 'dlg' so we need to wrap all the dlg method inside small methods in 'thisContr'
+ * 
+ */
+void LatexController::handleTexChanged(GtkWidget *widget, gpointer data)
+{
+	LatexController *thisContr = ((LatexController *)data);
```
andreasb242: add XOJ_CHECK_TYPE_OBJ(LatexController, data);

if there is a nullpointer or a wrong pointer this crashes here. So it's easier to find where the real bug is.

src/control/LatexController.cpp
```diff
@@ -184,44 +184,87 @@ void LatexController::showTexEditDialog()
...
+
+	thisContr->setCurrentTex(gtk_entry_get_text(GTK_ENTRY(widget)));
+	thisContr->deletePreviousRender();
+	thisContr->runCommand();
+	thisContr->insertTexImage(true);
+	thisContr->setImageInDialog(thisContr->getTemporaryRender()->getImage());
+}
+
+TexImage* LatexController::getTemporaryRender(){
+	return this->temporaryRender;
```
andreasb242: Please add 	XOJ_CHECK_TYPE(LatexController); to every method

src/control/LatexController.cpp
```diff
@@ -184,44 +184,87 @@ void LatexController::showTexEditDialog()
...
+}
+
 void LatexController::deleteOldImage()
 {
 	XOJ_CHECK_TYPE(LatexController);
 
 	if (selectedTexImage)
 	{
-		EditSelection* selection = new EditSelection(control->getUndoRedoHandler(), selectedTexImage, view, page);
+		EditSelection *selection = new EditSelection(control->getUndoRedoHandler(), selectedTexImage, view, page);
```
andreasb242: Usually we have the start at the type, not at the var name, as multi var declarations should not be used.

andreasb242:
@luco5826 Thank you!

There are some small issue, but they are not important, because of this I merged your pull already.
If you have time to fix this, create a new pull request.

luco5826:
I'm really happy you liked it!
I'm gonna fix all the stuff you suggested asap!

andreasb242:
Thank you :-)

PR #426, opened 2018-11-30
Fixed macro calls in LatexController methods

luco5826:
Hi!
I added the macro calls you suggested to add and fixed an error in the one called from the handler!


andreasb242:
Thank you!

Issue #143, opened 2015-10-02
Create distro packages with CPack


MarPiRK:
Since we're using CMake it's possible to easily add package builders with CPack.
I'm working right now on RPMs and after that someone with Debian based distro would create DEB.
It would be nice to add CTest building and coverage tests, too.

corna:
I've just tried to create a DEB package and it seems to work. Unfortunately the package does not contains the desktop files and icons: how can I add them? By now I've just added some informations in CMakeLists.txt:

    set(CPACK_PACKAGE_NAME "xournal++")
    set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Notetaking software designed around a tablet")
    set(CPACK_PACKAGE_VERSION "1.0.0-git20151007")
    set(CPACK_GENERATOR "DEB")
    set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Nicola Corna")
    
    include(CPack)

and ran cpack.

MarPiRK:
Yeah, I know it. Desktop files are added in `desktop_install` target. It needs to be moved to main install target.

corna:
I agree, Xournal++ is a graphical software and there's no need of having separate `install` and `desktop_install` targets

MarPiRK:
Unfortunately to properly install `xournal-thumbnailer` it's needed to run small shell script.
btw, xournal-thumbnailer isn't working for me. Maybe it's time to look up what's wrong with this script.

wbrenna:
Good idea, I've never actually used the thumbnailer...

MarPiRK:
Yet another reason to do something about `desktop-install` – it's violating some rules for Gentoo ebuilds.

PR #431, opened 2018-12-01
Debain packaging working

andreasb242:
#143


JonasDralle:
Can't locate the debian package...

LittleHuba:
See https://github.com/xournalpp/xournalpp#ubuntu-and-derivates

JonasDralle:
installing PPAs under Debian is [considered unsafe](https://wiki.debian.org/DontBreakDebian#Don.27t_make_a_FrankenDebian)

LittleHuba:
That would mean we have to support multiple debian packages independently on our own. See #959 for further details why we will not do it on our own.

JonasDralle:
installing PPAs under Debian is [considered unsafe](https://wiki.debian.org/DontBreakDebian#Don.27t_make_a_FrankenDebian)

PR #448, opened 2018-12-03
Fix installing files

youduda:
Some files are installed ignoring CMAKE_INSTALL_PREFIX and two files were missing.


peetCreative:
Does the  last x-xopt.desktop install works? I suspect it shouldn't because of the wrong name.

youduda:
What still could be improved is the redundant CMAKE_INSTALL_PREFIX in the install destination, but on the other hand, this way things are clearer therefore I left it unchanged.

andreasb242:
@youduda 
Thank You!

I knew it was not perfect, probably still not. But I had to remove the shell script, so packaging would work (even, not perfect yet, but a package with all files is created now)

PR #453, opened 2018-12-03
Improve CPack package generation and version number

youduda:
The second commit is only a proposal to change the version number to a valid one, as installing deb packages with the previous version is not possible. The version style is adopted from kodi (former xbmc), but of couse I can remove that part again.
By the way, their code could speedup implementing ppa support as their license match the xournalpp license.


andreasb242:
Thank you. Did you already create a debian Package?
For me its the first time, I createt the PPA yesterday, and the Launchpad Project, but it's a lot of work, and the custom Poppler Build is an issue (I may try to remove some PDF Features for this case)

PR #457, opened 2018-12-04
Fixed Error when launching LatexDialog when tex.png is missing

luco5826:
Hi!
I found out the problem which caused an error when LaTex dialog was lauched: basically I previously put `insertTexImage(true)` in the `LatexController->showTexEditDialog()` but when that method is executed, there isn't the tex.png image yet, so I removed it because the first effective tex.png file is created right when the Dialog appears, because a "changed" signal is launched at its creation!
I also added `.vscode` to the gitignore and raised the `POPPLER_MAX_VERSION` to 0.69 since it works!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ src/pdf/popplerdirect/workaround/workaround.sh
 # NetBeans
 nbproject/
 
+#Visual Studio Code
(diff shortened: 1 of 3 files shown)


andreasb242:
Thank you!

PR #461, opened 2018-12-05
Feature: show/hide Toolbars

luco5826:
Hi guys!
I saw that an issue marked as enhancement talked about the ability to hide and show the Toolbar, so I added it, now it can be done from the View menu or by presing F9.
It should work just fine, I also updated the Settings in order to save the state of the Toolbar after closing Xournalpp, but I don't understand one thing: when I de-select "Show Toolbar", the Toolbar disappear, but when I re-enable it, it appears with something like a frame around the document.
I don't know where the issue could be, I did simply hide/show widget on the Toolbar's widget!
Right now it works well, if someone knows how to fix that little issue... that'd be great!

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -60,6 +60,8 @@ void Settings::loadDefault()
 	this->mainWndWidth = 800;
 	this->mainWndHeight = 600;
 
+	this->showToolbar = true;
+
 	this->showSidebar = true;
 	this->sidebarWidth = 150;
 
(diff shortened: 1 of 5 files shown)


morrolinux:
Hi Luca,
I'm not sure if I understood what do you mean by "Hide the Toolbar":
I used to do it just fine via "view -> toolbars -> Empty Toolbar" 
see: https://imgur.com/a/7l16qAi
isn't this already a feature?

luco5826:
Awesome! Eheh didn't know about that.
In that case, I only added the F9 shortcut and the ability to hide or show the current toolbar without changing the Toolbar "preset", I understand if you don't think it could be mergeable!

morrolinux:
I am really sorry that you spent precious time doing something that was "already there" 
I also think code complexity should be kept at minimum when not needed.. I would normally advise against this merge, but we don't know yet what @andreasb242 and the others think of it.
Or maybe you could revert part of your changes and change the logic to switch back and forth from/to the current toolbar setting and the empty toolbar? What do you think about it? is it a bad idea?

luco5826:
No no! That is totally a good idea! I also thought about that change right after you told me the feature was already there!
No problem about my time, it was mu fault
I'm gonna fix it!

luco5826:
Thank you! I'm so glad to be in this project!

andreasb242:
For hiding Menubar I think F10 is the Standard key.

Hedronmx:
The standard key is actually the alt key

andreasb242:
Yes, you are right. Alt is the Menu key.
But not sure if this is easy possible, if not, F10 toggles also the menu on some applications.

Hedronmx:
No problem, a preference option could also work :)

PR #466, opened 2018-12-05
Add missing math header.

PellelNitram:
Upon compiling the most recent `master` branch, I found that the compilation found due to a missing header.

The call to `atan` in line 465 of the function `EditSelection::mouseMove()` caused the crash as the `math` header for `atan` was missing.

### DIFF
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -23,6 +23,7 @@
 #include <serializing/ObjectInputStream.h>
 
 #include <iostream>
+#include <math.h>
 using std::cout;
 using std::endl;
 


morrolinux:
Strangely enough, I cannot reproduce this issue. 
I think we can merge this though. 
Still, I'm not sure if it would be better <cmath> rather than <math.h>

PellelNitram:
That's strange! Is there any information I can provide about my build setup? (like gcc version etc?)

> Still, I'm not sure if it would be better  rather than <math.h>

What do you mean by that?

morrolinux:
Oh sorry, it seems like GitHub's markdown doesn't play well with angular brackets. I edited my comment.
Yes please, can you attach your compiler log along with its version?
What's even more strange is that even though the math library is not present, the build did work on Travis as well

PellelNitram:
This is my GCC version (with `gcc --version`):

> gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0
> Copyright (C) 2017 Free Software Foundation, Inc.
> This is free software; see the source for copying conditions.  There is NO
> warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

How do I attach a compiler log? (I did google, but haven't found anything).

> What's even more strange is that even though the math library is not present, the build did work on Travis as well

Is maybe any other of the included headers loading the library?

morrolinux:
> How do I attach a compiler log? (I did google, but haven't found anything).
> 
> >
You can copy/redirect the shell output and get it on pastebin.



> Is maybe any other of the included headers loading the library?

Probably. But I haven't had a good look at the code yet.

PR #467, opened 2018-12-05
Code documentation

PellelNitram:
I add Doxygen to generate a very basic HTML code documentation.

Everything is documented in `README.md` and I did one exemplary documentation of the class `Point()`.

A Doxygen file has many options. However, I set only the most basic ones so that further tweaking might be required.


andreasb242:
I know I merge the math.h include also, even if cmath would be better, more Documentation is better :-)

PR #471, opened 2018-12-06
rotation snapping every 45 degrees

morrolinux:
This will allow to have the rotation snapping to a grid every 45 degrees. 
Maybe we could use a checkbox by the dropdown menu to enable/disable this feature in the future.

### DIFF
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -319,6 +319,8 @@ void EditSelection::mouseUp()
 	PageRef page = this->view->getPage();
 	Layer* layer = page->getSelectedLayer();
 
+	snapRotation();
+
 	this->contents->updateContent(this->x, this->y, this->rotation, this->width, this->height, this->aspectRatio,
 								  layer, page, this->view, this->undo, this->mouseDownType);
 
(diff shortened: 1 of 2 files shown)


luco5826:
Awesome!
Maybe we could enable it by pressing CTRL while rotating it! I don't know how hard could it be implementing it but it would be perfect imho

src/control/tools/EditSelection.cpp
```diff
@@ -639,6 +641,43 @@ CursorSelectionType EditSelection::getSelectionTypeForPos(double x, double y, do
 	return CURSOR_SELECTION_NONE;
 }
 
+void EditSelection::snapRotation()
+{
+	double epsilon = 0.1;
+	if(std::abs(this->rotation) < epsilon)
```
andreasb242: do this values to a const double, and then do a loop, like:

const double ROTATION_LOCK = {0, M_PI / 2.0, M_PI ....};
....
for (int i = 0; i < sizeof(ROTATION_LOCK) / siezof(ROTATION_LOCK[0]); i++)
{
   if (std::abs(ROTATION_LOCK[i]) < epsilon)
      this->rotation = ROTATION_LOCK[i];
}

(yes, this formatting is not OK ;-))

morrolinux:
@luco5826 Yeah, but what if I'm not using a keyboard? :)
I often use my tablet this way when taking notes, and a dropdown setting from the tool menu would be useful in this scenario

src/control/tools/EditSelection.cpp
andreasb242: Hold down keys (CTRL, ALT, SHIFT) are within the mouse event, I did it wrong on my last change. You don't need key listing.

luco5826:
@morrolinux Yep you're right, didn't think about that

src/control/tools/EditSelection.cpp
morrolinux: > Hold down keys (CTRL, ALT, SHIFT) are within the mouse event, I did it wrong on my last change. You don't need key listing.

I don't get what your're talking about here

morrolinux:
Well... I don't really get how to check if shift is being hold down from within `EditSelection`
I may be too tired for today. @luco5826 feel free to contribute on that if you find it meanwhile :)

morrolinux:
> Additional (more work, not so easy) there should be a toolbar button, to switch this generally, for rotationg selection and maybe other which I currently forgot.

I think I'm gonna go for this. but maybe in a separate PR as it involves other parts too

morrolinux:
did it in #474

PR #473, opened 2018-12-07
Arrow rotation

morrolinux:
* Snap the arrow to a grid for rotation, just like the normal "segment" does
* ATM only each 90 degrees since I haven't figured out yet how to rotate the arrow head by 45 (any help on this would be nice)


morrolinux:
Implemented arrow rotation every 45 degrees

PR #474, opened 2018-12-07
add rotation snapping toolbar button

morrolinux:
* Snapping can now be enabled/disabled by `ROTATION_SNAPPING` toolbar button
* You can test it out with the `Ruler` tool: if it's set the line will snap every 45 deg. if not, it won't.
* I do realize the svg icon sucks, maybe some of you will come up with a better idea for representing it


src/control/Control.h
```diff
@@ -273,6 +276,7 @@ class Control : public ActionHandler,
 	UndoRedoHandler* undoRedo;
 	ZoomControl* zoom;
 	bool fullscreen;
+	bool snapRotation;
```
andreasb242: Initialization missing!

Issue #470, opened 2018-12-06
Crash on TextTool


luco5826:
Hi! I don't know if it's only me but I select the TextTool from the toolbar and add a Text box to my document; as I type Xournal++ crashes.
Do you experience the same issue?

andreasb242:
No, but do you have an exception log? it should be in your home, .xournalpp

is it reproducible on your system?

luco5826:
Yes, every time I add a Text box, right in the moment when I press a letter on my keyboard, it crashes. 
I would wait to create the PR about the Pen cursor in order to be sure that my changes didn't cause it, even if I changed only one file in a small portion

[errorlog.20181206-210729.log](https://github.com/xournalpp/xournalpp/files/2654677/errorlog.20181206-210729.log)

andreasb242:
You may set a breakpoint in TextEditor::onKeyPressEvent and then debug. I don't think this has something to do with the cursor... But yes, you can never be sure, It's C++ ;-)

luco5826:
I checked it before! 
I think that is 
`gtk_im_context_filter_keypress(this->imContext, event)` the point giving troubles

luco5826:
Hi @andreasb242 ! I found a workaround, and as I thought, the problem is not in Xournal++.
Googling around I found this very similar issue
[](url)https://github.com/p-e-w/finalterm/issues/364
and his fix worked for me!
As my debugging process found out, the problem is in `gtk_im_context_filter_keypress` but I don't know why this happens, neither how to permanently fix it!

andreasb242:
unset GTK_IM_MODULE fixes this issue, as a workaround.

What is then your value before you unset it?

mine is ibus.

$ echo $GTK_IM_MODULE
ibus

luco5826:
Mine is `xim` , but I sincerely don't know what that means, I'm on Debian Buster (testing) with 4.18.0-3-amd64

luco5826:
And now if I launch Xournal++ with
`GTK_IM_MODULE=ibus ./src/xournalpp ` it works!

andreasb242:
And if I set it to xim it crashes!

`GTK_IM_MODULE=xim xournalpp`

So this is not solved yet. If we cannot fix it, we should at least show an error to the user, that we don't support this configuration!

andreasb242:
https://unix.stackexchange.com/questions/260601/understanding-setting-up-different-input-methods

[...]

PR #492, opened 2018-12-08
Added Pen icon when in BigCursor mode

luco5826:
Hi! 
As the title says, i changed the icon of the cursor when in Big Cursor mode in order to make it look like a pencil.
I still have that problem in Text Box as in #470 , but even checking for imContext or event for NULL values didn't solve the issue. My debugger found a Segmentation Error in the 
`gtk_im_context_filter_keypress(this->imContext, event)` condition.


andreasb242:
Here is definitive no risk - so I merge it

Issue #478, opened 2018-12-07
Rescale LaTeX editor/preview 


morrolinux:
When writing "long" latex parts (> 10 words) the preview starts to get too big to fit the screen, and the editor is no longer usable (see the screenshot: https://imgur.com/a/xhbVNtc)
It would be nice for it to scale and stay under a certain size (and maybe even have a multiline textbox)

morrolinux:
@luco5826 Do you think you can acheive that?

luco5826:
@morrolinux I could totally try!
I should modify the LatexDialog and search for an option in mathex for the preferred output size!

andreasb242:
You can scale it wit cairo. Code hints (not compiling, but in general complete, copied from Xournal++ source)

```
	this->surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32,
											   width * this->pngDpi / 72.0,
											   height * this->pngDpi / 72.0);
	this->cr = cairo_create(this->surface);
	double factor = this->pngDpi / 72.0;
	cairo_scale(this->cr, factor, factor);

	cairo_set_source_surface(cr, ORIGINAL_GRAPHIC, 0, 0);
	cairo_paint(cr);
	cairo_destroy(this->cr);

```

morrolinux:
So.. It is better! but I would have scaled it more (or prevented it to go over a certain size)
See how it's changed on my previous example: https://imgur.com/a/srVKd3D
Do you have it significantly smaller? maybe you're on a HiDPI screen?

morrolinux:
Ok, I just realized that that happens when I edit an existing latex.
If I create a new one from scratch and paste the same text, here's the result: https://imgur.com/a/YCxY09C
Which is just perfect.
So I guess it's fine overall :)
Thank You! @luco5826

luco5826:
@morrolinux Awesome! 
I'm so happy you found it useful. It still needs some tweaks, including the multi-line editor which is harder then I thought, but when I'll have that implemented i'm gonna focus on the scale factor in order to optimize the size!

andreasb242:
g_source_remove(this->rescaleId); to remove it from queue and g_timeout_add with a timeout from 500ms or so on each key press.

So if the user 500ms does not type, the preview will be updated

luco5826:
Hi @andreasb242 sorry to bother you but I'm struggling with the timeout
```
void LatexController::handleTexChanged(GtkTextBuffer* buffer, LatexController* self)
{
[...]

PR #504, opened 2018-12-09
Fixed LaTex scaling for long text

luco5826:
Hi!
I managed to downscale the LaTex preview when long strings are inserted.
I still need to change the Dialog entry from a simple one-line entry to a small TextView (is this the right way?), but for now let me know if the scaling works a little better!
(I didn't change .po files, I don't know why they're different)

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: xournalpp 1.0.1\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-12-09 10:10+0100\n"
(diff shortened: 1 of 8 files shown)


andreasb242:
I think you have to merge against the master, then the diffs of the .po files should be gone.

luco5826:
Hi @andreasb242 sorry but I don't understand what should I do! I pushed my changes as I always do

andreasb242:
I thought you are not up-to-date, but you are.
The changes are OK, I did it today, but I don't understand why they are displayed in your pull...

Ignore them, if something will go wrong, this are only two texts... So not a problem, but I don't think so.
If you have tested the scaling, you can merge it.

morrolinux:
fixes #478

PR #517, opened 2018-12-11
Cleanup and Windows Build

andreasb242:


andreasb242:
@luco5826 You can still use Boost string, its used on a lot of places, removing it don't make sense.

luco5826:
Awesome! I knew you wanted to speed up the build process by removing unused libraries, so I thought that you wanted to remove the boost/string.hpp from the project! But if you say that it's still used in other sources it's fine!

Issue #489, opened 2018-12-08
png export with transparent background


morrolinux:
It would be nice ti have an option to export in png format with transparent background

Issue #490, opened 2018-12-08
PDF Export without paper style


morrolinux:
It would be nice to be able to export pdf(s) without paper style (e.g: dotted, graph, lined, ruled, plain..) 
I've thought about two ways of doing this:
* Create an option to apply a chosen paper style to all pages at once, and export with "plain" style applied
* Create a checkbox to decide weather the paper style should be exported or not

andreasb242:
Page Background: #69

I first need to fix the travis build for the new dependencies...

Rmano:
Yep. sorry, I've been cryptic. Suppose you want to prepare an overlay page with some data and maybe a signature with `xournalpp`, and then "stamp" it on a series of documents... I prepared an example, see this zip file: 
[stamping-pdfs.zip](https://github.com/xournalpp/xournalpp/files/2660905/stamping-pdfs.zip)



So, I have prepared a signature with `xournalpp` (see the file `signing.xopp`); 

![image](https://user-images.githubusercontent.com/6414907/49701606-5912e980-fbee-11e8-9129-8fb101a17cef.png)

then I use for example `pdftk` (available as a snap at https://snapcraft.io/pdftk) and I issue:

```
pdftk lilla.pdf stamp signing.pdf output stamped.pdf
``` 

![image](https://user-images.githubusercontent.com/6414907/49701613-73e55e00-fbee-11e8-8508-2ebfddd031c2.png)

as you can see, the result has the "signing" with the white background in it. 

I can solve the problem opening the `signing.pdf` file with `inkscape`, ungrouping everything, and delete the blank page underneath the signature (the resulting file is `signing-inkscape.pdf`); now when "stamping" it over the same file: 
```
pdftk lilla.pdf stamp signing-inkscape.pdf output stamped-inkscape.pdf
```
the result is quite different:

![image](https://user-images.githubusercontent.com/6414907/49701618-8790c480-fbee-11e8-903f-6117cb24cb53.png)

My idea is that it would be great to have a "transparent" or "no background" page, so that the resulting PDFs could be used to stamp over forms...

morrolinux:
This should be done via PNG export with transparent background, I've opened another ticket about it

andreasb242:
We paint always a Background to the PDF.
@morrolinux you would anayway a PDF export without Background, this may solve this issue as well without any additional work.

PR #525, opened 2018-12-13
PDF and PNG export with no background

morrolinux:
DO NOT MERGE JUST YET
* Export PNG with transparent background (via "export as")
* Export PDF with plain background (same)
* Fixing first page missing when doing "Export as" -> "PDF"


src/control/jobs/CustomExportJob.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include <i18n.h>
 #include <config-features.h>
+#include <iostream>
```
morrolinux: to be removed

src/control/jobs/PdfExportJob.cpp
```diff
@@ -61,11 +61,14 @@ bool PdfExportJob::isUriValid(string& uri)
 	return true;
 }
 
+
```
morrolinux: to be removed

src/view/background/TransparentBackgroundPainter.cpp
```diff
@@ -0,0 +1,39 @@
...
+	this->lineWidth = 1.5;
+	this->drawRaster1 = 14.17;
+}
+
+void TransparentBackgroundPainter::paint()
+{
+	XOJ_CHECK_TYPE(TransparentBackgroundPainter);
+
+	//paintBackgroundColor();
+	//paintBackgroundDotted();
```
morrolinux: to be removed alongside with the definition

src/control/jobs/CustomExportJob.h
```diff
@@ -17,6 +17,15 @@
 
 #include <PageRange.h>
 
+class Document;
+class PageType;
+
+const string EXPORT_PDF = "PDF files";
```
andreasb242: This strings needs to be translatable => _()
Static seems to be not working, so it should be intialized in the constructor

src/control/jobs/CustomExportJob.cpp
```diff
@@ -268,15 +280,19 @@ void CustomExportJob::run()
...
+		{
+			pdfe->setNoBackgroundExport(true);
+		}
 
 #ifdef ADVANCED_PDF_EXPORT_POPPLER
 		// Not working with ADVANCED_PDF_EXPORT_POPPLER
 		if (!pdfe->createPdf(this->filename))
 #else
-		if (!pdfe->createPdf(this->filename, exportRange))
+		if (!pdfe->createPdf(this->filename/*, exportRange*/))	//export range is called anyways
```
andreasb242: With this you probably break the Export with only some Page ranges (not PDF Export, export and Select PDF), did you test that?

src/control/jobs/CustomExportJob.h
morrolinux: Oh but I did that :) 
here:
`addFileFilterToDialog(_C(EXPORT_PDF), "*.pdf");`

src/control/jobs/CustomExportJob.cpp
morrolinux: Ok then there's something wrong... if I leave there `exportRange` an try to export, the output pdf is empty

morrolinux: Oh, no, ok.. there must be an index bound problem because the first page is always absent when exporting all pages

src/control/jobs/CustomExportJob.h
andreasb242: This will not work. Then remove the constants, and add the Text directly, gettext will not resolve constants.

andreasb242: The translation will work, but the source Text will not appear in the Translation editor. Thats what I mean

src/control/jobs/CustomExportJob.cpp
andreasb242: The first page has the Index 0
But the user enters page 1, because a normal user starts counting at 1, not 0. Only computer Scientists do :-)

morrolinux:
Fixes #489 and #490

PR #530, opened 2018-12-14
 Added Input method check (xim unsupported) 

### LINKED ISSUE #470: Crash on TextTool

luco5826:
Hi! I added a check on the env variable GTK_IM_MODULE which, if set to "xim", caused the Text tool to crash
Fixes #470


andreasb242:
Thank you!

PR #535, opened 2018-12-15
New input handling

andreasb242:
Should be working mostly
What I plan to do:
* Add a configurable timeout how long the pen disables the touchscreen.
* Disables touch gestures (and scrolling, need a good workaround)

Additional add a timeout handling, if there is no action on an input, release the lock for it, else no input will be possible anymore. So if we miss an event, input will not be possible anymore.

The input is now handled by device, therefor this is needed.


src/gui/Layout.cpp
```diff
@@ -50,12 +50,7 @@ void Layout::checkScroll(GtkAdjustment* adjustment, double& lastScroll)
 {
 	XOJ_CHECK_TYPE(Layout);
 
-	if (!view->shouldIgnoreTouchEvents())
```
andreasb242: This does issues, therefore I removed it, I plan to add something like this, but I need a better solution.

src/gui/PageView.cpp
```diff
@@ -619,44 +619,15 @@ bool XojPageView::onMotionNotifyEvent(GtkWidget* widget, GdkEventMotion* event,
 	return false;
 }
 
-void XojPageView::translateEvent(GdkEvent* event, int xOffset, int yOffset)
```
andreasb242: This is now done in the event handling classes

src/gui/XournalView.cpp
```diff
@@ -588,33 +576,6 @@ GtkContainer* XournalView::getParent()
...
-{
-	XOJ_CHECK_TYPE(XournalView);
-
-	this->lastPenAction = g_get_monotonic_time() / 1000;
-}
-
-/**
- * If the pen was active a short time before, ignore touch events
- */
-bool XournalView::shouldIgnoreTouchEvents()
```
andreasb242: this needs to be re-added in a better way

PR #545, opened 2018-12-16
update README for features

morrolinux:


morrolinux:
I Would also suggest to maybe change the screenshot with a new one featuring the latest features as well (maybe with "Toolbar Left" layout which already has them set) :
image

andreasb242:
Why only the toolbar left? Is not enough space on the other layouts?

I would prefer to add multiple Screenshots, I would also like to add a Windows Screeenshot when the packaging is done.

morrolinux:
Yeah why not!

PR #557, opened 2018-12-17
Add OpenSuse library instructions

nielstron:
I just installed it on my system and it works like a charm!

Only difference for me: `sudo make install` was necessary instead of the proposed `make install`


andreasb242:
@nielstron Thank you!

PR #566, opened 2018-12-22
Fixed some make warnings

luco5826:
Hi! 
I fixed some compilation warnings generated from deprecated methods. I also switched from `gtk_menu_popup` to `gtk_menu_popup_at_widget` because of the cleaner code and because I saw that only Ubuntu 16.X and earlier doesn't support GTK 3.22, so I thought this could be a good improvement! What do you think?

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -871,7 +871,7 @@ void Control::help()
 {
 	GError* error = NULL;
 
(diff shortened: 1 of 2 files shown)


andreasb242:
Our build is currently Running on Ubuntu 16.04, and Travis does not yet offer Ubuntu 18.04 (or I didn't find it)

PR #570, opened 2018-12-22
Concurrent job in travis CI

luco5826:
I always build Xournal++ on my system with `-j5` make parameter in order to speed up the compilation. I wanted to check if we get any improvement on Travis CI build time


andreasb242:
The Build is running in a Docker container on a shared system, it's really hard to test timing... But yes, you are right. for testing you may add the command cat /proc/cpuinfo, to see how many CPUs are available

May have a look at:
https://stackoverflow.com/questions/2527496/how-can-i-write-a-makefile-to-auto-detect-and-parallelize-the-build-with-gnu-mak

andreasb242:
This is only for Travis relevant, so if your build is working, it's correct.

Issue #416, opened 2018-11-29
Zooming with the slider changes page


peetCreative:
When I zoom with the slider the location in the document is changing to other page. Zooming back brings me back to the correct page.

peetCreative:
This seems to be a bug with the automatic gtk zooming..

andreasb242:
There is no GTK Zooming, this is a Custom Xournal++ implementation.

peetCreative:
mh... That's right.... Took me a while digging through the code. I suspect `Layout::layoutPages` and subsequent (`this->view->pagePosition->update(this->view->viewPages, len, height)`) to make the changes. 
Still don't completly understand what is going on there. The function is called three times, when pressing the ZoomIn, ZoomOut or two times when the slider is used. This is probably not the bug, but seems unefficient? 
One may also rewrite such function with std libraries, as it gets easier to understand is, or?

andreasb242:
> This is probably not the bug, but seems unefficient?
Yes, probably a Bug, but I sometimes it's not easy to get rid of this, and sometimes it's really needed (e.g. auto layouting boxed layouts)

>One may also rewrite such function with std libraries, as it gets easier to understand is, or?
About Which standard Libraries are you thinking about?

andreasb242:
Layout.cpp is mainly the layouting of the pages. It calculates the position of each page, depending on the Zoom and the layout options, like two pages.

The Layout itself does not have any container, the container is in XournalView, and its only an array:
	XojPageView** viewPages;

The whole Displaying is done by GTK, also which part are rendered or not.
The painting is done in gtk_xournal_draw:

Iterate over all Pages
>	while (it.hasNext())
>	{
...

If the page is not visible at all, skip it...
>		if (!clippingRect.intersects(pv->getRect()))
>		{
>			continue;
>		}

...

Paint a single page, the page is cached as image, and rendered in a background Thread.
We draw the whole page, but technical over Cairo / GTK only the part which is needed, is painted.
But this is all handled by GTK
>		pv->paintPage(cr, NULL);

>	}

But this is mostly not relevant for this bug.
The issue is relative simple:

If the Zoom changes, the surface size increases:
>	this->setLayoutSize(width, height);

So the Scrollbar gets updated (in Background by GTK)
But I think the Scrollbar keeps the current position. But the Page is not at this position anymore, because the page is bigger or smaller.

Therefor the solution would might be:
Before change the zoom, which page at which position is at the center of the screen vertically.
[...]

PR #576, opened 2018-12-22
fix the page is scrolling to the wrong position

peetCreative:
Fixes for me issue #416

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -22,8 +22,8 @@ ZoomControl::ZoomControl()
 
 	this->zoomWidgetPosX = 0;
 	this->zoomWidgetPosY = 0;
-	this->scrollPositionX = 0;
-	this->scrollPositionY = 0;
+	this->scrollPositionX = -1;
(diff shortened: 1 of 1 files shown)


andreasb242:
For me the Zoom is not not working correct anymore, neither by touch nor by mouse.

@luco5826 or @morrolinux can you recheck?

andreasb242:
@peetCreative Is this really working for you? How did you test this?

Even with the mouse and with the touch it doesn't work.
If I do a review, I look at code, like possible crashes, something which looks really wrong or so, and I probably don't see a bug.

If I analyze it now in detail, the only relevant change you did is:
zoomWidgetPosX and zoomWidgetPosY is scaled.

But this are coordinates which are on the widget side, not on the document side, therefore are independent of the zoom.

With your change now: the document moves down/right if I zoom in, or up/left if I zoom out.

luco5826:
I can only test by zooming via CTRL+Scroll and it always zooms to the top left corner of the document

andreasb242:
Ok. Thank you. I'll revert this commit.

PR #580, opened 2018-12-22
Removed env: GTK=3.22.30

luco5826:


andreasb242:
You may check your mails?
There are enough really old GTK functions which can be fixed. Have a look at existing builds.

PR #591, opened 2018-12-23
Remove global var

andreasb242:


andreasb242:
@morrolinux I fixed the global var issue. May you have a look at it, so you know what I did.

Issue #521, opened 2018-12-11
Export as Dialog don't change extension


Hedronmx:
When using "Exporting as ..."  If I change the format on the dropdown, the file extension doesn't change. 
The default extension is used which is .xopp
![Imgur](https://i.imgur.com/jnHYl1O.png)

I chose PNG but the file is being saved as .xopp

luco5826:
Hi!
Confirmed on my system, right now the only way is to set PNG graphics on the dropdown menu AND manually changing the extension from .xopp or .pdf to .png
We are gonna fix it!

luco5826:
@andreasb242 I found a problem:
when using the Export As.. dialog, Xournal++ does its checks only on the extension name, so we could set in the ComboBox PDF Document but, if we name the file with .png it will be exported as PNG.
So I suggest to modify the workflow, instead of checking on the extension after clicking on Export, we check on the value of the ComboBox appending the relative extension to the file name overwriting the one in the text entry
What do you think?

morrolinux:
I did figure this out while trying to add some export tweaks (like "export with no background"):

We need to get the current filter from `BaseExportJob`:
```
string BaseExportJob::getFilterName()
{
	XOJ_CHECK_TYPE(BaseExportJob);
	
	GtkFileFilter* filter = gtk_file_chooser_get_filter(GTK_FILE_CHOOSER(dialog));  //gets the filter
	return gtk_file_filter_get_name(filter);   //gets the name (you can get the extension as well)
}
```

And we can call this method somewhere when the dialog hasn't been destroyed yet, like, for example, in `PdfExportJob::isUriValid`:

```
string filterName = BaseExportJob::getFilterName();
std::cout<<"FilterName: "<<filterName<<"\n";
```
You get the choosen filter name printed out

morrolinux:
Next challenge will be to find a way to call it each time the filter has changed

morrolinux:
I think it would be acceptable :) what do you think @andreasb242 ?

andreasb242:
Yes, for me would this be OK, but you have to think about the "Do you want to overwrite" check, which you may need to reimplement.

I checked how GIMP handle this: They add additional to the filter a Type selection.
![grafik](https://user-images.githubusercontent.com/822958/50348893-63ac7780-053a-11e9-81ac-059ac15d9832.png)

If you don't add an extension, and don't select an extension, they don't allow to save you the image.
This would be the easiest solution.
[...]

PR #594, opened 2018-12-23
Workaround on Export extension

luco5826:
Hi!
I managed to solve issue #521 by making the final file format depending only on the selected filter by the ComboBox, I don't know if this is an improvement but at least a major portion of code is way easier now, because I also deleted un-called methods.
Problem is, right now we add the extension AFTER the user clicks on Save/Export, so the "confirm overwrite" event in the GtkFileChooser doesn't get triggered, so I added a method (`checkExistingFiles`) which check if the file exists and then warn the user, but doesn't give the chance to replace the existing file. 
I think we could add something like a YesNoDialog to Utils class in order to reproduce the "Do you want to replace the file?" popup. 
Do you think this is a bad idea? Maybe we could find another solution?


src/control/jobs/CustomExportJob.cpp
```diff
@@ -47,11 +47,11 @@ void CustomExportJob::addFilterToDialog()
 {
 	XOJ_CHECK_TYPE(CustomExportJob);
 
-	addFileFilterToDialog(EXPORT_PDF, "*.pdf");
-	addFileFilterToDialog(EXPORT_PDF_NOBG, "*.pdf");
-	addFileFilterToDialog(EXPORT_PNG, "*.png");
-	addFileFilterToDialog(EXPORT_PNG_NOBG, "*.png");
-	addFileFilterToDialog(EXPORT_XOJ, "*.xoj");
+	// Runs on every filter inside the filters map
```
andreasb242: Yes, much better than before!
Only one thing: C++11 also supports foreach on maps:
https://stackoverflow.com/questions/6963894/how-to-use-range-based-for-loop-with-stdmap

for (auto& filter : filters)

luco5826:
I just found out that simply saving a file doesn't add the .xopp extension. I'm gonna search for the problem

src/control/jobs/CustomExportJob.cpp
```diff
@@ -189,7 +186,7 @@ void CustomExportJob::exportPngPage(int pageId, int id, double zoom, DocumentVie
 		PdfView::drawPage(NULL, popplerPage, cr, zoom, page->getWidth(), page->getHeight());
 	}
 
-	if (this->chosenFilterName == EXPORT_PNG_NOBG)
+	if (this->chosenFilterName == _("PNG with transparent background"))
```
andreasb242: It would be nicer to add the values to the map, so for value use e.g. 

class  ExportType {
public:
  string extension;
  bool transparentBg
}

or something like this. Because if the texts are updated on one place, and somebody don't know he has to update it on a second place, it will fail.

andreasb242:
I reopened, because github don't allow me to submit my review on a closed Pull request ;-)

Generally I think it's good, I added some points on the code, nothing dramatic.

On Thing: I think the option "By extension" like GIMP has it, would be a good idea, as you have the extension in the map, you can search for it. But you can do this in an additional pull.

andreasb242:
I would add a nested class into BaseExportJob and make a constructor which initialized the vars, and make all vars public.

You can customize the GTK Dialog, like this sample from Control::openFile

´´´
		GtkWidget* dialog = gtk_message_dialog_new(getGtkWindow(),
												   GTK_DIALOG_MODAL,
													   GTK_MESSAGE_QUESTION, GTK_BUTTONS_NONE, "%s",
													   loadHandler.isAttachedPdfMissing()
															? _("The attached background PDF could not be found.")
															: _("The background PDF could not be found."));

		gtk_dialog_add_button(GTK_DIALOG(dialog), _("Select another PDF"), 1);
		gtk_dialog_add_button(GTK_DIALOG(dialog), _("Remove PDF Background"), 2);
		gtk_dialog_add_button(GTK_DIALOG(dialog), _("Cancel"), 3);
		gtk_window_set_transient_for(GTK_WINDOW(dialog), GTK_WINDOW(this->getWindow()->getWindow()));
		int res = gtk_dialog_run(GTK_DIALOG(dialog));
		gtk_widget_destroy(dialog);
´´´

And you can show the save dialog again. You simple have to call gtk_dialog_run and the same dialog is displayed to the user again, or if you do it the right oder, the dialog never disappear, so you can show the user an error, and let him fix it.

I would name the buttons in the dialog clear, like "Dou you which to overwrite blabla.xopp?"
-> Cancel
-> Choose other name (show the save dialog again)
-> Overwrite
or something like this.

luco5826:
Yes! It's exactly what i was referring to, if you think it could be a good idea then i'm going to implement it!

src/control/settings/Settings.cpp
```diff
@@ -1026,12 +1026,15 @@ string Settings::getDefaultSaveName()
 {
 	XOJ_CHECK_TYPE(Settings);
 
+<<<<<<< HEAD
```
andreasb242: Not merged!

luco5826:
OK! @andreasb242 I think we're there. I fixed all the stuff you suggested and the problems that came up from my first PR. 
Now the extension is always added after the user presses SAVE / EXPORT, so I added the `Util::replaceFileQuestion` method in order to manually handle overwrites. Now everything works! Let me know if there are problems or future issues!
(Should fix #521 )

src/control/jobs/CustomExportJob.cpp
```diff
@@ -39,6 +39,11 @@ CustomExportJob::~CustomExportJob()
 	}
 	exportRange.clear();
 
+	for(auto& filter : filters)
```
andreasb242: If you add the class by Value instead of pointer, you don't have to delete it. Even if this is slower, usually it doesen't matter, and you don't have to care about.
This is not a Problem, only a Hint ;-)

luco5826: Yes but I thought that since in the Contructor I create a `new ExportType`, I'd have to delete it in deconstruction phase, isn't it?

andreasb242: Yes, but if you dont write new, and remove the start in the type, it will be automatically be deleted.
Each time you access it (without reference) it will be copied, which is inefficent, but in such small cases it will be OK and easier to implement. It don't make sense, to change this now, as this is technical the better solution, but it's simpler the other way ;-)

PR #618, opened 2018-12-26
Draw tool: Coordinate system

PellelNitram:
I implemented a feature that lets one conveniently draw a coordinate system.

The feature resembles that of OneNote. The following video shows the behaviour of OneNote: https://www.youtube.com/watch?v=HPD5rPAPRXc.

Behaviour: When using the pen or highlighter, there is another tool available under the shape dropdown menu, namely a coordinate system.

What do you think of it?

**Comment:** There are two aspects that I was not able to cope with:

1. In `Actions.h` it is stated that one has to put the exact same order into `Tool.h`. When I do so, then the compilation works but a runtime error occurs upon startup straight away. I added the code to `Tool.h` but uncommented it stating the above problem.
2. I tried to implement the new feature similar to how drawing an arrow is implemented. Just an obervation on different bahaviour: When I close the program, my selection of the arrow tool is remembered but my selection of my newly implemented coordinate system tool is **not** remembered.

How to deal with the two above points? Also please let me know if I missed any other settings.

Just a side comment: I am really stunned of how much activity this project experiences recently (in particular by @andreasb242 !). As a first-time FOSS-contributer (this here is my second open-source contribution ever) I do feel the magic because of you guys. Merry Christmas :-).


andreasb242:
I got the source and check it.

andreasb242:
toolTypeFromString and toolTypeToString was missing, without this, the loading / saving does not work. I'll now check the runtime error you got. And then hopefully can merge, but I have to create a new Pull request.

andreasb242:
It's working now. See my fixes:
https://github.com/xournalpp/xournalpp/pull/619

Therefore I close this pull, as I used your commits for my changes, your name will still be in the History, only the Pull is closed.

Thank you, improvements are always welcome!

PR #665, opened 2018-12-31
 Italian translation, authors updated 

luco5826:
Hi!
I don't understand why everytime I edit the it.po file, `git` says that other translations have been edited too. I didn't change anything in other translations, can my changes get merged anyway?


andreasb242:
There is to much automated - I also notices this, and because I updated the german translation yesterday, you have now a merge conflict, and therefore you cannot merge.

I this case I create a new branch from upstream/master, copy my changes (in your case it.po and toolbar.ini) to the new branch, and create a new pull.

The CMake Script which does this is: /xournalpp/po/CMakeLists.txt
And it looks "standard", not much in there. If you want to fix it: Feel free, in my opinion the .po files should never be changed automatically, the .pot is Ok.

PR #715, opened 2019-01-06
Minor fix

luco5826:


src/util/StringUtils.cpp
```diff
@@ -5,12 +5,8 @@
 #include <string.h>
 
 string StringUtils::toLowerCase(string input)
-{
-	char* lower = g_utf8_strdown(input.c_str(), input.size());
-	string strLower = lower;
-	g_free(lower);
-
-	return strLower;
+{
```
andreasb242: This produces a memory leak

https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html#g-utf8-strdown

andreasb242: Please be really carefull with Memory stuff, such bugs are really hard to find

luco5826: Ok, so it needs to be like before this change?

andreasb242: Yes.

Issue #713, opened 2019-01-06
When pressure sensitivity is on, shape recognizer use always maximum width


Rmano:
The title says all (see the attached GIF also). 
It would be much better is the shape could be painted with the average line width of the hand-made path.

![recorded](https://user-images.githubusercontent.com/6414907/50739139-40d9be80-11dc-11e9-8eb3-3b8727194223.gif)

andreasb242:
Possible, but not sure if this really make sense. It's not much work, so we can try it.
We have to calculate the average pressure of the original object (which is anyway saved, for undoing) and then multiply it with the width.

andreasb242:
Yes... Its working really good. Thank you @luco5826 !

![bildschirmfoto vom 2019-01-08 19-39-08](https://user-images.githubusercontent.com/822958/50851523-60f4b380-137d-11e9-9c35-bbfeb97b8ad1.png)

PR #721, opened 2019-01-08
DO NOT MERGE - Stroke recognizer width

luco5826:
Hi! 
I think this could be a fix to #713 , but since I don't have a teblet to try this on yet, I'd like to ask you if this could be a solution and if not, which are the consequences.
If you could test this "fix" that would be perfect, even if I think this could mess up other things which don't involve Shape Recognition

### DIFF
--- a/src/model/Stroke.cpp
+++ b/src/model/Stroke.cpp
@@ -54,7 +54,7 @@ void Stroke::applyStyleFrom(const Stroke* other)
 {
 	setColor(other->getColor());
 	setToolType(other->getToolType());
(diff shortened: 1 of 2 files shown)


andreasb242:
Working Really good!

luco5826:
@andreasb242 Awesome! My doubt was that modifying that method could create some bugs when doing something like copy-paste a normal stroke, does it work anyway?

andreasb242:
Your right, this was the wrong line. I'll check this.

PR #723, opened 2019-01-08
Stroke recognizer size

andreasb242:
@luco5826 
This is the correct position.

But your code was all correct.


andreasb242:
Travis does not work...

luco5826:
Perfect!

PR #746, opened 2019-01-13
stroke style and fill rendered immediately while drawing

morrolinux:
* preserving incremental stroke rendering for "normal" strokes
* realtime rendering of stroke style and fill (better usability)


morrolinux:
@andreasb242  performance seems to be unaffected, is that ok for you?

andreasb242:
I would start now to see how to fix it, as the dotted lines needs to be completely redrawn each time... 
By the way: I would like to do the toolbar item like the page background one, but I need some time to implement it....

morrolinux:
Like the page background? I don't understand, you mean like the fill tool?

morrolinux:
as for the dotted line being redrawn each time... maybe we could instead re-draw small chunks of points each time? like in the current incremental redraw, but 4-8 points at a time instead of just the last two?

Issue #739, opened 2019-01-12
Draw coordinate system doesn't snap to grid.


frohro:
You can draw a coordinate system with the line tool, or the arrow tool, so maybe this feature is redundant, but if it is there, it would be nice if it worked as well as the other tools.  :-)

Thanks to all the developers!  You have some major improvements over xournal.
Rob

frohro:
The rectangle, the circle, and the coordinate system all lack this feature.  It is not so obvious unless you compile with the modification I made to the snap parameter, so it only snaps.   The arrow, thankfully has snap to angle, but not snap to grid.

andreasb242:
>unless you compile with the modification I made to the snap parameter

If you do fixes you can create a pull request, and we can merge it.
Developers are always welcome. The rules are simple: it needs to make sense, and don't break anything ;-)

The Coordinate System was also a Pull Request from a contributor. I already notice, that this feature is not fully integrated, but I personally merge them, because it also can be the base for improvements.

frohro:
Okay, I have snap to grid working for the square, the starting point of the arrow, (maybe I should do the end too).  I also fixed a bug with the square that it would only go up and to the left or down and to the right of the original point.  Shift to turn off snap to grid doesn't work with the rectangle because the shift was claimed to make a square.  It might be nice to change the key for one or the  other.  

I also need some directions on how you want me to do a pull request.  Right now I have these fixes committed to my local machine under a branch called frohro.

frohro:
I further edited it until now on the arrow, if you have snap to grid selected, but only want one of the ends snapped you can free the other by holding the shift down.  It works with either end.  

I created a repository here: https://github.com/frohro/xournalpp  I'm not quite sure how to do a pull request, except for repos I  have write permission on.  You probably know this much better.  Let me know if I can help you get the code.

frohro:
It is the frohro branch.

morrolinux:
https://help.github.com/articles/creating-a-pull-request-from-a-fork/

morrolinux:
You probably need to fix your tab size though. It's not consistent with the actual code

andreasb242:
Change the editor config to save Tabs instead of spaces. But the size (4) is correct.

PR #748, opened 2019-01-13
Frohro

frohro:
This is my work on #739, though I decided that the coordinate system wasn't really that important to me, so I just made snap to grid work on the rectangle, and arrow, as these are more important to me.  Making a coordinate system is easy with the arrow, and I can make ones with arrows on both ends, etc.  Snapping makes them look good.

### DIFF
--- a/src/control/tools/BaseStrokeHandler.cpp
+++ b/src/control/tools/BaseStrokeHandler.cpp
@@ -4,11 +4,66 @@
 #include "control/Control.h"
 #include "control/layer/LayerController.h"
 #include "undo/InsertUndoAction.h"
+#include <cmath>
 
 BaseStrokeHandler::BaseStrokeHandler(XournalView* xournal, XojPageView* redrawable, PageRef page)
  : InputHandler(xournal, redrawable, page)
 {
 	XOJ_INIT_TYPE(BaseStrokeHandler);
+}
+
+void BaseStrokeHandler::snapToGrid(double& x, double& y)
+{
+    XOJ_CHECK_TYPE(BaseStrokeHandler);
+
+    if (!xournal->getControl()->getSettings()->isSnapGrid())
+    {
+        return;
+    }
+
+    /**
+     * Snap points to a grid:
+     * If x/y coordinates are under a certain tolerance,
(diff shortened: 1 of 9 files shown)


morrolinux:
Ok so let me summarize it:
1) The arrow snapping is only working on the arrow tip for me (not both sides) 
2) The snapping tolerance should be configurable. Until we do this, I would not change it for consistency since it's so drastic.
3) Tab needs to be fixed 

Overall seems fine! 
Thanks for the contribution :) 
I can merge it when the above is fixed

frohro:
I'm not quite certain how to make it so you can have the first point off grid, but the arrow with snap to grid, so right now you can allow the arrow to not snap (by holding shift) and have the initial point snap to grid, but not the other way around.  Both ends snap to grid in that case.  

I don't have time to figure this out now.  Maybe later.  :-)

morrolinux:
Let me show you what happens with the arrow's grid snapping on my computer:
snapping

frohro:
It looks like the latest update didn't get included, i.e., you still have tolerance = gridsize/2.0; in the snapToGrid function in BaseStrokeHandler.  Did you do a new pull?

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -54,16 +109,16 @@ bool BaseStrokeHandler::onMotionNotifyEvent(const PositionInputData& pos)
...
 	this->redrawable->repaintRect(stroke->getX(), stroke->getY(),
-			stroke->getElementWidth(), stroke->getElementHeight());
+								  stroke->getElementWidth(), stroke->getElementHeight());
 
 	drawShape(currentPoint, pos.isShiftDown());
 	
 	rect.add(stroke->boundingRect());
 	double w = stroke->getWidth();
 
 	redrawable->repaintRect(rect.x - w, rect.y - w,
```
morrolinux: just an observation: this could be on a single line

morrolinux:
Oh sorry you were right. I wasn't following the branch, so `git pull` wasn't giving anything new.
I can merge this now!

PR #753, opened 2019-01-14
Frohro

frohro:
I added snap to grid to the circle and to the coordinate system.


src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -28,6 +28,7 @@ void BaseStrokeHandler::snapToGrid(double& x, double& y)
 	 */
 	double gridSize = 14.17;
 	double tolerance = 2.5; //gridSize/2.0; // if you want it to snap everywhere.
+	//tolerance = gridSize/2.0;  // The way frohro likes it.
```
morrolinux: "The way frohro likes it"? 

So.. I've added a setting for customizing grid snapping tolerance.
Please adjust to it by reading the value from the settings (see my last PR)

frohro: Sorry!  That comment should be cleaned out.  It was just for me.  

Am I to understand that you added a setting to allow adjusting this without messing with the source code?   That sounds really exciting.  I've been using xournalpp to make the slides for video [lectures](https://youtu.be/gN-d7Pzomzc) I'm doing for my Electronics class.  Its improvements over xournal are very helpful.   What do I need to do to change the setting for tolerance?

Thanks!

Rob

morrolinux: ahah I see :) 

Yeah, I've added an entry for changing rotation and grid snapping under the "Default" tab (but it might be moved under another tab in the future) 
You'll find an handy couple of spinners with values limited to a certain range that makes sense by the UI. 

I see that your code can be easily merged by resolving conflicts here on GitHub, I can handle it for you if you're uncertain on what to do.

PS: thanks for your contributions! It would be nice to also have arrows at either ends of the coordinate system.

morrolinux:
fixes #739

src/control/tools/BaseStrokeHandler.cpp
frohro: Thanks!  I am not very versed in github.  Please resolve the conflicts.  
On the arrows, here is what I'm thinking.  I'm not sure I like the growing arrow heads.  (They grow with the length of the arrows.)  It would be nice to have a setting that you could choose the size of the arrow head, which was then standard regardless of the length of the line.  The thing I have been doing to get around this is to use snap to grid to make it so I can extend lines with arrows that are a standard length, and so a standard arrow size.  What do you think?

morrolinux: Yeah, I don't like the proportional arrowhead either. 
I was already thinking about this, but I still need to find a really good solution, because it would probably be uncomfortable either if it was a general setting. 
I'll discuss this with @andreasb242 and see what he thinks of it

frohro: Good idea.  Another setting snap to grid needs is one to cut the grid size in half.  The squares are a bit too big for everything.  To me it seems a place on the toolbar for this might be more appropriate, because it is something that you want to change more frequently.

morrolinux: Agreed. could be a toggle to half the grid size when enabled

morrolinux: But this would work better if implemented alongside a full grid (painting) size management, which could store per-page squares size. Right now we only have a general template and the grid size is also duplicated in the code, so that's a thing that needs to be re-engineered first

frohro:
Do I need to do anything to finish this?  I may have messed the fork up by trying to sync with the repo here.

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -27,10 +27,11 @@ void BaseStrokeHandler::snapToGrid(double& x, double& y)
 	 * fix the point to the grid intersection value
 	 */
 	double gridSize = 14.17;
+  
 	double t = xournal->getControl()->getSettings()->getSnapGridTolerance();
 	double tolerance = (gridSize/2) - (1/t); 
-	
-	double xRem = fmod(x,gridSize);
+
+  double xRem = fmod(x,gridSize);
```
andreasb242: Formatting: here the indention looks wrong, and after x, a space is required.

morrolinux:
I just got the chance to test. Looks fine, works fine. 
Just tell me when you've fixed the above from Andreas so that I can merge!

src/control/tools/BaseStrokeHandler.cpp
Rmano: In my humble opinion, the arrowhead should be proportional to the line thickness of the line, not to the length. That way you can then add an arrow shape/size menu somewhere (In the end, it will be nice to have some configuration menu for all the shapes; for example for fill opacity, or for changing the anchors (draw circles by center and point or by bounding box), etc.   It could be a gear button near the shape selector or something like that. 
Just my 2 cents --- I do not think I have time to re-learn C++... and xournal++ is great now. It could be a killer app is someone is able to port it to android for the pen-enabled tablets...

andreasb242:
@frohro  so I know your timezone now... For me it's evening / night ;-)

>I think I repaired those, and pushed them.

Don't look like, but we are talking about some spaces, and @morrolinux already confirmed it's working.
So I merge.

src/control/tools/BaseStrokeHandler.cpp
andreasb242: @Rmano 
>for example for fill opacity

Already there! In the Toolbar or Menubar.

>and xournal++ is great now

Thank you :-) And really important: Only possible with the help of the community!

>It could be a killer app is someone is able to port it to android for the pen-enabled tablets...

Technical possible but not easy, and one of the issues: Probably even if you get the UI working as it is, it's may not the best.
When I would create a new Android App / UI and use the whole Xournal++ engine in the Background (drawing, editing undo/ redo etc.)
But even this will be much work.

PR #773, opened 2019-01-17
Pdf latex

andreasb242:
Removed Mathtex, use pdflatex, but still render to an image, so it does not solve the quality issue yet. But this is the first step, and it solves the license issue.

please check the pdf-latex branch, I don't use latex often, it seems to be working, even the preview works. But may I have escaping issues or something like this. Therefore please give me a feedback, It's may not ready for merging...


andreasb242:
@Rmano you gave the input about PDF Latex, you may would like to have a look at this, even if this does not solve your issue, as the PDF is internally still rendered to an image.
But it prepares for PDF rendering.

morrolinux:
Looks good to me! 
and finally we have latex on MacOS too :)

luco5826:
Awesome! It cleans up the project a lot!

Rmano:
Will try to check it this evening or tomorrow morning / European time; I have a quite busy day... ;-)

Rmano:
Yes, seems to still work ok. Great!

PR #801, opened 2019-01-22
Highlight cursor position - added

### LINKED ISSUE #786: Big Cursor with yellow opaque circle
Dear authors,
for screen-cast it would be great to have a "big cursor" with a yellow/red/whatever opaque circle. Since you overwrite the normal cursor-scheme DMZ-highlight is not working in Xournalpp. Joerg

luco5826:
Hi! I added the feature requested in #786 .
Now under Preferences -> View you can enable the highlighting option.
Right now I made it available only when in "Big Cursor" mode, but if you think that it should be available without that mode too, no problem, we only need to move one if outside a block ;)
Fixes #786

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: xournalpp 1.0.1\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-12-30 12:21+0100\n"
+"POT-Creation-Date: 2019-01-23 10:36+0100\n"
 "PO-Revision-Date: 2017-09-18 09:58+0200\n"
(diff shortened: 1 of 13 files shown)


JoergBitzer:
Thanks a lot. Just checked it out and it looks good. IMHO it would be better to have this option in both modes (small and big cursor). . However, at the moment the first item "Highlight cursor" depends on the second "big cursor". The view in preferences|view|cursor suggest that they are mutual independent.
 And since you said it is easily done. Your fast response is very appreciated. Thanks again!

campos20:
Is there an option the make the highlighting always on? And not just when we are writing?

campos20:
nvm

https://github.com/xournalpp/xournalpp/issues/1290#issuecomment-505781950

Thank you for your great work!

LittleHuba:
Not that I know of currently. But you can create a feature request if there is none already.

PR #809, opened 2019-01-27
modify Zoomslider behaviour

peetCreative:
- Add logarithmic scale function
- The changes are only every 5th time updated
- One sliding is one sequence

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -17,8 +17,8 @@ ZoomControl::ZoomControl()
 	this->zoomFitMode = true;
 
 	this->zoomStep = DEFAULT_ZOOM_STEP;
-	this->zoomMax = DEFAULT_ZOOM_MAX;
-	this->zoomMin = DEFAULT_ZOOM_MIN;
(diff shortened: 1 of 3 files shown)


andreasb242:
Technical working, but way are you updating only every 5 times?
I can change the slider, but the zoom does not change?

andreasb242:
Rendering is already in another thread, and rendering is delayed, as long as the zoom does not match the buffered image is fast scaled with cairo... So there should be no issue.

peetCreative:
Yes you are we kinda right. It's lacking a bit but its better than the not rendering.

peetCreative:
I found another issue, so don't merge... I will come back later with this pull request

andreasb242:
I'll try the performance on a slow notebook, but then I wait until you are finished.

peetCreative:
Mh as I played around with the Zoom calibration, shouldn't we have the min and max zoom depending the 100% zoom?
So minimum Zoom = 0.3 * zoom100 and maximum Zoom = 7 * zoom100

peetCreative:
OK I disabled the percentage which is still faulty. But you can merge or test.

andreasb242:
I'll test this on a high DPI and a standard display...

PR #813, opened 2019-01-28
fix #705 switch load of zoom and scroll

### LINKED ISSUE #705: Scrolling on startup
The current page is saved, if the document is opened again, it scrolls to the current page. The page displayed is the correct one, but it scrolls to the wrong one.
Probably is the zoom loaded later, and the position is therefore set to a complete wrong position.

peetCreative:
I have not extensivly tested the impact of this change, but it seems to solve problem as you intended in #705.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2207,8 +2207,8 @@ bool Control::loadMetadataCallback(MetadataCallbackData* data)
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


andreasb242:
WORKING!!!

Ahh that easy, I was searching really long for this, and didn't find it...

andreasb242:
Hi @peetCreative you may delete the unused branches on your repo, when I merge the pull - the branch will not be deleted.

Issue #808, opened 2019-01-27
Use PortAudio for Recording and Playback of audio


LittleHuba:
Switch to PortAudio library for audio handling

LittleHuba:
@andreasb242 could you point out a few things for me?

- How can I include the recorded data into the xopp format? Is there a preference on how to do that or can I simple create my own tag?

Is there something else to keep in mind except that it has to wrk (obviously) and it is cross platform?

andreasb242:
The Problem with the .xopp format is that it's an gzipped XML, which is a really bad start to put in an audio file. So I would currently still put it into a separate file, and keep the format as it is.

I would prefer to change the file format, but it's much work, and therefore I never did it. So for now I would stay at it and use an external file.

Instead of use a Library there would be also an option to integrate a tool which is better supported, e.g. a command line audio recorder / playback, add it as dependency.

I personal didn't look in details at audio recording, it's writte by @morrolinux .

VLC has also an API:
https://wiki.videolan.org/LibVLC/

If you have a good idea, or a sample (don't need to be fully working) post it here, before you do the full work, and the there is an issue (e.g. License, whatever).

I may missed some point, I was not prepared, and I didn't really expect that you really would do it ;-)
:+1:

LittleHuba:
What new file format are you thinking of? It might influence my approach as I'd prepare the code for those changes.

What I would like to prevent is having a huge dependency on another tool like VLC. Even if a lot of people have it installed on their system, it is still awful if you only have to install it for using xournalpp.

Additionally PortAudio is not as hard to use as it seems, it's actually one of the easiest audio libraries out there. The whole wrapper will be around 200-300 lines for recording and playback. So I'll most likely stick with it instead of just using another tool ;-)

I'll work it into xournalpp since I don't want to do the same work twice. ;-) Xournalpp is actually well prepared for the changes required, as far as I have seen.

morrolinux:
@LittleHuba that's great news! I had a look at PortAudio among the others few months ago, when I did the audio mapping stuff. I needed something to work with and I needed it quick (during lessons period) , so after a few trials went wrong I decided it was no option.
It's surely a good idea to incorporate this in xournalpp, and if you're familiar with it.. it probably won't take you as long, like you said. I'm looking forward to see your early changes as well

LittleHuba:
Ahh using a zip container makes it easy for me. Then I'll stick with the audio file.

[...]

PR #817, opened 2019-01-30
Move to portaudiocxx for recording audio

LittleHuba:
!!!DO NOT YET MERGE!!!

See issue #808 for reference

Only implementation to record audio using portaudio, and may not yet be compatible with playing the recordings.

### DIFF
--- a/src/util/audio/AudioQueue.cpp
+++ b/src/util/audio/AudioQueue.cpp
@@ -1,0 +1,62 @@
+#include "AudioQueue.h"
+
+AudioQueue::AudioQueue()
+{
+    XOJ_INIT_TYPE(AudioQueue);
+}
+
+AudioQueue::~AudioQueue()
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    XOJ_RELEASE_TYPE(AudioQueue);
+}
+
+void AudioQueue::reset()
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    this->notified = false;
+    this->streamEnd = false;
+    this->clear();
+}
+
+
+bool AudioQueue::empty()
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    return deque<int>::empty();
+}
+
+unsigned long AudioQueue::size()
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    return deque<int>::size();
+}
+
+void AudioQueue::push(int *samples, unsigned long nSamples)
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    unsigned long requiredSize = this->size() + nSamples;
+    if (this->max_size() < requiredSize)
+        this->resize(requiredSize);
+
+    for (long i = nSamples - 1; i >= 0; i--)
+        this->push_front(samples[i]);
+}
+
+std::vector<int> AudioQueue::pop(unsigned long nSamples)
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    std::vector<int> buffer(nSamples);
+    for (long i = nSamples - 1; i >= 0; i--)
+    {
+        buffer[i] = this->back();
+        this->pop_back();
+    }
+    return buffer;
+}
--- a/src/util/audio/AudioQueue.h
+++ b/src/util/audio/AudioQueue.h
@@ -1,0 +1,40 @@
+/*
+ * Xournal++
+ *
+ * Queue to connect an audio producer and an audio consumer
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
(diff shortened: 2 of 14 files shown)


src/util/audio/AudioQueue.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    return deque<int>::size();
+}
+
+void AudioQueue::push(int *samples, unsigned long nSamples)
+{
+    XOJ_CHECK_TYPE(AudioQueue);
+
+    unsigned long requiredSize = this->size() + nSamples;
+    if (this->max_size() < requiredSize)
```
andreasb242: We use to have brace everywhere, because it causes less bugs if you do this consequent... (but this is not a blocker, and I write this only once)

src/util/audio/AudioQueue.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+    AudioQueue();
+    ~AudioQueue();
+    void reset();
+    bool empty();
+    unsigned long size();
+    void push(int *samples, unsigned long nSamples);
+    std::vector<int> pop(unsigned long nSamples);
+
+    std::mutex queueLock;
```
andreasb242: If possible do this protected or private

src/util/audio/DeviceInfo.h
```diff
@@ -0,0 +1,32 @@
...
+#include <portaudiocpp/PortAudioCpp.hxx>
+#include <string>
+
+class DeviceInfo
+{
+public:
+    DeviceInfo(portaudio::Device *device, bool selected);
+    ~DeviceInfo();
+
+    const std::string deviceName;
```
andreasb242: Also, private / protected if possible

src/util/audio/PortAudioProducer.cpp
```diff
@@ -0,0 +1,123 @@
...
+    XOJ_RELEASE_TYPE(PortAudioProducer);
+}
+
+std::list<DeviceInfo> PortAudioProducer::getInputDevices()
+{
+    XOJ_CHECK_TYPE(PortAudioProducer);
+
+    std::list<DeviceInfo> deviceList;
+
+    for (portaudio::System::DeviceIterator i = this->sys.devicesBegin(); i != sys.devicesEnd(); ++i)
```
andreasb242: Nice detail: ++i! :+1:

src/util/audio/PortAudioProducer.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+int PortAudioProducer::recordCallback(const void *inputBuffer, void *outputBuffer, unsigned long framesPerBuffer, const PaStreamCallbackTimeInfo *timeInfo,
+                                      PaStreamCallbackFlags statusFlags)
+{
+    XOJ_CHECK_TYPE(PortAudioProducer);
+
+    if (statusFlags)
+    {
+        g_message(("PortAudioProducer: statusFlag: " + std::to_string(statusFlags)).c_str());
```
andreasb242: g_message supports formatting with printf style (only a hint)

andreasb242:
I added some comments - but in generally looks really good, I found only details. You have to change the travis config, else the build is not running (and the Readme, but this can be done later)

Thank you, good work :+1: !

LittleHuba:
I just fixed all the things you requested except the Readme. That would just be unecessary work.

LittleHuba:
I thought you meant the part about what is necessary for recording audio...

I'll update the build instructions, as this might prevent others of using Xournal++
Then you can merge the changes.

andreasb242:
Ok, and I'll do a build on Windows 10 and macOS to check if it's there still building...

LittleHuba:
Just updated the build instructions for Linux.
Could you update the instructions for Windows and Mac? Since I'm unable to build on either OS...

LittleHuba:
The same issue will be, with how I required SOX. I actually did not have any contact with this cmake command yet. I included libraries in a different way until now. But thanks for the info!

andreasb242:
I build everything for macOS, and for Windows I use as much as I can get from MingW.
But the Libs are also not available for Windows I think:

Therefore, I need this, right:
http://www.portaudio.com/download.html
http://www.portaudio.com/archives/pa_stable_v190600_20161030.tgz

Does this include your C++ wrapper? Or which Library do I need?

Is SOX this project?
http://sox.sourceforge.net/

LittleHuba:
Version 19 of Portaudio does include the necessary wrappers.
And yeah sox is found under that path.

andreasb242:
https://github.com/xournalpp/xournalpp/pull/821

For SOX I need to check tomorrow, I didn't find the source package yet... Probably somewhere hidden linked on the webpage or on the sourceforge page...

but not today...

PR #833, opened 2019-02-02
Highlight position in small cursor mode too

luco5826:
Hi!
As requested in PR #801 I added the "Highlight cursor position" not only in BigCursor mode but in small one too


JoergBitzer:
Great, you are really fast. Thanks again.

PR #838, opened 2019-02-03
Audio output using PortAudio

LittleHuba:
See issue #808 for reference


src/gui/dialog/SettingsDialog.cpp
```diff
@@ -269,6 +270,47 @@ void SettingsDialog::load()
...
+	{
+		if (this->audioOutputDevices[i].getSelected())
+		{
+			gtk_combo_box_set_active(GTK_COMBO_BOX(get("cbAudioOutputDevice")), i);
+		}
+	}
+
+	switch((int)settings->getAudioSampleRate())
+	{
+		default:
```
andreasb242: Usually default should be at the end, even modern compiler support it may this way (this is no blocker!)

src/gui/dialog/SettingsDialog.h
```diff
@@ -48,8 +50,12 @@ class SettingsDialog : public GladeGui
 	XOJ_TYPE_ATTRIB;
 
 	Settings* settings;
+	Control* control;
 	GtkWidget* callib;
 	int dpi;
+	std::vector<DeviceInfo> audioInputDevices;
```
andreasb242: for vector and string we have the using std::vector in the header already, so this is not needed, as these classes are used very often, but also, no issue...

src/util/audio/AudioPlayer.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    // Clean up after audio is played
+    stopThread = std::thread([&] {
+        while(this->portAudioConsumer->isPlaying())
+        {
+            Pa_Sleep(100);
+        }
+        this->stop();
+    });
+    //TODO we need something nicer here then detaching the thread
```
andreasb242: Is there no listener or something like this?

src/util/audio/PortAudioProducer.cpp
```diff
@@ -5,123 +5,160 @@ PortAudioProducer::PortAudioProducer(Settings *settings, AudioQueue *audioQueue)
...
-	this->inputChannels = static_cast<unsigned int>(device->maxInputChannels());
+    XOJ_CHECK_TYPE(PortAudioProducer);
+
+    try
+    {
+        return DeviceInfo(&sys.deviceByIndex(this->settings->getAudioInputDevice()), true);
+    }
+    catch(portaudio::PaException &e)
+    {
+        g_message("PortAudioProducer: Selected input device not found - fallback to default input device");
```
andreasb242: May add e.cause to the log, as there may another exception occur

LittleHuba: There is only one cause for this exception: The device, selected in the settings, got disconnected or is for some other reason not available.

Should I add a more verbose description or do you mean some special command I have not yet seen?

src/util/audio/AudioPlayer.cpp
andreasb242: You have an example in your code: Lambdas have a closure, so it should possible to use this to additional add all needed data.
(before lambdas: mem_fun can do this also)

I don't have a working example, but I'm really sure it will work ;-)

src/util/audio/PortAudioProducer.cpp
andreasb242: you should log e.what(), because you never know, may the next version provide a string or so, so usually I would log it, a normal user don't see it anyway. (add it to the existing message, nothing special)

PR #858, opened 2019-02-09
Eraser's cursor's size now matches the actual eraser size

### LINKED ISSUE #856: [Enhancemet] Set eraser cursor size to match area to erase
Presently is not easy to know which area is to be erased

luco5826:
Hi! 
This should  fix #856 
I thought one thing tho, right now we re-draw the cursor every time the cursor goes off and on the paper, even if it's not a really big issue, I think we could speed up things a little more by creating the cursor's image file in the .xournalpp folder only once, and then we could just load them when needed.
I'm saying this because we would only need 6 small pngs, 3 for the eraser's size, and 3 for the pen (big cursor, highlight position, both of them combined)
What do you think?

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1828,6 +1828,7 @@ void Control::showSettings()
 	bool verticalSpace = settings->getAddVerticalSpace();
 	bool horizontalSpace = settings->getAddHorizontalSpace();
 	bool bigCursor = settings->isShowBigCursor();
+	bool highlightPosition = settings->isHighlightPosition();
 
 	SettingsDialog* dlg = new SettingsDialog(this->gladeSearchPath, settings, this);
 	dlg->show(GTK_WINDOW(this->win->getWindow()));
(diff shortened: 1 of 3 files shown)


LittleHuba:
I'd ultimately want to allow users to select their own pen thickness by using a slider, while keeping the four current settings as a quick choose. That would be essentially harder if we would follow this new approach...

luco5826:
Hi @LittleHuba ! 
Yeah that would be a very good improvement! But I think that the cursor's size would be the last problem then, I don't know how strokes are saved to file or drawn, so I think that would be a very major commit!

LittleHuba:
Yeah that will most likely be a lot of work. But actually drawing the cursor once per entering the page is not that much overhead. 

What we could do is rendering the cursor when it is changed and just store it internally in a canvas or something like that. That would bring your speed improvement while not limiting the cursor to a few predefined sizes. 
You would have to do that anyway or the time saved by storing the cursor in a file would be consumed by actually opening that file every time.

andreasb242:
First: The tool size can be customized relative easy.

@luco5826 is using
>control->getToolHandler()->getSize();

Can you please use
>control->getToolHandler()->getThickness();

This returns a double, so if you customize the size, there is no issue!

And for @LittleHuba 
The size of ToolHandler::getThickness() is used for Eraser and Stroke.
The size is read from the tool, this needs to be changed, but it's not that much. The big problem would be to create an easy to use UI. I already used a slider for transparency, I don't think it's really easy to use...

andreasb242:
I uses the right size, so this is no issue. It's working, even it looks really strange with this monster cursor ;-)

PR #860, opened 2019-02-11
Add audio seeking capability

LittleHuba:
Adds the correct timestamp handling to the strokes associated to a recording.
Warn user when start of recording was not successful.

### DIFF
--- a/src/control/AudioController.cpp
+++ b/src/control/AudioController.cpp
@@ -44,7 +44,7 @@ void AudioController::recStartStop(bool rec)
 			return;
 		}
 
-		sttime = (g_get_monotonic_time() / 1000000);
+		sttime = static_cast<size_t>(g_get_monotonic_time() / 1000);
 
 		char buffer[50];
 		time_t secs = time(nullptr);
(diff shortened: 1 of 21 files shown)


src/control/AudioController.cpp
```diff
@@ -44,7 +44,7 @@ void AudioController::recStartStop(bool rec)
 			return;
 		}
 
-		sttime = (g_get_monotonic_time() / 1000000);
+		sttime = static_cast<size_t>(g_get_monotonic_time() / 1000);
```
LittleHuba: This might break compatibility to older recordings, as offset is changed.
The change is needed to get a correct sample offset, as seconds are not a small enough unit.
@andreasb242 how should I handle this?

andreasb242: LoadHandler::parseStroke() parses this time from file.
>	if (LoadHandlerHelper::getAttribInt("ts", true, this, ts))

The readed document version is stored in this->fileversion

``
void SaveHandler::writeHeader()
{
	this->root->setAttrib("creator", PROJECT_STRING);
	this->root->setAttrib("fileversion", "3");
	this->root->addChild(new XmlTextNode("title", "Xournal++ document - see " PROJECT_URL));
}
``
Here you can increase the version, then you can change the format, and read both files correct back.

LittleHuba: Thanks a lot!

LittleHuba:
Yeah I tested it and it does work :+1:

Issue #857, opened 2019-02-09
Making Flatpak package


Garbulix:
Having cross-distro package would be very nice. Haven't you thought about making Flatpak package? 
I cannot do that at the moment, because I am still not good at this stuff (compiling, making packages...).

andreasb242:
I'm against it, I think we should get to the offical repos, then also the update and all will be handled, or at least ppa. If possible also Windows Marked and Apple Store.

Garbulix:
I don't believe that will work, sorry. 
There is a lot of "big" distros and maintaining packages in every repo will need manpower - I guess there is not enough people for that. 
For example - there is no xournalpp in any of Fedora's repo (official, rpmfusion or even copr). 
I know that xpp from official repo could be better integrated with system (theme for example), but looking after all the distros would be too hard. Also - xpp isn't an app that needs some privileges - it is just a notebook. I would understand if that was some kind of system app, like Timeshift or VS Code (which sometimes need an access to terminal). 
Also, one of the pros of having a Flatpak is that there is one, nicely maintained Flathub. More apps there the better - it just looks good for people-new-in-linux.

Sorry, you can see that I am huge Flatpak fan. I believe that containerization and centralization can be step forward in making Linux systems usable for "normal people".
Eventually - I would deal with no Flatpak (or even Snap) package if there was a package in my Fedora's repository (a little selfish, I know).

andreasb242:
@Garbulix only flatpack fan or also developer?
> I cannot do that at the moment, because I am still not good at this stuff (compiling, making packages...).

I never used flatpack, and I do already the Windows and macOS Build, even if I'm not really using them.
If you whish to do it, no problem - we will help you, but I don't do it, and I don't think that one of the other developers would do it.

Garbulix:
I think that my previous post might sound little impudent - sorry for that. You are doing great job.
> @Garbulix only flatpack fan or also developer?  
Shamefully - currently only a fan. 
I want to dive into Linux programming (I was an IT student once...) in the future.
I hope I will be able to make some Flatpak packages in the future and maintain them and spread "good news".

andreasb242:
For this probably not even programming is required, or at least not much - usually you only need some commands, which may should be automated with a shellscript or so.

Did you build Xournal++ yourself? Then it also should not be that hard to create a package.
http://docs.flatpak.org/en/latest/first-build.html

For me this looks relative simple (yes I know, the details are always the problem):

>"build-commands": [
>                "make install"
[...]

PR #870, opened 2019-02-11
Flatpak: Initial manifest proposal

ezickler:
This a flatpak manifest i came up with. It is working for me, but as this is the very first app i wrote a flatpak  manifest for, this might be totally buggy. I'm also new to Xournalpp so i might not even notice if something is not working what should work. 

In #857 i posted a json manifest which i converted to yaml as it supports comments properly. I also dumped the sox dependecie and added libsndfile instead. 

To create a flatpak file the following to commands are needed.  ./flatpak-build and ./repo are arbitrary
folder names.
```
flatpak-builder ./flatpak-build org.github.xournalpp.xournalpp.yaml --force-clean --repo=./repo 
flatpak build-bundle ./repo  xournalpp.flatpak com.github.xournalpp.xournalpp
```

### DIFF
--- a/org.github.xournalpp.xournalpp.yaml
+++ b/org.github.xournalpp.xournalpp.yaml
@@ -1,0 +1,88 @@
+---
+app-id: com.github.xournalpp.xournalpp
+runtime: org.gnome.Platform
+runtime-version: '3.30'
+sdk: org.gnome.Sdk
+command: xournalpp
+rename-icon: xournalpp
+rename-desktop-file: xournalpp.desktop
+rename-appdata-file: xournalpp.xml
+finish-args:
+  # X11 + XShm access
+  - "--share=ipc"
+  - "--socket=x11"
+  # Wayland access
+  - "--socket=wayland"
+  # Needs to save files locally
+  - "--filesystem=xdg-documents"
+  - "--socket=pulseaudio"
+  # Allow dconf
+  - "--filesystem=xdg-run/dconf"
+  - "--filesystem=~/.config/dconf:ro"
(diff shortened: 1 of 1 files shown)


org.github.xournalpp.xournalpp.yaml
```diff
@@ -0,0 +1,88 @@
...
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: http://www.mega-nerd.com/libsndfile/files/libsndfile-1.0.27.tar.gz
+        sha256: a391952f27f4a92ceb2b4c06493ac107896ed6c76be9a613a4731f076d30fac0
+  # Tried git, but failed to get the autogen script to run.
+  #     - type: git
+  #       url: https://github.com/erikd/libsndfile
+  #       commit: 1d928bffdb65e827c3e7d17ece0e123b4a70ec6c
+  #       tag: 1.0.28
```
LittleHuba: A good way to find out how to build a library is looking at the PKGBUILD of Arch packages.
You might want to look at the dependencies as there could be a change to 1.0.27.

ezickler: Thx, nice tip. But the do exactly the same as i do. Download the .tar.gz from www.mega-nerd.com/libsndfile instead of using the git :)

LittleHuba: That was not to point out that you have to use git ;-) It was more a general tip, as you pointed out that you did not get a working build setup.

But for your information there is a package that uses git:
https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libsndfile-git

PR #890, opened 2019-02-16
Icon loading

andreasb242:


src/control/XournalMain.cpp
```diff
@@ -302,6 +302,15 @@ int XournalMain::run(int argc, char* argv[])
 
 	Control* control = new Control(gladePath);
 
+	if (control->getSettings()->isDarkTheme())
+	{
+		string icon = gladePath->getFirstSearchPath() + "/iconsDark/";
+		gtk_icon_theme_prepend_search_path(gtk_icon_theme_get_default(), icon.c_str());
+	}
+
+	string icon = gladePath->getFirstSearchPath() + "/icons/";
```
luco5826: Hi @andreasb242 ! I don't know if this is the line which causes this, but now "dark-themed" icons doesn't get loaded anymore, did you make it available only on "make install" installation?

andreasb242: I switched to GTK Icon loading. They are still installed within the ui Folder, which should be installed.
And I also tested it, but one thing I was not sure:
gtk_icon_theme_prepend_search_path or gtk_icon_theme_apend_search_path

After a few tries this solution works for me.

morrolinux: image
Hey there, I'm not sure what happened, but I've just installed xournalpp on a fresh new system and dark icons don't seem to get loaded. Where can I start digging?

PR #905, opened 2019-02-17
Update eraser's cursor's size on zoom change

### LINKED ISSUE #894: eraser cursor size should change with zoom
I really like the new eraser cursors from #856, but maybe it still needs some tweaks.

When I zoom out, the cursor size stays the same (in terms of pixels) and (I think) the amount being erased stays the same in terms of papersize (that is, a smaller area if terms of pixels).

luco5826:
Hi!
This should fix #894 
I think I'm going to implement the stuff we we're saying about "saving a pre-rendered icon" for some tools and use it when needed instead of re-draw them every time. At least I'd like to see if this could be less "resources extensive", not that it is right now but... why not?

### DIFF
--- a/src/.vscode/c_cpp_properties.json
+++ b/src/.vscode/c_cpp_properties.json
@@ -6,7 +6,8 @@
                 "${workspaceFolder}/**",
                 "/usr/include/libglade-2.0",
                 "/usr/include/glib-2.0",
-                "/usr/include/gtk-2.0"
+                "/usr/include/gtk-2.0",
+                "/usr/include/gtk-3.0"
             ],
             "defines": [],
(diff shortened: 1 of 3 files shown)


andreasb242:
What do you want to save?

luco5826:
Hi @andreasb242 ! I'd like to save something like the rendered pixbuf inside the Cursor class and, when needed, switch the pixbuf used for the actual cursor, instead of re-draw it every time, what do you think?

andreasb242:
Even drawing the background is much much more to do than this, so I don't think this brings anything.

luco5826:
Ook then, no problem

PR #922, opened 2019-02-22
Add buttons to control audio playback

LittleHuba:


LittleHuba:
@andreasb242 would you mind taking a look at the way I'm updating the ui elements with the approriate state?

andreasb242:
I'm already reviewing...

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -478,7 +478,11 @@ void ToolMenuHandler::initToolItems()
 	fontButton = new FontButton(listener, gui, "SELECT_FONT", ACTION_FONT_BUTTON_CHANGED, _("Select Font"));
 	addToolItem(fontButton);
 
-	ADD_CUSTOM_ITEM_TGL("RECSTOP", ACTION_RECSTOP, GROUP_REC, false, "rec", _("Rec / Stop"));
+	ADD_CUSTOM_ITEM_TGL("AUDIO_RECORDING", ACTION_AUDIO_RECORD, GROUP_AUDIO, false, "audio-record", _("Record Audio / Stop Recording"));
```
andreasb242: If you change something like this, we need backward compatibility.... (not checked yet)

src/util/audio/AudioPlayer.cpp
```diff
@@ -25,44 +26,78 @@ AudioPlayer::~AudioPlayer()
...
-								 while (this->portAudioConsumer->isPlaying())
-								 {
-									 Pa_Sleep(100);
-								 }
-								 this->stop();
-							 });
-	stopThread.detach();
+	if (status)
+	{
+		status &= this->play();
```
andreasb242: This probably should be logical or? Even if this probably works in 99% of all the cases with bool....

src/gui/toolbarMenubar/ToolMenuHandler.cpp
LittleHuba: Where and how do I add this?

andreasb242: ToolMenuHandler::load, there is already such a backward fix:

```
				// recognize previous name, V1.07 (Jan 2019) and earlier.
				if (name == "TWO_PAGES")
				{
					name = "PAIRED_PAGES";
				}
```

andreasb242:
Locks good!

Issue #937, opened 2019-02-25
New Fileformat - basic steps


andreasb242:
The main issue on he fileformat is, that it's not possible to embedded resources like PDFs, Audio Files, Images etc.

@LittleHuba now would start improving it, but I think we should make the plans public before start.
@morrolinux may also have a look at it.

Technical we would use .zip instead of .gz, which support multiple files within the .xopp file.
For this we thinks about the library libzip

1. Reading of the new fileformat needs to be enabled by a compiler option, only writing will be available, for backward compatibility. Previewer, open file with double click etc. need to be configured. (this should be for at least 2 Releases)
The contents of the new .zip file (have a look at .odt!):
/mimetype => Textfile with String "application/xournal++"
/Thumbnails/thumbnail.png => The .png Thubnail, which is now embedded as preview in the .xml
/content.xml => the current XML, with the same format
/META-INF/version => "6" Current document version
/META-INF/compatible => "6" Which version needs to be supported, to read the .xopp, so we decide here, if we do backward compatible changes or not.

2. We increase /META-INF/version and /META-INF/compatible and put e.g.the images into a separate Folder like "Pictures" for the images, and reference them in the content.xml
Also Audio files and .pdfs can be packed.
Audio files and images don't need to be compressed, they are already compressed.

LittleHuba:
> Reading of the new fileformat needs to be enabled by a compiler option, only writing will be available, for backward compatibility.

I would keep compatibility for the old format, as it requires virtually no overhead (10 lines of code). Therefore we do not need to keep anything hidden behind compiler flags.

LittleHuba:
What we could do is preventing saving of files in the old format after 2-5 releases, so we can then drop the support of the old format after some more releases to reduce code complexity. 

But to get no complains we must provide a tool to convert to the new format by then.

andreasb242:
No, we cannot drop the old format, we have an export as .xoj, which is the old format without some additions. .xoj is good supported by Xournal, by MrWrite, and may by other tool. Dropping the old format is therefore no Option.

tuxflo:
Maybe I'm late to the party but wouldn't it be possible to save the annotations directly to the pdf file? The windows application "xodo" seems to do this and I think that might be a great solution (for all files without audio an custom stuff).

Technius:
It is possible, but it would be very complicated. Some problems include:
[...]

PR #946, opened 2019-02-27
Add reading of new file type

LittleHuba:
Reading part of new file format (see #937)


LittleHuba:
Reading is now fully implemented and unit tests are added. Ready for inclusion into next release

src/control/xojfile/LoadHandler.h
```diff
@@ -111,9 +118,13 @@ class LoadHandler
 	ParserPosition pos;
 
 	string creator;
-	int fileversion;
+	int fileVersion;
```
andreasb242: You can initialize all vars here, instead of the constructor. But it's not required, only a tip for the future.

src/gui/PageViewFindObjectHelper.h
```diff
@@ -162,7 +162,11 @@ class PlayObject : public BaseSelectObject
 
 			if (!fn.empty())
 			{
-				view->getXournal()->getControl()->getAudioController()->startPlayback(Path::fromUri(view->settings->getAudioFolder()).str() + "/" + fn, (unsigned int) ts);
+				if (fn.rfind('/', 0) != 0)
```
andreasb242: Windows use Backslash, may use the Path class instead.Or is the Path an internal one?

src/control/xojfile/LoadHandler.h
LittleHuba: Just kept how it was previously :wink:

PR #956, opened 2019-03-04
remove xopp-recording.sh

flokli:
it's not used anymore in releases > 1.0.7.


flokli:
What's your opinion on removing the <= 1.0.7 instructions from the readme? I could add a followup commit removing these too…

andreasb242:
@LittleHuba we can merge this?

LittleHuba:
@flokli we should wait with removing the instructions for <= 1.0.7 releases for one or two more releases. Then all the users should at least be using a version that uses the new system.

PR #961, opened 2019-03-06
Edit selection movement between and going off page

JJones780:
This is in regard to the movement of selected objects between pages after selection.
 
This functionality was broken in the multilayout code changes.
This is a slightly different implementation to consider:
  It uses the page found under the cursor ( instead of the one with the biggest overlap );
  It doesn't delete the item if finalized off-page but instead resets to original page and position.


andreasb242:
Hi
Thank you very much!

If you would work more often on Xournal++ you might contact me over gitter.im or mail (you should see my Email address in the History)

PR #972, opened 2019-03-10
Layout get view at

JJones780:
This investigates adding binary search to getViewAt() but requires it to be part of the Layout class.

The experiment:

 I had speculated that a binary search wouldn't be beneficial until a larger number of pages were involved. I wanted to see what that number would be.
 From my experimenting it looks like the binary search is only more efficient for columns (or rows - or in the case of my original linear search, total pages ) greater than 250.
 

If you checkout and run commit eb1be15 it will print out timing comparisons of the linear and binary search as the mouse is moved around.

The design change:

 I've done my best to clean it up for inclusion if moving the pageposition functionality into Layout makes sense going forward. I'm not totally convinced myself and wouldn't be offended if this contribution wasn't included.


JJones780:
Note that the last commit REMOVES pageposition/* files and code referencing them as this functionality is moved into the Layout class.

Also: this makes running `cmake build` necessary

Issue #998, opened 2019-03-19
"File already exists. Do you want to overwrite it?" is shown twice


coldspark29:
If you want to save a file with a name, that already exists, you get two notifications. One notification from GTK3 I guess and the other from Xournal++. Might also be that it has to do with me being on KDE. There is also a spelling error in the German notification btw

peetCreative:
@luco5826 has fixed things for extension in #594 .
I would like to have the gtk dialog, because it's much more consistent.

andreasb242:
The GTK Dialog does not allow to select a file extension, therefore e.g. Gimp also uses a Custom export handling like this...

peetCreative:
According to Gimp: They also use the GTK overwrite dialog, but handle it using signals.
https://gitlab.gnome.org/GNOME/gimp/blob/master/app/dialogs/file-save-dialog.c

peetCreative:
I think one could implement export, save, open dialogs similar to gimp, using function connected to signals.

JJones780:
It looks like GIMP plays some tricks to have the dialog do the overwrite check on a modified filename. 
I tested the "response" signal but that occurs AFTER the overwrite warning which is why GIMP runs through the response twice, the second time around is with the modified filename and overwritewarning turned on . It seems like a bit too much hassle. Didn't GTK originate with GIMP.. why are they hacking a solution?  Maybe I'm wrong in how they are doing it.
I also looked into using the "update-preview"  signal... which happens on every little change but I think that's asking for trouble.

JJones780:
Pull request 1017 includes a simple fix (that keeps the existing custom dialog for now).

PR #1017, opened 2019-03-27
Fix double warnings

### LINKED ISSUE #1016: "This document not yet saved" shown twice when loading recent document
Create a new file then select File->Recent Documents->  and choose any recent document listed.
"This document not yet saved"  Save Discard Cancel  Dialog is shown twice in succession.

JJones780:
I've done a bit of testing and I'm fairly confident this fixes #1016 and #998.

#998 has comments about doing things the way GIMP does to use the GTK warning dialog which we may want to consider. See my comments there.  In the meantime I think this fix will be sufficient.


andreasb242:
Makes sense...

Issue #772, opened 2019-01-17
Zoom issue


luco5826:
Hi!
I got a problem while using Xournal++, I recently found out that when I open Xournal++ the zoom is centered way above the page and I cannot scroll down as the gif shows, if I mess with the zoom I get to move down, but then I'm not able to Zoom out, neither with panel button not with CTRL + scroll.
Right now I think the first part has something to do with the "Add vertical space" option, which I enabled.

![ezgif-4-cd9a40b50946](https://user-images.githubusercontent.com/38291180/51341678-2c29e000-1a92-11e9-925a-6d8c9e730fe0.gif)

luco5826:
Ok I checked, the first part is related to the "Add vertical space" option, if I disable it I don't have that proble anymore, but the Zoom one persists, I don't even know if they're correlated

GregDMeyer:
I have the same problem that I can't zoom out. Clicking zoom in or zoom out always zooms in. Even trying to pinch to zoom out still zooms in. Sometimes even just pressing save zooms in! and there is no way to zoom back out again.

andreasb242:
You may compare the two files, they are textfiles / human readable!

peetCreative:
@GregDMeyer have you found out sth in the files?

GregDMeyer:
Sorry it took me a bit, I had diff'ed them a few days ago but I guess I had customized a lot of settings because the diff was like the whole file ;)

It turns out the culprit is "Presentation Mode". When I turn it on, zooming is broken; when I turn it off, it works again. (The settings file stores an attribute for presentation mode, so that's why it was persistent between restarts). So the easy fix to get zoom working again is to turn off presentation mode!

peetCreative:
OK, so this issue is related to #671.
Anyhow in Presentationmode Zoom should be disabled.

GregDMeyer:
Sounds good. Should this issue be closed then as a duplicate?

andreasb242:
As this is a really specific issue: Let it open. There are anyway to many issues open again ;-)

LittleHuba:
Zomming is now completely broken. We need to rewrite it to handle the new columnized layouts and handle pages with different sizes in a better way.

peetCreative:
@LittleHuba What do you mean? For me it's not exactly zooming to the right position but roughly it works.

peetCreative:
Btw. what is the expected behavior for Rows/Colums layout for presentation mode?

andreasb242:
[...]

PR #1022, opened 2019-03-28
WIP:improve zoom behavior Presentation

### LINKED ISSUE #671: Presentation Mode not really working
- Zoom working strange - should be always 100% I think
- Toolbar should be selectable in the settings
- Create a Presentation toolbar, and select as default

peetCreative:
This rather bit too big pullrequest is supposed to fix various problems with zoom and view modes,
as well refactors zoom related parts:
Fix #671 #772
I would be happy, if someone has a look on it.

Präsentationmode:
- Zoom and Scroll is disabled
- Zoom that whole current page is viewed
- disable Paired, Rows or Columns settings
- Navigation with Keys: Up/next page,  down/previous page
- Zoom action Buttons are disabled instead of hidden 
Zoom Fit:
-  Zoom is adjusted by listining size-allocate signal of widget
Problems/Todo (to be created issues if merge):
- Presentation icon has to be designed. presentation-mode.svg is nothing saying. I'm not a graphic designer
- Sometimes after disabling Presentation-mode scroll is still disabled
- Layout::scrollTo does not work correctly.
-- Fitmode enabled and several columns, rows or paired: when going from one to the next page horizontal scroll is not working
-- Presentationmode and Document with horizontal differently sized pages gives same problems


LittleHuba:
Very nice to get this view mode actually running! I'll take a look sometime later.

JJones780:
### re: Layout::scrollTo does not work correctly.
XournalView::scrollTo  calls layout::ensureRectVisible.

In `void XournalView::scrollTo(size_t pageNo, double yDocument)`
 the yDocument argument is outdated and should be 0.

I'm slowly working through some cleanup leftover from the multilayout change and this is one of them.
If you want to leave the scrollTo issue and merge this code as is, I will put the fix above in with my changes.
I will also look at the other scrollTo issues you mentioned as soon as you're merged. I think some may already be fixed in my working branch.

JJones780:
re: zoom

Also, I've rewritten Layout::UpdateCurrentPage and find that zooming behaves better if this function does not select a new page in the visible area. My update will reflect this I expect. 
I don't think this affects your existing changes.

JJones780:
Does presentation mode require changing the layout of the document?
If so we should really create a 
void LayoutMapper::configureForPresentation(int numPages) which calls the generic  LayoutMapper::configure with some fixed settings  which is called when presentation mode is activated. 
Then when presentation mode is deactivated the user gets their original layout back.
Also, disabling the layout menu items during presentation mode would avoid confusion.
But:
I'd suggest considering NOT changing the layout for presentation.

peetCreative:
I will have a look at your comments tonight

src/control/Control.h
```diff
@@ -126,10 +126,8 @@ class Control :
 	void saveSettings();
 
 	void updateWindowTitle();
-	void calcZoomFitSize();
-	void setZoomFitButton(bool enabled);
 	void setViewPairedPages(bool continous);
-	void setViewPresentationMode(bool continous);
+	void setViewPresentationMode(bool);
```
andreasb242: Keep the var name also in the prototype, for documenation

src/control/zoom/ZoomControl.cpp
```diff
@@ -27,12 +27,18 @@ void ZoomControl::zoomOneStep(bool zoomIn, double x, double y)
 {
 	XOJ_CHECK_TYPE(ZoomControl);
 
+	if(this->zoomPresentationMode)
```
andreasb242: This probably solves the ticket: Presentation mode not working?!

src/gui/MainWindow.cpp
```diff
@@ -532,32 +531,6 @@ bool MainWindow::onKeyPressCallback(GtkWidget* widget, GdkEventKey* event, MainW
 		//editing text - give that control
 		return false;
 	}
-	else if (event->keyval == GDK_KEY_Down)
```
andreasb242: Why is this removed?

src/gui/MainWindow.cpp
```diff
@@ -576,17 +549,6 @@ bool MainWindow::deleteEventCallback(GtkWidget* widget, GdkEvent* event, Control
 	return true;
 }
 
-bool MainWindow::configureEventCallback(GtkWidget* widget, GdkEventConfigure* event, Control* control)
```
andreasb242: Why is this removed, was this dead code from GTK2? ("configure-event")

ui/icons/hicolor/scalable/actions/presentation-mode.svg
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
andreasb242: As there is no about comment: You checked the license of the icon or created it by yourself?

peetCreative: I created a new logo, but it's really nothing saying and I would request someone to create a new one.

src/gui/MainWindow.cpp
peetCreative: The same is done in XournalView.cpp:235 but this is called first.

src/control/zoom/ZoomControl.cpp
peetCreative: Well, it's just disables Zoom during presentationMode.
Maybe we should do this rather in zoomSequenceChanges?

ui/icons/hicolor/scalable/actions/presentation-mode.svg
andreasb242: The question was only about the license...

peetCreative: I would just say CC, I will push sth soon.

src/gui/MainWindow.cpp
peetCreative: I added this some commits before but it didn't catched when the application was resized using tiling. I moved to the signal size-allocate from the widget and the ZoomControl is listening to it. Now everything is working.

src/control/Control.cpp
```diff
@@ -1523,50 +1529,43 @@ void Control::changePageBackgroundColor()
...
-	settings->setPresentationMode(presentationMode);
-	fireActionSelected(GROUP_PRESENTATION_MODE, presentationMode ? ACTION_VIEW_PRESENTATION_MODE : ACTION_NOT_SELECTED);
+	if(enabled)
+	{
+		// Disable PresentationMode if ZoomValue could not calculated correctly
+		zoom->updateZoomPresentationValue();
+		//TODO: write out to settings previous layout
+		setViewPairedPages(false);
+		setViewRows(0);
+		setViewColumns(0);
```
LittleHuba: Shouldn't that be `setViewColumns(1)` so we have one column for pages?

src/control/zoom/ZoomControl.cpp
LittleHuba: I would do it at the earliest convenient position in the code, to decrease the overhead. But essentially it is negligible.

src/gui/MainWindow.cpp
LittleHuba: XournalView is the better position to do this. As that is actually the view that should react to the event.

src/gui/XournalView.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include <gdk/gdk.h>
 
+#include <tuple>
+#include <math.h>
```
LittleHuba: You should use cmath instead or we revive a bug previously fixed.

src/control/Control.cpp
JJones780: LittleHuba is correct but now that you've merged in #1037 you should not change the layout settings here.
You should remove the calls to  setViewRows, setViewColumns and setViewPairedPages.
PR 1037 did not configure the Paired Pages so you'll have to remedy that by changing   LayoutMapper::configureForPresentation() so that line
`	bool isPairedPages = settings->isShowPairedPages();`
instead reads:
`	bool isPairedPages = false;`

JJones780: When the user leaves presentation mode the layout will revert back to using the settings.
Because Presentation mode ignores the users layout settings we should also grey out the layout settings ( View->Layout and view->Cols/Rows ) in the menu when presentation mode is active. 
_Can do in another PR._

src/gui/XournalView.cpp
```diff
@@ -615,14 +616,25 @@ void XournalView::zoomChanged()
...
 	layout->layoutPages();
 
-	zoom->scrollToZoomPosition(view);
+	if(zoom->isZoomPresentationMode() || zoom->isZoomFitMode())
+	{
+		scrollTo(currentPage);
+	}
+	else
+	{
+		std::tuple<double, double> pos = zoom->getScrollPositionAfterZoom(view);
```
JJones780: You can use std:tie here to be a bit more readable:
		int  scrollX , scrollY;
		std::tie( scrollX , scrollY ) = zoom->getScrollPositionAfterZoom(view);
		if( scrollX != -1 &&  scrollY != -1)
		{
			layout->scrollAbs(  scrollX , scrollY );
		}

src/gui/XournalView.cpp
```diff
@@ -615,14 +616,25 @@ void XournalView::zoomChanged()
...
+		if(std::get<0>(pos) != -1 && std::get<1>(pos) != -1)
+		{
+			layout->scrollAbs(std::get<0>(pos), std::get<1>(pos));
+		}
+	}
```
JJones780: You can use std:tie here to be a bit more readable:
```
int scrollX , scrollY;
std::tie( scrollX , scrollY ) = zoom->getScrollPositionAfterZoom(view);
if( scrollX != -1 && scrollY != -1)
{
layout->scrollAbs( scrollX , scrollY );
}
```

JJones780:
I can't seem to be able to use the pgup or pgdown keys to go more than one page away?
Also, it doesn't zoom in on the current page when I go into Presentation mode?

src/gui/XournalView.cpp
```diff
@@ -451,9 +455,6 @@ void XournalView::scrollTo(size_t pageNo, double yDocument)
 	int height = v->getDisplayHeight();
 
 	layout->ensureRectIsVisible(x, y, width, height);
-
-	// Select the page
-	control->firePageSelected(pageNo);
```
JJones780: Restore this... we need to firePageSelected here! 
Page navigation won't work without it.

JJones780: Restore this... we need to firePageSelected here!
Page navigation won't work without it.

src/control/Control.cpp
LittleHuba: @JJones780 we should do this in another PR as this one is already heavy enough. Additionally it is coming up with Merge Conflicts where we should not introduce another feature in fear of even more conflicts.

peetCreative: I disabled the menuitems now, this should be enough.

src/gui/XournalView.cpp
peetCreative: reverted it.

src/control/Control.cpp
```diff
@@ -1523,50 +1533,67 @@ void Control::changePageBackgroundColor()
...
-void Control::setViewPairedPages(bool pairedPages)
-{
-	XOJ_CHECK_TYPE(Control);
+		setViewLayoutVert(settings->getViewLayoutVert());
+		setViewLayoutR2L(settings->getViewLayoutR2L());
+		setViewLayoutB2T(settings->getViewLayoutB2T());
 
-	settings->setShowPairedPages(pairedPages);
-	fireActionSelected(GROUP_PAIRED_PAGES, pairedPages ? ACTION_VIEW_PAIRED_PAGES : ACTION_NOT_SELECTED);
+		setViewPairedPages(settings->isShowPairedPages());
```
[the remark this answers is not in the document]
peetCreative: It is meant to restore the the Layout, Cols/Rows after coming back from Presentationmode

src/control/Control.cpp
```diff
@@ -1523,50 +1533,67 @@ void Control::changePageBackgroundColor()
...
-}
+	fireActionSelected(GROUP_PRESENTATION_MODE, enabled ? ACTION_VIEW_PRESENTATION_MODE : ACTION_NOT_SELECTED);
+	//Disable Zoom
+	fireEnableAction(ACTION_ZOOM_IN, !enabled);
+	fireEnableAction(ACTION_ZOOM_OUT, !enabled);
+	fireEnableAction(ACTION_ZOOM_FIT, !enabled);
+	fireEnableAction(ACTION_ZOOM_100, !enabled);
+	fireEnableAction(ACTION_FOOTER_ZOOM_SLIDER, !enabled);
+	//Disable Viewmodes without unsetting them
+	fireEnableAction(ACTION_VIEW_PAIRED_PAGES, !enabled);
```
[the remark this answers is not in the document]
peetCreative: Well this is just supposed to disable the button and menu for zoom.
I have to check the paired pages behavior, while presentation.

peetCreative:
>     * The zoom behaviour is also not right.. it scrolls and zooms simultaneously ( when using control with the mouse wheel ).

I can't reproduce this.

peetCreative:
> When setting presentation mode it doesn't scroll to the proper page. There seem to be other issues with page numbers so this is likely not associated with your changes.

I noticed that the current focused page is not updated anymore. So if one is just scrolling further and not drawing or sth. The current page is not updated. And if then Presentation is set it goes back to the first or previous used page, which is not at all visible.

peetCreative:
There are also lots of issues with paired pages in presentation and fit zoom, when the pages are not equally width-sized.

peetCreative:
I also noticed that size-allocate of the xournalview GTK widget is not thrown consistently and is sometimes working sometimes not and changing working on setting Presentationmode. Very strange.

Issue #1001, opened 2019-03-21
Hang & System Crash when setting Highlight Pen


MuTaTeD:
The application hangs and crashes the XServer on Kubuntu crashes when trying to change the value of Big Cursor or Highlight Cursor options in the Preferences -> View panel.
Need to delete the settings.xml file before the app can be restarted else it will Hang & Crash again.

Version: V1.0.9 Buillt: Mar 20 2019
OS: Kubuntu 18.04 LTS

MuTaTeD:
Same problem as #842

MuTaTeD:
The only change it made to the settings file as compared to default is 
`<property name="presentationHideElements" value="mainMenubar,sidebarContents"/>`
`<property name="showBigCursor" value="true"/>`   <--
`<property name="highlightPosition" value="false"/>`

As soon as the check box is ticked, the app hangs and crashed the X system and forces me to restart the machine

MuTaTeD:
Tried to debug at the cost of some 20+ restarts of the system and finally seem to get the fix
by changing the big/highlight cursor size in file: xournalpp/src/gui/Cursor.cpp line: 326
```
	if (big || highlightPosition)
	{
		height = width = 100;
	}
```
to 
```
	if (big || highlightPosition)
	{
		height = width = 60;
	}
```
Also need to update the highlight circle drawing code on line: 383
```		cairo_arc(cr, centerX, centerY, 45, 0, 2 * 3.1415);```
to
```		cairo_arc(cr, centerX, centerY, 30, 0, 2 * 3.1415);```

It seems that KDE/GTK theme has some problem when trying to draw a cursor more than 75 pixels in size. 60 pixels seems to be a better compromise.

Issue #842, opened 2019-02-05
Ubuntu with KDE freezes when selecting "big cursor" or "highlight cursor" mode


francesco-p:
Hello everyone,
I'm running Ubuntu 
`uname -a`
`Linux shuttle 4.15.0-45-generic #48-Ubuntu SMP Tue Jan 29 16:28:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux `
with KDE Plasma version `5.12.7`

I have the following problem: 
if I select both the settings : "big cursor" or "highlight cursor" the os completely freezes, absolutely nothing works anymore. The only way to unlock the situation is to press the power button and forcing the shutdown of the machine. 

Now everytime I run xournalpp the computer automatically freezes. I uninstalled the app, checked the logs of dmesg but unfortunately I can't find any useful information since it seems that the os crashes before it is able to log.

LittleHuba:
Xournal++ writes it's own logs to `/home/[user]/.xournalpp/errorlogs/`.
Please attach such a log, so we can look for the bug.

francesco-p:
> Xournal++ writes it's own logs to `/home/[user]/.xournalpp/errorlogs/`.
> Please attach the most recent log with the bug, so we can look for it.

I cannot provide any logs because it seems that when it freezes, the system completely stops. When I look for the file that you specified I cannot find the correct log (with the correct timestamp).

I think you should replicate my system through a VM and check there!

LittleHuba:
Is your system pure Ubuntu with KDE installed or some derivative like Kubuntu?
Would you mind pointing out some more details so I can set up an appropriate VM?

francesco-p:
It's pure Ubuntu then I switched to KDE everything is at its latest release.

You can download the latest Ubuntu release then install KDE through `sudo apt install kubuntu-desktop`.

andreasb242:
This usually means X Server sync issue, e.g. we do something witin a non UI Thread, this can Kill X11. But it makes no sense...

LittleHuba:
I tried it with KDE on Arch and there it has no problems. Therefore it has to be something special with the system. I'm just now setting up the VM.

LittleHuba:
I was able to reproduce the crash using Ubuntu 18.10 with `kubuntu-desktop` installed.

Fortunately my system did not crash and Xournal++ was able to write the errorlog:
```
Date: Tue Feb  5 17:33:48 2019
Error: signal 11
[...]

PR #1029, opened 2019-03-29
Fixed BigCursor  Crash on KDE

MuTaTeD:
Crash in KDE when Big Cursor or Highlight Cursor option is enabled in settings.
Ref: Issue #842 & #1001


LittleHuba:
Do you actually know why this fixes the bug?

LittleHuba:
Thanks for debugging and fixing!

PR #1030, opened 2019-03-30
Layout relative page navigation

JJones780:
This PR adds page navigation with respect to the position of the pages as laid out.

Explores  keymappings as below:

Arrow Keys:  
   no modifier - as before, scrolls view
   shifted arrow and
   numeric keypad arrow keys - - selects page found in given direction.

Note: removes ability for zoom control to have focus so that keys don't mysteriously stop working after you adjust zoom with the mouse.


LittleHuba:
The modifiers sound nice. It is intuitive for me.

andreasb242:
For me it also sounds OK

PR #1031, opened 2019-04-01
WIP  Fix dots

JJones780:
This PR filters very short ( in time and length ) strokes but allows successive ones through... so the user can still dot in a diagram or make quick short strokes to fill in a drawing etc.
Settings:
See strokeFilter* in settings file. I considered putting it in the settings dialog but unsure where... suggestions? Just the ability to turn it off... or adjust the settings ( fixed options or each of the three variables ( time, length, successive timeout).
The filter is done on the event times... so pause for debug and make a bunch of strokes and the filter will properly apply to saved up strokes upon continue.


LittleHuba:
Im not quite sure if we should add this fix as it only fights the symptoms and not the illness so to say. 

I triaged the bug for some time: sometimes events in GDKNewInputDevice get mixed up, which creates said random dots and scratches. We should fix the problem in that class and not filter out the resulting strokes. 

Aditionally I see a problem with writing. What happens to the dot of an i for example. I don't want to tab the screen 8 times to get the dot. Not even 2 times to tell the truth.

andreasb242:
Could be problematic, may somebody write a manual how we can *easy* debug GTK? ;-)

src/gui/inputdevices/NewGtkInputDevice.cpp
```diff
@@ -323,12 +324,13 @@ bool NewGtkInputDevice::eventHandler(GdkEvent* event)
 	else if (event->type == GDK_BUTTON_PRESS || event->type == GDK_TOUCH_BEGIN)
 	{
 		input->copyAxes(event);
-		input->actionStart();
+		guint32 time = event->type == GDK_BUTTON_PRESS ? ((GdkEventButton *)event)->time : ((GdkEventTouch *)event)->time;  //or call gdk_event_get_time()
```
JJones780: Different events have different structures.. only the first few entries ( including event->type ) are the same and otherrwise you need to check the event->type and then cast to the right type and get the wanted field if present.See gdkEvents.h
Functions getEventTime and getEventState makes it easier.

JJones780:
Trying out another interesting advanced user feature...  Allow the user to select items without having to switch to select tool.
test

JJones780:
I might feel comfortable with this for everyday use now.  It's not so much a dot filter as a dot vs selection alogrithm.

JJones780:
Replaced by #1087 and #1099

PR #1044, opened 2019-04-03
Fix the Draw Arrow/Draw Coordinate System issue/inconsistency in the Tools menu

siliconninja:
Minor bug: Under the Tools menu, the "Draw Arrow" is duplicated (the bottom one does "Draw Coordinate System")
I built the latest version from the `xournalpp-git` AUR package.

Screenshot of bug:
xournalpp-minor-naming-bug-20190403

This time, I was able to find the source of the issue! :) The PR contains the fix. I tested it by building from source. Tested on Arch Linux, built manually with this tutorial: https://github.com/xournalpp/xournalpp/blob/master/readme/LinuxBuild.md
I also edited the translation files to make sure that it was consistently recognizing it across all languages (some of them had "" in msgstr, would that mean it would replace with nothing, or would it replace it with the default string? I'm not very familiar with potfiles. I kept the name lowercase to make it consistent with other places where we have "Draw coordinate system" (in the arrow/dropdown near the shape tool button))
Feel free to edit (I checked "allow edits from maintainers") if you see any inconsistencies or things you want changed.

Here is a screenshot of the fix:
Screenshot_20190403_191656

Thank you!
- siliconninja

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -921,6 +921,7 @@ msgstr "Kreslit čáru"
 #: ../src/gui/dialog/ButtonConfigGui.cpp:108
 #: ../src/gui/toolbarMenubar/ToolDrawCombocontrol.cpp:34
 #: ../src/gui/toolbarMenubar/ToolMenuHandler.cpp:470
+#: ../ui/main.glade:1128
 msgid "Draw coordinate system"
 msgstr ""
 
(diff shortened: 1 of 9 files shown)


LittleHuba:
Thanks for the fix!
Empty msgstr mean `gettext` will use the original string from the file. It is a sign for not available translations.

PR #1045, opened 2019-04-04
fixes 1042: Mouse grab: Can't move page up slowly

JJones780:
The hand tool was scrolling with int but the scrolling works with doubles. The result was that you couldn't move the page up nor to the left slowly using the grab tool.

### DIFF
--- a/src/gui/inputdevices/InputSequence.cpp
+++ b/src/gui/inputdevices/InputSequence.cpp
@@ -152,10 +152,6 @@ void InputSequence::handleScrollEvent()
 	// use root coordinates as reference point because
 	// scrolling changes window relative coordinates
 	// see github Gnome/evince@1adce5486b10e763bed869
-	if (lastMousePositionX  == (int)rootX && lastMousePositionY == (int)rootY)
-	{
-		return;
-	}
 
 	if (scrollOffsetX == 0 && scrollOffsetY == 0)
 	{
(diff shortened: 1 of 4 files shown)


src/gui/inputdevices/InputSequence.cpp
```diff
@@ -152,10 +152,6 @@ void InputSequence::handleScrollEvent()
 	// use root coordinates as reference point because
 	// scrolling changes window relative coordinates
 	// see github Gnome/evince@1adce5486b10e763bed869
-	if (lastMousePositionX  == (int)rootX && lastMousePositionY == (int)rootY)
-	{
-		return;
-	}
 
 	if (scrollOffsetX == 0 && scrollOffsetY == 0)
```
LittleHuba: This comparison is not good as doubles normally have some imprecision that will cause them to never be exactly `0`.
Use `abs(scrollOffsetX) < threshold` with some very small threshold (e.g. `0.001`)

JJones780: I  tried this:
```
if (std::abs(lastMousePositionX - rootX) < 0.1 && std::abs( lastMousePositionY - rootY) < 0.1 )
{
	return;
}
```

and put a breakpoint on return and it never got hit - tried a bunch of zoom levels.
I then tried  <1 as the condition and it still never hit the breakpoint!

So I think we can remove this check safely. 
I've had a chance to test it now that I'm awake ;)  Thanks for taking a look.

LittleHuba: I think it is just a debouncer that may take effect in hardware that is older and sends multiple events very fast. But I'm actually not sure if we require it here. Would need to check more of the code to be able to say that for sure.

Issue #992, opened 2019-03-19
make fails on Debian Stretch


agtig:
When following the build instructions and installing the dependencies, the cmake works fine, but the make command fails with the following
```
xournalpp/src/control/tools/EditSelectionContents.cpp: In member function ‘void EditSelectionContents::finalizeSelection(double, double, double, double, bool, Layer*, PageRef, XojPageView*, UndoRedoHandler*)’:
/home/sand/Downloads/xournal/xournalpp/src/control/tools/EditSelectionContents.cpp:406:35: error: call of overloaded ‘abs(double&)’ is ambiguous
  bool rotate = (abs(this->rotation) > __DBL_EPSILON__);
```
And following this is many references to overloaded 'abs(double&)'

I don't know how to solve it easily?

peetCreative:
Can you try to build it from this fork? Probably sth like this happens because of this: 
https://stackoverflow.com/questions/1374037/ambiguous-overload-call-to-absdouble

agtig:
It doesn't seem to change anything. What kind of information can I provide that is helpful?
I git clone the repository (this time the fork), create a build directory, enters the build directory and run 'cmake ..'

I get the following output:

`[ 32%] Building CXX object src/CMakeFiles/xournalpp-core.dir/control/tools/EditSelectionContents.cpp.o
/home/sand/Downloads/xournal/fork/xournalpp/src/control/tools/EditSelectionContents.cpp: In member function ‘void EditSelectionContents::finalizeSelection(double, double, double, double, bool, Layer*, PageRef, XojPageView*, UndoRedoHandler*)’:
/home/sand/Downloads/xournal/fork/xournalpp/src/control/tools/EditSelectionContents.cpp:406:35: error: call of overloaded ‘abs(double&)’ is ambiguous
  bool rotate = (abs(this->rotation) > __DBL_EPSILON__);
                                   ^
In file included from /usr/include/c++/6/cstdlib:75:0,
                 from /usr/include/c++/6/ext/string_conversions.h:41,
                 from /usr/include/c++/6/bits/basic_string.h:5417,
                 from /usr/include/c++/6/string:52,
                 from /home/sand/Downloads/xournal/fork/xournalpp/src/util/XournalType.h:23,
                 from /home/sand/Downloads/xournal/fork/xournalpp/src/control/ToolEnums.h:14,
                 from /home/sand/Downloads/xournal/fork/xournalpp/src/control/ToolBase.h:15,
[...]

PR #1060, opened 2019-04-09
Fix errors on build on GCC 6.3.0: error: call of overloaded ‘abs(double)’ is ambiguous

JJones780:
Minor fix - discovered issue when compiling on latest MX Linux.
abs(num)    ->     std:abs(num)


LittleHuba:
Does this fix #992?

JJones780:
Yes, it fixes #992.  Thanks for pointing that out.

PR #1063, opened 2019-04-10
WIP  fixes #742 - Circles from center - use Control Key

### LINKED ISSUE #742: Circles would be nice to be drawn with the starting position as the center.
Most CAD tools I'm familiar with draw circles using the point you click as the center.  The way it is, I find it hard to make the circles the way I want.

JJones780:
Thought it would be a quick modification but turned out to be a tiny bit more work.

A few rough edges:
If you have the control key down it never gets to the code to start drawing. You have to press control after starting the circle.
Also - doesn't redraw the circle unless the mouse is moved..

### DIFF
--- a/src/control/tools/BaseStrokeHandler.cpp
+++ b/src/control/tools/BaseStrokeHandler.cpp
@@ -184,6 +184,13 @@ void BaseStrokeHandler::onButtonPressEvent(const PositionInputData& pos)
 }
 
 
+void BaseStrokeHandler::redrawShape( const PositionInputData& pos )
+{
+	XOJ_CHECK_TYPE(BaseStrokeHandler);
+	
+	this->drawShape( this->currPoint, pos );
+}
+
 void BaseStrokeHandler::modifyModifiersByDrawDir(double width, double height,  bool changeCursor)
 {
 	XOJ_CHECK_TYPE(BaseStrokeHandler);
--- a/src/control/tools/BaseStrokeHandler.cpp
+++ b/src/control/tools/BaseStrokeHandler.cpp
@@ -114,7 +114,7 @@ bool BaseStrokeHandler::onMotionNotifyEvent(const PositionInputData& pos)
 	this->redrawable->repaintRect(stroke->getX(), stroke->getY(),
 								  stroke->getElementWidth(), stroke->getElementHeight());
 
-	drawShape(currentPoint, pos.isShiftDown());
+	drawShape(currentPoint, pos);
 	
 	rect.add(stroke->boundingRect());
 	double w = stroke->getWidth();
--- a/src/control/tools/BaseStrokeHandler.cpp
+++ b/src/control/tools/BaseStrokeHandler.cpp
@@ -4,7 +4,10 @@
 #include "control/Control.h"
 #include "control/layer/LayerController.h"
 #include "undo/InsertUndoAction.h"
+#include "gui/Cursor.h"
 #include <cmath>
+
+#define FIX_DRAW_TYPE_LOCK_DISTANCE 50 
 
 BaseStrokeHandler::BaseStrokeHandler(XournalView* xournal, XojPageView* redrawable, PageRef page)
  : InputHandler(xournal, redrawable, page)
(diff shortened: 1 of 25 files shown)


morrolinux:
Just to keep consistency would you mind using Shift key instead of Ctrl for doing that? (almost) all the other shapes behave differently if shift is held down, so that'd be the right modifier

JJones780:
> Just to keep consistency would you mind using Shift key instead of Ctrl for doing that? (almost) all the other shapes behave differently if shift is held down, so that'd be the right modifier

"Ellipse"/Circle and Rectangle/Square now have similar capabilities with the same modifiers.
The rectangle tool was using shift to make squares so I copied that for the (now) ellipse tool to make circles.  They both use control to center and alt to temporarily toggle the grid snap setting.   
GIMP does it this way for Control and Shift at least.
Inkscape is the opposite.

It can be changed easily - I wont' merge until I hear back from a few of you on this and anything else.

Rmano:
I like this, but at the same time, I see it has a problem: how to interact with these tools if you have no keyboard? For example, a folded 2-in-1 device used to take notes. It would be great to have some additional control you can touch with the pen (like the rotation control) that let you do the same without a keyboard.

morrolinux:
@Rmano good point on that. I also have a 2 in 1 and would not benefit from this feature without some "button" or touch control. @JJones780 how about making the circle tool properties selectable like for the pen?
image

eliasrg:
I agree that it is best to minimize keyboard-only features (I also have a 2 in 1). From a UI standpoint, I'm thinking it might be best to make ellipses and circles two separate tools, and the same for rectangles and squares, and then adding a button "Center Shapes"? Or keep them the same tools and add another button to act as the Shift modifier (not sure about what to name it though).

JJones780:
Please continue the discussion on tablet mode modifiers in Issue #1066:
Access to modified ( shift, ctrl etc. ) tool options without a keyboard.
See you there.

src/control/tools/CircleHandler.cpp
```diff
@@ -37,10 +39,14 @@ void CircleHandler::drawShape(Point& c, const PositionInputData& pos)
 	}
 	else
 	{
+		
 		double width = c.x - this->startPoint.x;
 		double height = c.y - this->startPoint.y;
+		
+		this->setModifiers(width, height,  pos);	// sets this->modShift and this->modControl
```
JJones780: For classes derived from BaseStrokeHandler:  This is pretty much all that's needed to use Shift and Control modifiers along with initial directional modifier. It sets this->modShift and this->modControl for use below. There's an optional boolean not used here to turn off cursor updates.

src/gui/Cursor.cpp
```diff
@@ -408,3 +408,26 @@ GdkCursor* Cursor::createHighlighterOrPenCursor(int size, double alpha)
...
 
+
+
+void Cursor::setTempCursor(GdkCursorType type)
+{
+	XOJ_CHECK_TYPE(Cursor);
+	
+	MainWindow* win = control->getWindow();
+	
+	if( !win) return;
```
JJones780: Couldn't bare to put in curlys here...  this is pretty readable.

src/gui/Cursor.cpp
```diff
@@ -408,3 +408,26 @@ GdkCursor* Cursor::createHighlighterOrPenCursor(int size, double alpha)
...
+	
+	GdkWindow* window = gtk_widget_get_window(xournal->getWidget());
+	
+	if( !window ) return;
+	
+	GdkCursor* cursor = gdk_cursor_new_for_display(gdk_window_get_display(window), type);
+	gdk_window_set_cursor(gtk_widget_get_window(xournal->getWidget()), cursor);
+	gdk_window_set_cursor(window, cursor);
+	g_object_unref(cursor);
+}
```
JJones780: Pretty much copied this from elsewhere... a quick hack to try out cursors. Not well tested - happy to hear back on this.
Oops.. forgot.. Needs a filter to not keep setting over and over.

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -4,8 +4,11 @@
 #include "control/Control.h"
 #include "control/layer/LayerController.h"
 #include "undo/InsertUndoAction.h"
+#include "gui/Cursor.h"
 #include <cmath>
 
+#define FIX_DRAW_TYPE_LOCK_DISTANCE 50
```
JJones780: This will be a prefs setting if we decide to merge.

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -178,3 +181,60 @@ void BaseStrokeHandler::onButtonPressEvent(const PositionInputData& pos)
...
+						break;
+					case 3:
+						xournal->getCursor()->setTempCursor(GDK_TOP_RIGHT_CORNER);
+						break;
+					case 0:
+						xournal->getCursor()->setTempCursor(GDK_BOTTOM_LEFT_CORNER);
+						break;
+					case 2:
+						xournal->getCursor()->setTempCursor(GDK_TOP_LEFT_CORNER);
+						break;
```
JJones780: Pretty ugly cursors...  but at least they indicate something is happening.  Good enough for a default base class implementation?

src/control/tools/CircleHandler.cpp
```diff
@@ -5,6 +5,8 @@
 #include "undo/InsertUndoAction.h"
 #include <cmath>
 
+#define FIX_DRAW_TYPE_LOCK_DISTANCE 10
```
JJones780: oops.. leftover.. will remove

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -178,3 +181,60 @@ void BaseStrokeHandler::onButtonPressEvent(const PositionInputData& pos)
...
+		}
+	}
+	else
+	{
+		gestureShift = this->drawModifier & SHIFT;
+		gestureControl = this->drawModifier & CONTROL;
+	}
+	
+	this->modShift = pos.isShiftDown() ==  !gestureShift;
+	this->modControl = pos.isControlDown() == !gestureControl;
```
JJones780: The actual shift and Control keys will toggle any directional modifier... which may be confusing?  Not sure about this. try it out and comment.  Could make this only toggle when the directional modifier has been fixed ( which happens once the user has expanded the rect/circle past a certain point).

Rmano:
Works quite well! But I think a bit more visual feedback is needed when starting the stroke.

JJones780:
Set up the stroke tools to react to keypresses. They now redraw on shift and control without mouse movement.

**UI Change:**  Arrow and Ruler were using shift to toggle snapping. Changed to match other tools by using ALT.   -- comments?

src/control/tools/BaseStrokeHandler.h
```diff
@@ -28,12 +37,24 @@ class BaseStrokeHandler : public InputHandler
...
+	
 protected:
 	void snapToGrid(double& x, double& y);
+	Point currPoint;
+	/**
+	 * modifyModifiersByDrawDir  -  toggle shift and control modifiers depending on initial drawing direction/
+	 */
+	void modifyModifiersByDrawDir(double width, double height, bool changeCursor = true);
+	bool modShift = false;
+	bool modControl = false;
```
andreasb242: Please move the Attributes to the other attributes down, and don't mix it with methods

src/control/tools/CircleHandler.cpp
```diff
@@ -18,53 +19,90 @@ CircleHandler::~CircleHandler()
...
+			height = (width * signW) * signH;
 		}
-		// set resolution proportional to radius
-		double diameter = sqrt(pow(c.x - p.x, 2.0) + pow(c.y - p.y, 2.0));
-		int npts = (int) (diameter * 2.0);
-		double center_x = (c.x + p.x) / 2.0;
-		double center_y = (c.y + p.y) / 2.0;
-		double angle = atan2((c.y - p.y), (c.x - p.x));
 
+		double diameterX, diameterY;
```
andreasb242: C++ does not initialize vars, and C++ not always warns about unitialized vars (like Java does) therefore I would initialize all vars every time.

The optimizer will remove unused code, therefore it's not a performance issue etc.

src/control/tools/RectangleHandler.h
```diff
@@ -16,12 +16,14 @@
...
-	RectangleHandler(XournalView* xournal, XojPageView* redrawable, PageRef page);
+	RectangleHandler(XournalView* xournal, XojPageView* redrawable, PageRef page, bool flipShift = false, bool flipControl = false);
 	virtual ~RectangleHandler();
 
 private:
-	virtual void drawShape(Point& currentPoint, bool shiftDown);
-
+	virtual void drawShape(Point& currentPoint, const PositionInputData& pos);
+	Point startPoint;
+	bool started = false;
```
andreasb242: Also move down

JJones780:
@ eliasrg Thanks for your input.  We can also have multiple tools as you suggest. Support is included with this merge but creating the actual UI buttons is left for another ( for issue #1066 ).

PR #1067, opened 2019-04-11
fixing snapping to grid when a margin is used

morrolinux:


morrolinux:
So... that worked for me. I might be approching the problem in the wrong way, maybe check if it makes sense before merging

Rmano:
Works for me, too; nicely done.

morrolinux:
No major complaints from anyone so I guess I'm going to merge :)

Issue #1086, opened 2019-04-21
EditSelect buttons don't scale with different screen resolutions


JJones780:
EditSelect buttons don't scale with different screen resolutions.

PR #1087, opened 2019-04-21
Edit select

### LINKED ISSUE #1085: Negative scaling of strokes causes issues
Selecting a stroke and "scaling it over" ( i.e. negative scale on x or y axis ) causes the stroke to dissappear and other odd behaviour ( selecting multiple items causes others to disappear etc ).

JJones780:
Split out EditSelection related code from #1031 and made some more improvements.
fixes #1085 and #1086  ;) 
Also adds another handle to allow deleting of selected item.
More work yet to do on EditSelection to fix other issues I reported earlier.


JJones780:
Fixed the mouseover of rotated EditSelect so you can grab the handles and continue editing. Rotating again works well but the resizing does not.

PR #1088, opened 2019-04-21
WIP : PreStroke Pressure Filter

JJones780:
first try at pre-stroke filter to remove ugly 'tics' from stylus use.

### DIFF
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -13,6 +13,11 @@
 #include <gdk/gdk.h>
 #include <cmath>
 
+													//Testing stroke->removePrePressure() filter
+static float GAIN = 1.2;				// remove points until less than this increase in pressure.  >1 on way to peak, <1 after peak
+static int BACKOFF = 0;				// but keep this many points before above gain triggered.
+
+
 StrokeHandler::StrokeHandler(XournalView* xournal, XojPageView* redrawable, PageRef page)
  : InputHandler(xournal, redrawable, page),
    surfMask(NULL),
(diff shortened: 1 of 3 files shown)


JJones780:
You can see the strokes get cleaned up by this filter after slight delays:
Note the "st" in testing.
testingPreStrokeFilter

Note: This fixes #1089

It would be nice to filter immediately, but for now this demonstrates the filter much better. Immediate might not be an option depending on how the pressure data is handled.

LittleHuba:
I'll take a look at the code tomorrow. But even if we include this it should be opt-in. This is not an intuitive behaviour even if it happens to be an improvement in your case.

JJones780:
Be aware that this is marked Work In Progress (WIP).
It's got static variables for what will eventually be settings variables and probably just a simple checkbox in preferences to enable it if it's decided it is good to have in xournalpp. For now I/you can change the vars in the debugger to experiment with different values for the filtering. The current values seem to work quite well and result in the stroke the user intended to make.

LittleHuba:
Can we remove this PR now that you do not want to merge it anymore or have you further plans for it @JJones780?

schuhumi:
It would be great if it was possible to set a pressure curve just like in Krita for example: https://docs.krita.org/en/reference_manual/preferences/tablet_settings.html

That way you can:
 - Remove tics
 - adjust the pressure you want to put on the stylus for maximum line width
 - if the line can have curvature: adjust if the stylus feels like it had a soft/hard tip

This is the feature that I miss most in Xournal++

LittleHuba:
> It would be great if it was possible to set a pressure curve just like in Krita for example: https://docs.krita.org/en/reference_manual/preferences/tablet_settings.html
> 
> That way you can:
> 
> * Remove tics
> * adjust the pressure you want to put on the stylus for maximum line width
> * if the line can have curvature: adjust if the stylus feels like it had a soft/hard tip
> 
> This is the feature that I miss most in Xournal++

Please move this to a feature request. It is a very good idea but will get lost here 😉

Issue #1084, opened 2019-04-20
Music score


snino64:
¿Is it possible to add music score to set of papers background?
Thanks.

snino64:
That's great!
Thanks

peetCreative:
I have to adapt the old Page template settings. So I need bit more time... But I'm about it.

veecue:
This was fixed in #1309

PR #1090, opened 2019-04-21
Implement staves for music notes #1084

peetCreative:
Implemented staves (music scores) by extending lined templates.
I also rearranged some default settings so now default page format is lined with out border line. But with new default settings it's with borderline. We anyway should make the default page format selectable in settings.


andreasb242:
> We anyway should make the default page format selectable in settings.

You can customize the page settings already.
grafik

grafik


But we may should link it in the Settings, as this is mostly overseen.
There don't change the default in XojPage, it makes no sense, this is only an initial value, it should not really affect new pages, as they are applied from the settings on creation.

Technius:
Is this still being worked on?

Issue #1081, opened 2019-04-19
Crash on export


markwmuller:
I can reliably cause X++ to crash, by doing the following steps:

1. Open an XOPP file (or create a new one)
2. Click "file->export as pdf".
3. Enter a name, click "OK"
4. On "Export" dialog (where we can select page ranges), click "Cancel"
5. Draw a single stroke in X++
6. The program is now frozen. 

I'm running latest master on linux.

LittleHuba:
Unable to reproduce. Please check if this issue persists @markwmuller

markwmuller:
Seems fixed on latest! (specifically, I can't get that export dialogue to show up any more where one specifies which pages of the file to export to the PDF, which would precipitate the failure).

Thanks!

LittleHuba:
The dialog shows after selecting the path you want to export to (which is awkward by the way...)

The bug persists btw.

PR #1099, opened 2019-04-28
Tap select

JJones780:
Adds experimental feature to ignore short scratches which in turn allow the user to select items without changing from the drawing tool. 
tapSelect


JJones780:
This and #1081 replace  #1031

Issue #1074, opened 2019-04-16
Xournal loses all saved settings sometimes


xinxinw1:
Most of the times, when I close xournal, the settings I set (like pen color, eraser type and shape, stylus button settings) are saved and restored when I open it again. But seemingly randomly, sometimes when I open xournal again, my settings seem to be wiped and I see the default settings again. I'm not sure how to reproduce this, but it seems to happen every 8 opens and closes or so? It's really frustrating when I have to reapply all my settings every couple days :(

JJones780:
Had a quick look through... we're going to need some more hints.  Please run Xournalpp from the command line for the next while and let us know of any errors that are reported.
Which OS are you using?
Do you change locale on your machine.. ie between languages and/or numerical radix systems ( 100.00 vs 100,00 ) ?  Do you use multiple language keyboards?

xinxinw1:
Ok, I just switched to a new computer so let's see if it keeps happening, but details of my previous setup:
- OS: Arch Linux, Desktop environment: MATE
- Xournal version: 1.0.10
- Change locale: No, I always had locale set to en_US.UTF-8
- Language keyboards: Yes, I had a Chinese keyboard

xinxinw1:
Some good news, I was just able to reproduce this somewhat more frequently on my old computer:

It happens if I open around 3 xournalpp windows simultaneously, like if I open 3 terminals and run xournalpp on each one. Often two is also enough. I can tell the bug happened because I set my default pen color to black and on the new window it changes to blue.

Here is the command line output on each run (it is the same when the failure happens and when it doesn't happen):

```
$ xournalpp
ALSA lib pcm_dsnoop.c:638:(snd_pcm_dsnoop_open) unable to open slave
ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
[...]

PR #1104, opened 2019-04-30
save/read settings with fixed locale

JJones780:
This was the only thing that came to mind regarding #1074.
If the user changed his locale then xournalpp would have problems loading floating point settings the next time around.

This PR changes all of the settings double conversion to use 
g_ascii_dtostr instead of g_dtostr   
and g_ascii_strtoll instead of atof.  

The ascii variants always use locale 'C'.

We were already mainly using g_ascii_strtoll for settings.

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -179,7 +179,7 @@ void Settings::parseData(xmlNodePtr cur, SElement& elem)
 			}
 			else if (sType == "double")
 			{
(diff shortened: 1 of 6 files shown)


JJones780:
SaveHandler::saveTo()  temporarily sets the locale using 
`setlocale(LC_NUMERIC, "C"); `
and is used for saving XMLnodes but not when settings is saved.

PR #1106, opened 2019-04-30
Fix TextEditor not updating correctly after text is pasted

### LINKED ISSUE #1103: Pasted text appears only after after additional writing, or after unfocusing the textbox
The behaviour I have been noticing in the current Xournal++ version (I have a memory that this did not use to happen 1.5 - 2 months ago), is that if I paste text in a textbox [no matter whether new or containing some text already], the pasted text will not appear immediately, but only if:

- I click outside of the textbox (or anyway unfocus said textbox)

- I write additional text following my Ctrl+V action, so that instantaneously the pasted text will appear

Technius:
Since text pasting is asynchronous according to https://developer.gnome.org/gtk3/stable/GtkTextBuffer.html#gtk-text-buffer-paste-clipboard, the repaint calls are
triggered before the text is actually pasted. This commit attaches the repaint
calls to the `paste-done` signal that is fired by the text buffer once the
pasting completes.

This fixes #1103.

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -29,6 +29,8 @@ TextEditor::TextEditor(XojPageView* gui, GtkWidget* widget, Text* text, bool own
 	string txt = this->text->getText();
 	gtk_text_buffer_set_text(this->buffer, txt.c_str(), -1);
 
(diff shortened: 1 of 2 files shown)


LittleHuba:
Thanks for the fix!

PR #1112, opened 2019-05-03
Improve file dialog history handling

### LINKED ISSUE #1056: File open dialogue does not respect current working directory
Calling the file open dialogue while having a file already opened, I would expect the new file to be searched for in the same directory as the current file (which I call current working directory in the title of this report; CWD for short). Instead, the list of recently opened files is selected which is inconvenient. When wanting to edit another file in the CWD, one has to browse through the whole file system hierarchy. On the other hand, the list of recently opened files is always available as a special filter option in the file open dialogue.
This happens on Ubuntu Budgie 18.10 with 1.0.11 (compiled on April 5th 2019, 22:39:29). The desktop environment uses the Nautilus file browser if that should matter. The issue sort of reminds me of the now closed #990 but I do not recognize any obvious connection.
Thanks for your consideration!

### LINKED ISSUE #1108: Last save/image path is not saved properly in the settings file
While working on a fix for #1056, I noticed that the `lastSavePath` and `lastImagePath` fields of `Settings` are not saved properly:
```xml
  <property name="lastSavePath.str()" value=""/>
  <property name="lastOpenPath.str()" value=""/>
```
I'll probably fix this bug and fold it into a more general "fix and enhance open/save history" PR.

Edit: The culprit is how the [`WRITE_STRING_PROP` macro](https://github.com/xournalpp/xournalpp/blob/f2373684e62c9c5f2c0ae5a7b3c513f36504bf18/src/control/settings/Settings.cpp#L14) is being used:

https://github.com/xournalpp/xournalpp/blob/f2373684e62c9c5f2c0ae5a7b3c513f36504bf18/src/control/settings/Settings.cpp#L837

which will use its argument as the name of the key, hence `lastSavePath.str()`.

Technius:
Fixes #1108 and fixes #1056. In particular:
* Adds a new setting that stores the last opened file.
* Fixes `lastSaveFile` and `lastImageFile` being incorrectly saved in the settings.
* The file open dialog now displays the folder of the last opened file by default.
* When opening a file, the folder of the last saved file is added as a shortcut on the left sidebar. When saving, the folder of the last opened file is added.

### DIFF
--- a/src/control/settings/Settings.h
+++ b/src/control/settings/Settings.h
@@ -204,6 +204,12 @@ public:
 	void setLastSavePath(Path p);
 	Path getLastSavePath();
 
+    /**
+     * The last open path
+     */
+    void setLastOpenPath(Path p);
+	Path getLastOpenPath();
+
 	void setLastImagePath(Path p);
 	Path getLastImagePath();
 
@@ -514,6 +520,11 @@ private:
 	 *  The last saved folder
 	 */
 	Path lastSavePath;
+
+    /**
+	 *  The last opened folder
+	 */
+	Path lastOpenPath;
 
 	/**
 	 *  The last "insert image" folder
--- a/src/control/settings/Settings.h
+++ b/src/control/settings/Settings.h
@@ -204,10 +204,10 @@ public:
 	void setLastSavePath(Path p);
 	Path getLastSavePath();
 
-    /**
-     * The last open path
-     */
-    void setLastOpenPath(Path p);
+	/**
+	 * The last open path
+	 */
+	void setLastOpenPath(Path p);
 	Path getLastOpenPath();
 
 	void setLastImagePath(Path p);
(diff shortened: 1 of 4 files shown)


src/control/settings/Settings.h
```diff
@@ -204,6 +204,12 @@ class Settings
 	void setLastSavePath(Path p);
 	Path getLastSavePath();
 
+    /**
+     * The last open path
+     */
+    void setLastOpenPath(Path p);
+	Path getLastOpenPath();
```
LittleHuba: The indents are somewhat awkward. No deal breaker just something to look for in the future.

src/control/settings/Settings.h
```diff
@@ -515,6 +521,11 @@ class Settings
 	 */
 	Path lastSavePath;
 
+    /**
+	 *  The last opened folder
+	 */
+	Path lastOpenPath;
```
LittleHuba: Also the indents 😉

src/control/settings/Settings.h
Technius: Oh, looks like that was before I set up my editor configuration to match the coding conventions (I use spaces by default). I'll fix that shortly.

andreasb242: The formatting is not everywhere correct, even I did a complete reformatting some time ago.
There are so many developers now, it's hard to keep them im sync, I think features are more important than correct formatted code.... ;-)

Issue #1100, opened 2019-04-28
Inserting Latex Crashes Xournal++


13hannes11:
Trying to add Latex formula results in the error message `LaTeX PDF file, FileError: Failed to open file /home/hannes/.xournalpp/tex/tex.pdf No such file or directory`
After that it's possible to enter the formula which results in the program crashing with the same error message appearing again.

Latex package:
pdfTeX 3.14159265-2.6-1.40.18 (TeX Live 2017/Debian)
kpathsea version 6.2.3

That's the log:

`Date: Sun Apr 28 19:32:04 2019
Error: signal 11
[bt]: (0) xournalpp(+0x1a02cd) [0x55c3abfdb2cd]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x3ef20) [0x7fb21b0b5f20]
[bt]: (2) xournalpp(_ZN7Element16getElementHeightEv+0x1) [0x55c3abfb1291]
[bt]: (3) xournalpp(_ZN10Redrawable15rerenderElementEP7Element+0x1d) [0x55c3abf5c5cd]
[bt]: (4) xournalpp(_ZN15LatexController14insertTexImageEv+0xd9) [0x55c3abeda909]
[bt]: (5) xournalpp(_ZN15LatexController3runEv+0x2b8) [0x55c3abedacc8]
[bt]: (6) xournalpp(_ZN7Control8runLatexEv+0xa9) [0x55c3abed6059]
[bt]: (7) xournalpp(_ZN7Control15actionPerformedE10ActionType11ActionGroupP9_GdkEventP12_GtkMenuItemP14_GtkToolButtonb+0xd4e) [0x55c3abed75ce]
[bt]: (8) xournalpp(_ZN12AbstractItem9activatedEP9_GdkEventP12_GtkMenuItemP14_GtkToolButton+0xf3) [0x55c3abf8f623]
[bt]: (9) /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x19d) [0x7fb21d62510d]
[bt]: (10) /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2305e) [0x7fb21d63805e]
[bt]: (11) /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0xa75) [0x7fb21d640715]
[...]

PR #1113, opened 2019-05-03
Improve/fix Latex tool usability

### LINKED ISSUE #1095: Latex error popup when inserting backslash
I love the Latex feature, however I encounter a bug when the input provided to the input field does not compile to a valid PDF file. This happens either when I try to insert a single backslash (which happens frequently when trying to write something like "\int_0^1 f(x) dx) or when deleting the whole input. 

In either case, an error popup is displayed saying something along the lines of "Couldn't open LaTeX PDF file. PDF document is damaged". I suspect that this is due to the incomplete input when typing in a single backslash for example. 

Xournal++ doesn't crash, but it is annoying that you always have to dismiss the popup in order to continue typing. It happens rarely that the popup shows multiple times which is especially annoying. 

Thanks for the otherwise great app! 
Screenshot

Technius:
* [x] Improve error handling (Fixes #1095 somewhat. Possibly prevents #1100).
* [x] ~Set the rendered font color to the currently selected color in the toolbar~ (should be brought up in a separate issue)
* [x] Render asynchronously to avoid the noticeable pauses on keypresses (~100ms on my computer) 
* [x] ~Since pdflatex is relatively slow, delay rendering until 200ms have passed since the last key was entered~ (doesn't bring a noticeable improvement to the user experience)
* [x] ~Make LaTeX tool selectable, with click to bring up the dialog instead of instantly bringing up the dialog and placing the formula in the top left corner~ (too many changes; will bring up in another issue)
* [x] Disable OK button when the entered formula is invalid
* [x] Display a message that explains why the OK button is disabled
* [x] Generate Latex previews in a temporary directory instead of in the config directory
* [x] Include `amssymb` (#1128)

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -41,7 +41,6 @@ const char* LATEX_TEMPLATE_2 =
 LatexController::LatexController(Control* control)
 	: control(control),
 	  doc(control->getDocument()),
-	  // NOTE: Could get a race condition if multiple Xournal++ processes are running
 	  texTmp(Util::getTmpDirSubfolder("tex"))
 {
 	XOJ_INIT_TYPE(LatexController);
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -26,16 +26,16 @@ const char* LATEX_TEMPLATE_1 =
 	R"(\documentclass[crop, border=5pt]{standalone})" "\n"
 	R"(\usepackage{amsmath})" "\n"
 	R"(\usepackage{ifthen})" "\n"
-	R"(\begin{document})" "\n"
+	R"(\newlength{\pheight})" "\n"
 	R"(\def\preview{\(\displaystyle)" "\n";
 
 const char* LATEX_TEMPLATE_2 =
 	"\n\\)}\n"
-	R"(\newlength{\pheight})" "\n"
-	R"(\settoheight{\pheight}{\hbox{\preview}})" "\n"
-	R"(\ifthenelse{\pheight=0.0pt})" "\n"
+	R"(\begin{document})" "\n"
+	R"(\settoheight{\pheight}{\preview} %)" "\n"
+	R"(\ifthenelse{\pheight=0})" "\n"
 	R"({\GenericError{}{xournalpp: blank formula}{}{}})" "\n"
-	R"({\preview})" "\n"
+	R"(\preview)" "\n"
 	R"(\end{document})" "\n";
 
 LatexController::LatexController(Control* control)
@@ -303,7 +303,9 @@ void LatexController::onPdfRenderComplete(GPid pid, gint returnCode, PdfRenderCa
 
 void LatexController::setUpdating(bool newValue)
 {
+	XOJ_CHECK_TYPE(LatexController);
 	GtkWidget* okButton = this->dlg->get("texokbutton");
+	bool buttonEnabled = true;
 	if ((!this->isUpdating && newValue) || (this->isUpdating && !newValue))
 	{
 		// Disable LatexDialog OK button while updating. This is a workaround
@@ -311,12 +313,15 @@ void LatexController::setUpdating(bool newValue)
 		// is open; 2) the preview is generated asynchronously; and 3) the `run`
 		// method that inserts the TexImage object is called synchronously after
 		// the dialog is closed with the OK button.
-		gtk_widget_set_sensitive(okButton, !newValue);
+		buttonEnabled = !newValue;
 	}
+
 
 	// Invalid LaTeX will generate an invalid PDF, so disable the OK button if
 	// needed.
-	gtk_widget_set_sensitive(okButton, this->isValidTex);
+	buttonEnabled = buttonEnabled ? this->isValidTex : buttonEnabled;
+
+	gtk_widget_set_sensitive(okButton, buttonEnabled);
 
 	GtkLabel* errorLabel = GTK_LABEL(this->dlg->get("texErrorLabel"));
 	gtk_label_set_text(errorLabel, this->isValidTex ? "" : "The formula is empty when rendered or invalid.");
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -318,6 +318,9 @@ void LatexController::setUpdating(bool newValue)
 	// needed.
 	gtk_widget_set_sensitive(okButton, this->isValidTex);
 
+	GtkLabel* errorLabel = GTK_LABEL(this->dlg->get("texErrorLabel"));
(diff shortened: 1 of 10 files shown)


LittleHuba:
Rendering every 200ms is no good idea for battery consumption. But we should do it asynchronously after every keypress.

Technius:
To clarify, I meant to delay rendering until 200ms have passed after the last key was entered. I've updated the description to reflect that.

src/control/LatexController.cpp
```diff
@@ -333,6 +448,12 @@ TexImage* LatexController::loadRendered()
 	Path pdfPath = texTmp + "/tex.pdf";
 	GError* err = NULL;
 
+	if (!pdfPath.exists())
+	{
+		g_warning("LaTeX preview PDF file does not exist");
```
LittleHuba: Also show this to the user.

src/control/LatexController.cpp
```diff
@@ -14,15 +14,29 @@
...
+	R"(\documentclass[crop, border=5pt]{standalone})" "\n"
+	R"(\usepackage{amsmath})" "\n"
+	R"(\usepackage{ifthen})" "\n"
+	R"(\newlength{\pheight})" "\n"
+	R"(\def\preview{\(\displaystyle)" "\n";
 
 const char* LATEX_TEMPLATE_2 =
-"\n\\)\n"
-"\\end{document}\n";
+	"\n\\)}\n"
```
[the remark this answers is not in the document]
Technius: You can't insert newlines in an R string. It would have to be `"\n" R"(\))" "\n"` or something uglier (like splitting it across multiple lines).

src/control/LatexController.cpp
```diff
@@ -60,15 +74,13 @@ bool LatexController::findTexExecutable()
...
 {
 	XOJ_CHECK_TYPE(LatexController);
+	g_assert(!this->isUpdating);
 
 	string texContents = LATEX_TEMPLATE_1;
-	texContents += currentTex;
+	texContents += this->currentTex;
 	texContents += LATEX_TEMPLATE_2;
 
 	string texFile = texTmp + "/tex.tex";
```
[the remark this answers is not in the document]
Technius: Seems like a good issue to fix. One problem: latex generates multiple files for every PDF (there's a `.aux`, a `.div`, and `.log` file too) which could end up generated a lot of unwanted garbage.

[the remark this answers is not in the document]
Technius: Looks like an easy way is to dump the files in `g_get_temp_dir()` instead of `.xournalpp`.

src/control/LatexController.cpp
Technius: Actually, a huge portion of this file could be refactored and simplified, which would make this check unnecessary.

src/control/LatexController.cpp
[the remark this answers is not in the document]
LittleHuba: Ahh didn't think of that...

src/control/LatexController.cpp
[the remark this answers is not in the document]
LittleHuba: Definitely put these files into the temp directory as OSs will clean that directory up regularily. the `.xournalpp` directory is actually only for settings and files that should survive restarts.

src/control/LatexController.cpp
LittleHuba: You may either clean this up in this commit or in a follow-up. Thats for you to decide. I'm happy either way.

Technius: Refactored in one of the commits I pushed today.

src/control/LatexController.cpp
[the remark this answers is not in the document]
Technius: Just pushed this change.

Technius:
Looks like `g_uuid_string_random` was implemented in glib 2.52, but Ubuntu 16.04 has an older version of glib.

src/control/LatexController.cpp
```diff
@@ -110,147 +129,202 @@ void LatexController::findSelectedTexElement()
...
-	delete dlg;
+	delete this->dlg;
+}
+
+void LatexController::triggerImageUpdate()
+{
+	if (this->isUpdating)
+	{
+		return;
+	}
```
LittleHuba: It would be better to abort the current image generation and create a new image with the newer data. That could fix the issue with the generating lag. I also see a problem when you type the last character while the image is currently updating.

Technius: On the first point: this doesn't offer any significant advantage to the user experience. This means that if is the user is rapidly typing in the latex, then the user won't be able to see the preview until he or she is done typing. Also, aborting is a bit hard here with the possible race conditions we can have. From the few hours I've been using the new implementation, the lag hasn't been too noticeable.

On the second point: in the callback, after rendering the preview, we check to see if the latex has been updated since the last time an update was triggered. If it has, we trigger another update (see bottom of `onPdfRenderComplete`).

LittleHuba: Ahh okay then you thought about these cases. All good for me!

Technius:
Since the last review:
* Rebased onto master
* Use pid in temporary folder name (fixes an outstanding bug)
* Added `amssymb` to list of packages (addresses new feature request)
* Fixed a regression to the cancel button caused by my refactor
* Some code cleanup

This PR should be ready to merge with some more testing (any help welcome!)

veecue:
Great work!
Two minor suggestions:
 * Remove the default x^2
 * Add a small indication that the current code produced a build failure (eg a small red bar)

Rmano:
I tried this and it's nicely working. I am a quite heavy user of LaTeX; my normal way of entering formulas is to copy and paste from my documents - so I have a couple of suggestions, that can be seen as enhancements proposal for the future:
1 - think about a button to pause the continuous compilation. When in slower machines and with complex formulas, sometimes the lag when the formula is wrong is quite noticeable.
2 - Ideally, the formula text box should be multiline by default, and maybe you could add another text box (under an "expert" switch, maybe) to add text in the latex preamble (so you can add packages or whatever). 
Thanks!

Technius:
It seems like everything is working, so I'm going to merge this PR sooner rather than later. I'm going to be refactoring the code in a later PR as the code is still too complicated for what it's doing.

@Rmano A custom preamble feature is definitely going to be useful; I'll open a issue for that soon.

@veecue There's already some text that indicates when the latex is invalid
image
but it might be a good idea to make the notice more prominent (or by making the formula input have a red background).

veecue:
Ok, I actually didn't see the hint.
I would still remove the x^2 default or at least preselect it so that it vanishes when you start typing.

Technius:
Preselection is definitely a good idea. I'll look into it.

Issue #962, opened 2019-03-06
When "saving as"  file.pdf.xopp  saves as file.xopp


enricoar:
When annotating a file.pdf  I generally try to save the corresponding xournal file as 
file.pdf.xopp   (which by the way was the default option in xournal).
However, although I enter file.pdf.xopp in the "save as" dialog, the file is saved as file.xopp

Besides this, which seems a bug to me, it would be useful that the default filename when annotating a file.pdf should be file.pdf.xopp, instead of file.xopp, so that when exporting the annotated file to pdf, the background does not get annotated

PR #1117, opened 2019-05-06
Allow "*.pdf.xopp" as a filename

jonasBoss:
Fixes https://github.com/xournalpp/xournalpp/issues/962
as well as it fixes an issue where the AutoloadPdfXoj feature would only look for *.pdf.xopp which however couldn't exist.

If it is preferred not to keep the .pdf part of the filename only line 2351 of Control.cpp should be changed, as this will fix the autoload issue.
This might be a good behavior once the background is saved in the file itself.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2348,14 +2348,14 @@ bool Control::loadPdf(Path filename, int scrollToPage)
 	{
 		Path f = filename;
 		f.clearExtensions();
-		f += ".pdf.xopp";
+		f += ".xopp";
 		Document* tmp = loadHandler.loadDocument(f.str());
 
 		if (tmp == NULL)
 		{
 			f = filename;
 			f.clearExtensions();
-			f += ".pdf.xoj";
+			f += ".xoj";
 			tmp = loadHandler.loadDocument(f.str());
 		}
 
--- a/src/util/Path.cpp
+++ b/src/util/Path.cpp
@@ -138,11 +138,11 @@ void Path::clearExtensions()
 {
 	string plower = StringUtils::toLowerCase(path);
 
-	REMOVE_EXTENSION(".pdf.xoj");
-	REMOVE_EXTENSION(".pdf.xopp");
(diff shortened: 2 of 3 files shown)


LittleHuba:
You will at least need to update the tests for the Travis build to be successful.
I'll review the changes the next days.

veecue:
@LittleHuba Is there a reason why this isn't merged yet?

LittleHuba:
Mainly as I wanted to take another look at it. These filename sanitizers are there for some very good reasons and changing them can have high risks like people by accident overriding important files.

Sorry that this takes so long, I'm busy with university so it may take some more time. But I have not forgotten about it.

veecue:
If this implementation gives you headaches, I have another proposal to solve the basic issue:
Automatically open $(basename).xopp instead of $(basename).pdf.xopp. Then the saving functionality wouldn't have to be changed.

LittleHuba:
>Automatically open $(basename).xopp instead of $(basename).pdf.xopp

This actually sounds better.

jonasBoss:
> Mainly as I wanted to take another look at it. These filename sanitizers are there for some very good reasons and changing them can have high risks like people by accident overriding important files.
> 
> Sorry that this takes so long, I'm busy with university so it may take some more time. But I have not forgotten about it.

From my testing I did not find any unexpected behavior. Currently xournalpp happily allows a file to be named $(basename).whatever.xopp just not $(basename).pdf.xopp . To block $(basename).pdf.xopp as a valid filename seems a bit arbitrary. I actually prefer $(basename).pdf.xopp as a naming convention, to indicate the dependency on $(basename).pdf . 

Once the new file format, which includes the PDF file, is finished the behavior could be changed so that the suggested Filename is $(pdf-basename).xopp , but still allow files to be named $(basename).pdf.xopp. In order to allow a user to choose so.

LittleHuba:
I did not find any problems. Thus I'll merge it and we look if anybody reports problems. We will reconsider this when introducing the new file format.

PR #1125, opened 2019-05-09
Disable pen cursor

chentangchu:
Pen cursor in a tablet might not be necessary, so disable it if blankPenCursor is set true in the setting xml file.


Technius:
Thanks for the PR! You have a merge conflict in `src/gui/Cursor.cpp`, so you'll need to fix that before we can start reviewing.

LittleHuba:
Before merging this I'd like to add a better representation in the settings. Drop-down menu and everything.

JJones780:
The highlight option should still produce an effect (in this case, leaving just the highlight).

For the big pen/none maybe switching to a  drop down box to select from competing pointer images.
We could add in some other options here too - crosshairs, system pointer, lefthanded arrow, etc.

LittleHuba:
@Technius will you implement the changes I mentioned or should somebody else take over the PR?

Technius:
@chentangchu Are you still willing to make a dropdown menu, like @LittleHuba requested? If not, we can let someone else handle it.

JJones780:
I'll look into incorporating this.

JJones780:
Sorry... haven't had time for Xournal++ lately... still planning to get to this... soon.

JJones780:
Upon further reflection , I believe this feature should be part of a larger feature which would allow the user to define a customized cursor (including a blank one ).  The single dot cursor ( which changes to a normal cursor when the mouse is moved ) should suffice for the time being.

PR #1148, opened 2019-05-14
fixed typo in build-flatpack.sh

alexruetz:
Just fixed the filename in build-flatpak.sh so it really matches the flatpak yml filename.


LittleHuba:
Thanks!

Issue #1140, opened 2019-05-12
Preselect formula input text in latex dialog


Technius:
**Is your feature request related to a problem? Please describe.**
The default formula in the latex dialog is not selected by default. Typing into the input requires deleting the default text, which is annoying.

**Describe the solution you'd like**
The default formula should be selected by default when the dialog opens.

**Describe alternatives you've considered**
None.

**Additional context**
See discussion in #1113.

This should be a simple change in `src/gui/dialog/LatexDialog.cpp`, so feel free to submit a PR if you want. I'll get around to doing this in a few days if no one else wants to.

PR #1149, opened 2019-05-14
Preselect default text in LaTeX input dialog

veecue:
With this change, users won't have to delete the default x^2 LaTeX
formula before adding their own. The x^2 is selected so upon typing it
will vanish.

Fixes: #1140

### DIFF
--- a/src/gui/dialog/LatexDialog.cpp
+++ b/src/gui/dialog/LatexDialog.cpp
@@ -24,10 +24,11 @@ LatexDialog::~LatexDialog()
 	XOJ_RELEASE_TYPE(LatexDialog);
 }
 
-void LatexDialog::setTex(string texString)
+void LatexDialog::setTex(string texString, bool preselect)
 {
 	XOJ_CHECK_TYPE(LatexDialog);
 	this->theLatex = texString;
+	this->preselect = preselect;
 }
 
 string LatexDialog::getTex()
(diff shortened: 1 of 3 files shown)


src/gui/dialog/LatexDialog.cpp
```diff
@@ -97,6 +98,14 @@ void LatexDialog::load()
 {
 	XOJ_CHECK_TYPE(LatexDialog);
 	gtk_text_buffer_set_text(this->textBuffer, this->theLatex.c_str(), -1);
+
+	// preselect all text in the box if desired
+	if (this->preselect) {
```
LittleHuba: This flag is unecessary as the method `load()` is only called by the method `show()` which is called on dialog creation.

src/gui/dialog/LatexDialog.cpp
```diff
@@ -97,6 +98,14 @@ void LatexDialog::load()
...
+		GtkTextIter start, end;
+		gtk_text_buffer_get_start_iter(this->textBuffer, &start);
+		gtk_text_buffer_get_end_iter(this->textBuffer, &end);
+		gtk_text_buffer_select_range(this->textBuffer, &start, &end);
```
LittleHuba: Everything except these four lines can be removed from the commit.

veecue: You're missing something there. We only want to select the default x^2 and don't preselect anything when an existing formula is edited

LittleHuba: Ahh good call I haven't thought about that 😅

Issue #1137, opened 2019-05-11
Selection tools: double click to edit


Technius:
**Is your feature request related to a problem? Please describe.**
When I use the select object tool, I often want to move things around and edit them without switching tools. Currently, you have to move them, deselect, switch tool, and then edit.

**Describe the solution you'd like**
Double clicking on selected objects such as Text and LaTeX objects when using a selection tool should cause their corresponding editors to appear.

**Describe alternatives you've considered**
Select object tool could alternatively bring up a context menu that has an `Edit` option when right clicking (current behavior is to copy the object).

**Additional context**
Probably involves changes to input handling.

LittleHuba:
For double clicks we need to catch the event. Currently we only catch single clicks. (but two single clicks are generated before a double click). 
This is a minor change and should not be much work at all.

PR #1159, opened 2019-05-15
Allow double clicks on selected objects to begin editing

Technius:
Implements #1137 for the new input system only.
- [x] Handle double clicks for both mouse and stylus
- [x] Only catch double taps/left clicks

### DIFF
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -223,6 +223,18 @@ double EditSelection::getYOnView()
 	return this->y;
 }
 
+double EditSelection::getOriginalXOnView()
+{
+	XOJ_CHECK_TYPE(EditSelection);
+	return this->contents->getOriginalX();
+}
+
+double EditSelection::getOriginalYOnView()
+{
+	XOJ_CHECK_TYPE(EditSelection);
+	return this->contents->getOriginalY();
+}
+
 /**
  * get the width in document coordinates (multiple with zoom)
  */
--- a/src/control/tools/EditSelection.h
+++ b/src/control/tools/EditSelection.h
@@ -66,6 +66,18 @@ public:
 	double getYOnView();
 
 	/**
+	 * @return The original X coordinates of the provided view in document
+	 * coordinates.
+	 */
+	double getOriginalXOnView();
+
+	/**
+	 * @return The original Y coordinates of the provided view in document
+	 * coordinates.
+	 */
+	double getOriginalYOnView();
+
+	/**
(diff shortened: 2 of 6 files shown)


LittleHuba:
You will also need to modify `StylusInputHandler` to relay double click events.

Technius:
Will merge the next day if there are no objections.

veecue:
It doesn't work for me... Is there anything to watch out for? New input system is enabled

LittleHuba:
@veecue Could you attach event logs? One for each the mouse and the pen please.

Technius:
@veecue I only implemented this feature for double clicks/taps on selected objects when you are using a selection tool, so make sure you have the text / latex object selected first.

veecue:
Here is the log. What I did
 - Open xournal++
 - Create a Text
 - Change the tool to rectangular selection
 - Select the created text
 - Doubleclick the selected text element
Expected: Cursor apperears for editing
Experienced: Nothing

[debug_input.zip](https://github.com/xournalpp/xournalpp/files/3204688/debug_input.zip)

Technius:
Looks like the selection detection code doesn't work with different zoom levels, so I'll need to fix it.

Technius:
@veecue Is the issue is fixed now?

veecue:
It works now. Thanks! One minor remark: When editing the text and then clicking somewhere else the Text tool keeps being selected. This is not what I would expect.

veecue:
Oooh and I want to take this opportunity to say thank you to everyone involved in this project thanks for developement, maintainance and great support. This software is my daily driver at Uni, so I would be willing to pay a little (Uni, as I said) for this work. Do you have a Donation link or anything like that?

LittleHuba:
Thanks for the flowers @veecue 😉 
The best help is actually another developer. So if you know your way around C++ or are willing to learn it we would be happy to welcome you to our team.

Sorry for the OT by the way @Technius.

veecue:
You're welcome.
I actually know C++ but not Gtk3 (which seems very interesting btw). I already discoverd some other bugs and feature requests and I will submit and possibly implement them whenever I have spare time.
Good thing about Github is that you can always contribute, but if I know the project better and want to take more responsibility, I will get back to you

Technius:
@veecue I'm happy that xournalpp is working for you :+1: 

> One minor remark: When editing the text and then clicking somewhere else the Text tool keeps being selected. This is not what I would expect.

I agree, but unfortunately there's no easy way to do this at the time. It's definitely an improvement that we should look into in the future, though.

I'll be merging this change in now, as it seems to work properly.

Issue #804, opened 2019-01-24
Quickmenu (right mouse button)


masac:
https://drive.google.com/file/d/1yg0_Q7CNWRIl7L4Svb5HgdDt8TgJnNID/view?usp=sharing
In version 0.0.5 of the MrWriter program https://github.com/unruhschuh/MrWriter/releases , was added the option Quickmenu (right mouse button). Can this option be implemented in Xournalpp? Thanks!

andreasb242:
The buttons in Xournal++ are configurable, so generally reserve the right button is no option.
The Toolbars are also customizeable, so for this a new config option is needed, and a new "Toolbar".

Possible for sure, but also not a high priority feature. Currently we get more feature requests / bugs than we can implement / fix.

masac:
Thanks!

JJones780:
This is done... see #1161 !
You can now assign `Floating Toolbox` in `Prefs->Mouse->Mouse Buttons`  or `Prefs->Stylus->Stylus Buttons`,
Also note the experimental feature in` Prefs->DrawingArea->Action on Tool Tap` which lets you call up the toolBox without a button (via quick tap - great for stylus - use button for other things). Please try it out.
Once assigned to a Stylus/Mouse button or Tool Tap, you can drag and drop tool items to the Floating Toolbox via the usual `View->Toolbars->Customize`

PR #1161, opened 2019-05-17
Floating toolbox

JJones780:
I tried out an idea on a whim and it worked fairly well so I developed it a bit further... 

Addresses #804 ( Quickmenu )
GIFS below...
FloatingToolbox_0 5
FloatingToolboxConfig_0 5


JJones780:
It's not perfect - not all tools items will cooperate. WIP ! 
Connect it to a middle or right mouse button etc.
It works with the experimental "TryClick" when the pen is selected ( tap on the screen will select an object or if none will bring up the floating toolbox ). Not polished yet.
If you View->Toolbars->Customize  the floatingToolbox will be shown and you can drag and drop to it as well.  This behaviour is fixed at the moment but should only happen if the user wants a floating toolbox.

LittleHuba:
It would be great if the items of the toolbar are copied to the floating toolbox and not moved.

JJones780:
I agree.. copied in some way would be nice... and filter what items can be moved to the floating toolbox.
I've already made some fixes to make the toolbox size better... hope to get some changes here in a few days.. busy weekend.

JJones780:
Experimental Stage - ready for some code review. This started out as a hack so I plan to have another look at it after a few days with fresh eyes and see what I've missed. If anyone else wants to review some code please do!
This is minimally viable with regard to features and does not restrict what can be put in the floating toolbox - some things won't behave well; just take them back out ;)

**If the user doesn't enable the FloatingToolbox in some way it won't show up in normal use** ( _and when it does it is clearly marked as experimental_ ).
When customizing toolbars, the FloatingToolbox are will only appear if the user has:
  - assigned a mouse or stylus button to bring up the floatingToolbox;
  - or enabled tapAction and Show FloatingToolbox ( prefs->DrawingArea->ActionOnToolTap );
  - or already put tools in the FloatingToolbox.

Because it is unobtrusive by default I suggest that we consider merging into master in order to get further feedback.

Technius:
I briefly skimmed the changes just now, but I'll have more time for a detailed review tomorrow. Going to nitpick the formatting and syntax here:
* There's a lot of inconsistent spacing that should be fixed (particularly around `if` statements)
* Also a lot of redundant, unnecessary blank lines
* Pointer asterisks should go next to the type name, not the variable name (e.g. `int* foo` instead of `int *foo`)

JJones780:
> I briefly skimmed the changes just now, but I'll have more time for a detailed review tomorrow. Going to nitpick the formatting and syntax here:

Oops...
I'll run the new files through a source formatter via a git rebase asap.  I knew there was something I had forgot to do ;)

JJones780:
Formatting cleaned up - via git rebase within existing commit tree. Please grab entire branch again if you are doing a `git fetch upstream pull/1161/head:NEWfloatingToolbox`equivalent.

src/gui/FloatingToolbox.cpp
```diff
@@ -0,0 +1,217 @@
...
+void FloatingToolbox::show(bool showTitle)
+{
+	XOJ_CHECK_TYPE(FloatingToolbox);
+
+	gtk_widget_hide(this->floatingToolbox);		//force showing in new position
+	gtk_widget_show_all(this->floatingToolbox);
+
+	if (showTitle)
+	{
+		gtk_widget_show(this->mainWindow->get("labelFloatingToolbox"));
```
Technius: Duplicated code, refactor this outside of the if statement.

src/gui/FloatingToolbox.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+class FloatingToolbox 
+{
+public:
+	FloatingToolbox(MainWindow* theMainWindow, GtkOverlay *overlay);
+	virtual ~FloatingToolbox();
+
+public:
+	void show(int x, int y);
```
Technius: Add documentation here that indicates the coordinate system of x and y.

src/gui/MainWindow.cpp
```diff
@@ -40,6 +40,12 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control)
 	this->toolbarWidgets = new GtkWidget*[TOOLBAR_DEFINITIONS_LEN];
 	this->toolbarSelectMenu = new MainWindowToolbarMenu(this);
 
+	
+	
+	GtkOverlay *overlay = GTK_OVERLAY (get("mainOverlay"));
+	this->floatingToolbox = new FloatingToolbox (this, overlay);
```
Technius: No need to make `floatingToolbox` a pointer if it's only going to be allocated in the constructor and deallocated in the destructor.

src/gui/MainWindow.h
```diff
@@ -103,6 +108,8 @@ class MainWindow : public GladeGui, public LayerCtrlListener
 	static void toggleMenuBar(MainWindow* win);
 
 	void createToolbarAndMenu();
+	
+	void initFloatingToolbox();
```
Technius: This isn't called anywhere.

src/gui/FloatingToolbox.cpp
JJones780: fixed

src/gui/FloatingToolbox.h
JJones780: fixed

src/gui/MainWindow.cpp
JJones780: I started that way... But wanted to avoid a separate initialization function.  Critical during testing but might only be a preference now. Leaving for now at least.

src/gui/MainWindow.h
JJones780: oops... leftover from early stages.
Good catch. Thanks for looking over the code!

JJones780:
You can now assign Floating Toolbox in ` Edit->Prefs->Mouse->Mouse Buttons` or ` Edit->Prefs->Stylus->Stylus Buttons`.
Also note the experimental feature in `Edit->Prefs->DrawingArea->Action on Tool Tap` which lets you call up the _Floating Toolbox_ without a button (via quick tap - great for stylus - use button for other things). Please try it out.

Once assigned to a Stylus/Mouse button or Tool Tap the Floating Toolbox will appear during Toolbar customization. You can drag and drop tool items to the Floating Toolbox via the usual `View->Toolbars->Customize`.
There is a new default toolbar configuration option: `View->Toolbars->Floating Toolbox (experimental)`.

_Not all tools work perfectly well from the toolbox. This may be improved over time - issues welcome._

masac:
Thank you!

PR #1179, opened 2019-05-27
Multiple clicks in text editor

### LINKED ISSUE #841: Enable textbox double, triple click, to select word, line
As per the title, I think it would be useful to introduce in the Text tool a functionality such as double clicking to select a full word, or triple clicking to select the whole line / whole content of the textbox.

Technius:
Makes multiple clicks on during text editors select words and lines. This will implement and close #841. In particular:

* [x] Double clicking in a text field will select that word
* [x] Fix a zoom issue when double clicking a selected object to begin editing
* [x] Triple click in a text field will select a paragraph

### DIFF
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -502,7 +502,7 @@ bool XojPageView::onButtonDoublePressEvent(const PositionInputData& pos)
 	else if (toolHandler->getToolType() == TOOL_TEXT)
 	{
 		this->startText(x, y);
-		this->textEditor->selectWord();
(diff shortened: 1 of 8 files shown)


Technius:
This PR should be ready to merge now. If there are no objections, I will be merging this PR in three days.

As a side note, I am not happy with how GTK doesn't provide ways to look ahead at text or jump to the beginning of a line, nor am I happy with GTK's notion of a "paragraph." I'd welcome any suggestions for improving the monstrosity that is the paragraph selection code.

Issue #1176, opened 2019-05-26
Automated build and release process


Technius:
# The problem

As the project grows in size and complexity, we need to have a more comprehensive and automated build system. We currently make manual releases, which is time consuming and error prone.

* Builds for all platforms should be automated. Currently, we only have Travis building on Linux.
* Every version compiled by whatever CI platform we're using should be uploaded as an "unstable" or "nightly" release. This will be very helpful to people who want to try bleeding edge builds but don't have the technical knowledge to compile their own version.
* Any tag pushed to the repo should draft a release on GitHub; the release should then be approved by a maintainer. This will make it far easier to do a release and standardize the release binaries.
* We need to not just build the executable, but also distro-specific packages such as `.deb` files and `.rpm` files. We should also build a generic AppImage for other distros. Related issues: #1013, #1177
* Once the build is automated, we need a "Release Guide" which describes what changes are required to the source code (such as updating version numbers) to create a new release.

# My personal opinion

Travis supports what we want right now, but it's not exactly in the best hands right now. We should switch to a CI platform such as CircleCI, AppVeyor, Azure Pipelines, or GitLab CI (I prefer GitLab CI). Bleeding edge builds can be hosted on GitHub, but it would be better to host them on the CI platform that we pick (since GitHub can only store one bleeding edge release at a time, while CI platforms generally host all built versions).

In any case, I have some experience setting up CI for projects, so I'm willing to make these proposed changes. I would appreciate any feedback, so let me know what you think!

# Progress

This meta-issue will track progress on the aforementioned goals.
* [x] Azure Pipelines setup #1244 
* [x] AppImage support #944 
* [x] `.deb` packages #1013, #959 
* [x] `.rpm` packages
* [x] Windows builds #1127 
* [x] Mac builds
* [x] flatpak #1177 
* [x] Create draft releases on tag push
* [ ] Artifactory/bintray setup with nightly releases
* [ ] Release guide

JJones780:
Looks like Travis and Azure Pipelines are the only ones offering windows builds?
I also came across a mention of [buildkite.com](https://buildkite.com) on [YCombinator](https://news.ycombinator.com/item?id=18978251);  it does do Windows builds but they cheat a bit... it does the builds on machines we supply ( running agents to communicate with their servers ).

LittleHuba:
There is not really a requirement to host all distributions on one CI platform. It is some overhead to work on multiple but if it enables us to use better platforms that makes it worth.

[...]

PR #1182, opened 2019-05-28
Add appimagecraft YML

### LINKED ISSUE #944: Linux Appimage
Packaging as an [Appimage](https://appimage.org/) under Linux would be useful.

TheAssassin:
Just download [appimagecraft](https://github.com/TheAssassin/appimagecraft) (either install via `pip` or use the provided AppImage) and run it like `appimagecraft build`. An AppImage should be produced and end up in the root of your local clone.

**Note:** appimagecraft is a very very new tool I wrote to eliminate the need for writing some bash scripts etc., it's all handled by appimagecraft, all you need to do is provide some information in the form of a small YAML file.
It's unlike but if it should not work for you, your feedback is appreciated, please feel free to open issues in the tracker.

Fixes #944.
Contributes to #1176.

### DIFF
--- a/appimagecraft.yml
+++ b/appimagecraft.yml
@@ -1,0 +1,11 @@
+version: 1
+
+project:
+  name: com.github.xournalpp.xournalpp
+  version_command: git describe --tags
+
+build:
+  cmake:
+
+appimage:
+  linuxdeploy:


probonopd:
Interesting!

TheAssassin:
Just rebase-pulled master on my branch, AppImage now works fine since merge of #1202. Tested in a live ISO VM for xubuntu 18.10.

Technius:
I'll merge this for now since we don't have any kind of configuration for AppImage builds yet, but we'll need to revisit the configuration once we get some of the other infrastructure set up.

probonopd:
Thanks. Where can the resulting AppImages be downloaded?

appimagecraft.yml
```diff
@@ -0,0 +1,11 @@
+version: 1
+
+project:
+  name: com.github.xournalpp.xournalpp
+  version_command: git describe --tags
```
TheAssassin: @Technius I should've mentioned that this command to detect the current version was just a random guess I made, doesn't seem to work all that well since I just tried and the app said its version is 1.0.11, whereas this command gave me 1.0.8-...
This should be fixed so that both the version noted in the AppImage and the one in the app's about dialog are the same.

Technius:
@probonopd We haven't set up any deployment infrastructure yet, so binaries and packages need to be compiled and assembled manually for now.

appimagecraft.yml
Technius: Looks like we only tagged the last two releases on separate branches, so the tags on master didn't get updated. Hopefully this should be fixed by the next release.

Easy workaround for now would be to set up cmake to write the version string to a file during builds and then read that file to get the version.

PR #1197, opened 2019-06-03
Fix libzip dependency issue 

### LINKED ISSUE #1180: libzip dependency configured incorrectly in CMakeLists
Compiling on linux has problem with libzip library. `pkgconfig` finds the right version of libzip but it try to links with only library without path of libzip. Therefore trying to find libzip in linking step fails. Here is the fix:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 856b116..77b045c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ add_includes_ldflags ("${ZLIB_LIBRARIES}" "${ZLIB_INCLUDE_DIRS}")
 
 # libzip
 pkg_check_modules (ZIP REQUIRED "libzip >= 1.0.1")
-add_includes_ldflags ("${ZIP_LIBRARIES}" "${ZIP_INCLUDE_DIRS}")
+add_includes_ldflags ("${ZIP_LDFLAGS}" "${ZIP_INCLUDE_DIRS}")
 
 # pthreads
 find_package (Threads REQUIRED)
```

Technius:
Patch thanks to @moghiseh, closes #1180

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ add_includes_ldflags ("${ZLIB_LIBRARIES}" "${ZLIB_INCLUDE_DIRS}")
 
 # libzip
 pkg_check_modules (ZIP REQUIRED "libzip >= 1.0.1")
-add_includes_ldflags ("${ZIP_LIBRARIES}" "${ZIP_INCLUDE_DIRS}")
+add_includes_ldflags ("${ZIP_LDFLAGS}" "${ZIP_INCLUDE_DIRS}")
 
 # pthreads
 find_package (Threads REQUIRED)


LittleHuba:
This breaks builds on Arch.

LittleHuba:
This should fix the issue:
```
# zlib
find_package (ZLIB REQUIRED)
add_includes_ldflags ("${ZLIB_LDFLAGS}" "")
add_includes_ldflags ("${ZLIB_LIBRARIES}" "${ZLIB_INCLUDE_DIRS}")

# libzip
pkg_check_modules (ZIP REQUIRED "libzip >= 1.0.1")
add_includes_ldflags ("${ZIP_LDFLAGS}" "")
add_includes_ldflags ("${ZIP_LIBRARIES}" "${ZIP_INCLUDE_DIRS}")
```

It just adds all the flags which should statisfy all systems. But it has to be tested on the other machine since it could lead to problems. If that is the case we need to include the flags dependent on the OS.

Technius:
How does the build break? `ZIP_LDFLAGS` comes straight from `pkg-config --libs libzip` and should work.

Adding both `LDFLAGS` and `LIBRARIES` is going to duplicate the library link flags.

LittleHuba:
On Arch `ZLIB_LDFLAGS` is empty for some reason.

Technius:
For zlib, we're using `find_package` which just calls [`FindZLIB`](https://cmake.org/cmake/help/v3.6/module/FindZLIB.html), which only defines `ZLIB_LIBRARIES`. So just changing the flags for `libzip` should be sufficient.

Issue #1132, opened 2019-05-11
Maximizing the window causes scrolling issues


Technius:
**Affects versions :**
 - OS: Linux (Solus Linux), probably affects more versions
 - Version of Xournal++: `master` branch as of ` a5f033e`, as well as v1.0.10

**Describe the bug**
After maximizing the window for the first time, scrolling can break in one of two ways, depending on the situation:
1. Scrolling is completely disabled on the blank document displayed immediately after Xournal++ is opened. Sometimes the same issue occurs when a document with one page is opened
2. Scrolling within the page shown works, but scrolling between pages is disabled. In addition, selecting a page will only scroll within the current page but not change which page is displayed.

**To Reproduce**
Steps to reproduce situation 1:
1. Open Xournal++. There should be a blank document.
2. Maximize the window.
3. Attempt to scroll with the mouse wheel.

Steps to reproduce situation 2:
1. Open Xournal++. Open any document with more than one page.
2. Maximize the window.
3. Attempt to scroll to the next/previous page with the mouse wheel.
4. Attempt to select a different page.

**Expected behavior**
Scrolling should work.

**Screenshots of Problem**
N/A

**Additional context**
N/A

JJones780:
I can replicate this bug in the old input system.
It is not present in the new input system.

Technius:
I can still reproduce situation 1 with the new input system enabled.

PR #1213, opened 2019-06-04
 Fixed #1139, #1132

### LINKED ISSUE #1139: Resizing xournalpp does not update horizontal scrolling range
**Affects versions :**
 - OS: Ubuntu 18.04
 - Version of Xournal++: 1.0.11

**Describe the bug**
After xournalpp is resized, the horizontal scrolling range is not re-calculated until the drawing area is resized. In other words, the non-drawing region is not trimmed. As a result, it is possible to horizontally scroll out way beyond the drawing area.

**To Reproduce**
1. Maximize xournalpp
2. Un-maximize xournalpp so that it resizes horizontally
3. Drag the horizontal scrollbar to the right

**Expected behavior**
Horizontal scrolling should stop immediately after encountering the right border of the drawing area.

**Additional context**
Hitting Ctrl+(-) or Ctrl+Shift+(+) immediately fixes the issue.

Febbe:
Fixed #1139, #1132 by changing the widget which triggers the resize.
Fixed an untracked bug, that layoutPages() got the old allocation from the previous run, instead of the desired size allocation.

Refactored most of the files i touched to fix those issues.
Both were related to each other, so i could not split them into two commits.

Added "const" to the Range class.
Added a "noop" to the defines in XournalType, so ClangTidy wont show a warning because of an empty statement


src/control/zoom/ZoomControl.cpp
```diff
@@ -572,11 +579,17 @@ bool ZoomControl::onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScr
...
 }
 
-bool ZoomControl::onWidgetSizeChangedEvent(GtkWidget* widget, GdkRectangle *allocation, ZoomControl* zoom)
+bool ZoomControl::onWidgetSizeChangedEvent(GtkWidget* widget, GdkRectangle* allocation, ZoomControl* zoom)
 {
 	XOJ_CHECK_TYPE_OBJ(zoom, ZoomControl);
 
+	Rectangle r(allocation->x, allocation->y, allocation->width, allocation->height);
+	auto layout = gtk_xournal_get_layout(zoom->view->getWidget());
+	assert(widget != zoom->view->getWidget());
```
LittleHuba: Use `g_assert_true()` instead. We use it everywhere else so we shouldn't mix things.

src/gui/Layout.cpp
```diff
@@ -295,103 +276,85 @@ void Layout::layoutPages()
...
 	{
-		totalWidth += this->sizeCol[c] + XOURNAL_PADDING_BETWEEN;
-		this->sizeCol[c] = totalWidth;	//accumulated - absolute pixel location for use by getViewAt() and updateVisibility()
+		totalWidth += this->widthCols[c] + XOURNAL_PADDING_BETWEEN;
+		this->widthCols[c] = totalWidth;    //accumulated - absolute pixel location for use by getViewAt() and updateVisibility()
 	}
-	totalWidth += borderX - XOURNAL_PADDING_BETWEEN;
 
-	
+	//This seems to be wrong:
```
LittleHuba: Is this fixed now or is this a TODO?

src/gui/Layout.cpp
```diff
@@ -412,59 +375,59 @@ void Layout::ensureRectIsVisible(int x, int y, int width, int height)
...
-// 	int cb = cit - this->sizeCol.begin();
+// 	auto rit = std::lower_bound( this->heightRows.begin(),  this->heightRows.end(), y);
+// 	int rb = rit - this->heightRows.begin();	//get index
+// 	auto cit = std::lower_bound( this->widthCols.begin(),  this->widthCols.end(), x);
+// 	int cb = cit - this->widthCols.begin();
+
 
 
+	auto numRows = this->heightRows.size();//this->mapper.getRows();
+	auto numCols = this->widthCols.size();//this->mapper.getColumns();
```
LittleHuba: Remove the comments if no longer required.

src/gui/Layout.cpp
```diff
@@ -475,36 +438,46 @@ XojPageView* Layout::getViewAt(int x, int y)
...
+
+double Layout::getMinimalHeight()
+{
+	return this->minHeight;
+}
+
+double Layout::getMinimalWidth()
+{
+	return this->minWidth;
+}
```
LittleHuba: Add a newline here :wink:

src/gui/Layout.h
```diff
@@ -107,45 +113,40 @@ class Layout
...
+	XournalView* view{};
+	ScrollHandling* scrollHandling{};
+
+	std::vector<int> widthCols;
+	std::vector<int> heightRows;
 
-	double lastScrollHorizontal = -1;
-	double lastScrollVertical = -1;
+	double lastScrollHorizontal{-1};
+	double lastScrollVertical{-1};
```
LittleHuba: Do not use this initialization routine as it may not be supported by every compiler. 
Stick with `type variable = value;`

src/gui/LayoutMapper.cpp
```diff
@@ -19,52 +12,44 @@ LayoutMapper::~LayoutMapper()
...
 
+	};
 	//assemble bitflags for LayoutType
-	int type = isVertical ? LayoutBitFlags::Vertically : 0;
-	type |= isRightToLeft ? LayoutBitFlags::RightToLeft : 0;
-	type |= isBottomToTop ? LayoutBitFlags::BottomToTop : 0;
+	auto type = settings->isPresentationMode() ? Vertical :
+	            isVertical ? Vertical : 0u |
+	                                    isRightToLeft ? RightToLeft : 0u |
+	                                                                  isBottomToTop ? BottomToTop : 0u;
```
LittleHuba: This is not really nice as it takes some effort to understand. The previous way was better.

src/gui/Layout.cpp
Febbe: Oh right, this is a Todo. This code snipped is rather wrong but it definitely lacks of understanding. We may want to extract it / reformat it. It also is the reason, that we musst recalculate the layout every time "layoutPages" is called. It invalidates the precalculation.
I had to introduce a new variable to catch this.

```suggestion
```

src/gui/Layout.cpp
Febbe:
```suggestion
	auto numRows = this->heightRows.size();
	auto numCols = this->widthCols.size();
```

src/gui/Layout.h
Febbe: Mhh strange this is a c++11 feature and all compilers which stick to the standard should support it. It also is (in my opinion) cleaner to read and to write.

src/gui/LayoutMapper.cpp
Febbe: right its also wrong there are braces missing, i fixed it. It should be way better now. Thanks for the review.

src/gui/LayoutMapper.cpp
```diff
@@ -19,52 +12,44 @@ LayoutMapper::~LayoutMapper()
 }
 
 
-void LayoutMapper::configureFromSettings(int numPages, Settings* settings)
+void LayoutMapper::configureFromSettings(size_t numPages, Settings* settings)
 {
 	XOJ_CHECK_TYPE(LayoutMapper);
 
+	auto isPairedPages = settings->isShowPairedPages();
```
Technius: The variables here have different types, and the function names are not clear enough to accurately guess the types. The use of `auto` here is hurting readability; I would keep some of the explicit type annotations here.

src/gui/LayoutMapper.cpp
```diff
@@ -146,48 +131,64 @@ int LayoutMapper::getPairedPages()
...
+}
+
+bool LayoutMapper::isBottomToTop() const
+{
+	return (layoutType & BottomToTop) == BottomToTop;
+}
+
+bool LayoutMapper::isRightToLeft() const
+{
+	return (this->layoutType & RightToLeft) == RightToLeft;
```
Technius: Nitpicking, but please be consistent with `this->layoutType` or `layoutType` here.

src/gui/LayoutMapper.h
```diff
@@ -56,13 +59,14 @@ class LayoutMapper
...
 	 */
 
-	
-	void configureFromSettings(int numPages, Settings* settings);
-	
-	
+
+	void configureFromSettings(size_t numPages, Settings* settings);
+
+	[[deprecated ("use LayoutMapper::configureFromSettings")]]
```
Technius: Since this is internal code, you can just remove the method if it doesn't break anything.

src/gui/Layout.h
```diff
@@ -39,6 +39,7 @@ class Layout
 	/**
 	 * Adjusts the layout size to the given values
 	 */
+	// This should never happen from the outside, since it must recalculated for every change
```
Technius: The method should be marked private, if this is the case.

src/gui/LayoutMapper.cpp
Febbe: The variables have the same type, because I changed the type of the corresponding member variable to bool, since it was only used as bool and it is bad practice to check if an int is unequal to 0 just by wrighting the variable into the if statement. And a variable starting with or containing a verb is normally a bool. I changed the variable name to "showPairPages". This should be more readable.

src/gui/LayoutMapper.cpp
Febbe: Since we don't mark member variables with a "_" postfix it is better to use "this->variable". I don't think this is nitpicking ;) .

src/gui/Layout.h
Febbe: i removed it, because it only was a forward to layoutPages(w,h);

src/gui/Layout.h
LittleHuba: Well compilers unfortunately are not always compliant. Anyway we adhere to the mentioned standard in the rest of the code so we should not mix these code styles.

src/gui/LayoutMapper.cpp
Febbe: Ok i misunderstood you, i changed it now, is this better?
I still think it's not important for readability, whether a number is an int, unsigned, long or size_t.
Most of the time, it does not even matter if it's a float or a double. That what matters is the purpose, and for what the variable is used. For this pourpose an "auto const(&)" is fully sufficient. It also has the big advantage, that you can exchange the type without having troubles to change the type everywhere it is used or copied.

src/gui/Layout.cpp
```diff
@@ -137,153 +140,131 @@ Rectangle Layout::getVisibleRect()
...
 {
 	XOJ_CHECK_TYPE(Layout);
 
-	return layoutWidth;
-}
+	auto* settings = view->getControl()->getSettings();
+	auto len = view->viewPagesLen;
+	mapper.configureFromSettings(len, settings);
+	auto colCount = mapper.getColumns();
+	auto rowCount = mapper.getRows();
```
LittleHuba: Use auto only with pointers on initialization. Otherwise we have a hard time to:
a) check your code
b) maintaining said code as finding out which type a variable is will be a tedious task.

src/gui/Layout.h
```diff
@@ -107,45 +107,40 @@ class Layout
...
+	int minWidth = 0;
+	int minHeight = 0;
+
 	/**
 	 * cache the last GetViewAt() row and column.
 	 */
 	int lastGetViewAtRow = 0;
 	int lastGetViewAtCol = 0;
+
+	bool valid = false;
```
LittleHuba: Could you add a comment what this does?

src/gui/Layout.cpp
```diff
@@ -137,153 +140,131 @@ Rectangle Layout::getVisibleRect()
 	                 gtk_adjustment_get_page_size(scrollHandling->getVertical()));
 }
 
-/**
- * Returns the height of the entire Layout
- */
-double Layout::getLayoutHeight()
+int addif(int lhs, int rhs, bool b)
```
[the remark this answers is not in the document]
Febbe: Done, but this code snipped is so small, that it should be understandable just by reading the code.
It also is a source-file function so no one can access it from outside the source file.

src/gui/Layout.h
Febbe: Done, but i would rather remove it completely. (Part of a big refactoring / completely unrelated to zoom). This is just a workaround.

LittleHuba: If you can remove it without influencing the behaviour, just go for it. It is the main goal to decrease complexity of the code after all.

LittleHuba:
Seems like you tried to use a functionality only provided by newer versions of boost:
```
In file included from /home/travis/build/xournalpp/xournalpp/src/gui/Layout.h:19:0,
                 from /home/travis/build/xournalpp/xournalpp/src/gui/PageView.h:15,
                 from /home/travis/build/xournalpp/xournalpp/src/control/jobs/XournalScheduler.h:18,
                 from /home/travis/build/xournalpp/xournalpp/src/control/Control.h:25,
                 from /home/travis/build/xournalpp/xournalpp/src/util/audio/AudioPlayer.cpp:1:
/home/travis/build/xournalpp/xournalpp/src/gui/LayoutMapper.h:17:30: fatal error: boost/optional.hpp: No such file or directory
```

src/gui/Layout.h
Febbe: Unfortunately i would have to rewrite Layout::getViewAt and Layout::updateVisibility, that was a too big change for me, i also have other things to do in my private time. :D. But I added some todos, to indicate, that it could/should be refactored away.

LittleHuba: Ahh if it is marked that is already good. It can still be done in a separate PR. If you might have the time later on you can still try your hand. The work you provide is very good so far.

src/gui/Layout.cpp
```diff
@@ -434,27 +417,24 @@ XojPageView* Layout::getViewAt(int x, int y)
...
 		}
-
 	}
 
-	return NULL;
+	return nullptr;
 }
 
-int Layout::getIndexAtGridMap(int row, int col)
+boost::optional<size_t> Layout::getIndexAtGridMap(size_t row, size_t col)
```
LittleHuba: This is why the build fails. You will have to exchange this with something else...

Febbe:
> Seems like you tried to use a functionality only provided by newer versions of boost:
> 
> ```
> In file included from /home/travis/build/xournalpp/xournalpp/src/gui/Layout.h:19:0,
>                  from /home/travis/build/xournalpp/xournalpp/src/gui/PageView.h:15,
>                  from /home/travis/build/xournalpp/xournalpp/src/control/jobs/XournalScheduler.h:18,
>                  from /home/travis/build/xournalpp/xournalpp/src/control/Control.h:25,
>                  from /home/travis/build/xournalpp/xournalpp/src/util/audio/AudioPlayer.cpp:1:
> /home/travis/build/xournalpp/xournalpp/src/gui/LayoutMapper.h:17:30: fatal error: boost/optional.hpp: No such file or directory
> ```

Yes I saw this last night, but was just to tired to revert it. I would love to use the equivalent c++17 feature but this would certainly break old compilers. I also saw, that we don't even use boost anywhere.
I will just remove it and leave a todo, if we decide to use c++17 or a newer boost version.

LittleHuba:
We would also like to use the newer syntax and library versions of all our dependencies. The problem is that some distributions we ship to, just have really old versions of those. Hence we are forced to stick with the old ways.

> Yes I saw this last night, but was just to tired to revert it.

No worries just wanted to point it out to you :wink:

src/gui/Layout.cpp
Technius: Easy way out is `std::unique_ptr`. But this is temporary; using pointers for optional values is really terrible.

Febbe: replaced it with my own impl of an optional<size_t>.

> Easy way out is std::unique_ptr. But this is temporary; using pointers for optional values is really terrible.

I saw cases where the llvm/clang compiler optimized scoped std::unique_ptr completely away :D

src/gui/Layout.cpp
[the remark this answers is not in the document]
Febbe: Its used in two functions, i will refactor it to an inline function.
I thought that this function is so trivial, that every compiler will inline it automatically, but godbolt.org convinced me otherwise. The inline keyword works gcc/clang/msvc and the Intel compiler inlining the function.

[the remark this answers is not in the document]
Technius: It's inlined when optimizations are turned on (e.g. with `-O1`).

LittleHuba:
Your code looks quite good so far. My only problem is your abusive use of autotyping 😉 
I added a section to the Coding Conventions describing why I do have such a big problem with devs using a lot of auto types. If you ever had a bug caused by `auto`, you'll know why I'm that strict with it.

https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#using-auto-everywhere

Febbe:
> Your code looks quite good so far. My only problem is your abusive use of autotyping wink
> I added a section to the Coding Conventions describing why I do have such a big problem with devs using a lot of auto types. If you ever had a bug caused by `auto`, you'll know why I'm that strict with it.
> 
> https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#using-auto-everywhere

Ironically i've learned it the other way round. I got hits on my fingers at university, every time I did not used "auto const" or "auto const&" because its hard to refactor, to maintain, it can be less efficient, or even completely wrong, because you have some implicit type casts, at places where you forget to change the types, such as a double gets implicitly get cast to an int. That problem would not appear with auto. And I actually agree with them now.

I have a good article from Herb Sutter for you: https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/ there is also a section about readability.

LittleHuba:
Interesting as I learned to everytime be aware of which types you use and taking the usages into account when changing function signatures. That might be caused by my origin from developing in scientific computing where it is very important to actually be aware of what your code uses.

I've requested the opinion of the other devs and we will see for which direction we decide. Anyway thanks for the work and your point of view. I'll get back to you as soon as I have further information.

Technius:
Because there are outstanding todos, I would prefer if this PR is merged into a separate branch. Then we can address the remaining tasks in that branch before creating a new PR to merge back into master.

Febbe:
Yes we could open a e.q. "Layout and resize refactoring" branch, but the patch I preserved here works very well, there are still smal non intended side effects, from calling the resize inside the size_allocate routine of the scrolbarwidget, instead of using the appropriate signal from GtkWindow. But they aren't very problematic.  And I wont have time for the next 6 weeks to fix all the Todo's because I have a deadline. I would rather work at small problems instead of the big ones.

LittleHuba:
The general consent on auto typing is to keep it to a minimum. It has too many pitfalls to be considered a good coding style for now. The main concern is that we will have no chance to check PRs by hand anymore. That might not be a problem for your PR but yours is also of good quality what unfortunately cannot be said about all PRs.

We will allow auto typing for instatiations of new objects and iterators to allow better readability of code. All other variables should have the correct type.
Additionally we will switch to smart pointers to fix the potential memory leaks.

Sorry for the inconvenience.

LittleHuba:
I'll shortly create the branch, switch your target to the new branch and do the merge. 

@Febbe You may either decide to do a new PR on that branch, fixing things you want to fix or I can give you access to the branch to directly work on it.

Febbe:
> The general consent on auto typing is to keep it to a minimum.
Thats ok for me, there is also a way to use "using name = ugly_and_long_type". 

> I'll shortly create the branch, switch your target to the new branch and do the merge.
Ok perfect, then i can make a new PR with the simple memory leak fix.

> You may either decide to do a new PR on that branch, fixing things you want to fix or I can give you access to the branch to directly work on it.
I like it to just do a PR for greater changes, so others can review it. For small changes it could be good to just push directly to the new branch :) .

PR #1214, opened 2019-06-05
Make latex tool check for latex package dependencies

### LINKED ISSUE #1178: Latex tool should check for required latex packages
**Affects versions :**
 - OS: all
 - Version of Xournal++: `1.0.12`

**Describe the bug**
The LaTeX tool does not check for the required package used in the template (`standalone`). As a result, the tool will not function properly (see #1100) when required packages are missing. No relevant warning is presented to the user.

**To Reproduce**
1. Have LaTeX installed without the `standalone` package on the system.
2. Try to use the LaTeX tool.
3. No preview occurs.

**Expected behavior**
A dialog should pop up, warning the user that the `standalone` package is not installed and that additional system dependencies may be required (e.g. `texlive-latex-extra` on Debian derivatives).

The user guide and installation instructions should describe what dependencies are needed.

**Screenshots of Problem**
N/A

**Additional context**
Packages can be checked for existence using `kpsewhich`. For example, `kpsewhich standalone` check if the `standalone` LaTeX package is available on the system.

Technius:
Fixes #1178. In particular:

* Now requires `kpsewhich` to be installed (provided by most latex distributions)
* Uses `kpsewhich` to check for `standalone`.

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -61,20 +61,40 @@ LatexController::~LatexController()
 /**
  * Find the tex executable, return false if not found
  */
-bool LatexController::findTexExecutable()
+LatexController::FindDependencyStatus LatexController::findTexDependencies()
 {
 	XOJ_CHECK_TYPE(LatexController);
 
 	gchar* pdflatex = g_find_program_in_path("pdflatex");
 	if (!pdflatex)
 	{
-		return false;
+		string msg = _("Could not find pdflatex in PATH.\nPlease install pdflatex first and make sure it's in the PATH.");
+		return LatexController::FindDependencyStatus(false, msg);
 	}
-
 	this->pdflatexPath = pdflatex;
 	g_free(pdflatex);
 
-	return true;
+	// Check for 'standalone' latex package
+	static gchar* kpsewhichArgs[] = { g_strdup("kpsewhich"), g_strdup("standalone") };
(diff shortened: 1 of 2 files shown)


LittleHuba:
Please add this to the install instructions in the Wiki.
@andreasb242 do we need to specify the new dependency in some package files?

Technius:
There shouldn't be any new dependency; I believe that `kpsewhich` is already bundled with the basic `texlive-*` packages that we already depend on.

LittleHuba:
> There shouldn't be any new dependency; I believe that `kpsewhich` is already bundled with the basic `texlive-*` packages that we already depend on.

We should check this out.

Technius:
* For Debian-based systems, we depend on `texlive` which depends on `texlive-binaries`, which does bundle `kpsewhich`: https://packages.debian.org/stretch/amd64/texlive-binaries/filelist
* For Arch, we depend on `texlive-bin` which bundles `kpsewhich`: https://www.archlinux.org/packages/extra/x86_64/texlive-bin/
* For Fedora, we depend on `texlive-scheme-basic` which bundles `texlive-collection-basic` -> `texlive-kpathsea` which bundles `kpsewhich`: https://rpmfind.net/linux/RPM/mandriva/devel/cooker/x86_64/media/main/release/texlive-scheme-basic-20120810-1.noarch.html
* For MacOS, `kpsewhich` is included with `mactex`.
* For Windows, there's MiKTeX and texlive, both of which bundle `kpsewhich`.

LittleHuba:
Seems like all major distributions are covered. LGTM

PR #1222, opened 2019-06-09
Cursor cleanup - part 1

JJones780:
Recent cursor changes revealed growing pains related to the new input system.
This keeps the new cursor behaviour ( but consolidates the responsibility in XournalpppCursor ) and re-enables functionality that hadn't yet carried over to the new input system.
It also retains new and old functionality in the old input system.

### DIFF
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -7,6 +7,15 @@
 #include <pixbuf-utils.h>
 #include <cmath>
 
+
+/* Work In Progress:
+ * NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE
+ * 
+ * NOTE     Please consult with JJones780 before making any changes to this file or its header. 
+ * 
+ * NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE
+ */
+
 XournalppCursor::XournalppCursor(Control* control)
  : control(control)
 {
@@ -16,6 +25,14 @@ XournalppCursor::XournalppCursor(Control* control)
 XournalppCursor::~XournalppCursor()
 {
 	XOJ_RELEASE_TYPE(XournalppCursor);
+}
+
+
+void XournalppCursor::setInputDeviceClass(InputDeviceClass device)
+{
+	XOJ_CHECK_TYPE(XournalppCursor);
+
+	this->inputDevice = device;
 }
 
 void XournalppCursor::setMouseDown(bool mouseDown)
@@ -55,10 +72,7 @@ void XournalppCursor::setCursorBusy(bool busy)
 	XOJ_CHECK_TYPE(XournalppCursor);
 
 	MainWindow* win = control->getWindow();
-	if (!win)
-	{
-		return;
-	}
+	if (!win) return;
 
 	if (this->busy == busy)
 	{
@@ -118,22 +132,17 @@ void XournalppCursor::updateCursor()
 	XOJ_CHECK_TYPE(XournalppCursor);
 
 	MainWindow* win = control->getWindow();
-	if (!win)
-	{
-		return;
-	}
+	if (!win) return;
 
 	XournalView* xournal = win->getXournal();
-	if (!xournal)
-	{
-		return;
-	}
+	if (!xournal) return;
 
 	GdkWindow* window = gtk_widget_get_window(win->getWindow());
 
 	GdkCursor* cursor = NULL;
 	bool useDefault = false;
-
+	
+	
 	if (this->busy)
 	{
 		setStockCursor(GDK_WATCH);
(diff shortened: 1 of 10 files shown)


src/gui/XournalppCursor.cpp
```diff
@@ -118,119 +132,127 @@ void XournalppCursor::updateCursor()
...
 
 	GdkCursor* cursor = NULL;
-
+	bool useDefault = false;
+	
+	
 	if (this->busy)
 	{
-		cursor = gdk_cursor_new_for_display(gdk_window_get_display(window), GDK_WATCH);
+		setStockCursor(GDK_WATCH);
```
Technius: The combination of `setStockCursor` and `cursor = ...` is error prone. I suggest refactoring this so that only one of `setStockCursor` or `cursor = ...` is used.

src/gui/XournalppCursor.h
```diff
@@ -32,8 +33,9 @@ class XournalppCursor
 	void setMouseDown(bool mouseDown);
 	void setInvisible(bool invisible);
 	void setInsidePage(bool insidePage);
-	void setTempCursor(GdkCursorType type);
-	void setTempDrawDirCursor(bool shift, bool ctrl);
+	void setStockCursor(GdkCursorType type);
```
Technius: Document what `setStockCursor` does, the purpose is unclear right now.

src/gui/XournalppCursor.h
```diff
@@ -57,4 +63,14 @@ class XournalppCursor
...
 	bool invisible = false;
+	
+	// One shot drawDir custom cursor -drawn instead of pen/stylus then cleared.
+	bool drawDirActive = false;
+	bool drawDirShift = false;
+	bool drawDirCtrl = false;	
+	
+	
+	//combination to avoid making same cursor
+	void* lastCustomCursorAddress = NULL;	//for comparison only
```
Technius: This doesn't seem to be used anywhere.

src/gui/XournalppCursor.cpp
```diff
@@ -118,119 +132,127 @@ void XournalppCursor::updateCursor()
...
+			{
+				setStockCursor(GDK_ARROW);
+			}
+			else
+			{
+				if (type == TOOL_PEN)
+				{
+					cursor = getPenCursor();
+				}
+				else // must be:  if (type == TOOL_HILIGHTER)
```
Technius: Either add an assertion here or keep the highlighter tool check; this assumption may not be true in the future.

src/gui/inputdevices/MouseInputHandler.cpp
```diff
@@ -54,19 +54,12 @@ bool MouseInputHandler::handleImpl(InputEvent* event)
 	 * Trigger motion actions
 	 */
 	// Trigger motion action when pen/mouse is pressed and moved
-	if (event->type == MOTION_EVENT) //mouse or pen moved
+	if ( event->type == MOTION_EVENT) //mouse or pen moved
```
Technius: Unnecessary whitespace.

src/gui/XournalppCursor.cpp
```diff
@@ -254,8 +275,10 @@ void XournalppCursor::updateCursor()
 		{
 			gdk_window_set_cursor(gtk_widget_get_window(xournal->getWidget()), cursor);
 		}
-
-		gtk_widget_set_sensitive(xournal->getWidget(), !this->busy);
+		else if(useDefault)
```
Technius: `useDefault` should be renamed, since it does not "use a default cursor" but rather disables inputs.

src/gui/XournalppCursor.cpp
```diff
@@ -118,119 +132,127 @@ void XournalppCursor::updateCursor()
...
 			case CURSOR_SELECTION_BOTTOM:
-				cursor = gdk_cursor_new_for_display(gdk_window_get_display(window), GDK_SB_V_DOUBLE_ARROW);
+				setStockCursor(GDK_SB_V_DOUBLE_ARROW);
 				break;
 			default:
 				break;
 			}
-		}
-		else if (type == TOOL_PEN)
+		}
```
Technius: Trailing whitespace.

src/gui/XournalppCursor.cpp
```diff
@@ -302,16 +325,32 @@ GdkCursor* XournalppCursor::getEraserCursor()
 GdkCursor* XournalppCursor::getHighlighterCursor()
 {
 	XOJ_CHECK_TYPE(XournalppCursor);
+	
+	if (this->drawDirActive)
+	{
+		this->drawDirActive = false;
```
Technius: This is error prone because it modifies the state: updating the cursor twice a row will not create the correct cursor.

src/gui/XournalppCursor.h
```diff
@@ -57,4 +63,14 @@ class XournalppCursor
 
 	bool mouseDown = false;
 	bool invisible = false;
+	
+	// One shot drawDir custom cursor -drawn instead of pen/stylus then cleared.
+	bool drawDirActive = false;
+	bool drawDirShift = false;
+	bool drawDirCtrl = false;
```
Technius: Please leave comments explaining how the three are related.

JJones780:
@Technius,  Thanks for having a look. Sorry I didn't put a comment in to mention I was rewriting the XournalppCursor class next.

PR #1228, opened 2019-06-10
Place newly created latex objects in the center of the screen/page

Technius:
By default, newly created latex objects are placed in the top left corner of the current page. This PR now sets convenient default locations:

* If the center of the viewport is on the current page, then place the new latex in the center of the viewport.
* Otherwise, place the new latex in the center of the current page (which may possibly be out of the viewport).

The default placement can obviously be improved, so feedback is welcome.

(There's also the option of turning the LaTeX tool into a real selectable tool, but that would require some work.)

I will be merging this in three days if there are no objections.

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -196,6 +196,30 @@ void LatexController::findSelectedTexElement()
 			this->imgheight = this->selectedText->getElementHeight();
 		}
 	}
+	else
+	{
+		// This is a new latex object, so here we pick a convenient initial location
+		const double zoom = this->control->getWindow()->getXournal()->getZoom();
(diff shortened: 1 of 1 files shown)


LittleHuba:
I would add another option: 
Place it where the cursor is positioned.

Technius:
That works if you're using the keyboard shortcut, but not if you use the mouse to activate latex insert. I'll see if I can do something about the former...

Technius:
I think that placing the latex object where the cursor is more trouble than it's worth right now. It's non-trivial to get the location of a cursor, and I think a refactor may be needed so we can implement cursor placement for other tools such as the image tool.

Technius:
Will merge this tomorrow if there are no further objections.

PR #1235, opened 2019-06-12
Fix document close handling logic

### LINKED ISSUE #1233: Canceling open a file after discarding changes causes subsequent saves to create empty files
Hello!
First of all, thank you so much for this program! I'm a college student using Ubuntu and Xournal was killing me. This program is a huge help!

I think I have discovered a bug when trying to open a file while working on a current document that has not autosaved since last modification.

I will have a current document open and make changes before autosave has saved the file (since last autosave).
Next, I will try to open a new file, I will get a prompt if I want to discard any changes on the current document, I click discard.
I then close out of the "open a file" dialog box (nautilius on Ubuntu) and never actually open a file.
I'm now still on my original document I made modifications to before autosave (it never closed even though I clicked discard earlier). 
I go to click save and then it brings up a save dialog box and asks for a name to save the file (this shouldn't happen because this isn't a new file! It should simply save).
I proceed to give the file a new name (I don't want it to overwrite my current file incase something went wrong).

Sure enough, something goes wrong with the new file that is saved and it's only about 400 bytes whereas the original file is 5.7 MB. 

I go to open the new file with 400 bytes and it is just a blank Xournal++ screen. Was able to reproduce this multiple times.

Please see if others can reproduce this, thanks!

Technius:
Fixes #1233. Please test the PR and let me know if there is any regression!

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2232,7 +2232,7 @@ bool Control::newFile(string pageTemplate)
 {
 	XOJ_CHECK_TYPE(Control);
 
-	if (!this->close())
+	if (!this->close(true))
 	{
 		return false;
 	}
@@ -2292,7 +2292,7 @@ bool Control::openFile(Path filename, int scrollToPage, bool forceOpen)
 		return false;
 	}
 
-	if (!this->close())
+	if (!this->close(false))
 	{
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


Technius:
There's still an issue with an unnecessary close prompt after open > save > cancel that I need to fix.

Technius:
PR now ready to merge, and I'll do so this tomorrow unless someone finds a bug.

Summary of all changes in the PR:
* Fixes the open-discard-cancel save bug.
* Fixes an issue with the file save status not updating when `Control::save` is called with synchronization
* Simplified the prompt code in `Control::close`

PR #1241, opened 2019-06-12
Let's Do CSS

JJones780:
I've meant to look into this for a while...

This PR will load a CSS file called  ui/xournalpp.css

We can pretty up Xournal using it and the users can make adjustments as well.
Non-coding contributors can potentially contribute css stylings.
I'm going to be doing some cursors using it ( I've converted all except custom cursors to css  - my next PR ).

NOTE:  The CSS selector uses the "name" property NOT the id. So I had to change main.glade:
```
 <object class="GtkRevealer" id="floatingToolbox">
    <property name="name">floatingToolbox</property>
 ...
```
**Note the** `<property name="name">floatingToolbox</property>`
We will have to go through the glade files and do some proper naming in order to allow clean css selectors.

--------------
I've also changed the XojMsgBox to allow Pango markup.. and put in a bit of markup.  Rename the xournalpp.css file temporarily to see the result.
**This might cause problems for internationalization if markup characters are used that confuse Pango.
Comments?**

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -358,7 +358,8 @@ int XournalMain::run(int argc, char* argv[])
 	gtk_init(&argc, &argv);
 
 	GladeSearchpath* gladePath = new GladeSearchpath();
-	initResourcePath(gladePath);
+	initResourcePath(gladePath, "ui/about.glade");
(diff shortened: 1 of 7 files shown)


JJones780:
Also:  Anyone want to point out what needs to be done to make sure the xournalpp.css file gets installed into the proper place during installation?  I'll figure it out when I get more time.

JJones780:
Floating Toolbox Before:
image

and After: 
image

Can you spot the other css change?

PR #1243, opened 2019-06-12
Switch to new input system

LittleHuba:
This PR enables the new input system by default.
It removes the unecessary settings to enable it.

Additionally it does some cleanup:
- Separate touch scroll workaround from enabling touch drawing (redundant with the feature to assign input device classes)
- Remove functionality to assign devices as touchscreen which were not recognized as one (redundant with the feature to assign input device classes)
- Move touch scrolling workaround to input system tab as it is most likely searched there and not in touchscreen tab


LittleHuba:
This PR needs to be redone as the settings UI has changed tremendously since. This is not mergeable...

Issue #1198, opened 2019-06-03
Memory leak in BackgroundImage


Technius:
**Affects versions :**
 - OS: Linux
 - Version of Xournal++: master branch (as of eb6f2cb)

**Describe the bug**
With `-DENABLE_CPPUNIT=on` and `-fsanitize=address`, I ran `test/test-loadHandler` and got a few memory leaks.

**To Reproduce**
1. In `CMakeLists.txt`, insert the following code snippet somewhere after the `#Development options` line:
    ```
    if (DEV_MEMORY_CHECKING)
      set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
    endif (DEV_MEMORY_CHECKING)
     ```
2. In your build folder, run
     ```
    cmake .. -DENABLE_CPPUNIT=on && cmake --build .
    ```
3. Run `test/test-loadHandler`

**Expected behavior**
No memory leaks.

**Log files**

[memleak.log](https://github.com/xournalpp/xournalpp/files/3245894/memleak.log)

The log file indicates that the memory is allocated at
https://github.com/xournalpp/xournalpp/blob/eb6f2cb7328f55606cba221e1b8e390a566b3d8d/src/model/BackgroundImage.cpp#L47-L56

I'm guessing that `img` may not be destroyed properly in the destructor.

**Additional context**
I wanted to see if there were memory leaks not being caught by the current leak detection system. Behold, we _do_ have additional memory leaks, so we should look into enabling Address Sanitizer for development builds.

Febbe:
Should've fixed that in my local branch, i will make a pull request, after the actual pull request is accepted.

Febbe:
@Technius merged the PR thus I'll close this issue now.

PR #1245, opened 2019-06-12
Fixed Memory leak in BackgroundImage #1198

Febbe:
Moved impl of BackgroundImageContents into the BackgroundImage source file to reduce code bloat and increase encapsulation.
Removed unessesary header and source BackgroundImageContents.
Replaced custom memory management with a std::shared_ptr.
Refactored BackgroundImage.

Signed-off-by: Fabian Keßler <fabian_kessler@gmx.de>


src/model/BackgroundImage.h
```diff
@@ -11,44 +11,47 @@
...
-#include <XournalType.h>
+#include "XournalType.h"
 
 #include <gtk/gtk.h>
 
-class BackgroundImageContents;
+#include <memory>
 
-class BackgroundImage
+struct BackgroundImageContents;
```
Technius: Since this is internal, put it in a private section in `BackgroundImage`.

Febbe: You are right, already applied it :).

Febbe:
Tested it with the steps described by @Technius in #1198 and got 2 memory leaks without my patch and zero with the patch. I guess it works. Can I merge the request?

PR #1247, opened 2019-06-14
Update Readme file

JJones780:
See here:  https://github.com/JJones780/xournalpp/tree/ReadmeJune14-2019

Scaled down screenshots and put side by side using html instead of markdown.
Added Experimental feature notes.

**More to add? Change?**

_Note: ` pip install --user grip`  was a great way to preview the github markdown. 
https://github.com/joeyespo/grip_


LittleHuba:
It would be nice if the images in the table would all take the full size in their column. But I don't know if that is possible...

Also we should no longer recommend users to set TPCButton when the new input system is activated. Could you add this? In the experimental features this would be nice.

JJones780:
Removed the FAQ section as it only contained the outdated TPC Button issue.
Made the screenshots bigger.

Any other feature notes to add?

Issue #1250, opened 2019-06-14
Make use of clang-format


Febbe:
**Is your feature request related to a problem? Please describe.**
Developers tend to use the automatic format functions of their IDE's, those aren't configured the same way other developers configured theirs, and they even differ between your own workstations.

**Describe the solution you'd like**
Instead of letting the developers configure their format function, we should provide a clang-format configuration file, since clang-format is mostly IDE and platform independent. It also is supported by mostly every IDE.   

The provided eclipse configuration file is not sufficient and i think also out of date.

This will reduce merge conflicts due different formatting.
We can also reformat the whole project with one simple one liner over the terminal.

LittleHuba:
Sounds like a very good idea!
Would you mind creating said file?

Febbe:
Yes i can do that, but i also have some concerns and Questions about the current coding convention.

Why we use the \t character, the length depends always on the current editor and the code will look different in different editors if the programmers don't mix \t and \s in one line to ensure, that the spacing is editor independent. I would just use 4 spaces, since we accept a line width of 120 (2 for 80).

I also think that the current coding style eats lines, reducing the readability especially for short if-else statements, short functions and lambdas, i would just remove the additional line break before all {} code blocks and allow to inline those blocks if they are short and one liners.

I also would change the default placing of the const and volatile keyword after the type (or just don't care, but that could be confusing).
Since the only place where a const is allowed to be in front, when it stands in front of a type. In all other cases it must stand behind or will qualify something else or wont work:
E.g. a function void lock() const{ /*lock mutable mutex*/} i well formed, but void const lock(){ /*lock mutable mutex*/} wont compile.

or T getMyT() const{...} is a constant function which returns a copy of T, but T const getMyT(){...} will return a constant T (const T ==  T const) and the function itself cant be used in a const context anymore which would be unintended behaviour.

It also is better to read:
Ex 1: const T *const t;
Ex 2: T const* const t;

Assuming a read direction from left to right:
1) To a const T, a pointer is pointing to, which is const and named t.
2) T which is const, a pointer is pointing to, which is const, named t.

Assuming a read direction from right to left:
1) t is a constant pointer, to the type T which is const.
2) t is a constant pointer, to const T.
  
In both cases the second one is clearer to read.

veecue:
I want to drop my opinion here:
`if` should always use brackets and at least three lines for better readability:

> if (a) {
>      b;
> }

[...]

PR #1254, opened 2019-06-14
Make use of clang-format #1250

Febbe:
Added clang-format configuration file


LittleHuba:
Please change `MaxEmptyLinesToKeep: 1` to `MaxEmptyLinesToKeep: 2` for better formating of code sections. Except that small issue everything else looks good to me.
Please also update the Coding conventions in the wiki, so we do not have two different versions running around. You should also point out that people should run their changes once through clang-format before providing a PR.

Febbe:
Ok, I updated the coding conventions already. Shall I delete the old Coding Style instructions completely?

LittleHuba:
We should also schedule a PR freeze sometime in the future to run `clang-format` once on all the sources. But that has to wait until we have all the big PRs merged in.

There are a lot of conventions that are not handled by clang-format. You'll have to be careful what to remove and what to keep.

LittleHuba:
While I'm thinking about it I'll add a small automated check that makes sure people have formated their PRs before submitting them.

Febbe:
> We should also schedule a PR freeze sometime in the future to run clang-format once on all the sources. But that has to wait until we have all the big PRs merged in.

Definitely i had that problem before, resulting in a big chaos.  

> There are a lot of conventions that are not handled by clang-format. You'll have to be careful what to remove and what to keep. 

Ok, then I'll remove everything which is handled by ClangFormat.

> While I'm thinking about it I'll add a small automated check that makes sure people have formated their PRs before submitting them.

Great :)

Issue #1131, opened 2019-05-11
Right-click context menu for sidebar


Technius:
**Is your feature request related to a problem? Please describe.**
Right clicking a sidebar page preview should open a context menu. Currently, right clicking on a preview does nothing (I would expect _something_ to happen).

**Describe the solution you'd like**
The context menu affects only a single page, and it should have at least the following options:
* Insert page before
* Insert page after
* Move above
* Move below
* Delete

Here are some options that could be added, but they may used be too infrequently to be worth the time:
* Paper color
* Paper format
* Paper background

**Describe alternatives you've considered**
None.

**Additional context**

This can be implemented by
1. Designing a context menu that has the above options.
1. Listening to [`button-press-event`](https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget-button-press-event) instead of `clicked` in `SidebarPreviewBaseEntry` and performing the appropriate action based on the button pressed.

Technius:
Implemented by #1256.

PR #1256, opened 2019-06-15
Implement right-click context menu for the preview sidebar

Technius:
Implements #1131. In particular:

* Adds a context menu that appears when right clicking a sidebar preview page
* Context menu has the following functions:
    * [x] Insert page before/after
    * [x] Duplicate page
    * [x] Move up/down
    * [x] Delete page
    * [x] Edit page (only as a placeholder option--will be set to invisible by PR review time)


Technius:
Note to self:
* Remove `TODO` comments as those have already been fixed already.
* Document the memory management better.
* `gtk_menu_popup_at_pointer` is not available until 3.22, so use `gtk_menu_popup_for_device ()` instead

Technius:
Ready for review. Will merge in 3 days if there are no objections.

src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
```diff
@@ -11,13 +11,53 @@
...
+
+	for (const auto& pair: ctxMenuActions)
+	{
+		GtkWidget* const entry = gui->get(pair.first);
+		g_assert(entry != nullptr);
+
+		// Unfortunately, we need a fairly complicated mechanism to keep track
+		// of which action we want to execute.
+		typedef SidebarPreviewPages::ContextMenuData Data;
+		auto userdata = std::unique_ptr<Data>(new Data{this->toolbar, pair.second});
```
Febbe: use `auto userdata = mem::make_unique<Data>(this->toolbar, pair.second);`

src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
```diff
@@ -11,13 +11,53 @@
...
+
+		const auto callback =
+		        G_CALLBACK(+[](GtkMenuItem* item, Data* data) { data->toolbar->runAction(data->actions); });
+		const gulong signalId = g_signal_connect(entry, "activate", callback, userdata.get());
+		g_object_ref(entry);
+		this->contextMenuSignals.push_back(std::make_tuple(entry, signalId, std::move(userdata)));
```
Febbe: Try to use `this->contextMenuSignals.emplace_back(entry, signalId, std::move(userdata));`

src/gui/sidebar/previews/page/SidebarPreviewPages.h
```diff
@@ -14,6 +14,9 @@
 #include "gui/sidebar/previews/base/SidebarPreviewBase.h"
 
 #include <XournalType.h>
+#include <map>
```
Febbe:
```
#include <map>
#include <memory>
``` 
seems to be unused, move it to the source file.

src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
Technius: That doesn't work because `Data` is a struct, and trying to work around it is cumbersome.

Technius:
Rebased and will merge tomorrow if there aren't any more issues.

src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
Febbe: Just pass it as struct then, it will compile to the same assembler code.
Checkout this: https://www.godbolt.org/z/nweON9
It's a bit shorter and you don't need to use new :D.

src/gui/sidebar/previews/base/SidebarPreviewBaseEntry.cpp
```diff
@@ -14,7 +14,7 @@ SidebarPreviewBaseEntry::SidebarPreviewBaseEntry(SidebarPreviewBase* sidebar, Pa
 	XOJ_INIT_TYPE(SidebarPreviewBaseEntry);
 
 	this->widget = gtk_button_new();	// re: issue 1072
-	
+
```
LittleHuba: Remove this change as it is only formatting

src/gui/sidebar/previews/base/SidebarToolbar.cpp
```diff
@@ -22,32 +22,37 @@ SidebarToolbar::~SidebarToolbar()
 	XOJ_RELEASE_TYPE(SidebarToolbar);
 }
 
+void SidebarToolbar::runAction(SidebarActions actions)
+{
+	this->listener->actionPerformed(actions);
+}
+
```
LittleHuba: Why did you wrap this call?
I'm not really sure it has any real benefit.

ui/main.glade
```diff
@@ -2442,4 +2442,76 @@
       </object>
     </child>
   </object>
+  <object class="GtkMenu" id="sidebarPreviewContextMenu">
```
LittleHuba: You should add the name property @JJones780 introduced for CSS styling

Technius:
There's one more thing I forgot to mention. This PR should be a good starting point for a context menu for layers as well (might need some refactoring first though). I didn't add in a layers menu though since layers typically aren't modified frequently enough to warrant one.

I'll be merging this now.

Issue #1223, opened 2019-06-09
After you click on save with the pen, loading icon doesn't disappear until you write again


coldspark29:
- OS: Arch Linux
 - Version of Xournal++ : 1.0.11

After you click on save with the pen, loading icon doesn't disappear until you write again or hover the pen over the writing space.

1. Click on save with the pen
2. Hover pen over gray area
3. Hover pen over writing space

You save the file, then when you hover over the gray area the loading icon is displayed forever, unless you hover or write in the writing space again.

Technius:
Sounds like a cursor update is not triggered after the saving is finished.

LittleHuba:
Most likely this was until now hidden by the cursor being updated at every move. We should still trigger the update at the end of the task.

JJones780:
Got it..

The update is being triggered but it's falling through the cracks.

This bug dates back to Dec 8, 2018.. ( and now a comment in XournalppCursor.cpp makes much more sense to me).
I'm currently making a large number of changes to this file... I'll put the fix in with my restructuring and close this bug when it gets merged.

Thanks for reporting this bug!

coldspark29:
Great! Wouldn't have thought that the bug was that old.

I am happy to help! :) Reporting bugs is also the only thing I can do, since I lack the coding skills. I have a lot of ideas for improvement and new features. Sometimes I think that I annoy you with all my reports. Good to know that it's not the case ;)

JJones780:
This is fixed in newly merged #1257

@coldspark29 : Are you good at css...?

PR #1257, opened 2019-06-15
css cursors

JJones780:
Rewrote much of the cursor code to make it consistent. 

Uses css cursor definitions ( now preferred by gtk ) as well as Custom drawn cursors ( many of which can be migrated to pngs eventually).

Now compares requested cursor change to last cursor to avoid rebuilding and assigning cursors constantly.

Part 2 of cursorCleanup ( PR #1222 )


JJones780:
Note: also fixes #1223

PR #1259, opened 2019-06-15
 Removed memory leaks:

Febbe:
removed leak in MainWindow.cpp
removed leak in SettingsDialog.cpp

Refactored DeviceListHelper to a static factory,
because this was a factory object which owned self created data.

updated const corectness of InputDevice

### DIFF
--- a/src/gui/dialog/SettingsDialog.cpp
+++ b/src/gui/dialog/SettingsDialog.cpp
@@ -538,7 +538,8 @@ void SettingsDialog::save()
 	settings->setInputSystemTPCButtonEnabled(getCheckbox("cbInputSystemTPCButton"));
 	settings->setInputSystemDrawOutsideWindowEnabled(getCheckbox("cbInputSystemDrawOutsideWindow"));
 
-	unsigned scrollbarHideType = SCROLLBAR_HIDE_NONE;
+	auto scrollbarHideType =
+	        static_cast<std::make_unsigned<std::underlying_type<ScrollbarHideType>::type>::type>(SCROLLBAR_HIDE_NONE);
 	if (getCheckbox("cbHideHorizontalScrollbar"))
 	{
 		scrollbarHideType |= SCROLLBAR_HIDE_HORIZONTAL;
@@ -547,7 +548,7 @@ void SettingsDialog::save()
 	{
 		scrollbarHideType |= SCROLLBAR_HIDE_VERTICAL;
 	}
-	settings->setScrollbarHideType((ScrollbarHideType)scrollbarHideType);
+	settings->setScrollbarHideType(static_cast<ScrollbarHideType>(scrollbarHideType));
 
 	GdkRGBA color;
 	gtk_color_chooser_get_rgba(GTK_COLOR_CHOOSER(get("colorBorder")), &color);
--- a/src/gui/dialog/SettingsDialog.cpp
+++ b/src/gui/dialog/SettingsDialog.cpp
@@ -105,10 +105,9 @@ SettingsDialog::SettingsDialog(GladeSearchpath* gladeSearchPath, Settings* setti
 
 	initMouseButtonEvents();
 
-	auto deviceListHelper = new DeviceListHelper(false);
-	vector<InputDevice> deviceList = deviceListHelper->getDeviceList();
+	vector<InputDevice> deviceList = DeviceListHelper::getDeviceList();
 	GtkWidget* container = get("hboxInputDeviceClasses");
-	for (InputDevice const& inputDevice : deviceList)
+	for (InputDevice const& inputDevice: deviceList)
 	{
 		// Only add real devices (core pointers have vendor and product id NULL) and ignore keyboards
 		GdkDevice* device = inputDevice.getDevice();
@@ -332,7 +331,7 @@ void SettingsDialog::load()
 
 	string hidden = settings->getFullscreenHideElements();
 
-	for (const string& element : StringUtils::split(hidden, ','))
+	for (const string& element: StringUtils::split(hidden, ','))
 	{
 		if (element == "mainMenubar")
 		{
@@ -345,7 +344,7 @@ void SettingsDialog::load()
 	}
 
 	hidden = settings->getPresentationHideElements();
-	for (const string& element : StringUtils::split(hidden, ','))
+	for (const string& element: StringUtils::split(hidden, ','))
 	{
 		if (element == "mainMenubar")
 		{
@@ -460,7 +459,7 @@ string SettingsDialog::updateHideString(const string& hidden, bool hideMenubar, 
 
 	string newHidden;
 
-	for (const string& element : StringUtils::split(hidden, ','))
+	for (const string& element: StringUtils::split(hidden, ','))
 	{
 		if (element == "mainMenubar")
 		{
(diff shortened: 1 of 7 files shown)


LittleHuba:
> Refactored DeviceListHelper to a static factory,
because this was a factory object which owned self created data.

I don't really like that part :wink:
It will be quite the pain as it will conflict with #1242 🙈

Febbe:
Mhh darn, I'll look into that maybe i can provide a patch and merge it with it with your branch LittleHuba:fix_device_listing.

Febbe:
This should be able for merge since I added this bugfix also to @LittleHuba 's branch

JJones780:
Just my "2 cents":

 - Try to remember to fetch and merge master before you start a branch. 
 - Use git rebase to rebuild your changes on the most recent commit before you push.
 - Think of your commits as "paragraphs in a story" - they need to be about one thing.
 - For the near future; it's important NOT to clang-format all changed files... only changed lines!!! Use git-clang-format etc.
 - I see files that seem to have ONLY formatting changes. How? Why didn't you fix this?
 - Use `clang-format off` and `clang-format on` around manually aligned defines / case statements ( such as those found in ToolEnums.cpp .  Others have put an effort in to make those code sections readable and we have a responsibility to maintain them.
- Leave the clang-format until you are ready to push...  even your "replace NULL with nullptr" is full of clang-format changes that make it hard to see if all you changed were NULL's.


If you are having trouble with git rebase, clang-format  or other tools please don't hesitate to ask for help.  We're here as a team.

In the meantime it may be possible to weed out most of the unnecessary changes using a clang-format and a diff... then create a new branch and apply the reduced changes. Clang-format-off and on as well.  **I would be willing to help with this when you are almost ready to merge.**

LittleHuba:
I agree with the problems @JJones780 has mentioned.
We need to keep a balance between beautiful code and a readable history.

I think it would be best to cherrypick the commits that actually take care of the memory leaks and move all the other things into different PRs. As you mixed many different things into single commits you might even have to redo some work by hand. 

As this PR stands now I'm against merging it at any time. It will wreak havoc with all other PRs we still have out there.

Febbe:
Yes this is all on my Todo list since I created the PR, but I currently don't have the time to fix that. I'll do it in 3 weeks, if you didn't tried it before.

> Try to remember to fetch and merge master before you start a branch. 

I thought i've done this, this PR is just very old. And even better is to use git fetch upstream master && git reset --HARD upstream/master. So the merge doesn't even show up in the history. (<- did that.)

> I see files that seem to have ONLY formatting changes. How? Why didn't you fix this? 

Those are there, because I merged master into this branch and applied clang format after it.
So I applied clang format on the in master changed lines.

> I think it would be best to cherrypick the commits that actually take care of the memory leaks and move all the other things into different PRs. As you mixed many different things into single commits you might even have to redo some work by hand.

My plan was 
- to merge master into this branch
- create a completely new branch from upstream/master
- make a diff
- apply the diff to the new branch
- remove all NULL to nullptr only commit real changes.

or: use git rebase -i <commitdept> to delete the clang format changes and the merge with master.
then git rebase to rebase this branch onto master.
But im very unsure with git rebase, because didn't used it often.

LittleHuba:
> My plan was
to merge master into this branch
create a completely new branch from upstream/master
make a diff
apply the diff to the new branch
remove all NULL to nullptr only commit real changes.

This is the better approach as a rebase may go horribly wrong and then you won't be able to go back to how it was as you just rewrote the history 😉

Febbe:
> This is the better approach as a rebase may go horribly wrong and then you won't be able to go back to how it was as you just rewrote the history wink

 woopsy just done it (in a temporary branch) :D. Worked fine.

Febbe:
I'll merge this tomorrow, if there aren't any change requests.

LittleHuba:
Please wait for my review. Might take till Sunday...

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -539,7 +538,7 @@ void SettingsDialog::save()
 	settings->setInputSystemTPCButtonEnabled(getCheckbox("cbInputSystemTPCButton"));
 	settings->setInputSystemDrawOutsideWindowEnabled(getCheckbox("cbInputSystemDrawOutsideWindow"));
 
-	int scrollbarHideType = SCROLLBAR_HIDE_NONE;
+	unsigned scrollbarHideType = SCROLLBAR_HIDE_NONE;
```
LittleHuba: Please use `guint` as type otherwise Windows compilers may complain.

Febbe: I mostly develop c++ applications for windows this ist Not a Problem. Unsigned ist a inbuild type and also a compiler keyword. Also all enums derive drin unsigned If not other specified. Using guint would be a risky choice, because it could cause a typecast.

LittleHuba: The underlying type is implementation specific in C++11. Thus this might be the case for the compiler you are using.
As we are using `guint` as type for all other enums it is better if you stick with that.

Febbe: But this enum uses unsigned as underlying Type. And it is only a problem if you mix types, as assigning a gtype to an c++ type specifier. Thats again a reason, why it IS better to use always auto. In the case of enums auto wont work. Therefore i used the underlying type of the enum: unsigned. Guint can be problematik in this specific context.

Febbe: If we want to use guint as underlying type we must derive all enums explicitely from guint.

LittleHuba: `unsigned` is another name of `unsigned int` ([see here](https://de.cppreference.com/w/cpp/language/types)). Glib will also define `guint` as `unsigned int` if possible ([see here](https://people.gnome.org/~ryanl/glib-docs/glib-Basic-Types.html#guint)) but will use another definition if the compiler does not know of the type `unsigned int` which some Windows compilers do for some reason...

So this is only for compatibility reasons and nothing else...

andreasb242: Please don't discuss to much on such simple problem.
Our target should be to create a good note talking application, and not to discuss such simple details.

As in this specific case, where the value is uses as bitmask it makes absolutely **no difference if the value is signed or not**. And there are really bad compilers, e.g. Visual Studio (yes, the current versions are much better than the old ones), which have a lot of type issues.
Currently we are compiling with GCC on Windows, but as open source project we should write code which is as portable as possible, therefore @LittleHuba is right with portability.
@Febbe is also right, that an enum should be unsigned.

Because of this compatibility stuff it's really not bad to use standard types like int, uint_32_t (stdint.h) or gint, which is handled by GTK.

But I also checked "unsigned" is listed on the visual studio supported types: https://docs.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=vs-2019

However: please focus at the real problems!

Febbe: The problem I had with guint is, that it is not more portable, because an enums size is defined as at least as large as a "int" or "unsigned". If the compiler interprets the size differently, it will also interpret the size of an enum differently (as his int size). A "guint" is defined on each platform as uint_32_t, there will be a type difference and an implicit type cast. Which is bug prone.

Ok, in this case there will be no difference. But if we use a guint as type we should also derive the enum from it.

The best way would be something like this: 
```
std::make_unsigned_t<std::underlying_type_t<enum_T>> name = ...
```
But thats realy ugly :D.

PR #1273, opened 2019-06-18
Removed many memory leaks

Febbe:
Removed many memory leaks and possible undefined behaviour in undo / redo handling

Added a mem::make_unique<Type> helper function (equal to the stdc++14 version)


Febbe:
Please review this, I split all layout changes and replacements of NULL to nullptr into serate commits, so you should easily review my main changes in commit 380800f (Removed big...)

Febbe:
> documentation

yes at that point I still had automatic formatting on save activated, since it is required in other projects. But there all files are formatted from the beginning. Now its very hard to only remove them. I could do it, but I currently don't have the time. Thats why i split logic changes and format changes into multiple commits. Just look at commit 380800f

Febbe:
> XournalppCursor.h has no real changes and shouldn't be in this commit.
> XournalppCursor.cpp seems to be the same?

It's there because it got modified in master, I ran "git diff [placeholder] | clang-format-diff-8" on it, so only changed lines (including them of master) should be formatted. In the future I'll try to rebase my changes first, before I apply clang-format.

src/control/Control.cpp
```diff
@@ -1385,10 +1147,9 @@ void Control::deletePage()
 	this->doc->unlock();
 
 	updateDeletePageButton();
-	this->undoRedo->addUndoAction(new InsertDeletePageUndoAction(page, pNr, false));
+	this->undoRedo->addUndoAction(mem::make_unique<InsertDeletePageUndoAction>(page, pNr, false));
```
andreasb242: Can you explain why you are using make_unique?

src/control/Control.cpp
```diff
@@ -2485,8 +2101,8 @@ bool Control::loadMetadataCallback(MetadataCallbackData* data)
 
 void Control::loadMetadata(MetadataEntry md)
 {
-	MetadataCallbackData* data = new MetadataCallbackData();
-	data->md = md;
+	auto* data = new MetadataCallbackData();
+	data->md = std::move(md);
```
andreasb242: MetadataEntry does not have a move constructor, therefore I think this brings nothing?

src/control/Control.cpp
Febbe: - It's efficient (perfect forwarding)
- It's clear
- It's short
- It's correct
- I don't repeat myself 
- I don't need to use new explicitly.
- Everyone knows what it does.

src/control/Control.cpp
Febbe: It does. I don't have the source in front of me, but even the copy constructor isn't defined explicit it is defined implicit, because a child is move constructible. E.g. a std::string ord std::vector.

src/control/Control.cpp
Febbe: It also have more benefits, a reviewer, does not care, if a 'new' Object is deleted, with that he knows "ah that's a unique_ptr, perfect, I don't need to check, whether this Object is deleted or not".
Also we have a clear ownership transfer. The code is correct by design :) .

andreasb242: Ok, I didn't see the change of the List:
>std::deque<UndoActionPtr> undoList;

Then it's working.

src/control/Control.cpp
andreasb242: Ok. Tested and confirmed, it's faster.

```
	MetadataEntry e1;
	e1.path = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
	
	for(long i = 0; i < 10000000; i++) {
		MetadataEntry e2 = e1;
	}
```
and the second example with
```
		MetadataEntry e2 = std::move(e1);
```

```
time ./t1

real	0m0.285s
user	0m0.285s
sys	0m0.001s
```

And with move
```
time ./t2

real	0m0.131s
user	0m0.131s
sys	0m0.001s
```

Therefore this really brings something, even if it's really less...

Febbe: The different between copy and move is basically, copy makes a deep copy, move a flat copy + invalidating the old Object (set all ptrs to nullptr; ). So every Object owning pointers is faster to move than to copy, since we don't need to allocate and deallocate the old and new ptr. It also is exception safe.

PR #1274, opened 2019-06-18
Correct "dash-/ doted" typo in drop down menu

kylegodbey:
Simple fix, I think I found everywhere it was misspelled.


JJones780:
Thanks for catching the typo.. why didn't I see that?! ;)

PR #1275, opened 2019-06-19
Clang format

JJones780:
This .clang-format is very close to our existing format spec.
Try it with  `BinPackParameters: false` as well.
I put in `SortIncludes: false` to avoid changing existing files.. but for a new or massively changed file it might be nice to sort.

Todo: indent split argument lines more?

Note to self:
`clang-format-6.0 gui/XournalView.cpp > /tmp/clangdiff2.cpp ; kdiff3 /tmp/clangdiff2.cpp gui/XournalView.cpp`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -23,7 +23,7 @@ BreakConstructorInitializersBeforeComma: true
 ColumnLimit: 120
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 4
-ContinuationIndentWidth: 4
+ContinuationIndentWidth: 16
 
 DerivePointerAlignment: false # not in 3.4
 
--- a/.clang-format
+++ b/.clang-format
@@ -1,69 +1,62 @@
-﻿---
-BasedOnStyle: Google
-AccessModifierOffset: '-4'
-AlignAfterOpenBracket: Align
-AlignConsecutiveAssignments: 'false'
-AlignConsecutiveDeclarations: 'false'
-AlignEscapedNewlines: Left
-AlignOperands: 'true'
-AlignTrailingComments: 'true'
-AllowAllParametersOfDeclarationOnNextLine: 'true'
-AllowShortBlocksOnASingleLine: 'false'
-AllowShortCaseLabelsOnASingleLine: 'true'
-AllowShortFunctionsOnASingleLine: Inline
-AllowShortIfStatementsOnASingleLine: 'false'
-AllowShortLoopsOnASingleLine: 'false'
-AlwaysBreakAfterDefinitionReturnType: None
-AlwaysBreakAfterReturnType: None
-AlwaysBreakBeforeMultilineStrings: 'false'
-AlwaysBreakTemplateDeclarations: 'true'
-BinPackArguments: 'false'
-BinPackParameters: 'true'
-BreakBeforeBinaryOperators: None
-BreakBeforeBraces: Linux
-BreakBeforeTernaryOperators: 'false'
-BreakConstructorInitializers: AfterColon
-BreakInheritanceList: BeforeColon
-BreakStringLiterals: 'true'
-ColumnLimit: '120'
-CompactNamespaces: 'true'
-ConstructorInitializerAllOnOneLineOrOnePerLine: 'false'
-ConstructorInitializerIndentWidth: '8'
-ContinuationIndentWidth: '8'
-Cpp11BracedListStyle: 'true'
-DerivePointerAlignment: 'false'
-DisableFormat: 'false'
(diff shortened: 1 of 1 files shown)


Febbe:
I would revert UseTab: back to ForIndentation, since this ensures, that the indent is always the same with different tab length's in different Editors.

JJones780:
Thanks for spotting those.
Fixed.

Febbe:
For me it looks like you created a hole new file instead of correcting the old one.
There are several Options you kicked out and some which were right before, but are wrong now.
And there are some outdated Options:

```
AlignAfterOpenBracket: Align
AlignConsecutiveAssignments: false
AlignConsecutiveDeclarations: false
```
(lost)

replace: `AlignEscapedNewlinesLeft: true `(removed) with `AlignEscapedNewlines: Left`

`AlignOperands: true` (lost)

`AllowAllParametersOfDeclarationOnNextLine:`is differently handled by different devs and was not defined in the standard, but to reduce vertical spacing i would use true instead of false.
 
change `AllowShortIfStatementsOnASingleLine:` back to false, but consider it to make it true, when we format the hole project.

remove `AlwaysBreakAfterDefinitionReturnType`, since it is deprecated or set it to None (false is definitely wrong)

add missing `AlwaysBreakAfterReturnType: None` (replaced the option above) 

change `AlwaysBreakBeforeMultilineStrings:` to false, since we didn't done it before, and it increases vertical spacing

Set `BreakBeforeBinaryOperators:` to None, since we broke the line after the '&, | ...' or it was differently handled by different devs (not defined by our standard).

Set `BreakBeforeTernaryOperators:`  to false, because i think no one did that before.

Replace `BreakConstructorInitializersBeforeComma` (removed) with
```
BreakConstructorInitializers: AfterColon
BreakInheritanceList: AfterColon
```

Add `BreakStringLiterals: true` (lost)

Add `CompactNamespaces: true` (lost, place it after ColumnLimit: 120)

Here a opinion based change request, since it wasn't handled before, but 4 for constructors is to short and 16 generally to much (had pretty ugly changes in my edited files, when i reformatted the hole file):
```
ConstructorInitializerIndentWidth: 8
ContinuationIndentWidth: 8
```
I think 8 is a good tradeoff. It isn't that long, but it is different to the default indent, this ensures, that the indent of a linebreak is never the same as the following block. 

Add `Cpp11BracedListStyle: true``(lost)

Add `DisableFormat: false` (lost, but default false)

Set `ExperimentalAutoDetectBinPacking:` to true, this ensures, that a decision to unbinpack a function is not overiden by the formatter, it also has the advantage, that it reduces differences between old and new format (not defined and devs mixed it before).

Add `FixNamespaceComments: true` (lost)

Add `IncludeBlocks: Preserve` (lost and very important)

Add `IndentPPDirectives: None` (lost, previously `AfterHash`, but the desired option `BeforeHash` is only implemented in clang9)

Change `IndentWrappedFunctionNames: false` to true (more compliant to old code)

That's actually very cool: `PenaltyReturnTypeOnItsOwnLine: 10000` I'll experiment with that :), But i thought we only want this on a format change (when we reformat the project, leave it there since it wont have an effect currently).

Add `ReflowComments: false #true` (true would refactor old comments but looks better afterwards)

Remove `# PointerBindsToType: true` (removed by clang and it is a comment)

Add `SortUsingDeclarations: false #true`

Add `SpaceAfterTemplateKeyword: true` (lost)

Add `SpaceBeforeCpp11BracedList: false` (lost)

Add `SpaceBeforeCtorInitializerColon : true` (lost, differently handled, but mostly true)

Add `SpaceBeforeInheritanceColon: true` (lost, differently handled, but mostly true)

Add `SpaceBeforeRangeBasedForLoopColon: false` (lost, differently handled, but mostly false)

Add `SpacesInContainerLiterals: false` (lost)

After all this should match the old / current standard pretty well then :).

I will make a pull request with all the changes above, so you can accept it or ask for changes.
I also added a command to the wiki, which helps just to refactor the lines you edited :) .

If we use it, it will reduce differences through ClangFormat.

JJones780:
No, sorry I did not use the recent clang-format file. My only goal was to match our existing format with a clang-format file so that we can move forward from a matching start. 

I expect there will be many improvements _but any changes that don't match our current coding style should be considered wrong until approved by Andreas and the rest of the developers._

Thanks for looking it over. I'll do some comparisons against existing well-formatted code and report back.

Febbe:
> I expect there will be many improvements but any changes that don't match our current coding style should be considered wrong until approved by Andreas and the rest of the developers.

I aprove to that, but your file made more wrong than it helped. The old file was already very near to the standard.

JJones780:
My apologies.
I'm glad we're on the right track now though which wouldn't have happened if you hadn't started things off.

PR #1283, opened 2019-06-20
Support CSS:  Add name="name" property matching id's for all non-default id's

JJones780:
Now that there is a xournalpp.css file  someone might want to style parts of Xournal++
I've added CSS accessible names to the glade file which match the id's. Only objects with non-generic id's have been done.

Script used:
```
#do not run twice... be careful but do delete current name="names" lines before 
#running this again by using:
#    sed -r -i  '/.*name="name"/d' ui/*.glade
#Note: remove the -i to test.

gawk -i inplace 'match($0, /(\s*).*class="(\S+)".*id="(\S+)"/,arr){ print $0; if(arr[2] != "GtkAdjustment") print arr[1] "  <property name=\"name\">" arr[3] "</property>"; next}{print}'  ui/*.glade
#now remove lines generated from for non-generic id's:
for xx in sid label vbox hbox box separatormenuitem menuitem image grid menu button; \
do sed -r -i  "/.*name=\"name\".*\\b$xx[0-9]+\\b/d" ui/*.glade;done
```

### DIFF
--- a/ui/about.glade
+++ b/ui/about.glade
@@ -3,6 +3,7 @@
 <interface>
   <requires lib="gtk+" version="3.0"/>
   <object class="GtkDialog" id="aboutDialog">
+    <property name="name">aboutDialog</property>
     <property name="width_request">320</property>
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">About Xournal++</property>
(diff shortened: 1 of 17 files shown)


JJones780:
The end goal is to allow more CSS styling. See https://developer.gnome.org/gtk3/stable/chap-css-overview.html
Changes are also needed in the non-glade gtk objects added via code to fully support css styling.
gtk_widget_set_name(widget, name)
and this page might be of interest https://developer.gnome.org/gtk3/stable/gtk-migrating-GtkStyleContext-checklist.html

**GTK Inspector** is a very handy tool for experimenting with styles and identifying css objects to apply styles to. https://wiki.gnome.org/action/show/Projects/GTK/Inspector

There is the danger that opening up Xournal++ to css styling  will result in more issues being raised.

PR #1289, opened 2019-06-24
Bugfix: The undoList of UndoRedoHandler is not cleared after calling …

Febbe:
…UndoRedoHandler::clearContents().

As a result the undoList is not cleared after creating a new Document.

Signed-off-by: Fabian Keßler <fabian_kessler@gmx.de>


Febbe:
I'll merge this directly after the CI checks because it is a obvious bug caused by my previous PR.

PR #1303, opened 2019-06-28
Refactored XournalType to remove empty statement warnings of clang-tidy

Febbe:
this is a commit from refactor_layout_resize but has nothing to do with the layout or the resize.
But it is recommended change, since it removes nearly all warnings regarding the XournalType checks. 
(empty statement)


LittleHuba:
Nice! Those were awful for code checking...

Febbe:
You're welcome.

PR #1309, opened 2019-06-28
Enhancement staves (originally #1090)

LittleHuba:
Taking over as the original PR (#1090) was orphaned by @peetCreative


src/control/pagetype/PageTypeHandler.cpp
```diff
@@ -117,3 +121,82 @@ vector<PageTypeInfo*>& PageTypeHandler::getPageTypes()
 
 	return this->types;
 }
+
+PageTypeFormat PageTypeHandler::getPageTypeFormatForString(const string& format)
```
veecue: This is nicer to implement with a static const std::map

src/control/settings/PageTemplateSettings.cpp
```diff
@@ -3,6 +3,7 @@
 #include "control/xojfile/SaveHandler.h"
 
 #include <sstream>
+#include <control/pagetype/PageTypeHandler.h>
```
veecue: use dqoutes, group it to the others alphabetically and move the STL import to the top

src/control/xojfile/LoadHandler.cpp
```diff
@@ -13,6 +13,7 @@
 #include <gtk/gtk.h>
 
 #include <stdlib.h>
+#include <control/pagetype/PageTypeHandler.h>
```
veecue: user dquotes and group properly

src/control/xojfile/LoadHandler.cpp
```diff
@@ -693,7 +694,7 @@ void LoadHandler::parseStroke()
 
 	/** read stroke timestamps (xopp fileformat) */
 	const char* fn = LoadHandlerHelper::getAttrib("fn", true, this);
-	if (fn != NULL)
+	if (fn != NULL && strcmp(fn, "") != 0)
```
veecue: strlen would be more efficient, also use nullptr while you're at it

src/control/xojfile/LoadHandler.cpp
```diff
@@ -795,7 +796,7 @@ void LoadHandler::parseText()
 	text->setColor(color);
 
 	const char* fn = LoadHandlerHelper::getAttrib("fn", true, this);
-	if (fn != NULL)
+	if (fn != NULL && strcmp(fn, "") != 0)
```
veecue: same as above

src/control/xojfile/SaveHandler.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include <config.h>
 #include <i18n.h>
+#include <control/pagetype/PageTypeHandler.h>
```
veecue: use dquotes and group properly

src/control/xojfile/XojExportHandler.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <config.h>
 #include <i18n.h>
+#include <control/pagetype/PageTypeHandler.h>
```
veecue: use dquotes and group properly

src/model/PageType.h
```diff
@@ -13,11 +13,24 @@
 
 #include <XournalType.h>
 
+enum class PageTypeFormat {
```
veecue: All-caps for enums isn't what you would usually use for C++ and looks rather ugly to read tbw

src/view/background/LineBackgroundPainter.h
```diff
@@ -13,12 +13,13 @@
 
 #include "BaseBackgroundPainter.h"
 
+#include "LineBackgroundType.h"
```
veecue: properly group imports

src/view/background/LineBackgroundType.h
```diff
@@ -0,0 +1,18 @@
...
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+enum LineBackgroundType
+{
+	LINE_BACKGROUND_LINED,
```
veecue: Same as above, why not PascalCase for Enums? Also could make this an `enum class`

veecue:
And some comments after testing:
 - Functionwise everything looks good
 - I don't see a use case for staves with vertical line
 - I would prefer them to be light gray (like the graph) instead of blue
 - Usually staves have a small border at the left and right side (1-2cm), looks kind of weird this way

LittleHuba:
> Usually staves have a small border at the left and right side (1-2cm), looks kind of weird this way 

Then I'll have to undo some of the previous work. The staves painter is currently inheriting from the line painter, which is unaware of borders.

src/control/pagetype/PageTypeHandler.cpp
LittleHuba: `static const std::map` is unfortunately not a nice way to do this as it may throw a non-catchable exception at initialization. Additionally for a map of this size the runtime will be equal to running through a few conditions.

veecue:
Looks really good now, thanks! (sry, missed to re-review after your pushes)

PR #1310, opened 2019-06-28
Refactoring

Febbe:
Again some refactoring commits previously done in refactor_layout_resize, which do not belong there.

### DIFF
--- a/src/control/layer/LayerController.cpp
+++ b/src/control/layer/LayerController.cpp
@@ -10,8 +10,8 @@
 #include "util/cpp14memory.h"
 
 LayerController::LayerController(Control* control)
- : control(control),
-   selectedPage(size_t_npos)
+ : control(control)
+ , selectedPage(npos)
 {
 	XOJ_INIT_TYPE(LayerController);
 }
--- a/src/util/Path.cpp
+++ b/src/util/Path.cpp
@@ -1,20 +1,14 @@
 #include "Path.h"
+
 #include "StringUtils.h"
 
 #include <glib/gstdio.h>
-#include <string.h>
 
-Path::Path()
-{
-}
-
-Path::Path(const Path& other)
- : path(other.path)
-{
-}
+#include <cstring>
+#include <utility>
 
 Path::Path(string path)
- : path(path)
+ : path(std::move(path))
 {
 }
 
@@ -23,163 +17,37 @@ Path::Path(const char* path)
 {
 }
 
-Path::~Path()
+Path& Path::operator=(string p)
 {
+	this->path = std::move(p);
+	return *this;
 }
 
-/**
- * @return true if empty
- */
-bool Path::isEmpty()
+Path& Path::operator=(const char* p)
 {
-	return path.empty();
+	this->path = p;
+	return *this;
 }
 
-/**
- * Check if this is a file which exists
- */
-bool Path::exists()
+Path& Path::operator/=(const Path& p)
 {
-	return g_file_test(path.c_str(), G_FILE_TEST_EXISTS);
+	return *this /= p.str();
 }
 
-/**
- * Delete the file
- *
- * @return true if the file is deleted or does not exists
- */
-bool Path::deleteFile()
+Path& Path::operator/=(const string& p)
 {
-	if (!exists())
+	if (!path.empty())
 	{
-		// Does not exists, therefore cannot be deleted
-		return true;
+		path.reserve(path.size() + p.size() + 1);
+		path += G_DIR_SEPARATOR_S;
 	}
-
-	int result = g_unlink(c_str());
-	if (result == 0)
-	{
-		return true;
-	}
-
-	return false;
+	path += p;
+	return *this;
 }
 
-/**
- * Compare the path with another one
- */
-bool Path::operator ==(const Path& other)
+Path& Path::operator/=(const char* p)
 {
-	return this->path == other.path;
-}
-
-/**
- * Assign path
- */
-void Path::operator =(const Path& other)
-{
-	this->path = other.path;
-}
-
-/**
- * Assign path
- */
-void Path::operator =(const string& path)
-{
-	this->path = path;
-}
-
-/**
- * Assign path
- */
-void Path::operator =(const char* path)
-{
-	this->path = path;
-}
-
-/**
- * @return true if this file has .xopp or .xoj extension
- */
-bool Path::hasXournalFileExt()
-{
-	return hasExtension(".xoj") || hasExtension(".xopp");
-}
-
-/**
- * Check if the path ends with this extension
- *
- * @param ext Extension, needs to be lowercase
- * @return true if the extension is there
- */
-bool Path::hasExtension(string ext)
-{
-	if (ext.length() > path.length())
-	{
-		return false;
-	}
-
-	string pathExt = path.substr(path.length() - ext.length());
-	pathExt = StringUtils::toLowerCase(pathExt);
-
-	return pathExt == ext;
-}
-
-#define REMOVE_EXTENSION(ext) \
-	if (StringUtils::endsWith(plower, ext)) \
-	{ \
-		path = path.substr(0, path.length() - strlen(ext)); \
-		return;\
-	}
-
-/**
- * Clear the the last known extension (last .pdf, .pdf.xoj, .pdf.xopp etc.)
- */
-void Path::clearExtensions()
-{
-	string plower = StringUtils::toLowerCase(path);
-	
-	REMOVE_EXTENSION(".xoj");
-	REMOVE_EXTENSION(".xopp");
-}
-
-/**
- * Return the Path as String
- */
-const string Path::str() const
-{
-	return path;
-}
-
-/**
- * Return the Path as String
- */
-const char* Path::c_str() const
-{
-	return path.c_str();
-}
-
-/**
- * Get escaped path, all " and \ are escaped
- */
-string Path::getEscapedPath()
-{
-	string escaped = path;
-	StringUtils::replaceAllChars(escaped, {
-		replace_pair('\\', "\\\\"),
-		replace_pair('\"', "\\\"")
-	});
-
-	return escaped;
-}
-
-void Path::operator /=(Path p)
-{
-	*this /= p.str();
-}
-
-void Path::operator /=(string p)
-{
-	if (path.size() > 0)
+	if (!path.empty())
 	{
 		char c = path.at(path.size() - 1);
 		if (c != '/' && c != '\\')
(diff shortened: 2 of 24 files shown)


src/util/Path.cpp
```diff
@@ -1,20 +1,14 @@
...
-Path::Path()
-{
-}
-
-Path::Path(const Path& other)
- : path(other.path)
-{
-}
+#include <utility>
+#include <cstring>
```
veecue: properly group imports if you are refactoring them anyway:
 - at the beginning: corresponding header
 - first group STL
 - second group other libs (eg glib)
 - third group project files

src/util/Util.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
+
 #include "Util.h"
```
veecue: This would be the first line

src/util/Util.h
```diff
@@ -16,47 +16,42 @@
...
 #include <string>
-using std::string;
-
 #include <functional>
+#include <limits>
 
+using std::string;
 
-class Util
+namespace Util
```
veecue: if this is a namespace, please name it `util`

src/util/Path.cpp
Febbe: Right that's the common standard, but in this project it is:
- at the beginning: corresponding header
- first group project files
- second group other libs (eg glib)
- third group STL

src/util/Util.cpp
Febbe: damn auto includes :S

src/util/Util.h
Febbe: Normally I would do that and for project independent stuff I do it even in this project. But the naming convention in the Wiki says: "Names are CamelCase", since Util:: is a project dependant namespace I'll name it with CamelCase; It also does not make a different.

src/util/Util.cpp
Febbe: done

veecue:
Thanks for the clarifications, will help me when I finally have the time to contribute some more code :+1:

src/control/layer/LayerController.cpp
```diff
@@ -10,8 +10,8 @@
 #include "util/cpp14memory.h"
 
 LayerController::LayerController(Control* control)
- : control(control),
-   selectedPage(size_t_npos)
+ : control(control)
+ , selectedPage(npos)
```
andreasb242: The Comma should always be at the end of the line, not at the beginning.

src/util/Path.cpp
```diff
@@ -272,29 +249,29 @@ GFile* Path::toGFile()
...
  */
-Path Path::getParentPath()
+Path Path::getParentPath() const
 {
 	size_t separator = path.find_last_of("/\\");
 
 	if (separator == string::npos)
 	{
-		return "";
+		return Path{""};
```
andreasb242: I don't know this syntax, what are the advantages of this?

src/util/Path.cpp
```diff
@@ -272,29 +249,29 @@ GFile* Path::toGFile()
...
 
-	return path.substr(0, separator);
+	return Path{path.substr(0, separator)};
 }
 
 /**
  * Convert an uri to a path, if the uri does not start with file:// an empty Path is returned
  */
-Path Path::fromUri(string uri)
+Path Path::fromUri(const string& uri)
```
andreasb242: Do not change this, else it's not possible to call this from anything else than a std::string, keep it consistent over the whole API of the class.

src/util/Util.cpp
```diff
@@ -37,34 +37,37 @@ static bool execInUiThreadCallback(CallbackUiData* cb)
...
 	cairo_set_source_rgb(cr, r, g, b);
 }
 
 
 void Util::apply_rgb_togdkrgba(GdkRGBA& col, int color)
 {
-	col.red = ((color >> 16) & 0xFF) / 255.0;
-	col.green = ((color >> 8) & 0xFF) / 255.0;
-	col.blue = (color & 0xFF) / 255.0;
+	auto tmp = static_cast<unsigned>(color);
```
andreasb242: This casts are not needed, it makes the code more complicated, and therefore worse to read. Don't do this, except if you have a really good explanation.

src/util/Util.cpp
```diff
@@ -37,34 +37,37 @@ static bool execInUiThreadCallback(CallbackUiData* cb)
...
+	col.blue = (tmp & 0xFFu) / 255.0;
 	col.alpha = 1.0;
 }
 
 int Util::gdkrgba_to_hex(GdkRGBA& color)
 {
-	return (((int)(color.red * 255)) & 0xff) << 16 |
-			(((int)(color.green * 255)) & 0xff) << 8 |
-			(((int)(color.blue * 255)) & 0xff);
+	return ((static_cast<unsigned>(color.red * 255)) & 0xffu) << 16u |
```
andreasb242: Cast to unsigned and return as int makes the code only more complicated, and brings nothing. The same as above: Do you have a good reason?

src/control/layer/LayerController.cpp
Febbe: I know, but there is a bug in clang-format, so we decided to use this version until the bug is fixed, because it looks more similar to the old style. I could use /clang-format off

andreasb242: Ok

src/util/Path.cpp
Febbe: That's an explicit initializer, it replaced the old c/java styled () - braced style. Its deprecated to use those () braces or Type varname = Type(vars). But actually are no differences to Path("");

The correct way to initialize variables is:
Type typename{...};
for return Type{...}
in c++17 return {...}

src/util/Path.cpp
Febbe: Actually that's not correct. A `Type const&` binds to everything also to temporaries (as `auto&&` or `Type&&`) since string has no explicit Constructors (it has a converting) constructor it will work without a problem.
It is actually more efficient.

src/util/Util.cpp
Febbe: It removes false positive compiler warnings. Binary operators should not be used with unsigned values, because it is bug prone. That's why compilers and clang-tidy complain. Actually it would be better to return an unsigned, instead of an int. And i don't see a difference in readability between a cast to int or a cast to unsigned.

src/util/Util.cpp
Febbe: Shifting an unsigned behaves like you think the int behaves. But to shift an int behaves differently.
Shifting an int 1000 0000 1110 0000 to the right by one:
1000 0000 0111 0000 
Shifting an unsigned: 1000 0000 1110 0000 :
0100 0000 0111 0000
So that's why no one should shift signed values if he really don't know what he's doing.
And yellow marked code is very annoying.

src/util/Path.cpp
andreasb242: Checked an compiling with g++ 7.4

Not sure when this was changed: But this was definite not working with older versions, the main reason why I added nearly nowhere const to the parameter...

andreasb242: Then the additional const char* overload methods also should not be needed, right?
I meand operator/= etc.

Febbe: I'll do a rework to that class, seems like i can clean up more :D
And it should be supported since the first compiler supporting c++11. So somewhere around 2009.

src/util/Util.cpp
andreasb242: No, this is wrong.

´´´
#include <stdio.h>

int main() {
	unsigned a = 0b1000000011100000;
	int b = 0b1000000011100000;

	printf("a1=%04x\n", a);
	printf("b1=%04x\n", b);
	
	a = a >> 1;
	b = b >> 1;
	
	printf("a2=%04x\n", a);
	printf("b2=%04x\n", b);
	
	return 0;
}
´´´
g++ main.cpp -o main
./main

´´´
a1=80e0
b1=80e0
a2=4070
b2=4070
´´´

Febbe: Oh you misunderstood me, i used a 16bit int so i dont need to write so much 0's.
Look at this site: https://www.godbolt.org/z/fyP17H there is the proof for you.

Febbe:
```
int square(int num) {
    int i = 0xf000000f;
    return i>>31u;
} // returns -1

int square2(int num) {
    unsigned i = 0xf000000f;
    return i>>31u;
} //returns 1```

File: src/control/layer/LayerController.cpp (hunk 4 above)
Comment (Febbe) [reply src/control/layer/LayerController.cpp]: I think using `//clang-format off ... on` isn't the best idea, because no one will ever find and remove it. We should just wait for clang-9 or 10 fixing this issue.

File: src/util/Path.cpp (hunk 6 above)
Comment (Febbe) [reply src/util/Path.cpp]: Ok the functions with `char const*` aren't needed, but if I remove them, everytime someone passes a cstring it will be converted to a std::string first, which is not necessary.

Comment (andreasb242) [reply src/util/Path.cpp]: Ok

File: src/util/Path.cpp [hunk 9]
```diff
@@ -23,214 +17,153 @@ Path::Path(const char* path)
...
 }
 
-/**
- * @return true if empty
- */
-bool Path::isEmpty()
+Path& Path::operator=(const char* p)
 {
-	return path.empty();
+	this->path = p;
```
Technius: This string should be copied because `p` could be freed outside of `Path`.

Febbe: It is copied. this->path is a std::string, passing a char* to it will copy the C-String.

Technius:
```suggestion
	this->path = strdup(p);
```

Febbe: Your suggestion will lead to an memory leak.

src/util/Path.cpp
```diff
@@ -23,214 +17,153 @@ Path::Path(const char* path)
...
-bool Path::deleteFile()
+Path& Path::operator/=(const string& p)
 {
-	if (!exists())
+	path.reserve(path.size() + p.size() + 1);
+	if (!path.empty())
 	{
-		// Does not exists, therefore cannot be deleted
-		return true;
+		char const c = path.at(path.size() - 1);
```
Technius: What's this for?

src/util/Path.cpp
Technius: Nevermind, I forgot that `path` is a `std::string` and not a `char*`.

src/util/Path.cpp
Febbe: Actually nothing, thanks.

src/util/Path.h
```diff
@@ -20,57 +20,49 @@ using std::string;
...
-	void operator =(const char* path);
+	Path& operator=(const Path& other) = default;
+	Path& operator=(Path&& other) = default;
+	Path& operator=(string path);
+	Path& operator=(const char* path);
 
 	/**
 	 * Check if the path ends with this extension
 	 *
 	 * @param ext Extension, needs to be lowercase
```
Technius: Indicate in the documentation that `ext` must be prefixed with a `.` or else the extension check will not work correctly (it will only compare with the suffix of the path`).

Febbe: I just enabled the possibility to use both.

Febbe:
I'll merge this in the evening (in 12h) if no one complains.

Issue #1168, opened 2019-05-21
Presentation on second display


GregDMeyer:
This might be a little non-trivial to implement, but I thought I'd put it out there. It would be super cool if it was possible to show *only* the page area on a second display. What I am imagining is writing on my tablet with all the toolbars etc visible on the native display, and only what I am drawing being displayed on e.g. a projector.

Right now it is possible to give a presentation on a second display by just mirroring displays, but then all of the toolbars etc are visible, and in general this requires changing the resolution of the native display. Another option (on linux) is to use xrandr to export only a portion of the framebuffer, but this is pretty hacky ;)

I can poke around how to implement this if I have some spare time, but thought I would put the idea in public in case someone else sees an easy way to implement it.

LittleHuba:
This is actually not that hard as we are using the MVC pattern which supports such things. We only need to generate the window with this single view and change a few parts of the code to handle multiple such views.

Unfortunately we have a lot to do currently so this may still take some time.

JJones780:
Would the floatingToolbox (i.e. popup menu ) suffice for your use case? This is currently under development (#1161) but it allows one to use a full screen page only display. Here's a gifanim showing functionality without any sidewindows etc. You can put only the tools you need in the floating toolbar.
![allpage small](https://user-images.githubusercontent.com/2329541/58144018-0fe9ab80-7c0a-11e9-84c8-987c258504de.gif)

GregDMeyer:
@JJones780 Thanks! This could be a good temporary solution, though I'm not sure it does everything I am looking for---

There is another consideration aside from the toolbars. On my tablet, mirroring displays with e.g. a projector forces me to set my builtin display to a non-native resolution with an aspect ratio that doesn't match. Not only is this less visually nice, it also totally screws up the calibration of my tablet pen with position on the screen (so where I'm writing isn't where my pen is touching). Admittedly that's my problem ;) but I think it would be a nice, clean solution to export the page to a non-mirrored display. (I think some other note-taking software does this, but I don't have any concrete examples).

Thanks for your time and input!!

JJones780:
Are you using the same projector most of the time? You mentioned you are proficient with xrandr so it might be worthwhile to script an xrandr command to set it up ( see below )
[...]

PR #1311, opened 2019-06-29
Left and right key for navigation in presentation mode

wallento:
As the presentation remotes use the left and right keys for navigation, add the setting to enable them in presentation mode. The presenter can then use the remote and the xournalpp features, but not move elements with those keys anymore, hence its better to be an option.

This is a very tiny first step towards the wish list in #1168

### DIFF
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -318,7 +318,14 @@ bool XournalView::onKeyPressEvent(GdkEventKey* event)
 		}
 		else
 		{
-			layout->scrollRelative(-scrollKeySize, 0);
+			if (control->getSettings()->isPresentationMode())
+			{
+				control->getScrollHandler()->goToPreviousPage();
+			}
+			else
+			{
+				layout->scrollRelative(-scrollKeySize, 0);
+			}
 		}
 		return true;
 	}
@@ -331,7 +338,14 @@ bool XournalView::onKeyPressEvent(GdkEventKey* event)
 		}
 		else
 		{
-			layout->scrollRelative(scrollKeySize, 0);
+			if (control->getSettings()->isPresentationMode())
+			{
+				control->getScrollHandler()->goToNextPage();
+			}
+			else
+			{
+				layout->scrollRelative(scrollKeySize, 0);
+			}
 		}
 		return true;
 	}


src/gui/XournalView.cpp
```diff
@@ -257,14 +257,30 @@ bool XournalView::onKeyPressEvent(GdkEventKey* event)
 
 	if (event->keyval == GDK_KEY_KP_Left)
 	{
-		this->pageRelativeXY(-1, 0);
-		return true;
+		if (control->getSettings()->isPresentationMode() && control->getSettings()->getPresentationLeftRightEnabled())
+		{
+			control->getScrollHandler()->goToPreviousPage();
+			return true;
```
Technius: Bring the `return true` outside of the two if statements since they're duplicated.

src/gui/XournalView.cpp
```diff
@@ -257,14 +257,30 @@ bool XournalView::onKeyPressEvent(GdkEventKey* event)
...
-		return true;
+		if (control->getSettings()->isPresentationMode() && control->getSettings()->getPresentationLeftRightEnabled())
+		{
+			control->getScrollHandler()->goToNextPage();
+			return true;
+		}
+		else
+		{
+			this->pageRelativeXY(1, 0);
+			return true;
```
Technius: Same here: lift out the `return true`

src/gui/XournalView.cpp
wallento: Ah, right, I have fixed it, amended and force pushed, thanks!

wallento:
Thanks. Indeed you are right about the wording, I somehow assumed that left and right are moving the active element, because I never zoomed in :)

I will reword and force push later today or tomorrow with a better wording.

wallento:
Hm, so now that I am thinking of it, it probably shouldn't even be a setting. Pages are always zoom fit, hence left and right just have no effect, making left+right switch pages would just add something new and not hide an old feature, right? Happy to remove the setting if you agree.

Technius:
You should check if there are any actions (e.g. using left/right to move a selected object while in presentation mode) that the new keybindings overlap with. If there aren't any, there's no need for the toggle (otherwise, keeping the toggle should be fine).

JJones780:
Make note that the numpad arrow keys behave slightly differently (move by entire pages without pressing shift ) than the regular arrow keys and that the numpad arrow keys might already act appropriately ( except for the right not wrapping around).

It would be best not to change the behaviour of the numpad arrow keys unless required.

The presentation mode has a fixed layout and I don't think it was thought out how it would work for right to left languages (as the two column paired-pages mode was enabled toward the end).

wallento:
Yes, you are right, actually the patch will reduce to only Left+Right in Presentation. But at least I learned the code base a bit ;)

Regarding your second comment, I am not fully sure if the vision I have for the presentation mode actually fits the current cases entirely (see #1168). Maybe a "Presenter View" or "Classroom View" would be better suited. I will start this discussion in that issue or a new one once I have a better idea about how it will look. But it is pretty composable I think, once the functionality is there (currently working on the second screen).

wallento:
So, there we are. It was much easier than anticipated. The fundamental difference is that it only uses the left and right key and not their keypad variant too. Also, it respects the Shift-key combination to move within the dual page view. Masking the relative scrolling is okay as it didn't work in presenter view before (dual page) or is not needed (single page), I think.

JJones780:
Thanks for noting this issue and getting your hands dirty with the code.

I look forward with your investigation into the dual screen implementation.
One thing I had wondered about was the possibility of sending the info packaged up for the undo system to another copy of xournal++ ( and vice-versa ) as a quick and dirty multi-user editing session.. or a "read-only" second screen mirroring.

Issue #1285, opened 2019-06-22
"Export as..." crashes, glade error introduced with latest update


edo0:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Elementary OS 5.0 Juno (based on Ubuntu 18.04 LTS)
 - Version of Xournal++:   1.0.6-0 ~201906220514 ~ubuntu18.04.1 

**Describe the bug**
"Error loading glade file ..." (please refer to screenshot) when trying to "Export as..."

**To Reproduce**
1. Open a .xopp file
2. Ctrl + E      or     File > Export as...
3. Fill in name and press save

**Expected behavior**
A dialogue allowing you to choose which pages to export as PDF should appear.
In its place, the following error dialogue appears, and once you close the dialogue Xournal++ just closes down

**Screenshots of Problem**
![Screenshot from 2019-06-22 17 00 07](https://user-images.githubusercontent.com/16632292/59965550-ffca3200-950f-11e9-8107-73a8484a2b9c.png)

**Additional context**
The bug started presenting itself with today's release (1.0.6-0 ~201906220514 ~ubuntu18.04.1) - I am sure because I update the software daily using the PPA, and export notes as PDF on a daily basis.

Technius:
Regression introduced by #1283. @JJones780

Technius:
Fix should be to just remove this line:
https://github.com/xournalpp/xournalpp/blob/bd7f66472ba7e9ba5db03265e3cd8142144f9fc0/ui/exportSettings.glade#L315

PR #1317, opened 2019-07-02
Update exportSetting.glade to fix crash

edo0:
As per suggestion of @Technius, I have removed the seemingly incriminated line of code leading to this bug
https://github.com/xournalpp/xournalpp/issues/1285
As a layman in terms of programming, I do not have the skills to build and test this on my machine in a proper way


LittleHuba:
Just removing that line is no good idea as it will break styling with CSS. We'll have to check **why** this fails.

LittleHuba:
Okay seems like the object type `GtkEntryBuffer` does simply not have that attribute. Thus removing might be the only solution there is.

Issue #1236, opened 2019-06-12
[new input] touchscreen as hand tool doesn't work in Windows


cbm755:
**Affects versions :**
 - Windows
 - 1.0.12

**Describe the bug**
After enabling the new input system, I cannot make the touchscreen use the hand tool.  I think I could generally make this work in 1.0.10 (although it wasn't reliable and often started ignoring the touchscreen completely after some use, so I was hoping to experiment with the new input system).

Its not clear (to me) if the Preferences->Touchscreen tab is functional when I've enabled the New Input System.  Or am I supposed to configure inputs in some other way when using the New Input System.

**To Reproduce**
1. enable New Input System
2. Search for way to configure "touchscreen as hand tool" but cannot find.
3. "Touchscreen as hand tool" doesnt work OOTB (instead, it generally uses the selected tool).

**Input Logs**
"Follow this tutorial..."  Unfortunately, I don't know how to do any of this on Windows :(

cbm755:
I got the machine working with Fedora 30.  With git master, the touchscreen works as handtool OOTB.

In preferences->input system, the "Input Devices" section is populated, whereas on Windows (or maybe just on 1.0.12) it says something about a placeholder.

LittleHuba:
This is a duplicate of #1166. I'm currently working on a fix for that. I'll leave this ticket open as it has a more fitting title than #1166 and cares solely about that topic instead of mixing multiple ones.

LittleHuba:
@cbm755 please test the linked PR #1242 if it resolves your issue.

Generally the touchscreen tab stays active even if the new input system is enabled. In the future some functionalities should be incorporated into the input system tab. For example assigning devices as pen when they are recognized as touchscreens is now handled in the input system tab.

LittleHuba:
Please be aware that devices which were previously not listed have to be used once before opening the preferences to be listed.

cbm755:
I'm willing to help but I have no dev skills (on Windows anyway).   I guess #1127 would help if CI runs produced .exe as an artifact... some day.

LittleHuba:
Actually building on windows is not hard at all. Just follow these [instructions](https://github.com/xournalpp/xournalpp/blob/master/readme/WindowsBuild.md).

In the step "Get Sources and Build" 
you will have to use another repository to clone. 
https://github.com/LittleHuba/xournalpp.git 

And checkout  the branch `fix_device_listing` via `git checkout fix_device_listing`

[...]

PR #1318, opened 2019-07-02
Add device listing for devices not recognized by GDK

LittleHuba:
This inherits the changes from #1242. Including the fix to actually list the devices.

@michaelschufi please try it out.
There seems to still be a problem with the automatic mapping of the devices as a pen is currently automatically mapped to a mouse (or this is just Windows for you...)

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -6,6 +6,7 @@
 #include <config.h>
 #include <i18n.h>
 #include <Util.h>
+#include <util/DeviceListHelper.h>
 
 #define DEFAULT_FONT "Sans"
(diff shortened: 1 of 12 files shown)


LittleHuba:
Will ignore Clang Format check as it fails for some reason even if the file is correctly formatted...

cbm755:
I've tested this re: #1236 and it works for me!  GTK3 doesn't see my touchscreen (as [described here](https://github.com/xournalpp/xournalpp/issues/1236#issuecomment-504355971)).  With this PR, I can tell Xournalpp that my "System Aggregated pointer" is a touchscreen (and my pen is a pen), and then the new input system works for me.

Here's a screenshot:

Annotation 2019-07-06 213042

PR #1333, opened 2019-07-09
Generate Linux binaries (in tarballs) on release pipeline

Technius:
This PR adds configuration to generate tarballs for the Linux binary using CPack; the tarballs will be published in the (nightly) release pipeline.


LittleHuba:
Have you checked that running the `install` target still works correctly? You removed the call to `cmake/postinst` which might break it?

CMakeLists.txt
```diff
@@ -4,6 +4,9 @@ project ("Xournal++" CXX C)
 
 ## Also update changelog in debian folder!
 set (PROJECT_VERSION "1.0.13")
+set (CPACK_PACKAGE_VERSION_MAJOR "1")
```
TheAssassin: I guess `PROJECT_VERSION` should be generated from the other three values.

CMakeLists.txt
```diff
@@ -259,7 +260,10 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	install(FILES desktop/xournalpp.thumbnailer
 			DESTINATION ${CMAKE_INSTALL_PREFIX}/share/thumbnailers)
 
-	install(CODE "execute_process(COMMAND ${CMAKE_CURRENT_BINARY_DIR}/cmake/postinst configure)")
+	# TODO: Only package post install scripts with distro-specific packages
```
TheAssassin: Could be achieved with `CPACK_GENERATOR` if CPack is only ever used when a generator is specified explicitly (not sure how to handle the other case, we just use it that way).
See https://cmake.org/cmake/help/v3.0/module/CPack.html.

CMakeLists.txt
LittleHuba: Before doing that we would need to change the release pipeline as it currently parses the project version variable for determining the correct tag.

TheAssassin: You can fetch the calculated value from `CMakeCache.txt` after CMake has run, if that's possible.

LittleHuba: The problem is that the release pipeline never sees a version of the repository where `cmake` ran. It only gets the build artifacts of our pipeline "release" and parses the `CMakeLists.txt` directly from the github repository.
But it shouldn't be a problem to parse the three `CPACK_PACKAGE_VERSION` variables and concat them.

TheAssassin: The other way round is also possible.

I just intended to highlight the redundancy, which could in the end cause inconsistencies.

LittleHuba: Yeah the redundancy promises issues later on. Whichever way we decide we only need to keep the release pipeline in mind.

CMakeLists.txt
LittleHuba: `make install` should definitely include this command. Can we rewrite `postinst.in` to be distro-specific?

CMakeLists.txt
Technius: Where exactly does this version parsing happen?

Technius:
> Have you checked that running the `install` target still works correctly? You removed the call to `cmake/postinst` which might break it?

The main issue here is that CPack uses the `install` target to package binaries. This means that if we execute the postinstall script in the install step, it will trigger the postinstall script on the _computer doing the packaging_. This can cause undesirable problems.

The `install` target will still properly copy files, but it might not run the post configure script anymore. I think that if the user has already gone through the trouble to compile the program manually, it wouldn't be too much additional trouble to manually run the post-configure scripts (we just need to include this in the build/install instructions).

LittleHuba:
Thanks for the insights @Technius! We should modify the build instructions to use AppImages/flatpaks then. I'm not a big fan of doing manual installs but to test new translations for example you currently have to go that way as only starting the application without installing will use the old translation files.

But flatpak or AppImage should solve that too with the additional benefit of not messing up the system..

TheAssassin:
AppImages are used by LibreOffice for instance for testing daily builds, see also https://wiki.documentfoundation.org/QA/Testing_Daily_Builds.

LittleHuba:
So what we should do before merging this:

- [ ] Clean up the redundancy in `CMakeLists.txt` mentioned by @TheAssassin 
- [ ] Update the Release Pipeline to match the changes made in `CMakeLists.txt`
- [ ] Update the manual build instructions to either use flatpak or AppImage

CMakeLists.txt
LittleHuba: Check out the task `Set Variables` in https://dev.azure.com/xournalpp/xournalpp/_releaseDefinition?definitionId=1&_a=definition-tasks&environmentId=6

Technius: One simple solution is to upload a file containing the version number (generated by cmake) as an artifact, read it in the nightly pipeline, and then delete it before making the release.

By the way, it should be possible to define the nightly pipeline with a YAML file (we can probably do that in another PR).

LittleHuba: That is actually just one step more. Parsing it from a generated file is as hard as parsing it from the `CMakeLists.txt` but we don't have to care about that file.

TheAssassin: The place where version information is defined doesn't matter, as you will always have to parse something. But thinking about this, I now consider defining the version in an extra file for my own applications as well. Conceptually it seems like a good idea.

CMakeLists.txt
TheAssassin: Should `make install` automatically run this for *every* installation? Doesn't seem so. It shall be included in the final package, I guess.

AppImage creation uses `make install` as well, and it doesn't need the script at all. A tarball doesn't need it either I guess. It's only useful to native packages.

For AppImage and tarball, the script is therefore bloat. But that said, just packaging it doesn't harm anyone. Those few extra bytes don't hurt IMO.

In [AppImageLauncher](https://github.com/TheAssassin/AppImageLauncher/) I solved the issue by introducing [an extra variable](https://github.com/TheAssassin/AppImageLauncher/blob/master/resources/CMakeLists.txt#L20), so it's the duty of the pipeline job or packager to switch that on if the scripts are needed.

Technius:
I've updated the PR so that cmake will generate a `VERSION` file in the build directory; the file is uploaded as an artifact in pipelines and can be retrieved using a [download task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact?view=azure-devops). We can easily read in the relevant information from the file, e.g.

```bash
VER=$(cat VERSION | sed '1q;d')
TIMESTAMP=$(cat VERSION | sed '2q;d')
GIT_REV=$(cat VERSION | sed '3q;d')
```

Note that we can't update the nightly release pipeline until this PR is merged or else existing builds will break.

Technius:
I'll resolve the post install script situation in another PR (probably when I add `deb` packaging).

If there are no objections, tomorrow night I will merge this and update the release pipeline.

Technius:
Sorry for the delay, was busy last night.

I merged in the changes and updated the nightly pipeline (took way longer to do than I wanted, but whatever). The tar file should be uploaded now and with a sensible name; it should be simple to add more types of packages.

PR #1334, opened 2019-07-09
Improve desktop entry categories

TheAssassin:
At the moment, Xournal++ is listed in "Utility" on most desktops, which is IMO not a suitable location. I think it belongs to "Office". Unfortunately "TextEditor" has been specified to belong to "Utility", therefore it must be removed. A replacement might be "WordProcessor", which belongs to "Office", but on the other hand a top level category should be sufficient.

See also https://specifications.freedesktop.org/menu-spec/latest/apas02.html, https://specifications.freedesktop.org/menu-spec/latest/apa.html.

### DIFF
--- a/desktop/xournalpp.desktop
+++ b/desktop/xournalpp.desktop
@@ -20,4 +20,4 @@ Terminal=false
 StartupNotify=true
 MimeType=application/x-xoj;application/x-xojpp;application/x-xopp;application/x-xopt;application/pdf;
 Icon=xournalpp
-Categories=GNOME;GTK;Utility;TextEditor;
+Categories=Office;GNOME;GTK;


LittleHuba:
Seems like a good idea to get it into the office category. I would not search it in utilities. (But KDE has autosearch on typing so I never search manually for it 😉)

I can't find any other tool that would do a better mapping than top level category...
@andreasb242 what do you think about this mapping?

Technius:
I would also add the `Graphics` category since Xournal++ can be used to create posters, diagrams, drawings, etc.

TheAssassin:
If you check out the detailed list, you'll see that `Graphics` is more for drawing applications. There's subcategories for `2DGraphics` and `VectorGraphics` that fit to some extent, technically. But I guess these categories apply more to apps like Krita.

None of the subcategories that are shared with `Office` seems to fit (Xournal++ is not about photography or publishing), maybe `Viewer` might fit.

`Graphics` as a top level category might fit if we also add `Viewer`.

LittleHuba:
Yeah `Viewer` should also fit. We can display PDFs and Xournals after all.

andreasb242:
For me this looks good. Thank you.

PR #1337, opened 2019-07-09
NULL-terminate argument vector given to kpsewhich.

dermesser:
Otherwise glib will provide a (likely spurious) last element in argv,
causing EFAULT to be returned from execve in the child:

```
...
[pid 32237] execve("/usr/local/bin/kpsewhich", ["kpsewhich", "standalone", 0x1], 0x7ffe7141a078 /* 69 vars */) = -1 ENOENT (No such file or directory)
[pid 32237] execve("/usr/bin/kpsewhich", ["kpsewhich", "standalone", 0x1], 0x7ffe7141a078 /* 69 vars */) = -1 EFAULT (Bad address)
...
```

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -86,7 +86,7 @@ LatexController::FindDependencyStatus LatexController::findTexDependencies()
 	g_free(pdflatex);
 
 	// Check for 'standalone' latex package
-	static gchar* kpsewhichArgs[] = {g_strdup("kpsewhich"), g_strdup("standalone")};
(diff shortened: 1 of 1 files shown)


dermesser:
FWIW this completely prevents using LaTeX in xournal++ right now, if I understand correctly.

dermesser:
(I also want to note that I just recently found xournal++ as a previous user of xournal, and I want to say THANK YOU for making such an awesome piece of software!)

LittleHuba:
Thanks for the fix! Glib did not mention this peculiarity in a good way. Would have not seen it myself if I had not looked for this.

PR #1353, opened 2019-07-16
Fix & Enable Mac CI builds

LittleHuba:


LittleHuba:
Linking is successful but checks fail because of ambiguous use of abs.
I'll fix the errors in a separate PR as they are not part of enabling the MacOS pipeline.

PR #1359, opened 2019-07-17
update flatpak manifest

muelli:
this is in preparation for flathub proposal. I would like the flathub manifest to be as close to the upstream maintained manifest as possible. These proposed  changes achieve a closer relationship.

### DIFF
--- a/com.github.xournalpp.xournalpp.yaml
+++ b/com.github.xournalpp.xournalpp.yaml
@@ -1,4 +1,3 @@
----
 app-id: com.github.xournalpp.xournalpp
 runtime: org.gnome.Platform
 runtime-version: '3.32'
@@ -14,6 +13,7 @@ finish-args:
   # Wayland access
   - "--socket=wayland"
   # Needs to save files locally
+  # Why are portals not used? Does the app re-implement the open dialogue?
   - "--filesystem=xdg-documents"
   - "--socket=pulseaudio"
   # Allow dconf
@@ -22,7 +22,7 @@ finish-args:
   - "--talk-name=ca.desrt.dconf"
   - "--env=DCONF_USER_CONFIG_DIR=.config/dconf"
   - "--persist=.xournalpp"
-build-options:
+
 cleanup:
   - "/include"
   - "/lib/pkgconfig"
@@ -57,10 +57,14 @@ modules:
       - type: archive
         url: https://libzip.org/download/libzip-1.5.2.tar.xz
         sha256: b3de4d4bd49a01e0cab3507fc163f88e1651695b6b9cb25ad174dbe319d4a3b4
+
   - name: libportaudiocpp
     buildsystem: autotools
     config-opts:
       - "--enable-cxx" # compile with c++ headers!!!
+    make-args:
+        # seem parallel build is broken. At least it breaks for me, locally
+        - -j1
     sources:
       - type: git
         url: https://git.assembla.com/portaudio.git
@@ -76,6 +80,7 @@ modules:
       - "/share/gtk-doc"
       - "/share/man"
       - "/share/pkgconfig"
+
   - name: libsndfile
     buildsystem: autotools
     sources:
@@ -90,7 +95,6 @@ modules:
 
   - name: xournallpp
     buildsystem: cmake-ninja
-    config-opts: []
     sources:
       - type: git
         path: .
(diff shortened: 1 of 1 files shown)


ezickler:
I wrote the original version of the manifest and also sometimes experienced build failures with portaudio, but i hadn't figured out that it could be the parallel build. Seams better now. 

This was one of my first flatpak manifest, so i didn't know about the portal option for the file chooser. So feel free to change it to better solutions.

muelli:
> I wrote the original version of the manifest and also sometimes experienced build failures with portaudio, but i hadn't figured out that it could be the parallel build. Seams better now.

cool :)

> 
> This was one of my first flatpak manifest, so i didn't know about the portal option for the file chooser. So feel free to change it to better solutions.

It depends on the code. If the native GtkFilechooser is used, it should be able to access all files on the host. Does the app use a custom file chooser?

ezickler:
I tested this without the ` - "--filesystem=xdg-documents" ` an it works fine on my fedora gone desktop. So yes it seams that its the default gtk file chooser

alyberty:
The  `"--filesystemy=xdg-documents"` is part of the flatpack [sandboxing system](http://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access).

With this configuration, it only allows access to the documents folder, or if a user opens a file via xdg-open, the file will be made available to the sandbox.

The problem here is that, if the file references an external pdf, the pdf is not made available to the sandbox. This breaks expected behavior.

I'm not sure, what's the best option here. Ideally the sandbox is as restrictive as possible, without interfering in the user experience. But with the whole pdf referencing, I don't see the option to have any restrictions (apart from system directories).

muelli:
if the xournalpp files are not self-contained, i.e. opening a single file through the file chooser does not work, then the sandbox restrictions need to be lifted (as is currently done). Some other apps do indeed expose the full home directory of the user to the sandboxed app. That's not ideal, as you mention. But it does help to preserve the UX.

muelli:
FTR: There is a flathub app now: https://github.com/flathub/com.github.xournalpp.xournalpp
upstream involvement is highly appreciated.

LittleHuba:
> I tested this without the `- "--filesystem=xdg-documents"` an it works fine on my fedora gone desktop. So yes it seams that its the default gtk file chooser

Yes we are using the native GTK file chooser. Is this ready to be merged or are there changes pending?

muelli:
> Yes we are using the native GTK file chooser.

right, but it seems as if xournalpp files can reference further files. In that case, those additional files do not come from the file chooser.

>  Is this ready to be merged or are there changes pending?

yeah, I think this file is good enough for now.

LittleHuba:
> right, but it seems as if xournalpp files can reference further files. In that case, those additional files do not come from the file chooser.

Yes and this will never change as it is a behaviour that is at least included in our current file format. The new one will solve this partially but will still allow for external files as there are usecases where it is required to not pack them.

com.github.xournalpp.xournalpp.yaml
```diff
@@ -90,11 +95,9 @@ modules:
```
LittleHuba: Build works with CMake: Take a look at [this script](https://github.com/xournalpp/xournalpp/blob/mac-ci-builds/mac-setup/build-sndfile.sh). Even though it builds for MacOS it does work flawlessly.

com.github.xournalpp.xournalpp.yaml
```diff
@@ -14,6 +13,7 @@ finish-args:
   # Wayland access
   - "--socket=wayland"
   # Needs to save files locally
+  # Why are portals not used? Does the app re-implement the open dialogue?
```
LittleHuba: This should be changed to a description why we require this. E.g. Xournal++ supports references to external files within their fileformat, hence global file access is required.

com.github.xournalpp.xournalpp.yaml
```diff
@@ -57,10 +57,14 @@ modules:
       - type: archive
         url: https://libzip.org/download/libzip-1.5.2.tar.xz
         sha256: b3de4d4bd49a01e0cab3507fc163f88e1651695b6b9cb25ad174dbe319d4a3b4
+
   - name: libportaudiocpp
     buildsystem: autotools
     config-opts:
       - "--enable-cxx" # compile with c++ headers!!!
```
LittleHuba: This comment is also not really descriptive. Something like "Xournal++ uses the C++ headers of PortAudio, so compile them" would be better.

LittleHuba:
@Technius is this one obsolete now as you did the work over at the flathub repo?

Technius:
Yes.

PR #1367, opened 2019-07-21
Fix #1365 by correcting variable types from integer to double

### LINKED ISSUE #1365: Pasted graphic from clipboard is zoomed in
**Affects versions :**
 - OS: Arch Linux
 - Version of Xournal++: 1.0.12, 59f6e91876ccca2ebeaf7ea8dd6899709fb2b333 

**Describe the bug**
When pasting or dragging an image onto the drawing pane it is slightly zoomed in.

**To Reproduce**
1. Paste image

**Expected behavior**
Graphics size should match original image size when drawing area is at 100% zoom.

**Screenshots of Problem**
see #877 for animated gif

lehmanju:
Now images pasted aren't zoomed in anymore because element sizes are now correctly represented by doubles.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -3028,8 +3028,8 @@ void Control::clipboardPasteImage(GdkPixbuf* img)
 	auto image = new Image();
 	image->setImage(img);
 
-	int width = gdk_pixbuf_get_width(img);
-	int height = gdk_pixbuf_get_height(img);
+	auto width = static_cast<double>(gdk_pixbuf_get_width(img)) / settings->getDisplayDpi() * 72.0;
+	auto height = static_cast<double>(gdk_pixbuf_get_height(img)) / settings->getDisplayDpi() * 72.0;
 
 	int pageNr = getCurrentPageNo();
 	if (pageNr == -1)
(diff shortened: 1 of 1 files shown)


Febbe:
Hello Lehmanju,
thank you for your interest in xournal++.
Also thank you for the fix. 

Your code also looks very clean.
I just need to check if the behaviour is still correct.

Febbe:
Ok, the fix worked like intended and it did not changed the behaviour.

src/control/Control.cpp
```diff
@@ -3028,8 +3028,8 @@ void Control::clipboardPasteImage(GdkPixbuf* img)
 	auto image = new Image();
 	image->setImage(img);
 
-	int width = gdk_pixbuf_get_width(img);
-	int height = gdk_pixbuf_get_height(img);
+	auto width = static_cast<double>(gdk_pixbuf_get_width(img)) / settings->getDisplayDpi() * 72.0;
```
Technius: Why `72.0`? This should've been documented in a comment, or the constant should've been placed into an appropriately named `const` variable.

Febbe: I've looked into that, it is used everywhere in our codebase and brings the display dpi into relation to our pageView. Basically we assume a dpi of 72 as 100%.

Febbe: It is better to change this in a slip, but separated from this fix. We can write this out as a simple issue, a beginner can fix.

PR #1371, opened 2019-07-21
Fix coordinate precision to 8 decimal places to reduce file size

### LINKED ISSUE #1277: Too much precision in xopp files.
Low priority.. but do we really need 14 digits of precision in the xopp file?

example page coordinates:
168.71782794981516 503.4808047927973

cutting down to 4 would reduce file size substantially before and after compression.
168.7178 503.4808

Should be a setting ... and it might be good to have it keep accuracy for files loaded with more accuracy.

siliconninja:
Fixes #1277.
If you decide to close this PR because you decide not to have fixed coordinate precision, that is OK with me. I'm just opening this pull request since the code is complete, and it's ready for review.

### DIFF
--- a/src/util/Util.cpp
+++ b/src/util/Util.cpp
@@ -170,3 +170,8 @@ gboolean Util::paintBackgroundWhite(GtkWidget* widget, cairo_t* cr, void*)
 	return false;
 }
 
+gchar* Util::getCoordinateString(gchar* buff, gulong buffLen, double xVal, double yVal)
+{
+	g_snprintf(buff, buffLen, Util::PRECISION_FORMAT_STRING_XY, xVal, yVal);
+	return buff;
+}
--- a/src/util/Util.h
+++ b/src/util/Util.h
@@ -53,6 +53,19 @@ void execInUiThread(std::function<void()>&& callback);
 
 gboolean paintBackgroundWhite(GtkWidget* widget, cairo_t* cr, void* unused);
 
+/**
+ * Format coordinates to use 8 digits of precision https://m.xkcd.com/2170/
+ * This function works like g_ascii_formatd in that it stores the result in buff, and
+ * also returns the result.
+ */
+extern gchar* getCoordinateString(gchar* buff, gulong buffLen, double xVal, double yVal);
+
+constexpr const gchar* PRECISION_FORMAT_STRING = "%.8f";
+
+constexpr const gchar* PRECISION_FORMAT_STRING_XY = "%.8f %.8f";  // note the space delimiter
+
+constexpr const int PRECISION_FORMAT_BUFF_LEN = G_ASCII_DTOSTR_BUF_SIZE * 2 + 1;
+
 }  // namespace Util
 
 static const size_t npos = std::numeric_limits<size_t>::max();
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -780,7 +780,8 @@ xmlNodePtr Settings::savePropertyDouble(const gchar* key, double value, xmlNodeP
 	XOJ_CHECK_TYPE(Settings);
 
 	char text[G_ASCII_DTOSTR_BUF_SIZE];
-	g_ascii_dtostr( text, G_ASCII_DTOSTR_BUF_SIZE, value);	//  g_ascii_ version uses C locale always.
+	//  g_ascii_ version uses C locale always.
+	g_ascii_formatd(text, G_ASCII_DTOSTR_BUF_SIZE, Util::PRECISION_FORMAT_STRING, value);
 	xmlNodePtr xmlNode = saveProperty(key, text, parent);
 	return xmlNode;
 }
@@ -1043,7 +1044,8 @@ void Settings::save()
 	xmlSetProp(xmlFont, (const xmlChar*) "font", (const xmlChar*) this->font.getName().c_str());
 
 	char sSize[G_ASCII_DTOSTR_BUF_SIZE];
-	g_ascii_dtostr( sSize, G_ASCII_DTOSTR_BUF_SIZE, this->font.getSize());	//no locale
+
+	g_ascii_formatd(sSize, G_ASCII_DTOSTR_BUF_SIZE, Util::PRECISION_FORMAT_STRING, this->font.getSize());  // no locale
 	xmlSetProp(xmlFont, (const xmlChar*) "size", (const xmlChar*) sSize);
 
 
@@ -1100,7 +1102,7 @@ void Settings::saveData(xmlNodePtr root, string name, SElement& elem)
 			type = "double";
 
 			char tmp[G_ASCII_DTOSTR_BUF_SIZE];
-			g_ascii_dtostr( tmp, G_ASCII_DTOSTR_BUF_SIZE, attrib.dValue);
+			g_ascii_formatd(tmp, G_ASCII_DTOSTR_BUF_SIZE, Util::PRECISION_FORMAT_STRING, attrib.dValue);
 			value = tmp;
 		}
 		else if (attrib.type == ATTRIBUTE_TYPE_INT_HEX)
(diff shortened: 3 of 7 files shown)


src/util/Util.cpp
```diff
@@ -165,3 +166,17 @@ gboolean Util::paintBackgroundWhite(GtkWidget* widget, cairo_t* cr, void* unused
...
 }
 
+char* Util::getCoordinateString(double xVal, double yVal)
+{
+	char tmpX[G_ASCII_DTOSTR_BUF_SIZE];
+	g_ascii_formatd(tmpX, G_ASCII_DTOSTR_BUF_SIZE, Util::getFormatdPrecisionString(), xVal);
+	char tmpY[G_ASCII_DTOSTR_BUF_SIZE];
+	g_ascii_formatd(tmpY, G_ASCII_DTOSTR_BUF_SIZE, Util::getFormatdPrecisionString(), yVal);
+
+	return g_strdup_printf("%s %s", tmpX, tmpY);
```
JJones780: g_strdup_printf allocates memory that needs to be freed.  Our function should also mention this in a doxygen type comment block ( some IDE's will pop this up on mouseover).

**Better yet,**   use g_snprintf () instead and require the caller to allocate and pass us a buffer to use
 ( so that it becomes `getCoordinateString(gchar*buff, gulong buffLen, double xVal, double yVal) `    ) 
I see two benefits to using g_snprintf:  1) no hidden allocations, 2) Can re-use the buffer when calling this function repeatedly in a loop ( _although I see the original code misses this optimization.. maybe the compiler did it ;)_  )

One could also consider using a c++ object that free's in the destructor... but I think that's overkill here.

src/util/Util.cpp
```diff
@@ -165,3 +166,17 @@ gboolean Util::paintBackgroundWhite(GtkWidget* widget, cairo_t* cr, void* unused
...
+}
+
+char* Util::getFormatdPrecisionString()
+{
+	return g_strdup_printf("%%.%df", PRECISION_DECIMAL_PLACES);
```
JJones780: Avoid rebuilding this format string every time.
The only use of  PRECISON_DECIMAL_PLACES seems to be to build this string.  So it makes sense to define a char* constant instead.

Declare it in the header but you have to initialize it here:
`const char* Util::PRECISION_FORMAT_STRING = "%.8f";   //use 8 digits of precision   https://m.xkcd.com/2170/ `

src/util/Util.h
```diff
@@ -56,7 +56,12 @@ class Util
...
 
+	static char* getCoordinateString(double xVal, double yVal);
+	
+	static char* getFormatdPrecisionString();
 };
 
 static const size_t size_t_npos = static_cast<size_t>(-1);
 // for 64b systems it's 18446744073709551615 and for 32b – 4294967295
+
+static const int PRECISION_DECIMAL_PLACES = 8;
```
JJones780: Intead of int, I suggest you create a 
`static const char* PRECISION_FORMAT_STRING;`
and initialize it in .cpp

siliconninja:
Okay. I understand -- the change is so small that the commits could be combined into one, and trying/reverting ideas isn't worth including in the PR.

I will combine the commits later today. I would've replied earlier but GitHub was down.

siliconninja:
I will rebase and then squash the earlier commits after I have made those changes in the code review since that would be easier. I agree with the proposed changes -- for the first one, I think I'll just use `g_snprintf` because of those benefits that you mentioned.

src/util/Util.cpp
siliconninja: I plan to use `g_snprintf` because of those benefits that you mentioned.

src/util/Util.cpp
siliconninja: The xkcd link in the comment is also a nice easter egg :smile:

siliconninja:
I have made those changes and fixed the formatting with clang-format.

JJones780:
Oops, Do you have the non-squashed version still?  The clang format makes it very hard for someone to see what was done to implement the feature - either to learn from, review or to do regression testing.  You should only apply clang-format to lines you've changed ( using git-clang-format - see [ Xournal++ wiki on clang-format ](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#use-clang-format)) or for cases where you've made a lot of changes to one file do the clang-format as a separate commit.  Once we get the whole project reformatted with clang this shouldn't be an issue.
 If you've lost the pre-clang format version let me know and I can help.

siliconninja:
Yes, I have the non-squashed version. I saved it on a different branch in my fork called `issue_1277_archive`. I have run that command (`diff -U0 --no-color 583a317ff1d63b8d65d4019b75fb27801e5c0098 HEAD | clang-format-diff-8 -i -p1`) that you linked to in the documentation. I made further commits on that branch just in case I wanted to look at them later. I've now squashed my fork's `issue_1277` branch after I ran that command.

siliconninja:
CI has failed :(
I think there were more merge conflicts or extra code that didn't get added in properly.
I tried rebasing my `issue_1277_archive` branch and it's a mess of merge conflicts. In this case I think it's just easier to merge the final commit with upstream/master and resolve the conflicts on just that commit, then squash that onto my other `issue_1277` branch.
I'll try to get back to a working build later today. (I live in the US, so my timezone is very different from yours)

siliconninja:
Don't worry about the last comment, I was able to get everything up to date and built properly :relieved: 
Is this ready to be merged in?

siliconninja:
Btw, did a quick micro-note test w/ 8 decimal place precision. Everything seems good.

Output of `compare -verbose -metric PSNR 2019-07-24-Note-21-59.png 2019-07-24-Note-21-59-8dp.png diff.png`:

diff
Note how there's no dark red at the top (even though the ink is in red).
The only difference is at the bottom which I think is the new version of the page template in a newer version of Xournal++.

JJones780:
Hi @siliconninja ,

  I've been away lately.

  This is looking great. Very clean and understandable.

 What do you think about making getCoordinateString work the same way as g_ascii_formatd ( i.e. passing in a buffer to fill )?  It would let you re-use already allocated scratch buffer space and also use stack-allocated memory which we won't need to ( worry about it not being ) freed manually.
i.e. `getCoordinateString(gchar*buff, gulong buffLen, double xVal, double yVal) `


Thanks for doing the detective work on this. It's great to know we can shave off those digits without affecting image quality one bit ( pun intended ).

siliconninja:
Hi @JJones780,

Thank you! I think that making getCoordinateString like g_ascii_formatd is a good idea. I'll start working on that later today because I don't have much time right now.

siliconninja:
Alrighty, I've made those changes, and I'm ready for another look.

JJones780:
Looks good. You just missed one call to g_free after the allocation by g_strconcat().
But instead.... 
I'd recommend you just define another format string for '%.8f %.8f'.  _It avoids running code to rebuild that format string repeatedly._ I understand wanting to keep to one format string definition but in this case the second can be right beneath and is straightforward and easily maintained.

You can also do stuff like:
```
#define PFS "%.8f"
// then using it like:
constexpr const gchar* PRECISION_FORMAT_STRING    = PFS;
constexpr const gchar* PRECISION_FORMAT_STRING_XY = PFS " " PFS;
```
but as suggested, I'd just stick with :
```
constexpr const gchar* PRECISION_FORMAT_STRING    =  "%.8f";
constexpr const gchar* PRECISION_FORMAT_STRING_XY =  "%.8f %.8f";  //note space delimeter
```

siliconninja:
Thanks! I've made those changes. I agree, that makes the most sense. It's better not to rebuild the string and just have some duplication in this case.
Off-topic: The tip for concatenating strings using `#define` and spaces is pretty cool - I hadn't seen that before.

JJones780:
Looks great.
A few extra includes in Util.cpp leftover but that won't cause any problems.
I (or someone ) will be merging this soon... just heading out the door now though.

siliconninja:
To whoever's merging -- I've removed those imports (it still compiles fine without them). I forgot to do that, I had it planned though, thanks 🙂

JJones780:
Thanks for the code and sharing the research you did.

Issue #1350, opened 2019-07-16
Hand-drawn signature is pasted at the bottom of the document


yurivict:
I selected it, hit "copy", but the "paste" operation pastes it at the very bottom of the document, and not at the "current", visible location as expected.

Technius:
Related to #877 and #1165.

lehmanju:
@yurivict where do you copy your signature from? did you write it in xournal or is it a graphic from outside the programm? Could you describe your steps to reproduce this? Maybe try #1372 and check whether this fixes it.

yurivict:
> where do you copy your signature from

From another instance of ```xournalpp```. It was hand-drawn there.

yurivict:
> Maybe try #1372 and check whether this fixes it.

1.0.13-nightly.20190722 still has the problem.


> Could you describe your steps to reproduce this?

1. Draw something by hand in one document.
2. Select the drawing and copy it into the clipboard.
3. Open another instance of xournalpp
4. Paste into the middle of the page
5. Observe that it was pasted at the bottom of the page (or maybe it pastes at the exact same location it was in the original).

Technius:
@yurivict Nightly builds don't include PR changes, you'll have to compile #1372 manually. You can do so by cloning the PR repository ([instructions here](https://github.com/xournalpp/xournalpp/wiki/Installing#building-a-pull-request)) and following the build instructions for your platform (e.g. in the readme folder at the repository root)

yurivict:
#1372 improved the behavior, now it pastes exactly at the center of the screen.

IMO, it would be better if it would paste at the cursor location.

Issue #1184, opened 2019-05-29
Active page does not update when scrolling


Debrugger:
Scenario: open up a long pdf, scroll and read along with the lecture or searching for a passage, not writing anything yet. If you insert a page or change the display mode to two pages next to each other, the view jumps back to the active/outlined page. This behavior is very frustrating and unintuitive, especially when there's no clear reason for the 'active page' mechanism.

**Describe the solution you'd like**
Change the active page to the one currently visible when scrolling. It's not necessarily clear which page should be active when e.g. multiple are on screen, but anything is better than to suddenly be catapulted back 20 pages.

**Describe alternatives you've considered**
None

**Additional context**
None

LittleHuba:
The fix for this issue should consist of two separate things:

1. Change the active page when scrolling (this is required for backward compatibility with the old input system)
2. Remove the active page system as the new input system does not require it. When doing this we need to determine the position for pasted items.

To determine the page for pasting items I would use the following steps:

1. Use the current position of the cursor if it is within a page
2. Use the page that received the last input, if it is still visible
3. Use the currently visible page

For 2. and 3. we need to take the visible area of the page into account. We did this already but it got broken by the zooming changes. There should be another issue somewhere concerning this regression.

Technius:
Pinning this issue since it gets a lot of duplicates.

Issue #1165, opened 2019-05-21
Move pasted objects for clarity


michaelschufi:
The first time I copy-pasted an object in Xournal++ I didn't notice that the pasted object is right over the original one.

Therefore, I suggest moving the pasted object a bit, so it is clear the object got pasted.

What would be even better is to paste the object where the cursor is. (maybe make this behaviour optional).

sotte:
Just a small offset would be great!

Technius:
Current behavior is to paste in the center of the page. To accommodate this change, we'll need to add an option to the settings dialog that will make the paste behavior configurable.

LittleHuba:
I would not add another option for this. Pasting in the center of the page makes some planned projects very hard to solve (infinite page).
The proposed behaviour is intuitive and should suffice.

rolandlo:
For geometric constructions it is essential that the selection is pasted exactly where it was. So I would not introduce even a little offset. I suggest closing this issue.

Febbe:
Currently it is still pasted at the center of the page right? So closing this issue isn't the solution. We should implement the pasting behavior, so it pastes it either at the same position, if the position is in the view, otherwise at the center of the cursor/view

rolandlo:
Ok, there are two different ways to copy+paste:
1) Copy and then paste from the menu (or via Ctrl+c, Ctrl+v) will paste the object in the center of the page.
2) Copy and paste (in one step) via right mouse button (or stylus button 2, if you have one) at the same position of the selection.

I think 2) should remain as it is, since it is essential for geometric constructions. 

About 1): I only use this when I paste an object from a different page or from a different application. In that case there is no predefined place where to paste the object. Pasting it in the center of the page (or in the center of the current view, if it should be compatible with infinite pages) makes sense to me. Pasting at the cursor position makes also sense, but it has the problem, that if you paste from the menu you will have changed the position of the cursor. As a configurable option it may be fine.

TentativeConvert:
I'm using xournalpp to give online lectures, and for this I frequently paste material into xournalpp that I've prepared in advance.  So I'm copy-pasting from a different instance of xournalpp, or even from a different application.  By pasting at screen center or page center, I am almost always pasting on top of something that's already on the page. This is annoying for the viewers.  

[...]

PR #1372, opened 2019-07-22
Fix #877 by copying pasting behavior other parts

### LINKED ISSUE #877: Paste form clipboard can land outside of the viewport
When you paste with CTFL-V, if you are zoomed in, the pasted material (I checked with images) can land outside of the viewport, which is quite disconcerting (I pasted 21 times the same images before noticing...) 

recorded

lehmanju:
This fixes #877 by implementing the same pasting behavior used when pasting images/text/etc. from outside xournal (new generated xournal Elements). When pasting a selection though different behavior was used.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -3185,6 +3185,23 @@ void Control::clipboardPasteXournal(ObjectInputStream& in)
 		}
 		undoRedo->addUndoAction(std::move(pasteAddUndoAction));
 
+		double x = 0;
+		double y = 0;
+		//calculate x/y of paste target, see clipboardPaste(Element* e)
+		win->getXournal()->getPasteTarget(x, y);
+
+		x = std::max(0.0, x - selection->getWidth() / 2);
+		y = std::max(0.0, y - selection->getHeight() / 2);
+
+		//calculate difference between current selection position and destination
+		auto dx = selection->getXOnView() - x;
+		auto dy = selection->getYOnView() - y;
+
+		//for some reason selection moving is inverted (x -= dx,...), intended?
+		selection->moveSelection(dx, dy);
+		//update all Elements (same procedure as moving a element selection by hand and releasing the mouse button)
+		selection->mouseUp();
+
 		win->getXournal()->setSelection(selection);
 	}
 	catch (std::exception& e)


yurivict:
IMO, it would be better if it would paste at the cursor location, not at the center.

Technius:
Why not both? We can make this an option in the preferences.

Technius:
Instead of copying and pasting, can you refactor the logic into a separate function and use that in both places? Otherwise a future change to one part could make the behavior inconsistent.

yurivict:
I agree.

lehmanju:
@Technius basically the logic itself is already in a separate function, it just hasn't been used in this instance yet.

Regarding your suggestion to make this optional in settings I would recommend to open a new issue as that would be an enhancement and this is basically just a fix. I'd rather not mix those up ;)

src/control/Control.cpp
```diff
@@ -3185,7 +3185,24 @@ void Control::clipboardPasteXournal(ObjectInputStream& in)
 		}
 		undoRedo->addUndoAction(std::move(pasteAddUndoAction));
 
-		win->getXournal()->setSelection(selection);
+		double x = 0;
+		double y = 0;
+        //calculate x/y of paste target, see clipboardPaste(Element* e)
+		win->getXournal()->getPasteTarget(x, y);
+
+        x = MAX(0, x - selection->getWidth() / 2);
```
Febbe: Please use std::max and std::min in newly written code

src/control/Control.cpp
```diff
@@ -3185,7 +3185,24 @@ void Control::clipboardPasteXournal(ObjectInputStream& in)
...
+        y = MAX(0, y - selection->getHeight() / 2);
```
Febbe: Please use std::max and std::min in newly written code

Technius:
Tested the changes just now; it works well as long as you're on the same page. Unfortunately, there's some edge cases involving scrolling to a different page that can't be fixed until we implement #1184. I'd like to mention #1165 as that issue is still present even with this fix (but we're making progress!). 

Thanks for the PR!

Febbe:
Fixes #1350

PR #1380, opened 2019-07-27
Render application icon in xopp thumbnail previews

### LINKED ISSUE #1206: Add an icon to the thumbnailer
When you have several PDF files you are editing with `xournalpp`, it is difficult to say at a glance which files are PDF and which ones are `.xopp`:

image

Add a small icon into the thumbnail: 

icon

I know you can do it with a script (I have several, for raw image and for gimp; for example for raw images I do something like this: 
```bash
#!/bin/bash

work_dir=$(mktemp -d)

# check if tmp dir was created
if [[ ! "$work_dir" || ! -d "$work_dir" ]]; then
  echo "Could not create temp dir"
  exit 1
fi
cd $work_dir

# remove file if something strange happens
# trap 'rm -rf $work_dir' INT EXIT TERM

die() {
  echo >&2 "$@"
  rm -rf $tmpfile1
  exit 1
}

# The script will be called with parameters : -s %s %u %o
# -s %s: vertical size,  %u: input url, %o: output file

if [ $1 = "-s" ]; then
   size=$2
   uin=$3
   fout=$4
else
   size=32 #default size
   uin=$1
   fout=$2
fi

# uin es una url, eliminar "file://"
fin="${uin#file://*}"

# dcraw uses the same name / location of the raw file, so trick it
ln -s $fin . || die "link failed" 
# this will create a file with "*.thumb.jpg"
/usr/bin/dcraw -e $(basename $fin) || die "dcraw failed"
convert -auto-orient *.thumb.jpg -resize ${size}x${size} tmp1.png


composite -gravity southeast  \
   \(  -resize $((size/2))x$((size/2)) /usr/share/icons/raw.png \) \
   tmp1.png  "$fout" || die "convert failed"

# rm -rf $work_dir
exit 0
```

But it's complex, it needs to be reinstalled a lot of times, and sometime it will clash with the security settings (see for example https://askubuntu.com/questions/1088539/custom-thumbnailers-don-t-work-on-ubuntu-18-10-and-18-04). Having it embedded by default would be really handy.

Technius:
This PR makes the thumbnailer render the Xournal++ icon on top of previews (closes #1206).

It's still a work-in-progress because we need to refactor the resource search function `XournalMain::initResourcePath` into the `util` library so that the icon can be loaded. Also, the code is fairly ugly because I'm very new to cairo (suggestions to clean up the code are welcome--in particular, I'd like any suggestions to clean up the image loading code).


Technius:
Decided to just load the icon from the icon theme, since it should already be installed alongside the thumbnailer.

Rebased and ready for review.

src/xoj-preview-extractor/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ target_link_libraries (xournal-thumbnailer
   ${ZLIB_LIBRARIES}
   ${Glib_LDFLAGS}
   ${ZIP_LDFLAGS}
+  ${GTK_LDFLAGS}
```
LittleHuba: You will have to clean up the ld flags for macOS.
Here is how it is done:
```cmake
#
# DO NOT INCLUDE LIBRARIES WITH pkg_check_modules AFTER THIS LINE!!!
#
if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
    # Fix linker flag handling for MacOS builds where frameworks are separated from their flag
    string(REPLACE "-framework;" "-framework " xournalpp_LDFLAGS "${xournalpp_LDFLAGS}")
endif()
```

You could change this to a macro like add_include_ldflags and call it for all the LDFLAGS linked to the target.

LittleHuba: You will have to do so for all LDFLAGS and you should overwrite the variable as linking against it will lead to a failure anyway. Having the original variables available is actually bad as it can lead to errors.

Technius: First build error was because of the `-framework` argument handling. Second build failure was because I made a typo in `thumbnailer_GTK_LDFLAGS`. It builds now.

lehmanju:
Is this getting ready for 1.0.14 or does it need more time?

Technius:
Merging in 2 days if there are no objections.

PR #1381, opened 2019-07-27
Removed types header

Febbe:
Test compile time of xournalpp without XournalType.h PP Makros


Febbe:
The Microsoft CI is at least 1 minute faster, but the incremental build time is much faster.

Febbe:
This is finished, but we should wait to merge this until we freeze all PR's

Febbe:
Done in #1488

Issue #1101, opened 2019-04-28
Respect the XDG base directory specification on Linux


Technius:
The files in `.xournalpp` should be relocated to the appropriate locations specified by the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) (e.g. config files go in `~/.config/xournalpp/` and data files go in `~/.local/share/xournalpp`) on Linux.

Reasons:
* It makes it easier to synchronize dotfiles across multiple devices.
* One fewer dotfile in the home directory.
* Makes it easier to find the configuration files for new users (search in the standard location `~/.config/` instead of having to look it up online or run `find / -name '*xournalpp*'`)
* This standard is supported by many Linux programs and tools (e.g. Python, LibreOffice, NeoVim).

andreasb242:
This is correct, this should be cahnged, but the files should be moved on start, so that existing installations are migrated.

cbm755:
Maybe this project is young enough that us users would forgive you for not migrating such files, esp. if you had a prominent notice on the releases page.

LittleHuba:
We could add the moving algorithm for 2 releases and remove it afterwards. That would ensure that the biggest user community would not loose settings and we can remove the complexity afterwards.

avidseeker:
Just as a note, to fully comply with XDG standard:
* `$XDG_CONFIG_HOME/xournalpp/metadata` should be in `$XDG_STATE_HOME/xournalpp/metadata`
* `$XDG_CONFIG_HOME/xournalpp/plugins` should be in `$XDG_DATA_HOME/xournalpp/plugins`

PR #1384, opened 2019-07-30
Respect the XDG base directory specification

sikmir:
Fix https://github.com/xournalpp/xournalpp/issues/1101:
- [x] Removed build-time option `DEV_CONFIG_DIR` (run-time `XDG_CONFIG_HOME` env variable can be used instead)
- [x] Use `g_get_user_config_dir()` instead of `g_get_home_dir()` for building configuration file path
- [x] Tested on Linux
- [x] Migration (do we really need it?)
- [ ] Load custom plugins from `$XDG_DATA_HOME/xournalpp/plugins`
- [x] Save errorlogs to `$XDG_CACHE_HOME/xournalpp/errorlogs` (as far as these are "user specific non-essential data files")
- [ ] Test on Windows and macOS

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -6,6 +6,7 @@
 #include <gtk/gtk.h>
 #include <libintl.h>
 
+#include "filesystem.h"
 #include "control/jobs/ImageExport.h"
 #include "control/jobs/ProgressListener.h"
 #include "gui/GladeSearchpath.h"
@@ -69,6 +70,21 @@ void XournalMain::initLocalisation() {
     }
 #endif
     std::cout.imbue(std::locale());
+}
+
+void XournalMain::migrateSettings() {
+    Path newConfigPath(g_get_user_config_dir());
+    newConfigPath /= g_get_prgname();
+
+    if (!newConfigPath.exists()) {
+        Path oldConfigPath(g_get_home_dir());
+        oldConfigPath /= ".xournalpp";
+
+        if (oldConfigPath.exists()) {
+            fs::create_directories(newConfigPath.str());
+            fs::copy(oldConfigPath.str(), newConfigPath.str(), fs::copy_options::recursive);
+        }
+    }
 }
 
 void XournalMain::checkForErrorlog() {
@@ -260,6 +276,7 @@ auto XournalMain::exportPdf(const char* input, const char* output) -> int {
 auto XournalMain::run(int argc, char* argv[]) -> int {
     g_set_prgname("com.github.xournalpp.xournalpp");
     this->initLocalisation();
+    this->migrateSettings();
 
     GError* error = nullptr;
     GOptionContext* context = g_option_context_new("FILE");
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -6,7 +6,6 @@
 #include <gtk/gtk.h>
 #include <libintl.h>
 
-#include "filesystem.h"
 #include "control/jobs/ImageExport.h"
 #include "control/jobs/ProgressListener.h"
 #include "gui/GladeSearchpath.h"
@@ -25,6 +24,7 @@
 #include "config-dev.h"
 #include "config-paths.h"
 #include "config.h"
+#include "filesystem.h"
 #include "i18n.h"
 
 #if __linux__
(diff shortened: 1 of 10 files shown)


Technius:
Thanks for the PR! Unfortunately, we cannot merge this PR until we start moving on to 1.1.0 (which should be after the next release, 1.0.14).

We need migration code because we have a non-insignificant number of users. Also, the settings are getting more complicated and it can be a lot of trouble to reconfigure everything. Migration can be done during startup: if the folders don't exist in the XDG folder hierarchy, load the old configuration and save into the XDG folders. This should be followed by a dialog notifying the user of the change (so that they can go delete the old configuration folder, if desired).

sikmir:
@Technius I've added migration code, currently it cannot be built (on top of master branch) as far as it uses C++17 features (`std::filesystem`), but as far as I see there is https://github.com/xournalpp/xournalpp/pull/1437, so `v1.1.0` will use C++17, right?

Febbe:
We can't support c++17 std:: filesystem currently, because it is not supported by AppleClang 10

LittleHuba:
We will have to support AppleClang 10.1 at least, as we support macOS 10.13.

LittleHuba:
@sikmir we brought our code now to the new C++ standard. Please update this PR to work with this standard. You may talk with @Febbe for the inclusion on std::filesystem. He mentioned a fix for this specific problem.

Febbe:
Do your code use std::filesystem or a library you use? If it's only your code use the namespace fs instead of std:: filesystem. I'll then add a follow-up PR, which includes cmake configuration and the header file, containing the mapping from fs:: to std:: filesystem or respectively to std::experimental / boost::filesystem

src/control/XournalMain.cpp
```diff
@@ -32,6 +32,9 @@
 #endif
 
 #include <algorithm>  // std::sort
+#include <filesystem>
+
+namespace fs = std::filesystem;
```
Febbe: Replace with 
```
#include "filesystem.h"
``` 
Remove the fs namespace alias.

Febbe: FYI the file you will include is https://github.com/xournalpp/xournalpp/blob/pr-freeze/filesystem/src/filesystem.h.in . But don't Rebase your code upon this branch, because this branch needs a Rebase first.

sikmir: Done.

sikmir:
> Do you want to include the remaining points of the PR mentioned in the initial description?

Yes.

sikmir:
Updated and rebased on top of master.

sikmir:
@Febbe any plans for `pr-freeze/filesystem` branch to be merged?

Febbe:
Yes after the PR freeze.

sikmir:
> Yes after the PR freeze.

But I see only `pr-freeze/c++17` and it's already merged (#1485). Or I missed something?

Febbe:
The clang-format isn't applied.

Febbe:
@sikmir you can continue your work now. The support for stl::/boost::filesystem is implemented

src/control/XournalMain.cpp
```diff
@@ -70,8 +71,23 @@ void XournalMain::initLocalisation() {
     std::cout.imbue(std::locale());
 }
 
+void XournalMain::migrateSettings() {
+    Path newConfigPath(g_get_user_config_dir());
+    newConfigPath /= g_get_prgname();
```
Technius: `g_get_prgname()` is typically null here since the migration code is run before calling `gtk_init`. See
https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-prgname

Technius:
Hi @sikmir, this is a pretty important change and it's been sitting here for a while now. Would it be alright if I take over and finish the PR?

sikmir:
> Hi @sikmir, this is a pretty important change and it's been sitting here for a while now. Would it be alright if I take over and finish the PR?

Sure! Actually everything is ready, except tetsing on Windows and macOS.

Technius:
Runs fine on Windows with #1914 applied. Mac would probably also run without issue, but I haven't tested.

Technius:
Rebased on master with the Windows patch applied and some more error checking.

Technius:
Loading custom plugins is slightly out of scope here, as it's actually a feature enhancement that's more suitable for a follow-up pull request. Merging in 24 hours if there are no objections.

PR #1386, opened 2019-07-31
Fix #1385 search input not recognized

### LINKED ISSUE #1385: Keyboard input not recognized in document search
**Affects versions :**
- OS: Linux Mint 19.1 Cinnamon, Cinnamon version 4.0.10
- Hardware: Lenovo X1 Yoga
 - Version of Xournal++ 
   1.0.13
   20190729.1322
   b369cf06-dirty

**Describe the bug**.
No keyboard input is recognized in the document search field. This persists also with an external keyboard plugged in. For other functions of Xournal++ the keyboard works properly, for instance in text fields or when using Ctrl+F. 
On a second system with identical OS and hardware the problem exists as well. 

**To Reproduce**
- Open a blank page or pdf document
- Access the search field via Ctrl + F
- Type any text in the search field

lehmanju:
The "activate" signal is not necessary because as soon as text is changed
a new search is started anyway. key-press-events have to be handed to
the underlying widget after processing.

### DIFF
--- a/src/gui/SearchBar.cpp
+++ b/src/gui/SearchBar.cpp
@@ -30,23 +30,22 @@ SearchBar::SearchBar(Control* control)
 
 	// TODO: When keybindings are implemented, handle previous search keybinding properly
 	GtkWidget* searchTextField = win->get("searchTextField");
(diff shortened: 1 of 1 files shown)


andreasb242:
Thanks, looks good.

PR #1388, opened 2019-08-01
correctly handle extensions when exporting

veecue:
Since #1117, we allow for extension like ".pdf.xopp". This change led to
a bug where overwrithing an existing ".pdf" file in pdf export created a
new file with the extension ".pdf.pdf" instead of overwriting the old
one. This is fixed and additionally all current and future export
filetypes show the same behaviour now.


Technius:
Can you also add test cases for `.pdf.pdf` and `.xopp.xopp` (it might actually happen to some file)?

Issue #1393, opened 2019-08-04
Xournal++ does not open any document


tumb1er:
**Affects versions :**
 - OS: Ubuntu 19.04
 - Version of Xournal++ 1.0.13

**Describe the bug**
Error "XML Parser error: Wrong count of points (1)" on any opened document

**To Reproduce**
1. Create new document
2. Draw anything (single stroke is enough)
3. Save document
4. Reopen doc from "open" dialog or "Recent Documents"
5. An "XML Parser error" occurs, file not loaded.

**Expected behavior**
File is opened correctly

**Screenshots of Problem**
no.

**Additional context**
no.

veecue:
I'm afraid, I can't reproduce this. Can you specify your exact xournalpp version (commit id) and attach the .xopp file that creates the error (you have to zip it so you can upload it to github)?

tumb1er:
[2019-08-03-Note-16-17.zip](https://github.com/xournalpp/xournalpp/files/3466367/2019-08-03-Note-16-17.zip)

```
$ apt show xournalpp
Package: xournalpp
Version: 1.0.6-0~201908020631~ubuntu19.04.1
Priority: optional
Section: misc
Maintainer: Andreas Butti <andreasbutti@gmail.com>
Installed-Size: 4 253 kB
Depends: libc6 (>= 2.29), libcairo2 (>= 1.4.10), libgcc1 (>= 1:3.0), libgdk-pixbuf2.0-0 (>= 2.22.0), libglib2.0-0 (>= 2.37.3), libgtk-3-0 (>= 3.19.12), liblua5.3-0, libpango-1.0-0 (>= 1.14.0), libpangocairo-1.0-0 (>= 1.14.0), libpoppler-glib8 (>= 0.18.0), libportaudio2 (>= 19+svn20101113), libportaudiocpp0, libsndfile1 (>= 1.0.20), libstdc++6 (>= 6), libx11-6, libxi6 (>= 2:1.2.0), libxml2 (>= 2.7.4), libzip5 (>= 1.0), zlib1g (>= 1:1.1.4)
Download-Size: 955 kB
APT-Manual-Installed: yes
APT-Sources: http://ppa.launchpad.net/andreasbutti/xournalpp-master/ubuntu disco/main amd64 Packages
Description: Xournal++ is a handwriting Notetaking software with PDF annotation support. Supports Pen input like Wacom Tablets.
```

veecue:
[...]

PR #1396, opened 2019-08-05
write coordinates in C locale

veecue:
Since a6276cd2 we use g_snprintf for coordinate string formatting. This
function uses number formatting dependant on the current locale and
therefore might produce invalid files. This is fixed now.

Fixes: #1393
CC @siliconninja


src/control/xml/XmlStrokeNode.cpp
```diff
@@ -89,15 +89,12 @@ void XmlStrokeNode::writeOut(OutputStream* out)
...
 
-		gchar coordinateStr[Util::PRECISION_FORMAT_BUFF_LEN];
-		Util::getCoordinateString(coordinateStr, Util::PRECISION_FORMAT_BUFF_LEN, points[0].x, points[0].y);
-		out->write(coordinateStr);
+		Util::writeCoordinateString(out, points[0].x, points[0].y);
 
 		for (int i = 1; i < this->pointsLength; i++)
 		{
-			Util::getCoordinateString(coordinateStr, Util::PRECISION_FORMAT_BUFF_LEN, points[i].x, points[i].y);
+			Util::writeCoordinateString(out, points[i].x, points[i].y);
```
siliconninja: I like that you optimized the calls to write the coordinate string in addition to formatting it, it makes the code look neater, especially here.

test/control/LoadHandlerTest.cpp
```diff
@@ -303,6 +307,123 @@ class LoadHandlerTest : public CppUnit::TestFixture
...
+				CPPUNIT_ASSERT(false);
+				break;
+			}
+		}
+	}
+
+	void testLoadStoreLoadGerman()
+	{
+		constexpr auto testLocale = "de_DE.UTF-8";
+		bool hastTestLocale = false;
```
Technius: Is this a typo or intentional?

veecue:
@LittleHuba How can I check whether the test is executed on the CI so the multi-locale test is only forced there?

LittleHuba:
Just traf the log of the CI test. It lists all tests that were executed.

veecue:
@LittleHuba No, I want to do a check inside the test whether it is running on the CI so that users without a multilanguage setup don't get local test failures

LittleHuba:
The way to go then is an environment variable that we set in the CI environment.

src/control/xml/XmlStrokeNode.cpp
```diff
@@ -89,15 +89,12 @@ void XmlStrokeNode::writeOut(OutputStream* out)
...
 			out->write(" ");
```
JJones780: Good catch.. you changed the delimeter to after.

src/util/Util.h
```diff
@@ -55,17 +56,12 @@ gboolean paintBackgroundWhite(GtkWidget* widget, cairo_t* cr, void* unused);
...
 
-constexpr const gchar* PRECISION_FORMAT_STRING_XY = "%.8f %.8f";  // note the space delimiter
-
-constexpr const int PRECISION_FORMAT_BUFF_LEN = G_ASCII_DTOSTR_BUF_SIZE * 2 + 1;
-
 }  // namespace Util
 
-static const size_t npos = std::numeric_limits<size_t>::max();
\ No newline at end of file
+static const size_t npos = std::numeric_limits<size_t>::max();
```
JJones780: Glad you added the cr at eof !

veecue: Wasn't intentionally me but probably Qt creator or clang-format ;)

test/control/LoadHandlerTest.cpp
```diff
@@ -303,6 +310,125 @@ class LoadHandlerTest : public CppUnit::TestFixture
...
+		h.prepareSave(doc1);
+		Path tmp = Util::getTmpDirSubfolder() / "save.xopp";
+		h.saveTo(tmp);
+
+		// Create a second loader so the first one doesn't free the memory
+		LoadHandler handler2;
+		Document* doc2 = handler2.loadDocument(tmp.str());
+		auto elements2 = getElements(doc2);
+
+		auto coordEq = [](double a, double b) { return std::abs(a - b) <= 1e-6; };
```
JJones780: The file currently has values to 8 decimal places.

src/control/xml/XmlStrokeNode.cpp
veecue: Wow, thanks... Was neither found in my testing nor in the unit test... Wonder when the XmlStrokeNode is used at all...

test/control/LoadHandlerTest.cpp
```diff
@@ -303,6 +311,141 @@ class LoadHandlerTest : public CppUnit::TestFixture
...
+			}
+		}
+	}
+
+#ifdef __linux__
+	void testLoadStoreLoadGerman()
+	{
+		constexpr auto testLocale = "de_DE.UTF-8";
+		bool hasTestLocale = false;
+		for (auto available = g_get_language_names(); *available; available++)
```
JJones780: I've already got de_DE.utf8 installed yet g_get_language_names() only gave a subset of locales (which didn't include de_DE.UTF-8)
`locale -a`  on the command line verified it was installed.
Commenting out 426: CPPUNIT_ASSERT(hasTestLocale); resulted in a sucessful test.
inserting:
```
char* currLocale = setlocale(LC_ALL, testLocale);
        g_warning("curr Locale = %s",currLocale);
```
verified that the locale was being loaded ( it returns null for failed setLocale )
`** (process:29428): WARNING **: curr Locale = de_DE.UTF-8`
so...
 It may be better to just set the locale and assert the return string from setLocale() != null ??

JJones780: Thanks for requesting the review.. forced me to get aquainted with the testing apparatus .. which is something I should've done when I first started contributing. Thanks also for adding this test!

veecue: great suggestion

veecue:
Finally :) Will merge this soon if no objections

PR #1409, opened 2019-08-10
Remove touch workaround, scroll handling, old input system

lehmanju:
This removes the old input system together with custom scroll handling and the touch workaround.

This compiles and works without any issues.


LittleHuba:
I think this is ready to merge. We can only let the users with the devices requiring this workaround stumble on this and should it not work revert this commit. Probably squash it to ensure reverting is easy.

Additionally we should add an entry to our FAQs that explains how to use the new manual mapping feature for users that previously used the workaround.

lehmanju:
Strange, I had an issue where xournal was completely unresponsive but no errors and stack showed it was idle in gtk loop. Maybe a gtk issue or it was confused by settings. Works fine now.

LittleHuba:
I'm actually adverse to removing the old input system yet. I think it would be better to make another minor release where we switch to the new input system by default with this as backup if any problems arise for users. We still get tickets in. After we did this we should remove it before the following release. This switch seems to hard and prone for big problems.

LittleHuba:
Closed in favour of #1560. Removal of old input system will be in the release after the next to give a fallback to users if they encounter a problem.

PR #1410, opened 2019-08-11
Add option to disable overlay scrolling

lehmanju:
Scrollbars normally fade out after a short time which makes it
impossible to use the scrollbars on touchscreens.

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -84,6 +84,7 @@ void Settings::loadDefault()
 	this->highlightPosition = false;
 	this->darkTheme = false;
 	this->scrollbarHideType = SCROLLBAR_HIDE_NONE;
+	this->disableScrollbarFadeout = true;
 
 	//Set this for autosave frequency in minutes.
 	this->autosaveTimeout = 3;
@@ -562,6 +563,10 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur)
 			this->scrollbarHideType = SCROLLBAR_HIDE_NONE;
 		}
 	}
+	else if (xmlStrcmp(name, (const xmlChar*) "disableScrollbarFadeout") == 0)
+	{
+		this->disableScrollbarFadeout = xmlStrcmp(value, (const xmlChar*) "true") ? false : true;
+	}
 	else if (xmlStrcmp(name, (const xmlChar*) "audioSampleRate") == 0)
 	{
 		this->audioSampleRate = tempg_ascii_strtod((const char *) value, NULL);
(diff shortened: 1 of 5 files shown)


JJones780:
You've used   cbDisableScrollbarFadeout etc. for the gui code and text    and  overlayScrollingDisabled in the code and settings file.  This would be a good time to make the terms consistent.

I'd prefer you consolidated on the DisabledFadeout side...  _i.e.  disableScrollbarFadeout isScrollbarFadeoutDisabled() etc._     as  disableOverlayScrolling is misleading.

We use the overlay layer to show the floating toolbox and it doesn't scroll.

JJones780:
Consider making this setting default  true. Better to avoid frustrating users than the slight inconvenience of non-disappearing scrollbars (until they find this setting ).

If it is made on by default, would it also make sense to limit this to when the user has chosen the touchscreen to draw ( to avoid the inconvenience of the non-disappearing scrollbars ) or is it needed in other cases?

veecue:
I agree with @JJones780 to only make it default when the user wants to use the touchscreen to draw. However, it might be useful too to fix the two-finger gestures when the user is using the Touchscreen to draw.

lehmanju:
@veecue two-finger gestures work without any problems (except the zoom being laggy of course) . Even if touchscreen is set to mouse/pen for drawing.

I'd rather not couple overlay scrolling with input settings because you actually never know the users intention on why he might changed the input class. Furthermore, most of the time you use xournalpp in touchscreen mode anyways and scrollbars being invisible is just annoying.

src/control/settings/Settings.cpp
```diff
@@ -980,6 +985,7 @@ void Settings::save()
 		saveProperty((const char*) "scrollbarHideType", "none", root);
 	}
 
+	WRITE_BOOL_PROP(disableScrollbarFadeout);
```
JJones780: Needs to be moved up or down... currently it is between the scrollbarHideType setting and its comment line.  If you look in the ~/.xournalpp/settings.xml file you'll see:

```
 ...
  <property name="scrollbarHideType" value="none"/>
  <property name="disableScrollbarFadeout" value="true"/>
  <!--Hides scroolbars in the main window, allowed values: "none", "horizontal", "vertical", "both"-->
...
```

JJones780:
This looks great to me _( but do note the settings.cpp issue above)_ .  I agree that the fadeout is more annoying than useful and having this on by default is good for all use cases. 
I tried mixing and matching the scrollbar preference options and couldn't get any adverse behaviour with this code change. 
I think the two-finger behaviour @veecue is referring to is being able to move the page as well as zoom with two fingers. Currently we ( or just I ?) can only zoom.

lehmanju:
Good point! I will rearrange the code in settings.cpp and try to find a solution for two finger scrolling (probably in a separate PR though)

LittleHuba:
> @lehmanju The two-finger-scrolling should have been introduced in #1146 However the current implementation feels a little weird to me, since I always zoom while scrolling. Maybe a threshold would be good that switches between zoom and scroll mode

Scrolling while zooming should not be a problem as you can then zoom and move the content you want to see at the same time. Take a look at Android smartphones or other touch-aware and zoom-capable applications. They do not disable moving while zooming and do not take any threshold into account.

LittleHuba:
There are multiple PRs with changes to the settings.glade file... Prepare for some ugly merge conflicts...

lehmanju:
Would merge this tomorrow 18:00 CEST if there are no objections.

PR #1411, opened 2019-08-11
Fix zoom fit scroll

### LINKED ISSUE #941: Touch scrolling stops reacting regularly
When scrolling with the touchscreen sometimes there is no reaction. This happens about every 7th touch input. I don't know how to further narrow down the problem.

### LINKED ISSUE #1132: Maximizing the window causes scrolling issues

### LINKED ISSUE #1383: On window resize page scrolling is not working
**Affects versions :**
 - OS: Arch Linux
 - Version of Xournal++ master

**Describe the bug**
After starting xournalpp you can barely move the page around by touch and hold scrolling. Horizontal movement does not work at all. After CTRL+Mouse zooming everything works fine.

This happens when resizing the window, specifically from smaller to larger as the page zooms in as well it lands partly outside the viewport.

**To Reproduce**
1. Start xournal++
2. move page around

**Expected behavior**
Render the page centered on startup and be able to scroll without any issues.

**Screenshots of Problem**
Bildschirmfoto von 2019-07-28 20-20-43

lehmanju:
Fixes scrolling when using zoom fit and resizing window.

Underlying issue was missing repaint of child widget. (refactor_layout_resize should be merged first)

Closes #1383
Closes #1132 
Closes #941

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -590,13 +590,9 @@ bool ZoomControl::onWidgetSizeChangedEvent(GtkWidget* widget, GdkRectangle* allo
 	zoom->updateZoomPresentationValue();
 	zoom->updateZoomFitValue(r);
 
-	// Todo: remove after change to "expose_event"
 	auto layout = gtk_xournal_get_layout(zoom->view->getWidget());
-	GdkRectangle allNew = {allocation->x, allocation->y, std::max(allocation->width, layout->getMinimalWidth()),
-	                       std::max(allocation->height, layout->getMinimalHeight())};
-
 	layout->layoutPages(allocation->width, allocation->height);
-	gtk_widget_set_allocation(zoom->view->getWidget(), &allNew);
+	gtk_widget_queue_resize(zoom->view->getWidget());
 
 	return true;
 }


lehmanju:
Open for suggestions on how to fix the scroll position after zoom.

lehmanju:
```
XournalView::scrollTo
```
is bugged and does not take padding into account.

LittleHuba:
> Open for suggestions on how to fix the scroll position after zoom.

Well you could calculate a relative offset compared to the document dimensions before zomming and then use this to recalculate the absolute offset after zooming.

lehmanju:
@LittleHuba I don't think this is easily solvable, after some digging I found out that value calculation is only sometimes wrong and the scrollbars are set to a wrong value because they are rendered afterwards. Therefore I will leave this as is, it enables scrolling at least, and work on a drawing area rework (XournalWidget, ZoomControl, XournalView) with the appropriate callbacks on events.

LittleHuba:
We can solve the scrolling issue in another PR. But we should create an issue to not forget it.

PR #1414, opened 2019-08-16
Change appid to com.github.xournalpp.xournalpp

muelli:
this commit attempts to change the appid to be more in line with the appstream spec as it requires a reverse dns name.

It builds and installs fine for me, but I haven't checked whether things like the thumbnailer work as they did before.

The appdata file validates correctly.


Technius:
This will not remove the files from the previous naming scheme, so we'll need to include a note in the changelog.

PR #1416, opened 2019-08-17
fixes #1415  WIP: Adds 2 tools for pen sizes. 

### LINKED ISSUE #1415: 'Very thick' on toolbar
It seems that there are five pen thickness options (very thin, thin, medium, thick and very thick) but only the middle three of these have toolbar items. I usually swap between medium and very thick, and it's tedious to go into the menu every time I need the latter (especially since I like fullscreen).

Ref: https://github.com/xournalpp/xournalpp/blob/90f583a851265c52abc69aa7fcc81fe632ca23d1/src/gui/toolbarMenubar/ToolMenuHandler.cpp#L524-L526

JJones780:
Todo: needs icons and a bit more code.


JJones780:
Pen, Eraser and Highlighter now all have 5 thickness options for consistency.
Came across a bug,
Made the variable names more consistent.
I've also taken this opportunity to try changing the thickness icons - we've already got lots of circles for colours.  I tested out a bunch of options and a simple bent line looked the best IMHO. they are on the right side of the toolbar in the image below. If you have a suggestion please do a quick sketch/mock-up so we can compare. Another option would be to change the colour icons ( to squares or paint splatters?) .. or I guess leave everything as circles.

image

gomarcus:
Hey, how come this is missing from version 1.2.2 ??

rolandlo:
It's not missing. If you're looking for `very fine` or `very thick` you may be missing that they are not on all toolbars by default. You can customize the toolbar or find those thicknesses under menu `Tools`.

gomarcus:
ok. thanks! I did miss that. But is there a simple way to change the icon sizes? On Surface Pro 9 the buttons are so damn small.

rolandlo:
> ok. thanks! I did miss that. But is there a simple way to change the icon sizes? On Surface Pro 9 the buttons are so damn small.

There is no clean solution, but you may try [this suggestion](https://github.com/xournalpp/xournalpp/issues/3453#issuecomment-1067227494)

PR #1426, opened 2019-08-19
Fix calculation of stroke's bounding rectangle

### LINKED ISSUE #1425: Improper item selection bounds and clipping
**Affects versions :**
 All

**Describe the bug**
Strokes with thick lines are clipped on top and left but extra space on bottom and right sides

**To Reproduce**
Use a thick highlighter to draw a stroke then select it.

Also happens when selecting multiple strokes.

**Expected behavior**
bounding box evenly surrounds selected stroke(s).

image

JJones780:
fixes #1425 
Now properly centered
and takes into account variable (pressure) width strokes.

### DIFF
--- a/src/model/Stroke.cpp
+++ b/src/model/Stroke.cpp
@@ -549,7 +549,8 @@ bool Stroke::intersects(double x, double y, double halfEraserSize, double* gap)
 /**
  * Updates the size
  * The size is needed to only redraw the requested part instead of redrawing
- * the whole page (performance reason)
+ * the whole page (performance reason).
+ * Also used for Selected Bounding box.
  */
 void Stroke::calcSize()
 {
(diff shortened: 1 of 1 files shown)


src/model/Stroke.cpp
```diff
@@ -565,35 +565,44 @@ void Stroke::calcSize()
...
+	bool hasPressure = points[0].z != Point::NO_PRESSURE;
+	double halfThick = this->width / 2.0;  //  accommodate for pen width
+
+	for (int i = 0; i < this->pointCount; i++)
 	{
-		if (minX > points[i].x)
+		if (hasPressure) halfThick = points[i].z / 2.0;
+
+		double edge = points[i].x - halfThick;
+		if (minX > edge)
```
Technius: You can make this cleaner and clearer by using `std::min` and `std::max`:
```suggestion
		minX = std::min(minX, points[i].x - halfThick);
		minY = std::min(minY, points[i].y - halfThick);
		maxX = std::max(maxX, points[i].x + halfThick);
		maxY = std::max(maxY, points[i].y + halfThick);						
```

JJones780: Gee.. why didn't I see that!  Thanks!

PR #1427, opened 2019-08-20
improve hint for latex first timers.

JJones780:
Indicate the latex package that includes standalone when telling the user they need to install standalone.


LittleHuba:
Isn't the package OS-specific? This could actually confuse users even more...

Issue #1406, opened 2019-08-09
Have text with multiple colors in same text box


fxkrait:
**Is your feature request related to a problem? Please describe.**
I'm trying to use xournal for notetaking, and would like to be able to have keywords set to distinct colors from the main text color.

Most of the text I just drag and drop from a PDF ebook in Evince. The current way that I do this is simply to drag and drop (making a text box) with just the keyword, color it, then do the same for the main text, and try to line it up as best I can. However, this is a lot slower than being able to drag and drop the text in its entirety and then changing the color of certain keywords. It is also makes it more difficult to copy the entirety of the text into say an anki card afterword, as now I have to copy and paste multiple snippets of text. 

I believe this is partially related to #1092, in terms of having text with different styles in the same text field. However, that issue talks only about text styles (bold/italic/etc) and not text colors, so I believe this is slightly different. I can remove this if they are too similar, and add a comment to that issue.

**Describe the solution you'd like**
I would like to be able to select the keyword text, and click on the color icons to change it to the given color, while keeping the main text the same.

**Additional context**
Example of what I am doing:
![multipleColors](https://user-images.githubusercontent.com/7407672/62776257-e39e3800-ba5e-11e9-82ed-3d08075cb50c.png)

Technius:
This can be implemented much more easily than #1092 but is definitely limited in scope, so I'll leave this open for now.

JJones780:
See #1430 for an interesting hack that may lead us to a workable solution.

Let's take the discussion to #1092.  I'll let you close this issue and join over there.
Your feedback is appreciated!

fxkrait:
Sounds good.

Issue #1092, opened 2019-04-24
Add inline styling for text tool


blackhole89:
It would be immensely useful, and - I believe - somewhat backward-compatible with Xournal to make the text tool support inline styling rather than only a single style for the whole length of the input. In fact, Pango [already provides a markup language for this purpose](https://developer.gnome.org/pango/stable/PangoMarkupFormat.html): a really easy way to get a dirty version of this feature is to go into `TextView::drawText` and replace `pango_layout_set_text` with `pango_layout_set_markup`. This will result in text like `hello <b>world</b>` properly rendering in boldface when inert, but not when it is currently being edited, and making the editor (`src/gui/TextEditor.cpp`) support it involves significantly more work.

I'm willing to volunteer coding it up, but would like to coordinate with the maintainers a little bit before I actually sink the required effort. My current impression is that there is no way around doing something like the following:

- Replace `gtk_text_buffer_set_text` in TextEditor's constructor with some appropriate construction based on [gtk_text_buffer_insert_markup](https://developer.gnome.org/gtk3/stable/GtkTextBuffer.html#gtk-text-buffer-insert-markup). (Easy.)
- Unfortunately, gtk_text_buffer's internal data structures store rich text in a completely different format. We would therefore require two layers of glue code: one to transfer the gtk_text_buffer attributes to the pango_attr_list used for rendering (comparitively easy), and one to serialise the gtk_text_buffer as Pango markup when editing is finished. The latter is the annoying part: gtk_text_buffer [provides an API for this](https://developer.gnome.org/gtk3/unstable/GtkTextBuffer.html#gtk-text-buffer-register-serialize-format), but the only code on the internet actually using it seems to be [this 747-line mouthful](https://github.com/ptomato/osxcart/blob/master/osxcart/rtf-serialize.c) that outputs RTF.
- We'd also need to rewire some GUI signals to only apply attributes for boldface, font changes etc. at the current cursor position/selection.

The end result, ideally, would be that unstyled text continues being loadable in plain Xournal as before, and styled text will have tags that look like HTML appear when loaded in it. There are some alternatives to consider: e.g. to use something that looks like Markdown (and may even be keyboard-enterable) rather than the HTML-like Pango markup.

andreasb242:
Hi @blackhole89

This Idea is not new, but as you already mentioned, it's much work.
[...]

PR #1430, opened 2019-08-20
experiment with raw pango layout in text objects.

JJones780:
Make text objects aware of Pango markup

re: #1092  #1406 

xournalMarkup

### DIFF
--- a/src/view/TextView.cpp
+++ b/src/view/TextView.cpp
@@ -52,7 +52,18 @@ void TextView::drawText(cairo_t* cr, Text* t)
 
 	PangoLayout* layout = initPango(cr, t);
 	string str = t->getText();
-	pango_layout_set_text(layout, str.c_str(), str.length());
+
+	PangoAttrList* pangoAttr;
+	char* plainText;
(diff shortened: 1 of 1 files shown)


JJones780:
### Pango Markup is here:
https://developer.gnome.org/pango/stable/pango-Markup.html

LittleHuba:
Now we only need a way to edit the text nodes without writing markup... It would be nice if we could use a context sensitive menu that shows up at the border of the text field.

JJones780:
It wouldn't be too hard to do this the right way... by storing a separate attribute field for the text objects and editing it with small buttons that appear with the text editor to set colours, bold etc. It may break the file format though.

Technius:
This has the potential to break existing files if they already contain text that pango recognizes as markup (e.g. what if someone wrote notes about HTML?). I suggest deferring this to 1.1.0.

JJones780:
It's a hack.... It would at least get a prefs 'experimental' setting if we wanted to consider merging it ;)

For the time being it's only here to elicit discussion.

LittleHuba:
As it is only a really small change I think defering it to 1.1.0 seems like a good idea. I think merging this in as experminental is more hassle than it is worth for now.

PR #1437, opened 2019-08-23
cmake: Switch to C++17 standard

veecue:
Since we want to use C++17 features and the standard is available for
all major compilers, require it from cmake


Febbe:
add `add_definitions(-D_USE_MATH_DEFINES)` to CMakeLists.txt, then the CI Pipes should work again.

veecue:
@Febbe Thanks for the hint! However, I have got a different problem: Since the current CI (Ubuntu 16.04) uses gcc 5.4 which definitely doesn't support the whole C++17 standard, I would expect the build to fail during `cmake` (because `CMAKE_CXX_STANDARD_REQUIRED` ) is set. Then I would try to upgrade the CI (maybe 18.04) so that it properly builds. If you have any idea, why it doesn't work out this way, feel free to just push to the branch :).

Febbe:
Unfortunately I can't help you with the azure pipelines, @LitteHubba is the right person

LittleHuba:
Updating the CI is not actually good. We explicitly want an old environment (backward compatibility) 
You can search for an apt repository with a newer version of gcc though and install it from there. It is actually what we currently do as well.

veecue:
@Febbe I'll figure out the CI stuff, but I'll be glad if you find a way to make the Linux CI Build fail during cmake

veecue:
Ok, so apparently gcc 5.4 (with almost no C++17 support) is recognized to support std_cxx_17 by cmake 🤦‍♂

Febbe:
You can search for the last c++17 feature, which was implemented by gcc/clang/msvc and use that in the tests, splicing maps, and the filesystem lib for example.

Febbe:
Ok @veecue I fixed the Linux pipeline. Also all pipelines will fail now, if a compiler is used, which does not support the newest c++17 features

LittleHuba:
Please test on the native mac pipeline and not on the homebrew one. It will be removed shortly as it is not usable for our releases and takes more time than the native build.

azure-pipelines/CI.yml
```diff
@@ -34,7 +34,7 @@ stages:
 
       - job: 'Windows'
         pool:
-          vmImage: 'vs2017-win2016'
+          vmImage: 'windows-2019'
```
LittleHuba: Why do we need this change? It would be better to stay on the old system for testing.

src/util/logger/Logger.cpp
```diff
@@ -11,7 +11,7 @@
 #include <iostream>
 #include <stdio.h>
 
-#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
+#if defined(_WIN32) || defined(_WIN32) || defined(__WIN32__)
```
LittleHuba: `_WIN32` check is now duplicated

azure-pipelines/steps/build_mac.yml
```diff
@@ -6,9 +6,28 @@ steps:
...
+      brew install clang clang++
+      brew info clang
+      brew info clang++
+      echo 'export PATH="/usr/local/opt/llvm/bin:$PATH"' >> ~/.bash_profile
+      export CC=clang
+      export CXX=clang++
+      export LDFLAGS+=" -L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"
+      export CPPFLAGS+=" -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include/c++/v1/"
+    #xcode-select --install
+    displayName: 'Install llvm'
```
LittleHuba: Again the Homebrew Pipeline will be removed shortly. The current master does actually not use it anymore. Thus working on it makes absolutely no sense. Work on [the native pipeline](https://github.com/xournalpp/xournalpp/blob/master/azure-pipelines/steps/build_mac_native.yml) instead.

azure-pipelines/steps/build_linux.yml
```diff
@@ -7,13 +7,19 @@ parameters:
...
+      sudo apt install gcc-8 g++-8
       sudo apt-get install -y cmake ninja-build libcppunit-dev libgtk-3-dev libpoppler-glib-dev portaudio19-dev libsndfile-dev liblua5.3-dev libzip-dev
+      g++ --version
     displayName: 'Install dependencies'
   - bash: |
       mkdir build
     displayName: 'Create build directory'
   - bash: |
+      export CXX=g++-8
+      export CC=gcc-8
```
LittleHuba: We will have to add this requirement to our manual build instructions.

CMakeLists.txt
```diff
@@ -1,20 +1,25 @@
...
+cmake_minimum_required(VERSION 3.10)
 
 project ("Xournal++" CXX C)
 
+## Project related Variables
 ## Also update changelog in debian folder!
 set (CPACK_PACKAGE_VERSION_MAJOR "1")
 set (CPACK_PACKAGE_VERSION_MINOR "0")
-set (CPACK_PACKAGE_VERSION_PATCH "13")
+set(CPACK_PACKAGE_VERSION_PATCH "19")
```
LittleHuba: Where does the 19 come from? We are still in 1.0.13 and will switch to 1.0.14 at release but this switch will be done in a separate commit.

src/util/logger/Logger.cpp
Febbe: The hole line is sort of inflated `#ifdef(_WIN32)` is sufficient. And I just replaced all WIN32 defines with _WIN32 for testing purposes.

azure-pipelines/CI.yml
Febbe: There aren't that many dependency issues with different versions of windows. But the price using an old environment and not up to date compiler is way to high in my opinion.

CMakeLists.txt
Febbe: Its like 1.0.13 the development version of 1.1.0, instead of 1.0.14. And after we release 1.0.14 we will work on 1.1.

LittleHuba: Still this should be in its own commit. It is a version change and those should be very clear in the history. Additionally it should be done AFTER all commits of the version have been merged. This is so all commits can be sorted to one version without guessing.

azure-pipelines/CI.yml
LittleHuba: The thing is that you always test on the oldest supported system. Backward compatibility. Either we formally declare the Windows version no longer supported or we test for it. There are only these two ways we can take without a potential problem in the future. 
As declaring it no longer supported is not really necessary as Xournal++ works on it and we don't want to smaller our user base unnecessary we will have to test on it.

azure-pipelines/steps/build_linux.yml
Febbe: This has to be done at the time we merge.

azure-pipelines/steps/build_mac.yml
Febbe: This can also be done in another branch?

Febbe: deferred to pr-freeze/filesystem

cmake/find/FindCXX17.cmake
```diff
@@ -0,0 +1,68 @@
+
```
LittleHuba: Please add credits for the developer of this file

Febbe: I've made this file, but the FindFilesystem.cmake file isn't from me/us I really don't know who originally created it. I just modified round about 20%.

LittleHuba: You have found the file somewhere on Github if I'm not mistaken? Search your browser history and provide the original link of the file and the username of the author. Then write in it that you based your work on this file.
A quick search resulted in this file, which might be the one you based your work on:
https://github.com/vector-of-bool/neo-io/blob/master/cmake/FindFilesystem.cmake

LittleHuba: The problem is if we do not credit and somebody sees this we might be blocked from releasing the software on the repositories of a lot of OSes.

Febbe:
Done in #1488

PR #1438, opened 2019-08-23
prfreeze: Replace things with proper C++

lehmanju:


lehmanju:
Clang format was a bitch on some files, so I applied it to the whole file … as we are doing this anyway it shouldn't matter

lehmanju:
Closing in favor of #1486

PR #1441, opened 2019-08-23
prfreeze: Apply clang format on project

lehmanju:


lehmanju:
well nvm, better do this at the end

veecue:
I have the impression, we should do this before the major code refactors because the clang format applied pipeline will fail otherwise

Febbe:
Nah we can ignore the pipeline till the end and apply the format afterwards

Febbe:
We also need to discuss, how the format shall look like.

PR #1443, opened 2019-08-26
Simplify layout get view at

JJones780:
The fast algorithm was too difficult to understand/debug and prone to error on future code refactoring... so uncomment the concise stl-based binary search which was already there for testing and use it instead.

### DIFF
--- a/src/gui/Layout.cpp
+++ b/src/gui/Layout.cpp
@@ -9,6 +9,12 @@
 #include <algorithm>
 #include <numeric>
 #include <cstdlib>
+
+#include <chrono>
+typedef std::chrono::high_resolution_clock Clock;
+#include <iostream>
+
+
 /**
  * Padding outside the pages, including shadow
  */
@@ -23,6 +29,12 @@ constexpr size_t const XOURNAL_ROOM_FOR_SHADOW = 3;
  * Padding between the pages
  */
 constexpr size_t const XOURNAL_PADDING_BETWEEN = 15;
+
+
+double searchtime1 = 0;
+double searchtime2 = 0;
+double searchtime3 = 0;
+double plotindex = 0;
 
 
 Layout::Layout(XournalView* view, ScrollHandling* scrollHandling)
@@ -349,6 +361,7 @@ XojPageView* Layout::getViewAt(int x, int y)
 // 	auto cit = std::lower_bound( this->sizeCol.begin(),  this->sizeCol.end(), x);
 // 	int cb = cit - this->sizeCol.begin();
 
+	auto t0 = Clock::now();
 
 	auto fast_linear_search = [](std::vector<unsigned> const& container, size_t start, size_t value) {
 		auto row_i = std::next(begin(container), start);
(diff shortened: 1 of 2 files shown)


Febbe:
Mhh I simplified it already. And I tested it on all 2^32^2 combinations its always the same

Febbe:
And as I understand the comment there, the binary search has been exchanged because it was too slow

Febbe:
Can you test it with -O3?

Febbe:
I've run a benchmark, there is no difference between the new STL-Iterator like linear search version and the old index based Version. There is no STL overhead. The binary search version has a stable run time of O(log(n)) and is very small for all vector sizes (~60 to 120 ms). The linear search times are smaller for a difference below 100 to 150 pages. But are very high for a difference of 10000 pages (~ 8000 ms).
[search_bench.zip](https://github.com/xournalpp/xournalpp/files/3553571/search_bench.zip)

I think the best solution is to use the binary search, since the difference between the algorithms is very small between 0 and 150, above the linear search performs extremely bad.

src/gui/Layout.cpp
```diff
@@ -352,117 +338,17 @@ XojPageView* Layout::getViewAt(int x, int y)
...
 
-	auto t3 = Clock::now();
 
-	searchtime1 += std::chrono::duration_cast<std::chrono::nanoseconds>(t1 - t0).count();
-	searchtime2 += std::chrono::duration_cast<std::chrono::nanoseconds>(t2 - t1).count();
-	searchtime3 += std::chrono::duration_cast<std::chrono::nanoseconds>(t3 - t2).count();
-
-
-	if (plotindex == 0)
+	if (foundCol < this->widthCols.size() && foundRow < this->heightRows.size())
```
Febbe: redundant, remove this line, since it is checked in auto optionalPage = this->mapper.at({foundCol, foundRow});

src/gui/Layout.cpp
```diff
@@ -352,117 +338,17 @@ XojPageView* Layout::getViewAt(int x, int y)
...
-
-	//Binary Search ... too much overhead makes this a slower option in our use case.
-	/* This is over 2x faster than the first code block but 3x slower than the second code block*/
+	//Binary Search:
 	auto rit = std::lower_bound(this->heightRows.begin(), this->heightRows.end(), y);
-	int rb = rit - this->heightRows.begin();  //get index
+	int foundRow = rit - this->heightRows.begin();  //get index
 	auto cit = std::lower_bound(this->widthCols.begin(), this->widthCols.end(), x);
-	int cb = cit - this->widthCols.begin();
+	int foundCol = cit - this->widthCols.begin();
```
Febbe: write 
```
auto const& foundRow = std::distance(begin(this->heightRows),rit);
auto const& foundCol = std::distance(begin(this->widthCols),cit);
```

JJones780:
I think I see the difference between your benchmark and my measurements. My results were from measurements accumulated within a running Xournal++.  The difference is that in Xournal++ a vast majority of the time the user is moving at most one row and/or column away.

My results are consistently like this ( for a 500 page pdf ): 
2.40208e+06  new STL-iterator
1e+06   old index based ( from last week )
1.19468e+06  binary search
in ms where lower is better.

Here are some worst case for binary search:
index  agreeR             agreeC          new               old                       binary
6259    555=555=555     0=0=0   4.48116e+06     1.00205e+06     1.63577e+06
6448    301=301=301     0=0=0   4.72137e+06     1.00002e+06     2.08262e+06

The binary search is nice and clean - it's the one to go with.

Febbe:
Mmh the new STL algorithm is mostly the same as the old index based. With optimisations enabled it takes rather the same time or ist Up to 10 nanoseconds faster. I think both old benchmarks were created with no optimizations enabled and or in DEBUG mode. Thus iterators and datastructures and functions are build with a big boilerplate of debug and validation code. This also changes the inlining behavior.

src/gui/Layout.cpp
Febbe: I think with std::distance it is way clearer what the intension is. And we know the return type of std::distance. We also know, that the result is correct and that there are no casting mistakes. So using auto is generally fine and clear. Generally there is no difference between auto const and auto const& for rvalues and pvalues. The compiler will create exactly the same code.

Technius:
Since this is the last behavioral change going into 1.0.14, please make sure to test it thoroughly.

PR #1431, opened 2019-08-20
New Crowdin translations

LittleHuba:
Enjoy!

PR #1447, opened 2019-08-31
changed wording for preferences setting 

JJones780:
As pointed out  by Spanish translator Adolfo Jayme
Changed from:    Background color for window Background
To:            Background color between pages


Technius:
Actually, could we fold this into #1431?

Issue #1442, opened 2019-08-23
Tray Icon & Title Missing


andersonjwan:
**Affects versions :**
 - OS: [Linux - Fedora: 5.2.8-200.fc30.x86_64]
 - Version of Xournal++ [v1.0.13, Built on Aug 23, 2019]

**Describe the bug**
Starting a new instance of Xournal++ works as intended; however, in the GNOME taskbar, the tray icon and name is missing. The results are shown in the screenshots below.

**To Reproduce**
1. Install all required Fedora dependencies as listed.
2. Clone the repository.
3. Compile.
4. Install system-wide from build directory directly.

**Expected behavior**
The tray icon is expected to be on the left of the title. The title should be 'Xournal++' and not 'xournalpp'.

**Screenshots of Problem**
![image](https://user-images.githubusercontent.com/52432873/63629378-a1303b80-c5c6-11e9-8305-e5da166f8582.png)
![image](https://user-images.githubusercontent.com/52432873/63629431-fbc99780-c5c6-11e9-8cea-eb0b20b7ed30.png)

**Additional context**
Xournal++ does show up in the Activities bar, so it must be just a tray icon issue of finding the title and image. Also, please disregard the date of the second screenshot - this was done on a snapshot of a VM. It was indeed built on Aug 23, 2019.

If this is a duplicate issue, please let me know. I can take this down if it has been patched, but not pushed.

Technius:
In #1414, we changed the name of the xournalpp desktop and icon files to `com.github.xournalpp.xournalpp`. If you "installed" xournalpp instead of using a tarball release, you  will need to "reinstall" these files by renaming them (you could also delete them and run the install command again).

For reference, here are the files that need to be renamed (replace `xournalpp` with `com.github.xournalpp.xournalpp`):
```plain
${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps/xournalpp.svg
${CMAKE_INSTALL_PREFIX}/share/mime/packages/xournalpp.xml
${CMAKE_INSTALL_PREFIX}/share/applications/xournalpp.desktop
${CMAKE_INSTALL_PREFIX}/share/thumbnailers/xournalpp.thumbnailers
${CMAKE_INSTALL_PREFIX}/share/metainfo/xournalpp.appdata.xml
```

andersonjwan:
[...]

PR #1452, opened 2019-09-03
Update application metainfo

Technius:
This PR updates some of the project metainfo files.

* Fix the appdata file to hopefully reduce the chance of #1442 occurring
* Update list of authors
* Show git commit id in the about dialog
* Fix xournalpp icon in about dialog not showing due to id change


Technius:
Merging this in 2 days if there are no objections.

desktop/com.github.xournalpp.xournalpp.appdata.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright 2019 Tobias Mueller <tobiasmue@gnome.org>  -->
 <component type="desktop">
-  <id>com.github.xournalpp.xournalpp</id>
+  <id>com.github.xournalpp.xournalpp.desktop</id>
```
muelli: this is wrong, cf. https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic
the desktop suffix is not part of the appid.

desktop/com.github.xournalpp.xournalpp.appdata.xml
```diff
@@ -65,8 +65,13 @@
   </screenshots>
   <launchable type="desktop-id">com.github.xournalpp.xournalpp.desktop</launchable>
   <provides>
-    <id>xournalpp.desktop</id>
+    <binary>xournalpp</binary>
```
muelli: first of all: the `provides` section is optional.
It exists to ease transition to the new app id.
The old appid was `xournalpp.desktop`.
Changing it to a name of a binary is semantically not as correct as the previous appid.

desktop/com.github.xournalpp.xournalpp.appdata.xml
lehmanju: Looking at working files from Gnome applications this is widely used though.

desktop/com.github.xournalpp.xournalpp.appdata.xml
Technius: From the [Section 2.1.1 of the specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent):
> If a metainfo file contains a &lt;provides/&gt; tag, distributors must also ensure that the package providing the file contains all items referenced by that statement, or is installed by a metapackage depending on packages which provide these items.

Since `xournalpp.desktop` is an old id, it will presumably not be installed anymore. In either case, we should still include the binary tag because `xournalpp` _is_ installed.

desktop/com.github.xournalpp.xournalpp.appdata.xml
Technius: The `.desktop` suffix seems to be legacy ([Section 2.2.2 of the spec](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)). So actually this change is unneeded.

desktop/com.github.xournalpp.xournalpp.appdata.xml
Technius: Nevermind, I missed this bit ([Section 2.1.3](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides)):
> Contains the component-ID of another software component. The presence of this tag indicates that the software component containing it is able to provide all functionality of the one referenced in the &lt;provides/&gt; ↪ &lt;id/&gt; tag.
> This is useful in case a component-id had to be renamed in the past, e.g. because its domain name changed.

So the revision is to not remove the provides for `xournalpp.desktop`.

desktop/com.github.xournalpp.xournalpp.appdata.xml
muelli: > Looking at working files from Gnome applications this is widely used though.

what files are you looking at?

Technius:
Merging tomorrow if there are no further objections.

PR #1460, opened 2019-09-11
WIP: Add a custom pinch to zoom algorithm

### LINKED ISSUE #404: Zooming via pinch to zoom causes graphical glitches
When zooming via mulittouch pinch to zoom gestures, the page "glitches" it zooms correctly but it also shows the fullscreen view every few moments, which makes zooming really hard.
Also zooming in general causes the view to scroll quite often so even if I'm using the handle in the menu I end up on the last page even if I started in the middle or somewhere.

taaem:
Calculate the zoom value by hand and not use gtk_gesture for zooming.
Currently always zooms to center of screen.
This fixes #404, but the algorithm probably only works for my screen and I would be glad if someone could give some input there.

### DIFF
--- a/src/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/gui/inputdevices/TouchInputHandler.cpp
@@ -21,27 +21,43 @@ bool TouchInputHandler::handleImpl(InputEvent* event)
 {
 	XOJ_CHECK_TYPE(TouchInputHandler);
 
-	// Disallow multitouch
-	if (this->currentSequence && this->currentSequence != event->sequence)
+	// Don't handle more then 2 inputs
+	if (this->primarySequence && this->primarySequence != event->sequence
+		&& this->secondarySequence && this->secondarySequence != event->sequence)
 	{
 		return false;
 	}
 
-	if (event->type == BUTTON_PRESS_EVENT && this->currentSequence == nullptr)
+	if (event->type == BUTTON_PRESS_EVENT)
 	{
-		this->currentSequence = event->sequence;
-		actionStart(event);
+		if (this->primarySequence == nullptr)
+		{
+			this->primarySequence = event->sequence;
+			actionStart(event);
+		}
+
+		if (this->primarySequence && this->primarySequence != event->sequence && this->secondarySequence == nullptr)
+		{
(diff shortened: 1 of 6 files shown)


src/gui/inputdevices/TouchInputHandler.h
```diff
@@ -11,23 +11,31 @@
...
+	double priLastPosX = -1.0;
+	double priLastPosY = -1.0;
+  double secLastPosX = -1.0;
+  double secLastPosY = -1.0;
+  double lastDiff = 0.0;
 
 private:
 	void actionStart(InputEvent* event);
 	void actionMotion(InputEvent* event);
 	void actionEnd(InputEvent* event);
```
LittleHuba: Rename those to `scroll...` to make their purpose clear as we have two different functionalities in this class now.

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -101,8 +117,71 @@ void TouchInputHandler::actionEnd(InputEvent* event)
...
+	if (this->lastDiff == 0.0)
+	{
+		this->lastDiff = diff;
+		return;
+	}
+
+	ZoomControl* zoomControl = this->inputContext->getView()->getControl()->getZoomControl();
+
+	// TODO: Better algorithm for the zoom value
+	zoomControl->zoomSequenceChange(zoomControl->getZoom() + diff/80000, false);
```
LittleHuba: This new algortihm should make sure that the same point of the page stays under the finger when zooming. I'm sure you'll find a way to do this!

LittleHuba:
I'll try it out the next few days!

LittleHuba:
Just tested the functionality. The zooming works good! The things I mentioned above need to be solved though. 

- [ ] Zooming in a natural way that the same point of the page stays under the finger
- [ ] Scrolling while zooming

taaem:
@LittleHuba so technically these commits *should* fix your points, but I'm not totally sure how the scrolling works, because even though the center coordinates should be right it doesn't zoom correctly.
I don't know how zoomWidgetPosX and scrollPositionX  are different. Maybe you know better what is going on.

PR #1467, opened 2019-09-12
refactor(Util): add DPI_NORMALIZATION_FACTOR

### LINKED ISSUE #1368: Extract DPI normalization into an constexpr
As mentioned in  #1367 the normalization factor for the DPI is hardcoded everywhere in our codebase. It would be better to extract it into an constexpr:
```
constexpr auto const DPI_NORMALIZATION_FACTOR = 72.0;
```

Cogitri:
This spares us from hardcoding it everywhere.

fixes #1368

### DIFF
--- a/src/control/ClipboardHandler.cpp
+++ b/src/control/ClipboardHandler.cpp
@@ -1,6 +1,7 @@
 #include "ClipboardHandler.h"
 
 #include "Control.h"
+#include "Util.h"
 #include "view/DocumentView.h"
 
 #include <config.h>
@@ -203,7 +204,7 @@ bool ClipboardHandler::copy()
 
 	DocumentView view;
 
-	double dpiFactor = 1.0 / 72.0 * 300.0;
+	double dpiFactor = 1.0 / Util::DPI_NORMALIZATION_FACTOR * 300.0;
 
 	int width = selection->getWidth() * dpiFactor;
 	int height = selection->getHeight() * dpiFactor;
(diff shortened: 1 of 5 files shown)


Cogitri:
Actually, I don't think this is correct, `72.0` is only valid for a `1920x1080` screen. On my `3840x2160` screen the ruler in `Preferences->Zoom` is only correct if I set the `Display DPI Calibration` to `144`.

Cogitri:
Ah, it's not about resolution but what scaling you select.

Technius:
For reference, the scaling issue is #1472.

The `72.0` appears to be the [conversion factor from inches to points](https://en.wikipedia.org/wiki/Point_(typography)). Maybe the constant should be renamed for clarification.

Cogitri:
Ah, and that's always valid even when Xournalpp is scaled?

Technius:
Disclaimer: I don't know much about graphics/display. Feel free to correct me if I'm wrong.

From some quick reading, it seems that the points-per-inch scaling factor is independent of _pixels_-per-inch (try calculating the pixels per inch of your monitor--it will probably not be `72.0`). So yes, `72.0` points per inch is indeed a constant. But when rendering, we're concerned with _pixels_-per-inch (PPI); on a monitor, I think it's generally assumed that PPI = DPI, and the DPI scaling factor allows users to artificially increase PPI/DPI. [This article](https://wiki.gnome.org/HowDoI/HiDpi) states that "GTK+ makes hi-dpi work transparently in most places, but sometimes it may be necessary to do some extra work for best results," which I think we're seeing with the scaling issue.

Cogitri:
Ah, alright, thank you for the clarification! I don' t think this needs to be renamed then

LittleHuba:
I pointed out the file and lines where you need to take the scaling into account in the issue. This should be independent from the DPI.

Cogitri:
Thanks :)

I guess this is good to be merge as-is then?

PR #1469, opened 2019-09-13
Enabled newlines in Latex

nicolae-stroncea:
User can now use \newline,\\\,  etc, to write on a new line in Latex Mode. Previously, all text would go in 1 single line and fill the entire width, so user would have to start a new TexMode if equation was getting too long

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -22,7 +22,7 @@
  * the rendered formula is blank. Otherwise, a completely blank, sizeless PDF
  * will be generated, which Poppler will be unable to load.
  */
-const char* LATEX_TEMPLATE_1 = R"(\documentclass[crop, border=5pt]{standalone})"
+const char* LATEX_TEMPLATE_1 = R"(\documentclass[varwidth=true, crop, border=5pt]{standalone})"
                                "\n"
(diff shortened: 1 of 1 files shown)


Technius:
Thanks for the PR, but we're currently working on getting 1.0.14 released so we can't merge this PR until afterwards.

nicolae-stroncea:
Do you have a timeline in mind for this? I was looking into other minor bug fixes/ or adding other features. Is it better to wait for 1.0.14 to be released before I proceed with this?

LittleHuba:
The problem is that part of 1.0.14 is applying clang-format and some other changes to the code which apply to high percentages of the files. To preserve our git history and be able to check where a bug originated we want to do this in one go and not spread it.
So you should definitely wait since most likely your PRs you do now will not be mergable after these changes without a lot of manual modifications.

We hope to be finished with all this work within the week, but we have to make sure that a few changes don't mess up the systems of users. Especially one change in this release has the potential to break a lot of desktop environments so we might need some more time.

LittleHuba:
@nicolae-stroncea we will lift the PR-freeze shortly but we are not quite done yet. Please check out the main github page, in a week or so. We will remove the message about the PR-freeze as soon as we are done. Thanks in advance for all the work!

PR #1476, opened 2019-09-17
PenInputHandler: fix input with pen if display is scaled

### LINKED ISSUE #1472: Pen starts drawing at wrong position
**Affects versions :**
 - OS: Alpine Linux Edge (same with Flatpak), on GNOME 3.34 Wayland
 - Version of Xournal++ : 580ce2c79b8ca3ec35978bb82aa23d024790eb72

**Describe the bug**
The pen's starting point for drawing is incorrect if I use it with my stylus, it works fine with my mouse. The stylus (a HP Tilt Pen) works fine in xournal.

**Screenshots of Problem**
I touch the screen with the stylus in the black circle, but the actual drawing is in the left corner.

Screenshot from 2019-09-14 15-50-46

[debug_gdk_input.log](https://github.com/xournalpp/xournalpp/files/3612759/debug_gdk_input.log)
[debug_input.log](https://github.com/xournalpp/xournalpp/files/3612760/debug_input.log)

Cogitri:
fixes #1472

### DIFF
--- a/src/gui/inputdevices/PenInputHandler.cpp
+++ b/src/gui/inputdevices/PenInputHandler.cpp
@@ -29,6 +29,40 @@ PenInputHandler::~PenInputHandler()
 	XOJ_RELEASE_TYPE(PenInputHandler);
 }
 
+/**
+ * Get input data relative to current input page
+ */
+PositionInputData PenInputHandler::getInputDataRelativeToCurrentPage(XojPageView* page, InputEvent* event)
+{
+    XOJ_CHECK_TYPE(PenInputHandler);
+
+    GtkXournal* xournal = inputContext->getXournal();
+
+    gdouble eventX = event->relativeX;
+    gdouble eventY = event->relativeY;
+
+    gint scale = gtk_widget_get_scale_factor(&xournal->widget);
+
+    //take scroll offset into account
+    this->inputContext->getScrollHandling()->translate(eventX, eventY);
+
+    PositionInputData pos = {};
+    pos.x = eventX - page->getX() - xournal->x;
+    pos.y = eventY - page->getY() - xournal->y;
+    pos.pressure = Point::NO_PRESSURE;
+
+    if (this->inputContext->getSettings()->isPressureSensitivity())
+    {
+        pos.pressure = event->pressure;
+    }
+
+    pos.state = this->inputContext->getModifierState();
+    pos.timestamp = event->timestamp;
+
+    return pos;
+}
+
+
 void PenInputHandler::updateLastEvent(InputEvent* event)
 {
 	XOJ_CHECK_TYPE(PenInputHandler);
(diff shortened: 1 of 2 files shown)


src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -29,6 +29,40 @@ PenInputHandler::~PenInputHandler()
...
+    gdouble eventY = event->relativeY;
+
+    gint scale = gtk_widget_get_scale_factor(&xournal->widget);
+
+    //take scroll offset into account
+    this->inputContext->getScrollHandling()->translate(eventX, eventY);
+
+    PositionInputData pos = {};
+    pos.x = eventX - page->getX() - xournal->x;
+    pos.y = eventY - page->getY() - xournal->y;
```
Cogitri: Hm, multiplying this with the scale doesn't seem to fix it either... :thinking: 

It seems to me like the relative movement is fine, but the starting point is wrong

LittleHuba: Just output these values to the command line and check which are wrong. Most likely you need to only multiply `eventX` and `eventY` and not the whole statement.

Cogitri: Did that, but multiplying either of them with `scale` will result in the line getting longer than what I actually drew with my pen. I think the relative offset (which exentX and eventY are) are fine, but the starting point is wrong.

LittleHuba: Output all values of these lines to the command line and check if anything seems wrong. Maybe it is another value that is actually not set correctly for some reason.

LittleHuba: It could be `xournal->x` and `xournal->y`

LittleHuba: With no way to test I'm more or less gessing...

Cogitri: So I've been tinkering about a bit, I can't really seem to find a pattern in how it differs:

```
Correct output: 
eventX:463.219 pageX17 XournalX0
eventY:314.023 pageX10 XournalX0

Output with pen (2x scale):
eventX:122.215 pageX17 XournalX0
eventY:76.625 pageX10 XournalX0
```

Done with ` std::cout << "eventX:" << eventX << " pageX" << page->getX() << " XournalX" << xournal->x << "\n";`

LittleHuba: You could try saving the initial starting point and adding that with some scale on top of all further calculations. It is somewhat ugly but at least it would add the constant offset you seem to require.

LittleHuba:
@Cogitri is this the final fix now or does it need further work?

Cogitri:
Well, it doesn't work and it only happens on Wayland, so I think this is something we can't really fix in Xournalpp but needs to be done in GTK :/

LittleHuba:
Please create a ticket over at their repo

Cogitri:
Will do

-------- Original Message --------

PR #1482, opened 2019-09-19
Add `make` to list of Fedora dependencies

WillNilges:
I needed to install it before building Xournalpp

### DIFF
--- a/readme/LinuxBuild.md
+++ b/readme/LinuxBuild.md
@@ -14,7 +14,7 @@ poppler-glib texlive-bin texlive-pictures gettext libzip
 
 ### For Fedora/CentOS/RHEL:
 ```bash
-sudo dnf install gcc-c++ cmake gtk3-devel libxml2-devel cppunit-devel portaudio-devel libsndfile-devel \
+sudo dnf install make gcc-c++ cmake gtk3-devel libxml2-devel cppunit-devel portaudio-devel libsndfile-devel \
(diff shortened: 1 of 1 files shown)


Technius:
The installation instructions in the readme don't assume any specific build tool (other than CMake), but they do require `make`, `ninja`, or another supported CMake generator. I think it's better to have a small note indicating that such a tool must be installed instead of specifying `make` as a hard dependency.

WillNilges:
> The installation instructions in the readme don't assume any specific build tool (other than CMake), but they do require `make`, `ninja`, or another supported CMake generator. I think it's better to have a small note indicating that such a tool must be installed instead of specifying `make` as a hard dependency.

That'd probably be better tbh.

readme/LinuxBuild.md
```diff
@@ -6,6 +6,8 @@ Please create pull requests (or file issues) if you have more precise dependenci
 
 Lua is needed for plugins, if it is missing, the plugins will be disabled.
 
+The installation instructions don't assume any specific build tool (other than CMake), but they do require make, ninja, or another supported CMake generator. It is required that such a tool is installed in order to build xournalpp.
```
Technius: Very minor nitpick, but could you break this line into multiple lines with less than 80 (or 90) characters to be consistent with the rest of the file?

Technius:
LGTM besides the minor nitpick (see above).

LittleHuba:
Add notice that a tool like make is required for manual build to instructions

PR #1492, opened 2019-09-23
improved const correctness in Settings, probably fixes some segfaults.

### LINKED ISSUE #1491: Crash on opening files.
**Affects versions :**
 - OS: [Linux Mint 19.2 Cinnamon]
 - Version of Xournal++ [current master]

**Describe the bug**
Pressing the open button crashes the software

**To Reproduce**
compile with clang (in DEBUG Mode), open xournal++ try to open a file

**Expected behavior**
Do not crash.

fixed it already in a local branch, all `std::strings` in `Settings` are returned by value thus they only live after the line if not assigned. There was a .c_str() call on a temporary which is destructed immediately. 
Replaced all returns of a `std::string` by value in `Settings` with a `string const&`.
PR following.

Febbe:
Fixes #1491

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -7,7 +7,7 @@
 #include <i18n.h>
 #include <Util.h>
 #include <util/DeviceListHelper.h>
-
+#include <utility>
 #define DEFAULT_FONT "Sans"
 #define DEFAULT_FONT_SIZE 12
 
@@ -20,7 +20,7 @@
 const char* BUTTON_NAMES[] = {"middle", "right", "eraser", "touch", "default", "stylus", "stylus2"};
 
 Settings::Settings(Path filename)
- : filename(filename)
+ : filename(std::move(filename))
 {
 	XOJ_INIT_TYPE(Settings);
 	loadDefault();
@@ -374,7 +374,7 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur)
 	}
 	else if (xmlStrcmp(name, (const xmlChar*) "sidebarWidth") == 0)
 	{
-		this->sidebarWidth = g_ascii_strtoll((const char*) value, NULL, 10);
+		this->sidebarWidth = std::max<int>(g_ascii_strtoll((const char*) value, NULL, 10), 50);
 	}
 	else if (xmlStrcmp(name, (const xmlChar*) "sidebarOnRight") == 0)
 	{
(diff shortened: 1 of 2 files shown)


Febbe:
It's a regression of #1112 . Already fixed and merged it in another PR. This is the refactoring PR now.

src/control/settings/Settings.cpp
```diff
@@ -2005,20 +2005,17 @@ void Settings::setSidebarVisible(bool visible)
...
+int Settings::getSidebarWidth() const
 {
 	XOJ_CHECK_TYPE(Settings);
 
-	if (this->sidebarWidth < 50)
-	{
-		setSidebarWidth(150);
-	}
-	return this->sidebarWidth;
+	return this->sidebarWidth < 50 ? 150 : this->sidebarWidth;
```
LittleHuba: Is the behaviour of this method changed by accident or was this done with purpose? It saves against buggy settings files which may be out there.

Febbe: It was indeed done by purpose. The check should happen on write, not on read. I changed both functions.

LittleHuba: The thing is that if there is a buggy settings file out there then it will create an invalid state for the Xournal++ instance even if writing the file back in a correct way. You should definitely check this when reading the settings file. But it might be better to place this in the constructor than in the `getSidebarWidth()` method as it is a one-time fix.

Febbe: Or I have to be Shure, that this field is only written by set...(value).

LittleHuba: We are talking of files written with prior versions of Xournal++. Backwards-compatibility :wink:

Febbe: Year I know what you mean, I moved the check to the setter, but if the setter isn't used to initialize the value on the load of old files, I need to make Shure it does, or use the check there also.

Febbe: I am back home in 10 minutes, then I can check everything

src/control/settings/Settings.cpp
```diff
@@ -2005,20 +2006,17 @@ void Settings::setSidebarVisible(bool visible)
...
 }
 
 void Settings::setSidebarWidth(int width)
 {
 	XOJ_CHECK_TYPE(Settings);
+	width = width < 50 ? 150 : width;
```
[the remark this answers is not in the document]
Febbe: Done

src/control/settings/Settings.cpp
```diff
@@ -374,7 +374,7 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur)
 	}
 	else if (xmlStrcmp(name, (const xmlChar*) "sidebarWidth") == 0)
 	{
-		this->sidebarWidth = g_ascii_strtoll((const char*) value, NULL, 10);
+		this->sidebarWidth = std::max(g_ascii_strtoll((const char*) value, NULL, 10), 50L);
```
LittleHuba: you forgot to include the math header :wink:

Febbe: the header was somehow included already, over other headerfiles

PR #1493, opened 2019-09-23
Fix#1081

Febbe:


Technius:
Can you cherry pick the bug fix from the refactoring and move the refactoring into a separate PR?

Febbe:
Nearly every line I changed could be the problem, there is a mutex which isn't unlocked after an exception is thrown or before an return.

JJones780:
The error was in /src/control/jobs/CustomExportJob.cpp:118 which returned without calling doc->unlock()  when the dialog was cancelled. 
Add  `doc->unlock()` before line 118 should be all that's needed to fix bug #1493. 


What's the comment added in XojPdfDocument& Document::getPdfDocument() all about?
`//Todo: the locking is broken, fix it`
Any more details???

src/model/Document.h
```diff
@@ -88,7 +88,7 @@ class Document
 
 	void lock();
 	void unlock();
-	bool tryLock();
+	bool try_lock();
```
JJones780: try to avoid making personal preference changes in code you otherwise aren't changing.

src/model/Document.cpp
```diff
@@ -537,6 +534,7 @@ XojPdfDocument& Document::getPdfDocument()
 	return this->pdfDocument;
 }
 
+//Todo: the locking is broken, fix it
```
JJones780: details?

src/control/PageBackgroundChangeController.cpp
```diff
@@ -143,10 +143,10 @@ bool PageBackgroundChangeController::applyImageBackground(PageRef page)
 
 	Document* doc = control->getDocument();
 
-	doc->lock();
-	ImagesDialog dlg(control->getGladeSearchPath(), doc, control->getSettings());
-	doc->unlock();
-
+	auto dlg = [this, doc] {
+		std::lock_guard<Document> _{*doc};
```
JJones780: the underscore seems unnecessarily minimal... why not use "guard" as elsewhere or even something more descriptive such as "scopedGuard" since we aren't manually calling lock and unlock.

src/control/PageBackgroundChangeController.cpp
```diff
@@ -230,10 +230,10 @@ bool PageBackgroundChangeController::applyPdfBackground(PageRef page)
 		return false;
 	}
 
-	doc->lock();
-	PdfPagesDialog* dlg = new PdfPagesDialog(control->getGladeSearchPath(), doc, control->getSettings());
-	doc->unlock();
-
+	auto* dlg = [this, doc] {
```
JJones780: Why the lambda? Wouldn't a code block suffice?
Actually, it might be better to do a manual unlock ( and lock ?) here ( and similar cases ) for readability instead of relying on scope.

src/control/jobs/CustomExportJob.cpp
```diff
@@ -152,11 +151,11 @@ void CustomExportJob::run()
...
-		doc->lock();
-		h.prepareSave(doc);
-		h.saveTo(filename, this->control);
-		doc->unlock();
-
+		{
+			std::lock_guard<Document> _{*doc};
+			h.prepareSave(doc);
+			h.saveTo(filename, this->control);
+		}
```
JJones780: consider:   do a manual unlock instead of scope code block "side effect".

Technius: Manual lock/unlock is error-prone and should be avoided. The scoped code block follows RAII and is idiomatic (i.e. the lock the automatically released when it goes out of scope.

src/control/PageBackgroundChangeController.cpp
Technius: Looks like the lambda is used so that `dlg` can be assigned. This looks fine for readability imo.

Technius:
I would prefer adding a commit that inserts the `doc->lock()` as mentioned by @JJones780 and then making the refactor a followup commit or PR. This will lead to a cleaner commit history, and it would be easier to rollback in case the refactoring causes a regression.

src/control/PageBackgroundChangeController.cpp
JJones780: There's no reason not to simply call unlock() before the dlg->show().  
When going out of scope it won't unlock again.
No need for a lambda.

Technius: Manually locking and unlocking is error-prone (as you can see from the bug) and should be avoided if possible.

src/model/Document.cpp
Febbe: The problem is, that the lock and the unlock is not in the same scope. And this sort of synchronization handling is bug prone and unsafe to exceptions or code changes. So the hole part should be refactored some time.

src/control/PageBackgroundChangeController.cpp
Febbe: was very lazy :D, will change it.

src/control/PageBackgroundChangeController.cpp
Febbe: As @Technius pointed out, the locking / unlocking should only done over an RAII Interface like std::unique_lock, std::lock_guard or std::shared_lock. Without a RAII interface, the locking can break because of any exception or added return without unlocking. 
The lambda is indeed to assign a the variable without leaving it unassigned or setting it to nullptr before. It's like a returning scope block which does not exist in c++.

src/control/jobs/CustomExportJob.cpp
Febbe: Mmh, you think someone could not understand the purpose of this block and wants to remove it. Forgetting about the effect of the std::lock_guard? On the other hand that's exactly the purpose of std::lock_guard.

PR #1495, opened 2019-09-24
Fix toolbar manage dialog issues

Technius:
* Fixed a regression (caused by #1283) in the toolbar manage dialog causing the `View > Toolbars > Manage` menu option to crash the application
* Fixed the toolbar manage dialog icon having the wrong path due to the AppStream id refactor


Technius:
Merging after 24 hours if there are no objections.

Issue #1507, opened 2019-09-28
Undefined behaviour and heap-buffer-overflow


Febbe:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Linux Mint 19.03
 - Display Protocol X11
 - 3.22.x
 - Version: current master

**Describe the bug**
Undefined behaviour and crash when using two fingers or finger and stylus at the same time, or stylus(pen) and then finger(handtool)

**To Reproduce**
1. compile with CXXFLAGS=-stdlib=libstdc++ -g -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer
2. open xournalpp over console 
3. do the above described actions

**Expected behavior**
No undefined behaviour, no crash, no heap-buffer-overflow

**Additional context**
```
Starting program: /home/febbe/ClionProjects/xournalpp/cmake-build-sanitizers/src/xournalpp ..
warning: the debug information found in "/lib64/ld-2.23.so" does not match "/lib64/ld-linux-x86-64.so.2" (CRC mismatch).

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

** (xournalpp:23404): WARNING **: Unsupported input method: xim, changed to: ibus
[New Thread 0x7fffe3f34700 (LWP 23405)]
[New Thread 0x7fffe3733700 (LWP 23406)]
ALSA lib pcm_dsnoop.c:606:(snd_pcm_dsnoop_open) unable to open slave
ALSA lib pcm_dmix.c:1029:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2266:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
[New Thread 0x7fffdd802700 (LWP 23407)]
[Thread 0x7fffdd802700 (LWP 23407) exited]
[New Thread 0x7fffdd802700 (LWP 23408)]
[Thread 0x7fffdd802700 (LWP 23408) exited]
ALSA lib pcm_dmix.c:1029:(snd_pcm_dmix_open) unable to open slave
[New Thread 0x7fffdd802700 (LWP 23409)]
[Thread 0x7fffdd802700 (LWP 23409) exited]
[...]

PR #1511, opened 2019-10-03
Fix undefined behaviour

Febbe:
Replaced a plain old dynamic allocated C-array with a std::vector

### DIFF
--- a/src/model/Stroke.cpp
+++ b/src/model/Stroke.cpp
@@ -6,6 +6,7 @@
 #include <i18n.h>
 
 #include <cmath>
+#include <numeric>
 
 Stroke::Stroke()
  : AudioElement(ELEMENT_STROKE)
@@ -16,11 +17,6 @@ Stroke::Stroke()
 Stroke::~Stroke()
 {
 	XOJ_CHECK_TYPE(Stroke);
-
-	g_free(this->points);
-	this->points = NULL;
-	this->pointCount = 0;
-	this->pointAllocCount = 0;
 
 	XOJ_RELEASE_TYPE(Stroke);
 }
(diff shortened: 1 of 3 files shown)


src/model/Stroke.cpp
```diff
@@ -450,13 +411,13 @@ void Stroke::setPressure(const vector<double>& pressure)
 	XOJ_CHECK_TYPE(Stroke);
 
 	// The last pressure is not used - as there is no line drawn from this point
-	if (this->pointCount - 1 > (int)pressure.size())
+	if (this->points.size() - 1 != pressure.size())
```
Technius: Don't see any reason to change the condition. If we read a damaged file that has a mismatching pressure file, we should recover at least some of the pressure data instead of discarding all of it. (Yes, the previous condition refuses to load pressure information when the given data is larger than the number of points, but that could be improved by loading whatever can be loaded and then emitting a warning).

src/model/Stroke.cpp
```diff
@@ -355,7 +318,7 @@ void Stroke::rotate(double x0, double y0, double xo, double yo, double th)
 		p.y += y0;
 
 		p.x += xo-offset;	//center it
-		p.y += yo-offset;		
+		p.y += yo - offset;
```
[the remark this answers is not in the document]
Febbe: yes, thats the problem with clang-format, if i change it to the previous format, the check will fail. I could change this line also. Or we ignore it, since the format will be applied with the PR freeze.

src/model/Stroke.cpp
Febbe: mmh ok see your point. I'll look if I can make an even better decision. Or just revert that change

Febbe:
I'll extract the second commit into a new PR, since it is very intrusive, because it force-copies an event.
As we know from #1507 the heap buffer overflow is triggered by copying an wrong event.
But the current version is definitely wrong. It can lead to double deletion, because an uncopied event is dereferenced in the destructor of InputEvent.
I would rather remove the call to gdk_event_free().

Issue #1458, opened 2019-09-07
Xournal++ crashes when audio finishes playing


nicolae-stroncea:
**Affects versions :**
 - OS:  Fedora 30 Gnome Wayland And X11(occurs on both), Built and compiled
 - Version of Xournal++ : Most recent one, seems to be master on commi 6260dba

**Describe the bug**
After recording, when I click Play object, once the audio ends, it crashes horribly, in one case froze on the GUI and needed to restart PC.

**To Reproduce**
1. Step 1: Build and Compile xournalpp on Fedora 30, package as a .tar.gz
2. Step 2: Set audio input and output to pulse, Record, and then play audio.
3. Step 3: Once audio finishes playing, it will crash

**Expected behavior**
Not to crash
[errorlog.20190906-231602.log](https://github.com/xournalpp/xournalpp/files/3586213/errorlog.20190906-231602.log)

EDIT: To clarify about what I meant when I said: once the audio ends: It seems to occur both when general audio finishes playing, as well as when audio finishes playing for a particular object

LittleHuba:
If you run Xournal++ from a console do you get some additional output when it crashes?
(It is normal that there are some lines written when you start Xournal++ - so these do not count)

nicolae-stroncea:
Date: Wed Sep 18 16:32:58 2019
Error: signal 6
[bt]: (0) ./xournalpp() [0x769c50]
[bt]: (1) /lib64/libc.so.6(+0x37ec0) [0x7f221ece7ec0]
[bt]: (2) /lib64/libc.so.6(gsignal+0x145) [0x7f221ece7e35]
[bt]: (3) /lib64/libc.so.6(abort+0x127) [0x7f221ecd2895]
[bt]: (4) /lib64/libc.so.6(+0x22769) [0x7f221ecd2769]
[bt]: (5) /lib64/libc.so.6(+0x30566) [0x7f221ece0566]
[bt]: (6) /lib64/libportaudio.so.2(+0xe105) [0x7f221f35c105]
[bt]: (7) /lib64/libportaudio.so.2(+0x13ff3) [0x7f221f361ff3]
[bt]: (8) /lib64/libportaudio.so.2(+0x146c3) [0x7f221f3626c3]
[bt]: (9) /lib64/libpthread.so.0(+0x84c0) [0x7f221f3994c0]
[bt]: (10) /lib64/libc.so.6(clone+0x43) [0x7f221edac553]


Try to get a better stracktrace...
[...]

PR #1512, opened 2019-10-03
circular array based audio queue

morrolinux:
It basically worked on the first attempt, so I wasn't really spurred to do better anything in particular.
Please feel free to point out if you find anything stupid ;-)

### DIFF
--- a/src/util/audio/AudioQueue.h
+++ b/src/util/audio/AudioQueue.h
@@ -19,12 +19,16 @@
 #include <condition_variable>
 
 template <typename T>
-class AudioQueue : protected std::deque<T>
+class AudioQueue
 {
 public:
 	AudioQueue()
 	{
 		XOJ_INIT_TYPE(AudioQueue);
+
+		unsigned int size = 1024 * 200;
+		buf = (std::unique_ptr<T[]>(new T[size]));
+		buf_max_size = size;
 	}
 
 	~AudioQueue()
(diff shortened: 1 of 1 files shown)


Technius:
For future reference, this appears to be a fix for #1458.

LittleHuba:
@morrolinux is this included in #1520?

Febbe:
There are several things I don't like. 
I would rather use a std::array instead of a plain old c-array.
The objects in the queue aren't destructed, only overridden.
You don't use rangebased for loops.
You use some sort of pointer arithmetic.
The ringbuffer is still not accessed in a thread safe way. The bug could still happen, just not as often than before.

I would rather use the version from before with an write/shared read mutex and/or boost circular buffer.

morrolinux:
> @morrolinux is this included in #1520?

I must have branched out #1520 from this so that's not really intentional.. I'm gonna remove it from #1520 now.

> I would rather use the version from before with an write/shared read mutex and maybe with boost circular buffer.

Yeah sure no problem. As I said, I just needed something working ASAP since I'm very busy with classes right now. You can implement it your way, I like it more as well, no offence is taken :)

morrolinux:
> The ringbuffer is still not accessed in a thread safe way. The bug could still happen, just not as often than before.

I'm not sure about that, becuase the reader only reads and the writer advance the pointer once the write has finished, and no resources are accessed for write by two different threads. Or am I missing something?

LittleHuba:
Ringbuffers in itself are not necessarily thread-safe, but this specific implementation in this usecase should be thread-safe. The important part is that we only have one writer and one reader, and size is incremented after the buffer is written.

So in essence this is a valid solution, but all other problems mentioned by @Febbe stand and therefore I would stick to the Deque. Additionally I'm still not sure if a fixed buffer size is a good thing for all hardwares. I had some problems when I wrote this up since my desktop PC was way faster than my laptop.

Still thanks for the PR @morrolinux

PR #1520, opened 2019-10-06
Audio playback seek feature

morrolinux:
Since when taking notes we normally write *after* hearing something interesting, just seeking to the exact timestamp a stroke/text was created might just not be enough, so I've added seek forwards and backwards buttons to the playback bar, so that one can make better use of the feature.


morrolinux:
Screenshot attached 
image

LittleHuba:
LGTM

morrolinux:
Should I wait until the PR freeze is over before merging? 
What's with the Clang format error? I don't understand what's wrong with the test

LittleHuba:
You should most likely wait, just to make the transition smooth.
The Clang Format Error checks the code for right formatting to decrease the merge issues because formats are different. It is currently hard to get this check to signal green as the overall code is not formatted. Part of the PR Freeze is to fromat the code which will then in turn make this check useful. For now you can just ignore it.

LittleHuba:
@morrolinux please rebase on master to make this ready for merging.

morrolinux:
@LittleHuba 583 files changed... I'm having a bad feeling about this rebase...

LittleHuba:
Did you not force-push?

morrolinux:
> Did you not force-push?

What I did:
1) rebased my fork's master branch from upstream
2) rebased my feature branch from master
3) fixed conflicts, commit and pushed changes

Now, if I correctly understood what rebasing means, it should be allright, but then, how come that making a PR against upstream master shows all those files changed?
That's the bit I don't understand.

Should I redo it and just try merging my rebased master into my feature branch?

LittleHuba:
That is because you did most likely not force-push meaning you created merge commits for all your rebased commits. These are slightly different from the original commits and show therefore on github as changes. Fortunately fixing this is not really hard. You would need to reset your HEAD to before the rebase commits (34e6530), do the rebase locally again and then do the force-push (the force is here very important).

The command should be:
```
git reset --hard 34e6530
git rebase master
# Solve all conflicts
git push --force
```

morrolinux:
Alright, `AudioQueue.h` is unfortunately in the PR but as far as I can see the changes should match the master branch, does it look alright to you?

Is it normal for the tests to fail?

LittleHuba:
No the tests should definitely not fail. Please check out the test details to review why compilation failed.

src/control/AudioController.cpp
```diff
@@ -100,6 +100,21 @@ void AudioController::pausePlayback()
 	this->audioPlayer->pause();
 }
 
+void AudioController::seekForwards()
+{
+	XOJ_CHECK_TYPE(AudioController);
```
LittleHuba: Remove all XOJ checks as they are no longer valid

src/util/audio/VorbisProducer.cpp
```diff
@@ -36,23 +38,33 @@ bool VorbisProducer::start(std::string filename, unsigned int timestamp)
 			{
 				size_t numFrames = 1;
 				auto sampleBuffer = new float[1024 * this->sfInfo.channels];
-				std::unique_lock<std::mutex> lock(audioQueue->syncMutex());
-
-				while (!this->stopProducer && numFrames > 0 && !this->audioQueue->hasStreamEnded())
```
LittleHuba: You acidentially took the wrong choice in your merge here. You need both the lock and should use `numFrames`.

src/util/audio/VorbisProducer.cpp
```diff
@@ -36,23 +38,33 @@ bool VorbisProducer::start(std::string filename, unsigned int timestamp)
...
-					this->audioQueue->push(sampleBuffer, static_cast<size_t>(numFrames * this->sfInfo.channels));
+					if (this->seekSeconds != 0)
+					{
+						tot += seekSeconds * this->sfInfo.samplerate;
+						if (tot < 0) tot = 0;
+						sf_seek(this->sfFile, tot, SEEK_SET);
+						this->seekSeconds = 0;
+					}
+
+					this->audioQueue->push(sampleBuffer, static_cast<unsigned long>(numSamples * this->sfInfo.channels));
```
LittleHuba: Use `size_t` in the static cast.

Febbe:
If you have problems using the build in merge, you can specify a mergetool like meld. If you call git mergetool it will be opened instead of the console derivate. And you can try to safe your progress, by copiing your branch. By doing that, you'll have infinite tries to do it right.

morrolinux:
@Febbe Thanks for the suggestion and @LittleHuba Thanks for the review! I've fixed all of the above, and included the default audio seek setting. 
Everything seems fine, except for the Clang test for which I can't see what's wrong: pressing "Details" button only leads me to a bunch of code with no specific error. I'm not a big expert of Clang formatting myself so how should I interpret this? what to look for?

LittleHuba:
You're more then welcome!
The clang check is currently optional since the codebase is not yet formatted and therefore it does not really make sense yet. It's just that we don't get formatting related merge conflicts in the future. Once the whole codebase is formatted PRs can be formatted with a single command before merging. At that point we will make it mandatory.

LittleHuba:
By the way the bunch of code you see is a fit diff between your unformatted changes and the correct format.

morrolinux:
Aaah, ok, now I get it. I went to the code to fix the formatting and noticed that the codebase itself is indeed not coherent yet. Should I proceed merging then?

ui/settings.glade
```diff
@@ -3923,6 +3930,71 @@ If available select &lt;small&gt;&lt;tt&gt;pulse&lt;/tt&gt;&lt;/small&gt; as inp
...
+                                      <object class="GtkGrid" id="sid3">
+                                        <property name="visible">True</property>
+                                        <property name="can_focus">False</property>
+                                        <property name="row_spacing">10</property>
+                                        <property name="column_spacing">10</property>
+                                        <child>
+                                          <object class="GtkLabel" id="label3">
+                                            <property name="visible">True</property>
+                                            <property name="can_focus">False</property>
+                                            <property name="label" translatable="yes">Default Seek Time</property>
```
LittleHuba: I'm not really happy with this label. Better would be `Seeking offset in seconds` or along these lines.

src/util/audio/VorbisProducer.cpp
```diff
@@ -32,21 +33,32 @@ bool VorbisProducer::start(std::string filename, unsigned int timestamp)
...
 					while (this->audioQueue->size() >= this->sample_buffer_size && !this->audioQueue->hasStreamEnded() && !this->stopProducer)
 					{
 						audioQueue->waitForConsumer(lock);
 					}
 
+					if (this->seekSeconds != 0)
+					{
+						position += seekSeconds * this->sfInfo.samplerate;
+						if (position < 0) position = 0;
+						sf_seek(this->sfFile, position, SEEK_SET);
```
LittleHuba: Using `SEEK_CUR` you can just apply the offset which is then used from the current playback position. That saves you these awkward position calculations all throughout the code.

ui/settings.glade
LittleHuba: Am I right that this only modifies the offset applied when pressing the seek buttons?

morrolinux: Indeed, that option only modifies the offset applied when pressing the seek buttons.
I thought I'd be more appropriate to keep it separate from the actual track seeking when clicking on a stroke. Maybe I could add another configuration setting for that, but as of now I'm quite happy with it

src/util/audio/VorbisProducer.cpp
morrolinux: Damn it really was that simple? I'm gonna try this out when I get home.

LittleHuba: Yeah :joy: 
http://www.mega-nerd.com/libsndfile/api.html#seek

src/util/audio/VorbisProducer.cpp
```diff
@@ -32,21 +33,27 @@ bool VorbisProducer::start(std::string filename, unsigned int timestamp)
 	this->audioQueue->setAudioAttributes(this->sfInfo.samplerate, static_cast<unsigned int>(this->sfInfo.channels));
 
 	this->producerThread = new std::thread(
-			[&, filename]
+			[&, filename, seekPosition]
```
LittleHuba: This can be removed again as it is now unused in the lambda.

src/util/audio/VorbisProducer.cpp
```diff
@@ -32,21 +33,27 @@ bool VorbisProducer::start(std::string filename, unsigned int timestamp)
...
 
 					while (this->audioQueue->size() >= this->sample_buffer_size && !this->audioQueue->hasStreamEnded() && !this->stopProducer)
 					{
 						audioQueue->waitForConsumer(lock);
 					}
 
+					if (this->seekSeconds != 0)
+					{
+						sf_seek(this->sfFile,seekSeconds * this->sfInfo.samplerate, SEEK_CUR);
+						this->seekSeconds = 0;
```
LittleHuba: This creates a race condition on the variable `seekSeconds` but I think we can live with it as users are unable to click nearly as fast as this poll and reset happens.

morrolinux: Agreed. The worst that can happen is the program skipping a seek command if the user can click twice in a few nanoseconds (which I'm not even sure it's possibile to catch on a non real-time O.S.) so I wouldn't complicate it with synchronization constructs

PR #1526, opened 2019-10-08
Update version to 1.0.14

Technius:
Version 1.0.14 release. **Do not merge this PR; I will merge it.**

Please let me know if I forgot to update anything.


LittleHuba:
We should merge in all the bug fixes we have lying around before this.
At least #1511 and #1512.

Technius:
I don't think #1512 will be ready anytime soon. We can push it off to 1.1.0 or a minor release 1.0.16 with bugfix backports.

#1511 will be **the last PR to merge (if we include it), no exceptions**. We have to draw the line somewhere or else we'll never get 1.0.14 out.

LittleHuba:
Just closed #1512 since the current implementation is easy to fix and has some advantages.

I can create the PR for this fix in an hour. You can then decide if you still want to skip it for this release or include it.

PR #1528, opened 2019-10-10
Fix zoom handling

### LINKED ISSUE #404: Zooming via pinch to zoom causes graphical glitches

### LINKED ISSUE #1224: Zooming with touch gestures doesn't workt shortly after stroke
OS: Arch Linux
Version of Xournal++ : 1.0.11

After you have drawn a stroke, you can't immediately zoom. You have to scroll before and then can zoom again. 

        1. Draw a stroke
        2. Try to zoom with touch gesture
        3. Scroll with touch input and afterwards zooming is possible

### LINKED ISSUE #1286: Laggy zoom in new input system
Arch Linux 5.1
Xournal++ 1.0.13

Zooming is laggy. This doesn't occur in the old input system.

1. Turn on the new input system in the preferences
2. Zoom in or out

LittleHuba:
Originally PR #1460 by @taaem.
Moved to local branch for reworking of algorithm to fix two problems:

- [x] natural zooming (same point of page stays under the finger while zooming)
- [x] scrolling while zooming feels natural (no jumping)

Fixes #404, Fixes #1224, Fixes #1286


taaem:
Just tested your new algorithm and for me it works pretty well!

LittleHuba:
The zooming is not 100% natural but I can't find the reason for the slight shift. It seems to zoom sligthly faster than the fingers move...

I'd say it's good enough as it is 😉

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -7,102 +7,175 @@
...
+		this->secLastAbsX = event->absoluteX;
+		this->secLastAbsY = event->absoluteY;
 	}
-}
 
-void TouchInputHandler::actionEnd(InputEvent* event)
-{
-	XOJ_CHECK_TYPE(TouchInputHandler);
+	double sqDistance = std::pow(this->priLastAbsX - this->secLastAbsX, 2.0) + std::pow(this->priLastAbsY - this->secLastAbsY, 2.0);
+	double zoom = sqDistance / this->startZoomDistance;
```
taaem: The real distances are std::sqrt(distance) for both of these values and I think that's whats causing the non-natural zoom. I suggest to change either this line to:
```
	double zoom = std::sqrt(sqDistance / this->startZoomDistance);
```
or store both the distances directly as _real_ distances.
(length of an vector = sqrt(x²+y²))

LittleHuba: Ohh I just copied this from your approach without looking at it^^ 
The new algorithm is better but still not perfect. I have absolutely no idea where the divergence might come from.

LittleHuba: It should be good enough with the new approach. It diverges by 5mm for 10cm movement which is more or less beneath the finger anyway.

LittleHuba:
Will merge this in 24h if nobody complains.

Peetz0r:
Works for me but it's quite slow indeed. Still a huge improvement though :)

(fedora 31, gnome 3.34, wayland, thinkpad yoga 370)

LittleHuba:
The slowness is not caused by the zooming but by the rendering engine. So these fixes will not influence that part. But as we are currently working on a new fileformat which will also introduce a new rendering system we will solve these issues eventually.

PR #1529, opened 2019-10-10
Add internal locking to AudioQueue

### LINKED ISSUE #1458: Xournal++ crashes when audio finishes playing

LittleHuba:
@morrolinux Please test this on your machine

Fixes #1458

### DIFF
--- a/src/util/audio/AudioQueue.h
+++ b/src/util/audio/AudioQueue.h
@@ -16,6 +16,7 @@
 #include <vector>
 #include <deque>
 #include <mutex>
+#include <cmath>
 #include <condition_variable>
 
 template <typename T>
--- a/src/util/audio/AudioQueue.h
+++ b/src/util/audio/AudioQueue.h
@@ -82,7 +82,7 @@ public:
 
 		{
 			std::lock_guard<std::mutex> lock(internalLock);
-			returnBufferLength = std::min(nSamples, std::deque<T>::size() - std::deque<T>::size() % this->channels);
+			returnBufferLength = std::min(nSamples, (unsigned long)(std::deque<T>::size() - std::deque<T>::size() % this->channels));
 			for (unsigned long i = 0; i < returnBufferLength; i++)
 			{
 				returnBuffer[i] = this->back();
--- a/src/util/audio/AudioQueue.h
+++ b/src/util/audio/AudioQueue.h
@@ -16,7 +16,7 @@
 #include <vector>
 #include <deque>
 #include <mutex>
-#include <cmath>
+#include <algorithm>
 #include <condition_variable>
 
 template <typename T>
@@ -46,17 +46,17 @@ public:
 		return std::deque<T>::empty();
 	}
 
-	unsigned long size()
+	size_t size()
 	{
 		std::lock_guard<std::mutex> lock(internalLock);
 		return std::deque<T>::size();
 	}
 
-	void push(T* samples, unsigned long nSamples)
+	void push(T* samples, size_t nSamples)
 	{
 		{
 			std::lock_guard<std::mutex> lock(internalLock);
-			for (unsigned long i = 0; i < nSamples; i++)
+			for (size_t i = 0; i < nSamples; i++)
 			{
 				this->push_front(samples[i]);
 			}
@@ -68,7 +68,7 @@ public:
 		this->pushLockCondition.notify_one();
 	}
 
-	void pop(T* returnBuffer, unsigned long& returnBufferLength, unsigned long nSamples)
+	void pop(T* returnBuffer, size_t& returnBufferLength, size_t nSamples)
 	{
 		if (this->channels == 0)
 		{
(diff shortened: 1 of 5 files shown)


src/util/audio/AudioQueue.h
```diff
@@ -100,7 +100,7 @@ class AudioQueue : protected std::deque<T>
 
 		{
 			std::lock_guard<std::mutex> lock(internalLock);
-			returnBufferLength = std::min(nSamples, std::deque<T>::size() - std::deque<T>::size() % this->channels);
+			returnBufferLength = std::min(nSamples, (unsigned long)(std::deque<T>::size() - std::deque<T>::size() % this->channels));
```
Febbe: `returnBufferLength = std::min<unsigned>(nSamples, std::deque<T>::size() - std::deque<T>::size() % this->channels));` should work also. 
But I would rather change returnBufferLength to size_t.

src/util/audio/AudioQueue.h
```diff
@@ -91,11 +97,14 @@ class AudioQueue : protected std::deque<T>
...
 		}
 
-		returnBufferLength = std::min(nSamples, this->size() - this->size() % this->channels);
-		for (long i = 0; i < returnBufferLength; i++)
 		{
-			returnBuffer[i] = this->back();
-			this->pop_back();
+			std::lock_guard<std::mutex> lock(internalLock);
+			returnBufferLength = std::min(nSamples, std::deque<T>::size() - std::deque<T>::size() % this->channels);
+			for (unsigned long i = 0; i < returnBufferLength; i++)
```
Febbe: use size_t

src/util/audio/AudioQueue.h
```diff
@@ -52,23 +53,28 @@ class AudioQueue : protected std::deque<T>
...
 
 	void push(T* samples, unsigned long nSamples)
 	{
 		XOJ_CHECK_TYPE(AudioQueue);
 
-		for (unsigned long i = 0; i < nSamples; i++)
 		{
-			this->push_front(samples[i]);
+			std::lock_guard<std::mutex> lock(internalLock);
+			for (unsigned long i = 0; i < nSamples; i++)
```
Febbe: use size_t

src/util/audio/AudioQueue.h
```diff
@@ -52,23 +53,28 @@ class AudioQueue : protected std::deque<T>
 	{
 		XOJ_CHECK_TYPE(AudioQueue);
 
+		std::lock_guard<std::mutex> lock(internalLock);
 		return std::deque<T>::empty();
 	}
 
 	unsigned long size()
```
Febbe: return size_t

src/util/audio/AudioQueue.h
```diff
@@ -16,6 +16,7 @@
 #include <vector>
 #include <deque>
 #include <mutex>
+#include <cmath>
```
Febbe: is this for std::min? then I would include `<algorithm>`

Febbe:
I think it would be better, not to return internal states of the container like size() or empty(). For example the internal lock does not prevent the container size to change between a call to size() and pop(). I would only use functions like push, pop, try_pop, for_each, consume, etc.

LittleHuba:
> I think it would be better, not to return internal states of the container like size() or empty().

I'm actually against this for a good reason. These methods are inherently save by the way they are used in the code and the code is structured. 

We have a setup of exactly one producer and one consumer. 

`size`:
- If we use `size` on the producer side, we will check for an overflow. The consumer may only consume elements thus making the situation better for us even when the state is then not absolutely consistent. 
- If we check on the consumer side we check for an underflow which is either the case and we abort whatever we do or if not the producer may push new frames inbetween calling `size` and `pop` not changing the situation for the consumer.

empty:
- Using `empty` on the consumer side, we check if there currently are no samples available which might be used to pause the polling until samples are available again. When the producer pushes samples he will unblock the consumer, thus ensuring that it polls again. Should this not succeed since the notification arives between the call to `empty` and the call to `wait` the either the next push or the call to `signalEndOfStream` wakes the consumer up again.
- Using `empty` on the producer side would mean we want to make sure that the queue is empty before pushing anything to it which makes absolutely no sense at all. Even if we do this, again the same rule applies. The consumer may only consume elements, making the condition true even when we got a wrong state, again notifying the producer if it polls... I think you get what I mean.

So long story short: Changing this is only overhead and has no advantage.

morrolinux:
> @morrolinux Please test this on your machine
> 
> Fixes #1458

I'm sorry, for some odd reasons which I still have to figure out, I don't receive github emails anymore and therefore lose notifications!!
I'm gonna try this ASAP

morrolinux:
@LittleHuba Tested and works allright :)

PR #1535, opened 2019-10-11
Update WindowsBuild.md

Gansgar:
Wrong dependency is named. You need package `make` not `mingw-w64-x86_64_make`. This is correctly stated in the azure config.


Technius:
Thanks for the PR, but this is superceded by #1538.

PR #1538, opened 2019-10-12
Updating and clarifying Windows build readme

michaelschufi:
I updated the readme:
- We need `make` because PortAudio doesn't build using `mingw-w64-x86_64-make`.
- Added `lua-*` to gitignore.
- `cmake --build .` instead of `make`


Technius:
LGTM, thanks for the PR.

PR #1541, opened 2019-10-14
Improve format_diff_lines.pl utility script

Technius:
* Now autodetects whether clang-format is called `clang-format` or `clang-format-8`
* Fixes an issue where irrelevant lines are formatted due to changed lines being formatted one-at-a-time. This should make the CI format checks way more reliable.

Merging in 3 days if there are no objections. A `sample.cpp` file is provided to demonstrate the changes; it will be removed in a rebase before merging. You can try out the new formatting logic by messing up the formatting on specific lines and then running
```bash
git --no-pager diff -U0 -- '*.cpp' | perl azure-pipelines/util/format_diff_lines.pl
````

### DIFF
--- a/azure-pipelines/util/format_diff_lines.pl
+++ b/azure-pipelines/util/format_diff_lines.pl
@@ -13,6 +13,20 @@ my @hunks;
 # Current file
 my $current_file;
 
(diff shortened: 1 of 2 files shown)


LittleHuba:
LGTM

Issue #1532, opened 2019-10-11
Todo: Remove or change ArrayIterator<T>


Febbe:
The Array Iterator does not work with algorithms of the STL, therefore it is mostly useless. It can also be a source for errors and increases overhead.
Since normal pointers are also Iterators, I would prefer to use pointers with Iterator functions of the STL like std::next, std::prev, std::advance. Or even better replace c-arrays with std::array<T,size_t> or std::vector<T> and use their build in iterators.

Those also enable range based for loops, eg.:
for(auto&& var : container){...}

MrMallIronmaker:
I was looking for a good first issue and saw this. I'm wondering if this issue should this have been closed by PR #1545? Looks like the requested changes were made.

LittleHuba:
> should this have been closed by PR #1545?

Yeah thanks for the heads up! Hope to see your PR fixing another issue soon!

PR #1545, opened 2019-10-15
Remove Arrayiterator

st4ll1:
Issue #1532


src/control/ToolHandler.cpp
```diff
@@ -455,11 +455,18 @@ void ToolHandler::setColorFound()
...
 }
 
-ArrayIterator<Tool*> ToolHandler::iterator()
+std::array<Tool*, TOOL_COUNT>::const_iterator ToolHandler::begin()
 {
-	XOJ_CHECK_TYPE(ToolHandler);
+    XOJ_CHECK_TYPE(ToolHandler);
 
-	return ArrayIterator<Tool*> (tools, TOOL_COUNT);
+    return tools.begin();
```
Febbe: in general, use `begin(container)` and `end(container)`.

src/model/Stroke.h
```diff
@@ -72,8 +70,9 @@ class Stroke : public AudioElement
 	void setLastPoint(const Point& p);
 	int getPointCount() const;
 	void freeUnusedPointItems();
-	ArrayIterator<Point> pointIterator() const;
-	Point getPoint(int index) const;
+    std::vector<Point>::const_iterator iteratorBegin() const;
+    std::vector<Point>::const_iterator iteratorEnd() const;
```
Febbe: try to return `std::vector<Point> const&` instead of two iterators

src/model/Stroke.h
```diff
@@ -72,8 +70,9 @@ class Stroke : public AudioElement
...
+    Point getPoint(int index) const;
```
Febbe: return `Point const&`

src/view/StrokeView.cpp
```diff
@@ -171,13 +174,15 @@ void StrokeView::drawWithPressuire()
 {
 	int count = 1;
 	double width = s->getWidth();
-	ArrayIterator<Point> points = s->pointIterator();
+    std::vector<Point>::const_iterator points = s->iteratorBegin();
```
Febbe: generally use the auto type for known types for return values, like:
`auto p_iter = s->iteratorBegin();`

src/view/StrokeView.cpp
```diff
@@ -171,13 +174,15 @@ void StrokeView::drawWithPressuire()
...
-	Point lastPoint1 = points.next();
+    Point lastPoint1 = *points;
+    points++;
 	double dashOffset = 0;
-	while (points.hasNext())
+    while (points != s->iteratorEnd())
 	{
-		Point p = points.next();
+        Point p = *points;
+        points++;
```
Febbe: use a for loop like 
`for(auto&& p:points){` (in case of points being an container) or use
`for(auto i = s->iteratorBegin(), i != s->iteratorEnd(),++i){` in case of iterators

src/gui/XournalView.h
```diff
@@ -95,7 +93,9 @@ class XournalView : public DocumentListener, public ZoomListener
 	void repaintSelection(bool evenWithoutSelection = false);
 
 	TextEditor* getTextEditor();
-	ArrayIterator<XojPageView*> pageViewIterator();
+    std::vector<XojPageView*>::const_iterator iteratorBegin() const;
+    std::vector<XojPageView*>::const_iterator iteratorEnd() const;
```
Febbe: same as before try to return `container const&`

LittleHuba:
@st4ll1 thanks for your work so far. We'd like to merge this PR soon. Could you please implement the requested changes and resolve the merge conflicts by rebasing onto master?

src/control/ToolHandler.h
```diff
@@ -126,7 +129,8 @@ class ToolHandler
 private:
 	XOJ_TYPE_ATTRIB;
 
-	Tool* tools[TOOL_COUNT] = { 0 };
+    std::array<Tool*, TOOL_COUNT> tools;
```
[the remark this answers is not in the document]
st4ll1: replaced with std::array<std::unique_ptr<Tool>,TOOL_COUNT> tools;

src/view/StrokeView.cpp
st4ll1: added a helper template for loops with special cases for the first element,
since this was more then once the case,

https://raw.githubusercontent.com/st4ll1/xournalpp/remove_arrayiterator/src/util/LoopUtil.h

inspired by https://stackoverflow.com/questions/54831971/c-range-based-loop-with-special-case-for-first-item

i like this solution since there is no extra if in the loop

st4ll1:
I did not have much time this weekend to look into this PR,  

@Febbe I just run the clang format to your changes.

Febbe:
@LittleHuba, I think it is mergeable. Can you do it, if you think everything is fine. I will do it at 23:59 CEST

src/control/shaperecognizer/CircleRecognizer.cpp
```diff
@@ -45,24 +46,17 @@ double CircleRecognizer::scoreCircle(Stroke* s, Inertia& inertia)
...
+	auto const& pv = s->getPointVector();
+	if (pv.empty())
 	{
 		return 0;
 	}
 
-	Point p1 = it.next();
-
-	while (it.hasNext())
+	for (auto p1_i = begin(pv), p2_i = std::next(p1_i), p_end_i = end(pv); p2_i != p_end_i; ++p2_i, ++p1_i)
```
LittleHuba: Those names are not really descriptive...

src/control/tools/StrokeHandler.cpp
```diff
@@ -184,14 +184,11 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos)
 	// Backward compatibility and also easier to handle for me;-)
 	// I cannot draw a line with one point, to draw a visible line I need two points,
 	// twice the same Point is also OK
-	if (stroke->getPointCount() == 1)
+	if (auto const& pv = stroke->getPointVector(); pv.size() == 1)
```
LittleHuba: Descriptive naming...

Febbe: Actually in this case, the most would be over descriptive. "pv" is a scope variable with less than 10 lines, and the assignment is more than descriptive. I follow the guideline, if it can be inlined as good as using a temporary and if the scope is short(up to 15+-5 lines), use a short variable name. The use of this temporary is there to reduce duplicate calls to one long function.

PR #1547, opened 2019-10-15
Add a clang-tidy conf file

Febbe:
@JJones780 can you look over this?

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,34 @@
+Checks: "*,\
+-misc-unused-parameters,\
+-cppcoreguidelines-pro-type-union-access,\
+-clang-analyzer-alpha*,\
+-google-build-using-namespace,\
+-clang-analyzer-optin.osx*,\
+-clang-analyzer-osx*,\
+-llvm-include-order,\
+-llvm-header-guard,\
+-readability-named-parameter,\
+-cppcoreguidelines-pro-type-reinterpret-cast,\
+-cppcoreguidelines-pro-type-vararg,\
+-misc-move-const-arg,\
+-google-runtime-references,\
+-cert-err58-cpp,\
+-fuchsia-overloaded-operator,\
+-fuchsia-default-arguments,\
+-hicpp-vararg,\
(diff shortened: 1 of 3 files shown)


JJones780:
I haven't been using Clang-tidy regularly although I probably should as it's even built into [at least one of the IDE's I use](https://clang.llvm.org/extra/clang-tidy/Integrations.html).

I'm good with including a clang-tify config file in the git.  AFAIK it doesn't get used in the build but is very handy for developers to use to check as they go.

If someone wants to use their own .clang-tidy config they can set it up in their IDE or just put it in the src directory so it's seen before this one.

As far as the config file option themselves I only took a quick look over them.  
I tried running it on a few files and so far my only comments would be:
-  cppcoreguidelines-pro-type-cstyle-cast: This one gets tripped up by standard GTK practices.
-  -google-build-using-namespace:  this check could be included?

JJones780:
Also,  If we want to have the clang-tidy do a modernize run through the code ( [check the results of my script from an earlier test run that tried EVERY modernize option](https://github.com/JJones780/xournalpp/commits/testModernize) ) we should probably do that alongside a clang-format changeover which should happen someday soon too.

LittleHuba:
> Also, If we want to have the clang-tidy do a modernize run through the code ( [check the results of my script from an earlier test run that tried EVERY modernize option](https://github.com/JJones780/xournalpp/commits/testModernize) ) we should probably do that alongside a clang-format changeover which should happen someday soon too.

Yeah this is the actual PR-Freeze as we have to merge everything before this PR and open no new ones before it is done. Otherwise we might get inconsistent styles again.

Technius:
Can you add `-modernize-use-override`? A lot of methods are missing `override` specifiers.

Febbe:
Uh. I think this option is really useful. And the automatic clang-tidy cleanup, will replace all virtuals with override. Also missing flags are added automatically. So this won't be a problem animore

Febbe:
@LittleHuba that's exactly what I want. We can check for those rules as we check for the format.

PR #1549, opened 2019-10-16
Add flatpak build instructions and move manifest

Technius:
I moved the Flatpak manifest to https://github.com/flathub/com.github.xournalpp.xournalpp/ and updated it earlier today. This PR adds instructions on how to install the Flatpak and how to build it.


LittleHuba:
This means we have to update the manifest for each release? Should we automate this with the pipeline and create a PR automatically at the flathub repo when we push a new version?

Technius:
> This means we have to update the manifest for each release?

Yes, that is currently the case. That only thing that's required is to watch out for dependencies and commit/tags. If the build method doesn't change, then an automated PR changing the `tag` and `commit` fields of the manifest should suffice. However, if we add or remove dependencies, we may need to update the Flatpak manifest to package any added dependencies.

Technius:
That only requires updating the CI configuration, so I will merge this tomorrow if there are no objections.

LittleHuba:
Maybe open an issue though so we don't forget this?

PR #1555, opened 2019-10-18
Fix build configuration

### LINKED ISSUE #1553: Can't cmake on Ubuntu 18.04
Can't build latest master. I'm running Linux 18.04. 

Versions:
- X++ latest master
- **cmake --version** `cmake version 3.10.2`
- **gcc -v**:
`Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 7.4.0-1ubuntu1~18.04.1' --with-bugurl=file:///usr/share/doc/gcc-7/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-7 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1) `

I'm attaching CMake files:
[CMakeError.log](https://github.com/xournalpp/xournalpp/files/3744500/CMakeError.log)
[CMakeOutput.log](https://github.com/xournalpp/xournalpp/files/3744501/CMakeOutput.log)

**Output when running cmake:**
`CMake Error at cmake/find/FindCXX17.cmake:72 (target_compile_features):
  Cannot specify compile features for imported target "cxx17".
  CMakeLists.txt:64 (find_package)

-- Checking for module 'gtk-mac-integration'
--   No package 'gtk-mac-integration' found
Enable Xournal++ Plugins
Installing desktop files

Configuration:
	Compiler:                   /usr/bin/c++
	CppUnit enabled:            OFF

-- Configuring incomplete, errors occurred!
See also "/home/mark/Documents/Repos/External/xournalpp/build/CMakeFiles/CMakeOutput.log".
See also "/home/mark/Documents/Repos/External/xournalpp/build/CMakeFiles/CMakeError.log".`

Technius:
Fixes #1553 (tested on a Ubuntu 18.04 image on Docker).

Note that CI did not catch this issue because the "new toolchain" PPA was added before installing build dependencies.

`readme/LinuxBuild.md` will need to be updated to indicate the new dependency on a compiler that supports C++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,9 +61,6 @@ macro (add_includes_ldflags LDFLAGS INCLUDES)
     set (xournalpp_INCLUDE_DIRS ${xournalpp_INCLUDE_DIRS} "${INCLUDES}")
 endmacro (add_includes_ldflags LDFLAGS INCLUDES)
 
-find_package(CXX17 REQUIRED COMPONENTS optional)
-set(xournalpp_LDFLAGS ${xournalpp_LDFLAGS} cxx17)
-
 # libexec
 if (WIN32)
     set (xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-mwindows")
--- a/cmake/find/FindCXX17.cmake
+++ b/cmake/find/FindCXX17.cmake
@@ -23,13 +23,9 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
 
 # Normalize and check the component list we were given
 set(want_components ${CXX17_FIND_COMPONENTS})
-if (CXX17_FIND_COMPONENTS STREQUAL "")
-#Nothing to do here
-endif ()
 
 # Warn on any unrecognized components
-if (NOT extra_components STREQUAL "")
-    set(extra_components ${want_components})
+if (extra_components)
     list(REMOVE_ITEM extra_components optional filesystem map)
     foreach (component IN LISTS extra_components)
         message(WARNING "Extraneous find_package component for Filesystem: ${component}")
@@ -63,13 +59,6 @@ if ("map" IN_LIST want_components)
     endif ()
 endif ()
 
-if (TARGET cxx17)
-    # This module has already been processed. Don't do it again.
-    return()
-endif ()
-
-add_library(cxx17 INTERFACE IMPORTED)
-target_compile_features(cxx17 INTERFACE cxx_std_17)
 set(_found TRUE)
 
 cmake_pop_check_state()
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -3,6 +3,4 @@ unset (util_SOURCES)
 file (GLOB_RECURSE util_SOURCES *.cpp)
 
 add_library (util STATIC ${util_SOURCES})
-find_package(CXX17 REQUIRED)
-target_link_libraries(util PRIVATE cxx17)
 target_compile_features(util PUBLIC ${PROJECT_CXX_FEATURES})


Febbe:
I don't approve these changes. Can you make a switch between those cmake versions. And I would rather add the toolchain to the requirements.

markwmuller:
BTW: I tested, make works fine when cmake uses the default compiler (gcc-7). No need to explicitly spec a newer compiler than you get from the "simple" apt install.

Febbe:
Yes, we planned to use all cpp17 STL features before, but stepped back to only use optional. But generally there is no reason to use old toolchains, if it is simple to use newer ones.

Febbe:
I can recoment to install the newest ubuntu toolchain with the following commands:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update && sudo apt-get upgrade
sudo apt install gcc g++ gcc-8 g++-8
```
You can even use g++9 or g++10 if you want.
Then run
```
sudo apt-get install -y cmake ninja-build libcppunit-dev libgtk-3-dev libpoppler-glib-dev portaudio19-dev libsndfile-dev liblua5.3-dev libzip-dev
export CXX=g++-8
export CC=gcc-8

cmake ..
make/ninja
```

I prefer to set the CC and CXX Flags directly in my .bashrc or .profile preference file.

markwmuller:
Thank @Febbe , but note that this is not feasible in all cases (I am not willing to use a non-standard/non-default) compiler as default on my system, as it will likely break other things.

Febbe:
It shouldn't break anything since all modern C and C++ compilers are backwards compatible. It's probably better, if the compiler maintainer introduced better optimizations and bigfixes.

Rmano:
Hmmm... I know I am not a developer, but I try to test and help to debug. Please, try not to force everybody to use a PPA for such an important thing as a compiler chain. In a lot of computers here I would not be allowed to use them (nor to upgrade to a non-LTE distro) --- add at least a switch to make the thing compilable in current mainstream distro like Ubuntu 18.04. 
Just my 2 cents; thank a lot to all the developers here.

Febbe:
It was only a recommendation because it is very easy and fast 😅. But if you have problems with your institution you can still install the compiler and cmake into your home directory.
And for now it works to compile xournalpp with gcc-7, but we will use the filesystem library in the future. From then on gcc-8 or clang-7 are required to link against the STL implementation. However I added the possibility to link against the boost:: filesystem library if the STL version is not feasible.

LittleHuba:
Since we nowadays provide packages for distributions even in a nightly format this will not be of major impact to the big userbase. But I'd still say we should not make devs jump through too many hoops.

We should therefore provide the header and make CMake choose it as fallback if the compiler does not provide the functionality, without requiring further input of the dev. This should be transparent.

But I don't really know if the filesystem of C++ is worth all this effort... We do't really have a lot of contact with the filesystem outside of opening the few files and glib provides the necessary functionality as well. This is all just convenience paid for with restrictions on the code...

markwmuller:
Quick note: you can apt install g++-8 gcc-8 on 18.04, without adding any PPAs. This may allow the devs to allow use of 18.04 while keeping all the modern shiny features of latest compilers. One does then need to spec some additional cmake commands, to force it to use g++-8 and gcc-8.

PR #1558, opened 2019-10-21
Cleanup of repository

LittleHuba:
This PR cleans up the repository:
- Pipelines
  - Debian dockerfile
  - Remove travis
  - Naming of pipelines
- Remove Eclipse project files and add them to gitignore
- Remove check-source script
- Documentation
  - Prioritize native build over homebrew for MacOS
  - Improve windows documentation

Pending changes are moving the PPA files to one single directory.

Overall goal is to clean up the repository to give a development centric view and not a cluttered main directory where nobody finds what he searches.


LittleHuba:
**Right after merging this PR requires updating of the pipelines!**

LittleHuba:
I'll do the moving of the PPA files in a different PR, as we should first get it running again before we tinker with the files. Otherwise we might break even more.

Febbe:
we should also consider to cleanup the CMake files. All executables should be build in the base build directory or in bin, not in ./src/ for example

LittleHuba:
> we should also consider to cleanup the CMake files. All executables should be build in the base build directory or in bin, not in ./src/ for example

This should be done in a separate PR. Since I'm shifting pipelines we can not determine build status for this PR.

PR #1576, opened 2019-10-28
update traditional Chinese translation

doraeric:


LittleHuba:
Hi thanks for the translations! 

In the future you might want to use Crowdin for a better experience. It provides you with a graphical user interface and helps you translate the strings. Link is in the Readme!

doraeric:
Oh, I see. Thanks!

PR #1584, opened 2019-11-02
Update fr_FR.po

Maniform:
In french, when you close a file without saving it, xournal++ asks if you really want to close and gives you the choice to "Sauvegarder", "Annuler" or "Annuler". I changed the "Discard" translation from "Annuler" to "Abandonner".


LittleHuba:
Thanks for the PR!
For the future you can use our Crowdin project to edit translations online with a nice user-interface. Link is in the readme.

PR #1589, opened 2019-11-05
Simple clang-tidy fixes

Febbe:
only modernize-* changes, that could be done without errors.
This should be mergeable directly.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -8,6 +8,8 @@ Checks: "*,\
 -llvm-include-order,\
 -llvm-header-guard,\
 -readability-named-parameter,\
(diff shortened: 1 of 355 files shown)


Febbe:
I would like to merge those changes the next few days.

LittleHuba:
One quick question: Shouldn't the Clang Format check sign green with this? :wink:

Febbe:
> One quick question: Shouldn't the Clang Format check sign green with this? wink
Yes, but we want to swap the clang-format to the 4 indent google stile anyways.
Did the formating now for the last 40 commits.

Issue #967, opened 2019-03-09
Arrow:  tip change size as length increases.


deragon:
When creating an arrow, the tip of the arrow changes size as the length increases.  That means that multiple arrows of different lengths on the same page do not look the same.

At a minimum, the tip should remain the same regarding the length of the arrow.

Also, a tip selection tool would be welcomed.  I would like to be able to choose tips on both ends.  But I should create a separate issue about this.  On the main webpage of this fantastic tool (I discovered Xournal++ 30 minutes ago), there is no mention on how to get support.  That is why I am writing here more than I should here.  I would one go to discuss issue with the author(s) before submitting an official issue?

andreasb242:
> That is why I am writing here more than I should here. I would one go to discuss issue with the author(s) before submitting an official issue?

This issues are a good way to contact us, and we also do discussions here, at least on some issues.
So feel free to create an issue, also for discussion.

Additional there is a link to gitter on the Readme on top. But the most technical discussions happen here on the issues.

Back to the technical Question:
It should be relative easy to implement different tips, and let the user select them. But only BEFORE drawing, because an arrow will be drawed as 3 lines on the document (because of this you can also erase some parts), so changing the tips after drawing will not be possible.
Would this be Ok, or would you like to change the arrows after drawing? (You can request it - even if it will not be possible at the Moment ;-))

Rmano:
My opinion is that the tip size should be proportional to the line thickness, not to the length...

And yes, the most useful way to implements tips is to make them a property of the line, so that you will be able to change them after-the-fact. 

But just implementing the first point would make them usable (now they have little usability, I think.)

LittleHuba:
We could allow late modification of the tips by doing the following: 

Create another object class for arrows and store the actual arrow line separately from the lines of the tips. Allow modifying of the tips in that class, Erasing of sections of the arrow can also be handled in this class. 

Storing is more of a problem as we would need to increment the file version number and introduce a new element to the format. Therefore I'd say we postpone this until we have the new format. There we can also enforce requirements on the client reading a file and throw an error when finding an unknown element.

deragon:
Unless you are willing to implement shortly all the bells and whistles required for proper arrow support, I believe that this issue should just tackle a quick fix to ensure that the tip remains of the same size regardless of length.  Other issues could be created to tackle the more sophisticated and demanding changes, such as tip selection and tip change once the arrow is drawn.

[...]

PR #1606, opened 2019-11-16
Drawing tools

redweasel:
Closes: #967
Also the coordinate system drawing direction is now more natural.


src/control/tools/ArrowHandler.cpp
```diff
@@ -57,7 +57,7 @@ void ArrowHandler::drawShape(Point& c, const PositionInputData& pos)
 
 		// set up the size of the arrowhead to be 1/8 the length of arrow
```
LittleHuba: That comment is now missleading and should be changed to the new behaviour.

src/control/tools/CoordinateSystemHandler.cpp
```diff
@@ -66,7 +66,7 @@ void CoordinateSystemHandler::drawShape(Point& c, const PositionInputData& pos)
 		if ( !this->modControl )	//draw out from starting point
 		{			
 			stroke->addPoint(p1);
```
LittleHuba: This line has bad formatting. There seem to be many whitespaces at the end of the line that should not be there.

LittleHuba:
Sorry but you have now fixed the formatting for the whole files which is not something we want. Please revert the last commit and only change the two lines I have highlighted. Otherwise it will be a huge pain to later find the commit that caused a bug when commits change lines only for formatting.

src/control/tools/CoordinateSystemHandler.cpp
```diff
@@ -65,24 +65,15 @@ void CoordinateSystemHandler::drawShape(Point& c, const PositionInputData& pos)
 		
 		if ( !this->modControl )	//draw out from starting point
 		{
```
LittleHuba: No luck now it is that line :wink:

LittleHuba:
I just checked the other PRs and none conflict with your changes. So we can ignore formatting and do that in the Formatting PR that is still pending.
Once PR-Freeze is formally over this will be a lot easier!

Thanks for the work!

redweasel:
@LittleHuba thanks for the help. I'm very new to GitHub and messed up a couple times here and learned a lot.

LittleHuba:
No worries! Formatting is currently not really easy to get right as the whole codebase is not yet formatted correctly. That is what the PR-Freeze is about. But afterwards it will be one command to format the PR correctly.

Your actual work was very nicely done and I hope to see more work from you! We have a lot of issues which are suitable for beginners.

PR #1613, opened 2019-11-18
Reintroduce algorithm to select current page when scroll changes

### LINKED ISSUE #1184: Active page does not update when scrolling

LittleHuba:
Fixes #1184

### DIFF
--- a/src/gui/Layout.cpp
+++ b/src/gui/Layout.cpp
@@ -68,6 +68,10 @@ void Layout::updateVisibility()
 	int x1 = 0;
 	int y1 = 0;
 
+	// Data to select page based on visibility
+	size_t mostPageNr = 0;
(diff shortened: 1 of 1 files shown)


LittleHuba:
Will merge in 24h if no complaints.

Issue #1604, opened 2019-11-15
Can't open the built project in my language


Maniform:
Hi everyone,

I'm french and I built the project to try to improve some features. I edited the french translation but when I launched the built project, the language is still in english no matter what I'm doing.

My OS : Ubuntu 18.04.3 LTS

I tried to :
- launch the binaries -> ./xournalpp
- launch with a specific language : LANGUAGE=fr_FR ./xournalpp
- install built deb and launch it
- launch from the built deb with a specific LANGUAGE

I executed "make pot" and "make translations" before and after building the binairies.

Thank you for your help !

LittleHuba:
For translations to be updated and applied you have to provide the install prefix. You can set the install prefix to a directory within your installation directory and then run `make install` and run the program from the directory you specified. That should solve the issue.

LittleHuba:
Check out the Wiki for details on how to provide the prefix

Maniform:
Installing locally works fine, but I would prefer to install properly with the deb package, I don't understand why it doesn't work that way, but if it's too complex it's ok !

LittleHuba:
That is just how gettext works. You can still install the deb-package.

Maniform:
I tried to install from the deb package but it's still in english.

Maniform:
I read all the "Installing" wiki page. The only way I found to have the right localization is to install from source to the $HOME/.local folder. When I build a DEB package, the language is locked on English even if std::locale().name().c_str() return "fr_FR.UTF-8".

And we can't configure the project with a PREFIX and with a package target, cmake returns this error : "Current working directory cannot be established."

LittleHuba:
Did you follow the instructions in the Wiki on how to build the Debian package?

Maniform:
I followed the explanations from the LinuxBuild.md file (there is a ".deb packages" category), not from the wiki, I'll try from the wiki.

LittleHuba:
> I followed the explanations from the LinuxBuild.md file

Sorry that is what I meant.

LittleHuba:
Do you have another version of Xournal++ installed? The packages might collide.

Maniform:
I uninstalled every instance of xournalpp (or xournal++) before this, so yes. I still have xournal installed but it shouldn't be a problem.

For the LinuxBuild.md file, I exactly did these commands :
git clone https://github.com/xournalpp/xournalpp.git
cd xournalpp
mkdir build
cd build
[...]

PR #1617, opened 2019-11-19
Make sure translations are built before installation

TheAssassin:
Fixes the following issue in a clean checkout when trying to `make install`:

```
CMake Error at po/cmake_install.cmake:41 (file):
  file INSTALL cannot find
  ".../xournalpp/craftbuild/cmake-build/po/cs.gmo".
  cmake_install.cmake:102 (include)
```

I've discovered this bug while building a new AppImage for the last release only. I'm not sure if the dependency on the main binary makes too much sense, but on the other hand that's where the translations are needed. Please feel free to change the dependency to some other target if you think that's more suitable.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -94,6 +94,7 @@ add_executable (xournalpp
 add_dependencies (xournalpp
   xournalpp-core
   util
+  translations
 )
 
 target_link_libraries (xournalpp ${xournalpp_LDFLAGS})


LittleHuba:
I think the main target actually makes the most sense. Since those files are not needed for anything else than the main executable which is built by the main target.

LittleHuba:
But it seems like that fix is not actually working. I have the feeling this is the problem @Technius mentioned somewhere else. We will have to edit the Gettext file to fix the build command.

TheAssassin:
Well, any fix would be appreciated. This change works fine when building an AppImage with appimagecraft. Can you point me to the build log of the failed builds? I am not used to this Azure pipeline ... stuff...

LittleHuba:
You can get there by GitHub. Just go to the checks page and click on the link for more details.
Then once you are in azure you can click on one of the build steps which have the red x in front. Then it should show you the log.
Sorry I'm on my smartphone so I don't really have the best way to get you the direct link right now.

TheAssassin:
Seems some file is just not marked executable or some tool is simply missing. The log isn't too helpful. I hope @Technius can tell what the issue is.

LittleHuba:
If I'm not mistaken it was that we still have to supply a target for the gettext script. There was an issue he commented here in the repo.

Technius:
The comment in question is https://github.com/xournalpp/xournalpp/issues/1604#issuecomment-555224104. The issue concerned building fresh clones, which I fixed just now in 62fb7df.

Currently, I don't think the translation target is run at all, but we don't want the translation target to run every compile since the `po` files are annoying to deal with in git. The long-term solution is to create a "lazy translation" target that only builds the translation target when the timestamp of at least one translation file changes.

A quick workaround is to re-add the line removed in this commit: 84290e5

LittleHuba:
Maybe this helps:
https://stackoverflow.com/a/20902962

TheAssassin:
Quick fix available in #1729, but I'd rather like to see this merged instead.

Febbe:
Is it possible to build the translations into the build directory, instead of the source directory?
This would completely resolve the problem with git.

TheAssassin:
@Febbe not the way things are implemented. The `make translations` target actually updates the `.gmo` "source" files in the repo root. They're then processed as far as I can see into something more machine readable.

Technius:
Closing since this issue was fixed in #1737.

PR #1626, opened 2019-11-23
Fix #1583 Can't bold when Caps Lock is enabled

### LINKED ISSUE #1583: Can't bold text if CAPS LOCK is enabled
**Affects versions :**
 - OS: Arch Linux
 - DE: Gnome
 - Version of Xournal++: commit a3e8901 (latest)

**Describe the bug**
If I press CAPS LOCK and then try to bold it with CTRL+B before having started writing, it simply doesn't work.
I have to press CTRL+B and then CAPS LOCK in order for it to work.
Plus if you write something in CAPS and then try to bold it, it doesn't work too.

**To Reproduce**
1. Activate the CAPS LOCK
2. Activate the bold with CTRL+B
3. Start writing

Or

1. Activate the CAPS LOCK
2. Start writing
3. Press CTRL+B

**Expected behavior**
I should be able to activate bold in every condition, whether CAPS LOCK is active or not.

matepak:


Technius:
Thanks for the PR!

Febbe:
Unfortunately this is not the way to handle caps lock and other modifiers, please provide another PR with the changes suggested by gtk: https://developer.gnome.org/gtk3/stable/checklist-modifiers.html

matepak:
Ok, i will provide new PR in accordance with gtk.

PR #1627, opened 2019-11-25
Build windows release in Docker container

### LINKED ISSUE #1600: Windows 10 installation issues
Just tried the Windows installer on Windows 10 (1903) and it appears to be not working correctly.
First the installer does not add the program to the start menu.
After installation when trying to run the program manually from `C:\Program Files\Xournal++\bin\xournalpp.exe` you get two System Errors stating that `The code execution cannot proceed because libcrypto-1_1-x64.dll was not found.  Reinstalling the program may fix this problem.` and another one for libssl-1_1-x64.dll.

LittleHuba:
Fixes #1600

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -89,7 +89,7 @@ stages:
           
       - job: Windows
         pool:
-          vmImage: 'win2016-vs2017'
+          vmImage: 'vs2017-win2016'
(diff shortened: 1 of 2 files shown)


LittleHuba:
This (unfinished idea) would be a valid way but introduces a lot of overhead and maintenance.

PR #1630, opened 2019-11-25
Fixed test for modifier keys

matepak:
More proper way to test for modifier keys, according to [https://developer.gnome.org/gtk3/stable/checklist-modifiers.html](url)


Febbe:
Thank you!

PR #1641, opened 2019-11-29
Fix Debian build

pbirkants:
The fix for building translations added in #1602 breaks the build in a clean source tree.

```
$ git clone https://github.com/xournalpp/xournalpp.git
$ cd xournalapp/
# Generate .orig.tar.gz so debuild does not complain
$ dh_make -s --createorig -p xournalapp_1.0.6
$ debuild
[..]
   debian/rules override_dh_auto_build
make[1]: Entering directory '/tmp/git3/xournalpp'
make translations all
make[2]: Entering directory '/tmp/git3/xournalpp'
make[2]: *** No rule to make target 'translations'.  Stop.
make[2]: Leaving directory '/tmp/git3/xournalpp' 
```

This fixes the command as described in the [Debian New Maintainers Guide](https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#customrules):
>  If the Makefile in the source for gentoo requires you to specify build as its target to build it [52], you create an override_dh_auto_build target to enable this.
> ```
> override_dh_auto_build:
>        dh_auto_build -- build
> ```

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -3,5 +3,5 @@
 	dh $@
 
 override_dh_auto_build:
-	make translations all
+	dh_auto_build -- translations all
 


andreasb242:
Hi @pbirkants 

Thank you very much!
I'v searched for the right command a really long time ago, the Ubuntu PPA was not working anymore.
Now the Ubuntu PPA is built again since a long long time :-)

grafik

pbirkants:
Great :)

You should also update `debian/changelog` to at least match the current release version number.

LittleHuba:
> You should also update debian/changelog to at least match the current release version number

This should be checked in the release pipeline. Running a regex over the file checking if the version string is contained should be sufficient. If the string is not contained, the pipeline should throw an error should be thrown and not create a new release.

PR #1642, opened 2019-11-29
Optionalize audio

### LINKED ISSUE #1468: Optionalise Portaudio,sndfile&ALSA (and add Pulseaudio support?)
It'd be nice if portaudio,alsa&sndfile support was optional, since most systems only use of of those (or none of them and just use PA). I can give this a shot if you want to (and maybe have some pointers :)

LittleHuba:
Fixes #1468 

I tried to minimize the influence of this change in the codebase. Still more changes than I would like to have...

### DIFF
--- a/src/control/AudioControllerImpl.cpp
+++ b/src/control/AudioControllerImpl.cpp
@@ -1,0 +1,160 @@
+#include "AudioControllerImpl.h"
+#include "Util.h"
+
+#include <i18n.h>
+#include <XojMsgBox.h>
+
+AudioControllerImpl::AudioControllerImpl(Settings* settings, Control* control) : AudioController(settings, control)
+{
+	this->settings = settings;
+	this->control = control;
+	this->audioRecorder = new AudioRecorder(settings);
+	this->audioPlayer = new AudioPlayer(control, settings);
+}
+
+AudioControllerImpl::~AudioControllerImpl()
+{
+	delete this->audioRecorder;
+	this->audioRecorder = nullptr;
+
+	delete this->audioPlayer;
+	this->audioPlayer = nullptr;
+}
+
+auto AudioControllerImpl::startRecording() -> bool
+{
+	if (!this->isRecording())
+	{
+		if (getAudioFolder().isEmpty())
+		{
+			return false;
+		}
+
+		this->timestamp = static_cast<size_t>(g_get_monotonic_time() / 1000);
+
+		std::array<char, 50> buffer{};
+		time_t secs = time(nullptr);
+		tm* t = localtime(&secs);
+		// This prints the date and time in ISO format.
+		sprintf(buffer.data(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1, t->tm_mday,
+		        t->tm_hour, t->tm_min, t->tm_sec);
+		string data(buffer.data());
+		data += ".ogg";
+
+		audioFilename = data;
+
+		g_message("Start recording");
+
+		bool isRecording = this->audioRecorder->start(getAudioFolder().str() + "/" + data);
+
+		if (!isRecording)
+		{
+			audioFilename = "";
+			this->timestamp = 0;
+		}
+
+		return isRecording;
+	}
+	return false;
+}
+
+auto AudioControllerImpl::stopRecording() -> bool
+{
+	if (this->audioRecorder->isRecording())
+	{
+		audioFilename = "";
+		this->timestamp = 0;
+
+		g_message("Stop recording");
+
+		this->audioRecorder->stop();
+	}
+	return true;
+}
+
+auto AudioControllerImpl::isRecording() -> bool
+{
+	return this->audioRecorder->isRecording();
+}
+
+auto AudioControllerImpl::isPlaying() -> bool
+{
+	return this->audioPlayer->isPlaying();
+}
+
+auto AudioControllerImpl::startPlayback(string filename, unsigned int timestamp) -> bool
+{
+	this->audioPlayer->stop();
+	bool status = this->audioPlayer->start(filename, timestamp);
+	if (status)
+	{
+		this->control->getWindow()->getToolMenuHandler()->enableAudioPlaybackButtons();
+	}
+	return status;
+}
+
+void AudioControllerImpl::pausePlayback()
+{
+	this->control->getWindow()->getToolMenuHandler()->setAudioPlaybackPaused(true);
+
+	this->audioPlayer->pause();
+}
+
+void AudioControllerImpl::seekForwards()
+{
+	this->audioPlayer->seek(this->settings->getDefaultSeekTime());
+}
+
+void AudioControllerImpl::seekBackwards()
+{
+	this->audioPlayer->seek(-1 * this->settings->getDefaultSeekTime());
+}
+
+void AudioControllerImpl::continuePlayback()
+{
+	this->control->getWindow()->getToolMenuHandler()->setAudioPlaybackPaused(false);
+
+	this->audioPlayer->play();
+}
+
+void AudioControllerImpl::stopPlayback()
+{
+	this->control->getWindow()->getToolMenuHandler()->disableAudioPlaybackButtons();
+	this->audioPlayer->stop();
+}
+
+auto AudioControllerImpl::getAudioFilename() -> string
+{
+	return this->audioFilename;
+}
+
+auto AudioControllerImpl::getAudioFolder() -> Path
+{
+	string af = this->settings->getAudioFolder();
+
+	if (af.length() < 8)
+	{
+		string msg = _("Audio folder not set! Recording won't work!\nPlease set the "
+					   "recording folder under \"Preferences > Audio recording\"");
+		g_warning("%s", msg.c_str());
+		XojMsgBox::showErrorToUser(this->control->getGtkWindow(), msg);
+		return Path("");
+	}
+
+	return Path::fromUri(af);
+}
+
+auto AudioControllerImpl::getStartTime() const -> size_t
+{
+	return this->timestamp;
+}
+
+auto AudioControllerImpl::getOutputDevices() -> vector<DeviceInfo>
+{
+	return this->audioPlayer->getOutputDevices();
+}
+
+auto AudioControllerImpl::getInputDevices() -> vector<DeviceInfo>
+{
+	return this->audioRecorder->getInputDevices();
+}
--- a/src/control/AudioControllerImpl.h
+++ b/src/control/AudioControllerImpl.h
@@ -1,0 +1,57 @@
+/*
+ * Xournal++
+ *
+ * Audio Recording / Playing controller
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+#include <XournalType.h>
+#include "settings/Settings.h"
+#include "Control.h"
+#include "AudioController.h"
+#include <Path.h>
+#include <util/audio/AudioRecorder.h>
+#include <util/audio/AudioPlayer.h>
+#include <gui/toolbarMenubar/ToolMenuHandler.h>
+
+class AudioControllerImpl: public AudioController
+{
+public:
+	AudioControllerImpl(Settings* settings, Control* control);
+	~AudioControllerImpl() override;
+
+public:
+	bool startRecording() override;
+	bool stopRecording() override;
+	bool isRecording() override;
+
+	bool isPlaying() override;
+	bool startPlayback(string filename, unsigned int timestamp) override;
+	void pausePlayback() override;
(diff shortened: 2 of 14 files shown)


src/control/AudioControllerImpl.cpp
```diff
@@ -0,0 +1,160 @@
...
+	this->audioPlayer = new AudioPlayer(control, settings);
+}
+
+AudioControllerImpl::~AudioControllerImpl()
+{
+	delete this->audioRecorder;
+	this->audioRecorder = nullptr;
+
+	delete this->audioPlayer;
+	this->audioPlayer = nullptr;
```
Febbe: remove delete's, after replacing with  unique ptrs.
remove `this->[parameter] = nullptr;` after delete in destructors since its dead code anyway and no function can call delete on `this->[parameter]` after the constructor.
If it happens, there is a programming mistake, which should not be silently catched by a set to nullptr. The software musst crash then. 

Remove empty destructor completely. (Don't even keep defaulted destructor)

src/control/AudioControllerImpl.cpp
```diff
@@ -0,0 +1,160 @@
...
+			return false;
+		}
+
+		this->timestamp = static_cast<size_t>(g_get_monotonic_time() / 1000);
+
+		std::array<char, 50> buffer{};
+		time_t secs = time(nullptr);
+		tm* t = localtime(&secs);
+		// This prints the date and time in ISO format.
+		sprintf(buffer.data(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1, t->tm_mday,
```
Febbe: use the <chrono> library instead of <ctime>

src/control/AudioControllerImpl.h
```diff
@@ -0,0 +1,57 @@
+/*
```
Febbe: move file into AudioControler.cpp and use the PImpl patern.

src/control/Control.cpp
```diff
@@ -1,5 +1,9 @@
 #include "Control.h"
 
+#ifdef ENABLE_AUDIO
+#include "AudioControllerImpl.h"
+#endif
+
```
Febbe: Avoid to many preprozessor trees in different files by moving all into the AudioControler.cpp

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -365,6 +365,10 @@ void SettingsDialog::load()
 	touch.getInt("timeout", timeoutMs);
 	gtk_spin_button_set_value(GTK_SPIN_BUTTON(get("spTouchDisableTimeout")), timeoutMs / 1000.0);
 
+#ifndef ENABLE_AUDIO
+	GtkNotebook* notebook = GTK_NOTEBOOK(get("notebook1"));
+	gtk_notebook_remove_page(notebook, gtk_notebook_page_num(notebook, get("audioRecordingTabBox")));
+#else
```
Febbe: handle this over cpp code (`if constexpr ([stmt])` ), not over the preprozessor

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -626,6 +632,7 @@ void SettingsDialog::save()
 		settings->setAudioOutputDevice(
 		        static_cast<int>(this->audioOutputDevices[selectedOutputDeviceIndex].getIndex()));
 	}
+#endif
```
Febbe: same as above

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -614,6 +619,7 @@ void SettingsDialog::save()
 	settings->setSnapGridTolerance(
 	        static_cast<double>(gtk_spin_button_get_value(GTK_SPIN_BUTTON(get("spSnapGridTolerance")))));
 
+#ifdef ENABLE_AUDIO
```
Febbe: same as above

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -411,6 +415,7 @@ void SettingsDialog::load()
 
 	gtk_spin_button_set_value(GTK_SPIN_BUTTON(get("spAudioGain")), settings->getAudioGain());
 	gtk_spin_button_set_value(GTK_SPIN_BUTTON(get("spDefaultSeekTime")), settings->getDefaultSeekTime());
+#endif
```
Febbe: same as above

src/gui/dialog/SettingsDialog.h
```diff
@@ -11,13 +11,18 @@
...
+#endif
 #include <control/Control.h>
 #include "control/settings/Settings.h"
 #include "gui/GladeGui.h"
 #include "DeviceClassConfigGui.h"
 
 class ButtonConfigGui;
+#ifndef ENABLE_AUDIO
+class DeviceInfo;
+#endif
```
Febbe: move include to source file and leave `class DeviceInfo`, remove preprozessor stmts

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -194,6 +194,13 @@ void ToolMenuHandler::load(ToolbarData* d, GtkWidget* toolbar, const char* toolb
 				}
 				if (!found)
 				{
+#ifndef ENABLE_AUDIO
+					// Ignore unavailable audio buttons if audio is disabled.
+					if (StringUtils::startsWith(name, "AUDIO_") || name == "PLAY_OBJECT")
+					{
+						break;
+					}
+#endif
```
Febbe: handle over cpp code not preprozessor 4example over `if constexpr (<stmt>)`

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -451,12 +458,13 @@ void ToolMenuHandler::initToolItems()
...
 	ADD_CUSTOM_ITEM_TGL("SELECT_OBJECT", ACTION_TOOL_SELECT_OBJECT, GROUP_TOOL, true, "object-select", _("Select Object"));
 	ADD_CUSTOM_ITEM_TGL("VERTICAL_SPACE", ACTION_TOOL_VERTICAL_SPACE, GROUP_TOOL, true, "stretch", _("Vertical Space"));
-	ADD_CUSTOM_ITEM_TGL("PLAY_OBJECT", ACTION_TOOL_PLAY_OBJECT, GROUP_TOOL, true, "object-play", _("Play Object"));
 	ADD_CUSTOM_ITEM_TGL("HAND", ACTION_TOOL_HAND, GROUP_TOOL, true, "hand", _("Hand"));
 
 	fontButton = new FontButton(listener, gui, "SELECT_FONT", ACTION_FONT_BUTTON_CHANGED, _("Select Font"));
 	addToolItem(fontButton);
 
+#ifdef ENABLE_AUDIO
+	ADD_CUSTOM_ITEM_TGL("PLAY_OBJECT", ACTION_TOOL_PLAY_OBJECT, GROUP_TOOL, true, "object-play", _("Play Object"));
```
Febbe: preprozessor -> constexpr

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -468,6 +476,14 @@ void ToolMenuHandler::initToolItems()
...
 	                                          "audio-seek-backwards", _("Back"));
 	addToolItem(audioSeekBackwardsButton);
+#else
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuToolsPlayObject")));
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuAudioRecord")));
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuAudioPausePlayback")));
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuAudioStopPlayback")));
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuAudioSeekForwards")));
+	gtk_widget_destroy(GTK_WIDGET(gui->get("menuAudioSeekBackwards")));
+#endif
```
Febbe: preprozessor -> constexpr

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -576,6 +592,7 @@ auto ToolMenuHandler::getToolItems() -> vector<AbstractToolItem*>*
 
 void ToolMenuHandler::disableAudioPlaybackButtons()
 {
+#ifdef ENABLE_AUDIO
```
Febbe: preprozessor -> constexpr

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -587,10 +604,12 @@ void ToolMenuHandler::disableAudioPlaybackButtons()
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioStopPlayback")), false);
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioSeekForwards")), false);
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioSeekBackwards")), false);
+#endif
 }
 
 void ToolMenuHandler::enableAudioPlaybackButtons()
 {
+#ifdef ENABLE_AUDIO
```
Febbe: preprozessor -> constexpr

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -600,10 +619,13 @@ void ToolMenuHandler::enableAudioPlaybackButtons()
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioStopPlayback")), true);
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioSeekForwards")), true);
 	gtk_widget_set_sensitive(GTK_WIDGET(gui->get("menuAudioSeekBackwards")), true);
+#endif
```
Febbe: preprozessor -> constexpr

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -600,10 +619,13 @@ void ToolMenuHandler::enableAudioPlaybackButtons()
...
 }
 
 void ToolMenuHandler::setAudioPlaybackPaused(bool paused)
 {
+#ifdef ENABLE_AUDIO
 	this->audioPausePlaybackButton->setActive(paused);
 	gtk_check_menu_item_set_active(GTK_CHECK_MENU_ITEM(gui->get("menuAudioPausePlayback")), paused);
+#endif
```
Febbe: preprozessor -> constexpr

src/control/AudioControllerImpl.h
```diff
@@ -0,0 +1,57 @@
...
+	vector<DeviceInfo> getOutputDevices() override;
+	vector<DeviceInfo> getInputDevices() override;
+
+protected:
+	string audioFilename;
+	size_t timestamp = 0;
+	Settings* settings;
+	Control* control;
+	AudioRecorder* audioRecorder;
+	AudioPlayer* audioPlayer;
```
Febbe: use unique_ptrs

src/control/AudioControllerImpl.cpp
andreasb242: No, do NOT remove this nullpointer, this is really important, as C++ never checks if an instance is valid, you can cast every address to an instance, and only if the values are null, you get an error accessing it.
If they are not null, you e.g. get the Exception in the Audio Player, and it's not reproducible, why this happend!

Febbe: I don't agree with you neather in this case nor in general.
But this case first:
We have a destructor here, so accessing the pointer after the destructor is from the compiler point of view NOT POSSIBLE. The compiler will remove this line if optimizations are enabled since he (think to) knows, this value isn't accessed anymore.

In general:
Accessing a nullptr value is undefined behaviour as accessing a deleted pointer. 
Casting a nullptr_t to any type is completely legal and wont result in a crash. Such a cast won't even produce code.
Also assume we are not compiling with optimizations enabled and even with debug mode enabled. Then all those deleted pointers are set to some debug values like 0xDEADBEEF. Setting it to nullptr afterwards WILL HIDE ALL BUGS (like unplaned double deletion), since casting from nullptr to any type wont result in an error. Eather, it is just undefinded behaviour (compiler and platform implementation defined). So setting it to null will prevent from errors due (unwanted) double deletion or unwanted access.

The rule should be:
Is this pointer meant to be accessed after this line?
- Yes?: set it to nullptr, so we can check if the pointer is valid. Only access it if you know the pointer is valid. 
- No?  Better don't set it to nullptr, since it could hide bugs with wrongly written concurrent code or design and access mistakes.

LittleHuba:
This is not worth salvaging anymore. Closing as the rewrite will not need this anyway. Since we'll use plugin infrastructure anything not wanted can simply be left unlinked by users.

PR #1650, opened 2019-12-03
Image Insert Crash Resolved

### LINKED ISSUE #1648: Adding image crashes Xournal++
Version: 1.1.0+dev
OS: Linux (Kbuntu 18.04)
Desktop Environment: KDE
Build: Nov 30 2019, 10:06:11

Xournal++ crashes when I try to add an image in the document using the toolbar button while opening the File Choose dialog box

However I am able to drag and drop the image from the file system onto the document without any problems.

It seems that the crash is caused by the file choose dialog box, however I am able to open the file choose dialog boxes for Open, Save, Save As, Export As etc without any issue

MuTaTeD:
Resolved the application crash while adding an image to the document as reported in #1648
Fixes #1648

### DIFF
--- a/src/control/stockdlg/ImageOpenDlg.cpp
+++ b/src/control/stockdlg/ImageOpenDlg.cpp
@@ -24,7 +24,7 @@ auto ImageOpenDlg::show(GtkWindow* win, Settings* settings, bool localOnly, bool
 	}
 
 	GtkWidget* cbAttach = nullptr;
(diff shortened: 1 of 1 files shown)


LittleHuba:
Thanks for the PR!

Technius:
Regression caused by 985d094; only present in 1.1.0+dev versions.

PR #1663, opened 2019-12-12
Fixed possible PortAudio integration bugs

Febbe:
While fixing clang-tidy warnings, I found possible bugs I fixed:
- The AudioQueue's threads could be unaware of changes done in other threads, 
because those changes were done non atomic or not in a fixed memory ordering ensured by a mutex/lock.
- Same for VorbisConsumer and VorbisProducer
- Fixed possible ressource leak of the "SNDFILE_tag" - ressource
- Fixed memory leaks of undeleted std::thread's 
--
- Refactored all touched files.

### DIFF
--- a/src/control/AudioController.h
+++ b/src/control/AudioController.h
@@ -19,13 +19,16 @@
 #include <util/audio/AudioPlayer.h>
 #include <gui/toolbarMenubar/ToolMenuHandler.h>
 
-class AudioController
+class AudioController final
 {
 public:
-	AudioController(Settings* settings, Control* control);
-	virtual ~AudioController();
+	//Todo convert Pointers to reference (changes to control.cpp are neccessary)
+	AudioController(Settings* settings, Control* control)
+	 : settings(*settings)
+	 , control(*control)
+	{
+	}
 
-public:
 	bool startRecording();
 	bool stopRecording();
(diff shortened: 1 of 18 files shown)


src/control/AudioController.h
```diff
@@ -19,13 +19,15 @@
 #include <util/audio/AudioPlayer.h>
 #include <gui/toolbarMenubar/ToolMenuHandler.h>
 
-class AudioController
+struct AudioController final
```
LittleHuba: Why exactly are you changing this to a struct?

LittleHuba:
I haven't tested it yet on my machine. Will do so the next few days.

src/control/AudioController.h
Febbe: Because we write all member variables downside and constructors upside. Most constructors need to be public (in this case they need), so writing struct increases readability. Structs and classes are exactly the same except for the default visibility. So writing struct{...}; is the same as writing class{public:..}; struct is exactly what we want in all cases.

LittleHuba: I'm not really a fan of using structs here as it invites people to make attributes public by accident. Better would be to just reorder the classes, so that variables are top and everything else comes afterwards.

PR #1675, opened 2019-12-17
Zoom + Pen fix

StatErik:
Fix for issue #1668

I decided to call the reset method of the TouchInputHandler from HandRecognition, so it also works without the internal hand recognition turned on.

### DIFF
--- a/src/gui/inputdevices/HandRecognition.cpp
+++ b/src/gui/inputdevices/HandRecognition.cpp
@@ -145,7 +145,10 @@ void HandRecognition::penEvent()
 	if (touchState)
 	{
 		touchState = false;
-		disableTouch();
+		if (enabled)
+		{
+			disableTouch();
+		}
 		g_timeout_add(disableTimeout, reinterpret_cast<GSourceFunc>(enableTimeout), this);
 	}
 }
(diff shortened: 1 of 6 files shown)


src/gui/inputdevices/HandRecognition.cpp
```diff
@@ -155,14 +158,19 @@ void HandRecognition::penEvent()
...
+		{
+			inputContext->unblockDevice(InputContext::TOUCHSCREEN);
+		}
+		if (touchImpl)
+		{
+			touchImpl->enableTouch();
+		}
 	}
+	
+	inputContext->touchHandler->reset();
```
LittleHuba: I'm not really happy with this, as it is somewhere where it essentially shouldn't be. It should either be called in `InputContext::unblockDevice()` or even within the handler itself by overriding `AbstractInputHandler::block()`.

What you currently do is violating a few common programing paradigms. Namely the encapsulation paradigm. And violating this one leads to huuuge problems later on.

LittleHuba: Think about calling AbstractInputHandler::block() within the method when you override it. Otherwise you will break something else.

Febbe:
@LittleHuba is this PR now ok, or do you want to reject it?
@StatErik can you rebase those commits into one or two with
`git rebase -i HEAD~7`

LittleHuba:
Looks okay to me. I unfortunately can't check it currently. But the logic seems sound.

StatErik:
rebased

Febbe:
I couldn't find some bugs introduced by this, but it is not fixed completely.
The bug, that zooming isn't possible completely is fixed, but zooming and having the pen hovering is still broken. Also it does only work after the second zoom attempt.
@StatErik can you look into this?

Issue #975, opened 2019-03-10
Ctrl+G dialog is not usable


flyasky:
I'm drawing over long pdf. When I try to go to any page by Ctrl+G dialog  it opens  first page in document.

andreasb242:
I can not reproduce this - please try to describe the issue better. As long as we cannot reproduce this, we can not fix it...

flyasky:
1. Open any page in document
2. Gtrl+G and put number of any other page
3. Press Enter
4. You moved on first page (if Ok button used it will moves right)

andreasb242:
Ok, Enter is not working, but OK is. this was the Information I needed.

Technius:
This was fixed by #1678.

PR #1678, opened 2019-12-19
Enable Ctrl-G dialog to be closed by "Enter"

MrMallIronmaker:
This fixes bug #975.
Key changes:
- removed the "accelerator" that linked the enter key to the OK button. It appeared this forced the dialog to close before the value in the spin button was handled.
- used `gtk_entry_set_activates_default` to specify that when the Enter key is pressed in the spin box, the default option should be activated.
- included some settings that specified that the OK button was the one to activate by default (i.e. when the Enter key is pressed)

Additional changes:
- the Close button was changed to a Cancel button to be consistent with other dialogs
- Instead of the buttons creating values 1 and 2 (hardcoded), the OK and Close buttons response codes were replaced with the GTK-specified constants. (Unfortunately, this means there are weird -5 and -6 in the `.glade` file, but hopefully no one is manually editing those anyways...)

Two concerns:
- There are a few lines of fiddling I did that I'm not sure are necessary changes. Is it worth going back and finding the smallest sufficient set of changes?
- For some reason, opening and editing the `ui/goto.glade` file in Glade adds `type="action"` in places that seems to break the page. It works as it is now, but if anyone ever edits this `ui/goto.glade` file again, it won't work without manually removing those lines. I assume that's a problem, but I don't know enough about Glade to resolve it.

### DIFF
--- a/src/gui/dialog/GotoDialog.cpp
+++ b/src/gui/dialog/GotoDialog.cpp
@@ -16,10 +16,10 @@ auto GotoDialog::getSelectedPage() const -> int
 void GotoDialog::show(GtkWindow* parent)
 {
 	gtk_window_set_transient_for(GTK_WINDOW(this->window), parent);
-    gtk_entry_set_activates_default(GTK_ENTRY(get("spinPage")), TRUE);
+	gtk_entry_set_activates_default(GTK_ENTRY(get("spinPage")), TRUE);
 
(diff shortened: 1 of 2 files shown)


MrMallIronmaker:
All good! I was able to fix the `type=action` problem by converting to 3.18. I'm excited to have contributed!

LittleHuba:
LGTM
Thanks for the PR!

PR #1685, opened 2019-12-30
Initial Commit: Use std::filesystem or fallback to boost::filesystem

Febbe:
Please check if the licensing is ok.


Technius:
Licensing should be OK (Boost is compatible with GPLv2) but the mac build is failing.

PR #1687, opened 2019-12-30
adjusted formatting

Febbe:
I adjusted the readability-identifier-naming options to match our coding convention:
>* Names are CamelCase (namespaces, classes, structs, enums, etc.)
>* Members are camelBack
>* Functions are camelBack
>* No underlines are used

note: I left some aNy_CasE to give some versatility in source files. So 
> * No underlines are used

may not correct any more.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -42,259 +42,69 @@ HeaderFilterRegex: '*.(h|hpp|hxx)'
 AnalyzeTemporaryDtors: false
 CheckOptions:
   - key:             readability-identifier-naming.AbstractClassCase
-    value:           lower_case
-  - key:             readability-identifier-naming.AbstractClassPrefix
-    value:           ''
-  - key:             readability-identifier-naming.AbstractClassSuffix
-    value:           ''
+    value:           CamelCase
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -42,259 +42,69 @@ HeaderFilterRegex: '*.(h|hpp|hxx)'
...
-  - key:             readability-identifier-naming.EnumSuffix
-    value:           ''
+    value:           CamelCase
   - key:             readability-identifier-naming.FunctionCase
-    value:           lower_case
-  - key:             readability-identifier-naming.FunctionPrefix
-    value:           ''
-  - key:             readability-identifier-naming.FunctionSuffix
-    value:           ''
+    value:           camelBack
```
Technius: Will this break named requirements in some STL functions? e.g. `try_lock`?

Febbe: It will effect such functions, but clang-tidy will only report warnings/errors, in our code(, as long the CMakeLists.txt is configured correctly). We can write a `// NOLINT[readability-identifier-naming]` to suppress those warnings.

Issue #1221, opened 2019-06-08
Draw polygon tool


alexandredias3d:
**Is your feature request related to a problem? Please describe.**
Sometimes I want to highlight text with a different polygon than the rectangle tool. To do this, I have to select the line tool and draw the polygon manually without any help of the program to restrict movement like Inkscape and GIMP do. Also, I could not find a way to fill the contents of a "shape" created by the line tool.

**Describe the solution you'd like**
A new drawing tool that works similarly to how the *Draw bezier curves and straight lines (Shift + F6)* tool in Inkscape work could be added to Xournal++. This tool works as follows. The user chooses the first point.  After that, following points are chosen and they form a a line with the previous one. The last selected point closes the polygon.

**Additional context**
Here is a screenshot of this feature in Inkscape.
![A quadrilateral on the left and a triangle on the right](https://user-images.githubusercontent.com/25330126/59149035-aed21e00-89e6-11e9-8e2e-e24cbc6ce027.png)

It is also present on GIMP in the Free Select tool.

LittleHuba:
Sounds like a very useful tool.

Recommendation for implementation:
To stop drawing further points the line is finished when a double click is received.

Technius:
The line drawing part was implemented in #1688. Polygon fill has not been implemented yet.

Technius:
Actually, polygon fill _has_ been implemented (I tested just now), so I'm closing this issue.

PR #1688, opened 2020-01-02
add basic spline tool

rolandlo:
This is an implementation of a very basic spline tool. The latter was on the wishlist for new features, see  [issue #1221](https://github.com/xournalpp/xournalpp/issues/1221). So far it is only for linear splines (i.e. polgyon courses), but it can be extended from there on. The inkscape "Bezier curves and straight lines" tool can serve as a model for further development, although for a note taking tool it probably doesn't need to get as refined as for a vector graphics tool.

SnapToGrid is not yet implemented, nor is selectByTap. The (linear) spline is finished when a double click is received. It is my first commit, so I tried to keep it as basic as possible, hoping for some feedback from more experienced developers and collaboration.


src/control/tools/SplineHandler.cpp
```diff
@@ -0,0 +1,125 @@
...
+	page->fireElementChanged(stroke);
+
+	stroke = nullptr;
+
+	xournal->getCursor()->updateCursor();
+}
+
+void SplineHandler::drawShape(Point& c, const PositionInputData& pos) {
+	int count = stroke->getPointCount();
+	if (count > 1) {
```
Technius: Please add a comment indicating why this is necessary (e.g. that the last point is at the cursor position, which changes dynamically).

src/gui/PageView.cpp
```diff
@@ -498,8 +515,16 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
...
 
-        delete this->inputHandler;
-        this->inputHandler = nullptr;
+		ToolHandler* h = control->getToolHandler();
+		if (h->getDrawingType()==DRAWING_TYPE_SPLINE)
+		{
+			// Do NOT delete inputHandler yet, since drawing is not finished
+		} 
+		else 
+		{
```
Technius: Nitpick: use `if (h->getDrawingType() != DRAWING_TYPE_SPLINE)` and change the comment to indicate that the spline tool manually finalizes drawing.

rolandlo:
Thanks a lot for your helpful feedback!  I made the requested changes in the 2nd commit. 

- Now the splines can also be finalized via the escape key. For those like me using xournal++ often without keyboard, but with a stylus, I think the doublePressButton finalizing option is still useful. Also I don't mind that in this way a new vertex is created. One simply needs to think about it before placing the last vertex. It's the same behavior as in Inkscape, so it shouldn't be strange.

- While drawing, all vertices but the dynamically changing one are circled. The first vertex is drawn in a different color (red), because it is needed to close the spline (by clicking inside the red circle). After finalizing the spline it takes a moment until the circles are gone. I would prefer those circles to go away immediately, but I didn't figure out yet, how to do it. What redraw method should I call?

- As a next step I would implement the actions to move around or remove the last circled vertex via the arrow keys resp. the backspace key.

src/control/tools/SplineHandler.cpp
```diff
@@ -0,0 +1,210 @@
...
+SplineHandler::~SplineHandler() = default;
+
+void SplineHandler::draw(cairo_t* cr) {
+    if (stroke) {
+        double zoom = xournal->getZoom();
+        int dpiScaleFactor = xournal->getDpiScaleFactor();
+
+        cairo_scale(cr, zoom * dpiScaleFactor, zoom * dpiScaleFactor);
+        // draw circles around knot points
+        double lineWidth = 0.5;
```
Technius: In my opinion, the circles should scale with the zoom; they're difficult to see on larger screens.

src/control/tools/SplineHandler.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    Layer* layer = page->getSelectedLayer();
+
+    UndoRedoHandler* undo = control->getUndoRedoHandler();
+
+    undo->addUndoAction(mem::make_unique<InsertUndoAction>(page, layer, stroke));
+
+    layer->addElement(stroke);
+    page->fireElementChanged(stroke);
+    this->redrawable->repaintRect(stroke->getX(), stroke->getY(), stroke->getElementWidth(),
```
Technius: Use `rerenderRect` here. The difference is that `repaintRect` will draw to the view's buffer, while `rerenderRect` will draw the view's buffer to the canvas/screen (according to the documentation in `Redrawable.h`).

rolandlo: Thanks for the help!

src/control/tools/SplineHandler.cpp
rolandlo: This is now done.

Febbe:
I think finalizing on cancelation(e.g. changing tool) is better, since one does simply delete it afterwards. If it gets deleted instead, all the work is lost.

rolandlo:
I tried to make the requested changes. When the tool or drawingType is changed the spline is now deleted as soon as the draw method is called again. This happens in particular after motion is detected.
If the spline should disappear at once (after the tool/drawing type has changed), this would have to be called from the ToolHandler, I guess. I haven't figured out how to do it in that way.

rolandlo:
The last circled vertex can now be moved (via arrow keys) or deleted (via backspace key).

src/control/tools/SplineHandler.cpp
```diff
@@ -0,0 +1,275 @@
...
+        case GDK_KEY_BackSpace: {
+            if (pointCount > 2) {
+                // delete last non dynammically changing point
+                stroke->deletePoint(pointCount - 2);
+                this->redrawable->repaintRect(x, y, w, h);
+            }
+            break;
+        }
+        case GDK_KEY_Up: {
+            if (pointCount > 2) {
```
Technius: There's a lot of code duplication here. The common parts should be factored out into a helper function.

src/control/tools/SplineHandler.cpp
Technius: Oops, I wasn't clear enough. It should scale _inversely_ with zoom, such that if the zoom is decreased, the circles will increase in size (so that they will still be visible).

src/control/tools/SplineHandler.cpp
rolandlo: I have now factored the common parts from the arrow key events into a helper function called movePoint

src/control/tools/SplineHandler.cpp
rolandlo: I have changed that accordingly now. Please check if this looks good on higher resolution screens now.

rolandlo:
Tool switching now finalizes the spline. Is everything fine now?

PR #1691, opened 2020-01-04
Fix zoom fit infinitely recursing on page size change

### LINKED ISSUE #1660: Xournalpp freezes when changing following page size from A4 to A5
**Affects versions :**
 - OS: Kubuntu 18.04.3 LTS
 - KDE/PLASMA shipped with Kubuntu 18.04.3 LTS
 - libgtk 3.0 (3.22.30-1ubuntu4)
 - Version of Xournal++ 1.1.0+dev

**Describe the bug**
xournalpp window freezes, when changing the page size from a following page to a smaller format.

**To Reproduce**
1. Open new document.
2. Insert a second page.
3. Change the second page size from e.g. A4 to A5.

**Expected behavior**
xournal++ window freezes, after about 30s it terminates with the console error message 
"Speicherzugriffsfehler (Speicherabzug geschrieben)"

Shall I attach the core dump? It's a file of more than 5MB! There is no error log made by xournalpp.

**Additional context**
With the version 1.0.16 it's no problem to create a document with a second smaller page.
I have a few journals created with a 1.0.x version, that I can't open with 1.1.0+dev. When I try to open these old journals, the xournalpp window freezes.
When I unzipped these journals, changed all lines containing 

`<page width="611.99998270" height="791.99998270">`

to 

`<page width="595.27000000" height="841.89000000">`

then I can reopen these old journals with the 1.1.0+dev version. The freezing seems to belong to different page sizes.

Technius:
~Fixes~ Actually, this is a partial fix for #1660, but the freezing issue only. There is still a bug causing the zoom/scroll to "ping-pong" when scrolling across pages of different sizes.

This is kind of a hacky solution that doesn't cover all cases at the moment (there may still be freezing).

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -173,8 +173,8 @@ auto ZoomControl::getZoom() const -> double { return this->zoom; }
 
 auto ZoomControl::getZoomReal() const -> double { return this->zoom / this->zoom100Value; }
 
-void ZoomControl::setZoom(double zoom) {
-    this->zoom = zoom;
+void ZoomControl::setZoom(double newZoom) {
+    this->zoom = newZoom;
     fireZoomChanged();
 }
 
@@ -256,7 +256,7 @@ void ZoomControl::zoom100() {
 
 void ZoomControl::zoomFit() {
     if (this->zoomFitMode && !this->zoomPresentationMode && this->zoom != this->zoomFitValue) {
-        startZoomSequence(-1, -1);
+        this->startZoomSequence(-1, -1);
         this->zoomSequenceChange(this->zoomFitValue, false);
         endZoomSequence();
     }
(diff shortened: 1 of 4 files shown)


Technius:
This bug is very tricky to fix: for example, suppose that there are two pages, the first page larger than the other. The larger page is initially selected; the user attempts to scroll down. The smaller page then occupies most of the viewport, causing the zoom to fit to the smaller page. But then this causes the larger page to occupy most of the viewport, shifting selection back and causing zoom to change again.

With the `updatingZoomFit` hack, this causes the user to be unable to scroll down to the smaller page. Without it, the window will freeze due to infinite recursion.

Technius:
I _think_ the recent changes should fix the freezing issues in zoom fit and presentation mode. I removed the `updatingZoomFit` hack because it didn't actually do anything useful; the new solution, which is just to defer scrolling updates, works much more nicely. Scrolling is still janky though: scroll too fast and the page selection will jump around. There may also be issues when scrolling down from a smaller page to a larger page.

We'll probably need to rewrite the zoom/scroll system so that zooming and scrolling is performed simultaneously, not separately. The fact that the page views have inconsistent sizes after zooming makes it difficult to do anything with the layout after a zoom (e.g. after the zoom changes, an `XojPageView`'s x and y coordinates are not properly adjusted for the new zoom value, causing `Layout::updateVisibility` to select the wrong page).

Technius:
Handling this in #1706.

Issue #683, opened 2019-01-03
ubuntu: touch screen not drawing


gregdymek:
drawing not working on the touchscreen, it's scrolling instead, normal xournal is working ok however icons are too small on uhd screen

gregdymek:
I'm using hp spectre x360

andreasb242:
But your tablet has also pen input, right? And this does drawing?

Xournal is still on GTK2, Xournal++ is on GTK3, and the GTK3 Scrolled Container automatically takes the Touch event for Scrolling. The Xournal Widget only gets an event where you touch ("click"), but not the dragging. And this is not an option, this is hardcoded in GTK3.

I already noticed this (it's the same behavoir on all devices), and I already tried to make this configurable.
So you need a really good reason, why you would use the touch for drawing, else I probably wont spend time to this.

Stiffy2000:
Many people could be using a tablet + Xournal++ to do notes and the like.
Maybe if there is another touch handler we can use?

andreasb242:
We can remove the scrolled panel and to it ourselv, with this we remove all standard scrolling stuff. We had this before. And then let the user select.

morrolinux:
@andreasb242 I have to say that after the last X11 update my tablet's stylus gets (erroneously) detected as a touchscreen device (by x11, not a xournal++ bug). With the actual status I can configure it under the "Input" section of xournal, but cannot use it to write because of this.
I think indeed that could be useful to let the user pick one "touchscreen" device from the "Input" tab, and handle it manually.
would this be possible?

andreasb242:
@morrolinux We can do this, but this also means to disable all gestures, and it wont fix your issue, because on the Touschreen no pressure is provided.

morrolinux:
Umh I see... wouldn't be too much of a problem for pen pressure as I always disable this (me personally of course) but with no gestures.. yeah.. it would have been useful if it was possible to do it just "per-device"

andreasb242:
If it would be work, it may also solves #404 if enabled.

andreasb242:
By the way, it's working on Windows, at least on the cheap chinese 7" Tablets...

andreasb242:
It's already planed. I'll hope I'll can fix this soon.

andreasb242:
Technical I removed the GTK Scroll handling, therefore removed the full Touch scrolling, Scroll wheel, Scrollbars etc.
[...]

Issue #1694, opened 2020-01-05
Presentation mode incorrectly enabled after program start


rolandlo:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Ubuntu 19.10
 - X11
 - libgtk 3.24.12.1
- Version of Xournal++: 1.1.0+dev

**Describe the bug**
When the settings.xml file has presentationMode enabled, Xournal++ starts with scrolling like in presentationMode, but zoomed to full page width and with presentationMode not ticked in the menu entry.

**To Reproduce**
1. Set view to presentation mode (via F5 key)
2. close Xournal++
3. open new instance of Xournal++

**Expected behavior**
the new instance of Xournal++ is set in presentation mode, the corresponding menu entry is ticked and the document is zoomed like in presentation mode

**Additional context**
The console shows a warning (xournalpp:27218): WARNING **: 23:20:47.216: Error calculating zoom value. This warning is issued by the setViewPresentationMode method in Control.cpp

rolandlo:
It appears that replacing "false" with "true" in
`    if (!page) {
        // no page
        return false;
    }
`
(lines 219-222 in ZoomControl::updateZoomPresentationValue) would solve this issue. I hope it doesn't create other problems, though. So far, I don't see any. The user is not allowed to delete the last page.

rolandlo:
Fixed by PR #2413

Issue #1433, opened 2019-08-21
Refactor XournalView, XournalWidget, ZoomControl


lehmanju:
This should also solve issues with wrong scroll position after window resize etc. (#1411)

My plan is:
- use `GtkDrawingArea` instead of custom widget and only listen to signals we need
- recalculate `Layout` only when needed (scroll position updates or widget area resizes)
- XournalView and PageView should properly do the rendering

elivanf:
I have a problem viewing the files.
When the page is not fully visible the file is not displayed.
How to solve this problem?

LittleHuba:
@elivanf please open a separate ticket for this with a more detailed description of the problem. Best would be steps we can reproduce. Thanks!

elivanf:
Sorry I didn't know how to use the tool to solve my problem. Thanks.

PR #1706, opened 2020-01-16
Viewport rewrite: Round 1

Technius:
As I've mentioned in multiple issues, we'll need to refactor some of the rendering/zooming/scrolling code to fix the outstanding bugs and to prepare for the new file format features. Basically, we can start out by refactoring some of the existing code, i.e. dealing with #1433.

This PR removes the custom GTK widget in `src/gui/widgets/XournalWidget.cpp` and replaces it with a wrapper around a `GtkDrawingArea`.

TODO:
* [x] Renders as expected
* [x] Fix scroll bar adjustments: the GTK adjustments seemed to be set correctly for the older widgets, but not for the drawing area
* [x] Accommodate touchscreen drawing
* [ ] Test for regressions

Notes:
* This PR will be as minimal as possible since it touches a lot of files.
* `GtkDrawingArea` is connected to the `GDK_CONFIGURE` event by default, which is fired on window visibility changes. This is useful for figuring out when to re-layout the document, but it's also causing several new problems.
* #1694 shows up here, whereas it didn't with the previous widget. Probably related to the `GDK_CONFIGURE` thing.
* I also fixed an edge case in input handling: it used to assume every event has a device, but `GDK_CONFIGURE` does *not* have a device!
* This PR changes the default scroll handling to use the updated `GtkDrawingArea` scroll handling. "Enable touch workaround" needs to be renamed to "Enable touch drawing." Consequently, application restart is no longer required to change this setting. As far as I tested with simulated inputs, touchscreens seem to work properly with default scroll handling, and `ScrollhandlingXournalpp` is probably obsolete now (see https://github.com/xournalpp/xournalpp/pull/1706#issuecomment-575968020).

**Next steps:**

You may have noticed "Round 1" in the title of this PR. The next step would be to unify the zooming and scrolling code into a "viewport" handler, since we fundamentally cannot handle zooming and scrolling separately (see https://github.com/xournalpp/xournalpp/pull/1691#issuecomment-570880995).

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -231,8 +231,8 @@ void MainWindow::initXournalWidget() {
     }
     // Todo configure-event
 
-    Layout* layout = gtk_xournal_get_layout(this->xournal->getWidget());
-    scrollHandling->init(this->xournal->getWidget(), layout);
+    Layout* layout = this->xournal->getWidget()->getLayout();
+    scrollHandling->init(this->xournal->getWidget()->getGtkWidget(), layout);
 }
 
 void MainWindow::setTouchscreenScrollingForDeviceMapping() {
@@ -283,7 +283,7 @@ void MainWindow::initHideMenu() {
     }
 }
 
-auto MainWindow::getLayout() -> Layout* { return gtk_xournal_get_layout(GTK_WIDGET(this->xournal->getWidget())); }
+auto MainWindow::getLayout() -> Layout* { return this->xournal->getWidget()->getLayout(); }
 
 auto cancellable_cancel(GCancellable* cancel) -> bool {
     g_cancellable_cancel(cancel);
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -193,6 +193,8 @@ void MainWindow::toggleMenuBar(MainWindow* win) {
 void MainWindow::initXournalWidget() {
     GtkWidget* boxContents = get("boxContents");
 
+    // TODO: Rename "Touch Workaround" to "Touch Drawing"; remove deprecated scroll handling and add option to always
+    // show scrollbars
     if (control->getSettings()->isTouchWorkaround()) {
         GtkWidget* box1 = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
         gtk_container_add(GTK_CONTAINER(boxContents), box1);
@@ -236,6 +238,8 @@ void MainWindow::initXournalWidget() {
 }
 
 void MainWindow::setTouchscreenScrollingForDeviceMapping() {
+    gtk_scrolled_window_set_kinetic_scrolling(GTK_SCROLLED_WINDOW(winXournal),
+                                              !this->getControl()->getSettings()->isTouchWorkaround());
     for (InputDevice const& inputDevice: DeviceListHelper::getDeviceList(this->getControl()->getSettings())) {
         InputDeviceClass deviceClass = InputEvents::translateDeviceType(inputDevice.getName(), inputDevice.getSource(),
                                                                         this->getControl()->getSettings());
--- a/src/gui/widgets/XournalWidget.h
+++ b/src/gui/widgets/XournalWidget.h
@@ -13,12 +13,6 @@
 
 #include <gtk/gtk.h>
 
-G_BEGIN_DECLS
-
-#define GTK_XOURNAL(obj) G_TYPE_CHECK_INSTANCE_CAST(obj, gtk_xournal_get_type(), GtkXournal)
-#define GTK_XOURNAL_CLASS(klass) GTK_CHECK_CLASS_CAST(klass, gtk_xournal_get_type(), GtkXournalClass)
-#define GTK_IS_XOURNAL(obj) G_TYPE_CHECK_INSTANCE_TYPE(obj, gtk_xournal_get_type())
-
 class AbstractInputDevice;
 class EditSelection;
 class Layout;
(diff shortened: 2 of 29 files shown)


Technius:
Two things in this latest update. The scroll issue was fixed by resizing the drawing area on layout change. I added some doc comments since this part of the code is undocumented and hard to understand at a first glance. Note that this affects the default / non-touch-workaround scrolling.

For the "Touch Workaround" scrolling: the bad news is that it doesn't work anymore and I can't figure out a non-trivial way to make it work. The good news is that the "Touch Workaround" scrolling (added in #797 and meant to fix #683) is probably obsolete now. I noticed that the `kinetic-scrolling` property in `GtkScrolledWindow` can be set to false to disable scrolling on touch screen dragging. I went ahead and did so, and scrolling seems to work property when touchscreen input emulation is enabled in the GTK inspector. I can't test it on a real device, though.

Technius:
Update on "Touch Workaround": it didn't work because I had zoom fit enabled (which seems to be breaking everything...). After disabling zoom fit, the scrolling worked as expected.

Febbe:
Wasn't the touch workaround deprecated anyway since gtk+-3.x? And was supposed to be removed?

Technius:
> Wasn't the touch workaround deprecated anyway since gtk+-3.x? And was supposed to be removed?

I actually have no idea. I think deprecation was planned in #1409 but it was never removed. My attempted fix didn't work (https://github.com/xournalpp/xournalpp/issues/683#issuecomment-575969619) and some people still need the workaround. I can't effectively come up with a fix because I don't have any devices with touchscreens.

Febbe:
Mhh I have one. And I always had troubles with the touch workaround. Maybe I can look I to that. Anyway there must be a better way as the current solution.

Technius:
I another a look at the input code, and I found that there is a semantic difference in the way the input is handled with the touch workaround enabled.

https://github.com/xournalpp/xournalpp/blob/8052bee481f202f3d148ec8337f560e3719dce2f/src/gui/inputdevices/InputContext.cpp#L110-L116

I tested this fact by setting my mouse's device class to touchscreen. With the workaround off, I couldn't  draw anything, which seems like the correct behavior if I only wanted to scroll with a touchscreen. With the workaround on, I was able to make strokes.

It's probably safe to delete the `ScrollHandlingXournal` and just change the workaround option to choose between scroll/edit input handlers. But again, I can't test this because I don't have a touchscreen.

Technius:
Touch drawing works as far as I am able to test it with simulated touch inputs (see the second commit for the changes I made). I've left the "Touch Workaround" scroll handler in for now, but we could deprecate and remove it either here or in a followup PR.

src/gui/MainWindow.cpp
```diff
@@ -244,6 +246,7 @@ void MainWindow::setTouchscreenScrollingForDeviceMapping() {
             break;
         }
     }
```
Technius: We might not need this block of code anymore with the updated touch settings, but I'm not 100% sure.

src/gui/MainWindow.cpp
```diff
@@ -193,7 +193,9 @@ void MainWindow::toggleMenuBar(MainWindow* win) {
 void MainWindow::initXournalWidget() {
     GtkWidget* boxContents = get("boxContents");
 
-    if (control->getSettings()->isTouchWorkaround()) {
+    // TODO: Awaiting review
+    if (false) {
+//    if (control->getSettings()->isTouchWorkaround()) {
```
Technius: This is just a quick hack to force the updated scroll handling. I will remove this block entirely if we decide to deprecate the touch workaround.

lehmanju:
As far as I know the touch workaround is supposed to work for devices that expose a "normal mouse pointer" to GTK which in reality is a touchscreen. To make the touchscreen work those events need to be translated. The original touch workaround did exactly that. However, with the new input system you can also change device classes of you input devices in settings. This basically enables a new touch workaround by simply changing the device class of your touchscreen to "Touchscreen".

Technius:
Ok, that's good to know.

>  Deleting the old touch workaround is therefore no problem at all, has to be done in conjunction with deleting the old input system though.

Why is this the case? As far as I've tested, the old input system is compatible with the changes I've made. Yes, they should both be deprecated, but in a follow-up PR and not this one.

src/gui/MainWindow.cpp
Technius: Actually, the added line should go above this, since if we set a touchscreen to a different device class, there should be no kinetic scrolling regardless if touch draw is disabled or not.

Febbe:
> Why is this the case? As far as I've tested, the old input system is compatible with the changes I've made. Yes, they should both be deprecated, but in a follow-up PR and not this one.

Maybe we should remove it first and then make this PR as followup. So this PR can fully handle the Layout issue without interfering with other sections (input).

src/gui/widgets/XournalWidget.h
```diff
@@ -29,63 +23,46 @@ class XournalView;
...
+    XournalWidget(XournalView* view, std::unique_ptr<InputContext> inputContext);
+    XournalWidget(XournalView* view, ScrollHandling* scrollHandling);
+    virtual ~XournalWidget();
 
-    /**
-     * Scrollbars
-     */
-    ScrollHandling* scrollHandling;
+    /** Current selection */
+    EditSelection* selection;
```
Febbe: This is undefined behaviour, since selection is used after initialisation.
Todo: Either use smart pointers or initialize the variable with `EditSelection* selection{}`;

Technius: Good catch, I missed this when refactoring. I will probably get around to updating the PR this weekend.

Technius: The best thing we can do right now (without refactoring literally everything else) is to initialize `selection` to null, since `XournalView::setSelection` passes in a raw pointer and assumes that the widget will take ownership.

src/gui/MainWindow.cpp
Technius: Doing it in followup PR.

Technius:
Going to merge this into a `view-rewrite` branch tomorrow, since all the work is not done yet.

PR #1709, opened 2020-01-19
WIP: test std::deque for Strokes

### LINKED ISSUE #1707: Pen lag upon starting a stroke
linux Elementary OS
libgtk-3-bin	3.22.30-1ubuntu4
Xournal++ 1.1.0 +dev

When using the pen input at the start of the stroke the pen output is laggy and inaccurate for about 500-700 ms from which onwards the stroke is smooth and fluent

This happens with the start of every pen stroke.
Pen input works fine with other apps with no lag so I don't think it's input side.

I've tried all new input settings (on and off) but to no avail.

Febbe:
may fix #1707

### DIFF
--- a/src/control/shaperecognizer/CircleRecognizer.cpp
+++ b/src/control/shaperecognizer/CircleRecognizer.cpp
@@ -45,7 +45,8 @@ auto CircleRecognizer::scoreCircle(Stroke* s, Inertia& inertia) -> double {
     double x0 = inertia.centerX();
     double y0 = inertia.centerY();
 
-    auto const& pv = s->getPointVector();
+    auto&& [begi, endi] = s->getPoints();
+    std::vector<Point> pv{begi, endi};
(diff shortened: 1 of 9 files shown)


Febbe:
This PR realy seems to make a different, but I don't know why it only works in combination with the touch and scroll workaround.

Technius:
Is this with or without optimizations enabled?

I think we need a systematic way to profile, otherwise we're just playing optimization guessing games.

The touch and scroll workaround renders differently and could be slower, which could result in visible performance improvements from using a deque (this is pure conjecture).

fsimonis:
I had a deeper look at this and tried to profile the time of the insertion. This is <1 microsecond on my system for general additions and up to 100 microseconds for insertions with reallocation on paths longer than 4000 points.

So the insertion really isn't a bottleneck here.
I also checked the time from the event handler to the insertion and they took about 150 microseconds.

Thus the data structure is not an issue here.

PR #1749, opened 2020-02-22
Avoid overwriting settings file during its parsing

Guldoman:
Fixes xournalpp/xournalpp#1074


Technius:
LGTM, thanks for the PR.

Issue #1746, opened 2020-02-18
Dark theme not working on Windows


R4ygen:
Hello, i just installed the latest **release** of **Xournal++ (1.0.17)** on my **Windows 10 Pro** machine, as an alternative to OneNote.
So far so good apart for one thing: I can't make the dark theme working.

If i check the option under _Preferences -> View -> Dark theme_ and reboot Xournalpp, nothing changes.
![Immagine](https://user-images.githubusercontent.com/49517480/74753680-c4376180-5270-11ea-89d5-266fa6d972b6.jpg)

So far i've been able to change the page color and the background, but i would like to change the toolbar, menu and window colors too.

I've found some config files in the directory _\Program Files\Xournal++\ui_ but i couldn't figure out if there's some values i can manually change there to edit the theme colors.
I've already tried to reinstall the program (always through the release automated installer) and check for steps i could have skip in the installation instructions and FAQs, but nothing.

Thanks in advance

Technius:
Themes are controlled by GTK, so if you can somehow set the environment variable `GTK_THEME=Adwaita:dark` before launching Xournal++, it should show up dark. Unfortunately, this isn't something we can do within Xournal++ itself, but we might be able to change the Windows launcher application (`xournalpp_launcher.exe`) to make it easier.

Apparently the "Dark Theme" setting only changes the icons: https://github.com/xournalpp/xournalpp/blob/8052bee481f202f3d148ec8337f560e3719dce2f/src/control/XournalMain.cpp#L322-L325

R4ygen:
Oh i just got confused by the option name then, i though it as a whole dark theme.
Not a bug at least.

Well since it's a missing feature this could become a low priority enhancement request, yeah.
Sadly if it's more complex than editing a config file is still above my technical level. But i'll try to google a little about it.

Technius:
> Oh i just got confused by the option name then, i though it as a whole dark theme.
> Not a bug at least.

I would say it's a bug since it's misleading; it should really be renamed as "Use dark mode icons."

badshah400:
This setting seems to be not working at all — even for toolbar icons — on my Linux installation. I have turned this setting on, and launched xournalpp (at commit 46a4dc82) with
> GTK_THEME=Adwaita:dark xournalpp

Nonetheless, I see the same toolbar icons being used (i.e. not the icons from iconsDark). See, e.g., the icon for the Text tool or the brush sizes both of which should be white in colour if used from iconsDark.

[...]

PR #1767, opened 2020-03-11
Prepend iconsDark path after setting default path.

badshah400:
When the "dark theme" option is set, the iconsDark directory should be
the first dir where the icon theme should search for the icon. By
prepending this path after the default icons path has been set, we make
sure that this dir is indeed the first in the list of dirs to look up
icons.

Fixes issue #1746.


Technius:
LGTM, thanks for the PR.

PR #1774, opened 2020-03-16
Remove Y adjustment in XournalView::scrollTo

### LINKED ISSUE #1764: PDF in presentation mode, scroll doesn't move by a PDF page exactly.
- OS: Debian Sid
 - Desktop environment: none, display run by X11 not Wayland
 - Which version of libgtk do you use: ?
 - Version of Xournal++: 1.0.17
 - Installation method: debian package (as provided by xournalpp developers)

My use case:
  I want to present a PDF, full screen, one page a time, occasionally drawing-scribbling on it
My solution:
  full screen, presentation mode, zoom fit to height
My problem:
  scrolling - whenever I scroll using, e.g.,  PgDown the scroll amount is less than one page and larger and larger portions of the previous page are left in sight. 

I have to remark that `xournal` behaves perfectly with respect to my use case, possibly because it has an option to turn off the continuous display of a document that `xournalpp` apparently has dropped, so it's not a show stopper for me.

ohthehugemanatee:
This fixes the page scrolling units issue for me.

I'm honestly not clear on what the `yDocument` variable was supposed to represent, and why it was always added to the y dimension of new page positions. At least in my testing, this patch makes all "go to page" operations align the viewer with the *top* of the new page, rather than "sort of" the bottom. 

Closes #1764

### DIFF
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -391,7 +391,7 @@ void XournalView::scrollTo(size_t pageNo, double yDocument) {
     Layout* layout = gtk_xournal_get_layout(this->widget);
 
     int x = v->getX();
-    int y = v->getY() + std::lround(yDocument);
+    int y = v->getY();
     int width = v->getDisplayWidth();
     int height = v->getDisplayHeight();
 


ohthehugemanatee:
@Technius  any chance of a review on this one-line PR?

Technius:
I don't think this is correct. The `yDocument` is the offset from the top of the page. We need this for the Ctrl+F search bar, for example. If we ignore it, the scroll will just go to the top of the page instead of to the searched text.
https://github.com/xournalpp/xournalpp/blob/46a4dc8292fbca0bce24b8ade6dceee83db16791/src/gui/SearchBar.cpp#L119-L121

ohthehugemanatee:
Aha, thank you for explaining that! So really, in page next/prev, `yDocument` should be zero. That would get us the same behavior. On it.

PR #1779, opened 2020-03-19
Switch CI to build on MacOS 10.15

Technius:
Azure Pipelines is phasing out 10.13

Merging if build succeeds.

### DIFF
--- a/azure-pipelines/continuous-integration.yml
+++ b/azure-pipelines/continuous-integration.yml
@@ -54,7 +54,7 @@ stages:
 
       - job: macOS
         pool:
-          vmImage: 'macOS-10.13'
+          vmImage: 'macOS-10.15'
         displayName: 'Test Xournal++ on MacOS'
         steps:
           - template: steps/build_mac.yml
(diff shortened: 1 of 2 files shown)


Technius:
@LittleHuba The build script failed to uninstall Homebrew. Is Homebrew uninstalled because it messes up the build?

By the way, the 10.13 image is going to be removed on March 23, which is only a few days away. So we need to get this upgraded ASAP (or else no MacOS nightlies in the near future).

LittleHuba:
Yeah Homebrew messes up the build. So installing it is necessary. Read up the build of GTK+ to check this out. You will also have to update the GTK+ blob used to macOS 10.15 otherwise it will fail as well.

Technius:
GTK supports a build through `meson` instead of `jhbuild`, so we may not need to uninstall Homebrew (I don't have enough experience with MacOS to debug this).

LittleHuba:
I might be able to dust of my VM to set this up once more. Will check in the next few days.

Technius:
I'm going to close this for now and try again with a more serious attempt.

PR #1785, opened 2020-03-20
Add Mouse+Keyboard device class

### LINKED ISSUE #1769: Text input broken with wireless mouse+keyboard combo
**Affects versions :**
 - OS: Linux
 - Desktop environment: Gnome X11
 - Which version of libgtk do you use: libgtk-3-0/bionic-updates,now 3.22.30-1ubuntu4 amd64 
 - Version of Xournal++: 1.0.17
 - Installation method: compiled locally

**Describe the bug**
The text input does not work.

You can type the text in another window, ctrl-c and ctrl-v **do** work. But typing directly is dead (?)
Will try to recompile, but... this is quite strange. 

**To Reproduce**
1. Launch xournalpp
2. choose T
3. click in the canvas
4. try to type anything

**Screenshots of Problem**
Peek 2020-03-13 09-32

Technius:
Fixes #1769 by introducing a "Mouse + Keyboard Combo" device class. This also includes a refactoring of the input device class settings code.


Rmano:
Checked with a Logitech Keyboard + Combo ant it works here; it fixes issue #1769 for me.

src/gui/inputdevices/KeyboardInputHandler.cpp
```diff
@@ -13,10 +13,10 @@ KeyboardInputHandler::KeyboardInputHandler(InputContext* inputContext): Abstract
 KeyboardInputHandler::~KeyboardInputHandler() = default;
 
 auto KeyboardInputHandler::handleImpl(InputEvent* event) -> bool {
-    auto keyEvent = reinterpret_cast<GdkEventKey*>(event->sourceEvent);
     GtkXournal* xournal = inputContext->getXournal();
 
-    if (keyEvent->type == GDK_KEY_PRESS) {
+    if (event->sourceEvent->type == GDK_KEY_PRESS) {
+        auto keyEvent = reinterpret_cast<GdkEventKey*>(event->sourceEvent);
```
LittleHuba: Could you remove the indirection for ever call and store the source event in a variable? It would prevent future accidential uses of the wrong event.

ui/settingsDeviceClassConfig.glade
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.22.1 -->
 <interface>
-  <requires lib="gtk+" version="3.0"/>
+  <requires lib="gtk+" version="3.18"/>
```
LittleHuba: Downgrade gtk+ library for backwards compatibility.

Technius: I'm not sure this matters given that we already require GTK >= 3.18, but I've changed it back for now.

src/gui/inputdevices/KeyboardInputHandler.cpp
Technius: Not entirely sure what you meant, but I changed the code to use `GdkEvent* gdkEvent = event->sourceEvent;` and then replaced every `event->sourceEvent` with `gdkEvent`.

LittleHuba: That was exactly what I meant.

ui/settingsDeviceClassConfig.glade
LittleHuba: Ahh then it would not have mattered. We had a problem a while back with a too new version string here causing issues. We should maybe make these consistent in a future update.

PR #1792, opened 2020-03-23
Fix warnings calling system in callbacks without checking the result.…

palvaradomoya:
…  For that task the wrapper Util::systemWithMessage was created.

### DIFF
--- a/src/util/Util.cpp
+++ b/src/util/Util.cpp
@@ -1,6 +1,7 @@
 #include "Util.h"
 
 #include <array>
+#include <cstdlib>
 #include <utility>
 
 #include <unistd.h>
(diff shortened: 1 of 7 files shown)


src/util/Util.cpp
```diff
@@ -160,3 +161,10 @@ void Util::writeCoordinateString(OutputStream* out, double xVal, double yVal) {
     g_ascii_formatd(coordString.data(), G_ASCII_DTOSTR_BUF_SIZE, Util::PRECISION_FORMAT_STRING, yVal);
     out->write(coordString.data());
 }
+
+void Util::systemWithMessage(const char* command) {
+    if (std::system(command) != 0) {
+        string msg = FS(_F("Error executing system command: {1}") % command);
```
Febbe: Can you add the errorcode
```
if (auto errc = std::system(command); errc != 0){
string msg = FS(_F("Error executing system command: {1} with Error {2}") % command % errc);
}
```

palvaradomoya: Hallo Fabian, I did your suggested changes, with a slightly different message.  I also squashed the commits.

Febbe: Ok thanks, this looks good for me, but you created a bit of a mess with your squash attempt.
I fixed it for you.

PR #1798, opened 2020-03-25
Implementation of cubic splines

rolandlo:
This commit implements cubic splines. At every knot you can adjust the tangent by click-dragging. The longer the drawn tangent, the stronger its influence. If all tangents are trivial (only click, no click-drag), it will be a linear spline as before.
The last set knot and the last set tangent can be modified using the keyboard (e, s, r, b for extend, shrink, rotate and back rotate).
If the spline is closed the tangent of the first knot will be used for the last knot as well, so the join will be smooth (if the tangent of the first knot is non-trivial).

### DIFF
--- a/src/control/tools/SplineHandler.cpp
+++ b/src/control/tools/SplineHandler.cpp
@@ -9,66 +9,91 @@
 #include "undo/InsertUndoAction.h"
 #include "util/cpp14memory.h"
 
-/**
- * @brief A class to handle splines
- *
- * Drawing of a spline is started by a ButtonPressEvent. After every ButtonReleaseEvent,
- * a new knot point is added. The spline is finished through a ButtonDoublePressEvent,
- * the escape key or a ButtonPressEvent near the first knot. The latter event closes the spline.
- *
- * Splines segments can be linear or cubic (as in Inkscape). Join of two cubic segments
- * is supposed to be smooth. As for now only linear splines are implemented.
- */
 SplineHandler::SplineHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page):
         InputHandler(xournal, redrawable, page) {}
 
 SplineHandler::~SplineHandler() = default;
 
+constexpr double RADIUS_WITHOUT_ZOOM = 10.0;
+constexpr double LINE_WIDTH_WITHOUT_ZOOM = 2.0;
+
 void SplineHandler::draw(cairo_t* cr) {
-    if (!stroke) {
+    if (!stroke || this->knots.empty()) {
         return;
     }
 
     double zoom = xournal->getZoom();
-    double radius = radiusConst / zoom;
+    double radius = RADIUS_WITHOUT_ZOOM / zoom;
     if (xournal->getControl()->getToolHandler()->getDrawingType() != DRAWING_TYPE_SPLINE) {
         g_warning("Drawing type is not spline any longer");
-        finalizeSpline();
-        //        stroke = nullptr;
-        //        xournal->getCursor()->updateCursor();
+        this->finalizeSpline();
+        this->knots.clear();
+        this->tangents.clear();
         return;
     }
 
     int dpiScaleFactor = xournal->getDpiScaleFactor();
     cairo_scale(cr, zoom * dpiScaleFactor, zoom * dpiScaleFactor);
 
-    // draw circles around knot points
-    double lineWidth = 2 / zoom;
-    int pointCount = stroke->getPointCount();
+    double lineWidth = LINE_WIDTH_WITHOUT_ZOOM / zoom;
     cairo_set_line_width(cr, lineWidth);
-    Point firstPoint = stroke->getPoint(0);
-    Point currPoint = stroke->getPoint(pointCount - 1);
-    double dist = currPoint.lineLengthTo(firstPoint);
-    cairo_set_source_rgb(cr, 1, 0, 0);                               // use red color for first knot
-    cairo_move_to(cr, firstPoint.x + radius, firstPoint.y);          // move to start point of circle arc;
-    cairo_arc(cr, firstPoint.x, firstPoint.y, radius, 0, 2 * M_PI);  // draw circle
-    if (dist < radius && pointCount > 3) {  // current point lies within the circle around the first knot
+    const Point& firstKnot = this->knots.front();
+    const Point& lastKnot = this->knots.back();
+    const Point& firstTangent = this->tangents.front();
+    const Point& lastTangent = this->tangents.back();
+    double dist = this->currPoint.lineLengthTo(firstKnot);
+
+    // draw circles around knot points
+    cairo_set_source_rgb(cr, 0.3, 0.3, 0.3);           // use gray color for all knots except first one
+    for (auto p: knots) {                              // circle all knots, circle around first knot will be redrawn
+        cairo_move_to(cr, p.x + radius, p.y);          // move to start point of circle arc;
+        cairo_arc(cr, p.x, p.y, radius, 0, 2 * M_PI);  // draw circle
+    }
+    cairo_stroke(cr);
+    cairo_set_source_rgb(cr, 1, 0, 0);                             // use red color for first knot
+    cairo_move_to(cr, firstKnot.x + radius, firstKnot.y);          // move to start point of circle arc;
+    cairo_arc(cr, firstKnot.x, firstKnot.y, radius, 0, 2 * M_PI);  // draw circle
+    if (dist < radius && this->getKnotCount() > 1) {  // current point lies within the circle around the first knot
         cairo_fill(cr);
     } else {
         cairo_stroke(cr);
     }
 
-    cairo_set_source_rgb(cr, 0.3, 0.3, 0.3);       // use gray color for all knots except first one
-    for (size_t i = 1; i < pointCount - 1; i++) {  // dynamically changing knot is not circled
-        Point p = stroke->getPoint(i);
-        cairo_move_to(cr, p.x + radius, p.y);          // move to start point of circle arc;
-        cairo_arc(cr, p.x, p.y, radius, 0, 2 * M_PI);  // draw circle
+    // draw dynamically changing segment
+    cairo_set_source_rgb(cr, 0.3, 0.3, 0.3);  // use gray color
+    const Point& cp1 = Point(lastKnot.x + lastTangent.x, lastKnot.y + lastTangent.y);
+    const Point& cp2 = (dist < radius && this->getKnotCount() > 1) ?
+                               Point(firstKnot.x - firstTangent.x, firstKnot.y - firstTangent.y) :
+                               this->currPoint;
+    SplineSegment changingSegment = SplineSegment(lastKnot, cp1, cp2, this->currPoint);
+    changingSegment.draw(cr);
+
+    // draw dynamically changing tangent
+    cairo_move_to(cr, lastKnot.x - lastTangent.x, lastKnot.y - lastTangent.y);
+    cairo_line_to(cr, lastKnot.x + lastTangent.x, lastKnot.y + lastTangent.y);
+
+    cairo_stroke(cr);
+
+
+    // draw other tangents
+    cairo_set_source_rgb(cr, 0, 1, 0);
+    for (size_t i = 0; i < this->getKnotCount(); i++) {
+        cairo_move_to(cr, this->knots[i].x - this->tangents[i].x,
+                      this->knots[i].y - this->tangents[i].y);  // draw dynamically changing segment
+        cairo_line_to(cr, this->knots[i].x + this->tangents[i].x, this->knots[i].y + this->tangents[i].y);
     }
     cairo_stroke(cr);
 
-    // draw spline
-    view.drawStroke(cr, stroke, 0);
+    // create stroke and draw spline
+    this->updateStroke();
+    if (this->getKnotCount() > 1) {
+        view.drawStroke(cr, stroke, 0);
+    }
 }
+
+constexpr double SHIFT_AMOUNT = 1.0;
+constexpr double ROTATE_AMOUNT = 5.0;
+constexpr double SCALE_AMOUNT = 1.05;
 
 auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
     if (!stroke ||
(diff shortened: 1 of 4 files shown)


src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <cmath>
+
+SplineSegment::SplineSegment(const Point& p, const Point& q) {
+    firstKnot = p;
+    firstControlPoint = p;
+    secondKnot = q;
+    secondControlPoint = q;
+}
+
+SplineSegment::SplineSegment(const Point& p, const Point& fp, const Point& sp, const Point& q) {
```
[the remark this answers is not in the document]
rolandlo: This suggested change gives me a segfault error upon execution. The error comes from invoking `SplineSegment firstPart = SplineSegment(firstKnot, b0, c0, d0);` in the `SplineSegment::subdivide`-method.

rolandlo:
Thanks a lot for the feedback. I agree with everything except with the use of the member initialization list syntax in `SplineSegment::SplineSegment(const Point& p, const Point& fp, const Point& sp, const Point& q)`, see my comment above. Should I leave this constructor as is or change the code in another way?

src/model/SplineSegment.cpp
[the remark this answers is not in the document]
Technius: The member initialization list should be equivalent to the assignments. Did you copy the above change verbatim? This constructor assigns `fp` and `sp` as well.

If the member initialization list syntax doesn't work, it's fine to just forget it.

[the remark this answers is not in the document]
rolandlo: The constructor that does not work (although it compiles fine) is
```
SplineSegment::SplineSegment(const` Point& p, const Point& fp, const Point& sp, const Point& q): 
        firstKnot(p), firstControlPoint(fp), secondKnot(sp), secondControlPoint(q) {}

```
It does assign `fp` and `sp` as well. So I will leave it as it is.

rolandlo:
The follow-up commit contains some code-polishing (like using `->` consistently in function headings) as well. Are any more changes required?

Technius:
Just the one last code simplification above and then it's good to go!

src/model/SplineSegment.cpp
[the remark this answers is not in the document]
Febbe: I also tried it, and for me it works. There might be another problem. Do you tried to recompile the complete project?
```
SplineSegment::SplineSegment(const Point& p, const Point& fp, const Point& sp, const Point& q):
        firstKnot(p), secondKnot(q), firstControlPoint(fp), secondControlPoint(sp) {}
```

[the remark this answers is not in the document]
rolandlo: Yes, I did. The program still crashes once I set the second knot (when the first tangent is non-trivial).

src/control/tools/SplineHandler.cpp
```diff
@@ -77,123 +94,113 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
...
             break;
         }
         case GDK_KEY_Left: {
-            movePoint(-1, 0);
+            this->movePoint(-1, 0);
+            break;
+        }
+        case GDK_KEY_r:
+        case GDK_KEY_R: {  // r like "rotate"
+            double angle = (event->state == GDK_SHIFT_MASK) ? -5 : 5;
```
Technius: The event state is a bit field, so you need to use
```suggestion
            double angle = (event->state & GDK_SHIFT_MASK) ? -5 : 5;
```

src/control/tools/SplineHandler.cpp
```diff
@@ -77,123 +94,113 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
...
+            double xOld = tangents.back().x;
+            double yOld = tangents.back().y;
+            double xNew = cos(angle * M_PI / 180) * xOld + sin(angle * M_PI / 180) * yOld;
+            double yNew = -sin(angle * M_PI / 180) * xOld + cos(angle * M_PI / 180) * yOld;
+            this->modifyLastTangent(Point(xNew, yNew));
+            break;
+        }
+        case GDK_KEY_s:
+        case GDK_KEY_S: {  // s like "scale"
+            double factor = (event->state == GDK_SHIFT_MASK) ? 1 / 1.1 : 1.1;
```
Technius: Same here.
```suggestion
            double factor = (event->state & GDK_SHIFT_MASK) ? 1 / 1.1 : 1.1;
```

src/model/SplineSegment.cpp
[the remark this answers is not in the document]
Febbe: The thing is, Technius and my code sniped must not result in a segfault. Your version and our's also produce exactly the same binary output (with optimizations enabled). The only case it can crash there is, when one of the Point references are deleted. Since 2 of them are members of the SplineSegment, the SplineSegment may be deleted. And that would be the real Problem. I'll try to recompile it with sanitizers on and look for the real cause.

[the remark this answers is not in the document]
Febbe: Ok I found the problem: It's a combination of using recursion, many growing vectors in a recursive function and the move into the return vector which may eliminate return value optimization. I'll list my change wish the following review comments.

src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+        std::pair<SplineSegment, SplineSegment>&& twoSegments = subdivide(0.5);
+        pointArray = std::move(twoSegments.first.toVectorOfPoints());
+        pointArray.pop_back();
+        std::vector<Point>&& secondArray = twoSegments.second.toVectorOfPoints();
+        pointArray.insert(pointArray.end(), std::make_move_iterator(secondArray.begin()),
+                          std::make_move_iterator(secondArray.end()));
+    }
+    return pointArray;
+}
```
Febbe: Generally use early return as it increases readability, also it might be more efficient here.
Use "auto const&" for long constant return values or introduce a type alias and use it as "T const&".
Try to avoid to move into return value container, as it breaks possible return value optimisation. Instead assign it to a new local variable. 
The pop_back can be optimized out if the whole function only returns n-1 points.
The move_iterator is not necessary since Points are not movable.
Also vectors have a very bad scaling when resized often better use std::list or std::deque. 

Here an example:

```suggestion
auto SplineSegment::toPointSequence() const -> std::list<Point> {
    auto result = toPointSequenceImpl();
    result.emplace_back(secondKnot);
    return result;
}

auto SplineSegment::toPointSequenceImpl() const -> std::list<Point> {
    if (isFlatEnough()) {
        return {firstKnot};
    }
    auto const& childSegments = subdivide(0.5);
    auto left_points = childSegments.first.toPointSequenceImpl();
    left_points.splice(end(left_points), childSegments.second.toPointSequenceImpl());
    return left_points;
}
```

src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+auto SplineSegment::subdivide(float t) -> std::pair<SplineSegment, SplineSegment> {
```
Febbe: This function can be const:
```suggestion
auto SplineSegment::subdivide(float t) const -> std::pair<SplineSegment, SplineSegment> {
```

src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,76 @@
...
+    SplineSegment firstPart = SplineSegment(firstKnot, b0, c0, d0);    // first point of each step
+    SplineSegment secondPart = SplineSegment(d0, c1, b2, secondKnot);  // last point of each step
+    return std::make_pair(firstPart, secondPart);
+}
+
+auto SplineSegment::linearInterpolate(const Point& p, const Point& q, float t) -> Point {
+    return Point(p.x * (1 - t) + q.x * t, p.y * (1 - t) + q.y * t);
+}
+
+auto SplineSegment::evaluate(float t) -> Point {
```
Febbe: This function is unused and can be deleted.
Also it can be const:
```suggestion
auto SplineSegment::evaluate(float t) const -> Point {
```

src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,76 @@
...
+    Point b2 = SplineSegment::linearInterpolate(secondControlPoint, secondKnot, t);
+
+    Point c0 = SplineSegment::linearInterpolate(b0, b1, t);
+    Point c1 = SplineSegment::linearInterpolate(b1, b2, t);
+
+    Point d0 = SplineSegment::linearInterpolate(c0, c1, t);  // This is the interpolated point
+    return d0;
+}
+
+auto SplineSegment::isFlatEnough() -> bool {
```
Febbe:
```suggestion
auto SplineSegment::isFlatEnough() const -> bool {
```

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+ * Every spline segment is defined by two knot point and two control points.
+ * This contrasts to the cairo-draw command cairo_curve_to, where everything is relative to the current position of
+ * another knot,
+ */
+
+
+class SplineSegment {
+public:
+    SplineSegment() = default;
+    ~SplineSegment() = default;
```
Febbe: It doesn't really matter, but I prefer to remove the "custom defaulted destructor" as it makes no difference for the compiler, but many people not using our .clang-tidy file will get into a warning hell, because the "rule of 5" is not applied.

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+     * @param sp the second control point
+     * @param q the second knot
+     */
+    SplineSegment(const Point& p, const Point& fp, const Point& sp, const Point& q);
+
+    /**
+     * @brief draw the spline segment via the cairo_curve_to command
+     * @param cr the cairo context on which the spline segment ought to be drawn
+     */
+    void draw(cairo_t* cr);
```
Febbe:
```suggestion
    void draw(cairo_t* cr) const;
```

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+     * @return A point list which represents the spline segment without the start point.
+     */
+    std::vector<Point> toVectorOfPoints();
+
+    /**
+     * @brief Subdivide the spline into two parts with respect to parameter t.
+     * @param t the parameter between 0 and 1, which corresponds to the point where the spline is split
+     * @return A pair of two spline segements corresponding to the two parts of the spline
+     */
+    std::pair<SplineSegment, SplineSegment> subdivide(float t);
```
Febbe:
```suggestion
    std::pair<SplineSegment, SplineSegment> subdivide(float t) const;
```

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+     * @return A point on the line segment which corresponds to the parameter value t
+     */
+    static Point linearInterpolate(const Point& p, const Point& q, float t);
+
+    /**
+     * @brief evaluate the function defining the spline at value t
+     * @param t the parameter between 0 and 1, which corresponds to the point on the spline segment
+     * @return A point on the spline segment which corresponds to the parameter value t
+     */
+    Point evaluate(float t);
```
Febbe:
```suggestion
    Point evaluate(float t) const;
```

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * @brief checks if the spline segment is flat enough so that it can be drawn as a straight line
+     * @return true, if the spline segment is flat enough; false otherwise
+     */
+    bool isFlatEnough();
```
Febbe:
```suggestion
    bool isFlatEnough() const;
```

src/model/SplineSegment.h
```diff
@@ -0,0 +1,114 @@
...
+     * @brief draw the spline segment via the cairo_curve_to command
+     * @param cr the cairo context on which the spline segment ought to be drawn
+     */
+    void draw(cairo_t* cr);
+
+    /**
+     * @brief Convert the spline segment to a vector of points.
+     * @return A point list which represents the spline segment without the start point.
+     */
+    std::vector<Point> toVectorOfPoints();
```
Febbe: Here the declaration of my suggestion from above:
Also I did not understand, why the first point is missing.

```suggestion
    /**
     * @brief Convert the spline segment to a vector of points.
     * @return A point list which represents the spline segment.
     */
    std::list<Point> toPointSequence() const;

    /**
     * @brief Convert the spline segment to a vector of points.
     * @return A point list which represents the spline segment without the end point.
     */
    auto toPointSequenceImpl() const -> std::list<Point>;
```

rolandlo: That "without the start point" comment was a mistake (I planned it differently earlier).

rolandlo:
Thanks, Febbe, for all your suggestions and explanations. I am sorry to say, however, that even with all those improvements the member initialization list gives me a segfault error upon execution. That's why I left the constructor with assignments.

src/control/tools/SplineHandler.cpp
```diff
@@ -202,51 +215,111 @@ void SplineHandler::finalizeSpline() {
...
-    if (pointCount > 1) {
-        // remove dynamically changing point at (previous) cursor position
-        stroke->deletePoint(pointCount - 1);
+    // convert collection of segments to stroke
+    stroke->deletePointsFrom(0);
+    stroke->addPoint(knots.front());
+    size_t indexLast = 0;
+    for (auto s: segments) {
+        for (auto p: s.toPointSequence()) {
+            if (validMotion(p, stroke->getPoint(indexLast))) {
```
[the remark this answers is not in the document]
rolandlo: I now added such a check in the flatness criterion, so distances between two subsequent points won't get too small.

src/control/tools/SplineHandler.cpp
```diff
@@ -202,51 +215,111 @@ void SplineHandler::finalizeSpline() {
...
+        segments.push_back(SplineSegment(knots[i], cp1, cp2, knots[i + 1]));
+    }
 
-    int pointCount = stroke->getPointCount();
-    if (pointCount > 1) {
-        // remove dynamically changing point at (previous) cursor position
-        stroke->deletePoint(pointCount - 1);
+    // convert collection of segments to stroke
+    stroke->deletePointsFrom(0);
+    stroke->addPoint(knots.front());
```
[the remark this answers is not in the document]
rolandlo: It is no longer needed now that the check on the distance to the previous point has been removed.

src/model/SplineSegment.cpp
```diff
@@ -0,0 +1,64 @@
...
+    SplineSegment secondPart = SplineSegment(d0, c1, b2, secondKnot);  // last point of each step
+    return std::make_pair(firstPart, secondPart);
+}
+
+auto SplineSegment::linearInterpolate(const Point& p, const Point& q, float t) -> Point {
+    return Point(p.x * (1 - t) + q.x * t, p.y * (1 - t) + q.y * t);
+}
+
+auto SplineSegment::isFlatEnough() const -> bool {
+    double eps = 1.001;
```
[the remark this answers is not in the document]
rolandlo: In practice an "average" spline segments will now be converted to a list of roughly 70 points. I hope this is fine.

rolandlo:
One last thing: I changed the key binding for moving the last drawn knot around. Instead of the arrow keys (which are used for page navigation) I use the x- and y-keys (in combination with SHIFT for reversing its effect) to move the knot in x- resp. y-direction. It was really troublesome to move knots up and  down in documents with more than one page with the arrow keys.
Is everything alright now?

src/model/SplineSegment.cpp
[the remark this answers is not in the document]
Febbe: Yes, that's fine, the best would be to make this a preference, but until this is not the case, we should make a good compromise between usability, pleasant appearance, and performance / energy consumption. Otherwise we get either tickets for "high energy consumption" or a bad "look and feel".

src/control/tools/SplineHandler.cpp
```diff
@@ -202,51 +215,111 @@ void SplineHandler::finalizeSpline() {
...
+    size_t indexLast = 0;
+    for (auto s: segments) {
+        for (auto p: s.toPointSequence()) {
+            if (validMotion(p, stroke->getPoint(indexLast))) {
+                stroke->addPoint(p);
+                indexLast++;
+            }
```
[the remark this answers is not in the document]
Febbe: Done

Febbe:
Mostly everything is fine, that with the keys is indeed a bug in your code: Your Key Handler does not  return true if the key has been handled, this should be changed for all Actions. For the change to x and y I would consume @Technius. I prefer still "Up" and "Down".

There are some stylistic things like the use of `container.size()>0` which should be `!container.empty()`. Or the use of magic numbers (constants). Which should be constexpr variables with a descriptive name. But that's nitpicking on high niveau.
I don't expect you to change this.

Do you know how to [squash](https://www.internalpointers.com/post/squash-commits-into-one-git) your commit into one? You have to force push the squashed commit(s) to your branch.
For fixups or clang format I prefer to just use fixup(f) instead of squash(s).  
That would be the last step before merge:
1. Clean your history up (squash or fixup commits)
2. Rebase onto current master.

Technius:
> I use the x- and y-keys (in combination with SHIFT for reversing its effect) to move the knot in x- resp. y-direction. It was really troublesome to move knots up and down in documents with more than one page with the arrow keys.

Using <kbd>x</kbd> and <kbd>y</kbd> is really bad UX, but I know what the problem is. Review comment following soon...

src/control/tools/SplineHandler.cpp
```diff
@@ -77,123 +94,115 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
...
+            double xNew = xOld * factor;
+            double yNew = yOld * factor;
+            this->modifyLastTangent(Point(xNew, yNew));
             break;
         }
     }
-    this->redrawable->rerenderRect(x - 1, y - 1, w + 2, h + 2);
+    this->redrawable->repaintRect(rect.x - 1, rect.y - 1, rect.width + 2, rect.height + 2);
     return false;
```
Technius: This always returns false, which means that the event will continue to be propagated even when it has been handled here. You will need to return true if you've done something with the keypress.

rolandlo: I switched back to arrow keys and corrected the return statements.

rolandlo:
First time I squashed my commits into one, rebased on current master and force-pushed the squashed commit to my branch. But it seems it worked, didn't it?

rolandlo:
@Technius Anything left to be done, until it can be merged?

Febbe:
Sorry I think neither @technius nor me had time to review a last time. I currently have to work for example😅. But it should be perfectly mergeable. I also think, that additional changes should be done in a fixup PR when there are more. Because your work seems to be very clean, good and ready as is.

Technius:
LGTM, merging now. Thanks for the PR!

PR #1800, opened 2020-03-25
Replaced PageRef with STL smart pointer

Febbe:
The PageRef was originally inconsistently used, sometimes with manual referencing and sometimes over the self made smart pointer. Also PageRef did not implemented a move constructor, thus a efficient noexept move of a container was not possible. This enables further optimization possibilities.
Which are not applied yet.


Technius:
This won't interfere with the viewport refactor...I think.

Let me know when you want a review.

Technius:
CI build failed, please fix that first.

Febbe:
@Technius fixed it.

Issue #1705, opened 2020-01-15
Redesign large size Xournal++ icon


luyatshimbalanga:
**Is your feature request related to a problem? Please describe.**
The current icons used on Xournal++ is set at 48x48 size causing it to not properly display on Gnome Shell.

**Describe the solution you'd like**
Modernize the  Xournal++ icon set at least to 128x128 in SVG format

**Describe alternatives you've considered**
Simply resize the current application icon to 128x128.

**Additional context**
A new Xournal++ logo in addition of refreshed tool icons will improve the appearance of the application while complying to [Gnome Icon and Artwork guideline.](https://developer.gnome.org/hig/stable/icons-and-artwork.html.en)

luyatshimbalanga:
![Screenshot from 2020-01-19 00-32-36](https://user-images.githubusercontent.com/7230131/72677613-3a1e9200-3a53-11ea-95d0-eca397b2dbc3.png)

A preview of the new redesigned Xournal++ icon retaining some elements from the old logo while adhering to Gnome Icon guideline.
Additionally, a icon from the development version is also available.

Febbe:
LGTM

Fatih20:
Cool.

Fatih20:
@luyatshimbalanga So is it going anywhere? Have you made some pull request or what, because I'm looking into doing this as well.

luyatshimbalanga:
![com github xournalpp xournalpp-screenshot](https://user-images.githubusercontent.com/7230131/82125595-d5064f00-975b-11ea-8edb-217a2a8a2ef3.png)
Updated screenshot

PR #1808, opened 2020-03-29
New icon app

luyatshimbalanga:
Xournal++ icon needed a redesign to fit in a modern desktop environment. Rational is available at referred issue #1705 . Both PNG and SVG version are  128x128 size.


Technius:
Any reason why the new icons use a different path?

Fatih20:
In my opinion, the pen needs another color. It _kinda_ blend in with the white paper and looked monotone.

Fatih20:
Also, are you going to redesign the app's files icon as well? Because it also looked old.

luyatshimbalanga:
Sorry for the delay, I was very busy due to the pandemic event.

> Also, are you going to redesign the app's files icon as well? Because it also looked old.
I will.

> Any reason why the new icons use a different path?
They follow the guideline from Freedesktop and hicolor i.e. /usr/share/icons/hicolor/{128x128,scalable}/app for the new icon and /usr/share/icons/symbolic/app for it symbolic version. See https://developer.gnome.org/hig/stable/icons-and-artwork.html.en

Fatih20:
Any further update? I could help you with this.

luyatshimbalanga:
New icons are updated.
@Technius 
The reason for the different path is visibility of the "X".

LittleHuba:
Thanks for the work so far @luyatshimbalanga!
I'm unfortunately not really happy with the look and feel of the icon yet. I'd prefer a "flat" look of the pencil and slightly thicker strokes for the X. Also a less aggressive color for the X would be preferable to the current orange.
With those changes we should have an icon that looks modern and fresh and should last for some time.

Fatih20:
> Thanks for the work so far @luyatshimbalanga!
> I'm unfortunately not really happy with the look and feel of the icon yet. I'd prefer a "flat" look of the pencil and slightly thicker strokes for the X. Also a less aggressive color for the X would be preferable to the current orange.
> With those changes we should have an icon that looks modern and fresh and should last for some time.

If @luyatshimbalanga were to flatten the pen, IMHO, he should change the pen color to something else because it could be kinda lost to the paper, preferably orange to match the cross on the paper.

luyatshimbalanga:
Thank you for the feedback. The pen inspired from the BIC brand is used for illustration and the X has a darker red thicker stroke. Check the latest commit on this [forked branch](https://github.com/fedoradesign/xournalpp)

LittleHuba:
This already looks way better!
Two more things:
1. You missed to update the `.png` files
2. The `X` seems somewhat unfinished currently. I'd add some dynamic there by not using straight lines. Another idea would be to actually use the Pen as one of the lines making up the X and some other tool as the second stroke. No idea how that actually turns out though.

Fatih20:
>     2. The `X` seems somewhat unfinished currently. I'd add some dynamic there by not using straight lines. Another idea would be to actually use the Pen as one of the lines making up the X and some other tool as the second stroke. No idea how that actually turns out though.

Regarding this, an x like this would be interesting.
https://qph.fs.quoracdn.net/main-qimg-229af91fa609577c27c69f9e240cfd62
Also, about dynamics, the Live Path Effects in Inkscape named Power Stroke and Taper Stroke could help achieving this.

luyatshimbalanga:
@Fatih20 and @LittleHuba ,
Sorry for the delayed update. I finally have the opportunity to update the icon based on Fatih20's suggestion. Let know what you think. 
Preview

Fatih20:
> @Fatih20 and @LittleHuba ,
> Sorry for the delayed update. I finally have the opportunity to update the icon based on Fatih20's suggestion. Let know what you think.

IMHO, the color of the x is too dull. If you look at GNOME's app icons, they always looked vibrant but doesn't stab your eye with oversaturation (like with the x in current xournal icons). There's a balance that needs to be met there. The GNOME icon palette would be the first thing you should look into since you're making the icons in their style.

Alternatively, you could just make the x black and make the pen colored instead. I personally lean more into this, since you'd probably write stuff in black when using xournal since it's not a drawing app.

luyatshimbalanga:
> > @Fatih20 and @LittleHuba ,
> > Sorry for the delayed update. I finally have the opportunity to update the icon based on Fatih20's suggestion. Let know what you think.
> 
> IMHO, the color of the x is too dull. If you look at GNOME's app icons, they always looked vibrant but doesn't stab your eye with oversaturation (like with the x in current xournal icons). There's a balance that needs to be met there. The GNOME icon palette would be the first thing you should look into since you're making the icons in their style.
> 
The icon uses GNOME HIG icon palette.

> Alternatively, you could just make the x black and make the pen colored instead. I personally lean more into this, since you'd probably write stuff in black when using xournal since it's not a drawing app.

I agree about setting the "x" black. Here is the possible choice of colour for the pen: 
Preview-v2
Inspiration from the current logo.

Preview-v3
Slight variant of the current concept.

Preview-v4
Inspiration from the original Xournal.

Fatih20:
@luyatshimbalanga I'd take the first one because that's the color of the typical pencil.

luyatshimbalanga:
@Fatih20 and @LittleHuba ,
Icon updated. Both png (256x256) and svg are included.

rolandlo:
Very nice work! So clean. I would also pick the first choice for the color for the pen.

rolandlo:
What is the status of this PR? Any work left to do?

luyatshimbalanga:
Awaiting @LittleHuba and @Technius for feedback.

Technius:
I agree with the others in that the orange color pencil is the best! I think this is ready for merge. Just one minor criticism: the "wood" color on the pencil blends in slightly with the paper, making it hard to see on the smaller sizes. This shouldn't be too serious though, since the "x" and the pencil are distinguishing features on the small icons.

Technius:
Merging in 3 days unless there's an objection.

LittleHuba:
This PR had major issues. Windows builds ended up with no icon at all. I'll revert the commit for this reason and it will have to be reworked before we merge it once more.

LittleHuba:
@luyatshimbalanga please recheck your work and provide a new PR.

PR #1813, opened 2020-03-30
[WIP] View rewrite

### LINKED ISSUE #1795: View refactor metaissue
This issue tracks progress on the viewport refactoring and facilitates discussion on how it should be done.

**Why?**
* The current scrolling and zoom code is a hot mess of spaghetti code. There are many bugs that cannot be fixed without rewriting the entire zoom/scroll system (see below).
* One of the goals of the new file format is to support infinite pages and lazy loading, which will not be possible under the current system.

**Potentially fixable issues**
* Poor scroll/zoom interaction: #1501, #1608, #1660, #1668, #1699, #1753, #1777
* Page selection behavior: #1658
* Presentation mode: #1035, #1764
* Render issues: #1762

**How**
See #1433 for some details.

Some other points to consider:
* The scroll and zoom code has a potential infinite loop, causing #1660. In short, scrolling triggers page selection, which then triggers zoom code, which then triggers scrolling...
* To avoid the zoom/scroll part of the loop, `ScrollHandler` and `ZoomControl` should be combined into a single `Viewport` class, which will perform *both* scrolling and zooming simultaneously (e.g. a standard affine transform).
* To avoid selection/viewport looping, we must distinguish between *viewport update* and *navigation*. The former *only* changes what will be rendered, while the latter will also force a page selection. A viewport update should *not* cause a page selection.
* Zooming and scrolling should interpolate between pages of different sizes, otherwise we will run into jumpy scrolling behavior.

lehmanju:
partial credit goes to @Technius 

Refactor view, Closes #1795 
Removes old input system, scroll handling and touch workaround (can be achieved by setting corresponding device to pen)


lehmanju:
Uml diagrams can be found here, created with umbrello: https://github.com/lehmanju/xournalpp/tree/view-rewrite/development/uml

lehmanju:
So I have been thinking a lot about a better architecture and came across the [Flux Pattern](https://facebook.github.io/flux/docs/in-depth-overview/). Separating state and view updates allows for a cleaner application and data flow. More importantly, it allows swapping out different components and even adding secondary XournalWidgets as they do not carry any state whatsoever. In this rewrite I will however only focus on the main widget part, split its state and behavior and write an adapter that feeds actions into a newly created dispatcher. For the future I was thinking about a two stage action dispatching system. The first stage is basically the input system, its actions are GtkEvents. Those will then be converted into actions based on the first stages state and then fed into the main dispatcher that will update the application model and view.

LittleHuba:
Before you put a lot of work in there give me a chance to provide you with the corresponding API of the new library. Otherwise you might need to redo parts which would be unnecessary overhead. I'll send you the stuff by tomorrow at the latest.

Technius:
`TouchDrawingInputHandler` shouldn't be deleted since there are still people interested in using the touch screen to draw.

lehmanju:
@Technius it works differently with the new view, so there is no need for it anymore.

lehmanju:
I will remove most of the old stuff, get it working and then readd if necessary.

lehmanju:
main

State is only captured in various storages and the InputHandlers. Everything else will be stateless.

lehmanju:
Alright, so I added a bunch of adapter methods to `XournalView` which will either be noops or create an action for the relevant storages. This results in minimal change to other files to get this working. We can then expand from there on to use the pattern more widely.

lehmanju:
Closed in favor of #1925

PR #1817, opened 2020-03-31
Make Rectangle a template class

lehmanju:
Better type safety this way.

### DIFF
--- a/src/util/Rectangle.cpp
+++ b/src/util/Rectangle.cpp
@@ -4,25 +4,24 @@
 
 #include "Range.h"
 
-Rectangle::Rectangle() = default;
+template <class T>
+Rectangle<T>::Rectangle() = default;
 
-Rectangle::Rectangle(double x, double y, double width, double height): x(x), y(y), width(width), height(height) {}
+template <class T>
+Rectangle<T>::Rectangle(T x, T y, T width, T height): x(x), y(y), width(width), height(height) {}
 
-Rectangle::Rectangle(const Range& rect):
+template <class T>
+Rectangle<T>::Rectangle(const Range& rect):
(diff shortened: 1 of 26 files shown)


src/util/Rectangle.cpp
```diff
@@ -4,18 +4,20 @@
...
 
-Rectangle::~Rectangle() = default;
-
-auto Rectangle::intersects(const Rectangle& other, Rectangle* dest) const -> bool {
-    double destX = NAN, destY = NAN;
-    double destW = NAN, destH = NAN;
+template <class T>
+auto Rectangle<T>::intersects(const Rectangle<T>& other, Rectangle<T>* dest) const -> bool {
+    T destX = NAN, destY = NAN;
+    T destW = NAN, destH = NAN;
```
Technius, line 20: These declarations should be removed and the `destX = std::max(...)` lines should become declarations. `NAN` is not a valid value for any non-floating point number.

src/util/Rectangle.h
```diff
@@ -1,7 +1,7 @@
 /*
  * Xournal++
  *
- * A rectangle with double precision
+ * A Rectangle<double> with double precision
```
Technius, line 4: The comment here should be about the `Rectangle` class--the new comment is written as if it is a specific instantiation of `Rectangle`.

src/util/Rectangle.h
```diff
@@ -1,7 +1,7 @@
 /*
  * Xournal++
  *
- * A rectangle with double precision
+ * A Rectangle with variable precision
```
Technius, line 4:
```suggestion
 * A rectangle data structure over a numeric data type.
```

Issue #1818, opened 2020-03-31
"Hitbox" of arrow heads not accurate.


Septeract7:
**Affects versions :**
 - OS: Windows 10
 - libgtk v3.0
 - Version of Xournal++: 1.1.0+dev, cb964dc4
 - Installation method: windows installer (official release, nightly)

**Describe the bug**
The "hitbox" of the arrow head seems to stretch too far down the arrow shaft. When erasing (with the stroke eraser) *only near* the arrow head, without the eraser touching the lines of the arrow, the arrow gets erased anyways.

**To Reproduce**
1. Use Arrow drawing tool to draw an arrow.
2. Select *stroke* eraser.
3. Erase near the arrow head without touching the arrow with the eraser.
4. The whole arrow will be deleted.

**Expected behavior**
The stroke eraser should only erase the arrow if it touches any line of the arrow. 

**Additional context**
I am using a Wacom Intuos Tablet. But i don't think this is input related.

Technius:
An arrow is actually treated as a single stroke, which is why the whole arrow is deleted. Here are the vertices of the stroke:
![image](https://user-images.githubusercontent.com/1066652/78073072-0f0ac580-7355-11ea-84ae-f3bd0f62bb7d.png)

Is there a particular use case you have in mind for deleting only part of the arrow? In my experience, it's more useful to have the entire arrow deleted.

rolandlo:
@Technius I think you misunderstand the problem raised. The problem is not that the whole arrow is deleted. The problem is that the arrow is deleted even if the eraser Rectangle does not hit any of the line segments forming the arrow.
I can reproduce this behaviour. It is not restricted to arrows, but it happens with hand drawn strokes as well. Looking at the code this behaviour seems to result from the intersection-check
`auto Stroke::intersects(double x, double y, double halfEraserSize, double* gap) -> bool`
from stroke.cpp
In my opinion the code therein does not what the creator intended. Both 

1.  the computation of the distance p of the center (x,y) of the Eraser Box from the straight line passing through the two subsequent points of the stroke
2. the computation of the center (midpoint) between the two subsequent points of the stroke

are flawed. 
In 1. the divisor is the length of the wrong vector and 
in 2. the center (x,y) of the eraser box should be replaced by the second point (px, py) on the stroke

A quick fix would be to replace line 300 of strokes.cpp
[...]

PR #1819, opened 2020-04-01
fix eraser bug

rolandlo:
This should fix the bug from issue #1818. 
I can't say what this padding is supposed to be for (it seems to work fine without), that the original author coded (more than 9 years ago). Should I simply remove it?


Technius:
LGTM, thanks for the PR.

Issue #1778, opened 2020-03-18
Laser


alexisflesch:
**Is your feature request related to a problem? Please describe.**
I'm using xournalpp as a virtual board to stream with my students. I know it is not its first purpose but it does the job quite well with a wacom tablet. I would love it if it had an option like google jamboard to show a laser (with a latency of approximately 1s).

**Describe the solution you'd like**
A button in the toolbar to start the laser.

Technius:
This should be fairly easy to impement.

With the current events and schools moving to online classes, I'm going to make this a high priority.

alexisflesch:
This is very nice of you, thanks a lot !

Technius:
I started working on this today and it should be done within the next few days.
Check out my PR for progress.

Technius:
@alexisflesch There's been some discussion in #1822 about the scope of this feature and I think it would be great if you could provide some feedback.

StHante:
I am using a tool called PenAttention to do this. I actually have the "laser" on even when I'm writing.

Izerpizer:
What is the current status of the laser cursor? I am eagerly awaiting this feature!

frohro:
I'm also interested in this feature.  In the mean time I am using the lasso select to highlight things in online lectures.

JPYamamoto:
Are there any updates regarding this feature? I would also greatly benefit from it.

ronandalton:
I've taken some inspiration from #1822 but so far it seems to be a pretty big pull request in comparison to that. I currently have changes across 23 files, since adding a new tool seems to require a lot of code changes in various places. I'm just trying to get an initial version working and then I'll make a new pull request.

biosaat:
As the implementation seems to have such a big impact: 

What do you think about an alternative approach which is perhaps easier to implement:
One could have a new button/functionality to normally highlight with the red bold textmarker---but the entire highlighted line is automatically removed one second after it was finished. 

This is not exactly a laser pointer (where the tail starts disappearing as you draw) but perhaps something to start with.

renatocan:
I think this alternative would work. When I was looking for this feature, I was imagining something like the tool used in this point of this video: https://www.youtube.com/watch?v=OrBEon3VlQg&t=1455s (BTW, does someone know what application is this one, used in the video?)

As I understand, this alternative that was proposed is similar, except for the fade effect that is used to remove the highlighted line.

Thanks!

Rmano:
[...]

Issue #919, opened 2019-02-22
Configure Shortcuts


Raffinade:
For me it would be fine to configure shortcuts in the frontend even for the colors, so that i dont have to put the pointer on the color to change it.
I hope it will be possible.

Greets Raffinade

LittleHuba:
We should add a section to the settings for this and make all shortcuts we have adjustable.

LittleHuba:
This will require a global system listening for shortcuts. They can not be bound to the buttons themselves anymore (which might be better anyway).

relunctance:
I'm looking forward to this shortcuts/keybindings . When can I used it

Narice:
Yes, that would be a must!

ave-63:
The only reason I am still using MS windows is because it has PDF Annotator, which has keyboard shortcut customization for almost every action. I use lines, four colors of pens, four colors of highlighters, lasso select, copy, paste, undo, delete, and a few other things this way. I have them bound to regular keys (asdfg, ASDF, qwert, etc). 

LittleHuba says they cannot be bound to regular keys in xournalpp, and that is a shame! I'm a math teacher that uses PDF Annotator for regular lectures, and I will get sore fingers from holding down shift+ctrl all day.

Anyway, I was learning how to use git and cmake so that I could implement this myself, but it looks like others are already working on it? It will take me a ton of time to get up to speed because I'm not a software developer.

asanakoy:
My vote for customizable shortctus as well. 
Or at least adding shortcuts for changing Pen line style would be helpful

tobiasBora:
I would also love custom shortcuts, because I'm using several tools (xournalpp, drawpile...) and I'd love to have consistent, short, shorcuts.

But at least, as a first simple thing, would it be possible to bind "Ctrl-Shift-E" to "Export as PDF"? (the shorcut would even be consistent with Gimp shortcuts) I use that feature always and I'm tired of clicking with the mouse.

Thanks a lot!

kamdh:
PgUp and PgDn should automatically go up or down a page. Ctrl-PgUp/PgDn is cumbersome

JaviMaciasG:
I would also like to have all drawing elements to be accessible by keyboard. Most tools are already this way, but I heavily miss this feature for colors and pen options. I am using xournal++ for synchronous online and offline teaching (in a public University) and constant mouse movements to access some drawing features are distracting to students who are used to following my cursor around.

Thank you very much for your superb application!!

[...]

PR #1822, opened 2020-04-02
[WIP] Presentation laser

Technius:
Implements #1778. This adds a new "presentation laser" that follows after the mouse cursor.

As an exception due to the current circumstances, this new feature will be backported to the `1.0.x` series.

TODO:
* [ ] Toolbar button to toggle the presentation laser.
* [ ] Configuration options for the laser, e.g. color, size, speed, etc.
* [ ] Better graphical effects for the laser, such as a laser "trail" (see Google Presentation pointer)


Technius:
Here's a preview of the laser:
xpp_laser_preview

Rmano:
This is nice...!
But please let me share my idea about the laser which is quite different from what I see this. I think of the laser as a tool to temporarily highlight part of the drawing, so in my opinion it should be a new tool, so when you choose it (for example instead of the pen) you have the following behavior:

- if I hover the tablet pen / move the mouse I have just the red  dot where the pointer is
- if I "draw" the tablet pen / move the mouse clicking the dot leave a 1-2 seconds fading trail, so that if I draw a circle around a formula I'll have a nice fading circle around it.

I do not find the red dot following the pen while writing very useful; the idea of a laser is to using the pointing laser while presenting, and this is in a different phase than writing. 

Just my 2 cents and I understand that different people can have different opinions to this thing, and that your current implementation would be really useful already... So thanks!

rolandlo:
The google jamboard laser pointer mentioned in issue #1778 looks indeed quite different (like explained by @Rmano). See this [youtube video](https://www.youtube.com/watch?v=5p5yR8lGTdY). The demonstration of the laser pointer starts at 6:00.

Technius:
@Rmano @rolandlo 
Thanks for the valuable feedback--I put up this PR early specifically because I didn't have a clear idea of what people want. I suppose my preview GIF was not a good example, since this feature will probably be used together with a stylus.

> I think of the laser as a tool to temporarily highlight part of the drawing, so in my opinion it should be a new tool [...] if I "draw" the tablet pen / move the mouse clicking the dot leave a 1-2 seconds fading trail, so that if I draw a circle around a formula I'll have a nice fading circle around it.

This would be certainly more useful than what is currently here! However, I think that we shouldn't just copy what Jamboard does but also ensure that the feature fits the purposes of Xournal++ well. At least, one question I have is, "to what extent does an evaporating stroke have an advantage over simply...annotating the document?"

> I do not find the red dot following the pen while writing very useful; the idea of a laser is to using the pointing laser while presenting, and this is in a different phase than writing.

I agree with this; my example was misleading.

> The google jamboard laser pointer mentioned in issue #1778 looks indeed quite different (like explained by @Rmano). See this youtube video. The demonstration of the laser pointer starts at 6:00.

Thanks for the video--I couldn't easily find a good demonstration.

---

I think it may be useful to outline a few pros and cons of having the red dot so we can at least see what this basic prototype has the potential to bring to the table:

Pros:
* The standard stylus cursor is a single pixel, which is very difficult to see. The dot highlights the position of the cursor and is far more visible.
* There is a slight advantage of the dot over the large stylus cursor because the large cursor is monotone. Setting the dot to the appropriate color will increase the visibility of the dot (and can direct attention more effectively).

Cons:
* When a presenter is giving a presentation on a projected screen, a physical laser pointer is strictly superior to anything we can emulate with software, since the laser will be significantly brighter than whatever the projector can produce.
* I say "colored" because this dot will be indistinguishable from a background of the same color. A red dot is not visible on a red background (but a laser pointer on a projected screen _is_).
* When paired with a mouse cursor, the presentation laser is just visual clutter. A better alternative would be to just replace the mouse cursor entirely instead of tacking on a new feature.

Of course, everything is subject to improvement and more feedback would be appreciated.

rolandlo:
The purpose of a laser pointer/evaporating stroke (or similar) lies in turning the attention of the viewer temporarily to a certain region of the document. Annotating the document has not the same temporary effect and could overload the content of the document if used too much.
While teaching (in normal school environment) I use a stylus to write and the projector to show the content of my screen. My pupils find it very helpful when the position of the stylus on the screen is highlighted (with a yellow circle). Like that they can follow me even if I don't write and only want to highlight something temporarily. This is already integrated in xournal++. It must only be activated in the settings (under "view" -> "cursor").
So for my purposed I don't really need another tool for the same purpose. The laser pointer/evaporating stroke à la jamboard is more flashy (somehow cool). But maybe it is also more distracting than the highlighting of the stylus position via the yellow circle. I am not sure, it is any better. So we should carefully evaluate what would be the advantage of a laser pointer/evaporating stroke over what is already there. Of course it is nice to have options, but as a con, adding another tool will make the toolbar more cluttered. I never had trouble of visibility with the yellow circle, since I wouldn't write or draw something in yellow.

Rmano:
@rolandlo @Technius --- yes, the "show the cursor is nice", and cab fulfill 80% of the laser effect:
image
The only problem is that I find it distracting while writing (for the students), and it is quite difficult to switch on-off. 
Having it on a key for temporary showing would be great and basically all that's need for me. Being able to assign it to one of the pen's button (show the circle while I keep it pressed, or even toggle it on a click) would be perfect.

Technius:
> @rolandlo: The purpose of a laser pointer/evaporating stroke (or similar) lies in turning the attention of the viewer temporarily to a certain region of the document. Annotating the document has not the same temporary effect and could overload the content of the document if used too much.

This type of effect could be achieved by annotating the document and then using undo to remove the added annotations. I've done something similar with in-person meetings with whiteboards: emphasize something temporarily by circling/underlining some words written on the whiteboard, and then erasing the added lines afterwards.

Making these additional annotations permanent could also have a nice side effect: they would serve as notes to you (the presenter) about how you could improve your document in the future, or highlight ways in which your current document is lacking. You can avoid this "overloading of content" by drawing your presentation annotations on another layer--then it would be easy to hide your emphasis, making your document clearer.

---

> @Rmano: Having it on a key for temporary showing would be great and basically all that's needed for me. Being able to assign it to one of the pen's button (show the circle while I keep it pressed, or even toggle it on a click --- although the "on while pressed" would be better) would be perfect.

In general, I think we need both 1) more options to bind to buttons and 2) more buttons to bind options to. Hopefully we can get to #919 some time in the near future...

---

> @rolandlo: So for myself I don't really need another tool for the same purpose. [...] So we should carefully evaluate what would be the advantage of a laser pointer/evaporating stroke over what is already there. Of course it is nice to have options, but as a con, adding another tool will make the toolbar more cluttered.

I agree with this in general, but toolbar clutter isn't really an issue: the toolbar can be customized and we can make the laser pointer tool something that is not placed there by default. The real problem is that it adds more code to the codebase, which translates to a maintenance burden on us. Less is more, and we would rather not have the feature if it's subsumed by others.

alexisflesch:
Hi,

Sorry I didn't answer earlier, for some reason I didn't get notified of this conversation until today.

That being said, thanks a lot for the work. I do totally agree with everything that has been said here. I understand that more code means more maintenance. Here is what I think :
- a laser like jamboard (with a trail) makes it easy to show something without having to draw and remove (if you have a lot of stuff to show it will take time to ctrl+Z everything and make your presentation slow)
- if we could easily toggle the "show the cursor" behavior (with a mouseclick or stylus click), it could indeed replace this feature but would be a little less handy : with a trail I can make a virtual box around a result easily.
- the idea of a "jamboard laser" that would be available on a click would be perfect for my usage, but this might be very specific to me in which case it shouldn't be implemented.

michaelcadilhac:
How timely ;-)  Here's my use case: I'm also recording classes for students, on a ChromeOS (crostini).  As hovering is not passed to apps on ChromeOS, I have basically no way to point at things, least I plug a mouse.  So I took your great PR, and butchered in a button in the toolbar, implementing what I think is the "jamboard laser" above—that's what I need:
laser

Personally, I don't need the "delay" in the laser, so I will probably set the speed to a big number.

Looking forward to seeing that in the code base soon!

michaelcadilhac:
I've cleaned up a bit my thing, just out of personal sanity.  It's over there:

https://github.com/xournalpp/xournalpp/compare/master...michaelcadilhac:laser-tool?expand=1

I started using the floating toolbox, because of ChromeOS nonsensical hiding of the pen buttons, so I put tap detection in the handler too.  By the way, how come taping is not part of InputHandler, and is duplicated in BaseStrokeHandler and StrokeHandler? This redundancy is … strange.

H-M-H:
I like the solution from @michaelcadilhac,  I do not need any fancy animations, just a tool to quickly highlight something. I would vouch to keep it simple do it like that.
Also thanks for providing the changes you made @michaelcadilhac I will probably use your version for now.

michaelcadilhac:
@H-M-H Cheers.  I've also implemented double tap to bring the floating box and quickly select the tool; an example use case: https://youtu.be/ks-xknKgDTg?t=3557

Izerpizer:
What's the current status of this feature? I am eagerly awaiting its implementation.

Technius:
I've somewhat neglected this because I've been so busy, but I'm hoping to update this during the weekend. Not quite sure what the best solution should look like so far, but I'll implement one or both of the following based on how much time I have:

* A button to toggle cursor highlight, which seems to be sufficient for most people.
* A trail that follows the laser movements, to differentiate the laser from the cursor highlight. We could also make it a tool like @michaelcadilhac did.

Adding temporary/fading writing to the laser would involve some more work since we'd have to intercept input events.

Izerpizer:
The solution that I think is being looked for is the one that is mentioned [here](https://github.com/xournalpp/xournalpp/pull/1822#issuecomment-608111962), which is definitely the one that I am looking for as well; the ability to draw a line and have it fade away after some latency period, as mentioned [here](https://github.com/xournalpp/xournalpp/pull/1822#issuecomment-612358243). Solely highlighting the position of the cursor, while it may be useful to have, I'm not sure is the desire of the feature.

Rmano:
There is an intermediate possibility --- how about:

1. I click a button (probably a stylus button) and I activate laser-mode;  the cursor is highlighted.
2. At this moment, the pen is changed to the default highlighter (or configurable one) and an undo-point is stored.
3. I can draw around with the highlighter.
4. when I release (or click again, not sure about it) the button the undo rewinds removing the things drawn in laser-mode, and the highlight of the cursor is off.

That way there is no time-related code as needed for a fading trace, but the functionality of temporarily select things is here. If the click is stateful (I mean --- one click for selecting the mode, click again to exit) you can basically draw whatever you want and that will disappear when back to normal mode...

redweasel:
@Rmano there is a function called `onMotionCancelEvent()` in the code which can remove a stroke before adding it to the page. It would be trivial (I think) to turn any pen/highlighter into that kind of laser. I hope to get the laserpointer you first invisioned above (the comment with so many reactions), so I don't think an intermediate solution would be good.

Fun fact: Instead of undo you can also currently use touch drawing to get temporary strokes. You can draw with your finger and the stroke drawn will disappear as soon as you touch the screen with a second finger to start zooming.

Technius:
> @Technius If you can just make a pointer tool which shows a red dot as the mouse cursor when enabled, then I can try to pick it up and make the fading effect in collaboration with you in a new PR.

Yes, let's do that. I don't have much time to work on this PR due to Real Life Things™, and I'm trying to prioritize bugfixes for the 1.1.0 release.

Rmano:
> @Technius If you can just make a pointer **tool** which shows a red dot as the mouse cursor when enabled, then I can try to pick it up and make the fading effect in collaboration with you in a new PR.

Yes, I think that can be a quite reasonable stop-gap solution for now. Also, consider (I do not know if it's more complex or not) the option I suggested in https://github.com/xournalpp/xournalpp/pull/1822#issuecomment-608270083 --- If I could have the highlighter on button no.2 of my pen (I have the eraser on button 1) that would basically solve the problem.

redweasel:
@Rmano if it's a tool, you can make a button on your pen change to that tool in the preferences.

Rmano:
> @Rmano if it's a tool, you can make a button on your pen change to that tool in the preferences.

The problem is that the pointer highlighter is an option of the normal pen, not a tool --- so to change it you have to go into the preference. At least *I think*...

redweasel:
That's what I meant in my comment by making the **tool** bold. Because it is important, that it is it's own tool.

Rmano:
> That's what I meant in my comment by making the **tool** bold. Because it is important, that it is it's own tool.

Ah, ok, got it. Yes, a different tool will do, and you are right, it fits better in the philosophy of the application.

rinakra:
First of all, thank you so much for all the work your doing on this feature. I want to say that *just* having a red dot laser pointer, while useful, isn't really as useful as having a laser similar to what I've seen in Goodnotes with ipads. 

I use Xournalpp to teach math and if I want a red-dot laser I use the "[find-cursor](https://github.com/arp242/find-cursor)" program with a keyboard shortcut and some options. (for anyone curious, download the package and run "find-cursor --repeat 0 --follow --distance 1 --line-width 16 --size 16 --color red")

So I just wanted to strongly advocate for a laser pointer that actually writes some kind of bright or transparent (highligher-like maybe?) line and then fades away after a couple seconds. That would be really extremely useful.

src/gui/PresentationLaser.cpp
```diff
@@ -0,0 +1,62 @@
...
+    // The laser will move, so redraw the current position
+    const Rectangle<double>&& oldBounds = this->getDrawBounds();
+    gtk_widget_queue_draw_area(this->widget, oldBounds.x, oldBounds.y, oldBounds.width, oldBounds.height);
+
+    const double angle = std::atan2(distY, distX);
+    const double dx = std::abs(distX) < this->speed ? distX : speed * std::cos(angle);
+    const double dy = std::abs(distY) < this->speed ? distY : speed * std::sin(angle);
+    this->x += dx;
+    this->y += dy;
+
```
mjg, line 39: If you limit xspeed and yspeed separately this can lead to a strange effect: If you move almost horizontally or almost vertically (only one direction faster than speed) then the laser will move at a steeper angle, rich that horizontal or vertical line and then move strictly horizontally or vertically.

ShamanthMurali:
Is this kinda laser possible

(Jamboard by google)

kauesena:
The behaviour suggested by @ShamanthMurali would be great. The app notability for ipads also has a similar behaviour. What is drawn with the laser pointer fades alway uniformely after a short while. [This video](https://youtu.be/sET8lFnWy5Y?t=406) shows the feature.

K4LCIFER:
> Is this kinda laser possible  (Jamboard by google)

The issue with this one is that the text often disappears before you finish writing it. I think it would be better to have more of a global timer for the laser, so when ever you write with the laser pen, it resets the timer countdown; when you are finshed writing the timer will expire, and, when it does, all the text fades away at once.

There is an existing program that does exactly this. I cant remember off the top of my head what the program is called, but it is the notetaking program that is often used on iPads.

RunningOffCaffeine:
@K4LCIFER that app is called Good notes, I think. I agree that would be a better implementation, so that you can temporarily write out a full sentence, and it only disappears after you are done.

Rmano:
(commenting here prompted by https://github.com/xournalpp/xournalpp/issues/1778#issuecomment-1519130603).

I think there are many possible "presentation laser" possibilities here. I think a tool that makes it possible to temporarily write or highlight on the page (and then remove the things) is the basic needed function. I am now using https://github.com/bk138/gromit-mpx for my online presentations --- it does what's needed, although the problem is that it's keyboard-bound and so quite cumbersome to use when also writing with a tablet.

I envision a tool that I can select, starts "temporary write/highlight" mode, and that "opens" a sub-tool option board (to exit, clear screen, and change from several drawing mode like line, arrow, highlight) exactly like `gromit-mpx` but controlled just with the tablet. I would avoid fancy things like fading (or at least it can be added to the writing modes...)

Psi505:
5 years and still nothing :'(

SirGamsay:
The Feature to enable writing with the Pen and then it disapears after a short while would be perfect.

bhennion:
Superseeded by #6391

PR #1830, opened 2020-04-04
Fix zooming with ctrl + scroll 

### LINKED ISSUE #1587: Mouse Wheel zoom changes pages
**Affects versions :**
 - OS: Arch Linux
 - In case of linux, which desktop environment XFCE4
 - Which version of libgtk do you use. gtk-xfce-engine 2.10.1-2
 - Xournal++ 1.0.15

**Describe the bug**
When using a large PDF (100+ pages) and trying to zoom with ctrl + mousewheel, each mouse wheel step, in addition to changing the zoom, will also change the page in view forward or backward, depending on zooming in or out

**To Reproduce**
1. Open up a large PDF (100+ pages)
2. Go randomly in the middle
3. Look at the current page.
4. ctrl + mousewheel zoom
5. Look at the current page. It changed

**Expected behavior**
The page should stay the same, just the zoom should change

### LINKED ISSUE #1753: Zooming in with Ctrl+Scroll causes the document to scroll down
**Affects versions :**
 - OS: Arch Linux
 - Desktop environment: X11
 - Which version of libgtk do you use: 3.24.14+9+g429a6287e0-1
 - Version of Xournal++: 46a4dc82
 - Installation method: Arch User Repository

**Describe the bug**
Whenever I try to zoom in by pressing Ctrl + scrolling up/down with the scroll wheel, the document does zoom in/out, but also scrolls. This happens with both a mouse, and a touchpad. On small documents, this is not so important, but when using larger documents, it is more noticeable.
Zooming in with touchscreen works perfectly

**To Reproduce**
1. Make a new journal with many pages
2. Press Ctrl and use the scroll wheel to zoom in

**Expected behavior**
I would expect the document to zoom like it does, but stay focused on the same page

**Screenshots of Problem**
xournal
Initially on page 5, zooming causes it to go to page 10

**Additional context**
I have this issue ever since I started using Xournal++ (about october 2019)

jvwowd:
Fixes #1753
Fixes #1587 (which is listed as a duplicate of #1516 though I'm not sure it is)

The 'zoomWidgetPosX' and 'zoomWidgetPosY' are actually offsets, not
positions, and the current code was not taking that into account when
setting the variables.

In addition, I don't believe the size of the toolbars (essentially
'wx' and 'wy' from what I can tell) are relevant. I tested and found
less scrolling when these were removed.

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -75,8 +75,8 @@ void ZoomControl::startZoomSequence(double centerX, double centerY) {
         this->zoomWidgetPosX = rect.width / 2;
         this->zoomWidgetPosY = rect.height / 2;
     } else {
-        this->zoomWidgetPosX = centerX;
-        this->zoomWidgetPosY = centerY;
+        this->zoomWidgetPosX = centerX - rect.x;
+        this->zoomWidgetPosY = centerY - rect.y;
     }
 
     this->scrollPositionX = (rect.x + this->zoomWidgetPosX) / this->zoom;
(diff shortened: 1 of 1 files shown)


jvwowd:
While this fixes the issue, it seems that there is still some jitter when scrolling, but this applies with either the toolbar buttons or with ctrl+scroll.

Technius:
This is a pretty good hotfix for the issue while we wait for the viewport refactor. Thanks for the PR!

PR #1837, opened 2020-04-06
Fixed memory leak in InputContext and PenInputHandler.

Febbe:
Refactored InputEvent.

### DIFF
--- a/src/gui/inputdevices/InputContext.cpp
+++ b/src/gui/inputdevices/InputContext.cpp
@@ -70,70 +70,67 @@ auto InputContext::eventCallback(GtkWidget* widget, GdkEvent* event, InputContex
 auto InputContext::handle(GdkEvent* sourceEvent) -> bool {
     printDebug(sourceEvent);
 
-    InputEvent* event = InputEvents::translateEvent(sourceEvent, this->getSettings());
+    InputEvent event = InputEvents::translateEvent(sourceEvent, this->getSettings());
 
     // Add the device to the list of known devices if it is currently unknown
     GdkDevice* sourceDevice = gdk_event_get_source_device(sourceEvent);
     GdkInputSource inputSource = gdk_device_get_source(sourceDevice);
     if (inputSource != GDK_SOURCE_KEYBOARD && gdk_device_get_device_type(sourceDevice) != GDK_DEVICE_TYPE_MASTER &&
-        this->knownDevices.find(string(event->deviceName)) == this->knownDevices.end()) {
+        this->knownDevices.find(string(event.deviceName)) == this->knownDevices.end()) {
 
-        this->knownDevices.insert(string(event->deviceName));
+        this->knownDevices.insert(string(event.deviceName));
         this->getSettings()->transactionStart();
-        auto deviceClassOption = this->getSettings()->getDeviceClassForDevice(string(event->deviceName), inputSource);
+        auto deviceClassOption = this->getSettings()->getDeviceClassForDevice(string(event.deviceName), inputSource);
         this->getSettings()->setDeviceClassForDevice(sourceDevice, deviceClassOption);
         this->getSettings()->transactionEnd();
     }
 
     // We do not handle scroll events manually but let GTK do it for us
-    if (event->type == SCROLL_EVENT) {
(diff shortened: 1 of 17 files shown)


LittleHuba:
Sorry currently no time to test anything. Please check if drawing continues when you start on a page leave it inbetween and reenter the same page while doing a continuous stroke. Also check if you can draw over multiple pages with a continuous stroke. If both work reliably your changes should be good.

Febbe:
No problem, I asked you, because you are the main author of this section. And you have the greatest overview over the design. Can you have a look into #1507 . My investigations to this topic is the reason for this PR, but it does not address it, I just found some bugs I spontaneous fixed on my way :D.

LittleHuba:
When I've got time I will look into it. But as a first comment: Xournal++ itself is not programmed in a way to handle simulatneous inputs. We would need to attach identifiers to all the handling code to support this. So easiest solution is most likely to just filter out these simultaneous events (even if that is not the goal in the long run).

src/gui/inputdevices/InputContext.cpp
```diff
@@ -70,70 +70,67 @@ auto InputContext::eventCallback(GtkWidget* widget, GdkEvent* event, InputContex
...
 
-    if (event->deviceClass == INPUT_DEVICE_IGNORE) {
+    if (event.deviceClass == INPUT_DEVICE_IGNORE) {
         return true;
     }
-
-    delete event;
-
-    // We received an event we do not have a handler for
+    g_warning("We received an event we do not have a handler for");
```
Technius, line 133: Can't this get very spammy? Maybe we should hide this behind the `DEBUG_INPUT` flag.

src/gui/inputdevices/InputEvents.cpp
```diff
@@ -97,48 +72,48 @@ auto InputEvents::translateDeviceType(GdkDevice* device, Settings* settings) ->
...
+    targetEvent.deviceClass = translateDeviceType(device, settings);
 
-    targetEvent->deviceName = const_cast<gchar*>(gdk_device_get_name(device));
+    targetEvent.deviceName = const_cast<gchar*>(gdk_device_get_name(device));
 
     // Copy both coordinates of the event
-    gdk_event_get_root_coords(sourceEvent, &(targetEvent->absoluteX), &(targetEvent->absoluteY));
-    gdk_event_get_coords(sourceEvent, &(targetEvent->relativeX), &(targetEvent->relativeY));
+    gdk_event_get_root_coords(sourceEvent, &(targetEvent.absoluteX), &(targetEvent.absoluteY));
+    gdk_event_get_coords(sourceEvent, &(targetEvent.relativeX), &(targetEvent.relativeY));
```
Technius, line 91: Very minor, but the `&(targetEvent.XYZ)` style is not consistent with the `&targetEvent.pressure` below.

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -302,7 +286,7 @@ auto PenInputHandler::actionEnd(InputEvent* event) -> bool {
             if (!this->lastHitEvent) {
                 return false;
             }
-            currentPage = getPageAtCurrentPosition(this->lastHitEvent);
+            currentPage = getPageAtCurrentPosition(&this->lastHitEvent);
```
Technius, line 289: `AbstractInputHandler::getPageAtCurrentPosition` has a null check for the event argument. This could potentially change the semantics since `lastHitEvent` is now just initialized to some garbage event data.

src/gui/inputdevices/InputContext.cpp
Febbe: Mhh yes indeed, I wanted users to report that, since we should know when we miss events unintentionaly.

src/gui/inputdevices/PenInputHandler.cpp
Febbe: For that exact case, I added an `operator bool()` which returns `true`, if the underlying `GtkEvent*` is valid.

Febbe: fixed that thank you for pointing that out

Issue #1821, opened 2020-04-02
Handwriting on Wayland looks worse than under X11


anp369:
**Affects versions :**
 - OS: Ubuntu 18.04
 - (Linux only) Desktop environment: started with Ubuntu on Wayland
 - Which version of libgtk do you use: 3.22.30 (output of apt list)
 - Version of Xournal++: 1.0.17
 - Installation method: .deb from the github release page

**Describe the bug**
Text written with xournalpp under X11 looks smoother and cleaner than under Wayland.
For Comparison I added a screenshot with one sentence, written in X11 and in Wayland.
The X11 text is readable way better and feels more like what i actually wrote while the wayland text looks like the pointer jumps sometimes.

**To Reproduce**
1. Login to ubuntu using Wayland
2. Open xournalpp and write a sentence
3. logout and relogin using X11.
4. open the document and write the sentence again

**Expected behavior**
Both sentenced should look the same and should be equally readable

**Screenshots of Problem**
![Comparison of two versions of the sentence](https://user-images.githubusercontent.com/23340440/78274741-f0512e00-7510-11ea-8a96-1a59a90eac45.png)

**Additional context**
Tried different input system settings, None of them seems to fix the problem. The problem appears with both the old and the new input system.

lehmanju:
This might actually be caused by your gtk version. Gtk wayland support has made big steps the last few releases. Xournalpp does not differentiate between X11 and Wayland events, they are all generated by gtk.

anp369:
This just happened on a fresh Ubuntu 18.04 LTS installation. Strange that nobody else seems to have this problem and that even if the events should be the same, they turn out to have different results. Is there anything I could try to fix this? Are you interested in specific logs?

lehmanju:
The only thing you can try to do is install the latest gtk version (might have to add a ppa as your current version is not the latest).

anp369:
Thanks for your response,
I'll might try this in the next days. Did you experience this issue yet?

dadosch:
I think (at least that seems to be the case for me) that under wayland there is no pressure sensitivity…

Issue #1662, opened 2019-12-11
Problem with stylus writing


Maniform:
**Affects versions :**
 - OS: Linux
 - Version of Xournal++ 1.1.0+dev git commit 6e279639

**Describe the bug**
When I write with my stylus, when my stylus touches the screen, a line appears very often like you can see on the screeshots attached (it's more visible on circles).

**To Reproduce**
Just write with a random stylus, it will happens very often.

**Input Logs**
[debug_input.log](https://github.com/xournalpp/xournalpp/files/3952337/debug_input.log)
[debug_gdk_input.log](https://github.com/xournalpp/xournalpp/files/3952338/debug_gdk_input.log)
[xinput_list.txt](https://github.com/xournalpp/xournalpp/files/3952339/xinput_list.txt)

**Expected behavior**
I don't want theses lines to appear when I'm wrtting. It's harder when I try to write, because small letters become very difficult to be read.

**Screenshots of Problem**
![Capture d’écran de 2019-12-09 16-03-56](https://user-images.githubusercontent.com/44653826/70656824-177ea980-1c5b-11ea-84a9-884cdd40dc42.png)
![Capture d’écran de 2019-12-09 16-03-56_old](https://user-images.githubusercontent.com/44653826/70656945-4c8afc00-1c5b-11ea-8bb3-73316103172a.png)
![Capture d’écran de 2019-12-11 21-09-25](https://user-images.githubusercontent.com/44653826/70656803-0f266e80-1c5b-11ea-8fb0-17451952dd63.png)

Maniform:
I just upgraded from Ubuntu 18.04 LTS to Ubuntu 19.10 and discovered that it wasn't a problem with xournal++ but an problem with Ubuntu 18.04 which is not up to date with stylus writting.

PR #1854, opened 2020-04-09
Add option to ignore the first events of the stylus when touching the screen.

hard-zero1:
Add option to ignore the first events of the stylus when touching the screen to avoid artifacts at the beginnings of strokes. The number of ignored events can be set in the preferences.

I had this problem on Kubuntu 18.04: When the stylus is pressed on the screen, the accuracy and update rate of its recognized position increases. Because of the lower precision before, when hitting the screen the cursor might jump from its previous position to the more precise point (this is especially far if the pen tip was in rapid movement right before touching), while beeing recognized as writing. This can result in artifacts that e.g. make handwritten text very hard to read. To work around this, the first events since the triggering event can be ignored.

This works great for me (when ignoring the first 3 events) and should also be a workaround for #1662 and #1821 as they seem to be the same problem.

### DIFF
--- a/src/gui/dialog/SettingsDialog.cpp
+++ b/src/gui/dialog/SettingsDialog.cpp
@@ -34,6 +34,12 @@ SettingsDialog::SettingsDialog(GladeSearchpath* gladeSearchPath, Settings* setti
                      }),
                      this);
 
+    g_signal_connect(get("cbIgnoreFirstStylusEvents"), "toggled",
+                     G_CALLBACK(+[](GtkToggleButton* togglebutton, SettingsDialog* self) {
+                         self->enableWithCheckbox("cbIgnoreFirstStylusEvents", "spNumIgnoredStylusEvents");
+                     }),
+                     this);
+
 
     g_signal_connect(get("btTestEnable"), "clicked", G_CALLBACK(+[](GtkButton* bt, SettingsDialog* self) {
                          Util::systemWithMessage(gtk_entry_get_text(GTK_ENTRY(self->get("txtEnableTouchCommand"))));
@@ -202,6 +208,7 @@ void SettingsDialog::load() {
     loadCheckbox("cbHideVerticalScrollbar", settings->getScrollbarHideType() & SCROLLBAR_HIDE_VERTICAL);
     loadCheckbox("cbDisableScrollbarFadeout", settings->isScrollbarFadeoutDisabled());
     loadCheckbox("cbTouchWorkaround", settings->isTouchWorkaround());
+    loadCheckbox("cbIgnoreFirstStylusEvents", settings->getIgnoredStylusEvents() != 0);  // 0 means disabled, >0 enabled
     loadCheckbox("cbNewInputSystem", settings->getExperimentalInputSystemEnabled());
     loadCheckbox("cbInputSystemTPCButton", settings->getInputSystemTPCButtonEnabled());
     loadCheckbox("cbInputSystemDrawOutsideWindow", settings->getInputSystemDrawOutsideWindowEnabled());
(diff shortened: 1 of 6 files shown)


src/gui/dialog/SettingsDialog.cpp
```diff
@@ -202,6 +208,7 @@ void SettingsDialog::load() {
     loadCheckbox("cbHideVerticalScrollbar", settings->getScrollbarHideType() & SCROLLBAR_HIDE_VERTICAL);
     loadCheckbox("cbDisableScrollbarFadeout", settings->isScrollbarFadeoutDisabled());
     loadCheckbox("cbTouchWorkaround", settings->isTouchWorkaround());
+    loadCheckbox("cbIgnoreFirstStylusEvents", settings->getIgnoredStylusEvents() != 0);  // 0 means disabled, >0 enabled
```
Technius, line 211: Repeated use of `settings->getIgnoredStylusEvents() != 0` is error-prone. Please store it in a local variable `const bool ignoreStylusEventsEnabled = ...` and reuse on line 227.

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -216,6 +223,13 @@ void SettingsDialog::load() {
     GtkWidget* spAutosaveTimeout = get("spAutosaveTimeout");
     gtk_spin_button_set_value(GTK_SPIN_BUTTON(spAutosaveTimeout), settings->getAutosaveTimeout());
 
+    GtkWidget* spNumIgnoredStylusEvents = get("spNumIgnoredStylusEvents");
+    if (settings->getIgnoredStylusEvents() == 0) {  // 0 means checkbox disabled, the spinButton's value should be >= 1
```
Technius, line 227:
```suggestion
    if (!ignoreStylusEventsEnabled) {
```

src/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -49,7 +55,16 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
...
+        if (this->eventsToIgnore >= 0) {
+            if (this->eventsToIgnore == 0) {
+                this->eventsToIgnore = -1;
+                this->actionStart(event);
+            } else {
+                this->eventsToIgnore--;
+            }
+        } else {
+            this->actionMotion(event);
+        }
```
Technius, line 67: You can flatten the nested ifs:
```c++
if (this->eventsToIgnore == 0) {
    // ...
} else if (this->eventsToIgnore > 0) {
    // ...
} else {
    // ...
}
```

src/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -110,7 +127,9 @@ void StylusInputHandler::setPressedState(InputEvent const& event) {
     {
         switch (event.button) {
             case 1:
-                this->deviceClassPressed = true;
+                if (!this->eventsToIgnore >= 0) {
+                    this->deviceClassPressed = true;
+                }
```
Technius, line 132: Looks like a typo here? As it is currently written, `!this->eventsToIgnore` is the same as `this->eventsToIgnore != 0` but then `>= 0` always evaluates to true.

hard-zero1:
The mentioned typo was more than a typo, sorry. It should be fixed now.

src/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -92,8 +90,10 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
     // Trigger end of action if pen tip leaves screen or mouse button is released
     if (event.type == BUTTON_RELEASE_EVENT) {
         if (event.button == 1 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
-            if (this->inputRunning) {
+            if (this->eventsToIgnore < 0) {
                 this->actionEnd(event);
+            } else {
+                this->eventsToIgnore = -1;
```
LittleHuba, line 96: Should this not be done in either case?

hard-zero1: It is -1 already in the first case. But i can also set it every time if you want?

LittleHuba: Nah if it is -1 already then its good.

Technius:
Merging now, thanks for the PR!

Issue #1867, opened 2020-04-13
When moving to last page of an PDF document, xournal hangs (segfaults)


magsilva:
When moving to last page of an PDF document (https://dx.doi.org/10.1109/TSE.1985.232210, temporarily available at [here](https://www.dropbox.com/s/jde5gm6qxcag3nx/PROUST%3A%20Knowledge-based%20program%20understanding.pdf?dl=0)), xournal++ hangs (segfaults) when scrolling to the last page.

**Affects versions :**
 - OS: Linux (Fedora 31)
 - Desktop environment: X11
 - Which version of libgtk do you use: gtk3-3.24.13
 - Version of Xournal++:  git 4798e25 (from 2020/01/16), also with latest in trunk (commit d3e6bbe758dd55a1787b8aeb010df2b3aa5da6f5)
 - Installation method: Luya's COPR (https://download.copr.fedorainfracloud.org/results/luya/xournalpp/)

**To Reproduce**
1. Step 1: Load PDF document indicated at the beginning of the report
2. Step 2: Scroll down to latest page (or simply set page number to 9 in the spinpage box
3. Step 3: If scrolling down, you may have difficulties even to get to last page, but you can move the vertical scrollbar to reach it. If using the spinpage box, you can set that page directly. Either way, xournal++ will hang (and eventually segfault).

**Additional context**
When trying to debug the problem, I noticed that several events were emitted regarding the page change. After some fiddling around the code, I could narrow it down to the latest statement of 
`Layout::updateVisibility()`: `this->view->getControl()->firePageSelected(mostPageNr)`. For some reason, this was triggering an infinite storm of events.

Technius:
This is actually caused by #1660. Note that the last page of the PDF has a different size, which causes the issue when zoom fit is enabled.

Closing as duplicate.

PR #1868, opened 2020-04-13
Fix issue #1867.

magsilva:
When moving to the last page of the following PDF document, https://dx.doi.org/10.1109/TSE.1985.232210,
xournal++ hangs (segfaults). You can try to move to the last page either by scrolling
or by setting the page number in the spinpage box, the result will be the same.

After some debugging, I narrowed it down to the latest statement of Layout::updateVisibility():
this->view->getControl()->firePageSelected(mostPageNr). For some reason, this was triggering an
infinite storm of events.

My solution was to avoid firing the page selected event by checking if any change indeed occurred
in the layout. I used the variable mostPageNr to detect if there was any change: if it wasn't
changed by the updateVisibility()'s algorithm, the value would be the same as of its initilization.
By not calling firePageSelected if not change has happened, this issue was partially handled.

However, when scrolling, there was some issues, as if scrolling was malfunctioning. I had the
same issue with another PDF document (https://www.dropbox.com/s/f5f22fvgldcm2gm/Learning%20to%20program%20%3D%20learning%20to%20construct%20mechanisms%20and%20explanations.pdf?dl=0).
When looking at the change I did in Layout::updateVisibility(), I noticed that the page number
for 'mostPageNr' was off by one. After fixing this, I could scroll without problems to the
last page of the document (and actually in any PDF document).

### DIFF
--- a/src/gui/Layout.cpp
+++ b/src/gui/Layout.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cstdlib>
 #include <numeric>
+#include <optional>
 
 #include "control/Control.h"
 #include "gui/scroll/ScrollHandling.h"
@@ -68,7 +69,7 @@ void Layout::updateVisibility() {
     int y1 = 0;
 
     // Data to select page based on visibility
-    size_t mostPageNr = 0;
+    std::optional<size_t> mostPageNr;
     double mostPagePercent = 0;
 
     for (size_t row = 0; row < this->heightRows.size(); ++row) {
@@ -107,7 +108,9 @@ void Layout::updateVisibility() {
         x1 = 0;
     }
 
-    this->view->getControl()->firePageSelected(mostPageNr);
+    if (mostPageNr) {
+        this->view->getControl()->firePageSelected(mostPageNr.value());
+    }
 }
 
 auto Layout::getVisibleRect() -> Rectangle<double> {


Technius:
Thanks for the attempt, but the current selection behavior is actually correct. Observe the sidebar with the fix included in this PR:

xpp_scroll_off

As I've remarked in #1867, the real issue is actually #1660. I tried to fix this previously in #1691, but the solution was too hacky/complicated and I abandoned it. A better solution would be to temporarily block one of the zoom/scroll/select events to avoid the infinite loop (easiest would be the scroll events in `Layout`, using `gtk_signal_handler_block` and `gtk_signal_handler_unblock`).

magsilva:
@Technius, thank you for explaining the issue with the sidebar and providing further information on the actual issue. I took a look at issue #1795  and pull request #1813 (view rewrite) (and I will keep one eye on them). For now, I reverted my off-by-one "fix", but kept the remaining changes as they solved my current issue (although it is not a proper solution to the real problem).

Febbe:
An index of 0 is still valid. Changing to page 0 must therefore fire that event.

magsilva:
@Febbe, that could be addressed by initializing `mostPageNr` to an invalid page number and only firing the event if it is actually updated by the algorithm.

src/gui/Layout.cpp
```diff
@@ -62,7 +62,7 @@ void Layout::updateVisibility() {
     int y1 = 0;
 
     // Data to select page based on visibility
-    size_t mostPageNr = 0;
+    size_t mostPageNr = -1;
```
Febbe, line 65: please use something like `std::numeric_limits<size_t>::max()` and use a predefined constant instead of `-1` (which is also a negative number for unsigned values).
```suggestion
    size_t mostPageNr = INVALID_PAGE_IDX;
```
and define `INVALID_PAGE_IDX` as `constexpr auto INVALID_PAGE_IDX = std::numeric_limits<size_t>::max()` on the top of the source file.

You could also use an std::optional<size_t>, which is a way more readable and only uses 8 byte more on the stack.

src/gui/Layout.cpp
```diff
@@ -101,7 +101,9 @@ void Layout::updateVisibility() {
         x1 = 0;
     }
 
-    this->view->getControl()->firePageSelected(mostPageNr);
+    if (mostPageNr != -1) {
```
Febbe, line 104:
```suggestion
    if (mostPageNr != INVALID_PAGE_IDX) {
```

src/gui/Layout.cpp
magsilva: Thank you, I didn't know about std::optional. I rewrote the patch using it.

PR #1889, opened 2020-04-21
Fixes #1870

### LINKED ISSUE #1870: Linux Build Instructions should clone from https
**Affects versions :**

Building from GitHub clone

**Describe the bug**

Instructions should state to clone from https, not http.

**Additional context**

https://github.com/xournalpp/xournalpp/blame/master/readme/LinuxBuild.md#L60

rickhg12hs:
Fixes #1870

### DIFF
--- a/readme/LinuxBuild.md
+++ b/readme/LinuxBuild.md
@@ -73,9 +73,7 @@ common available, therefore the Cairo PDF Export is without PDF Bookmarks.
 
 The binary executable will be placed in the `build/src/` subdirectory.
 
-To place the binary executable and support files in systemwide directories, execute `sudo make install`.
-
-After updating xournalpp, with perhaps `git pull`, execute `cmake ..`, `cmake --build .`, and optionally `sudo make install` again.
+After updating xournalpp, with perhaps `git pull`, execute `cmake ..` and `cmake --build .`.
 
 ## Packaging and Installation
 


readme/LinuxBuild.md
```diff
@@ -73,6 +73,10 @@ common available, therefore the Cairo PDF Export is without PDF Bookmarks.
 
 The binary executable will be placed in the `build/src/` subdirectory.
 
+To place the binary executable and support files in systemwide directories, execute `sudo make install`.
+
+After updating xournalpp, with perhaps `git pull`, execute `cmake ..`, `cmake --build .`, and optionally `sudo make install` again.
+
```
Technius, line 79: These are steps that we explicitly discourage below.

rickhg12hs: I did read that, but I don't understand why.  The procedure I mention works well for me.  Why shouldn't I do it?

Technius: There is no clean way to uninstall after running `sudo make install`. If you upgrade or uninstall xournalpp, you risk leaving behind files that could cause problems down the road. Additionally, `sudo make install` is likely to be incompatible with your system's package manager.

There is a `rpm/fedora/xournalpp.spec` file if you would like to build an RPM package.

rickhg12hs: OK, I made some changes.

readme/LinuxBuild.md
```diff
@@ -73,6 +73,8 @@ common available, therefore the Cairo PDF Export is without PDF Bookmarks.
 
 The binary executable will be placed in the `build/src/` subdirectory.
 
+After updating xournalpp, with perhaps `git pull`, execute `cmake ..` and `cmake --build .`.
```
rickhg12hs, line 76: Acceptable?

Febbe: You should also mention to cd into the builddir. Also an "touch ../CMakeLists.txt" should be called before "cmake .."

PR #1919, opened 2020-05-01
Add feature to create new xopp file from cli

### LINKED ISSUE #1872: new document filename specified from command line
It would be great if one could provide a file name for a new document via command line. 
So doing 
```
xournalpp new_file.xopp
```
could start xournalpp and the document would already be named `new_file.xopp`. In this case when the user would do a save, no new file-browsing pop-up would appear requesting for naming the doc. 
This is of course standard behavior for editors, etc. 
thanks

idotobi:
This pull request resolves #1872  .

I chose to create a new file by default when the filename does not yet exist, as this is the standard behaviour for most command-line-tools I know (e.g. vim). However, I realized that inkscape and gimp through an error if the file does not yet exist. Thus, I'm sure what the best solution is. Using a flag like `--new` to explicitly request a new file would also be an easy addition to this pull request

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1896,7 +1896,7 @@ void Control::showSettings() {
     delete dlg;
 }
 
-auto Control::newFile(string pageTemplate) -> bool {
+auto Control::newFile(string pageTemplate, Path fileName) -> bool {
     if (!this->close(true)) {
         return false;
     }
@@ -1905,6 +1905,9 @@ auto Control::newFile(string pageTemplate) -> bool {
 
     this->doc->lock();
     *doc = newDoc;
+    if (!fileName.isEmpty()) {
+        this->doc->setFilename(fileName);
+    };
     this->doc->unlock();
 
     addDefaultPage(std::move(pageTemplate));
--- a/src/control/Control.h
+++ b/src/control/Control.h
@@ -69,7 +69,7 @@ public:
 
 public:
     // Menu File
-    bool newFile(string pageTemplate = "");
+    bool newFile(string pageTemplate = "", Path fileName = "");
     bool openFile(Path filename = "", int scrollToPage = -1, bool forceOpen = false);
     bool annotatePdf(Path filename, bool attachPdf, bool attachToDocument);
     void print();
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -390,7 +390,11 @@ auto XournalMain::run(int argc, char* argv[]) -> int {
         g_object_unref(file);
 
         if (!p.isEmpty()) {
-            opened = control->openFile(p, openAtPageNumber);
+            if (g_file_test(optFilename[0], G_FILE_TEST_EXISTS)) {
+                opened = control->openFile(p, openAtPageNumber);
+            } else {
+                opened = control->newFile("", optFilename[0]);
+            }
         } else {
             string msg = _("Sorry, Xournal++ cannot open remote files at the moment.\n"
                            "You have to copy the file to a local directory.");


Febbe:
I think the default behaviour is not to create that file in the filesystem.
Most software will just create a new file internally with a predefined save path specified.

src/control/XournalMain.cpp
```diff
@@ -346,7 +346,12 @@ auto XournalMain::run(int argc, char* argv[]) -> int {
         g_object_unref(file);
 
         if (!p.isEmpty()) {
-            opened = control->openFile(p, openAtPageNumber);
+            if (g_file_test(optFilename[0], G_FILE_TEST_EXISTS)) {
+                opened = control->openFile(p, openAtPageNumber);
+            } else {
+                GString* fileName = g_string_new(optFilename[0]);
+                opened = control->newFile("", fileName->str);
```
Febbe, line 353: Don't create a GString on the heap, this is a memory leak. Just pass optFilename.
```suggestion
                opened = control->newFile("", optFilename);
```

src/control/Control.h
```diff
@@ -69,7 +69,7 @@ class Control:
 
 public:
     // Menu File
-    bool newFile(string pageTemplate = "");
+    bool newFile(string pageTemplate = "", string fileName = "");
```
Febbe, line 72:
```suggestion
    bool newFile(string pageTemplate = "", fs::path fileName = "");
```

```suggestion
    bool newFile(string pageTemplate = "", Path fileName = "");
```

src/control/Control.cpp
```diff
@@ -1898,7 +1898,7 @@ void Control::showSettings() {
     delete dlg;
 }
 
-auto Control::newFile(string pageTemplate) -> bool {
+auto Control::newFile(string pageTemplate, string fileName) -> bool {
```
Febbe, line 1901:
```suggestion
auto Control::newFile(string pageTemplate, Path fileName) -> bool {
```
```suggestion
auto Control::newFile(string pageTemplate, fs::path fileName) -> bool {
```

idotobi:
@Febbe :  I'm not fluent with `c++` and neither with `glib` so thanks for the detailed feedback.

On the general comment: I think my comment was misleading. As far as I think my code change would not directly create the file when it's missing, but rather open xournalpp and already call the file as given in the cli argument. Only when you save it should than write something on disk. But I'll double check, just to make sure.

Febbe:
> @Febbe : I'm not fluent with `c++` and neither with `glib` so thanks for the detailed feedback.
> 
> On the general comment: I think my comment was misleading. As far as I think my code change would not directly create the file when it's missing, but rather open xournalpp and already call the file as given in the cli argument. Only when you save it should than write something on disk. But I'll double check, just to make sure.

No problem. 
For me it also looks like, that your changes won't create a file directly.

src/control/Control.cpp
idotobi: I went with `Path`

src/control/Control.h
idotobi: I went with `Path`

src/control/XournalMain.cpp
idotobi: Good point. My intention was to make it more readable, but that's not really worth it in this case.

idotobi:
@Febbe : I incorporated your suggestions.

---

> For me it also looks like, that your changes won't create a file directly.

I just double checked. The workflow is as we said:
 1. `xournalpp my-new-file.xopp`  -> open new file in xournalpp and set the name to `my-new-file.xopp`
 2. Click `Save` in application -> create/save `my-new-file.xopp` on disk

I think this is quite intuitive behaviour. To take @Technius comment in the issue into account, we could implement a flag that prevents opening non existing files (e.g. `xournalpp --if-exists my-old-file.xopp`). Thus anyone who does not like the default behaviour could simply create an alias in their shell of choice. What do you think?

---

(@Febbe :  My commit is now linked to my account, right?)

LittleHuba:
I don't really see anybody having problems with the default usecase. But you can add that flag if you want to.

idotobi:
@LittleHuba: I think so, too. I just wanted to make sure that I don't pass over @Technius' concern. However, I'd say we can merge it and in case anyone really has trouble with that behaviour we can still add a flag as part of a new issue/PR.

idotobi:
FYI: already updated the commit based on @Febbe's change requests.

idotobi:
Does anyone have an idea why the CI fails on Windows?

Technius:
Looks like Azure Pipelines was updated yesterday to include msys2 by default, so our manual msys2 installation script is failing to work.

Febbe:
Can you do a `git rebase upstream/master` and `git rebase -i HEAD~3` to squash everything. Then do a `git push --force`

idotobi:
Done :)

PR #1925, opened 2020-05-02
Complete application Rewrite

lehmanju:
Redux pattern, value based semantics, glade files as resource and much more! ;)


CMakeLists.txt
```diff
@@ -75,9 +75,9 @@ find_package(Boost REQUIRED)
 
 pkg_check_modules(xournalpp_deps REQUIRED IMPORTED_TARGET GLOBAL
         "glib-2.0 >= 2.32.0"
-        "gtk+-3.0 >= 3.22.0"
+        "gtkmm-3.0 >= 3.24.0"
```
LittleHuba, line 78: Do not require a too recent version or we will have problems with non-up-to-date operating systems like Debian or OpenSUSE.

Febbe: In prinziple this is fine, because it will require a long time until we can release that version (milestone 2.0). Then we could backport it anyways.

base/include/xournalpp/view/Viewport.h
```diff
@@ -0,0 +1,37 @@
...
+    double rawScale;
+};
+
+struct Resize {
+    int width;
+    int height;
+};
+
+using ViewportAction = std::variant<Scroll, Resize, Scale>;
+using ViewportResult = std::pair<Viewport, lager::effect<ViewportAction>>;
```
arximboldi, line 35: You can use `lager::result<Viewport, ViewportAction>` for convenience here.

base/src/xournalpp/Xournalpp.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+auto update(AppState model, Action action) -> XournalppResult {
+    // To add more update functions see https://sinusoid.es/lager/modularity.html
+    // Consider doing this on the lowest layer of hierarchy possible
+    return std::visit(lager::visitor{[&](const ViewportAction& viewportAction) -> XournalppResult {
+                          auto [newViewport, effect] = viewportUpdate(model.viewport, viewportAction);
+                          model.viewport = newViewport;
+                          return {model, effect};
+                      }},
+                      action);
```
arximboldi, line 30: For less boilerplate and slightly nice syntax, you can use:
```
return lager::match(action)(
   [&](const ViewportAction& a) { ... },
   ...
);
```

base/src/xournalpp/gui/widgets/XournalWidget.cpp
```diff
@@ -0,0 +1,104 @@
+#include "XournalWidget.h"
+
+#include <cmath>
+#include <utility>
+
+#include <gdk/gdk.h>
+
+enum Direction { HORIZONTAL, VERTICAL };
+
+XournalWidget::XournalWidget(const lager::reader<Settings>& settings, lager::reader<Viewport> viewport,
```
arximboldi, line 10: Why take `settings` by `const&` and `viewport` by value? I'd take both by value (they have reference semantics under the hood anyways as you know.)

arximboldi: Oh, I see, because you are not copying `settings` into a local variable. You can still move `settings` into `operator[]` below though.

CMakeLists.txt
LittleHuba: Not quite sure Debian will have 3.24 by then but we can risk it 😉

LittleHuba: What creates more headaches is the usage of `zug` but that train has already departed (pun intended)

Febbe: Haha fun fact I wrote my thesis over delay prediction in passenger train systems.

lehmanju: @LittleHuba zug is header only as well and the plan is to use gtk4 as soon as its released.

LittleHuba: Ahh well we should download those headers autmatically then (for devs). If not even Arch had a package before I created one we can assume no other OS will have. So we should make the life of potential devs not harder than it is and remove that hurdle.

lehmanju:
This will hibernate for some time as we have to get the new library working first. Additionally, I planned to use GTK4 for this which unfortunately is not finished yet. However, it looks very promising it is getting released this year. With a lot to do on the library it doesn't make sense develop a GUI first when there is no actual way to store and render anything.

fsimonis:
Can you please elaborate on the state and plan of this PR a bit?
What will change in the code, and what will become obsolete?

I have been trying out a few things with the code base lately, but I am hesitant to invest the extra time it takes to finish changes and turn them into proper PRs. It would be pointless to invest time into code that will be removed in the future.

lehmanju:
We are basically writing a new application at this point. It will feature a Rust library, infinite pages and GTK4/Qt5.  The current code has proved to be too bad to do this in the current app. This PR will therefore not have anything in common with the current state of Xournal++.

fsimonis:
I recently attempted to migrate the application to using unique ownership via `unique_ptr` for page elements, so I can see why you would like to rewrite the application. The code is challenging.
Given the rewrite, there is no real reason to invest additional time into extending the current state of the code.

Regarding the new version:
* Are contributions to the rewrite appreciated?
* Is there anything specific contributors could give you a hand?

lehmanju:
@fsimonis Your help would be highly appreciated. Join us at https://matrix.to/#/!cBnigogrPmZFqkDJnj:matrix.org?via=matrix.org&via=tchncs.de&via=msrd0.de and I can invite you to our developer room.

Our Rust library isn't in a usable state right now, first milestone would be loading/saving files (although it will be a completely different file format). However, to test it, we would need a QT application that really only handles events and displays a pixelbuffer from the Rust library. We would be happy if you are interested.

lehmanju:
Since this is gaining more interest, here is the current plan:

Since drawing always has direct responses on the screen (displaying the drawn line) it can be seen as a game although a simple one. Therefore, we will use some patterns from game development for our rewrite. Furthermore, we will use Rust as language to improve maintainability in the future.

This has following implications:
- bevy/amethyst (game engines in Rust) both work with winit for input operations
- winit does not include pen input api
- client-toolkit used by winit for wayland does not have pen input either (that I am currently implementing)

Once at least one platform supports pen input I can start prototyping the rewrite.

Technius:
This is quite ambitious and too large for a single PR. I think this should be closed for now.

Febbe:
Maybe close it and create a milestone. There you can link this pr and all relating issues

Technius:
I think "whole application rewrite" warrants an entirely new repo / codebase, so I'm just going to close this.

PR #1933, opened 2020-05-04
Fix wording of new page template dialog

### LINKED ISSUE #1524: Reword "page template" dialogue
The "Page template" dialogue's settings are confusing. Specifically, there's text for each of the two possible check boxes saying "Use the settings above for the first Page" where the check box says "Copy last page settings". 

How can a setting be used for the first page if it depends on the last? 

Also "settings above" for the second check box -- does this include *all* settings above? (It doesn't, I think, only the last 2).

Finally, the form says "settings will be used when you create a new document", which to me suggests a blank document, rather than annotating a PDF. 

Suggest to describe what the effect is clearer, e.g.: "Copy current page's size when inserting a new page" and "Copy current page's style (background) when inserting a new page".

Remove "Settings will be used if you create a new document" to "Settings apply on restart" or similar (I'm not sure what is meant by the sentence, but it may be this?)

**Additional context**
I'm not the only one confused, see https://github.com/xournalpp/xournalpp/issues/1517#issuecomment-538848606

ETA: Current dialogue, for reference:
gnome-shell-screenshot-IN288Z

Technius:
Fixes #1524. Merging in 24h if there are no objections.

### DIFF
--- a/ui/pageTemplate.glade
+++ b/ui/pageTemplate.glade
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
-  <requires lib="gtk+" version="3.0"/>
+  <requires lib="gtk+" version="3.16"/>
   <object class="GtkAdjustment" id="adjustment1">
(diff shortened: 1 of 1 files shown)


Technius:
Preview:
image

PR #1941, opened 2020-05-05
Fix offset of normal pen cursor

idotobi:
This commit fixes the offset of the pen cursor.
The previous implementation was based on `centerX` and `centerY`.
However, `cairo_rectangle` expects to get the `topLeft` X and Y
coordinates.

### DIFF
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -422,8 +422,9 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
     }
 
     cairo_set_source_rgba(cr, r, g, b, alpha);
-    // Correct the offset of the coloured dot for big-cursor mode
-    cairo_rectangle(cr, centerX, centerY, size, size);
+    double topLeftX = centerX - size/2.0, 
+           topLeftY = centerY - size/2.0;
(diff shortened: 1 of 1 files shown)


idotobi:
As I debugged this I had a few questions.

 1. How do you nicely debug values inside functions that run in a constant loop (e.g. XournalppCursor::createHighlighterOrPenCursor). Can I print the value to some field in some toolbar, or is it better to use the debugger? How do you usually do this?
 2. Is there a reason for using a rectangle instead of a circle?
 3. Is there a reason for not showing the recangle in the cursorSize that is selected, especially when using the Highlighter.

I'm pretty sure inside a Pull Request is the wrong place to ask, but I also don't know were the right place would be. Any suggestions?

(add 2./3. if not I'd be happy to work on that)

LittleHuba:
Closed in favor of #1945

LittleHuba:
Some answers:
1. For testing either use debugger or print to console and remove before you submit the PR
2. Most likely because it once was 2px big and a circle with 2px diameter is not drawn very well
3. I don't really get what you mean with this

idotobi:
@LittleHuba : Thanks for the information! :pray: 
3. was regarding #1945 . I was trying to ask if there's a reason for #1945 not being implemented yet (as I didn't see the issue before writing the comment). But I take that you merging #1945 is answer enough :innocent:

PR #1944, opened 2020-05-06
Mini Change: Replace hardcoded pi value with preprocessor const

idotobi:


LittleHuba:
Thanks for killing this monster!

Issue #1852, opened 2020-04-09
LaTex - in Black Background isn't readable at all. The Font size and color can't be set manually.


dishantpandya777:
The issue is quite reproducible. 

When you are using default or white Paper / Background... There's no issue.

But when you change to say Black BackGround, then there is no way to cope up with LaTeX. The writing is not readable at all because of the Black Fonts in LaTex Output and the Black Background.

I tried to give the command to change the LaTex fonts color but they reverted with error. So, it seems there's no other option at present.

Please help as I needed to Create a Video Lecture and unable to handle this situation.

I am using in Linux (for above situation).

Thanks.

tattsan:
The following method is a workaround.

Xournalpp uses standalone.cls for temporary .tex file. And standalone.cls includes standalone.cfg.
So if you write  \Requirepackage{color} or  \Requirepackage{xcolor} in standalone.cfg, you can use color latex commands.

eldad-a:
> The following method is a workaround.
> 
> Xournalpp uses standalone.cls for temporary .tex file. And standalone.cls includes standalone.cfg.
> So if you write \Requirepackage{color} or \Requirepackage{xcolor} in standalone.cfg, you can use color latex commands.

Thanks!

Is there a way to have a specific `standalone.cfg` file  for `xournalpp`? 

Modifying the system-wide `standalone.cfg` file works. However, creating one in the `.xournalpp` folder, where the configuration files are, showed no effect. 
This is on `Ubuntu 18.04.4 LTS bionic`; xournalpp installed from the ppa, following the [instructions](https://github.com/xournalpp/xournalpp).
Had to install the dependencies listed [here](https://github.com/xournalpp/xournalpp/blob/master/readme/LinuxBuild.md), otherwise xournalpp crashes when opening `preferences` from the `edit` menu; this is reported in issue [#917](https://github.com/xournalpp/xournalpp/issues/917).


BTW, on my system the `p` in `\Requirepackage` needed to be upper-case, that is:
```LATEX
\RequirePackage{xcolor}
```

Technius:
I'm planning on adding configuration options to the latex tool in the upcoming week(s).

tattsan:
[...]

PR #1952, opened 2020-05-10
Latex tool enhancements

### LINKED ISSUE #1188: LaTeX preamble customization
I want to be able to customize the preamble of the LaTeX template so I can e.g. add additional packages or define custom variables.

There should be a setting that allows users to customize the preamble. This setting could be global, or preferably, per-document (for example, you may want to define variables specific to the document being worked on).

We can implement a more limited subset of the preamble feature, such as packages or custom variable names.

**Additional context**
Per-document settings would first require the addition of a "Document Settings" dialog that can be used to configure options for a particular document.

Technius:
This PR implements various enhancements for the LaTeX tool, including:
* [x] Configuration settings panel for the LaTeX tool
* [x] Customizable template (global). Closes #1188, #1852.
  * [x] Load and use user-specific template file
  * [ ] Project-specific settings
  * [x] Provide variable for current font color
* [x] ~Multiline LaTeX input, which can support large formulas as well as graphics (e.g. tikz).~
  * Apparently this is already supported by pressing <kbd>Shift</kbd><kbd>Enter</kbd>
* [x] Button to test LaTeX configuration (useful for debugging LaTeX setup/installation)
* [x] ~Switch default PDF generation command from `pdflatex` to `latexmk` by default~ (see https://github.com/xournalpp/xournalpp/pull/1952#issuecomment-631205182)
* [x] Customizable `kpsewhich` and PDF generation commands
  * [x] Remove custom `kpsewhich` setting
  * [ ] Improve error feedback
* [ ] Documentation

### DIFF
--- a/resources/default_template.tex
+++ b/resources/default_template.tex
@@ -8,6 +8,10 @@
 \usepackage{ifthen}
 \newlength{\pheight}
 
+% Color support
+\usepackage{xcolor}
+\definecolor{xpp_font_color}{HTML}{%%XPP_TEXT_COLOR%%}
+
 % User input
 \def\preview{\(
   \displaystyle
@@ -21,5 +25,5 @@
   \ifthenelse{\pheight=0}{\GenericError{}{xournalpp:blankformula}{}{}}
 
   % Render the user input
-  \preview
+  \textcolor{xpp_font_color}{\preview}
 \end{document}
--- a/resources/default_template.tex
+++ b/resources/default_template.tex
@@ -1,0 +1,25 @@
+\documentclass[varwidth=true, crop, border=5pt]{standalone}
+
+% Packages
+\usepackage{amsmath}
+\usepackage{amssymb}
+
+% Blank formula checking
+\usepackage{ifthen}
+\newlength{\pheight}
+
+% User input
+\def\preview{\(
+  \displaystyle
+  %%XPP_TOOL_INPUT%%
+  \)
+}
+
+\begin{document}
+  % Check if the formula is empty
+  \settoheight{\pheight}{\preview}
+  \ifthenelse{\pheight=0}{\GenericError{}{xournalpp:blankformula}{}{}}
+
+  % Render the user input
+  \preview
+\end{document}
--- a/resources/default_template.tex
+++ b/resources/default_template.tex
@@ -4,6 +4,9 @@
 \usepackage{amsmath}
 \usepackage{amssymb}
 
+% for storing in memory verbatim content to be reused later
+\usepackage{scontents}
+
 % Blank formula checking
 \usepackage{ifthen}
 \newlength{\pheight}
@@ -13,17 +16,18 @@
 \definecolor{xpp_font_color}{HTML}{%%XPP_TEXT_COLOR%%}
 
 % User input
-\def\preview{\(
-  \displaystyle
-  %%XPP_TOOL_INPUT%%
-  \)
-}
+\begin{scontents}[store-env=preview]
+	\( 
+	\displaystyle
+    %%XPP_TOOL_INPUT%%
+	\)
+\end{scontents}
 
 \begin{document}
   % Check if the formula is empty
-  \settoheight{\pheight}{\preview}
+  \settoheight{\pheight}{\getstored[1]{preview}}
   \ifthenelse{\pheight=0}{\GenericError{}{xournalpp:blankformula}{}{}}
 
   % Render the user input
-  \textcolor{xpp_font_color}{\preview}
+  \textcolor{xpp_font_color}{\getstored[1]{preview}}
 \end{document}
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -19,41 +19,6 @@
 #include "i18n.h"
 #include "pixbuf-utils.h"
 
-/**
- * First half of the LaTeX template used to generate preview PDFs. User-supplied
- * formulas will be inserted between the two halves.
- *
- * This template is necessarily complicated because we need to cause an error if
- * the rendered formula is blank. Otherwise, a completely blank, sizeless PDF
- * will be generated, which Poppler will be unable to load.
- */
-const char* LATEX_TEMPLATE_1 = R"(\documentclass[varwidth=true, crop, border=5pt]{standalone})"
-                               "\n"
-                               R"(\usepackage{amsmath})"
-                               "\n"
-                               R"(\usepackage{amssymb})"
-                               "\n"
-                               R"(\usepackage{ifthen})"
-                               "\n"
-                               R"(\newlength{\pheight})"
-                               "\n"
-                               R"(\def\preview{\(\displaystyle)"
-                               "\n";
-
-const char* LATEX_TEMPLATE_2 = "\n\\)}\n"
-                               R"(\begin{document})"
-                               "\n"
-                               R"(\settoheight{\pheight}{\preview} %)"
-                               "\n"
-                               R"(\ifthenelse{\pheight=0})"
-                               "\n"
-                               R"({\GenericError{}{xournalpp: blank formula}{}{}})"
-                               "\n"
-                               R"(\preview)"
-                               "\n"
-                               R"(\end{document})"
-                               "\n";
-
 LatexController::LatexController(Control* control):
         control(control),
         settings(control->getSettings()->latexSettings),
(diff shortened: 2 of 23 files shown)


Technius:
Here's how to try it out if you're curious. The settings are located in the LaTeX tab of the settings dialog. Currently, only the template file feature is implemented (rest are WIP). The default template file is available in `resources/default_template.tex` (it is not detected automatically yet).
image

src/gui/dialog/LatexSettingsPanel.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+LatexSettingsPanel::~LatexSettingsPanel() {
+    g_object_unref(this->cbAutoDepCheck);
+    g_object_unref(this->globalTemplateChooser);
+}
+
+void LatexSettingsPanel::load(const LatexSettings& settings) {
+    gtk_toggle_button_set_active(this->cbAutoDepCheck, settings.autoCheckDependencies);
+    if (!settings.globalTemplatePath.empty()) {
+        gtk_file_chooser_set_filename(this->globalTemplateChooser, settings.globalTemplatePath.c_str());
```
Technius, line 19: Should be `settings.globalTemplatePath.u8string().c_str()` since glib assumes UTF-8 encoded strings.

src/control/settings/Settings.cpp
```diff
@@ -809,6 +818,15 @@ void Settings::save() {
     WRITE_BOOL_PROP(inputSystemTPCButton);
     WRITE_BOOL_PROP(inputSystemDrawOutsideWindow);
 
+    WRITE_BOOL_PROP(latexSettings.autoCheckDependencies);
+    // Inline WRITE_STRING_PROP(latexSettings.globalTemplatePath) since it
+    // breaks on Windows due to the native character representation being
+    // wchar_t instead of char
+    fs::path& p = latexSettings.globalTemplatePath;
+    xmlNode = saveProperty("latexSettings.globalTemplatePath", p.empty() ? "" : p.string().c_str(), root);
```
Technius, line 826: TODO: use `u8string()` instead of `string()`

Technius:
Rebased on master. Text colors are now available through the `%%XPP_TEXT_COLOR%%` string, which is set to the color of the text tool.

Technius:
Refactored the LaTeX code to be cleaner and more modular; it also now uses the more modern `GSubprocess` API. Custom PDF generation commands are now supported, but we actually can't switch to `latexmk` by default (yet) since its error codes do not distinguish between missing files and bad input. The alternative is to supply better error message feedback, so we don't have to pop up a dialog whenever there's an error.

rolandlo:
Hi @Technius 
Finally I took some time to test out your latex enhancements. It's awesome! I love to use tikz and pgfplots inside, for example to create graphs. You may consider to add something like this to the default file:
```
%for defining commands
\usepackage{xargs}

% for drawing and plotting
\usepackage{tikz}
\usepackage{pgfplots}
\pgfplotsset{compat=newest} % Allows to place the legend below plot
\newcommandx{\graph}[3][1=,2=]{
	\begin{tikzpicture}
	\begin{axis}[xlabel=$x$,ylabel=$y$, axis lines=center,samples=100, #2]
	\addplot[#1]{#3};
	\end{axis}
	\end{tikzpicture}
}

```
Then one can simply write something like `\graph{x^2}` or `\graph[domain=-2:2, blue][[x=1cm,y=1cm]{x^2}` to get a function graph. It will make it very easy to create graphs, see here:
grafik

Alternatively you could create a second template file (e.g. `tikz_template.tex`) which preloads the tikz and pgfplots packages and defines some useful shortcuts.

Technius:
Rebased on master with some bug fixes (now tested on Windows). I think it might be worth holding off on project-specific settings until we get a document-specific configuration UI. This PR should be ready for general review now.

@rolandlo Thanks for the feedback, I'm glad it works well! I only want to include the bare minimum in the default template, as it's likely that each person will want to set up their own defaults and workflow-specific commands. That's why the global template can be customized ;)

rolandlo:
> @rolandlo Thanks for the feedback, I'm glad it works well! I only want to include the bare minimum in the default template, as it's likely that each person will want to set up their own defaults and workflow-specific commands. That's why the global template can be customized ;)

Ok, I see your point. I just would like to make sure that even people who have little idea about LaTeX can use this feature. Among math teachers, for example, everyone can write something simple like `\graph{x^2}` or `\graph[domain=-3,3]{x^2}`, but not that many have used tikz and pgfplots and stuff like that before actually (in my experience). It needs to be explained somewhere. Maybe this would be a case for the new website?!

Technius:
A minor change before I take a few months-long break on open source: global template path is now set by to the default template by default.

@rolandlo @Febbe @LittleHuba Feel free to update/continue/merge this PR, as I don't have the time to work on it now.

rolandlo:
I will now have a close look at @Technius work and try to add the check he suggested.

resources/default_template.tex
```diff
@@ -0,0 +1,29 @@
...
+% Color support
+\usepackage{xcolor}
+\definecolor{xpp_font_color}{HTML}{%%XPP_TEXT_COLOR%%}
+
+% User input
+\def\preview{\(
+  \displaystyle
+  %%XPP_TOOL_INPUT%%
+  \)
+}
```
rolandlo, line 20: This will not be copied verbatim, which is a problem if a user wants to use e.g. code highlighting via the listings package or file contents inclusion via the filecontents package. It's better to use the scontents package to copy the user input verbatim.

ui/latexSettings.glade
```diff
@@ -0,0 +1,299 @@
...
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="halign">start</property>
+                            <property name="label" translatable="yes">The additional *.tex file defined for the journal.</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">1</property>
+                            <property name="top_attach">3</property>
+                          </packing>
+                        </child>
```
rolandlo, line 165: This is not yet used in the code.

ui/settings.glade
```diff
@@ -4396,6 +4396,30 @@ If available select &lt;small&gt;&lt;tt&gt;pulse&lt;/tt&gt;&lt;/small&gt; as inp
...
+              </object>
+              <packing>
+                <property name="position">10</property>
+              </packing>
+            </child>
+            <child type="tab">
+              <object class="GtkLabel" id="latexTabLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label" translatable="yes">Latex</property>
```
rolandlo, line 4416: Should be written "LaTeX" for consistency.

src/model/TexImage.cpp
```diff
@@ -66,93 +50,48 @@ auto TexImage::cairoReadFunction(TexImage* image, unsigned char* data, unsigned
...
-        this->read = 0;
+    const std::string type = binaryData.substr(1, 3);
+    if (type == "PDF") {
+        // Note: binaryData must not be modified while pdf is live.
+        this->pdf = poppler_document_new_from_data(this->binaryData.data(), this->binaryData.size(), nullptr, err);
+        if (!pdf || poppler_document_get_n_pages(this->pdf) < 1) {
+            return false;
+        }
+        PopplerPage* page = poppler_document_get_page(this->pdf, 0);
+        poppler_page_get_size(page, &this->width, &this->height);
```
rolandlo, line 77: The width and height should only be set for newly created TexImages. If the TexImage is loaded from a file, the saved width and height should not be overwritten.

src/gui/dialog/LatexSettingsPanel.cpp
```diff
@@ -0,0 +1,82 @@
...
+                msg = _("Sample LaTeX file generated successfully.");
+            } else {
+                msg = FS(_F("Error: {1}. Please check the contents of {2}") % err->message % tmpDir.c_str());
+                g_error_free(err);
+            }
+            g_object_unref(*proc);
+        } else if (auto* err = std::get_if<LatexGenerator::GenError>(&result)) {
+            msg = err->message;
+        }
+    }
```
rolandlo, line 76: Here (as indicated by @Technius ) a check is needed if the template path name is a regular file (not a folder, for instance).

src/control/LatexController.cpp
```diff
@@ -63,84 +36,21 @@ LatexController::~LatexController() { this->control = nullptr; }
...
     }
-    if (kpsewhichStatus != 0) {
-        string msg = FS(_F("Could not find the LaTeX package 'standalone'.\nPlease install standalone (found in "
-                           "texlive-latex-extra) and make sure "
-                           "it's accessible by your LaTeX installation."));
+    this->latexTemplate = std::string(std::istreambuf_iterator<char>(is), {});
+    if (!is.good()) {
+        string msg = _("Failed to read global template file. Please check your settings.");
         return LatexController::FindDependencyStatus(false, msg);
     }
```
rolandlo, line 49: Here (as indicated by @Technius ) a check is needed if the template path name is a regular file (not a folder, for instance).

josephwright:
Would it be possible to specify the path to LaTeX using this interface? On the Mac, GUIs don't inherit the path in any meaningful way.

Technius:
> Would it be possible to specify the path to LaTeX using this interface? On the Mac, GUIs don't inherit the path in any meaningful way.

You can configure the command used to run LaTeX. If it's insufficient, we can add another option to specify the relevant environment variables or something.

Uroc327:
Awesome work, thanks! Do you have any plans, when this feature will make it into a release?

Technius:
> Awesome work, thanks! Do you have any plans, when this feature will make it into a release?

This will be included in 1.1.0, but we currently don't have an ETA for 1.1.0's release. Probably sometime within the next few months. You could try the nightly builds if you really need this feature.

Felipe-Math:
Hello people,

I have the following situation: the background of my pages is dark (#202020), so I write (mostly) in white color. This generates 2 problems:

1. the LaTeX expression added has a white background and white font color, so I cant see my equation in the page (just a white rectangle);
2. the LaTeX preview window has white background, so I can't even preview the equation.

I solved the first situation inserting the command `\nopagecolor`, just after `\begin{document}` in the template .tex file. Now, my LaTeX equation is inserted with the correct font color (white) and transparent background, as expected.
This solves my first problem, but the second problem remains, because the LaTeX preview window still has a white background.

I would like to know if there is some way to change the background color of LaTeX preview window. I though in two possibilities:

1- Perhaps the preview window could detect the current page background color and use it; or
2- Perhaps the preview window background could be modified to be transparent, as the equation itself.

I dont know how to implement any of this solutions, so could someone help me? Or sugests another solution to my problem?

Thanks

p.s.: I'm not a english native speaker, so sorry for any mistake.

rolandlo:
@Felipe-Math Looks like what you want is already in PR #2809. It won't make it for the 1.1.0 version that is about to be released, but it will certainly be high on the priority list after that.

Felipe-Math:
@rolandlo 0k, thanks.

PR #1955, opened 2020-05-10
Add command for libgtk version to bug template

idotobi:
This is taken from the stack overflow answer: https://stackoverflow.com/a/126193

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -12,8 +12,10 @@ assignees: ''
 **Affects versions :**
  - OS: [e.g. Linux (which distro), Mac, Windows]
  - (Linux only) Desktop environment: [e.g. Wayland, Gnome-Wayland, XWayland, X11]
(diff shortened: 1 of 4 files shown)


Technius:
This command will only work if the appropriate `*-dev` or `*-devel` package is installed, which is not a reasonable expectation.

idotobi:
@Technius : Good point, then I'm gonna include it in the output of `xournalpp --version` as it would be nice to have an easy way for the users to obtain everything they need to know for a complete Bug report.

I'd also add it to the `About` window to enable even users who are not comfortable with the command line or do you think that's overdoing it?

LittleHuba:
It should definitely be part of the about window. Many people don't know their way around a terminal.

idotobi:
@LittleHuba : Would you mind reviewing my code change? I'd be happy to have your second opinion.

LittleHuba:
LGTM 
Nice and concise code.

idotobi:
Should I merge then?

idotobi:
Since there were no objections, I'll go ahead and merge.

PR #1967, opened 2020-05-11
Fixes #1961

### LINKED ISSUE #1961: Button-related settings are saved and used in the wrong order.
**Affects versions :**
 - OS: Linux (Kubuntu 20.04)
 - Installation method: deb from github

**Describe the bug**
The button-related settings entered from the settings dialog are saved and used in the wrong order.

**To Reproduce**
1. Edit->Preferences->Mouse->"**Right Mouse** Button" = "Insert Image"
2. Open .xournalpp/settings.xml, now `<data name=`**`"eraser"`**`><attribute name="tool" type="string" value="image"/></data>`

3. etc.

**Expected behavior**
Saving the settings for, e.g., Right Mouse should affect the behaviour of Right Mouse, etc.

**Additional context**
After a very quick look at the code, I think the problem comes from the fact that in [Settings.cpp](https://github.com/xournalpp/xournalpp/blob/bed6c310f46903c7e3a98b036a0f1bd07593508e/src/control/settings/Settings.cpp) the order of the button settings is inconsistent.
In particular, at line 25 the order is "**middle**", "**right**", "**eraser**", "touch", "default", "stylus", "stylus2" (this order is used in lines 471 and 627 to load/save the `buttonConfig` array from/to the config file) while lines 108-121 and 1458-1470 imply: **Eraser**, **Middle**, **Right**, Touch, Default, Pen button 1, Pen button 2 (this order is used to read from the `buttonConfig` array in the interface methods, e.g. `Settings::getEraserButtonConfig`, that the `Settings` class exposes to the rest of the code).
I suggest the entries in line 25 are reordered as follows:

        constexpr char const* BUTTON_NAMES[] = {"eraser", "middle", "right", "touch", "default", "stylus", "stylus2"};

lciti:
This pull request addresses issue #1961 (button-related settings are saved and used in the wrong order).

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -22,8 +22,6 @@
     com = xmlNewComment((const xmlChar*)(var)); \
     xmlAddPrevSibling(xmlNode, com);
 
-constexpr char const* BUTTON_NAMES[] = {"middle", "right", "eraser", "touch", "default", "stylus", "stylus2"};
-
 Settings::Settings(Path filename): filename(std::move(filename)) { loadDefault(); }
 
 Settings::~Settings() {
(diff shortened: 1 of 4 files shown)


src/control/settings/Settings.cpp
```diff
@@ -464,11 +469,32 @@ void Settings::loadDeviceClasses() {
...
+        case BUTTON_DEFAULT:
+            return "default";
+        case BUTTON_STYLUS:
+            return "stylus";
+        case BUTTON_STYLUS2:
+            return "stylus2";
+        default:
+            return "unknown";
+    }
+}
```
Febbe, line 491: Make this function constexpr.
Note: Also the best would be to return string_views.

src/control/settings/Settings.h
```diff
@@ -136,6 +145,7 @@ class Settings {
 
     void saveData(xmlNodePtr root, const string& name, SElement& elem);
 
+    static const char* getButtonConfigNames(int button);
```
Febbe, line 148: The function is only used in the sourcefile, so no explicit forward declaration is necessary.

src/control/settings/Settings.cpp
lciti: I am a bit rusty with C++ but will give it a try

src/control/settings/Settings.h
lciti: Sorry, I am not following you (as I said my C++ is rusty). How can I define it as a class member if not in the class declaration? Are you suggesting it should be a standalone function within the cpp file rather than a static class member? If that's the case, I am not sure I agree because semantically it's part of the class (any derived class would need to be aware of such mapping and possibly be able to use it, without having to "include" the cpp file).

Febbe: @lciti 
 > Are you suggesting it should be a standalone function within the cpp file rather than a static class member?

Yes it should be a standalone function in the source file.

 > If that's the case, I am not sure I agree because semantically it's part of the class (any derived class would need to be aware of such mapping and possibly be able to use it, without having to "include" the cpp file).

Yet no class derives from this and it is less likely this will happen. Also inheritance should be avoided in 99% of any cases. Also if that's the case the complete function can be moved into a header.
I also think the Settings class is not an EnumMapper, so basically this function belongs right to the Enum definition as free function, if used at many places.

Febbe: constexpr functions can't be split above source and header files, when they are used in many compilation units, because they must be available for each using compilation unit on compile time.
You would write such function like:   
`constexpr MyType my_func(Input in) {...}`   
When it is static:   
`constexpr static MyType my_func(Input in) {...}`

src/control/settings/Settings.cpp
lciti: The constexpr is fine but I get an error with string_view because string_views are not necessarily null-terminated while the function `s.child()` that takes the output of `getButtonConfigName` needs something that can be converted to `string` (its signature is `SElement::child(const string& name)`). On the other hand `string` cannot be `constexpr` in C++17. The two options I can see are `constexpr const char*` or `string` (like, e.g., in `drawingTypeToString` in ToolEnums.cpp). Any preference? I've gone with the first one for now but happy to change if you prefer.
PS: Also, I have now used `auto` with a trailing data type as this seems to be the style used in the rest of the file.

src/control/settings/Settings.h
lciti: Sorry, I missed your replies as I had not updated the page

lciti: > I also think the Settings class is not an EnumMapper, so basically this function belongs right to the Enum definition as free function, if used at many places.

That's a good point. If I understand correctly, you are saying the function should be in the header right after the definition of the corresponding enum.

src/control/settings/Settings.cpp
Febbe: Yes I knew about that habbit. Thats why I marked it as note. I prefer a char const* as return value. But if you still want to return a string_view you can still convert it into a string. A string can also be constructed from a char array without a null termination by providing the size of the char array.

lciti:
@Febbe please see if new version is fine.

Febbe:
When you change the parameter `int` of the buttonToString function to an Button, this PR is a LGTM.

src/control/settings/Settings.h
Febbe: Exactly 😬.

lciti:
> When you change the parameter `int` of the buttonToString function to an Button, this PR is a LGTM.

This breaks the `for` loops in `Settings::saveButtonConfig` and `Settings::loadButtonConfig`. Shall we keep `buttonToString(int)` or cast the loop iterator to the enum type: `buttonToString((Buttons)i)` or `buttonToString(static_cast<Buttons>(i))`?

Febbe:
The latter use modern c++11 casts

lciti:
> The latter use modern c++11 casts

Done

Febbe:
Very good work , thank you and keep on :).

PR #1968, opened 2020-05-12
Fix cursor update to apply always

### LINKED ISSUE #1954: Cursor size update does not occur when moving from toolbar to page

**Affects versions :**
 -   Operating System: Arch Linux
      Kernel: Linux 5.6.11-arch1-1
      Architecture: x86-64

 - X11
 - Which version of libgtk do you use: `3.24.20`
 - Installation method: build

**Describe the bug**
The pen/highlighter/eraser-cursors size does not update immediately after moving the cursor back to the page from the toolbar. Only if another type of cursor (e.g. mouse pointer, hand) is triggered in between, by (for example getting out of the page area) the cursor is updated upon reentry.
This does not effect the size of the drawn lines just of the cursor.

**To Reproduce**
1. Change Pen/Highlighter/Eraser cursor size in toolbar
2. Move back to page

**Expected behavior**
Cursor size is update right when moving back to the page.

**Screenshots of Problem**

xournalpp-cursor_size-update

xournalpp-cursor_size-update_eraser

**Additional context**

For the pen and highlighter cursor this was introduced by my previous PR #1945 as the size was never updated before.
However the erase had this issue already before.

idotobi:
This PR fixes #1954.
**Root Cause:**
The functions `createHighlighterOrPenCursor` and `getEraserCursor` were
returning early if the cursor type had not changed. This prevented the
update of the cursor size, based on the new setting.

### DIFF
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -307,12 +307,7 @@ void XournalppCursor::updateCursor() {
 
 
 auto XournalppCursor::getEraserCursor() -> GdkCursor* {
-
-    if (CRSR_ERASER == this->currentCursor) {
-        return nullptr;  // cursor already set
-    }
     this->currentCursor = CRSR_ERASER;
-
 
     // Eraser's size follow a quadratic increment, so the cursor will do the same
     double cursorSize = control->getToolHandler()->getThickness() * 2 * control->getZoomControl()->getZoom();
@@ -336,7 +331,6 @@ auto XournalppCursor::getHighlighterCursor() -> GdkCursor* {
     if (this->drawDirActive) {
         return createCustomDrawDirCursor(48, this->drawDirShift, this->drawDirCtrl);
     }
-
 
     return createHighlighterOrPenCursor(5, 120 / 255.0);
 }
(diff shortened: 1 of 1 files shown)


Febbe:
This function is called very often, did you checked if your change has a visible performance change?
LGTM.

idotobi:
@Febbe : So that was the reason for this check. I didn't notice any big perfomance drop when I tested it. But I also didn't specifically look out for it. I'd rather test it a little more before merging then.

Febbe:
I don't know, it looks like someone did an premature optimization. So probably this is not a problem. Premature optimizations shouldn't be done anyways. I would say we merge this and if someone figures out that he gets a performance drop. We should optimize it. But this time correctly.

idotobi:
I just looked at the code and I think we could even have a test if an update is necessary on the level of `XournalppCursor::updateCursor()` (https://github.com/xournalpp/xournalpp/blob/372ce72adfe8bc53d8cbb0b1b0baec2d1c1c22bf/src/gui/XournalppCursor.cpp#L186). Thus we the single functions don't need to care, because they would only be called if needed.

To clarify some pseudo code (aka. python):

```py
def updateCursor():
    if cursor_changed():
        return
    ...calling the specific update functions
```

idotobi:
> I don't know, it looks like someone did an premature optimization. So probably this is not a problem. Premature optimizations shouldn't be done anyways. I would say we merge this and if someone figures out that he gets a performance drop. We should optimize it. But this time correctly.

I just tested again. I didn't observe a big difference when comparing `xournalpp-1.0.18` with this branch.

Just out of curiosity: Would a decrease in performance result in the cursor lagging behind the pen position?

idotobi:
I'd say we can merge it. I would still create a refactor issue for myself to add a check in `void XournalppCursor::updateCursor()`. It shouldn't hurt and I would get to understand the code structure of the cursor file on a higher level then just individual functions.

idotobi:
@Febbe : Should I go ahead an merge it then?

LittleHuba:
Yeah I think you can go ahead. If any problems in performance are reported we can always fix them in a follow up PR.

PR #1980, opened 2020-05-14
Allow "tapping" when using the Select Rectangle and Select Region tools

lciti:
This PR implements the following feature: when using the Select Rectangle or Select Region tools, clicking on an object (without dragging to create the rectangle or the selection region) selects the object. This is similar to the experimental feature "Tapping" that allows the selection of an object when drawing.
I find this particularly useful as it allows me to use a single selection tool (I associate Select Rectangle to the middle mouse button) both to select multiple objects (dragging a rect around them) or a single one (with a single click).

### DIFF
--- a/src/control/tools/Selection.cpp
+++ b/src/control/tools/Selection.cpp
@@ -84,6 +84,8 @@ void RectSelection::currentPos(double x, double y) {
 
     this->ex = x;
     this->ey = y;
+
+    this->userTapped = false;
 }
 
 void RectSelection::paint(cairo_t* cr, GdkRectangle* rect, double zoom) {
@@ -165,6 +167,8 @@ void RegionSelect::currentPos(double x, double y) {
 
     // at least three points needed
     if (this->points && this->points->next && this->points->next->next) {
+
+        this->userTapped = false;
 
         auto* r0 = static_cast<RegionPoint*>(this->points->data);
         double ax = r0->x;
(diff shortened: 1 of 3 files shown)


Technius:
Does it work with the `Stylus > Ignore the first X events` setting enabled?

lciti:
> Does it work with the `Stylus > Ignore the first X events` setting enabled?

It does not work reliably with the stylus regardless of that setting (actually that setting seems to improve things a little bit). If I keep tapping on an object, it selects it once in five times or so. Even the standard Tapping feature in drawing mode is hit and miss for me (half of the time I manage to select, half of the time I draw a tiny segment).

Do you think it's due to the fact that the stylus generates more events than a mouse?
If so, one possible solution is to keep track of the maximum distance of any new point (recorded by `XxxxSelection::currentPos`) from the point from which the drag was initiated (the argument of the constructor, i.e. `this->sx/sy` for Rect and `this->points` for Region).
One issue is that the distance threshold would need to be computed in device space rather than drawing space (like the coordinates passed to `XxxxSelection::currentPos`). I could make `userTapped` into a function that takes `zoom` as argument so it can check whether the distance covered is larger than a threshold in device space (the caller `XojPageView::onButtonReleaseEvent` is aware of the zoom factor unlike the `XxxxSelection` object).
I'll do this in a new commit so you can see if this makes sense.

lciti:
@Technius The new version works with a stylus too. If you prefer this version, we can squash the two commits.

Technius:
Awesome, thanks for the PR!

Issue #1947, opened 2020-05-08
Too slow when open a note file with pdf background which has nearly 1000 pages 


McGiRG3z:
**Affects versions :**
 - OS: Manjaro Linux
 - (Linux only) Desktop environment: X11
 - Which version of libgtk do you use: 1:3.24.20-1
 - Version of Xournal++: 1.0.18
 - Installation method: AUR

**Describe the bug**
Too slow when open a note file with pdf background which has more than 1000 pages.

**To Reproduce**
1. Download this example pdf with 914 pages: https://libgen.is/book/index.php?md5=3886A4C1A684AB63E25E82E0A2D63DE9
2. Open it in Xournal++ as background
3. Take some notes on it, save and close
4. Reopen the note file

**Additional context**
It takes more than 10 seconds to reopen the note file, really slow...

PR #2002, opened 2020-05-21
Speedup Document Loading

fsimonis:
This PR speeds up the loading of PDFs and annotated PDFs significantly.

Main changes:
* Introduced an index from pdf page number to document page number, which reduces the lookup times of `Document::findPdgPage()` from `O(n)` to constant time.
  This index is generated on document read and on demand.
  Interactive document changes such as adding/inserting/deleting a page simply invalidate this index to keep the user interface snappy.
  This speeds-up the load-time of PDFs with many labels such as books.
* Changed the `xojfile/LoadHandler` to add pages in a bulk after parsing instead of calling `addPage` once per page while parsing. This required to add a bulk version `addPages`, which is templated and thus visible in the header. I hope that's not against some coding rules. 

Starting point for this PR was the issue #1947.
I tested this on my machine (X1 Yoga Gen 4) with `RelWithDebInfo` and `-march=native -fno-omit-frame-pointer` using clang 10. I also assume that the files were cached by the linux kernel during the following tests.

The _very vague_ baseline of the load time for an annotated version of the 914 page document was ~6 seconds. With these changes, the document loads in ~1 second.

_edit: sentence got chopped off somehow._

### DIFF
--- a/src/control/xojfile/LoadHandler.cpp
+++ b/src/control/xojfile/LoadHandler.cpp
@@ -225,6 +225,9 @@ auto LoadHandler::parseXml() -> bool {
 
     g_markup_parse_context_free(context);
 
+    // Add all parsed pages to the document
+    this->doc.addPages(pages.begin(), pages.end());
+
     if (this->pos != PASER_POS_FINISHED && this->lastError.empty()) {
         lastError = _("Document is not complete (maybe the end is cut off?)");
         return false;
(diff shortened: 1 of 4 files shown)


Technius:
Does this work fine with duplicated PDF pages? Otherwise, this looks fine to me.

fsimonis:
The original version of `findPdfPage()`returned the first document page number matching the pdf page number. The new indexing of the document indexes pages starting at the first document page. `unordered_map::emplace` does not overwrite existing keys, thus both algorithms result in the same mapping.

Old version:
https://github.com/xournalpp/xournalpp/blob/79c2fe7f15e7ba2578c5ee9ad66045f4d41bc925/src/model/Document.cpp#L172-L179

New version:
https://github.com/xournalpp/xournalpp/blob/54ac2b083ded8bad0707c455bcd2e4b6df196f06/src/model/Document.cpp#L221-L227

Technius:
LGTM, thanks for the PR!

Issue #888, opened 2019-02-16
Make color selectable why eraser is on


coldspark29:
When you choose a color, while the eraser is selected, the color should change accordingly and switch back to the pen. I happen to try this often, but then have to select the pen first in order to change the color

andreasb242:
Yes I agree, the disabled colors are badly visible... I also do this from time to time ;-)

cbm755:
As a long time xournal user, I'll just add that it behaves as @coldspark29 describes and I'll +1 this.

coldspark29:
I see that it has been implemented and it works great. In my opinion the crossed out color indicators are not necessary. There is already an indicator for the eraser being selected and it just makes things messy

PR #2004, opened 2020-05-21
Change color selection behaviour if tool is Eraser

### LINKED ISSUE #1546: Tool switching doesn't work correctly when using the erasor with a shortcut previously
**Affects versions :**
 - OS: Linux
 - In case of linux, which desktop environment: GNOME Wayland
 - Which version of libgtk do you use: 3.24.12
 - Version of Xournal++: 1.0.14

**Describe the bug**
When I set one of the buttons my of my pen (stylus?) to be a shortcut to eraser switching colours (or other tools rdoesn't work correctly)

**To Reproduce**
1. Set shortcut as mentioned above (in my case use Key2 for eraser)
2. Erase something, triggered by the shortcut
3. Stop pressing the key to switch back to the pen
4. Select another colour
5. When starting to draw it will switch back to the previous colour instead of using the newly selected colour. You have to draw once with the pen, then selecting the colour works.

**Expected behavior**

**Screenshots of Problem**
(Optional) If applicable, add screenshots to help explain your problem.

**Additional context**

### LINKED ISSUE #1712: Eraser button does not respect changes in writing tool
**Affects versions :**
 - OS: [e.g. Linux (which distro), Mac, Windows]: Kubuntu 19.04
 - In case of linux, which desktop environment [Wayland, Gnome-Wayland, XWayland, X11]: KDE
 - Which version of libgtk do you use: 3.24.8-1ubuntu1, 2.24.32-3ubuntu1 (not sure which one xournal++ uses)
 - Version of Xournal++ [e.g. 1.0.10, specific commit, specific branch]: 1.1.0+dev
    

**Describe the bug**
If you are using a stylus and change the input tool immediately after erasing, the app disregards your manual change and restores the input tool to what it was before erasing began.

**To Reproduce**
1. Select one input tool (e.g. highlighter) and draw something
2. Use the eraser button on a stylus to erase it
3. While still holding the eraser button, change the input tool to something else (e.g. pencil). Alternatively, move the stylus away from the screen so it is no longer detected by the computer and then change the input. I've found both cause the problem consistently. 
4. if you bring the stylus close to the screen (over the writing area), the input automatically switches to the highlighter even though I manually selected the pencil tool.

**Expected behavior**
I expect the app to respect the manual selection instead of overriding it with the input it had cached before the erase button was pressed. I'm guessing there's a simple fix (like checking if the current tool is still an eraser before automatically loading the previous tool) that would fix this.  

**Additional context**
Using a Lenovo Yoga 720 with the Lenovo Active Pen

idotobi:
This PR fixes  #1546 and #1712:

**Old Behaviour:**

 - Eraser selected from toolbar:
   Change of colour possible in toolbar. This, however, results in a switch to the pen regardless if you were just using the Highlighter or the Pen
 - Eraser selected from Stylus Button:
   Change of colour possible in toolbar. However, no effect on selected tool when previous tool was pen. When previous tool was Highlighter it even changes the current tool to the pen in some special edge cases.

**New Behaviour:**

 - Eraser selected from toolbar:
   No change of colour possible anymore as eraser does not have colour and
   it's not obvious for which other tool to set the colour.
 - Eraser selected via stylus:
   Allow change of colour in the toolbar. This colour will actually set
   the tools colour which was active before the button press. This makes
   sense as upon release of the eraser button this tool will be selected
   again.

@Febbe: I've seen that you introduced the _switch to Pen_ behaviour. Thus, I guess you will have the strongest opinion on this change. Please let me know what you think of this new behaviour.


Febbe:
Oh I did not implemented this, it was @andreasb242 to fix #888. I also think changing the color should swap to the pen. For me it look like the author of #1712 wants the same behav. like in #888, just for his corner case. 
Another question, I don't see how the color selection is related to #1712. In #1712 the problem is, that "old tool" is not overiden on selection, while holding a pen button and changing the tool in the toolbar. I think we should rename "old tool" to something like a "toolbar selected tool" and remove the dependency to the the eraser tool and other tools completely.

idotobi:
@Febbe: I like the idea of having "toolbar selected tool" and then "stylus selected tool" this should make the code way easier to understand. I'm gonna refactor my PR Accordingly. 

Also you're right about #1712. It seems like I misread the issue :facepalm: . Actually the PR should be linked to #1546.

The switching to pen behaviour feels still weird to me. To be honest, even after 3 years of extensive xournalpp usage I've never noticed that this behaviour existed :sweat_smile: 
I guess it feels weird to me because it's not obvious on first usage which tool you'll switch to since there are two tools (Highlighter and Pen) which have the capability to have colour.

idotobi:
@Febbe: I refactored the code using "toolbarSelectedTool" rather then "lastSelectedTool". In my opinion this improved this part of the code quite a bit and the behaviour when selecting colour is now exactly what I would expect.

The only thing I didn't implement is allowing to choose colour when the eraser was selected from the toolbar, as this doesn't make to much sense to me and also would not be implementable with "toolbarSelectedTool".

Febbe:
> @Febbe: I refactored the code using "toolbarSelectedTool" rather then "lastSelectedTool". In my opinion this improved this part of the code quite a bit and the behaviour when selecting colour is now exactly what I would expect.

Yes, I also think it is a way better. Maybe adding an additional Pointer "penSelectedTool" and just "swapping" the pointer into currentTool could remove all `if(this->toolbarSelectedTool)  statements

> The only thing I didn't implement is allowing to choose colour when the eraser was selected from the toolbar, as this doesn't make to much sense to me and also would not be implementable with "toolbarSelectedTool".

I would still prefer to ask the other devs what we want to define as intended behaviour. But for now this is good as is.

src/control/Tool.h
```diff
@@ -29,6 +29,8 @@ class Tool: public ToolBase {
 
     double getThickness(ToolSize size);
 
+    ToolType getToolType();
+
```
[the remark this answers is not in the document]
idotobi: Well spotted. This was actually a reminder from my previous attempt of solving the issue. Thus I'll delete it again. However, I should still make `const` a habit :sweat_smile: .

src/control/Tool.cpp
```diff
@@ -30,3 +30,5 @@ void Tool::setCapability(int capability, bool enabled) {
 auto Tool::hasCapability(ToolCapabilities cap) const -> bool { return (this->capabilities & cap) != 0; }
 
 auto Tool::getThickness(ToolSize size) -> double { return this->thickness[size - TOOL_SIZE_VERY_FINE]; }
+
+auto Tool::getToolType() -> ToolType { return this->type; }
```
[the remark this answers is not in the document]
idotobi: Same as above

idotobi:
> Yes, I also think it is a way better. Maybe adding an additional Pointer "penSelectedTool" and just "swapping" the pointer into currentTool could remove all `if(this->toolbarSelectedTool) statements

Thanks for the suggestions :pray: . That worked quite nicely. 

> I would still prefer to ask the other devs what we want to define as intended behaviour. But for now this is good as is.

Definitely as this could potentially be a breaking change for some peoples usage this should be agreed upon by all devs. Should I raise it on gitter, or how do you usually deal with such changes?

Febbe:
Yes in gitter

idotobi:
Short summary:

**Old behaviour:**
When eraser is selected in Toolbar you can change the colour and it also changes your tool to pen.

When eraser is selected by stylus button, changing colour is allowed but results in inconsistent behaviour.

**New Behaviour:**
When eraser is selected in Toolbar you cannot choose the colour. 

When eraser is selected by stylus button you can choose the colour and it sets the colour of the tool that was selected previously from the toolbar. Thus when you release the stylus button you get the previous tool with the changed colour. (also works with changing the tool in the toolbar)

src/control/ToolHandler.cpp
```diff
@@ -250,37 +256,38 @@ void ToolHandler::setLineStyle(const LineStyle& style) {
  * 			and therefore should not be applied to a selection
  */
 void ToolHandler::setColor(int color, bool userSelection) {
-    this->current->setColor(color);
+    this->toolbarSelectedTool->setColor(color);
+    this->buttonSelectedTool->setColor(color);
```
Febbe, line 260: I think this should only change the current selected tools color. Since it might be a bit confusing, having the high lighter on the pen button and changing with it the color of the pen.

idotobi: Hmm, good point. Since I'm only using the pen button as eraser I didn't think of this. I guess best would be:

```cpp
if (this->buttonSelectedTool->tool == TOOL_ERASER)
   this->toolbarSelectedTool->setColor(color);
```

What do you think?

Febbe:
```suggestion
    if (this->buttonSelectedTool->tool == TOOL_ERASER) {
        this->toolbarSelectedTool->setColor(color);
    } else {
        this->currentTool->setColor(color);
    }
```

idotobi: Nice one! Yeah that would be ideal. I will only be able to wrap it up next week. I would include this change and then rebase on master.

idotobi: It Turned out that it was actually more beneficial to do the following, as setting the color for the current Tool even if it's a Eraser (selected by the button) ensures that in the toolbar the correct color is highlighted.

```cpp
    if (this->buttonSelectedTool->tool == TOOL_ERASER) {
        this->toolbarSelectedTool->setColor(color);
    }
    this->currentTool->setColor(color); 
```

idotobi:
@Febbe : I finally managed to get back to this. Now it also works when the button selected tool is a tool which also has color (e.g. Highlighter).

If you approve I'd merge the pull request.

idotobi:
@Febbe : I just thought back on the naming of the variables and maybe it would be better to define:

* `mainTool` := `tollbarSelectedTool`
* `buttonTool` := `buttonSelectedTool`

It's slightly less obvious but more concise :thinking:

idotobi:
The last commit also fixes #2095 and make the whole behaviour more generic by differentiating tools based on capabilities not on toolType.

For a code refactor I think it would be handy to rewrite the the getter Functions in ToolHandler to include an optional argument which determines which for which tool to get the properties:

For example:

```cpp
auto ToolHandler::hasCapability(ToolCapabilities cap, bool mainTool) -> bool {
    if (mainTool == false) {
        return (currentTool->capabilities & cap) != 0;
    } else {
        return (toolbarSelectedTool->capabilities & cap) != 0;
    }
}
```

Could be rewritten to:

```cpp
enum ToolPointerEnum{CURRENT, MAIN, BUTTON};

auto ToolHandler::hasCapability(ToolCapabilities cap, ToolPointerEnum toolpointerEnum = CURRENT) -> bool {
    Tool* tool= getToolPointer(toolpointerEnum);
    return (tool->capabilities & cap) != 0;
}
```

with `CURRENT` being the default to ensure backward compatibility.

idotobi:
I just remembered  this pull request :sweat_smile: 
@Febbe : Could you please have a look over my last comment and code change. I'd value your opinion :)

Febbe:
Oh actually I thought this PR is merged already. I have an exam the next days, so this will take a while, until I can look over it.
@Technius do you have time for this?

idotobi:
@Febbe : No worries. Good Luck with the exam! :crossed_fingers: 

@Technius : In case you have a look at the PR, please also consider if you'd prefer to have the refactor of the ToolHandler getter Functions in this PR as well (see https://github.com/xournalpp/xournalpp/pull/2004#issuecomment-653742511 ). If you think that would be good I could still do it :innocent:

Technius:
Looks good so far. I think doing the refactor here would be good since it's only a few lines.

idotobi:
@Technius : The Refactor is done :smiley:

Technius:
I just tested this. Everything works for me except the behavior when the eraser is switched to from the stylus:
xpp_pr_bug

I get the same result with the commit before the refactor.

idotobi:
That is strange for me it works :thinking: 

Peek 2020-09-30 18-52

What Kind of Stylus and configuration do you use?

I have following Setup: x230T with default stylus and I use the eraser-button at the end for switching to eraser

image

idotobi:
@Technius : I just watched your GIF again. What strikes me is that even though the colour items are enabled and correctly show the pen icon, there is no "on-hover" animation when you hover over them and no active state change in the colour toolbar. 

Can you select anything else in the toolbar while the button on your stylus is pressed?

Technius:
> With my eraser button I can select items in the toolbar but with the side button pressed I cannot anywhere in the toolbar (e.g. saving the file).

I think this is the issue: I usually have my eraser bound to the side button. It works correctly if I bind eraser to the back button.

The side button eraser actually doesn't work on `master` anyways, so I think this PR is ready for merge. We could fix the side button eraser in a follow-up PR, if desired (I personally don't mind that much because I don't switch colors when I'm erasing).

apandada1:
> That is strange for me it works thinking
> 
> Peek 2020-09-30 18-52

Does this only work with a stylus?

idotobi:
@apandada1 : From the implementation side it should work with all all input devices. However, the Stylus is the only one where the button might stay active even when the cursor moves to the toolbar (at least on my device). Hence, the behaviour is only observable with the Stylus.

Is this different to what you experience on your device?

apandada1:
@idotobi I don't have a stylus, I use mouse. The behaviour changed when I upgraded from 1.0.19 -> 1.1.0+dev, that's why I wanted to know.

idotobi:
@apandada1 : Could you explain what the behaviour was on 1.0.19 and what it is on 1.1.0+dev? Then I could check it with the code.

apandada1:
@idotobi See https://github.com/xournalpp/xournalpp/issues/2527

PR #2005, opened 2020-05-22
Windows CI quickfix

Technius:
Delete the existing msys64 installation before building. This is a temporary hack until we figure out a better solution.


Technius:
Deleting and unpacking our artifact it is too slow (~8 minutes), so I'll just copy the install script here to see if it's faster...

PR #2009, opened 2020-05-23
Clear PDF cache when document changes

### LINKED ISSUE #1964: When annotating a PDF and opening new one, the new one won't show on the paper
**Affects versions :**
 - OS: Windows 10

**Describe the bug**
I grade homework that is submitted as PDF files. After I'm done grading one, I want to open the next one to annotate. When opening the new one, the paper resizes and the old PDF stays on the page. The new PDF won't load in. I have to close the program and open the new PDF again.

**To Reproduce**
1. Open a PDF with `File` > `Annotate PDF`
2. Annotate the PDF (I do it with the pen tool and my graphics tablet).
3. Open a new PDF with `File` > `Annotate PDF`

Reproduction rate: 100%

**Expected behavior**
The new PDF opens just like the first one.

clemdc:
This prevent previous PDF pages being rendered instead of new PDF ones.

Fixes #1964

### DIFF
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -699,6 +699,9 @@ void XournalView::documentChanged(DocumentChangeType type) {
     }
     viewPages.clear();
 
+    delete this->cache;
+    this->cache = new PdfCache(control->getSettings()->getPdfPageCacheSize());
+
     Document* doc = control->getDocument();
     doc->lock();
 
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -699,8 +699,7 @@ void XournalView::documentChanged(DocumentChangeType type) {
     }
     viewPages.clear();
 
-    delete this->cache;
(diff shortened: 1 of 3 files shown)


src/gui/XournalView.cpp
```diff
@@ -699,6 +699,9 @@ void XournalView::documentChanged(DocumentChangeType type) {
     }
     viewPages.clear();
 
+    delete this->cache;
+    this->cache = new PdfCache(control->getSettings()->getPdfPageCacheSize());
```
Technius, line 703: This works, but I think it would be better to make `PdfCache::clearCache` public and call it instead, as it makes the intent clearer.

clemdc:
Would you want me to submit a new PR with the public change and the modification in the side view as well ?

Febbe:
No its better to just do it in this PR.

Technius:
LGTM, thanks for the PR.

Issue #1972, opened 2020-05-12
Problem with Snapping


Fatih20:
Grid Snapping does nothing, when moving objects. I've been told that it enables snapping when objects are being moved to be constrained to an invisible grid in the document. Yet, as I demonstrated  in the attached video below, it didn't snap.

https://open.lbry.com/Screen-record-from-2020-05-12-19.12.20#9ce4e31b5bcd0ebc814b2443dd82469e73403834

The solution to this seems to be pretty obvious and I hope it can get fixed because it's a bad user experience to have a feature not working.

Febbe:
You should really work on your temper. Currently you are rude and unfriendly. In general this will lead to people not interested to help you out with that, or generally to deal with you. Also it will be very embarrassing if your problem was an RTFM or layer 8 problem.
Also you should stick to our bug template, currently I don't know if this could be a new bug, or a already fixed one. 

As first hint you should upgrade and then visit the preferences first. There are some additional options now. Also rewrite your issue if it still persists.

Fatih20:
> You should really work on your temper. Currently you are rude and unfriendly. 
Sorry, I was just using this app and got really frustrated about this problem.
> RTFM or layer 8 problem.
The user guide and the FAQ doesn't say anything about snapping, https://github.com/xournalpp/xournalpp/wiki/User-Manual#snapping
> As first hint you should upgrade and then visit the preferences first. There are some additional options now. Also rewrite your issue if it still persists.
I'm using the version available through the ppa. I did decrease snapping tolerance but it stayed the same when moving object. Effects of grid snapping can only be felt when making new object.

Febbe:
> > RTFM or layer 8 problem.

 > The user guide and the FAQ doesn't say anything about snapping, https://github.com/xournalpp/xournalpp/wiki/User-Manual#snapping

Oh clicking on that link is really frustrating. Expected just old infos, not empty chapters with an header.
Also this was just an example, bad behavior everytime falls back on your feet and even harder, if you aren't right.

Thanks for the additional informations. The work on the snapping function is still in progress, but it got already massive improvements the last few weaks. And I am aware of the problem, that some edit tools just don't respect the snapping.

It would be a great help, if you edit your issue and summarize all the tools which are currently affected.

rolandlo:
[...]

Issue #807, opened 2019-01-26
It would be nice if when snap to grid is selected, that moved objects would snap to grid.


frohro:
I'm creating slides for video lectures using xournalpp.  It is much nicer than xournal, which I used before, and getting much better all the time.  I often create a figure and then copy/paste it, and move the pasted object to a new location, for example when making schematic diagrams.  Then I modify the pasted figure.  I use snap to grid to make things line up nicely.  I have to be extremely careful when I move items to try and align them again on the grid or my modifications don't line up properly.

andreasb242:
There was already a ticket to this topic, @morrolinux did a config option or something?

For me it's not clear whats the problem.

morrolinux:
I think we're talking about a different thing @andreasb242 : he's talking about grid snapping for moving objects, not drawing.
I might have an idea on how to do this, but I'm busy with exams right now so I would do it in a couple of weeks or so I think...

frohro:
It appears that the VerticalToolHandler needs fixed too, because when you have snap to grid selected, you want it to snap to grid too.

rolandlo:
This has been fixed in #2011.

PR #2011, opened 2020-05-25
snapping for selections and vertical space

rolandlo:
This first commit implements snapping for moving selections, addressing issue #807 (and the duplicate #1972). The selection is snapped according to the the nearest corner of the reduced bounding box (discarding padding and stroke width). This still needs some fixes (in particular for scaling/rotating objects and for selections which do not merely consist of strokes) and improvements.

I do plan to add snapping for vertical space (mentioned in issue #807).

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -163,6 +163,9 @@ void Settings::loadDefault() {
     this->doActionOnStrokeFiltered = false;
     this->trySelectOnStrokeFiltered = false;
 
+    this->snapRecognizedShapesEnabled = false;
+    this->restoreLineWidthEnabled = false;
+
     this->inTransaction = false;
 }
 
@@ -459,6 +462,10 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         this->latexSettings.globalTemplatePath = fs::u8path(v);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("latexSettings.genCmd")) == 0) {
         this->latexSettings.genCmd = reinterpret_cast<char*>(value);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("snapRecognizedShapesEnabled")) == 0) {
+        this->snapRecognizedShapesEnabled = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("restoreLineWidthEnabled")) == 0) {
+        this->restoreLineWidthEnabled = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
     }
 
     xmlFree(name);
@@ -825,10 +832,12 @@ void Settings::save() {
     WRITE_INT_PROP(strokeFilterIgnoreTime);
     WRITE_DOUBLE_PROP(strokeFilterIgnoreLength);
     WRITE_INT_PROP(strokeFilterSuccessiveTime);
-
     WRITE_BOOL_PROP(strokeFilterEnabled);
     WRITE_BOOL_PROP(doActionOnStrokeFiltered);
     WRITE_BOOL_PROP(trySelectOnStrokeFiltered);
+
+    WRITE_BOOL_PROP(snapRecognizedShapesEnabled);
+    WRITE_BOOL_PROP(restoreLineWidthEnabled);
 
     WRITE_INT_PROP(numIgnoredStylusEvents);
 
@@ -1663,6 +1672,15 @@ auto Settings::getDoActionOnStrokeFiltered() const -> bool { return this->doActi
 void Settings::setTrySelectOnStrokeFiltered(bool enabled) { this->trySelectOnStrokeFiltered = enabled; }
 
 auto Settings::getTrySelectOnStrokeFiltered() const -> bool { return this->trySelectOnStrokeFiltered; }
+
+void Settings::setSnapRecognizedShapesEnabled(bool enabled) { this->snapRecognizedShapesEnabled = enabled; }
+
+auto Settings::getSnapRecognizedShapesEnabled() const -> bool { return this->snapRecognizedShapesEnabled; }
+
+
+void Settings::setRestoreLineWidthEnabled(bool enabled) { this->restoreLineWidthEnabled = enabled; }
+
+auto Settings::getRestoreLineWidthEnabled() const -> bool { return this->restoreLineWidthEnabled; }
 
 
 void Settings::setIgnoredStylusEvents(int numEvents) {
--- a/src/model/Element.cpp
+++ b/src/model/Element.cpp
@@ -11,11 +11,17 @@ Element::~Element() = default;
 
 auto Element::getType() const -> ElementType { return this->type; }
 
-void Element::setX(double x) { this->x = x; }
+void Element::setX(double x) {
+    this->x = x;
+    this->sizeCalculated = false;
+}
 
-void Element::setY(double y) { this->y = y; }
+void Element::setY(double y) {
+    this->y = y;
+    this->sizeCalculated = false;
+}
 
-auto Element::getX() -> double {
+auto Element::getX() const -> double {
     if (!this->sizeCalculated) {
         this->sizeCalculated = true;
         calcSize();
@@ -23,20 +29,28 @@ auto Element::getX() -> double {
     return x;
 }
 
-auto Element::getY() -> double {
+auto Element::getY() const -> double {
     if (!this->sizeCalculated) {
         this->sizeCalculated = true;
         calcSize();
     }
     return y;
 }
+auto Element::getSnappedBounds() const -> Rectangle<double> {
+    if (!this->sizeCalculated) {
+        this->sizeCalculated = true;
+        calcSize();
+    }
+    return this->snappedBounds;
+}
 
 void Element::move(double dx, double dy) {
     this->x += dx;
     this->y += dy;
+    this->snappedBounds = this->snappedBounds.translated(dx, dy);
 }
 
-auto Element::getElementWidth() -> double {
+auto Element::getElementWidth() const -> double {
     if (!this->sizeCalculated) {
         this->sizeCalculated = true;
         calcSize();
@@ -44,7 +58,7 @@ auto Element::getElementWidth() -> double {
     return this->width;
 }
 
-auto Element::getElementHeight() -> double {
+auto Element::getElementHeight() const -> double {
     if (!this->sizeCalculated) {
         this->sizeCalculated = true;
         calcSize();
@@ -52,7 +66,7 @@ auto Element::getElementHeight() -> double {
     return this->height;
 }
 
-auto Element::boundingRect() -> Rectangle<double> {
+auto Element::boundingRect() const -> Rectangle<double> {
     return Rectangle<double>(getX(), getY(), getElementWidth(), getElementHeight());
 }
 
--- a/src/model/Image.cpp
+++ b/src/model/Image.cpp
@@ -7,7 +7,7 @@
 
 #include "pixbuf-utils.h"
 
-Image::Image(): Element(ELEMENT_IMAGE) { this->sizeCalculated = true; }
+Image::Image(): Element(ELEMENT_IMAGE) {}
 
 Image::~Image() {
     if (this->image) {
@@ -27,13 +27,20 @@ auto Image::clone() -> Element* {
     img->data = this->data;
 
     img->image = cairo_surface_reference(this->image);
+    img->calcSize();
 
     return img;
 }
 
-void Image::setWidth(double width) { this->width = width; }
+void Image::setWidth(double width) {
+    this->width = width;
+    this->calcSize();
+}
 
-void Image::setHeight(double height) { this->height = height; }
+void Image::setHeight(double height) {
+    this->height = height;
+    this->calcSize();
+}
 
 auto Image::cairoReadFunction(Image* image, unsigned char* data, unsigned int length) -> cairo_status_t {
     for (unsigned int i = 0; i < length; i++, image->read++) {
@@ -76,7 +83,8 @@ auto Image::getImage() -> cairo_surface_t* {
     return this->image;
 }
 
-void Image::scale(double x0, double y0, double fx, double fy) {
+void Image::scale(double x0, double y0, double fx, double fy, double rotation,
+                  bool) {  // line width scaling option is not used
     this->x -= x0;
     this->x *= fx;
     this->x += x0;
@@ -86,9 +94,10 @@ void Image::scale(double x0, double y0, double fx, double fy) {
 
     this->width *= fx;
     this->height *= fy;
+    this->calcSize();
 }
 
-void Image::rotate(double x0, double y0, double xo, double yo, double th) {}
+void Image::rotate(double x0, double y0, double th) {}
 
 void Image::serialize(ObjectOutputStream& out) {
     out.writeObject("Image");
@@ -119,6 +128,10 @@ void Image::readSerialized(ObjectInputStream& in) {
     this->image = in.readImage();
 
     in.endObject();
+    this->calcSize();
 }
 
-void Image::calcSize() {}
+void Image::calcSize() const {
+    this->snappedBounds = Rectangle<double>(this->x, this->y, this->width, this->height);
+    this->sizeCalculated = true;
+}
--- a/src/model/Snapping.cpp
+++ b/src/model/Snapping.cpp
@@ -6,6 +6,16 @@ namespace Snapping {
 
 [[nodiscard]] inline double roundToMultiple(double val, double multiple) { return val - std::remainder(val, multiple); }
 [[nodiscard]] inline double distance(Point const& a, Point const& b) { return std::hypot(b.x - a.x, b.y - a.y); }
+double snapVertically(double y, double gridSize, double tolerance) {
+    double ySnapped = roundToMultiple(y, gridSize);
+    return abs(ySnapped - y) < tolerance * gridSize ? ySnapped : y;
+}
+
+double snapHorizontally(double x, double gridSize, double tolerance) {
+    double xSnapped = roundToMultiple(x, gridSize);
+    return abs(xSnapped - x) < tolerance * gridSize ? xSnapped : x;
+}
+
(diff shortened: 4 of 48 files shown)


rolandlo:
Currently snapping for moving selections looks like in the demo video below:

snappingDemo

rolandlo:
Now snapping works for selections consisting only of strokes, even when they have been scaled and rotated. Moreover you can now make selections as small as minSize (5 pixels in view coordinates) and do not lose the aspect ratio while scaling them up again.

A few things left:
- If the selection contains other elements than strokes, the snapping does not work properly yet.
- While the rotation angle is non-zero, resizing the selection results in a additional translation movement. This is due to the fact that resizing results in a change of the rotation center (two rotations with different centers and opposite angle combined yield a translation). This is not new and is not so problematic, because the selection can quickly be moved and snapped again.
- The cursor for resizing should be rotated along with the selection (issue #2021)
- It might be nice to have snapping for resizing selections as well. However, I am not sure, what the best behavior is, in particular for rotated selections. 
- Currently scaling fixes the corner or side of the full bounding box opposite to the anchor you use. Maybe it should fix the corresponding corner of the reduced bounding box (the non-visible one used for snapping). 
- Maybe the rotation center should be the center of the reduced bounding box.
- It would be nice to have the option to keep lines width (e.g. while holding the SHIFT key) for scaled selections.

rolandlo:
The last commit closes issue #2021.

rolandlo:
Now snapping should work for all selections, not just those consisting of strokes.

rolandlo:
Although there may still be room for improvement, this PR seems reasonably complete. So please check it out. 
@frohro @Fatih20 you may want to check, if the snapping behaviour suits your needs.

Fatih20:
> Although there may still be room for improvement, this PR seems reasonably complete. So please check it out.
> @frohro @Fatih20 you may want to check, if the snapping behaviour suits your needs.
I'll download the fork and compile it to test it later, thanks for the heads up!

frohro:
Thanks for your work on this Roland!  Here are my comments:

- I had to do a complete cmake, make, make install after switching to the extend-snapping branch.  i had linker issues if I did not redo the cmake after mistakenly building the master branch.
- The cursor render for resizing is not at the correct angle for the box you select for scaling.  See the video.  It seems to happen after rotating, but I'm not exactly sure yet.  
output

- I agree that it would be nice if scale changes kept the lines the same width, or at least there was an option  for that.  For me it would be the preferred behavior.  As it is, I don't use resizing at all on objects with lines, because of this.
- I agree that resizing snapping would be nice too.
Again, this is a great improvement!  Thanks for all your work Roland.  It will make my online classes better.

Rob

rolandlo:
Thanks @frohro for your feedback and for pointing out a bug with the angle for the cursor. I think I will find an easy fix for that. It should come down to resetting the angle to 0 after a selection has been finalized.

As for the line width I see three ways. 
1) make it configurable in the settings or 
2) have a key modifier (SHIFT or CTRL) or 
3) have both a setting for the usually desired behavior and have a key modifier to change it occasionally.

The downside of 2. is that it is not practical if you work mostly without keyboard. So I lean to going for 1 or 3. 

For resizing snapping: Do you have a clear idea what the behavior should be? I have a clear idea about it when the selection is not rotated, but no idea, if it is rotated. Maybe I could implement resizing snapping only for not rotated selections.

rolandlo:
@frohro: The cursor angle bug is now fixed. If you have time, please check it out again.

frohro:
The cursor bug is fixed for me.

Because xournalpp is in pretty constant flux, and the documentation is really lacking, I don't like Ctrl or Shift key features.  They are really hard to discover by users that aren't reading the code, which is most everyone.  So I vote for option one or three, and it might be nice to have a note in the settings that says Shift or Ctrl swaps the behavior while it is pushed if option three is pursued.

For resize snapping, you could snap the corner or even side closest to what is grabbed to the grid.  That way a user could resize first one side to fit the grid, and then the other, if they want two sides to fit the grid, at the expense of changing the size.  I think the size is the thing that should have the least precedence, since in xournalpp it is all relative anyway.  I really like how you made the snap to grid work for the side or corner closest to what you grabbed.  This would fit in nicely with your already constructed paradigm.

Thanks for all your work on this.  For me at least it is a tremendously useful feature. 

Rob

rolandlo:
Line with scaling can now be enabled or disabled. See in the drawing area settings.

frohro:
Thanks Roland,
I find the setting in the Drawing Area setting, but it behaves the same way with it checked or not.
Rob

rolandlo:
You will only see the difference after finalizing the selection (by clicking outside of it). Before you only see a preview (elements are not updated, only the cairo-surface is transformed). Take a look:
lineWidthDemo

frohro:
Ah!  I like it.  It makes the objects look so much better!  Thanks for the explanation.  We need to think about how to document this feature, as a lot of guys will be like me, and not unselect it until we like it.  :-)

Thanks!!
Rob

rolandlo:
resizing and rotation is now based on the reduced bounding box. This makes it much easier to get snapping for resizing selections as well.
The bug #2034 is still present in the PR and must be fixed there as well.

rolandlo:
Snapping for resizing selections is now implemented. If the selection is not horizontal the snapping acts like the rotation were not there. 
There are some problems with undo actions, in particular bug #2034 
Moreover while resizing the cursor sticks to the reduced bounding box and can be far from the resizing anchors. This may need correction.

Nevertheless snapping should work and may be useful, please check it out, Rob @frohro

frohro:
This looks pretty nice Roland.  I played around with it for a while, and I can get it to do what I want!  It is not quite how I envisioned it, but it works just as well, as I can still make end points snap where I want them to, and make things as large as I like, etc.  I played with it for about half an hour and could find no problems.

rolandlo:
Now undo actions work as expected.

rolandlo:
It is time for some review from one of the active developers @LittleHuba @Febbe @Technius

frohro:
Hi Roland,
There is one more thing I found that needs to snap to grid.  The shape recognizer does not step the recognized shapes to the grid.
Thanks for all your great work on this.  You are making xournalpp much better for my purposes at least.
Rob

rolandlo:
Hi Rob
Great to hear that my efforts to improve snapping help for your purposes. I am aware that the shape recognizer doesn't support snapping yet. It seemed to me (at first sight at least) to be a bit more complicated to do that, so I excluded it from this PR. I will think about it in future, though.
If the shape recognizer should support snapping, then there are two ways to approach it. Either snap the (points defining the) recognized shape to grid. Or try to recognize a shape which is aligned to grid to start with. The second approach is most likely more complicated. In my experience the shape recognizer only recognizes circles, straight lines and rectangles, although there is code in xournalpp supporting more complex shapes. 
Would it suffice for you if snapping for the shape recognizer would only snap recognized straight lines, nearly horizontal or vertical rectangles and circles? 
- in case of straight lines it would simply snap the end points to the grid.
- in case of nearly horizontal or vertical rectangles it would rotate them to be exactly horizontal/vertical and then snap the four corners to the grid
- in case of a circle it would snap the two points on the horizontal diameter to the grid.
Or what would you need it for exactly?

frohro:
Hi Roland,
For me the purpose of shape recognition is just so I don't have to select the right tool to make the shape neat.  
I, likewise, find the only shapes that actually work are the circle, rectangle, and line.  Sometimes I wish the arrow would work too, but I've never been able to make it work.  For me the recognition parameters are not loose enough to get good recognition about half the time, even trying hard to make it neat.  It would be an improvement if there were a sloppiness adjustment in the settings that would make it recognize sloppier shapes.
You described what I want for recognized shapes exactly.  This would make drawing block diagrams and schematics easier.  
Thanks again for all your work on xournalpp!
Rob

rolandlo:
There was a problem with resizing horizontal and vertikal strokes, which is fixed with the latest commit. Moreover I made the cursor stick to the visible bounding box, which improves the ux.

rolandlo:
Hi Rob
Please check out the snapping for shape recognizer. I found a simple concept for it: just snap the reduced bounding box to grid. This works in all cases (line, rectangle, circle) equally well.

frohro:
Thanks Roland!  I played with it for half an hour, trying to practice my box writing, etc.  It is very much better because you can have the items where you want them, but the recognition is still too spotty to make it actually useful.  Circles are easier to draw than boxes, but now sometimes my boxes get recognized as circles.  My issue, now is, even with practice, I make so many figures that are not recognized or not recognized correctly that the time I save by not having to select the tools is outweighed by the undo and redo operations I have to do to correct mistakes.  This addition is a big improvement, though its benefits won't be fully useful until the recognition gets better.  :-)  At least it makes the snap to grid complete, which I feel is a major benefit.

rolandlo:
Thanks for the feedback, Rob!
The shape recognizer code is pretty old. It was written by Denis Auroux, the main author of xournal, and hardly modified since then. The algorithm was documented at [the xournal repository](https://sourceforge.net/p/xournal/support-requests/6/) Some of the recognition features (like triangle recognition) have been disabled in xournalpp. I think, some work should be invested into improving this algorithm. I am not sure that fiddling around with the "sloppiness" constants (seen in ShapeRecognizerConfig.h) will bring much.
I did a quick test, drawing circles, line segments and rectangles by hand, 10 of each kind. From these 10/10 were recognized as circles, 9/10 as line segments, but only 4/10 as rectangles. So in my opinion one should try to improve on the rectangle recognition part. 
I have other plans though for the near future, though.
By the way I never had a box (rectangle) recognized as a circle.

src/model/Stroke.cpp
```diff
@@ -168,40 +168,32 @@ void Stroke::move(double dx, double dy) {
...
+        cairo_matrix_transform_point(&rotMatrix, &p.x, &p.y);
     }
     // Width and Height will likely be changed after this operation
     calcSize();
 }
 
-void Stroke::scale(double x0, double y0, double fx, double fy) {
-    double fz = sqrt(fx * fy);
+void Stroke::scale(double x0, double y0, double fx, double fy, double rotation, bool doScale) {
+    double fz = (doScale) ? sqrt(fx * fy) : 1;
```
Febbe, line 186: Just some notes, for me the functions needs to be evaluated for a refactor:
 - doesn't fx*fy == 1 imply, that doScale is false?
 - it look some kind of odd, that there is a rotation used to scale, this should be irrelevant, when previous translations are applied, or the element childs are positioned relative to to the element, isn't it?
 - the thickness `fz` should be calculated from the angle it has (Law of cosines), not just the pythagorean theorem. Currently the scaling is only correct for perfectly diagonal lines.
 - also this function looks more like a general translation function, maybe renaming it / creating a new would be more clear.

src/model/Text.cpp
```diff
@@ -109,3 +120,10 @@ void Text::readSerialized(ObjectInputStream& in) {
 
     in.endObject();
 }
+
+void Text::updateSnapping() {
+    this->xSnap = this->x;
+    this->ySnap = this->y;
+    this->wSnap = this->width;
+    this->hSnap = this->height;
+}
```
Febbe, line 129: add a newline

src/model/Text.cpp
```diff
@@ -65,9 +75,10 @@ void Text::scale(double x0, double y0, double fx, double fy) {
     this->font.setSize(size);
 
     this->sizeCalculated = false;
+    this->updateSnapping();
```
Febbe, line 78: call calculateSize before this line or remove it, since the width and height is not calculated yet.

src/control/settings/Settings.cpp
```diff
@@ -158,6 +158,7 @@ void Settings::loadDefault() {
     this->strokeFilterIgnoreLength = 1;
     this->strokeFilterSuccessiveTime = 500;
     this->strokeFilterEnabled = false;
+    this->scaleLineWidthEnabled = false;
```
Febbe, line 161: You changed the default behaviour of xournal++, this will be very unexpected, resizing written text is not possible without changing it back.

rolandlo:
Hi @febbe, thanks a lot for your helpful comments and explanations.

src/control/settings/Settings.cpp
rolandlo: Ok, I will set this to true as default.

src/model/Stroke.cpp
rolandlo: To scale the line width and pressure of points with the geometric mean of the scaling factors was not my idea, but makes some sense to me.

- In the special case where fx*fy==1 then the line width and pressure of points is left unmodified. The boolean doScale is about whether line width and pressure should be potentially modified after a selection has been resized.
- The rotation angle for this generalized scale function is only used for the Undo Handler (see ScaleUndoAction.cpp:46). The Undo Handler does not remember how the selection was rotated, when the scaling was done, it just remembers what it should undo/redo with the elements of the selection. Note that this function does scale in a rotated coordinate system. So instead of scaling coordinates by fx in direction of the usual x-axis and by fy in direction of the usual y-axis, it does scaling with respect to the rotated x- and y-axis.
- I think the thickness fz should be the same for the whole stroke, not depend on the angle of the single line segments. Otherwise, if you want to do it with "correct scaling", you get ugly strokes. This is already visible if you scale a rectangle (with round line caps) only in x-direction (or y-direction). See the following example:
grafik

- If you prefer, I can remove the rotation and change the undo / redo action. However the rotation function also uses a rotation center, so it is also not just standard rotation around the origin.

src/model/Image.cpp
```diff
@@ -76,7 +83,7 @@ auto Image::getImage() -> cairo_surface_t* {
     return this->image;
 }
 
-void Image::scale(double x0, double y0, double fx, double fy) {
+void Image::scale(double x0, double y0, double fx, double fy, double rotation, bool doScale) {
```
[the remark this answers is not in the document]
rolandlo: I know, but rotation may also be used by the Image class eventually. The boolean doScale will never be used, though. It is a virtual function, so all derived classes (from the base class Element) must use the same arguments, don't they? How should I go about it, if the argument doScale is only used by the derived class Stroke?

src/model/Element.cpp
```diff
@@ -31,9 +37,28 @@ auto Element::getY() -> double {
...
+    return xSnap;
+}
+
+auto Element::getSnapY() -> double {
+    if (!this->sizeCalculated) {
+        this->sizeCalculated = true;
+        calcSize();
+    }
+    return ySnap;
+}
```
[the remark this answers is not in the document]
rolandlo: I don't know yet what to do about that. We have the same issue with the unsnapped values as well. Those getters are called quite often and on various places. The snapped values are rarely called, in contrast.

rolandlo:
For some requested changes (about getters and about the scaling function) I have formulated questions (see above) and will wait for some feedback. The other changes are implemented.

src/model/Element.cpp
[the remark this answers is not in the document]
rolandlo: I tried to make all getters related to the bounding boxes const by making all cache variables mutable and all functions called by the getters const, but still failed. The compiler complained about the static function TextView::calcSize()

rolandlo:
@Febbe: Now I have made all requested changes except those you suggested for the scaling function (see my questions above). Tell me, if you still want me to change that after seeing my comments.

frohro:
Roland,
I was using your work this morning making figures for python notebooks, and snapping moved objects to the grid is just fantastic.  Thanks for all the time you put into this.  It really helps make xournalpp great for my applications.  I can now copy and paste repeated parts of the figure easily, and they end up where I want them.  It was so nice!
Thanks!
Rob

frohro:
I've been waiting for this to be accepted since May.  In the meantime, I have been using @rolandlo's fork that has these changes.  I have been using it a lot.  Just today, I used the ability to select a picture and move it over a certain number of squares in my Signals & Systems class.  it was great!  I hope it doesn't molder until merging it is a problem.  Operationally, it is great!

Febbe:
Thanks frohro for the reminder. @rolandlo, you have a better overview, what's missing in your repository before we merge it. ( You should squash some commits together, but what else is open?)

rolandlo:
The only real thing that comes to mind: The snapping for recognized shapes may need to be made optional. I often have the shape recognizer (and snapping to grid) turned on even if I just write some text, so that lines are straight if I underline a title (or something else). Then I don't want this line to be snapped to grid.

Maybe rotation snapping could be made for 15 degree steps (currently it is 45 degrees). That would allow for easy creation of equilateral triangles and things like that. I am not sure whether that should be made configurable in the settings or not. It can wait for a follow-up PR as well.

Otherwise I don't see why it shouldn't be merged. Seems @frohro  has done quite a bit of testing and it works as expected.

So @Febbe, if you agree, I will make snapping for recognized settings optional by adding a checkbox in the drawing area panel in the settings, and then squash commits together.

Febbe:
Maybe a switch between offset snapping as extension, instead of snapping to the grid. That would allow to start shapes exact at the pen position, and the snapping is relative to the first point. If the position is bad someone can still snap it to the grid by moving the object.

rolandlo:
That offset switch idea sounds like an interesting suggestion, but most likely not trivial to implement and requiring decent testing. I would prefer to leave that for a follow-up PR and finish this PR up rather sooner than later.

rolandlo:
I have made the changes I wanted (configurable snapping for shape-recognizer, 15 degree rotation snapping). Some (minor) merge conflicts have to be solved. Seems I have messed up with git trying to squash the commits and solve the merge conflicts.

Febbe (approved):

src/model/Snapping.cpp
```diff
@@ -6,15 +6,25 @@ namespace Snapping {
...
+double snapVertically(double y, double gridSize, double tolerance) {
+    double ySnapped = roundToMultiple(y, gridSize);
+    return abs(ySnapped - y) < tolerance * gridSize ? ySnapped : y;
+}
+
+double snapHorizontally(double x, double gridSize, double tolerance) {
+    double xSnapped = roundToMultiple(x, gridSize);
+    return abs(xSnapped - x) < tolerance * gridSize ? xSnapped : x;
+}
+
```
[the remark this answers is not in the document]
rolandlo: I know, but I think these methods will diverge in future if we introduce snapping to other grids. Should I still make 1 function out of it?

rolandlo:
Seems I have managed to squash and rebase on current master, resolving all merge conflicts.

Febbe:
Ok can you squash the last commit. And then you can merge it.

rolandlo:
Thanks a lot Febbe for taking the time to look into it. Wonderful that this is finished now.

PR #2014, opened 2020-05-27
Fix MacOS 10.13 compatibility issues (build and CI)

Technius:
This PR fixes the MacOS builds. Although the issues have actually been present since the forced move to MacOS 10.15 due to the Azure Pipelines update, they had not been exposed until we also updated the regular CI runner to check backwards compatibility with 10.13.

* Switch from Boost filesystem to https://github.com/gulrak/filesystem, which is mostly compliant with C++17 filesystem. Note that the supplementary filesystem library is only used if `std::filesystem` is not supported. The new library will be downloaded automatically by CMake at compile time.
* Fix some backwards compatibility issues.

### DIFF
--- a/cmake/find/FindFilesystem.cmake
+++ b/cmake/find/FindFilesystem.cmake
@@ -1,6 +1,11 @@
 # Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
 # file Copyright.txt or https://cmake.org/licensing for details.
 
+# Original file from CMake Community Modules by vector-of-bool
+# (https://github.com/vector-of-bool/CMakeCM)
+#
+# With modifications by the Xournal++ Team
+
 #[=======================================================================[.rst:
 
 FindFilesystem
@@ -105,6 +110,7 @@ endif ()
 include(CMakePushCheckState)
 include(CheckIncludeFileCXX)
 include(CheckCXXSourceCompiles)
+include(ExternalProject)
 
 cmake_push_check_state()
 
@@ -121,7 +127,7 @@ endif ()
 
 # Warn on any unrecognized components
 set(extra_components ${want_components})
-list(REMOVE_ITEM extra_components Final Experimental Boost)
+list(REMOVE_ITEM extra_components Final Experimental Boost ghc)
 foreach (component IN LISTS extra_components)
     message(WARNING "Extraneous find_package component for Filesystem: ${component}")
 endforeach ()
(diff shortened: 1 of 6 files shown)


cmake/find/FindFilesystem.cmake
```diff
@@ -1,6 +1,9 @@
 # Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
 # file Copyright.txt or https://cmake.org/licensing for details.
 
+# Original file from Kitware CMake
+# With modifications by Bryan Tan
+
```
Febbe, line 6: Add my name or replace by Xournal++ Team, since I added the boost stuff.

cmake/find/FindFilesystem.cmake
```diff
@@ -236,13 +240,40 @@ if (NOT _found AND "Boost" IN_LIST want_components)
...
-        set(CXX_FILESYSTEM_HEADER boost/filesystem.hpp CACHE STRING "The header that should be included to obtain the filesystem APIs")
-        set(CXX_FILESYSTEM_NAMESPACE boost::filesystem CACHE STRING "The C++ namespace that contains the filesystem APIs")
+        # Bug fix by Bryan Tan
+        set(CXX_FILESYSTEM_HEADER boost/filesystem.hpp CACHE STRING "The header that should be included to obtain the filesystem APIs" FORCE)
+        set(CXX_FILESYSTEM_NAMESPACE boost::filesystem CACHE STRING "The C++ namespace that contains the filesystem APIs" FORCE)
+        # End bug fix
         add_library(std_filesystem INTERFACE)
         target_compile_features(std_filesystem INTERFACE cxx_std_17)
         target_link_libraries(std_filesystem INTERFACE Boost::boost Boost::filesystem)
+        target_include_directories(std_filesystem INTERFACE ${Boost_INCLUDE_DIRS})
```
Febbe, line 250: This line should be obsolete, since it is part of the target Boost::filesystem and Boost::boost. Or is this not working, since we have an Interface of an Interface library?

LittleHuba:
As you modified the pipelines already please add one more change:
- Split up the Release Pipeline to three individual pipelines all doing the same job as the current Release pipeline but only for one OS each.

When you added this and the PR is merged, we can make the release process more agnostic to errors in one OS. Meaning we get nightlies even when one OS fails. The Release will then just reuse the old version of the one where the build of the current version failed. When you merge this PR please notify me in advance so I can edit the release pipeline.

cmake/find/FindFilesystem.cmake
```diff
@@ -236,13 +240,40 @@ if (NOT _found AND "Boost" IN_LIST want_components)
...
+    set(CXX_FILESYSTEM_NAMESPACE ghc::filesystem CACHE STRING "The C++ namespace that contains the filesystem APIs" FORCE)
+
+    add_library(std_filesystem INTERFACE)
+    add_dependencies(std_filesystem ghcFilesystem_git)
+    target_compile_features(std_filesystem INTERFACE cxx_std_17)
+    ExternalProject_Get_Property(${GHC_FILESYSTEM} SOURCE_DIR)
+    file(MAKE_DIRECTORY "${SOURCE_DIR}/include")
+    target_include_directories(std_filesystem INTERFACE "${SOURCE_DIR}/include")
+
+    add_library(std::filesystem ALIAS std_filesystem)
```
Febbe, line 276: I think this could be simplified, since ExternalProject_Add creates a target, containing all dependencies.

src/util/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ unset (util_SOURCES)
 file (GLOB_RECURSE util_SOURCES *.cpp)
 
 add_library (util STATIC ${util_SOURCES})
-target_link_libraries(util PUBLIC std::filesystem cxx17)
+target_link_libraries(util PUBLIC cxx17)
```
Febbe, line 6: I would leave this here, since we need it there anyways when switching completely to std::filesystem.

cmake/find/FindFilesystem.cmake
Technius: `ExternalProject_Add` only builds and installs the subproject; it does _not_ import the targets of the subproject. The preferred option is [`FetchContent`](https://cmake.org/cmake/help/git-stage/module/FetchContent.html) but it is only available on CMake 3.11+ (we unfortunately require 3.10).

cmake/find/FindFilesystem.cmake
Technius: I will have to check this later, because my first attempt at making it compile with Boost needed this line.

Technius: Turns out to be unnecessary.

Technius:
Rebased on master and addressed @Febbe's comments.

> As you modified the pipelines already please add one more change:
> 
>     * Split up the Release Pipeline to three individual pipelines all doing the same job as the current Release pipeline but only for one OS each.
> 
> 
> When you added this and the PR is merged, we can make the release process more agnostic to errors in one OS. Meaning we get nightlies even when one OS fails. The Release will then just reuse the old version of the one where the build of the current version failed. When you merge this PR please notify me in advance so I can edit the release pipeline.

The only problem I see with this is that we want the entire build to fail for full releases. This makes releases very complicated. I wonder if we can get Azure Pipelines to just ignore the build failure...there should be a way to do this now that the deployment pipelines can be specified in yaml.

LittleHuba:
We will just have to separate those pipelines as well. For the nightly we can continue if one should fail, for the release one we stop.

The problem with the merged pipeline is, that we already do continue building for the other OSs but the build is marked as fail. The Release pipeline itself takes the last successful build (which is the oldest one where all managed to work)

PR #2032, opened 2020-06-01
Fix Adding Pages wont scroll to new page

### LINKED ISSUE #2016: Adding pages doesn't bring me to the new page
**Affects versions :**
 - OS: Linux elementaryOS Hera 5.1
     - 3.22.30
 - Version of Xournal++: [e.g. 1.0.16, specific commit, specific branch]
     - 1.1.0+dev
 - Installation method: PPA

**Describe the bug**
Adding new pages doesn't bring me to the newly added pages

**To Reproduce**
1. Make a new document
2. Add many pages
3. Delete all of them
4. Repeat step 2

**Expected behavior**
Adding pages bring me to the newly added pages

**Screenshots of Problem**
Screen record from 2020-05-28 12 24 25

Febbe:
Should fix #2016
Todo inherited from https://developer.gnome.org/gtk3/stable/GtkScrollable.html#GtkScrollablePolicy :

> All scrollable widgets should do the following:
 - [ ] When a parent widget sets the scrollable child widget’s adjustments, the widget should populate the adjustments’ “lower”, “upper”, “step-increment”, “page-increment” and “page-size” properties and connect to the “value-changed” signal.
>
- [x] Because its preferred size is the size for a fully expanded widget, the scrollable widget must be able to cope with underallocations. This means that it must accept any value passed to its GtkWidgetClass.size_allocate() function.
>
- [ ] When the parent allocates space to the scrollable child widget, the widget should update the adjustments’ properties with new values.
>
- [x] When any of the adjustments emits the “value-changed” signal, the scrollable widget should scroll its contents.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1229,7 +1229,7 @@ void Control::insertPage(const PageRef& page, size_t position) {
     int visibleHeight = 0;
     scrollHandler->isPageVisible(position, &visibleHeight);
 
-    if (visibleHeight < 10) {
+    if (visibleHeight < 100) {
         Util::execInUiThread([=]() { scrollHandler->scrollToPage(position); });
     }
     firePageSelected(position);
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -567,7 +567,10 @@ void XournalView::pageInserted(size_t page) {
     viewPages.insert(begin(viewPages) + page, pageView);
 
     Layout* layout = gtk_xournal_get_layout(this->widget);
+    GtkAllocation allocation;
+    gtk_widget_get_allocation(this->widget, &allocation);
     layout->recalculate();
+    layout->layoutPages(allocation.width, allocation.height);
     layout->updateVisibility();
 }
 


src/control/Control.cpp
```diff
@@ -1246,7 +1246,7 @@ void Control::insertPage(const PageRef& page, size_t position) {
     int visibleHeight = 0;
     scrollHandler->isPageVisible(position, &visibleHeight);
 
-    if (visibleHeight < 10) {
+    if (visibleHeight < 100) {
```
Technius, line 1249: What does this constant do?

Febbe: When 100pt are visible don't scroll. Its the height of the visible area of the page, not zoomed. I am not sure why we have that behaviour.

LittleHuba: I assume this prevents scrolling while pages are not yet rendered correctly. But considering that we actually know the size of the page from the start this does not really make sense. Would be interesting when this was introduced. Might give a hint why it is required.

Febbe:
The problem is wide bigger than expected:
The zoom to the new page requires, that the actual GtkAlignment is set but this requires that the page is exposed once. But before that the Layout needs to be recalculated. So the Widget can be exposed to the correct size. Currently it does not work to emplace the realignment after the expose. Also overriding the alignment does not work. 
We need to rewrite the scrolling and the layouting so, that the following jobs can be executed in the specific order:

add page > recalculate layout and bounds > expose > reallign > redraw .

Technius:
That's right, see the original rationale for the viewport rewrite #1795.

rolandlo:
I guess this PR can be closed, since PR #2346 does the job.

PR #2055, opened 2020-06-09
Set a darker graph color when a black background is used

robinp7720:
Using a black background with light gray lines in the graph view is quiet jarring. This will change the graph color to a darker color when a black background is set.
Black graph background


src/view/background/GraphBackgroundPainter.cpp
```diff
@@ -8,6 +8,21 @@ GraphBackgroundPainter::GraphBackgroundPainter() = default;
...
+void GraphBackgroundPainter::updateGraphColor() {
+    unsigned int backgroundColor = this->page->getBackgroundColor();
+
+    // The background color returned by getBackgroundColor is sometimes offset by 0xff000000
+    if (backgroundColor == 0x000000 || backgroundColor == 0xff000000) {
+        this->foregroundColor1 = 0x202020;
+        return;
+    }
+
+    this->foregroundColor1 = 0xBDBDBD;
```
Febbe, line 23: This only works for color 0x0. It would be better to check for the distance.
```suggestion
   auto greyscale = [](uint32_t color){
       return ((0xff & color) + (0xff & (color >> 8)) + (0xff & (color >> 16))) / 3;
   };

this->foregroundColor1 = std::abs(int32_t(greyscale(backgroundColor)) - int32_t(greyscale(this->foregroundColor1))) < 30 ? 0x202020 : 0xBDBDBD;
```

Issue #2056, opened 2020-06-10
Performance problems


Krzmbrzl:
**Affects versions :**
 - OS: KDE Neon (Linux)
 - (Linux only) Desktop environment:  X11
 - Which version of libgtk do you use
     - 3.22.30
 - Version of Xournal++: 1.1.0+dev
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Installation method: PPA

**Describe the bug**
Not really sure whether that counts as a bug but in the last couple of weeks (probably about a month now) I have seen a (perceived) drop in performance or rather an increase in CPU consumption when writing.

When I first started using this application, everything worked just fine but right now it seems to get worse and worse. When I am running a screen recorder alongside Xournal++, I'm getting a lack between the stylus input and the line being drawn of several seconds.
My CPU usage is also pretty high in these cases (90% or higher). I didn't switch the recording program and it used to work just fine (on the same laptop).

Did you change something in the past moth or so that might be related?
(I used the snapshot PPA from the beginning)

Krzmbrzl:
I can confirm that the performance problems seem to have vanished in the current master branch. Thank you! :)

Issue #2003, opened 2020-05-21
[Regression] Erasing with Pen Button isn't directly executed


jaykijay:
**Affects versions :**
 - OS: Archlinux
 - KDE
 - Which version of libgtk do you use: 1:3.24.20-1
 - Version of Xournal++: 1.1.0+ (a26797b4)
 - Compiled from AUR

**Describe the bug**
In the last Version i used (like 2Months old) everything worked fine. 
Now when i set the first pen button to be a eraser (erase whole lines) it does not erase it directly but just after i stopped pressing the Button. This is very uncomfortable.
If i choose the eraser tool manually it works perfectly fine..
I made a small gif to show the issue: First step: Erasing with Button, second Step: Erasing with manually choosing the tool.

**To Reproduce**
1. Configure Pen Button 1 to be a eraser
2. Write something
3. Press the button to delete it

**Expected behavior**
It should be deleted directly while still pressing the button. As it was before.

**Screenshots of Problem**

![Peek 2020-05-21 13-55](https://user-images.githubusercontent.com/24495677/82560936-ad0b5700-9b72-11ea-9c1b-07c2dde9e602.gif)

4VRDriver:
I can confirm this issue, but it only appears when the new input system is enabled. Try to disable it under Settings > Input system.

In my case the behaviour is slightly different: If the new input system is enabled, no strokes will be erased whatsoever even after letting the button go.

Technius:
I can't reproduce this on my computer. Can one of you run `git bisect` to find the bad commit?

4VRDriver:
I did `git bisect` and in my case it narrows down to commit `baabf791a21aadbec9344558565ab8ad6ee26fb7`, so around when the new input system was added.  (so bisect wasn't a big help I guess :P)

But I have found something else that could be very interesting: There is a setting in the KDE wacom settings called "Tap to execute action". (where my cursor is)
![Screenshot_20200522_140938](https://user-images.githubusercontent.com/44267643/82666643-9d554680-9c36-11ea-93d9-b4c902b9be2c.png)

If this is selected and the old input system is enabled it behaves just as expected. But enabling the new input system causes to not to be able to erase strokes whatsoever by pressing the stylus button.

If this checkbox is not selected and the old input system is enabled, strokes get erased when the stylus isn't even touching the tablet. By enabling the new input system however, the eraser behaves now as expected (erasing only when the stylus is touching the tablet and by pressing the button).

[...]

PR #2057, opened 2020-06-10
fix bug #2056

rolandlo:
This commit should fix bug #2056 (and possibly bug #2003) while still updating the cursor after the eraser (or pen/highlighter) size changes. Check it out!

### DIFF
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -309,7 +309,7 @@ void XournalppCursor::updateCursor() {
 auto XournalppCursor::getEraserCursor() -> GdkCursor* {
 
     // Eraser's size follow a quadratic increment, so the cursor will do the same
-    double cursorSize = control->getToolHandler()->getThickness() * 2 * control->getZoomControl()->getZoom();
+    double cursorSize = control->getToolHandler()->getThickness() * 2.0 * control->getZoomControl()->getZoom();
     gulong flavour = static_cast<gulong>(64 * cursorSize);
 
     if (CRSR_ERASER == this->currentCursor && flavour == this->currentCursorFlavour) {
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -307,10 +307,17 @@ void XournalppCursor::updateCursor() {
 
 
 auto XournalppCursor::getEraserCursor() -> GdkCursor* {
-    this->currentCursor = CRSR_ERASER;
 
     // Eraser's size follow a quadratic increment, so the cursor will do the same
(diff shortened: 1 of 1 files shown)


Febbe:
Looks also good to me. But since @idotobi introduced this with a bugfix, he should check, if his bugfix still works as expected.

src/gui/XournalppCursor.cpp
```diff
@@ -307,10 +307,17 @@ void XournalppCursor::updateCursor() {
 
 
 auto XournalppCursor::getEraserCursor() -> GdkCursor* {
-    this->currentCursor = CRSR_ERASER;
 
     // Eraser's size follow a quadratic increment, so the cursor will do the same
     double cursorSize = control->getToolHandler()->getThickness() * 2 * control->getZoomControl()->getZoom();
```
LittleHuba, line 312: Please change this to `2.0`.

idotobi:
> Looks also good to me. But since @idotobi introduced this with a bugfix, he should check, if his bugfix still works as expected.

I will only be able to check this PR next week. If it's pressing you can go ahead and merge it already. I wouldn't expect it to break my bugfix. And if, it would be quite apparent (cursor size would not update).

src/gui/XournalppCursor.cpp
rolandlo: @LittleHuba Could you please explain me, (if and) why that makes any difference? As I understand, the int 2 will be casted to a double implicitly. Should I always write integer values with a decimal point when it is multiplied with (or added to/subtracted from, divided by) a double?

Technius: This is just a maintenance and readability thing. It's pretty annoying to have to chase down all of the type definitions just to figure out if/when there are casts. If we happen to change the type signature (which is possible, but unlikely), we can immediately avoid buggy implicit conversions.

LittleHuba: @Technius not necessarily...

What you will get is not as easy as "it will be cast to double". Depending on where the `int` is, the exact opposite can happen and you end up with casting all the operands to `int` and then multiply them to finally cast them back to `double` for storage. 

Additionally you might introduce overflows if you use `int` and `long` within the same formula. 

Generally stick with the biggest type (`int + long -> long`) and the most precise (`int + float -> float`) to not run in these issues. If you need to cast, do so after you did all the arithmetics and only need to store the result.

Technius: Right, which is why we want to either make every cast explicit or eliminate casting altogether. Rounding errors are easily exposed when everything is explicit.

Technius:
Merging, thanks for the PR!

PR #2060, opened 2020-06-11
Add Ubuntu 20.04 to release builds

Technius:
Adds Ubuntu 20.04 to release builds. This should help discoverability and resolve a lot of installation problems.

Also removes the `condition: always()` since the only thing that it achieves is preventing build cancellations.

After merging, this needs a change in the deployment pipeline in the renaming step.

```diff
- for f in linux-ubuntu-packages/xournalpp-* linux-debian-packages/xournalpp-*; do
+ for f in linux-ubuntu-*-packages/xournalpp-* linux-debian-packages/xournalpp-*; do
```

### DIFF
--- a/azure-pipelines/jobs/release_ubuntu.yml
+++ b/azure-pipelines/jobs/release_ubuntu.yml
@@ -1,0 +1,37 @@
+# Template for running Ubuntu release jobs
+parameters:
+  - name: ubuntu_version
+    type: string
+  - name: job_name
+    type: string
+  - name: build_appimage
+    type: boolean
+    default: false
+
+jobs:
+  - job: '${{ parameters.job_name }}'
+    pool:
+      vmImage: 'ubuntu-${{ parameters.ubuntu_version }}'
+    displayName: 'Build for Ubuntu ${{ parameters.ubuntu_version }}'
+    steps:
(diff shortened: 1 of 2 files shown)


LittleHuba:
> Also removes the condition: always() since the only thing that it achieves is preventing build cancellations.

I introduced this to see if other OSs may run through even if one fails. But as we want to separate those pipelines anyway that is obsolete now.

For testing we will need to merge this first into a new branch on the main repository and then do a manual run of the release pipeline. This will not trigger an actual release since we only trigger those if built from the master-branch.

Technius:
> For testing we will need to merge this first into a new branch on the main repository and then do a manual run of the release pipeline. This will not trigger an actual release since we only trigger those if built from the master-branch.

Oops, I forgot to mention that I already tested this in a separate Azure Pipelines setup. We could test it in another branch if you want to be extra sure.

LittleHuba:
Nah if you tested it already I'm good. You can go ahead an merge it then. Please make the release change as well so the next nightly doesn't break. You've got around 14 hours to do so. Rebuild is at midnight german time.

Technius:
Missed the window, but I guess I'll just merge and see if it works in 22 hours.

Issue #2037, opened 2020-06-04
I want to write a plugin


createyourpersonalaccount:
**Is your feature request related to a problem? Please describe.**
I write math documents in LaTeX and often I want to include images. I don't like having to "code" the images (with tikz, asymptote, gnuplot, etc). I would much rather be able to draw them; this is what I currently do in xournal++, but the exporting is not as streamlined as could be.

**Describe the solution you'd like**
I would like to write a plugin that makes this easier. It would work like this: You choose an area with the area selector and click "export to png"; it should automatically be placed in project/figures/N.png where N is an incrementing number (or some other such schema).

I have programming experience in C, C++ and other languages (hobbyist), but I have not contributed to any projects before. I was wondering if someone could help me with the relevant information: How to write a plugin for xournal++, and whether the idea I have in mind is feasible for someone like me.

Thank you for your time, and this awesome project!

rolandlo:
@createyourpersonalaccount 

That should be rather simple, I think, although it may depend on the system you use. With a plugin you can execute system commands. Use `gnome-screenshot` or `scrot` to take a screenshot of a user-selected area and save it into a file. 
Your plugin will have an .ini file (see any of the existing plugins in the source code) and a lua-scriptfile. The latter (using `scrot`) would look about like this:

```
-- Register all Toolbar actions and intialize all UI stuff
function initUi()
  app.registerUi({["menu"] = "Take and save Screenshot", ["callback"] = "shot"});
end

function shot()
  os.execute("scrot -s '%Y-%m-%d-%T.png' -e 'mv $f ~/project/figures/' ")
end
```
Explanation: 

- scrot is a screenshot tool (under Linux)
- the option -s lets you specify the area
- the string '%Y-%m-%d-%T.png' indicates the filename defined by year (Y), month (m), day(d) and time of the day (T)
- the option -e executes the given command
- mv moves the file
- $f is the scrot-internal filename
- ~/project/figures is the path you want your file to be stored

If you want a more sophisticated file naming convention, you can use some lua scripting to produce it.

@Spiriz00 You can combine the screenshot-tool `scrot` with the convert-tool `convert` to get a pdf image

createyourpersonalaccount:
@rolandlo Thank you, I will look into this. I am glad (and a bit surprised!) that @Spiriz00 already showed support for the idea.

I will try to make this happen within the week.

Spiriz00:
[...]

PR #2086, opened 2020-06-24
QuickSnapshot plugin added to xournal++.

createyourpersonalaccount:
This is the plugin promised in #2037

### DIFF
--- a/plugins/QuickScreenshot/main.lua
+++ b/plugins/QuickScreenshot/main.lua
@@ -1,0 +1,112 @@
+-- The QuickScreenshot plugin:
+--
+-- Allows the user to save a region to disk as png image.
+-- To be used so that users can save various diagrams, etc
+-- to their directory of choice; useful when creating a lot
+-- of diagrams for presentation.
+--
+-- TODO:
+--  o Support for Windows and macOS is missing. To add support
+--    you must edit the functions 'existsUtility' and 'go' to
+--    use the right system programs.
+--
+
+-- Returns 'true' if the system utility exists;
+-- returns 'false' otherwise. This is a dangerous function,
+-- since the argument to popen is not sanitized.
+-- The programmer must ensure that the argument to
+-- existsUtility() is sanitized (not malicious).
+function existsUtility(utility)
+  if operatingSystem == "Windows" then
+    -- Windows not supported yet.
+  elseif operatingSystem == "Darwin" then
+    -- macOS not supported yet.
+  else
+    local f = assert(io.popen("command -v " .. utility))
+    local s = assert(f:read('*a'))
+    f:close()
+    return s ~= ""
+  end
+end
+
+-- Return one of "Windows", "Linux", "Darwin", "FreeBSD",
+-- "OpenBSD" or "NetBSD", etc.
+function findOS()
+  if package.config:sub(1,1) == '\\' then
+    return "Windows"
+  else
+    local f = assert(io.popen("uname -s", 'r'))
+    local s = assert(f:read('*a'))
+    f:close()
+    return s
+  end
+end
+
+-- Register all Toolbar actions and intialize all UI stuff
+function initUi()
+  ref = app.registerUi({["menu"] = "QuickScreenshot", ["callback"] = "go", ["accelerator"] = "<Shift><Alt>t"});
+  operatingSystem = findOS() -- What's the operating system?
+end
+
+-- This is the callback that gets executed when the user
+-- activates the plugin via the menu or hotkey.
+--
+-- This function calls 'app.saveAs' to create
+-- a "Save As" dialog for the user to choose the
+-- filename where the image should reside. Then
+-- it executes a screenshot utility to capture the
+-- region the user selects.
+function go()
+  local windowsUtilities = {} -- list of windows programs here
+  local macUtilities = {} -- list of macOS programs here
+  local elseUtilities = -- list of programs for other systems
+    { "scrot --overwrite --select --silent "
+-- gnome-screenshot behaves oddly (pollutes the window)
+-- the line is added here to be indicative of what
+-- adding another utility looks like.
+--
+--  , "gnome-screenshot --area --file="
+    }
+
+  -- Different operating systems have different
+  -- screenshot utilities, so we check the OS here
+  if operatingSystem == "Windows" then
+    app.msgbox("Windows not supported yet.", {[1] = "OK"})
+  elseif operatingSystem == "Darwin" then
+    app.msgbox("macOS not supported yet.", {[1] = "OK"})
+  else
+    -- This becomes true if at least one screenshot
+    -- utility is available on the system
+    local foundUtility = false
+    -- Check utility availability and use it
+    for i,command in ipairs(elseUtilities) do
+      utilityName = command:match("%S+")
+      if existsUtility(utilityName) then
+        local tmpFilename = os.tmpname()
+        local runCommand = assert(io.popen(command .. tmpFilename .. " &"))
+        runCommand:read('*all')
+        runCommand:close()
+
+        -- Launch the "Save As" dialog
+        local filename = app.saveAs()
+        if not filename then
+          os.remove(tmpFilename)
+          return
+        end
+
+        os.rename(tmpFilename, filename)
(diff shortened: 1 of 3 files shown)


plugins/QuickScreenshot/main.lua
```diff
@@ -0,0 +1,115 @@
...
+  else
+    -- This becomes true if at least one screenshot
+    -- utility is available on the system
+    local foundUtility = false
+    -- Check utility availability and use it
+    for i,command in ipairs(elseUtilities) do
+      utilityName = command:match("%S+")
+      if existsUtility(utilityName) then
+        local tmpFilename = os.tmpname()
+        os.execute(command .. tmpFilename .. " &")
```
rolandlo, line 98: The following code change will make your plugin wait for the user to finish the area selection:

```suggestion
        local file = assert(io.popen(command .. tmpFilename .. " &"))
        file:read('*all')
        file:close()
```
Make this change and launch the "Save as" dialog between this code block and the renaming of the file. That should solve your problem.

createyourpersonalaccount: @rolandlo I tried the suggested code changes, but the issue persists, see the attached image. I think this has to do with the single-thread lua process stalling the xournal++ redrawing. 
Untitled

rolandlo: Hmm, strange, that does not happen on my computer. I have no "pollution" at all, it just works fine. Just to make clear: Have you moved the "Save as dialog" launching right in front of renaming the file? E.g. like this?

```
    for i,command in ipairs(elseUtilities) do
      utilityName = command:match("%S+")
      if existsUtility(utilityName) then
        local tmpFilename = os.tmpname()
        local file = assert(io.popen(command .. tmpFilename .. " &"))
        file:read('*all')
        file:close()
          -- Launch the "Save As" dialog 
        local filename = app.saveAs()
        if not filename then
          return
        end
	      os.rename(tmpFilename, filename)
	      foundUtility = true
	      break
      end
    end

```
More questions:
- Does the pollution happen with both gnome-screenshot and scrot? 
- Can you try taking a screenshot of some area which includes parts outside of the Xournal++ window? So do the screenshot tools pollute parts outside of Xournal++ as well?

createyourpersonalaccount: This does not happen with scrot! The pollution does not happen outside the Xournal++ window. I think scrot does the trick. I will remove gnome-screenshot from the lua code, since that behavior is not acceptable.

createyourpersonalaccount:
The security issue has been dealt with by using a temporary filename and the Lua function `os.rename()` which is safe. The pollution issue discussed above was addressed by using `scrot` instead of `gnome-screenshot`.

rolandlo:
Great that scrot works for you without pollution as well!
Further changes will be required, though:

1. The plugin does not work currently. When I run the plugin, an empty temporary file is generated in /tmp. After choosing the selection, the screenshot is not saved in this file, but in a file with _000 appended to the filename. When the file dialog is closed, the original tmp-file is moved (which is empty) to the correct place.
2. You should choose a different shortcut than Ctrl-t, since that one is already used by the TestPlugin. See [the GTK3 reference](https://developer.gnome.org/gtk3/stable/gtk3-Keyboard-Accelerators.html#gtk-accelerator-parse) for how to format a more sophisticated shortcut like <Ctrl><Shift>1.
3. Apply Clang format.
4. Make sure your lua code uses correct indentation.

createyourpersonalaccount:
I don't understand how to use ~git-clang-format~... I tied `git add .` followed by the command but I get the response: ~no modified files to format~. Am I doing something wrong? The modified files are `main.lua` (which I don't expect clang to work on) and ~src/plugin/luapi_application.h~.

rolandlo:
For clang formatting the header file (or any cpp file) you may just execute 
`clang-format-9 -i --style=file ./src/plugin/luapi_application.h` in your xournalpp folder. See several other options discussed in #1929.
For lua files: all spaces, 2 per indentation (compare e.g. xournalpp/plugins/ToggleGrid/togglegrid.lua)

createyourpersonalaccount:
Thank you @rolandlo. I tried to fix the issues you brought up. I am using <Shift><Alt>t instead. Unfortunately the commit log has too many commits. I am not very good with git and I messed up in that way, I hope that's okay.

rolandlo:
@createyourpersonalaccount Once I have checked that everything is OK, I can "squash and merge" from within github, if you want me to. Otherwise I would ask you to do it yourself by following [this tutorial](https://www.internalpointers.com/post/squash-commits-into-one-git).

createyourpersonalaccount:
Sounds good, I'll do the squash then!

rolandlo:
The plugin works very nicely (on Linux) now. The screenshot is now saved correctly. I have checked that after uninstalling scrot I get the error message printed to console. 
I have also tested the plugin on Windows and expected to see the "Windows not supported yet" error message. However even running Xournal++ from the console there is no console output in case Xournal++ was not built from source. Most Xournal++ users on windows will not build from source, so they will not see an error message and get confused by it. 
Moreover I found some lines in your main.lua files where you indented by 3 spaces instead of 2. Therefore I ask:

- please change the error messages on line 73 and 75 of main.lua so that they use the app.msgbox function (c.f. the Example plugin, which also uses that function). 
- fix the indentation in main.lua (line numbers 34, 93-94), 

Afterwards I think the plugin can be merged.

createyourpersonalaccount:
Okay, will do. I think on windows the executable to run is `%WINDIR%\System32\SnippingTool.exe`. I can't test it unfortunately but I will add it as a potential candidate in the comments.

rolandlo:
With SnippingTool.exe you can select an area using the /clip option. But how can you save it to a specified file? Not sure it can do that.

createyourpersonalaccount:
Ah, thank you for checking. I could not check because I had no Windows access and I can't find the manual online. 

Eventually it would be nice to replace the system invocations with Xournal++ code, but for now to add Windows support one option would be to use winAPI to take the screenshot.

rolandlo:
> Eventually it would be nice to replace the system invocations with Xournal++ code, but for now to add Windows support one option would be to use winAPI to take the screenshot.

The question is why would you make this (coding and code maintaining) effort, if you can also just run your favorite screencapture tool via a shortcut and use its GUI to save the file. It is only a question of a couple of clicks you can save, not more, unless you add more functionality or make it communicate with Xournal++ (like pasting the screenshot into the Xournal++ document).

createyourpersonalaccount:
You are right, it's not very important. I'm happy enough with this solution which is what I am going to be using from now on :)

plugins/QuickScreenshot/main.lua
```diff
@@ -70,8 +71,10 @@ function go()
   -- Different operating systems have different
   -- screenshot utilities, so we check the OS here
   if operatingSystem == "Windows" then
+    app.msgbox("Windows not supported yet.", {})
```
rolandlo, line 74: You need one button, otherwise you cannot close the msgbox.
```suggestion
    app.msgbox("Windows not supported yet.", {[1] = "OK"})
```

plugins/QuickScreenshot/main.lua
```diff
@@ -70,8 +71,10 @@ function go()
...
     print("Windows not supported yet.")
   elseif operatingSystem == "Darwin" then
+    app.msgbox("macOS not supported yet.", {})
```
rolandlo, line 77: Same here.

rolandlo: I do not think printing the error message to the console is necessary if the same message is already communicated via the msgbox. But that is a minor thing.

createyourpersonalaccount: Fixed!

createyourpersonalaccount:
I just realized that I can commit the suggestions you make! Is that important in terms of attribution?

rolandlo:
It does make a difference in terms of attribution, but it is not important for me.

Everything looks fine now. Can you please squash your commits into one once more? I will merge the PR then (unless you tell me that you still want to change something).

createyourpersonalaccount:
Sure! Thanks, this was a positive experience.

PR #2087, opened 2020-06-27
Move config directory to .config/xournalpp

### LINKED ISSUE #1999: Strange (new) location of config files?
**Affects versions :**
 - Linux Ubuntu 18.04, Gnome-X11
 - Version of Xournal++: a26797b4aed
 - Installation method: compiled by source
 - Run method: in the build directory, using `./src/xournalpp` to check before installing

**Describe the bug**

The application informed me that the config files were moved (good and nice!) to `.config/com.github.xournalpp.xournalpp` (?).

This could be ok, but all the other applications I have installed use a `.config/app-name` directory, like `.config/xournalpp`.

**To Reproduce**

1. Compile
2. Execute `./src/xournalpp`

**Expected behavior**

All the other application follow the `.config/xournalpp` style of config directory

**Screenshots of Problem**

y2kmurali:
Really short PR to resolve #1999 

Tested by running the app locally, got the dialog that the config directory was moved to ~/.config/xournalpp, and verified that any settings changes were reflected in this new config file location and not the old one.

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -290,7 +290,7 @@ auto XournalMain::exportPdf(const char* input, const char* output) -> int {
 }
 
 auto XournalMain::run(int argc, char* argv[]) -> int {
-    g_set_prgname("com.github.xournalpp.xournalpp");
+    g_set_prgname("xournalpp");
     this->initLocalisation();
     MigrateResult migrateResult = this->migrateSettings();
 


rolandlo:
@y2kmurali Testing your PR I got an error:

```
** (xournalpp:241333): WARNING **: 15:59:31.476: Failed to load "colornames.ini" (/home/roland/.config/xournalpp//colornames.ini): Datei oder Verzeichnis nicht gefunden


** (xournalpp:241333): WARNING **: 15:59:31.488: configfile does not exist /home/roland/.config/xournalpp//settings.xml

```
Note the two slashes before the filenames. Any idea what went wrong?

y2kmurali:
I'm honestly not sure, I don't receive that message, but that might just be because I happened to develop this on a fresh install of the app :confused: 

I'm guessing somehow an extra '/' must've gotten into the config path, what's the output of `getConfigFolder()`? Mine's just /home/muralim/.config/xournalpp which is why my computer isn't complaining about the paths having a double slash.

aDogCalledSpot:
`getConfigFolder()` returns ~/.config/xournalpp as expected but then doing `~/.config/xournalpp /= ""` returns `~/.config/xournalpp/` (the empty string is the subfolder of `colornames.ini` or `settings.xml`). Then doing ` p /= "colornames.ini"` returns `"~/.config/xournalpp//colornames.ini".

I honestly don't think though that this is an issue though. Double-slashes are allowed in file paths and it happens frequently when using environment variables. I don't understand why glib doesn't support this as `fopen` allows double-slashes without any issues.

We could fix this here quickly by adjusting the `/=` operator to also only append a separator if the path which should be appended is also non-empty.

rolandlo:
@aDogCalledSpot Thanks for the analysis. I also think we need to adjust the /= operator. If we look at the code
https://github.com/xournalpp/xournalpp/blob/0313ce563947980598149b33dfbbf3c7acfd4edd/src/util/Path.cpp#L26-L44

we see that the two methods behave differently. The second one only appends the separator, if there isn't one already, whilst the first one ignores that problem. So the first method should be fixed to match the behavior of the second method.

@y2kmurali Can you do this?
@aDogCalledSpot Are you fine with this way of fixing the problem as well?

Febbe:
@rolandlo bevor you merge this, can you make shure, that the user gets a warning and the directory is copied/renamed also. Just changing the application name will recreate all configurations. Another thing: there should be such a behavior from the previous renaming/"configuration folder change" which should be modified to support both changes of the path.

rolandlo:
@Febbe Ok, I will check it carefully. There is the XournalMain::migrateSettings() method, which should take care of that, but yes, it must be modified. 
One question: Is it acceptable if the path is copied (not renamed) and the old path (with files) is just left in the filesystem? I think renaming (rather than copying) would lead to problems when you use different (old and new) versions of Xournalpp. Maybe it would be good to leave a warning about the (potentially) unused path left in the filesystem. What do you think?

Febbe:
Yes that's perfectly fine. Very good idea!

aDogCalledSpot:
> @aDogCalledSpot Are you fine with this way of fixing the problem as well?

I think it's good but looking at the code there I noticed that `operator/` and `operator+=` should be doing the same thing except that a new object is returned. We should make sure the implementation matches there too.

Some brief unit tests for each operator would probably be in order.

Febbe:
This part of the code will be and can be replaced by std::filesystem. std:: filesystem uses /=

aDogCalledSpot:
> This part of the code will be and can be replaced by std::filesystem. std:: filesystem uses /=

Is this being worked on currently? Otherwise I would be happy to create a PR implementing this.

Febbe:
No currently noone is working on it. I planed to do this.

You are welcome to help us with that.

y2kmurali:
Just saw the comments, I am comfortable with std::filesystem as a way to fix the pathing issues.

@rolandlo what other changes need to be made to the migrateSettings function? I assumed that it was functional and it seemed to work fine - was it just adding that extra information message, or is there something else? Either way let me know and I can look into those changes.

rolandlo:
> @rolandlo what other changes need to be made to the migrateSettings function? I assumed that it was functional and it seemed to work fine - was it just adding that extra information message, or is there something else? Either way let me know and I can look into those changes.

The thing is that we change the path for the second time. It was already changed in PR #1384. For instance the config files under Linux used to be in the ~/.xournalpp folder, then they got migrated to the ~/.config/com.github.xournalpp.xournalpp folder and now they should be migrated to the ~/.config/xournalpp folder. 
Now there are users which have already made the first migration step and users that haven't. Moreover from the first group there are users who have removed the old folder and those who haven't. 
So the migration code should first check if the first migration step has already been done. If yes, migrate the files from there. If no it should migrate the files from the original place (before the first migration step).

If the migration seems functional to you, then you probably haven't removed the old folder and the files from the old folder got copied. That's not how it should work.

Febbe:
Ok, the filesystem change is now implemented. You can implement the behaviour to copy the config directory after an update.

Febbe:
> was it just adding that extra information message

You need to add the current configuration directory on top of the array defined in the function.
The function will search for old directories and copy them to the new path.

Febbe:
This can't be merged, because it breaks: #1914 

It must be solved by starting xournalpp as described in the following tutorial: 
https://developer.gnome.org/gtk3/stable/ch01s04.html

Issue #2091, opened 2020-07-01
PortAudioProducer: Unable to open stream


step-:
**Affects versions :**
 - OS: Linux
 - (Linux only) Desktop environment: X11
 - Which version of libgtk do you use gtk3-3.22.30-x86_64-1
 - Version of Xournal++: a320e1c
 - Installation method: self-compiled

**Describe the bug**
Xournalpp can't start recording on my Linux box neither with the default sound card mic nor with my USB webcam's mic.

**To Reproduce**
1. Compile
2. Run, set sound directory preference, click the recording button.

**Expected behavior**
Recording works.

**What happens instead**
Error dialog "Recorder could not be started".

**Discussion**
Xournalpp can't start recording on my Linux box neither with the default sound card mic nor with my USB webcam's mic.

So I compiled [libportaudio stable](http://www.portaudio.com/archives/pa_stable_v190600_20161030.tgz) from [portaudio.com](http://www.portaudio.com/download.html) with `--enable-debug-output` and debug symbols, and xournalpp a320e1c with debug symbols.

`--enable-debug-output` helped troubleshooting thanks to this message:
```
PaAlsaStreamComponent_InitialConfigure: Wanted 16000.000, closest sample rate was 1.000
```
By default the Xournalpp GUI allows setting the minimum sampling rate preference to 44100 Hz (default), which seems to cause this issue. After manually editing `settings.xml` and changing:
```
<property name="audioSampleRate" value="16000.00000000" />
```
from its default value `44100.00000000`, and restarting xournalpp, I was able to successfully record audio from my webcam's mic.

I tested again, this time setting `audioSampleRate` to `22050` but it produced the same "Unable to open stream" error.

Therefore, it seems that the Xournalpp GUI should allow for `16000` Hz as the minimum sampling rate and set that as the conservative default value, which the users can always increase if necessary.

Console log output attached.
[console.log](https://github.com/xournalpp/xournalpp/files/4857075/console.log)

Chris00:
Thanks for the work on this.  I installed the last _nightly build_ (<code>1.1.0+dev~git20200921.0005-nogitfound-buster-1</code>) and changed in the “Audio recording” settings the sample rate to “16000 Hz” but, when trying to start recording, I still have in the message “Recorder could not be started” and, in the shell, 
```
** Message: 22:21:16.140: Start recording
** Message: 22:21:16.140: PortAudioProducer: Unable to open stream

[...]

PR #2092, opened 2020-07-01
Add 16kHz sample rate and fix 91kHz CB label.

step-:
This commit implements the change suggested in #2091. It also changes
the unmatched combo box label value "91000" to "96100" to reflect the
integer value that the `load` settings function expects.

### DIFF
--- a/ui/settings.glade
+++ b/ui/settings.glade
@@ -4243,8 +4243,9 @@ If available select &lt;small&gt;&lt;tt&gt;pulse&lt;/tt&gt;&lt;/small&gt; as inp
                                             <property name="can_focus">False</property>
                                             <property name="active">0</property>
                                             <items>
+                                              <item id="16000">16000 Hz</item>
                                               <item id="44100">44100 Hz</item>
-                                              <item id="91000">91000 Hz</item>
+                                              <item id="96000">96000 Hz</item>
                                               <item id="192000">192000 Hz</item>
                                             </items>
                                           </object>
--- a/src/gui/dialog/SettingsDialog.cpp
+++ b/src/gui/dialog/SettingsDialog.cpp
@@ -391,15 +391,18 @@ void SettingsDialog::load() {
     }
 
     switch (static_cast<int>(settings->getAudioSampleRate())) {
-        case 96100:
-            gtk_combo_box_set_active(GTK_COMBO_BOX(get("cbAudioSampleRate")), 1);
+        case 16000:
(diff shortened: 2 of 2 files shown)


step-:
@rolandlo I was under the impression that the default value is saved to the user's configuration file on first run, therefore existing users wouldn't be affected by the new default because they already have a configuration file.  But it doesn't matter what the default value is, as long as 16 kHz is offered as a choice.  I committed the change you requested.

rolandlo:
@step- Ok, you are right about that. It only occurs after deleting the settings file. Anyway 44100 Hz is a very commonly used sample rate.

You have changed back the default in the Settings.cpp file, but not in the SettingsDialog.cpp file. So please change that as well.

Looking at various sources about sample rates (e.g. [wikipedia](https://en.wikipedia.org/wiki/Sampling_(signal_processing)#Applications)) I find that 16000 Hz, 44100 Hz and 19200 Hz are all commonly used sample rates, but 96100 Hz is not. It should be 96000 Hz instead. So you should change the number 96100 to 96000 (consistently). 
Once you have made the changes you may want to change the PR status from "Draft" to "Ready for review" so that it can be merged.

step-:
@rolandlo, yeah, I was wondering too why the original code used 96.1kHz instead of the more common 96kHz but since I found some internet sources citing 96.1kHz I didn't raise any question.  Anyway, I've implemented the changes you requested.

rolandlo:
Ok, everything works fine now. Can you please squash your commits into one (following [this tutorial](https://www.internalpointers.com/post/squash-commits-into-one-git))? Alternatively I can do it for you from within github. After the squashing I will merge your PR.

ui/settings.glade
```diff
@@ -4243,8 +4243,9 @@ If available select &lt;small&gt;&lt;tt&gt;pulse&lt;/tt&gt;&lt;/small&gt; as inp
                                             <property name="can_focus">False</property>
                                             <property name="active">0</property>
                                             <items>
+                                              <item id="16000">16000 Hz</item>
                                               <item id="44100">44100 Hz</item>
-                                              <item id="91000">91000 Hz</item>
+                                              <item id="96100">96000 Hz</item>
```
rolandlo, line 4248: Sorry to request one more change, but I noticed that the item id is set inconsequently for the 96000 Hz sample rate.
```suggestion
                                              <item id="96000">96000 Hz</item>
```

step-:
@rolandlo done and squashed in c252a7f1.  The following commit ac041671 didn't change anything but was necessary to sync my local repo.

rolandlo:
@step- Thanks for your commit. I have merged it.

PR #2102, opened 2020-07-07
Drop-down menus for right and left handers

MichelReims:
Windows configuration for drop-down menus for right and left-handers does not work.


rolandlo:
@MichelReims
As you can see something's wrong here. Your PR consists of 250+ commits changing 764 files. That can't be right. Give it another try. If you need some help, have a look at [this site](https://www.freecodecamp.org/news/how-to-make-your-first-pull-request-on-github-3/).

Issue #2093, opened 2020-07-02
Cursor shape


mauricebis:
Thanks for the app, which I've just started using. I like it.

Now, I find that the cursor moving constantly between an arrow and a dot when the pen is in contact with the tablet gives much instability for the eye sight (I'm not that young) when handwriting. Would it make sense to have an option where the cursor is a small pointing arrow which remains stable whether the pen is in contact with the tablet or not.

4VRDriver:
#1671 is an duplicate of this issue.

I agree that there should be an option to not switch the cursor to a dot/pen-symbol. There is an option to enable a pen symbol instead of the dot, but it still switches back and fourth. A consistent pen arrow would make sense, but I think some users would prefer to just use the normal cursor for writing. So there should be some settings so that the user can decide what fits best for himself.

duncanawoods:
Adding my request here as advised by @rolandlo 

My preference would be similar to writing on an iPad i.e. while in pen-mode, there should be absolutely no mouse cursor at all for the drawing area whether the stylus is in contact with the page or not.

Rationale:

- when writing quickly, the mouse cursor constantly flashing up as lift the stylus for each new word is incredibly distracting
- when lifting the stylus for crossing t's dotting i's, the mouse cursor obscures where I wish to write harming accuracy
- the mouse cursor is redundant, the stylus tip is literally a pointing device!

If possible, the mouse cursor would still appear on window chrome because the hover-state does help trigger tool-tops and such.

Would a pull-request for this behaviour enabled by a preference flag be accepted or is addressing this imminent?

rolandlo:
@duncanawoods Some questions:
1. With pen-mode do you mean 
- the mode when the ToolType is set to PEN or HIGHLIGHTER 
or 
- that the ToolType is set to PEN or HIGHLIGHTER and additionally the DrawingType is STROKE_RECOGNIZER or DEFAULT? The cursor does not disturb when using the rectangle, circle, arrow, ruler (=line segment), coordinate system or spline DrawingType, does it?
2. With "absolutely no mouse cursor" you mean 
- that there is not even a dot indicating the stylus tip? So you have no clue where your cursor position is, if you are not drawing? 
or 
- that you still see a tiny dot? 

The least change in code would be required if you want to have the first option in 1. and the second option in 2. 
You basically only need to add the ui for the setting, connect it to the code and change line 254 in XournalppCursor.cpp  
[...]

PR #2111, opened 2020-07-11
Support "no icon" stylus cursor type #2093

duncanawoods:
# Purpose

Support a "no icon" cursor type for stylus use for a more "like paper and pen" experience with less distraction.

See issue #2093

# Implementation

I have replaced the `showBigCursor` boolean setting with a `stylusCursorType` enum and updated the settings and preference pages.

- Followed the settings config for `scrollbarHideType`
  - used toString/fromString methods to avoid repetition of constants
  - saved the setting as text rather than an int to be readable
  - add an xml comment for the available options
- Followed the ui config for `cbAudioSampleRate`
- Followed the dialog layout for the stylus screen
  - used a frame and full width combo
  - I have copied the alignment
- Replaced use `showBigCursor` cursor with an enum
- Defaulted the enum to the current DOT cursor
- I have run git-clang-format

# Testing

I have tested:
- tool modes: pen, highlight, shapes, stroke-recogniser
- changing the setting and ensuring it is saved/loaded properly

# Queries:

- I haven't done anything to patch older settings files so if someone has chosen "big cursor" settings, it will be reset to a dot
- I would personally define things like combo-box indices and view ids as constants but chose to follow similar code elsewhere to be consistent
- I don't throw any type of error if the xml cursor type setting isn't valid, I just silently default to a dot. Maybe add a log message?
- I have no `.glade` experience so just copying what I see for other settings
- You may have a preferred location of the enum toString/fromString functions I have added in Settings.h
- I didn't think `strcmp` was advisable but it's used elsewhere so I have been consistent

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -69,7 +69,7 @@ void Settings::loadDefault() {
     this->menubarVisible = true;
 
     this->autoloadPdfXoj = true;
-    this->showBigCursor = false;
+    this->stylusCursorType = STYLUS_CURSOR_DOT;
     this->highlightPosition = false;
     this->cursorHighlightColor = 0x80FFFF00;  // Yellow with 50% opacity
     this->cursorHighlightRadius = 30.0;
@@ -341,8 +341,8 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         this->presentationMode = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("autoloadPdfXoj")) == 0) {
         this->autoloadPdfXoj = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
-    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("showBigCursor")) == 0) {
-        this->showBigCursor = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("stylusCursorType")) == 0) {
+        this->stylusCursorType = stylusCursorTypeFromString(reinterpret_cast<const char*>(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("highlightPosition")) == 0) {
         this->highlightPosition = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("cursorHighlightColor")) == 0) {
@@ -739,7 +739,9 @@ void Settings::save() {
     WRITE_STRING_PROP(presentationHideElements);
     WRITE_COMMENT("Which gui elements are hidden if you are in Presentation mode, separated by a colon (,)");
 
-    WRITE_BOOL_PROP(showBigCursor);
+    xmlNode = saveProperty("stylusCursorType", stylusCursorTypeToString(this->stylusCursorType), root);
+    WRITE_COMMENT("The cursor icon used with a stylus, allowed values are \"none\", \"dot\", \"big\"");
+
     WRITE_BOOL_PROP(highlightPosition);
     WRITE_UINT_PROP(cursorHighlightColor);
     WRITE_UINT_PROP(cursorHighlightBorderColor);
@@ -1018,15 +1020,15 @@ void Settings::setDrawDirModsRadius(int pixels) {
     save();
 }
 
+auto Settings::getStylusCursorType() const -> StylusCursorType { return this->stylusCursorType; }
 
-auto Settings::isShowBigCursor() const -> bool { return this->showBigCursor; }
-
-void Settings::setShowBigCursor(bool b) {
-    if (this->showBigCursor == b) {
+void Settings::setStylusCursorType(StylusCursorType type) {
+    if (this->stylusCursorType == type) {
         return;
     }
 
-    this->showBigCursor = b;
+    this->stylusCursorType = type;
+
     save();
 }
 
--- a/src/control/settings/SettingsEnums.cpp
+++ b/src/control/settings/SettingsEnums.cpp
@@ -1,0 +1,15 @@
+#include "SettingsEnums.h"
+
+auto stylusCursorTypeFromString(const string& stylusCursorTypeStr) -> StylusCursorType {
+    if (stylusCursorTypeStr == "none") {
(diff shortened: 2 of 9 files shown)


duncanawoods:
Thanks @rolandlo I get on these tomorrow:

- [ ] Add a log message if the settings xml contains an unknown type
- [ ] Create a `SettingsEnum` class
- [ ] Use `string&` not `char*` for stylusCursorTypeStr
- [ ] Make the combo box entries translatable
- [ ] Move the cursor icon combo box to the same line should be on the same line as the label
- [ ] Move the combo box to the top of the cursor section
- [ ] Set other cursor settings "visible" based on whether a cursor icon is chosen

duncanawoods:
Hi @rolandlo I have completed the requested changes.

I'm not sure if this is how a pull-request that has been updated from upstream should look. I'm a bit new to pull-requests & git. If I need to clean up the commit history then please give me a quick pointer for what I need to do. Thanks!

rolandlo:
@duncanawoods The commit history of your PR should only list those commits that come from you, not also those which have been merged into master in between. I fixed this following this [stackoverflow answer](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch)
I will now take a close look at your changes. So far it looks good to me.

duncanawoods:
Oh great thanks! I have made this fix to #2112 as well.

src/control/settings/Settings.cpp
```diff
@@ -739,7 +739,9 @@ void Settings::save() {
     WRITE_STRING_PROP(presentationHideElements);
     WRITE_COMMENT("Which gui elements are hidden if you are in Presentation mode, separated by a colon (,)");
 
-    WRITE_BOOL_PROP(showBigCursor);
+    saveProperty("stylusCursorType", stylusCursorTypeToString(this->stylusCursorType), root);
```
rolandlo, line 742: The comment gets written on the wrong position in the settings file. Fix:
```suggestion
    xmlNode = saveProperty("stylusCursorType", stylusCursorTypeToString(this->stylusCursorType), root);
```

src/control/settings/SettingsEnums.h
```diff
@@ -0,0 +1,99 @@
...
+        case STYLUS_CURSOR_DOT:
+            return "dot";
+        case STYLUS_CURSOR_BIG:
+            return "big";
+        default:
+            return "unknown";
+    }
+}
+
+StylusCursorType stylusCursorTypeFromString(const string& stylusCursorTypeStr);
```
rolandlo, line 99: Add a new line at the end of the file:
```suggestion
StylusCursorType stylusCursorTypeFromString(const string& stylusCursorTypeStr);

```

src/gui/dialog/SettingsDialog.cpp
```diff
@@ -188,6 +193,13 @@ void SettingsDialog::customHandRecognitionToggled() {
     gtk_widget_set_sensitive(get("boxCustomTouchDisableSettings"), touchMethod == 2);
 }
 
+void SettingsDialog::customStylusIconTypeChanged() {
+    GtkWidget* cbStylusCursorType = get("cbStylusCursorType");
+    int stylusCursorType = gtk_combo_box_get_active(GTK_COMBO_BOX(cbStylusCursorType));
+    bool showCursorHighlightOptions = stylusCursorType != STYLUS_CURSOR_NONE;
+    gtk_widget_set_visible(get("highlightCursorGrid"), showCursorHighlightOptions);
```
rolandlo, line 200: Maybe it would be better to only make these settings insensitive than invisible.
```suggestion
    gtk_widget_set_sensitive(get("highlightCursorGrid"), showCursorHighlightOptions);
```

src/control/settings/SettingsEnums.cpp
```diff
@@ -0,0 +1,15 @@
+#include "SettingsEnums.h"
+
+StylusCursorType stylusCursorTypeFromString(const string& stylusCursorTypeStr) {
```
rolandlo, line 3:
```suggestion
auto stylusCursorTypeFromString(const string& stylusCursorTypeStr) -> StylusCursorType {
```

duncanawoods:
- [ ] stylusCursorTypeToString function to be defined in the cpp-file not in the header file.
- [ ] return a const string, not const char*

@rolandlo 

> Question: what does context=... in the settings.glade toolbox items do?

Ah - it's the Glade designer box labelled "context for translation". It seemed sensible given some translation tools just present a table of raw strings leading to odd translations given ambiguous meanings and no context. Shall I remove them?

rolandlo:
> > Question: what does context=... in the settings.glade toolbox items do?
> 
> Ah - it's the Glade designer box labelled "context for translation". It seemed sensible given some translation tools just present a table of raw strings leading to odd translations given ambiguous meanings and no context. Shall I remove them?

@duncanawoods Makes sense. You can leave it. I just did not know about it.

duncanawoods:
@rolandlo 

I have made the same change to the existing `buttonToString(Buttons button)` as well.

Febbe:
> Optional change: I would prefer the stylusCursorTypeToString function to be defined in the cpp-file not in the header file. This would be more in style with existing code. Moreover make it return a const string, not const char*. You will need to add .c_str() when you call it in the settings file with saveProperty.

I need to intervene here, since stylusCursorTypeToString was defined as constexpr it must return a constexpr type. Also it must be implemented in the header file. Consistency with the old code is also not a good reason to follow bad design. And converting an enum type to an constant string is indeed a good case to use a constexpr. So either still use char const* const or return a std::string_view to a constexpr char* based - string.
And since all of the strings are const static / constexpr they won't need the lifetime guard of a std::string.

duncanawoods:
@Febbe @rolandlo I am happy to make the change you agree on.

>  Can you squash your commits into one following this recipe?

Apparently, for me, the answer is "no". I followed the recipe and it triggered merge conflicts. I resolved them and my local commit history looked fine but when pushing, it triggered another merge. I now appeared to have added two more commits. Sigh.

I think it's because my commit history contains merge resolutions. I've tried some of the approaches here with no luck:

https://stackoverflow.com/questions/3133449/why-does-git-rebase-give-me-merge-conflicts-when-all-im-doing-is-squashing-comm

Do you have any suggestions? e.g. maybe a nuclear option of backup the code, hard reset my fork and reapply the changes manually?

rolandlo:
@duncanawoods Please follow @Febbe's suggestions. He is much more experienced than me, both what concerns code style and git handling.

duncanawoods:
Ok, thank you for patience and encouragement @rolandlo, much appreciated!

Febbe:
Ok, do you have time at the evening (9pm CET) I can help you then.

duncanawoods:
Yes (8pm UK time) thank you @Febbe. I'll make the string changes now.

Febbe:
I am in gitter rn.

Issue #2074, opened 2020-06-17
Hide Toolbar toggle shortcut (like F10 is for Menu), used together with Floating Toolbox


tjavdar:
Now that one has floating toolboxes, one should be able to pop/hide the static toolbar with a single shortcut. This allows to use all the screen area while having all the tools in a toolbox at hand. 

The problem at present is that hiding a menu is a long-fetched action:
                         (F10, if menus are hidden)->View->Toolbars->Empty toolbar

Desired solution: Create a shortcut to pop/hide the static toolbar (e.g., Shift-F10, as F10 is for Menu)

Workflow if implemented:
   0. Configure first a  personal floating toolbox by: View->Toolbars->Floating Toolbox->Customize,
       then assign it to a mouse button (File->Preferences->Mouse) or the stylus (File->Prefs->Stylus).
       This is  done once.
   1. At session start: Hide Menu (F10) and Toolbar (Shift-F10).
   2. Start writing/sketching using the whole screen.
   3. Invoke the toolbox to change colors, pens, actions, when needed. Go to 2 -> cycle.

duncanawoods:
I'm interested in this too. I've had a look and created a menu item that toggles the toolbars on f9.

@rolandlo is this desirable?

rolandlo:
@duncanawoods That sounds like a logical choice and desirable to me. There was even a PR for this #461, but somehow was closed without merging. I will take a look at your PRs next week.

duncanawoods:
Oh neat. #461 looks similar but better than mine - it saves the setting and toggles visibility more neatly.

rolandlo:
@duncanawoods If #461 is a better implementation for the toolbar toggle shortcut, then we should ask the author @luco5826 to open his PR again.

luco5826:
Hi @rolandlo , thanks for calling me in! If I recall correctly (jk, I just re-read that PR) the problem was about adding "a lot of code" for a feature that was already there (as OP said, through `View->Toolbars->Empty toolbar`).

It's been a long time since that PR and I did not follow this project consistently enough to be sure that #461 would not break anything. I could reopen that but it was more than one year ago, if @duncanawoods had the time and the patience to recheck it or merge it with its work that would be awesome!

duncanawoods:
Hi @luco5826 ok I will take it on.

rolandlo:
Closed by #2112

Issue #2025, opened 2020-05-30
Allow pages to be 100% fullscreen in presentation mode


stevenbell:
I'm hoping to use Xournal++ to draw on my PDF slides as I present. To do this, I'd like the slides to be 100% fullscreen. When I hide the toolbars, use "presentation mode" and go fullscreen, there is still a border around my slide and I can see a sliver of the next or previous slide, as shown in the screenshot.

![Screenshot from 2020-05-30 12-29-22](https://user-images.githubusercontent.com/1936591/83333978-779cf280-a271-11ea-9c5a-a919ed13c5cf.png)

I tried using "Fit to screen", but this appears essentially the same (except that I can continuously scroll through the pages instead of flipping one to the next as in presentation mode).

Setting the background and "current page" highlights to black is a reasonable workaround, but it wastes some pixels.

I'm running Xournal++ 1.0.16 (installed via snap) on Ubuntu 18.04.

stevenbell:
I hacked on the code a bit and got pretty close.  By setting [`XOURNAL_PADDING`](https://github.com/xournalpp/xournalpp/blob/master/src/gui/Layout.cpp#L16) and [`XOURNAL_PADDING_BETWEEN`](https://github.com/xournalpp/xournalpp/blob/master/src/gui/Layout.cpp#L26) to 0 and removing the `+ 14.0` in [ZoomControl.cpp](https://github.com/xournalpp/xournalpp/blob/master/src/control/zoom/ZoomControl.cpp#L225), most of the extra space is gone.

This is almost a fix, but:
* `XOURNAL_PADDING` is obviously used in the other modes, so setting it to zero should be conditional on being in presentation mode.
* Is it intended that "fit to screen" causes the page to fill the entire screen?  This is what I'd like, but the current  `+ 14.0` padding is clearly intentional, not just a side effect.
* Can the border be turned off in presentation mode?  There's only one active page, so highlighting it seems superfluous.
*  There are still a couple rows of empty pixels at the bottom, but this appears to be in GTK rather than in the page rendering/layout.  Any clues where to start looking at this would be appreciated.

duncanawoods:
I'm happy to take a look at this. @rolandlo can you help on some answers to @stevenbell questions please?

> Is it intended that "fit to screen" causes the page to fill the entire screen?

This is what I would want/expect too.

> Can the border be turned off in presentation mode?

[...]

PR #2112, opened 2020-07-11
Add "Show toolbars" menu toggle item bound to F9

duncanawoods:
# Purpose

This is a suggested fix for #2074 to hide toolbars as easily as other screen furniture. It is based on the toggle menu item for hiding the sidebar and bound to the F9 accelerator.

# Queries 

- I have cached the currently visible toolbar as `ToolbarData* lastToolbar`. I don't know the implications of storing that (e.g. could it become invalidated?) or where it should best be kept. Could just store the id if necessary.
- Unlike the `show sidebar` I haven't created a persistent setting for hidden toolbars
- It's kind of odd that menu accelerators no longer work if you hide menus
- I think it would be neater to have _Zen Mode_ to disable toolbars/menu/side-bar in one step
- If this is accepted, we can also remove the `Empty Toolbar` toolbar type which I find a bit weird

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -62,6 +62,8 @@ void Settings::loadDefault() {
     this->showSidebar = true;
     this->sidebarWidth = 150;
 
+    this->showToolbar = true;
+
     this->sidebarOnRight = false;
 
     this->scrollbarOnLeft = false;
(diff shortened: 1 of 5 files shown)


duncanawoods:
Combine with aspects of similar PR #461

- [ ] create a persistent setting for toolbar visibility
- [ ] change toolbar visibility without caching the last toolbar

duncanawoods:
@rolandlo @tjavdar 

Ok ready for review.

duncanawoods:
@rolandlo fantastic, thanks!

> One idea would be to address issue #2025

I'll take a look and do it separately. I have some questions that I'll add to the issue.

PR #2125, opened 2020-07-20
Add a lower limit for thickness of strokes.

### LINKED ISSUE #2118: Add a lower limit for thickness of strokes
**Affects versions :**
 - OS: Manjaro Linux
 - (Linux only) Desktop environment: KDE Plasma, X11
 - Which version of libgtk do you use: 3.24.21
 - Installation method: GIT

**Describe the bug**
For this bug you need a pressure sensitive pen.
There are cases when a stroke is drawn but the tip of the stylus does not sense any pressure. This can be caused from a bug (#2053) or when a input event is very short in time. In my case this happens a lot when i try to write the dot over the letter 'i'. In 20% of the times the dot will not be recognized.
The issue here is that the stroke will get drawn but with the width of "0.00010757" which is extremely thin. Xournalpp does not display such thin lines at all. However they are in the xopp file and when exported as PDF they will be shown by most of the PDF viewers as a very thin line. (I will add a xopp file)

A quick fix for the problem would be to add a lower limit to the thickness of all strokes. That way the user can see the stroke at least in Xournalpp and erase it if it wasn't on purpose.
For testing purposes I added those lines to StrokeHandler.cpp:71
```
if (Point::NO_PRESSURE != pos.pressure && stroke->getToolType() == STROKE_TOOL_PEN) {
    const double LOWER_LIMIT = 0.2;
    double pressure = pos.pressure;
    if(pressure < LOWER_LIMIT) {
        pressure = LOWER_LIMIT;
    }
    stroke->setLastPressure(pressure * stroke->getWidth());
}
```
And voila no invisible strokes any more. Also all dots over the letter 'i' get recognized now (but they are sometimes really thin tho).

**To Reproduce**
1. Open the xopp file I uploaded.
2. Verify that you see a blank page.
3. Export as PDF and see the thin stroke.

[thin_line.xopp.txt](https://github.com/xournalpp/xournalpp/files/4941368/thin_line.xopp.txt)
[i.xopp.txt](https://github.com/xournalpp/xournalpp/files/4941370/i.xopp.txt)

**Additional**
Adding more pressure steps in general may make a lower pressure limit obsolete, but this has to be tested.

4VRDriver:
Fixes #2118 

The PR is for testing purposes.

The actual value of the lower limit (here: 0.2) should be discussed.

### DIFF
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -68,6 +68,11 @@ auto StrokeHandler::onMotionNotifyEvent(const PositionInputData& pos) -> bool {
     }
 
     if (Point::NO_PRESSURE != pos.pressure && stroke->getToolType() == STROKE_TOOL_PEN) {
+        const double LOWER_LIMIT = 0.2;
+        double pressure = pos.pressure;
+        if(pressure < LOWER_LIMIT) {
+            pressure = LOWER_LIMIT;
+        }
         stroke->setLastPressure(pos.pressure * stroke->getWidth());
     }
 
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -73,7 +73,7 @@ auto StrokeHandler::onMotionNotifyEvent(const PositionInputData& pos) -> bool {
         if(pressure < LOWER_LIMIT) {
             pressure = LOWER_LIMIT;
         }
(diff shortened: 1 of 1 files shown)


rolandlo:
In your approach you set a lower bound for the pressure (20% of the maximal pressure), not for the stroke width of the line segments making up the stroke. Not sure this is what you intended.

Issue #2144, opened 2020-08-01
Adding new pdf pages does not work


rolandlo:
**Affects versions :**
 - OS: Ubuntu 20.04
 - (Linux only) Desktop environment:: X11
 - Which version of libgtk do you use: 3.24.20
  - Version of Xournal++: dev1.1.0+dev 
- Installation method: PPA

**Describe the bug**
Adding newly created pdf pages in a pdf was possible by adding new pages with "pdf background". Now this does not work any more. The app freezes after hitting the insert button.

**To Reproduce**
1.Open a pdf in Xournal++
2. Save the xopp-file
3. Add some pages to the pdf file (externally, for example using pdftk)
4. Open the xopp-file again
5. Set the background option for inserted pages to "pdf background"
6. Click on "Insert page" and choose one of the new pdf pages
7. Observe that the app freezes

**Expected behavior**
I expect that the app does not freeze and the new pdf page is inserted into the document (like it used to be)

**Additional context**
Running git-bisect shows that the regression was introduced with PR #1613. The freeze results from an infinite loop, in which firePageSelected and fireZoomChanged methods call each other.

rolandlo:
A temporary fix is provided in PR #2146.

Technius:
Same source problem as #1660.

rolandlo:
Seems like PR #2312 fixes the bug. The temporary fix in PR #2146 (which made use of the fact that the focus shifted to the first page) can be removed then.

Issue #1588, opened 2019-11-04
[Feature Request] Update underlying PDF file


vinieich:
I have a PDF that have one of my classes topics and my professor changes it after every class.

Sometimes he updates it with 50+ pages and I use this pdf on Xournalpp but the problem is that the only way to add the new pages is manually selecting one by one.

I've tried to delete the PDF and then selecting the new one as the background (using the xournalpp dialog that complains for the missing pdf after opening the file) but the page count is still the same.

PS:
I put the [Feature Request] tag because I honestly searched everywhere that I could but couldn't find how to do this. If there is some way to do this then I apologize.
English is not my primary language so maybe the text is a little confuse. I can try to reformulate if necessary.

LittleHuba:
This is actually a duplicate of #889. Stil thanks for reporting this.

I'm not quite sure it is worth adding this functionality before moving to the new rendering system. It is quite a huge modification and we would have to do it twice...

For now you might want to open the new pdf in another instance of Xournal++ and just move all the annotations via copy&paste if it is not too much hassle.

PR #2146, opened 2020-08-02
append new pdf pages via menu

rolandlo:
The idea of this PR is to provide a convenient option to append all new pdf pages to a Xournal++ document after the pdf has been updated externally with new pages. This is already possible by adding one page after the other, but that is not so practical if lots of pages have been added, see issue #1588.

Unfortunately there is bug #2144 which needs to be fixed. I have only made a temporary fix, which softens the problem, but does not quite solve it. With the temporary fix you won't have any freezes, but the page number is not updated if you change to the very first page of your document. This would have to be addressed before the PR can be merged.


LittleHuba (changes requested):
Looks good to me so far. The warnings should definitely be provided as a dialog though. Most users do not see such messages posted to stderr.

rolandlo:
@LittleHuba The errors are now reported in XojMsgBox'es. Do you have an idea about the freezing bug #2144 (which I "fixed" at the cost of pages not updating when scrolling to page 1)? It was introduced after you reintroduced the "select page after scrolling" algorithm. I haven't figured out yet, whether we should introduce a condition to break the infinite loop or whether there is something wrong with the algorithm to compute the most visible page (when there is a pdf background).

LittleHuba:
That is a taff problem. I'll look into that issue tonight. Most likely we have to separate the algorithm into two different ones.

src/control/Control.cpp
```diff
@@ -1235,6 +1238,48 @@ void Control::deletePage() {
...
+    // find last page with pdf background and get its pdf page number
+    PageRef page;
+    int lastPdfPage = -1;
+    for (int i = pageCount - 1; i >= 0; --i) {
+        page = doc->getPage(i);
+        if (page && page->getBackgroundType().isPdfPage()) {
+            lastPdfPage = page->getPdfPageNr();
+            break;
+        }
+    }
```
Febbe, line 1252: All of those return values are size_t, so don't cast to int implicitly without bound checks.

```suggestion
    auto pageCount = this->doc->getPageCount();
    // find last page with pdf background and get its pdf page number
    auto currentPdfPageCount = [&]() {
        for (size_t i = pageCount; i != 0; --i) {
            if (auto page = doc->getPage(i - 1); page && page->getBackgroundType().isPdfPage()) {
                return page->getPdfPageNr() + 1;
            }
        }
        return 0ULL;
    }();
```

src/control/Control.cpp
```diff
@@ -1235,6 +1238,48 @@ void Control::deletePage() {
...
+    if (lastPdfPage == -1) {
+        string msg = FS(_F("Document has no pdf pages."));
+        XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+        return;
+    }
```
Febbe, line 1257: I think in this cornercase should we should add the new pages. Imagine a document with all PDF pages deleted, and you add new pages to the PDF, now all the pdfs should be added. Am I right?

```suggestion
```

src/control/Control.cpp
```diff
@@ -1235,6 +1238,48 @@ void Control::deletePage() {
...
+
+    int pdfPageCount = this->doc->getPdfPageCount();
+    int numPages = pdfPageCount - lastPdfPage - 1;
+
+    if (numPages <= 0) {
```
Febbe, line 1262: Also use auto or size_t.
```suggestion
    auto pdfPageCount = this->doc->getPdfPageCount();
    auto insertCount = pdfPageCount - currentPdfPageCount;

    if (insertCount == 0) {
```

src/control/Control.cpp
```diff
@@ -1235,6 +1238,48 @@ void Control::deletePage() {
...
+    for (int i = 1; i <= numPages; ++i) {
+
+        doc->lock();
+        XojPdfPageSPtr pdf = doc->getPdfPage(lastPdfPage + i);
+        doc->unlock();
+
+        if (pdf) {
+            auto newPage = std::make_shared<XojPage>(pdf->getWidth(), pdf->getHeight());
+            newPage->setBackgroundPdfPageNr(lastPdfPage + i);
+            insertPage(std::move(newPage), pageCount + i - 1);
```
Febbe, line 1275:
```suggestion
    for (size_t i = 0; i != insertCount; ++i) {

        doc->lock();
        XojPdfPageSPtr pdf = doc->getPdfPage(currentPdfPageCount + i);
        doc->unlock();

        if (pdf) {
            auto newPage = std::make_shared<XojPage>(pdf->getWidth(), pdf->getHeight());
            newPage->setBackgroundPdfPageNr(currentPdfPageCount + i);
            insertPage(newPage, pageCount + i);
```

Febbe (changes requested):

src/gui/Layout.cpp
```diff
@@ -108,7 +108,8 @@ void Layout::updateVisibility() {
         x1 = 0;
     }
 
-    if (mostPageNr) {
+    if (mostPageNr &&
+        mostPageNr.value() != 0) {  // temporary fix for bug #2144. Page number won't update if scrolling to first page
```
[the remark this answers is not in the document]
rolandlo: Yes, we don't want that. It should be fixed. That's why I added the "Do not merge label". 
What do you mean by "empty pdf pages"? Do you mean pdf pages with no actual content (just white background)? I never have empty pdf pages in my documents and I can't think of documents that have, unless the creator wanted them to be there. In any case, deleting empty pages at the end is a lot easier than adding pdf pages that have been created externally (for instance by adding new content to a LaTeX document or adding content to a Word- or Powerpoint-document and re-exporting to pdf).

src/control/Control.cpp
rolandlo: This syntax is new to me. What does the `[&](){ .... }();` stuff do?

src/control/Control.cpp
rolandlo: Ok, I am fine with adding all pdf pages in that case. If all pdf pages are deleted and you save/reload the document there won't be any pdf pages to add, though. It only occurs if you remove the last pdf page in the document and don't save.

src/control/Control.cpp
Febbe: > This syntax is new to me. What does the `[&](){ .... }();` stuff do?

Oh that's a lamda expression. It is part of the standard since c++11.
Basically I inlined a unnamed function here, returning the value for "currentPdfPageCount" and with the prceding `()` I directly executed it.
With that no predeclared variable is necessary and the returnvalue of the lambda function is directly executed.

https://en.cppreference.com/w/cpp/language/lambda

src/gui/Layout.cpp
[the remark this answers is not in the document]
Febbe: I misunderstood #2211

src/control/Control.cpp
```diff
@@ -1247,7 +1247,7 @@ void Control::appendNewPdfPages() {
                 return page->getPdfPageNr() + 1;
             }
         }
-        return 0ULL;
+        return 0UL;
```
Febbe, line 1250: This is only correct on 32bit or Linux machines.
`return ULL;` was already correct, since we don't support 32 bit.
But you can support all machines / compiler:
```suggestion
        return size_t{0U};
```

Febbe:
LGTM after a fixup-rebase.

src/control/Control.cpp
rolandlo: Thanks for the fix. It didn't compile with `return 0ULL` on my (64-bit Linux) machine (gcc 9.2.1). That's why I changed it. I am glad you checked it once more. With `size_t{0U}` it works.

rolandlo:
After bug #2144 has been fixed, this PR looks ready for merging to me. I have removed the workaround for bug #2144 in the code, so it just works without problems.

PR #2170, opened 2020-08-16
Allow the user to disable cursor blinking.

gebner:
This uses the same setting which is also used by the built-in text entry widget.

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -31,6 +31,7 @@ TextEditor::TextEditor(XojPageView* gui, GtkWidget* widget, Text* text, bool own
     gtk_text_buffer_place_cursor(this->buffer, &first);
 
     GtkSettings* settings = gtk_widget_get_settings(this->widget);
+    g_object_get(settings, "gtk-cursor-blink", &this->cursorBlink, nullptr);
     g_object_get(settings, "gtk-cursor-blink-time", &this->cursorBlinkTime, nullptr);
     g_object_get(settings, "gtk-cursor-blink-timeout", &this->cursorBlinkTimeout, nullptr);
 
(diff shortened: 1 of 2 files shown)


src/gui/TextEditor.cpp
```diff
@@ -865,6 +866,12 @@ void TextEditor::repaintCursor() { repaintEditor(); }
  * Blink!
  */
 auto TextEditor::blinkCallback(TextEditor* te) -> gint {
+    if (!te->cursorBlink) {
+        te->cursorVisible = true;
+        te->repaintCursor();
```
LittleHuba, line 871: This is a very bad way of doing this. `repaintCursor()` forces a full page redraw which is very expensive and might lead to more than just the cursor flashing... 

This PR would need repaintCursor to be reimplemented to at least only redraw the area of the TextEditor or even better only the area of the cursor.

LittleHuba (changes requested):

src/gui/TextEditor.cpp
gebner: Ok, I've reimplemented `repaintCursor` to only repaint the area of the current text editor.  Replacing `repaintEditor` is harder because in general we need to repaint more than current area (for example when the user deletes a character).

Just to make it clear, the first version of the PR did not increase the number of page redraws (the number is even reduced if cursor blinking is disabled).  The text editor did a page redraw every time the cursor blinked.

LittleHuba (approved):
Sorry for taking so long!

LGTM! Still have to do a functional test in the next few days.

LittleHuba (approved):
Sorry for the long wait!
Functionality looking good. This can be merged!

LittleHuba:
I'll merge it tomorrow if nobody has any issues with it.

LittleHuba:
Thanks for the work!

Issue #2152, opened 2020-08-08
Recognition of handwritten text


rolandlo:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a feature to extract the handwritten text from a Xournal++ document, so you can reuse it in a text editor. 

**Describe the solution you'd like**
I would like to have a Xournal++ plugin, in which you can select pages (or the whole document) and get the text extracted from the selected pages and save it to a .txt file or copy it to clipboard.

**Describe alternatives you've considered**
Instead of using a Xournal++ plugin we might create an external tool that takes a xopp-file (and page numbers) as input and returns the extracted text as output.

**Additional context**
This idea has also been discussed in #51 and #1953. 
Using the google API to recognize text the task might be much easier than expected. See [this online demo](https://www.chenyuho.com/project/handwritingjs/) and note that the input sent to the google API could be very easily produced from the list of strokes that Xournal++ produces, see [this description](https://github.com/ChenYuHo/handwriting.js#capture-trace).

rolandlo:
Got a first working version, already. Currently upon activating the plugin, the recognized text (a list of possibilities, sorted by likelyhood) is written to the console. See 
![FirstImpression](https://user-images.githubusercontent.com/40485433/89757082-8fe88380-dae4-11ea-9894-cf0cd083840c.png)

A PR will follow.

atticus-sullivan:
Well this looks pretty amazing. The only little concern that I'm having, is that handwriting data is quite personal and I personally would like to know more about how this works (is the data used to improve the AI? Will the selection the user makes which was the correct guess be sent back?).

So I think this is a really nice feature (searching ones own Notes is awesome), but in my opinion this should be (if possible) something like a plugin or something the user can turn on and off, if he/she has concerns.

(Don't get me wrong this is amazing, but I think one should be careful with this)

rolandlo:
@atticus-sullivan 
You are right about privacy concerns. With this plugin the handwritten data (coordinates of the points making up the strokes) is sent to the google server and processed there. We have no control what google is using that data for. 
It is designed as a Lua Plugin and deactivated by default. You first have to activate it and only after clicking on the name of the plugin the data is sent to the google server. In a more refined version you would first get a dialog where you can choose the pages/layers which should get recognized and then you have to click a button to send the data to the google server. You would see a clear warning there.
When you select the correct guess no data is sent to the google server.

rolandlo:
[...]

PR #2176, opened 2020-08-19
[WIP] Lua Plugin for Handwriting recognition

rolandlo:
This is a Lua Plugin that uses the Google IME API for handwriting recognition.

The current initial version only supports recognition of the handwritten text on the selected layer, although the UI shows more options. The language and the number of returns (i.e. the number of alternatives for the recognized text, that are displayed) already work.

This Plugin requires the following lua-modules 
- `lgi `(for creating the UI)
- `socket`,` ssl.https` and `ltn12` (for the `https-request`)

These modules, which can be installed with the `luarocks` package manager, will be searched for in` /usr/local/sahre/lua/5.3/`
If that is the wrong path, the `initUi `function of `recognition.lua` must be modified.

### DIFF
--- a/plugins/HandwritingRecognition/plugin.ini
+++ b/plugins/HandwritingRecognition/plugin.ini
@@ -1,0 +1,15 @@
+[about]
+## Author / Copyright notice
+author=Roland Lötscher
+
+description=Issue https://github.com/xournalpp/xournalpp/issues/2152
+
+## If the plugin is packed with Xournal++, use
+## <xournalpp> then it gets the same version number
+version=<xournalpp>
+
+[default]
+enabled=false
+
+[plugin]
+mainfile=recognition.lua
--- a/plugins/HandwritingRecognition/Dialog.glade
+++ b/plugins/HandwritingRecognition/Dialog.glade
@@ -1,0 +1,558 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
+  <object class="GtkDialog" id="dialog1">
+    <property name="can_focus">False</property>
+    <property name="title" translatable="yes">Text recognition tool: Settings</property>
+    <property name="resizable">False</property>
+    <property name="destroy_with_parent">True</property>
+    <property name="type_hint">dialog</property>
+    <child type="titlebar">
+      <placeholder/>
+    </child>
+    <child internal-child="vbox">
+      <object class="GtkBox" id="mainBox">
+        <property name="can_focus">False</property>
+        <property name="margin_left">18</property>
+        <property name="margin_right">18</property>
+        <property name="margin_top">18</property>
+        <property name="margin_bottom">18</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">18</property>
+        <child internal-child="action_area">
+          <object class="GtkButtonBox" id="actionButtonBox">
+            <property name="can_focus">False</property>
+            <property name="layout_style">end</property>
+            <child>
+              <object class="GtkButton" id="btOk">
+                <property name="label" translatable="yes">Ok</property>
+                <property name="name">btn_ok</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">True</property>
+              </object>
+              <packing>
+                <property name="expand">True</property>
+                <property name="fill">True</property>
+                <property name="position">2</property>
+              </packing>
+            </child>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">False</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkBox" id="contentBox">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="orientation">vertical</property>
+            <property name="spacing">18</property>
+            <child>
+              <object class="GtkFrame">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label_xalign">0</property>
+                <property name="shadow_type">none</property>
+                <child>
+                  <object class="GtkBox">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="spacing">12</property>
+                    <child>
+                      <object class="GtkImage" id="icon_selection">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="pixel_size">60</property>
+                        <property name="icon_name">text-x-generic</property>
+                        <property name="icon_size">6</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkBox">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="orientation">vertical</property>
+                        <property name="spacing">6</property>
+                        <child>
+                          <object class="GtkRadioButton" id="rd_1">
+                            <property name="label" translatable="yes">on the selected layer</property>
+                            <property name="visible">True</property>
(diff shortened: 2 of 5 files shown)


rolandlo:
Current progress can be seen from this video: 
textRecognitionDemo
Basically everything works, just some fine-tuning left.

rolandlo:
As it turns out, the google server is only able to handle at most 150 strokes (roughly 2 handwritten sentences) at the time. That is quite unfortunate. Splitting up the strokes into chunks of roughly 150 strokes is possible, but it is not trivial to patch the recognized text pieces together.

amirji:
I was thinking about this API:
https://developer.myscript.com/ 
Much more supported than google API.

rolandlo:
@amirji I have seen the myscript API. It certainly looks pretty powerful. But only a limited number of requests per month are free. After that it quickly gets expensive. Moreover you need a subscription and only iOS, Android, Windows and Web are supported; Linux support is missing. That makes it problematic.

rolandlo:
This will stall for a while until I have extended the Lua Plugin API. It is better to do that in a consistent manner than to extend the API time by time just to make one single plugin work.

DeveloperOl:
Why are you using the google online recognition and not something like _libtesseract_ from https://github.com/tesseract-ocr/tesseract ?

rolandlo:
@DeveloperOl See #2152 for the reason. In short: Tesseract is off-line text recognition.

rolandlo:
An interesting alternative to the Google server would be [MathPix](https://docs.mathpix.com/). It has a documented API and not only returns handwritten text, but also math. So you potentially could use the selection tool in Xournal++ to select a handwritten formula, activate the plugin via menu or via keyboard accelerator, get the formula translated to LaTeX by Mathpix and Xournal++ would paste it as an editable LaTeX object. Similarly with handwritten text (or even a combination of both).

abrac:
*Just brainstorming*: I found this cool open-source program called [CellWriter](https://github.com/risujin/cellwriter). It is able to do character recognition based on input from a mouse/pen. I wonder if it is possible to pipe out handwritten strokes from Xournal++ to CellWriter... 🤔

rolandlo:
@abrac The problem is, that CellWriter expects the letters to be written in cells, whereas in Xournal++ you can write freely across the whole document.

abrac:
That's true! Maybe the code from CellWriter could still be adapted to somehow work with Xournal++ 🤷. But it would probably take a whole PhD on its own to implement it that way! 🤣

yingshaoxo:
Hey, guys!

How about we do not recognize the text.

Instead, we recognize math symbols, which are more important than pure text:


https://mathpix.com/

https://github.com/ritheshkumar95/im2latex-tensorflow

rolandlo:
@yingshaoxo 
Yes, mathpix is an interesting option (see [above](https://github.com/xournalpp/xournalpp/pull/2176#issuecomment-740031495))
im2latex looks like a weaker solution, which would probably not work with handwritten formulas.

amirji:
@yingshaoxo @rolandlo 
I am using mathpix daily and it is great but it is not free. Only the first 100 tries in each month are free. I am curious if we could make one by ourselves or not, the trained data can be downloaded and no need for a remote API.

yingshaoxo:
We could make an additional local api server by making a python package to
PyPI, then any users who want to use it could just install a python package
to use it. (this step is easy, I'm an Python export.

But the problem is, I can not understand the logic of tensorflow, I could
only understand keras....such a pain...

yingshaoxo:
I found an app: https://math.microsoft.com/en

One it recognized an image, you long press that image, you'll get latex copied into your clipboard.

And this app does not require login, we may use it to recognize.(if anyone knows how to capture network data packages, we could simulate that request to microsoft servers)

personalizedrefrigerator:
@rolando Please break the changes to the plugin API into a separate PR (if you have time :)). It would be nice to have the additional stroke functionality for things like #2850.

rolandlo:
> @rolando Please break the changes to the plugin API into a separate PR (if you have time :)). It would be nice to have the additional stroke functionality for things like #2850.

Not sure how `applib_getStrokes` (which returns a list of all strokes from a Xournal++ document in a Lua table) would help for #2850. In that issue the goal is to put strokes (obtained from some image) into the document, which is the other way round.

It wouldn't be much work to break the changes to the plugin API into a separate PR, if there is need for it. However there are still some things I am unsure about and which may become more difficult to change later, if there are already some plugins based on it:
- Currently a stroke in the Lua table consists of a table for the x-coordinates of the points and a table for the y-coordinates of the points. That's how the Google server (and other recognition software) want the strokes to be passed. The pressure data is ignored (which I think shouldn't). A stroke could also consist of a table of points, where each point is a table consisting of an x-, y- (and pressure-)value. I am not sure what is better.
- `applib_getStrokes` should eventually also yield the width of each stroke, the line style, the color, the fill opacity and the stroke tool (pen, highlighter or eraser) used. This doesn't need to be implemented all at once, but at least in a way, that doesn't require plugin authors to update their plugin each time the API is extended. 
-  Eventually there should also be `applib_getText`, `applib_getImage`, `applib_getTeXImage`
 - Eventually it should be possible to create and filter selections from the plugin API, to modify elements of a selection and to add elements to a selection. That's what will be needed for #2850. All in all this will require a decent amount of work.

personalizedrefrigerator:
It looks like [QtVirtualKeyboard](https://doc.qt.io/qt-5/handwriting.html) uses [Lipi toolkit](http://lipitk.sourceforge.net/lipi-toolkit.htm) for its handwriting recognition... Maybe we could use this...

WillNilges:
Hello, I'm wondering if y'all are still working on this? I'm starting to pick back up on working on this and would love to help out. (Hi it's the guy from #2850)

rolandlo:
@WillNilges Hi! No, I am not working on this currently and I have no plan to do so in the near future. You are welcome to work on this.

kauesena:
That would be very nice as then files would be searchable!

IloveJesusmyKing:
Has anyone looked into the recognition technology that Microsoft uses for OneNote. I have found that their setup is efficient in processing lots of handwriting. It allows for searching handwriting.

All3xJ:
Any news about this amazing plugin?

yingshaoxo:
Don't you guys ever think about using Python?

https://github.com/JaidedAI/EasyOCR

WillNilges:
I've thought about it, and personally, I'd really like it if we could figure out a way to do completely local recognition that didn't rely on proprietary software. This could be a viable option. And I'm looking at integrating python OpenCV into https://github.com/willnilges/inkpath so hopefully I'll have some relevant experience doing that soon.

amirji:
We can use one of these engines for local recognition:

https://github.com/kba/awesome-ocr

However, still, local math recognition will be a challenge.

Zireael07:
I think math support could be done as a follow-up thing?

rolandlo:
Here is a [link](https://wicg.github.io/handwriting-recognition/) to a suggested Handwriting Recognition API that seems to fit well with what we would love to achieve in Xournal++. It includes segmentation information about each transcription, which means that the strokes in Xournal++ could be matched with text characters.

plugins/HandwritingRecognition/plugin.ini
```diff
@@ -0,0 +1,15 @@
+[about]
+## Author / Copyright notice
+author=Roland Lötscher
+
+description=Issue https://github.com/xournalpp/xournalpp/issues/2152
+
+## If the plugin is packed with Xournal++, use
+## <xournalpp> then it gets the same version number
+version=<xournalpp>
```
lucgiffon, line 9: Hello! I don't understand. We should replace <xournallpp> with the actual version number?

Captain-Pumpkinhead:
Any updates?  I'd love to use this plugin.  It'd make integrating Xournal++ with Obsidian much easier.

michucz:
Can't wait to test this once it is available. One question, are you planning, or is it even possible to male the handwritten text searhable after it has been recognized? That would be amazing.

rolandlo:
Please check out https://github.com/PellelNitram/xournalpp_htr for a more promising approach than this plugin.

rolandlo:
Closing in favor of https://github.com/PellelNitram/xournalpp_htr

Issue #2104, opened 2020-07-08
Bad positioning of eterogeneous elements in the exported pdf by a xournal file


LittarruFrancesco:
**Affects versions :1.0.15**
 - OS: Ubuntu 18.04 LTS
 - (Linux only) Desktop environment: MATE
 - version of libgtk: 3.22.30-1ubuntu4
 - Version of Xournal++: 1.0.15
 - Installation method: official release

**Describe the bug**
I made a xournalpp page with eterogeneous elements like images, text, and markers.
The page is just one, but with many small elements.
When exported in pdf, some elements are overlapping and others are moved downwards (text over images and marker in different positions).

**To Reproduce**
1. Open a new xopp file
2. Edit a page with eterogeneous elements, in my case: texts, images, markers
3. Export in pdf
4. The pdf file should look exactly like the xopp one.

**Expected behavior**
I expect the pdf to look identical to the xopp file I designed.

**Screenshots of Problem**
![xournal_bug_comparison_bad_positioning_after_pdf_exportation](https://user-images.githubusercontent.com/11428019/86901265-aa59c680-c10c-11ea-8073-b24d710c809b.png)

Edit: the screen res is 1080p. (PCSpecialist Cosmos VII)

rolandlo:
duplicate of  #1390

rolandlo:
I may have closed this issue prematurely. Could you please provide the xopp-file for testing purposes? You can do it by renaming the file as .txt and dragging it into the text field that you use to answer.

LittarruFrancesco:
Yes sure!
[Input_Capture_Mode.txt](https://github.com/xournalpp/xournalpp/files/4967887/Input_Capture_Mode.txt)

I also tried to open it in another computer with the same OS and the same Xournal version but with screen res of 1366*768 instead of 1080p and the file in xournal++ is formatted even worse than the pdf version of that in 1080p

rolandlo:
It is a bit complicated, since we apparently get different output with different versions and different computers. Still one central question is whether the bug is only specific to the text handling (line spacing) or not. From my tests with your document so far I would bet on the former. 

I have extracted the text into one xopp-file 
[Input_Capture_Mode_Text.txt](https://github.com/xournalpp/xournalpp/files/4980924/Input_Capture_Mode_Text.txt)
and all the rest into another xopp file.
[...]

Issue #2073, opened 2020-06-17
Line spacing gets smaller during export to PDF


E-Wal:
**Affects versions**
 - OS: Linux (Ubuntu 18.04)
 - Desktop environment: Ubuntu Standard [Gnome-Wayland]
 - Version of libgtk: `xournalpp --version` crashed with unknown option
 - Version of Xournal++: 1.0.16, git commit 07cb970f
 - Installation method: snap store

**Describe the bug**
When exporting a document with textfield, the text field in the PDF is vertically smaller than it has been in Xournal++, whereas highlighting stays at it's place. That courses problems in long textfields: Words further down in the textfield seem to be highlighted in Xournal++, but are not highlighted in the exported PDF - instead the word one or two lines further down is highlighted...

**To Reproduce**
1. Write three words in a textfield and bring one of them (through blanklines) to the middle of the page and the last one down to the bottom.
2. Highlight each word.
3. Export to PDF.
4. See the difference: The highlighting stays on it's place, whereas the second and especially the third word appear further above.

**Expected behavior**
Highlighted words should be still highlighted after exporting to PDF.

**Screenshots of Problem**
![Auswahl_355](https://user-images.githubusercontent.com/66143769/84916509-7582c280-b0be-11ea-8165-7a91a989a816.png)
![Auswahl_356](https://user-images.githubusercontent.com/66143769/84916513-761b5900-b0be-11ea-8071-fcaad8593736.png)

Technius:
Duplicate of #1390.

Issue #1390, opened 2019-08-02
Text spaciing changes on export to PDF


markwmuller:
**Affects versions :**
 - OS: Linux
 - Version of Xournal++ current master (Latest commit a6276cd)

**Describe the bug**
The spacing of text entered using the "text" tool changes on export to PDF. See attached xournal file, and the resulting pdf. Note that the numbers are on either side of the lines in the xournal file, but no longer in the PDF. 

**To Reproduce**
Draw vertical lines with pen. Open text box, write numbers interspaced with spaces over each line. Export to PDF. See how numbers no longer align with lines. 

See attached files (zipped because .xopp upload not allowed): [textAndLines.zip](https://github.com/xournalpp/xournalpp/files/3462124/textAndLines.zip)

**Expected behavior**
PDF looks like xournal file.

Technius:
Do you know if this only affect text, or does it affect other objects such as images and latex?

markwmuller:
I only noticed with text, it came up because I was typing data into a PDF.

Technius:
This may be caused by changing DPI scaling without restarting the application. Does this occur if you open Xournal++, load the file, and then export the PDF without changing DPI settings?

Technius:
Interestingly enough, I can't reproduce this if I follow your steps.

But if I open the file you provided, I see the following in Xournal++:
![image](https://user-images.githubusercontent.com/1066652/67431647-93cd0880-f599-11e9-8e0e-3267bb791bcf.png)

Can you also insert an image and see if the problem persists? I want to make sure this issue is limited only to text before trying to figure it out. I suspect that system DPI scaling may be affecting Pango (which we use to render text).

markwmuller:
See attached. 

Image seems OK, text still odd. Confusingly, it worked OK the first time I tried it, then I rebooted and created a new file (which has original issue)
[spacing_img.zip](https://github.com/xournalpp/xournalpp/files/3764956/spacing_img.zip)

Here's the time it worked OK, in case you can diagnose something from that:
[GRID.zip](https://github.com/xournalpp/xournalpp/files/3764965/GRID.zip)

new-on-github:
Same problem here:
Build 1.1.0 dev from 31.03.2020

Don't know why, but the I think the text changes his stretching/size and the drawing stays the same.

By the way: Thanks for this amazing free software!

rickhg12hs:
[...]

Issue #1118, opened 2019-05-07
Text/font size doesn't scale with dpi calibration change in zoom settings.


CobaltSpace:
I am on version 1.0.10 on Arch Linux.

Technius:
Does it scale properly when you save the file, restart the program, and load the file?

CobaltSpace:
Yes.

This issue shows up with opening the same file on two different computers with different display dpi calibration.

VidalDuval:
I get a possibly related issue in version 1.0.12 (on Arch Linux as well). I discovered the effect after setting the Display DPI Calibration (Preferences -> Zoom) to 152. Annotated PDFs created previously with Xournal were then showing much bigger text size. When the DPI calibration is set to 100, text size looks the same as it did in Xournal.

alfazaz:
Same problem on uptodate archlinux... It's an annoying bug (when I open a last edited .xopp file with a different dpi calibration, text size is changed -> it's necessary to find the oldest dpi used to keep old texts in right place with right size...).

PR #2182, opened 2020-08-21
Make line spacing be the same in export and on screen

rolandlo:
This short PR aims to fix ~two annoying (high priority) bugs and another (yet unreported) minor bug~ the following bug:
- line spacing bug, leading to different results in screen view and pdf export of text elements:#1390,  #2073, #2104 (and #25)

The following bugs, that were previously part of this PR, have already been addressed in #2724
- dpi scaling bug, leading to different text size with different dpi settings: #1118
- search mark bug, leading to search marks that are one character too long (unreported)

The bugs are all easy to reproduce and the results can be compared with the fixed version. For instance take the following zip-file, consisting of a xopp-file with its pdf-export:
[TextLinesOfIncreasingSize.zip](https://github.com/xournalpp/xournalpp/files/5111212/TextLinesOfIncreasingSize.zip)

The columns are written with different font sizes (14.0 pt, 14.1 pt, 14.2 pt, ..., 15.0 pt from left to right). With the PR the screen output and the pdf export look the same: the columns get longer and longer from left to right with every 0.1 pt increase of the font size. With the current (unfixed) version, most columns will have the same height in the screen output. In the pdf export it is the same like in the PR. So clearly the screen output was flawed in the current version of Xournal++, not the pdf export.

### DIFF
--- a/src/view/TextView.cpp
+++ b/src/view/TextView.cpp
@@ -34,6 +34,10 @@ void TextView::updatePangoFont(PangoLayout* layout, const Text* t) {
     PangoFontDescription* desc = pango_font_description_from_string(t->getFontName().c_str());
     pango_font_description_set_absolute_size(desc, t->getFontSize() * PANGO_SCALE);
 
+#if PANGO_VERSION_CHECK(1, 48, 5)  // see https://gitlab.gnome.org/GNOME/pango/-/issues/499
+    pango_layout_set_line_spacing(layout, 1.0);
+#endif
+
     pango_layout_set_font_description(layout, desc);
     pango_font_description_free(desc);
 }


rolandlo:
Since there currently is no answer about the problem with empty lines from the pango team, I will have to postpone the line spacing fix to another PR and make the other 2 fixes first. However it may be better to provide some kind of migration tool for xopp-files which contain text written with other dpi-settings than 72. I think I will write a little Lua-Plugin, which does that migration task on a document basis.

LittleHuba:
A quick (and quite dirty) fix might be to temporarily put a whitespace in empty lines before rendering?

rolandlo:
Yes, that would fix the line spacing.

LittleHuba:
You should be able to add this in the [rendering method](https://github.com/xournalpp/xournalpp/blob/2cccc612aff0c93a348d573406e2e4e8cdf7a59b/src/view/TextView.cpp#L46). Once pango has it fixed we can remove it again.

rolandlo:
There are side effects for text editing though. Pressing right key arrow to advance to the next character would behave differently. Somehow Inkscape manages this differently, but it seems they make a lot of layout computations themselves, sidestepping pango. There text layout code is pretty complex.

LittleHuba:
If you just do it in the rendering method and do not apply it to the actual text but the local copy it should not make any differences.

rolandlo:
I will try.

Technius:
Awesome work, but how does this affect existing files (i.e. backwards compatibility)?

rolandlo:
Existing files written with dpi settings other than 72 will need to be fixed. My suggestion above: 

> However it may be better to provide some kind of migration tool for xopp-files which contain text written with other dpi-settings than 72. I think I will write a little Lua-Plugin, which does that migration task on a document basis.

How would you deal with this problem?

Technius:
I think your suggested solution sounds reasonable. But we probably won't want to include this PR in `stable` yet since a required migration would not be appropriate for a "minor" release.

rolandlo:
Ok, no problem. We'll keep it in the mind for a major release then.

Technius:
If we bump up the file format version, we could try auto-migrating files on older versions of the format upon confirmation by the user.

rolandlo:
> If we bump up the file format version, we could try auto-migrating files on older versions of the format upon confirmation by the user.

Maybe this is the way to go. However it won't be possible to figure out the scaling factor automatically with certainty. The user still needs to be asked about it, since he may have created different documents with different dpi settings (which are not recorded in the file). Moreover a user who never changed the standard dpi setting (from 72) will not need conversion and may be irritated by being asked about keeping the scaling factor for each document.

Febbe:
In my opinion, the dpi should be completely independent of the file format. It should be applied after rendering the document as scale factor.

rolandlo:
> In my opinion, the dpi should be completely independent of the file format. It should be applied after rendering the document as scale factor.

This is what my PR does. However currently this is not the case, so we need to take care of it in the migration step. Currently a font displayed with 18pt font size and 72 dpi setting will be stored as 18pt in the file (as it ought to be); with a 144dpi setting to achieve the same visible font size, the font is stored as 9pt size (which is wrong).

LittleHuba:
Could we provide a plugin that does the scaling adjustments if activated? That would actually leave it to the user if he wants the adjustment or not. I would then present a dialog to the user asking him to activate this or leave it deactivated.

rolandlo:
My idea was to provide a plugin for making the scaling adjustment once. If the user realizes that his font sizes are wrong, he should activate the plugin from the menu. The plugin shows a dialog asking for the scaling factor (and providing some information). It could suggest a scaling factor (current dpi / 72), but the user may have to correct it, because he may have created the document with a different dpi setting. Afterwards the user should save the file and doesn't need to use the plugin any more.

LittleHuba:
That is quite exactly what I meant. Users not interested or not requiring this plugin can simply deactivate it and will not be bothered with any dialogs.

Technius (approved):

g-kartik:
The problem is still there

rolandlo:
> The problem is still there

What Pango version do you use? 1.48.5 or higher is required.

g-kartik:
Yes pango is installed of version 1.48.7.

g-kartik:
@rolandlo can you please confirm whether you too face this issue?

rolandlo:
@g-kartik 
I'm on Ubuntu 21.04 and don't have Pango 1.48.5 or later installed. @tattsan what is your experience with line spacing on screen and in pdf? Is there any difference?

tattsan:
I haven't noticed any differerence. I think some sample files needed.

niknah:
I haven't noticed any difference in existing text boxes.  The spacing is more accurate in newly created text boxes, but not exactly right.  It's either too big or too small.  Samples attached.
[xournal_text_spacing.zip](https://github.com/xournalpp/xournalpp/files/7125821/xournal_text_spacing.zip)

PR #2185, opened 2020-08-25
Update WindowsBuild.md

rolandlo:
`make` is required to build the portaudio library (creating a `make.exe` copy of `mingw32-make.exe` doesn't work)


Febbe:
LGTM

PR #2186, opened 2020-08-26
Check if folder exists in ensure folder exists

aDogCalledSpot:
std::filesystem::create_directories returns false if no folders were created which can be the case if the folder already exists (which will be most of the time). This simply adds a check beforehand to ensure no errors are presented if none occured.


Febbe:
Thank you!

PR #2195, opened 2020-08-28
Apply to all pages

MichelReims:
When I choose "Apply to all pages" in the "Journal/paperback" menu, all pages  turn blank. I can't choose the same paperback for all page in an existing document


rolandlo:
That is not a pull request. It is a question/issue. So please use an issue to raise your problem or ask on gitter.im.

MichelReims:
Yes, I do

PR #2251, opened 2020-09-18
Plugin for cycling through a color list

### LINKED ISSUE #1835: Add color control in plugin interface
I want to change the marker color via hotkeys. Since there are not hotkeys/shortcuts I want to implement this in a plugin. However, plugins only can access actions defined in `actionPerformed` (`Color.cpp:392`). However, color isn't handled there: `// nothing to do here, the color toolbar item handles the color` and I can't find another way to access color control via plugin or shortcuts.

An extension of the plugin interface to allow for (marker) color control would be great.

* General hotkeys / shortcuts for controls such as color selection would be nice too and would make my plugin plan obsolete (in a good way).
* Instead of "properly"(?) extending the plugin interface one could just implement an extra function for marker color like was done for background color (`applib_changeCurrentPageBackground` in `luapi_application.h:185`).

**Additional context**
The reason I was trying to make this plugin is that I use a graphics tablet with stylus to write / make presentations now. I has buttons that I would like to use e.g. to cycle through colors or to change tools. 

Edit: I imagined something like this in `luapi_application.h`, I just don't have any idea where to get the necessary ColorToolItem from ...

    static int applib_changeColor(lua_State* L) {
        int color = 1;
        ColorToolItem* cti = ???
        cti->enableColor(color);
    
        return 1;
    }

rolandlo:
This has been on the wish list for quite a while, see #919 and #1835.
Closes #1835.

### DIFF
--- a/src/plugin/luapi_application.h
+++ b/src/plugin/luapi_application.h
@@ -272,6 +272,7 @@ static int applib_changeColor(lua_State* L) {
 
     if (toolType == TOOL_NONE) {
         g_warning("tool \"%s\" is not valid or no tool has been selected", toolTypeToString(toolType).c_str());
+        lua_pop(L, 3);
         return 0;
     }
 
@@ -288,6 +289,9 @@ static int applib_changeColor(lua_State* L) {
     } else {
         g_warning("tool \"%s\" has no color capability", toolTypeToString(toolType).c_str());
     }
+
+    // Make sure to remove all vars which are put to the stack before!
+    lua_pop(L, 3);
 
     return 1;
 }
--- a/src/plugin/luapi_application.h
+++ b/src/plugin/luapi_application.h
@@ -237,14 +237,14 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
 /**
  * Change color of a specified tool or of the current tool
  */
-static int applib_changeColor(lua_State* L) {
+static int applib_changeToolColor(lua_State* L) {
 
     // discard any extra arguments passed in
     lua_settop(L, 1);
     luaL_checktype(L, 1, LUA_TTABLE);
 
     lua_getfield(L, 1, "selection"); /* either true or false, for changing selection color
-                                       defaults to true*/
+                                       defaults to false*/
     lua_getfield(L, 1, "tool");      /* "pen", "hilighter", "text"
                                       "select_rect", "select_object", "select_region"
                                       if omitted, current Tool is used */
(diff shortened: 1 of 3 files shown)


src/plugin/luapi_application.h
```diff
@@ -233,13 +234,75 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
...
+    //   -2 = "pen"
+    //   -1 = 0xff0077
+
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    ToolHandler* toolHandler = ctrl->getToolHandler();
+
+    bool selection = false;
+    if (lua_isboolean(L, -3)) {
+        selection = lua_toboolean(L, -3);
```
Technius, line 264: Should report a warning if `selection` is a non-nil, non-boolean value.

src/plugin/luapi_application.h
```diff
@@ -233,13 +234,75 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
...
+
+    if (toolType == TOOL_NONE) {
+        g_warning("tool \"%s\" is not valid or no tool has been selected", toolTypeToString(toolType).c_str());
+        lua_pop(L, 3);
+        return 0;
+    }
+
+    int color = 0x000000;
+    if (lua_isinteger(L, -1)) {
+        color = lua_tointeger(L, -1);
```
Technius, line 281: Check if the color is valid RGBA.

Technius (approved):
Looks good so far, just a few minor comments:
* `changeColor` is too vague, I think the Lua function should be renamed to `changeToolColor`.
* Needs more error handling (see the lines I indicated)

src/plugin/luapi_application.h
rolandlo: RGB you mean?

rolandlo:
I have made the suggested changes. Can I squash and merge?

src/plugin/luapi_application.h
Technius: Yeah, RGB.

It's too easy to mix up the formats, so we should replace all (un)signed int colors with typedefs like `rgb_t` or `rgba_t` in the future.

Technius:
Yes, go for it! :)

lcswillems:
Hi, how can I customize the colors? I can't find the lua file on my computer.

rolandlo:
@lcswillems What OS do you use? Under Linux you should find it in `/usr/share/xournalpp/plugins/`

lcswillems:
That's right!!

So the path given here is wrong I think: https://xournalpp.github.io/guide/plugins/

PR #2252, opened 2020-09-19
Rewrite main to use the gtk3 initialisation instead of the deprecated gtk2 method

Febbe:
Since this is a huge change and this change should be merged into the next release we need to test it well.
Here a suggested list what's to test:

#### Test for Linux:
 - [x] Xournalpp starts (build_dir).
 - [x] Plugins are loaded (build_dir).
 - [x] All commands of the command-line-interface work.
 - [x] All icons are still displayed.
 - [x] everything works after installation

#### Test for Window:
 - [x] All commands of the command-line-interface work.
 - [x] All icons are still displayed.
 - [x] Xournalpp starts.
 - [x] everything works after installation
 - [x] Plugins are loaded

#### Test for Mac:
 - ~~All commands of the command-line-interface work.~~
 - ~~All icons are still displayed.~~
 - ~~Xournalpp starts.~~
 - ~~everything works after installation~~
 - ~~Plugins are loaded~~
 - ~~Append this list if desired.~~

This may fix problems with icons.
It will use the config folder name xournalpp instead of com.github.xournalpp.xournalpp

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -1,9 +1,11 @@
 #include "XournalMain.h"
 
+#include <algorithm>
 #include <memory>
 
 #include <glib/gstdio.h>
 #include <gtk/gtk.h>
+#include <gui/toolbarMenubar/model/ToolbarColorNames.h>
 #include <libintl.h>
 
 #include "control/jobs/ImageExport.h"
@@ -11,7 +13,6 @@
 #include "gui/GladeSearchpath.h"
 #include "gui/MainWindow.h"
 #include "gui/XournalView.h"
-#include "gui/toolbarMenubar/model/ToolbarColorNames.h"
 #include "pdf/base/XojPdfExport.h"
 #include "pdf/base/XojPdfExportFactory.h"
 #include "undo/EmergencySaveRestore.h"
@@ -30,15 +31,33 @@
 #if __linux__
 #include <libgen.h>
 #endif
+namespace {
 
-#include <algorithm>  // std::sort
+constexpr auto APP_FLAGS = GApplicationFlags(G_APPLICATION_SEND_ENVIRONMENT | G_APPLICATION_NON_UNIQUE);
 
+/// Configuration migration status.
+enum class MigrateStatus {
+    NotNeeded,  ///< No migration was needed.
+    Success,    ///< Migration was carried out successfully.
+    Failure,    ///< Migration failed. */
+};
 
-XournalMain::XournalMain() = default;
+struct MigrateResult {
+    MigrateStatus status{};
+    std::string message;  ///< Any additional information about the migration status.
+};
 
-XournalMain::~XournalMain() = default;
+auto migrateSettings() -> MigrateResult;
 
-void XournalMain::initLocalisation() {
+void checkForErrorlog();
+void checkForEmergencySave(Control* control);
+
+auto exportPdf(const char* input, const char* output) -> int;
+auto exportImg(const char* input, const char* output) -> int;
+
+void initResourcePath(GladeSearchpath* gladePath, const gchar* relativePathAndFile, bool failIfNotFound = true);
+
+void initLocalisation() {
 #ifdef ENABLE_NLS
 
 #ifdef _WIN32
@@ -61,56 +80,59 @@ void XournalMain::initLocalisation() {
 
 #endif  // ENABLE_NLS
 
-    // Not working on Windows! Working on Linux, but not sure if it's needed
-#ifndef _WIN32
+    // Not working on GNU g++(mingww) forWindows! Only working on Linux/macOS and with msvc
     try {
         std::locale::global(std::locale(""));  // "" - system default locale
     } catch (std::runtime_error& e) {
-        g_warning("XournalMain: System default locale could not be set.\nCaused by: %s", e.what());
+        g_warning("XournalMain: System default locale could not be set.\n - Caused by: %s\n - Note that it is not "
+                  "supported to set the locale using mingw-w64 on windows.\n - This could be solved by compiling "
+                  "xournalpp with msvc",
+                  e.what());
     }
-#endif
     std::cout.imbue(std::locale());
 }
 
-XournalMain::MigrateResult XournalMain::migrateSettings() {
+auto migrateSettings() -> MigrateResult {
     fs::path newConfigPath = Util::getConfigFolder();
 
     if (!fs::exists(newConfigPath)) {
-        std::array<fs::path, 1> oldPaths = {
+        std::array oldPaths = {
+                Util::getConfigFolder().parent_path() /= "com.github.xournalpp.xournalpp",
+                Util::getConfigFolder().parent_path() /= "com.github.xournalpp.xournalpp.exe",
                 fs::u8path(g_get_home_dir()) /= ".xournalpp",
         };
         for (auto const& oldPath: oldPaths) {
             if (!fs::is_directory(oldPath)) {
                 continue;
             }
-            g_message("Migrating configuration from %s to %s", oldPath.c_str(), newConfigPath.c_str());
+            g_message("Migrating configuration from %s to %s", oldPath.string().c_str(),
+                      newConfigPath.string().c_str());
             Util::ensureFolderExists(newConfigPath.parent_path());
             try {
                 fs::copy(oldPath, newConfigPath, fs::copy_options::recursive);
                 constexpr auto msg = "Due to a recent update, Xournal++ has changed where it's configuration files are "
                                      "stored.\nThey have been automatically copied from\n\t{1}\nto\n\t{2}";
-                return {MigrateStatus::Success,
-                        FS(_F(msg) % oldPath.u8string().c_str() % newConfigPath.u8string().c_str())};
+                return {MigrateStatus::Success, FS(_F(msg) % oldPath.u8string() % newConfigPath.u8string())};
             } catch (fs::filesystem_error const& except) {
                 constexpr auto msg =
                         "Due to a recent update, Xournal++ has changed where it's configuration files are "
                         "stored.\nHowever, when attempting to copy\n\t{1}\nto\n\t{2}\nmigration failed:\n{3}";
                 g_message("Migration failed: %s", except.what());
                 return {MigrateStatus::Failure,
-                        FS(_F(msg) % oldPath.u8string().c_str() % newConfigPath.u8string().c_str() % except.what())};
+                        FS(_F(msg) % oldPath.u8string() % newConfigPath.u8string() % except.what())};
             }
         }
     }
     return {MigrateStatus::NotNeeded, ""};
 }
 
-void XournalMain::checkForErrorlog() {
+void checkForErrorlog() {
     fs::path errorDir = Util::getCacheSubfolder(ERRORLOG_DIR);
     if (!fs::exists(errorDir)) {
         return;
     }
 
-    vector<fs::path> errorList;
+    std::vector<fs::path> errorList;
     for (auto const& f: fs::directory_iterator(errorDir)) {
         if (f.is_regular_file() && f.path().stem() == "errorlog") {
             errorList.emplace_back(f);
@@ -127,11 +149,9 @@ void XournalMain::checkForErrorlog() {
                     _("There is an errorlogfile from Xournal++. Please send a Bugreport, so the bug may be fixed.") :
                     _("There are errorlogfiles from Xournal++. Please send a Bugreport, so the bug may be fixed.");
     msg += "\n";
-#if defined(GIT_BRANCH) && defined(GIT_REPO_OWNER)
-    msg += FS(_F("You're using {1}/{2} branch. Send Bugreport will direct you to this repo's issue tracker.") %
-              GIT_REPO_OWNER % GIT_BRANCH);
+    msg += FS(_F("You're using \"{1}/{2}\" branch. Send Bugreport will direct you to this repo's issue tracker.") %
+              GIT_ORIGIN_OWNER % GIT_BRANCH);
     msg += "\n";
-#endif
     msg += FS(_F("The most recent log file name: {1}") % errorList[0].string());
 
     GtkWidget* dialog = gtk_message_dialog_new(nullptr, GTK_DIALOG_MODAL, GTK_MESSAGE_QUESTION, GTK_BUTTONS_NONE, "%s",
@@ -144,6 +164,7 @@ void XournalMain::checkForErrorlog() {
     gtk_dialog_add_button(GTK_DIALOG(dialog), _("Cancel"), 5);
 
     int res = gtk_dialog_run(GTK_DIALOG(dialog));
+    gtk_widget_destroy(dialog);
 
     auto const& errorlogPath = Util::getCacheSubfolder(ERRORLOG_DIR) / errorList[0];
     if (res == 1)  // Send Bugreport
@@ -159,19 +180,17 @@ void XournalMain::checkForErrorlog() {
     } else if (res == 4)  // Delete Logfile
     {
         if (!fs::exists(errorlogPath)) {
-            string msg = FS(_F("Errorlog cannot be deleted. You have to do it manually.\nLogfile: {1}") %
-                            errorlogPath.u8string());
+            msg = FS(_F("Errorlog cannot be deleted. You have to do it manually.\nLogfile: {1}") %
+                     errorlogPath.u8string());
             XojMsgBox::showErrorToUser(nullptr, msg);
         }
     } else if (res == 5)  // Cancel
     {
         // Nothing to do
     }
-
-    gtk_widget_destroy(dialog);
 }
 
-void XournalMain::checkForEmergencySave(Control* control) {
+void checkForEmergencySave(Control* control) {
     auto file = Util::getConfigFile("emergencysave.xopp");
 
     if (!fs::exists(file)) {
@@ -206,13 +225,12 @@ void XournalMain::checkForEmergencySave(Control* control) {
     gtk_widget_destroy(dialog);
 }
 
-auto XournalMain::exportImg(const char* input, const char* output) -> int {
+auto exportImg(const char* input, const char* output) -> int {
     LoadHandler loader;
 
     Document* doc = loader.loadDocument(input);
     if (doc == nullptr) {
         g_error("%s", loader.getLastError().c_str());
-        return -2;
     }
 
     fs::path const path(output);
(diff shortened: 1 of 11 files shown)


src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
```diff
@@ -1,50 +1,26 @@
...
         g_error_free(error);
         return;
     }
-
-    g_key_file_set_string(this->config, "info", "about", "Xournalpp custom color names");
+    g_key_file_set_string(config, "info", "about", "Xournalpp custom color names");
 }
 
-void ToolbarColorNames::saveFile(fs::path const& file) {
+void safe(fs::path const& file, GKeyFile* config, std::unordered_map<Color, std::string>& predefinedColorNames) {
```
Technius, line 21: Typo, should be `save`. Also needs to be marked `static`.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
```diff
@@ -1,50 +1,26 @@
...
-    return *instance;
-}
-
-void ToolbarColorNames::freeInstance() {
-    delete instance;
-    instance = nullptr;
-}
-
-void ToolbarColorNames::loadFile(fs::path const& file) {
+void load(fs::path const& file, GKeyFile* config, std::unordered_map<Color, std::string>& predefinedColorNames) {
```
Technius, line 11: Mark as `static`.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
```diff
@@ -56,22 +32,51 @@ void ToolbarColorNames::saveFile(fs::path const& file) {
...
+
+ToolbarColorNames::ToolbarColorNames(fs::path readFile): file(std::move(readFile)) {
+    this->config = g_key_file_new();
+    g_key_file_set_string(this->config, "info", "about", "Xournalpp custom color names");
+    initPredefinedColors();
+    load(this->file, this->config, predefinedColorNames);
+}
+
+ToolbarColorNames::~ToolbarColorNames() {
+    saveFile(file);
```
Technius, line 44: Saving the file should be explicit in the application shutdown--don't do it here in the destructor.

src/util/PathUtil.cpp
```diff
@@ -174,12 +177,15 @@ auto Util::getCacheSubfolder(const fs::path& subfolder) -> fs::path {
...
     auto p = fs::u8path(g_get_user_data_dir());
+    assert(!p.empty());
     p /= g_get_prgname();
     p /= subfolder;
 
     return Util::ensureFolderExists(p);
 }
 
+
+#include <iostream>
```
Technius, line 188: This should go at the top of the file.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
```diff
@@ -56,22 +32,51 @@ void ToolbarColorNames::saveFile(fs::path const& file) {
     g_free(data);
 }
 
+
+ToolbarColorNames::ToolbarColorNames(fs::path readFile): file(std::move(readFile)) {
+    this->config = g_key_file_new();
+    g_key_file_set_string(this->config, "info", "about", "Xournalpp custom color names");
+    initPredefinedColors();
+    load(this->file, this->config, predefinedColorNames);
+}
```
Technius, line 41: This should be a secondary constructor. What if we don't want to load toolbar colors from a file?

src/control/XournalMain.cpp
```diff
@@ -560,3 +415,146 @@ void XournalMain::initResourcePath(GladeSearchpath* gladePath, const gchar* rela
...
+    GtkApplication* app = gtk_application_new("com.github.xournalpp.xournalpp", APP_FLAGS);
+
+    g_signal_connect(app, "activate", G_CALLBACK(&on_activate), &app_data);
+    g_signal_connect(app, "command-line", G_CALLBACK(&on_command_line), &app_data);
+    g_signal_connect(app, "open", G_CALLBACK(&on_open_files), &app_data);
+    g_signal_connect(app, "startup", G_CALLBACK(&on_startup), &app_data);
+    g_signal_connect(app, "shutdown", G_CALLBACK(&on_shutdown), &app_data);
+    g_signal_connect(app, "handle-local-options", G_CALLBACK(&on_handle_local_options), &app_data);
+
+    std::array<GOptionEntry, 6> options = {
```
Technius, line 549: Use a plain const array, no advantage to a C++ array here. We would have to change the size here every time we add/remove flags.

src/control/XournalMain.cpp
```diff
@@ -276,238 +297,73 @@ auto XournalMain::exportPdf(const char* input, const char* output) -> int {
...
-        g_printf("└──%s: %d.%d.%d \n", "libgtk", gtk_get_major_version(), gtk_get_minor_version(),
-                 gtk_get_micro_version());
-        return 0;
-    }
-
-    // Checks for input method compatibility
-
+    std::unique_ptr<GladeSearchpath> gladePath;
+    std::unique_ptr<Control> control;
+    std::unique_ptr<MainWindow> win;
```
Technius, line 320: These don't need to be pointers.

Technius (changes requested):
Didn't look at everything yet, but skimmed the code briefly:
* Did not pass CI on MacOS.
* The functions in `XournalMain` either need to be `static` or namespaced.
* `ToolbarColorNames` should not carry `file` within it, it's unnecessary complexity. File IO should be explicit in the application.

src/control/XournalMain.cpp
Febbe: The problem is, that they must initialized delayed and the pointers must persist across the signals.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
Febbe: > Also needs to be marked `static`.

The function is already static.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
Febbe: This is a singelton class, we always want to load the same file. I could even hardcode it into the class itself.

src/CMakeLists.txt
```diff
@@ -17,10 +17,7 @@ add_subdirectory (xoj-preview-extractor)
 add_definitions (-g -Wreturn-type -Wuninitialized -Wunused-value -Wunused-variable)
 
 if (WIN32)
-  # optimize for size (the Windows .exe is really big)
-  set (xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-Os -s")
 elseif (APPLE OR LINUX)
-  # Nothing to do for APPLE and linux
```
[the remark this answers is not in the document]
Febbe: Thanks, wasn't my intend.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
Febbe: renamed and namespaced.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
Febbe: namespaced instead

src/util/PathUtil.cpp
Febbe: removed, was only for debugging

src/control/XournalMain.cpp
Febbe: The disadvantage is, that clang-tidy complains. Marking everything as `// NOLINT(<xyz>)` is only applicable, if it can't be fixed due to a dependency. I would propose to either wait till c++20 to use `std::make_array` or implement it.

Febbe:
Thanks @Technius for the review, also thanks for the hint, that static does a lot more than I thought. 
I decided to use anonymous namespaces to fulfill the purpose.

I have some design questions about `ToolbarColorNames`:
This class is used to initialize a Singleton and before I changed the signature it has been manually initialized to explicitly load the file. Hence this always happened, I don't see the reason why we should not inline the code into the constructor.
If it's required to create an instance without the colornames.ini file we still could create one. But now the constructor will be dead code.

Regarding the save: There is indeed no reason to do fileIO in the destructor except, to ensure, that the file is saved on application shutdown. Maybe we should add the safe also to the crash handling.

src/control/XournalMain.cpp
Technius: Looks like the easiest option is to just leave it as `std::array`.

Technius:
> This class is used to initialize a Singleton

Singletons are a maintenance nightmare for a code base as large as Xournal++ (as we may have seen with other things we've refactored...). `ToolbarColorNames` should be refactored to not be a singleton, but this is a secondary concern that should be addressed in a future PR.

> Regarding the save: There is indeed no reason to do fileIO in the destructor except, to ensure, that the file is saved on application shutdown. Maybe we should add the safe also to the crash handling.

The current behavior is to save the color names at application shutdown (bottom of `XournalMain::run`), so we could just save right before application shutdown in `Control::quit`. It's not critical enough to justify saving during crash handling.

LittleHuba:
> The current behavior is to save the color names at application shutdown (bottom of XournalMain::run), so we could just save right before application shutdown in Control::quit. It's not critical enough to justify saving during crash handling.

My two cents on this are that you should keep the crash handler as simple and small as possible. You never know what part did crash the application and rerunning that part potentially in the crash handler would cause it to fail as well. Hence only the file should be tried to be saved there and nothing else. Like in a emergency you safe people and not their cars and stuff.

Febbe:
I think `ToolbarColornames` should be moved into Settings. 
But this is a job for a followup PR.
Therefore, I think it is irrelevant, if I remove the filename again, especially because it is currently only used to load and safe from the same file.
@LittleHuba @Technius I already made that change. The `ToolbarColorNames` class does not safe in the destructor anymore.

src/control/XournalMain.cpp
Febbe: Turns out writing `std::array options = <initializer-list>` does the job! :)

Febbe:
Command line option -n does not work. We need to check if this is a bug in the previous version.

Febbe:
Ignore Clang Format CI check, the old clang-format is bugged for non parametrized function parameters.

Febbe:
@rolandlo or @LittleHuba can you test the windows build for me again?

LittleHuba:
I'll take a look later this week.

Febbe:
Maybe we should Hard-code the application-name into Xournal++.
Viewable name: Xournal++
Native name: xournalpp
AppId: com.github.xournalpp.xournalpp
And then let the gtk options AND paths depend on them, instead of setting the paths by requesting them from the gtk-application, after setting them there first.

rolandlo:
Compiling this PR with MinGW64 under Windows I get a lot of compile warnings like `warning: unknown escape sequence '\9' constexpr auto GIT_COMMIT_ID="\91263c3cc HEAD";`

It does compile though. During startup I get some warnings as well:

````term
** (xournalpp.exe:1612): WARNING **: 11:33:05.071: XournalMain: System default locale could not be set.
Caused by: locale::facet::_S_create_c_locale name not valid
Note that this is not supported using mingw-w64 on windows.
Try to compile xournalpp with msvc

** (xournalpp.exe:1612): WARNING **: 11:33:05.087: configfile does not exist C:\Users\Roland\AppData\Local\xournalpp\settings.xml

** Message: 11:33:05.118: Using default latex template in C:\msys64\home\Roland\Pulls\xournalpp\resources\default_template.tex

** (xournalpp.exe:1612): WARNING **: 11:33:05.241: No device found. Is Xournal++ running in debugger / Eclipse...?
Probably this is the reason for not finding devices!


** (xournalpp.exe:1612): WARNING **: 11:33:05.256: Failed to load "colornames.ini" (C:\Users\Roland\AppData\Local\xournalpp\colornames.ini): No such file or directory


(xournalpp.exe:1612): Gtk-WARNING **: 11:33:05.272: Theme parsing error: <data>:1:22: Junk at end of value for font-family

(xournalpp.exe:1612): Pango-WARNING **: 11:33:05.657: couldn't load font "C059 Not-Rotated 8", falling back to "Sans Not-Rotated 8", expect ugly output.
````

What exactly do you want me to test, @Febbe?

**Comment by rolandlo** [comment]
Under Input Devices I only see this:
grafik

**Comment by LittleHuba** [comment]
> Under Input Devices I only see this:
> grafik

That is unfortunately normal for Windows. GTK only exposes the aggregated system pointer instead of the actual devices.

**Comment by Febbe** [comment]
There is a list on the top of the page, but I realized, that we need to check if the config folder is moved from com.github.xournalpp.xournalpp(.exe) to just xournalpp

**Comment by rolandlo** [comment]
Seems I got some warnings that only appeared on first startup. Now I have fewer warnings:
````term
$ build/src/xournalpp

** (xournalpp.exe:13420): WARNING **: 20:56:50.896: XournalMain: System default locale could not be set.
Caused by: locale::facet::_S_create_c_locale name not valid
Note that this is not supported using mingw-w64 on windows.
Try to compile xournalpp with msvc

(xournalpp.exe:13420): Gtk-WARNING **: 20:56:51.408: Theme parsing error: <data>:1:22: Junk at end of value for font-family

(xournalpp.exe:13420): Pango-WARNING **: 20:56:52.837: couldn't load font "C059 Not-Rotated 8", falling back to "Sans Not-Rotated 8", expect ugly output.
```` 
The icons are all there. No font is selected. After setting a font and font size, text works. Selecting the text gives a strange result though:
grafik
Here I wrote "Tagi" with Sans Normal 12. It is displayed normally, when not selected.
grafik

Plugins work fine. Pen, highlighter, eraser work, all drawing types work (both with stylus and mouse).
Saving and loading a xopp-file works. Inserting an image works. 
Hand tool, vertical space and selection work.

When opening the preferences, I get the error message:
````term
** (xournalpp.exe:6524): WARNING **: 21:21:03.127: filesystem error: directory iterator cannot open directory: No such file or directory [../share/locale/]
```` 
which does not look good.

The Dpi settings ruler does not work properly. The Dpi remains at 72, but the ruler marks change.

I haven't tested LaTeX (currently not installed on my Windows partition).

I had some problems saving a pdf as xopp-file and then  opening it. It displayed the error message, that I needed to replace the background.

The pdf and image crerating form the terminal work. I haven't tested the gtk- and gapplication options. 

I also haven't installed xournalpp yet.

**Comment by rolandlo** [comment]
Running the build-setup.sh script, I got some error messages:

````term
Copy locale zh
cp: der Aufruf von stat für '/usr/share/locale/zh/LC_MESSAGES/glib20.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gdk-pixbuf.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gtk30.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gtk30-properties.mo' ist nicht möglich: No such file or directory
Copy locale zh_TW
````

and a warning:

````term
Processed 1 file, writing output (x86-ansi):
warning 7998: ANSI targets are deprecated
```` 

**After installation:**

After installing via xournalpp-setup.exe I get the following message on first startup: 
grafik

Fonts seem to work fine. Translations work fine. Annotating a pdf, saving the annotations as a xopp-file and loading the xopp file seem to work fine.

I can't find the global LaTeX template file. Is it not included in the windows installer?
There are no plugins listed in the Plugin manager. If I start from the MinGW-console I see the error message:
grafik

Other than that I haven't noticed differences between the installed version and the version run from the build directory.

Recording does not work. It also does not work when running from the build directory:
grafik
grafik

**Comment by rolandlo** [comment]
One more warning that I receive on startup of the installed version:
grafik

Moreover if I open a file with the installed version from the MinGW command line or via double clicking the file in the file manager I get this window:
grafik
After clicking OK, it loads normally. 

Similarly if I use the -p option to export to pdf or the -i option to export as an image I get three debug info windows, that I must close before the export is made. Exporting works.

The -n option (or --page) option does not seem to work. I am not quite sure about the syntax though (`--page=2` or `-page 2`, `-n=2` or `-n 2`)

If I use the --name option like ` bin/xournalpp.exe --name="MyTestName"`
(which I expected to only change the window title) I get an error message like this:

grafik

The --g-fatal-warnings option does what it promises to do (it quits the application after the first warning).

I am not sure how I should test other command line options like --class, --gdk-debug, --gdk-no-debug, --gtk-module, --gtk-debug, --gtk-no-debug and --gapplication-service.

**Comment by Febbe** [comment]
Do not test the glib and gapplication command line options, since we probably can't fix them if they are wrong. Also they should generally do what they promise.

**Comment by Febbe** [comment]
The page command was also an issue before. Can you test the original version for those issues? We should open additionally issues for those who were there befor

**Comment by LittleHuba** [comment]
```
Copy locale zh
cp: der Aufruf von stat für '/usr/share/locale/zh/LC_MESSAGES/glib20.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gdk-pixbuf.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gtk30.mo' ist nicht möglich: No such file or directory
cp: der Aufruf von stat für '/mingw64/share/locale/zh/LC_MESSAGES/gtk30-properties.mo' ist nicht möglich: No such file or directory
Copy locale zh_TW
```

That is not related to the fixes of @Febbe. It is rather just a sign for a version of glib and GTK missing translations for zh_TW locale. The users installing Xournal++ on their systems with these locales will have to deliver them some other way if they require it.

LittleHuba:
Missing files are most likely an issue of the installer script. It has to be extended to include the additional files.

rolandlo:
> The page command was also an issue before. Can you test the original version for those issues? We should open additionally issues for those who were there befor

The page command is working in version 1.0.18 and 1.0.19 (at least on Linux, but probably on Windows as well). It does not work in version 1.1.0+dev (same on Linux and on Windows). It already didn't work in the June 12 2020 version (git commit 8afb3841). I will file a bug report.

Febbe:
> > The page command was also an issue before. Can you test the original version for those issues? We should open additionally issues for those who were there befor
> 
> The page command is working in version 1.0.18 and 1.0.19 (at least on Linux, but probably on Windows as well). It does not work in version 1.1.0+dev (same on Linux and on Windows). It already didn't work in the June 12 2020 version (git commit [8afb384](https://github.com/xournalpp/xournalpp/commit/8afb384191d1b546312c5e1b98ac7d5a80bf631c)). I will file a bug report, see #2445.

I'll hide this, since it is a regression from another bug.

Febbe:
Ok tested the points @rolandlo found, but I couldn't reproduce any of them. Neither the locale issue. The mingw-w64 team may have fixed that already. Also, our own issues also seem to be fixed.

PR #2268, opened 2020-09-22
Fixup #2254

### LINKED ISSUE #2254: Short strokes can't be erased.
This is a regression with the commit 6e5469874290871c501f14dbbca36099e96231d8 .

**Affects versions :**
 - OS: Arch Linux
 - Version of Xournal++: 1.1.0+dev  cdebec61151f968e6fafb2074be182aff970b0af

**Describe the bug**
Small strokes can't be erased.

**To Reproduce**
1. Start with Pen tool.
2. Create extremely short strokes.
3. Try to erase them with an eraser tool (in 'standard" or "delete stroke"  mode).
4. You can't erase them, if the strokes are sufficiently short.

rolandlo:
Fixes #2254

### DIFF
--- a/src/model/Element.cpp
+++ b/src/model/Element.cpp
@@ -75,7 +75,14 @@ void Element::setColor(Color color) { this->color = color; }
 auto Element::getColor() const -> Color { return this->color; }
 
 auto Element::intersectsArea(const GdkRectangle* src) -> bool {
-    GdkRectangle rect = {gint(getX()), gint(getY()), gint(getElementWidth()), gint(getElementHeight())};
+    // compute the smallest rectangle with integer coordinates containing the bounding box and having width, height > 0
+    auto x = getX();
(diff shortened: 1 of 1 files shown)


Febbe (commented):
The thing is, that the bounding box will be too big for even numbers of x, y, width, height. This might be insignificant, but also the Element height or width should not be 0 so just floor the value should be sufficient. I added the `std::max(1, x2 - x1)` in my suggestion to ensure, that my assumption, that the height or width is actually greater zero is correct.

rolandlo:
You meant for "integer numbers" (not "even numbers"), right? I am happy with your change. The code should do what it promises to do in the comment.

Febbe:
> You meant for "integer numbers" (not "even numbers"), right? I am happy with your change. The code should do what it promises to do in the comment.

right.
You can rebase-merge after the squash

Issue #2267, opened 2020-09-22
Breaks on i386 architectrures: non-constant-expression cannot be narrowed from type 'unsigned long long' to 'size_t'


yurivict:
```
/wrkdirs/usr/ports/graphics/xournalpp/work/xournalpp-cdebec61151f968e6fafb2074be182aff970b0af/src/util/audio/VorbisProducer.cpp:28:33: error: non-constant-expression cannot be narrowed from type 'unsigned long long' to 'size_t' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]
        size_t const bufferSize{1024ULL * sfInfo.channels};
                                ^~~~~~~~~~~~~~~~~~~~~~~~~
/wrkdirs/usr/ports/graphics/xournalpp/work/xournalpp-cdebec61151f968e6fafb2074be182aff970b0af/src/util/audio/VorbisProducer.cpp:28:33: note: insert an explicit cast to silence this issue
        size_t const bufferSize{1024ULL * sfInfo.channels};
                                ^~~~~~~~~~~~~~~~~~~~~~~~~
                                static_cast<size_t>(     )
1 error generated.
```

Log: http://beefy17.nyi.freebsd.org/data/head-i386-default/p549547_s365984/logs/xournalpp-1.0.18.20200819.log (IPv6 URL)

FreeBSD 11/12/13
clang-8/9/10/11

Febbe:
Xournal++ is only supported on 64bit machines. Even if you can compile it, it will show an empty document. But you can provide a very simple PR to fix that: `size_t const bufferSize{size_t(1024U) * sfInfo.channels};`

yurivict:
https://github.com/xournalpp/xournalpp/pull/2270

PR #2270, opened 2020-09-22
Fix build on i386 systems.

yurivict:


Febbe:
Was this really the only cause?

Febbe:
Thank you for your PR

yurivict:
Thank you for merging this PR so quickly.

I'll open more PRs in case there would be other i386-related failures.

Febbe:
Fixed #2267

PR #2276, opened 2020-09-25
Fixed some file_chooser issues including a wrong recent path for image import.

### LINKED ISSUE #2266: Default location for image import (?)
When importing an image, I now get "Folder contents cannot be displayed" and I have to walk down the tree from my home directory to find the desired image.  This occurs even if I launch xournal++ from the desired working directory.

Previous behavior was (if I recall correctly) to open the working directory.

It would also be nice if one could specify a default image working directory.

Febbe:
- Fixes #2266
 - gtk_file_chooser_set_x got utf8 encoded paths, but only supports local encoded paths
 - replaced gtk_file_chooser_set_x_uri with gtk_file_chooser_set_x on several places

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2258,12 +2258,12 @@ auto Control::showSaveDialog() -> bool {
 
     this->doc->lock();
     auto suggested_folder = this->doc->createSaveFolder(this->settings->getLastSavePath());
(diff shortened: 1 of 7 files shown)


Febbe:
Will merge in 12h to 16h.

PR #2277, opened 2020-09-25
fix copy selection when rotated and scaled

### LINKED ISSUE #2090: Rotation of shapes if not copied
**Affects versions :**
 - OS: Debian
 - Desktop environment: X11
 - Which version of libgtk do you use: I do not know; It is not given in the about window; `xournalpp --version` errors with
```
** (xournalpp:377189): ERROR **: 09:37:11.507: Unknown option --version
Trace/breakpoint trap
```
 - Installation method: apt repo

**Describe the bug**
Create a rectangle. Select it, e.g. with a lasso selection. Rotate it by 45°. Copy it. paste it -> The copy is not rotated.

**Expected behavior**
The copied rectangle should be rotated.

Thanks!

rolandlo:
Fixes #2090

### DIFF
--- a/src/control/tools/EditSelectionContents.cpp
+++ b/src/control/tools/EditSelectionContents.cpp
@@ -62,10 +62,25 @@ void EditSelectionContents::addElement(Element* e, Layer::ElementIndex order) {
     }
 }
 
+void EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, Layer::ElementIndex>> newInsertOrder) {
+    this->selected.clear();
+    this->selected.reserve(newInsertOrder.size());
+    std::transform(begin(newInsertOrder), end(newInsertOrder), std::back_inserter(this->selected),
+                   [](auto const& e) { return e.first; });
+    this->insertOrder = std::move(newInsertOrder);
+}
+
 /**
- * Returns all containing elements of this selections
+ * Returns all containing elements of this selection
  */
 auto EditSelectionContents::getElements() -> vector<Element*>* { return &this->selected; }
+
+/**
+ * Returns the insert order of this selection
+ */
+auto EditSelectionContents::getInsertOrder() const -> std::deque<std::pair<Element*, Layer::ElementIndex>> const& {
+    return this->insertOrder;
+}
 
 /**
  * Sets the tool size for pen or eraser, returs an undo action
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -16,6 +16,7 @@
 #include "serializing/ObjectOutputStream.h"
 #include "undo/ColorUndoAction.h"
 #include "undo/FontUndoAction.h"
+#include "undo/InsertUndoAction.h"
 #include "undo/SizeUndoAction.h"
 #include "undo/UndoRedoHandler.h"
 
@@ -293,9 +294,16 @@ void EditSelection::addElement(Element* e, Layer::ElementIndex order) {
 }
 
 /**
- * Returns all containing elements of this selections
+ * Returns all containing elements of this selection
  */
 auto EditSelection::getElements() -> vector<Element*>* { return this->contents->getElements(); }
+
+/**
+ * Returns the insert order of this selection
+ */
+auto EditSelection::getInsertOrder() const -> std::deque<std::pair<Element*, Layer::ElementIndex>> const& {
+    return this->contents->getInsertOrder();
+}
 
 /**
  * Finish the current movement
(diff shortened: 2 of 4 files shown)


src/control/tools/EditSelectionContents.cpp
```diff
@@ -62,11 +62,27 @@ void EditSelectionContents::addElement(Element* e, Layer::ElementIndex order) {
...
 }
 
+void EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, Layer::ElementIndex>> newInsertOrder) {
+    std::vector<Element*> newElements;
+    for (auto [e, index]: newInsertOrder) {
+        newElements.emplace_back(e);
+    }
+    this->selected = newElements;
+    this->insertOrder = newInsertOrder;
+}
```
Febbe, line 72:
```suggestion
void EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, Layer::ElementIndex>> newInsertOrder) {
    this->selected.clear();
    this->selected.reserve(newInsertOrder.size());
    std::transform(begin(newInsertOrder), end(newInsertOrder), std::back_inserter(this->selected), [](auto const& e,auto const&){return e;});
    this->insertOrder = std::move(newInsertOrder);
}
```

Febbe (commented):
In general this complete structure needs to be refactored. A selection should not take ownership of the elements, but that's another topic.

rolandlo:
Thanks for the review and the improvements @Febbe.
The `std::transform` bit doesn't compile for me (using gcc 9.2.1). 

```
 /usr/include/c++/9/bits/stl_algo.h: In instantiation of ‘_OIter std::transform(_IIter, _IIter, _OIter, _UnaryOperation) [with _IIter = std::_Deque_iterator<std::pair<Element*, long int>, std::pair<Element*, long int>&, std::pair<Element*, long int>*>; _OIter = std::back_insert_iterator<std::vector<Element*> >; _UnaryOperation = EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, long int> >)::<lambda(const auto:1&, auto:2)>]’:
[build] ../src/control/tools/EditSelectionContents.cpp:69:63:   required from here
[build] /usr/include/c++/9/bits/stl_algo.h:4337:24: error: no match for call to ‘(EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, long int> >)::<lambda(const auto:1&, auto:2)>) (std::pair<Element*, long int>&)’
[build]  4337 |  *__result = __unary_op(*__first);
[build]       |              ~~~~~~~~~~^~~~~~~~~~
[build] ../src/control/tools/EditSelectionContents.cpp:69:20: note: candidate: ‘template<class auto:1, class auto:2> EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, long int> >)::<lambda(const auto:1&, auto:2)>’
[build]    69 |                    [](auto const& e, auto const) { return e; });
[build]       |                    ^
[build] ../src/control/tools/EditSelectionContents.cpp:69:20: note:   template argument deduction/substitution failed:
[build] In file included from /usr/include/c++/9/algorithm:62,
[build]                  from ../src/util/Rectangle.h:14,
[build]                  from ../src/model/Element.h:22,
[build]                  from ../src/control/tools/EditSelectionContents.h:19,
[build]                  from ../src/control/tools/EditSelectionContents.cpp:1:
[build] /usr/include/c++/9/bits/stl_algo.h:4337:24: note:   candidate expects 2 arguments, 1 provided
[build]  4337 |  *__result = __unary_op(*__first);
[build]       |              ~~~~~~~~~~^~~~~~~~~~
```

src/control/tools/EditSelectionContents.cpp
```diff
@@ -62,11 +62,25 @@ void EditSelectionContents::addElement(Element* e, Layer::ElementIndex order) {
     }
 }
 
+void EditSelectionContents::replaceInsertOrder(std::deque<std::pair<Element*, Layer::ElementIndex>> newInsertOrder) {
+    this->selected.clear();
+    this->selected.reserve(newInsertOrder.size());
+    std::transform(begin(newInsertOrder), end(newInsertOrder), std::back_inserter(this->selected), [](auto const& e,auto const&){return e;});
```
Febbe, line 68:
```suggestion
    std::transform(begin(newInsertOrder), end(newInsertOrder), std::back_inserter(this->selected), [](auto const& e){return e.first;});
```

rolandlo: Thanks

Technius:
Why does the insert order deque need to store the index of each element? The indices don't appear to be used anywhere.

rolandlo:
> Why does the insert order deque need to store the index of each element? The indices don't appear to be used anywhere.

The indices are used in `Layer::insertElement`to get the elements of the elements vector of a layer be added in the right order. The order in which the elements are added determines the order in which the elements are painted (which element covers which other element).

Technius:
> > Why does the insert order deque need to store the index of each element? The indices don't appear to be used anywhere.
> 
> The indices are used in `Layer::insertElement`to get the elements of the elements vector of a layer be added in the right order. The order in which the elements are added determines the order in which the elements are painted (which element covers which other element).

What I'm trying to say is that we can just have a `std::deque<Element*>` instead of `std::deque<std::pair<Element*, Layer::ElementIndex>>` because the index part of the pair is not being used anywhere.

rolandlo:
@Technius Have you taken into account `EditSelectionContents::finalizeSelection` as well? 

  ```      
if (index == Layer::InvalidElementIndex) {
    // if the element didn't have a source layer (e.g, clipboard)
    layer->addElement(e);
} else {
    layer->insertElement(e, index);
}
```

Febbe:
The index determines the insert order in each Layer. We should add a comment to clarify that. The deque itself determines the drawing order.

Technius:
Ah, there we go. Somehow I missed that `EditSelectionContents` already uses an `insertOrder` field internally.

Technius:
What's the status on this?

rolandlo:
> What's the status on this?

It's fully functional. But according to @Febbe's comment from Sep 25 some part in `EditSelection::copySelection` needs refactoring.

> In general this complete structure needs to be refactored. A selection should not take ownership of the elements, but that's another topic.

Technius:
I think we should merge this as-is (after resolving the merge conflicts, of course) and then do the refactoring in a follow up PR.

rolandlo:
I have resolved the merge conflicts and have checked if it still works fine (which it does).

Technius (approved):

PR #2286, opened 2020-09-27
fix resizing bug after pasting object

### LINKED ISSUE #2281: Resizing of 'the selection just after pasted' behaves oddly
This is a regression with the commit 6e5469874290871c501f14dbbca36099e96231d8 .

**Affects versions :**
 - OS: Arch Linux
 - Version of Xournal++: master (after 6e5469874290871c501f14dbbca36099e96231d8 )
 - Installation method: built from source

**Describe the bug**
When a copied object is pasted, it appears as a selected object.
But if you click the corner or the edge to resize it, the object
will jump to another place or disappear.

**To Reproduce**
1. Create a stroke. Call the object AAA.
2. Select AAA and copy it to the clipboard.
3. Paste the object in the clipboard. Call the appeared object BBB.
4. BBB appears as a selected object. Click one of the corner of it with a pen.

**Additional context**
1. If you click the corner with a mouse, this bug does't appears.

2. If you use LaTeX tool, the object generated by LaTeX  also
appears as a selected object. This object also gets this bug.

3. If you apply the commit 6f29f1cc85a576461d927f1ce347da4293179d36 ,
this bug seems to be avoided if the paste is the first time.
But if you try another paste succesively, it gets this bug again.

rolandlo:
Fixes #2281

### DIFF
--- a/src/control/tools/EditSelectionContents.cpp
+++ b/src/control/tools/EditSelectionContents.cpp
@@ -522,7 +522,11 @@ void EditSelectionContents::serialize(ObjectOutputStream& out) {
 void EditSelectionContents::readSerialized(ObjectInputStream& in) {
     in.readObject("EditSelectionContents");
 
-    this->originalBounds = Rectangle<double>{in.readDouble(), in.readDouble(), in.readDouble(), in.readDouble()};
+    double originalX = in.readDouble();
+    double originalY = in.readDouble();
+    double originalW = in.readDouble();
+    double originalH = in.readDouble();
+    this->originalBounds = Rectangle<double>{originalX, originalY, originalW, originalH};
 
     this->relativeX = in.readDouble();
     this->relativeY = in.readDouble();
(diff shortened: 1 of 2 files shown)


src/control/tools/EditSelectionContents.cpp
```diff
@@ -523,13 +522,7 @@ void EditSelectionContents::serialize(ObjectOutputStream& out) {
...
     in.readObject("EditSelectionContents");
 
-    double originalWidth = in.readDouble();
-    double originalHeight = in.readDouble();
-
-    double originalX = in.readDouble();
-    double originalY = in.readDouble();
-
-    this->originalBounds = Rectangle<double>{originalX, originalY, originalWidth, originalHeight};
+    this->originalBounds = Rectangle<double>{in.readDouble(), in.readDouble(), in.readDouble(), in.readDouble()};
```
Technius, line 525: This is less readible; side effect order should be explicit.

Technius (approved):

src/control/tools/EditSelectionContents.cpp
rolandlo: Ok, I will change it back then.

rolandlo:
Thanks for looking at it @Technius. I have made the effect order explicit again. Shall I squash and merge?

Technius:
I'll did it for you to save time ;)

PR #2304, opened 2020-10-08
Drop fs::exists check for recent files

### LINKED ISSUE #2225: Recently used file with "no such device" causes Xournal++ to crash at start.
**Affects versions :**
 - OS: Arch Linux
 - Desktop environment: Gnome-Wayland
 - Version of Xournal++: 45a619d83f97205c92ac146f13d5ae00af83af7e
 - Installation method: AUR ([xournalpp-git](https://aur.archlinux.org/packages/xournalpp-git/))

**Describe the bug**
When a file exists in the "recently used files" where the device is not available (f.e. a Samba mount) Xournal++ crashes at start.

**To Reproduce**
1. Mount a samba drive using a VPN interface
2. Open the file in Xournal++, close it
3. Make sure the VPN interface is not available without unmounting the Samba drive (`no such device`)
4. Re-open Xournal++

**Expected behavior**
Xournal++ ignores the unavailable file.

**Additional context**
Crash:
```
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: status: No such device [/mnt/samba/mount/that/is/not/available/file.pdf]

** (com.github.xournalpp.xournalpp:26191): WARNING **: 12:39:29.368: [Crash Handler] Crashed with signal 6

** (com.github.xournalpp.xournalpp:26191): WARNING **: 12:39:29.369: [Crash Handler] Wrote crash log to: $HOME/.cache/com.github.xournalpp.xournalpp/errorlogs/errorlog.20200908-123929.log
```

Error log: https://fb.hash.works/HMqP0W

File path is from `./.local/share/recently-used.xbel`:
```
<bookmark href="file:///mnt/samba/mount/that/is/not/available/file.pdf" added="2020-09-02T23:54:55Z" modified="2020-09-03T00:00:03Z" visited="1969-12-31T23:59:59Z">
```

This sounds like something that should've been fixed with #1730, but as I stated above I'm using the latest version in master.

hashworks:
Resolves #2225.

Please note that while this solves the "crash and burn" issue Xournal++ encounters when even a simple `fs::exists` failes, in #2225 I also had quite high timeouts – 10s or 30s _for every affected file in recents_.

10s alone is quite a bummer, but it adds up. My case is quite special, but let's assume that one just has a very slow network drive. 10 files that take 3s each for `fs::exists`, that's 30s just to start Xournal++. Maybe we just want to drop `fs::exists` here and only call that when we actually want to open a single file?

### DIFF
--- a/src/control/RecentManager.cpp
+++ b/src/control/RecentManager.cpp
@@ -130,7 +130,7 @@ auto RecentManager::filterRecent(GList* items, bool xoj) -> GList* {
         auto p = Util::fromUri(uri);
 
         // Skip remote files
-        if (!p || !fs::exists(*p)) {
+        if (!p) {
             continue;
         }
 


hashworks:
I've added a future timeout of 1s for every `fs::exists` call.

hashworks:
Any comments on this? :)

rolandlo:
@Febbe is our filesystem specialist, so I guess you will have to wait until he finds time to look at your PR

Technius:
Sorry, I would've looked at this earlier, but I've been crazy busy this month (and will be also crazy busy next month).

To improve the timeout logic, you'll want to wrap the check for each file in separate futures and then wait for them in parallel; otherwise the files will be checked sequentially, making the timeout not particularly useful.

I feel like the timeout is a sort of band-aid fix for the underlying problems you identified in https://github.com/xournalpp/xournalpp/issues/2225#issuecomment-705651474. In my opinion, we should properly fix the issue by changing the behavior of the recent files logic:
* The current behavior is to enumerate the saved recent files listing and then filter out all the "invalid" files. As a user, I would be surprised if a recent file entry suddnely disappears from the list. For example, if I rename a file, I would like an error to pop up when I try to access the old file, as a reminder that I renamed the file. What use is a recent files list if it doesn't remember my recent files?
* The current behavior is inconsistent with other applications, most of which simply present a dumb list and then indicate an error when a user chooses a nonexistent file.
* The new behavior of the recent files entry should be a dumb list that only remembers the chosen paths, doing no filtering or anything of the sort. Entries should only be deleted to make space for new ones.

Febbe:
Especially, when a file was located on an network device, the user might want to attach the device again, instead of loosing his recent files list.

hashworks:
> * The new behavior of the recent files entry should be a dumb list that only remembers the chosen paths, doing no filtering or anything of the sort. Entries should only be deleted to make space for new ones.

Sounds reasonable. I would just drop lines [141-153](https://github.com/xournalpp/xournalpp/blob/fe5b48dbf2064278bab5ec51c0ffb45fde45471d/src/control/RecentManager.cpp#L141), namely the `fs::exists` check. We would still filter by extension.

hashworks:
@Technius Force-pushed the removal and renamed the PR.

Technius:
Awesome, I'll take a look tonight and merge if there are no glaring issues.

Technius:
LGTM, thanks for the PR.

PR #2308, opened 2020-10-10
remove old input system and touch workaround

rolandlo:
This PR completely removes the old input system (for easier code maintainance), including the corresponding setting whether to enable the new input system or not. I hope I didn't miss anything.


Technius:
I think this should be merged after we release a beta/milestone version of current master so that the transition is not so sudden.

rolandlo:
The settings glade file will be fixed before it is merged. It takes too much work to resolve the merge conflicts every time.

rolandlo:
Merging this in 24h if no objections are made.

kubajaburek:
I personally don't think it's a good idea to remove the old input system when stylus buttons are not working correctly in the new one, as I already reported in https://github.com/xournalpp/xournalpp/issues/2628#issuecomment-761606275.

Unfortunately I don't really understand how the input system works and how to debug it, so I wasn't able to try to fix the issue myself.

rolandlo:
@kubajaburek It's better to fix the new input system than keeping using (and getting bug reports about) a much more buggy old input system.

PR #2309, opened 2020-10-11
fixup borders from empty toolbars

### LINKED ISSUE #2291: window borders and palm

**Affects versions :**
 - OS: Ubuntu
 - Desktop environment: X11 (Gnome)
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Version of Xournal++: 1.1.0+dev version of Sep 28 (commit d0f58e0), has been an issue for at least a month I think
 - Installation method: PPA and compiled from source

**Describe the bug**
The Xournal++ window has very thick borders on Gnome. These borders allow moving the window and double-clicking to (un)maximize. When writing on the right edge of the screen (as right-handed user) it sometimes happens that I accidentally (un)maximize the window with my palm.

**Expected behavior**
1.0.18 does not have these thicker windows

**Screenshots of Problem**
1.1.0-dev:
image

1.0.18:
image

rolandlo:
Fixes #2291 by only setting those toolbar widgets visible, which have at least one child.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -500,7 +500,10 @@ void MainWindow::setToolbarVisible(bool visible) {
 
     settings->setToolbarVisible(visible);
     for (int i = 0; i < TOOLBAR_DEFINITIONS_LEN; i++) {
-        gtk_widget_set_visible(this->toolbarWidgets[i], visible);
+        auto widget = this->toolbarWidgets[i];
+        if (!visible || GTK_IS_CONTAINER(widget) && gtk_container_get_children(GTK_CONTAINER(widget))) {
+            gtk_widget_set_visible(widget, visible);
+        }
     }
 
     GtkWidget* w = get("menuViewToolbarsVisible");


rolandlo:
@duncanawoods If you have time please check this PR. It should fix something we both missed in https://github.com/xournalpp/xournalpp/commit/1ba572969e27948264f5ada821f558531c330fd9.

duncanawoods:
Oh wow, good find. Sorry for not noticing this at the time. It took me a few minutes to even see the difference between side-by-side screenshots!

It fixes the problem on my testing. Minor questions:

1. Are children ever added to empty toolbars dynamically? Could we get in trouble not restoring a necessary toolbar because it only happens to be empty at the point it is tested?

2. It is a toggle method so the change will mean it also won't hide empty toolbars. If guess it is a non-issue assuming that empty toolbars are already hidden, but could add a `&& visible` if only wanting to make sure we don't show them.

rolandlo:
> It fixes the problem on my testing. Minor questions:
> 
>     1. Are children ever added to empty toolbars dynamically? Could we get in trouble not restoring a necessary toolbar because it only happens to be empty at the point it is tested?

They are (in `ToolMenuHandler::load`), but the toolbar is made visible at the end if the element count is greater than 0. I think this should not be a real concern, since after adding to (empty) toolbars dynamically one can set the toolbar visible (again).
> 
>     2. It is a toggle method so the change will mean it also won't hide empty toolbars. If guess it is a non-issue assuming that empty toolbars are already hidden, but could add a `&& visible` if only wanting to make sure we don't show them.

This means `!visible ||` for the `gtk_widget_set_visible` call. Here I agree with you. I have changed that.

duncanawoods:
Yep looks good.

rolandlo:
Thanks for looking at it once more. I have merged it now.

Issue #2211, opened 2020-09-02
Xournal++ crashes when a blank page is added in "Annotate PDF" mode


jgrnst:
**Affects versions :**
 - OS: Linux Kubuntu 20.04
 - (Linux only) KDE
 - Which version of libgtk do you use
     - 3.24.20
 - Version of Xournal++: 1.1.0+dev
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Installation method:
       Package built from source

**Describe the bug**
When a blank page is added in "Annotate PDF" mode xournal++ freezes and eventually crashes.

**To Reproduce**
1. Launch xournal++
2. Open a PDF file in "Annotate PDF" mode
3. Open the pull-down menu next to "Insert page" button at the bottom of the screen
4. Select any blank page format (tried with plain, ruled, ruled with a vertical line)'
5. Click "Insert page" button 

**Expected behavior**
A blank page should be added to the document

**Screenshots of Problem**

** Message: 15:39:17.874: lastOpenPath set
[1]+  Segmentation fault      (core dumped) xournalpp

Error log is not created

rolandlo:
Most likely this is the same bug as reported in #2144. The problem is that Xournal++ enters an infinite loop of zooming/scrolling methods calling each other. The freeze is fixed in pull request #2146 but at the cost of not getting page number update if you move to page 1 of the document.

rolandlo:
In fact, as I checked now, with PR #2146 there is no freeze when adding a blank page in "Annotate PDF" mode. It is interesting, that this happens only when the background is a pdf in the unfixed version.

nikkopomata:
I've been experiencing essentially the same bug (version 1.1.0+dev, libgtk 3.24.20, Mint 20, XFCE 4.14)

The following DOES NOT work, i.e. xournal++ crashes:
- Adding non-pdf-background pages to a pdf-background document (existing .xopp, new Annotate Document, etc) when xournal++ was opened on a blank-background document (e.g. opened without an argument, or opened with a reference to an existing blank-background document) (exception below)
- Adding image-background pages to a blank-background document when xournal++ was opened on a blank-background document

The following DOES work fine:
- Adding blank-style pages to a blank-background document, apparently any context
- Adding pdf-background pages to a pdf-background document, apparently any context
- Adding blank-style pages to a pdf-background document *when the focus is on a blank-background page* (which has been added in a previous session)
[...]

Issue #1926, opened 2020-05-02
adding a new page sometimes shifts focus to beginning of the document


petergthatsme:
**Affects versions :**
I'm using latest version (master) on ubuntu 18.04. 
My documents are usually in landscape format; but this should not be a requirement to see this bug. 

**Describe the bug**
Adding a new page (very often, but not always) via the +page button both adds a new page (this is a good thing), but also moves the user/interface/focus to the first page of the document. 

**To Reproduce**
1. Step 1
Create a new document, and scribble something.
2. Step 2
add a new page; it will be added, but sometimes the interface stays focused on the first page. 
3. Step 3
sometimes a few pages have to be added for this to be observed; also I notice that this happens more often (I think) if one scrolls to the bottom of the "current" page before adding a new one. 

**Expected behavior**
After adding a new page, the page should be added, and now in focus. 

thanks.

michaelcadilhac:
I can confirm this, but same as @petergthatsme , not sure what's the correct recipe.

Rmano:
Yes, it happens to me too. I think that it is related to copy and paste. I will attach a blank file here in a zip).

[Pizarra-Xournal++.zip](https://github.com/xournalpp/xournalpp/files/4581067/Pizarra-Xournal%2B%2B.zip)

Steps:
1. open the file
2. draw something in the first page
3. select it and do ctrl-c
4. go to page 2, ctrl-v
5. go to page 3, ctrl-v
6. optionally click to exit the selection mode (does not change things)
7. add a page, you'll jump to the first one.

![test](https://user-images.githubusercontent.com/6414907/81073479-fd828380-8ee7-11ea-8a12-3ea755008d38.gif)

michaelcadilhac:
I wasn't using copy/paste, but it's good that we have a recipe!  Following exactly @Rmano 's instructions, I do jump back, although not to the top of the first page, but somewhere on the first page.

An easier recipe, for me, seems to open @Rmano 's document, scrolling and selecting the last page, and adding pages repetitively.   This is very erratic; first page is inserted while not moving, second jumps at the beginning, then it alternates between new page and first page, then I can't even follow the pattern :-D

Technius:
Looks like another case of #1658.

neldredge:
[...]

PR #2312, opened 2020-10-11
Force page layout update after inserting page

### LINKED ISSUE #2289: I come back to the top of the document when I add a new page at the end of a document
I work under windows 10 with the nightly-automated release of the 28 september.
Sometimes, when I add a new page at the end of a document, I come back to the top of the document.

rolandlo:
This PR should fix  #2289 and #1926 about focus shift to the beginning of the document after inserting a page. It may also fix the freeze from #2211 and  #2144 when inserting a blank page in a pdf document.

The first commit updates the page layout immediately after inserting the page. This only happens after some time (usually too late) without this fix. It also removes unnecessary calls to `firePageSelected` (which is called from `scrollToPage` anyway) and `layoutPages` which gets triggered after the `gtk_window_move_resize` anyway.
It does not quite solve the problem of scrolling to the new page though, since sometimes scrolling goes to the end of the current page. More work is required.


Rmano:
Testing. 
BTW, to test PR I do the following: 
```bash 
git fetch origin pull/2312/head:pr-2312
git checkout pr-2312
cd build
make
```
Is this correct? Should I also remove the `build` direectory?

Rmano:
MUCH better now. I didn't manage to make it jump back (congrats!).
It's not perfect though.  Notice in the animation below: 
1 - adding the second page should (in my opinion) scroll to show the new page (probably aligned with the bottom)
2 -adding the third page does not scroll at all, making the new page invisible unless you manually scroll
3 - adding the 4th page show what is in my opinion the correct behavior --- scrolling so that the last page is full on screen.

Peek 2020-10-12 12-30


But nevertheless, this is a nice step forward, thanks to all the contribs that helped find this.

rolandlo:
@Rmano 
Thanks for testing. You observe the same behavior as I do. Let's see if we can make it perfect then.

rolandlo:
Messed it up too much. I will try it with another PR.

rolandlo:
Closed in favor of PR #2346.

Issue #1886, opened 2020-04-21
Edit layer names


atticus-sullivan:
Hi,

I'm not sure if this would be difficult to implement (regarding the file structure), but it would be really nice, if yo were able to edit the layer names to something which describes the content better than layer 1, layer 2,...

Technius:
This was implemented in #2321 but there is still room for improvement.

atticus-sullivan:
Quite late (just cleaning up my issues), but for me with this renaming functionality this is solved

PR #2321, opened 2020-10-16
[WIP]: Editable layer name

Mazurel:
This pr refers to issue #1886.
I have simply added an option to rename layers, the option is available in Journal menu and can be used with Ctrl+R shortcut.
When clicked or run via shortcut, it runs prompt for new layer name and when accepted changes currently selected layer's name to new one.
New names are saved into xopp file but are not necessary, so backward compatibility is not a problem.

Sadly I have an issue with the undo option. 
The names are changed when undo-ed but when I try to refresh currently displayed layer names xournalpp freezes on finishing job mutex on sidebarentry destructor. I think that it may be due to document lock on the other thread, but I am not sure how it works (screenshot of gdb are below). 
Right now I can't figure it out on my own. Currently the refresh just doesn't work even though the name changes.

Menu entry and example names
image

Dialog
image

Gdb Thread 1
image

Gdb Thread 15
image


rolandlo:
Thanks for your PR. I will take a look at it, when I find time (maybe this weekend).

rolandlo (commented):
After some tests I can provide some comments. I haven't studied the code in detail yet. It should be enough to give you some idea of what needs to be improved.

- Generally speaking I like the feature and the Layer rename dialog. Thanks for your work to implementing it. I found some wrong behavior in addition to the problem you pointed out with undoing layer renaming.
- It is not only that the undo operation for renaming layer results in a freeze/crash, but also the undo operations for inserting a new layer (or copying a layer) and for moving layers. Those undo operations worked fine previously. So you must have introduced some regressions with respect to that.
- After creating a new document or adding a new page, there should be only the background layer. A first layer should only be created when starting to write/draw. In the dropdown menu from the toolbar you should only find the background layer. In your PR I see layer 1 listed instead, no background layer is listed. This is a mistake. Moreover in the sidebar layer preview I see both the background layer and layer 1 in your PR, when there should only be the background layer.

- In the rename layer dialog, I suggest to make the Enter key confirm the new name.
- The layer name attribute is also exported in the .xoj compatibilty export mode. It should be checked carefully if that leads to problems loading those files in Xournal. After a first test it seems to work fine.
- I am wondering if the rename layer operation (and the delete layer operation) should also be listed in the dropdown menu from the toolbar.

Mazurel:
Thanks for your response, especially for pointing the fact that the issue also affects other layer actions. I will try to look into this as fast as I will be able to.

Mazurel:
Currently there is still an issue with the document locking when undo-ing any layer related action when sidebar with layer preview is opened. The issue is created due to destruction of `SidebarPreviewLayerEntry` object which locks document and waits for some job to finish (although there are no other jobs active).  The problem is I don't really know how async jobs work in xournalpp, so when I will figure it out I will be able to fix it. 
Sadly I have had a lot of work to do lately so I couldn't give this problem enough time, but I will try to fix it as soon as possible.
Other than that everything seems to be working OK.
Also I will apply `clang-format` when the bug will be fixed.

Mazurel:
So I have fixed the bug, but I am not sure at what cost. The cause of the bug was as I mentioned before some mutex problem. I have fixed it by commenting out the code in `src/control/jobs/XournalScheduler.cpp:64`.
It doesn't seem to have broken anything and have fixed all the layer issues.
I will look more into it, but I would appreciate if someone with more knowledge about this subject would help me with this `finishTask` function/method.

rolandlo:
@Mazurel Can you please rebase on current master and fix the conflicts?

rolandlo (changes requested):
Some more comments:
- Great that you found a way to avoid the freeze
- Some things don't quite work as expected yet:
  - If I start with a new document (or a new page) and open the layer preview (in the sidebar), a new layer is generated. This should not happen. Moreover the background layer is called "Layer 1" in the Xournal++ toolbar. This is wrong.
  - In the toolbar dropdown and in the sidebar layer preview, layers without names remain unlabelled, in the Xournal++ toolbar, they are labeled. This is at least inconsistent. Maybe it would be better to have a provisional labeling for unnamed layers.
- The layer names should also be made available to the plugin system. In `luapi_applicatoin.h` there are already some lines in `applib_getDocumentStructure` that only need to be uncommented. Background layer names should be added, though. Moreover it would be nice to have an `applib_getLayerName` function in the API, that could be called like `app.getLayerName("Corrections")` in a Lua plugin. If you need help with that, let me know.

Mazurel:
>If I start with a new document (or a new page) and open the layer preview (in the sidebar), a new layer is generated. This should not happen. Moreover the background layer is called "Layer 1" in the Xournal++ toolbar. This is wrong.

Has been fixed.

>In the toolbar dropdown and in the sidebar layer preview, layers without names remain unlabelled, in the Xournal++ toolbar, they are labeled. This is at least inconsistent. Maybe it would be better to have a provisional labeling for unnamed layers.

Was related to the previous one, it has also been fixed. All layers have default name Layer {n}.

> The layer names should also be made available to the plugin system. In luapi_applicatoin.h there are already some lines in applib_getDocumentStructure that only need to be uncommented. Background layer names should be added, though. Moreover it would be nice to have an applib_getLayerName function in the API, that could be called like app.getLayerName("Corrections") in a Lua plugin. If you need help with that, let me know.

I have implemented basic API for writing new layer/background name, although I am not sure how to make function return value in LUA. Would love if you would help me with that.

Right now there is also apparently build fail, I will try to resolve it.

rolandlo:
Xoj-compatibility exports of documents with named background layers can't be opened in Xournal. The other layer names appear to be no problem, but the background layer names are.

rolandlo:
The Lua plugin stuff works well according to my tests. Are you asking about how to make a return value in a function of the `.lua`-file? Or did you ask about the API?

Mazurel:
I will try to fix the compatibility with Xournal (maybe today).

About the lua, what I was wondering is if it makes sense to add `getCurrentLayerName` function. I don't have any experience with lua and I wonder if `getDocumentStructure` is not enough (I don't want to over-complicate the API).

rolandlo:
> I will try to fix the compatibility with Xournal (maybe today).
> 
> About the lua, what I was wondering is if it makes sense to add `getCurrentLayerName` function. I don't have any experience with lua and I wonder if `getDocumentStructure` is not enough (I don't want to over-complicate the API).

The way that return values work is that you put everything into a Lua table which is put onto the stack and the plugin reads it from there. 
In my opinion we should keep the API simple and things that can be retrieved from `getDocumentStructure` don't need a separate getter. Otherwise we would have to create lots of them. So I strongly advise against creating `applib_getCurrentLayerName`

Mazurel:
Okay. In this case will focus on the Xournal compatibility.

src/model/XojPage.cpp
```diff
@@ -55,7 +56,6 @@ void XojPage::removeLayer(Layer* layer) {
     }
     this->currentLayer = npos;
 }
-
 void XojPage::setSelectedLayerId(int id) { this->currentLayer = id; }
```
rolandlo, line 59: Please add back the empty line above this line.

src/undo/LayerRenameUndoAction.h
```diff
@@ -0,0 +1,37 @@
...
+
+public:
+    virtual string getText();
+    bool undo(Control* control) override;
+    bool redo(Control* control) override;
+
+private:
+    Layer* layer;
+    string newName;
+    string oldName;
```
rolandlo, line 36:
```suggestion
    LayerRenameUndoAction(Layer* layer, const std::string& newName, const std::string& oldName);
    virtual ~LayerRenameUndoAction();

public:
    std::string getText() override;
    bool undo(Control* control) override;
    bool redo(Control* control) override;

private:
    Layer* layer;
    std::string newName;
    std::string oldName;
```

src/control/xojfile/XojExportHandler.h
```diff
@@ -32,6 +32,6 @@ class XojExportHandler: public SaveHandler {
     virtual void writeHeader();
     virtual void writeSolidBackground(XmlNode* background, PageRef p);
     virtual void writeTimestamp(AudioElement* audioElement, XmlAudioNode* xmlAudioNode);
-
+    virtual void writeBackgroundName(XmlNode* background, PageRef p);
```
rolandlo, line 35: Please use the `override` keyword (at the end) and don't use the `virtual` keyword for these 5 functions.

rolandlo (changes requested):
Now mostly everything looks fine. I have just a few minor nitpicks. Also please apply clang format. After that you can squash your commits, create a good commit description and rebase on current master.

rolandlo (approved):
Thanks! Everything looks fine now. I will merge.

Issue #2082, opened 2020-06-21
^Z/^Y-undo/re-do shortcuts stop working when Menu is hidden (F10)


tjavdar:
**Affects versions :**
 - OS: Archlinux
 - uname -a: Linux xe 5.7.4-arch1-1, x86_64 GNU/Linux
 - Desktop environment: i3
 - Which version of libgtk:  1:3.24.20-1
 - Version of Xournal++: 1.0.18, Git commit: 535d261d
 - Installation method:  Arch repo

**Describe the bug**
The shortcuts
       Ctrl-Z (undo last stroke) and Ctrl-Y (redo undone stroke)
stop working after the menu is hidden via F10.

**To Reproduce**
1. Start xournal and make a stroke. Press ^Z to undo it and ^Y to redo it.
2. Press F10 to hide the "File  Edit View ..." top menu.
3. Pressing ^Z and ^Y stop working

**Expected behavior**
 Ctrl-Z (undo) and Ctrl-Y (redo) shortcuts should work no matter if the Menu is on-screen or not

s1-ranjan:
I would like to add one more important observation in this most of the shortcut keys are non functional when menu is hidden.

duncanawoods:
I'll take a look at this. I believe it is a default GTK behaviour which can be fixed by binding the accelerators to the main window, not the menu items:

https://stackoverflow.com/questions/19707775/accelerators-stop-responding-when-menubar-gets-hidden

edo0:
Duplicate of  #1002

duncanawoods:
I've taken a look at this. 

I've tried to iterate through the menu items use them to generate global shortcuts but my attempts to invoke the bound GClosure haven't worked and the menu item widget itself isn't valid when the menu is hidden. I don't think there is an alternative to manually creating the bindings duplicating/replacing the menu accelerators. 

We can't have both the global shortcuts and the menu accelerators bound to the same keys at the same time.

There is an argument for removing the menu accelerators. To match the current UI we would then have to manually add labels to indicate the global keys.

For now, I have implemented a way to enable/disable global accelerators when the menubar is shown/hidden and it works but I would like some confirmation to proceed before I implement the rest of the accelerators. It uses an intermediary representation of the accelerators which will be a good beginning for key maps #919 where most of the work will be the settings pages to configure the shortcuts.

```cpp
typedef void(*AccelCallback)(MainWindow* mainWindow);

struct GlobalAccelerator {
    AccelCallback callback; 
[...]

Issue #1002, opened 2019-03-21
Hiding menu bar with F10 disables "Ctrl + " shortcuts


edo0:
When I hide the tools Menu (File, Edit, View etc.) by pressing F10 key (images attached depict transition), I lose the ability of making in-program commands involving keyboard shortcuts.
E.g.: Ctrl+S, Ctrl+Z, Ctrl+E, Ctrl+C will not produce anything
Whereas other system-wide shortcuts (E.g. Alt-tab) still work fine.
App is Xournal 1.0.9 built 20/03, on Elementary 5.0 (Ubuntu 18.04 based)

![Screenshot from 2019-03-21 17 17 04](https://user-images.githubusercontent.com/16632292/54767662-eb6b5600-4bfd-11e9-9fe7-1d92e4b7f4bd.png)
![Screenshot from 2019-03-21 17 22 01](https://user-images.githubusercontent.com/16632292/54767670-f0c8a080-4bfd-11e9-9c4c-edf95f85ea68.png)

peetCreative:
This is probably do to that the shortcuts are bound to the elements in this toolbar.
This issue should be resolved #803

LittleHuba:
This will be solved by #919

bhennion:
Closing as I think it's been resolved.

PR #2324, opened 2020-10-19
Rebind accelerators when the menubar is hidden

duncanawoods:
## Problem

Fix for #2082 and #1002 and progress towards user defined keymaps #919 for @rolandlo to test.

Keyboard accelerators for things like copy/paste/fullscreen/tools are bound by the menubar definition in glade. When the menubar is hidden, they no longer work.

## Implementation

There is no way to use the menubar programmatically when hidden and no other way I could find to hide it that preserved the accelerators so the only path I could see was to manually rebind them when the menubar is hidden. We can't have menubar and manual accelerators bound at the same time so they have to be toggled on/off when the menubar is reshown.

This means we have duplication of the accelerator definitions in glade and `initGlobalAccelerators` that must be kept in sync. If we removed the accelerators from glade then we would lose the UI integration that displays the short-cut key. If implementing user-defined key bindings then I think we would be forced to remove the glade accelerators and manually create UI labels to show the user's chosen short-cut.

I have reused the `actionPerformed` infrastructure where possible. This does mean providing dummy information for the parameters we don't have like the gtk event and the button being pressed which is currently fine but a future developer might change the action implementations to use these parameters causing a problem.

It looks like we could almost auto generate these bindings by parsing the glade file but it would make assumptions about how the accelerators are implemented that could easily be broken so lambdas makes it flexible to provide any necessary behaviour.

References:

https://stackoverflow.com/questions/19707775/accelerators-stop-responding-when-menubar-gets-hidden

## Testing

I have tested the accelerators with/without the menubar and it looks ok. 

I found one conflict with `f11` and `XournalView::onKeyPressEvent` which caused double toggling. I have removed it and it looks ok. Areas to examine would be any other view states that might manually handle key events.

## Remaining

I haven't changed the current `MainWindow::initHideMenu` which manually rebinds f10. I think I can remove it and make everything consistent.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -313,6 +313,8 @@ void Control::initWindow(MainWindow* win) {
 
     this->pluginController->registerMenu();
 
+    win->rebindMenubarAccelerators();
+
     fireActionSelected(GROUP_SNAPPING, settings->isSnapRotation() ? ACTION_ROTATION_SNAPPING : ACTION_NONE);
     fireActionSelected(GROUP_GRID_SNAPPING, settings->isSnapGrid() ? ACTION_GRID_SNAPPING : ACTION_NONE);
 }
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -35,7 +35,7 @@
 #include "util/DeviceListHelper.h"
 
 MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control):
-        GladeGui(gladeSearchPath, "main.glade", "mainWindow"), ignoreNextHideEvent(false) {
+        GladeGui(gladeSearchPath, "main.glade", "mainWindow") {
     this->control = control;
     this->toolbarWidgets = new GtkWidget*[TOOLBAR_DEFINITIONS_LEN];
     this->toolbarSelectMenu = new MainWindowToolbarMenu(this);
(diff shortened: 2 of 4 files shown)


rolandlo:
Thanks for this very welcome PR. It works fine so far. I will take more time for testing and checking the code, though.

rolandlo (commented):
Testing your PR I found one major issue: Accelerators defined in plugins still don't work when the menu is hidden.

Have you considered using [gtk-accel-groups-from-object](https://developer.gnome.org/gtk3/stable/gtk3-Keyboard-Accelerators.html#gtk-accel-groups-from-object) to retrieve all accel groups which are attached to the main window (instead of redefining them manually)? If that works, this would solve the problem with plugin accelerators and make maintainability easier.

duncanawoods:
I did write something that walked the menu and extracted all the `GtkAccelGroups` but I believe that invoking the bound `GClosure` when it was hidden didn't work. I suspect it's dependent on the parent control being enabled. I have since deleted that version but I have some scratch notes around that I will resurrect and check again.

Can you see a way to extract the `ActionType` from a `GtkAccelGroup`? I can't but that would be an alternative to reusing the `GClosure`.

> Testing your PR I found one major issue: Accelerators defined in plugins still don't work when the menu is hidden.

Ah. I don't know anything about plugins.

duncanawoods:
Ok good news. I have rewritten it to walk the menubar and rebind each accelerator to a new closure that invokes the menuitem widget itself. This has allowed removing the special handling of the menubar toggle and it's double event workaround.

I'm not sure how to test plugins. I believe I have ColorCycle plugin enabled and I have looked at the code for the shortcut but `alt-c` doesn't seem to do anything on the main-branch. Is there some more setup I need to do test them from source?

rolandlo:
Great! Plugin accelerators don't work yet, but I found a fix for that: simply wait for the plugin controller to register menus
    `this->pluginController->registerMenu();`
(line 314 in Control.cpp)
until you invoke `win->rebindMenubarAccelerators();`
(on the next line). Don't forget to provide access to the `rebindMenubarAccelerators()` to the Control class, since it is declared as a private method of the `MainWindow` class.

If you have the `ColorCycle` plugin enabled you should see it in the menubar (menu plugin, below `Plugin Manager`). Note that after enabling the plugin you must restart Xournal++ once before it is registered. If you don't see the plugin after enabling it and restarting Xournal++ you should see an error message in the console when starting Xournal++. 

Once the `ColorCycle` plugin is enabled you can press `Alt+c` to change color. You will see the change of color in the toolbar. You can also simply draw something with the pen, press `Alt+c`, draw something more, press `Alt+c`, draw more ... and observe that the color changes every time.

duncanawoods:
Thanks! I have force pushed the plugin fix.

I have also extended the menubar walking to recur on submenus. At the moment there are no accelerators in cascading menus so it wasn't strictly necessary but it felt like a bug waiting to happen in future.

rolandlo:
It works fine with the plugin accelerators as well now. Great! I will take another look in the evening, but if I don't find anything concerning in the code, this looks mergeable to me. Or do you have still plans to change something?

duncanawoods:
it's ready for merging from my POV and of course I'm happy to address any style or other changes you would like.

duncanawoods:
Thank you for your support @rolandlo!

rolandlo:
You're welcome. Thanks for your great PR. Always a pleasure of working with you.

Issue #1797, opened 2020-03-24
Strokes are drawn when exiting a selection


coldspark29:
**Affects versions :**
OS: Manjaro
(Linux only) Desktop environment: X11
Which version of libgtk do you use
gtk3 1:3.24.14+9+g429a6287e0-1
gtk2 2.24.32-2
Version of Xournal++: 1.1.0+dev 86d5178b
Installation method: AUR

**Describe the bug**
To exit a selection you have to tap somewhere with the pen. No stroke should be drawn then. @LittleHuba already fixed that once and someone broke it again.

Febbe:
I can't reproduce this with the newest master (drew something -> selected it -> tapped outside of the selection).
This seems to be an issue with specific configurations.
Do you have special options activated like the old Input System or the touch and scroll workaround?
Or even better can you provide your configuration files?

coldspark29:
No I am on the new input system and have none of the mentioned configurations activated. 


[settings.zip](https://github.com/xournalpp/xournalpp/files/4401470/settings.zip)

Febbe:
Hello Jan, 
thanks for sharing your configuration. I can reproduce it now. It is not a problem with the selection itself, rather with the fact, that you configured your first pen button to the selection tool and the tool is changed back to pen as soon you release the first pen button.
Holding the the button while clicking should be a first workaround.

coldspark29:
Okay, that makes sense. I have configured it like this, because I have the Bamboo Ink (in my opinion the best pen out there) and the buttons on it are inverted. Thank you for the advice!

Technius:
This has been addressed by the recently added setting `Stylus > Ignore the first X events`.

coldspark29:
No, that hasn’t solved it for me.

Gruß,
Jan Claußen

LittleHuba:
The setting would not fix it if set correctly. It discarded the input for everything even canceling a selection. 
We should fix this in the input system. It has to check if there is a selection that is being canceled and then discard the current input.

coldspark29:
Yes, I also think that it introduced a bug. I activated the setting at once when I read about the merge. Now when I make a selection, resize it and then duplicate it, it breaks the resizing operation previously made. I can't confirm right now if it happened due to the setting, but it fits Ulrich's description and he knows his input system best!

coldspark29:
I just checked it. The bug I described is not related to the setting.

PR #2339, opened 2020-10-24
Fix regression default button

### LINKED ISSUE #2303: Clicking on "default tool" (from the menu or via toolbar) does not change the tool at all
Under windows 10 

I created a shortcut with the "Shift +Ctrl+ D" keys for the "Default Tool" menu command. When I use it, I see that the "Default Tool" command is correctly selected, but as soon as I move the stylus or the mouse, it is the eraser that appears instead of the stylus (which I programmed in "Default Tool"). 
When I use the shortcut and move the stylus, it's switch automatically from pen to eraser.

 Clicking on "default tool" (from the menu or via toolbar) does not change the tool at all.
This appareas since the last commit.

Maybe @idotobi will find a quick fix for it

idotobi:
This PR fixes #2303 by giving every toolbutton an independent Tool instance to operate on (see commit message 2a2a3e5b1ce608bd9e23de5d7bc977ba82968749).

### 1. Applying changes in the settings of the Buttons requires Restart

@Febbe , @rolandlo : Do you know a way to call a function from the Controler or ToolHandler upon closing the preferences menu?

### 2. Different Eraser Types for different Buttons are not yet supported

This would require an additional refactor and I didn't have the time for this yet.


idotobi:
The second limitation mentioned above is fixed by 063b80ef069b4a14b3ee15c2b06b75e6bd753cc5.

Now the only limitation left is:

1. Applying changes in the settings of the Buttons requires Restart

src/control/ToolHandler.cpp
```diff
@@ -105,8 +108,13 @@ ToolHandler::~ToolHandler() {
     this->settings = nullptr;
 }
 
-void ToolHandler::setEraserType(EraserType eraserType) {
-    this->eraserType = eraserType;
+void ToolHandler::initButtonTool(ToolPointer tp, ToolType type) {
+    setToolPointer(new Tool(tools[type - TOOL_PEN].get()), tp);
+}
```
idotobi, line 113: I think this should rather be something like:

```cpp
Tool* tool = getToolPointer(tp);
delete tool;
setToolPointer(new Tool(tools[type - TOOL_PEN].get()), tp);
```

Otherwise, this would be a memory leak. However, right now this doesn't work because the "copy constructor" `Tool(Tool* t)` doesn't do a deepcopy of the `thickness` array (which is a member of `Tool`). 

I would see two solutions:
1. Include deepcopy of `thickness` array in copy constructor: I tried this but for whatever reason it did crash the program in other parts (probably a programming mistake but somehow this solution also feels ugly).
2. Use `unique_ptr`'s for the the button pointers rather than normal pointers: This sounds like the right solution to me but I didn't use smart pointers before, so I'm a little bit unsure how to approach this.

@Febbe , @LittleHuba : What would be the correct way to approach this? If you could give me pointer on how to do it either way that would help me a lot.

idotobi: Little update:

I managed to implement the `unique_ptr` and the `deepcopy` :tada:

idotobi:
:tada: I think this PR is ready for Review now. I just resolved the last issue (now buttontools get updated when closing the Settings window).

idotobi:
@Febbe , @LittleHuba : Would one of you have time to review the PR?

LittleHuba:
Can do that earliest coming weekend. If @Febbe hasn't reviewed then already I'll take a look.

src/control/Control.cpp
```diff
@@ -2468,6 +2470,26 @@ void Control::applyPreferredLanguage() {
 #endif
 }
 
+void Control::initButtonTool() {
+    ButtonConfig* cfg = settings->getEraserButtonConfig();
+    cfg->initButton(this->toolHandler, Button::eraser);
+
+    cfg = settings->getStylusButton1Config();
+    cfg->initButton(this->toolHandler, Button::one);
```
LittleHuba, line 2478: Please make this `Button::stylusOne` and so on.

src/control/ToolHandler.cpp
```diff
@@ -95,18 +95,34 @@ void ToolHandler::initTools() {
     tools[TOOL_FLOATING_TOOLBOX - TOOL_PEN] = std::make_unique<Tool>("showFloatingToolbox", TOOL_FLOATING_TOOLBOX,
                                                                      Color{0x000000U}, TOOL_CAP_NONE, nullptr);
 
-    this->buttonSelectedTool = &getTool(TOOL_PEN);
+    this->eraserButtonTool = std::make_unique<Tool>(tools[TOOL_HILIGHTER - TOOL_PEN].get());
+    this->button1Tool = std::make_unique<Tool>(tools[TOOL_HILIGHTER - TOOL_PEN].get());
```
LittleHuba, line 99: Same here: `stylusButton1Tool`

src/control/ToolHandler.cpp
```diff
@@ -446,10 +481,17 @@ void ToolHandler::loadSettings() {
     }
 }
 
-void ToolHandler::pointCurrentToolToButtonTool() { this->currentTool = this->buttonSelectedTool; }
+void ToolHandler::pointCurrentToolToButtonTool(Button button) {
+    Tool* tool = getButtonTool(button);
+    this->activeTool = tool;
+
+    this->listener->toolColorChanged(false);
```
LittleHuba, line 488: Sure the color does never change?

src/control/ToolHandler.h
```diff
@@ -23,7 +23,8 @@
 #include "Tool.h"
 #include "XournalType.h"
 
-enum ToolPointer { current, toolbar, button };
+enum SelectedTool { active, toolbar };
```
LittleHuba, line 26: I'm not really happy with "active". Am I right with assuming this is a tool that was selected from the toolbar but then modified?

LittleHuba (commented):
So I couldn't do any actual test but I trust in you having tried out what you did. Just a few minor nitpicks and potential issues I spotted.

I'm not that deep in the tool system so I just asked whatever question did not answer itself through your changes.

src/control/Control.cpp
idotobi: Agreed. That's the better wording :+1:

src/control/ToolHandler.cpp
idotobi: :+1:

src/control/ToolHandler.cpp
idotobi: I had to look this up.

This is actually slightly misleading when one reads it the actual signature of that function is:

```cpp
virtual void toolColorChanged(bool userSelection);
```

And `userSelection` basically denotes whether the there is currently something selected for which this change in color should be applied.

idotobi: If you have a good idea on how this part can be refactored to be more readable I'd be happy to refactor it :smile: .

src/control/ToolHandler.h
idotobi: The logic here is the following

There are two kinds of pointers
* `Button` Pointers: will not change the tool they are pointing to over their lifetime unless the settings are changed
* `SelectedTool` Pointers: will mostly change the tool they are pointing to over their lifetime

The meaning of the button pointers is self explanatory (I hope).

For the Toolpointer it goes as follows:
* `toolbar`: Points to the tool currently selected in the toolbar
* `active`: Points to the tool currently used when drawing or interacting with the application. Hence `active` points either to `toolbar` selected tool or two one of the `Button`s

idotobi: Is there a nice `cpp` conform way to document this in the code? I guess in this case it would be very hard to make the variable name itself self descriptive enough.

idotobi:
@LittleHuba : Thanks for your comments (and trust). I'll refactor the code according to your comments. To make sure I didn't oversee anything while testing I'm gonna use this branch for a book summary I'll have to do this week. If I don't spot anything out of the ordinary I would call it fine from my side then.

src/control/ToolHandler.cpp
LittleHuba: Ahh okay. That is a really bad way of mixing those two things together. Better would be to refactor both functionalities into their own functions and call those which are required. Mixing multiple actions within one function leads to eaxactly this issue of not really knowing what a function does.

src/control/ToolHandler.h
LittleHuba: Okay that is a really complex way of doing stuff. Looks like this system grew some features and now ended up with three arms and one and a half leg instead of two of each.
It would be great if we could simplify this. So it's up to you if you want to do this in this PR or a follow-up or leave it for somebody else but I think this desperately needs some simplification. How exactly though would require some looking into the code and its requirements.

src/control/ToolHandler.cpp
idotobi: True. That is a good approach. I'll refactor that part.

src/control/ToolHandler.h
idotobi: You're right some simplification would be good for this part of the code. It was quite complex when I started and I might have even added to the complexity with my recent changes.
I would probably still do it outside of this PR to fix the regression first. Even though the Code is complex at least it works right now. While I'll do the other changes I'll already think about how this could be simplified.

src/control/Control.cpp
idotobi: Done :heavy_check_mark:

src/control/ToolHandler.cpp
idotobi: Done :heavy_check_mark:

src/control/ToolHandler.cpp
idotobi: Done :heavy_check_mark:

src/control/ToolHandler.cpp
```diff
@@ -446,13 +483,20 @@ void ToolHandler::loadSettings() {
...
 
-void ToolHandler::pointCurrentToolToButtonTool() { this->currentTool = this->buttonSelectedTool; }
+void ToolHandler::pointCurrentToolToButtonTool(Button button) {
+    Tool* tool = getButtonTool(button);
+    if (this->activeTool == tool)
+        return;
+    this->activeTool = tool;
+
+    this->fireToolChanged();
+}
```
idotobi, line 493: @LittleHuba : This could be quite big performance improvement as this function is called in every call of `PenInputHandler::actionMotion` via `changeTool` which is basically called continuously while drawing for reasons I haven't figured out yet (see current discussion in https://github.com/xournalpp/xournalpp/pull/2393/files#diff-a03baeadae3022a320f793c7565ed67b183f6755149f324133430b629ee8956bL172). If you know why `changeTool` needs to be called throughout the Motion that would be great. If not I'll play around with it.

src/control/ToolHandler.cpp
```diff
@@ -446,13 +483,20 @@ void ToolHandler::loadSettings() {
...
 
 void ToolHandler::pointCurrentToolToToolbarTool() {
-    this->currentTool = this->toolbarSelectedTool;
+    if (this->activeTool == this->toolbarSelectedTool)
+        return;
+    this->activeTool = this->toolbarSelectedTool;
 
-    this->listener->toolColorChanged(false);
-    this->listener->toolSizeChanged();
```
idotobi, line 455: These two calls are actually part of of `fireToolChanged`. Hence they are superfluous.

src/control/ToolHandler.cpp
LittleHuba: Just gave my two cents on that comment. You are welcome to play around with that though. If you manage to reduce those calls (definitely possible) it would help a lot.

idotobi:
The last refactor https://github.com/xournalpp/xournalpp/pull/2339/commits/e9160afe5c7ef8eb0ce91da35065cfacab045141 should be the last big change on this PR. However, I will still test it a bit to ensure I didn't oversee any problem as I changed quite a bit of code again.

@David96: The fixes for #2393 are also included now and I even managed to reduce the calls to fireToolchanged() further :innocent: . If you could try out this PR to check if this still works for you as well that would be great.

@LittleHuba : As there are no further changes planned it would be nice if you could have a second cross check. Especially the new `InputUtils` would be good to check because I'm not sure if they way I used `#include` is correct (after my last commit unexpectedly failed :sweat_smile: ).

src/control/ToolHandler.cpp
```diff
@@ -242,8 +259,18 @@ auto ToolHandler::getThickness(ToolPointer toolPointer) -> double {
     return 0;
 }
 
-void ToolHandler::setSize(ToolSize size, ToolPointer toolPointer) {
-    Tool* tool = getToolPointer(toolPointer);
+void ToolHandler::setSize(ToolSize size) {
+    Tool* tool = this->activeTool;
+    if (size < TOOL_SIZE_VERY_FINE || size > TOOL_SIZE_VERY_THICK) {
+        g_warning("ToolHandler::setSize: Invalid size! %i", size);
```
LittleHuba, line 265: Please clamp the value only. Somebody might switch from very low to (too) high. Best would be to still end up as high as possible in this scenario. If we get a bug here then it would be low priority instead of the high priority for not changing tool size at all, as would occur with your current code.

src/control/ToolHandler.cpp
```diff
@@ -270,43 +297,57 @@ void ToolHandler::setLineStyle(const LineStyle& style) {
  * 			and therefore should not be applied to a selection
  */
 void ToolHandler::setColor(Color color, bool userSelection) {
-    if ((this->buttonSelectedTool->capabilities & TOOL_CAP_COLOR) == 0) {
+    if (this->activeTool != this->toolbarSelectedTool && (this->activeTool->capabilities & TOOL_CAP_COLOR) == 0) {
```
LittleHuba, line 300: Why not use `toolHandler->hasCapability(TOOL_CAP_COLOR)`?

src/control/ToolHandler.cpp
```diff
@@ -495,15 +541,58 @@ auto ToolHandler::isSinglePageTool(ToolPointer toolPointer) -> bool {
...
+        case Buttons::BUTTON_RIGHT:
+            this->mouseRightButtonTool.reset(new Tool(tool));
+            break;
+        case Buttons::BUTTON_TOUCH:
+            this->touchDrawingButtonTool.reset(new Tool(tool));
+            break;
+        default:
+            g_error("This button is not defined for ToolHandler.");
     }
 }
```
LittleHuba, line 598: Please always add a newline at the end of a file. Nowadays C-Compilers don't really require it anymore but it is still a good thing to do for compatibility. You never know what toolchain somebody tries to use. And it removes that ugly sign...

src/control/settings/ButtonConfig.cpp
```diff
@@ -20,12 +20,34 @@ auto ButtonConfig::getDrawingType() -> DrawingType { return this->drawingType; }
 
 auto ButtonConfig::getAction() -> ToolType { return this->action; }
 
-void ButtonConfig::acceptActions(ToolHandler* toolHandler) {
+void ButtonConfig::initButton(ToolHandler* toolHandler, Buttons button) {
```
LittleHuba, line 23: Shouldn't that class be called `Button`?

src/control/settings/ButtonConfig.h
```diff
@@ -45,4 +49,4 @@ class ButtonConfig {
 
     friend class Settings;
     friend class ButtonConfigGui;
-};
+};
```
LittleHuba, line 52: Add a newline please :)

src/gui/inputdevices/MouseInputHandler.cpp
```diff
@@ -111,21 +113,22 @@ auto MouseInputHandler::changeTool(InputEvent const& event) -> bool {
...
+    else if (modifier3 /* Right Button */ && !xournal->selection)
+        InputUtils::applyButton(toolHandler, settings, Buttons::BUTTON_RIGHT, pressed, toolChanged, configChanged);
 
-    if (cfg && cfg->getAction() != TOOL_NONE) {
-        toolHandler->pointCurrentToolToButtonTool();
-        cfg->acceptActions(toolHandler);
-    } else {
-        toolHandler->pointCurrentToolToToolbarTool();
+    if (!pressed || !configChanged) {
+        toolChanged |= toolHandler->pointCurrentToolToToolbarTool();
```
LittleHuba, line 126: This is a direct result of applying the button. Hence it should be part of this method. Or why did you leave this here?

src/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -168,24 +169,25 @@ auto StylusInputHandler::changeTool(InputEvent const& event) -> bool {
...
+        InputUtils::applyButton(toolHandler, settings, Buttons::BUTTON_ERASER, pressed, toolChanged, configChanged);
     }
 
-    if (cfg && cfg->getAction() != TOOL_NONE) {
-        toolHandler->pointCurrentToolToButtonTool();
-        cfg->acceptActions(toolHandler);
-    } else {
-        toolHandler->pointCurrentToolToToolbarTool();
+    if (!pressed || !configChanged) {
+        toolChanged = toolHandler->pointCurrentToolToToolbarTool();
```
LittleHuba, line 188: See MouseInputHandler

src/gui/inputdevices/TouchDrawingInputHandler.cpp
```diff
@@ -90,11 +94,12 @@ auto TouchDrawingInputHandler::changeTool(InputEvent const& event) -> bool {
...
     if (cfg && cfg->getAction() != TOOL_NONE) {
-        toolHandler->pointCurrentToolToButtonTool();
-        cfg->acceptActions(toolHandler);
+        InputUtils::applyButton(toolHandler, settings, Buttons::BUTTON_TOUCH, pressed, toolChanged, configChanged);
     } else {
-        toolHandler->pointCurrentToolToToolbarTool();
+        toolChanged = toolHandler->pointCurrentToolToToolbarTool();
     }
 
+    if (toolChanged)
```
LittleHuba, line 102: See MouseInputHandler

src/gui/inputdevices/TouchDrawingInputHandler.cpp
```diff
@@ -72,8 +73,11 @@ auto TouchDrawingInputHandler::handleImpl(InputEvent const& event) -> bool {
 
 auto TouchDrawingInputHandler::changeTool(InputEvent const& event) -> bool {
     Settings* settings = this->inputContext->getSettings();
-    ButtonConfig* cfgTouch = settings->getTouchButtonConfig();
+    ButtonConfig* cfgTouch = settings->getButtonConfig(Buttons::BUTTON_TOUCH);
     ToolHandler* toolHandler = this->inputContext->getToolHandler();
+    bool toolChanged = false;
+    bool pressed = false;
+    bool configChanged = true;
```
LittleHuba, line 80: It is bad style to force callers to reserve space for data they do not require. Add a minimized signature for `applyButton` instead.

src/gui/inputdevices/old/InputSequence.cpp
```diff
@@ -423,13 +428,13 @@ auto InputSequence::changeTool() -> bool {
...
     }
 
-    if (cfg && cfg->getAction() != TOOL_NONE) {
-        h->pointCurrentToolToButtonTool();
-        cfg->acceptActions(h);
-    } else {
-        h->pointCurrentToolToToolbarTool();
+
+    if (!pressed || !configChanged) {
+        toolChanged = h->pointCurrentToolToToolbarTool();
```
LittleHuba, line 433: Same as MouseInputHandler

src/gui/inputdevices/old/InputSequence.cpp
```diff
@@ -391,30 +393,33 @@ void InputSequence::stopInput() {
...
     } else if (button == 3 /* Right Button */ && !xournal->selection) {
-        cfg = settings->getRightButtonConfig();
+        InputUtils::applyButton(h, settings, Buttons::BUTTON_RIGHT, pressed, toolChanged, configChanged);
     } else if (gdk_device_get_source(device) == GDK_SOURCE_ERASER) {
         penDevice = true;
-        cfg = settings->getEraserButtonConfig();
+        InputUtils::applyButton(h, settings, Buttons::BUTTON_ERASER, pressed, toolChanged, configChanged);
     } else if (cfgTouch->device == gdk_device_get_name(device)) {
-        cfg = cfgTouch;
+        pressed = true;
```
LittleHuba, line 417: Very obscure way of calling some specific methods later on. Just call them directly here.

LittleHuba (changes requested):
So I had time to go through this with a finer comb than last time. So don't be shocked by there being a lot more comments. Your code is mostly good but there are some areas that need some more work. 


Some general advice I want to give you and also other devs out there:
Try to make your code as simple as possible. Sometimes directly calling a method in multiple branches is better than condensing code by setting some (obscure) variables to trigger the call somewhere later in the function. Doing it via such variables forces a reader to go through the whole method body to understand what actually happens. Further setting the variable in different lines might trigger different stuff, which is easily missed. It's not about keeping code short but about keeping it straight forward. Shortening is done by compiler optimizations way better than you could ever hope for in a language such as C.

Also please document your code some more. I know this is generally what devs try to avoid at all cost but it is a crucial part of keeping software maintainable. Tend to cluster your code in blocks and add a comment above those blocks describing in a concise but thorough way what the following lines do and why it is required. The only lines not requiring a comment are lines that are so selfexplainatory that you could just write the line of code itself as its comment. Those comments are also a good point in time for you to rethink if you missed anything or might even do it in a better way. Often you see that the code you wrote actually does not do what you actually wanted to achieve.

Also signatures should always have a comment describing what a method does and what parameters are for.

I'm aware that the current code base absolutely lacks such comments but this is THE reason why we have spaghetti code, code duplication and tons of bugs. So it would be great if we could leave that place and head for the light 😉

src/gui/inputdevices/InputUtils.h
```diff
@@ -1,5 +1,3 @@
-#pragma once
```
LittleHuba: **NEVER NEVER NEVER EVER remove the include barrier in a header file!!!**

I can't stress that point enough.

If you have to do that, this means you have circular dependencies in your headers. You may fix them by predeclaring structs or classes in this header. If you just remove the include barrier you might end up with duplicate definitions of classes or structs with really obscure compiler warnings. Believe me that is something you do not want.

David96:
Thanks, this works for me (regarding #2393).

src/control/Control.cpp
```diff
@@ -1803,19 +1806,14 @@ void Control::toolLineStyleChanged() {
...
-void Control::toolColorChanged(bool userSelection) {
+
+void Control::toolColorChanged() {
     fireActionSelected(GROUP_COLOR, ACTION_SELECT_COLOR);
     getCursor()->updateCursor();
+}
 
-    if (userSelection && this->win && toolHandler->getColor() != Color(-1)) {
+void Control::changeColorOfSelection() {
+    if (this->win && toolHandler->getColor() != Color(-1)) {
```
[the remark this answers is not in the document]
idotobi: You're right. This is actually code I just adapted when splitting the function in it's two functionalities. But I should have checked that could while I was at it :sweat_smile: .
Will be changed :+1:

src/control/Control.cpp
```diff
@@ -2468,6 +2466,19 @@ void Control::applyPreferredLanguage() {
 #endif
 }
 
+void Control::initButtonTool() {
+    vector<Buttons> buttons{Buttons::BUTTON_ERASER, Buttons::BUTTON_STYLUS, Buttons::BUTTON_STYLUS2,
+                            Buttons::BUTTON_MIDDLE, Buttons::BUTTON_RIGHT,  Buttons::BUTTON_TOUCH};
```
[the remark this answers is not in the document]
idotobi: I used the names that were already in SettingsEnum to not add to much changes in this already quite big PR. But I can do that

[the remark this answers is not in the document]
idotobi: I'll update `BUTTON_STYLUS` and `BUTTON_STYLUS2` to `BUTTON_STYLUS_ONE` and `BUTTON_STYLUS_TWO` respectively as this was also something I didn't like to much about the previous naming scheme.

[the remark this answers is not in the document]
idotobi: Should I also rename `BUTTON_ERASER` to `BUTTON_STYLUS_ERASER` while I'm at it, or does that not make sense?

src/control/Tool.cpp
```diff
@@ -12,6 +12,24 @@ Tool::Tool(string name, ToolType type, Color color, int capabilities, double* th
     setColor(color);
 }
 
+Tool::Tool(Tool* t) {
+    this->name = t->name;
+    this->type = t->type;
+    if (t->thickness) {
+        this->thickness = new double[5];
```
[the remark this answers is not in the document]
idotobi: Nice! Thanks for sharing that link.
This actually also helps to get that magic number out of the `ToolHandler.cpp`.

One question though:  Right now I've added this as a `public static int toolSizes = 5` to `Tool`, which in this case makes sense. If it would be a generic const. what the bes practice to namespace it to avoid naming conflicts?

```cpp
class Tool: public ToolBase {
public:
    Tool(string name, ToolType type, Color color, int capabilities, double* thickness);
    Tool(Tool* t);
    virtual ~Tool();

    static const int toolSizes = 5;
```

src/control/ToolHandler.cpp
idotobi: Sounds good. I'll use `std::clamp` from `<algorithm>` and still raise the warning to make it not fully silent.

src/control/ToolHandler.cpp
idotobi: Honestly, I think when I was writing that piece of Code I probably had forgotten about that function :sweat_smile:

src/control/ToolHandler.cpp
idotobi: I adapted the code according to the comments. It took a while but I think now `fireToolChanged()` is almost only called when one would expect.

src/control/ToolHandler.cpp
idotobi: Well, not coming from the cpp world I can definitely add this to my TIL, thanks :pray: 

In case someone else stumbles upon this: This is how you set it in vscode https://stackoverflow.com/a/54270768.

src/control/settings/ButtonConfig.h
idotobi: Done :heavy_check_mark:

src/gui/inputdevices/InputUtils.h
idotobi: This was exactly the pointer I was missing. Being not that familiar with the cpp toolchain yet, I wasn't shure while the linking was throwing an error with `undefined reference`. I guess this was a result of a circular dependency then.

It felt weird that suddenly withouth the `#pragma` it worked that's why I asked explicitly for feedback that. Thanks. And thank you even more for the general advice :pray: .

I guees I'll find a better place then for this util function :detective:

src/control/Control.cpp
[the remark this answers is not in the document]
LittleHuba: Better touch everything that is connected in one PR than have two PRs for one change. I'd leave the eraser as it is as it may be theoretically a different device (even though it is often integrated with the stylus).

src/control/Tool.cpp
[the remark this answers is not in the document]
LittleHuba: `static const int` makes here most sense as the variable is actually strongly connected to the class. Hence it should be a member of the class. You would make it a `const int` without the class association if it would be some global array. In the end it will be handled the same way by the compiler (shifted to the text section of the binary). So your solution is best in our case.

src/control/ToolHandler.cpp
LittleHuba: Yeah exactly! It should not be fully silent but as it is only a warning it should still do mostly what the user would expect.

src/gui/inputdevices/InputUtils.h
LittleHuba: The C compiler goes through your source code (.c file) and includes headers mentioned with `#include`. This operation is literally pasting the content of the file at the line of the macro. C itself has the restriction that all structs, classes and signatures, used for an invocation, must be declared priorly. This is because the compiler stores each struct or signature in a table to match jumping addresses. The actual definition may then follow at some later point. In your case this means you can just predeclare the necessary structs, classes and functions within the header. 

https://www.geeksforgeeks.org/what-are-forward-declarations-in-c/

idotobi: One question in general: Would you consider it good practice to have a `#pragma once` at the top of every header file, or should it only be used in specific scenarios.

idotobi:
@LittleHuba : Thank you very much for your  detailed review and comments. I didn't manage to get through all yet but I'll hope I'll get to it on the weekend. On the things I already incorporated locally I left a comment.

Regarding your general comments I fully agree with both of them. One could basically summarize them with the phrase "Write code for humans and not machines". I generally try to write code in the way that comments are not necessary and rather wrap something in a function that has a speaking name as function-/variable-names are basically compiled comments, but I noticed that when using `cpp` (not my strongest programming language) the hurdle to write extra functions still sometimes keeps me from following that ideal.
In anyway, I shouldn't use it as excuse for not writing comments at all :sweat_smile: .

Regarding documenting code signatures what is the best practice in `cpp`:
* Should I document the function in the header file, the source file, or both (I've seen different things in the code base)
* What is the best format for documenting signatures? Is there a standardized way that we try to follow for this codebase?

LittleHuba:
You are very welcome! Happy to show you help you out with that stuff.

> Regarding documenting code signatures what is the best practice in cpp:

C++ does not have any convention for comments. Normally you write signature comments and comments for anything that is accessible from outside into the header. That part is not implementation specific and the interface of a class should stay the same even if you modify internal stuff. Implementation specific stuff (line comments) are part of the source file. So kind of the same as with all the code in C++.

> What is the best format for documenting signatures?

C++ does not have a common format for comments. Javadoc is the one used in most projects and has best support in IDEs so stick with that one.

src/control/settings/ButtonConfig.cpp
idotobi: Agreed seems also more natural to me.

src/gui/inputdevices/MouseInputHandler.cpp
idotobi: I left it at this place because the logic is actually a little bit more subtle. But I take that comment that the structure is not clear enough yet :sweat_smile: 
I think I found a quite clear solution now.

src/gui/inputdevices/StylusInputHandler.cpp
idotobi: Will be adressed.

src/gui/inputdevices/InputUtils.h
LittleHuba: Every header should have a `#pragma once` or be encapsulated in a

```
#ifdef __HEADER_XYZ__
#define __HEADER_XYZ__

...

#endif
```

It is crucial that headers are only ever included once per source file. As generally header files reference other header files this is not always the case. Hence what you do is to tell the preprocessor to only include a file once even if it is referenced multiple times.

src/gui/inputdevices/MouseInputHandler.cpp
LittleHuba: > I left it at this place because the logic is actually a little bit more subtle.

Please add comments until this becomes clear. Left this way I would simplify this when I next touch the code. Another good reason to add comments 😉

src/gui/inputdevices/old/InputSequence.cpp
LittleHuba: You can actually ignore this as it is part of the old input system which is listed for removal. See #2308

LittleHuba: Thinking of that - it might be best if we remove the old input system first and then merge your PR.

rolandlo: In PR #2308 @Technius wrote:

>  I think this should be merged after we release a beta/milestone version of current master so that the transition is not so sudden.

I don't think this PR (#2339) should wait until a beta/milestone version of current master is released. So you two @Technius @LittleHuba should agree on when the old input system gets removed.

idotobi (commented):
@LittleHuba : I think I adressed all your comments now. Feel free to have a second check. I'll use this version as my daily driver as of now to check for any regressions.

idotobi:
The recent commit https://github.com/xournalpp/xournalpp/pull/2339/commits/61ffd13467c0749d24cb872171f2a7555812d0eb should also fix #1797 as I found the right place during a refactor.

src/gui/inputdevices/InputUtils.h
```diff
@@ -1,9 +1,43 @@
...
+     * @return false if the tool was not changed as it already was pointing to the correct tool.
+     */
+    static bool applyButton(ToolHandler* toolHandler, Settings* settings, Button button);
+
+    /**
+     * @brief Warn the user in case of Questionable Settings in for the Touchscreen "Tool"
+     * Warn the user in case all of the following are true:
+     *  - Disable Drawing for this device is set for Touchscreen
+     *  - Tool - don't change is set for Touchscreen
+     *  - current Tool in the toolbar is a Drawing tool (i.e. Pen/Highlighter/Eraser)
```
LittleHuba, line 35: Sure that this is the combination of settings where this happens? I thought if we disable drawing for touchscreen then it should not draw at all (independent of which tool is selected). But maybe I have that wrong in my head.

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -120,16 +122,14 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
...
 
-
         xournal->view->clearSelection();
-        if (changeTool(event)) {
-            // Do not handle event in any further way to make click only deselect selection
-            return true;
-        }
+        changeTool(event);
+        // stop early to prevent drawing when clicking outside of the selection with the intention of deselecting
+        if (toolHandler->hasCapability(TOOL_CAP_SIZE))
```
LittleHuba, line 131: I assume you actually check here for whether the event originates from a pen or highlighter. If this is the case, this is definitiely the wrong method.Checking via side-effects is always bad since it breaks very easily as nobody expects this kind of usage. Instead check the toolType as is done in line 108.

src/gui/inputdevices/old/InputSequence.cpp
```diff
@@ -249,14 +249,14 @@ auto InputSequence::actionStart(guint32 time) -> bool {
...
+            return isEventHandled;
         }
 
 
         xournal->view->clearSelection();
-        if (changeTool()) {
-            return true;
-        }
+        // stop early to prevent drawing when clicking outside of the selection with the intention of deselecting
+        if (toolHandler->hasCapability(TOOL_CAP_SIZE))
```
LittleHuba, line 258: Same as in PenInputHandler

src/gui/inputdevices/old/InputSequence.cpp
LittleHuba: Then this one first as it is not a breaking change but a fix. But #2308 should follow with small delay as this PR most likely breaks the old system even more.

src/gui/inputdevices/MouseInputHandler.cpp
idotobi: Did you check the latest version I pushed to this PR?
I think it's quite clean there, or would you still add comments to it?

```cpp
void MouseInputHandler::changeTool(InputEvent const& event) {
    Settings* settings = this->inputContext->getSettings();
    bool isClickOnSelection = this->inputContext->getXournal()->selection;
    bool toolChanged = false;

    if (modifier2 && !isClickOnSelection)
        toolChanged = InputUtils::applyButton(toolHandler, settings, Button::BUTTON_MOUSE_MIDDLE);
    else if (modifier3 && !isClickOnSelection)
        toolChanged = InputUtils::applyButton(toolHandler, settings, Button::BUTTON_MOUSE_RIGHT);
    else
        toolChanged = toolHandler->pointActiveToolToToolbarTool();

    if (toolChanged)
        toolHandler->fireToolChanged();
}
```

src/gui/inputdevices/old/InputSequence.cpp
idotobi: Just to be sure this is not missed: 
The addition of the fix for "Drawing when trying to unselect" (#1797) that I added in one of the last commits could be experienced as breaking change by some users even though in my opinion it's rather a fix to a usability issue. Or at least it is for my usage of xournalpp.
I would still keep it in but in case you would rather want to have it separately I can also revert that fix.

src/gui/inputdevices/PenInputHandler.cpp
idotobi: It's actually the Pen, Highlighter and Eraser :sweat_smile: 
But you're of course right. This was again the old habit of trading shortness for expressiveness :see_no_evil:. Will be changed.

src/gui/inputdevices/old/InputSequence.cpp
idotobi: Will be changed :smile:

src/gui/inputdevices/InputUtils.h
idotobi: I just double checked and I basically left the logic as is (maybe apart from not returning early which only complicated the code).

This is the corresponding piece of code previous to my changes is:

```cpp
        // If an action is defined we do it, even if it's a drawing action...
        if (cfg->getDisableDrawing() && cfg->getAction() == TOOL_NONE) {
            ToolType tool = toolHandler->getToolType();
            if (tool == TOOL_PEN || tool == TOOL_ERASER || tool == TOOL_HILIGHTER) {
                g_message("ignore touchscreen for drawing!\n");
                return true;
            }
        }
```

To be honest I don't fully understand this feature anyway but at least the comment that was there initially suggests that this behaviour was intentional.

src/gui/inputdevices/PenInputHandler.cpp
idotobi: When working with the code the repetition of  `tool == TOOL_PEN || tool == TOOL_ERASER || tool == TOOL_HILIGHTER` seemed like to much code duplication for my taste, so I created a function `isDrawingTool` for `Tool` and a proxy for `ToolHandler`. That should give the best of both worlds :crossed_fingers:

src/gui/inputdevices/old/InputSequence.cpp
LittleHuba: For me this is also a fix rather than a change. We added other fixes you could see as "breaking" changes between earlier minor versions so no issue here.

src/gui/inputdevices/InputUtils.h
LittleHuba: Well that is kind of my bad as I did not comment this part good enough as it seems. What this code does is drop all inputs from the touchscreen if drawing via touchscreen is not activated. You might want to take a look into `cfg->getDisableDrawing()` as it is not named really well and might have warped logic.

src/control/ToolEnums.cpp
```diff
@@ -141,8 +141,8 @@ auto toolTypeFromString(const string& type) -> ToolType {
     if (type == "eraser") {
         return TOOL_ERASER;
     }
-    if (type == "hilighter") {
```
LittleHuba, line 144: Leave this case in to fix following breaking change.

ui/toolbar.ini
```diff
@@ -17,7 +17,7 @@
 # Color: COLOR(0xffffff),COLOR(0xffff00),COLOR(0xff8000),COLOR(0xff00ff),COLOR(0x00ff00),COLOR(0x00c0ff),COLOR(0x808080),COLOR(0x008000),COLOR(0xff0000),COLOR(0x3333cc),COLOR(0x000000),COLOR_SELECT
 #  Notice: This are the default Xournal colors, each other color in HEX can also be used, eg COLOR(0x12ABCF);
 #
-# Tools: ERASER,PEN,HILIGHTER,IMAGE,TEXT,IMAGE,SELECT,SELECT_REGION,SELECT_RECTANGLE,VERTICAL_SPACE,HAND,DEFAULT_TOOL
+# Tools: ERASER,PEN,HIGHLIGHTER,IMAGE,TEXT,IMAGE,SELECT,SELECT_REGION,SELECT_RECTANGLE,VERTICAL_SPACE,HAND,DEFAULT_TOOL
```
LittleHuba, line 20: This is a breaking change. See above comment to fix.

src/control/ToolEnums.cpp
idotobi: Could we allow for both spellings during Input and only use the new/correct spelling as output?
This way we would have a smooth transition for old users and new users would already use the corrected spelling.

ui/toolbar.ini
idotobi: I would propose to still fix it here but apply the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) (see upper comment https://github.com/xournalpp/xournalpp/pull/2339#discussion_r531128098) in the parsing. Would that work, or am I missing something.

src/gui/inputdevices/InputUtils.h
idotobi: Hmm, not sure I understand. `getDisableDrawing()` seems to be quite simple actually.

```cpp
auto ButtonConfig::getDisableDrawing() const -> bool { return this->disableDrawing; }
```

What I don't understand in the above logic is: Why is it not sufficient to check only for `if (cfg->getDisableDrawing())` if the drawing should be simply disabled when that flag is set. In the current implementation (even prior to my change) setting a drawing tool in the Touch config would allow to draw again or having for example having the hand tool selected in the toolbar.

Is this really the expected bahviour? The early stopping I can easily implement.

idotobi: One cpp noob question:

I was looking throught the usages of `disableDrawing` and I was wonering why it only has a getter but no setter.

I also checked and to my understanding it is a private variable https://github.com/xournalpp/xournalpp/blob/ddbb7c6b12b4bca61be65281219767b52e67e0ae/src/control/settings/ButtonConfig.h#L41

However, there are pieces of the code (https://github.com/xournalpp/xournalpp/blob/c0855853aa49000c14c61257930cb40cc4503e34/src/control/settings/Settings.cpp#L675) where it is still set directly. I always thought that is not possible? What am I missing here?

David96: Look up what a friend class is ;)

ui/toolbar.ini
LittleHuba: Yes that is what I meant. Read both versions but only write the correct one.

src/gui/inputdevices/InputUtils.h
LittleHuba: Best would be if you talk with the author of #2435 so your modifications match in this area. What I implemented back then was what the old system provided in functionality. There is definitely room for improvement.

src/control/ToolEnums.cpp
LittleHuba: Yes that is what I meant with my comment.

src/gui/inputdevices/InputUtils.h
LittleHuba: > Look up what a friend class is ;)

To tell the truth I do not really like friend classes. They are a good tool for unit testing as you can check for correctness even in the private state of a class but in all other cases they are quite often a sign for bad design choices. Needing access to private state of a class is just a badly designed interface.

idotobi: Hi @personalizedrefrigerator,

Looking at the files changed in your PR I don't see that we should get a conflict in functionality. But I think @LittleHuba is right, and it's better to double check.
I think the two biggest changes that are contained in my PR that could potentially lead to conflicts are:
 * the `changeTool` methods in the Input Handlers no longer return a bool
 * I refactored out the part in the `TouchDrawingInputHandler` (https://github.com/xournalpp/xournalpp/pull/2339/files#diff-c3385b4accee35899531b66ef3248e28162ee01e5d1d8ca69e5954eaedf8b13eL78)

Also if you could have a look at the comment above this comment-thread and it's implementation (https://github.com/xournalpp/xournalpp/pull/2339/files#diff-cd8df78806cf06e70a7662cb62424d7d95c728d13950575c785befaf9b345fd3R25) that would be great. Maybe you understand what the purpose of these lines was (original before refactoring: https://github.com/xournalpp/xournalpp/pull/2339/files#diff-c3385b4accee35899531b66ef3248e28162ee01e5d1d8ca69e5954eaedf8b13eL83). Otherwise I would for now try to leave the original logic untouched.

personalizedrefrigerator: The only change to `TouchDrawingInputHandler::changeTool` (relative to your changes) should be adding `this->updateKineticScrollingEnabled()` near line `86` (this might lead to a merge conflict, but it should be easy to resolve ☺).

Thank you for checking!


As to `if (cfg->getDisableDrawing() && cfg->getAction() == TOOL_NONE) {`, I am very confused...

`cfg` should be a `ButtonConfig`. [`getDisableDrawing`](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/control/settings/ButtonConfig.cpp#L17) just returns `this->disableDrawing`, which is only set in two places:
 * In the [`ButtonConfig` constructor](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/control/settings/ButtonConfig.cpp#L17) (set to `false`).
 * In [`ButtonConfigGui`](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/gui/dialog/ButtonConfigGui.cpp#L239).
    * Set based on the contents of a checkbox in [`/ui/settingsButtonConfig.glade`](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/ui/settingsButtonConfig.glade#L48).
    * `settingsButtonConfig.glade` is only referenced in `po/*.po` files and [`ButtonConfigGui.cpp`](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/gui/dialog/ButtonConfigGui.cpp#L27).
        * It seems only to be constructed in [`SettingsDialog.cpp`](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/gui/dialog/SettingsDialog.cpp#L146), which is shown in `Control::showSettings`, which is called by `Control::actionPerformed`.
    * The `ButtonConfigGui`s are stored in `SettingsDialog::buttonConfigs` and are only used to call `bcg->saveSettings` on `SettingsDialog::save()`.
    * `ButtonConfigGui` has a [`show` method](https://github.com/xournalpp/xournalpp/blob/dbc96baec2b733b9bc050f8e6d5cc385f3a2116d/src/gui/dialog/ButtonConfigGui.cpp#L190), with the contents `// Not implemented! This is not a dialog!`.

From the above, I am confused. Why does `settingsButtonConfig.glade` exist if `ButtonConfigGui` is never shown? How can one access the settings in `ButtonConfigGui`?

idotobi: @personalizedrefrigerator: 
I played around a bit with the debugger and I think I can explain it.

When starting the application `cfg->disableDrawing` is set in: https://github.com/xournalpp/xournalpp/blob/master/src/control/settings/Settings.cpp#L541-L548

The ButtonConfigGui is used, I think it only says `// Not implemented!` as it is not a standalone dialog but part of the Settings Dialog.
image

Within ButtonConfigGui it is then changed within https://github.com/xournalpp/xournalpp/blob/master/src/gui/dialog/ButtonConfigGui.cpp#L236-L240

So how it is set I think is explained by this.


Regarding the logic @LittleHuba I checked out the master branch before I started contributing and there the `ignore touchscreen for drawin` message was also shown even though the drawing was still happening. I then realized that this might have been a bug and I think I found the right spot where to fix it (https://github.com/xournalpp/xournalpp/pull/2339/commits/c2b94fc765672c250f8fd739496df68b628e9d94#diff-a03baeadae3022a320f793c7565ed67b183f6755149f324133430b629ee8956bR73-R74).

One code style question: For this functionality to work I had to implement all `changeTool`s to return a `bool`. Theoretically this could also be handled by throwing and catching errors, but this might be overseen by others more easily. Is the returning `bool` approach clean?

LittleHuba: Ahh nice that you worked all of that out so quickly. Returning a bool is preferable to throwing errors here as error catching has quite the time penalty attached to it which we would like to avoid here.

idotobi:
@LittleHuba : I'd like to resolve the merge conflict before finally merging the PR. Is it better to merge master back into this branch or to rebase my branch?

src/control/ToolHandler.cpp
idotobi: This is fixed in 38d509654dd5b017f84bf8b17f59b0a907d83638.

src/control/ToolHandler.cpp
idotobi: This was added in 38d509654dd5b017f84bf8b17f59b0a907d83638 and will be automatically applied in every code I touch :innocent:

src/control/settings/ButtonConfig.cpp
idotobi: Fixed in 38d509654dd5b017f84bf8b17f59b0a907d83638. Please resolve if you agree with the solution.

src/gui/inputdevices/MouseInputHandler.cpp
idotobi: Fixed in 38d509654dd5b017f84bf8b17f59b0a907d83638. Please resolve if you like the solution :)

src/gui/inputdevices/old/InputSequence.cpp
idotobi: Please resolve if you think this is fine :) (The cleanup happened in38d509654dd5b017f84bf8b17f59b0a907d83638)

src/gui/inputdevices/InputUtils.cpp
```diff
@@ -22,13 +22,15 @@ bool InputUtils::applyButton(ToolHandler* toolHandler, Settings* settings, Butto
     return toolChanged;
 }
 
-void InputUtils::warnIfQuestionableTouchDrawingSettings(ToolHandler* toolHandler, Settings* settings) {
+bool InputUtils::touchDrawingDisallowed(ToolHandler* toolHandler, Settings* settings) {
     ButtonConfig* cfg = settings->getButtonConfig(Button::BUTTON_TOUCH);
     if (cfg->getDisableDrawing() && cfg->getAction() == TOOL_NONE && toolHandler->isDrawingTool()) {
         g_message("Ignoring touchscreen for drawing.\n"
```
LittleHuba, line 28: Could you modify this to only ever print this message once per instance? Otherwise the console might be flooded and printing to the console is quite slow in comparison.

LittleHuba (commented):
A few more nitpicks and then this is good to go.

LittleHuba:
> (Update: It seems that something, without my interaction resolved the merge conflict. It might be that [e46cb82](https://github.com/xournalpp/xournalpp/commit/e46cb82e4055002968063ed81508c7234e7878ec) caused this but I don't understand yet how 🤔 )

Cool that it resolved itself. You should still check the resulting GUI though. Git uses line based conflict resolution which is good for text document but might be still problematic for something like the GUI description. Your element might appear somewhere else or the GUI might not be buildable after all. 

> @LittleHuba : I'd like to resolve the merge conflict before finally merging the PR. Is it better to merge master back into this branch or to rebase my branch?

Generally speaking it is always better to rebase your work as you then do not touch the previous already merged code. If you merge it into your PR you might introduce bugs in other features just through a mistake while merging.

src/gui/inputdevices/MouseInputHandler.h
```diff
@@ -33,10 +33,12 @@ class MouseInputHandler: public PenInputHandler {
      *  - no button is pressed
      *  - or the settings for that button are "No Toolchange"
      *  - or the click is happening on top of a selection on the canvas
-     * this function ensures that the tool selected in the toolbar resmains selected.
+     * this function ensures that the tool selected in the toolbar remains selected.
      *
      * @param event
+     * @return true if tool was changed successfully
+     * @return false never
```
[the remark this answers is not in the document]
idotobi: Makes sense. Is done in latest commit.

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -139,7 +138,7 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
     }
 
     // not handled
```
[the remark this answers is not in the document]
idotobi: True. I removed it.

src/gui/inputdevices/InputUtils.cpp
idotobi: Done :)

idotobi:
> Generally speaking it is always better to rebase your work as you then do not touch the previous already merged code. If you merge it into your PR you might introduce bugs in other features just through a mistake while merging.

You're right rebasing is the cleanest the only thing I don't like is that one has to force push and thus change the history of the reviews that were already conducted. I guess this is the lesser of two evils.

> Cool that it resolved itself. You should still check the resulting GUI though. Git uses line based conflict resolution which is good for text document but might be still problematic for something like the GUI description. Your element might appear somewhere else or the GUI might not be buildable after all.

So I guess I'll rebase then and check if it still works.

idotobi:
@LittleHuba : I rebased and I didn't find any unexpected behaviour. From my side it would be ready to merge.

LittleHuba:
> > Generally speaking it is always better to rebase your work as you then do not touch the previous already merged code. If you merge it into your PR you might introduce bugs in other features just through a mistake while merging.
> 
> You're right rebasing is the cleanest the only thing I don't like is that one has to force push and thus change the history of the reviews that were already conducted. I guess this is the lesser of two evils.

There is a faction supporting merges and one supporting rebases. As you already discussed in your answer both have their drawbacks but I think the ones of rebasing are less, as you only mess up your own work and know hopefully how to fix it again. With merging you have to fix work of others that you do not know the intricacies of. So merging is more error prone and doesn't work well if there are a lot of people working together on the same code base.

One further thing:
I don't really feel comfortable squashing all your commits as they take care of various different issues. Therefore it would be great if you could do one more rebase run where you squash the commits that take care of the same issue. Also try to associate formatting fixes with the commits where you changed those lines/files.

LittleHuba (approved):

Febbe:
Great work! With that fix we get closer to release 1.1.0 finally.

PR #2290, opened 2020-09-29
fix utf8 problems on windows, loading files with special characters

Febbe:
Fixes #2280

Issue #2280, opened 2020-09-26
Problem with Files names containing accents in Windows 10


MichelReims:
OS : Windows 10
Version of Xournalpp : xournalpp-1.1.0+dev-nightly.20200924-windows.exe

Problem :
Under windows 10, when I double-click on a file whose name contains accents the pages are blank.
Under Windows 10, when I open it from menu "File->Open", I have the message "Could not open file"

When I rename the file without accents, there is no problem.

[Corrige.zip](https://github.com/xournalpp/xournalpp/files/5286274/Corrige.zip)

Febbe:
It might be, that I fixed that unintentionally yesterday with this commit: https://github.com/xournalpp/xournalpp/commit/2a3581c5abde46464ce3f95c111225744d300abf
The PR: https://github.com/xournalpp/xournalpp/pull/2276
Can you test it? You could also just wait for the new nightly

MichelReims:
I have compiled the last version. The filename "Corrigé.xopp" becomes "corrigÃ©.xopp"

Febbe:
Do you know your encoding, I'll take a closer look into this, but I don't have a Windows OS till Wednesday

MichelReims:
I don't know where I find it.
I am in France (western europe)

Febbe:
> I don't know where I find it.
> I am in France (western europe)

Ok I am also located in western Europe, I think I can reproduce and fix this quite fast as soon I have time. Thank you for your report. If you want to acquire some experience in c++ you could also try to fix it, this is very easy to fix.

MichelReims:
in previous versions of xournalpp,  I didn't have this problem when I compiled

Febbe:
We made a huge refactoring regarding paths. I think that's the reason for the problems.
Eevery progress can always have (temporary) regressions and there is mostly no way to prevent that completely :/ .

MichelReims:
I understand

MichelReims:
Sorry. I have no experience in c++

MichelReims:
I think there are still issues with annotated PDF files that contain accents

PR #2346, opened 2020-10-27
Fix page layout update

### LINKED ISSUE #2289: I come back to the top of the document when I add a new page at the end of a document

rolandlo:
Try once more to fix the page layout updating. I had to open a new PR, because I messed up the old one (#2312).

This PR should fix #2289 and #1926 about focus shift to the beginning of the document after inserting a page. It may also fix the freeze from #2211 and #2144 when inserting a blank page in a pdf document.

The page layout now gets updated immediately after inserting the page. That only happens after some time (usually too late) without this fix. It also removes an unnecessary call to firePageSelected, since this action is called from scrollToPage anyway.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1224,18 +1224,18 @@ void Control::insertNewPage(size_t position) { pageBackgroundChangeController->i
 
 void Control::insertPage(const PageRef& page, size_t position) {
     this->doc->lock();
-    this->doc->insertPage(page, position);
+    this->doc->insertPage(page, position);  // insert the new page to the document and update page numbers
     this->doc->unlock();
+
+    // notify document listeners about the inserted page; this creates the new XojViewPage, recalculates the layout
+    // and creates a preview page in the sidebar
     firePageInserted(position);
 
     getCursor()->updateCursor();
 
-    int visibleHeight = 0;
-    scrollHandler->isPageVisible(position, &visibleHeight);
-
-    if (visibleHeight < 10) {
-        Util::execInUiThread([=]() { scrollHandler->scrollToPage(position); });
-    }
+    // make the inserted page fully visible (or at least as much from the top which fits on the screen),
+    // and make the page appear selected
+    scrollHandler->scrollToPage(position);
     firePageSelected(position);
 
     updateDeletePageButton();
--- a/src/gui/Layout.h
+++ b/src/gui/Layout.h
@@ -111,6 +111,10 @@ protected:
 private:
     static void checkScroll(GtkAdjustment* adjustment, double& lastScroll);
 
+    /**
+     * Calls either the ScrollHandlingGtk or (when the Touch Workaround is enabled) the ScrollHandlingXournalpp
+     * method to set the layout size by updating the horizontal and vertical GtkAdjustments
+     */
     void setLayoutSize(int width, int height);
 
 private:
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -471,7 +471,7 @@ auto XournalView::getVisibleRect(XojPageView* redrawable) -> Rectangle<double>* 
 auto XournalView::getHandRecognition() -> HandRecognition* { return handRecognition; }
 
 /**
- * @returnScrollbars
+ * @return Scrollbars
  */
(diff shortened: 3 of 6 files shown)


rolandlo:
Scrolling to the new page after inserting a page should work perfectly now. @Rmano Please check it out!

The problem, that remained from my previous attempt was, that the layout size change (the upper value of a GtkAdjustment) was only put into a queue and not executed immediately. So I changed that to fix it. I hope, I didn't create another problem. This needs more testing.

rolandlo:
Previously I believed the call to `firePageSelected` was unnecessary, since it was already called from `scrollToPage` anyway. While testing it turned out that it does make a different for layouts with multiple rows and columns (for instance two row layout). So I added it back again.

rolandlo:
@Febbe, could you please take a look at the code? It seems to work very well now.

MichelReims:
I compile this PR. There is problem with accents in filenames under windows 10 (xopp and anotated files). This is the issue #2280 that as not be solved by #2290

rolandlo:
@MichelReims This has nothing to do with this PR. So please report the problem in issue #2280 or open up a new one.

Febbe:
I did not test it now, but that looks very good. Can you directly back port it to 1.0.20-dev? I think, we should start to back port fixes immediately.

Technius:
For future reference, can you leave some comments in the code explaining why the code is necessary? I always forget how the scrolling code works every time I come back to it...

rolandlo:
Ok, I will add some comments and wait for @Rmano to test the PR out (he announced to do that). I can also back port it to 1.0.20-dev once I have merged the PR into master. I agree that it is a good idea to back port fixes immediately. It will also make it easier to keep track of the necessary changes for the changelog.

Rmano:
Works well almost always. The only problem I got (ok, that could be nitpicking) is that if you add the second page when **all** of the first page is in view, you have no scroll to the second page. After that, or if  the first page is not completely in view it works well.

Peek 2020-10-30 12-13

Rmano:
Another nitpicking: for the N added page, it would be nice that when you need to scroll to show the new page, some of the previous page (N-1)  is still in view; maybe putting the top of the new page at 20% position of the canvas? So you have a bit of view of the old one too.

rolandlo:
Thanks for the feedback @Rmano 
- The reason you observe no scrolling when the page from which you insert is already all in view is that the scrolling is only executed if the visible height of the inserted page is at most 10 pt. So if a little fraction of the inserted page is already visible you have no scrolling. Not sure why we have this condition in the code. I will try out what happens if we remove it.
- Your second suggestion would not work in presentation mode. It would also not work for layouts with multiple rows and columns where the inserted page and the previous page are not necessarily next to each other in the layout. So I would rather not change that.

Could you check, if the PR fixes the freezes observed in #2211? It seems to me to be the case (and would perfectly make sense from the code point of view).

Rmano:
> Could you check, if the PR fixes the freezes observed in #2211? It seems to me to be the case (and would perfectly make sense from the code point of view).

I tried but I can't trigger the bug (to be honest, I never saw it either). I am not sure which difference there is between opening a PDF and "Annotate PDF" thing...

rolandlo:
There are two differences I am aware of:
- If you choose "annotate pdf" you get the "attached mode" checkbox. Enabling it means that the file path of the annotated pdf is relative to the directory of the xopp-file.
- With "annotate pdf" you can annotate a pdf, where you already have a xopp-file, when you want to make annotations in a different xopp-file. If you just open the pdf, you will open the xopp file with the same file name stem.

Ok, if you never got any crashes from inserting empty pages in pdf-documents, then you have nothing to test with that regard.

rolandlo:
Ok, I made Xournal++ always scroll to the inserted page regardless of whether it is already visible without scrolling or not. To me this looks much more natural. .
@Rmano Does that meet your expectations (about your first suggestion)?

Moreover I removed executing the scrolling in the UI thread, since it can be executed right away, now we don't have a race between scrolling and updating the Gtk adjustment upper values.

Rmano:
Yes, I like it now... looks great, wonderful work! Thanks!

rolandlo:
Fixes #2289, fixes #1926, fixes #2211, fixes #2144.

PR #2178, opened 2020-08-20
Update CHANGELOG.md

rolandlo:
I have added most changes since version 1.0.18 got released. The changelog for version 1.0.18 is still missing.

PR #2353, opened 2020-10-30
Updated dependencies and description in debian/control

apandada1:
- Copied the dependencies from the official debian package. Otherwise, the generated debian package (in the PPA) had  dependencies which were unavailable in Ubuntu 18.04, and as a result, the (stable releases) PPA package for 1.0.19 could not be installed in Ubuntu 18.04.
I have verified that these updated dependencies make it possible to use the PPA in Ubuntu 18.04, and I could not notice any regression in either 18.04 or 20.04.

- Also updated suggests field
- Also updated description

This will automatically add current dependencies to the official daily builds PPA.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,8 +8,34 @@ Homepage: https://github.com/xournalpp/xournalpp/
 
 Package: xournalpp
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}
-Description: Xournal++ is a handwriting Notetaking software with PDF annotation support. Supports Pen input like Wacom Tablets.
-
-
-
+Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6
+Suggests: texlive-base, texlive-latex-extra
+Description: Xournal++ - Open source hand note-taking program
+ Xournal++ is a hand note taking software written in C++ with the target of 
+ flexibility, functionality and speed. Stroke recognizer and other parts are 
+ based on Xournal Code, which you can find at sourceforge. 
+ It supports Linux (e.g. Ubuntu, Debian, Arch, SUSE), macOS and Windows 10. 
+ Supports pen input from devices such as Wacom Tablets.
+ .
+ Xournal++ features:
+ .
+  - Support for Pen pressure, e.g. Wacom Tablet
+  - Support for annotating PDFs
+  - Fill shape functionality
+  - PDF Export (with and without paper style)
+  - PNG Export (with and without transparent background)
+  - Allows maping different tools/colors etc. to stylus/mouse buttons
+  - Sidebar with Page Previews with page sorting, PDF Bookmarks and Layers
+  - Enhanced support for image insertion
+  - Eraser with multiple configurations
+  - LaTeX support (requires a working LaTeX install)
+  - Bug reporting, autosave, and auto backup tools
+  - Customizeable toolbar, with multiple configurations
+  - Page Template definitions
+  - Shape drawing (line, arrow, circle, rect)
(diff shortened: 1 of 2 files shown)


Technius:
Thanks for the PR!

Just to clarify, this is for the `1.0.x` branch? We would have to set the dependencies separately for master and `1.0.x`.

apandada1:
I did not know that the dependencies of the master branch are now different.
This should be for the 1.0.x branch. Should I set the target branch to `1.0.x`?

I will check what are the dependencies for the master branch (from the nightly released .deb package), and soon send another PR for master.

apandada1:
If I change target branch to 1.0.x, github will try to merge all commits on `master` to the `1.0.x` branch. 

Should I close this pull request and open another one for the 1.0.x branch?

Technius:
Try rebasing on the `1.0.x` branch.

apandada1:
I will copy the dependencies of the latest nightly build and force push that to my master branch, so that this PR won't be wasted.

I have cherry-picked the commit into my 1.0.x branch. I will send another pull request from my 1.0.x ---> this repo's 1.0.x

Is this a good idea? I am waiting to know your opinion before I open another pull request

Technius:
Sounds good to me! Are these immediately ready for merge?

apandada1:
This PR is **not ready** because it is supposed to be for master, but it has dependencies of 1.0.x. I will force push to correct the dependencies.
#2357 is ready for merge.

apandada1:
I checked that the `nightly` .deb package has the same dependencies which I proposed in this PR.

Technius:
Awesome, thanks for the PR.

Note that we use `+` instead of `-` (i.e. `1.1.0+dev~{revtime}`), so the PPA should also use `+` for consistency.

CC @rolandlo PPA update. We should also update the deb changelog (by copying the revision history in #2178, for example).

apandada1:
Updating the .deb changelog would be a good idea because then one would not have to manually change version number in Launchpad. Anyway, `{debupstream}+dev~{revtime}` will automatically make the version number `1.1.0+dev~{revtime}` (once the `debian/changelog` is updated. This has the added advantage that it will automatically pick up `1.2+dev~{revtime}` when 1.2 is released.

rolandlo:
@Technius You want me to change the launchpad PPA recipe? I think only the owner of the recipe can do that.

rolandlo:
> Awesome, thanks for the PR.
> 
> Note that we use `+` instead of `-` (i.e. `1.1.0+dev~{revtime}`), so the PPA should also use `+` for consistency.
> 
> CC @rolandlo PPA update. We should also update the deb changelog (by copying the revision history in #2178, for example).

The PPA recipe (master build) is now updated to use the `+dev`. We should push the 1.1.0 tag on the master branch, I think, to make it `1.1.0+dev` (instead of `1.0.6+dev`). @LittleHuba Did you have a concern about that and if so, why?

LittleHuba:
Pleite only push that tag once we make the release as the tag signifies a release.

rolandlo:
We can still [move the tag](https://stackabuse.com/git-push-tags-to-a-remote-repo/) once we make the release tough. How can we have the correctly labeled PPA version otherwise?

LittleHuba:
Normally the development branch has no different label than the last official release, as at that point it is not determined which release number the next release will get. The release number is determined by the significance of the changes made at the point of time the release is created. That we have 1.0.6 as the latest tag in our dev-branch is a direct result of our awkward branching logic. It would therefore be better to wait with updating the label until we make the 1.1.0 release from the master branch. After this we should change our branching logic to the more common one where we only work on the master branch and create a release branch where the master branch is merged into when we make a release. Hotfixes are then merged with both the master and release branch, triggering a hotfix release and making sure that the changes are also part of the next major release.

rolandlo:
Ok, I think I got it now. I am looking forward for the simpler branching logic.

apandada1:
I would like to suggest two alternatives of creating a new tag.
1. In the Launchpad recipe, edit `# git-build-recipe format 0.4 deb-version {debupstream}+dev~{revtime}` to 
`# git-build-recipe format 0.4 deb-version 1.1.0+dev~{revtime}`, and edit again when 1.1.1 or 1.2 is be released.
2. Add an entry for 1.1.0+dev in `debian/changelog`, and Launchpad will pick it up automatically.

Issue #2355, opened 2020-10-30
Can't replace an existing PDF on exporting


Ev1lbl0w:
(Please complete the following information, and then delete this line)

**Affects versions :** 1.0.19
 - OS: Arch Linux
 - (Linux only) Desktop environment: X11
 - Which version of libgtk do you use: 3.24.23
 - Version of Xournal++: 1.0.19
 - Installation method: official package from [arch](https://www.archlinux.org/packages/community/x86_64/xournalpp/)

**Describe the bug**
When I try to export a PDF of my notes, and I wish to override an already existing one, the confirmation window appears, but clicking "Override" won't do anything, as if that option meant "Don't Override"

**To Reproduce**
1. Export notes to, for example, `notes.pdf`
2. Export again, overriding file
3. Nothing happens
4. If the file is deleted from the file explorer, then it can be exported w/o issue.

**Expected behavior**
To be able to override a vile

**Screenshots of Problem**
N/A

**Additional context**
N/A

rolandlo:
This doesn't happen in the 1.1.0+dev version.

Ev1lbl0w:
Can confirm this bug is fixed on the latest `master`, closing it...

rolandlo:
It should get fixed in the next stable release (1.0.20) as well, so I will reopen it.

ThamirysOlv:
This is also happening with png files.

BMicraft:
> Can confirm this bug is fixed on the latest `master`, closing it...

This is still not fixed in 1e314821c04d1ba86bdecbadf14e8eb3ee18d230 for me

Joheyy:
Shouldn't this issue be closed because of #2529 ?

PR #2360, opened 2020-10-31
1.0.x - fix page layout update after inserting a new page

rolandlo:
backport of 23aee5dd847d10204b62fd8b406c91888b114a9e

I did this by updating my local 1.0.x branch (via 
```term
git checkout 1.0.x
git fetch upstream
git merge upstream/1.0.x
git push origin 1.0.x
```
and then I cherry-picked via 
```term
git cherry-pick -x 23aee5dd847d10204b62fd8b406c91888b114a9e
```
and resolved all conflicts. 

Is that the way that is done in general? To solve the merge conflicts was tedious, since the git patched the incoming changes mostly at the wrong line numbers.

Anyway I managed to solve all conflicts. I tested out the back-ported fix and it worked fine.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1444,13 +1444,7 @@ void Control::insertPage(const PageRef& page, size_t position)
 
 	getCursor()->updateCursor();
 
-	int visibleHeight = 0;
-	scrollHandler->isPageVisible(position, &visibleHeight);
-
-	if (visibleHeight < 10)
-	{
-		Util::execInUiThread([=]() { scrollHandler->scrollToPage(position); });
-	}
+    scrollHandler->scrollToPage(position);
 	firePageSelected(position);
 
 	updateDeletePageButton();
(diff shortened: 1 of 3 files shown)


Technius:
> Is that the way that is done in general? To solve the merge conflicts was tedious, since the git patched the incoming changes mostly at the wrong line numbers.

Yup, backports are really tedious because of the big formatting changes we made after we split off the branch. When I backport, I usually 1) `git cherry-pick` the desired commit; 2) revert the changes to the HEAD of `1.0.x`; and then 3) rewrite the changes manually. This is usually faster than trying to deal with all the merge conflicts.

Hopefully we can put out 1.1.0-beta1 or 1.1.0 soon so we don't have to deal with this any longer.

Technius (approved):

apandada1:
Then release a new version like 1.0.20?

rolandlo:
Maybe #2355 should be fixed first before releasing a 1.0.20 version.

Issue #2350, opened 2020-10-29
The new background pattern is too distracting under a dark page color.


sammdu:
## Affects versions :
 - OS: Linux/Ubuntu18.04.5
 - GNOME3 X11
 - Which version of libgtk do you use
     - 3.22.30
 - Version of Xournal++: 
     - 1.0.19
 - Installation method: bionic.deb via GitHub releases

## Describe the bug
New paper background (graph, dotted, etc) too distracting on top of a dark background color.

In an older version **this was not the case**. I always used a dark page background and the patterns would be very subtle and nice. The older version that worked well was `1.0.6` through the Unstable Ubuntu PPA:
```
http://ppa.launchpad.net/andreasbutti/xournalpp-master/ubuntu bionic/main amd64 xournalpp amd64 1.0.6-0~202008180544~ubuntu18.04.1
```

## To Reproduce
Choose a dark page background color (I use #111111)
Choose a paper background pattern, such as dotted or graph, but particularly bad with graph:
![image](https://user-images.githubusercontent.com/10665890/97528044-34ab8c00-1983-11eb-9937-9aaf79cbd095.png)

## Expected behavior
The background pattern should be subtle and not distracting:
![image](https://user-images.githubusercontent.com/10665890/97528303-c3200d80-1983-11eb-9ca4-4089a57d86aa.png)

## Suggestions
I know that depending on the light/darkness of the page background, the lightness of the patterns that is preferable will change. So maybe allow the use the set the color for background patterns?

sammdu:
What is interesting is: in the past I was able to use background patterns with both light and dark page backgrounds, and the patterns would always be the right lightness, not distracting. But since upgrading to the new version, This behavior has changed. The background patterns are only subtle under a light background.

rolandlo:
Would PR #2061 solve your problem?

sammdu:
> Would PR #2061 solve your problem?

Looks like it might! Of course I haven't gotten a chance to try it but I hope it will work! I will wait for a newer release.

rolandlo:
Speaking of newer releases: You may want to try out version 1.1.0+dev. Seems it already matches your desired pattern.

sammdu:
Yep. Seems to work!
Although I don't see a UI element to set the foreground color, nor do I see anything in `pagetemplates.ini` (maybe I've overlooked), but I'm guessing a default was hardcoded?
The default works well, though. Thanks!

robinp7720:
[...]

Issue #1853, opened 2020-04-09
Paper background colors


witekcebularz:
**Is your feature request related to a problem? Please describe.**
I have an old, inkjet printer and whenever I use "Graph" paper background (which is essential for me) the print doesn't turn out very good because of the shade of the grid.

**Describe the solution you'd like**
I think it's a cool idea to be able to change the color of the grid itself (from the grey shade to any custom color, or at least to some pre-made ones). This can also apply for the rest of the paper backgrounds. If the paper background has more than one color (Ruled with Vertical Line) it would be nice to specify the second color as well. And also saving the edited background and a way to make it a default.

Other than a useful feature, it can also bring a lot of esthetic value to the program and the documents created with it.

**Describe alternatives you've considered**
If that's impossible for some reason it'd be nice to see the blue colored-grid from the first Xournal back, as an option.

**Additional context**
I'm sorry if that's already a thing, I googled for it and also searched the config files for the appropriate parameter but I couldn't find it. And yes, I know I can change the background to anything I want, but wouldn't it be so much more convenient to have this as a feature?

Technius:
It shouldn't be too hard to add an option to select the line colors.

bhennion:
This is actually already possible, for now only by editing a configuration file.
If you add an entry
```
[myBackgroundName]
name=Ruled with vertical line, custom colors
format=lined
config=f1=0x00ff00,af1=0xff00ff,f2=0xffff00,af2=0x0000ff
```
in the `pagetemplate.ini` file (not sure where to find it on your installed system), you'll get custom colors.
On the config line above:

- f1 is the RGB hexadecimal code for the color of the horizontal lines, on light background colors.
- af1 is the RGB hexadecimal code for the color of the horizontal lines, on dark background colors.
- f2 is the RGB hexadecimal code for the color of the vertical line, on light background colors.
- af2 is the RGB hexadecimal code for the color of the vertical line, on dark background colors.

This is obviously not so handy, but at least it is possible.

witekcebularz:
Oh, thank you, it worked without a problem!

Mine was located in `/usr/share/xournalpp/ui/pagetemplates.ini`
Thanks a lot!

bhennion:
> Mine was located in `/usr/share/xournalpp/ui/pagetemplates.ini` Thanks a lot!

There may even be a user config directory where a copy of this file could be put to get user-wide templates (as opposed to ystem-wide). I don't know.

[...]

PR #2366, opened 2020-11-02
Add option to change primary foreground color to journal menu

robinp7720:
This adds the ability to set config options in the BackgroundConfig class using the ```setValue``` and ```setValueHex``` methods. Using this, the page foreground color can dynamically be changed by updating the config string of the PageType of the XojPage.
This PR only adds the ability to change the primary foreground color. It doesn't add the ability to change the secondary foreground color. I don't believe showing the ability to change the secondary foreground color for all views would be a good idea since most backgrounds only have one foreground color.
Relevant issues: #2350, #1853
I would also ignore #2352 if this were accepted as I don't believe having user configurable foreground colors and "alternative dark bg colors" are very user friendly.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -10,6 +10,7 @@
 #include <gio/gio.h>
 #include <glib/gstdio.h>
 #include <gtk/gtk.h>
+#include <view/background/BackgroundConfig.h>
 
 #include "gui/TextEditor.h"
 #include "gui/XournalView.h"
@@ -1402,7 +1403,12 @@ void Control::changePageForegroundColor() {
     dlg.show(GTK_WINDOW(this->win->getWindow()));
 
     if (auto optColor = dlg.getSelectedColor(); optColor) {
-        p->setForegroundColor(*optColor);
+        PageType pageType = p->getBackgroundType();
+        auto *backgroundConfig = new BackgroundConfig(pageType.config);
+        backgroundConfig->setValueHex("f1", ((uint32_t) *optColor) - 0xff000000);
+        pageType.config = backgroundConfig->toString();
+        p->setBackgroundType(pageType);
+
         firePageChanged(pNr);
     }
 }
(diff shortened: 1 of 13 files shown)


src/control/Control.cpp
```diff
@@ -1378,6 +1382,37 @@ void Control::changePageBackgroundColor() {
...
+    }
+
+    SelectBackgroundColorDialog dlg(this);
+    dlg.show(GTK_WINDOW(this->win->getWindow()));
+
+    if (auto optColor = dlg.getSelectedColor(); optColor) {
+        PageType pageType = p->getBackgroundType();
+        auto* backgroundConfig = new BackgroundConfig(pageType.config);
+        backgroundConfig->setValueHex("f1", ((uint32_t)*optColor) - 0xff000000);
+        pageType.config = backgroundConfig->toString();
```
Technius, line 1409: Won't this leak memory? Am I missing a reason for why you can't just stack allocate it, e.g.
`BackgroundConfig backgroundConfig(pageType.config)`?

src/view/background/BackgroundConfig.cpp
```diff
@@ -66,3 +67,34 @@ auto BackgroundConfig::loadValueHex(const string& key, uint32_t& value) -> bool
...
+bool BackgroundConfig::setValue(const string& key, string value) {
+    this->data[key] = value;
+
+    return true;
+}
+
+bool BackgroundConfig::setValueHex(const string& key, uint32_t value) {
+    char output[6];
+
+    sprintf(output, "%X", value);
```
Technius, line 80: Use `snprintf` instead, since `output` has a fixed size.

Technius (commented):
Just some quick comments for now, will test when I have some free time.

src/control/Control.cpp
robinp7720: I'm still fairly new to c++, so I'm still learning. I've changed the foreground color config setter to use stack allocation for the config parser.

PR #2379, opened 2020-11-08
Customizable color palette using .gpl config file

### LINKED ISSUE #1812: Suggestion for A Better Palette
The default palette provided by Xournal doesn't look good. Here's a video describing it.
https://open.lbry.com/@Fatih109:4/Xournal-Palette:3?r=7YWrkr8hJMjm8335kNGjStx3D2ZoNEHm
As you can see here, almost all of them have the maximum value of saturation. Colors with maximum saturation can attracts attention because of how "flashy" it is but it can also cause eye strain when being seen for a long time because of the same reason. 
As such, almost all of these colors aren't ideal for note taking, where often time you're rereading the note you've taken for a long time.

I have a suggestion for a better color palette heavily inspired by elementary OS color palette, here it is side by side with the old palette.
Xournal Palette

You might think it's worse because it looked more gray-ish and bland but that's precisely why it's better. Because of that, it drastically reduces eye strain. Proof of this, here's a comparison between the same note but with 2 different color. One is from the old palette and one is from the new palette.
Xournal Palette Comparison
Screenshot from 2020-03-30 21 00 02

Feel free to leave any feedback!

idotobi:
Currently this is a draft to eventually resolve #1812 . If you're really curious about this you can already try it out. The main functionality is working. In particular on first start of xournalpp a `palette.gpl` file is created in your respective `.config` folder. This you can either replace or adjust to your desired palette. 

However, note:
 - the parsing might still cause errors
 - there might still be memory leaks
 - if you're palette has less colors than you have `colorToolItems` in your Toolbar xournalpp will show a warning and cycle through the palette as often as required
 - the behaviour in combination with the customize toolbar dialog is not yet consistent


kno10:
Thank you!
I was able to replace the color palette with one that I like much better with your patch.

idotobi:
I updated the PR. It's still not ready for review as there are some cleaning up tasks I need to do first (e.g. writing nice comments). But if you want to try it out I'm always happy about feedback :innocent: .
The biggest change coming with this update is that the customization now allows you to have the same colour multiple times in your toolbars and that you can define the order of the colours using using the toolbar customization window.

:warning: Pleas note, with the latest commit xournalpp automatically migrates all the toolbarsettings to the new toolbar.ini format. In case you have elaborate Colour Settings (special order and non default colours) in your toolbar.ini, this might shuffle your colours a bit.

idotobi:
I'm getting closer to finalizing this, but I'm still a bit confused with the whole toolbar customization code. Maybe you, @Febbe  or @lehmanju , can help me as you've been working on that code previously:

1. Question: Where is this ColortoolItem destructed? To me it looks like it never is.
https://github.com/xournalpp/xournalpp/blob/89021c9828b6aef0bb6d6dda8d8cfb353fcc5c9b/src/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp#L282-L283

2. Question: In the `ToolbarCustomizeDialog` quite a few objects are allocated with `gtk` methods but also there I don't see them destructed or there references even tracked. Is gtk automagically taking care of that or is that something that should rather be implemented?
E.g. https://github.com/xournalpp/xournalpp/blob/89021c9828b6aef0bb6d6dda8d8cfb353fcc5c9b/src/gui/dialog/toolbarCustomize/ToolbarCustomizeDialog.cpp#L275-L279

Febbe:
>     1. Question: Where is this ColortoolItem destructed? To me it looks like it never is.

The pointer is copied into a function at the end of the block you refer.
There it is added into a list. So the pointer is tracked there.
Functions, where such an Object is deleted are:   
In `void ToolMenuHandler::freeDynamicToolbarItems()` ToolMenuHandler.cpp:74   
In `void ToolMenuHandler::removeColorToolItem(AbstractToolItem* it)` ToolMenuHandler.cpp:199
But you are right, this should be refactored to use a unique_ptr also those cast's along the way look quite strange. 
They shouldn't be required.

src/util/NamedColor.cpp
```diff
@@ -0,0 +1,89 @@
...
+};
+
+
+NamedColor::NamedColor(const NamedColor& other):
+        paletteIndex{other.paletteIndex}, isPaletteColor{other.isPaletteColor}, name{other.name} {
+    this->color = other.color;
+    this->colorU16 = other.colorU16;
+}
+
+NamedColor::~NamedColor(){};
```
Febbe, line 51: It would be better to remove all custom defined destructors, copy/move constructors/asignment operators.
Those are created automatically, if all used types also implement those.
Mark the class optionally as final.

src/util/NamedColor.h
```diff
@@ -0,0 +1,136 @@
...
+     * @brief Construct a new NamedColor instance
+     * This automatically sets the isPaletteColor to negative as NamedColors from a color palette can only be
+     * constructed using the input operator.
+     * As it is no paletteColor the Name of the Color will be set to "Custom Color" and the paletteIndex to 0.
+     *
+     * @param color color to set for the custom NamedColor instance
+     */
+    NamedColor(const Color& color);
+
+    ~NamedColor();
```
Febbe, line 68:
```suggestion
```

Febbe: Alternatively mark all as defaulted.
In my opinion, it is not required to write documentation for trivial constructors / destructors.

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+        // call to private variable paletteIndex allowed because of friend class
+        color.paletteIndex = namedColors.size();
+        namedColors.push_back(color);
+        return true;
+    }
+    return false;
+}
+
+
+NamedColor Palette::getColorAt(size_t i) const {
```
Febbe, line 62:
```suggestion
NamedColor const& Palette::getColorAt(size_t i) const {
```

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+NamedColor* Palette::getColorPointerAt(size_t i) {
+    if (i >= namedColors.size()) {
+        i = i % namedColors.size();
+        g_warning("There are more Coloritems in the Toolbar then your Palette defines.\n"
+                  "Hence, cycling through palette from the beginning.");
+    }
+
+    return &namedColors.at(i);
+}
```
Febbe, line 80: I think this function is obsolete/duplicate code, as soon you return the const reference.
Returning a pointer is good, if the function can fail e.g no `Color`` at index `i`, thus returning a nullptr.

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+size_t Palette::size() const { return namedColors.size(); }
```
Febbe, line 82: I would remove those and only return a `std::vector<NamedColor> const&`

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+    std::string line;
+    Header tmp;
+    if (std::getline(str, line)) {
+        std::istringstream iss{line};
+        if (getline(iss, tmp.attribute, ':') && getline(iss, tmp.value, '\n')) {
+            tmp.attribute = StringUtils::trim(tmp.attribute);
+            tmp.value = StringUtils::trim(tmp.value);
+
+            // parsing was successful hence results can be stored permanently
+            header.swap(tmp);
```
Febbe, line 140:
```suggestion
            header = std::move(tmp);
```

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+            str.setstate(std::ios::failbit);
+        }
+    }
+    return str;
+}
+
+void Header::swap(Header& other) {
+    std::swap(attribute, other.attribute);
+    std::swap(value, other.value);
+}
```
Febbe, line 154: If that function is only used to swap in the input stream function, it can be removed.

src/gui/toolbarMenubar/model/ColorPalette.h
```diff
@@ -0,0 +1,157 @@
...
+     */
+    std::string value;
+
+    /**
+     * @brief Swap all properties between two Header instances
+     *
+     * @param other other Header instance to swap properties with
+     */
+    void swap(Header& other);
+};
```
Febbe, line 45: If this struct is only used in on CU I would move it into the source file.

```suggestion
struct Header {

    friend std::istream& operator>>(std::istream& str, Header& header);

    std::string attribute; ///< Attribute name given in gpt file
    std::string value; ///< Value for corresponding attribute in gpt file
};
```

src/util/Color.h
```diff
@@ -137,6 +138,12 @@ constexpr auto Util::GdkRGBA_to_rgb(const GdkRGBA& color) -> Color {
            floatToUIntColor(color.blue);
 }
 
+constexpr auto Util::colorU16_to_rgb(const ColorU16& color) -> Color {
+    return color.red << 16U |  //
+           color.blue << 8U |  //
+           color.green;
+}
+
```
Febbe, line 146:
```suggestion
constexpr auto Util::colorU16_to_argb(const ColorU16& color) -> Color {
    return color.alpha << 24U | 
           color.red << 16U |  //
           color.blue << 8U |  //
           color.green;
}

```

src/util/NamedColor.cpp
```diff
@@ -0,0 +1,89 @@
+#include "NamedColor.h"
+
+#include <sstream>
+
+#include <util/StringUtils.h>
```
Febbe, line 5:
```suggestion
#include "util/StringUtils.h"
```

src/util/NamedColor.cpp
```diff
@@ -0,0 +1,89 @@
...
+auto NamedColor::getName() const -> std::string { return name; };
+
+void NamedColor::swap(NamedColor& other)  // C++03 as this answer was written a long time ago.
+{
+    std::swap(colorU16, other.colorU16);
+    std::swap(color, other.color);
+    std::swap(name, other.name);
+    std::swap(paletteIndex, other.paletteIndex);
+    std::swap(isPaletteColor, other.isPaletteColor);
+}
```
Febbe, line 89: Moving instead of swapping is most likely the faster choice.

src/util/NamedColor.cpp
```diff
@@ -0,0 +1,89 @@
...
+    std::string line;
+    NamedColor tmp;
+    if (std::getline(str, line)) {
+        std::istringstream iss{line};
+        if (iss >> tmp.colorU16.red >> tmp.colorU16.blue >> tmp.colorU16.green && std::getline(iss, tmp.name)) {
+            tmp.name = StringUtils::trim(tmp.name);
+            tmp.color = Util::colorU16_to_rgb(tmp.colorU16);
+            tmp.isPaletteColor = true;
+            /* OK: All read operations worked */
+            namedColor.swap(tmp);  // C++03 as this answer was written a long time ago.
```
Febbe, line 63:
```suggestion
            namedColor = std::move(tmp); 
```

src/control/Control.h
```diff
@@ -338,6 +339,7 @@ class Control:
     ZoomControl* zoom = nullptr;
 
     Settings* settings = nullptr;
+    Palette* palette;
```
Febbe, line 342:
```suggestion
    Palette* palette = nullptr;
```

```suggestion
    std::unique_ptr<Palette> palette;
```

src/control/settings/Settings.cpp
```diff
@@ -596,6 +598,15 @@ auto Settings::load() -> bool {
     loadButtonConfig();
     loadDeviceClasses();
 
+    paletteFilePath = PALETTE_FILE;
+    auto paletteFile = Util::getConfigFile(paletteFilePath);
+    if (!fs::exists(paletteFile)) {
+        Palette::create_default(paletteFile);
+    }
+    this->palette = new Palette(std::move(paletteFile));
```
Febbe, line 606:
```suggestion
    this->palette = std::make_unique<Palette>(std::move(paletteFile));
```

src/control/settings/Settings.cpp
```diff
@@ -596,6 +598,15 @@ auto Settings::load() -> bool {
...
+    this->palette->load();
```
Febbe, line 607: **Question:** is there a failover, if the file couldn't be load?

src/control/settings/Settings.cpp
```diff
@@ -596,6 +598,15 @@ auto Settings::load() -> bool {
     loadButtonConfig();
     loadDeviceClasses();
 
+    paletteFilePath = PALETTE_FILE;
+    auto paletteFile = Util::getConfigFile(paletteFilePath);
```
Febbe, line 602:
```suggestion
    auto paletteFile = Util::getConfigFile(PALETTE_FILE);
```

src/control/settings/Settings.h
```diff
@@ -887,4 +888,8 @@ class Settings {
      * e.g. "en_US"
      */
     std::string preferredLocale;
+
+public:
+    Palette* palette;
+    fs::path paletteFilePath;
```
Febbe, line 894:
```suggestion
```

src/control/settings/Settings.h
```diff
@@ -887,4 +888,8 @@ class Settings {
      * e.g. "en_US"
      */
     std::string preferredLocale;
+
+public:
+    Palette* palette;
```
Febbe, line 893: **Question:** Who is the owner of the Palette, Control or Settings or both? Or are both Palette* distinct?

src/gui/dialog/toolbarCustomize/ToolbarCustomizeDialog.cpp
```diff
@@ -314,20 +313,22 @@ void ToolbarCustomizeDialog::rebuildColorIcons() {
...
-            continue;
-        }
+    for (int i{}; i < palette->size(); i++) {
+        // Here we need the pointer to ensure that the memory still contains the relevant data
+        // when the toolitemColorDragBegin callback is raised later on.
+        // If we would use NamedColor namedColor it would automatically fall out of scope at the end of this function
+        // and then toolitemColorDragBegin would point to memory that has already been deallocated.
+        NamedColor* namedColor = palette->getColorPointerAt(i);
+        Color c = namedColor->getColor();
+        GtkWidget* icon = ColorSelectImage::newColorIcon(c, 16, true);
```
Febbe, line 325: > // If we would use NamedColor namedColor it would automatically fall out of scope at the end of this function

Yes but you can use a reference instead `NamedColor const& namedColor` using `NamedColor namedColor` will also copy everything into the memory of `namedColor` on assignment.

```suggestion
    for (auto const& namedColor : palette->colors()) {
        Color c = namedColor.getColor();
        GtkWidget* icon = ColorSelectImage::newColorIcon(c, 16, true);
```

src/gui/toolbarMenubar/ColorToolItem.h
```diff
@@ -23,7 +23,7 @@ class ColorSelectImage;
 
 class ColorToolItem: public AbstractToolItem {
 public:
-    ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent, Color color,
+    ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent, const NamedColor& namedColor,
```
Febbe, line 26: You could move the value in the constructor.

```suggestion
    ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent, NamedColor namedColor,
```

src/gui/toolbarMenubar/ColorToolItem.cpp
```diff
@@ -6,23 +6,19 @@
...
 
-ColorToolItem::ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent, Color color,
-                             bool selektor):
+ColorToolItem::ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent,
+                             const NamedColor& namedColor, bool selektor):
         AbstractToolItem("", handler, selektor ? ACTION_SELECT_COLOR_CUSTOM : ACTION_SELECT_COLOR),
-        color(color),
         toolHandler(toolHandler),
-        parent(parent) {
+        namedColor{namedColor} {
```
Febbe, line 20:
```suggestion
        namedColor{std::move(namedColor)} {
```

src/gui/toolbarMenubar/ColorToolItem.cpp
```diff
@@ -6,23 +6,19 @@
...
 #include "StringUtils.h"
 #include "Util.h"
 #include "i18n.h"
 
 bool ColorToolItem::inUpdate = false;
 
-ColorToolItem::ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent, Color color,
-                             bool selektor):
+ColorToolItem::ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, GtkWindow* parent,
+                             const NamedColor& namedColor, bool selektor):
```
Febbe, line 17:
```suggestion
                             NamedColor namedColor, bool selektor):
```

src/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -88,6 +90,9 @@ void ToolMenuHandler::unloadToolbar(GtkWidget* toolbar) {
 void ToolMenuHandler::load(ToolbarData* d, GtkWidget* toolbar, const char* toolbarName, bool horizontal) {
     int count = 0;
 
+    Palette* palette = this->control->getSettings()->palette;
+    int colorIndex{};
```
Febbe, line 94: use `size_t` for indices

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+    return false;
+}
+
+auto Palette::parseColorLine(const std::string& line) -> bool {
+    NamedColor color;
+    std::istringstream iss{line};
+    if (iss >> color) {
+        // call to private variable paletteIndex allowed because of friend class
+        color.paletteIndex = namedColors.size();
+        namedColors.push_back(color);
```
Febbe, line 55:
```suggestion
        namedColors.emplace_back(std::move(color));
```

src/gui/toolbarMenubar/model/ColorPalette.cpp
```diff
@@ -0,0 +1,154 @@
...
+        g_error(".gpt file needs to start with \"GIMP Palette\" in the "
+                "first line");
+        return false;
+    }
+    return true;
+}
+
+auto Palette::parseHeaderLine(const std::string& line) -> bool {
+    Header headerEntry;
+    NamedColor color;
```
Febbe, line 40:
```suggestion
```

src/gui/toolbarMenubar/model/ToolbarItem.cpp
```diff
@@ -27,6 +27,8 @@ ToolbarItem::~ToolbarItem() = default;
 
 auto ToolbarItem::getName() -> string { return this->name; }
 
+void ToolbarItem::setName(const string& name) { this->name = name; }
```
Febbe, line 30:
```suggestion
void ToolbarItem::setName(string name) { this->name = std::move(name); }
```

src/util/StringUtils.cpp
```diff
@@ -78,3 +78,11 @@ auto StringUtils::iequals(const string& a, const string& b) -> bool {
...
     return result == 0;
 }
+
+auto StringUtils::isNumber(const string& input) -> bool {
+    for (char const& c: input) {
+        if (std::isdigit(c) == 0)
+            return false;
+    }
+    return true;
+}
```
Febbe, line 88: Check if it is better to use [`std::from_chars`](https://en.cppreference.com/w/cpp/utility/from_chars) here.
Especially if a conversion follows this function and when it's called in a loop. `std::from_chars` may the faster version.

src/util/NamedColor.cpp
idotobi: Makes sense. Will be done :heavy_check_mark: .
Didn't know the `final` specifier before thanks for letting me learn something new :pray:

src/util/NamedColor.h
idotobi: I'll remove them (see above).

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: I think here it makes sense to write the wrapper function. This way the cycling behaviour does not need to be implemented by the caller

```cpp
    if (i >= namedColors.size()) {
        i = i % namedColors.size();
```

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: You meant `const NamedColor& Palette::getColorAt(size_t i) const`, right?
That is definitely way prettier. Thanks :pray: !

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: With `const NamedColor& Palette::getColorAt(size_t i) const` this is not needed anymore. Will be removed :heavy_check_mark:

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: Only `const NamedColor& Palette::getColorAt(size_t i) const` will be kept.

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: Agreed. This is better :+1:

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: Will be done :heavy_check_mark: .

src/gui/toolbarMenubar/model/ColorPalette.h
idotobi: I see your point. I'd still tend to keep it in the Header file for now. I think at some point the Header should also be used, for example for showing information about the currently active palette.
I'm not sure if I'll implement this in this PR (probably not yet) but it should still come. Then the Header class would need to be public anyways, right?

src/util/Color.h
idotobi: :heavy_check_mark:

src/util/NamedColor.cpp
idotobi: Yep, that's a stupid mistake :sweat_smile:

src/util/NamedColor.cpp
idotobi: Yes agreed :+1:

src/util/NamedColor.cpp
idotobi: :heavy_check_mark:

src/control/settings/Settings.cpp
idotobi: Yes, the the topmost line in your snippet calls it.

Basically the behaviour is:
1. Try to read palette file in `.config/xournalpp` foler (or similar)
      a. if file does not exist create default palette file at specified path from xournalpp colorscheme (useful for soft migration)
2. Parse file

The only case that is not caught is when the palette file exists but has a non parseable format. For this case we still should define the expected behaviour.

I'd propose in case of exception or empty palette (no colors). Show warning and use default color scheme. What do you think?

src/control/Control.h
idotobi: Yes! Way better :+1:

src/control/settings/Settings.cpp
idotobi: Agree :heavy_check_mark:

src/control/settings/Settings.cpp
idotobi: I did this to give easy access to `paletteFilePath`  for every code having access to Settings. But as this is nowhere used right now I'm open to removing it. It would only be helpful for future functionality of showing palette info, or even selecting it dynamically. What do you think?

src/control/settings/Settings.h
idotobi: Oh, the one in control was only a leftover, from the initial implementation attempt. There should be only one palette and it's in settings.

src/gui/dialog/toolbarCustomize/ToolbarCustomizeDialog.cpp
idotobi: :sweat_smile: I actually did it this way because of another narrowing conversion warning by clang-tidy further down the function.

I changed it now to iterate over `size_t` and to to the casting then explicitly when the `i` variable is actually used for the placement of the icon.

```cpp
int ii{};
        if (i >= std::numeric_limits<int>::max())
            g_error("Int overflow because of two many colors defined in Palette");
        else
            ii = static_cast<int>(i);
```

src/gui/toolbarMenubar/ToolMenuHandler.cpp
idotobi: :heavy_check_mark:

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: :heavy_check_mark:

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: Copy paste error :see_no_evil: .

src/gui/toolbarMenubar/model/ToolbarItem.cpp
idotobi: For the caller both approaches would yield the same behaviour, right?

```cpp
ToolbarItem toolbarItem{};
std::string name{"test"};
toolbarItem.setName(name);
REQUIRE(name == "test");
```

* `ToolbarItem::setName(const string& name)` passes name by reference but then copy assigns `this->name = name;`
* `ToolbarItem::setName(string name)` passes name by value (hence I guess copying it) and then we move it into `this->name = std::move(name);`

It that's the case I would prefer the `const string& name` as it clearly states that the string name will not be changed in the process. But I guess there's something else I'm not aware of :sweat_smile:

src/util/StringUtils.cpp
idotobi:
```cpp
                    } else if (StringUtils::isNumber(arg)) {
                        paletteIndex = std::stoi(arg);
                        namedColor = palette->getColorAt(paletteIndex);
```

This is how it's used in the `ToolMenuHandler` when the palette is read. I think here it is OK to trade speed for readability (unless someone has a lot of colors in the toolbar).

With the `std::from_chars` method the control flow would be more complicated as it's not as strict as I would like it to be.

I assume the could would need to look like the following:

```cpp
int tmp;
auto [p, ec]  = std::from_chars(arg.begin(), arg.end(), tmp);
if(p == arg.end()) paletteIndex = tmp;
...
```

src/gui/toolbarMenubar/ColorToolItem.h
idotobi: I think I'm probably missing something, namely the same thing as in https://github.com/xournalpp/xournalpp/pull/2379#discussion_r548090298.

idotobi:
@Febbe : Thank you for the the detailed feedback :pray: . I tried to address all your comments either by directly implementing them or by answering with additional questions :innocent: .
I'd still want to polish a few things before updating the PR, so right now I wouldn't consider the PR ready for review yet. But I would let you know once I have everything ready. If you could review it then that would be great.

src/gui/toolbarMenubar/model/ColorPalette.cpp
Febbe: There is no difference between `T const&` and `const T&`. It's completely a matter of flavor. I prefer the first, because a preceding `const` only exists before "named types". Thus, the first version is in my opinion more consistent/consequent. E.g.:
`T const* const* const` `==` `const T* const* const`

src/control/settings/Settings.cpp
Febbe: > The only case that is not caught is when the palette file exists but has a non parseable format. For this case we still should define the expected behaviour.

Yes I would use the same behavior, override the file and use the default palette.

But it may be better to just load the default settings from constexpr variables. And then save the file.

```
// try {
color_palatte.load();
if (<onfail>) /* } catch(std::exeption const& ...) */ {
    color_palatte = generate_default_palette();
    color_palatte.save()
}

```

src/control/settings/Settings.cpp
Febbe: It's unlikely that it is ever needed. And if we will decide to store such strings/paths in the future, they should be stored at another position. 
Also, it is very cheap to recreate the path.

src/control/settings/Settings.h
Febbe: Ok then it should be a std::unique_ptr

src/gui/dialog/toolbarCustomize/ToolbarCustomizeDialog.cpp
Febbe: That's definitely the better way. Mostly I just write 
```
assert(i >= std::numeric_limits<int>::max());
function_call(static_cast<int>(i));
```
The else branch is not required after a g_error call. Just write the following or cast the `i` in place:

```
        if (i >= std::numeric_limits<int>::max())
            g_error("Int overflow because of two many colors defined in Palette");
        auto ii = static_cast<int>(i);
```

src/gui/toolbarMenubar/model/ToolbarItem.cpp
Febbe: > It that's the case I would prefer the `const string& name` as it clearly states that the string name will not be changed in the process. But I guess there's something else I'm not aware of sweat_smile

Both variants guarantee, that the original value is not changed in the process.
But "pass by value and move" has the advantage, that this allows the user to prevent any copies at all, since it moves the point of copy directly in front of the function. Also, it marks the function as a value copying function.

This allows the following:

```
    ToolbarItem toolbarItem{};
    std::string name{"test"};
    toolbarItem.setName(std::move(name));   //< No copy here! name is invalidated
    return toolbarItem;
} // delete empty / invalidated name 
```
But removing the `std::move`:

```
    ToolbarItem toolbarItem{};
    std::string name{"test"};
    toolbarItem.setName(name);   //< Implicit Copy here! name will still be unchanged
    return toolbarItem;
} // destroy name
```

src/util/StringUtils.cpp
Febbe: > I assume the could would need to look like the following:
> 
> ```c++
> int tmp;
> auto [p, ec]  = std::from_chars(arg.begin(), arg.end(), tmp);
> if(p == arg.end()) paletteIndex = tmp;
> ...
> ```

I am not exactly sure, but it looks good. I would probably check for the error, depending on the invariant at the end of the function call.
It also has the advantage, to catch overflow errors, and it works with the same locale ("C") as `std::atoi` / `std::stoi` If you want to use the latter you actually should check the errno flag. And this is not thread safe.

src/gui/toolbarMenubar/model/ColorPalette.cpp
idotobi: Thanks for explaining. I looked it up and it seems that there are some other benefits to this way of writing it ([stackoverflow](https://stackoverflow.com/questions/2640446/why-do-some-people-prefer-t-const-over-const-t)). I'll try it out.

src/control/settings/Settings.cpp
idotobi: I though about overriding at first as well. But I think that would annoy users quite a bit.

Imagine you tweaked your palette but did a simple formatting mistake. You start xournalpp. Upps, now your palette file is overwritten with the default :see_no_evil: 

I'd rather show a "warning" message with `g_message` and load the default palette but not overwrite the palette file.

src/control/settings/Settings.cpp
idotobi: Ok. You convinced me. Will be done :heavy_check_mark:

src/control/settings/Settings.h
idotobi: :heavy_check_mark:

src/control/settings/Settings.h
idotobi: :heavy_check_mark:

src/gui/toolbarMenubar/model/ToolbarItem.cpp
idotobi: Now I get it

1. Approach one: "pass by value"
     a. "... and copy": one copy, hence less efficient; `name` still intact be used after calling setName
     b. "... and move": no copy more,  hence efficient; `name` no longer useful as empty
2. Approach two: "pass by reference"
     a. one copy, hence less efficient; `name` still intact be used after calling setName

So basically by using "pass by value" in the function signature you give the programmer the option whether to move or copy depending on the usecase, whereas with "pass by reference" you force the copy on them :bulb:

idotobi: Will be done :heavy_check_mark:

src/gui/toolbarMenubar/ColorToolItem.cpp
idotobi: :heavy_check_mark:

src/gui/toolbarMenubar/ColorToolItem.h
idotobi: Question resolved. Will be done :heavy_check_mark:

src/util/StringUtils.cpp
idotobi: I think I would rewrite it in this way then:

```cpp
// check for old color format of toolbar.ini
                    if (StringUtils::startsWith(arg, "0x")) {
                        namedColor = palette->getColorAt(colorIndex);
                        paletteIndex = namedColor.getIndex();

                        // set new COLOR Toolbar entry
                        std::ostringstream newColor("");
                        newColor << "COLOR(" << paletteIndex << ")";
                        dataItem->setName(newColor.str());
                        colorIndex++;
                    } else {
                        std::istringstream colorStream(arg);
                        colorStream >> paletteIndex;
                        if (!colorStream.eof() || colorStream.fail()) {
                            g_warning("Toolbar:COLOR(...) has to start with 0x, get color: %s", arg.c_str());
                            continue;
                        }
                        namedColor = palette->getColorAt(paletteIndex);
                    }
```

I would still have two questions for my understanding:
1. Why would I need to check errno? Wouldn't it simply through an exception if there were an `std::out_of_range` error?
2. Why would thread safety be of concern here?

idotobi:
@Febbe : After a busy start in the year I finally managed to incorporate all the review comments (and rebased on master). From my side there's nothing missing anymore, so it would be ready for a second round of review :nerd_face: 

If you need color palettes to try out, some nice ones can be found here: https://lospec.com/palette-list

To change the palette, you need to replace the `.config/xournalpp/palette.gpl`.
In case you want to get the original one back just delete the file.

idotobi:
This PR, should only be merged after https://github.com/xournalpp/xournalpp/pull/2860 as they will create merge conflicts and https://github.com/xournalpp/xournalpp/pull/2860 is more important.

rolandlo:
> This PR, should only be merged after #2860 as they will create merge conflicts and #2860 is more important.

#2860 has been merged. So can you please rebase and solve the conflicts?

jcguu95:
@idotobi Wow we almost got this PR done! Thank you for your great effort!

idotobi:
@jcguu95 Thanks. It didn't really fit in this release anymore. But I'll restart work on it once v.1.1.0 is out :innocent:

wedwo:
Where would this palette.gpl file be located on a Windows installation by default?  I can't seem to find it anywhere.  Thanks!

LittleHuba:
I'll merge this into the main development branch, if there are no objections within 24 hours.

src/control/settings/Settings.cpp
LittleHuba: A dialog would be even better. g_messages are quite hard to get to on Windows.

idotobi:
@LittleHuba : I wanted to double check, tomorrow morning if there's still anything that would need polish. Then (SA 25.09 ~2pm) I think it should be ready for merging. Just want to make sure that this doesn't cause problems for anyone following the main dev branch :sweat_smile: .

src/control/settings/Settings.cpp
idotobi: Good idea I'll try that out.

idotobi:
@wedwo : Sorry for the long waiting time I wasn't that active in the last 2 months.

Since I don't have a Windows machine anymore I couldn't verify myself, but according to https://xournalpp.github.io/guide/file-locations/ it should be under `C:\Users\<user>\AppData\Local\xournalpp`.

LittleHuba:
Could you place the function to create an error dialog somewhere more central in the application? We need that functionality quite often and this gives us the means to centralize this functionality.

Also errorDialog itself should create the error message in the log.

Febbe:
@littleHubba currently all our "custom" error dialog/Error H
Händlers can't be ported to gtk4. This is, because they removed blocking dialogs(dialog.run()). Therefore I would recommend, to write a custom dialog, without using the run function.

LittleHuba:
> @littleHubba currently all our "custom" error dialog/Error H
Händlers can't be ported to gtk4. This is, because they removed blocking dialogs(dialog.run()). Therefore I would recommend, to write a custom dialog, without using the run function.

Ahh good to know. Then we should create a new dialog class in a follow-up PR and move from such custom error handlers to this new class.

LittleHuba:
@idotobi Please do not change the behavior of the error handling then. You are welcome to create the PR for the error dialog class though. If @Febbe has not already worked on that?

idotobi:
@LittleHuba : Ok. Should I still move the function to a more central place? If yes where to you think would be a good place for it?

LittleHuba:
> @LittleHuba : Ok. Should I still move the function to a more central place? If yes where to you think would be a good place for it?

I don't think there is any benefit in that. We will have to walk the full codebase for such dialogs anyway to centralize them. So leave it as it is for now.

LittleHuba:
#2735 will have to be modified because of cadeab3

idotobi:
Ok, So I just did a last check:
 - Does it work with no `toolbar.ini` yet (fresh install scenario)? :heavy_check_mark: 
 - Does it work with `toolbar.ini` based on old `COLOR(0xffffff)` format, and does migration work (migration scenario)? :heavy_check_mark: 
 - Can I customize colors as expected (i.e. same color multiple times, self defined order)? :heavy_check_mark: 
 - Do I get the error Dialogue? :heavy_check_mark: 

Unless I missed anything it should be ready for merging :crossed_fingers:

LittleHuba:
Could you whip up a test suite for the parser? We need to add tests for such stuff to get some reliability.

Febbe:
Partially I worked on this, bit only to get the dialogs to work with gtk4.

LittleHuba:
> Partially I worked on this, bit only to get the dialogs to work with gtk4.

Would you mind separating this work into a PR we can merge in the near future? It would be sensible to get this done as soon as possible since there are other PRs that would benefit of this work.

idotobi:
> Could you whip up a test suite for the parser? We need to add tests for such stuff to get some reliability.

I'd love to! But I probably won't manage to get it done today. Haven't worked with CppUnit before, and I have not much experience with testing in cpp yet :sweat_smile: . But creating a test framework for the parser would be a nice learning exercise.

LittleHuba:
You can look at the existing tests and #3390 for some inspiration. There is no real hurry though. So take your time.

idotobi:
@LittleHuba : Please wait with merging I just found one bug while writing a test...

idotobi:
So I also checked the one thing now that I previously missed:
 - Does application work if both `palette.gpl` and `tollbar.ini` are not existing? :heavy_check_mark: 

I also started on the test suite, which resulted in the three last fixes (even though none of these bugs was user facing :sweat_smile: ).
From my side it's ready for merging now :nerd_face: . 
The test suite will come in a separate PR. Unless of course you think it's better to wait on the testsuite.

LittleHuba:
It is definitely better to wait on the testsuite.

idotobi:
@LittleHuba : Testsuite is included now :innocent:

LittleHuba:
> @LittleHuba : Testsuite is included now 😇

And failing on both MacOS and Windows 😉

idotobi:
> And failing on both MacOS and Windows :wink: 

:man_facepalming:  :sweat_smile: those lovely platform differences...

Technius:
Looks like some sort of linking/dependency issue on Mac:

<details>

```
test 3
    Start 3: ColorPalette

3: Test command: /Users/runner/work/1/s/build/test/test-palette
3: Test timeout computed to be: 10000000
3: dyld: Library not loaded: @rpath/libpoppler-glib.8.dylib
3:   Referenced from: /Users/runner/work/1/s/build/test/test-palette
3:   Reason: image not found
3/3 Test #3: ColorPalette .....................Subprocess aborted***Exception:   0.14 sec
```
</details>

idotobi:
@Technius : Yeah you're right. My best guess is that up until now the tests  didn't depend on `poppler` and hence it's not installed on the CI yet.

I'm currently trying to install it in the CI and get it going by trial and error :sweat_smile: . If someone knows how to do this, help would be be appreciated (the iteration cycles with CI config are quite tedious :zzz: )

idotobi:
:tada: a missing `install_name_tool` was missing.

```
              install_name_tool -add_rpath /Users/xournal-dev/gtk/inst/lib/. test/test-palette
```

LittleHuba:
Your code looks good so far. The only issue that I still see is a lot of missing comments within your code. Please describe what it does. For people not familiar with your work, this will help them to make their own modifications later on without breaking your stuff. It also helps the rest to understand what you want to achieve with your code and if it actually does what you expect it to do.

idotobi:
Sure. I'll go through it once more and turn `--verbose` mode on :wink:. I probably will get to it on the weekend.

DaJazzuwu:
Hi, I'm a extreme newbie in all of this. I want to customize the color palette, but I read this and I really don't now where to start (not even what to search), if someone could help me a bit with a youtube tutorial of what I should know before even trying I would be really thankful.

idotobi:
> Hi, I'm a extreme newbie in all of this. I want to customize the color palette, but I read this and I really don't now where to start (not even what to search), if someone could help me a bit with a youtube tutorial of what I should know before even trying I would be really thankful.

@DaJazzuwu : Until this PR is merged: The best thing for you to do based on the currently released xournalpp is to.
 1. Checkout where your config files are located on your OS (see https://xournalpp.github.io/guide/file-locations/)
 2. edit the `toolbar.ini` (if it is not existent yet, just go to `View` -> `Toolbars` -> `Customize` inside xournalpp and create a custom toolbar, this should create a `toolbar.ini` for you)
 3. Search for COLOR(0x`??????`) and change the hex values, where `??????` is a hex RGB value.
 4. Start xournalpp

This or you wait a bit until it's merged, then it should be ready for you to play around with. This should be not later than Christmas :santa:

Technius:
@idotobi Awesome work!

@LittleHuba I wrote some WIP docs, see https://github.com/xournalpp/xournalpp.github.io/pull/58

ja-he:
Hey, I wanted to try this out and encountered a crash, so I investigated a bit;
maybe you already know this about the WIP state, otherwise hopefully this helps :slightly_smiling_face:.

<details><summary>More datails about the crashes, which commits I experienced them on, etc.</summary>

> Linux system (Manjaro on 5.14 kernel),
> i3 window manager (also tried XFCE for most recent commit),
> libgtk version 3.24.30

For this I used the example file contents found in <https://github.com/xournalpp/xournalpp.github.io/pull/58/files> stemming from #1812:

```
GIMP Palette
Name: Xournal New Palette.gpl
#
255 225 107 #FFE16B
255 161  84 #FFA154
205 158 247 #CD9EF7
155 219  77 #9BDB4D
100 186 255 #64BAFF
128 128 128 #808080
 58 145   4 #3A9104
237  83  83 #ED5353
  0  46 153 #002E99
  0   0   0 #000000
255 255 255 #FFFFFF
```

Results of testing _some_ of the commits, by recency:

- 07c2e23c:
  - ✓ works great:
  - warns: 'There are more Coloritems in the Toolbar then your Palette defines.'
- 1f95622e:
  - ? fails to read the palette:
  - shows an error in a dialog window
    > There has been a problem parsing the colour palette [...]
    > line 3 is malformed. [...]
    > The application will start with the default color palette
  - starts normally when the error is dismissed with 'OK', default color palette
- 39002773:
  - ✗ not usable:
  - shows an error in a dialog window
    > There has been a problem parsing the colour palette [...]
    > line 3 is malformed. [...]
    > The application will start with the default color palette
  - **segfaults when the error dialog is dismissed with 'OK'**
- b184fd26:
  - ✗ not usable:
  - segfaults, no error dialog, just the crash handler warning
- 84750c62:
  - ✗ not usable:
  - segfaults, no error dialog, just the crash handler warning
- d06aa5b3:
  - ✗ not usable:
  - segfaults, no error dialog, just the crash handler warning
  - note: also doesn't work with no 'palette.gpl' present at all, in which case from the previous states it seems it should actually generate a default file, instead it crashes all the same; (might also be the case for the other commits that segfault)

</details>

Since I'm already commenting, might as well nitpick the error messages:
- color - colour inconsistency
- 'then' should be 'than'

Thanks for your work on this, i would  love this feature.

idotobi:
@ja-he : Thanks a lot for the detailed bug-report :pray: ! Also for the nitpicks :wink: 
I have to rebase on master anyways as quite some stuff changed in the meantime. Then I'll look into the fix. If I'm not way to overconfident you should have something working by the end of today :crossed_fingers:

ui/toolbar.ini
```diff
@@ -36,8 +39,13 @@ name = Portrait
 name[de] = Hochformat
 name[it] = Ritratto
 name[ru] = Портретная
+<<<<<<< HEAD
```
idotobi, line 42: Seems my rebasing skills are still lacking :man_facepalming: I'm pretty sure this was the culprit. 
@ja-he: I'll test it out, just to be sure, then I'll update the PR.

idotobi:
@ja-he : Ready for testing. Let me know if you see anything unexpected. I also added some documentation at https://github.com/xournalpp/xournalpp.github.io/pull/58 in case you're curious.

ja-he:
works great @idotobi

Febbe:
@idotobi can you squash your commits?
Unfortunately, this can't be done by GitHub.

Febbe:
I also can do this for you if you want. I'll merge it then, just before #2570.

idotobi:
@Febbe : Do you mean I should squash and force push, or directly squash and merge?

idotobi:
@Febbe : I squashed and force pushed. Should be ready for merging.

idotobi:
There's one thing that users should be made aware of when the update to this version:

> In case you ordered your colors in the toolbar in a custom way this order is lost in the migration. However, if you reorder them afterwards this order will be persisted, so it's a one time effort.

When we do the release we might want to put this in the release notes.

daholzfeind:
Why did you revert my changes from https://github.com/xournalpp/xournalpp/commit/2fb95d75d69e8aa92d25fe13f7e5ecae569539f6#diff-33b9de3205649dabd01cebf47f90185d2efde29266a2ecc7032da0e78824102d ? Is there any reason for it? If no, can i undo it again?

rolandlo:
> @idotobi or @Febbe Why did you reverted my changes from [2fb95d7#diff-33b9de3205649dabd01cebf47f90185d2efde29266a2ecc7032da0e78824102d](https://github.com/xournalpp/xournalpp/commit/2fb95d75d69e8aa92d25fe13f7e5ecae569539f6#diff-33b9de3205649dabd01cebf47f90185d2efde29266a2ecc7032da0e78824102d) ? Is there any reason for it? If no, can i undo it again?

Probably a rebasing mistake.

Febbe:
@daholzfeind thanks for the catch, this happened unintentionally then.

> If no, can i undo it again?

src/util/NamedColor.h
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @brief Construct a new NamedColor instance
+     * This automatically sets the isPaletteColor to negative as NamedColors from a color palette can only be
+     * constructed using the input operator.
+     * As it is no paletteColor the Name of the Color will be set to "Custom Color" and the paletteIndex to 0.
+     *
+     * @param color color to set for the custom NamedColor instance
+     */
+    NamedColor(const Color& color);
```
marioalexis84, line 56: On 32 bit system, I cannot compile because size_t is uint32_t, so NamedColor::NamedColor (const Color &) cannot overload NamedColor::NamedColor (const size_t &). I can bypass the problem by replacing the type of the paletteIndex member with uint64_t, but surely there must be a better way to fix it.

daholzfeind:
> Yes please, you know better how to correct it.
> Pls mention me or @idotobi after providing the PR.

Done, see: https://github.com/xournalpp/xournalpp/pull/3614

I was confused because I use the PDF export every day and with the newest devel release it was gone ;)

idotobi:
@daholzfeind : Thanks for reporting. And yes it was a rebase mistake :see_no_evil: I'll review your change right away. If it works as expected I'll go ahead and merge it.

src/util/NamedColor.h
idotobi: Thanks for reporting this :pray: .
I think the easiest is to use `uint16_t` instead of `size_t` for the `paletteIndex`. I guess it's rather unlikely that we'll ever have more than 65536 colors in a palette. I'm already on it and will raise a PR soon.

idotobi: @LittleHuba : Well I guess that's another reason I shouldn't use `size_t` that much anymore :sweat_smile:

idotobi: @marioalexis84 : I prepared a fix at #3616. Please check if that works for you.

LittleHuba: > @LittleHuba : Well I guess that's another reason I shouldn't use `size_t` that much anymore :sweat_smile: 

Yeah one of the common issues of C(++). If I remember correctly, the only thing keeping us from making official 32-bit releases is the use of size_t in various places.

Febbe: > Well I guess that's another reason I shouldn't use size_t that much anymore 😅

Don't be that harsh to yourself: Using `size_t` for indices, especially when accessing containers, is completely fine.
Also, the usage of size_t is only problematic, if it is mixed with, or assigned to, other types.

This is rather an example to use stronger types than to not use size_t.
E.g. use my Color debug struct, to enforce that really every rgba-color is an uint32_t.

SidharthaDR:
Hello, im late, but can i know instructions on how to download this or use this feature?

SidharthaDR:
> Hello, im late, but can i know instructions on how to download this or use this feature?

Done i got it.

**Instructions to customize color palatte:**
The file **"palette.gpl"** is already in folder where xournalpp puts its config files.
You can open it in notepad and edit colors as you like!

||Linux (.deb installer)| MacOS (.app-bundle)| MS Windows (.exe installer)|
|--------| --------| -------|-----|
|Config folder| /home/user/.config/xournalpp | /Users/user/.config/xournalpp | C:\Users\<user>\AppData\Local\xournalpp|

source for config file location: https://xournalpp.github.io/guide/file-locations/
Docs for reference: https://xournalpp.github.io/guide/config/toolbar-colors/#toolbar-palette-files

Thank you for this feature!

PR #2380, opened 2020-11-09
Minor touchups for C++17

createyourpersonalaccount:
I made the following changes:

- Changes some C headers to C++ headers,
- `NULL` changed to `nullptr`
- `sprintf` changed to `snprintf` (not in every occurrence, but some where I deemed it OK)
- Changed `%x` format specifiers of `sprintf` with `PRIuxN` where appropriate (where `uintN_t` was used).
- Removed a use of `g_strdup_printf` thus avoiding the dynamic allocation.

### DIFF
--- a/src/control/AudioController.cpp
+++ b/src/control/AudioController.cpp
@@ -1,5 +1,7 @@
 #include "AudioController.h"
 
+#include <cinttypes>
+
 #include "Util.h"
 #include "XojMsgBox.h"
 #include "i18n.h"
@@ -16,8 +18,8 @@ auto AudioController::startRecording() -> bool {
         time_t secs = time(nullptr);
         tm* t = localtime(&secs);
         // This prints the date and time in ISO format.
-        sprintf(buffer.data(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1, t->tm_mday,
-                t->tm_hour, t->tm_min, t->tm_sec);
+        snprintf(buffer.data(), buffer.size(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1,
+                 t->tm_mday, t->tm_hour, t->tm_min, t->tm_sec);
         string data(buffer.data());
         data += ".ogg";
 
--- a/src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
+++ b/src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
@@ -1,5 +1,6 @@
 #include "ToolbarColorNames.h"
 
+#include <cinttypes>
 #include <fstream>
 
 #include <glib/gstdio.h>
@@ -12,9 +13,7 @@ ToolbarColorNames::ToolbarColorNames() {
     initPredefinedColors();
 }
 
-ToolbarColorNames::~ToolbarColorNames() {
-    g_key_file_free(this->config);
-}
+ToolbarColorNames::~ToolbarColorNames() { g_key_file_free(this->config); }
 
 static ToolbarColorNames* instance = nullptr;
 
@@ -61,14 +60,14 @@ void ToolbarColorNames::addColor(Color color, const std::string& name, bool pred
         this->predefinedColorNames[color] = name;
     } else {
         char colorHex[16];
-        sprintf(colorHex, "%06x", uint32_t{color});
+        snprintf(colorHex, sizeof(colorHex), "%06" PRIx32, uint32_t{color});
         g_key_file_set_string(this->config, "custom", colorHex, name.c_str());
     }
 }
 
 auto ToolbarColorNames::getColorName(Color color) -> std::string {
     char colorHex[16];
-    sprintf(colorHex, "%06x", uint32_t{color});
+    snprintf(colorHex, sizeof(colorHex), "%06" PRIx32, uint32_t{color});
 
     std::string colorName;
     char* name = g_key_file_get_string(this->config, "custom", colorHex, nullptr);
--- a/src/util/logger/Logger.cpp
+++ b/src/util/logger/Logger.cpp
@@ -6,12 +6,11 @@
 
 #ifdef DEV_CALL_LOG
 
+#include <cstdio>
 #include <fstream>
 #include <iostream>
 #include <sstream>
 #include <string>
-
-#include <stdio.h>
 
 #ifdef _WIN32
 
@@ -41,7 +40,7 @@ inline std::string NowTime() {
     tm r;
     strftime(buffer, sizeof(buffer), "%X", localtime_r(&tv.tv_sec, &r));
     char result[100];
-    sprintf(result, "%s.%06ld", buffer, (long)tv.tv_usec);
+    snprintf(result, sizeof(result), "%s.%06ld", buffer, (long)tv.tv_usec);
     return result;
 }
 
--- a/src/util/serializing/HexObjectEncoding.cpp
+++ b/src/util/serializing/HexObjectEncoding.cpp
@@ -1,5 +1,6 @@
 #include "HexObjectEncoding.h"
 
+#include <cinttypes>
 #include <cstdio>
 
 HexObjectEncoding::HexObjectEncoding() = default;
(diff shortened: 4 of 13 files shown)


createyourpersonalaccount:
I can't figure out how to use `git` to fix the clang test mistake. Feel free to do what you feel like with this PR, I'm giving up on git...

LittleHuba:
I'll give you a hand:
See this log: https://dev.azure.com/xournalpp/xournalpp/_build/results?buildId=7759&view=logs&j=8af2c752-b8c3-5474-5b0a-7b1f58b7bf90&t=0928bceb-a691-5a8f-98f6-852338e7281a

There is a diff showing all the formatting issues in your PR. You can either go and apply the diff using a tool like patch or change the offending lines by hand.

LittleHuba:
Or you just use the `clang-format` tool :wink:

createyourpersonalaccount:
Why did this happen with `clang-format`?
https://github.com/xournalpp/xournalpp/pull/2380/files#diff-0d5ed1bcf0bb37c261e3e6439809a1b040c391ab0ee5fe764cab02b640de738eL15-R16
It looks like it shouldn't have happened, and ` ToolbarColorNames::~ToolbarColorNames()` should've stayed a 3-line function...

LittleHuba:
That is actually okay. Functions that only contain one single small line of code just get wrapped that way. No real use in keeping it a 3-line one.

Mostly this just helps us all use the same coding style. This reduces merge conflicts caused by differently formatted code which is a real issue with a project of this size.

src/control/AudioController.cpp
```diff
@@ -16,8 +18,8 @@ auto AudioController::startRecording() -> bool {
         time_t secs = time(nullptr);
         tm* t = localtime(&secs);
         // This prints the date and time in ISO format.
-        sprintf(buffer.data(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1, t->tm_mday,
-                t->tm_hour, t->tm_min, t->tm_sec);
+        snprintf(buffer.data(), buffer.size(), "%04d-%02d-%02d_%02d-%02d-%02d", t->tm_year + 1900, t->tm_mon + 1,
+                 t->tm_mday, t->tm_hour, t->tm_min, t->tm_sec);
```
Febbe, line 22: Try to use `std::get_time` and `std::put_time` in c++17 for time conversions.

src/gui/toolbarMenubar/model/ToolbarColorNames.cpp
```diff
@@ -61,14 +60,14 @@ void ToolbarColorNames::addColor(Color color, const std::string& name, bool pred
...
     } else {
         char colorHex[16];
-        sprintf(colorHex, "%06x", uint32_t{color});
+        snprintf(colorHex, sizeof(colorHex), "%06" PRIx32, uint32_t{color});
         g_key_file_set_string(this->config, "custom", colorHex, name.c_str());
     }
 }
 
 auto ToolbarColorNames::getColorName(Color color) -> std::string {
     char colorHex[16];
```
Febbe, line 69: If you are touching this, you could directly replace old plain c arrays with c++11 std::arrays.
Another idea would be to completely get rid of c-format `sprintf` functions with std::format, but this is still not available on most compilers. Or use `std::stringstream`. But usecases of `std::stringstream` will be replaced mostly by std::format either. So for now this change is good.

src/util/logger/Logger.cpp
```diff
@@ -41,7 +40,7 @@ inline std::string NowTime() {
     tm r;
     strftime(buffer, sizeof(buffer), "%X", localtime_r(&tv.tv_sec, &r));
     char result[100];
```
Febbe, line 42:
```suggestion
    std::array<char,100> result;
```

src/util/serializing/HexObjectEncoding.cpp
```diff
@@ -10,8 +11,8 @@ void HexObjectEncoding::addData(const void* data, int len) {
     char* buffer = static_cast<char*>(g_malloc(len * 2));
```
Febbe, line 11: Replace this by `std::vector<std::byte>`

src/util/serializing/HexObjectEncoding.cpp
```diff
@@ -10,8 +11,8 @@ void HexObjectEncoding::addData(const void* data, int len) {
...
 
     for (int i = 0; i < len; i++) {
-        int x = static_cast<uint8_t const*>(data)[i];
-        sprintf(&buffer[i * 2], "%02x", x);
+        uint8_t x = static_cast<uint8_t const*>(data)[i];
+        sprintf(&buffer[i * 2], "%02" PRIx8, x);
     }
 
     g_string_append_len(this->data, buffer, len * 2);
```
Febbe, line 18: replace data by `std::vector<std::byte>`

createyourpersonalaccount: That would change the signature of `HexObjectEncoding()`. Who uses `HexObjectEncoding()`? I tried to search for it in Xournal++'s source code, but it only references itself, and it's not used anywhere else in the source code other than the corresponding source and header files. Is it a library used in some system?

src/control/AudioController.cpp
createyourpersonalaccount: I do not have the knowledge to do this change unfortunately. I have no understanding of locale issues and what `AudioController` does. I am confident that the `snprintf` change is equivalent, but I would not be so confident using `get_time` and `put_time` myself.

src/util/logger/Logger.cpp
createyourpersonalaccount: I don't know if that is necessary. `char[100]` is also C++ and the array template has a lot more features than what is needed in `NowTime()`. I wanted to rewrite `NowTime()` entirely in C++ (avoiding POSIX `localtime_r`) but unfortunately C++17 `<chrono>` does not have a way to do this. It seems to be something that C++0a might bring to the table.

Febbe: Both compile to the same binary. The only difference is, that the c++ array enables more static analysis and also provides good debug routines in debug mode. Also some operations are range checked by default.

Febbe:
@createyourpersonalaccount do you want to do some suggestions / nitpicks I made? Or shall I merge it?

createyourpersonalaccount:
Sure, I'd like to hear them. I was waiting for your reply on a question I made on `HexObjectEncoding()`. As for `std::array`, again I think it would change the signature of the concerned function. I tried to have small local changes. I want to keep looking at Xournal++s code to get a hang of what's going on and trying to improve things here and there. This is just a minor "fix", but I hope to find some bugs later on, or at least make the source code nicer.

src/util/logger/Logger.cpp
Febbe: Yes cpp-20 brings a way better time location and calendar support. I would suggest to wait until that day we switch to cpp-20. But this would require to drop some old operating systems like macOS10.13, and Ubuntu 16.04 maybe also 18.04.

src/util/serializing/HexObjectEncoding.cpp
Febbe: Currently, no-one is using it. Not sure, if we should just remove that class then.
Since this project isn't a library it is not a problem to switch the signature of classes.
It's more matter of how many work do you have to do, to refactor the code. E.g. changing the signature by a change of private fields, this would only require changes in the corresponding source file. But changing a function requires changes in many places.

createyourpersonalaccount: There's no point in fixing the code for `HexObjectEncoding.cpp` if it's just going to be removed though. It might be older code that used it. Maybe it's actually used for something and it would be a mistake to remove it. If, as you say, it's not needed, I think it should be removed.

src/util/logger/Logger.cpp
createyourpersonalaccount: Okay, to resolve this, I will keep it as `char result[100];`. I'm not feeling like changing it right now because I'm a bit tired from something else.

Febbe:
Ok give me a note, if you want to defer the suggestions into another commit.

Issue #1772, opened 2020-03-14
Input Pen not detected on MacOS: No devices found


andrew67912:
**Affects versions :**
 - MacOS High Sierra, Ver. 10.13.6
 - Xournal++, 1.0.17, Feb 3 2020
 - Pen Wacom Intuous

**Describe the bug**
The pen device is not properly detected. In [Preferences > Input System > Input Devices] the following note shows up:
"No devices were found. This seems wrong - may file a bug report?"

Nevertheless, the pen is working but at only low resolution and without pressure sensitivity. 

**Input Logs**
(As we use the pre-built binary, we do not have a log report file. If required, we could try to rebuild and create a log.) 

A workaround/solution to this bug would be great as we use Xournal++ on multiple OS.

adundovi:
I can confirm this, although on a different platform.

- OS: Fedora 31, KDE
- Version: Xournal++ 1.0.16 (RPM from the official repository)
- Wacom Cintiq 16

No tablet features (sensitivity, button, eraser) available within the app, and the stylus acts like a regular mouse. In the settings. the same message as above is displayed. KDE apps recognise it normally, including Krita. GIMP also works with all features, but Inkscape doesn't... maybe a GTK issue after all?

fredowski:
I can confirm this 

  * MacOS High Sierra 10.13.6
  * Xournal++ 1.0.17
  * Wacom Cintiq 13 HD

Thanks for your efforts!

gioiadc:
Same here, 

- MacOS High Sierra, Ver. 10.15.3

- Xournal++, 1.0.17, Feb 3 2020

- Wacom Intuos

thanks!

lagrange2art:
Same with me,

- maxOS Catalina, Ver. 10.15.3
- Wacom Intuos S (Driver Ver. 6.3.38-3) 

The tablet attached via an USBc hub from Aukey. It is properly configured and performs well in OneNote and other programs. There is no version of xournal++ for which the Wacom is recognised as input-device. 
- Xournal++ 1.0.17: no pressure sensitivity, pen buttons cannot be customised
- Xournal++ 1.0.8:  there is pressure sensitivity (although tablet is not recognised) however with     low resolution and "lagging" behaviour.

airbete:
Same here:
- OS: Fedora 30, KDE
- Version: Xournal++ 1.1.0+ dev (22a0dd89)
- Wacom Graphire4

[...]

PR #2390, opened 2020-11-11
MacOS: Update the build environment to gtk 3.24.23 - switch to macports

fredowski:
The MacOS build does not recognize the Wacom pen due to a problem
which is fixed in gtk 3.24.23. I did not manage to create the
build environment with jhbuild so I switched to macports. The Readme
explains how to create the build environment. The gtk-mac-integration
and the gtk-mac-bundler is still used, i.e. this build produces an
application bundle which runs standalone.

See: https://github.com/xournalpp/xournalpp/issues/1772

### DIFF
--- a/mac-setup/Readme.md
+++ b/mac-setup/Readme.md
@@ -1,0 +1,17 @@
+# How to build for MacOS
+
+The MacOS build will build most prerequisites via macports. Some updated versions are installed via the manual configure/make/make install. The build environment is located in $HOME/gtk/inst for historical reasons. You need to have Apple XCode installed.
+
+```
+cd mac-setup
+./installmacports.sh
+./cleanmacports.sh
+./build-gettext.sh
+./build-mac-integration.sh
+```
+
+Then your build environment is ready and you can create the app bundle
+
+```
+./complete-build.sh $HOME/gtk
+```
--- a/mac-setup/build-app.sh
+++ b/mac-setup/build-app.sh
@@ -72,6 +72,9 @@ for locale in */ ; do
   fi
 done
 
+## Fix the calling script in the bundle
+sed -i '' '/export DYLD_LIB.*/d' ./Xournal++.app/Contents/MacOS/xournalpp
+
 echo "Create zip"
 zip -r Xournal++.zip Xournal++.app
 
--- a/mac-setup/build-gettext.sh
+++ b/mac-setup/build-gettext.sh
@@ -1,0 +1,23 @@
+export PATH="/usr/bin:/bin"
+
+gettextversion=0.21
+rm -rf ./gettext
+curl -L https://ftp.gnu.org/pub/gnu/gettext/gettext-$gettextversion.tar.xz -o gettext.tar.xy
+tar -xJf ./gettext.tar.xy
+rm ./gettext.tar.xy
+mv gettext-$gettextversion gettext
+cd gettext
+./configure --prefix=$HOME/gtk/inst \
+  --disable-csharp \
+  --disable-java \
+  --disable-native-java \
+  --disable-openmp \
+  --without-emacs \
+  --with-included-gettext \
+  --with-included-glib \
+  --with-included-libcroco \
+  --with-included-libunistring \
+  --with-included-libxml
+
+make -j8
+make install
--- a/mac-setup/cleanmacports.sh
+++ b/mac-setup/cleanmacports.sh
@@ -1,0 +1,21 @@
+#!/bin/bash -xve
+
+# Cleanup the macports build environment
+
+
+# Copyright (C) 2020 Friedrich Beckmann
+# Released under GNU General Public License, either version 3
+# or any later option
+
+topdir=$HOME/gtk
+
+# This is the installation directory which will be used as macports prefix
+# and as pspp configure prefix.
+bundleinstall=$topdir/inst
+
+export PATH=$bundleinstall/bin:$bundleinstall/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
+
+for i in  {1..10}; do port -N uninstall leaves || true ; done
+
+rm -rf $bundleinstall/var/macports
+rm -rf $topdir/macports-ports
--- a/mac-setup/installmacports.sh
+++ b/mac-setup/installmacports.sh
@@ -1,0 +1,102 @@
+#!/bin/bash -xve
+
+# Install macports and all dependencies for xournalpp in $HOME/gtk/inst
+# macports is all recompiled because it uses the quartz backend instead of x11
+
+portindexversion=8e06b840b02731c3ef712f4aa6c4ff138a64a604
+
+# Copyright (C) 2020 Friedrich Beckmann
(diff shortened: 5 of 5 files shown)


LittleHuba:
Did you test it on a machine/user account without macports installed?

fredowski:
I build with macports but the application bundle also works if I rename the macports install directory, i.e. the libraries are really taken from the bundle. The "installmacports.sh" script produces a completely independent new macports install in $HOME/gtk/inst.

fredowski:
I added a pull request for the binary build environment. I build this on MacOS 10.13.6.

https://github.com/xournalpp/xournalpp-pipeline-dependencies/pull/1

LittleHuba:
That sounds good enough. Symbolic links should be broken by renaming so then you should actually have a "clean" environment where apps either work standalone or break completely. Could you do some testing just to make sure? Open a file write stuff and save it? Audio recording as well if possible. Then we know for sure.

But thanks for the effort so far!

fredowski:
Opening pdf, annotating, file open and save works. Pen sensitivity works. Eraser detection by turning the Wacom Pen also works. The audio recording does not work - but it does not work with the released bundle 1.0.19 either.

fredowski:
I have not tested the cppunit regression - only building the bundle with "complete-build.sh". But I added the cppunit to the dependencies.

LittleHuba:
Ahh yeah audio recording had some bug previously. So not your bad. 

Could you please check out this config of the pipeline? I assume it has to be changed as well for the new build env to work. This should be the last change before we can merge this in.
https://github.com/xournalpp/xournalpp/blob/master/azure-pipelines/steps/build_mac.yml

fredowski:
Unfortunately I have no idea how I can check the azure pipeline. Looking at the yaml file I do not see a reason why it should not work. As far as I understand it my changes are only relevant for creating the build environment but the pipeline uses the pre-build environment in the xournalpp/xournalpp-pipeline-dependencies repository.

fredowski:
Ohh. The libraries in the installation directory $HOME/gtk/inst are linked to my user name. So before the gtk-mac-bundler replaces these hard links the path has to be the same, i.e. the user has to be "fritz". 

Fredo:lib fritz$ otool -L libpsl.dylib 
libpsl.dylib:
	/Users/fritz/gtk/inst/lib/libpsl.5.dylib (compatibility version 9.0.0, current version 9.3.0)
	/Users/fritz/gtk/inst/lib/libiconv.2.dylib (compatibility version 9.0.0, current version 9.1.0)
	/Users/fritz/gtk/inst/lib/libintl.8.dylib (compatibility version 10.0.0, current version 10.5.0)
	/Users/fritz/gtk/inst/lib/libunistring.2.dylib (compatibility version 4.0.0, current version 4.0.0)
	/Users/fritz/gtk/inst/lib/libidn2.0.dylib (compatibility version 4.0.0, current version 4.7.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1455.255.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
Fredo:lib fritz$

LittleHuba:
> Ohh. The libraries in the installation directory $HOME/gtk/inst are linked to my user name. So before the gtk-mac-bundler replaces these hard links the path has to be the same, i.e. the user has to be "fritz".
> 
> Fredo:lib fritz$ otool -L libpsl.dylib
> libpsl.dylib:
> /Users/fritz/gtk/inst/lib/libpsl.5.dylib (compatibility version 9.0.0, current version 9.3.0)
> /Users/fritz/gtk/inst/lib/libiconv.2.dylib (compatibility version 9.0.0, current version 9.1.0)
> /Users/fritz/gtk/inst/lib/libintl.8.dylib (compatibility version 10.0.0, current version 10.5.0)
> /Users/fritz/gtk/inst/lib/libunistring.2.dylib (compatibility version 4.0.0, current version 4.0.0)
> /Users/fritz/gtk/inst/lib/libidn2.0.dylib (compatibility version 4.0.0, current version 4.7.0)
> /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1455.255.0)
> /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
> Fredo:lib fritz$

That is no big issue but not perfect either. The current package has the same problem but we found a way around. It is not the best thing that your name is in there though. People tend to interpret stuff in things not matching with their expectations. Could you recreate the build environment with a path independent of the username?

LittleHuba:
And could you also provide a step by step instruction for recreating the setup? So we can update the GTK version in the future. Just add it as a file next to the blob.

LittleHuba:
Without step by step guide this can not be merged!

fredowski:
Ah  - I forgot about this...

Technius:
We should make the binary blob build automated and part of CI.

mschinca:
I managed to compile and build and compile using the method described in this PR.
Here's my milage:
- app starts fine
- attempting to open a file located in Desktop, Documents etc. triggers an error: "Access Denied". This seems to be due to [TCC](https://apple.stackexchange.com/questions/395631/allow-all-applications-to-access-desktop-documents-and-downloads-in-catalina)
- granting full disk access to ```/bin/sh``` as described in the link above seems to work around the issue
- saving a note works
- exporting a brand new note to pdf works, but
- opening any existing pdf file and exporting as pdf causes the app to crash with error:
```
libc++abi.dylib: terminating with uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in equivalent: Operation not supported [/Users/me/Documents/test_export_from_original_pdf.pdf] [/Users/me/Documents/test_export_from_new.pdf]

** (com.github.xournalpp.xournalpp:39500): WARNING **: 12:53:38.390: [Crash Handler] Crashed with signal 6

** (com.github.xournalpp.xournalpp:39500): WARNING **: 12:53:38.390: [Crash Handler] Wrote crash log to: /Users/me/.cache/com.github.xournalpp.xournalpp/errorlogs/errorlog.20210330-125338.log
```
log attached.

Tested on macOS Big Sur 11.2 and macOS Catalina 10.15.7
[errorlog.20210330-125338.log](https://github.com/xournalpp/xournalpp/files/6228815/errorlog.20210330-125338.log)

Febbe:
Looks like macports also includes X11, therefore, gtk will build/ link for X11 instead of quartz. This might fix some performance issues with the current gtk3+ quartz edition. But I don't know if we want this.

Technius:
> Looks like macports also includes X11, therefore, gtk will build/ link for X11 instead of quartz. This might fix some performance issues with the current gtk3+ quartz edition. But I don't know if we want this.

Based on some testing with my M1 Mini, using the x11 variant doesn't work out of the box (I think it needs `xorg-server` to be installed too, but I haven't tried it yet).

Febbe:
We don't want to use X11 on MacOS I think 🤔.

Technius:
Superceded by #4371

Issue #2354, opened 2020-10-30
Todo Metaissue


Febbe:
This issue is a summary of several To-dos found by developers.

 - [ ] Replace path strings with `std::filesystem::path` / `fs::path`
     - [ ] `AudioController`: replace `std::string` with `fs::path`
     - [ ] ...
 - [ ] Replace `fs::path::string` calls with `fs::path::u8string`, if the result is used to display paths.
 - [ ] Create pipeline for msvc or clang for windows based on msvc, because mingw64 is very buggy and does not support all language features of c++11 to c++20 
 - [ ] Release the windows versions build by msvc or clang
 - [ ] Remove `(xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-Os -s")` and fix the windows pipeline to use the RelWithSize CMAKE flag

Issue #1155, opened 2019-05-15
Loading custom plugins


enricoar:
Maybe it's already possible, but I didn't find out why. Is it possible to make own  plugins and put them somewhere so they are recognized by xournalpp. At the moment it seems to work only if I put them in the plugin directory of the cloned source. It would be useful if one could put it, e.g. in the  .xournalpp directory.
Thanks

Technius:
This can now be easily accomplished thanks to #1384.

Currently, the plan is to load plugins from *all* of the following paths:
* `$XDG_DATA_HOME/xournalpp/plugins`. This will cover e.g. `~/.local/share/xournalpp/plugins`.
* `$DIR/xournalpp/plugins` where `$DIR` is `a folder from `$XDG_DATA_DIRS`. This will cover e.g. `/usr/share/xournalpp/plugins`.
* `$CONFIG_FOLDER/plugins` where `$CONFIG_FOLDER` is the xournalpp configuration folder, in case the configuration is synced across computers.

Febbe:
I would also add `./plugins` and `../plugins` and `../../plugins`

Febbe:
What shall we do, if there are duplicates?
I propose to add the path to the PluginManager, so both plugins are closeable, but they must be mutually exclusive I think.

Technius:
> I would also add `./plugins` and `../plugins` and `../../plugins`

Do you mean relative to the `xournalpp` binary? If you instead mean relative to the current directory, I would see no reason for why this would be desirable.

> What shall we do, if there are duplicates?

The most straightforward way I can think of is to load only the plugin that comes first in a priority list on the search paths (e.g. load the plugins from the user directories first, then ignore plugins with names similar to the ones already loaded in the following directories).

Febbe:
> Do you mean relative to the `xournalpp` binary? If you instead mean relative to the current directory, I would see no reason for why this would be desirable.

Yes relative to the source, this would make xournalpp relocatable.

> The most straightforward way I can think of is to load only the plugin that comes first in a priority list on the search paths (e.g. load the plugins from the user directories first, then ignore plugins with names similar to the ones already loaded in the following directories).

Interestingly it seems to work very well to load plugins with the same name.
But in the case we only show and load it once, which search order shall we use?
[...]

PR #2391, opened 2020-11-11
Refactored Plugin related code

Febbe:
- use std::unique_ptr
 - removed XournalType.h
 - fixed many clang-tidy warnings.
 - consistently use fs::path
 - mark functions override instead of virtual

### DIFF
--- a/src/plugin/Plugin.cpp
+++ b/src/plugin/Plugin.cpp
@@ -1,12 +1,10 @@
 #include "Plugin.h"
+#ifdef ENABLE_PLUGINS
 
 #include <utility>
 
-#include <config.h>
-
+#include "config.h"
 #include "i18n.h"
-
-#ifdef ENABLE_PLUGINS
 
 extern "C" {
 #include <lauxlib.h>
@@ -16,47 +14,22 @@ extern "C" {
 
 #include "luapi_application.h"
 
-#define LOAD_FROM_INI(target, group, key)                                 \
-    {                                                                     \
-        char* value = g_key_file_get_string(config, group, key, nullptr); \
-        if (value != nullptr) {                                           \
-            target = value;                                               \
-            g_free(value);                                                \
-        }                                                                 \
-    }
-
 /*
  ** these libs are loaded by lua.c and are readily available to any Lua
  ** program
  */
-static const luaL_Reg loadedlibs[] = {{"app", luaopen_app}, {nullptr, nullptr}};
+constexpr std::array loadedlibs{luaL_Reg{"app", luaopen_app}};
 
 Plugin::Plugin(Control* control, std::string name, fs::path path):
         control(control), name(std::move(name)), path(std::move(path)) {
     loadIni();
 }
 
-Plugin::~Plugin() {
-    if (lua) {
-        // Clean up, free the Lua state var
-        lua_close(lua);
-        lua = nullptr;
-    }
-
-    for (MenuEntry* m: menuEntries) {
-        delete m;
-    }
-    menuEntries.clear();
-}
-
-/**
- * Get Plugin from lua engine
- */
-Plugin* Plugin::getPluginFromLua(lua_State* lua) {
+auto Plugin::getPluginFromLua(lua_State* lua) -> Plugin* {
     lua_getfield(lua, LUA_REGISTRYINDEX, "Xournalpp_Plugin");
 
     if (lua_islightuserdata(lua, -1)) {
-        Plugin* data = (Plugin*)lua_touserdata(lua, -1);
+        auto* data = static_cast<Plugin*>(lua_touserdata(lua, -1));
         lua_pop(lua, 1);
         return data;
     }
@@ -64,9 +37,6 @@ Plugin* Plugin::getPluginFromLua(lua_State* lua) {
     return nullptr;
 }
 
-/**
- * Register toolbar item and all other UI stuff
- */
 void Plugin::registerToolbar() {
     if (!this->valid || !this->enabled) {
         return;
@@ -74,8 +44,8 @@ void Plugin::registerToolbar() {
 
     inInitUi = true;
 
-    lua_getglobal(lua, "initUi");
-    if (lua_isfunction(lua, -1) == 1) {
+    lua_getglobal(lua.get(), "initUi");
+    if (lua_isfunction(lua.get(), -1) == 1) {
         if (callFunction("initUi")) {
             g_message("Plugin «%s» UI initialized", name.c_str());
         } else {
@@ -88,10 +58,6 @@ void Plugin::registerToolbar() {
     inInitUi = false;
 }
 
-
-/**
- * Register all menu entries to the menu
- */
 void Plugin::registerMenu(GtkWindow* mainWindow, GtkWidget* menu) {
     if (menuEntries.empty() || !this->enabled) {
         // No entries - nothing to do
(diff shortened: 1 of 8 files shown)


rolandlo:
I was just looking over the code. Could you tell me, what you changed with your last commit/force-push? 
One remark: I needed to add `#include <memory>` to `PluginDialogEntry.h` to make it compile with gcc 9.2.1

Febbe:
> I was just looking over the code. Could you tell me, what you changed with your last commit/force-push?
> One remark: I needed to add `#include <memory>` to `PluginDialogEntry.h` to make it compile with gcc 9.2.1

I just added the #include <memory> in 2 files and fixed the clang-format. I normally use clang-12 to compile my cpp code. And clang is very friendly when it comes to missing header files. So it always compiles on my computer :D.

Febbe:
First of all I need feedback if all plugins still load and work as expected. I also need feedback if plugins in special non-ASCII folder names still work on Windows. Unfortunately I can't test this very often because my Windows PC is rarely accessible.

rolandlo:
> First of all I need feedback if all plugins still load and work as expected. I also need feedback if plugins in special non-ASCII folder names still work on Windows. Unfortunately I can't test this very often because my Windows PC is rarely accessible.

I have run your PR (compiled from source) from the terminal via `build/src/xournalpp`. All plugins in the `plugins` folder in the same `xournalpp` directory load as expected. To see filenames displayed lexically normal is certainly a plus. Otherwise all the plugins seem to work as before. 
One thing that bugs me is that accelerators like `<Release>z` don't seem to work. Firstly it is displayed like `Z` (as if there was no `<Release>` before the `z` key in the menu. Secondly pressing and releasing the `z` key does not execute the plugin. I don't know if I misunderstand something, but in the [GTK documentation](https://developer.gnome.org/gtk3/stable/gtk3-Keyboard-Accelerators.html#gtk-accelerator-parse) it is written 

> The format looks like “\<Control\>a” or “\<Shift\>\<Alt\>F1” or “\<Release\>z” (the last one is for key release).

So I would expect it to work.

Another question: Could we add a user plugin folder? Currently plugins are only loaded from one specific folder (which is a system folder in case Xournal++ has been installed on the system, and which cannot be modified without administrator rights).

I will now check how your PR works on Windows.

rolandlo:
Under windows there is something wrong with the terminal messages: 
grafik

rolandlo:
Special folder names do not seem to work on Windows. I renamed the ColorCycle plugin to `àÄ@fix~d$$äö!è`, after which it was still correctly displayed in the Plugin Manager, but could not be loaded upon activating it and restarting Xournal++, see
grafik

Febbe:
@rolandlo the terminal in windows don't support utf8 by default. Sometimes it isn't even ASCII/ANSI. To fix that all << and >> musst be replaced by "

Febbe:
> One thing that bugs me is that accelerators like `<Release>z` don't seem to work. Firstly it is displayed like `Z` (as if there was no `<Release>` before the `z` key in the menu. Secondly pressing and releasing the `z` key does not execute the plugin. I don't know if I misunderstand something, but in the [GTK documentation](https://developer.gnome.org/gtk3/stable/gtk3-Keyboard-Accelerators.html#gtk-accelerator-parse) it is written

Has the issue been there before?  I only changed the path representation, the rest is code cleanup. I also don't quite understand what the problem is. What is the difference between accelerators like `<Release>z' and those that work.

rolandlo:
> Has the issue been there before? I only changed the path representation, the rest is code cleanup. I also don't quite understand what the problem is. What is the difference between accelerators like `z' and those that work.

Yes, this issue has been there before. The idea of using accelerators like `<Release>z` is for temporary activation of some tool. For instance you want to switch to the eraser tool when pressing `z` and switch back to the previously used tool when releasing this key. Accelerators work, but the key release accelerators (that should act upon releasing the key) don't.

Febbe:
> Another question: Could we add a user plugin folder? Currently plugins are only loaded from one specific folder (which is a system folder in case Xournal++ has been installed on the system, and which cannot be modified without administrator rights).

I think that's a great idea. There is already a user config folder in `~/.config/xournalpp` there should be another folder "plugins" in there, I can do this.

Febbe:
> Yes, this issue has been there before. The idea of using accelerators like `<Release>z` is for temporary activation of some tool. For instance you want to switch to the eraser tool when pressing `z` and switch back to the previously used tool when releasing this key. Accelerators work, but the key release accelerators (that should act upon releasing the key) don't.

I didn't wrote the plugin code, so I don't know where the accelerators are handled. I can imagine, that the accelerators won't be handled, if they are already handled. This must be investigated. You can add this to #2354 by editing my post.

src/plugin/Plugin.cpp
```diff
@@ -282,33 +203,33 @@ void Plugin::loadScript() {
...
     addPluginToLuaPath();
 
     // Run the loaded Lua script
-    if (lua_pcall(lua, 0, 0, 0) != LUA_OK) {
-        const char* errMsg = lua_tostring(lua, -1);
-        map<int, string> button;
-        button.insert(std::pair<int, string>(0, _("OK")));
+    if (lua_pcall(lua.get(), 0, 0, 0) != LUA_OK) {
+        const char* errMsg = lua_tostring(lua.get(), -1);
+        map<int, std::string> button;
```
Technius, line 231: For consistency, this should also be `std::map`.

src/plugin/Plugin.cpp
```diff
@@ -317,17 +238,14 @@ void Plugin::loadScript() {
...
+    lua_getglobal(lua.get(), fnc.c_str());
 
     // Run the function
-    if (lua_pcall(lua, 0, 0, 0)) {
-        const char* errMsg = lua_tostring(lua, -1);
-        map<int, string> button;
-        button.insert(std::pair<int, string>(0, _("OK")));
+    if (lua_pcall(lua.get(), 0, 0, 0)) {
+        const char* errMsg = lua_tostring(lua.get(), -1);
+        map<int, std::string> button;
```
Technius, line 247: `std::map` here too.

src/plugin/Plugin.cpp
```diff
@@ -16,66 +14,38 @@ extern "C" {
...
-/**
- * Get Plugin from lua engine
- */
-Plugin* Plugin::getPluginFromLua(lua_State* lua) {
+auto Plugin::getPluginFromLua(lua_State* lua) -> Plugin* {
     lua_getfield(lua, LUA_REGISTRYINDEX, "Xournalpp_Plugin");
 
     if (lua_islightuserdata(lua, -1)) {
-        Plugin* data = (Plugin*)lua_touserdata(lua, -1);
+        auto* data = (Plugin*)lua_touserdata(lua, -1);
```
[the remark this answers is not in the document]
Technius: Or `static_cast`, whichever you think is more readable here (I have no opinion, they seem equally clear to me).

Technius:
> > Another question: Could we add a user plugin folder? Currently plugins are only loaded from one specific folder (which is a system folder in case Xournal++ has been installed on the system, and which cannot be modified without administrator rights).
> 
> I think that's a great idea. There is already a user config folder in `~/.config/xournalpp` there should be another folder "plugins" in there, I can do this.

See #1155 for the relevant issue.

src/plugin/Plugin.cpp
Febbe: done, also added a todo in our metaissue

Febbe:
> Special folder names do not seem to work on Windows. I renamed the ColorCycle plugin to `àÄ@fix~d$$äö!è`, after which it was still correctly displayed in the Plugin Manager, but could not be loaded upon activating it and restarting Xournal++, see
> grafik
> 
> **Edit** even with a simple folder name like `René` or `Ähnlichkeit` the plugin cannot be loaded under Windows.

It seems like, this is a dependency issue caused by mingw-64. If the path exist, fopen should be able to open path::string::c_str

PR #2393, opened 2020-11-12
Fix strokes only appearing after lifting pen

### LINKED ISSUE #2242: Strokes sometimes appear only after the pen is lifted
**Affects versions :**
 - OS: Linux Debian 10 with gnome desktop
 - Version of Xournal++ 1.0.18
 - Hardware: Thinkpad Yoga X1 gen1
**Describe the bug**
In xournal++ a line sometimes gets only drawn after the pen is lifted and the stroke is finished, this does not happen in Xournal (see Screenshots)

**To Reproduce**
1. Open Xournal++
2. Draw a stroke
3. there appers no stroke on the screen but sometimes a pixel gets dark where the stroke will be
4. The Pen gets lifted
5. The stroke appears after a slight delay

**Input Logs**

[debug_gdk_input.log](https://github.com/xournalpp/xournalpp/files/5218260/debug_gdk_input.log)
[debug_input.log](https://github.com/xournalpp/xournalpp/files/5218266/debug_input.log)

**Expected behavior**
The stroke should be visible as soon as it gets drawn.

**Screenshots of Problem**
Xournal
xournal
Xournal++
xournal++

**Additional context**
This happens primalary when on battery power, but sometimes even while i charge it. Maybe this happens because xournal++ is demanding not enough resources from the cpu? 

Edit: I noticed that the live previews of all tools are impacted.

David96:
Fixes #2242 

Also fixes strokes not disappearing until lifting the eraser.

The main problem was "changeTool" being called during the motion event,
for some reason firing those events there did cause the stroke to not
appear.

Only removing that call though resulted in the first stroke after
starting the application not appearing before lifting the stroke
(inverting my issue, which already was a big improvement). Also removing
those events then fixed that, too, as well as the eraser issue.

I honestly don't see, why pointCurrentToolToButtonTool and
pointCurrentToolToToolbarTool should be different here and I didn't
experience any issues so far but still, this is more an empirical "fix"
than actually understanding the real issue at hand (still took me hours
to debug).

Of course someone actually familiar with this code base should verify that this change doesn't break anything else. At least for me, drawing in general feels a lot smoother but that might also be my imagination and happiness to have finally fixed my problems with pen input.

### DIFF
--- a/src/control/ToolHandler.cpp
+++ b/src/control/ToolHandler.cpp
@@ -448,13 +448,7 @@ void ToolHandler::loadSettings() {
 
 void ToolHandler::pointCurrentToolToButtonTool() { this->currentTool = this->buttonSelectedTool; }
 
-void ToolHandler::pointCurrentToolToToolbarTool() {
-    this->currentTool = this->toolbarSelectedTool;
-
-    this->listener->toolColorChanged(false);
-    this->listener->toolSizeChanged();
-    this->fireToolChanged();
-}
+void ToolHandler::pointCurrentToolToToolbarTool() { this->currentTool = this->toolbarSelectedTool; }
 
 auto ToolHandler::getToolThickness(ToolType type) -> const double* { return this->tools[type - TOOL_PEN]->thickness; }
 
--- a/src/gui/inputdevices/PenInputHandler.cpp
+++ b/src/gui/inputdevices/PenInputHandler.cpp
@@ -169,8 +169,6 @@ auto PenInputHandler::actionMotion(InputEvent const& event) -> bool {
     GtkXournal* xournal = this->inputContext->getXournal();
     ToolHandler* toolHandler = this->inputContext->getToolHandler();
 
-    this->changeTool(event);
-
     if (toolHandler->getToolType() == TOOL_HAND) {
         if (this->deviceClassPressed) {
             this->handleScrollEvent(event);


idotobi:
That (#2242) is a weird bug. I cannot claim that I understand all the parts of that part of the code but at least the `pointCurrentToolToButtonTool` I've only added recently. I'll have a look at your PR this weekend.

David96:
I did a little more digging and apparently this minimal patch also fixes my problem:
```patch
diff --git a/src/gui/XournalppCursor.cpp b/src/gui/XournalppCursor.cpp
index 603f41b6..def248c4 100644
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -305,7 +305,7 @@ void XournalppCursor::updateCursor() {
         gtk_widget_set_sensitive(xournal->getWidget(), !this->busy);
     }

-    gdk_display_sync(gdk_display_get_default());
+    //gdk_display_sync(gdk_display_get_default());

     if (cursor != nullptr) {
         g_object_unref(cursor);
```
As you can see, I removed the gdk_display_sync from the updateCursor function. I noticed, that updateCursor was called by all three event handlers called by that pointCurrentToolTo… function. At least, I think it's performance wise very bad to call that function three times on every motion event, as it "Flushes any requests queued for the windowing system and waits until all requests have been handled".
I'm still not sure, why this causes the display to actually **not** be updated - maybe sway is at fault here getting overwhelmed by the flood of sync requests? That's just a wild guess though.

src/control/ToolHandler.cpp
```diff
@@ -448,13 +448,7 @@ void ToolHandler::loadSettings() {
 
 void ToolHandler::pointCurrentToolToButtonTool() { this->currentTool = this->buttonSelectedTool; }
 
-void ToolHandler::pointCurrentToolToToolbarTool() {
-    this->currentTool = this->toolbarSelectedTool;
-
-    this->listener->toolColorChanged(false);
-    this->listener->toolSizeChanged();
-    this->fireToolChanged();
```
idotobi, line 456: I double checked and the `toolColorChanged` and `toolSizeChanged` are superfluous as they are actually called withing `fireToolChanged`.
`fireToolChanged`, however, is important for the toolbar to always show the most current state so this should not be removed.

I also played arround with it and I think following change might help quite a lot performance wise:

```cpp
void ToolHandler::pointCurrentToolToButtonTool(Button button) {
    Tool* tool = getButtonTool(button);
    if (this->activeTool == tool)
        return;
    this->activeTool = tool;

    this->fireToolChanged();
}

void ToolHandler::pointCurrentToolToToolbarTool() {
    if (this->activeTool == this->toolbarSelectedTool)
        return;
    this->activeTool = this->toolbarSelectedTool;

    this->fireToolChanged();
}
```

If you don't mind I would include this change in my current ongoing PR (#2339).
If you could check if my change improves the situation in any way that would be great :nerd_face:

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -169,8 +169,6 @@ auto PenInputHandler::actionMotion(InputEvent const& event) -> bool {
     GtkXournal* xournal = this->inputContext->getXournal();
     ToolHandler* toolHandler = this->inputContext->getToolHandler();
 
-    this->changeTool(event);
```
idotobi, line 172: I agree it seems weird that `changeTool` would be called as part of this function.
Here, I would be a little more cautious though because I'd assume someone put it there for a reason.

Maybe it would be good to have this as a separate mini-PR so multiple people can check it on their system if they experience any weird behaviour. Performance-wise it would be quite an improvement. Good catch :+1:

wisp3rwind:
> I'm still not sure, why this causes the display to actually **not** be updated - maybe sway is at fault here getting overwhelmed by the flood of sync requests? That's just a wild guess though.

Is this issue compositor-specific? In #2242 you mentioned "gnome desktop" which would suggest mutter rather than sway. CPU usage might also skyrocket due to the GTK/sway issue https://github.com/xournalpp/xournalpp/issues/2398. (I didn't really look into this PR, your fixes may very well also be useful to improve performance.)

As you are already wondering whether GDK might be sending too many requests to the compositor, you could get a more useful log with `WAYLAND_DEBUG=1 xournalpp`.

David96:
@idotobi Thanks for your comments, I will take a look at them during the week.

@wisp3rwind #2242 wasn't reported by me ;) I'm using sway.
The WAYLAND_DEBUG thing is how I figured out that the repeated display sync calls might be a problem in the first place (see my comment above and the minimal patch included in it).
I also experience a very high CPU usage of sway (hadn't checked before), but that appears to be a different issue from this one.
This still doesn't explain, why (too) many display sync requests cause the display **not** to be updated properly, but I don't know a whole lot about this anyway.

wisp3rwind:
> @wisp3rwind #2242 wasn't reported by me ;) I'm using sway.

Oh, sorry, I got that mixed up :facepalm: 

> The WAYLAND_DEBUG thing is how I figured out that the repeated display sync calls might be a problem in the first place (see my comment above and the minimal patch included in it).

Yeah, I had only seen the logs in the original post of #2242.

> I also experience a very high CPU usage of sway (hadn't checked before), but that appears to be a different issue from this one.
> This still doesn't explain, why (too) many display sync requests cause the display **not** to be updated properly, but I don't know a whole lot about this anyway.

I was just browsing xournal++ issues, and this one sounded like it might only be exposed due to high CPU usage in the first place, which is why I left the comment. Even if this is indeed not observable with that GTK/sway issue fixed, it's surely a good thing to avoid unnecessary requests sent to the compositor :)

src/gui/inputdevices/PenInputHandler.cpp
LittleHuba: That is actually on purpose. Some stylised do not send ENTER or LEAVE events for some reason. For those we must catch the motion events to handle the necessary updates when a pointer enters the drawing area. Also moving from background to a page in the drawing area must update the pointer which was so far handled by this call. 

There might be room to improve this nowadays. That code is at least two years old.

David96:
Alright, so using @idotobi's proposed patch (after fixing it up) also fixes my issue with strokes. Erasing doesn't happen instantaneously, but that is not as bad as the strokes. Erasing works, when removing the changeTool call from the motion handler but if that's actually needed for some pens to work properly I get that you don't want to remove it (I'll see whether the GTK fix mentioned by @wisp3rwind will change the situation there).
So I'd suggest I either change this PR to only include @idotobi's patch or close it and @idotobi includes it in his other PR.

Fixed patch:
```diff
diff --git a/src/control/ToolHandler.cpp b/src/control/ToolHandler.cpp
index 3fdfdc35..e0d413cb 100644
--- a/src/control/ToolHandler.cpp
+++ b/src/control/ToolHandler.cpp
@@ -446,13 +446,19 @@ void ToolHandler::loadSettings() {
     }
 }
 
-void ToolHandler::pointCurrentToolToButtonTool() { this->currentTool = this->buttonSelectedTool; }
+void ToolHandler::pointCurrentToolToButtonTool() {
+    if (this->currentTool == this->buttonSelectedTool)
+        return;
+    this->currentTool = this->buttonSelectedTool;
+
+    this->fireToolChanged();
+}
 
 void ToolHandler::pointCurrentToolToToolbarTool() {
+    if (this->currentTool == this->toolbarSelectedTool)
+        return;
     this->currentTool = this->toolbarSelectedTool;
 
-    this->listener->toolColorChanged(false);
-    this->listener->toolSizeChanged();
     this->fireToolChanged();
 }
```

PR #2406, opened 2020-11-15
Extend the Lua Plugin API for page and layer operations

rolandlo:
This PR adds various new functions to the Lua Plugin API. Moreover a plugin is added for cloning all handwritten layers to the next page. 

The **new Plugin API functions** are:
- `app.sidebarAction` for actions accessible in the sidebar
- `app.layerAction` for actions accessible in the layer controller
- `app.changePdfBackgroundPageNr` to change the pdf background page number
- `app.getDocumentStructure` to get lots of useful info on the document, also used for applying operations on all/selected pages
- `app.scrollToPage`scrolls relatively or absolutely to a page
- `app.scrollToPos` scrolls relatively or absolutely to a new position on the same page
- `app.setCurrentPage` sets the given page as new current page
- `app.setPageSize` to set the size of the current page
- `app.setCurrentLayer` sets the given layer as the new current layer, and updates visibility if specified
- `app.setLayerVisibility` sets the visibility of the current layer

In order for the plugin to work correctly, the `XournalView::pageDeleted` function had to be fixed, so that the page layout is updated immediately. This will not be necessary after merging PR #2425 

A test plugin is now available which tests out the functionality of various API functions, mostly combined with each other. I am not sure if the test plugin should be merged or whether it is just for the reviewer to have an easier job.

- [x] Still some more testing is required
- [x] Some error messages will have to be included as well, if the user wants to execute impossible actions.

**Edit 1**: added `app.setLayerVisibility`
**Edit 2**: added `app.setPageSize`

**Edit 3**: The PR is now ready for review.

### DIFF
--- a/src/plugin/luapi_application.h
+++ b/src/plugin/luapi_application.h
@@ -15,7 +15,9 @@
 #include "control/Control.h"
 #include "control/PageBackgroundChangeController.h"
 #include "control/Tool.h"
+#include "control/layer/LayerController.h"
 #include "control/pagetype/PageTypeHandler.h"
+#include "gui/widgets/XournalWidget.h"
 
 #include "StringUtils.h"
 #include "XojMsgBox.h"
@@ -24,6 +26,8 @@ using std::map;
 /**
  * Create a 'Save As' native dialog and return as a string
  * the filepath of the location the user chose to save.
+ *
+ * Example: local filename = app.saveAs()
  */
 static int applib_saveAs(lua_State* L) {
     GtkFileChooserNative* native;
@@ -57,9 +61,8 @@ static int applib_saveAs(lua_State* L) {
 }
 
 /**
- * Example:
- * app.msgbox("Test123", {[1] = "Yes", [2] = "No"})
- * Return 1 for yes, 2 for no in this example
+ * Example: local result = app.msgbox("Test123", {[1] = "Yes", [2] = "No"})
+ * Pops up a message box with two buttons "Yes" and "No" and returns 1 for yes, 2 for no
  */
 static int applib_msgbox(lua_State* L) {
     const char* msg = luaL_checkstring(L, 1);
@@ -89,7 +92,11 @@ static int applib_msgbox(lua_State* L) {
 
 
 /**
- * Allow to register menupoints or toolbar icons, this needs to be called from initUi
+ * Allow to register menupoints, this needs to be called from initUi
+ *
+ * Example: app.registerUi({["menu"] = "HelloWorld", callback="printMessage", accelerator="<Control>a"})
+ * registers a menupoint with name "HelloWorld" executing a function named "printMessage",
+ * which can be triggered via the "<Control>a" keyboard accelerator
  */
 static int applib_registerUi(lua_State* L) {
     Plugin* plugin = Plugin::getPluginFromLua(L);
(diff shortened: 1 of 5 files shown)


src/plugin/luapi_application.h
```diff
@@ -214,7 +237,58 @@ static int applib_uiActionSelected(lua_State* L) {
...
+            {"NEW_BEFORE", SIDEBAR_ACTION_NEW_BEFORE},
+            {"NEW_AFTER", SIDEBAR_ACTION_NEW_AFTER},
+    };
+    const char* actionStr = luaL_checkstring(L, 1);
+    if (actionStr == nullptr) {
+        luaL_error(L, "Missing action!");
+    }
+    auto pos = actionMap.find(actionStr);
+    if (pos == actionMap.end()) {
+        luaL_error(L, "Action unknown!");
```
Technius, line 261: This message would be more helpful if the value of `actionStr` is shown.

src/plugin/luapi_application.h
```diff
@@ -214,7 +237,58 @@ static int applib_uiActionSelected(lua_State* L) {
...
+/**
+ * Execute action from layer controller
+ *
+ * Example: app.layerAction("ACTION_DELETE_LAYER")
+ * deletes the current layer
+ */
+static int applib_layerAction(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+
+    ActionType action = ActionType_fromString(luaL_checkstring(L, 1));
```
Technius, line 279: This should warn if the action is `ACTION_NONE` (i.e. invalid action).

src/plugin/luapi_application.h
```diff
@@ -305,13 +385,381 @@ static int applib_changeToolColor(lua_State* L) {
...
+    if (lua_isboolean(L, 3)) {
+        relative = lua_toboolean(L, 3);
+    }
+
+    if (relative) {
+        width += page->getWidth();
+        height += page->getHeight();
+    }
+
+    if (width > 0 and height > 0) {
```
Technius, line 670: `&&` instead of `and` for consistency.

src/plugin/luapi_application.h
```diff
@@ -305,13 +385,381 @@ static int applib_changeToolColor(lua_State* L) {
...
+    return 1;
+}
+
+/*
+ * Sets the visibility of the current layer
+ *
+ * Example: app.setLayerVisibility(true)
+ * makes the current layer visible
+ */
+static int applib_setLayerVisibility(lua_State* L) {
```
Technius, line 727: I think this would be more useful if you could supply an optional layer number parameter.

Technius (commented):
This is great! I did a brief pass through the code for now, will take a look in more detail when I have time.

src/plugin/luapi_application.h
rolandlo: If the action is invalid then `ActionType_fromString` will show an error message. It won't use `ACTION_NONE` as return value. I could check manually if `action` is one of the Action types handled in `LayerController::actionPerformed` (those are `ACTION_NEW_LAYER` , `ACTION_DELETE_LAYER`, `ACTION_GOTO_PREVIOUS_LAYER` and `ACTION_GOTO_TOP_LAYER`), but if more actions are added or the actions are renamed, this check will have to be updated again. So what should I do?

src/plugin/luapi_application.h
rolandlo: On the other hand, one can simply use `app.setCurrentLayer` followed by `app.setLayerVisibility` to achieve the same thing (one may need to set the current layer back at the end though). In general I tried to keep the API simple and let the Lua code set (or iterate through) layers and pages. See also e.g. `applib_setPageSize`, where also no page number is specified.
If you insist, I can change it, of course.

Technius: Fair enough. We could easily add an extra parameter if someone complains that the API is clunky.

src/plugin/luapi_application.h
Technius: Ah, I didn't see that error message.

Technius (approved):
LGTM

rolandlo:
I have removed the plugin to test the API. For further use I will place it here in zipped form. In particular it contains the function to change all page sizes simultaneously to the current page size, and a demo plugin for autoscrolling
[TestAPI.zip](https://github.com/xournalpp/xournalpp/files/5721261/TestAPI.zip)
.

PR #2407, opened 2020-11-16
Changing the console encoding to utf-8 on windows

### LINKED ISSUE #2395: Under windows there is something wrong with the terminal messages:
Under windows there is something wrong with the terminal messages: 
grafik

_Originally posted by @rolandlo in https://github.com/xournalpp/xournalpp/issues/2391#issuecomment-726762477_

Febbe:
- This only works for windows v. >= 1804

Fixes #2395

### DIFF
--- a/src/Xournalpp.cpp
+++ b/src/Xournalpp.cpp
@@ -24,7 +24,7 @@ auto main(int argc, char* argv[]) -> int {
 #ifdef _WIN32
     // Show and hide the console here. Otherwise, gspawn-win32-helper will create annoying console popups.
     AllocConsole();
-    SetConsoleOutputCP(CP_UTF8); //< Quick fix for #2395
+    SetConsoleOutputCP(CP_UTF8); // Quick fix for #2395
     ShowWindow(GetConsoleWindow(), SW_HIDE);
 #endif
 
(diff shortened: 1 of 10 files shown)


Febbe:
This is only a Quick fix. In my opinion, all `»` and `«` should be replaced by upper `"`, which are more internationally. 
`»` and `«` are mostly used in middle Europe. And there they are used differently: By France and Switzerland `«text»`. In Germany and Austria »text«.

Technius:
> This only works for windows v. >= 1804

If this is the case, it would be better to just replace the double angle bracket characters with double quotes. There's no reason to break compatibility so significantly for such a trivial issue (99% of Windows users will not care about console output).

Febbe:
The thing is, that this would break all translations with those "Guillemets".

Febbe:
> > This only works for windows v. >= 1804
> 
> If this is the case, it would be better to just replace the double angle bracket characters with double quotes. There's no reason to break compatibility so significantly for such a trivial issue (99% of Windows users will not care about console output).

I think xournalpp will still run, but the console output just does not change to utf-8.

rolandlo (approved):
Console messages look fine with this PR on Windows.

LittleHuba:
Will merge tomorrow if nobody has an issue with it.

PR #2413, opened 2020-11-16
fix presentation mode after startup

### LINKED ISSUE #1690: From last update(201912241606) I can't edit.
**Affects versions :**
 - OS: Lubuntu 18.04.1
 - libgtk 3.22.30-1ubuntu4
 - Xournal++ 1.0.6-0-201912241606-ubuntu18.04.1
(201912291941 the same problem)

**Describe the bug**
It opens the files but I can't edit, nor existing neither new ones.

rolandlo:
Fixes #1690 according to my tests. 
I included a warning so that we would see it right away if this code change created any unintended complications. The warning can be removed later.

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -169,8 +169,8 @@ auto ZoomControl::getZoomFitValue() const -> double { return this->zoomFitValue;
 auto ZoomControl::updateZoomPresentationValue(size_t pageNo) -> bool {
     XojPageView* page = view->getViewFor(view->getCurrentPage());
     if (!page) {
-        // no page
-        return false;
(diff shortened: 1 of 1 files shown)


Technius:
As far as I can tell, this makes it so that when presentation mode is enabled and no page in the document exists, the application will not try to select the nonexistent page (e.g. such as immediately during startup or before a file is finished loading). Is my understanding correct? Need to know for changelog purposes.

rolandlo:
If presentation mode is enabled and no page in the document exists, the application will set zoom presentation mode (disable zoom widgets for zooming in and out and for setting zoom fit mode and zoom100 mode) without actually computing the zoom presentation value. The zoom presentation value is computed once the document loads.

Before that change, during startup, zoom presentation mode was disabled because the zoom presentation value could not be computed with no page available.

PR #2416, opened 2020-11-17
Revert "New icon app"

LittleHuba:
Reverts xournalpp/xournalpp#1808 in response to comment on merge commit showing major issues. Windows releases do not have icons anymore. What I can see, there is a divergence in the file list.


Technius:
Looks like the icon png got deleted. This shouldn't cause the icon to disappear though--wouldn't GTK just load the svg icon instead?

PR #2422, opened 2020-11-18
Add several plugin loading paths

### LINKED ISSUE #1155: Loading custom plugins

Febbe:
- [x] Make duplicates differentiable
 - [x] Do not open all duplicates with the same name.

fixes #1155

### DIFF
--- a/src/core/plugin/Plugin.h
+++ b/src/core/plugin/Plugin.h
@@ -80,6 +80,9 @@ public:
     /// Plugin version
     auto getVersion() const -> std::string const&;
 
+    /// @return the path to the plugin folder. Each plugin should have a unique path.
+    auto getPath() const -> fs::path const&;
+
     /// The plugin is enabled
     auto isEnabled() const -> bool;
 
@@ -134,5 +137,5 @@ private:
 };
 
 #else
-struct Plugin final {};  ///< empty struct, since Plugin is not compiled
+struct Plugin final {};  ///< empty struct, since Plugins are not available
 #endif
--- a/src/core/plugin/PluginController.cpp
+++ b/src/core/plugin/PluginController.cpp
@@ -1,8 +1,13 @@
 #include "PluginController.h"
 
-#include <algorithm>
+#include <memory>
+#include <tuple>
+#include <utility>
 
-#include <config-features.h>
+
+#ifdef ENABLE_PLUGINS
+#include <algorithm>
+#include <filesystem>
 
 #include "control/Control.h"
 #include "gui/GladeSearchpath.h"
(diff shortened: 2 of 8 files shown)


rolandlo:
It would be nice if this PR could be finished. Otherwise, rather than having no custom plugins folder, we could use #3063

Technius:
1.1 is out so this could be revisited.

Febbe:
@rolandlo I think it is 90% finished.
I only have to remove some changes which do not belong into this PR.

Febbe:
grafik

Febbe:
I did not add #3063 since it is easy to add the extra path to this version, after it is merged.
So @UweStrohbeck has the chance to contribute his work without throwing it away :) .

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
-#endif
-}
-
-void PluginController::loadPluginsFrom(fs::path const& path) {
-#ifdef ENABLE_PLUGINS
-    Settings* settings = control->getSettings();
-    std::vector<std::string> pluginEnabled = StringUtils::split(settings->getPluginEnabled(), ',');
-    std::vector<std::string> pluginDisabled = StringUtils::split(settings->getPluginDisabled(), ',');
+namespace {
+auto load_awailable_plugins_from(fs::path const& path, Control* control) -> std::vector<std::unique_ptr<Plugin>> {
```
Technius, line 17: typo: `load_available_plugins_from`

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
+    // Todo(fabian) move those search paths into PathUtils
+    auto searchPath = control->getGladeSearchPath()->getFirstSearchPath();
+    auto searchPaths = {fs::weakly_canonical(searchPath /= "../plugins"),  //
+                        fs::weakly_canonical(searchPath /= "/plugins"),    //
+                        Util::getConfigSubfolder("plugins")};
+
+    for (auto&& path: searchPaths) {
+        for (auto&& plugin: load_awailable_plugins_from(path, control)) {
+            emplace_sorted_if_not_exists(plugins, std::move(plugin));
+        }
```
Technius, line 76: This is O(n^2). Could we improve this by using an ordered map (indexed by path of plugin folder) to track the loaded plugins?

src/core/plugin/Plugin.h
```diff
@@ -80,6 +80,9 @@ class Plugin final {
     /// Plugin version
     auto getVersion() const -> std::string const&;
 
+    /// @return the path to the plugin folder
```
Technius, line 83:
```suggestion
    /// @return the path to the plugin folder. Each plugin should have a unique path.
```

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
+    }
+
+    Settings* settings = control->getSettings();
+    // enable plugins
+    std::vector<std::string> pluginEnabled = StringUtils::split(settings->getPluginEnabled(), ',');
+    for (auto&& plugin_path: pluginEnabled) {
+        if (auto iter = std::find_if(begin(plugins), end(plugins),
+                                     [&plugin_path](auto&& plugin) { return plugin->getPath() == plugin_path; });
```
Technius, line 84: This is `O(len(enabled) * n) ~= O(n^2)` assuming `enabled = n`. If the map above is defined, then I think we can turn this into a map lookup.

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
+            iter != end(plugins)) {
+            (*iter)->setEnabled(true);
         }
-    } catch (fs::filesystem_error const& e) {
-        g_warning("Could not open plugin dir: \"%s\"", path.string().c_str());
-        return;
+    }
+    // disable plugins
+    std::vector<std::string> pluginDisabled = StringUtils::split(settings->getPluginDisabled(), ',');
+    for (auto&& plugin_path: pluginDisabled) {
```
Technius, line 91: Same O(len(enabled) * n) problem here.

src/core/plugin/PluginController.cpp
Febbe: Yes this can be improved, for example with an boost::sorted_vector(?) Or by using std::lower_bound.

I decided to use a vector over a map, because we first won't have many Plugins so that it matters. 
Second, a std::map operation might be 100 times slower.
Third the effort to implement my own container is not worth it (using boost is not an option currently).

Ps until yesterday, I used a std::map for this.

src/core/plugin/Plugin.h
```diff
@@ -133,6 +138,10 @@ class Plugin final {
     bool valid = false;           ///< Flag if the plugin is valid / correct loaded
 };
 
+constexpr bool operator==(Plugin const& lhs, Plugin const& rhs) noexcept { return tieer(lhs) == tieer(rhs); }
+
+constexpr bool operator<(Plugin const& lhs, Plugin const& rhs) noexcept { return tieer(lhs) == tieer(rhs); }
```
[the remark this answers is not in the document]
Febbe: I forgot, that the comparison is by dept first. But does it matter in this case?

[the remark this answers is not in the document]
Technius: In particular, I think this can be fixed with
```suggestion
constexpr bool operator<(Plugin const& lhs, Plugin const& rhs) noexcept { return tieer(lhs) < tieer(rhs); }
```
Otherwise sorting will not work correctly.

src/core/plugin/Plugin.h
```diff
@@ -116,6 +119,8 @@ class Plugin final {
     /// Get Plugin from lua engine
     static auto getPluginFromLua(lua_State* lua) -> Plugin*;
 
+    constexpr friend auto tieer(Plugin const& plugin) { return std::tie(plugin.name, plugin.version, plugin.path); }
```
Technius, line 122: I think it's more readable to make the return type of this explicit. I think it's probably something like:
```c++
using key_type = std::tuple<const std::string&, const std::string&, const fs::path&>;
```

I would also rename this to something more descriptive like `getKeyTuple()` and add a description:
```c++
/// Returns a tuple that is suitable for identifying this plugin (e.g., to be used in a map, set, or sorted vector).
```

Lastly, to avoid unnecessary comparisons, we should use the most common differentiator (the path) as the first component of the tuple.

src/core/plugin/PluginController.cpp
Technius: This is kind of a small perf improvement thing but also a low hanging fruit. I think the easiest and most effective improvement for now is to use `std::lower_bound` (as long as we ensure that plugins is pre-sorted, which should not be difficult to implement and maintain). For the low plugin count case, the overhead is negligible; and this should hopefully scale better in the high plugin count case.

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
+    auto it = std::find_if(std::begin(plugins), endit,
+                           [&path](auto const& p) { return fs::canonical(p->getPath()) == fs::canonical(path); });
+    if (it != endit) {
+        return;
+    }
+    plugins.emplace_back(std::move(plugin));
+    std::sort(begin(plugins), end(plugins), [](auto const& lhs, auto const& rhs) {
+        auto tieer = [](auto&& p) { return std::tie(p->getName(), p->getPath()); };
+        return tieer(lhs) < tieer(rhs);
+    });
```
Technius, line 59: Could this be improved by using `std::lower_bound` instead of inserting and sorting? Also, why not `*lhs < *rhs`?

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,106 @@
...
+            iter != end(plugins)) {
+            (*iter)->setEnabled(false);
+        }
+    }
+
+    for (auto&& plugin: plugins) {
+        if (!plugin->isEnabled()) {
+            continue;
+        }
+        plugin->loadScript();
```
Technius, line 103: nitpick:
```suggestion
        if (plugin->isEnabled()) {
            plugin->loadScript();
        }
```

src/core/plugin/PluginController.cpp
Technius: Or use `std::lower_bound` as noted above.

src/core/plugin/Plugin.h
Technius: Oh, then again, changing the order of the tuple will also modify the behavior of `<`. Can you clarify the reasoning behind the current ordering?

Technius:
Also, please add some unit tests that check the plugin loading logic.

src/core/plugin/PluginController.h
```diff
@@ -52,7 +52,8 @@ class PluginController final {
     Control* control;
 
     /**
-     * All loaded Plugins
+     * All loaded Plugins, sorted by name and path
+     * Todo: replace with boost::sorted_vector
```
Febbe, line 56:
```suggestion
     * Todo: replace with boost::flat_map
```

src/core/plugin/PluginController.cpp
Febbe: Replaced by a Comparator Class

Febbe:
> 

That's very hard, the hole Plugin loading logic is based on Control and Settings. A test independently to those would require a huge refactoring. I only could test the static functions. That would expose them to the global namespace, unfortunately.

I wished that we would have some sort of [[test]] tag in c++ similar to rust.

src/core/plugin/PluginController.cpp
```diff
@@ -1,54 +1,137 @@
...
 
-            plugin->loadScript();
 
-            this->plugins.emplace_back(std::move(plugin));
+PluginController::PluginController(Control* control): control(control) {
+#ifdef ENABLE_PLUGINS
+    // Todo(fabian) move those search paths into PathUtils
+    auto searchPath = control->getGladeSearchPath()->getFirstSearchPath();
+    auto searchPaths = {fs::weakly_canonical(searchPath /= "../plugins"),  //
+                        fs::weakly_canonical(searchPath /= "/plugins"),    //
```
rolandlo, line 100: The `searchPath` is the `ui` subfolder of the [Shared resources folder](https://xournalpp.github.io/guide/file-locations/#shared-resources-folder), at least if it exists. If the `ui`-subfolder does not exist, it would be the shared resources folder itself, but I don't think Xournal++ can run without the `ui`-subfolder.
So I don't understand why adding `searchPath /= "/plugins"` is needed.

rolandlo (approved):
Thanks @Febbe for taking your time to finishing this PR. 
I have only one minor question about the code (see below). Otherwise it looks all good to me.

Febbe:
You can merge, after the pipeline successes.

src/core/plugin/PluginController.cpp
Febbe: Fixed

Febbe:
Ignore clang-format, my clang-format is newer.

PR #2425, opened 2020-11-19
Fix page layout update in a unified way

### LINKED ISSUE #1777: page changing when switching between paired and non paired display
being on a page, in paired pages view, desactivating such mode changes page to another one, without an apparent criterion. switching back to paired pages mode bring to another page, different from the previous ones

### LINKED ISSUE #2411: Zooming looses the spot in the document
**Affects versions :**
 - Xubuntu 18.04 
 - Desktop: Xfce
 - version of libgtk 3.22.30
 - Version of Xournal++: 1.0.19 
 - Installation method: PPA

**Describe the bug**
Changing from 100% zoom to fit page zoom, jumps the view around in the document. 

**To Reproduce**
Steps to reproduce the behaviour:
Open a multi page document. 
Do a zoom to 100%, scroll a few pages. 
Do a zoom to fit. 
The view of the document jumps back to where you started scrolling.  

**Expected behavior**
doing the above should show the same bit of document just in different zoom depths.

rolandlo:
Fixes #1777. 
Partially fixes #2411.

Also fixes page update after page deletion (which is relevant for navigating through pages in plugins) and after other changes of page layout than switching from paired view to non-paired view.

I also simplified the code in `XournalMain::run`, since the timing issue is solved with the new page layout method anyway.

Note that zooming in and out may still change the active page (although not drastically any more). This is due to the fact, that the `Layout::updateVisibility() method computes the most visible page and makes that page the active page and that `scrollToPage` makes the top of the page visible.

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -404,9 +404,8 @@ auto XournalMain::run(int argc, char* argv[]) -> int {
     checkForErrorlog();
     checkForEmergencySave(control);
 
-    // There is a timing issue with the layout
-    // This fixes it, see #405
-    Util::execInUiThread([=]() { control->getWindow()->getXournal()->layoutPages(); });
+    // There was a timing issue with the layout, see #405
+    control->getWindow()->getXournal()->layoutPages();
 
     if (migrateResult.status != MigrateStatus::NotNeeded) {
         Util::execInUiThread([=]() { XojMsgBox::showErrorToUser(control->getGtkWindow(), migrateResult.message); });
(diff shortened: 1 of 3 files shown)


pindakaas42:
Tested this and it does improve #2411 but it does not fully solve the problem, the zoom to fit still jumps to the top of the page, way better, but it means if I zoom to the bottom of a page in the 100% zoom and then do a zoom to fit, I then have scroll all the way down the page again.

rolandlo:
I think this PR definitively doesn't do any harm and solves some problems. So I merge it.

PR #2435, opened 2020-11-23
Better Touchscreen Zooming

personalizedrefrigerator:
## Addresses these issues:
 * Previously, when touch drawing was enabled, the touchscreen could not be used to pan and zoom. 
 * Additionally, zoom with touch caused the page to scroll up and down by large amounts (zoom centered at top-left of document).
      * **Edit:** This only happened when touch drawing/the touchscreen workaround was enabled.
 * Zooming can be very slow when editing large PDFs.
      * **Only partially fixes this issue** 

## Adds these features:
  * Two-finger touch panning/scrolling when touch drawing is enabled.
  * An option to clear the PDF page cache less frequently.
  * Pressure inference (faster strokes -> lower pressure).
      * An option to toggle this in settings.

## Screenshot
image

## Questions I Have:
 * What is the reason for the workaround that enables touch drawing? Is it reasonable to allow panning and zooming when the user uses multi-touch with this workaround enabled?

### DIFF
--- a/src/control/PdfCache.cpp
+++ b/src/control/PdfCache.cpp
@@ -1,13 +1,15 @@
 #include "PdfCache.h"
 
+#include <algorithm>
 #include <cstdio>
 #include <utility>
 
 class PdfCacheEntry {
 public:
-    PdfCacheEntry(XojPdfPageSPtr popplerPage, cairo_surface_t* img) {
+    PdfCacheEntry(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
         this->popplerPage = std::move(popplerPage);
         this->rendered = img;
+        this->zoom = zoom;
     }
 
     ~PdfCacheEntry() {
@@ -15,6 +17,8 @@ public:
         cairo_surface_destroy(this->rendered);
         this->rendered = nullptr;
     }
+
+    double zoom;
     XojPdfPageSPtr popplerPage;
     cairo_surface_t* rendered;
 };
@@ -30,14 +34,9 @@ PdfCache::~PdfCache() {
     this->size = 0;
 }
 
-void PdfCache::setZoom(double zoom) {
-    if (this->zoom == zoom) {
-        return;
-    }
-    this->zoom = zoom;
+void PdfCache::setZoom(double zoom) { this->zoom = zoom; }
 
-    clearCache();
-}
+void PdfCache::setZoomingClearsCache(bool clears) { this->zoomClearsCache = clears; }
 
 void PdfCache::clearCache() {
     for (PdfCacheEntry* e: this->data) {
@@ -46,24 +45,26 @@ void PdfCache::clearCache() {
     this->data.clear();
 }
 
-auto PdfCache::lookup(const XojPdfPageSPtr& popplerPage) -> cairo_surface_t* {
+auto PdfCache::lookup(const XojPdfPageSPtr& popplerPage) -> PdfCacheEntry* {
     for (PdfCacheEntry* e: this->data) {
         if (e->popplerPage->getPageId() == popplerPage->getPageId()) {
-            return e->rendered;
+            return e;
         }
     }
 
     return nullptr;
 }
 
-void PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img) {
-    auto* ne = new PdfCacheEntry(std::move(popplerPage), img);
-    this->data.push_front(ne);
-
-    while (this->data.size() > this->size) {
+PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
+    while (this->data.size() >= this->size) {
         delete this->data.back();
         this->data.pop_back();
     }
+
+    auto* ne = new PdfCacheEntry(std::move(popplerPage), img, zoom);
+    this->data.push_front(ne);
+
+    return ne;
 }
 
 void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoom) {
@@ -71,28 +72,36 @@ void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoo
 
     this->setZoom(zoom);
 
-    cairo_surface_t* img = lookup(popplerPage);
-    if (img == nullptr) {
-        img = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, popplerPage->getWidth() * this->zoom,
-                                         popplerPage->getHeight() * this->zoom);
+    PdfCacheEntry* cacheResult = lookup(popplerPage);
+
+    if (cacheResult == nullptr || this->zoom < cacheResult->zoom / 2.0 && this->zoom > 1.0 ||
+        this->zoom > cacheResult->zoom * 2.0
+
+        || this->zoomClearsCache && this->zoom != cacheResult->zoom) {
+
+        double renderZoom = std::max(zoom, 1.0);
+
+        auto* img = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, popplerPage->getWidth() * renderZoom,
+                                               popplerPage->getHeight() * renderZoom);
         cairo_t* cr2 = cairo_create(img);
 
-        cairo_scale(cr2, this->zoom, this->zoom);
+        cairo_scale(cr2, renderZoom, renderZoom);
         popplerPage->render(cr2, false);
         cairo_destroy(cr2);
-        cache(popplerPage, img);
+
+        cacheResult = cache(popplerPage, img, renderZoom);
     }
 
     cairo_matrix_t mOriginal;
     cairo_matrix_t mScaled;
     cairo_get_matrix(cr, &mOriginal);
     cairo_get_matrix(cr, &mScaled);
-    mScaled.xx = 1;
-    mScaled.yy = 1;
+    mScaled.xx = this->zoom / cacheResult->zoom;
+    mScaled.yy = this->zoom / cacheResult->zoom;
     mScaled.xy = 0;
     mScaled.yx = 0;
     cairo_set_matrix(cr, &mScaled);
-    cairo_set_source_surface(cr, img, 0, 0);
+    cairo_set_source_surface(cr, cacheResult->rendered, 0, 0);
     cairo_paint(cr);
     cairo_set_matrix(cr, &mOriginal);
 
--- a/src/control/PdfCache.cpp
+++ b/src/control/PdfCache.cpp
@@ -6,11 +6,17 @@
 
 class PdfCacheEntry {
 public:
-    ///  Cache [img], the result of rendering [popplerPage] with
-    /// the given [zoom].
-    ///  A change in the document's zoom causes a change in the
-    /// quality of the PDF backgrounds (zoomed in => need a higher
-    /// quality rendering).
+    /**
+     *   Cache [img], the result of rendering [popplerPage] with
+     * the given [zoom].
+     *  A change in the document's zoom causes a change in the
+     * quality of the PDF backgrounds (zoomed in => need a higher
+     * quality rendering).
+     *
+     * @param popplerPage
+     * @param img is the result of rendering popplerPage
+     * @param zoom is the zoom at which the page was rendered.
+     */
     PdfCacheEntry(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
         this->popplerPage = std::move(popplerPage);
         this->rendered = img;
@@ -41,7 +47,9 @@ PdfCache::~PdfCache() {
 
 void PdfCache::setZoom(double zoom) { this->zoom = zoom; }
 
-void PdfCache::setZoomingClearsCache(bool clears) { this->zoomClearsCache = clears; }
+void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
+
+void PdfCache::setAnyZoomChangeCausesRecache(bool b) { this->zoomClearsCache = b; }
 
 void PdfCache::clearCache() {
     for (PdfCacheEntry* e: this->data) {
@@ -78,16 +86,21 @@ void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoo
     this->setZoom(zoom);
 
     PdfCacheEntry* cacheResult = lookup(popplerPage);
+    bool needsRefresh{cacheResult == nullptr};
 
-    if (cacheResult == nullptr
+    if (cacheResult != nullptr) {
+        double averagedZoom = (this->zoom + cacheResult->zoom) / 2.0;
+        double percentZoomChange = std::abs(cacheResult->zoom - this->zoom) * 100.0 / averagedZoom;
+
         // If we do have a cached result, is its rendering quality
         // acceptable for our current zoom?
-        || this->zoom < cacheResult->zoom / 2.0 && this->zoom > 1.0 ||
-        this->zoom > cacheResult->zoom * 2.0
+        needsRefresh = this->zoom > 1.0 && percentZoomChange > this->zoomRefreshThreshold
 
-        // Has the user requested that we **always** clear the cache on zoom?
-        || this->zoomClearsCache && this->zoom != cacheResult->zoom) {
+                       // Has the user requested that we **always** clear the cache on zoom?
+                       || this->zoomClearsCache && this->zoom != cacheResult->zoom;
+    }
 
+    if (needsRefresh) {
         double renderZoom = std::max(zoom, 1.0);
 
         auto* img = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, popplerPage->getWidth() * renderZoom,
--- a/src/control/PdfCache.cpp
+++ b/src/control/PdfCache.cpp
@@ -6,6 +6,11 @@
 
 class PdfCacheEntry {
 public:
+    ///  Cache [img], the result of rendering [popplerPage] with
+    /// the given [zoom].
+    ///  A change in the document's zoom causes a change in the
+    /// quality of the PDF backgrounds (zoomed in => need a higher
+    /// quality rendering).
     PdfCacheEntry(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
         this->popplerPage = std::move(popplerPage);
         this->rendered = img;
@@ -56,7 +61,7 @@ auto PdfCache::lookup(const XojPdfPageSPtr& popplerPage) -> PdfCacheEntry* {
 }
 
 PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
-    while (this->data.size() >= this->size) {
+    while (this->data.size() > this->size) {
         delete this->data.back();
         this->data.pop_back();
     }
@@ -74,9 +79,13 @@ void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoo
 
     PdfCacheEntry* cacheResult = lookup(popplerPage);
 
-    if (cacheResult == nullptr || this->zoom < cacheResult->zoom / 2.0 && this->zoom > 1.0 ||
+    if (cacheResult == nullptr
+        // If we do have a cached result, is its rendering quality
+        // acceptable for our current zoom?
+        || this->zoom < cacheResult->zoom / 2.0 && this->zoom > 1.0 ||
         this->zoom > cacheResult->zoom * 2.0
 
+        // Has the user requested that we **always** clear the cache on zoom?
         || this->zoomClearsCache && this->zoom != cacheResult->zoom) {
 
         double renderZoom = std::max(zoom, 1.0);
--- a/src/control/ToolHandler.cpp
+++ b/src/control/ToolHandler.cpp
@@ -14,11 +14,13 @@
 
 ToolListener::~ToolListener() = default;
 
-ToolHandler::ToolHandler(ToolListener* listener, ActionHandler* actionHandler, Settings* settings) {
+ToolHandler::ToolHandler(ToolListener* stateChangeListener, ActionHandler* actionHandler, Settings* settings) {
     this->settings = settings;
     initTools();
-    this->listener = listener;
+    this->toolChangeListeners = std::vector<std::function<void(ToolType)> >();
     this->actionHandler = actionHandler;
+
+    this->stateChangeListener = stateChangeListener;
 }
 
 void ToolHandler::initTools() {
@@ -167,9 +169,15 @@ void ToolHandler::selectTool(ToolType type) {
 }
 
 void ToolHandler::fireToolChanged() {
-    if (listener) {
-        listener->toolChanged();
+    for (auto listener: this->toolChangeListeners) {
+        listener(this->activeTool->type);
     }
+
+    stateChangeListener->toolChanged();
+}
+
+void ToolHandler::addToolChangedListener(std::function<void(ToolType)> listener) {
+    toolChangeListeners.push_back(listener);
 }
 
 auto ToolHandler::getTool(ToolType type) -> Tool& { return *(this->tools[type - TOOL_PEN]); }
@@ -204,7 +212,7 @@ void ToolHandler::setPenSize(ToolSize size) {
     this->tools[TOOL_PEN - TOOL_PEN]->setSize(size);
 
     if (this->activeTool->type == TOOL_PEN) {
-        this->listener->toolSizeChanged();
+        this->stateChangeListener->toolSizeChanged();
     }
 }
 
@@ -212,7 +220,7 @@ void ToolHandler::setEraserSize(ToolSize size) {
     this->tools[TOOL_ERASER - TOOL_PEN]->setSize(size);
 
     if (this->activeTool->type == TOOL_ERASER) {
-        this->listener->toolSizeChanged();
+        this->stateChangeListener->toolSizeChanged();
     }
 }
 
@@ -220,7 +228,7 @@ void ToolHandler::setHighlighterSize(ToolSize size) {
     this->tools[TOOL_HIGHLIGHTER - TOOL_PEN]->setSize(size);
 
     if (this->activeTool->type == TOOL_HIGHLIGHTER) {
-        this->listener->toolSizeChanged();
+        this->stateChangeListener->toolSizeChanged();
     }
 }
 
@@ -228,7 +236,7 @@ void ToolHandler::setPenFillEnabled(bool fill, bool fireEvent) {
     this->tools[TOOL_PEN - TOOL_PEN]->setFill(fill);
 
     if (this->activeTool->type == TOOL_PEN && fireEvent) {
-        this->listener->toolFillChanged();
+        this->stateChangeListener->toolFillChanged();
     }
 }
 
@@ -242,7 +250,7 @@ void ToolHandler::setHighlighterFillEnabled(bool fill, bool fireEvent) {
     this->tools[TOOL_HIGHLIGHTER - TOOL_PEN]->setFill(fill);
 
     if (this->activeTool->type == TOOL_HIGHLIGHTER && fireEvent) {
-        this->listener->toolFillChanged();
+        this->stateChangeListener->toolFillChanged();
     }
 }
 
@@ -269,7 +277,7 @@ void ToolHandler::setSize(ToolSize size) {
 
     Tool* tool = this->toolbarSelectedTool;
     tool->setSize(clippedSize);
-    this->listener->toolSizeChanged();
+    this->stateChangeListener->toolSizeChanged();
 }
 
 void ToolHandler::setButtonSize(ToolSize size, Button button) {
@@ -279,12 +287,12 @@ void ToolHandler::setButtonSize(ToolSize size, Button button) {
 
     Tool* tool = getButtonTool(button);
     tool->setSize(clippedSize);
-    this->listener->toolSizeChanged();
+    this->stateChangeListener->toolSizeChanged();
 }
 
 void ToolHandler::setLineStyle(const LineStyle& style) {
     this->tools[TOOL_PEN - TOOL_PEN]->setLineStyle(style);
-    this->listener->toolLineStyleChanged();
+    this->stateChangeListener->toolLineStyleChanged();
 }
 
 void ToolHandler::setColor(Color color, bool userSelection) {
@@ -293,17 +301,17 @@ void ToolHandler::setColor(Color color, bool userSelection) {
     }
     Tool* tool = this->activeTool;
     tool->setColor(color);
-    this->listener->toolColorChanged();
+    this->stateChangeListener->toolColorChanged();
     if (userSelection)
-        this->listener->changeColorOfSelection();
-    this->listener->setCustomColorSelected();
+        this->stateChangeListener->changeColorOfSelection();
+    this->stateChangeListener->setCustomColorSelected();
 }
 
 void ToolHandler::setButtonColor(Color color, Button button) {
     Tool* tool = this->getButtonTool(button);
     tool->setColor(color);
-    this->listener->toolColorChanged();
-    this->listener->setCustomColorSelected();
+    this->stateChangeListener->toolColorChanged();
+    this->stateChangeListener->setCustomColorSelected();
 }
 
 auto ToolHandler::getColor() -> Color {
@@ -512,9 +520,9 @@ void ToolHandler::setSelectionEditTools(bool setColor, bool setSize, bool setFil
 
     if (this->activeTool->type == TOOL_SELECT_RECT || this->activeTool->type == TOOL_SELECT_REGION ||
         this->activeTool->type == TOOL_SELECT_OBJECT || this->activeTool->type == TOOL_PLAY_OBJECT) {
-        this->listener->toolColorChanged();
-        this->listener->toolSizeChanged();
-        this->listener->toolFillChanged();
+        this->stateChangeListener->toolColorChanged();
+        this->stateChangeListener->toolSizeChanged();
+        this->stateChangeListener->toolFillChanged();
         this->fireToolChanged();
     }
 }
--- a/src/control/ToolHandler.cpp
+++ b/src/control/ToolHandler.cpp
@@ -17,7 +17,6 @@ ToolListener::~ToolListener() = default;
 ToolHandler::ToolHandler(ToolListener* stateChangeListener, ActionHandler* actionHandler, Settings* settings) {
     this->settings = settings;
     initTools();
-    this->toolChangeListeners = std::vector<std::function<void(ToolType)> >();
     this->actionHandler = actionHandler;
 
     this->stateChangeListener = stateChangeListener;
@@ -169,15 +168,15 @@ void ToolHandler::selectTool(ToolType type) {
 }
 
 void ToolHandler::fireToolChanged() {
-    for (auto listener: this->toolChangeListeners) {
+    for (auto&& listener: this->toolChangeListeners) {
         listener(this->activeTool->type);
     }
 
     stateChangeListener->toolChanged();
 }
 
-void ToolHandler::addToolChangedListener(std::function<void(ToolType)> listener) {
-    toolChangeListeners.push_back(listener);
+void ToolHandler::addToolChangedListener(ToolChangedCallback listener) {
+    toolChangeListeners.emplace_back(std::move(listener));
 }
 
 auto ToolHandler::getTool(ToolType type) -> Tool& { return *(this->tools[type - TOOL_PEN]); }
--- a/src/control/ToolHandler.h
+++ b/src/control/ToolHandler.h
@@ -62,6 +62,7 @@ class ActionHandler;
 
 class ToolHandler {
 public:
+    using ToolChangedCallback = std::function<void(ToolType)>;
     ToolHandler(ToolListener* stateChangedListener, ActionHandler* actionHandler, Settings* settings);
     virtual ~ToolHandler();
 
@@ -226,7 +227,7 @@ public:
      * @param listener A callback, called when the user/client 
      *  changes tools.
      */
-    void addToolChangedListener(std::function<void(ToolType)> listener);
+    void addToolChangedListener(ToolChangedCallback listener);
 
     /**
      * @brief Get the Tool of a certain type
@@ -364,8 +365,7 @@ private:
     std::unique_ptr<Tool> mouseRightButtonTool;
     std::unique_ptr<Tool> touchDrawingButtonTool;
 
-    std::vector< std::function<void(ToolType)> > toolChangeListeners;
-
+    std::vector<ToolChangedCallback> toolChangeListeners;
 
     ToolListener* stateChangeListener = nullptr;
     ActionHandler* actionHandler = nullptr;
--- a/src/control/ToolHandler.h
+++ b/src/control/ToolHandler.h
@@ -12,6 +12,7 @@
 #pragma once
 
 #include <array>
+#include <functional>
 #include <memory>
 #include <string>
 #include <vector>
@@ -61,7 +62,7 @@ class ActionHandler;
 
 class ToolHandler {
 public:
-    ToolHandler(ToolListener* listener, ActionHandler* actionHandler, Settings* settings);
+    ToolHandler(ToolListener* stateChangedListener, ActionHandler* actionHandler, Settings* settings);
     virtual ~ToolHandler();
 
     /**
@@ -216,8 +217,19 @@ public:
      */
     void fireToolChanged();
 
+    /** 
+     * @brief Listen for tool changes.
+     * 
+     * Different from the listener given to the constructor -- [listener]
+     * here only listens for when the current tool is changed to another.
+     * 
+     * @param listener A callback, called when the user/client 
+     *  changes tools.
+     */
+    void addToolChangedListener(std::function<void(ToolType)> listener);
+
     /**
-     * @brief Get the Tool of a certain typ
+     * @brief Get the Tool of a certain type
      *
      * @param type
      * @return Tool&
@@ -352,7 +364,10 @@ private:
     std::unique_ptr<Tool> mouseRightButtonTool;
     std::unique_ptr<Tool> touchDrawingButtonTool;
 
-    ToolListener* listener = nullptr;
+    std::vector< std::function<void(ToolType)> > toolChangeListeners;
+
+
+    ToolListener* stateChangeListener = nullptr;
     ActionHandler* actionHandler = nullptr;
     Settings* settings = nullptr;
 };
--- a/src/control/ToolHandler.h
+++ b/src/control/ToolHandler.h
@@ -218,13 +218,13 @@ public:
      */
     void fireToolChanged();
 
-    /** 
+    /**
      * @brief Listen for tool changes.
-     * 
+     *
      * Different from the listener given to the constructor -- [listener]
      * here only listens for when the current tool is changed to another.
-     * 
-     * @param listener A callback, called when the user/client 
+     *
+     * @param listener A callback, called when the user/client
      *  changes tools.
      */
     void addToolChangedListener(ToolChangedCallback listener);
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -95,8 +95,9 @@ void ZoomControl::startZoomSequence(utl::Point<double> zoomCenter) {
     auto const& view_pos = utl::Point{rect.x, rect.y};
 
     this->zoomWidgetPos = zoomCenter - view_pos;
-    this->scrollPosition = (view_pos + this->zoomWidgetPos) / this->zoom;
     this->zoomSequenceStart = this->zoom;
+
+    setScrollPositionAfterZoom(view_pos);
 }
 
 void ZoomControl::zoomSequenceChange(double zoom, bool relative) {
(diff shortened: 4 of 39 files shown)


src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -119,8 +129,9 @@ void TouchInputHandler::zoomStart() {
     }
 
     Rectangle zoomSequenceRectangle = zoomControl->getVisibleRect();
+    auto topLeft = utl::Point<double>{zoomSequenceRectangle.x, zoomSequenceRectangle.y};
```
LittleHuba, line 132: Please merge this with `center` as it does not make sense to have this variable declared as center if it actually isn't.

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -142,13 +142,33 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
     return false;
 }
 
+double PenInputHandler::inferPressure(PositionInputData const& pos, XojPageView* page) {
```
LittleHuba, line 145: Two things with infering pressure. I generally like the idea behind it, as drawing quickly is often associated with less pressure. But not all users might like this as it could be understood as unintuitive. Hence I'd appreciate it if you could make this optional via a preference in the preferences dialog.

Additionally this feature is quite specific to the TouchDrawingInputHandler and should therefore reside in this class. If you add it to the PenInputHandler, you would block any new tool that might actually not want pressure by design.

LittleHuba (changes requested):
Hi!
Thanks for the work! This are actually some amazing additions. I could not actually test the features so far, but got around to at least give your code a once-over.

LittleHuba:
> What is the reason for the workaround that enables touch drawing? Is it reasonable to allow panning and zooming when the user uses multi-touch with this workaround enabled?

The workaround you can specifically select in the preferences is a feature of the old input system. The new system does not require nor use this feature. Instead it uses the list of devices where you can assign device classes. So if you assign a touchscreen the pen input class it will actually draw for touch inputs. This is interesting for tablets where you don't actually have a stylus to draw with. In that manner it would also be interesting to enable panning and zooming for multi-touch gestures. But it should be optional and not interfere with the response of the normal inputs. You'll have to do some magic for this where you treat the input initially as normal input and the invalidate the input when the second finger touches the screen.

You can text me on Gitter or Matrix if you want some more details or didn't understand what I tried to describe here.

src/gui/inputdevices/TouchInputHandler.cpp
personalizedrefrigerator: Done! Interestingly, however, this breaks zooming when touch drawing/the workaround is disabled.

For now, I change the center here only if the workaround/touch zooming is enabled.

src/gui/inputdevices/PenInputHandler.cpp
personalizedrefrigerator: I think it might make sense for pressure to be enabled for multiple input sources. For example, mice also lack pressure sensitivity.

I have added this to the preferences dialog.

personalizedrefrigerator:
I updated the preferences dialog! I, however, have not updated any of the translations...

personalizedrefrigerator:
@LittleHuba I think I've made the requested changes!

There are still some issues, however:
 * When touch drawing is enabled, writing outside the window doesn't work.
 * New settings in the SettingsDialog lack translations.

src/gui/MainWindow.cpp
```diff
@@ -282,15 +284,44 @@ void MainWindow::initXournalWidget() {
...
+void MainWindow::setGtkTouchscreenScrollingEnabled(bool enabled) {
+    if (enabled == gtkTouchscreenScrollingEnabled || usingTouchWorkaround || winXournal == nullptr) {
+        return;
+    }
+
+    gtkTouchscreenScrollingEnabled = enabled;
+
+    // Run on the main thread:
+    // 0 = timeout (run as soon as possible)
+    gdk_threads_add_timeout(
```
Febbe, line 310: **nitpick (non-blocking):** use `gdk_threads_add_idle*` here instead of `gdk_threads_add_timeout`

src/gui/MainWindow.cpp
```diff
@@ -282,15 +284,44 @@ void MainWindow::initXournalWidget() {
...
+            0,
+            [](gpointer data) -> gboolean {
+                MainWindow* self = static_cast<MainWindow*>(data);
+
+                gtk_scrolled_window_set_kinetic_scrolling(GTK_SCROLLED_WINDOW(self->winXournal),
+                                                          self->gtkTouchscreenScrollingEnabled);
+
+                return FALSE;
```
Febbe, line 318: Macros are discouraged since c++11 and mostly obsolete since c++17. Please use the in build types.

```suggestion
                return false;
```

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -118,12 +128,22 @@ void TouchInputHandler::zoomStart() {
...
+        center += utl::Point<double>{zoomSequenceRectangle.x, zoomSequenceRectangle.y};
+    }
 
     zoomControl->startZoomSequence(center);
 }
 
 void TouchInputHandler::zoomMotion(InputEvent const& event) {
+    if (!inputContext->getSettings()->isZoomGesturesEnabled()) {
+        return;
+    }
```
Febbe, line 146: **Discussion:** This looks like a workaround for me, isn't it. In my opinion this function or this InputHandler shouldn't be called in that case.

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -142,10 +162,13 @@ void TouchInputHandler::zoomMotion(InputEvent const& event) {
...
 
     zoomControl->setScrollPositionAfterZoom(offset);
-    lastZoomScrollCenter = center;
 }
 
 void TouchInputHandler::zoomEnd() {
+    if (!inputContext->getSettings()->isZoomGesturesEnabled()) {
+        return;
+    }
+
```
Febbe, line 171: **Discussion:** same here

src/control/PdfCache.cpp
```diff
@@ -46,53 +45,63 @@ void PdfCache::clearCache() {
...
+PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
+    while (this->data.size() >= this->size) {
         delete this->data.back();
         this->data.pop_back();
     }
+
+    auto* ne = new PdfCacheEntry(std::move(popplerPage), img, zoom);
+    this->data.push_front(ne);
+
+    return ne;
```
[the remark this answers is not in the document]
personalizedrefrigerator: The change here was to make `PdfCache::cache` return a `PdfCacheEntry*` and cache the `zoom`... The `new` isn't new!

The surrounding diff looks like this:
```
-  void PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img) {
-  auto* ne = new PdfCacheEntry(std::move(popplerPage), img);
-    this->data.push_front(ne);
-
-    while (this->data.size() > this->size) {
+ PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
+    while (this->data.size() >= this->size) {
+        delete this->data.back();
+        this->data.pop_back();
+    }
+
+    auto* ne = new PdfCacheEntry(std::move(popplerPage), img, zoom);
+    this->data.push_front(ne);
+
+    return ne;
```

I agree with you, though!!! Refactoring would be nice here!!!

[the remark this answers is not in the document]
personalizedrefrigerator: I'm resolving this issue. If you have further comments, feel free to un-resolve it!

src/gui/MainWindow.cpp
personalizedrefrigerator: I tried `gtk_threads_add_idle(`, but the priority it put calling the function at too low of a priority!

I'm looking into using `gtk_threads_add_idle_full` with a priority of `G_PRIORITY_HIGH_IDLE`!

personalizedrefrigerator: According to the GTK documentation, `gtk_threads_add_idle_full` has the following signature:
```
guint
gdk_threads_add_idle_full (gint priority,
                           GSourceFunc function,
                           gpointer data,
                           GDestroyNotify notify);
```

Personally, `gdk_threads_add_timeout` seems more readable here; however, I understand that this might not be a question of readability.

Which do you think is more readable/better?

src/gui/inputdevices/TouchInputHandler.cpp
personalizedrefrigerator: I was duplicating logic found at the beginning of `zoomStart`! 

I can change this!

src/gui/MainWindow.h
```diff
@@ -165,6 +167,9 @@ class MainWindow: public GladeGui, public LayerCtrlListener {
     GtkWidget* winXournal = nullptr;
     ScrollHandling* scrollHandling = nullptr;
 
+    bool usingTouchWorkaround;
+    bool gtkTouchscreenScrollingEnabled;
```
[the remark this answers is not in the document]
personalizedrefrigerator: According to its description in the settings dialog, the touchscreen workaround should only take effect when the application restarts. 

`usingTouchWorkaround` is cached at program run to provide this functionality -- `settings->isTouchWorkaround()` reflects the current state of the setting. It might make more sense for `settings->isTouchWorkaround()` to return the current **effective state** of `touchWorkaroundEnabled`. Unfortunately, the preferences dialog uses this to set the state of the touch workaround checkbox... 

Similarly, `gtkTouchscreenScrollingEnabled` is necessary as (I think) it isn't stored anywhere else (except GTK).

src/control/PdfCache.cpp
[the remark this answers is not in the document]
Febbe: No, I didn't see that, this should be resolved in another PR.
With that the `std::list<PdfCacheEntry*>` should also be replaced by `std::deque<PdfCacheEntry>`

Febbe:
@LittleHuba I may want to take a closer look at the zoom changes in `TouchInputHandler.cpp`, since I rewrote parts of the zooming handling. 
@personalizedrefrigerator that's the only reason, why I also made a small first review. Mostly nitpicks and suggestions. Your code looks very well and I like your pressure inference feature. Keep going :).

src/gui/inputdevices/TouchInputHandler.cpp
personalizedrefrigerator: Changed!

src/gui/MainWindow.cpp
Febbe: It is a question of understandability and this is a subset of readability. `gtk_threads_add_idle` looks more like adding a task to the scheduler. A timeout looks like, that the code must be executed after a specific amount of time with a specific frequency. But also with your above comment this should be the same.
I leave that to you and @LittleHuba.

Regarding my opinion: `gtk_threads_add_idle` is more readable, since it also tells the reader which priority must be used. Also, the `GDestroyNotify` can be a `nullptr` as I know. This would only add one parameter and will replace the magic number `0` with a named constant.

personalizedrefrigerator: I found `Util::execInUiThread`! I'm adding a `priority` parameter and using that!

LittleHuba:
@Febbe of course go ahead with that part. You most likely know that system way better than me.

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -142,13 +143,33 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
...
 
+double PenInputHandler::inferPressureIfEnabled(PositionInputData const& pos, XojPageView* page) {
+    // Infer pressure
+    if (pos.pressure == Point::NO_PRESSURE && this->inputContext->getSettings()->isPressureGuessingEnabled()) {
+        PositionInputData lastPos = getInputDataRelativeToCurrentPage(page, this->lastEvent);
+
+        double dt = std::min((pos.timestamp - lastPos.timestamp) / 10.0, 2.0);
+        double inverseSpeed = (dt) / std::sqrt(std::pow(pos.x - lastPos.x, 2) + std::pow(pos.y - lastPos.y, 2) + 0.001);
+
+        double newPressure = 3.142 / 2.0 + std::atan(inverseSpeed * 3.14 - 1.3);
```
LittleHuba, line 154: You are using very expensive methods here (pow, sqrt, atan) within a function that is called very very often. It would be good to simplify this calculation to a quick approximation. The result does not need to be perfect as it has no huge influence on the look if the pressure is 0.00000002 or 0.00000003.

src/gui/MainWindow.cpp
LittleHuba: `Util::execInUiThread` is actually meant for exactly this usecase.

LittleHuba (changes requested):
Nice work so far. What I'm missing in general are comments. I know they are a huge bother to write every time but it just makes the work for me and the next one working on this code so much easier.

src/gui/inputdevices/PenInputHandler.cpp
Febbe: `std::pow(x,2)` can be replaced by x*x, which is much faster. Unfortunately there is no constexpr version of `<cmath>` functions. Then it wouldn't make any difference.
Also `<cmath>` supports the [`std::hypot`](https://en.cppreference.com/w/cpp/numeric/math/hypot) function, which is optimized for calculating the 2nd norm. If you require precision instead of a numeric approximation, it is still a way faster than a manual `sqrt(pow(x,2) + pow(y,2))` (in this case just use `utl::Point::distance` or `Point::lineLengthTo).

personalizedrefrigerator:
@Febbe Most of the changes to `TouchInputHandler.cpp` were an attempt to fix view translation while scrolling large documents:

out

I think this is because the distance between pages is constant, even while zooming. As such, zooming out increases the **relative** distance between pages. Neither `ZoomControl` nor `Layout` take this into account. 

As such, zooming does not work expected near the end of many-page documents.

src/control/PdfCache.cpp
```diff
@@ -46,53 +45,63 @@ void PdfCache::clearCache() {
...
     return nullptr;
 }
 
-void PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img) {
-    auto* ne = new PdfCacheEntry(std::move(popplerPage), img);
-    this->data.push_front(ne);
-
-    while (this->data.size() > this->size) {
+PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom) {
+    while (this->data.size() >= this->size) {
```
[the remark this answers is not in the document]
personalizedrefrigerator: Oops! I must have changed this while debugging! I can change it back!

src/control/ToolHandler.cpp
```diff
@@ -152,9 +154,15 @@ void ToolHandler::selectTool(ToolType type, bool fireToolChanged, bool triggered
 }
 
 void ToolHandler::fireToolChanged() {
-    if (listener) {
-        listener->toolChanged();
+    for (auto listener: this->toolChangeListeners) {
```
Febbe, line 157: This copies all listeners in `this->toolChangeListeners`. Was this intentional?
Better use:
```suggestion
    for (auto&& listener: this->toolChangeListeners) {
```

personalizedrefrigerator: No!!!! Thank you!!!!

src/control/ToolHandler.cpp
```diff
@@ -152,9 +154,15 @@ void ToolHandler::selectTool(ToolType type, bool fireToolChanged, bool triggered
...
+        listener(this->currentTool->type);
     }
+
+    stateChangeListener->toolChanged();
+}
+
+void ToolHandler::addToolChangedListener(std::function<void(ToolType)> listener) {
+    toolChangeListeners.push_back(listener);
```
Febbe, line 165:
```suggestion
    toolChangeListeners.emplace_back(std::move(listener));
```

src/control/PdfCache.cpp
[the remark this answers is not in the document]
personalizedrefrigerator: Reverted!

src/control/ToolHandler.h
```diff
@@ -41,7 +42,7 @@ class ActionHandler;
 
 class ToolHandler {
 public:
-    ToolHandler(ToolListener* listener, ActionHandler* actionHandler, Settings* settings);
+    ToolHandler(ToolListener* stateChangedListener, ActionHandler* actionHandler, Settings* settings);
```
Febbe, line 45:
```suggestion
    using ToolChangedCallback = std::function<void(ToolType)>;
    ToolHandler(ToolListener* stateChangedListener, ActionHandler* actionHandler, Settings* settings);
```

src/control/ToolHandler.h
```diff
@@ -93,6 +94,10 @@ class ToolHandler {
     ToolType getToolType(ToolPointer toolpointer = ToolPointer::current);
     void fireToolChanged();
 
+    // Different from the listener given to the constructor -- [listener]
+    // here only listens for when the current tool is changed to another.
+    void addToolChangedListener(std::function<void(ToolType)> listener);
```
Febbe, line 99:
```suggestion
    void addToolChangedListener(ToolChangedCallback listener);
```

src/control/ToolHandler.h
```diff
@@ -149,7 +154,8 @@ class ToolHandler {
      * not in the list, so its a "custom" color for us
      */
 
-    ToolListener* listener = nullptr;
+    std::vector<std::function<void(ToolType)> > toolChangeListeners;
```
Febbe, line 157:
```suggestion
    std::vector<ToolChangedCallback> toolChangeListeners;
```

src/gui/MainWindow.cpp
personalizedrefrigerator: Changed!

src/gui/inputdevices/TouchDrawingInputHandler.cpp
```diff
@@ -96,5 +156,18 @@ auto TouchDrawingInputHandler::changeTool(InputEvent const& event) -> bool {
...
+void TouchDrawingInputHandler::updateKineticScrollingEnabled() {
+    auto* control = inputContext->getView()->getControl();
+    auto* mainWindow = control->getWindow();
+    auto* toolHandler = this->inputContext->getToolHandler();
+
+    // Kinetic scrolling is nice; however, we need to disable it so we can draw.
+    if (mainWindow != nullptr && control->getSettings()->getTouchDrawingEnabled()) {
+        mainWindow->setGtkTouchscreenScrollingEnabled(toolHandler->getToolType() == TOOL_HAND);
+    }
+}
```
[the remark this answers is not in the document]
personalizedrefrigerator: Fixed!

LittleHuba:
> @LittleHuba I think I've made the requested changes!
> 
> There are still some issues, however:
> 
> * When touch drawing is enabled, writing outside the window doesn't work. **Edit:** This has been fixed.
> * New settings in the SettingsDialog lack translations.
> * (**Edit**) `Recent Documents` isn't working. **This might have been broken before. Please test in `xournalpp:master` before merging.**

The missing translations will be fixed by the community in time. If there exists no translation, the english tag is used. People can introduce new translations easily using our Crowdin project.

Recent documents was bugged some time ago. I'm not quite sure if the fix is already merged. But that should not be of concern for this PR as it is a completely different topic.

src/control/ToolHandler.cpp
```diff
@@ -12,11 +12,13 @@
 
 ToolListener::~ToolListener() = default;
 
-ToolHandler::ToolHandler(ToolListener* listener, ActionHandler* actionHandler, Settings* settings) {
+ToolHandler::ToolHandler(ToolListener* stateChangeListener, ActionHandler* actionHandler, Settings* settings) {
     this->settings = settings;
     initTools();
-    this->listener = listener;
+    this->toolChangeListeners = std::vector<std::function<void(ToolType)> >();
```
Febbe, line 18:
```suggestion
```

src/control/ToolHandler.cpp
```diff
@@ -152,9 +154,15 @@ void ToolHandler::selectTool(ToolType type, bool fireToolChanged, bool triggered
...
-    if (listener) {
-        listener->toolChanged();
+    for (auto&& listener: this->toolChangeListeners) {
+        listener(this->currentTool->type);
     }
+
+    stateChangeListener->toolChanged();
+}
+
+void ToolHandler::addToolChangedListener(std::function<void(ToolType)> listener) {
```
Febbe, line 164:
```suggestion
void ToolHandler::addToolChangedListener(ToolChangedCallback listener) {
```

LittleHuba:
> @Febbe Most of the changes to `TouchInputHandler.cpp` were an **attempt** to fix view translation while scrolling large documents:
> 
> out
> 
> I think this is because the distance between pages is constant, even while zooming. As such, zooming out increases the **relative** distance between pages. Neither `ZoomControl` nor `Layout` take this into account.
> 
> As such, zooming does not work expected near the end of many-page documents.
> 
> ### This is still an issue.

Would you be interested in doing a follow-up PR that fixes this issue?

src/gui/inputdevices/PenInputHandler.cpp
personalizedrefrigerator: `pos` and `lastPos` are of type `PositionInputData`. As such, `utl::Point::distance` does not apply without a conversion.

I am considering making `PositionInputData` extend `utl::Point<double>`...

personalizedrefrigerator:
> 
> Would you be interested in doing a follow-up PR that fixes this issue?

Yes. This issue has been bothering me. I can try to fix it.

src/gui/inputdevices/PenInputHandler.cpp
LittleHuba: I think extending `util::Point<double>` would be a valid solution.

personalizedrefrigerator: I'm still not sure what to do about `std::atan`...

LittleHuba:
We'll either have to squash the commits before merging the PR or you do some rebasing-magic to clean up your commits. Currently it is a lot of commits just fixing previous errors.

src/gui/inputdevices/PenInputHandler.cpp
Febbe: You could use a local object of type utl::Point<double>. Most likely this will completely inlined by the compiler :).

personalizedrefrigerator: Extending `utl::Point<double>` makes the code much cleaner. I think I'll do that for now...

personalizedrefrigerator:
> We'll either have to squash the commits before merging the PR or you do some rebasing-magic to clean up your commits. Currently it is a lot of commits just fixing previous errors.

I'm working on this!

src/gui/dialog/SettingsDialog.h
```diff
@@ -35,9 +35,15 @@ class SettingsDialog: public GladeGui {
     void setDpi(int dpi);
 
     /**
-     * Autosave was toggled, enable / disable autosave config
+     * Set active regions
      */
     void enableWithCheckbox(const string& checkbox, const string& widget);
+    void disableWithCheckbox(const string& checkbox, const string& widget);
+
+    /**
```
Febbe, line 43: **Nitpick (non-blocking):** I would not use `/**`, since this marks the comment as documentation for the next function and will be associated with it. Since the content is a description for the whole block, we do not want this.

```suggestion
    /*
```

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -142,13 +143,33 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
     return false;
 }
 
+double PenInputHandler::inferPressureIfEnabled(PositionInputData const& pos, XojPageView* page) {
+    // Infer pressure
+    if (pos.pressure == Point::NO_PRESSURE && this->inputContext->getSettings()->isPressureGuessingEnabled()) {
+        PositionInputData lastPos = getInputDataRelativeToCurrentPage(page, this->lastEvent);
+
+        double dt = std::min((pos.timestamp - lastPos.timestamp) / 10.0, 2.0);
+        double inverseSpeed = (dt) / (pos.distance(lastPos) + 0.001);
```
Febbe, line 152:
```suggestion
        double inverseSpeed = (dt) / (utl::Point<double>(pos.x,pos.y).distance(utl::Point<double>(lastPos.x,lastPos.y)) + 0.001);
```

src/gui/inputdevices/PositionInputData.h
```diff
@@ -14,7 +14,9 @@
 
 #include <gtk/gtk.h>
 
-class PositionInputData {
+#include "util/Point.h"
+
+class PositionInputData: public utl::Point<double> {
```
Febbe, line 19: Do not use inheritance here:
 1. `utl::Point<>` does not provide a virtual destructor, which will introduce unnecessarily a v-table for such a trivial struct.
 2. `PositionInputData` is not a Point it contains one. So make `utl::Point<double>` a member with the name `pos` or `position`. A doc is not required, since this is self explaining.

po/README.md
```diff
@@ -0,0 +1,3 @@
+Please use [Crowdin](https://crowdin.com/project/xournalpp) to create or edit translations. 
+Should a language not be available, you can reach out to the developers by opening a new 
+issue or Gitter/Matrix (see links in main README).
```
LittleHuba, line 3: Looks like one of your rebases was not correct. This change is from a different PR and should not be part of your changes.

personalizedrefrigerator: I merged changes to `xournalpp:master` into my branch partway through work on my PR. I probably shouldn't have done this.

image

These changes probably come from the commits labeled "**LittleHubba** authored and **personalizedrefrigerator committed**".

personalizedrefrigerator: I may have to fork `xournalpp:master` again and apply my changes to the fork... Should I do this and open a new PR?

LittleHuba: Normally you do not merge both branches but rebase on top of master.
You can use the git reflog to undo the merge and then do the rebase.

personalizedrefrigerator: Okay. Thank you for catching this!

LittleHuba: If using the reflog does not work, just cherry pick your changes to another branch locally and force push to your PR branch.

Febbe: Yes instead of using `git merge upstream:master` one should ever use `git rebase upstream:master` this will keep the history clean and you won't mess with foreign commits.

personalizedrefrigerator:
As it relates to the topic of this PR, I have included a **partial** fix for the moving-viewport-while-zooming issue (I think issue #2411). See [this commit](https://github.com/xournalpp/xournalpp/pull/2435/commits/550375a3bb71c6f3b052be7c5ab251154608117d).

src/control/zoom/ZoomControl.cpp
```diff
@@ -78,15 +79,26 @@ auto ZoomControl::getVisibleRect() -> Rectangle<double> {
...
+    this->scrollPosition = (scrollPos - this->unscaledPixels + this->zoomWidgetPos) / this->zoom;
 }
 
 auto ZoomControl::getScrollPositionAfterZoom() const -> utl::Point<double> {
     if (this->zoomSequenceStart == -1) {
         return {-1, -1};
     }
 
-    return (this->scrollPosition * this->zoom) - this->zoomWidgetPos;
+    return (this->scrollPosition * this->zoom) - this->zoomWidgetPos + this->unscaledPixels;
```
LittleHuba, line 101: Not a very good idea to use the variable for this. This forces the call of `void ZoomControl::setScrollPositionAfterZoom(utl::Point<double> scrollPos)` before any call to `ZoomControl::getScrollPositionAfterZoom()` may happen or it returns the wrong result. As getPaddingAbovePage is also const you should just call the method here as well.

src/gui/sidebar/previews/base/SidebarPreviewBase.cpp
```diff
@@ -128,6 +128,6 @@ auto SidebarPreviewBase::scrollToPreview(SidebarPreviewBase* sidebar) -> bool {
     return false;
 }
 
-void SidebarPreviewBase::pageDeleted(int page) {}
+void SidebarPreviewBase::pageDeleted(size_t page) {}
```
LittleHuba, line 131: For those really large documents with more than 2147483647 pages? 😉 
But leave it there, it doesn't hurt.

personalizedrefrigerator: I was getting a warning in the mac build CI check :)

src/control/zoom/ZoomControl.cpp
personalizedrefrigerator: Notice that `ZoomControl::startZoomSequence` calls `ZoomControl::setScrollPositionAfterZoom`! 

Of course, I can still make this change -- it gets rid of an unnecessary member variable ☺.

personalizedrefrigerator: `ZoomControl::startZoomSequence` calls `ZoomControl::setScrollPositionAfterZoom`. To call `getPaddingAbovePage`, we also need to call `XournalView::getWidget()`, which is not `const` ☺.

personalizedrefrigerator: I should document this!

personalizedrefrigerator: Done!

I should note that this change is only a partial fix. While the zooming in/out very far into a large document no longer causes the viewport to scroll significantly, it does cause the viewport to rapidly move up and down small amounts.

For anyone working on this in the future: Having set `XOURNAL_PADDING_BETWEEN` and tested before making these changes, I saw that this behavior occurs without this new partial fix.

LittleHuba (approved):
This seems to be the best fix for now. In the long run we should improve on this. But as we need to revamp the rendering system anyway it does not make sense to put more work in this now.

I'll do a functional test in the next few days and given everything works we can merge this.

personalizedrefrigerator:
image

The "Pressure Inference" frame lacks a border. This stylistic fix really should be a part of commit 1450a7c923b69055de8f0c677b06b2aec8e4e752. As such, I'm making the change and force-pushing the modified history.

LittleHuba:
> image
> 
> The "Pressure Inference" frame lacks a border. This stylistic fix really should be a part of commit 1450a7c923b69055de8f0c677b06b2aec8e4e752 (and only involves changes to a few lines of `settings.glade`). As such, I'm making the change and force-pushing the modified history.

This should actually be part of the previous frame.

personalizedrefrigerator:
Already changed!

personalizedrefrigerator:
This is what the changed parts of the settings dialog currently looks like:
image
image
image
image

LittleHuba:
> image

This should be part of Drawing Area as it is part of the rendering system and has not really anything to do with loading or saving a file.

personalizedrefrigerator:
> This should be part of Drawing Area as it is part of the rendering system and has not really anything to do with loading or saving a file.

This has been changed!

personalizedrefrigerator:
image
image

personalizedrefrigerator:
@Febbe @eliasrg  Can you check whether the same issue fixed by c1edabf2ccaa64606cacf2e29790d7425978ec3c is also fixed here? 
In this PR, multi-touch gestures are used even when touch-gesture zooming is disabled. As such, I would prefer to not ignore all multi-touch gestures when touch-gesture zooming is disabled.

eliasrg:
@personalizedrefrigerator Yes, #2509 is also fixed here. In fact, with zoom gestures and touch drawing turned off, it seems to work exactly like c1edabf (just ignoring the second finger). So in principle, I have no objection at all!

However, I had a look at your `TouchInputHandler.cpp` and I think I found a small issue with it. I will make a separate comment.

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -36,22 +38,27 @@ auto TouchInputHandler::handleImpl(InputEvent const& event) -> bool {
...
 
         if (event.sequence == this->primarySequence) {
-            this->primarySequence = nullptr;
+            // If secondarySequence is nullptr, this sets primarySequence
+            // to nullptr. If it isn't, then it is now the primary sequence!
+            this->primarySequence = this->secondarySequence;
+            this->secondarySequence = nullptr;
+
+            this->priLastAbs = this->secLastAbs;
+            this->priLastRel = this->secLastRel;
```
eliasrg, line 61: If I understand this part correctly, I should, if zoom gestures and touch drawing are disabled, be able to scroll with one finger, put a second finger on the screen, release the first finger and continue scrolling with the second finger. But this code doesn't accomplish that, because `secLastAbs` and `secLastRel` are only kept up to date if `zoomGesturesEnabled` is true. I wrote a quick fix for this: https://github.com/eliasrg/xournalpp/commit/b9c50aab62ce58d157b5960f8d9c23dd1f98e7c4 and it does work when I test it. However, this logic is getting a bit confusing/unnatural, so it might be better to restructure the file more drastically...

personalizedrefrigerator: I think `TouchInputHandler::scrollMotion` should already be keeping these up to date:
```cpp
void TouchInputHandler::scrollMotion(InputEvent const& event) {
    // Will only be called if there is a single sequence (zooming handles two sequences)
    auto offset = [&]() {
        auto absolutePoint = utl::Point{event.absoluteX, event.absoluteY};
        if (event.sequence == this->primarySequence) {
            auto offset = absolutePoint - this->priLastAbs;
            this->priLastAbs = absolutePoint;
            return offset;
        } else {
            auto offset = absolutePoint - this->secLastAbs;
            this->secLastAbs = absolutePoint;
            return offset;
        }
    }();
```

I might be misunderstanding this...

eliasrg: But `scrollMotion` is not called for the second finger if there are two fingers on the screen, only for the one actually scrolling.

I think you can see the difference if you compile both versions and try switching fingers while scrolling (zoom gestures and touch drawing turned off)?

personalizedrefrigerator: Yes! With touch-zooming off, I am hitting the issue! Thank you for catching this!

I agree with you about re-structuring the file -- it is getting rather difficult to understand what is going on.

LittleHuba:
After merging #2339 we got some conflicting files. Could you clean those up? I'd like to merge your PR as one of the next few so it will be included in the next release.

src/gui/inputdevices/TouchInputHandler.cpp
eliasrg: Okay, I tried to restructure this in a nice way, but I encountered strange bugs, and eventually I realised that I had misunderstood how the touch input works. Because it doesn't happen only in this file. If you comment out the whole of `handleImpl` except `return false`, you can still scroll with one finger without problem. But as soon as you put a second finger on the screen, it stops working. This is because there is a [`GtkScrolledWindow`](https://developer.gnome.org/gtk3/stable/GtkScrolledWindow.html) which has built-in support for one-finger touches, and even provides kinetic scrolling (which the code in `scrollMotion` doesn't, as you can check by scrolling after switching fingers in my "quick fix" https://github.com/eliasrg/xournalpp/commit/b9c50aab62ce58d157b5960f8d9c23dd1f98e7c4).

So I don't know why there are two competing systems for scrolling... it seems like a bad idea unless I am missing something. Can anybody explain the reason for this? My feeling is that we should either only handle zooming here and leave scrolling for the `GtkScrolledWindow` (probably the best), or get rid of the scrolled window and handle _all_ scrolling ourselves (more work and we wouldn't have kinetic scrolling, but it'd play nicer with zooming and multitouch).

Does that make sense?

personalizedrefrigerator: Thank you for looking into this!
The "enable touch workaround" setting switches between the two scrolling methods.

personalizedrefrigerator: It seems like a proper fix for the transition from two-finger panning to one-finger scrolling will require some refactoring... 
Do you think it could be appropriate to make this fix the subject of a follow-up PR?

LittleHuba: > So I don't know why there are two competing systems for scrolling... it seems like a bad idea unless I am missing something. Can anybody explain the reason for this?

@eliasrg This is actually a fix for a very old problem that was ported to the new input system. When you enable touch drawing (the touch workaround) you were essentially unable to scroll at all. Back in the day this was fixed with this convoluted system of handling scrolling manually in this case. As you mentioned GTK actually would handle this for us but only for "single finger touches" which are used for drawing in the workaround.

I see two solutions here:
1. We stick with this convoluted system and fix it as well as possible. This would keep our code somewhat smaller and leave stuff like kinetic scrolling with GTK. Arguably it would make the input system harder to understand but also minimize potential areas where we need to fix stuff.
2. We throw out GTK scroll handling completely and implement stuff like kinetic scrolling in our own handlers. I would then make it optional if scrolling should be done with two fingers or one (one being default as it is consistent with GTK). This would increase our code size but make the system straight forward.

@Febbe, @idotobi, @rolandlo, @Technius, @personalizedrefrigerator and @eliasrg what is your opinion on this? Lets make a poll with reaction smileys (🎉 for 1 and 🚀 for 2)

eliasrg: @LittleHuba If we are okay with not having two-finger scrolling (personally, I don't need it, but if you are writing with your finger you probably do), then I'd say let GTK do all scroll handling and only handle zoom ourselves. If we do need two-finger scrolling, it seems we've outgrown GTK's scrolling and should throw it out! Since we already have a working scroll handler (which I am working on simplifying), the most difficult part to implement would probably be kinetic scrolling, if indeed we need that. We'd also get a slightly nicer feel overall, with smooth transitions between scrolling and zooming.

LittleHuba: Okay the verdict is quite clear. Well go forward with option two. As this PR is already big enough I'd recommend a follow-up PR for this. @personalizedrefrigerator and @eliasrg would you like to join up and work on this?

LittleHuba:
With the scrolling question in the comment handled I assume this is good to merge @personalizedrefrigerator? Or do you have anything else you want to fix before doing so?

personalizedrefrigerator:
@LittleHuba I think this is good to merge. However, to be clear, I have not applied [eliasrg@b9c50aa](https://github.com/eliasrg/xournalpp/commit/b9c50aab62ce58d157b5960f8d9c23dd1f98e7c4). Should I do this before merging? Should this be a part of the follow-up PR?

LittleHuba:
As this is essentially a bug in your code the fix should be included. Even for the small timeframe it will actually be active.

personalizedrefrigerator:
@LittleHuba Applied!

LittleHuba:
I'll go through the code once more tonight and make the functional test. If nothing major pops up I'll merge it then.

personalizedrefrigerator:
Testing the PR in KDE Plasma, Wayland, I see that the `setGtkTouchscreenScrollingEnabled`'s `Util::execInUiThread` with `G_PRIORITY_HIGH` is not called soon enough after a two-finger pan for the next stroke not to scroll. In other words, two-finger panning while drawing can cause single-finger gestures immediately after to scroll the view.

LittleHuba:
> Testing the PR in KDE Plasma, Wayland, I see that the `setGtkTouchscreenScrollingEnabled`'s `Util::execInUiThread` with `G_PRIORITY_HIGH` is not called soon enough after a two-finger pan for the next stroke not to scroll (at least, I think this is what's happening...). As such, two-finger panning while drawing sometimes causes the following single-finger gesture to scroll, rather than draw.
> Enabling the touch workaround fixes this, but this requires action by the user.

I was unable to reproduce this. Are you sure your finger actually leaves the screen inbetween? You can take a look at the event stream from the debug output to see if the events are sent.

The whole scrolling part seems to work fine. One problem I see is with the PDF caching. You should at least redraw once the zooming has stopped. Currently you only trigger redraws if a certain zoom delta is reached. This delta is chosen too high and leads to bad renders. Keep this logic but decrease the delta (or make it modifiable by the user) and then add a fixed redraw once zooming is finished and the delta is higher than some second very small delta. This way the user might have bad renders inbetween but a good one when he annotates again.

personalizedrefrigerator:
> You should at least redraw once the zooming has stopped. Currently you only trigger redraws if a certain zoom delta is reached. This delta is chosen too high and leads to bad renders.

Thank you for catching this!

personalizedrefrigerator:
@LittleHuba I've added a setting that controls how often pages are re-rendered. 

I think re-rendering the current page in the PDF when done zooming will require some refactoring. I haven't done this yet!

## Changed parts of the settings dialog
image
image

LittleHuba:
Thanks for the update! I'll retest the behavior tomorrow and get back to you with some feedback if necessary!

personalizedrefrigerator:
> I think re-rendering the current page in the PDF when done zooming will require some refactoring. I haven't done this yet!

I've been using the feature that limits PDF re-rendering (to the cached image) to edit large PDF files (~10 MB over 2 or 3 pages), for which a re-render can take a few seconds when zoomed in. Would leaving this setting as is be reasonable?

An alternative might be to replace it with a setting that prevents re-renders when the user has zoomed in more than a specified amount -- a greater zoom causes the PDFCache to render to a larger surface and therefore takes longer to render.

Febbe:
@LittleHuba any updates on this?

LittleHuba:
Sorry not yet. Currently busy with other stuff. I'll return to this PR shortly though.

Febbe:
The parts I had my fingers in, look very good. 
LGTM from me.

personalizedrefrigerator:
The most recent changes apply pressure inference at the start and end of strokes, addressing https://github.com/xournalpp/xournalpp/pull/2512#issuecomment-749090044.

LittleHuba:
Your changes have made a huge impact on zooming behaviour! There are two small things I'd like to change though.

With the current default value (1%) and the 5% steps you end up at 6%, 11% which is ugly. I do like the 5% steps though as they are crude enough for the normal user. People interested in fine adjustments can still type the value.

Instead I'd like the default value to be 5% as this keeps backgrounds sharp but reduces redrawing of the background by factor 5.

Once this is changed this PR should be good to go! Thanks for the good work!

personalizedrefrigerator:
Thank you for the feedback! I've changed this!

rolandlo:
I'm sorry, but with the latest merge of #2592  we have created some merge conflicts in the settings file. Could you please solve them?

personalizedrefrigerator:
@rolando I'm working on it!

## Additional changes

> Testing the PR in KDE Plasma, Wayland, I see that the `setGtkTouchscreenScrollingEnabled`'s `Util::execInUiThread` with `G_PRIORITY_HIGH` is not called soon enough after a two-finger pan for the next stroke not to scroll (at least, I think this is what's happening...).

I was able to reproduce this on GNOME Wayland. It was due to a mistake in `TouchInputHandler.cpp` (not properly setting `primarySequence` on touch end during multi-touch). I have fixed this.

In changing this, I also changed `gtkTouchscreenScrollingEnabled` to an `std::atomic_bool`, as it has potential to be passed between threads:
```c++
// Line 295
void MainWindow::setGtkTouchscreenScrollingEnabled(bool enabled) {
    if (enabled == gtkTouchscreenScrollingEnabled.load() || usingTouchWorkaround || winXournal == nullptr) {
        return;
    }

    gtkTouchscreenScrollingEnabled.store(enabled);

    Util::execInUiThread(
            [=]() {
                gtk_scrolled_window_set_kinetic_scrolling(GTK_SCROLLED_WINDOW(winXournal),
                                                          gtkTouchscreenScrollingEnabled.load());
            },
            G_PRIORITY_HIGH);
}
```

As such, `MainWindow.h` now includes `atomic`:
```c++
// Line 14
#include <atomic>
```

Febbe:
grafik
Both points require the restart also the point obviously requires also the touch workaround to be disabled not only the new input system.

Febbe:
> Zooming can be very slow when editing large PDFs.
>
>    Only partially fixes this issue

This may be fixed by #2577 or #2541

personalizedrefrigerator:
> ~Both points require the restart~ also the point obviously requires also the touch workaround to be disabled not only the new input system.

I'm confused... With the new input system and the workaround enabled, touch drawing should work. I think there's something I'm misunderstanding!

LittleHuba:
> > ~Both points require the restart~ also the point obviously requires also the touch workaround to be disabled not only the new input system.
> 
> I'm confused... With the new input system and the workaround enabled, touch drawing should work. I think there's something I'm misunderstanding!

You got it right @personalizedrefrigerator. Only the new input system requires the restart. Once it is enabled, switching this on and off functions without restart. In the long run we should remove these warnings from the descriptions and simply show the user a warning at the bottom of the settings dialog as soon as he has changed a setting that requires a restart.

@Febbe if you do not have anything additional I'd go ahead and merge this.

rolandlo:
@LittleHuba Does this PR need some squashing before it gets merged? It still has 12 commits.

LittleHuba:
Best would be to clean up some of the later commits by doing a Git history rewrite and hunk-based commits. That requires some higher knowledge of Git though and some planning in advance to know what change goes to what commit. Especially the fixup commits should be merged with the earlier ones.

If that is to much to request, we should just squash and merge it.

personalizedrefrigerator:
@LittleHuba I would rather have the commits squashed.

Febbe:
> I'm confused... With the new input system and the workaround enabled, touch drawing should work. I think there's something I'm misunderstanding!

The thing is, that the checkbox is checked, disabled but visible, when the Gtk-Touch-Workaround is checked. Only unchecking the Gtk-Touch-Workaround activates the checkbox.

personalizedrefrigerator:
> Only unchecking the Gtk-Touch-Workaround activates the checkbox. So the new input system is required to be checked and Gtk-Touch-Workaround must be unchecked. But that's not mentioned there.

The touch drawing setting has no effect on the old input system. The workaround, however, does. 

I'm not sure how to make this relationship clearer in the settings dialog...

LittleHuba:
@Febbe could you provide some information on how to improve this?

Febbe:
Maybe add a hint when the checkbox is disabled, that the Touch Workaround needs to be disabled to modify this option.

personalizedrefrigerator:
## Old input system enabled, workaround enabled

image

## Old input system enabled, workaround disabled

image

## Old input system disabled, workaround enabled

image

## Old input system disabled, workaround disabled

image

Febbe:
Awesome 😊!

LittleHuba:
Thanks @personalizedrefrigerator for your hard work! I'm looking forward to future work of yours!

personalizedrefrigerator:
This was a rather large PR! Thank you for taking the time to review it!

Issue #2373, opened 2020-11-07
export layers in different PDF pages


sanette:
I'm starting using Xournalpp for a seminar presentation.
I'm using the layers to progressively add new content to a page, this works nicely by
* make sure initially the page has only "Layer 1" activated (and background of course)
* use CTRL-SHIFT-RIGHT ('next layer') to activate next layer
* when everyhing has been shown, I go to next page and start again.

My request is: it would be nice that PDF export could save one page per layer, so that when we display the PDF page after page, this has the same effect as what is described above. This way, the presentation could be used on a computer that doesn't have xournalpp installed.

fujtajlfik:
I would also appreciate this as a possibility for exporting PFD.

sanette:
I have implemented this (on my local copy). I like it, I think it will make me use xournalpp more often for presentations!
My question now: how do we choose between this PDF export mode and the usual PDF export mode?
1. add a flag in the Preferences
2. add another choice in the File menu (like 'Export Layers to PDF')
3. ask with a popup when selecting 'Export as PDF' (something like 'Do you want to export all layers as separate pages?')

sanette:
the code is here: https://github.com/sanette/xournalpp

sanette:
Here is an example:

[2020-12-23-Note-09-06.pdf](https://github.com/xournalpp/xournalpp/files/5734090/2020-12-23-Note-09-06.pdf)

rolandlo:
> EDIT: I have chosen option 4. You have to use "save as..." then select PDF, and the dialog will let you check a box "export all layers as separate pages"

Could you also add a command-line option?

sanette:
> Could you also add a command-line option?

ok, done!
`xournalpp -p output.pdf --export-layers file.xopp`

rolandlo:
Great! Please make a Pull request to the master branch (after you have pushed your latest commit including the command-line option).

Technius:
Fixed in #2589.

ventu06:
When writing a presentation, I noticed that sometimes I need more flexibility, for example, having layers that only appear in certain slides but not in subsequent ones.
So I wrote a small plugin that uses the layer names to decide which slides they should appear in and which they should not. You can find it [here](https://github.com/ventu06/xournalpp-plugins)!

PR #2449, opened 2020-11-27
Add command line export options

bhennion:
Hey,
Here is a first commit implementing a --export-range=... option for the command line. It works as is but I have a question.

As this new option (obiously) uses the same backend as the ExportDialog, I also fixed a bug (not listed) making the ExportDialog fail to understand a page range of the form "n-" (from page n until the end). Now, there are two ways of handling such a page range:
1) Set the end of the range to -1 and implement this possibility in the export job
2) Set the end of the range to the number of the last page **when we parse the string**

I chose option 2 (it was simpler to implement, and I don't yet see a possible benefit of option 1). However, andreasb123, who implemented PageRange::parse (in [this commit](https://github.com/xournalpp/xournalpp/commit/2d412eec34f5cb2f21bcb29db9d1d107dcc276c3#diff-56f37a5274511463372c1a4502e7cd937a822676995d9fc560766855ee3ef0f8)) seemed to have had option 1 in mind (although the ExportJob's don't support it yet).

Does anyone have an opinion?

**Edit**
I now implemented the following command line arguments:
--export-range   (a string used as a page range vector for exports. e.g. "-2,4-6,8,10-" will only export pages 1,2,4,5,6,8 and all pages after no 10)
--export-png-dpi   (an integer used as dpi for png exports)
--export-png-width   (an integer used as the width in pixels of png exports)
--export-png-height   (an integer used as the height in pixels of png exports)

if --export-png-dpi is set, then --export-png-width and --export-png-height are ignored.
if --export-png-width is set, then --export-png-height is ignored.
In particular, the exported png files always have the original page's aspect ratio.

**Further possible improvements:**
Add the dpi, width and height parameters to the GUI's ExportDialog

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -52,8 +52,9 @@ auto migrateSettings() -> MigrateResult;
 void checkForErrorlog();
 void checkForEmergencySave(Control* control);
 
-auto exportPdf(const char* input, const char* output) -> int;
-auto exportImg(const char* input, const char* output) -> int;
+auto exportPdf(const char* input, const char* output, const char* range, bool noBackground) -> int;
+auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
+               bool noBackground) -> int;
 
 void initResourcePath(GladeSearchpath* gladePath, const gchar* relativePathAndFile, bool failIfNotFound = true);
 
@@ -225,7 +226,22 @@ void checkForEmergencySave(Control* control) {
     gtk_widget_destroy(dialog);
 }
 
-auto exportImg(const char* input, const char* output) -> int {
+/**
+ * @brief Export the input file as a bunch of image files (one per page)
+ * @param input Path to the input file
+ * @param output Path to the output file(s)
+ * @param range Page range to be parsed. If range=nullptr, exports the whole file
+ * @param pngDpi Set dpi for Png files. Non positive values are ignored
+ * @param pngWidth Set the width for Png files. Non positive values are ignored
+ * @param pngHeight Set the height for Png files. Non positive values are ignored
+ * @param noBackground If true, the exported image file has transparent background
+ *
+ *  The priority is: pngDpi overwrites pngWidth overwrites pngHeight
+ *
+ * @return 0 on success, -2 on failure opening the input file, -3 on export failure
+ */
+auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
+               bool noBackground) -> int {
     LoadHandler loader;
 
     Document* doc = loader.loadDocument(input);
@@ -242,10 +258,26 @@ auto exportImg(const char* input, const char* output) -> int {
     }
 
     PageRangeVector exportRange;
-    exportRange.push_back(new PageRangeEntry(0, int(doc->getPageCount() - 1)));
+    if (range) {
+        exportRange = PageRange::parse(range, int(doc->getPageCount()));
+    } else {
+        exportRange.push_back(new PageRangeEntry(0, int(doc->getPageCount() - 1)));
+    }
+
     DummyProgressListener progress;
 
-    ImageExport imgExport(doc, path, format, false, exportRange);
+    ImageExport imgExport(doc, path, format, noBackground, exportRange);
+
+    if (format == EXPORT_GRAPHICS_PNG) {
+        if (pngDpi > 0) {
+            imgExport.setQualityParameter(EXPORT_QUALITY_DPI, pngDpi);
+        } else if (pngWidth > 0) {
+            imgExport.setQualityParameter(EXPORT_QUALITY_WIDTH, pngWidth);
+        } else if (pngHeight > 0) {
+            imgExport.setQualityParameter(EXPORT_QUALITY_HEIGHT, pngHeight);
+        }
+    }
+
     imgExport.exportGraphics(&progress);
 
     for (PageRangeEntry* e: exportRange) {
@@ -263,7 +295,16 @@ auto exportImg(const char* input, const char* output) -> int {
     return 0;  // no error
 }
 
-auto exportPdf(const char* input, const char* output) -> int {
+/**
+ * @brief Export the input file as pdf
+ * @param input Path to the input file
+ * @param output Path to the output file
+ * @param range Page range to be parsed. If range=nullptr, exports the whole file
+ * @param noBackground If true, the exported pdf file has white background
+ *
+ * @return 0 on success, -2 on failure opening the input file, -3 on export failure
+ */
+auto exportPdf(const char* input, const char* output, const char* range, bool noBackground) -> int {
     LoadHandler loader;
 
     Document* doc = loader.loadDocument(input);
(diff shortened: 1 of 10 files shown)


bhennion:
Okay, I think I'm done. I'm happy to take any and all comments you may have. 
I tested those functionalities (see the above post) on various files, with or without a PDF backgound. I also tested that I did not break the GUI's export functionalities.

rolandlo (changes requested):
Thanks for your great work! Everything works as expected. The code looks clear and your design decisions seem good. I have some minor nitpicks though:

1) About code comments: Please format signature comments in header files compatible with `doxygen`, using keywords like `@brief`, `@param`, `@return`. See the `SplineSegment` class header file for how we would like to have it. I know signature comments are missing in big parts of the code base, but we try to improve.

2) In `ImageExport::createSurface` there are some implicit casts between double and int. Please make them explicit. Moreover rounding to the nearest integer would possibly make more sense than rounding down (via implicit casting).

3) It would be nice to have the dpi options in the GUI export dialog as well. As I understand you have chosen not to do it, so it will be left for the future. I am fine with that, but I really hope someone (else) will add it.

Otherwise everything looks fine to me.

bhennion:
@rolando: I will do 1. and 2. as soon as possible.

About 3.: I though about it, and there is an obvious quick solution. However, I fear this is not going to be a beautiful solution. For a good solution (in my opinion), some work should be done. Here is what I mean:

a. There should be a common motherclass to all Export classes (XojPdfCairoExport, ImgExport, Xoj, other classes for other formats to come) containing (among others) a list of available options for each available formats (page range, background, dpi stuff and so on).

b. Both the CLI options and the GUI (probably CustomExportJob) then access a table of all available formats and options. Those options would become available to both as soon as they are implemented.

Of course, this requires some amount of work, but in the long run, should save people's time. Was it yesterday that was mentionned adding a new format?
I guess the job would essentially consist of moving some code around and remove some duplicate code (for instance, every Export class has a function createExport or so, and they all have the same structure: export pages one by one).

I assume this strategy was the idea behind the XojPdfExport class (that seems to have been used so that two pdfExporter could coexist), or of BaseExportJob (except a Job interacts with the GUI, which I assume was the reason for creating ImgExport so that exporting becomes available to the CLI).

If you think this is reasonible, I can try and think about it some more. I could also implement quickly the ugly solution, and still think about cleaning afterwards.

rolandlo:
@bhennion 
I think you are right about 3. Nevertheless I would suggest to implement the quick ugly solution for now. It may take a long time otherwise until pure GUI users profit from dpi export settings. We don't know when somebody is going to touch the GUI export refactoring, since there are lots of more urgent things to do.

bhennion:
Okay, this last commit should take care of rolandlo's 3 remarks.

I do have a question about how to handle Gtk. I have an enumeration (src/control/jobs/ImageExport.h)
`enum ExportQualityCriterion { EXPORT_QUALITY_DPI, EXPORT_QUALITY_WIDTH, EXPORT_QUALITY_HEIGHT };`
which is used (among other things) to match items of a GtkListStore (exported from Glade). For my code to function, those lists have to agree (and the order of their elements too). This means for instance that the second item in the GtkListStore must correspond to setting the width.
Is there a clever way of making sure of that? (so that whoever adds a new item to the list cannot introduce a bug by simply misplacing an item). Of course, I added a comment warning about this, but I'd prefer a more reliable solution.

Febbe:
blocked by #2252

rolandlo (changes requested):
Very nice high quality work! I am fine with all your changes. It works like a charm. 
Please solve the merge conflicts that have appeared after merging #2252 

About your question: I don't know what the most secure way to have these two lists agree would be. We have some other places in the code, where two lists must be in the same order and it was also just solved with a comment in the code. In your case I think it would be really quite obvious and easy to fix, if someone accidentally permuted the list in the glade file. So I am fine with how it is right now.

bhennion:
> Very nice high quality work! I am fine with all your changes. It works like a charm.

Thanks!

> Please solve the merge conflicts that have appeared after merging #2252

Done.

> About your question: I don't know what the most secure way to have these two lists agree would be. We have some other places in the code, where two lists must be in the same order and it was also just solved with a comment in the code. In your case I think it would be really quite obvious and easy to fix, if someone accidentally permuted the list in the glade file. So I am fine with how it is right now.

Ok. Fine by me. Should I squash it all together (and re-clang it all, apparently there is a misplace space)?

mjg:
Thanks a lot for your work, these features are exactly what allows me to switch to xournalpp from xournal.

For xournal (legacy) there was a config option about exporting the background. This omitted any paper background but did export a PDF background. In contrast, xournalpp's "export as PDF with plain background" is really plain, and the command line switch in this PR inherits that behavior.

Especially for exporting from command line, xournal (legacy)'s behavior may be more useful: When do you want to export only the annotations to a PDF? On the other hand, writing on a lined paper but printing on a blank one seems very common.

I don't suggest adding xournalpp's established behavior, but I'm wondering whether `--export-no-paper` would make a nice and easy addition to this PR (together with "export as PDF without paper"), especially for documents with mixed pages.

rolandlo:
> Ok. Fine by me. Should I squash it all together (and re-clang it all, apparently there is a misplace space)?

Yes, you should make sure Clang Format is applied (the problem is only the spacing on the second last line of the `exportOptions` array) and squash into one commit.
i haven't quite understood @mjg's suggestion yet. You may wait for the upcoming patch (suggested by him, I guess?!).

mjg:
> > Ok. Fine by me. Should I squash it all together (and re-clang it all, apparently there is a misplace space)?
> 
> Yes, you should make sure Clang Format is applied (the problem is only the spacing on the second last line of the `exportOptions` array) and squash into one commit.
> i haven't quite understood @mjg's suggestion yet. You may wait for the upcoming patch (suggested by him, I guess?!).

Sorry for being unclear. xournal documents can have several types of backgrounds: PDF pages, paper backgrounds (ruled/graph/...) and images. It might be that all of them are "paper backgrounds" in xournalpp terminology, I was referring to ruled/graph/... as "paper backgrounds" which may be the source of confusion. Let's call them ruling, so we have PDF, ruling and image as possible backgrounds.

In xournal (legacy), when you unselect "print paper ruling" then the ruling is not printed or exported but a PDF background is.
In xournalpp, when you "export as PDF with plain background" then neither ruling nor PDF background are exported - the background is truely plain. Both is as expected from the wording of the options.

What I suggest is an additional export mode for xournalpp where PDF backgrounds are exported but the paper ruling ist not: Often, the ruling is used just as a guide while writing but you don't want it in the output. You do want the PDF that you are annotating to be visible. And you may have different backgrounds on different pages.

And that's what I have a patch for: an additional export mode and the command line switches. I understand now that I have to reword the mode ("export without paper ruling" instead of "export without paper background"), and I'm wondering whether we really want the "shut all background off" mode as a single option on the command line (as above, plus another "PDF yes ruling" no mode), or rather two independent options for PDF and ruling (and possible one for images).

I'm new to xournalpp and have to check how the three backgrounds are drawn exactly since DocumentView::drawBackground() does only 2 out of the 3 types and defers the 3rd to PdfView, and I haven't looked at image backgrounds nor svg export yet.

rolandlo:
@mjg 
Aha, I didn't think about documents with different page backgrounds. So you say when you have a document, which has some pages with pdf background and other pages with ruled/graph/... background, then there is no option in Xournal++ to keep the pdf background in the exported file while getting rid of the background for pages with ruled/graph/... background. 
I think this has a point (as does exporting layers separately #2373, which Xournal can also do), but I think it would be better to do (and discuss) this in a new PR. So if you don't mind, let's merge this PR first and then you can make your PR on top of it.

bhennion:
Here you go. I re-clanged and squashed. Didn't need to (re)rebase, so you should be good to go.

rolandlo (approved):
I will merge it now. Thanks again.

PR #2453, opened 2020-11-28
Fix and simplify metadata manager read

### LINKED ISSUE #2445: Command line option --page not working in 1.1.0+dev
**Affects versions :**
 - OS: Linux, Windows 
 - Version of Xournal++: 1.1.0+dev (at least from git commit 8afb3841 to the latest version)
 - Installation method: All of PPA, flatpak and building from source
**Describe the bug**
The command line options `-n` (in short) and `--page` (in long) for scrolling to the specified page do not work any more in version 1.1.0+dev. Simply no scroll is taking place.

**To Reproduce**
1. Create a multipage document, say `foo.xopp` with >2 pages
2. Run xournalpp with the --page option, e.g. `xournalpp --page=2 foo.xopp`
3. Observe that the document loads without scrolling to the 2nd page.

**Expected behavior**
I would expect scrolling to the specified page after the document has loaded.

**Additional context**
- In version 1.0.18 and 1.0.19 the options
`--page=2` or `--page 2` or `-n 2` work, the option `-n=2` does not (and probably is not supposed to work). 
In version 1.1.0+dev the page number is still correctly read with the former 3 options, but the scrolling does not take place.

- Even with working --page (or -n) option the page number is shifted. According to `xournalpp --help` the first page number should be 1, but it is 0 indeed. So `--page=2` scrolls to the 3rd page instead of the 2nd one. That should be corrected, too.

- The error seems to happen because of an invalid Metadata entry.

rolandlo:
Fixes #2445 
Without this fix the metadata manager simply does not work.

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -294,7 +294,7 @@ struct XournalMainPrivate {
     XournalMainPrivate() = default;
     XournalMainPrivate(XournalMainPrivate&&) = delete;
     XournalMainPrivate(XournalMainPrivate const&) = delete;
-    auto operator=(XournalMainPrivate&&) -> XournalMainPrivate = delete;
+    auto operator=(XournalMainPrivate &&) -> XournalMainPrivate = delete;
     auto operator=(XournalMainPrivate const&) -> XournalMainPrivate = delete;
 
     ~XournalMainPrivate() {
@@ -307,7 +307,7 @@ struct XournalMainPrivate {
     gchar* pdfFilename{};
     gchar* imgFilename{};
     gboolean showVersion = false;
-    int openAtPageNumber = -1;
+    int openAtPageNumber = 0;  // when no --page is used, the document opens at the page specified in the metadata file
     std::unique_ptr<GladeSearchpath> gladePath;
     std::unique_ptr<Control> control;
     std::unique_ptr<MainWindow> win;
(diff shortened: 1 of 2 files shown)


src/control/XournalMain.cpp
```diff
@@ -382,7 +382,7 @@ auto XournalMain::run(int argc, char* argv[]) -> int {
 
         try {
             if (fs::exists(p)) {
-                opened = control->openFile(p, openAtPageNumber);
+                opened = control->openFile(p, openAtPageNumber - 1);  // First page for user is page 1
```
Febbe, line 385: What happens if you use zero as input?
Make sure this doesn't fail. Or if it fails, that it will fail verbose.

if `openAtPageNumber` is a signed type, you could use something like `std::max(0, openAtPageNumber - 1)` if it's a unsigned type `std::max(1, openAtPageNumber) - 1` could do the job.

Febbe (approved):

src/control/XournalMain.cpp
rolandlo: If zero (or any negative integer) is used as input, `Control::fileLoaded` will eventually load the document opened at the same page as it was loaded last time (without any complaints) if the metadata is valid. If the metadata is invalid, it will open on the first page.

Maybe I could initialize `openAtPageNumber` by 0 instead of -1, since -1 is subtracted now. But it does not really make any difference, since -1 and -2 are both negative.

PR #2468, opened 2020-11-30
Remove using std::* statements in header files

rolandlo:
As wished by @Febbe on the TODO list, I removed using `std::` statements in all header files, and added `std::` accordingly throughout the header files. In the source files I added `std::` when there were only few occurences. Otherwise I added `using std::` in the source file.


src/util/XournalType.h
```diff
@@ -26,7 +26,4 @@
 // Include string and vector everywhere
 #include <cstring>  // Todo: Use std::string_view if possible
 #include <string>
-#include <vector>
-
-using std::string;
-using std::vector;
+#include <vector>
```
Febbe, line 29: Remove this header completely.

Febbe:
LGTM

src/util/XournalType.h
rolandlo: Done. I suggest to wait for  #2339 to get merged.

Febbe: Yes, this is a very invasive change, breaking the code of nearly every PR.

Febbe (approved):

LittleHuba:
This should be the last commit making it into 1.1.0.

All PRs not part of the 1.1.0 release can then update their work without further delaying the release of 1.1.0.

Technius:
This is pretty low priority imo. Let's leave this for after the release--we don't want to introduce random breakage from refactoring changes.

rolandlo:
> This is pretty low priority imo. Let's leave this for after the release--we don't want to introduce random breakage from refactoring changes.

In my opinion it's time to merge this PR. Although this PR changes lots of files and may lead to some merge conflicts in other PR's, solving these merge conflicts is almost trivial (stick to your own PR and add `std::` where needed). I went through all changes once more and verified that the PR does nothing else than removing the XournalType header and `using std::` statements in header files, which is compensated by 
- adding `std::` to `string`, `vector`, `list`, `ofstream` at lots of places, 
- adding `using std::string` or `using std::vector` in some `.cpp` files, where it seemed advantageous.
-  adding header files includes where needed

 As long as it compiles this cannot have introduced any bugs.

So, if no objections are raised, I will merge this PR in 24 hours.

Febbe:
Yes do it.

PR #2525, opened 2020-12-10
Fix Segfault due to missing nullptr

bhennion:
I had a segfault message due to the lack of nullptr at the end of the option array. See
[Gtk's documentation](https://developer.gnome.org/gio/stable/GApplication.html#g-application-add-main-option-entries):

`void
g_application_add_main_option_entries (GApplication *application,
                                       const GOptionEntry *entries);`

>application: a GApplication
>
>entries: (array zero-terminated=1) (element-type GOptionEntry) a NULL-terminated list of GOptionEntrys

**Edit** Fixed link and added quote


rolandlo:
Thanks for the PR. I will merge it.

PR #2530, opened 2020-12-12
Take toolsize and erasertype from corresponding toolbar tool when triggered by Button

### LINKED ISSUE #2521: Eraser via stylus button doesn't reflect eraser size setting.
**Affects versions :**
 - OS: Arch Linux
 - Desktop environment: X11
 - libgtk: 3.24.24
 - Version of Xournal++: 1.1.0+dev ( bd0b5bb717351c9c2a9462ad6c2e4778db99c9ed )

**Describe the bug**
If you use the eraser via stylus button, its eraser size is not the same as the eraser setting, but it is the same as the pen setting.

**To Reproduce**
1. Configure stylus button 1 to invoke eraser. Choose size "don't change".
2. Set pen size 'very thick' and eraser size 'very fine'.
3. If you use eraser via menu or Ctrl-Shift-E, the eraser size is 'very fine'.
4. If you use eraser via stylus button 1, the eraser size is 'very thick'.

**Expected behavior**
At the step4 above, the eraser size should be 'very fine'.

idotobi:
Fix the behaviour of "Don't Change" settings for the Buttons to use the corresponding Tool from the toolbar as reference.

This fixes #2521 .

### DIFF
--- a/src/control/settings/ButtonConfig.cpp
+++ b/src/control/settings/ButtonConfig.cpp
@@ -28,18 +28,18 @@ void ButtonConfig::initButton(ToolHandler* toolHandler, Button button) {
     toolHandler->resetButtonTool(this->action, button);
 
     if (this->action == TOOL_PEN || this->action == TOOL_HIGHLIGHTER || this->action == TOOL_ERASER) {
-
-        if (this->drawingType != DRAWING_TYPE_DONT_CHANGE) {
-            toolHandler->setButtonDrawingType(this->drawingType, button);
-        }
-
         if (this->size != TOOL_SIZE_NONE) {
             toolHandler->setButtonSize(this->size, button);
         }
-        toolHandler->setButtonColor(this->color, button);
-    }
-    if (this->action == TOOL_ERASER && this->eraserMode != ERASER_TYPE_NONE) {
-        toolHandler->setButtonEraserType(this->eraserMode, button);
+        if (this->action == TOOL_PEN || this->action == TOOL_HIGHLIGHTER) {
+            toolHandler->setButtonColor(this->color, button);
+            if (this->drawingType != DRAWING_TYPE_DONT_CHANGE)
+                toolHandler->setButtonDrawingType(this->drawingType, button);
+        }
+        if (this->action == TOOL_ERASER) {
+            if (this->eraserMode != ERASER_TYPE_NONE)
+                toolHandler->setButtonEraserType(this->eraserMode, button);
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


rolandlo:
Yes, according to my tests this is exactly the behavior I'd like to see and that most users would expect. Let me just check carefully if everything is fine.

src/control/settings/ButtonConfig.cpp
```diff
@@ -28,18 +28,18 @@ void ButtonConfig::initButton(ToolHandler* toolHandler, Button button) {
     toolHandler->resetButtonTool(this->action, button);
 
     if (this->action == TOOL_PEN || this->action == TOOL_HIGHLIGHTER || this->action == TOOL_ERASER) {
```
idotobi, line 30: @rolandlo : This part I only changed to make the flow similar to `ButtonConfig::applyNoChangeSettings`, which was the function that actually needed to change.
So in this function the changes should be logically equivalent to the version before.

src/control/settings/ButtonConfig.cpp
```diff
@@ -50,18 +50,18 @@ void ButtonConfig::applyConfigToToolbarTool(ToolHandler* toolHandler) {
     toolHandler->selectTool(this->action);
 
     if (this->action == TOOL_PEN || this->action == TOOL_HIGHLIGHTER || this->action == TOOL_ERASER) {
```
idotobi, line 52: @rolandlo : This part I only changed to make the flow similar to `ButtonConfig::applyNoChangeSettings`, which was the function that actually needed to change.
So in this function the changes should be logically equivalent to the version before.

src/control/settings/ButtonConfig.cpp
```diff
@@ -70,17 +70,20 @@ bool ButtonConfig::applyNoChangeSettings(ToolHandler* toolHandler, Button button
         return false;
     }
     if (this->action == TOOL_PEN || this->action == TOOL_HIGHLIGHTER || this->action == TOOL_ERASER) {
-
-        if (this->drawingType == DRAWING_TYPE_DONT_CHANGE) {
-            toolHandler->setButtonDrawingType(toolHandler->getDrawingType(SelectedTool::toolbar), button);
-        }
+        Tool& correspondingTool = toolHandler->getTool(this->action);
```
idotobi, line 73: @rolandlo : This is the key as the settings should be taken from the corresponding Tool not just the currently active tool

rolandlo (approved):
Please change the title of the PR to reflect its content more precisely. The changes LGTM.

PR #2536, opened 2020-12-13
Fix regression default tool at startup

### LINKED ISSUE #2535: Regression: Xournal++ forgets pen options
**Affects versions :**
 - Version of Xournal++:
    the problem occurs in 7ab24766206264c179d0a1e405cb2c1d43c7e83c
    It was working fine in ce4a13a0d0ca6aeb5456daaef782e0167a19fc1e
 - OS: Arch Linux (5.4.82-1-lts)
 - (Linux only) Desktop environment: KDE Plasma 5.20.4 on X11
     3.24.24
 - Installation method: [e.g. PPA, Flatpak, official release, etc.]
     Arch Linux PKGBUILD with custom release.

**Describe the bug**
Xournalpp always starts up with a black pen of small size. I don't know whether xournalpp previously remebered that per document or globally, but is does not seem to be saved anywhere anymore.
Quick edit: It also forgets shape settings like which shape was selected and whether it was enabled.

**To Reproduce**
1. Change the pen to a color != black or a size > small, or enable a shape tool
2. Save doc and close xournal
3. Open the same doc. 

**Expected behavior**
I would expect Xournalpp to behave like it did previously and remember my pen settings.

idotobi:
This commit changes the behaviour back to always using the last tool
selected at startup, rather then the default button Tool.

This fixes #2535.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2493,9 +2493,6 @@ void Control::initButtonTool() {
         cfg = settings->getButtonConfig(b);
         cfg->initButton(this->toolHandler, b);
     }
-
-    cfg = settings->getButtonConfig(Button::BUTTON_DEFAULT);
-    cfg->applyConfigToToolbarTool(this->toolHandler);
 }
 
 auto Control::askToReplace(fs::path const& filepath) const -> bool {


rolandlo (approved):
LGTM

Issue #2066, opened 2020-06-14
Scrolling not working properly on new documents


nicola-bastianello:
**Affects versions :**
 - OS: Xubuntu 18.04
 - Desktop environment: XFCE
 - Which version of libgtk do you use: 3.22.30
 - Version of Xournal++: 1.1.0+dev
 - Installation method: PPA

**Describe the bug**
After launching Xournal++ on a new document scrolling does not work; when launching it by opening a saved document scrolling works. Zooming out solves the issue.

**To Reproduce**
1. Open Xournal++ on a new document
2. Scroll does not work:
    * the scroll bar can be moved up and down with the mouse, but does not affect which part of the document is shown
    * with the two-finger gesture on the laptop's touchpad the scroll bar moves rapidly and erratically up and down, but still only the top of the page is shown
4. Try opening a previously saved document: scroll still does not work
3. Zoom out (CTRL+minus) and scrolling starts to work normally (tried after seeing -- possibly related -- issue #2062)
4. Close Xournal++, and open a previously saved document (even a blank one): scrolling works normally

**Additional context**
The issue is the same when launching Xournal++ from the program menu and a terminal. When opening from a terminal some error messages are printed, see [attached file](https://github.com/xournalpp/xournalpp/files/4776821/terminal.txt))

Technius:
Does this occur when zoom fit is disabled?

Possible regression #1383.

Technius:
@Febbe maybe the recent zoom changes caused this?

Technius:
Do you have the `Touch Workaround` enabled?

nicolabastianello:
1. Indeed, it happens only when zoom fit is on.
2. No, I don't have the touch workaraound enabled.

wolfgang-noichl:
**Version**

- Xubuntu 20.04
- libgtk 3.24.18-1ubuntu1
- Xournal 1.0.6-0\~202005080350\~ubuntu19.10.1

**Description**

Reproduced the above. In more detail: If the GTK Touch / Scrolling workaround is enabled, zooming does not restore the ability to scroll. "Zoom gestures" settings don't seem to have any influence.

BMicraft:
This seems to be the same issue as 2138. I think this issue can be close in favor of the other one.

rolandlo:
This is probably fixed by PR #2541. I will close it in favor of #2138.

Issue #1699, opened 2020-01-08
Zoom fet to screen- no scroll


masac:
**Affects versions :**
 - OS:Linux Mint 19.3 Mate
 - Version of Xournal++ 1.1.0+dev

**Describe the bug**
In zoom fet to screen, I cannot use scroll.

**To Reproduce**
https://drive.google.com/file/d/1tcnbeFB051O2Q64ymjZY-u27XeDFkz6G/view?usp=sharing

Technius:
Same cause as #1668.

BMicraft:
I don't think this is quite the same as #1608.

This seems more like #2138 and can probably be closed in favor of that.

rolandlo:
Closed in favor of #2138. Note there is a fix available, PR #2541.

PR #2541, opened 2020-12-13
Fixes zoomFitValue update hindering scrolling

### LINKED ISSUE #2138: Not able to scroll notes until I change the zoom
**Affects versions :**
 - OS: Windows 10
 - Installation method: Official Windows release

**Describe the bug**
When I open a saved file .xopp there is no way to scroll pages until I change zoom.

**To Reproduce**
With Xournal++ closed, open a saved .xopp file and try to scroll up/down.

**Expected behavior**
I would like to scroll pages without need changing zoom.

**Screenshots of Problem**
This is the default zoom grade when program opens up
image

Decreasing or increasing zoom allows me to scroll pages
image

### LINKED ISSUE #2318: Go back to 'Normal Drawing'
Hi,
when I draw an object (a circle, a box..) I cannot find a fast way to go back to 'normal drawing' (without stroke recognition).
I usually click 'Ctrl+1' two times, but maybe a more direct way exists... if so, please tell me!
Using 'Default Tool' is another possibility but resets also color and pen size
I would suggest adding an entry 'Normal Drawing' in the 'Shape' list, with an associated shortcut

Thanks,
pnc

rolandlo:
Fixes #2138 (and its duplicates #1699 and #2066)

The cause of the problem was that for the users that reported the problem the allocation width that was passed to `updateZoomFitValue` did not match with the width of the visible rectangle, which made the zoomFitValue alternate in rapid succession between two different values, which invoked `XournalView::scrollTo` method to keep the top part of the current page visible (if the whole page did not fit on the screen).

It must be checked carefully if we don't reintroduce a bug (#1139, #1132 ) that has been fixed in #1213. So far it seems, it doesn't reintroduce these bugs.

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -102,7 +102,7 @@ void LatexController::findSelectedTexElement() {
         const double centerX = visibleBounds.x + 0.5 * visibleBounds.width;
         const double centerY = visibleBounds.y + 0.5 * visibleBounds.height;
 
-        if (layout->getViewAt(centerX, centerY) == this->view) {
+        if (layout->getPageViewAt(centerX, centerY) == this->view) {
             // Pick the center of the visible area (converting from screen to page coordinates)
             this->posx = (centerX - this->view->getX()) / zoom;
             this->posy = (centerY - this->view->getY()) / zoom;
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -580,7 +580,7 @@ auto EditSelection::getPageViewUnderCursor() -> XojPageView* {
 
 
     Layout* layout = gtk_xournal_get_layout(this->view->getXournal()->getWidget());
-    XojPageView* v = layout->getViewAt(hx, hy);
+    XojPageView* v = layout->getPageViewAt(hx, hy);
 
     return v;
 }
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -140,9 +140,7 @@ void ZoomControl::setZoom100Value(double zoom100Val) {
     fireZoomRangeValueChanged();
 }
 
-auto ZoomControl::updateZoomFitValue(size_t pageNo) -> bool { return updateZoomFitValue(getVisibleRect(), pageNo); }
-
-auto ZoomControl::updateZoomFitValue(const Rectangle<double>& widget_rect, size_t pageNo) -> bool {
+auto ZoomControl::updateZoomFitValue(size_t pageNo) -> bool {
     if (pageNo == 0) {
         pageNo = view->getCurrentPage();
     }
@@ -151,6 +149,7 @@ auto ZoomControl::updateZoomFitValue(const Rectangle<double>& widget_rect, size_
         return false;
     }
 
+    Rectangle widget_rect = getVisibleRect();
     double zoom_fit_width = widget_rect.width / (page->getWidth() + 20.0);
     if (zoom_fit_width < this->zoomMin || zoom_fit_width > this->zoomMax) {
         return false;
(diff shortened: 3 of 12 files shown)


tattsan:
> Fixes # 2318 (and its duplicates # 1699 and # 2066)
Is #2318  a typo for #2138 ?

rolandlo:
> > Fixes # 2318 (and its duplicates # 1699 and # 2066)
> 
> Is #2318 a typo for #2138 ?

Yes, thanks for pointing it out! I corrected it.

Technius:
Seems like this would also fix the alternating/jumpy zoom problem with pages of different size (#1660, which was previously patched with a hotfix). Do you know if it does?

rolandlo:
I don't know. I can try removing the hotfix and see how it goes.

rolandlo:
I have removed the hotfix and don't get any crashes. The behavior is still the same. In zoomFitMode when scrolling from one page to the other page of different width (which changes the zoom) the page is still scrolled to the top (if it is not fully visible). The comes from the fact that in `XournalView::zoomChanged()` we call `scrollTo(currentPage)` in case of zoomFitMode.

Febbe:
I don't see any crashes, but the Rectangle returned by `getVisibleRect()` seems to be wrong.
I think that was the reason I decided to use the `GdkRectangle` instead of the `getVisibleRect()` function.


To reproduce the behavior: 
maximize the window with view100%, then minimize, then move the mouse into the window: Layout resizes to a wrong layout.

rolandlo:
> I don't see any crashes, but the Rectangle returned by `getVisibleRect()` seems to be wrong.
> I think that was the reason I decided to use the `GdkRectangle` instead of the `getVisibleRect()` function.
> 
> To reproduce the behavior:
> maximize the window with view100%, then minimize, then move the mouse into the window: Layout resizes to a wrong layout.

Not for me. It seems for me `getVisibleRect` always reports the same width and height like the `GdkRectangle` does. Can you create a little gif showing the difference?

Febbe:
Ok seems like I still had the Gtk-Touch-Workaround enabled.
But we want to remove that.

rolandlo:
@Febbe So can I merge it then?

Febbe:
It would be cleaner to remove getVisibleRect and use the correct allocation function however, but that might be to intruding / complicated to fix.

Febbe:
Ok rechecked it on Windows. There, the bug still occur.

Febbe:
bug

rolandlo:
Sorry, but the less dirty hack doesn't work at all (right now). Try inserting pages.

Febbe:
That's strange, I testet exactly that on my Windows machine. I'm gonna take a Look.

rolandlo:
Ok, scrolling after inserting a page works again. But the UI becomes unresponsible (I can't draw for a while) after inserting a few pages.

Febbe:
blocked by #2435

Febbe:
@rolandlo can you review my part. I think we can merge it after #2435 if everything is alright.

Febbe (approved):

rolandlo:
@Febbe Since #2577 contains this fix wouldn't it be better to close this PR and just proceed with PR #2577

Febbe:
I would do it the other way around. I also think that your and my commits require each other

rolandlo:
What do you mean with "the other way round"? First merge this PR, then PR #2577?

Febbe:
Since both PRS contain the Same content, merge this PR and ignore the new

rolandlo:
They do not contain the same content. This one is just a part of the new one. It contains less additions and deletions, and it doesn't work.

Febbe:
Ok somehow the push didn't work over the git gui, or I made something wrong. Now they are the same. Thanks for the catch.

rolandlo:
Ok, I will review your changes then.

rolandlo:
@Febbe I am happy with your changes. Only minor nitpick: The last comment in `Layout.h` seems incomplete (and contains a typo: `laotPages`)

I have tested the PR on my laptop and on my Pinephone. Even on the Pinephone (with 2 GB RAM) it ran pretty smoothly and things like inserting pages, changing page layout or paper format worked fine.

Febbe:
Merged because @personalizedrefrigerator agreed to rebase his work.

Issue #795, opened 2019-01-21
Export as image from CLI


anishmittal2020:
1. It will be great if we could export (and import) the notes as `SVG`? 
    -   SVG is supported by modern web browsers
    -   SVG is text based
    -   there are Javascript libraries which allow drawing SVG in browsers (like [this one](https://github.com/SVG-Edit/svgedit)
    -   Markdown allows embedding SVG as images

---

2. Also, can we export as `png` (and `svg`) from CLI?  

I am motivated by this as I want to use `xournalpp` only for handwritten notes. I want to use  `Markdown` and `Pandoc` as a container of my notes. I want to make small handwritten notes as separate files and embed them inside my larger markdown file.

andreasb242:
SVG Export was removed because there was an issue in exporting stuff (it was there, but not working).
But this is now fixed, SVG export is no Problem, as this is supported by cairo.
It's not that much work.

Commandline arguments is also not a big issue.

SVG Import is much work, because currently we do not support SVG rendering, so this is a new implementation. (and is not supported by cairo).

You can expect the SVG export in a much shorter time than the import. But also then, you may noticed how many tickets are currently written withing really short time...

andreasb242:
Working, one issue: we need to export as SVG 1.2, and e.g. EOG does not display it correctly, Firefox does.
See xournalpp -h for the commands.

PR #2575, opened 2020-12-20
Allow exports without ruling (but with other backgrounds)

mjg:
The aim of this series is to allow exports without ruling but with the other backgrounds, since the former are often auxiliary only, but the latter are essential. If you have documents with mixed background types or batch process documents with different background types then the new `--export-no-ruling` option should come in handy.

The 1st commit is an independent but technically related bugfix (do not export PDF background to PNG when asked not to).

The 2nd commit marks an existing bug as TODO (PDF backgrounds are never exported to SVG). I don't know if you prefer to carry TODOs in code. A proper fix would be to render the PDF to an image cairo surface and attach the image to the SVG surface, which is over my cairo head, unfortunately.

The remaining commits address the actual purpose: refactor and introduce the new export mode.

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -356,7 +356,7 @@ struct XournalMainPrivate {
     XournalMainPrivate() = default;
     XournalMainPrivate(XournalMainPrivate&&) = delete;
     XournalMainPrivate(XournalMainPrivate const&) = delete;
-    auto operator=(XournalMainPrivate &&) -> XournalMainPrivate = delete;
+    auto operator=(XournalMainPrivate&&) -> XournalMainPrivate = delete;
     auto operator=(XournalMainPrivate const&) -> XournalMainPrivate = delete;
 
     ~XournalMainPrivate() {
@@ -374,9 +374,8 @@ struct XournalMainPrivate {
     int exportPngDpi = -1;
     int exportPngWidth = -1;
     int exportPngHeight = -1;
-    gboolean exportNoBackground =
-            false;  // don't use bool, see
-                    // https://stackoverflow.com/questions/21152042/is-glib-command-line-parsing-order-sensitive
+    gboolean exportNoBackground = false;
+    gboolean exportNoRuling = false;
     gboolean progressiveMode = false;
     std::unique_ptr<GladeSearchpath> gladePath;
     std::unique_ptr<Control> control;
@@ -584,13 +583,17 @@ auto on_handle_local_options(GApplication*, GVariantDict*, XMPtr app_data) -> gi
 
     if (app_data->pdfFilename && app_data->optFilename && *app_data->optFilename) {
         return exportPdf(*app_data->optFilename, app_data->pdfFilename, app_data->exportRange,
-                         app_data->exportNoBackground ? EXPORT_BACKGROUND_NONE : EXPORT_BACKGROUND_ALL,
+                         app_data->exportNoBackground ? EXPORT_BACKGROUND_NONE :
+                         app_data->exportNoRuling     ? EXPORT_BACKGROUND_UNRULED :
+                                                        EXPORT_BACKGROUND_ALL,
                          app_data->progressiveMode);
     }
     if (app_data->imgFilename && app_data->optFilename && *app_data->optFilename) {
         return exportImg(*app_data->optFilename, app_data->imgFilename, app_data->exportRange, app_data->exportPngDpi,
                          app_data->exportPngWidth, app_data->exportPngHeight,
-                         app_data->exportNoBackground ? EXPORT_BACKGROUND_NONE : EXPORT_BACKGROUND_ALL);
+                         app_data->exportNoBackground ? EXPORT_BACKGROUND_NONE :
+                         app_data->exportNoRuling     ? EXPORT_BACKGROUND_UNRULED :
+                                                        EXPORT_BACKGROUND_ALL);
     }
     return -1;
 }
@@ -639,6 +642,10 @@ auto XournalMain::run(int argc, char** argv) -> int {
                          _("Export without background\n"
                            "                                 The exported file has transparent or white background,\n"
                            "                                 depending on what its format supports\n"),
+                         0},
+            GOptionEntry{"export-no-ruling", 0, 0, G_OPTION_ARG_NONE, &app_data.exportNoRuling,
+                         _("Export without ruling\n"
+                           "                                 The exported file has no paper ruling\n"),
                          0},
             GOptionEntry{"export-layers-progressively", 0, 0, G_OPTION_ARG_NONE, &app_data.progressiveMode,
                          _("Export layers progressively\n"
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -52,10 +52,10 @@ auto migrateSettings() -> MigrateResult;
 void checkForErrorlog();
 void checkForEmergencySave(Control* control);
 
-auto exportPdf(const char* input, const char* output, const char* range, bool noBackground, bool progressiveMode)
-        -> int;
+auto exportPdf(const char* input, const char* output, const char* range, ExportBackgroundType exportBackground,
+               bool progressiveMode) -> int;
 auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
-               bool noBackground) -> int;
+               ExportBackgroundType exportBackground) -> int;
 
 void initResourcePath(GladeSearchpath* gladePath, const gchar* relativePathAndFile, bool failIfNotFound = true);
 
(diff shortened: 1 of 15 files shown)


rolandlo:
About the TODO: 
In principle it looks to me that Xournal++ is already doing it right to create a svg surface out of the pdf background. The code basically does exactly what the (working) `pdf2svg` tool does, see [code](https://github.com/dawbarton/pdf2svg/blob/0515541debd0ffcbad129d0c8f3cf64bdcd1ca8b/pdf2svg.c#L50-L81). An output is created, which resembles the one from pdf2svg when you open it up as a text-file, but something must be wrong.

mjg:
> About the TODO:
> In principle it looks to me that Xournal++ is already doing it right to create a svg surface out of the pdf background. The code basically does exactly what the (working) `pdf2svg` tool does, see [code](https://github.com/dawbarton/pdf2svg/blob/0515541debd0ffcbad129d0c8f3cf64bdcd1ca8b/pdf2svg.c#L50-L81). An output is created, which resembles the one from pdf2svg when you open it up as a text-file, but something must be wrong.

This turned out to be quite interesting: xournalpp paints a white background behind the PDF even for export (but nor for printing). It does so by painting it *after* the PDF is rendered, specifying `CAIRO_OPERATOR_DEST_OVER` as the compositing operator to get it below the PDF rendering. When rendering to a SVG 1.2 surface cairo turns this into a `comp-op:dst-over` style.

So what's the problem? SVG 1.2 never got accepted as a standard, and thus it is supported only partially. The resulting SVG does not even work in inkscape.

If we render to an SVG 1.1 surface instead then cairo renders the PDF to an image (as I suggested originally) and everything looks fine, but this is suboptimal for "vector PDFs", of course.

I'm experimenting with the drawing order in `src/view/PdfView.cpp:PdfView::drawPage()` right now, but this is my first foray into cairo, and there must be a reason why it was done in this "backwards" manner.

`pdf2svg` does not paint a white background behind the PDF, by the way. One could argue either way for xournalpp - after all, xournalpp does not export transparent PNG backgrounds on an additional white background either.

rolandlo:
> This turned out to be quite interesting: xournalpp paints a white background behind the PDF even for export (but nor for printing). It does so by painting it _after_ the PDF is rendered, specifying `CAIRO_OPERATOR_DEST_OVER` as the compositing operator to get it below the PDF rendering. When rendering to a SVG 1.2 surface cairo turns this into a `comp-op:dst-over` style.
> 
> So what's the problem? SVG 1.2 never got accepted as a standard, and thus it is supported only partially. The resulting SVG does not even work in inkscape.
> 
> If we render to an SVG 1.1 surface instead then cairo renders the PDF to an image (as I suggested originally) and everything looks fine, but this is suboptimal for "vector PDFs", of course.
> 



Interesting. See also the comment in #795:

> Working, one issue: we need to export as SVG 1.2, and e.g. EOG does not display it correctly, Firefox does.
> See xournalpp -h for the commands.

rolandlo:
Can you please rebase on current master and solve merge conflicts?

mjg:
> Can you please rebase on current master and solve merge conflicts?

Yes, the layered export is a cool feature to rebase on :-) It works well together with export-no-background. This makes me wonder about the UI though:
- Should `--export-no-background` and `--export-no-ruling` be presented in the UI as options just like `--export-presentation` since they only modify the export (but present no action such `-p` or `-i`), rather than full-blown export types?
- Should `export-presentation` be named `--export-layered` (and the UI option "export layers as pages") since "presentation" is only a specific use case (and PDF display mode), technically independent of the export mode (layered or not)?

(Sorry I wasn't aware of the layer feature while it was inflight.) I'd answer "yes" to both and would do the refactoring along with the rebase if I'm not the only one answering "ses".

bhennion:
> * Should `--export-no-background` and `--export-no-ruling` be presented in the UI as options just like `--export-presentation` since they only modify the export (but present no action such `-p` or `-i`), rather than full-blown export types?

I would agree with you there.

> * Should `export-presentation` be named `--export-layered` (and the UI option "export layers as pages") since "presentation" is only a specific use case (and PDF display mode), technically independent of the export mode (layered or not)?

^^ I precisely argued against this [here](https://github.com/xournalpp/xournalpp/pull/2589#pullrequestreview-558908375). I agree that "presentation" is a specific use case. However "export layers as pages" is IMO misleading: this is not exporting each layer to a different page. It's quite difficult to find a proper and concise description of this feature... Maybe a long(er) description (either in CLI or in a hover bubble) could include an example.

rolandlo:
> 
> Yes, the layered export is a cool feature to rebase on :-) It works well together with export-no-background. This makes me wonder about the UI though:
> 
>     * Should `--export-no-background` and `--export-no-ruling` be presented in the UI as options just like `--export-presentation` since they only modify the export (but present no action such `-p` or `-i`), rather than full-blown export types?

Good discussion point. Thinking about it, in my opinion there should be only one export type per file format (pdf, png, svg, xoj)  selectable in the drop-down menu from the file dialog, and everything else presented as options in the export dialog. 
Moreover the menu item `"Export as PDF"` may need to be renamed to `"Export as PDF (basic mode)"`, since users have found it hard to discover that `"Export as ..."` will present them with more options for PDF export as well. I am open for other suggestions though. @Febbe @Technius @LittleHuba @bhennion @sanette do you have an opinion about that?
> 
>     * Should `export-presentation` be named `--export-layered` (and the UI option "export layers as pages") since "presentation" is only a specific use case (and PDF display mode), technically independent of the export mode (layered or not)?
>
I see your point, `--export-layered` sounds understandable. For the UI option I think `export layers as pages` is not quite clear (it sounds like there will be one layer per page). `export successive layers to separate pages` would be more clear, but maybe a bit long. Otherwise something shorter like `Layered` with the concise description in the hover bubble would be still better than `export layers as pages`. What do you think @mjg @sanette?
> 
> (Sorry I wasn't aware of the layer feature while it was inflight.) I'd answer "yes" to both and would do the refactoring along with the rebase if I'm not the only one answering "ses".

bhennion:
> Good discussion point. Thinking about it, in my opinion there should be only one export type per file format (pdf, png, svg, xoj) selectable in the drop-down menu from the file dialog, and everything else presented as options in the export dialog.
> Moreover the menu item `"Export as PDF"` may need to be renamed to `"Export as PDF (basic mode)"`, since users have found it hard to discover that `"Export as ..."` will present them with more options for PDF export as well. I am open for other suggestions though. @Febbe @Technius @LittleHuba @bhennion @sanette do you have an opinion about that?

This most likely goes beyond the scope of this PR, but... IMO, a good export dialog would look like this:

1. It should prompt for a file name (and ofc allow for browsing to select a file), allow for an explicite choice of export format (and if not specified, infer the format from the file's extension). In particular, the filter should by default select every available export formats. The Ok button is only available once a format has been selected/infered (and the filename is not empty).
2. The export options should not be in a second dialog, but in an expander (e.g. GtkExpander) at the bottom of the first dialog (hidden by default). Only the options available for the selected/infered file format should be displayed.
3. Everything which is not a format (SVG/PDF/PNG/XOJ and maybe others) is an option (e.g. with/without background).

With such an export dialog, "Export to PDF" then simply preselects the export format, and advanced options are available there in the expander, should the user need them.

Technius:
> Moreover the menu item "Export as PDF" may need to be renamed to "Export as PDF (basic mode)", since users have found it hard to discover that "Export as ..." will present them with more options for PDF export as well.

 I think it would be better to replace `Export as PDF` with a `Quick Export As..` submenu, which will display multiple export options like PDF, PNG, etc. when hovered over: `Quick Export As > PDF`, `Quick Export As > PNG`, etc.

> This most likely goes beyond the scope of this PR

I agree with this, so changing the export menu should go in a separate or follow up PR. In fact, I think it's sufficient to add in the new export options as-is. We should _eventually_ overhaul the export dialog so that we don't have to add an entry for every possible combination of options.

mjg:
So, here's the rebase on top of current master (with progressive mode) plus the refactoring of all background export options to live in the export options dialog, reducing the proper export types to the 3 file types again.

rolandlo (commented):
Please fix the clang-format and squash your commits. Everything else looks good now.

mjg:
Fixed the new clang issue (due to rebase) and re-pushed.

By "squash", do you mean squashing the whole series into 1 single commit? I'm just asking because with a real merge:
- you always have `git diff A B` for viewing the total diff
- you always have `git log A..B` for viewing the combined log
- you always have `git log --first-parent` and `git log --merges` on the main branch for a quick overall picture
- you keep the break-down into small readable commits especially if they separate clean-up, refactoring and new features

It's your call for your project, of course, for me a `git merge --squash` is done quickly.

LittleHuba:
Please keep the commits as they are. They are nicely separated and only take care of one problem at a time.

rolandlo:
I am fine with the breaking into commits. Just as I made a last test, it seems there is a problem with "Annotate Pdf". Pdf files that open normally with "Open" do not open normally with "Annotate Pdf". I need to check it more closely.

rolandlo:
I think I have found a bug with "Annotate PDF", which I can reproduce. I will make a bug report. It is not related to this PR though. So I will merge this.

rolandlo:
Thanks @mjg for the great contribution. Export options have improved a lot with the latest export related PR's.

PR #2576, opened 2020-12-21
Fix an displaying issue for ARM and 32bit devices.

### LINKED ISSUE #2514: No writing sheet on ARMv7 and 32bit devices.
**Affects versions :**
 - OS: Raspberry Pi OS
 - version of libgtk 3.24.5
 - Installation method: builded directly according the Linux Build for Ubuntu/Debian

**Describe the bug**
When you open Xournal++, the zoom is fixed at 274% : no page, the background is grey so you think that the program doesn't work. You have to decrease at 100% to see the page.

**To Reproduce**
1. Launch Xournal++
2. The background of work surface is grey (no white page as expected)
3. Decrease the zoom level at 100%
3. You can see now the white page.

**Expected behavior**
Display the page all the time.

Febbe:
This commit replaced some implicit conversions with (in debug mode) checked casts.
I also had to remove some explicit type declarations, since they were wrong on ARM devices.

The problem from #2514 is, that `size_t` is defined as the return type of `size_of`, which is somehow defined as `unsigned long` which is not an aligned 8byte type, even if the target machine is a 64bit one. This may a bug in raspbianOS
For reference: https://www.keil.com/support/man/docs/armcc/armcc_chr1359125009502.htm
This results in some casting issues, when `size_t` is used in calculations where 64 bit types are used.

This commit blindly removed all implicit casts, by applying first a cast to the target signedness and then a size cast.
That way it was way more cost-efficient as to search for the implicit cast which produced the bug. All implicit casts should be removed anyways.

fixes #2514


Febbe:
blocked by #2435

Technius:
For future reference, can you elaborate on what the problems?

Febbe:
> 
> 
> For future reference, can you elaborate on what the problems were?

Done

LittleHuba:
@Febbe can you please fix the conflict. #2435 was merged so this PR is no longer blocked.

Febbe:
> @Febbe can you please fix the conflict. #2435 was merged so this PR is no longer blocked.

Done, but it must be retested by some ARMv7 users.

LittleHuba:
If nobody has an ARM device we could test this with, we can also just merge it and wait for complaints.

PR #2512, opened 2020-12-09
Support for input stabilizing

bhennion:
This follows up on #2320.
I implemented basic support for smoothing algorithms, as well as a a version of Gimp's smoothing algorithm (see #2320 for more details on this algorigthm). You can see the results in the following gifs (no smoothing vs Gimp's smoothing).
rolandlo
TentativeConvert

This is pretty rudimentary and there is no GUI control for now. To try out the smoother, you need to use --smoothing=GIMP in your command line. There are two parameters you can tweak: 

- A (double) coefficient sigma (through --smoothing-sigma=0.2 for instance). In the first gif, it was set to 1, in the second gif, it was set to 2.
- The size of the buffer used by the algorithm (through --smoothing-buffer-size=20). In the examples above, I let to 20.

[You can also try out --smoothing=ARITHMETIC which is fun but not very useful]

**Edit:** the parameters are now set in the GUI settings dialog (Stylus tab). The CLI arguments are no longer valid. **End of Edit**

There are a couple of issues to work through:

- [x] The rate events arrive depends on the input device (for Wacom tablets, I've seen it vary from 1 event per 3ms to 1 event per 10ms). Since the buffer has a predetermined size, this rate influences the smoothing parameters. It should not. For instance, one could dynamically change the buffer's length (based on what criteria?)

- (No fix, but not important) Gimp's algorithm uses velocities as a parameter. Here, the velocity is computed using the event's timestamps. However the timestamps are not quite precise enough and different events can have the same timestamp. This results in a division by 0 when computing the velocity. See the console log.
- (No fix, but not important) Because StrokeHandler receives the events in coordinate relative to the page's dpi (right?), the parameter sigma has to be changed if using a different dpi (I think...). I think I made it zoom-independent, but I might have messed that up too.

And more generic TODOs:
- [x] Add parameter selection to the GUI and get rid of the ugly CLI parameter workaround
- [x] Add a version of Krita's deadzone stabilizer (see #2320 for details).
- [x] Implement those smoothing algorithms either as lamdas, function objects, std::functions, Concepts / Interfaces. Do not determine it on every loop cycle.
- [x] Use range based for loops.
- [x] Add a stroke finisher (see [this post](https://github.com/xournalpp/xournalpp/issues/2320#issuecomment-742081126) and the next)
- [x] Remove Glade's default sid1234 names to avoid conflicts with other PR

Known bugs:
[...]

PR #2589, opened 2020-12-23
export layers in different PDF pages

sanette:
Enhancement proposal: see Issue https://github.com/xournalpp/xournalpp/issues/2373

In order to save all layers as separate pages, two options:
1. use "Save as..." in the File menu, select "PDF File" or "PDF File with plain background", and in the "Document export setting" dialog, check the box "Export all layers as separate pages"
2. use the command-line, as in 
    `xournalpp -p output.pdf --export-layers file.xopp`

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -52,7 +52,8 @@ auto migrateSettings() -> MigrateResult;
 void checkForErrorlog();
 void checkForEmergencySave(Control* control);
 
-auto exportPdf(const char* input, const char* output, const char* range, bool noBackground) -> int;
+auto exportPdf(const char* input, const char* output, const char* range, bool noBackground, bool presentationMode)
+        -> int;
 auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
                bool noBackground) -> int;
 
@@ -301,10 +302,13 @@ auto exportImg(const char* input, const char* output, const char* range, int png
  * @param output Path to the output file
  * @param range Page range to be parsed. If range=nullptr, exports the whole file
  * @param noBackground If true, the exported pdf file has white background
+ * @param presentationMode If true, then for each xournalpp page, instead of rendering one PDF page, the page layers are
+ * rendered one by one to produce as many pages as there are layers.
  *
  * @return 0 on success, -2 on failure opening the input file, -3 on export failure
  */
-auto exportPdf(const char* input, const char* output, const char* range, bool noBackground) -> int {
+auto exportPdf(const char* input, const char* output, const char* range, bool noBackground, bool presentationMode)
+        -> int {
     LoadHandler loader;
 
     Document* doc = loader.loadDocument(input);
@@ -327,14 +331,14 @@ auto exportPdf(const char* input, const char* output, const char* range, bool no
         // Parse the range
         PageRangeVector exportRange = PageRange::parse(range, doc->getPageCount());
         // Do the export
-        exportSuccess = pdfe->createPdf(path, exportRange);
+        exportSuccess = pdfe->createPdf(path, exportRange, presentationMode);
         // Clean up
         for (PageRangeEntry* e: exportRange) {
             delete e;
         }
         exportRange.clear();
     } else {
-        exportSuccess = pdfe->createPdf(path);
+        exportSuccess = pdfe->createPdf(path, presentationMode);
     }
 
     if (!exportSuccess) {
@@ -370,7 +374,10 @@ struct XournalMainPrivate {
     int exportPngDpi = -1;
     int exportPngWidth = -1;
     int exportPngHeight = -1;
-    bool exportNoBackground = false;
+    gboolean exportNoBackground =
+            false;  // don't use bool, see
+                    // https://stackoverflow.com/questions/21152042/is-glib-command-line-parsing-order-sensitive
+    gboolean presentationMode = false;
     std::unique_ptr<GladeSearchpath> gladePath;
     std::unique_ptr<Control> control;
     std::unique_ptr<MainWindow> win;
@@ -577,7 +584,7 @@ auto on_handle_local_options(GApplication*, GVariantDict*, XMPtr app_data) -> gi
 
     if (app_data->pdfFilename && app_data->optFilename && *app_data->optFilename) {
         return exportPdf(*app_data->optFilename, app_data->pdfFilename, app_data->exportRange,
-                         app_data->exportNoBackground);
+                         app_data->exportNoBackground, app_data->presentationMode);
     }
     if (app_data->imgFilename && app_data->optFilename && *app_data->optFilename) {
         return exportImg(*app_data->optFilename, app_data->imgFilename, app_data->exportRange, app_data->exportPngDpi,
@@ -630,6 +637,11 @@ auto XournalMain::run(int argc, char** argv) -> int {
                          _("Export without background\n"
                            "                                 The exported file has transparent or white background,\n"
                            "                                 depending on what its format supports\n"),
+                         0},
+            GOptionEntry{"export-presentation", 0, 0, G_OPTION_ARG_NONE, &app_data.presentationMode,
+                         _("Export in presentation mode\n"
+                           "                                 In PDF export, layers are rendered one by one,\n"
+                           "                                 suitable for presentation\n"),
                          0},
             GOptionEntry{"export-range", 0, 0, G_OPTION_ARG_STRING, &app_data.exportRange,
                          _("Only export the pages specified by RANGE (e.g. \"2-3,5,7-\")\n"
--- a/ui/exportSettings.glade
+++ b/ui/exportSettings.glade
@@ -328,6 +328,24 @@ Set export parameters</property>
                     <property name="top-attach">0</property>
                   </packing>
                 </child>
+
+		<child>
+                  <object class="GtkCheckButton" id="cbPresentationMode">
+                    <property name="label" translatable="yes">Export as a presentation</property>
+                    <property name="visible">True</property>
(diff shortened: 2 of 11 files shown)


rolandlo:
@bhennion Could you please give some feedback whether this PR is compatible with [your commit](https://github.com/xournalpp/xournalpp/pull/2512/commits/d48f5b4a376261bc092480818a227b850c3750e5). Or what is your general idea about it?

bhennion:
> @bhennion Could you please give some feedback whether this PR is compatible with [your commit](https://github.com/xournalpp/xournalpp/pull/2512/commits/d48f5b4a376261bc092480818a227b850c3750e5). Or what is your general idea about it?

First, my commit is a half-backed hack. I am not sure the feature it adds has an audience (besides me and my tests). If still you'd like to merge it (once improved), then here is my point of view:

1. I don't think there is a conflict in the code itself.
2. The features follow fairly different philosophies 
      1. my commit selects a range of layers to export, with in mind a single page export. This kinda works for image exports (each page ends up in a separate document anyway), but I fail to see the point of that for multipages exports (e.g. pdf): we can't assume all pages have the same layer structure. (counter-argument: if you stretched a little bit, you could imagine someone willing to export every layer but the first, on every page.)
     2. this PR implements a self contained flag to get a beamer-like presentation. IMO, this only really makes sense when exporting to a multi-pages format (for us pdf).
    
So I would say they are compatible. I would also say that what this PR implements is more useful than what my silly commit does, so this PR takes precedence.

ui/exportSettings.glade
```diff
@@ -327,6 +327,24 @@ Set export parameters</property>
...
                   </packing>
                 </child>
+
+		<child>
+                  <object class="GtkCheckButton" id="cbExportLayers">
+                    <property name="label" translatable="yes">Export all layers as separate pages</property>
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <property name="receives_default">False</property>
+                    <property name="tooltip_text" translatable="yes">If enabled, all layers will be rendered successively as different pages, like for a powerpoint or LaTeX beamer document. The resulting PDF file can be used for a presentation.</property>
```
bhennion, line 337: I understand the label as "each layer will occupy a different page". This is not what the feature does. I know it's hard to give a concise descritpion of this feature. IMO, an undescriptive sentence is better than a misleading one:
```suggestion
                    <property name="label" translatable="yes">Export as a presentation</property>
                    <property name="visible">True</property>
                    <property name="can_focus">True</property>
                    <property name="receives_default">False</property>
                    <property name="tooltip_text" translatable="yes">If enabled, the layers of each page will be added one by one. The resulting PDF file can be used for a presentation.</property>
```

src/control/XournalMain.cpp
```diff
@@ -631,6 +635,11 @@ auto XournalMain::run(int argc, char** argv) -> int {
                            "                                 The exported file has transparent or white background,\n"
                            "                                 depending on what its format supports\n"),
                          0},
+            GOptionEntry{"export-layers", 0, 0, G_OPTION_ARG_NONE, &app_data.exportLayers,
+                         _("Export layers as pages\n"
+                           "                                 Layers are rendered successively as different pages,\n"
+                           "                                 suitable for presentation\n"),
```
bhennion, line 641: I find the description misleading.

bhennion (commented):
This is great work! Thanks!

Maybe you could make the feature descritpions more accurate. I know it's difficult in this case.
I'd suggest giving this feature a proper name (such as "export as a presentation") and changing the variable name `exportLayers` accordingly (e.g. `presentationMode`).

src/control/XournalMain.cpp
sanette: I see what you mean, each time I wrote this I had trouble finding the best way to express it, and this holds for variable names too

sanette:
ok so I followed @bhennion 's suggestion and changed all 'exportLayers' to 'presentationMode'

rolandlo:
```diff
diff --git a/src/pdf/base/XojCairoPdfExport.cpp b/src/pdf/base/XojCairoPdfExport.cpp
index 73c3fbfd..321788f5 100644
--- a/src/pdf/base/XojCairoPdfExport.cpp
+++ b/src/pdf/base/XojCairoPdfExport.cpp
@@ -1,5 +1,6 @@
 #include "XojCairoPdfExport.h"
 
+#include <map>
 #include <sstream>
 #include <stack>
 
@@ -124,24 +125,21 @@ void XojCairoPdfExport::exportPageLayers(size_t page) {
     PageRef p = doc->getPage(page);
 
     // We keep a copy of the layers initial Visible state
-    vector<Layer*>* l = p->getLayers();
-    int nol = l->size();
-    bool initialVisibilities[nol];
-    for (int i = 0; i < nol; i++) {
-        Layer* layer = (*l)[i];
-        initialVisibilities[i] = layer->isVisible();
+    std::map<Layer*, bool> initialVisibility;
+    for (const auto& layer: *p->getLayers()) {
+        initialVisibility[layer] = layer->isVisible();
         layer->setVisible(false);
     }
 
     // We draw as many pages as there are layers. The first page has
     // only Layer 1 visible, the last has all layers visible.
-    for (int i = 0; i < nol; i++) {
-        (*l)[i]->setVisible(true);
+    for (const auto& layer: *p->getLayers()) {
+        layer->setVisible(true);
         exportPage(page);
     }
 
     // We restore the initial visibilities
-    for (int i = 0; i < nol; i++) (*l)[i]->setVisible(initialVisibilities[i]);
+    for (const auto& layer: *p->getLayers()) layer->setVisible(initialVisibility[layer]);
 }
 
 auto XojCairoPdfExport::createPdf(fs::path const& file, PageRangeVector& range, bool presentationMode) -> bool {

```

rolandlo (changes requested):
Please add some margin (12pt) in the export dialog (like in 3452fddacfc43aea3bac242cb1549e35fde73ec4) and use range-based for loops in `XojCairoPdfExport::exportPageLayers` as suggested in the comment. Otherwise your PR looks fine. I have checked that it works. 
After applying the changes please squash your commits and rebase on master. We will probably have to wait for #2435 to get merged.

I noticed that the `clone handwritten layers` function from the `layerActions` plugin will have to be modified to be ready for your `presentationMode` export. This will require some work though (layers will have to be merged) and is not part of this PR.

sanette:
> Please add some margin (12pt) in the export dialog (like in [3452fdd](https://github.com/xournalpp/xournalpp/commit/3452fddacfc43aea3bac242cb1549e35fde73ec4))

I think I already did this in the other PR https://github.com/xournalpp/xournalpp/pull/2592

rolandlo:
> > Please add some margin (12pt) in the export dialog (like in [3452fdd](https://github.com/xournalpp/xournalpp/commit/3452fddacfc43aea3bac242cb1549e35fde73ec4))
> 
> I think I already did this in the other PR #2592

Ok, then you should rebase on current master.

sanette:
> use range-based for loops in `XojCairoPdfExport::exportPageLayers` as suggested in the comment.

ok, that's fine with me. I'm really not fluent in C++, so don't hesitate to correct.
I suppose that accessing a vector with index `i` is faster than looking up a `map`, but since the number of layers is (probably?) very small, this should not make a difference

sanette:
I have incorporated the new loops and hopefully rebased everything on current master

rolandlo (approved):
LGTM

rolandlo:
You may want to hand-tune your commit message though.

rolandlo:
A little help for updating the commit message: Use `git commit --amend` (and `git push --force`) as described [here](https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/changing-a-commit-message)

sanette:
> You may want to hand-tune your commit message though.

Thanks, you make me improve my git abilities ;)

sanette:
> We will probably have to wait for #2435 to get merged.

this is not clear to me, why do we have to wait for #2435?

rolandlo:
> > We will probably have to wait for #2435 to get merged.
> 
> this is not clear to me, why do we have to wait for #2435?

I thought it might create merge conflicts in #2435. But as I see now there are no files in common that have been modified, so I will merge now.

rolandlo:
I have merged it. Thanks for your contribution. Always welcome.

sanette:
> ok so I followed @bhennion 's suggestion and changed all 'exportLayers' to 'presentationMode'

oops I just realized there is already something else called "Presentation Mode" (in the View menu)... maybe this will cause some confusion. And the variable 'presentationMode' is already used! I think we should use another name. Any idea?

mjg:
I'm sorry I (re)started the bike-shedding in #2575. At least I'll take it here where it belongs ... (It is a cool feature btw!)

I understand why @bhennion objected against "layered" originally since layers are not exported individually as pages but as progressive overlays on top of each other. At the same time, "presentationMode" is used for something else already, so I would argue against reusing it.

Maybe *progressive* is a term that conveys both "on top of each other" as well as the dynamic nature (in contrast to the complete stack of layers). So, the suggestion would be:
- "Export layers progressively" as the settings name as well as the cli option short text
- `progressiveMode` instead of `presentationMode`
- `--export-layers-progressively` instead of `--export-presentation`
- "Render layers progressively one by one. This results in N export pages per page with N layers, building up the layer stack progressively." for the settings tooltip as well as the cli option long text, the latter maybe amended by "PDF mode only."

sanette:
Happy New Year ;)
I'm not completely satisfied with "progressive" because it's a bit far from conveying the idea of producing a pdf for a beamer presentation, but I don't have a better idea... So let's do it! The most urgent anyway is to remove "presentationMode".

sanette:
new PR => https://github.com/xournalpp/xournalpp/pull/2609

PR #2590, opened 2020-12-23
Reword "fill transparency" to "fill opacity"

### LINKED ISSUE #2381: Change "trasparency" to "opacity" in pen option -> fill transparency dialog
This is a very, VERY low priority thing...

**Affects versions :**
 - OS: Linux
 - 1.0.19 (and before)

**Describe the bug**
When setting the transparency on fills (which, by the way, should be better discoverable --- maybe a right click on the bucket tool?) you are asked to set "transparency", while really you set "opacity" (when put to 100% the result is solid fill, and 0% means totally transparent).

**To Reproduce**
Just select "Tools->Pen options->Fill transparency"

**Expected behavior**
The option should be called "opacity" or the values should act the other way around ;-)

**Screenshots of Problem**
image

Dettorer:
Fixes #2381

Some things worth noting (and maybe changing) about this PR:
- I didn't use the `generateConvert.php` script to regenerate the enum files, doing so even on master changes the files a lot and breaks the compilation
- my glade version looks pretty different from the one the `.glade` files were last modified with, and it renamed a lot of things, is that ok or should I change the `.glade` files manually (that would be very easy to do for the goal of this PR)


rolandlo (changes requested):
For backwards compatibility could you please accept the `"ACTION_TOOL_FILL_TRNASPARENCY"`(and others) also as read (but not write) values (like in [this commit](https://github.com/xournalpp/xournalpp/pull/2339/commits/f97f8ff69d5d605d70d1c96ec3dfdce7ff7e8d81)? Otherwise one may experience trouble when upgrading (with old settings file). 

Otherwise the code looks good to me. I don't think we need to bother about glade-files changing with newer versions and I also wouldn't use the convertGenerate.php-script (which is a bit outdated, i think). I will try your PR out after you have made the requested change.

rolandlo (changes requested):
Some more comments, after trying it out:
1) Could you please add some margin to the fill opacity dialog (in the glade file), c.f. #2592 
2) I realize that the change from "HILIGHTER..." to "HIGHLIGHTER..." in #2339 didn't preserve backwards compatibility in the `ActionType.Generated.cpp` and `ActionGroup.Generated.cpp` files. This leads to crashes when opening the application with an old settings file.
Could you add the old spellings for "HILIGHTER..." there as well? In the case of `"ACTION_TOOL_HIGHLIGHTER_FILL_OPACITY" there would have to be 3 accepted values in total (add "ACTION_TOOL_HILIGHTER_FILL_TRANSPARENCY").
3) In the end we will have to squash to commits. Do you know how to do that?

Dettorer:
Sure, I'll squash the commits related to the rewording and keep the "HILIGHTER" changes in a separate one.

Dettorer:
@rolandlo, could you share the settings file you used to reproduce the crashes with the "hilighter" spelling? I'm not sure where to look to find all the related options, and I think I can't find any related to opacity/transparency.

rolandlo:
Hmm, I will try to reconstruct it. Maybe it was just a problem with a corrupt settings file I had.

Dettorer:
In the meantime, here is the opacity-related changes.

rolandlo:
Turns out I was wrong about the crash being created by the settings file. It comes from the `main.glade` file. Apparently I had two different versions installed using the same `main.glade` file. That shouldn't happen when one upgrades Xournal++ normally. So I think it is not necessary to create backward compatibility. You can remove the backward compatibility in `ActionType.generated.cpp`

rolandlo (approved):

rolandlo:
This is ready for merge. Do you agree @Dettorer ?

Dettorer:
Yes, I've got nothing to add. I eyeballed the margins of the second commit a bit though, I used the same values as the other PR you mentioned and it looks OK but did you take a look to make sure it was what you expected?

rolandlo:
Yes, it was what I expected. Thanks for your contribution! I will merge.

Issue #1328, opened 2019-07-07
preferences dialog is too wide for my screen


cbm755:
**Affects versions :**
 - OS: Windows, Fedora GNU/Linux]
 - Version of Xournal++ branch `LittleHuba:fix_device_listing_new`, master a37524e2626e269e1c1defbdf5abe3305d29e863

**Describe the bug**
On Windows 10, I have a portrait screen that is 1200 pixels wide.  But I have some big fonts and/or other scaling selected (could check details if needed).  With this setup, the preferences dialog doesn't fit horizontally on the screen (screenshot below).

On my Fedora laptop, I checked the master branch and the minimum size of the preferences dialog is 968 pixels.  But I suspect this depends on OS fontsize preferences etc.

**To Reproduce**
1. select a low res screen (I guess something like 800 pixels wide)
2. open preferences

**Expected behavior**
If the preferences dialog were a bit more compact or was able to shrink a little more then it would fit.  Or the panes would scroll horizontally as they already do vertically.

**Screenshots of Problem**
![pref_dialog_too_wide](https://user-images.githubusercontent.com/818622/60764226-903d6200-a03a-11e9-985e-783ed02d38db.png)

sanette:
I see two potential solutions
1. slightly reorganize the grid of the "Action on Tool Tap" setting, which is the largest element in the Settings.
2. allow horizontal scrolling
I think we can do both, what do you think?

sanette:
by the way, there is a missing GtkScrolledWindow in inputTabBox, I can add it too

rolandlo:
This was fixed in PR #2592

PR #2592, opened 2020-12-24
settings dialog styling

sanette:
Trying to improve the presentation of the Settings dialog by adding a bit of margin,
make the "LaTeX" tab behave more like the other tabs, and make the labels of subsections more recognizable.
Screenshot_20201224_173935
Screenshot_20201224_174005

### DIFF
--- a/src/gui/dialog/SettingsDialog.cpp
+++ b/src/gui/dialog/SettingsDialog.cpp
@@ -166,6 +166,33 @@ void SettingsDialog::setDpi(int dpi) {
 void SettingsDialog::show(GtkWindow* parent) {
     load();
 
+    // detect display size. If large enough, we enlarge the Settings
+    // Window up to 1000x740.
+    GdkDisplay* disp = gdk_display_get_default();
+    if (disp != NULL) {
+        GdkWindow* win = gtk_widget_get_window(GTK_WIDGET(parent));
+        if (win != NULL) {
+            GdkMonitor* moni = gdk_display_get_monitor_at_window(disp, win);
+            GdkRectangle workarea;
+            gdk_monitor_get_workarea(moni, &workarea);
+            gint w = -1;
+            gint h = -1;
+            if (workarea.width > 1100) {
+                w = 1000;
+            } else if (workarea.width > 920) {
+                w = 900;
+            }
+            if (workarea.height > 800) {
+                h = 740;
+            } else if (workarea.height > 620) {
+                h = 600;
+            }
+            gtk_window_set_default_size(GTK_WINDOW(this->window), w, h);
+        } else {
+            g_message("Parent window does not have a GDK Window. This is unexpected.");
+        }
+    }
+
     gtk_window_set_transient_for(GTK_WINDOW(this->window), parent);
     int res = gtk_dialog_run(GTK_DIALOG(this->window));
 
--- a/ui/exportSettings.glade
+++ b/ui/exportSettings.glade
@@ -46,6 +46,7 @@
         <property name="can-focus">False</property>
         <property name="orientation">vertical</property>
         <property name="spacing">4</property>
+	<property name="margin">12</property>
         <child internal-child="action_area">
           <object class="GtkButtonBox" id="dialog-action_area2">
             <property name="name">dialog-action_area2</property>
--- a/ui/latexSettings.glade
+++ b/ui/latexSettings.glade
@@ -11,8 +11,8 @@
   <object class="GtkScrolledWindow" id="latexSettingsPanel">
     <property name="visible">True</property>
     <property name="can_focus">True</property>
-    <property name="shadow_type">in</property>
-    <property name="min_content_height">500</property>
+    <property name="min_content_height">450</property>
+    <property name="min_content_width">500</property>
     <child>
       <object class="GtkViewport">
         <property name="visible">True</property>
@@ -22,11 +22,13 @@
             <property name="visible">True</property>
             <property name="can_focus">False</property>
             <property name="orientation">vertical</property>
+	    <property name="vexpand">True</property>
             <child>
               <object class="GtkFrame">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="label_xalign">0</property>
+                <property name="label_xalign">0.01</property>
+                <property name="margin">12</property>
                 <child>
                   <object class="GtkAlignment">
                     <property name="visible">True</property>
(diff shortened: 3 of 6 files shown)


Technius:
This looks great! Is there anything else you want to add before I merge this?

rolandlo:
Can you do something about #1328?

sanette:
> Is there anything else you want to add before I merge this?

I could add the corrections I made to answer https://github.com/xournalpp/xournalpp/issues/1328#
or if you prefer I can put them in another PR

rolandlo:
> > Is there anything else you want to add before I merge this?
> 
> I could add the corrections I made to answer #1328
> or if you prefer I can put them in another PR

I think it fits well in this PR.

sanette:
I didn't how to do this on the glade file, so I added a command in the cpp file to set the default size to 900x600, after checking that the monitor size is large enough. (otherwise, the min size is something like 600x500, which should fit in any monitor I guess!)

rolandlo:
Some feedback:
- I tested the preferences dialog in your PR at various resolutions (and with different language settings). It's a huge improvement for small screens.
- With screen resolution 896 x 672 it worked still very well.
- With lower screen resolution (like 800 x 600) it gets problematic, since the dialog is too high to show all panels and the "Cancel" and "OK" buttons.
- The 900 x 600 resolution looks optimal with English language. With other languages (like German, which has longer words), it is a bit too narrow. It is not hard to resize the settings panel though. For me this is still fine, although not ideal.
- I am curious to test it out on my Pinephone (with 1440 x 720 resolution). This has been a major headache so far and I guess it will work fine now, which is great.

rolandlo:
On the Pinephone the settings dialog now works very nicely both in landscape and portrait mode.  I get a warning `gdk_monitor_get_workarea: assertion `GDK_IS_MONITOR (monitor) failed` , in the terminal, but it doesn't seem to create problems.

sanette:
thanks for the feedback! 
* concerning very small resolution. Currently the min size for the tab content is 500x500, which means, after adding "Cancel/OK" buttons, the tab selector on the left, and the OS window decorations, something like 715x620 (this will depend on the font size, Desktop theme, etc.). Concerning the vertical size (620), shrinking it down to 600 is no problem with a couple of adjustment. If we wanted to accommodate for even smaller screens, then we would have to put the whole dialog in a ScrolledWindow. But do we really need to adapt to smaller than 600 in height?
* concerning larger screens, we could refine the monitor resolution test and set the width to 1000 when available. This would be better for German, and probably no harm for English

rolandlo:
> thanks for the feedback!
> 
>     * concerning very small resolution. Currently the min size for the tab content is 500x500, which means, after adding "Cancel/OK" buttons, the tab selector on the left, and the OS window decorations, something like 715x620 (this will depend on the font size, Desktop theme, etc.). Concerning the vertical size (620), shrinking it down to 600 is no problem with a couple of adjustment. If we wanted to accommodate for even smaller screens, then we would have to put the whole dialog in a ScrolledWindow. But do we really need to adapt to smaller than 600 in height?
> 

I think screens smaller than 600 pixel in height are very rare these days and I wouldn't add support for it. Even 896 x 672 (which works well) is probably pretty rare.

>     * concerning larger screens, we could refine the monitor resolution test and set the width to 1000 when available. This would be better for German, and probably no harm for English

That sounds good to me.

bhennion:
> I think screens smaller than 600 pixel in height are very rare these days and I wouldn't add support for it. Even 896 x 672 (which works well) is probably pretty rare.

Even if the screen themselves have higher resolutions, modern desktops have a global zoom option (think of HiDpi). Users with bad eyesight may set this zoom to a fairly high value (say 300%). If I'm not mistaken, as fas as GTK is concerned, this would divide the resolution by 3 (in this example). So IMO, very small screen resolutions should be taken into account (say for accessibility), even though actual screens with this resolution barely exist anymore.

sanette:
here is a new version which should be OK for height>=600.
I agree about the HiDPI issue. I myself have a 3200x1800 resolution on my laptop, and in order to get a nicely scaled interface I launch xournalpp with  `GDK_SCALE=2 GDK_DPI_SCALE=0.5`; then the reported size = 1600x865 (it seems to substract the height of the taskbar). If I try `GDK_SCALE=3` I get 1066x576, and the Settings dialog fits nicely.

rolandlo:
In the drawing area panel (Action on Tool Tap) you put everything into a 3x4 grid. This makes the "Ignore time" spin button quite wide (as wide as the checkbox and "Try to select object first" label. This could be improved by letting the checkbox and label expand to the whole rest of the row. That would also save some horizontal space in the drawing area panel.

Setting the screen resolution to 800 x 600 the preferences dialog doesn't fit to screen in my case (the gnome shell panel and title bar take too much space).

sanette:
> This makes the "Ignore time" spin button quite wide (as wide as the checkbox and "Try to select object first" label.

good catch! I've corrected this

sanette:
> Setting the screen resolution to 800 x 600 the preferences dialog doesn't fit to screen in my case (the gnome shell panel and title bar take too much space).

could you tell me more precisely which part is too big? (height or width or both?)

rolandlo:
> > Setting the screen resolution to 800 x 600 the preferences dialog doesn't fit to screen in my case (the gnome shell panel and title bar take too much space).
> 
> could you tell me more precisely which part is too big? (height or width or both?)

Only height. I cannot access the Cancel and OK buttons.

sanette:
> > > Setting the screen resolution to 800 x 600 the preferences dialog doesn't fit to screen in my case (the gnome shell panel and title bar take too much space).
> > 
> > 
> > could you tell me more precisely which part is too big? (height or width or both?)
> 
> Only height. I cannot access the Cancel and OK buttons.

that's weird, on my laptop the whole Dialog including window decoration (KDE) is 574px high

sanette:
the only solution I see is to make the GtkNoteBook scrollable. I just tried, it's not too bad actually.

rolandlo:
That's how it looked for me.

Screenshot_SettingsDialog_800x600

rolandlo:
> the only solution I see is to make the GtkNoteBook scrollable. I just tried, it's not too bad actually.

Ok, I will check

rolandlo:
It looks pretty nice, actually. I like it:
Screenshot_ScrollablePreferences_800x600

rolandlo:
Looks like everything is done, right? Can you squash the commits then, if you feel the same?

sanette:
I'm very bad at git, but somehow it seems that I was able to squash...

rolandlo:
> I'm very bad at git, but somehow it seems that I was able to squash...

Yes, that worked. I will merge then.

rolandlo:
Merged. Thanks for your very valuable contribution. Great work!

sanette:
I'm glad I was able to help. I have been using xournalpp during lockdown for seminars and online courses, and I realized, why not try and improve it! Anyway thanks for making this great app.

PR #2600, opened 2020-12-29
Fix AppImages not running on more recent Linux distros

rolandlo:
The first commit replaces the use of custom AppImage build and run tools by the use of linuxdeploy-plugin-gtk. This simplifies our own task. We can profit from the AppImage community improving this plugin and fixing bugs when new distro releases appear, while contributing back by testing the plugin on various distros with our app.


rolandlo:
The most recent version of the linuxdeploy GTK plugin appears to be broken, see https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/issues/7
Using the older version though seems to fix our AppImage problems. I have created the AppImage directory on Ubuntu 18.04 and ran it successively on Ubuntu 20.10.
It can be customized, but we should try out if the Azure Pipelines really produces a working AppImage using this PR.

azure-pipelines/util/build_appimage.sh
```diff
@@ -11,8 +11,13 @@ if [ -z "$APPDIR" ]; then
     exit 1
 fi
 
-# linuxdeploy location
-LINUXDEPLOY=${LINUXDEPLOY:-"linuxdeploy.AppImage"}
+# get linuxdeploy and linuxdeploy-plugin-gtk
+# Currently broken: wget -c "https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh", see https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/issues/7
+wget -c "https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/c008df652946408f357f502bab67bfcf6f303b4e/linuxdeploy-plugin-gtk.sh"
+wget -c "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage"
```
Technius, line 17: This should check if the file has not been downloaded before trying to download it.

rolandlo:
I had to install `librsvg2-dev` to make the plugin run. I will add it to the dependencies.

azure-pipelines/util/build_appimage.sh
Technius, line 17: The version of linuxdeploy should be set to a specific commit if possible so we don't get random breakages.

Technius (approved):
LGTM, the linuxdeploy GTK plugin is better than what we have right now.

rolandlo:
By the way note that currently the generated AppImage uses the Adwaita icon theme. This could be customized.

azure-pipelines/util/build_appimage.sh
rolandlo: That does not look possible. There is only a continuous release build, see https://github.com/linuxdeploy/linuxdeploy/releases

rolandlo:
@Technius Should we merge already to see, if we get functional AppImages from the Azure Pipelines?

rolandlo:
Need to fix something...

rolandlo:
Done.

Technius:
> Should we merge already to see, if we get functional AppImages from the Azure Pipelines?

I can run this on my private "Release test" pipeline to see if it works correctly.

Technius:
You can find the resulting artifact here when it's done running: https://dev.azure.com/technius/xournalpp/_build/results?buildId=397&view=results

rolandlo:
> You can find the resulting artifact here when it's done running: https://dev.azure.com/technius/xournalpp/_build/results?buildId=397&view=results

The pipeline has run successfully, but I can't see any artifacts.

Technius:
Here's a direct link: https://artprodcus3.artifacts.visualstudio.com/Aa25d95a8-7742-4e55-942f-fe2977ca150d/d07f160b-25db-4823-83e6-50b84810bac2/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3RlY2huaXVzL3Byb2plY3RJZC9kMDdmMTYwYi0yNWRiLTQ4MjMtODNlNi01MGI4NDgxMGJhYzIvYnVpbGRJZC8zOTcvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip

For future reference, the artifacts can be found by clicking one of the two buttons here:
image

You could also try setting up your own "release test" pipeline on your own Azure Pipelines repository using `azure_pipelines/release.yml` as the configuration file.

rolandlo:
> For future reference, the artifacts can be found by clicking one of the two buttons here:

I know of that, but I can't access them (probably missing the right to do so). I can click on the 5 sections as long as I wish, it doesn't show them.
grafik

The direct link you sent me only contains the Windows artifact, not the appimage I wanted to test out. I will try to setup my own "release test" pipeline then.

rolandlo:
Using my own Azure Pipeline I tested the AppImage generated for Ubuntu 18.04 on various distros. It runs successfully on Ubuntu 18.04, 20.04, 20.10 and Debian 10 (Buster). Localizations are missing but other than that I haven't experienced any issues.

LittleHuba:
> Localizations are missing but other than that I haven't experienced any issues.

Any idea why that is the case? Are the files not imported or do we miss localization information from the OS the AppImage is running on?

TheAssassin:
Just for the record, the Gtk plugin has been updated, the issue _should_ be fixed. Contributions welcome, by the way. If you find issues, please don't hesitate to report them (and maybe even send a fix, if you can).

rolandlo:
> > Localizations are missing but other than that I haven't experienced any issues.
> 
> Any idea why that is the case? Are the files not imported or do we miss localization information from the OS the AppImage is running on?

As I understand, the problem is that the files (that are correctly packaged into the AppImage) are not imported. The translation files are looked for in the directory `PACKAGE_LOCALE_DIR` from the file `config-paths.h` generated from `config-paths.h.in`. The latter is defined at configure time (via setting `DCMAKE_INSTALL_PREFIX` to which `share/locale` gets appended). In the `release_ubuntu.yml` file we set `-DCMAKE_INSTALL_PREFIX=$PWD/staging`. There is no way the AppImage should know where to look for the translation files (other than `$PWD/staging/share/locale` which is wrong).

To make it clear: The AppImage shouldn't install (or expect) any translation files for Xournal++ on the system. It should use the ones that are packaged. I haven't figured out how to do it.

Technius:
> As I understand, the problem is that the files (that are correctly packaged into the AppImage) are not imported. The translation files are looked for in the directory `PACKAGE_LOCALE_DIR` from the file `config-paths.h` generated from `config-paths.h.in`. The latter is defined at configure time (via setting `DCMAKE_INSTALL_PREFIX` to which `share/locale` gets appended). In the `release_ubuntu.yml` file we set `-DCMAKE_INSTALL_PREFIX=$PWD/staging`.

We could instead modify the application startup to load this path from an environment variable at runtime, if the environment variable is defined (defaults to this configure time path otherwise). Then maybe we could set the environment variable to point to the appropriate directory when the AppImage is run.

TheAssassin:
Do you use `gettext`? https://github.com/linuxdeploy/linuxdeploy/pull/160, dedicated plugin on the way.

rolandlo:
> Do you use `gettext`? [linuxdeploy/linuxdeploy#160](https://github.com/linuxdeploy/linuxdeploy/pull/160), dedicated plugin on the way.

Yes, we do use `gettext`.

TheAssassin:
See https://github.com/linuxdeploy/misc-plugins/blob/master/gettext/linuxdeploy-plugin-gettext.sh. Should help with your use case. Contributions welcome!

rolandlo:
> See https://github.com/linuxdeploy/misc-plugins/blob/master/gettext/linuxdeploy-plugin-gettext.sh. Should help with your use case. Contributions welcome!

Thanks! Looks like what we need. I will check it out.

rolandlo:
With the latest PR Xournal++ will use the environment variable TEXTDOMAINDIR (if it is set) for the gettext file path. The AppImage has correct localizations now.

X0rg:
@rolandlo I updated the linuxdeploy-plugin-gettext.sh in linuxdeploy/misc-plugins@e757980469bc9f3857152376cc4126caee65fe7a, I think this plugin was not working as expected.

rolandlo:
> @rolandlo I updated the linuxdeploy-plugin-gettext.sh in [linuxdeploy/misc-plugins@e757980](https://github.com/linuxdeploy/misc-plugins/commit/e757980469bc9f3857152376cc4126caee65fe7a), I think this plugin was not working as expected.

It actually worked as expected, the variable was correctly exported. Hopefully it still does.

TheAssassin:
Yep, it'll still work, from what I can see.

@X0rg hooks are `source`d, not called. If they were called, the `export` would never propagate into the calling script. That's why you don't need a shebang. It's harmless, though: it will be taken as a comment and therefore be ignored.

rolandlo:
Anyway, I checked that the AppImage produced with the latest plugin(s) runs fine on Ubuntu 20.10, 20.04, 18.04 and Debian 10, all with correct localization. Thanks @TheAssassin and @X0rg for making our lives much easier than it would be without the plugins. 

@Technius Can you please take another look at it? Should I remove the `g_message` that I introduced in `PathUtil.cpp` to display the paths involved?

X0rg:
BTW, if something is missing for your AppImage, feel free to add stuff to https://github.com/linuxdeploy/misc-plugins. :wink: 

> @X0rg hooks are `source`d, not called. If they were called, the `export` would never propagate into the calling script. That's why you don't need a shebang. It's harmless, though: it will be taken as a comment and therefore be ignored.

@TheAssassin Ah, my bad. linuxdeploy-plugin-gstreamer.sh and linuxdeploy-plugin-gtk.sh use shebang and `$APPDIR`, which made me think of a mistake. But the shebang can be useful for shellcheck, it avoids to impose a shell dialect for hooks.

azure-pipelines/util/build_appimage.sh
pktiuk: I think it is recommended to have some backup of this tool just in case, because it may break one day.  
Example: https://github.com/linuxdeploy/linuxdeploy/issues/143.  
You can use older binaries from link in that issue: https://artifacts.assassinate-you.net//linuxdeploy/

TheAssassin: Please note that the artifacts storage is making no guarantees about its stability. For instance, I've recently changed from (the annoying) Artifactory to a plain HTTP server. If you are afraid that linuxdeploy is going to go down, you need to host your own binaries.

rolandlo:
The `g_message` can be removed later, if it is disturbing. I will merge now.

Issue #2515, opened 2020-12-09
High to extreme memory usage an d when zooming


felurx:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Manjaro (based on Arch)
 - (Linux only) Desktop environment: X11 (apparently)
 - Which version of libgtk do you use: 3.24.23
 - Version of Xournal++: 1.0.19
 - Installation method: sudo pacman -S xournalpp
**Describe the bug**
Zooming into documents 

**To Reproduce**
1. Step 1
2. Step 2
3. etc.

**Expected behavior**

**Screenshots of Problem**
(Optional) If applicable, add screenshots to help explain your problem.

**Additional context**
Add any other information or context about the problem here.

felurx:
Dang, accidentally submitted it without completing the text. Sorry.

personalizedrefrigerator:
#2435 doesn't fix it:
![image](https://user-images.githubusercontent.com/46334387/101650827-3d1acc00-39f1-11eb-93a1-f1e94abc127e.png)

felurx:
> Unfortunately, #2435 doesn't fix it:
Sorry for the misunderstanding, I'm just hoping for it to fix the rapid zooming, not the memory issues.

Technius:
This is due to how to a combination of problems:
* The page size is multiplied by the zoom factor, which means a 3000x6000 document will use 4x as many pixels with 2x zoom. 
* Zooming will trigger a relayout of the _whole_ document, not just the nearby or visible pages.
* No lazy loading, so everything is kept in memory.

We can address this as part of the viewport refactor (#1795).

Explorare:
I did some tests just now and can confirm there are mainly two issues here.
1. The memory usage per page is related to the rendering size. The memory used per page grows 4x with a 2x zoom, which does make sense.
2.  The option "Clear cached paged while scrolling." under View > Performance didn't work. The memory usage will keep growing and capped at 1400MB even with pre-load all set to `1`. But, the memory will be released soon after a re-render caused by zoom. 
3. By the way, scanned and compressed PDF documents (like djvu) rendering is very slow, uses a lot RAM and CPU.

Here are some details:
_[Calculus Vol 1](https://assets.openstax.org/oscms-prodcms/media/documents/Calculus_Volume_1_-_WEB_68M1Z5W.pdf)_ by [Openstax](https://openstax.org/details/books/calculus-volume-1) 
[...]

PR #2603, opened 2020-12-29
Rewrite and improve page load caching algorithm

### LINKED ISSUE #2412: Preemptive load of pages around the current one
(This is taken from #1477, which describes this problem perfectly)
When annotating a PDF and switching between pages, the entire page is momentarily replaced by a big white screen with "Loading...".

I am using xournalpp to record video lectures where I advance through PDF slides and annotate them with the pen tools. This loading screen issue is particularly distracting because my slides have black background and the "Loading" screen is white.

I wonder if @michaelcadilhac 's code at the bottom of #1477 ([here](https://github.com/xournalpp/xournalpp/issues/1477#issuecomment-663921991)) could be considered for integration. The code basically preloads a given number of pages pages around the current one(s).

Technius:
This PR updates the page cache algorithm and adds a page preloading feature. Fixes #2412.

TODO:
* [x] Configuration option for `n` pages to preload before current page
* [x] Configuration option for `n` pages to preload after current page
* [ ] Configuration option for `n` additional pages to keep in LRU cache
* [x] Address performance concerns when rapidly scrolling through pages (pages are loaded eagerly, so they consume a lot of memory).

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -139,6 +139,9 @@ void Settings::loadDefault() {
 
     this->pageRerenderThreshold = 5.0;
     this->pdfPageCacheSize = 10;
+    this->preloadPagesBefore = 3U;
+    this->preloadPagesAfter = 5U;
+    this->eagerPageCleanup = true;
 
     this->selectionBorderColor = 0xff0000U;  // red
     this->selectionMarkerColor = 0x729fcfU;  // light blue
@@ -398,6 +401,12 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         this->pageRerenderThreshold = g_ascii_strtod(reinterpret_cast<const char*>(value), nullptr);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("pdfPageCacheSize")) == 0) {
         this->pdfPageCacheSize = g_ascii_strtoll(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("preloadPagesBefore")) == 0) {
+        this->preloadPagesBefore = g_ascii_strtoull(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("preloadPagesAfter")) == 0) {
+        this->preloadPagesAfter = g_ascii_strtoull(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("eagerPageCleanup")) == 0) {
+        this->eagerPageCleanup = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("selectionBorderColor")) == 0) {
         this->selectionBorderColor = Color(g_ascii_strtoull(reinterpret_cast<const char*>(value), nullptr, 10));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("selectionMarkerColor")) == 0) {
@@ -843,6 +852,9 @@ void Settings::save() {
 
     WRITE_INT_PROP(pdfPageCacheSize);
     WRITE_COMMENT("The count of rendered PDF pages which will be cached.");
+    WRITE_UINT_PROP(preloadPagesBefore);
+    WRITE_UINT_PROP(preloadPagesAfter);
+    WRITE_BOOL_PROP(eagerPageCleanup);
 
     WRITE_COMMENT("Config for new pages");
     WRITE_STRING_PROP(pageTemplate);
@@ -1621,6 +1633,36 @@ void Settings::setPdfPageCacheSize(int size) {
         return;
     }
     this->pdfPageCacheSize = size;
+    save();
+}
+
+auto Settings::getPreloadPagesBefore() const -> unsigned int { return this->preloadPagesBefore; }
+
+void Settings::setPreloadPagesBefore(unsigned int n) {
+    if (this->preloadPagesBefore == n) {
+        return;
+    }
+    this->preloadPagesBefore = n;
+    save();
+}
+
+auto Settings::getPreloadPagesAfter() const -> unsigned int { return this->preloadPagesAfter; }
+
+void Settings::setPreloadPagesAfter(unsigned int n) {
+    if (this->preloadPagesAfter == n) {
+        return;
+    }
+    this->preloadPagesAfter = n;
+    save();
+}
+
+auto Settings::isEagerPageCleanup() const -> bool { return this->eagerPageCleanup; }
+
+void Settings::setEagerPageCleanup(bool b) {
+    if (this->eagerPageCleanup == b) {
+        return;
+    }
+    this->eagerPageCleanup = b;
     save();
 }
 
--- a/src/control/settings/Settings.h
+++ b/src/control/settings/Settings.h
@@ -330,6 +330,15 @@ public:
     int getPdfPageCacheSize() const;
     [[maybe_unused]] void setPdfPageCacheSize(int size);
 
+    unsigned int getPreloadPagesBefore() const;
+    void setPreloadPagesBefore(unsigned int n);
+
+    unsigned int getPreloadPagesAfter() const;
+    void setPreloadPagesAfter(unsigned int n);
+
+    bool isEagerPageCleanup() const;
+    void setEagerPageCleanup(bool b);
+
     string const& getPageTemplate() const;
     void setPageTemplate(const string& pageTemplate);
 
@@ -937,4 +946,18 @@ private:
      * e.g. "en_US"
      */
     std::string preferredLocale;
+    /**
+     * The number of pages to pre-load before the current page.
+     */
+    unsigned int preloadPagesBefore{};
+
+    /**
+     * The number of pages to pre-load after the current page.
+     */
+    unsigned int preloadPagesAfter{};
+
+    /**
+     * Whether to evict from the page buffer cache when scrolling.
+     */
+    bool eagerPageCleanup{};
 };
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -1,5 +1,6 @@
 #include "XournalView.h"
 
+#include <algorithm>
 #include <cmath>
 #include <memory>
 #include <tuple>
@@ -23,6 +24,14 @@
 #include "Util.h"
 #include "XournalppCursor.h"
 #include "filesystem.h"
+
+std::pair<size_t, size_t> XournalView::preloadPageBounds(size_t page, size_t maxPage) {
+    const size_t preloadBefore = this->control->getSettings()->getPreloadPagesBefore();
+    const size_t preloadAfter = this->control->getSettings()->getPreloadPagesAfter();
+    const size_t lower = page > preloadBefore ? page - preloadBefore : 0;
+    const size_t upper = std::min(maxPage, page + preloadAfter);
+    return {lower, upper};
+}
 
 XournalView::XournalView(GtkWidget* parent, Control* control, ScrollHandling* scrollHandling):
         scrollHandling(scrollHandling), control(control) {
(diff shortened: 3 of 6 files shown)


Technius:
We probably don't need to do anything fancy for the scrolling performance. I monitored the performance of scrolling through a large textbook PDF (~35MB) with PgDown held down: CPU & memory usage was slightly higher than Okular and Firefox (PDF.js). However, I have 32 GB of memory so we will need to ensure that it won't cause crashes when only "old laptop amounts" of ram (e.g. 4GB) is available (very easy to trigger, see #2515).

Technius:
The cache is cleaned up when a new page is selected. This should have an unnoticeable performance impact and will save a lot of memory usage when scrolling, but I have left it as an option just in case.

Technius:
I'm not really convinced we need to keep additional pages in cache (outside of the visible/preloaded pages) anyways--random access is probably a relatively infrequent scenario. Merging this in 48 hours unless there are objections.

rolandlo:
Two remarks:
- The "Cache extra pages setting" should be removed from the glade file if it is not used, I guess.
- Have you measured the memory consumption in any way? I have checked it with the System Monitor (3.36.0). After startup Xournal++ consumes around 23.4 MB of memory. After I load up a large pdf (15.2 MB in size; 55 pages) with annotations (1.5 MB in size), it rises to roughly 54 MB (sometimes more). If I close the document (by clicking on Edit -> New) Xournal++ still takes up the same 54 MB of memory. If I step through all the pages (by pressing page down about once each second), the memory consumption rises to something like 180 MB (or more). If I close the document again (by clicking on Edit -> New) the occupied memory either stays at the same level or goes back slightly, but it still remains rather high. There must be something wrong, I think.

Technius:
> The "Cache extra pages setting" should be removed from the glade file if it is not used, I guess.

Yes, you're right.

> Have you measured the memory consumption in any way?

Nothing too concrete, I just monitored `htop`'s memory metric while scrolling through a document. But your test suggests that we might need to clear the cache when we close / load the document.

> If I step through all the pages (by pressing page down about once each second), the memory consumption rises to something like 180 MB (or more).

The memory consumption should definitely increase as you are stepping through, but it will be bounded by the cache size (= visible pages + preload pages). The metrics don't track actual memory consumption though, only page count.

src/control/settings/Settings.h
```diff
@@ -887,4 +896,18 @@ class Settings {
      * e.g. "en_US"
      */
     std::string preferredLocale;
+    /**
+     * The number of pages to pre-load before the current page.
+     */
+    size_t preloadPagesBefore{};
```
LittleHuba, line 902: `size_t` seems to be a little bit overkill here 😉

src/control/settings/Settings.h
```diff
@@ -887,4 +896,18 @@ class Settings {
...
+
+    /**
+     * The number of pages to pre-load after the current page.
+     */
+    size_t preloadPagesAfter{};
```
LittleHuba, line 907: `size_t` seems to be a little bit overkill here 😉

src/gui/XournalView.cpp
```diff
@@ -86,40 +95,38 @@ void XournalView::staticLayoutPages(GtkWidget* widget, GtkAllocation* allocation
...
+    constexpr auto cmp = [](XojPageView* a, XojPageView* b) {
+        return a->getLastVisibleTime() < b->getLastVisibleTime();
+    };
+    std::vector<XojPageView*> toRemove;
+    // Keep in cache: current page + preload_before + preload_after + LRU capacity
+    for (size_t i = 0; i < this->viewPages.size(); i++) {
+        auto&& page = this->viewPages[i];
+        size_t pageNum = i + 1;
+        bool isPreload = pagesLower <= pageNum && pageNum <= pagesUpper;
+        if (page->getLastVisibleTime() > 0 && !isPreload) {
```
LittleHuba, line 118: Flip this for faster comparison. If the first check is already wrong, the second check will not be performed in a `&&`. `!isPreload` is one cycle, compared to a few hundred for accessing the pointer of `page->getLastVisibleTime()`.

src/gui/XournalView.cpp
```diff
@@ -24,6 +25,14 @@
 #include "XournalppCursor.h"
 #include "filesystem.h"
 
+std::pair<size_t, size_t> XournalView::preloadPageBounds(size_t page, size_t maxPage) {
+    static const size_t preloadBefore = this->control->getSettings()->getPreloadPagesBefore();
+    static const size_t preloadAfter = this->control->getSettings()->getPreloadPagesAfter();
+    const size_t lower = page > preloadBefore ? page - preloadBefore : 1;
```
LittleHuba, line 31: Do page numbers really start with 1? Awkward... In IT all enumerations normally start at 0.

src/gui/XournalView.cpp
```diff
@@ -86,40 +95,38 @@ void XournalView::staticLayoutPages(GtkWidget* widget, GtkAllocation* allocation
...
-            auto* v = static_cast<XojPageView*>(l->data);
-
-            if (pixel <= 0) {
-                v->deleteViewBuffer();
-            } else {
-                pixel -= v->getBufferPixels();
-            }
+    // Remove oldest pages exceeding (LRU cache capacity)
+    const size_t cacheCapacity = 0;
+    if (toRemove.size() > cacheCapacity) {
```
LittleHuba, line 125: This does not make sense. Either you have a cache or you don't but this is some kind of really awkward mix.

src/gui/XournalView.cpp
```diff
@@ -377,6 +384,19 @@ void XournalView::pageSelected(size_t page) {
     control->updatePageNumbers(currentPage, pdfPage);
 
     control->updateBackgroundSizeButton();
+
+    if (control->getSettings()->isEagerPageCleanup()) {
+        this->cleanupBufferCache();
```
LittleHuba, line 389: Please do not do this here. `getCurrentPage` is a method often used for input handling and should therefore be fast. What you do here is attaching a huge time penalty to each and every call of this method. Instead, introduce a background job that does this regularly.

src/gui/XournalView.cpp
```diff
@@ -377,6 +384,19 @@ void XournalView::pageSelected(size_t page) {
...
+    }
+
+    // Load surrounding pages if they are not
+    const auto& [pagesLower, pagesUpper] = preloadPageBounds(page, this->viewPages.size());
+    g_assert(pagesLower <= pagesUpper);
+    for (size_t i = pagesLower; i < pagesUpper; i++) {
+        if (this->viewPages[i]->getBufferPixels() == 0) {
+            this->viewPages[i]->rerenderPage();
+        }
+    }
```
LittleHuba, line 399: Do also not do this here. Same reason as above. Instead put this into scroll handling.

LittleHuba (changes requested):
Your PR is an interesting one. You introduce a cache system to speed up rendering. The way you integrated it though slows down nearly all other code. Please change the points where you preload pages to reduce the negative impact of caching those pages.

LittleHuba:
> Nothing too concrete, I just monitored htop's memory metric while scrolling through a document. But your test suggests that we might need to clear the cache when we close / load the document.

Yes this is definitely missing. Your current system relies on going through all available pages and clearing the ones not seen for longer time. If you close a document you can not iterate through those pages anymore and leak them accidently.

src/gui/XournalView.cpp
```diff
@@ -24,6 +25,14 @@
 #include "XournalppCursor.h"
 #include "filesystem.h"
 
+std::pair<size_t, size_t> XournalView::preloadPageBounds(size_t page, size_t maxPage) {
+    static const size_t preloadBefore = this->control->getSettings()->getPreloadPagesBefore();
+    static const size_t preloadAfter = this->control->getSettings()->getPreloadPagesAfter();
```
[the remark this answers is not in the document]
Technius: This was a typo from when I was prototyping this (when I just had them as constants in the file, not in the settings). Thanks for the catch.

src/gui/XournalView.cpp
Technius: What does this have to do with `getCurrentPage`? Is this code not in `pageSelected` or am I missing something?

src/gui/XournalView.cpp
Technius: This should've been named `extraCacheCapacity`, but I was going to remove this entire snippet anyways.

src/gui/XournalView.cpp
Technius: Additionally, the enclosing function is doing file I/O which will take _far_ more cycles than garbage collecting the cache. I'm not sure whether this is going to have much effect in the grand scheme of things.

src/gui/XournalView.cpp
Technius: Oops, that's a typo.

src/gui/XournalView.cpp
Technius: The timer is insufficient for handling garbage collection, because rapidly scrolling through pages can cause entire RAM + swap to be consumed before the cache is GC'd.

ui/settings.glade
```diff
@@ -43,6 +43,16 @@
...
+  <object class="GtkAdjustment" id="adjustmentPreloadPagesAfter">
+    <property name="upper">99</property>
+    <property name="step-increment">1</property>
+    <property name="page-increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustmentPreloadPagesBefore">
+    <property name="upper">99</property>
+    <property name="step-increment">1</property>
+    <property name="page-increment">10</property>
+  </object>
```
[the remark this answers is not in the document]
Technius: Glade does not output the `lower` property when it is set to 0. GTK probably assumes it to be `0` when it is missing.

Technius:
> You introduce a cache system to speed up rendering.

I did not introduce anything--this repurposes the existing system so that it can load pages in advance to prevent noticeable flickering / stuttering.

> If you close a document you can not iterate through those pages anymore and leak them accidently.
 
I double-checked the code: render buffers are already deleted when clearing pages, so the memory is going somewhere else. I wonder if this is an actual memory leak or if it's simply some overhead somewhere.

---

`htop`reports that with 1 preload page before and 1 preload page after, memory usage was 405 MB (physical) when scrolling through a large textbook. Stepping up to the default of 3 before / 5 after increased usage to 455 MB (physical).

In comparison, without this patch, scrolling consumes 2.8 GB (physical) and settles down at 1.6 GB a few seconds after scrolling finishes.

src/gui/XournalView.cpp
LittleHuba: > What does this have to do with getCurrentPage? Is this code not in pageSelected or am I missing something?

Sorry my bad... Github folded the code with the header of getCurrentPage right above the fold and I didn't see the other method signature. In that case this is okay.

src/gui/XournalView.cpp
```diff
@@ -86,40 +95,34 @@ void XournalView::staticLayoutPages(GtkWidget* widget, GtkAllocation* allocation
...
+        return a->getLastVisibleTime() < b->getLastVisibleTime();
+    };
+    std::vector<XojPageView*> toRemove;
+    // Keep in cache: visible pages + preload_before + preload_after
+    for (size_t i = 0; i < this->viewPages.size(); i++) {
+        auto&& page = this->viewPages[i];
+        const size_t pageNum = i + 1;
+        const bool isPreload = pagesLower <= pageNum && pageNum <= pagesUpper;
+        if (!isPreload && page->getLastVisibleTime() > 0) {
+            toRemove.insert(std::upper_bound(toRemove.begin(), toRemove.end(), page, cmp), page);
```
LittleHuba, line 119: As you clear the full list, there is no need to sort the pages anymore.

LittleHuba (commented):
Sorry for the confusion. The previous change requests were an oversight by me. Just one further optimization to point out, resulting from your simplification of the algorithm. Then I think this is good to go.

LittleHuba:
> I double-checked the code: render buffers are already deleted when clearing pages, so the memory is going somewhere else. I wonder if this is an actual memory leak or if it's simply some overhead somewhere.

Unlikely that it is just some overhead somewhere, as it should always be freed once the document is closed.

src/gui/XournalView.cpp
Technius: Fixed. Now it just deletes the buffers in the loop.

Technius:
I found an additional problem: the pages are not preloaded asynchronously, so there's some stuttering during scrolling.

Technius:
Nevermind, that seems to be a different page size / scrolling bug. I think this PR is good to go; merging in 48 hours if there are no more comments.

rolandlo:
Still, if I open a document and then close it (via menu File -> New) again, I would expect that the amount of memory consumed is again roughly the same as before I opened the document. This is not the case. 

To reproduce:
1) Start Xournal++ and observe the amount of memory consumed by it
2) Open a (large) document
3) Press Menu File -> New
4) Observe the amount of memory consumed and notice that it is much higher than it was in the beginning.

Even simpler:
1) Start Xournal++ and observe the amount of memory consumed by it
2) Press Menu File -> New
3) Observe the amount of memory consumed and notice that it is higher than it was in the beginning.

Technius:
> Still, if I open a document and then close it (via menu File -> New) again, I would expect that the amount of memory consumed is again roughly the same as before I opened the document. This is not the case.

This memory consumption issue is also present on `master` and is therefore irrelevant to this PR. Please file another issue.

rolandlo:
Ok, I will.

src/control/settings/Settings.cpp
```diff
@@ -382,6 +385,12 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         this->presentationHideElements = reinterpret_cast<const char*>(value);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("pdfPageCacheSize")) == 0) {
         this->pdfPageCacheSize = g_ascii_strtoll(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("preloadPagesBefore")) == 0) {
+        this->preloadPagesBefore = g_ascii_strtoull(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("preloadPagesAfter")) == 0) {
+        this->preloadPagesAfter = g_ascii_strtoull(reinterpret_cast<const char*>(value), nullptr, 10);
+    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("eagerPageCleanup")) == 0) {
```
Febbe, line 392:
```suggestion
        this->preloadPagesBefore = g_ascii_strtoul(reinterpret_cast<const char*>(value), nullptr, 10);
    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("preloadPagesAfter")) == 0) {
        this->preloadPagesAfter = g_ascii_strtoul(reinterpret_cast<const char*>(value), nullptr, 10);
    } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("eagerPageCleanup")) == 0) {
```

Technius: This function doesn't exist, and it's unlikely for a document to have more than `2^32 - 1` pages in practice.

Technius:
Rebased on master. Merging in 24 hours.

Issue #2585, opened 2020-12-23
Add Xournal to the "Open With" context menu on Windows


LCarmi:
**Is your feature request related to a problem? Please describe.**
On Windows 10, if Xournal is not your default pdf editor and but you want to open a file with Xournal, this actions are needed:
Right click on file -> Open With -> Choose another app -> xournalpp.exe
Many common applications are shown as suggestions in the "Right click on file -> Open With " menu, which is much more useful from a UX point of view

**Describe the solution you'd like**
I looked forward to see how to add Xournal to that menu.

While debugging the Registry Entries, I found out this solution:
[https://superuser.com/questions/1199648/strange-open-with-list-inconsistent-with-registry/1256482#1256482](https://superuser.com/questions/1199648/strange-open-with-list-inconsistent-with-registry/1256482#1256482)

By adding an entry ending in .ApplicationCompany as suggested by the linked resource, i fixed the problem. 
However it is to be noted that this fix is only temporary because it seems that Windows 10 refreshes the MuiCache, and that probably such a fix should be done during the app registration in Windows.

Unfortunately I'm no expert of windows installation procedures and so on, but still I'm sharing this usability improvement if useful to be integrated in future releases.

Moreover, the .FriendlyAppName can be changed to "Xournal++" instead of "xournalpp.exe"

Technius:
Anyone interested in contributing, this can be done by editing the the NSIS install script in `windows-setup/xournalpp.nsi`. The windows install script actually needs an overhaul in my opinion...

rolandlo:
> Anyone interested in contributing, this can be done by editing the the NSIS install script in `windows-setup/xournalpp.nsi`. The windows install script actually needs an overhaul in my opinion...

I agree. In particular the plugin folder and resource folder should get installed.

Technius:
I installed a fresh version of Xournal++ and I got a `Edit Xournal++ file` menu option when right clicking an xopp file. Do you not have that option?

DUOLabs333:
@Technius I think LCarmi is referring to opening PDFs in Xournal++.

Technius:
This has been fixed in #2606.

PR #2606, opened 2020-12-30
Update windows installer

rolandlo:
- [x] Install plugins and resource folder
- [x] Generate Windows icon file xournalpp.ico automatically
- [x] Add Xournal++ to the "Open With" Context for PDF files (#2585)
- [x] Remove the launcher
- [x] Test everything carefully


Febbe:
Can we get rid of the launcher? I would like to have only one executable.

Technius:
The launcher should be safe to remove. It was needed previously because some paths were hardcoded into the application, but ever since the installation became relocatable, the launcher became unnecessary.

Technius:
I found some time today to set up a Windows VM and change some of the Windows build settings. Would you like me to commit the changes to this PR directly or open a new one?

rolandlo:
@Technius Yes, please submit your changes to this PR directly. Yes, the ImageMagick convert doesn't work yet.

Technius:
See the commit message for details on what I did. I updated the build script to use the CMake install target as a base, so it just needs to copy the libraries and runtime dependencies. Tested on my Windows VM, but I could've missed some things when I copied my patch over to my Linux installation.

I think we can safely remove the launcher at this point.

src/CMakeLists.txt
```diff
@@ -69,6 +69,28 @@ file (GLOB xournalpp_SOURCES_RECURSE
   pdf/popplerapi/*.cpp
 )
 
+if (WIN32)
```
Technius, line 72: This code for generating the icons could probably be combined into the one below.

windows-setup/package.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+## Windows Setup script
```
Technius, line 3: I forgot to update this, will do it later.

src/CMakeLists.txt
```diff
@@ -69,6 +69,28 @@ file (GLOB xournalpp_SOURCES_RECURSE
...
+  # generate .ico from .png
+  set (ICON_SIZES 16 32 48 256)
+  foreach (SIZE IN LISTS ICON_SIZES)
+    set (RESIZED_ICON ${PROJECT_BINARY_DIR}/xournalpp-icon-${SIZE}.png)
+    add_custom_command (
+	  OUTPUT ${RESIZED_ICON}
+	  COMMAND rsvg-convert ${PROJECT_SOURCE_DIR}/ui/pixmaps/com.github.xournalpp.xournalpp.svg -o ${RESIZED_ICON}
```
rolandlo, line 79: You should specify width and height here. Otherwise all icons are identic, 128x128 pixel.

rolandlo:
I have tested the version from my last commit (without launcher and with the .ico file created from .png files with different resolutions) and it worked fine. The icon was generated correctly and Xournal++ ran normally, no problem with pressure sensitivity of the pen. Icons are displayed correctly.

src/CMakeLists.txt
rolandlo: I addressed this.

src/CMakeLists.txt
rolandlo: I fixed this.

rolandlo:
The last commit (about opening pdf-files via the "Open With" context-menu) doesn't quite work as I hoped for. It must probably be reverted or at least modified.

Technius:
Xournal++ is now an entry in the `Open With...` popup. See the commit description for the full changelog.

One thing we are missing is version information in the installer. We will need to insert the version info into the `src/win32/xpp.rc` file during the CMake configuration.

Technius:
Last thing we need to do is test the installer.

@rolandlo Can you do a release test to ensure that the installer will build correctly?

rolandlo:
@Technius Sure, I will do the testing.

rolandlo:
Some comments from testing:
- The release pipeline now generates the installer corrrectly (I had to update the name of the script in `release.yml`)
- The version is correctly displayed in the installer
- The uninstaller works now
- The new icon is displayed everywhere except in the start menu, where I still get the old icon displayed
- Attaching to the the taskbar works fine
- "Open with" works now
- The folder in which Xournal++ is executed (via the start menu link) is wrong. It should be executed in the `bin` subfolder. Otherwise localizations are not detected and the language cannot be changed to anything different than English
- The folder structure has changed (as far as I can see). When installing with the new installer into the same folder, some old subfolders are left (in particular the `ui` folder is left in the `Xournal++` directory)
- The LaTeX settings panel does not open the right folder (at first) when attempting to change the template
- I have only tested "Install for all users"

Technius:
> The new icon is displayed everywhere except in the start menu, where I still get the old icon displayed

I noticed this too, except I didn't get any icon at all. I haven't figured out if this is a problem due to the legacy install or if this is due to a missing registry key, etc.

Technius:
I made some more fixes (see commit message for detailed changes). Installing on top of the legacy version should generally work now. I have not fixed the localization issue though, and would prefer putting it off into a follow-up PR since this issue is blocking a number of other PRs / potential fixes.

> The new icon is displayed everywhere except in the start menu, where I still get the old icon displayed

This has been fixed now. Part of the problem was due to the old start menu entries not being removed, another part was due to the `SHChangeNotify` issue I pointed out above.

> The LaTeX settings panel does not open the right folder (at first) when attempting to change the template

Not sure why this is happening.

> I have only tested "Install for all users"

We should probably disable or delete the "Install for current user" setting. I currently don't have the time/intention to support it.

Technius:
@rolandlo Let's get this merged in, say, 48 hours. Do you have any remaining concerns / tests you want to do?

rolandlo:
@Technius I have just tested it and the icon issue is fixed. Great! The LaTeX template path also worked. Maybe I had a bad path in my settings file before. I am fine with disabling or deleting the "Install for current user" setting.
I don't have any remaining concerns.

rolandlo:
I have also checked once more that the PR builds successfully on Azure Pipelines. It looks all good.

Technius:
I'll disable the "install for current user settings" tomorrow and merge this PR afterwards.

Issue #2363, opened 2020-10-31
Allow overwriting background PDF files


pktiuk:
**Describe the solution you'd like**
It would be good to allow users overwriting background PDF files (with proper warning about impossibility of further editing and cleaning/removing `.xopp` file).

**Describe alternatives you've considered**
Exporting into another pdf file and removing the previous file after ending work with file is a messy solution.

**Additional context**
My workflow during working with PDF-s at university is like this, during lecture I am opening presentation (in form of pdf) provided by the teacher and put notes and valuable comments during lecture, I am saving file (and I am not going to work with it using xournal++ anymore), later these files are synchronized with cloud and I would like to have an access to annotated versions on devices without xournal++.

Technius:
> I would like to have an access to annotated versions on devices without xournal++

It seems to me that the core issue here is having some sort of auto-export feature. Maybe this could be done with a plugin.

pktiuk:
For me, it is just annoying that app doesn't let me to overwrite this file, even if I am aware of its consequences.

I think the simplest solution would be adding here button "force overwritting" (or something similar) and ensure clearing/deleting old `.xopp` file after this.
![image](https://user-images.githubusercontent.com/45544416/98166957-94a4b400-1ee8-11eb-873b-1e4b733fc978.png)

Technius:
I don't really like the force overwriting solution because it would effectively "desync" the annotations from the PDF, making the annotation file essentially useless.

> But in my opinion plugin dealing with auto-exporting would be great, although I am not sure how it should be done to still allow editing file without any problems.

Here are my thoughts on how the plugin might work:
* Let's suppose we have PDF background file `lecture.pdf` and annotations `notes.xopp`.
* The plugin (or whatever the final implementation ends up being) will allow the user to set an "auto export" directory `$MY_EXPORT_DIR`.
* At regular time intervals, or whenever the user saves the annotations, the plugin will export `notes.pdf` (or some other configurable file name) to `$MY_EXPORT_DIR`.

This way, the annotations can always be edited because the auto-export plugin only creates a copy.

Phabbits:
I believe both of these features have their use cases, and both should be implemented.

[...]

PR #2618, opened 2021-01-03
Add publish action for PDF export

dhmint:
Adds a publish action along with menu item and timer callback which can be configured in settings. Publish is essentially ExportAsPdf except that it is a silent operation (no user interaction, file choosing is done automatically like autosave) and that it optionally executes a program/script on the exported file. We have written this extension for online courses in math and computer science and think it would be of interest to other users as well.


src/control/Control.h
```diff
@@ -385,6 +388,11 @@ class Control:
     int autosaveTimeout = 0;
     fs::path lastAutosaveFilename;
 
+    /**
+     * The publish handler ID
+     */
+    int publishTimeout = 0;
```
Technius, line 394: This name is confusing; I initially thought this was the publication interval. Can you rename this to something like `autoPublishHandler`?

Technius (changes requested):
Thanks for the PR! Related: #2363

Something like this is better implemented as a plugin (input wanted from @rolandlo), although we would need a minor addition to the plugin API. This would avoid having to add a separate publish action, which will just cause confusion in the future.

It doesn't make sense to add a `silent` flag to `showFileChooser` when the whole point of `showFileChooser` is to, well, present a dialog for the user to select a file.

Thoughts?

rolandlo:
> Something like this is better implemented as a plugin (input wanted from @rolandlo), although we would need a minor addition to the plugin API. This would avoid having to add a separate publish action, which will just cause confusion in the future.

In principle I agree it makes sense to implement this as a plugin. It will need a different UI though and there is one issue, which may be quite non-trivial. Suggestion/questions:

- I would add an `app.ExportAsPdf(filename)` command to the plugin API, where all you need to specify is an absolute filename. 
- The filename can be generated flexibly by the plugin. The plugin should probably have access to the filename of the xopp-file. So should we add another API command `app.getFilepath()` or should this be part of `app.getDocumentStructure()`?
- Code to execute after the export could be written in Lua as well (using the `os.execute` or `io.popen` for system calls, but also capable to interact more closely with Xournal++ than pure system calls). **However how should we make the code only run after the export job has been finished?** Currently this happens in `PdfPublishJob::afterExport()`. That looks non-trivial.
- Via the plugin we would need to create an independent little GUI, which we can do using the Lua `lgi`-module. If you have Xournal++ cover the full screen, this plugin GUI would have to be focused again if you want to change the settings or trigger a "publish" action.
- Should the settings (for the time interval), be saved in an external file (or is it fine if the time interval is forgotten when restarting Xournal++)?

LittleHuba:
> However how should we make the code only run after the export job has been finished?

Is it possible to provide a callback in the plugin that can be called from our side once the export is finished?

rolandlo:
> > However how should we make the code only run after the export job has been finished?
> 
> Is it possible to provide a callback in the plugin that can be called from our side once the export is finished?

In principle, this should be possible, sure. The plugin would have to register a listener for this signal and then we could call the function. It means the beginning of the introduction of a signaling system, extending our home grown plugin engine. This means some work, but also benefits with regards to other use cases of plugins.

bhennion:
I have given some thoughts to this. I think an ideal Publish plugin/job should

1. Allow for exports in every supported format.
2. For formats allowing it, only export the pages that have changed since the last publication (to avoid further lags like in #2675)

A use case of mine: publish my notes as a webpage. Each page is a small png file updated only when needed.

Should the plugin callback function be implemented (as suggested by @LittleHuba), then the first item will not be an issue. The second point however, requires a up-to-date hasPageBeenModified flag exposed to the plugins.

Would that be doable?

rolandlo:
> I have given some thoughts to this. I think an ideal Publish plugin/job should
> 
>     1. Allow for exports in every supported format.
> 
>     2. For formats allowing it, only export the pages that have changed since the last publication (to avoid further lags like in [Temporary input lag when autosaving a big file #2675](https://github.com/xournalpp/xournalpp/issues/2675))
> 
> 
> A use case of mine: publish my notes as a webpage. Each page is a small png file updated only when needed.
> 
> Should the plugin callback function be implemented (as suggested by @LittleHuba), then the first item will not be an issue. The second point however, requires a up-to-date hasPageBeenModified flag exposed to the plugins.
> 
> Would that be doable?

@bhennion It would be easy to add a new document listener, that stores (for each page) the time, when it has last been updated. The plugin API can provide a function exposing these update times and the plugin can check for pages that need update in regular time intervals (for example once each second). Does that sound like a good solution to you? It would be easier than to notify the plugin directly about page updates.

bhennion:
@rolandlo yes, that sounds like a better way to so that!

Issue #2619, opened 2021-01-03
Stylus pressure multiplier


caph1993:
**Is your feature request related to a problem? Please describe.**

My typical stroke renders very thin on the screen, even with the thick tool.

![image](https://user-images.githubusercontent.com/5473460/103489903-095f7600-4de6-11eb-9437-bb47b371d127.png)

**Describe the solution you'd like**

A menu for selecting the **input-output pressure curve**, like those used in image editors would be fantastic.

Example:

![image](https://user-images.githubusercontent.com/5473460/103490314-0ca83100-4de9-11eb-9838-1e88cd83f5c9.png)

**Describe alternatives you've considered**

A simple scale factor for the stylus pressure may work. I had the same issue using [another app](http://www.styluslabs.com/), and they provided a 'pressure scale factor' that improved the minimum width a lot, although slower and harder strokes became extremely wide.

personalizedrefrigerator:
#2622 doesn't add a pressure curve, but I think it addresses this!

tattsan:
The PR #2622 works fine!  Thank you so much.  The line thickness of  draw tool is the one for 100% pressure, but normal writing pressure is not so high. For this reason, when both are mixed, the balance of thickness becomes unbalanced.  The PR solved this issue! 

The PR will fix #1015 and #2545 .  The solution is similar to #1361 , but from a different point of view.

PR #2622, opened 2021-01-04
Add minimum pressure and pressure multiplier settings

personalizedrefrigerator:
## Summary
Adds settings for minimum pressure and a pressure multiplier.

image
image
image

Although it does not add pressure curves, it should address #2619.


rolandlo (approved):
A very good addition. It was indeed a problem, that the strokes drawn with a pressure sensitive stylus could never be quite as thick as those drawn with the mouse on devices like mine (and maybe the other way round on others).

This will also fix #2118. 

The PR looks good to me. Nothing I would change. Is it ready from your side as well?

personalizedrefrigerator:
Yes. I think the PR is ready.

rolandlo:
Thanks, good job!

PR #2624, opened 2021-01-06
Added vertical mode for the pagespinner tool

### LINKED ISSUE #1156: Toolbar Pagetool not Rotating
I like to have my Tools on the left side the problem is that the Page Selection tool isn't rotating.

This is how it looks now:
Active_Implementation

And here a possible implementation:
Posible_Implementation

Thomi7:
Fixes #1156

image

### DIFF
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -37,8 +37,7 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
 auto ToolPageSpinner::newItem() -> GtkToolItem* {
     GtkOrientation orientation = horizontal ? GTK_ORIENTATION_HORIZONTAL : GTK_ORIENTATION_VERTICAL;
 
-    GtkWidget* spinner = this->pageSpinner->getWidget();
-    gtk_widget_unparent(spinner);  // make spinner re-addable
+    GtkWidget* spinner = this->pageSpinner->newWidget();
     gtk_orientable_set_orientation(reinterpret_cast<GtkOrientable*>(spinner), orientation);
 
     GtkWidget* pageLabel = gtk_label_new(_("Page"));
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -35,24 +35,51 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
 }
 
 auto ToolPageSpinner::newItem() -> GtkToolItem* {
-    GtkToolItem* it = gtk_tool_item_new();
+    GtkOrientation orientation = horizontal ? GTK_ORIENTATION_HORIZONTAL : GTK_ORIENTATION_VERTICAL;
 
-    GtkWidget* hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 1);
+    GtkWidget* spinner = this->pageSpinner->getWidget();
+    gtk_widget_unparent(spinner);  // make spinner re-addable
+    gtk_orientable_set_orientation(reinterpret_cast<GtkOrientable*>(spinner), orientation);
 
     GtkWidget* pageLabel = gtk_label_new(_("Page"));
-    GtkWidget* spinner = this->pageSpinner->getWidget();
+    this->lbPageNo = gtk_label_new(this->lbPageNo ? gtk_label_get_text(GTK_LABEL(lbPageNo)) : "");
 
-    gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), pageLabel, false, false, 7);
+    if (horizontal) {
+        gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
+        gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
+        gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
+    } else {
+        gtk_widget_set_halign(pageLabel, GTK_ALIGN_BASELINE);
+        gtk_widget_set_halign(spinner, GTK_ALIGN_BASELINE);
+        gtk_widget_set_halign(this->lbPageNo, GTK_ALIGN_BASELINE);
+    }
 
-    gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), spinner, false, false, 0);
+    GtkWidget* box = gtk_box_new(orientation, 1);
+    gtk_box_pack_start(GTK_BOX(box), pageLabel, false, false, 7);
+    gtk_box_pack_start(GTK_BOX(box), spinner, false, false, 0);
+    gtk_box_pack_start(GTK_BOX(box), this->lbPageNo, false, false, 7);
 
-    this->lbPageNo = gtk_label_new("");
-    gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), this->lbPageNo, false, false, 7);
-
-    gtk_container_add(GTK_CONTAINER(it), hbox);
+    GtkToolItem* it = gtk_tool_item_new();
+    gtk_container_add(GTK_CONTAINER(it), box);
 
     return it;
+}
+
+auto ToolPageSpinner::createItem(bool horizontal) -> GtkToolItem* {
+    this->horizontal = horizontal;
+
+    this->item = createTmpItem(horizontal);
+    g_object_ref(this->item);
+
+    if (GTK_IS_TOOL_BUTTON(this->item) || GTK_IS_TOGGLE_TOOL_BUTTON(this->item)) {
+        g_signal_connect(this->item, "clicked", G_CALLBACK(&toolButtonCallback), this);
+    }
+
+    return this->item;
+}
+
+auto ToolPageSpinner::createTmpItem(bool horizontal) -> GtkToolItem* {
+    GtkToolItem* item = AbstractToolItem::createTmpItem(horizontal);
+    g_object_ref(item);
+    return item;
 }
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -72,6 +72,9 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
     GtkWidget* pageLabel = gtk_label_new(_("Page"));
     this->lbPageNo = gtk_label_new("");
 
+    box = gtk_box_new(orientation, 1);
+    updateLabels();
+
     if (orientation == GTK_ORIENTATION_HORIZONTAL) {
         g_clear_object(&this->lbVerticalPdfPage);
         gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
@@ -88,9 +91,6 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
         gtk_widget_set_halign(lbVerticalPdfPage, GTK_ALIGN_BASELINE);
     }
 
-    updateLabels();
-
-    box = gtk_box_new(orientation, 1);
     gtk_box_pack_start(GTK_BOX(box), pageLabel, false, false, 7);
     gtk_box_pack_start(GTK_BOX(box), spinner, false, false, 0);
     gtk_box_pack_start(GTK_BOX(box), this->lbPageNo, false, false, 7);
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -73,11 +73,14 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
     this->lbPageNo = gtk_label_new("");
 
     if (orientation == GTK_ORIENTATION_HORIZONTAL) {
-        this->lbVerticalPdfPage = nullptr;
+        g_clear_object(&this->lbVerticalPdfPage);
         gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
         gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
         gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
     } else {
+        if (this->lbVerticalPdfPage) {
+            g_object_unref(this->lbVerticalPdfPage);
+        }
         this->lbVerticalPdfPage = gtk_label_new("");
         gtk_widget_set_halign(pageLabel, GTK_ALIGN_BASELINE);
         gtk_widget_set_halign(spinner, GTK_ALIGN_CENTER);
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -22,9 +22,36 @@ ToolPageSpinner::~ToolPageSpinner() {
 
 auto ToolPageSpinner::getPageSpinner() -> SpinPageAdapter* { return pageSpinner; }
 
-void ToolPageSpinner::setText(const string& text) {
+void ToolPageSpinner::setPageInfo(const size_t pagecount, const size_t pdfpage) {
+    this->pagecount = pagecount;
+    this->pdfpage = pdfpage;
     if (lbPageNo) {
-        gtk_label_set_text(GTK_LABEL(lbPageNo), text.c_str());
+        updateLabels();
+    }
+}
+
+void ToolPageSpinner::updateLabels() {
+    string ofString = FS(C_F("Page {pagenumber} \"of {pagecount}\"", " of {1}") % this->pagecount);
+    if (this->horizontal) {
+        string pdfString;
+        if (this->pdfpage > 0) {  // zero means that theres no pdf currently
+            pdfString = string(", ") + FS(_F("PDF Page {1}") % this->pdfpage);
+        }
+        gtk_label_set_text(GTK_LABEL(lbPageNo), (ofString + pdfString).c_str());
+    } else {
+        gtk_label_set_text(GTK_LABEL(lbPageNo), ofString.c_str());
+        if (this->pdfpage > 0) {  // zero means that theres no pdf currently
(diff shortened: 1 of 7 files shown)


src/gui/widgets/SpinPageAdapter.cpp
```diff
@@ -37,7 +34,16 @@ void SpinPageAdapter::pageNrSpinChangedCallback(GtkSpinButton* spinbutton, SpinP
     adapter->lastTimeoutId = g_timeout_add(100, reinterpret_cast<GSourceFunc>(pageNrSpinChangedTimerCallback), adapter);
 }
 
-auto SpinPageAdapter::getWidget() -> GtkWidget* { return this->widget; }
+auto SpinPageAdapter::newWidget() -> GtkWidget* {
+    g_object_unref(this->widget);
```
LittleHuba, line 38: Not quite happy with this. This method does not make it obvious that through the creation of a new widget the old one is dropped into a partially unconnected state. Also the widget is in this case the listener and should therefore be connected to the adapter. The adapter should have nothing to do with creating the widget. Therefore create the widget in the ToolPageSpinner itself and then attach it to the SpinPageAdapter. For this you can create a new interface for connecting widgets to this adapter. Then either make it obvious that the adapter can only handle one attached widget or allow for multiple widgets to be attached.

LittleHuba (changes requested):
Thanks for your work. The logic seems to be somewhat awkward between the PageSpinAdapter and the widget but the rest looks good!

tattsan:
Thanks, this feature is what I've been waiting for!
I tried this PR.

image

When annotating PDF files, the pagespinner seems to be still too wide.

Thomi7:
> Thanks, this feature is what I've been waiting for!
> I tried the PR.
> 
> image
> 
> When annotating PDF files, the pagespinner seems to be still too wide.

image

It looks like this now. I was able to save some space by removing the "Page" from this label. The size of the spinner should now be correct too.

Thomi7:
@LittleHuba I'm gonna look in to your review later.

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -35,24 +62,57 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
...
+    g_object_ref(this->item);
+
+    if (GTK_IS_TOOL_BUTTON(this->item) || GTK_IS_TOGGLE_TOOL_BUTTON(this->item)) {
+        g_signal_connect(this->item, "clicked", G_CALLBACK(&toolButtonCallback), this);
+    }
+
+    return this->item;
+}
+
+auto ToolPageSpinner::createTmpItem(bool horizontal) -> GtkToolItem* {
```
LittleHuba, line 114: As far as I can see this is only used once in the whole code. It would make sense to not have this in its own method then.

LittleHuba (changes requested):
This is looking way cleaner now. One small change and then this should be good to go. I'll do the functional test once you integrated my request. Thanks for the good work so far!

src/gui/toolbarMenubar/ToolPageSpinner.cpp
Thomi7: I removed some obsolete code in my latest commit. I think that we need this method because it overrides the Method of AbstractToolItem. When I remove it or move the code to the createItem Method I get GTK assertions because the g_object_ref is missing.

LittleHuba: Okay that might be the case. Then the override flag is missing in the header. It would have made this restriction obvious.

Thomi7: I've added it to all overridden methods in the ToolPageSpinner header.

src/gui/widgets/SpinPageAdapter.cpp
```diff
@@ -33,11 +27,31 @@ void SpinPageAdapter::pageNrSpinChangedCallback(GtkSpinButton* spinbutton, SpinP
...
+
+    gtk_spin_button_set_range(GTK_SPIN_BUTTON(this->widget), min, max);
+    gtk_spin_button_set_value(GTK_SPIN_BUTTON(this->widget), this->page);
+}
+
+void SpinPageAdapter::removeWidget() {
+    assert(this->hasWidget());
+
+    g_signal_handler_disconnect(this->widget, this->pageNrSpinChangedHandlerId);
+    this->widget = nullptr;
```
Technius, line 53:
```suggestion
    g_clear_object(&this->widget);
```

src/gui/widgets/SpinPageAdapter.cpp
```diff
@@ -2,18 +2,12 @@
...
+    this->min = 0;
+    this->max = 1;
     this->page = -1;
 }
 
-SpinPageAdapter::~SpinPageAdapter() {
-    g_object_unref(this->widget);
-    this->widget = nullptr;
-}
+SpinPageAdapter::~SpinPageAdapter() { this->widget = nullptr; }
```
Technius, line 10:
```suggestion
SpinPageAdapter::~SpinPageAdapter() { g_clear_object(&this->widget); }
```

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -35,24 +62,52 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
...
+    gtk_orientable_set_orientation(reinterpret_cast<GtkOrientable*>(spinner), orientation);
 
     GtkWidget* pageLabel = gtk_label_new(_("Page"));
-    GtkWidget* spinner = this->pageSpinner->getWidget();
+    this->lbPageNo = gtk_label_new("");
 
-    gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), pageLabel, false, false, 7);
+    if (orientation == GTK_ORIENTATION_HORIZONTAL) {
+        this->lbVerticalPdfPage = nullptr;
```
Technius, line 76:
```suggestion
        g_clear_object(&this->lbVerticalPdfPage);
```

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -35,24 +62,52 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
...
+    if (this->pageSpinner->hasWidget()) {
+        this->pageSpinner->removeWidget();
+    }
+    GtkWidget* spinner = gtk_spin_button_new_with_range(0, 1, 1);
+    this->pageSpinner->setWidget(spinner);
+    gtk_orientable_set_orientation(reinterpret_cast<GtkOrientable*>(spinner), orientation);
 
     GtkWidget* pageLabel = gtk_label_new(_("Page"));
-    GtkWidget* spinner = this->pageSpinner->getWidget();
+    this->lbPageNo = gtk_label_new("");
```
Technius, line 73: This will leak memory when `newItem` is called again.

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -35,24 +62,52 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
...
 
-    gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), pageLabel, false, false, 7);
+    if (orientation == GTK_ORIENTATION_HORIZONTAL) {
+        this->lbVerticalPdfPage = nullptr;
+        gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
+        gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
+        gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
+    } else {
+        this->lbVerticalPdfPage = gtk_label_new("");
```
Technius, line 81:
```suggestion
        if (this->lbVerticalPdfPage) {
            g_object_unref(this->lbVerticalPdfPage);
        }
        this->lbVerticalPdfPage = gtk_label_new("");
```

Technius (changes requested):
Haven't run this yet, but memory management is iffy here.

Febbe:
I think this is to hot for 1.0.x, deferring it might be better.

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -41,19 +68,24 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
...
         gtk_widget_set_halign(pageLabel, GTK_ALIGN_BASELINE);
         gtk_widget_set_halign(spinner, GTK_ALIGN_BASELINE);
         gtk_widget_set_halign(this->lbPageNo, GTK_ALIGN_BASELINE);
+        gtk_widget_set_halign(lbVerticalPdfPage, GTK_ALIGN_BASELINE);
     }
 
-    GtkWidget* box = gtk_box_new(orientation, 1);
+    updateLabels();
+
+    box = gtk_box_new(orientation, 1);
```
tattsan: These two lines should be in reverse order. Otherwise, if ``box`` is null,  ``updateLabels()`` causes crash at ``gtk_box_pack_start`` in it. In addition, these two lines should be moved before the ``if-else clause`` just above them. Otherwise, parent widget checks in ``updateLabels()`` fails, and PDF page number won't shown.

src/gui/toolbarMenubar/ToolPageSpinner.cpp
```diff
@@ -35,24 +62,52 @@ auto ToolPageSpinner::getNewToolIcon() -> GtkWidget* {
...
+    }
 
-    gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), spinner, false, false, 0);
+    updateLabels();
 
-    this->lbPageNo = gtk_label_new("");
-    gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
-    gtk_box_pack_start(GTK_BOX(hbox), this->lbPageNo, false, false, 7);
+    box = gtk_box_new(orientation, 1);
```
tattsan, line 90: ``updateLabels();`` line and ``box = gtk_box_new(orientation, 1);`` line should be in reverse order. Otherwise, if box is null, ``updateLabels()`` causes crash at ``gtk_box_pack_start`` in it. In addition, these two lines should be moved before the ``if-else clause`` just above them. Otherwise, parent widget checks in ``updateLabels()`` fails, and PDF page number won't shown.

src/gui/toolbarMenubar/ToolPageSpinner.cpp
Thomi7: If i understand that correctly we would need a lot more of these gobj unrefs. For example in line 75 for that box = gtk_box_new(orientation, 1); 
Is that right?

Technius: Yeah, the important thing is to understand that [gtk objects are reference counted](https://developer.gnome.org/programming-guidelines/stable/memory-management.html.en). You need an unref if you are going to "lose" the reference, or a ref if you're going to store an additional one. For example:
```
GtkWidget* box = gtk_new_new(orientation, 1); // this has a reference count of 1
gtk_widget_add(someContainer, box); // someContainer will hold a reference of box and increase the reference count to 2

// we don't need box anymore, so decrease the reference count. Otherwise we'll get a memory leak later!
g_object_unref(box); // box ref count 2 --> 1

// when the container is destroyed, ref count of box drops from 1 --> 0 and box is deallocated
```

Second example:
```
GtkLabel* label = gtk_label_new(); // ref count of 1
this->label = label; // here we "transfer" the reference to the class field
// we don't need to do a unref here because it should be unref'd at some other point (e.g. destructor)
```

Thomi7: Ok, thanks for that explanation. I'm gonna try to fix that.

Thomi7:
@Technius I've tried to fix the memory management. The code in ToolPageSpinner::newItem makes sense for me but i get GTK assertions in that part. The only reason I could think of is that gtk_box_pack_start does not increase the ref count for some reason. Is that the problem?

To finish this I also need help with ToolPageSpinner::createItem and ToolPageSpinner::createTmpItem. I don't really understand their function nor their memory management.

Technius:
Thanks for the work, looks good to me so far. I had time to look into this today and identified the remaining problem. Please try this patch:
<details>
  <summary>Patch</summary>

```diff
diff --git a/src/gui/toolbarMenubar/ToolPageSpinner.cpp b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
index 665adc20..8dd66fde 100644
--- a/src/gui/toolbarMenubar/ToolPageSpinner.cpp
+++ b/src/gui/toolbarMenubar/ToolPageSpinner.cpp
@@ -17,11 +17,9 @@ ToolPageSpinner::ToolPageSpinner(GladeGui* gui, ActionHandler* handler, string i
 
 ToolPageSpinner::~ToolPageSpinner() {
     delete this->pageSpinner;
-    if (this->lbVerticalPdfPage) {
-        g_object_unref(this->lbVerticalPdfPage);
-    }
-    g_object_unref(this->lbPageNo);
-    g_object_unref(this->box);
+    g_clear_object(&this->lbVerticalPdfPage);
+    g_clear_object(&this->lbPageNo);
+    g_clear_object(&this->box);
 }
 
 auto ToolPageSpinner::getPageSpinner() -> SpinPageAdapter* { return pageSpinner; }
@@ -71,24 +69,28 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
     }
     GtkWidget* spinner = gtk_spin_button_new_with_range(0, 1, 1);
     gtk_orientable_set_orientation(reinterpret_cast<GtkOrientable*>(spinner), orientation);
+    g_object_ref_sink(spinner);
 
-    GtkWidget* pageLabel = gtk_label_new(_("Page"));
 
     if (this->lbPageNo) {
         g_object_unref(this->lbPageNo);
     }
     this->lbPageNo = gtk_label_new("");
+    g_object_ref_sink(this->lbPageNo);
 
     if (this->lbVerticalPdfPage) {
-        g_object_unref(this->lbVerticalPdfPage);
+        g_clear_object(&this->lbVerticalPdfPage);
     }
+    
+    GtkWidget* pageLabel = gtk_label_new(_("Page"));
     if (orientation == GTK_ORIENTATION_HORIZONTAL) {
-        this->lbVerticalPdfPage = nullptr;
         gtk_widget_set_valign(pageLabel, GTK_ALIGN_BASELINE);
         gtk_widget_set_valign(spinner, GTK_ALIGN_BASELINE);
         gtk_widget_set_valign(this->lbPageNo, GTK_ALIGN_BASELINE);
     } else {
         this->lbVerticalPdfPage = gtk_label_new("");
+        g_object_ref_sink(this->lbVerticalPdfPage);
+
         gtk_widget_set_halign(pageLabel, GTK_ALIGN_BASELINE);
         gtk_widget_set_halign(spinner, GTK_ALIGN_CENTER);
         gtk_widget_set_halign(this->lbPageNo, GTK_ALIGN_BASELINE);
@@ -97,11 +99,11 @@ auto ToolPageSpinner::newItem() -> GtkToolItem* {
     this->pageSpinner->setWidget(spinner);  // takes ownership of spinner reference
 
     if (this->box) {
-        g_object_unref(box);
+        g_object_unref(this->box);
     }
-    box = gtk_box_new(orientation, 1);
+    this->box = gtk_box_new(orientation, 1);
+    g_object_ref_sink(this->box);
     gtk_box_pack_start(GTK_BOX(box), pageLabel, false, false, 7);
-    g_object_unref(pageLabel);
     gtk_box_pack_start(GTK_BOX(box), spinner, false, false, 0);
     gtk_box_pack_start(GTK_BOX(box), this->lbPageNo, false, false, 7);
 
diff --git a/src/gui/widgets/SpinPageAdapter.cpp b/src/gui/widgets/SpinPageAdapter.cpp
index f5eb3062..2a7555df 100644
--- a/src/gui/widgets/SpinPageAdapter.cpp
+++ b/src/gui/widgets/SpinPageAdapter.cpp
@@ -40,6 +40,7 @@ bool SpinPageAdapter::hasWidget() { return this->widget != nullptr; }
 void SpinPageAdapter::setWidget(GtkWidget* widget) {
     // only one widget is supported and the previous widget has to be removed via removeWidget
     assert(!this->hasWidget());
+    g_assert_nonnull(widget);
 
     this->widget = widget;
     this->pageNrSpinChangedHandlerId =
@@ -55,8 +56,7 @@ void SpinPageAdapter::removeWidget() {
 
     g_signal_handler_disconnect(this->widget, this->pageNrSpinChangedHandlerId);
 
-    g_object_unref(this->widget);
-    this->widget = nullptr;
+    g_clear_object(&this->widget);
 }
 
 auto SpinPageAdapter::getPage() const -> int { return this->page; }
diff --git a/src/gui/widgets/SpinPageAdapter.h b/src/gui/widgets/SpinPageAdapter.h
index d175d7e5..6fc1fb4e 100644
--- a/src/gui/widgets/SpinPageAdapter.h
+++ b/src/gui/widgets/SpinPageAdapter.h
@@ -29,6 +29,10 @@ public:
 
 public:
     bool hasWidget();
+
+    /**
+     * Assumes ownership of widget
+     */
     void setWidget(GtkWidget* widget);
     void removeWidget();
 
```
</details>

In GTK, creating a new widget will creating a "floating reference" that is initially unowned. Container or widgets will automatically take ownership of floating references. Thus, when the widgets are created and stored in member fields like `this->box` or `this->lbPageNo`, the class does not actually have ownership of the widgets. The widgets are therefore being deallocated before the class destructor is called, causing a double free. The fix is to use `g_object_ref_sink` function to assume ownership of the floating references.

Thomi7:
> Thanks for the work, looks good to me so far. I had time to look into this today and identified the remaining problem. Please try this patch:
> Patch
> 
> In GTK, creating a new widget will creating a "floating reference" that is initially unowned. Container or widgets will automatically take ownership of floating references. Thus, when the widgets are created and stored in member fields like `this->box` or `this->lbPageNo`, the class does not actually have ownership of the widgets. The widgets are therefore being deallocated before the class destructor is called, causing a double free. The fix is to use `g_object_ref_sink` function to assume ownership of the floating references.

Ok, thanks for that explanation. I applied your patch. This should be ready now.

Technius:
Awesome. I'm a little busy this week so I'll take a took after this weekend.

Technius (approved):
LGTM. I tested it and it seems to work fine, though we'll probably need to merge #2860 first and then rebase this to see if there are any lingering memory issues.

@Febbe I think this should be okay to include in 1.1.0, it's not too much of a code change.

rolandlo:
@Thomi7 #2860 has been merged. So could you please squash your commits and rebase on current master? Then we can take another look at potential memory issues.

Technius:
Rebased and tested; works without any glaring issues. Merging in 24 hours.

PR #2641, opened 2021-01-09
Make line style modifiable after drawing

rolandlo:
I have often struggled with line styles, because they could not be modified after the strokes have been drawn. With this PR the line style of a whole selection is made modifiable similarly as for the color of a selection. See animation.
LineStyleDemo

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2748,12 +2748,8 @@ void Control::setLineStyle(const string& style) {
         sel = this->win->getXournal()->getSelection();
     }
 
-    // TODO(fabian): allow to change selection
     if (sel) {
-        //		UndoAction* undo = sel->setSize(size, toolHandler->getToolThickness(TOOL_PEN),
-        //										toolHandler->getToolThickness(TOOL_HIGHLIGHTER),
-        //										toolHandler->getToolThickness(TOOL_ERASER));
-        //		undoRedo->addUndoAction(undo);
+        undoRedo->addUndoAction(sel->setLineStyle(stl));
     }
 
     this->toolHandler->setLineStyle(stl);
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -17,6 +17,7 @@
 #include "undo/ColorUndoAction.h"
 #include "undo/FontUndoAction.h"
 #include "undo/InsertUndoAction.h"
+#include "undo/LineStyleUndoAction.h"
 #include "undo/SizeUndoAction.h"
 #include "undo/UndoRedoHandler.h"
 
@@ -257,6 +258,12 @@ auto EditSelection::setSize(ToolSize size, const double* thicknessPen, const dou
 auto EditSelection::setFill(int alphaPen, int alphaHighligther) -> UndoAction* {
     return this->contents->setFill(alphaPen, alphaHighligther);
 }
+
+/**
+ * Set the line style of all elements, return an undo action
+ * (Or nullptr if nothing done)
+ */
+auto EditSelection::setLineStyle(LineStyle style) -> UndoActionPtr { return this->contents->setLineStyle(style); }
 
 /**
  * Set the color of all elements, return an undo action
--- a/src/control/tools/EditSelection.h
+++ b/src/control/tools/EditSelection.h
@@ -20,6 +20,7 @@
 #include "model/Font.h"
 #include "model/PageRef.h"
 #include "model/Snapping.h"
+#include "undo/UndoAction.h"
 #include "view/ElementContainer.h"
 
 #include "CursorSelectionType.h"
@@ -31,7 +32,6 @@ class Layer;
 class XojPageView;
 class Selection;
 class Element;
-class UndoAction;
 class EditSelectionContents;
 class DeleteUndoAction;
 
@@ -121,6 +121,12 @@ public:
      */
     UndoAction* setSize(ToolSize size, const double* thicknessPen, const double* thicknessHighlighter,
                         const double* thicknessEraser);
+
+    /**
+     * Set the line style of all strokes, return an undo action
+     * (Or nullptr if nothing done)
+     */
+    UndoActionPtr setLineStyle(LineStyle style);
 
     /**
      * Set the color of all elements, return an undo action
--- a/src/control/tools/EditSelectionContents.cpp
+++ b/src/control/tools/EditSelectionContents.cpp
@@ -18,6 +18,7 @@
 #include "undo/FillUndoAction.h"
 #include "undo/FontUndoAction.h"
 #include "undo/InsertUndoAction.h"
+#include "undo/LineStyleUndoAction.h"
 #include "undo/MoveUndoAction.h"
 #include "undo/RotateUndoAction.h"
 #include "undo/ScaleUndoAction.h"
@@ -230,6 +231,37 @@ auto EditSelectionContents::setFont(XojFont& font) -> UndoAction* {
     }
     delete undo;
     return nullptr;
+}
+
+/**
+ * Set the line style of all strokes, return an undo action
+ * (Or nullptr if nothing done)
+ */
+auto EditSelectionContents::setLineStyle(LineStyle style) -> UndoActionPtr {
+    auto undo = std::make_unique<LineStyleUndoAction>(this->sourcePage, this->sourceLayer);
+
+    bool found = false;
+
+    for (Element* e: this->selected) {
+        if (e->getType() == ELEMENT_STROKE) {
+            auto s = static_cast<Stroke*>(e);
+            auto lastLineStyle = s->getLineStyle();
+            s->setLineStyle(style);
+            undo->addStroke(s, lastLineStyle, s->getLineStyle());
+
+            found = true;
+        }
+    }
+
+    if (found) {
+        this->deleteViewBuffer();
+        this->sourceView->getXournal()->repaintSelection();
+
+        return undo;
+    }
+
+
+    return {};
 }
 
 /**
--- a/src/control/tools/EditSelectionContents.h
+++ b/src/control/tools/EditSelectionContents.h
@@ -19,6 +19,7 @@
 #include "model/Element.h"
 #include "model/Font.h"
 #include "model/PageRef.h"
+#include "undo/UndoAction.h"
 #include "view/ElementContainer.h"
 
 #include "CursorSelectionType.h"
@@ -29,7 +30,6 @@ class Layer;
 class XojPageView;
 class Selection;
 class Element;
-class UndoAction;
 class EditSelectionContents;
 class DeleteUndoAction;
 
@@ -40,6 +40,12 @@ public:
     virtual ~EditSelectionContents();
 
 public:
+    /**
+     * Sets the line style for all strokes, returs an undo action
+     * (or nullptr if nothing is done)
+     */
+    UndoActionPtr setLineStyle(LineStyle style);
+
     /**
      * Sets the tool size for pen or eraser, returs an undo action
      * (or nullptr if nothing is done)
--- a/src/undo/LineStyleUndoAction.cpp
+++ b/src/undo/LineStyleUndoAction.cpp
@@ -1,0 +1,71 @@
+#include "LineStyleUndoAction.h"
+
+#include "gui/Redrawable.h"
+#include "model/PageRef.h"
+#include "model/Stroke.h"
+
+#include "Rectangle.h"
+#include "i18n.h"
+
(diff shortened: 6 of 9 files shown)


src/undo/LineStyleUndoAction.h
```diff
@@ -0,0 +1,40 @@
...
+
+public:
+    virtual bool undo(Control* control);
+    virtual bool redo(Control* control);
+    virtual string getText();
+
+    void addStroke(Stroke* s, LineStyle originalStyle, LineStyle newStyle);
+
+private:
+    std::vector<LineStyleUndoActionEntry*> data;
```
Technius, line 38: There's no need to hold pointers, just store them directly.

Technius (commented):
LGTM besides one small issue.

src/control/Control.cpp
```diff
@@ -2748,12 +2748,8 @@ void Control::setLineStyle(const string& style) {
         sel = this->win->getXournal()->getSelection();
     }
 
-    // TODO(fabian): allow to change selection
     if (sel) {
-        //		UndoAction* undo = sel->setSize(size, toolHandler->getToolThickness(TOOL_PEN),
-        //										toolHandler->getToolThickness(TOOL_HIGHLIGHTER),
-        //										toolHandler->getToolThickness(TOOL_ERASER));
-        //		undoRedo->addUndoAction(undo);
+        undoRedo->addUndoAction(UndoActionPtr(sel->setLineStyle(stl)));
```
Febbe, line 2752:
```suggestion
        undoRedo->addUndoAction(sel->setLineStyle(stl));
```

src/control/tools/EditSelection.cpp
```diff
@@ -258,6 +259,12 @@ auto EditSelection::setFill(int alphaPen, int alphaHighligther) -> UndoAction* {
     return this->contents->setFill(alphaPen, alphaHighligther);
 }
 
+/**
+ * Set the line style of all elements, return an undo action
+ * (Or nullptr if nothing done)
+ */
+auto EditSelection::setLineStyle(LineStyle style) -> UndoAction* { return this->contents->setLineStyle(style); }
```
Febbe, line 266: It is intended to use only smart ptr's. Since the target requires a smart pointer and this is new code; use smart pointers here:

```suggestion
auto EditSelection::setLineStyle(LineStyle style) -> UndoActionPtr { return this->contents->setLineStyle(style); }
```

src/control/tools/EditSelection.h
```diff
@@ -122,6 +122,12 @@ class EditSelection: public ElementContainer, public Serializeable {
     UndoAction* setSize(ToolSize size, const double* thicknessPen, const double* thicknessHighlighter,
                         const double* thicknessEraser);
 
+    /**
+     * Set the line style of all strokes, return an undo action
+     * (Or nullptr if nothing done)
+     */
+    UndoAction* setLineStyle(LineStyle style);
```
Febbe, line 129:
```suggestion
    UndoActionPtr setLineStyle(LineStyle style);
```

src/control/tools/EditSelectionContents.cpp
```diff
@@ -232,6 +233,38 @@ auto EditSelectionContents::setFont(XojFont& font) -> UndoAction* {
     return nullptr;
 }
 
+/**
+ * Set the line style of all strokes, return an undo action
+ * (Or nullptr if nothing done)
+ */
+auto EditSelectionContents::setLineStyle(LineStyle style) -> UndoAction* {
```
Febbe, line 240:
```suggestion
auto EditSelectionContents::setLineStyle(LineStyle style) -> UndoActionPtr {
```

src/control/tools/EditSelectionContents.cpp
```diff
@@ -232,6 +233,38 @@ auto EditSelectionContents::setFont(XojFont& font) -> UndoAction* {
...
+    auto* undo = new LineStyleUndoAction(this->sourcePage, this->sourceLayer);
```
Febbe, line 241:
```suggestion
    auto undo = std::make_unique<LineStyleUndoAction>(this->sourcePage, this->sourceLayer);
```

src/control/tools/EditSelectionContents.cpp
```diff
@@ -232,6 +233,38 @@ auto EditSelectionContents::setFont(XojFont& font) -> UndoAction* {
...
+
+    bool found = false;
+
+    for (Element* e: this->selected) {
+        if (e->getType() == ELEMENT_STROKE) {
+            auto s = dynamic_cast<Stroke*>(e);
```
Febbe, line 247: If you are sure this Element is a Stroke, use `static_cast`. If not, check if `s` is not nullptr. 

```suggestion
            auto s = static_cast<Stroke*>(e);
```
```suggestion
       if (auto s = dynamic_cast<Stroke*>(e); e->getType() == ELEMENT_STROKE && s) {
            
```

src/control/tools/EditSelectionContents.cpp
```diff
@@ -232,6 +233,38 @@ auto EditSelectionContents::setFont(XojFont& font) -> UndoAction* {
...
+    if (found) {
+        this->deleteViewBuffer();
+        this->sourceView->getXournal()->repaintSelection();
+
+        return undo;
+    }
+
+
+    delete undo;
+    return nullptr;
```
Febbe, line 265:
```suggestion
return {};
```

src/control/tools/EditSelectionContents.h
```diff
@@ -40,6 +40,12 @@ class EditSelectionContents: public ElementContainer, public Serializeable {
     virtual ~EditSelectionContents();
 
 public:
+    /**
+     * Sets the line style for all strokes, returs an undo action
+     * (or nullptr if nothing is done)
+     */
+    UndoAction* setLineStyle(LineStyle style);
```
Febbe, line 47:
```suggestion
    UndoActionPtr setLineStyle(LineStyle style);
```

src/undo/LineStyleUndoAction.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "gui/Redrawable.h"
+#include "model/PageRef.h"
+#include "model/Stroke.h"
+
+#include "Rectangle.h"
+#include "i18n.h"
+
+class LineStyleUndoActionEntry {
+public:
```
Febbe, line 11:
```suggestion
struct LineStyleUndoActionEntry {
```

src/undo/LineStyleUndoAction.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+LineStyleUndoAction::~LineStyleUndoAction() {
+    for (LineStyleUndoActionEntry* e: this->data) {
+        delete e;
+    }
+}
+
+void LineStyleUndoAction::addStroke(Stroke* s, LineStyle originalStyle, LineStyle newStyle) {
+    this->data.push_back(new LineStyleUndoActionEntry(s, originalStyle, newStyle));
```
Febbe, line 31: Always use emplace over push / insert. Especially after changing `data`'s type to a vector holding structs.
```suggestion
    this->data.emplace_back(s, originalStyle, newStyle);
```

src/undo/LineStyleUndoAction.cpp
```diff
@@ -0,0 +1,86 @@
...
+LineStyleUndoAction::LineStyleUndoAction(const PageRef& page, Layer* layer): UndoAction("LineStyleUndoAction") {
+    this->page = page;
+    this->layer = layer;
+}
+
+LineStyleUndoAction::~LineStyleUndoAction() {
+    for (LineStyleUndoActionEntry* e: this->data) {
+        delete e;
+    }
+}
```
Febbe, line 28:
```suggestion
```
Since you forwarded LineStyleUndoActionEntry, it could be required to default the destructor here.:
```suggestion
LineStyleUndoAction::~LineStyleUndoAction() =default;
```

src/undo/LineStyleUndoAction.cpp
```diff
@@ -0,0 +1,86 @@
...
+        return true;
+    }
+
+    LineStyleUndoActionEntry* e = this->data.front();
+    double x1 = e->s->getX();
+    double x2 = e->s->getX() + e->s->getElementWidth();
+    double y1 = e->s->getY();
+    double y2 = e->s->getY() + e->s->getElementHeight();
+
+    for (LineStyleUndoActionEntry* e: this->data) {
```
Febbe, line 45:
```suggestion
    for (LineStyleUndoActionEntry& e: this->data) {
```

src/undo/LineStyleUndoAction.cpp
```diff
@@ -0,0 +1,86 @@
...
```
Febbe, line 71:
```suggestion
    for (LineStyleUndoActionEntry& e: this->data) {
```

src/undo/LineStyleUndoAction.h
```diff
@@ -0,0 +1,40 @@
...
+
+class LineStyleUndoActionEntry;
+class Stroke;
+class Layer;
+class Redrawable;
+
+class LineStyleUndoAction: public UndoAction {
+public:
+    LineStyleUndoAction(const PageRef& page, Layer* layer);
+    virtual ~LineStyleUndoAction();
```
Febbe, line 28: Rule of 5:
```suggestion
    LineStyleUndoAction(LineStyleUndoAction const&) = delete;
    LineStyleUndoAction(LineStyleUndoAction&&) = delete;
    LineStyleUndoAction& operator=(LineStyleUndoAction const&) = delete;
    LineStyleUndoAction& operator=(LineStyleUndoAction&&) = delete;
    ~LineStyleUndoAction() override = default;
```

src/undo/LineStyleUndoAction.h
```diff
@@ -0,0 +1,40 @@
...
+
+public:
+    virtual bool undo(Control* control);
+    virtual bool redo(Control* control);
```
Febbe, line 32: Do not use virtual on inherited and overridden functions.
```suggestion
    bool undo(Control* control) override;
    bool redo(Control* control) override;
```

src/undo/LineStyleUndoAction.h
```diff
@@ -0,0 +1,40 @@
...
+    virtual string getText();
```
Febbe, line 33:
```suggestion
    string getText() override;
```

src/undo/LineStyleUndoAction.h
Febbe:
```suggestion
    std::vector<LineStyleUndoActionEntry> data;
```

Febbe (changes requested):

src/control/tools/EditSelection.cpp
rolandlo: @Febbe 
The compiler complains 
```
error: ‘UndoActionPtr’ does not name a type; did you mean ‘UndoAction’ 
```
Should I duplicate the `using` statement from the UndoRedoHandler class? However there are more errors coming along. Everything compiles when I change back to `UndoAction*` (4 places).

Febbe: This warning means, that the compiler does not know jet what this type is.
To copy the using statement could result into a name conflict, when suddenly both statements occur in the same compilation unit. So the best would be to either use the complete name or just move the usng UndoActionPtr statement into the UndoAction.h and include that.

rolandlo: Ok, I moved the `using UndoActionPtr` into the `UndoAction.h` header file. In order to make the line 
`auto undo = std::make_unique<LineStyleUndoAction>(this->sourcePage, this->sourceLayer);` compile I had to move the `LineStyleUndoActionEntry` struct to the header file. Can you please take a look if that was fine?

Febbe: > compile I had to move the LineStyleUndoActionEntry struct to the header file. Can you please take a look if that was fine?

Yes that's fine.
But it was not required (I think so), the reason it can't compile is, that the it is not known, how to destruct the vector in your class, since you only forward declared the LineStyleUndoActionEntry, but you try to delete the vector already in the header file.

To resolve that there are 2 ways: Do not forward declare the class and define it in the header (like you did).
Or move the definition (not the declaration) of the destructor of LineStyleUndoAction to the source file.

LineStyleUndoAction.h: `~LineStyleUndoAction();` <= Rule of 5 is recommended but not mandatory
LineStyleUndoAction.cpp: `~LineStyleUndoAction() = default;` <= Always Requires Rule of 5

rolandlo: As I understand everything is fine now as it is, right? So I can merge?

Febbe: Yes, Ship it

PR #2645, opened 2021-01-10
don't impose label width

sanette:
This is consistent with
https://github.com/xournalpp/xournalpp/pull/2592
(and should have been done there, but somewhat I forgot to)


rolandlo:
Thanks, I merged it.

PR #2651, opened 2021-01-11
Add touchpad pinch gesture support

David96:
This adds zooming on a touchpad pinch gesture.

This doesn't seem to work when the touch workaround is enabled, I guess the events are handled somewhere else then. If someone is able to point me in the right direction, it would be nice :)

About some things I'm not quite sure, see review comments.

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -26,6 +26,35 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
     return zoom->isZoomPresentationMode();
 }
 
+auto onTouchpadPinchEvent(GtkWidget* widget, GdkEventTouchpadPinch* event, ZoomControl* zoom) -> bool {
+    if (event->type == GDK_TOUCHPAD_PINCH && event->n_fingers == 2) {
+        utl::Point<double> center;
+        switch (event->phase) {
+            case GDK_TOUCHPAD_GESTURE_PHASE_BEGIN:
+                if (zoom->isZoomFitMode()) {
+                    zoom->setZoomFitMode(false);
+                }
+                center = {event->x, event->y};
+                // Need to adjust the center because the event coordinates are relative
+                // to the widget, not to the zoomed (and translated) view
+                center += {zoom->getVisibleRect().x, zoom->getVisibleRect().y};
+                zoom->startZoomSequence(center);
+                break;
+            case GDK_TOUCHPAD_GESTURE_PHASE_UPDATE:
+                zoom->zoomSequenceChange(event->scale, true);
+                break;
+            case GDK_TOUCHPAD_GESTURE_PHASE_END:
+                zoom->endZoomSequence();
+                break;
+            case GDK_TOUCHPAD_GESTURE_PHASE_CANCEL:
+                zoom->cancelZoomSequence();
+                break;
+        }
+        return true;
+    }
+    return false;
+}
+
 
 // Todo: try to connect this function with the "expose_event", it would be way cleaner and we dont need to align/layout
 //       the pages manually, but it only works with the top Widget (GtkWindow) for now this works "fine"
@@ -113,6 +142,13 @@ void ZoomControl::endZoomSequence() {
     zoomSequenceStart = -1;
 }
 
+void ZoomControl::cancelZoomSequence() {
+    if (zoomSequenceStart != -1) {
+        setZoom(zoomSequenceStart);
+        endZoomSequence();
+    }
+}
+
 auto ZoomControl::getVisibleRect() -> Rectangle<double> {
     GtkWidget* widget = view->getWidget();
     Layout* layout = gtk_xournal_get_layout(widget);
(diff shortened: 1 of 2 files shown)


src/control/zoom/ZoomControl.cpp
```diff
@@ -26,6 +26,33 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
...
+        utl::Point<double> center;
+        switch (event->phase) {
+            case GDK_TOUCHPAD_GESTURE_PHASE_BEGIN:
+                if (zoom->isZoomFitMode()) {
+                    zoom->setZoomFitMode(false);
+                }
+                center = {event->x, event->y};
+                // Need to adjust the center because the event coordinates are relative
+                // to the widget, not to the zoomed (and translated) view
+                center += {zoom->getVisibleRect().x, zoom->getVisibleRect().y};
```
David96, line 40: So when doing the zooming for the touchscreen, this is wrapped in an if statement, checking whether "getGtkTouchscreenScrollingEnabled()" is false. In my testing, adding this offset was always necessary. With the touch work around enabled, the whole pinch gesture doesn't work at all though.

src/control/zoom/ZoomControl.cpp
```diff
@@ -157,7 +184,9 @@ void ZoomControl::addZoomListener(ZoomListener* l) { this->listener.emplace_back
 void ZoomControl::initZoomHandler(GtkWidget* window, GtkWidget* widget, XournalView* v, Control* c) {
     this->control = c;
     this->view = v;
+    gtk_widget_add_events(widget, GDK_TOUCHPAD_GESTURE_MASK);
     g_signal_connect(widget, "scroll-event", G_CALLBACK(onScrolledwindowMainScrollEvent), this);
+    g_signal_connect(widget, "event", G_CALLBACK(onTouchpadPinchEvent), this);
     g_signal_connect(window, "configure-event", G_CALLBACK(onWindowSizeChangedEvent), this);
```
David96, line 190: Not sure whether I'm just not able to find it or it really isn't there/is that bad, but I couldn't really find any documentation on this GTK/GDK event stuff. I'm not really sure about what exactly the "event" parameter does, but changing it to something like "touchpadpinch-event" didn't work. Also, I don't know whether this is the right place to do the gtk_widget_add_events call.

David96: Alright, found the signals it can connect to (https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget.signals) - as there doesn't seem to be anything specific to touchpad gestures I think we have to keep this general event handler.
Question to you then is: should I rename onTouchpadPinchEvent as it actually is a general event handler or keep it this way as it effectively only handles pinch events?

LittleHuba: Please keep the name as it is. If we ever need to handle more events it should be obvious that this handler is only for the pinch events.

src/control/zoom/ZoomControl.cpp
LittleHuba: You may keep it as it is then. `getGtkTouchscreenScrollingEnabled()` will be removed soon anyway, as this workaround got revamped into better internal handling of scrolling on touchscreens.

src/control/zoom/ZoomControl.cpp
```diff
@@ -26,6 +26,33 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
...
+                // Need to adjust the center because the event coordinates are relative
+                // to the widget, not to the zoomed (and translated) view
+                center += {zoom->getVisibleRect().x, zoom->getVisibleRect().y};
+                zoom->startZoomSequence(center);
+                break;
+            case GDK_TOUCHPAD_GESTURE_PHASE_UPDATE:
+                zoom->zoomSequenceChange(event->scale, true);
+                break;
+            case GDK_TOUCHPAD_GESTURE_PHASE_END:
+            case GDK_TOUCHPAD_GESTURE_PHASE_CANCEL:
```
LittleHuba, line 47: If the gesture is canceled, the zoom should actually be reverted and not finished. Canceling might be triggered, when GTK decides this is actually a different event.

LittleHuba (changes requested):

src/control/zoom/ZoomControl.cpp
David96: So in the cancel case I should just call zoomSequenceUpdate with a scale of 1 and end the sequence after that, right?
I'm not sure whether this is better from a users perspective - I wouldn't want anything jumping around, that's why I just ended the sequence for the cancel event, too. Not sure how late in the sequence it's possible that the event is cancelled though.

LittleHuba: Please add a method `cancelZoomSequence` in [ZoomControl.cpp](https://github.com/xournalpp/xournalpp/blob/master/src/control/zoom/ZoomControl.cpp) which reverts the zoom sequence. If I'm not mistaken all the necessary data to revert it fully is already available in this class.

David96: Alright, I hope that's fine now. Thanks for your comments!

src/control/zoom/ZoomControl.cpp
```diff
@@ -113,6 +142,11 @@ void ZoomControl::endZoomSequence() {
     zoomSequenceStart = -1;
 }
 
+void ZoomControl::cancelZoomSequence() {
```
LittleHuba, line 145: Please fence this against `zoomSequenceStart == -1` to avoid entering an invalid state should this ever be called with no current zoom sequence.

David96: Done

LittleHuba:
The implementation does unfortunately not work with my device. I assume this is GDK though and not your code. I'll investigate some more and get back to you with hopefully more information.

LittleHuba:
@David96 thanks for your work. I added a note to the commit about the missing support for the combination of Xorg and libinput. As this is not fixable on our side there is no reason to not merge this PR. We'll just have to make sure to mention this in the changelog of the release.

David96:
Alright thanks. Just out of curiosity/a bit offtopic: the commit appears as verified and has me as author - imo this kind of defeats the whole purpose of signing as the commit was clearly altered (and apparently also signed) by you. I also do have a gpg key associated with my GitHub account so I would expect Github to only write verified behind commits authored by me, that are also actually signed by me. Is this by design/do you know how that happened?

Technius:
We usually merge by rebase (as in this case), so the commit id changed and GitHub probably re-signed it with its own gpg key.

Issue #1745, opened 2020-02-18
Select text from PDF


sotte:
**Is your feature request related to a problem? Please describe.**
xournalpp is great for annotating PDFs (highligthing, adding text, whatnot) but often I want to **select text from the PDF, copy it,** and paste it into another program. xournalpp doesn't have a text selection tool like most PDF readers.

**Describe the solution you'd like**
Add a text selection tool.

**Describe alternatives you've considered**
Currently I open a PDF reader to select the text that way.

easyw:
@sotte 
this feature is the most one I'm missing too...
+1 from my side

caliarim:
+1

DystopianLphant:
+1, I have to keep my pdf viewer opened to copy the text I need

sotte:
Any ideas how to proceed here? "Simple text selection" can be quite tricky (multi column documents). Could we leverage the underlying pdf engine?

LittleHuba:
You can use the underlying pdf engine theoretically. In the current code of Xournal++ this is hard though, as the PDF is rendered as background and after that all state of the engine is discarded.

ozel:
same here. the only occasion that I am using another PDF reader sometimes is for coping things like title or author names for completing the Zotero database entry. Since the text search feature works pretty well, it looks like not all of the PDF engines state is discarded... any pointers on where to start looking in the source code?

ReddingZH:
This feature would be a huge plus for xournalpp and its use cases.

zeidoo:
+1. Would be really nice.

All3xJ:
up. It should be a must

Rots:
As per recommendation of @rolandlo just a notification in this thread that there are also annotation use cases with PDFs. I've written down my use case [in discussions](https://github.com/xournalpp/xournalpp/discussions/2681).

rmarabini:
+1

AndersSpringborg:
+1

VivekThazhathattil:
In the lightweight pdf reader MuPDF, text selection is implemented as simply right clicking and dragging to  draw a rectangle over the text that you want to copy and then pressing ctrl+C to copy it. Taking inspiration from this, I tried tweaking xournalpp's code to allow select rectangle tool to act in a similar way. Then I was able to use poppler api command poppler_page_get_selected_text() given the popplerRectangle to get  the required text. Maybe a similar method would do the trick!

adhuliya:
+1
[...]

Issue #1027, opened 2019-03-29
follow pdf hyperlinks


enricoar:
Hi,
it would be great if it was possible to follow the hyperlinks of the annotated pdf background, for example when clicking with the "select object" arrow. Hyperlinks could be internal (like references), these are the most useful ones  for example leading to equations. Or also external via the www.

Also, I think it was already mentioned, it would be nice to have the possibility to include hyperlinks (internal and/or external)
Thanks
Enrico

LittleHuba:
This is unfortunately not as easy as it sounds. It will take some time but we'll add that feature.

lukyluket:
I also agree this is the only major feature I really miss, for the rest, I like the program very much.
Thanks for the work on the project

johncbowman:
Yes, if one could follow both internal and external hyperlinks, xournalpp would become a perfect tool for the online delivery of PDF content. I need this feature for opening up links to 3D WebGL content, as well referring to previous theorems and examples in university lecture notes. Hopefully this issue can be given high priority. Thanks in advance!

Technius:
See proof-of-concept in #2660.

mizhozan:
This might be also related to this missing feature: insert link. Something it would be great to insert a link either to the external webpage or to just another page/section in the xournal document.

Technius:
> This might be also related to this missing feature: insert link. Something it would be great to insert a link either to the external webpage or to just another page/section in the xournal document.

As great as this feature would be, it would require changes to the format and therefore be a lot more work than just following existing links in the document. We could tackle this when implementing #1092, #729, or the new file format.

didierm:
As I am unable to provide any coding help, my feedback is without any merit, but nevertheless : 

unable to follow **internal** hyperlinks within a PDF document prohibits efficient navigation within a large document, and hence renders Xournal++ quite cumbersome as an annotation tool for any PDF larger than a few pages.

For that reason, I am currently falling back to KDE Okular (which has a very extensive set of annotations tools, but lacks a bit in the freehand drawing, i.e. handwritten text, department, due to - presume - interpolation issues when detecting curves : https://bugs.kde.org/show_bug.cgi?id=437401).

TornaxO7:
Are there any updates for this feature?

Technius:
[...]

PR #2660, opened 2021-01-12
[Proof of concept] Enhanced PDF tools

Technius:
This PR adds additional features for PDF interaction and notetaking. In particular, this extends the `Select Rectangle` tool to interact with objects in the background PDF (if it exists) when clicking with Alt held down (keybindings subject to change).

Currently, this PR is a proof-of-concept. Ideas, feedback, and additional help are welcome. We don't want to feature-creep this too much--the idea is to get something merged into `master` within a reasonable timeframe.

TODO:
* [ ] Interaction with links in document (#1027)
  * [x] URL support
  * [ ] Local document references (awaiting integration with Technius/xournalpp#1)
  * [ ] Button to convert into text element
  * [ ] Button to copy URL to clipboard
* [x] ~Text selection~ #1745) superceded by #3326
* [ ] Image selection
  * [ ] Popup
  * [ ] Button to copy to clipboard
  * [ ] Button to convert into image element

Demo:
Peek 2021-01-12 15-15

### DIFF
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -6,6 +6,7 @@
 #include <memory>
 
 #include <gdk/gdk.h>
+#include <glib/gprintf.h>
 
 #include "control/Control.h"
 #include "control/SearchControl.h"
@@ -545,8 +546,52 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
         } else {
             double zoom = xournal->getZoom();
             if (this->selection->userTapped(zoom)) {
-                SelectObject select(this);
-                select.at(pos.x / zoom, pos.y / zoom);
+                if (pos.isAltDown()) {
+                    // Attempt PDF selection
+                    auto& pdfDoc = this->xournal->getDocument()->getPdfDocument();
+                    if (this->getPage()->getPdfPageNr() != npos) {
+                        auto page = pdfDoc.getPage(this->getPage()->getPdfPageNr());
+
+                        // Search for selected link
+                        const auto links = page->getLinks();
+                        for (auto&& [rect, uri]: links) {
+                            const double pageX = pos.x / zoom;
+                            const double pageY = pos.y / zoom;
+                            if (!(rect.x1 <= pageX && pageX <= rect.x2 && rect.y1 <= pageY && pageY <= rect.y2)) {
+                                continue;
+                            }
+
+                            char* uriStr = g_uri_to_string(uri);
+                            char* uriText = g_markup_escape_text(uriStr, -1);
+                            char* labelMarkup = g_strdup_printf("<a href=\"%s\">%s</a>", uriStr, uriText);
+                            g_free(uriStr);
+                            g_free(uriText);
+                            GtkWidget* label = gtk_label_new(nullptr);
+                            gtk_label_set_markup(GTK_LABEL(label), labelMarkup);
+
+                            GtkWidget* box = gtk_box_new(GTK_ORIENTATION_VERTICAL, 5);
+                            gtk_container_add(GTK_CONTAINER(box), label);
+                            gtk_widget_show_all(box);
+
+                            GtkWidget* popover = gtk_popover_new(this->getXournal()->getWidget());
+                            gtk_container_add(GTK_CONTAINER(popover), box);
+
+                            auto x = static_cast<int>(this->getX() + rect.x1 * zoom);
+                            auto y = static_cast<int>(this->getY() + rect.y1 * zoom);
+                            auto w = static_cast<int>((rect.x2 - rect.x1) * zoom);
+                            auto h = static_cast<int>((rect.y2 - rect.y1) * zoom);
+                            GdkRectangle canvasRect{x, y, w, h};
+                            gtk_popover_set_pointing_to(GTK_POPOVER(popover), &canvasRect);
+                            gtk_popover_set_constrain_to(GTK_POPOVER(popover), GTK_POPOVER_CONSTRAINT_WINDOW);
+                            gtk_popover_popup(GTK_POPOVER(popover));
+
+                            break;
+                        }
+                    }
+                } else {
+                    SelectObject select(this);
+                    select.at(pos.x / zoom, pos.y / zoom);
+                }
             }
             delete this->selection;
             this->selection = nullptr;
--- a/src/pdf/base/XojPdfPage.h
+++ b/src/pdf/base/XojPdfPage.h
@@ -37,6 +37,8 @@ public:
     XojPdfPage();
     virtual ~XojPdfPage();
 
+    using Link = std::pair<XojPdfRectangle, GUri*>;
+
 public:
     virtual double getWidth() = 0;
     virtual double getHeight() = 0;
@@ -44,6 +46,8 @@ public:
     virtual void render(cairo_t* cr, bool forPrinting = false) = 0;
 
     virtual vector<XojPdfRectangle> findText(string& text) = 0;
+
+    virtual auto getLinks() -> std::vector<Link> = 0;
 
     virtual int getPageId() = 0;
 
--- a/src/pdf/popplerapi/PopplerGlibPage.cpp
+++ b/src/pdf/popplerapi/PopplerGlibPage.cpp
@@ -1,5 +1,7 @@
 #include "PopplerGlibPage.h"
 
+#include <poppler-page.h>
+
 
 PopplerGlibPage::PopplerGlibPage(PopplerPage* page): page(page) {
     if (page != nullptr) {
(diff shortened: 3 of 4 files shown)


LittleHuba:
Could you please also add a way that does not require the ALT key? Some people do not have access to a keyboard when annotating (me for example).

Technius:
> Could you please also add a way that does not require the ALT key? Some people do not have access to a keyboard when annotating (me for example).

Here are some ideas:
* Double or triple tap/click on link: this could interfere with other features
* Middle/right mouse click on link: could interfere when a tool is bound to this button
* Add a dedicated PDF interaction tool: this won't conflict with anything but requires yet another tool

Ideally, PDF interaction behavior should be keybindable.

bhennion:
> Could you please also add a way that does not require the ALT key? Some people do not have access to a keyboard when annotating (me for example).

Devices without a keyboard most likely have a touchscreen. So maybe a long touch would work?

personalizedrefrigerator:
We already have a "play sound associated with object" tool. Could we use this tool for link interaction as well (after renaming)?

Technius:
> > Could you please also add a way that does not require the ALT key? Some people do not have access to a keyboard when annotating (me for example).
> 
> Devices without a keyboard most likely have a touchscreen. So maybe a long touch would work?

This is a good idea, thanks.

> We already have a "play sound associated with object" tool. Could we use this tool for link interaction as well (after renaming)?

I like this idea; we could rename "play sound" to "element interaction," so it only uses but not modifies existing elements.

In general, we need to consolidate Select Object and Select Rectangle since the latter can already fulfil the purpose of the former.

johncbowman:
This looks very promising. Let me know once it compiles so I can try it out:

In file included from /usr/local/src/xournalpp/src/pdf/base/XojPdfDocumentInterface.h:18,
                 from /usr/local/src/xournalpp/src/pdf/base/XojPdfDocument.h:17,
                 from /usr/local/src/xournalpp/src/model/Document.h:22,
                 from /usr/local/src/xournalpp/src/control/xojfile/SaveHandler.h:18,
                 from /usr/local/src/xournalpp/src/util/CrashHandler.cpp:8:
/usr/local/src/xournalpp/src/pdf/base/XojPdfPage.h:40:45: error: 'GUri' was not declared in this scope
   40 |     using Link = std::pair<XojPdfRectangle, GUri*>;
      |                                             ^~~~
/usr/local/src/xournalpp/src/pdf/base/XojPdfPage.h:40:50: error: template argument 2 is invalid
   40 |     using Link = std::pair<XojPdfRectangle, GUri*>;
      |                                                  ^
/usr/local/src/xournalpp/src/pdf/base/XojPdfPage.h:50:44: error: 'Link' was not declared in this scope; did you mean 'link'?
   50 |     virtual auto getLinks() -> std::vector<Link> = 0;
      |                                            ^~~~
      |                                            link
/usr/local/src/xournalpp/src/pdf/base/XojPdfPage.h:50:48: error: template argument 1 is invalid
   50 |     virtual auto getLinks() -> std::vector<Link> = 0;
      |                                                ^
/usr/local/src/xournalpp/src/pdf/base/XojPdfPage.h:50:48: error: template argument 2 is invalid

johncbowman:
It looks like a new version (>= 2.65) of glib2-devel is required to compile this PR.

Technius:
> It looks like a new version (>= 2.65) of glib2-devel is required to compile this PR.

I basically went for the easiest URI library that I had available on my system for this proof-of-concept. To get this PR ready for actual inclusion, we'll need to use a URI library available on Ubuntu 18.04, like [`uriparser`](https://uriparser.github.io/).

src/pdf/base/XojPdfPage.h
```diff
@@ -37,6 +37,8 @@ class XojPdfPage {
     XojPdfPage();
     virtual ~XojPdfPage();
 
+    using Link = std::pair<XojPdfRectangle, GUri*>;
```
personalizedrefrigerator, line 40: Could we use a `std::pair<XojPdfRectangle, PopplerActionUri*>` here (or even a `std::pair<XojPdfRectangle, std::string>`)?
[Documentation](https://developer.gnome.org/poppler/unstable/poppler-PopplerAction.html#PopplerAction).

src/pdf/popplerapi/PopplerGlibPage.cpp
```diff
@@ -78,3 +80,23 @@ auto PopplerGlibPage::findText(string& text) -> vector<XojPdfRectangle> {
...
+
+auto PopplerGlibPage::getLinks() -> std::vector<Link> {
+    std::vector<Link> results;
+    const double height = getHeight();
+
+    GList* links = poppler_page_get_link_mapping(this->page);
+    for (GList* l = links; l != NULL; l = g_list_next(l)) {
+        auto* link = static_cast<PopplerLinkMapping*>(l->data);
+        if (link->action->type == POPPLER_ACTION_URI && link->action->uri.uri) {
+            GUri* uri = g_uri_parse(link->action->uri.uri, G_URI_FLAGS_NONE, nullptr);
```
personalizedrefrigerator, line 92: Would it be safe to store `link->action->uri.uri` as a string? We would be avoiding the need for a URI-parsing library. I think I'm missing something here...

src/gui/PageView.cpp
```diff
@@ -545,8 +546,52 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
         } else {
             double zoom = xournal->getZoom();
             if (this->selection->userTapped(zoom)) {
-                SelectObject select(this);
-                select.at(pos.x / zoom, pos.y / zoom);
+                if (pos.isAltDown()) {
```
personalizedrefrigerator, line 549: Could we put this in `PageView::onButtonDoublePressEvent` (and only show when using the select tool)? To me, this seems more intuitive than holding alt and clicking.

personalizedrefrigerator:
Aside from the above comments, this looks good so far! Thank you for working on this!

src/gui/PageView.cpp
personalizedrefrigerator: Never mind! See https://github.com/Technius/xournalpp/pull/1.

src/pdf/popplerapi/PopplerGlibPage.cpp
Technius: If we could, we would need to ensure the following:
* The URI is a valid URI. Does poppler always ensure that the PDF links are valid URIs?
* The URI is properly encoded.

michaelsjackson:
What I am missing in Xournal++ mostly is being able to copy text from background pdf, in two forms mainly:
1.  copy selected text to clipboard, binding: F13, for example
2.  copy current page text completely to clipboard, binding: F14, for example

In both cases some background listening TTS engine could read the text, for example, or just for pasting elsewhere.  I like the listening method, while annotating, the computer could do some useful work as well.

I was surprised seeing there is a text search function (control-f) but no text select with a following copy.

I can mention another useful technique, as related somehow here:
- using Xournal / Xournal++ for foreground annotation
- using Okular for background annotation, both can work at the same time.  In Okular I do not like I can not copy from hotkey the selected text, I need one more click after the must pop up each time. If Xournal++ would have even text selection based colored highlighting of text I might not need Okular even, except its cool page gliding / scrolling method.  Thanks friends for such a great tool and the hard work so far!

michaelsjackson:
My favourite style for text copying in Xournal++ would be following technique:
- using another tool:  you select some rectangle on foreground layer
- Xournal++ selects the text in rectangle region, but in background pdf layer
- Xournal++ copies also this text into the clipboard
In short three actions in one.  You simply select your rectangle, done.  That would be my dream style how any annotation program should work, also Okular could benefit from this style, in Okular you always need one more click.  Any step you repeat very often, should be as easy as possible.  This is the main thought.  Thanks for reading so far.  Have a nice day friends.

michaelsjackson:
I guess Xournal++  has it I guess, via its plugin system, if we could only have a "universal plugin" where we can simply define our own shell command.  Then we could construct our background process required for the current page. So Xournal++ needs to send only its current page number to its plugin, the rest the triggered shell command would do.

michaelsjackson:
I would like also more fluid zoom in/out or gliding between page regions, similar as in Okular (regarding the nice gliding).  Then, you could simply use SimpleScreenRecorder and record everything, but this time with fluid zooming, nice gliding, who needs Explain Everything then? Audio optimizations you can do anyway perfectly in Reaper DAW, which is the best DAW (digital audio workstation) on universe, this is a fact.

michaelsjackson:
> I'll probably wait until #3326 is merged before I continue with this.

Cool how text highlighting is already working. Hopefully this jumps into official Xournal++ soonish.

Technius:
Closing in favor of #3851.

PR #2674, opened 2021-01-16
Layer preview: show PDF background in background layer only

mjg:
A PDF background lives in the background layer just like any other
background. In the sidebar layer preview, however, PDF backgrounds are
drawn for every layer; all other backgrounds (ruling, image) are drawn
for the background layer only.

Move the corresponding codeblock to drawPage() where all the logic is
about what to draw in the preview, and make it draw PDF backgrounds on
the background layer (and combined page view) only.

**Note**: If we can rely on -100 and -1 being the only negative layer values then `layer == -100 || layer == -1` can be simplified to `layer < 0`, of course.

### DIFF
--- a/src/control/jobs/PreviewJob.cpp
+++ b/src/control/jobs/PreviewJob.cpp
@@ -62,6 +62,11 @@ void PreviewJob::drawBackgroundPdf(Document* doc) {
 void PreviewJob::drawPage(int layer) {
     DocumentView view;
     PageRef page = this->sidebarPreview->page;
+    Document* doc = this->sidebarPreview->sidebar->getControl()->getDocument();
+
+    if (page->getBackgroundType().isPdfPage() && (layer == -100 || layer == -1)) {
+        drawBackgroundPdf(doc);
+    }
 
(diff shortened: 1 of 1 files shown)


bhennion:
Hey,
Thanks for your PR!

Maybe the previous behaviour represented an actual use case: to chose between various layers of annotations, some users may need to see the background to properly understand what the annotations mean. Imagine someone highlighting various things on different layers for instance.

Perhaps the best way to handle this would be to have a check box somewhere on the layer preview panel, called something like "Show background", which displays or hides the background in the layer preview.

mjg:
Well, 2 out of 3 background types do not show up in the preview of upper layers. This PR is just meant to treat them all the same.

Note that all background types (including PDF) always show up in the preview of the bottom layer and in the page preview.

As for other use cases: I think that especially now with the progressive output option, an option to make the layer previews incremental would be helpful - then you would have a preview of the later output "pages/slides". This option would cover the "background+annotation" scenario, as well.

Technius:
This is very useful, thanks for the PR.

PR #2678, opened 2021-01-18
Enable OpenBSD support

cmburn:
I can try to write a port and get this into our package repo sometime in the next week or so as well!!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,11 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "FreeBS
     set (xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-lX11 -lXi")
 endif ()
 
+# OpenBSD needs the Xlibs too, but is different about where it stores them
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -69,6 +69,11 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "FreeBS
     set (xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-lX11 -lXi")
 endif ()
 
+# OpenBSD needs the Xlibs too, but is different about where it stores them
+if (${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
+    set (xournalpp_LDFLAGS ${xournalpp_LDFLAGS} "-L/usr/X11R6/lib -lX11 -lXi")
```
Febbe, line 74: Generally it is better to not link libraries manually, they should be linked over CMake targets.
There is already a find script in Cmake, hopefully this also works out of the box.
Can you try to use https://cmake.org/cmake/help/latest/module/FindX11.html
```suggestion
    find_package(X11 Required) 
    list(APPEND xournalpp_LDFLAGS X11::X11)
```

Febbe (changes requested):

PR #2695, opened 2021-01-23
Allow for panning during touchpad pinch gesture

David96:
Just noticed that this doesn't respect natural scrolling which might feel weird. Not sure whether there is any method of checking whether the user has this enabled.
As in this case the touchpad really works like a touchscreen, I made it always use natural scrolling for now. I would find it nice, if this would depend on the libinput/whatever is being used setting but a quick search didn't yield anything helpful there.

If you want, I could add a setting for that.

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -28,7 +28,7 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
 
 auto onTouchpadPinchEvent(GtkWidget* widget, GdkEventTouchpadPinch* event, ZoomControl* zoom) -> bool {
     if (event->type == GDK_TOUCHPAD_PINCH && event->n_fingers == 2) {
-        utl::Point<double> center;
+        utl::Point<double> center, visibleRectPos;
         switch (event->phase) {
             case GDK_TOUCHPAD_GESTURE_PHASE_BEGIN:
                 if (zoom->isZoomFitMode()) {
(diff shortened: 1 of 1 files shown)


redweasel:
I guess after https://github.com/xournalpp/xournalpp/pull/2743 lands your patch would be:
```diff
diff --git a/src/control/zoom/ZoomControl.cpp b/src/control/zoom/ZoomControl.cpp
index 42243569..903ccfa1 100644
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -46,7 +46,7 @@ auto onTouchpadPinchEvent(GtkWidget* widget, GdkEventTouchpadPinch* event, ZoomC
                 zoom->startZoomSequence(center);
                 break;
             case GDK_TOUCHPAD_GESTURE_PHASE_UPDATE:
-                zoom->zoomSequenceChange(event->scale, true);
+                zoom->zoomSequenceChange(event->scale, true, {event->dx, event->dy});
                 break;
             case GDK_TOUCHPAD_GESTURE_PHASE_END:
                 zoom->endZoomSequence();```

David96:
Thanks :) Will wait for your patch to get merged before modifying this one though.

LittleHuba:
> Just noticed that this doesn't respect natural scrolling which might feel weird. Not sure whether there is any method of checking whether the user has this enabled.
As in this case the touchpad really works like a touchscreen, I made it always use natural scrolling for now. I would find it nice, if this would depend on the libinput/whatever is being used setting but a quick search didn't yield anything helpful there.
>
> If you want, I could add a setting for that.

We will at least require a setting for users. It would be very confusing otherwise. Better would be if you could find out whether it is activated through GDK.

LittleHuba:
I'll close this because there was no work done for at least six months. Please reopen if you are interested in continuing your work.

PR #2699, opened 2021-01-24
Fix page preview only updating one page

bhennion:
When writing on 2 (or more) different pages between two refreshing of the page preview (5 seconds between two successive refreshing), only the first modified page gets its preview updated. 
This PR fixes that.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1622,12 +1622,9 @@ void Control::undoRedoChanged() {
 }
 
 void Control::undoRedoPageChanged(PageRef page) {
-    if (!this->changedPages.empty() &&
-        std::find(begin(this->changedPages), end(this->changedPages), page) == end(this->changedPages)) {
-        return;
+    if (std::find(begin(this->changedPages), end(this->changedPages), page) == end(this->changedPages)) {
+        this->changedPages.emplace_back(std::move(page));
     }
-
-    this->changedPages.emplace_back(std::move(page));
 }
 
 void Control::selectTool(ToolType type) {


src/control/Control.cpp
```diff
@@ -1623,7 +1623,7 @@ void Control::undoRedoChanged() {
 
 void Control::undoRedoPageChanged(PageRef page) {
     if (!this->changedPages.empty() &&
-        std::find(begin(this->changedPages), end(this->changedPages), page) == end(this->changedPages)) {
+        std::find(begin(this->changedPages), end(this->changedPages), page) != end(this->changedPages)) {
         return;
     }
```
bhennion, line 1629: Wouldn't this be better if `Control::changedPages` was an std::unordered_set to begin with?
I may be wrong here, with `changedPages` being usually very small.

rolandlo: I guess using `std::unordered_set` would be cleaner, but performance-wise I don't think it makes a big difference.

src/control/Control.cpp
```diff
@@ -1623,7 +1623,7 @@ void Control::undoRedoChanged() {
 
 void Control::undoRedoPageChanged(PageRef page) {
     if (!this->changedPages.empty() &&
```
rolandlo, line 1625: You are right about the fix. The check whether `changedPages` is non-empty can be removed then, however. So please remove it.

rolandlo (changes requested):

src/control/Control.cpp
bhennion: Done

rolandlo:
Wait, the `==` should still be a `!=`

src/control/Control.cpp
bhennion: We may even make things worse performance-wise (because of the hash computations and the allocation overheads). Let's leave it like that.

bhennion:
> Wait, the `==` should still be a `!=`

No, I move the statement from after the `if(...) { return; }` to inside the `if` statement.

rolandlo:
Ups, missed that.

rolandlo:
So it's ready to be merged, right?

bhennion:
Yes yes, go ahead!

rolandlo:
Ok, thanks for the fix.

Technius:
Changelog note: this and #2630 are regression fixes. (please correct me if I'm wrong)

rolandlo:
Correct, both are regression fixes.

Issue #2166, opened 2020-08-14
Calligraphic Pen


cuppajoeman:
Hey, I think this program is great! I've been using it to take notes, but I want to emulate my fountain pen.

I'm just requesting this feature, but I'd love to help out too. If I wanted to help implement this what would I do? Would I be creating a plugin, or something else?

Thanks.

rolandlo:
Hi! Great that you like Xournal++ and would love to help out. Could you share in more details, what you would like to achieve? I guess the goal is to write with the pen/stylus in a way that the ink looks like from a calligraphic pen. Do you have some specific behavior in mind that you know from other software? Maybe [inkscape](https://inkscape.org/doc/tutorials/calligraphy/tutorial-calligraphy.html)?

Most likely this can not be implemented as a plugin, since it would rely on too much of Xournal++'s core functionality (so it would require too much effort to extend the plugin API). It would have to be implemented in the C++ base.
The strokes in Xournal++ are composed of (small) line segments of variable width. What you could want to achieve is to make this variable width dependent (in a non-trivial way) from several parameters (pen pressure, angle, curvature etc.) A feature that has already been requested is a "pressure curve" #1170. In Inkscape there are also parameters like "tremor", "wiggle and mass". 
A little warning: pen tilt support is not supported by GTK+ (according to #2065). So at least currently it would have to emulate the fountain pen without detecting tilt. 

Do you have something in mind like that? Please share your thoughts.

cuppajoeman:
Hey rolando, thanks for the reply! 

I looked over your inkscape link and wow that looks nice. For me personally, I am using an old thinkpad (X200T) and it's pen is not pressure sensitive, so for my use case and experience in programming, I think the best thing for me to look at would be a minimal working calligraphy pen, that is, a constant angle while writing (can be tweaked with a slider/input when not writing) and a thickness: 

![image](https://user-images.githubusercontent.com/19198104/90302562-52398100-de7d-11ea-9518-adcc363298c2.png)

I started thinking about a high level implementation for this:

![image](https://user-images.githubusercontent.com/19198104/90302689-629e2b80-de7e-11ea-91b0-6eedd0541510.png)

Given the movement vector (if that exists) for the pen, we simply check the angle with it and the specified angle that the user has provided for the pen, if they are moving at the same angle (+ pi * k) then the thickness of the pen is scaled down to almost nothing, on the other extreme, the movement is perpedicular to the given angle then the thickness is not reduced at all.

[...]

Issue #2320, opened 2020-10-16
Stroke smoothing


TentativeConvert:
I have a very shaky hand, even more on a tablet than on paper.  Gimp has a nice feature that they call "stroke smoothing", which really helps improve my handwriting on my tablet:

![Gimp-vs-Xournalpp](https://user-images.githubusercontent.com/16051841/96257661-09a45e00-0fbb-11eb-94ca-eb2292121ade.png)

Is there any chance a similar feature could be implemented in Xournal++?

rolandlo:
Generally speaking that's something I consider working on (or at least looking in to) in the future. There are more urgent tasks(like geometry tools and fixing annoying bugs) though, so don't expect anything from my side soon.

rolandlo:
Anyone willing to work on this is advised to take a look [in this opentoonz issue](https://github.com/opentoonz/opentoonz/issues/81), where different implementations are discussed.

bhennion:
I was looking into the existing smoothing algorithms, trying to determine which was more adapted to handwritten words. I applied the various stabilizers Krita has to offer, with various parameters, to a recorded input sequence (using libinput record/replay)
![stabilizers](https://user-images.githubusercontent.com/17818041/101196856-9685a280-3661-11eb-9021-368e6fd3b86c.png)

In the case of my handwritting (and my input hardware), it feels like the first line (without stabilizer) is better. Still, I am sure there are cases, like @TentativeConvert's, where a stabilizer could help.

In order to better document the issue, it would help to have other libinput recordings on which I could test various smoothing techniques. So if other linux users could try and use [libinput record](https://people.freedesktop.org/~whot/libinput-rtd/tools.html#libinput-record), write a line or two and post the resulting yml, it would be great.

TentativeConvert:
Here's my  [libinput record](https://people.freedesktop.org/~whot/libinput-rtd/tools.html#libinput-record)ing of the following two lines:

![This-is-a-test-stroke](https://user-images.githubusercontent.com/16051841/101240210-207b4d00-36ee-11eb-99bc-5cb6df9a7709.png)

(I needed to zip the yml file before uploading it.)

[teststroke.zip](https://github.com/xournalpp/xournalpp/files/5646887/teststroke.zip)

bhennion:
Thanks!

I made some research and experiments about a couple of open source algorithms. This first comment is experimental. There will be a second comment explaining how (I think) those algorithms work.

[...]

PR #2702, opened 2021-01-24
[WIP] add in calligraphic stroke

cuppajoeman:
This pull request is in relation to #2166 . @rolandlo helped me develop a standalone .cpp file which generates a calligraphic stroke given a set of points, nib angle, and thickness. The file contains the function itself, along with the main function where I was generating the stroke as an svg, this is probably why the tests fail on my code.

@rolandlo [has said](https://github.com/xournalpp/xournalpp/issues/2166#issuecomment-766433648) that he will be integrating my code into to xournalpp, as he has the familiarity with the code base.


rolandlo:
So, I baked in the calligraphic stroke style into Xournal++. There is no way to switch to normal drawing yet.
First impressioin:
- There is a serious problem with performance. The drawing takes too long and is lagging behind. See video:

https://user-images.githubusercontent.com/40485433/105735322-dbc4a300-5f33-11eb-8ca4-ab33145a5401.mp4



 
- Once the drawing is complete, the result looks kind of nice. Example picture:
grafik


@cuppajoeman Could you please make a rigorous analysis, how much time each step of the drawing algorithm takes? Also think about how the performance could be improved.

cuppajoeman:
@rolandlo If I were to hazard I guess I would say the issue is the fact that I create `back_and_forth_list`. The reason for it's creation was so that I could iterate over all the points first forwards, then backwards, and well that probably takes a while to do. I'll try to modify it so that we don't have to create the extra data structure, and we can see how that preforms.

I am in class right now, so I might not be able to get it done straight away, but I will try to get to it later today.

rolandlo:
Take your time. But I would first analyze the time needed. It could also be that the actual filling by Cairo takes most time.

cuppajoeman:
Do you have any recommendations on how I could test the runtime of certain parts of the code?

rolandlo:
I would use something like this: https://www.geeksforgeeks.org/measure-execution-time-function-cpp/

rolandlo:
By the way I have observed that calligraphic stroke style already works nicely with splines. There performance issues are not felt.

cuppajoeman:
Hi @rolandlo, I wrapped the three main parts of the code and ran it on a huge list (78528 lines with four points per line), here were the results:

> Construction of backAndForthList took 20190 microseconds 
> Time to construct the calligraphic path: 34368 microseconds 
> Time for cairo to fill in the shape: 2186 microseconds 

Based on this it seems like cairo isn't the part of the code the takes the longest.

In terms of the analysis of the code, suppose N is the size of the list of points coming in.

Creating Back forth list iterates over the list two times adding the point to the end of the vector, that will take 2 * N steps, under the assumption that adding to the end of a vector takes constant time, then this would take 2 * N * v (where v is the time for a computer to add to the end of the vector) and take up 2 * N * p space where p is the size of the point struct.

We iterate over the backAndForthList whose length is 2* N, inside the loop some basic comparisons are done. Also note that by the time the loop completes we will called cairo_line_to approximately 2* N times for the top and bottom of the stroke. 

When a switch happens, as we draw two lines on that switch point in general with a smooth stroke I don't believe the # of switch points is not large compared to the number of non switch points, for example a perfect circle would only have two, also my name has 20 switch points, which would be small compared to the number of times it sampled the input device as it wrote

Let C represent the time cairo_line_to to run then the cost of this whole loop is 2 * N * (C).

I am not entirely sure how `cairo_fill` works, but based on the timing, it doesn't seem to be too much of an issue in terms of the runtime.

cuppajoeman:
In other words, nothing seems to weird with the runtime of the function. I will try to remove the dependence on backAndForthList by modifying the main for loop to use indices and then index into the list they pass in.

cuppajoeman:
@rolandlo I just removed `backAndForthList` and you can see it [here](https://gitlab.com/cuppajoeman/calligraphic-experimentation/-/blob/master/calligraphic.cpp). Feel free to see if this reduces the lag.

rolandlo:
The lag hasn't reduced drastically. We shouldn't build up the entire path every time from scratch, I think. We should find a solution where we only append to it.

cuppajoeman:
Right, I suppose it might help if I can get a bit more context on how my function is being used in xournalpp. I assume that it is being called on every "update" of the pen (like a new point is added)?

Does the normal stroke have a way of dealing with this?

cuppajoeman:
If that is the case, (runs everytime a point is added), then the runtime would be in theta(n^2) runtime, because we iterate over the whole list each time 1 + 2 + 3 + 4 + ... + n = (n (n-1))/2. So I can see why performance would be an issue. Like you said we need each update to run in constant time so our solution takes theta(n) runtime

rolandlo:
When a point is added, the stroke `std::vector` is updated and a render job is scheduled. Your function is called whenever the stroke is rendered, which happens quite frequently. It is the same for the normal stroke.

rolandlo:
I just discovered that the drawing of calligraphic strokes works mostly fine if we combine it with the the fill tool. If we set opacity very low, one can hardly see the filling. Only in rare moments when the whole calligraphic stroke is repainted there is a lag.

rolandlo:
I will quickly present the results from modifying the `StrokeHandler::onMotionNotify` function. Drawing seems to work pretty well now. There are only some artifacts at the end of calligraphic strokes, which requires some correction. But it looks quite amazing, I would say. Stay tuned for the video i will record!

rolandlo:
Here a video impression. I only recorded it with 5 frames per second. It felt a lot smoother while drawing than it is shown in the video. Only at the end of the second word there was some lag.

https://user-images.githubusercontent.com/40485433/106167082-22610a00-618d-11eb-804c-9e040e91f954.mp4

rolandlo:
Compare this to the same stroke drawn the normal way:
grafik

or with line width scaled for better comparison:

grafik

cuppajoeman:
Yes, I really like the way it looks, though I do see the artifact at the end of the stroke: 

image

Do you have any idea where this comes from? ~~Maybe it's to do with the pressure as they lift their pen off the page?~~

Actually it seems to be always happening as the pen moves, see this mid stroke:

image

In terms of the lag you are still experiencing, I might need to start reading the xournalpp code to understand how it works exactly to be able to make it faster, but it seems like you have found way to reduce it significantly. I'm glad we have this, but do you still think we should consider an algorithm that doesn't re-draw the whole path? 

I think if we were to try that again, then the algorithm I developed might not be too useful since I fully designed it from the point of view that I had a path ready to go and then I just draw the calligraphic stroke on top of that rather than incrementally doing it. Although maybe that might be a good use for @psychon's algorithm?

I remember at the start of the issue for the Calligraphic stroke you had suggested to make it a single polygon 
> when you create a filled shape with a contour of different color (or opacity) than the filling inside.

Maybe we could use @pshychon's algorithm to have a smooth experience when writing, and then use the alternative single polygon once the stroke is done?

Anyways just throwing some ideas out there.

I'm glad to see how far this has come though!

bhennion:
Great tool!

About the artefacts: I didn't look under the hood, but I know (from working on #2320) that @rolandlo's device emits an event with pressure 0 at the end of each stroke. Could this be a source of issues here?

psychon:
> Although maybe that might be a good use for @psychon's algorithm?

Quick thought: Sure! My code has a "mask" as its intermediate state and adds each part of the stroke to this mask. Thus, by just caching the mask, one can easily add more components to it. However, I do not know how expensive applying this mask is. This might still be slow, because it has to go through all pixels and do some computation for them. This might need fine-tuning (e.g.: One can easily approximate "only the part inside of this small rectangle here actually changed and needs redrawing").

> About the artefacts: 

Random thought that you surely already thought yourself: It looks like the path crosses on itself, i.e. "left" and "right" are switched. Since this also happens mid-drawing, this should not be pressure dependent, but I didn't look at any of the code.

rolandlo:
The artifcats should be gone now. I have just replaced the pressure values by their absolute values. If a point has no reported pressure, the pressure value assigned was -1. That probably caused the path crossings on the ends.

rolandlo:
I had a look at the pressures of my points when drawn with the stylus. The last one has pressure -1 (i.e. no pressure) and the previous points have low pressure. This completely explains the artifacts:

```
** Message: 23:49:34.799: Pressure of point 133: 1.183485
** Message: 23:49:34.799: Pressure of point 134: 1.149839
** Message: 23:49:34.799: Pressure of point 135: 1.099331
** Message: 23:49:34.799: Pressure of point 136: 1.042632
** Message: 23:49:34.799: Pressure of point 137: 1.023940
** Message: 23:49:34.799: Pressure of point 138: 0.112800
** Message: 23:49:34.799: Pressure of point 139: -1.000000
```

cuppajoeman:
@rolandlo Should I still be trying to fix the lag?

rolandlo:
I think yes. On my laptop it works reasonably, but on lower end laptops and tablets it is probably too slow. Moreover I also tested while recording a video. Then the lag was much bigger. Not sure why. But the tool should work while recording videos, otherwise we get complaints from those doing remote teaching.

cuppajoeman:
You said that we should be doing incremental filling instead of the whole thing each time, the algorithm that we came up with really only had drawing the full shape in mind, and so we would probably have to rethink the algorithm from the drawing board. Another option is trying out psychon 's algorithm.

I know you mentioned benefits of it being a single polygon, in that case maybe we could use psychon's algorithm for drawing, then once the stroke is done we turn it into the single polygon representation? 

What do you think?
C

rolandlo:
Trying out different options is a good idea. In the end we will use the one that works best.

cuppajoeman:
Here is link to @psychon's [code](https://github.com/xournalpp/xournalpp/issues/2166#issuecomment-765926959), I am not really sure how to test it out in inkscape, so either you or I or anyone willing could try it out, if you can tell how to test out the different algorithm in xournalpp

rolandlo:
@cuppajoeman Do you know how to build this PR? Were you able to test it out?

Having the stroke as one polygon saves drawing operations and it's easier to draw a contour. At least that is what I thought.

In principle if we would draw the stroke from the other side, we could just append points on both sides of the path. We could use a `std::deque` for that, but the problem is that the cairo path does not use this data structure and does not support adding stuff at the beginning efficiently. 

> maybe we could use psychon's algorithm for drawing, then once the stroke is done we turn it into the single polygon representation?

This would make sense, I think.

cuppajoeman:
@rolandlo the machine that has the drawing tablet is windows, I'm not quite sure how to build it for windows.  If you can tell point me where I can learn how to do it, I can try adding psychon's stuff. Although I also believe that it was designed to add the style to the stroke after the fact. @psychon do you think it could be modified so that it only draws the newest point?

psychon:
> do you think it could be modified so that it only draws the newest point?

Just `cairo_clip()` to a (pixel-aligned) rectangle and cairo should very quickly discard "the outside".

Alternatively, it should be possible to cache some intermediate state, but I do not know how that fits into xournalpp.

rolandlo:
@cuppajoeman You can build on Windows as described [in this wiki](https://github.com/xournalpp/xournalpp/blob/master/readme/WindowsBuild.md). It will take a while until you manage to build the first time (since you need to install a lot of stuff and compile portaudio), but once you manage, it will all become much easier and faster.

rolandlo:
I have added @psychon's method of drawing a calligraphic stroke. I didn't apply any optimizations yet. Comparing it side by side on half my laptop screen, it seemed to me that the @psychon's method was slower, but still working more or less fine. The difference became much larger when I tried the two methods on the full screen. There @psychon's method (without optimizations) became unusable.

grafik

The fancy fill pattern was only displayed some time after finishing the stroke. Before it was just blue like in @cuppajoeman 's method.
One difference: @psychon's method does not take into account the pressure. That's why there are no artifacts near the end of the strokes.

rolandlo:
I also tested the two methods on "Wayland on X". Here @cuppajoeman's method worked flawlessly (no lag felt, no artifacts) whereas @psychon's method (with not optimizations applied) was unusable even on a smaller screen because of the high lag. 

I will test @cuppajoeman's method while recording the screen and try to find the differences between what X11 does and what wayland does.

Technius:
What's the status of this PR?

rolandlo:
> What's the status of this PR?

It's WIP, that was abandoned for some time because of the focus on the 1.1.0 release. We have to solve some performance issues and properly integrate this feature into the GUI. The performance will also have to be tested on say MacOS Catalina, where normal drawing/writing is still fine, but slower than on Linux and Windows.

rolandlo:
@cuppajoeman I'd like to resume working on this feature. Would you like to collaborate on it further?

rolandlo:
I have tested @cuppajoeman 's method on MacOS Catalina. It wasn't quite performant enough, but reasonably close, so it might become so after some opimizations.
@psychon's method was much too slow, but there may be more room for improvement when not recreating the whole cairo surface each time.

LittleHuba:
@rolandlo, will you continue on this PR?

rolandlo:
> @rolandlo, will you continue on this PR?

Not in the foreseeable future. I may possibly resume when the setsquare/compass are finished and the spline approximation is finished as well, although even then there are things to do with higher priorities. Feel free to close the PR for now.

LittleHuba:
Then I'll close this. If anybody wants to take up this work, feel free to do so.

cuppajoeman:
> @cuppajoeman I'd like to resume working on this feature. Would you like to collaborate on it further?

Going through some of my old PR's and remembered this one, I would like to continue with this PR, but I'll have to get some context about where it was at. I think we were at a point where I had an algorithm, but we were checking to see if it was performant enough?

rolandlo:
@cuppajoeman Sorry for the late reply. The code is still in this PR. Can you please rebase on current master? Then I can test the performance.

cuppajoeman:
@rolandlo No problem, and my apologies for the same. I can try to rebase on the master branch, maybe the performance can be better by segmenting it as we draw so it only draws small pieces at a time, but I can't remember if xournal++ already does this.

PR #2704, opened 2021-01-25
Remove console window flash at startup on Windows

jaburjak:
I have noticed that the Windows build uses `AllocConsole` at Xournal++ startup to prevent `gspawn-win64-helper` from creating its own console popups, for example when rendering LaTeX. However, this makes the console window flash briefly on the screen, which is annoying especially in tablet mode. When you launch Xournal++ (using the start menu or by opening a file associated with it), a console window is opened and then hidden immediately.

In tablet mode, there is no classic desktop, instead a full-screen start menu is displayed when no window is in the foreground. This means that after hiding the console window, the screen drops back to the start menu. Then the actual Xournal++ GUI is opened in the background and you must switch to it by tapping its icon in the task bar.

Unfortunately there is no way to force `AllocConsole` to create an already-hidden console window, so if we want to solve this, we have to spawn our own console process manually and attach to it instead. `CreateProcess` allows us to start `cmd.exe` hidden, so it won’t flash on the screen. Then we can attach to it using `AttachConsole`. That’s what I’m doing in this PR.

I know this adds a lot of code, but when I noticed that #2606 will remove the launcher (which caused a similar issue, except it was the launcher that flashed on the screen), I thought it would be nice to get rid of this annoying behavior entirely.

Thanks for considering my PR!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -89,6 +89,14 @@ if (WIN32)
   unset (ICON_SIZES)
   unset (ICON_DEPS)
   add_custom_target (xournalpp_icon DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/win32/xournalpp.ico)
+
+  # Concatenate SOURCES lists
+  set (xournalpp_SOURCES ${xournalpp_SOURCES_RECURSE} ${xournalpp_SOURCES})
+  unset (xournalpp_SOURCES_RECURSE)
+
+  file (GLOB xournalpp_SOURCES_RECURSE
+    win32/*.cpp
+  )
 endif()
 
 # Concatenate SOURCES lists
--- a/src/Xournalpp.cpp
+++ b/src/Xournalpp.cpp
@@ -18,14 +18,13 @@
 #include "Stacktrace.h"
 
 #ifdef _WIN32
-#include <windows.h>
+#include "win32/console.h"
 #endif
 
 auto main(int argc, char* argv[]) -> int {
 #ifdef _WIN32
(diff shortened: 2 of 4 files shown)


Technius:
Thanks for the PR, I couldn't find a nice way to get rid of the console popups (I'm not too familiar with Windows APIs). I think we'll want to wait until #2606 is merged first, then we can move this code over to the new `src/win32` folder to keep the code a bit cleaner.

rolandlo:
Just a notice: PR #2606 has already been merged, so moving this code to `src/win32` as suggested by @Technius is possible now.

kubajaburek:
Thanks, I moved the code.

LittleHuba:
We are currently working on some bigger PRs that block all other PRs. We'll get back to your PR as soon as these are finished. Thanks in advance for your patience.

Febbe:
Is this also required, when using 
`add_executable(<name> WIN32 <sources>)`?

And what happens, if I want to start xournalpp from console with the calling console connected?

Febbe:
@Technius ok, even when the property in cmake fixes that, it will be an improvement then. Currently it does not attach to the current console.

Technius:
Sorry for the delay. I tested this and it seems to work fine in a Windows 10 virtual machine. We just need to rebase this (there shouldn't be any conflicts).

jaburjak:
Thanks for testing it. I rebased the patch onto the lastest master.

Febbe:
@technius, I think we can add this then to 1.1.1

LittleHuba:
LGTM merge in 24 hours

PR #2707, opened 2021-01-26
Make circle go up to cursor when drawing from center

tim-clifford:
When drawing a circle from the center, it makes sense to put the edge of the circle wherever the cursor is, but the current implementation is to take the maximum of the width and height, which corresponds to the cursor only at the principal axes, and makes a smaller circle everywhere else.

Apologies if there's an issue or PR about this already, there were a lot about circles to look through!

### DIFF
--- a/src/control/tools/CircleHandler.cpp
+++ b/src/control/tools/CircleHandler.cpp
@@ -45,7 +45,8 @@ void CircleHandler::drawShape(Point& c, const PositionInputData& pos) {
         {
             int signW = width > 0 ? 1 : -1;
             int signH = height > 0 ? 1 : -1;
-            width = std::max(width * signW, height * signH) * signW;
+            width = (this->modControl) ? sqrt(pow(width, 2) + pow(height, 2)) :
+                                         std::max(width * signW, height * signH) * signW;
             height = (width * signW) * signH;
         }
 


Technius:
Thanks for this PR! There's one problem, though. The behavior with <kbd>Shift</kbd> but no <kbd>Ctrl</kbd> has become less intuitive:
xpp_circle_new
Compare this to the old behavior:
xpp_circle_old

So I think we want to keep the existing max behavior when <kbd>Ctrl</kbd> is not held down.

tim-clifford:
Good spot! That's fixed now.

src/control/tools/CircleHandler.cpp
```diff
@@ -45,7 +45,11 @@ void CircleHandler::drawShape(Point& c, const PositionInputData& pos) {
         {
             int signW = width > 0 ? 1 : -1;
             int signH = height > 0 ? 1 : -1;
-            width = std::max(width * signW, height * signH) * signW;
+            if (this->modControl) {
+                width = sqrt(pow(width, 2) + pow(height, 2));
+            } else {
+                width = std::max(width * signW, height * signH) * signW;
+            }
```
rolandlo, line 52: This can be expressed in one line with:
```suggestion
            width = (this->modControl) ? sqrt(pow(width, 2) + pow(height, 2)) : std::max(width * signW, height * signH) * signW;
```

tim-clifford: Yes, I was aware of this. However that would take the line length to 128 characters, which makes it much more difficult to read. Having a look at the rest of the source, it seems that the convention is to stick to 120 characters or less. Perhaps a good middle ground is the following (although they will compile to the same thing anyway iirc):
```suggestion
            width = (this->modControl)
                    ? sqrt(pow(width, 2) + pow(height, 2))
                    : std::max(width * signW, height * signH) * signW;
```

rolandlo: You can just clang-format your changes (which you must do anyway). The result is spread on two lines, separated by the double point (which is still on the first line).

tim-clifford: Ah, I'm sorry, I was not aware of this.

Technius:
LGTM, merging.

PR #2712, opened 2021-01-27
make vim keybindings behave exactly like arrows

kaymmm:
This just makes the hjkl keys emulate the arrow keys instead of having separate behavior that doesn't also work in presentation view.


Technius:
Tested and works as expected, thanks for the PR.

PR #2721, opened 2021-01-28
fix touch drawing with pen

### LINKED ISSUE #2123: Drawing with touch works with highlighter, other tools, but not pen tool
**Affects versions :**
 - OS: Linux (arch)
 - (Linux only) Desktop environment: X11, XFCE, OpenBox
 - Which version of libgtk do you use: unknown (but arch is up to date)
 - Version of Xournal++: 1.0.18, commit 535d261d
 - Installation method: pacman

**Describe the bug**
Drawing with touch is not working. I have a Lenovo Yoga 460, which has a version of the touch screen that is not compatible with most styli. I bought a stylus that looked like it might be compatible, but it's not, and instead the stylus works like touching with my finger. I still liked it more than the Lenovo stylus so I tried to set it up to draw with touch. I can draw the highlighter with this stylus (and my finger), and the eraser works too, as well as other tools I've tried, but the pen tool does not write with my finger (or this stylus).

**To Reproduce**
I tried toggling a bunch of settings:
Input System -> Enable new input system: Checked or unchecked (same result either way)
Input System -> Input Devices: Tried assigning the touchscreen to pen, as well as the default setting (same result either way)
Touchscreen -> Enable GTK Touch / Scrolling workaround: Checked or unchecked (when checked, I was able to use highlighter, eraser, lasso select, with finger, but not pen with finger. When unchecked, I couldn't use any tools with finger.)

**Expected behavior**
I expected to be able to draw with my finger.

**Additional context**
I am hesitant to submit this issue because it is a much lower priority for me than customizable keyboard shortcuts (issue 919), and I am hoping this doesn't distract a contributor from that, but I feel it is my duty to submit this bug report to help someone fix it some day!

rolandlo:
Fixes #2123.
Thanks to @bhennion for finding the cause of the problem.

### DIFF
--- a/src/gui/inputdevices/InputEvents.cpp
+++ b/src/gui/inputdevices/InputEvents.cpp
@@ -103,17 +103,19 @@ auto InputEvents::translateEvent(GdkEvent* sourceEvent, Settings* settings) -> I
         gdk_event_get_keyval(sourceEvent, &targetEvent.button);
     }
 
-    // Copy the event sequence if there is any
-    if (sourceEventType == GDK_TOUCH_BEGIN || sourceEventType == GDK_TOUCH_UPDATE || sourceEventType == GDK_TOUCH_END ||
-        sourceEventType == GDK_TOUCH_CANCEL) {
-        targetEvent.sequence = gdk_event_get_event_sequence(sourceEvent);
-    }
 
     // Copy the timestamp
     targetEvent.timestamp = gdk_event_get_time(sourceEvent);
 
     // Copy the pressure data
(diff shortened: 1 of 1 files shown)


bhennion:
Argh! It actually does not seem to work on my device. Let me have a closed look.

bhennion:
> Argh! It actually does not seem to work on my device. Let me have a closer look.

Well, I think it's a different bug (maybe not in xournal). The thing is: it does not work when I use the touch screen, but it does work when I replay the touch sequence using libinput. That's fairly weird but surely unrelated to this PR.

I think you can merge this.

Issue #2718, opened 2021-01-28
Enable mirroring by either a button or transformation


coldspark29:
There is no possibility of creating a mirror image automatically. Here I am trying to transform the drawing to flip it around to get the mirror image for the discrete convolution, but I have to redraw it. I think there should be a button for this and the transformation should also support this.

![mirror](https://user-images.githubusercontent.com/19779511/106100172-30d30580-613c-11eb-86b6-aa3b59c514bf.gif)

redweasel:
yes please! This is my most missed feature by far! (could potentially be a simple fix and a good first issue to tackle)

rolandlo:
It is not so trivial, since there are a lot of calculations made for snapping and there are rotation, stretching, translation and undo actions involved, and previews are displayed, which all must work. I see an approach to simplify the currently used code, which would make it easier. But it is definitively not on the difficulty level of a good first issue. In Xournal the approach is much simpler. There is no rotation and no previews while transforming the contents are displayed.

redweasel:
Ok I'm already looking into it, and came across some very strange code (gigantic if statements in a place where there really should not be a branch). I am already trying to simplify it down. I have worked on similar code in blender already so I hope to make a good contribution here too.

coldspark29:
Nice work!

Issue #1440, opened 2019-08-23
making mirror image with resize tool


ehremington:
In the old xournal, you could take a selection and by grabbing and dragging one side of the selection box, you could mirror and resize the image. This seems like a minor thing, but I have found myself missing this feature in xournal++ in my physics class. Maybe hard to introduce but it is the only place that I've found to be a regression from the original program.

And while I'm hear let me express my great thanks to everyone working on the project! The addition of a rotation tool to selections was a huge improvement over the original and all of the update going here have been wonderful. So glad this project has taken off!

JJones780:
I can't remember the exact problem, but there might even be a blocking issue to this enhancement. Maybe it wasn't worth doing until rotation was cleaned up?

LittleHuba:
If I remember correctly it was that the file format does not support it. You mentioned transformation matrices as a possible solution.

JJones780:
Yes, I believe there are a few issues which would all be fixed by using ( and storing ) a matrix transform instead of recalculating points and thicknesses.
Off the top of my head I think it included
 - allow the rotation of images ( and other imported objects )
 - editing already rotated objects ( using the handles )
 - editing already scaled objects (using the handles )
 - mirroring objects using scale handles.

Technius:
Related to #918.

rolandlo:
Cloning a selection is already possible. Just right click with your mouse or use pen button 2 (you can remap pen buttons with `xsetwacom` if needed). So I will close this issue.

ehremington:
Wait cloning a section was not what this feature request was about. I would like the a functionality that was present in the original xournal to be present if possible in xournal++. Let me describe again the feature: Upon making a selection, the handle on the side of the selection that allow resizing should also be able to be dragged "through" the unmoving side of the selection to produce a mirror image. This may seems like a minor thing, but in my experience teaching a live physics class with many drawings, I found that I missed it upon moving from xournal to xournal++. I believe that this feature is low priority but is closing this issue the same as a "won't fix"?

rolandlo:
Sorry I misunderstood. I agree, that this would be useful.

rolandlo:
@ehremington Am I assuming correctly that you only need this mirroring feature for strokes? For images, TeXImages and text it would be much harder to implement, I think. Xournal also only mirrors strokes and just moves and scales the other elements.

PR #2723, opened 2021-01-28
add EditSelection mirror functionality for strokes and images

redweasel:
fixes https://github.com/xournalpp/xournalpp/issues/2718
fixes https://github.com/xournalpp/xournalpp/issues/1440

Simplifies the edit selection logic. Then adds in the mirror functionality, but
just like `aspectRatio` is fixed for some objects, also `mirror` is fixed for some objects.

The handling of the corners is now linear for all cases. Before this PR it was non linear for the corners, resulting in weird behavior when the mouse is not exactly on the grabbed corner.

Also adjusted cursors to match the mirrored scale directions in the corners.

### DIFF
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -27,6 +27,9 @@
 
 constexpr size_t MINPIXSIZE = 5;  // smallest can scale down to, in pixels.
 
+constexpr int DELETE_PADDING = 20;  // ui button padding
+constexpr int ROTATE_PADDING = 8;
+
 EditSelection::EditSelection(UndoRedoHandler* undo, const PageRef& page, XojPageView* view):
         snappingHandler(view->getXournal()->getControl()->getSettings()),
         x(0),
@@ -114,6 +117,7 @@ void EditSelection::contstruct(UndoRedoHandler* undo, XojPageView* view, const P
     this->sourceLayer = this->sourcePage->getSelectedLayer();
 
     this->aspectRatio = false;
+    this->mirror = true;
 
     this->relMousePosX = 0;
     this->relMousePosY = 0;
@@ -130,6 +134,7 @@ void EditSelection::contstruct(UndoRedoHandler* undo, XojPageView* view, const P
 
     cairo_matrix_init_identity(&this->cmatrix);
     this->view->getXournal()->getCursor()->setRotationAngle(0);
+    this->view->getXournal()->getCursor()->setMirror(false);
 }
 
 EditSelection::~EditSelection() {
@@ -294,6 +299,10 @@ void EditSelection::addElement(Element* e, Layer::ElementIndex order) {
         this->aspectRatio = true;
     }
 
+    if (!e->rescaleWithMirror()) {
+        this->mirror = false;
+    }
+
     if (e->getType() != ELEMENT_STROKE) {
         // Currently only stroke supports rotation
         supportRotation = false;
@@ -367,10 +376,12 @@ void EditSelection::mouseMove(double mouseX, double mouseY, bool alt) {
         // find corner of reduced bounding box in rotated coordinate system closest to grabbing position
         double cx = this->snappedBounds.x;
         double cy = this->snappedBounds.y;
-        if (this->relMousePosRotX > this->snappedBounds.width / 2) {  // closer to the right side
+        if ((this->relMousePosRotX > this->snappedBounds.width / 2) ==
+            (this->snappedBounds.width > 0)) {  // closer to the right side
             cx += this->snappedBounds.width;
         }
-        if (this->relMousePosRotY > this->snappedBounds.height / 2) {  // closer to the lower side
+        if ((this->relMousePosRotY > this->snappedBounds.height / 2) ==
+            (this->snappedBounds.height > 0)) {  // closer to the lower side
             cy += this->snappedBounds.height;
         }
 
@@ -391,8 +402,7 @@ void EditSelection::mouseMove(double mouseX, double mouseY, bool alt) {
 
         // move
         moveSelection(p.x - cx, p.y - cy);
-    } else if (this->mouseDownType == CURSOR_SELECTION_ROTATE && supportRotation)  // catch rotation here
-    {
+    } else if (this->mouseDownType == CURSOR_SELECTION_ROTATE && supportRotation) {  // catch rotation here
         double rdx = mouseX / zoom - this->snappedBounds.x - this->snappedBounds.width / 2;
         double rdy = mouseY / zoom - this->snappedBounds.y - this->snappedBounds.height / 2;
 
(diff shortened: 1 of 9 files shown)


tattsan:
Thanks, this is the feature I've been waiting for! I tried the PR, and find the following phenomenon.

https://user-images.githubusercontent.com/6355486/106225136-dfa04000-6227-11eb-9528-3fa41b120234.mp4

If I continue extending the reflected selection, it suddenly jumps to next page.

rolandlo:
Wow, great work so far! 

I have noticed that snapping does not work properly while moving mirrored selections. You can test it out with a rectangle. The corner closest to the position where you grabbed the selection should get snapped to grid. It still works when the selection is not mirrored.

redweasel:
Also found https://github.com/xournalpp/xournalpp/issues/1440 which is the same feature request. In the newest version of this patch, I also fixed an issue with the selection not scrolling the page correctly when rotated (along with the here reported issue, as they were related).

Good that I initially though this would be trivial. Otherwise I would probably not have done it.

Still working on getting clang-format work... I have clang-format-11 installed.

tattsan:
There is also #2653 in 'Discussions'.  This discusses not continuous deformation, but -100% transformation by one operation.

redweasel:
> There is also #2653 in 'Discussions'. This discusses not continuous deformation, but -100% transformation by one operation.

I don't think adding the buttons is in the scope of this pr. I will soon (after this merged) make another one for the mirror buttons.

tattsan:
I've used this feature for the last few days with no issue.

> I will soon (after this merged) make another one for the mirror buttons.

Thank you so much!

Technius:
Is this backwards compatible? If not, we'll have to defer it to after the 1.1.0 release (ETA: end of this month).

redweasel:
Yes, I tested it and the mirrored strokes (as expected) are just strokes like normal. Even the mirrored images behave normal so they are being rendered correctly, you are just not able to "unmirror" them. I think I would like to see it in 1.1.0, but I am mostly using self compiled versions anyway.

tattsan:
I've tested the .xopp file with Xournal++ of older version.  There's a one different behavior. 
- Select the vertically mirrored image object.
- Click the bottom handle, and pull it down.
- Then the top of the selection will be extended upwards.

If you do the same operation with the new Xournal++, the bottom of the selection will be extended downwards. Only image files show this difference. Pen strokes are normal.

rolandlo:
The mirrored image sometime disappears from the screen after unselecting it. 
Moreover the snapping issue I mentioned above (wrong corner is snapped when moving mirrored image) is still unresolved.
It would be nice to have the mirror functionality in the 1.1.0 release, but there are several issues that need to be solved yet.

redweasel:
@rolandlo Oh now I see. Seems like I never understood snapping enough! I fixed that snapping did entirely not work, but missed that it's different for each corner. Will fix it in a moment!

redweasel:
I fixed it. It is now using the correct corner when translating the mirrored selection. I did however find some more unstable snapping issues not related to this PR while testing. Will report them now.

redweasel:
The fix for image rendering was a bit harder. I introduced getters for `getVisualX/Y/Width/Height` into `Element`. That looked like the most reliable way to do this. Hope I didn't miss a spot to swap the normal getter for the visual getter.

I have to retract my claim that this is backwards compatible. It is almost backward compatible, but images mirrored images will only render on full page rerenders in older versions.

redweasel:
This should be ready for a final review. I disabled mirror for everything except for strokes for now.

src/control/tools/EditSelection.cpp
```diff
@@ -407,136 +414,73 @@ void EditSelection::mouseMove(double mouseX, double mouseY, bool alt) {
...
+            int xMul = (xSide + 1) / 2;
+            int yMul = (ySide + 1) / 2;
+            double xOffset =
+                    (rx - this->x) - this->width * xMul;  // x-offset from corner/side that is used for resizing
+            double yOffset =
+                    (ry - this->y) - this->height * yMul;  // y-offset from corner/side that is used for resizing
+
+            // calculate scale factor using dot product
+            double f = (xOffset * nx + yOffset * ny + diag) / diag;
+            f = std::copysign(std::max(std::abs(f), minSize / std::min(this->width, this->height)), f);
```
rolandlo, line 459: I thinks this needs correction. You will see the difference if your selection consists of an ellipse with a large eccentricity.
```suggestion
            f = std::copysign(std::max(std::abs(f), minSize / std::min(std::abs(this->width), std::abs(this->height))), f);
```

src/control/tools/EditSelection.cpp
```diff
@@ -731,20 +688,20 @@ auto EditSelection::getSelectionTypeForPos(double x, double y, double zoom) -> C
         return CURSOR_SELECTION_BOTTOM_RIGHT;
     }
 
-    if (x1 - (20 + this->btnWidth) - BORDER_PADDING <= x && x1 - (20 + this->btnWidth) + BORDER_PADDING >= x &&
-        y1 - BORDER_PADDING <= y && y1 + BORDER_PADDING >= y) {
+    if (xmin - (20 + this->btnWidth) - BORDER_PADDING <= x && x <= xmin - (20 + this->btnWidth) + BORDER_PADDING &&
```
rolandlo, line 691: Could you please define `constexpr` for these magical numbers 20 and 8 which are used (here and elsewhere) for the position of the delete and rotate rectangles? That would make it more readable.

rolandlo (changes requested):
Except of the two minor change requests I made everything seems to work very well. I have tested copying/pasting, cloning, undo/redo, save/load, snapping, rotation, resizing and moving. For undo/redo to work properly I just needed to rebase on current master.
I love your work, both for the mirror capability (which I will use in geometry class often) and for the simplifications in the code.

So please make the two requested changes and rebase on current master.

rolandlo (approved):
Ok, great! Thanks for this contribution. Do you want to squash any commits (or make other changes) before I merge it?

redweasel:
Just to be sure, what is the general rule around here for commits in a PR? In other projects it was always one commit per feature. I would squash the commits If that's preferred here.

rolandlo:
We often have only one commit per feature. If the commits (of a larger PR) are nicely separated and only take care of one problem at a time we will accept multiple commits. Fixups of previous commits like "Apply clang format" or "Fix the thing I forgot previously" should definitely get squashed away. I think for this PR it would make sense to squash it all into one commit. If you would prefer to have the fix of automatic scroll separated from the rest, this would probably justify to squash the PR into two commits. 
Note that we have sometimes forgotten to squash the commits before we merged it. So if you look at the commit history you will find exceptions to what I just wrote.

redweasel:
wait a sec, do not merge now. I made a mistake!

redweasel:
Ok the PR is ready for merge now.

rolandlo:
Thanks again. Great job!

danielgomez3:
I'm not sure how to flip my selection in xournalpp.. apparently this feature was added? I'm using Xournalpp 1.2.3 on NixOS.

rolandlo:
> I'm not sure how to flip my selection in xournalpp.. apparently this feature was added? I'm using Xournalpp 1.2.3 on NixOS.

For what was implemented in this PR, see the screen recording https://github.com/xournalpp/xournalpp/pull/2723#issuecomment-769538152

There is a Lua plugin for flipping selections of strokes and a Lua plugin for flipping images. What are you interested in?

PR #2724, opened 2021-01-29
Fix font size dpi scaling bug and search mark bug and add a plugin for migration of font sizes

rolandlo:
This PR replaces PR #2182. 
It does not take care of line spacing in pdf export (for which we would need Pango's `pango_layout_set_line_spacing` to work correctly). 

Currently the plugin will derive the scaling factor from the display dpi setting (in the zoom panel of the preferences). This will usually be the correct choice, but it is not if the user has changed his display dpi setting since he created the document. So I think a display dpi that can be set from the GUI should be included. This could be done either by adding a GUI element in the Lua plugin API or by creating the GUI element in the plugin using the `lgi` module. In the latter case it may make sense to add the `lgi` module to the project. It is always needed when creating a non-trivial GUI.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,6 +9,7 @@ Homepage: https://github.com/xournalpp/xournalpp/
 Package: xournalpp
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2-2 (>= 2.40)
+Recommends: lua-lgi
 Suggests: texlive-base, texlive-latex-extra
 Description: Xournal++ - Open source hand note-taking program
  Xournal++ is a hand note taking software written in C++ with the target of 
--- a/plugins/MigrateFontSizes/dialog.glade
+++ b/plugins/MigrateFontSizes/dialog.glade
@@ -1,0 +1,341 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
+  <object class="GtkAdjustment" id="adjOldDpi">
+    <property name="lower">13</property>
+    <property name="upper">400</property>
+    <property name="value">223</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjScaleFactor">
+    <property name="lower">0.17999999999999999</property>
+    <property name="upper">5.5499999999999998</property>
+    <property name="value">3.0899999999999999</property>
+    <property name="step_increment">0.01</property>
+    <property name="page_increment">1</property>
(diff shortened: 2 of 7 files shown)


rolandlo:
I have added a GUI dialog using the lgi Lua module. It must be installed with `sudo luarocks install lgi` on Linux. Can we somehow bundle the lgi-module with Xournal++? It would be useful for creating plugins with a non-trivial GUI.

rolandlo:
If you want to see how it works, here is a screen recording:

https://user-images.githubusercontent.com/40485433/106520338-eefbe380-64dc-11eb-8c75-4559ecb75f8b.mp4

Technius:
This is awesome!

> I have added a GUI dialog using the lgi Lua module. It must be installed with `sudo luarocks install lgi` on Linux. Can we somehow bundle the lgi-module with Xournal++? It would be useful for creating plugins with a non-trivial GUI.

It seems like `lgi` needs some native code bindings, so it would be hard to bundle it with Linux at least (and wouldn't work well with distro packagers). On Linux, it would be better to add it as an optional/recommended dependency in the package file (e.g. `lua-lgi` on Debian, Ubuntu, Fedora). If lgi cannot be found at runtime, the dialog could show a warning dialog.

We could bundle it on Flatpak, Mac, and Windows without problem, though.

rolandlo:
I have implemented the warning dialog for the case when the lgi module is not found:
grafik

rolandlo:
The lgi-module via `sudo luarocks instlal lgi` was in `/usr/local/share/lua/5.3`, the lgi module installed via `sudo apt install lua-lgi` is found in `/usr/share/lua/5.3/` I will add both to the path.

rolandlo:
I have improved the warning dialog to show the full path of the main.lua file and to add some spacing to make it less crowded. Moreover the lua lgi module is now (by default) searched in both `/usr/share/lua/5.3` and `/usr/local/share/lua/5.3`

rolandlo:
> We could bundle it on Flatpak, Mac, and Windows without problem, though.

Unfortunately the `lgi`-module is only available for Lua 5.1 in MinGw64 (on Windows), see [here](https://packages.msys2.org/base/mingw-w64-lua-lgi). This is a problem, since we use Lua 5.3

Technius:
We could patch the lgi package in msys to support Lua 5.3 (preferably as a PR on the MINGW-packages repo). [For example,`lua-lpeg` supports both 5.1 and 5.3](https://github.com/msys2/MINGW-packages/blob/821c4ab6f2741b07b58191b4f1a503989fe0f68b/mingw-w64-lua-lpeg/PKGBUILD).

rolandlo:
> We could patch the lgi package in msys to support Lua 5.3 (preferably as a PR on the [package repo](https://github.com/msys2/MINGW-packages)). [For example,`lua-lpeg` supports both 5.1 and 5.3](https://github.com/msys2/MINGW-packages/blob/821c4ab6f2741b07b58191b4f1a503989fe0f68b/mingw-w64-lua-lpeg/PKGBUILD).

I don't see why you think lua-lpeg supports lua 5.3. 

In any case the font size migration can also be done without the nice lgi GUI. It's enough to use the DisplayDPI from the preferences and the direct scaling from the first menu entry of the plugin.

So is there still something we should improve before this can get merged?

Technius:
Okay, fair enough. Looks fine to me for now--although I have some reservations about the text scaling directly part of the plugin API. I suppose we can delete it at some later point when migration is no longer necessary.

rolandlo:
Fixes #1118

Issue #2332, opened 2020-10-21
Latex makes crash the App


ErgastiAlex:
**Affects versions :**
 - OS: Windows
 - libgtk: 3.24.23
 - Xournal: 1.1.0+ dev
 - Installation method: Official Release, nigthly Build.

 **Description**
When open the latex window if i press "Cancel" before the image of the example x^2 loads the app'll crash

**To Reproduce**
1- Open Math TEX
2- Press immediately cancel or the X symbols

**Expected behavior**
Simply close the windows without make the app crashes.

**Screenshots of Problem**
![image](https://user-images.githubusercontent.com/20249175/96725859-ab5bee80-13b1-11eb-8fb9-9c0292474e24.png)
Press now the X button or "Annulla"

**Additional context**
I've follow the instruction to add the default_template.tex, i also have MikTex installed.

rolandlo:
Instead of pressing cancel or the X symbol you can also press the ESC key to get this crash.

Technius:
Simplest but dirtiest fix is to disable the close button in the dialog.

The proper fix would be to cancel the render finish callback (the pdflatex process callback) if the dialog is closed.

PR #2728, opened 2021-01-30
fix crash when closing the latex dialog early

redweasel:
fixes https://github.com/xournalpp/xournalpp/issues/2332 by aborting the render instead of calling an assert to crash the program deliberately.

### DIFF
--- a/src/control/LatexController.cpp
+++ b/src/control/LatexController.cpp
@@ -149,7 +149,8 @@ void LatexController::triggerImageUpdate(const string& texString) {
         XojMsgBox::showErrorToUser(this->control->getGtkWindow(), err->message);
     } else if (auto** proc = std::get_if<GSubprocess*>(&result)) {
         updating_cancellable = g_cancellable_new();
-        g_subprocess_wait_check_async(*proc, updating_cancellable, reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), this);
+        g_subprocess_wait_check_async(*proc, updating_cancellable,
+                                      reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), this);
     }
 
     updateStatus();
@@ -182,8 +183,7 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
             // the render was canceled
             g_error_free(err);
             return;
-        }
-        else if (!g_error_matches(err, G_SPAWN_EXIT_ERROR, 1)) {
+        } else if (!g_error_matches(err, G_SPAWN_EXIT_ERROR, 1)) {
             // The error was not caused by invalid LaTeX.
             string message =
                     FS(_F("Latex generation encountered an error: {1} (exit code: {2})") % err->message % err->code);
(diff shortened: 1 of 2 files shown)


rolandlo:
It still crashes if I press ESC shortly (maybe 1/2 second) after opening the LaTeX dialog. Your check does not make Xournal++ return then and signal 11 is emitted while calling `self->dlg.getBufferContents`

redweasel:
Ok I have tried to reproduce your crash, but in my build it just does not want to crash from that dialog anymore(with this patch), no matter what I try. Maybe someone else has to continue on this.

OS: Linux Debian testing
DE: KDE Plasma 5.20.5

Technius:
A better solution would be to connect to the dialog's delete event and prevent the event from propagating when the latex is updating.

redweasel:
would stopping the dialog's delete event not stop it from reacting to the close attempt? I would rather go with a more responsive approach of just invalidating/stopping the render task.

Technius:
> I would rather go with a more responsive approach of just invalidating/stopping the render task.

Yes, this is the best approach overall. You will need to cancel the render callback entirely (triggered by `LatexGenerator::asyncRun` on line 147) when the dialog is closed. If the callback is called after the dialog is closed, the application will crash because `LatexController` will already have been destroyed.

redweasel:
Did anybody test the updated version? @rolandlo does it still crash for you?

rolandlo:
Just tested it. It doesn't crash any more.

rolandlo:
I have to correct: It doesn't crash anymore, but the OK button is always disactivated. So I can't insert any LaTeX formulas into the document any more.

rolandlo:
It works fine now. Is this ready to be merged?

Technius:
I think this is relying on undefined behavior to work. I managed to cause a crash by pressing ESC quickly enough.

src/control/LatexController.cpp
```diff
@@ -166,22 +169,28 @@ void LatexController::handleTexChanged(GtkTextBuffer* buffer, LatexController* s
 }
 
 void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, LatexController* self) {
-    g_assert(self->isUpdating);
+    if (!self->isUpdating()) {
+        return;
+    }
```
Technius, line 174: The destructor is being called before the callback is invoked here, so `self` has technically already been deallocated. The check here therefore should not access any members of `self`.

One clean alternative is to pass in a pair of `LatexController*` and `GCancellable*` to `g_subprocess_wait_check_async` and do the cancel check here.

Technius (changes requested):
Ok, so it doesn't crash but I confirmed that it's using undefined behavior, which can lead to unpredictable crashing/breakage on different systems.

src/control/LatexController.cpp
redweasel: I don't think I even need that check now, since the canceled error will be thrown on the previous render attempt, so only one render at a time is running, right?

Technius (approved):
LGTM. Anything else you want to change before we merge?

redweasel:
No, I'm fine with this. To be sure, I will keep an eye out on the bug tracker.

Technius:
Great, then I'll merge this now.

Issue #997, opened 2019-03-19
Eraser doesn't work with marker


jclsn:
When I use the eraser on paint drawn with the marker, it rearranges the colour in a very strange way, instead of erasing it.

LittleHuba:
Please provide a video or screenshot.

coldspark29:
Before erasing:
![image](https://user-images.githubusercontent.com/19779511/54666960-80315f00-4aeb-11e9-916d-f1a30d92708c.png)

While erasing:
![image](https://user-images.githubusercontent.com/19779511/54666969-86274000-4aeb-11e9-9107-dc734c6ecaff.png)

After letting go of the eraser:
![image](https://user-images.githubusercontent.com/19779511/54666976-8a535d80-4aeb-11e9-8134-e858fad77a8f.png)

masac:
https://drive.google.com/file/d/1mp4jXg66ZphWJN0f3P_cQIXqbd7zUBU6/view?usp=sharing

eliasrg:
I am not familiar with the implementation details, but this seems like a difficult issue to fix directly, without completely losing the overlap behaviour of highlighter strokes:
![997-erase-marker-1](https://user-images.githubusercontent.com/7657466/55954560-41607600-5c5f-11e9-9387-261e5fecc674.png)
Whiteout seems to be the best tool for erasing parts of a highlighter stroke, but unfortunately it also whites out the background. Maybe that is worth changing?

LittleHuba:
The bug is twofold:
- The rendering while erasing is broken
- The gap in the stroke created by the eraser is to small it should create a gap with the following rule:
All points of the stroke within a distance of `1/2 (eraserWidth + strokeWidth)` of the eraser path will be removed

jvwowd:
@eliasrg  I have a concern about changing the whiteout as suggested. Maybe there is another way, but currently using whiteout is the only way I know how to remove part of a background PDF when annotating which I sometimes find useful.

For example, when taking notes for a lecture, it is sometimes useful to duplicate a page, and remove part of the background to show more work.

eliasrg:
@jvwowd One way would be to use the pen with a white colour, enabling fill, and making sure to set "fill transparency" to 100% in Tools/Pen Options. It might even be more convenient than using whiteout!

jvwowd:
[...]

Issue #2578, opened 2020-12-21
Spline interpolation of strokes for better quality rendering


bhennion:
This follows up on #2508 and is related to #2320 (and the associated PR #2512).

**Is your feature request related to a problem? Please describe.**
When drawing fast with a high latency device, we can see the line segments forming the stroke
For instance, this image was generated with my touchpad (in Xournal++)
![segments](https://user-images.githubusercontent.com/17818041/102780709-afe35a00-4396-11eb-8c15-4bc0393eafcb.png)

**Describe the solution you'd like**
Instead of piecewise linear strokes, we could have [splines](https://en.wikipedia.org/wiki/Spline_(mathematics)). From the above input, inkscape generated this spline
![spline](https://user-images.githubusercontent.com/17818041/102780931-15374b00-4397-11eb-991c-7454da871a6e.png)

The algorithm inkscape used here is extremely simple, but there exist way more evolved ones. I'll list some references below.
To make the rendering of those strokes as good as possible, we would need to keep in mind that they are splines. We should allow for spline segments of various degrees in strokes (degree 1 = line segment, degree 2 = quadratic spline segment, degree 3 = cubic spline segment).

**Benefits:**

1. Rendered strokes would be perfectly rescalable (and zooming in to much would not show angular points)
2. This paves the way for optimization by approximation algorithms. Those can drastically reduce the number of spline segments needed to encode a stroke, therefore cutting the file size. Consider those two .svg files contained in this archive [svg.zip](https://github.com/xournalpp/xournalpp/files/5724555/svg.zip). They contain the same sentence (recorded using libinput). One is generated by xournal++ and has size 297 Kib The other has been generated with Krita's calligraphy tool (which is vectorial) and has size 68 Kib. This means a **77%** decrease in file size for a similar (or better...) output quality.
3. This would allow an implementation of smoothing/stabilizing (as in #2320 and #2512) using algorithms like [this one](https://en.wikipedia.org/wiki/Smoothing_spline)

**Road map:** In practice, this breaks into smaller steps

- [ ] Adapt the Stroke class,
- [ ] Extend the .xopp file format.
- [ ] Research further and implement 2.
- [ ] Research further and implement 3.

**Some references:**

- A [library](https://people.math.sc.edu/Burkardt/cpp_src/spline/spline.html) for interpolation/approximation by splines
[...]

Issue #2233, opened 2020-09-12
Packaged File Format Project Status


Geo25rey:
I'm interested in contributing to Packaged File Format Project, but it seems to have not been active for about a year and it's unclear what the next step is further progression. It looks @LittleHuba was working on it, so I'd like your input on where I can begin. 

Based off of what has been discussed in the issues marked as "done", it seems like there is no detailed specification for the proposed file format. Would that be a good starting point?

LittleHuba:
@Geo25rey sorry for the late answer. If you are still interested in helping, please reach out to me on Matrix (@littlehuba:matrix.org).

Geo25rey:
> @LittleHuba:matrix.org

Done

Issue #1515, opened 2019-10-04
Requirements Analysis for File Format


LittleHuba:
Here is a public place to discuss about the requirements for the new file format.

Following is a list of requirements I have identified so far. Feel free to add further or discuss the validity of any. Implementation of the requirements is not yet important. See this as a wish list.

- support for partial reads (jit)
- continued autosave
- support for both pages and infinity paper
- support for both packed attachments and external attachments
- clear and simple versioning (including backwards compatibility)
- undo/redo preserved over saves (can be switched on or off as it will increase filesize)
- support for external files (OneDrive, Cloud, URL,...)
- clear OO-programming
- compression of attachments

The file format and all code handling it will be outsourced into an own library. We follow common practice with this step and keep a clear separation between handling, modifying and viewing the file.

The discussion will be closed on 09.11.19.

andreasb242:
>support for partial reads (jit)
Which means probably random access, which is usually not possible if the file is compressed. (Huffmann does not allow it, so e.g. with GZip this is not possible, with zip it is possible, because each file is compressed itself, with tar.gz or 7-zip it's therefore not possible).

The Library should be Plattform independent, so e.g. an Android Port will be possible.

Febbe:
zlib or szip should be capable to allow this. The HDF group also used it to compress their random access dataformat

andreasb242:
https://support.hdfgroup.org/doc_resource/SZIP/

SZip is may an option.
zlib is definitive not seekable:
https://stackoverflow.com/questions/14225751/random-access-to-gzipped-files
https://zlib.net/manual.html#Stream

ZLib is always a Stream (yes, you can skip, but for this you decompress everything from the beginning to this position)

LittleHuba:
> Implementation of the requirements is not yet important

As mentioned above this is absolutely independent from the implementation details. With all the requirements listed, I'll then search for viable candidates to cover all aspects.

LittleHuba:
> undo/redo preserved over saves (can be switched on or off as it will increase filesize)

I'm thinking of loosening this to a certain amount of steps (100-200) as otherwise file sizes might explode over time. And no usecases where you actually need to undo 200 steps are coming to mind so far...

LittleHuba:
Multi-User support might be interesting as well. This would make the code quite complex though. Hence I'm not really a fan of this. Your thoughts on this would be very welcome!

Febbe:
[...]

PR #2729, opened 2021-01-30
[WIP] Add support for spline approximation of strokes

bhennion:
Hi every one,
This PR follows up on #2578. In the current state of things, any new stroke is approximated by a spline and stored as such.
The eraser does not yet fully work (well, apparently it does, but under the hood, it falls back to strokes as a bunch of points), and splines cannot be saved as such yet.

As you write and strokes are created, you will get interesting statistics (e.i. nb of points vs nb of spline segments) in the terminal.

There is yet a lot of things to do:

- [ ] Adapt the eraser tool
- [ ] Add spline to the file format. Add a "Save without splines" option for backward compatibility.
- [ ] Refactor `Spline`, `PartialSplineSegment` and `SplineSegment` for optimization (e.g. `Spline = std::vector<Point>` and the segments are obtained by recasting)
- [ ] Adapt `Stroke::rotate` `Stroke::scale` `Stroke::calcSize` `Stroke::isInSelection` `Stroke::intersects` (with gap)
- [ ] Create classes `StrokeWithPressure` and `StrokeWithoutPressure` to avoid testing at every iteration and reduce the memory footprint of the latter.
- [ ] Fix pressure variations (see [this comment](https://github.com/xournalpp/xournalpp/issues/2578#issuecomment-757336585))
- [ ] Add cusp detection (to get a smaller number of spline segments, see for instance the article referenced [here](https://github.com/xournalpp/xournalpp/issues/2578#issuecomment-749152479))
- [ ] Benchmark all that

You are of course more than welcome to give your opinion on or even tackle any of the above tasks, or to either add other tasks I forgot.


bhennion:
I made progress on the eraser tool. It now works fine when no pressure is involved. It partially fixes #997 (the highligher is displayed as it should while erasing). For strokes with pressure, they are displayed without pressure while erasing. The pressure is back once we stop erasing. See here (besides the poor quality and flickers of the gif)
Eraser1

@rolandlo: your long winded strokes [issue](https://github.com/xournalpp/xournalpp/discussions/2767#discussioncomment-347387) should be fixed too. If you have a minute to try it out. [Note that fixing the pressure issues may reintroduce some lag later. Hopefully not]

rolandlo:
> @rolandlo: your long winded strokes [issue](https://github.com/xournalpp/xournalpp/discussions/2767#discussioncomment-347387) should be fixed too. If you have a minute to try it out. [Note that fixing the pressure issues may reintroduce some lag later. Hopefully not]

The eraser works very nicely with (long winded) strokes that I draw with the version from this PR. If I take the long winded stroke that I drew with the current master there is some strange behavior when erasing: The eraser acts like it were pretty thick and after a short while the whole stroke disappears completely.

bhennion:
> If I take the long winded stroke that I drew with the current master there is some strange behavior when erasing: The eraser acts like it were pretty thick and after a short while the whole stroke disappears completely.

Yes, that's expected at this stage. The new eraser only works (for now) with strokes that are splines, and stroke are not converted to splines on file load. I did not do anything in the save/load direction. This actually raises the interesting question of how to handle backward compatibility.

redweasel:
I would like to raise a concern here. I HATE if the stroke would slightly move every time I release the pen after a stroke! A lot of Software packages do this and i absolutly hate it!

On the other hand I LOVE bezier curves and would like to see more of them in xournalpp. The reduced memory consumption is also a bonus, that I don't wanna miss out on.

Here is what I think you can do: (correct me if I'm wrong)
(Assuming your splines are of degree 3 always, so 2 controlpoints between the real points
1. Approximate spline until a second real point is created
2. Fix the first one in place and only approximate the stroke from there.
3. Draw approximated spline segment instead of the original points

This would help increase performance on usual handwriting. But more importantly, the jitter only happens very close to the tip of the pen. I wonder how the error tolerance will look like with this approach. It might need to be set to a smaller value, depending on the implementation.

The question of backwards compatibility is a very critical one here as well! Since from my understanding the spline tool which is already in, is converting the spline into points and saving it like that, there is no way to save on the memory and have full backwards compatibility. If you fix yourself to only use 3. order bezier curves (2 control points) you could write the points and control points out in the usual point list for each stroke. Then loading it from the new version would understand to draw the bezier curves and loading it from an old version would draw a polyline, which might get close to the real shape, depending on the approximation error value. That's the best I came up with. Sorry if that wasn't helpful.

bhennion:
> I HATE if the stroke would slightly move every time I release the pen after a stroke!

I don't like it either, but I don't know how to avoid it in a satisfactory manner. We kinda had this discussion in #2578. One possible idea (other than yours) is to differ the replacement of the raw stroke by the spline until the next full repaint (zoom in/out, scroll, etc), so that it becomes somewhat invisible to the user. I'm not sure I like this solution, but it's on the table, I think.

> Here is what I think you can do: (correct me if I'm wrong)
> (Assuming your splines are of degree 3 always, so 2 controlpoints between the real points
> 
> 1. Approximate spline until a second real point is created
> 2. Fix the first one in place and only approximate the stroke from there.
> 3. Draw approximated spline segment instead of the original points

I don't really understand what you mean by "a second real point". When is an input `move` event a real point? There seem to be "on the fly" approximation algorithms out there (e.g. a patented one that we mentionned in #2578), but we could not find an open source one. (Note that any on-the-fly approximation will be less efficient than an after-the-fact algorithm).
I could try to implement something along your lines (I think, tell me if this was not what you had in mind):

Apply a Schneider-like least squares method to fit a Bézier curve to the sample (a queue of `move` events). As events are triggered and added to the sample, this curve will change. Whenever the best fitted curve is to far from the data points, store the last best fitted curve and flush all the corresponding events from the sample (i.e. all but the last in).

All in all, I think this would cost significantly more. This could be worth try, especially if we can have an efficient answer to the question: how to compute the distance between the Bézier curve and the piecewise-linear path (reterred to as the path in the following) drawn by the raw events? In Schneider's algorithm, it works as follows.

1. Compute the parameter of each point in the path. The parameter is the cumulated length of all segments between the starting point and the current point, over the total length of the path. You get some `double t` between 0 and 1. Compute the point on the spline corresponding to that parameter `t` and compute the distance between the spline point and the path point.
2. If the maximum `maxDistance` of those distances is lower than your fixed error tolerance `maxError`, you're fine. If `maxError < maxDistance < 4maxError`, then it may be worth reparametrizing. This means adapting the parameter `t` associated to each point of the path (and thus the corresponding point on the spline) by some Newton-Raphson method. Recompute the distance and try again. If `maxDistance` is still bigger than `maxError` after a couple (often 4) iteration of this process, then the Bézier curve is not considered a good fit.

The point is that the second step is crucial for Schneider's algorithm to give a small number of Bézier segments in the end. It comes with a cost though. Doing all that at every `move` event seems to costful.

> The question of backwards compatibility is a very critical one here as well! Since from my understanding the spline tool which is already in, is converting the spline into points and saving it like that, there is no way to save on the memory and have full backwards compatibility. If you fix yourself to only use 3. order bezier curves (2 control points) you could write the points and control points out in the usual point list for each stroke. Then loading it from the new version would understand to draw the bezier curves and loading it from an old version would draw a polyline, which might get close to the real shape, depending on the approximation error value. That's the best I came up with. Sorry if that wasn't helpful.

I don't think this can work. The spline can be fairly far from its control points and it would distord the handwrittings to much. For instance, a circle can be approximated (up to e-4) by a spline with 12 points. Drawing as you suggest would render an ugly octogon. In my mind, the files would not end up being backwards compatible: the splines would be stored as such. An option to "export to pre-1.1.1" format would be available (and of course a tool to convert). This probably means a 2-phased release: first we can read spline files and only save stroke as splines if the user asks for it, second writting as spline becomes the default. I don't really know what else can be done.

redweasel:
@bhennion I like how you propose to handle backwards compatibility (though it will potentially leave legacy code in the future codebase)

I'm not sure about the idea to just hide the change. I feel like that would cause some invalid bug reports...
A similar solution would be to approximate on save. Again not any different really.

I can see you did get my point. Thanks for taking the time to explain some details. I have also looked a little at your code now, but there is a lot. I will not pretend to have understood everything at this late evening (it's impressive). One point you didn't mention in your comment, can you cache the last result and use newton method from there or is it part of the curve fitting and needs to be done often? That way only one newton iteration would be needed as it is done on every move event. You can also introduce a delay between the approximation calls, like 50ms or something like that and batch process very small batches. In the meantime the usual stroke points would be shown. After all I don't know how this would affect the approximation quality (the reduced file size goal) as Schneider's algorithm works on the entire path usually.

Anyway I would love to see an incremental construction instead of the jump at the end. I don't think I can be much of a help right now. Let me finish with this link to a shader of a signed distance function for bezier curves that I found https://www.shadertoy.com/view/4sKyzW (it is using newton iteration as far as I can tell)

bhennion:
Status update: the last commit removed `Stroke::points` and adapted almost all geometric shape tools. The new and shiny eraser also works for piecewise-linear strokes (you can try it out on rectangles/arrows/lines from the dedicated tools) as well as for splines.

In effect, the pressure values are no longer used when drawing and save and load operations now result in crashes (hence the pipeline tests failing). It won't stay that way, obviously. I'll also update the todo-list in the first message above.

redweasel:
https://github.com/xournalpp/xournalpp/pull/2729/commits/f5ba9266636aee0942aaf5db9bec3ac083692176 seems to be an unrelated issue. Can you make a Pull Request for that one. It might fix some render issues.

Also about your error function. One alternative to the method you described, would be to calculate the (approximated) intersection points of the spline with the stroke, and then calculate the sum of the area in the enclosed regions as an error value. I think that might be faster, as you only need to sample the bezier curve once (avoiding too many samples on straight lines if possible), then find intersection points using sweep and prune (quite fast), then calulcate the areas.

bhennion:
> [f5ba926](https://github.com/xournalpp/xournalpp/commit/f5ba9266636aee0942aaf5db9bec3ac083692176) seems to be an unrelated issue. Can you make a Pull Request for that one. It might fix some render issues.

Done: see #2846.

> Also about your error function. One alternative to the method you described, would be to calculate the (approximated) intersection points of the spline with the stroke, and then calculate the sum of the area in the enclosed regions as an error value. I think that might be faster, as you only need to sample the bezier curve once (avoiding too many samples on straight lines if possible), then find intersection points using sweep and prune (quite fast), then calulcate the areas.

This idea crossed my mind too. To compute the intersections, I imagined something involving one Newton-Raphson per segments in the stroke. I don't know much about Sweep and Prune, but it seems to depend on a reliable time parameter (close to the parametrization issue we have with the other method). Am I wrong?

Another option is to sample the Bézier curve and compute the distance between each sampled spline point and the strokes segments. This seems quite efficient.

redweasel:
sweep and prune is just a good/fast scanline algorithm to quickly detect which line segments could intersect (broad phase collision detection). I can see problems with this approach though, when there is self-intersections or when there is geometry which does not have area like a line going back on itself, it could create wrong results and think it matched it perfectly.

Your last Idea is also pretty good, you just need a lot of evenly spaced samples (maybe up to twice the amount of the stroke?). But to compute the distance of the bezier points to the polyline efficiently, you need a fancy algorithm. Just brute force, checking every edge of the polyline is O(n^2). If you want to try this, go for the brute force way first and check if this approach is good in practice. A fancier algorithm for the distance check would be to do a broad phase with a very simple binary BVH tree with spheres or AABBs around the edges (3D spheres/AABBs since you need a 3D solution for pressure right?), You can use that BVH tree to get rid of a lot of checks, so you will get O(n log n) time for BVH construction and approx. O(log n) time per point distance check, giving you a total of O(n log n) for your error function calculation. If you want to go this route and need an explanation of the BVH stuff, I can give you a better one if you ask me. I have implemented such things already in the past. One advantage of the BVH Tree is also, since the stroke points never move, you can keep a lot of the structure if you do incremental construction.

bhennion:
> Your last Idea is also pretty good, you just need a lot of evenly spaced samples (maybe up to twice the amount of the stroke? But you can go one at a time in a spacially efficient way and abort when one is outside the tolerance). But to compute the distance of the bezier points to the polyline efficiently, you need a fancy algorithm. Just brute force, checking every edge of the polyline is O(n^2). If you want to try this, go for the brute force way first and check if this approach is good in practice. A fancier algorithm for the distance check would be to do a broad phase with a very simple binary BVH tree with spheres or AABBs around the edges (3D spheres/AABBs since you need a 3D solution for pressure right?), You can use that BVH tree to get rid of a lot of checks, so you will get O(n log n) time for BVH construction and approx. O(log n) time per point distance check, giving you a total of O(n log n) for your error function calculation. If you want to go this route and need an explanation of the BVH stuff, I can give you a better one if you ask me. I have implemented such things already in the past. One advantage of the BVH Tree is also, since the stroke points never move, you can keep a lot of the structure if you do incremental construction.
> 
> But in that sense, using https://www.shadertoy.com/view/4sKyzW would not be a bad idea to since you can get your distance of the stroke points to the bezier curve in basically constant time, so the check would be O(n). Your current check is also O(n) AFAICT. (these are only valid if there is only one bezier segment, otherwise it would both be O(n*m) with m the number of bezier segments, I guess. Correct if I'm wrong on this one)

Giving it more thougths, it is no longer clear that a distance function like this is enough for this feature (live spline-approximation near the stylus tip). Let me reflect on this a little bit.

Anyway, having an efficient distance function would still be useful elsewere (for detecting clicks on/near the stroke, for implementing a circular eraser, probably other things). More precisely, it would often be about finding the parameter (along the spline) corresponding to the point of spline closest to a given point. This is similar to the shadertoy @redweasel pointed out. IMO, this should be achieved with a good polynomial roots finding library (open source and cross platforms, ofc). To my surprise, I could not find any. Does anyone know of such a library?

redweasel:
Eigen is an open source library for matrix operations. It can do very efficient matrix/vector operations and also solve for eigenvalues. That way you can use the https://en.wikipedia.org/wiki/Companion_matrix to calculate the roots of a polynomial. Otherwise there is also https://eigen.tuxfamily.org/dox/unsupported/group__Polynomials__Module.html to solve polynomials. The library also contains optimizers and other useful things and is very popular as one can see here https://eigen.tuxfamily.org/index.php?title=Main_Page.

bhennion:
Thanks! I had seen the unsupported Polynomials module of Eigen in my research. I didn't look under its hood, but I thought matrix based algorithm were more expensive: without clever optimisations, each iteration is in O(n²), for n the degree. With something like what you found here [https://www.shadertoy.com/view/4sKyzW](https://www.shadertoy.com/view/4sKyzW), the complexity is also O(n^2) (because we need to solve the derivatives to isolate the roots), but we can drop some roots we don't care about (i.e. not between 0 and 1) earlier in the process (once we know the roots of the derivative).

Of course, we will only need small degrees (say n <= 6), so those considerations may not be so relevant. I can't tell.

bhennion:
Hi everyone!
I made some progress here. The eraser is now fully functional, pressure-sensitive strokes are back, Save/Load handlers are fixed (but do not yet save a spline properly, nor do they convert a loader old-school stroke to a spline).

I must say I am quite happy about how smooth the eraser became, with somewhat simpler code!

bhennion:
> I guess we should benchmark those methods...

Alright, I benchmarked two polynomial solvers. The first one is Eigen's (unsupported) `PolynomialSolver` that relies on the QR-algorithm to compute the eigen-values of the companion matrix. The second one is handmade, based on Halley's method.

The stats are staggering... Here for finding the roots in (0,1) of 100,000 degree 6 polynomials
```
Eigen Loop:  99999
Found 44825 roots in 121586747 microseconds.

Handmade Loop:  99999
Found 44890 roots in 2200424 microseconds. Maximal error: 0.000001
```
I also compared the results when feeding them the same polynomial, ofc. They match.

The big difference here is that the matrix (Eigen) method computes every root, and then only keep the real ones between 0 and 1. The handmade one only ever computes real roots between 0 and 1. This makes the number of application of Halley's method drastically drop (to a handful per polynomial).

I'll commit and push the handmade solver once I've run some more test and tried some more possible optimisations.

bhennion:
Hey everyone!
I finely took the time to improve the polynomial solver I cooked up. This lead to significant (~80%) performance improvement. This is now used to compute the distance between a point and a spline (for instance when clicking to select or play the audio associated to a stroke). It all seems to work pretty fine.

bhennion:
Hi everyone!
So, I've rebased onto master and cleaned up a little the commit list to see things more clearly. I should have some time in the next month or so to work on this PR a little bit.

The biggest step yet to make is the adaptation of the file format. As @redweasel mentionned in #3021, it'd be best to minimize the number of file format changes.
I don't know how realistic this is, but ideally, this would be merged at the same time as @LittleHuba's [work on a file format](https://github.com/xournalpp/xournalpp/projects/2) (see also #1515). @LittleHuba, has there been any progress with this? (see also #2233)

Should this PR be put on hold, some parts of it may be retrofitted and merged independently. I am thinking about the erasure rework that improves the current one (see #2767 for a list of erasure related issues).

bhennion:
I just pushed a first attempt at a "live approximator", meaning that the splines are found on the fly (as opposed to after the stroke is finished).

@redweasel it should correspond to the suggestion you made above. Does it feel like you expected ?

If you look at your terminal, you'll see that: it actually gives better splines (i.e. less segments per stroke), which is good, but it costs significantly more (~4x more cpu time than the simple implementation of Schneider's algorithm).

You can also see that there are some flickers and lags (especially when making a long stroke). Most of this is due to the dirtiness of the patch and can most likely be avoided.

Still, there will most likely remain severe lags on lower end devices with such an algorithm. I'll try to improve this, but I'm running low on ideas here. I'll describe the whole algorithm in another post in case someone has an idea.

rolandlo:
Agreed, it's currently too laggy when using the stylus. With the mouse (where less input samples are generated) it feels OK (on a reasonably fast laptop, on Linux). I can try the previous commit on MacOS Catalina, where performance is worse in general, but still acceptable (without spline approximation).

By the way, when the `fill` option is on, the app crashes as soon as releasing the pen.
```
Error: signal 11
[bt]: (0) build/src/xournalpp(+0x22f2eb) [0x559c103ee2eb]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x41040) [0x7f4dbe0b7040]
[bt]: (2) build/src/xournalpp(_ZNK5Point14relativeLineToERKS_d+0x38) [0x559c103b1288]
[bt]: (3) build/src/xournalpp(_ZN6Spline14addLineSegmentERK5Point+0x49) [0x559c103b2e49]
[bt]: (4) build/src/xournalpp(_ZN6Spline16LiveApproximator8finalizeEv+0x195) [0x559c103b7075]
[bt]: (5) build/src/xournalpp(_ZN13StrokeHandler20onButtonReleaseEventERK17PositionInputData+0x25f) [0x559c1031195f]
[bt]: (6) build/src/xournalpp(_ZN11XojPageView20onButtonReleaseEventERK17PositionInputData+0x4c) [0x559c1033fbfc]
[bt]: (7) build/src/xournalpp(_ZN15PenInputHandler9actionEndERK10InputEvent+0xd0) [0x559c10379640]
[bt]: (8) build/src/xournalpp(_ZN17MouseInputHandler10handleImplERK10InputEvent+0x8e) [0x559c103786ce]
[bt]: (9) build/src/xournalpp(_ZN12InputContext6handleEP9_GdkEvent+0x4ac) [0x559c1037781c]
[bt]: (10) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3e94a8) [0x7f4dbf2584a8]
```

rolandlo:
I have tested your PR without live approximation (commit 9377348648d96c3600eb2a8fcad86d5f304c62ce) on MacOS Catalina (on a 7 year old MacBook Pro). To my big surprise while drawing (before releasing the pen and triggering the spline approximation) the experience was much more fluid than on current master. Have you found a way to avoid full redraws on MacOS while drawing?

The difference can be measured by the number of input points that are collected while drawing a screen-big circle in normal speed (about 1 sec) with my Wacom Intuos Pro M tablet. 

- With your PR I get around 130 input points (and the shape doesn't look edgy), 
- On current master I only get around 30 input points (and the shape looks roughly like a polygon with about 15 edges). 

On Linux (on a faster laptop)I don't see this difference (probably because there are no full redraws). I get close to 300 input points.

I will test now on BigSur, where the lag on current master is even bigger (about 12 input points for a circle).

bhennion:
> Have you found a way to avoid full redraws on MacOS while drawing?

If so, that's an accident (I don't have any Mac). The lines most likely to have changed something like that are in `StrokeHandler.cpp`. I'll make a review of the file highlighting what I think you could try out directly on master.

src/control/tools/StrokeHandler.cpp
```diff
@@ -132,27 +124,21 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
         cairo_fill(crMask);
 
         view.drawStroke(crMask, stroke, 0, 1, true, true);
+        this->redrawable->repaintRect(stroke->getX() - width, stroke->getY() - width,
+                                      stroke->getElementWidth() + 2 * width, stroke->getElementHeight() + 2 * width);
```
bhennion, line 129: Added this. I don't remember why...

src/control/tools/StrokeHandler.cpp
```diff
@@ -132,27 +124,21 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
...
-            lastSegment.setWidth(stroke->getWidth());
-
-            cairo_set_operator(crMask, CAIRO_OPERATOR_OVER);
-            cairo_set_source_rgba(crMask, 1, 1, 1, 1);
-
-            view.drawStroke(crMask, &lastSegment, 0, 1, false);
-        }
+        view.drawStroke(crMask, &lastSegment, 0, 1, false);
+        this->redrawable->repaintRect(lastSegment.getX(), lastSegment.getY(), lastSegment.getElementWidth(),
+                                      lastSegment.getElementHeight());
```
bhennion: Same here

src/control/tools/StrokeHandler.cpp
```diff
@@ -334,7 +308,7 @@ void StrokeHandler::onButtonPressEvent(const PositionInputData& pos) {
     double width = page->getWidth() * zoom * dpiScaleFactor;
     double height = page->getHeight() * zoom * dpiScaleFactor;
 
-    surfMask = cairo_image_surface_create(CAIRO_FORMAT_A8, width, height);
+    surfMask = cairo_image_surface_create(CAIRO_FORMAT_A8, (int)std::ceil(width), (int)std::ceil(height));
```
bhennion, line 342: Made this change too. It could be relevant if Cairo reacts badly when try to paint outside the bounds of the surface. I don't know.

rolandlo:
The changes you highlighted applied to current master didn't have a noticeable impact on performance on MacOS Catalina. There must be something else that makes your code more performant.

bhennion:
I improved quite a bit the live approximator. It is now fluid on my laptop (but it's quite high end). I quite like the feeling of it now!

In terms of statistic, it gives an average of 1 spline segment (~3 points) for 50 points with this laptop's pen. It means something like a 94% in data size in the end!

I have another lead to yet optimize the live approximator, but it'd be based on how many points are merged into on spline segment on average. This of course depends on the device (from 5 on my touchpad to 50 on my pen), so I need to think about how to infer this. If you have an idea, shoot!

rolandlo:
`SplineLiveApproximator2.h` is missing. It doesn't compile.

bhennion:
> `SplineLiveApproximator2.h` is missing. It doesn't compile.

Yes of course I forgot to add the new files... Repushed an amended commit. Sorry about that!

rolandlo:
It's indeed amazingly fluent on Linux! Well done! :muscle:  Let me try on MacOS.

By the way, the eraser doesn't work on horizontal and vertical strokes, it seems.

bhennion:
> By the way, the eraser doesn't work on horizontal and vertical strokes, it seems.

Could you describe the bug more precisely? I can't reproduce. For instance when I draw a rectangle, I can erase anything I want.

There is another bug though: when doing a single-click stroke. This case needs to be handles separatly.

rolandlo:
When doing a single click, a horizontal stroke is generated (at least that's how it appears on the screen) and these strokes can't be deleted, neither with the standard eraser mode nor with the delete stroke eraser mode. 

On MacOS the performance was also very fluid. I got a (seemingly random) crash (segfault) in `Document::drawElement` once when drawing for a long time (although not particularly long strokes). 

One thing that could possibly be improved is the blinking when the stroke is finished. While drawing that doesn't happen, just at the end when releasing the pen. I can observe it on Linux as well.

bhennion:
Ok, so the issue seems to be the mishandling of the single clicks. I can imagine it triggering a SegFault, since nothing is done to handle this very peculiar case for now.

> One thing that could possibly be improved is the blinking when the stroke is finished.
That's quite easy to solve in itself. This blinking comes from the rerendering I added because of the a posteriori spline approximation. This can be removed when using the live approximator.

It raises a more complex question: how to handle the various modes (no spline, approximation after the facts or live approximation). I add todos in the first post :-)

bhennion:
Remastered the spline approximators (both a after-the-fact and live approximators) to factor code chunks. Performances have also been improved (quite a lot).
The flicker issue has been solve, and dashed and filled strokes now work (except when they correspond to single clicks).

The live approximation is enabled (hardcoded) with this commit, but everything is in place to add a setting.

It'd be helpful if people could try this commit out on their respective machines, and tell me if they experience any lag with the live approximation.

rolandlo:
I have tested 8951119c6927770859a026d5861a1998372a3220 on my Lenovo Yoga convertible on Ubuntu 21.04 (on X11 and Wayland) and MS Windows 10 and on my MacBook Pro on MacOS BigSur. On my convertible I tried it with the mouse, with the integrated stylus and with my Wacom Intuos Pro M tablet. On the MacBook I only had the mouse and the Wacom tablet available.

The blinking is gone and the issue with single clicks is also resolved, which is great! As I noticed the live approximator draws the anchor points it uses as little circles. They disappear after a full redraw (some seconds after the stroke has finished). I guess this was done on purpose (for debugging)

With normal zoom setting (A4 paper scaled to fit width) the live approximator worked very well on all devices. There is lag with higher zoom. It depends on the device used for drawing. I made sure the dpi setting was set correctly in the preferences (on my Lenovo convertible I have 209 dpi, on my MacBook Pro I have 147 dpi)

On Linux (both X11 and Wayland) and Windows the experience was good up to zoom levels of:
- 600% for mouse input
- 400% for integrated stylus input
- 300% for the Wacom tablet stylus

On higher zoom levels the strokes only appear with some lag, which in extreme cases means that the strokes only appear some seconds after the stroke has been finished. The number of spline segments didn't seem to depend on what zoom level I drew.
With fill option it was more extreme, I had to reduce the zoom level by about 100% to not experience lag. The highlighter didn't (without fill option) didn't seem to make a difference.

On Mac OS Big Sur (without toolbar) there was lag on lower zoom levels (as expected)
- 180% for mouse input
- 140% for the Wacom tablet stylus

I have encountered an issue with the eraser on Mac OS Big Sur. It couldn't erase some single points. I get lots of messages "Improbable case 2" and the erasing shows some lag. The lag while erasing doesn't happen on Linux and Windows and the messages "Improbable case 2" only appear very rarely on these operating systems. 

On Mac OS BigSur I have encountered a (seemingly random) crash (like reported in a previous version) after drawing for a while on a slightly too high zoom level. Here is the crash log:

```
Date: Sun Aug  1 10:01:06 2021
Error: signal 11
[bt]: (0) 0   xournalpp                           0x000000010e279a9d _ZL12crashHandleri + 125
[bt]: (1) 1   libsystem_platform.dylib            0x00007fff2062dd7d _sigtramp + 29
[bt]: (2) 2   ???                                 0x3fb999999999999a 0x0 + 4591870180066957722
[bt]: (3) 3   xournalpp                           0x000000010e26381b _ZNK12DocumentView11drawElementEP6_cairoP7Element + 43
[bt]: (4) 4   xournalpp                           0x000000010e263b6a _ZN12DocumentView9drawLayerEP6_cairoP5Layer + 250
[bt]: (5) 5   xournalpp                           0x000000010e264508 _ZN12DocumentView8drawPageENSt3__110shared_ptrI7XojPageEEP6_cairobbbb + 392
[bt]: (6) 6   xournalpp                           0x000000010dfba054 _ZN10PreviewJob8drawPageEv + 756
[bt]: (7) 7   xournalpp                           0x000000010dfba438 _ZN10PreviewJob3runEv + 72
[bt]: (8) 8   xournalpp                           0x000000010dfb8b69 _ZN3Job7executeEv + 25
[bt]: (9) 9   xournalpp                           0x000000010dfc2d83 _ZN9Scheduler17jobThreadCallbackEPS_ + 595
[bt]: (10) 10  libglib-2.0.0.dylib                 0x000000010eec4d02 g_thread_proxy + 66
[bt]: (11) 11  libsystem_pthread.dylib             0x00007fff205e9950 _pthread_start + 224
[bt]: (12) 12  libsystem_pthread.dylib             0x00007fff205e547b thread_start + 15
```

rolandlo:
One more thing I have noticed: After selecting a stroke switching the pen size does not work.

bhennion:
Thanks a lot for this thorough testing!

> As I noticed the live approximator draws the anchor points it uses as little circles. They disappear after a full redraw (some seconds after the stroke has finished). I guess this was done on purpose (for debugging)

Yes, this is on purpose and will be removed.

I'm not sure why the zoom has any effect on the lag on Linux or Windows. I'll try something out. On MacOs, I can imagine the "live segment" (being replaced at every new event) takes a bigger area of the screen thus triggering a repaint too large for Quartz.  If so, I don't really know what I can do besides resolving to an approximation after the fact. I'll think about it.

> I have encountered an issue with the eraser on Mac OS Big Sur. It couldn't erase some single points. I get lots of messages "Improbable case 2" and the erasing shows some lag. The lag while erasing doesn't happen on Linux and Windows and the messages "Improbable case 2" only appear very rarely on these operating systems.

Hum... This is weird bug. I managed to reproduce it (but very rarely). It will be fairly difficult to pinpoint the cause, so it'll take some time.

> On Mac OS BigSur I have encountered a (seemingly random) crash (like reported in a previous version) after drawing for a while on a slightly too high zoom level.

This is also gonna be difficult if I can't reproduce. I'll add some asserts here and there and aks you to test again in a while.

> After selecting a stroke switching the pen size does not work.

This should be easy. I added this (and all the above) to the todo-list

bhennion:
I think I corrected all the eraser's `Improbable case 2` bugs I could trigger on Linux. One was due to numerical instability in the polynomial solver, the other was due to vertical/horizontal strokes that had an empty bounding box.
I still don't know why they'd appear more often on macOS. @rolandlo, could you try once more? Should you still get some `Improbable case 2`'s, you should have a pretty verbose output in the terminal that could help me figure out what's going on.

I added some `assert`'s in the `Spline` class to try and catch the `SegFault` @rolandlo reported.

I also added an attempt at reducing the lag on high zoom ratio. I can't test it right now as I don't have my tablet with me. I'll make some proper testing in a week or so. I'm not really satisfied with this solution (if it is one) anyway: it simply drops every event that happens (zoom agnostically) to close to the last one.

I'm thinking about another and more satisfying solution: instead of approximating after every input event, I should approximate in `StrokeHandler::draw` (which, at least on my laptop, is called ~60 times a second). Depending on the input device, this should significantly reduce the number of calls to `SingleSegmentFitter::findBestCubicSegment()`.

bhennion:
Another issue related to approximation and zoom: for now, the constants (e.g. `MAX_ERROR`) are expressed in page coordinates. In particular, the same input sequence at different zoom levels gives different splines. This probably causes the differences in performance that various zoom levels show.

I think the behaviour should be zoom-agnostic. However, the original behaviour (on `master`) is not zoom-agnostic: the constant `InputHandler::PIXEL_MOTION_THRESHOLD` that determines whether or not an event should be ignored is expressed in page coordinates (so the shortest possible segment looks bigger in higher zoom levels). Does anyone know if this was the result of an actual choice, or just happened to be?

rolandlo:
> I think I corrected all the eraser's `Improbable case 2` bugs I could trigger on Linux. One was due to numerical instability in the polynomial solver, the other was due to vertical/horizontal strokes that had an empty bounding box.
> I still don't know why they'd appear more often on macOS. @rolandlo, could you try once more? Should you still get some `Improbable case 2`'s, you should have a pretty verbose output in the terminal that could help me figure out what's going on.

I don't get these messages any more with the latest commit.

> 
> I added some `assert`'s in the `Spline` class to try and catch the `SegFault` @rolandlo reported.
> 
I haven't yet got that SegFault any more.

> I also added an attempt at reducing the lag on high zoom ratio. I can't test it right now as I don't have my tablet with me. I'll make some proper testing in a week or so. I'm not really satisfied with this solution (if it is one) anyway: it simply drops every event that happens (zoom agnostically) to close to the last one.

It looks to me that there is a modest improvement on MacOS, but the issue is still there.

> 
> I'm thinking about another and more satisfying solution: instead of approximating after every input event, I should approximate in `StrokeHandler::draw` (which, at least on my laptop, is called ~60 times a second). Depending on the input device, this should significantly reduce the number of calls to `SingleSegmentFitter::findBestCubicSegment()`.

bhennion:
Hey!

I started investigating the lag when zoomed in @rolandlo found. I managed to reproduce something, hopefully it is the same bug. First, as you can see on the video below, it seems to be a purely graphical bug: the spline approximation is still running in the console and showing the `failed` messages (those messages are not error messages. They are nominal behaviour).

https://user-images.githubusercontent.com/17818041/136507386-0cc08710-c796-484f-85e8-2176c92b2def.mp4

@rolandlo Could you confirm that the lag you encounter behaves similarly? The last commit should print the same messages as in the video.

Now, from there I do not know where to go. I don't think any of the changes I made in the `StrokeView` class could trigger this zoom related bug. It feels like the issue comes from the view classes.

bhennion:
@rolandlo do you have strokes recognition on? If so, the segfault is expected.

As for the messages, and just to be clear, you do get them even when the stroke view is falling behind (like in the video)? So that it shows the approximation algorithm is not the source of the lag and it's rather a display bug?
(I'm not sure how to formulate my question clearly, I'm sorry. Please do tell if I need to try and reformulate)

bhennion:
I think I managed to fix the lag. I still don't know why it worstened with higher zoom levels, but it seems to be gone now. @rolandlo, could you please confirm it works for you too now (even on BigSur)? Thanks a lot!

rolandlo:
@bhennion Indeed the live approximator (for ordinary pen with plain line style and no filling) is very smooth on both Linux and Windows on all zoom levels now. I get around 40-60 frames per second. Great job!
On MacOS it's still laggy and worse on BigSur (where the live approximator is smooth up to around 150%) than on Catalina (where the live approximator is smooth up to around 200%). The frame rate is quite a bit smaller (around 15-20 frames per second).

bhennion:
@rolandlo Thanks a lot for the testing!! I'll see what I can do on MacOs, but I'm not very optimistic.

bhennion:
> I'm currently quite confused by inconsistent observations I make. Somehow the latest commit works pretty OK when run first time after being built (there is lag on higher zoom levels, but after writing ends and the rendering catches up I still get precisely what I wanted to write/draw), but poorly (on higher zoom levels) from the 2nd time on. I first tested on BigSur and now on Catalina, where I seem to have found a way to reproduce the problem consistently. Probably it's something that changes in the settings. I will try to find out.

This is indeed quite confusing.

In the meantime, I applied to this PR the changes of #3480 that could reduce lag on high zoom levels. Hopefully this will help.

personalizedrefrigerator:
Copy-pasting a stroke that is not a spline **may** be throwing a `std::exception` with this PR.

The build of Xournal++ I use on my tablet incorporates this PR (but I have spline approximation turned off in settings). When I try to copy-paste a stroke, Xournal++ seems to crash.

I plan to try re-building from this PR (without any other PRs merged in) this weekend to see if it actually is this PR causing the issue.

bhennion:
I rebased on current master here.
Apparently, the test `testReadStroke` (from `test/unit_tests/util/ObjectIOStreamTest.cpp`) is failing, but only on MacOs.
I don't understand how that could be. Is there a way to get more verbose output from the pipeline test upon failure?

bhennion:
> Copy-pasting a stroke that is not a spline **may** be throwing a `std::exception` with this PR.
> 
> The build of Xournal++ I use on my tablet incorporates this PR (but I have spline approximation turned off in settings). When I try to copy-paste a stroke, Xournal++ seems to crash.
> 
> I plan to try re-building from this PR (without any other PRs merged in) this weekend to see if it actually is this PR causing the issue.

First, thanks a lot for checking this PR out!

I can't reproduce the bug you mention, though. How are you copy-pasting? `Ctrl + C` and `Ctrl + V`? Cloning with the right button?

bhennion:
This was rebased on #3532. I added a todo item to remind us of this.

bhennion:
Rebased on master after the merge of #3532.

I also made some code readability improvements, and added some benchmarks (see f5c4c09 and 282d7fc).

There seem to be an issue with the CI build on Linux:
```
/home/vsts/work/1/s/src/core/control/tools/splineapproximation/SplineApproximatorSingleSegmentFitter.cpp:74:21: error: ‘transform_reduce’ is not a member of ‘std’
```
despite the `#include <numeric>`. Does anyone know why?

moelski:
Hi !
This sounds like a really great feature for Xournal. If I get it right, it would smooth my hand drawings, correct? 
Any idea when it will be in a first downloadable version?

bhennion:
> Hi ! This sounds like a really great feature for Xournal. If I get it right, it would smooth my hand drawings, correct? Any idea when it will be in a first downloadable version?

Well, this is gonna take a while (and a long one). In the meantime, there are Stroke stabilization algorithms in the current version. You can enable them in the Input system configuration panel. It's kind of difficult to know which parameters to put in though. There were some discussion about that happening in #3721.

PR #2730, opened 2021-01-30
Fix confusing behavior of "Apply [page type] to current/all pages"

### LINKED ISSUE #951: Paper background "apply to all pages" strange behavior?
I supposed that "apply to all pages" should be a simple switch --- but it's acting (and looking) quite strange here (recently compiled, says it's 1.0.9): 

recorded

(do not mind the flashing lines, it's a `byzanz` (the gif recorder) artifact.

Technius:
This PR fixes the confusing behavior of the `Apply to current/all pages` buttons. Fixes #951.

* Fixes a bug where the `Journal > Paper background` apply page type button would use the page type from the new page type dropdown. The fix is a bit hacky but it's simple and straightforward.
* Fixes a bug where the selected page type in `Journal > Paper background` would not update after using `Apply to all pages` from the new page type dropdown.
* Removes the confusing check mark next to the Apply to current / all pages button.

Merging in 48 hours if there are no objections.

### DIFF
--- a/src/gui/dialog/PageTemplateDialog.cpp
+++ b/src/gui/dialog/PageTemplateDialog.cpp
@@ -1,6 +1,7 @@
 #include "PageTemplateDialog.h"
 
 #include <fstream>
+#include <memory>
 #include <sstream>
 
 #include "control/pagetype/PageTypeHandler.h"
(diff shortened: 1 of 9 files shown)


src/gui/dialog/PageTemplateDialog.cpp
```diff
@@ -158,8 +156,6 @@ void PageTemplateDialog::showPageSizeDialog() {
 
         updatePageSize();
     }
-
-    delete dlg;
```
personalizedrefrigerator, line 162: It looks like dlg is still being allocated with 'new' on line 147... Who is responsible for freeing dlg?

personalizedrefrigerator (commented):
Thank you!

src/gui/dialog/PageTemplateDialog.cpp
Technius: Nice catch, thanks.

Technius:
Split up the bugfix commit for more granularity and did a little bit more cleanup in `PageTemplateDialog`.

PR #2738, opened 2021-02-01
Use librsvg directly in thumbnailer to add icon to previews

### LINKED ISSUE #1612: Thumbnails do not show up on Gnome anymore
**Affects versions :**
 - OS: Linux
 - Gnome (X11)
 - Version of Xournal++ : 1.0.14 and 1.0.16

**Description**
Since 1.0.14, Thumbnails for .xopp files do not show up in the file manager (Nautilus) anymore

**To Reproduce**
1. Open the file browser and look for a .xopp file.

**Expected behavior**
The .xopp file has a thumbnail that shows the contents of the first page of the file.

**Actual behavior**
It uses the generic icon of Xournalpp instead of thumbnails.

Technius:
Fixes #1612.

Old behavior: crash during gtk initialization.

New behavior: emit warning message and render the thumbnail without the icon if the icon cannot be found.

The icon search code is not very comprehensive, but this is just a quick fix for the bug.


Technius:

Issue #2703, opened 2021-01-25
Pinch-to-zoom results in trembling effect in large PDFs


coldspark29:
**Affects versions :**
 - OS: Manjaro
 - (Linux only) Desktop environment: [e.g. Wayland, Gnome-Wayland, XWayland, X11]
 - Which version of libgtk do you use
     3.24.24
 - Version of Xournal++
     01107eb
 - Installation method: AUR

**Describe the bug**
This occurs since the scrolling bug while zooming was fixed. It now just scrolls up and down rapidly which results in a trembling affects.

**To Reproduce**
1. Open large PDF > 50 pages
2. Pinch-to-zoom
3. See the trembling effect

Technius:
Are you sure you're using the right version? ` 1.0.15.r49.gf2535dd5-1` is pretty old.

coldspark29:
Are you referring to the 1.0.15? I guess the version shown in the AUR has not been updated in a while. It actually builds from source. I just did it again and have 01107eb which is the latest commit.

Technius:
Do you know which PR/commit `scrolling bug while zooming was fixed` came from?

coldspark29:
Found it https://github.com/xournalpp/xournalpp/issues/1843. I already forgot about it. @Febbe marked it as fixed and then I complained. Before it was just scrolling up endlessly. I think he did something to make up for the scrolling. It seems like it immediatley scrolls down again. More like a workaround than a fix.

personalizedrefrigerator:
#2435 touched code related to this. See https://github.com/xournalpp/xournalpp/pull/2435#issuecomment-734790455 for the explanation.

This trembling happened prior to the PR, though, when the margin between pages was set to zero.

This may or may not be relevant, but I hope it is useful!

coldspark29:
It is fixed! After all these years :D

PR #2743, opened 2021-02-03
Fix Zoom Jitter

redweasel:
fixes https://github.com/xournalpp/xournalpp/issues/2703

The fix was to make layout calculate the page height/width with double precision, because otherwise on page N a random error of ~N/2 would have been accumulated.

~~I also refactored some of the space conversion code to be less error prone to double-precision-errors. Also with the old code I noticed, that when zooming with the mouse on the last page, it would randomly jump to prior pages. That is fixed with the refactor. I did test a lot of things, but I am sure I left something untested, so testing is highly encouraged!~~

~~If requested I can split this PR into the refactor/jump fix and the jitter fix as they are mostly separate in code.~~

**Edit:** This PR got split.

### DIFF
--- a/src/gui/Layout.cpp
+++ b/src/gui/Layout.cpp
@@ -145,8 +145,8 @@ void Layout::recalculate_int() const {
         auto const& c = raster_p.first;
         auto const& r = raster_p.second;
         XojPageView* v = view->viewPages[pageIdx];
-        pc.widthCols[c] = std::max(pc.widthCols[c], as_unsigned_strict(v->getDisplayWidth()));
-        pc.heightRows[r] = std::max(pc.heightRows[r], as_unsigned_strict(v->getDisplayHeight()));
+        pc.widthCols[c] = std::max(pc.widthCols[c], v->getDisplayWidthDouble());
+        pc.heightRows[r] = std::max(pc.heightRows[r], v->getDisplayHeightDouble());
     }
 
     // add space around the entire page area to accommodate older Wacom tablets with limited sense area.
@@ -158,8 +158,8 @@ void Layout::recalculate_int() const {
     pc.minWidth = as_unsigned(2 * hPadding + as_signed_strict((pc.widthCols.size() - 1) * XOURNAL_PADDING_BETWEEN));
     pc.minHeight = as_unsigned(2 * vPadding + as_signed_strict((pc.heightRows.size() - 1) * XOURNAL_PADDING_BETWEEN));
 
-    pc.minWidth = std::accumulate(begin(pc.widthCols), end(pc.widthCols), pc.minWidth);
-    pc.minHeight = std::accumulate(begin(pc.heightRows), end(pc.heightRows), pc.minHeight);
+    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), pc.minWidth));
+    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), pc.minHeight));
     pc.valid = true;
 }
 
@@ -197,8 +197,8 @@ void Layout::layoutPages(int width, int height) {
     auto const centeringYBorder = (height - as_signed(pc.minHeight)) / 2;
 
     using SBig = decltype(as_signed(h_padding * centeringXBorder));
-    auto const borderX = std::max<SBig>(h_padding, centeringXBorder);
-    auto const borderY = std::max<SBig>(v_padding, centeringYBorder);
+    auto const borderX = static_cast<double>(std::max<SBig>(h_padding, centeringXBorder));
+    auto const borderY = static_cast<double>(std::max<SBig>(v_padding, centeringYBorder));
 
     // initialize here and x again in loop below.
     auto x = borderX;
@@ -217,11 +217,11 @@ void Layout::layoutPages(int width, int height) {
                 XojPageView* v = this->view->viewPages[*optionalPage];
                 v->setMappedRowCol(strict_cast<int>(r),
                                    strict_cast<int>(c));  // store row and column for e.g. proper arrow key navigation
-                auto vDisplayWidth = v->getDisplayWidth();
+                auto vDisplayWidth = v->getDisplayWidthDouble();
                 {
-                    int64_t paddingLeft = 0;
-                    int64_t paddingRight = 0;
-                    auto columnPadding = as_signed(this->pc.widthCols[c]) - vDisplayWidth;
+                    auto paddingLeft = 0.0;
+                    auto paddingRight = 0.0;
+                    auto columnPadding = this->pc.widthCols[c] - vDisplayWidth;
 
                     if (isPairedPages && len > 1) {
                         // pair pages mode
@@ -233,15 +233,15 @@ void Layout::layoutPages(int width, int height) {
                             paddingLeft = XOURNAL_ROOM_FOR_SHADOW;
                             paddingRight = XOURNAL_PADDING_BETWEEN - XOURNAL_ROOM_FOR_SHADOW + columnPadding;
                         }
-                    } else {                                                            // not paired page mode - center
-                        paddingLeft = XOURNAL_PADDING_BETWEEN / 2 + columnPadding / 2;  // center justify
-                        paddingRight = XOURNAL_PADDING_BETWEEN - paddingLeft + columnPadding / 2;
+                    } else {  // not paired page mode - center
+                        paddingLeft = XOURNAL_PADDING_BETWEEN / 2.0 + columnPadding / 2.0;  // center justify
+                        paddingRight = XOURNAL_PADDING_BETWEEN - paddingLeft + columnPadding / 2.0;
                     }
 
                     x += paddingLeft;
 
-                    v->setX(strict_cast<int>(x));  // set the page position
-                    v->setY(strict_cast<int>(y));
+                    v->setX(floor_cast<int>(x));  // set the page position
+                    v->setY(floor_cast<int>(y));
 
                     x += vDisplayWidth + paddingRight;
                 }
@@ -278,7 +278,7 @@ auto Layout::getPaddingAbovePage(size_t pageIndex) const -> int {
 
     // User-configured padding above all pages.
     auto const paddingAbove =
-            sumIf(XOURNAL_PADDING, settings->getAddHorizontalSpaceAmount(), settings->getAddVerticalSpace());
+            sumIf(XOURNAL_PADDING, settings->getAddVerticalSpaceAmount(), settings->getAddVerticalSpace());
 
     // (x, y) coordinate pair gives grid indicies. This handles paired pages
     // and different page layouts for us.
@@ -291,25 +291,24 @@ auto Layout::getPaddingLeftOfPage(size_t pageIndex) const -> int {
     bool isPairedPages = this->mapper.isPairedPages();
     const Settings* settings = this->view->getControl()->getSettings();
 
-    auto const paddingBefore =
-            sumIf(XOURNAL_PADDING, settings->getAddVerticalSpaceAmount(), settings->getAddHorizontalSpace());
+    auto paddingBefore =
+            sumIf(XOURNAL_PADDING, settings->getAddHorizontalSpaceAmount(), settings->getAddHorizontalSpace());
 
     auto const pageXLocation = as_signed(this->mapper.at(pageIndex).first);
 
     // No page pairing or we haven't rendered enough pages in the row for
     // page pairing to have an effect,
-    if (!isPairedPages || pageXLocation == 0) {
-        return strict_cast<int>(pageXLocation * XOURNAL_PADDING_BETWEEN + as_signed(paddingBefore));
+    if (!isPairedPages) {
+        return strict_cast<int>(pageXLocation * XOURNAL_PADDING_BETWEEN + XOURNAL_PADDING_BETWEEN / 2 +
+                                as_signed(paddingBefore));
     } else {
-        // We have a greater separation between pairs of pages. Handle this here,
-        //  Note that pageXLocation - 1 >= 0 because we take the if branch above when
-        // pageXLocation == 0.
-        auto paddingBetweenPairs = as_signed(pageXLocation - 1) / 2 * XOURNAL_PADDING_BETWEEN;
-
-        // The two pages within each pair have a smaller separation, XOURNAL_ROOM_FOR_SHADOW.
-        auto shadowRoomInsidePairs = pageXLocation / 2 * XOURNAL_ROOM_FOR_SHADOW;
-
-        return strict_cast<int>(paddingBetweenPairs + shadowRoomInsidePairs + paddingBefore);
+        auto columnPadding =
+                XOURNAL_PADDING_BETWEEN + strict_cast<int>(pageXLocation / 2) * (XOURNAL_PADDING_BETWEEN * 2);
+        if (pageXLocation % 2 == 0) {
+            return strict_cast<int>(columnPadding - XOURNAL_ROOM_FOR_SHADOW + paddingBefore);
+        } else {
+            return strict_cast<int>(columnPadding + XOURNAL_ROOM_FOR_SHADOW + paddingBefore);
+        }
     }
 }
 
--- a/src/gui/Layout.cpp
+++ b/src/gui/Layout.cpp
@@ -158,8 +158,8 @@ void Layout::recalculate_int() const {
     pc.minWidth = as_unsigned(2 * hPadding + as_signed_strict((pc.widthCols.size() - 1) * XOURNAL_PADDING_BETWEEN));
     pc.minHeight = as_unsigned(2 * vPadding + as_signed_strict((pc.heightRows.size() - 1) * XOURNAL_PADDING_BETWEEN));
 
-    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), pc.minWidth));
-    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), pc.minHeight));
+    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), double(pc.minWidth)));
+    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), double(pc.minHeight)));
     pc.valid = true;
 }
 
--- a/src/gui/PageView.h
+++ b/src/gui/PageView.h
@@ -119,11 +119,13 @@ public:
      * on the display taking into account the current zoom
      */
     int getDisplayWidth() const;
+    double getDisplayWidthDouble() const;
     /**
      * Returns the height of this XojPageView as displayed
      * on the display taking into account the current zoom
      */
     int getDisplayHeight() const;
+    double getDisplayHeightDouble() const;
 
     /**
      * Returns the x coordinate of this XojPageView with
--- a/src/gui/Layout.h
+++ b/src/gui/Layout.h
@@ -42,8 +42,8 @@ public:
         // The width and height of all our pages
         size_t minWidth = 0;
         size_t minHeight = 0;
-        std::vector<unsigned> widthCols;
-        std::vector<unsigned> heightRows;
+        std::vector<double> widthCols;
+        std::vector<double> heightRows;
         bool valid = false;
     };
 
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -867,6 +867,12 @@ auto XojPageView::getDisplayHeight() const -> int {
     return std::lround(this->page->getHeight() * this->xournal->getZoom());
 }
 
+auto XojPageView::getDisplayWidthDouble() const -> double { return this->page->getWidth() * this->xournal->getZoom(); }
+
+auto XojPageView::getDisplayHeightDouble() const -> double {
+    return this->page->getHeight() * this->xournal->getZoom();
+}
+
 auto XojPageView::getSelectedTex() -> TexImage* {
     EditSelection* theSelection = this->xournal->getSelection();
     if (!theSelection) {
--- a/src/gui/XournalView.cpp
+++ b/src/gui/XournalView.cpp
@@ -486,7 +486,6 @@ void XournalView::zoomChanged() {
     } else {
         auto pos = zoom->getScrollPositionAfterZoom();
         if (pos.x != -1 && pos.y != -1) {
-            // Todo: This could be one source of all evil:
             Layout* layout = gtk_xournal_get_layout(this->widget);
             layout->scrollAbs(pos.x, pos.y);
         }


src/control/zoom/ZoomControl.cpp
```diff
@@ -184,10 +195,9 @@ auto ZoomControl::getScrollPositionAfterZoom() const -> utl::Point<double> {
         return {-1, -1};
     }
 
-    return (this->scrollPosition * this->zoom) - this->zoomWidgetPos + this->unscaledPixels;
+    return this->scrollPosition * this->zoom - this->zoomWidgetPos + unscaledPixels;
```
personalizedrefrigerator, line 198: For consistency, it might be nice to change `unscaledPixels` to `this->unscaledPixels`...

src/gui/PageView.h
```diff
@@ -124,6 +124,16 @@ class XojPageView: public Redrawable, public PageListener {
...
+    /**
+     * Returns the width of this XojPageView as displayed
+     * on the display taking into account the current zoom
+     */
+    double getDisplayWidthDouble() const;
+    /**
+     * Returns the height of this XojPageView as displayed
+     * on the display taking into account the current zoom
+     */
+    double getDisplayHeightDouble() const;
```
personalizedrefrigerator, line 136: Since these have the same documentation strings as the methods above them, would it be reasonable to use the same docstring for both? E.g.

```c++
    /**
     * Returns the width of this XojPageView as displayed
     * on the display taking into account the current zoom
     */
    int getDisplayWidth() const;
    double getDisplayWidthDouble() const;
    /**
     * Returns the height of this XojPageView as displayed
     * on the display taking into account the current zoom
     */
    int getDisplayHeight() const;
    double getDisplayHeightDouble() const;
```

David96:
Thanks a lot, from a quick testrun, zooming appears to be a lot smoother. I'll continue to use this PR and report back in case I find any issues.
Out of curiosity: why did you replace the getVisibleRect calls by gdk_window_get_root_coords? I didn't have any issues with wrongly centered zooming so I assumed the coordinates were correct.

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -137,13 +127,18 @@ void TouchInputHandler::zoomStart() {
         zoomControl->setZoomFitMode(false);
     }
 
+    auto center = (this->priLastAbs + this->secLastAbs) / 2.0;
+
     auto* mainWindow = inputContext->getView()->getControl()->getWindow();
+    int rx, ry;
+    gdk_window_get_root_coords(*mainWindow, 0, 0, &rx, &ry);
+    center.x -= rx;
+    center.y -= ry;
```
personalizedrefrigerator, line 136: ~~This part might warrant a comment. (I'm not very familiar with GTK, though, so it could just be that :)).~~

Consider moving this code into a method in `MainWindow.h/MainWindow.cpp`. I think this will make it easier to understand what's going on here & limit coupling between the input system and GTK.

src/control/zoom/ZoomControl.cpp
```diff
@@ -110,22 +112,23 @@ void ZoomControl::zoomScroll(ZoomDirection zoomIn, utl::Point<double> zoomCenter
...
-    double newZoom = this->zoom + (zoomIn ? this->zoomStepScroll : -this->zoomStepScroll);
+    double newZoomStep = zoomIn ? this->zoomStepScroll : -this->zoomStepScroll;
+    double newZoom = this->zoom * std::exp(0.5 * newZoomStep);
     this->zoomSequenceChange(newZoom, false);
 }
 
 void ZoomControl::startZoomSequence() {
     Rectangle rect = getVisibleRect();
-    startZoomSequence({rect.x + rect.width / 2.0, rect.y + rect.height / 2.0});
+    startZoomSequence({rect.width / 2.0, rect.height / 2.0});
```
personalizedrefrigerator, line 122: Why don't we need `rect.x` and `rect.y` here?

src/control/zoom/ZoomControl.cpp
```diff
@@ -18,7 +18,9 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
                 (event->direction == GDK_SCROLL_UP || (event->direction == GDK_SCROLL_SMOOTH && event->delta_y < 0)) ?
                         ZOOM_IN :
                         ZOOM_OUT;
-        zoom->zoomScroll(direction, {event->x, event->y});
+        int rx, ry;
+        gdk_window_get_root_coords(gtk_widget_get_window(widget), 0, 0, &rx, &ry);
+        zoom->zoomScroll(direction, {event->x_root - rx, event->y_root - ry});
```
personalizedrefrigerator, line 23: Please comment this piece! Some developers (like myself) have little experience with GTK! The documentation for `gdk_window_get_root_coords`. Why aren't we using `event->x` and `event->y`?

redweasel:
I replaced `event->x` with `event->x_root` as well, so instead of calculating the offset as double it calculates the center as integers independent of the scroll value. I don't understand where `event->x` comes from and replacing it in this way seemed to fix the page jump problem when zooming with ctrl+mousewheel. But on the other hand, from googling it seemed like `event->x_root` would be in window space, but it was in screen space for me...

personalizedrefrigerator (commented):
From a quick test, this seems to work very well! Thank you!

I have a few comments. These are mostly just me not understanding changes -- additional commenting/moving things into methods with descriptive names could be useful!

src/gui/inputdevices/TouchInputHandler.cpp
personalizedrefrigerator: > I replaced `event->x` with `event->x_root` as well, so instead of calculating the offset as double it calculates the center as integers independent of the scroll value. I don't understand where `event->x` comes from and replacing it in this way seemed to fix the page jump problem when zooming with ctrl+mousewheel. But on the other hand, from googling it seemed like `event->x_root` would be in window space, but it was in screen space for me...

I just saw this! It might be worth documenting this here!

src/control/zoom/ZoomControl.cpp
personalizedrefrigerator: > I replaced `event->x` with `event->x_root` as well, so instead of calculating the offset as double it calculates the center as integers independent of the scroll value. I don't understand where `event->x` comes from and replacing it in this way seemed to fix the page jump problem when zooming with ctrl+mousewheel. But on the other hand, from googling it seemed like `event->x_root` would be in window space, but it was in screen space for me...

Okay! It might be best to factor things related to this into a single place, so it's easier to change/read. Thank you!

personalizedrefrigerator:
On non-PDFs, I'm having an issue with ctrl+mousewheel zooming... Try opening a blank document, adding a few pages, scrolling down, then ctrl+mousewheel zooming in and out. When I did this, the view moved to the top of one of the pages as soon as I started zooming! It doesn't do this in master.


https://user-images.githubusercontent.com/46334387/106765120-53a66380-65ed-11eb-9688-cff5bcd3c594.mp4

David96:
I looked a bit into this event->x vs event->x_root thing, and the difference seems to be that x_root is relative to the window (as in window managed by a window manager, a GdkWindow might be something different and therefore the confusion with the docs?) and event->x/y is relative to the scroll view, so on large documents event->y becomes very big (which if my understanding of mantissa/exponent is correct could result in less precision for double calculations). Furthermore the values don't seem to have double precision for me anyway, but rather the _root coordinates return always something in the form SomeNumber.0 and the other coordinates return some higher precision value but always the same numbers after the "."

Another weird thing I observed which could hint to a bug in the zooming code (and maybe be related to the bug just reported by @personalizedrefrigerator) is that when scrolling at the end of a document with ctrl+scroll the event->y coordinate jumps to a very small number (basically meaning the document was scrolled up) and then back to the high number which it is supposed to be (without me actually seeing any jumping). Looking at the code, one problem I see is that for mouse wheel scroll, the soom sequence is never ended. Therefore the zoomSequenceStart is never properly set after the first mousewheel zooming. One would need to figure out when the scrolling has stopped, maybe listen to a button release event on ctrl? I think this might fix @personalizedrefrigerator's issue.

This is the quickndirty patch I used to observe the coordinates:
```diff
diff --git a/src/control/zoom/ZoomControl.cpp b/src/control/zoom/ZoomControl.cpp
index 9c8713e4..9c16d823 100644
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -20,6 +20,7 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
                         ZOOM_OUT;
         int rx, ry;
         gdk_window_get_root_coords(gtk_widget_get_window(widget), 0, 0, &rx, &ry);
+        printf("y_root: %f, y: %f\n", event->y_root, event->y);
         zoom->zoomScroll(direction, {event->x_root - rx, event->y_root - ry});
         return true;
     }
```

src/control/zoom/ZoomControl.cpp
personalizedrefrigerator: I don't know why, but reverting this seems to stop the view from jumping at the beginning of a control+wheel... It does, however, make the view scroll down while doing a ctrl+wheel...

src/control/zoom/ZoomControl.cpp
```diff
@@ -85,14 +86,15 @@ void ZoomControl::zoomOneStep(ZoomDirection direction, utl::Point<double> zoomCe
     this->setZoomFitMode(false);
 
     startZoomSequence(zoomCenter);
-    double newZoom = (direction == ZOOM_IN) ? this->zoom + this->zoomStep : this->zoom - this->zoomStep;
+    double newZoomStep = (direction == ZOOM_IN) ? this->zoomStep : -this->zoomStep;
+    double newZoom = this->zoom * std::exp(0.5 * newZoomStep);
```
personalizedrefrigerator, line 90: This makes ctrl+wheel zooming so much better!!!! Thank you!

src/control/zoom/ZoomControl.cpp
redweasel: I changed `startZoomSequence` to use window coordinates not page relative coordinates. AFAIK `rect.x` would make it page relative. I tested it and it works this way.

src/control/zoom/ZoomControl.cpp
redweasel: I have no experience with GTK as well. Where (which file) would you put the function for translating screen coordinates to window coordinates?

redweasel:
I think I might have fixed the issue with the page jumps now by always starting the zoom-sequence appropriately (?). It was checking for a change in mouse position previously, but that only introduced bugs and wasn't helping with performance at all (which I think might have been the reason it was there in the first place)

Roadmap for Zooming:
- There is some error of ~3 pixels in `getPaddingLeftOfPage()` I think.
- mouse zoom steps are now different -> reflect that in the settings gui strings and code comments
- Fix/Remove the Gtk touch workaround (it is not working for me at all) (bug not introduced here)
- Fix touch drawing/zoom scroll (does zoom and scroll weirdly, when enabled and three fingers are touched down)
- Make layout system independent of zoom level -> no "`unscaledPixels`" between pages

src/control/zoom/ZoomControl.cpp
personalizedrefrigerator: Okay! Thank you!

src/control/zoom/ZoomControl.cpp
personalizedrefrigerator: I think [MainWindow](https://github.com/xournalpp/xournalpp/blob/master/src/gui/MainWindow.h) might be reasonable. I don't see anything like this already in that file, though...
It might also make sense to put this in something in `src/util`, but I don't see a file that provides similar/related functionality.

redweasel: Can we leave it as is for this PR and make a separate cleanup PR with no functional changes after this one merged? As I pointed out, this is not the only thing that needs cleanup.

personalizedrefrigerator: I think that's fine!

src/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -137,16 +132,25 @@ void TouchInputHandler::zoomStart() {
         zoomControl->setZoomFitMode(false);
     }
 
+    auto center = (this->priLastAbs + this->secLastAbs) / 2.0;
+
     auto* mainWindow = inputContext->getView()->getControl()->getWindow();
+    // use screen pixel coordinates for the zoom center
+    // as relative coordinates depend on the changing zoom level
+    int rx, ry;
+    gdk_window_get_root_coords(*mainWindow, 0, 0, &rx, &ry);
```
David96, line 141: This is wrong. The mainWindow here is of type MainWindow, not GtkWindow. We need the gtk window of the scroll widget, which seems to be a bit harder to get. Otherwise this will return {0,0} instead of the right root_coords which causes a little offset in the zoom center. 

This fixes that but without caring about the touch workaround (in that case, another GtkWidget seems to be used in MainWindow). Also, maybe there is another way of getting the right widget/root coords without adding an additional method to MainWindow.
```diff
diff --git a/src/gui/MainWindow.cpp b/src/gui/MainWindow.cpp
index 3e03ee08..6dc30672 100644
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -340,6 +340,8 @@ void MainWindow::initHideMenu() {
 
 auto MainWindow::getLayout() -> Layout* { return gtk_xournal_get_layout(GTK_WIDGET(this->xournal->getWidget())); }
 
+auto MainWindow::getWinXournal() -> GtkWidget* { return this->winXournal; }
+
 auto cancellable_cancel(GCancellable* cancel) -> bool {
     g_cancellable_cancel(cancel);
 
diff --git a/src/gui/MainWindow.h b/src/gui/MainWindow.h
index 875ac3fc..bc8245fc 100644
--- a/src/gui/MainWindow.h
+++ b/src/gui/MainWindow.h
@@ -98,6 +98,7 @@ public:
 
     bool isGestureActive();
 
+    GtkWidget* getWinXournal();
 
     /**
      * Disable kinetic scrolling if there is a touchscreen device that was manually mapped to another enabled input
diff --git a/src/gui/inputdevices/TouchInputHandler.cpp b/src/gui/inputdevices/TouchInputHandler.cpp
index 73a0737a..9573ebbe 100644
--- a/src/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/gui/inputdevices/TouchInputHandler.cpp
@@ -138,10 +138,11 @@ void TouchInputHandler::zoomStart() {
     // use screen pixel coordinates for the zoom center
     // as relative coordinates depend on the changing zoom level
     int rx, ry;
-    gdk_window_get_root_coords(*mainWindow, 0, 0, &rx, &ry);
+    gdk_window_get_root_coords(gtk_widget_get_window(mainWindow->getWinXournal()), 0, 0, &rx, &ry);
     center.x -= rx;
     center.y -= ry;

     // When not using touch drawing, we're using a different scrolling method.
     // This requires different centering.
     /*if (!mainWindow->getGtkTouchscreenScrollingEnabled()) {
 ```

**Comment by redweasel** [comment]
Thanks @David96. I included your patch (2bfeaaf fix window coordinates n TouchInputHandler)

I have also fixed the wrong padding calculation.
I also made the zoom steps comply to the settings, so from 100% Zoom, one step will add 10% if thats whats in your settings. The mouse scroll zoom behavior that @personalizedrefrigerator seemed to really like is unchanged by this.

**Comment by David96** [comment]
Looking at it again, it might be better to do something like this instead of modifying MainWindow. This should also handle the touch workaround case.

But in the end, someone actually associated with this project should comment on that.

```diff
diff --git a/src/control/zoom/ZoomControl.cpp b/src/control/zoom/ZoomControl.cpp
index 42243569..18943302 100644
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -207,6 +207,7 @@ void ZoomControl::addZoomListener(ZoomListener* l) { this->listener.emplace_back
 void ZoomControl::initZoomHandler(GtkWidget* window, GtkWidget* widget, XournalView* v, Control* c) {
     this->control = c;
     this->view = v;
+    this->widget = widget;
     gtk_widget_add_events(widget, GDK_TOUCHPAD_GESTURE_MASK);
     g_signal_connect(widget, "scroll-event", G_CALLBACK(onScrolledwindowMainScrollEvent), this);
     g_signal_connect(widget, "event", G_CALLBACK(onTouchpadPinchEvent), this);
@@ -226,6 +227,8 @@ void ZoomControl::fireZoomRangeValueChanged() {
     }
 }
 
+auto ZoomControl::getWidget() -> GtkWidget* { return this->widget; }
+
 auto ZoomControl::getZoom() const -> double { return this->zoom; }
 
 auto ZoomControl::getZoomReal() const -> double { return this->zoom / this->zoom100Value; }
diff --git a/src/control/zoom/ZoomControl.h b/src/control/zoom/ZoomControl.h
index a83980b8..084c381a 100644
--- a/src/control/zoom/ZoomControl.h
+++ b/src/control/zoom/ZoomControl.h
@@ -169,6 +169,8 @@ public:
     /// Get visible rect on xournal view, for Zoom Gesture
     Rectangle<double> getVisibleRect();
 
+    GtkWidget* getWidget();
+
     void setZoomStep(double zoomStep);
 
     void setZoomStepScroll(double zoomStep);
@@ -192,6 +194,7 @@ private:
     XournalView* view = nullptr;
     Control* control = nullptr;
     std::vector<ZoomListener*> listener;
+    GtkWidget* widget = nullptr;
 
     /**
      * current Zoom value
diff --git a/src/gui/inputdevices/TouchInputHandler.cpp b/src/gui/inputdevices/TouchInputHandler.cpp
index 73a0737a..fcd49ccc 100644
--- a/src/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/gui/inputdevices/TouchInputHandler.cpp
@@ -134,11 +137,10 @@ void TouchInputHandler::zoomStart() {
 
     auto center = (this->priLastAbs + this->secLastAbs) / 2.0;
 
-    auto* mainWindow = inputContext->getView()->getControl()->getWindow();
     // use screen pixel coordinates for the zoom center
     // as relative coordinates depend on the changing zoom level
     int rx, ry;
-    gdk_window_get_root_coords(*mainWindow, 0, 0, &rx, &ry);
+    gdk_window_get_root_coords(gtk_widget_get_window(zoomControl->getWidget()), 0, 0, &rx, &ry);
     center.x -= rx;
     center.y -= ry;
 ```

File: src/gui/Layout.cpp [hunk 8]
```diff
@@ -199,12 +199,12 @@ void Layout::layoutPages(int width, int height) {
...
-    auto const borderX = std::max<SBig>(h_padding, centeringXBorder);
-    auto const borderY = std::max<SBig>(v_padding, centeringYBorder);
+    auto const borderX = static_cast<double>(std::max<SBig>(h_padding, centeringXBorder));
+    auto const borderY = static_cast<double>(std::max<SBig>(v_padding, centeringYBorder));
 
     // initialize here and x again in loop below.
-    auto x = borderX;
-    auto y = borderY;
+    double x = borderX;
+    double y = borderY;
```
Febbe, line 207:
```suggestion
    auto x = borderX;
    auto y = borderY;
```

src/gui/Layout.cpp
```diff
@@ -260,13 +260,13 @@ void Layout::layoutPages(int width, int height) {
 
 
     // accumulated - absolute pixel location for use by getViewAt() and updateVisibility()
-    auto totalWidth = borderX;
+    double totalWidth = borderX;
```
Febbe, line 263:
```suggestion
    auto totalWidth = borderX;
```

src/gui/Layout.cpp
```diff
@@ -260,13 +260,13 @@ void Layout::layoutPages(int width, int height) {
...
     std::transform(
             begin(this->pc.widthCols), end(this->pc.widthCols), begin(this->colXStart), [&totalWidth](auto&& widthCol) {
                 return strict_cast<std::remove_reference_t<decltype(widthCol)>>(totalWidth +=
                                                                                 widthCol + XOURNAL_PADDING_BETWEEN);
             });
-    auto totalHeight = borderY;
+    double totalHeight = borderY;
```
Febbe, line 269:
```suggestion
    auto totalHeight = borderY;
```

src/gui/Layout.cpp
```diff
@@ -293,25 +293,24 @@ auto Layout::getPaddingLeftOfPage(size_t pageIndex) const -> int {
...
 
     auto const pageXLocation = as_signed(this->mapper.at(pageIndex).first);
 
     // No page pairing or we haven't rendered enough pages in the row for
     // page pairing to have an effect,
-    if (!isPairedPages || pageXLocation == 0) {
-        return strict_cast<int>(pageXLocation * XOURNAL_PADDING_BETWEEN + as_signed(paddingBefore));
+    if (!isPairedPages) {
+        return strict_cast<int>(pageXLocation * XOURNAL_PADDING_BETWEEN + XOURNAL_PADDING_BETWEEN / 2 +
+                                as_signed(paddingBefore));
```
Febbe, line 305: Page 0 should have no `XOURNAL_PADDING_BETWEEN` as offset from the left view anchor.
Where comes the `XOURNAL_PADDING_BETWEEN / 2` from?

Febbe: Also the `XOURNAL_PADDING_BETWEEN` is the distance between two pages.

src/gui/Layout.cpp
```diff
@@ -293,25 +293,24 @@ auto Layout::getPaddingLeftOfPage(size_t pageIndex) const -> int {
...
-        auto shadowRoomInsidePairs = pageXLocation / 2 * XOURNAL_ROOM_FOR_SHADOW;
-
-        return strict_cast<int>(paddingBetweenPairs + shadowRoomInsidePairs + paddingBefore);
+        auto columnPadding =
+                XOURNAL_PADDING_BETWEEN + strict_cast<int>(pageXLocation / 2) * (XOURNAL_PADDING_BETWEEN * 2);
+        if (pageXLocation % 2 == 0) {
+            return strict_cast<int>(columnPadding - XOURNAL_ROOM_FOR_SHADOW + paddingBefore);
+        } else {
+            return strict_cast<int>(columnPadding + XOURNAL_ROOM_FOR_SHADOW + paddingBefore);
+        }
```
Febbe, line 313: `columnPadding` should be equal to `XOURNAL_PADDING_BETWEEN`

Febbe (changes requested):
I would aprechiate, if you split up your PR, since I am currently not happy with the change from using local positions to global positions.

Also, I don't have much time to review this in detail. So we have to wait for the next release (1.1.1) instead of 1.1. When you split it, we can at least merge the fix before the release of 1.1.0

src/gui/Layout.cpp
redweasel: My version is correct AFAICT. See the layout code, to understand how this actually plays out.

src/gui/Layout.cpp
redweasel: My version is correct AFAICT. See the layout code, to understand how this actually plays out.

redweasel:
I heavily disagree on the usage of `auto` here, but here you go, I changed it. The usage of `auto` was after all the root source of this issue, and the reason why nobody before me found it.

I will split this PR up into the layout part and the zoom input part and squash the commits when I'm finished.

Febbe:
@redweasel as I understand the use of integral values for zoomed ranges was the issue. 

"Using auto everywhere" has the advantage, 

 - That it first is always the correct type
 - That there will never be implicit casts
 - That it defers the need of casts to to place the value is required

Also the use of explicit types caused a huge bug on 32 bit and arm64 systems, because they defined size_t different.
So using `auto` is not the problem it's more like, that we allow implicit casts.

redweasel:
We do not (really) allow lossy implicit casts. There is always a compiler warning if an implicit cast could cause problems. The problem is more like: *we don't listen to the compiler enough.*

*(I found a bunch of dead code in xournalpp regarding file attachment loading, using the compiler warnings, introduced by a clang-tidy operation, but since I wasn't able to produce a bug I didn't mention it yet)*

In this particular case it has to be double and double only! If it would be float, we would have a problem right here again even though it behaves the same, the precision would not suffice (so 0.0 better be double on every possible platform out there or this little change to auto already caused a new issue). `auto` just makes entry harder for new developers. I know this project uses auto everywhere so I comply.

Remember `auto` is controversial to talk about (just google it, there are many people on both ends of the spectrum with good reasons. I'm on the anti part of the spectrum. Hope I am still welcome here xD).

Febbe:
@redweasel Yes I also know both sides :D. I mostly don't care much, but in this case (Layout calculation) we had a huge bug because of named but implementation defined types.

Newcomers are also welcome to ask questions if they do not understand things.

For the split of your PR: I'll try to check all pitfalls, if your code performs better, and I do not find bugs, I'll approve it. So we might merge it before 1.1.0.

Regarding the `XOURNAL_PADDING_BETWEEN` when no one changed the Layout behavior it should be still the distance between two pages. But I'll check the code again on my machine. But for the future it would be way easier, if you explain, why your code is correct, instead of claiming it is correct and linking to the source.

 > Hope I am still welcome here xD
 
Of course, you are :D. Should've marked some `auto` changes as nitpicks.
In particular the `double value = 0;` to `auto value = 0.0` which is only a beauty/consistency nitpick.

Febbe:
If you want to improve the hole Layout / zoom scroll code you could make a plan, we want to rewrite it completely, using some sort of matrix transformation instead of separated zooming and scrolling. Also, it could be good to split the Layout into a 2PageLayout / SinglePageLayout / PresentationLayout.

But I am not sure if we should wait until we change to gtk4. The event system will be changed there fundamentally.

redweasel:
I have now only included the relevant changes to Layout in this PR. I will create another PR for the Zoom Control fixes. (If you try this PR you will immediately see the issues). The issue that this PR really fixes now is the jitter and the wrong zoom center in different configurations.

Sorry for not explaining the code for the padding. I just copied the code directly from the Layout algorithm and removed the unneseccary parts. I don't know why it look like this, it's just what the Layout code did.

@Febbe Yes I was thinking about rewriting the Layout/Zoom code using matrices as well. Xournalpp can have some abstraction around the GTK+ events I think. The Zoom/Layout code should not touch any gtk events from its side. I will make a plan as an issue as soon as I get to it.

Febbe:
Looks good so far, but I think, if we chose to use doubles in the Layout, we shoul do it consequently: 

**nitpick**, because it's constexpr: *https://github.com/xournalpp/xournalpp/blob/d2aaf62208d9c991bdd066d222d2b68b2cfe35fc/src/gui/Layout.cpp#L14-L27

change values to 10.0

**bugprone:** implicit cast, integer division
https://github.com/xournalpp/xournalpp/blob/d2aaf62208d9c991bdd066d222d2b68b2cfe35fc/src/gui/Layout.cpp#L237-L238

Use `2.0` as divisor not `2`

https://github.com/xournalpp/xournalpp/blob/d2aaf62208d9c991bdd066d222d2b68b2cfe35fc/src/gui/Layout.cpp#L161-L162

`pc.minWidth` is still a size_t type, consider to change this also to double or calculate the value with a temporary.
```
    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), double(pc.minWidth)));
    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), double(pc.minHeight)));
```

src/gui/Layout.cpp
```diff
@@ -158,8 +158,8 @@ void Layout::recalculate_int() const {
     pc.minWidth = as_unsigned(2 * hPadding + as_signed_strict((pc.widthCols.size() - 1) * XOURNAL_PADDING_BETWEEN));
     pc.minHeight = as_unsigned(2 * vPadding + as_signed_strict((pc.heightRows.size() - 1) * XOURNAL_PADDING_BETWEEN));
 
-    pc.minWidth = std::accumulate(begin(pc.widthCols), end(pc.widthCols), pc.minWidth);
-    pc.minHeight = std::accumulate(begin(pc.heightRows), end(pc.heightRows), pc.minHeight);
+    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), pc.minWidth));
+    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), pc.minHeight));
```
Febbe, line 162: Currently, you still use a `size_t` as starting value, this results in casting each value of the vectors to be casted to `size_t` and then added to the starting value.

```suggestion
    pc.minWidth = floor_cast<size_t>(std::accumulate(begin(pc.widthCols), end(pc.widthCols), double(pc.minWidth)));
    pc.minHeight = floor_cast<size_t>(std::accumulate(begin(pc.heightRows), end(pc.heightRows), double(pc.minHeight)));
```

Febbe:
@redweasel thank you for your contribution!

PR #2746, opened 2021-02-04
Fix GTK touch workaround

### LINKED ISSUE #2745: GTK Touch Workaround Broken
**Affects versions :**
 - OS: Manjaro Linux
 - (Linux only) Desktop environment: Plasma/X11
 - liibgtk 3.24.24
 - Version of Xournal++: Current master (ea943783fbf990bf5f22c17bbe0cf8245da25722)
 - Installation method: Built from source

**Describe the bug**
See https://github.com/xournalpp/xournalpp/commit/67927ead3c36bcf6afa7723d7f80ca47650fffda#r46743703 -- `gtk_xournal_get_preferred_width/gtk_xournal_get_preferred_height` return the size of the layout, even when not in a scrolled widget. When we're handling scrolling ourselves (when the workaround is enabled), we should be returning a value much smaller than the layout size.

**To Reproduce**
1. Enable touch workaround
2. Restart app
3. Try to scroll

**Expected behavior**
I should be able to scroll

**Additional context**
Regression from 67927ead3c36bcf6afa7723d7f80ca47650fffda /gtk_xournal_get_preferred_width/gtk_xournal_get_preferred_height. Should return a constant value when the workaround is enabled, but doesn't.

personalizedrefrigerator:
See https://github.com/xournalpp/xournalpp/commit/67927ead3c36bcf6afa7723d7f80ca47650fffda#r46743703 -- `gtk_xournal_get_preferred_width/gtk_xournal_get_preferred_height` return the size of the layout, even when not in a scrolled widget. When we're handling scrolling ourselves (when the workaround is enabled), we should be returning a value much smaller than the layout size.

This should fix #2745.

**Edit:** The difference between this PR and #2613 is that `ScrollHandler` has a different implementation if we're using the touch workaround than if we're not. #2613 fixed this by setting the `minimal_width` and `minimal_height` to a constant regardless of whether the workaround was enabled.

### DIFF
--- a/src/gui/widgets/XournalWidget.cpp
+++ b/src/gui/widgets/XournalWidget.cpp
@@ -163,12 +163,12 @@ static void gtk_xournal_init(GtkXournal* xournal) {
 
 static void gtk_xournal_get_preferred_width(GtkWidget* widget, gint* minimal_width, gint* natural_width) {
     GtkXournal* xournal = GTK_XOURNAL(widget);
-    *minimal_width = *natural_width = xournal->layout->getMinimalWidth();
+    *minimal_width = *natural_width = xournal->scrollHandling->getPreferredWidth();
 }
 
 static void gtk_xournal_get_preferred_height(GtkWidget* widget, gint* minimal_height, gint* natural_height) {
     GtkXournal* xournal = GTK_XOURNAL(widget);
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
Closing in favor of #2308

PR #2749, opened 2021-02-05
Fix #1679 -- Pen activates after exiting page when using floating toolbar

### LINKED ISSUE #1679: Experimental floating toolbox: Pen activates automatically on crossing boundary edges of page.
**Affects versions :**
 - OS: Manjaro linux 18.1.4
 -  libgtk : gtk-2.0/2.10.0
 - Version of Xournal++ : 1.0.16

**Describe the bug**
I have set up the experimental toolbox (floating) and it is great. But one issue I have found is that after the toolbox closes, and if I cross the edges of the page (eg from one page to the next, or side edges), I find that the pen starts to write by itself even though it is not touching the surface of the tablet I am using.

**Screenshots of Problem**
1.Opening toolbox and selecting pen tool:
image
2.Pen writing by itself:
image

### LINKED ISSUE #2762: Pen activates automatically on crossing page edges using new input system
**Affects versions :**
 - Ubuntu 20.04, Gnome X11
- Xournal++ 1.1.0+dev
- libgtk: 3.24.20
- PPA

**Describe the bug**
I use a Wacom Stylus with the "merge button events with stylus tip events" setting on and the new input system. I use the "merge button events" option because it enables me to use the buttons on the stylus without actually pressing the stylus on the tablet.
 
After using a button on the stylus that is mapped to any action (tested with "Select rectangle" and "Hand"), the last tool(pen or highlighter) will activate itself after crossing a page boundary. Similar issue will happen if I move my cursor into my right hand toolbar, and then return to the canvas. The result of this is that i get scribbles whenever I cross a page, which I then have to delete manually.

This issue seems related: #1679.

**To Reproduce**
1. Activate new input system.
2. Use a button on the stylus. Move to another page.
3. Pen will start writing on new page automatically

**Expected behavior**
Pen does not activate itself.

**Additional context**:

This does not happen when using the old input system

personalizedrefrigerator:
This should fix #1679. Previously, calls to `PenInputHandler::actionMove` would result in drawing even when not `PenInputHandler::inputRunning`. This is a bit of a workaround. 
`PenInputHandler::actionMove` probably shouldn't be called if we're not drawing/the pen isn't down!

This should also fix #2762.

### DIFF
--- a/src/gui/inputdevices/PenInputHandler.cpp
+++ b/src/gui/inputdevices/PenInputHandler.cpp
@@ -182,6 +182,11 @@ double PenInputHandler::filterPressure(PositionInputData const& pos, XojPageView
 }
 
 auto PenInputHandler::actionMotion(InputEvent const& event) -> bool {
+    // If we're not handling input, do nothing!
+    if (!this->inputRunning) {
+        return false;
+    }
+
     /*
      * Workaround for misbehaving devices where Enter events are not published every time
      * This is required to disable outside scrolling again


personalizedrefrigerator:
@Technius This and #2760 are significant usability fixes for me. I hope this PR can be reviewed soon!

rolandlo:
It does solve #2762, but I can still see some strange behavior. If I cross the border between two pages rapidly, the stroke on the old page becomes a ghost stroke, which only disappears once this screen section is repainted. For example if I press the pen and go rapidly from page 2 to page 1 and back and come close to the stroke painted on page 2, then it slowly disappears.
I guess a repaint/rerender is missing somehwhere.

personalizedrefrigerator:
I having trouble reproducing this... Does similar behavior happen with Xournal++ before this PR?

rolandlo:
Yes, similar behavior happens before this PR as well. So maybe we can merge this one and fix the problem in another PR.

To reproduce: just draw forth and back across two pages without releasing the mouse/stylus and you should sooner or later observe this strange behavior.

nicolae-stroncea:
To add to this, I noticed the same behaviour on the text tool, whereas scrolling on a new page will activate a small text box near the border. It's not a real issue compared to the pen and higlighter which actually change the page, though

src/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -182,6 +182,11 @@ double PenInputHandler::filterPressure(PositionInputData const& pos, XojPageView
 }
 
 auto PenInputHandler::actionMotion(InputEvent const& event) -> bool {
+    // If we're not handling input, do nothing!
+    if (!this->inputRunning) {
+        return false;
+    }
+
```
personalizedrefrigerator, line 189: At present, this PR only changes/adds these 5 lines. As long as this doesn't cause a regression, I would like to keep this PR as small as possible. :)

nicolae-stroncea: Actually it seems like your fix does prevent the behaviour from happening with other tools(including text tool). Would be great to have this PR merged since it apparently also prevents a destructive action. If you try the same behaviour but with the "Vertical Space" tool, it seems like it actually deletes the elements that it shifts, and they're not recoverable with the "Undo" tool.

rolandlo:
Ok, thanks for the fix.

Issue #2755, opened 2021-02-06
Minor UX issues


nicolae-stroncea:
I'm a big fan of Xournal++. I've been using it for 2 years now. These are some minor UX issues I've observed over time and that are currently applicable on 1.1.0 dev.

1. `Shift+Ctrl+Left` shortcut switches layers when writing text. This shortcut should either be changed or disabled(probably disabled) when writing text because it is very commonly used to select the word to your left. This is especially confusing if you're not used to layers. This is related to #2679 where another Xournal++ shortcut interferes with common text editing shortcuts.
2. Drawing shapes are not aligned in the UI according to their shortcuts. I think this is common in all default toolbars, but is especially noticeable in the "Right Hand Note Taking" Toolbar. Intuitively, I would expect that I cycle through them starting at 1 in increasing order. Current order:
    * Circle(Ctrl+3)
    * Rectangle(Ctrl+2)
    * Arrow(Ctrl+4)
    * Line(Ctrl+6)
3. Default icon to start recording makes it seem like you are currently being recorded. Usually the red circle is used to indicate you are recording live. Possible solutions are to either make the icon black, or change it to another icon. A possible UI flow might be: 
    * Inactive(icon is black), active(icon is red)
    * Inactive(icon is black), active(icon is black and has a grey outline as current active tools have).
![image](https://user-images.githubusercontent.com/39338488/107108210-3dbebb80-6804-11eb-8d90-5c73a15f0cd6.png)
4. The option to "Copy current page size instead" in "Configure new page template" should be enabled by default. It seems intuitive that when a user clicks to create a new page, they'll get a page that mirrors the current one. What currently happens in some of my PDFs is that I'll click on "New Page" and I'll get a huge page compared to the ones I have currently in the PDF. Other reasons this should be the default is that it looks better for a PDF document to have all pages be the same size (which is the case in pretty much any PDF documents I've ever used). A possible issue with current default might also be if you're printing it and it has different-sized pages(may not be an issue at all, not sure what would happen here).
    * *Side-note*: It would be great if common page sizes (Template options + "current page size") are listed in the drop-down for the  "Insert Page" button in the toolbar. This would both reduce the number of clicks and makes that drop-down the primary place you need for quick new page adjustments.

schachmett:
[...]

Issue #2679, opened 2021-01-18
Ctrl+Del shortcut should conform to usual text editing conventions


schachmett:
_This comment was edited: Ctrl+Backspace is actually working while Ctrl+Del is not. In the original comment, I accidentally got that backwards_

**Affects versions :**
OS: Linux Ubuntu 20.04 (Gnome)
Versions: Xournal++ 1.0.20,  libgtk: 3.24.20 
Installation method: I think it was regular apt official release

**Describe the bug**
Xournalpp overwrites the default Ctrl+Del shortcut. Usually, Ctrl+Del deletes the word succeeding the cursor. Xournal++ however overwrites this so that it deletes the current page.

**To Reproduce**
1. add a text field to a document
2. write something
3. hit Ctrl+Del
4. the page disappears (not sure if this is also the case in single-page documents)

**Expected behavior**
The word succeeding the cursor should be deleted

**Describe the solution you'd like**
Please remove the current Ctrl+Del shortcut or change it to different keys. The default functionality should then automatically work (I guess the OS does this).

**Additional context**
I use Xournal++ mainly for annotating and highlighting regular pdfs, so I use the text feature often. It can be a bit of a shock when you are typing something and suddenly the whole page vanishes. It can also lead to severe data loss, depending on when you figure out what happened.
For Ctrl+Backspace, the word preceding the cursor is deleted. This is the normal behavior and is working as expected in Xournal++.

Sorry if this is in the wrong category. I was not sure if this is bug or a feature request. To me it feels like a bug.

personalizedrefrigerator:
For me, Ctrl+Backspace works, but Ctrl+Del doesn't (it deletes the current page). I'm running KDE Neon/Plasma X11 and Xournal built from a recent commit.

Please try version 1.1.0+dev. Nightly AppImages can be found here: https://github.com/xournalpp/xournalpp/releases/tag/nightly.

personalizedrefrigerator:
Related issue: https://github.com/xournalpp/xournalpp/issues/919

schachmett:
Thank you and sorry, I actually meant it the other way around. Ctrl+Backspace works, but Ctrl+Del does not.
Would it be enough to just change the glade file? What would be a sensible alternative? Maybe Ctrl+Shift+Del?
If that would be all it takes and if it is okay to just change this shortcut, I could make a PR.

Issue #2756, opened 2021-02-06
Zoom in shortcut doesn't work


nicolae-stroncea:
**Affects versions :**
 - Ubuntu 20.04, Gnome X11
- Xournal++ 1.1.0+dev
- libgtk: 3.24.20
- PPA

**Describe the bug**
`Ctrl + +` zoom in shortcut does not work. 

**To Reproduce**

Zoom in using the shortcut.

personalizedrefrigerator:
For anyone interested in working on this, these links might be helpful:
  * [Some relevant areas of the codebase for adding keyboard shortcuts to existing actions](https://github.com/xournalpp/xournalpp/issues/2679#issuecomment-763130247).
  *  #2766

nicolae-stroncea:
@personalizedrefrigerator  where did you find the: `Ctrl+Shift++` shortcut? I looked through the code and can't find it.

In main.glade, it looks mapped correctly to : `Ctrl++`:
```
                          <object class="GtkImageMenuItem" id="menuViewZoomIn">
                            <property name="label">gtk-zoom-in</property>
                            <property name="name">menuViewZoomIn</property>
                            <property name="visible">True</property>
                            <property name="can-focus">False</property>
                            <property name="use-underline">True</property>
                            <property name="use-stock">True</property>
                            <signal name="activate" handler="ACTION_ZOOM_IN" swapped="no"/>
                            <accelerator key="plus" signal="activate" modifiers="GDK_CONTROL_MASK"/>
                            <accelerator key="KP_Add" signal="activate" modifiers="GDK_CONTROL_MASK"/>
                          </object>
```
however it doesn't seem to respond to the shortcut at all.

Is is possible that it doesn't respond because it's mapped to "Ctrl+Shift++" instead?

personalizedrefrigerator:
We might need to map it to "Ctrl ="... This might be un-intuitive on keyboards for which '=' and '+' use different keys, however:
![Keyboard layout with different keys mapped to zoom in](https://user-images.githubusercontent.com/46334387/109106129-0e54ed80-76e4-11eb-99a7-bc7a750b8331.png)

... as compared to...

![EN-US layout with + and = mapped to the same key](https://user-images.githubusercontent.com/46334387/109106239-40fee600-76e4-11eb-8b0f-ac9d6395389e.png)

nicolae-stroncea:
@personalizedrefrigerator  you're right. My keyboard has the `+` on the same key as `=`.

[...]

PR #2766, opened 2021-02-07
Changed shortcuts for page deletion and layer navigation

schachmett:
This is a proposition to change the` Ctrl+Del` shortcut for page deletion to `Ctrl+Shift+Del` to not interfere with text editing shortcuts according to #2679 .
The same was done for `Ctrl+Shift+Left` --> `Shift+PageDown`,  ` Ctrl+Shift+Right` --> `Shift+PageUp`  (see #2755). 
I don't mean to change the shortcut rationale randomly, please comment and I will change the PR as requested.

### DIFF
--- a/ui/main.glade
+++ b/ui/main.glade
@@ -1041,7 +1041,7 @@
                             <property name="label" translatable="yes">_Previous Layer</property>
                             <property name="use-underline">True</property>
                             <signal name="activate" handler="ACTION_GOTO_PREVIOUS_LAYER" swapped="no"/>
-                            <accelerator key="Left" signal="activate" modifiers="GDK_SHIFT_MASK | GDK_CONTROL_MASK"/>
+                            <accelerator key="Page_Down" signal="activate" modifiers="GDK_SHIFT_MASK"/>
                           </object>
                         </child>
                         <child>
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
@rolando It would be nice to have this. If you have time, please review!

rolandlo:
@schachmett I agree with your changes. There is another issue with shortcuts:
 
In the text editor (in Xournal++) the code 
https://github.com/xournalpp/xournalpp/blob/bb6e237ed9720b9fb8e1919a3bc53019df498148/src/gui/TextEditor.cpp#L275-L290
says that
- Ctrl+b toggles bold/not bold (this works)
- Ctrl+plus increases the text size by 1 (this zooms in instead)
- Ctrl+minus decreases the text size by 1 (this zooms out instead)

Do you have a solution for that as well?

nicolae-stroncea:
This will close #2861

nicolae-stroncea:
@rolandlo  one other issue with `Ctrl+` shortcut is that it doesn't work on North American/British Keyboards, see also #2756

personalizedrefrigerator:
@rolando @Febbe This is a review request! I frequently try to use Ctrl+Shift+Left and Ctrl+Shift+Right when editing text and am finding this issue rather annoying.

rolandlo (approved):
@personalizedrefrigerator Sorry for the delay. I hoped to get a solution for other accelerators as well, but that can also be treated later.
If no objections are made, I will merge this within 22 hours.

PR #2771, opened 2021-02-08
Automatically request dark/light GTK theme based on "use dark theme" settings option

### LINKED ISSUE #2764: Xournal 1.1.0 Update broke Dark Theme
Windows 10, Xournal 1.1.0. The newest update broke the dark theme I was using at the time, detailed [here](https://github.com/xournalpp/xournalpp/issues/1746#issuecomment-612533873). So, instead of everywhere being dark, everywhere is light mode.

personalizedrefrigerator:
This should fix #2764 and #3278.

<strike>**Note that, at present, this PR selects the theme that matches the icon theme, even if the user has not checked/unchecked 'dark mode' in settings.** I think this is reasonable behavior -- otherwise, we end up with difficult-to-see icons, but could be seen as a breaking change. </strike> (The dark mode toggle no longer just changes the icon theme and Xournal++ now uses its own icons in place of the system's).

**Potential issues**
 * The default GTK theme must also have a dark variant installed.

**Edit:** Xournal++ now uses its own icons by default, making earlier issues mentioned here no longer relevant.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -44,11 +44,13 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control):
     g_object_ref(mainContentWidget);
     g_object_ref(sidebarWidget);
 
+    GtkSettings* appSettings = gtk_settings_get_default();
+    g_object_set(appSettings, "gtk-application-prefer-dark-theme", control->getSettings()->isDarkTheme(), NULL);
+
     loadMainCSS(gladeSearchPath, "xournalpp.css");
 
     GtkOverlay* overlay = GTK_OVERLAY(get("mainOverlay"));
     this->floatingToolbox = new FloatingToolbox(this, overlay);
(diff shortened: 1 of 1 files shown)


suisseWalter:
did you really want to include #3281 and not #3278 in your edit?

personalizedrefrigerator:
> did you really want to include #3281 and not #3278 in your edit?

Yes! Thank you for catching that! :)

Febbe:
@LittleHuba I think, this should be at least in 1.1.2. Maybe we can add it even to 1.1.1?

LittleHuba:
I would prefer 1.1.2 as 1.1.1 should be release really soon an this PR is not yet finished. Or am I mistaken?

Febbe:
Looks like it is.

rolandlo (approved):
LGTM

LittleHuba:
Okay then I guess we can merge it into 1.1.1 as well.

PR #2776, opened 2021-02-08
Layer preview fixes

mjg:
This PR solves a few issues with the sidebar, some of which are only noticable if you have a document with bookmarks:

- sidebar is not initialised properly (toolbar shows up even if it should not)
- toolbar is not reenabled properly (if it was disabled and you switch to page preview)
- toolbar tooltips refer to pages even when they act on layers

As a sidebar fix, PR #2674 could go on top of this, but because of the concerns mentioned there I'll save it for later (when I implement another sidebar mode with stacked previews).


Technius:
LGTM, thanks for the PR.

PR #2779, opened 2021-02-08
fix snapping in horizontal/vertical direction

rolandlo:
This fixes a bug in the application of the grid snapping tolerance in horizontal and vertical direction. This is experienced when resizing a selection horizontally or verticalling.


Technius:
Is this a regression or new bug fix?

rolandlo:
This was a new bug fix.

PR #2727, opened 2021-01-30
Move close button to top of sidebar

personalizedrefrigerator:
Moving the sidebar's close button to the top of the sidebar allows the user to make the sidebar smaller than previously possible, while moving the "close" button further from the delete button.

<center>

</center>

PR #2791, opened 2021-02-09
Fix regression large sidebar buttons in Adapta theme

personalizedrefrigerator:
Some themes (e.g. Adapta) draw large buttons when their container is missing the `linked` style class.

**Before**
Shadow in the wrong place and large buttons at the bottom of the sidebar

**After**
Buttons are smaller (as before) and the shadow no longer looks out-of-place

This should address https://github.com/xournalpp/xournalpp/pull/2727#issuecomment-776191629.

Note that while this decreases the size of the sidebar in some themes, adding the `linked` class slightly increases the size of the sidebar in others.
For example, with the Breeze GTK theme,

**After**
more padding between buttons

**Before**
less space between buttons

### DIFF
--- a/ui/main.glade
+++ b/ui/main.glade
@@ -2128,7 +2128,7 @@
                                     <property name="can-focus">False</property>
                                   </object>
                                   <packing>
-                                    <property name="expand">False</property>
+                                    <property name="expand">True</property>
                                     <property name="fill">True</property>
                                     <property name="position">0</property>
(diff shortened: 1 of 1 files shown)


Technius:
LGTM. Just need to fix the merge conflicts now.

personalizedrefrigerator:
> LGTM. Just need to fix the merge conflicts now.

I've done this!

Issue #2775, opened 2021-02-08
"g_unlink() failed" when trying to write LaTeX annotations


DUOLabs333:
Windows 10, Xournal 1.1.0 dev. When I'm typing up LaTeX annotations, every second or so, it says "Could not save .tex file: [path] could not be saved. g_unlink() failed. Permission denied". The annotation does not even load in the preview. Running in administrator mode does not work. I did not keep the folder in "Program Files", if that matters.

rolandlo:
I can reproduce the problem with the nightly from Feb 08, 2021 on Windows. It doesn't happen with the nightly from the day before (Feb 07). The regression must have been introduced in #2728. 
@redweasel Do you have a Windows testing environment?

redweasel:
uh... @rolandlo I could setup a windows testing environment. But lets not conclude to fast. On first glance this does not seem to come from my commit.

@DUOLabs333 what version of Xournal++ do you use exactly? (you can check that in the "help > about" dialog)

DUOLabs333:
@redweasel I am using 1.1.0+ dev version that was built today at 00:18:29

redweasel:
Ok thanks. I assume your compiled commit is 6c0aa8332c7da94d9336ae247c54384ceb93cee3. I will investigate a little.

redweasel:
This is the relevant code which sent the message is in `LatexGenerator.cpp`.
@DUOLabs333 Can you check your console output for warnings? Because according to the following code in `PathUtil.cpp` it should have shown a Message saying "Could not create folder ..." and wrote a warning into the console if the folder which is used to output the LaTeX can not be created.

If that is not the case then please check if there is a Temporary folder for xournalpp with a file called "tex.tex" in it. As I do not use Windows, I don't know where that folder is though, so a quick google might help. I also don't know which programs are allowed to use the Temporary Folder in Windows, which might be relevant, as this is seemingly a permission issue.

Also, which version did work for you? What did you change except the version?

```c++

auto Util::getTmpDirSubfolder(const fs::path& subfolder) -> fs::path {
    auto p = fs::u8path(g_get_tmp_dir());
    p /= FS(_F("xournalpp-{1}") % Util::getPid());
    p /= subfolder;
    return Util::ensureFolderExists(p);
}

auto Util::ensureFolderExists(const fs::path& p) -> fs::path {
    try {
        fs::create_directories(p);
    } catch (fs::filesystem_error const& fe) {
        Util::execInUiThread([=]() {
[...]

PR #2798, opened 2021-02-11
Fix LaTeX dialog threading for Windows

redweasel:
fixes https://github.com/xournalpp/xournalpp/issues/2775

I was debating if I should let the current (outdated) render finish or, if I should abort it to start the updated render. In the end, after some testing, letting the outdated render finish feels smoother for the user. That also means there is this simple fix.


Technius:
LGTM. Tested working on Windows.

PR #2804, opened 2021-02-13
Remove unnecessary copy of stroke data in Stroke::freeUnusedPointItems()

bhennion:
`Stroke::freeUnusedPointItems` copies the strokes points to a new location. Using `std::vector::shrink_to_fit()` instead is (way) cheaper with the same memory footprint shrinking.

The line `this->points = {begin(this->points), end(this->points)};` was introduced with commit
> commit 742499a20b99bc1d6d103c7a16376c4a2ef053a8
> Author: Fabian Keßler <fabian_kessler@gmx.de>
> Date:   Fri Oct 11 09:41:38 2019 +0200
>
>     Fixed undefined behaviour
>     by replacing dynamic allocated and resized c-array with a std::vector

@Febbe: was there a specific reason for using `this->points = {begin(this->points), end(this->points)};` and not `std::vector::shrink_to_fit()`?

### DIFF
--- a/src/model/Stroke.cpp
+++ b/src/model/Stroke.cpp
@@ -158,7 +158,13 @@ auto Stroke::getPoint(int index) const -> Point {
 
 auto Stroke::getPoints() const -> const Point* { return this->points.data(); }
 
-void Stroke::freeUnusedPointItems() { this->points = {begin(this->points), end(this->points)}; }
+void Stroke::freeUnusedPointItems() {
+    size_t capacity = this->points.capacity();
+    // Reduce the vector's memory footprint if it has to much unused memory
+    if (capacity > 1024 && 4 * this->points.size() < 3 * capacity) {
+        this->points = {this->points.begin(), this->points.end()};
+    }
+}
 
 void Stroke::setToolType(StrokeTool type) { this->toolType = type; }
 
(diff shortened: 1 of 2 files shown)


Febbe:
Yes there indeed is. Shrink to fit is implementation defined and will mostly do nothing. This is not the behavior the original author intended with this function. In my opinion it should be never called btw.

bhennion:
> Yes there indeed is. Shrink to fit is implementation defined and will mostly do nothing. This is not the behavior the original author intended with this function. In my opinion it should be never called btw.

Isn't the original capacity of a `vector` also implementation specific? If so, there is no guaranty that `this->points = {begin(this->points), end(this->points)};` actually does what it's supposed to here either.

> You can read more here:
> https://stackoverflow.com/questions/23502291/is-shrink-to-fit-the-proper-way-of-reducing-the-capacity-a-stdvector-to-its

Thanks for the link. Instructive indeed!

Febbe:
The default size of the default constructor is. However, when initializing the vector with a specific size or range it does exactly what you request. Maybe with a small offset.

The problem is more the filling of a vector. Starting with a size of 10 it increases in size with a specific heuristic value.

10 > 20 > 40 > 80 > 160 ... 512 > 1024 (f=2) this will invoke a copy for each increase and the size may be way to big. 

You could improve this function by using my code, but doing this size change only for large vectors with a large unused space. Maybe from a stroke count of 1024 shrink every vector, which has more than 25% unused space.

bhennion:
Done.

bhennion:
Ok, with @Febbe's explanation and all, I don't think this is worth merging anymore.

Issue #2628, opened 2021-01-07
Mouse buttons in xournal++ don't work after pressing a pen button


GuentherE:
Affected versions: 1.0.20 and 1.1.0-nightly-20210106
OS: Windows 10

Hi,

Thanks for this great software. I like it very much. I use it under Ubuntu 18.04 with no problems with my XP-Pen Artist 12 Pro tablet. 

Now I have a problem as a customer project forces me to use Windows 10. Installation was flawless and using the tablet with my pen and my mouse works also. But after I have pressed one of the pen buttons my mouse stops working: I can use it in other programs but in xournal++ pressing the mouse buttons has no effect. After restarting xournal++ the mouse is working again.

As a workaround I have tried to disable the pen buttons and use the shortcut buttons on the tablet for switching between eraser and highlighter, but this is also not working. It seems that the signal of the pen buttons is influenicing the mouse behaviour even there is no action assigned.

Using the same szenario with e.g. MS whiteboard works.

Is this a configuration problem or a bug?

TIA,
Guenther

personalizedrefrigerator:
Thank you for reporting this! When broken, does the mouse open/close the dropdown menus at the top of the screen? Are you using the new input system?

GuentherE:
I can open/close the dropdown menus of other programs but can do nothing in xournal++. Even closing the app is not possible.

personalizedrefrigerator:
Try running Xournal++ from `cmd.exe` or similar. Some warnings/messages should be logged in its output.

GuentherE:
Unfortunately no messages when started from command line. Same behaviour.

The mouse cursor vanishes when entering xournal++ and is a small dot when in the title bar area. Outside xournal++ it shows the normal cursor.

personalizedrefrigerator:
@rolandlo Could you look at this? I don't have access to a windows machine with pen input. Thank you!

personalizedrefrigerator:
@GuentherE Can you post the version of libgtk Xournal++ is using (see Help > About).

GuentherE:
Version 1.0.20 with libgtk 3.24.23

GuentherE:
I've tested it with 1.0.18 and 1.0.12: same effect.

rolandlo:
Some comments: 
- Under Windows there are only warnings/messages when building from source (as described in the [wiki](https://github.com/xournalpp/xournalpp/blob/master/readme/WindowsBuild.md)).
[...]

PR #2805, opened 2021-02-13
Fix Selection Tool in New Input System

### LINKED ISSUE #2539: Regression: Selection Tool not working with Styuls Button 2
**Affects versions :**
 - OS: Arch
 - Which version of libgtk do you use: libgtk: 3.24.23 
 - Version of Xournal++: Xournal++ 1.1.0+dev 
 - Installation method: build from 

**Describe the bug**
Lasso tool does no longer work when activated via Stylus Button 2:
image

**To Reproduce**
1. Set Stylus Button 2 to Select Region
2. Save Settings
3. Try to use Button 2 from Stylus

**Expected behavior**
Selected Region is shown while drawing and contained strokes are selected after completion.

**Screenshots of Problem**
No selected region is shown and nothing can be selected.
Peek 2020-12-13 17-31

**Additional context**
It seems that 925acf9eefce1775bb3b174d2742d620243a3412 introduced this regression. @Febbe could you please check where this could come from?

idotobi:
This PR fixes #2539.

Previously only `event.button == 1` was handled in the new input system,
but in the old system all non-zero values where considered.
This lead to drawing events not being handled if a button was pressed
in the new input system.
This is fixed in this commit by handling any `event.button > 0`.

### DIFF
--- a/src/gui/inputdevices/StylusInputHandler.cpp
+++ b/src/gui/inputdevices/StylusInputHandler.cpp
@@ -25,8 +25,7 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
 
     // Trigger start of action when pen/mouse is pressed
     if (event.type == BUTTON_PRESS_EVENT) {
-
-        if (event.button == 1 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
+        if (event.button > 0 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
             this->eventsToIgnore = this->inputContext->getSettings()->getIgnoredStylusEvents();
             if (this->eventsToIgnore > 0) {
                 this->eventsToIgnore--;  // This is already the first ignored event
@@ -90,7 +89,7 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
 
     // Trigger end of action if pen tip leaves screen or mouse button is released
     if (event.type == BUTTON_RELEASE_EVENT) {
-        if (event.button == 1 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
+        if (event.button > 0 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
             if (this->eventsToIgnore < 0) {
                 this->actionEnd(event);
             } else {


idotobi:
@rolandlo : I added this to v1.1.0release Milestone as this would be a regression if not fixed before the old input system is turned of.

rolandlo:
> @rolandlo : I added this to v1.1.0release Milestone as this would be a regression if not fixed before the old input system is turned of.
I agree with that. 
Have you tested your PR on Windows as well? Does it change anything with respect to #2628? Otherwise I will make a quick test.

Technius:
Sidenote: these magic constants for button need to be refactored.

idotobi:
> Have you tested your PR on Windows as well? Does it change anything with respect to #2628? Otherwise I will make a quick test.

No I couldn't test it on windows, as I don't have a windows machine :sweat_smile: . If you could do it that would be great.

---

**Regarding #2628**

 I don't think this will fix it completely. I just checked with current master:
If I keep pressing the stylus button that did not work prior to the fix then clicking on anything in the UI of xournalpp did not show any reaction. Hence it could be related.
However, I could still close the app (though that might be special for Linux), and the problem did not persist when I didn't press the Stylus button anymore. So there is probably an additional root cause to this.

You could check if it is linked by putting a debug flag at https://github.com/xournalpp/xournalpp/blob/891c972c27a335113f8a424ae7d3adcbc1ed3f18/src/gui/inputdevices/StylusInputHandler.cpp#L29 or simply printing the `event.button` value to console right before the `if`. If `event.button` gets stuck on a non `1` value than this is part of the cause. In this case, however, we should also find out why it gets stuck as this would cause additional problems.

idotobi:
> Sidenote: these magic constants for button need to be refactored.

Agreed. Best is to create a followup issue for cleaning that up a bit. I can write the issue.
Would you suggest to use enums instead of the magic constants or is there an even better way?

rolandlo:
I have trouble testing the PR on Windows, since currently no stylus buttons work at all (with or without PR). Maybe I have messed up the Windows pen settings.
I have tested the PR on Ubuntu in an 
- X11 session (where my non-eraser stylus button corresponds to the middle mouse button, i.e. GdkEvent Button 2) and in a 
- Wayland session (where my non-eraser stylus button corresponds to the right mouse button, i.e. GdkEvent Button 3). 

When I assign the selection tool to stylus button 2 (i.e. GdkEvent Button 3) in the preferences in a Wayland session, the selection tool will not only be activated if I press the button and touch the screen, but also when I press the button and just hover. This behavior should only occur with `TPCButtonEnabled`.
In conclusion I think this PR would introduce a bug rather than solve a problem.

idotobi:
@rolandlo : You're right :see_no_evil: . I wasn't aware of the TPCButton configuration setting. Luckily @tattsan pointed me to this https://github.com/xournalpp/xournalpp/issues/2539#issuecomment-779197877.

I'll close the PR then. However, I think it might be good to document this both for users and developers. I'll continue the discussion on that in the linked issue.

Issue #2526, opened 2020-12-10
Closing then reopening file removes all line breaks in "edit LaTeX" mode


DUOLabs333:
I'm using Windows 10 and Xournal 1.1.0. When I add LaTeX annotations, I usually add line breaks to the input (though not the actual output), so when I'm typing, I don't have to edit one run-on line. However, when I close the file, then reopen it, all those line breaks are removed, so I have to put them back in all over again. I can't tell if this is expected behavior or not though.

rolandlo:
I can reproduce it on Linux as well. Looks like a (minor) bug.

DUOLabs333:
@rolandlo Oh, ok. I thought this was one of those "it's not a bug, it's a feature" type of things.

PR #1703, opened 2020-01-12
Add Fedora

luyatshimbalanga:

PR #2809, opened 2021-02-15
LaTeX Dialog: Use GtkSourceView and autoresize preview

### LINKED ISSUE #2542: Ctrl+Z doesn't undo in "Add/Edit TeX"
I use Windows 10, with version 1.1.0. When adding or editing TeX on a document, I tried to do Ctrl+Z to undo (and similarly, Ctrl+Shift+Z to redo) in the text box, but nothing happened.

personalizedrefrigerator:
The LaTeX dialog now has a command output tab and can use a GtkSourceView if compiled in

## Summary
 * If either of the`gtksourceview-3` or `gtksourceview-4` libraries are available at build time, a GtkSourceView is used to provide syntax highlighting for <math>LaTeX</math> a user has entered.
   * Introduces settings (line numbers, syntax highlight, etc.) for this.
       * Settings are added to Settings Dialog/LaTeX
 * ~~Allows multi-line LaTeX entry.~~ Changes `OK` shortcut to `Ctrl+Enter`.
 * Wraps the `GtkTextView` and the TeX preview in a `GtkPaned` and resizes the preview, as appropriate.
     * This is **very** useful for writing large paragraphs of formatted text.
     * This is **very** useful for writing long pieces of formatted text
 * Adds a "Command Output" tab to the LaTeX dialog.
 * Changes the TeX preview background to black when the default TeX color is light (this is, and was, before this PR, based on the color of the text tool).
     * Light TeX is now visible in the TeX preview.

## To-do
 * [x] Move dependency changes to a separate commit.
 * [x] Update Linux installation readme (new (optional) dependency).
 * [x] Update Debian/RPM package configuration.
 * [x] Move SourceView theme setting to `LatexSettings.h/.cpp` from global settings.
 * [x] Separate section in Linux build README for CentOS (and RHEL?) with `libgtksourceview3-devel` dependency.
 * [x] Change `gtksourceview4` dependency for OpenSUSE to `gtksourceview4-devel`

## Need review
 * [x] 46dda4e507e74ee499f2708797b3e94478dc21e4 needs verification 
     -  [x] changes to deb 
     -  [x] and rpm dependencies
 * [x] af342f17346e252817f17c3021d6d228234b7d26 may have errors
     - [x] Check Fedora`gtksourceview4-devel` dependency
     - [x] Check CentOS `gtksourceview3-devel` dependency.
     - [x] Check RHEL `gtksourceview3-devel` dependency (should be the same as CentOS)
     - [x] Check Arch `gtksourceview4` dependency
     - [x] Check Ubuntu/Debian `libgtksourceview-4-dev` dependency
     - [x] Check Raspberry Pi OS `libgtksourceview-4-dev` dependency
     - [x] Check openSUSE `gtksourceview` dependency
     - [x] Check Solus `libgtksourceview-devel` dependency
 
 I think the above changes are correct (and, as such, marked as resolved), but a tripple-check might be worth while!

When built with GtkSourceView, this should fix #2542.


rolandlo:
Looks great! One question about "Allows multi-line LaTeX entry."
Is this now possible with Enter? This was realized before with Shift+Enter.

personalizedrefrigerator:
> Looks great! One question about "Allows multi-line LaTeX entry."
> Is this now possible with Enter? This was realized before with Shift+Enter.

This is done with Enter! I was unaware of Shift+Enter!

personalizedrefrigerator:
At least with the `Adwaita-Dark` GTK theme and default SourceView theme, comments are rather difficult to see:
Dark blue comments on a dark gray background

This could be fixed by either adding a setting to set the SourceView style (perhaps using [StyleChooserButton or StyleChooserWidget](https://developer.gnome.org/gtksourceview/3.24/GtkSourceStyleSchemeChooserWidget.html)).

Technius:
This is too major of a change (new dependency) to include in 1.1.0.

personalizedrefrigerator:
> This is too major of a change (new dependency) to include in 1.1.0.

Xournal++ will still build without the `GtkSourceView` dependency. If unable to find `gtksourceview-3` or `gtksourceview-4`, it uses a `TextView`, as it did before.

src/gui/dialog/LatexSettingsPanel.cpp
```diff
@@ -4,19 +4,39 @@
...
+#ifdef HAVE_GTK_SOURCEVIEW_LIB
+#include <gtksourceview/gtksource.h>
+#endif
+
+LatexSettingsPanel::LatexSettingsPanel(GladeSearchpath* gladeSearchPath, Settings* settings):
         GladeGui(gladeSearchPath, "latexSettings.glade", "latexSettingsPanel"),
         cbAutoDepCheck(GTK_TOGGLE_BUTTON(this->get("latexSettingsRunCheck"))),
-        globalTemplateChooser(GTK_FILE_CHOOSER(this->get("latexSettingsTemplateFile"))) {
+        globalTemplateChooser(GTK_FILE_CHOOSER(this->get("latexSettingsTemplateFile"))),
+        globalSettings(settings) {
```
personalizedrefrigerator, line 21: I don't like this... `LatexSettingsPanel` should only know about `LaTeX`-specific settings. 

Although it makes sense to include the `SourceView` settings in the `LatexSettingsPanel` (as that's the only place it's used), this should probably be changed.

Technius: Let's keep it simple for now and have it in the `LatexSettingsPanel`. We can make it global when we need it.

Technius:
I haven't reviewed the actual code yet but some general comments:
* The dependency configuration changes should be split out into a separate commit.
* Also, the Linux installation readme needs to be updated to account for the new dependency. Same with Windows (I think it should have an msys package available). Mac build instructions are being changed in another PR, so we can hold off on Mac for now.
* Debian / RPM package configuration also needs to be updated.
* > At least with the Adwaita-Dark GTK theme and default SourceView theme, comments are rather difficult to see:
  
  Maybe we could set a default background color (e.g. white) to avoid this problem.

Yeah, dependency updates are a headache, which is why I try to avoid them right before a major update.

personalizedrefrigerator:
> I haven't reviewed the actual code yet but some general comments:
> 
>     * The dependency configuration changes should be split out into a separate commit.
> 
>     * Also, the Linux installation readme needs to be updated to account for the new dependency. Same with Windows (I think it should have an msys package available). Mac build instructions are being changed in another PR, so we can hold off on Mac for now.
> 
>     * Debian / RPM package configuration also needs to be updated.
> 
>     * > At least with the Adwaita-Dark GTK theme and default SourceView theme, comments are rather difficult to see:
>       
>       
>       Maybe we could set a default background color (e.g. white) to avoid this problem.
> 
> 
> Yeah, dependency updates are a headache, which is why I try to avoid them right before a major update.

Converting to a draft until the above are done.

src/gui/dialog/LatexSettingsPanel.cpp
personalizedrefrigerator: Done!

personalizedrefrigerator:
Please review commits https://github.com/xournalpp/xournalpp/pull/2809/commits/94447be126cc61c7be9630d50449a90d748a1a9c and https://github.com/xournalpp/xournalpp/pull/2809/commits/4631db3ba1687b0aa4725eb2658176bfd0eb2bbe! These probably have issues as they update dependencies for Linux distributions I have not tested with!
 * **Most of these rpm/deb/README changes are untested.**

readme/LinuxBuild.md
```diff
@@ -24,34 +24,36 @@ The minimum required CMake version is 3.10, but we recommend to use >=3.15.
...
+gtksourceview-devel
 ```
 
 #### For Solus:
 ```bash
 sudo eopkg it -c system.devel
 sudo eopkg it cmake libgtk-3-devel libxml2-devel poppler-devel libzip-devel \
 portaudio-devel libsndfile-devel alsa-lib-devel cppunit-devel lua-devel \
-librsvg-devel gettext
+librsvg-devel gettext gtksourceview-devel
```
personalizedrefrigerator, line 56: Please verify

rpm/fedora/xournalpp.spec
```diff
@@ -34,6 +34,7 @@ BuildRequires:  pkgconfig(lua)
 BuildRequires:  pkgconfig(poppler-glib)
 BuildRequires:  pkgconfig(portaudiocpp) >= 12
 BuildRequires:  pkgconfig(sndfile)
+BuildRequires:  pkgconfig(gtksourceview) >= 3.18
```
personalizedrefrigerator, line 37: This might also be wrong -- it might need to be, instead `BuildRequires:  pkgconfig(gtksourceview4)`...

debian/control
```diff
@@ -3,12 +3,12 @@ Maintainer: Andreas Butti <andreasbutti@gmail.com>
...
 Priority: optional
 Standards-Version: 3.9.2
-Build-Depends: debhelper (>= 9), cmake, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile-dev, liblua5.3-dev, libzip-dev, librsvg2-dev
+Build-Depends: debhelper (>= 9), cmake, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile-dev, liblua5.3-dev, libgtksourceview-4-dev | libgtksourceview-3.0-dev, libzip-dev, librsvg2-dev
 Homepage: https://github.com/xournalpp/xournalpp/
 
 Package: xournalpp
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2-2 (>= 2.40)
+Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libgtksourceview-4-0 | libgtksourceview-3.0-1 (>= 3.18.0), libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2-2 (>= 2.40)
```
personalizedrefrigerator, line 11: I think this is fine, but it still needs to be checked...

rolandlo: - The changes to `debian/control` look fine. I can install those packages on Ubuntu 20.10 and I have checked that the alternative you provide work on older Ubuntu versions as well.

- In #2814 I got the comment by @Technius: 
> I would be careful about updating the rpm file, as it's used for the Fedora equivalent of PPA: #1703 (comment)
I think this applies here as well.

rolandlo:
From [this overview](https://pkgs.org/search/?q=gtksourceview) it looks like CentOS doesn't have the `gtksourceview4-devel` dependency (Fedora does).

personalizedrefrigerator:
> From [this overview](https://pkgs.org/search/?q=gtksourceview) it looks like CentOS doesn't have the `gtksourceview4-devel` dependency (Fedora does).

Okay... It seems to only have `gtksourceview3-devel`, and `gtksourceview3`...

rolandlo:
In OpenSuse the package seems to be named `gtksourceview4`, see [this link](https://build.opensuse.org/search?search_text=gtksourceview4&search_for=0&name=1&attrib_type_id=)

In Solus it is named `libgtksourceview-devel`, see [this link](https://solus.pkgs.org/rolling/solus-shannon-x86_64/libgtksourceview-devel-4.8.0-23-1-x86_64.eopkg.html)

The info about Arch looks OK. For RHEL I don't know.

tattsan:
I use Arch Linux. I installed gtksourceview4 package, and could build the PR fine.

TikZ sometimes draw HUUUUUGE curves, so I'm very gratefull for the autoresize feature.

rpm/fedora/xournalpp.spec
personalizedrefrigerator: I can try to test the modified `rpm` over the weekend.

personalizedrefrigerator: Changed to `pkgconfig(gtksourceview-4)`

personalizedrefrigerator:
image

It looks like it's `gtksourceview4-devel` on OpenSUSE Tumbleweed...

tattsan:
The font setting is great!  12pt was too small for my 4K monitor, but I didn't want to increase the font size except in the LaTeX window.

About autoresize feature:
In the vertical direction, the preview is resized to the full height of preview pane.  But in the horizontal direction,  there seems to exist another upper limit. It is not resized to the full width of preview pane if the pane has sufficient width.  Can it be resized to the pane width?

Technius:
> In the vertical direction, the preview can be resized to the full height of preview pane. But in the horizontal direction, there seems to exist another upper limit. It can not be resized to the full width of preview pane if the pane has sufficient width. Can it be resized to the pane width?

Better than this is to compute both scale factors and then set the actual scale factor used to the smaller of the two. This will accommodate both vertical and horizontal scaling, so that your best choice is to resize the dialog to fit the aspect ratio of the image.

tattsan:
>Better than this is to compute both scale factors and then set the actual scale factor used to the smaller of the two. This will accommodate both vertical and horizontal scaling, so that your best choice is to resize the dialog to fit the aspect ratio of the image.

I agre, that's what I meant too.

personalizedrefrigerator:
> In the vertical direction, the preview can be resized to the full height of preview pane. But in the horizontal direction, there seems to exist another upper limit.

The upper limit applies to both directions (i.e. it only shrinks the preview to fit). I'll try to remove this.

> The font setting is great! 12pt was too small for my 4K monitor, but I didn't want to increase the font size except in the LaTeX window.

Oops! It wasn't using `12pt`! It was using `12px`! Hopefully, in GTK, `12pt` is dpi-aware... Regardless, there should be an option to use the font from the GTK theme.

Thank you for the feedback!

tattsan:
>(i.e. it only shrinks the preview to fit).

I've made the LaTeX window fullscreen. Then the preview was shown as described above.

> Oops! It wasn't using 12pt!

 :sweat_smile:

personalizedrefrigerator:
The font issue should be fixed, I haven't done anything to the preview yet.

personalizedrefrigerator:
After saving and re-opening a document with LaTeX, all formulas are condensed into one line. This should probably be fixed...

For example,
**before**
multiple lines of TeX

**after**
single line of TeX

after a save and re-open.

rolandlo:
Yes, this should be fixed. the issue has already been reported in #2526

tattsan:
I was mistaken. This feature only shrinks the image that is too large, it does not enlarge the image if the preview window is large. This is reasonable.

personalizedrefrigerator:
I think the issue is that we're storing the text as an attribute, which probably needs line breaks escaped...

This is how normal text is stored, for comparison:
normal text, a value, not an attribute

personalizedrefrigerator:
> it does not enlarge the image if the preview window is large.

I think it would be nice if it made the preview content larger for a larger preview. :)

personalizedrefrigerator:
### Status Update

Currently (in xournalpp:master), the TeX color is based on the text tool color.
A recent commit changes the background color to black when a light TeX color is used:
**Before**
TeX is invisible -- background is always light

**After**
TeX is visible on a dark background

personalizedrefrigerator:
@rolando @Technius  I would love to have this be a part of Xournal++. It adds a dependency, though (so it may be best to wait to merge it if we're releasing a minor version soon). When should we merge this?

LittleHuba:
Is this ready for a review/merge?

Technius:
This is ready for review, but I would prefer holding off on such a large change until after 1.1.1 is out.

LittleHuba:
We can merge this for the next minor release (1.2.0) right after the review. This will give nightly users time to test it out.

Patch releases 1.1.x are not blocked by such a merge.

personalizedrefrigerator:
@bhennion @idotobi  @rolandlo I would like to merge this soon (if possible). Please review this if you get time!

personalizedrefrigerator:
## A quick note on the MacOS pipelines

This PR adds an (optional) dependency on `gtksourceview-3-dev` or `gtksourceview-4-dev`. I've added this dependency to the builds for Windows and Linux, but I'm unsure how to add it to the build process for MacOS. Help with this would be appreciated!

rolandlo:
> ## A quick note on the MacOS pipelines
> 
> This PR adds an (optional) dependency on `gtksourceview-3-dev` or `gtksourceview-4-dev`. I've added this dependency to the builds for Windows and Linux, but I'm unsure how to add it to the (automated) build process for MacOS. Help with this would be appreciated!

There are two ways: 
1. build `gtksourceview` from source during the build process (in `build-mac.yml`)
2. use `jhbuild` to build `gtksourceview` in advance and add it to the Gtk blob that @Febbe uploaded to the [xournalpp-pipeline-dependencies repo](https://github.com/xournalpp/xournalpp-pipeline-dependencies/raw/macos-10.15-jhbuild)

The second way would be preferable, if the build takes longer than a few seconds.

On the Gnome repo for [gtksourceview-4-8](https://gitlab.gnome.org/GNOME/gtksourceview/-/tree/gtksourceview-4-8) it's written that to build the latest version of `GtkSourceView` plus its dependencies from Git, `Jhbuild` is recommended. So maybe in any case we must use the second way.

bhennion:
This looks great! I haven't been through the code yet, but I do have a resquest:

When compiling with `gtksourceview3`, but `gtksourceview` is not installed (anymore) upon execution, we get
```
./install/bin/xournalpp: error while loading shared libraries: libgtksourceview-3.0.so.1: cannot open shared object file: No such file or directory
```
and Xournal++ never launches.

Not having the library at runtime should not be blocking. I'm thinking of downstream package maintainer and optional dependencies: the package is compiled with every thing in it, but run on various machine that may not have the optional library.

Febbe:
First: Don't introduce optional libs (only when platform dependent). Either we add that feature or we don't.
Instead of linking dynamically, there is another solution; uncommon utility libs could be added and linked statically. This will increase binary size, but does not force the user to install every dep. Also, on Windows and mac, those need to be shipped anyway.

Febbe:
Optional platform independent libs are only useful, when: 
 - They are called over a system call (completely independent executable)
 - Or linked at runtime via dlopen(). Functions must be called via their address.
They are not useful: 
 - When they depend on the building system. (via macros etc.)

bhennion:
@febbe I'm confused as to what you opinion is.

If I understand correctly what you're saying, there are two sensible options:

1. We should decide whether we rely on `gtksourceview3` or `gtksourceview4` and make it a mandatory build and runtime dependency.
2. We should implement a runtime dependency load (i.e. use `dlopen` or such). If this fails, then we should disable the functionality.

Correct?

I would say that overall, 2. is preferable, but it is of course more work (especially as `dlopen` is not totally portable). See #3262 for a comment in this direction.
Of course, in the specific case of this PR, the library is small (<1MB), but IMO some other rather specific bigger dependencies would benefit from being made optional (e.g. `portaudio`, `tex`, ...). Anyway, this is work for another time.

Febbe:
Of cause point 2 would be top notch, but the opportunity cost would be to high to use this. Therefore, stick with 1.

personalizedrefrigerator:
> @Febbe I'm confused as to what you opinion is.
> 
> If I understand correctly what you're saying, there are two sensible options:
> 
>     1. We should decide whether we rely on `gtksourceview3` or `gtksourceview4` and make it a mandatory build and runtime dependency.
> 
>     2. We should implement a runtime dependency load (i.e. use `dlopen` or such). If this fails, then we should disable the functionality.
> 
> 
> Correct?
> 
> I would say that overall, 2. is preferable, but it is of course more work (especially as `dlopen` is not totally portable). See #3262 for a comment in this direction. Of course, in the specific case of this PR, the library is small (<1MB), but IMO some other rather specific bigger dependencies would benefit from being made optional (e.g. `portaudio`, `tex`, ...). Anyway, this is work for another time.

If we want to support Ubuntu 18.04, we need to choose `gtksourceview3`. Otherwise, we can choose `gtksourceview4`.
For reference, here's a link to the `gtksourceview` library: https://gitlab.gnome.org/GNOME/gtksourceview .

Febbe:
Do they have different ABIs?

rolandlo:
It looks like some rebasing errors have been introduced and the PR doesn't build anymore. The `add_includes_ldflags` macro does not exist anymore, so it can't be used for `gtksourceview`. 
@personalizedrefrigerator Could you please make it build again?

rolandlo:
Building gtksourceview 4.8 on MacOS via `JhBuild` turned out to be no problem at all. Simply run
```
jhbuild build gtksourceview3
``` 
in a jhbuild shell. So we can bundle it when we upload an updated binary blob for MacOS. Ideally we do it in combination with Gtk `3.24.31`, which not only solves the issues on MacOS Monterey, but should also boost performance on MacOS BigSur. It will probably take a week or two until `gtk-osx` is updated to Gtk 3.24.31. The update to 3.24.30 happened on 2021, July 26 (in [this commit](https://gitlab.gnome.org/GNOME/gtk-osx/-/commit/02cc20cfe26b1641c15f0c57d2c1c86e2b5b69d6)), which is 18 days after the Gtk 3.24.30 release

rolandlo:
@personalizedrefrigerator Any plans on making this PR build again in the coming weeks? Soon we will update the MacOS binary blob (with Gtk 3.24.31) and may include `gtksourceview` into the blob. It would be important to test your PR to check if it all works well.

personalizedrefrigerator:
> @personalizedrefrigerator Any plans on making this PR build again in the coming weeks? Soon we will update the MacOS binary blob (with Gtk 3.24.31) and may include `gtksourceview` into the blob. It would be important to test your PR to check if it all works well.

I'm relatively busy right now. I'll do so if I have time!

rolandlo:
@personalizedrefrigerator FYI: `gtksourceview` 4.8.1 is now part of the current Gtk blob for MacOS.

azure-pipelines/dependencies/msys2-blob.yml
```diff
@@ -21,6 +21,7 @@ steps:
     pacman --noconfirm -Syuu
     pacman --noconfirm -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-make mingw-w64-x86_64-ninja patch
     pacman --noconfirm -S mingw-w64-x86_64-poppler mingw-w64-x86_64-gtk3 mingw-w64-x86_64-libsndfile mingw-w64-x86_64-libzip
+    pacman --noconfirm -S mingw-w64-gtksourceview4
```
rolandlo, line 24: This should be mingw-w64-x86_64-gtksourceview4, I think

rolandlo:
On Ubuntu 21.10 it works fine. On MacOS there are somehow no styles to choose from
Screenshot 2022-02-19 at 20 13 53
This results in crashes when closing the preferences window or when testing the LaTeX configuration (by pressing the corresponding button). Any idea what is the problem?

```shell
(xournalpp-bin:13487): GtkSourceView-CRITICAL **: 20:21:51.302: gtk_source_style_scheme_chooser_set_style_scheme: assertion 'GTK_SOURCE_IS_STYLE_SCHEME (scheme)' failed

(xournalpp-bin:13487): GtkSourceView-CRITICAL **: 20:21:53.691: gtk_source_style_scheme_get_id: assertion 'GTK_SOURCE_IS_STYLE_SCHEME (scheme)' failed

** (xournalpp-bin:13487): WARNING **: 20:21:53.691: [Crash Handler] Crashed with signal 11
```   
and the crashlog:
```
[bt] #1 1   xournalpp-bin                       0x0000000103e7aef2 _ZL12crashHandleri + 1874
[bt] #2 2   libsystem_platform.dylib            0x00007fff6b68d5fd _sigtramp + 29
[bt] #3 3   ???                                 0x00007f95a0ce1c50 0x0 + 140280624716880
[bt] #4 4   libc++.1.dylib                      0x00007fff687955ef _ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE6assignEPKc + 21
[bt] #5 5   xournalpp-bin                       0x00000001040374fd _ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEaSEPKc + 29
[bt] #6 6   xournalpp-bin                       0x0000000104037089 _ZN18LatexSettingsPanel4saveER13LatexSettings + 361
[bt] #7 7   xournalpp-bin                       0x0000000104037577 _ZN18LatexSettingsPanel9checkDepsEv + 87
[bt] #8 8   xournalpp-bin                       0x000000010403820d _ZZN18LatexSettingsPanelC1EP15GladeSearchpathENK3$_0clEP10_GtkWidgetPS_ + 29
[bt] #9 9   xournalpp-bin                       0x00000001040381dd _ZZN18LatexSettingsPanelC1EP15GladeSearchpathEN3$_08__invokeEP10_GtkWidgetPS_ + 29
[bt] #10 10  libgobject-2.0.0.dylib              0x0000000106c3256d _g_closure_invoke_va + 205
[bt] #11 11  libgobject-2.0.0.dylib              0x0000000106c4943b g_signal_emit_valist + 1243
[bt] #12 12  libgobject-2.0.0.dylib              0x0000000106c49ef2 g_signal_emit + 130
[bt] #13 13  libgtk-3.0.dylib                    0x0000000105807022 gtk_real_button_released + 274
[bt] #14 14  libgobject-2.0.0.dylib              0x0000000106c3256d _g_closure_invoke_va + 205
[bt] #15 15  libgobject-2.0.0.dylib              0x0000000106c4943b g_signal_emit_valist + 1243
[bt] #16 16  libgobject-2.0.0.dylib              0x0000000106c49ef2 g_signal_emit + 130
[bt] #17 17  libgtk-3.0.dylib                    0x00000001058076f8 multipress_released_cb + 104
[bt] #18 18  libgtk-3.0.dylib                    0x00000001057bb5fa _gtk_marshal_VOID__INT_DOUBLE_DOUBLEv + 202
[bt] #19 19  libgobject-2.0.0.dylib              0x0000000106c3256d _g_closure_invoke_va + 205
[bt] #20 20  libgobject-2.0.0.dylib              0x0000000106c4943b g_signal_emit_valist + 1243
[bt] #21 21  libgobject-2.0.0.dylib              0x0000000106c49ef2 g_signal_emit + 130
[bt] #22 22  libgtk-3.0.dylib                    0x00000001058f51b7 gtk_gesture_multi_press_end + 199
[bt] #23 23  libgobject-2.0.0.dylib              0x0000000106c35be1 g_cclosure_marshal_VOID__BOXEDv + 177
[bt] #24 24  libgobject-2.0.0.dylib              0x0000000106c3256d _g_closure_invoke_va + 205
[bt] #25 25  libgobject-2.0.0.dylib              0x0000000106c4943b g_signal_emit_valist + 1243
[bt] #26 26  libgobject-2.0.0.dylib              0x0000000106c49ef2 g_signal_emit + 130
[bt] #27 27  libgtk-3.0.dylib                    0x00000001058f2a34 gtk_gesture_handle_event + 756
[bt] #28 28  libgtk-3.0.dylib                    0x00000001058f72ea gtk_gesture_single_handle_event + 618
[bt] #29 29  libgtk-3.0.dylib                    0x00000001058ae8ca gtk_event_controller_handle_event + 154
[bt] #30 30  libgtk-3.0.dylib                    0x0000000105ac996c _gtk_widget_run_controllers + 140
[bt] #31 31  libgtk-3.0.dylib                    0x00000001057b53c5 _gtk_marshal_BOOLEAN__BOXEDv + 181
```

personalizedrefrigerator:
> On Ubuntu 21.10 it works fine. On MacOS there are somehow no styles to choose from Screenshot 2022-02-19 at 20 13 53 This results in crashes when closing the preferences window or when testing the LaTeX configuration (by pressing the corresponding button). Any idea what is the problem?
> 
> ...

Thank you for testing it! It looks like we might need to determine where themes are stored on MacOS and call [gtk_source_style_scheme_manager_append_search_path](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/method.StyleSchemeManager.append_search_path.html).

Additionally, [this website](http://scribes.sourceforge.net/themegenerator.php) should be able to generate `GtkSourceView`-compatible themes. As such, it would be easy to create a `xournalpp` theme and add it to the list of choices on all platforms (ensuring that MacOS can access some themes).

rolandlo:
> Thank you for testing it! It looks like we might need to determine where themes are stored on MacOS and call [gtk_source_style_scheme_manager_append_search_path](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/method.StyleSchemeManager.append_search_path.html).

I'll try to find out. For reference under Ubuntu 21.10 I have stored them under `/usr/share/gtksourceview-4/styles/`
In the `Xournal++.app` bundle they should hence be copied to `Xournal++.app/Contents/Resources/share/gtksourceview-4/styles`. Currently the `share` folder has no `gtksourceview-4` subfolder.

LittleHuba:
Maybe you need to add the path to the bundling script?

rolandlo:
> Maybe you need to add the path to the bundling script?

Yes, I think adding the following lines to `xournalpp.bundle` should do the trick.
```
  <data>
    ${prefix}/share/gtksourceview-4
  </data>
```

personalizedrefrigerator:
> > Maybe you need to add the path to the bundling script?
> 
> Yes, I think adding the following lines to `xournalpp.bundle` should do the trick.
> 
> ```
>   <data>
>     ${prefix}/share/gtksourceview-4
>   </data>
> ```

I've done this! How/whether to support `Ubuntu 18.04` is something else that needs to be considered... We could include building `gtksourceview-4` as a step for building (if `gtksourceview-4` is not available for a distro).

rolandlo:
> > > Maybe you need to add the path to the bundling script?
> > 
> > 
> > Yes, I think adding the following lines to `xournalpp.bundle` should do the trick.
> > ```
> >   <data>
> >     ${prefix}/share/gtksourceview-4
> >   </data>
> > ```
> 
> I've done this! How/whether to support `Ubuntu 18.04` is something else that needs to be considered... We could include building `gtksourceview-4` as a step for building (if `gtksourceview-4` is not available for a distro).

The problem with that is that `gtksourceview-4` depends on `Gtk >= 3.24` and Ubuntu 18.04 is shipped with Gtk 3.22, see [here](https://packages.ubuntu.com/search?suite=bionic&section=all&arch=any&keywords=libgtk-3&searchon=names).

rolandlo:
> > > > Maybe you need to add the path to the bundling script?
> > > 
> > > 
> > > Yes, I think adding the following lines to `xournalpp.bundle` should do the trick.
> > > ```
> > >   <data>
> > >     ${prefix}/share/gtksourceview-4
> > >   </data>
> > > ```
> > 
> > 
> > I've done this! How/whether to support `Ubuntu 18.04` is something else that needs to be considered... We could include building `gtksourceview-4` as a step for building (if `gtksourceview-4` is not available for a distro).
> 
> The problem with that is that `gtksourceview-4` depends on `Gtk >= 3.24` and Ubuntu 18.04 is shipped with Gtk 3.22, see [here](https://packages.ubuntu.com/search?suite=bionic&section=all&arch=any&keywords=libgtk-3&searchon=names).

What other options do we have? Could we make these distros use gtksourceview 3 instead or is that out of question?

personalizedrefrigerator:
> > > > > Maybe you need to add the path to the bundling script?
> > > > 
> > > > 
> > > > Yes, I think adding the following lines to `xournalpp.bundle` should do the trick.
> > > > ```
> > > >   <data>
> > > >     ${prefix}/share/gtksourceview-4
> > > >   </data>
> > > > ```
> > > 
> > > 
> > > I've done this! How/whether to support `Ubuntu 18.04` is something else that needs to be considered... We could include building `gtksourceview-4` as a step for building (if `gtksourceview-4` is not available for a distro).
> > 
> > 
> > The problem with that is that `gtksourceview-4` depends on `Gtk >= 3.24` and Ubuntu 18.04 is shipped with Gtk 3.22, see [here](https://packages.ubuntu.com/search?suite=bionic&section=all&arch=any&keywords=libgtk-3&searchon=names).
> 
> What other options do we have? Could we make these distros use gtksourceview 3 instead or is that out of question?

We can use gtksourceview3 on these distros, but we were running into issues with linking against one (and it not being present at runtime): https://github.com/xournalpp/xournalpp/pull/2809#issuecomment-964144891

Technius:
I don't think it's worth trying to support gtksourceview-3. It's too much work to maintain tests, build infrastructure, and compatibility for different versions of the library. And it's likely that we'll drop support for Ubuntu 18.04 within the year anyways since Ubuntu 22.04 will come out in a few months (dropping support will need to wait until after Xournal++ 1.2 to avoid surprises, though) .

Technius:
Ok, so my comment earlier was premature and not very well thought out.

I think we have a few options here:

* Require gtksourceview-4. Drop support for Ubuntu 18.04. This is a fairly drastic option that would require us to push this feature at least after the 1.2 release (which I think nobody would want to wait for).
* Require gtksourceview-3, but not gtksourceview-4. I don't know what the difference is between the two versions are, but the former would likely be available on more platforms?
* Make gtksourceview-4 optional (e.g., with a build flag), and do not use gtksourceview-3. Distro packagers can choose to enable the gtksourceview feature if they want, but to ensure maximum compatibility, our own builds will not have the feature enabled. Will need extra testing and build configuration to check builds with/without the flag.

I am mostly in favor of the last option. I understand why @Febbe may be opposed to it, but it will at least unblock this PR and will allow motivated users to have access to the feature with the least friction.

> @Technius So you suggest to drop support of Ubuntu 18.04 before it is EOL (which is only in April 2023)? That will probably receive some pushback from the users of Ubuntu 18.04 (and derivatives like Linux Mint 19.3) In that case the AppImage will have to be built on Ubuntu 20.04 as well.

@rolandlo I do not think we should drop support for Ubuntu 18.04 just to support this feature. But I have several good reasons for why we should consider dropping support for Ubuntu 18.04 after the 1.2 release (which is offtopic and can be discussed in another issue).

rolandlo:
> * Make gtksourceview-4 optional (e.g., with a build flag), and do not use gtksourceview-3. Distro packagers can choose to enable the gtksourceview feature if they want, but to ensure maximum compatibility, our own builds will not have the feature enabled. Will need extra testing and build configuration to check builds with/without the flag.

I tend to agree with this solution, but can't we enable the `gtksourceview` feature in all our builds except the one for Ubuntu 18.04? I really don't see the point why we should disable the feature on MacOS or Windows, for example.

Technius:
> I tend to agree with this solution, but can't we enable the gtksourceview feature in all our builds except the one for Ubuntu 18.04? I really don't see the point why we should disable the feature on MacOS or Windows, for example.

Sure. We can and should turn on the feature for MacOS and Windows since we control the distribution for that. The purpose of the build flag is to ensure that the application builds and runs on Ubuntu 18.04 while we still support it.

With just an opt-in build flag, we do not need to fiddle with any of the Debian package metadata or worry about missing libraries at runtime. We can remove the build flag and make the dependency required when we are ready to drop support for older distributions.

Febbe:
@xournalpp/core @personalizedrefrigerator what's the status of this PR?

personalizedrefrigerator:
> @xournalpp/core @personalizedrefrigerator what's the status of this PR?

A build flag needs to be added to disable linking to GtkSourceView! Without this, the PR won't build on Ubuntu 18.04, which lacks gtksourceview4.

Febbe:
@xournalpp/core, in my opinion, it is not required to build on ubuntu18.04 anymore. It is superseded by 2 LTS versions: ubuntu20.04 and ubuntu22.04.

Febbe:
@personalizedrefrigerator what's the problem of a feature testing variable, which conditionally disables this?

rolandlo:
> @xournalpp/core, in my opinion, it is not required to build on ubuntu18.04 anymore. It is superseded by 2 LTS versions: ubuntu20.04 and ubuntu22.04.

We build the AppImage on Ubuntu 18.04, whichis a very convenient way to test Xournal++ on multiple platforms without installing it. What to do about that then?

personalizedrefrigerator:
> @personalizedrefrigerator what's the problem of a feature testing variable, which conditionally disables this?

Sorry! I've been busy recently! I'll try to make time to add that tomorrow!

personalizedrefrigerator:
> @personalizedrefrigerator what's the problem of a feature testing variable, which conditionally disables this?

I've added the `USE_GTK_SOURCEVIEW` CMake flag!

While I've set `USE_GTK_SOURCEVIEW=OFF` for the Ubuntu 18.04 CI pipeline, we're still building the Linux AppImages on Ubuntu 18.04. As such, I currently expect the release pipelines for Linux to fail. Is it possible to move the Linux CI to Ubuntu 20.04 or Ubuntu 22.04?

rolandlo:
On Ubuntu 22.04 I have tested out the version with `USE_GTK_SOURCVIEW=OFF` (and no `libgtksourceview-4-dev` package installed) and the version with `USE_GTK_SOURCVIEW=ON` (and `libgtksourceview-4-dev` package installed). Both versions work fine. It's a great step ahead! The LaTeX tool is really getting more and more powerful. 
One thing I noticed: When I uncheck `Use font from system theme` and choose a font (in the LaTeX panel of the preferences), I get error messages in the console:
```
(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.984: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.984: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
```
The app continues to work fine though. Is that something we should worry about?

With the version that  uses gtksourceview I also got some warnings like:
```
(xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined

(xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined

(xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined
```

I will continue testing.

personalizedrefrigerator:
> On Ubuntu 22.04 I have tested out the version with `USE_GTK_SOURCVIEW=OFF` (and no `libgtksourceview-4-dev` package installed) and the version with `USE_GTK_SOURCVIEW=ON` (and `libgtksourceview-4-dev` package installed). Both versions work fine. It's a great step ahead! The LaTeX tool is really getting more and more powerful. One thing I noticed: When I uncheck `Use font from system theme` and choose a font (in the LaTeX panel of the preferences), I get error messages in the console:
> 
> ```
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.984: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> 
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.984: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> 
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> 
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> 
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> 
> (xournalpp:31652): GLib-GObject-CRITICAL **: 22:45:14.985: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
> ```
> 
> The app continues to work fine though. Is that something we should worry about?
> 
> With the version that uses gtksourceview I also got some warnings like:
> 
> ```
> (xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined
> 
> (xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined
> 
> (xournalpp:36414): GtkSourceView-WARNING **: 23:03:52.779: in file /usr/share/gtksourceview-4/language-specs/latex.lang: style 'def:text' not defined
> ```
> 
> I will continue testing.

Thank you for testing! Do the warnings (first group) appear also when using the font picker for the text tool (toolbar button)?

rolandlo:
> Do the warnings (first group) appear also when using the font picker for the text tool (toolbar button)?

No, they don't.

src/core/gui/dialog/LatexDialog.cpp
```diff
@@ -13,26 +13,140 @@
...
+    if (styleScheme) {
+        gtk_source_buffer_set_style_scheme(GTK_SOURCE_BUFFER(this->textBuffer), styleScheme);
+    }
+#endif
+
+    // Enable/disable word-wrap.
+    GtkWrapMode texBoxWrapMode = GTK_WRAP_NONE;
+    if (settings.editorWordWrap) {
+        texBoxWrapMode = GTK_WRAP_WORD_CHAR;
+    }
```
rolandlo, line 89:
```suggestion
    GtkWrapMode texBoxWrapMode = settings.editorWordWrap ? GTK_WRAP_WORD_CHAR : GTK_WRAP_NONE;
```
looks simpler to me.

src/core/gui/dialog/LatexDialog.cpp
```diff
@@ -86,6 +243,69 @@ auto LatexDialog::getBufferContents() -> std::string {
...
+    cairo_matrix_t matTransformed;
+    cairo_get_matrix(cr, &matOriginal);
+    cairo_get_matrix(cr, &matTransformed);
+
+    matTransformed.xx = scaleFactor;
+    matTransformed.yy = scaleFactor;
+    matTransformed.xy = 0;
+    matTransformed.yx = 0;
+    matTransformed.x0 += translateX;
+    matTransformed.y0 += translateY;
```
rolandlo, line 296: That could also be written as follows:
```suggestion
    cairo_matrix_init(&matTransformed, scaleFactor, 0, 0, scaleFactor, matOriginal.x0 + translateX, matOriginal.y0 + translateY);
```

rolandlo (commented):
I have done more testing on Linux and Windows. Regarding MacOS: The Gtk blob currently misses GtkSourceView (it was present in the previous blob; this is my fault), but that is unrelated to this PR.

In general the PR works pretty well. The settings introduced in the LaTeX panel work flawlessly. I tested it both with the USE_GTK_SOURCEVIEW flag turned on and off. Code-wise I have found nothing serious and only added some minor comments.

One thing that I find a bit disturbing: Whenever an existing LaTeX formula is edited, the preview dimensions and LaTeX dialog dimensions are reset. Given that the user has intentionally scaled the preview and LaTeX dialog it seems better UX that the same dimensions are used again. So maybe each TeXImage element should store these dimensions/factors. 


Some things I noticed while testing the PR more and which are already present in current master:

1) When an existing LaTeX formula is edited, the TexImage inserted into the document keeps the same width and height. That is frequently unwanted. Say the existing LaTeX formula consists of one line of math and the updated formula consists of multiple lines. Then it makes no sense to keep the same height. I would expect the first line written to keep the same height.
A simple solution would be to remove the lines
https://github.com/xournalpp/xournalpp/blob/962fd4be4f40061462a75404447c5c469e56153e/src/core/control/LatexController.cpp#L102-L103
On the other hand, if a TeXImage was scaled intentionally by the user, the scaling would be undone by any modification of the LaTeX formula. So it might make sense to store the user-applied scaling factors in x- and y-direction and reapply them rather than to keep width and height the same.


2) When `pdflatex` is not in the path, I get`the message:
```
(xournalpp.exe:4512): GLib-CRITICAL **: 20:47:15.080: g_error_free: assertion 'error != NULL' failed
````
while opening the LaTeX tool. This message was obtained on Windows, but Linux shows the same assertion violation.

File: src/core/control/LatexController.cpp [hunk 8]
```diff
@@ -149,16 +152,26 @@ void LatexController::triggerImageUpdate(const string& texString) {
...
     auto result = generator.asyncRun(this->texTmpDir, texContents);
     if (auto* err = std::get_if<LatexGenerator::GenError>(&result)) {
         XojMsgBox::showErrorToUser(this->control->getGtkWindow(), err->message);
     } else if (auto** proc = std::get_if<GSubprocess*>(&result)) {
         updating_cancellable = g_cancellable_new();
-        g_subprocess_wait_check_async(*proc, updating_cancellable,
-                                      reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), this);
+        char* stdinBuff = nullptr;  // No stdin
+
+        g_subprocess_communicate_utf8_async(*proc, stdinBuff, updating_cancellable,
```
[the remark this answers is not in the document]
personalizedrefrigerator:
```
// Render the TeX and capture the process' output.
```
has been added on line 170.

src/core/gui/dialog/LatexDialog.cpp
personalizedrefrigerator: As above, I personally find the longer/multi-line form easier to read. Of course, if you disagree, I'm willing to change it!

src/core/gui/dialog/LatexDialog.h
```diff
@@ -51,6 +63,47 @@ class LatexDialog: public GladeGui {
...
+     */
+    void setupCSS();
+
+private:
+    /**
+     * @brief Called on 'draw' signal.
+     * @param widget is the target of the event; the GtkDrawingArea we're rendering to.
+     * @param cr is drawn to
+     * @param renderedTexPtrPtr should be a cairo_surface_t** containing the rendered
+     *  preview.
```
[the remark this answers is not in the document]
personalizedrefrigerator: Thank you for catching this! It should be fixed!

src/core/gui/dialog/LatexSettingsPanel.cpp
```diff
@@ -12,15 +17,39 @@
...
+#ifdef USE_GTK_SOURCEVIEW
+    this->sourceViewThemeSelector = gtk_source_style_scheme_chooser_button_new();
+    gtk_container_add(themeSelectionBoxContainer, sourceViewThemeSelector);
+
+    gtk_label_set_text(GTK_LABEL(this->get("lbSourceviewSettingsDescription")), _("LaTeX editor theme:"));
+#else
+    this->sourceViewThemeSelector = nullptr;
+
+    gtk_label_set_text(GTK_LABEL(this->get("lbSourceviewSettingsDescription")),
+                       _("GtkSourceView was disabled when building Xournal++!"
```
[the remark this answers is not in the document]
personalizedrefrigerator: Thank you for catching this! I've fixed it!

src/core/gui/dialog/LatexDialog.cpp
rolandlo: Those two were only suggestions, a matter of taste. I don't mind if you keep it the way it is.

rolandlo:
> > Whenever an existing LaTeX formula is edited, the preview dimensions and LaTeX dialog dimensions are reset. Given that the user has intentionally scaled the preview and LaTeX dialog it seems better UX that the same dimensions are used again. So maybe each TeXImage element should store these dimensions/factors.
> 
> I agree. I'm hesitant to add more commits to this PR (as it's already quite large). Perhaps this could be done in a follow-up PR.

That's fine for me.

rolandlo (approved):
I suggest we merge the PR now.

rolandlo:

rolandlo:
I think we should add the build flag ` -DUSE_GTK_SOURCEVIEW=OFF` to `release.yml` and `create-installers.yml` (3 lines below `- template: steps/build_linux.yml` in both files), otherwise the release build and the installer creation are not working any more for Ubuntu/Debian. That's in accordance with https://github.com/xournalpp/xournalpp/pull/2809#issuecomment-1050686763

By the way I have updated the Gtk blob. It contains `gtksourceview` 4.8.1 again.

rolandlo:
The flag is also needed in `release_ubuntu.yml`. Forgot to mention it with the other two files.

Febbe:
Why is this flag not autodetected by cmake?

CMakeLists.txt
```diff
@@ -93,6 +93,13 @@ list(APPEND pkg-module-spec "portaudiocpp >= 12")
 list(APPEND pkg-module-spec "sndfile >= 1.0.25")
 list(APPEND pkg-module-spec "librsvg-2.0 >= 2.40")
 
+# Ubuntu 18.04 does not package gtksourceview-4
+option (USE_GTK_SOURCEVIEW "Link with gtksourceview-4 (for TeX syntax highlighting)" ON)
+if (USE_GTK_SOURCEVIEW)
+  list(APPEND pkg-module-spec "gtksourceview-4 >= 4.0.0")
+  add_compile_definitions(USE_GTK_SOURCEVIEW)
+endif(USE_GTK_SOURCEVIEW)
+
```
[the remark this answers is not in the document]
rolandlo: @Febbe @personalizedrefrigerator Ok, how do we proceed from here. Fix the CMakeLists.txt issue before merging or merging first and then fixing the issue with a new PR?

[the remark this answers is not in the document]
Technius: Let's do that in a follow up.

rolandlo:
Thanks @personalizedrefrigerator for the wonderful work! Just merged it and looking forward to using the improved LaTeX tool.

CMakeLists.txt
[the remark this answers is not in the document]
Febbe: As long someone feels responsible to do this quite fast. Such merges are always ok I think. Does someone of you want to fix it?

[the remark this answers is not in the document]
rolandlo: I will take care of it.

PR #2814, opened 2021-02-17
Fix librsvg dependency

rolandlo:
The PPA (for the latest master) does not build since February 4, 2021 because `librsvg-2.0` is missing. It is required for the thumbnailer. So we must ensure that the package gets installed.


Technius:
LGTM for the Debian side. I would be careful about updating the rpm file, as it's used for the Fedora equivalent of PPAs: https://github.com/xournalpp/xournalpp/pull/1703#discussion_r366662268

rolandlo:
Ok, I undid the changes in the rpm file. Should I merge now?

Issue #2711, opened 2021-01-26
Assignment of "-1" to unsigned values


yurivict:
**Affects versions :**
All

**Describe the bug**
Assigning -1 to ```size_t``` value here https://github.com/xournalpp/xournalpp/blob/master/src/gui/XournalView.h#L167 is wrong. Compiler catches this:
```
/usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/ClipboardHandler.h:66:24: warning: implicit conversion changes signedness: 'int' to 'gulong' (aka 'unsigned long') [-Wsign-conversion]
    gulong hanlderId = -1;
                     ~ ^~
```

This problem is systematic:
```
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/tools/StrokeHandler.cpp:8:
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/Control.h:22:
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/jobs/XournalScheduler.h:21:
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/gui/sidebar/previews/page/SidebarPreviewPageEntry.h:17:
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/gui/sidebar/previews/page/SidebarPreviewPages.h:19:
/usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/gui/sidebar/previews/base/SidebarPreviewBase.h:113:28: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
    size_t selectedEntry = -1;
                         ~ ^~
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/tools/StrokeHandler.cpp:8:
In file included from /usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/Control.h:31:
/usr/ports/graphics/xournalpp/work/xournalpp-6ffcf8be/src/control/ClipboardHandler.h:66:24: warning: implicit conversion changes signedness: 'int' to 'gulong' (aka 'unsigned long') [-Wsign-conversion]
    gulong hanlderId = -1;
                     ~ ^~
[...]

PR #2821, opened 2021-02-17
Fixup Zoom loop, when using 2 different page sizes

### LINKED ISSUE #2770: Segmentation Fault on scrolling through page of different size
**Affects versions :**
* Ubuntu 20.04, Gnome X11
* Xournal++ 1.1.0+dev
* libgtk: 3.24.20
* PPA

**Describe the bug**
Xournalpp has a segmentation fault when it scrolls through a page of different size to other pages

**To Reproduce**
1. Create a document with 3 pages: (page 1: A4, page 2: A3, page 3: A4).
2. Scroll through document.
3. Xournal++ crashes

**Additional context**
* According to @rolandlo Xournal++ enters an infinite (scrolling, zooming) loop.
* Issue does not occur with Flatpak(1.0.20).

Febbe:
This fixup introduces a check, whether the old selected page is still visible, only if not, the new zoom_to_fit value is calculated.
This has the drawback, that it defers the zooming to the page size until the old page is not visible anymore.

fixes #2770

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -101,7 +101,7 @@ void ZoomControl::zoomScroll(ZoomDirection zoomIn, utl::Point<double> zoomCenter
         return;
     }
 
-    if (this->zoomFitMode) {
+    if (this->isZoomFitMode()) {
         this->setZoomFitMode(false);
     }
 
@@ -256,7 +256,7 @@ auto ZoomControl::updateZoomFitValue(size_t pageNo) -> bool {
 
     this->zoomFitValue = zoom_fit_width;
     fireZoomRangeValueChanged();
-    if (this->zoomFitMode && !this->zoomPresentationMode) {
+    if (this->isZoomFitMode() && !this->zoomPresentationMode) {
         this->zoomFit();
     }
     return true;
@@ -303,7 +303,7 @@ void ZoomControl::zoom100() {
 }
 
 void ZoomControl::zoomFit() {
-    if (this->zoomFitMode && !this->zoomPresentationMode && this->zoom != this->zoomFitValue) {
+    if (this->isZoomFitMode() && !this->zoomPresentationMode && this->zoom != this->zoomFitValue) {
         startZoomSequence();
         this->zoomSequenceChange(this->zoomFitValue, false);
         endZoomSequence();
@@ -321,15 +321,24 @@ void ZoomControl::zoomPresentation() {
 void ZoomControl::setZoomFitMode(bool isZoomFitMode) {
     if (this->zoomFitMode != isZoomFitMode) {
         this->zoomFitMode = isZoomFitMode;
-        this->control->fireActionSelected(GROUP_ZOOM_FIT, isZoomFitMode ? ACTION_ZOOM_FIT : ACTION_NOT_SELECTED);
+        this->control->fireActionSelected(GROUP_ZOOM_FIT, this->zoomFitMode ? ACTION_ZOOM_FIT : ACTION_NOT_SELECTED);
     }
 
-    if (isZoomFitMode) {
+    if (this->isZoomFitMode()) {
         zoomFit();
     }
 }
 
-auto ZoomControl::isZoomFitMode() const -> bool { return this->zoomFitMode; }
+auto ZoomControl::isZoomFitMode() const -> bool {
+    // Todo(fabian): Remove this fix and make both modes possible in parallel, after fixing the infinite loop.
+    //  Explanation: First of all, zoomFit never worked with paired pages. Also, using both resulted in a stackoverflow
+    //               when different page sizes are used. This is caused by a logical loop.
+    //               We decided to deactivate it in PR#2821 & I#2770. instead of fixing it, to get release 1.1.0 ready.
+    //               Zoom presentation mode is also excluded, because it was never intended to work together.
+    //               It is also excluded everywhere else (duplicate code).
+    auto infiniteLoopFixup = !this->zoomPresentationMode && !this->control->getSettings()->isShowPairedPages();
+    return this->zoomFitMode && infiniteLoopFixup;
+}
 
 void ZoomControl::setZoomPresentationMode(bool isZoomPresentationMode) {
     this->zoomPresentationMode = isZoomPresentationMode;
@@ -352,7 +361,7 @@ void ZoomControl::pageSizeChanged(size_t page) {
 
 void ZoomControl::pageSelected(size_t page) {
     // Todo (fabian): page selected should do nothing here, since Zoom Controls, which page is selected.
-    //               This results in a logical loop
+    //                This results in a logical loop. See PR#2821 & I#2770
     if (current_page != page) {
         this->last_page = this->current_page;
         this->current_page = page;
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
This seems to make presentation mode not zoom in to smaller pages. See below.

https://user-images.githubusercontent.com/46334387/108274012-33b48b00-7129-11eb-9a20-81d454b27efb.mp4

Technius:
Nice!

> This fixup introduces a check, whether the old selected page is still visible, only if not, the new zoom_to_fit value is calculated. This has the drawback, that it defers the zooming to the page size until the old page is not visible anymore.

I think this is a very reasonable approach to the problem and I would say the _expected behavior of zoom fit_ by default. When you're reading/annotating and scrolling, you typically want to finish reading/annotating the page at the same zoom level without a sudden jump to the next page.

In contrast, in presentation mode (as noted by @personalizedrefrigerator), scrolling down should immediately change the zoom because _presentation mode expects exactly one page to be shown_.

src/control/zoom/ZoomControl.cpp
```diff
@@ -351,6 +351,16 @@ void ZoomControl::pageSizeChanged(size_t page) {
...
+    //               This results in a logical loop
+    if (current_page != page) {
+        this->last_page = this->current_page;
+        this->current_page = page;
+    }
+
+    if (view->isPageVisible(this->last_page, nullptr)) {
+        return;
+    }
+    updateZoomPresentationValue(this->current_page);
```
Technius, line 364: I'm guessing that this should be moved before the visibility check to address the presentation mode zoom problem.

Febbe: done.

Febbe:
> I think this is a very reasonable approach to the problem and I would say the expected behavior of zoom fit by default. When you're reading/annotating and scrolling, you typically want to finish reading/annotating the page at the same zoom level without a sudden jump to the next page.

Adobe even chose the way to never adopt that value automatically. Only when the user decides to press that button.

> In contrast, in presentation mode (as noted by @personalizedrefrigerator), scrolling down should immediately change the zoom because presentation mode expects exactly one page to be shown.

In my opinion Presentation mode should never have more than 1 page in the view.

Febbe:
The infinite loop still exist. Now it is triggered when enabling 2 page view.
I still think we should merge this, since the 2page view never worked with zoom to fit.

Technius:
> The infinite loop still exist. Now it is triggered when enabling 2 page view.
> I still think we should merge this, since the 2page view never worked with zoom to fit.

Let's also try to prevent this because someone is eventually going to trigger this infinite loop behavior. I think a good workaround for now is to turn off zoom fit when 2page is enabled.

Febbe:
We need to get rid of the logical loop:
grafik

Technius:
> ZoomHandler::zoomChanged should not change the selected page, since zooming in and out on a fixpoint, should not trigger the change of the page. Even if it does not zoom in on a fixpoint, because we are zooming in and out at the border of a document. This should not change the selected page while zooming.

I mean, we're well aware of this--wasn't this the point of the viewport rewrite? Except we can't delay 1.1.0 any longer as it includes way too many essential features/bug fixes.

Febbe:
@personalizedrefrigerator can you retest/review the pr again?

Febbe:
@Technius correct, not only that, but mostly yes, the thing is, the logical loop should be removed, and when I prevent it, noone can track it down anymore, that's why I posted the stacktrace. we should add this to the viewport metaissue

personalizedrefrigerator:
This is a lot better! Thank you!

<details>
<summary>Issue I found (present before this PR</summary>

Document causing issues

**It looks like this issue was present before this PR:** When presenting in paired pages mode, I am unable to view page "3" in the above document. Attempting to do so moves the view to either page 2 or page 6.

</details>

src/control/zoom/ZoomControl.h
```diff
@@ -236,4 +236,7 @@ class ZoomControl: public DocumentListener {
      * depends on dpi (REAL_PERCENTAGE_VALUE * zoom100Value)
      */
     double zoomMin = DEFAULT_ZOOM_MIN;
+
+    size_t current_page = static_cast<size_t>(-1);
+    size_t last_page = static_cast<size_t>(-1);
```
personalizedrefrigerator, line 241: Could we use [`npos`](https://github.com/xournalpp/xournalpp/issues/2711) here? 
It looks like [it's equivalent to static_cast<size_t>(-1)](https://www.cplusplus.com/reference/string/string/npos/)...

src/control/zoom/ZoomControl.cpp
```diff
@@ -351,6 +355,16 @@ void ZoomControl::pageSizeChanged(size_t page) {
 }
 
 void ZoomControl::pageSelected(size_t page) {
-    updateZoomPresentationValue(page);
-    updateZoomFitValue(page);
+    // Todo (fabian): page selected should do nothing here, since Zoom Controls, which page is selected.
+    //               This results in a logical loop
```
personalizedrefrigerator, line 359: It might be useful to add a link to this PR or the issue it fixes here. E.g.
```c++
// Todo (fabian): page selected should do nothing here, since Zoom Controls, which page is selected.
//               This results in a logical loop. See issue #2770.
```

src/control/zoom/ZoomControl.cpp
```diff
@@ -321,15 +321,19 @@ void ZoomControl::zoomPresentation() {
...
-    if (isZoomFitMode) {
+    if (this->isZoomFitMode()) {
         zoomFit();
     }
 }
 
-auto ZoomControl::isZoomFitMode() const -> bool { return this->zoomFitMode; }
+auto ZoomControl::isZoomFitMode() const -> bool {
+    // Todo(fabian): Remove this fix and make both modes possible in parallel, after fixing the infinite loop
+    auto infiniteLoopFixup = !this->zoomPresentationMode && !this->control->getSettings()->isShowPairedPages();
```
personalizedrefrigerator, line 334: It would be nice to have a comment that explains why `infiniteLoopFixup` is set this way...

personalizedrefrigerator (approved):
This looks a lot better! Aside from a few minor suggestions, this looks good to me!

src/control/zoom/ZoomControl.h
Febbe: `npos` is a string constant and should only be used parsing strings. We could use `std::numeric_limits<size_t>::max();`
But mostly this is overkill and `static_cast<size_t>(-1)` or `size_t(-1)` will to the same job without reducing readability.

Febbe:
> 
> 
> This is a lot better! Thank you!
> I did find one issue, however, it seems to have been present before this PR.
> 
> Document causing issues
> 
> **It looks like this issue was present before this PR:** When presenting in paired pages mode, I am unable to view page "3" in the above document. Attempting to do so moves the view to either page 2 or page 6.

Did you create an issue for that?

personalizedrefrigerator:
> Did you create an issue for that?

#2847

Febbe:
Ready to be merged after pipeline succeeds.

rolandlo:
Happy you fixed it @Febbe.

PR #2822, opened 2021-02-18
fix deb package with non-satisfiable dependency

rolandlo:
I am not sure it fixes the problem, but I don't see a way to try it out without merging the commit, letting it build on Launchpad and trying out the package.

Since `librsvg` is only used for the thumbnailer an option would be to add this dependency only to the `Recommends` field.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,7 +8,7 @@ Homepage: https://github.com/xournalpp/xournalpp/
 
 Package: xournalpp
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg-2.0 (>= 2.40)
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -8,7 +8,7 @@ Homepage: https://github.com/xournalpp/xournalpp/
 
 Package: xournalpp
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg-2.0 (>= 2.40)
+Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2 (>= 2.40)
```
Technius, line 11: [It goes by `librsvg2-2` according to the package search](https://packages.ubuntu.com/search?suite=bionic&searchon=names&keywords=librsvg).

rolandlo: Thanks for showing me, how to find the correct package name. Let's try it with `librsvg2-2` then.

PR #2826, opened 2021-02-18
Fix attributes of IM's preedit strings in TextEditor.

tattsan:
This PR deals with the attributes for preedit strings of IM.
1. Fix that attributes were not reflected.
2. Use attributes for the preedit strings set by IM.

**On the first point**
Seeing TextEditor.cpp, IM's preedit strings are to be displayed with underline.
But subsequent code for attributes of selections overrides it, and no underline appears.
Because TextSelection and PreEditingTexts are exclusive operation, there's no needs to compose their attributes.
So I moved the TextSelection code to PreEditing code's else clause.

**On the second point**
There are various IM programs, and they set attributes for preedit strings.  So Xournal++ should not set attributes for preedit string, but use IM's attributes. The PR uses only  IM's attributes, and does not intend to compose them with other attributes. When we need another kind of attributes, it'll be rewritten.

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -91,6 +91,10 @@ TextEditor::~TextEditor() {
         g_object_unref(this->layout);
         this->layout = nullptr;
     }
+
+    if (this->preeditAttrList) {
+        pango_attr_list_unref(this->preeditAttrList);
+    }
 }
 
 auto TextEditor::getText() -> Text* {
@@ -183,6 +187,15 @@ void TextEditor::iMPreeditChangedCallback(GtkIMContext* context, TextEditor* te)
      */
     gtk_im_context_get_preedit_string(context, &str, &attrs, &cursor_pos);
 
+    if (attrs == nullptr) {
+        attrs = pango_attr_list_new();
+    }
+    if (te->preeditAttrList) {
+        pango_attr_list_unref(te->preeditAttrList);
+    }
+    te->preeditAttrList = attrs;
+    attrs = nullptr;
+
     if (str && str[0] && !gtk_text_iter_can_insert(&iter, true)) {
         gtk_widget_error_bell(te->widget);
         goto out;
@@ -198,7 +211,6 @@ void TextEditor::iMPreeditChangedCallback(GtkIMContext* context, TextEditor* te)
 
 out:
 
-    pango_attr_list_unref(attrs);
     g_free(str);
 }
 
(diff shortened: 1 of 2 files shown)


tattsan:
I'll compare two IM's -- ibus-skk and fcitx-skk .
**ibus-skk**

https://user-images.githubusercontent.com/6355486/108408474-6add8600-7268-11eb-8ff9-8f6e48c801eb.mp4

**fcitx-skk**

https://user-images.githubusercontent.com/6355486/108408535-7b8dfc00-7268-11eb-9350-da28cd79ba68.mp4

src/gui/TextEditor.cpp
```diff
@@ -960,53 +961,47 @@ void TextEditor::paint(cairo_t* cr, GdkRectangle* repaintRect, double zoom) {
...
-            list = pango_attr_list_new();
+            if (list == nullptr) {
+                list = pango_attr_list_new();
+                pango_layout_set_attributes(this->layout, list);
+            }
+            pango_attr_list_insert(list, attrib);
+        } else {
+            // remove all attributes
+            PangoAttrList* list = pango_attr_list_new();
             pango_layout_set_attributes(this->layout, list);
```
Technius, line 1003: As far as I understand, we need a `pango_attr_list_unref(list)` here since `list` is owned here and `pango_layout_set_attributes` only borrows `list`.

Technius (changes requested):
Thanks for the PR, this looks great. I have some concerns about the memory management.

src/gui/TextEditor.cpp
```diff
@@ -193,6 +193,7 @@ void TextEditor::iMPreeditChangedCallback(GtkIMContext* context, TextEditor* te)
     } else {
         te->preeditString = "";
     }
+    te->preeditAttrList = attrs;
```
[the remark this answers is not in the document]
tattsan: I agree.

src/gui/TextEditor.cpp
```diff
@@ -960,53 +961,47 @@ void TextEditor::paint(cairo_t* cr, GdkRectangle* repaintRect, double zoom) {
...
-        attrib->start_index = pos;
-        attrib->end_index = pos + preeditString.length();
-
-        if (list == nullptr) {
-            list = pango_attr_list_new();
-            pango_layout_set_attributes(this->layout, list);
+        PangoAttrList* list = pango_attr_list_new();
+        PangoAttrList* preedit_list = this->preeditAttrList;
+        if (list != nullptr) {
+            pango_attr_list_splice(list, preedit_list, pos, preeditString.length());
```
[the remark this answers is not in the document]
tattsan: ``list`` in the line 973 is a typo for ``preedit_list`` .  There may be some IM which set no attribute on preedit string. But this is not a suitable place, I'll moved the code for that into ``iMPreeditChangedCallback`` .

src/gui/TextEditor.cpp
```diff
@@ -960,53 +961,47 @@ void TextEditor::paint(cairo_t* cr, GdkRectangle* repaintRect, double zoom) {
...
-        attrib->start_index = getByteOffset(gtk_text_iter_get_offset(&start));
-        attrib->end_index = getByteOffset(gtk_text_iter_get_offset(&end));
+            attrib->start_index = getByteOffset(gtk_text_iter_get_offset(&start));
+            attrib->end_index = getByteOffset(gtk_text_iter_get_offset(&end));
 
-        if (list == nullptr) {
-            list = pango_attr_list_new();
+            if (list == nullptr) {
+                list = pango_attr_list_new();
+                pango_layout_set_attributes(this->layout, list);
```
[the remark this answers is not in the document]
tattsan: I agree.

src/gui/TextEditor.cpp
tattsan: I agree.

tattsan:
@Technius  Thank you so much for all your help. I modified the PR according to your comments.
I also made another commit to rename ``list`` to ``attrlist`` for code readability.

tattsan:
I've tested another IM (uim), and run into crash. Error log showed that it occured around  pango_attr_list_splice .

src/gui/TextEditor.cpp
```diff
@@ -183,6 +183,13 @@ void TextEditor::iMPreeditChangedCallback(GtkIMContext* context, TextEditor* te)
      */
     gtk_im_context_get_preedit_string(context, &str, &attrs, &cursor_pos);
 
+    if (attrs == nullptr) {
+        attrs = pango_attr_list_new();
+    }
+    te->preeditAttrList = attrs;
+    pango_attr_list_unref(attrs);
```
Technius, line 190: This `pango_attr_list_unref` might be causing the new crash you got. AFAICT ownership is transferred to `te->preeditattrList`, so this unref will drop the reference count to 0 and then lead to double free! I

src/gui/TextEditor.cpp
```diff
@@ -960,53 +966,51 @@ void TextEditor::paint(cairo_t* cr, GdkRectangle* repaintRect, double zoom) {
...
-        if (list == nullptr) {
-            list = pango_attr_list_new();
-            pango_layout_set_attributes(this->layout, list);
+            if (attrlist == nullptr) {
+                attrlist = pango_attr_list_new();
+                pango_layout_set_attributes(this->layout, attrlist);
+                pango_attr_list_unref(attrlist);
+                attrlist = nullptr;
+            }
+            pango_attr_list_insert(attrlist, attrib);
```
Technius, line 1006: This doesn't seem quite right. If `attrlist` is null above, it will be set to null again.
```suggestion
            const bool isNewAttrlist = attrlist == nullptr;
            if (isNewAttrlist) {
                attrlist = pango_attr_list_new();
                pango_layout_set_attributes(this->layout, attrlist);
            }
            pango_attr_list_insert(attrlist, attrib);
            if (isNewAttrlist) {
                pango_attr_list_unref(attrlist);
                attrlist = nullptr;
            }
```

src/gui/TextEditor.cpp
tattsan: Bingo! Dropping the ``pango_attr_list_unref`` makes ``uim`` work properly.

src/gui/TextEditor.cpp
tattsan: How fool I am! Thanks.

src/gui/TextEditor.cpp
tattsan: Instead ``pango_attr_list_unref(te->preeditAttrList);`` is needed before ``te->preeditAttrList = attrs;``, isn't it?

tattsan:
I've checked ``uim`` as IM, with anthy as it's backend engine.

https://user-images.githubusercontent.com/6355486/108592603-fe6e9e00-73b1-11eb-867e-35fe1401a50d.mp4

I've also checked uim-skk, ibus-anthy, fcitx-anthy.

tattsan:
After the final review, I'll squash all the commits to one.

Technius:
`preeditAttrList` is not being freed in the destructor.

src/gui/TextEditor.cpp
Technius: I think the `pango_attr_list_unref` line should be removed entirely. If you call it twice, you get a double-free segfault, which suggests that calling the unref here is going to free attrs.

tattsan: All right. I will remove the line.

tattsan:
I've removed ``pango_attr_list_unref`` from ``iMPreeditChangedCallback``, and squashed all the commits to one. Thank you so much @Technius !

src/gui/TextEditor.cpp
```diff
@@ -183,6 +183,12 @@ void TextEditor::iMPreeditChangedCallback(GtkIMContext* context, TextEditor* te)
      */
     gtk_im_context_get_preedit_string(context, &str, &attrs, &cursor_pos);
 
+    if (attrs == nullptr) {
+        attrs = pango_attr_list_new();
+    }
+    te->preeditAttrList = attrs;
```
Technius, line 189: I just realized that we need to unref the previous attribute list.
```suggestion
    if (te->preeditAttrList) {
        pango_attr_list_unref(te->preeditAttrList);
    }
    te->preeditAttrList = attrs;
```

Technius (commented):
No problem, happy to help. There's still two issues I can see. One of them is in a review comment below. The other issue is that the `te->preeditAttrList` needs to be unref'd in the destructor if it is not null.

tattsan:
>The other issue is that the ``te->preeditAttrList`` needs to be unref'd in the destructor if it is not null.

Done.  I had no experience with C++. It was a great learning experience.

Technius:
LGTM, thanks.

Issue #2844, opened 2021-02-22
Cannot choose language in "xournalpp-1.1.0+dev-nightly.20210221"


MichelReims:
OS : Windows 10
Version of xournalpp : xournalpp-1.1.0+dev-nightly.20210221

Issue : It is no longer possible to choose the language when it was possible in previous versions

rolandlo:
@MichelReims Can you say what was the last version where you could choose the GUI language?

MichelReims:
For example I can choose language in "xournalpp-1.1.0+dev-nightly.20210204-windows", but not in "xournalpp-1.1.0+dev-nightly.20210207-windows".
I think the problem appeared when during the installation of Xournalpp there is a new dialog box : "Install for all users"

rolandlo:
Ok, that's when we overhauled the Windows installer. Probably the location of the localization files is not found. I will take a look how to fix it.

MichelReims:
Thank you

rolandlo:
My observation (under Windows):
- If Xournal++ from the Start Menu localizations work normally.
- If Xournal++ is run from the Command Prompt from the `bin` subfolder of the installation folder (typing `xournalpp.exe`) it runs fine as well
- If Xournal++ is run from the Command Prompt from a different folder (e.g. from the installation folder using `bin\xournalpp.exe`) localizations are not found
- If Xournal++ is run via the context menu of a .xopp-file in the file manager localizations are not found.

The reason is simple: Xournal++ will search the localizations either 
1) from the path specified in the TEXTDOMAINDIR environment variable (which we do not set in Windows; it is used for the AppImage) or 
2) relative to the present working directory in `..\share\locale`. The present working directory should be the folder which contains `xournalpp.exe`, otherwise this won't work.

@MichelReims Can you please verify my findings so far?

rolandlo:
One more observation:
If you create an environment variable TEXTDOMAINDIR with value `/c/Program Files/Xournal++/share/locale` then the GUI picks up the system language, but you cannot change the language in the preferences

MichelReims:
On my computer, the only time xournalpp works normally is when I run it from "bin" subfolder of the installation folder. In other cases it does not work normally.
It's already good like that

rolandlo:
Did you try to uninstall Xournal++ before reinstalling the new version? Maybe the start menu entry is from an older version.

MichelReims:
Yes I uninstalled and reinstalled it. It does not change anything.
[...]

PR #2845, opened 2021-02-22
Make TEXTDOMAINDIR work under Windows

rolandlo:
The searchpath separator under Windows is not `:` like on Linux/Unix, but `;`
To distinguish OS's we can use the `G_SEACHPATH_SEPARATOR` variable, see [this link](https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#G-SEARCHPATH-SEPARATOR:CAPS).


rolandlo:
Fixes #2844

PR #2846, opened 2021-02-22
Fix Rectangle::unite not properly calculating the union

bhennion:
Intersection of rectangles was not computed properly. This fixes it.

Example: `{1, 0, 1, 0}.unite({0, 0, 1, 0})` resulted in `{0, 0, 1, 0}` instead of `{0, 0, 2, 0}`.


rolandlo (approved):
LGTM. Thanks for the fix.

PR #2849, opened 2021-02-22
Escape linebreaks when saving LaTeX text

### LINKED ISSUE #2526: Closing then reopening file removes all line breaks in "edit LaTeX" mode

personalizedrefrigerator:
image
image

On file load, linebreaks in XML attributes are ignored. Linebreaks were not escaped on file save. This PR escapes linebreaks on file save. 

This should be backwards compatible because we use a `GMarkupParseContext` to parse XML (on file load), which handles escape sequences for us.

Should fix #2526.

### DIFF
--- a/src/control/xml/TextAttribute.cpp
+++ b/src/control/xml/TextAttribute.cpp
@@ -15,6 +15,7 @@ void TextAttribute::writeOut(OutputStream* out) {
                                             replace_pair('\"', "&quot;"),
                                             replace_pair('<', "&lt;"),
                                             replace_pair('>', "&gt;"),
+                                            replace_pair('\n', "&#13;"),
                                     });
     out->write(v);
 }


rolandlo (approved):
LGTM. Nice fix.

DUOLabs333:
@personalizedrefrigerator I'm not sure exactly what you meant by "backwards compatible", but textboxes that had line breaks before the update, don't get line breaks upon the load. It's only for any saved textboxes with line breaks going forward.

personalizedrefrigerator:
@DUOLabs333 I meant that old versions of Xournal++ could open & edit the LaTeX without an error (i.e. there is no breaking change in the file format).

PR #2857, opened 2021-02-23
Fix invalid stroke width and image dimensions

### LINKED ISSUE #2464: Invalid stroke width or image dimensions leading to disappearing selections and missing parts in pdf export
**Affects versions :**
 - OS: doesn't matter, I guess
  - Which version of libgtk do you use: doesn't matter, I guess
 - Version of Xournal++: 1.0.18, 1.0.19, 1.1.0+dev (possibly others)
 - Installation method: doesn't matter, I guess

**Describe the bug**
Xournal++ sometimes stores `-NaN` as a stroke width (see #2372, #2168), which causes trouble using the selection tools (parts of the document disappear, the cursor update looks flawed, and exporting the pdf can make Xournal++ crash

Xournal++ also sometimes stores images with 0 width and height (and 0's as x and y-offsets) and strokes with all 0's as width, see [this file](https://files.gitter.im/5647729516b6c7089cbab6bf/UfoH/aula11.xopp) reported on the Gitter channel, which leads to similar misbehavior.

**To Reproduce**
It's not clear yet, how these strokes and images are created. Once they are created it's easy to reproduce following the steps described in #2372. 

**Expected behavior**
Width `-NaN` should not be possible. Such strokes should get filtered out immediately when they are created. Similarly for strokes with all 0's as width and images with 0 width and/or height. 

**Screenshots of Problem**
See #2372. 

**Additional context**
The problem with images of zero width and height may be related to undo operation. If you insert an image and undo its insertion, you will see the edit selection at the top left corner of the document with 0 width and height. I couldn't reproduce getting it saved in the xopp-file though.

rolandlo:
Fixes #2464. 
See #2464 for details.

### DIFF
--- a/src/control/tools/EditSelectionContents.cpp
+++ b/src/control/tools/EditSelectionContents.cpp
@@ -553,6 +553,11 @@ void EditSelectionContents::serialize(ObjectOutputStream& out) {
     out.writeDouble(this->originalBounds.width);
     out.writeDouble(this->originalBounds.height);
 
+    out.writeDouble(this->lastSnappedBounds.x);
+    out.writeDouble(this->lastSnappedBounds.y);
+    out.writeDouble(this->lastSnappedBounds.width);
+    out.writeDouble(this->lastSnappedBounds.height);
+
     out.writeDouble(this->relativeX);
     out.writeDouble(this->relativeY);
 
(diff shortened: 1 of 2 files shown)


Technius (approved):

Technius:
Actually, it might not be enough to just serialize it. We also have to account for zero `lastSnappedBounds` values in the undo action.

rolandlo:
> Actually, it might not be enough to just fix the (de)serialization. We also have to account for zero `lastSnappedBounds` values in the undo action.

Which undo action? With zero `lastSnappedBounds` the `scale` variable will be set to `false`, so no `ScaleUndoAction` is added.

rolandlo:
@Technius 
Nice catch. I haven't noticed before that undo/redo didn't work for scaling horizontal and vertical strokes (without any copying/pasting) involved. 
Can you please take another look at it.

PR #2859, opened 2021-02-23
Fix highlighter initial rendering extremely thin

### LINKED ISSUE #2858: Highlighter is not rendered properly while using it
**Describe the bug**
Using the highlighter, only a very thin line is drawn. Once the stroke is finished and a repaint happened, the storke is as is should be.

Screenshot_20210223_213613

This is most likely related to my PR #2512. 

Edit: It seems to only happen with pressure inference on.

personalizedrefrigerator:
## Summary
 * We weren't setting the tool type when drawing a segment of a stroke! As such, it defaulted to `STROKE_TOOL_PEN`, causing highlighter strokes to be treated like pen strokes (drawn based on pressure, etc.).

Should fix #2858

### DIFF
--- a/src/model/eraser/EraseableStroke.cpp
+++ b/src/model/eraser/EraseableStroke.cpp
@@ -33,9 +33,12 @@ void EraseableStroke::draw(cairo_t* cr) {
 
     double w = this->stroke->getWidth();
 
+    // Some tools ignore pressure sensitivity.
+    bool usePressure = this->stroke->getToolType() != STROKE_TOOL_HIGHLIGHTER;
+
     for (GList* l = tmpCopy->data; l != nullptr; l = l->next) {
         auto* part = static_cast<EraseableStrokePart*>(l->data);
-        if (part->getWidth() == Point::NO_PRESSURE) {
+        if (!usePressure || part->getWidth() == Point::NO_PRESSURE) {
             cairo_set_line_width(cr, w);
         } else {
             cairo_set_line_width(cr, part->getWidth());
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -141,6 +141,7 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
             lastSegment.addPoint(prevPoint);
             lastSegment.addPoint(point);
             lastSegment.setWidth(stroke->getWidth());
+            lastSegment.setToolType(stroke->getToolType());
 
             cairo_set_operator(crMask, CAIRO_OPERATOR_OVER);
             cairo_set_source_rgba(crMask, 1, 1, 1, 1);


bhennion:
This works fine, and it is probably the best way to fix this! Thanks a lot!!

(For posterity: the bug was introduced because the stabilizers always set the point's pressure value to the value of the incoming event. Before #2512, this value was defaulted to `Point::NO_PRESSURE` unless the `toolType` was the pen.)

rolandlo:
Try out the standard eraser on such a highlighter stroke. It looks strange.

personalizedrefrigerator:
> Try out the standard eraser on such a highlighter stroke. It looks strange.

Strange behavior on erase: Seems to switch to pen rendering.

It does. I'll try to look into it.

src/model/eraser/EraseableStroke.cpp
```diff
@@ -33,9 +33,12 @@ void EraseableStroke::draw(cairo_t* cr) {
 
     double w = this->stroke->getWidth();
 
+    // Some tools ignore pressure sensitivity.
+    bool usePressure = this->stroke->getToolType() != STROKE_TOOL_HIGHLIGHTER;
+
     for (GList* l = tmpCopy->data; l != nullptr; l = l->next) {
         auto* part = static_cast<EraseableStrokePart*>(l->data);
-        if (part->getWidth() == Point::NO_PRESSURE) {
+        if (!usePressure || part->getWidth() == Point::NO_PRESSURE) {
```
personalizedrefrigerator, line 41: This should fix the eraser issue, but I'm not very satisfied with the solution -- I would prefer that EraseableStroke not need to take into account the type of tool that made the stroke, if possible.

rolandlo:
> This works fine, and it is probably the best way to fix this! Thanks a lot!!
> 
> (For posterity: the bug was introduced because the stabilizers always set the point's pressure value to the value of the incoming event. Before #2512, this value was defaulted to `Point::NO_PRESSURE` unless the `toolType` was the pen.)

What is the idea of setting the point's pressure to something else than `Point::NO_PRESSURE` for highlighter strokes?

bhennion:
> What is the idea of setting the point's pressure to something else than `Point::NO_PRESSURE` for highlighter strokes?

It really is not a design but rather a flaw and a mistake. As is, the stabilizers are completeley unaware of the strokes properties.
They average everything they're given (position and pressure). They are just fed the events as they come, so here we go.

Fixing the highlighter problem could be done 

1. with @personalizedrefrigerator's proposal, with in mind the philosophy: whatever the context, when we draw a stroke, we should draw it differently if it's a highlighter.
2. We could also simply set the pressure to `Point::NO_PRESSURE` in `StrokeHandler::drawSegmentTo` (so after stabilizing)
3. or in the `StrokeHandler::onMotionNotify` (before passing the event to the stabilizer).

This raises another possible issue with the stabilizers: even if the input events have all `Point::NO_PRESSURE` values, those values will still get averaged, and with rounding errors and all that, may not longer be `== Point::NO_PRESSURE` in the end (I never saw it happen, but in theory it could, I think). Ideally, the stabilizers would not do anything on the pressure (to save flops) if it's not needed, but I don't know yet how to do that efficiently (i.e. not testing in every tiny loop). The may mean templating everything and feed it points without pressure. I'll create a separate issue for this.

The ideal solution would be, IMO, to do 3. together with fixing the above averaging issue. This may take time, so for now, maybe 1. or 2. are simpler.

bhennion:
I implemented a version of 2. as above in #2863.

personalizedrefrigerator:
Closing in favor of #2863. Thank you!

PR #2871, opened 2021-02-24
update msys2 to get cairo 1.17.4-2

### LINKED ISSUE #2601: Export to PDF creates corrupted PDFs with Xournal++ 1.0.20
**Affects versions :**
 - OS: Windows
 - Version of Xournal++: 1.0.20 / 1.0.19
 - Installation method: Official Windows Installer

**Describe the bug**
Xournal++ "Export to PDF" feature will create partially corrupted PDFs with Xournal++ 1.0.20.
Using Xournal++ 1.0.19 will export PDFs working correctly.

I compared the PDFs created by Xournal++ 1.0.19 and 1.0.20. They will be identical with exception of the date and the **FontName** and **BaseFont** entries

Examples from 1.0.19 (each **/FontName** will have a matching **/BaseFont** string not printed here)
```
/FontName /GZMBAL+LMRoman10-Regular
/FontName /HQLXHP+CMMI10
/FontName /IGGRRB+CMMI8
/FontName /TJYQRQ+CMR6
/FontName /ZBREYI+CMR10
/FontName /AWBOZU+CMR8
/FontName /WIZMEQ+Tahoma
```
Same locations in PDF file with Xournal++ 1.0.20
```
/FontName /;(5@A6+LMRoman10-Regular
/FontName /:16*:2+CMMI10
/FontName /9;;00@+CMMI8
/FontName /.8)101+CMR6
/FontName /(@0=)9+CMR10
/FontName /A+@3(-+CMR8
/FontName /+9(5=1+Tahoma
```

**Additional context**
- The 1.0.20 exported PDFs can be opened with some viewers (like *Chrome*, *SumatraPDF*)
- Official *Adobe Acrobat* will just give a "*there has been some error processing this page*" message and show nothing
- *Firefox* will render the page only until the point where the first annotation added with Xournal++ would show

I assume there is some set of characters *Acrobat* and *Firefox* expect when parsing font names, so they will fail loading the PDF if they encounter forbidden characters.

rolandlo:
Fixes #2601 by updating MSYS2 to include mingw-w64-x86_64-cairo-1.17.4-2 instead of mingw-w64-x86_64-cairo-1.17.4-1.

I just checked running the msys2-blob to check if the Cairo version gets really bumped to 1.17.4-2 (it does). 

**Edit:** The continuos integration pipeline shows the same result. The downside is, that it take about 15 minutes longer to build on Windows.

**Edit2:** I now replaced `pacman -Syuu` (twice) by `pacman -Sy` (once) to only update the package index without a full system upgrade.

### DIFF
--- a/azure-pipelines/steps/build_windows.yml
+++ b/azure-pipelines/steps/build_windows.yml
@@ -8,6 +8,7 @@ steps:
   # Copy of msys2-blob as a temporary fix for Windows CI
   - script: |
       set PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem
+      pacman --noconfirm -Sy
       pacman --noconfirm -S mingw-w64-x86_64-imagemagick mingw-w64-x86_64-ninja mingw-w64-x86_64-cppunit mingw-w64-x86_64-poppler mingw-w64-x86_64-gtk3 mingw-w64-x86_64-libsndfile mingw-w64-x86_64-libzip mingw-w64-x86_64-lua
     env:
       MSYS2_ARCH: x86_64


Technius:
It might be better to simply update the package index (`pacman -Sy`) instead of doing a full system upgrade (`pacman -Syuu` twice).

rolandlo:
Ok, let's try.

DUOLabs333:
@rolandlo This seems to have fixed the corrupted PDF export issues.

PR #2872, opened 2021-02-25
added shortcut for default tool

### LINKED ISSUE #2754: Add shortcut for Default Tool

I primarily use xournalpp through key bindings. It would be great if the "Default Tool" also had a shortcut (option: `Shift+Ctrl+D`).

This would be consistent with all the other primary tools. It is currently the only non-mapped tool (from Tools section in menu-bar), with the other non-mapped tools being audio-related: play object, record/stop, etc.

This request is different from  #919, which requires that keybindings are configurable. I simply ask that there is a default shortcut for the default tool. 

P.S sorry for the barrage of feature requests and bug reports all at the same time.

nicolae-stroncea:
Fixes #2754.
Added shortcut "Ctrl+Shift+D" for the default tool, which makes it accessible by keyboard.

I know that it's also possible to write a plugin for it, but imo this makes the UI more consistent since the majority of other tools(pen, eraser, etc) also have shortcuts, and if the default tool is considered important/popular it should have a built-in shortcut as well.

The only other main tools left that don't have shortcuts are: "Play Object" and "Record/Stop".

### DIFF
--- a/ui/main.glade
+++ b/ui/main.glade
@@ -1377,6 +1377,7 @@
                             <property name="label" translatable="yes">_Default Tools</property>
                             <property name="use-underline">True</property>
                             <signal name="activate" handler="ACTION_TOOL_DEFAULT" swapped="no"/>
+                            <accelerator key="d" signal="activate" modifiers="GDK_SHIFT_MASK | GDK_CONTROL_MASK"/>
                           </object>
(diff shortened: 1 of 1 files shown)


rolandlo:
I agree with you. Thanks for the PR.

PR #2873, opened 2021-02-25
Daily snapshot release for Fedora

luyatshimbalanga:
This spec file is set for daily snapshot with very minimal maintenance. The comment is added to describe the purpose. The benefit is bringing more Fedora testers to find possible bugs and test improvement. The only change is `version_string` set for development release i.e. 1.1.0.


rolandlo:
Thanks!

Issue #2866, opened 2021-02-24
Highlight currently playing object


nicolae-stroncea:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I use the "Play Object" tool, and as the audio changes to the next objects, I  do not know anymore what object the audio is related to. Therefore, I can't just click start on an object and follow along with the audio and what I was writing. Instead my workflow changes to clicking on  objects I'm interested in and hoping that they have some audio, which I would say is less efficient and requires a lot more user effort.

**Describe the solution you'd like**
* Once I start playing an object, it would be great to have the currently playing object highlighted, especially because the object will frequently change. A further useful enhancement would be to automatically change the page to the currently playing object, which might be useful if you want to sit back and relisten to your lecture alongside viewing your annotations.

Benefits:
* Following along as the highlighted object changes enhances learning and recollection as you now have an audio story of your progression through the lesson. This is great for reviewing notes later.
* When I click the "Forward" and "Backward" buttons, I can now see what those next objects are.

nicolae-stroncea:
@LittleHuba  could you reopen this?

#2874 highlights objects that *are* playable, but does not highlight the *currrently playing* object.

PR #2874, opened 2021-02-25
Fade all annotations without audio when clicking play object

nicolae-stroncea:
If you record audio selectively, when you click on the `Play Object` tool, you have to essentially click on random objects until you find one with audio.

This commit fades (alpha=50/255) all of the annotations that don't have audio whenever `Play Object` tool is enabled, making it easier to find the objects with audio. Annotations revert to full opacity once another tool is chosen.

**Regular page:**

image

**`Play object` is turned on**:

image

Only the bright green text on the left, and the shapes on the right have related audio.

Notes that don't require any action:
* Images are not AudioElements, therefore they are currently always faded.
* Highlights are also always faded. This is because it doesn't seem like they ever store audio information.
* I think both of the above are by *current* design.

Notes that may require action:
* TexImages are currently always opaque. I couldn't find how to make the Tex translucent too since i'm not too familiar with how poppler works, would appreciate some pointers. While not necessary, would be great to get them translucent too when the tool is activated.

### DIFF
--- a/src/view/DocumentView.cpp
+++ b/src/view/DocumentView.cpp
@@ -57,7 +57,7 @@ void DocumentView::drawStroke(cairo_t* cr, Stroke* s, int startPoint, double sca
     sv.paint(this->dontRenderEditingStroke);
 }
 
-void DocumentView::drawText(cairo_t* cr, Text* t) {
+void DocumentView::drawText(cairo_t* cr, Text* t) const{
     cairo_matrix_t defaultMatrix = {0};
     cairo_get_matrix(cr, &defaultMatrix);
     if (t->isInEditing()) {
@@ -65,13 +65,18 @@ void DocumentView::drawText(cairo_t* cr, Text* t) {
     }
 
     cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
-    applyColor(cr, t);
+    if(this->markAudioStroke && t->getAudioFilename().length() == 0){
+        applyColor(cr, t, AudioElement::opacityNoAudio);
+    }
+    else{
+        applyColor(cr, t);
+    }
 
     TextView::drawText(cr, t);
     cairo_set_matrix(cr, &defaultMatrix);
 }
 
-void DocumentView::drawImage(cairo_t* cr, Image* i) {
+void DocumentView::drawImage(cairo_t* cr, Image* i) const {
     cairo_matrix_t defaultMatrix = {0};
     cairo_get_matrix(cr, &defaultMatrix);
 
@@ -87,7 +92,11 @@ void DocumentView::drawImage(cairo_t* cr, Image* i) {
     cairo_scale(cr, xFactor, yFactor);
 
     cairo_set_source_surface(cr, img, i->getX() / xFactor, i->getY() / yFactor);
-    cairo_paint(cr);
+    if(this->markAudioStroke){
+        cairo_paint_with_alpha(cr, AudioElement::opacityNoAudio / 255.0);
+    } else{
+        cairo_paint(cr);
+    }
 
     cairo_set_matrix(cr, &defaultMatrix);
 }
--- a/src/view/StrokeView.cpp
+++ b/src/view/StrokeView.cpp
@@ -45,23 +45,27 @@ void StrokeView::changeCairoSource(bool markAudioStroke) {
     if (s->getFill() != -1 && s->getToolType() != STROKE_TOOL_HIGHLIGHTER) {
         cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
 
-        // Set the color and opacity
-        DocumentView::applyColor(cr, s, s->getFill());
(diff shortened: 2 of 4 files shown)


Technius:
Nice!

I think a major improvement would be to also fade out all non-audio elements when the Play Object tool is _selected_.

> I couldn't find how to make the Tex translucent

Maybe something along these lines?
```c++
cairo_push_group(cr); // switch to temporary surface
poppler_page_render(cr, page); // render page
cairo_pattern_t* texPattern = cairo_pop_group(cr); // return to original surface
cairo_set_source(cr, texPattern); // use temporary surface as pattern
cairo_paint_with_alpha(cr, alpha); // paint the temporary surface
cairo_pattern_destroy(texPattern); // discard the temporary surface
```

nicolae-stroncea:
Thanks @Technius,  your suggestion worked perfectly. TexElement now also grows translucent when the "Play Object" tool is enabled.

image


I wanted to clarify something as I think my PR name may be unclear:
> I think a major improvement would be to also fade out all non-audio elements when the Play Object tool is selected.

This is actually exactly what this PR is meant to fix. Essentially as long as the `Play Object` tool is selected, everything that doesn't have audio will fade away. Then once user chooses another tool, opacity for all elements returns to normal. It's meant to make it easier for the user to see the elements that carry audio so they don't click around randomly looking for them.

This PR doesn't fix #2866, as it doesn't actually highlight the currently playing element in any way and doesn't fix the issue raised there (at least to a good standard).

Fatih20:
This would be great for the software! I was creating the app new homepage and found it difficult to show how Xournal++ can record audio notes because objects with audio recording isn't highlighted.

personalizedrefrigerator:
I would like to have this in version 1.2, if possible. @nicolae-stroncea please rebase on `xournalpp:master` and [apply `clang-format`](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#use-clang-format). Thank you for your work!

<details>



image
image
image
image

</details>

src/view/DocumentView.cpp
```diff
@@ -116,7 +125,14 @@ void DocumentView::drawTexImage(cairo_t* cr, TexImage* texImage) {
 
         cairo_translate(cr, texImage->getX(), texImage->getY());
         cairo_scale(cr, xFactor, yFactor);
-        poppler_page_render(page, cr);
+        if(this->markAudioStroke){ // make non-Audio strokes translucent when markAudioStroke
+            cairo_push_group(cr); // switch to temporary surface
+            poppler_page_render(page, cr); // render page
+            cairo_pop_group_to_source(cr); // return to original surface and set temp surface as pattern
+            cairo_paint_with_alpha(cr, AudioElement::opacityNoAudio / 255.0); // paint the temporary surface w/ opacity level
```
personalizedrefrigerator, line 132: Thank you for adding comments! I think this would be more readable if the comments were on a separate line from the code.
```suggestion
        // make non-Audio strokes translucent when markAudioStroke
        if(this->markAudioStroke){
            // Switch to a temporary surface, render the page, then switch back. This sets the
            // temporary surface to the current pattern.
            cairo_push_group(cr);
            poppler_page_render(page, cr);
            cairo_pop_group_to_source(cr);
            
            // paint the temporary surface w/ opacity level
            cairo_paint_with_alpha(cr, AudioElement::opacityNoAudio / 255.0);
```

src/view/StrokeView.cpp
```diff
@@ -42,25 +42,29 @@ void StrokeView::changeCairoSource(bool markAudioStroke) {
     ///////////////////////////////////////////////////////
     // Fill stroke
     ///////////////////////////////////////////////////////
+    bool audioStrokebutNoAudio = markAudioStroke && s->getAudioFilename().length() == 0;
```
bhennion, line 45: This variable's name is a bit confusing, I think. Maybe a simpler `fadeOutStroke` would be clearer?

bhennion (changes requested):
This is great! It obviously needs to be rebased and reviewed again after that.

One bug still bothers me on pressured strokes (it was already in `master`, but should be fixed in this PR, I think):
sshot-2
In this screenshot,the two strokes above do not have audio recordings attached and the Play Audio tool is on.

This is due to the strokes segments being painted separately. The intersection of two consecutive segments is then darker than the rest. It gives a weird thing on the denser parts of the stroke.

A solution would be to use `cairo_push_group`, paint the stroke (with `s->getFill()` as alpha), and  then `cairo_pop_group_to_source`and `cairo_paint_with_alpha(cr, AudioElement::opacityNoAudio / 255.0);` (like you did for Tex formulas).

bhennion:
Rebased on master, applied clang format and various nitpicks (including @personalizedrefrigerator's comments).

Merging into master in 24h if no objections are raised.

LittleHuba:
No need to lengthen the period before merging, the changes just add some comments and change that variable name. You can batch commit them right before merging. The rest is good with me.

PR #2885, opened 2021-02-27
Fix Backup on Save Mechanism

### LINKED ISSUE #2883: X++ changes underlying pdf name resulting in background PDF could not be found
**Affects versions :**
 - OS: Linux
 - (Linux only) Desktop environment: Ubuntu 20.04,  Gnome X11
 - Which version of libgtk do you use: libgtk: 3.24.20
 - Version of Xournal++: 
     - Xournal++ 1.1.0+dev
 - Installation method: From source

**Describe the bug**
When a crash happens, if you restore and save the .xopp file with the same original(default suggested) .xopp name, the underlying pdf `original.pdf` is renamed to `original.pdf~` which results in the pdf not being found anymore with UI error: "The background PDF could not be found".

**To Reproduce**
1. Open a pdf file.
2. Draw a stroke, do not save.
3. Simulate a crash by sending `kill -11 <process pid` 
4. Reopen X++, you get a dialog "Xournal++ crashed last time. Would you like to restore the last edited file?".  Choose "Restore file".
6. Save with the default settings, i.e keep suggested name. 
7. Check in your file explorer that the original original pdf is now renamed  with a `~` appended at the end. Next time you open the .xopp file, you will not see the background PDF anymore as the background pdf has been renamed.

**Expected behavior**
Original pdf name is not changed
**Screenshots of Problem**
image

idotobi:
This PR fixes #2883

After the fix:
Backup will be created for the filepath that will be written to while saving.
Note backup here means aka. moving to `filename` + `~`.

Before the fix this was not always the case as non `.xopp` extensions where overwritten to `.xopp` upon saving, but only after the backup was already created for to original filepath.

### DIFF
--- a/src/control/jobs/SaveJob.cpp
+++ b/src/control/jobs/SaveJob.cpp
@@ -90,17 +90,19 @@ auto SaveJob::save() -> bool {
     fs::path const filepath = doc->getFilepath();
     doc->unlock();
 
+    auto const target = fs::path{filepath}.replace_extension(".xopp");
+
     if (doc->shouldCreateBackupOnSave()) {
         try {
-            Util::safeRenameFile(filepath, fs::path{filepath} += "~");
+            // Note: The backup must be created for the target as this is the filepath
+            // which will be written to. Do not use the `filepath` variable!
+            Util::safeRenameFile(target, fs::path{target} += "~");
         } catch (fs::filesystem_error const& fe) {
             g_warning("Could not create backup! Failed with %s", fe.what());
             return false;
         }
         doc->setCreateBackupOnSave(false);
     }
-
-    auto const target = fs::path{filepath}.replace_extension(".xopp");
 
     doc->lock();
     h.saveTo(target, this->control);


idotobi:
For future reference: The easiest way to easily reproduce emergency recoveries is to create an `emergencysave.xopp` in `.config/xournalpp/` via `killall -11 xournalpp`.

Then

```bash
cd .config/xournalpp
mv emergencysave.xopp emergencysave.xopp_backup
```

And then simply copying the backup back in place whenever needed before starting xournalpp

```bash
cd .config/xournalpp
cp emergencysave.xopp_backup emergencysave.xopp
```

Technius (approved):

Febbe:
Mhh weird. The filename variable shouldn't end on .pdf; only on a xournal/xournalpp extension.
After this change, we can't create backups from the old xournal format.

This may be irrelevant, since we only store as *.xopp. So the old file won't be overridden.

idotobi:
@Febbe : True, usually `filepath` should be a `.xopp` file. The problem occurs when you open a `.pdf` directly with xournalpp. Then it is not and it seems that https://github.com/xournalpp/xournalpp/pull/2885/files#diff-4166f0b7dc1905a808adc6c926a9d4d79fa5ee9225dfb61e15e0631a2593e4afL103 was explicitly written to mitigate this case.

My first attempt was to basically just include a new check for `if(filepath.extension() == ".pdf")` and then skip the backup, but I would worry that if there will ever be another file format we allow as background this would be overseen resulting in a regression risk :see_no_evil: 

Apart from this, it kind of makes sense to me that the Backup mechanism could be specified as follows:

When a user is saving a Document, Xournalpp createas a backup of the previous file version which would be overwritten otherwise. The backup is created by moving the previous file version from `<filepath>` to `<fileapath>~` right before saving the new version.

This would be in line with:

> This may be irrelevant, since we only store as *.xopp. So the old file won't be overridden.

Do you think this is fine or is there any other behaviour that you can think of that would depend on the previous implementation/behaviour? :thinking:

Febbe:
I think it will be fine for now, but better would a check, which checks if the save name is equal to the actual. This will guard every future filetype.

Issue #2889, opened 2021-02-28
Moving selections is unstable/highly sensitive when selection is only partially in view


nicolae-stroncea:
**Affects versions** :
* OS: Linux
* Desktop environment: Ubuntu 20.04, Gnome X11
* libtk: libgtk: 3.24.20
* Xournal version: Xournal++ 1.1.0+dev
 * Installation method: From source

**Describe the bug**

Moving a selection (a selected image, group of shapes, text, etc) throughout the page is unstable if you move a selection that is only partially on display(with the rest of it being out of your current view but still on the same page). For the best example of this consider when the  zoom level is more than 100%. 

By unstable I mean that the selection moves extremely quickly, maybe about 6-7 pages per second, if you even _lightly/accidentally_ move some selected region which is not fully on display _in the direction of where the rest of the selection is(i.e rest of selection is above, and you move selection slightly up._ This is especially inconvenient and disorienting when you make a large selection, which makes the issue much more likely to occur.

**Expected behavior**

Selections are moved **significantly slower** for this case.

**Screenshots of Problem**

Below is a screencast of 2 examples. For reference, the selection moves quicker than shown here, where it seems slowed down due to framerate.
[webcast_of_issue.zip](https://github.com/xournalpp/xournalpp/files/6058203/webcast_of_issue.zip)

Photon89:
I can confirm and have the same issue. I guess, the algorithm starts scrolling when the upper or lower edge of the object comes into the vicinity of the upper or lower edge of the visible part of the page, but when the object is already at (or even over) the edge, the scrolling is triggered immediately and doesn't stop as the object keeps being over the edge.

Proposal for a fix: If the object's edge is already past the visible page's edge, start scrolling but don't move the object till its edge gets visible. Once the object's edge is visible, switch to usual mode.

personalizedrefrigerator:
See
https://github.com/xournalpp/xournalpp/blob/5e78a44f8668418ad34468b2e9bfa27f9a61740e/src/gui/inputdevices/PenInputHandler.cpp#L416-L449

nicolae-stroncea:
Wanted to add a further detail. If you hit "Undo" once this happens, X++ freezes up for a couple of seconds. It doesn't crash and recovers but you get a long stack trace with warnings

tattsan:
Confirmed. I also observed that unintentional copies are created.
Perhaps these are the mentioned objects which the backtrace says 
``Could not remove element from layer, it's not on the layer!``

[...]

PR #2893, opened 2021-03-01
Fix panning speed when selection is on the edge of the screen

Technius:
This PR is a hack for fixing #2889. Currently a proof of concept.

In particular, this replaces the hacky edge panning logic with less terrible (but still bad) timer-based logic: instead of scrolling to "ensure that the selection is visible," the logic has been changed to "scroll up/down at a fixed speed when the selection is slightly out of visible page bounds"

* [x] Do not pan when selection is at the top/bottom of document
* [x] Fix selection position jitter when moving the mouse during panning
* [x] Implement horizontal panning
* [x] Add settings for adjusting panning speed
* [x] Fix selection teleporting when moving across pages. This is caused by the action not resetting when a page is crossed; the current event system assumes that any given action occurs on the same page.

As a workaround for jitter and teleporting, panning will block the next mouse movement from causing the selection to move. The teleportation is fixed entirely, while the jitter is reduced by a significant amount (there is still some present but it is only barely noticeable).

### DIFF
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -740,6 +740,7 @@ bool EditSelection::handleEdgePan(EditSelection* self) {
 
 
     Layout* layout = gtk_xournal_get_layout(self->view->getXournal()->getWidget());
+    const Settings* const settings = self->getView()->getXournal()->getControl()->getSettings();
     const double zoom = self->view->getXournal()->getZoom();
 
     // Helper function to compute scroll amount for a single dimension, based on visible region and selection bbox
@@ -754,7 +755,7 @@ bool EditSelection::handleEdgePan(EditSelection* self) {
         double mult = 0.0;
 
         // Calculate bonus scroll amount due to proportion of selection out of view.
-        const double maxMult = 5.0;
+        const double maxMult = settings->getEdgePanMaxMult();
         int panDir = 0;
         if (aboveMax) {
             panDir = 1;
@@ -765,7 +766,7 @@ bool EditSelection::handleEdgePan(EditSelection* self) {
         }
 
         // Base amount to translate selection (in document coordinates) per timer tick
-        const double panSpeed = 20.0;
+        const double panSpeed = settings->getEdgePanSpeed();
         const double translateAmt = visLen * panSpeed / (100.0 * PAN_TIMER_RATE);
 
         // Amount to scroll the visible area by (in layout coordinates), accounting for multiplier
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -30,9 +30,11 @@
 
 using std::vector;
 
-constexpr size_t MINPIXSIZE = 5;  // smallest can scale down to, in pixels.
+/// Smallest can scale down to, in pixels.
+constexpr size_t MINPIXSIZE = 5;
 
-constexpr int DELETE_PADDING = 20;  // ui button padding
+/// Padding for ui buttons
+constexpr int DELETE_PADDING = 20;
 constexpr int ROTATE_PADDING = 8;
 
 EditSelection::EditSelection(UndoRedoHandler* undo, const PageRef& page, XojPageView* view):
@@ -697,10 +699,7 @@ void EditSelection::moveSelection(double dx, double dy) {
     this->view->getXournal()->repaintSelection();
 }
 
-/**
- * If the selection is outside the visible area correct the coordinates
- */
-void EditSelection::ensureWithinVisibleArea() {
+auto EditSelection::getBoundingBoxInView() const -> Rectangle<double> {
     int viewx = this->view->getX();
     int viewy = this->view->getY();
     double zoom = this->view->getXournal()->getZoom();
@@ -714,11 +713,15 @@ void EditSelection::ensureWithinVisibleArea() {
     double minx = cx - w / 2.0;
     double miny = cy - h / 2.0;
 
+    return {viewx + minx * zoom, viewy + miny * zoom, w * zoom, h * zoom};
+}
+
+void EditSelection::ensureWithinVisibleArea() {
+    const Rectangle<double> viewRect = this->getBoundingBoxInView();
     // need to modify this to take into account the position
     // of the object, plus typecast because XojPageView takes ints
-    this->view->getXournal()->ensureRectIsVisible(static_cast<int>(viewx + minx * zoom),
-                                                  static_cast<int>(viewy + miny * zoom), static_cast<int>(w * zoom),
-                                                  static_cast<int>(h * zoom));
+    this->view->getXournal()->ensureRectIsVisible(static_cast<int>(viewRect.x), static_cast<int>(viewRect.y),
+                                                  static_cast<int>(viewRect.width), static_cast<int>(viewRect.height));
 }
 
 /**
--- a/src/control/tools/EditSelection.cpp
+++ b/src/control/tools/EditSelection.cpp
@@ -9,6 +9,7 @@
 #include "gui/PageView.h"
 #include "gui/XournalView.h"
 #include "gui/XournalppCursor.h"
+#include "gui/widgets/XournalWidget.h"
 #include "model/Document.h"
 #include "model/Element.h"
 #include "model/Layer.h"
@@ -36,6 +37,9 @@ constexpr size_t MINPIXSIZE = 5;
 /// Padding for ui buttons
 constexpr int DELETE_PADDING = 20;
 constexpr int ROTATE_PADDING = 8;
+
+/// Number of times to trigger edge pan timer per second
+constexpr unsigned int PAN_TIMER_RATE = 30;
 
 EditSelection::EditSelection(UndoRedoHandler* undo, const PageRef& page, XojPageView* view):
         snappingHandler(view->getXournal()->getControl()->getSettings()),
@@ -162,6 +166,11 @@ EditSelection::~EditSelection() {
 
     this->view = nullptr;
     this->undo = nullptr;
+
+    if (this->edgePanHandler) {
+        g_source_destroy(this->edgePanHandler);
+        g_source_unref(this->edgePanHandler);
+    }
 }
 
 /**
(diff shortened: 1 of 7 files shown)


redweasel:
As you are already at it, can you make it scroll with a speed proportional to the relative covered area of the object? Also an option for the scrolling speed in the preferences would be preferred over a const in the code.
Remember, nothing more permanent than a temporary fix.

Technius:
I still have to implement the configuration settings, but this should be mostly ready. This seems to work reasonably well for me now.

rolandlo:
I have tested the new scrolling, but not looked at the code yet. It works quite well for me, too. I'm not 100% convinced that a configuration setting is needed, since the user has some control over the scrolling speed as well and since every preference has a cost.

What I find a bit confusing though is that I can move the selection slightly over one edge (say the bottom edge) and then when pulling slightly in the other direction (north) it goes in direction of the edge (south) and not in the direction I actually pull. This confusing behaviour stops if I pull harder, but I may not realize that I need to do so.

Febbe:
> since every preference has a cost

Interesting, how far the reach of some Microsoft bloggers are, did we read the same article?😂

rolandlo:
> > since every preference has a cost
> 
> Interesting, how far the reach of some Microsoft bloggers are, did we read the same article?joy

Microsoft blogger? I read [this article](https://www.omgubuntu.co.uk/2021/07/why-gnome-does-what-gnome-does-by-tobias-bernard) written by a Gnome dev.

Febbe:
Oh right, meant that article. But a Microsoft dev stated the same thing. (That's why they removed the startsound)

Technius:
> I'm not 100% convinced that a configuration setting is needed, since the user has some control over the scrolling speed as well and since every preference has a cost.

Yes, the defaults seem fine, but there's the possibility that some users may find them too fast or slow. I would expect at least the max scroll speed and scroll speed scaling to be configurable.

> What I find a bit confusing though is that I can move the selection slightly over one edge (say the bottom edge) and then when pulling slightly in the other direction (north) it goes in direction of the edge (south) and not in the direction I actually pull. This confusing behaviour stops if I pull harder, but I may not realize that I need to do so.

Interesting to hear this viewpoint; I find the current behavior intuitive. I treat the pulling behavior as _acceleration_ and _deceleration_: pulling in the same direction accelerates the panning, while pulling away decelerates the panning. If I want to pan in the opposite direction, I would move the selection upwards. I feel like this behavior is closer to the use case of actually moving the selection, since I typically only want to pan the screen in some direction if I want to place the selection at a location in the same direction.

rolandlo:
> Interesting to hear this viewpoint; I find the current behavior intuitive. I treat the pulling behavior as _acceleration_ and _deceleration_: pulling in the same direction accelerates the panning, while pulling away decelerates the panning. If I want to pan in the opposite direction, I would move the selection upwards. I feel like this behavior is closer to the use case of actually moving the selection, since I typically only want to pan the screen in some direction if I want to place the selection at a location in the same direction.

Not sure I completely understood what you mean. I will add a video recording to make my point. 

https://user-images.githubusercontent.com/40485433/126900671-ed0c8f19-3e87-4053-ac43-070a8cf0ee64.mp4

The first 2-3 seconds are irrelevant. After roughly 3 seconds into the video, the selection is not moving at all. It is slightly extending over the lower edge of the window. Then I pick the selection with the mouse and pull upwards, but the selection moves downwards. It's not decceleration.

> Yes, the defaults seem fine, but there's the possibility that some users may find them too fast or slow. I would expect at least the max scroll speed and scroll speed scaling to be configurable.

Ok, go ahead. I can live with that.

src/control/tools/EditSelection.cpp
```diff
@@ -690,17 +704,134 @@ void EditSelection::moveSelection(double dx, double dy) {
...
+    } else if (afterRight) {
+        xScrollMult += (bboxRight - visRectRight) / bbox.width * MAX_MULT;
+    }
+
+    // How much we'll move the selection and layout.
+    const double SCROLL_SPEED = 5.0;
+    double scrollX = scrollXDir * (SCROLL_SPEED * xScrollMult);
+    double scrollY = scrollYDir * (SCROLL_SPEED * yScrollMult);
+    double layoutScrollX = zoom * scrollX;
+    double layoutScrollY = zoom * scrollY;
```
rolandlo, line 787: This part of the code confuses me a lot. I don't quite get why `yScrollMult` and `xScrollMult` should be initialized with `1.0`. Should those values stand for the x- and y-relative proportions of the selection out of view? Then these values would be `0.0` under normal circumstances; and if the selection is not moved completely out of view, the compultation would have to yield values between `0.0` and `1.0`.
It should be made 100% clear what `SCROLL_SPEED` and `MAX_MULT` mean. I would interpet these names more according to this suggestion:
```suggestion
    // The maximal speed in pt per update, with which the selection is moved, when it 
    // intersects or at least touches the view.
    const double SCROLL_SPEED = 8.0;
    // The maximal number, with with the scrolling speed can be multiplied, when the 
    // selection is moved far out of view.
    const double MAX_MULT = 3.0;
    
    // Compute relative distance in y- and x-direction of the farthest selection edge 
    // from the view. These values will be between 0.0 and 1.0 when the selection 
    // intersects or touches the view and are bounded by MAX_MULT
    double yScrollMult = 0.0;
    double xScrollMult = 0.0;
 
    if (aboveTop) {
        std::min(yScrollMult = (visRect.y - bbox.y) / bbox.height, MAX_MULT);
    } else if (belowBottom) {
        std::min(yScrollMult = (bboxBot - visRectBot) / bbox.height, MAX_MULT);
    }

    if (beforeLeft) {
        std::min(xScrollMult = (visRect.x - bbox.x) / bbox.width, MAX_MULT);
    } else if (afterRight) {
        std::min(xScrollMult = (bboxRight - visRectRight) / bbox.width, MAX_MULT);
    }

    double scrollX = scrollXDir * (SCROLL_SPEED * xScrollMult);
    double scrollY = scrollYDir * (SCROLL_SPEED * yScrollMult);
    double layoutScrollX = zoom * scrollX;
    double layoutScrollY = zoom * scrollY;```

File: src/control/tools/EditSelection.cpp [hunk 2]
```diff
@@ -690,17 +704,134 @@ void EditSelection::moveSelection(double dx, double dy) {
...
-
     updateMatrix();
 
     this->view->getXournal()->repaintSelection();
 }
 
+void EditSelection::setEdgePan(bool pan) {
+    if (pan && this->edgePanHandler == nullptr) {
+        this->edgePanHandler = g_timeout_source_new(1000 / 30);
+        g_source_set_callback(this->edgePanHandler, reinterpret_cast<int (*)(void*)>(EditSelection::handleEdgePan),
```
rolandlo, line 715: Why not simply like this?
```suggestion
        g_source_set_callback(this->edgePanHandler, reinterpret_cast<GSourceFunc>(EditSelection::handleEdgePan),
```

src/control/tools/EditSelection.cpp
```diff
@@ -690,17 +704,134 @@ void EditSelection::moveSelection(double dx, double dy) {
...
 
-    ensureWithinVisibleArea();
-
     updateMatrix();
 
     this->view->getXournal()->repaintSelection();
 }
 
+void EditSelection::setEdgePan(bool pan) {
+    if (pan && this->edgePanHandler == nullptr) {
```
rolandlo, line 713:
```suggestion
    if (pan && !this->edgePanHandler) {
```

rolandlo (changes requested):
Some comments about the code. Although it works pretty OK, I feel it should be made more readable what concerns the scrolling computation.

src/control/tools/EditSelection.cpp
Technius: GSourceFunc is not available until glib 2.58.

Technius:
I rebased and added the preference settings. I will update the code for clarity later today.

src/control/tools/EditSelection.cpp
Technius: As the panning is currently implemented, the pan amount is calculated as `base amount + bonus amount from out-of-view`. We can definitely change it to just scale linearly instead. I'll have to play around with it to see which feels better first.

Technius:
> The first 2-3 seconds are irrelevant. After roughly 3 seconds into the video, the selection is not moving at all. It is slightly extending over the lower edge of the window. Then I pick the selection with the mouse and pull upwards, but the selection moves downwards. It's not decceleration.

This is occurring because the element is not fully visible when the first selection ends. Hence, the second selection immediately triggers panning. We could try to avoid this problem by snapping the scroll to ensure that the element is fully visible.

src/control/tools/EditSelection.cpp
rolandlo: You probably mean [`G_SOURCE_FUN`](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#G-SOURCE-FUNC:CAPS). We use `reinterpret_cast<GSourceFunc>` already in various other source files.

Technius:
Readability should be greatly improved now, but there are still a few minor changes I want to make.

src/control/tools/EditSelection.cpp
Technius: You're right. Fixed.

rolandlo:
It works really very nicely now and the readability is greatly improved. The only confusing/misleading thing now:
Say `MAX_MULT = 5`. From the description 
`Multiply the panning speed by up to this amount, based on how much of the selected element is out of view.` in the glade file I would assume that the panning speed can be up to 5 times as fast as the base panning speed. In reality it can be up to 6 times as fast, since this amount is added to the base panning speed.

For extreme case `MAX_MULT=0` the description would say that there can be no panning (since the panning speed is at most 0), whereas it is just the base panning speed.

Technius:
I finally had some time to compare adding `1.0` to mult and adding nothing to mult. The former feels better in my opinion because it allows the user to pan faster with more context, but the latter allows the user to control the panning more precisely. To make the mechanism easier to understand, I just went with the latter.

rolandlo (approved):
From my perspective everything looks fine now.

Technius:
Doing some preliminary rebasing, will finish later today.

Technius:
Rebased onto master and squashed into 5 commits. Merging in 24 hours if there are no objections.

PR #2902, opened 2021-03-04
Fix several LaTeX-dialog-related memory bugs

personalizedrefrigerator:
Most of the issues: `poppler_document_get_page()` results were not `g_object_unref`ed.
Others:
 * Process object wasn't unref'ed in `LatexController`
 * On LaTeX dialog close, a null `GCancellable` object pointer could be unreffed, causing a warning.
 * `TexImage::clone()` set its clone's `pdf` pointer to its own PDF pointer, without increasing its reference count.


Technius:
LGTM, thanks.

Issue #2823, opened 2021-02-18
ARM Architecture release


gaperez64:
The current Mac release immediately crashes for me after trying to run it. I suppose it is because of the new architecture (yes, I am running an M1 chip). I compiled from source using the convenient homebrew solution and that indeed works.

I propose to add a release targeting the new architecture for convenience (and for uses who cannot or do not want to compiler Xournal++ themselves).

Additionally, it would be nice to have a build-app script which works on top of the homebrew solution since the current one only works if you use the jhbuild recipe (at least out of the box... it's straightforward to modify it I think).

Technius:
#2390 switches to a macports based build, which should be nicer. I'm not sure if we can do a native ARM build--I don't think Azure Pipelines supports ARM builds for Mac. It's hard for us to support Macs right now because we don't have Mac devices.

gaperez64:
I can try the macports solution once it's merged. For the ARM-based Mac release, I agree that the best would be to wait until Azure pipelines supports the architecture so that it's just a new item in the YAML file.

gaperez64:
I am happy to try and help set up a minimal working version of Xournal++ using homebrew. To this end, here is some more feedback on my side:

- I got the [homebrew version](https://github.com/xournalpp/xournalpp/blob/master/readme/MacBuild.md#xournal-mac-homebrew-build) to compile, though it required to remove "cairo/" from all <cairo/*> includes
- I can run xournalpp after that but I have already found two issues with it:

1. It crashes when I try to export the current document as a pdf. I am attaching both the terminal messages and the crash's log.
[crash.log](https://github.com/xournalpp/xournalpp/files/6027382/crash.log)
[crash-exportpdf.pdf](https://github.com/xournalpp/xournalpp/files/6027384/crash-exportpdf.pdf)
2. When writing using my Wacom tablet (Wacom One) the lines are "squiggly", that is less straight than if I use the builtin trackpad. I am attaching an image of their comparison. There is no special/particular output on the terminal nor do I see error message when writing though so I am at a loss as to what other information I can provide for this. 
<img width="760" alt="Screenshot 2021-02-23 at 08 40 46" src="https://user-images.githubusercontent.com/2659561/108815108-2af20800-75b4-11eb-9a79-60cf3889dad1.png">

gaperez64:
I pulled the most recent source code and recompiled as described above, now the second problem is no longer present :) No idea which recent changes fixed this though. Problem 1 is still present.

[...]

PR #2906, opened 2021-03-04
Fix Crash on permission failure

Febbe:
- replaced `fs::equivalent` with 2 more robust calls to `fs::weakly_canonical`
 - catches an error on failure.

### DIFF
--- a/src/control/jobs/BaseExportJob.cpp
+++ b/src/control/jobs/BaseExportJob.cpp
@@ -29,14 +29,22 @@ void BaseExportJob::addFileFilterToDialog(const string& name, const string& patt
 auto BaseExportJob::checkOverwriteBackgroundPDF(fs::path const& file) const -> bool {
     auto backgroundPDF = control->getDocument()->getPdfFilepath();
     // If there is no background, we can return
-    if (!fs::exists(backgroundPDF)) {
-        return true;
-    }
-    // If the new file name (with the selected extension) is the previously selected pdf, warn the user
-    if (fs::equivalent(file, backgroundPDF)) {
-        string msg = _("Do not overwrite the background PDF! This will cause errors!");
-        XojMsgBox::showErrorToUser(control->getGtkWindow(), msg);
-        return false;
+    try {
+        if (!fs::exists(backgroundPDF)) {
+            return true;
+        }
+        // If the new file name (with the selected extension) is the previously selected pdf, warn the user
+
+        if (fs::weakly_canonical(file) == fs::weakly_canonical(backgroundPDF)) {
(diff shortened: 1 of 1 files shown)


Febbe:
will merge in 24h if pipeline succeeds and if there are no objections
@Technius @rolandlo

Technius (commented):
Can you elaborate on where this "permission failure" is coming from?

rolandlo:
Will this fix https://github.com/xournalpp/xournalpp/issues/2823#issuecomment-783982789?

Febbe:
> Will this fix [#2823 (comment)](https://github.com/xournalpp/xournalpp/issues/2823#issuecomment-783982789)?

Very likely. I still don't know when and why this happens. For me, it seems like a permission failure. But the error message (unsupported operation) was not very specific, and it occurred in 90% of all cases.

createyourpersonalaccount:
@Febbe maybe it has to do with SIP protection? https://support.apple.com/en-us/HT204899
Here's a github issue from a different project that may be related to what was happening: https://github.com/gulrak/filesystem/issues/77

Technius:
We actually use that library for file system handling on Mac, but we're too close to our next release to bump up a dependency version and risk regressions. I propose just having this band-aid fix for now and bumping the library version after we're done with the 1.1.0 release.

Febbe:
It's also a problem on linux mint 20(Ubuntu 20.04)

gaperez64:
> > Will this fix [#2823 (comment)](https://github.com/xournalpp/xournalpp/issues/2823#issuecomment-783982789)?
> 
> Very likely. I still don't know when and why this happens. For me, it seems like a permission failure. But the error message (unsupported operation) was not very specific, and it occurred in 90% of all cases. So I just decided to use a similar approach which is a lot more safe, but a bit more expensive. Since it is not in a loop, this is fine.

I can confirm that with the change to `fs::weakly_canonical` I am unable to reproduce the error.  To be precise, the following code does not give the error anymore:
```cpp
#include <iostream>
#include <cstdint>
#include <filesystem>
#include <cassert>

namespace fs = std::filesystem;

int main(int argc, char* argv[])
{
    assert(argc >= 3); 
    // hard link equivalency
    fs::path p1 = argv[1];
    fs::path p2 = argv[2];
    // if (fs::equivalent(p1, p2)) {
    if (fs::weakly_canonical(p1) == fs::weakly_canonical(p2)) {
        std::cout << p1 << " is equivalent to " << p2 << '\n';
        return 1;
    } else {
        std::cout << p1 << " is NOT equivalent to " << p2 << '\n';
        return 0;
    }   
}
```

PR #2907, opened 2021-03-05
Fix crash on scroll with sidebar preview open

### LINKED ISSUE #2887: Crash when scrolling through PDF with sidebar open on Layer/Layerstack Preview
**Affects versions** :
* OS: Linux
* Desktop environment: Ubuntu 20.04, Gnome X11
* libtk: libgtk: 3.24.20
* Xournal version: Xournal++ 1.1.0+dev
 * Installation method: From source

**To Reproduce**

1. Open PDF with X++.
2. Open sidebar on **Layer preview**.
3. Scroll through PDF. *Note*: It typically crashes within 1-10 seconds of continuous scrolling.
4. X++ crashes.

Random PDF I found online to test on, but I think this problem occurs on any pdf.
[Report_Writing_Process_principles_and_styles.pdf](https://github.com/xournalpp/xournalpp/files/6057657/Report_Writing_Process_principles_and_styles.pdf)

**Additional Details**
* I've gotten this crash many times in the past but this is the first time I was able to find a reproducible patern for it.
* Don't think it occurs for 1.0.20

personalizedrefrigerator:
A `PreviewJob`'s `SidebarPreviewBaseEntry` could be deleted before the `PreviewJob` could run (when scrolling quickly). This is handled by adding an `OnDestroyListener` to the `SidebarPreviewBaseEntry` class.

This is a bit of a band-aid fix.

Should fix #2887.

### DIFF
--- a/src/control/jobs/PreviewJob.cpp
+++ b/src/control/jobs/PreviewJob.cpp
@@ -3,21 +3,34 @@
 #include "control/Control.h"
 #include "gui/Shadow.h"
 #include "gui/sidebar/previews/base/SidebarPreviewBase.h"
-#include "gui/sidebar/previews/base/SidebarPreviewBaseEntry.h"
 #include "gui/sidebar/previews/layer/SidebarPreviewLayerEntry.h"
 #include "model/Document.h"
 #include "view/DocumentView.h"
 #include "view/PdfView.h"
 
-PreviewJob::PreviewJob(SidebarPreviewBaseEntry* sidebar): sidebarPreview(sidebar) {}
+PreviewJob::PreviewJob(SidebarPreviewBaseEntry* sidebar): sidebarPreview(sidebar) {
+    g_object_ref(sidebar->widget);
 
-PreviewJob::~PreviewJob() { this->sidebarPreview = nullptr; }
+    previewDestroyListenerID = sidebarPreview->addOnDestroyListener([&]() { this->sidebarPreview = nullptr; });
+}
+
+PreviewJob::~PreviewJob() {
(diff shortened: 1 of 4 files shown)


rolandlo:
Great! I can confirm that it doesn't crash any more with this PR.

Technius (commented):
I guess this fixes the issue, but this fix seems like way overkill.

It might be better to just remove the preview job from the scheduler when the sidebar base entry is destroyed. See `XournalScheduler::removeSource`, where the source of a `PreviewJob` is `this->sidebarPreview` (i.e. just pass `this` when calling from `SidebarPreviewBaseEntry`).

The scheduler/job system is probably due for a rewrite but that's something to save for after the 1.1.0 release.

personalizedrefrigerator:
> It might be better to just remove the preview job from the scheduler when the sidebar base entry is destroyed. See `XournalScheduler::removeSource`, where the source of a `PreviewJob` is `this->sidebarPreview` (i.e. just pass `this` when calling from `SidebarPreviewBaseEntry`).

I think it's already trying to do this:

https://github.com/xournalpp/xournalpp/blob/5d2683662f560b5254d52973b9f4281036e43e24/src/control/jobs/XournalScheduler.cpp#L14-L18

https://github.com/xournalpp/xournalpp/blob/5d2683662f560b5254d52973b9f4281036e43e24/src/gui/sidebar/previews/base/SidebarPreviewBaseEntry.cpp#L30-L35

personalizedrefrigerator:
Closing in favor of #2955.

PR #2927, opened 2021-03-09
Macapp

Febbe:
pipeline version of macports


Febbe:
@fredowski is there a way to reduce the pipeline overhead, so we can use your macports buildscripts in the production pipelines?

fredowski:
Hi, sorry for not continuing this work - I am quite busy at the moment. I am not too sure what your plan is. Building all dependencies from macports takes quite a while. So I would not do this in CI. Is that your question?

Febbe:
wont work, since macports requires too much time to install

PR #2934, opened 2021-03-11
CMake: Make X11 an optionnal dependency

lanodan:
This was tested on a libX11-less wayland linux and another linux system with X11+Wayland.


Febbe:
X11 should be removed in the future, or at least added as a module.

PR #2949, opened 2021-03-14
fix localizations for Azure pipelines builds

### LINKED ISSUE #2945: localizations don't work in released .deb packages
**Affects versions :**
 - OS: Linux

**Describe the bug**
After installing an officially relased .deb package (built on Azure Pipelines) localizations don't work, i.e. Xournal++ doesn't have the right path for the localizations and setting the language via `language` panel in the preferences is not possible. 

**To Reproduce**
1. Download one of the .deb-packages from the [releases page](https://github.com/xournalpp/xournalpp/tags)
2. Install via `dpkg -i PACKAGE_NAME`
3. Start xournalpp and try to change the language in the preferences

**Expected behavior**
Localizations should work out of the box like they do with the `.deb` packages from the PPA

**Additional context**
The problem is that we set the flag 
```
-DCMAKE_INSTALL_PREFIX=$PWD/staging
``` 
in the `release.yml` and `release_Ubuntu.yml` files, and use 
```
set (PACKAGE_DATA_DIR "${CMAKE_INSTALL_PREFIX}/share")
```
and 
```
set (PACKAGE_LOCALE_DIR ${PACKAGE_DATA_DIR}/locale PARENT_SCOPE)
```
in the `CMakeLists.txt` files.

Then you can see from the message (after starting xournalpp from the console):

``` 
** Message: 15:51:46.203: TEXTDOMAINDIR = (null), PACKAGE_LOCALE_DIR = /__w/1/s/build/staging/share/locale, chosen directory = /__w/1/s/build/staging/share/locale
```
that the wrong `locale` folder is chosen. 

A workaround is to set the environment variable `TEXTDOMAINDIR` to `/usr/share/locale/`, but we better repair our `cmake`scripts and/or `.yml` files to fix this.

rolandlo:
Fixes #2945 
Not sure if it works, but we will see it from testing the Azure pipelines.

### DIFF
--- a/azure-pipelines/jobs/release_ubuntu.yml
+++ b/azure-pipelines/jobs/release_ubuntu.yml
@@ -18,7 +18,7 @@ jobs:
       - template: ../steps/build_linux.yml
         parameters:
           build_type: 'RelWithDebInfo'
-          cmake_flags: '-DCMAKE_DEBUG_INCLUDES_LDFLAGS=ON -DENABLE_CPPUNIT=ON -DDEBUG_COMPILE=ON -DCMAKE_INSTALL_PREFIX=$PWD/staging -DCPACK_GENERATOR="TGZ;DEB"'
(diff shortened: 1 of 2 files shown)


rolandlo:
It works. I tested everything on Azure Pipelines including the AppImage as well.

PR #2953, opened 2021-03-15
Typo in Lua API documentation for scrollToPos

raw-bacon:
In one of the two given examples the name of the function was correct, in the other it was `scrollToPage` rather than `scrollToPos`.

Let me know if these PRs are too silly.


rolandlo:
Thanks for the fix!

PR #2955, opened 2021-03-16
Fix crash on scroll with sidebar preview open

### LINKED ISSUE #2887: Crash when scrolling through PDF with sidebar open on Layer/Layerstack Preview

personalizedrefrigerator:
## Summary
 * A minor scheduler refactoring to use `std::deque`, `std::mutex`, and `std::condition_variable`.
 * Fixes a crash on scrolling through long PDFs with the sidebar layers preview open.
     * The crash was caused by a `nullptr` (or, possibly sometimes not, we weren't using a `std::atomic` to sync the object that was set to null between threads, but it **did** involve accessing freed memory) de-ref in `PreviewJob.cpp`.  
     * I think a part of this was not waiting for preview jobs to finish, possibly freeing the SidebarPreview (that PreviewJob uses) before letting PreviewJobs finish running.
 
Fixes #2887.
See also #2907.

### DIFF
--- a/src/control/jobs/Job.h
+++ b/src/control/jobs/Job.h
@@ -72,6 +72,12 @@ protected:
      */
     virtual void afterRun();
 
+    /**
+     * Called when the job is to be removed from the
+     * scheduler.
+     */
+    virtual void onDelete();
+
 private:
     static bool callAfterCallback(Job* job);
 
--- a/src/control/jobs/Scheduler.cpp
+++ b/src/control/jobs/Scheduler.cpp
@@ -1,5 +1,6 @@
 #include "Scheduler.h"
 
+#include <cassert>
 #include <cinttypes>
 
 #include <config-debug.h>
@@ -13,20 +14,7 @@
 Scheduler::Scheduler() {
     this->name = "Scheduler";
 
-    // Thread
-    g_cond_init(&this->jobQueueCond);
-
-    g_mutex_init(&this->jobQueueMutex);
-    g_mutex_init(&this->jobRunningMutex);
-    g_mutex_init(&this->schedulerMutex);
-    g_mutex_init(&this->blockRenderMutex);
-
     // Queue
-    GQueue init = G_QUEUE_INIT;
-    this->queueUrgent = init;
-    this->queueHigh = init;
-    this->queueLow = init;
-    this->queueNone = init;
     this->jobQueue[JOB_PRIORITY_URGENT] = &this->queueUrgent;
     this->jobQueue[JOB_PRIORITY_HIGH] = &this->queueHigh;
     this->jobQueue[JOB_PRIORITY_LOW] = &this->queueLow;
@@ -67,7 +55,7 @@ void Scheduler::stop() {
         return;
     }
     this->threadRunning = false;
-    g_cond_broadcast(&this->jobQueueCond);
+    this->jobQueueCond.notify_all();
 
     if (this->thread) {
         g_thread_join(this->thread);
@@ -77,38 +65,42 @@ void Scheduler::stop() {
 void Scheduler::addJob(Job* job, JobPriority priority) {
     SDEBUG("Adding job...");
 
-    g_mutex_lock(&this->jobQueueMutex);
+    {
+        std::lock_guard lock{this->jobQueueMutex};
 
-    job->ref();
-    g_queue_push_tail(this->jobQueue[priority], job);
-    g_cond_broadcast(&this->jobQueueCond);
+        job->ref();
+        this->jobQueue[priority]->push_back(job);
+    }
 
-    SDEBUG("add job: %" PRId64, (uint64_t)job);
-
-    g_mutex_unlock(&this->jobQueueMutex);
+    SDEBUG("add job: %" PRId64 "; type: %" PRId64, (uint64_t)job, (uint64_t)job->getType());
+    this->jobQueueCond.notify_all();
 }
 
 auto Scheduler::getNextJobUnlocked(bool onlyNotRender, bool* hasRenderJobs) -> Job* {
     Job* job = nullptr;
 
     for (int i = JOB_PRIORITY_URGENT; i < JOB_N_PRIORITIES; i++) {
+        std::deque<Job*>& queue = *this->jobQueue[i];
+
         if (onlyNotRender) {
-            for (GList* l = this->jobQueue[i]->head; l != nullptr; l = l->next) {
-                job = static_cast<Job*>(l->data);
+            for (auto it = queue.begin(); it != queue.end(); ++it) {
+                job = *it;
 
                 if (job->getType() != JOB_TYPE_RENDER) {
-                    g_queue_delete_link(this->jobQueue[i], l);
+                    queue.erase(it);
                     return job;
                 }
-                if (hasRenderJobs) {
+
+                if (hasRenderJobs != nullptr) {
                     *hasRenderJobs = true;
                 }
             }
-        } else {
-            job = static_cast<Job*>(g_queue_pop_head(this->jobQueue[i]));
-            if (job) {
-                return job;
-            }
+        } else if (!queue.empty()) {
+            job = queue.front();
+            queue.pop_front();
+            assert(job != nullptr);
+
+            return job;
         }
     }
 
@@ -118,17 +110,17 @@ auto Scheduler::getNextJobUnlocked(bool onlyNotRender, bool* hasRenderJobs) -> J
 /**
  * Locks the complete scheduler
  */
-void Scheduler::lock() { g_mutex_lock(&this->schedulerMutex); }
+void Scheduler::lock() { this->schedulerMutex.lock(); }
 
 /**
  * Unlocks the complete scheduler
  */
-void Scheduler::unlock() { g_mutex_unlock(&this->schedulerMutex); }
+void Scheduler::unlock() { this->schedulerMutex.unlock(); }
 
 #define ZOOM_WAIT_US_TIMEOUT 300000  // 0.3s
 
 void Scheduler::blockRerenderZoom() {
-    g_mutex_lock(&this->blockRenderMutex);
+    std::lock_guard lock{this->blockRenderMutex};
 
     if (this->blockRenderZoomTime == nullptr) {
         this->blockRenderZoomTime = g_new(GTimeVal, 1);
@@ -136,23 +128,22 @@ void Scheduler::blockRerenderZoom() {
 
     g_get_current_time(this->blockRenderZoomTime);
     g_time_val_add(this->blockRenderZoomTime, ZOOM_WAIT_US_TIMEOUT);
-
-    g_mutex_unlock(&this->blockRenderMutex);
 }
 
 void Scheduler::unblockRerenderZoom() {
-    g_mutex_lock(&this->blockRenderMutex);
+    {
+        std::lock_guard lock{this->blockRenderMutex};
 
-    g_free(this->blockRenderZoomTime);
-    this->blockRenderZoomTime = nullptr;
-    if (this->jobRenderThreadTimerId) {
-        g_source_remove(this->jobRenderThreadTimerId);
-        this->jobRenderThreadTimerId = 0;
+        g_free(this->blockRenderZoomTime);
+        this->blockRenderZoomTime = nullptr;
+
+        if (this->jobRenderThreadTimerId) {
+            g_source_remove(this->jobRenderThreadTimerId);
+            this->jobRenderThreadTimerId = 0;
+        }
     }
 
-    g_mutex_unlock(&this->blockRenderMutex);
-
-    g_cond_broadcast(&this->jobQueueCond);
+    this->jobQueueCond.notify_all();
 }
 
 /**
@@ -179,12 +170,13 @@ auto g_time_val_diff(GTimeVal* t1, GTimeVal* t2) -> glong {
 auto Scheduler::jobRenderThreadTimer(Scheduler* scheduler) -> bool {
     scheduler->jobRenderThreadTimerId = 0;
 
-    g_mutex_lock(&scheduler->blockRenderMutex);
-    g_free(scheduler->blockRenderZoomTime);
-    scheduler->blockRenderZoomTime = nullptr;
-    g_mutex_unlock(&scheduler->blockRenderMutex);
+    {
+        std::lock_guard lock{scheduler->blockRenderMutex};
+        g_free(scheduler->blockRenderZoomTime);
+        scheduler->blockRenderZoomTime = nullptr;
+    }
 
-    g_cond_broadcast(&scheduler->jobQueueCond);
+    scheduler->jobQueueCond.notify_all();
 
     return false;
 }
@@ -192,12 +184,15 @@ auto Scheduler::jobRenderThreadTimer(Scheduler* scheduler) -> bool {
 auto Scheduler::jobThreadCallback(Scheduler* scheduler) -> gpointer {
     while (scheduler->threadRunning) {
         // lock the whole scheduler
-        g_mutex_lock(&scheduler->schedulerMutex);
+        std::unique_lock schedulerLock{scheduler->schedulerMutex};
+        SDEBUG("Job Thread: Blocked scheduler.");
 
-        g_mutex_lock(&scheduler->blockRenderMutex);
-        bool onlyNoneRenderJobs = false;
+        bool onlyNonRenderJobs = false;
         glong diff = 1000;
         if (scheduler->blockRenderZoomTime) {
+            std::lock_guard lock{scheduler->blockRenderMutex};
+            SDEBUG("Zoom re-render blocking.");
+
             GTimeVal time;
             g_get_current_time(&time);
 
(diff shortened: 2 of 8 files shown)


src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
 XournalScheduler::~XournalScheduler() = default;
 
 void XournalScheduler::removeSidebar(SidebarPreviewBaseEntry* preview) {
-    // Dont await for tasks, as it completly blocks whole program.
-    // Not sure why.
-    removeSource(preview, JOB_TYPE_PREVIEW, JOB_PRIORITY_HIGH, false);
+    removeSource(preview, JOB_TYPE_PREVIEW, JOB_PRIORITY_HIGH, true);
```
personalizedrefrigerator, line 15: This change should fix the crash on scroll -- resources in use by a PreviewJob could be freed before the job finished. SidebarPreviewBaseEntry calls XournalScheduler::removeSidebar in its destructor, so if we don't wait for tasks to finish (the true parameter), the sidebar's destructor continues, freeing resources required by a potentially-still-running PreviewJob.

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
 XournalScheduler::~XournalScheduler() = default;
 
 void XournalScheduler::removeSidebar(SidebarPreviewBaseEntry* preview) {
-    // Dont await for tasks, as it completly blocks whole program.
```
personalizedrefrigerator, line 15: I haven't noticed this blocking after making this change (possibly due to the refactor?), but I don't know how often/what caused this blocking to occur...

src/control/jobs/Scheduler.h
```diff
@@ -111,25 +114,31 @@ class Scheduler {
...
 
-    GQueue queueUrgent{};
-    GQueue queueHigh{};
-    GQueue queueLow{};
-    GQueue queueNone{};
+    /**
+     * Jobs of each priority. New jobs
+     * are added to the back of each queue.
+     */
+    std::deque<Job*> queueUrgent{}, queueHigh{}, queueLow{}, queueNone{};
```
personalizedrefrigerator, line 131: These are `std::deque`s and not `std::queue`s. This is because we use iteration. See https://stackoverflow.com/questions/1259099/stdqueue-iteration.

src/control/jobs/XournalScheduler.cpp
```diff
@@ -66,28 +75,21 @@ void XournalScheduler::removeSource(void* source, JobType type, JobPriority prio
...
         finishTask();
     }
-
-    g_mutex_unlock(&this->jobQueueMutex);
 }
 
 auto XournalScheduler::existsSource(void* source, JobType type, JobPriority priority) -> bool {
     bool exists = false;
-    g_mutex_lock(&this->jobQueueMutex);
+    std::lock_guard<std::mutex> lock(this->jobQueueMutex);
```
personalizedrefrigerator, line 82: We could use `std::scoped_lock` here (and in similar locations), but I think I agree with this post: https://stackoverflow.com/posts/60172828/timeline#history_38bf9e37-80b9-49e6-ba53-299043622aea.

src/control/jobs/Scheduler.h
Febbe: It's good to use std::deque :+1: 
But please don't declare more than one variable of the same type in one line.

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
...
-    // Not sure why.
-    removeSource(preview, JOB_TYPE_PREVIEW, JOB_PRIORITY_HIGH, false);
+    removeSource(preview, JOB_TYPE_PREVIEW, JOB_PRIORITY_HIGH, true);
 }
 
 void XournalScheduler::removePage(XojPageView* view) { removeSource(view, JOB_TYPE_RENDER, JOB_PRIORITY_URGENT); }
 
 void XournalScheduler::removeAllJobs() {
-    g_mutex_lock(&this->jobQueueMutex);
+    std::lock_guard<std::mutex> lock(this->jobQueueMutex);
```
Febbe, line 21: **nitpick:** Very good change :+1: But it's not required in cpp17 to write the template parameter, also use `{}` to initialize objects.
```suggestion
    std::unique_lock guard{this->jobQueueMutex};
```

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
...
             if (type == JOB_TYPE_PREVIEW || type == JOB_TYPE_RENDER) {
                 job->deleteJob();
-                g_queue_remove(this->jobQueue[priority], job);
+
+                it = queue.erase(it);
+
                 job->unref();
+                job = nullptr;
+            } else {
+                ++it;
```
Febbe, line 41: Use `std::erase` (cpp20) or `std::remove_if` in combination with `container.erase`

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
...
     }
-
-    g_mutex_unlock(&this->jobQueueMutex);
 }
 
 void XournalScheduler::finishTask() {
-    g_mutex_lock(&this->jobRunningMutex);
-    g_mutex_unlock(&this->jobRunningMutex);
+    this->jobRunningMutex.lock();
+    this->jobRunningMutex.unlock();
```
Febbe, line 49: We may want to use `std::condition_variable::wait`

Febbe: also, don't call lock/unlock manually
```suggestion
    std::unique_lock guard g{this->jobRunningMutex};
```

src/control/jobs/XournalScheduler.cpp
Febbe: why not using `std::unique_lock`?

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
...
 }
 
 void XournalScheduler::removeSource(void* source, JobType type, JobPriority priority, bool awaitFinishTask) {
-    g_mutex_lock(&this->jobQueueMutex);
+    {
+        std::lock_guard<std::mutex> lock(this->jobQueueMutex);
```
Febbe, line 54:
```suggestion
        std::unique_lock lock{this->jobQueueMutex};
```

src/control/jobs/XournalScheduler.cpp
```diff
@@ -12,51 +12,60 @@ XournalScheduler::XournalScheduler()
...
+        auto it = this->jobQueue[priority]->begin();
 
-    int length = g_queue_get_length(this->jobQueue[priority]);
-    for (int i = 0; i < length; i++) {
-        Job* job = static_cast<Job*>(g_queue_peek_nth(this->jobQueue[priority], i));
+        while (it != this->jobQueue[priority]->end()) {
+            Job* job = *it;
+
+            if (job->getType() == type && job->getSource() == source) {
+                it = this->jobQueue[priority]->erase(it);
```
Febbe, line 62: use `std::remove_if` and `container.erase`

src/control/jobs/XournalScheduler.cpp
```diff
@@ -66,28 +75,21 @@ void XournalScheduler::removeSource(void* source, JobType type, JobPriority prio
...
 auto XournalScheduler::existsSource(void* source, JobType type, JobPriority priority) -> bool {
     bool exists = false;
-    g_mutex_lock(&this->jobQueueMutex);
+    std::lock_guard<std::mutex> lock(this->jobQueueMutex);
 
-    int length = g_queue_get_length(this->jobQueue[priority]);
-    for (int i = 0; i < length; i++) {
-        Job* job = static_cast<Job*>(g_queue_peek_nth(this->jobQueue[priority], i));
+    for (auto it = this->jobQueue[priority]->begin(); it != this->jobQueue[priority]->end(); ++it) {
+        Job* job = *it;
```
Febbe, line 85:
```suggestion
    for(Job* job:this->jobQueue[priority]){
```

Febbe: You could also use `std::find_if` and remove the complete forloop.

Febbe (changes requested):

src/control/jobs/Scheduler.cpp
```diff
@@ -179,78 +170,83 @@ auto g_time_val_diff(GTimeVal* t1, GTimeVal* t2) -> glong {
...
+        SDEBUG("do job: %" PRId64, (uint64_t)job);
         job->execute();
-
         job->unref();
-        g_mutex_unlock(&scheduler->jobRunningMutex);
+        scheduler->jobRunningMutex.unlock();
 
         // unlock the whole scheduler
-        g_mutex_unlock(&scheduler->schedulerMutex);
+        scheduler->schedulerMutex.unlock();
```
Febbe, line 249: only lock mutexes over guards

src/control/jobs/XournalScheduler.cpp
personalizedrefrigerator: To me,
```c++
        queue.erase(std::remove_if(queue.begin(), queue.end(),
            [&](Job* job) {
                if (job->getType() == type && job->getSource() == source) {
                    job->deleteJob();
                    job->unref();

                    return true;
                }   

                return false;
            }), queue.end());
```

makes less sense than

```c++
        auto it = this->jobQueue[priority]->begin();

        while (it != this->jobQueue[priority]->end()) {
            Job* job = *it;

            if (job->getType() == type && job->getSource() == source) {
                it = this->jobQueue[priority]->erase(it);

                job->deleteJob();
                job->unref();
                job = nullptr;
            } else {
                ++it;
            }   
        } 
```

because we're calling `job->unref()` in the predicate given to `std::remove_if`. I'm willing to change it to the `queue.erase(std::remove_if(...))` version, if you disagree, though.

src/control/jobs/XournalScheduler.cpp
personalizedrefrigerator: I think locking on the `jobRunningMutex` is better here; we don't want this to block when a job isn't already running.

I made the suggested change, though!

src/control/jobs/XournalScheduler.cpp
personalizedrefrigerator: Nice catch!

src/control/jobs/XournalScheduler.cpp
personalizedrefrigerator: Same comment as above!

src/control/jobs/XournalScheduler.cpp
Febbe: Maybe, because u started to use "lock guards" which will automatically unlock the mutex when the guard is destroyed.

src/control/jobs/XournalScheduler.cpp
Febbe: We need a refactor of the job system. It should use `std::shared_ptr`s. The actions executed here have to be done for each destruction of Jobs, so they should be moved into the destructor.
Basically `Job::deleteJob()` must be removed. And `Job::unref` has to be handled by a `shared_ptr`.

Febbe: Regarding the while over `remove_if` question. This is indeed ideological. 
In my opinion, a loop whether `while` or `for` does not tell anything about what happens there.

`remove_if` exactly tells the reader, that you want to erase objects from the container.
Also, removing elements from the list while iterating is first prone to bugs, second it can be slow for some containers. (Not in case of deque's).

> I would very much like the Job* job parameter to be a const Job* job in the lambda
Yes, I agree, that the parameter should be `const`, but that's not possible, because the `Job`s are designed wrongly.

> I don't think remove_if guarantees that the predicate will be called exactly once for each job

Even if not: 2*O(1) is still O(1) and 2*O(n) is still O(n) (2*O(n²) = O(n²))...
So the disadvantage would be minimal, but the `erase(remove_if` is designed to make deleting many objects from the list as optimal as possible. It works by moving all the objects to be removed to the end of the list. Then the complete end is removed at once.

src/control/jobs/XournalScheduler.cpp
Febbe: >                     `// I think the best solution would be to use shared_ptrs to track`
                   `// jobs, but I think this is out of the scope of this PR.`

haha lol: exactly what I said above :D

src/control/jobs/Scheduler.cpp
```diff
@@ -77,38 +65,42 @@ void Scheduler::stop() {
...
+
         if (onlyNotRender) {
-            for (GList* l = this->jobQueue[i]->head; l != nullptr; l = l->next) {
-                job = static_cast<Job*>(l->data);
+            for (auto it = queue.begin(); it != queue.end(); ++it) {
+                job = *it;
 
                 if (job->getType() != JOB_TYPE_RENDER) {
-                    g_queue_delete_link(this->jobQueue[i], l);
+                    queue.erase(it);
```
Febbe, line 90: This is a bug, the iterators are invalidated here, so `it` has to be reassigned.

src/control/jobs/Scheduler.h
```diff
@@ -111,25 +114,34 @@ class Scheduler {
...
+    std::deque<Job*> queueLow{};
+    std::deque<Job*> queueNone{};
 
-    GQueue* jobQueue[JOB_N_PRIORITIES]{};
+    /**
+     * Map (JobPriority) -> queue of that priority.
+     * For example,
+     *   jobQueue[JOB_PRIORITY_URGENT] is &queueUrgent.
+     */
+    std::deque<Job*>* jobQueue[JOB_N_PRIORITIES]{};
```
Febbe, line 141: **nitpick:** use std::array here

src/control/jobs/XournalScheduler.cpp
Febbe: Ok just read the [documentation](https://en.cppreference.com/w/cpp/algorithm/remove): You are right, 
```

p | - | unary predicate which returns ​true  if the element should be removed.  The expression p(v) must be convertible to bool for every argument v of type (possibly const) VT, where VT is the value type of ForwardIt, regardless of value category, and must not modify v. Thus, a parameter type of VT&is not allowed, nor is VT unless for VT a move is equivalent to a copy (since C++11). ​
-- | -- | --
```

src/control/jobs/XournalScheduler.cpp
Febbe: If it is fixed yet, can this change be reverted?

personalizedrefrigerator: Reverting this should make the issue re-appear. 

In its destructor, `SidebarPreviewBaseEntry` calls `removeSidebar` **before freeing other resources**. **PreviewJobs** expect these resources to be available until they finish. If we return before all jobs using the preview have finished, we have potential use-after-frees -- the preview jobs expect the preview to still exist.

I'm adding a comment to the docstring for `removeSidebar`. Additionally, `removePage` already acts like this -- it waits for running jobs to finish.

src/control/jobs/Scheduler.h
personalizedrefrigerator: Good idea!

src/control/jobs/XournalScheduler.cpp
personalizedrefrigerator: Reverting the change does break the fix, but it (at least on my machine) isn't causing a crash.
All pages say "Loading..." in the document and in the preview.
The scheduler blocks indefinitely after scrolling through a large PDF with layer preview or layerstack preview open.

I think this is happening because we're trying to lock a freed mutex (line 36):
```c++
 35 void PreviewJob::finishPaint() {
 36     g_mutex_lock(&this->sidebarPreview->drawingMutex);
 37 
 38     if (this->sidebarPreview->crBuffer) {
 39         cairo_surface_destroy(this->sidebarPreview->crBuffer);
 40     }
 41     this->sidebarPreview->crBuffer = crBuffer;
 42 
 43     // The preview widget can be referenced after this is deleted.
 44     // Only it should be referenced in the callback.
 45     GtkWidget* previewWidget = this->sidebarPreview->widget;
 46     g_object_ref(previewWidget);
 47 
 48     Util::execInUiThread([previewWidget]() {
 49         gtk_widget_queue_draw(previewWidget);
 50         g_object_unref(previewWidget);
 51     });
 52 
 53     g_mutex_unlock(&this->sidebarPreview->drawingMutex);
 54 }

```

```c++
 30 SidebarPreviewBaseEntry::~SidebarPreviewBaseEntry() {
 31     this->sidebar->getControl()->getScheduler()->removeSidebar(this);
 32     this->page = nullptr;
 33 
 34     gtk_widget_destroy(this->widget);
 35     this->widget = nullptr;
 36 
 37     if (this->crBuffer) {
 38         cairo_surface_destroy(this->crBuffer);
 39         this->crBuffer = nullptr;
 40     }
 41 }
```

If we don't block on line 31, the SidebarPreviewBaseEntry, and its `drawingMutex`, are freed. This can happen in the middle of a `PreviewJob`, before we attempt to `g_mutex_lock(&this->sidebarPreview->drawingMutex)` the freed mutex.

Febbe:
I think we can add this to 1.1.0 Since it is more a bugfix, than a refactoring.
@personalizedrefrigerator can you add this fix to the changelog and add the refactor of jobs to use `shared_ptr`s to the todo list.

src/control/jobs/Job.h
```diff
@@ -72,6 +72,12 @@ class Job {
      */
     virtual void afterRun();
 
+    /**
+     * Called when the job is to be removed from the
+     * scheduler.
+     */
+    virtual void onDelete();
+
```
Febbe, line 80: Is this necessary? It violates https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-ctor-virtual

personalizedrefrigerator:
> can you add this fix to the changelog

This problem doesn't seem to have been present in 1.0.20 (or was, at least, much harder to trigger). `v1.0.20` only updates the layerstack preview after a user has stopped scrolling and interacted with the current page, and, as such, this crash doesn't seem to happen when quickly scrolling through a document.

> add the refactor of jobs to use `shared_ptr`s to the todo list.

> * [ ] Rework Job system

is already on the to-do list. @Febbe can you edit it (maybe a sub-task) to reflect the refactor to use `shared_ptr`s? (I don't think I can edit the to-do list.)

src/control/jobs/Scheduler.cpp
```diff
@@ -118,41 +110,40 @@ auto Scheduler::getNextJobUnlocked(bool onlyNotRender, bool* hasRenderJobs) -> J
 /**
  * Locks the complete scheduler
  */
-void Scheduler::lock() { g_mutex_lock(&this->schedulerMutex); }
+void Scheduler::lock() { this->schedulerMutex.lock(); }
```
Febbe, line 113: I think this function should return a unique_lock

Febbe: I think, we also could defer that. I'll add this to the to-do list :)

PR #2967, opened 2021-03-20
Isometric grid fix

### LINKED ISSUE #2964: Isometric grid does not line up in landscape
**Affects versions :**
 - OS: Fedora Linux 33
 - Desktop environment: none (X11/i3)
 - libgtk: 3.24.27
 - Xournal++ 1.1.0+dev

**Describe the bug**
In an isometric grid, 3 lines are supposed to intersect in one point. In landscape mode, the lines do not always intersect in one point but leave a little trangle instead (the 3 lines intersect pairwise, for the 3 pairs, in 3 different points).

**To Reproduce**
1. Choose isometric grid
2. Switch journal page to landscape
3. Observe the intersections

**Expected behavior**
3 lines are supossed to intersect in 1 point.

**Screenshots of Problem**
xopp (zipped for GH's sake)
[isometricLandscape.zip](https://github.com/xournalpp/xournalpp/files/6170571/isometricLandscape.zip)
pdf
[isometricLandscape.pdf](https://github.com/xournalpp/xournalpp/files/6170562/isometricLandscape.pdf)

**Additional context**
The problem shows up in xournalpp itself (document view) as well as the PDF export.
It shows up in parts of the page only which indicates some sort of rounding problem.

I also noticed that "grid snapping" uses the rectangular grid, not the isometric one, but this may be completely unrelated.

mjg:
Fixes #2964

### DIFF
--- a/ui/pagetemplates.ini
+++ b/ui/pagetemplates.ini
@@ -34,13 +34,11 @@ format=dotted
 name=Isometric Dotted
 name[de]=Isometrisch Gepunktet
 format=isodotted
-config=r1=20
 
 [predefinedIsoGraph]
 name=Isometric Graph
 name[de]=Isometrisches Raster
 format=isograph
-config=r1=20
 
 [graphWithBorder]
 name=Graph with border
--- a/ui/pagetemplates.ini
+++ b/ui/pagetemplates.ini
@@ -5,12 +5,12 @@ name=Plain
 name[de]=Einfarbig
 format=plain
 
-[predefinedLined]
+[predefinedRuled]
 name=Ruled
 name[de]=Liniert
 format=ruled
 
-[predefinedLinedVLine]
+[predefinedLined]
 name=Ruled with vertical line
 name[de]=Liniert mit Randlinie
 format=lined
@@ -25,33 +25,26 @@ name=Graph
 name[de]=Kariert
 format=graph
 
-[XournalppDotted]
+[predefinedDotted]
 name=Dotted
 name[de]=Gepunktet
 format=dotted
 
 [predefinedIsoDotted]
 name=Isometric Dotted
-name[de]=Isometisch Gepunktet
+name[de]=Isometrisch Gepunktet
 format=isodotted
 config=r1=20
 
 [predefinedIsoGraph]
 name=Isometric Graph
-name[de]=Isometisches Raster
+name[de]=Isometrisches Raster
 format=isograph
-config=lw=1,r1=20
+config=r1=20
 
-[ruledWithBorder]
+[graphWithBorder]
 name=Graph with border
 name[de]=Kariert mit Rand
 format=graph
 config=m1=40,rm=1
-
-
-#[XournalppRedTest]
-#name=Red dotted
-#name[de]=Rot gepunktet
-#format=dotted
-#config=f1=FF0000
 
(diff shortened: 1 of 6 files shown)


rolandlo (approved):
Thanks for the fix, which looks good. I also check all cases (even-even, odd-even, even-odd, odd-odd).
Minor nitpick in the commit description: `take of that extra diagonal when
needed.` (of -> off)
Could you please also fix the German spelling in the `pagetemplates.ini`? `Isometisch(es)` should be `Isometrisch(es)`. 

By the way, I have no idea why the side length of the equilateral triangles is set to 20pt = 5/18in in the `pagetemplates.ini`. If you want to change that value to something more sensible, please go ahead. It's not a big deal though, since it can easily be modified by the user.

mjg:
So far I've pushed the commit message fix ("take care of") and a few clean ups which do not change the behaviour.

The isometric grid specifies a different default in source than the ini file does. It's been like that since it's inception, which is strange. The default in source is the same as for graph paper. Alas, isometric grid has a margin and graph paper does not, so even with the same grid spacing they do not line up. Which made me look at some more parameters...

There is "rm" (roundMargin) but only "graph" obeys it. Well, kind of: `style="graph" config="m1=20,rm=1"` does not do on A4 paper what I think it is supposed to do. (Looks like a wrong shift.)

Isometric ground rounds no matter what and has a margin no matter what (as if "rm" and "m1" were hardcoded).

Speaking of which: "m1" is obeyed by "graph" only.

And the vertical line of "lined" could be an option, turning "lined" into "ruled" with "vline=72" to have the position in config.

So, the more I look at this the more I want to change and make uniform. BUT: Even just removing "r1=20" for the isomeric grid means that existing xopp-files may display differently. Unless I change the default in source to 20, of course. I'm not sure how to handle this in the interest of "stability" of the file format. Maybe postpone a larger overhaul for a new file format version?

mjg:
Turns out I was overly cautious. Existing xopp files store the extra parameters from the ini file. So I'll add another commit which removes the ini default, changing the display only for new files with isometric grid.

rolandlo:
@mjg 

I don't think that removing "r1=20" in the `pagetemplates.ini` (or replacing it with another value) means that existing xopp-files display differently than before. That value is saved in the xopp-file. 
In general I agree that the page types and parameters need a larger overhaul and this will most likely affect the file format, which may be better to do when introducing a new file format version. 

Let me remind you that you still haven't fixed the `Isometisch/Isometisches` in `pagetemplates.ini`.

ui/pagetemplates.ini
```diff
@@ -34,24 +34,15 @@ format=dotted
 name=Isometric Dotted
 name[de]=Isometisch Gepunktet
```
rolandlo, line 35:
```suggestion
name[de]=Isometrisch Gepunktet
```

ui/pagetemplates.ini
```diff
@@ -34,24 +34,15 @@ format=dotted
...
 format=isodotted
-config=r1=20
 
 [predefinedIsoGraph]
 name=Isometric Graph
 name[de]=Isometisches Raster
```
rolandlo, line 40:
```suggestion
name[de]=Isometrisches Raster
```

rolandlo (changes requested):

rolandlo (approved):
Ready for merge?

rolandlo:
Ok, thanks a lot for the fixes!

PR #2973, opened 2021-03-25
Update MacBuild.md

zhangguiyu:
Updated Homebrew Build dependencies list by adding

libzip portaudio libsndfile


Febbe:
Thank you for your contribution. Can you also check, if plugins are enabled? if not, try to add lua.

PR #2980, opened 2021-03-27
Fix off-by-one error in toolbar customization

### LINKED ISSUE #2970: Editing toolbar button configuation - button locations get mixed up
**Affects versions :**
 - OS: WIN10, Linux Mint 19
 - (Linux only) Desktop environment:  X11
 - Version of Xournal++: xournalpp-1.1.0.dev-nightly.20210318
 - Installation method: AppImage / Win-Exe

**Describe the bug**
Changes in toolbar configuration (-> location of buttons) are not stored after restart of Xournal++.
See also attached images

**To Reproduce**
1. Step 1
Open toolbar configuration editor

2. Step 2
Add and rearrange buttons for page-moving |<    <    X    >     >|
(see --> Image 01

3. Step 3
Close toolbar configuration editor

4. Step 4
Check location of buttons (seems to be ok --> Image 02)

5. Step 5
Restart Xournal++

**Expected behavior**
Location of butons should be as in Step 4.

But position of buttons are mixed --> Image 03
Re-editing of toolbar and restarting of Xournal++ don't solve the issue.

**Screenshots of Problem**
Image 01
Image01

Image 02 - as it should be
Image02

Image 03 - as it is
Image03

**Additional context**
toolbar.ini

================================================================

## Here you can customize the Toolbars
# Delete this file to generate a new config file with default values
#
# Available buttons:
# File: NEW,SAVE,OPEN,PRINT
#
# Edit: CUT,COPY,PASTE,SEARCH,UNDO,REDO,INSERT_NEW_PAGE
#
# Navigation: GOTO_FIRST,GOTO_BACK,GOTO_NEXT,GOTO_LAST,GOTO_NEXT_ANNOTATED_PAGE
#
# Zoom: ZOOM_OUT,ZOOM_IN,ZOOM_FIT,ZOOM_100,FULLSCREEN,PAIRED_PAGES
#
# Color: COLOR(0xffffff),COLOR(0xffff00),COLOR(0xff8000),COLOR(0xff00ff),COLOR(0x00ff00),COLOR(0x00c0ff),COLOR(0x808080),COLOR(0x008000),COLOR(0xff0000),COLOR(0x3333cc),COLOR(0x000000),COLOR_SELECT
#  Notice: This are the default Xournal colors, each other color in HEX can also be used, eg COLOR(0x12ABCF);
#
# Tools: ERASER,PEN,HIGHLIGHTER,IMAGE,TEXT,IMAGE,SELECT,SELECT_REGION,SELECT_RECTANGLE,VERTICAL_SPACE,HAND
#  Notice: ERASER also has a drop down menu to select the eraser type, SELECT are all selection tools, with drop down menu
#
# Tool settings: SHAPE_RECOGNIZER,RULER,FINE,MEDIUM,THICK,SELECT_FONT
#
# Components: PAGE_SPIN,ZOOM_SLIDER,LAYER
#  PAGE_SPIN: The page spiner, incl. current page label
#  ZOOM_SLIDER: The zoom slider
#  LAYER: The layer dropdown menu
#
#

[Right hand Note Taking Copy Copy]
toolbarTop1=SAVE,NEW,OPEN,SEPARATOR,CUT,COPY,PASTE,SEPARATOR,UNDO,REDO,SEPARATOR,SEPARATOR,FULLSCREEN,HAND,SEPARATOR,DEFAULT_TOOL,SEPARATOR,PEN,HILIGHTER,IMAGE,ERASER,SEPARATOR,SHAPE_RECOGNIZER,TEXT,SEPARATOR,SELECT_FONT,INSERT_NEW_PAGE,DELETE_CURRENT_PAGE,SEPARATOR,GOTO_BACK,GOTO_NEXT_ANNOTATED_PAGE,GOTO_FIRST,GOTO_NEXT,GOTO_LAST,PAGE_SPIN,SEPARATOR,SEPARATOR
toolbarLeft1=COLOR(0xffffff),COLOR(0xffff00),COLOR(0xff8000),COLOR(0xff00ff),COLOR(0x00ff00),COLOR(0x00c0ff),COLOR(0x808080),COLOR(0x008000),COLOR(0x3333cc),COLOR(0xff0000),COLOR(0x000000),COLOR_SELECT,SEPARATOR,ZOOM_OUT,ZOOM_IN,ZOOM_FIT,ZOOM_SLIDER,ZOOM_100,SEPARATOR
toolbarLeft2=VERY_FINE,FINE,MEDIUM,THICK,VERY_THICK,SEPARATOR,TOOL_FILL,DRAW_CIRCLE,SEPARATOR,DRAW_RECTANGLE,DRAW_ARROW,RULER,SEPARATOR,DRAW_COORDINATE_SYSTEM,DRAW_SPLINE,ROTATION_SNAPPING,GRID_SNAPPING,SEPARATOR,VERTICAL_SPACE,SELECT_REGION,SEPARATOR,SELECT_RECTANGLE,SELECT_OBJECT,DELETE
name=Test_V01 

================================================================

idotobi:
Fix the behaviour that any addition to the toolbar during customization
would show one position further to the right after reopening the
application.
(Notable exception: the very first position)

This PR fixes #2970


src/gui/toolbarMenubar/model/ToolbarEntry.cpp
```diff
@@ -49,6 +45,13 @@ auto ToolbarEntry::removeItemById(int id) -> bool {
 
 auto ToolbarEntry::insertItem(string item, int position) -> int {
     auto* it = new ToolbarItem(std::move(item));
+    // This is required as the position generated by gtk_toolbar_get_drop_index
+    // is 1 indexed with the notable exception that the beginning of the toolbar
+    // is still 0.
+    // Hence, the possible input positions are [0, 2, 3, 4, ...] which need to be mapped
+    // to [0, 1, 2, 3, ...] to work consistently with the the entries vector.
+    if (position >= 1)
+        position--;
```
Febbe, line 54: Dafuq? Isn't there be a more sane approach? What is the cause of this inconsistency?

Febbe: Let us assume, the implementation of gtk_toolbar_get_drop_index is correct, then we have a bug in the input values or with the handling of the returnvalue.
So we should fix that, instead of creating a fixup *after* the call.

I think the bug is in `ToolbarAdapter::toolbarDragDataReceivedCb` and `ToolbarAdapter::toolbarDragMotionCb`

For me, the following looks wrong:
```
    // x,y are already in toolbar coordinates as the gtk_toolbar_get_drop_index() specification requires.
    // However, without this translation we have an unwanted vertical offset.
```

idotobi: You're right it's not a bug in gtk --that would have been weird.

However, it took me a while to find the culprit. The `ToolbarAdapter` adds a hidden`GtkToolItem` at construction at the `0`th index in the toolbar (see https://github.com/xournalpp/xournalpp/blob/22e60aef9d2860aacefd20f27041c72c3569fb00/src/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp#L111-L126)

From the context of the code it seems that this was used at some point to give the toolbar some size in case no other items were in the toolbar yet.
However, I checked and I didn't find any weird functionallity when simply removing everything apart from 

```cpp
gtk_widget_show(this->w);
```

@Febbe : Do you know if this piece of code serves any other purpose that I didn't see?

If not I would vote for removing it.

---
https://github.com/xournalpp/xournalpp/blob/22e60aef9d2860aacefd20f27041c72c3569fb00/src/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp#L264-L269

Also seems weird to me but without it the position of dropped items in vertical toolbars becomes wrong.

Febbe: I think, setting the size should be done by setting the size directly, instead of inserting an dummy item.

The dummy item / setting of the size directly may be required to customize the toolbars.

The fix is nothing else than a base transformation, so it's correct, when the target base is the same as the calling widget of those signals.

Febbe: I vote for removing it. I checked the case I mentioned, and it does not make any difference.

idotobi: :sweat_smile: Wow that's old. Yeah I think that some of the css styling capabilities might have made this superfluous at some point.

I'll remove it then. 

I'll still analyze why https://github.com/xournalpp/xournalpp/blob/22e60aef9d2860aacefd20f27041c72c3569fb00/src/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp#L264-L269 is required, as I would naively expect that a event handler of a toolbar already creates properly transformed coordinates. However, if I don't get anywhere with this I'll update the PR the latest by the weekend.

Febbe: I think it has something to do with the fact, that drag and drop is used at minimum across the hole application. But first I am confused, why this is not documented by GTK anywhere, it is also not required to be in AppWindow coordinates, since this callback is only triggered for the connected widget.

idotobi: I played around with this some more and I'm starting to be convinced that this is a bug in gtk :see_no_evil: 

The x coordinates are handled within the coordinate system of the toolbar, which can be seen when customizing the floating toolbar.
Thus the current version had even a bug because it translated `x` coordinates as well (my last commit fixes this).
The y coordinates, however, somehow require the AppWindow coordinate system otherwise the `pos`ition is off (see vertical toolbars).

I added a wrapper around this weird behaviour of gtk in my latest commit. Hopefully this is verbose enough :innocent:

idotobi: @Febbe : I rebased the PR yesterday and I think it should be good to go again. What do you think?

src/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp
```diff
@@ -325,3 +299,18 @@ void ToolbarAdapter::toolbarDragDataReceivedCb(GtkToolbar* toolbar, GdkDragConte
...
+    // since the y coordinate is only relevant in vertical toolbars the transformation is NOT required for horizontal
+    // toolbars
+    if (horizontal) {
+        return gtk_toolbar_get_drop_index(toolbar, x, y);
+    } else {
+        gint wx = 0;
+        gint wy = 0;
+        gtk_widget_translate_coordinates(GTK_WIDGET(toolbar), gtk_widget_get_toplevel(GTK_WIDGET(toolbar)), x, y, &wx,
+                                         &wy);
+        return gtk_toolbar_get_drop_index(toolbar, x, wy);
```
LittleHuba, line 314:
```suggestion
        return gtk_toolbar_get_drop_index(toolbar, wx, wy);
```

I would feel more comfortable if you used both translated coordinates here. This might be a really hard to find bug, if at some point the x-coordinate becomes important (though I do not know if this would ever happen)

src/gui/toolbarMenubar/model/ToolbarEntry.cpp
LittleHuba: Please add a comment directly above the line of this bug and reference it to [this](https://github.com/xournalpp/xournalpp/pull/2980#discussion_r611072211) comment of yours. This way we will know why it was done this way.

Issue #2088, opened 2020-06-27
New Input System: Lines Disappearing


redweasel:
**Affects versions :**
 - OS: Linux Debian bullseye
 - Stylus Driver: xserver-xorg-input-wacom 1:0.36.1
 - Version of Xournal++ 1.0.18, Release and current master 1.1.0+

**Describe the bug**
If a line is drawn with full pressure and very fast from start to finish and then a new line is started the previous line sometimes disappears. This only happens with the new input system no matter which setting I tried. Ignore first stylus event is on in 1.1.0+ but it has the same behaviour in 1.0.18 where that feature was not yet implemented afaict.

This is a 3x slowed down screencapture of this happening. Note there are these strange lines which appear on starting the stroke that deletes the previous one.
![xournalpp_bug](https://user-images.githubusercontent.com/54547948/85921988-723ac400-b880-11ea-93bc-3f7e702a1908.gif)

Also I noticed that using the touchpad while before the deleting/terminating stroke will result in funny drawings when the pen is still hovering. When I terminate this faulty state with a rightclick of the pen which normally activates the lasso selection tool, it will keep the drawings.
![xournalpp_bug2](https://user-images.githubusercontent.com/54547948/85922379-07d75300-b883-11ea-8099-fb761ae9663c.gif)

I don't know how well it is visible on the gifs but all the faulty stokes have a thick ending. So basically the last segment of the stroke gets full thickness if the stylus is moved too fast.

The behaviour is a little different if pressure sensitivity is disabled, but in principle similar.

**To Reproduce**
Try doing the same as in the video.
1. draw a line with full pressure and very fast from start to finish
2. try moving the cursor with a touchpad
3. start next stroke to terminate the faulty one
4. repeat the process but terminate by switching the tool

Technius:
Reopened since #2984 was reverted in #3178.

PR #2984, opened 2021-03-29
Handle buggy driver events better to avoid vanishing lines

redweasel:
Fixes https://github.com/xournalpp/xournalpp/issues/2088

My pen device is weird and my driver is probably also weird (see issue for detail), so I had lots of problems with the new input system, that were not there with the old/removed input system. The problems include:
- disappearing last strokes when a new stroke is started
- touchpad input was drawing without pressed button after some strokes
- strokes sometimes become very jaggy and handwriting becomes unreadable
- when starting a new stroke a little spike to the last segment is drawn

These problems are all fixed for me by this patch. The problem was my driver sending some weird events. Here is a collection of things I recorded:
- Press event was sent, motion events were sent starting with positive pressure and ending with zero pressure, no release event was sent, next press was sent. This caused the last line to disappear.
- Press event was sent with positive pressure, motion events were sent with positive pressure, one motion event with negative pressure was sent, more motion events with positive pressure. This may have caused the jaggy unreadable lines?
- Press event was sent with positive pressure, motion events were sent with positive pressure, motion event with negative pressure was sent, motion events with zero pressure were sent, no release event. (weird common edge case on my device)

My solution was to first of all make sure the last stroke is ended when creating a new one. 
Also negative pressure motion events will be ignored if pressure sensitivity is on (may need to reflect that in the setting tooltip).

Then also end a stroke if the pressure is zero. This may not be necessary, but may be a problem for other devices, so please test this patch.

### DIFF
--- a/src/gui/inputdevices/StylusInputHandler.cpp
+++ b/src/gui/inputdevices/StylusInputHandler.cpp
@@ -27,6 +27,12 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
     if (event.type == BUTTON_PRESS_EVENT) {
 
         if (event.button == 1 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
+            if (this->eventsToIgnore < 0 && this->inputRunning) {
+                // end last action if it was not ended already
+                this->actionEnd(event);
+                g_message(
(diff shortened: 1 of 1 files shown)


Febbe:
We might want to filter those events before. Maybe in a `class FilterBuggedEvents: EventHandler`. 
This is a general refactor idea, your solution looks good to me.

PR #2990, opened 2021-03-30
allow using the system Lua package path

### LINKED ISSUE #2968: MigrateFontSizes plugin does not work with lua other than 5.3
This is with the current git version of xournalpp, any platform. The reason is:
- `src/plugin/Plugin.cpp` resets the module search path "for security and compatibility reasons"
- `MigrateFontSizes` needs a module (lgi) and sets the path assuming lua 5.3

This affects all plugins which need to require modules. Now, it would be easy to patch this plugin to use the lua `_VERSION`, but I'm wondering: If a plugin can simply set `package.path` then what is the point of resetting it in `Plugin.cpp`?

In fact, relying on the system path appears to be much more secure than allowing a plugin to change it freely, doesn't it?

rolandlo:
Fixes #2968

### DIFF
--- a/plugins/MigrateFontSizes/main.lua
+++ b/plugins/MigrateFontSizes/main.lua
@@ -3,9 +3,6 @@ function initUi()
   app.registerUi({["menu"] = "Migrate font sizes with factor displayDPI / 72", ["callback"] = "migrate"});
   app.registerUi({["menu"] = "Show font size migration dialog", ["callback"] = "showDialog"});
 
-  lgiInstallPaths = "/usr/share/lua/5.3/?.lua; /usr/local/share/lua/5.3/?.lua" -- change this if your lgi.lua file is located in a different directory
-  package.path = package.path .. ";../?.lua;" .. lgiInstallPaths 
-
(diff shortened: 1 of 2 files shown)


rolandlo:
@mjg If you have time, can you please check if the PR works with Lua 5.4?

mjg:
> @mjg If you have time, can you please check if the PR works with Lua 5.4?

Yep, this works. Only glitch I noticed if I show the font size dialogue: it says "72 / 72 = 3,0900" (old/norm=scale factor), and if I just "apply" it indeed scales by that factor. As soon as I change any number in this dialogue the other one is computed correctly, so it's just a wrong initial value, due to the fact that the plugin's `showDialog()` fails to call `ui.spbtScaleFactor:set_value(oldDpi/72)` or the like during init.

If, instead, I use "migrate font sizes with..." it correctly displays "72/72=1" and scales by a factor of 1.

rolandlo:
Thanks for testing and for catching this glitch @mjg. I fixed it.

Technius (approved):

PR #2997, opened 2021-04-02
Sidebar Toolbar: Fix abrupt background color change in Adwaita theme

personalizedrefrigerator:
## Before
image

## After
image

The `inline-toolbar` CSS class was previously applied to the parent of this toolbar (containing outline, layer preview, etc.), rather than the toolbar itself, causing a sudden change in background color. This PR moves the `inline-toolbar` class from the toolbar's parent to the toolbar itself.


Febbe:
Not sure if this is better now :shrug:

PR #3015, opened 2021-04-07
[FR] translate Discard into Ignorer

bobot:
`Annuler` is already used for `Cancel`. Otherwise when quitting the possibilities are `save`, `Annuler` and `Annuler`. I believe the one in the middle is `Discard`.


rolandlo:
Translations are handled in the [Crowdin project](https://crowdin.com/project/xournalpp). Please make your translation improvements there.

PR #3017, opened 2021-04-08
fix page preview for same-length PDFs

mjg:
If you open xournalpp (thus have a 1 page "empty" document) and then
annotate or open a 1 page PDF then the page preview is not updated. The
same happens if you open a document of any length and then annotate a
PDF with the same page length.

The reason is that SidebarPreviewPages::updatePreviews() does not do
anything if the number of pages has not changed. This check has been
there for a long time but is questionable: updatePreviews() is called
only for one reason, namely to update the previews.

So remove the check and fix the issue.

Note that this issue does not arise if you open another document
(instead of opening or annotating a PDF, which is the same): In this
case, the underlying Document object is newly created.

### DIFF
--- a/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
+++ b/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
@@ -187,11 +187,6 @@ void SidebarPreviewPages::updatePreviews() {
     doc->lock();
     size_t len = doc->getPageCount();
 
-    if (this->previews.size() == len) {
-        doc->unlock();
-        return;
-    }
-
     for (SidebarPreviewBaseEntry* p: this->previews) { delete p; }
     this->previews.clear();
 


mjg:
That was a weird one to pin down...

Marked as draft because the "real" fix might be different, namely tearing down the old Document object and creating a new one in annotatePdf the same way as when opening a new document. In any case, the page length comparison is not a good criterion for updating the previews, me thinks.

Technius:
I think a better solution would be to subclass DocumentListener so we can clear the preview when the document changes.

mjg:
Do you suggest to add something like a PreviewListener as a subclass and work with that? I'm not sure I have the necessary overview of the codebase to implement that.

But maybe nobody else has encountered this bug anyways, it's benign.

mjg:
I guess this can be closed now that the fix is in 2780815e8 ("fix sidebar page preview stuck", 2021-11-20) verbatim, though without the commit message. I should have learned by now that you don't like commit messages ;-)

Technius:
Sorry for leaving this alone for almost a year, @mjg. I have been busy in the past few months and didn't notice that the issue had been fixed in #3593. Ironically, the merged PR fixed _this_ issue but did not fix the bug that it was actually intended to fix. I'm going to close this PR as the change has already been incorporated.

mjg:
Yes, I noticed the patch is actually the same and fixed a different bug, too!

For none of us xournalpp is the main job paying the bills, I know the time constraints very well.

PR #3023, opened 2021-04-09
Toggle highlight position

mjg:
Here's a suggestion for a very lean implementation of a highlight pointer ("laser"): make the existing position highlighter (which so far is in the settings menu only) available as a quick toggle. This may partly serve requests like issue #1778 and PR #1822 until a new laser tool is implemented.

When there is a more general system for keybindings the plugin can be retired while the action itself may still be useful.

The best use for the keybinding may actually be: map a tablet (pad) button to that key sequence.


mjg:
In addition, I noticed that `STYLUS_CURSOR_NONE` does not draw the position highlight, and the settings dialog indicates that this is intentional. I'm wondering why, though?

Not only for the feature here, it can be useful to highlight the position without drawing the dot or pen. (In fact, for the laser I would like to toggle the dot off when I toggle the dot on.)

rolandlo:
`STYLUS_CURSOR_NONE` does not draw the position highlight because the highlight of the cursor position may add unnecessary visual clutter when writing with the pen on the screen (so you always know where the pen is physically).

mjg:
> `STYLUS_CURSOR_NONE` does not draw the position highlight because the highlight of the cursor position may add unnecessary visual clutter when writing with the pen on the screen (so you always know where the pen is physically).

I fully understand that not everyone wants the position highlight. It is off by default anyway. What I don't understand is: In the other two cursor modes (small dot, big pen) the user can turn the position hightlight on and off. Why shouldn't the user be allowed to make the same choice in the third cursor mode (CURSOR_NONE)?

rolandlo:
@mjg Ok, I get the question now. Frankly, I don't see a reason for not allowing cursor highlight in the "no dot" mode. Probably we just wanted to have the (destraction-free) no dot mode in #2111 and didn't think about allowing cursor highlight any further. 
@duncanawoods Please correct me if I am wrong.

duncanawoods:
The goal was a typical iPad/paper-and-pen like stylus experience with no visual indicator distracting from the writing experience. The goal was to not draw the cursor at all. From an implementation POV, the highlight is part of the cursor so the options are sub-configuration only relevant for visible cursors.

When it came to updating the settings page I had to support the default "dot" behaviour and the old "big icon" checkbox and it seemed clearest to create the combo box in terms of "icon" which I can see suggests the icon and highlight are independent when the original intent was the cursor as a whole.

It would make sense to adjust the cursor rendering code so that the "no icon" truly behaves as "no icon" instead of "no cursor". 

p.s. I was annoyed enough by a dot, I don't know what mad people want a big colourful splodge :grin:

mjg:
I had the same reaction as @duncanawoods when I first encountered the position highlight... It's just that it can be useful for presenting now (not writing).

I threw in a few highlighter fixes. But before introducing yet more API glues so that a plugin can turn off the cursor (the dot/pen) when turning on the highlight I'm wondering whether it makes more sense to go down the "tool path" right away: introduce a new tool that is just a "non-drawing pen" (called "laser" or "pointer") with the position highlight on (and no dot/pen/... cursor) that reuses as much of the existing code as possible. This avoids plugins, and later on it could be amended by a fading trail or other temporary painting which helps on platforms where the (highlight) cursor might not be captured by the recording/screen sharing.

Technius:
>  introduce a new tool that is just a "non-drawing pen" (called "laser" or "pointer") with the position highlight on (and no dot/pen/... cursor) that reuses as much of the existing code as possible

This is in line with the suggestions in #1822, and I think it is a pretty good solution that is worth trying out.

Technius (approved):
This seems to work reasonably well from some sort tests. The only problem I had was that I became confused about why the highlight toggle wasn't working with my mouse, when I remembered that the cursor highlight only works when drawing a stroke or hovering a stylus over...

A rebase and then this should be fine to merge, in my opinion.

@rolandlo Any other thoughts on this PR?

MichelReims:
I don't show hightlight position under windows 10

MichelReims:
Yes, I will try. I thought it was a laser

PR #2886, opened 2021-02-28
Merge in new action icons

Technius:
This is #1735 rebased on master, with the accompanying code changes. Huge thanks to @Fatih20 for the icon redesign, who contributed the new icons.

TODO:
~* [ ] Adjust icon license metadata to indicate they're under GPLv2~
* [x] Add option to set whether monochrome icons or colored icons are preferred
* [x] Check if the rebase selected the correct icons
* [x] Address feedback from comments below
* [ ] squash and rebase on current master

PR #3025, opened 2021-04-10
add tools to draw gauss/sinus/poly math functions

TimoJoFischer:
- tool to draw gauss functions
- tool to draw poly functions (x² x³ x⁴ x⁵)
- tool to draw sinus functions from 0 to 2pi/3pi/4pi/5pi

using the shift and ctrl key to change the properties

Would be nice if further math functions could be implemented or have a popup to set properties of a function.

Example:
Screenshot_20210411_015040


rolandlo (commented):
Thanks for this very nice contribution! It looks pretty useful for teaching math. I think it fits very well into Xournal++. The code looks clean.

Among more functions to add I would consider particularly:
1. power functions with negative exponent (-1, -2, -3, -4) 
2. exponential function

Minor nitpick: The `svg` for the sine function doesn't look easy to recognize (it resembles a handwritten letter m). Maybe use sin(x) with x in [0, 2*pi]

In #2886 we will introduce new icons. The icons have a more systematic naming scheme. According to that naming scheme, your icons should be named `draw-poly`, `draw-gauss` and `draw-sinus`. Better use the new naming scheme, as otherwise we will soon have to change it.

TimoJoFischer:
:warning: This should be improved

I now added power functions with negative exponent, ln and exp

Problem: this functions are only working correctly for a specific range. I don't know how to fix it. e.g 1/x only draw the correct function in range [0.25-4]. Has someone a idea?

Example:
Screenshot_20210411_174541

rolandlo:
> warning This should be improved
> 
> I now added power functions with negative exponent, ln and exp
> 
> Problem: this functions are only working correctly for a specific range. I don't know how to fix it. e.g 1/x only draw the correct function in range [0.25-4]. Has someone a idea?

Yes, I have an idea (which seems to work). For power functions with negative exponents the second point should not be on the graph, but it should be the intersection point of the asymptotes (i.e. the coordinate system origin). 

For instance for 1/x:
```C++
            // 1/x
            for (int j = 0; j <= npts; j++) {
                double x = 0.25 + j*3.75/npts;
                double y = 1/x;
                stroke->addPoint(Point(start_x-(x-1)*width, start_y-(y-1)*height));
            }
```

mjg:
This is certainly nice, but limited to a hardcoded set of functions and does not scale. I'm wondering how much it would take to have something more like the LaTeX tool: a graph plotter taking a lua expression. This would also solve your issue with the possible range: the user would have to decide :)

Technius:
> This is certainly nice, but limited to a hardcoded set of functions and does not scale. I'm wondering how much it would take to have something more like the LaTeX tool: a graph plotter taking a lua expression.

This would probably make a good candidate for a plugin that invokes a program like gnuplot.

rolandlo:
As far as I see this could easily be generalized. I will use parametric functions (x(t), y(t)). Ordinary functions are included via settings x(t)=t and y(t) the function term in the variable t.

Each function is defined by:
- function terms `x(t)` and `y(t)`
- one interval [`t_min`, `t_max`] in which the image of the parametric curve ought to be drawn (in future we may allow more than one interval)
- two points `(x0, y0)` and `(x1,y1)` with `dx:=x1-x0` and `dy:=y1-y0` non-zero. These coordinates are defined with respect to the coordinate system in which x(t) and y(t) live, and will made to correspond to the point `startPoint` which is placed first on the screen and the current cursor position `currentPoint`, respectively.
- a constant `fpts` used for computing the number of points required to draw the curve

 
Then we use the following pseudo-code to generate the points on the Stroke

``` 
set a constant minimal number of points nptsMin (e.g. nptsMin = 24)
set width = currentPoint.x - startPoint.x
set height = currentPoint.y - startPoint.y
set npts = max (fpts* (length of diagonal of rectangle with above width and height), nptsMin) 

remove previous stroke points

sample t linearly in the interval [t_min, t_max] with npts samples
for each sample t 
   evaluate x(t) and y(t) and set
   p.x = startPoint.x + (x-x0)*width/dx
   p.y = startPoint.y + (y-y0)*height/dy
   add the point (p.x,p.y) to the stroke
draw the stroke
```

Ideally we would have a flexible solution like with the toolbars. 

- some .ini-file (or similar) where you can define your (sets of 1-4) function and names of svg-files
- the .ini-file is read and the menu entries/toolbar combo entries are created at startup
- some dialog where you can define a function inside Xournal++

The .ini file might look roughly as follows:

```.ini
[Some user defined power functions, exponents are 2,3,-1 and -2]
name=power functions
name[de]=Potenzfunktionen
icon = draw-negPower.svg

# first curve, drawn with no modifier
x[1] = t
y[1] = t^2
domain[1] = [-3:3]
firstPoint[1] = (0,0)
secondPoint[1] = (2.5, 6.25)
fpts[1] = 2.0

# second curve, drawn with control modifier
x[2] = t
y[2] = t^3
domain[2] = [-2:2]
firstPoint[2] = (0,0)
secondPoint[2] = (2.5,15.625)
fpts[2] = 3.0

# third curve, drawn with shift modifier
x[3] = t
y[3] = t^(-1)
domain[3] = [0.25:4]
firstPoint[3] = (1,1)
secondPoint[3] = (0,0)
fpts[3] = 1.5

# fourth curve, drawn with control and shift modifiers
x[4] = t
y[4] = t^(-2)
domain[4] = [0.25:4]
firstPoint[4] = (1,1)
secondPoint[4] = (0,0)
fpts[4] = 2.4
```

In that way, everybody could use his own preferred functions and parametric curves while it would still be fast to draw these graphs/parametric curves.

personalizedrefrigerator:
> some dialog where you can define a function inside Xournal++

I agree! This would be very nice!

Technius:
Thinking about this more, I think we really should handle this as a plugin and not as a built-in feature. I agree with @mjg's concerns about scalability and @rolandlo's comments about customizability. Here are some of my thoughts:
* We could extend the plugin API to allow plugins to create custom stroke elements, which could then be used for plotting functions.
* Since we want plotting to be a tool, we could also allow plugins to register custom stroke types (e.g. similar to the Draw Ellipse, Draw Rectangle, etc. types that we already have).
* To account for different parameters, we will need some way to bring up a "tool customization" menu of sorts, either as a built-in or with a custom plugin API.
This is quite a bit more work than what is included with just this PR, but I feel like it would open up a lot of interesting possibilities.

rolandlo:
Ok, let us do some brainstorming how we could implement this via plugins.

So the idea is to let a plugin determine what stroke (e.g. function graph, a particular shape or whatever) is drawn depending on the user input and the C++ code takes care of all details.

So we could write a `PluginShapeHandler` and a `PluginPlotHandler` class derived from the `InputHandler` class that communicate with the plugin and do the actual rendering, input handling, snapping, tapping detection, undo action management, zooming etc. The Lua Plugin API would define two functions `app.registerShape` and `app.registerPlot` which would specify everything needed.

Here is a (purely fictional) example plugin using both a custom shape and a function plotter:

```lua
-- Register shape and function plots
function initUi()
app.registerShape({["name"]="triangle shape", ["icon"]="arc.svg", ["motion"]="mo", ["motionDrag"]="md", ["buttonPress"]=bp, ["buttonRelease"]=br})

app.registerPlot({["name"]="power functions", ["icon"]="powerFunctions.svg", ["domain"]="dom", ["x"] ="xCoord", ["y"]="yCoord",  ["firstPoint"] = "fp", ["secondPoint"] = "sp"})
end


--- code for shape
-- A simple implementation of a circle arc
-- the first button press point determines the center; 
-- the first button release point determines the start point of the circle arc
-- the second button press point projected onto the circle determines the end of the circle arc

local angles = {}    -- stores the angles defining the start and end point of the circle arc
local center = {}    -- stores the coordinates of the center of the circle arc
local radius         -- the radius of the circle arc

function circleArc(c,r, a)   -- helper function to produce circle arc points
  local stroke={c}
  local da=a[2]-a[1]
  local n = math.max(24, math.floor(2*r))
  for i = 1, n do
   table.append(stroke, {c[1]+r*cos(a[1]+da/n*i), c[2]+r*sin(a[1]+da/n*i)}) 
  end 
  return stroke
end

function mo(x,y)   -- react on motion without dragging
    local dx = x - center[1]
    local dy = y - center[2]
    return circleArc(center, radius, {angles[1], math.atan(dy,dx)})   -- temporary circle arc
end

function md(x,y)   -- react on dragging
  local dx = x - center[1]
  local dy = y - center[2]
  local dist = math.sqrt(dx^2+dy^2)
  return circleArc(center, dist, 0, 2*math.pi)   -- temporary circle
end

function bp(x,y) -- react on buttonPressEvent
 
  if #center== 0 then
    center == {x,y}  
    return {}  -- nothing to draw yet
  else
    local dx = x - center[1]
    local dy = y - center[2]
    angles[2] = math.atan(dy,dx)
    return circleArc(center, radius, angles), true    -- definite circle arc; second argument "true" means drawing is finished
  end

end

function br(x,y)  -- react on buttonReleaseEvent
  local dx = x - center[1]
  local dy = y - center[2]
  radius = math.sqrt(dx^2+dy^2)
  angles[1] = math.atan(dy,dx)
  return circleArc(center, radius, 0, 2*math.pi)   -- definite circle on which the arc is 
end


-- code for plotter
-- Some user defined power functions, exponents are 2,3,-1 and -2

function dom(mod)
  local domains = {{-3,3}, {-2,2}, {0.25,4}, {0.25,4}} 
  return domains[mod]
end

function xCoord(t, mod)
  return t
end

function yCoord(t, mod)
  if mod == 1 then return t^2
  elseif mod == 2 then return t^3
  elseif mod == 3 then return t^(-1)
  elseif mod == 4 then return t^(-2)
  end 
end

function fp(mod)
  local points = {{0,0}, {0,0}, {1,1}, {1,1}}
  return points[mod]
end

function sp(mod)
  local points = {{2.5,6.25}, {2.5,15.625}, {0,0},{0,0}}
  return points[mod]
end
```

If you like it visually (for the circle arc):
grafik

Fatih20:
> `draw-poly`, `draw-gauss` and `draw-sinus`

If this feature is finished, I'll make a pull request for these icons.

TimoJoFischer:
this feature can still be improved, but right now i dont have any time

wolfallein:
Hi,

This is an amazing feature! I'll certainly use it a lot. Looking forward to seeing it implemented.

Thank you for the effort.

rolandlo:
> This is certainly nice, but limited to a hardcoded set of functions and does not scale. I'm wondering how much it would take to have something more like the LaTeX tool: a graph plotter taking a lua expression. This would also solve your issue with the possible range: the user would have to decide :)

I have written a basic function plotter realized via a Lua plugin, see #3811. It doesn't make the present PR obsolete, but it's certainly related.

profmatxavier:
Excelente. Como faço para utilizar este incrível recurso no meu Xournal no Windows 10?

kaliasgoldmedal:
> * tool to draw gauss functions
> * tool to draw poly functions (x² x³ x⁴ x⁵)
> * tool to draw sinus functions from 0 to 2pi/3pi/4pi/5pi
> 
> using the shift and ctrl key to change the properties
> 
> Would be nice if further math functions could be implemented or have a popup to set properties of a function.
> 
> Example: Screenshot_20210411_015040

How to install this version. I have installed downloaded exe in my desktop. But this plugin is not available. Please help me

kaliasgoldmedal:
How to install this version. I have installed downloaded exe in my desktop. But this plugin is not available. Please help me

PR #3033, opened 2021-04-13
Fix simple implicit type conversions

TuningYourCode:
My first contribute to a c++ project - hopefully quiet some "useful".

Only changed places which seem to be local variables and were using wrong type. 

My IDE (Clion - JetBrains) does no longer complain about incorrect type of these variables, building works and application was tested for some minutes with different tasks including wacom tablet.

If my contribution is ok for you i could also try to do more major stuff like easy issues or fixing some of the other implicit type conversions. 

For me as beginner to this project it is quiet strange why page-related variables (count, current page id etc.) are (signed) integer instead of size_t which should be on all common plattforms at least that big like int and will show any errors/tips in IDEs.

### DIFF
--- a/src/model/Document.cpp
+++ b/src/model/Document.cpp
@@ -255,11 +255,11 @@ auto Document::fillPageLabels(GtkTreeModel* treeModel, GtkTreePath* path, GtkTre
         return false;
     }
 
-    int page = doc->findPdfPage(link->dest->getPdfPage());
+    auto page = doc->findPdfPage(link->dest->getPdfPage());
 
     gchar* pageLabel = nullptr;
     if (page != -1) {
-        pageLabel = g_strdup_printf("%i", page + 1);
+        pageLabel = g_strdup_printf("%lu", page + 1);
     }
     gtk_tree_store_set(GTK_TREE_STORE(treeModel), iter, DOCUMENT_LINKS_COLUMN_PAGE_NUMBER, pageLabel, -1);
     g_free(pageLabel);
--- a/src/model/Document.cpp
+++ b/src/model/Document.cpp
@@ -258,7 +258,7 @@ auto Document::fillPageLabels(GtkTreeModel* treeModel, GtkTreePath* path, GtkTre
     auto page = doc->findPdfPage(link->dest->getPdfPage());
 
     gchar* pageLabel = nullptr;
-    if (page != -1) {
+    if (page != npos) {
         pageLabel = g_strdup_printf("%lu", page + 1);
     }
     gtk_tree_store_set(GTK_TREE_STORE(treeModel), iter, DOCUMENT_LINKS_COLUMN_PAGE_NUMBER, pageLabel, -1);
--- a/src/undo/InsertDeletePageUndoAction.cpp
+++ b/src/undo/InsertDeletePageUndoAction.cpp
@@ -70,7 +70,7 @@ auto InsertDeletePageUndoAction::deletePage(Control* control) -> bool {
     // It's not great practise but it works.
     // doc->lock();
     auto pNr = doc->indexOf(page);
-    if (pNr == -1) {
+    if (pNr == npos) {
         //	doc->unlock();
         // this should not happen
         return false;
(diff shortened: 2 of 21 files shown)


mjg:
Fixing these warnings is certainly nice because it helps catch real "errors". I'm wondering, though, whether `auto` is the more readable solution for most of these. The codebase is "organically grown", so you will see different styles. But, e.g. the local variable in a `for` loop is a good candidate for `auto` (unless you go full throttle and convert to range loops).

Febbe:
> I'm wondering, though, whether auto is the more readable solution for most of these.

This is indeed an opinionated question and there is no correct answer to it, I can only throw in my opinion: Using auto everywhere is better, since it prevents many casting errors. Also as maintainer, I want to prove the logic of the implementation and not the semantic. Using auto means for me, that I don't need to bother, if the type is correct. There is also a blog post of Herb Sutter to this: https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

Other maintainers here have of course another opinion. Every one of us can handle the AAA-Style and the opposite, only if it's too unreadable for the reviewer, you must change this. Often it is marked as **nitpick**.

Febbe:
Very good, we will merge this after we released version 1.1.0. This was also on my to-do list.

rolandlo:
@TuningYourCode Can you please rebase on current master and solve the merge conflicts?

src/model/Document.cpp
```diff
@@ -255,11 +255,11 @@ auto Document::fillPageLabels(GtkTreeModel* treeModel, GtkTreePath* path, GtkTre
         return false;
     }
 
-    int page = doc->findPdfPage(link->dest->getPdfPage());
+    auto page = doc->findPdfPage(link->dest->getPdfPage());
 
     gchar* pageLabel = nullptr;
     if (page != -1) {
```
rolandlo, line 261:
```suggestion
    if (page != npos) {
```

src/undo/InsertDeletePageUndoAction.cpp
```diff
@@ -69,7 +69,7 @@ auto InsertDeletePageUndoAction::deletePage(Control* control) -> bool {
     // unlocking the UndoRedoHandler's lock inside here.
     // It's not great practise but it works.
     // doc->lock();
-    int pNr = doc->indexOf(page);
+    auto pNr = doc->indexOf(page);
     if (pNr == -1) {
```
rolandlo, line 73:
```suggestion
    if (pNr == npos) {
```

src/undo/PageBackgroundChangedUndoAction.cpp
```diff
@@ -30,7 +30,7 @@ auto PageBackgroundChangedUndoAction::undo(Control* control) -> bool {
     this->newH = this->page->getHeight();
 
     Document* doc = control->getDocument();
-    int pageNr = doc->indexOf(this->page);
+    auto pageNr = doc->indexOf(this->page);
     if (pageNr == -1) {
```
rolandlo, line 34:
```suggestion
    if (pageNr == npos) {
```

src/undo/PageBackgroundChangedUndoAction.cpp
```diff
@@ -55,7 +55,7 @@ auto PageBackgroundChangedUndoAction::undo(Control* control) -> bool {
 auto PageBackgroundChangedUndoAction::redo(Control* control) -> bool {
     Document* doc = control->getDocument();
 
-    int pageNr = doc->indexOf(this->page);
+    auto pageNr = doc->indexOf(this->page);
 
     if (pageNr == -1) {
```
rolandlo, line 60:
```suggestion
    if (pageNr == npos) {
```

rolandlo (changes requested):
I have some nitpicks, see in the code. The rest looks good to me.

rolandlo:
Anything else you'd like to address in this PR? Or is it ready to be merged from your side?

TuningYourCode:
Feel free to merge it :) If i find time again i will just open another PR.

rolandlo:

Issue #3020, opened 2021-04-09
Option for opening last used file


cylonrob:
First of all, this app is a lifesaver for remote teaching.

For users who frequently work on the same file over many sessions it would  be convenient to have an option to automatically load the last used file. 

My personal solution works, but it's obviously not general; it was to make an alias 

`alias xournalpp $(grep /home /home/robert/.xournalpp/metadata/$(ls -t /home/robert/.xournalpp/metadata/ | head -n 1))`

but an official feature would be nice. Thanks for a great program!

PR #3040, opened 2021-04-15
Add open most recent document on application startup

TuningYourCode:
Implements #3020.

I'm open for improvements as it is my second PR for a c++ project.

Changes on the settings dialog could be adjusted to be a completely new container to not touch existing things but i though it makes sense to put this setting and autoloading journal (pdf) into the same container.

image

### DIFF
--- a/src/control/RecentManager.cpp
+++ b/src/control/RecentManager.cpp
@@ -105,6 +105,10 @@ auto RecentManager::getMostRecent() -> GtkRecentInfo* {
     GList* filteredItemsXoj = filterRecent(gtk_recent_manager_get_items(gtk_recent_manager_get_default()), true);
     auto mostRecent = static_cast<GtkRecentInfo*>(filteredItemsXoj->data);
 
+    gtk_recent_info_ref(mostRecent);
+    for (GList* l = filteredItemsXoj; l != nullptr; l = l->next) {
+        gtk_recent_info_unref(static_cast<GtkRecentInfo*>(l->data));
+    }
     g_list_free(filteredItemsXoj);
 
     return mostRecent;
--- a/src/control/RecentManager.cpp
+++ b/src/control/RecentManager.cpp
@@ -101,6 +101,15 @@ auto RecentManager::sortRecentsEntries(GtkRecentInfo* a, GtkRecentInfo* b) -> gi
     return tp_a != tp_b ? (tp_a < tp_b ? 1 : -1) : 0;
 }
 
+auto RecentManager::getMostRecent() -> GtkRecentInfo* {
+    GList* filteredItemsXoj = filterRecent(gtk_recent_manager_get_items(gtk_recent_manager_get_default()), true);
+    auto mostRecent = static_cast<GtkRecentInfo*>(filteredItemsXoj->data);
+
+    g_list_free(filteredItemsXoj);
+
+    return mostRecent;
+}
+
 auto RecentManager::filterRecent(GList* items, bool xoj) -> GList* {
     GList* filteredItems = nullptr;
 
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -560,6 +560,10 @@ void on_startup(GApplication* application, XMPtr app_data) {
             XojMsgBox::showErrorToUser(static_cast<GtkWindow*>(*app_data->win), msg);
             opened = app_data->control->newFile("", p);
         }
+    } else if (app_data->control->getSettings()->isAutoloadMostRecent()) {
+        if (auto p = Util::fromUri(gtk_recent_info_get_uri(app_data->control->getRecentManager()->getMostRecent()))) {
+            opened = app_data->control->openFile(*p);
+        }
     }
 
     app_data->control->getScheduler()->start();
--- a/ui/settings.glade
+++ b/ui/settings.glade
@@ -645,14 +645,15 @@ If the file was not yet saved you can find it in your home directory, in ~/.xour
                                         <property name="can_focus">False</property>
                                         <property name="orientation">vertical</property>
                                         <child>
-                                          <object class="GtkLabel" id="sid15">
+                                          <object class="GtkCheckButton" id="cbAutoloadXoj">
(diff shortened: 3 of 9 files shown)


src/control/RecentManager.cpp
```diff
@@ -101,6 +101,12 @@ auto RecentManager::sortRecentsEntries(GtkRecentInfo* a, GtkRecentInfo* b) -> gi
     return tp_a != tp_b ? (tp_a < tp_b ? 1 : -1) : 0;
 }
 
+auto RecentManager::getMostRecent() -> GtkRecentInfo * {
```
personalizedrefrigerator, line 104: `clang-format`'s complaining about this line.
See https://dev.azure.com/xournalpp/xournalpp/_build/results?buildId=12744&view=logs&j=8af2c752-b8c3-5474-5b0a-7b1f58b7bf90&t=0928bceb-a691-5a8f-98f6-852338e7281a

Aside from that, this PR looks good to me!

src/control/XournalMain.cpp
```diff
@@ -557,6 +557,12 @@ void on_startup(GApplication* application, XMPtr app_data) {
             XojMsgBox::showErrorToUser(static_cast<GtkWindow*>(*app_data->win), msg);
             opened = app_data->control->newFile("", p);
         }
+    } else if (app_data->control->getSettings()->isAutoloadMostRecent()) {
+        auto p = Util::fromUri(gtk_recent_info_get_uri(app_data->control->getRecentManager()->getMostRecent()));
+
+        if (p) {
+            opened = app_data->control->openFile(*p);
+        }
```
Febbe, line 565: We try to modernize our codebase, so we want new code as modern as possible:
Please use [cpp-17 - init-statements](https://en.cppreference.com/w/cpp/language/if)

```suggestion
        if (auto p = Util::fromUri(gtk_recent_info_get_uri(app_data->control->getRecentManager()->getMostRecent()));
            p) {
            opened = app_data->control->openFile(*p);
        }
```

Febbe (changes requested):
The c++ code of your PR looks good, but can you describe, why you changed all names from minus-case to snake_case in the ui files?
If you didn't do this, please try to use the newest glade version.

ui/settings.glade
```diff
@@ -1,201 +1,201 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.22.2 -->
```
personalizedrefrigerator, line 2: For a smaller diff, consider using a newer version of glade. Version 3.22 uses '_'s in many places where 3.38 uses '-'s. :)

TuningYourCode:
thanks for your feedback and sorry for the delay. should be fixed now and even the changes to settings.glade should be less as of Glade upgrade on my side.

src/control/RecentManager.cpp
```diff
@@ -101,6 +101,12 @@ auto RecentManager::sortRecentsEntries(GtkRecentInfo* a, GtkRecentInfo* b) -> gi
     return tp_a != tp_b ? (tp_a < tp_b ? 1 : -1) : 0;
 }
 
+auto RecentManager::getMostRecent() -> GtkRecentInfo* {
+    GList* filteredItemsXoj = filterRecent(gtk_recent_manager_get_items(gtk_recent_manager_get_default()), true);
+
+    return static_cast<GtkRecentInfo*>(filteredItemsXoj->data);
```
Technius, line 107: I think this will leak memory.

Technius (commented):
You may need to rebase on master due to #3050.

src/control/RecentManager.cpp
TuningYourCode: i'm not that used to c++. Would make sense to extend `filterRecent` to give it an amount/count of needed results (in this case 1) and then only return only the filepath and before returning destroying the single `GtkRecentInfo` object or which flow would you suggest to not leak memory in this case?

Technius: In my opinion, the easiest change is to just free the other recent items and then keep the recent info that we want.
1. Store the `GtkRecentInfo*` we want in a local variable, and then set the corresponding data pointer of the GList to null.
2. Call [`g_free_list_full`](https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-free-full) on `filteredItemsXoj`
3. Return the saved `GtkRecentInfo*`.

Technius: This is really more like C rather than C++, since glib and GTK are C libraries. See https://developer.gnome.org/programming-guidelines/stable/memory-management.html.en

TuningYourCode: hopefully i changed everything as it should be.

i'm more used to "simple" languages like php, javascript and ruby. :)

src/control/RecentManager.cpp
```diff
@@ -101,6 +101,15 @@ auto RecentManager::sortRecentsEntries(GtkRecentInfo* a, GtkRecentInfo* b) -> gi
     return tp_a != tp_b ? (tp_a < tp_b ? 1 : -1) : 0;
 }
 
+auto RecentManager::getMostRecent() -> GtkRecentInfo* {
+    GList* filteredItemsXoj = filterRecent(gtk_recent_manager_get_items(gtk_recent_manager_get_default()), true);
+    auto mostRecent = static_cast<GtkRecentInfo*>(filteredItemsXoj->data);
+
+    g_list_free(filteredItemsXoj);
+
```
Technius, line 109: Need to free the unused list elements.
```suggestion

    gtk_recent_info_ref(mostRecent);
    for (GList* l = filteredItemsXoj; l != nullptr; l = l->next) {
        gtk_recent_info_unref(static_cast<GtkRecentInfo*>(l->data));
    }
    g_list_free(filteredItemsXoj);

```

Technius (changes requested):
Sorry for the delay, I have been very busy the past month. The PR looks good to me except for one thing.

Technius (approved):
LGTM. Thanks for the PR!

PR #3042, opened 2021-04-16
Replace some GList with STL structs

Marmare314:
Since I am new to the Xournal++ code I thought I could get started by doing some refactoring :)

- ClipboardHandler
    - replaced GList by std::set because it is sorted
    - added some static_cast's to make clang-tidy happy
- RegionSelect
    - replaced GList by std::vector since only push_back is needed
    - **reserve has not been called**


Marmare314:
Unfortunately I can't seem to figure out how to install clang-format-12 on Linux Mint 20.
And apparently clang-format-10 with 'AlignOperands: true' yields a different result.

rolandlo:
> Unfortunately I can't seem to figure out how to install clang-format-12 on Linux Mint 20.
> And apparently clang-format-10 with 'AlignOperands: true' yields a different result.

I fixed it. Do you have `clang-format-11` on Linux Mint 20? That may be enough.

Marmare314:
It does! clang-format-11 seems to work fine while git-clang-format-11 complains about the AlignOperators Argument.
But I guess that's good enough.

Febbe:
you can install the newest (not-stable) version, which brings meta-packages from here: https://apt.llvm.org/
 
> git-clang-format-11 complains about the AlignOperators Argument
 
There is a bug (in my opinion);  git-clang-format-11 calls the meta-package clang-format, instead of clang-format-11. When you use the canonical provided meta package, you'll have clang-format-10 installed. so git-clang-format-11 will call clang-format which calls clang-format-10.

Installing the llvm provided meta-package will fix everything.

Marmare314:
I got clang-format-12 running now, however I had no luck with the meta-package. 
But `git-clang-format --binary clang-format-12` works well.

Also: If I do some similar refactoring, should I create a new pull request, or push it to this one?

rolandlo:
> I got clang-format-12 running now, however I had no luck with the meta-package.
> But `git-clang-format --binary clang-format-12` works well.
> 
> Also: If I do some similar refactoring, should I create a new pull request, or push it to this one?

I suggest you push it to this one. Just squash the commits which belong together and keep commits apart, if they belong to different refactoring steps.

Marmare314:
I think that should be everything worth replacing.

src/control/xml/XmlNode.h
```diff
@@ -20,11 +24,9 @@ class ProgressListener;
 class XmlNode {
 public:
     XmlNode(const char* tag);
-    virtual ~XmlNode();
 
-private:
-    XmlNode(const XmlNode& node);
-    void operator=(const XmlNode& node);
+    XmlNode(const XmlNode& node) = delete;
+    void operator=(const XmlNode& node) = delete;
```
Febbe, line 29: Since you are using `std::unique_ptr` those are deleted anyways. But defining them as deleted requires to fulfill the rule of 5. So best would be to remove them. This will implicitly allow to move the object.

src/control/xml/XmlPointNode.h
```diff
@@ -11,23 +11,23 @@
...
 class XmlPointNode: public XmlAudioNode {
 public:
     XmlPointNode(const char* tag);
-    virtual ~XmlPointNode();
 
-private:
-    XmlPointNode(const XmlPointNode& node);
-    void operator=(const XmlPointNode& node);
+    XmlPointNode(const XmlPointNode& node) = delete;
+    void operator=(const XmlPointNode& node) = delete;
```
Febbe, line 25: Same here, remove or fulfill the rule of 5.

src/control/xml/XmlPointNode.cpp
```diff
@@ -23,13 +14,12 @@ void XmlPointNode::writeOut(OutputStream* out) {
 
     out->write(">");
 
-    for (GList* l = this->points; l != nullptr; l = l->next) {
-        auto* p = static_cast<Point*>(l->data);
-        if (l != this->points) {
+    for (auto pointIter = points.begin(); pointIter != points.end(); pointIter++) {
+        if (pointIter != points.begin()) {
```
Febbe, line 18: I would move the condition out of the loop, also prepend the ++ (increment) in c++, because val++ will create a copy of val before increment. And not every compiler is capable to remove the extra code.

src/control/xml/XmlPointNode.cpp
```diff
@@ -2,18 +2,9 @@
...
-    for (GList* l = this->points; l != nullptr; l = l->next) {
-        auto* p = static_cast<Point*>(l->data);
-        delete p;
-    }
-    g_list_free(this->points);
-    this->points = nullptr;
-}
-
-void XmlPointNode::addPoint(const Point* point) { this->points = g_list_append(this->points, new Point(*point)); }
+void XmlPointNode::addPoint(const Point* point) { points.emplace_back(*point); }
```
Febbe, line 7:
```suggestion
void XmlPointNode::addPoint(Point point) { points.emplace_back(std::move(point)); }
```

src/control/xojfile/SaveHandler.cpp
```diff
@@ -333,11 +308,9 @@ void SaveHandler::saveTo(OutputStream* out, const fs::path& filepath, ProgressLi
     out->write("<?xml version=\"1.0\" standalone=\"no\"?>\n");
     root->writeOut(out, listener);
 
-    for (GList* l = this->backgroundImages; l != nullptr; l = l->next) {
-        auto* img = static_cast<BackgroundImage*>(l->data);
-
-        auto tmpfn = (fs::path(filepath) += ".") += img->getFilepath();
-        if (!gdk_pixbuf_save(img->getPixbuf(), tmpfn.u8string().c_str(), "png", nullptr, nullptr)) {
+    for (BackgroundImage& img: backgroundImages) {
```
Febbe, line 311: img is only used as const&:
```suggestion
    for (BackgroundImage const& img: backgroundImages) {
```

Febbe (changes requested):
Here, a bulk of change requests, thank you for the PR

src/undo/PageLayerPosEntry.h
```diff
@@ -11,22 +11,17 @@
...
-        this->pos = pos;
-        this->layer = layer;
-    }
+class Layer;
 
-    ~PageLayerPosEntry() {}
+template <class T>
+struct PageLayerPosEntry {
+    // TODO: constructor could be removed with C++20
+    PageLayerPosEntry<T>(Layer* layer, T* element, int pos): layer(layer), element(element), pos(pos) {}
```
Febbe, line 19: Are you sure? I think this is already replaceable since c++14 or at least c++17.

src/undo/EraseUndoAction.h
```diff
@@ -11,34 +11,33 @@
...
-    virtual string getText();
+    string getText() override;
 
 private:
-    GList* edited = nullptr;
-    GList* original = nullptr;
+    std::multiset<PageLayerPosEntry<Stroke>, decltype(&PageLayerPosEntry<Stroke>::cmp)> edited{
+            PageLayerPosEntry<Stroke>::cmp};
+    std::multiset<PageLayerPosEntry<Stroke>, decltype(&PageLayerPosEntry<Stroke>::cmp)> original{
+            PageLayerPosEntry<Stroke>::cmp};
```
Febbe, line 42: This is wrong, using a three-way comparator here will violate the [Compare Requirement](https://en.cppreference.com/w/cpp/named_req/Compare).
Generally, I prefer structs to functions, because it is not required to initialize them manually.

```suggestion
    std::multiset<PageLayerPosEntry<Stroke>> edited{};
    std::multiset<PageLayerPosEntry<Stroke>> original{};
```

src/undo/EraseUndoAction.cpp
```diff
@@ -1,86 +1,48 @@
...
-    for (GList* l = this->original; l != nullptr;) {
-        auto* p = static_cast<PageLayerPosEntry<Stroke>*>(l->data);
-        GList* del = l;
-        l = l->next;
-
-        if (p->element->getPointCount() == 0) {
-            this->edited = g_list_delete_link(this->edited, del);
-            delete p;
-            p = nullptr;
+    for (auto entryIter = original.begin(); entryIter != original.end(); entryIter++) {
```
Febbe, line 25: why not using a foreach loop here:

```suggestion
    for (auto&& entry: original) {
```

src/undo/EraseUndoAction.cpp
```diff
@@ -1,86 +1,48 @@
...
-            for (GList* ls = stroke; ls != nullptr; ls = ls->next) {
-                auto* copy = static_cast<Stroke*>(ls->data);
-                p->layer->insertElement(copy, pos);
-                this->addEdited(p->layer, copy, pos);
+            int pos = static_cast<int>(entryIter->layer->removeElement(entryIter->element, false));
+
+            EraseableStroke* e = entryIter->element->getEraseable();
+            std::vector<std::unique_ptr<Stroke>> strokeList = e->getStroke(entryIter->element);
+            for (auto& stroke: strokeList) {
+                // TODO (Marmare314): should use unique_ptr in layer
```
Febbe, line 36: Thank you for the todo :); Yes that is a refactoring, which should go into another pr :)

src/undo/DeleteUndoAction.h
```diff
@@ -35,6 +32,7 @@ class DeleteUndoAction: public UndoAction {
     string getText() override;
 
 private:
-    GList* elements = nullptr;
+    std::multiset<PageLayerPosEntry<Element>, decltype(&PageLayerPosEntry<Element>::cmp)> elements{
+            PageLayerPosEntry<Element>::cmp};
```
Febbe, line 36: Like below this violates the Compare Requirement. (Reference below)
```suggestion
    std::multiset<PageLayerPosEntry<Element>> elements{};
```

src/undo/AddUndoAction.h
```diff
@@ -35,6 +32,7 @@ class AddUndoAction: public UndoAction {
     string getText() override;
 
 private:
-    GList* elements = nullptr;
+    std::multiset<PageLayerPosEntry<Element>, decltype(&PageLayerPosEntry<Element>::cmp)> elements{
+            PageLayerPosEntry<Element>::cmp};
```
Febbe, line 36: Same here.

src/model/eraser/EraseableStroke.cpp
```diff
@@ -243,88 +231,75 @@ auto EraseableStroke::erasePart(double x, double y, double halfEraserSize, Erase
...
+            if (!strokeList.empty()) {
+                strokeList.back()->addPoint(lastPoint);
             }
-            s = new Stroke();
-            s->setColor(original->getColor());
-            s->setToolType(original->getToolType());
-            s->setLineStyle(original->getLineStyle());
-            s->setWidth(original->getWidth());
-            list = g_list_append(list, s);
+            auto& newStroke = strokeList.emplace_back(std::move(std::make_unique<Stroke>()));
```
Febbe, line 291: `std::make_unique<Stroke>()` is already a rvalue-reference, no need to call move here.
```suggestion
            auto& newStroke = strokeList.emplace_back(std::make_unique<Stroke>());
```

Febbe (changes requested):

Febbe:
remove `po/xournalpp.pot` from the commit history

src/model/eraser/EraseableStrokePart.h
```diff
@@ -11,33 +11,24 @@
...
-
-private:
-    EraseableStrokePart(const EraseableStrokePart& part);
 
 public:
     void addPoint(Point p);
     double getWidth() const;
 
-    GList* getPoints();
+    std::vector<Point>& getPoints();
```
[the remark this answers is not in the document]
Marmare314: In EraseableStroke::erasePart points are erased. A `const` version could be added though.

src/model/eraser/EraseableStrokePart.cpp
```diff
@@ -79,22 +55,15 @@ auto EraseableStrokePart::getElementHeight() const -> double { return this->elem
...
 void EraseableStrokePart::clearSplitData() {
-    for (GList* l = this->points->next; l->next != nullptr;) {
-        auto* p = static_cast<Point*>(l->data);
-        delete p;
-        GList* link = l;
-        l = l->next;
-
-        this->points = g_list_delete_link(this->points, link);
-    }
+    points.erase(points.begin() + 1, points.end() - 1);  // deletes everything but endpoints
```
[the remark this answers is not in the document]
Marmare314: Should `assert` from `<cassert>` be used or `g_assert`?

src/undo/PageLayerPosEntry.h
Marmare314: At least `vector::emplace` does not support it yet.
But I think this is cleaner than dealing with it in the files using it.

arv-pyrna:
@Marmare314 : There are some files you did not change for example: 

PopplerGlibPage.cpp,
ToolbarCustomizeDialog.cpp, 
MainWindow.cpp,
RecentManager.cpp
DeviceListHelper.cpp

In these place, is there any specific reason why you avoided it ? or is it not worth changing there ? 
If you don't want to work on them, maybe i will it up.

Marmare314:
> @Marmare314 : There are some files you did not change for example:
> 
> PopplerGlibPage.cpp,
> ToolbarCustomizeDialog.cpp,
> MainWindow.cpp,
> RecentManager.cpp
> DeviceListHelper.cpp
> 
> In these place, is there any specific reason why you avoided it ? or is it not worth changing there ?
> If you don't want to work on them, maybe i will it up.

- PopplerGlibPage: its not worth replacing it as poppler_page_find_text returns a glist an it is only iterated over once
- MainWindow, DeviceListHelper, ToolbarCustomizeDialog: same thing here
- RecentManager: i forgot about this one

It will be a while before I will have time to replace it, so feel free to do it.

Febbe:
With c++20, we will get std::ranges and Sentinels, this will make it simple, to just create a 
`template <GListLike C> GListView;`, without wrapping it. Then we can also get rid of some copies.

rolandlo:
This PR should be rebased on current master, while solving the merge conflicts.

Febbe:
Rebased on current master

Febbe:
Thank you for the PR. I finally got time to merge it.

PR #3050, opened 2021-04-19
fix settings comments placement

mjg:
The comments for `scrollbarHideType` and for new pages are misplaced in
`settings.xml`, for two different reasons: comments arise as nodes
attached to the node which they comment on in the tree but are output
before the node in a textual representation. Therefore, the code
generating the tree can generate a node comment only after the node it
comments on (1st reason). The usage of macros hides this fact, so when
you circumvent the macros you have to assign the node variable manually,
or else the comment will be attached to the wrong node (2nd reason).
Fix both.

### DIFF
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -838,19 +838,18 @@ void Settings::save() {
     WRITE_BOOL_PROP(disableScrollbarFadeout);
 
     if (this->scrollbarHideType == SCROLLBAR_HIDE_BOTH) {
-        saveProperty("scrollbarHideType", "both", root);
+        xmlNode = saveProperty("scrollbarHideType", "both", root);
     } else if (this->scrollbarHideType == SCROLLBAR_HIDE_HORIZONTAL) {
-        saveProperty("scrollbarHideType", "horizontal", root);
+        xmlNode = saveProperty("scrollbarHideType", "horizontal", root);
     } else if (this->scrollbarHideType == SCROLLBAR_HIDE_VERTICAL) {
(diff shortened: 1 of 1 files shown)


src/control/settings/Settings.cpp
```diff
@@ -838,19 +838,19 @@ void Settings::save() {
...
     } else if (this->scrollbarHideType == SCROLLBAR_HIDE_HORIZONTAL) {
-        saveProperty("scrollbarHideType", "horizontal", root);
+        xmlNode = saveProperty("scrollbarHideType", "horizontal", root);
     } else if (this->scrollbarHideType == SCROLLBAR_HIDE_VERTICAL) {
-        saveProperty("scrollbarHideType", "vertical", root);
+        xmlNode = saveProperty("scrollbarHideType", "vertical", root);
     } else {
-        saveProperty("scrollbarHideType", "none", root);
+        xmlNode = saveProperty("scrollbarHideType", "none", root);
     }
```
Technius, line 849: Very minor, but can you delete this space to make it clear that the comment is attached to `scrollbarHideType`?

Technius (approved):
Thanks for the PR. I would be willing to merge this as-is, but I have some minor comments.

* I wonder if it would be better to rename the `WRITE_COMMENT` macro to `ATTACH_COMMENT` macro.
* See minor formatting concern below.

mjg:
Renaming may also help an intermediate goal: abstract the saving away into a class which has those macros as methods, and provide an alternative subclass of that abstract class which stores into and reads from a lua table...

Technius:
LGTM, thanks.

PR #3062, opened 2021-04-22
avoid crash from wrong action group/type spelling

rolandlo:
Fixes a crash reported in https://github.com/xournalpp/xournalpp/issues/919#issuecomment-821897950 resulting from wrong action type or group names, than can be called from a plugin.


rolandlo:

PR #3063, opened 2021-04-22
custom plugin folder implemented

UweStrohbeck:
implemented a user selectable custom plugin folder. The user can choose if he wants to use the standard plugin folder or a selectable custom plugin folder. Changes are made to the UI also.
Tests were made on Ubuntu 18.04 only.
Because PluginController::loadPluginsFrom can no deal with url file path names I used the gtk_file_chooser_get_filename to retrieve the path name (see 
char* pFolderName = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(get("fcPluginCustomFolder"))); 
in settingsdialog.cpp. Uncertain if this will work on Win and Mac also.

### DIFF
--- a/src/plugin/PluginController.cpp
+++ b/src/plugin/PluginController.cpp
@@ -10,12 +10,29 @@
 
 #include "Plugin.h"
 #include "StringUtils.h"
+#include "filesystem.h"
 
 
 PluginController::PluginController(Control* control): control(control) {
 #ifdef ENABLE_PLUGINS
-    auto searchPath = control->getGladeSearchPath()->getFirstSearchPath();
-    loadPluginsFrom((searchPath /= "../plugins").lexically_normal());
+    fs::path searchPath;
+    Settings* settings = control->getSettings();
+    // check if we have to use a customize plugin folder
+    if(settings->isPluginCustomFolderEnabled() == true)
+    {
+      // get customized plugin folder name
+      searchPath = settings->getPluginCustomFolder();
+    }
+    else
+    {
+      // use standard plugin folder
+      // get first subdirectory of the applicaiton main directory    
+      searchPath = control->getGladeSearchPath()->getFirstSearchPath();
+      // select plugin subdirectory
+      searchPath = searchPath /= "../plugins";
+    }
+    // load plugins
+    loadPluginsFrom((searchPath).lexically_normal());
 #endif
 }
 
--- a/Doxyfile
+++ b/Doxyfile
@@ -995,13 +995,13 @@ USE_MDFILE_AS_MAINPAGE =
 # also VERBATIM_HEADERS is set to NO.
 # The default value is: NO.
 
-SOURCE_BROWSER         = NO
+SOURCE_BROWSER         = YES
 
 # Setting the INLINE_SOURCES tag to YES will include the body of functions,
 # classes and enums directly into the documentation.
 # The default value is: NO.
 
-INLINE_SOURCES         = NO
+INLINE_SOURCES         = YES
 
 # Setting the STRIP_CODE_COMMENTS tag to YES will instruct doxygen to hide any
 # special comment blocks from generated source code fragments. Normal C, C++ and
@@ -2331,7 +2331,7 @@ INCLUDED_BY_GRAPH      = YES
 # The default value is: NO.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-CALL_GRAPH             = NO
+CALL_GRAPH             = YES
 
 # If the CALLER_GRAPH tag is set to YES then doxygen will generate a caller
 # dependency graph for every global function or class method.
@@ -2343,7 +2343,7 @@ CALL_GRAPH             = NO
 # The default value is: NO.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-CALLER_GRAPH           = NO
+CALLER_GRAPH           = YES
 
 # If the GRAPHICAL_HIERARCHY tag is set to YES then doxygen will graphical
 # hierarchy of all classes instead of a textual one.
--- a/src/control/settings/Settings.cpp
+++ b/src/control/settings/Settings.cpp
@@ -157,6 +157,7 @@ void Settings::loadDefault() {
     this->audioGain = 1.0;
     this->defaultSeekTime = 5;
 
+    this->pluginCustomFolderEnabled = false;
     this->pluginEnabled = "";
     this->pluginDisabled = "";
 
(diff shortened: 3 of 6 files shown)


src/plugin/PluginController.cpp
```diff
@@ -10,12 +10,29 @@
...
+      searchPath = settings->getPluginCustomFolder();
+    }
+    else
+    {
+      // use standard plugin folder
+      // get first subdirectory of the applicaiton main directory    
+      searchPath = control->getGladeSearchPath()->getFirstSearchPath();
+      // select plugin subdirectory
+      searchPath = searchPath /= "../plugins";
+    }
```
rolandlo, line 33:
```suggestion
    Settings* settings = control->getSettings();
    // check if we have to use a customize plugin folder
    auto searchPath = (settings->isPluginCustomFolderEnabled()) ?
                              fs::path{settings->getPluginCustomFolder()} :
                              control->getGladeSearchPath()->getFirstSearchPath() /= "../plugins";
```

rolandlo (changes requested):
Great to have custom plugin folders! I am looking forward to it.
Some comments:
- I tested it on Ubuntu 21.04, Windows 10 and MacOS Catalina. It worked on all these systems.
- In order that the setting takes effect, restart is required. This should be mentioned in the preferences.
- Try to use the latest glade version to not change the more modern minus-case to snake_case in the ui files.
- Apply clang format. Compare the comments in #3042
- The idea of @Febbe in #2422 was to use the custom plugin folder additionally to the default plugin folder and to take care of duplicated plugins. In my opinion your approach, which is quite a bit simpler, doesn't have any major drawbacks. The user can copy the plugins from the default plugin folder to the user plugin folder if he wishes so.
- Could you please comment on your changes to the Doxyfile?
- I think the changes to `PluginController.cpp` should be simplified (see code suggestion)

UweStrohbeck:
Hi Roland,
thanks for taking your time to review my pull request. 
- I was uncertain about the path desciptor on different platforms. Thanks for testing on different platforms. 
- I can add a info to the gui that the changes take effect only after restart. How to proceed? I make the changes in my local file and pushing this to my fork on github, and then?
- Regarding glade I do not understand what is the point. I installed glade on ubuntu and modified the ui file. I do not understand what you mean with minus-case and snake-case?
- Not sure if I got the point. I have to install clang on my machine and use it before generating a pull request?
- The idea of Febbe has a major drawback on ubuntu. If you install Xournal++ via the Software center (as suggested in the doc) you will end up installing xournal in the snap environment. In this closed environment even with root access you can not add you own plugins into the standard plugin folder. Therefore you have no chance to modify installed plugins in the standard plugin folder.
- I made changes to the doxygen file in order to activate the caller graph and the reference to the sources. As a beginner I found it helpful get better understanding about the sources
- your changes to the PluginController.cpp are highl appreciated

Sorry for bothering you with my 'entry-level' questions.
Br
Uwe

personalizedrefrigerator:
>     * I installed glade on ubuntu and modified the ui file. I do not understand what you mean with minus-case and snake-case?

The version of Glade available from `apt` in Ubuntu 20.04 is older than the version of Glade available from `snap`. Install it using `sudo snap install glade` or from the `Ubuntu Software` app (it should be version 38.x.x or greater!)


>     * Not sure if I got the point. I have to install clang on my machine and use it before generating a pull request?

We're using the program `clang-format` to apply consistent formatting to our code. You should be able to [`sudo apt install clang-format-10` on Ubuntu 20.04](https://packages.ubuntu.com/search?keywords=clang-format-10).

rolandlo:
> 
> > * Not sure if I got the point. I have to install clang on my machine and use it before generating a pull request?
> 
> We're using the program `clang-format` to apply consistent formatting to our code. You should be able to [`sudo apt install clang-format-10` on Ubuntu 20.04](https://packages.ubuntu.com/search?keywords=clang-format-10).
> 
> See https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#use-clang-format
> (@rolando The above link says we're still using clang-format-9, but I don't think that's true anymore).

Yes, that's outdated. We use `clang-format-12`, while `clang-format-11` still seems to be recent enough for what I have observed so far. However `clang-format-10` doesn't always produce the results that satisfies the pipeline check. In the Azure pipeline we install `clang-format-12` from PPA, see https://github.com/xournalpp/xournalpp/pull/2732/files

@UweStrohbeck 
- Yes, you change your files locally and push to your fork. Then the changes should be automatically visible when you load the page on Github next time. There is no need to do anything more.
- You can apply clang format before you create the pull request, but you can also do it later as a separate commit. In the end this fixup commit should be `squashed` with the previous commit in order to keep a clean history, using `git rebase -i HEAD~<number of commits>`. I can explain the squashing later to you. Maybe there will be more to squash.

mjg:
Maybe a bit late, but some thoughts:
- Do we need default plugins at all? By all means, shipping default plugins is fine, but typically they are shipped as example code at a location where they are not run from.
- A typical location for plugins is `${CONFIGFOLDER}/plugins}`. A user would copy an example plugin or their own into that folder to activate it.

At least that's what many other programmes do. That way we need neither a preference for the folder nor a way of diabling default plugins (nor a way of disabling or enabling any).

UweStrohbeck:
Hi Michael,

> A typical location for plugins is ${CONFIGFOLDER}/plugins}. A user would copy an example plugin or their own into that folder to activate it.

In my point of view this will not work when you installed the SW via Ubuntu snap. As far as I understood snap correctly it 'simulates' a  file system and let you not modify anything within this file system by yourself. Even not as a root.

Br
Uwe

mjg:
> Hi Michael,
> 
> > A typical location for plugins is ${CONFIGFOLDER}/plugins}. A user would copy an example plugin or their own into that folder to activate it.
> 
> In my point of view this will not work when you installed the SW via Ubuntu snap. As far as I understood snap correctly it 'simulates' a file system and let you not modify anything within this file system by yourself. Even not as a root.
> 
> Br
> Uwe

Well, that is true for the "system part", and you and I both want to avoid writing to that part. But snap is not a hypervisor or full sandbox mechanism.

The config folder is per user in their home directory - I do hope that snap apps can access files in your home directory, or they would be useless.

UweStrohbeck:
Thanks for the interessting discussion.
Finally a decision has to be made. The current solution can be seen as a bug, Because plugin development within ubuntu/snap is currently not possible.

Option A
Us a predefined plugin folder within the 'user' directory. Not sure, if we have to differ between the different OS (getting user names, installation for all users or only selected users on a machine, ...). The user cannot change the plugin directory.

Option B
use a custom plugin folder in addition to the standard plugin folder. Still the delivered plugin examples can not be modified for ubuntu users, Copying the examples to a custom folder may end up in duplicates.

Option C
Keep the standard behaviour as it is - may be appropriate for alot of users. If the user decides to change the plugin folder, give him maximum degree of flexibility

br
Uwe

rolandlo:
> Thanks for the interessting discussion.
> Finally a decision has to be made. The current solution can be seen as a bug, Because plugin development within ubuntu/snap is currently not possible.
> 
> Option A
> Us a predefined plugin folder within the 'user' directory. Not sure, if we have to differ between the different OS (getting user names, installation for all users or only selected users on a machine, ...). The user cannot change the plugin directory.
> 
> Option B
> use a custom plugin folder in addition to the standard plugin folder. Still the delivered plugin examples can not be modified for ubuntu users, Copying the examples to a custom folder may end up in duplicates.
> 
> Option C
> Keep the standard behaviour as it is - may be appropriate for alot of users. If the user decides to change the plugin folder, give him maximum degree of flexibility
> 
> br
> Uwe

I like option C (i.e. the one suggested by this PR currently) best. It's simple, flexible and well in line with other things we have in Xournal++ like setting the audio directory or the LaTeX template file. Compared to option A (suggested by @mjg) it has the advantage, that when you install Xournal++, default plugins are available right away and you don't start with an empty plugin folder and have to figure out if plugins really work on your machine. 
Moreover the current way of enabling/disabling plugins is more convenient than copying/removing plugins in the directory tree. Maybe we will manage to do without restarting Xournal++ in future, which would make it even more convenient. 

I would like to hear the opinion of @Febbe and @Technius though before we proceed.

Also let me remind that the current behavior is not only a problem for Ubuntu/Snap users. It's also a problem on other systems where the user doesn't have admin privileges.

Technius:
@Febbe and I discussed this previously, see https://github.com/xournalpp/xournalpp/issues/1155#issuecomment-623234420

Febbe:
Additionally there is also a pending PR to implement this, which is stalled, until we release version 1.1

rolandlo:
Closing due to lack of activity and in view of #2422

PR #3076, opened 2021-04-26
Fix capitalization in menus

kenballus:
This patch consists of 9 capitalization fixes scattered around various menus. I fixed them using Glade, so it also switched around the order of a few properties in the XML. I think that's just due to a Glade update, and will have no effect on the UI.


rolandlo:
LGTM. Translations will have to be regenerated though, which may take a while. So I'm not 100% sure we should merge it before the 1.1.0 release.

rolandlo:
@Technius Should we merge it?

Technius (approved):

PR #3088, opened 2021-05-02
Fix Windows CI

Technius:
An attempt to fix the Windows builds. The recent errors are due to a package version conflict with packages such as GCC, which I suspect are happening because we were not syncing the package repositories.

### DIFF
--- a/azure-pipelines/steps/build_windows.yml
+++ b/azure-pipelines/steps/build_windows.yml
@@ -8,7 +8,7 @@ steps:
   # Copy of msys2-blob as a temporary fix for Windows CI
   - script: |
       set PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem
-      pacman --noconfirm -Sy
+      pacman --noconfirm -Syu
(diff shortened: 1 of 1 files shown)


rolandlo:
As observed in https://github.com/xournalpp/xournalpp/pull/2871#issue-579579613 this may increase build times by <= 15 minutes. But Ok, what else can we do? Have you checked the PR in your own pipeline?

Technius:
This doesn't seem to work, so the issue is elsewhere. Maybe it's an Azure Pipelines issue. Here's the full error for context:
```
resolving dependencies...
error: failed to prepare transaction (could not satisfy dependencies)
looking for conflicting packages...
:: installing mingw-w64-x86_64-gcc-libs (10.2.0-11) breaks dependency 'mingw-w64-x86_64-gcc-libs=10.2.0-10' required by mingw-w64-x86_64-gcc
:: installing mingw-w64-x86_64-gcc-libs (10.2.0-11) breaks dependency 'mingw-w64-x86_64-gcc-libs=10.2.0-10' required by mingw-w64-x86_64-gcc-libgfortran
```

Technius:
According to [this](https://github.com/msys2/setup-msys2#context), the msys setup is updated on GitHub virtual environments (which Azure Pipelines uses) about every 10 days. So one solution would be to just wait.

PR #3089, opened 2021-05-02
Fix Windows Azure Pipeline

rolandlo:
Removing the updating of the package index like in this PR works, as my test on my personal dev.azure pipeline shows, and it does now uses Cairo 1.17.4-2. This is needed to have good pdf exports. So I suggest to merge this PR. From [the pipeline](https://dev.azure.com/rolandlo/xournalpp/_build/results?buildId=51&view=logs&j=5502ea34-1a31-5d2e-665e-0754f4413bc7&t=4d4fe241-2267-54df-e534-736dc4ac6f29):
```
downloading mingw-w64-x86_64-cairo-1.17.4-2-any.pkg.tar.zst...
```

### DIFF
--- a/azure-pipelines/steps/build_windows.yml
+++ b/azure-pipelines/steps/build_windows.yml
@@ -8,7 +8,6 @@ steps:
   # Copy of msys2-blob as a temporary fix for Windows CI
   - script: |
       set PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem
-      pacman --noconfirm -Sy
       pacman --noconfirm -S mingw-w64-x86_64-imagemagick mingw-w64-x86_64-ninja mingw-w64-x86_64-cppunit mingw-w64-x86_64-poppler mingw-w64-x86_64-gtk3 mingw-w64-x86_64-libsndfile mingw-w64-x86_64-libzip mingw-w64-x86_64-lua
     env:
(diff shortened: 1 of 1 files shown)


rolandlo:
Replaces #3088

Technius:
Yeah, this makes more sense to me. Since the msys environment is updated every 10 days, we don't want to risk a partial upgrade by updating the package index. In the future, it would probably just be easier to wait those 10 days before we rely on a newer dependency. If an update causes a breakage, we could pin an older version.

Mac builds seem to be broken now for reasons unrelated to this PR:
```
tar: Error opening archive: Unrecognized archive format
```
so something might be going wrong with the Mac binary blob.

rolandlo:
> Yeah, this makes more sense to me. Since the msys environment is updated every 10 days, we don't want to risk a partial upgrade by updating the package index. In the future, it would probably just be easier to wait those 10 days before we rely on a newer dependency. If an update causes a breakage, we could pin an older version.
> 
> Mac builds seem to be broken now for reasons unrelated to this PR:
> 
> ```
> tar: Error opening archive: Unrecognized archive format
> ```
> 
> so something might be going wrong with the Mac binary blob.

We've had this previously at the beginning of a month, probably related to bandwidth which is reset at the beginning of the month. It was fixed automatically after a couple of days.

Technius:
Oh, I remember seeing that. We should probably document these common breakages so that maintenance is easier in the future.

PR #3094, opened 2021-05-04
Streamline package descriptions and update appdata xml

rolandlo:
We have four variants of a package description in Xournal++
1. the one from the `package_description` file
2. the one in the `README.md`
3. the one in `debian/control`
4. the one in the `com.github.xournalpp.xournalpp.appdata.xml` file (that is used for the `flatpak`). 

I streamlined them and made them all equal up to formatting details. I have also updated the appdata xml-file to some degree.

The appdata xml-file refers to four images, which are already seven years old, i.e. pretty outdated. We should create new images, that show how Xournal++ looks now and what we can do with it now.  Has anybody a nice .xopp-file that could be used to create such an image?

I think we need at least:
- an image for notetaking, maybe with multiple layers
- an image for PDF-annotation
- an image for using LaTeX formulas, maybe something showing off the power we gained with the customizable template, like code-highlighting or drawing with `TikZ` or function graphs via `TikZ/pgfplots`
- maybe something with customizable toolbars or floating toolbox (although that is not new)

Any suggestions?

Three of the 7-year old images that are currently referred to are located on [imgur.com](https://imgur.com/). Is that a good place for these images or should be include the new images in the repository or somewhere else? Note that these images will be shown to tons of Linux users when looking for Xournal++ in the software center or on the flathub website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,26 +40,27 @@ Stroke recognizer and other parts are based on Xournal Code, which you can find 
 
 Xournal++ features:
 
-- Support for pen pressure, e.g. Wacom Tablet
-- Support for annotating PDFs
+- Supports pressure senstive styluses and digital pen tables (e.g. Wacom, Huion, XP Pen, etc. tablets)
+- Paper backgrounds for notetaking, scratch paper, or whiteboarding
+- Annotate on top of PDFs
+- Export to a variety of formats including SVG, PNG and PDF, both from the GUI and command line
+- Different drawing tools (e.g.pen, highlighter) and stroke styles (e.g. solid, dotted)
+- Shape drawing (line, arrow, circle, rectangle, spline)
 - Fill shape functionality
-- PDF Export (with and without paper style)
-- PNG Export (with and without transparent background)
-- Allows mapping different tools / colors etc. to stylus buttons / mouse buttons
-- Sidebar with Page Previews with advanced page sorting, PDF Bookmarks and Layers (can be individually hidden, editing layer can be selected)
+- Shape resizing and rotation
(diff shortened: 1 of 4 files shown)


rolandlo:
I have created 3 xopp-files and taken screenshots, that could be uploaded and linked to in the `com.github.xournalpp.xournalpp.appdata.xml`

**For note-taking:**
note-taking

**For pdf annotation:**
pdf_annotation

**For LaTeX and customized preamble:**
LaTeX_with_customized_preamble

Any oppinion @Technius @Febbe @idotobi ?

Technius:
Looks awesome! If you want to show them on Flatpak store or whatever, you can just directly use the link generated by GitHub since GitHub will need to store the images for your comment. We'll need somewhere else to store the xopp files though.

rolandlo:
> Looks awesome! If you want to show them on Flatpak store or whatever, you can just directly use the link generated by GitHub since GitHub will need to store the images for your comment. We'll need somewhere else to store the xopp files though.

Thanks! The pdf background and the LaTeX preamble used will have to be stored along with the xopp-files, I think. These should probably be stored within the repository. I don't see a folder where it belongs, but I could create a new one.

Technius:
I would put the resources in another repository instead (e.g. `xournalpp-examples` or something), unless we're also going to use them for unit tests.

rolandlo:
> I would put the resources in another repository instead (e.g. `xournalpp-examples` or something), unless we're also going to use them for unit tests. Besides that, everything looks good to me.

Done! I have subtracted the resources from this PR and added them to a `xournalpp-examples`repository.

PR #2837, opened 2021-02-20
Fix zoom jumping and zoom steps (Partial Zoom Refactor)

redweasel:
This is a splitup part of https://github.com/xournalpp/xournalpp/pull/2743 .
see https://github.com/xournalpp/xournalpp/pull/2743 for older comments.

This patch fixes the issues with zoom jumping to previous/different pages. It also changes zoom steps to be relative for a much better zooming experience.

Internally this is a change from local to global coordinates, because the local coordinates had weird position spikes in them. This is honestly mostly due to my missing understanding of where the local coordinates come from. I am sure that part of this patch can be done differently.

I have also simplified some of the logic to avoid addition/subtraction of large floating point values and simpler access of the zoom calls.

PR #3085, opened 2021-04-29
Add CONTRIBUTING.md

Technius:
Fixes #3065. This is an initial draft, looking to get some feedback first.

PR #3099, opened 2021-05-08
Temporary hotfix for zoom fit loop

### LINKED ISSUE #3036: Crash when scrolling over slides with different format
Affects versions :

    OS: Linux 20.04
    (Linux only) Desktop environment: Gnome 3.36, X11
    Which version of libgtk do you use: 3.24.20
    Version of Xournal++: 1.1.0+dev
    Installation method: sudo apt 

**Describe the bug**
I work on Powerpoint Slides. When I insert a new blank page this has another Format than the Powerpoint slides.
If I let Xournal maximize the slides and scroll up and down through the different size slides it crashes. And it really crashes, it unfortunately does not save the current state (as it normally does :-) ), and when you restart you can easily recover it, as a pop up just gives you the option. It really crashes and then you can only search for the recovery file yourself, which is often missing a lot, since it was autosaved a minute before it crashed.

**To Reproduce**
1. Open the file I uploaded
[Crash-Test-Slides.zip](https://github.com/xournalpp/xournalpp/files/6309858/Crash-Test-Slides.zip)

2. Make sure 'Zoom fit to screen' is activated
3. Scroll up and down through the empty slides with different formats
4. For me it crashes every time

**Expected behavior**
Normal Scrolling from slide to slide. Changing automaically the Zoom fit to screen.

**Additional context**
Is there an option to make Xournal insert new empty pages with the same size as the page before?
(At the moment this can only be done with first duplicating page and then apply 'Plain' to current page. This is more of a feature request, but I think it would be in general very nice if it just inserted pages of the same size as a standard.

Technius:
Fixes #3036. This is a **temporary** fix for the infinite recursing zoom fit issue so that users of 1.1.0 won't get crashes. A quick run with the test file provided in #3036 suggests that the infinite recursing issue is caused, in part, by jumpy/nonlinear scrolling during the zoom (which should be improved by #2837).

### DIFF
--- a/src/control/zoom/ZoomControl.cpp
+++ b/src/control/zoom/ZoomControl.cpp
@@ -303,11 +303,18 @@ void ZoomControl::zoom100() {
 }
 
 void ZoomControl::zoomFit() {
+    // TODO: properly fix the zoom fit infinite recursion
+    if (this->isZoomFittingNow) {
+        return;
+    }
+
+    this->isZoomFittingNow = true;
     if (this->isZoomFitMode() && !this->zoomPresentationMode && this->zoom != this->zoomFitValue) {
         startZoomSequence();
         this->zoomSequenceChange(this->zoomFitValue, false);
         endZoomSequence();
     }
+    this->isZoomFittingNow = false;
 }
 
 void ZoomControl::zoomPresentation() {
--- a/src/control/zoom/ZoomControl.h
+++ b/src/control/zoom/ZoomControl.h
@@ -239,4 +239,5 @@ private:
 
     size_t current_page = static_cast<size_t>(-1);
     size_t last_page = static_cast<size_t>(-1);
+    bool isZoomFittingNow = false;
 };


rolandlo (approved):
This works fine with all documents I have tested and makes sense. Only minor nitpick: We have some coding conventions about TODOs that are not followed
https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#todo-comments
(regarding the name)

Technius:
> Only minor nitpick: We have some coding conventions about TODOs that are not followed

I do not find that convention to be very useful -- it adds noise without much information. We should discuss this more in #3085.

rolandlo:
> > Only minor nitpick: We have some coding conventions about TODOs that are not followed
> 
> I do not find that convention to be very useful -- it adds noise without much information. We should discuss this more in #3085.

Ok, we will need to overthink these conventions in #3085 then.

rolandlo:
@Febbe Will you look into this PR or shall we merge it?

rolandlo:

PR #3113, opened 2021-05-13
Updated my copyiright year :-)

andreasb242:


mjg:
I'm always wondering why people are keen on these date ranges. Technically, if you use a range then the end date is the date of the last revision, not the end of copyright validity. You had no change in 2021, but again that does not make your copyright end in 2020 (though the statement you make about 2021 could be questioned). The change to about.glade "updating" the copyright would not qualify as a "revision" if it mattered. Is there any jurisdiction where it would?

Issue #1594, opened 2019-11-07
Set default binding of middle mouse button to hand tool


JulianGro:
**Is your feature request related to a problem? Please describe.**
When having a tool like the text tool selected, you cannot move the page with the middle mouse button. This is pretty annoying when you are zoomed in and want to make a new text box in a different place.

**Describe the solution you'd like**
If the middle mouse button is not used for a function of a tool, it should be used for moving the view.

**Additional context**
In this screenshot I was typing on the right side of the table and wanted to read text on the left of the table. To move the view I had to select the hand tool, rather than just using the middle mouse button.
![Bildschirmfoto vom 2019-11-04 12-29-17](https://user-images.githubusercontent.com/11144627/68383179-754c3d00-0155-11ea-98e6-74e00a95bd85.png)

Technius:
You can set the middle mouse button to the Hand Tool in the settings:
![image](https://user-images.githubusercontent.com/1066652/68640957-23414980-04be-11ea-9ca7-9a501ff616db.png)

If this doesn't work for you, we'd appreciate more feedback.

jug007:
It does work the main machine I use xournalpp on. I do feel like this should be set by standard.
Also the hand doesn't work on my main desktop machineat all. Probably because of #1501

Technius:
Since the middle mouse button is not set by default, letting the default be the hand tool seems like a reasonable change.

Technius:
Fixed by #3121.

PR #3121, opened 2021-05-18
Set default binding of middle mouse button to hand tool

BHatGuy:
Set the default binding of the middle mouse button to 'hand tool' as proposed in #1594


Technius:
LGTM, thanks for the PR.

PR #3131, opened 2021-05-27
fix text selection bug (#3071)

### LINKED ISSUE #3071: Text Tool: Selection/Highlighting does not always match actual selected text.
**Affects versions :**
 - OS: Linux/Manjaro
 - Which version of libgtk do you use: 3.24.28
 - Version of Xournal++: 1.1.0+dev (built from fork)
 - Installation method: Build from source

**Describe the bug**
Highlighted regions in selected regions of text do not always match the actual selected region.

**To Reproduce**
1. Create a blank document.
2. Add the text: "This is a test..." with the text tool.
3. Select (using Shift + Left_Arrow * 9) "a test..."
4. Unselect "a test..." (using Shift + Right_Arrow * 2)
5. Press backspace.

**Expected behavior**
The "a " should not be highlighted after unselecting it.

**Screenshots of Problem**
See bug description

VivekThazhathattil:
This commit fixes #3071. 
Before the change, using `pango_layout_get_attributes` meant that the previously selected text's changed background color persisted even when it was unselected. Now its working as expected.

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -61,13 +61,9 @@ TextEditor::~TextEditor() {
 
     if (this->ownText) {
         UndoRedoHandler* handler = gui->getXournal()->getControl()->getUndoRedoHandler();
-        for (TextUndoAction& undo: this->undoActions) {
-            handler->removeUndoAction(&undo);
-        }
+        for (TextUndoAction& undo: this->undoActions) { handler->removeUndoAction(&undo); }
     } else {
-        for (TextUndoAction& undo: this->undoActions) {
-            undo.textEditFinished();
-        }
+        for (TextUndoAction& undo: this->undoActions) { undo.textEditFinished(); }
     }
     this->undoActions.clear();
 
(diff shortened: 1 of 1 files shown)


VivekThazhathattil:
Bug:
text_selection_bug

Fixed:
text_selection_fixed

rolandlo (approved):
LGTM. I checked that it works.

Technius:

Technius:
Still need to know:
> Is this a regression or a bug that also exists in 1.0.x?

Febbe:
I have one question for you @VivekThazhathattil, you replaced the current setting, instead of merging them. What implications does this have on current settings? Or is this irrelevant, since this is the only setting ever written, and it is safe to just override it?

Technius:
> I have one question for you @VivekThazhathattil, you replaced the current setting, instead of merging them. What implications does this have on current settings? Or is this irrelevant, since this is the only setting ever written, and it is safe to just override it?

On second thought, this may introduce potential problems with pre-edit strings with some languages. I will check if this PR introduces bugs with some inputs methods.

Technius (approved):
I had time to look and test this carefully. Works with ibus-pinyin input method with no obvious errors.

> you replaced the current setting, instead of merging them. What implications does this have on current settings? Or is this irrelevant, since this is the only setting ever written, and it is safe to just override it?

It seems like the attributes are only ever changed in `paint`, so this should be safe. However, if we want to set attributes elsewhere in the future, we will need to adjust the code so that it copies or modifes the old attribute list.

Technius:

PR #3156, opened 2021-06-15
Removed references of macOs High Sierra

Febbe:


rolandlo (approved):
We will have to update the screenshots as well (using the new icons), but that can be done in another PR.

PR #3159, opened 2021-06-15
 Fixes #1916

### LINKED ISSUE #1916: Corrupted file on closing without warning during pdf export
**Affects versions :**
 - OS: Linux Debian 9
 - (Linux only) Desktop environment: KDE
 - Which version of libgtk do you use: 3.22.11-1 (I think)
 - Version of Xournal++: 1.0.16
 - Installation method: snap

**Describe the bug**
When editing a pdf (after opening it in xournal via the normal "open" dialogue), and then using "Export as PDF" to export the pdf with annotations, one can close xournalpp _during_ the export without any warning. The pdf file is then corrupted.

**To Reproduce**
See description.

**Expected behavior**
Either refuse to close while exporting pdfs, or warn the user that a pdf is currently exported and if he/she wants to stop the export and close.

**Additional context**
This is very serious as people can lose data by this. I almost lost three hours of work, but fortunately I could recover thanks to the autosave feature.

Febbe:
- Reordered some calls, to prevent segfaults on export.   
 - Added explicit wait.
 - Fixed some minor casting issues

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2451,15 +2451,13 @@ void Control::quit(bool allowCancel) {
         return;
     }
 
-    this->closeDocument();
-
-    this->scheduler->lock();
-
     audioController->stopRecording();
-    settings->save();
-
+    this->scheduler->lock();
     this->scheduler->removeAllJobs();
     this->scheduler->unlock();
+    this->scheduler->stop();  // Finish current task. Must be called to finish pending saves.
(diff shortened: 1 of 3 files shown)


rolandlo (approved):
LGTM. With regards to #1916 I checked that it works: Xournal++ waits with closing the document until the PDF export is finished.

Technius:

PR #3162, opened 2021-06-16
Port thickness-thicker to light theme

### LINKED ISSUE #3161: How use option stock icons ?
There is no change when I check in preferences "use stock icons" (windows 10)

Fatih20:
To fix #3161.


rolandlo:
Thanks, @Fatih20, I will merge it right away.

PR #3170, opened 2021-06-19
Quick fix for making cloned background images load

### LINKED ISSUE #3165: Images used as background are not saved.

**Affects versions :**
 - OS: Debian Buster
 - Which version of libgtk do you use: 3.24.5
 - Version of Xournal++: 1.1.0+dev (nightly 20210618)
 - Installation method: Automated nightly build

**Describe the bug**
I use a PNG file as a background for some pages. After applying the background image, saving the document and reopening it, only the first page has the background image. All following pages have a gray "PDF background missing" instead.

**To Reproduce**
1) Go to Journal -> Paper background -> Image.
2) Select the PNG image and then select Journal -> Paper background -> Apply to all pages.
3) Confirm the background page individually for each page by clicking the OK button (50 times in my case) to apply the background to all pages 
4) save the document and close xournalpp
5) reopen the file in xournalpp

**Expected behavior**
The document should look like as it has been saved including the background images.
Before saving
After loading the same document

**Screenshots of Problem**
(Optional) If applicable, add screenshots to help explain your problem.

**Additional context**

rolandlo:
Fixes #3165 

~I have found two problems, one must have been there for a long time, the other is a regression only in 1.1.0+dev. They are both related to this line:~

https://github.com/xournalpp/xournalpp/blob/28faadd6352aa77b5596fb3e698a5ce8a79693aa/src/control/xojfile/LoadHandler.cpp#L381

~The regression comes from the fact, that `filepath` is of type `fs::path` and not the filename string (which, for a cloned background image, is a string representation of a number).~

The ~other~ problem was that the pages are added too late to the document. So `getPage()` will return a `nullpointer`.

The pages are only added to the document once the whole `xopp`-file has been parsed. 
https://github.com/xournalpp/xournalpp/blob/28faadd6352aa77b5596fb3e698a5ce8a79693aa/src/control/xojfile/LoadHandler.cpp#L231-L233

My PR fixes this by using the `pages` variable instead of `getPage()`. It's a quick ~and dirty~ fix and may have to be improved, but it seems to work.

### DIFF
--- a/src/control/xojfile/LoadHandler.cpp
+++ b/src/control/xojfile/LoadHandler.cpp
@@ -332,6 +332,8 @@ void LoadHandler::parseBgSolid() {
 void LoadHandler::parseBgPixmap() {
     const char* domain = LoadHandlerHelper::getAttrib("domain", false, this);
     const fs::path filepath(LoadHandlerHelper::getAttrib("filename", false, this));
+    // in case of a cloned background image, filename is a string representation of the page number from which the image
+    // is cloned
 
     if (!strcmp(domain, "absolute") || (!strcmp(domain, "attach") && this->isGzFile)) {
         fs::path fileToLoad;
@@ -378,7 +380,13 @@ void LoadHandler::parseBgPixmap() {
 
         this->page->setBackgroundImage(img);
     } else if (!strcmp(domain, "clone")) {
-        PageRef p = doc.getPage(stoull(filepath.string()));
+        gchar* endptr = nullptr;
+        auto const& filename = filepath.string();
+        size_t nr = static_cast<size_t>(g_ascii_strtoull(filename.c_str(), &endptr, 10));
+        if (endptr == filename.c_str()) {
+            error("%s", FC(_F("Could not read page number for cloned background image: {1}.") % filepath.string()));
+        }
+        PageRef p = pages[nr];
 
         if (p) {
             this->page->setBackgroundImage(p->getBackgroundImage());


Technius:
> which, for a cloned background image, is a string representation of a number

Can we add a comment that makes this assumption explicit?

rolandlo:
> > which, for a cloned background image, is a string representation of a number
> 
> Can we add a comment that makes this assumption explicit?

Sure, I have amended it to the previous commit.

src/control/xojfile/LoadHandler.cpp
```diff
@@ -378,7 +381,8 @@ void LoadHandler::parseBgPixmap() {
 
         this->page->setBackgroundImage(img);
     } else if (!strcmp(domain, "clone")) {
-        PageRef p = doc.getPage(stoull(filepath.string()));
+        size_t nr = static_cast<size_t>(stoull(filename));
```
Technius, line 384: If `filename` is an invalid number (which is possible if the file is ill-formed), then `stoull` will return 0. I think [`g_ascii_strtoull`](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-ascii-strtoull) would be a better option because it lets you check for invalid values.

Technius (commented):
Maybe I'm missing something, but I don't understand how this fixes the bug. Why does the old behavior--converting to a path and then converting back to a string--not work?

I also have a concern about the existing integer parsing behavior (see below).

src/control/xojfile/LoadHandler.cpp
rolandlo: Turns out you are right about the conversion to path and converting back to string. I changed it back.

I have now introduced checking for invalid values using `g_ascii_strtoull`

src/control/xojfile/LoadHandler.cpp
```diff
@@ -378,7 +380,13 @@ void LoadHandler::parseBgPixmap() {
 
         this->page->setBackgroundImage(img);
     } else if (!strcmp(domain, "clone")) {
-        PageRef p = doc.getPage(stoull(filepath.string()));
+        gchar* endptr = nullptr;
+        const gchar* filename = filepath.string().c_str();
+        size_t nr = static_cast<size_t>(g_ascii_strtoull(filename, &endptr, 10));
+        if (endptr == filename) {
```
Febbe, line 386: This is dangling, filename doesn't exit anymore, after thar line. Use: 
```suggestion
```suggestion
        auto const& filename = filepath.string();
        size_t nr = static_cast<size_t>(g_ascii_strtoull(filename.c_str(), &endptr, 10));
        if (endptr == filename.c_str()) {
```
        const auto filename = filepath.string();
```

rolandlo: Thanks!

rolandlo:
Merging in 24 hours if no more objections are raised.

Technius (approved):

PR #3172, opened 2021-06-20
Update Layout.cpp

The1234Team:


The1234Team:
THNX !

Technius (changes requested):
I appreciate your enthusiasm, but it would be better to implement options in the settings rather than changing the hardcoded defaults for everyone.

Issue #3039, opened 2021-04-15
StylusHandler Design


redweasel:
This design document will discuss a redesign of the StylusHandler and the relevant options in the preferences. This came up in https://github.com/xournalpp/xournalpp/issues/2717 as the current design is kind of weird, especially for new users. It will need another design task for better touch handling in time with the view refactor.

**If there is anything in here which you don't agree with, please comment below. This is a design document for that exact purpose.**

## Input Events (and their reliability)
- Every stylus has pressure sensitivity. Otherwise it is a pointing device (-> should be Mouse (for drawing) or Touch (for dragging) not Stylus)
  - Remove that option and make stylus with mouse option use the correct handler (it is, but somehow pressure sensitivity still works, which should be considered a bug)
- Pressed event when the pen touches the surface (`TabletPCButton on`) (do not rely on position data for drawing strokes)
- Pressed event when a pen button is pressed (even in hover) (`TabletPCButton off`) (do not rely on position data for drawing strokes)
- Move event with (most of the time) available pressure data and reliable (positive) position data, but sometimes way too dense/frequent for normal strokes.
- Release event when the pen is released from the surface (this can sometimes be left out)
- Enter/Leave event for entering and leaving the drawing area (do not rely on the coordinates or pressure data here)
- Proximity in/out for bringing the pen in proximity of the device and leaving proximity. (do not rely on coordinates data here, also do not rely on every driver sending this, but use it if there)

## Default Behavior
- The stylus will only draw (pressure sensitive tools) when it is pressed down on the surface (positive pressure data is received for drawing, unconnected to the pressed event).
- The stylus will act with tools (non pressure sensitive like "lasso" or "draw rectangle") if the stylus button is pressed down on the surface (positive pressure, not in hover, even if `TabletPCButton off` is configured).

## Additional options
- for each stylus button (1, 2, eraser) there should be a separate option to enable hover actions. Meaning that if enabled, you should be able to draw with a highlighter or select with a lasso in hover. Note in the tooltip that `TabletPCButton off` is needed for this to work (I assume... I think for the eraser buttons it's not strictly needed). If you get too far away it will `abort` the action. If you release the button it will `confirm` the action. `abort` should confirm strokes, but cancel selections.
- "Enable Pressure Inference" should be an option in the stylus tab or be enabled by default as a fallback since it does not affect normal stylus input (again: every stylus device is expected to have pressure sensitivity).

## Removed Preferences
[...]

Issue #2717, opened 2021-01-28
Holding second pen button (switching momentarily to hand tool) does not allow moving around page


michaelsjackson:
Linux, Ubuntu 18.04, experienced this in older versions as well, in certain cases.  This bug seems to have an older age. Latest night build.

michaelsjackson:
Just found out, if I touch with the pen the graphics tablet then it is working, but normally, also in original Xournal (without ++) this is working also when the pen is in the air, meaning without touching the tablet.  But at least there is a solution for now, better than nothing.  Thanks friends.

Mittagskogel:
I have the same issue. In the stable 1.0.20-hotfix release (Commit adaa97ff), this still works correctly: Pressing the pen button while hovering is enough to execute the corresponding action. I use this with the select region and eraser tools. I just compiled Xournal++ (newest commit) for my Raspberry Pi, and now I need to touch the pen to the tablet in combination with pressing the corresponding button in order to use the function (Select region, Eraser, and Hand tool). Perhaps this problem appeared with the new input system. I use the old input system in the 1.0.20-hotfix release.

Mittagskogel:
As a note, I can't use the older 1.0.20-hotfix release on my raspi because of the disappearing page bug (https://github.com/xournalpp/xournalpp/issues/2875)

rolandlo:
You should enable the `Merge button events with stylus button settings` checkbox in menu `Edit -> Preferences` (input system panel)

Mittagskogel:
Awesome, that works, thank you! Apparently I had not tried that in the newest version, even though I thought I had.
As a side note: `xsetwacom get 6 TabletPCButton` returns `off` for me, which is probably why I didn't enable the checkbox in the first place.

Mittagskogel:
Sorry, I did find more issues with this. In commit e5cdd5f2, everything works correctly; I can hover select and hover erase. In the latest commit, hover erase mostly works (although it seems like I need to hold the pen *extra* close to the tablet). Hover select does not work.
I'm happy using the older build, but let me know if there is anything I can do to help!

rolandlo:
@Mittagskogel 
In case you manage to build from source (as described in [the wiki](https://github.com/xournalpp/xournalpp/blob/master/readme/LinuxBuild.md)) you could check if commit ea92dc7340c76237213b0e0fdc521647a5bfaf2f has introduced a regression with respect to the `Merge button events with stylus button settings` mode.

[...]

PR #3178, opened 2021-06-22
Revert "handle buggy driver events better to avoid vanishing lines"

rolandlo:
This reverts commit e442d57eb2312a91eeb4849f27c66e7b25eb1894.

In my opinion the cure was worse than the disease. Some users couldn't write any more, because strokes were ended suddenly. On Windows there were problems with stylus buttons as well. So I think it is better to revert this commit and wait for the more careful approach of #3039 to solve the issues addressed in that commit.

### DIFF
--- a/src/gui/inputdevices/StylusInputHandler.cpp
+++ b/src/gui/inputdevices/StylusInputHandler.cpp
@@ -27,12 +27,6 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
     if (event.type == BUTTON_PRESS_EVENT) {
 
         if (event.button == 1 || this->inputContext->getSettings()->getInputSystemTPCButtonEnabled()) {
-            if (this->eventsToIgnore < 0 && this->inputRunning) {
-                // end last action if it was not ended already
-                this->actionEnd(event);
-                g_message(
-                        "Impossible press event without release event - this is a sign of bugged hardware or drivers");
-            }
             this->eventsToIgnore = this->inputContext->getSettings()->getIgnoredStylusEvents();
             if (this->eventsToIgnore > 0) {
                 this->eventsToIgnore--;  // This is already the first ignored event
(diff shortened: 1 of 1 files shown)


rolandlo:
@Febbe @Technius Any opinion on whether to revert or not to revert?

By the way the Clang format error looks like a temporary thing. The file is formatted correctly.

Febbe:
Is it easy to fix (quick fix is ok) -> fix it.
Does it require a big refactoring: revert

rolandlo:
> Is it easy to fix (quick fix is ok) -> fix it.
> Does it require a big refactoring: revert

I don't see a quick fix that does anything else than reverting more or less exactly what is the content of the commit. More analysis will be required to find a good solution which works well for everybody including the author of the commit @redweasel That will be part of #3039

Technius:
I'm not an expert regarding the input system, so I'll defer judgment to @rolandlo. However, I would prefer the option where the input is decent for most users (with some users experiencing quirks/weird bugs) over an option where the input is great for some users but terrible/non-functional for everyone else.

> Some users couldn't write any more, because strokes were ended suddenly. On Windows there were problems with stylus buttons as well.

Can you point to concrete issues/comments?

rolandlo:
> > Some users couldn't write any more, because strokes were ended suddenly. On Windows there were problems with stylus buttons as well.
> 
> Can you point to concrete issues/comments?

Pro reverting the commit:
- https://gitter.im/xournalpp/xournalpp?at=60d0dbdabcb3e8060ed85572 I had a live meeting with this user and it was really impossible to write without reverting the commit
- https://github.com/xournalpp/xournalpp/issues/2717#issuecomment-815653866 (Hover select did not work without reverting the commit).
- personal experience I had from using stylus buttons on MS Windows (where only dots appeared, having assigned the highlighter to a stylus button). That doesn't seem to make a difference with the patched GTK anymore (which is work in progress).

Contra reverting the commit:
- https://github.com/xournalpp/xournalpp/pull/2984#issue-602674880 (experience of user @redweasel, who made the PR to fix the problems with his bugged hardware)

PR #3201, opened 2021-07-01
fix #3200

### LINKED ISSUE #3200: "Previously selected language not available anymore" all the time with Arabic after simplifying the names
**Affects versions:**
 - OS: Linux, Ubuntu 20.04.
 - Desktop environment: both KDE5 and Unity (Gnome3).
 - libgtk: 3.24.20.
 - Version of Xournal++: 1.1.0+dev (the latest commit "New Crowdin updates and simplification of names" of 2021.06.28).
 - Installation method: both the official PPA, and by manually building from the repo.

**Describe the bug:**

After the recent simplification of language names, when the Language is set to Arabic, this happens:

Whenever I open Preferences, it complains that "Previously selected language not available anymore!", and if I don't select Arabic again, it is reset to "System Default" upon restarting.

But if I use Xournal++ without ever opening Prefrences, no issues occur at all (and Arabic remains after restarting).

Side note: "System Default" uses the English strings in KDE5 regardless of KDE's language, but the Arabic strings in Unity if the DE's primary language is set to Arabic.

It may be because the name of the language is still `ar_SA` somewhere. I thought it is the 9th line in [`l10n_master/po/ar.po`](https://raw.githubusercontent.com/xournalpp/xournalpp/l10n_master/po/ar.po) (as it still is `"Language: ar_SA\n"`), but changing that and building doesn't solve the problem.

**To Reproduce:**
1. Install the latest version from the official PPA.
2. Launch Xournal++, Edit > Preferences > Language, and choose "ar", then "OK".
3. Restart Xournal++. (The language now is Arabic.)
4. Select the second menu («تحرير», Edit) > the last item («التفضيلات», Preferences).
5. It complains that «اللغة المختارة سابقا لم تعد متاحة», which is "Previously selected language not available anymore".
6. Click «نعم» (OK) to dismiss the message. Now it opens the Preferences window.
7. Click «نعم» (OK) to close the Preferences window, and restart Xournal++.
8. The language is now set to System Default (which likely uses the English strings).

An interesting observation: if you close the Preferences window with ESC or "Cancel" («ألغِ») instead of "OK" («نعم»), it doesn't change the language.

noureddin:
Please see https://github.com/xournalpp/xournalpp/issues/3200#issuecomment-872438041

### DIFF
--- a/src/gui/dialog/LanguageConfigGui.cpp
+++ b/src/gui/dialog/LanguageConfigGui.cpp
@@ -63,7 +63,7 @@ LanguageConfigGui::LanguageConfigGui(GladeSearchpath* gladeSearchPath, GtkWidget
     // Set the current locale if previously selected
     auto prefPos = availableLocales.begin();
     if (auto preferred = settings->getPreferredLocale(); !preferred.empty()) {
-        prefPos = std::lower_bound(availableLocales.begin(), availableLocales.end(), preferred);
+        prefPos = std::find(availableLocales.begin(), availableLocales.end(), preferred);
         if (*prefPos != preferred) {
             XojMsgBox::showErrorToUser(nullptr, _("Previously selected language not available anymore!"));
 


noureddin:
This PR, earlier, changed `std::lower_bound` to `std::find`, which solved the problem, but had a worse complexity (`O(n)` for `find` vs `O(log(n))` for `lower_bound`); thanks for [Febbe (Fabian Keßler) for pointing that](https://matrix.to/#/!mIWOZbAzBNmtxiyIii:gitter.im/$mgZqDga5OZq7bH_SYqMZbMtzg8QKviXECgt1RJMO4CI?via=gitter.im&via=matrix.org).

So, I reverted to the original `lower_bound`, and tried to solve it another way: rename the first locale from `_("System Default")` to `" "` (a space), after adding it the UI, so it's always the first, without affecting its visible content.

Technius (commented):
I don't like this hack. Just switch the `std::lower_bound` to `std::find`; the increased maintenance cost of the currently proposed solution significantly outweighs the performance improvement.

noureddin:
I agree that `std::lower_bound` is not _significantly_ more performant than `std::find` in this specific case, and I agree that `std::find` is more understandable, and thus more maintainable.

But I don't have enough experience with large projects to see how this line (with its comment, and possibly a reference to the issue) would _significantly_ increase the maintenance cost.

I agree with both solutions, but now I'm receiving two different messages from the core team.

I'll be happy to implement the solution you, the team, see fit. Also I'll happy to know if there is a better solution than both. :smiley:

Technius:
As I see it, using `std::find` is the best solution. It adds zero net maintenance cost (we simply swap `std::lower_bound` with `std::find`), with a negligible performance penalty (e.g. opening the preferences window,  which is an uncommon event, becomes 0.01s faster--unnoticeable by any user). Asymptotics do not matter here--as far as we are concerned, `n` is a constant, finite number of translations, so both `std::find` and `std::lower_bound` are constant time in this context.

Sure, we can keep using `std::lower_bound` by adding an extra line of code. But some of the drawbacks include:
* The use of the function `std::lower_bound` requires us to maintain the invariant that the list is sorted. This invariant can easily be violated in a future change (e.g. what if we want to add another non-language option similar to "System Default")?
* Furthermore, the intent of using `std::lower_bound` is unclear, and at worse, downright misleading (since it would imply that we are searching for the "smallest" value for some definition of "small").
* Setting the "System Default" line to blank adds a special case, which means that we now have to add comments, explanations, or references explaining this special case. Then, if there happens to be a bug regarding translations in the future, people who try to fix the bug will need to understand what the snippet in this PR does. This would require reading through the comments on #3200 and on this PR, which could take up a lot of valuable free time.

noureddin:
Thank you for your detailed explanation. I just restored the first commit (that changed `std::lower_bound` to `std::find` with no other changes). :smiley:

Technius:

rolandlo (approved):
LGTM

Issue #3000, opened 2021-04-04
Let's have a place with full control over layers


Photon89:
**Is your feature request related to a problem? Please describe.**
There are several places where layers can be manipulated:

- in the Journal menu you can create, delete and rename a layer
- in the little status bar menu you can choose the active layer, show/hide layers and create a layer
- in the layer side bar you can choose the active layer, move, duplicate, delete and show/hide a layer

**Describe the solution you'd like**
There should be a place, presumably the side bar, where you have access to all layer related actions. So I'd propose to add actions which the side bar is still missing: rename and create a layer. They could be added to the button bar on the bottom of the side bar and also in a context menu which would appear if you right click a layer.

**Describe alternatives you've considered**
A similar problem is described here: https://github.com/xournalpp/xournalpp/issues/2494

Technius:
The context menu should be fairly straightforward to implement and would probably be similar to what I did in #1256 a while back.

ja-he:
As an update, in #3204 there is a new sidebar context menu `SidebarLayersContextMenu` (extending a new base context menu) which has move, remove and merge (the subject of that PR).

Extending it with further actions should be straightforward, once merged.

Issue #1911, opened 2020-04-28
Feature request: Clone all Layers with custom drawings to the next existing page


flo2k:
I do use xournal++ to annotate my PDF slides which I created with the latex beamer package. I heavily use the \pause and \visibility features to uncover the individual items. The resulting PDF with the slides contains one page per uncovered item. When I draw (annotate) something that belongs to item 1 and I uncover item 2, which means that I go the next page, then my custom drawings (annotations) are away. I would like to have a mechanism that allows me to reuse my previous drawings (annotations) on the next page easily.

Two buttons with: copy drawings (annotations) (with all layers - except of the background layer) to the next or from the previous slide would be perfect. 
I do also need shortcuts for that to be able to use it from my wacom tablet with the express keys.

I would really appreciate such a feature. Maybe this is already possible, but I don't know how.

LittleHuba:
Duplicate of #1647

Technius:
Not a duplicate--there is currently no way to duplicate or copy/paste layers.

flo2k:
Thank you for the hint to https://github.com/xournalpp/xournalpp/issues/1647. I think the problem is the same: we would like to have the existing drawings (annotations) within one slide (which can consist of several pages) visible.

My first approach was to think that I can do this manually by copying the existing annotations from the previous page to the current, or from the current to the next page as I described above.

But if xournal++ automatically detects that the slide consists of several pages than this is definitely also a good solution for me, if not the better one.

rolandlo:
What you currently can do in order to clone all handwritten layers to the next page (which is assumed to be without annotations): 

1. copy the whole page including background (in the sidebar), 
2. replace the background of the copied page with the next page background (from the menu journal->background)
3. delete the next page without annotations (in the sidebar)

flo2k:
@rolandlo, you are right. This is a solution if you need this only once. However, if you are in a lecture where you need this several times, then this is not a good solution.

michaelcadilhac:
I've spent a couple hours doing something along these lines, precisely for the same reasons.  This needs testing and proper icons, but this fits my needs so far.  The features are:
- Copy current layer to next page
- Copy all (equivalent to some "select all + copy")
- Paste in place; this allows things that live at one position on a slide to be pasted at the same position on another slide.

See https://github.com/michaelcadilhac/xournalpp/tree/incremental-layers

rolandlo:
[...]

PR #3204, opened 2021-07-02
Implement layer merging

### LINKED ISSUE #2784: Be Able to Merge Two Layers Together
It would be nice to be able to combine/merge two layers.

Perhaps when in the layer viewer, there could be a button, or you could right click a layer and select something like "merge up" or "merge down", which as the name implies, merges the layer in question with the layer directly above it or directly below it. The order in which objects will be combined on the layer depends on the order that the two layers in question are in the layer viewer. If one layer is on top of the other, then when combined, the layer which was on top, will have its content drawn over top of the layer below it when they are combined.

ja-he:
### Basics

This functionality would previously have to be emulated by copy, switch layer, paste, select old layer, delete layer.  
This PR  implements it (hopefully) properly with corresponding undo/redo actions.
It's split into 2 initial commits implementing functionality and GUI hookup respectively.

edit: resolves #2784 

### Where to put it in the GUI?

_Edit: I've removed outdated pondering here, instead here's where it is in the GUI..._

The functionality is available through:

* A new context menu opened by right-clicking on a layer preview in the sidebar
* The 'Journal' Menu
* The shortcut `<c-m>` (`CTRL_M`)

2021-07-23-170927_3000x2224_scrot
2021-07-23-171659_3000x2224_scrot

### TODO

I've done some basic testing on this with strokes and text, testing the merge itself, the undo and the redo. Everything seemed to work as expected with no warnings.

There are some TODOs left, some related to i18n, the others can probably just be removed, but I figured I'd leave them in for the WIP state.

Any feedback on this would be appreciated.


rolandlo:
@ja-he I will just make some first comments. Once version 1.1.0 is released, I can take a closer look. 

- Merging layers is really missing in Xournal++, so I am happy you made this PR. 
- Closes #2784. Your PR is related to [this suggestion](https://github.com/xournalpp/xournalpp/issues/1911#issuecomment-689187172)
- First impression: The code is easy readable and the feature seems to work fine.
- Just noticed: Undo/redo does work properly with layer renaming. That seems unrelated to your PR though.
- I think layer merging should be made accessible from the menubar, so that it can also be used via the plugin system. The `LayerActions` plugin allows to clone all non-background layers to the next page. By default it should merge the layers, I think. so your merging functionality comes in handily.
- In general we have to figure out the best solution to #3000. Probably this solution would use a context menu.

ja-he:
I appreciate your feedback and you taking the time to point me to relevant issues/PRs (I did actually look for an issue when I initially wanted this feature, but I was too lazy to make one myself and never got around to implementing this until yesterday).

I'm having a look at #1256 based on @Technius' comment in #3000, looking to implement the basic context menu. This GUI stuff is fairly foreign to me, so I'll probably just copy-paste with the obvious adjustments.

rolandlo:
@ja-he Can you please rebase on current master and solve the merge conflicts?

ja-he:
@rolandlo sure, I'm on it.

ja-he:
One sec on this, I shot myself in the foot trying to be clever in the rebase.

ja-he:
Alright, apologies for the messy commits (my first PR). Now everything should be up to date.

ja-he:
Alright, this should now be fine and up to date (but the UI hookup is basically a sketch, so this shouldn't be merged just yet).
Apologies for the messy commits here, it's my first PR. (However I take no responsibility for the formatting confusion since my `clang-format` literally gives me different output depending on if I run `cat <file> | clang-format` or `clang-format <file>`.)

rolandlo:
> Alright, this should now be fine and up to date (but the UI hookup is basically a sketch, so this shouldn't be merged just yet).

The dropdown menu works fine. Can you add an entry to the menubar (in menu Journal)? As already pointed out, that would be useful for plugins as well.

> Apologies for the messy commits here, it's my first PR. (However I take no responsibility for the formatting confusion since my `clang-format` literally gives me different output depending on if I run `cat <file> | clang-format` or `clang-format <file>`.)

Could it be that you forgot the `--style=file` option? This will apply clang format according to our `.clang.format` file in the root directory of the repository.

ja-he:
> Could it be that you forgot the `--style=file` option? This will apply clang format according to our `.clang.format` file in the root directory of the repository.

I thought looking for `.clang-format` would be the default behaviour, but regardless, the difference persists even with the flag.
It can't decide where to put the `#include` (top: filename given, bottom: piped):

Separate issue is that some of the source files aren't properly formatted in the first place, so after formatting I had to stage only the hunks I actually had something to do with.

ja-he:
> Can you add an entry to the menubar (in menu Journal)? As already pointed out, that would be useful for plugins as well.

Done. It's good you mentioned this again because I missed/misunderstood that suggestion in your initial notes.

Is the context menu desired in this PR then? I only added it to have a way to invoke the behavior from the GUI.

personalizedrefrigerator:
> Is the context menu desired in this PR then? I only added it to have a way to invoke the behavior from the GUI.

Personally, I like the context menu. Since we already have a context menu for page previews, I think a context menu like this makes sense.

It would be nice to have an alternative way to access the same functionality (maybe under the Edit menu). This makes the feature available to plugins and makes it more usable on some touchscreen-only devices (long press seems to not automatically open a context menu on Fedora/GNOME and Manjaro/KDE).

rolandlo:
> > Can you add an entry to the menubar (in menu Journal)? As already pointed out, that would be useful for plugins as well.
> 
> Done. It's good you mentioned this again because I missed/misunderstood that suggestion in your initial notes.

Good! Just change (or remove) the keyboard accelerator, since CTRL+SHIFT+L is already used for deleting a layer.
@personalizedrefrigerator  It's already implemented, in menu `Journal`.
> 
> Is the context menu desired in this PR then? I only added it to have a way to invoke the behavior from the GUI.

Yes, the context menu is desired. We can extend on that in a later PR. It would be nice if the commits for adding the context menu and for adding the menubar entry remained clearly separated.

ja-he:
> Good! Just change (or remove) the keyboard accelerator, since CTRL+SHIFT+L is already used for deleting a layer.

Done, I chose `CTRL_M` since a search for `<accelerator\s*key="m"` showed no results so I'm assuming it's free (and intuitive).

> It would be nice if the commits for adding the context menu and for adding the menubar entry remained clearly separated.

I'll clean up the commits a bit then.

ja-he:
I'll finish this up some time next week, when I find the time.
From my perspective here's what's left to do on this:

* [ ] Polish up the context menu
    * [ ] Add the other basic items (might as well, while we're at it)
    * [ ] Figure out how to enable the items only when available (e.g. merge down only when not on lowest or background layer); I already have looked at how to do this and have a rough idea, couldn't get it to work right just yet.
* [ ] Clean up commits
* [ ] i18n stuff? _I don't know but I imagine somewhere something needs to be added_

rolandlo:
`i18n` stuff is already taken care of. The code line with `_("Merge layer down")` will automatically trigger an update of the localization template `xournalpp.pot` and the localizations in the supported languages. Then our translators can update the individual localization files over [Crowdin](https://Crowdin.com/project/xournalpp). The updates will be automatically uploaded in a Github pull request, which we will merge after a while.

ja-he:
From my perspective this is ready for review (this time for real). I'd like to think this could get merged at this point (though from what I understand you're targeting the end of October, so there's no rush).

Some pointers:
 * adds accelerator `CTRL+M`, compatible with other PRs?
 * the new context menu for layers is a little inconsistent from the one for pages (I don't mind straightening out the naming, if required, but further cleanup would probably involve generifying some stuff from the `...pages`/`...layers` code back into their `...base` and more thought given to not double(triple?)-registering callbacks)
 * I've tested moderately, and no longer come across any issues, besides #3290, which affects 1.1.0 in general; further testing welcome
 * any feedback welcome :slightly_smiling_face:

src/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
```diff
@@ -7,20 +7,82 @@
 #include "SidebarPreviewLayerEntry.h"
 #include "i18n.h"
 
+// Since we have two layer sidebars, we need to make sure we don't
+// double-register callbacks. Since we also need to make sure we only do the
+// proper deallocation stuff on destuction of the last SidebarPreviewLayers
+// object, we use a static counter which we count up in construction and down
+// in destruction.
+static int layerSidebarsActive = 0;
```
personalizedrefrigerator, line 15: Thank you for the detailed comment! I don't like having static, mutable variables like this, though...

I think I might like the following better:
 * Create `BaseContextMenu` and `SidebarLayersContextMenu` classes. Give a `std::shared_ptr<SidebarLayersContextMenu>` to both `SidebarPreviewLayers`.
    * I (or anyone else who wants to) could refactor `SidebarPreviewPages` in a follow-up PR
    * This simplifies the code in `SidebarPreviewLayers` and gets rid of the `static` variable.

An issue with my suggestion is that, until we refactor `SidebasrPreviewPages` to use `BaseContextMenu`, there will be a potentially-confusing inconsistency.

src/gui/sidebar/previews/layer/SidebarPreviewLayerEntry.cpp
```diff
@@ -6,13 +6,29 @@
...
+            auto mouseEvent = reinterpret_cast<GdkEventButton*>(event);
+            if (mouseEvent->button == 3) {
+                self->mouseButtonPressCallback();
+                self->sidebar->openPreviewContextMenu();
+                return true;
+            }
+        }
+        return false;
+    });
+    g_signal_connect_after(this->widget, "button-press-event", clickCallback, this);
```
personalizedrefrigerator, line 30: Because we now have `clickCallback`s in both `SidebarPreviewLayerEntry` and `SidebarPreviewPageEntry`, I think it might make sense to factor this code out into `SidebarPreviewBaseEntry`.

`void openPreviewContextMenu(SidebarPreviewBaseEntry&)` could be a `virtual` method of `SidebarPreviewBase` that we `override` in `SidebarPreviewLayer` and `SidebarPreviewPage`.

This design would be particularly useful if we want to make long taps open the context menu in the future (on systems like Linux/Wayland).

src/gui/sidebar/previews/layer/SidebarPreviewLayers.h
```diff
@@ -82,4 +89,31 @@ class SidebarPreviewLayers: public SidebarPreviewBase, public LayerCtrlListener
...
+    struct ContextMenuData {
+        SidebarToolbar* toolbar;
+        SidebarActions actions;
+    };
+
+    /**
+     * The signals connected to the context menu items. This must be kept track
+     * of so the data can be deallocated safely.
+     */
+    std::vector<std::tuple<GtkWidget*, gulong, std::unique_ptr<ContextMenuData>>> contextMenuSignals;
```
personalizedrefrigerator, line 118: See earlier comment: I think it would be nice to factor this out into a `SidebarPreviewLayersContextMenu` class (or similar) that both instances of `SidebarPreviewLayers` have a pointer to.

src/undo/MergeLayerDownUndoAction.h
```diff
@@ -0,0 +1,50 @@
...
+
+    // TODO: Did not copy non-default destructor definition:
+    //       There is a non-default destructor defined in
+    //       'src/undo/MoveLayerUndoAction.{h,cpp}' which simply sets all
+    //       pointer members to `nullptr`. Is there any reason to do this?
+    //       All I can think of is assuming that some other part of our program
+    //       still holds a pointer to this destroyed undo/redo action and if its
+    //       access doesn't segfault it might try to access these old non-null
+    //       pointers which could then cause a segfault one level deeper.
+    //       Seems convoluted to me, but maybe I'm missing something?
```
personalizedrefrigerator, line 35: I don't think we need a `virtual` destructor here (even though most other undo actions do have a virtual destrouctor).

If the reason for including a virtual destructor was to ensure that `virtual UndoAction::~UndoAction` was called, [it should be fine to omit it](https://stackoverflow.com/a/2198400/17055750).

src/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
```diff
@@ -179,6 +245,37 @@ void SidebarPreviewLayers::layerSelected(size_t layerIndex) {
...
+    if (layerIndex < bgIndex) {
+        actions |= SIDEBAR_ACTION_COPY;
+        actions |= SIDEBAR_ACTION_DELETE;
+    }
+
+    gtk_widget_set_sensitive(this->contextMenuMoveUp, actions & SIDEBAR_ACTION_MOVE_UP);
+    gtk_widget_set_sensitive(this->contextMenuMoveDown, actions & SIDEBAR_ACTION_MOVE_DOWN);
+    gtk_widget_set_sensitive(this->contextMenuMergeDown, actions & SIDEBAR_ACTION_MERGE_DOWN);
+    gtk_widget_set_sensitive(this->contextMenuDuplicate, actions & SIDEBAR_ACTION_COPY);
+    gtk_widget_set_sensitive(this->contextMenuDelete, actions & SIDEBAR_ACTION_DELETE);
```
personalizedrefrigerator, line 278: There might be a bug here! In my tests, none of the menus are desensitized when I'm using the **Layer Stack Preview**. It works fine with the other preview pane.

None of the background page's menus are desensitized

Move first page up menu isn't desensitized

src/control/layer/LayerController.h
```diff
@@ -106,6 +107,10 @@ class LayerController: public DocumentListener {
      */
     void ensureLayerExists(PageRef page);
 
+    // Allows us to rebuild the layer menu from the undo action class upon
+    // redoing a merge.
+    friend class MergeLayerDownUndoAction;
```
personalizedrefrigerator, line 112: I think it might make more sense to add a public `rebuildLayerMenu` or `fireLayerUpdate` method to `LayerControler` than adding `MergeLayerDownUndoAction` as a `friend` class... I'm not sure though...

personalizedrefrigerator (changes requested):
Thank you for working on this! (I really want this feature).

There are a few things I would like to be resolved before we merge this:

ja-he:
@personalizedrefrigerator thanks for the review and the good suggestions; I'll dedicate some time to this this weekend.

src/gui/sidebar/previews/layer/SidebarPreviewLayerEntry.cpp
ja-he: > [abstract `clickCallback`]

Sure. To do that as far as I can tell we'll need to either
- let `SidebarPreviewBaseEntry` be a `friend` of `SidebarPreviewBase` (I'll do this for now)
- make `SidebarPreviewBase::openPreviewContextMenu()` public

src/control/layer/LayerController.cpp
```diff
@@ -222,6 +227,54 @@ void LayerController::moveCurrentLayer(bool up) {
     fireRebuildLayerMenu();
 }
 
+void LayerController::mergeCurrentLayerDown(void) {
+    control->clearSelectionEndText();
+
+    PageRef page = getCurrentPage();
+    auto pageID = selectedPage;
+    if (!page) {
+        // if there is no current page (nullptr) return
```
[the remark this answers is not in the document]
ja-he: :+1: especially the noisy comment; just to note: some other comparisons in that file will remain `!page`

src/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
ja-he: You are correct; they're apparently only sensitized based on the selection in the layer preview. There are also some failed assertions `gtk_widget_set_sensitive: assertion 'GTK_IS_WIDGET (widget)' failed`.

src/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
ja-he: This is a lot cleaner for sure. I did the static counter mostly because I didn't want to touch too much "other" code, but this was misguided considering how unwieldy this variable would have been. Thanks for the nudge.

ja-he: Switching the context menu for the pages sidebar would look something like this: 

```diff
diff --git a/src/gui/sidebar/previews/page/SidebarPagesContextMenu.cpp b/src/gui/sidebar/previews/page/SidebarPagesContextMenu.cpp
new file mode 100644
index 00000000..b5ff69aa
--- /dev/null
+++ b/src/gui/sidebar/previews/page/SidebarPagesContextMenu.cpp
@@ -0,0 +1,46 @@
+#include "gui/sidebar/previews/page/SidebarPagesContextMenu.h"
+
+#include <memory>
+
+SidebarPagesContextMenu::SidebarPagesContextMenu(GladeGui* gui, SidebarToolbar* toolbar):
+        SidebarBaseContextMenu(gui->get("sidebarPreviewContextMenu")) {
+
+    // Connect the context menu actions
+    const std::map<std::string, SidebarActions> ctxMenuActions = {
+            {"sidebarPreviewDuplicate", SIDEBAR_ACTION_COPY},
+            {"sidebarPreviewDelete", SIDEBAR_ACTION_DELETE},
+            {"sidebarPreviewMoveUp", SIDEBAR_ACTION_MOVE_UP},
+            {"sidebarPreviewMoveDown", SIDEBAR_ACTION_MOVE_DOWN},
+            {"sidebarPreviewNewBefore", SIDEBAR_ACTION_NEW_BEFORE},
+            {"sidebarPreviewNewAfter", SIDEBAR_ACTION_NEW_AFTER},
+    };
+
+    for (const auto& pair: ctxMenuActions) {
+        GtkWidget* const entry = gui->get(pair.first);
+        g_assert(entry != nullptr);
+
+        // Unfortunately, we need a fairly complicated mechanism to keep track
+        // of which action we want to execute.
+        using Data = ContextMenuData;
+        auto userdata = std::make_unique<Data>(Data{toolbar, pair.second});
+
+        const auto callback =
+                G_CALLBACK(+[](GtkMenuItem* item, Data* data) { data->toolbar->runAction(data->actions); });
+        const gulong signalId = g_signal_connect(entry, "activate", callback, userdata.get());
+        g_object_ref(entry);
+        this->contextMenuSignals.emplace_back(entry, signalId, std::move(userdata));
+
+        if (pair.first == "sidebarPreviewMoveDown") {
+            this->contextMenuMoveDown = entry;
+        } else if (pair.first == "sidebarPreviewMoveUp") {
+            this->contextMenuMoveUp = entry;
+        }
+    }
+    g_assert(this->contextMenuMoveDown != nullptr);
+    g_assert(this->contextMenuMoveUp != nullptr);
+}
+
+void SidebarPagesContextMenu::setActionsSensitivity(SidebarActions actions) {
+    gtk_widget_set_sensitive(this->contextMenuMoveUp, actions & SIDEBAR_ACTION_MOVE_UP);
+    gtk_widget_set_sensitive(this->contextMenuMoveDown, actions & SIDEBAR_ACTION_MOVE_DOWN);
+}
diff --git a/src/gui/sidebar/previews/page/SidebarPagesContextMenu.h b/src/gui/sidebar/previews/page/SidebarPagesContextMenu.h
new file mode 100644
index 00000000..31c2eaba
--- /dev/null
+++ b/src/gui/sidebar/previews/page/SidebarPagesContextMenu.h
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <map>
+#include <memory>
+#include <vector>
+
+#include <gtk/gtk.h>
+
+#include "gui/sidebar/previews/base/SidebarBaseContextMenu.h"
+#include "gui/sidebar/previews/base/SidebarToolbar.h"
+
+
+class SidebarPagesContextMenu: public SidebarBaseContextMenu {
+public:
+    SidebarPagesContextMenu(GladeGui* gui, SidebarToolbar* toolbar);
+
+    void setActionsSensitivity(SidebarActions actions) override;
+
+private:
+    GtkWidget* contextMenuMoveUp = nullptr;
+    GtkWidget* contextMenuMoveDown = nullptr;
+};
diff --git a/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp b/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
index 5b9126d4..61fc4700 100644
--- a/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
+++ b/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
@@ -13,53 +13,8 @@
 
 SidebarPreviewPages::SidebarPreviewPages(Control* control, GladeGui* gui, SidebarToolbar* toolbar):
         SidebarPreviewBase(control, gui, toolbar),
-        contextMenu(gui->get("sidebarPreviewContextMenu")),
-        iconNameHelper(control->getSettings()) {
-    // Connect the context menu actions
-    const std::map<std::string, SidebarActions> ctxMenuActions = {
-            {"sidebarPreviewDuplicate", SIDEBAR_ACTION_COPY},
-            {"sidebarPreviewDelete", SIDEBAR_ACTION_DELETE},
-            {"sidebarPreviewMoveUp", SIDEBAR_ACTION_MOVE_UP},
-            {"sidebarPreviewMoveDown", SIDEBAR_ACTION_MOVE_DOWN},
-            {"sidebarPreviewNewBefore", SIDEBAR_ACTION_NEW_BEFORE},
-            {"sidebarPreviewNewAfter", SIDEBAR_ACTION_NEW_AFTER},
-    };
-
-    for (const auto& pair: ctxMenuActions) {
-        GtkWidget* const entry = gui->get(pair.first);
-        g_assert(entry != nullptr);
-
-        // Unfortunately, we need a fairly complicated mechanism to keep track
-        // of which action we want to execute.
-        using Data = SidebarPreviewPages::ContextMenuData;
-        auto userdata = std::make_unique<Data>(Data{this->toolbar, pair.second});
-
-        const auto callback =
-                G_CALLBACK(+[](GtkMenuItem* item, Data* data) { data->toolbar->runAction(data->actions); });
-        const gulong signalId = g_signal_connect(entry, "activate", callback, userdata.get());
-        g_object_ref(entry);
-        this->contextMenuSignals.emplace_back(entry, signalId, std::move(userdata));
-
-        if (pair.first == "sidebarPreviewMoveDown") {
-            this->contextMenuMoveDown = entry;
-        } else if (pair.first == "sidebarPreviewMoveUp") {
-            this->contextMenuMoveUp = entry;
-        }
-    }
-    g_assert(this->contextMenuMoveDown != nullptr);
-    g_assert(this->contextMenuMoveUp != nullptr);
-}
-
-SidebarPreviewPages::~SidebarPreviewPages() {
-    for (const auto& signalTuple: this->contextMenuSignals) {
-        GtkWidget* const widget = std::get<0>(signalTuple);
-        const guint handlerId = std::get<1>(signalTuple);
-        if (g_signal_handler_is_connected(widget, handlerId)) {
-            g_signal_handler_disconnect(widget, handlerId);
-        }
-        g_object_unref(widget);
-    }
-}
+        contextMenu(std::make_shared<SidebarPagesContextMenu>(gui, toolbar)),
+        iconNameHelper(control->getSettings()) {}
 
 void SidebarPreviewPages::enableSidebar() {
     SidebarPreviewBase::enableSidebar();
@@ -299,11 +254,9 @@ void SidebarPreviewPages::pageSelected(size_t page) {
         this->toolbar->setHidden(false);
         this->toolbar->setButtonEnabled(static_cast<SidebarActions>(actions));
 
-        gtk_widget_set_sensitive(this->contextMenuMoveUp, actions & SIDEBAR_ACTION_MOVE_UP);
-        gtk_widget_set_sensitive(this->contextMenuMoveDown, actions & SIDEBAR_ACTION_MOVE_DOWN);
+        this->contextMenu->setActionsSensitivity(SidebarActions(actions));
     }
 }
 
-void SidebarPreviewPages::openPreviewContextMenu() {
-    gtk_menu_popup(GTK_MENU(this->contextMenu), nullptr, nullptr, nullptr, nullptr, 3, gtk_get_current_event_time());
-}
+// TODO: does this get used?!
+void SidebarPreviewPages::openPreviewContextMenu() { this->contextMenu->open(); }
diff --git a/src/gui/sidebar/previews/page/SidebarPreviewPages.h b/src/gui/sidebar/previews/page/SidebarPreviewPages.h
index d5c36241..eb19689b 100644
--- a/src/gui/sidebar/previews/page/SidebarPreviewPages.h
+++ b/src/gui/sidebar/previews/page/SidebarPreviewPages.h
@@ -17,13 +17,12 @@
 #include <vector>
 
 #include "gui/sidebar/previews/base/SidebarPreviewBase.h"
+#include "gui/sidebar/previews/page/SidebarPagesContextMenu.h"
 #include "util/IconNameHelper.h"
 
-
 class SidebarPreviewPages: public SidebarPreviewBase {
 public:
     SidebarPreviewPages(Control* control, GladeGui* gui, SidebarToolbar* toolbar);
-    virtual ~SidebarPreviewPages();
 
 public:
     /**
@@ -69,28 +68,8 @@ private:
      */
     void unselectPage();
 
-    /**
-     * The context menu to display when a page is right-clicked.
-     */
-    GtkWidget* const contextMenu = nullptr;
-
-    GtkWidget* contextMenuMoveUp = nullptr;
-    GtkWidget* contextMenuMoveDown = nullptr;
-
-    /**
-     * The data passed to the menu item callbacks.
-     */
-    struct ContextMenuData {
-        SidebarToolbar* toolbar;
-        SidebarActions actions;
-    };
-
+    std::shared_ptr<SidebarPagesContextMenu> contextMenu;
 
-    /**
-     * The signals connected to the context menu items. This must be kept track
-     * of so the data can be deallocated safely.
-     */
-    std::vector<std::tuple<GtkWidget*, gulong, std::unique_ptr<ContextMenuData>>> contextMenuSignals;
 
 private:
     IconNameHelper iconNameHelper;
```

src/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
ja-he: This issue is fixed by implementing the context menu as its own class such that the layer sidebars each have a pointer to the same context menu instance (`SidebarLayersContextMenu`).

ja-he:
I implemented most of the suggestions from the review by @personalizedrefrigerator now. For now they're temporarily their own commits but I'll integrate them into the three commits (pure functionality, context menu, dropdown menu) as they fit. (At least this is my plan; first PR and all I'm hoping this isn't sand in the review gears :slightly_smiling_face: ).

Outstanding:
* [ ] there's a bit of a bug where when the context menu is opened the first time the sensitization isn't right because the layer selection has not changed yet; this was a touch more tricky to get right so I'm leaving it for another day for now
* [ ] fix build by rebasing on current main
* [ ] make layer menu rebuilding public, probably

ja-he:
Hm, the windows-pipeline failed because of connection-timeout trying to build 'pipeaudio'?

personalizedrefrigerator:
> Hm, the windows-pipeline failed because of connection-timeout trying to build 'pipeaudio'?

The Windows pipeline is broken right now (for other PRs, too).

src/gui/sidebar/previews/layer/SidebarPreviewLayers.h
```diff
@@ -71,6 +80,8 @@ class SidebarPreviewLayers: public SidebarPreviewBase, public LayerCtrlListener
     virtual void pageChanged(size_t page);
 
 private:
+    SidebarActions getViableActions(size_t layerIndex, size_t layerCount) const;
```
personalizedrefrigerator, line 83:
```suggestion
    /**
     * @return things that can reasonably be done to a given layer (e.g. merge down, copy, delete, etc.)
     */
    SidebarActions getViableActions(size_t layerIndex, size_t layerCount) const;
```

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
personalizedrefrigerator, line 1:
```suggestion
/*
 * Xournal++
 *
 * Context menu layers sidebars
 *
 * @author Xournal++ Team
 * https://github.com/xournalpp/xournalpp
 *
 * @license GNU GPLv2 or later
 */
 
#pragma once
```

src/gui/sidebar/previews/base/SidebarBaseContextMenu.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
```
personalizedrefrigerator, line 1:
```suggestion
/*
 * Xournal++
 *
 * Abstract base class for sidebar context menus
 *
 * @author Xournal++ Team
 * https://github.com/xournalpp/xournalpp
 *
 * @license GNU GPLv2 or later
 */
 
#pragma once
```

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.cpp
```diff
@@ -0,0 +1,53 @@
...
+        SidebarBaseContextMenu(gui->get("sidebarPreviewLayersContextMenu")) {
+
+    // Connect the context menu actions
+    const std::map<std::string, SidebarActions> ctxMenuActions = {
+            {"sidebarPreviewLayerMoveUp", SIDEBAR_ACTION_MOVE_UP},
+            {"sidebarPreviewLayerMoveDown", SIDEBAR_ACTION_MOVE_DOWN},
+            {"sidebarPreviewMergeDown", SIDEBAR_ACTION_MERGE_DOWN},
+            {"sidebarPreviewLayerDuplicate", SIDEBAR_ACTION_COPY},
+            {"sidebarPreviewLayerDelete", SIDEBAR_ACTION_DELETE},
+            // TODO: the behavior for SIDEBAR_ACTION_NEW_BEFORE/AFTER is only defined for pages
```
personalizedrefrigerator, line 13: See our [coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#todo-comments)' section on `TODO`s:

```suggestion
            // TODO(ja-he): the behavior for SIDEBAR_ACTION_NEW_BEFORE/AFTER is only defined for pages
```

personalizedrefrigerator (changes requested):
Thank you for continuing to work on this!

I haven't noticed any issues while testing this.

I have a few comments, but other than those, this PR seems ready to be merged!

src/control/layer/LayerController.h
LittleHuba: Definitely do not use a friend class here. While some people think this is okay, it makes understanding the code really hard and is not worth the "added protection" of not having this method be public.

src/gui/sidebar/previews/base/SidebarPreviewBase.h
```diff
@@ -127,4 +132,5 @@ class SidebarPreviewBase: public AbstractSidebarPage {
     bool enabled = false;
 
     friend class SidebarLayout;
+    friend class SidebarPreviewBaseEntry;
```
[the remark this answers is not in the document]
ja-he: I'll remove that then and make the relevant member function public?

src/control/layer/LayerController.h
LittleHuba: Just as a follow-up for clarification:
From the top of my mind, there are only a few very limited cases where you would be interested in using a friend class. And those are really specific cases like:

1. An iterator for multiple custom classes that must access private state within these classes
2. A really big class you want to split up into smaller units for maintenance but treat them as one big class from the outside
3. A unit test requires a check of an internal state and you really have no good way of giving it access without going against common sense (and this is really already a code smell since your unit tests should only ever cover the public interfaces of your units - that is their whole point)

So whenever you happen to stumble upon a friend class/function, you most likely look either at a quick-and-dirty hack around encapsulation or a system that has grown over time, and refactoring would equal rewriting the full codebase.

And while risking an angry mob outside my window: No friend classes do not mirror the concept of `protected` from Java... Whoever got that idea out into the world...

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.cpp
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+
+    for (const auto& [name, action]: ctxMenuActions) {
+
+        GtkWidget* const entry = gui->get(name);
+        g_assert(entry != nullptr);
+
+        // Unfortunately, we need a fairly complicated mechanism to keep track
+        // of which action we want to execute.
```
personalizedrefrigerator, line 23: It might make more sense here to describe the code below. For example,
```suggestion
        // Create callbacks, store the entry
```

personalizedrefrigerator (approved):
This looks good to me!

<details><summary>Potential topics for a <b>follow-up PR</b>:</summary>

 * Rename layer action in the context menu
 * New layer above action in the context menu
 * Open the context menu on tap+hold on Linux touchscreen devices (I think Windows simulates a right-click)

</details>

personalizedrefrigerator:
@xournalpp/core Merging in 72 hours unless there are objections!

src/control/layer/LayerController.cpp
```diff
@@ -222,6 +227,53 @@ void LayerController::moveCurrentLayer(bool up) {
...
+    auto pageID = selectedPage;
+    if (page == nullptr) {
+        return;
+    }
+
+    // layerID value:
+    //    ...
+    //    2: layer 2
+    //    1: layer 1
+    //    0: background
```
LittleHuba, line 243: Please use block comments for multi-line comments. Using single-line comments for multiple lines can lead to serious problems down the way. (MISRA coding style)
```
/*
 *
 */
```

src/control/layer/LayerController.cpp
```diff
@@ -222,6 +227,53 @@ void LayerController::moveCurrentLayer(bool up) {
     fireRebuildLayerMenu();
 }
 
+void LayerController::mergeCurrentLayerDown(void) {
+    control->clearSelectionEndText();
+
+    PageRef page = getCurrentPage();
+    auto pageID = selectedPage;
```
LittleHuba, line 234: Add a comment what this is required for

src/gui/sidebar/previews/base/SidebarBaseContextMenu.h
```diff
@@ -0,0 +1,62 @@
...
+    SidebarBaseContextMenu(const SidebarBaseContextMenu&) = delete;
+    SidebarBaseContextMenu& operator=(SidebarBaseContextMenu&&) = delete;
+    SidebarBaseContextMenu& operator=(const SidebarBaseContextMenu&) = delete;
+
+    /**
+     * Sets the sensitivity of the actions in the menu according to the actions
+     * specified by the passed `actions`.
+     *
+     * If the context menu had widgets (entries) for the actions A, B, C, D
+     * this were passed the actions value (A|C) then it would sensitize (make
```
LittleHuba, line 34:
```suggestion
     * and this method was passed the actions value (A|C) then it would sensitize (make
```

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.cpp
```diff
@@ -0,0 +1,52 @@
...
+        using Data = ContextMenuData;
+        auto userdata = std::make_unique<Data>(Data{toolbar, action});
+
+        const auto callback =
+                G_CALLBACK(+[](GtkMenuItem* item, Data* data) { data->toolbar->runAction(data->actions); });
+        const gulong signalId = g_signal_connect(entry, "activate", callback, userdata.get());
+        g_object_ref(entry);
+        this->contextMenuSignals.emplace_back(entry, signalId, std::move(userdata));
+
+        if (name == "sidebarPreviewLayerMoveUp") {
```
LittleHuba, line 32: Please do not duplicate strings in the classes. This can lead to really awful bugs caused by typos. Instead, predefine the strings in a shared context and reuse them through the class:

```
constexpr char* constString = "something";
```

src/undo/MergeLayerDownUndoAction.cpp
```diff
@@ -0,0 +1,62 @@
...
+    // properly displayed in the sidebar. Without this, the visually displayed
+    // selection defaults to the top layer, but if the user were to delete the
+    // current layer via `shift+ctrl+l` the merged layer would disappear, as it
+    // is the one actually selected.
+    layerController->fireRebuildLayerMenu();
+
+    return true;
+}
+
+auto MergeLayerDownUndoAction::redo(Control* control) -> bool {
```
LittleHuba, line 49: This should not reimplement the action but use the actual implementation of the action. Otherwise, changes to the actual action might miss the implementation here. To limit the coupling of the modules, you might want to keep the implementation here and use it within the controller instead of the other way around.

LittleHuba (changes requested):
Looks quite nice already. Just some nitpick and a few potential future issues I'd like to tackle earlier rather than later.

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.cpp
ja-he: Fair point, again the code structure mirrors the implementation of the pages context menu I used as a referenced.
I do believe this should be `constexpr const char* asdf = "asdf"` though, else the compiler warns, rightfully so, because I could just do `asdf[0] = 'x'` afterwards without the additional `const`.

LittleHuba: Oh yeah sure that const makes absolute sense!
Using the old implementation as a reference is a double-edged sword. While it is good to look at it, there is no promise that it is really that good. We have some serious mistakes in the codebase that are only slowly removed over time.

LittleHuba:
Please also rebase your PR on the latest commit of `master`. Another PR made significant changes to the repository structure. You'll most likely have to edit some include paths and move some files to the appropriate directories.

src/gui/sidebar/previews/layer/SidebarLayersContextMenu.cpp
Febbe: Writing `constexpr auto* asdf = "asdf";` is also valid.

src/undo/MergeLayerDownUndoAction.cpp
ja-he: As I don't see any other of the layer actions implement it this way, I have to clarify: have the actual implementation of the action in the Undo Action's `redo()` member function and then in the Layer Controller's implementation initialize the undo action, call `redo()` to do the action initially and then give that undo action to the Undo Redo Handler?

LittleHuba: Yes that would be my way of doing this. It keeps the full logic of the action contained in one class and whenever you want to do or undo it you have to use the class.

src/control/layer/LayerController.cpp
ja-he: Done, but I don't understand the reasoning yet; can you expand on this?  
From looking at MISRA C++ 2008 I can only find Section 6.2.7, which states nothing like this. I wasn't familiar with MISRA before this, but I did look at Xournalpp's Coding Conventions wiki page, which doesn't state anything about this.

src/undo/MergeLayerDownUndoAction.cpp
ja-he: Despite the small semantic confusion of `redo` = do for the first time, I do like this. However, it does expose some of the murkyness around some of this:

I call `LayerController::fireRebuildLayerMenu` to ensure that the selection (red outline) _shown_ in the sidebar actually reflects the current selection. I call `XournalView::layerChanged` to ensure that the page is rerendered (e.g. if the layer we merge with is invisible, the merge shouid change what the page looks like, hence rerender needed).  
__However, if I call `layerChanged` _before_ `fireRebuildLayerMenu` the program freezes.__

The rough cause of the freeze is that as `fireRebuildLayerMenu` goes through the `LayerControlListener`s and prompts them each to rebuild the menu, this call to the third listener never returns.

I'd like to investigate this a bit more (as opposed to just saying "it has to be this order, else it'll freeze, who knows why"), especially because it seems like it might be related to #3290?
Probably won't have much time to before next weekend, though.

If anybody has insight on this, feel free to contribute it.

ja-he: In the implementation of the undo/redo action if we call `fireRebuildLayerMenu`
_after_ `layerChanged` it will freeze the program, e.g. if we have the regular
layer sidebar open, because upon destruction of the old previews we end up
waiting forever (I didn't quite wait this long) on `jobRunningMutex` in
`XournalScheduler::finishTask`.

### Callstack:

    LayerController::fireRebuildLayerMenu
        SidebarPreviewLayers::rebuildLayerMenu
            SidebarPreviewLayers::updatePreviews
                SidebarPreviewBaseEntry::~SidebarPreviewBaseEntry
                    XournalScheduler::removeSidebar
                        XournalScheduler::removeSource
                            XournalScheduler::finishTask

ja-he: From what I can tell, to execute the render job that ends up being added because
of the `layerChanged` call, `jobRunningMutex` is locked.
The render job however is never completed, because to execute it the document
needs to be locked, which it already has been by the undo redo handler that
calls our undo/redo action, and so we get stuck here in `Document::lock`.
Thus the layer menu rebuild waits on the job lock which waits on the document
lock which waits on the overall caller (the undo/redo handler) to be done, which
of course can't happen until the layer rebuild returns.
Thus we are deadlocked if we add this render job before rebuilding the layer menu.

ja-he: Understanding this, I'm okay with just having the calls in the 'correct' order in the undo/redo action implementation, probably with a comment warning about the order. There's really no way for me to change anything else without touching stuff that has nothing to do with the PR (besides being over my head in complexity anyways).

To stick with the pattern of implementing actions in their undo/redo actions `redo()` member, this issue might pop up again.
I'd say it stems from the fact that as I implement it for example, the undo or redo really does multiple things:
- change the document (e.g. do the layer merge)
- update the UI (e.g. trigger page rerender and update layer menu)

Maybe it would make sense to give add some kind of member `post_action()` to the undo/redo action interface that the undo/redo handler could call after having executed the action and unlocked the document again. This could then do whatever additional updating might seem necessary.

@LittleHuba or @personalizedrefrigerator thoughts?

src/core/undo/MergeLayerDownUndoAction.cpp
```diff
@@ -0,0 +1,94 @@
...
+    return true;
+}
+
+auto MergeLayerDownUndoAction::redo(Control* control) -> bool {
+    /*
+     * Required to end a text edit in progress.
+     * The edited text will be included in the merge.
+     */
+    control->clearSelectionEndText();
+
```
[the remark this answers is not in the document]
ja-he: removed for now, because odd behavior when doing something odd (trying to redo a merge after starting a text edit) is better than a freeze

Technius:
Due to merge of the PDF toolbox feature, there are now merge conflicts with the main UI glade files. Please rebase and resolve; I can also provide assistance if needed.

ja-he:
> Let me know when you're ready for a review.

Should be.

For that, let me point out [this thread](https://github.com/xournalpp/xournalpp/pull/3204#discussion_r753872249) especially. As I note in the last comment, maybe a solution to the nearly-deadlock would be a `post_action()` the undo-redo handler could call e.g. for setting off UI refreshes after the document is unlocked.

A glance at [this](https://github.com/xournalpp/xournalpp/pull/3204#discussion_r768816960) might also be worthwhile.

Technius (approved):
Looks fine to me. In general, the current sidebar implementation is a giant ball of spaghetti code, so I don't think there much more we can do in this PR without refactoring the whole sidebar logic.

Several points that we will need to clarify in the future:
* Need to clearly define the expectations of undo/redo: when should model update happen? When should UI updates happen?
* The sidebar preview job logic could be better. For example, allocate the sidebar preview entry widget wrapper in a `shared_ptr` instead of with `new`, and then make sure each preview job has a reference to the widget wrapper. When the preview list is rebuilt, the individual entry widgets can mark themselves as invalid instead of waiting for current scheduler jobs to finish, so that any currently running preview jobs can abort. This idea is not entirely fleshed out, but it's one potential way to fix the whole mess.

src/undo/MergeLayerDownUndoAction.cpp
ja-he: I followed this up with #3844

ja-he:
Thanks for the review!

Here my final thoughts on follow-ups to this:
- I followed up on the idea for the `UndoAction` interface in #3844, which for me was the last unresolved aspect of this PR.
- The pages sidebar context menu (which currently lives in `SidebarPreviewPages` should probably become `SidebarPagesContextMenu` based on the `SidebarBaseContextMenu` that this PR adds; in https://github.com/xournalpp/xournalpp/pull/3204#discussion_r748914771 I have an old draft on what that could look like 
- `SIDEBAR_ACTION_NEW_BEFORE`/`AFTER` are not defined for layers, but are candidates for the context menu if they ever are (or the other way around: if they're desired in the new context menu, they'll have to be implemented for layers first)
- something like `SIDEBAR_ACTION_RENAME[_LAYER]` might be convenient, so renaming layers could be done via the new context menu
- as you say, the sidebar is a bit messy. it might make sense to shift the other actions' implementations into their `UndoAction`s, just like this PR ended up doing it.
- @personalizedrefrigerator [suggested](https://github.com/xournalpp/xournalpp/pull/3204#pullrequestreview-810701870) a tap-and-hold to trigger the context menu

I have nothing more to add to this PR :slightly_smiling_face:

Technius:

PR #3209, opened 2021-07-05
Update authors list and about dialog

rolandlo:
See AUTHORS file and old and new (current) image of the about dialog:

old:
grafik

new:
grafik

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,13 +1,162 @@
-Andreas Butti
-Wilson Brenna (since 2012)
-Marek Pikuła (since 2015)
-Moreno Razzoli (since 2018)
-Luca Errani (since 2018)
-Ulrich Huber (since 2019)
-Justin Jones (since 2019)
-Bryan Tan (since 2019)
-Justus Rossmeier (since 2019)
-Fabian Keßler (since 2019)
-lehmanju (since 2019)
-Roland Lötscher (since 2020)
-Tobias Hermann (since 2020)
+Developer Team:
+
+2010-2021, Andreas Butti <andreas.butti@gmail.com>
+2012-2018, Wilson Brenna <wilson@wbrenna.homelinux.com>
+2015-2016, Marek Pikuła <marek@pikula.tk>
+2017-2019, Peter Klausing <peetCreative@users.noreply.github.com>
+2018-2019, Moreno Razzoli <morrolinux@users.noreply.github.com>
+2018-2019, Luca Errani <luca.errani1@gmail.com>
+2019-2021, Ulrich Huber <ulrich@huberulrich.de>
+2019-2019, Justin Jones <bjustjones@netscape.net>
+2019-2021, Bryan Tan <Technius@users.noreply.github.com>
+2019-2019, Justus Rossmeier <veecue@users.noreply.github.com>
+2019-2021, Fabian Keßler <Fabian_Kessler@gmx.de>
+2019-2020, Julius Lehmann <internet@devpi.de>
+2020-2021, Tobias Hermann <idotobi@users.noreply.github.com>
+2020-2021, Roland Lötscher <roland_loetscher@hotmail.com>
+           
+
+Contributors from the community:
+
+Adolfo Jayme Barrientos <fitojb@ubuntu.com>
+Alex Veltman <alexveltman@protonmail.com>
+Andreas Stallinger <astallinger@posteo.net>
+Antonin Décimo <antonin.decimo@gmail.com>
+Archisman Panigrahi <apandada1@gmail.com>
+Ari Breitkreuz <ari.breitkreuz@pm.me>
+Atri Bhattacharya <badshah400@gmail.com>
+Awais Lodhi <awais.lodhi@gmail.com>
+Azure Pipelines <azuredevops@microsoft.com>
+Barak A. Pearlmutter <barak+git@pearlmutter.net>
+Baseng0815 <1shedex2@gmail.com>
+Ben Kallus <bkallus@hamilton.edu>
+Benjamin Hennion <benjamin.hennion@wanadoo.fr>
+Casey Jao <casey.jao@gmail.com>
+ChiDal <chinmay.dalal.22012001@gmail.com>
+Clement Deschamps <46494175+clemdc@users.noreply.github.com>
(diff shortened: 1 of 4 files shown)


rolandlo:
Suggestions for further improvement:
@LittleHuba suggested to mention the full list of authors (from the AUTHORS file)
@idotobi asked to add the script to generate the AUTHORS file. I simply used the script from https://gist.github.com/vladikoff/d3f5ec1f570e66fb7faff329d01012b7, listed the (core) developer team from the copyrights file and removed all email addresses that appeared to belong to one of the core developers in the list of community contributors.

LittleHuba:
Please make the version info left aligned and add a horizontal line between the upper info and the author section. Also check why the two columns seem to not be vertically aligned with each other.

rolandlo:
New version:

grafik

LittleHuba:
Looking good!
I'd vertically center the version info, if that is possible, and change the string of the link to the full list of authors to "See the full list of contributors"

Also, both repository links should be https!

rolandlo:
> Looking good!
> I'd vertically center the version info, if that is possible, and change the string of the link to the full list of authors to "See the full list of contributors"
> 
> Also, both repository links should be https!

Thanks for the feedback. I will implement it. 
One more question I have: Should we really name the years the devs have been active in the about dialog (which seems a bit ununsual) or do the names suffice? We already have the years in the linked AUTHORS file.

LittleHuba:
I'd say the names absolutely suffice. There is no obligation to even note the years anywhere and if anybody is really interested, he can grind the git history for this.

rolandlo:
> "See the full list of contributors"
Without the dots?

LittleHuba:
Yeah without the dots. Dots normally have the meaning that a dialog will be opened that requires further input of the user before the action is triggered. As we directly open the browser, no dots should be there.

rolandlo:
Now it looks like this:

grafik

rolandlo:
By the way, does @peetCreative have a real name, too?

LittleHuba:
Two final nitpicks and then I should be done 😅

For people that are not into open source or IT, "Github repository" does not ring any bells. Instead of "Github repository" I would go with "Source Code / Bug Tracker".

And generally, headlines have all non-glue-words with an starting with an upper case letter. ("Icon Designers")

> By the way, does @peetCreative have a real name, too?

Peter Klausing, what I gathered from one of his commits to a personal repository
https://github.com/peetCreative/esvr2/commit/7593dc6843ff9150b5a33d8851ed470ef9bcb9ed

LittleHuba:
Oh, and can we link to the license text of the GNU GPL v2? That would be nice as well.

rolandlo:
> Oh, and can we link to the license text of the GNU GPL v2? That would be nice as well.

This link? https://www.gnu.org/licenses/old-licenses/gpl-2.0.html

LittleHuba:
Better would be: https://raw.githubusercontent.com/xournalpp/xournalpp/master/LICENSE
Then we make sure it at least links to our version if there should for some crazy reason be discrepancies between both.

rolandlo:
One last thing that is bothering me: After "Authors", "Icon Designers", "Website" and "Source Code / Bug Tracker" we have a colon (":"), after other headlines we don't. I will remove the colons, as they seem unnecessary.

rolandlo:
That's how it looks now: 
grafik

LittleHuba:
The distance between the links and the headlines in the right column seems to be wrong in some instances. The license link is way too low and the website link should be nearer to its headline than the following one.

rolandlo:
> The distance between the links and the headlines in the right column seems to be wrong in some instances. The license link is way too low and the website link should be nearer to its headline than the following one.

I agree with that, but I haven't figured out how to fix it. The Links are put into a vertical box with the label, and even though there is 0 padding, the distance is quite big between the two box contents (label and link).

rolandlo:
Ok, fixed it. 

grafik

LittleHuba:
Okay final final nitpick ^^

Could you center the close button? Looks kind of awkward if it is all the way to the right.

rolandlo:
> Okay final final nitpick ^^
> 
> Could you center the close button? Looks kind of awkward if it is all the way to the right.

Do you like it better like this?

grafik

rolandlo:
Here is the current `diff` between the `AUTHORS` file and the `AUTHORS_raw` file created via the following command, which takes the git shortlog (summary, with email adresses) and removes leading numbers and spaces and writes it to `AUTHORS_raw`.
```
```

[diff.txt](https://github.com/xournalpp/xournalpp/files/6772377/diff.txt)

The diff should show that I have only removed authors from the `AUTHORS_raw` file that are already in the developer team.

The `AUTHORS` file should show that I didn't forget to remove any of the developers from the community contributors section of the `AUTHORS_raw` file.

LittleHuba:
Looks perfect! From my side this is good to go!

rolandlo:

idotobi:
@rolandlo Looks good to me as well :+1: And it fits my screen again :tada: 

image

The only thing that could be improved is to have separate links for "Source Code" and "Bug Tracker".
By only linking to the main section of the repo (aka. Source Code) we implicitly assume that every user knows how bugs/issues are tracked in github. Pointing also directly to the issue page would make that as obvious as possible.
However, I guess this is definitely close to overthinking the about dialog :sweat_smile: and might actually not add much value.

rolandlo:
> The only thing that could be improved is to have separate links for "Source Code" and "Bug Tracker".
> By only linking to the main section of the repo (aka. Source Code) we implicitly assume that every user knows how bugs/issues are tracked in github. Pointing also directly to the issue page would make that as obvious as possible.
> However, I guess this is definitely close to overthinking the about dialog sweat_smile and might actually not add much value.

In my opinion that would be too much. Someone who does not find the bug tracker from the main page of the repository is unlikely to submit a useful bug report or to extract useful information from existing bug reports anyway.

idotobi:
> In my opinion that would be too much. Someone who does not find the bug tracker from the main page of the repository is unlikely to submit a useful bug report or to extract useful information from existing bug reports anyway.

Sounds reasonable. From my side this is good to go then.

Technius:
LGTM.

PR #3210, opened 2021-07-06
fix comment field case and remove unused entry in copyright.txt

barak:


rolandlo:
@barak The CC-BY-SA-4.0 licence is used in https://github.com/xournalpp/xournalpp/blob/1989fb2a1225b123744b4729b513c373df35968c/copyright.txt#L47-L57

So why do you suggest to remove the license text from `copyrights.txt`?

Technius:
Looks like the removed license is CC-BY-4.0, which is a different license from CC-BY-SA-4.0. Instead, the CC-BY-4.0 text will need to be swapped with that of CC-BY-SA-4.0.

barak:
This change was mechanical, suggested by lintian which noticed a License: stanza that was not referenced by any Files: stanza. Maybe the real issue was that some Files: stanza was referencing the wrong license?

rolandlo:
> Looks like the removed license is CC-BY-4.0, which is a different license from CC-BY-SA-4.0. Instead, the CC-BY-4.0 text will need to be swapped with that of CC-BY-SA-4.0.

The CC-BY-SA-4.0 text is already contained in it. So the PR looks like it does the right thing.

rolandlo (approved):

barak:
The fact that CC-BY-SA-4.0 and CC-BY-4.0 are different (but very similar) licenses from the same organization is pretty confusing.

Issue #3104, opened 2021-05-09
Cannot open file with PDF extension (renaming to pdf works)


ngoonee:
Am on Linux, in KDE
1.1.0+dev and libgtk-3.24.29
Installed using PKGBUILD from AUR (xournalpp-git) on 17 April

When opening a file with the extension PDF (all capital letters) xournalpp opens a blank document. A blank messagebox shows with no context except the 'OK' button. Renaming the exact same document to having .pdf extension, xournalpp has no problems opening it.

I've tried with various other combinations of capital letters, it seems any capital letters in the extension fail.

bhennion:
Fixed on master

PR #3214, opened 2021-07-08
 Fix some blank messageboxes that shows no context except the 'OK' button

VivekThazhathattil:
This is a solution to one of the issues mentioned in #3104. The reason why the message box was empty was because `g_markup_escape_text()` wasn't called on the message strings before passing it to `gtk_message_dialog_set_markup()`.
Before:
image
After:
image

### DIFF
--- a/src/util/XojMsgBox.cpp
+++ b/src/util/XojMsgBox.cpp
@@ -22,7 +22,7 @@ void XojMsgBox::showErrorToUser(GtkWindow* win, const string& msg) {
 
     GtkWidget* dialog =
             gtk_message_dialog_new_with_markup(win, GTK_DIALOG_MODAL, GTK_MESSAGE_ERROR, GTK_BUTTONS_OK, nullptr);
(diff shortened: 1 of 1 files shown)


Technius (commented):
Thanks for the PR! Just one problem: `g_markup_escape_text` will allocate new strings, so they need to be freed with `g_object_unref`.

VivekThazhathattil:
@Technius Should I use `g_free` or `g_object_unref`?

Technius:
Either one works, but `g_object_unref` is less error prone since it only decreases the ref count (which will trigger a free anyway if the new ref count is 0).

VivekThazhathattil:
using `g_object_unref(formattedMsg)` crashes xournalpp for some reason :(

Technius (approved):
Ah, I was mistaken; glib strings are not reference counted. So `g_free` should be used in this context (and `g_object_ref`/`unref` should _not_ be used).

I have tested the PR and the fix appears to work correctly.

VivekThazhathattil:
Hi @Technius, will this PR be merged soon?

Technius:
Yes, thanks for the reminder.

PR #3215, opened 2021-07-08
fix working directory for Windows start menu entry

rolandlo:
This PR sets the working directory of the start menu link for Xournal++ to be the bin subfolder of the installaioin folder. This makes "../share/locale" point  to the correct locale directory.

It only works for launching Xournal++ from the start menu. It does not work from the context menu of a xopp-file or when running Xournal++ from terminal, when the present working directory is not the bin subfolder of the installation directory.

An alternative would be to use the TEXTDOMAINDIR environment variable or to use the execution path or maybe some resource path.

**Edit:** Fixed some typos


Technius (approved):
This change makes sense for start menu launches.

> It does not work from the context menu of a xopp-file or when running Xournal++ from terminal, when the present working directory is not the bin subfolder of the installation directory.

I would ideally like to see this fixed too (to allow for portable installations), but we can do that in a future PR.

Issue #2911, opened 2021-03-06
Xournal++ PDF auto-open: Not opening [filename].xopp, only [filename].pdf.xopp


personalizedrefrigerator:
**Is your feature request related to a problem? Please describe.**

![image](https://user-images.githubusercontent.com/46334387/110197598-7355c500-7e01-11eb-9e50-d74837305552.png)
Xournal++ requires PDF files to have a corresponding `.pdf.xopp` file to auto-load the Xournal++ file with the same name. This is not the default filename when saving an annotated PDF for the first time.

**Describe the solution you'd like**
I would like Xounal++ to also auto-load `[filename].xopp`, in addition to `[filename].pdf.xopp`.

**Describe alternatives you've considered**
The default export name could be changed back to `[filename].pdf.xopp`/Xournal++ could stop removing the trailing `.pdf` before appending `.xopp` when exporting (which is what I think it used to do).

**Additional context**

ngoonee:
This would be very convenient (and I think not very hard).

PR #3217, opened 2021-07-09
fix no autoloading journal for "filename.xopp" and "filename.xoj" problem

VivekThazhathattil:
This commit does the following:

If one were to enable journal autoloading (Edit -> Preferences -> Load/Save -> Autoloading Journals)  and if the xournalpp file corresponding to `<name>.pdf` were named `<name>.xopp`, `<name>.xoj`,  `<name>.pdf.xopp` or  `<name>.pdf.xoj`, then loading the pdf would autoload the xournalpp file as well. 

Previously it autoloaded only `<name>.pdf.xopp` and `<name>.pdf.xoj`. The default save file name is `<name>.xopp` for the latest build, so it would be good if the xournalpp file with that name autoloaded too.

### DIFF
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2077,7 +2077,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
 
 auto Control::loadPdf(const fs::path& filepath, int scrollToPage) -> bool {
     LoadHandler loadHandler;
-  
+
     if (settings->isAutoloadPdfXoj()) {
         fs::path f;
         Document* tmp;
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -2079,13 +2079,12 @@ auto Control::loadPdf(const fs::path& filepath, int scrollToPage) -> bool {
     LoadHandler loadHandler;
 
     if (settings->isAutoloadPdfXoj()) {
-        fs::path f;
         Document* tmp;
-        std::vector<std::string> exts = {".xopp", ".xoj", ".pdf.xopp", ".pdf.xoj"};
-        for (int i = 0; i < exts.size(); i++) {
-            f = filepath;
+        const std::vector<std::string> exts = {".xopp", ".xoj", ".pdf.xopp", ".pdf.xoj"};
+        for (const std::string& ext: exts) {
+            fs::path f = filepath;
             Util::clearExtensions(f, ".pdf");
-            f += exts[i];
+            f += ext;
             tmp = loadHandler.loadDocument(f);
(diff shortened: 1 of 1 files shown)


Technius (changes requested):
There's too much code duplication here. Can we simplify this by moving the candidate file paths into a vector and then using aloop?

src/control/Control.cpp
```diff
@@ -2079,18 +2079,17 @@ auto Control::loadPdf(const fs::path& filepath, int scrollToPage) -> bool {
...
     if (settings->isAutoloadPdfXoj()) {
-        fs::path f = filepath;
-        Util::clearExtensions(f);
-        f += ".xopp";
-        Document* tmp = loadHandler.loadDocument(f);
-
-        if (tmp == nullptr) {
+        fs::path f;
+        Document* tmp;
+        std::vector<std::string> exts = {".xopp", ".xoj", ".pdf.xopp", ".pdf.xoj"};
```
Technius, line 2084:
```suggestion
        const std::vector<std::string> exts = {".xopp", ".xoj", ".pdf.xopp", ".pdf.xoj"};
```

src/control/Control.cpp
```diff
@@ -2079,18 +2079,17 @@ auto Control::loadPdf(const fs::path& filepath, int scrollToPage) -> bool {
     LoadHandler loadHandler;
 
     if (settings->isAutoloadPdfXoj()) {
-        fs::path f = filepath;
-        Util::clearExtensions(f);
-        f += ".xopp";
-        Document* tmp = loadHandler.loadDocument(f);
-
-        if (tmp == nullptr) {
+        fs::path f;
```
Technius, line 2082: Move this into the loop since it's only used there.

src/control/Control.cpp
```diff
@@ -2079,18 +2079,17 @@ auto Control::loadPdf(const fs::path& filepath, int scrollToPage) -> bool {
...
+        Document* tmp;
+        std::vector<std::string> exts = {".xopp", ".xoj", ".pdf.xopp", ".pdf.xoj"};
+        for (int i = 0; i < exts.size(); i++) {
```
Technius, line 2085: `i` is only used for `exts[i]`, so prefer the range based for loop:
```
for (const std::string& ext: exts) {
```

Technius (commented):
Nice work, this is much better. I just have some minor style suggestions (see below).

VivekThazhathattil:
I think it fixes #2911.

Technius (approved):
Thanks for the find. I've confirmed that this PR fixes the issue and works as intended.

PR #3221, opened 2021-07-11
Replaced hashmap to trivial key type by an std::vector.

Febbe:
Remove std::unordered_map to reduce caching issues / increase performance


Febbe:
@rolandlo can you check this out?

rolandlo:
Ok, I will build it on Azure Pipelines and see, if it makes a difference on BigSur.

Febbe:
I think, this will slow down zooming. 
But I can replace the `std::unordered_map` with a `std::vector`, which will even improve zooming speed. 
When this fixes the drawing issues on Mac, I try to fix it before the weekend.

rolandlo:
Can you rebase on current master, please?

rolandlo:
Something is odd, I can't write with my pen any more (on Ubuntu). It moves the pen around like the hand tool. Mouse works fine.

Febbe:
I did not changed that

rolandlo:
Ok, it was something in my input system settings. Don't quite understand why the installed Xournal++ version ran normally with it.

rolandlo:
I have tested it on BigSur (after building your PR on BigSur). The lag is still roughly the same. Maybe the lag comes from the rendering, not from the layout computation.

Febbe:
The layout calculation is part of the rendering.

Febbe:
But it seams, that this isn't the cause or my attempt is not faster than looking into the unordered_map

Febbe:
Can you retry, this should be faster anyway.

rolandlo:
> @rolandlo can you retry, this should be faster anyway now?

I got a crash after a couple of seconds (adding some pages and switching to 2-column layout)

```
Date: Fri Jul 16 18:38:51 2021
Error: signal 11
[bt]: (0) build/src/xournalpp(+0x216f2b) [0x5563d97fcf2b]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x41040) [0x7f5a702f7040]
[bt]: (2) build/src/xournalpp(_ZNK11XojPageView13containsPointEiib+0x60) [0x5563d975ce10]
[bt]: (3) build/src/xournalpp(_ZN6Layout13getPageViewAtEii+0x106) [0x5563d9755e96]
[bt]: (4) build/src/xournalpp(_ZN17MouseInputHandler15setPressedStateERK10InputEvent+0x13) [0x5563d9797c63]
[bt]: (5) build/src/xournalpp(_ZN17MouseInputHandler10handleImplERK10InputEvent+0x2b) [0x5563d9797d2b]
[bt]: (6) build/src/xournalpp(_ZN12InputContext6handleEP9_GdkEvent+0x4ac) [0x5563d9796edc]
[bt]: (7) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3e94a8) [0x7f5a714984a8]
[bt]: (8) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x1294) [0x7f5a70bfa724]
[bt]: (9) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f5a70bfa893]
[bt]: (10) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3b05fb) [0x7f5a7145f5fb]
[bt]: (11) /lib/x86_64-linux-gnu/libgtk-3.so.0(gtk_main_do_event+0xd3f) [0x7f5a7130209f]
[bt]: (12) /lib/x86_64-linux-gnu/libgdk-3.so.0(+0x39733) [0x7f5a70fe3733]
[bt]: (13) /lib/x86_64-linux-gnu/libgdk-3.so.0(+0x70e36) [0x7f5a7101ae36]
[bt]: (14) /lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_context_dispatch+0x25b) [0x7f5a718b38eb]
[bt]: (15) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0xa8d28) [0x7f5a71906d28]
[bt]: (16) /lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_context_iteration+0x33) [0x7f5a718b1023]
[bt]: (17) /lib/x86_64-linux-gnu/libgio-2.0.so.0(g_application_run+0x1cd) [0x7f5a70d0027d]
[bt]: (18) build/src/xournalpp(_ZN11XournalMain3runEiPPc+0x4d2) [0x5563d96cae82]
[bt]: (19) /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xd5) [0x7f5a702de565]
[bt]: (20) build/src/xournalpp(_start+0x2e) [0x5563d96a055e]
```

Febbe:
should be fixed now

src/gui/LayoutMapper.h
```diff
@@ -21,23 +21,20 @@
...
 #include "XournalType.h"
 
-namespace std {
-template <>
-class hash<std::pair<size_t, size_t>> {
-public:
-    size_t operator()(std::pair<size_t, size_t> const& data) const {
-        size_t seed = 0;
-        boost_c::hash_combine(seed, data.first);
-        boost_c::hash_combine(seed, data.second);
```
rolandlo, line 31: Since these were the only two occurences of `hash_combine` it would make sense to remove `hashcombine.h` and the license reference in `copyright.txt` (lines 89-93)

src/gui/LayoutMapper.h
```diff
@@ -99,33 +96,32 @@ class LayoutMapper {
...
 
-private:
+public:
     struct internal_data {
         size_t cols = 0;
         size_t rows = 0;
         size_t actualPages = 0;
 
-        int offset = 0;
+        uint32_t offset = 0;
```
rolandlo, line 106: Now `getFirstPageOffset()` should have return type `uint32_t` as well or be removed (it isn't used anywhere).

src/gui/LayoutMapper.h
Febbe: is it ok to just remove the entry or do I have to execute the python script?

rolandlo: I think it is OK. Running the python script would only have told you that the `copyright.txt` should be updated. We can update the `latest_commit_hash` another time.

src/gui/LayoutMapper.h
rolandlo: Now it doesn't compile any more. Maybe some missing include (#include <cstdint> I guess)?
```  
./src/gui/LayoutMapper.h:99:9: error: 'uint32_t' does not name a type; did you mean 'wint_t'?
   99 |         uint32_t offset = 0;
```

rolandlo:
Except of the missing header file it looks all fine to me. The code has become cleaner,. I can't tell though, how big the performance gain is.

Febbe:
> Except of the missing header file it looks all fine to me. The code has become cleaner,. I can't tell though, how big the performance gain is.

It is small, but visible via perf ~0,5% perf gain. In contrast to this, my other attempts had performance gains of (between) 7% (and 21% on X11).

Febbe:

PR #3234, opened 2021-07-16
Optimized the calculation of stroke sizes

Febbe:
@rolandlo pls test

### DIFF
--- a/src/model/Stroke.cpp
+++ b/src/model/Stroke.cpp
@@ -8,6 +8,34 @@
 
 #include "i18n.h"
 
+template <typename Float>
+constexpr void updateBounds(Float& x, Float& y, Float& width, Float& height, Rectangle<Float>& snap, Point const& p,
+                            double half_width) {
+    {
+        Float x2 = x + width;
+        Float y2 = y + height;
+
+        x = std::min(x, p.x - half_width);
+        y = std::min(y, p.y - half_width);
+        x2 = std::max(x2, p.x + half_width);
+        y2 = std::max(y2, p.y + half_width);
+        width = x2 - x;
+        height = y2 - y;
+    }
+    {
+        Float snapx2 = snap.x + snap.width;
+        Float snapy2 = snap.y + snap.height;
+
+        snap.x = std::min(snap.x, p.x);
+        snap.y = std::min(snap.y, p.y);
+        snapx2 = std::max(snapx2, p.x);
+        snapy2 = std::max(snapy2, p.y);
+        snap.width = snapx2 - snap.x;
+        snap.height = snapy2 - snap.y;
+    }
+}
+
+
 Stroke::Stroke(): AudioElement(ELEMENT_STROKE) {}
 
 Stroke::~Stroke() = default;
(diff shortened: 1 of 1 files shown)


src/model/Stroke.cpp
```diff
@@ -359,42 +395,32 @@ void Stroke::calcSize() const {
...
-        maxY = std::max(maxY, p.y + halfThick);
-
         minSnapX = std::min(minSnapX, p.x);
         minSnapY = std::min(minSnapY, p.y);
-
         maxSnapX = std::max(maxSnapX, p.x);
         maxSnapY = std::max(maxSnapY, p.y);
     }
 
+    halfThick = points[0].z != Point::NO_PRESSURE ? halfThick / 2.0 : this->width / 2.0;
```
rolandlo, line 413: This looks flawed to me. `halfThick` was set to `0.0`, so `halfThick / 2.0` will simply be `0.0`. Was your intention to take `halfThick` to be the maximal `p.z` divided by 2?
This will potentially enlarge the `bounds` rectangle (compared to the current implementation) when pressure is non-constant.

Also do you assume that either all points of a stroke have no pressure or all points have pressure?

Febbe: > Was your intention to take halfThick to be the maximal p.z divided by 2
Yes, thanks for the catch.

> This will potentially enlarge the bounds rectangle (compared to the current implementation) when pressure is non-constant.
I am aware of that, but I think this is not such a big problem;
 - The thickness will have a small range
 - The previous attempt also was an approximation.
 - The rectangle is mostly used to shorten hard computations

src/model/Stroke.cpp
```diff
@@ -359,42 +395,32 @@ void Stroke::calcSize() const {
...
     double maxSnapX = DBL_MIN;
     double minSnapY = DBL_MAX;
     double maxSnapY = DBL_MIN;
 
-    bool hasPressure = points[0].z != Point::NO_PRESSURE;
-    double halfThick = this->width / 2.0;  //  accommodate for pen width
+    auto halfThick = 0.0;
 
+    //#pragma omp parralel
     for (auto&& p: points) {
```
Febbe, line 406:
```suggestion
    for (auto&& p: points) {
        halfThick = std::max(halfThick, p.z)
```

src/model/Stroke.cpp
Febbe: > Also do you assume that either all points of a stroke have no pressure or all points have pressure?

This hasn't changed, so that assumption was made before.

Febbe: Merging tomorrow, if no objections.

Febbe:

src/model/Stroke.cpp
```diff
@@ -137,16 +165,23 @@ void Stroke::setLastPoint(const Point& p) {
 
 void Stroke::addPoint(const Point& p) {
     this->points.emplace_back(p);
-    this->sizeCalculated = false;
+    updateBounds(Element::x, Element::y, Element::width, Element::height, Element::snappedBounds, p,
+                 hasPressure() ? p.z / 2.0 : this->width);
```
rolandlo, line 169:
```suggestion
I think this should be
                 hasPressure() ? p.z / 2.0 : this->width / 2.0);
```

src/model/Stroke.cpp
```diff
@@ -137,16 +165,23 @@ void Stroke::setLastPoint(const Point& p) {
...
 }
 
 auto Stroke::getPointCount() const -> int { return this->points.size(); }
 
 auto Stroke::getPointVector() const -> std::vector<Point> const& { return points; }
 
-void Stroke::deletePointsFrom(int index) { points.resize(std::min(size_t(index), points.size())); }
+void Stroke::deletePointsFrom(int index) {
+    points.resize(std::min(size_t(index), points.size()));
```
rolandlo, line 177: In principle in case `index > points.size()` it is not necessary to invalidate the cache. Not sure that is relevant though.

Febbe: Yes, that's right, but using an index, larger than `points.size() - 1` can be assumed to be a programming error.
So you are also right with point 2:
> Not sure that is relevant though.

src/model/Stroke.cpp
Febbe: Good catch.

src/model/Stroke.cpp
LittleHuba: > Also do you assume that either all points of a stroke have no pressure or all points have pressure?

This assumption should be true currently. The only case where no pressure is available is if you use a tool that has no pressure sensitivity or where GDK+ does not supply it. If GDK+ does not supply it, this is as far as I have seen, true for all events. Though there is nothing that ensures this, so if this should ever change, it will be a hard-to-find bug.

bhennion: [This line](https://github.com/xournalpp/xournalpp/blob/97c20f62c5a5c16ce6dbb21ffd19240f4fe55961/src/control/tools/StrokeHandler.cpp#L121) ensures that the pressure is never set if the first event"s pressure is not available.

PR #3235, opened 2021-07-16
Remove X11 callback to flush the X11 context, which happens way too often

Febbe:
Also, this function call should be obsolete, since it will be called by gtk after each drawing cycle.

### DIFF
--- a/src/gui/XournalppCursor.cpp
+++ b/src/gui/XournalppCursor.cpp
@@ -306,8 +306,9 @@ void XournalppCursor::updateCursor() {
         }
         gtk_widget_set_sensitive(xournal->getWidget(), !this->busy);
     }
-
-    gdk_display_sync(gdk_display_get_default());
+    // Performance removal!
+    // If this line is ever required, make sure this function is never called, when the cursor is not set.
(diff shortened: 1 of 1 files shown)


Technius:
Using this snippet:
```c++
    const auto start = g_get_real_time();
    gdk_display_sync(gdk_display_get_default());
    const auto end = g_get_real_time();
    g_message("Elapsed: %ld us", end - start);
```
On my computer, it seems like the performance overhead is between 10-20 microseconds if no cursor change occurs and 500-1500 microseconds if there is a cursor change. So the overhead is quite negligible, although I think we should remove this call anyways to keep the code cleaner.

Febbe:
@Technius on my machine, it had a huge impact:
image

Febbe:
Shall I keep the comment?

Febbe:

Technius:
> on my machine, it had a huge impact, it had 3/5 of the input handling time:

Wow, that's crazy.

> Shall I keep the comment?

It wouldn't hurt, although I would also add a reference to this PR.

> Merging tomorrow, if no objections.

Let's save this to after the release since we don't have time to test it. We can always publish a minor bugfix release afterwards.

Issue #3224, opened 2021-07-12
regression when window switching/rescaling


suisseWalter:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Ubuntu 20.04 LTS
 - (Linux only) Desktop environment: I3-X11
 - Which version of libgtk do you use 3.22.30 /(3.24.20 when built from source)
 - Version of Xournal++: [e.g. 1.0.16, specific commit, specific branch]
     1a60605e(nightly build 11 july 2021)
 - Installation method: appimage

**Describe the bug**
using I3 as my WM I have noticed some massive lags with switching to and from the latest version of nightly. 
It takes about a second to load the window when switching to it. the same goes if it's re-scaled(closing another window beside it).
I can also produce it by force-closing the about menu(clsing it by keyboard comand, not be close button), but not while opening it. 
note moving it around the same workspace(without changing the size) works without a problem. 

I experience similar issues while switching to and from graphics heavy games, that aren't really made for re-scaling their windows.

this behavior was introduced between nightly 20210615 and the latest one(20210711)

**To Reproduce**
1) open latest nightly in I3
2a) remove windows on the same workspace.
2b) open the about menu and then close it again.
2c) switch to the workspace xournalpp is on from a different one.
3) see xournalpp behaving slowly. and leaving black bars for a couple of seconds.

**Expected behavior**
Xournalpp should immediately rescale windows and update it's content. 
the corect behaviour is already happening in stable as well as in nightly 20210615(and before)

**Screenshots of Problem**

**Additional context**
I would be happy to try all the nightly build between 6 of june and today to find the one that introduced the problem, should that be needed.

rolandlo:
@suisseWalter We haven't changed much which might explain the regression you observe from June 15, 2021 and July 11, 2021. Commit 07f2aa685db21540b46b1dee3652a2fea7e8e821 is the only one which comes to mind on first sight, and even that commit doesn't look likely to me. It is possible that some dependency update created the regression. You would find out if you build from source as described in the [wiki](https://github.com/xournalpp/xournalpp/blob/master/readme/LinuxBuild.md). Let me know if you need help with it.

Are you sure you can exclude a system update on your side as the problem?

suisseWalter:
I can have both appimages running in parallel and observe it(or just the new one). so there shouldn't be a problem with dependencies. I can try to build from source. did it in the past so should be possible. 
I was also looking at 6354a33 but is glade only used for creating files and not part if the final package?

[...]

PR #3236, opened 2021-07-16
Performance gdkeventguard

Febbe:
The previous PR'S are still in this PR, so I had better profiling information.

### DIFF
--- a/src/gui/inputdevices/InputEvents.cpp
+++ b/src/gui/inputdevices/InputEvents.cpp
@@ -4,6 +4,8 @@
 
 #include "InputEvents.h"
 
+#include "control/settings/Settings.h"
+
 auto InputEvents::translateEventType(GdkEventType type) -> InputEventType {
     switch (type) {
         case GDK_MOTION_NOTIFY:
--- a/src/gui/inputdevices/InputEvents.h
+++ b/src/gui/inputdevices/InputEvents.h
@@ -11,11 +11,14 @@
 
 #pragma once
 
+#include <memory>
+
 #include <gdk/gdk.h>
 #include <glib.h>
 
-#include "control/settings/Settings.h"
 #include "model/Point.h"
+
+class Settings;
 
 enum InputEventType {
     UNKNOWN,
@@ -46,48 +49,25 @@ enum InputDeviceClass {
 
 struct GdkEventGuard {
     static inline GdkEvent* safeRef(GdkEvent* source) { return gdk_event_copy(source); }
-    static inline GdkEvent* safeMove(GdkEvent*& source) {
-        auto ret = source;
-        source = nullptr;
-        return ret;
-    }
-    static inline void safeDelete(GdkEvent* source) {
-        if (source != nullptr) {
-            gdk_event_free(source);
-        }
-    }
 
     GdkEventGuard() = default;
-    [[maybe_unused]] explicit GdkEventGuard(GdkEvent* source): event(safeRef(source)) {}
-    GdkEventGuard(GdkEventGuard const& other): event(safeRef(other.event)) {}
-    GdkEventGuard(GdkEventGuard&& other): event(safeMove(other.event)) {}
+
+    [[maybe_unused]] explicit GdkEventGuard(GdkEvent* source): event(safeRef(source), &gdk_event_free) {}
 
     GdkEventGuard& operator=(GdkEvent* source) {
-        safeDelete(event);
-        event = safeRef(source);
-        return *this;
-    }
-    GdkEventGuard& operator=(GdkEventGuard const& other) {
-        safeDelete(event);
-        event = safeRef(other.event);
-        return *this;
-    }
-    GdkEventGuard& operator=(GdkEventGuard&& other) {
-        safeDelete(event);
-        event = safeMove(other.event);
+        event = {safeRef(source), &gdk_event_free};
         return *this;
     }
 
-    operator GdkEvent*() const { return event; }
+    operator GdkEvent*() const { return event.get(); }
 
-    ~GdkEventGuard() { safeDelete(event); }
-
-    GdkEvent* event{};
+    // it's more performant to manage the GdkEvent over C++ than over gdk
+    // Since the gdk_copy is extreme expansive
+    std::shared_ptr<GdkEvent> event{};
 };
 
-
 struct InputEvent final {
-    /*explicit(false)*/ explicit operator bool() const { return sourceEvent.event; }
+    /*explicit(false)*/ explicit operator bool() const { return !!sourceEvent.event; }
 
     GdkEventGuard sourceEvent;
 


Febbe:
@rolandlo you may want to test the performance now on macOs

Technius:
The next release is very close, so let's delay the merge of such nontrivial refactoring to after the release.

rolandlo:
I have let build this PR on Azure Pipelines and compared it to the latest nightly release. On my MacBook there is still significant lag in both builds when using a toolbar (I used the portrait toolbar). Using my Wacom Intuos Pro M tablet I can't write normally and big circles I draw (with normal speed) become very edgy (maybe 8-10 edges for one circle). Without toolbar handwriting looks acceptable on both versions and circles I draw look more like circles (around 20-24 edges).
I don't observe a performance gain from this PR on my MacBook. Comparing results it looks more like the opposite, although the decisive factor is having the toolbar displayed or not. There may be Mac users that will report otherwise (for one user with an M1 chip the toolbar didn't make a difference), but at least this PR doesn't solve the lag issue for all Mac users.

I agree with @Technius that we should delay the merge of this PR.

Febbe:
Damn, than I need you to profile x++ on macOs. 
On X11 I have a large performance gain :)

rolandlo:
> Damn, than I need you to profile x++ on macOs.
> On X11 I have a large performance gain :)

Yes, there is seems to be no way around profiling Xournal++ on MacOS BigSur. On Linux when I draw big circles, they look completely smooth. No edges visible at all. Looking at the `gunzip`-ed .xopp-file a circle-like shape I drew on Linux consists of about 300 points. On MacOS BigSur it's about 30 points, on MacOS Catalina about 85 points. Since I use different laptops on Linux and MacOS it's not directly comparable, but still gives a hint.

suisseWalter:
it seams like this branch causes some crashes for me. (X11+I3,ubuntu20.04)
one of the errorlogs created is attached bellow. 
the crashes can happen at any point. but I can provoke them by moving the window around in I3 and it most often crashes as soon as the window is next to the left edge of my display. But I was also able to cause the crash by just writing on it. 
I compiled all 4 different commits in this PR. and it's only the last one that causes the problem, so the gtkeventguard stuff. 
(I just compiled the febbe/xournalpp fork, I hope this is correct. but it seams like this is the only difference to master) 
the console output before crashing was the following:
```
xournalpp: /home/username/Documents/sandbox/xournalbug/xournalpp/src/gui/LayoutMapper.cpp:171: std::optional<long unsigned int> LayoutMapper::at(const GridPosition&) const: Assertion `index < rasterToPage.size()' failed.

** (xournalpp:101914): WARNING **: 20:02:10.230: [Crash Handler] Crashed with signal 6

** (xournalpp:101914): WARNING **: 20:02:10.231: [Crash Handler] Wrote crash log to: /home/username/.cache/xournalpp/errorlogs/errorlog.20210720-200210.log

** (xournalpp:101914): WARNING **: 20:02:10.430: Trying to emergency save the current open document…

```

[errorlog.20210720-191819.log](https://github.com/xournalpp/xournalpp/files/6850170/errorlog.20210720-191819.log)

the reason for me compiling this fork was to figure out if it may contain a alternative fix for #3224 (was curious about the X11 callback part), but it does not.

Febbe:
@suisseWalter Thanks for testing, the crash happens, because I assumed, that the structure I refactored is not called with an illegal argument. Just like calling the vectors subscription operator with an out of bounds index.

But, when it does not crash, it also hadn't a performance gain?

Febbe:
> I compiled all 4 different commits in this PR. and it's only the last one that causes the problem, so the gtkeventguard stuff.

It crashes because bac6746 not the gdk_event_guard.

Febbe:
> 
> 
> > I compiled all 4 different commits in this PR. and it's only the last one that causes the problem, so the gtkeventguard stuff.
> 
> It crashes because [bac6746](https://github.com/xournalpp/xournalpp/commit/bac6746ed86f5ff8613bab91af4d7e2cb9121e7e) not the gdk_event_guard.

It crashes here: 
https://github.com/xournalpp/xournalpp/commit/bac6746ed86f5ff8613bab91af4d7e2cb9121e7e#diff-e21b577a460a5ce8a9cab36f500fd33d7c17414dcc50a06d0bbc73c898596405R171

suisseWalter:
> But, when it does not crash, it also hadn't a performance gain?

I'm running on linux, with a rather powerfull laptop. so even if there would be a expected performance increase it probably wouldn't be significant and therefore it wouldn't be noticable. but no, I'm not noticing a performance increase. where should I notice it? as I wasn't testing it in use, for obvious reasons, just switching it between workspaces etc. if you would like to, I could compile it to a appimage and then test it on a old laptop(year2012) and see if I can spot a visible difference. 

> It crashes because bac6746 not the gdk_event_guard.

weird, thought I had compiled with that and it didn't cause the problem. but if you found the problem I was probably wrong.

rolandlo:
@Febbe So what to do about the crash noticed by @suisseWalter ?
The performance issue on MacOS is partly fixed now (in #2729). Even on BigSur drawing is now fluent if using the empty toolbar, or the minimal left toolbar from PR #3267. The problems have to do with bad performance of the Quartz backend used on MacOS. Rendering Cairo surfaces is expensive on MacOS, and the cost depends pretty much on the size of  the rectangles that are renderd. It will apparently improve considerably with Gtk 4, but for now we can't expect the same performance on MacOS like on Linux with respect to the rendering.

Your changes make sense to me. If you can avoid the crash and rebase on current master, I can test once more on MacOS, Windows and Linux. I'm not sure how to measure the performance exactly though. How did you measure the performance gain mentioned in https://github.com/xournalpp/xournalpp/pull/3221#issuecomment-881739943?

Febbe:
I am on vacation in Sweden, since we are here with a Camper, I don't have much power for my laptop and only mobile internet.

Febbe:
@rolandlo, I would change the assert to an regular if

Febbe:
Done. I would merge this into the new release branch

LittleHuba:
> Done. I would merge this into the new release branch

Please merge it into a feature-branch, so we can take another look into that crash noticed by @suisseWalter

Febbe:
The crash is fixed already

LittleHuba:
Then we can merge this directly with `master`.

Febbe:
This is rather a bugfix, than a refactoring

LittleHuba:
But is it a bugfix that requires the release of a hotfix?

Febbe:
Here is the question how fine-grained our release cycles should be. I would release it with 1.1.x and not 1.2.x a hot fix is something like 1.1.1.x (for me) since it "fixes" a critical in a release

LittleHuba:
Let's move this discussion to Gitter.

rolandlo:
I have trouble checking this PR out. When I do
```
 git fetch upstream pull/3236/head:febbe
git checkout febbe
git log
```
I don't see the 4 commits introduced in this PR, but only
```
8430f45516181f1fdbb106be0972c1ee1291ee54 (grafted, HEAD -> febbe, Febbe/performance_gdkeventguard) Optimize GdkEventGuard
```
Why does that happen? Is that because @Febbe's [master branch](https://github.com/Febbe/xournalpp) already contains changes that are not in the main repo master?

Febbe:
have you tried `gitk --all`?

rolandlo:
> have you tried `gitk --all`?

I didn't know about that tool. That's how the graph looks like:
grafik

Febbe:
Ah, you checked out my commit, without feching the history. So the commit is "unrelated". But it contains all previous commits.

PR #2839, opened 2021-02-20
Update changelog to prepare for 1.1.0

Technius:
This currently includes the changes made since the last changelog update. Please add suggestions if I missed anything / got anything wrong.

We still need to check against some of the older commits since November 2019.

PR #3241, opened 2021-07-18
Update changelog to prepare for 1.1.0

rolandlo:
This is  #2839 squashed and rebased. While fixing the merge conflict I added the (second last) line about fixing the BigSur startup crash and removing the integration into the Mac menu bar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,80 @@
 
 ## 1.1.0 / Nightly (Unreleased)
 
+This is a new major version of Xournal++ with many new features, improvements,
+and bug fixes thanks to over one year's worth of contributions from the
+community.
+
 * **Breaking changes**:
     * Xournal++ now follows the [XDG Base Directory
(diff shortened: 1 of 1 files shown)


rolandlo:
@Technius In my opinion this PR is ready to be merged. Any objections?

Technius:
LGTM.
* If MSYS updated in time, I would add a line referring to the GTK dependency update.
* I would add a line to "Breaking Changes" explaining that the old input system has been removed, which may affect some users.

rolandlo:
> LGTM.
> 
>     * If MSYS updated in time, I would add a line referring to the GTK dependency update.
Unfortunately, MSYS2 on Azure Pipelines only updated from `mingw-w64-x86_64-gtk3-3.24.29-1` to `mingw-w64-x86_64-gtk3-3.24.29-2` (3 days ago). That won't contain the improved input handling on Windows.
> 
>     * I would add a line to "Breaking Changes" explaining that the old input system has been removed, which may affect some users.
Ok, I will amend it to the current commit.

rolandlo:
Ok, done!

PR #3263, opened 2021-07-24
Try to find out what improved macOS behaviour in #2729

bhennion:
This follows up on the discussion started in [this post](https://github.com/xournalpp/xournalpp/pull/2729#issuecomment-885898684) about how something in #2729 improves the fluidity of strokes under macOS. 

@rolandlo Could you try out those two commits and tell me if the improvement is still there?
It should work with the second commit, but I also hope it works with the first...

(with the second commit, some features will not work. Some functions are just dummies so that it compiles)

### DIFF
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -120,6 +120,14 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
 
     stroke->addPoint(this->hasPressure ? point : Point(point.x, point.y));
 
+    double width = stroke->getWidth();
+
+    assert(stroke->getPointCount() >= 2);
+    const Point& prevPoint(stroke->getPoint(stroke->getPointCount() - 2));
+
+    Range rg(prevPoint.x, prevPoint.y);
+    rg.addPoint(point.x, point.y);
+
     if ((stroke->getFill() != -1 || stroke->getLineStyle().hasDashes()) &&
         !(stroke->getFill() != -1 && stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER)) {
         // Clear surface
@@ -132,27 +140,26 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
         cairo_fill(crMask);
 
         view.drawStroke(crMask, stroke, 0, 1, true, true);
+
+        const Point& firstPoint = stroke->getPointVector().front();
+        rg.addPoint(firstPoint.x, firstPoint.y);
     } else {
-        if (auto const pointCount = stroke->getPointCount(); pointCount > 1) {
-            Point prevPoint(stroke->getPoint(pointCount - 2));
+        Stroke lastSegment;
 
-            Stroke lastSegment;
+        lastSegment.addPoint(prevPoint);
+        lastSegment.addPoint(point);
+        lastSegment.setWidth(width);
 
-            lastSegment.addPoint(prevPoint);
-            lastSegment.addPoint(point);
-            lastSegment.setWidth(stroke->getWidth());
+        cairo_set_operator(crMask, CAIRO_OPERATOR_OVER);
+        cairo_set_source_rgba(crMask, 1, 1, 1, 1);
 
-            cairo_set_operator(crMask, CAIRO_OPERATOR_OVER);
-            cairo_set_source_rgba(crMask, 1, 1, 1, 1);
-
-            view.drawStroke(crMask, &lastSegment, 0, 1, false);
-        }
+        view.drawStroke(crMask, &lastSegment, 0, 1, false);
     }
 
-    const double w = stroke->getWidth();
+    width = prevPoint.z != Point::NO_PRESSURE ? prevPoint.z : width;
 
-    this->redrawable->repaintRect(stroke->getX() - w, stroke->getY() - w, stroke->getElementWidth() + 2 * w,
-                                  stroke->getElementHeight() + 2 * w);
+    this->redrawable->repaintRect(rg.getX() - 0.5 * width, rg.getY() - 0.5 * width, rg.getWidth() + width,
+                                  rg.getHeight() + width);
 }
 
 void StrokeHandler::onMotionCancelEvent() {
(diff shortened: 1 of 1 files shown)


bhennion:
Oh, btw, the CI test are failing. That's expected with this half baked version. It's just to reduce the haystack in which to look for the needle.

rolandlo:
I've just tested. The improvement is seen in Test2, but not in Test1.

bhennion:
> I've just tested. The improvement is seen in Test2, but not in Test1.

Thanks! It's quite a mystery... I really have not idea what change I made could be responsible.
I'll try and give you other commits to test.

rolandlo:
The improvement is seen in Test3 as well. The haystack is getting smaller. :smile:

bhennion:
Ok, here are commits to bissect. I think they all build (but I only checked some).
I guess the improvement should appear either with f4fa61 or 41b006f

rolandlo:
Results (will get updated):
- In general, the LoadHandler test failes (can be seen from Azure Pipelines)
- commit `StrokeHandler::drawSegmentTo` contains the performance improvement.  I noticed two bugs: the first point is always without pressure. After the stroke is finished, the line gets thicker.

rolandlo:
So in summary the performance improvement must stem from commit `StrokeHandler::drawSegmentTo` 41b006f

rolandlo:
> So in summary the performance improvement must stem from commit `StrokeHandler::drawSegmentTo` [41b006f](https://github.com/xournalpp/xournalpp/commit/41b006fb8abac953a199e68a9f9b0158bd9ab073)

(at least when drawing with a pressure-sensitive stylus)

bhennion:
> Results (will get updated):
> 
> * In general, the LoadHandler test fails (can be seen from Azure Pipelines)
> * commit `StrokeHandler::drawSegmentTo` contains the performance improvement.  I noticed two bugs: the first point is always without pressure. After the stroke is finished, the line gets thicker.
> * commit `StrokeView` does not contain the improvement.
> * commit `StrokeHandler::paintTo` does not contain the improvement.

Thanks! 
Would you mind also trying to apply only the commit `Stroke::drawSegmentTo` (and not the other 4)? It should still build.

rolandlo:
> Would you mind also trying to apply only the commit `Stroke::drawSegmentTo` (and not the other 4)? It should still build.

Yes, that commit alone also gives the performance improvement.

bhennion:
Ok, I think I got it. It's essentially the changes I said "I did not remember why I did that" about earlier. It's about reducing as much as you can the size of the rectangle feed to `this->redrawable->repaintRect`.

Could you check if it also works with `Fill` on or with the highlighter?

rolandlo:
> Ok, I think I got it. It's essentially the changes I said "I did not remember why I did that" about earlier. It's about reducing as much as you can the size of the rectangle feed to `this->redrawable->repaintRect`.
> 
> Could you check if it also works with `Fill` on or with the highlighter?

I will check. The improvement is also seen without the two `this->redrawable->repaintRect` statements though.

bhennion:
> I will check. The improvement is also seen without the two `this->redrawable->repaintRect` statements though.

I'm confused here. On linux, the ongoing stroke is not displayed at all if I remove this (or those, depending on the patch) `this->redrawable->repaintRect`. You can see the stroke without it under MacOS? Or maybe you meant something else...

rolandlo:
> > I will check. The improvement is also seen without the two `this->redrawable->repaintRect` statements though.
> 
> I'm confused here. On linux, the ongoing stroke is not displayed at all if I remove this (or those, depending on the patch) `this->redrawable->repaintRect`. You can see the stroke without it under MacOS? Or maybe you meant something else...

You are right. Seems I forgot to save the file when commenting out before recompiling on MacOS (I'm used that files get saved automatically on Linux). I agree with the conclusion now. It seems painting with Cairo is quite time-costly on MacOS and that's why the lag appeared. 

I think we can make the rectangle even a bit smaller by replacing `width` by `0.5*width` (and `2*width` with `width`) in the rectangle coordinates. 

There is a little issue with filled highlighter strokes, but there is also an issue with that in master. These highlighter strokes are only filled when completed (in contrast to strokes). And it may take a little longer until they are filled after completion in comparison with master. But filled highlighter strokes is not something I would normally use.

bhennion:
> It seems painting with Cairo is quite time-costly on MacOS and that's why the lag appeared.
This is pretty weird. Do you have a way of knowing whether Cairo uses the GPU (in general on MacOS and specifically with Xournal++)? This would explain the drastic difference in performance.

> I think we can make the rectangle even a bit smaller by replacing `width` by `0.5*width` (and `2*width` with `width`) in the rectangle coordinates.

Done

> 
> There is a little issue with filled highlighter strokes, but there is also an issue with that in master. These highlighter strokes are only filled when completed (in contrast to strokes). And it may take a little longer until they are filled after completion in comparison with master. But filled highlighter strokes is not something I would normally use.

Added a full stroke repaint after a filled stroke with Highlighter is completed. It deals with the delay, but the filling still only happens once the stroke is complete.

rolandlo:
> > It seems painting with Cairo is quite time-costly on MacOS and that's why the lag appeared.
> 
> This is pretty weird. Do you have a way of knowing whether Cairo uses the GPU (in general on MacOS and specifically with Xournal++)? This would explain the drastic difference in performance.
> 
> > I think we can make the rectangle even a bit smaller by replacing `width` by `0.5*width` (and `2*width` with `width`) in the rectangle coordinates.
> 
> Done
> 
> > There is a little issue with filled highlighter strokes, but there is also an issue with that in master. These highlighter strokes are only filled when completed (in contrast to strokes). And it may take a little longer until they are filled after completion in comparison with master. But filled highlighter strokes is not something I would normally use.
> 
> Added a full stroke repaint after a filled stroke with Highlighter is completed. It deals with the delay, but the filling still only happens once the stroke is complete.

Ok, makes sense to me. 
I will now build on MacOS BigSur. I would expect that when the toolbar is not visible, it should be reasonably fluent as well.

rolandlo:
As expected, the performance on MacOS BigSur has also improved significantly. On BigSur there is still lag when the toolbar is shown, but that is a different problem. When the toolbar is not shown, drawing works reasonably smooth.

To me, this PR looks ready. Did you still want to change something?

bhennion:
> To me, this PR looks ready. Did you still want to change something?

No, I'm happy with it! Thanks for all the testing!

rolandlo (approved):
I will merge in 24 hours if no objections are raised.
Thanks @bhennion for this valuable fix and the patience to get down to the very root of the problem.

Issue #3242, opened 2021-07-18
Big Icon toolbar icons create lag in application


idotobi:
**Is your feature request related to a problem? Please describe.**

This is basically a reminder to check why big toolbar icon sizes affect the lag of the application when drawing.
It could be that this is a symptom of unnecessary redrawing of toolbar icons which could be avoided.

**Describe the solution you'd like**
Research how toolbar icons are affecting lag.

**Describe alternatives you've considered**
n.a.

**Additional context**
src: https://github.com/xournalpp/xournalpp/pull/3230#issuecomment-881064542
Internal Gitter context: https://gitter.im/xournalpp/dev-internal?at=60f15d1c7b5a415e65ebe157

rolandlo:
I suggest adding a minimal toolbar (only one column on the left used) to the list of toolbars. For me on MacOS BigSur after #3263 is merged this more or less solves the lag issue. A minimal toolbar could be useful for other users as well,

Issue #2139, opened 2020-07-29
Xournal++ on PinePhone


rolandlo:
The [PinePhone](https://www.pine64.org/pinephone/) is a new open source smartphone developed by [Pine64](https://www.pine64.org/) running a mainline linux kernel. You can run various linux distributions on the PinePhone. So you can have linux apps running on your phone, which is pretty cool. Moreover the phone is quite affordable (149 USD). I recently bought one (UBports Community Edition) and was curious to find out if Xournal++ would run on the phone. So here are my first impressions after checking out the flatpak version in [Mobian](https://wiki.mobian-project.org/) (= Debian for mobile devices):

Xournal++ is running without fatal errors. The UI would need some tweaks to fit to the screen of the PinePhone (1440x770 resolution). Displaying pdfs and xopp-files in full screen mode works pretty nicely. I currently use a bluetooth mouse for drawing. Scrolling via touch and pinch to zoom work fine as well. The settings menu takes too much space to change any settings. So I haven't tried to draw via touch (or via a stylus) or to use the floating toolbox. In Mobian it is easy to switch between portrait and landscape mode. Depending on what you do, you may want to switch between these two modes. Here is a visual impression in landscape mode:

![FirstTry](https://user-images.githubusercontent.com/40485433/88791564-3ace7880-d19a-11ea-8391-b35a58dc72f3.jpg)

So if anyone is interested in creating a nice UX for Xournal++ on the PinePhone, get your hands on now! :-)

rolandlo:
As suggested in the [Tweaks section of Mobian](https://wiki.mobian-project.org/doku.php?id=tweaks) the apps can be scaled. I couldn't make it work with scale-to-fit. But it worked with changing `scale = 2` to `scale = 1` in `/etc/phosh/phoc.ini` (line 5, i.e. below `[OUTPUT:DSI-1]`) and rebooting the phone. This does change all the other apps, too, unfortunately, so I will need to reset it. 

After that rescaling operation the settings can be changed in Xournal++'s settings dialog without any problems. It seems that only the old input system works (although this may have changed in earlier versions; this flatpak is from Nov 25 of year 2019). 

No input devices were recognized, so I cannot use touch to draw. This is the most serious problem remaining. I can use touch for panning and zooming and I can draw with the bluetooth mouse (and write with the bluetooth keyboard, if I want). Pdf-background and image insertion work perfectly and so does pdf export. The floating toolbox works, too. 

[...]

PR #3267, opened 2021-07-25
Add minimal toolbar

rolandlo:
For performance reasons #3242 (on MacOS BigSur), for having a distraction-free toolbar for basic needs (see https://github.com/xournalpp/xournalpp/discussions/3262) and for devices with a low screen-resolution (such as the pinephone #2139) I am adding this minimal toolbar. It looks like this (with low screen resolution):
grafik

For everything else the menubar can be used. It's easy for users to create a customized toolbar that extends on this one.


rolandlo:
Actually for the Pinephone it would be better to have the same tools horizontally (top toolbar). Maybe add a horizontal minimal toolbar as well.

Technius (commented):
I really like this, although I feel like it would be better if there were more zoom options.

rolandlo:
> I really like this, although I feel like it would be better if there were more zoom options.

I can add the zoom slider. There are marks for the distinguished zoom levels. My goal was that the toolbar fits onto my screen when I use 200% scaling on my 2560x1440 screen (so effectively 1280x720 resolution). It won't fit with the zoom slider (or with any other additional tool), but it will still fit on a slightly larger screen, which is good enough for MacOS users.

What do you think about the idea of adding a `Minimal Left` and a `Minimal Top` toolbar with the same tools (one of them with the tools on the left, the other with the tools on the top)?

rolandlo:
This toolbar fits with 1600x900 resolution.
grafik

On the top it fits on even smaller screen resolutions (like 1280x720, with Gnome dock on the left).
grafik

Technius:
This looks good for a default. Users with more exotic devices will probably customize the toolbar anyways.

rolandlo:

Issue #3251, opened 2021-07-20
Enable scrolling without touching the tablet


redtrumpet:
**Is your feature request related to a problem? Please describe.**
I want to scroll the page without touching the screen. With 1.0.20 and the old input system, I applied the Hand tool to the button of my pen, and I was able to scroll by clicking the button without touching the screen.

This is not possible anymore with the new input system, I tried it with both v1.0.20 and v1.1.0.

**Describe the solution you'd like**
I'd like to be able to press the button on my pen, and then scroll the page by hovering the pen above my tablet.

**Describe alternatives you've considered**
I tried enabling the "Merge button events with stylus tip events", which makes this behaviour possible. However this is not feasible, because then other things go wrong. If I scroll outside the page, and then move the cursor back to the page the main tool will be activated. This makes it impossible to use the Vertical Space tool, since the whole page will suddenly shift.

**Additional context**
#2901 seems to be a similar problem regarding the floating toolbox. It may be appropriate to merge the two issues.

linux-addict:
This is a probably a duplicate of #1939 and #2664.

Issue #1939, opened 2020-05-05
Add a setting to drag the canvas without the pen touching the tablet


minecraft2048:
**Is your feature request related to a problem? Please describe.**
Currently if I set the button in the pen to activate the hand tool in preference>stylus>stylus buttons, after clicking the button I set, I need to touch the pen to the tablet to drag the canvas around

**Describe the solution you'd like**
Basically I want it to act like how Krita did, if I click a button in my pen and move the pen it moves the canvas around while the pen is still hovering.

Munzu:
That's weird, I can drag the canvas without touching just fine. I have my stylus button mapped to the hand tool as well

LittleHuba:
Please check your settings and untick the option "Merge button events with tip events" in the input system tab. You will also have to check for the wacom setting though.

minecraft2048:
> Please check your settings and untick the option "Merge button events with tip events" in the input system tab. You will also have to check for the wacom setting though.

It is already unchecked though:
![image](https://user-images.githubusercontent.com/2445636/82169707-b3e16200-9905-11ea-93b6-b7a123c7a8f7.png)

I'm currently using the PPA build though, built on May 17 2020, 03:16:38 and I'm not using a Wacom tablet



> That's weird, I can drag the canvas without touching just fine. I have my stylus button mapped to the hand tool as well

For some reason in my Arch Linux AUR install it works just like you described, version 1.0.18 git commit 535d261d, but I'm not sure whether it is intended or not, because when I press the button to drag the screen and I touched the pen to the tablet, it draws instead of drags

LittleHuba:
Did you switch the wacom setting mentioned in our settings dialog to off? It resets itself on a system restart by the way. We'd also require more info on what platform you are using. Libinput? Gnome?...

minecraft2048:
I'm not using Wacom tablet, I'm using XP-Pen G640 tablet, so `xsetwacom` doesn't work here, and for this installation I'm using Kubuntu 20.04 LTS with KDE, and libinput is installed

LittleHuba:
Okay please provide separate [input logs](https://github.com/xournalpp/xournalpp/wiki/How-to-provide-debug-info-for-input-related-bugs) for these situations:

1. Tap your screen once with the tip of your pen without pressing any button
2. Press the first button on your pen while hovering it over the screen. (The pen has to be within proximity of the screen). Release the button again without ever touching the screen.
[...]

Issue #1877, opened 2020-04-19
Stylus buttons only "halfway activated" when clicked


thesleort:
**Affects versions :**
 - OS: [Linux (Ubuntu 20.04)]
 - (Linux only) Desktop environment: X11
 - Which version of libgtk: 3.24.17-3ubuntu1
 - Version of Xournal++: 1.1.0+dev
 - Installation method: PPA

**Describe the bug**
When using the stylus the button doesn't work as intended for any of the 'Stylus Buttons' settings in preferences. 
Lets say that the `Hand` is selected when `Button 1` is clicked, then the pointer-icon only shows up, when the stylus hits the surface/screen. This means that it never "grabs" the page, so that I can scroll. (It's similar to selecting the `Hand`-tool with a mouse. If you are not holding down the mouse button, then it shows like a pointing hand, and when you hold the mouse button down it enters the grabbing state and shows a "grabbing hand". However, here when you use the stylus and press the stylus button it only shows as a pointing hand). 
_Above is an example with the hand tool, however, the behavior is the same for all tools available to the stylus buttons_

In the program `Xournal` grabbing the page using the button works as intended, so it cannot be a driver issue.

Also, it works correctly in Wayland on Ubuntu 20.04, I.e. it doesn't happen there.

**To Reproduce**
1. Use X11
2. Assign stylus some functionality like "Hand"
3. Click assigned stylus button and drag on the page.
4. Page should now scroll, but it doesn't because the cursor still shows a "Pointer" instead of "Grabbing hand".

**Expected behavior**
When the stylus button is clicked, it should become active and not enter some sort of "hovering state"

thesleort:
Have investigated further and it seems that when the xorg configuration for the wacom tablet/stylus is set to `Option "TPCButton" "on"`, this bug occurs.

When the option is set to `Option "TPCButton" "off"` it works as expected. 

I still regard it as a bug, as the old program `xournal` works no matter what the xorg option is. Also as mentioned it works in Wayland with both xorg options.

LittleHuba:
Please go into the settings and to the tab "Input System" all the way down there is an option to handle this exact problem.

LittleHuba:
You are welcome to provide a PR that checks automatically what setting is set within the system. But be aware it has to take into account that there might be multiple devices with different settings.

thesleort:
Pardon me for not finding that. Maybe this setting should be more clear.

[...]

PR #3270, opened 2021-07-25
show floating toolbox on button click

### LINKED ISSUE #2901: Bring up the floating toolbox without tapping the screen
For now, the Floating toolbox is considered a tool that you have to tap in order to bring up, which for example prevents the user from keeping some strokes selected and applying color via the toolbox, as using the floating toolbox deselects everything.

I think bringing up the toolbox directly whenever the corresponding stylus button is pressed would be more natural, and also resolve the problem above.

None.

**Additional context**
None.

Simons102:
Pop up the Floating Toolbox (if configured) by pressing a button on your pen without tapping the screen.

Even though this seems to work fine, I am not sure if this is the right way to do this, and open to suggestions how it could be done better.

Closes #2901

### DIFF
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -396,6 +396,8 @@ auto XojPageView::onButtonClickEvent(const PositionInputData& pos) -> bool {
 
         control->getWindow()->floatingToolbox->show(wx, wy);
     }
+
+    return true;
 }
 
 auto XojPageView::onButtonDoublePressEvent(const PositionInputData& pos) -> bool {
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -358,14 +358,7 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
         ImageHandler imgHandler(control, this);
         imgHandler.insertImage(x, y);
     } else if (h->getToolType() == TOOL_FLOATING_TOOLBOX) {
-        gint wx = 0, wy = 0;
-        GtkWidget* widget = xournal->getWidget();
-        gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
-
-        wx += std::lround(pos.x) + this->getX();
-        wy += std::lround(pos.y) + this->getY();
-
-        control->getWindow()->floatingToolbox->show(wx, wy);
+        this->showFloatingToolbox(pos);
     }
 
     return true;
@@ -387,14 +380,7 @@ auto XojPageView::onButtonClickEvent(const PositionInputData& pos) -> bool {
     ToolHandler* h = control->getToolHandler();
 
     if (h->getToolType() == TOOL_FLOATING_TOOLBOX) {
-        gint wx = 0, wy = 0;
-        GtkWidget* widget = xournal->getWidget();
-        gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
-
-        wx += std::lround(pos.x) + this->getX();
-        wy += std::lround(pos.y) + this->getY();
-
-        control->getWindow()->floatingToolbox->show(wx, wy);
+        this->showFloatingToolbox(pos);
     }
 
     return true;
@@ -537,12 +523,7 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
 
             if (doAction)  // pop up a menu
             {
-                gint wx = 0, wy = 0;
-                GtkWidget* widget = xournal->getWidget();
-                gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
-                wx += std::lround(pos.x + this->getX());
-                wy += std::lround(pos.y + this->getY());
-                control->getWindow()->floatingToolbox->show(wx, wy);
+                this->showFloatingToolbox(pos);
             }
         }
 
@@ -956,4 +937,17 @@ void XojPageView::elementChanged(Element* elem) {
     } else {
         rerenderElement(elem);
     }
+}
+
+void XojPageView::showFloatingToolbox(const PositionInputData& pos) {
+    Control* control = xournal->getControl();
+
+    gint wx = 0, wy = 0;
+    GtkWidget* widget = xournal->getWidget();
+    gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
+
+    wx += std::lround(pos.x) + this->getX();
+    wy += std::lround(pos.y) + this->getY();
+
+    control->getWindow()->floatingToolbox->show(wx, wy);
 }
--- a/src/gui/PageView.cpp
+++ b/src/gui/PageView.cpp
@@ -371,6 +371,33 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
     return true;
 }
 
+auto XojPageView::onButtonClickEvent(const PositionInputData& pos) -> bool {
+    Control* control = xournal->getControl();
+    double x = pos.x;
+    double y = pos.y;
+
+    if (x < 0 || y < 0) {
+        return false;
+    }
+
+    double zoom = xournal->getZoom();
+    x /= zoom;
+    y /= zoom;
+
+    ToolHandler* h = control->getToolHandler();
+
+    if (h->getToolType() == TOOL_FLOATING_TOOLBOX) {
+        gint wx = 0, wy = 0;
+        GtkWidget* widget = xournal->getWidget();
+        gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
+
+        wx += std::lround(pos.x) + this->getX();
+        wy += std::lround(pos.y) + this->getY();
+
+        control->getWindow()->floatingToolbox->show(wx, wy);
+    }
+}
+
 auto XojPageView::onButtonDoublePressEvent(const PositionInputData& pos) -> bool {
     // This method assumes that it is called after onButtonPressEvent but before
     // onButtonReleaseEvent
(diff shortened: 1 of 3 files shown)


src/gui/PageView.cpp
```diff
@@ -371,6 +371,33 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
+    if (h->getToolType() == TOOL_FLOATING_TOOLBOX) {
+        gint wx = 0, wy = 0;
+        GtkWidget* widget = xournal->getWidget();
+        gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);
+
+        wx += std::lround(pos.x) + this->getX();
+        wy += std::lround(pos.y) + this->getY();
+
+        control->getWindow()->floatingToolbox->show(wx, wy);
+    }
```
personalizedrefrigerator, line 398: This code is the same as code in `XojPageView::onButtonPressEvent` (and also `XojPageView::onButtonReleaseEvent`). I think it makes sense to add a private function to `PageView.cpp`:
```c++
/**
 * Shows the floating toolbox at the location of an input event.
 */
static void showFloatingToolbox(XournalView* xournal, const PositionInputData& pos) {
    Control* control = xournal->getControl();

    gint wx = 0, wy = 0;
    GtkWidget* widget = xournal->getWidget();
    gtk_widget_translate_coordinates(widget, gtk_widget_get_toplevel(widget), 0, 0, &wx, &wy);

    wx += std::lround(pos.x) + this->getX();
    wy += std::lround(pos.y) + this->getY();

    control->getWindow()->floatingToolbox->show(wx, wy);
}
```

...that we can use to reduce code duplication!

src/gui/PageView.h
```diff
@@ -152,6 +152,11 @@ class XojPageView: public Redrawable, public PageListener {
     bool onMotionNotifyEvent(const PositionInputData& pos);
     void onMotionCancelEvent();
 
+    /**
+     * for popup windows like the floating toolbox
+     */
+    bool onButtonClickEvent(const PositionInputData& pos);
```
personalizedrefrigerator, line 158: **Documentation:**
If the comment in `StylusInputHandler` on lines 46-60 is applied, this documentation comment should describe the difference between `onButtonClickEvent` and `onButtonPressEvent`, `onButtonReleaseEvent` (e.g. `this event fires even when no input sequence is actively running`).

---

**Design:**

For future programmers trying to add additional click functionality, I think it could be confusing that `onButtonClick` is only called by `StylusInputHandler` (while other events like `onButtonPressEvent` are also used in, for example, `PenInputHandler`).

Do you think it would make sense for `onButtonClickEvent` to also be called in `onButtonReleaseEvent`?
  * If (in StylusInputHandler) we only call `onButtonClickEvent` directly when there's no input sequence running, we won't be double-calling `onButtonClickEvent`.
  * Additionally, this would let us remove some of the duplicated code I commented about in `PageView`.

personalizedrefrigerator (commented):
Thank you for working on this! I have a few design comments (mostly about `onButtonClickEvent`). It will be very nice to have this!

personalizedrefrigerator:
Thank you for the pull request! This change should make Xournal++ much easier to use.

rolandlo:
While this feature is certainly nice to have, I'd like to point out that there are several related issues #2901, #3251, #1939, #1877 (and most likely others) that may have to be addressed at once. It's not only the floating toolbox, but also other tools that should be usable when clicking without touching the screen.

Simons102:
Thanks for the detailed review @personalizedrefrigerator :)
Totally missed that return statement in `onButtonClickEvent`.
By the way, is there any specific reason to use `auto` as the return type?
I added `XojPageView::showFloatingToolbox`, although not as a static function because of the calls to `this->getX()` and `this->getY()`, and removed the check if we're using the floating toolbox from `StylusInputHandler::handleImpl` as you suggested.
I agree with you that `onButtonClickEvent` only being called by `StylusInputHandler` could cause confusion; maybe a good idea would be to also call it wherever pressing buttons on a drawing tablet is handled (which I can't test as I only own a convertible with stylus input), or to give it a different name?
As for calling `onButtonClickEvent` in `onButtonReleaseEvent` too: I'm not sure if I understood your suggestion correctly, but think that a 'click' should either happen when pressing a button or releasing it, and not both on different occasions.
Again, thank you for taking the time take a look at this!
@rolandlo, I'm not sure if some of those Issues are out of the scope of this pull request (for example #1939), as to me the floating toolbox seems more like a one-shot thing, compared to keeping a tool active while a stylus button is pressed. I think the best thing is to either decide on a tool by tool basis which tools should be activated (e.g. hand) while pressing a button and which shouldn't (like for example the pen or highlighter, which would start drawing on the screen), or add an option to the settings where this behaviour can be configured.
I could try to implement one of those and open another pull request though, as I currently have some time to spare :)

personalizedrefrigerator:
> As for calling onButtonClickEvent in onButtonReleaseEvent too: I'm not sure if I understood your suggestion correctly, but think that a 'click' should either happen when pressing a button or releasing it, and not both on different occasions.

I'll have to test this, but I think that if <b>onButtonReleaseEvent</b> is called (currently), then <b>onButtonClickEvent</b> will not be called for that event.

If we make <b>onButtonReleaseEvent</b> call <b>onButtonClickEvent</b> for "click like" logic, we have that tools that always start an input sequence when clicking (e.g. mouse) will call <b>onButtonClickEvent</b> in addition to the stylus.
 * If I'm right that this won't lead to calling <b>onButtonClickEvent</b> twice for the same input sequence (e.g. mouse down, mouse move, mouse up is an input sequence), this change could make it easier to add click actions in the future.

Simons102:
While it seems like this doesn't call onButtonClickEvent twice for the same input sequence, it leaves us with this behaviour on mouse down, mouse move, mouse up:

https://user-images.githubusercontent.com/43406797/127057975-51a93a6f-db23-4e87-b6cf-dbb85278e6a6.mp4

which feels a little weird.

It might be a good idea to rename`onButtonClickEvent`, as it's called when a button was pressed and not when a button was pressed and released again, which is more of a complete click. But I have no idea what it could be renamed to.

Would calling `onButtonClickEvent` from `onButtonPressEvent` instead of `onButtonReleaseEvent` and moving the code that handles tools that respond to click actions from `onButtonPressEvent` to `onButtonClickEvent` be another option?
Although I can't think of another tool aside from the floating toolbox that does this.
That would preserve the current behaviour for mouse based input, (most likely) not call `onButtonClickEvent` twice for the same input sequence and not make more than one call to e.g. `FloatingToolbox::show` for the same event.

Simons102:
This should be ready now.

personalizedrefrigerator:
I'm noticing that when I use the floating toolbox as I would normally (stylus button down, touch pen to screen, stylus button up, pen above screen), any subsequent taps (with or without the button down) open the floating toolbox.

personalizedrefrigerator:
> This should be ready now.

Thank you for working on this! It looks like it's almost ready! See the issue I mentioned above.

Simons102:
Weirdly enough I can only reproduce this on Wayland and not on Xorg.
I don't think that this PR caused that bug, just that it made the bug easier to notice, because it's also possible to get into that state in the current release by pushing the button, touching the stylus to the screen (will open the floating toolbox), removing the stylus from the screen, touching the stylus to the screen (on the floating toolbox), releasing the button and then removing the stylus from the screen again.

Calling `inputHandler->onButtonReleaseEvent` when calling `XojPageView::onButtonClickEvent` from `StylusInputHandler::handleImpl` seems impossible because `inputHandler` was always `null` in that situation.

The issue is that `StylusInputHandler::handleImpl` isn't called for the button release event that is fired when the stylus button is released while the stylus is touching the floating toolbox. That in turn causes `StylusInputHandler::setPressedState` to not be called which means that in `StylusInputHandler` `this->modifier2` or `this->modifier3` simply stay true, and thus `StylusInputHandler` thinks that the stylus button is still being pressed.

I'll try to fix that, but where that `ButtonReleaseEvent` goes is currently beyond me.

personalizedrefrigerator:
> I'll try to fix that, 

Fixing it could go in a different pull request. 
This PR looks good to me!

LittleHuba:
Please also fix the last mentioned issue. This could lead to buggy input behavior which I do not want to introduce to the development branch.

Simons102:
I don't think that I can fix the last mentioned issue.
That behaviour can be reproduced on the current 1.1.0 release of Xournal++, but so far I could only trigger it on Wayland (Gnome).
It seems like Wayland doesn't give us a button release event for a stylus button outside of the widget that received the corresponding button press event, if the stylus tip was lifted from the screen after pressing and before releasing the button.
I guess that's more likely something that can only properly be fixed upstream.
Nevertheless I created #3302.

Simons102:
If we do something like
```c++
g_signal_connect(this->window, "button-release-event", G_CALLBACK(onButtonReleaseCallback), this);
```
in `MainWindow::MainWindow`, and then in `onButtonReleaseCallback` filter for the device class `INPUT_DEVICE_PEN` or `INPUT_DEVICE_ERASER`, and only in that case pass the event on to `InputContext::handle` (or maybe directly to the `Stylus Handler`), the mentioned issue seems to mostly vanish, but can still be triggered by opening a dropdown from the floating toolbox and then releasing the stylus button.
But doing that causes InputContext::handle to be called twice for a button release event when the stylus is over the document area, and seems kind of hacky.
The issue still remains that we cannot know if a stylus button was released while the stylus was outside of the application's window.

LittleHuba:
Okay I thought this was a result of your work. As it already appears in 1.1.0 as well I have no more reason to block this PR.

LittleHuba:
Would you mind checking for a ticket over at GTK? If you find none, feel free to create one and reference it with #3302.

Simons102:
Alright, I didn't find a ticket for this at GTK and will open one as soon as I can reproduce this with one of the [demos](https://gitlab.gnome.org/GNOME/gtk/-/tree/master/demos), or have a small and self contained example ready that I can attach to it. Which might take a while.

Simons102:
There now exists a [ticket at GTK](https://gitlab.gnome.org/GNOME/gtk/-/issues/4157), but I'm not sure how I should go about referencing it with #3302, other than putting a link to it in the comments.
Hopefully that actually is an issue with GTK and can be fixed soon.

personalizedrefrigerator:
Overall, this PR (for me) makes it easier to use the floating toolbox on Wayland. I would like to merge it in 48 hours, if that seems appropriate.

PR #3271, opened 2021-07-26
generateConvert script: Change «, » to "

personalizedrefrigerator:
## Summary
 * Makes `generateConvert.php`'s output match the current contents of the `*.generated.cpp` files.


personalizedrefrigerator:

PR #3276, opened 2021-07-27
Update Portaudio to version 19.7 on MS Windows

rolandlo:
Portaudio has been updated to version 19.7 on April 6, 2021, see http://files.portaudio.com/download.html. On MS Windows Portaudio is built from source in the build process and the build script downloads from a hard-coded tar-ball filename. I have updated the filename and the corresponding checksum.


rolandlo:
Looking at the CI pipeline it seems that the build time for Portaudio is reduced by roughly 50% (from about 3 min to 1.5 min)

rolandlo:

Issue #3279, opened 2021-07-28
Stroke Recognizer should preserve endpoints for recognized lines.


kenballus:
Affects Versions: All

The Bug: When drawing a line with the stroke recognizer on, the endpoints of the line are not preserved. 99% of the time that I'm trying to draw a line, I can get the endpoints to be right; I need help only with drawing a straight line between them. This is the behavior in many competing stroke recognizers (i.e. the one used in GoodNotes) This behavior is especially frustrating when trying to draw a straight line between two preexisting points.

To Reproduce: Use the stroke recognizer to draw a line that is straightest in the middle. The endpoints will change to match the middle.

Expected Behavior: Whenever a stroke is recognized as a line, it is transformed by preserving its endpoints and extending a straight line between them.

rolandlo:
Your suggestion makes sense to me as an enhancement and shouldn't be hard to implement. I wouldn't call it a bug, though.

kenballus:
I'm glad to see that you like the suggestion. I just submitted PR #3282, which implements this new behavior.

PR #3282, opened 2021-07-29
Update stroke recognizer so recognized lines preserve endpoints.

kenballus:
Fixes issue #3279.

This patch changes the behavior of the stroke recognizer when drawing lines. Instead of replacing the user's stroke with a line of best fit, the user's stroke is now replaced with a straight line connecting the first and last point in the stroke.


kenballus:
Converted to draft because this removes the ability to draw straight vertical and horizontal lines.

PR #3284, opened 2021-07-29
Bugfix: corrected behaviour of StringUtils::rtrim

er888kh:
StringUtils::rtrim previously only deleted the first (from left)
TRIM_CHAR from the input string


LittleHuba:
Could you please add a reference to an open issue?
Default behavior of string::erase() is to delete all characters from the provided position. As far as I can see, this behavior is the same from the time when this method was added. Further, your modification does not make any difference, as you supply as second argument what is there by default. 

Sorry, but I don't really get the point of this change.

er888kh:
You are absolutely right. The default behavior is the same; for a moment I thought std::string::erase behaves like the one for some of the containers.

PR #3285, opened 2021-07-29
Update stroke recognizer to preserve line endpoints.

kenballus:
Implements enhancement suggested in issue #3279.

Commit breakdown:
 - 43cc161: Changes stroke recognizer so that when a line is recognized, the endpoints are preserved a straight line is extended between them. Introduces a bug that makes it so nearly vertical/horizontal lines don't snap to the vertical/horizontal axis.
 - 831895c: Merge
 - b6dcf26: Removes an unnecessary "using std::map" from a header file, as specified in issue #2354.
 - 6a5ce40: Fixes the bug introduced in 43cc161.
 - de2589d: Fixes formatting of if statement to satisfy clang formatting.


src/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -297,22 +297,33 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke) -> ShapeRecognizerResult
...
-            s->addPoint(Point(rs->x1, rs->y1));
-            s->addPoint(Point(rs->x2, rs->y2));
+            if (aligned)
+            {
+                s->addPoint(Point(rs->x1, rs->y1));
+                s->addPoint(Point(rs->x2, rs->y2));
+            } else {
+                s->addPoint(Point(stroke->getPoints()[0].x, stroke->getPoints()[0].y));
+                s->addPoint(Point(stroke->getPoints()[stroke->getPointCount() - 1].x,
+                                  stroke->getPoints()[stroke->getPointCount() - 1].y));
```
LittleHuba, line 323: Please store the vector returned from `stroke->getPoints()` in a local variable before creating the points using this local vector. Doing it this way makes it hard to read and also forces several derefenciations of pointers that cost a lot of time.

kenballus:
There are a lot of commits here, and I can imagine that this could clutter the commit history. I know just enough git to get by, so if you'd like me to reformat some of this, I'm happy to do that.

src/gui/toolbarMenubar/ToolPageLayer.h
```diff
@@ -18,8 +18,6 @@
 
 #include "AbstractToolItem.h"
 
-using std::map;
-
```
LittleHuba, line 22: As this has not really anything to do with the rest of your PR, please revert this and create a separate PR for this.

src/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -297,22 +297,33 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke) -> ShapeRecognizerResult
...
         if (n == 1)  // current stroke is a line
         {
+            bool aligned = false;
             if (fabs(rs->angle) < SLANT_TOLERANCE)  // nearly horizontal
             {
                 rs->angle = 0.0;
                 rs->y1 = rs->y2 = rs->ycenter;
+                aligned = true;
             }
             if (fabs(rs->angle) > M_PI / 2 - SLANT_TOLERANCE)  // nearly vertical
```
LittleHuba, line 307: This should be an `else if`. You could also merge the logic for adding the point with this logic but that is not really necessary.

LittleHuba (changes requested):
Thanks for the work. Here are some pointers on stuff to improve.

LittleHuba:
The number of commits is not really important for this size of the PR. We can simply squash it once we merge it.

src/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -297,22 +297,30 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke) -> ShapeRecognizerResult
...
 
-            s->addPoint(Point(rs->x1, rs->y1));
-            s->addPoint(Point(rs->x2, rs->y2));
+            if (aligned) {
+                s->addPoint(Point(rs->x1, rs->y1));
+                s->addPoint(Point(rs->x2, rs->y2));
+            } else {
+                auto points = stroke->getPoints();
+                s->addPoint(Point(points[0].x, points[0].y));
+                s->addPoint(Point(points[stroke->getPointCount() - 1].x, points[stroke->getPointCount() - 1].y));
```
LittleHuba, line 321: You can still simplify this:
```suggestion
                s->addPoint(points[0]);
                s->addPoint(points[stroke.size() - 1]);
```

LittleHuba (changes requested):

src/control/shaperecognizer/ShapeRecognizer.cpp
kenballus: Thanks! I didn't notice Stroke::getPointVector, so I'm using that instead of Stroke::getPoints. Now I can just use vector::front and vector::back.

LittleHuba (approved):

PR #3287, opened 2021-07-29
Remove using std::map from header

kenballus:
Remove the last using std::something from a header file, as requested in #2354.


LittleHuba:
Just as a side-note: Please never merge `master` into your PR-branch. Instead, rebase your branch on top of master.
This is because, in a merge, you must potentially edit code from others. This may lead to you accidentally introducing a bug into the work of somebody else. If you perform a rebase instead, you only edit your own work, which we will look over before merging.

In this instance all is good and you can leave it as it is. We will squash the commits and you only changed the one thing you wanted to change.

LittleHuba:
LGTM merge in 24 hours.

rolandlo (approved):

Issue #2913, opened 2021-03-06
Control over pen size?


Kreijstal:
As far as I know  there are only 5 options, `very fine, fine, medium, thick, very thick`, which imho it's pretty limited, is this due to a inherent limitation or could the user choose an arbitrary value?

rolandlo:
The file format would support arbitrary pen size, but the Gui only has these 5 options.

Technius:
Related: #1361

personalizedrefrigerator:
A workaround available in 1.1.0+dev:

![](https://user-images.githubusercontent.com/46334387/111021106-feddd180-837e-11eb-91dd-207db42bc4da.png)
Change the "Pressure Multiplier" under Settings > Stylus > Pressure Sensitivity. Unfortunately, this requires opening and closing settings.

Technius:
Closing as duplicate of #1361.

Issue #1361, opened 2019-07-20
Stroke thickness slider


Technius:
**Is your feature request related to a problem? Please describe.**
There are currently only three selectable _hardcoded_ stroke thickness levels in the interface. It would be useful to be able to choose more specific thickness levels. Current workarounds include manually editing `.xopp` files, which is extremely annoying, tedious, and undesirable.

**Describe the solution you'd like**
The three defaults are decent (in my opinion) and do not need to be changed. I propose adding a new "thickness slider" button that, when pressed, will present a small popup window (non-modal) with a thickness slider inside of it. The slider should range from some very small size (e.g. 1 pixel wide) to a very large size (e.g. 1/3 of the page width); this should be sufficient for most use cases.

When a stroke is selected, the thickness slider should also affect the selected stroke.

**Describe alternatives you've considered**
We could remove the stroke buttons and color buttons altogether and introduce some sort of "properties" pane in the style of GIMP or Photoshop. While this would be far more powerful and useful for future tools, it would also be less intuitive and would more work to implement than just a thickness slider.

**Additional context**
Thickness is already stored and handled as a `double` internally, so this is really just a UI change.

JJones780:
For stylus/touch users having the toolbuttons to select pre-set thickness ( and colours ) is a necessity. 
There are 5 presets for the pen tool available via the menu and as you noticed, with #1416 they will all have toolbuttons ( and the highlighter and eraser will be bumped up to 5 as well).

I think having a slider as well as pref settings for the thickness values would be ideal.

atticus-sullivan:
Is this already in a branch or even on the master branch (just only here on github and not in the repos)?
I'd really like to have this, since right now thick is too big and medium is too thin to mark the lines of the document I'm currently working in.

Technius:
No one has started working on this yet, but it should be fairly easy and quick to implement if someone does.

atticus-sullivan:
Oh, I don't work with github that long and I missread the mentions for some sort of pull requests.

PR #3269, opened 2021-07-25
WIP: Add window opacity slider tool

personalizedrefrigerator:
## Summary
_see also_ #3312.

 * Implements suggestion 10 from #3259
Screenshot of semi-transparent Xournal++

## To-do
 * [x] Error handling for when the user doesn't have a compositor running.
 * [x] Remove code duplication (maybe `ZoomSlider` and the opacity slider should share a slider base-class?)
     * `AbstractSliderItem` (currently used by just the opacity slider)
 * [ ] Setting for whether to make the toolbar transparent, in addition to the Xournal widget.
     * Potential reference: https://stackoverflow.com/questions/16832581/how-do-i-make-a-gtkwindow-background-transparent-on-linux

PR #3288, opened 2021-07-30
Add custom thickness slider

kenballus:
Implements a custom thickness option as specified in #1361 and #2913.

This is very much WIP. This patch has no polish and probably introduces a lot of bugs. Some code is copied and pasted, mostly from the fill opacity dialog. I do not claim that this is good code, but it seems like it works.
To try it out, go to Tools->Pen Options->Set Custom Thickness, then set a thickness. (The slider is in serious need of calibration, so the pen will probably be huge.)

I would be very appreciative if someone could provide constructive criticism or suggestions for making this more presentable.


kenballus:
This is starting to come together. To do:
 - icons (currently in progress)
 - more convenient UI. As it is, setting a custom thickness requires digging through a lot of menus. Ideally, you could click again on the custom thickness button once it's selected and have the slider window pop up.

Any ideas about what the icons should look like, or how to make the UI better would be greatly appreciated!

kenballus:
Added a logarithmic scale. Now, I don't think the window needs to be wider. Remaining to do are icons and putting the slider in a more convenient spot.

kenballus:
I still think the slider could go in a more convenient place, but I'm pretty happy with the level of polish this has for now. The icon could probably use a little work, too.

rolandlo:
Nice to have custom thickness. One thing I noticed: The existing line styles don't work well with large stroke width. I'm wondering, if we can/need to do something about it.

kenballus:
> Nice to have custom thickness. One thing I noticed: The existing line styles don't work well with large stroke width. I'm wondering, if we can/need to do something about it.

Thanks for pointing this out. It turns out that the line styles don't work well with the built-in "very thick" pen setting either. So I think that issue should qualify as a separate bug. I'm looking into fixing it right now. Do you think it's reasonable to set for all thicknesses the ratio of line width to gap length between dashes to match the ratio that the "medium" thickness currently has? Or should the relationship be nonlinear?

rolandlo:
In my opinion when the "Preserve line width" option in the "Drawing Area" panel of the preferences is switched off, scaling a dashed/dotted/dashdotted line (via one of the corner anchors of the selection tool) should just do a uniform scaling as one would expect. Then one has to look at the code what changes are needed to achieve that and whether backwards compatibility is an issue.

rolandlo:
Fixing the line style scaling should go into a different PR, indeed.

Fatih20:
> I still think the slider could go in a more convenient place, but I'm pretty happy with the level of polish this has for now. The icon could probably use a little work, too.

Which icon are you talking about? I'll get to work on it once this feature is finalized and ready to merge.

kenballus:
> > I still think the slider could go in a more convenient place, but I'm pretty happy with the level of polish this has for now. The icon could probably use a little work, too.
> 
> Which icon are you talking about? I'll get to work on it once this feature is finalized and ready to merge.

Thanks. I was referring to the icon for the new button that this patch adds. It goes right next to the "thick" width in the default toolbar. It's called "xopp-thickness-custom.svg" in the icons directories.

Fatih20:
> > > I still think the slider could go in a more convenient place, but I'm pretty happy with the level of polish this has for now. The icon could probably use a little work, too.
> > 
> > 
> > Which icon are you talking about? I'll get to work on it once this feature is finalized and ready to merge.
> 
> Thanks. I was referring to the icon for the new button that this patch adds. It goes right next to the "thick" width in the default toolbar. It's called "xopp-thickness-custom.svg" in the icons directories.

What does the icon represent? What action does it do when I click on it?

kenballus:
It switches the current tool thickness to a custom thickness which was set previously by the user.

Fatih20:
> It switches the current tool thickness to a custom thickness which was set previously by the user.

Oh, so it doesn't pop out a slider or anything?

kenballus:
> > It switches the current tool thickness to a custom thickness which was set previously by the user.
> 
> Oh, so it doesn't pop out a slider or anything?

I guess I'm still looking for suggestions on that front. What do you think is the way to do it? As it is, the slider popup is in a menu, and the button just changes the thickness to the one you set in the menu.
Do you think there should be a slider button in the toolbar? If so, would there be a separate button to switch to the thickness set by the slider once it's set? Or maybe a button that you click once to switch to custom thickness, and again to open the slider?

Fatih20:
> > > It switches the current tool thickness to a custom thickness which was set previously by the user.
> > 
> > 
> > Oh, so it doesn't pop out a slider or anything?
> 
> I guess I'm still looking for suggestions on that front. What do you think is the way to do it? As it is, the slider popup is in a menu, and the button just changes the thickness to the one you set in the menu.
> Do you think there should be a slider button in the toolbar? If so, would there be a separate button to switch to the thickness set by the slider once it's set? Or maybe a button that you click once to switch to custom thickness, and again to open the slider?

Now I'm no software engineer, so my suggestion may be one hell of an engineering challenge. But yes, I would prefer a slider in the UI. How does the slider correspond to the hardcoded thickness selection? Simple, at least UX wise, the pencil, highlighter, and other tools would always uses the slider to define its thickness. The hardcoded the thickness selection merely makes the slider jump to a specific point when clicked.

kenballus:
> How does the slider correspond to the hardcoded thickness selection?

This patch adds a 6th thickness selection called "Custom." This thickness defaults to the same value as Medium for each tool. Then, by clicking on Tools->[Pen|Eraser|Highlighter] Options->Custom Thickness, you can open a slider window that allows you to set a new thickness for that Custom thickness selection. The preset thickness options remain unchanged. Would you rather we remove the hardcoded thickness options and just have a slider with tickmarks? That would require a little more work, but I think it's totally doable. If others agree that that's better UX, then I'll implement it.

Fatih20:
> > How does the slider correspond to the hardcoded thickness selection?
> 
> This patch adds a 6th thickness selection called "Custom." This thickness defaults to the same value as Medium for each tool. Then, by clicking on Tools->[Pen|Eraser|Highlighter] Options->Custom Thickness, you can open a slider window that allows you to set a new thickness for that Custom thickness selection. The preset thickness options remain unchanged. Would you rather we remove the hardcoded thickness options and just have a slider with tickmarks? That would require a little more work, but I think it's totally doable. If others agree that that's better UX, then I'll implement it.

That was a rethorical question that I answered in the next sentence as a suggestion on how the hardcoded thickness selection should reconcile with the slider.

Fatih20:
> > How does the slider correspond to the hardcoded thickness selection?
> 
> This patch adds a 6th thickness selection called "Custom." This thickness defaults to the same value as Medium for each tool. Then, by clicking on Tools->[Pen|Eraser|Highlighter] Options->Custom Thickness, you can open a slider window that allows you to set a new thickness for that Custom thickness selection. The preset thickness options remain unchanged. Would you rather we remove the hardcoded thickness options and just have a slider with tickmarks? That would require a little more work, but I think it's totally doable. If others agree that that's better UX, then I'll implement it.

What I meant is that the hardcoded selection is right there, next to the actual slider. Clicking one of them would cause the slider to jump around.

kenballus:
> > > How does the slider correspond to the hardcoded thickness selection?
> > 
> > 
> > This patch adds a 6th thickness selection called "Custom." This thickness defaults to the same value as Medium for each tool. Then, by clicking on Tools->[Pen|Eraser|Highlighter] Options->Custom Thickness, you can open a slider window that allows you to set a new thickness for that Custom thickness selection. The preset thickness options remain unchanged. Would you rather we remove the hardcoded thickness options and just have a slider with tickmarks? That would require a little more work, but I think it's totally doable. If others agree that that's better UX, then I'll implement it.
> 
> What I meant is that the hardcoded selection is right there, next to the actual slider. Clicking one of them would cause the slider to jump around.

I think that sounds like a really good idea.

Fatih20:
> > > > How does the slider correspond to the hardcoded thickness selection?
> > > 
> > > 
> > > This patch adds a 6th thickness selection called "Custom." This thickness defaults to the same value as Medium for each tool. Then, by clicking on Tools->[Pen|Eraser|Highlighter] Options->Custom Thickness, you can open a slider window that allows you to set a new thickness for that Custom thickness selection. The preset thickness options remain unchanged. Would you rather we remove the hardcoded thickness options and just have a slider with tickmarks? That would require a little more work, but I think it's totally doable. If others agree that that's better UX, then I'll implement it.
> > 
> > 
> > What I meant is that the hardcoded selection is right there, next to the actual slider. Clicking one of them would cause the slider to jump around.
> 
> I think that sounds like a really good idea.

I like to think it is because that's how many apps handle it. They have sliders and they have discrete sizes to choose. Is that feasible from an engineering perspective though? Since you'd have to put the slider right there in the actual toolbar.

@mjg What do you think?

personalizedrefrigerator:
> Do you think there should be a slider button in the toolbar?

#3269 abstracts some of the toolbar-slider logic (see [AbstractSliderItem.h](https://github.com/xournalpp/xournalpp/blob/8792242a67e5414cc85549bac94398476334fc04/src/gui/toolbarMenubar/AbstractSliderItem.h)/[AbstractSliderItem.cpp](https://github.com/xournalpp/xournalpp/blob/8792242a67e5414cc85549bac94398476334fc04/src/gui/toolbarMenubar/AbstractSliderItem.cpp)).

If this would be useful here, I could move this refactor to a separate PR! :)

mjg:
We have sliders for page zoom, fill opacity and then again in settings for several things.
If I operate the page zoom slider and then +/- it steps through different values than before, so it's not quite yet what we want for thickness.
For things like opacity, I often know I want a specific value but have to fiddle around with the slider. So, that could benefit from a revamp, too.
In settings, we also have fields where you can enter a value as well as operate with +/-.

Ideally, we could combine these features:
- easy quick adjustment (slider type)
- easy selection of discrete values/presets
- a way to specify an exact value

darktable has perfected that, but that's quite a target to aim for. I don't know glade, so I can't judge what is possible within this framework.

src/control/ToolEnums.h
```diff
@@ -21,6 +21,7 @@ enum ToolSize {
     TOOL_SIZE_MEDIUM,
     TOOL_SIZE_THICK,
     TOOL_SIZE_VERY_THICK,
+    TOOL_SIZE_CUSTOM,
```
personalizedrefrigerator, line 24: Be sure to handle `TOOL_SIZE_CUSTOM` in `ToolEnums.cpp`!

kenballus: Thanks!

kauesena:
Is there some file one could edit to create some intermediate size between two already existing sizes? Say, create a new size between FINE and MEDIUM. For my handwriting, a little finer then MEDIUM would be perfect.

kenballus:
> Is there some file one could edit to create some intermediate size between two already existing sizes? Say, create a new size between FINE and MEDIUM. For my handwriting, a little finer then MEDIUM would be perfect.

As of right now, I think there's no easy way to add extra thickness options without editing the source code in a bunch of places. This PR does that, but it's pretty outdated now and there's probably new places that need to be edited. As a workaround, you could just change the MEDIUM thickness to be a little thinner. If you don't mind recompiling, the way to do that is to edit [src/control/ToolHandler.cpp](https://github.com/xournalpp/xournalpp/blob/master/src/control/ToolHandler.cpp). Lines 28-54 establish the preset thicknesses for each tool. For example, to make MEDIUM ~30% thinner for the pen tool, you could change line 30 from
```C++
thickness[TOOL_SIZE_MEDIUM] = 1.41;
```
to
```C++
thickness[TOOL_SIZE_MEDIUM] = 1.00;
```

Technius:
Hi @kenballus, do you plan to update this PR any time soon? If not, I can help finish the PR.

kenballus:
> Hi @kenballus, do you plan to update this PR any time soon? If not, I can help finish the PR.

Hi @Technius. I don't have plans to update this anytime soon, since I think I went about doing it in the wrong way. This PR adds a sixth thickness for each tool that can be changed with a slider. I think a better solution is to have there be only one thickness value, and have the preset thicknesses just change that value. That design makes adding a slider trivial, and allows for the addition of more presets in the future without having to change the code in 50 places. Implementing that change requires touching more of the code than I am comfortable doing, so I'm not sure if I'm the right person for the job. Also, I won't have a lot of time to work on this until at least mid-March, so it might be better for someone else to take over on this PR. Thanks a lot.

kenballus:
This is a pretty critical missing feature, and I'm looking to give this another try. I think there are two viable paths forward:
1. Add a 6th element to `Tool::thickness` that corresponds to a custom thickness value set by the user. (Less work, less satisfying)
2. Add `double Tool::current_thickness`, which is the current thickness of the tool. Remove all places in the program that assume that the thickness will be one of a fixed, named set. Refactor as many things that rely on `ToolSize` as possible. (More work, more satisfying)

Can someone with authority weigh in on which solution is better? I'm leaning toward the second option, but I'd hate to do the work just for someone up the chain to shoot it down.

rolandlo:
> This is a pretty critical missing feature, and I'm looking to give this another try. I think there are two viable paths forward:
> 
>     1. Add a 6th element to `Tool::thickness` that corresponds to a custom thickness value set by the user. (Less work, less satisfying)
> 
>     2. Add `double Tool::current_thickness`, which is the current thickness of the tool. Remove all places in the program that assume that the thickness will be one of a fixed, named set. Refactor as many things that rely on `ToolSize` as possible. (More work, more satisfying)
> 
> 
> Can someone with authority weigh in on which solution is better? I'm leaning toward the second option, but I'd hate to do the work just for someone up the chain to shoot it down.

To me the second solution looks indeed better/cleaner. But let's hear from @Technius as well.

Technius:
I like the second approach more; I think it makes sense to refactor the tools so that thickness is stored as an actual number and not a choice from a set of predefined constants. Based on a quick skim through the code, I estimate that it should be possible to implement without touching too many places. Some of the things that will need to be changed are:

* `ToolBase` should store the `size` as a `double`.
* We can keep around `Tool.thickness`, but it needs to be renamed to something more accurate like `thicknessPresets`.
* Add a `ToolBase::setSize(double)` overload to directly set the size (without the preset).
* `ToolHandler::getThickness()` should return the size directly.
* The widgets for the thickness sizes should only be marked as active when the size matches (+/- some tolerance for float rounding errors)

kenballus:
University is on winter break, so I'm working on thistoday. Looking to have something workable by end of day.

Lolol4572:
@kenballus Just curious, not asking for anything to be rushed, but what is the status of this issue?

atticus-sullivan:
I'd still love to have this feature. Is there anything I can do to help?

Or is this PR too broken after such a long time and we should start over again?

atticus-sullivan:
Alright, I'll try to rework this. But I won't do it right now, so if someone else wants to work on it, just do it (maybe leave a short note to avoid that multiple people work simultaneously on this, I'll do the same).

kenballus:
I'm closing this PR because it's really out of date, and I don't plan to revive it. Hope this feature makes it into master somehow!

PR #3297, opened 2021-08-01
Fix undo layer rename

### LINKED ISSUE #3257: Undo/redo does not work properly with layer renaming
**Affects versions :** 1.1.0
 - OS: Ubuntu 21.04 (probably doesn't matter)
 - (Linux only) Desktop environment: Wayland, X11 (probably doesn't matter)
 - libgtk: 3.24.25 (probably doesn't matter)
 - Installation method: PPA (probably doesn't matter)

**Describe the bug**
After renaming multiple layers and renaming a layer, undoing the renaming does not only rename the renamed layer, but also the currently selected layer

**To Reproduce**
1. Open a new document
2. Create multiple (at least 2) layers via the dropdown menu in the toolbar (and use the "layer preview" in the sidebar to see their names)
3. Rename one of the layers via CTRL+R
4. Select a different layer in the layer preview
5. Undo via CTRL+Z
6. Observe that not only the renamed layer got its original name, but also the selected layer gets the original name of the renamed layer

**Expected behavior**
The undo operation should only change the name of the layer that was renamed. Similarly with redo.

**Additional context**
I have already mentioned that issue in #3204.

ja-he:
Fixes #3257. Since nobody touched that issue in a few days, here's a PR with what I suggested there.

1. There was an definite error in the undo/redo logic, simple fix.
2. For the correct change to properly reflect in the UI, I figured calling `fireRebuildLayerMenu()` would be good, so we give the undo action a `LayerController*` as  well and let it reach in there when it need to. (Seems fine?)

### DIFF
--- a/src/control/layer/LayerController.h
+++ b/src/control/layer/LayerController.h
@@ -112,4 +112,6 @@ private:
     std::list<LayerCtrlListener*> listener;
 
     size_t selectedPage;
+
+    friend class LayerRenameUndoAction;
 };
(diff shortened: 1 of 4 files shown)


rolandlo (approved):
LGTM. I don't see any downside of your approach.

PR #3299, opened 2021-08-03
Release Helper Scripts

LittleHuba:
We will switch to git-flow with modified branches. Active development will continue on _master_ and releases will be published to the branch _releases_. You can read up on git-flow here: https://nvie.com/posts/a-successful-git-branching-model/

This PR introduces various helper scripts for preparing and publishing new releases.

They enforce git-flow and should be somewhat resilient to mistakes. I wrote them in bash, only using what should be commonly available on Linux-based OSs. This way no prerequisites are required (Please tell me otherwise if I missed anything).

None of the scripts push anything to the remote repository, so feel free to try them locally and tell me if anything does not work as expected.

I also fixed the versioning scheme of the releases from our unstable PPA to be in accordance with the rules from Debian.

What the scripts do:
- **validate_version.sh**
    Validates a provided version to be higher than the current version in CMakeLists.txt The comparison follows the same procedure as is applied in package-managers
- **bump_version.sh**
    Updates the version strings in all relevant files and adds the necessary version information to the changelogs. The summary of changes must be supplied manually after this script was executed. Changes to the version strings or changelogs are not committed.
- **prepare_release.sh**
    Creates a release branch and bumps the version within this release branch. Makes sure that no release branch for the supplied version exists or that the version was already released. Ensures that it is executed from _master_.
- **prepare_hotfix.sh**
    Creates a hotfix branch from a prior release and bumps the version within this hotfix branch. Makes sure that no hotfix branch for the supplied version exists or that the hotfix was already released. Ensures that it is executed from _master_.
- **publish_release.sh**
    Merges the branch with _releases_ and creates the appropriate tag for the release. Makes sure that it is executed from a release-/hotfix-branch and that the release was not previously published. Also starts the merge into _master_ if requested by the user.
    
Should you miss a script, please tell me about it.


LittleHuba:
I will update the pipelines once the scripts have been merged. Additionally, I will introduce a new pipeline that publishes nightly builds of active release branches. These releases can be seen as beta and will stop once the release is published and the release-branch is deleted.

Technius:
When you update the pipeline, can you also change the tag format so that releases are prefixed with a `v` (e.g. `v1.1.1`)? This is to help emphasize that the version number refers to a specific tag.

LittleHuba:
> ... can you also change the tag format so that releases are prefixed with a v (e.g. v1.1.1)?

I could do that. But then we should update the existing tokens as well. Otherwise we have a mixed scheme which is awful.

Technius:
If we update the existing tags, we might break something downstream though.

LittleHuba:
Same happens if we change the tags for the coming releases. As we only tag releases anyway, I do not actually see a problem with just using the version tag without the preceeding 'v'.

scripts/bump_version.sh
```diff
@@ -0,0 +1,63 @@
...
+
+####################
+# Bump version
+####################
+
+# Update version in the CMakeLists.txt
+sed -i "s/set (CPACK_PACKAGE_VERSION_MAJOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MAJOR \"$MAJOR_VERSION\")/g" $SCRIPT_PATH/../CMakeLists.txt
+sed -i "s/set (CPACK_PACKAGE_VERSION_MINOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MINOR \"$MINOR_VERSION\")/g" $SCRIPT_PATH/../CMakeLists.txt
+sed -i "s/set (CPACK_PACKAGE_VERSION_PATCH \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_PATCH \"$PATCH_VERSION\")/g" $SCRIPT_PATH/../CMakeLists.txt
+sed -i "s/set (VERSION_SUFFIX \"[^\"]*\")/set (VERSION_SUFFIX \"$VERSION_SUFFIX\")/g" ${SCRIPT_PATH}/../CMakeLists.txt
```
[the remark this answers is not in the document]
Technius: We set `PROJECT_VERSION` using `CPACK_PACKAGE_VERSION_*`. https://github.com/xournalpp/xournalpp/blob/fffaa72e432a529cee7d343f0e337f491d98b431/CMakeLists.txt#L12

This seems appropriate to me since we actually care about the major/minor/patch versions.

scripts/bump_version.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Update Changelog
+if ! grep -Fxq "## $1" $SCRIPT_PATH/../CHANGELOG.md; then
+    sed -i "N;N;s/# Changelog\n\n## Nightly (Unreleased)/# Changelog\n\n## Nightly (Unreleased)\n\n## $VERSION\n\nTODO/g" $SCRIPT_PATH/../CHANGELOG.md
+    echo "WARNING: Please make sure to update CHANGELOG.md"
+else
+    echo "ERROR: CHANGELOG.md already contains this version number"
```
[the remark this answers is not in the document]
Technius: The way I understand it:
* When publishing a release, it may not be listed in the changelog, so we emit a warning.
* If we publish a release but it is already in the changelog, it has likely to have already been published. So raise an error.

I think it might make more sense to make both of them errors, so that a release _must_ be accompanied by a changelog update.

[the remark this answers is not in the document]
personalizedrefrigerator: It might also make sense to `exit 1` at the end of this script if we encounter an error. E.g.

```sh
EXIT_STATUS=0

displayError() {
    echo "ERROR: $1"
    EXIT_STATUS=1
}

...

else
    displayError "Some error"
fi

exit $EXIT_STATUS
```

> I think it might make more sense to make both of them errors, so that a release must be accompanied by a changelog update.

I think this makes sense. In this case, however, the above suggestion (`exit $EXIT_STATUS`) could be replaced with `exit 1` and wouldn't be particularly useful.

scripts/prepare_release.sh
```diff
@@ -0,0 +1,76 @@
...
+    exit 1
+fi
+
+# Check if branch already exists
+if git rev-parse --quiet --verify release-$1 > /dev/null; then
+    echo "ABORT: Release branch \`release-$1\` already exists"
+fi
+
+if git tag | grep -Fxq "$1"; then
+    echo "This version was already released"
```
personalizedrefrigerator, line 47: Consider having `ERROR: ` prefix error messages (as above).

personalizedrefrigerator (commented):

scripts/bump_version.sh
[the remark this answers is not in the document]
Febbe: Since the behaviour of CPack changed to:
[```
CPACK_PACKAGE_VERSION_MAJOR

    Package major version. This variable will always be set, but its default value depends on whether or not version details were given to the project() command in the top level CMakeLists.txt file. If version details were given, the default value will be CMAKE_PROJECT_VERSION_MAJOR. If no version details were given, a default version of 0.1.1 will be assumed, leading to CPACK_PACKAGE_VERSION_MAJOR having a default value of 0.

```](https://cmake.org/cmake/help/v3.21/module/CPack.html#variable:CPACK_PACKAGE_VERSION_MAJOR)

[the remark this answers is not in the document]
Comment (LittleHuba) [reply scripts/bump_version.sh:30]: As long as the version is written somewhere I can parse it from...

I will update this logic once #2570 is merged.

File: scripts/bump_version.sh (hunk 2 above)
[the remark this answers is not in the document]
Comment (LittleHuba) [reply scripts/bump_version.sh:37]: I do not see the changelog as a source of truth. Instead the major scripts use git tags and/or CMake to validate the version string. 

The reasoning behind this is, that the changelog must be updated manually and therefore somebody might choose to do this prematurely. In either case, a manual check of the changelog is required in all scenarios, so this should be a warning after all and no error.

**Comment by LittleHuba** [comment]
I merged all the scripts into one script. Because it is quite late at my place, I did not have the time to check for functionality yet. So expect this to not work yet. You can look through the script and check for obvious problems though. I'll do testing and debugging tomorrow.

@mjg please also take a look, so we can be sure that the new versioning scheme is actually correct and follows your proposal.

**Comment by LittleHuba** [comment]
Of course, I'll squash this before merging.

**Comment by mjg** [comment]
> I merged all the scripts into one script. Because it is quite late at my place, I did not have the time to check for functionality yet. So expect this to not work yet. You can look through the script and check for obvious problems though. I'll do testing and debugging tomorrow.
> 
> @mjg please also take a look, so we can be sure that the new versioning scheme is actually correct and follows your proposal.

Correctness is in the eye of the beholder :)

It seems you are mostly following "git-flow" with their "master" named "releases" here and their "develop" named "master" here. In that scheme, I never quite understood the role of the "releases" banch, to be honest: it appears to merely keep track of release tags in a linear order which seizes to make sense once there are hotfixes or backports for releases other than the last one, so that you can't keep those there. Github's Release feature might be a substitute, but I dunno.

As for the versionning scheme: It's a bit hard to tell from the script alone because the result depends on the input to the script. Is it supposed to verify its arguments strictly, i.e. should I throw arbitrary version numbers on it (being on arbitrary branches) and the script either bails out or does the right thing?

**Comment by LittleHuba** [comment]
> I never quite understood the role of the "releases" banch

This branch exists because release branches are not retained indefinitely in git-flow. Thus, if you tag your release you can not do this on the release branch. Now you could think: "Why not place the tag on `master` then? When we merge in the release?" But assume you previously merged new features to `master`. Then the merge commit will not be suitable for your tag.

> it appears to merely keep track of release tags in a linear order which seizes to make sense once there are hotfixes or backports for releases other than the last one, so that you can't keep those there

This is a valid concern. Seems like we must diverge from git-flow further than I thought...

> Is it supposed to verify its arguments strictly

Yes, it is supposed to verify the provided arguments strictly. It would be better if you take a look at the script and try to understand it. Though I do not know how good you are with bash scripting.

**Comment by Febbe** [comment]
Back porting won't be an option in the future and hot fixes are merged into releases.

**Comment by mjg** [comment]
> Yes, it is supposed to verify the provided arguments strictly. It would be better if you take a look at the script and try to understand it. Though I do not know how good you are with bash scripting.

I don't know. But it does not run and I can at least fix it to run (a missing `)`, misnamed function) without arguments.

Still it would help to know the expected outcome. For example, in detached state,
```
if [[ $(git branch --contains HEAD | grep -Fxq "releases") ]]
```
matches only if the current `HEAD` is contained in the branch `releases`, which fails as soon as you put a commit on top. Is that intended, or was this supposed to be the other way round?

**Comment by LittleHuba** [comment]
> Back porting won't be an option in the future and hot fixes are merged into the release-branch. So every release will be in the history of release.

It should still be an option to publish a release 1.1.5 while we have an unstable major version 2.0.0 released. Seems like this could come up in the future if we make some major changes like a completely new file format.

**Comment by LittleHuba** [comment]
@mjg I should have fixed those bugs now. As I mentioned, I did not have the time yesterday. I'll start with some testing myself now.

**Comment by mjg** [comment]
`is_detached` is backwards, I think.

You can define your return codes any way you like, of course. But deviating from the standard is a typical source of confusion (which seems to have happened here). In particular, return codes (`$?`) are not returned as results by something like `grep -q` (it never returns a result), so you can't compare its result like you do in `is_detached()`.

The standard is so that logical chaining works, e.g. `is_detached && echo "watch your HEAD"` should work, and
```
if [[ $(is_detached) -eq 0 ]]; then
```
should just be
```
if is_detached; then
```
which requires no knowledge about special return values of the function (and assumes it follows the standard).

**Comment by Febbe** [comment]
@LittleHuba yes, but only if the desired way of releasing the fix in 1.5.x and then merging it into 2.x is not possible for some reason.

**Comment by LittleHuba** [comment]
> @LittleHuba yes, but only if the desired way of releasing the fix in 1.5.x and then merging it into 2.x is not possible for some reason.

But that is exactly the point @mjg made. With git flow you cant release a 1.5.x once you released any version that is higher, as you will be unable to merge 1.5.x into the releases branch (merge conflicts).

**Comment by Febbe** [comment]
I think merge conflicts are ok and should be solved, but mistakes, like merging a bugfix first into dev is problematic.

**Comment by Technius** [comment]
Why not [GitLab flow](https://docs.gitlab.com/ee/topics/gitlab_flow.html#release-branches-with-gitlab-flow), which addresses the release and hotfix problem?

**Comment by LittleHuba** [comment]
Quite interesting. Gilab flow is in essence git flow but with release branches that remain forever and no releases branch. Also that is what I ended up with.

@Febbe you can not merge those conflicts as you would then modify your release. The commits must apply without conflict.

**Comment by LittleHuba** [comment]
This state should now fix the prepare phase from the main development branch completely. I also updated the changelogs to have the correct state.
Do not expect the publish phase to work yet or the prepare phase from published releases.

**Comment by Febbe** [comment]
@LittleHuba why it should modify the release? The bugfix is applied on the release-branch and then merged into the feature-2.0 branch. There is no conflict after an release. Even if, just release a new release after the merge.

**Comment by LittleHuba** [comment]
Okay, this is ready for testing now. All functionality works on my side.

You can follow these steps to see the full functionality:
1. Merge this branch with `master`
2. Prepare the next minor release with `scripts/release_helper.sh prepare 1.2.0`
3. Publish the minor release 1.2.0 with `scripts/release_helper.sh publish`
4. Try the merge back on `master` (This step is difficult. If you have an idea how to simplify this please tell)
5. Check out the published release with `git checkout v1.2.0`
6. Create a hotfix with `scripts/release_helper.sh prepare 1.2.0+test`
7. Publish this hotfix with `scripts/release_helper.sh publish`
8. Check out the release branch for 1.2 releases with `git checkout release-1.2`
9. Publish another minor release 1.2.1 with `scripts/release_helper.sh publish`

That should cover it all. You should take a look at the automatic commits done by the script so that the changes to the files are correct in your opinion. Before we can use this for the 1.1.1 release, I must create the appropriate branch with the correct version information. I'll do so once this is merged.

**Comment by LittleHuba** [comment]
> @LittleHuba why it should modify the release? The bugfix is applied on the release-branch and then merged into the feature-2.0 branch. There is no conflict after an release. Even if, just release a new release after the merge.

I think this is simply a miscommunication. The release must be tagged on the release-branch. Then it is merged back into the main development branch. But the main development branch has different version information because if the release is 1.2.0 then this branch will be 1.3.0~dev. Hence we get a merge conflict.

The main point from @mjg and myself was, that we can not use git-flow in the sense of having a `releases` branch where all the tagged releases are. Because once we publish a version 1.2.0 we could never publish a lower version 1.1.1. Not even a hotfix, as we would then be unable to tag this release on the `releases` branch. That is the case because we can not linearly place the 1.1.1 release after the 1.2.0 release on this branch because 1.1.1 is not based on 1.2.0.

I circumvented this by simply dropping the `releases` branch. There is no real benefit of having it anyway. Just having the tags is absolutely sufficient.

**Comment by Febbe** [comment]
> There is no real benefit of having it anyway. Just having the tags is absolutely sufficient.

Exactly, I also thought this is a misunderstanding of git-flow and that the releases branch was never meant to have all releases in *linear* order. 

> But the main development branch has different version information because if the release is 1.2.0 then this branch will be 1.3.0~dev. Hence we get a merge conflict.

This can only happen, if the advantage of version numbers are done concurrently. And on a version merge conflicts, just choose the version number of the target branch. But this won't happen if a merge only happens to branches with higher version numbers.
E.g. 1.1.1 -> 1.1.2 -> 1.2.1 -> 1.2.2 ->  2.0
Assuming we only support at maximum 2/3 versions it will look like this:

1.1.2 -> 1.2.2 (-> 2.0~dev)
Unmerged feature branches or special dev branches can be rebased on merge or directly, when no one else works on them.

**Comment by Febbe** [comment]
> 4. Try the merge back on master (This step is difficult. If you have an idea how to simplify this please tell)

Replace master with current/original branch and don't merge, just fast forward this branch.

**Comment by LittleHuba** [comment]
> > But the main development branch has different version information because if the release is 1.2.0 then this branch will be 1.3.0~dev. Hence we get a merge conflict.
> 
> This can only happen, if the advantage of version numbers are done concurrently. And on a version merge conflicts, just choose the version number of the target branch. But this won't happen if a merge only happens to branches with higher version numbers.
> E.g. 1.1.1 -> 1.1.2 -> 1.2.1 -> 1.2.2 -> 2.0
> Assuming we only support at maximum 2/3 versions it will look like this:

No this is a common cause. Once a release branch is created for a release `1.2.0` (`release-1.2`) the main development branch is advanced to `1.3.0~dev`. This way it is possible to merge changes from other feature-branches that are not meant for the `1.2.0` release.

**Comment by LittleHuba** [comment]
> > 1. Try the merge back on master (This step is difficult. If you have an idea how to simplify this please tell)
> 
> Replace master with current/original branch and don't merge, just fast forward this branch.

That does not work as soon as there is one single commit on master that is later than the commit where the release branches off. Unfortunately that will be the case every single time because of the versioning change). But even without this commit, it would not work as soon as we merge a feature for the next release to `master`.

**Comment by Febbe** [comment]
> Unfortunately that will be the case every single time because of the versioning change).

I would move the versioning change directly after the release.

**Comment by LittleHuba** [comment]
> I would move the versioning change directly after the release.

If you mean "directly after the release on the release-branch", this does not really work. Since then somebody using the nightlies would get features that are intended for the next release as part of a version that is denoted as the development version of this release.

**Comment by LittleHuba** [comment]
The last push fixed behavior when preparing a new release branch that is not the syntactically next minor release (1.1.0->1.2.0).

**Comment by LittleHuba** [comment]
Some details on the workings of this script:

- `prepare 1.2.0` on `master`
  1. Create the release branch `release-1.2`
  2. Bump the version of `master` to `1.3.0~dev`
  3. Bump the version of `release-1.2` to `1.2.0~dev`
- `publish` on branch `release-1.2`
  1. Bump the version from `1.2.0~dev` to `1.2.0`
  2. Tag the release as `v1.2.0`
  3. Bump the version to `1.2.1~dev`
  4. Propose the merge of the release commit back to `master`
- `prepare 1.2.0+test` on commit of release `v1.2.0`
  1. Create branch `hotfix-1.2.0+test`
  2. Bump version to `1.2.0+test~dev`
- `publish` on branch `hotfix-1.2.0+test`
  1. Bump the version from `1.2.0+test~dev` to `1.2.0+test`
  2. Tag the release as `v1.2.0+test`
  3. Propose the merge of the release commit back to `master`

This setup allows us to merge any features that are not meant for this minor release directly after the minor release was branched off. This gives us maximum time for testing and the best possible parallelization of the development process. The drawback is that the merging of the release branch is not overly simple.

**Comment by Febbe** [comment]
I thought ~dev is meant as postfix to symbolize a "is derived from" relationship. This looks now like a "prepares" relationship.

**Comment by LittleHuba** [comment]
> I thought ~dev is meant as postfix to symbolize a "is derived from" relationship. This looks now like a "prepares" relationship.
> 
> So point 1 should be moved after point 2 for each publish type.

What you mean would be +dev. ~dev is "working towards". Maybe we should use the +dev on the `master` branch so we save the version bump there? On the release branches using ~dev makes more sense as we specifically work towards a release.

**Comment by Febbe** [comment]
In the future, I would derive the version completely from the last tags on the current branch.
Without manual version bumping.

**Comment by LittleHuba** [comment]
In the future, I would derive the version completely from the last tags on the current branch.
Without manual version bumping.

That is impossible because we have changelog files that require to have the version written into them. So we will have to bump them manually even if the rest of the system derives it from the tags.

**Comment by Febbe** [comment]
Ok yep. That's true. But it is possible to verify the manual bumping by hand.
But back to topic, what's missing to merge this?

**Comment by LittleHuba** [comment]
In essence some testing by others to catch cases I might have missed. Also a final decision on `+dev` or `~dev` versioning scheme.

Then this should be good to go.

**Comment by LittleHuba** [comment]
Okay, this should be good enough for testing. I had to do a little bit of magic in the background to get the merging of releases back to the main development branch right. The benefit is, that we now have validation that the versions are not messed up during merging.

On the main development branch, we now work with `+dev` suffix and on release branches with `~dev`.

**Comment by rolandlo** [comment]
I followed https://github.com/xournalpp/xournalpp/pull/3299#issuecomment-909578538, but got some error messages while merging back.

```
roland@roland-ThinkPad-X1-Yoga-3rd:~/Dokumente/Programmierung/Reviews/xournalpp$ scripts/release_helper.sh publish
Are you sure you want to publish release 1.2.0? [y/N] y
[release-1.2 a70af0126] Release 1.2.0
 4 files changed, 5 insertions(+), 5 deletions(-)
Tagging the release
[release-1.2 29e5751a9] Automated version bump to 1.2.1~dev
 4 files changed, 11 insertions(+), 2 deletions(-)
SUCCESS: Release was published locally!
To publish the release globally push your changes with:

    git push --follow-tags origin release-1.2

Do you want to merge the release back to the main development branch now? [Y/n] Y

Be careful while patching! Read these instructions carefully before starting the merge:
- Merge your version as if it will be the new version of the main development branch.
  The version will be modified to a development version automatically!
Start the merging process? [y/N] y

.git/hooks/commit-msg: 3: source: not found
.git/hooks/commit-msg: 9: current_version: not found
.git/hooks/commit-msg: 9: [[: not found
The version information in CMakeLists.txt is not correct. The current version should be 1.2.0
Merge wurde nicht committet; benutzen Sie 'git commit', um den Merge abzuschließen.
```

Do I need to remove `.git/hooks/commit-msg` after preparing the release?

**Comment by LittleHuba** [comment]
Did you merge the helper script branch into the main development branch before following the steps?

**Comment by rolandlo** [comment]
> Did you merge the helper script branch into the main development branch before following the steps?

I merged the helper script branch into `master` as step 1, yes. Output of `git log` (on `master` branch):

```
ac0688037a2b234e848ce07d754a722bf4143983 (HEAD -> master, origin/master, release-helper-scripts) Create release helper script
a5904ac058819b82cdab54ff378d601fe97712ae Create proper initial status
a9117fab3ccd5b30d52b03480dce3baba5b91023 Update and fix the version of the debian release
466082db423f8a5abeb9bf1f46f2e52cf1a7c5ba (upstream/master) Fix typo from 'orignal' to 'original'
ab57ddda2489182039b4f267fc194dfe42bdccf8 Remove horizontal and vertical space compensation
```

**Comment by LittleHuba** [comment]
Could you remove the git hook manually and then add a comand to output the current directory in line 392?

Maybe your git version executes the git hooks within another directory?

**Comment by rolandlo** [comment]
> Could you remove the git hook manually and then add a comand to output the current directory in line 392?
> 
> Maybe your git version executes the git hooks within another directory?

I removed the `commit-msg` hook and added the line
```
echo $(pwd)
```

Now the output is:
```
Start the merging process? [y/N] y

/home/roland/Dokumente/Programmierung/Reviews/xournalpp
.git/hooks/commit-msg: 5: source: not found
.git/hooks/commit-msg: 11: current_version: not found
.git/hooks/commit-msg: 11: [[: not found
The version information in CMakeLists.txt is not correct. The current version should be 1.2.0
Merge wurde nicht committet; benutzen Sie 'git commit', um den Merge abzuschließen.
```

The directory looks correct.

**Comment by rolandlo** [comment]
> `prepare 1.2.0` on `master`
> 
>     1. Create the release branch `release-1.2`
> 
>     2. Bump the version of `master` to `1.3.0~dev`
> 
>     3. Bump the version of `release-1.2` to `1.2.0~dev`

I don't see the bump of the `master` branch to `1.3.0~dev` (neither from the bash script nor from inspecting the master branch). Is this information outdated?

The problem I experience is still the same.

**Comment by LittleHuba** [comment]
That part is outdated. We now use a +dev versioning scheme on master. So the version stays the same when creating a release branch. On the release branch we use a ~dev versioning scheme to signify the active work towards a specific release.

Could you post the git hook that was created?

**Comment by rolandlo** [comment]
> That part is outdated. We now use a +dev versioning scheme on master. So the version stays the same when creating a release branch. On the release branch we use a ~dev versioning scheme to signify the active work towards a specific release.
> 
> Could you post the git hook that was created?

Here it is 
[commit-msg.zip](https://github.com/xournalpp/xournalpp/files/7153648/commit-msg.zip)

**Comment by LittleHuba** [comment]
Okay your git hooks were not executed as bash script. Give it another try please.

**Comment by rolandlo** [comment]
> Okay your git hooks were not executed as bash script. Give it another try please.

The issue is fixed indeed. Now when following all steps from  https://github.com/xournalpp/xournalpp/pull/3299#issuecomment-909578538 I get merge conflicts in step 7
> Publish this hotfix with `scripts/release_helper.sh publish`

in all files that have been modified. For example in `CHANGELOG.md` I have 
`1.2.0+dev (Unreleased)` as Current Changes and
`1.2.0+test` as Incoming Changes.

Did I do something wrong?

**Comment by LittleHuba** [comment]
> Now when following all steps from [#3299 (comment)](https://github.com/xournalpp/xournalpp/pull/3299#issuecomment-909578538) I get merge conflicts in step 7
> 
> > Publish this hotfix with `scripts/release_helper.sh publish`
> 
> in all files that have been modified. For example in `CHANGELOG.md` I have
> `1.2.0+dev (Unreleased)` as Current Changes and
> `1.2.0+test` as Incoming Changes.
> 
> Did I do something wrong?

No this behavior is as expected. When the release `1.2.0` was created, there was no notion of `1.2.0+dev`. As the hotfix release `1.2.0+test` is based upon `1.2.0` you now have two release versions following the base version `1.2.0`. Since Git stores modifications based on diffs, you get two diffs for the same lines of the relevant files. 

What you will have to do is merge the set of changes. You'll want the development version `1.2.0+dev` at the start of the file, followed by `1.2.0+test` which is followed by `1.2.0`. I'll improve the instructions the script provides for you before starting the merge.

**Comment by Technius** [comment]
Just a minor comment: I think it would be a good idea to add a `readme/releases.md` that summarizes 1) how the publish process works and 2) how the versioning scheme works.

**Comment by LittleHuba** [comment]
> Just a minor comment: I think it would be a good idea to add a `readme/releases.md` that summarizes 1) how the publish process works and 2) how the versioning scheme works.

Yes, that is something I will add.

**Comment by LittleHuba** [comment]
Please give the readme file a once-over and comment on places I should improve.

I'll do so in a follow-up PR. I don't think this should delay the progress of others any more than it already has.
The script should work now as expected. If there are no complaints within the next 48 hours, I'll go ahead and merge it.

File: scripts/release_helper.sh [hunk 4]
```diff
@@ -0,0 +1,601 @@
+#!/bin/bash
+
+# Get the path of the script
+SCRIPT_PATH=$(dirname $(realpath -s $0))
```
Technius, line 4:
```suggestion
SCRIPT_PATH=$(dirname "$(realpath -s "$0")")
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+
+# Parse current version string
+function current_major_version() {
+    echo $(sed -n 's/^set (CPACK_PACKAGE_VERSION_MAJOR "\([0-9]\+\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt)
```
Technius, line 8:
```suggestion
    sed -n 's/^set (CPACK_PACKAGE_VERSION_MAJOR "\([0-9]\+\)")/\1/p' "${SCRIPT_PATH}"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+}
+
+function current_minor_version() {
+    echo $(sed -n 's/^set (CPACK_PACKAGE_VERSION_MINOR "\([0-9]\+\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt)
```
Technius, line 12:
```suggestion
    sed -n 's/^set (CPACK_PACKAGE_VERSION_MINOR "\([0-9]\+\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+}
+
+function current_patch_version() {
+    echo $(sed -n 's/^set (CPACK_PACKAGE_VERSION_PATCH "\([0-9]\+\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt)
```
Technius, line 16:
```suggestion
    sed -n 's/^set (CPACK_PACKAGE_VERSION_PATCH "\([0-9]\+\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt
```

scripts/release_helper.sh
Technius: Without this, a space in a directory name will result in the wrong `SCRIPT_PATH`.

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+}
+
+function current_suffix_version() {
+    echo $(sed -n 's/^set (VERSION_SUFFIX "\([+~][^\"]*\)")/\1/p' ${SCRIPT_PATH}/../CMakeLists.txt)
```
Technius, line 20:
```suggestion
    sed -n 's/^set (VERSION_SUFFIX "\([+~][^\"]*\)")/\1/p' "${SCRIPT_PATH}"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    local replace=$2
+
+    if ! [[ "$1" =~ ~dev|\+dev$ ]]; then
+        local publish=1
+    else
+        local publish=0
+    fi
+
+    # Update version in the CMakeLists.txt
+    sed -i "s/set (CPACK_PACKAGE_VERSION_MAJOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MAJOR \"$(parse_major_version $1)\")/g" $SCRIPT_PATH/../CMakeLists.txt
```
Technius, line 128:
```suggestion
    sed -i "s/set (CPACK_PACKAGE_VERSION_MAJOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MAJOR \"$(parse_major_version $1)\")/g" "$SCRIPT_PATH"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    sed -i "s/set (CPACK_PACKAGE_VERSION_MINOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MINOR \"$(parse_minor_version $1)\")/g" $SCRIPT_PATH/../CMakeLists.txt
```
Technius, line 129:
```suggestion
    sed -i "s/set (CPACK_PACKAGE_VERSION_MINOR \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_MINOR \"$(parse_minor_version $1)\")/g" "$SCRIPT_PATH"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    sed -i "s/set (CPACK_PACKAGE_VERSION_PATCH \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_PATCH \"$(parse_patch_version $1)\")/g" $SCRIPT_PATH/../CMakeLists.txt
```
Technius, line 130:
```suggestion
    sed -i "s/set (CPACK_PACKAGE_VERSION_PATCH \"[0-9]\+\")/set (CPACK_PACKAGE_VERSION_PATCH \"$(parse_patch_version $1)\")/g" "$SCRIPT_PATH"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    sed -i "s/set (VERSION_SUFFIX \"[^\"]*\")/set (VERSION_SUFFIX \"$(parse_suffix_version $1)\")/g" ${SCRIPT_PATH}/../CMakeLists.txt
```
Technius, line 131:
```suggestion
    sed -i "s/set (VERSION_SUFFIX \"[^\"]*\")/set (VERSION_SUFFIX \"$(parse_suffix_version $1)\")/g" "${SCRIPT_PATH}"/../CMakeLists.txt
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+
+    # From now on current_*_version functions return the new version
+
+    # Update Changelog
+    if [ $replace -eq 0 ]; then
+        sed -i "N;N;s/# Changelog\n\n/# Changelog\n\n## $(current_version) (Unreleased)\n\n/g" $SCRIPT_PATH/../CHANGELOG.md
```
Technius, line 137:
```suggestion
        sed -i "N;N;s/# Changelog\n\n/# Changelog\n\n## $(current_version) (Unreleased)\n\n/g" "$SCRIPT_PATH"/../CHANGELOG.md
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    else
+        if [ $publish -eq 0 ]; then
+            sed -i "s/## ${prior_version} (Unreleased)/## $(current_version) (Unreleased)/g" $SCRIPT_PATH/../CHANGELOG.md
```
Technius, line 140:
```suggestion
            sed -i "s/## ${prior_version} (Unreleased)/## $(current_version) (Unreleased)/g" "$SCRIPT_PATH"/../CHANGELOG.md
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+        else
+            sed -i "s/## ${prior_version} (Unreleased)/## $(current_version)/g" $SCRIPT_PATH/../CHANGELOG.md
```
Technius, line 142:
```suggestion
            sed -i "s/## ${prior_version} (Unreleased)/## $(current_version)/g" "$SCRIPT_PATH"/../CHANGELOG.md
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+        abort 2 "Could not read current git user - Please set with 'git config --set user.name <name>'"
+    fi
+    local git_mail=$(git config --get user.email)
+    if [ $? -ne 0 ]; then
+        abort 2 "Could not read current git user email - Please set with 'git config --set user.email <email>'"
+    fi
+    local date=$(date --rfc-2822)
+
+    if [ $replace -eq 0 ]; then
+        sed -i "1i xournalpp ($(current_version)-1) UNRELEASED; urgency=medium\n\n  * \n\n -- ${git_user} <${git_mail}>  ${date}\n" $SCRIPT_PATH/../debian/changelog
```
Technius, line 158:
```suggestion
        sed -i "1i xournalpp ($(current_version)-1) UNRELEASED; urgency=medium\n\n  * \n\n -- ${git_user} <${git_mail}>  ${date}\n" "$SCRIPT_PATH"/../debian/changelog
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    else
+        if [ $publish -eq 0 ]; then
+            sed -i "s/xournalpp (${prior_version}-1) UNRELEASED; urgency=medium/xournalpp ($(current_version)-1) UNRELEASED; urgency=medium/g" $SCRIPT_PATH/../debian/changelog
+            sed -i "/xournalpp ($(current_version)-1)/,/xournalpp/s/ --.*/ -- ${git_user} <${git_mail}>  ${date}/g" $SCRIPT_PATH/../debian/changelog
```
Technius, line 162:
```suggestion
            sed -i "/xournalpp ($(current_version)-1)/,/xournalpp/s/ --.*/ -- ${git_user} <${git_mail}>  ${date}/g" "$SCRIPT_PATH"/../debian/changelog
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    if [ $? -ne 0 ]; then
+        abort 2 "Could not read current git user email - Please set with 'git config --set user.email <email>'"
+    fi
+    local date=$(date --rfc-2822)
+
+    if [ $replace -eq 0 ]; then
+        sed -i "1i xournalpp ($(current_version)-1) UNRELEASED; urgency=medium\n\n  * \n\n -- ${git_user} <${git_mail}>  ${date}\n" $SCRIPT_PATH/../debian/changelog
+    else
+        if [ $publish -eq 0 ]; then
+            sed -i "s/xournalpp (${prior_version}-1) UNRELEASED; urgency=medium/xournalpp ($(current_version)-1) UNRELEASED; urgency=medium/g" $SCRIPT_PATH/../debian/changelog
```
Technius, line 161:
```suggestion
            sed -i "s/xournalpp (${prior_version}-1) UNRELEASED; urgency=medium/xournalpp ($(current_version)-1) UNRELEASED; urgency=medium/g" "$SCRIPT_PATH"/../debian/changelog
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+            sed -i "/xournalpp ($(current_version)-1)/,/xournalpp/s/ --.*/ -- ${git_user} <${git_mail}>  ${date}/g" $SCRIPT_PATH/../debian/changelog
+        else
+            sed -i "s/xournalpp (${prior_version}-1) UNRELEASED; urgency=medium/xournalpp ($(current_version)-1) unstable; urgency=medium/g" $SCRIPT_PATH/../debian/changelog
```
Technius, line 164:
```suggestion
            sed -i "s/xournalpp (${prior_version}-1) UNRELEASED; urgency=medium/xournalpp ($(current_version)-1) unstable; urgency=medium/g" "$SCRIPT_PATH"/../debian/changelog
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+            sed -i "/xournalpp ($(current_version)-1)/,/xournalpp/s/ --.*/ -- ${git_user} <${git_mail}>  ${date}/g" $SCRIPT_PATH/../debian/changelog
```
Technius, line 165:
```suggestion
            sed -i "/xournalpp ($(current_version)-1)/,/xournalpp/s/ --.*/ -- ${git_user} <${git_mail}>  ${date}/g" "$SCRIPT_PATH"/../debian/changelog
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+        fi
+    fi
+
+    # Update Appdata
+    local date=$(date +%Y-%m-%d)
+
+    if [ $replace -eq 0 ]; then
+        sed -i "1,/^    <release .*$/ {/^    <release .*$/i\
+        \ \ \ \ <release date=\"$date\" version=\"$(current_version)\" />
+        }" $SCRIPT_PATH/../desktop/com.github.xournalpp.xournalpp.appdata.xml
```
Technius, line 175:
```suggestion
        }" "$SCRIPT_PATH"/../desktop/com.github.xournalpp.xournalpp.appdata.xml
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+    else
+        sed -i "s/\ \ \ \ <release date=\".*\" version=\"${prior_version}\" \/>/\ \ \ \ <release date=\"$date\" version=\"$(current_version)\" \/>/g" $SCRIPT_PATH/../desktop/com.github.xournalpp.xournalpp.appdata.xml
```
Technius, line 177:
```suggestion
        sed -i "s/\ \ \ \ <release date=\".*\" version=\"${prior_version}\" \/>/\ \ \ \ <release date=\"$date\" version=\"$(current_version)\" \/>/g" "$SCRIPT_PATH"/../desktop/com.github.xournalpp.xournalpp.appdata.xml
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+                    abort 7 "Could not checkout release branch. BEWARE you are on 'master' now!"
+                fi
+                exit 0
+            fi
+
+            echo ""
+
+            # These git hooks will be called before the merge commit is created. They will take care of bumping the version to the correct value.
+            # Create the commit-msg hook
+            cat << EOF > $SCRIPT_PATH/../.git/hooks/commit-msg
```
Technius, line 391:
```suggestion
            cat << EOF > "$SCRIPT_PATH"/../.git/hooks/commit-msg
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+
+# Remove the git hooks
+rm .git/hooks/commit-msg
+
+# Signify git that it can proceed with the merge
+exit 0
+EOF
+
+            # The post-commit hook must be executable
+            chmod +x $SCRIPT_PATH/../.git/hooks/commit-msg
```
Technius, line 440:
```suggestion
            chmod +x "$SCRIPT_PATH"/../.git/hooks/commit-msg
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+                    abort 7 "Could not checkout release branch. BEWARE you are on 'master' now!"
+                fi
+                exit 0
+            fi
+
+            echo ""
+
+            # These git hooks will be called before the merge commit is created. They will take care of bumping the version to the correct value.
+            # Enable the pre-merge-commit hook as it will call the pre-commit hook
+            mv $SCRIPT_PATH/../.git/hooks/pre-merge-commit.sample $SCRIPT_PATH/../.git/hooks/pre-merge-commit
```
Technius, line 465:
```suggestion
            mv "$SCRIPT_PATH"/../.git/hooks/pre-merge-commit.sample $SCRIPT_PATH/../.git/hooks/pre-merge-commit
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+            # Create the pre-commit hook
+            cat << EOF > $SCRIPT_PATH/../.git/hooks/pre-commit
```
Technius, line 467:
```suggestion
            cat << EOF > "$SCRIPT_PATH"/../.git/hooks/pre-commit
```

scripts/release_helper.sh
```diff
@@ -0,0 +1,601 @@
...
+# Remove the git hooks
+mv .git/hooks/pre-merge-commit .git/hooks/pre-merge-commit.sample
+rm .git/hooks/pre-commit
+
+# Signify git that it can proceed with the merge
+exit 0
+EOF
+
+            # The pre-commit hook must be executable
+            chmod +x $SCRIPT_PATH/../.git/hooks/pre-commit
```
Technius, line 508:
```suggestion
            chmod +x "$SCRIPT_PATH"/../.git/hooks/pre-commit
```

Technius (commented):
LGTM.

Shellcheck raises a lot of warnings. Here are the warnings that I think are relevant:

LittleHuba:
If there are no objections in the next 24 hours, I'll go ahead and merge this.

LittleHuba:
Steps to take when merging this:

- [ ] Merge the PR
- [ ] Branch off `release-1.1`
- [ ] Bump version on `release-1.1` to `1.1.1~dev`
- [ ] Update the Release Pipeline to only work on `release-*` and `hotfix-*` branches
- [ ] Decide whether we want to provide `stable-dev` releases that are nightlies from the latest `release-*` branch

PR #3306, opened 2021-08-06
Preparation for GTK4

Febbe:
@reviewers, please review commit by commit.

https://docs.gtk.org/gtk4/migrating-3to4.html

 - [x] Do not use deprecated symbols
 - [ ] Enable diagnostic warnings
 - [x] Do not use GTK-specific command line arguments
 - [ ] Do not use widget style properties
 - [x] Review your window creation flags
 - [ ] Stop using direct access to GdkEvent structs
 - [x] Stop using `gdk_pointer_warp()`
 - [x] Stop using non-RGBA visuals
 - [x] Stop using `gtk_widget_set_app_paintable`
 - [x] Stop using `GtkBox` padding, fill and expand child properties
 - [x] Stop using the state argument of `GtkStyleContext` getters
 - [x] Stop using `gdk_pixbuf_get_from_window()` and `gdk_cairo_set_source_window()`
 - [ ] Stop using `GtkButton`'s image-related API
 - [ ] Stop using `GtkWidget` event signals
 - [x] Set a proper application ID
 - [x] Stop using `gtk_main()` and related APIs
 - [ ] Reduce the use of `gtk_widget_destroy()`
 - [x] Stop using the GtkWidget.destroy vfunc
 - [ ] Reduce the use of generic container APIs
 - [ ] Review your use of icon resources
 
 ==== HArder to fullfill, since the API is not fully replaced ====
 
  - [x] Stop using GdkScreen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON) # enable clang tidy to use correct include
 set_directory_properties(PROPERTIES EP_BASE "${PROJECT_BINARY_DIR}/external")
 
 add_definitions(-D_USE_MATH_DEFINES)
+add_definitions(-DGDK_DISABLE_DEPRECATED -DGTK_DISABLE_DEPRECATED)
 
 # package version
 include(Version)
--- a/src/core/control/ClipboardHandler.cpp
+++ b/src/core/control/ClipboardHandler.cpp
@@ -245,7 +245,7 @@ void ClipboardHandler::setCopyPasteEnabled(bool enabled) {
 }
 
 void ClipboardHandler::ownerChangedCallback(GtkClipboard* clip, GdkEvent* event, ClipboardHandler* handler) {
-    if (event->type == GDK_OWNER_CHANGE) {
+    if (gdk_event_get_event_type(event) == GDK_OWNER_CHANGE) {
         handler->clipboardUpdated(event->owner_change.selection);
     }
 }
--- a/src/core/control/RecentManager.cpp
+++ b/src/core/control/RecentManager.cpp
@@ -10,11 +10,9 @@
 #include "util/i18n.h"
 #include "util/safe_casts.h"
 
-#define MIME "application/x-xoj"
-#define MIME_PDF "application/x-pdf"
-#define GROUP "xournal++"
-
-using std::string;
+constexpr auto const* MIME = "application/x-xoj";
+constexpr auto const* MIME_PDF = "application/x-pdf";
+constexpr auto const* GROUP = "xournal++";
 
 namespace {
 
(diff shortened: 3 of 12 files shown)


rolandlo:
>     * [ ]  Stop using `gdk_pixbuf_get_from_window()` and `gdk_cairo_set_source_window()`

We already don't use it. There is only a comment about using `gdk_pixbuf_get_from_window()`in `pixbuf-utils.cpp`

Febbe:
> > ```
> > * [ ]  Stop using `gdk_pixbuf_get_from_window()` and `gdk_cairo_set_source_window()`
> > ```
> 
> We already don't use it. There is only a comment about using `gdk_pixbuf_get_from_window()`in `pixbuf-utils.cpp`

Right, I also stumbled over this, maybe mark this after I removed the comment.

Febbe:
> Stop using direct access to GdkEvent structs

That will be very hard, since the new input system hardly relies on that.
@LittleHuba any suggestions how to fix that?
We basically must(?) write own GDKInputHandlers for that.

LittleHuba:
As I see it, this only means that we must use getters to retrieve the information from GdkEvent structs. The input system is rigged to move all data from these GdkEvent structs to our own struct. All our input handlers work on our own structs. So we would only need to make sure that this conversion uses the appropriate getters.

See https://github.com/xournalpp/xournalpp/blob/3fe13126e893f88f461cf48cf891f74654b73b63/src/gui/inputdevices/InputEvents.cpp#L77

Febbe:
@LittleHuba oh meant actually this: `Stop using GtkWidget event signals` Gtk4 completely changed the event system / handling.

LittleHuba:
The description here does not sound that different from what we currently do...
https://docs.gtk.org/gtk4/input-handling.html

LittleHuba:
Could you provide me with some resources on the changes that cause your concerns?

Febbe:
That the way how we should do it, is to write our own eventControllers, if the current ones don't suffice. But the only way to port the NewInput System is to connect it to a LegacyInputController. Dunno, which implications this has.

LittleHuba:
Can't we reuse the same concept we currently use and simply forward all events from a new eventController to our input system? As the events are propagated from parent to child, this controller would sit on the widget of the drawing area. Same as before as far as I remember.

Febbe:
Yes, that's the same as using a LegacyInputController. But it is very dirty. Most of the things we do in the input system and what we also planned to do in the future, is already part of the new EventControler system. So we should transform the input system to adapt those.

LittleHuba:
In that case, I would go with the LegacyInputController for now and switch to the new concept in a follow-up PR. Otherwise, this PR will grow very big and stop progress for other PRs as it influences many parts of the source.

Febbe:
I think this PR is already ready to merge, all other changes will require a switch to gtk4 directly.

Technius:
Needs to be rebased.

Febbe:
@Technius it's rebased, I also rebased the gtk4 branch, but I suppose I broke many things.

Febbe:
@Technius I'll merge this in 24h hours if there are no objections

rolandlo:
On commit 3625e7bc1e5a28bb49d65cbb078682196cae0d9a: There are more `GdkEvents` structs that are accessed directly, e.g. in `SidebarPreviewLayerEntry.cpp`, `SidebarPreviewBaseEntry.cpp`, `SetsquareInputHandler.cpp`,

Febbe:
Yes, that issue is too big for this first step. Also many Event structs aren't even implented in gtk3.

Febbe:
Honestly it would be better to rewrite the application from scratch than porting it😔

LittleHuba:
> Honestly it would be better to rewrite the application from scratch than porting it😔

We can always consider restarting on the GUI. The file format and input handling are quite decoupled from it.

Febbe:
> input handling are quite decoupled from it
We also have to restructure the input handling, since gtk3 events are deprecated and should be replaced by Event controllers.
But this might be easier than I expect.

PR #3310, opened 2021-08-06
bump version for the devel branch

mjg:
Nightly and local builds should classify as "more recent" than the last
released version, both for clarity and for updates from ppa/copr for
those who enable nightlty/devel repositories.

So, bump the minor (non-patch) level and append "+dev" as usual.

Note:
I'm not sure I caught all places which need a bump. In particular, `po/xournalpp.pot` might get a bump from somewhere else (or not).


LittleHuba:
I'm against bumping the version number of the nightlies prematurely.

In general terms it is not decided what the next release will have as version number, before this release is actively prepared (The version is bumped). So increasing the version number of the nightlies should wait as long as we do not commit to this step.

If we bump now to 1.1.1+dev, and later decide that we actually wanted to make the release 1.2.0 the premature version-bump of the nightlies is confusing. Worse is if we bump the nightly to 1.1.1+dev and later release version 1.1.0-debug for whatever reason. Then 1.1.0-debug would be perceived by package managers as lower than all 1.1.1+dev versions. Even those released before 1.1.0-debug.

Now assume there is a user that installed one nightly version manually an is normaly on the stable branch. This use will never receive this additional release even though it might contain a fix he is waiting for.

So in my opinion the version should stay as it is. We can discuss the +dev suffix but that is as far as I would go.

mjg:
Maybe I was not clear enough: versioning the devel version as 1.1.0 after 1.1.0 is released is plainly wrong and leads to problems. Just look at the tags tab here: You find 1.1.0 released and 1.1.0 nightlies, so what is it?

The way in which xournalpp used the "+dev" suffix in the past was not "plus additions", but "this is development leading up to". If master becomes "1.2.0-dev" now then this does not exclude intermediate patch or hotfix versions "1.1.x" at all - in fact, they typically are based off master or carry cherry-picked fixes only, and in both cases the version sorts correctly.

On the other hand, having master versioned 1.1.0 forces all packagers to come up with their own versioning, or else recent devel builds will not update release builds!
[For the proposed PR, there are standard packaging procedures.]

Finally, master was versioned 1.1.0+dev *before* 1.1.0 was released and switched to 1.1.0 when it was released, so it would be really strange to switch back after.

Febbe:
Mhh, in my opinion, it should be less than every future release_tag and also greater than the latest release.
I would propose a nightly counter, either as patch or after `~dev`.
But since cmake does not support dev, maybe we should then use tweaks

The two proposed versions:
`<latest_major>.<latest_minor>.<latest_patch>.<nightly_count_after_latest_release>`
`<latest_major>.<latest_minor>.<latest_patch>~dev-<nightly_count_after_latest_release>`

LittleHuba:
From AUR I know that they have no issue using the versioning scheme we currently deploy. They simply use the long description which follows from the last release (1.1.0-38) where 38 describes the number of commit since the last official release. This seems a way more stable approach than bumping the version number in advance for the preparation of a release where we most often do not know what version the next release will actually be.

LittleHuba:
> Mhh, in my opinion, it should be less than every future release_tag and also greater than the latest release.
> I would propose a nightly counter, either as patch or after `~dev`.
> But since cmake does not support dev, maybe we should then use tweaks
> 
> The two proposed versions:
> `<latest_major>.<latest_minor>.<latest_patch>.<nightly_count_after_latest_release>`
> `<latest_major>.<latest_minor>.<latest_patch>~dev-<nightly_count_after_latest_release>`
> 
> E.g.:
> 
> `1.1.0.64`
> `1.1.0~dev-64`

That would force us to increment the tweak each night. Leading to tons of commits that just increase the tweak by one. This is something package maintainers can solve quite nicely be deriving the package version using `git describe --tags --exclude nightly`.

Febbe:
> That would force us to increment the tweak each night. Leading to tons of commits that just increase the tweak by one. This is something package maintainers can solve quite nicely be deriving the package version using git describe --tags --exclude nightly.

Not, if we derive that value before calling `project`. We can do this also by using `git describe --tags --exclude nightly`

mjg:
Well, the point is what the version in the source means, and that meaning seems to have changed around the release of 1.1.0. Otherwise it could not have been 1.1.0 both before and after the release.

Packagers can work with both meanings, of course. But it has to have a defined meaning.

Packagers don't need in-source bumps for every commit or nightly, all they need is a defined versioning scheme for releases (we have that), and for in between releases which either says "based on" or "leading up to".

Febbe:
**Off-topic:** I would additionally propose, to move every package variable like version, contact, dependencies into an external file, which is read before calling project.
With that, we could implement a dependency management.
And only handle dynamic values in the cmake script.

Febbe:
> "based on"

=> maybe `1.1.0-1~dev` ?

"leading up to"

=> stays `1.1.2~dev[-X]`

**?**

Febbe:
@mjg thanks for rising this discussion. We should really stick to a set standard, but the old worked well for 1.1.0 but to be honest it has problems with a more dynamic release policy. Therefore, we either need to break the old, or modify it to fulfill all arising problems. 

@rolandlo @LittleHuba @Technius  @the_rest_who_cares, what do you say we should do?

LittleHuba:
Yes @mjg the meaning did change. Partially because the bumping of the version was not done completely right. But in hindsight, I do not want to change it back, for the above reasons. We are currently introducing scripts to help us with version bumping, which will keep the meaning stable in the future. Sorry for the inconvenience if we caused any.

From my point of view, the version number should mean "is based upon", because that is the truth after all. Is leading up to, is somewhat missleading if we do not know ourselves to what release we actually work to.

So from my side 1.1.0~dev is correct right now and was incorrect before the release of 1.1.0. Since we release 1.1.1 within the next few days, this should not be a problem for long.

mjg:
So, I did some more log digging (to check my own claims), and it seems that `VERSION_SUFFIX` (`+dev`) is somewhat recent, introduced 93 commits after 1.0.15 and 731 commits before 1.1.0. Before that, the same "problem" existed.

In any case, I think the version in source should indicate whether it is a release or not. Typically, this is done by assigning a version stem (like major.minor) to the development branch and working towards its release (major.minor.0), optionally keeping release branches for patch backporting.

The introduction of 1.1.0+dev seemed like a switch to that theme from the previous one (which was bump on release and keep that version), which is why I suggested this PR.

I guess this was not fully intentional (according to @LittleHuba 's comment coming in right now), so I'll retract this PR. Thanks for clarifying :)

(One could add +dev now but that seems moot with 1.1.1 coming up.)

So for the record:
- in-source version is the base version
- `VERSION_SUFFIX` indicates whether it is as released (``) or with additional developement on top (`+dev`)

LittleHuba:
Thanks for checking this once more!

I think it is preferable to stay with the setup as it is now. We'll add the +dev back with #3299 . I'm not quite sure if the PR is already set up correctly for this but I'll add that `+dev` suffix if it is missing.

Technius:
For the record, I was the one who introduced the `+dev` suffix in #1599. @mjg's reasoning in https://github.com/xournalpp/xournalpp/pull/3310#issuecomment-894173426 matches with my intention in making the aforementioned change. I modeled the version suffix after the Java package repository naming convention, e.g.:
* `1.1.0-SNAPSHOT` in Java means "in-development version that will eventually be released as version 1.1.0"
* `1.1.0-M1` means "in-development version that is the 1st milestone for version 1.1.0"
* `1.1.0-rc1` means "1st release candidate for version 1.1.0"
* etc. 

I modified this scheme so that we:
* replace `-` with `+` in order to be more consistent with [Debian package versioning](http://www.fifi.org/doc/debian-policy/policy.html/ch-versions.html). Note that `-` is commonly used to indicate Debian revisions, so I instead used `+`.
* Use `dev` instead of `SNAPSHOT` to more clearly indicate a development version.

> Is leading up to, is somewhat missleading if we do not know ourselves to what release we actually work to.

This does not matter with the new branching scheme. We have separate branches for releases; on these branches we know _exactly_ what release we are working towards. The next release on a minor branch will be versioned as `1.1.(x+1)`, and the next major release will be versioned as `1.2.0` (or some other "greater" version that is _not_ `1.1.y`).

Furthermore, without changing the version numbers on the development branch, how are we supposed to distinguish between different branches? In particular, I refer to the "nightly" builds. Let's assume we have a single nightly version, which corresponds to the next major release in development. If I were to download this nightly, I would assume that it may contain backwards incompatible changes. Furthermore, I would like a _clear indication as such_; a version number of `1.2.x` gives me this information. If the build was instead labeled as some `1.1.x` version, now the exact feature set included in this build would be ambiguous and misleading.

Finally, there is no big hurdle in changing the _final_ version number of an in-development version, if we desire to change the it at the last minute.

LittleHuba:
Thanks for clearing this up!

> 1.1.0-SNAPSHOT in Java means "in-development version that will eventually be released as version 1.1.0"


But that system won't work with Debian as each `1.2.0+dev` release is seen as being higher than 1.2.0 itself. So in essence users won't ever be able to switch to the main release as it is seen as a downgrade instead of an upgrade.

What should work (not tested yet) is `1.2.0~dev`, as it should sort before `1.2.0`.

> on these branches we know exactly what release we are working towards.

On release branches we do know the next version. And on these branches the first thing that is done is bumping the version.
I had in mind that we release these versions as release candidates. Something like `1.2.0~rc`.

> If I were to download this nightly, I would assume that it may contain backwards incompatible changes. Furthermore, I would like a clear indication as such; a version number of 1.2.x gives me this information.

That goes in line with semantic versioning. A really bad way of versioning, from what I read recently. But your point stands and makes quite some sense.

So should we bump on the release branch to `1.1.(x+1)~rc` and on master when creating the release branch to `1.(x+1).0~dev`?

mjg:
It seems I stirred up quite a discussion, but hopefully for the good of the project :)

I think you should decide on the branching model first - you might have done so already with the new release branch. E.g., one could use `master` for development and then branch off `release-1.1` for 1.1.0 and following patch releases 1.1.x, cherry-picking from `master`. (Before, there seemed to be only one main line with consecutive releases.)

Next: What does the about dialogue say, i.e. how do you communicate the version to the end user in case of a snapshot? git revs are definitive but not informative, not even the output of `git describe --tags` (because 1.1.0 plus 10 commits could be hotfixes as well as mainline development), so this needs to be some form of "1.1.0+dev" (indicating base plus development) or "1.2-dev" (development leading up to 1.2 or higher). In any case, this needs to be user parsable.

Finally, there is the package version. Any packager will have to amend the above info by something which sorts correctly for their distro's resolver. E.g., Fedora (and other dnf suers) have versioning schemes both for "snapshot leading up to" and "snapshot based on", and I'm sure Debian and others have the same for deb. The "~" sorting before numbers (to make RCs at work) is common to both, e.g.

Ideally, the "in-source version" is both informative for the user and a good base for the package version, but that's not a strict necessity, I would think.

So, if my assumption on the branching model is right, and answering @LittleHuba 's question, you could have the development branch at `1.1~dev`, then at release branch time `git branch releases-1.1.x master` (or whatever you want to name the branch for the minor releases) and bump to `1.2~dev` on `master`. This way, you can keep `1.1~dev` for doing some wrangling on release branch if necessary before cutting a `1.1.0~rc1` which sorts correctly, finally a `1.1.0` for the release. Then `1.1.1~dev` on the release branch if necessary, or directly `1.1.1~rc1` etc. `1.2~dev` from the mainline always sorts newer.

Using `1.1~dev` on the mainline development helps the user distinguish it from release branch development with `1.1.0~dev`, but both would work for the sorting if you bump mainline at release branch time.

BTW, since this is too long already: This (clarifying branching/versioning) might be a good time to rename the mainline branch from `master` to something like `devel` or `main` (as in "main development", not "main version"). Personally, I have no problems with `master` at all, but I understand why some people do, and a different name can be more descriptive.

PR #3312, opened 2021-08-06
Toolbar: `ZoomSlider` refactor

personalizedrefrigerator:
## Summary
 * Intended to facilitate creation of new slider toolbar tools.
 * Applies the refactor from #3269 to `ZoomSlider`.
 * This refactor may also be useful to #3288.

### DIFF
--- a/src/gui/toolbarMenubar/AbstractSliderItem.cpp
+++ b/src/gui/toolbarMenubar/AbstractSliderItem.cpp
@@ -1,0 +1,189 @@
+#include "AbstractSliderItem.h"
+
+#include <cassert>
+#include <iomanip>
+#include <sstream>
+#include <tuple>
+
+class AbstractSliderItem::Impl {
+public:
+    Impl(AbstractSliderItem* publicApi_, SliderRange range);
+
+    static void onSliderChanged(GtkRange* range, AbstractSliderItem* parent);
+    static bool onSliderButtonPress(GtkRange* range, GdkEvent* event, AbstractSliderItem* parent);
+    static bool onSliderButtonRelease(GtkRange* range, GdkEvent* event, AbstractSliderItem* parent);
+    static bool onSliderHoverScroll(GtkRange* range, GdkEventScroll* event, AbstractSliderItem* parent);
+    static gchar* formatSliderValue(GtkRange* range, double value, AbstractSliderItem* parent);
+
+    /**
+     * Listen for input events on the new slider.
+     *
+     * If this already has a slider, this disconnects listeners
+     * on events for the current slider.
+     *
+     * @param slider The new slider to listen for events on.
+     */
+    void setSlider(GtkRange* slider);
+
+    void setEnabled(bool enabled);
+
+    /**
+     * Constructs a new toolbar item and slider.
+     * Does not take ownership.
+     */
+    std::tuple<GtkToolItem*, GtkRange*> newItem(bool horizontal) const;
+
+    // Conversions to/from range_ (internal values) to external scale values:
+    double getScaleMax() const { return publicApi_->scaleFunc(range_.max); };
+    double getScaleMin() const { return publicApi_->scaleFunc(range_.min); };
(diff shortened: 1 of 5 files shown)


src/gui/toolbarMenubar/AbstractSliderItem.cpp
```diff
@@ -0,0 +1,194 @@
...
+    auto [item, slider] = pImpl->newItem(horizontal);
+    g_object_ref(item);
+
+    gtk_widget_show_all(GTK_WIDGET(item));
+    return item;
+}
+
+// TODO(personalizedrefrigerator): Remove this (currently required by AbstractToolItem
+auto AbstractSliderItem::newItem() -> GtkToolItem* { return createTmpItem(true); }
+
```
personalizedrefrigerator, line 99: Currently, `AbstractToolItem` requires an implementation of `newItem` in addition to `createItem` and `createTmpItem`. 

In `newItem` it is not specified whether the item is `horizontal` or not and `newItem` only returns one thing, a `GtkToolItem` (no pointers to internal widgets, e.g. a slider). As such, it isn't clear whether `newItem` is intended to modify the state of `this`, or whether it should create an empty toolbar item...

src/gui/toolbarMenubar/AbstractSliderItem.cpp
```diff
@@ -0,0 +1,194 @@
...
+    parent->onSliderHoverScroll();
+    return false;
+}
+
+auto AbstractSliderItem::Impl::formatSliderValue(GtkRange* range, double value, AbstractSliderItem* parent) -> gchar* {
+    std::string text = parent->formatSliderValue(parent->scaleFuncInv(value));
+
+    // TODO: At present, gchar* is equivalent to (const char*). Is there a better/won't break
+    // way of doing this cast?
+    return g_strdup(static_cast<const gchar*>(text.c_str()));
```
personalizedrefrigerator, line 193: I don't like this conversion from `std::string` to `gchar*`. Is there a better way to do this?

src/gui/toolbarMenubar/AbstractSliderItem.h
```diff
@@ -0,0 +1,112 @@
...
+        // Number of fine/coarse steps between min and max.
+        double fineSteps;
+        double coarseSteps;
+    };
+
+public:
+    /**
+     * @param id TODO(personalizedrefrigerator; @see AbstractToolItem::AbstractToolItem)
+     * @param handler TODO ^^
+     * @param type TODO ^^
```
personalizedrefrigerator, line 38: Documentation TODOs that need to be resolved before merging.

src/gui/toolbarMenubar/ToolZoomSlider.cpp
```diff
@@ -1,182 +1,158 @@
...
 
-ToolZoomSlider::ToolZoomSlider(ActionHandler* handler, std::string id, ActionType type, ZoomControl* zoom,
+constexpr double SCALE_LOG_OFFSET = 0.20753;
+
+constexpr double FINE_STEP_COUNT = 100.0;
+constexpr double COARSE_STEP_COUNT = 10.0;
+
+// TODO(personalizedrefrigerator): Update the range to reflect the min and max zoom settings.
+constexpr auto SLIDER_RANGE =
+        AbstractSliderItem::SliderRange{DEFAULT_ZOOM_MIN, DEFAULT_ZOOM_MAX, FINE_STEP_COUNT, COARSE_STEP_COUNT};
```
personalizedrefrigerator, line 18: Before this PR, the zoom slider's range was also from `DEFAULT_ZOOM_MIN` to `DEFAULT_ZOOM_MAX`. As such, while this should be updated to match the setting, this should not cause a regression.

personalizedrefrigerator:
Merging in 24-ish hours unless there are objections.

personalizedrefrigerator:
Now merging in 48 hours unless there are objections.

PR #3313, opened 2021-08-06
[Experiment] CAIRO_FILTER_FAST, CAIRO_ANTIALIAS_FAST

personalizedrefrigerator:
This PR **may** make rendering/drawing faster when zoomed in/out far. It may also have little/no effect.

**TODO**:
 * [ ] Benchmark and compare speed
 * [ ] Any reduction in quality?


personalizedrefrigerator:
Unfortunately, this (and inserting similar configuration just after creating most other `cairo` contexts) doesn't seem to improve the performance of #2729 (live approximation while zoomed in).

PR #3315, opened 2021-08-07
Extend input method check for gcin

changhc:
Following #470 and #530, this PR extends the input method check to cover any input method that is not ibus.

In my case, I'm using `gcin`, an input method for Traditional Chinese. This is not covered in the current check, so xournalpp crashes every time I try to add some text to my pdf documents. My change will make the check more comprehensive than checking only against `xim`.

Edit: checking against `xim` and `gcin` only now.

#### Test output
```
$ echo $GTK_IM_MODULE
gcin
$ ./src/xournalpp
** Message: 18:15:22.754: TEXTDOMAINDIR = (null), PACKAGE_LOCALE_DIR = /usr/local/share/locale, chosen directory = /usr/local/share/locale            
                                                                                                                                                      
** (xournalpp:18484): WARNING **: 18:15:22.754: Unsupported input method: gcin, changed to: ibus                                                      
ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave                                                                                     
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear                                                                                
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe                                                                          
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side                                                                                
ALSA lib pcm_route.c:869:(find_matching_chmap) Found no matching channel map                                                                          
ALSA lib pcm_oss.c:377:(_snd_pcm_oss_open) Unknown field port                                                                                         
ALSA lib pcm_oss.c:377:(_snd_pcm_oss_open) Unknown field port                                                                                         
ALSA lib pcm_usb_stream.c:486:(_snd_pcm_usb_stream_open) Invalid type for card                                                                        
ALSA lib pcm_usb_stream.c:486:(_snd_pcm_usb_stream_open) Invalid type for card                                                                        
ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave                                                                                     
** Message: 18:15:22.788: Plugin "MigrateFontSizes" UI initialized 
```

### DIFF
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -383,10 +383,16 @@ using XMPtr = XournalMainPrivate*;
 
 /// Checks for input method compatibility and ensures it
 void ensure_input_model_compatibility() {
+    char* unsupportedImModules[] = {"xim", "gcin"};
     const char* imModule = g_getenv("GTK_IM_MODULE");
-    if (imModule != nullptr && strcmp(imModule, "ibus") != 0) {
-        g_setenv("GTK_IM_MODULE", "ibus", true);
-        g_warning("Unsupported input method: %s, changed to: ibus", imModule);
+    if (imModule != nullptr) {
+        for (int i = 0; i < 2; i++) {
+            if (strcmp(imModule, unsupportedImModules[i]) == 0) {
+                g_setenv("GTK_IM_MODULE", "ibus", true);
+                g_warning("Unsupported input method: %s, changed to: ibus", imModule);
+                break;
+            }
+        }
     }
 }
 
--- a/src/control/XournalMain.cpp
+++ b/src/control/XournalMain.cpp
@@ -384,9 +384,9 @@ using XMPtr = XournalMainPrivate*;
 /// Checks for input method compatibility and ensures it
 void ensure_input_model_compatibility() {
     const char* imModule = g_getenv("GTK_IM_MODULE");
-    if (imModule != nullptr && strcmp(imModule, "xim") == 0) {
+    if (imModule != nullptr && strcmp(imModule, "ibus") != 0) {
         g_setenv("GTK_IM_MODULE", "ibus", true);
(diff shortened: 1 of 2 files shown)


tattsan:
This PR seems to exclude all input methods other than `ibus`. However I don't use `ibus` on Xournalpp, but `uim`.
It is violent to "eliminate everything except `ibus`", isn't it?

LittleHuba:
I don't really get the point of this PR as it moves from blocking one specific input method ('`xim`) to blocking all input methods (including `xim`). 

So it does obviously not improve the situation of the author, but worsens the situation of everybody else not using `xim`?

changhc:
You're right. I didn't know other input methods work normally with xournalpp. I'll change this PR to block `gcin` only.

src/control/XournalMain.cpp
```diff
@@ -383,10 +383,16 @@ using XMPtr = XournalMainPrivate*;
 
 /// Checks for input method compatibility and ensures it
 void ensure_input_model_compatibility() {
+    char* unsupportedImModules[] = {"xim", "gcin"};
     const char* imModule = g_getenv("GTK_IM_MODULE");
```
personalizedrefrigerator, line 387: I would be more comfortable with this if we used `std::string` here. :)

E.g.
```c++
const char* imModuleEnv = g_getenv("GTK_IM_MODULE");

if (imModuleEnv != nullptr) {
  std::string imModule { imModuleEnv };
  if (imModule == ... || imModule == ...) {
   ...
  }
}
```

changhc: Updated. Thanks for your comment!

personalizedrefrigerator:
I've applied clang-format!

This looks good to me! Merging in 24 hours unless there are objections.

Issue #1405, opened 2019-08-09
Repeated white background flash on larger colored background documents


fxkrait:
**Affects versions :**
 - Arch Linux
 - xournalpp 1.0.12-1

**Describe the bug**
Background repeatedly flashes white on colored background document that contains multiple pages and multiple text items.

**To Reproduce**
1. Make background light gray color for all pages
2. make 13 pages (just to make a lot of pages, it can happen with less)
3. On first page, create a large text block, and copy and paste it multiple times (to create a lot of text in document)
![firstPage](https://user-images.githubusercontent.com/7407672/62771250-a8493c80-ba51-11e9-9d7c-6718a843448b.png)

4. go to second page, and make some new text
5. Notice flashing

**Expected behavior**
Background should not flash white.

**Screenshots of Problem**
gif of flashing:
![flash](https://user-images.githubusercontent.com/7407672/62771301-c151ed80-ba51-11e9-8c39-323ae2916e3d.gif)

**Additional context**
Note: This does go away if the file is re-opened.
However, I do believe it will soon happen again in the same document upon further editing (and additions).
The flash can also occur when scrolling around the document.

lehmanju:
Related to #1107

ehremington:
This very annoying since it means that the projector screen is flashing at my class full of students.

This happens for me when I have zoomed in to a portion of a page and then start writing, although it seems sporadic even in that case.

LittleHuba:
I isolated the problematic section of code:
```
void XojPageView::paintPageSync(cairo_t* cr, GdkRectangle* rect)
{
	XOJ_CHECK_TYPE(XojPageView);

	if (this->crBuffer == nullptr)
	{
		drawLoadingPage(cr);
		return;
	}
```

For performance reasons we have a cleaning job that removes the buffers of pages that were not rendered for a longer time. This will remove the variable crBuffer which triggers a repainting of the Loading-Page (the white flicker...)

I'm still searching for the best way to fix this.

LittleHuba:
If you are not interested in the technicalities you may ignore this post. 

Okay so the problem is as follows:
Xournal++ draws pages in two steps:
1. Render the contents onto a buffer surface to minimize the time consumed by the overall drawing process
2. Draw the actual view
2.1. Draw the background of the view
2.2. Draw the shadows around the visible pages
2.3. Draw the visible pages using the buffers from step 1

[...]

PR #3326, opened 2021-08-14
[WIP]: PDF Text Selection Util

aeghn:
Following @Technius's suggestion, I decided to open a draft pull request.

As the title says, this pull request is to add interaction with the PDF text, like copying text, highlighting text, etc. 

This PR is going to close this issue: #1745 

Preview: 

https://user-images.githubusercontent.com/51843252/128634365-ddc3e7fd-255a-4ea9-933e-4a1c1257afaf.mp4

As the video shows, I added the following features
1. select text and draw text area in real time
2. pop-up menu after the selection is done, supporting respectively
    - Copy text directly
    - Highlighting the selection area
    - Add underline to the selection area
    - Adding strikethrough to the selection area
3. After selecting the area, you can  select another color if you change your mind
 
And the last three are new elements added directly, with a direct call to strokeHandler, and each line is a separate object. Because it seems to me that these three markup tools are designed to help people using computers to mark up quickly and should be on the same level as other manual markup tools. Of course this can be discussed and replaced by a better solution.

But this does not have all the features yet.

I'm now thinking of adding the ability to select text in a region, which could be useful when we select part of the text from a table.
Final result should like be

https://user-images.githubusercontent.com/51843252/129444907-e2e2516a-7404-4caa-8ea8-4d3e6255e42f.mp4

Thanks to the help of @Technius, I could clean the previous selection, but I find there is another problem is how to determine when the re-rendering work is done. To get the demo above, I let the thread sleep for 20 ms, 
the [method](https://github.com/aeghn/xournalpp/blob/8d21edce2eede3dc9314397e29731e2d567e6ab1/src/control/PDFTextSelectControl.cpp#L253) is
```c++
void PDFTextSelectControl::reselect() {
    // invoke view->rerenderRect();
    this->rerenderBox(this->selectTextRecs);
    // invoke view->repaintPage();
    this->repaintPage();

    // FIXME It should happen after rerender job
    std::chrono::milliseconds dur(20);
    std::this_thread::sleep_for(dur);

    this->paint(this->cr, this->zoom, this->zoom);
    this->selectPdfText();
}
```

but I know this is not allowed, so I hope you can give some hints. Thanks in advance!


Technius:
See this comment https://github.com/xournalpp/xournalpp/issues/1405#issuecomment-530076810 on a high-level overview of how the rendering system works.

Once you update an element, you only need to request a rerender (of the buffer), and the rendering system should handle the rerender and redraw for you. For your code snippet above, you'll want to call `this->selectPdfText()` first and then ask for just `rerenderBox/Page`. Let me know if you need more help and I can take a close look at the code.

aeghn:
@Technius 
Sorry, I didn't explain my thoughts on this, I'll restate them below

First of all the key point of the operation is that I need the selection area to exist until the next operation (marking, copying, changing the selection style, etc.) is performed. If I call `rerender()` directly after the drawing is done, then the content just drawn will be cleaned up, which is not suitable for this operation.

Next is my complete idea of the logic of this operation.

1. the user completed the selection (that is, the following happens after on the call `Selection::finalize()` ==> `PdfTextSelectControl::finalize()` method, rather than the previous drag selection operation), at this time the selection area has been fixed and needs to be displayed (because the user may need to check whether the selection area is correct) So this does not allow the call `rerenderPage ()`/`rerenderRect ()`, because it will directly clear the previous highlighted selection
2. pop-up menu
3. the user starts the next operation (marking, copying, changing the selection style), which clears the highlighted selection from the previous step
   - If it's mark and copy, then just clear the selection area and do the corresponding action, and hide the menu, now we could just invoke `rerenderRect()` method
   - If the selection type is changed (e.g. from selecting a paragraph to selecting text in the area)
       1) clear the selection area of the previous step
       2) redraw the new selection area (while the new selection area also needs to wait for the next operation to be cleared)

     Now here is the problem, I can't tell when the `rerender()` job will finish, and if it happens slower than `b.`, then the newly drawn area will be cleared again.

Technius:
I briefly skimmed through the code and found a few issues.

* The `draw*` methods are confusingly named because they create stroke elements, not draw them. I would suggest renaming them to be, e.g. `createFoo`.
* A rerender will trigger a repaint, so you don't need to call any of the `repaint*` methods.
* `PDFTextSelectControl` should not store the cairo context. Furthermore, you should not be invoking `paint` directly within the `PdfTextSelectControl`. You can rely on the fact that a rerender will call `PageView::paintPageSync`, which will then call the `paint` method on the selection.
* In the `PDFTextSelectControl::finalize` method, you can just call `rerenderRect` on the selected text area.

> Now here is the problem, I can't tell when the rerender() job will finish, and if it happens slower than b., then the newly drawn area will be cleared again.

I think this will no longer be an issue once you address the problems above. On a reselection, you need to store the bounds of the current selected page. Then you select the new text area, and finally you can trigger a rerender on both 1) the previously selected area; and 2) the newly selected area. Since the only active selection is the new text area, the previous selection should be cleared and the new selection should be drawn on the next rerender.

aeghn:
Thank you for your suggestions, I will refer to them to modify the unreasonable parts of my code

aeghn:
Hi @Technius, I've refactored this section. The result is:

1. `PdfFloatingToolbox` for floating toolbox, creating strokes and copying text.
2. `PdfTextSelection`, used to interact with `PopplerGlibPage` and draw the selection.
   The reason why I extracted this class from `Selection.cpp` is `PdfTextSelection` and `Selection` don't have the same lifetime. When the button is released, `Selection` object will be destroyed, while `PDFTextSelection` still need to keep the highlighted selection(I haven't noticed that before, so I watsed a lot of time on it). I do not want to change the previous logic, so I added a new member in `XojPageView`, called `pdfTextSelection`, its logic is similar to `view->selection`, but there are two other restrictions.
   - isFinilized: whether the user releases the button.
   - isFinished: whether the user creates a stroke, copy the text or exit the menu, then `pdfTextSelection` should be deconstructed.
3. delete `PDFTextSelectControl`, I think this can make the structure more clear.

Now that it has worked, I will then focus on how to choose a more accurate area. If you have any ideas, please let me know. :)

Technius:
Sounds good from a high level. Let me know when you're ready for me to take a closer look.

aeghn:
I think I've accomplished what I expected when I started this.

These days I've abandoned the deprecated `poppler_page_get_selection_region` and started using `poppler_page_get_selected_region` instead to make selections and use `poppler_page_get_text_layout_for_area` after the selection.

aeghn:
As I mentioned earlier, I'm working on the area text selection. But this is not easy to do.

poppler officially provides `poppler_page_get_text()` and `poppler_page_get_text_layout()` (return each graph position) interfaces, the two interfaces return two arrays are one-to-one correspondence, so we can use the two methods to do area selection. Then I found that these two interfaces only work properly with ASCII encoding, so I may need to look into a dirtier way to make area selection for other languages now.

aeghn:
With @liangyongxiang's help, now it is possible to do area selection and copy the right text (only if when `poppler_page_get_text_layout()` returns utf-8 text).

Preview:

https://user-images.githubusercontent.com/51843252/131220791-0618bcaa-1277-4736-bd13-cdf88cd0f24e.mp4

Technius:
Nice work! I think we can assume UTF-8 for now since that's what glib does. I can help you clean things up later.

aeghn:
@Technius I finished all my ideas at this stage.

So I think this draft merge request is ready and it's time to start reviewing it.

ui/xournalpp.css
```diff
@@ -7,9 +7,23 @@
  *  See: https://wiki.gnome.org/action/show/Projects/GTK/Inspector
  */
 
-/*Style the Floating Toolbox!*/
+#pdfFloatingToolGrid {
+    padding: 7px;
+}
+
+#pdfFloatingToolGrid button {
```
personalizedrefrigerator, line 14: Make sure you handle the 'window.darkMode' case, too! (See lines 67-85)

src/control/tools/PdfTextSelection.cpp
```diff
@@ -0,0 +1,140 @@
...
+PdfTextSelection::PdfTextSelection(double x, double y, XojPageView* view) {
+    this->sx = x;
+    this->sy = y;
+    this->ex = x;
+    this->ey = y;
+
+    this->view = view;
+
+    this->isFinalized = false;
+    this->isFinished = false;
```
personalizedrefrigerator, line 14: (<i>Really minor:</i> Feel free to ignore!) Personally, I think the following is a bit easier to read. :)

```suggestion
PdfTextSelection::PdfTextSelection(double x, double y, XojPageView* view):
        sx(x),
        sy(y),
        ex(x),
        ey(y),

        view(view),
        isFinalized(false),
        isFinished(false) {
```

src/gui/PdfFloatingToolbox.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+void PdfFloatingToolbox::copyText() {
+    GtkClipboard* clipboard = gtk_widget_get_clipboard(this->theMainWindow->getWindow(), GDK_SELECTION_CLIPBOARD);
+
+    const gchar* text = this->pdfTextSelection->getSelectedText().c_str();
+
+    gtk_clipboard_set_text(clipboard, text, -1);
+}
+
+void PdfFloatingToolbox::createStrokes(PdfMarkerStyle position, PdfMarkerStyle width, int markerOpacity) {
```
personalizedrefrigerator, line 138: It looks like highlighting/underlining is broken when input stabilization is on:
segfault on getStroke: this=0x0

<details><summary>My input stabilization settings</summary>

image

</details>

src/gui/PdfFloatingToolbox.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if (this->pdfTextSelection->getSelectedTextRects().empty())
+        return;
+
+    auto strokeHandler =
+            new StrokeHandler(this->pdfTextSelection->getPageView()->getXournal(),
+                              this->pdfTextSelection->getPageView(), this->pdfTextSelection->getPageView()->getPage());
+    Stroke* stroke = nullptr;
+
+    auto p = new PositionInputData();
+    p->pressure = -1;
```
personalizedrefrigerator, line 148:
```suggestion
    p->pressure = Point::NO_PRESSURE;
```

src/gui/PdfFloatingToolbox.cpp
```diff
@@ -0,0 +1,188 @@
...
+    p->state = GDK_RELEASE_MASK;
+
+    for (XojPdfRectangle rect: this->pdfTextSelection->getSelectedTextRects()) {
+        // the center line position of stroke
+        double h = position == PdfMarkerStyle::POS_TEXT_BOTTOM ? std::max(rect.y1, rect.y2) :
+                   position == PdfMarkerStyle::POS_TEXT_MIDDLE ? (rect.y1 + rect.y2) / 2 :
+                                                                 std::min(rect.y1, rect.y2);
+
+        // the width of stroke
+        double w = width == PdfMarkerStyle::WIDTH_TEXT_LINE ? 1 : std::abs(rect.y2 - rect.y1);
```
personalizedrefrigerator, line 158: I think I find the following might be easier to understand:

```suggestion
    for (XojPdfRectangle rect: this->pdfTextSelection->getSelectedTextRects()) {
        double topOfLine = std::min(rect.y1, rect.y2);
        double middleOfLine = (rect.y1 + rect.y2) / 2;
        double bottomOfLine = std::max(rect.y1, rect.y2);
        double rectWidth = std::abs(rect.y2 - rect.y1);
        
        // the center line position of stroke
        double h = position == PdfMarkerStyle::POS_TEXT_BOTTOM ? bottomOfLine :
                   position == PdfMarkerStyle::POS_TEXT_MIDDLE ? middleOfLine : topOfLine;

        // the width of stroke
        double w = width == PdfMarkerStyle::WIDTH_TEXT_LINE ? 1 : rectWidth;
```

This is subjective! If you disagree with this change, please ignore it!

ui/main.glade
```diff
@@ -124,6 +124,168 @@
...
+          <packing>
+            <property name="left-attach">2</property>
+            <property name="top-attach">0</property>
+            <property name="width">1</property>
+            <property name="height">1</property>
+          </packing>
+        </child>
+      </object>
+    </child>
+  </object>
```
personalizedrefrigerator, line 288: Please add tool-tips here!
(e.g. `<property name="tooltip-text" translatable="yes">Strikethrough text</property>`)

personalizedrefrigerator (changes requested):
Thank you for working on this! I have a few suggestions after a quick look through the code!

src/control/tools/PdfTextSelection.cpp
Technius: Yes, I would move these initializations to the declaration line as suggested above. `clang-format` will reformat these to be on one line, making the code clearer.

src/gui/PdfFloatingToolbox.cpp
Technius: We can clean this up further by making `PdfMarkerStyle` an `enum class` and then using a `switch` statement. The compiler will warn about missing cases, which can potentially catch some bugs.

Technius:
This is a pretty big PR so it'll take me a few days to do a detailed review. Some high level comments based on a brief skim through the code:
* I would rename the `Select PDF Text` tool to `Select PDF Element` because we can eventually use it for more features, like clicking on PDF links and copying images.
* The application crashes if I use the `Select PDF` tool on a new document with no PDF.
* I'm not a big fan of duplicating the floating toolbox for the PDF popup. I would've liked the tool to open a popover instead, but the current solution works and is not worth the effort to change right now.
* The icon and the code copied from poppler need to be added to `copyright.txt`.
* The new `Select PDF` tool should either be 1) placed next to the other select tools; or 2) it should be placed under the select tools dropdown.
* There seems to be quite a bit of duplication in the new utility functions added to `PopplerGlibPage`. Is there some way to simplify it?

Technius (changes requested):
This is a pretty big PR so it'll take me a few days to do a detailed review. Some high level comments based on a brief skim through the code:
* I would rename the `Select PDF Text` tool to `Select PDF Element` because we can eventually use it for more features, like clicking on PDF links and copying images.
* The application crashes if I use the `Select PDF` tool on a new document with no PDF.
* I'm not a big fan of duplicating the floating toolbox for the PDF popup. I would've liked the tool to open a popover instead, but the current solution works and is not worth the effort to change right now.
* The icon and the code copied from poppler need to be added to `copyright.txt`.
* The new `Select PDF` tool should either be 1) placed next to the other select tools; or 2) it should be placed under the select tools dropdown.
* There seems to be quite a bit of duplication in the new utility functions added to `PopplerGlibPage`. Is there some way to simplify it?

aeghn:
@personalizedrefrigerator @Technius Thanks for the code review, this is my first attempt to work on a C++ application (to be honest, I have no experience in C++ and only a little knowledge of C before), so I realized that there are really a LOT of problems inside, although I tried my best and spent a lot of my free time. All in all, I see this as a learning opportunity, and I will take your suggestions into consideration and think about my previous shortcomings.

ui/xournalpp.css
aeghn: Oh, okay. In my opinion the floating toolbox should have same background color with the pdf and I couldn't find how to make pdf background black, so I didn't handle that case(I want to try to add the function to siwtch pdf light/dark mode on the fly and then add the `window.darkMode` for pdf floating toolbox then), thus you point out this it's okay to add dark mode for toolbox now.

aeghn:
@Technius 
> The application crashes if I use the Select PDF tool on a new document with no PDF.

I will work on this soon.

> I'm not a big fan of duplicating the floating toolbox for the PDF popup. I would've liked the tool to open a popover instead

I think floating toolbox is more harmonious with selection, and popover is suitable for click operation(like clicked for a link).

> The icon and the code copied from poppler need to be added to copyright.txt.

There is no icon is copied from poppler

> The new Select PDF tool should either be 1) placed next to the other select tools; or 2) it should be placed under the select tools dropdown.

How about only show this when only open a pdf document?

> There seems to be quite a bit of duplication in the new utility functions added to `PopplerGlibPage`. Is there some way to simplify it?

There are four methods added to the `PopplerGlibPage`. The last one is for Area Selection, and the other three is used for selection from start to end. I didn't see the need to streamline them.

ronandalton:
Good job on this feature so far. Here are a few suggestions/issues I found while trying out the current implementation:
- Undo action only deletes one line at a time - should delete all lines
- Clicking away should cancel the selection / start new one
- Show an `I` cursor when selecting normally and `+` when using box select
- Selecting text across pages crashes the program
- Popup tool moves strangely when zooming out far
- The highlight and box select toggle icons aren't well suited
- The popup tool should follow the selection when panning view
- Double/triple click [and drag] would be nice to allow selecting whole words/lines
- The select PDF text icon could be improved with an `I` and a PDF logo

LittleHuba:
Fixes #3369

personalizedrefrigerator:
@aeghn I would love to help work on this feature. Are you fine with me pushing commits directly to 20210814_pdf_selection?
Thank you for your contribution!

Technius:
I'd also like to help out--there are some things that would be done faster if I were to write the code directly instead of writing paragraphs of feedback.

aeghn:
@personalizedrefrigerator @Technius 

Of course, it would be a great help if you could also participate in this PR directly, which will ensure a more uniform style of code and better and more stable functionality.

src/gui/PdfFloatingToolbox.cpp
Technius: I rewrote the stroke creation logic, can you check again?

aeghn:
Trying to fix this:
> Selecting text across pages crashes the program

But it looks like the cross-page selection isn't so easy to do.

What do you think? @Technius @personalizedrefrigerator 

Implementing a single page selection first (like the other selection tools)?

Technius:
> Implementing a single page selection first (like the other selection tools)?

Yes, let's keep it simple for now. We can make it more complex and robust in a future PR.

aeghn:
- [ ] Undo action only deletes one line at a time - should delete all lines (fixed by @Technius)
- [ ] Clicking away should cancel the selection / start new one f97e297 and ca854d7
- [ ]  Selecting text across pages crashes the program 6f472a8 
       We just support select in single page now 
- [ ] Popup tool moves strangely when zooming out far 6ceeb8b 
- [ ] Double/triple click [and drag] would be nice to allow selecting whole words/lines 6faf552

@ronandalton  Thanks for the test and please recheck if those problems have been fixed .

Please let me know if this has any problems.

raffaem:
I have tested it and it works wonders. A few notice:

1. Sometimes there is a delay from the moment I press the highlight / cross / underline button until the action is fully carried over if the text selected is substantial (a discrete amount of lines)
2. Sometimes there is a delay in selecting the text for the double click / triple click
3. Would it be possible to have help messages for the menu that pops over when you select text? So that if you hover over "copy" a text message appears that say "over" etc.
4. The "i" button for highlighting text I find misleading. Would it be possible to have something like [this](https://freesvg.org/img/fthighlighting.png)?

Thank you very much for your work. It is very useful indeed :)

raffaem:
Uhm when you highlight a word it highlights too much. For example, this is the result of highlighting the word "from":

image

Would it make sense to have the highlight a perfect rectangle without rounded borders?

raffaem:
Arg, if you select text from a page, you scroll to the next page, and you press the highlight button, the text on that next page is highlighted, not the text on the original page that you selected.

aeghn:
> the text on that next page is highlighted, not the text on the original page that you selected.

@raffaem Have you pull the latest code?

aeghn:
> Following our discussion on chat, [here is](https://youtu.be/HbOQTAQF1ns) a video recording of the issue.

@raffaem 
I have tried to solve this issue, please retry it

aeghn:
@raffaem 

> The pop-up now disappears and the text is un-selected if I scroll over to the next page

Yes, this is the right result. When you scroll to the other page, the selection should be cleared and the pop menu disappears.

> Would it be possible to have help messages for the menu that pops over when you select text? So that if you hover over "copy" a text message appears that say "copy" etc.

>The "i" button for highlighting text I find misleading. Would it be possible to have something like this?

> Would it make sense to have the highlight a perfect rectangle without rounded borders?

These `graphic issues` should to be considered after we can make sure the whole part works normally.

> Sometimes there is a delay from the moment I press the highlight / cross / underline button until the action is fully carried over if the text selected is substantial (a discrete amount of lines)
> Sometimes there is a delay in selecting the text for the double click / triple click.

I'm not in a position to fix xournalpp's rendering-related issues, I'm still not familiar with that part of the code and might need help from other contributors.

ronandalton:
>     * [x]  Undo action only deletes one line at a time - should delete all lines (fixed by @Technius)
> 
>     * [x]  Clicking away should cancel the selection / start new one [f97e297](https://github.com/xournalpp/xournalpp/commit/f97e2971e780dd4022cec2e94026896d50dcd5b5) and [ca854d7](https://github.com/xournalpp/xournalpp/commit/ca854d7db082b09a7d21d0a1245de6982191dbf9)
> 
>     * [x]   Selecting text across pages crashes the program [6f472a8](https://github.com/xournalpp/xournalpp/commit/6f472a89cac2423b7362043bf852132d3d912c88)
>       We just support select in single page now
> 
>     * [x]  Popup tool moves strangely when zooming out far [6ceeb8b](https://github.com/xournalpp/xournalpp/commit/6ceeb8b61d6d831ec9102faf7d19a122f8e25b2a)
> 
>     * [x]  Double/triple click [and drag] would be nice to allow selecting whole words/lines [6faf552](https://github.com/xournalpp/xournalpp/commit/6faf552aff83f14a887d8618f6b9a9a64b0de9d3)
> 
> 
> @ronandalton Thanks for the test and please recheck if those problems have been fixed .
> 
> Please let me know if this has any problems.

Nice! I've tested it and several of those issues appear to be fixed. A few points I noticed:
- The click away action doesn't appear to work if you click outside a page.
- Switching tools (to say the pen tool) while something is highlighted makes things go funky.
- The effect of double/triple clicking is only seen after releasing the mouse. Most text editors select the whole word/line while the selection is taking place.
- The strange behavior of the popup menu when zooming out far seems to still be there.

ronandalton:
> > @ronandalton @aeghn I should have fixed this:
> 
> Thanks for the effort. @raffaem
> 
> > The strange behavior of the popup menu when zooming out far seems to still be there.
> 
> @ronandalton Could you make a screen recording for this?

Here is a screen record:
https://user-images.githubusercontent.com/86718942/136301042-997dfd62-264c-4e7c-8895-beadc3c45a40.mp4
It might not matter that much. Just thought it was a bit weird that's all.

aeghn:
> git push says I don't have write access to this PR.
Do I have to branch this PR?

You should open a pr in my fork.

Emm... The scope of this is a bit wide, for example handtool should not clear the current selection. I will make the change after merging your commits.

aeghn:
> The "Select PDF Area Text" does not work. It does nothing at all. What it is supposed to do?

Sorry, it's a bug. fixed now.

> The toolbox moves with the hand if you select the hand tool after selecting some text. I think also the hand tool should close the toolbox

Maybe we should import some more logic to handle different cases, like to hide the floating toolbox when the page changed?

>  How are you supposed to remove an underline/highlight/strikethrough?

Actually, all markers was written into the xopp file, and they are like the other tools used to make markers in the xournalpp.
So, 
   - `Ctrl-Z` when you just made the underline/highlight/strikethrough
   -  Use rect/region select tool and delete them
   - Use the eraser

raffaem:
> It looks like the double/triple click and drag doesn't work at all now - only single word or single line selections are possible. Is that intentional? Also sometimes triple clicking doesn't register properly and it's treated as a double click. This might be caused by the popup toolbox stealing the last click.

What is double/triple clicking and dragging suppose to do?

Here it does nothing except moving the toolbox around

ronandalton:
> What is double/triple clicking and dragging suppose to do?

It highlights whole words/lines. I suggest trying it in other programs such as the web browser you are using to view this text right now.

ronandalton:
> What is dragging supposed to do?

Highlight multiple words/lines.

aeghn:
> Would it make sense to have the highlight a perfect rectangle without rounded borders?

@raffaem  fixed 6d814f5

aeghn:
> It looks like the double/triple click and drag doesn't work at all now - only single word or single line selections are possible. Is that intentional?

@ronandalton  I have added a workaround for this, please try it.  41e5d2d

ronandalton:
> > It looks like the double/triple click and drag doesn't work at all now - only single word or single line selections are possible. Is that intentional?
> 
> @ronandalton I have added a workaround for this(because poppler doesn't support this), please try it. [41e5d2d](https://github.com/xournalpp/xournalpp/commit/41e5d2db0d96cfc36c6b96620d54ca770b283815)

I tried it and it's working great except for if you don't drag the mouse and just double/triple click one spot, it doesn't work.

aeghn:
@Technius  @personalizedrefrigerator 

What is the logic for saving xopp files?

I tried to trace the `serialize()` method of `Stroke`, but I didn't catch it.

Thanks.

aeghn:
I won't add any more features into this PR, instead I will focus on:
 - fix the log/program bugs
 - adjust the illogical code structure
 - add more comments

If you find any problems, please do not hesitate and tell me immediately.

aeghn:
> Show an I cursor when selecting normally and + when using box select

fixed 06ada7c

LittleHuba:
We also experience such a delay when rendering LaTeX formulas. The delay seems to be specific to Windows. So if you manage to find a fix for this, we would be interested in porting this to our LaTeX editor.

raffaem:
> We also experience such a delay when rendering LaTeX formulas. The delay seems to be specific to Windows. So if you manage to find a fix for this, we would be interested in porting this to our LaTeX editor.

It is not specific to Windows because I'm on Fedora 34 Workstation and the previous video recording was take on Fedora 34 Workstation

raffaem:
is this going to be reviewed?

Technius:
Sorry, I've been busy recently and haven't had time to look at this.

You'll have to rebase your PR to match the new directory structure. Apologies for the trouble; let @Febbe know if you need further help.

@xournalpp/core Someone please take over review of this PR.

aeghn:
1. I found it wasn't easy to do rebase work with so many commits, so I merged them into one

2. The rendering issue mentioned before should have been fixed by other contributors @raffaem

raffaem:
It works wonderfully and the delay in the selection has been fixed.

@aeghn thank you so much.

I hope this can get reviewed soon.

> Here is a 5 second delay between the time I press the highlight button and the time the text is fully highlighted.

This seems to be now fixed.

src/core/control/tools/PdfElemSelection.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
```
personalizedrefrigerator, line 1:
```suggestion
/*
 * Xournal++
 *
 * Tool for selecting PDF content
 *
 * @author Xournal++ Team
 * https://github.com/xournalpp/xournalpp
 *
 * @license GNU GPLv2 or later
 */
#pragma once
```

src/core/gui/PdfFloatingToolbox.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include "control/Control.h"
+#include "control/tools/PdfElemSelection.h"
+#include "gui/XournalView.h"
+
+#include "GladeGui.h"
+
+
+class PdfFloatingToolbox {
```
personalizedrefrigerator, line 10: It might make sense to have both the floating toolbox tool and the PdfFloatingToolbox share a base class...

personalizedrefrigerator (changes requested):

personalizedrefrigerator:
Thank you for working on this!!! I should have time to contribute to this over the weekend!

aeghn:
> It might make sense to have both the floating toolbox tool and the PdfFloatingToolbox share a base class...

I'm trying to replace this floating toolbox with the popup toolbox that @Technius is talking about, because I think it could 

1. Maybe it should add a popover after jumping to another location via a link or toc to help the user go back (after this part is merged), I think they (pdf related tools) would be better using the same technique.
2. It looks like the popover is a bit more lightweight.

But I'm to busy these days,  please let me know how you think about it @personalizedrefrigerator

raffaem:
I still hope this can get reviewed :(

rolandlo:
I'm late to the party and reviewing the whole PR may be above my skill level, but I want to give at least some feedback. This looks like a great addition to Xournal++. Thanks so much for your work, you put into it @aeghn . Also thanks to @raffaem and @ronandalton for providing valuable feedback, which helped to fix several issues.
- As far as I can tell, it works all nicely (on Linux; will have to test on Windows and MacOS as well).
- I don't like the exit button in the floating toolbox. Is it really needed? You can simply click somewhere else on the page to exit.
- Pressing the `Escape` key should close the floating toolbox as well, I think.
- The "highlight selected text" icon looks inappropriate. I would have no clue what it does.
- The tooltip "Select with the other method" is too unspecific. It should instead name the other method depending on what method is currently used and ideally change the icon, too.
- I just got a crash pressing the "Select the other method" button. Here is the backtrace:
```- Date: Tue Jan  4 17:36:23 2022
Error: signal 6
[bt]: (0) ./xournalpp(+0x29267c) [0x562884c8667c]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x46520) [0x7f8beee2a520]
[bt]: (2) /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0xf8) [0x7f8beee7e808]
[bt]: (3) /lib/x86_64-linux-gnu/libc.so.6(raise+0x16) [0x7f8beee2a476]
[bt]: (4) /lib/x86_64-linux-gnu/libc.so.6(abort+0xd7) [0x7f8beee107b7]
[bt]: (5) /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa1a31) [0x7f8bef1ada31]
[bt]: (6) /lib/x86_64-linux-gnu/libstdc++.so.6(+0xad4ec) [0x7f8bef1b94ec]
[bt]: (7) /lib/x86_64-linux-gnu/libstdc++.so.6(+0xad557) [0x7f8bef1b9557]
[bt]: (8) /lib/x86_64-linux-gnu/libstdc++.so.6(+0xad7f9) [0x7f8bef1b97f9]
[bt]: (9) /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa4448) [0x7f8bef1b0448]
[bt]: (10) /lib/x86_64-linux-gnu/libstdc++.so.6(+0x148c61) [0x7f8bef254c61]
[bt]: (11) ./xournalpp(_ZN15PopplerGlibPage17selectAreaFinallyERK15XojPdfRectanglePP13_cairo_regionPSt6vectorIS0_SaIS0_EEPNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x810) [0x562884d50546]
[bt]: (12) ./xournalpp(_ZN16PdfElemSelection13selectFinallyEv+0xb4) [0x562884dc7558]
[bt]: (13) ./xournalpp(_ZN18PdfFloatingToolbox18switchSelectTypeCbEP10_GtkButtonPS_+0x91) [0x562884cfb2d7]
[bt]: (14) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x1290) [0x7f8bef60b970]
[bt]: (15) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f8bef60bad3]
``` 
Unfortunately I couldn't reproduce it a second time.
- I checked compatibility of the file format. Older versions of Xournal++ do open newly generated files (that have the capStyle field for strokes) fine. They just use the default rounded corners.
- Clang format should be corrected. Let us know if you need help with it.

That's it for the moment.

LittleHuba:
Thanks, @rolandlo for the high-level review. I'll do a code review in the next few days. @aeghn if you want to fix the things @rolandlo mentioned before my review, please write me a short note.

rolandlo:
Meanwhile I have tested the PR on MS Windows and MacOS as well. I didn't find any additional problems.

Since the icons in the floating toolbox depend on the icon theme used and I expressed dislike of the "highlight" button icon, let me show you how it really looks like (with the Yaru theme):
grafik

aeghn:
Given that the Chinese Lunar New Year is approaching and I need to finish my current work, I've been very busy lately and won't have much time to participate in.

In the meantime, if you feel the pr is still immature or poorly coded, please just close the pr, after about half a year I'm about to lose interest in it and I don't know if I have the patience to continue implementing my other ideas.

Thank you to all those who participated in this pr.

raffaem:
> Given that the Chinese Lunar New Year is approaching and I need to finish my current work, I've been very busy lately and won't have much time to participate in.
> 
> In the meantime, if you feel the pr is still immature or poorly coded, please just close the pr, after about half a year I'm about to lose interest in it and I don't know if I have the patience to continue implementing my other ideas.
> 
> Thank you to all those who participated in this pr.

I fear this is going to end like the support for Bluetooth audio codec in PulseAudio.

Technius:
I have some free time now; I've played around with the current code a bit and it works pretty well even for an initial version. I'm sure that some of us will agree that it's useful even if it's not perfect at the moment. Let's see if we can get this merged in a timely manner!

The changes I pushed above include:
* Rebased onto master again
* Cleaned up some of the code and ran the formatter
* Cleaned up some of the memory management

Here's what I'll do next:
* Split up the commits so it's easier to review. It's hard to reason about 1300 lines of code split up across 30 files.
* Replace the highlighter in the toolbox with the highlighter tool icon
* Further clean up the memory management--it's not clear what has ownership over the PDF element selection.

raffaem:
Tested again, it's working wonderfully.

Also @aeghn is pushing again, thanks :)

> Replace the highlighter in the toolbox with the highlighter tool icon

I can manage that. What I need to do?

Technius:
> Also @aeghn is pushing again, thanks :)

No, that's just me revising @aeghn's work and force pushing the changes.

> I can manage that. What I need to do?

I have that done, don't worry about it.

Technius:
@rolandlo I made some more UI changes. Specifically, I removed the exit button and changed the highlight iocn. Please take a look.

After thinking about the code for a bit, I can't think of any meaningful way to split up the PR further. Thus, we'll just have to clean up the code some more and merge afterwards.

Remaining items (@xournalpp/core feel free to edit this todo list):
* [ ] Identify and remove redundant methods from PdfFloatingToolbox and PdfElemSelection
* [ ] Clean up the new helper methods in PopplerGlibPage - they are too feature-focused to belong there
  * [ ] Add appropriate license text & changes to `copyright.txt` for the code copied from poppler
  * [ ] Refactor helper methods into PdfElemSelection
* [ ] Add code documentation and rename some methods to be more understandable

rolandlo:
> @rolandlo I removed the exit button and changed the highlight icon. Please take a look.

Looks good!

rolandlo:
> > @rolandlo I removed the exit button and changed the highlight icon. Please take a look.
> 
> Looks good!

Hitting the ESC key should exit though, which it doesn't currently.

Technius:
I cleaned up the PR significantly: code readability should be much improved, but at the cost of the selection being slightly less accurate than the previous version. I believe the quality is high enough for a merge now.

Everyone who helped out, please try the PR one last time. I will merge in 72 hours unless there are objections/concerns.

> Hitting the ESC key should exit though, which it doesn't currently.

This is a minor issue that would be best addressed in a follow up PR.

raffaem:
> Everyone who helped out, please try the PR one last time

I think it is not working anymore.

Here is what I get:

https://user-images.githubusercontent.com/54762742/151689771-dc076e12-4307-40a8-93e7-a09d24969b7c.mov

Here is what I expected to get, and what I got in the previous version, both in terms of selection (in blue color) and actual highlighting (in yellow color):

Screenshot from simple_aac_recording

Technius:
Thanks for the catch. I fixed some more issues caused by my refactoring:
* Area text selection should now be more accurate and copied text is now consistent with what you see. However, the selection highlight algorithm is slightly different; the selection highlight may be more prone to be being "broken up" compared to the previous version (functionally, it works fine).
* Fixed double/triple drag select not working.
* Fixed issues with the copied text being inconsistent with the selection highlight.
* [New bugfix!] Linear text selection (including double/triple click) now works correctly on double column articles.

raffaem:
I don't understand why if I select a word, the box surrounding it is all contiguous, whilch if I use the rectangle select tool I have gaps here and there.

Here is a video:

https://user-images.githubusercontent.com/54762742/151722995-4f59880c-608e-4ba5-b239-89f4ee7ddda6.mov

Here is a screenshot where the gaps have been circled:

Screenshot from 2022-01-31 00-43-30

> However, the selection highlight algorithm is slightly different; the selection highlight may be more prone to be being "broken up" compared to the previous version (functionally, it works fine).

I appreciate very much your intellectual honesty of admitting that your implementation is worse than @aeghn one, but why then we should retain your implementation instead of going back to @aeghn one? I don't understand that.

> Fixed double/triple drag select not working.

What is a double/triple "drag"? Do you mean double/triple click? How do I test this?

Technius:
> I don't understand why if I select a word (either by keeping the left mouse button down and dragging the mouse, or by double clicking) the box surrounding it is all contiguous, while if I use the rectangle select tool I have gaps here and there.

Linear selection uses Poppler's implementation. Poppler does not provide an API for area selection, so a custom implementation is required for area selection.

Both area selection implementations are based on `poppler_page_get_text_layout`, which provides a list of bounding boxes drawn around each symbol in the PDF. This means that we have to manually compute the areas that encapsulate each line. The original implementation constructs lines by grouping up bounding boxes that share the same lower y position (ignoring the "height" of the boxes). The new implementation is similar, but groups up boxes that share _both_ lower y position and height.

The difference is that the original implementation is more likely to place text on the same "line" into groups (as it does not check height), but this can lead to incorrect results (see below).

> I appreciate very much your intellectual honesty of admitting that your implementation is worse than @aeghn one, but why then we should retain your implementation instead of going back to @aeghn one? I don't understand that.

Two reasons: first, the original implementation is not very well documented and is fairly complex. I imagine that it would be difficult to maintain.

Secondly, the original implementation does not work correctly when there are inline symbols like code or equations, due to the way it groups up bounding boxes.



This has been fixed in the refactored version, which groups up symbols into lines based on _both_ lowest y position and height; a side effect is that some symbols may not be considered to be on the same "line" due to minor differences in the text layout. My opinion is that is better to have something that is functional than to have something that is pretty but not entirely functional.



PDFs are complicated and (consequently) libraries that deal with PDFs are not very good.

Technius:
> What is a double/triple "drag"? Do you mean double/triple click? How do I test this?

With the "Select Lienar PDF Text" tool, double/triple click on a word, and then drag the mouse to select by word/line.

raffaem:
@Technius  thanks for the clarifications.

I tested some more and generally it works and works well.

> With the "Select Linear PDF Text" tool, double/triple click on a word, and then drag the mouse to select by word/line.

Yeah it works for me. Thanks.

----

Please try to [download this PDF](https://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf).

Can you select only equation 108?

I was not able to select it without selecting the entire column, here is a video:

https://user-images.githubusercontent.com/54762742/152012520-feda4c01-aaf1-49c8-83ce-7511b504c57a.mov

Technius:
Since everything seems good, I'll be squashing the bugfix commits and merging in 24 hours. Biggest thanks go to @aeghn for the PR itself, and many thanks to everyone else for providing detailed feedback and testing.

> Can you select only equation 108?

Not with linear selection; area selection can select it to some extent (with gaps) This looks to be a Poppler issue; I get the same problem when I'm using Okular. I'm guessing this is because the equations are typeset in LaTeX as a table or an array.

raffaem:
🤩

src/core/pdf/popplerapi/PopplerGlibPage.cpp
```diff
@@ -78,3 +88,194 @@ auto PopplerGlibPage::findText(std::string& text) -> std::vector<XojPdfRectangle
...
+    PopplerRectangle* rectArray = nullptr;
+    guint numRects = 0;
+    if (style == XojPdfPageSelectionStyle::Area) {
+        PopplerRectangle area{rect.x1, rect.y1, rect.x2, rect.y2};
+        if (!poppler_page_get_text_layout_for_area(this->page, &area, &rectArray, &numRects)) {
+            return {.region = cairo_region_create(), .rects = textRects};
+        }
+    } else {
+        if (!poppler_page_get_text_layout(this->page, &rectArray, &numRects)) {
+            return {.region = cairo_region_create(), .rects = textRects};
```
Febbe, line 187: This is c++20 code and not allowed currently

Technius: Both gcc and clang support the C99 style designated initializers, and this code has already been merged.

PR #3333, opened 2021-08-19
Fixed #3330: Allow characters after final dot in filename

### LINKED ISSUE #3330: Request save as: Exercise.5.12  --saves--> Exercise.xopp (silently ignoring all after the first dot)
**Affects versions :**
 - OS: Archlinux
 - (Linux only) Desktop environment: X11 under i3
 - Which version of libgtk: 3.24.30
 - Version of Xournal++: 1.1.0
 - Installation method: official Arch repo

**Describe the bug**
Request for saving from File->Save[ as] (or Ctrl-S/Ctrl-U) under a

filename.with.dots

silently ignores everything after the first dot and saves under

filename.xopp.

**To Reproduce**
1. Start xournalpp with an empty canvas and scribble something
2. Choose menu File->Save and write in the filename field:  Exercise.5.12
3. Click save and exit xournalpp
4. Check what actually have been saved (e.g., ls) gives: Exercise.xopp

**Expected behavior**
Filenames with dots should be saved w/o truncation at the first dot.

Aniruddha-Deb:
This PR removes .pdf, .xopp and .xoj extensions (if any) and then appends xopp to the filename, rather than the simple replace_extension that was happening previously (because of which the text after the last dot in the filename was lost, as replace_extension used to replace that text thinking it was the extension).

Old code (in `SaveJob::save`)
```cpp
fs::path const filepath = doc->getFilepath();
auto const target = fs::path{filepath}.replace_extension(".xopp");
```
New code:
```cpp
fs::path filepath = doc->getFilepath();
Util::clearExtensions(filepath, ".pdf");
auto const target = fs::path{filepath}.concat(".xopp");
```

### DIFF
--- a/src/control/jobs/SaveJob.cpp
+++ b/src/control/jobs/SaveJob.cpp
@@ -87,10 +87,11 @@ auto SaveJob::save() -> bool {
 
     doc->lock();
     h.prepareSave(doc);
-    fs::path const filepath = doc->getFilepath();
+    fs::path filepath = doc->getFilepath();
     doc->unlock();
 
-    auto const target = fs::path{filepath}.replace_extension(".xopp");
+    Util::clearExtensions(filepath, ".pdf");
+    auto const target = fs::path{filepath}.concat(".xopp");
 
     if (doc->shouldCreateBackupOnSave()) {
         try {


personalizedrefrigerator:
This looks good to me! I can confirm that this fixes #3330.

LittleHuba:
Could you check how this behaves with .xoj import/export? I do remember that we already touched this once because it did not work as expected.

personalizedrefrigerator:
> Could you check how this behaves with .xoj import/export? I do remember that we already touched this once because it did not work as expected.

`.xoj` export/import still works with this change.

personalizedrefrigerator:
Thank you for your work! I'm merging this in 24 hours unless there are objections.

PR #3334, opened 2021-08-19
Allow Vertical Spacing tool to move elements above line

Technius:
Currently, the Vertical Spacing tool works as follows:
* The user clicks and holds to set an "anchor" line.
* By dragging down some distance `y`, all elements on the same page below the anchor line are pushed down by `y` units.
* If the user drags _up_ above the anchor line instead, all elements on the same page below the anchor line are _pulled_ up by `y` units.

**Updated PR description**:

This PR allows users to also select elements above the anchor line using Ctrl. The behavior is the same as when selecting below the line; i.e. if selecting below the line, the elements are translated in the same direction as the mouse movement and by the same amount.

TODO:
* [x] Select elements above anchor line using <kbd>Ctrl</kbd>.
* [x] Clean up the code

---

**Old PR description**:

This PR modifies the behavior of Vertical Spacing so that dragging above the anchor line by `y` units will instead push all elements _above_ the anchor line upwards by `y` units.

Comparison (**Note: the videos are outdated and correspond to the "Old PR description"**): 

https://user-images.githubusercontent.com/1066652/130134724-15af6906-0e29-4c4a-a0e6-ae4d421930a8.mp4

https://user-images.githubusercontent.com/1066652/130134743-86624a4f-9bf2-40bb-8151-4fb3a532d95b.mp4

### DIFF
--- a/src/core/control/tools/VerticalToolHandler.cpp
+++ b/src/core/control/tools/VerticalToolHandler.cpp
@@ -21,11 +21,7 @@ VerticalToolHandler::VerticalToolHandler(Redrawable* view, const PageRef& page, 
 
     for (Element* e: this->elements) {
         this->layer->removeElement(e, false);
-
-        this->jumpY = std::max(this->jumpY, e->getY() + e->getElementHeight());
     }
-
-    this->jumpY = this->page->getHeight() - this->jumpY;
 
     this->crBuffer = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, this->page->getWidth() * zoom,
                                                 (this->page->getHeight() - y) * zoom);
@@ -57,16 +53,8 @@ void VerticalToolHandler::paint(cairo_t* cr, GdkRectangle* rect, double zoom) {
 
     gdk_cairo_set_source_rgba(cr, &selectionColor);
 
-    double y = NAN;
-    double height = NAN;
-
-    if (this->startY < this->endY) {
-        y = this->startY;
-        height = this->endY - this->startY;
-    } else {
-        y = this->endY;
-        height = this->startY - this->endY;
-    }
+    const double y = std::min(this->startY, this->endY);
+    const double height = std::abs(this->startY - this->endY);
 
     cairo_rectangle(cr, 0, y * zoom, this->page->getWidth() * zoom, height * zoom);
 
--- a/src/core/control/tools/VerticalToolHandler.cpp
+++ b/src/core/control/tools/VerticalToolHandler.cpp
@@ -3,36 +3,28 @@
 #include <cmath>
 #include <memory>
 
+#include <cairo.h>
+
 #include "model/Layer.h"
 #include "undo/UndoRedoHandler.h"
 #include "view/DocumentView.h"
 
 VerticalToolHandler::VerticalToolHandler(Redrawable* view, const PageRef& page, Settings* settings, double y,
-                                         double zoom):
-        view(view), page(page), layer(this->page->getSelectedLayer()), snappingHandler(settings) {
+                                         bool initiallyReverse, double zoom):
+        view(view),
+        page(page),
+        layer(this->page->getSelectedLayer()),
+        spacingSide(initiallyReverse ? Side::Above : Side::Below),
+        snappingHandler(settings) {
     double ySnapped = snappingHandler.snapVertically(y, false);
     this->startY = ySnapped;
     this->endY = ySnapped;
-    for (Element* e: this->layer->getElements()) {
-        if (e->getY() >= y) {
-            this->elements.push_back(e);
-        }
-    }
+    this->zoom = zoom;
+    this->crBuffer = cairo_image_surface_create(
+            CAIRO_FORMAT_ARGB32, static_cast<int>(this->page->getWidth() * this->zoom),
+            static_cast<int>(std::max(this->startY, this->page->getHeight() - this->startY) * this->zoom));
 
-    for (Element* e: this->elements) {
-        this->layer->removeElement(e, false);
-    }
-
-    this->crBuffer = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, this->page->getWidth() * zoom,
-                                                (this->page->getHeight() - y) * zoom);
-
-    cairo_t* cr = cairo_create(this->crBuffer);
-    cairo_scale(cr, zoom, zoom);
-    cairo_translate(cr, 0, -y);
-    DocumentView v;
-    v.drawSelection(cr, this);
-
-    cairo_destroy(cr);
+    this->adoptElements(this->spacingSide);
 
     view->rerenderPage();
 }
@@ -46,6 +38,46 @@ VerticalToolHandler::~VerticalToolHandler() {
     }
 }
 
+void VerticalToolHandler::adoptElements(const Side side) {
+    this->spacingSide = side;
+
+    // Return current elements back to page
+    for (Element* e: this->elements) { this->layer->addElement(e); }
+    this->elements.clear();
+
+    // Add new elements based on position
+    for (Element* e: this->layer->getElements()) {
+        if ((side == Side::Below && e->getY() >= this->startY) ||
+            (side == Side::Above && e->getY() + e->getElementHeight() <= this->startY)) {
+            this->elements.push_back(e);
+        }
+    }
+
+    for (Element* e: this->elements) { this->layer->removeElement(e, false); }
+
+    cairo_t* cr = cairo_create(this->crBuffer);
+    cairo_scale(cr, this->zoom, this->zoom);
+
+    // Clear the buffer first
+    cairo_save(cr);
+    cairo_set_source_rgba(cr, 0, 0, 0, 0);
+    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
+    cairo_paint(cr);
+    cairo_restore(cr);
+
+    // if below, render elements translated so that startY is 0.
+    // if above, 0 is already the top of the page.
+    if (this->spacingSide == Side::Below) {
+        cairo_translate(cr, 0, -this->startY);
+    } else {
+        g_assert(this->spacingSide == Side::Above);
+    }
+    DocumentView v;
+    v.drawSelection(cr, this);
+
+    cairo_destroy(cr);
+}
+
 void VerticalToolHandler::paint(cairo_t* cr, GdkRectangle* rect, double zoom) {
     GdkRGBA selectionColor = view->getSelectionColor();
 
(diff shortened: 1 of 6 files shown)


Technius:
Looks like a gcc 8 bug is causing the Linux build to fail. It works fine with gcc 10.

bhennion:
> Allow user to press a modifier key to pull instead of push (is this actually useful though?).

I imagine a posible use case is what follows: The user wants to add something. They push some content down, write whatever they wanted to add. They now realize they had added too much space and want to pull stuff back up.

To be honest, it seems to me a more useful to pull up than to push up: the user often start taking notes at the top of the page and there is no room to push up. (but I personally never used the tool, so what do I know)

Technius:
Yeah, for this tool, I think the only important part is that there should be some way to move things up and some way to move things down. So the main issue here is: which is the best default?

For downwards movement, the best default is to push things down (status quo). My use case is similar to the one you mentioned: to insert additional space into the middle of a page of notes so that I can add some additional details.

> To be honest, it seems to me a more useful to pull up than to push up: the user often start taking notes at the top of the page and there is no room to push up.

I think upwards movement is debatable, but we can have that debate in a separate issue. On second thought, the current design is probably intentional, so I will change the behavior to be:

* Default down = push (same as old)
* Default up = pull (same as old)
* Modified down = pull (entirely new behavior)
* Modified up = push (entirely new behavior)

i.e., the modifier key will keep the direction but invert the action.

Technius:
Squashed and rebased onto master, with the new behavior added.

src/control/tools/VerticalToolHandler.cpp
```diff
@@ -50,45 +42,106 @@ VerticalToolHandler::~VerticalToolHandler() {
...
+        this->view->rerenderRect(0, std::min(oldEnd, ySnapped), this->page->getWidth(), this->page->getHeight());
+    } else {
+        this->view->rerenderRect(0, 0, this->page->getWidth(), std::max(oldEnd, ySnapped));
+    }
+}
+
+bool VerticalToolHandler::onKeyPressEvent(GdkEventKey* event) {
+    if ((event->keyval == GDK_KEY_Control_L || event->keyval == GDK_KEY_Control_R) && this->spacingSide > 0) {
+        this->adoptElements(-1.0);
+        this->view->rerenderPage();
```
LittleHuba, line 132: Could we reduce this to `rerenderRect`? Rendering the full page should be a last resort.

src/control/tools/VerticalToolHandler.cpp
```diff
@@ -50,45 +42,106 @@ VerticalToolHandler::~VerticalToolHandler() {
...
+        return true;
+    }
+    return false;
+}
+
+bool VerticalToolHandler::onKeyReleaseEvent(GdkEventKey* event) {
+    if ((event->keyval == GDK_KEY_Control_L || event->keyval == GDK_KEY_Control_R) && this->spacingSide < 0) {
+        this->adoptElements(1.0);
+        this->view->rerenderPage();
```
LittleHuba, line 141: Could we reduce this to `rerenderRect`? Rendering the full page should be a last resort.

src/control/tools/VerticalToolHandler.h
```diff
@@ -42,15 +55,25 @@ class VerticalToolHandler: public ElementContainer {
...
-     * When we create a new page
+     * Indicates whether to move elements above (negative) or below (positive) the anchor line.
+     */
+    double spacingSide = 0;
+
+    /**
+     * Current zoom level.
      */
-    double jumpY = 0;
+    double zoom;  // TODO: listen to zoom change in case zoom is changed during spacing
```
LittleHuba, line 76: This TODO should be done before we consider merging this.

src/control/tools/VerticalToolHandler.cpp
```diff
@@ -50,45 +42,106 @@ VerticalToolHandler::~VerticalToolHandler() {
     }
 }
 
+void VerticalToolHandler::adoptElements(double side) {
```
LittleHuba, line 45: Please use an enum for this.

LittleHuba (changes requested):

src/control/tools/VerticalToolHandler.cpp
Technius: It's not obvious to me how that would be done. All elements below the line are reset, requiring the region below the line to be rerendered. Then all elements above the line are shifted, requiring the region above the line to be rerendered.

Technius:
@xournalpp/core Could someone take over this PR? Thanks in advance.

Technius:

rolandlo:
I got a signal 6 crash after a while using the vertical tool:
```shell
Date: Sun Feb  6 21:13:54 2022
Error: signal 6
[bt]: (0) ./xournalpp(+0x28c4dc) [0x5617d3e044dc]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x46520) [0x7f833a206520]
[bt]: (2) /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0xf8) [0x7f833a25a808]
[bt]: (3) /lib/x86_64-linux-gnu/libc.so.6(raise+0x16) [0x7f833a206476]
[bt]: (4) /lib/x86_64-linux-gnu/libc.so.6(abort+0xd7) [0x7f833a1ec7b7]
[bt]: (5) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x1eb7c) [0x7f833a89eb7c]
[bt]: (6) ./xournalpp(_ZN19VerticalToolHandler12redrawBufferEv+0x60) [0x5617d3fd857c]
[bt]: (7) ./xournalpp(_ZN19VerticalToolHandler13adoptElementsENS_4SideE+0x243) [0x5617d3fd8505]
[bt]: (8) ./xournalpp(_ZN19VerticalToolHandlerC2EP10RedrawableRKSt10shared_ptrI7XojPageEP8SettingsdP11ZoomControlP10_GdkWindow+0x187) [0x5617d3fd8163]
[bt]: (9) ./xournalpp(_ZN11XojPageView18onButtonPressEventERK17PositionInputData+0x7c2) [0x5617d3f4b8dc]
[bt]: (10) ./xournalpp(_ZN15PenInputHandler11actionStartERK10InputEvent+0x2f7) [0x5617d3e828d3]
[bt]: (11) ./xournalpp(_ZN17MouseInputHandler10handleImplERK10InputEvent+0x5a) [0x5617d3e81dec]
[bt]: (12) ./xournalpp(_ZN20AbstractInputHandler6handleERK10InputEvent+0x71) [0x5617d3f78f8f]
[bt]: (13) ./xournalpp(_ZN12InputContext6handleEP9_GdkEvent+0x3f3) [0x5617d3e7f475]
[bt]: (14) ./xournalpp(_ZN12InputContext13eventCallbackEP10_GtkWidgetP9_GdkEventPS_+0x2b) [0x5617d3e7f07f]
[bt]: (15) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3e6cf8) [0x7f833b489cf8]
[bt]: (16) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x1290) [0x7f833a9e7970]
[bt]: (17) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f833a9e7ad3]
[bt]: (18) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3ae43b) [0x7f833b45143b]
[bt]: (19) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x2514e0) [0x7f833b2f44e0]
[bt]: (20) /lib/x86_64-linux-gnu/libgtk-3.so.0(gtk_main_do_event+0xd3a) [0x7f833b2f538a]
[bt]: (21) /lib/x86_64-linux-gnu/libgdk-3.so.0(+0x39763) [0x7f833afd7763]
[bt]: (22) /lib/x86_64-linux-gnu/libgdk-3.so.0(+0x70b26) [0x7f833b00eb26]
[bt]: (23) /lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_context_dispatch+0x26b) [0x7f833a8d58bb]
[bt]: (24) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0xa8f08) [0x7f833a928f08]
[bt]: (25) /lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_context_iteration+0x33) [0x7f833a8d3003]
[bt]: (26) /lib/x86_64-linux-gnu/libgio-2.0.so.0(g_application_run+0x19d) [0x7f833ab1ec1d]
[bt]: (27) ./xournalpp(_ZN11XournalMain3runEiPPc+0x723) [0x5617d3e0b6d5]
[bt]: (28) ./xournalpp(main+0x29) [0x5617d3e04342]
[bt]: (29) /lib/x86_64-linux-gnu/libc.so.6(+0x2dfd0) [0x7f833a1edfd0]
[bt]: (30) /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x7d) [0x7f833a1ee07d]
[bt]: (31) ./xournalpp(_start+0x25) [0x5617d3e04255]
```

Terminal messages:
```shell
** Message: 21:09:42.899: Plugin "Example" UI initialized
** Message: 21:12:43.650: Info: autosave document...
** Message: 21:12:43.650: Autosaving to /home/roland/.cache/xournalpp/autosaves/70064.autosave.xopp
**
ERROR:/home/roland/Dokumente/Programmierung/Reviews/xournalpp/src/core/control/tools/VerticalToolHandler.cpp:62:void VerticalToolHandler::redrawBuffer(): 'this->crBuffer' should not be nullptr
Bail out! ERROR:/home/roland/Dokumente/Programmierung/Reviews/xournalpp/src/core/control/tools/VerticalToolHandler.cpp:62:void VerticalToolHandler::redrawBuffer(): 'this->crBuffer' should not be nullptr

** (xournalpp:70064): WARNING **: 21:13:54.277: [Crash Handler] Crashed with signal 6

** (xournalpp:70064): WARNING **: 21:13:54.288: [Crash Handler] Wrote crash log to: /home/roland/.cache/xournalpp/errorlogs/errorlog.20220206-211354.log

** (xournalpp:70064): WARNING **: 21:13:54.566: Trying to emergency save the current open document…

** (xournalpp:70064): WARNING **: 21:13:55.429: Successfully saved document to "/home/roland/.config/xournalpp/emergencysave.xopp"
```

rolandlo:
I can reproduce the crash consistently by starting up Xournal++ (with a fresh document) and use the vertical space tool to move in both directions .

Technius:
I rebased on latest commit and fixed the issue.

Technius:
The CTRL problem should be fixed now.

rolandlo:
The CTRL problem is fixed indeed. I have found a new problem though, which appears when I use 200% scaling. When I use the vertical space tool, only part of the objects to move are rendered. When I release the mouse they are rendered fully again.

Technius:
> The CTRL problem is fixed indeed. I have found a new problem though, which appears when I use 200% scaling. When I use the vertical space tool, only part of the objects to move are rendered. When I release the mouse they are rendered fully again.

I can't reproduce your problem even when I use the mouse wheel to zoom in and out while using the vertical space tool. Can you provide a video showing the problem?

rolandlo:
> > The CTRL problem is fixed indeed. I have found a new problem though, which appears when I use 200% scaling. When I use the vertical space tool, only part of the objects to move are rendered. When I release the mouse they are rendered fully again.
> 
> I can't reproduce your problem even when I use the mouse wheel to zoom in and out while using the vertical space tool. Can you provide a video showing the problem?

I was talking about DPI scaling, not about zoom. But OK, I will create a video.

rolandlo:
renderingProblem

Technius:
Thanks for the catch. I didn't know that `gdk_window_create_similar_image_surface` does not actually scale the buffer size.

rolandlo:
One more (and probably last) issue I discovered: the vertical space tool mixes up the insertion order of elements. It should preserve it, I think. Otherwise even if I don't move anything at all, the insertion order is mixed up and doesn't even get restored when hitting CTRL+Z. See the dark blue line in the following screen recording:

insertionProblem

Technius:
The original implementation is also affected, so that bug is out of scope of this PR. Will tackle in a follow-up.

Merging in 24 hours unless there are further objections.

PR #3344, opened 2021-08-24
Fix a bug that makes single point stroke pressureless

### LINKED ISSUE #1534: No pressure sensitivity for dots
**Affects versions :**
 - OS: Linux (Fedora 30)
 - Gnome-Wayland 3.32.2
 - Which version of libgtk do you use: not sure
 - Version of Xournal++: 1.0.10, version `0+git.26c7205f` from snapcraft.io

**Describe the bug**
When trying to draw small dots using the stylus, the resulting dots always have the same size (which is too big), hence ignoring the stylus pressure, as can be seen on the screenshot below. They also seem to appear with a small delay, like when using the shape recogniser.
I cannot figure out if this is a feature, an issue with my tablet, or some X++ bug. It affect the original Xournal too.

**To Reproduce**
1. Quickly tap the tablet/screen with the stylus to produce a dot.
2. For reference, repeat step 1, but this time move the cursor very slightly while applying pressure.

**Expected behavior**
Step 1 generates a small dot, with a size similar to normal strokes and with pressure sensitivity. The result should look similar to that of step 2.

**Actual behaviour**
Step 1 generates a large, fixed size dot like in the screenshot below. 

**Screenshot of Problem**
Screenshot from 2019-10-11 18-26-28

**Attachments**

* Zipped `.xopp` file: [Dots.zip](https://github.com/xournalpp/xournalpp/files/3718748/Dots.zip)

bhennion:
This supposedly fixes #1534:

dots

Some workarounds are needed for pressureless devices as well and pressure inference. I will add them later if this works on all available pressure sensitive devices.

Testing on various devices is necessary. For anyone trying this PR out:

- it's easier to get single-point strokes when zoomed in to maximum.
- if you open `Xournal++` from a terminal, you get a "single dot" message for every such single-point stroke.

### DIFF
--- a/src/control/tools/StrokeHandler.cpp
+++ b/src/control/tools/StrokeHandler.cpp
@@ -76,17 +76,26 @@ void StrokeHandler::paintTo(const Point& point) {
         Point endPoint = stroke->getPoint(pointCount - 1);
         double distance = point.lineLengthTo(endPoint);
         if (distance < PIXEL_MOTION_THRESHOLD) {  //(!validMotion(point, endPoint)) {
+            if (pointCount == 1 && this->hasPressure && endPoint.z < point.z) {
+                // Record the possible increase in pressure for the first point
+                // Nb: at this point, neither point.z nor endPoint.z has been multiplied by this->stroke->getWidth()
+                this->stroke->setLastPressure(point.z);
+                this->firstPointPressureChange = true;
+
+                // Paint the dot for the user to see
+                this->paintDot(endPoint.x, endPoint.y, this->stroke->getWidth() * point.z);
+            }
             return;
         }
         if (this->hasPressure) {
             /**
              * Both device and tool are pressure sensitive
              */
-            if (endPoint.z != Point::NO_PRESSURE) {
-                /**
-                 * Avoid issues at the beginning of the stroke
-                 */
-
+            if (this->firstPointPressureChange) {
+                // Avoid shrinking if we recorded a higher pressure event at the beginning of the stroke
+                this->firstPointPressureChange = false;
+                stroke->setLastPressure(std::max(endPoint.z, point.z) * stroke->getWidth());
+            } else {
                 if (const double widthDelta = (point.z - endPoint.z) * stroke->getWidth();
                     - widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
                     /**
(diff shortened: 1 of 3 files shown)


ViniciusGiroto:
Seems to be working. It is kinda hard to draw the thicker dots, but that is probably my inability.

image
(some of these are not dots)

Tested with a Wacon One on Fedora 34

bhennion:
Uh... The seems to be an issue downloading portaudio for Windows, making the build fail
```
--2021-08-24 19:59:47--  (try:20)  http://files.portaudio.com/archives/pa_stable_v190700_20210406.tgz
Connecting to files.portaudio.com (files.portaudio.com)|107.22.210.58|:80... failed: Connection timed out.
Giving up
```
Hopefully it's just a server down. I will rerun it in a day.

personalizedrefrigerator:
> So just a fixed dot width for when pressure inference is on then. I still don't know how to find out if the device is using pressure inference from within StrokeHandler though.

It looks like there's a bug in pressure inference!
image

For large <i>Δt</i>s (Δt is in tenths of a second), pressure inference estimates the same pressure.

Note that previously, the maximum `dt` considered by pressure inference was `2.0` (0.2 seconds):
```c++ 
...

        double dt = std::min((pos.timestamp - lastPos.timestamp) / 10.0, 2.0);
        double distance = utl::Point<double>(pos.x, pos.y).distance(utl::Point<double>(lastPos.x, lastPos.y));
        double inverseSpeed = (dt) / (distance + 0.1);

...
```

Even with the `std::min` removed, however, it's still giving the same final pressure (on mouse up). As such, the solution here is probably fixing pressure inference for large `Δt`s.

personalizedrefrigerator:
Even with a better formula, however, the single-point stroke doesn't seem to change size with respect to the second point's pressure...

bhennion:
> Even with a better formula, however, the single-point stroke doesn't seem to change size with respect to the second point's pressure...

That's to be expected with the previous commits. With pressure inference on, you are (for now) supposed to get the same dot as with pressure value 1 every time.
The ButtonRelease event's pressure was never read (on the devices I have access to, this event's pressure value is always 0).

I added a line testing if this value is bigger than the recorded pressure. I also added a commit with your patch.
It seems to works. The bigger question is: how does it feels when writing using pressure inference? Are the dots "the right size"?

LittleHuba (approved):
Merge in 24 hours into v1.1.1 if there are no objections.

PR #3350, opened 2021-08-29
Fixed 'PDF background missing' text showing

### LINKED ISSUE #3185: "PDF background missing" displayed behind transparent image background
**Affects versions :**
 - OS: Linux (doesn't matter)
 - X11 (doesnt' matter) 
 - Which version of libgtk do you use: 3.24.25 (doesn't matter)
 - Version of Xournal++: 1.1.0+dev, 1.0.19 (probably all versions)
 - Installation method: PPA (doesn't matter)

**Describe the bug**
The message "PDF background missing" is not only displayed on PDF pages where the background is missing, but also on pages with image background. The message is only visible if the image background is transparent.

**To Reproduce**
1. Open a new document
2. Choose an image background with a transparent image
3. Observe that behind the image the message "PDF background missing" is displayed

**Expected behavior**
No such message should be displayed for image backgrounds.

**Screenshots of Problem**
grafik

ronandalton:
Stops the 'PDF background missing' warning text from being shown when a
transparent background image is used. Fixes #3185.

### DIFF
--- a/src/control/jobs/RenderJob.cpp
+++ b/src/control/jobs/RenderJob.cpp
@@ -114,7 +114,7 @@ void RenderJob::run() {
         int height = this->view->page->getHeight();
 
         bool backgroundVisible = this->view->page->isLayerVisible(0);
-        if (backgroundVisible) {
+        if (backgroundVisible && this->view->page->getBackgroundType().isPdfPage()) {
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
This seems to fix the issue! @rolando merging in 24 hours if there are no objections.

rolandlo:
> This seems to fix the issue! Merging in 24 hours if there are no objections.

I agree.

PR #3355, opened 2021-08-31
Fix inconsistent display of filled highlighter strokes

### LINKED ISSUE #2904: Inconsistent Behaviour of Highlighter When Filling Is Activated
**Affects versions :**
 - Kubuntu 20.10 (5.8.0-44-generic)
 - Xournal++ 1.1.0+dev libgtk: 3.24.23
 - Installation method: PPA (I think)
 - I also remember seeing the problem on a windows system a few months ago, but currently don't have access to a windows system.

**Describe the bug**
The highlighter combined with the fill function sometimes produces borders and sometimes not, depending on the color selected.

**To Reproduce**
1. Activate the highlighter and filler
2. Draw a cirlce with the bright green -> You won't get a border.
3. Draw a circle with the dark green -> You will get a border.

**Expected behavior**
No boundary would be best, but if there is a reason for these boundaries, it would make sense for all highlighters to create a boundary. In short, the behaviour should be consistent.

**Screenshots of Problem**
Example with the rectangle tool:
ezgif-2-ff0e13ed5ce1

Example with hand-drawn circle:
ezgif-2-bb3f2f646dfb

**Additional context**
Maybe this is feature I don't understand :smile:. Feel free to close the issue if this is the case, but even then I think it would be great to have a option to make this consistent (in other words turn the borders off).

bhennion:
Fix #2904

This is a quick fix. Some refactoring of `StrokeView` would also be beneficial to improve the logic, optimize and reduce the computational cost, but that'd require more work and goes beyond the scope of this simple bugfix PR.

### DIFF
--- a/src/view/StrokeView.cpp
+++ b/src/view/StrokeView.cpp
@@ -80,6 +80,8 @@ void StrokeView::drawNoPressure() {
         // Do not apply the alpha here, else the border and the fill
         // are visible instead of one homogeneous area
         DocumentView::applyColor(cr, s, 255);
+        // The operator will be reset (to CAIRO_OPERATOR_MULTIPLY) by cairo_pop_group_to_source below
+        cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
         drawFillStroke();
         group = true;
     }
@@ -95,7 +97,6 @@ void StrokeView::drawNoPressure() {
 
     if (group) {
         cairo_pop_group_to_source(cr);
-
         if (noAlpha) {
             // Currently drawing -> transparent applied on blitting
             cairo_paint(cr);


src/view/StrokeView.cpp
```diff
@@ -80,6 +80,8 @@ void StrokeView::drawNoPressure() {
         // Do not apply the alpha here, else the border and the fill
         // are visible instead of one homogeneous area
         DocumentView::applyColor(cr, s, 255);
+        // The operator will be reset (to CAIRO_OPERATOR_MULTIPLY) by cairo_pop_group_to_source below
+        cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
```
LittleHuba, line 84: What if `group` is false?

personalizedrefrigerator: We're setting group to `true` on line 86. Moving this to just after line 79 might make this clearer:
```cpp
    cairo_push_group(cr);
    group = true;
    ... Lines 80-84 ...
    drawFillStroke();
}
```

bhennion:
Ok, I changed the logic of `StrokeView::drawNoPressure` to make things more clear. It also saves a loop through the stroke's points for filled highlighter.

LittleHuba (approved):
LGTM - This class needs some serious cleanup but that is outside of the scope of this PR.

PR #3357, opened 2021-09-01
Implement Page>Edit

personalizedrefrigerator:
## Summary
 * Previously `Page [Right Click]> Edit` was labeled `Edit (not implemented yet)`
Edit wasn't implemented
 * `Edit` now calls `Control::pageFormat()`
 * `pageFormat` now shows a dialog similar to the page template dialog, allowing for background customization, in addition to size
 
Single page config

 * May still need some cleanup


src/gui/dialog/pageFormat/PageTemplateDialog.cpp
```diff
@@ -0,0 +1,239 @@
+#include "PageTemplateDialog.h"
+
```
personalizedrefrigerator, line 2: This file is a renamed & refactored version of `PageTemplateDialog.cpp`

LittleHuba:
I think it would be better if we went a completely different route with the Page Template Dialog. Could you merge this dialog with the Preferences Dialog?

personalizedrefrigerator:
> I think it would be better if we went a completely different route with the Page Template Dialog. Could you merge this dialog with the Preferences Dialog?

To clarify, are you suggesting that we keep the new `Page Format` dialog as it is in this PR, while moving `Page Template` to the settings dialog? If so, I agree that this makes sense!

LittleHuba:
> To clarify, are you suggesting that we keep the new Page Format dialog as it is in this PR, while moving Page Template to the settings dialog?

Yes, that is what I propose. Currently, all global settings are within the preferences dialog except the Page Template. This is quite confusing if you are new to Xournal++.

personalizedrefrigerator:
> > To clarify, are you suggesting that we keep the new Page Format dialog as it is in this PR, while moving Page Template to the settings dialog?
> 
> Yes, that is what I propose. Currently, all global settings are within the preferences dialog except the Page Template. This is quite confusing if you are new to Xournal++.

Button to open the page format dialog has been added to settings

I've added a button that opens the page template dialog to settings. Although adding the contents of the dialog directly to the settings window would be ideal, that is more refactoring than I would like to do at the moment. Perhaps this can be done in a follow-up PR.

LittleHuba:
> Currently the "Saved Separately" behavior that results from this could be confusing. As such, I am unsure whether moving the contents of the dialog to the settings panel should be delayed until a future PR.

Yes with that in mind it would be better to postpone it to a future PR instead of making the situation even more complex before solving it completely.

PR #3359, opened 2021-09-02
Hide, rather than disable, contents tab when unavailable

personalizedrefrigerator:
# Summary
 * Hides, rather than disables, the "contents" tab in the sidebar when the tab would be empty
 * This saves space in the tab area at the top of the toolbar

## XOPP Without PDF Background
Contents tab is invisible when not editing a PDF with contents

## PDF Without Bookmarks
Contents tab still invisible: PDF with no table of contents

## PDF With Bookmarks
Contents tab is shown


Technius (approved):

personalizedrefrigerator:

PR #3364, opened 2021-09-04
Extend Lua API to get info on tools

### LINKED ISSUE #3332: Feature request: A way to track what tool is active through the Plugin API
My drawing tablet stylus buttons do not work in the program, so I have developed a plugin to reproduce eraser toggling behavior. However, this will only work if I hard code which tools to switch between (e.g., pen and eraser, highlighter and eraser, etc.), and thus breaks if I were to switch from the pen to the highlighter, for example. There is no way to track what tool is selected, and hence I created this request. 

Add a function to the Plugin API which returns the active tool.
Example:
```Lua
activeTool = app.getActiveTool()
print(activeTool)
if activeTool == "TOOL_PEN" then
...
```

I've tried binding two separate buttons, one to select the pen, and one to select the eraser, in my drawing tablet software, but found this to be uncomfortable and unnatural. In my research, I found that the stylus button issues should be fixed, but I have tried multiple different configurations with my stylus buttons in Xournal++ and no solution has arisen.

**Additional context**
I am using a HUION Inspiroy H1060P if it means anything.
For reference, my plugin implementation is here: [EraserToggle](https://github.com/dylif/EraserToggle)

rolandlo:
New API:
```lua
app.getToolInfo("active")
app.getToolInfo("pen")
app.getToolInfo("highlighter")
app.getToolInfo("eraser")
app.getToolInfo("text")
```

Example output (from the test plugin):

```
active tool: eraser
active size (name): thin
active size (value): 2.83
active color: 0x000000
active fill opacity: -1
active drawing type: default
active line style: plain
pen size (name): medium
pen size (value): 1.41
pen filled: false
pen fill opacity: 127
pen color: 0x00c0ff
pen drawing type: strokeRecognizer
pen line style: dash
highlighter size (name): medium
highlighter size (value): 8.5
highlighter filled: false
highlighter fill opacity: 87
highlighter color: 0x00c0ff
highlighter drawing type: arrow
eraser type: default
eraser size (name): thin
eraser size (value): 2.83
text font (name): Sans
text font (size): 20.0
text color: 0x00c0ff
```

Closes #3332

**Edit:** updated example output and API function calls

### DIFF
--- a/src/plugin/luapi_application.h
+++ b/src/plugin/luapi_application.h
@@ -21,6 +21,8 @@
 #include "control/pagetype/PageTypeHandler.h"
 #include "gui/XournalView.h"
 #include "gui/widgets/XournalWidget.h"
+#include "model/Font.h"
+#include "model/StrokeStyle.h"
 #include "model/Text.h"
 
 #include "StringUtils.h"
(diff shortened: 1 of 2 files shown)


src/plugin/luapi_application.h
```diff
@@ -475,6 +476,164 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
     return 1;
 }
 
+/*
+ * Returns a table encoding all info on the pen in a Lua table of the shape
+ * {
+ *   "size" = string
+_*   "fill" = bool
```
personalizedrefrigerator, line 483:
```suggestion
 *   "fill" = bool
```

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,164 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
...
+    lua_settable(L, -3);
+
+    return 1;
+}
+
+/*
+ * Returns a table encoding all info on the highlighter in a Lua table of the shape
+ * {
+ *   "size" = string
+_*   "fill" = bool
```
personalizedrefrigerator, line 519:
```suggestion
 *   "fill" = bool
```

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,164 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
...
+ * {
+ *   "tool" = string
+ *   "color" = integer
+ *   "fillOpacity" = integer,
+ *   "drawingType" = string
+ *   "lineStyle" = string
+ *   "thickness" = number
+ * }
+ *
+ * Example: local activeToolInfo = app.getActiveToolInfo()
```
personalizedrefrigerator, line 593: Although perhaps better suited for the website, documenting some of the possible values for 'tool', 'color', etc. could be helpful here.
```suggestion
 * Example 1: local activeToolInfo = app.getActiveToolInfo()
 * Example 2: app.getActiveToolInfo()["tool"]
 *  - Returns the name of the active tool. See /src/control/ToolEnums.cpp#toolTypeToString for possible values.
```

personalizedrefrigerator (approved):
Thank you for working on this! I like the API.

The comments below either correct typos or are documentation-related and subjective.
This will be nice to have!

rolandlo:
@personalizedrefrigerator 
Can you please take another look. I tried to improve the documentation a bit, renamed "tool" to "type" for the active tool and added the "size" key to the active tool.
Clang format should be Ok. Looks like a temporary thing, since Azure pipelines cannot find `clang-format-12` currently.

```
E: Unable to locate package clang-format-12
##[error]Bash exited with code '100'.
Finishing: Install clang-format
```

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,194 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
     return 1;
 }
 
+/*
+ * Returns a table encoding all info on the pen in a Lua table of the shape
+ * {
+ *   "size" = string
+ *   "fill" = bool
```
personalizedrefrigerator, line 483:
```suggestion
 *   "filled" = bool
```

I think `filled` makes more sense here; I associate `fill` with `fill details` (e.g. fill opacity, fill color, etc).

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,194 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
...
+
+    return 1;
+}
+
+
+/*
+ * Returns a table encoding all info on the active tool in a Lua table of the shape
+ * {
+ *   "type" = string
+ *   "size" = string
```
personalizedrefrigerator, line 604: It might make more sense to make `thickness` a part of size:
```lua
  size = {
    name = string,
    value = number,
  },
```

Although we're currently using `size` to refer to an enum value in the codebase and `thickness` to refer to the corresponding line thickness, I think grouping them might make the API easier to understand.

Possible issue: If we add `app.setActiveToolInfo(info)`, a size of ` { name = small, thickness = 1e8 }` could be invalid. Does it set the thickness to `1e8` or use the thickness associated with `small`?

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,194 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
...
+ *   "color" = integer
```
personalizedrefrigerator, line 605: It might make sense to break `color` up into `red`, `green`, and `blue` components. Alternatively, document how to access the different components of `color`.

src/plugin/luapi_application.h
```diff
@@ -475,6 +476,194 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
...
+ *   "fillOpacity" = integer
```
personalizedrefrigerator, line 606: From this line, it isn't clear that `fillOpacity` ranges from `0` to `255` (inclusive).

Additionally, it might make sense to make `fillOpacity` range from `0.0` to `1.0`, instead. This seems more intuitive (and more closely matches the GUI where `fillOpacity` ranges from `0.0` to `100.0`).

personalizedrefrigerator: If we break `color` up into `{ red = number, green = number, blue = number }`, we should also have each component have the same range as `fillOpacity`.

src/plugin/luapi_application.h
rolandlo: I think we should remain consistent with how we pass colors in `app.changeToolColor`
```lua
app.changeToolColor({["color"] = 0xff00ff, ["tool"] = "PEN"})
```
From the current implementation one can get the string representation via 
```lua
local color = app.getActiveToolInfo()["color"]
local hexString = string.format("0x%06x", color)
```
and from there one can extract the r,g,b values (between 0 and 255) via
```lua
red = tonumber("0x" .. string.sub(hexString,3,4))
green = tonumber("0x" .. string.sub(hexString,5,6))
blue = tonumber("0x" .. string.sub(hexString,7,8))
```

So I think we should just add it to the documentation (in one of the examples)

src/plugin/luapi_application.h
rolandlo: I agree that it should be clarified that `fillOpacity` ranges from 0 to 255. I am not so happy with converting it to the 0.0 to 1.0 range when it is internally stored as an integer from 0 to 255. There may be rounding problems with the conversion.

src/plugin/luapi_application.h
rolandlo: I need to take a look at the code of what name/value pairs are currently possible and what #3288 will change before I can answer on this.

src/plugin/luapi_application.h
rolandlo: Or more directly:
```lua
local color = app.getActiveToolInfo()["color"]
local red = color >> 16 & 0xff
local green = color >> 8 & 0xff
local blue = color & 0xff
```

src/plugin/luapi_application.h
personalizedrefrigerator: > There may be rounding problems with the conversion.

Okay! That makes sense!

src/plugin/luapi_application.h
personalizedrefrigerator: Okay. That makes sense.

It might also be worth [creating a `Color` type for `lua`](https://www.lua.org/pil/28.html) with `__tonumber` and `getR()`, `getG()`, and `getB()` methods. In this case, `changeToolColor` could accept an `integer` or a `xournalpp.Color`.

The benefits of a custom type:
 * Allows us to add additional methods to `Color` in the future (e.g. hueRotate, invert, etc).
 * Easier to get/set `RGB` components.

Some downsides:
 * More complicated to implement.
   * I don't think we're defining custom types via the plugin API yet...
 * Need to update `app.changeToolColor`

personalizedrefrigerator: Overall, I think using a number is fine here. Let's just make sure to document it on the website :).

src/plugin/luapi_application.h
rolandlo: @personalizedrefrigerator I agree with the idea to make `thickness` part of size and applied it in my last commit. Moreover I added the info about the thickness of the pen, highlighter and eraser.
The issues you mentioned don't look too critical to me.
- There is no need to add `app.setActiveToolInfo`, since those properties can (and should) be set through `app.uiAction` and `app.changeToolColor`. The thickness is an exception and we will have to add API to set the custom thickness implemented in #3288 
- Looking at the current implementation of #3288 suggests to me that getting the info on the individual tool sizes will continue to work as expected.

personalizedrefrigerator:
LGTM!

nicolae-stroncea:
@rolandlo would `getActiveToolInfo` ever return *Text* as the active tool?

rolandlo:
> @rolandlo would `getActiveToolInfo` ever return _Text_ as the active tool?

It would return "text" as the active tool, yes. You can find the list of tools here:

https://github.com/xournalpp/xournalpp/blob/623d8654bacbc8db264f2cf9a76aead20be26889/src/control/ToolEnums.cpp#L96-L137

nicolae-stroncea:
I think an `app.getTextInfo()` may be also be useful, probably in the same contexts as the other `getInfo`'s

rolandlo:
> I think an `app.getTextInfo()` may be also be useful, probably in the same contexts as the other `getInfo`'s

What information should `app.getTextInfo()` provide?

nicolae-stroncea:
I would say information regarding the customizable options of the Text tool:
* Color
* Font size
* Font type

Would be useful for any plugin that reacts to the tool settings and then maybe modifies them in some way

rolandlo:
> I would say information regarding the customizable options of the Text tool:
> 
>     * Color
> 
>     * Font size
> 
>     * Font type
> 
> 
> Would be useful for plugins that reacts to the tool settings and then maybe modifies them in some way

I have added that to the API in the latest commit.

Technius (commented):
Would it be possible to define a single `getExtraToolInfo(toolname)` (or some better name) function instead of having so many `get${TOOL}Info` functions?

rolandlo:
> Would it be possible to define a single `getExtraToolInfo(toolname)` (or some better name) function instead of having so many `get${TOOL}Info` functions?

@Technius Done!

LittleHuba:
> With regards to adding tests: Can that wait until we have decided on the testing framework?

Yes, that seems reasonable. Though we should not forget about this. Our Testing suites are insufficient and should have high priority.

rolandlo:
Rebased on current master and squashed into two commits, where the second commit just removes the API testing from the first commit in the example plugin (where it does not really fit). Use the first commit to try out the API.

Merging in 24 hours into master if no objections are raised.

PR #3372, opened 2021-09-06
Fix touchscreen zooming when using vertical/horziontal spacing. Fixes #3298

### LINKED ISSUE #3298: Centered pinch-to-zoom
While zooming with the scroll wheel the page zooms around the mouse pointer. Ideally, when zooming with pinch-to-zoom the zoom would be centred about the midpoint of the user's fingers. Currently, the zoom is relative to the top-left corner of the document.

robinp7720:
The spacing compensation was added to the zoom center in touchInputHandler and in ZoomControl when starting a touch pinch gesture. This caused the pinch to zoom gesture to not be centered on the zoom origin when using vertical/horizontal spacing.
This PR simply removes the duplicated offset compensation from the touchInputHandler since it's already handled by the generic zoom controller.

### DIFF
--- a/src/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/gui/inputdevices/TouchInputHandler.cpp
@@ -107,15 +107,7 @@ void TouchInputHandler::scrollMotion(InputEvent const& event) {
 }
 
 void TouchInputHandler::zoomStart() {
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator (approved):

personalizedrefrigerator:
Thank you for working on this! Merging in 48 hours unless there are objections!

PR #3390, opened 2021-09-11
Reworked PageRange to use C++17 STL.

createyourpersonalaccount:
Summary of changes:

* The function `PageRange::parse` was modified to use STL;
  - Fixes a potential bug were ranges larger than `pageCount`
    could be stored. (It is unclear if this bug could occur)
  - Eliminates a potential memory leak by avoiding `new`.

* Added documentation for `PageRange::parse`.

* `PageRangeEntry` was simplified to a small struct.
  - Its element have type `size_t` now. This matches other
    parts of the source code where pages or page counts are
    mentioned, and avoids casts between `int` and `size_t`.

* Reworked other parts of the source code to use `PageRange` with
  the new implementation.


createyourpersonalaccount:
Looks like the `int`/`size_t` conversions weren't thoroughly checked, and it fails on macOS. I'm not sure why there's a failure to convert from `int` to `size_t`, but I'll take a look a little later...

src/pdf/base/XojCairoPdfExport.cpp
```diff
@@ -143,7 +144,8 @@ void XojCairoPdfExport::exportPageLayers(size_t page) {
     for (const auto& layer: *p->getLayers()) layer->setVisible(initialVisibility[layer]);
 }
 
-auto XojCairoPdfExport::createPdf(fs::path const& file, PageRangeVector& range, bool progressiveMode) -> bool {
+auto XojCairoPdfExport::createPdf(fs::path const& file, const std::vector<PageRangeEntry>& range, bool progressiveMode)
+        -> bool {
```
flomnes, line 148: Why not declare this function as
```cpp
bool XojCairoPdfExport::createPdf(/*...*/)  {/*...*/}
```
?

I find this syntax a bit odd, but maybe it's just me.

src/pdf/base/XojCairoPdfExport.cpp
```diff
@@ -153,20 +155,18 @@ auto XojCairoPdfExport::createPdf(fs::path const& file, PageRangeVector& range,
...
-    int c = 0;
-    for (PageRangeEntry* e: range) {
-        for (int i = e->getFirst(); i <= e->getLast(); i++) {
-            if (i < 0 || i >= static_cast<int>(doc->getPageCount())) {
-                continue;
+    size_t c = 0;
+    for (auto& e: range) {
+        for (size_t i = e.first; i <= e.last; i++) {
+            if (i >= doc->getPageCount()) {
+                break;
```
flomnes, line 169: Change of behavior at line 169. This makes this PR more than a refactor.

src/util/PageRange.cpp
```diff
@@ -1,76 +1,67 @@
...
+ * The function silently corrects input that is outside the allowed range 1 - pageCount.
+ * For example, the input
+ *
+ *     1-3, 6, 4-, -5, 6-5, 99-, -
+ *
+ * results in a vector containing the following tuples, given `pageCount = 10`.
+ *
+ *     {0, 2}, {5, 5}, {3, 9}, {0, 4}, {5, 5}, {9, 9}, {0,9}
+ *
+ */
```
flomnes, line 22: I would add some unit testing to make sure that `parse` does exactly this.

createyourpersonalaccount:
Looks like the build fails on Windows because `poppler.h` is missing. I don't understand this error because I don't think I have modified anything related to Poppler. Does anyone know what happened?

robinp7720:
> Looks like the build fails on Windows because `poppler.h` is missing. I don't understand this error because I don't think I have modified anything related to Poppler. Does anyone know what happened?

I don't believe that it's anything to do with your code changes. The Windows CI build seems to be slightly broken right now.

src/control/XournalMain.cpp
```diff
@@ -1,6 +1,7 @@
 #include "XournalMain.h"
 
 #include <algorithm>
+#include <cstddef>
```
LittleHuba, line 4: Are you sure you need this header?

createyourpersonalaccount:
I'm trying to reply to some of the reviewer comments but my comments have been in 'Pending' status for a while now, is it my mistake?

LittleHuba:
> I'm trying to reply to some of the reviewer comments but my comments have been in 'Pending' status for a while now, is it my mistake?

This means you started a review. You can submit the comments in the `Files changed` tab with the `Submit review` button somewhere at the top right of the page. Just select `Comment` there.

src/pdf/base/XojCairoPdfExport.cpp
LittleHuba: There are voices in the developer community that see this style as the future. I'm not quite sure myself if it really will be but we'll see. For now we do not restrict it that heavily, though we might take that step in the future.

src/pdf/base/XojCairoPdfExport.cpp
LittleHuba: This is actually quite the improvement to the algorithm. `continue` would test all pages of the document against this condition even though it is clear that any following page will not be part of the selection if the previous wasn't. `break` instead aborts the inner loop after the first violation of the condition.

As our documents do generally not have negative page numbers, there is no negative difference in the behavior compared to the previous version.

test/util/PageRangeTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+class PageRangeTest: public CppUnit::TestFixture {
+    CPPUNIT_TEST_SUITE(PageRangeTest);
+
+    CPPUNIT_TEST(testParse);
+
+    CPPUNIT_TEST_SUITE_END();
+
+public:
+    void testParse() {
+        std::string ranges = "1-3, 6; 4-  : -5,, 6-5, 99-, -";
```
LittleHuba, line 39: You should add some more different ranges to this test. Try to think of corner cases. For example, what will happen if:

- One of the ranges is invalid but all others are valid
- No range at all is given
- Invalid characters are provided (letters, ...)
- ...

src/control/XournalMain.cpp
```diff
@@ -329,12 +326,9 @@ auto exportPdf(const char* input, const char* output, const char* range, ExportB
 
     if (range) {
         // Parse the range
-        PageRangeVector exportRange = PageRange::parse(range, doc->getPageCount());
+        auto exportRange = PageRange::parse(range, doc->getPageCount());
```
LittleHuba, line 329: Please make this variable `exportRanges`. As it is a vector of ranges and you use `auto` it should be straightforward, that these might be multiple ranges.

LittleHuba: Please also check all other places. I did see this multiple times throughout your code.

src/control/jobs/CustomExportJob.cpp
```diff
@@ -25,14 +25,9 @@ CustomExportJob::CustomExportJob(Control* control): BaseExportJob(control, _("Cu
...
 CustomExportJob::~CustomExportJob() {
-    for (PageRangeEntry* e: exportRange) {
-        delete e;
-    }
     exportRange.clear();
 
-    for (auto& filter: filters) {
-        delete filter.second;
-    }
+    for (auto& filter: filters) { delete filter.second; }
```
LittleHuba, line 30: Still an unnecessary style change.

src/gui/dialog/ExportDialog.h
```diff
@@ -26,7 +29,7 @@ class ExportDialog: public GladeGui {
     virtual void show(GtkWindow* parent);
     void initPages(int current, int count);
     bool isConfirmed() const;
-    PageRangeVector getRange();
+    std::vector<PageRangeEntry> getRange();
```
LittleHuba, line 32:
```suggestion
    std::vector<PageRangeEntry> getRanges();
```

src/util/PageRange.cpp
```diff
@@ -1,76 +1,75 @@
...
+            if (entry.last > pageCount)
+                entry.last = pageCount;
+            if (entry.last < entry.first)
+                entry.last = entry.first;
+            if (entry.first != 0) {
+                entry.first--;
+            }
+            if (entry.last != 0) {
+                entry.last--;
+            }
```
LittleHuba, line 70: Please document what all these changes do. Even though it might be obvious after reading the full code, comments should be added wherever you do not understand the meaning right away.

LittleHuba (changes requested):

src/util/PageRange.cpp
createyourpersonalaccount: Thanks, I'm working on it. I discovered a bug: `pageCount=0` would still push elements into the returned vector, but it should not. I've explained the situation in comments.

As soon as I can rework some repository issue I'm having right now, I'll push the changes.

createyourpersonalaccount: OK comments added.

src/control/XournalMain.cpp
createyourpersonalaccount: But that would be a variable names change; should that be a different PR?

test/util/PageRangeTest.cpp
createyourpersonalaccount: What do you mean by invalid range? I'll try these.

src/control/jobs/CustomExportJob.cpp
createyourpersonalaccount: Fixed.

src/gui/dialog/ExportDialog.h
createyourpersonalaccount: I'm going to wait for your response in the other similar review comment.

src/control/XournalMain.cpp
LittleHuba: No, you may of course add any change into this PR that is connected to the intent of your PR. So any renaming is okay, as long as it has something to do with your actual modifications to the behavior of the code. Here, the usage of auto instead of the actual type removes the implicit knowledge that this is a vector of ranges. Hence, renaming the variable is crucial for good coding as otherwise, one might think this is a single range.

Also with a PR like yours that is not really big, it is a good idea to fix such problems straight away instead of opening another PR. If your PR would be covering complex changes in lets say 20 files, I would agree that you should do this in another PR.

test/util/PageRangeTest.cpp
LittleHuba: Should you still be confused, `--1;5-6` would be an example. Further examples are:

- `;2-`
- `-1-`

src/control/XournalMain.cpp
createyourpersonalaccount: But then I will change `exportRange` (for example) to `exportRanges`, even though there will still be versions of it with `exportRange`. Wouldn't it be better to just replace all occurences? However, that means I'm changing files that I haven't changed in this PR.

LittleHuba: You will have to add all relevant files. But as you modified the return type of `PageRange::parse`, all files you must make additional changes to, should already be part of this PR.

We don't really care about minimum touched files in a PR. All we care about is that the intent of the PR matches with the changes. Because the intent of your PR is to update and fix the class `PageRange` any modification to the codebase that is in relation to this class is completely valid.

createyourpersonalaccount: OK I will update those occurrences to match the plural form of the name.

src/control/XournalMain.cpp
createyourpersonalaccount: I removed it, it was not needed.

src/control/XournalMain.cpp
createyourpersonalaccount: I am not very sure about this...

You mentioned something about `auto`.  What if I replace `auto` by `std::vector<PageRangeEntry>`, will it then be enough? (I.e. can I leave the variable names unchanged?)

test/util/PageRangeTest.cpp
createyourpersonalaccount: `-1-` is not bad input, because everything after a valid read is ignored (that was the old behavior too).

It should however be documented, and of course the unit test should show it.

createyourpersonalaccount: OK I was right after all. I documented the behavior though in the function documentation string, and added those testcases as well.

src/control/XournalMain.cpp
LittleHuba: Then there is still the issue with confusing variable names. You should make sure that within your code (and for the purpose of this PR within the whole class `PageRange`) there are no confusing variable names. Even better would be if you clean up all code that is using your code.

LittleHuba: You can do it in a separate commit, for now, so we can look at the impact on the codebase.

test/util/PageRangeTest.cpp
LittleHuba: Please modify the behavior to not accept any range if there is an error parsing it. The reason is, that a user does not purposefully input an invalid range string. Hence, if you have an error while parsing it, it will most likely be a typo by the user. At such a time we should throw an error instead of silently continuing. Otherwise, users might end up with a PDF that has the wrong pages without knowing about it.

LittleHuba: Please also confirm that the GUI notifies the user in the event of such a wrong input.

createyourpersonalaccount: > Please modify the behavior to not accept any range if there is an error parsing it. 

When does this happen? I'm not sure what you refer to.

LittleHuba: I'm talking about the general behavior of `PageRange::parse`. It should not accept faulty input. Instead, it should warn the user that the input could not be parsed.

createyourpersonalaccount: Ah, OK I understand.

When is this function used? I don't know how it is used, when does the user get to interact with it?

LittleHuba: If I'm not mistaken this is part of the printing dialog. Your IDE might help you out here. Most modern IDEs allow you to search for functions that use another function. Just right-click on `PageRange::parse` and take a look at the context menu.

createyourpersonalaccount: I think the place that you want it to warn the user is in `src/gui/dialog/ExportDialog.cpp`, where it is called by `ExportDialog::getRange()`. Do you know which user action invokes this function?

Would the warning have to use gtk or does xournal have a function to show a warning message?

LittleHuba: 1. Go to `File -> Export as...`
2. Select in the file dialog as file type `PNG`
3. Save the file

This will open the dialog you refer to. The best would be if you disable the `OK` button if the ranges are wrong and display a warning sign within the input field. You can use [gtk_entry_set_icon_from_pixbuf ()](https://developer-old.gnome.org/gtk3/stable/GtkEntry.html#gtk-entry-set-icon-from-pixbuf) for that.

 Another dialog on top seems to not be the best experience.

LittleHuba: For an appropriate icon you best speak with @luyatshimbalanga

LittleHuba: Or @luyatshimbalanga 

Both have designed icons for us previously.

createyourpersonalaccount: OK I've done some work on this. Try it out if you'd like and let me know what you think of it. I avoided using an icon, instead I used `gtk_style_context_*_class(context, "error");` to add/remove the error style in the text form. I also grayed out the OK button when appropriate. 

Since you probably last checked, I've also changed the parse function, now it emits exceptions in case of an issue, such as a pageCount equal to 0, or when the argument string is not a valid page range.

I still have a few minor things I'd like to change, i.e. I'd like to add a few more test cases, and some other minor things, but let me know what you think of it so far.

LittleHuba: Nice! Your work looks good from the first look at your code. Adding the context class is also a nice touch. I'll try it out live tomorrow.

createyourpersonalaccount:
@LittleHuba If there's updates let me know, did you try the right commit?

LittleHuba:
Sorry for the delay. I'm currently finishing my master thesis. I haven't forgotten about this and will get back to it as soon as I have some free time.

createyourpersonalaccount:
Oh nice! Okay then, I think that might be a priority :)

bhennion (changes requested):
This works almost perfectly here.

I have two requests:
1. Would it possible to add out of range page numbers to the list of forbidden strings in the dialog? The number 0 in particular seems to be treated as 1, and numbers above the number of pages are treated as the number of pages. Basically, I'd want "0,2-4" to throw, but also "6" if the document only has 4 pages.
2. Could you catch the error for invalid strings when exporting from CLI? Right now, a call to `xournalpp foo.xopp --create-pdf=output.pdf --export-range="--1"` triggers a signal 6 crash. Displaying an error message like `"--1" is not a valid page range list` would be preferable.
Changes for that would be in the file `control/XournalMain.cpp`, around the lines you already mdified.

src/gui/dialog/ExportDialog.cpp
```diff
@@ -84,22 +126,21 @@ auto ExportDialog::getBackgroundType() -> ExportBackgroundType {
...
+        std::vector<PageRangeEntry> range;
+        range.push_back(PageRangeEntry{this->currentPage - 1, this->currentPage - 1});
         return range;
     }
 
-
-    PageRangeVector range;
-    range.push_back(new PageRangeEntry(0, this->pageCount - 1));
+    std::vector<PageRangeEntry> range;
+    range.push_back(PageRangeEntry{0, this->pageCount - 1});
```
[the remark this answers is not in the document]
createyourpersonalaccount: What error code should I return to the shell?

bhennion:
> What error code should I return to the shell?

@createyourpersonalaccount Sorry I forgot to answer this! A simple -1 would do, I think.

src/util/PageRange.cpp
```diff
@@ -1,76 +1,110 @@
...
+     *   3) -m,  parsed by leftOpenRange,
+     *   4) n-m, parsed by pageRange,
+     *   5) -,   parsed by bothOpenRange,
+     *   6) For everything else, an exception is thrown.
+     */
+    const std::regex singlePage("\\s*(\\d+)\\s*");
+    const std::regex rightOpenRange("\\s*(\\d+)\\s*-\\s*");
+    const std::regex leftOpenRange("\\s*-\\s*(\\d+)\\s*");
+    const std::regex pageRange("\\s*(\\d+)\\s*-\\s*(\\d+)\\s*");
+    const std::regex bothOpenRange("\\s*-\\s*");
```
Technius, line 62: I think this is unnecessarily complicated. Why not a regex with three capture groups?
```
\\s*(\\d+)?\\s*(-)?\\s*(\\d+)?\\s*
```

src/util/PageRange.cpp
```diff
@@ -1,76 +1,110 @@
...
+ *     parse("1, 2-, -3, 4-5, -, 0-42", 10)
+ *     ===>
+ *     {0, 0}, {1, 9}, {0, 2}, {3, 4}, {0, 9}, {0, 9}
+ *
+ * Note that the page numbers are parsed in the format 1 - pageCount
+ * (permissive) and the return values are in the range 0 - (pageCount-1).
+ *
+ * @param s The string containing page ranges to parse.
+ * @param pageCount The largest page that is possible to refer to.
+ * @exception std::logic_error If `pageCount == 0`.
```
Technius, line 33: Error handling can be performed better by breaking this function into two smaller, private functions:
1. `parse`: Parse the input string into a page range data structure, returning an error message if failing.
2. `validate`: Validate that the page range data structure is valid (e.g., check start <= end, clamp to max page num).

Then a public function should call `parse` and then `validate`, which allows the error handling logic to be cleaner. 

This design also enables the constructor of `PageRange` to be private, so that  `PageRange` can only be constructed through a builder function that validates the provided range.

Technius (changes requested):

src/util/PageRange.cpp
createyourpersonalaccount: I would argue that the code documents itself. Originally it was like you suggested, I opted for the other option because the resulting code further below was cleaner.

src/util/PageRange.cpp
```diff
@@ -1,76 +1,110 @@
...
+ * @exception std::invalid_argument If the input doesn't match any acceptable
```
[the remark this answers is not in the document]
createyourpersonalaccount: In what way?

src/util/PageRange.cpp
```diff
@@ -1,76 +1,110 @@
...
+ * page ranges.
+ * @return A vector containing the page ranges.
+ */
+std::vector<PageRangeEntry> PageRange::parse(const std::string& s, size_t pageCount) {
```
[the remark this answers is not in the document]
createyourpersonalaccount: I think it's personal preference.

src/util/PageRange.cpp
createyourpersonalaccount: I agree.

createyourpersonalaccount:
I do not intend to work on this further, I appreciate all your comments.

bhennion:
> 1. Would it possible to add out of range page numbers to the list of forbidden strings in the dialog? The number 0 in particular seems to be treated as 1, and numbers above the number of pages are treated as the number of pages. Basically, I'd want "0,2-4" to throw, but also "6" if the document only has 4 pages.
> 2. Could you catch the error for invalid strings when exporting from CLI? Right now, a call to `xournalpp foo.xopp --create-pdf=output.pdf --export-range="--1"` triggers a signal 6 crash. Displaying an error message like `"--1" is not a valid page range list` would be preferable.
>    Changes for that would be in the file `control/XournalMain.cpp`, around the lines you already mdified.

If I'm not mistaken, those are still open issues, right?

Febbe:
@bhennion you already implemented this partially. Would you mind to either reject or rebase this PR?

bhennion:
> @bhennion you already implemented this partially. Would you mind to either reject or rebase this PR?

I think you meant @rolandlo, but I'm on it anyway. I'll push a rebased version by tomorrow.

bhennion:
Alright, I rebased on master. I tested it briefly (there is also a CI test).

PR #3392, opened 2021-09-12
(Minor) Close dangling file handlers.

createyourpersonalaccount:
In certain cases where an error occurs, the
file handlers obtained from `zip_fopen` were left
open. We close them. There is also an instance
of `zip_open` that we `zip_close`.


Technius (approved):

LittleHuba:
Part of next minor release.

Issue #729, opened 2019-01-11
Rich text support


PellelNitram:
I had the following idea which would make Xournal++ even more competitive to other (commercial) software.

The idea is rich text support. While I am not all sure if that's the right wording, here is what I thought of: When typing text, it would be amazing to have lists and word-wise formatting (such _as_ **this**). I call it "rich text support" because I know that RST or Markdown introduce a simple syntax that can be rendered to look like formatted text and lists upon rendering.

Neither do I have an idea how much work it is to implement it nor do I know what you think of it :-).

What I do know is that it would make Xournal++ even more competitive and that I would enjoy the feature. The reason for the latter is that I, amongst others, take study notes with Xournal++ and end up using "x)" or "*" as bullet points because I use lists a lot.

andreasb242:
@PellelNitram I fully agree with you.
And when I started Xournal++, I even thought I'll add Rich Text Support.

But it will be incompatible with .xoj, at that time this was the killer, I woulden't brake the support. But now we have .xopp, which is anyway incompatible, we have a .xoj export, which removes all stuff which is not compatible, so this will not be an issue anymore.

And because I would add RTF, I uses from beginning on Pango. So technical Xournal++ already is ready for RTF, if you select Text, we set the background Color, this is not a selection!

So first:
Load / Store needs to be implemented, and the .xoj Export, needs to be implemented.
And there needs a good UI Solution, a popup-toolbar or something like this, it does not fit the current customize toolbar.
But therefore it's much less work, than it may look like.

But don't expect it tomorrow, it's still complicated.... And it's definitive a hard task, which takes time....

PellelNitram:
Oh, that's amazing news to hear!

I will keep my eyes open and once I find a task that I can solve, I will create another PR :-).

PellelNitram:
Oh, that's amazing news to hear!

I will keep my eyes open and once I find a task that I can solve, I will create another PR :-).

andreasb242:
@PellelNitram If you need help you can ask in the issues, we usually respond.

But this task is probably a really bad start ;-) Definitive not a beginner task.

nicolae-stroncea:
[...]

Issue #1490, opened 2019-09-23
Why is the user input wrapped inside the variable \preview in Latex in LatexController.cpp


nicolae-stroncea:
I'm currently working on adding support for markdown through a latex template, but am hitting roadblocks because it seems that all of the user input has to  wrapped inside a variable \preview.

To use markdown inside latex, text has to be formatted like this:
```
\begin{markdown}
text here
\end{markdown}
```
I haven't been able to put the variable \preview inside the markdown environment since it is defined outside of it so the environment doesn't have access to it and just compiles with the text "\preview" instead of the actual value of the var.
```
\begin{markdown}
\preview
\end{markdown}
```
I'd like to know what the purpose of \preview is and if there's ways around it to achieve what I want

LittleHuba:
@technius you are the pro in this area 😉

nicolae-stroncea:
So I assume that's why there's a $x^2$ in the beginning.

I've got some follow-up questions on that:
* what's the relationship between blank formulas and the error checking that is going on? I assume that if there is an error, pdf would  be blank and Poppler wouldn't be able to load it due to size. Is it the same mechanism?
* If they are different things, is setting the layout to a certain height(pheight) used to solve the error checking problem or the blank formulas problem?

Technius:
If there is a blank formula, the error checking code in the template will raise an error, so `pdflatex` will exit with status 1. If `pdflatex` exits with status 1, then we assume that the formula is invalid and display the blank/invalid formula warning message; we do not attempt to load the PDF (in fact, we will delete it afterwards to prevent more errors).

Without the error checking code, `pdflatex` will exit with status 0 and generate a blank PDF when there is a blank formula, thereby crashing Poppler.

Sidenote: the `$x^2$` is purely cosmetic; the default formula could actually be set to anything.

Technius:
For now, I'm closing this issue. Feel free to reopen/reply if you have any additional questions.

PR #3408, opened 2021-09-21
Working version of markdown

nicolae-stroncea:
This PR introduces Markdown formatting through LaTeX. More customized text editing has been a feature that was long requested: https://github.com/xournalpp/xournalpp/issues/729, https://github.com/xournalpp/xournalpp/issues/1092, and makes Xournal++ text editing significantly more powerful.

To enable it, a user just has to change the template to `markdown_template.tex`

Here is what is now possible:

image

Here is the text I wrote in the Latex box:
```markdown

An h1 header
============


Paragraphs are separated by a blank line. 

2nd paragraph. *Italic*, **bold**, and `monospace`. Itemized lists look like:

* this one
* that one
* the other one

> Block quotes are
> written like so.
>
> They can span multiple paragraphs,
> if you like.

Let's use a horizontal line to separate the text:

***

An h2 header
------------

Here's a numbered list

1. first item
2. second item
3. third item

Here's \textcolor{red}{red} text and \textcolor{green}{green} text. Here's some \colorbox{yellow}{highlighted} text.

Here's a code sample you can do by indenting 4 spaces or using 3 backticks or 3 strikethrough symbols:

~~~python

for i in 1 .. 10:
  print("Hello World")
~~~


And here's a **table**:  \newline

| Right | Left |
|:------:|:-----:|
|  12   |  12  |
| 123   |  123 |
|   1   |    1 | 

```

Summarizing introduced possibilities:
* H1 and H2 headings
* Ordered and Unordered Lists
* In-line bold, italic, and monospace
* Colouring of specific text instead of everything in the textbox. You can use `xcolor` commands such as `\color{green}` to color everything green under the command, `\textcolor{red}{Hello}` to color just the text contained inside brackets
* Tables
* Highlighted code
* Hyperlinks and footnotes

**all within the same textbox and by using markdown syntax**

------------

**Some Notes**:

* Testing the configuration generates an error, despite the code working correctly. The error appears to be caused by the markdown template I wrote but I haven't been able to figure out why. 
image
* Adding `#` anywhere in the markdown box invalidates the whole LaTex, so we should warn users not to use it. Looks like a [known issue](https://www.overleaf.com/learn/latex/Articles/How_to_write_in_Markdown_on_Overleaf). There may be a couple more special characters that the markdown environment doesn't accept: `^,&,_``. Instead of using `#` for headers, use `===` for level 1 and `---` for level 2 headings.
* I've added a starting string that works for both the LaTex template and the markdown template, `3x` instead of `x^2` since the `^` is not accepted by the markdown environment.
* It would be really useful to have each template specify their own starting string (within the TEX template as a first line comment maybe), but I'm not sure how difficult that would be to implement. The markdown template will not render the first line properly(no clue why) unless you start with a blank line, so a better starting string for the markdown template would be `\n3x`. This won't work for the LaTeX template, so users will have to manually add a new line on first row if they want first line to be rendered correctly. Another reason this will be useful is we can create a template for tables, another one for writing code, which would be mostly identical to the markdown template but have different starting strings.
* I've added the option to colour the markdown text based on the current Text color, same way as the Tex tool, but I purposefully haven't enabled it (lines 34-35 in `markdown_template.tex` are commented out) yet.  It seems better to have the text be black by default and then allow users to colour it whichever way they want through the xcolor commands.

### DIFF
--- a/src/control/settings/LatexSettings.h
+++ b/src/control/settings/LatexSettings.h
@@ -20,8 +20,8 @@ public:
     bool autoCheckDependencies{true};
     fs::path globalTemplatePath{};
 #ifdef __APPLE__
-    std::string genCmd{"/Library/TeX/texbin/pdflatex -halt-on-error -interaction=nonstopmode '{}'"};
+    std::string genCmd{"/Library/TeX/texbin/pdflatex -halt-on-error --shell-escape -interaction=nonstopmode '{}'"};
 #else
-    std::string genCmd{"pdflatex -halt-on-error -interaction=nonstopmode '{}'"};
+    std::string genCmd{"pdflatex -halt-on-error --shell-escape -interaction=nonstopmode '{}'"};
 #endif
 };
--- a/resources/markdown_template.tex
+++ b/resources/markdown_template.tex
@@ -1,0 +1,48 @@
+% This template uses the scontents package, which is only available on relatively recent TeX distributions. In case it is not available on your system, use the legacy_template.tex
+\documentclass[varwidth=true, crop, border=5pt]{standalone}
+
+% Packages
+% \usepackage{amsmath}
+% \usepackage{amssymb}
+\usepackage{xcolor}
+
+\usepackage[fencedCode,definitionLists,pipeTables,tableCaptions]{markdown}
+% code syntax highlighting
+\usepackage{minted}
+
+
+% for storing in memory verbatim content to be reused later
+\usepackage{scontents}
+
+% Blank formula checking
+\usepackage{ifthen}
+\newlength{\pheight}
+
+% User input
+\begin{scontents}[store-env=preview]
+    %%XPP_TOOL_INPUT%%
+\end{scontents}
+
+% Command to set colour globally
+\makeatletter
+\newcommand{\globalcolor}[1]{%
+  \color{#1}\global\let\default@color\current@color
+}
+\makeatother
+
+% Set the color globally
+% \definecolor{xpp_font_color}{HTML}{%%XPP_TEXT_COLOR%%}
+% \AtBeginDocument{\globalcolor{xpp_font_color}}
+
+\begin{document}
+
+
+  % Check if the formula is empty
+  \settoheight{\pheight}{\getstored[1]{preview}}%
+  \ifthenelse{\pheight=0}{\GenericError{}{xournalpp:blankformula}{}{}}
+
+  % Render the user input
+  \begin{markdown}
+    %%XPP_TOOL_INPUT%%
+  \end{markdown}
+\end{document}


LittleHuba:
Nice! That is a big step in the right direction. If we can manage to hide this behind a graphical editor in the future we would have great text editing!

Thanks for the work so far. I'll check out the code and give you some ideas on your problems.

Witiko:
> 1. Disable latex in the markdown template.

Assuming you are talking about [the `hybrid` option of the Markdown package](https://witiko.github.io/markdown/#option-hybrid), then please note that this does not close the security issue. You can end the markdown environment by writing `\end{markdown}` and then execute arbitrary TeX code regardless of whether or not the support for TeX in markdown text is enabled. This is not just a security issue; it also means that you can not write `\end{markdown}` in your input. At all.

> @Witiko suggested as a fix saving the user input in a text file and then doing `\markdownInput{input.md}` in the markdown template

This closes the security issue and also enables other things such as writing two or more spaces at the end of a line to produce a hard line break, which do not work in `\begin{markdown} ... \end{markdown}` due to limitations of TeX. For more information, see [Section 2.3.1 of the Markdown documentation](https://mirrors.ctan.org/macros/generic/markdown/markdown.pdf#page=35).

> Also, how much would this affect performance, given that we would have to keep rewriting the `input.md`.

From the standpoint of performance, they are equivalent. With pdfTeX, `\begin{markdown} ... \end{markdown}` will buffer the text into a file and then execute the equivalent of `\markdownInput`.

Witiko:
As the developer of the [witiko/markdown](https://github.com/witiko/markdown) LaTeX package, I am always happy to help with developing FOSS. Please, feel free to ping me any time with your questions and feature requests.

Technius:
This is a really cool idea!

> How serious is this for Xournal++'s threat model?

Security has unfortunately been neglected in Xournal++ in general. With respect to the LaTeX tool, the formula is substituted into the template without any escaping. This means that an attacker can easily create a malicious `.xopp` file (note that rendered LaTeX is cached in the file; an attack can only occur if the user attempts to edit a malicious formula). And given the complexity and power of LaTeX, I would not expect any workaround to be very useful or effective.

> It would be really useful to have each template specify their own starting string

I think this would be a good idea even outside of the context of using markdown. We can have templates for different applications&mdash;mathematical equations, tikz drawings, chemical formulas, code snippets, etc.&mdash;all of which would be better with their own custom starting strings.

In the long term, I wonder if it would be better to generalize the "LaTeX tool" into a "Markup tool". There's nothing LaTeX-specific about how the LaTeX tool actually works: all we need is a way to substitute user input into a template and then invoke some command that generates a PDF. For example, we could use `pdflatex` for latex, `pandoc` for markdown, and `pygments` for code snippets.

LittleHuba:
> And given the complexity and power of LaTeX, I would not expect any workaround to be very useful or effective.

Depending on the environment we could actually prevent escaping it. But all in all the more such templates we introduce, the bigger our attack surface gets. So maybe we should consider securing the current LaTeX template first before adding additional templates?

introt:
> In the long term, I wonder if it would be better to generalize the "LaTeX tool" into a "Markup tool". There's nothing LaTeX-specific about how the LaTeX tool actually works: all we need is a way to substitute user input into a template and then invoke some command that generates a PDF. For example, we could use `pdflatex` for latex, `pandoc` for markdown, and `pygments` for code snippets.

Over at [Zim](https://github.com/zim-desktop-wiki/zim-desktop-wiki) we have something similar called ImageGenerator plugins. These present to the user as a menu option as well as a dialog to edit and test the generated images' source. The source file gets attached to the page along with a png render. The latter is displayed in the wiki page. These plugins are really easy to code[1] and make it really quick and simple to create simple diagrams and make small fixes to big ones[2].

image
_Some equations created with the Equation Editor plugin, as seen in the wiki page as well as its attachments. The source LaTeX was copied from Wikipedia_

I'd really love a Lilypond "Markup tool" -- you could mark the score with Xournal++ while practicing and later edit the score without having to change the background document -- though for that particular use-case I'd like to open the source file in the editor of my choise and not some simple dialog, which'd mean the source couldn't (as easily) be embedded in the document.

Anyhow, nice wok nicolae-stroncea! Been longing for a quick way to input rich text since Xournal and were wondering when it would get picked up after reading #3397.

Please mark this comment as off-topic if you'd like to move the discussion re: generalizing LaTeX tool into "PDF-generators" elsewhere.

---

0. Current selection included out-of-the-box ranges from LaTeX and GNU R Plots to Ditaa diagrams and Lilypond musical scores. However, we ship none of those - system commands are detected during runtime. For a complete list, see the *Editor plugins in the manual: https://zim-wiki.org/manual/Plugins.html (please ignore the broken <a> tags, we're due for a website update along with a  `.1` release soon)
1. See eg. the "Ditaa Editor"'s source, it defines a total of 4 functions with like 15 lines of code _combined_: https://github.com/zim-desktop-wiki/zim-desktop-wiki/blob/develop/zim/plugins/ditaaeditor.py
2. Just right-click the object and select "Edit" to open the editor dialog - the image gets re-rendered on save.

Witiko:
> * Adding `#` anywhere in the markdown box invalidates the whole LaTex, so we should warn users not to use it. Looks like a [known issue](https://www.overleaf.com/learn/latex/Articles/How_to_write_in_Markdown_on_Overleaf). There may be a couple more special characters that the markdown environment doesn't accept: `^,&,_`. Instead of using `#` for headers, use `===` for level 1 and `---` for level 2 headings.
> * I've added a starting string that works for both the LaTex template and the markdown template, `3x` instead of `x^2` since the `^` is not accepted by the markdown environment.

The Overleaf article is outdated. The character `#` can always be used for headings and it can also be used in the text unless you enable the `hybrid` option, which enables you to mix LaTeX and Markdown markup and then you would need to write `\\#`.

LittleHuba:
I see this situation kind of like the Macro-feature of the MS Office Suite. Maybe we should load untrusted documents in a read-only fashion, where such unsafe content is shown with static images from within the file. If the user trusts the document, he can enable editing of the document. From that point on, it is the users' responsibility to not open any such "macro" element and save it with code that is malicious.

LittleHuba:
> Over at Zim we have something similar called ImageGenerator plugins. These present to the user as a menu option as well as a dialog to edit and test the generated images' source, similar to Xournal++'s "Insert LaTeX" one. The source file gets attached to the page along with a png render. The latter is displayed in the wiki page. These plugins are really easy to code[1] and make it really quick and simple to create simple diagrams and make small fixes to big ones[2].

This is where I wanted to go with the whole concept of Xournal++. Making any element just the result of a plugin. That would open the documents of Xournal++ to many cool features.

introt:
> This is where I wanted to go with the whole concept of Xournal++. Making any element just the result of a plugin. That would open the documents of Xournal++ to many cool features.

You might want to also take a deeper look at how Zim is designed and structured for further inspiration: most stuff is implemented as plugins, including basic stuff like the Page Index. This way the core doesn't get bloaty and unmaintainable over time, and the users have fine grained control over the UX (something which the poster in eg. #3262 would appreciate). It's kinda beautiful to me, really, as both an user and a (minor) developer.

Shipping features as plugins, with sane defaults and the most complicated ones disabled out-of-the-box, avoids giving the user an information overload or getting them stuck in decision paralysis. Not that I don't enjoy using Xournal++, but opening it up for the first time after using Xournal for years did freak me out a little :)

nicolae-stroncea:
Submitted a commit that fixes the new line issue, but `#` still doesn't work, even without the hybrid option.

@Witiko did `#` always work, or was it enabled after a certain version?

I'm wondering if maybe I'm running an outdated version either of the latex package, or one of its dependencies maybe.

Witiko:
It always worked. It would be useful if you could attach the standard error output of pdfLaTeX.

nicolae-stroncea:
Looking more into the issue with writing a `#`, if I include some text with a heading directly into `markdown_template.tex` the heading appears correctly. Example:
```
\begin{markdown}
# Hello world
%%XPP_TOOL_INPUT%%
\end{markdown}
```
But if I include the same text `# Hello world` directly in the LaTeX textbox, I get "The formula is empty when rendered or invalid"

nicolae-stroncea:
Unfortunately I can't easily grab a hold of the error. It's not logged to the output if I run `xournalpp` from CLI, and I'm not sure if it's logged somewhere else. @rolandlo are the errors generated when running `pdflatex` saved somewhere?

Witiko:
Can you get a hold of the produced .tex file?

nicolae-stroncea:
Here's the log files when I write: "# 3x"

[log_files.zip](https://github.com/xournalpp/xournalpp/files/7220839/log_files.zip)

The error is: "You can't use `macro parameter character #' in restricted horizontal mode"

Technius:
@nicolae-stroncea The log should be saved to `/tmp/xournalpp-${PID}/tex/tex.log` (you'll have to find the right PID for your xournalpp process).

Witiko:
> Here's the log files when I write: "# 3x"
> 
> [log_files.zip](https://github.com/xournalpp/xournalpp/files/7220839/log_files.zip)
> 
> The error is: "You can't use `macro parameter character #' in restricted horizontal mode"

Here is the unpacked `tex.tex` file. You will notice that your issue is with the `scontents` environment, not the Markdown package>

``` tex
% This template uses the scontents package, which is only available on relatively recent TeX distributions. In case it is not available on your system, use the legacy_template.tex
\documentclass[varwidth=true, crop, border=5pt]{standalone}

% Packages
% \usepackage{amsmath}
% \usepackage{amssymb}
\usepackage{xcolor}

\usepackage[fencedCode,definitionLists,pipeTables,tableCaptions]{markdown}
% code syntax highlighting
\usepackage{minted}


% for storing in memory verbatim content to be reused later
\usepackage{scontents}

% Blank formula checking
\usepackage{ifthen}
\newlength{\pheight}

% User input
\begin{scontents}[store-env=preview]
# 3x
\end{scontents}

% Set the color globally
% \definecolor{xpp_font_color}{HTML}{3333cc}
% \AtBeginDocument{\globalcolor{xpp_font_color}}

\begin{document}


% Check if the formula is empty

% Render the user input
\begin{markdown}
# 3x
\end{markdown}
\end{document}
```

nicolae-stroncea:
I see, thank you for the help debugging! The `scontents` package is something I was long wondering about (https://github.com/xournalpp/xournalpp/issues/1490).  @Technius at the time said we're using it as a hack essentially to avoid an error from `pdflatex` on an empty string. Here's how I think we can avoid this issue altogether:

We can modify `triggerImageUpdate` to disable the ok button on an empty string, and to write a message to the user that "the formula must not be empty".

```cpp
void LatexController::triggerImageUpdate(const string& texString) {
    if (this->isUpdating()) {
        return;
    }
    if (texString.empty()) {
        GtkWidget* okButton = this->dlg.get("texokbutton");
        gtk_widget_set_sensitive(okButton, false);
        GtkLabel* errorLabel = GTK_LABEL(this->dlg.get("texErrorLabel"));
        gtk_label_set_text(errorLabel, N_("The formula must not be empty."));
        return;
    }
```

Here's how it would look:
image

The last generated image is still embedded, but I think the message makes it clear enough to the user what is going on, and why they can't submit ok. With this we could get rid of scontents entirely and all the code related to it(setting the height, etc).

@Technius what are your thoughts on this?

nicolae-stroncea:
Just added a commit which stores the markdown in a file instead of in plaintext, as @Witiko suggested. 

> Depending on the environment we could actually prevent escaping it. But all in all the more such templates we introduce, the bigger our attack surface gets. So maybe we should consider securing the current LaTeX template first before adding additional templates?

@LittleHuba the markdown template now both has hybrid mode disabled(i.e LaTeX commands disabled) and is written to a file so escaping should not be possible anymore.

src/control/settings/LatexSettings.h
```diff
@@ -20,8 +20,8 @@ class LatexSettings {
     bool autoCheckDependencies{true};
     fs::path globalTemplatePath{};
 #ifdef __APPLE__
-    std::string genCmd{"/Library/TeX/texbin/pdflatex -halt-on-error -interaction=nonstopmode '{}'"};
+    std::string genCmd{"/Library/TeX/texbin/pdflatex -halt-on-error --shell-escape -interaction=nonstopmode '{}'"};
 #else
-    std::string genCmd{"pdflatex -halt-on-error -interaction=nonstopmode '{}'"};
+    std::string genCmd{"pdflatex -halt-on-error --shell-escape -interaction=nonstopmode '{}'"};
```
LittleHuba, line 25: This additional flag should only be enabled for the Markdown template. This worsens the attack surface of the LaTeX template tremendously.

LittleHuba (changes requested):
It would make sense to separate this into another tool. Mixing Markdown and LaTeX tools does not really make any sense in my opinion.

Witiko:
Please note that the `-shell-escape` flag must only be enabled for
pdflatex. With lualatex, it can be dropped completely.

src/control/settings/LatexSettings.h
Witiko: Please note that the `-shell-escape` flag must only be enabled for pdflatex. With lualatex, it can be dropped completely.

LittleHuba: We'd have to look up if lualatex is available on all platforms we ship to. Also I don't really think it is an issue if we use `--shell-escape` for markdown, as it is quite secure with the Markdown in its separate file. My worries are solely with using this flag for LaTeX, as it allows literally any command being issued in the shell.

Witiko: The LuaTeX binary is [available since at least TeX Live 2016](http://www.luatex.org/). It predates the Markdown package, i.e. in a full TeX Live installation, LuaTeX should be available whenever the Markdown package is available.

LittleHuba: @xournalpp/core what is your opinion on switching to LuaTeX?

Technius: I don't see any problem changing the default command as long as it produces non-zero error codes on build failures.

rolandlo: luatex and lualatex are also included in TinyTeX (which we currently use for the snap package).

RJDan:
> To enable it, a user just has to change the template to `markdown_template.tex`

Hello...
Where do we find this 'markdown_template.tex'?

linnabraham:
> > To enable it, a user just has to change the template to `markdown_template.tex`
> 
> Hello... Where do we find this 'markdown_template.tex'?

I got a version by checking out the Files Changed tab in the top of this pull request page. But it still doesnt work for me when I try to test the configuration. Gives me an error about missing $ inserted

archisman-panigrahi:
It has been three years, and issues like https://github.com/xournalpp/xournalpp/issues/729 are being closed in favor of this PR. Any updates?

rolandlo:
> It has been three years, and issues like #729 are being closed in favor of this PR. Any updates?

Looks like this PR was forgotten. I opened #729 once more.

Generally speaking I agree with @Technius and @LittleHuba that we need to generalize the LaTeX tool, so that different PDF generators (latex, pandoc, typst, pygments, ...) can be used simultaneously.

MattiaGrigoli:
So is this thing implemented or not?
Because I have an issue with the text: when writing long lines it doesn't break the line and go to the next line, instead it keeps writing over the page.

PR #3414, opened 2021-09-25
Create pipeline for releases via comments

LittleHuba:
This allows collaborators to create installers for all supported OSs in PR's by commenting `\AzurePipelines run create-installers`

This should help with PRs that target Windows or macOS where it is quite the task for somebody to set up a build environment. Hopefully, this will increase testers. I restricted the creation of such installers to collaborators, as not to tax our pipelines too much.


LittleHuba:
\Azure-Pipelines run create-installers

PR #3420, opened 2021-09-26
(Dialog was not freed after allocation.

createyourpersonalaccount:


LittleHuba:
Please use a smart pointer for `dlg` instead.

LittleHuba (approved):

src/control/jobs/CustomExportJob.cpp
```diff
@@ -70,7 +72,7 @@ auto CustomExportJob::showFilechooser() -> bool {
 
     Document* doc = control->getDocument();
     doc->lock();
-    auto* dlg = new ExportDialog(control->getGladeSearchPath());
+    auto dlg = std::make_unique<ExportDialog>(control->getGladeSearchPath());
```
bhennion, line 75: Is a pointer even necessary here? 
```suggestion
    ExportDialog dlg(control->getGladeSearchPath());
```
(with of course change each `dlg->foo` by a `dlg.foo`)

Technius (approved):
The only concern would be if the destructor of `ExportDialog` has side effects, but it doesn't. LGTM.

LittleHuba:
Can you please check the size of `ExportDialog` Depending on its size it might be a bad idea to store this on the stack.

createyourpersonalaccount:
> Can you please check the size of `ExportDialog`? Depending on its size it might be a bad idea to store this on the stack.

Good point, from a quick look: I see 3 ints in the class and it inherits GladeGui which has three pointers. I think it’s okay to have as a stack variable.

LittleHuba:
Merge in 24 hours if there are no objections.

PR #3444, opened 2021-10-08
Update Authors lists

rolandlo:
Here is the output of
```
diff AUTHORS AUTHORS_raw
```
[diff.txt](https://github.com/xournalpp/xournalpp/files/8009657/diff.txt)

The diff should show that I have only removed authors from the AUTHORS_raw file that are already in the developer team. One exception is @fabian-thomas who has used 2 different email adresses, where I only kept one.

@bhennion @personalizedrefrigerator 

**Edit**: Updated on Feb 6, 2022.


rolandlo:
Squashing into 1 commit and "rebase and merge"ing in 24 hours if no objections are raised.

Issue #2548, opened 2020-12-17
Collaboration mode


Clemens-Dautermann:
**This feature is not related to a problem**
I just think it would be awesome to have a collaborative mode to enable multiple peple to edit one document

**Describe the solution you'd like**
One could implement a collaborative mode. Kind of like in JetBrains CodeWithMe.

**Describe alternatives you've considered**
I have considered using online editors but they either are garbage or closed source or don't work with my tablet.

**Additional context**
This is propably a huge  project, really complicated and low priority. I would love to help though.

LittleHuba:
We are actually working on such a tool. It is much work and will therefore take some time.

Clemens-Dautermann:
I thought as much. Thank you very much!

sanette:
meanwhile, did you try Write from stylus-lab? Having experience with Write would be a good way to select what we need for xournalpp

Clemens-Dautermann:
No I did not, but I will, thanks!

sanette:
I think an "easy" way to implement collaboration is the following:

1. assume both authors are already using a file-syncing app like `dropbox` or `seafile`
2. the xournalpp session has an "owner", say person A; the other one is B
3. B has to detect that A is owner and xournalpp puts itself in `read only` mode for B (this can be done with a shared lock file)
4. A can modify the file and save it. B will see the modifications once the file is saved and sync-ed.
5. A has a button to yield ownership to B (or a variant: the button simply gives temporary write access to B). During this period, the file is `read only` for A.
6. and so on...

Of course, this is 'slow' sharing, A and B cannot both write on the document at the same time, and the modifications cannot be seen 'on the fly'. All of these would be very nice, of course. But the "slow" mode is already enough for all real-life situations I need, where each one can write in turn.

thibaultmol:
is this still being worked on?

sanette:
Well, not really; it currently "works". (as described above) but since it seems this wasn't the direction that people wanted to go, I didn't extend it further.

tobiasBora:
I've been using drawpile for collaborative open-source drawing app, and it works great but:
1. you need to host your own server as the public ones are often overwhelmed. While it is nice to have this option, it requires extra steps from users that are not trivial (open ports etc). Something based on WebRTC based on a TURN server seems like a simpler solution: users then just need to share a link with their favorite app/email/matrix room and don't need to have any running server.
[...]

PR #3450, opened 2021-10-11
Collaboration mode

sanette:
**First attempt at a "slow" collaboration mode...**

Following up the discussion here: https://github.com/xournalpp/xournalpp/issues/2548

This PR is mainly for building on various targets so that the feature can be tested.


sanette:
great, thanks!

PR #3460, opened 2021-10-13
Fix ToolType serialization bug.

createyourpersonalaccount:
The following serialization invariant was found to not
hold for ToolType:

    fromString(toString(x)) == x

The reason was a mismatch in serialization:

    toolTypeToString(TOOL_PLAY_OBJECT) == "PlayObject"
    toolTypeFromString("playObject") == TOOL_PLAY_OBJECT)

This probably had the effect of incorrectly serializing. We correct this
by changing the return value of  toolTypeToString to match the camelCase
style seen in its other return values.


LittleHuba:
Please rebase this on `release-1.1`. See https://github.com/xournalpp/xournalpp/blob/master/readme/Releases.md#branching-strategy for an explanation of our branching strategy.

PR #3465, opened 2021-10-13
Change Tool internals, increase static analysis.

createyourpersonalaccount:
This depends on #3464.


LittleHuba:
> You might want to consider using size_t instead of unsigned int in most cases, but I don't think that it really matters here.

What cases do you mean? If you speak of the enum then I kindly disagree. There is no requirement to increase the memory consumption for something like an enum flag field. He has at least 32 bits -> 32 enum values to set in an `unsigned int`. Way more than he requires.

PR #3480, opened 2021-10-16
Improve Stroke's display in StrokeView and StrokeHandler

bhennion:
This PR solves various stroke's display issues by refactoring the class `StrokeView`.
The solved issues are:

- Filled highlighter strokes were not filled when being drawn
- Various display issues of the fadeout supposed to happen when a stroke has no audio attached and the `play Audio` tool is selected.

Here is a video with both behaviours (none of the strokes have audio attached here...).

https://user-images.githubusercontent.com/17818041/137598650-8c54c0e1-1fee-4a6d-a0c0-96e7db227721.mp4

Obviously, there is an intersection with #2874 (but only in `StrokeView`). The two PRs could be merged. In particular, this solves the issue raised [in this post](https://github.com/xournalpp/xournalpp/pull/2874#pullrequestreview-751185320)

### DIFF
--- a/src/config-debug.h.in
+++ b/src/config-debug.h.in
@@ -42,3 +42,8 @@
  * Draw a border around all painted rects
  */
 #cmakedefine DEBUG_SHOW_PAINT_BOUNDS
+
+/**
+ * Draw the mask in StrokeView
+ */
+#cmakedefine DEBUG_SHOW_MASK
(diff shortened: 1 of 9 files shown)


bhennion:
Apparently there is a building issue in the Linux pipeline due to the type `uint8_t` being unknown. Not sure why this happens: there are ofc other instances of `uint8_t` in the code (see e.g. `DocumentView::applyColor`) and it compiles well on my laptop.

LittleHuba:
> `StrokeHandler` paints on a mask that is later blitted on the screen. Preivously this mask's size was the entire page's size * zoom. On high zoom levels, it meant having to allocate up to hundreds of MB (depending on the page size, ofc) whenever a stroke is started. This could be the source of some lags.

Wow, that is one heavy corpse in the code you eliminated there!

bhennion:
I'd like to merge this this weekend. If anyone objects to that, please say so!

Issue #1288, opened 2019-06-24
Better svg import


jensMF:
**Is your feature request related to a problem? Please describe.**
When I import .svg files I get ugly, pixelated and hardly readable results.

**Describe the solution you'd like**
I would like to have a nice vectorized graphic.

In the example, it is readable, but I head other examples where it was not.

<img width="1280" alt="Screenshot" src="https://user-images.githubusercontent.com/1620034/60021985-44320b00-9693-11e9-90d3-9c7422bd1318.png">

Technius:
The issue here is that the image tool loads images using GdkPixbuf, which rasterizes SVG files.

https://github.com/xournalpp/xournalpp/blob/90f583a851265c52abc69aa7fcc81fe632ca23d1/src/control/tools/ImageHandler.cpp#L49-L67

This can be fixed by loading SVG files using `cairo_svg_surface_create` and then use `image->setImage` on the surface. This change should be done in `ImageHandler::insertImage`.

Technius:
I forgot that cairo can't load SVGs by itself. The _real_ fix is to use rsvg to load the SVG and then rerender the pixbuf whenever the dimensions change. This requires more work and possibly needs an additional dependency.

LittleHuba:
This would actually be good to implement with #1344 as we need the same functionality there.

lukyluket:
How do I actually import svg?

I am not able to do it

Technius:
> How do I actually import svg?
> 
> I am not able to do it

Click somewhere with the Image Tool and choose an svg file.

lukyluket:
Thanks,
I wanted to point out that there is a new application tuhi that allows
downloading files from wacom slate tablets,
I think it would be fantastic to use those tablet for when offline from
computer and being able to import the notes in Xournalpp files,

I am suggesting to offer native Xournal export but maybe we could think of
importing the svg file they produce
https://github.com/tuhiproject/tuhi

LittleHuba:
Import of SVGs is easier than connecting to tuhi in some other way. If somebody is willing to connect tuhi to X++ in a more sofisticated way he is more than welcome though.

lukyluket:
I have tried that but it imports the full document as an image, and is not
editable.
In reality it should be a connection of strokes, and ideally it would be
nice to make it editable from inside X++

LittleHuba:
This ticket is for rescaling SVGs when the zoom of the view is adjusted. It is not about editable SVGs. 

[...]

PR #3503, opened 2021-10-24
Refactor view/DocumentView

bhennion:
The PR is an attempt to refactor view/DocumentView.

~~It is based on both #3480 and #3502, so the first commits come from there.
I'll rebase on master if and when anyone of those PR's get merged.~~ Edit: This is done now.

About the code now: I took care of all the easy things.
Things'll get more messy with the background and page painters. I don't exactly know how to proceed.
@LittleHuba @Febbe @rolandlo, I know you guys have been discussing the View refactor for a while now, so you may have a more mature idea on how to move forward from here.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -78,8 +78,6 @@ Control::Control(GApplication* gtkApp, GladeSearchpath* gladeSearchPath): gtkApp
 
     this->applyPreferredLanguage();
 
-    TextView::setDpi(settings->getDisplayDpi());
-
     this->pageTypes = new PageTypeHandler(gladeSearchPath);
     this->newPageType = std::make_unique<PageTypeMenu>(this->pageTypes, settings, true, true);
 
@@ -1970,8 +1968,6 @@ void Control::showSettings() {
 
     getWindow()->getXournal()->getHandRecognition()->reload();
     getWindow()->updateColorscheme();
-
-    TextView::setDpi(settings->getDisplayDpi());
 
     delete dlg;
 }
--- a/src/core/control/jobs/PreviewJob.cpp
+++ b/src/core/control/jobs/PreviewJob.cpp
@@ -7,6 +7,7 @@
 #include "gui/sidebar/previews/layer/SidebarPreviewLayerEntry.h"
 #include "model/Document.h"
 #include "view/DocumentView.h"
+#include "view/LayerView.h"
 #include "view/PdfView.h"
 
 PreviewJob::PreviewJob(SidebarPreviewBaseEntry* sidebar): sidebarPreview(sidebar) {}
@@ -96,6 +97,8 @@ void PreviewJob::drawPage() {
             break;
     }
 
+    auto context = xoj::view::Context::createDefault(cr2);
+
     switch (type) {
         case RENDER_TYPE_PAGE_PREVIEW:
             // render all layers
@@ -109,7 +112,8 @@ void PreviewJob::drawPage() {
                 view.drawBackground();
             } else {
                 Layer* drawLayer = (*page->getLayers())[layer];
-                view.drawLayer(cr2, drawLayer);
+                xoj::view::LayerView layerView(drawLayer);
+                layerView.draw(context);
             }
             view.finializeDrawing();
             break;
@@ -120,7 +124,8 @@ void PreviewJob::drawPage() {
             view.drawBackground();
             for (int i = 0; i <= layer; i++) {
                 Layer* drawLayer = (*page->getLayers())[i];
-                view.drawLayer(cr2, drawLayer);
+                xoj::view::LayerView layerView(drawLayer);
+                layerView.draw(context);
             }
             view.finializeDrawing();
             break;
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -13,7 +13,6 @@
 #include "gui/XournalView.h"
 #include "undo/InsertUndoAction.h"
 #include "undo/RecognizerUndoAction.h"
-#include "view/DocumentView.h"
 #include "view/StrokeView.h"
 
 #include "StrokeStabilizer.h"
@@ -48,10 +47,19 @@ void StrokeHandler::draw(cairo_t* cr) {
         auto context = xoj::view::Context::createColorBlind(crMask);
         strokeView->draw(context);
     }
-    DocumentView::applyColor(cr, stroke);
 
-    cairo_set_operator(
-            cr, stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER ? CAIRO_OPERATOR_MULTIPLY : CAIRO_OPERATOR_OVER);
+    // set the stroke color and blend mode
+    if (stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER) {
+        if (auto fill = stroke->getFill(); fill != -1) {
+            Util::cairo_set_source_rgbi(cr, stroke->getColor(), static_cast<double>(fill) / 255.0);
+        } else {
+            Util::cairo_set_source_rgbi(cr, stroke->getColor(), xoj::view::StrokeView::OPACITY_HIGHLIGHTER);
+        }
+        cairo_set_operator(cr, CAIRO_OPERATOR_MULTIPLY);
+    } else {
+        Util::cairo_set_source_rgbi(cr, stroke->getColor());
+        cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
+    }
 
     cairo_mask_surface(cr, surfMask, 0, 0);
 }
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -13,6 +13,8 @@
 #include "gui/XournalView.h"
 #include "undo/InsertUndoAction.h"
 #include "undo/RecognizerUndoAction.h"
+#include "view/DocumentView.h"
+#include "view/StrokeView.h"
 
 #include "StrokeStabilizer.h"
 #include "config-features.h"
@@ -43,8 +45,8 @@ void StrokeHandler::draw(cairo_t* cr) {
         cairo_set_operator(crMask, CAIRO_OPERATOR_CLEAR);
         cairo_paint(crMask);
 
-        cairo_set_operator(crMask, CAIRO_OPERATOR_SOURCE);
-        view.drawStroke(crMask, stroke, true);
+        auto context = xoj::view::Context::createColorBlind(crMask);
+        strokeView->draw(context);
     }
     DocumentView::applyColor(cr, stroke);
 
@@ -158,7 +160,9 @@ void StrokeHandler::drawSegmentTo(const Point& point) {
         lastSegment.addPoint(point);
         lastSegment.setWidth(width);
 
-        view.drawStroke(crMask, &lastSegment, true);
+        auto context = xoj::view::Context::createColorBlind(crMask);
+        xoj::view::StrokeView sView(&lastSegment);
+        sView.draw(context);
     }
 
     width = prevPoint.z != Point::NO_PRESSURE ? prevPoint.z : width;
@@ -326,6 +330,10 @@ void StrokeHandler::onButtonPressEvent(const PositionInputData& pos) {
 
         this->hasPressure = this->stroke->getToolType() == STROKE_TOOL_PEN && pos.pressure != Point::NO_PRESSURE;
         this->fullRedraw = this->stroke->getFill() != -1 || stroke->getLineStyle().hasDashes();
+
+        if (this->fullRedraw) {
+            strokeView.emplace(stroke);
+        }
 
         stabilizer->initialize(this, zoom, pos);
     }
--- a/src/core/view/ElementView.cpp
+++ b/src/core/view/ElementView.cpp
@@ -1,0 +1,29 @@
+#include "model/Element.h"
+#include "model/Image.h"
+#include "model/Stroke.h"
+#include "model/TexImage.h"
+#include "model/Text.h"
+
+#include "ImageView.h"
+#include "StrokeView.h"
+#include "TexImageView.h"
+#include "TextView.h"
+#include "View.h"
+
+using namespace xoj::view;
+
+auto ElementView::createFromElement(const Element* e) -> std::unique_ptr<ElementView> {
+    switch (e->getType()) {
+        case ELEMENT_STROKE:
+            return std::make_unique<StrokeView>(dynamic_cast<const Stroke*>(e));
+        case ELEMENT_TEXT:
+            return std::make_unique<TextView>(dynamic_cast<const Text*>(e));
+        case ELEMENT_IMAGE:
+            return std::make_unique<ImageView>(dynamic_cast<const Image*>(e));
+        case ELEMENT_TEXIMAGE:
+            return std::make_unique<TexImageView>(dynamic_cast<const TexImage*>(e));
+        default:
+            assert(false && "ElementView::getFromElement: Unknown element type!");
+            return nullptr;
+    }
+}
--- a/src/core/view/ImageView.h
+++ b/src/core/view/ImageView.h
@@ -1,0 +1,30 @@
+/*
+ * Xournal++
+ *
+ * Displays a Image Element
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+#include "View.h"
+
+class Image;
+
+class xoj::view::ImageView: public xoj::view::ElementView {
+public:
+    ImageView(const Image* image);
+    virtual ~ImageView();
+
+    /**
+     * Draws an Image element
+     */
+    void draw(const Context& ctx) const override;
+
+private:
+    const Image* image;
+};
--- a/src/core/view/LayerView.h
+++ b/src/core/view/LayerView.h
@@ -1,0 +1,36 @@
+/*
+ * Xournal++
+ *
+ * Displays a Layer
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+#include "util/Rectangle.h"
+
+#include "View.h"
+
+class Layer;
+
+class xoj::view::LayerView {
+public:
+    LayerView(const Layer* layer);
+
+    /**
+     * @brief Draws the portion of a Layer determined by drawArea
+     */
+    void draw(const Context& ctx, const xoj::util::Rectangle<double>& drawArea) const;
+
+    /**
+     * @brief Draws the entire Layer
+     */
+    void draw(const Context& ctx) const;
+
+private:
+    const Layer* layer;
+};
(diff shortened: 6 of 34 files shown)


bhennion:
Rebased on master

bhennion:
Rebased on master.
As mentioned in [this post](https://github.com/xournalpp/xournalpp/issues/1288#issuecomment-1066045585), this PR could be reviewed and merged as is to allow for further edits of ImageView to achieve #1288. The rest of the DocumentView class splitting can be done in a follow up PR.
@Technius Would this help?

src/core/view/View.h
```diff
@@ -0,0 +1,43 @@
...
+#include <gtk/gtk.h>
+
+namespace View {
+
+enum NonAudioTreatment : bool { FADE_OUT_NON_AUDIO_ = true, NORMAL_NON_AUDIO = false };
+enum EditionTreatment : bool { SHOW_CURRENT_EDITING = true, HIDE_CURRENT_EDITING = false };
+enum ColorTreatment : bool { COLORBLIND = true, NORMAL_COLOR = false };
+
+struct Context {
+    cairo_t* cairoContext;
```
Technius, line 23: This will be used everywhere. Should be named `cr`.

src/core/view/View.h
```diff
@@ -0,0 +1,43 @@
...
+    NonAudioTreatment fadeOutNonAudio;
+    EditionTreatment showCurrentEdition;
+    ColorTreatment noColor;
+};
+
+class ElementView {
+public:
+    virtual ~ElementView() = default;
+    virtual void draw(const Context& context) const = 0;
```
Technius, line 32: Rename to `ctx` since it will be used frequently.

Technius (changes requested):
* In the code, our "views" are actually conflating two different functions in the MVC model: controller (i.e., handling state changes) and view (i.e., rendering). For example, "ImageView" and "TextView" are actually like views in the MVC model. DocumentView and PageView are more like controllers that contain view logic, and should really be renamed to "DocumentController" and "PageController".
* We've started to use `namespace xoj` in new code, so the `View` namespace should be renamed to something like `xoj::view`. (Personally, I think we should have `namespace xournalpp` as it is much more obvious than "xoj").
* Views (in the renderer sense) should depend on elements but not vice versa, otherwise we'll end up with cyclic dependency spaghetti code. If there's any private data/methods we need to expose, we can write public getter functions.

src/core/view/View.h
bhennion: Done

bhennion:
> *DocumentView and PageView are more like controllers that contain view logic, and should really be renamed to "DocumentController" and "PageController".

I agree with the analysis. I think they should, in the end, be splitted into a controller and a view.
I did not rename them in this PR though. This'll wait for a follow-up PR, that should remove the remaining View bits in DocumentView and make DocumentView an actual controller.

The rest of your remarks have been implemented.
I'm turning this PR into a "non-Draft". A follow-up PR will pick up the work an finish cleaning up DocumentView.
By merging this PR like that, we do not delay the splitting of ImageView required in #3782.

bhennion:
I forgot to mention:

> * Views (in the renderer sense) should depend on elements but not vice versa, otherwise we'll end up with cyclic dependency spaghetti code. If there's any private data/methods we need to expose, we can write public getter functions.

I implemented the change anywhere I could, but... there is any issue with Text::calcSize that relies on TextView (it basically fakes a rendering and gets the bounding box of the rendering).
This is not great, but I do not see any other way of handling this.
I'm open to suggestions.

src/core/view/LayerView.h
```diff
@@ -0,0 +1,37 @@
...
+#include "util/Rectangle.h"
+
+#include "View.h"
+
+class Layer;
+
+class xoj::View::LayerView {
+public:
+    LayerView(const Layer* layer);
+    ~LayerView();
```
Technius, line 23: Don't need to declare (and define) the destructor if the default one is being used.

src/core/control/Control.cpp
```diff
@@ -78,7 +78,7 @@ Control::Control(GApplication* gtkApp, GladeSearchpath* gladeSearchPath): gtkApp
 
     this->applyPreferredLanguage();
 
-    TextView::setDpi(settings->getDisplayDpi());
+    xoj::View::TextView::setDpi(settings->getDisplayDpi());
```
Technius, line 81: I don't like how we have what is effectively a global variable controlling the dpi. How involved would it be to remove the global variable?

src/core/control/jobs/PreviewJob.cpp
```diff
@@ -96,6 +97,9 @@ void PreviewJob::drawPage() {
             break;
     }
 
+    xoj::View::Context context{cr2, xoj::View::NORMAL_NON_AUDIO, xoj::View::HIDE_CURRENT_EDITING,
+                               xoj::View::NORMAL_COLOR};
```
Technius, line 101: Since we might want to initialize contexts in a variety of settings, I suggest creating a series of helper methods to create "preset" context settings, e.g.

* `xoj::view::Context::createDefault(cr)`
* `xoj::view::Context::createForExport(cr)`

If necessary, the caller can modify the returned context before using it with the view(s).

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -43,13 +44,23 @@ void StrokeHandler::draw(cairo_t* cr) {
...
+        xoj::View::Context context{crMask, xoj::View::NORMAL_NON_AUDIO, xoj::View::HIDE_CURRENT_EDITING,
+                                   xoj::View::COLORBLIND};
+        xoj::View::StrokeView strokeView(stroke);
+        strokeView.draw(context);
     }
-    DocumentView::applyColor(cr, stroke);
 
-    cairo_set_operator(
-            cr, stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER ? CAIRO_OPERATOR_MULTIPLY : CAIRO_OPERATOR_OVER);
+    if (stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER) {
```
Technius, line 53: I think a comment here to explain the code would be helpful:
```suggestion
    // set the stroke color and blend mode
    if (stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER) {
```

src/core/view/SetsquareView.cpp
```diff
@@ -26,8 +26,11 @@ void SetsquareView::paint(cairo_t* cr) {
 
 void SetsquareView::drawTemporaryStroke(cairo_t* cr) {
     if (stroke) {
-        DocumentView docView;
-        docView.drawStroke(cr, stroke, 0);
+        ;
```
Technius, line 29: This looks like a typo.

Technius (commented):
Thanks for the changes! I did a quick skim through the code, will leave more comprehensive comments over the weekend.

`xoj::View` should be renamed to `xoj::view`, as the current capitalization suggests a class/struct and not a namespace.

> I implemented the change anywhere I could, but... there is any issue with Text::calcSize that relies on TextView (it basically fakes a rendering and gets the bounding box of the rendering). This is not great, but I do not see any other way of handling this. I'm open to suggestions.

I think the solution to this will be more complicated and will require more thought, so let's put it off until a later PR.

src/core/view/TextView.h
```diff
@@ -49,4 +48,8 @@ class TextView {
      * Sets the font name from Text model
      */
     static void updatePangoFont(PangoLayout* layout, const Text* t);
+
+private:
+    const Text* text;
+    static int DPI;
```
LittleHuba, line 54: This should be part of the settings class. We should not duplicate this info within the application.

src/core/view/TextView.cpp
```diff
@@ -43,54 +44,71 @@ void TextView::updatePangoFont(PangoLayout* layout, const Text* t) {
     pango_font_description_free(desc);
 }
 
-void TextView::drawText(cairo_t* cr, const Text* t) {
-    cairo_save(cr);
+void TextView::draw(const Context& ctx) const {
+    if (text->isInEditing()) {
+        return;
+    }
```
LittleHuba, line 50: What is the reasoning for this fast exit?

src/core/view/TextView.cpp
```diff
@@ -43,54 +44,71 @@ void TextView::updatePangoFont(PangoLayout* layout, const Text* t) {
...
+        return {};
+    }
+
     cairo_surface_t* surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 1, 1);
     cairo_t* cr = cairo_create(surface);
 
     PangoLayout* layout = initPango(cr, t);
-    string str = t->getText();
-    pango_layout_set_text(layout, str.c_str(), str.length());
+    std::string str = t->getText();
```
LittleHuba, line 86: Please make the variable name more descriptive. Also, reuse the variable further down.

src/core/view/TextView.cpp
```diff
@@ -104,8 +122,8 @@ void TextView::calcSize(const Text* t, double& width, double& height) {
...
```
LittleHuba, line 125: Please make the variable name more descriptive.

src/core/control/Control.cpp
bhennion: You mean removing the variable xoj::view::TextView::DPI, right?
This is a little bit more complicated than expected. We'd need TextView to have a pointer to the Settings instance, and to do so, we'd need every class creating a TextView to have a pointer to the Settings.
The worst case is model/Text (needs TextView for calcSize(), see discussoin above).

I think we should do a deeper refactoring of this class, but I'd leave is like that for now.

src/core/view/TextView.h
bhennion: See above

bhennion (commented):
Alright, I implemented most of the changes you recommended, with the one exception of the TextView::DPI global variable (see other comment).
I'll have a look into the MVC-push @LittleHuba suggested in a follow-up PR.

src/core/control/Control.cpp
LittleHuba: We should consider a singleton approach for the Settings class. That way we would not need to forward it through half the application.

LittleHuba:
I would assume you don't see changes because your DPI is set to 1 by default and this replacing it by 1 does not change anything?

LittleHuba:
In my opinion, you can leave the DPI stuff as it is because the solution of this is outside the scope of this PR. Well tackle this in a follow-up.

Febbe:
I remember, that there were many issues on export/printing, so maybe search for closed issues regarding the dpi. Also read https://docs.gtk.org/PangoCairo/func.context_set_resolution.html 72 results in an 1 to 1 scaling. Don't replace it with 1, replace it with 92, to use the default value, or a much larger value like 184.

bhennion:
> I would assume you don't see changes because your DPI is set to 1 by default and this replacing it by 1 does not change anything?

Nice try, but no ^^ My DPI is 95. I also tried setting my DPI to other values (much larger and much smaller), no changes.

bhennion:
> I remember, that there were many issues on export/printing, so maybe search for closed issues regarding the dpi.

Will do.

> Also read https://docs.gtk.org/PangoCairo/func.context_set_resolution.html 72 results in an 1 to 1 scaling.

Yeah, I read the doc, ofc. But it ain't so precise. I'm parsing through Pango's code right now, but this is huge.

> Don't replace it with 1, replace it with 92, to use the default value, or a much larger value like 184.

I tried that too!

bhennion:
> In my opinion, you can leave the DPI stuff as it is because the solution of this is outside the scope of this PR. We'll tackle this in a follow-up.

Yes of course. I was just toying around

bhennion:
> I've been looking at this DPI variable business. I actually don't understand what it's for. It is used excactly once, in this line of `TextView::initPango`
> 
> https://github.com/xournalpp/xournalpp/blob/4dacaee99b6d5ba95570656ae18d344e0dd180b3/src/core/view/TextView.cpp#L27
> 
> Now, replacing DPI with the constant value 1 here, I do not see any difference (neither in xournal++'s window, nor in the terminal where I printed the bounding boxes). Does anyone know enough about Pango to understand why?

It was added by this commit https://github.com/xournalpp/xournalpp/commit/eb9d52686391ed40a43bcf4379cac049ec01bd36
It was commited directly, not through a PR, so no more info there.

Anyway, that's a fight for another day.

Febbe:
It could be, that the dpi set there is either superseded by gtk3 or by other custom scaling code by us, so using the doubled size will be neglected by a dynamic rescaling to fit a specific rectangle. Unfortunately we can't just ask @wbrenna I think.

rolandlo:
In commit https://github.com/xournalpp/xournalpp/pull/2724/commits/4b60b78f8ae4ed1ecba621cf40225f1661699bfe fixing #1118 there were two ways to fix the issue (in which the size of the font was not the same in document coordinates when using different displayDpi values in the settings). The one I used was to use `pango_font_description_set_absolute_size` instead of the (experimental) `pango_font_description_set_size`. The alternative solution (as far as I remember) was to keep `pango_font_description_set_size`, but to replace `textDpi` by the constant `72`. In the latter case you can get rid of the `textDpi` variable.

bhennion:
> pango_font_description_set_absolute_size

I think you pointed out the right thing here: `pango_font_description_set_absolute_size` sets the size in cairo coordinates, while `pango_font_description_set_size` sets the size in pixel size. The latter thus depends on the DPI, while the former does not.

I think this is why the DPI value does not matter.
I removed the DPI variable entirely, and text seems to work as expected when changing the DPI setting. @rolandlo would you mind confirming that I did not reintroduce #1118 in the process?

I also created a BackgroundImageView, in case @Technius wants to change the image background handling too.

This is ready for another round of reviews, whenever you have time @Technius, @LittleHuba.

rolandlo:
> I think this is why the DPI value does not matter.
> I removed the DPI variable entirely, and text seems to work as expected when changing the DPI setting. @rolandlo would you mind confirming that I did not reintroduce #1118 in the process?

I can confirm that you did not reintroduce #1118.

bhennion:
> I can confirm that you did not reintroduce #1118.

Thanks a lot!

src/core/view/ElementView.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include "ImageView.h"
+#include "StrokeView.h"
+#include "TexImageView.h"
+#include "TextView.h"
+#include "View.h"
+
+using namespace xoj::view;
+
+auto ElementView::getFromElement(const Element* e) -> std::unique_ptr<ElementView> {
```
Technius, line 15: nitpick: `createFromElement` is more descriptive.

src/core/view/ImageView.h
```diff
@@ -0,0 +1,30 @@
...
+
+class Image;
+
+class xoj::view::ImageView: public xoj::view::ElementView {
+public:
+    ImageView(const Image* image);
+    virtual ~ImageView();
+
+    /**
+     * Draws a Image model
```
Technius, line 24: minor nitpick:
```suggestion
     * Draws an Image element
```

src/core/view/TextView.cpp
```diff
@@ -43,54 +38,72 @@ void TextView::updatePangoFont(PangoLayout* layout, const Text* t) {
...
-    do {
-        pos = StringUtils::toLowerCase(text).find(srch, pos + 1);
-        if (pos != -1) {
-            XojPdfRectangle mark;
-            PangoRectangle rect = {0};
-            pango_layout_index_to_pos(layout, pos, &rect);
-            mark.x1 = (static_cast<double>(rect.x)) / PANGO_SCALE + t->getX();
-            mark.y1 = (static_cast<double>(rect.y)) / PANGO_SCALE + t->getY();
+    size_t pos = text.find(pattern);
+    while (pos != std::string::npos) {
```
Technius, line 92: nitpick: I think this might be better as a for loop.

Technius (approved):
Thanks for these changes; this is a great improvement over the current code. I just have a few minor nitpicks. One more thing: I think we should merge `release-1.1` back into `master` and then rebase this PR, as there is a bugfix involving Text elements [`38258a4`](https://github.com/xournalpp/xournalpp/commit/38258a498f9edfc8cea0e842da745dcaa0fa7505).

---

Regarding @LittleHuba's comment:
> I don't really like that a view is created for each drawing operation. This makes drawing potentially very expensive. The lifetime of the view should be as long as that of the text itself.

Adding to this, I believe the elements that are more expensive to render (e.g., images, TeX, text) can benefit from some form of retained mode rendering. For elements that we always want to render using immediate mode rendering, we can eliminate the view struct/class altogether and use a freestanding function.

LittleHuba:
Sorry I'm really busy currently with moving. I'll try to give it a look later today but I can't promise anything. Regarding the comment from @Technius about free functions for some views. I disagree on that. While I understand the point of saving some complexity in storage, this step would add complexity in the rendering path because we have to differ between element types again. That is definitely unwanted as we then pay this by branch mispredictions and thus loss of performance.

bhennion:
I took care of @Technius' nitpicks.

> Regarding the comment from @Technius about free functions for some views. I disagree on that. While I understand the point of saving some complexity in storage, this step would add complexity in the rendering path because we have to differ between element types again. That is definitely unwanted as we then pay this by branch mispredictions and thus loss of performance.

I guess an argument for @Technius' point of view is that most element views (typically, strokes) do not involve any caching. I should point out here that this should change once splines are used: to draw a stroke stored as a spline (with pressure values), we need to transform it into a bunch of line segments of various width. Those line segments should be cached by the StrokeView.
So, if most element views involve caching, I guess @LittleHuba's complete tree seems simpler and more efficient.


I had another question related to this (and a follow up PR), concerning the background "painters". Right now, there is, in DocumentView, one instance of each possible type of background painter (for the simple backgrounds, not pdf nor image).
If we follow @LittleHuba's tree structure to the letter, this should be replaced by one instance of the right `BackgroundView` per `PageView` (an actual View, not the current PageView that is actually a controller). Except most note-taking documents would typically have several pages, all with the same background. It looks like a waste to me to have one `BackgroundView` per `PageView` (although the logic would be cleaner).
Any thoughts on this?

LittleHuba (commented):
Thanks for your continued work. You made amazing progress! 
Just one issue that I'm not yet comfortable with but otherwise this can go in as it is.

A note for later on:
In a follow-up PR we should create some RAII helpers to take care of the correct handling of cairo_save() and cairo_restore() but this is out of scope for this PR.

bhennion:
Thanks for the feedback!
I removed the BackgroundView::draw taking a rectangle as a parameter. In fact, we need to fill the mask entirely, so any other (future) BackgroundView can actually infer the rectangle from the mask's matrix.

I'll wait for #3782 to be merged and rebase thereon.

LittleHuba:
Thanks for your work. This is a huge step towards a more clean code base!

Please add a comment when somebody should reapprove.

bhennion:
I rebased on master. Besides that, since your last review, I removed unecessary includes in BackgroundView/ImageBackgroundView and renamed BackgroundImageView to ImageBackgroundView.
The renaming is in prevision for a unification of background handling. We'll have have a `class SomethingBackgroundView: public BakgroundView` for every knid of backgrounds (image, pdf, graph, isometric, ...)

@LittleHuba @Technius could you give it a last review before we merge this?

LittleHuba (approved):
LGTM. There is definitely some work left for follow-up PRs but all of it is out of the scope of this PR. Thanks for the great work!

bhennion:
Alright, merging in 48 hours unless an objection is raised.

Issue #3500, opened 2021-10-24
Crash when using the text editor tool


Rio6:
**Affects versions :**
 - OS: Artix Linux
 - (Linux only) Desktop environment: X11, awesomewm
 - IM: gcin 2.9.0
 - Which version of libgtk do you use: 3.24.30
 - Version of Xournal++: Found on 1.1.0, git bisect points to c237d3e5f8f190cea9e4098e62cf7f7362a29ed8
 - Installation method: occurs both for Archlinux package and compile from source

**Describe the bug**
Xournal++ crashes when clicking on the page using text editor tool.

**To Reproduce**
1. Open Xournal++
2. Select text editor
3. Click anywhere on the page

**Expected behavior**
Text editor get activated on page.

**Screenshots of Problem**

https://user-images.githubusercontent.com/15951245/138579919-b393266e-0170-478f-b7b3-35626350ceed.mp4

**Additional context**
Following outputs are from c237d3e5f8f190cea9e4098e62cf7f7362a29ed8

Stdout:
```
(xournalpp:7646): Gdk-CRITICAL **: 00:10:15.007: gdk_device_get_source: assertion 'GDK_IS_DEVICE (device)' failed

(xournalpp:7646): Gdk-CRITICAL **: 00:10:15.007: gdk_device_get_name: assertion 'GDK_IS_DEVICE (device)' failed
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid

** (xournalpp:7646): WARNING **: 00:10:15.007: [Crash Handler] Crashed with signal 6
```
Log file:
```
Date: Sat Oct 23 23:50:32 2021
Error: signal 6
[bt]: (0) ./src/xournalpp(+0x431945) [0x55a05baf4945]
[bt]: (1) /usr/lib/libc.so.6(+0x3cf80) [0x7fc9d0b18f80]
[bt]: (2) /usr/lib/libc.so.6(gsignal+0x145) [0x7fc9d0b18ef5]
[bt]: (3) /usr/lib/libc.so.6(abort+0x116) [0x7fc9d0b02862]
[bt]: (4) /usr/lib/libstdc++.so.6(+0x99802) [0x7fc9d0d5d802]
[bt]: (5) /usr/lib/libstdc++.so.6(+0xa5c8a) [0x7fc9d0d69c8a]
[bt]: (6) /usr/lib/libstdc++.so.6(+0xa5cf7) [0x7fc9d0d69cf7]
[...]

PR #3511, opened 2021-10-26
Add null check to device in InputEvent's handler

Rio6:
Didn't have much progress on #3507 so opening this instead for now.

This PR makes `InputContext::handle` to return early if the event device is null, which would cause translateEvent to throw otherwise.

Related issue #3500


Rio6:
retriggered ci and it's good now. Good to merge?

bhennion:
LGTM! Merging in 24 if no objections are raised.

PR #3518, opened 2021-10-27
Fixing typo

Varpie:
When opening Xournal++ after an update changing where the configuration files are located, users are greeted with a message that contains an error: it says "Due to a recent update, Xournal++ has changed where **it's** configuration files are stored."

With these changes, the message is corrected so that users are not greeted by a typo.


personalizedrefrigerator:
Merging (_and squashing_) in 24 hours unless there are any objections.

bhennion:
> Merging (_and squashing_) in 24 hours unless there are any objections.

LGTM! Just think of rebasing to `release-1.1` before merging this

PR #3530, opened 2021-10-30
Code cleanup around ShapeRecognizer

bhennion:
With 07ed4f7 and the creation of the `InputHandler`'s derived classes (e.g. `StrokeHandler`), the Shape recognizer feature lost the possibility to combine older strokes into the recognized pattern. For instance, before that commit, one could get a rectangle by drawing 4 lines, even if one lifted the pen between those lines.

This is not possible anymore because since 07ed4f7 the `ShapeRecognizer` gets created with every starting stroke, and thus does not remember previously recognized lines.

This PR **does not reinstate this behaviour**. What we do here is simply to clean up some bits of code that were made pointless with the change described above.

I only changed the easy stuff, without going into the inner cogs of the ShapeRecognizer.

The functionalities are (supposedly) not altered in any way by this PR.

Some further simplifications of the recognizer's code would be possible. They may come in a follow up PR.

[PS: I came to have a look at this because #2729 needs to adapt the Recognizer to splines]


bhennion:
Rebased on master

bhennion:
Unless someone objects, I'll merge this removal of dead code in 72 hours.

Issue #949, opened 2019-02-28
Input lag while using the eraser on surface pro


alexanderi96:
The eraser results very laggy after long session of usage on my surface pro 2017

andreasb242:
Which operating system are you using? Addional: X11, Wayland etc?

alexanderi96:
Right now I'm using Ubuntu 18.10, the latest jakeday's surface kernel (4.19.23) and yes, I'm running under Wayland. Actually I must check if the problem persists with x11

alexanderi96:
I can confirm that the problem persists under x11 as well (even if it is less evident)

GregDMeyer:
I've noticed this as well, but only in HiDPI mode. Setting GDK_SCALE=1 I have no issue.

LittleHuba:
Please provide a video of the problem. Maybe this will help to get to the actual problem. With the current description it can be anywhere in the whole stack.

saengowp:
My stylus is also laggy under Wayland (but not X11). The first continuous stroke is responsive while the next stroke wouldn't be displayed until the pen is lifted. 

I resolved it by

- Use a mouse to move the cursor out of the Xournal++ window then use the stylus only inside the drawing page. This resolve the issue
- **OR** Change the preference to enable the new input system. This also resolve the issue.

**Environment**
- Hunion H420 Drawing Tablet
- Xournalpp 1.0.17
- WM: Sway (Wayland)
- OS: Arch Linux

violoncelloCH:
I'm just checking the latest nightly build (xournalpp-1.1.0.dev-nightly.20210103-x86_64.AppImage) and I'm getting an Input lag also when using the new input system. However I only get the lag for the eraser and not while drawing. I can easily reproduce it in an empty document by drawing some parallel strokes and then trying to erase them. The more individual lines I cross with my eraser stroke, the longer it takes until the strokes actually disappear. (e.g. ~ 2 sec for erasing 25 lines). 
To be specific: this is when I do a fast eraser stroke and the lines all disappear at the same time after the delay. However when erasing with a slow stroke the lines disappear almost immediately and one by one.
I also need to mention that this only happens on Wayland not for X11.
I'm happy to provide further information; just let me know what could help debugging this.

caleb-allen:
I'm on Wayland and I'm experiencing exactly what @violoncelloCH describes

LittleHuba:
[...]

PR #3532, opened 2021-10-31
Rewrite Eraser (backport from #2729)

### LINKED ISSUE #997: Eraser doesn't work with marker

bhennion:
This is a backport of the new eraser developed for #2729. It has several advantages over the current eraser (see video below):

- Fix #997
- Fix the lag mentionned in [this post](https://github.com/xournalpp/xournalpp/discussions/2767#discussioncomment-347387)
- It could help with #949.
- Better display of dashed strokes (still quite imperfect though). See #3537.
- It is ready for when the splines will attack.

https://user-images.githubusercontent.com/17818041/139596304-b43c374d-cd40-42dc-974d-5778f17edbd6.mp4


LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMTYzMjgvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMTYzMjgvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMTYzMjgvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 18.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMTYzMjgvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0xOC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMTYzMjgvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)

LittleHuba:
Nice work!

There seems to be a bug though: Please check out the blue highlighter at the end of your video in the new implementation. The stroke changes its appearance way after parts of it were erased...

bhennion:
> There seems to be a bug though: Please check out the blue highlighter at the end of your video in the new implementation. The stroke changes its appearance way after parts of it were erased...

Yes, the change is happening when I release the mouse's button. The issue is quite simple.
It starts with an observation that has nothing to do with how the eraser is implemented: when we erase part of a stroke A, the stroke is split in two (or more) parts that will become strokes B and C. Now, B and C may overlap leading to a change in appearance on their intersection. See for instance those rectangles of which I erased a bit:

Screenshot_20211031_211220

To fix this would go beyond the eraser: one would need to remember that two strokes come from the same one, and always draw them on the same mask. That is possible but requires changes in the file format.

Now, without fixing this more general question, we could make sure that when using the eraser, what you see is what you get. Here, the issue is that I only rerender a small area around the eraser, so if there was some far away intersection of the new strokes B and C, it will not be displayed properly until the button is released and the strokes are fully repainted. I see two solutions:

1. Rerender the entire bounding box of the stroke at every eraser movement. This would cost to much, IMO, especially for MacOS users.
2. Compute where the new components of the stroke intersect and redraw this area. Again, quite costful, I think.

I'm open to any suggestions!

LittleHuba:
I would go with solution 2. You can quite aggressively cut the cases where a costly check for intersection of the actual stroke segments must be done by first checking the bounding boxes for intersection.

bhennion:
I finally implemented a binary-tree-based algorithm for finding out where to rerender when cutting a highlighter stroke in two. The result:

https://user-images.githubusercontent.com/17818041/141655291-fdf67b18-0207-416f-9121-d8f026fe62b0.mp4

bhennion:
Some progress here, but still one issue (that I know of). See

https://user-images.githubusercontent.com/17818041/142190000-c9395135-43fa-438b-858a-9a801ef3949c.mp4

So, as you can see:

- Filled strokes are now rerendered when they should. This means quite a large area can be rerendered, so this will possibly lag on MacOs.
- Closed strokes (typically rectangles and ellipses) are now treated as such and we can no longer see where the first and last points meet.

But two things left to do:

- [ ] Figure out how to render properly filled highlighter strokes being erased. The issue here is that those strokes are painted on a mask and then blitted, so all the different bits of the stroke being erased are blitted at the same time.
- [ ] Move `ErasableStroke::draw` and `ErasableStroke::drawFilling` to a new class `view/ErasableStrokeView`

bhennion:
I think this is ready for a round of tests. All the kind of strokes should now be rendered when erasing "as it will look like once you lift the pen". It works very smoothly on my machine.

To anyone willing to help: here is a file with various kind of strokes [eraser-test.zip](https://github.com/xournalpp/xournalpp/files/7565115/eraser-test.zip). Could you run this PR (some builds are available in the post above), open this file (or any other), try erasing stuff (with the standard eraser) and tell me if something feels wrong (lag, bug or other UX oddities you dislike)?

Some lags are kinda expected when erasing filled strokes on MacOs, hopefully not to much... I don't think I can do anything about it anyway.

Because of #997, I made some choices (see below) about what parts of a stroke are erased when the eraser box is near a stroke. If you could tell me how it feels for you, particularly with very thick highlighter strokes, it would help!

The particular choices I made are explained in the file `src/model/eraser/README-Eraser-and-padded-box.pdf` that I reproduce here (in the code, the padding is `0.4 * stroke.width`, so much much smaller than in the picture)
README-Eraser-and-padded-box

bhennion:
Rebased on master. I'd still like to get some feedback on this (especially the UX, lags or inconveniences one might encounter).

bhennion:
I improved the data structures for better performances. There are some things left to do here:

- [ ] Fix rarely reproducible bug
- [ ] Write comments: explain how this works!
- [ ] Write test units
    - [ ] intersection with padded boxes
    - [ ] overlap trees
    - [ ] partial clones (getStrokes)
    - [ ] TinyVector & SmallVector

bhennion:
I think this is ready for a first review.
I still have to write a couple of test units, but that shouldn't be an issue for a review.

bhennion:
Rebased on master. This is still up for review.

Febbe:
The problem here are the 2500+ lines of code. I mostly have time for short reviews.
Can you elaborate, which changes should be taken care of? 
Also, @LittleHuba did already many reviews. 
What is your status there @LittleHuba?

src/core/model/Point.cpp
```diff
@@ -8,18 +8,15 @@ Point::Point(double x, double y, double z): x(x), y(y), z(z) {}
...
-    y *= factor;
-    x += this->x;
-    y += this->y;
-
-    return Point(x, y);
+auto Point::lineTo(const Point& p, double length) const -> Point { return relativeLineTo(p, length / lineLengthTo(p)); }
+
+auto Point::relativeLineTo(const Point& p, const double ratio) const -> Point {
+    const double oneMinusRatio = 1 - ratio;
+    return Point(oneMinusRatio * x + ratio * p.x, oneMinusRatio * y + ratio * p.y, oneMinusRatio * z + ratio * p.z);
```
Febbe, line 15: Why not just x1 + (x2 - x1) * ratio?

src/core/control/tools/StrokeStabilizer.cpp
```diff
@@ -118,10 +118,8 @@ void StrokeStabilizer::Active::quadraticSplineTo(const Event& ev) {
     /**
      * The quadratic control point is converted into two cubic control points
      */
-    // Equivalent to fp = B.lineTo(Q, 2/3*distance), but avoids recomputing the norms
-    Point fp((Q.x - B.x) * 2 / 3 + B.x, (Q.y - B.y) * 2 / 3 + B.y);
-    // Equivalent to sp = C.lineTo(Q, 2/3*distance), but avoids recomputing the norms
-    Point sp((Q.x - C.x) * 2 / 3 + C.x, (Q.y - C.y) * 2 / 3 + C.y);
+    Point fp = B.relativeLineTo(Q, 2 / 3);
+    Point sp = C.relativeLineTo(Q, 2 / 3);
```
Febbe, line 122: 2 / 3 is 1. You have to use 2.0 / 3.0
```suggestion
    Point fp = B.relativeLineTo(Q, 2.0 / 3.0);
    Point sp = C.relativeLineTo(Q, 2.0 / 3.0);
```

bhennion: Right, I'll change that too

src/util/include/util/UnionOfRectangles.h
```diff
@@ -0,0 +1,52 @@
...
+        data.erase(std::remove_if(std::next(begin(data)), end(data),
+                                  [=](auto const& other_rect) {
+                                      // intersecting rects are united in data.front();
+                                      //                                       other_rect.contains(rect);
+                                      return other_rect.x <= rect.x && other_rect.y <= rect.y &&
+                                             other_rect.x + other_rect.width >= rect.x + rect.width &&
+                                             other_rect.y + other_rect.height >= rect.y + rect.height;
+                                  }),
+                   end(data));
+    }
```
bhennion, line 49: This proposal does not pass `ctest`.
There needs to be several passes because as the `last_united` grows, it may start intersecting (without containing) another rectangle that we already tested. See the test case in test/unit_tests/util/RectangleTest.cpp.

Let me try to find a better version of that using more std::algorithm than my original code.

bhennion: I made a new proposal that passes the `ctest` and is more clear, I think.

bhennion:
Thanks for starting this review @febbe! That's greatly appreciated!

> Can you elaborate, which changes should be taken care of?

Besides the code improvements, I think some feedback on how the eraser actually feels (UX-wise) could help.
Also, if you feel some code portions would benefit some more comments, I'll be happy to add some more.

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
...
+        s->points.reserve(upperBound.index - lowerBound.index + 2);
+
+        // First point
+        s->points.emplace_back(this->getPoint(lowerBound));
+
+        // end point of segment of index lowerBound.index
+        auto it = this->points.cbegin() + (std::ptrdiff_t)lowerBound.index + 1;
+
+        // end point of segment of index upperBound.index
+        auto endIt = this->points.cbegin() + (std::ptrdiff_t)upperBound.index + 1;
```
Febbe, line 100:
```suggestion
        auto endIt = std::next(begin(this->points, std::ptrdiff_t(upperBound.index + 1));
```

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
...
+
+        std::copy(it, endIt, std::back_inserter(s->points));
+
+        // Last point
```
Febbe, line 104: I think, this is over explaining:

```suggestion
        s->points.reserve(upperBound.index - lowerBound.index + 2);
        s->points.emplace_back(this->getPoint(lowerBound));
        auto const begi = std::next(begin(this->points, std::ptrdiff_t(lowerBound.index + 1));
        auto const endi= std::next(begin(this->points, std::ptrdiff_t(upperBound.index + 1));
        std::copy(begi, endi, std::back_inserter(s->points));
```

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
...
+std::unique_ptr<Stroke> Stroke::cloneSection(const PathParameter& lowerBound, const PathParameter& upperBound) const {
+    auto s = std::make_unique<Stroke>();
+    s->applyStyleFrom(this);
+    s->width = this->width;  // stroke width, not bounding box width
+
+    if (upperBound.index == lowerBound.index) {
+        const Point& p = this->getPoint(lowerBound);
+        const Point& q = this->getPoint(upperBound);
+        s->points = {p, q};
+    } else {
```
Febbe, line 90: Assuming the function is used correct, can it happen, that this case isn't handled also by the else branch?
In my opinion, `upperBound.t != 0.0` should never be true.

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
 
 auto Stroke::clone() -> Element* { return this->cloneStroke(); }
 
+std::unique_ptr<Stroke> Stroke::cloneSection(const PathParameter& lowerBound, const PathParameter& upperBound) const {
```
Febbe, line 81: Might be a bit confusing, having c++ iterators always point to the next element and an invalid last iterator pointing to the first element, out of bounds. And then this function takes an end argument, which points to the second last element.

I would rewrite the get Point function to be static and to accept 2 Points. From that, the result is generated.
upperBound.t is assumed to be >0.0 and lowerBound.t < 1.0. upperBound should be then either last or last + 1.

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
...
+    std::copy(startIt, this->points.cend() - 1, std::back_inserter(s->points));
+
+    auto endIt = this->points.cbegin() + (std::ptrdiff_t)endParam.index + 1;
+    std::copy(this->points.cbegin(), endIt, std::back_inserter(s->points));
+
+    // Last point
+    s->points.emplace_back(this->getPoint(endParam));
+
+    // Remove unused pressure value
+    s->points.back().z = Point::NO_PRESSURE;
```
Febbe, line 138: **Nitpick (also the comment above): ** Use std::next, and explain why you have to do something, not what.

For example, why does the startIter >= endIter? I think, this is even clear in this context.

src/core/model/Stroke.cpp
```diff
@@ -72,6 +78,68 @@ auto Stroke::cloneStroke() const -> Stroke* {
...
+
+    s->points.reserve(this->points.size() - startParam.index + endParam.index + 1);
+
+    // First point
+    s->points.emplace_back(this->getPoint(startParam));
+
+    auto startIt = this->points.cbegin() + (std::ptrdiff_t)startParam.index + 1;
+    assert(startIt != this->points.cend());
+    std::copy(startIt, this->points.cend() - 1, std::back_inserter(s->points));
+
```
Febbe, line 130: You skipped the last point, why?
Shouldn't it be added, but with the previous z value?

src/core/model/Stroke.cpp
```diff
@@ -198,6 +266,21 @@ auto Stroke::getPoint(int index) const -> Point {
     return points.at(index);
 }
 
+Point Stroke::getPoint(PathParameter parameter) const {
+    if (parameter.index >= this->points.size() - 1) {
+        return this->points.back();
+    }
```
Febbe, line 272: I would just assert this. In my opinion, this is wrong usage.

src/core/model/Stroke.h
```diff
@@ -116,6 +162,12 @@ class Stroke: public AudioElement {
 protected:
     void calcSize() const override;
 
+private:
+    static std::optional<Interval<double>> intersectLineWithRectangle(const Point& p, const Point& q,
+                                                                      const xoj::util::Rectangle<double>& rectangle);
+    static TinyVector<double, 2> intersectLineSegmentWithRectangle(const Point& p, const Point& q,
```
Febbe, line 168: How does this compile? TinyVector should be incomplete.

src/core/model/Stroke.h
```diff
@@ -20,6 +22,18 @@ enum StrokeTool { STROKE_TOOL_PEN, STROKE_TOOL_ERASER, STROKE_TOOL_HIGHLIGHTER }
 enum StrokeCapStyle { ROUND, BUTT, SQUARE };
 
 class ErasableStroke;
+struct PaddedBox;
+struct PathParameter;
+
+template <class T, unsigned int N>
+class TinyVector;
```
Febbe, line 29: I think, the dependency is pulled in somewhere else. I would include it directly. Otherwise, the compilation will fail, if the indirect dep is not included there anymore.

src/util/include/util/UnionOfRectangles.h
Febbe: Good you made those tests. I am just not sure, if it would be faster to only allow one repaint rectangle. Currently, we will have at worse a runtime of P(S) (power set), which is extremely dangerous. Maybe a more heuristic but not so correct approach is better. 

(I would leave a comment that this can be a pitfall, we can fix that later if required.)

src/core/model/Stroke.cpp
bhennion: I assume you meant `upperBound.t == 0.0`, right? Even then, there is (at least) one case where it can happen: `lowerBound = (0, 0.0)` and `upperBound = (0, 0.0)`. This is obviously a corner case, but still.

Also, I thought having this function (and the other similar functions like `ErasableStrokeView::draw`) as generic as possible would limit the number of corner-case-tests when computing those bounds.

src/core/model/Stroke.cpp
bhennion: By assumption here, the stroke's first point equals its last point (e.g. the stroke is a rectangle or an ellipse). Our clone starts somewhere on the stroke, pass by the starting/end point, and end somewhere else. The point is: we want only one copy of that starting/end point in our new stroke. I chose the first for the pressure value business.

I agree about the last-has-point-has-no-pressure situation. This should be fixed with #2729 (for a spline, the last point's pressure matters, and I made piecewise linear paths behave similarly).

src/core/model/Stroke.cpp
bhennion: The logic is built to consider the path as a succession of segments (of which any consecutive two happen to share an endpoint, both geometrically and in memory). The value PathParameter.index is the segment's index (there is one less segments than points).
For a perfect C++ like upper bound, we'd have to do something like (n, std::nextafter(t, 2.0)). I don't think it makes much sense.

PS: the segment oriented logic stems from #2729: it makes more sense to think about spline segments than about points. Then, for homogeneity, the same is applied to piecewise linear paths.

src/core/model/Stroke.cpp
bhennion: You're right this case could be dealt with by the `else` branch (thus removing the if statement), removing the `if (upperBound.t != 0.0)` below, and asserting that `upperBound.t` can only be 0.0 if `lowerBound == upperBound`.

src/util/include/util/UnionOfRectangles.h
bhennion: Yeah, it's really not clear how much we gain from being greedy on the rerendered areas. Clearly, there's the MacOs hardware acceleration issue playing a part in those choices.

Maybe it'd be interesting to get how many rectangles are (usually) rerendered in by a render job (see `PageView` L700+ and `RenderJob`). If this is close to 1, then this class can be dropped altogether, and replaced with a single rectangle to which we add everything.

src/core/model/Stroke.h
bhennion: I think forward declaration of the return type of a **static** member function is allowed. The header for TinyVector is definitely not included, even indirectly, in this file. Its actually never included in any header file in the project.

bhennion:
> Will take a look into this when I have time.
Thanks a lot!

> Did you checked the debug stacktrace?
No. I did try to set the `UNDO_TRACE` flag to get some debug, but it does not seem to compile with the variable #define'd anymore.

src/core/model/Stroke.cpp
```diff
@@ -375,6 +446,153 @@ auto Stroke::intersects(double x, double y, double halfEraserSize, double* gap)
...
+            result.emplace_back(0U, 0.0);
+            result.emplace_back(0U, 0.0);
+            return result;
+        }
+        return IntersectionParametersContainer();
+    }
+    return this->intersectWithPaddedBox(box, 0, pointCount - 2);
+}
+
+auto Stroke::intersectWithPaddedBox(const PaddedBox& box, size_t firstIndex, size_t lastIndex) const
```
Febbe, line 463: Can you split this up for me, either into local lambdas or static (file local) functions, with good explaining names 😅 .. 

For a reviewer, this function is horrible, since it has many branches high level of nesting, etc. also it is most likely, that a bug is hiding there, but neither the implementor nor any reviewer will find it.

Therefore, many people defined some metrics to prevent that:
 - e.g. functions should not be longer than 20 lines (of code),
 - reduce multiline comments
 - explain why something must or must not happen, not what.
 - and last but not least: [clang-tidy-check](https://clang.llvm.org/extra/clang-tidy/checks/readability-function-cognitive-complexity.html) [paper cognitive-complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf)

Febbe: for example: ` * Initialize with the half tangent at the first point ` can be a function with exact that name.

src/core/model/Stroke.cpp
```diff
@@ -424,18 +642,76 @@ void Stroke::calcSize() const {
...
+        // Horizontal segment
+        double y = p.y;
+        if (y <= rectangle.y || y >= rectangle.y + rectangle.height) {
+            return std::nullopt;
+        }
+        double norm = 1.0 / (q.x - p.x);
+        double tUp = (rectangle.x - p.x) * norm;
+        double tBottom = tUp + rectangle.width * norm;
+        return Interval<double>::getInterval(tUp, tBottom);
+    }
```
Febbe, line 685: First, this is duplicate code and should be moved into a more generic lambda.
Second, I don't think, that both branches are too unlikely to specify a specific function for them.
The generic case might be enough.

src/core/model/Stroke.h
```diff
@@ -116,6 +162,12 @@ class Stroke: public AudioElement {
 protected:
     void calcSize() const override;
 
+private:
+    static std::optional<Interval<double>> intersectLineWithRectangle(const Point& p, const Point& q,
```
Febbe, line 166: What is the meaning of Interval::min / max, how do they relate to p,q, rectangle? The optional is obvious.

src/util/include/util/Interval.h
```diff
@@ -0,0 +1,139 @@
...
+    const_iterator begin() const { return reinterpret_cast<const_iterator>(target.data()); }
+    const_iterator end() const {
+        if (target.size() % 2) {
+            return reinterpret_cast<const_iterator>(target.data() + target.size() - 1);
+        } else {
+            return reinterpret_cast<const_iterator>(target.data() + target.size());
+        }
+    }
+
+    size_t size() const { return static_cast<size_t>(end() - begin()); }
```
Febbe, line 135:
```suggestion
    size_t size() const { return std::distance(begin(), end()); }
```

src/util/include/util/Interval.h
```diff
@@ -0,0 +1,139 @@
...
+    bool operator<(const Interval<T>& other) const { return !(other <= *this); }
+    bool operator>(const Interval<T>& other) const { return !(*this <= other); }
+};
+
+/**
+ * Template class that converts a sorted contiguous container into an iteratable of type Interval
+ * It (virtually) pairs up consecutive elements of the vector 2 by 2 to make Interval's.
+ */
+template <class contiguous_container_type>
+class IntervalIteratable final {
```
Febbe, line 118:
```suggestion
class IntervalView final {
```

src/util/include/util/Interval.h
```diff
@@ -0,0 +1,139 @@
...
+public:
+    IntervalIteratable(const contiguous_container_type& target): target(target) {}
+    ~IntervalIteratable() = default;
+
+    using value_type = Interval<typename contiguous_container_type::value_type>;
+    using const_iterator = const value_type*;
```
Febbe, line 124: It would be cleaner, to Implement the Iterator as a pair of 2 iterators, returning an Interval as copy.

Febbe:
23/40, note that some comments are minor nit's even I did not write it explicitly.

src/core/model/Stroke.cpp
bhennion: No, those cases do happen as soon as the rectangle tool is used, or the pdf ruler, or a line is recognized and snapped to the grid, etc.
Then, you're right, it could be made into a lambda or so.

src/core/model/Stroke.h
bhennion: Will add a comment

src/core/model/Stroke.cpp
bhennion: Let me try :-)

src/util/include/util/Interval.h
bhennion: Maybe, be unnecessarily less efficient. For a dereferentiation, it'd go from "1 dereferentiation" to "2 dereferentiations + 2 copies into newly allocated mem". I'd rather keep it that way.

LittleHuba:
@Febbe I see that you already started on the review so I'll leave this to you. No sense in both of us going over it.

src/util/include/util/Interval.h
Febbe: Don't underestimate the compiler. 
References are often cost free, since they are optimized away. Also you would get rid of the reinterpreted cast, which is extremely dangerous here. Actually this can be UB in the case, that the alignment of the data types are different.
And last but not least, the returned object will be returned by value, with return value optimisation, it will be a noop. And when the object ist used directly, it might not even touch the stack, instead it can stored directly in an register.

src/core/model/Stroke.h
bhennion: done

src/core/model/Stroke.cpp
bhennion: done

src/core/model/Stroke.h
bhennion: done

src/util/include/util/Interval.h
bhennion: Ok. I removed the IntervalView class altogether.

bhennion: (Note: we'll still need to have this conversatio for #2729, there are other View's there based on the same principle -- see Spline::SegmentIteratable. I don't really understand your argument yet)

src/core/model/Stroke.cpp
bhennion: I made an attempt. You can have a look whenever you have time

src/util/include/util/UnionOfRectangles.h
bhennion: I made some measurements. The average number of rectangles rendered by a single RenderJob run is around 1.01. (I once reached 1.02 :-). So yeah, we should just unite all rectangles.

bhennion: I removed UnionOfRectangles in the last commit.

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+        if (nb < N) {
+            dataArray[nb] = std::move(value);
+        } else if (nb == N) {
+            dataVector.reserve(N + 1);
+            std::move(dataArray.begin(), dataArray.end(), std::back_inserter(dataVector));
+            dataVector.push_back(std::move(value));
+        } else {
+            dataVector.push_back(std::move(value));
+        }
+        ++nb;
```
Febbe, line 77: Just call emplace_back here

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+        if (nb > N) {
+            if (nb == N + 1) {
+                std::move(dataVector.begin(), std::prev(dataVector.end()), dataArray.begin());
+                dataVector.clear();
+            } else {
+                dataVector.pop_back();
+            }
+        }
+        --nb;
+    }
```
Febbe, line 102: **Minor nit:** I know this is way easier, but keeping the data in the vector might be faster (hardly depends on N).
E.g. when having N*sizeof(T) is less than one cache line, this makes probably no difference.
Also, dataVector.clear(); is faster, because it does nothing (besides a change of the end-iterator). If you planned to free memory, you should re-assign an empty vector: `dataVector = std::vector<value_type>(0);`.

I'll leave that to you.

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+    void clear() {
+        nb = 0;
+        /**
+         * Reset capacity. If clear() is called a lot and more than N objects are usually pushed, don't use this class!
+         */
```
Febbe, line 107: Move that comment to the class description.

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+        dataVector = {};
```
Febbe, line 108: In the current state, since the invariant that all data is located at the stack is true, when nb<=N, you could just ignore dataVector, or also use `dataVector.clear()`. Just make sure you make the same decision as in `pop_back()`

src/util/include/util/SmallVector.h
bhennion: Done. Same in TinyVector

src/util/include/util/SmallVector.h
bhennion: Done

src/util/include/util/SmallVector.h
bhennion: I find it easier to keep the invariant "if nb <= N then the data is in the stack". Also less confusing for the future user of this class. I'll keep this behaviour (with the suggested `dataVector = std::vector<value_type>(0);`).

src/util/include/util/SmallVector.h
bhennion: done

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include <array>
+#include <vector>
+
+template <class T, unsigned int N>
+class SmallVector {
+public:
+    SmallVector() = default;
+
+    using size_type = unsigned int;
```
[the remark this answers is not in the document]
bhennion: I also added the missing Container member types:
```
    using size_type = size_t;
    using value_type = T;
    using reference = T&;
    using iterator = T*;
    using const_iterator = const T*;
    using difference_type = ptrdiff_t;
```

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+
+    T& operator[](size_type pos) { return dataArray[pos]; }
+    const T& operator[](size_type pos) const { return dataArray[pos]; }
+    T& front() { return dataArray.front(); }
+    const T& front() const { return dataArray.front(); }
+    T& back() { return dataArray[nb - 1]; }
+    const T& back() const { return dataArray[nb - 1]; }
+
+    using const_iterator = const T*;
+    using iterator = T*;
```
Febbe: You should use the array iterator here. It has more guarantees for debug compilation.

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+
+    iterator begin() { return dataArray.data(); }
+    const_iterator begin() const { return dataArray.data(); }
+    iterator end() { return dataArray.data() + nb; }
+    const_iterator end() const { return dataArray.data() + nb; }
```
Febbe, line 42:
```suggestion
    iterator begin() { return begin(dataArray); }
    const_iterator begin() const { return begin(dataArray); }
    iterator end() { return std::next(begin(dataArray), nb); }
    const_iterator end() const { return std::next(begin(dataArray), nb); }
```

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+
+    void push_back(const T& value) {
+        if (nb < N) {
+            dataArray[nb++] = value;
+        } else {
+            throw std::length_error("TinyVector overflow");
+        }
+    }
```
Febbe: std::array accesses are tested in debug mode, so there is no reason to waste computation power here:
```suggestion
    void push_back(const T& value) {
        new (&dataArray[nb++]) T(value); 
    }
```

Febbe: Also just call emplace_back for both push_back's

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+    template <class... Args>
+    void emplace_back(Args&&... args) {
+        if (nb < N) {
+            dataArray[nb++] = T(std::forward<Args>(args)...);
+        } else {
+            throw std::length_error("TinyVector overflow");
+        }
+    }
```
Febbe, line 51: The following ensures real emplacement, it is only allowed, because `std::array` is always uninitialized:
```suggestion
    template <class... Args>
    void emplace_back(Args&&... args) {
        new (&dataArray[nb++]) T(std::forward<Args>(args)...);
    }
```

Febbe, line 51:
```suggestion
    void emplace_back(Args&&... args) {
        dataArray[nb++] = T(std::forward<Args>(args)...);
    }
```

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <array>
+#include <stdexcept>
+
+template <class T, unsigned int N>
+class TinyVector {
+public:
+    TinyVector() = default;
+
+    using size_type = unsigned int;
```
Febbe: **Bugprone:**
```suggestion
    using size_type = std::array<T, N>::size_type;
```

src/util/include/util/TinyVector.h
```diff
@@ -0,0 +1,68 @@
...
+        if (nb < N) {
+            dataArray[nb++] = T(std::forward<Args>(args)...);
+        } else {
+            throw std::length_error("TinyVector overflow");
+        }
+    }
+    void clear() { nb = 0; }
+
+private:
+    std::array<T, N> dataArray{};
```
Febbe, line 57: https://en.cppreference.com/w/cpp/types/aligned_storage

src/util/include/util/SmallVector.h
```diff
@@ -0,0 +1,115 @@
...
+    void clear() {
+        nb = 0;
+        /**
+         * Reset capacity. If clear() is called a lot and more than N objects are usually pushed, don't use this class!
+         */
+        dataVector = {};
+    }
+
+private:
+    std::array<T, N> dataArray{};
```
Febbe, line 96: https://en.cppreference.com/w/cpp/types/aligned_storage

I would use TinyVector here. And fix that in Tiny Vector

src/util/include/util/Interval.h
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * @brief Computes the convex envelop of this and another value
+     * @param t The other value
+     */
+    [[maybe_unused]] void envelop(const T& t) {
+        if (t >= max) {
+            max = t;
+        } else {
+            min = std::min(min, t);
+        }
```
Febbe, line 66: Easier to read:
```suggestion
        max = std::max(max, t);
        min = std::min(min, t);
```

src/core/model/eraser/ErasableStroke.cpp
```diff
@@ -4,300 +4,386 @@
...
-        }
+    // Need to compute the bounding box
+    Point p = this->stroke.getPoint(section.min);
+    double minX = p.x;
+    double maxX = p.x;
+    double minY = p.y;
+    double maxY = p.y;
+
+    auto data = this->stroke.getPointVector();
+    auto endIt = std::next(data.cbegin(), (std::ptrdiff_t)section.max.index + 1);
```
Febbe, line 297: Why + 1 and not minus 1.

src/core/model/eraser/ErasableStroke.cpp
```diff
@@ -4,300 +4,386 @@
...
+        maxX = std::max(maxX, ptIt->x);
+        minY = std::min(minY, ptIt->y);
+        maxY = std::max(maxY, ptIt->y);
     }
 
+    Point q = this->stroke.getPoint(section.max);
+    minX = std::min(minX, q.x);
+    maxX = std::max(maxX, q.x);
+    minY = std::min(minY, q.y);
+    maxY = std::max(maxY, q.y);
```
Febbe, line 309: Isn't the point already computed in the for loop?

src/core/model/eraser/ErasableStroke.cpp
```diff
@@ -4,300 +4,386 @@
...
-            changed = true;
-        } else {
-            splitPoints.back().push_back(*pointIter);
-        }
-    }
-    if (splitPoints.back().empty()) {
-        splitPoints.pop_back();
-    }
+    const double width = maxX - minX + halfStrokeWidth;
+    const double height = maxY - minY + halfStrokeWidth;
```
Febbe, line 320:
```
    const double stroke_width = this->stroke.getWidth();
    const double width = maxX - minX + stroke_width
    const double height = maxY - minY + stroke_width;
```

src/core/model/eraser/ErasableStrokeOverlapTree.cpp
```diff
@@ -0,0 +1,239 @@
...
+        this->maxX = p2.x;
+    }
+
+    if (p1.y > p2.y) {
+        this->minY = p2.y;
+        this->maxY = p1.y;
+    } else {
+        this->minY = p1.y;
+        this->maxY = p2.y;
+    }
```
Febbe, line 170:
```suggestion
    std::tie(this->minX, this->maxX) = std::minmax(p1.x, p2.x);
    std::tie(this->minY, this->maxY) = std::minmax(p1.y, p2.y);
```

Febbe:
Got a crash, after removing random filled strokes: grafik

bhennion:
> Got a crash, after removing random filled strokes: grafik

Hmm... Not great.
I think there could be some very rare numerical instability occurences that could trigger this. I never witnessed it though. Maybe it reacts differently on Windows. Let me try to build a failing test case.

Does this happens regularly for you? If so, I'll make a debug version with relevant output to try and determine the source of the issue.

src/core/model/eraser/ErasableStroke.cpp
bhennion: `std::next(data.cbegin(), (std::ptrdiff_t)section.max.index)` points to `data[section.max.index]` which is the penultimate point in the section. The ultimate point `this->stroke.getPoint(section.max);` is between `data[section.max.index]` and `data[section.max.index + 1]` (more precisely, `data[section.max.index].relativeLineTo(data[section.max.index], section.max.t)`).

So this +1 is correct.

src/core/model/eraser/ErasableStroke.cpp
bhennion: See above.

src/core/model/eraser/ErasableStroke.cpp
bhennion: Done.

src/core/model/eraser/ErasableStrokeOverlapTree.cpp
bhennion: Done

src/util/include/util/Interval.h
bhennion: Done

src/util/include/util/TinyVector.h
bhennion: I'm working on that.
A question: do we not care `aligned_storage` being deprecated? (It's only for C++23 and we are not yet using C++20, so we could say we have time and we'll see then :-)

bhennion:
> > Got a crash, after removing random filled strokes: grafik
> 
> Hmm... Not great. I think there could be some very rare numerical instability occurences that could trigger this. I never witnessed it though. Maybe it reacts differently on Windows. Let me try to build a failing test case.
> 
> Does this happens regularly for you? If so, I'll make a debug version with relevant output to try and determine the source of the issue.

I fixed a possible source of this assert failing (and added a test case :-). Not convinced this case is what happened for you though. This seem way to improbable for you to experience it (like, ever).

src/util/include/util/TinyVector.h
bhennion: > std::array accesses are tested in debug mode, so there is no reason to waste computation power here:

I'm not sure of what you mean here. Without this test, I just get a SegFault, nothing more (with CMAKE_BUILD_TYPE = Debug).

src/util/include/util/TinyVector.h
bhennion: Can I actually do both using `std::array::iterator` and using `std::aligned_storage`? If so, I don't know how.

Febbe: Right, that's not possible. You'll have to write your own then.

Febbe:
> > > Got a crash, after removing random filled strokes: grafik
> > 
> > 
> > Hmm... Not great. I think there could be some very rare numerical instability occurences that could trigger this. I never witnessed it though. Maybe it reacts differently on Windows. Let me try to build a failing test case.
> > Does this happens regularly for you? If so, I'll make a debug version with relevant output to try and determine the source of the issue.
> 
> I fixed a possible source of this assert failing (and added a test case :-). Not convinced this case is what happened for you though. This seem way to improbable for you to experience it (like, ever).

It still happens, this time I were in debug mode.
The Stoke had 2 Points.

```
inner_box
297.72931374773287
455.78315310140147

outer_box
296.82531374773288
454.87915310140147

p0
296.49764914748346
455.58475059851605

p1
296.82531374773288
455.26527761327287
```

bhennion:
Ha! I can reproduce now! I introduced a bug when remodeling `Stroke::intersectWithPaddedBox` to make more readable. Let me try and find exactly what's happening

bhennion:
> Ha! I can reproduce now! I introduced a bug when remodeling `Stroke::intersectWithPaddedBox` to make more readable. Let me try and find exactly what's happening

Actually, what I see happen (very rarely) on my laptop is a numerical error making the assert fail. The fix is to simply enlarge the assert's box by  `1e-10`. 
I also added a lot of debug output in case it still happens and there was another issue.

Febbe:
Even if the assert will not fail, there is still a bug. That branch whére the assert fails requires the stroke to be longer than 2 points.

bhennion:
I refactored TinyVector and SmallVector to use std::aligned_storage and created suitable iterators util/include/util/BasePointerIterator.h. 
Those classes can be rereviewed.

bhennion:
> I could make a view (iterator adaptor) to make clear that the loop and indices are thought for segment-based iteration.

This is done. It shoudl clarify this index business.

Note also that the (temporary) debug info printed out in the console introduces a lag when erasing long strokes. The lag disapears when removing the debug info (by commenting out the `#define ENABLE_ERASER_DEBUG` in Stroke.cpp:L533

Febbe:
I skimmed through the code, and it looks good so far. I'll test it, and please also test it again.
When everything is fine, please squash all your commits into one and then rebase it onto the current master.
When all pipelines work, you or other members of @xournalpp/core can merge it :). Thank you for the hard work.

Febbe:
Also, great that you created so many tests.

Febbe:
Had again a floating point missmatch error: 
grafik

rolandlo:
> When everything is fine, please squash all your commits into one and then rebase it onto the current master.

That sounds like a lot of stuff going into one single commit. I think history can be rewritten better than that.

bhennion:
> Had again a floating point missmatch error

Damn! Do you happen to have the terminal output corresponding to this crash? There should be debug info there that would help me figure out what the f is happening.
I assume you were not doing anything more than erasing random strokes, right? I can't seem to reproduce...

bhennion:
> > When everything is fine, please squash all your commits into one and then rebase it onto the current master.
> 
> That sounds like a lot of stuff going into one single commit. I think history can be rewritten better than that.

I agree. I'll see what I can do about this once I fixed the crash @Febbe mentioned.

Febbe:
replaced the assert with:

```
        assert([&]() -> bool {
            auto ret = lastPoint.isInside(Rectangle<double>(outerBox.x - 1e-10, outerBox.y - 1e-10,
                                                            outerBox.width + 2e-10, outerBox.height + 2e-10));
            if (ret)
                return true;
            std::cout << "| Last point is not in outerbox (with a 1e-10 margin)\n";
            std::cout << "| * innerBox = (";
            std::cout << std::setprecision(5);
            std::cout << std::setw(9) << innerBox.x << " ; ";
            std::cout << std::setw(9) << innerBox.x + innerBox.width << ") -- (";
            std::cout << std::setw(9) << innerBox.y << " ; ";
            std::cout << std::setw(9) << innerBox.y + innerBox.height << ")\n";
            std::cout << "| * outerBox = (";
            std::cout << std::setw(9) << outerBox.x << " ; ";
            std::cout << std::setw(9) << outerBox.x + outerBox.width << ") -- (";
            std::cout << std::setw(9) << outerBox.y << " ; ";
            std::cout << std::setw(9) << outerBox.y + outerBox.height << ")\n";
            std::cout << "| * lastPoint = (";
            std::cout << std::setw(9) << lastPoint.x << " ; ";
            std::cout << std::setw(9) << lastPoint.y << ")\n";
            std::cout << std::flush;
            return false;
        }());
```

And got something like this:
``` text
| Last point is not in outerbox (with a 1e-10 margin)
| * innerBox = (   431.33 ;    448.33) -- (   222.25 ;    239.25)
| * outerBox = (   430.43 ;    449.24) -- (   221.35 ;    240.16)
| * lastPoint = (   393.99 ;    247.76)
```
innerbox und outerbox overlap somehow and the rectangles are not normalized, arent they?

Your output:

```
Stroke::intersectWithPaddedBox debug:
| * flags.isInsideOuter              = true
| * flags.wentInsideInner            = true
| * flags.lastSegmentEndedOnBoundary = false
| * Segment   0: (430.42824 ; 237.88888) -- (428.64414 ; 238.44641)
|  |__** result.size() =   1
| * Segment   1: (428.64414 ; 238.44641) -- (419.55381 ; 241.24344)
|  |__** result.size() =   1
| * Segment   2: (419.55381 ; 241.24344) -- (408.36570 ; 244.73972)
|  |__** result.size() =   1
| * Segment   3: (408.36570 ; 244.73972) -- (395.77909 ; 247.53675)
|  |__** result.size() =   1
| * Segment   4: (395.77909 ; 247.53675) -- (393.99499 ; 247.75976)
|  |__** result.size() =   1
```

Febbe:
> 

Yep, obviously rewriting the history, would be the cleanest solution, but either way, a squash from all is a good start for the rewrite.

bhennion:
The good news is: I know what's happening now!
This is the same numerical instability I fixed for the last point of the stroke, but happening to the first point. You can see in both your crashes, the stroke's first point is on the boundary of the `outerBox` (at least up to visible digits):
```
firstPoint.y == 756.31795 == outerBox.y
```

There could be other cases where numerical instability messes things up, I guess. I think I'll just add a sanity check at the end (instead of the assert). If the list of intersection points we get is not consistent, we can just drop it all and not erase anything at this iteration.
At the next iteration, the eraser box will have moved a tiny bit, most likely making the error disappear. Whatever as under the eraser box will get erased then. The user won't see the difference.

bhennion:
Rebased on master and cleaned up the commit tree. I kept in separate commits the preliminary work.

bhennion:
Alright, I implemented a simple sanity check  to account for unforeseen numerical imprecisions that would trigger a crash (or more likely the unwanted erasure of some bits of the stroke).

bhennion:
Alright: I fixed the sanity check. Now, whenever something goes wrong when computing the intersection points of the stroke with the eraser box (most likely some rounding error...), we disregard the results. During the next iteration, the eraser box will have moved even so slightly, and the rounding errors will no longer be an issue.
This all happens in 1/100s, so we can't see a thing.

I also fixed the newly add various stroke caps, that were mishandled before.

I think this PR should wait for #3503 to be merged. I'd rather rebase this one on top of #3503 than the other way around.

bhennion:
Btw, and OT: it is quite weird that the PdfToolBox strokes are all filled highlighter. They should be either a simple highligher, or a normal stroke. This has some computational cost when rendering...

bhennion:
I rebased on master, now that #3503 has been merged.
@Febbe do you wanna give this PR a last look over?
@Febbe @rolandlo Are you happy with the commit tree as it is now?

rolandlo:
@rolandlo Are you happy with the commit tree as it is now?

The commit tree looks pretty nice now. The last commit is still quite big, but other than maybe splitting off the unit tests and the Readme off into a separate commit, it would probably not be easy to split it up.

By the way, about the Readme: For the Setsquare I put the Readme into the folder `development/documentation`. You put the Readme for the eraser directly into the eraser folder. I don't know which way is preferable (I can see pros and cons for both), but maybe we better try to be consistent.

The code looks really pretty clean and performance-wise it's a giant improvement. I really admire the work you did here. Great job!

bhennion:
> The last commit is still quite big, but other than maybe splitting off the unit tests and the Readme off into a separate commit,

I splitted the commit as suggested

> By the way, about the Readme: For the Setsquare I put the Readme into the folder `development/documentation`. You put the Readme for the eraser directly into the eraser folder. I don't know which way is preferable (I can see pros and cons for both), but maybe we better try to be consistent.

That's a good point. I moved the Readme to `development/documentation`, it feels cleaner than to leave it in the source tree. Not sure any future developer will see it though.

> The code looks really pretty clean and performance-wise it's a giant improvement. I really admire the work you did here. Great job!

Thanks a lot!!

Febbe:
LGTM :)

bhennion:
Great! Thanks again for the reviewing!

PR #3564, opened 2021-11-12
Improve unit test build time

idotobi:
This PR makes the `src/util` directory a sink in the include graph (see below for details) and splits up the unit-tests per project "component" (i.e. `util`, `model`, `conrtol`).

This enables to only build `src/util` when building `unit-tests-util`, avoiding unnecessary recompilation 🎉. Especially for utils this improves the build-test cycle substantially as changes in utils usually cause most of the project to be rebuilt

E.g.: `Rectangle.h` has the following dependencies throughout the project

Rectangle_8h__dep__incl

## Details

Before the Refactor the dependency graph looked like this. The red arrows basically broke the assumption that utils should only be referred to from the rest of the code but not the other way around:

2021-11-12_11-19_util_deps

To do this some files were moved to other directories in `src` where they seemed to fit semantically.

Now it looks like:

2021-11-12_12-04_dep_graph

2021-11-12_12-04_dep_graph_big


idotobi:
Note: The new location of the files I moved out of `src/util` was chosen based on my limited knowledge and are up for discussion if you disagree.

idotobi:
@Febbe : I think this PR might be interesting to you (in light of #2570). If I understand #2570 correctly you did something similar. I would propose that #2570 goes first and then I'll  try to resolve the merge conflicts in here.

Technius:
This has been obsoleted by the directory structure refactor. `util` no longer has a circular dependency.

Technius:
Closing for now; feel free to open a new PR if there's still a problem.

PR #3476, opened 2021-10-14
Migrate from CPPUnit to gtest

idotobi:
- include gtest to cmake files and remove cppunit
 - include gtest in azure-pipelines and remove cppunit
 - move tests from test/ to gtest/unit_tests/ and adapt to gtest
 - add two testcases from XojPreviewExtractorTest that were previously
   skipped and fix them

Resolves #3439

PR #3576, opened 2021-11-16
Prevent installation of Google test

### LINKED ISSUE #3561: Installing package after running tests includes GoogleTest files
**Affects versions :**
 - OS: Only tested on Linux (Arch)
 - Version of Xournal++: 269167da8a0a7a9c3322157220cb8676ca0fefaa or newer
 - Installation method: Build from source

**Describe the bug**
Installing after running tests causes gtest and gmock files to also be installed, such as `/usr/include/gmock/*`, `/usr/include/gtest/*`, `/usr/lib/cmake/GTest/*`, etc.

**To Reproduce**
1. Build xournalpp with `cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..; cmake --build .`
2. Run tests with `make test-units; make test`
3. Install with `make install`

**Expected behavior**
GoogleTest files should not be included in order to not conflict with system files.

EDIT: oops just noticed I still had the flag for `CPPUNIT` instead of `GTEST`. Guess I copied from the old script before I tried patching it for `GTEST`.

idotobi:
Fixes #3561 

 - move `FetchContent`, ... to `./test/CMakeLists.txt`
 - `add_subdirectory` for tests with `EXCLUDE_FROM_ALL`
 - set option `INSTALL_GTEST` to `OFF`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,19 +148,10 @@ add_includes_ldflags ("${SNDFILE_LDFLAGS}" "${SNDFILE_INCLUDE_DIRS}")
 ## Additional features ##
 # Testing Framework: gtest
 option (ENABLE_GTEST "Enable gtest build for xournalpp application" OFF)
-if (ENABLE_GTEST)
-    include(FetchContent)
-    FetchContent_Declare(
-        googletest
-        URL https://github.com/google/googletest/archive/refs/tags/release-1.11.0.zip
-    )
-	# Prevent reloading if already downloaed
-	set(FETCHCONTENT_UPDATES_DISCONNECTED ON)
-    # For Windows: Prevent overriding the parent project's compiler/linker settings
-    set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-    FetchContent_MakeAvailable(googletest)
+if(ENABLE_GTEST)
 	enable_testing()
-endif (ENABLE_GTEST)
+endif()
+
 
 # Plugins / scripting
 find_package (Lua 5.3 EXACT)
@@ -269,6 +260,9 @@ add_subdirectory (src)
 
 ## Man page generation ##
 add_subdirectory (man)
+
+## Tests ##
+# ./test is added via src to enable cleaner includes
 
 ## Final targets and installing ##
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -139,5 +139,5 @@ install (TARGETS xournalpp
 )
 
 if (ENABLE_GTEST)
-  add_subdirectory (${CMAKE_SOURCE_DIR}/test ${CMAKE_BINARY_DIR}/test)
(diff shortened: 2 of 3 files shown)


idotobi:
This should be fixed with this. I repeated the commands from https://github.com/xournalpp/xournalpp/issues/3561#issue-1051299123 locally and no `gtest`/`gmock` files are installed.

The failing Windows CI comes from www.portaudio.com being down. I'm not sure if we should wait on this for merging the fix or not.

MithicSpirit:
Can confirm this fixes #3561 on Arch. Tested with docker using the following:

```sh
#!/usr/bin/env sh

sudo pacman --noconfirm -Syu git cmake texlive-bin gtk3 poppler-glib libxml2 portaudio libsndfile lua libzip

echo "LS-ing HERE =================================="
ls /usr/include/gmock
ls /usr/include/gtest
ls /usr/lib/cmake/GTest

echo "BUILDING HERE ================================"

# Current master branch
#git clone 'https://github.com/xournalpp/xournalpp.git'

# PR #3576
git clone 'https://github.com/idotobi/xournalpp.git'
git checkout origin/fix/3561-prevent-install-of-gtest

mkdir xournalpp/build
cd xournalpp/build
cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..; cmake --build .
make test-units; make test
echo "INSTALLING HERE =============================="
make install

echo "LS-ing HERE =================================="
ls /usr/include/gmock
ls /usr/include/gtest
ls /usr/lib/cmake/GTest
```

In a shell, put the above script in the file `./repro` and make it executable, then run `sudo docker run -it --rm -v "$PWD:/repro" archlinux:base-devel /repro/repro`.

The expected behavior is that the directories `/usr/include/gmock`, `/usr/include/gtest`, and `/usr/lib/cmake/GTest` do not exist in the beginning or the end, which happens when using the `PR #3576` section, but not when using `Current master branch` (where in the end the directories do exist). I believe that there may be other directories that gtest may attempt to install to, but considering that these three directories worked I believe that it is reasonable to assume that this PR is working.

idotobi:
Last force push, removed some unnecessary diffs (empty lines).

idotobi:
@xournalpp/core : As I can tell this is ready for merging. If someone could review, it would be great, as this bug could negatively affect some people following master.

personalizedrefrigerator:
I haven't tried building with this PR, but from a quick look through the changes, it looks good to me.

The indentation in CMaleLists.txt files is inconsistent right now (tabs vs spaces vs fewer spaces). A follow-up PR can fix this.

idotobi:
@bhennion : Yeah there are quite a few inconsistencies (e.g. `if (` vs. `if(`). I would say it's better to fix these formatting issues in a separate PR. That keeps the change set cleaner.

bhennion:
LGTM! I checked the list of installed files before #3476 and after this PR. They match.

PR #3586, opened 2021-11-18
Update page update handling

### LINKED ISSUE #3585: Sidebar page preview stuck and not responding
**Affects versions :**
I guess all versions.

**Describe the bug**
When annotating a pdf with the same amount of pages as the document edited before the sidebar page preview is stuck and does not respond on clicking on page previews.

**To Reproduce**
1. Open a fully new Xournalpp instance
2. Annotate any pdf with exactly one page

fabian-thomas:
- fixed a bug where the first page on the preview wasn't responding on clicking (closes #3585)
- removed dubious locking code from Document initialisation

I need your input on the code I removed from the Document initialisation. I'm gonna comment on what I think the lines were intended to do.


src/model/Document.cpp
```diff
@@ -407,13 +407,6 @@ auto Document::operator=(const Document& doc) -> Document& {
     buildContentsModel();
     updateIndexPageNumbers();
 
-    bool lastLock = tryLock();
```
fabian-thomas, line 410: I guess it is assumed here that tryLock returns false if we already had the lock. But this is not the case. It's actually undefined.

GMutex is neither guaranteed to be recursive nor to be non-recursive, i.e. the return value of g_mutex_trylock() could be both FALSE or TRUE, if the current thread already has locked mutex. Use GStaticRecMutex, if you need recursive mutexes.
https://developer.gimp.org/api/2.0/glib/glib-Threads.html#g-mutex-trylock

src/model/Document.cpp
```diff
@@ -407,13 +407,6 @@ auto Document::operator=(const Document& doc) -> Document& {
...
-    unlock();
```
fabian-thomas, line 411: Here we unlock in any of the two cases.

src/model/Document.cpp
```diff
@@ -407,13 +407,6 @@ auto Document::operator=(const Document& doc) -> Document& {
...
-    this->handler->fireDocumentChanged(DOCUMENT_CHANGE_COMPLETE);
```
fabian-thomas, line 412: I don't understand why we should fire document changed here. It's already called in Control::newFile. And there we for sure know that we have the lock.

src/model/Document.cpp
```diff
@@ -407,13 +407,6 @@ auto Document::operator=(const Document& doc) -> Document& {
...
-    if (!lastLock)  // document was locked before
```
fabian-thomas, line 413: Now we lock again if we had the lock before calling fireDocumentChanged on the assumption that the behavior of tryLock is defined. Since this is not the case this whole code seems really weird. I don't know if removing all of this breaks anything else but in my testing everything has been fine.

src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
```diff
@@ -187,11 +187,6 @@ void SidebarPreviewPages::updatePreviews() {
     doc->lock();
     size_t len = doc->getPageCount();
 
-    if (this->previews.size() == len) {
```
fabian-thomas, line 190: This is the part that resolves the related issue. Comparing the sizes here gives no clue about whether the document is new. We would need to provide that as argument to the listener, hence I removed the whole block.

src/model/Document.cpp
bhennion: It's worst than that. From Gtk doc: 
> Calling `g_mutex_unlock()` on a mutex that is not locked by the current thread leads to undefined behaviour.

So if the previous `trylock` returned false (i.e. the mutex is already locked, we don't know by whom), we're in uncharted territories.

src/model/Document.cpp
bhennion: Agreed. It's also not the model's role the lock itself. The caller should lock the document before calling the operator.
Please make sure that this is the case anywhere the operator is used.

bhennion (commented):
Thanks for the PR! LGTM. Please check that the document is always locked by the caller when the `operator=` is called, just to make sure.

fabian-thomas:
> Thanks for the PR! LGTM. Please check that the document is always locked by the caller when the `operator=` is called, just to make sure.

Do you know how I can do something like assert(thread_has_lock(gmutex))?

Overwriting the = operator seems to be really bad style to me. It took me a bit to find out about that definiton while debugging. Is there any benefit by doing that?

fabian-thomas:
I've found a lot more code that uses tryLock and some other dubious lines in the Document class. I'm gonna look into that once I know how to assert for having the lock.

bhennion:
> Do you know how I can do something like assert(thread_has_lock(gmutex))?

I don't think you can.
Also, and I should have mentionned it earlier, a PR #3382 that changed every `g_mutex` into `std::mutex` was merged in the release-1.1 branch. It has not yet been cherry-picked to master, so it will cause some merge conflicts once it is.
@xournalpp/core Maybe we could do this cherry-picking in advance?

> Overwriting the = operator seems to be really bad style to me. It took me a bit to find out about that definiton while debugging. Is there any benefit by doing that?

There are pros and cons. In this case, more cons than pros, I think. The operator is used only in `Control.cpp` (3 times) and every time it plays the role of a `std::swap` rather than of an assignment. I think it'd be reasonable to make it `Document::swap` and actually swap everything in it. We'd save some unnecessary copy operations and make things clearer, I think.

bhennion:
> Also, and I should have mentionned it earlier, a PR #3382 that changed every `g_mutex` into `std::mutex` was merged in the release-1.1 branch. It has not yet been cherry-picked to master, so it will cause some merge conflicts once it is. @xournalpp/core Maybe we could do this cherry-picking in advance?

For the moment, you could cherry-pick c8d690c2d at the root of your PR to work with those changes already.

Febbe:
Why not merging release-1.1 into master?

fabian-thomas:
I decided to move the actual bug fix to another PR since this one will get big.

bhennion:
> Why not merging release-1.1 into master?

Yes, of course, merging is better!

fabian-thomas:
> > Why not merging release-1.1 into master?
> 
> Yes, of course, merging is better!

I've done it a bit different. I've cherry-picked my commits to release-1.1.

LittleHuba:
That is really not what you want. This PR can never be merged into a minor release as it has a way too high potential to break something. This should be seen more like a new feature than a bugfix and is therefore best merged into `master`.

Febbe:
You could write some enclosing guard named Document with the following function:

template <class Fn>
void doTask(Fn&& fn){
    std::lock_guard guard{mutex};
    fn(internalDocument);
}

With the private member internalDocument, which contains all the fields in the doc.

LittleHuba:
We could actually overload the `->` operator for the wrapper class. That is a common way to do something like that.

fabian-thomas:
> That is really not what you want. This PR can never be merged into a minor release as it has a way too high potential to break something. This should be seen more like a new feature than a bugfix and is therefore best merged into `master`.

Oh, yes. So how do we do that?

I base this PR on master and merge release-1.1. And then we merge this PR when we merge release-1.1 onto master?

src/model/Document.cpp
Febbe: > It's also not the model's role the lock itself

It's actually the other way round. Optimally, lockable's should not be accessible from the public.

fabian-thomas: I agree on that. Document should probably be implemented as monitor. I guess this is what @LittleHuba meant:
> We could actually overload the -> operator for the wrapper class. That is a common way to do something like that.

The only problem I see with that is that we might get some overhead if a lot of functions are called on document in a series. Then it would be better to lock the document, do our stuff and then return the lock. We could fix that with a reentrant lock, but this would somewhat break the idea of monitors.

LittleHuba:
No, we will have to merge `release-1.1` in its own PR. If you include it in this PR that is sub-optimal as you will mix a lot of different changes. I'll see if I have the time to do this merge in the next few days. Hopefully, there are not that many conflicts.

Once I merged both branches, you can just fast-forward your branch to the HEAD of `master`. I'll mention you, so you get notified once I merged the branches.

src/model/Document.cpp
LittleHuba: The problem is actually that you can't do multiple actions atomically. A better solution would be a "Monitor" that resembles a RAII-lock. It should enable exclusive access to the document during its lifetime and relinquish the lock automatically during destruction. We should make it reentrant to cover the current use-cases.

Febbe:
@fabian-thomas you can start your work on top of master now.

fabian-thomas:
I've began to look into this. I started by searching for a monitor like implementation that doesn't use some crazy cpp syntax and is easy to understand. What also mattered was that we need some reentrant functionality (locking the document for multiple function calls). Also a reader-writer functionality would be nice to have.

I've settled on the following:
We have one recursive_mutex for the Document class.
In each function in Document we either do a
`std::shared_lock<std::recursive_mutex> lock(mutex);` (function only reads)
or a
`std::unique_lock<std::recursive_mutex> lock(mutex);` (function also writes).
This solves concurrency and reader-writer differentiation.

For the reentrant functionality we just do something like:
```
std::lock_guard<std::recursive_mutex> lock(doc.mutex);
doc.function1();
doc.function2();
```

What do you think about this approach?

fabian-thomas:
Another question:
I've seen a lot of function declarations using the following syntax: `auto function() -> int {`. Does this has any benefit to `int function() {`?

If not I would suggest that we use the latter one since the syntax is easier to understand for people that have no knowledge of cpp (like me).

Febbe:
> Another question: I've seen a lot of function declarations using the following syntax: `auto function() -> int {`. Does this has any benefit to `int function() {`?
> 
> If not I would suggest that we use the latter one since the syntax is easier to understand for people that have no knowledge of cpp (like me).

For normal functions and methods, there is no logical difference, only visual.
You can actually use what you want. Currently, we don't force someone to use either one of these.
But people like me actually have a preference:
Use trailing return types for types longer than 4 for several reasons:
 - It aligns better with other functions.
 - For me, it is better readable, because I don't care about types, I care about flow, so knowing what happens there has a higher preference, than the actual type.
 - For lambdas only trailing return type work, so use the trailing return type for consistency.
 - For some template functions, only trailing return types work.
     - trying to use the old way might drive a new one crazy.
     - would not happen, when using trailing return types
 - trailing return types are more "modern"

I would recommend, to leave functions as they are and choose your preferred style for new functions.
But when we decide to enforce a style, it will be "use trailing return types".

Febbe:
> I've began to look into this. I started by searching for a monitor like implementation that doesn't use some crazy cpp syntax and is easy to understand. 

You can ask your questions here if you have one.

> For the reentrant functionality we just do something like:
> 
> ```
> std::lock_guard<std::recursive_mutex> lock(doc.mutex);
> doc.function1();
> doc.function2();
> ```
> 
> What do you think about this approach?

The document's Mutex must not be accessible. It would be better to implement the lock and try_lock function, but not unlock.
Those functions shall return a guard like: `std::unique_lock` / `std::shared_lock`.

Using a recursive lock has a smell, but it's ok, to fix a deadlock (when a better aproach is e.g. not feasible because a large amount of code has to be refactored).

That's why an observer based approach would be better, but your approach with my change request is way better than the current situation.

fabian-thomas:
> Using a recursive lock has a smell, but it's ok, to fix a deadlock (when a better aproach is e.g. not feasible because a large amount of code has to be refactored).

What is wrong with recursive locks, apart from their minimal overhead?

> That's why an observer based approach would be better, but your approach with my change request is way better than the current situation.

Could you elaborate that observer approach further? I guess we should go for the best pattern, since we'll already refactor a lot of code.

src/core/model/Document.cpp
```diff
@@ -148,10 +190,24 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
...
 
+/*
+ * TODO
+ *
+ * Read operation.
+ */
+cairo_surface_t* Document::getPreview() {
+    std::shared_lock<std::recursive_mutex> lock(mutex);
+    return this->preview;
+}
```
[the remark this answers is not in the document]
fabian-thomas: Yeah, I've rewritten it a little bit to make sure that I will take a look at this later.
In the best case we would write a wrapper for this cairo_surface that holds the document lock until it itself is freed.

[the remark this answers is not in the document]
fabian-thomas: But I think this is not feasible. There are some other classes that would need this treatment too. Does somebody know another way to do this?

Febbe:
> What is wrong with recursive locks, apart from their minimal overhead?

It's a sign of bad design: It basically means, that none knows, and can say, which thread has when and where access to the data.
But this is actually something which can be controlled when the access patterns are well-defined.
It also has an overhead.

The best solution, would be, to have the `class Document;` itself, without any Lockable's, a `class DocumentHandle;`, owning a (shared)reference to a Document and a Lockable. The `class DocumentHandle;` implements all functions of the Document with a lock_guard. It also implements a template function, which accepts a Function, acquires a lock_guard, and executes that function with a reference to the real Document.

fabian-thomas:
> > What is wrong with recursive locks, apart from their minimal overhead?
> 
> It's a sign of bad design: It basically means, that none knows, and can say, which thread has when and where access to the data. But this is actually something which can be controlled when the access patterns are well-defined. It also has an overhead.
> 
> The best solution, would be, to have the `class Document;` itself, without any Lockable's, a `class DocumentHandle;`, owning a (shared)reference to a Document and a Lockable. The `class DocumentHandle;` implements all functions of the Document with a lock_guard. It also implements a template function, which accepts a Function, acquires a lock_guard, and executes that function with a reference to the real Document.

But that would require a lot of redundant definitions of each function. In DocumentHandle header and source + Document header and source.
How does this pattern solve reentrance?

Febbe:
> But that would require a lot of redundant definitions of each function. In DocumentHandle header and source + Document header and source.

Yep, that's a solution which is easy, understandable and explicit.
Another approach would be something like a 
```
template<class std::shared_ptr<std::pair<T, std::mutex>>> WriteHandle`, which overloads the `auto operator->() -> T*`.
The Handle releases the lock on destruction.

> How does this pattern solve reentrance?

The document is itself not accessible from outside, it is also not lockable. But a function which is transferred to the DocumentHandle will retrieve a reference to the Document. 
So, inside the function, the Document can be modified without acquiring a lock.
This is safe, because the function can only be executed by the Document Handle, which will acquire the lock.

**Comment by fabian-thomas** [comment]
This will take a while to complete...
After experimenting some more with different monitor implementations I found the one with overloading to be the best.

The most interesting class to review by now is therefore the Monitor one. Pls provide feedback on that one (as stated in the class I've copied the most interesting code from stackoverflow).

For the process of refactoring I need some more information on what is working concurrently on the Document. Currently all calls to the Document will be synchronized. If it would turn out that only some fields in the Document need to be synchronized this would add a big(???) overhead.

If there are no bigger flaws with my work until now I would continue refactoring from time to time to get to a point where I can build the application again.

File: src/core/model/Monitor.h [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+            Monitor* mon = nullptr;
+            std::unique_lock<std::mutex> lock;
+    };
+
+
+    LockedMonitor operator->();
+
+    LockedMonitor lock();
+
+    TryLockedMonitor tryLock();
```
Febbe, line 37: I think an optional will force the accessory to check / explicitly dereference the optional.
Also, `try_lock` should be written in snake case for convenience (c++ STL does so). 

```suggestion
    std::optional<LockedMonitor> try_lock();
```

src/core/model/Monitor.h
```diff
@@ -0,0 +1,44 @@
...
+
+    T& getUnsafeAccess();
```
Febbe, line 39: We actually don't want such a function, since access should be guarded. But since it helps while refactoring, mark it `[[deprecated]]`:
```suggestion
    [[deprecated]] T& getUnsafeAccess();
```

src/core/model/Monitor.h
```diff
@@ -0,0 +1,44 @@
...
+
+private:
+    T model;
+    std::mutex mutex;
```
Febbe, line 43: We want a read and write access, to reduce contention.
```suggestion
    std::shared_mutex mutex;
```

src/core/model/Monitor.h
```diff
@@ -0,0 +1,44 @@
...
+        void ReplaceModel(T model);
+        bool lockAcquired;
+        private:
+            Monitor* mon = nullptr;
+            std::unique_lock<std::mutex> lock;
+    };
+
+
+    LockedMonitor operator->();
+
```
Febbe, line 34: Add 
```
SharedMonitor lock_shared();
std::optional<SharedMonitor> try_lock_shared();
```

src/core/model/Monitor.h
```diff
@@ -0,0 +1,44 @@
...
+    struct TryLockedMonitor
+    {
+        TryLockedMonitor(Monitor* monitor);
+        T* operator->();
+        void ReplaceModel(T model);
+        bool lockAcquired;
+        private:
+            Monitor* mon = nullptr;
+            std::unique_lock<std::mutex> lock;
+    };
```
Febbe, line 30:
```suggestion
    struct SharedMonitor
    {
        SharedMonitor(Monitor* monitor);
        T const * operator->();
        void ReplaceModel(T model);
        bool lockAcquired;
        private:
            Monitor* mon = nullptr;
            std::shared_lock<std::shared_mutex> lock;
    };
```

src/core/model/Monitor.cpp
```diff
@@ -0,0 +1,57 @@
...
+T* Monitor<T>::TryLockedMonitor::operator->() {
+    assert(lockAcquired);
+    return &mon->model;
+};
+
+template <class T>
+void Monitor<T>::TryLockedMonitor::ReplaceModel(T model) {
+    assert(lockAcquired);
+    mon->model = model;
+}
```
Febbe, line 44: I think, overloading operator== is what we want. Also, for the Monitor class itself.

Febbe (changes requested):

src/core/model/Monitor.cpp
```diff
@@ -0,0 +1,57 @@
...
+auto Monitor<T>::operator->() -> typename Monitor<T>::LockedMonitor { return LockedMonitor(this); }
+
+template <class T>
+auto Monitor<T>::lock() -> typename Monitor<T>::LockedMonitor { return LockedMonitor(this); }
+
+template <class T>
+auto Monitor<T>::tryLock() -> typename Monitor<T>::TryLockedMonitor { return TryLockedMonitor(this); }
+
+template <class T>
+T& Monitor<T>::getUnsafeAccess() { return model; }
```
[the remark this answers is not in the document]
fabian-thomas: This actually works (it compiles for me) but is really annoying to read. Should I just remove the header and move everything to the source file again?

src/core/model/Monitor.h
fabian-thomas: Do we really need this differentiation? This will add a lot of complexity.

When we add something like that the caller is again responsible for only doing read operations while holding the shared monitor. We have no way of forcing a uniquely locked monitor for some function calls to Document.

Febbe: When the `SharedMonitor` only returns a `Document const*` the caller can't access non const functions.

bhennion:
@fabian-thomas What is the status of this PR?

Making the document thread-safe(r) seems like an important thing to do, so merging a PR along those lines would be great!

fabian-thomas:
I absolutely agree. Currently I have no time for working on this. Additionnaly, I think that it might be better if someone who has a better overview of the whole codebase would do this.

I'm gonna try to finish and merge the other PRs that I've currently open in the next few weeks.

PR #3590, opened 2021-11-19
Detect .PDF as pdf

### LINKED ISSUE #3548: Unable to open PDF with capitalized extension "foo.PDF"
**Affects versions :**
 - OS: Linux openSuse Tumbleweed
 - (Linux only) Desktop environment: KDE X11
 - Version of Xournal++: 1.1.0

**Describe the bug**
Cannot open an otherwise perfectly working PDF.
This PDF (attached) can be opened without problem with Okular and with Master PDF Editor 5.
Opening with Xournal++ I get an xmlparser error (see screenshot below).

**To Reproduce**
1. Open the attached PDF file with Xournal++

**Expected behavior**
The file should be opened and shown by Xournal++

**Screenshots of Problem**
image

[Dichiarazione sostitutiva certificato di morte pdf_9 T.PDF](https://github.com/xournalpp/xournalpp/files/7491392/Dichiarazione.sostitutiva.certificato.di.morte.pdf_9.T.PDF)

fabian-thomas:
Closes #3548

### DIFF
--- a/src/util/PathUtil.cpp
+++ b/src/util/PathUtil.cpp
@@ -71,8 +71,12 @@ auto Util::getEscapedPath(const fs::path& path) -> std::string {
 }
 
 auto Util::hasXournalFileExt(const fs::path& path) -> bool {
-    auto extension = path.extension();
+    auto extension = StringUtils::toLowerCase(path.extension().string());
     return extension == ".xoj" || extension == ".xopp";
+}
+
+auto Util::hasPdfFileExt(const fs::path& path) -> bool {
+    return StringUtils::toLowerCase(path.extension().string()) == ".pdf";
 }
 
 auto Util::clearExtensions(fs::path& path, const std::string& ext) -> void {
--- a/src/control/Control.cpp
+++ b/src/control/Control.cpp
@@ -1983,7 +1983,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         return loadXoptTemplate(filepath);
     }
 
-    if (filepath.extension() == ".pdf") {
+    if (Util::hasPdfFileExt(filepath)) {
         return loadPdf(filepath, scrollToPage);
     }
 
--- a/src/control/RecentManager.cpp
+++ b/src/control/RecentManager.cpp
@@ -51,7 +51,7 @@ void RecentManager::addRecentFileFilename(const fs::path& filepath) {
     std::array<gchar*, 2> groups = {group_name.data(), nullptr};
     std::string app_name = g_get_application_name();
     std::string app_exec = std::string(g_get_prgname()) + " %u";
-    std::string mime_type = (filepath.extension() == ".pdf") ? std::string(MIME_PDF) : std::string(MIME);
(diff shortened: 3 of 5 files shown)


src/util/PathUtil.cpp
```diff
@@ -72,7 +73,14 @@ auto Util::hasXournalFileExt(const fs::path& path) -> bool {
     return extension == ".xoj" || extension == ".xopp";
 }
 
-auto Util::clearExtensions(fs::path& path, const std::string& ext) -> void {
+auto Util::hasPdfFileExt(const fs::path& path) -> bool {
+    return std::find(Util::PDF_EXTENSIONS.begin(), Util::PDF_EXTENSIONS.end(), path.extension()) !=
+           Util::PDF_EXTENSIONS.end();
+}
```
bhennion, line 79: Would it not be simpler to just add a `StringUtils::toLowerCase` as is done in the lambda `rm_ext` below? This would also cover all the cases like `.Pdf` an so on.

src/util/PathUtil.cpp
```diff
@@ -72,7 +73,14 @@ auto Util::hasXournalFileExt(const fs::path& path) -> bool {
...
+
+auto Util::clearExtensions(fs::path& path) -> void { clearExtensions(path, std::vector<std::string>()); }
+
+auto Util::clearExtensions(fs::path& path, const std::vector<std::string>& extensions) -> void {
     auto rm_ext = [&path](const std::string ext) {
         if (StringUtils::toLowerCase(path.extension().string()) == StringUtils::toLowerCase(ext)) {
             path.replace_extension("");
```
bhennion, line 86: See here for the lambda `rm_ext`

src/util/PathUtil.cpp
fabian-thomas: Yeah, I thought about that before implementing this with a list of pdf extensions. At that point I thought it would be cleaner to do it like I did. I'm gonna change that.

personalizedrefrigerator (commented):
I haven't tested this yet, but from a quick look through the changes, this looks good to me!

LittleHuba:
@fabian-thomas please rebase to `release-1.1`. See https://github.com/xournalpp/xournalpp/blob/master/readme/Releases.md#branching-strategy

fabian-thomas:
That helped a little bit with understanding the different branches and tags. Although I'm not totally sure if I understood that release_helper script correctly. Is this something I should usually use/use in this case?

As I understand this I should just rebase this branch on release-1.1 so that it can be merged into that branch.

LittleHuba:
No that script is solely for us maintainers. You should never have to use it.

LittleHuba:
> As I understand this I should just rebase this branch on release-1.1 so that it can be merged into that branch.

Yes exactly.

fabian-thomas:
@LittleHuba Have I done this correctly? I decided to cherry-pick the commits since rebasing looked like it would not make sense.

LittleHuba:
Yes cherry-picking is also a valid solution. Your changes seem to be okay. 

Why did you remove the test case though?

fabian-thomas:
I accidently included it while cherry-picking the changes. It came from master. Should I include it again?

Febbe:
Thanks for the PR.

PR #3593, opened 2021-11-20
Fix sidebar page preview stuck and not responding

### LINKED ISSUE #3585: Sidebar page preview stuck and not responding

fabian-thomas:
- fixed a bug where the sidebar page preview was stuck and not responding (closes #3585)

### DIFF
--- a/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
+++ b/src/gui/sidebar/previews/page/SidebarPreviewPages.cpp
@@ -187,11 +187,6 @@ void SidebarPreviewPages::updatePreviews() {
     doc->lock();
     size_t len = doc->getPageCount();
 
-    if (this->previews.size() == len) {
-        doc->unlock();
-        return;
-    }
-
     for (SidebarPreviewBaseEntry* p: this->previews) { delete p; }
     this->previews.clear();
 


Febbe:
This is actually a fix for release1.1, can you cherry-pick this?

fabian-thomas:
> This is actually a fix for release1.1, can you cherry-pick this?

Like that?

Febbe:
Yes like that. 

Actually the code you removed makes a bit sense, no recreation of dynamic objects if the size did not change.
This means your change has some performance implicattions for large files. It enforces a complete recreation of all `SidebarPreviewBaseEntry`s. Isn't it possible to fix them instead of reallocating all? (Using a `std::vector<SidebarPreviewBaseEntry>` instead of std::vector<SidebarPreviewBaseEntry*> may kill the overhead here.)

Also I don't see, why there should be a problem with the mutex behavior. That is correct here (unlock locked mutex on return).

fabian-thomas:
> Yes like that.
> 
> Actually the code you removed makes a bit sense, no recreation of dynamic objects if the size did not change. This means your change has some performance implicattions for large files. It enforces a complete recreation of all `SidebarPreviewBaseEntry`s. Isn't it possible to fix them instead of reallocating all? (Using a `std::vector<SidebarPreviewBaseEntry>` instead of std::vector<SidebarPreviewBaseEntry*> may kill the overhead here.)

Yeah, that's true. But when and how often is this function actually called? I've attached a debugger and tried to trigger updatePreviews. After 2 minutes of testing it was only called when creating a new document.

Could you evaluate your idea with `std::vector<SidebarPreviewBaseEntry>` further?
The only solution I see is that we pass a boolean value to `updatePreviews` which gives information about a full document change. We just don't have that information in that function.

> 
> Also I don't see, why there should be a problem with the mutex behavior. That is correct here (unlock locked mutex on return).

The inner block is totally fine but don't needed anymore if we remove the if condition.

Febbe:
> Could you evaluate your idea with std::vector<SidebarPreviewBaseEntry> further?
Currently, all SidebarPreviewBaseEntry`s may have a random position on the heap and must be freed, allocated every run.
But by dropping the pointer, all SidebarPreviewBaseEntry's are located next to each other, also the space is not allocated on each insertion, since vectors won't shrink. This is then nearly equal to just updating the Elements.
[It also is faster, because it removes one indirection step and increases locality](https://en.wikipedia.org/wiki/Locality_of_reference)

> The inner block is totally fine but don't needed anymore if we remove the if condition.

Ah, thought, that the discussion in your other thread over g_mutex is because of this issue. Never mind then.

> Yeah, that's true. But when and how often is this function actually called? I've attached a debugger and tried to trigger updatePreviews. After 2 minutes of testing it was only called when creating a new document.

Then we should just merge your fix and leave the Pointer removal to your refactoring?

Febbe:
Thank you for your work!

PR #3603, opened 2021-11-22
check for empty tex string and simplify latex template

nicolae-stroncea:
This PR removes the scontents package and greatly simplifies the latex template


rolandlo:
Your PR doesn't seem to work with the `listings` package for code highlighting. See https://xournalpp.github.io/guide/tools/latex/ for how to use it.
That was one of the main points using the `scontents` package. Further analysis is required why it doesn't work though. If I replace `textcolor` by `color` it works.

PR #3610, opened 2021-11-23
Fix #3606: Set stream locale to 'C' when parsing colors

### LINKED ISSUE #3606: Palette migration fails, leading to crash
**Affects versions :**
 - master [1496256](https://github.com/xournalpp/xournalpp/commit/149625635708e5b08eb635a934c4dc7d160726c2)
 - OS: ArchLinux

**Describe the bug**
Trying to run the latest master, I get a popup saying
```
There has been a problem parsing the color palette file at /home/benjamin/.config/xournalpp/palette.gpl

What happened:
The line 4 is malformed.
What to do:
Please fix your palette file, or rename it so xournalpp creates a new default palette file for you. This file can then be used as a template.
What will happen now:
The application will start with the default color palette.
```

The app then crashes.

I certainly never modified the incriminated palette file. Here it is, just in case [palette.gpl](https://github.com/xournalpp/xournalpp/files/7586110/palette.txt)

**To Reproduce**
Launch and see...

**Additional context**
The logs:
[CLI_log.txt](https://github.com/xournalpp/xournalpp/files/7586130/CLI_log.txt)
[errorlog.20211123-072740.log](https://github.com/xournalpp/xournalpp/files/7586133/errorlog.20211123-072740.log)

@idotobi Could you have a look?

bhennion:


src/util/NamedColor.cpp
```diff
@@ -26,9 +26,11 @@ auto operator>>(std::istream& str, NamedColor& namedColor) -> std::istream& {
     NamedColor tmp;
     if (std::getline(str, line)) {
         std::istringstream iss{line};
+        iss.imbue(std::locale::classic());
```
idotobi, line 29: Maybe add a comment on why this is needed as it might be a bit surprising here on first sight.

idotobi (commented):
Could you also a add a testcase for this, which verifies that the default_palette can be loaded even with a french locale. That would basically be a copy of `TEST(ColorPalette, testDefaultLoad)` just with a different locale set.

bhennion:
To be honest, I don't think this is the best solution. I think we should not be setting the global C++ locale at all. Then every std::stream would be initialized with the classic locale `"C"` and those annoying locale related issues would disappear.

idotobi:
That would sound like the better option to me, too.

Does anyone else in @xournalpp/core know why `std::locale::global(std::locale(""))` was set in the first place?

https://github.com/xournalpp/xournalpp/blob/f72fe7c85bf6ac0525350362b946e10df7a9371c/src/core/control/XournalMain.cpp#L75

bhennion:
> Could you also a add a testcase for this, which verifies that the default_palette can be loaded even with a french locale. That would basically be a copy of `TEST(ColorPalette, testDefaultLoad)` just with a different locale set.

This feels weird. Why choose fr_FR and not any other of the hundreds of locales out there? Also, we'd need to make sure the locales are installed on the pipeline. I'm not sure how various OS's would react to an `imbue` with an absent locale as an argument. This may cause more trouble.

IMO, we should merge this PR as is (with the comment @idotobi asked for), and remove the global locale setting in a follow up PR carefully tested.

bhennion:
Merging in 24h provided the tests clear and noone objects.

idotobi:
@bhennion : I added a respective issue (#3611) and assigned you and me, so we don't forget about this.

idotobi (approved):
LGTM :+1:

Febbe:
What about the outstream? It should also imbue the same locale

Febbe:
@idotobi exactly for that purpose, so that all strings in all streams are written and read with the c locale. But this is honestly not what we want. We actually want to use the systems locale for the app and the c locale for serdes.

bhennion:
> > Does anyone else in @xournalpp/core know why std::locale::global(std::locale("")) was set in the first place?
> 
> @idotobi exactly for that purpose, so that all strings in all streams are written and read with the c locale. But this is honestly not what we want. We actually want to use the systems locale for the app and the c locale for serdes.

Except this sets the locale to the system default, not to `C`.

bhennion:
> What about the outstream? It should also imbue the same locale

I don't think there is any. For now, configuration is done via directly editing the file. Xournal++ is never writing in the file, I think. @idotobi, is that right?

Febbe:
> I don't think there is any. For now, configuration is done via directly editing the file. Xournal++ is never writing in the file, I think. @idotobi, is that right? 

Then it LGTM.

Febbe:
> Except this sets the locale to the system default, not to C.

Then it is that what we actually want (until we add a locale editor to the settings, to override the system locale).
All Strings displayed to the user should be formatted in their common way, like (09.11.2021 for Germans and 11/9/21 for American people).

idotobi:
> > What about the outstream? It should also imbue the same locale
> 
> I don't think there is any. For now, configuration is done via directly editing the file. Xournal++ is never writing in the file, I think. @idotobi, is that right?

@bhennion : Actually there is. Namely when writing the default palette.

@Febbe : Thanks for catching this inconsistency!

https://github.com/xournalpp/xournalpp/blob/f72fe7c85bf6ac0525350362b946e10df7a9371c/src/core/gui/toolbarMenubar/model/ColorPalette.cpp#L136-L139

bhennion:
Thanks @idotobi @Febbe. Fixed the ostream locale.

Febbe:
Ready to be merged I think.

PR #3616, opened 2021-11-24
Use uint16_t for PaletteIndex rather than size_t

idotobi:
This fixes a problem on 32bit machines reported at https://github.com/xournalpp/xournalpp/pull/2379#pullrequestreview-814967220
where the two constructors for `NamedColor(const size_t&)` and `NamedColor(const Color&)` are ambiguous as `Color == uint32_t`.

### DIFF
--- a/src/core/gui/toolbarMenubar/model/ColorPalette.h
+++ b/src/core/gui/toolbarMenubar/model/ColorPalette.h
@@ -112,9 +112,9 @@ struct Palette {
     /**
      * @brief Get the number of namedColors part of the palette
      *
-     * @return size_t number of namedColors in palette
+     * @return uint16_t number of namedColors in palette
      */
-    size_t size() const;
+    uint16_t size() const;
 
     /**
      * @brief Get a NamedColor from the palette
@@ -122,7 +122,7 @@ struct Palette {
      * @param i palette index
      * @return NamedColor of palette at palette index
      */
-    NamedColor const& getColorAt(size_t i) const;
+    NamedColor const& getColorAt(uint16_t i) const;
 
 
 private:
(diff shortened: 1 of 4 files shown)


marioalexis84:
@idotobi: Rebased and compiled on 32-bit system without errors.

idotobi:
Merging in 24h unless there are any objections.

src/core/gui/toolbarMenubar/model/ColorPalette.h
```diff
@@ -133,17 +133,17 @@ struct Palette {
     fs::path filepath;
 
     /**
-     * @brief Vector containing all colors of the palette
```
idotobi, line 136: Note: This change fixed a clang-tidy warning in the file, but is otherwise not relevant for the change.

Febbe:
**Minor nitpick**: I would only change the constructor and the caller. 
~~Or: Make Color strong typed and the constructor explicit.~~

  => I'll do that myself, since I know, what I have to change. 
When the latter version works, I would propose to use that.

idotobi:
> **Minor nitpick**: I would only change the constructor and the caller.

Why would that be preferable? Wouldn't this than cause inconsistencies with the getters or at least allow misinterpretation by other programmers on how big paletteIndex can actually be?

> ~Or: Make Color strong typed and the constructor explicit.~
> 
> => I'll do that myself, since I know, what I have to change. When the latter version works, I would propose to use that.

Ok, now I'm confused :sweat_smile: . Should I wait with the merge and you give your solution a shot, or should I merge this fix regardless?

Febbe:
> Ok, now I'm confused sweat_smile . Should I wait with the merge and you give your solution a shot, or should I merge this fix regardless?

Jep, I try to get one ready till 18:00 MET

> Why would that be preferable? Wouldn't this than cause inconsistencies with the getters or at least allow misinterpretation by other programmers on how big paletteIndex can actually be?

I think regarding the change of the type to uint16_t. There is no good solution: Either we have to do many casts (bad) or we have to live with inconsistency (bad) :D.

Issue #3637, opened 2021-12-04
Copied texts are pasted in reversed order


youngminpark2559:
Hi, I'm running xournal++ of 1.1.0+dev built on Dec 1 2021 on Ubuntu 20.04.3 LTS
And I encountered this reversed pasted text problem from copied texts.
Thanks.

https://user-images.githubusercontent.com/30585126/144695560-76bf9953-d216-472b-a1e7-a0a5bdf62962.mp4

bhennion:
I can reproduce. It seems to be a regression. A build from October 24th (83c0c40f) does not have this bug on my machine (Archlinux).
@youngminpark2559 would you have time to bisect and figure out which commit broke this? If you don't know how, I'll do it when I find the time.

youngminpark2559:
@bhennion Umm, sorry, I don't have the ability to technically investigate which commit caused this problem in source code level. Please check this when you have the time.

bhennion:
Alright. The guilty commit is 0156eb6d2 that came in with #3042.
@Marmare314 @Febbe Do you know what's going on?

bhennion:
A fix is available in #3639. I will most likely be merged rapidly.

PR #3639, opened 2021-12-06
Fix 3637 - properly order copied text in clipboard

bhennion:
#3042 introduced #3637 because `g_list` and `std` do not have the same standard of `Compare` functions.
In a `g_list`, A < B if `Compare(A,B) < 0`. With `std`, A < B if `Compare(A, B) == true`.


Febbe (commented):
Oh missed that in my review. Thanks!

**Non-blocking minor nitpick:**
In general, using something like that would be less error-prone:
```
auto isElementLess(Element* a, Element* b) -> bool {
    auto tier = [](Element* a){return std::tie(a->getX(), a->getY();};
    return tier(a) < tier(b);
}
```
You can just merge this as is, if you want.

PR #3641, opened 2021-12-06
Fix #3640 - Change ErasableStroke::PartList from std::vector to std::list

### LINKED ISSUE #3640: Crash using eraser
**Affects versions :**
 - OS: Linux (Ubuntu 21.10)
 - observed on X11
 - Version of Xournal++: 1.1.0+dev (from Dec. 1, 2021)
 - Installation method: PPA

**Describe the bug**
While annotating a PDF the app sometimes (randomly) crashes (with signal 11) when I use the eraser via stylus button. This looks like a regression, as I didn't experience that problem before.

**To Reproduce**
1. Write something onto a PDF
2. Erase using the stylus button
3. Keep doing that until the app crashes

**Expected behavior**
No crash

**Additional context**
Here are two errorlogs: [errorlogs.2021_12_06.zip](https://github.com/xournalpp/xournalpp/files/7659359/errorlogs.2021_12_06.zip)

bhennion:
Fix #3640

### DIFF
--- a/src/core/model/eraser/ErasableStroke.cpp
+++ b/src/core/model/eraser/ErasableStroke.cpp
@@ -233,16 +233,14 @@ auto ErasableStroke::erasePart(double x, double y, double halfEraserSize, PartLi
 
     std::vector<std::vector<Point>> splitPoints{{}};
 
-    for (auto pointIter = points.begin(); pointIter != points.end();) {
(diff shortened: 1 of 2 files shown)


rolandlo (approved):
LGTM. It fixes the crash indeed.

bhennion:
Just cleaned up a couple of other weird iterator operations.

PR #3644, opened 2021-12-06
Fix negative pressure values of strokes

### LINKED ISSUE #3642: Strokes sometimes disappear when recognized as lines by the recognizer
**Affects versions :**
 - OS: Linux
 - Ubuntu 21.10
 - Version of Xournal++: 1.1.0+dev (from December 1, 2021)
 - Installation method: PPA
 
**Describe the bug**
When I draw strokes that are fairly straight and (should) get recognized as lines, those lines sometimes disappear immediately after the pen tip is released.

**To Reproduce**
1. Turn on the shape recognizer.
2. Draw lots of (nearly) lines by hand
3. Observe that some of them disappear right when they should be recognized

**Expected behavior**
The lines should not disappear.

**Additional context**
It is probably a regression, either from #3042 or #3285

### LINKED ISSUE #3643: Negative pressure values in strokes
With latest master, strokes can get negative pressure values. See e.g. the attached file
[2021-12-06-Note-21-40.zip](https://github.com/xournalpp/xournalpp/files/7663470/2021-12-06-Note-21-40.zip)

For now, I only tested with a mouse and pressure inference enabled.
The negative values seem to only appear at the beginning of strokes.

This is a regression. 

(This possibly relates to #3642. See comment there)

bhennion:
Fixes #3643 and fixes #3642

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -313,8 +313,8 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke) -> ShapeRecognizerResult
                 s->addPoint(Point(rs->x2, rs->y2));
             } else {
                 auto points = stroke->getPointVector();
-                s->addPoint(points.front());
-                s->addPoint(points.back());
+                s->addPoint(Point(points.front().x, points.front().y));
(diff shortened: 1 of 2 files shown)


bhennion:
#3643 was due to a mistake when merging the release branch (It's on me. I should have checked this particular point). This fixes that.
#3642 is also fixed along the way.

Merging in 36 hours if nobody complains

PR #3647, opened 2021-12-08
Refactored ErasableStroke

Febbe:


src/core/model/eraser/ErasableStroke.cpp
```diff
@@ -1,14 +1,45 @@
...
+    return points;
+}
 
-#include "ErasableStrokePart.h"
+}  // namespace
 
-ErasableStroke::ErasableStroke(Stroke* stroke): stroke(stroke) {
 
+ErasableStroke::ErasableStroke(Stroke* stroke): stroke(stroke) {
+    // parts.reserve(stroke->getPointCount());
```
bhennion, line 42:
```suggestion
    parts.reserve(stroke->getPointCount() - 1);
```

bhennion (changes requested):
Here are preliminary code comments.

More importantly, it does not work as expected

https://user-images.githubusercontent.com/17818041/145164680-fa421c7d-6ef2-48f2-b5e3-dc30638bb85a.mp4

The bugs I noticed:
1. Line segments (drawn with the line tool) don't get erased in their middle anymore, unless we have already touched one of their endpoints.
2. Line segments are shortened at the end while erasing their beginning.
3. A gap is created when erasing the beginning of a stroke.

Presumably, 2. and 3. are one and the same bug, the gap appearing at the end of the first segment of the stroke.

bhennion:
To be perfectly Candide, I think this PR should just change the container type and make sure the iterators don't stay invalid. The rest is a waste of time and efforts if #3532 is to be merged one day.

Febbe:
@bhennion oh completely forgot about that 😅😓

Febbe:
superseeded

PR #3650, opened 2021-12-08
Background pdf overwriting

### LINKED ISSUE #2363: Allow overwriting background PDF files

fabian-thomas:
Closes #2363

When trying to overwrite the background pdf the user gets a choice window with 3 buttons:
- Cancel
- Overwrite, don't make backup
- Overwrite, but make backup

When the user selects "don't backup" the background pdf will be saved to a tmp file. This should then get deleted once the export is through. Then xournalpp opens the newly created pdf, therefore discarding the stored changes.

When the user selects "make backup" the background pdf is just reloaded and the user can do further edits.

TODO:
- [ ] add a note that modifications are not possible anymore after selecting "no backup"
- [x] delete tmp file
- [x] discard changes when tmp file used
- [x] when using "make backup" it might be annoying for the user that we are still asked if we want to discard the pdf when loading a new one / quitting (solved through faking that we saved the document)
- [ ] make backup extension consistent with other backups
- [ ] further TODOs in code

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1927,7 +1927,7 @@ void Control::showSettings() {
 }
 
 auto Control::newFile(string pageTemplate, fs::path filepath) -> bool {
-    if (!this->close(true)) {
+    if (!this->closeDocumentInteractively(true)) {
         return false;
     }
 
@@ -1977,7 +1977,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         return false;
     }
 
-    if (!this->close(false)) {
+    if (!this->closeDocumentInteractively(false)) {
         return false;
     }
 
@@ -2051,7 +2051,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         }
     }
 
-    this->closeDocument();
+    this->discardDocument();
 
     this->doc->lock();
     this->doc->clearDocument();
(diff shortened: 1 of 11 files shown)


src/core/control/Control.cpp
```diff
@@ -2441,12 +2474,12 @@ void Control::quit(bool allowCancel) {
...
     this->scheduler->unlock();
     this->scheduler->stop();  // Finish current task. Must be called to finish pending saves.
-    this->closeDocument();    // Must be done after all jobs has finished (Segfault on save/export)
+    this->discardDocument();  // Must be done after all jobs has finished (Segfault on save/export)
     settings->save();
     g_application_quit(G_APPLICATION(gtkApp));
 }
 
-auto Control::close(const bool allowDestroy, const bool allowCancel) -> bool {
+auto Control::closeDocumentInteractively(const bool allowDestroy, const bool allowCancel) -> bool {
```
fabian-thomas, line 2482: We could consider renaming that method to close again if we would add something like forceClose as argument.

tjiagoM:
Hi @fabian-thomas I know it has been quite a while for this PR here, but was wondering whether you have any plans on coming back to this very useful feature. Sorry I cannot offer my help as I don't know c++ :/

rolandlo:
@fabian-thomas If you have given up on this PR, please close it.

PR #3652, opened 2021-12-09
Fix #3651 - remove pressure value on Highlighter strokes first point

### LINKED ISSUE #3651: The first point of Highlighter strokes has a pressure value
**Affects versions :**
 - OS: all
 - Version of Xournal++: `release-1.1` since #3344 and `master` since #3644.

**Describe the bug**
On a pressure sensitive device (or with pressure inference), the first point of highlighter strokes gets a pressure value assigned. This leads to `stroke::hasPressure` to return true, thus messing up saved files by adding a useless array of pressure values.

**To Reproduce**
Create a highlighter stroke with a pressure sensitive device, save and inspect the file.

A fix is coming soon :-)

bhennion:
Merging in `release-1.1` in 24h unless someone objects.

This will also need to be added to master ASAP: the bug corrupts files (with little harm, but still). How should I do that? Another merge, cherry-picking?

### DIFF
--- a/src/control/tools/InputHandler.cpp
+++ b/src/control/tools/InputHandler.cpp
@@ -53,9 +53,11 @@ void InputHandler::createStroke(Point p) {
         }
     } else if (h->getToolType() == TOOL_HIGHLIGHTER) {
         stroke->setToolType(STROKE_TOOL_HIGHLIGHTER);
+        p.z = Point::NO_PRESSURE;
     } else if (h->getToolType() == TOOL_ERASER) {
         stroke->setToolType(STROKE_TOOL_ERASER);
         stroke->setColor(0xffffffU);
+        p.z = Point::NO_PRESSURE;
     }
 
     stroke->addPoint(p);


Febbe:
Rebase this onto release-1.1, then merge release-1.1 into master

src/control/tools/StrokeHandler.cpp
```diff
@@ -369,9 +369,14 @@ void StrokeHandler::onButtonPressEvent(const PositionInputData& pos) {
...
-        createStroke(Point(this->buttonDownPoint.x, this->buttonDownPoint.y, pos.pressure));
+        this->hasPressure = this->xournal->getControl()->getToolHandler()->getToolType() == TOOL_PEN &&
+                            pos.pressure != Point::NO_PRESSURE;
 
-        this->hasPressure = this->stroke->getToolType() == STROKE_TOOL_PEN && pos.pressure != Point::NO_PRESSURE;
+        if (this->hasPressure) {
+            createStroke(Point(this->buttonDownPoint.x, this->buttonDownPoint.y, pos.pressure));
+        } else {
+            createStroke(Point(this->buttonDownPoint.x, this->buttonDownPoint.y));
+        }
```
Febbe, line 379: What's the difference to this?:
```suggestion
            createStroke(Point(this->buttonDownPoint.x, this->buttonDownPoint.y, pos.pressure));
```

bhennion: When the tool is the highlighter, then the event `pos` may have a pressure value, but we do not want to give the point that pressure value.
The boolean `this->hasPressure` is always `false` when the tool is the highlighter, and `true` when the tool is the normal pen and the device is pressure sensitve.

Febbe: Ok, looked into the source of the function.
Wouldn't it an idea to move that into it? Then the pressure is not set wrongly at other places?

Febbe: If not this LGTM and you can merge it ASAP.

Febbe:
LGTM

PR #3664, opened 2021-12-15
Fix: clear undoRedo properly on close

### LINKED ISSUE #2878: When switching from one document to a pdf document confirmation is asked twice
**Affects versions :**
 - Version of Xournal++: 1.1.0+dev (doesn't affect 1.0.19)

**Describe the bug**
When I switch from one Xournal++ document (with or without pdf background) to a pdf-document by clicking `Open` and selecting a pdf, I need to confirm twice to `Discard` the current document.

**To Reproduce**
1. Open a new document and draw something
2. Click on menu `File > Open` and select a pdf-file
3. Observe that you need to confirm twice to discard the current document

**Expected behavior**
One confirmation instead of two

**Additional context**
The problem is that first `Control::openFile` calls `Control::close` and then `Control::annotatePdf` does the same thing (without noticing that it doesn't need to display the dialog any more).

changhc:
Fixes #2878.

#### Explanation
`Control::close` checks if there is anything not yet saved and closes the document if a user chooses to discard the changes. However, it does not actually clean up `undoRedo` if the target file should not be destroyed. I'm not sure about the rationale behind this.

Anyway, I change the code to clear `undoRedo` whenever a user chooses to discard unsaved changes. It is `Control::closeDocument`, which is called by `Control::close`, that actually determines whether a file should be destroyed. It was defaulted to destroy all files when it was called. I just changed it to letting the caller decide whether the file should be cleaned up.

The question now is: is there any use case where we don't want to clear `undoRedo` after we close a file?

#### Test
The bug case mentioned in the issue is fixed with the changes.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2485,17 +2485,17 @@ auto Control::close(const bool allowDestroy, const bool allowCancel) -> bool {
         }
     }
 
-    if (allowDestroy && discard) {
-        this->closeDocument();
+    if (discard) {
+        this->closeDocument(allowDestroy);
     }
     return true;
 }
 
-void Control::closeDocument() {
+void Control::closeDocument(bool destroy) {
     this->undoRedo->clearContents();
 
     this->doc->lock();
-    this->doc->clearDocument(true);
+    this->doc->clearDocument(destroy);
     this->doc->unlock();
 
     this->undoRedoChanged();
--- a/src/core/control/Control.h
+++ b/src/core/control/Control.h
@@ -327,7 +327,7 @@ private:
     /**
      * "Closes" the document, preparing the editor for a new document.
      */
-    void closeDocument();
+    void closeDocument(bool destroy = true);
 
     /**
      * Applies the preferred language to the UI


rolandlo:
Thanks for the PR, which is very welcome. I agree that it fixes #2878. Testing out your PR I noticed that the app crashes sometimes via the following steps.
1. Open a xopp-file that annotates a pdf
2. Draw something
3. Use menu `File Open` and open another .xopp-file that annotates a pdf, while discarding the changes. 
4. Notice that the app (sometimes) crashes when you click on `discard`.

I am not sure the crash was introduced by your PR, but I didn't get a crash using the latest version (from PPA) in the same manner. So could you please investigate this issue?

changhc:
Thanks for the comment. By "sometimes", do you mean you repeated exactly the same steps for a few times and only some attempts failed?

rolandlo:
> Thanks for the comment. By "sometimes", do you mean you repeated exactly the same steps for a few times and only some attempts failed?

Yes, that's what I observed (if I didn't make a mistake). The crash was actually a freeze and it happened around 50% of the times I tried.

changhc:
I tried to reproduce what you said but things were totally normal on my side. This is what I did:
- Prepare two pdf files A and B
- Create an .xopp file for both A and B
- Open A.xopp and draw something
- Open B.xopp and discard changes made in A.xopp
- Draw something in B.xopp
- Open A.xopp and discard changes made in B.xopp
- Repeat

I repeated the draw-and-discard steps for more than 20 times and I didn't observe any crash. I'm using Ubuntu 20.04, though I don't think this is platform-dependent. Can anyone else try to reproduce the situation?

rolandlo:
I discovered that the freeze doesn't happen randomly, but it depends on which page is open (and modified) for the document (with pdf background) whose changes you discard. If you discard changes on the first page, there is no freeze. If you discard changes on a different page, I get a freeze. The new document I try to open does not matter.
Take these files for testing:
[freezeTestDocs.zip](https://github.com/xournalpp/xournalpp/files/7741034/freezeTestDocs.zip)

Can you reproduce the freeze now?

changhc:
Yup now I see the issue. I modified my code to not calling `this->doc->clearDocument` if `allowDestroy` is false, and then the problem was solved. There's something in `this->doc->clearDocument` (not `this->doc->lock`; I tried not calling the lock but it didn't help) that causes the freeze. I'll look into it.

Technius:
Side note (unrelated to this PR): the use of mutexes in `Document::closeDocument` and `Document::operator=` is incorrect: https://github.com/xournalpp/xournalpp/blob/6a6435e9f221c3ab624b2779d92ecc523ca97a29/src/core/model/Document.cpp#L84-L90

bhennion:
What is the status of this PR? It looks like a useful bugfix to me.

changhc:
Hey sorry for the late reply. I wanted to look into the `allowDestroy` flag as mentioned above but just got busy since then and forgot about this. If anyone has any idea, I'd appreciate your comment.

Issue #3611, opened 2021-11-23
Remove global locale


idotobi:
**Is your feature request related to a problem? Please describe.**

Setting a global cpp locale caused some problems in the past (see https://github.com/xournalpp/xournalpp/issues/3606) and required some fixes (see https://github.com/xournalpp/xournalpp/pull/3610). However, these rather cured the symptom than the cause.

**Describe the solution you'd like**

A cleaner way would be to remove the global locale, by adapting

https://github.com/xournalpp/xournalpp/blob/f72fe7c85bf6ac0525350362b946e10df7a9371c/src/core/control/XournalMain.cpp#L61-L89

**Describe alternatives you've considered**

* adding `imbue` to a lot of places in the codebase and testing several locales whenever input/output is implemented

**Note**
This change should be followed by extensive testing, as it might have far reaching unintended side effects.

Febbe:
The thing is, that we always want to have the C locale for serdes and always want to have the default locale for display.
When we touch this, we should create 2 template functions:
```
template<class Stream, class ...Args>
auto xoj::init_serdes_stream(Args... args) -> Stream{
    Stream stream{std::forward<Args>(args)...};
    stream.imbue(std::locale::default());
}

template<class Stream, class ...Args>
auto xoj::init_display_stream(Args... args) -> Stream;
```

Regarding the global locale: I don't know if we want to set it to the C locale, since it **could** affect the locale of gtk.

Febbe:
> Regarding the global locale: I don't know if we want to set it to the C locale, since it could affect the locale of gtk.

Gtk itself sets the c-locale to the system's locale, so we should not override that after [e.g. gtk_init](https://docs.gtk.org/gtk3/func.disable_setlocale.html).

bhennion:
> The thing is, that we always want to have the C locale for serdes and always want to have the default locale for display. When we touch this, we should create 2 template functions:
> 
> ```
> template<class Stream, class ...Args>
> auto xoj::init_serdes_stream(Args... args) -> Stream{
>     Stream stream{std::forward<Args>(args)...};
>     stream.imbue(std::locale::classic());
>     return stream;
> }
> 
> template<class Stream, class ...Args>
> auto xoj::init_display_stream(Args... args) -> Stream;
> ```
> 
> Regarding the global locale: I don't know if we want to set it to the C locale, since it **could** affect the locale of gtk.

I'd do it the other way around: leave the default C++ locale to "C", and imbue the display streams with the system locale.
[...]

PR #3676, opened 2021-12-21
Fix #3655: Use stod for loading doubles

### LINKED ISSUE #3655: Floating point numbers are loaded incorrectly from pagetemplates.ini
**Affects versions :**
 - OS: Ubuntu
     - 3.24.20
 - Version of Xournal++: [e.g. 1.0.16, specific commit, specific branch]
     - latest of 1.1-release branch (01737d1)
 - Installation method: from source

**Describe the bug**
Floating point numbers in page template configs are converted to integers when they're loaded by Xournal++. That is a problem because the line width config needs floating point numbers to be set in a meaningful way.

I think the problem is located [here](https://github.com/xournalpp/xournalpp/blob/bc5f6047910a9978b440a6234eaebd831062bdd4/src/core/view/background/BackgroundConfig.cpp#L45). The string is converted to a (unsigned) long, truncating the decimal places. That method is used by [BaseBackgroundPainter](https://github.com/xournalpp/xournalpp/blob/bc5f6047910a9978b440a6234eaebd831062bdd4/src/core/view/background/BaseBackgroundPainter.cpp#L57) to load the value in.

After building release-1.1 from source and changing that line so it converts to double (stod) everything works as expected for me.

**To Reproduce**
Put this in your pagetemplates.ini and select the page background in Xournal++ (or use a plugin to change your background config). If you change the line width to 1.0, the lines should be visible again. If you set it to anything between 0 and less than 1, no lines should be visible.

```
[FloatingPointBug]
name=Example of line width not working
format=graph
config=lw=0.9
```

In my case there are no lines since the .9 seems to be truncated, only leaving a 0 as the value.

**Expected behavior**
The floating point values should be loaded correctly and the lines should be clearly visible when setting the line width to 0.9.

DielN:
Fix BackgroundConfig::loadValue for doubles (used for loading in background configs from pagetemplates.ini) by using stod instead of stoul to preserve the decimal places of the configured value, which would be truncated otherwise (see #3655)

This relies on the locale for LC_NUMERIC to be "C" (e2a1f97327d94818cdecb19795308f8aed3177eb). There's been some discussion about this in #3611. Because of that, I'm not sure if using stod is the best way to fix this (as stod is locale-dependent).

If the locale was a locale which uses a comma as the decimal separator (for example de_DE.UTF-8), there would be two problems:
- Different configs work differently on different locales (lw=1.2 will work on en_US, but if that config is shared with a user with a de_DE locale, it won't work for that user without changing the decimal separator)
- Users using a locale with comma as decimal separator will only be able to use scientific notation, since "," would be interpreted as a separator for the key-value pairs of the config (e.g. config=lw=1.2,m1=40)

So maybe using GLib's [ascii_strtod](https://docs.gtk.org/glib/func.ascii_strtod.html) would be a better option. 
Another way would be using istringstream an imbuing C-locale. Doesn't seem like a bad option either.

### DIFF
--- a/src/core/view/background/BackgroundConfig.cpp
+++ b/src/core/view/background/BackgroundConfig.cpp
@@ -1,8 +1,10 @@
 #include "BackgroundConfig.h"
 
+#include <sstream>
 #include <utility>
 
 #include "util/StringUtils.h"
+#include "util/serdesstream.h"
 
 using std::string;
 
@@ -42,7 +44,8 @@ auto BackgroundConfig::loadValue(const string& key, int& value) const -> bool {
 auto BackgroundConfig::loadValue(const string& key, double& value) const -> bool {
     string str;
     if (loadValue(key, str)) {
-        value = std::stoul(str, nullptr, 10);
+        auto valueStream = serdes_stream<std::istringstream>(str);
+        valueStream >> value;
         return true;
     }
 


rolandlo:
I would go with `ascii_strtod` 
@bhennion What's your opinion?

bhennion:
In general, I'd avoid using `glib` for this kind of generic things that are now covered by the `std` libraries.
So I'd go for a stream, contructed using `serdes_stream` introduced in 4b741987 (see [this file](https://github.com/xournalpp/xournalpp/blob/master/src/util/include/util/serdesstream.h)).


As a sidenote: the question also extends to the rest of the `BackgroundConfig` class, as string-to-integer-type conversion may depend on the locale (see e.g. [this page](https://en.cppreference.com/w/cpp/string/byte/strtoul)):

> Additional numeric formats may be accepted by the currently installed C locale.

Changing this however may go beyond this PR's scope.

DielN:
I changed it to use serdes_stream. Seems to be working for me, even when removing the global "C"-locale.

Regarding the other functions in `BackgroundConfig`, I think that should be done in a new PR when there's a decision about #3611 (I'm unsure if there's any final conclusion to that issue yet?).

rolandlo (approved):
LGTM

rolandlo:
Thanks for the fix, @DielN. I merged it.

PR #3677, opened 2021-12-21
Add pageTypeConfig to app.getDocumentStructure()

DielN:
I've been playing around with the plugin API and background configs. 
Bizarrely, you can get the format (graph, dotted, ...) of your page/background with getDocumentStructure(), but not the config (line width etc.). This should be included, since a full PageType has both. You can pass both to changeCurrentPageBackground via the plugin API.

I've added it and it seems to work fine, feel free to test it with the changed example plugin in 4a53df667fcc8c8d2cb95813cd99a373f0115161.

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -764,6 +764,8 @@ static int applib_getToolInfo(lua_State* L) {
  *       "pageHeight" = number,
  *       "isAnnoated" = bool,
  *       "pageTypeFormat" = string,
+ *       "pageTypeConfig" = string,
+ *       "backgroundColor" = integer,
  *       "pdfBackgroundPageNo" = integer (0, if there is no pdf background page),
  *       "layers" = {
  *         [0] = {
@@ -818,6 +820,14 @@ static int applib_getDocumentStructure(lua_State* L) {
         PageType pt = page->getBackgroundType();
         std::string pageTypeFormat = PageTypeHandler::getStringForPageTypeFormat(pt.format);
         lua_pushstring(L, pageTypeFormat.c_str());
+        lua_settable(L, -3);
+
+        lua_pushliteral(L, "pageTypeConfig");
+        lua_pushstring(L, pt.config.c_str());
+        lua_settable(L, -3);
+
+        lua_pushliteral(L, "backgroundColor");
+        lua_pushinteger(L, int(uint32_t(page->getBackgroundColor())));
         lua_settable(L, -3);
 
         lua_pushliteral(L, "pdfBackgroundPageNo");


DielN:
I've had a change of mind. I'm working on a plugin which allows users to configure their page templates and export them so they can put it in their pagetemplates.ini. That plugin would use this change to getDocumentStructure().
I could change this PR to be a draft for that plugin and include my WIP code soon.

You can check out the code added to the plugin API for including pageTypeConfig in getDocumentStructure() meanwhile. 
Please let me know if this PR should be converted to one for the plugin (while still including the change to getDocumentStructure() of course) or that should be a separate PR.

rolandlo:
Please create a separate PR. I will review the current PR soon and merge it, if it works.

rolandlo:
Your PR works fine. I noticed that the background color (accessible via `page->getBackgroundColor()`) is also missing. If you would like, add that one as well.

DielN:
I added the backgroundColor of the page to getDocumentStructure().

For pushing the color as an int to lua, I did it the same way it's done in getToolInfo().
I'm just not sure why the color is converted to int after being converted to uint32_t. It seems to work with just pushing uint32_t as well. It even stays unsigned in lua. If converted to int, it seems like it's converted to a signed 64 bit number in lua, that's why the output looks a bit weird when printing the color as hex in lua.

It doesn't really matter, since the important 32 bits/4 bytes (for the color values) stay unchanged, but wouldn't it be better to just leave it as uint32_t or would that break something in some edge case?

rolandlo:
> I added the backgroundColor of the page to getDocumentStructure().
> 
> For pushing the color as an int to lua, I did it the same way it's done in getToolInfo(). I'm just not sure why the color is converted to int after being converted to uint32_t. It seems to work with just pushing uint32_t as well. It even stays unsigned in lua. If converted to int, it seems like it's converted to a signed 64 bit number in lua, that's why the output looks a bit weird when printing the color as hex in lua.
> 
> It doesn't really matter, since the important 32 bits/4 bytes (for the color values) stay unchanged, but wouldn't it be better to just leave it as uint32_t or would that break something in some edge case?

Hmm, I'm not sure about that. `lua_pushinteger` takes a `lua_Integer` as its second argument, which has  a platform-specific size. 
> Standard Lua uses 64-bit integers and double-precision (64-bit) floats, but you can also compile Lua so that it uses 32-bit integers and/or single-precision (32-bit) floats. The option with 32 bits for both integers and floats is particularly attractive for small machines and embedded systems. (See macro LUA_32BITS in file luaconf.h.)

So what is happening on a 32-bit machine? The practice of first converting to `uint32_t`, then to `int`, was introduced by @Febbe in https://github.com/xournalpp/xournalpp/pull/3623/commits/fe810967e14623e03444864e2971b32ea0dc0b72
So @Febbe could you please explain why we do that?

rolandlo:
Since the PR is fully functional, I will merge it once it is squashed into one commit.

DielN:
Rebased on current master & squashed into one commit, feel free to merge.

rolandlo:
Thanks for your contribution! Just merged it.

Issue #2850, opened 2021-02-22
Automated Whiteboard Transcription


WillNilges:
**Is your feature request related to a problem? Please describe.**

I frequently do out my homework on the whiteboard, but then take a picture of it. It'd be nice if I could have Xournalpp scan a photo, and export the strokes to a .xopp file.

**Describe the solution you'd like**
A button that would allow you to drop in an image, and have xournalpp vectorize it, and place it in the working document.

**Describe alternatives you've considered**
I am investigating writing a program to do this using [autotrace](http://autotrace.sourceforge.net/). Work on that can be found [here](https://github.com/willnilges/imboard) (Spoiler: I had this idea literally two days ago at a hackathon then spent the weekend working on a completely different idea)

**Additional context**
Autotrace seems to work pretty well as a backend, at least for monochrome images. Additional colors also work, but aren't as good, since then you have to specify a higher color-count.

```
$ autotrace circles02.jpg --color-count 2 --output-format svg > outputimage.svg
```

![circles02](https://user-images.githubusercontent.com/42927786/108782363-71eee780-7539-11eb-824a-b597fdc63306.jpg)

![outputimage](https://user-images.githubusercontent.com/42927786/108782395-7ddaa980-7539-11eb-8b78-fa2a413f8779.jpg)

Anyone interested in helping me out with this?

personalizedrefrigerator:
It would be awesome to have this!

personalizedrefrigerator:
This seems kindof like the functionality I would expect a plugin to have (but I'm not sure Lua is the best choice for this...).

WillNilges:
Oh, the Lua thing is an artifact of my hackathon project. This will be pure C or C++.

personalizedrefrigerator:
> Oh, the Lua thing is an artifact of my hackathon project. This will be pure C or C++.

Here is a list of plugin-related resources:
 * [Example plugins](https://github.com/xournalpp/xournalpp/tree/master/plugins). 
 * [C-based plugin API](https://github.com/xournalpp/xournalpp/blob/master/src/plugin/luapi_application.h).
 * [A pull request that extends the Plugin API with new functionality](https://github.com/xournalpp/xournalpp/pull/2176/files).

WillNilges:
Sweet, thanks! I'll comment on that other PR, but this does kinda seem like the kind of thing that can't (easily) be a plugin.

Unless, a Lua API can somehow interface with my code.

[...]

PR #3688, opened 2021-12-28
Inkpath api extensions

WillNilges:
Adds additional functionality to the Lua API to make [inkpath](https://github.com/willnilges/inkpath) work. The plugin uses [autotrace](https://github.com/yy502/autotrace) to vectorize whiteboard strokes and insert them in the working document. See discussion of the feature [here](https://github.com/xournalpp/xournalpp/issues/2850). And [here](https://github.com/WillNilges/inkpath/issues/10) is the current roadmap for the project.

- `getFilePath`: Opens a GTK File Chooser dialogue used to prompt user to pick an image to transcribe. Allows filtering based on file type to make it useful for other file-choosing purposes.
- `addStrokes`: Places a batch of strokes on the canvas when given a list of lists of X, Y, and optionally pressure coordinates. Also takes in pen tool options per-stroke. Also allows specifying per-stroke undoing or stroke batch undoing.
- `addSplines`: Places a batch of spline-based strokes on the canvas given a list of lists of splineSegment coordinates. Also optionally takes in pen tool options per-stroke. Also allows specifying per-stroke undoing or stroke batch undoing.
- `addStrokeHelper`: Reduces code duplication between the above two functions by parsing user-provided tool options before adding an element to the document.
- `refreshPage`: Notifies Xournalpp that the page changed (Used after either of the `add` functions)

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -24,8 +24,10 @@
 #include "gui/XournalView.h"
 #include "gui/widgets/XournalWidget.h"
 #include "model/Font.h"
+#include "model/SplineSegment.h"
 #include "model/StrokeStyle.h"
 #include "model/Text.h"
+#include "undo/InsertUndoAction.h"
 #include "util/StringUtils.h"
 #include "util/XojMsgBox.h"
 #include "util/safe_casts.h"
@@ -101,6 +103,56 @@ static int applib_saveAs(lua_State* L) {
     // Destroy the dialog and free memory
     g_object_unref(native);
 
+    return args_returned;
+}
+
+/**
+ * Create a 'Open File' native dialog and return as a string
+ * the filepath the user chose to open.
+ *
+ * Examples:
+ *   path = app.getFilePath()
+ *   path = app.getFilePath({'*.bmp', '*.png'})
+ */
+static int applib_getFilePath(lua_State* L) {
+    GtkFileChooserNative* native =
+            gtk_file_chooser_native_new(_("Open file"), nullptr, GTK_FILE_CHOOSER_ACTION_OPEN, nullptr, nullptr);
+    gint res;
+    int args_returned = 0;  // change to 1 if user chooses file
+    char* filename;
+
+    // Get vector of supported formats from Lua stack
+    std::vector<std::string> formats;
+    // stack now contains: -1 => table
+    lua_pushnil(L);
+    // stack now contains: -1 => nil; -2 => table
+    while (lua_next(L, -2)) {
+        // stack now contains: -1 => value; -2 => key; -3 => table
+        const char* value = lua_tostring(L, -1);
+        formats.push_back(value);
+        lua_pop(L, 1);
+        // stack now contains: -1 => key; -2 => table
+    }
+    // stack now contains: -1 => table
+    lua_pop(L, 1);  // Stack is now the same as it was on entry to this function
+    if (formats.size() > 0) {
+        GtkFileFilter* filterSupported = gtk_file_filter_new();
+        gtk_file_filter_set_name(filterSupported, _("Supported files"));
+        for (std::string format: formats) gtk_file_filter_add_pattern(filterSupported, format.c_str());
+        gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(native), filterSupported);
+    }
+
+    // Wait until user responds to dialog
+    res = gtk_native_dialog_run(GTK_NATIVE_DIALOG(native));
+    // Return the filename chosen to lua
+    if (res == GTK_RESPONSE_ACCEPT) {
+        filename = static_cast<char*>(gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(native)));
+        lua_pushlstring(L, filename, strlen(filename));
+        g_free(static_cast<gchar*>(filename));
+        args_returned = 1;
+    }
+    // Destroy the dialog and free memory
+    g_object_unref(native);
     return args_returned;
 }
 
@@ -202,7 +254,7 @@ static int applib_registerUi(lua_State* L) {
 /**
  * Execute an UI action (usually internally called from Toolbar / Menu)
  * The argument consists of a Lua table with 3 keys: "action", "group" and "enabled"
- * The key "group" is currently only used for debugging purpose and can savely be omitted.
+ * The key "group" is currently only used for debugging purpose and can safely be omitted.
  * The key "enabled" is true by default.
  *
  * Example 1: app.uiAction({["action"] = "ACTION_PASTE"})
@@ -326,6 +378,429 @@ static int applib_layerAction(lua_State* L) {
     ctrl->getLayerController()->actionPerformed(action);
 
     return 1;
+}
+
+/**
+ * Helper function for addStroke API. Parses pen settings from API call, taking
+ * in a Stroke and a chosen Layer, sets the pen settings, and applies the stroke.
+ */
+static void addStrokeHelper(lua_State* L, Stroke* stroke) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    PageRef const& page = ctrl->getCurrentPage();
+    Layer* layer = page->getSelectedLayer();
+
+    std::string size;
+    double thickness;
+    int fillOpacity;
+    bool filled;
+    Color color;
+    std::string lineStyle;
+    ToolHandler* toolHandler;
+    const char* tool;
+    std::string allowUndoRedoAction;
+
+    // Get attributes.
+    lua_getfield(L, -1, "tool");
+    lua_getfield(L, -2, "width");
+    lua_getfield(L, -3, "color");
+    lua_getfield(L, -4, "fill");
+    lua_getfield(L, -5, "lineStyle");
+
+    tool = luaL_optstring(L, -5, "");  // We're gonna need the tool type.
+
+    toolHandler = ctrl->getToolHandler();
+
+    // TODO: (willnilges) Handle DrawingType?
+    // TODO: (willnilges) Break out Eraser functionality into a new API call.
+
+    // Set tool type
+    if (strcmp("highlighter", tool) == 0) {
+        stroke->setToolType(STROKE_TOOL_HIGHLIGHTER);
+
+        size = toolSizeToString(toolHandler->getHighlighterSize());
+        thickness = toolHandler->getToolThickness(TOOL_HIGHLIGHTER)[toolSizeFromString(size)];
+
+        fillOpacity = toolHandler->getHighlighterFill();
+        filled = toolHandler->getHighlighterFillEnabled();
+
+        Tool& tool = toolHandler->getTool(TOOL_HIGHLIGHTER);
+        color = tool.getColor();
+    } else {
+        if (!(strcmp("pen", tool) == 0))
+            g_warning("%s", FC(_F("Unknown stroke type: \"{1}\", defaulting to pen") % tool));
+
+        stroke->setToolType(STROKE_TOOL_PEN);
+
+        size = toolSizeToString(toolHandler->getPenSize());
+        thickness = toolHandler->getToolThickness(TOOL_PEN)[toolSizeFromString(size)];
+
+        fillOpacity = toolHandler->getPenFill();
+        filled = toolHandler->getPenFillEnabled();
+
+        Tool& tool = toolHandler->getTool(TOOL_PEN);
+        color = tool.getColor();
+        lineStyle = StrokeStyle::formatStyle(tool.getLineStyle());
+    }
+
+    // Set width
+    if (lua_isnumber(L, -4))  // Check if the width was provided
+        stroke->setWidth(lua_tonumber(L, -4));
+    else
+        stroke->setWidth(thickness);
+
+    // Set color
+    if (lua_isinteger(L, -3))  // Check if the color was provided
+        stroke->setColor(Color(lua_tointeger(L, -3)));
+    else
+        stroke->setColor(color);
+
+    // Set fill
+    if (lua_isinteger(L, -2))  // Check if fill settings were provided
+        stroke->setFill(lua_tointeger(L, -2));
+    else if (filled)
+        stroke->setFill(fillOpacity);
+    else
+        stroke->setFill(-1);  // No fill
+
+    // Set line style
+    if (lua_isstring(L, -1))  // Check if line style settings were provided
+        stroke->setLineStyle(StrokeStyle::parseStyle(lua_tostring(L, -1)));
+    else
+        stroke->setLineStyle(StrokeStyle::parseStyle(lineStyle.data()));
+
+    lua_pop(L, 5);  // Finally done with all that Lua data.
+
+    // Add the stroke
+    layer->addElement(stroke);
+    return;
+}
+
+/**
+ * Given a table containing a series of splines, draws a batch of strokes on the canvas.
+ * Expects a table of tables containing eight coordinate pairs, along with attributes of the stroke.
+ *
+ * Required Arguments: splines
+ * Optional Arguments: pressure, tool, width, color, fill, lineStyle
+ *
+ * If optional arguments are not provided, the specified tool settings are used.
+ * If the tool is not provided, the current pen settings are used.
+ * The only tools supported are Pen and Highlighter.
+ *
+ * The function expects 8 points per spline segment. Due to the nature of cubic
+ * splines, you must pass your points in a repeating pattern:
+ * startX, startY, ctrl1X, ctrl1Y, ctrl2X, ctrl2Y, endX, endY, startX, startY, ...
+ *
+ * The function checks that the length of the coordinate table is divisible by eight, and will throw
+ * an error if it is not.
+ *
+ * Example: app.addSplines({
+ *            ["splines"] = { -- The outer table is a table of strokes
+ *                ["coordinates"] = { -- Each inner table is a coord stream that represents SplineSegments that can be
+ * assembled into a stroke
+ *                  [1] = 880.0, // Every eight coordinates (4 pairs) is a SplineSegment
+ *                  [2] = 874.0,
+ *                  [3] = 881.3295,
+ *                  [4] = 851.5736,
+ *                  [5] = 877.2915,
+ *                  [6] = 828.2946,
+ *                  [7] = 875.1697,
+ *                  [8] = 806.0,
+ *                  ... -- A spline can be made up of as many SplineSegments as is necessary.
+ *                },
+ *                -- Tool options are also specified per-stroke
+ *                ["width"] = 1.4,
+ *                ["color"] = 0xff0000,
+ *                ["fill"] = 0,
+ *                ["tool"] = "pen",
+ *                ["lineStyle"] = "plain"
+ *            },
+ *            ["allowUndoRedoAction"] = "grouped", -- Each batch of splines can be grouped into one undo/redo action (or
+ * "individual" or "none")
+ *        })
+ */
+
+static int applib_addSplines(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    std::vector<Element*> strokes;
+    const char* allowUndoRedoAction;
+
+    // Discard any extra arguments passed in
+    lua_settop(L, 1);
+    luaL_checktype(L, 1, LUA_TTABLE);
+
+    lua_getfield(L, 1, "splines");
+    if (!lua_istable(L, -1))
+        luaL_error(L, "Missing spline table!");
+
+    size_t numSplines = lua_rawlen(L, -1);
+    for (size_t a = 1; a <= numSplines; a++) {
+        std::vector<double> coordStream;
+        Stroke* stroke = new Stroke();
+        // Get coordinates
+        lua_pushnumber(L, a);
+        lua_gettable(L, -2);
+        lua_getfield(L, -1, "coordinates");
+        if (!lua_istable(L, -1))
+            luaL_error(L, "Missing coordinate table!");
+        size_t numCoords = lua_rawlen(L, -1);
+        for (size_t b = 1; b <= numCoords; b++) {
+            lua_pushnumber(L, b);
+            lua_gettable(L, -2);
+            double point = lua_tonumber(L, -1);
+            coordStream.push_back(point);  // Each segment is going to have multiples of 8 points.
+            lua_pop(L, 1);
+        }
+        // pop value + copy of key, leaving original key
+        lua_pop(L, 1);
+        // Handle those points
+        // Check if the list is divisible by 8.
+        if (coordStream.size() % 8 != 0)
+            luaL_error(L, "Point table incomplete!");
+
+        // Now take that gigantic list of splines and create SplineSegments out of them.
+        long unsigned int i = 0;
+        while (i < coordStream.size()) {
+            // start, ctrl1, ctrl2, end
+            Point start = Point(coordStream.at(i), coordStream.at(i + 1), Point::NO_PRESSURE);
+            Point ctrl1 = Point(coordStream.at(i + 2), coordStream.at(i + 3), Point::NO_PRESSURE);
+            Point ctrl2 = Point(coordStream.at(i + 4), coordStream.at(i + 5), Point::NO_PRESSURE);
+            Point end = Point(coordStream.at(i + 6), coordStream.at(i + 7), Point::NO_PRESSURE);
+            i += 8;
+            SplineSegment segment = SplineSegment(start, ctrl1, ctrl2, end);
+            std::list<Point> raster = segment.toPointSequence();
+            for (Point point: raster) stroke->addPoint(point);
+            // TODO: (willnilges) Is there a way we can get Pressure with Splines?
+        }
+
+        if (stroke->getPointCount() >= 2) {
+            // Finish building the Stroke and apply it to the layer.
+            addStrokeHelper(L, stroke);
+            strokes.push_back(stroke);
+        } else
+            g_warning("Stroke shorter than two points. Discarding. (Has %d)", stroke->getPointCount());
+        // Onto the next stroke
+        lua_pop(L, 1);
+    }
+
+    lua_pop(L, 1);  // Stack is now the same as it was on entry to this function
+
+    // Check how the user wants to handle undoing
+    lua_getfield(L, 1, "allowUndoRedoAction");
+    allowUndoRedoAction = luaL_optstring(L, -1, "grouped");
+    if (strcmp("grouped", allowUndoRedoAction) == 0) {
+        PageRef const& page = ctrl->getCurrentPage();
+        Layer* layer = page->getSelectedLayer();
+        UndoRedoHandler* undo = ctrl->getUndoRedoHandler();
+        undo->addUndoAction(std::make_unique<InsertsUndoAction>(page, layer, strokes));
+    } else if (strcmp("individual", allowUndoRedoAction) == 0) {
+        PageRef const& page = ctrl->getCurrentPage();
+        Layer* layer = page->getSelectedLayer();
+        UndoRedoHandler* undo = ctrl->getUndoRedoHandler();
+        for (Element* element: strokes) undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, element));
+    } else if (strcmp("none", allowUndoRedoAction) == 0)
+        g_warning("Not allowing undo/redo action.");
+    else
+        g_warning("Unrecognized undo/redo option: %s", allowUndoRedoAction);
+    lua_pop(L, 1);
+    return 0;
+}
+
+/**
+ * Given a table of sets of points, draws a batch of strokes on the canvas.
+ * Expects three tables of equal length: one for X, one for Y, and one for
+ * stroke pressure, along with attributes of the stroke. Each stroke has
+ * attributes handled individually.
+ *
+ * Required Arguments: X, Y
+ * Optional Arguments: pressure, tool, width, color, fill, lineStyle
+ *
+ * If optional arguments are not provided, the specified tool settings are used.
+ * If the tool is not provided, the current pen settings are used.
+ * The only tools supported are Pen and Highlighter.
+ *
+ * The function checks for consistency among table lengths, and throws an
+ * error if there is a discrepancy
+ *
+ * Example:
+ *
+ * app.addStrokes({
+ *     ["strokes"] = { -- The outer table is a table of strokes
+ *         {   -- Inside a stroke are three tables of equivalent length that represent a series of points
+ *             ["x"]        = { [1] = 110.0, [2] = 120.0, [3] = 130.0, ... },
+ *             ["y"]        = { [1] = 200.0, [2] = 205.0, [3] = 210.0, ... },
+ *             ["pressure"] = { [1] = 0.8,   [2] = 0.9,   [3] = 1.1, ... },
+ *             -- Each stroke has individually handled options
+ *             ["tool"] = "pen",
+ *             ["width"] = 3.8,
+ *             ["color"] = 0xa000f0,
+ *             ["fill"] = 0,
+ *             ["lineStyle"] = "solid",
+ *         },
+ *         {
+ *             ["x"]        = { [1] = 310.0, [2] = 320.0, [3] = 330.0, ... },
+ *             ["y"]        = { [1] = 300.0, [2] = 305.0, [3] = 310.0, ... },
+ *             ["pressure"] = { [1] = 3.0,   [2] = 3.0,   [3] = 3.0, ... },
+ *             ["tool"] = "pen",
+ *             ["width"] = 1.21,
+ *             ["color"] = 0x808000,
+ *             ["fill"] = 0,
+ *             ["lineStyle"] = "solid",
+ *         },
+ *         {
+ *             ["x"]        = { [1] = 27.0,  [2] = 28.0,  [3] = 30.0, ... },
+ *             ["y"]        = { [1] = 100.0, [2] = 102.3, [3] = 102.5, ... },
+ *             ["pressure"] = { [1] = 1.0,   [2] = 1.0,   [3] = 1.0, ... },
+ *             ["tool"] = "pen",
+ *             ["width"] = 1.0,
+ *             ["color"] = 0x00aaaa,
+ *             ["fill"] = 0,
+ *             ["lineStyle"] = "dashdot",
+ *         },
+ *     },
+ *     ["allowUndoRedoAction"] = "grouped", -- Each batch of strokes can be grouped into one undo/redo action (or
+ * "individual" or "none")
+ * })
+ */
+static int applib_addStrokes(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    std::vector<Element*> strokes;
+    const char* allowUndoRedoAction;
+
+    // Discard any extra arguments passed in
+    lua_settop(L, 1);
+    luaL_checktype(L, 1, LUA_TTABLE);
+
+    lua_getfield(L, 1, "strokes");
+    if (!lua_istable(L, -1))
+        luaL_error(L, "Missing stroke table!");
+    size_t numStrokes = lua_rawlen(L, -1);
+    for (size_t a = 1; a <= numStrokes; a++) {
+        std::vector<double> xStream;
+        std::vector<double> yStream;
+        std::vector<double> pressureStream;
+        Stroke* stroke = new Stroke();
+
+        // Fetch table of X values from the Lua stack
+        lua_pushnumber(L, a);
+        lua_gettable(L, -2);
+
+        lua_getfield(L, -1, "x");
+        if (!lua_istable(L, -1))
+            luaL_error(L, "Missing X-Coordinate table!");
+        size_t xPoints = lua_rawlen(L, -1);
+        for (size_t b = 1; b <= xPoints; b++) {
+            lua_pushnumber(L, b);
+            lua_gettable(L, -2);
+            double value = lua_tonumber(L, -1);
+            xStream.push_back(value);
+            lua_pop(L, 1);
+        }
+        lua_pop(L, 1);
+
+        // Fetch table of Y values form the Lua stack
+        lua_getfield(L, -1, "y");
+        if (!lua_istable(L, -1))
+            luaL_error(L, "Missing Y-Coordinate table!");
+        size_t yPoints = lua_rawlen(L, -1);
+        for (size_t b = 1; b <= yPoints; b++) {
+            lua_pushnumber(L, b);
+            lua_gettable(L, -2);
+            double value = lua_tonumber(L, -1);
+            yStream.push_back(value);
+            lua_pop(L, 1);
+        }
+        lua_pop(L, 1);
+
+        // Fetch table of pressure values from the Lua stack
+        lua_getfield(L, -1, "pressure");
+        if (lua_istable(L, -1)) {
+            size_t pressurePoints = lua_rawlen(L, -1);
+            for (size_t b = 1; b <= pressurePoints; b++) {
+                lua_pushnumber(L, b);
+                lua_gettable(L, -2);
+                double value = lua_tonumber(L, -1);
+                pressureStream.push_back(value);
+                lua_pop(L, 1);
+            }
+        } else
+            g_warning("Missing pressure table. Assuming NO_PRESSURE.");
+
+        lua_pop(L, 1);
+
+        // Handle those points
+        // Make sure all vectors are the same length.
+        if (xStream.size() != yStream.size()) {
+            luaL_error(L, "X and Y vectors are not equal length!");
+        }
+        if (xStream.size() != pressureStream.size() && pressureStream.size() > 0)
+            luaL_error(L, "Pressure vector is not equal length!");
+
+        // Check and make sure there's enough points (need at least 2)
+        if (xStream.size() < 2) {
+            g_warning("Stroke shorter than two points. Discarding. (Has %ld/2)", xStream.size());
+            return 1;
+        }
+        // Add points to the stroke. Include pressure, if it exists.
+        if (pressureStream.size() > 0) {
+            for (long unsigned int i = 0; i < xStream.size(); i++) {
+                Point myPoint = Point(xStream.at(i), yStream.at(i), pressureStream.at(i));
+                stroke->addPoint(myPoint);
+            }
+        } else {
+            for (long unsigned int i = 0; i < xStream.size(); i++) {
+                Point myPoint = Point(xStream.at(i), yStream.at(i), Point::NO_PRESSURE);
+                stroke->addPoint(myPoint);
+            }
+        }
+
+        // Finish building the Stroke and apply it to the layer.
+        addStrokeHelper(L, stroke);
+        strokes.push_back(stroke);
+        // Onto the next stroke
+        lua_pop(L, 1);
+    }
+
+    // Check how the user wants to handle undoing
+    lua_getfield(L, 1, "allowUndoRedoAction");
+    allowUndoRedoAction = luaL_optstring(L, -1, "grouped");
+    if (strcmp("grouped", allowUndoRedoAction) == 0) {
+        PageRef const& page = ctrl->getCurrentPage();
+        Layer* layer = page->getSelectedLayer();
+        UndoRedoHandler* undo = ctrl->getUndoRedoHandler();
+        undo->addUndoAction(std::make_unique<InsertsUndoAction>(page, layer, strokes));
+    } else if (strcmp("individual", allowUndoRedoAction) == 0) {
+        PageRef const& page = ctrl->getCurrentPage();
+        Layer* layer = page->getSelectedLayer();
+        UndoRedoHandler* undo = ctrl->getUndoRedoHandler();
+        for (Element* element: strokes) undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, element));
+    } else if (strcmp("none", allowUndoRedoAction) == 0)
+        g_warning("Not allowing undo/redo action.");
+    else
+        g_warning("Unrecognized undo/redo option: %s", allowUndoRedoAction);
+
+    lua_pop(L, 1);  // Stack is now the same as it was on entry to this function
+
+    return 0;
+}
+
+/**
+ * Notifies program of any updates to the working document caused
+ * by the API.
+ *
+ * Example: app.refreshPage()
+ */
+static int applib_refreshPage(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    PageRef const& page = ctrl->getCurrentPage();
+    if (page)
+        page->firePageChanged();
+    else
+        g_warning("Called applib_refreshPage, but there is no current page.");
+    return 0;
 }
 
 /**
@@ -1257,6 +1732,10 @@ static const luaL_Reg applib[] = {{"msgbox", applib_msgbox},
                                   {"scaleTextElements", applib_scaleTextElements},
                                   {"getDisplayDpi", applib_getDisplayDpi},
                                   {"export", applib_export},
+                                  {"addStrokes", applib_addStrokes},
+                                  {"addSplines", applib_addSplines},
+                                  {"getFilePath", applib_getFilePath},
+                                  {"refreshPage", applib_refreshPage},
                                   // Placeholder
                                   //	{"MSG_BT_OK", nullptr},
 


rolandlo:
Suggested edit:
```diff
diff --git a/src/core/plugin/luapi_application.h b/src/core/plugin/luapi_application.h
index b11b3a34e..61283249f 100644
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -124,19 +124,15 @@ static int applib_getFilePath(lua_State* L) {
     // Push another reference to the table on top of the stack (so we know
     // where it is, and this function can work for negative, positive and
     // pseudo indices
-    lua_pushvalue(L, -1);
     // stack now contains: -1 => table
     lua_pushnil(L);
     // stack now contains: -1 => nil; -2 => table
     while (lua_next(L, -2)) {
         // stack now contains: -1 => value; -2 => key; -3 => table
-        // copy the key so that lua_tostring does not modify the original
-        lua_pushvalue(L, -2);
-        // stack now contains: -1 => key; -2 => value; -3 => key; -4 => table
-        const char* value = lua_tostring(L, -2);
+        const char* value = lua_tostring(L, -1);
         formats.push_back(value);
         // pop value + copy of key, leaving original key
-        lua_pop(L, 2);
+        lua_pop(L, 1);
         // stack now contains: -1 => key; -2 => table
     }
     // stack now contains: -1 => table (when lua_next returns 0 it pops the key
@@ -407,19 +403,15 @@ static int applib_drawSplineStroke(lua_State* L) {
     // Push another reference to the table on top of the stack (so we know
     // where it is, and this function can work for negative, positive and
     // pseudo indices
-    lua_pushvalue(L, -1);
     // stack now contains: -1 => table
     lua_pushnil(L);
     // stack now contains: -1 => nil; -2 => table
     while (lua_next(L, -2)) {
         // stack now contains: -1 => value; -2 => key; -3 => table
-        // copy the key so that lua_tostring does not modify the original
-        lua_pushvalue(L, -2);
-        // stack now contains: -1 => key; -2 => value; -3 => key; -4 => table
-        const char* value = lua_tostring(L, -2);
-        coordStream.push_back(std::stod(value));
+        double value = lua_tonumber(L, -1);
+        coordStream.push_back(value);
         // pop value + copy of key, leaving original key
-        lua_pop(L, 2);
+        lua_pop(L, 1);
         // stack now contains: -1 => key; -2 => table
     }
     // stack now contains: -1 => table (when lua_next returns 0 it pops the key
@@ -450,6 +442,7 @@ static int applib_drawSplineStroke(lua_State* L) {
     }
     // If there aren't at least 4 points, then don't add the stroke.
     printf("applib_drawSplineStroke: myStroke shorter than four points. Will discard.\n");
+    delete myStroke;
     myStroke = nullptr;
     return 1;
 }
@@ -473,22 +466,15 @@ static int applib_drawStroke(lua_State* L) {
     // Get the table from the Lua stack
     std::vector<double> coordStream;
 
-    // Push another reference to the table on top of the stack (so we know
-    // where it is, and this function can work for negative, positive and
-    // pseudo indices
-    lua_pushvalue(L, -1);
     // stack now contains: -1 => table
     lua_pushnil(L);
     // stack now contains: -1 => nil; -2 => table
     while (lua_next(L, -2)) {
         // stack now contains: -1 => value; -2 => key; -3 => table
-        // copy the key so that lua_tostring does not modify the original
-        lua_pushvalue(L, -2);
-        // stack now contains: -1 => key; -2 => value; -3 => key; -4 => table
-        const char* value = lua_tostring(L, -2);
-        coordStream.push_back(std::stod(value));
+        double value = lua_tonumber(L, -1);
+        coordStream.push_back(value);
         // pop value + copy of key, leaving original key
-        lua_pop(L, 2);
+        lua_pop(L,1);
         // stack now contains: -1 => key; -2 => table
     }
     // stack now contains: -1 => table (when lua_next returns 0 it pops the key
@@ -504,6 +490,7 @@ static int applib_drawStroke(lua_State* L) {
     }
     myStroke->setWidth(1.5);
     layer->addElement(myStroke);
+    delete myStroke;
     myStroke = nullptr;
     return 0;
 }

```

WillNilges:
Hey, thanks for the review! I'm still not that familiar with Lua, so, admittedly, I got that stack manipulation code off of StackOverflow. I also _fully forgot_ about C++'s `delete` keyword :facepalm: Additionally, I'll remember to run `valgrind` in the future.

Question, though. If we're `delete`ing, then do we need the `nullptr` line?

```diff
+    delete myStroke;
     myStroke = nullptr;
```

rolandlo:
> Hey, thanks for the review! I'm still not that familiar with Lua, so, admittedly, I got that stack manipulation code off of StackOverflow. I also _fully forgot_ about C++'s `delete` keyword facepalm Additionally, I'll remember to run `valgrind` in the future.
> 
> Question, though. If we're `delete`ing, then do we need the `nullptr` line?
> 
> ```diff
> +    delete myStroke;
>      myStroke = nullptr;
> ```

Yes, we need it. Otherwise we get undefined behavior. See [coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#set-pointer-to-nullptr-after-delete)

rolandlo:
An important next step is to decide how the strokes and splines should be passed from Lua to the API function. I'm not quite happy how it is done now. Keep in mind that the API should be used by everyone who wants to add strokes and splines to a Xournal++ document from a Lua plugin. In particular assuming that points have no pressure and that the stroke width is `1.5` is unacceptable. 

Suggestion for `drawStroke` (in final form):

```
stroke = {
x = {[1] = number, [2] = number, ...., [N] = number}
y = {[1] = number, [2] = number, ...., [N] = number}
pressure = {[1] = number, [2] = number, ...., [N] = number}
width = number
color = integer
fill = integer
tool = string
lineStyle = string
}
```
I'm not 100% sure whether it is better to have the x-coordinates, y-coordinates and pressure values separated, or to create a table of points, where each point is a table consisting of x, y-coordinates and pressure-value. The first approach is the one used by the Google IME API and other handwriting recognition software. 

It is not necessary to implement all of that at once, but at least everything that is assigned explicitly (like the stroke width) should be implemented.

WillNilges:
I just had the _opposite_ of a brain blast.

I Refactored my API function to take in a dataset like what you posted, but now I'm getting a really weird crash that seems to happen after the plugin finishes. I've written a test plugin that just hands off 8 points to turn into a stroke while I was working on the refactor.

### TL;DR: I'm not supposed to be `delete`ing the strokes if they're supposed to be added to a layer, should I?

```Lua
-- Register all Toolbar actions and intialize all UI stuff
function initUi()
  ref = app.registerUi({
          ["menu"] = "Stroke Test",
          ["callback"] = "strokeTest",
          ["accelerator"] = "<Control><Alt>r"
  });
  print("StrokeTest registered\n");
end

-- Callback if the menu item is executed
function strokeTest()
-- <stroke tool="pen" ts="0" fn="" color="#000000ff" width="1.50000000">288.50000000 430.40000000 288.34300412 430.65101293 288.19502325 430.86073320 288.04719216 431.03537672 287.89064560 431.18115940 287.71651834 431.30429716 287.51594515 431.41100590 287.28006078 431.50750155</stroke>
--  -- In the current code, the only things that matter are the pressure and the width. Nothing else is used right now
        app.drawStroke({
                ["x"] = {
                        [1] = 101.0,
                        [2] = 102.0,
                        [3] = 103.0,
                        [4] = 104.0,
                        [5] = 105.0,
                        [6] = 106.0,
                        [7] = 107.0,
                        [8] = 108.0
                },
                ["y"] = {
                        [1] = 100.0,
                        [2] = 100.0,
                        [3] = 100.0,
                        [4] = 100.0,
                        [5] = 100.0,
                        [6] = 100.0,
                        [7] = 100.0,
                        [8] = 100.0
                },
                ["pressure"] = {
                        [1] = 0.5,
                        [2] = 0.4,
                        [3] = 0.3,
                        [4] = 0.4,
                        [5] = 0.5,
                        [6] = 0.6,
                        [7] = 0.6,
                        [8] = 0.5
                },
                ["width"] = 1.4,
                ["color"] = 0xff0000,
                ["fill"] = 0,
                ["tool"] = "STROKE_TOOL_PEN",
                ["lineStyle"] = "default"
        })
        app.refreshPage()
        print("done")
end
```

The crash seems to come after I call `page->firePageChanged();` from within `app.refreshPage()`. The reason for this seems to be the `delete myStroke;` line you suggested (I suspect it wasn't crashing before when you and I were testing this a few days ago because the suggested line didn't actually get called when the program ran correctly). **Now that I type that out, it makes perfect sense.**

It was added here
```C
    // Make sure there are enough points in the stroke to not ruin the file
    // (See Xournalpp LoadHandler.cpp:913)
    if (myStroke->getPointCount() >= 4) {
        myStroke->setWidth(1.5);
        layer->addElement(myStroke);
        delete myStroke; // I added this line, and it started crashing
        myStroke = nullptr;
        return 0;
    }
    // If there aren't at least 4 points, then don't add the stroke.
    printf("applib_drawSplineStroke: myStroke shorter than four points. Will discard.\n");
    delete myStroke; // Your addition. Does not crash, because the stroke is not used.
    myStroke = nullptr;
    return 1;
```

WillNilges:
@rolandlo I have some new code to push. Two questions before I do:
1. Do you have a better idea for how to initialize and deal with that `new Stroke();`?
```C++
Stroke* myStroke = new Stroke(); 
```
2. I should only be `delete`ing aforementioned `Stroke` when it's not being used? Such as here:
```C++
    // Make sure there are enough points in the stroke to not ruin the file
    // (See Xournalpp LoadHandler.cpp:913)
    if (myStroke->getPointCount() >= 4) {
        myStroke->setWidth(1.5);
        layer->addElement(myStroke); // Because it's getting added to the layer, I shouldn't delete it.
        myStroke = nullptr;
        return 1;
    }
    // If there aren't at least 4 points, then don't add the stroke.
    g_warning("Stroke shorter than four points. Discarding. (Has %d/4)", myStroke->getPointCount());
    delete myStroke; // But here, since it's useless, I should.
    myStroke = nullptr;
    return 1;
```

rolandlo:
> @rolandlo I have some new code to push. Two questions before I do:
> 
>     1. Do you have a better idea for how to initialize and deal with that `new Stroke();`?
> 
> 
> ```c++
> Stroke* myStroke = new Stroke(); 
> ```
> 
>     2. I should only be `delete`ing aforementioned `Stroke` when it's not being used? Such as here:
> 
> 
> ```c++
>     // Make sure there are enough points in the stroke to not ruin the file
>     // (See Xournalpp LoadHandler.cpp:913)
>     if (myStroke->getPointCount() >= 4) {
>         myStroke->setWidth(1.5);
>         layer->addElement(myStroke); // Because it's getting added to the layer, I shouldn't delete it.
>         myStroke = nullptr;
>         return 1;
>     }
>     // If there aren't at least 4 points, then don't add the stroke.
>     g_warning("Stroke shorter than four points. Discarding. (Has %d/4)", myStroke->getPointCount());
>     delete myStroke; // But here, since it's useless, I should.
>     myStroke = nullptr;
>     return 1;
> ```

You're right. Just don't delete strokes, that are added to a layer. The data they point to is still needed, so it shouldn't be destroyed. Sorry for the confusion.

WillNilges:
Finished adding all the attributes to both the point-based and spline-based functions. But as I was reviewing, I had a thought: Would it be better to make the Spline-based function simply compute the Splines, then pass those back to the Lua stack and call the Point-based function, thereby reducing code duplication?

OR

Could I unify the functions by somehow making it detect a Spline array vs a Point Array?

The splines are very important to me (obviously), but I'm basically wondering if I should integrate the functionality into the one function, or if I should treat it more as an "addon," or if I should continue to treat it as an entirely separate function, and just deal with the code duping?

WillNilges:
I did it both ways. Check the last two commits.  Both work pretty well, but I think rolling the spline function into the other one would make it easier to maintain. It just has to be well documented that you can pass a table of splines _or_ two tables of points.

Perhaps we could use X and Y tables for both, and then pass a separate boolean or something that dictates the use of splines vs strokes, rather than having to pass `["splines"]` or `["x"]` and `["y"]` .

rolandlo:
> I did it both ways. Check the last two commits. Both work pretty well, but I think rolling the spline function into the other one would make it easier to maintain. It just has to be well documented that you can pass a table of splines _or_ two tables of points.
> 
> Perhaps we could use X and Y tables for both, and then pass a separate boolean or something that dictates the use of splines vs strokes, rather than having to pass `["splines"]` or `["x"]` and `["y"]` .

I would prefer to keep it separate, since it looks more user friendly (the user either wants to have splines or strokes and most likely not mix it). You can still introduce some helper functions in the `luapi_application.h` that take care of the code duplication. 
Note that the implementation of `applib_drawSplineStroke` will not use rasterization anymore when the file format supports splines (PR #2729). 

Some comments on the code (for commit 0b2c727)
- I would suggest to rename `drawStroke` to `addStroke` and `drawSplineStroke` to `addSpline`, since it is more about adding a stroke/spline to the layer than to draw it (and SplineStroke seems unnecessarily complicated)
- It would be nice to have most fields optional with default values
   - lineStyle should default to "solid"
   - if no "pressure" key is given, pressure for all points should be `Point::NO_PRESSURE`
   - tool should default to "pen"
   - fill should default to -1 (i.e. no fill)   
   - color should default to `#000000` (black color)
  Use `luaL_optstring` (and similar) for that purpose.
- I think strokes with 2 and 3 points are fine. Only those with less than 2 points are bad. The LoadHandler complains if it reads less than 4 coordinates, not points. Strokes consisting of a single point are handled as strokes with a doubled point in Xournal++.
- Please make sure to handle the (bad) case when the user passes coordinate lists, where there are not an equal number of x-coordinates and y-coordinates
- Please include an example of how to pass a spline
- I'm not 100% certain that the current implementation using a coordinates stream for passing splines is the best solution. Admittedly using 8 different tables for `startX`, `startY`, `ctrl1X`, `ctrl1Y`, `ctrl2X`, `ctrl2Y`, `endX`, `endY` would look a bit cumbersome.

WillNilges:
Sorry for the delay. I split the functions again and added a helper function. I figure when #2729 gets merged, I'll update the Spline function to leverage that change. I also added the documentation and default params, along with some error handling.

src/core/plugin/luapi_application.h
```diff
@@ -103,6 +104,62 @@ static int applib_saveAs(lua_State* L) {
...
+ *
+ * Examples:
+ *   path = app.getFilePath()
+ *   path = app.getFilePath({'*.bmp', '*.png'})
+ */
+static int applib_getFilePath(lua_State* L) {
+    GtkFileChooserNative* native =
+            gtk_file_chooser_native_new(_("Open file"), nullptr, GTK_FILE_CHOOSER_ACTION_OPEN, nullptr, nullptr);
+    gint res;
+    int args_returned = 1;  // change to 1 if user chooses file
```
rolandlo, line 119:
```suggestion
    int args_returned = 0;  // change to 1 if user chooses file
```

WillNilges: Why change this?

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+ * Notifies program of any updates to the working document caused
+ * by the API.
+ *
+ * Example: app.refreshPage()
+ */
+static int applib_refreshPage(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* ctrl = plugin->getControl();
+    PageRef const& page = ctrl->getCurrentPage();
+    page->firePageChanged();
```
rolandlo, line 640: Please emit a warning if there is no page (and don't execute `page->firePageChanged`).

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+            g_warning("%s", FC(_F("Unknown stroke type: \"{1}\", assuming pen") % tool));
+        }
+        LineStyle lineStyle = StrokeStyle::parseStyle(style);
+        stroke->setLineStyle(lineStyle);
+        if (strcmp("", style) != 0 && strcmp("dash", style) != 0 && strcmp("dashdot", style) != 0 &&
+            strcmp("dot", style) != 0)
+            g_warning("%s", FC(_F("Unknown style: \"{1}\", assuming solid") % style));
+        layer->addElement(stroke);
+        stroke = nullptr;
+        return 1;
```
rolandlo, line 433: If you make the return type `void` you will also have to change the return statements (here and below).
```suggestion
        return;
```

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+    }
+    // If there aren't at least 2 points, then don't add the stroke.
+    g_warning("Stroke shorter than two points. Discarding. (Has %d/2)", stroke->getPointCount());
+    delete stroke;
+    stroke = nullptr;
+    return 1;
```
rolandlo, line 439: Here as well.
```suggestion
    return;
```

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+        Point end = Point(coordStream.at(i + 6), coordStream.at(i + 7), Point::NO_PRESSURE);
+        i += 8;
+        SplineSegment segment = SplineSegment(start, ctrl1, ctrl2, end);
+        std::list<Point> raster = segment.toPointSequence();
+        for (Point point: raster) stroke->addPoint(point);
+        // TODO: (willnilges) Is there a way we can get Pressure with Splines?
+    }
+    // Finish building the Stroke and apply it to the layer.
+    addStrokeHelper(L, stroke, layer);
+    return 1;
```
rolandlo, line 522: The number of return values is 0, isn't it?
```suggestion
    return 0;
```

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+        return 1;
+    }
+    // Add points to the stroke
+    for (long unsigned int i = 0; i < xStream.size(); i++) {
+        Point myPoint = Point(yStream.at(i), xStream.at(i), pressureStream.at(i));
+        stroke->addPoint(myPoint);
+    }
+    // Finish building the Stroke and apply it to the layer.
+    addStrokeHelper(L, stroke, layer);
+    return 1;
```
rolandlo, line 627: Return the number of returned values (which appears to be 0).
```suggestion
    return 0;
```

rolandlo (changes requested):
We're getting closer. I have some further comments (see code).

src/core/plugin/luapi_application.h
rolandlo: Each API function should return the number of return values (that are pushed onto the Lua stack). If the user chooses no file, there is no filename pushed onto the Lua stack.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+    // Make sure we have enough points to form a stroke
+    if (stroke->getPointCount() >= 2) {
+        // Get the rest of the attributes. Default to something if none is provided
+        // TODO: (willnilges) Test this. Is this robust? Should we be doing this one field at a time?
+        lua_getfield(L, 1, "width");
+        lua_getfield(L, 1, "color");
+        lua_getfield(L, 1, "fill");
+        lua_getfield(L, 1, "tool");
+        lua_getfield(L, 1, "lineStyle");
+        double width = luaL_optnumber(L, -5, 1.5);
```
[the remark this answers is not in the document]
WillNilges: Ah, thanks! Didn't realize you could do that. 1.5 was chosen out of ignorance :sweat_smile:

Will change :+1:

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+        stroke->setLineStyle(lineStyle);
+        if (strcmp("", style) != 0 && strcmp("dash", style) != 0 && strcmp("dashdot", style) != 0 &&
+            strcmp("dot", style) != 0)
+            g_warning("%s", FC(_F("Unknown style: \"{1}\", assuming solid") % style));
+        layer->addElement(stroke);
+        stroke = nullptr;
+        return 1;
+    }
+    // If there aren't at least 2 points, then don't add the stroke.
+    g_warning("Stroke shorter than two points. Discarding. (Has %d/2)", stroke->getPointCount());
```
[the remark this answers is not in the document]
WillNilges: It was meant to be "This stroke has only 1 out of 2 points." I can remove it if you think that's too verbose.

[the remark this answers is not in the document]
rolandlo: Got it, I misunderstood. Please remove it.

src/core/plugin/luapi_application.h
WillNilges: The reason for the refreshPage function is to update the document so that the strokes added actually show up. See this comment in the issue for context.
https://github.com/xournalpp/xournalpp/issues/2850#issuecomment-950342758

Is there another, better way to fix this problem?

rolandlo: I think you do it correctly. Just the case where there is no page (so that `page->firePageChanged()` would lead to a segfault) should be treated separately.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+    delete stroke;
+    stroke = nullptr;
+    return 1;
+}
+
+/**
+ * Given a table containing a series of splines, draws a stroke on the canvas.
+ * Expects a table of coordinate pairs along with attributes of the stroke.
+ * Attributes are optional.
```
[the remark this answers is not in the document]
WillNilges: Going to add this to the documentation.
```
The function expects 8 points per spline. Due to the nature of quadratic                                                                                
splines, you must pass your points in a repeating pattern:
startX, startY, ctrl1X, ctrl1Y, ctrl2X, ctrl2Y, endX, endY, startX, startY, ...

The function checks that the spline table is divisible by eight, and will throw
an error if it is not.       
```

src/core/plugin/luapi_application.h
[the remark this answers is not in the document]
WillNilges: Now the color and size, if no others are passed, are based off of the currently selected options. Thought that might be a nice touch.

I was thinking that maybe, by default, everything should be based off of current options.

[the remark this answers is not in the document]
WillNilges: The helper will now check what the current pen thickness and color are, and apply those options if there were none provided.

Going further, I think it might be best to just default entirely to the current pen options.

WillNilges:
I think at this point, we're "reviewing," right?

WillNilges:
It's also worth noting that school has started up for me again, so apologies if I take a while to do things.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +384,263 @@ static int applib_layerAction(lua_State* L) {
...
+    Control* ctrl = plugin->getControl();
+    PageRef const& page = ctrl->getCurrentPage();
+    Layer* layer = page->getSelectedLayer();
+    Stroke* stroke = new Stroke();
+
+    // Discard any extra arguments passed in
+    lua_settop(L, 1);
+    luaL_checktype(L, 1, LUA_TTABLE);
+
+    lua_getfield(L, 1, "splines");
```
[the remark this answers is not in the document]
rolandlo: Why use the plural form then? The array of numbers build up one spline (but multiple segments).

src/core/plugin/luapi_application.h
```diff
@@ -385,28 +379,35 @@ static int applib_layerAction(lua_State* L) {
...
-        // TODO: (willnilges) Test this. Is this robust? Should we be doing this one field at a time?
         lua_getfield(L, 1, "width");
         lua_getfield(L, 1, "color");
         lua_getfield(L, 1, "fill");
         lua_getfield(L, 1, "tool");
         lua_getfield(L, 1, "lineStyle");
-        double width = luaL_optnumber(L, -5, 1.5);
-        long long int color = luaL_optinteger(L, -4, 0);
+        double width = luaL_optnumber(L, -5, currentPenSize);
+        Color color = lua_isinteger(L, -4) ? Color(lua_tointeger(L, -4)) : currentColor;
```
rolandlo, line 403: Since there is a "tool" field, the user may not only add pen strokes, but also highlighter or eraser strokes. It looks wrong to use e.g. the current pen size for a highlighter stroke. You should use the specific values for the highlighter tool when `tool=="highlighter"`. Similarly with eraser strokes (and pen strokes).

rolandlo:
> It's also worth noting that school has started up for me again, so apologies if I take a while to do things.

No problem, take your time.

src/core/plugin/luapi_application.h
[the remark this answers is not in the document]
rolandlo: > Re: how to pass this data in the first place I've thought about it some more, and what if we made a struct? One probably already exists in the Spline PR, right? It would be annoying, but not difficult to pass a table of tables. That might be cleaner and perhaps easier to understand for a first time user. It would also make it more logical to add cubic splines and the like. I don't have any problem with the current approach, but it's not super extensible, and is entirely dependent on the interpretation done by the function.

I have thought about it, too. The downside of passing a table of tables (consisting of 8 coordinates each) is, that it takes more computation time to check, whether all individual tables are valid. Moreover it will also take more memory to store/pass it like that. So probably it is better to keep it as one single table consisting of coordinates. For the user it is not that a big difference, I think. Efficiency is more important.

rolandlo:
> I think at this point, we're "reviewing," right?

Yes, I would call it like that.

src/core/plugin/luapi_application.h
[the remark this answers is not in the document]
WillNilges: Maybe I should say that it's an array of numbers that, when assembled properly, form **some number of splines**. A `SplineSegment` is just a spline, right? And if you can pass multiple splines in one go (each spline you pass gets added to the same stroke, though), then I think "Splines" fits.

src/core/plugin/luapi_application.h
WillNilges: We talked about this earlier and came up with[ these sets of defaults.](https://github.com/xournalpp/xournalpp/pull/3688#issuecomment-1003541104) I've been playing around with this, trying to figure out how to use the active tool's settings as "defaults" of sorts because it seems like that'd be way better. To do that, though, I think we'd need to require a tool parameter as input.

I'm thinking it goes like this:
If the user provides no options, then the current active settings are used.
If the user provides a tool and no options, then that tool's current settings are used.
If the user provides a tool and any number of options, then their provided options override the current tool's settings.
If the user provides no tool and any number of options, **the current active settings are used.**

WillNilges: Basically, you need to provide a tool, or get the default settings

rolandlo: > I'm thinking it goes like this: 
If the user provides no options, then the current active settings are used. 
If the user provides a tool and no options, then that tool's current settings are used. 
If the user provides a tool and any number of options, then their provided options override the current tool's settings. 
If the user provides no tool and any number of options, **the current active settings are used.**

Using the "current active settings" has the problem, that the current active tool can be something else than pen, highlighter or eraser (e.g. the "hand" tool). With that in mind, it may be wiser to assume tool "pen" when no tool is specified, since it will be the most used one by far.

WillNilges: Works for me. Let me see what I can do. I'm looking at how `getToolInfo` works (and pretty much just stealing the logic).

WillNilges: So I've just about got a solution, but for whatever reason, when I set the stroke type to Eraser, instead of erasing the stroke, this happens:
image

Not sure what I'm doing wrong. I set the stroke type as eraser, get the width from the current settings, but still it just draws black.

WillNilges: I pushed my code if you want to take a look.

WillNilges:
Update to above commit: Highlighter seems to work.

src/core/plugin/luapi_application.h
rolandlo: > So I've just about got a solution, but for whatever reason, when I set the stroke type to Eraser, instead of erasing the stroke, this happens: image
> 
> Not sure what I'm doing wrong. I set the stroke type as eraser, get the width from the current settings, but still it just draws black.

We can only add eraser strokes in "whiteout" mode. These strokes are like pen strokes, just in White color and thick (by default). You must specify the color.
See here for the difference in the saved file. The first two eraser strokes came from a plugin. The last one was hand-drawn (in whiteout mode).

```xml
<stroke tool="eraser" color="#000000ff" width="2.00000000">150.00000000 100.00000000 154.38421326 108.76842652</stroke>
<stroke tool="eraser" color="#000000ff" width="2.00000000">170.48390270 140.96780539 200.00000000 200.00000000 80.00000000 300.00000000 300.00000000 100.00000000</stroke>
<stroke tool="eraser" color="#ffffffff" width="12.00000000">237.76500567 88.88411427 238.25880631 88.63721395 238.75260694 88.14341332 241.96231107 86.66201141 245.17201519 85.92131046 249.12242027 84.43990856 256.77633011 81.47710475 261.46743614 79.50190221 266.15854217 
77.52669967 273.81245201 74.56389586 276.03455487 73.82319491 278.75045836 72.83559364 279.49115931 72.58869332 279.98495995 72.34179300 281.21946154 71.84799237 281.46636185 71.60109205 281.96016249 71.60109205 282.94776376 71.60109205</stroke>
```

WillNilges: Why is that? I haven't actually delved too deeply into the eraser source code yet, so I don't know how the normal eraser works. I'll fix what I have in the meantime.

rolandlo: - The only eraser type that adds a stroke to the layer, is the "whiteout" mode.
- In "delete stroke" mode each stroke touched by the eraser is removed from the layer.
- In "standard" mode each stroke touched by the eraser is cut into two smaller pieces, that replace the original stroke.

It would be possible to use the EraseHandler's `erase` function to erase at each point  from a list of points (like for a stroke) passed via the Lua plugin API. That has nothing to do with `addStroke` though, and should not be in this PR.

WillNilges: Gotcha, thanks.

So you think that those `erase` functionalities should be in other functions (and hence PRs)? Seems a bit counterintuitive to split up eraser functionality into different API calls, since those functions are grouped the same way and used the same way. Personally, I'd either add another field to this function, or split all eraser functionality into an `eraseStroke` function, including Whiteout.

rolandlo: In that case it is better to split all eraser functionality into an `erase` function (including Whiteout, which will probably not be used much anyway).

rolandlo: I will have to test the new API with all options, so there could be more things that need change, but right now there is nothing else I want you to change.

WillNilges:
Removed the eraser code. Will open a new PR for that sometime this week hopefully.

src/core/plugin/luapi_application.h
WillNilges: I don't know if this would be helpful, but I wrote a simple Lua plugin to test the Stroke API (requires visual inspection of course :P but maybe it could be expanded to save the file then parse the contents?)

https://github.com/WillNilges/inkpath/tree/5d55e501b455d4667ba62d3c26ccc9256ecb0006/HACKING/StrokeTest

rolandlo:
> Removed the eraser code. Will open a new PR for that sometime this week hopefully.

I'm trying to figure out if the eraser functionality will be used in the plugin API. Do you already see an application for that?

WillNilges:
> > Removed the eraser code. Will open a new PR for that sometime this week hopefully.
> 
> I'm trying to figure out if the eraser functionality will be used in the plugin API. Do you already see an application for that?

I could see someone wanting to do _something_ with it. Maybe some kind of color-coding plugin where red strokes are removed or something? Regardless, I think any work towards allowing Xournal++ to be controlled completely programmatically is a good thing.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +378,335 @@ static int applib_layerAction(lua_State* L) {
...
+    // Finish building the Stroke and apply it to the layer.
+    addStrokeHelper(L, stroke, layer);
+    return 0;
+}
+
+/**
+ * Given a set of points, draws a stroke on the canvas.
+ * Expects three tables of equal length: one for X, one for Y, and one for
+ * stroke pressure, along with attributes of the stroke.
+ * Attributes are optional. Pressure is optional.
```
rolandlo, line 579: Please add a comment on what default values are used.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +378,335 @@ static int applib_layerAction(lua_State* L) {
...
+ * Attributes are optional.
+ *
+ * The function expects 8 points per spline. Due to the nature of quadratic
+ * splines, you must pass your points in a repeating pattern:
+ * startX, startY, ctrl1X, ctrl1Y, ctrl2X, ctrl2Y, endX, endY, startX, startY, ...
+ *
+ * The function checks that the spline table is divisible by eight, and will throw
+ * an error if it is not.
+ *
+ * Example: app.addStroke({
```
rolandlo, line 504:
```suggestion
 * Example: app.addSpline({
```

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +378,335 @@ static int applib_layerAction(lua_State* L) {
...
+ * Expects a table of coordinate pairs along with attributes of the stroke.
+ * Attributes are optional.
+ *
+ * The function expects 8 points per spline. Due to the nature of quadratic
+ * splines, you must pass your points in a repeating pattern:
+ * startX, startY, ctrl1X, ctrl1Y, ctrl2X, ctrl2Y, endX, endY, startX, startY, ...
+ *
+ * The function checks that the spline table is divisible by eight, and will throw
+ * an error if it is not.
+ *
```
rolandlo, line 503: Please add a comment on optional values and defaults.

rolandlo (changes requested):
I have tested the API for adding strokes and splines. It works fine.  The behaviour for using defaults looks pretty good.

See my suggestions for some minor improvements. It shouldn't be much work to implement them.

WillNilges:
There's a couple more comments I need to work out locally—please hold.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +378,335 @@ static int applib_layerAction(lua_State* L) {
...
+        // Set line style
+        if (lua_isstring(L, -1)) // Check if line style settings were provided
+            stroke->setLineStyle(StrokeStyle::parseStyle(lua_tostring(L, -1)));
+        else
+            stroke->setLineStyle(StrokeStyle::parseStyle(lineStyle.data()));
+
+        lua_pop(L, 5); // Finally done with all that Lua data.
+
+        // Add the Stroke
+        layer->addElement(stroke);
```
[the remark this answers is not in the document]
WillNilges: I don't suppose there's a way to allow a whole bunch of strokes to be counted as one "undoing" action?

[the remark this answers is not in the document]
rolandlo: What people want to do depends very much on the use case. If you add only a small number of strokes/splines, it may be more convenient, if all of them can be undone individually. In your use case of image transcription undoing the whole thing at once would be better.
I can't follow your link (is it valid?), but yes, `InsertsUndoAction` already supports undoing/redoing a bunch of elements at once. The only reasonable way of using that would be to pass all splines/strokes in a huge table 
```Lua
{
strokes = {stroke1, stroke2, ... }
allowUndoRedoAction = "together"      -- (or "individual" or "none")
}
```
(similarly with splines) where each stroke/spline in the table has the same shape as before, just without the `allowUndoRedoAction` key.
The functions would have to be renamed to `addStrokes` and `addSplines`, respectively.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +379,347 @@ static int applib_layerAction(lua_State* L) {
...
+    }
+    // If there aren't at least 2 points, then don't add the stroke.
+    g_warning("Stroke shorter than two points. Discarding. (Has %d)", stroke->getPointCount());
+    delete stroke;
+    stroke = nullptr;
+    return;
+}
+
+/**
+ * Given a table containing a series of splines, draws a stroke on the canvas.
```
rolandlo, line 506:
```suggestion
 * Given a table containing a series of spline segments, draws a stroke on the canvas.
```

src/core/plugin/luapi_application.h
rolandlo: I wanted you to comment, that the defaults come from the current settings of the pen/highlighter tool, if none are specified.
It should also be added, that only "pen" or "highlighter" are allowed for the tool key and that "pen" is default.

src/core/plugin/luapi_application.h
```diff
@@ -327,6 +379,347 @@ static int applib_layerAction(lua_State* L) {
...
+}
+
+/**
+ * Given a set of points, draws a stroke on the canvas.
+ * Expects three tables of equal length: one for X, one for Y, and one for
+ * stroke pressure, along with attributes of the stroke.
+ *
+ * Required Arguments: X, Y
+ * Optional Arguments: pressure, tool, width, color, fill, lineStyle
+ *
```
rolandlo, line 595: Same here: Tell what defauls are used if optional arguments are unspecified.

rolandlo (approved):
Please decide whether you want to allow passing multiple strokes/splines, so that they can be undone all at once. For me the current implementation is fine, but if you feel that it is important to allow undoing all strokes/splines at once, then go ahead.
Other than that I have added only very few minor comments on improving the documentation. Functionality it looks all fine now.

WillNilges:
Sick, thanks for the feedback!

I think I figured out how to allow multiple strokes to be passed in at once. I'm thinking that I'll just include both functions. I'll commit in a min.

WillNilges:
Getting a weird error when trying to make an `InsertsUndoAction`. I'm not really sure what the problem is, but my attention is drawn to the second `note`.

```
/mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:80: note:   no known conversion for argument 3 from 'vector<Stroke*>' to 'vector<Element*>'
```


```
/usr/include/c++/10/bits/unique_ptr.h: In instantiation of 'typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = InsertsUndoAction; _Args = {const std::shared_ptr<XojPage>&, Layer*&, std::vector<Stroke*, std::allocator<Stroke*> >&}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<InsertsUndoAction, std::default_delete<InsertsUndoAction> >]':
/mnt/xopp_dev/xournalpp/src/core/plugin/luapi_application.h:701:85:   required from here
/usr/include/c++/10/bits/unique_ptr.h:962:30: error: no matching function for call to 'InsertsUndoAction::InsertsUndoAction(const std::shared_ptr<XojPage>&, Layer*&, std::vector<Stroke*>&)'
  962 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /mnt/xopp_dev/xournalpp/src/core/plugin/luapi_application.h:29,
                 from /mnt/xopp_dev/xournalpp/src/core/plugin/Plugin.cpp:16:
/mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:5: note: candidate: 'InsertsUndoAction::InsertsUndoAction(const PageRef&, Layer*, std::vector<Element*>)'
   38 |     InsertsUndoAction(const PageRef& page, Layer* layer, std::vector<Element*> elements);
      |     ^~~~~~~~~~~~~~~~~
/mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:80: note:   no known conversion for argument 3 from 'vector<Stroke*>' to 'vector<Element*>'
   38 |     InsertsUndoAction(const PageRef& page, Layer* layer, std::vector<Element*> elements);
      |                                                          ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
/mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:36:7: note: candidate: 'InsertsUndoAction::InsertsUndoAction(const InsertsUndoAction&)'
   36 | class InsertsUndoAction: public UndoAction {
      |       ^~~~~~~~~~~~~~~~~
/mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:36:7: note:   candidate expects 1 argument, 3 provided
```

rolandlo:
> Getting a weird error when trying to make an `InsertsUndoAction`. I'm not really sure what the problem is, but my attention is drawn to the second `note`.
> 
> ```
> /mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:80: note:   no known conversion for argument 3 from 'vector<Stroke*>' to 'vector<Element*>'
> ```
> 
> ```
> /usr/include/c++/10/bits/unique_ptr.h: In instantiation of 'typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = InsertsUndoAction; _Args = {const std::shared_ptr<XojPage>&, Layer*&, std::vector<Stroke*, std::allocator<Stroke*> >&}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<InsertsUndoAction, std::default_delete<InsertsUndoAction> >]':
> /mnt/xopp_dev/xournalpp/src/core/plugin/luapi_application.h:701:85:   required from here
> /usr/include/c++/10/bits/unique_ptr.h:962:30: error: no matching function for call to 'InsertsUndoAction::InsertsUndoAction(const std::shared_ptr<XojPage>&, Layer*&, std::vector<Stroke*>&)'
>   962 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
>       |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> In file included from /mnt/xopp_dev/xournalpp/src/core/plugin/luapi_application.h:29,
>                  from /mnt/xopp_dev/xournalpp/src/core/plugin/Plugin.cpp:16:
> /mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:5: note: candidate: 'InsertsUndoAction::InsertsUndoAction(const PageRef&, Layer*, std::vector<Element*>)'
>    38 |     InsertsUndoAction(const PageRef& page, Layer* layer, std::vector<Element*> elements);
>       |     ^~~~~~~~~~~~~~~~~
> /mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:38:80: note:   no known conversion for argument 3 from 'vector<Stroke*>' to 'vector<Element*>'
>    38 |     InsertsUndoAction(const PageRef& page, Layer* layer, std::vector<Element*> elements);
>       |                                                          ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
> /mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:36:7: note: candidate: 'InsertsUndoAction::InsertsUndoAction(const InsertsUndoAction&)'
>    36 | class InsertsUndoAction: public UndoAction {
>       |       ^~~~~~~~~~~~~~~~~
> /mnt/xopp_dev/xournalpp/src/core/undo/InsertUndoAction.h:36:7: note:   candidate expects 1 argument, 3 provided
> ```

The reason seems clear: 
```shell
no known conversion for argument 3 from 'vector<Stroke*>' to 'vector<Element*> 
```
Just define `strokes` as a `vector<Element*>`.

WillNilges:
Fixed it. Thanks!

WillNilges:
Thoughts on changing `addStrokes` to take in a singular table of tables? Each table represents a stroke, and the values in the table alternate `X, Y, pressure, X, Y, pressure, ...`. That would be more in line with how `addSplines` works, and would also be a bit easier to program.

rolandlo:
> Thoughts on changing `addStrokes` to take in a singular table of tables? Each table represents a stroke, and the values in the table alternate `X, Y, pressure, X, Y, pressure, ...`. That would be more in line with how `addSplines` works, and would also be a bit easier to program.

That does not sound like a good idea to me. The interpretation of the list would change when not specifying pressure (which should remain optional).

rolandlo:
> Fixed it. Thanks! I still need to write documentation and do the raster API.

Seems you changed the way to specify a single spline. Options like width, color etc. should NOT be assumed to be the same for all splines. Only "allowUndoRedoAction" should be common. A spline will consist of several spline segments and I would advise you to keep the "coordinates" key for the spline and not make a table of segments.

Also the `addSplines` and `addStrokes` should be enough. No need to keep `addSpline` and `addStroke`.

WillNilges:
> > Fixed it. Thanks! I still need to write documentation and do the raster API.
> 
> Seems you changed the way to specify a single spline. Options like width, color etc. should NOT be assumed to be the same for all splines. Only "allowUndoRedoAction" should be common. A spline will consist of several spline segments and I would advise you to keep the "coordinates" key for the spline and not make a table of segments.
> 
> Also the `addSplines` and `addStrokes` should be enough. No need to keep `addSpline` and `addStroke`.

Hold on, I think there's some confusion between how we both understand what this function is supposed to do, and I think it's because I mislabeled some variables. Right now, `addSplines` is supposed to pass an entire batch of strokes in at once, through a big table of tables. Each inner table contained a set of SplineSegments. I guess the function that I wrote today should be called `addStrokes()` or `addSplineStrokes()` or even `addStrokesFromSplines()`.

Here's how that table of tables should look
```
strokes = { // So the outer table is a table of strokes
    { // Each inner table is a stroke
        [1] = 880.0, // Every eight pairs is a SplineSegment
        [2] = 874.0,
        [3] = 881.3295,
        [4] = 851.5736,
        [5] = 877.2915,
        [6] = 828.2946,
        [7] = 875.1697,
        [8] = 806.0,
        ... // A Stroke can be made up of as many SplineSegments as is necessary.
    },
    ... // You can pass as many independent strokes into this function as you want.
}
```

 Previously, the way that my plugin passed each stroke was by calling `addSpline` which accepted a table of length 8n. The data in the table was coordinates for SplineSegments. To transcribe the whole image, I would call this `addSpline` function over and over and over again, passing in a single stroke (which was composed of spline segments) each time. Because of this, I couldn't add them all to an `InsertsUndoAction`.

So I think what I want is a function to pass a batch of strokes that are composed of splines to Xournalpp (which is what I have now). `addSpline` still does them separately like you want, but has the aforementioned problem. I can't think of a way to do what you're asking, unless you can somehow return a stroke pointer to Lua?

Does that make sense? I should do some refactoring to help clarify this.

WillNilges:
> (similarly with splines) where each stroke/spline in the table has the same shape as before, just without the allowUndoRedoAction key.

I think I might see what you're getting at. Would each stroke have width, color, etc settings specified?

rolandlo:
> > (similarly with splines) where each stroke/spline in the table has the same shape as before, just without the allowUndoRedoAction key.
> 
> I think I might see what you're getting at. Would each stroke have width, color, etc settings specified? Basically, you'd be sending a big table containing multiple iterations of this?
> 
> ```
>  * Example: app.addSplines({
>  *            ["coordinates"] = {
>  *              [1] = 880.0,
>  *              [2] = 874.0,
>  *              [3] = 881.3295,
>  *              [4] = 851.5736,
>  *              [5] = 877.2915,
>  *              [6] = 828.2946,
>  *              [7] = 875.1697,
>  *              [8] = 806.0,
>  *              [...] = ...,
>  *            },
>  *            ["width"] = 1.4,
>  *            ["color"] = 0xff000000,
>  *            ["fill"] = 0,
>  *            ["tool"] = "pen",
>  *            ["lineStyle"] = "solid"
>  *        })
> ```
> 
> That could work, I suppose.
> 
> **Edit:** correct names.

Yes, that's what I suggested. When your image transcription finally works fully, the splines will also have different color, width etc. Also imagine a plugin that imports from a different file format. Then it would also need to batch add multiple splines/strokes with different properties.

WillNilges:
Wow this is really difficult.

I can't figure out how to pass a table to C, get that table onto the stack, and then parse values in that table. I feel like it should be as simple as getting other fields out of the table, like i do in addBatchStrokesFromSplines, but I just can't figure it out.

WillNilges:
Wait hold on

WillNilges:
Alright, there we go. Something more in line with that?

rolandlo:
> Alright, there we go. Something more in line with that?

Yes!
> 
> I still need to do `addStrokes`.

Go ahead.

WillNilges:
Done. Thoughts?

WillNilges:
(sneaky update w/ master)

rolandlo:
> (sneaky update w/ master)
> 
> Do you want me to rebase this at all?

You shouldn't have merged master into it. That creates a huge mess. Please undo the merge commit. Then rebase on upstream/master and solve the merge conflict. If you need help with it, let me know.

rolandlo:
> Fixed the conflicts, do you want them all squashed?

Eventually yes, I think the commits should all be squashed into one commit. Please wait with squashing until I review the latest changes though.

rolandlo:
> > Fixed the conflicts, do you want them all squashed?
> 
> Eventually yes, I think the commits should all be squashed into one commit. Please wait with squashing until I have reviewed the latest changes though.

rolandlo:
Please fix the error reported from Azure Pipelines:

```
/home/vsts/work/1/s/src/core/plugin/luapi_application.h: At global scope:
/home/vsts/work/1/s/src/core/plugin/luapi_application.h:1721:50: error: ‘applib_drawStroke’ was not declared in this scope
                                   {"drawStroke", applib_drawStroke},
                                                  ^~~~~~~~~~~~~~~~~
/home/vsts/work/1/s/src/core/plugin/luapi_application.h:1721:50: note: suggested alternative: ‘applib_addStrokes’
                                   {"drawStroke", applib_drawStroke},
                                                  ^~~~~~~~~~~~~~~~~
                                                  applib_addStrokes
/home/vsts/work/1/s/src/core/plugin/luapi_application.h:1722:49: error: ‘applib_addStroke’ was not declared in this scope
                                   {"addStroke", applib_addStroke},
                                                 ^~~~~~~~~~~~~~~~
/home/vsts/work/1/s/src/core/plugin/luapi_application.h:1722:49: note: suggested alternative: ‘applib_addStrokes’
                                   {"addStroke", applib_addStroke},
                                                 ^~~~~~~~~~~~~~~~
                                                 applib_addStrokes

```

WillNilges:
Oops, that must've been an artifact of the rebase. Sorry about that.

rolandlo (changes requested):
As I have noticed there are bugs reading the coordinates for strokes (see comments in the code). That must be fixed. Please check the result carefully.

rolandlo:
Suggested edit:
```diff
diff --git a/src/core/plugin/luapi_application.h b/src/core/plugin/luapi_application.h
index 795d74a01..ec6ed245d 100644
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -536,14 +536,19 @@ static int applib_addSplines(lua_State* L) {
     lua_getfield(L, 1, "strokes");
     if (!lua_istable(L, -2))
         luaL_error(L, "Missing coordinate table!");
-    lua_pushnil(L);
-    while (lua_next(L, -2)) {
+
+    size_t numSplines = lua_rawlen(L, -1);
+    for (size_t a = 1; a<=numSplines; a++) {
         std::vector<double> coordStream;
         Stroke* stroke = new Stroke();
         // Get coordinates
+        lua_pushnumber(L, a);
+        lua_gettable(L, -2);
         lua_getfield(L, -1, "coordinates");
-        lua_pushnil(L);
-        while (lua_next(L, -2)) {
+        size_t numCoords = lua_rawlen(L, -1); 
+        for (size_t b = 1; b<=numCoords; b++) {
+            lua_pushnumber(L, b);
+            lua_gettable(L, -2);
             double point = lua_tonumber(L, -1);
             coordStream.push_back(point);  // Each segment is going to have multiples of 8 points.
             lua_pop(L, 1);

```

WillNilges:
Nice, thank you. Odd that I've never seen `addSplines` behave badly (that's the one I've tested the most :sweat:), and that `addStrokes` misbehaves in the same way every time (I managed to reproduce the bug). I figured the issue had something to do with how those `while` loops worked, but I couldn't find any time today to play with it. I'll see if I can implement the fix tomorrow.

rolandlo (changes requested):
I found some mistakes with the error messages (see code). Also please rename "strokes" to "splines" in the `addSplines` function.

rolandlo (approved):
LGTM. In case you don't intend to change anything on the code anymore, can you please squash all commits into one now?
Also please update the description in the first post of the Conversation.

WillNilges:
Alrighty, done. How does that look?

rolandlo:
Looks like your last commit(s) (warnings and renaming) weren't included in the squashed commit. 
Also please remove all lines with `drawingType` in `addStrokeHelper`. They are not needed anywhere. I forgot to mention that.

WillNilges:
Oops, fixed that.

rolandlo (approved):

rolandlo:
Thanks @WillNilges for the great work! I think this addition to the plugin API will prove very useful. It wouldn't be hard to implement a function plotter plugin, for example. Also importing from a different file format would be possible. Wish you all the best with the image transcription. Merging now.

PR #3689, opened 2021-12-30
update macos build scripts

rolandlo:
Updating the build scripts for poppler (including openjpeg), libzip and portaudio, so they use the latest version.

### DIFF
--- a/mac-setup/build-libzip.sh
+++ b/mac-setup/build-libzip.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -x
 export PATH="$HOME/.new_local/bin:$PATH"
 
-curl -L https://libzip.org/download/libzip-1.5.2.tar.gz -o libzip.tar.gz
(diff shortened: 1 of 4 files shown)


LittleHuba:
Was this tested on your side @rolandlo? If so, feel free to merge this right away.

rolandlo:
Yes, it was tested, also with the once more updated poppler version. I will push some according changes to the [pipeline dependency script](https://github.com/xournalpp/xournalpp-pipeline-dependencies/blob/base/gtk/package-gtk-bin.sh) as well. Since I have an updated Gtk blob for MacOS ready, that should work for MacOS Monterey as well (it has Gtk 3.24.30 + [patches](https://gitlab.gnome.org/GNOME/gtk-osx/-/blob/master/modulesets-stable/gtk-osx.modules#L301-302) that have been applied in the meantime in the gtk-osx repo that make it work on Monterey), I think I should replace the blob (even though it doesn't contain Gtk 3.24.31 yet).

PR #3718, opened 2022-01-14
Add maim&xclip option to screenshot plugin

Quoteme:
Add the default option to use "maim" and "xclip" to copy selections from the screen into the system clipboard.
This allows for a Linux user who has maim and xclip installed to simply select part of the screen and paste it using ctrl+v or by pasting using the "paste" button in xournal++.
This is better than before because it does not require saving temporary files. The previous option of using scrot and saving in a temp file still exists and is used as a fallback

### DIFF
--- a/plugins/QuickScreenshot/main.lua
+++ b/plugins/QuickScreenshot/main.lua
@@ -59,6 +59,8 @@ function go()
   local windowsUtilities = {} -- list of windows programs here
   local macUtilities = {} -- list of macOS programs here
   local elseUtilities = -- list of programs for other systems
+                        -- maim and xclip are not listed, because they
+                        -- are the default
     { "scrot --overwrite --select --silent "
 -- gnome-screenshot behaves oddly (pollutes the window)
 -- the line is added here to be indicative of what
@@ -74,6 +76,13 @@ function go()
   elseif operatingSystem == "Darwin" then
     app.msgbox("macOS not supported yet.", {[1] = "OK"})
   else
+    -- Skip all the detection of tools in case maim and xclip are
+	-- installed. Then a screenshot is made with maim and saved to the
+	-- clipboard, skipping all other tools
+    if existsUtility("maim") and existsUtility("xclip") then
+      os.execute("maim -s | xclip -selection clipboard -t image/png")
+      return
+    end
     -- This becomes true if at least one screenshot
     -- utility is available on the system
     local foundUtility = false


LittleHuba:
Thanks for this nice work!
Could you please change the commit description (headline) to something more descriptive?

Quoteme:
I have done that. Is it okay now?

rolandlo:
I'm struggling to come up with the ideal solution, but here are my thoughts:
- There are different use cases for a screenshot plugin: 
     1) making a screenshot to paste it from the clipboard (in Xournal++ or somewhere else)
     2) making a screenshot to store it in a file selected by the user
     3) making a screenshot to store it in some folder with automatic filename creation
- 1) is the preferred behavior by @Quoteme and realized in this PR, 
  2) is the status quo and 
  3) was the original intention by @createyourpersonalaccount 
- The simplest way to realize i. would be to `os.execute` the command `gnome-screenshot -a -c` (like pointed out already in [this discussion](https://github.com/xournalpp/xournalpp/discussions/2024#discussioncomment-127412), but that tool may be only available on `gnome` based distros. Using `maim` and `xclip` like in this PR is not much more difficult either. Instead of `maim` we could also continue to use `scrot`. @Quoteme Did you have a particular reason why you preferred `maim` over `scrot`?
- It would easily be possible to have 2 or 3 different menu entries (and keyboard accelerators) for realizing the different behaviours i-ii (and if needed iii) in the plugin. That looks better than just overriding behavior ii when `maim` and `xclip` are present, although it has also some drawbacks (cluttering the menu and using different shortcuts)
- A problem with the screenshot plugin (in general), is that it does not work on Wayland. If there is a solution that also works on Wayland, then that solution should be preferred.

Quoteme:
> * @Quoteme Did you have a particular reason why you preferred `maim` over `scrot`?

Yes. Maim allows one to pipe the screenshot directly into xclip (see: `maim -s | xclip -selection clipboard -t image/png` from my commit). Therefor no additional temporary file needs to be created just to store a screenshot (like we need with scrot). This is the same as `gnome-screenshot -c`. The problem with gnome-screenshot is, that it only runs inside gnome shell and fails with ` Unable to use GNOME Shell's builtin screenshot interface, resorting to fallback X11.` otherwise.

rolandlo:
Let's go forward. Merging in 24 hours if no objections are raised.

rolandlo:
Thanks for your contribution @Quoteme . I merged it.

PR #3738, opened 2022-01-22
fix: Crash on startup with no recent files

### LINKED ISSUE #3734: Xournal++ crashes on startup when autoloadMostRecent is true and no recent files are available
**Affects versions :**
 - OS: Linux (Ubuntu 20.04)
 - (Linux only) Desktop environment: KDE
 - libgtk: 3.22.30
 - Xournal++ 1.1.0
 - Installation method: Appimage, official Release

**Describe the bug**
When Xournal++ is installed for the very first time and the setting "autoloadMostRecent" is set to true, but no file is actually ever saved and thus no recent files are available, Xournal++ crashes on the next start and can not be started again. After manually editing `~/.config/xournalpp/settings.xml` and resetting "autoloadMostRecent" everything works as expected.
I was able to reproduce this issue by removing all relevant recent files from `~/.local/share/recently-used.xbel`.

**To Reproduce**
1. Activate autoloadMostRecent
2. Close Xournal++
3. Remove all entries referring to Xournal++ files (*.xopp) from `~/.local/share/recently-used.xbel` (This is the state when you first install Xounal++ and have never opened any *.xopp files). It seems that the file ending is actually the relevant part that is used for discovering recent files. Just changing the file ending was enough for me to reproduce the error. Might be better to remove the complete `<bookmark></bookmark>` section though. The mime-type should be `application/x-xoj`, but changing the mime-type does apparently not influence discovery of recent files.
4. Try to restart Xournal++
5. Xournal++ instantly crashes. The only way to recover is to manually edit ~/.config/xournalpp/settings.xml and set `<property name="autoloadMostRecent" value="false"/>`

**Expected behavior**
Xournal++ can be started and does not crash.

namlit:
- Add nullptr check to getMostRecent method
- Add nullptr check to on_startup function
- Change method descriptions to make explicit that nullptrs can be returned

Fixes #3734

### DIFF
--- a/src/control/RecentManager.cpp
+++ b/src/control/RecentManager.cpp
@@ -105,6 +105,9 @@ auto RecentManager::sortRecentsEntries(GtkRecentInfo* a, GtkRecentInfo* b) -> gi
 
 auto RecentManager::getMostRecent() -> GtkRecentInfo* {
     GList* filteredItemsXoj = filterRecent(gtk_recent_manager_get_items(gtk_recent_manager_get_default()), true);
+    if (filteredItemsXoj == nullptr) {
+        return nullptr;
+    }
     auto mostRecent = static_cast<GtkRecentInfo*>(filteredItemsXoj->data);
 
     gtk_recent_info_ref(mostRecent);
--- a/src/control/RecentManager.h
+++ b/src/control/RecentManager.h
@@ -80,6 +80,7 @@ public:
 
     /**
      * Returns the most recent xoj item from the underlying GtkRecentManager
+     * or nullptr, if no recent files exist
      */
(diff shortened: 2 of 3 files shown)


bhennion:
This looks great! Thanks for the catch and the fix!
Would you mind rebasing this onto the branch `release-1.1` so that it'd be included in our next minor release?

namlit:
> This looks great! Thanks for the catch and the fix! Would you mind rebasing this onto the branch `release-1.1` so that it'd be included in our next minor release?

Just rebased and changed the destination branch for the merge.

bhennion:
Thanks!

@xournalpp/core Merging in 48 hours unless someone objects.

PR #3740, opened 2022-01-22
Merge bugfixes from 1.1.x into main dev branch

Febbe:


mac-setup/Info.plist
```diff
@@ -2,88 +2,67 @@
...
     <key>CFBundleName</key>
     <string>Xournal++</string>
-    <key>CFBundlePackageType</key>
-    <string>APPL</string>
-    <key>CFBundleSignature</key>
+    <key>CFBundleDisplayName</key>
     <string>Xournal++</string>
+
+    <key>CFBundleShortVersionString</key>
+    <string>1.1.1~dev</string>
```
Febbe, line 19:
```suggestion
    <string>1.1.0+dev</string>
```

mac-setup/Info.plist
```diff
@@ -2,88 +2,67 @@
...
+    <key>CFBundleVersion</key>
+    <string>1.1.1~dev.0</string>
```
Febbe, line 21:
```suggestion
    <string>1.1.0+dev.0</string>
```

rpm/fedora/xournalpp.spec
```diff
@@ -6,7 +6,7 @@
 #This spec file is intended for daily development snapshot release
 %global	build_repo https://github.com/xournalpp/xournalpp/
 %global	build_branch master
-%global	version_string 1.1.0
+%global	version_string 1.1.1~dev
```
Febbe, line 9:
```suggestion
%global	version_string 1.1.0+dev
```

src/core/control/XournalMain.cpp
```diff
@@ -457,8 +457,11 @@ void on_startup(GApplication* application, XMPtr app_data) {
             opened = app_data->control->newFile("", p);
         }
     } else if (app_data->control->getSettings()->isAutoloadMostRecent()) {
-        if (auto p = Util::fromUri(gtk_recent_info_get_uri(app_data->control->getRecentManager()->getMostRecent()))) {
-            opened = app_data->control->openFile(*p);
+        auto most_recent = app_data->control->getRecentManager()->getMostRecent();
+        if (most_recent != nullptr) {
```
Technius, line 461: Minor nitpick, but this can be simplified to
```suggestion
        if (auto* most_recent = app_data->control->getRecentManager()->getMostRecent()) {
```

Technius (approved):
This looks good to me, but it needs a rebase.

src/core/control/XournalMain.cpp
Febbe: Jep, but I am not shure, if I should fix the code from other PR's in a merge.

bhennion: in general, I don't think we should. It could make things more complicated if another fix, say in 1.1.2, touches this bit of code. Then merging this other fix into master would require even more care.

In the present case, it's pretty simple and the above scare may be unjustified. Still, I think we should stick to this basic principle.

Febbe:
Will rebase this after release 1.1.1

Issue #2278, opened 2020-09-25
Todo: Refactor `ObjectInputStream` to use std::stringstream


Febbe:
The `ObjectInputStream` class uses a self implemented serialization mechanism.
 - The read data is `reinterpret_cast`ed to the read values this results in a wrong alignment this is UB. 
 - Currently prone to bugs and should be replaced with a correct implementation like std::stringstream.

```text
ObjectInputStream.cpp

Load of misaligned address 0x6190005ad782 for type 'int', which requires 4 byte alignment
    ObjectInputStream::readString() ObjectInputStream.cpp:110
Load of misaligned address 0x6190005ad7bf for type 'double', which requires 8 byte alignment
    ObjectInputStream::readDouble() ObjectInputStream.cpp:86
Load of misaligned address 0x6190005ad88f for type 'int', which requires 4 byte alignment
    ObjectInputStream::readInt() ObjectInputStream.cpp:74
Load of misaligned address 0x6190005ad89d for type 'size_t' (aka 'unsigned long'), which requires 8 byte alignment
    ObjectInputStream::readSizeT() ObjectInputStream.cpp:98
Load of misaligned address 0x6190005ad8bf for type 'int', which requires 4 byte alignment
    ObjectInputStream::readData(void**, int*) ObjectInputStream.cpp:129
Load of misaligned address 0x6190005ad8c3 for type 'int', which requires 4 byte alignment
    ObjectInputStream::readData(void**, int*) ObjectInputStream.cpp:132
    
    Stroke::readSerialized(ObjectInputStream&) : Stroke.cpp:68
    Control::clipboardPasteXournal(ObjectInputStream&) Control.cpp:2638
```

Febbe:
May cause #2262

theanshm:
I would like to work on this issue if it is open to work on. I am new to contributing to open source and thus would require a little push in the right direction but I am willing to learn.

bhennion:
One remark: from the current version of the code in `Stroke::readSerialized` (and probably elsewhere too)
```
    in.readData(reinterpret_cast<void**>(&p), &count);
    this->points = std::vector<Point>{p, p + count};
```
The `ObjectInputStream::readData` method is not compatible with modern containers. This results in the data being copied twice on deserialization. This probably doubles the cost of the process for no reason. This should be fixed along with the above.

Febbe:
Closed by #2870

PR #2870, opened 2021-02-24
Refactor and test ObjectInputString

jetSett:
Refactoring ObjectInputString to use std::stringstream and adding unit test to it.

cf #2278

PR #3742, opened 2022-01-25
Attempt to fix copies #2278(comment)

karina-ranadive:
I found this error on the issues page of Xournal++: "Refactor ObjectInputStream to use std::stringstream #2278". I did research into this error and referenced the two links below when coding a potential fix. I am running into errors when it comes to my solution as it may have to do with the actual build of the project itself or the code but I feel based on my research, this could serve as a potential solution. I basically used alignment structs paired with istringstream objects to correct the misalignment issue mentioned in #2278.  

https://stackoverflow.com/questions/28727914/what-does-misaligned-address-error-mean#:~:text=%22Misaligned%20address%22%20usually%20means%20that,bit%20integer%20from%20address%200x1001). 

https://stackoverflow.com/questions/11983311/c-4-bytes-aligned-data

### DIFF
--- a/readme/MacBuild.md
+++ b/readme/MacBuild.md
@@ -1,5 +1,5 @@
 # Xournal++ MacOS .app Build
-hi
+
 Do not install macports or homebrew. If you have installed it, you need to
 create a new user, and use this for the whole process. jhbuild does not work,
 if there is such an environment installed.
--- a/readme/MacBuild.md
+++ b/readme/MacBuild.md
@@ -1,4 +1,5 @@
 # Xournal++ MacOS .app Build
+hi
 Do not install macports or homebrew. If you have installed it, you need to
 create a new user, and use this for the whole process. jhbuild does not work,
 if there is such an environment installed.
--- a/src/util/include/util/serializing/ObjectInputStream.h
+++ b/src/util/include/util/serializing/ObjectInputStream.h
@@ -39,6 +39,16 @@ public:
 
     void readData(void** data, int* len);
     cairo_surface_t* readImage();
+    
+    struct AlignmentInt{ //referenced https://stackoverflow.com/questions/11983311/c-4-bytes-aligned-data & https://stackoverflow.com/questions/28727914/what-does-misaligned-address-error-mean#:~:text=%22Misaligned%20address%22%20usually%20means%20that,bit%20integer%20from%20address%200x1001).
+        char c;
+        __attribute__((__aligned__(4))) int32_t member;
+    }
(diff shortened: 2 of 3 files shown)


danemtsov:
I think https://github.com/xournalpp/xournalpp/issues/2278#issue-709022140 has been fixed with #2870.
The only task left is avoiding the second copy of the deserialized data as described here: https://github.com/xournalpp/xournalpp/issues/2278#issuecomment-777661990.

Febbe:
Yes, as you already mentioned, you hit an already fixed, but not closed issue; I am sorry, but your work seems to be obsolete.
But I can give you some explaining feedback.

Febbe (commented):
The previous implementation used a `int my_int = &reinterpret_cast<int const*>(pointer_into_unalligned_continuous_char_sequence)`

What happens here is: 
 1. `pointer_into_unalligned_continuous_char_sequence` is a pointer, and it's value describes the position of a character.
 2. It is casted to an `const int*` (`int const*` are the same), so the result is a pointer, which value seem to point to an `const int`
 3. Then the pointer is dereferenced, a hint for the compiler, that you want to access the data of the pointer. 
 4. The `int` at the pointer position is copied.

The UB happens at point 4., when the address points to a position in the memory, which has not the alignment of int. So changing the target type to a special aligned type won’t change anything, because the reinterpret_cast should've been casted to the specific alignment. Since there are 4 possible alignments, this is not predictable without knowing the target architecture.

A proper solution would be to just use `std::memcpy`, which will originally copy all bytes from address x to address y. And when the target is properly aligned, nothing can go wrong.

The reason why this was UB is described here: https://en.cppreference.com/w/cpp/language/reinterpret_cast

Febbe:
> The only task left is avoiding the second copy of the deserialized data as described here: [#2278 (comment)](https://github.com/xournalpp/xournalpp/issues/2278#issuecomment-777661990).

There are still copies: ObjectInputStream.cpp(24,4): 
```
auto ObjectInputStream::read(const char* data, int data_len) -> bool {
    istream.clear();
    len = (size_t)data_len;
    std::string dataStr = std::string(data, len); // Copy
    istream.str(dataStr); // Copy
```
The second copy is elideable. The first not (we have to wait for c++23 or write our own streambufferobject / use [boost](https://stackoverflow.com/questions/3122901/non-copying-istringstream)  )
```
auto ObjectInputStream::read(const char* data, int data_len) -> bool {
    istream.clear();
    len = (size_t)data_len;
    istream.str(std::string(data, len)); // Only one copy, the constructed string is a rvalue (prvalue) and "moved into the stream"
```

Febbe:
I marked your PR as draft, if you want to fix the copy, you can reupload it.

Febbe:
Closing, because there is no evidence, that this PR will get finished soon.

PR #3745, opened 2022-01-25
Move git repo info to separate config header

danemtsov:
This reduces build tasks after checkout from ~120 to 2 object files and their dependencies.


Technius (approved):
LGTM, thanks.

PR #3746, opened 2022-01-25
Build console application for Debug configuration on Windows

danemtsov:
This makes debugging a bit easier when using a build setup that doesn't itself require an MSYS console (e.g. VS Code + CMake tools).

Currently based on #3745.


MichelReims:
Hello,
I use Xournalpp (last nightly version) under Windows 7 and 10 and I have a problem. Xournalpp create a directory named "d:_temp\msys\msys64\var\cache\fontconfig", when I run it. When I close Xournal, this directory is not deleted.  Have you this problem ?

Febbe:
Yes, my wish is, to have someday a msvc pipeline and a msvc generated release. I'll take a look

src/CMakeLists.txt
```diff
@@ -78,7 +78,6 @@ if (WIN32)
   # windows specific source files
   file (GLOB xournalpp-sources-win32 exe/win32/*.cpp exe/win32/*.h)
   target_sources(xournalpp PRIVATE ${xournalpp-sources-win32})
-  target_compile_options(xournalpp PRIVATE -mwindows)
```
Febbe, line 81: This is indeed obsolete (duplication)

src/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ add_executable(xournalpp exe/Xournalpp.cpp)
 target_link_libraries(xournalpp PRIVATE xoj::core xoj::util)
 set_target_properties(xournalpp PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
         MACOSX_BUNDLE $<IF:$<PLATFORM_ID:Darwin>,TRUE,${CMAKE_MACOSX_BUNDLE}>
-        WIN32_EXECUTABLE $<IF:$<PLATFORM_ID:Windows>,TRUE,${CMAKE_WIN32_EXECUTABLE}>
+        WIN32_EXECUTABLE $<IF:$<AND:$<PLATFORM_ID:Windows>,$<NOT:$<CONFIG:Debug>>>,TRUE,${CMAKE_WIN32_EXECUTABLE}>
```
Febbe, line 46: Xournalpp should be connecting to the console, which started it, does this not work for you?
It should be connected to the vscode terminal.

danemtsov:
So trying to get *some* console output while debugging eventually evolved into making available *all* console output and even implementing rudimentary crash handlers for the proof of concept.

The idea is:
* When debugging, attach to parent process (-> debugger) console
* Use the (hidden) console spawned on startup by default
  - It is not hidden if `--show-console` is specified as first cl arg. In this case the spawned console is also used when debugging.
  - It can be shown by selecting *Help > Debug > Show console* from the menu bar.
  - It will be shown if a crash occurs (that's what the crash handlers are for).

For keeping the console clean I implemented a simple helper executable that is spawned instead of cmd. All it does is swallowing input.

src/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ target_compile_options(defaults INTERFACE -Wall -Wreturn-type -Wuninitialized -W
 target_include_directories(defaults INTERFACE ${PROJECT_BINARY_DIR}/src)
 add_library(xoj::defaults ALIAS defaults)
 if (MSVC)
-  target_compile_definitions(defaults INTERFACE PUBLIC NOMINMAX WIN32_LEAN_AND_MEAN)
+  target_compile_definitions(defaults INTERFACE NOMINMAX WIN32_LEAN_AND_MEAN)
```
danemtsov, line 16: Actually this is unrelated. Forgot to unstage...

src/core/control/win32/WinConsole.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#ifdef _WIN32
+
+#include <string>
+
+// All functions defined in /src/exe/win32/console.cpp
+
+namespace xpp_win32 {
```
Febbe, line 9:
```suggestion
namespace xoj::win32 {
```

src/exe/win32/console.cpp
```diff
@@ -38,9 +60,10 @@ void attachConsole() {
             } else {
                 // It takes a short time before the console becomes ready to be attached, unfortunately
                 // there is no API that would let us wait for it.
-                for (unsigned short i = 0; i < 20; i++) {
+                for (unsigned short i = 0; i < 40; i++) {
```
Febbe, line 63: **Question:** This will block the main thread, therefore it might be better to extrude this into a different thread?

src/core/control/win32/WinConsole.h
danemtsov: I did that already :)

Actually I wanted feedback more on the general idea outlined [here](https://github.com/xournalpp/xournalpp/pull/3746#issuecomment-1035699055) than on the code itself, which is just a messy proof of concept.

src/exe/win32/console.cpp
danemtsov: I implemented a more sophisticated (but potentially overcomplicated) version which I can push later.

I.a. it uses inter-process syncing via

src/exe/win32/console.cpp
```diff
@@ -66,8 +89,99 @@ void attachConsole() {
...
+
+    auto reopenOutput = [](FILE* stdFile, std::basic_ios<char>& stream, std::basic_ios<wchar_t>& wstream,
+                           const char* streamDesc) {
+        FILE* dummyFile = nullptr;
+        if (auto r = freopen_s(&dummyFile, "CONOUT$", "w", stdFile)) {
+            // Continuing with potentially closed stdout/stderr will lead to UB
+            //  => crash the process
+
+            // For debugging
+            std::system_error e(r, std::generic_category(), std::string("Failed to reopen ") += streamDesc);
```
Febbe, line 127: you could throw this. It will automatically call abort if not catched.

src/exe/win32/console.cpp
```diff
@@ -1,24 +1,46 @@
...
+
+namespace xpp_win32 {
+
+static void myAllocConsole(bool hidden) {
+    // Path of Xournal++ executable
+    std::vector<wchar_t> moduleFileNameData(128, 0);
+
+    // GetModuleFileName returns number of chars excluding trailing \0 if path
+    // fits into buffer including trailing \0, else returns buffer size.
+    while (GetModuleFileNameW(NULL, moduleFileNameData.data(), moduleFileNameData.size()) ==
```
Febbe, line 24: It most likely wont ever happen, but can you guard the loop from executing infinitelly, like increasing the size only up to 2^16? It can also happen, that the function returns 0 and sets the error flag. This should be guarded too.

src/exe/win32/console.cpp
```diff
@@ -1,24 +1,46 @@
...
 #include <windows.h>
 
-void attachConsole() {
-    if (GetConsoleWindow() != NULL) {
-        // Console is already attached.
-        return;
+#include "filesystem.h"
+
+static bool _haveOwnConsoleWindow = false;
+static HANDLE _consoleHelperJob = NULL;
```
Febbe, line 14: use `nullptr` instead of `NULL`

src/core/control/win32/WinConsole.h
Febbe: Ah, ok. I think it is very awesome, but I have objections/questions regarding the correctness (absence of undefined behavior). Also I don't know why we require an additional executable for the console. Shouldn't it be enough to spawn a cmd-terminal? 

Regarding undefined behavior: The low level signal handling is implemented via the windows API, so this might be ok (depends on the specification): When using std::signal (the C-API) it is undefined behavior to use any stream operators or functions.

danemtsov:
This is my state so far. The last commit is the relevant one for the console, the rest is just preparations (haven't had the time to further split it up). It's not ready yet but the main ideas are there:
A platform-specific console manager object (`PlatformConsole`) is initialized right at the start of main(). On windows, this performs the necessary steps to get a console attached as outlined [here](https://github.com/xournalpp/xournalpp/pull/3746#issuecomment-1035699055).
By default, the manager object will show the console window upon destruction (which will make cl arguments like --help, --version etc. actually useful on Windows, the output shows in a new console and we still are a well-behaved GUI app that doesn't touch the console it was started from). If execution continues after ```handle-local-options``` (i.e. the main window shows up), the destruction behavior is set to 'keep console hidden' (and kill console helper process).
For Unix the implementation does nothing yet, but maybe a similar feature might be useful if Xournal++ isn't started from a shell, especially if some plugins might want to output something to a console?

Please don't feel offended when it says "guard from GPL", I want to leave the possibility to use the code in GPL-incompatible projects. I haven't conclusively thought about licensing yet...

Regarding your questions:

* About the second executable:
  The actual console window is not managed by cmd but by the OS itself (via conhost.exe). For a GUI program to get a console it must either request one itself (with AllocConsole(), no way to initially hide it) or attach to an existing one. The latter is what we do with spawning a console process with hidden window state (console-helper.exe in this PR, previously cmd.exe).
  Also if we want to show the console after process exit (whether it be a crash or after printing --help), there must be at least one process attached to the console or else the OS will destroy the console.
  In a nutshell the second executable is needed for creating the console potentially hidden and to keep it alive after Xournal++ exits.
Additionally I implemented a deadlock alert as with previous versions of GTK/pango/harfbuzz, Xournal++ would often deadlock before any window showed.

* I'm not sure what exactly you mean here:

  > The low level signal handling is implemented via the windows API, so this might be ok (depends on the specification): When using std::signal (the C-API) it is undefined behavior to use any stream operators or functions.

  I haven't seen anything regarding what is (not) allowed in win32 unhandled exception filter or C++ terminate handler?

I'm not yet satisfied about the overall design, specifically:
* PlatformConsole as an abstract base class just so that crash handler (which is also platform specific) can access internals without pulling in more dependencies. This could be structured better by having platform-specific code grouped at one place.
* The actual PlatformConsole instance needs to have static storage duration so that the destructor is called on process exit (. ConsoleCtl references this instance, but when a crash occurs, the destructor of the PlatformConsole needs to be called in the crash handler, which formally leaves a dangling reference in the ConsoleCtl.

Febbe:
> Please don't feel offended when it says "guard from GPL", I want to leave the possibility to use the code in GPL-incompatible projects. I haven't conclusively thought about licensing yet...

The reason, because we use GPL is, because we must, when we want to use other GPL software. But besides that, I don't know international law, but here in Germany, the writer of the code is the owner, and it is not even possible to give the copyrights to someone else.

Febbe:
@danemtsov what do you think of boost::io_service as signal handler for both linux and windows? But that dependency might be too big for signal handling only.

PR #3754, opened 2022-01-30
Rendering: Tiling

### LINKED ISSUE #3123: Infinite page option?
Would it be possible to add an infinite page option? I picture it like in OneNote where the page gets bigger and bigger the more you scroll horizontally / vertically.
This would help with long flowcharts or big images since you are not restricted by the size of a DIN A4 page.

LittleHuba:
This is a very early implementation of tiling. Expect some serious bugs and inconveniences...

Given some more work, this is the foundation for supporting infinite canvas and arbitrary page sizes.

See commit messages for more detailed information.

Fixes #3123

### DIFF
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -1,135 +1,168 @@
 #include "RenderJob.h"
 
 #include <cmath>
+#include <future>
 
 #include "control/Control.h"
 #include "control/ToolHandler.h"
 #include "gui/PageView.h"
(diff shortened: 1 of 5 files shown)


raffaem:
How can I test this? The page is still A4

LittleHuba:
As I said this is early work in progress. You still have to manually resize your pages

LittleHuba:
I will not continue work on this as it makes no sense to do this before we move to GTK4. With the node-based rendering of GTK4 this would just further complicate the switch and would be completely unusable once we switched.

PR #3760, opened 2022-01-31
Use `fs::path::u8string` everywhere

danemtsov:
... since conversion in `fs::path::string` is unspecified if `value_type` is not char (like on Windows).

This also fixes an "Illegal byte sequence" exception in my Clang build when using funny file names (e.g. '🌍🌎🌏🛰🚀.xopp') thrown here:
https://github.com/xournalpp/xournalpp/blob/2bea77ca5c91fb0f2dad2a4f2dec91e382eeedd5/src/core/control/settings/MetadataManager.cpp#L172
...and maybe others I haven't hit.

### DIFF
--- a/src/core/control/AudioController.cpp
+++ b/src/core/control/AudioController.cpp
@@ -30,7 +30,7 @@ auto AudioController::startRecording() -> bool {
 
         g_message("Start recording");
 
-        bool isRecording = this->audioRecorder->start((getAudioFolder() / data).string());
+        bool isRecording = this->audioRecorder->start((getAudioFolder() / data).u8string());
 
         if (!isRecording) {
             audioFilename = "";
--- a/src/core/control/latex/LatexGenerator.cpp
+++ b/src/core/control/latex/LatexGenerator.cpp
@@ -49,7 +49,7 @@ auto LatexGenerator::asyncRun(const fs::path& texDir, const std::string& texFile
         return res;
     };
 
-    auto texFilePath = (Util::getLongPath(texDir) / "tex.tex").string();
+    auto texFilePath = (Util::getLongPath(texDir) / "tex.tex").u8string();
     for (auto i = cmd.find(u8"{}"); i != std::string::npos; i = cmd.find(u8"{}", i + texFilePath.length())) {
         cmd.replace(i, 2, texFilePath);
     }
--- a/src/core/control/settings/MetadataManager.cpp
+++ b/src/core/control/settings/MetadataManager.cpp
@@ -2,7 +2,7 @@
 
 #include <algorithm>  // std::sort
 #include <fstream>
-#include <iomanip> // std::quoted
+#include <iomanip>  // std::quoted
 #include <sstream>
 
 #include <fcntl.h>
--- a/src/core/control/settings/MetadataManager.cpp
+++ b/src/core/control/settings/MetadataManager.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>  // std::sort
 #include <fstream>
+#include <iomanip> // std::quoted
 #include <sstream>
 
 #include <fcntl.h>
@@ -169,7 +170,7 @@ void MetadataManager::storeMetadata(MetadataEntry* m) {
 
     ofstream out(path);
     out << "XOJ-METADATA/1.0\n";
-    out << m->path.u8string() << "\n";
+    out << std::quoted(m->path.u8string()) << "\n";
     out << "page=" << m->page << "\n";
     out << "zoom=" << m->zoom << "\n";
     out.close();
--- a/src/core/control/settings/MetadataManager.cpp
+++ b/src/core/control/settings/MetadataManager.cpp
@@ -28,13 +28,13 @@ MetadataManager::~MetadataManager() { documentChanged(); }
 void MetadataManager::deleteMetadataFile(fs::path const& path) {
     // be careful, delete the Metadata file, NOT the Document!
     if (path.extension() != ".metadata") {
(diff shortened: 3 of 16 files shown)


src/core/control/settings/MetadataManager.cpp
```diff
@@ -169,7 +169,7 @@ void MetadataManager::storeMetadata(MetadataEntry* m) {
 
     ofstream out(path);
     out << "XOJ-METADATA/1.0\n";
-    out << m->path << "\n";
+    out << m->path.u8string() << "\n";
```
danemtsov, line 172: Not sure if there are any more like this though...

danemtsov: 'Quoted'ness fixed; pretty sure this is the only place.

danemtsov:
> the windows console couldn't display utf-8

IMHO this shouldn't be cared about too much since, by default, stdout/stderr aren't a thing anyway in Windws GUI apps (except when launched from a shell running inside mintty, more on that maybe later in my other PR).

On the other hand 'normal' users ideally shouldn't have to install a (multi?) GB MSYS environment just to get some debug output, so either a way to direct stdout/stderr to a log file is needed (which then should be UTF-8), or we just add an action to show the console window that is created on startup and direct stdout/stderr there. And since nothing else is using this console (except the processes that Xournal++ spawns) we can also just `SetConsoleOutputCP(CP_UTF8)`, which should work well enough.

What do you think?

Regardless, this
https://github.com/xournalpp/xournalpp/blob/2bea77ca5c91fb0f2dad2a4f2dec91e382eeedd5/src/core/control/settings/MetadataManager.cpp#L172
and the reading counterpart need to be changed so that the metadata can be associated with a file whose name doesn't fit into single byte encoding.

Febbe:
> IMHO this shouldn't be cared about too much since, by default, stdout/stderr aren't a thing anyway in Windws GUI apps. (except when launched from a shell running inside mintty, more on that maybe later in my other PR).

Actually, it is a bug, that it does not connect to the calling tty like cmd or powershell :/. So it should be a thing.
But location issues in a debug output aren't really a thing.

I only wanted to explain the original reasons why we didn't choose `u8string` in the first place.
Also, we will run into trouble with c++20. We might switch to it in the near future. But then all current strings and conversions will bite us.

The `SetConsoleOutputCP(CP_UTF8)` function is not part of early windows7 / 8 / 10 versions. Aren't they?

PR looks good so far, but can you squash all your commits?

src/core/control/AudioController.cpp
```diff
@@ -30,7 +30,7 @@ auto AudioController::startRecording() -> bool {
 
         g_message("Start recording");
 
-        bool isRecording = this->audioRecorder->start((getAudioFolder() / data).string());
+        bool isRecording = this->audioRecorder->start((getAudioFolder() / data).u8string());
```
Febbe, line 33: **Note @xournalpp/core :** 
`bool start(const std::string& filename);` should be `bool start(fs::path const& filename);`

test/unit_tests/util/PathTest.cpp
```diff
@@ -55,65 +55,65 @@ TEST(UtilPath, testClearExtensions) {
...
     Util::clearExtensions(a);
-    EXPECT_EQ(old_path.string(), a.string());
+    EXPECT_EQ(old_path.u8string(), a.u8string());
 
 
     // The following tests use the generic separator which works on all systems
     auto b = fs::path("/test/asdf.TXT");
     Util::clearExtensions(b);
-    EXPECT_EQ(string("/test/asdf.TXT"), b.string());
+    EXPECT_EQ(string("/test/asdf.TXT"), b.u8string());
```
Febbe, line 70: Either the literal must be also in u8 or b.string() must be used.

```suggestion
    EXPECT_EQ(string(u8"/test/asdf.TXT"), b.u8string());
```

This applies for all assertions.

Febbe (changes requested):
Ok, I found some problems, which must be fixed.

src/core/control/latex/LatexGenerator.cpp
```diff
@@ -49,7 +49,7 @@ auto LatexGenerator::asyncRun(const fs::path& texDir, const std::string& texFile
         return res;
     };
 
-    auto texFilePath = (Util::getLongPath(texDir) / "tex.tex").string();
+    auto texFilePath = (Util::getLongPath(texDir) / "tex.tex").u8string();
```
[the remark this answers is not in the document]
Technius: Yes, this has to be system encoding. Should probably leave a comment about that here.

> g_shell_parse_argv
>
> command_line: The value is a file system path, using the OS encoding.

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -555,7 +556,7 @@ void LoadHandler::parseStroke() {
         } else {
             auto tempFile = getTempFileForPath(fn);
             if (!tempFile.empty()) {
-                stroke->setAudioFilename(tempFile.string());
+                stroke->setAudioFilename(tempFile.u8string());
```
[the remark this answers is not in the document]
Technius: Any time there is an exception like this, we should document it with a comment or two.

Or change the method name to make it explicit: `stroke->setAudioFilenameSystemEncoding`, `stroke->setAudioFilenameU8`

Febbe:
#3785 fixes one of the above problems with utf8 paths.

Febbe:
@[danemtsov](https://github.com/danemtsov), do you plan to work on this further?

PR #3779, opened 2022-02-06
fixup warnings

Febbe:


Technius:
Can you break this into several PRs, each one focused on a specific set of files (e.g., each commit in this PR is a separate PR). I don't want to end up with a situation like in #3326 where the PR snowballed so much that it became unmanageable to review & merge. Also, this can introduce a lot of merge conflicts.

Febbe:
The rest of the fixups here are either too much related or are very trivial changes. (e.g. rule of 5)

Febbe:
Ignore clang-format, my version is clang-format 14, so differences are expected.

src/core/audio/AudioQueue.h
```diff
@@ -121,9 +121,9 @@ class AudioQueue {
      * @return std::pair<double, int>,
```
Technius, line 121: Update this comment.

src/core/audio/VorbisConsumer.cpp
```diff
@@ -27,7 +27,7 @@ auto VorbisConsumer::start(const std::string& filename) -> bool {
 
     this->consumerThread = std::thread([this, sfFile = std::move(sfFile), channels = channels] {
         auto lock{audioQueue.acquire_lock()};
-        auto buffer_size{static_cast<size_t>(std::max(0, 64 * channels))};
+        auto buffer_size{size_t(64 * channels)};
```
Technius, line 30:
```suggestion
        size_t buffer_size(64 * channels);
```

src/core/pdf/base/XojCairoPdfExport.cpp
```diff
@@ -160,15 +160,17 @@ auto XojCairoPdfExport::createPdf(fs::path const& file, PageRangeVector& range,
...
     if (this->progressListener) {
         this->progressListener->setMaximumState(count);
     }
 
     int c = 0;
-    for (PageRangeEntry* e: range) {
-        for (int i = e->getFirst(); i <= e->getLast(); i++) {
+    for (auto const& e: range) {
+        for (size_t i = e.getFirst(); i <= e.getLast(); i++) {
             if (i < 0 || i >= static_cast<int>(doc->getPageCount())) {
```
Technius, line 174: Is this cast still needed?

src/util/PageRange.cpp
```diff
@@ -3,29 +3,21 @@
...
-auto PageRangeEntry::getLast() const -> int { return this->last; }
-
-auto PageRangeEntry::getFirst() const -> int { return this->first; }
-
 auto PageRange::isSeparator(char c) -> bool { return (c == ',' || c == ';' || c == ':'); }
 
-auto PageRange::parse(const char* str, int pageCount) -> PageRangeVector {
+
+// Todo (rename): what does this? Neither name or code is explaining.
+auto PageRange::parse(std::string_view str, size_t pageCount) -> PageRangeVector {
```
Technius, line 10: I believe this parses the string that's typed in the page range field of the export dialog. See `ExportDialog.cpp`.
image

Technius: Ideally, there should be unit tests for this method.

src/core/audio/VorbisConsumer.cpp
Febbe: This cast is actually required, since `64 * channels` is not size_t.

PR #3780, opened 2022-02-06
Fix minor typo: Tex to TeX

tattsan:
The same typos were fixed in several places before, but it has been still there.


Technius:
LGTM, thanks. I wanted to merge this into 1.1.1 but the commit history got too complicated. So merging into master for now.

Issue #3777, opened 2022-02-06
 Calligraphic Pen 2


odd-orb:
Hello,

I saw someone writing in some other software the other day and really liked the qualities of the pen tool they were using. I figured that the basic effect is quite simple: moving the pen in the down direction results in a heavier stroke than moving it up with some interpolation in between. I looked at the features of this app and the issues. I saw #2166, but that is about a different mechanic.

I then started developing and soon had the first version running. I called the feature **heavy** *downstroke* Here are some impressions from my testing so far:

![Hello, World.](https://user-images.githubusercontent.com/76651420/152674333-b21295af-6150-4620-a4e4-fb5a8b76dcd9.png)

The following image contains three examples with the same strokes copied and the line style changed back to the default for each of them (as some kind of before-after-comparison). The three examples read:

> Guten Abend Xournal++

> Hello, World.

> This is how I imagined the tool to be used mainly.
> For normal Notes, to achieve a simple and consistent but more dynamic look. I am writing this with pressure sensitivity deactivated and quite thick pen width.

![Three examples for the heavy downstroke pen with comparison (i. e. the strokes with the default line style).](https://user-images.githubusercontent.com/76651420/152674900-2399193f-9310-4840-860f-64c7bc01e962.png)

Last but not least, the changed dropdown menu:

![The changed dropdown menu with the new "heavy downstroke"-tool selected](https://user-images.githubusercontent.com/76651420/152674953-826531c5-13cf-48c4-8788-6095712ffcba.png)

I just created a fork to get a first pull request going. But not all formating/parsing or serializing/deserializing features are implemented completely jet. I am currently disregarding backward compatibility at these places. That will change very soon, too. But I first want to see how it will break compatibility.

**Planned Features / Possible Changes**

- Heavy Direction: currently, a line that points straight down will be the thickest, a line pointing straight up will be the thinnest. I want to make that direction variable and probably set the default direction to point a little bit more to the lower left instead of straight down.
- Pressure Weight Behaviour: I am currently thinking about how this feature should really interact with the weight calculated from the line pressure. Currently, I am just scaling the weight that would normally be the line width just before it is applied:

   ```
   double width = p1i->z != Point::NO_PRESSURE ? p1i->z : s->getWidth();
   
   if (s->getLineStyle().hasHeavyDownstroke()) {
     double r = s->getLineStyle().getHeavyDownstrokeRatio();
     double dx = p2i->x - p1i->x;
[...]

PR #3781, opened 2022-02-06
Added feature: Heavy Downstroke stroke style (implementation of Calligraphic Pen 2 #3777)

odd-orb:
This is a first implementation of the features discussed in #3777.

Hello, World.

I will create this pull request as a draft since there are currently potentially compatibility-breaking changes (formatting/parsing and serializing/deserializing). Compatibility testing is certainly needed.

&mdash; David


odd-orb:
How would you imagine backward compatibility at serialization and deserialization?

```cpp
void LineStyle::serialize(ObjectOutputStream& out) const {
    out.writeObject("LineStyle");

    out.writeData(this->dashes, this->dashCount, sizeof(double));
    out.writeDouble(this->heavyDownstrokeRatio);

    out.endObject();
}

void LineStyle::readSerialized(ObjectInputStream& in) {
    in.readObject("LineStyle");

    g_free(this->dashes);
    this->dashes = nullptr;
    this->dashCount = 0;
    in.readData(reinterpret_cast<void**>(&this->dashes), &this->dashCount);

    this->heavyDownstrokeRatio = in.readDouble();

    in.endObject();
}
```

I changed these parts by adding the lines containing `writeDouble` and `readDouble`. This will simply throw an error when it trys to deserialize an old serialization, i. e. when there is no double after the double list representing the dashes. I could directly include a check for that and fix this issue that way. Also, if old code tries to deserialize a new serialized line style, it will expect the end of an object to early and fail. Do we care for these cases anyway? It seems to only be used for copy and pasting or something like that.

rolandlo:
> Also, if old code tries to deserialize a new serialized line style, it will expect the end of an object to early and fail. Do we care for these cases anyway? It seems to only be used for copy and pasting or something like that.

Yes, it is only used for copying/pasting. You don't need to care about compatibility. You would have to run two different versions of Xournal++ to get into trouble.

odd-orb:
How do I even create a custom Line Style in Xournalpp? Am I right to assume that that feature is basically unavailable currently? How did you test formatting and parsing until now?

odd-orb:
I implemented the formatting and parsing including backward compatibility. I will do further testing now, but I think this could be regarded as ready. It might be smart to wait until I implemented the angle and adjusted the formatting and parsing accordingly though.

A changed stroke style in the file;

```xml
<stroke tool="pen" ... style="cust: 6 3 0.5 3 hd_ratio(0.5)">
```

had the expected effect.

grafik

(this line was drawn with the mouse through a layer of GWSL (Graphical Windows Subsystem for Linux) which resulted in the tremble)

odd-orb:
I will quickly do one more commit with the correct formatting.

src/core/view/StrokeView.cpp
```diff
@@ -172,10 +185,10 @@ void StrokeView::paint(bool dontRenderEditingStroke, bool markAudioStroke, bool
         cairo_set_operator(crEffective, CAIRO_OPERATOR_SOURCE);
     }
 
-    if (s->getErasable() && !dontRenderEditingStroke) {
+    if (s->getErasable() and !dontRenderEditingStroke) {
         // don't render erasable for previews
         drawErasableStroke(crEffective, s);
-    } else if (s->hasPressure() && !highlighter) {
+    } else if ((s->hasPressure() or s->getLineStyle().hasHeavyDownstroke()) and !highlighter) {
```
bhennion, line 191: I think we should keep the && and || symbols here (rather than `or` and `and`). It's not explicitly in the coding conventions, but those keywords don't appear anywhere in the code. You can ignore this comment if you want.

bhennion (changes requested):
Sorry for the late review.

I like this tool, and the code is clean.
There is an implementation decision I disagree with though: I don't think heavy downstroke should be implemented as a LineStyle, but rather as a tool that applies a factor to the pressure values upon writing (depending on whether you go up or down).

I see three reasons for this:

- It reduces the cost of StrokeView::draw, which can be called often, by computing the factor only once, when the stroke is being created.
- It allows for stroke's rotations. With the current implementation, rotating a stroke results in a changed aspect. (solvable by remembering the angle)
- It involves no changes to the file format: the stroke is stored like any other stroke.

The downside is, of course, that changing the stroke style after the facts becomes less flexible: one could still apply heavy downstroke to an existing normal stroke, but reverting this becomes complicated (undo would work, but say you want to revert this change after reopening the file).

The rest of the team might disagree with me on this: @xournalpp/core what do you think?

Technius:
I would also prefer a non-breaking implementation method; however, I don't think adjusting only the pressure would be equivalent, as the "heavy downstroke" weights the width dynamically:
https://github.com/xournalpp/xournalpp/blob/4a8fd94ec5f7fead9ad88d2c6f9f64fa24fc5d95/src/core/view/StrokeView.cpp#L56-L64

Technius:
On second thought, if the "Heavy Downstroke" is a stroke modifier, then we could perform this weighting during stroke creation, so I guess that's not a problem.

odd-orb:
Thank you for your reviews. I think performing the weighting during stroke creation/stoke style selection might have a few advantages:

- better performance
- no problem with rotation
- no need to change the file format

But I would like to argue for a dynamic application of the effect.

1. **Applying and switching stroke styles is non-destructive for all stroke styles for now...**

2. **Do we need to optimize this feature? Is this the only way of optimizing it?**

   I don't know how bad the impact of this implementation on performance really is. I used this version of Xournal++ for a while now and didn't have any problems with performance. That said, I don't have much experience with really large files.

   The current version would work with just taking an absolute value instead of calculating the norm using multiplying and the square root. This would be faster. With a cheaper version of the calculation, I don't think that calculating the weights on the fly would do any harm, comparing it to the complexity of the calls surrounding the weighting logic. I could be mistaken. Maybe the cheap way could even be the standard if it looks good enough. I would like to add the option to change the angle. This should be symmetric, i.e. the effect with an angle of 20 deg should look the same as if I applied it to the -20deg rotated stroke with an angle of 0 deg.

3. **Is changing the file format really a problem?**

   I don't think we should keep the name `hd_ratio` (or Heavy Downstroke for that matter), but I think this way of extending the format looks pretty promising. I think we should extend the format when adding a new stroke style. A stroke style carries options. We didn't jet make those options visible and modifiable to the user, but I guess that's the plan. Of course, if we decide to make this a destructive stroke style, things are different.

**About the problem with rotation:** I think local coordinates for every stroke could be plausible. A stroke is made out of potentially hundreds of floats already. At most 4 more would not change much, I guess... (remembering just the angle might be a little more space efficient, but it would be more costly to apply the effects accordingly)

I will try to get working on this feature again soon. I have no problem with any of the suggestions regarding coding convention.

Technius:
I just realized that you initially created this PR a few months ago; sorry for the late feedback. Thank you for working on this feature, as this brings some valuable discussion to the table.

I think there is some confusion going on here. We are talking about "line styles" and "stroke styles", but they are not the same thing. Xournal++ has strokes, and strokes may have _line styles_. A line style (which I'll admit is confusingly named) only controls the degree to which a stroke is drawn with dashes. That is, the line style has no effect on the position and pressure of a stroke.

Not everyone has the same definition of "stroke style", but my understanding of "stroke style" is that it is an abstract concept of how a stroke should look like with no direct counterpart in the application. For example, I would consider solid lines, dashed lines, highlighter strokes, and "heavy downstroke"-type strokes to be of different "stroke styles".

I think we're all missing a fundamental problem: how do we want to go about implementing a new "stroke style"? Based on my argument above, it seems inappropriate to add "heavy downstroke" as a line style. But on the other hand, we haven't settled on a standard way to implement new stroke styles. I suppose we could add a new tool (similar to highlighter), but this would not scale past one or two new stroke styles.

---

> Is changing the file format really a problem

We care a lot about backwards compatibility, and modifying the file format means that we have to bump up the file format version. The plan is to bump the file format version with the release of the new file format version in #937, but who knows how long that will take.

That being said, I have been thinking that we should create some sort of "experimental file format version" (e.g., 99999) that is always marked as "too new" so that we can experiment with new features. If we introduce such a version, I think it'd be okay to extend the file format, provided we have some support code that can detect when "experimental" features are used.

Regarding your concerns about the calculations: arithmetic is relatively cheap on modern computers, so let's put the performance as a lower priority. We can figure out how to optimize it if it turns out to be too slow.

> I think we should extend the format when adding a new stroke style

I agree that this information may be needed for individual stroke styles (e.g., we could enhance highlighter with "gradient" or other fun parameters). However, this should only be required if custom rendering logic is required. A more modular approach for strokes that use the normal stroke rendering logic is to provide some sort of plugin API for postprocessing strokes.

PR #3783, opened 2022-02-07
Fix inconsistent path usage for audio files

Febbe:


Febbe:
testet with audiofolder: `P:\♀☺ÊÌ◙┬─┬ノ( ◕◡◕ ノ)`

src/core/audio/SNDFileCpp.h
```diff
@@ -0,0 +1,47 @@
...
+#define ENABLE_SNDFILE_WINDOWS_PROTOTYPES 1
+#endif
+
+#include <memory>
+
+#include <sndfile.h>
+
+#include "filesystem.h"
+
+namespace xoj {
```
Technius, line 26: `namespace xoj::audio` so we don't have to move this again in the future.

src/core/gui/PageViewFindObjectHelper.h
```diff
@@ -149,16 +149,12 @@ class PlayObject: public BaseSelectObject {
         if ((s->intersects(x, y, 15, &tmpGap))) {
             size_t ts = s->getTimestamp();
 
-            std::string fn = s->getAudioFilename();
-
+            auto fn = s->getAudioFilename();
             if (!fn.empty()) {
```
Technius, line 153:
```suggestion
            if (auto fn = s->getAudioFilename(); !fn.empty()) {
```

Technius (approved):
LGTM. But one question: what if the file path is mounted over SSH or some non-file URI?

Technius:
Actually, let's just sidestep the whole non-`file://` URI edge case by assuming that the user will mount such URIs with gvfs.

src/core/audio/SNDFileCpp.h
```diff
@@ -0,0 +1,47 @@
...
+
+using SNDFileGuard = std::unique_ptr<SNDFILE, SNDFILE_Deleter>;
+
+
+inline SNDFILE* sf_open_pp(std::string path, int mode, SF_INFO* info) { return sf_open(path.c_str(), mode, info); }
+#if (defined(ENABLE_SNDFILE_WINDOWS_PROTOTYPES) && ENABLE_SNDFILE_WINDOWS_PROTOTYPES)
+inline SNDFILE* sf_open_pp(std::wstring path, int mode, SF_INFO* info) {
+    return sf_wchar_open(path.c_str(), mode, info);
+}
+#endif
```
danemtsov, line 40: Is there a benefit to taking a string instead of a path and keeping the narrow string overload around?

Otherwise a copy could be avoided (`fs::path::c_str` is equivalient to `.native().c_str()` and `fs::path::native` returns a `const string_type&`):

```suggestion
#ifdef _WIN32
#if (defined(ENABLE_SNDFILE_WINDOWS_PROTOTYPES) && ENABLE_SNDFILE_WINDOWS_PROTOTYPES)
inline SNDFILE* sf_open_pp(fs::path const& path, int mode, SF_INFO* info) {
    return sf_wchar_open(path.c_str(), mode, info);
}
#else // ENABLE_SNDFILE_WINDOWS_PROTOTYPES
#error Inconsistent path char types
#endif // ENABLE_SNDFILE_WINDOWS_PROTOTYPES
#else // _WIN32
inline SNDFILE* sf_open_pp(fs::path const& path, int mode, SF_INFO* info) { return sf_open(path.c_str(), mode, info); }
#endif // _WIN32
```

Febbe: Thanks for the catch, but the functions should take a `StringType const&` with that, automatic overload resolution will find the correct function automatically.

Febbe:
@Technius mounted files aren't a problem, they will have a valid local path specifier. But URI's won't work, they never did. Since this will require different stream overloads, for each URI type different to `file:///`. We can't use any Library relying on e.g. `f_open`.

Febbe:
Will squash, and rebase after pipeline succeeds.

src/core/audio/VorbisConsumer.cpp
```diff
@@ -16,11 +20,9 @@ auto VorbisConsumer::start(const std::string& filename) -> bool {
     sfInfo.format = SF_FORMAT_OGG | SF_FORMAT_VORBIS;
     sfInfo.samplerate = static_cast<int>(this->settings.getAudioSampleRate());
 
-    auto SNDFILE_closer = [](SNDFILE* tag) { sf_close(tag); };
-    std::unique_ptr<SNDFILE, decltype(SNDFILE_closer)> sfFile{sf_open(filename.c_str(), SFM_WRITE, &sfInfo),
-                                                              std::move(SNDFILE_closer)};
+    auto sfFile = audio::make_snd_file(file.native(), SFM_WRITE, &sfInfo);
```
danemtsov, line 23: This is the wrong place to call `.native()`, right? Should be called inside `make_snd_file` instead.
(Do I understand correctly that implicit conversion will create a copy because there is only an `operator string_type()` but no `operator string_type&()`?)

Sry for only half catching this earlier (:

Febbe: You are right, but you could do this in your PR.

PR #3785, opened 2022-02-07
Bugfix obvious enum comparison

Febbe:
`enum constant in boolean context [-Wint-in-bool-context]`


Technius:
Does this change any of the tool behaviors?

Febbe:
> Does this change any of the tool behaviors?

It depends, the expression⁰ I fixed will always evaluate to true. If the same expression¹ is made somehow before, this will change nothing. But an expression evaluating to true is definitely not what the original Autor intended there.

Technius:
This breaks double-click-to-edit on newly created LaTeX objects. The fix should be to remove the `isSelectTool` check altogether.

Technius (approved):

PR #3786, opened 2022-02-08
Replace Job ref count mutex with atomic integer

Technius:
Code cleanup to provide simpler atomic ref counting in Job with better performance.

Reviewers: since this affects a vital part of the application, please review this PR carefully.

### DIFF
--- a/src/core/control/jobs/Job.cpp
+++ b/src/core/control/jobs/Job.cpp
@@ -1,30 +1,19 @@
 #include "Job.h"
 
-#include <mutex>
+#include <gdk/gdk.h>
 
-#include <glib.h>
-#include <gtk/gtk.h>
-
-Job::Job() {}
+Job::Job(): refCount(1) {}
 
 Job::~Job() = default;
 
 void Job::unref() {
-    this->refMutex.lock();
-    this->refCount--;
-
-    if (this->refCount == 0) {
-        this->refMutex.unlock();
+    if (refCount.fetch_sub(1, std::memory_order_acq_rel) == 1) {
         delete this;
-    } else {
-        this->refMutex.unlock();
     }
 }
 
 void Job::ref() {
-    this->refMutex.lock();
-    this->refCount++;
-    this->refMutex.unlock();
+    refCount.fetch_add(1, std::memory_order_relaxed);
 }
 
 void Job::deleteJob() {
@@ -56,13 +45,7 @@ void Job::callAfterRun() {
     }
 
     this->ref();
-
     this->afterRunId = gdk_threads_add_idle(reinterpret_cast<GSourceFunc>(Job::callAfterCallback), this);
 }
 
-/**
- * After run will be called from UI Thread after the Job is finished
- *
- * All UI Stuff should happen here
- */
 void Job::afterRun() {}
--- a/src/core/control/jobs/Job.h
+++ b/src/core/control/jobs/Job.h
@@ -11,15 +11,24 @@
 
 #pragma once
 
-#include <mutex>
-#include <string>
-#include <vector>
+#include <atomic>
 
 enum JobType { JOB_TYPE_BLOCKING, JOB_TYPE_PREVIEW, JOB_TYPE_RENDER, JOB_TYPE_AUTOSAVE };
 
+/**
+ * A manually ref-counted class representing an asynchronous job to be used with
+ * the Xournal++ scheduler.
+ *
+ * Instances of Job (and its subclasses) should be constructed exclusively using
+ * the `new` keyword.
+ */
 class Job {
 public:
     Job();
+    Job(const Job&) = delete;
+    Job& operator=(const Job&) = delete;
+    Job(Job&&) = delete;
+    Job& operator=(Job&&) = delete;
 
 protected:
     virtual ~Job();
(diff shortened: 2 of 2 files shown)


src/core/control/jobs/Job.h
```diff
@@ -11,15 +11,24 @@
...
 
 enum JobType { JOB_TYPE_BLOCKING, JOB_TYPE_PREVIEW, JOB_TYPE_RENDER, JOB_TYPE_AUTOSAVE };
 
+/**
+ * A manually ref-counted class representing an asynchronous job to be used with
+ * the Xournal++ scheduler.
+ *
+ * Instances of Job (and its subclasses) should be constructed exclusively using
+ * the `new` keyword.
+ */
```
Febbe, line 24: Maybe it would be better to use `std::shared_ptr` or even std::shared_from_this (depends, if this is passed to C functions, and if this is done in an async OP; then maybe not). 

Also, a `JopPtr create(Args... args)` and a private/protected constructor will prevent the usage of stack allocations.

src/core/control/jobs/Job.cpp
```diff
@@ -1,30 +1,19 @@
...
-    } else {
-        this->refMutex.unlock();
     }
 }
 
 void Job::ref() {
-    this->refMutex.lock();
-    this->refCount++;
-    this->refMutex.unlock();
+    refCount.fetch_add(1, std::memory_order_relaxed);
```
Febbe, line 16: I am not 100% sure but:

having t1 with count == 1 fetch_add(1) and a lock free relaxed transmission to t2, cnt.fetch_add is allowed to be reordered after the transmission and vice versa. t2 reads transmission, unrefs the cnt => 0?

src/core/control/jobs/Job.h
Technius: We're not going to pull in an extra dependency for a simple refactor.

src/core/control/jobs/Job.h
```diff
@@ -78,11 +87,13 @@ class Job {
     virtual void onDelete();
```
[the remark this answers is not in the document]
Technius: This is not called during the constructor or destructor.

src/core/control/jobs/Job.h
Technius: I read over the docs for `std::shared_ptr` again and found that its ref count is actually handled with atomics, so it would be better to use `std::shared_ptr`. But that would involve quite a bit more effort.

src/core/control/jobs/Job.cpp
Technius: The atomic is only there to ensure the _ref count_ does not experience data races. It can't prevent incorrect use of the ref counting (that's what a RAII wrapper helps prevent).

The acquire/release on unref prevents a ref from being reordered after an unref:
```
t1: ref, unref     <---- ref cannot be moved after unref (potentially unsafe)
```

Relaxed is used on ref because it does not matter whether two refs are exchanged:
```
t2: ref1, ref2     <---- reorder ok: end result is ref count is added by 2
```

The atomics do not prevent bad use of reference counting:
```
t1: unref  <------ memory freed here
t2:           ref     <---- use after free
```

Technius:
Looks like `Job` is not used in too many places, so it's probably better to just replace with `shared_ptr` everywhere.

src/core/control/jobs/Job.cpp
Febbe: I meant this situation, but I think I forgot a paragraph from the memory ordering rules, since shared_ptr's do the same:

| t1 code  | t1 reordered | t2 code |
|----------|--------------|---------|
| cnt = 1  | cnt = 1      |         |
| ref      | transmit     |         |
| transmit |              | recv    |
| unref    |              | do_x()  |
|          |              | unref   |
|          | ref          |         |
|          | unref        |         |

Technius: Sending data to another thread probably involves a memory barrier, so there cannot be reordering of that form.

Technius: Well, I'm no expert on concurrency, so let me know if there's anything wrong. Otherwise this sort of PR will cause unexplainable bugs.

Febbe: I think, shared_ptr's will also have this problem. But this is generally not a concern, since shared_ptrs should not be sent to another thread without synchronization.

Code like:

```
static std::shared_ptr<X> shared_data;
static std::atomic<bool> send{};


void t1(){
    std::shared_ptr<X> my_data;
    shared_data = my_data;
    send.store(true, std::memory_order::relaxed);
    my_data.reset();
}

void t2(){
    while(!send.load(std::memory_order::relaxed));
    shared_data.reset();
}
```
is insane :D.

https://en.cppreference.com/w/cpp/atomic/memory_order is the related site. And the sequenced before part is the issue.

Febbe:
I think you can merge this, I assume there is no transmission without release-aquire semantics. 
Asked this also on https://stackoverflow.com/questions/71048755/memory-ordering-rules-what-do-i-missed-here?noredirect=1#comment125598001_71048755

Technius:
This PR is solely for replacing the mutex that protects the refcount from data races with an atomic that does the same thing. The atomic _does not prevent incorrect reference counting_ and _does not prevent data races on the actual data_. I found this quite helpful to review when thinking about the memory ordering: https://www.cs.utexas.edu/~bornholt/post/memory-models.html

PR #3791, opened 2022-02-08
Fix Console not working if not using msys2

Febbe:
This connects the console to the parent process if it is a console application.
Fixes the case, when xournalpp is started by a cmd not a msys console.

### DIFF
--- a/src/win32/console.cpp
+++ b/src/win32/console.cpp
@@ -1,12 +1,30 @@
 #include "console.h"
 
+#include <cstdio>
+
 #include <windows.h>
+//
+// #include <wchar.h>
+// #include <wincon.h>
+
 
 void attachConsole() {
     if (GetConsoleWindow() != NULL) {
         // Console is already attached.
         return;
     }
+    // https://www.tillett.info/2013/05/13/how-to-create-a-windows-program-that-works-as-both-as-a-gui-and-console-application/
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        if (GetStdHandle(STD_OUTPUT_HANDLE) != INVALID_HANDLE_VALUE &&
+            GetStdHandle(STD_ERROR_HANDLE) != INVALID_HANDLE_VALUE) {
+            // Redirect stdout and stderr to the console.
+            freopen("CONOUT$", "w", stdout);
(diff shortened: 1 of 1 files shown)


Febbe:
@jaburjak can you take a look?

Febbe:
@danemtsov this patch will connect xpp to the vscode console. I just tested it with that.

danemtsov:
I currently have a more complete fix for this in the works (e.g. will only attach to parent process' console if debugging, otherwise provide menu entry to show console window). Reasoning is that retroactively attaching to a console may interfere with the console's own output.

Febbe:
> I currently have a more complete fix for this in the works (e.g. will only attach to parent process' console if debugging, otherwise provide menu entry to show console window). Reasoning is that retroactively attaching to a console may interfere with the console's own output.

The problem with that approach would be, that the most significant information will occur on startup. In some circumstances, the app crashes before the user has the ability to start a new console.

danemtsov:
It will use the console that is created on startup. I had thought about having a command line flag to not not hide it and to not couple its lifetime to xournalpp process lifetime so it can be examined after a crash.

> Reasoning is that retroactively attaching to a console may interfere with the console's own output.

The most important bit here is that the console won't be blocked and you can still start other processes from it.

Febbe:
> The most important bit here is that the console won't be blocked and you can still start other processes from it.
This is what I'd like to avoid:

Yes, windows does shit in this case. There is no real perfect solution. I also researched possibilities, all have their drawbacks.

1. Either we use 2 executables, a .exe or a console executable, which starts the first. Drawback: You have to type xournalpp, not xournalpp.exe
2. recursive call of xournalpp.exe, when it starts from a console. Interferes with gtk_application etc. no unique process ID ...
3. starting it as console program: visible console (flash)
4. starting it as is: attached console is no-blocking

The decision depends on what the lesser evil is.
The most clean solution would probably a 2nd executable.

Technius:
I remember there were some issues with using a wrapper executable (environment variables, missing icon, path setup, etc.), so we moved _away_ from a wrapper to just the plain exe.

Febbe:
Superseded I think. Even the currently used method might not be the best.

PR #3792, opened 2022-02-08
Fix clang-tidy modernize-use-override

Febbe:
For a more modern codebase.
Used ` run-clang-tidy -fix -checks=-*,modernize-use-override -header-filter=.*`


Febbe:
@Technius this does not contain some contextual changes. 
I think we can merge it. Even when it will introduce conflicts, they will be trivially to resolve.

Technius (approved):
LGTM, I skimmed through quickly and I didn't see anything go wrong.

Issue #1170, opened 2019-05-22
Pressure curve


JJones780:
_Originally posted by @schuhumi in https://github.com/xournalpp/xournalpp/pull/1088#issuecomment-492230778_

 It would be great if it was possible to set a pressure curve just like in Krita for example: https://docs.krita.org/en/reference_manual/preferences/tablet_settings.html

That way you can:
 - Remove tics
 - adjust the pressure you want to put on the stylus for maximum line width
 - if the line can have curvature: adjust if the stylus feels like it had a soft/hard tip

This is the feature that I miss most in Xournal++

_Originally posted by @schuhumi in https://github.com/xournalpp/xournalpp/pull/1088#issuecomment-492230778_

JJones780:
I'll close this and see if I can get the original author to create the issue.

schuhumi:
Thank you very much for making this an official enhancement @JJones780 !

I didn't move it to a feature request myself yet because I want to propose something. Have a look at this screenshot:
https://i.imgur.com/G8Hn2ev.png
This is from an Android app called "Lecturenotes", which I used a lot on Android X86. What you can see is that it is very easy to resemble behaviours of real-world drawing tools like pens, pencils, highlighters etc.
The setup dialog for each color looks like this (annotated translations):
https://i.imgur.com/lS7cnfP.png

What you can do there is basically the same as (or rather a useful subset of) having a pressure curve for width and another for alpha. The developers of Lecturenotes just hide it behind 3 sliders.

I know having individual curves for thickness and alpha for each color is a long shot and a big effort. But maybe when you look at the screenshots you can see how natural and appealing the results look. When you print such a document with a color printer you can barely tell the difference. I just wanted to throw this in before someone implements a global pressure curve, so it can be considered beforehand.

Thank you all for this great application!!

LittleHuba:
https://github.com/draekko-rand/Gtk3Curve

Take a look at this widget. It could be used as base for the implementation.

JJones780:
@schuhumi The LectureNotes app is interesting.  We could have an additional setting for each stroke to tell it how the existing pressure information is to be interpreted - so I guess this wouldn't be too much of a change to the current implementation. We'd have to consider a more flexible way to select pens. 

I'm glad you presented it as a feature from another note-taking app... xournal++ isn't meant to be an artists tool, in my opinion.

@LittleHuba - thanks for pointing out the Gtk3Curve!

[...]

PR #3798, opened 2022-02-09
[Draft] Add simple pressure curve adjustment setting (softness)

gurrgur:
This pull request adds a stylus setting called "Softness", which enables users to increase or decrese the softness of the tip by applying a static transfer function to the pressure level. #1170 proposes more complex and in-depth pressure curve adjustments but in my opinion, for most users this simple softness setting could suffice.

The transfer function looks like this for various softness settings:

pressure-curve-plot

At 1 and -1 you get a circular arc, between -1 and 1 the shape becomes similar to a straight line and for abs(softness) > 1 the curve gets steeper. Finally, 0 yields the identity.

Here is a writing sample comparing various softness settings using a medium pen and no system or driver level pressure curve adjustments:

xournalpp-softness-example


ui/settings.glade
```diff
@@ -75,6 +75,13 @@
     <property name="step-increment">0.05</property>
     <property name="page-increment">0.5</property>
   </object>
+  <object class="GtkAdjustment" id="adjustmentPressureCurve">
```
Technius, line 78: The slider needs a tooltip explaining its effect.

Technius (commented):
Thanks for this PR. I really like the simplicity.

gurrgur:
Not sure if the explanation of the transfer function is too detailed but I think it useful to have information about what it does and where it came from.

Technius:
The detailed comment is exactly what I was looking for, thanks.

rolandlo:
I have my doubts about the concept of this PR. The pressure is cut off at the value 1.0 (when the curve parameter is non-zero). I see two issues with that:
- There are styluses that report pressure well above 1 (like 6-7 or so). 
- There is the pressure multiplier. Personally I don't want to apply maximal pressure only to get strokes of the same width as if they were drawn without pressure. I want to have average pressure near that value. Therefore I use the pressure multiplier to get pressures around 0-2 with average pressure near 1. Now if you cut off everything above 1 that will not be a nice experience at all.

gurrgur:
Your concern with the limit imposed on using the pressure multiplier as means to tune the working point is understandable. I always thought the multiplier was meant to normalize pressure to [0, 1] but it seems I was wrong.

Perhaps the pressure multiplier could be replaced with a working point setting. The working point would be defined by the mapping `[input pressure level p --> 1.0]`. Any curve setting (e.g. softness, steepness) would act in vicinity of that working point while keeping the mapping `[input pressure level p --> 1.0]` intact. When no curve adjustments are in effect, the working point setting would result in a pressure multiplier of 1 divided by working point input pressure.

PR #3810, opened 2022-02-13
Fix TextEditor repaint bounds

Technius:
This PR fixes a performance issue with the text tool.


bhennion:
LGTM

Technius:
Actually, I'll rebase this onto `release-1.1` instead so we can get it into a `1.1.2` release.

Technius:

PR #3823, opened 2022-02-16
Fix capitalization

flleeppyy:


Technius (approved):
LGTM, thanks for the fix.

Note: merging into master branch and not release-1.1, as it's unclear whether the translation files will be updated for release-1.1.

Issue #977, opened 2019-03-10
Ctrl+shift+(Left/right arrow) does not select text, it changes layers


siliconninja:
I need to select a word of text (for example pressing Ctrl+shift+right arrow with the cursor at word1 selects "word1"), so I press Ctrl+Shift+left arrow or Ctrl+Shift+right arrow and that sometimes changes the layer instead. Is there a way to possibly change this default shortcut or possibly include it in the scripting feature (#919)?
I get this issue when I draw or click on the bottom > "New Layer". If I just use the background layer, this works perfectly.
I don't have much time to fully focus on contributing code because I'm in school, but I'll see what I can do if I get some free time. There is no error because the switching layers is probably configured with these keyboard shortcuts by default.

It might have something to do with ACTION_GOTO_PREVIOUS_LAYER and ACTION_GOTO_NEXT_LAYER:
https://github.com/xournalpp/xournalpp/blob/a071b52ca614e815857fe12f06d8195ecffe8f79/src/enums/ActionType.enum.h

LittleHuba:
We can do this when we work on #841

Technius:
It's much more common to edit text than to switch layers, so I think it would be better to have <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>&#8592;</kbd> and <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>&#8594;</kbd> select text when editing a text object The current behavior can remain for everything else.

LittleHuba:
This can be handled with #919

Technius:
This is actually a bit tricky, since the layer menu item is receiving (and handling) the <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>←</kbd> and <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>→</kbd> keypresses before the event can be handled by the `TextEditor`.

twitwi:
I encountered a comparable issue while experimenting with plugins.
A plugin that registers a plain key (let's say "e") captures this key even if the user is editing text, which technically makes it impossible to type the letter "e".

I found a workaround on https://stackoverflow.com/a/61584287/2297277 which I tested (I'm not sure where the best place would be for this code so I did not create a pull request):

```C
g_signal_connect(window,"key-press-event",G_CALLBACK(gtk_window_propagate_key_event),NULL);
g_signal_connect(window,"key-release-event",G_CALLBACK(gtk_window_propagate_key_event),NULL);
```
[...]

PR #3841, opened 2022-02-20
Make TextEditor pass events to IM first

Rio6:
When using gcin as IM, I usually use arrows key to select candidates, or escape key to leave candidate menu. But in xournal++ text editor if I wanted to do the same, the cursor gets moved instead with arrow keys, and the whole text box gets closed with escpae key. This PR move event handling of IM before the text editor event handling and escap key handling so character selection.

With this PR, gcin now works pretty well with xournal++, so it can probably be removed from here https://github.com/xournalpp/xournalpp/blob/3d8dc4bc8b1cbc1230062937dd9ff19039834b00/src/core/control/XournalMain.cpp#L306-L308
(I don't have xim installed to check, but this might fix its issue as well).

@tattsan tagging you because you were testing with xim on my other PR, maybe you can help me test this one as well. Much thanks.

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -624,6 +624,20 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
 }
 
 auto XojPageView::onKeyPressEvent(GdkEventKey* event) -> bool {
+    if (this->textEditor) {
+        if (this->textEditor->onKeyPressEvent(event)) {
+            return true;
+        }
+    } else if (this->inputHandler) {
+        if (this->inputHandler->onKeyEvent(event)) {
+            return true;
+        }
+    } else if (this->verticalSpace) {
+        if (this->verticalSpace->onKeyPressEvent(event)) {
+            return true;
+        }
+    }
+
     // Esc leaves text edition
     if (event->keyval == GDK_KEY_Escape) {
         if (this->textEditor) {
@@ -637,20 +651,6 @@ auto XojPageView::onKeyPressEvent(GdkEventKey* event) -> bool {
 
         return false;
     }
-
-    if (this->textEditor) {
-        return this->textEditor->onKeyPressEvent(event);
-    }
-
-
-    if (this->inputHandler) {
-        return this->inputHandler->onKeyEvent(event);
-    }
-
-    if (this->verticalSpace) {
-        return this->verticalSpace->onKeyPressEvent(event);
-    }
-
 
     return false;
 }
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -303,7 +303,7 @@ void ensure_input_model_compatibility() {
     const char* imModule = g_getenv("GTK_IM_MODULE");
     if (imModule != nullptr) {
         std::string imModuleString{imModule};
-        if (imModuleString == "xim" || imModuleString == "gcin") {
+        if (imModuleString == "xim") {
             g_warning("Unsupported input method: %s", imModule);
         }
     }
--- a/src/core/gui/TextEditor.cpp
+++ b/src/core/gui/TextEditor.cpp
@@ -253,10 +253,6 @@ auto TextEditor::imDeleteSurroundingCallback(GtkIMContext* context, gint offset,
 }
 
 auto TextEditor::onKeyPressEvent(GdkEventKey* event) -> bool {
-    if (gtk_bindings_activate_event(G_OBJECT(this->textWidget), event)) {
-        return true;
-    }
-
     bool retval = false;
     bool obscure = false;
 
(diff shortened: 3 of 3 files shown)


tattsan:
**Non xim environment**
I've tested the PR on `uim`, `ibus`, `fcitx` of their native (`non xim`) mode, but I didn't notice any issue.

**Xim general**
There's many implementations for `xim`, but I've tested only xim mode in `uim` -- say `uim-xim` --, and xim mode in `fcitx` -- say `fcitx-xim`. In your previous #3511, I've reported that `xim` had an issue concerning its candidate window, but now I've confirmed that this is a `uim-xim` specific issue, not happens for `fcitx-xim`. So I think we can say Xournal++ basically supports `xim`.

**Uim-xim**
`Uim-xim` can't get cursor position in Xournal++, so `xim-uim` displays it's candidate window at unexpected place.
I've resolved the similar issue in #2851 for several IME's, but the issue remains unresolved for `xim-uim` .

This behavior is the same before and after the current PR , and not a target of the PR.

**Other contexts**
We also have to consider the input priority for global binding and TextEditor bindings. I use control-G for escaping (like Emacs) in IME, but it is sometimes recognized as `Go to page` command binded globally. Of course, this issue is not the target of the current PR.

Rio6:
Sounds like not all xim implementation are working well enough, so I'll just remove gcin from giving warning message and keep xim there.

For Ctrl-G, I'm not familiar enough with gtk to figure out where that's being handled. Does it happen both before and after this PR? I thought it would be handled here
https://github.com/xournalpp/xournalpp/blob/9b5bfe70aab98bd19a4409db453a55c09886161a/src/core/gui/TextEditor.cpp#L271
but if this PR doesn't fix it then obviously it's not.

Anyways, much thanks for testing this @tattsan

Technius:
> For Ctrl-G, I'm not familiar enough with gtk to figure out where that's being handled. Does it happen both before and after this PR? I thought it would be handled here

If I recall, the application window itself is first in the priority list for keybindings, so some keybindings may not be ever seen by the text editor. For example, `Shift+Ctrl+Left` never gets passed to the text editor (#977). As @tattsan mentioned, this issue would be best fixed in another PR.

src/core/gui/PageView.cpp
```diff
@@ -624,6 +624,20 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
...
+    } else if (this->inputHandler) {
+        if (this->inputHandler->onKeyEvent(event)) {
+            return true;
+        }
+    } else if (this->verticalSpace) {
+        if (this->verticalSpace->onKeyPressEvent(event)) {
+            return true;
+        }
+    }
+
```
Technius, line 640: There may be unintended consequences due to the change in key handling behavior here, but I think these should be _good_ consequences, as some events that we've ignored in the past will now be forwarded to the appropriate places.

src/core/gui/PageView.cpp
```diff
@@ -624,6 +624,20 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
 }
 
 auto XojPageView::onKeyPressEvent(GdkEventKey* event) -> bool {
+    if (this->textEditor) {
+        if (this->textEditor->onKeyPressEvent(event)) {
+            return true;
+        }
```
Technius, line 630: Nitpicking, but these can be simplified to
```suggestion
    if (this->textEditor && this->textEditor->onKeyPressEvent(event)) {
        return true;
```

Technius (approved):
I trust @tattsan's judgment on this one, and the changes to PageView look good to me. I just have a few minor nitpicks.

src/core/gui/PageView.cpp
Technius: (no need to take action on this)

src/core/gui/PageView.cpp
Rio6: The behavior would be a bit different because the use of `else if`. I try to make it close to the original behavior, ie. only one of the textEditor, inputHandler, and verticalSpace gets called.

tattsan:
> Does it happen both before and after this PR?

Yes. As @Technius mentioned, #977 would be the right place to discuss on this point.

src/core/gui/PageView.cpp
Technius: The behavior in the PR is already different from original behavior because it's checking the result of `onKeyPressEvent` instead of returning it directly (the new behavior is better though).

The suggested change is just an equivalent, but simpler version of the new behavior.

Technius:
Thanks for adding the comment. As a last step, could you squash the following commits:
* `2d2ea1b` & `cec0738` into a single commit
* `3d8dc4b` & `9b5bfe7` into a single commit
* `fdb14f1` is fine as-is
Or if you prefer, I can do it for you.

Rio6:
Squashed

src/core/gui/PageView.cpp
Rio6: With my current change, when `textEditor->onKeyPressEvent` returns false `inputHandler->onKeyEvent` does not get called. But with the suggested change, it does get called. (similiar thing with `verticalSpace->onKeyPressEvent`) The first behaviour matches the original was what I meant.

Unless you mean that it is better for the later handlers to be called when the previous ones return false? If that's the case I can update the code.

Technius (approved):
LGTM. Merging in 24 hours if there are no objections.

PR #3851, opened 2022-02-22
WIP: Support PDF Links

personalizedrefrigerator:
# Summary
 * Rebase of https://github.com/Technius/xournalpp/pull/1

# Status
 * [ ] Use the new PDF tool to allow opening links, instead of a different pop-up GUI

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -854,7 +854,7 @@ bool XojPageView::displayLinkPopover(std::shared_ptr<XojPdfPage> page, double pa
             size_t pageId = doc->findPdfPage(pdfPage);
             doc->unlock();
 
-            GtkWidget* button = gtk_button_new_with_label(FC(_F("Scroll to page {1}") % pageId));
+            GtkWidget* button = gtk_button_new_with_label(FC(_F("Scroll to page {1}") % (pageId + 1)));
             gtk_container_add(GTK_CONTAINER(box), button);
 
             g_signal_connect(
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -3,7 +3,9 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdlib>
+#include <iomanip>
 #include <memory>
+#include <tuple>
 
 #include <gdk/gdk.h>
 #include <glib/gprintf.h>
@@ -30,6 +32,7 @@
 #include "gui/widgets/XournalWidget.h"
 #include "model/Image.h"
 #include "model/Layer.h"
+#include "model/LinkDestination.h"
 #include "model/PageRef.h"
 #include "model/Stroke.h"
 #include "model/Text.h"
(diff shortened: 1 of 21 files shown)


src/core/gui/PageView.h
```diff
@@ -188,6 +190,26 @@ class XojPageView: public Redrawable, public PageListener {
...
+    /**
+     * @brief Display a popover with link-related actions, if one
+     *  is at a given location on the page.
+     *
+     * @param page is the current page
+     * @param targetX
+     * @param targetY the link must contain (targetX, targetY)
+     * @returns true iff a URI link exists near/at (targetX, targetY) => a popover was shown
+     */
+    bool displayLinkPopover(std::shared_ptr<XojPdfPage> page, double targetX, double targetY);
```
Technius, line 211: I think this function could be made static (with a few modifications in the implementation).

src/core/gui/PageView.cpp
```diff
@@ -793,6 +816,96 @@ void XojPageView::drawLoadingPage(cairo_t* cr) {
...
+                        std::shared_ptr<LinkDestination> dest;
+                        GtkWidget* popover;
+                        std::tie(self, dest, popover) = *state;
+
+                        self->getXournal()->getControl()->getScrollHandler()->scrollToLinkDest(*dest);
+                        gtk_popover_popdown(GTK_POPOVER(popover));
+
+                        free(state);
+                    }),
+                    new std::tuple(this, dest, popover));
```
Technius, line 879: Why use `this` instead of the the `page` parameter (and vice versa)?

Technius (commented):
Thanks for picking up my work! I'm very excited to see these changes. In general, I think we should focus on the feature in this PR (as you've been doing so far) and try to put off the temptation of refactoring to a later PR. I left some early feedback, which I hope you'll find helpful.

Regarding the integration with the PDF toolbox, I think we'll need to spend some time on the UX in follow up work. We'll want to show the popover when it's convenient, but we also want to avoid annoying the user with too much popover (e.g., on a page with a lot of links). Here are some ideas for when to show the popover:
* When the PDF selection is finalized, show a popover if there is exactly one link in the selection and the link is fully contained within the selection region.
* When there is an existing selection, show a popover if the user clicks on a link that has a nonempty intersection with the selection region.

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -70,6 +72,47 @@ void Sidebar::buttonClicked(GtkToolButton* toolbutton, SidebarPageButton* button
...
+
+    doc->lock();
+    XojPdfPageSPtr pdf = doc->getPdfPage(pdfPage);
+    doc->unlock();
+
+    if (pdf) {
+        auto page = std::make_shared<XojPage>(pdf->getWidth(), pdf->getHeight());
+        page->setBackgroundPdfPageNr(pdfPage);
+        control->insertPage(std::move(page), position);
+    }
```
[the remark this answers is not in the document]
Technius: Oh, `control->insertPage` already locks the document, so you can ignore this.

We need to come up with a clear and consistent guide for when and where the document should be locked.

Technius:
I created a rebased version in #4412, and will try to wrap up this work there so that it doesn't bitrot and get thrown away.

Technius:
Superceded by #4412, which was merged.

PR #3882, opened 2022-03-01
Properly handle blocking of devices for setsquare

rolandlo:
This PR make the setsquare handle blocking of devices properly, which is currently used for the automatic hand recognition (an option that can be turned on in the preferences, Touchscreen panel). It fixes an issue reported by @lxdb in #3082, where using the stylus resulted in input events temporarily not getting processed at all (since everything was blocked by the setsquare input handler). 

@lxdb Please test the PR on your device (with automatic hand recognition turned on).

### DIFF
--- a/src/core/gui/inputdevices/InputContext.cpp
+++ b/src/core/gui/inputdevices/InputContext.cpp
@@ -7,6 +7,7 @@
 #include "control/DeviceListHelper.h"
 
 #include "InputEvents.h"
+#include "SetsquareInputHandler.h"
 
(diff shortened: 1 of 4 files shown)


lxdb:
After merging this PR everything works as usual with internal hand recognition enabled. Thank you for solving the problem so quickly!

rolandlo:
> @rolandlo After merging this PR everything works as usual with internal hand recognition enabled. Thank you for solving the problem so quickly!

Thanks for testing @lxdb 

PR #3892, opened 2022-03-06
Updated spec file

luyatshimbalanga:
Here is the updated spec file. 
Also add ` https://copr.fedorainfracloud.org/webhooks/github/27610/d8866b0a-4783-4d65-a0e4-0d46f0961d12/ ` in the webhook so the build will sync with  the commit. See 
Screenshot from 2022-03-06 11-36-08

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,178 @@
+# Contributing to Xournal++
+
+Thanks for your interest in contributing to Xournal++! The project is run
+completely by volunteers in their spare time, so any contribution--no matter how
+small--is greatly appreciated. In this file, we'll outline how you can
+contribute to the project.
+
+## Overview
+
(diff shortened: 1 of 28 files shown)


Technius:
Does #3896 fix the gtest issues?

luyatshimbalanga:
> Does #3896 fix the gtest issues?

gtest support is disabled as a workaround. You can turn it on once the codes from #3896 are merged.

Technius:
Merging in 24 hours if there are no objections. (By the way, I merged #3896, so you may want to rerun the build just to check).

TODO:
* [ ] Enable webhook as suggested in the OP

luyatshimbalanga:
The [build is successful](https://copr.fedorainfracloud.org/coprs/g/design-suite/xournalpp/build/3707080/) with the recent commit using the gtest system.

Technius:
Thanks for testing!

PR #3894, opened 2022-03-07
fix "suggest parantheses around && within ||"

rolandlo:
Fixes some compiler warnings related to "suggest parantheses around && within ||"

### DIFF
--- a/src/core/gui/inputdevices/TouchDrawingInputHandler.cpp
+++ b/src/core/gui/inputdevices/TouchDrawingInputHandler.cpp
@@ -32,7 +32,7 @@ auto TouchDrawingInputHandler::handleImpl(InputEvent const& event) -> bool {
 
     // Do we need to end the touch sequence?
     bool mustEnd = event.type == BUTTON_RELEASE_EVENT;
-    mustEnd = mustEnd || event.type == GRAB_BROKEN_EVENT && this->deviceClassPressed;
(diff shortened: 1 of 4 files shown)


rolandlo:
merging in 24 hours if no objections are raised

src/core/gui/inputdevices/TouchDrawingInputHandler.cpp
```diff
@@ -32,7 +32,7 @@ auto TouchDrawingInputHandler::handleImpl(InputEvent const& event) -> bool {
 
     // Do we need to end the touch sequence?
     bool mustEnd = event.type == BUTTON_RELEASE_EVENT;
-    mustEnd = mustEnd || event.type == GRAB_BROKEN_EVENT && this->deviceClassPressed;
+    mustEnd = mustEnd || (event.type == GRAB_BROKEN_EVENT && this->deviceClassPressed);
```
Technius, line 35: Might be cleaner as
```suggestion
    mustEnd |= (event.type == GRAB_BROKEN_EVENT && this->deviceClassPressed);
```

Technius (approved):

src/core/gui/inputdevices/TouchDrawingInputHandler.cpp
rolandlo: That doesn't actually compile.

Technius: Ah, probably should've been single pipe equal. Well, it's a very minor thing--no worries then.

PR #3911, opened 2022-03-13
Remove unecessary mask in StrokeHandler

bhennion:
This PR improves performance and mem usage in StrokeHandler for strokes that require a full redraw at each event, by removing an unecessary mask.

It also fixes the display of single-point strokes that require a full redraw, and removes the corresponding warnings.

This is ready for a review.

### DIFF
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -360,10 +360,10 @@ void StrokeHandler::onButtonPressEvent(const PositionInputData& pos) {
 
         cairo_surface_set_device_offset(surfMask, (0.5 * strokeWidth - visibleRect->x) * ratio,
                                         (0.5 * strokeWidth - visibleRect->y) * ratio);
+        cairo_surface_set_device_scale(surfMask, ratio, ratio);
 
         crMask = cairo_create(surfMask);
 
-        cairo_scale(crMask, ratio, ratio);
 
         // Paint the starting point
         this->paintDot(this->buttonDownPoint.x, this->buttonDownPoint.y,
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -22,48 +22,43 @@ using namespace xoj::util;
 
 guint32 StrokeHandler::lastStrokeTime;  // persist for next stroke
 
-
 StrokeHandler::StrokeHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page):
         InputHandler(xournal, redrawable, page),
         snappingHandler(xournal->getControl()->getSettings()),
         stabilizer(StrokeStabilizer::get(xournal->getControl()->getSettings())) {}
 
-StrokeHandler::~StrokeHandler() {
-    destroySurface();
-}
-
 void StrokeHandler::draw(cairo_t* cr) {
-    if (!stroke) {
-        return;
-    }
+    assert(stroke && stroke->getPointCount() > 0);
 
-    if (this->fullRedraw) {
-        /**
-         * Erase the mask entirely in this case
-         */
-        cairo_set_operator(crMask, CAIRO_OPERATOR_CLEAR);
-        cairo_paint(crMask);
-
-        auto context = xoj::view::Context::createColorBlind(crMask);
-        strokeView->draw(context);
-    }
-
-    // set the stroke color and blend mode
-    if (stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER) {
-        if (auto fill = stroke->getFill(); fill != -1) {
-            Util::cairo_set_source_rgbi(cr, stroke->getColor(), static_cast<double>(fill) / 255.0);
+    auto setColorAndBlendMode = [stroke = this->stroke, cr]() {
+        if (stroke->getToolType() == STROKE_TOOL_HIGHLIGHTER) {
+            if (auto fill = stroke->getFill(); fill != -1) {
+                Util::cairo_set_source_rgbi(cr, stroke->getColor(), static_cast<double>(fill) / 255.0);
+            } else {
+                Util::cairo_set_source_rgbi(cr, stroke->getColor(), xoj::view::StrokeView::OPACITY_HIGHLIGHTER);
+            }
+            cairo_set_operator(cr, CAIRO_OPERATOR_MULTIPLY);
         } else {
-            Util::cairo_set_source_rgbi(cr, stroke->getColor(), xoj::view::StrokeView::OPACITY_HIGHLIGHTER);
+            Util::cairo_set_source_rgbi(cr, stroke->getColor());
+            cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
         }
-        cairo_set_operator(cr, CAIRO_OPERATOR_MULTIPLY);
+    };
+
+    if (this->mask) {
+        setColorAndBlendMode();
+        cairo_mask_surface(cr, mask->surf, 0, 0);
     } else {
-        Util::cairo_set_source_rgbi(cr, stroke->getColor());
-        cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
+        if (this->stroke->getPointCount() == 1) {
+            // drawStroke does not handle single dots
+            const Point& pt = this->stroke->getPoint(0);
+            double width = this->stroke->getWidth() * (this->hasPressure ? pt.z : 1.0);
+            setColorAndBlendMode();
+            this->paintDot(cr, pt.x, pt.y, width);
+        } else {
+            strokeView->draw(xoj::view::Context::createDefault(cr));
+        }
     }
-
-    cairo_mask_surface(cr, surfMask, 0, 0);
 }
-
 
 auto StrokeHandler::onKeyEvent(GdkEventKey* event) -> bool { return false; }
 
@@ -99,8 +94,12 @@ void StrokeHandler::paintTo(const Point& point) {
                 this->stroke->setLastPressure(point.z);
                 this->firstPointPressureChange = true;
 
-                // Paint the dot for the user to see
-                this->paintDot(endPoint.x, endPoint.y, this->stroke->getWidth() * point.z);
+                double width = this->stroke->getWidth() * point.z;
+                if (mask) {
+                    this->paintDot(mask->cr, endPoint.x, endPoint.y, width);
+                }
+                // Trigger a call to `draw`. If mask == nullopt, the `paintDot` is called in `draw`
+                this->redrawable->repaintRect(endPoint.x - 0.5 * width, endPoint.y - 0.5 * width, width, width);
             }
             return;
         }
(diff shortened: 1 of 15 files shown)


src/core/control/tools/StrokeHandler.cpp
```diff
@@ -32,26 +32,24 @@ StrokeHandler::~StrokeHandler() {
...
-        return;
-    }
-
-    if (this->fullRedraw) {
-        /**
-         * Erase the mask entirely in this case
-         */
-        cairo_set_operator(crMask, CAIRO_OPERATOR_CLEAR);
-        cairo_paint(crMask);
+    assert(stroke && stroke->getPointCount() > 0);
```
Technius, line 35: I'm conflicted about this assertion. If I understand correctly, on one hand, we want to avoid having to draw a stroke if the user has not initiated stroke input. On the other hand, it does not seem obvious to me that draw is only invoked in the above situation, which means that a user could get segfaults when drawing stuff.

The assertion can be improved as follows:
```suggestion
    g_assert(stroke && stroke->getPointCount() > 0 && "can only draw a stroke if user has initiated stroke input!");
```

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -363,20 +349,39 @@ void StrokeHandler::onButtonDoublePressEvent(const PositionInputData& pos) {
...
+    const double ratio = xournal->getZoom() * static_cast<double>(xournal->getDpiScaleFactor());
+
+    std::unique_ptr<Rectangle<double>> visibleRect(xournal->getVisibleRect(redrawable));
+
+    // We add a padding to limit graphical bugs when scrolling right after completing a stroke
+    const double strokeWidth = this->stroke->getWidth();
+    const int width = static_cast<int>(std::ceil((visibleRect->width + strokeWidth) * ratio));
+    const int height = static_cast<int>(std::ceil((visibleRect->height + strokeWidth) * ratio));
+
+    surfMask = cairo_image_surface_create(CAIRO_FORMAT_A8, width, height);
```
Technius, line 378: I think `surfMask` needs to be destroyed first if it already exists, otherwise there may be a memory leak.

Also, would it be better to use `gdk_window_create_similar_image_surface`?

src/core/control/tools/StrokeHandler.h
```diff
@@ -72,6 +72,8 @@ class StrokeHandler: public InputHandler {
     SnapToGridInputHandler snappingHandler;
 
 private:
+    void createMask();
```
Technius, line 75: This needs a doc comment. Also, I think this should be renamed to some more descriptive like `createStrokeRenderBuffer`. I get that technically we're applying a cairo mask to the page render buffer surface, but that obscures the real intention (we draw the incoming points to a buffer that we can use for full repaint).

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -363,20 +349,39 @@ void StrokeHandler::onButtonDoublePressEvent(const PositionInputData& pos) {
 
 void StrokeHandler::destroySurface() {
     if (surfMask || crMask) {
+        assert(!this->fullRedraw);
```
Technius, line 352: What is this assert guarding against?

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -89,8 +87,12 @@ void StrokeHandler::paintTo(const Point& point) {
...
                 this->firstPointPressureChange = true;
 
-                // Paint the dot for the user to see
-                this->paintDot(endPoint.x, endPoint.y, this->stroke->getWidth() * point.z);
+                double width = this->stroke->getWidth() * point.z;
+                if (!this->fullRedraw) {
+                    // Paint the dot for the user to see
+                    this->paintDot(crMask, endPoint.x, endPoint.y, width);
+                }
+                this->redrawable->repaintRect(endPoint.x - 0.5 * width, endPoint.y - 0.5 * width, width, width);
```
Technius, line 95: Do we need to repaint around the end point if `fullRedraw` is true? The endpoint is only drawn if false.

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -330,29 +332,13 @@ void StrokeHandler::onButtonPressEvent(const PositionInputData& pos) {
...
-
-        // Paint the starting point
-        this->paintDot(this->buttonDownPoint.x, this->buttonDownPoint.y,
-                       this->hasPressure ? this->stroke->getWidth() * pos.pressure : this->stroke->getWidth());
+    double width = this->hasPressure ? this->stroke->getWidth() * pos.pressure : this->stroke->getWidth();
+    if (!this->fullRedraw) {  // Strokes that require a full redraw don't need a mask
+        this->createMask();
+        this->paintDot(crMask, this->buttonDownPoint.x, this->buttonDownPoint.y, width);
     }
+    this->redrawable->repaintRect(this->buttonDownPoint.x - 0.5 * width, this->buttonDownPoint.y - 0.5 * width, width,
```
Technius, line 340: See my above question about full repaint.

src/core/gui/PageView.cpp
```diff
@@ -853,36 +853,32 @@ void XojPageView::paintPageSync(cairo_t* cr, GdkRectangle* rect) {
...
         cairo_save(cr);
-        cairo_scale(cr, zoom, zoom);
         this->search->paint(cr, rect, zoom, getSelectionColor());
         cairo_restore(cr);
     }
 
     if (this->inputHandler) {
         cairo_save(cr);
-        int dpiScaleFactor = xournal->getDpiScaleFactor();
-        cairo_scale(cr, 1.0 / dpiScaleFactor, 1.0 / dpiScaleFactor);
```
Technius, line 885: Isn't scaling by `1.0 / dpiScaleFactor` different from scaling by `zoom`?

Technius (commented):
I'm not familiar with the stroke rendering and handling code, but I took a look and left some feedback.

* Nice work with handling the zoom more uniformly; this will help avoid some confusion in the future. However, I would like to see doc comments (on the appropriate methods) explaining when we need to multiply by zoom and when we do not need to multiply by zoom.
* Can you clarify what you mean by "the corresponding warnings" that occur for single point strokes?

Further comments on `StrokeHandler.{cpp,h}` in general (no need to take action on these):
* I don't really understand what the `fullRedraw` flag is and when it is set. We should add doc comments that explain the logic in the future.
* We need to explain what the "masking surface" is in more detail. I had to read through the file multiple times to understand that it is a temporary buffer used to render the stroke that is currently being drawn.

bhennion:
Thanks for the feedback!

> I'm not familiar with the stroke rendering and handling code, but I took a look and left some feedback.
> 
> * Nice work with handling the zoom more uniformly; this will help avoid some confusion in the future. However, I would like to see doc comments (on the appropriate methods) explaining when we need to multiply by zoom and when we do not need to multiply by zoom.

I'll try and add comments. Basically, the idea is to keep the following assumption true:

- The cairo context fed to `draw`/`paint` is always in page coordinates (i.e. not further scaling or offset is ever needed to paint something in `draw`).

This all thing could be simplified by applying a `cairo_surface_set_device_scale` on the `XojPageView::crBuffer` upon creation, and removing most of the `cairo_scale`'s. It's however not quite that simple (because the zoom changes), and the necessary changes for that go to deep into the view refactor. I'd rather leave that either to a later PR, or to @LittleHuba.

> * Can you clarify what you mean by "the corresponding warnings" that occur for single point strokes?

Try starting a filled stroke, and look at the terminal output. You should get some 
```
** (xournalpp:29090): WARNING **: 08:54:56.617: DocumentView::drawStroke empty stroke...
```
This is due to a view.drawStroke being called on a stroke with less than 2 points.

> Further comments on `StrokeHandler.{cpp,h}` in general (no need to take action on these):
> 
> * I don't really understand what the `fullRedraw` flag is and when it is set. We should add doc comments that explain the logic in the future.
> * We need to explain what the "masking surface" is in more detail. I had to read through the file multiple times to understand that it is a temporary buffer used to render the stroke that is currently being drawn.

src/core/control/tools/StrokeHandler.cpp
bhennion: `repaintRect` triggers the call to `draw`, where the `paintDot` will be called in the case we don't use the mask.
I'll add a comment.

src/core/control/tools/StrokeHandler.cpp
bhennion: Here for historic reasons :) Will get rid of it

src/core/control/tools/StrokeHandler.cpp
bhennion: I don't think `surfMask` should ever exist before that.
The entire class was structured before the switch to `GTK3`. Before that, the InputHandlers used to survive a single stroke creation: there was one IntpuHandler of each type for each page.
Now, the class is created upon ButtonPressEvent (see `XojPageView::onButtonPressEvent`). An entire refactor of all the InputHandlers would help clearing all that out: someday, maybe.

I'll add an assert.

bhennion: About `gdk_window_create_similar_image_surface`: at this point, we do not have access to any surface to mimic. We only have a surface in the `draw` loop.

src/core/control/tools/StrokeHandler.cpp
bhennion: I don't think `draw` can ever be called if the stroke has not yet been created. The stroke is created, with an initial point, at the same time as the StrokeHandler is created (see `XojPageView::onButtonPressEvent`).

bhennion:
I added a new subclass StrokeHandler::Mask to handler construction and destruction of the mask safely. The mask is now an std::optional<Mask>. The existence of a mask also replaces the flag fullRedraw (== !mask).

Also took care of your other comments (I think).

bhennion:
I rebased on the latest master.

I also simplified further the zoom handling in `PageView::paintPageSync`. Now, the VerticalSpaceTool is fed, like every other tool, a cairo context set up in page coordinates.

@Technius You can go a rereview it all, whenever you have time.

LittleHuba:
> This all thing could be simplified by applying a cairo_surface_set_device_scale on the XojPageView::crBuffer upon creation, and removing most of the cairo_scale's. It's however not quite that simple (because the zoom changes), and the necessary changes for that go to deep into the view refactor. I'd rather leave that either to a later PR, or to @LittleHuba.

I think you are doing an awesome job here. It is also a very good idea to keep your PRs small. Better have five small incremental PRs than one big PR. They will be easier to review and faster to merge. Then I see no reason why you couldn't approach this task yourself :slightly_smiling_face:

bhennion:
> I think you are doing an awesome job here. It is also a very good idea to keep your PRs small. Better have five small incremental PRs than one big PR. They will be easier to review and faster to merge. Then I see no reason why you couldn't approach this task yourself slightly_smiling_face

I'll try to keep the PRs contained :-)

src/core/gui/PageView.cpp
Technius: Please add a doc comment somewhere that explicitly states this assumption ((maybe in PageView? not sure where an appropriate place would be). It will help greatly when we revisit the code!

Technius (approved):
LGTM except one minor comment. I did not have time to test, but I didn't see any obvious problems with the code.

bhennion:
Thanks @Technius for the review!

I added a comment in PageView. Ideally (maybe in a follow up PR, someday), I think all those painters should inherit from a common class, where this comment should be moved.

PR #3940, opened 2022-04-02
To make the cursor invisible on eraser motion only (not in erase acti…

a-mehrabian:
**Context:**
Chances are, The user does not want to see the annoying cursor all the time when the stylus cursor type is set to STYLUS_CURSOR_NONE.
This works fine on pen but eraser. Especially when the stylus is in the proximity of the surface. 
However, the eraser cursor should reappear on in-erase action as it provides the estimation of where the eraser is being applied.

**Solution:**
To implement the logic in the device handler as it is aware of the device status (in motion but in-erase mode).
Since the XournalppCursor is not aware of the device status, then it is not a good candidate to incorporate the logic.

### DIFF
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -273,7 +273,12 @@ void XournalppCursor::updateCursor() {
                 }
             }
         } else if (type == TOOL_ERASER) {
-            cursor = getEraserCursor();
+            if (this->invisible) {
+                setCursor(CRSR_BLANK_CURSOR);
+                cursor = nullptr;
+            } else {
+                cursor = getEraserCursor();
+            }
         }
 
         else if (type == TOOL_TEXT) {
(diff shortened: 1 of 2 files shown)


rolandlo:
Thanks for the contribution. I think the goal of this PR is fine. The implementation does not quite work on my system (Ubuntu 22.04, Wayland) though. When erasing with `STYLUS_CURSOR_NONE`, the eraser cursor is only shown very briefly when I touch the screen with the stylus. While erasing stuff (and moving the cursor around to do that) I don't see an eraser cursor.
On what system did you test your PR?

rolandlo:
Thanks for the contribution. I think the goal of this PR is fine. The implementation does not quite work on my system (Ubuntu 22.04, Wayland) though. When erasing with `STYLUS_CURSOR_NONE`, the eraser cursor is only shown very briefly when I touch the screen with the stylus. While erasing stuff (and moving the cursor around to do that) I don't see an eraser cursor.
On what system did you test your PR?

a-mehrabian:
> 

Hi rolandlo,
I only tested it, on the Ubuntu 20.04 with the X11.

a-mehrabian:
> Thanks for the contribution. I think the goal of this PR is fine. The implementation does not quite work on my system (Ubuntu 22.04, Wayland) though. When erasing with `STYLUS_CURSOR_NONE`, the eraser cursor is only shown very briefly when I touch the screen with the stylus. While erasing stuff (and moving the cursor around to do that) I don't see an eraser cursor. On what system did you test your PR?

Ubuntu 20, X11

Issue #2362, opened 2020-10-31
Pen, double arrow


rdabhi:
Kindly consider adding a double arrow (pointing both sides) to the list of shapes that can be drawn with pen (& highlighter)
I find myself drawing double arrows all the time, with arrow + line (for drawing the other side of arrow)

PR #3946, opened 2022-04-08
Add double-ended arrow tool

C34A:
I've been using xournal++ for a while and I was annoyed that while there is an arrow tool, there isn't a double ended arrow tool. Since there appears to already be an issue for this (#2362) which has been around for a long time, I thought I'd try to add it. This code seems to work, but I am not very familiar with this codebase or GTK so it seems likely there is something wrong with my implementation.

This reuses the icon for a single-ended arrow since I didn't know how to make/where to find one for a double-ended arrow.


src/core/control/tools/ArrowHandler.cpp
```diff
@@ -30,6 +30,12 @@ void ArrowHandler::drawShape(Point& c, const PositionInputData& pos) {
...
             // remove previous points
+            if (doubleEnded) {
+                stroke->deletePoint(8);
+                stroke->deletePoint(7);
+                stroke->deletePoint(6);
+                stroke->deletePoint(5);
+            }
             stroke->deletePoint(4);
             stroke->deletePoint(3);
             stroke->deletePoint(2);
```
bhennion, line 41: Use `stroke->deletePointsFrom(1);` rather

bhennion:
Thanks for the PR! (and sorry noone could make time to review it before now).

Besides the above comment, the code looks clean, and the feature useful enough. Of course, the question of the icon needs to be addressed. The icons are simple .svg files that you could edit for instance with Inkscape. There needs to be two versions: a light and a dark one.
You can start by copying and editing `ui/iconsColor-dark/hicolor/scalable/actions/xopp-draw-arrow.svg` and simply add a symmetric copy of the arrow head, or something of the sort.

C34A:
Thanks for the feedback! I'll give that a shot.

C34A:
Ok, your comment is resolved and I added the needed icons. They probably aren't perfect, but they appear to match the other icons.

bhennion:
Look's good to me! The feature seems to work as expected. I think it can be merged as is.

Unless an objection is raised (by you @C34A, or by anyone else) in the meantime, I'll be squashing and merging this PR in 48 hours.

bhennion (approved):

bhennion:
Just merged the PR. Thanks a lot for the contribution @C34A!

PR #3952, opened 2022-04-12
Fix image loading regression

### LINKED ISSUE #3951: Images distorted and cropped in version 1.1.1+dev
**Affects versions:**
 - OS: Ubuntu 20.04
 - Desktop environment: GNOME Desktop
 - Version of libgtk: 3.24.20
 - Version of Xournal++: 1.1.1+dev

**Describe the bug**
Images (imported or pasted) get distorted: the image is compressed horizontally and the bottom part is cropped. Sometimes they are also compressed vertically.

**To Reproduce**
1. Save image or copy it in clipboard
2. Paste or import image into document

**Expected behavior**
The image should be imported or copied without being distorted

**Screenshots of Problem**
1.1.1+dev
Schermata da 2022-04-11 19-55-30

1.1.0+dev
Schermata da 2022-04-11 20-10-39

**Additional context**
This happens with files created with previous versions of Xournal++ as well. Rolling back to 1.1.0+dev solves the issue. Files created with 1.1.1+dev still have distorted images, old files return normal if they haven't been edited, otherwise they stay distorted. It affects jpg and npg.

Technius:
Fix #3951.

Oops, accidentally swapped height and width when generating the image buffer.

### DIFF
--- a/src/core/model/Image.cpp
+++ b/src/core/model/Image.cpp
@@ -130,8 +130,8 @@ auto Image::getImage() const -> cairo_surface_t* {
         this->imageSize = {gdk_pixbuf_get_width(pixbuf), gdk_pixbuf_get_height(pixbuf)};
 
         // TODO: pass in window once this code is refactored into ImageView
-        this->image = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, gdk_pixbuf_get_height(pixbuf),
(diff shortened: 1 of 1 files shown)


Technius:
Merging immediately since this bug is very high priority.

LittleHuba:
Removing the assets should suffice. The automatic pipeline should then release the new nightly right at midnight.

PR #3962, opened 2022-04-17
Refactor setsquare

rolandlo:
This PR refactors the Setsquare following the MVC pattern as demanded by @LittleHuba in https://github.com/xournalpp/xournalpp/pull/3082#discussion_r809361933


src/core/control/SetsquareController.h
```diff
@@ -0,0 +1,169 @@
...
+class SetsquareController {
+public:
+    SetsquareController(std::unique_ptr<SetsquareView>& setsquareView, std::unique_ptr<Setsquare>& s);
+    ~SetsquareController() = default;
+
+public:
+    double getHeight() const;
+    double getRotation() const;
+    double getTranslationX() const;
+    double getTranslationY() const;
```
LittleHuba, line 31: These should not be part of the controller.

Instead the model (Setsquare) should be used to get these values. This promotes wrong usage.

LittleHuba (commented):
Cool that you continue on this front!

I unfortunately didn't have much time to look through this, sorry. But I saw one mistake during my quick skim over your work. 

You have the three classes. Controller, model and view. The model is the stateful object that stores the current context and the attributes. So to get the current state you should always query the model. No facade hiding this or anything.

The controller is there to modify the model. It does only contain state that is temporary and methods to modify the model. You can see it as kind of a firewall that only allows users to change the model from one correct state to another correct state.

And then lastly the view that also does not contain any state but visualizes the state contained in the model.

src/core/control/SetsquareController.h
rolandlo: > Cool that you continue on this front!
> 
> I unfortunately didn't have much time to look through this, sorry. But I saw one mistake during my quick skim over your work.
> 
> You have the three classes. Controller, model and view. The model is the stateful object that stores the current context and the attributes. So to get the current state you should always query the model. No facade hiding this or anything.
> 
> The controller is there to modify the model. It does only contain state that is temporary and methods to modify the model. You can see it as kind of a firewall that only allows users to change the model from one correct state to another correct state.
> 
> And then lastly the view that also does not contain any state but visualizes the state contained in the model.

Thanks for the help, which is very useful. I found it hard to find reliable sources on the MVC model. There seem to be lots of variants and explanations I found on the internet often do not take into account implementation details and/or are designed with a different use case in mind.

rolandlo: > These should not be part of the controller.
> 
> Instead the model (Setsquare) should be used to get these values. This promotes wrong usage.


How exactly should the `SetsquareInputHandler` access the model? Currently it has no reference to the model (it can only access the controller via the `InputContext`) and that's why I added those methods. The `SetsquareInputHandler` is created at the start of the application (like all input handlers), whereas the `SetsquareController` (and the `SetsquareView` and `Setsquare` itself) is only created when it is needed.
Should I 

- pass a pointer of the `Setsquare` to the `SetsquareInputHandler` once it is created 
- keep a reference to the setsquare in the `XournalWidget` (so that it can be accessed via the input context like the controller)
- add a `getSetsquare` method to the `SetsquareController`

or would you advise something different?

rolandlo: I'm also unsure about ownership (via `unique_ptr`). Currently:
- The `XournalWidget` owns the `SetsquareController`
- The `SetsquareController` owns the `SetsquareView` and `Setsquare`
- The `InputContext` owns the `SetsquareInputHandler`

Does that look alright?

LittleHuba: I think the ownership is the clue here. What you normally do with an MVC is that you have a view-tree which I mentioned in a previous ticket. So the hierarchically higher view should own all contained views.

The same hierarchy is mirrored by the model. Because every view has a model (not necessarily the other way round) you can mirror the hierarchy mostly.

The controller does not really have a fixed place. It is more of a "place it where you need it".  In special cases it could even be discussed to have multiple concurrent controllers for the same model. Though, they must then take this concurrency into account to modify the model correctly.

LittleHuba: I'd prefer the push variant of the MVC, where you subscribe your view to the model and get updates for free. In this context, your input handler would then be a view and should be subscribed to the model.

LittleHuba: I can draw you a UML diagram the next few days that shows you the dependencies of this scenario if you'd like.

rolandlo: > I can draw you a UML diagram the next few days that shows you the dependencies of this scenario if you'd like.

Yes, please.

LittleHuba: https://cloud.huberulrich.de/s/ZWnikJQEofPJzeW

Here is an in-depth explanation of the MVC pattern in general. If you need it applied to this example drop me a note and I'll apply it for you.

LittleHuba: Also of note: There is never a hard one or the other between the two variants. You can also make a mix, where the model provides the data partially to the view via `update()` and part of the data is then queried from the model by the view itself. For example in a ListView you could only provide the index of the updated item in `update()` and the View would then query the data of this item from the model.

rolandlo: Thanks for the explanations. I will first try to apply it to this example myself and then ask more specific questions later on.

bhennion:
Can this be closed in favor of #4211?

rolandlo:
> Can this be closed in favor of #4211?

Yes.

PR #3964, opened 2022-04-18
MacOS: fix linker flags when using MacPorts

Technius:
The CMake dependencies found with pkgconfig will incorrectly set the external
module LINK_OPTIONS to be of the form `-framework A B C`, which is incorrect.
This commit adds some checks to correct this CMake/pkgconfig error.

Furthermore, some redundant linker flags have been removed from
xournalpp-thumbnailer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12...${CMAKE_VERSION})
+cmake_minimum_required(VERSION 3.12)
 cmake_policy(VERSION ${CMAKE_VERSION})
 
 ## Project related Variables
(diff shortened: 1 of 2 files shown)


Febbe:
First of all, we should replace pkgconfig with find_package. Second, shouldn't this be fixed in a newer cmake version?

CMakeLists.txt
```diff
@@ -108,6 +108,14 @@ if (#[[${CMAKE_VERSION} VERSION_LESS "3.18" AND]] ${CMAKE_SYSTEM_NAME} MATCHES "
     string(REPLACE "-framework;" "-framework " libs "${libs}")
     message(STATUS "PkgConfig::ExternalModules::INTERFACE_LINK_LIBRARIES fixed: ${libs}")
     set_target_properties(PkgConfig::ExternalModules PROPERTIES INTERFACE_LINK_LIBRARIES "${libs}")
+
+
+    if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13")
```
Febbe, line 113: This looks good so far, but it is fixed since [3.18](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/3814), so either exclude newer versions or raise the minimum required for macOSX.

Technius: It is not fixed. I am using CMake 3.22.

Technius:
This passes CI and fixes my issue locally, so merging now.

Technius:
This passes CI and fixes my issue locally, so merging now.

Issue #3542, opened 2021-11-05
Option to move the background vertical line to right


HirbodBehnam:
**Is your feature request related to a problem? Please describe.**
Currently the background option "ruled with vertical line" places a vertical line in the left side of paper. In right to left languages like Persian or Arabic this line must on right.
I also had a look at the #2137 and opened the xml inside the xopp file but it doesn't have any options:
```xml
<background type="solid" color="#000000ff" style="lined"/>
```

**Describe the solution you'd like**
It would be really cool if there was an option to specify if user wants the vertical line on left or right of paper.

**Describe alternatives you've considered**
I currently simply use "Ruled" background without vertical line.

Thanks for the great project! And sorry if this is an duplicate; I searched for similar issued but found nothing about this...

Issue #1713, opened 2020-01-20
Add option to change background lined ruling/spacing


y2kmurali:
**Is your feature request related to a problem? Please describe.**
Some writing applications (i.e. math) require different line spacings than others (i.e. written responses). Xournal currently only has one default line ruling; if I want to change the ruling (e.g. college ruled instead of wide ruled), the only real thing I can do is download a pdf of a college ruled paper and set that as my background, for every document I make.

**Describe the solution you'd like**
I'd like there to be a option under the menu for setting the ruling size manually (i.e. having a menu for ruling width in some unit (font points?)).

**Describe alternatives you've considered**
An alternative might include having a few preset options (wide ruled, college ruled, extra wide ruled, etc.) but that might be extra work for less flexibility for the user.

renatobellotti:
Is there no way to use squared backgrounds instead of lines? I couldn't find the option. For me, this is a deal breaker for not using Xournal++.

rolandlo:
I also think that the ruling size should be made configurable. That's not hard, but requires extending the file format.

@renatobellotti There is the graph background (with or without border). You find it in the menu `journal->page background`

Febbe:
@renatobellotti We already have a preset of default backgrounds, including squared ruling. Those currently have an exact size of 0.5cm(print size)

renatobellotti:
> @renatobellotti We already have a preset of default backgrounds, including squared ruling. Those currently have an exact size of 0.5cm(print size)

Thanks for letting me know!

i-zanis:
Hello, 

Is there a way to reduce the height on the "Ruled" option? 
It enforces letter to become larger.

rolandlo:
@i-zanis It's possible by modifying `pagetemplates.ini`. See https://github.com/xournalpp/xournalpp/issues/2137#issuecomment-799956788

PR #3969, opened 2022-04-19
View refactor 2: Generated backgrounds

bhennion:
This PR follows #3503.

In here, I refactored the way backgrounds were painted, to get a single base class for all background views (pdf, image, and generated ones -- lined, ruled, graph, and so on) **Edit:** PDF backgrounds will be dealt with in a follow up PR #3990.

About the generated backgrounds, the functionalities are not changed, with the following exceptions

1. Fixed a bug in the graph background with border: on the left, the old behaviour, on the right, the new one. Note that the crossings are still aligned with the snap grid.
Screenshot_20220419_143313
2. Added a margin parameter for the lined background, to customize the position of the vertical line. A negative value puts the line on the right hand side. I added a default page template with the line on the right, fixing #3542.
3. Added a parameter controlling the spacing between the horizontal lines in ruled and lined backgrounds, fixing #1713.
4. When asked to draw a background to a cairo context, the BackgroundView's only paint the bare minimum, and not the entire page (i.e. we only draw the lines that pass within the cairo context's clip extents).
This will be needed if and when we implement some kind of tilling-based rendering, for very large pages.
 
The added parameters are controlled by the configuration file `ui/pagetemplates.ini`, using the same variable names as for the graph backgrounds (m1 = margin, r1 = line spacing, in 1/72 th of a inch).
Note that the file format already supported those parameters (although they were obviously not used by the background painters). It gives
```
<background type="solid" color="#ffffffff" style="lined" config="m1=-72,r1=14.17,f1=0x00ff00"/>
```
for a lined background, with green horizontal lines every 5mm (=14.17/72 inch), and a margin on the right hand side.

**More importantly**, the case of the pdf backgrounds requires a discussion. @xournalpp/core I'd love you input on this.

For now, the pdf cache is not used. The way I see things, the cache should disappear, and the PdfBackgroundView should become the cache.
In general, there should be a PageView per page (an actual PageView, not the current weird thing) per document view, and each PageView should point to a BackgroundView, some LayerView's, and so on.
To handle the cache size, I imagine the DocumentView would create and destroy PageView's as needed: at a given time, we'd have in memory a couple of PageView's, corresponding to the pages surrounding the currently displayed page.
If the background is a Pdf file, PdfBackgroundView should store its last rendering and use it when possible (with some leniency on the zoom level, like is done in PdfCache for now).

@LittleHuba @Technius Is this what you had in mind?

**Edit:** Added item 4- above


LittleHuba:
Your general approach makes a lot of sense. This would also clean up the architecture of our graphical user interface. 

The only thing you should keep in mind is the reaction time for fast scrolling through pages.

Still have to look through your code but that is a bigger task with 82 changed files...

bhennion:
> The only thing you should keep in mind is the reaction time for fast scrolling through pages.

Yes, this is of course something I want to focus on. I'm not sure how to best do that yet.
There is also the question of infinite canvas that has been asked for by many. This is, of course, orthogonal to the pdf cache question.

I forgot to mention in my first post (I'll add it there too for posterity): there is another change I made to the generated backgrounds. When asked to draw a background to a cairo context, the BackgroundView's only paint the bare minimum, and not the entire page (i.e. we only draw the lines that pass within the cairo context's clip extents).
This will be needed if and when we implement some kind of tilling-based rendering, for very large pages.
(For now, only the eraser and the Spline tool -- I don't know why -- actually call for partial rerenderings. See the terminal output when using those tools)

> Still have to look through your code but that is a bigger task with 82 changed files...

Quite a few files are only changed to add a const qualifier to a function somewhere, or something of the sort. I could split those changes to another more trivial PR if it helps.
The important changes are in the view directory.

bhennion:
I split the previous content of this PR into 3: #3985, this one and one to come with the PDF stuff. This should make things easier to review.

This PR is now focused on the generated backgrounds. It is based on #3985. 

- [ ] Merge #3985 and rebase.

bhennion:
@LittleHuba @Technius This PR is now ready for review. I split it as much as I could without leaving the code in to weird of an intermediate state.

As mentionned above, it now solely focuses on refactoring generated backgrounds (i.e. everything besides pdf and image backgrounds). The only changes in behaviour are the ones listed in the first post.

src/core/view/background/BaseIsometricBackgroundView.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto [indexMinY, indexMaxY] =
+            getIndexBounds(minY - halfLineWidth - ystep - contentYOffset, maxY + halfLineWidth + ystep - contentYOffset,
+                           ystep, 0.0, contentHeight);
+    if ((indexMinY + indexMinX) % 2) {
+        if (indexMinX) {
+            --indexMinX;
+        } else {
+            --indexMinY;
+        }
+    }
```
Technius, line 58: What is this code snippet trying to do?

Technius (commented):
Thanks for these changes. Although I didn't look too closely at the painter implementations, the overall design looks clean to me.

src/util/include/util/Color.h
```diff
@@ -63,6 +63,8 @@ struct ColorU8 {
         rhs = val;
         return is;
     };
+
+    constexpr auto isLight() const -> bool { return uint32_t(red) + uint32_t(green) + uint32_t(blue) > 0x180U; }
```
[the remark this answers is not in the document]
bhennion: I don't know. I just took the constant previously used in the old `BackgroundPainter`s to determine which color set should be used for the rulings.
Changing it would result in some old documents not being displayed the way they used to (the foreground ruling colors could change).

src/core/view/background/StavesBackgroundView.cpp
```diff
@@ -0,0 +1,54 @@
...
+    // Get the bounds of the mask, in page coordinates
+    double minX;
+    double maxX;
+    double minY;
+    double maxY;
+    cairo_clip_extents(cr, &minX, &minY, &maxX, &maxY);
+
+    const double staffHeight = lineWidth + 4 * LINES_SPACING;
+
+    // The + 4 * lineWidth is added to preserve the previous behaviour.
```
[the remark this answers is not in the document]
bhennion:
```
const double vOffsetBetweenStaves = STAVES_SPACING + staffHeight + 4 * lineWidth;
```
The point is: this formula makes little to no sense to me: the `+ 4 * lineWidth` was most likely a mistake in the original implementation. We cannot fix this, however, as it would alter how the background is painted in old documents (moving some straves up). Hence my comment. I'll extend the comment to make this more explicit.

src/core/view/background/BaseIsometricBackgroundView.cpp
bhennion: I added a comment. Hopefully it's clear enough

bhennion:
Thanks for the review! 

> Although I didn't look too closely at the painter implementations, the overall design looks clean to me.

Most of the insides of each painter has simply been copied and pasted from the old painter, not sure they require so much attention.

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -37,18 +38,20 @@ void PageTypeMenu::loadDefaultPage() {
...
 
     auto page = std::make_shared<XojPage>(previewWidth / zoom, previewHeight / zoom);
+    page->setBackgroundType(pt);
 
     cairo_surface_t* surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, previewWidth, previewHeight);
     cairo_t* cr = cairo_create(surface);
     cairo_scale(cr, zoom, zoom);
 
-    bgPainter->paint(pt, cr, std::move(page));
+    auto bgView = xoj::view::BackgroundView::createForPage(std::move(page), pt.config, 2.0);
```
Technius, line 53: This seems like a hack: we're creating a new page so that we can draw with the desired width, height, background color, and background type. Could this be refactored?

Technius (approved):
LGTM except for one issue below.

src/core/control/pagetype/PageTypeMenu.cpp
bhennion: Done!

bhennion:
There is one issue I just noticed (not introduced by this PR):

The previews in the page template configuration menu (journal->Configure page template->"drop down menu on the line starting with Background") do not have the selected background color. Their background color is always white.

This should be fixed someday, but as I could not find a simple fix, this will be postponed to another PR (I'll create an issue).

Besides that, this PR seems to be approved for merging, so I'll merge it in 48 hours unless someone objects.

PR #3970, opened 2022-04-19
Remove x and y from GtkXournal struct

rolandlo:
Those variables `xournal->x` and `xournal->y` have been `0` all the time (for many years) and made the code look more confusing. They are pointless, so I suggest to remove them via this PR.


bhennion:
LGTM!

rolandlo:
Merging in 24 hours, if no objections are raised.

Issue #3921, opened 2022-03-20
Show the filepath in the titlebar along with the filename


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
If you have multiple windows each editing files of the same name, but in different directories, it is difficult to differentiate between them.

**Describe the solution you'd like**
I would like for the filepath of the file currently being edited to be shown in the titlebar along with the name of the file.

PR #3934, opened 2022-03-29
3921 show the filepath in the titlebar along with the filename

p3tkovicn:
Made required enhancement per  **Show the filepath in the titlebar along with the filename #3921** issue.
To avoid breaking default behavior added "Show filepath in titlebar" in **view** tab in settings. 
Default behavior remains same i.e. only filename is shown.
Edited behavior when **Show filepath in titlebar** setting changes and settings are saved, change in titlebar is applied immediately.

Translations to other languages were not added.

PR #3976, opened 2022-04-22
#3921 Show filepath in titlebar along with the filename

lisalcao:
Addressing issue [#3921](https://github.com/xournalpp/xournalpp/issues/3921), changed code to show the full file path in the window's title bar so that if the user has multiple windows, each editing files of the same name in different directories, it is easier for the user to differentiate between them.


Technius (changes requested):
This looks like a revert of #3934 -- are you sure the PR contents are correct?

PR #3988, opened 2022-04-29
Fixes the ordering of recent files

### LINKED ISSUE #3987: RecentHandler doesn't sort elements anymore. Bug in GListView
While refactoring `GList`s to use a c++20 like `GListView` the sorting have been removed. Gtk's RecentManager does not sort the values by modification date.

Also, the equality operator of the GListView::GListViewIterator can't compile if used.

Febbe:
Also fixes the equality operator of GListView
Fixes #3987

### DIFF
--- a/src/core/control/RecentManager.cpp
+++ b/src/core/control/RecentManager.cpp
@@ -1,26 +1,44 @@
 #include "RecentManager.h"
 
+#include <algorithm>
 #include <array>
+#include <cassert>
+#include <functional>
+#include <new>
+#include <vector>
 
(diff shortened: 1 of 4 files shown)


MichelReims:
Is it possible to clear recent files list ?

Febbe:
> Is it possible to clear recent files list ?

We only retrieve the data from the GTK manager. So probably it is possible from the file chooser. An alternative is, you download a Software which does this, like http://ssokolow.com/gtk-recent-scrubber/

PR #3999, opened 2022-05-01
Added a workaround for a reference issue.

### LINKED ISSUE #3997: SegFault when quitting
**Affects versions :**
 - latest master (commit f757458)
 - OS: Archlinux, KDE X11, up to date.

**Describe the bug**
The app crashes (SegFault) when quitting, if the document has just been saved (or has never been modified).
The crash does not happen if the document has been modified since the last time it was saved (and a save or discard dialog is then shown).

**To Reproduce**
1. Open xournal++ from a terminal
2. Close it (e.g. hit `Alt` + `F4`, or click in the upper right cross)
3. It crashes.

The backtrace (see below) seems to point towards the SetsquareInputHandler. This makes little sense to me. @rolandlo: any clue?
```
Error: signal 11
[bt]: (0) ./install/bin/xournalpp(+0xbad7a) [0x55d4f0964d7a]
[bt]: (1) /usr/lib/libc.so.6(+0x42560) [0x7fe4af4ea560]
[bt]: (2) ./install/bin/xournalpp(_ZN12InputContext7getViewEv+0) [0x55d4f09ae150]
[bt]: (3) ./install/bin/xournalpp(_ZN21SetsquareInputHandler6handleERK10InputEvent+0x1c) [0x55d4f09b279c]
[bt]: (4) ./install/bin/xournalpp(_ZN12InputContext6handleEP9_GdkEvent+0xa7) [0x55d4f09ae407]
[bt]: (5) /usr/lib/libgtk-3.so.0(+0x87bb9) [0x7fe4b1056bb9]
[bt]: (6) /usr/lib/libgobject-2.0.so.0(g_signal_emit_valist+0x12c6) [0x7fe4afce56b6]
[bt]: (7) /usr/lib/libgobject-2.0.so.0(g_signal_emit+0x94) [0x7fe4afce5824]
[bt]: (8) /usr/lib/libgtk-3.so.0(+0x345f2e) [0x7fe4b1314f2e]
[bt]: (9) /usr/lib/libgtk-3.so.0(gtk_widget_send_focus_change+0xb7) [0x7fe4b1310d27]
[bt]: (10) /usr/lib/libgtk-3.so.0(+0x355837) [0x7fe4b1324837]
[bt]: (11) /usr/lib/libgtk-3.so.0(+0x355e8a) [0x7fe4b1324e8a]
[bt]: (12) /usr/lib/libgobject-2.0.so.0(g_cclosure_marshal_VOID__OBJECTv+0x7f) [0x7fe4afcc696f]
[bt]: (13) /usr/lib/libgobject-2.0.so.0(g_signal_emit_valist+0x12c6) [0x7fe4afce56b6]
[bt]: (14) /usr/lib/libgobject-2.0.so.0(g_signal_emit+0x94) [0x7fe4afce5824]
[bt]: (15) /usr/lib/libgtk-3.so.0(+0x34c29d) [0x7fe4b131b29d]
[bt]: (16) /usr/lib/libgobject-2.0.so.0(g_object_run_dispose+0x52) [0x7fe4afcd5bb2]
[bt]: (17) ./install/bin/xournalpp(_ZN8GladeGuiD1Ev+0x39) [0x55d4f09a0c59]
[bt]: (18) ./install/bin/xournalpp(_ZN10MainWindowD0Ev+0x9) [0x55d4f09a2719]
[bt]: (19) ./install/bin/xournalpp(_ZN11XournalMain3runEiPPc+0x5bf) [0x55d4f096799f]
[bt]: (20) /usr/lib/libc.so.6(+0x2d310) [0x7fe4af4d5310]
[bt]: (21) /usr/lib/libc.so.6(__libc_start_main+0x81) [0x7fe4af4d53c1]
[bt]: (22) ./install/bin/xournalpp(_start+0x25) [0x55d4f0964405]
```

Febbe:
- since the xournal widget is in disposing state,
 the event handlers can still be called

fixes #3997

### DIFF
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -44,8 +44,7 @@ XournalView::XournalView(GtkWidget* parent, Control* control, ScrollHandling* sc
 
     InputContext* inputContext = new InputContext(this, scrollHandling);
     this->widget = gtk_xournal_new(this, inputContext);
-    // we need to refer widget here, because we unref it somewhere twice!?
-    g_object_ref(this->widget);
+    g_object_ref_sink(this->widget);  // take ownership without increasing the ref count
 
     gtk_container_add(GTK_CONTAINER(parent), this->widget);
     gtk_widget_show(this->widget);
--- a/src/core/gui/inputdevices/InputContext.cpp
+++ b/src/core/gui/inputdevices/InputContext.cpp
@@ -26,6 +26,9 @@ InputContext::InputContext(XournalView* view, ScrollHandling* scrollHandling) {
 }
 
 InputContext::~InputContext() {
+    // Destructor is called in xournal_widget_dispose, so it can still accept events
+    g_signal_handler_disconnect(this->widget, signal_id);
+
     delete this->stylusHandler;
     this->stylusHandler = nullptr;
 
@@ -43,6 +46,7 @@ InputContext::~InputContext() {
 }
 
 void InputContext::connect(GtkWidget* pWidget) {
+    assert(!this->widget);
     this->widget = pWidget;
     gtk_widget_set_support_multidevice(widget, true);
 
@@ -60,7 +64,7 @@ void InputContext::connect(GtkWidget* pWidget) {
 
     gtk_widget_add_events(pWidget, mask);
 
-    g_signal_connect(pWidget, "event", G_CALLBACK(eventCallback), this);
+    signal_id = g_signal_connect(pWidget, "event", G_CALLBACK(eventCallback), this);
 }
 
 auto InputContext::eventCallback(GtkWidget* widget, GdkEvent* event, InputContext* self) -> bool {
--- a/src/core/gui/inputdevices/InputContext.h
+++ b/src/core/gui/inputdevices/InputContext.h
@@ -40,6 +40,7 @@ class SetsquareInputHandler;
 class InputContext {
 
 private:
+    gulong signal_id{0};
     std::unique_ptr<SetsquareInputHandler> setsquareHandler;
     StylusInputHandler* stylusHandler;
     MouseInputHandler* mouseHandler;


src/core/gui/XournalView.cpp
```diff
@@ -45,7 +45,7 @@ XournalView::XournalView(GtkWidget* parent, Control* control, ScrollHandling* sc
     InputContext* inputContext = new InputContext(this, scrollHandling);
     this->widget = gtk_xournal_new(this, inputContext);
     // we need to refer widget here, because we unref it somewhere twice!?
-    g_object_ref(this->widget);
+    // g_object_ref(this->widget);
```
bhennion, line 48: I'd say we need to call `g_object_ref_sink` here, and a `g_object_unref` in `~XournalView`.

Did we ever consider using gtkmm (and cairomm, and popplermm, and so on) to stop worrying about those ref count headaches?

Febbe: > I'd say we need to call g_object_ref_sink here, and a g_object_unref in ~XournalView.

Not sure if references from g_object_new are floating. But I'll read the documentation again.
To use g_object_unref would be the correct way, but since something is still holding a reference the deletion is delayed. Control is already deleted then and we have the same mess. So to fix that we have to dig way deeper into the cleanup procedure, or live with that for know.

About gtkmm we thought about that and had a huge debate @lehmanju made an attempt to rewrite the GUI with gtkmm4.
So in principle there is only the fact against it, that it's a lot of work.

danemtsov: GTK widget constructors return floating references.

How about just banning g_object_ref/unref/... in the wild and use a unique_ptr-like construct for owning references? (btw this is not the only place where ref management is a bit odd)

bhennion: Well, wrapping every ref/unref in RAII structures would still be quite a bit of work (either by using glibmm's Glib::RefPtr, or by implementing our own).

Febbe: Right, just read it again, I will change that. 

To get rid of the `gtk_widget_destroy` function, the widget also needs to be removed from the parent container.

@danemtsov I already created a similar construct:
https://github.com/xournalpp/xournalpp/blob/fb6254599268dbb71ebfa51d415e64a3f328777c/src/util/include/util/PathUtil.h#L24

But for gtkwidgets we must rewrite an own smart ptr. Something like this: https://www.boost.org/doc/libs/1_78_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ptr

Febbe: `Glib::RefPtr` is only a alias to `std::shared_ptr` (gtkmm4). So this wont even work we would have to replace everything with the corresponding gtkmm type (also we have to rewrite the xournal widget or wrap it).

Febbe:
@bhennion if you are interested in porting this application to gtkmm, I would write a clang-tooling tool, which does this automatically. When you start this by hand, it will be a huge effort. Also, you have to rebase every change to master into the porting branch. Having a program doing this for you is much better.

PR #4000, opened 2022-05-02
Fix TextEditor render delay regression

### LINKED ISSUE #3977: Paint lags of text tool under large zoom maginification.
This is not so serious.
**Affects versions :**
 - OS: Linux
 - Desktop environment: X11 with xmonad
 - libgtk: 3.24.33
 - Version of Xournal++: 1.1.1+dev (176e95ac05e30fbf5a5971af6ce2750cf90fe50f or after)
 - Installation method: built from source

**Describe the bug**
When the zoom magnification is set to maximum and a wide letter like W is typed in text tool, the right half of the letter may not be displayed for about 0.5 seconds. If you type letters in succession with key repeats, only the last letter get this delay.

**Screenshots of Problem**

https://user-images.githubusercontent.com/6355486/164603514-8c0ada12-3604-443d-945b-b809f5ef19b2.mp4

**Additional context**
This was introduced by the commit 176e95ac05e30fbf5a5971af6ce2750cf90fe50f.

Technius:
Fixes #3977.

@tattsan can you test to see if this fixes the issue for you?

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -908,7 +908,7 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
 
 void TextEditor::repaintEditor() {
     auto rect = text->boundingRect();
-    this->gui->repaintRect(rect.x, rect.y, rect.width, rect.height);
+    this->gui->rerenderRect(rect.x, rect.y, rect.width, rect.height);
 }
 
 /**


Technius:
This needs to be merged/cherry picked back into master afterwards.

bhennion:
I can confirm this fixes the issue on my machine.

There is another issue, tangentially related: the bounding box of italic text is miscalculated (both on the release branch and on master):

Screenshot_20220502_135220

(The letter goes beyond the bounding box on the right). You may fix this in this PR if you want, or I'll open a dedicated ticket.

tattsan:
I've also confirmed that the issue has been resolved. (Sorry, I've checked master branch only.)
Thank you so much!

tattsan:
@bhennion  This is the correct cursor position. I wonder if the character display bounding box needs to be obtained in another way.
image

Technius:
> I wonder if the character display bounding box needs to be obtained in another way.

The bounding box is obtained using `pango_layout_get_size` which just calls [`pango_layout_get_extents`](https://docs.gtk.org/Pango/method.Layout.get_extents.html). The latter is described as:
> Logical extents are usually what you want for positioning things.

So for pixel-perfect bounding box calculation, we'll want to use something else to determine the size. That seems like enough work that it should go in a separate issue / PR.

Technius:
Merging for now. Thanks for testing.

PR #4004, opened 2022-05-03
Fetch PopplerPage from PdfCache instead of recreating it

bhennion:
This PR improves the behaviour of PdfCache (and callers).

Before this PR, whenever a pdf background was to be rendered:

- a PopplerPage was extracted from the PDF document (using the page number, through a call to `poppler_document_get_page(document, number)`).
- the PopplerPage was passed on to PdfCache, which tested if the cache contained another page with the same number

The PR removes the unnecessary creation of a new PopplerPage (meaning the extraction from the PDF document), and simply searches the cache for a PopplerPage with the same number.

I am not sure how expensive a call to `poppler_document_get_page(document, number)` is, but this should slightly improve the performance of the process.

I tested the following behaviours:

- Open a file without a pdf background
- Open a file with a pdf background
- Change a single page's background (to ruled background, or to another page of the pdf)
- Change file (in both directions, and from pdf to pdf)
- Export a file with pdf background, in all possible formats (pdf, png, svg)
- Zoom in and out (and check the cache is used, through some debug `cout`'s)
- Open a large pdf file (500 pages), scroll and check it's fluid
- Check the previews

Tell me if I forgot something obvious...

Otherwise, this PR is ready for review :-)

### DIFF
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -5,6 +5,10 @@
 #include <cstdio>
 #include <utility>
 
+#include "control/settings/Settings.h"
+#include "pdf/base/XojPdfDocument.h"
+#include "util/i18n.h"
+
 class PdfCacheEntry {
 public:
     /**
@@ -35,27 +39,38 @@ public:
     cairo_surface_t* rendered;
 };
 
-PdfCache::PdfCache(size_t size) { this->size = size; }
+PdfCache::PdfCache(const XojPdfDocument& doc, Settings* settings): pdfDocument(doc) { updateSettings(settings); }
 
 PdfCache::~PdfCache() {
     clearCache();
     this->size = 0;
 }
 
-void PdfCache::setZoom(double zoom) { this->zoom = zoom; }
-
 void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
 
-void PdfCache::setAnyZoomChangeCausesRecache(bool b) { this->zoomClearsCache = b; }
+void PdfCache::setMaxSize(size_t newSize) {
+    this->size = newSize;
+    while (this->data.size() > this->size) {
+        delete this->data.back();
+        this->data.pop_back();
+    }
+}
+
+void PdfCache::updateSettings(Settings* settings) {
+    if (settings) {
+        setMaxSize(settings->getPdfPageCacheSize());
+        setRefreshThreshold(settings->getPDFPageRerenderThreshold());
+    }
+}
 
 void PdfCache::clearCache() {
     for (PdfCacheEntry* e: this->data) { delete e; }
     this->data.clear();
 }
 
-auto PdfCache::lookup(const XojPdfPageSPtr& popplerPage) -> PdfCacheEntry* {
+auto PdfCache::lookup(size_t pdfPageNo) const -> PdfCacheEntry* {
     for (PdfCacheEntry* e: this->data) {
-        if (e->popplerPage->getPageId() == popplerPage->getPageId()) {
+        if (static_cast<size_t>(e->popplerPage->getPageId()) == pdfPageNo) {
             return e;
         }
     }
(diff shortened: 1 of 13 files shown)


src/core/control/PdfCache.cpp
```diff
@@ -75,15 +78,15 @@ PdfCacheEntry* PdfCache::cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img,
...
-void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoom) {
-    this->renderMutex.lock();
+void PdfCache::render(cairo_t* cr, size_t pageNo, double zoom, double pageWidth, double pageHeight) {
+    std::lock_guard<std::mutex> lock(this->renderMutex);
 
     this->setZoom(zoom);
 
-    PdfCacheEntry* cacheResult = lookup(popplerPage);
+    PdfCacheEntry* cacheResult = lookup(pageNo);
     bool needsRefresh{cacheResult == nullptr};
```
Technius, line 87: Random unrelated nitpick, but the brace initialization syntax here makes this unnecessarily hard to read. Could you fix this to use `=` ?

src/core/control/PdfCache.cpp
```diff
@@ -98,30 +101,40 @@ void PdfCache::render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoo
     if (needsRefresh) {
         double renderZoom = std::max(zoom, 1.0);
 
+        auto popplerPage = cacheResult ? cacheResult->popplerPage : pdfDocument.getPage(pageNo);
+
+        if (!popplerPage) {
+            g_error("PdfCache::render Could not get the pdf page %zu from the document", pageNo);
```
Technius, line 107: nitpick: `g_warning` would probably be better here, since this is not a fatal error.

src/core/control/PdfCache.h
```diff
@@ -53,12 +54,16 @@ class PdfCache {
...
+    static void renderMissingPdfPage(cairo_t* cr, double pageWidth, double pageHeight);
+
 private:
     void setZoom(double zoom);
-    PdfCacheEntry* lookup(const XojPdfPageSPtr& popplerPage);
+    PdfCacheEntry* lookup(size_t pageNo) const;
     PdfCacheEntry* cache(XojPdfPageSPtr popplerPage, cairo_surface_t* img, double zoom);
 
 private:
+    XojPdfDocument pdfDocument;
```
Technius, line 65: Is this meant to be owned by PdfCache or is this supposed to be a reference?

src/core/control/PdfCache.h
```diff
@@ -19,22 +19,23 @@
...
+    PdfCache(const XojPdfDocument& doc, size_t size);
     virtual ~PdfCache();
 
 private:
     PdfCache(const PdfCache& cache);
     void operator=(const PdfCache& cache);
 
 public:
-    void render(cairo_t* cr, const XojPdfPageSPtr& popplerPage, double zoom);
+    void render(cairo_t* cr, size_t pageNo, double zoom, double pageWidth, double pageHeight);
```
Technius, line 38: Could you add a comment explaining the meaning of these parameters?

src/core/control/PdfCache.h
```diff
@@ -53,12 +54,16 @@ class PdfCache {
      */
     void setRefreshThreshold(double percentDifference);
 
+    static void renderMissingPdfPage(cairo_t* cr, double pageWidth, double pageHeight);
+
 private:
     void setZoom(double zoom);
-    PdfCacheEntry* lookup(const XojPdfPageSPtr& popplerPage);
+    PdfCacheEntry* lookup(size_t pageNo) const;
```
Technius, line 61: I assume `pageNo` refers to the page of the PDF, not the Xournal++ document?

src/core/control/jobs/ImageExport.cpp
```diff
@@ -147,14 +146,17 @@ void ImageExport::exportImagePage(size_t pageId, size_t id, double zoomRatio, Ex
         return;
     }
 
-    if (page->getBackgroundType().isPdfPage() && (exportBackground >= EXPORT_BACKGROUND_UNRULED)) {
+    if (page->getBackgroundType().isPdfPage() && (exportBackground != EXPORT_BACKGROUND_NONE)) {
```
Technius, line 149: Does this change the behavior? Looks like a bug fix.

src/core/control/jobs/ImageExport.cpp
```diff
@@ -147,14 +146,17 @@ void ImageExport::exportImagePage(size_t pageId, size_t id, double zoomRatio, Ex
...
         auto pgNo = page->getPdfPageNr();
         XojPdfPageSPtr popplerPage = doc->getPdfPage(pgNo);
-
-        PdfView::drawPage(nullptr, popplerPage, cr, zoomRatio, page->getWidth(), page->getHeight());
+        if (!popplerPage) {
+            this->lastError = _("Error while export pdf background: I cannot find the pdf page.");
```
Technius, line 153: It would be helpful to include the page number in the error message as well.

src/core/gui/XournalView.cpp
```diff
@@ -717,10 +722,13 @@ void XournalView::documentChanged(DocumentChangeType type) {
     for (auto&& page: viewPages) { delete page; }
     viewPages.clear();
 
-    this->cache->clearCache();
+    this->cache.reset();
 
     Document* doc = control->getDocument();
     doc->lock();
+    if (doc->getPdfPageCount() != 0) {
+        cache = std::make_unique<PdfCache>(doc->getPdfDocument(), control->getSettings()->getPdfPageCacheSize());
```
Technius, line 730: Do we also need to update the PdfCache's size when the settings change?

Technius (commented):
Thanks for these changes. I looked through the code, will give a more comprehensive test tomorrow.

src/core/control/PdfCache.h
bhennion: XojPdfDocument is itself some kind of smart pointer (It's pretty unreadable and would deserve a do-over, but that's out of scope). So I do want to own one of them, to make sure the pdf document does not get destroyed before the cache does. Both are eventually destroyed when the document is replaced by a newly opened one.

src/core/control/jobs/ImageExport.cpp
bhennion: EXPORT_BACKGROUND_UNRULED = 1 and EXPORT_BACKGROUND_NONE = 0, so there is no change in behaviour. I just find this easier to parse.

src/core/gui/XournalView.cpp
bhennion: It'd be better indeed. As it is, the cache size is only applied to newly opened documents.
This is a papercut, but there is another issue with the settings: see other comment (above or below, I don't know :)

src/core/gui/XournalView.cpp
```diff
@@ -474,8 +477,10 @@ void XournalView::zoomChanged() {
     XojPageView* view = getViewFor(currentPage);
 
     ZoomControl* zoom = control->getZoomControl();
-    this->cache->setAnyZoomChangeCausesRecache(false);  // We're zooming -- no need for repeated re-renders.
-    this->cache->setRefreshThreshold(control->getSettings()->getPDFPageRerenderThreshold());
+    if (this->cache) {
+        this->cache->setAnyZoomChangeCausesRecache(false);  // We're zooming -- no need for repeated re-renders.
+        this->cache->setRefreshThreshold(control->getSettings()->getPDFPageRerenderThreshold());
+    }
```
bhennion, line 483: This should only happen when the settings are changed, see other comment.
This has a cost and the refresh threshold should just be updated upon settings change.
Is there a SettingsChanged event dispatch somewhere in the code?

(also the setAnyZoomChangeCausesRecache should be removed altogether, and the corresponding variable in PdfCache should be replaced by false all over)

bhennion (commented):
Thanks for the feedback @Technius!
I'll take care of the nitpicks. The SettingsChanged question is more important, but I'm not sure how to solve it.

src/core/gui/XournalView.cpp
bhennion: Ok. I found a way to only update those parameters when the settings are actually changed. Possibly not the best way, you tell me!

src/core/control/PdfCache.cpp
```diff
@@ -35,27 +38,31 @@ class PdfCacheEntry {
...
+}
 
 void PdfCache::clearCache() {
     for (PdfCacheEntry* e: this->data) { delete e; }
     this->data.clear();
 }
 
-auto PdfCache::lookup(const XojPdfPageSPtr& popplerPage) -> PdfCacheEntry* {
+auto PdfCache::lookup(size_t pdfPageNo) const -> PdfCacheEntry* {
     for (PdfCacheEntry* e: this->data) {
```
Technius, line 64: For a follow up PR: This looks rather inefficient given that this is the lookup method on a _cache_, should probably change `data` to a map of some sort.

src/core/gui/XournalView.cpp
```diff
@@ -717,10 +726,14 @@ void XournalView::documentChanged(DocumentChangeType type) {
...
 
-    this->cache->clearCache();
+    this->cache.reset();
 
     Document* doc = control->getDocument();
     doc->lock();
+    if (doc->getPdfPageCount() != 0) {
+        this->cache = std::make_unique<PdfCache>(doc->getPdfDocument(), control->getSettings()->getPdfPageCacheSize());
+        this->cache->setRefreshThreshold(control->getSettings()->getPDFPageRerenderThreshold());
+    }
```
Technius, line 736: Please refactor this (and the above) into a (private) helper function in case we add another setting to PdfCache in the future.

src/core/gui/dialog/SettingsDialog.cpp
```diff
@@ -914,4 +915,5 @@ void SettingsDialog::save() {
 
     this->control->getWindow()->setGtkTouchscreenScrollingForDeviceMapping();
     this->control->initButtonTool();
-}
\ No newline at end of file
+    this->control->getWindow()->getXournal()->onSettingsChanged();
```
Technius, line 918: Not sure whether this is the best solution, but I think this is fine for now. It may be possible that this will lead to deadlocks in the future, but we can revisit this decision if that happens.

Technius (approved):

src/core/control/PdfCache.cpp
bhennion: Well, it actually turns out the cache size is (always) 10. The setting parameter we fetch by calling `control->getSettings()->getPdfPageCacheSize()` has only a default value, and no way of changing it (besides maybe editing the config file by hand).

With those kind of sizes, a map would probably actually be slower

bhennion:
Thanks a lot for the review @Technius! I made the required change by creating a function `PdfCache::updateSettings(Settings* settings)`. Any additional settings should just be handled in this funciton.

Provided the CI test passes and noone objects, I'll be squashing and merging this in 48 hours.

PR #4010, opened 2022-05-05
Copy paste selection anchors

mjg:
This series focusses on copy/paste usability in a "pen-only" or "pen-mostly" workflow, such as with a pen tablet display without keyboard (in reach). We already have "delete selection" as an anchor (handle) on a selection box, and this series implements cut/copy/paste/duplicate to go along with that. It works best with quick access to selections, such as by tapping.

In terms of placement and styling, I followed the existing delete anchor. In particular, I use the same padding, which is different from the padding of the rotate anchor (for reasons unknown to me). The drawn symbols are meant to resemble "minus" (cut), "plus" (copy), "left arrow" (paste), "plus with arrow" (duplicate, i.e. copy+paste).


mjg:
And here is what it looks like:

CopyPaste

Technius:
Very nice feature; this would be really convenient to use. Will take a look at the code over the week.

Some comments about the icons as shown here:
* I think the icons should be replaced with more standard looking icons, since the user would probably not understand what they do at first glance. How about something like scissors (cut), solid box + dashed box (copy), two overlapping solid boxes (paste), and two overlapping solid boxes with an arrow (duplicate)?
* We don't need to restrict ourselves to the style of the current delete icon. In the past, some users have complained about the delete box being too small / hard to press, so we should strive for better UI/UX even if it means changing the style.

mjg:
Yes, first thing I checked after using this in class was the size of the boxes ... I might be difficult to find a good balance between usability, observability and non-obtrusiveness - I think the existing style of the delete and rotate box (i.e. more line art than icon) is quite good in that respect already, and maybe we can do you what you suggest in that style. In any case, the size is dpi dependent but we should increase the current upper size limit, I guess.

src/core/control/tools/EditSelection.cpp
```diff
@@ -994,9 +1034,16 @@ void EditSelection::paint(cairo_t* cr, double zoom) {
     drawAnchorRect(cr, x + width, y + height, zoom);
 
 
-    drawDeleteRect(cr, std::min(x, x + width) - (DELETE_PADDING + this->btnWidth) / zoom, y, zoom);
+    drawAnchorDelete(cr, std::min(x, x + width) - (DELETE_PADDING + this->btnWidth) / zoom, y, zoom);
```
Technius, line 1037: Lots of duplication here. I think the code would be easier to read if some of the common computations were extracted out, e.g.
```
const double xmin = std::min(x, x + width);
const double xmax = std::max(x, x + width);
const double scaledBtnOffs = (DELETE_PADDING + this->btnWidth) / zoom;
```

src/core/control/tools/EditSelection.cpp
```diff
@@ -878,6 +891,33 @@ auto EditSelection::getSelectionTypeForPos(double x, double y, double zoom) -> C
         return CURSOR_SELECTION_DELETE;
     }
 
+    if (xmin - (DELETE_PADDING + this->btnWidth) - BORDER_PADDING <= x &&
```
Technius, line 894: I think the code would be easier to read if some of the common computations were extracted out, e.g.
```c++
const double leftBtnMin = xmin - (DELETE_PADDING + this->btnWidth) - BORDER_PADDING;
const double leftBtnMax = xmin - (DELETE_PADDING + this->btnWidth) + BORDER_PADDING;
// etc.
```

src/core/control/tools/EditSelection.cpp
```diff
@@ -1038,6 +1084,74 @@ void EditSelection::drawDeleteRect(cairo_t* cr, double x, double y, double zoom)
...
+/**
+ * draws an indicator where you can cut the selection
+ */
+void EditSelection::drawAnchorCut(cairo_t* cr, double x, double y, double zoom) const {
+    cairo_set_source_rgb(cr, 0, 0, 0);
+    cairo_rectangle(cr, x * zoom - (this->btnWidth / 2), y * zoom - (this->btnWidth / 2), this->btnWidth,
+                    this->btnWidth);
+    cairo_stroke(cr);
+    cairo_set_source_rgb(cr, 1, 0, 0);
+    cairo_move_to(cr, x * zoom - (this->btnWidth / 2), y * zoom - (this->btnWidth / 2));
```
Technius, line 1096: Since we'll be making the style consistent, the "setup" logic could be extracted into a helper function:

```c++
namespace {
void prepareDrawAnchor(cairo_t* cr, double x, double y, double zoom, double btnWidth) {
    cairo_set_source_rgb(cr, 0, 0, 0);
    cairo_rectangle(cr, x * zoom - (btnWidth / 2), y * zoom - (btnWidth / 2), btnWidth,
                    btnWidth);
    cairo_stroke(cr);
    cairo_set_source_rgb(cr, 1, 0, 0);
    cairo_move_to(cr, x * zoom - (btnWidth / 2), y * zoom - (btnWidth / 2));
}
} // namespace
```

Technius (commented):
The code changes look very reasonable to me. Some refactoring nitpicks below.

Technius:
Random thought, but is paste actually useful? I'm having a hard time imagining a situation where I would want to paste (a different element) by selecting an existing element.

src/core/control/tools/EditSelection.cpp
mjg: If you want to refactor, I'd suggest to refactor the existing code first, then rebase/recode anything new on top. That way we know which (possibly unexpected) changes in behaviour come from which code changes.

mjg:
> Random thought, but is paste actually useful? I'm having a hard time imagining a situation where I would want to paste (a different element) by selecting an existing element.

I use it all the time. It's super useful to:

- paste the same selection multiple times: I often use the same "base" multiple times but scribble in between. Without quick "paste" acces, I would have to do the selection again, which can be time consuming or complicated if it's a lasso selction
- transport far away: yes, you can drag+scroll by hitting the edge of the screen, but this is difficult to controll in terms of scroll speed. It's much easier to copy, then scroll by any means, then select *anything* (e.g. by tapping) and insert what you have copied before.

Besides, it's a matter of consistency to have "paste" where you have "copy". In fact: Without "paste", there is no point in having "copy" nor "cut" at all. That would reduce us to "duplicate", limitting the usefulness quite a bit.

In summary: Just use it with a pen focussed workflow (keyboard out of reach) and you'll understand ;)

danemtsov:
Some brainstorming about obtrusiveness:

In OneNote, you can tap the selection to bring up a mini context menu:
&emsp;Screenshot 2022-05-16 121618
Menu expanded:
&emsp;Screenshot 2022-05-16 121645

Also right now, duplication can be done with a right-click on the selection or the equivalent action with a pen, though I agree that discoverability of this feature is 'by chance' at best – some sort of context menu item or anchor would certainly be better.

Note that the context menu described above would show on left-click or normal tap, the latter of which might make input handling a bit tricky (conflict with scroll handler?). But left-click could be enough for a start.

Alternative ideas:
* Replace _right-click → duplicate_ with a straightforward _right-click → context menu_ (but I'm not sure every pen has a convenient way to 'right-click'; there would still be some hassle to trigger it with touch only)
* Always show mini menu on selection, but obtrusiveness... (maybe there could be a setting for this?)

What do you think?

mjg:
> Some brainstorming about obtrusiveness:
 
With small handles, I don't find the current version to be obtrusive. This might change with the suggested larger icons, of course.

>
> In OneNote, you can tap the selection to bring up a mini context menu:  Screenshot 2022-05-16 121618 Menu expanded:  Screenshot 2022-05-16 121645

This is similar to the suggestion in this PR, with the menu items arranged around the selection.

> 
> Right now in Xournal++, duplication can be done with a right-click on the selection or the equivalent action with a pen, though I agree that discoverability of this feature is 'by chance' at best – some sort of context menu item or anchor would certainly be better.

This does not work over here, no matter how I configure the right button - you assign the button functions within xournalpp.

> 
> Note that the context menu described above would show on left-click or normal tap, the latter of which might make input handling a bit tricky (conflict with scroll handler?). But left-click could be enough for a start.

If I understand this correctly, you would need to make the selection, then tap to bring up the menu, then tap to execute the action? That's one tap more. We often have a conflict between convenience and obtrusiveness, of course.

I had thought about putting the cut/copy/... actions in the floating toolbox. But (IIRC) the selection is deselected once you tap to bring up the toolbox, so this would work only if you bring up the toolbox in a different way.

src/core/control/tools/EditSelection.cpp
LittleHuba: Feel free to refactor within a separate commit that is then part of this PR.

bhennion (commented):
I agree with what @Technius said:

- Those are great additions, and improving the keyboard-free UX is absolutely necessary.
- I find the icons not very explicit. They would benefit some more "common" forms. This includes the handles already present on master. Somehow, the cursor changing is what really helps understanding the rotation handle for instance.
- I also wonder about the `Paste` handle. It seems that the paste button in a toolbar/floating toolbox is more natural "goto" place for the feature. IMO, whatever is surrounding a given selection should only concern that selection (while pasting from clipboard does not).

mjg:
> I agree with what @Technius said:
> 
> * Those are great additions, and improving the keyboard-free UX is absolutely necessary.
> * I find the icons not very explicit. They would benefit some more "common" forms. This includes the handles already present on master. Somehow, the cursor changing is what really helps understanding the rotation handle for instance.

Yes, I'm all for better forms or shapes or placements.

> * I also wonder about the `Paste` handle. It seems that the paste button in a toolbar/floating toolbox is more natural "goto" place for the feature. IMO, whatever is surrounding a given selection should only concern that selection (while pasting from clipboard does not).

I have to strongly disagree here, based on my own usage of the feature in a hybrid classroom setting over more than one term: quick access is the key point of this feature. You need the same quick access for paste in order to paste (what was copied) a second time or at a different place. And from a UI perspective, you look for "paste" at the same place where you look for "copy", so it's natural to have it there.

Just think about it this way: select and copy something, then select something else and press CTRL+v - what is going to be pasted? Not the current selection but what is in the clipboard (what you had copied). It's realy the same with the paste handle.

Floating toolbox would be OK (for all handles, including copy) if it didn't deselect the selection ... So, yes, my feature is more like using the space around the selection as a copy&paste toolbox since it works for both copy and paste.

bhennion:
> You need the same quick access for paste in order to paste (what was copied) a second time or at a different place.

- If your pasting something a second time, then the "Duplicate" handle of what you just pasted does that already.
- If you're pasting at a totally different place, then you need to select a random thing and paste. This is weird and definitely not intuitive. Moreover you can't do that if there is nothing to select (e.g. pasting to a new page).
     Also, should you paste via the handles or via the floating toolboxes, whatever you selected will no longer be selected in the end.

> And from a UI perspective, you look for "paste" at the same place where you look for "copy", so it's natural to have it there.

I don't quite agree on that. It's true for desktop interfaces (toolbar oriented), but false for a phone interface: you usually get a copy/cut as soon as something is selected. To paste, go somewhere, long press and a menu pops up. This is what I encountered from most touch-only devices I've interacted within the last years.

> Just think about it this way: select and copy something, then select something else and press CTRL+v - what is going to be pasted? Not the current selection but what is in the clipboard (what you had copied). It's realy the same with the paste handle.

It really isn't the same. Keyboard accelerations are not visually linked to your selection, while the handles are.

> Floating toolbox would be OK (for all handles, including copy) if it didn't deselect the selection

It is true that the floating toolbox has some UI issues which make it somewhat unreliable at this point. Still, fixing those issues seems like the way to go for the pasting.

Anyway, having a paste handle or not is not an important enough disagreement to block this PR. I won't oppose to its merge (after code review) once we find a better look for those handles.
How about a circular handle (I guess border and filling colors would be theme-dependent), with the theme's corresponding icons inside? Can we make this small enough to limit obstructiveness?

mjg:
Rebasing for good measure, before I look into UI alternatives (icons etc).

jerdoe:
> > Some brainstorming about obtrusiveness:
> 
> With small handles, I don't find the current version to be obtrusive. This might change with the suggested larger icons, of course.
> 
> > In OneNote, you can tap the selection to bring up a mini context menu:  Screenshot 2022-05-16 121618 Menu expanded:  Screenshot 2022-05-16 121645
> 
> This is similar to the suggestion in this PR, with the menu items arranged around the selection.
> 
> > Right now in Xournal++, duplication can be done with a right-click on the selection or the equivalent action with a pen, though I agree that discoverability of this feature is 'by chance' at best – some sort of context menu item or anchor would certainly be better.
> 
> This does not work over here, no matter how I configure the right button - you assign the button functions within xournalpp.
> 
> > Note that the context menu described above would show on left-click or normal tap, the latter of which might make input handling a bit tricky (conflict with scroll handler?). But left-click could be enough for a start.
> 
> If I understand this correctly, you would need to make the selection, then tap to bring up the menu, then tap to execute the action? That's one tap more. We often have a conflict between convenience and obtrusiveness, of course.
> 
> I had thought about putting the cut/copy/... actions in the floating toolbox. But (IIRC) the selection is deselected once you tap to bring up the toolbox, so this would work only if you bring up the toolbox in a different way.

I have assigned the floating toolbox to my second pen button, and when I bring it up, it does not cancel the selection. Does that differ when you implement a tap on the selection ? Couldn't it be something more related to the drag and drop handler to move the selection ?

PR #4022, opened 2022-05-10
Add ruled page template with vertical line on the right

lennonhill:
This adds a variant of the lined page template that features a vertical bar on the right hand side. I find it more convenient to use on devices that feature a physical touch screen when using the page's full width, as my hand doesn't have to go as much over the screen's edge.


bhennion:
Thanks for the PR @lennonhill, but as you mentionned #3969 has been merged and implemented that feature too. Closing this.

Issue #1917, opened 2020-04-30
request: add "Select All on Page" to Edit menu


kortschak:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I am using xournal as a distance education tool scratch pad and I want to delete an entire page.

**Describe the solution you'd like**
I would like to be able to select all elements on a page so that I can delete them with a <kbd>Ctrl</kbd>+<kbd>A</kbd> <kbd>Del</kbd>.

**Describe alternatives you've considered**
The current approach would be to use Page->New Page After menu item. This would work for the case that I have if there were key bindings for this (or user defined key bindings were a thing).

**Additional context**
This is in the category of ponies.

PR #4024, opened 2022-05-12
Add Select All option to edit menu

JulianHeuser:
Addresses issue #1917

Currently this feature selects all elements on the current page and layer. The menu item is under Edit. The shortcut is set to Ctrl+A.

If needed, it would probably be pretty easy to make it select everything on all pages instead, or change the menu location. I just chose what I thought made the most sense and is most consistent with other gtk apps.

image

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1054,12 +1054,13 @@ void Control::selectAllOnPage() {
         return;
     }
 
+    this->doc->lock();
     XojPageView* view = win->getXournal()->getViewFor(pageNr);
     if (view == nullptr) {
+        this->doc->unlock();
         return;
     }
 
-    this->doc->lock();
     PageRef page = this->doc->getPage(pageNr);
     Layer* layer = page->getSelectedLayer();
 


bhennion (changes requested):
Thanks for the PR! This is a long overdue addition of a very basic and important feature.

I have some small comments on the code itself. See below.

bhennion:
Also, if you want, you could add a `Unselect all` button too (with accelerator `Ctrl + Shift + A`)

src/core/control/Control.cpp
```diff
@@ -1045,6 +1048,37 @@ void Control::clearSelectionEndText() {
...
+        return;
+    }
+
+    PageRef page = this->doc->getPage(pageNr);
+    Layer* layer = page->getSelectedLayer();
+
+    win->getXournal()->clearSelection();
+
+    std::vector<Element*> elements;
+    for (Element* e: layer->getElements()) { elements.push_back(e); }
```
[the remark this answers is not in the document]
JulianHeuser: I tried it this way and for some reason it creates issues with selecting more than 2 things at once. This is the error I'm getting:
`** (xournalpp:31877): WARNING **: 14:27:31.405: Could not remove element from layer, it's not on the layer!`
and sometimes:
`** (xournalpp:31877): WARNING **: 14:38:16.294: Layer::insertElement() try to add an element twice!`
Then some items aren't selected or don't move properly after selection.

Assuming this is a bug, I think it's caused by how elements are ordered in the layer's element vector. This is also why I have to reverse the sort order (or at least re-instate it) in https://github.com/xournalpp/xournalpp/pull/4024#discussion_r872554054, because I'm reversing it here to get it to work. If someone fixed this we could make this change.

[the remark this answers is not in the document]
bhennion: Ah right, I see. The issue is here

https://github.com/xournalpp/xournalpp/blob/7da9a156b5e9bf9ee5167559966f7cde9c497321/src/core/control/tools/EditSelection.cpp#L88-L100

The elements get removed from the layer (so to our referenced vector) as they are added to the selection content. This is the reason for your messed up objects order too: in your situation, every call to `Layer::indexOf` return 0, a we already popped out the previous elements from the layer...

This is pretty inefficient in many respects:

- The copy aforementioned
- Successive calls to `Layer::removeElement` mean a complexity in O(n²) instead of O(n)
- Successive calls to `Layer::indexOf` mean a complexity in O(n²) instead of O(n) (we already know the index, since we are iterating on the layer content itself)

I think the most efficient solution would be:

1. To add a construct an empty `EditSelection` (using the first constructor) and do something like
```
Layer::Elementindex n = 0;
for (auto* e:layer->getElements()) {
    selection->addElement(e, n++);
}
layer->clear(false);
```
2. To add a `Layer::clear(bool free)`(used above) removing every element of a given layer (and optionally deleting them).

Alternatively, 1. could be a new constructor for the class `EditSelection` taking a `Layer*` as a parameter (among others).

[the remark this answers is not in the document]
JulianHeuser: Thanks for the suggestions, everything seems to be working now.

src/core/control/tools/EditSelection.h
```diff
@@ -41,6 +41,8 @@ class EditSelection: public ElementContainer, public Serializable {
     EditSelection(UndoRedoHandler* undo, Selection* selection, XojPageView* view);
     EditSelection(UndoRedoHandler* undo, Element* e, XojPageView* view, const PageRef& page);
     EditSelection(UndoRedoHandler* undo, const std::vector<Element*>& elements, XojPageView* view, const PageRef& page);
+    EditSelection(UndoRedoHandler* undo, const std::vector<Element*>& elements, XojPageView* view, const PageRef& page,
+                  Layer* layer);
```
bhennion, line 45:
```suggestion
    EditSelection(UndoRedoHandler* undo, XojPageView* view, const PageRef& page,
                  Layer* layer);
```

src/core/control/tools/EditSelection.cpp
```diff
@@ -99,6 +99,23 @@ EditSelection::EditSelection(UndoRedoHandler* undo, const vector<Element*>& elem
     view->rerenderPage();
 }
 
+EditSelection::EditSelection(UndoRedoHandler* undo, const vector<Element*>& elements, XojPageView* view,
+                             const PageRef& page, Layer* layer):
+        snappingHandler(view->getXournal()->getControl()->getSettings()) {
```
bhennion, line 104: The vector `elements` must be the layer's content (otherwise, you get a memory leak).

```suggestion
EditSelection::EditSelection(UndoRedoHandler* undo, XojPageView* view,
                             const PageRef& page, Layer* layer):
        snappingHandler(view->getXournal()->getControl()->getSettings()) {
    const auto& elements = layer->getElements();
```

src/core/model/Layer.h
```diff
@@ -52,6 +52,11 @@ class Layer {
      */
     ElementIndex removeElement(Element* e, bool free);
 
+    /**
+     * Removes all Element%s from the Layer
+     */
+    void clear();
```
bhennion, line 58: The fact it does not delete the elements must be abundantly clear

```suggestion
     * Removes all Elements from the Layer *without freeing them*
     */
    void clearNoFree();
```

src/core/control/Control.cpp
```diff
@@ -1045,6 +1048,34 @@ void Control::clearSelectionEndText() {
...
+    if (pageNr == npos) {
+        return;
+    }
+
+    XojPageView* view = win->getXournal()->getViewFor(pageNr);
+    if (view == nullptr) {
+        return;
+    }
+
+    this->doc->lock();
```
bhennion, line 1062: One last thing: I think the document should be locked already for this step.
```suggestion
    this->doc->lock();
    XojPageView* view = win->getXournal()->getViewFor(pageNr);
    if (view == nullptr) {
        this->doc->unlock();
        return;
    }

```

bhennion (approved):

bhennion:
Alright! Thanks a lot for the PR!
If you don't wanna add anything to it (and if noone else objects), I'll be merging this in 48 hours.

JulianHeuser:
Sound good, thanks for the help! If I decide to add a deselect I'll make another pull request.

PR #4025, opened 2022-05-12
Const-qualify Element::... and fix sign-conversion warnings relative to layer indices

bhennion:
The code-cleanup PR is two-fold:

- const-qualify getters in Element and derived classes
- Fix sign-conversion warnings relative to layer indices

Most of the changes are trivial, but a couple of points would require a rapid review (all in the second commit, relative to sign-conversions). I'll highlight those points with review comments

### DIFF
--- a/src/core/control/jobs/PreviewJob.cpp
+++ b/src/core/control/jobs/PreviewJob.cpp
@@ -66,7 +66,7 @@ void PreviewJob::drawPage() {
     PageRef page = this->sidebarPreview->page;
     Document* doc = this->sidebarPreview->sidebar->getControl()->getDocument();
     PreviewRenderType type = this->sidebarPreview->getRenderType();
-    int layer;
+    Layer::Index layer = 0;
 
     doc->lock();
 
@@ -79,7 +79,7 @@ void PreviewJob::drawPage() {
     // switch block can go away and the layer assignment into the remaining switch block.
     switch (type) {
         case RENDER_TYPE_PAGE_LAYER:
-            if (layer != -1) {
+            if (layer != 0) {
                 break;  // out
             }
             [[fallthrough]];
@@ -107,10 +107,10 @@ void PreviewJob::drawPage() {
         case RENDER_TYPE_PAGE_LAYER:
             // render single layer
             view.initDrawing(page, cr2, true);
-            if (layer == -1) {
+            if (layer == 0) {
                 view.drawBackground();
             } else {
-                Layer* drawLayer = (*page->getLayers())[layer];
+                Layer* drawLayer = (*page->getLayers())[layer - 1];
                 xoj::view::LayerView layerView(drawLayer);
                 layerView.draw(context);
             }
@@ -121,7 +121,7 @@ void PreviewJob::drawPage() {
             // render all layers up to layer
             view.initDrawing(page, cr2, true);
             view.drawBackground();
-            for (int i = 0; i <= layer; i++) {
+            for (Layer::Index i = 0; i < layer; i++) {
                 Layer* drawLayer = (*page->getLayers())[i];
                 xoj::view::LayerView layerView(drawLayer);
                 layerView.draw(context);
--- a/src/core/control/layer/LayerController.cpp
+++ b/src/core/control/layer/LayerController.cpp
@@ -30,18 +30,13 @@ void LayerController::pageSelected(size_t page) {
     fireRebuildLayerMenu();
 }
 
-void LayerController::insertLayer(PageRef page, Layer* layer, int layerPos) {
+void LayerController::insertLayer(PageRef page, Layer* layer, Layer::Index layerPos) {
     page->insertLayer(layer, layerPos);
     fireRebuildLayerMenu();
 }
 
 void LayerController::removeLayer(PageRef page, Layer* layer) {
     page->removeLayer(layer);
-    fireRebuildLayerMenu();
-}
-
-void LayerController::addLayer(PageRef page, Layer* layer) {
-    page->addLayer(layer);
     fireRebuildLayerMenu();
 }
 
@@ -78,8 +73,8 @@ auto LayerController::actionPerformed(ActionType type) -> bool {
 
         case ACTION_GOTO_NEXT_LAYER: {
             PageRef p = getCurrentPage();
-            int layer = p->getSelectedLayerId();
-            if (layer < static_cast<int>(p->getLayerCount())) {
+            auto layer = p->getSelectedLayerId();
+            if (layer < p->getLayerCount()) {
                 switchToLay(layer + 1, true);
             }
         }
@@ -87,7 +82,7 @@ auto LayerController::actionPerformed(ActionType type) -> bool {
 
         case ACTION_GOTO_PREVIOUS_LAYER: {
             PageRef p = getCurrentPage();
-            int layer = p->getSelectedLayerId();
+            auto layer = p->getSelectedLayerId();
             if (layer > 0) {
                 switchToLay(layer - 1, true);
             }
@@ -125,7 +120,7 @@ void LayerController::hideAllLayer() { hideOrHideAllLayer(false); }
  */
 void LayerController::hideOrHideAllLayer(bool show) {
     PageRef page = getCurrentPage();
-    for (size_t i = 1; i <= page->getLayerCount(); i++) { page->setLayerVisible(i, show); }
+    for (Layer::Index i = 1; i <= page->getLayerCount(); i++) { page->setLayerVisible(i, show); }
 
     fireLayerVisibilityChanged();
     control->getWindow()->getXournal()->layerChanged(selectedPage);
@@ -139,7 +134,7 @@ void LayerController::addNewLayer() {
     }
 
     auto* l = new Layer();
-    int layerPos = p->getSelectedLayerId();
+    auto layerPos = p->getSelectedLayerId();
     p->insertLayer(l, layerPos);
 
     control->getUndoRedoHandler()->addUndoAction(std::make_unique<InsertLayerUndoAction>(this, p, l, layerPos));
@@ -157,8 +152,8 @@ void LayerController::deleteCurrentLayer() {
         return;
     }
 
-    int lId = p->getSelectedLayerId();
-    if (lId < 1) {
+    auto lId = p->getSelectedLayerId();
+    if (lId == 0) {
         return;
     }
     Layer* l = p->getSelectedLayer();
@@ -184,19 +179,19 @@ void LayerController::moveCurrentLayer(bool up) {
         return;
     }
 
-    int lId = p->getSelectedLayerId();
+    auto lId = p->getSelectedLayerId();
     Layer* currentLayer = p->getSelectedLayer();
-    if (lId < 1) {
+    if (lId == 0) {
         // Background cannot be moved
         return;
     }
 
-    if (lId < 2 && !up) {
+    if (lId == 1 && !up) {
         // bottom layer cannot be moved down
         return;
     }
 
-    if (lId == static_cast<int>(p->getLayerCount()) && up) {
+    if (lId == p->getLayerCount() && up) {
         // top layer cannot be moved up
         return;
     }
@@ -206,7 +201,7 @@ void LayerController::moveCurrentLayer(bool up) {
     // Layer IDs are a bit strange, because background is 0
     // so the first layer is 1, technical the first layer is still
     // index 0 in the vector... confusing...
-    int newIndex = up ? lId : lId - 2;
+    auto newIndex = up ? lId : lId - 2;
     p->insertLayer(currentLayer, newIndex);
 
     MainWindow* win = control->getWindow();
@@ -255,8 +250,8 @@ void LayerController::mergeCurrentLayerDown() {
      * Layer indices in the vector are off by one from the layer IDs because
      * the background is not in the vector, so layer 1 has index 0 and so on.
      */
-    const size_t layerBelowIndex = ((size_t)layerBelowID) - 1;
-    Layer* layerBelow = page->getLayers()->at(layerBelowIndex);
+    const Layer::Index layerBelowIndex = layerBelowID - 1;
+    Layer* layerBelow = (*page->getLayers())[layerBelowIndex];
 
     UndoActionPtr undo_redo_action =
             std::make_unique<MergeLayerDownUndoAction>(this, page, currentLayer, layerBelow, layerID - 1, pageID);
@@ -274,8 +269,8 @@ void LayerController::copyCurrentLayer() {
         return;
     }
 
-    int lId = p->getSelectedLayerId();
-    if (lId < 1) {
+    auto lId = p->getSelectedLayerId();
+    if (lId == 0) {
         return;
     }
     Layer* l = p->getSelectedLayer();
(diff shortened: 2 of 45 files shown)


src/core/control/layer/LayerController.cpp
```diff
@@ -40,11 +40,6 @@ void LayerController::removeLayer(PageRef page, Layer* layer) {
     fireRebuildLayerMenu();
 }
 
-void LayerController::addLayer(PageRef page, Layer* layer) {
-    page->addLayer(layer);
-    fireRebuildLayerMenu();
-}
-
```
bhennion, line 47: This function was never used

src/core/control/layer/LayerController.h
```diff
@@ -35,9 +35,8 @@ class LayerController: public DocumentListener {
     void pageSelected(size_t page) override;
 
 public:
-    void insertLayer(PageRef page, Layer* layer, int layerPos);
+    void insertLayer(PageRef page, Layer* layer, size_t layerPos);
     void removeLayer(PageRef page, Layer* layer);
-    void addLayer(PageRef page, Layer* layer);
```
bhennion, line 40: Never used, see comment above.

src/core/control/jobs/PreviewJob.cpp
```diff
@@ -108,10 +108,10 @@ void PreviewJob::drawPage() {
         case RENDER_TYPE_PAGE_LAYER:
             // render single layer
             view.initDrawing(page, cr2, true);
-            if (layer == -1) {
+            if (layer == 0) {
                 view.drawBackground();
             } else {
-                Layer* drawLayer = (*page->getLayers())[layer];
+                Layer* drawLayer = (*page->getLayers())[layer - 1];
```
bhennion, line 114: index shift of one to avoid negative numbers. See consistent changes in SidebarPreviewLayerEntry

src/core/control/jobs/PreviewJob.cpp
```diff
@@ -122,7 +122,7 @@ void PreviewJob::drawPage() {
             // render all layers up to layer
             view.initDrawing(page, cr2, true);
             view.drawBackground();
-            for (int i = 0; i <= layer; i++) {
+            for (size_t i = 0; i < layer; i++) {
```
bhennion, line 125: index shift of 1. See above comment

src/core/gui/sidebar/previews/layer/SidebarPreviewLayerEntry.cpp
```diff
@@ -66,7 +66,7 @@ void SidebarPreviewLayerEntry::checkboxToggled() {
     }
 
     bool check = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(cbVisible));
-    (dynamic_cast<SidebarPreviewLayers*>(sidebar))->layerVisibilityChanged(layer + 1, check);
+    (dynamic_cast<SidebarPreviewLayers*>(sidebar))->layerVisibilityChanged(layerId, check);
```
bhennion, line 69: shift of 1, see above comment

src/core/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
```diff
@@ -107,12 +107,12 @@ void SidebarPreviewLayers::updatePreviews() {
...
 
-    int layerCount = page->getLayerCount();
+    auto layerCount = page->getLayerCount();
 
     size_t index = 0;
-    for (int i = layerCount; i >= 0; i--) {
+    for (auto i = layerCount + 1; i--;) {
         std::string name = lc->getLayerNameById(i);
-        SidebarPreviewBaseEntry* p = new SidebarPreviewLayerEntry(this, page, i - 1, name, index++, this->stacked);
+        SidebarPreviewBaseEntry* p = new SidebarPreviewLayerEntry(this, page, i, name, index++, this->stacked);
```
bhennion, line 115: Shift of 1, see above comment

src/core/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
```diff
@@ -107,12 +107,12 @@ void SidebarPreviewLayers::updatePreviews() {
         return;
     }
 
-    int layerCount = page->getLayerCount();
+    auto layerCount = page->getLayerCount();
 
     size_t index = 0;
-    for (int i = layerCount; i >= 0; i--) {
+    for (auto i = layerCount + 1; i--;) {
```
bhennion, line 113: Decreasing loop on unsigned variable. Is there a more readable way?

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
```diff
@@ -82,30 +82,24 @@ void ToolPageLayer::addSpecialButtonTop() {
...
-        }
-    }
-
-    if (layerId < 0) {
+    if (it == layerItems.end()) {
         g_warning("Invalid Layer Menu selected - not handled");
         return;
     }
-
+    auto layerId = it->first;
```
bhennion, line 99: Replace with std::find. Please double check what I wrote doesn't change the behaviour.

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
```diff
@@ -123,26 +117,21 @@ void ToolPageLayer::layerMenuShowClicked(GtkWidget* menu) {
...
-    }
+    auto it = std::find_if(showLayerItems.begin(), showLayerItems.end(),
+                           [menu](std::pair<const size_t, GtkWidget*>& a) { return a.second == menu; });
 
-    if (layerId < 0) {
+    if (it == showLayerItems.end()) {
         g_warning("Invalid Layer Show Menu selected - not handled");
         return;
     }
+    auto layerId = it->first;
```
bhennion, line 127: Replace with std::find. Please double check what I wrote doesn't change the behaviour.

src/core/plugin/luapi_application.h
```diff
@@ -941,7 +942,7 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
             luaL_error(L, "Current page has no pdf background, cannot use relative mode!");
         }
     }
-    if (selected >= 0 && selected < static_cast<int>(doc->getPdfPageCount())) {
+    if (selected < doc->getPdfPageCount()) {
```
bhennion, line 945: typeof(selected) == size_t

bhennion (commented):
Besides those comments, the remaining changes are trivial.
I tested everything worked as expected:

- Adding/removing/merging/renaming layers, and undo/redo of those operations
- The previews (entire page, individual layers and stack of layers)
- Hiding/showing individual layers

src/core/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
Febbe: The modern-cpp approach would be a reverse iterator. 
But this might be more readable:?
```suggestion
    for (auto i = layerCount + 1; i != 0;) {
        --i;
```

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
Febbe: You might want to create a lambda variable for the binary_operator like `auto menu_finder = [menu](std::pair<const size_t, GtkWidget*>& a) { return a.second == menu; };`

Since you require it more than once.

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
```diff
@@ -183,8 +172,8 @@ void ToolPageLayer::updateMenu() {
 
     addSpecialButtonTop();
 
-    int layerCount = lc->getLayerCount();
-    for (int layer = layerCount; layer > 0; layer--) {
+    auto layerCount = lc->getLayerCount();
+    for (auto layer = layerCount; layer > 0; layer--) {
```
Febbe, line 176: Shouldn't this be   
```suggestion
    auto layerCount = lc->getLayerCount();
    for (auto layer = layerCount; layer != 0; ) {
        layer--
```

Febbe: Must the order be reversed?

src/core/gui/toolbarMenubar/ToolPageLayer.h
```diff
@@ -75,8 +75,8 @@ class ToolPageLayer: public AbstractToolItem, public LayerCtrlListener {
     GtkWidget* layerButton = nullptr;
     GtkWidget* menu = nullptr;
 
-    std::map<int, GtkWidget*> layerItems;
-    std::map<int, GtkWidget*> showLayerItems;
+    std::map<size_t, GtkWidget*> layerItems;
+    std::map<size_t, GtkWidget*> showLayerItems;
```
Febbe, line 79: **Question:** What does size_t represent here, an index, Count, hash, some norm, an Apple?

Maybe use something like:
```suggestion
    using AppleJuiceIndex = size_t;
    std::map<AppleJuiceIndex, GtkWidget*> layerItems;
    std::map<AppleJuiceIndex, GtkWidget*> showLayerItems;
```

Also, both std::map and std::unordered_map have huge drawbacks.    
Do holes appear? No + starts with 0? -> std::vector.    
Yes? Must it be sorted? -> leaving it as map might be ok for now (A sorted vector might be better).    
Else -> use std::unordered_map (google::dense_map might be better)

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
bhennion: No, not this one: the case layer == 0 (i.e. the background) is treated separatly just belonw this chunk of code.

src/core/gui/toolbarMenubar/ToolPageLayer.h
bhennion: Hmm... You're right, there was no reason to have a std::map here to begin with. I replaced it with a std::vector. Good catch!

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
```diff
@@ -123,26 +116,20 @@ void ToolPageLayer::layerMenuShowClicked(GtkWidget* menu) {
...
-            break;
-        }
-    }
-
-    if (layerId < 0) {
+    if (it == showLayerItems.end()) {
         g_warning("Invalid Layer Show Menu selected - not handled");
         return;
     }
+    size_t layerId = (size_t)std::distance(showLayerItems.begin(), it);
```
Febbe, line 125: **minor nitpick:** prefer `static_cast<T>(x)` over `T(x)` over `(T)x`

src/core/model/XojPage.cpp
```diff
@@ -36,80 +36,67 @@ void XojPage::addLayer(Layer* layer) {
...
 
-void XojPage::removeLayer(Layer* layer) {
-    for (unsigned int i = 0; i < this->layer.size(); i++) {
-        if (layer == this->layer[i]) {
-            this->layer.erase(this->layer.begin() + i);
-            break;
-        }
+void XojPage::removeLayer(Layer* l) {
+    if (auto it = std::find(layer.begin(), layer.end(), l); it != layer.end()) {
+        this->layer.erase(it);
```
Febbe, line 51: That's dangerous. Both layer.end() and it are invalidated after erase. Therefore, you must use the iterator returning from erase. But there is an even better solution:

cont.erase + std::remove(if):

```suggestion
    auto const& endi = end(layer);
    this->layer.erase(std::remove(begin(layer),endi, l),endi);
```

src/core/model/XojPage.h
```diff
@@ -56,17 +56,16 @@ class XojPage: public PageHandler {
...
     Color getBackgroundColor() const;
 
     std::vector<Layer*>* getLayers();
-    size_t getLayerCount();
-    int getSelectedLayerId();
-    void setSelectedLayerId(int id);
-    static bool isLayerVisible(Layer* layer);
-    bool isLayerVisible(int layerId);
+    size_t getLayerCount() const;
+    size_t getSelectedLayerId();
```
Febbe, line 66: Does this getter modify something? Or can it be const.

src/core/undo/InsertLayerUndoAction.cpp
```diff
@@ -9,10 +9,12 @@
...
 
 InsertLayerUndoAction::InsertLayerUndoAction(LayerController* layerController, const PageRef& page, Layer* layer,
-                                             int layerPosition):
-        UndoAction("InsertLayerUndoAction"), layerPosition(layerPosition), layerController(layerController) {
+                                             size_t layerPosition):
+        UndoAction("InsertLayerUndoAction"),
+        layerPosition(layerPosition),
+        layerController(layerController),
+        layer(layer) {
     this->page = page;
```
Febbe, line 17:
```suggestion
        layer(layer). 
        page(page) {
```

src/core/undo/RemoveLayerUndoAction.cpp
```diff
@@ -9,11 +9,9 @@
 #include "util/i18n.h"
 
 RemoveLayerUndoAction::RemoveLayerUndoAction(LayerController* layerController, const PageRef& page, Layer* layer,
-                                             int layerPos):
-        UndoAction("RemoveLayerUndoAction"), layerController(layerController) {
+                                             size_t layerPos):
+        UndoAction("RemoveLayerUndoAction"), layerController(layerController), layer(layer), layerPos(layerPos) {
     this->page = page;
```
Febbe, line 14:
```suggestion
        UndoAction("RemoveLayerUndoAction"), layerController(layerController), layer(layer), layerPos(layerPos),
        page(page) {
```

Febbe:
Overall, this LGTM after you resolve all conversations.

src/core/gui/sidebar/previews/layer/SidebarPreviewLayers.cpp
Technius: I would make it explicit that it's unsigned.

Secondly, a good way to iterate backwards over an unsigned counter is to "shift" the count by 1:
```c++
for (unsigned i = n + 1; i > 0; i--) {
  (void)foo[i - 1];
}

// or decrement inside loop

for (unsigned i = n + 1; i > 0;) {
  i--;
  (void)foo[i];
}
```

src/core/model/XojPage.h
```diff
@@ -56,17 +56,16 @@ class XojPage: public PageHandler {
...
+    void setSelectedLayerId(size_t id);
```
Technius, line 67: We should typedef the layer ids to avoid confusion, e.g.

```
    void setSelectedLayerId(Layer::Id id);
```

src/core/model/XojPage.h
Technius: Looks like `getSelectedLayerId` will insert a layer if there are no layers in the page. That's pretty undesirable imo, we should refactor it (perhaps in a followup) so that the page layers are always in a valid state.

src/core/undo/InsertLayerUndoAction.cpp
bhennion: No can do without refactoring the undo business. `page` is a member of the base class UndoAction (and some derived classes don't set it (e.g. `EmergencySaveRestore`). Changing this would go to deep for this PR.

src/core/undo/RemoveLayerUndoAction.cpp
bhennion: same as above

src/core/model/XojPage.h
bhennion: I agree there is something here. For a follow up PR.

src/core/model/XojPage.h
bhennion: We should have a clear scheme for this kind of things (maybe add it to the coding convention page). If we adopt `Layer::Index`, then the `Layer::ElementIndex` member type should probably be moved to `Element::Index`.

Febbe: > If we adopt Layer::Index, then the Layer::ElementIndex member type should probably be moved to Element::Index

Exactly

src/core/undo/InsertLayerUndoAction.cpp
Febbe: Ok, I didn't know that this is a base class member.

src/core/model/XojPage.h
bhennion: Done!

bhennion:

Issue #3765, opened 2022-02-02
No input in textbox after scrolling gives focus to other page


fpesari:
**Affects versions :**
 - OS: openSUSE Tumbleweed
 - (Linux only) Desktop environment: KDE Plasma 5.23.5 on X11
 - Which version of libgtk do you use: 3.24.31
 - Version of Xournal++: 1.1.0
 - Installation method: openSUSE repository

**Describe the bug**
Text boxes stop taking any input when scrolling gives focus to a different page than the one which contains said textbox

**To Reproduce**
1. Open a PDF with more than one page to annotate (or create an empty document with > 1 pages)
2. Put a textbox inside a page
3. Scroll with the mouse wheel until the next (or previous) page is focused (red outline)
4. Type something on your keyboard. The text cursor will keep blinking inside the text box but nothing will happen

**Expected behavior**
Either:
1. (my preferred option) The textbox too goes out of focus and becomes "closed"; in other words, editing mode for that textbox is turned off (the equivalent of pressing Escape after typing something)
2. You can keep writing into the textbox, despite its page being unfocused

bhennion:
This has been fixed in the release branch.

PR #4027, opened 2022-05-12
Unfocus text boxes after the selected page changes

JulianHeuser:
Addresses #3765

This pr stops editing of the textbox once you change pages. Before, it would stay selected but you couldn't type until you went back to the page it was on.


Technius:
Thanks for the PR! Can you change the target branch to `release-1.1`?

Technius (approved):

JulianHeuser:
Still new to this but got there in the end, re-tested and it works. Dunno if there's a better way to change branch than a force push but it's one line so no big deal.

PR #4037, opened 2022-05-19
page template: correct German word order

mjg:


rolandlo (approved):
Thanks for the correction. Merging in 24 hours, if no objections are raised.

PR #4042, opened 2022-05-20
Fix data race in InputHandler's

### LINKED ISSUE #3579: Possible data race in `StrokeHandler`, `BaseStrokeHandler`, `SplineHandler` (possible others)

### LINKED ISSUE #4038: Frequent crashes, Wacom note taking on Arch Linux
**Affects versions :**
 - OS: Arch Linux (Kernel 5.15.33-1-lts)
- Xournal++ 1.1.1
└──libgtk: 3.24.33
 - Installation method: pacman

**Describe the bug**
I use xournalpp for notetaking in ~2 hour sessions (university) with a Wacom Intuous BT S tablet. Throughout ~50% of these sessions, Xournalpp will crash upon beginning a stroke. I saw it happening 5 minutes in a session and also 55 minutes in, so runtime of Xournalpp does not seem to have a big influence. Xournalpp manages to emergency save every time. Journal messages say "Stroke incomplete!".

**To Reproduce**
I wish I knew. Possibly related to Bluetooth drawing tablet? I really don't know. Happy to assist in debugging this however, just provide guidance (debug build, etc.)

**Additional context**

Journal output of a crashing session:

```
May 19 12:10:12 gg dbus-daemon[423]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.4591' (uid=1000 pid=597551 comm="xournalpp")
May 19 12:15:26 gg xournalpp[599234]: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale
May 19 12:15:26 gg xournalpp[599234]: Plugin "MigrateFontSizes" UI initialized
May 19 12:15:32 gg dbus-daemon[423]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.4598' (uid=1000 pid=599234 comm="xournalpp")
May 19 12:18:27 gg xournalpp[599234]: Info: autosave document...
May 19 12:18:27 gg xournalpp[599234]: Autosaving to ...
May 19 13:00:27 gg xournalpp[599234]: Info: autosave document...
May 19 13:00:27 gg xournalpp[599234]: Autosave renamed from ...
May 19 13:00:27 gg xournalpp[599234]: Autosaving to ....
May 19 13:03:27 gg xournalpp[599234]: Info: autosave document...
May 19 13:03:27 gg xournalpp[599234]: Autosave renamed from ...
May 19 13:03:27 gg xournalpp[599234]: Autosaving to...
May 19 13:06:27 gg xournalpp[599234]: Info: autosave document...
May 19 13:06:27 gg xournalpp[599234]: Autosave renamed from ....
May 19 13:06:27 gg xournalpp[599234]: Autosaving to ....
May 19 13:09:27 gg xournalpp[599234]: Info: autosave document...
May 19 13:09:27 gg xournalpp[599234]: Autosave renamed from ...
May 19 13:09:27 gg xournalpp[599234]: Autosaving to ...
May 19 13:10:57 gg xournalpp[599234]: [Crash Handler] Crashed with signal 11
May 19 13:10:57 gg xournalpp[599234]: [Crash Handler] Wrote crash log to: /home/.../.cache/xournalpp/errorlogs/errorlog.20220519-131057.log
May 19 13:10:57 gg xournalpp[599234]: Trying to emergency save the current open document…
May 19 13:10:57 gg xournalpp[599234]: Successfully saved document to "/home/.../.config/xournalpp/emergencysave.xopp"
May 19 13:11:00 gg xournalpp[601025]: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale
May 19 13:11:00 gg xournalpp[601025]: Plugin "MigrateFontSizes" UI initialized
May 19 13:11:07 gg dbus-daemon[423]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.4621' (uid=1000 pid=601025 comm="xournalpp")

<I guess here I restarted> 

May 19 13:13:48 gg dbus-daemon[423]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.4623' (uid=1000 pid=601025 comm="xournalpp")
May 19 13:14:01 gg xournalpp[601025]: Info: autosave document...
May 19 13:14:01 gg xournalpp[601025]: Autosaving to ...
May 19 13:14:39 gg xournalpp[601025]: Stroke incomplete!
```

And the errorlog:

```
Date: Thu May 19 13:10:57 2022
Error: signal 11
[bt]: (0) xournalpp(+0x1ad08b) [0x563caa7b808b]
[bt]: (1) /usr/lib/libc.so.6(+0x42560) [0x7fa80a156560]
[bt]: (2) xournalpp(_ZNK12DocumentView11drawElementEP6_cairoP7Element+0x21) [0x563caa7ae021]
[bt]: (3) xournalpp(_ZN12DocumentView9drawLayerEP6_cairoP5Layer+0x85) [0x563caa7ae255]
[bt]: (4) xournalpp(_ZN12DocumentView8drawPageESt10shared_ptrI7XojPageEP6_cairobbbb+0xf5) [0x563caa7ae365]
[bt]: (5) xournalpp(_ZN10PreviewJob8drawPageEv+0x3cb) [0x563caa6d665b]
[bt]: (6) xournalpp(_ZN10PreviewJob3runEv+0x75) [0x563caa6d6795]
[bt]: (7) xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0x127) [0x563caa6d17a7]
[bt]: (8) /usr/lib/libglib-2.0.so.0(+0x84da5) [0x7fa80b6cdda5]
[bt]: (9) /usr/lib/libc.so.6(+0x8d5c2) [0x7fa80a1a15c2]
[bt]: (10) /usr/lib/libc.so.6(clone+0x44) [0x7fa80a226584]


[bt] #1 xournalpp(+0x1ad554) [0x563caa7b8554]
??:0
[bt] #2 /usr/lib/libc.so.6(+0x42560) [0x7fa80a156560]
??:0
[bt] #3 xournalpp(_ZNK12DocumentView11drawElementEP6_cairoP7Element+0x21) [0x563caa7ae021]
??:0
[bt] #4 xournalpp(_ZN12DocumentView9drawLayerEP6_cairoP5Layer+0x85) [0x563caa7ae255]
??:0
[bt] #5 xournalpp(_ZN12DocumentView8drawPageESt10shared_ptrI7XojPageEP6_cairobbbb+0xf5) [0x563caa7ae365]
??:0
[bt] #6 xournalpp(_ZN10PreviewJob8drawPageEv+0x3cb) [0x563caa6d665b]
??:0
[bt] #7 xournalpp(_ZN10PreviewJob3runEv+0x75) [0x563caa6d6795]
??:0
[bt] #8 xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0x127) [0x563caa6d17a7]
??:0
[bt] #9 /usr/lib/libglib-2.0.so.0(+0x84da5) [0x7fa80b6cdda5]
??:0
[bt] #10 /usr/lib/libc.so.6(+0x8d5c2) [0x7fa80a1a15c2]
??:0
[bt] #11 /usr/lib/libc.so.6(clone+0x44) [0x7fa80a226584]
??:0
```

bhennion:
This PR should fix #3579 (and possibly #4038).

It is not as clean as I'd want, and I'd like some advice on how this kind of mutex handling is best implemented. 
At least, there does not seem to be any deadlock happening (but I only tested it briefly).

@talkdirty: would you mind checking this PR out and report on whether or not this fixes #4038?

### DIFF
--- a/src/control/tools/BaseStrokeHandler.cpp
+++ b/src/control/tools/BaseStrokeHandler.cpp
@@ -16,13 +16,15 @@ guint32 BaseStrokeHandler::lastStrokeTime;  // persist for next stroke
 BaseStrokeHandler::BaseStrokeHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page, bool flipShift,
                                      bool flipControl):
         InputHandler(xournal, redrawable, page),
-        snappingHandler(xournal->getControl()->getSettings()),
         flipShift(flipShift),
-        flipControl(flipControl) {}
+        flipControl(flipControl),
+        snappingHandler(xournal->getControl()->getSettings()) {}
 
 BaseStrokeHandler::~BaseStrokeHandler() = default;
 
 void BaseStrokeHandler::draw(cairo_t* cr) {
+    std::lock_guard lock(this->strokeMutex);
+
     if (!stroke) {
         return;
     }
@@ -36,8 +38,6 @@ void BaseStrokeHandler::draw(cairo_t* cr) {
 
 auto BaseStrokeHandler::onKeyEvent(GdkEventKey* event) -> bool {
     if (event->is_modifier) {
-        Rectangle<double> rect = stroke->boundingRect();
-
         PositionInputData pos{};
         pos.x = pos.y = pos.pressure = 0;  // not used in redraw
         if (event->keyval == GDK_KEY_Shift_L || event->keyval == GDK_KEY_Shift_R) {
@@ -52,18 +52,18 @@ auto BaseStrokeHandler::onKeyEvent(GdkEventKey* event) -> bool {
             return false;
         }
 
-        this->redrawable->repaintRect(stroke->getX(), stroke->getY(), stroke->getElementWidth(),
-                                      stroke->getElementHeight());
+        Rectangle<double> rect;
+        {  // lock_guard scope
+            std::lock_guard lock(this->strokeMutex);
 
+            rect = stroke->boundingRect();
 
-        Point malleablePoint = this->currPoint;  // make a copy as it might get snapped to grid.
-        this->drawShape(malleablePoint, pos);
+            Point malleablePoint = this->currPoint;  // make a copy as it might get snapped to grid.
+            this->drawShape(malleablePoint, pos, lock);
 
-
-        rect.unite(stroke->boundingRect());
-
-        double w = stroke->getWidth();
-        redrawable->repaintRect(rect.x - w, rect.y - w, rect.width + 2 * w, rect.height + 2 * w);
+            rect.unite(stroke->boundingRect());
+        }
+        redrawable->repaintRect(rect.x, rect.y, rect.width, rect.height);
 
         return true;
     }
@@ -71,38 +71,41 @@ auto BaseStrokeHandler::onKeyEvent(GdkEventKey* event) -> bool {
 }
 
 auto BaseStrokeHandler::onMotionNotifyEvent(const PositionInputData& pos) -> bool {
-    if (!stroke) {
-        return false;
-    }
 
     double zoom = xournal->getZoom();
     double x = pos.x / zoom;
     double y = pos.y / zoom;
-    int pointCount = stroke->getPointCount();
 
     Point currentPoint(x, y);
-    Rectangle<double> rect = stroke->boundingRect();
+    Rectangle<double> rect;
 
-    if (pointCount > 0) {
-        if (!validMotion(currentPoint, stroke->getPoint(pointCount - 1))) {
-            return true;
+    {  // lock_guard scope
+        std::lock_guard lock(this->strokeMutex);
+
+        if (!stroke) {
+            return false;
         }
+
+        if (int pointCount = stroke->getPointCount(); pointCount > 0) {
+            if (!validMotion(currentPoint, stroke->getPoint(pointCount - 1))) {
+                return true;
+            }
+        }
+
+        rect = stroke->boundingRect();
+
+        drawShape(currentPoint, pos, lock);
+
+        rect.unite(stroke->boundingRect());
     }
 
-    this->redrawable->repaintRect(stroke->getX(), stroke->getY(), stroke->getElementWidth(),
-                                  stroke->getElementHeight());
-
-    drawShape(currentPoint, pos);
-
-    rect.unite(stroke->boundingRect());
-    double w = stroke->getWidth();
-
-    redrawable->repaintRect(rect.x - w, rect.y - w, rect.width + 2 * w, rect.height + 2 * w);
+    redrawable->repaintRect(rect.x, rect.y, rect.width, rect.height);
 
     return true;
 }
 
 void BaseStrokeHandler::onMotionCancelEvent() {
+    std::lock_guard lock(this->strokeMutex);
     delete stroke;
     stroke = nullptr;
 }
@@ -110,10 +113,11 @@ void BaseStrokeHandler::onMotionCancelEvent() {
 void BaseStrokeHandler::onButtonReleaseEvent(const PositionInputData& pos) {
     xournal->getCursor()->activateDrawDirCursor(false);  // in case released within  fixate_Dir_Mods_Dist
 
+    std::lock_guard lock(this->strokeMutex);
+
     if (stroke == nullptr) {
         return;
     }
-
 
     Control* control = xournal->getControl();
     Settings* settings = control->getSettings();
@@ -150,7 +154,6 @@ void BaseStrokeHandler::onButtonReleaseEvent(const PositionInputData& pos) {
         BaseStrokeHandler::lastStrokeTime = pos.timestamp;
     }
 
-
     // This is not a valid stroke
     if (stroke->getPointCount() < 2) {
         g_warning("Stroke incomplete!");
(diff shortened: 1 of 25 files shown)


Febbe (commented):
**Literature**

For Concurrency tips you should read [this](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cpcon-concurrency)   
A book which is very good: Concurrency in Action from Anthony Williams

Important is: [CP.31](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-data-by-value) [CP.43](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-time) [CP.44](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-name) [CP.50](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-mutex)

I think you acquire the lock for too long. Try to find as many blocks which have a minimum size.

src/control/tools/InputHandler.cpp
```diff
@@ -32,7 +31,7 @@ void InputHandler::resetShapeRecognizer() {
     // Does nothing here. Implemented in the extending classes
 }
 
-void InputHandler::createStroke(Point p) {
+void InputHandler::createStroke(Point p, const std::lock_guard<std::recursive_mutex>&) {
```
Febbe, line 34: **General Todo for a refactoring:** I am not happy with the naming: createXXX should either return a Stroke or a unique / shared pointer to a Stroke. Not void.

Febbe: this for example is a perfect way to show, that the locking is too wide. The only thing, which must be protected is the line 37: `stroke = new Stroke();`
also this line can be moved to the end of the function / the return value by replacing the line with `auto stroke = std::make_unique<Stroke>(); ...; this->stroke = stroke.release();`

Febbe, line 34: Don't provide lock_guard's in the function header, certainly never as const reference. It may be useful, to assure, that a lock is acquired. But then it has to be either replaced with std::unique_lock (value semantic or non-const reference).

bhennion:
I guess all this means we'll need a deeper PR, which may no longer be suitable for the release branch.
I'll try to integrate your feedback while keeping the changes to a minimum. We'll see if I can figure something out.

Febbe:
No, I think the easiest solution would be, to replace that pointer with a shared_ptr. And you just use a std::swap on it, when it is updated. But that means that it is not allowed to change the stroke without making a copy of it (and swap the new shared_ptr into it). Then you only have to protect the write operation on the shared_ptr (unless you are using atomic<shared_ptr>)

Febbe:
Also removing all those obsolete references to the lock_guard will reduce the diff.

bhennion:
I made a second attempt. Things are probably more like @Febbe recommended in StrokeHandler and BaseStrokeHandler. However, I do not see how to do things properly in SplineHandler without refactoring.
Basically, the mutex needs to protect all the data (the stroke, the knots, and the tangent vectors), and those are used (either read or write) on almost every line of code.

I think the best way to handle this would be to split the data into "old and fixed points" and "moving spline segment". The old and fixed data should be painted onto a mask in the main loop (the mask begin protected by a mutex, ofc). The moving segment should be shared by value with the drawing thread. Those kind of changes go beyond what can be done on the release branch, I think.

I'd say we have a choice to make for the SplineHandler: either we leave the data race be, or we use this "overlocking" version.

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -150,30 +139,31 @@ void BaseStrokeHandler::onButtonReleaseEvent(const PositionInputData& pos) {
...
-
-
     control->getLayerController()->ensureLayerExists(page);
 
     Layer* layer = page->getSelectedLayer();
 
     UndoRedoHandler* undo = control->getUndoRedoHandler();
 
-    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke));
+    Stroke* s;
```
Febbe, line 148:
```suggestion
    std::unique_ptr<Stroke> s;
```

src/control/tools/BaseStrokeHandler.cpp
```diff
@@ -150,30 +139,31 @@ void BaseStrokeHandler::onButtonReleaseEvent(const PositionInputData& pos) {
...
+    {
+        std::lock_guard lock(strokeMutex);
+        s = stroke.release();
```
Febbe, line 151:
```suggestion
        s = std::move(stroke);
```

src/control/tools/BaseStrokeHandler.h
```diff
@@ -59,8 +61,11 @@ class BaseStrokeHandler: public InputHandler {
 protected:
     DocumentView view;
     Point currPoint;
-    Point buttonDownPoint;  // used for tapSelect and filtering - never snapped to grid. See startPoint defined in
-                            // derived classes such as CircleHandler.
+    Point buttonDownPoint;  // used for tapSelect and filtering - never snapped to grid.
+    Point startPoint;       // May be snapped to grid
+
+    std::mutex strokeMutex;  // protects InputHandler::stroke
```
Febbe, line 67: Why is this mutex not located next to the stroke?

bhennion: Because BaseStrokeHandler::draw uses the stroke, so the stroke needs protecting in this derived class (of InputHandler). On the other hand StrokeHandler::draw uses a mask, so the stroke does not need protecting in this derived class.
I don't know how to make this cleanly...

PR #4043, opened 2022-05-21
Fixed X11 Hand Recognition not working 

### LINKED ISSUE #4040: Hand Recognition does not work since a while now

**Affects versions :**
 - OS: Linux
 - Which version of libgtk do you use : irrelevant
 - git:xournalpp/master
 - apt / self compiled

**Describe the bug**
Hand Recognition doesn't work properly anymore. Now, when I press my hand to the screen, it interups the drawing of the pen

grafik

**To Reproduce**
1. Use Hand Recognition
2. start to draw
3. Pres your hand on the screen

**Expected behavior**
Clear, not interrupted line. No hand gestures

**Additional context**
I am bisecting it currently, and it is a regression.

Febbe:
Fixes #4040


Technius:
Does the bug affect 1.1.x as well?

Febbe:
Shouldn't be.

PR #4045, opened 2022-05-22
Fixed Undefined Behaviour in ColorToolItem.cpp:142

Febbe:
Will merge after I found out, if this also happens in the newest release

### DIFF
--- a/src/core/gui/MainWindowToolbarMenu.cpp
+++ b/src/core/gui/MainWindowToolbarMenu.cpp
@@ -1,5 +1,6 @@
 #include "MainWindowToolbarMenu.h"
 
+#include "control/settings/Settings.h"
 #include "gui/toolbarMenubar/ToolMenuHandler.h"
 #include "gui/toolbarMenubar/model/ToolbarData.h"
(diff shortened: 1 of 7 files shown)


Febbe:
Since this is a regression, I will merge it. There are other cases of UB I want to fix directly after this.

idotobi:
Thanks for the fix :pray: 
Could you give me a pointer why this was undefined behaviour? Just so I don't do it again :sweat_smile:

Febbe:
> Thanks for the fix :pray: 
> Could you give me a pointer why this was undefined behaviour? Just so I don't do it again :sweat_smile: 

You returned a string by value, not by const reference. Then, you called cstr on it, without keeping the string value. The cstr ptr was then a dangling pointer.

idotobi:
I see :bulb:  Thanks for explaining.
 I'll check the my `c_str` usage in the upcomming PR. Might very well have used it too liberally elsewhere.

Febbe:
You could also compile xournalpp with `-g -O1 -fno-omit-frame-pointer -fsanitize=address -fsanitize=undefinded`. This will automatically crash the app, when such an error occurs.

PR #4054, opened 2022-05-27
Fix instability in ErasableStroke.testIntersectWithPaddedBox

### LINKED ISSUE #4050: ErasableStroke.testIntersectWithPaddedBox test fails with `-O2 -march=native`
**Affects versions :**
 - OS: Arch Linux (probably affects all Linux distros)
 - (Linux only) Desktop environment: WM-only X11
 - Which version of libgtk do you use: 3.24.34
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Version of Xournal++: latest commit (though has happened for quite a while)
     - can be obtained by running `xournalpp --version` or in the **About** window

**Describe the bug**
When compiling and running tests with `-O2 -march=native` (or higher optimization level), the test `ErasableStroke.testIntersectWithPaddedBox` fails.

**To Reproduce**
1. Make sure docker is installed and the `archlinux:base-devel` image is available
2. Open a terminal and place the following file in the PWD: 
```sh
#!/usr/bin/env sh

export MAKEFLAGS="-j5"

# echo "======================================================================="
sudo pacman --noconfirm -Syu git cmake help2man gtest texlive-bin gtk3 \
	poppler-glib libxml2 portaudio libsndfile lua53 lua53-lgi libzip
cd ~


echo "CLONING ==============================================================="
git clone 'https://github.com/xournalpp/xournalpp.git' --depth=1
cp -r xournalpp xournalpp-o1
cp -r xournalpp xournalpp-o2
cp -r xournalpp xournalpp-o3

echo "BUILDING =============================================================="
cd ~/xournalpp
mkdir build; cd build
export CFLAGS=" -march=native "
export CXXFLAGS=" $CFLAGS "
cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..
cmake --build .
echo "RUNNING TESTS ========================================================="
cmake --build . --target test-units
cmake --build . --target test

echo "BUILDING (O1) ========================================================="
cd ~/xournalpp-o1
mkdir build; cd build
export CFLAGS=" -march=native -O1 "
export CXXFLAGS=" $CFLAGS "
cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..
cmake --build .
echo "RUNNING TESTS (O1) ===================================================="
cmake --build . --target test-units
cmake --build . --target test

echo "BUILDING (O2) ========================================================="
cd ~/xournalpp-o2
mkdir build; cd build
export CFLAGS=" -march=native -O2 "
export CXXFLAGS=" $CFLAGS "
cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..
cmake --build .
echo "RUNNING TESTS (O2) ===================================================="
cmake --build . --target test-units
cmake --build . --target test

echo "BUILDING (O3) ========================================================="
cd ~/xournalpp-o3
mkdir build; cd build
export CFLAGS=" -march=native -O3 "
export CXXFLAGS=" $CFLAGS "
cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DENABLE_GTEST=ON ..
cmake --build .
echo "RUNNING TESTS (O3) ===================================================="
cmake --build . --target test-units
cmake --build . --target test
```
3. Run `sudo docker run -it --rm -v "$PWD:/repro" --cpuset-cpus="0-4" archlinux:base-devel /repro/repro`
4. The test will fail for the O2 and O3 cases, but not for O1 and no optimization.

**Expected behavior**
The tests should at least succeed on `-O2`, since that is usually considered stable (failing on `-O3` is not ideal but understandable).

**Screenshots of Problem**
N/A

**Additional context**
My CPU is an AMD Ryzen 5 2600, though I suspect that this can be at least reproduced on other Ryzen 5 CPUs.

bhennion:
Fix #4050 

The issue was due to numerical imprecisions, which, for some reason, only appear with those compile flags.
They are irrelevent for the app (there is a safe-guard that forbids wrongly erasing parts of a stroke, precisely to catch those kind of cases).

This PR simply modifies the test case to remove those imprecision issues.
The borderline cases that trigger such issues: when a stroke point is exactly on the eraser box's boundary. As it seems the compiler options (harmlessly) change the app behaviour in those cases, they simply should not be part of the test unit.

### DIFF
--- a/test/unit_tests/model/ErasableStrokeTest.cpp
+++ b/test/unit_tests/model/ErasableStrokeTest.cpp
@@ -199,8 +199,8 @@ TEST(ErasableStroke, testIntersectWithPaddedBox) {
     boxes[1] = {Point(1, 1), 1, 3};
     expectedResult[1] = {{0, 0.0}, {1, 2.0 / 3.0}};
     // Start in padding. Pass by inner box. End in padding going toward inner box
-    boxes[2] = {Point(1, 1), 2, 3};
-    expectedResult[2] = {{0, 0.0}, {1, 2.0 / 3.0}, {9, 1.0}, {10, 1.0}};
+    boxes[2] = {Point(1, 1), 1.9, 3.5};
+    expectedResult[2] = {{0, 0.0}, {1, 5.0 / 6.0}, {9, 5.0 / 6.0}, {10, 1.0}};
     // Start on padding boundary, pass by inner box
     boxes[3] = {Point(1, 1), 0.5, 1};
     expectedResult[3] = {{0, 0.0}, {0, 1.0}};
@@ -226,8 +226,8 @@ TEST(ErasableStroke, testIntersectWithPaddedBox) {
         printf("Test box %u: ", i);
         for (auto p: res) { printf("(%zu, %4.2f) ", p.index, p.t); }
         printf("\n");
-        ASSERT_EQ(res.size(), expectedResult[i].size());
-        for (unsigned int j = 0; j < res.size(); ++j) {
+        EXPECT_EQ(res.size(), expectedResult[i].size());
+        for (unsigned int j = 0; j < std::min(res.size(), expectedResult[i].size()); ++j) {
             printf("   Found (%zu, %4.2f) - expected (%zu, %4.2f)\n", res[j].index, res[j].t,
                    expectedResult[i][j].index, expectedResult[i][j].t);
             EXPECT_EQ(res[j].index, expectedResult[i][j].index);


bhennion:
There an unrelated CI failure during the Windows build
```
5/325] Generating win32/xournalpp.ico
FAILED: src/win32/xournalpp.ico D:/a/1/s/build/src/win32/xournalpp.ico 
cmd.exe /C "cd /D D:\a\1\s\build\src && convert D:/a/1/s/build/xournalpp-icon-16.png D:/a/1/s/build/xournalpp-icon-32.png D:/a/1/s/build/xournalpp-icon-48.png D:/a/1/s/build/xournalpp-icon-256.png D:/a/1/s/build/src/win32/xournalpp.ico"
```

MithicSpirit:
Can confirm this fixes #4050 for me.

Issue #2808, opened 2021-02-15
Make all arrowheads solid, regardless of pen type


tim-clifford:
**Is your feature request related to a problem? Please describe.**
The arrowheads for dashed, dash-dotted and dotted arrows are ugly.

**Describe the solution you'd like**
Make all arrowheads solid, regardless of pen type.

**Describe alternatives you've considered**
Make arrowheads non-solid, but consistent between arrow lengths.

**Additional context**
I'm happy to implement this, if you agree with either of the suggested
improvements. If you could point me in the right direction of the files to
look at that would be amazing.

rolandlo:
Each stroke has a (single) line style: PLAIN, DASH, DASH_DOT or DOT. Arrows are just strokes that have been drawn by connecting 4 points in some order. To make the arrowheads solid you would probably need to create an arrow as consisting of two strokes. This would make it annoying to move arrows.
It may be quite a bit harder to implement this feature than you would have thought of.

tim-clifford:
Thanks for explaining the details @rolandlo. I did get the impression it might
be difficult. I'll chew on this for a bit and look closely at the code to see
if there's anything I can do without separating them into two strokes. I do
find the current behaviour very annoying so even if the arrowheads could just
be symmetric I think that would be a big improvement.

Febbe:
It should be doable to create an `Element` of two strokes, the question is, if our filetype supports groups of Elements.
@LittleHuba do you know more?

LittleHuba:
The current fileformat does not support this. It would be one of the major goals of the new fileformat though.

tim-clifford:
@LittleHuba what's the timeline on the new fileformat? Is there somewhere I can track the progress?
If it will be much easier to implement this with the new fileformat, I'm happy to close this issue for now and reopen it again when that's possible.

LittleHuba:
The work is currently discontinued on my side. I never got to implement the fileformat and therefore it is quite far in the future. You can leave the issue open though, as we can use it to keep track of this feature request. You can check out the progress here: https://github.com/xournalpp/xournalpp/projects/2

I do hope somebody else from the team picks development up.

Issue #3876, opened 2022-02-27
Allow selecting end points of straight lines


introt:
**Is your feature request related to a problem? Please describe.**
Rearranging mind maps with lines between nodes is pretty tedious. Currently, the easiest way is to erase the lines and redraw them after moving nodes.

**Describe the solution you'd like**
Being able to select the start and/or end points of straight lines and arrows, and having lines be redrawn when one of them is moved relative to the other.

**Additional context**
![image](https://user-images.githubusercontent.com/26069787/155884986-b33cfe2f-f9fa-4862-9db7-aa4084700683.png)

Note that the line end points aren't "holding onto" any invisible node box like in many specialized graph/mind map drawing tools; they're just floating freely.

Technius:
I think drawing diagrams is a common enough use case that it might be useful to have a dedicated "shape select" tool that is used for manipulating shape/stroke data. I've been using dedicated diagramming tools at work, and sometimes I get annoyed when it takes over 10 minutes to use the drag & drop GUI to draw a simple diagram that would've only taken 1 minute if I had access to Xournal++ pen tools...

Technius:
Actually, I would argue that a dedicated tool or at least a special selection mode is going to be required to implement this "point selection" feature anyways.

Let's assume that we actually implement point selection is performed by the rectangular selection tool. Currently, the rectangular selection tool is designed for selecting _whole_ elements (e.g., all points in a stroke). This is fundamentally at odds with point selection, which is a sort of _partial_ element selection. If we modify the rectangular select tool to select lines that are partially in the selection region, then the selection may be much more frustrating to use when trying to pick out a specific set of strokes that are surrounded by other strokes (e.g., such as when trying to select the numerator of a handwritten fraction).

We currently don't have any dedicated tools/modes for partial element selection. Adding one will open up the opportunity to implement not only implement this feature, but also features like editing splines.

/cc @rolandlo

introt:
I agree that a special selection mode is needed; the selection should be `"current rectangle selection" ⊆ "line shape end points"`, and as easily movable, copy-able, resizable and tilt-able as the current selections are (i.e. this tool is not for manipulating individual points).  

Other applications' stroke point selection tools, like Inkscape's "Edit paths by nodes" usually only allow selecting the points of whole paths, so we might be onto something novel here :)

rolandlo:
> the selection should be `"current rectangle selection" ⊆ "straight line/arrow end points"` (never partial strokes!), 

[...]

PR #4055, opened 2022-05-27
scale small arrow heads by line length

mjg:
So far, the size of arrow heads is determined purely by the line width,
which is good for medium sized and long arrows. For very small arrows,
this leads to "interesting effects", especially for double headed
arrows.

To avoid these effects, scale the arrow head by the line length for
short arrows, i.e.: make the heads no larger than 0.4 * line length.
This suites double arrows as well, and besides, 0.4 is roughly
1 - (sqrt(5) - 1) / 2 which is the golden choice.

nomin
withmin

### DIFF
--- a/src/core/control/tools/ArrowHandler.cpp
+++ b/src/core/control/tools/ArrowHandler.cpp
@@ -19,18 +19,28 @@ ArrowHandler::~ArrowHandler() = default;
 auto ArrowHandler::createShape(bool isAltDown, bool isShiftDown, bool isControlDown)
         -> std::pair<std::vector<Point>, Range> {
     Point c = snappingHandler.snap(this->currPoint, this->startPoint, isAltDown);
+    const double lineLength = std::hypot(c.x - this->startPoint.x, c.y - this->startPoint.y);
+    const double thickness = control->getToolHandler()->getThickness();
+    const double slimness = lineLength / thickness;
 
     // We've now computed the line points for the arrow
-    // so we just have to build the head
+    // so we just have to build the head:
+    // arrowDist is the distance between the line's and the arrow's tips
+    // delta is the angle between each arrow leg and the line
 
-    // set up the size of the arrowhead to be 7x the thickness of the line
-    double arrowDist = control->getToolHandler()->getThickness() * 7.0;
+    // set up the size of the arrow head to be 7x the thickness of the line (regime 1)
+    const double THICK1 = 7, LENGTH2 = 0.4;
+    double arrowDist = thickness * THICK1;
+    // but not too large compared to the line length (regime 2)
+    if (slimness < THICK1 / LENGTH2) {
+        arrowDist = lineLength * LENGTH2;
+    }
 
-    // an appropriate delta is Pi/3 radians for an arrow shape
-    double delta = M_PI / 6.0;
-    double angle = atan2(c.y - this->startPoint.y, c.x - this->startPoint.x);
+    // an appropriate opening angle 2*delta is Pi/3 radians for an arrow shape
+    const double delta = M_PI / 6.0;
+    const double angle = atan2(c.y - this->startPoint.y, c.x - this->startPoint.x);
 
-    std::pair<std::vector<Point>, Range> res;
+    std::pair<std::vector<Point>, Range> res; // members initialised below
     std::vector<Point>& shape = res.first;
 
     shape.reserve(doubleEnded ? 9 : 5);
--- a/src/core/control/tools/ArrowHandler.cpp
+++ b/src/core/control/tools/ArrowHandler.cpp
@@ -29,7 +29,7 @@ auto ArrowHandler::createShape(bool isAltDown, bool isShiftDown, bool isControlD
     // delta is the angle between each arrow leg and the line
 
     // an appropriate opening angle 2*delta is Pi/3 radians for an arrow shape
-    const double delta = M_PI / 6.0;
+    double delta = M_PI / 6.0;
     // We use different slimness regimes for proper sizing:
     const double THICK1 = 7, THICK3 = 1.6;
     const double LENGTH2 = 0.4, LENGTH4 = (doubleEnded ? 0.5 : 0.8);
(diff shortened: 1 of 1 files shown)


tattsan:
I was faced with this "too large arrowhead" problem while creating a video about geometric vectors for students. So I like this PR, thank you so much!

If there is a problem, it is when the line width is large, as in the extra thick setting of the highlighter. 

arrows-new


If the width of the arrow is too wide and the arrow is too short, as shown in the figure, it will not look like an arrow in this PR setting. I think it would be better to make the arrowhead a little larger in this type of case.

mjg:
I tend to say we should scale the linewidth fpr short arrows ... With thick short lines, what would be a good choice for the arrow head size?

Technius:
Thanks for the PR, this significantly improves the arrow usability.

> With thick short lines, what would be a good choice for the arrow head size?

I think changing the length of the arrow head won't solve the problem. If we make the arrow head shorter, then the thickness of the stem will completely overshadow the arrow head (middle arrow in @tattsan's example). In contrast, if we make the arrow head longer, then we end up with the original situation where the arrow head is way too large so it ends up looking like a triangle (imagine the last arrow in @tattsan's example).

Perhaps also changing the angle (only on large thickness) might be an improvement? It might be a little too complicated though.
<details>
<summary>Quick mockup in gimp:</summary>

</details>

bhennion:
The PR already greatly improves on the current behaviour, with minimal changes to the code, so that's great! I'm not sure we can find a good and simple solution for very thick highlighter strokes, but that's alright for now, I think.


To put in my two pennies worth: in the long run, I think a proper solution to ugly arrowheads wouold be to add stroke tips to the file format. Then, the user could chose, like, e.g. in Libreoffice

Screenshot_20220529_115659

This would allow to have clean arrow heads or tails (for instance, with a pointy end, not a round one) and solve the highlighter issue: something like this (without the overlapping artefact between the line and the head)
Screenshot_20220529_121336

 It'd also allow for curved arrows (simply any stroke, with an arrowhead added), fix #2808, make #3876 possible for arrows (see [this post](https://github.com/xournalpp/xournalpp/issues/3876#issuecomment-1055825021)), allow for proper arrow resizing, and so on.
All this, of course, goes way beyond the scope of this PR.

Technius:
To summarize, I think we all agree that the way to go with this PR is 1) scaling by line length and 2) adjusting the angle on thicker lines (e.g., with @tattsan's patch).

bhennion (commented):
@mjg Do you want to add @tattsan's angle idea to this PR?
Otherwise, it's already much better than the current behaviour of master and I guess it could me merged.

mjg:
> @mjg Do you want to add @tattsan's angle idea to this PR? Otherwise, it's already much better than the current behaviour of master and I guess it could me merged.

Sorry, I missed the agreement. By force-pushing rebases I did not mean to ignore it ;-)

Yes, I can fold in the angle adjustment, or (better) put that on top. I'm more of a commit-early-commit-often-rebase-and-merge-gitter than a squasher ...

mjg:
Intermediate report: I wrapped my head around @tattsan 's patch now because the multiple "divide by fraction" and interaction between the cases and parameters triggered me to do so, and also because arrow shapes should scale properly if we scale both linewidth (thickness) and linelength by the same factor. I am rewriting everything in terms of the shape ratio linelength/thickness ("slimness").
My patch had two slimness regimes: scale the headsize ("arrowDist" in the code) linearly by thickness for slim arrows, linearly by linelength for fat arrows.
@tattsan introduced two more thresholds separating my "fat regime" into three regions where we alternate between thickness and linelength as the scale factor (with different scaling constants). Granted, one of them is only almost linear, as much as sin is between pi/6 and pi/3.
I'm not convinced those additional alternations are necessary - basically, the arrow wings should still protrude beyond the thick baseline, which is why we scale by thickness for fat arrows. We can help that effect by adjusting the angle.
Once the line is shorter than it is thick, there is not much we can do.

So, I'll spend more time on this than planned, hoping for better readibility. Codwise, spelling out the nested max/min constructs as ifs should not impact with a good compiler. max/min are just that anyways.

mjg:
Feel free to try out without the last commit and to adjust the thresholds between the slimness regimes :-)

tattsan:
I drawed the same arrows as in https://github.com/xournalpp/xournalpp/pull/4055#issuecomment-1140396860 with the new patch.

arrows-new3

mjg:
Thanks @tattsan . That's great for comparison, for checking whether it meets your needs and, potentially, for adjusting the parameters further.

src/core/control/tools/ArrowHandler.cpp
```diff
@@ -19,15 +19,38 @@ ArrowHandler::~ArrowHandler() = default;
 auto ArrowHandler::createShape(bool isAltDown, bool isShiftDown, bool isControlDown)
         -> std::pair<std::vector<Point>, Range> {
     Point c = snappingHandler.snap(this->currPoint, this->startPoint, isAltDown);
+    double lineLength = sqrt(pow(c.x - this->startPoint.x, 2) + pow(c.y - this->startPoint.y, 2));
```
bhennion, line 22:
```suggestion
    double lineLength = std::hypot(c.x - this->startPoint.x, c.y - this->startPoint.y);
```

src/core/control/tools/ArrowHandler.cpp
```diff
@@ -19,15 +19,38 @@ ArrowHandler::~ArrowHandler() = default;
...
+    // so we just have to build the head:
+    // arrowDist is the distance between the line's and the arrow's tips
+    // delta is the angle between each arrow leg and the line
 
     // an appropriate delta is Pi/3 radians for an arrow shape
     double delta = M_PI / 6.0;
+    // set up the size of the arrowhead to be 7x the thickness of the line
+    double arrowDist = thickness * 7;
+    if (slimness >= 7 / 0.4) {
+        // arrow is not too long compared to the line length
```
bhennion, line 36:
```suggestion
        // arrow head is not too long compared to the line length
```

src/core/control/tools/ArrowHandler.cpp
```diff
@@ -19,15 +19,38 @@ ArrowHandler::~ArrowHandler() = default;
...
+    } else if (slimness >= 1.6 / 0.4) {
+        // arrow is not too short compared to the thickness
+        arrowDist = lineLength * 0.4;
+    } else if (slimness >= 1.6 / (doubleEnded ? 0.5 : 0.8)) {
+        // arrow is not too thick compared to the line length
+        arrowDist = thickness * 1.6;
+        // help visibility by widening the angle
+        delta = (1 + (slimness - 1.6 / 0.4) / (1.6 / (doubleEnded ? 0.5 : 0.8) - 1.6 / 0.4)) *  M_PI / 6.0;
+	// which allows to shorten the tips and keep the horizonzal distance
+	arrowDist *= sin(M_PI / 6.0) / sin(delta);
```
bhennion, line 46:
```suggestion
        // which allows to shorten the tips and keep the horizonzal distance
        arrowDist *= sin(M_PI / 6.0) / sin(delta);
```
Something's wrong with our clang-format test unit... It should have caught this.
Could you apply clang-format to your changes?

bhennion (changes requested):
Thanks for the changes!

There is a clang-format issue (we need to fix the CI apparently).

Also, I guess the constants should be named, so that one understands what 7, or 7 / 0.4 corresponds to.

mjg:
I also commented on a false clang warning. We could use `#pragma` here (3 of them), but that seemed like overengineering things.

bhennion (approved):
Oh... I just realized I had left my review pending here. Sorry.

Just one small thing (see below) and then we'll be good to go.

mjg:
Thanks for catching the last concrete number ;)
Rebase error fixed.
I wasn't sure about the "resolve conversation" button, so I just re-pushed things.

bhennion:
Thanks a lot for the nice work!

Squashing and merging in 24 hours.

Issue #3694, opened 2022-01-03
spec versioning does not follow Fedora guidelines


mjg:
**Affects versions :**
 - OS: Fedora Linux
 - Version of Xournal++: that is the question ;) 
 
**Describe the bug**
`1a070ffd4 ("Update xournalpp.spec", 2021-12-27)` and the preceding commit changed the versioning of the `spec` file for post releases, referring to the Fedora packaging guidelines and suggesting (in the comment) what to do for pre and post releases. Unfortunately, the versioning scheme does not appear to follow the guidelines, and neither do the comments.

Fedora has two possible schemes:
- the current one
- the "traditional one" (basically for RHEL 7 and other systems with old rpms)

The versioning in `xournalpp.spec` is a mix of both: the caret notation for post release is supported in the current scheme only; on the other hand, in that scheme the `Release` rpm tag ist strictly for the package release (and would have to be `Release: 1%{?dist}` here). But the spec puts versioning info there (as you would in the traditional scheme) and suggests a prerelease package version 0 (as in the traditional scheme, but not the current scheme).

I would submit a PR, but I can't tell which scheme the spec wants to follow. (Possibly the "current scheme", and possibly I'll a submit a PR for that anyways...)

LittleHuba:
Thanks for the heads up.
Please also check the script for updating the version so we do not revert your fix by accident later on.
Feel free to submit a PR for the "current scheme"

mjg:
Sorry for the late response. It took me a while until I realised that `scripts/release_helper.sh` is more recent in branch release-1.1 than in master, and that in the former it indeed rewrites the spec file ...

Also, I don't want to stomp on @luyatshimbalanga 's feet. Luya, is your main focus with the in-tree spec to have a spec which builds binary or source rpms locally (rather than feed copr's rpkg or such), and stays otherwise close to the spec in Fedora's rawhide dist-git?

luyatshimbalanga:
@mjg Yes that is the focus, feel free to improve the spec file.

mjg:
Me, somewhat confused again by the branch structure:
```
git log --oneline --left-right --cherry-mark  upstream/master...upstream/release-1.1 -- scripts/release_helper.sh rpm/fedora/xournalpp.spec
> c1ae6c312 Fix wrong version suffix in relevant files
> db7a16eec Update release helper script to set correct versions for Fedora and MacOS
> be1dab569 Update version info in Fedora build recipe
[...]

PR #4056, opened 2022-05-27
Rpkg

mjg:
So here is the rebased rpkg branch related to #3694. The spec is almost completely due to @luyatshimbalanga , with adaptions to use rpkg instead of lua for computing the version etc. Note:

- This gives a valid spec file only after processing with `rpkg` (which is the intended use case for automatic builds on copr).
- The computed versions sort earlier than the ones created so far - users of the copr repo need to "downgrade" to the new versions. So, either decide that this is acceptable for "devel users" (it should be) or hold this merge back until the next release.

### DIFF
--- a/rpm/fedora/xournalpp.spec
+++ b/rpm/fedora/xournalpp.spec
@@ -127,7 +127,7 @@ appstream-util validate-relax --nonet %{buildroot}%{_metainfodir}/com.github.%{n
 * Sun Mar 6 2022 Luya Tshimbalanga <luya@fedoraproject.org>
 - Port enhanced spec file from Michael J Gruber version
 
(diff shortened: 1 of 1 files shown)


Technius:
Thanks for the PR.

We currently don't have a communication channel for the development/master branch (although we should...maybe a mailing list or RSS feed?), so I think we should hold this back until the next release.

rolandlo:
@Technius Should this PR be merged after (or before) the 1.2 release?

Technius:
Now that 1.2 is released, I think we can revisit this PR now.

Technius (approved):
Thanks! Merging in 24 hours if there are no objections.

PR #4065, opened 2022-05-31
Change float format string to use g specifier

Technius:
This commit changes the float precision format string to "%.8g" to save space
when serializing.

The float precision format string is defined as "%.8f", which always prints
floats with 8 trailing digits (padding with zeros if needed). However, in many
cases, fewer than 8 trailing digits are needed. This means that by using the
"%.8g" specifier, which does not print trailing digits and may switch to
scientific notation to be shorter, the size of the serialized stroke strings may
be reduced.

As an experiment, I compared two files consisting solely of stroke data:

* File A: a 5-page, 1.4 MB document with worked out math problems.
* File B: a 16-page, 3.3 MB document full of mathematical proofs.

On File A, the total size of the stroke string data was reduced by 23.64% with
no change in the float values; this lead to a total file size reduction of
16.24% (uncompressed) and 6.40% (compressed).

On File B, the total size of the stroke string data was reduced by 23.11% with
negligible change in the float values (less than 1e-5 % relative difference);
this lead to a total file size reduction of 15.48% (uncompressed) and
18.99% (compressed).


Technius:
If you'd like to compare your own files, you can use the script that I used:

<details>
<summary>Python script to compare f vs g specifier</summary>

To use, copy the script into `compare.py` and then run `compare.py <file saved with master branch> <file saved with this PR>`

```python
#!/usr/bin/env python3

from dataclasses import dataclass
from types import SimpleNamespace
from typing import List
import xml.etree.ElementTree as ETree
import sys
import gzip
from pathlib import Path

def get_stats(p: Path):
    strokes = []
    with gzip.open(p) as f:
        fcontents = f.read()
        root = ETree.fromstring(fcontents)
        for stroke in root.iterfind('**/stroke'):
            if not stroke.text:
                continue
            contents: str = stroke.text
            points = contents.split(' ')
            strokes.append(SimpleNamespace(
                strlen=len(contents),
                points=[float(s) for s in points],
            ))
    return SimpleNamespace(
        strokes=strokes,
        text=fcontents,
        fsize=p.stat().st_size
    )

slf = get_stats(Path(sys.argv[1]))
slg = get_stats(Path(sys.argv[2]))

assert len(slf.strokes) == len(slg.strokes)

@dataclass
class FStats:
    tot_strlen: int = 0

def pcnt_diff(old, new):
    return 100 * (new - old) / old

total_strokes = len(slf.strokes)
total_points = 0
fstats = FStats()
gstats = FStats()
min_diff = float('inf')
max_diff = float('-inf')
diff_list = []
for sf, sg in zip(slf.strokes, slg.strokes):
    total_points += len(sf.points) - len(sg.points)
    fstats.tot_strlen += sf.strlen
    gstats.tot_strlen += sg.strlen
    assert len(sf.points) == len(sg.points)
    for p1, p2 in zip(sf.points, sg.points):
        d = pcnt_diff(p2, p1)
        diff_list.append(d)
        min_diff = min(min_diff, d)
        max_diff = max(max_diff, d)

print('F stroke strlen: {}'.format(fstats.tot_strlen))
print('G stroke strlen: {}'.format(gstats.tot_strlen))
print('stroke strlen % reduction: {:.2f}%'.format(
    -pcnt_diff(fstats.tot_strlen, gstats.tot_strlen)))
print('total % uncompressed file size reduction: {:.2f}%'.format(-pcnt_diff(len(slf.text), len(slg.text))))
print('total % compressed file size reduction: {:.2f}%'.format(-pcnt_diff(slf.fsize, slg.fsize)))
print()
print('min diff: {:.8f}%'.format(min_diff))
print('max diff: {:.8f}%'.format(max_diff))
diff_list.sort()
print('median diff: {:.8f}%'.format(diff_list[len(diff_list) // 2]))

```

</details>

Technius:
The test failure is due to `LoadHandlerTest.cpp::testSaveLoad()` making an exact comparison between the stroke values (it's nice that we have a unit test to catch file format changes!). But this is a false positive; with my script above I got the following results:
```plain
F stroke strlen: 31242
G stroke strlen: 24126
stroke strlen % reduction: 22.78%
total % uncompressed file size reduction: 18.23%
total % compressed file size reduction: 15.04%

min diff: -0.00000458%
max diff: 0.00000437%
median diff: 0.00000000%
```

bhennion (approved):

rolandlo:
> > are files written with `%.8g` readable by the master branch?
> 
> Yes. The float values are parsed using `g_ascii_strtod` which supports values like `1.234` and `-59.28e-20`, so older versions are forward compatible with this change.

What about backward compatibility with Xournal (.xoj-file export in compatibility mode)? Can these files still be read by Xournal?

Technius:
I can't get a xopp file on master to open in Xournal, so I don't know. I think Xournal also uses `g_ascii_strtod`, so it's likely that  there isn't a problem with loading floats printed with the g specifier.

rolandlo:
> I can't get a xopp file on master to open in Xournal, so I don't know. I think Xournal also uses `g_ascii_strtod`, so it's likely that there isn't a problem with loading floats printed with the g specifier.

As far as I can see, the .xoj-export will also use the new number format. It's fine, though. I checked both a modified `.xoj`-file opened in Xournal and the source code of Xournal (`xo-file.c`).

What about the failing tests? 
```The following tests FAILED:
	 15 - ControlLoadHandler.testLoadStoreLoadDefault (Failed)
	 16 - ControlLoadHandler.testLoadStoreLoadGerman (Failed)
```

Technius:
> What about the failing tests?

Either 1) we replace the tests; or 2) we relax the test code so it checks relative difference below some threshold instead of checking exact equality.

Maybe an even better way is to treat the existing tests like backwards compatibility tests and then add new tests that have already been saved in the new representation.

Technius:
Rebased and updated the tests.

rolandlo (approved):
Minor nitpick:
`testLoadStoreLoadHelper` is not only applied to the `suite.xopp` file, so the comment above this function and in the function body are slightly misleading.

Other than that LGTM.

Technius:
Thanks, I clarified the description a little bit and removed a comment that I forgot to remove.

PR #4080, opened 2022-06-08
Prevent fractional translation of mask surface

lennonhill:
Discussed here: https://github.com/xournalpp/xournalpp/discussions/4072


bhennion:
Thanks a lot @lennonhill!

> Previously finished strokes would change their appearance shortly after they were drawn.

I cannot observe this change in appearance, I assume it is mostly visible on (relatively) low DPI displays.

In any case, this integral offset seems more resonable, and is probably cheaper (avoids antialiasing when blitting the mask).

I'll be merging this in 48 hours unless an objection is raised.

PR #4081, opened 2022-06-09
View Refactor 4 - Code cleanup & VerticalToolHandler

bhennion:
This is a code cleanup PR in the View Refactor series :)

There are non-trivial changes in VerticalToolHandler that limit the unnecessary rerenderings (see 4bef970d).

@Technius: I'm sorry I've been asking you a lot of reviews lately, but I think you're the most qualified for the VerticalToolHandler business in this PR. :-)

### DIFF
--- a/src/core/control/tools/VerticalToolHandler.cpp
+++ b/src/core/control/tools/VerticalToolHandler.cpp
@@ -8,6 +8,7 @@
 #include "control/zoom/ZoomControl.h"
 #include "model/Layer.h"
 #include "undo/UndoRedoHandler.h"
+#include "util/LoopUtil.h"
 #include "view/DebugShowRepaintBounds.h"
 #include "view/SelectionView.h"
 
@@ -29,7 +30,9 @@ VerticalToolHandler::VerticalToolHandler(Redrawable* view, const PageRef& page, 
 
     this->adoptElements(this->spacingSide);
 
-    view->rerenderPage();
+    if (auto rect = this->getElementsBoundingRect()) {
+        view->rerenderRect(rect->x, rect->y, rect->width, rect->height);
+    }
 }
 
 VerticalToolHandler::~VerticalToolHandler() {
@@ -114,7 +117,7 @@ void VerticalToolHandler::paint(cairo_t* cr) {
                         cairo_image_surface_get_height(this->crBuffer));
         cairo_set_source_rgba(cr, 1.0, 0, 0, 0.3);
         cairo_fill(cr);
-    })
+    });
 }
 
 void VerticalToolHandler::currentPos(double x, double y) {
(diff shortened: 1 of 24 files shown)


src/core/control/tools/VerticalToolHandler.cpp
```diff
@@ -155,27 +161,46 @@ bool VerticalToolHandler::onKeyReleaseEvent(GdkEventKey* event) {
...
+        this->view->repaintRect(0, min, this->page->getWidth(), max - min);
+        return nullptr;
+    }
     double dY = this->endY - this->startY;
 
     auto undo =
             std::make_unique<MoveUndoAction>(this->layer, this->page, &this->elements, 0, dY, this->layer, this->page);
 
+    std::optional<xoj::util::Rectangle<double>> rect = this->getElementsBoundingRect();
+    assert(rect);
```
Technius, line 187: nit: `g_assert` gives a better error message.

Also, if you move up the rect line right after the `this->elements->empty()` block, I think it will be clearer that rect should not be the nullopt.

src/core/control/tools/VerticalToolHandler.cpp
```diff
@@ -28,7 +30,10 @@ VerticalToolHandler::VerticalToolHandler(Redrawable* view, const PageRef& page,
 
     this->adoptElements(this->spacingSide);
 
-    view->rerenderPage();
+    auto rect = this->getElementsBoundingRect();
+    if (rect) {
```
Technius, line 34: nitpick:
```suggestion
    if (auto rect = this->getElementsBoundingRect()) {
```

src/core/control/tools/VerticalToolHandler.cpp
```diff
@@ -108,12 +113,12 @@ void VerticalToolHandler::paint(cairo_t* cr) {
...
-                    cairo_image_surface_get_height(this->crBuffer));
-    cairo_set_source_rgba(cr, 1.0, 0, 0, 0.3);
-    cairo_fill(cr);
-#endif
+    IF_DEBUG_REPAINT({
+        cairo_rectangle(cr, 0, elemY, cairo_image_surface_get_width(this->crBuffer),
+                        cairo_image_surface_get_height(this->crBuffer));
+        cairo_set_source_rgba(cr, 1.0, 0, 0, 0.3);
+        cairo_fill(cr);
+    })
```
Technius, line 121: nit: add a semicolon to avoid weird editor line breaking behavior (e.g., to pretend this is a statement).

src/core/control/tools/VerticalToolHandler.cpp
```diff
@@ -155,27 +161,46 @@ bool VerticalToolHandler::onKeyReleaseEvent(GdkEventKey* event) {
...
+
     for (Element* e: this->elements) {
         e->move(0, dY);
 
         this->layer->addElement(e);
     }
+    this->elements.clear();
 
-    view->rerenderPage();
+    view->rerenderRect(rect->x, rect->y + dY, rect->width, rect->height);
```
Technius, line 196: I think this could be further improved by only rerendering around the element bboxes, but that will have to be done in a future PR (along with convincing benchmarks/profiling), as I don't think we have such a fine-grained rerendering method on `XojPageView` yet.

Technius (approved):
LGTM, some minor style nitpicks.

The first time I wrote this, I think I didn't realize that the tool handlers are drawn in repaint, so thanks for noticing the inefficiency here.

bhennion:
Thanks a lot for the review, @Technius! I applied your suggestions and fixed a small bug in the debug output (visible when compiling with `-DDEBUG_SHOW_REPAINT_BOUNDS`).

Merging in 48 hours, unless an objection is raised.

PR #4082, opened 2022-06-09
Add RAII wrappers for cairo classes in gui/PageView and jobs/RenderJob

### LINKED ISSUE #3903: Stroke delete eraser takes a long time to erase with 200% display scaling
- OS: Arch Linux
 - Desktop environment: GNOME 41, wayland or X11
 - libgtk version 3.24.33
 - Xournal++ version 1.1.1 (commit 4f113e94)
 - Installation method: Arch repo (pacman)

When using 200% display scaling, the "delete stroke" eraser takes several seconds to delete strokes. If you want to erase a line of text for example, you would run the eraser across that line of text, wait a couple of seconds, and then it will all erase at the same time. This happens with both wayland and X11. Setting the display scaling to 100% in GNOME settings fixes this issue; with 100% scaling, strokes are deleted immediately when touched with the eraser.

1. Set display scaling to 200% in GNOME settings
2. Open Xournal++
3. Open or create a document with at least a few lines of handwriting
4. Select the "delete stroke" eraser
5. Try to erase a line of text with the eraser tool. It will take a second or more for the line to be deleted

Expected behavior: Strokes should be deleted as soon as the eraser touches them, rather than a few seconds after. This is how it works with 100% display scaling.

Video of the issue: https://youtu.be/KIoZmCGSCxI

If there is anything I can do to make troubleshooting this easier, let me know

bhennion:
This PR creates RAII wrappers for cairo_t and cairo_surface_t (as well as a template for further applications -- e.g. GObject...).

Those wrappers are then used in gui/PageView and to refactor jobs/RenderJob.

One major change with this PR concerns HiDPI settings: jobs/RenderJob is now able to rerender only a portion of the page (it was only the case for GDK_SCALE=1 before this PR).
This should drastically improve performances on HiDPI displays and fix #3903 and may help with other performance related issues as well.

**Note:** ~~This PR is based on #4081. Only the last 2 commits actually belong to this PR.~~

- [x] Merge #4081 and rebase

### DIFF
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -7,16 +7,17 @@
 
 #include <cairo.h>  // for cairo_create, cairo_destroy, cairo_...
 
-#include "control/Control.h"      // for Control
-#include "control/ToolEnums.h"    // for TOOL_PLAY_OBJECT
-#include "control/ToolHandler.h"  // for ToolHandler
-#include "control/jobs/Job.h"     // for JOB_TYPE_RENDER, JobType
-#include "gui/PageView.h"         // for XojPageView
-#include "gui/XournalView.h"      // for XournalView
-#include "model/Document.h"       // for Document
-#include "util/Rectangle.h"       // for Rectangle
-#include "util/Util.h"            // for execInUiThread
-#include "view/DocumentView.h"    // for DocumentView
+#include "control/Control.h"          // for Control
+#include "control/ToolEnums.h"        // for TOOL_PLAY_OBJECT
+#include "control/ToolHandler.h"      // for ToolHandler
+#include "control/jobs/Job.h"         // for JOB_TYPE_RENDER, JobType
+#include "gui/PageView.h"             // for XojPageView
+#include "gui/XournalView.h"          // for XournalView
+#include "model/Document.h"           // for Document
+#include "util/Rectangle.h"           // for Rectangle
+#include "util/Util.h"                // for execInUiThread
+#include "util/raii/CairoWrappers.h"  // for CairoSurfaceSPtr, CairoSPtr
+#include "view/DocumentView.h"        // for DocumentView
 
 using xoj::util::Rectangle;
 
@@ -25,54 +26,34 @@ RenderJob::RenderJob(XojPageView* view): view(view) {}
 auto RenderJob::getSource() -> void* { return this->view; }
 
 void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
-    double zoom = view->xournal->getZoom();
-    Document* doc = view->xournal->getDocument();
+    const double ratio = view->xournal->getZoom() * this->view->xournal->getDpiScaleFactor();
 
     /**
-     * Make sure the mask is big enough
-     * The +1 covers examples like rect.x = 0.4, rect.width = 1 and zoom = 1
-     * We need a mask of width 2 pixels for that...
+     * The +1 makes sure the mask is big enough
+     * For example, if rect.x = m + 0.9, rect.width = n + 0.2 and ratio = 1 and m and n are integers
+     * We need a mask of width n+2 pixels for that...
      **/
-    auto x = int(std::floor(rect.x * zoom));
-    auto y = int(std::floor(rect.y * zoom));
-    auto width = int(std::ceil(rect.width * zoom)) + 1;
-    auto height = int(std::ceil(rect.height * zoom)) + 1;
+    const auto x = std::floor(rect.x * ratio);
+    const auto y = std::floor(rect.y * ratio);
+    const auto width = int(std::ceil(rect.width * ratio)) + 1;
+    const auto height = int(std::ceil(rect.height * ratio)) + 1;
 
-    cairo_surface_t* rectBuffer = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
-    cairo_t* crRect = cairo_create(rectBuffer);
-    cairo_translate(crRect, -x, -y);
-    cairo_scale(crRect, zoom, zoom);
+    xoj::util::CairoSurfaceSPtr rectBuffer(cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height));
+    cairo_surface_set_device_offset(rectBuffer.get(), -x, -y);
+    cairo_surface_set_device_scale(rectBuffer.get(), ratio, ratio);
 
-    DocumentView v;
-    Control* control = view->getXournal()->getControl();
-    v.setMarkAudioStroke(control->getToolHandler()->getToolType() == TOOL_PLAY_OBJECT);
-    v.setPdfCache(view->xournal->getCache());
+    renderToBuffer(rectBuffer.get());
 
-    doc->lock();
-    v.drawPage(view->page, crRect, false);
-    doc->unlock();
+    std::lock_guard lock(this->view->drawingMutex);
+    xoj::util::CairoSPtr crPageBuffer(cairo_create(view->crBuffer.get()));
 
-    cairo_destroy(crRect);
-
-    view->drawingMutex.lock();
-
-    cairo_t* crPageBuffer = cairo_create(view->crBuffer);
-
-    cairo_set_operator(crPageBuffer, CAIRO_OPERATOR_SOURCE);
-    cairo_set_source_surface(crPageBuffer, rectBuffer, x, y);
-    cairo_rectangle(crPageBuffer, x, y, width, height);
-    cairo_fill(crPageBuffer);
-
-    cairo_destroy(crPageBuffer);
-
-    cairo_surface_destroy(rectBuffer);
-
-    view->drawingMutex.unlock();
+    cairo_set_operator(crPageBuffer.get(), CAIRO_OPERATOR_SOURCE);
+    cairo_set_source_surface(crPageBuffer.get(), rectBuffer.get(), 0, 0);
+    cairo_rectangle(crPageBuffer.get(), rect.x, rect.y, rect.width, rect.height);
+    cairo_fill(crPageBuffer.get());
 }
 
 void RenderJob::run() {
-    double zoom = this->view->xournal->getZoom();
-
     this->view->repaintRectMutex.lock();
 
     bool rerenderComplete = this->view->rerenderComplete;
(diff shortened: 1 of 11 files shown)


LittleHuba:
Nice! This is amazing 😍
I'll reserve some time this weekend to give this a review. If I understood you right, this could even remove a big blocker for infinite pages.

bhennion:
> If I understood you right, this could even remove a big blocker for infinite pages.

Yes, it's definitely a step towards infinite pages.

> I'll reserve some time this weekend to give this a review

Thanks! Keep in mind that only the last two commits are relevent. So review commit by commit, at least until #4081 is merged.

test/unit_tests/util/RAIIWrappersTest.cpp
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    static int nbData;
+};
+int TestData::nbData = 0;
+};  // namespace
+
+using TestRAII = xoj::util::RAII_Wrapper<TestData, TestData::ref, TestData::unref>;
+
+TEST(UtilsRAII, testWrappers) {
```
LittleHuba, line 33: Please test only one thing per test. And then give the test a very descriptive name, please.

LittleHuba (commented):
Nice work! Just some minor nitpicks.

src/util/include/util/RAII_Wrappers.h
```diff
@@ -0,0 +1,147 @@
...
+    }
+    RAII_Wrapper& operator=(RAII_Wrapper&& other) {
+        if (this == &other) {
+            return *this;
+        }
+        if (p) {
+            unref(p);
+        }
+        p = other.p;
+        other.p = nullptr;
```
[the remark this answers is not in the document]
Febbe: Better than swap semantic is std::exchange

src/core/control/jobs/RenderJob.h
```diff
@@ -45,6 +46,8 @@ class RenderJob: public Job {
 
     void rerenderRectangle(xoj::util::Rectangle<double> const& rect);
 
+    void renderToBuffer(xoj::util::CairoSurfaceWrapper& buffer) const;
+
```
Febbe, line 50: Pass smart_ptrs by value or its content by (const)-reference. In this case, I would pass the `cairo_surface_t*` itself.

Febbe (changes requested):

src/util/include/util/RAII_Wrappers.h
```diff
@@ -0,0 +1,147 @@
...
+ * @brief Simple template class for RAII wrappers of Cairo/GTK classes
+ * @param T The wrapper will store a pointer of type T
+ * @param ref A function that increases the ref-count
+ * @param unref A function that decreases the ref-count
+ *
+ * WARNING: For GTK classes that often come as a floating ref.
+ *      e.g. https://docs.gtk.org/gobject/class.Object.html
+ *    call ref_sink before passing it to RAII_Wrapper(T*) or reset(T*).
+ */
+template <typename T, T* ref(T*), void unref(T*)>
```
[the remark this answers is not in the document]
bhennion: I changed to template<typename T, class H> where H contains all the required functions (ref, unref and optionally ref_sink)

src/util/include/util/RAII_Wrappers.h
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Assumes ownership is already established
+     *
+     * WARNING: This is not fit for GTK classes that come as a floating ref.
+     *      e.g. https://docs.gtk.org/gobject/class.Object.html
+     *    For such a pointer, call ref_sink before passing it to RAII_Wrapper(T*) or reset(T*).
+     */
+    RAII_Wrapper(T* p): p(p) {}
```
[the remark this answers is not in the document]
bhennion: I used this idea. I am not sure it is actually safer, unless we remove the default `Adopt = adopt`. Maybe the default should depend on the existence of H::ref_sink?

src/core/control/jobs/RenderJob.h
bhennion: What the rational behind this? If ref/unref can be costful, then passing them by reference seems a sound idea.

Febbe: The rational behind this is, that the function actually does nothing with the smart_ptr. It does something with the object itself. Therefore it is cleaner to pass the object* instead of the smart ptr. Only if the function want to modify the smart_ptr itself, it is required to pass it.

src/util/include/util/RAII_Wrappers.h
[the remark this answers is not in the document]
Febbe: Actually a good idea.

src/util/include/util/RAII_Wrappers.h
```diff
@@ -0,0 +1,77 @@
...
+
+class CairoSurfaceHandler {
+public:
+    constexpr static auto ref = cairo_surface_reference;
+    constexpr static auto unref = cairo_surface_destroy;
+};
+
+template <class object_type>
+class GInitiallyUnownedHandler {
+    static_assert(std::is_base_of<GInitiallyUnowned, object_type>::value, "Class must inherit GInitiallyUnowned");
```
Febbe, line 40: I don't think this will work, because C inheritance is self implemented and cannot be determined by the compiler.

src/util/include/util/RAII_Wrappers.h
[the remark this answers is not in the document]
Febbe: In most cases, the programmer want to reference the pointer. 
If a widget (GInitiallyUnowned) is created, it's often passed to the future parent after the initialization. After that, the pointer has to be referenced anyway. 

So probably I would make Ref the default one.
Making it dependent may be counterintuitive, then we have a container with the same name but different behaviour. 
Smells like std::vector<bool>.

Also it would be cool, if you could add constexpr and noexcept were ever possible.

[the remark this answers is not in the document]
bhennion: > In most cases, the programmer want to reference the pointer.

This is just not true for Cairo objects (or even Poppler objects), for which the user almost never wants to add a reference.

Also, even for GWidget's, it is not clear that you would pass the widget first to its parent, and then to the smart pointer. If the user first creates the smart pointer, initializes a couple of parameters of the widget, and then adds it to the parent, then we need to ref_sink.
All in all, is the ref has already been sunk (by the parent, or anyone else), ref_sink is equivalent to ref and is still safe. I do not see any case where defaulting to ref could be better than defaulting to ref_sink (if ref_sink exists).

I think the bottom line is: we (I) want `SmartPointer p(create_object(...));` to be leak free: i.e. ref_sink for GObjects, and adopt for cairo/poppler things.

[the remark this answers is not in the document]
Febbe: Mhh, and when we add a type to H:

`using Default = Adopt` for example?
Then the constructor and the reset method must be templated against.

src/util/include/util/raii/CLibrariesSPtr.h
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    constexpr static struct Adopt {
+    } adopt = Adopt();
+    constexpr static struct Ref {
+    } ref = Ref();
+    constexpr static struct RefSink {
+    } refsink = RefSink();
+
+    CLibrariesSPtr(T* p, Adopt = adopt): p(H::adopt(p)) {}
```
bhennion, line 70: I think I found an acceptable solution: I added to the handler class H a function T* adopt(T*) which:
- For cairo objects: simply returns the input pointer
- For GObjects: tests if the ref is floating and sinks if necessary.

This makes both `GObjectSPtr p(g_object_new(...))` and `WidgetSPtr p(gtk_create_some_widget(....))` safe (note that g_object_new does not return a floating ref, while gtk's widget creating functions typically do -- with the notable exception of our own gtk_xournal_new).

Febbe: The thing is, that this check should not be necessary, because it is well documented, in which state the gobject is. I would only do an assert for debug builds.

bhennion: `g_object_take_ref` this is what we want :-) https://docs.gtk.org/gobject/method.Object.take_ref.html

bhennion: Except the linux CI does not know it exists... :-(

bhennion: > Except the linux CI does not know it exists... :-(

`g_object_take_ref` was introduced in glib 2.70. Ubuntu 18.04 LTS has glib 2.26 and ubuntu 20.04 has glib 2.64... So we can't rely on it. To bad!

bhennion: I went for using `g_object_take_ref` as soon as the glib version is high enough, with a fallback to the previous handmade "if floating sink, otherwise do nothing".

In the case of outdated glib2's, I would not replace the test by an assert for the following reason:
- this is the only way to actually get the exact same outcome as a call to `g_object_take_ref`
- us devs typically have an up-to-date library, and we would not see the assertion failures
- it offers more flexibility and safety

src/core/control/jobs/RenderJob.cpp
```diff
@@ -25,54 +26,34 @@ RenderJob::RenderJob(XojPageView* view): view(view) {}
...
      * We need a mask of width 2 pixels for that...
      **/
-    auto x = int(std::floor(rect.x * zoom));
-    auto y = int(std::floor(rect.y * zoom));
-    auto width = int(std::ceil(rect.width * zoom)) + 1;
-    auto height = int(std::ceil(rect.height * zoom)) + 1;
+    const auto x = int(std::floor(rect.x * ratio));
+    const auto y = int(std::floor(rect.y * ratio));
+    const auto width = int(std::ceil(rect.width * ratio)) + 1;
+    const auto height = int(std::ceil(rect.height * ratio)) + 1;
```
Febbe, line 39: 1st. Too model, that we require at least a mask of two would be done in regular with a std::max(2, value).
2nd `cairo_surface_set_device_offset` only accepts doubles, why do we cast this here first into na int and then back to an double?

src/core/control/jobs/RenderJob.h
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <cairo/cairo.h>  // for cairo_surface_t
```
Febbe, line 14: Is it not possible to forward declare this?

src/util/include/util/raii/CairoWrappers.h
```diff
@@ -0,0 +1,65 @@
...
+#include <cairo.h>
+
+#include "CLibrariesSPtr.h"
+
+namespace xoj::util {
+
+inline namespace raii {
+namespace specialization {
+
+template <typename T>  // C++20: replace with std:identity()
```
Febbe, line 25:
```suggestion
template <typename T>  // Todo (cpp20): replace with std::identity()
```

Febbe (approved):
I think there are only nitpicks left.

src/core/control/jobs/RenderJob.h
bhennion: Except cairo_surface_t is typedef'd in cairo.h. I don't know how to forward declare a typedef.

src/core/control/jobs/RenderJob.cpp
bhennion: It's not about size 2, this is just a dumb example to explain the +1. The same issue appears with anything like rect.x = n + 0.9 and rect.width = m + 0.2 (with m and n integers). In this situation, we need a width of m+2.
I'll make the comment more precise.

About the casting: you're right, it's not necessary for x and y. Only the fact that they are integer values matters, to avoid unnecessary antialiasing.

bhennion:
Incorporated @Febbe's nitpicks and cleaned up the commit tree.

Merging in 48 hours unless someone objects

PR #4088, opened 2022-06-13
Minor Bugfixes:

Febbe:
- Fixed memory leak
- Fixed Crash with utf8 paths on windows


bhennion (commented):
LGTM! It needs testing on diffrerent platforms before merging, just to make sure.
Could a test unit be added to do that (parse a test .tex file with a utf8 file name)?

Febbe:
> LGTM! It needs testing on diffrerent platforms before merging, just to make sure. Could a test unit be added to do that (parse a test .tex file with a utf8 file name)?

Utf8 files on Windows never worked, they may cause a crash. This just adds an exception handler in this case.

Issue #2625, opened 2021-01-06
Line Style for Pen as Default Tool


dtreffenstaedt:
An additional setting to be able to select a default line style when selecting the pen as default tool would be nice in order to not have to select the line style separately to the default tool.

From a brief look through the code this seems to be possible to be done through changing the implementation for the LineStyle to one more consistent with the other tool settings by creating an additional enum for LineStyle and passing this as a parameter to the ButtonConfig objects. I understand this would necessitate some internal changes due to the current LineStyle class implementation, though it looks like there are only four distinct styles implemented and identified through strings and so it seems to me like the current design doesn't offer major benefits over an implementation through an enum.

rolandlo:
Good suggestion. That would speed up my workflow as well. It's not only about the line style, see #4177 for more characteristics.

PR #4089, opened 2022-06-13
Fix SelectionTools not considering LineStyle

### LINKED ISSUE #3793: When changing the line style of something which is already drawn, if you already have the line style that you want selected, you must switch off the current line style and back to be able to change the items line style

**Affects versions :**
 - OS: Arch Linux
 - (Linux only) Desktop environment: Gnome Wayland
 - Which version of libgtk do you use: 3.24.31
 - Version of Xournal++: 1.1.0
 - Installation method: Official package from the Arch Linux repositories

**Describe the bug**
If you have an item which is drawn with a specific line style, and you select the item to switch the line style, if you currently have the line style that you want the item to be switched to already seleced, and you try and select it again to switch the item, nothing happens. Instead you must switch off of the line style, select the item, then reselect the desired line style.

**To Reproduce**
Using a circle as an example
1. Select the pen tool
2. Select the standard line style
3. select the draw ellipse tool
4. Draw an ellipse
5. switch the pen to the dashed line line-style
6. select the ellipse
7. try and switch the ellipse to the dased line style by selecting the already selected dased line style, and observe that you cant switch the line style of the ellipse
8. switch back to the standard line style
9. reselect the ellipse
10. select the dased line style and observe that the ellipse is now using the dashed line style

**Expected behavior**
I expect that the selected item will change to whatever line style that I select regardless of what line style is currently selected.

forgedOctopus:
Fixes #3793.

Before the fix, the currently selected LineStyle of the Pen is marked as active in the toolbar during a selection. This prevents changing the LineStyle of the selection in certain situations:

https://user-images.githubusercontent.com/65341804/173325173-41a82a8d-d126-46f5-8fcc-158496136d97.mp4

The PR has the goal to be simple and make only few changes to the source code. It introduces a flag in the LineStyle class. The flag indicates whether the in the toolbar selected LineStyle should be cleared or not:
- if all selected stroke elements have the same LineStyle, the flag is not set and this Style will be marked as active in the toolbar
- if the selected stroke elements have different LineStyles, the flag is set and no LineStyle will be marked active in the toolbar
- if no elements are selected, the original behavior is kept

https://user-images.githubusercontent.com/65341804/173325164-e7210707-a0a0-45bb-80ab-ea1834ad0454.mp4

While the PR provides a simple fix, a larger re-write of the LineStyle implementation would probably make sense. An implementation using Enums like ToolSize's does, for instance, should make the LineStyle implementation more consistent with other Tool implementations and solve #2625.

If there is interest in such a re-write, I'm happy to look into it.

Looking forward to your feedback.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1989,22 +1989,9 @@ void Control::toolFillChanged() {
 }
 
 void Control::toolLineStyleChanged() {
-    const LineStyle& lineStyle = toolHandler->getTool(TOOL_PEN).getLineStyle();
-    string style = StrokeStyle::formatStyle(lineStyle);
+    std::optional<string> style = getLineStyleToSelect();
 
-    std::optional<string> styleOfSelection = std::optional(style);
-    if (win) {
-        const EditSelection* sel = win->getXournal()->getSelection();
-
-        if (sel) {
-            styleOfSelection = getLineStyleOfSelection(sel);
-            if (styleOfSelection && styleOfSelection.value() != "pen") {
-                style = styleOfSelection.value();
-            }
-        }
-    }
-
-    if (!styleOfSelection) {
+    if (!style) {
         fireActionSelected(GROUP_LINE_STYLE, ACTION_NONE);
     } else if (style == "dash") {
         fireActionSelected(GROUP_LINE_STYLE, ACTION_TOOL_LINE_STYLE_DASH);
@@ -2017,9 +2004,21 @@ void Control::toolLineStyleChanged() {
     }
 }
 
-auto getLineStyleOfSelection(const EditSelection* sel) -> std::optional<string> const {
+auto Control::getLineStyleToSelect() -> std::optional<string> const {
+    const LineStyle& lineStyle = toolHandler->getTool(TOOL_PEN).getLineStyle();
+    string style = StrokeStyle::formatStyle(lineStyle);
+
+    if (!win) {
+        return style;
+    }
+
+    const EditSelection* sel = win->getXournal()->getSelection();
+    if (!sel) {
+        return style;
+    }
+
     bool isFirstPenStrokeElement = true;
-    string previous_style = "pen";
+    string previous_style = "none";
 
     // Todo(cpp20) Replace with std::ranges::filter_view and for_first_then_for_each
     for (const Element* e: sel->getElements()) {
@@ -2027,7 +2026,7 @@ auto getLineStyleOfSelection(const EditSelection* sel) -> std::optional<string> 
             const auto* s = dynamic_cast<const Stroke*>(e);
 
             if (s->getToolType() == STROKE_TOOL_PEN) {
-                const string style = StrokeStyle::formatStyle(s->getLineStyle());
+                style = StrokeStyle::formatStyle(s->getLineStyle());
 
                 if (isFirstPenStrokeElement) {
                     previous_style = style;
@@ -2041,7 +2040,7 @@ auto getLineStyleOfSelection(const EditSelection* sel) -> std::optional<string> 
         }
     }
 
-    return previous_style;
+    return style;
 }
 
 void Control::toolColorChanged() {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1839,6 +1839,9 @@ void Control::toolChanged() {
         toolColorChanged();
     }
 
+    if (toolHandler->getToolType() == TOOL_PEN)
+        toolLineStyleChanged();
+
     ActionType rulerAction = ACTION_NOT_SELECTED;
     if (toolHandler->getDrawingType() == DRAWING_TYPE_STROKE_RECOGNIZER) {
         rulerAction = ACTION_SHAPE_RECOGNIZER;
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2857,7 +2857,9 @@ void Control::clipboardPasteXournal(ObjectInputStream& in) {
         g_warning("could not paste, Exception occurred: %s", e.what());
         Stacktrace::printStracktrace();
         if (selection) {
-            for (Element* el: selection->getElements()) { delete el; }
+            for (Element* el: selection->getElements()) {
+                delete el;
+            }
             delete selection;
         }
     }
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2878,9 +2878,7 @@ void Control::clipboardPasteXournal(ObjectInputStream& in) {
         g_warning("could not paste, Exception occurred: %s", e.what());
         Stacktrace::printStracktrace();
         if (selection) {
-            for (Element* el: selection->getElements()) {
-                delete el;
-            }
+            for (Element* el: selection->getElements()) { delete el; }
             delete selection;
         }
     }
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1980,11 +1980,6 @@ void Control::toolFillChanged() {
 }
 
 void Control::toolLineStyleChanged() {
-    if (toolHandler->getLineStyle().getClearLineStyle()) {
-        fireActionSelected(GROUP_LINE_STYLE, ACTION_NONE);
-        return;
-    }
-
     const LineStyle& lineStyle = toolHandler->getTool(TOOL_PEN).getLineStyle();
     string style = StrokeStyle::formatStyle(lineStyle);
 
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2021,6 +2021,7 @@ auto getLineStyleOfSelection(const EditSelection* sel) -> std::optional<string> 
     bool isFirstPenStrokeElement = true;
     string previous_style = "pen";
 
+    // Todo(cpp20) Replace with std::ranges::filter_view and for_first_then_for_each
     for (const Element* e: sel->getElements()) {
         if (e->getType() == ELEMENT_STROKE) {
             const auto* s = dynamic_cast<const Stroke*>(e);
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1993,28 +1993,28 @@ void Control::toolLineStyleChanged() {
     string style = StrokeStyle::formatStyle(lineStyle);
 
     bool isLineStyleSameForAll = true;
-	
-    if (win) {
-        EditSelection* sel = win->getXournal()->getSelection();
-        if (sel) {
-			bool isFirstStrokeElement = true;
-			string previous_style = "none";
-			
-			for (Element* e: sel->getElements()) {
-				if (e->getType() == ELEMENT_STROKE) {
-					auto* s = dynamic_cast<Stroke*>(e);
-					style = StrokeStyle::formatStyle(s->getLineStyle());
 
-					if (isFirstStrokeElement) {
-						previous_style = style;
-						isFirstStrokeElement = false;
-					} else {
-						if (style != previous_style) {
-							isLineStyleSameForAll = false;
-						}
-					}
-				}
-			}
+    if (win) {
+        const EditSelection* sel = win->getXournal()->getSelection();
+        if (sel) {
+            bool isFirstStrokeElement = true;
+            string previous_style = "none";
+
+            for (const Element* e: sel->getElements()) {
+                if (e->getType() == ELEMENT_STROKE) {
+                    const auto* s = dynamic_cast<const Stroke*>(e);
+                    style = StrokeStyle::formatStyle(s->getLineStyle());
+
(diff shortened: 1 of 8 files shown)


forgedOctopus:
The PR should be ready for review again, sorry. 

I rebased the PR on the current master. For this, I did a force-push overwriting the PR's old history.

During the rebase, I made only one change:
I fixed the conflict in [705ac2c](https://github.com/xournalpp/xournalpp/pull/4089/commits/705ac2cd31f5e84789254545257dc529c5a236d6). In the master the import section of `src/core/gui/XournalView.cpp` had been updated. This resulted in a conflict with the by me in the PR added import of `model/StrokeStyle.h`.

Looking forward to your feedback.

forgedOctopus:
Thanks for reaching out. No worries, I totally understand this.

In any case, feel free to ask if there's anything unclear or needs to be done.

forgedOctopus:
> Looking forward to your feedback.

I realized, I might have formulated this badly (English is not my native language). To avoid misunderstandings: I didn't want to pressure you with this. I'm happy for any feedback, but I'm totally aware, that this can take time and this is not a matter of course. 

So no worries if your feedback takes some time.

src/core/control/Control.cpp
```diff
@@ -1839,6 +1839,9 @@ void Control::toolChanged() {
         toolColorChanged();
     }
 
+    if (toolHandler->getToolType() == TOOL_PEN)
+        toolLineStyleChanged();
```
bhennion, line 1843: Please always encapsulate conditional code with braces (see [our coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#indentation))
```suggestion
    if (toolHandler->getToolType() == TOOL_PEN) {
        toolLineStyleChanged();
    }
```

src/core/model/LineStyle.h
```diff
@@ -61,4 +70,9 @@ class LineStyle: public Serializable {
      * Dash count (0 for no dash)
      */
     int dashCount = 0;
+
+    /**
+     * Flag to clear LineStyle selection
+     */
+    bool clearLineStyle = false;
```
bhennion, line 77: I don't like this: the flag has nothing to do with the LineStyle itself, so it should not be part of the data structure.

I guess the way I'd do this is by adding a tool capability (see ToolEnums.h::ToolCapabilities and Tool::setCapability)

src/core/control/ToolHandler.cpp
```diff
@@ -534,13 +534,18 @@ void ToolHandler::setSelectionEditTools(bool setColor, bool setSize, bool setFil
         t->setSize(TOOL_SIZE_NONE);
         t->setColor(Color(-1));
         t->setFill(false);
+
+        LineStyle newStyle = t->getLineStyle();
+        newStyle.setClearLineStyle(!setLineStyle);
+        t->setLineStyle(newStyle);
```
bhennion, line 540:
```suggestion
        t->setCapability(TOOL_CAP_LINE_STYLE, setLineStyle);
        t->setSize(TOOL_SIZE_NONE);
        t->setColor(Color(-1));
        t->setFill(false);
```

with TOOL_CAP_LINE_STYLE declared in ToolEnums.h

src/core/model/LineStyle.cpp
```diff
@@ -27,6 +27,8 @@ void LineStyle::operator=(const LineStyle& other) {
 
     other.getDashes(dashes, dashCount);
     setDashes(dashes, dashCount);
+
+    clearLineStyle = other.clearLineStyle;
```
bhennion, line 31:
```suggestion
```

src/core/model/LineStyle.cpp
```diff
@@ -88,3 +90,8 @@ void LineStyle::setDashes(const double* dashes, int dashCount) {
  * @return true if dashed
  */
 auto LineStyle::hasDashes() const -> bool { return this->dashCount > 0; }
+
+
+void LineStyle::setClearLineStyle(bool clearLineStyle) { this->clearLineStyle = clearLineStyle; }
+
+bool LineStyle::getClearLineStyle() const { return clearLineStyle; }
```
bhennion, line 97:
```suggestion
```

src/core/model/LineStyle.h
```diff
@@ -51,6 +51,15 @@ class LineStyle: public Serializable {
...
 
+    /**
+     * Set flag to clear the LineStyle selection in the toolbar
+     *
+     * @param clearLineStyle true, to clear toolbar LineStyle selection
+     */
+    void setClearLineStyle(bool clearLineStyle);
+
+    bool getClearLineStyle() const;
+
```
bhennion, line 62:
```suggestion
```

src/core/control/Control.cpp
```diff
@@ -1976,9 +1979,27 @@ void Control::toolFillChanged() {
 }
 
 void Control::toolLineStyleChanged() {
+    if (toolHandler->getLineStyle().getClearLineStyle()) {
```
bhennion, line 1982:
```suggestion
    if (!toolHandler->hasCapability(TOOL_CAP_LINE_STYLE)) {
```

bhennion (changes requested):
First, thanks for this great PR!

There is one thing I dislike (see below). Otherwise, just nitpicks.

forgedOctopus:
Thanks a lot for your review and helpful suggestions. 
I will start implementing the required changes.

src/core/control/Control.cpp
forgedOctopus: Fixed by 923338c2.

src/core/model/LineStyle.h
forgedOctopus: Resolved by 76fab8d3808b620d2d7faeef98bfe48b56f03254.

src/core/model/LineStyle.cpp
forgedOctopus: Resolved by 76fab8d3808b620d2d7faeef98bfe48b56f03254.

src/core/model/LineStyle.cpp
forgedOctopus: Resolved by 76fab8d3808b620d2d7faeef98bfe48b56f03254.

src/core/control/ToolHandler.cpp
forgedOctopus: Resolved by baffc58cbe83b76317ee8a0c8f5497dd5b52c85b.

src/core/control/Control.cpp
forgedOctopus: Resolved by fea96d645547dc7db4ef2605aa2ad8c802c8b945 and 660d3f958c816c90568a3594831aa74ada84761c.

src/core/model/LineStyle.h
forgedOctopus: I agree.

I decided to re-write the fix using a new tool capability as you suggested.
In my opinion, the new fix makes way more sense than the previous workaround and does not introduce code where it doesn't belong (e.g. in `XournalView::setSelection()`, `ToolHandler::setSelectionEditTools())`).

forgedOctopus:
Based on your feedback, I decided to re-write the fix using a new tool capability for line style as you suggested.

I think the new fix makes way more sense now than the previous workaround. Furthermore, it improves the behavior: same as with tool size and color, the line style can now only be selected, when the tool actually supports the capability. I added the capability only for the pen tool. Hence, changing the line style is now only possible for the pen.

The video below illustrates the new behavior:

https://user-images.githubusercontent.com/65341804/180599610-d6bb27fb-6e29-41e9-bfa0-0a8fa45996d7.mp4

src/core/control/Control.cpp
```diff
@@ -1979,7 +1992,35 @@ void Control::toolLineStyleChanged() {
...
 
-    if (style == "dash") {
+    bool isLineStyleSameForAll = true;
+	
+    if (win) {
+        EditSelection* sel = win->getXournal()->getSelection();
+        if (sel) {
+			bool isFirstStrokeElement = true;
+			string previous_style = "none";
+
```
bhennion, line 2002: Just as a reminder, for when we pass to c++ 20

```suggestion
			
	\\ Todo(cpp20) Replace with std::ranges::filter_view and for_first_then_for_each
```

src/core/control/Control.cpp
```diff
@@ -1979,7 +1992,35 @@ void Control::toolLineStyleChanged() {
...
+				if (e->getType() == ELEMENT_STROKE) {
+					auto* s = dynamic_cast<Stroke*>(e);
+					style = StrokeStyle::formatStyle(s->getLineStyle());
+
+					if (isFirstStrokeElement) {
+						previous_style = style;
+						isFirstStrokeElement = false;
+					} else {
+						if (style != previous_style) {
+							isLineStyleSameForAll = false;
```
bhennion, line 2013:
```suggestion
							isLineStyleSameForAll = false;
							break;
```

src/core/control/Control.cpp
```diff
@@ -1979,7 +1992,35 @@ void Control::toolLineStyleChanged() {
...
+    if (win) {
+        EditSelection* sel = win->getXournal()->getSelection();
+        if (sel) {
+			bool isFirstStrokeElement = true;
+			string previous_style = "none";
+			
+			for (Element* e: sel->getElements()) {
+				if (e->getType() == ELEMENT_STROKE) {
+					auto* s = dynamic_cast<Stroke*>(e);
+					style = StrokeStyle::formatStyle(s->getLineStyle());
```
bhennion, line 2006: You may wanna filter with `s->getToolType() == STROKE_TOOL_PEN` as well here, to exclude highlighter or whiteout strokes.

bhennion (changes requested):
Thanks for the changes! This is much cleaner indeed.

I still have a couple of comments (see below). Mainly: highlighter strokes should not count in "are there strokes of different line styles?".

I also have another request (it may be out of scope). See this video:

https://user-images.githubusercontent.com/17818041/180601196-01a7de7b-b1ca-4b88-bb13-5add4e67550b.mp4

When we change the line style of selected strokes to, say, dashes, it also changes the line style of the pen tool (so that strokes drawn after that are also dashed).

I'd rather the tool's line style was not affected by whatever happens to the selection (like for colors or stroke width).
This is not a bug introduced by your PR, so you can chose to declare it out of scope :-)

src/core/gui/XournalView.cpp
```diff
@@ -637,16 +638,17 @@ void XournalView::setSelection(EditSelection* selection) {
             if (s->getToolType() != STROKE_TOOL_ERASER) {
                 canChangeColor = true;
                 canChangeFill = true;
+                canChangeLineStyle = true;
```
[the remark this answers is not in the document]
forgedOctopus: True, oversaw this one, thanks.

Resolved by 3c93210ce040a2e7074824f9dbd9b2e7d18eeeca.

src/core/control/Control.cpp
forgedOctopus: Resolved by 018fdef29d82d5a9ac214621cc5d6681ff840bad.

src/core/control/Control.cpp
forgedOctopus: Thanks, resolved by 67e5e9dea630e9ed617d84e44be6078f89b6ecee and the later refactorings.

src/core/control/Control.cpp
forgedOctopus: Added in a1f5563482acc2a3b6becf0801ac0b6a92f22363.

Sorry, due to the refactoring I'm not entirely sure if I put it in the correct spot.

Can you quickly check if it's correct?

bhennion: Yes, that's the right place

forgedOctopus:
Thanks for your quick review and comments!

I could resolve all comments with your suggestions.

I was also able to solve the issue with the tool line style being affected by the line style set for a selection (d4a9f69128e281b7326cf776179808bbcfae7e64).

In addition, I did some more code polish (f973f2689a844081c33b49df2e7a0feb345a0299, 018fdef29d82d5a9ac214621cc5d6681ff840bad, e498971daf98002a794e728600305b20f6da9cf5, ed3c18022169c0ed43615a668bf574b86bb20067).

I kept 018fdef29d82d5a9ac214621cc5d6681ff840bad and e498971daf98002a794e728600305b20f6da9cf5 both instead of one large refactoring commit, so you can see my thought process and decide which to keep / not to keep.  And if you don't think a certain refactoring was a good idea, I can easily revert it.

Below a video to illustrate the current behavior:

https://user-images.githubusercontent.com/65341804/180640363-eb8eac5a-bda8-48ba-8ad9-f4fb95245057.mp4

bhennion (approved):
LGTM.

If you're satisfied as well, we'll give other devs a couple of days to review if they want to, and then squash and merge.

@xournalpp/core Merging in 72 hours unless an objectino is raised.

forgedOctopus:
Sounds good! :)
Thanks for taking your time and providing such helpful reviews.

Issue #4095, opened 2022-06-15
Unnecessary includes in header files should be removed or replaced by forward declarations


bhennion:
The header files contain a lot of unnecessary includes, and a lot of them can be replaced by forward declarations.

Fixing this would greatly improve on build time, especially when rebuilding after minor changes in a small part of the code.

As an example, building the app, touching src/util/RAII_wrappers.h in #4082 and rebuilding triggers a rebuild of almost every .o targets. Replacing the include of gui/PageView.h by a forward declaration in every header reduces this rebuild to a dozen of files.

There are tools out there to automatize this process ([cppclean](https://github.com/myint/cppclean) or [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use))
 
@xournalpp/core Does anyone have experience with either of those tools (or another one)?
Would applying one of them to our source tree be a good idea?

LittleHuba:
Yes this is something that is a problem since quite some time.

I have used IWYU previously. It does it's job. Cppclean seems like a tool that will tell us a bunch of other stuff as well that has nothing to do with includes.

Keep in mind that both tools will just give you hints what to change but you'll still have to change it yourself.

Febbe:
After reading the documentation, it should be very easy to use:

```
cd /path/to/xournalpp
git switch -c iwuu_test
cd build
cmake -GNinja ..
iwyu_tool.py -p .
```

This is, because we already create a compilation database for the use with clang-tidy.

PR #4098, opened 2022-06-16
Apply include-what-you-use to the code tree

bhennion:
In this PR, I applied IWYU (see #4095) to the entire project. Every single source file is touched.

The process is not so streamlined and some intervention was needed.
I spotted some includes that could be replaced by forward declaration and that IWYU did not detect. I also spotted some includes that IWYU added and that were just not necessary (e.g. it added `#include <filesystem>` to `view/StrokeView.cpp`. I removed it).

So there are surely other issues of the sort. I doubt we (or anyone) can double-check everything, so I have mixed feelings about this.

@xournalpp/core would you care to give it a glance and tell me how it feels to you? If that's globally an improvement, I'll either rebase or redo the process on latest master.

In any case, this definitely conflicts with every other PR. Those conflicts are relatively easy to fix, but still.

**Edit**: the command I ran from a new build dir (with the mappingfile.imp cat'ed below inside)
```
cmake GNinja ..
iwyu-tool -j 10 -p . -- -Xiwyu --mapping_file=mappingfile.imp > IWYU.out
iwyu-fix-includes --nosafe_headers < IWYU.out
```
mappingfile.imp:
```
[
    { include: ["<filesystem>", "public", "\"filesystem.h\"", "public"] }
]
```
Adding other lines to the mappingfile to avoid some overly precise include splitting (like fetching subincludes of a library -- Portaudio is a good example) is probably doable too.


LittleHuba:
I feel uncomfortable when doing this for all headers at the same time. It is really taffy to spot all issues. I'd like it more when we do this for smaller units. Not quite sure if that is simple though. I'd assume that includes in one headers might influence includes in another header...

Febbe:
> 

Then it should not compile. 
Conceptually, the order of includes should not matter. (There are few exceptions, when using internal/private api headers like them of mswindows). But also then, the program stops to compile.
The only thing I can imagine are bugs due to the same identifiers in the same compilation unit. And that's only getting better with include what you use.

bhennion:
> Then it should not compile.

Exactly. I tried doing it by hand on a couple of files, just to gauge how messed up things were. I ended up having to edit dozens of files, just to clean up a couple.

One possibility would be to first run `iwyu-fix-includes` without the  `--nosafe_headers` flag. It only adds headers where they should be included.
Then, in a second more careful step, we rerun it with the `--nosafe_headers` flag on a file-by-file (or maybe folder-by-folder) basis. This will remove unecessary headers and replace some of them with forward declarations.

There are a couple of issues to fix, even with the first step. As you can see, the CI build fail because it included files like `<bits/utility.h>` that are compiler specific. There should be a way to avoid that with careful use of the mapping file (see https://github.com/include-what-you-use/include-what-you-use/issues/1045)

LittleHuba:
What I meant is that some headers include some requirements of other headers that are forwarded transitively. So what @bhennion meant: We can't fix one header without having to fix ten others.

Febbe:
> What I meant is that some headers include some requirements of other headers that are forwarded transitively. So what @bhennion meant: We can't fix one header without having to fix ten others.

That's the reason, I would not do this by hand. Instead let iwuu do that for you.

rolandlo:
> > 
> 
> Then it should not compile. Conceptually, the order of includes should not matter. (There are few exceptions, when using internal/private api headers like them of mswindows). But also then, the program stops to compile. The only thing I can imagine are bugs due to the same identifiers in the same compilation unit. And that's only getting better with include what you use.

Some care has to be taken with precompiler flags. When the code compiles, it doesn't necessarily mean that it would compile with a different set of precompiler flags set. I would care too much about the merge conflicts. It's worth the pain if we can significantly reduce compile times.

Febbe:
> Some care has to be taken with precompiler flags.

Exactly, I also would say, that this is ok, as long every pipeline works. Any problem occurring beside that can be fixed in a fix-up PR.

LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjAzNzIvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjAzNzIvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjAzNzIvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 18.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjAzNzIvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0xOC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjAzNzIvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)

bhennion:
Alright, this now pass the CI tests. I have tried to be careful about the build flags.

For posterity, here is how I proceeded.
I created mapping files for various libs (or updated the existing gcc mapping files) to make sure no implementation defined headers were included. Those mapping files are in here: [mapping.tar.gz](https://github.com/xournalpp/xournalpp/files/8934291/mapping.tar.gz)

Then I ran 
```
mkdir build
cd build
cmake GNinja .. -DENABLE_PLUGINS=on -DDEBUG_INPUT=on -DDEBUG_INPUT_GDK_PRINT_EVENTS=on -DDEBUG_INPUT_PRINT_ALL_MOTION_EVENTS=on -DDEBUG_RECOGNIZER=on -DDEBUG_SHEDULER=on -DDEBUG_SHOW_ELEMENT_BOUNDS=on -DDEBUG_SHOW_REPAINT_BOUNDS=on -DDEBUG_SHOW_PAINT_BOUNDS=on -DDEBUG_SHOW_MASK=on -DDEBUG_ERASABLE_STROKE_BOXES=on -DENABLE_ERASER_DEBUG=on -DDEV_CALL_LOG=on -DENABLE_NLS=on
iwyu-tool -p . -- -w -Xiwyu --no_default_mapping -Xiwyu --mapping_file=/path/to/total-map.imp -Xiwyu --cxx17ns > iwyu-all-flags.out
```
and a second time with all the cmake flags off to get another output file iwyu-no-flags.out.
I compared the two outputs (actually, on a .cpp-file per .cpp-file basis -- only 19 .cpp file were handled differently in the two outputs) and kept the includes that appeared in both the IWYU listings.

I then fixed some bugs ("filesystem.h" and <*stream> were mishandled, in particular). Then the OS specific includes were fixed using the CI error logs.

This is most likely not perfect: there are `// #define MY_OWN_FLAG` all over the code. If and when anything goes wrong with one of them, it should straightforward to rectify.

bhennion:
So... Do we want to merge this?

Febbe:
Can you squash them? I wouldn't wait too long to merge this actually.

bhennion:
> Can you squash them? I wouldn't wait too long to merge this actually.

Done. I only kept appart: the two preliminary commits (useful for IWYU to apply correctly) and the last commit that reduces the include-tree but was made by hand and touches a little bit more than includes.

Merging in 24 hours, unless someone objects.

src/core/audio/AudioPlayer.cpp
```diff
@@ -9,6 +9,13 @@
 
 class Settings;
 
+AudioPlayer::AudioPlayer(Control& control, Settings& settings):
+        control(control),
+        settings(settings),
+        audioQueue(std::make_unique<AudioQueue<float>>()),
+        portAudioConsumer(std::make_unique<PortAudioConsumer>(*this, *audioQueue)),
+        vorbisProducer(std::make_unique<VorbisProducer>(*audioQueue)) {}
```
Febbe, line 17: Question: Does such refactorings only happen to classes that have a non default destructor?

Febbe:
Previous:
grafik
Now:
grafik

src/core/audio/AudioPlayer.cpp
bhennion: Yes. Only AudioPlayer, AudioRecorder and AudioController. Previously, the unique_ptr's were initialized in the header file.
Those classes can not have default constructor, because of the references to Settings and Control.

Febbe: I talked from destructors which are relevant.

bhennion: Oh sorry, I misread. Actually, AudioController has a default destructor indeed (the other two don't).

> Classes having a implicitly default destructor should not be refactored in such a way.

Why is that an issue?

Anyway, I can reverse this last commit.

Febbe: No that's fine. There is a tradeoff: compile time vs runtime. By moving the destructor into the source the compiler can't see anymore, that the destructor is trivial, preventing optimisations. Also then, the - rule of 5 applies. But since std::unique_ptr break Copy ctors and all ctors are deleted this is fine.

Febbe: Btw you can merge this. I just wanted to make sure, that this sort of changes should be decided by hand.

bhennion: Thanks for the explanation @Febbe!

PR #4121, opened 2022-06-28
Forward-declare XojPage in PageRef and protect headers

bhennion:
This PR is a followup of #4098: it cleans up a couple of includes in a couple of header files that do not have a .cpp counterpart.

As the changes are completely harmless, I'll be merging this in 24 hours unless someone objects.


Febbe:
When PR's/MR's like this compile on every pipeline, you can merge them theoretically directly :).

Febbe:
Also, thank you for all those code cleanups.

PR #4128, opened 2022-07-03
If you press alt while you are selecting a shape, center the shape on the x axis.

ShaanMakoon:


Febbe:
You can center a shape on the x-axis via snapping.

ShaanMakoon:
Should I change it so it centres on the middle of the y-axis as well? This would be useful if there is not a centring tool.

bhennion:
Hey. First, thanks for the PR. 

First, and I am sorry to say this, but I am not quite sure about the targetted feature. This looks like a very niche feature that would better be implemented as a plugin.

Second, there are several things that do not work with this code:

1. I don't quite know what it does when several strokes are selected. Try and guess where the strokes will end up in the following video:

    https://user-images.githubusercontent.com/17818041/178427611-9dc7d3cb-3651-44cb-b490-1955e4f39dfd.mp4

2. Whenever I hit a keyboard letter key and nothing is selected (try typing in a text box...), I get a SegFault crash [errorlog.20220712-085034.log](https://github.com/xournalpp/xournalpp/files/9090338/errorlog.20220712-085034.log).

ShaanMakoon:
Thanks for your thoughts, really useful as this is my first PR. Should I leave this issue then ?

PR #4130, opened 2022-07-04
Drop old mimelnk MIME types

pinotree:
Those desktop MIME types were needed only with KDE up to 3.x, as it
used to have its own desktop-based MIME type system. KDE 3 is EOL for
many years now, and there are already XDG MIME types.


bhennion:
Thanks for the PR!
This looks good to me.

Merging in 72 hours unless someone objects.

PR #4136, opened 2022-07-05
Fix text edition display bug when an IM is used

### LINKED ISSUE #4135: Pre-edit strings are left unerased
**Affects versions :**
 - OS: Linux
 - Version of Xournal++: release-1.1 branch, since the commit ~eddc80620cd2d5ce2851fcb831a755336c507601~ 951cc92bd40a752f5cfffc4babe2f82af8d76840 

**Describe the bug**
Pre-edit strings are left unerased when you use IME.

**Screenshots of Problem**
Before the commit ~eddc80620cd2d5ce2851fcb831a755336c507601~ 951cc92bd40a752f5cfffc4babe2f82af8d76840:

https://user-images.githubusercontent.com/6355486/177324607-be0c1d9a-d39a-47aa-8995-1fc8b1f0fe05.mp4

After the commit ~eddc80620cd2d5ce2851fcb831a755336c507601~ 951cc92bd40a752f5cfffc4babe2f82af8d76840:

https://user-images.githubusercontent.com/6355486/177324636-575d3232-3bd0-4af7-8b9d-e6f5c4ff0270.mp4

bhennion:
Fix #4135.

The issue was the following: when computing the area to repaint, the TextEditor was relying on Text::calcSize. The problem is that the Text element does not contain the current pre-edition string of the Input Method, leading to the repainting of to small a box.

The fix is thus to properly compute the bounding box of the text.

### DIFF
--- a/src/gui/TextEditor.cpp
+++ b/src/gui/TextEditor.cpp
@@ -6,6 +6,7 @@
 
 #include "control/Control.h"
 #include "undo/ColorUndoAction.h"
+#include "util/Rectangle.h"
 #include "view/DocumentView.h"
 #include "view/TextView.h"
 
@@ -904,8 +905,62 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
     return false;
 }
 
+auto TextEditor::computeBoundingRect() -> Rectangle<double> {
+    /**
+     * We draw on a fake surface to get the size of the printed text
+     * See also TextView::calcSize
+     *
+     * NB: we cannot rely on TextView::calcSize directly, since it would not take the IM preeditString into account.
+     */
+    cairo_surface_t* surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 1, 1);
+    cairo_t* cr = cairo_create(surface);
+    auto* textElement = this->getText();
+
+    PangoLayout* pl = TextView::initPango(cr, textElement);
+
+    GtkTextIter cursorIter = {nullptr};
+    GtkTextMark* cursor = gtk_text_buffer_get_insert(this->buffer);
+    gtk_text_buffer_get_iter_at_mark(this->buffer, &cursorIter, cursor);
+
+    if (!this->preeditString.empty()) {
+        string text = textElement->getText();
+        int offset = gtk_text_iter_get_offset(&cursorIter);
+        int pos = gtk_text_iter_get_line_index(&cursorIter);
+
+        for (gtk_text_iter_set_line_index(&cursorIter, 0); gtk_text_iter_backward_line(&cursorIter);) {
+            pos += gtk_text_iter_get_bytes_in_line(&cursorIter);
+        }
+        gtk_text_iter_set_offset(&cursorIter, offset);
+        string txt = text.substr(0, pos) + preeditString + text.substr(pos);
+
+        PangoAttrList* attrlist = pango_attr_list_new();
+        PangoAttrList* preedit_attrlist = this->preeditAttrList;
+        pango_attr_list_splice(attrlist, preedit_attrlist, pos, preeditString.length());
+        pango_layout_set_attributes(pl, attrlist);
+        pango_attr_list_unref(attrlist);
+        attrlist = nullptr;
+        pango_layout_set_text(pl, txt.c_str(), txt.length());
+    } else {
+        string txt = textElement->getText();
+        pango_layout_set_text(pl, txt.c_str(), txt.length());
+    }
+
+    int w = 0;
+    int h = 0;
+    pango_layout_get_size(pl, &w, &h);
+    double width = (static_cast<double>(w)) / PANGO_SCALE;
+    double height = (static_cast<double>(h)) / PANGO_SCALE;
+    g_object_unref(pl);
+
+    cairo_destroy(cr);
+    cairo_surface_destroy(surface);
+
+    return Rectangle<double>(textElement->getX(), textElement->getY(), width, height);
+}
+
+
 void TextEditor::repaintEditor() {
-    auto rect = this->text->boundingRect();
+    auto rect = this->computeBoundingRect();
     this->previousBoundingBox.unite(rect);
     const double zoom = this->gui->getXournal()->getZoom();
     const double padding = (BORDER_WIDTH_IN_PIXELS + PADDING_IN_PIXELS) / zoom;
--- a/src/gui/TextEditor.h
+++ b/src/gui/TextEditor.h
@@ -19,6 +19,8 @@
 #include "undo/UndoAction.h"
 
 class XojPageView;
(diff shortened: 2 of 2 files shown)


rolandlo:
> NB: the CI build are still failing on Windows due to PortAudio's website being offline. We should fix this ASAP.

Hmm, that should have been fixed 21 days ago, see https://github.com/PortAudio/portaudio/issues/714. Seems it is down again. :(

tattsan:
It works fine for me. Thank you so much!

src/gui/TextEditor.h
```diff
@@ -60,6 +62,9 @@ class TextEditor {
     UndoAction* setColor(Color color);
 
 private:
+    void setTextToPangoLayout(PangoLayout* pl) const;
```
Technius, line 65: It's not obvious to me what this does, so I suggest adding a brief doc comment explaining its purpose.

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+    cairo_destroy(cr);
+    cairo_surface_destroy(surface);
+
+    return Rectangle<double>(textElement->getX(), textElement->getY(), width, height);
+}
+
+
 void TextEditor::repaintEditor() {
-    auto rect = this->text->boundingRect();
+    auto rect = this->computeBoundingRect();
```
Technius, line 967: In the future, we may want to consider caching the bounding rect, since it will currently be recalculated even if the contents do not change (e.g., cursor blink or moving the cursor will cause the bounding rect to be recalculated). Not obvious what type of performance impact this will cause--but I think we are currently still much better compared to when we were rerendering the whole page.

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+        attrlist = nullptr;
+        pango_layout_set_text(pl, txt.c_str(), txt.length());
+    } else {
+        string txt = this->text->getText();
+        pango_layout_set_text(pl, txt.c_str(), txt.length());
+    }
+}
+
+auto TextEditor::computeBoundingRect() -> Rectangle<double> {
+    /**
```
Technius, line 938: nitpick: double `*` should be reserved for documentation comments only.
```suggestion
    /*
```

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+     * We draw on a fake surface to get the size of the printed text
+     * See also TextView::calcSize
+     *
+     * NB: we cannot rely on TextView::calcSize directly, since it would not take the IM preeditString into account.
```
Technius, line 942: I think this is clearer:
```suggestion
     * NB: we cannot rely on TextView::calcSize directly, since it would not take the size changes due to the IM preeditString into account.
```

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
     return false;
 }
 
+void TextEditor::setTextToPangoLayout(PangoLayout* pl) const {
```
Technius, line 908: Note: Although the original code is undocumented, I think this would be a good opportunity to add some comments clarifying the intention. This PR is targeting release-1.1, but we'll be merging into master at some point anyways.

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+    if (!this->preeditString.empty()) {
+        GtkTextIter cursorIter = {nullptr};
```
Technius, line 910:
```suggestion
        // If using an IM with a preedit string, we need to insert the preedit string at the current cursor position.

        // Get current cursor position
        GtkTextIter cursorIter = {nullptr};
```

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+        GtkTextMark* cursor = gtk_text_buffer_get_insert(this->buffer);
+        gtk_text_buffer_get_iter_at_mark(this->buffer, &cursorIter, cursor);
+
+        string text = this->text->getText();
+        int offset = gtk_text_iter_get_offset(&cursorIter);
+        int pos = gtk_text_iter_get_line_index(&cursorIter);
+
+        for (gtk_text_iter_set_line_index(&cursorIter, 0); gtk_text_iter_backward_line(&cursorIter);) {
```
Technius, line 918:
```suggestion
        // Calculate the byte offset into the text contents. Since gtk doesn't provide a method to do this, we need to manually count bytes per line.
        for (gtk_text_iter_set_line_index(&cursorIter, 0); gtk_text_iter_backward_line(&cursorIter);) {
```

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
...
+            pos += gtk_text_iter_get_bytes_in_line(&cursorIter);
+        }
+        gtk_text_iter_set_offset(&cursorIter, offset);
+        string txt = text.substr(0, pos) + preeditString + text.substr(pos);
```
Technius, line 922: nit:
```suggestion
        std::string txt = text.substr(0, pos) + preeditString + text.substr(pos);
```

src/gui/TextEditor.cpp
```diff
@@ -904,8 +905,66 @@ auto TextEditor::blinkCallback(TextEditor* te) -> gint {
     return false;
 }
 
+void TextEditor::setTextToPangoLayout(PangoLayout* pl) const {
+    if (!this->preeditString.empty()) {
+        GtkTextIter cursorIter = {nullptr};
+        GtkTextMark* cursor = gtk_text_buffer_get_insert(this->buffer);
+        gtk_text_buffer_get_iter_at_mark(this->buffer, &cursorIter, cursor);
+
+        string text = this->text->getText();
```
Technius, line 914:
```suggestion
        std::string text = this->text->getText();
```

Technius (approved):
LGTM. Some minor nitpicking below.

src/gui/TextEditor.cpp
bhennion: I agree. I think there are a lot of possible optimizations of this class. Overall, a global refactor would be useful (e.g. removing GtkTextBuffer and use std::string instead -- or maybe a stream, using Pango directly for cursor position, and more)

bhennion:

Issue #3684, opened 2021-12-26
Bidirectionnal linking and nested subpage like Notion


slimhk45:
[[]] to create a nested subpage. Suggest a pop-up to show all already created page with the name typed to just link to It instead of creating a new.

Issue #2020, opened 2020-05-29
Presentation Window


H-M-H:
**Is your feature request related to a problem? Please describe.**
I use Xournal++ to record lessons and want my students to see only the area I draw on. Right now I use OBS Studio to record the Xournal++ window and crop it so only the drawing area is visible.

**Describe the solution you'd like**
I propose an option to show a second window that only displays what is being drawn. Like that I could easily capture this new window and would not need to worry about cropping. Additionally this is probably useful if you do a live presentation using two screens.

**Describe alternatives you've considered**
Of course on could go for implementing a videorecord function using ffmpeg but this is probably a lot more involved and less flexible than just having a second window for presentation. Like that things like choosing the correct audioinput or adding effects can be handled by external programs like OBS Studio.

LittleHuba:
I agree with you on integrating our own recording feature being wide out of scope. There are some amazing apps out there that do a way better job than we would be able to realize.

We plan on making multiple views of a document possible in an upcoming version. As this will take quite some time though, a simple hack up of this feature might be interesting. As long as the view does not have to have editing capabilities it should be possible with reasonable effort. Some other dev than me has to take that up tough. I'm currently bogged down with writing that new version mentioned above.

frohro:
This is very similar to #4239.

bhennion:
Duplicate of #936

Issue #1638, opened 2019-11-29
Zoom window like GoodNote has


MisterPresident:
Hello,
I realized that my writing looks different regarding the zoom I use (which doesn't work that good for me :-( #1516). Moreover, when I zoom closer the lines look thicker and in general it looks more beautiful. I think the pressure sensitivity and zoom do not work properly together. However I don't mind this (bug?) very much.

Today in my lecture I saw someone using GoodNote on an iPad. She was using an feature called "Zoom window". There you can move an rectangle to select a region, which is zoomed in to the bottom of the main window. In this zoomed region you can do the same interactions like you can do in the main window anyways but you are closer and therefore you have more precision while still enjoying the overview over the zoomed out document.

When something is unclear please use Google-Images and search for "GoodNotes zoom window". I think then my feature request is pretty clear.

I don't know how complicated this feature is but I would love to see this feature because it would improve the productivity so much! 

Looking forward to hear from you!
Thanks!

LittleHuba:
We can implement this quite easily once the new rendering system is available. It was designed with multiple views in mind.

LittleHuba:
BTW it is not a bug that zooming changes the line thickness. The pen writes with a global thickness which is of course influenced by the zoom. 
Also precision differs because your hardware only has a certain precision which is mapped to the visible area of the document. This area gets smaller the further you zoom in. Hence more precision for smaller area which in turn results in more precise strokes.

MisterPresident:
That sounds great! Looking forward to the new rendering system! Is there a brunch to try it as early as possible?

LittleHuba:
I'm currently just writing up the code skeleton. The rendering system will be contained in its own library once the library is finished we will merge it with the graphical user interface. Most likely that last step will be done in a branch on this repository which you can try out. But until then a few months will pass.

fopor:
That sounds great!
I miss the zoom box that some iOS apps have (like Good Notes)

mkarpiarz:
Any progress on this one @LittleHuba? Is there a public feature branch (even a WIP one) people can help testing or contributing to?

LittleHuba:
Unfortunately no. I'm bogged down with work since start of this year and have literally no time to work on this project for now. Though you might want to ask @lehmanju since he was involved back than as well and might have resumed work by now.

lguenth:
Hey @lehmanju, any updates on this? Any way we can contribute?

lehmanju:
[...]

PR #4137, opened 2022-07-06
Split geometric shape handlers into Controller-View

### LINKED ISSUE #3579: Possible data race in `StrokeHandler`, `BaseStrokeHandler`, `SplineHandler` (possible others)

bhennion:
This PR is a proposed first step towards splitting gui/PageView into a view and a controller.
Before this split could be possible, we need to split every tool handler: gui/PageView owns tool handlers which will need to be split into a view, owned by a future view/PageView, and a handler, owned by a future control/PageController.

In this PR, we split the derived classes of BaseStrokeHandler (i.e. the handlers for all geometric shapes -- Rectangle, ellipse, arrow and so on -- but not StrokeHandler). A lot of planning has gone into this PR. I'll try my best to describe the plan, stop me if you object to any parts of it.

- The gui/PageView has a pool of `OverlayView`'s, which should, in time, contain what I call overlays: everything that needs to be painted but is not actually in the page model (e.g. current tools, selections, search results highlighting). For now, this pool contains at most 1 element, corresponding, if there is one, to the active BaseStrokeHandler.
- In order to be able, in a distant future, to have several views of the current document (e.g. to implement #1638 or #2020), each BaseStrokeHandler now has a pool of views, and dispatches repaint requests to the entire pool. Similar things should be implemented for other tools (and other overlays) in the future.
- The OverlayView's inherit from a base class Repaintable, which should, in due time, take over gui/Redrawable. ~~The class Repaintable includes a possible change of coordinates, so that repaint request are always expressed in local coordinates, and transfered up to a parent Repaintable after changing from child coordinates to parent coordinates (think of selections, pages in a layout, or maybe one day, subpages #3684).~~ Edit: This was removed in an iteration of the PR.
- Once this strategy is applied to all other tool handlers (at least StrokeHandler, SplineHandler, SetsquareHandler, Selection, EditSelection, PdfElemSelection, VerticalToolHandler, SearchControl), then we can actually split gui/PageView into a view and a controller. The view view/PageView would inherit from Repaintable, and control/PageController would have a pool of view/PageView's to send repaint/rerender/change requests to.

More specific implementation comments:
- To accomodate the various dispatch needs, a very generic DispatchPool class is implemented. (I'd love feedback on the implementation here)
- ~~The data of the BaseStrokeHandler's is also now protected by a mutex, to address #3579, at least for those tools.~~
    **Edit:** this is actually not necessary, see #3579.
- The class BaseStrokeHandler is renamed to BaseShapeHandler in order to fix its confusing name.

### DIFF
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -5,7 +5,7 @@
 
 #include "control/Control.h"                       // for Control
 #include "control/settings/Settings.h"             // for Settings
-#include "control/tools/BaseStrokeHandler.h"       // for BaseStrokeHandler
+#include "control/tools/BaseShapeHandler.h"        // for BaseShapeHandler
 #include "control/tools/SnapToGridInputHandler.h"  // for SnapToGridInputHan...
 #include "gui/XournalView.h"                       // for XournalView
 #include "gui/inputdevices/PositionInputData.h"    // for PositionInputData
@@ -16,7 +16,7 @@ class XojPageView;
 
 EllipseHandler::EllipseHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page, bool flipShift,
                                bool flipControl):
-        BaseStrokeHandler(xournal, redrawable, page, flipShift, flipControl) {}
+        BaseShapeHandler(xournal, redrawable, page, flipShift, flipControl) {}
 
 EllipseHandler::~EllipseHandler() = default;
 
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -1,96 +1,108 @@
 #include "EllipseHandler.h"
 
 #include <algorithm>  // for max
-#include <cmath>      // for abs, pow, M_PI, cos
+#include <cassert>
+#include <cmath>  // for abs, pow, M_PI, cos
 
 #include "control/Control.h"                       // for Control
 #include "control/settings/Settings.h"             // for Settings
 #include "control/tools/BaseShapeHandler.h"        // for BaseShapeHandler
 #include "control/tools/SnapToGridInputHandler.h"  // for SnapToGridInputHan...
-#include "gui/XournalView.h"                       // for XournalView
-#include "gui/inputdevices/PositionInputData.h"    // for PositionInputData
-#include "model/Stroke.h"                          // for Stroke
+#include "model/Point.h"                           // for Point
 
-class XojPageView;
-
-
-EllipseHandler::EllipseHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page, bool flipShift,
-                               bool flipControl):
-        BaseShapeHandler(xournal, redrawable, page, flipShift, flipControl) {}
+EllipseHandler::EllipseHandler(Control* control, const PageRef& page, bool flipShift, bool flipControl):
+        BaseShapeHandler(control, page, flipShift, flipControl) {}
 
 EllipseHandler::~EllipseHandler() = default;
 
-
-void EllipseHandler::drawShape(Point& c, const PositionInputData& pos) {
-    this->currPoint = c;
-
+auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isControlDown)
+        -> std::pair<std::vector<Point>, Range> {
     /**
      * Snap point to grid (if enabled - Alt key pressed will toggle)
      */
-    c = snappingHandler.snapToGrid(c, pos.isAltDown());
+    Point c = snappingHandler.snapToGrid(this->currPoint, isAltDown);
 
-    if (!this->started)  // initialize first point
-    {
-        this->startPoint = c;
-        this->started = true;
-        stroke->addPoint(c);  // avoid complaints about <2 points.
-    } else {
-        double width = c.x - this->startPoint.x;
-        double height = c.y - this->startPoint.y;
+    double width = c.x - this->startPoint.x;
+    double height = c.y - this->startPoint.y;
 
 
-        this->modShift = pos.isShiftDown();
-        this->modControl = pos.isControlDown();
+    this->modShift = isShiftDown;
+    this->modControl = isControlDown;
 
-        Settings* settings = xournal->getControl()->getSettings();
-        if (settings->getDrawDirModsEnabled())  // change modifiers based on draw dir
-        {
-            this->modifyModifiersByDrawDir(width, height, true);
-        }
-
-        if (this->modShift)  // make circle
-        {
-            int signW = width > 0 ? 1 : -1;
-            int signH = height > 0 ? 1 : -1;
-            width = (this->modControl) ? sqrt(pow(width, 2) + pow(height, 2)) :
-                                         std::max(width * signW, height * signH) * signW;
-            height = (width * signW) * signH;
-        }
-
-        double diameterX = 0;
-        double diameterY = 0;
-        int npts = 0;
-        double center_x = 0;
-        double center_y = 0;
-        double angle = 0;
-
-        // set resolution proportional to radius
-        if (!this->modControl) {
-            diameterX = width;
-            diameterY = height;
-            npts = static_cast<int>(std::abs(diameterX * 2.0));
-            center_x = this->startPoint.x + width / 2.0;
-            center_y = this->startPoint.y + height / 2.0;
-            angle = 0;
-        } else {  // control key down, draw centered at cursor
-            diameterX = width * 2.0;
-            diameterY = height * 2.0;
-            npts = static_cast<int>(std::abs(diameterX) + std::abs(diameterY));
-            center_x = this->startPoint.x;
-            center_y = this->startPoint.y;
-            angle = 0;
-        }
-        if (npts < 24) {
-            npts = 24;  // min. number of points
-        }
-
-        // remove previous points
-        stroke->deletePointsFrom(0);
-        for (int j = 0; j <= npts; j++) {
-            int i = j % npts;  // so that we end exactly where we started.
-            double xp = center_x + diameterX / 2.0 * cos((2 * M_PI * i) / npts + angle + M_PI);
-            double yp = center_y + diameterY / 2.0 * sin((2 * M_PI * i) / npts + angle + M_PI);
-            stroke->addPoint(Point(xp, yp));
-        }
+    Settings* settings = control->getSettings();
+    if (settings->getDrawDirModsEnabled()) {
+        // change modifiers based on draw dir
+        this->modifyModifiersByDrawDir(width, height, true);
     }
+
+    if (this->modShift) {
+        // make circle
+        width = (this->modControl) ? std::hypot(width, height) :
+                                     std::copysign(std::max(std::abs(width), std::abs(height)), width);
+        height = std::copysign(width, height);
+    }
+
+    double radiusX = 0;
+    double radiusY = 0;
+    double center_x = 0;
+    double center_y = 0;
+
+    if (!this->modControl) {
+        radiusX = 0.5 * width;
+        radiusY = 0.5 * height;
+        center_x = this->startPoint.x + radiusX;
+        center_y = this->startPoint.y + radiusY;
+    } else {
+        // control key down, draw centered at cursor
+        radiusX = width;
+        radiusY = height;
+        center_x = this->startPoint.x;
+        center_y = this->startPoint.y;
+    }
+
+    /*
+     * Set resolution depending on the radius (heuristic)
+     */
+    auto nbPtsPerQuadrant =
+            static_cast<unsigned int>(std::ceil(5 + 0.1 * std::max(std::abs(radiusX), std::abs(radiusY))));
+    const double stepAngle = M_PI_2 / nbPtsPerQuadrant;
+    const double excentricity = std::abs(radiusY / radiusX);
+
+    std::pair<std::vector<Point>, Range> res;
+    std::vector<Point>& shape = res.first;
+
+    /*
+     * This call to reserve() makes the calls to std::transform() below safe.
+     * DO NOT REMOVE
+     * NB: the +1 is necessary to add a copy of the first point and close the ellipse.
+     */
+    shape.reserve(4 * nbPtsPerQuadrant + 1);
+
+    shape.emplace_back(center_x + radiusX, center_y);
+    for (unsigned int j = 1U; j < nbPtsPerQuadrant; j++) {
+        const double tgtAngle = stepAngle * j;
+        // This formula dispatches the points, so that points are denser where the ellipse has higher curvature.
+        const double centerAngle = 0.5 * (std::atan(excentricity * std::tan(tgtAngle)) + tgtAngle);
+        double xp = center_x + radiusX * std::cos(centerAngle);
+        double yp = center_y + radiusY * std::sin(centerAngle);
+        shape.emplace_back(xp, yp);
+    }
+    shape.emplace_back(center_x, center_y + radiusY);
+
+    // The following std::transform() are only safe because no reallocations will happen (see reserve() above).
+    // Symmetry for second quadrant
+    assert(shape.capacity() >= 2 * shape.size() - 1);
+    std::transform(std::next(shape.rbegin()), shape.rend(), std::back_inserter(shape),
+                   [&](const Point& p) { return Point(2 * center_x - p.x, p.y); });
+
+    // Symmetry for the second half
+    assert(shape.capacity() >= 2 * shape.size() - 1);
+    std::transform(std::next(shape.rbegin()), shape.rend(), std::back_inserter(shape),
+                   [&](const Point& p) { return Point(p.x, 2 * center_y - p.y); });
+
+    Range rg(center_x + radiusX, center_y + radiusY);
+    rg.addPoint(center_x - radiusX, center_y - radiusY);
+    res.second = rg;
+
+    return res;
 }
--- a/src/core/control/tools/InputHandler.cpp
+++ b/src/core/control/tools/InputHandler.cpp
@@ -9,54 +9,47 @@
 #include "control/Control.h"          // for Control
 #include "control/ToolEnums.h"        // for TOOL_ERASER, TOOL_HIGHLIGHTER
 #include "control/ToolHandler.h"      // for ToolHandler
-#include "gui/XournalView.h"          // for XournalView
 #include "model/Point.h"              // for Point, Point::NO_PRESSURE
 #include "model/Stroke.h"             // for Stroke, STROKE_TOOL_ERASER, STR...
 #include "util/Color.h"               // for Color
 
 #include "filesystem.h"  // for path
 
-InputHandler::InputHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page):
-        xournal(xournal), redrawable(redrawable), page(page), stroke(nullptr) {}
+InputHandler::InputHandler(Control* control, const PageRef& page): control(control), page(page) {}
 
 InputHandler::~InputHandler() = default;
 
-/**
- * @return Current editing stroke
- */
-auto InputHandler::getStroke() -> Stroke* { return stroke; }
+auto InputHandler::getStroke() const -> Stroke* { return stroke.get(); }
 
-void InputHandler::createStroke(Point p) {
-    ToolHandler* h = xournal->getControl()->getToolHandler();
+auto InputHandler::createStroke(Control* control) -> std::unique_ptr<Stroke> {
+    ToolHandler* h = control->getToolHandler();
 
-    stroke = new Stroke();
-    stroke->setWidth(h->getThickness());
-    stroke->setColor(h->getColor());
-    stroke->setFill(h->getFill());
-    stroke->setLineStyle(h->getLineStyle());
+    auto s = std::make_unique<Stroke>();
+    s->setWidth(h->getThickness());
+    s->setColor(h->getColor());
+    s->setFill(h->getFill());
+    s->setLineStyle(h->getLineStyle());
 
     if (h->getToolType() == TOOL_PEN) {
-        stroke->setToolType(STROKE_TOOL_PEN);
+        s->setToolType(STROKE_TOOL_PEN);
 
-        if (xournal->getControl()->getAudioController()->isRecording()) {
-            fs::path audioFilename = xournal->getControl()->getAudioController()->getAudioFilename();
-            size_t sttime = xournal->getControl()->getAudioController()->getStartTime();
+        if (control->getAudioController()->isRecording()) {
+            fs::path audioFilename = control->getAudioController()->getAudioFilename();
+            size_t sttime = control->getAudioController()->getStartTime();
             size_t milliseconds = ((g_get_monotonic_time() / 1000) - sttime);
-            stroke->setTimestamp(milliseconds);
-            stroke->setAudioFilename(audioFilename);
+            s->setTimestamp(milliseconds);
+            s->setAudioFilename(audioFilename);
         }
     } else if (h->getToolType() == TOOL_HIGHLIGHTER) {
-        stroke->setToolType(STROKE_TOOL_HIGHLIGHTER);
-        p.z = Point::NO_PRESSURE;
+        s->setToolType(STROKE_TOOL_HIGHLIGHTER);
     } else if (h->getToolType() == TOOL_ERASER) {
-        stroke->setToolType(STROKE_TOOL_ERASER);
-        stroke->setColor(Color(0xffffffU));
-        p.z = Point::NO_PRESSURE;
+        s->setToolType(STROKE_TOOL_ERASER);
+        s->setColor(Color(0xffffffU));
     }
 
-    stroke->addPoint(p);
+    return s;
 }
 
 auto InputHandler::validMotion(Point p, Point q) -> bool {
-    return hypot(p.x - q.x, p.y - q.y) >= PIXEL_MOTION_THRESHOLD;
+    return std::hypot(p.x - q.x, p.y - q.y) >= PIXEL_MOTION_THRESHOLD;
 }
--- a/src/core/control/tools/RectangleHandler.cpp
+++ b/src/core/control/tools/RectangleHandler.cpp
@@ -4,7 +4,7 @@
 
 #include "control/Control.h"                       // for Control
 #include "control/settings/Settings.h"             // for Settings
-#include "control/tools/BaseStrokeHandler.h"       // for BaseStrokeHandler
+#include "control/tools/BaseShapeHandler.h"        // for BaseShapeHandler
 #include "control/tools/SnapToGridInputHandler.h"  // for SnapToGridInputHan...
 #include "gui/XournalView.h"                       // for XournalView
 #include "gui/inputdevices/PositionInputData.h"    // for PositionInputData
@@ -14,7 +14,7 @@ class XojPageView;
 
 RectangleHandler::RectangleHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page, bool flipShift,
                                    bool flipControl):
-        BaseStrokeHandler(xournal, redrawable, page, flipShift, flipControl) {}
+        BaseShapeHandler(xournal, redrawable, page, flipShift, flipControl) {}
 
 RectangleHandler::~RectangleHandler() = default;
 
--- a/src/core/control/tools/RectangleHandler.cpp
+++ b/src/core/control/tools/RectangleHandler.cpp
@@ -6,70 +6,54 @@
 #include "control/settings/Settings.h"             // for Settings
 #include "control/tools/BaseShapeHandler.h"        // for BaseShapeHandler
 #include "control/tools/SnapToGridInputHandler.h"  // for SnapToGridInputHan...
-#include "gui/XournalView.h"                       // for XournalView
-#include "gui/inputdevices/PositionInputData.h"    // for PositionInputData
-#include "model/Stroke.h"                          // for Stroke
+#include "model/Point.h"                           // for Point
 
-class XojPageView;
 
-RectangleHandler::RectangleHandler(XournalView* xournal, XojPageView* redrawable, const PageRef& page, bool flipShift,
-                                   bool flipControl):
-        BaseShapeHandler(xournal, redrawable, page, flipShift, flipControl) {}
+RectangleHandler::RectangleHandler(Control* control, const PageRef& page, bool flipShift, bool flipControl):
+        BaseShapeHandler(control, page, flipShift, flipControl) {}
 
 RectangleHandler::~RectangleHandler() = default;
 
-void RectangleHandler::drawShape(Point& c, const PositionInputData& pos) {
-    this->currPoint = c;
-
+auto RectangleHandler::createShape(bool isAltDown, bool isShiftDown, bool isControlDown)
+        -> std::pair<std::vector<Point>, Range> {
     /**
      * Snap point to grid (if enabled)
      */
-    c = snappingHandler.snapToGrid(c, pos.isAltDown());
+    Point c = snappingHandler.snapToGrid(this->currPoint, isAltDown);
 
-    if (!this->started)  // initialize first point
-    {
-        this->startPoint = c;
-        this->started = true;
-        stroke->addPoint(c);  // avoid complaints about <2 points.
-    } else {
-        double width = c.x - this->startPoint.x;
-        double height = c.y - this->startPoint.y;
+    double width = c.x - this->startPoint.x;
+    double height = c.y - this->startPoint.y;
 
 
-        this->modShift = pos.isShiftDown();
-        this->modControl = pos.isControlDown();
+    this->modShift = isShiftDown;
+    this->modControl = isControlDown;
 
-        Settings* settings = xournal->getControl()->getSettings();
-        if (settings->getDrawDirModsEnabled())  // change modifiers based on draw dir
-        {
-            this->modifyModifiersByDrawDir(width, height, true);
-        }
-
-        if (this->modShift)  // make square
-        {
-            int signW = width > 0 ? 1 : -1;
-            int signH = height > 0 ? 1 : -1;
-            width = std::max(width * signW, height * signH) * signW;
-            height = (width * signW) * signH;
-        }
-
-        Point p1;
-        if (!this->modControl) {
-            p1 = this->startPoint;
-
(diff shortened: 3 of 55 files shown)


bhennion:
> I'll have to take a closer look at the last commit. That is quite big and needs some time to focus on all parts.

Of course! No hurry there.

> Regarding the dispatch pool: Please avoid macros whenever possible. They lead to really complex bugs that are amazingly hard to debug.

Yeah. My first attempt had no macro, creating a listener base class, but I failed to make something satisfactory.
The issue is the following. Consider the following code
```
template<class T> class DispatchPool {
    template <typename... Args>
    void dispatch(Args&&... args) const {
        for (auto* v: this->pool) {
            v->on(std::forward<Args>(args)...);
        }
    }
    std::vector<T*> pool;
};
template<class T> class Listener {
    void addToPool(DispatchPool<T>*);
    void removeFromPool();
    DispatchPool<T>* pool;
};

class MyClass: public Listener<MyClass>{
   void on(const MyEventType& event);
};
```
This works fine, except `Listener<T>::addToPool` has to recast its `this` to a `T*` in order to add it to the pool. 
This is unsafe unless there is a way (say a static_assert) to make sure that any instantiation of `Listener<T>` is actually of type `T`. Is there such a way?
Of course, I could resolve to a dynamic_cast, but I'd like to get rid of the oveheads.

bhennion:
I think I found an ok solution for getting rid of the macro DispatchPool.

src/core/control/tools/InputHandler.cpp
```diff
@@ -16,47 +16,44 @@
...
-        xournal(xournal), redrawable(redrawable), page(page), stroke(nullptr) {}
+InputHandler::InputHandler(XournalView* xournal, const PageRef& page): xournal(xournal), page(page) {}
 
 InputHandler::~InputHandler() = default;
 
-/**
- * @return Current editing stroke
- */
-auto InputHandler::getStroke() -> Stroke* { return stroke; }
+auto InputHandler::getStroke() const -> Stroke* { return stroke.get(); }
```
Febbe, line 23: It looks like, that stroke will never be initialized anymore, how about removing this then?

src/util/Range.cpp
```diff
@@ -1,33 +1,47 @@
...
-    this->y1 = y;
-    this->y2 = y;
-}
+Range::Range():
+        minX(std::numeric_limits<double>::max()),
+        minY(std::numeric_limits<double>::max()),
+        maxX(std::numeric_limits<double>::lowest()),
+        maxY(std::numeric_limits<double>::lowest()) {}
+
+Range::Range(double x, double y): minX(x), minY(y), maxX(x), maxY(y) {}
```
Febbe, line 14: Move to header, since this does not rely on another header file (faster compile time).
```suggestion
```

src/util/include/util/Range.h
```diff
@@ -11,11 +11,18 @@
...
+class Rectangle;
+};
 
-class Range {
+class Range final {
 public:
+    Range();  // Empty range
     Range(double x, double y);
-    virtual ~Range();
+    Range(double minX, double minY, double maxX, double maxY);
```
Febbe, line 23:
```suggestion
    Range(double minX, double minY, double maxX, double maxY): minX(minX), minY(minY), maxX(maxX), maxY(maxY) {}
```

src/util/include/util/Range.h
```diff
@@ -11,11 +11,18 @@
...
+namespace xoj::util {
+template <typename Float>
+class Rectangle;
+};
 
-class Range {
+class Range final {
 public:
+    Range();  // Empty range
     Range(double x, double y);
```
Febbe, line 22:
```suggestion
    Range(double x, double y): minX(x), minY(y), maxX(x), maxY(y) {}
```

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,83 @@
...
+template <class T>
+class DispatchPool {
+public:
+    using listener_type = T;
+    DispatchPool() = default;
+    DispatchPool(DispatchPool&&) = default;
+    DispatchPool(const DispatchPool&) = default;
+    DispatchPool& operator=(DispatchPool&&) = default;
+    DispatchPool& operator=(const DispatchPool&) = default;
+
```
Febbe, line 29: Either make this class final *and* remove all default special member functions, or add a defaulted virtual destructor

src/core/view/overlays/ShapeToolView.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace xoj::view {
+struct Context;
+class StrokeView;
+
+class ShapeToolView: public OverlayView, public Repaintable, public xoj::util::Listener<ShapeToolView> {
+
+public:
+    ShapeToolView(BaseShapeHandler* toolHandler, const Repaintable* parent);
+    virtual ~ShapeToolView() noexcept;
```
Febbe, line 32: Use override instead of virtual. Also, can this class be defined final?
```suggestion
    ~ShapeToolView() override noexcept;
```

src/core/view/overlays/ShapeToolView.h
```diff
@@ -0,0 +1,56 @@
...
+     * @brief Draws the overlay to the given context
+     */
+    void draw(const Context& ctx) const override;
+
+    bool isViewOf(const OverlayBase* overlay) const override;
+
+    /**
+     * Listener interface
+     */
+    static constexpr struct PAINT_REQUEST_t {
```
Febbe, line 44: Please use snake_case or PascalCase for types. `snake_case` for STL like/mimic structures. `PascalCase` for Xournal internal classes.

src/core/model/OverlayBase.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+class OverlayBase {
+public:
+    virtual ~OverlayBase() = default;
+};
```
Febbe, line 17: What's the use case of an empty base class, without virtual functions to call?

src/core/gui/PageView.h
```diff
@@ -39,8 +41,9 @@ class Element;
 class PositionInputData;
 class Range;
 class TexImage;
+class OverlayBase;
 
-class XojPageView: public Redrawable, public PageListener {
+class XojPageView: public Redrawable, public PageListener, public xoj::view::Repaintable {
```
Febbe, line 46: We already have Redrawable, why we need an additional Repaintable.

src/core/gui/PageView.cpp
```diff
@@ -302,26 +306,43 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
+                break;
+            case DRAWING_TYPE_COORDINATE_SYSTEM:
+                shapeHandler = new CoordinateSystemHandler(this->xournal, this, getPage());
+                break;
+            default:
+                this->inputHandler = new StrokeHandler(this->xournal, this, getPage());
+        }
+        if (shapeHandler) {
+            this->inputHandler = shapeHandler;
+            this->overlayViews.emplace_back(std::make_unique<xoj::view::ShapeToolView>(shapeHandler, this));
```
Febbe, line 341: It really feels bad that we recreate all Handlers, when a button is pressed. This looks like a huge performance loss.
It would be better to either have a std::variant, containing all. Or have them all pre-allocated and only std::exchange them in.

Febbe (changes requested):
I mostly gave language related feedback, since @LittleHuba is the code owner of this part. So he can give you the best feedback on this.

But I have also a question, did you look at the gtk4 drawing-API? They want to replace Cairo with a node based, approach. Therefore, your changes should not increase the porting effort.

src/util/Range.cpp
```diff
@@ -1,33 +1,47 @@
...
+
+void Range::addPadding(double padding) {
+    this->minX -= padding;
+    this->maxX += 2.0 * padding;
+    this->minY -= padding;
+    this->maxY += 2.0 * padding;
+}
 
-auto Range::getY2() const -> double { return this->y2; }
+auto Range::empty() const -> bool { return this->minX > this->maxX || this->minY > this->maxY; }
```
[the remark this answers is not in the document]
bhennion: I'm confused. First, is could be `invalid`, maybe, but certainly not `valid`.
Secondly, a Range corresponds to the area of points (x,y) such that minX <= x <= maX and minY <= y <= maxY, right?
So if minX > maxX then the range represents an empty area of the plane (and the same with y's).
Also, as soon as `range.addPoint` has been called at least once, then `range.empty` will return false (whatever the previous state was).

How is `empty` not a good description then?

src/core/model/OverlayBase.h
bhennion: Wait no! There was a use case: see `XojPageView::deleteViewOf(OverlayBase*)`. I just needed a common pointer type for everything that will be an overlay (for now, only InputHandler, but more should come).

bhennion:
> But I have also a question, did you look at the gtk4 drawing-API? They want to replace Cairo with a node based, approach. Therefore, your changes should not increase the porting effort.

I don't think this PR influences (in any direction) the difficulty of porting to GTK4. Basically, this just regroups all the drawing operations in view/, but does not change anything to how the painting is done.

There will be more reflection to do when it'll be time to refactor XojPageView itself. I don't quite get how to replace the current buffer (a cairo_surface_t) when passing to GTK4.

src/core/model/OverlayBase.h
Febbe: Then OverlayBase should named ViewBase, and it should require all view functions defined from Repaintable.

src/core/view/CoordinateChange.h
```diff
@@ -0,0 +1,38 @@
...
+
+class Range;
+
+namespace xoj::view {
+
+class CoordinateChange final {
+public:
+    CoordinateChange() = default;
+    CoordinateChange(const utl::Point<double> origin, double scaleFactor);
+    ~CoordinateChange() = default;
```
Febbe, line 24: Remove default destructor for finalized type
```suggestion
```

Febbe:
For now, Cairo still exist in gtk4, it just requires an extra indirection and must be added as node into the drawing tree. But it might get deprecated some time.   
The larger problem is, that gtk4 globalized input. Now every input is caught globally and is fed into gtk-input-controlers, which then determine, if they have to accept the event. Multiple Controllers can accept the same event, and it is also incompatible with the current approach. The old approach to get events from a widget does not work anymore. Widgets don't emit events.

src/core/gui/PageView.cpp
Febbe: It's often safer to recreate, than to reset a structure, but I would just call something like this:

`*handler = Handler{};`
instead of `handler = new Handler{};`
or `new(handler) Handler{};`

src/core/gui/PageView.h
Febbe: Ok, how about renaming `Redrawable` -> `LegacyRedrawable` and `Repaintable` -> `Redrawable` ? :D

Febbe: At least the name wouldn't be "stupid" and the other name states, that the type is about to replaced some time.

src/core/model/OverlayBase.h
bhennion: No, it's not a view. It's a base for some controllers: those controllers which can have one or more views. This includes all the tool controllers (by that I mean things in control/tools), but also SearchControl for instance (to highlight the search results).

It is obviously not clear or we wouldn't be having this discussion, so something needs to be done to clarify the situation. Not sure what yet.

bhennion:
> For now, Cairo still exist in gtk4, it just requires an extra indirection and must be added as node into the drawing tree. But it might get deprecated some time.

As long as [GtkDrawingArea](https://docs.gtk.org/gtk4/class.DrawingArea.html) relies on cairo for its draw function, we are safe.

> The larger problem is, that gtk4 globalized input. Now every input is caught globally and is fed into gtk-input-controlers, which then determine, if they have to accept the event. Multiple Controllers can accept the same event, and it is also incompatible with the current approach. The old approach to get events from a widget does not work anymore. Widgets don't emit events.

This seems like a problem for the input system rather then for our tool handlers, does it not?

Febbe:
> This seems like a problem for the input system rather then for our tool handlers, does it not?

Those tool handlers rely on, or are part of, the input system.

bhennion:
> Those tool handlers rely on, or are part of, the input system.

Of course they rely on it. My point was: whatever is in gui/inputdevices should play the role of an abstraction layer, so that whatever relies on it should not be touched by any changes in the library.

bhennion:
I refactored the view classes a little bit, for better code sharing between this and the classes to come in #4158.

Technius:
This is rather large, so I'll take a look over the week.

bhennion:
Just rebased after the preliminary code cleanups were merged into master. I also added some documentation in the code.
@Technius Thanks a lot!
I think some files deserve more attention than the rest of them: namely util/DispatchPool, core/control/BaseShapeHandler and core/view/overlays/*. This is where the actual View-Controller stplitting happens.

bhennion:
To help with the reviewing of this PR, as well as #4158, #4159 and future PR's, here is a very coarse diagram of the intended structure:
ViewRefactor

This PR, and #4158 and #4159 concern only the blue (OverlayView's) and green (Overlay's, including InputHandler -- which IMO should be renamed) portions.


Now, in the last commit [e85c908](https://github.com/xournalpp/xournalpp/pull/4137/commits/e85c908690ba31d743d156fd98d11b5013d985ca), I fixed a blinking problem upon shape finalization. I do not like the way this is done, so I'd be delighted if someone had a better idea. I'll describe the issue in detail in a second post.

bhennion:
@xournalpp/core As I explained in #3579, I made a mistake then and never rechecked. Let it be a lesson for me.

In any case, I need to remove the mutexes from this PR. Don't review it just yet... I'm sorry about that.

bhennion:
> In any case, I need to remove the mutexes from this PR. Don't review it just yet... I'm sorry about that.

Well... Maybe I should not remove the mutexes just yet. There may be multiple threads involved if and when we allow several views of the same document.

src/core/view/overlays/BaseShapeOrSplineToolView.cpp
```diff
@@ -0,0 +1,84 @@
...
+        return mask.cr;
+    } else {
+        cairo_set_line_cap(cr, CAIRO_LINE_CAP_ROUND);
+        cairo_set_line_join(cr, CAIRO_LINE_JOIN_ROUND);
+        cairo_set_line_width(cr, this->strokeWidth);
+
+        const double* dashes = nullptr;
+        int dashCount = 0;
+        this->lineStyle.getDashes(dashes, dashCount);
+        assert((dashCount == 0 && dashes == nullptr) || (dashCount != 0 && dashes != nullptr));
```
Technius, line 59: Same concern here about the assertion.

src/core/view/overlays/BaseStrokeToolView.h
```diff
@@ -0,0 +1,61 @@
...
+#include "model/LineStyle.h"
+#include "util/Color.h"
+#include "view/overlays/OverlayView.h"
+
+class Stroke;
+
+namespace xoj::view {
+class Repaintable;
+
+class BaseStrokeToolView: public OverlayView {
```
Technius, line 25: Needs a doc comment.

What are intended subclasses of `BaseStrokeToolView` other than the `BaseShapeOrSpineToolView` family?

src/core/view/overlays/BaseStrokeToolView.h
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @brief All that's required to draw a stroke in cairo
+     */
+    cairo_operator_t cairoOp;
+    Color strokeColor;
+    LineStyle lineStyle;
+    double strokeWidth;
+
+
+    class Mask {
```
Technius, line 45: Needs a doc comment.

src/core/view/overlays/BaseStrokeToolView.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const int xOffset = -static_cast<int>(std::floor(visibleRange.minX * scale));
+    const int yOffset = -static_cast<int>(std::floor(visibleRange.minY * scale));
+    const int width = static_cast<int>(std::ceil(visibleRange.maxX * scale)) + xOffset;
+    const int height = static_cast<int>(std::ceil(visibleRange.maxY * scale)) + yOffset;
+
+    return Mask(width, height, xOffset, yOffset, scale);
+}
+
+BaseStrokeToolView::Mask::Mask(int width, int height, int xOffset, int yOffset, double scale) {
+    cairo_surface_t* surf = cairo_image_surface_create(CAIRO_FORMAT_A8, width, height);
```
Technius, line 47: For performance reasons, we might want to make it possible to use something like [`cairo_surface_create_similar()`](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-create-similar), otherwise we might be forcing the image buffer to be allocated in Xournal++'s heap instead of e.g. the display server or VRAM etc.

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,80 @@
...
+class Listener {
+private:
+    // Keep the constructor private: only T can inherit Listener<T>
+    Listener() = default;
+    friend T;
+
+    using pool_type = xoj::util::DispatchPool<T>;
+
+public:
+    void registerToPool(const std::shared_ptr<pool_type>& newpool) {
```
Technius, line 62: Doc comment should mention that this will remove it from the current pool.

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    bool empty() const { return pool.empty(); }
+
+private:
+    std::vector<listener_type*> pool;
+};
+
+template <class T>
+class Listener {
```
Technius, line 53: Needs a doc comment.

Is my understanding of this correct: Listener is a CRTP style class. An example implementation would be `class MyListener : public Listener<MyListener> { ... }`. Subclasses of Listener are required to implement `on()` (which can potentially be overloaded), which can then be invoked by calling the `DispatchPool<MyListener>::dispatch()` method. The fun part is that because this is templated, we can abuse types like `std::enable_if_t`.

However, it seems like a concrete subclass can only implement `Listener<...>` once. Is this a problem, or do you think the overlays/views will each only be implementing one listener?

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,80 @@
...
+#pragma once
+
+#include <algorithm>
+#include <cassert>
+#include <memory>
+#include <vector>
+
+namespace xoj::util {
+template <class T>
+class DispatchPool final {
```
Technius, line 21: Needs a doc comment.

I initially found the type parameter name `T` to be confusing, until I saw the `using listener_type = T` line below. I would suggest renaming `T` to something more understandable like `ListenerT`.

Technius (commented):
Thanks for the work, this is sorely needed. I think the overall design looks reasonable. I haven't checked everything thoroughly, but I did look over most of the major parts.

* I have a few questions about the DispatchPool/Listener thing, as well as some of the overlay code (see below).
* I mentioned some spots that I think need doc comments, but it would probably be better to defer writing those until the overall design is finalized.

src/core/view/overlays/BaseShapeOrSplineToolView.cpp
```diff
@@ -0,0 +1,84 @@
...
+        BaseStrokeToolView(parent, safeGetStroke(toolHandler)),
+        fillingAlpha(getFillingAlpha(toolHandler)),
+        mask(this->fillingAlpha != 0.0 && safeGetStroke(toolHandler).getToolType() == STROKE_TOOL_HIGHLIGHTER ?
+                     createMask() :
+                     Mask()) {
+    if (mask) {
+        const double* dashes = nullptr;
+        int dashCount = 0;
+        this->lineStyle.getDashes(dashes, dashCount);
+        assert((dashCount == 0 && dashes == nullptr) || (dashCount != 0 && dashes != nullptr));
```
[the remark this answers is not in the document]
bhennion: Yes to everything!

src/core/view/overlays/BaseStrokeToolView.h
bhennion: There will be StrokeToolView, in #4158

src/util/include/util/DispatchPool.h
bhennion: > Is my understanding of this correct: Listener is a CRTP style class. An example implementation would be `class MyListener : public Listener<MyListener> { ... }`. Subclasses of Listener are required to implement `on()` (which can potentially be overloaded), which can then be invoked by calling the `DispatchPool<MyListener>::dispatch()` method. The fun part is that because this is templated, we can abuse types like `std::enable_if_t`.

Yes, you are correct.

> However, it seems like a concrete subclass can only implement `Listener<...>` once. Is this a problem, or do you think the overlays/views will each only be implementing one listener?

If a class needs to be two kinds of listeners, you can do that
```
class A : public Listener<A> {...};  // define virtual functions void on(...); in here
class B : pulbic Listener<B> {...};  // define virtual functions void on(...); in here - types must differ from the ones in A
class C : public A, public B {...};    // overload everything here
```
This ain't a big deal, because both `A` and `B` should be also used separately elsewhere (otherwise, don't split...).

src/core/view/overlays/BaseStrokeToolView.cpp
bhennion: TBH, I have no idea which of cairo's surfaces are stored where. The documentation does not mention this at all.
They also do not mention performance differences. Do you have a source (or benchmarks, or so) that shows a difference?
In an ideal world, I agree that `cairo_surface_create_similar` should give us the best surface for the job...

Performance aside, the downside of `cairo_surface_create_similar` is that we need a cairo_surface_t instance upon creation of the mask.
I guess this can be done, by initializing the mask during the first call to ShapeToolView::draw.

src/core/control/tools/RectangleHandler.cpp
```diff
@@ -4,72 +4,66 @@
...
+    shape.emplace_back(p1);
+    shape.emplace_back(p1.x, p2.y);
+    shape.emplace_back(p2);
+    shape.emplace_back(p2.x, p1.y);
+    shape.emplace_back(p1);
+
+    Range rg(p1.x, p1.y);
+    rg.addPoint(p2.x, p2.y);
+
+    return {shape, rg};
```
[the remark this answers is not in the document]
bhennion: I was counting on some (hypothetical) copy elision kind of compiler optimisation here. I assumed that an `std::move` would block such optimisation.
I can always make the pair before, and emplace points into `first`, but that's definitely less readable.

src/core/control/tools/EllipseHandler.cpp
```diff
@@ -2,95 +2,84 @@
...
+        radiusY = height;
+        center_x = this->startPoint.x;
+        center_y = this->startPoint.y;
+    }
+
+    // set resolution proportional to radius
+    auto npts = static_cast<size_t>(std::round((std::abs(radiusX) + std::abs(radiusY)) * 2.0));
+    if (npts < 24) {
+        npts = 24;  // min. number of points
+    }
```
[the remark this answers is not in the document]
bhennion: Changed by an optimal formula for the number of points.

src/core/view/overlays/BaseStrokeToolView.cpp
bhennion: I just pushed a commit making the masks through `cairo_surface_create_similar` and printing out the surface type in a terminal.
FYI, on my laptop: KDE X11 gives xlib surfaces while KDE Wayland gives image surfaces.

I'd be interested if you get different kind of surfaces depending on the OS, DE and x11 or wayland. To get this information in a terminal, you need to draw a filled highlighter shape (e.g. highlighter, with filling, rectangle).

rolandlo: On Ubuntu 22.04 wayland I get `Creating mask of type: image`

rolandlo: On Ubuntu 22.04 x11 I get `Creating mask of type: xlib`

rolandlo: On MS Windows 10 I get `Creating mask of type: win32`

rolandlo: On MacOS Catalina I get `Creating mask of type: quartz`
This may change when Gtk 3.24.35 is released and updated in `gtk-osx`.

bhennion: Thanks a lot @rolandlo for the data!

I run some benchmarks I found [here](https://openbenchmarking.org/test/pts/cairo-demos) (only relevant for X11, I think), and toyed around. It turns out, the benefits of using `cairo_surface_create_similar` over `cairo_surface_create_similar_image` (on xlib surfaces) are enormous. For instance
```
$./flower-demo-xlib --benchmark
flowers: 125.03 fps
```
vs
```
$./flower-demo-image --benchmark
flowers: 14.42 fps
```
I don't quite know where things get so much slower, probably the under-use of graphics acceleration.

Anyway, this seems to be saying that we need to replace most occurences of `cairo_image_surface_create` by a `cairo_surface_create_similar`, as @Technius was suggesting. It will be pretty complicated to know the kind of surfaces we need in all the places we use image surfaces though.

I'm getting out of the PR's scope :)

bhennion:
Here are the last changes I made:

1. Split off the mask class to a dedicated file. Then, in future PR's, we can use it for every mask or surface buffer (e.g. in StrokeView, PageView, VerticalToolHandler).
2. I removed the mutex. As I mentionned above, there is actually oly one thread involved. The structure is not changed, so putting the mutex back in if we need to later won't be hard at all.

This is ready for another round of review, for whoever feels like it :)

src/core/view/StrokeViewHelper.cpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+double xoj::view::StrokeViewHelper::drawWithPressure(cairo_t* cr, const std::vector<Point>& pts,
+                                                     const LineStyle& lineStyle, double dashOffset) {
+    const double* dashes = nullptr;
+    int dashCount = 0;
+    lineStyle.getDashes(dashes, dashCount);
+    assert((dashCount == 0 && dashes == nullptr) || (dashCount != 0 && dashes != nullptr));
+
+    if (dashes) {
+        for (const auto& [p, q]: PairView(pts)) {
```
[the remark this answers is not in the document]
Technius: Makes sense. Although for the branching issue, the pattern seems simple enough that modern CPU branch predictors should be able to optimize it.

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include <algorithm>
+#include <cassert>
+#include <memory>
+#include <vector>
+
+namespace xoj::util {
+
+/**
+ * @brief Generic template for dispatchers.
```
Technius, line 22:
```suggestion
 * @brief Utility class for dispatching a method call to multiple "listeners".
```

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,111 @@
...
+ * The listeners must implement functions
+ *      void on(Arg&&...);
+ * that the dispatcher will call through dispatch().
+ */
+template <class ListenerT>
+class DispatchPool final {
+public:
+    using listener_type = ListenerT;
+
+    template <typename... Args>
```
Technius, line 33: Needs a doc comment.
```suggestion
    /**
     * @brief Invokes the `on()` method of all registered `ListenerT`s.
     */
    template <typename... Args>
```

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,111 @@
...
+    void dispatch(Args&&... args) const {
+        for (auto* v: this->pool) {
+            v->on(std::forward<Args>(args)...);
+        }
+    }
+
+    void add(listener_type* v) {
```
Technius, line 40:
```suggestion
    /**
     * The listener cannot already be registered.
     */
    void add(listener_type* v) {
```

src/util/include/util/DispatchPool.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief CRTP-style class for listener
+ * Usage:
+ *  class A: Listener<A> {
+ *      (virtual) ~A() { unregisterFromPool(); }
+ *      void on(...);  // Signal receiver
+ *  };
+ *
+ * WARNING: Always unregisterFromPool() in derived class destructor.
```
Technius, line 68: This seems like a pretty big caveat. Would it be a big problem to add a virtual destructor in Listener that automatically does this?

src/core/view/Mask.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <cassert>
+#include <cmath>
+
+#include <cairo.h>
+
+#include "util/Range.h"
+
+using namespace xoj::view;
+
+// #define DEBUG_MASKS
```
Technius, line 12:
```suggestion
```

src/core/view/Mask.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#include <cairo.h>
+
+#include "util/Range.h"
+#include "util/raii/CairoWrappers.h"
+
+namespace xoj::view {
+
+class Mask {
```
Technius, line 21: Needs doc comments.

Thanks for adding this, we could probably simplify several of the tool handlers with this.

In a follow-up: it would also be useful to have the ability to resize the cairo_surface (such as is done in VerticalToolHandler).

src/util/include/util/DispatchPool.h
bhennion: I'd rather not impose a vtable on every listener type.
More importantly, this would open a data race upon destruction:
the base class gets destroyed last, so in a multithreading situation, a function on() of the derived class could be called (by the dispatcher) between the destruction of the derived class and before the destruction of the base (hence the call to unregisterFromPool()).

I don't have a proper solution for this.

src/core/view/Mask.h
bhennion: > In a follow-up: it would also be useful to have the ability to resize the cairo_surface (such as is done in VerticalToolHandler).

Yes I agree. Also in the other tools handlers actually: the tool handlers should listen to zoom or scroll events and adapt their masks accordingly, like the VerticalToolHandler does.

bhennion: Also: This class could be placed in util rather than core/view. Not sure which makes more sense.

src/core/model/Stroke.cpp
```diff
@@ -302,6 +302,22 @@ Point Stroke::getPoint(PathParameter parameter) const {
 
 auto Stroke::getPoints() const -> const Point* { return this->points.data(); }
 
+void Stroke::setPointVector(const std::vector<Point>& other, const Range* const snappingBox) {
+    this->points = other;
+    if (!snappingBox || this->points.empty() || this->points.front().z != Point::NO_PRESSURE) {
+        // We cannot deduce the bounding box from the snapping box if the stroke has pressure values
+        this->sizeCalculated = false;
+    } else {
+        assert(snappingBox->isValid());
```
rolandlo, line 311: This assertion is not satisfied for a horizontal or vertical straight line (drawn with the "Line" drawing type, with grid snapping on) and the app crashes.

bhennion: Thanks for the catch! I'll get back to this PR in a moment.

bhennion:
Hey there!

I cleaned up the includes here and there, and tidied the commit tree. @Technius @Febbe @LittleHuba: would any of you give this another review or do you think/trust it is good to go?

src/core/gui/PageView.h
```diff
@@ -61,12 +61,19 @@ class XojPageView: public LegacyRedrawable, public PageListener, public xoj::vie
...
+     */
+    void drawAndDeleteToolView(xoj::view::ToolView* v, const Range& rg) override;
+
     int getDPIScaling() const override;
     double getZoom() const override;
     Range getVisiblePart() const override;
 
     double getWidth() const override;
     double getHeight() const override;
+    //
```
Technius, line 76: nit: unnecessary comment

src/core/view/Repaintable.h
```diff
@@ -41,5 +42,11 @@ class Repaintable {
      * @brief Flag a region as dirty. Dirty regions will get redrawn at the next screen refresh.
      */
     virtual void flagDirtyRegion(const Range& rg) const = 0;
+
+    /**
+     * @brief Remove a tool view repaint the given range
```
Technius, line 47:
```suggestion
     * @brief Remove a tool view and repaint the given range
```

src/core/view/overlays/ShapeToolView.h
```diff
@@ -43,6 +43,14 @@ class ShapeToolView final: public BaseShapeOrSplineToolView, public xoj::util::L
     } FLAG_DIRTY_REGION = {};
     void on(FlagDirtyRegionRequest, const Range& rg);
 
+    /**
+     * @brief Called before the corresponding BaseShapeHandler's destruction
+     * @param rg The bounding box of the entire stroke
+     */
```
Technius, line 49: Shouldn't the doc comment go on the function? Also, I think the `Finalization` struct itself also needs a doc comment.

src/util/include/util/DispatchPool.h
```diff
@@ -39,6 +39,21 @@ class DispatchPool final {
...
+     * The deleteOn() is supposed to delete the listener
+     * (or at the very least remove any reference to the dispatcher in the listener)
+     */
+    template <typename... Args>
+    void dispatchAndClear(Args&&... args) const {
+        // We cannot iterate on the this->pool: the listener's deletion would trigger calls to DispatchPool::remove()
+        // leading to possible segfaults.
+        auto p = std::move(this->pool);
+        for (auto* v: p) {
+            v->deleteOn(std::forward<Args>(args)...);
```
Technius, line 54: Unless I'm missing something, shouldn't `deleteOn` also be on the list of required methods in the class's doc comment?

Technius (approved):
I assume that you didn't modify anything when you rebased and only looked at the last too commits. LGTM if so, besides one small assert issue + some minor documentation issues.

bhennion:
> I assume that you didn't modify anything when you rebased

That is correct!

I pushed a fixup commit with your comments. Thanks for the review!!

Should we plan on merging this then?

Technius:
Yes, I think this can be merged unless other people have any objections.

bhennion:
Merging this in 5 days, unless someone objects.

PR #4139, opened 2022-07-07
Add legacy snippet to read files containing PageTypeFormat::Copy

### LINKED ISSUE #4127: Unable open files created with nightly version of xournal
I work on windows 10. 
I created files with a January 2021 version of xournal. 
I can no longer open them with the xournalpp-1.1.1+dev-nightly.20220629-windows.exe version, but I can open them with the xournalpp- 1.1.1-windows.exe
Below an example of file
[Correction TD 1 (Exos 1-5).zip](https://github.com/xournalpp/xournalpp/files/9032577/Correction.TD.1.Exos.1-5.zip)

bhennion:
Fixes #4127.


src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -293,7 +293,25 @@ void SaveHandler::writeSolidBackground(XmlNode* background, PageRef p) {
...
+        if (!this->errorMessage.empty()) {
+            this->errorMessage += "\n";
+        }
+        this->errorMessage += _("Page type format is PageTypeFormat::Copy - converted to PageTypeFormat::Plain to "
+                                "avoid corrupted file");
+
+        background->setAttrib("style", PageTypeHandler::getStringForPageTypeFormat(PageTypeFormat::Plain));
+    } else {
+        background->setAttrib("style", PageTypeHandler::getStringForPageTypeFormat(fmt));
+    }
```
Febbe, line 314: Is it even possible, when this is already fixed on import?

bhennion: Yes, because #4142 is still out there (and there might be other sources we don't yet know about).

Febbe: Mhh, actually, `PageTypeFormat` must not contain copy at all to fix this.

bhennion: > Mhh, actually, `PageTypeFormat` must not contain `Copy` at all to fix this.

Agreed, but this can't be done without a refactor of the page background menus.

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -337,6 +337,24 @@ void LoadHandler::parseBgSolid() {
...
 
+    if (bg.format == PageTypeFormat::Copy) {
+        /*
+         * PageTypeFormat::Copy is just a placeholder for the various background related menus, indicating that the
+         * background should be copied from another page.
+         * IT SHOULD NEVER APPEAR IN AN ACTUAL PAGE MODEL OR A FORTIORI SAVED FILE
+         *
+         * Before https://github.com/xournalpp/xournalpp/commit/257b935b3faa61f54a38f79dce734d666f7ab738
+         * a bug was creating files, the pages of which had background type PageTypeFormat::Copy.
+         * Those pages were displayed as PageTypeFormat::Plain.
```
Technius, line 348: I don't think this description is entirely accurate, the bug is still present even after that commit.
```suggestion
         * Due to the bug https://github.com/xournalpp/xournalpp/issues/4127
         * it is possible for some older files to (incorrectly) contain pages with background type PageTypeFormat::Copy.
         * Such pages were displayed as PageTypeFormat::Plain.
```

src/core/control/xojfile/SaveHandler.cpp
Technius: Let's discuss the problem of Copy in #4142.

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -337,6 +337,24 @@ void LoadHandler::parseBgSolid() {
...
+         *
+         * This snippet is legacy code to recover said corrupted files.
+         */
+        g_warning("The opened page is (mildly) corrupted. Fixing background type from PageTypeFormat::Copy to "
+                  "PageTypeFormat::Plain.");
```
Technius, line 353:
```suggestion
        g_warning("The opened page has background type PageTypeFormat::Copy, which should not happen. Converting to "
                  "PageTypeFormat::Plain.");
```

Technius (approved):
LGTM, some minor improvements below.

Technius:
If it's not too troublesome, could you also attach a unit test that checks whether the Copy background type is converted successfully?

bhennion:
> If it's not too troublesome, could you also attach a unit test that checks whether the Copy background type is converted successfully?

Done.

PR #4143, opened 2022-07-08
Add support for layer selection on CLI exports

### LINKED ISSUE #4131: select which layers to export to pdf from command line
When exporting to PDF from the GUI, it is possible to select the layers of interest as visible and these are the only layers that will be exported. This is a great feature for creating incremental illustrations (for example for research papers). It would be great if this functionality was made available from the command line. This way, we could have a single source file and a Makefile that automatically creates all the PDFs that we need.

As a comparison, in Xfig, you can run:

`fig2dev -L pdf -D +/-rangelist input.fig output.pdf`

where D stands for depths (which corresponds to layers in xfig terminology). The `rangelist` argument can be complicated, for example +10,40,55:70,80 keeps in total 18 layers, -15 only removed layer 15.

This is different from the `--export-layers` option for two reasons:
1. we usually want to export a combination of layers onto each illustration (there may be a common layer for example, and some additions)
2. we want to export to single-page PDF files for inclusion of each illustration in, for example, a research paper.

Thanks in advance

bhennion:
Fixes #4131.

This is an old piece of code I had written a while back for exporting specific layers from the CLI. The syntax is not the one asked for in #4131, but it'll have to do :-)


bhennion:
For the syntax, just run `xournalpp --help-export`. If the help is not enough, then it needs more details.

PS: can we update the CI clang-format version?

Febbe:
> can we update the CI clang-format version?

Of course, we can. I have always done this before. But currently I don't have much time.

rolandlo (changes requested):
Nice work! Please make the Lua API aware of the layer range (see comment in the code)). Except of that I only have some nitpicks, mostly related to error messages and documentation comments.

One thing I am uncertain of: If the user uses the --export-no-background option and specifies a layer range that does not intersect with the layer list on one page, should it really export an empty page (current behaviour) or should that page be skipped?

rolandlo:
Forgot to mention: please fix the `PageRangeTest`.

src/core/plugin/luapi_application.h
```diff
@@ -1702,9 +1702,9 @@ static int applib_export(lua_State* L) {
     auto extension = file.extension();
 
     if (extension == ".pdf") {
-        ExportHelper::exportPdf(doc, outputFile, range, bgType, progressiveMode);
+        ExportHelper::exportPdf(doc, outputFile, range, nullptr, bgType, progressiveMode);
     } else if (extension == ".svg" || extension == ".png") {
-        ExportHelper::exportImg(doc, outputFile, range, pngDpi, pngWidth, pngHeight, bgType);
+        ExportHelper::exportImg(doc, outputFile, range, nullptr, pngDpi, pngWidth, pngHeight, bgType);
```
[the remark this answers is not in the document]
bhennion: I tried something. I'm not so sure about how the lua stack works, so could you please check I didn't mess it up?

src/core/control/ExportHelper.h
```diff
@@ -23,6 +23,9 @@ namespace ExportHelper {
  * @param doc Document to export
  * @param output Path to the output file(s)
  * @param range Page range to be parsed. If range=nullptr, exports the whole file
+ * @param layerRange Layer range to be parsed. Will only export those layers, for every exported page.
+ *                  If a number is to high for the number of layers on a given page, it is just ignored.
```
rolandlo, line 27:
```suggestion
 *                  If a number is too high for the number of layers on a given page, it is just ignored.
```

src/core/control/ExportHelper.h
```diff
@@ -32,22 +35,25 @@ namespace ExportHelper {
...
+int exportImg(Document* doc, const char* output, const char* range, const char* layerRange, int pngDpi, int pngWidth,
+              int pngHeight, ExportBackgroundType exportBackground);
 
 /**
  * @brief Export the input file as pdf
  * @param doc Document to export
  * @param output Path to the output file
  * @param range Page range to be parsed. If range=nullptr, exports the whole file
+ * @param layerRange Layer range to be parsed. Will only export those layers, for every exported page.
+ *                  If a number is to high for the number of layers on a given page, it is just ignored.
```
rolandlo, line 47:
```suggestion
 *                  If a number is too high for the number of layers on a given page, it is just ignored.
```

src/core/control/XournalMain.cpp
```diff
@@ -248,6 +248,9 @@ void checkForEmergencySave(Control* control) {
  * @param input Path to the input file
  * @param output Path to the output file(s)
  * @param range Page range to be parsed. If range=nullptr, exports the whole file
+ * @param layerRange Layer range to be parsed. Will only export those layers, for every exported page.
+ *                  If a number is to high for the number of layers on a given page, it is just ignored.
```
rolandlo, line 252:
```suggestion
 *                  If a number is too high for the number of layers on a given page, it is just ignored.
```

src/core/control/XournalMain.cpp
```diff
@@ -257,38 +260,41 @@ void checkForEmergencySave(Control* control) {
...
-    return ExportHelper::exportImg(doc, output, range, pngDpi, pngWidth, pngHeight, exportBackground);
+    return ExportHelper::exportImg(doc, output, range, layerRange, pngDpi, pngWidth, pngHeight, exportBackground);
 }
 
 /**
  * @brief Export the input file as pdf
  * @param input Path to the input file
  * @param output Path to the output file
+ * @param layerRange Layer range to be parsed. Will only export those layers, for every exported page.
+ *                  If a number is to high for the number of layers on a given page, it is just ignored.
```
rolandlo, line 280:
```suggestion
 *                  If a number is too high for the number of layers on a given page, it is just ignored.
```

src/core/plugin/luapi_application.h
```diff
@@ -1673,14 +1673,16 @@ static int applib_export(lua_State* L) {
 
     lua_getfield(L, 1, "outputFile");
     lua_getfield(L, 1, "range");
+    lua_getfield(L, 1, "layerRange");
```
rolandlo, line 1676: Could you please replace `["range"] = "1-2"` by `["layerRange"] = "1-2"` in Example 3 above? Otherwise those writing a plugin might miss that there is an option to export only selected layers.

src/core/plugin/luapi_application.h
```diff
@@ -1673,14 +1673,16 @@ static int applib_export(lua_State* L) {
...
     lua_getfield(L, 1, "background");
     lua_getfield(L, 1, "progressiveMode");
     lua_getfield(L, 1, "pngDpi");
     lua_getfield(L, 1, "pngWidth");
     lua_getfield(L, 1, "dpiHeight");
 
-    const char* outputFile = luaL_optstring(L, -7, nullptr);
-    const char* range = luaL_optstring(L, -6, nullptr);
+    const char* outputFile = luaL_optstring(L, -8, nullptr);
```
rolandlo, line 1683: Please adjust the number in `lua_pop(L, 7)` below. It should be `lua_pop(L,8)`.

rolandlo (changes requested):
Some nitpicks left.
The clang format check failed because of `DocumentView.cpp`. Otherwise LGTM.

bhennion:
All done. Thanks a lor for the review and lua_help!

rolandlo (approved):
The Clang-format test still fails. Is that on purpose? I didn't follow the discussions in #4144.
Except of that (potentially) everything looks fine now.

bhennion:
> The Clang-format test still fails. Is that on purpose? 

Yes it is. It's about the CI's clang-format being outdated. The CI test will come around :)

PR #4144, opened 2022-07-08
update clang-format check

Febbe:


Febbe:
Since clang-14 the git-clang-format module supports return codes. It returns 1 if there is at least one diff which has not been formatted.

bhennion:
> Since clang-14 the git-clang-format module supports return codes. It returns 1 if there is at least one diff which has not been formatted.

So azure-pipelines/util/format_diff_lines.pl can be deleted, right?

Overall, this looks like a sound simpliciation. Did you test the output when clang format has not been applied? (just to make sure)

bhennion (approved):

azure-pipelines/clang-format-applied.yml
```diff
@@ -20,44 +20,16 @@ stages:
...
-        # Apply clang-format, modifying all badly formatted files
-        if [ ! -z "$FILES" ]; then
-            git --no-pager diff -U0 $BASE_COMMIT -- $FILES | perl azure-pipelines/util/format_diff_lines.pl
-        fi
+        git switch $(System.PullRequest.TargetBranch)
+        git switch $(System.PullRequest.SourceBranch)
+        BASE_COMMIT=$(git merge-base $(System.PullRequest.SourceBranch) $(System.PullRequest.TargetBranch))
+        echo $BASE_COMMIT
 
+        git clang-format --diff $BASE_COMMIT --
```
Technius, line 34: `git clang-format` sometimes has a bug where exit code 1 is returned but there are no changes, e.g. stdout of this command is empty (I haven't found a reliable way to reproduce this bug, so I haven't reported it upstream yet). A check is needed for this case.

Source: I ran into this problem while setting up a clang-format check at work a few weeks ago...

Technius (commented):
Thanks for this change.

* We should update the VM image to Ubuntu 20.04, [which already has clang-format 12 installed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#installed-software).
* Path include trigger should be set to `*.c, *.cpp, *.h, *.hpp` to avoid unnecessary runs of the format check.
* See comment below on caveat of `git clang-format`.
* Please attach a sample file (to be removed before merge) to test the format check.

azure-pipelines/clang-format-applied.yml
Febbe: > We should update the VM image to Ubuntu 20.04, [which already has clang-format 12 installed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#installed-software).

This method only works with llvm14+ https://github.com/llvm/llvm-project/commit/edbb8a843c130e60d71cb73e56a33d5ba2cc0ec9 https://github.com/llvm/llvm-project/issues/53220

But we could set the clang-format binary manually.

> Path include trigger should be set to *.c, *.cpp, *.h, *.hpp to avoid unnecessary runs of the format check.

Shouldn't be a problem: I think. git-clang-format and clang-format should already filter them, so that only compatible files are checked. But I can check this.

> Please attach a sample file (to be removed before merge) to test the format check.
Already done.

Febbe: There was a bug regarding the return code: https://github.com/llvm/llvm-project/issues/54758 Not sure if this was the reason.

Febbe: default extensions:   

```
default_extensions = ','.join([
      # From clang/lib/Frontend/FrontendOptions.cpp, all lower case
      'c', 'h',  # C
      'm',  # ObjC
      'mm',  # ObjC++
      'cc', 'cp', 'cpp', 'c++', 'cxx', 'hh', 'hpp', 'hxx',  # C++
      'ccm', 'cppm', 'cxxm', 'c++m',  # C++ Modules
      'cu', 'cuh',  # CUDA
      # Other languages that clang-format supports
      'proto', 'protodevel',  # Protocol Buffers
      'java',  # Java
      'js',  # JavaScript
      'ts',  # TypeScript
      'cs',  # C Sharp
      'json',  # Json
      ])
```
can be filtered with: 

```
The following git-config settings set the default of the corresponding option:
  clangFormat.binary
  clangFormat.commit
  clangFormat.extensions
  clangFormat.style
```

So this might be very straightforward.

Febbe:
@Technius I think you can re-review it: 
 - VM is updated, 
 - triggers updated, 
 - caveat of git clang-format is possible only a bug which is already fixed in the nightly LLVM versions (llvm-14.06 still contain it). I filed a backport request.
 - Sample file attached
 
 I would not restrict users to use clang-format-12, but we could test for all: clang-format 12,14 and latest.

PR #4146, opened 2022-07-09
Add option to disable X11 linking

Technius:
This adds a CMake option to disable linking with X11, enabling builds on some platforms (e.g., Mac M1 builds using the Nix package manager).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,9 +107,9 @@ pkg_check_modules(ExternalModules REQUIRED IMPORTED_TARGET ${pkg-module-spec})
 if (#[[${CMAKE_VERSION} VERSION_LESS "3.18" AND]] ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     # bugfix for old pkg_config versions
     get_target_property(libs PkgConfig::ExternalModules INTERFACE_LINK_LIBRARIES)
-    message(STATUS "PkgConfig::ExternalModules::INTERFACE_LINK_LIBRARIES: ${libs}")
+    # message(STATUS "PkgConfig::ExternalModules::INTERFACE_LINK_LIBRARIES: ${libs}")
     string(REPLACE "-framework;" "-framework " libs "${libs}")
-    message(STATUS "PkgConfig::ExternalModules::INTERFACE_LINK_LIBRARIES fixed: ${libs}")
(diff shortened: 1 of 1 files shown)


Febbe:
This is already implemented via: [`CMAKE_DISABLE_FIND_PACKAGE_<PackageName>`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html) and [CMAKE_REQUIRE_FIND_PACKAGE_<PackageName>](https://cmake.org/cmake/help/latest/variable/CMAKE_REQUIRE_FIND_PACKAGE_PackageName.html#variable:CMAKE_REQUIRE_FIND_PACKAGE_%3CPackageName%3E)

Also, it does not feel good to add so much boilerplate code.

CMakeLists.txt
```diff
@@ -140,6 +139,35 @@ if (GtkSourceView_FOUND)
...
+        message(STATUS "Found X11, checking required dependencies...")
+        if (NOT X11_Xi_FOUND)
+            message(SEND_ERROR "X11 support enabled, but required library libXi not found")
+        endif ()
+        if (NOT X11_Xext_FOUND)
+            message(SEND_ERROR "X11 support enabled, but required library libXext not found")
+        endif ()
+    endif ()
+endif ()
+
```
Febbe, line 170: I would modify the linking, so that the generator expression awaits both libraries, if X11 has been found.

```suggestion
# The following variables are not required, just to show the user that they are there.
set(CMAKE_DISABLE_FIND_PACKAGE_X11 OFF CACHE BOOL "Disable X11")
set(CMAKE_REQUIRE_FIND_PACKAGE_X11 OFF CACHE BOOL "Force X11")
find_package(X11 #[[COMPONENTS Xi Xext X11]]) # component might not work, and must be tested (not documented).
```

Technius:
Thanks for the review.

* I wasn't aware of `CMAKE_DISABLE/REQUIRE_FIND_PACKAGE_*`; thanks for pointing it out. Modified the build to use those two files.
* Added Xi and Xext as components (it didn't seem very obvious from the FindX11.cmake script that they are components). Tested working with my local build.

Febbe:
> Added Xi and Xext as components (it didn't seem very obvious from the FindX11.cmake script that they are components). Tested working with my local build.

Yes, I also didn't know if they work.

Technius:

PR #4149, opened 2022-07-12
Format source files with clang-format

createyourpersonalaccount:
The command used at the root directory was

    find src/                                 \
         -regex '.*\.\(h\|hpp\|c\|cpp\)'      \
         -execdir clang-format -i             \
                               -style=file {} \;

### DIFF
--- a/src/core/audio/AudioQueue.h
+++ b/src/core/audio/AudioQueue.h
@@ -87,14 +87,18 @@ public:
     void waitForProducer(std::unique_lock<std::mutex>& lock) {
         // static_assert(lock.mutex() == &this->queueLock);
         assert(lock.mutex() == &this->queueLock);
-        while (!this->pushNotified && !hasStreamEnded()) { this->pushLockCondition.wait(lock); }
+        while (!this->pushNotified && !hasStreamEnded()) {
+            this->pushLockCondition.wait(lock);
+        }
         this->pushNotified = false;
     }
 
     void waitForConsumer(std::unique_lock<std::mutex>& lock) {
         // static_assert(lock.mutex() == &this->queueLock);
         assert(lock.mutex() == &this->queueLock);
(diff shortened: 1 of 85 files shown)


eldipa:
I like the idea of automating the formatting of the code but as you can see, adding a commit that modifies 85 files may break `git blame`. Worst, as I understand, this was a command that you ran manually: this means that it is a matter of time for the code eventually drifts from the formatted version. 

A safer approach would be:
 - add a `pre-commit` hook that call `clang-format -i` on each commit, so any developer will stick to the formatting
 - add a call to `clang-format` (without `-i`) in the CI system and if a file does not stick to the expected format, fail the build.
 - one of the developers with push access should run `clang-format -i` on all the files and do a commit. Then, write a `.git-blame-ignore-revs` file in the root directory with the full hash of the commit that contains the reformatting. Each developer can configure his/her `git` to read that file and ignore that commit when doing a `git blame`.

createyourpersonalaccount:
> Worst, as I understand, this was a command that you ran manually: this means that it is a matter of time for the code eventually drifts from the formatted version.

The code style is automatically checked by the CI system, I believe I was just formatting old files. Git blame can be fixed by ignoring the commit with [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

Febbe:
There is also the problem, that all merge requests will face conflicts, when they modify the same unformatted section. This is btw. the reason, why we never fixed the format for all files.

eldipa:
> There is also the problem, that all merge requests will face conflicts, when they modify the same unformatted section. This is btw. the reason, why we never fixed the format for all files.

Good point.

rolandlo:
This PR is out of sync and contested. Closing therefore.

PR #4153, opened 2022-07-17
mac: only initialize env vars to paths that exist

Technius:
The environment setup logic forcefully overrides the environment variables used
by dependencies (e.g., GDK_PIXBUF_MODULE_FILE) to paths assumed to conform to
an installed version of Xournal++. As a result, variables like
GDK_PIXBUF_MODULE_FILE may point to files that don't exist, resulting in a
startup crash.

This commit fixes the following issues:

* Detection for the bundle file has been changed to also check for the
  "Contents" folder name. Only checking for "Resources" is insufficient since
  the repository contains a folder named "resources" at the top level.

* Environment variables pointing to a path are only set if the path exists.


src/osx/setup-env.cpp
```diff
@@ -28,14 +52,16 @@ void setupEnvironment() {
...
+    // Override any environment variables corresponding to install prefix paths.
+    // Only do this if the paths don't exist, otherwise we could run into issues
+    // (e.g., nonexistent pixbuf module file will lead to a startup crash)
+    setenvPathIfExists("XDG_CONFIG_DIRS", xdgPath.string().c_str(), 1);
+    setenvPathIfExists("XDG_DATA_DIRS", dataPath.string().c_str(), 1);
+    setenvPathIfExists("GTK_DATA_PREFIX", base.string().c_str(), 1);
+    setenvPathIfExists("GTK_EXE_PREFIX", base.string().c_str(), 1);
+    setenvPathIfExists("GTK_PATH", base.string().c_str(), 1);
+    setenvPathIfExists("GTK_IM_MODULE_FILE", imModuleFile.string().c_str(), 1);
+    setenvPathIfExists("GDK_PIXBUF_MODULE_FILE", pixbufModuleFile.string().c_str(), 1);
```
Technius, line 64: We should probably log these paths to avoid debug issues later down the line.

src/osx/setup-env.cpp
```diff
@@ -9,17 +9,41 @@
 
 #include "filesystem.h"
 
+namespace {
+void setenvPathIfExists(const char* name, fs::path const& path, bool override) {
+    if (fs::exists(path)) {
+        setenv(name, path.c_str(), 1);
```
rolandlo, line 15:
```suggestion
        setenv(name, path.c_str(), override);
```

src/osx/setup-env.cpp
```diff
@@ -9,17 +9,41 @@
...
     auto base = Stacktrace::getExePath().parent_path();  // Xournal++.app/Contents or $HOME/gtk/inst
 
-    if (fs::exists(base / "Resources")) {  // app-bundle
+    /* We consider the application to launch from a bundle iff the parent
+     * directory is named "Content" and it has a "Resources" folder.
+     * Note that checking the latter only is insufficient, e.g. it will
+     * falsely detect a bundle from the build folder, since the repository root
+     * has a folder named "resources"
+     */
+    const bool isInBundle = base.filename() == "Content" && fs::exists(base / "Resources");
```
rolandlo, line 36:
```suggestion
     * directory is named "Contents" and it has a "Resources" folder.
     * Note that checking the latter only is insufficient, e.g. it will
     * falsely detect a bundle from the build folder, since the repository root
     * has a folder named "resources"
     */
    const bool isInBundle = base.filename() == "Contents" && fs::exists(base / "Resources");
```

src/osx/setup-env.cpp
```diff
@@ -9,17 +9,41 @@
...
+
+}  // namespace
 
 void setupEnvironment() {
     /**
      * see https://gitlab.gnome.org/GNOME/gtk-mac-bundler/-/blob/master/examples/gtk3-launcher.sh
      * and https://gitlab.gnome.org/GNOME/gtk-mac-bundler/-/issues/12
      */
+
+    // Get the folder containing the application executable
```
rolandlo, line 27:
```suggestion
    // Get the parent folder of the folder containing the application executable
```

rolandlo (commented):
I don't have my MacBook with me right now, so I can't test the PR. So please give me time until Saturday for testing.
For clarity: does the PR allow to run from the build folder without `make install`? Is it run via `./xournalpp`  or via `jhbuild run ./xournalpp` then?

The logic for checking if the app is running from an app bundle should also be changed in `Util::getDataPath()` (inside `PathUtil.cpp`). 

There is one more bug (relying on a false assumption) further down (my fault):
`   lang.replace(lang.find("-"), 1, "_");  // e.g. "de_DE"` results in a crash, if `lang` does not contain `-`. I assumed that it will always be of the form `ab-CD`, but it can also just be of the form `ab` (language without country code). We should fix that, too.

Technius:
@rolandlo I've haven't tested the application bundle build, so your help testing it would be greatly appreciated.

>  For clarity: does the PR allow to run from the build folder without `make install`?

Yes, the PR fixes a crash that would've happened if trying to run from the build folder.

> The logic for checking if the app is running from an app bundle should also be changed in `Util::getDataPath()` (inside `PathUtil.cpp`).

Thanks for the catch, fixed.

> There is one more bug (relying on a false assumption) further down (my fault): ` lang.replace(lang.find("-"), 1, "_"); // e.g. "de_DE"` results in a crash, if `lang` does not contain `-`. I assumed that it will always be of the form `ab-CD`, but it can also just be of the form `ab` (language without country code). We should fix that, too.

Also fixed.

src/osx/setup-env.cpp
```diff
@@ -28,14 +58,22 @@ void setupEnvironment() {
...
-    setenv("XDG_CONFIG_DIRS", xdgPath.string().c_str(), 1);
-    setenv("XDG_DATA_DIRS", dataPath.string().c_str(), 1);
-    setenv("GTK_DATA_PREFIX", base.string().c_str(), 1);
-    setenv("GTK_EXE_PREFIX", base.string().c_str(), 1);
-    setenv("GTK_PATH", base.string().c_str(), 1);
+    // Override any environment variables corresponding to install prefix paths.
+    // Only do this if the paths don't exist, otherwise we could run into issues
+    // (e.g., nonexistent pixbuf module file will lead to a startup crash)
+    setenvPathIfExists("XDG_CONFIG_DIRS", xdgPath, 1);
+    setenvPathIfExists("XDG_DATA_DIRS", dataPath, 1);
```
Technius, line 65: The behavior of overriding (instead of prepending) is too aggressive in the use case I am interested in (xournalpp with Nix), since `XDG_DATA_DIRS` may contain the directories with the system themes (e.g., Adwaita and hicolor theme).

For example, xournalpp logs the following on startup:
```
Skip override env var XDG_CONFIG_DIRS, path doesn't exist: "/nix/store/j4b0xb53icgpa60pybfvdpl8kd1w6mlq-xournalpp-1.1.1/etc/xdg"
Overriding env var XDG_DATA_DIRS: "/nix/store/j4b0xb53icgpa60pybfvdpl8kd1w6mlq-xournalpp-1.1.1/share"
```

However, my `XDG_DATA_DIRS` already consists of:
```
/nix/store/j4b0xb53icgpa60pybfvdpl8kd1w6mlq-xournalpp-1.1.1/share:/nix/store/iiiib5v00l61fbf7fp12hnd2nfkp2kf9-hicolor-icon-theme-0.17/share:/nix/store/f0j672ahrxq9ag0ldq3w771l40fr90ls-adwaita-icon-theme-42.0/share:
```

So for these "DIRS" variables, it would be better to prepend the paths. The downside is that the user may have potentially already set one of these variables, which can cause confusion.

src/osx/setup-env.cpp
```diff
@@ -9,17 +9,47 @@
...
+    const bool isInBundle = base.filename() == "Contents" && fs::exists(base / "Resources");
+    if (isInBundle) {
+        // app-bundle, e.g. Xournal++.app/Contents
+        // Install prefix is actually the Resources folder
         base = base / "Resources";
-    }  // Now base is Xournal++.app/Contents/Resources or $HOME/gtk/inst
+        std::cerr << "xournalpp is running from a bundle\n";
+    } else {
+        // Assume the "base dir" is the install prefix, corresponding to
+        // e.g., build folder or $HOME/gtk/inst
```
Technius, line 49: I just realized there's a case missing here. To summarize:
* the bundle has the executable at `Contents/bin/xournalpp`.
* in master branch, a source build has the executable located at `$build/xournalpp.`
* (The missing case) for release-1.1, the executable in the build folder is placed at `$build/src/xournalpp`

So maybe a special case check for `src` as a parent name is also needed here.

Technius: We could also only set `XDG_DATA_DIRS` if the application is in a bundle, since the bundle should include any themes that may be needed anyways.

rolandlo: > I just realized there's a case missing here. To summarize:
> 
> * the bundle has the executable at `Contents/bin/xournalpp`.
> * in master branch, a source build has the executable located at `$build/xournalpp.`
> * (The missing case) for release-1.1, the executable in the build folder is placed at `$build/src/xournalpp`
> 
> So maybe a special case check for `src` as a parent name is also needed here.

I agree, that there is a missing case for release-1.1. Instead of `Contents/bin/xournalpp` you mean `Contents/MacOS/xournalpp`. It doesn't matter though.

Currently running the PR via `jhbuild run src/xournalpp` from within the build folder results in the messages:
```
xournalpp is running from some non-bundle path
Skip override env var XDG_CONFIG_DIRS, path doesn't exist: "/Users/xournal-dev/xournalpp/build/etc/xdg"
Skip override env var XDG_DATA_DIRS, path doesn't exist: "/Users/xournal-dev/xournalpp/build/share"
Skip override env var GTK_IM_MODULE_FILE, path doesn't exist: "/Users/xournal-dev/xournalpp/build/lib/gtk-3.0/3.0.0/immodules.cache"
Skip override env var GDK_PIXBUF_MODULE_FILE, path doesn't exist: "/Users/xournal-dev/xournalpp/build/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache"
** Message: 16:51:11.404: Continue with GDK_PIXBUF_MODULE_FILE = (null)
** Message: 16:51:11.408: Setting LANG and LC_MESSAGES to de_CH.UTF-8
** Message: 16:51:11.613: TEXTDOMAINDIR = (null), Platform-specific locale dir = /Users/xournal-dev/xournalpp/build/share/locale, chosen directory = /Users/xournal-dev/xournalpp/build/share/locale
** Message: 16:51:11.649: Plugin "MigrateFontSizes" UI initialized

However, even if you remove the `/src` from those paths, they don't exist. So the approach seems flawed. I think if the app is running from a local build, it should still use the prefix `$HOME/gtk/inst`. 
```

Technius: Does `jhbuild` set any environment variables when it runs the command?

rolandlo: Yes, it sets a ton of environment variables. This can already been seen by executing `jhbuild run env` versus `env`.
In particular `XDG_DATA_DIRS`, `XDG_CONFIG_DIRS` are set and the `PATH` has several additions. It's 81 variables vs 19 variables in one test.

Technius:
For future reference, here are most of the variables set by jhbuild: https://gitlab.gnome.org/GNOME/jhbuild/-/blob/master/jhbuild/environment.py

We'll want to make sure we set the relevant ones in the launcher function.

rolandlo:
Closing in favor of #7358

PR #4158, opened 2022-07-21
Split tools/StrokeHandler into View-Controller

bhennion:
Follows #4137.

This PR splits tools/StrokeHandler into a View and a Controller. It also optimizes the rendering costs ~~and protects the data with a mutex (see #3579)~~.

- [x] Merge #4137 and rebase

~~**Edit:** the CI failing is due to GCC 8 not handling `std::transform_reduce`. Will be fixed with #4160~~   done

### DIFF
--- a/src/core/control/tools/SplineHandler.h
+++ b/src/core/control/tools/SplineHandler.h
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <memory>    // for unique_ptr
 #include <optional>  // for optional
 #include <vector>    // for vector
 
@@ -22,6 +23,7 @@
 #include "model/Point.h"     // for Point
 #include "util/Rectangle.h"  // for Rectangle
 #include "view/StrokeView.h"
+#include "view/overlays/OverlayView.h"
 
 #include "InputHandler.h"            // for InputHandler
 #include "SnapToGridInputHandler.h"  // for SnapToGridInputHandler
@@ -29,6 +31,11 @@
 class PositionInputData;
 class LegacyRedrawable;
 class XournalView;
+
+namespace xoj::view {
+class OverlayView;
+class Repaintable;
+}  // namespace xoj::view
 
 /**
  * @brief A class to handle splines
@@ -50,6 +57,10 @@ public:
     ~SplineHandler() override;
 
     void draw(cairo_t* cr) override;
+
+    std::unique_ptr<xoj::view::OverlayView> createView(xoj::view::Repaintable* parent) const override {
+        return nullptr;
+    }
 
     void onSequenceCancelEvent() override;
     bool onMotionNotifyEvent(const PositionInputData& pos, double zoom) override;
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -19,10 +19,6 @@
 #include "control/shaperecognizer/ShapeRecognizer.h"  // for ShapeRecognizer
 #include "control/tools/InputHandler.h"               // for InputHandler::P...
 #include "control/tools/SnapToGridInputHandler.h"     // for SnapToGridInput...
-#include "control/zoom/ZoomControl.h"
-#include "gui/MainWindow.h"
-#include "gui/PageView.h"                        // for XojPageView
-#include "gui/XournalView.h"                     // for XournalView
 #include "gui/inputdevices/PositionInputData.h"  // for PositionInputData
 #include "model/Document.h"                      // for Document
 #include "model/Layer.h"                         // for Layer
@@ -32,11 +28,12 @@
 #include "undo/InsertUndoAction.h"               // for InsertUndoAction
 #include "undo/RecognizerUndoAction.h"           // for RecognizerUndoA...
 #include "undo/UndoRedoHandler.h"                // for UndoRedoHandler
-#include "util/Color.h"                          // for cairo_set_sourc...
+#include "util/DispatchPool.h"                   // for DispatchPool
 #include "util/Range.h"                          // for Range
 #include "util/Rectangle.h"                      // for Rectangle, util
-#include "view/StrokeView.h"                     // for StrokeView, Str...
-#include "view/View.h"                           // for Context
+#include "view/overlays/StrokeToolFilledHighlighterView.h"  // for StrokeToolFilledHighlighterView
+#include "view/overlays/StrokeToolFilledView.h"             // for StrokeToolFilledView
+#include "view/overlays/StrokeToolView.h"                   // for StrokeToolView
 
 #include "StrokeStabilizer.h"  // for Base, get
 
@@ -44,47 +41,15 @@ using namespace xoj::util;
 
 guint32 StrokeHandler::lastStrokeTime;  // persist for next stroke
 
-StrokeHandler::StrokeHandler(Control* control, XojPageView* pageView, const PageRef& page):
+StrokeHandler::StrokeHandler(Control* control, const PageRef& page):
         InputHandler(control, page),
         snappingHandler(control->getSettings()),
         stabilizer(StrokeStabilizer::get(control->getSettings())),
-        pageView(pageView) {}
+        viewPool(std::make_shared<xoj::util::DispatchPool<xoj::view::StrokeToolView>>()) {}
 
-void StrokeHandler::draw(cairo_t* cr) {
-    assert(stroke && stroke->getPointCount() > 0);
-
-    auto setColorAndBlendMode = [stroke = this->stroke.get(), cr]() {
-        if (stroke->getToolType() == StrokeTool::HIGHLIGHTER) {
-            if (auto fill = stroke->getFill(); fill != -1) {
-                Util::cairo_set_source_rgbi(cr, stroke->getColor(), static_cast<double>(fill) / 255.0);
-            } else {
-                Util::cairo_set_source_rgbi(cr, stroke->getColor(), xoj::view::StrokeView::OPACITY_HIGHLIGHTER);
-            }
-            cairo_set_operator(cr, CAIRO_OPERATOR_MULTIPLY);
-        } else {
-            Util::cairo_set_source_rgbi(cr, stroke->getColor());
-            cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
-        }
-    };
-
-    if (this->mask) {
-        setColorAndBlendMode();
-        cairo_mask_surface(cr, mask->surf, 0, 0);
-    } else {
-        if (this->stroke->getPointCount() == 1) {
-            // drawStroke does not handle single dots
-            const Point& pt = this->stroke->getPoint(0);
-            double width = this->stroke->getWidth() * (this->hasPressure ? pt.z : 1.0);
-            setColorAndBlendMode();
-            this->paintDot(cr, pt.x, pt.y, width);
-        } else {
-            strokeView->draw(xoj::view::Context::createDefault(cr));
-        }
-    }
-}
+StrokeHandler::~StrokeHandler() = default;
 
 auto StrokeHandler::onKeyEvent(GdkEventKey* event) -> bool { return false; }
-
 
 auto StrokeHandler::onMotionNotifyEvent(const PositionInputData& pos, double zoom) -> bool {
     if (!stroke) {
@@ -103,9 +68,12 @@ auto StrokeHandler::onMotionNotifyEvent(const PositionInputData& pos, double zoo
     return true;
 }
 
-void StrokeHandler::paintTo(const Point& point) {
+void StrokeHandler::paintTo(Point point) {
+    if (this->hasPressure && point.z > 0.0) {
+        point.z *= this->stroke->getWidth();
+    }
 
-    int pointCount = stroke->getPointCount();
+    auto pointCount = stroke->getPointCount();
 
     if (pointCount > 0) {
         Point endPoint = stroke->getPoint(pointCount - 1);
@@ -113,16 +81,8 @@ void StrokeHandler::paintTo(const Point& point) {
         if (distance < PIXEL_MOTION_THRESHOLD) {  //(!validMotion(point, endPoint)) {
             if (pointCount == 1 && this->hasPressure && endPoint.z < point.z) {
                 // Record the possible increase in pressure for the first point
-                // Nb: at this point, neither point.z nor endPoint.z has been multiplied by this->stroke->getWidth()
                 this->stroke->setLastPressure(point.z);
-                this->firstPointPressureChange = true;
-
-                double width = this->stroke->getWidth() * point.z;
-                if (mask) {
-                    this->paintDot(mask->cr, endPoint.x, endPoint.y, width);
-                }
-                // Trigger a call to `draw`. If mask == nullopt, the `paintDot` is called in `draw`
-                this->pageView->repaintRect(endPoint.x - 0.5 * width, endPoint.y - 0.5 * width, width, width);
+                this->viewPool->dispatch(xoj::view::StrokeToolView::THICKEN_FIRST_POINT_REQUEST, point.z);
             }
             return;
         }
@@ -130,34 +90,25 @@ void StrokeHandler::paintTo(const Point& point) {
             /**
              * Both device and tool are pressure sensitive
              */
-            if (this->firstPointPressureChange) {
-                // Avoid shrinking if we recorded a higher pressure event at the beginning of the stroke
-                this->firstPointPressureChange = false;
-                stroke->setLastPressure(std::max(endPoint.z, point.z) * stroke->getWidth());
-            } else {
-                if (const double widthDelta = (point.z - endPoint.z) * stroke->getWidth();
-                    -widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
-                    /**
-                     * If the width variation is to big, decompose into shorter segments.
-                     * Those segments can not be shorter than PIXEL_MOTION_THRESHOLD
-                     */
-                    double nbSteps = std::min(std::ceil(std::abs(widthDelta) / MAX_WIDTH_VARIATION),
-                                              std::floor(distance / PIXEL_MOTION_THRESHOLD));
-                    double stepLength = 1.0 / nbSteps;
-                    Point increment((point.x - endPoint.x) * stepLength, (point.y - endPoint.y) * stepLength,
-                                    widthDelta * stepLength);
-                    endPoint.z *= stroke->getWidth();
-                    endPoint.z += increment.z;
-                    stroke->setLastPressure(endPoint.z);
+            if (const double widthDelta = point.z - endPoint.z;
+                - widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
+                /**
+                 * If the width variation is to big, decompose into shorter segments.
+                 * Those segments can not be shorter than PIXEL_MOTION_THRESHOLD
+                 */
+                double nbSteps = std::min(std::ceil(std::abs(widthDelta) / MAX_WIDTH_VARIATION),
+                                          std::floor(distance / PIXEL_MOTION_THRESHOLD));
+                double stepLength = 1.0 / nbSteps;
+                Point increment((point.x - endPoint.x) * stepLength, (point.y - endPoint.y) * stepLength,
+                                widthDelta * stepLength);
+                endPoint.z += increment.z;
 
-                    for (int i = 1; i < static_cast<int>(nbSteps); i++) {  // The last step is done below
-                        endPoint.x += increment.x;
-                        endPoint.y += increment.y;
-                        endPoint.z += increment.z;
-                        drawSegmentTo(endPoint);
-                    }
+                for (int i = 1; i < static_cast<int>(nbSteps); i++) {  // The last step is done below
+                    endPoint.x += increment.x;
+                    endPoint.y += increment.y;
+                    endPoint.z += increment.z;
+                    drawSegmentTo(endPoint);
                 }
-                stroke->setLastPressure(point.z * stroke->getWidth());
             }
         }
     }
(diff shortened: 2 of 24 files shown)


bhennion:
Now that #4137 has been merged, this is up for review.

bhennion:
Rebased and resolved conflicts

bhennion:
Resolved conflicts.

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -101,61 +66,47 @@ auto StrokeHandler::onMotionNotifyEvent(const PositionInputData& pos, double zoo
...
-                                    widthDelta * stepLength);
-                    endPoint.z *= stroke->getWidth();
+            if (const double widthDelta = point.z - endPoint.z;
+                - widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
+                /**
+                 * If the width variation is to big, decompose into shorter segments.
+                 * Those segments can not be shorter than PIXEL_MOTION_THRESHOLD
+                 */
+                double nbSteps = std::min(std::ceil(std::abs(widthDelta) / MAX_WIDTH_VARIATION),
+                                          std::floor(distance / PIXEL_MOTION_THRESHOLD));
```
Technius, line 98: Should add a comment explaining that `nbSteps >= 2.0` because `distance >= PIXEL_MOTION_THRESHOLD` and `abs(widthDelta) > MAX_WIDTH_VARIATION`. This fact was not very clear to me on the first reading.

Technius (commented):
I only read the code, but largely LGTM. Will test it out tomorrow.

src/core/control/tools/SplineHandler.h
```diff
@@ -51,6 +58,10 @@ class SplineHandler: public InputHandler {
 
     void draw(cairo_t* cr) override;
 
+    std::unique_ptr<xoj::view::OverlayView> createView(xoj::view::Repaintable* parent) const override {
+        return nullptr;
```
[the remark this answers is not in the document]
bhennion: The SplineHandler specialization is implemented with #4159.

src/core/control/tools/StrokeHandler.cpp
bhennion: No, nbSteps could still be 1 (e.g. if distance == 1.1 * PIXEL_MOTION_THRESHOLD), but this is not a problem: the last (and only) point is added by the call to `drawSegmentTo` at the end of the function.

bhennion: Added a comment

bhennion:
@Technius Have you found time to try this out?

Technius:
Based on some light testing, when using a filled stroke with the stroke recognizer, the fill remains as the pre-recognized stroke fill and not the final stroke fill. The stroke lines correctly disappear but the fill (outside of the recognized stroke) remain until the next repaint of that region. I found this easiest to reproduce with circles, but squares also work.

I didn't notice any other obvious problem.

bhennion:
Thanks for the testing! I am sorry I let something that obvious slip through my check.

It is fixed now

https://user-images.githubusercontent.com/17818041/210957695-6d063c02-8570-4c54-ba8d-ec88a3b17af4.mp4

src/core/control/tools/StrokeHandler.cpp
Technius: Right, since the operation is `min`, your example would make `nbSteps == 1`.

Technius (approved):

bhennion:
Thanks @Technius for the review and testing!

Merging in 72 hours unless an objection is raised.

bhennion:
I can reproduce the `hasNoViewOf` failing assertion when writing between pages. Let's see if I can find a reasonable fix.

The invalid range one, I haven't encountered. @rolandlo were you doing something special when it happened? Do you know if your ongoing stroke was big/small/just starting/just finished?

bhennion:
> I can reproduce the `hasNoViewOf` failing assertion when writing between pages. Let's see if I can find a reasonable fix.

Here is a fix. Ideally, one would be able to paint across page boundaries, but that'd be more work (and it does not work on master or on the release branch either).

rolandlo:
> I can reproduce the `hasNoViewOf` failing assertion when writing between pages. Let's see if I can find a reasonable fix.
> 
> The invalid range one, I haven't encountered. @rolandlo were you doing something special when it happened? Do you know if your ongoing stroke was big/small/just starting/just finished?

I only remember going through the different drawing types and testing them out. Nothing extraordinary. The stroke may have just finished, but I'm not sure. I was rather surprised when the window went away (below was the VS Code window). I will simply try to reproduce the failing assertion.

bhennion:
@hrdl-github It's okay, I can reproduce under Wayland, with "Merge button events with stylus tip events" checked, but only with one of the two buttons on my stylus.
Do you have "Merge button events with stylus tip events" checked?

A rapid with the master branch tells me the bug was there before, but the new assert is catching it: in the following sequence, I touched the screen with the stylus, pressed button 1 then released it, pressed button 2 then released it. 
I just logged the calls to PenInputHandler::actionStart and actionEnd.
```
Action start        <-- tip touch
Action end          <-- Button 1 pressed
Action start         <-- Button 1 pressed
Action end          <-- Button 1 released
Action start        <-- Button 1 released
Action start        <-- Button 2 pressed
Action end         <-- Button 2 released
Action end         <-- tip lifted from the screen
```
Now, there should never ever be two actions started at the same time (the input handler is not made to handle two parallel actions).

rolandlo:
> Now, there should never ever be two actions started at the same time (the input handler is not made to handle two parallel actions).

This is reported in #4377.

bhennion:
> This is reported in https://github.com/xournalpp/xournalpp/issues/4377.

Oh right! I forgot, thanks !

bhennion:
> > Do you have "Merge button events with stylus tip events" checked?
> 
> I did not -- enabling this option prevents xournalpp from crashing on the built-in digitiser on [5e78bdb](https://github.com/xournalpp/xournalpp/pull/4158/commits/5e78bdbd75aceb52acba6cd0b567d28c2844430d).
> Does this make sense?

TBH, nothing makes sense about this option... On my device, on Wayland, the option should be checked for one button to work, and unchecked for the other to work. (and if it's checked, the other button makes the assertion fail)
This option would need a very careful refactor, but it's not an easy task, seeing the variety of devices out there and the various OS's (Windows, Mac, Linux X, Linux Wayland). This goes way beyond this PR.

I might just uglily replace the assertion by a safety check that removes preexisting views for this PR.

bhennion:
Yay!! I finally managed to reproduce @rolandlo's crash!
You can actually get an invalid Range if you're drawing on a page that is not visible :-) Yes, this is possible!

- Open a new file. Add a new page.
- Check that the top of the second page is right under the bottom of the toolbar. The bottom of the first page is not visible, "hidden" by the toolbar.
- Start drawing on the second page, and go to the toolbar while keeping your stylus tip on the screen (or your mouse button pressed).
- See the crash, with the proper assert failing.

bhennion:
> Yay!! I finally managed to reproduce @rolandlo's crash! You can actually get an invalid Range if you're drawing on a page that is not visible :-) Yes, this is possible!
> 
> * Open a new file. Add a new page.
> * Check that the top of the second page is right under the bottom of the toolbar. The bottom of the first page is not visible, "hidden" by the toolbar.
> * Start drawing on the second page, and go to the toolbar while keeping your stylus tip on the screen (or your mouse button pressed).
> * See the crash, with the proper assert failing.

Except this does not cause a crash with the version @rolandlo got his crash. This has been introduced with 5e78bdb.

I really don't see how you got your crash then... (Still, I need to fix the above now)

rolandlo:
@bhennion I tried quite hard to reproduce the crash I reported, but failed. Maybe ignore it for now.

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -291,25 +219,23 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
...
+    doc->lock();
+    layer->addElement(stroke.get());
+    doc->unlock();
 
-    Stroke* s = stroke.release();
-    layer->addElement(s);
-    page->fireElementChanged(s);
+    // Blitt the stroke to the page's buffer and delete all views.
+    // Passing the empty Range() as no actual redrawing is necessary at this point
+    this->viewPool->dispatchAndClear(xoj::view::StrokeToolView::FINALIZATION_REQUEST, Range());
```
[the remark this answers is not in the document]
bhennion: We don't need to pass on a Range. The rerendered area should just be the Stroke's bounding box.
I am currently thinking of various cases where special rerendering would be required at the end of a Stroke sequence:
- If part of the stroke is outside the visible area, then the mask is too small and naively blitting it to the page buffer is not enough.
- As you say, if the stroke is not on the top layer.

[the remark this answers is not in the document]
bhennion: See e40bd19

bhennion:
I think I fixed all the reported issues (most likely including @rolandlo's slippery crash). This can be tested once more.

bhennion:
Thanks for the catch. I'm done trying to fix the input system in this PR, so I just replaced the assertion with a safety check removing all the existing views. That's not beautiful but it should work.

src/core/gui/PageView.cpp
```diff
@@ -342,42 +336,44 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
-            assert(hasNoViewOf(overlayViews, inputHandler));
+            /**
+             * Due to https://github.com/xournalpp/xournalpp/issues/4377
+             * some devices under some configurations can start an action while another one has already started
+             * This is a workaround to avoid mem leaks and segfaults
+             * This `if` statement can probably be replaced by an `assert` once #4377 is fixed
+             */
+            g_warning("InputHandler already exists upon XojPageView::onButtonPressEvent. Deleting it");
+            std::remove_if(overlayViews.begin(), overlayViews.end(),
+                           [ih = this->inputHandler](auto& v) { return v->isViewOf(ih); });
```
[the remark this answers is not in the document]
bhennion: Ah right! I keep forgetting std::remove does not erase stuff... Thanks a lot for the catch and the debugging!

src/core/gui/PageView.cpp
```diff
@@ -664,7 +659,16 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
...
-            assert(hasNoViewOf(overlayViews, inputHandler));
+            if (!hasNoViewOf(overlayViews, inputHandler)) {
+                /*
+                 * This is a workaround to avoid mem leaks and segfaults
+                 * This `if` statement can probably be replaced by an `assert` once #4377 is fixed
+                 */
+                g_warning("InputHandler still has a View before deletion in XojPageView::onButtonReleaseEvent. "
+                          "Deleting it");
+                std::remove_if(overlayViews.begin(), overlayViews.end(),
+                               [ih = this->inputHandler](auto& v) { return v->isViewOf(ih); });
```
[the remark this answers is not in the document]
bhennion: Actually, just to be safe, we need to erase all possible matches:
```suggestion
                overlayViews.erase(std::remove_if(overlayViews.begin(), overlayViews.end(),
                               [ih = this->inputHandler](auto& v) { return v->isViewOf(ih); }), overlayViews.end());
```
I created a static function `eraseViewsOf` doing just that, to avoid making the same mistake again and again.

bhennion:
I also reverted https://github.com/xournalpp/xournalpp/commit/0159d076e0d6d627bb134d2d21e57ee2987dbe06 as you suggested. My silly mistake was most likely causing the issue.

bhennion:
I rebased and cleaned up the commit tree.

I'll be merging this as is in 7 days, unless an objection is raised.

rolandlo (commented):
Sorry for the late reply, but I was pretty busy the last couple of weeks.

Codewise I can't see a flaw. I have tested it without issues on Ubuntu 22.10 (both Wayland and X11). However testing it now on my MacBook (with Catalina) I got an assertion failure when drawing filled highlighter strokes rapidly. For filled highlighter strokes it is quite a bit slower than Linux, but my MacBook is also quite a bit older than the Lenovo ThinkPad I use on Linux.

```term
Assertion failed: (isInitialized()), function wipe, file /Users/admin/xournalpp/src/core/view/Mask.cpp, line 85.

** (xournalpp:2820): WARNING **: 20:56:17.257: [Crash Handler] Crashed with signal 6

** (xournalpp:2820): WARNING **: 20:56:17.258: [Crash Handler] Wrote crash log to: /Users/admin/.cache/xournalpp/errorlogs/errorlog.20230209-205617.log
```
Here's the error log:

```term
[bt] #1 1   xournalpp                           0x000000010e8b24f2 _ZL12crashHandleri + 1874
[bt] #2 2   libsystem_platform.dylib            0x00007fff6aa075fd _sigtramp + 29
[bt] #3 3   ???                                 0x0000000300000000 0x0 + 12884901888
[bt] #4 4   libsystem_c.dylib                   0x00007fff6a8dd808 abort + 120
[bt] #5 5   libsystem_c.dylib                   0x00007fff6a8dcac6 err + 0
[bt] #6 6   xournalpp                           0x000000010ec37aea _ZN3xoj4view4Mask4wipeEv + 74
[bt] #7 7   xournalpp                           0x000000010ec5594b _ZNK3xoj4view31StrokeToolFilledHighlighterView4drawEP6_cairo + 251
[bt] #8 8   xournalpp                           0x000000010ea59d48 _ZN11XojPageView9paintPageEP6_cairoP20_cairo_rectangle_int + 984
[bt] #9 9   xournalpp                           0x000000010eb4ba4c _ZL16gtk_xournal_drawP10_GtkWidgetP6_cairo + 1004
[bt] #10 10  libgtk-3.0.dylib                    0x000000011080abb5 gtk_widget_draw_internal + 421
[bt] #11 11  libgtk-3.0.dylib                    0x0000000110591181 gtk_container_propagate_draw + 625
[bt] #12 12  libgtk-3.0.dylib                    0x00000001105918e3 gtk_container_draw + 227
[bt] #13 13  libgtk-3.0.dylib                    0x00000001106f55c6 _gtk_pixel_cache_draw + 1430
[bt] #14 14  libgtk-3.0.dylib                    0x0000000110802988 gtk_viewport_render + 232
[bt] #15 15  libgtk-3.0.dylib                    0x00000001105998b2 gtk_css_custom_gadget_draw + 146
[bt] #16 16  libgtk-3.0.dylib                    0x000000011059edc7 gtk_css_gadget_draw + 1447
[bt] #17 17  libgtk-3.0.dylib                    0x0000000110801e56 gtk_viewport_draw + 134
[bt] #18 18  libgtk-3.0.dylib                    0x000000011080abb5 gtk_widget_draw_internal + 421
[bt] #19 19  libgtk-3.0.dylib                    0x0000000110591181 gtk_container_propagate_draw + 625
[bt] #20 20  libgtk-3.0.dylib                    0x00000001105918e3 gtk_container_draw + 227
[bt] #21 21  libgtk-3.0.dylib                    0x000000011072ee3c gtk_scrolled_window_render + 492
[bt] #22 22  libgtk-3.0.dylib                    0x00000001105998b2 gtk_css_custom_gadget_draw + 146
[bt] #23 23  libgtk-3.0.dylib                    0x000000011059edc7 gtk_css_gadget_draw + 1447
[bt] #24 24  libgtk-3.0.dylib                    0x000000011072aa3c gtk_scrolled_window_draw + 92
[bt] #25 25  libgtk-3.0.dylib                    0x000000011080abb5 gtk_widget_draw_internal + 421
[bt] #26 26  libgtk-3.0.dylib                    0x0000000110591181 gtk_container_propagate_draw + 625
[bt] #27 27  libgtk-3.0.dylib                    0x00000001105918e3 gtk_container_draw + 227
[bt] #28 28  libgtk-3.0.dylib                    0x00000001105389fe gtk_box_draw_contents + 62
[bt] #29 29  libgtk-3.0.dylib                    0x00000001105998b2 gtk_css_custom_gadget_draw + 146
[bt] #30 30  libgtk-3.0.dylib                    0x000000011059edc7 gtk_css_gadget_draw + 1447
[bt] #31 31  libgtk-3.0.dylib                    0x00000001105376cc gtk_box_draw + 92
```

bhennion:
The last commit should fix it. I'm surprised you're encountering this when simply drawing rapidly though.
It probably means that `XojPageView::getVisiblePart()` returns an empty range quite often. I don't understand why that would happen...

rolandlo:
> The last commit should fix it. 

Looking at the code I'm convinced it does. I haven't managed to reproduce the failing assertion even without the commit. Seems I have a talent to produce crashes when trying out things the first time, which are hard to reproduce afterwards. ;-)

There are still issues with filled highlighter strokes though, not only on MacOS.

On Linux sometimes when I draw "dots" (with the mouse, just one click without moving), the dots are not shown until I grab them with a (rectangle) selection. In the terminal I get messages like:
```term
(xournalpp:17308): Gtk-CRITICAL **: 08:20:13.257: gtk_widget_queue_draw_area: assertion 'height >= 0' failed

(xournalpp:17308): Gtk-CRITICAL **: 08:20:13.898: gtk_widget_queue_draw_area: assertion 'height >= 0' failed
```

On MacOS there is a clear performance regression for filled highlighter strokes (comparing this PR with the latest nightly build). I will try once more with the PR rebased on master in the evening.

rolandlo:
Actually the "dots" can also disappear when selecting them. Moreover the selection can extend to include the left upper page corner. It also happens with the pen and without filling.

bhennion:
> On MacOS there is a clear performance regression for filled highlighter strokes (comparing this PR with the latest nightly build). I will try once more with the PR rebased on master in the evening.

That's bad! Do you also observe a performance change with the other tools (pen, filled/not filled, dashed or not)?
I'll try to find out what specific change made this regression.

The dots issue is most likely a missing padding somewhere. I'll figure it out.

bhennion:
Also, would you mind testing [this branch](https://github.com/bhennion/xournalpp/tree/pr/RemoveDeviceScale) on MacOs configured with -DDEBUG_MASKS=on? It should improve performances by using Quartz surfaces instead of image surfaces

bhennion:
https://github.com/xournalpp/xournalpp/pull/4158/commits/f4482026c11bfcdb5d08e756463b5b020f386cda might improve performances

rolandlo:
> [f448202](https://github.com/xournalpp/xournalpp/commit/f4482026c11bfcdb5d08e756463b5b020f386cda) might improve performances

It looks to me, that this is indeed the case. It will require some more testing.

rolandlo:
It looks like the dots issue is related to some strange behavior in `Points::addPoint` when less than 2 points have already been added. I don't understand it yet, but the result are wrong snappedBounds.

The issue is simple to reproduce (on Linux):

1) Start Xournal++ (PR version)
2) Add a "dot" using the pen or highlighter tool (with or without fill) with one click of the mouse
3) Select the "dot" and observe that the selection rectangle is too large.

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -256,7 +182,9 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
         const Point& pt = pv.front();
         if (this->hasPressure) {
             // Pressure inference provides a pressure value to the last event. Most devices set this value to 0.
-            this->stroke->setLastPressure(std::max(pt.z, pos.pressure) * this->stroke->getWidth());
+            const double newPressure = std::max(pt.z, pos.pressure * this->stroke->getWidth());
+            this->stroke->setLastPressure(newPressure);
+            this->viewPool->dispatch(xoj::view::StrokeToolView::THICKEN_FIRST_POINT_REQUEST, newPressure);
         }
         stroke->addPoint(pt);
```
rolandlo, line 189: The issue with "dots" getting wrong bounds and being only drawn after the 5 seconds interval has passed can be fixed by changing this line:
```suggestion
       drawSegmentTo(pt);
```

bhennion: I con't reproduce the bug you are reporting. Also, this fix doesn't quite make sense: either the stroke is pressure-sensitive, in which case 
```
    this->viewPool->dispatch(xoj::view::StrokeToolView::THICKEN_FIRST_POINT_REQUEST, newPressure);
```
repaints the dot, or the stroke is not pressure sensitive, in which case nothing needs to be repainted: the dot should have already been painted upon ButtonPress.

bhennion: Oh, I see! The issue lies in `Stroke::updateBounds()` which fails to properly compute the stroke's bounds when `Stroke::addPoint()` adds the first point. So the initial repainted Range, based on `Stroke::boundingRect()`, is not the right one.

bhennion: It was a little bit more sneaky than that. I was basically doing `points.emplace_back(points.front())`, which is UB: `emplace_back()` caused the vector to reallocate so the reference `points.front()` is dangling by the time it's used.
This is fixed by d94e19e

rolandlo: Unfortunately now the "dots" are rendered with delay on MacOS (where it worked previously with no delay). Something must still be wrong.

rolandlo: I must correct myself. There is no difference between MacOS and Linux. Just for filled highlighter "dots" there is no immediate rendering (the "dot" is only rendered after the max. 5 second period). For filled pen "dots" there is no issue. I don't see where a flagDirtyRegion is apparent.

rolandlo: @hrdl-github The nightly build works fine on MacOS (at least in that aspect).

bhennion: Fixed with the last commit. The `singleDot==true` case was not handled specifically in `StrokeToolFilledHighlighterView::draw()`. Thanks a lot for the catch and the investigation!

rolandlo: > I found another issue when testing [e28d71d](https://github.com/xournalpp/xournalpp/commit/e28d71ddd13e9b5e4be061e045f31b0acf814f28): when creating strokes with > 1 dots using the filled highlighter I get a dotted boundary and no filling. When repainted a few seconds later I get the expected result.

I just wanted to report the same thing.

bhennion: ^^Yeah, I realized this right fter posting. The "fixup" commit should solve this.

rolandlo: Except that in https://github.com/xournalpp/xournalpp/commit/d94e19ead6ca56f1b11ef21c5a1198ca4673cf8d this was still fine, only in https://github.com/xournalpp/xournalpp/commit/e28d71ddd13e9b5e4be061e045f31b0acf814f28 it is not

rolandlo (approved):
Looks all good to me know. Performance on MacOS is good as well.
Maybe the last fixup commits could be squashed before merging.

bhennion:
Rebased on master. I also
- Reduced further the mask wipeout area for filled highlighter strokes, to improve performances.
- Fixed a bug where he bounding boxes of single dots were not updated when those dots grew fatter.

rolandlo:
@bhennion I have tested it once more and haven't found any issues. From my side you can merge at any moment.

bhennion:
Thanks a lot! I'll merge in 72 hours unless a(nother) problem is raised.

PR #4159, opened 2022-07-21
Split tools/SplineHandler into View-Controller

### LINKED ISSUE #3579: Possible data race in `StrokeHandler`, `BaseStrokeHandler`, `SplineHandler` (possible others)

bhennion:
This follows both #4158 and #4137.

This PR splits tools/SplineHandler into a view and a controller. It also optimizes (a lot) the rendering costs by only repainting the necessary areas (as opposed to rerendering to big portions of pages). (although some further optimizations should come with splines and #2729).

~~The data is moreover protected by a mutex. Together with #4137 and #4158, this addresses #3579 and #4038.~~
**Edit**: This is actually not necessary. See #3579.

- [ ] Merge #4137 and #4158 and rebase


Technius:
The "[Finalize spline upon tool change or page focus change](https://github.com/xournalpp/xournalpp/pull/4159/commits/187392bfb674a91ef5a0093ec244f1cbe3c05eda)" commit seems to fix an unreported bug, thanks.

I also noticed that selecting elements (e.g., with Ctrl+A) while drawing a spline does not finalize it; is that also fixed with this PR?

bhennion:
> I also noticed that selecting elements (e.g., with Ctrl+A) while drawing a spline does not finalize it; is that also fixed with this PR?

No it's not fixed here. I was not aware of this issue. From what I could gather, this seems to be a bigger isssue: nothing gets finalized when hitting Ctrl+A (Textbox, Spline, even normal strokes). Might be more suited for a dedicated PR.

src/core/control/tools/SplineHandler.cpp
```diff
@@ -204,87 +149,102 @@ auto SplineHandler::onMotionNotifyEvent(const PositionInputData& pos, double zoo
...
-    rect.unite(this->computeRepaintRectangle());
-    this->redrawable->repaintRect(rect.x, rect.y, rect.width, rect.height);
-
+    this->viewPool->dispatch(xoj::view::SplineToolView::FLAG_DIRTY_REGION, rg);
     return true;
 }
 
-void SplineHandler::onSequenceCancelEvent() { stroke.reset(); }
+void SplineHandler::onSequenceCancelEvent() {
+    //  Touch screen sequence changed from normal to swipe/zoom/scroll sequence
```
rolandlo, line 159: This comment is misleading as `onSequenceCancelEvent` can also happen without touchscreen.

rolandlo (changes requested):
I have observed two issues with rendering artifacts:
1) When zooming in or out while drawing a spline, the (sizes of the) circles are not updated until the cursor position changes.
2) There are rendering artifacts near the filled red circle. The circle should get filled as a whole when the cursor is close enough to its center, but it is currently not.

Two questions:
1) Why  did you decide to make the SplineHandler close the spline on ButtonRelease instead of ButtonPress?
2) Why do the named colors need alpha values?

bhennion:
> 1. When zooming in or out while drawing a spline, the (sizes of the) circles are not updated until the cursor position changes.

I'll fix that

> 2. There are rendering artifacts near the filled red circle. The circle should get filled as a whole when the cursor is close enough to its center, but it is currently not.

I'm not quite sure I understand what you mean here. Are you referring to the disk appearing slightly smaller than the circle? If so, I just fixed that.


> 1. Why  did you decide to make the SplineHandler close the spline on ButtonRelease instead of ButtonPress?

I don't fully remember my motivation back then. I think it was about UX: buttons and things of this sort are only activated once you release the button (so you can always cancel an ongoing "click" by releasing outside of the active area). I just followed this pattern, I think.

> 2. Why do the named colors need alpha values?

Simply to be able to use Util::cairo_set_source_argb on them without caring about the alpha values

rolandlo:
> I'm not quite sure I understand what you mean here. Are you referring to the disk appearing slightly smaller than the circle? If so, I just fixed that.

What I meant wasn't fixed (at least not fully). See here (at low zoom):

https://user-images.githubusercontent.com/40485433/219902715-922c1f1f-ea6c-4a54-975d-64f948a0ec14.mp4

bhennion:
@rolandlo I think I fixed the two bugs you reported.

rolandlo:
> @rolandlo I think I fixed the two bugs you reported.

That works great now!
I have found one very minor issue though: When you click into the red circle and move out of it (to cancel the spline closing) and then zoom in or out, the spline is closed (which I guess it shouldn't) although editing continues.

bhennion:
> > @rolandlo I think I fixed the two bugs you reported.
> 
> That works great now! I have found one very minor issue though: When you click into the red circle and move out of it (to cancel the spline closing) and then zoom in or out, the spline is closed (which I guess it shouldn't) although editing continues.

It's actually the same without zoom in or out, isn't it? The stroke still displays as closed until the button is released. Or are you referring to something else?

I'm not sure I wanna change that though. I don't see what else could happen in this case. Dragging the end of the spline around while pressed seems odd, as releasing won't do anything (and it should not do anything).

rolandlo:
> > 
> > That works great now! I have found one very minor issue though: When you click into the red circle and move out of it (to cancel the spline closing) and then zoom in or out, the spline is closed (which I guess it shouldn't) although editing continues.
> 
> It's actually the same without zoom in or out, isn't it? The stroke still displays as closed until the button is released. Or are you referring to something else?
> 

It is the same, right. I had observed something different, but not any more.

bhennion:
Thanks for the review!

PR #4160, opened 2022-07-21
Bump gcc version from 8 to 10 in CI

bhennion:


bhennion:
This version bump allows for c++17 parallel execution policies and std::transform_reduce, which were only introduced in GCC 9.

PR #4162, opened 2022-07-22
Fix: move events being processed twice

bhennion:
Some Motion events were being transfered twice to the same InputHandler, for tools such that isSinglePageTool == true (e.g. rectangle, ellipse, spline tools). This patch fixes this.

It also fixes ToolHandler::isSinglePageTool, were a boolean operation was obviously wrong.

Merging to the release branch in 48 hours, unless an objection is raised.

### DIFF
--- a/src/control/ToolHandler.cpp
+++ b/src/control/ToolHandler.cpp
@@ -532,15 +532,14 @@ auto ToolHandler::isSinglePageTool() -> bool {
     ToolType toolType = this->getToolType();
     DrawingType drawingType = this->getDrawingType();
 
-    return toolType == (TOOL_PEN && (drawingType == DRAWING_TYPE_ARROW || drawingType == DRAWING_TYPE_ELLIPSE ||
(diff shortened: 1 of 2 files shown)


Technius (approved):
LGTM, nice catch.

Just to make sure I understand the bug: in the original code, the `onMotionNotifyEvent` will fire for a single page tool (input will be clamped to the _input start_ page), and the event will also fire due to the pen being in the canvas (input will be considered for the _current_ page). Thus, for a single page tool, the input start page and the current page may be the same. Since the function does not return immediately in either case, the event is triggered for a total of two times.

bhennion:
> Just to make sure I understand the bug: in the original code, the `onMotionNotifyEvent` will fire for a single page tool (input will be clamped to the _input start_ page), and the event will also fire due to the pen being in the canvas (input will be considered for the _current_ page). Thus, for a single page tool, the input start page and the current page may be the same. Since the function does not return immediately in either case, the event is triggered for a total of two times.

Yes, exactly!

PR #4163, opened 2022-07-24
Code cleanup before handlers splitting

bhennion:
This is a code cleanup PR coming from #4137, #4158 and #4159, to simplify the reviewing.

Two things in here:
1. Some const-correctness in various classes in gui/
2. Fix util/PairView to work on const qualified containers

Since nothing is really happening in this PR, I'll be merging it in 48 hours unless someone objects.


Technius (approved):
LGTM. I think this is safe to merge right away.

PR #4164, opened 2022-07-25
Fix low resolution filled highlighter stroke bug

bhennion:
Fixes a bug where filled highlighter strokes do not have the right resolution when zoomed in.
        The bug was introduced by fc17f8555.

### DIFF
--- a/src/core/view/StrokeView.cpp
+++ b/src/core/view/StrokeView.cpp
@@ -110,27 +110,42 @@ void StrokeView::draw(const Context& ctx) const {
          */
 
         /**
-         * We need to rescale the mask according to the scaling ratio of the target cairo context.
-         * We find out this scaling by looking at the transformation matrix
+         * The mask needs to have to right resolution: the number of pixels per page coordinate units.
+         *
+         * This value can be recovered from the given canvas: the mask must have the exact same resolution as the canvas
+         * This resolution combines both the surface scale and the transformation matrix zoom ratio.
          */
-        cairo_matrix_t matrix;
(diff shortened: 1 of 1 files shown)


src/core/view/StrokeView.cpp
```diff
@@ -118,19 +118,25 @@ void StrokeView::draw(const Context& ctx) const {
         // We assume the matrix is diagonal (i.e. only scaling, no rotation)
         assert(matrix.xy == 0 && matrix.yx == 0);
 
+        double scaleX;
+        double scaleY;
+        cairo_surface_get_device_scale(cairo_get_target(ctx.cr), &scaleX, &scaleY);
```
Technius, line 123: I would add a comment along the lines of "The mask needs to be scaled by both zoom and device scale (e.g., to be same as the canvas), otherwise the stroke will rendered at a lower resolution". This comment could be combined with the one a few ones above.

I also think the code here would be easier to understand if you moved this snippet above the matrix snippet, so that it's clear that `scaleX, scaleY` refer to the final scale values:
```c++
Define the scale variables
Initialize them to the device scale
Get the transformation matrix
Multiply the scale values by the matrix values
```

Technius (approved):
LGTM. Minor maintainability concern below.

bhennion:

bhennion:
The clang-format CI test is broken, with message
```
The following packages have unmet dependencies:
 clang-format : Depends: clang-format-15 (>= 15~) but it is not installable
```

PR #4165, opened 2022-07-25
 Added functionality to the missing PDF background dialog to propose a replacement file

Malex14:
If there is a PDF file with the same name as the missing PDF in the directory where the journal file is saved, the dialog now proposes that file to the user. If not the dialog stays as before.

This is especially useful when moving the journal file and the background to a new location. Or in my case where I switch between Linux and Windows a lot, the background can be automatically found even tho Linux and Windows are using a different path structure

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2130,25 +2130,41 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
     if ((loadedDocument != nullptr && loadHandler.isAttachedPdfMissing()) ||
         !loadHandler.getMissingPdfFilename().empty()) {
         // give the user a second chance to select a new PDF filepath, or to discard the PDF
-
         const fs::path missingFilePath = fs::path(loadHandler.getMissingPdfFilename());
 
-        std::string msg;
+        string parentFolderPath;
+        string filename;
 
+        // since POSIX systems detect the whole Windows path as a filename, this checks whether missingFilePath contains
+        // a Windows path
+        GMatchInfo* matchInfo = nullptr;
+        GRegex* regex = g_regex_new(R"([A-Z]:\\.*\\(.*))", static_cast<GRegexCompileFlags>(0),
+                                    static_cast<GRegexMatchFlags>(0), nullptr);
+        g_regex_match(regex, missingFilePath.filename().c_str(), static_cast<GRegexMatchFlags>(0), &matchInfo);
+
+        if (g_match_info_matches(matchInfo)) {
+            parentFolderPath = missingFilePath.filename();
+            filename = g_match_info_fetch(matchInfo, 1);
+        } else {
+            parentFolderPath = missingFilePath.parent_path().string();
+            filename = missingFilePath.filename().string();
+        }
+
+        std::string msg;
         if (loadHandler.isAttachedPdfMissing()) {
             msg = FS(_F("The attached background file \"{1}\" could not be found. It might have been moved, "
                         "renamed or deleted.") %
-                     missingFilePath.filename().string());
+                     filename);
         } else {
             msg = FS(_F("The background file \"{1}\" could not be found. It might have been moved, renamed or "
                         "deleted.") %
-                     missingFilePath.filename().string());
+                     filename);
         }
 
-        msg += FS(_F("\nIt was last seen at: \"{1}\"") % missingFilePath.parent_path().string());
+        msg += FS(_F("\nIt was last seen at: \"{1}\"") % parentFolderPath);
 
         // try to find file in current directory
-        auto proposedPdfFilepath = filepath.parent_path() / missingFilePath.filename();
+        auto proposedPdfFilepath = filepath.parent_path() / filename;
         bool proposedPdfFileExits = fs::exists(proposedPdfFilepath);
         if (proposedPdfFileExits) {
             msg += FS(_F("\nProposed replacement file: \"{1}\"") % proposedPdfFilepath.string());
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2148,7 +2148,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         }
 
         // try to find file in current directory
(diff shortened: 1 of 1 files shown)


src/core/control/Control.cpp
```diff
@@ -2132,30 +2132,43 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+        if (proposedPdfFileExits) {
+            msg += FS(_F("\nProposed replacement file: \"{1}\"") % proposedPdfFilepath.string());
+        }
+        GtkWidget* dialog = gtk_message_dialog_new(getGtkWindow(), GTK_DIALOG_MODAL, GTK_MESSAGE_QUESTION,
+                                                   GTK_BUTTONS_NONE, "%s", msg.c_str());
+
+        if (proposedPdfFileExits) gtk_dialog_add_button(GTK_DIALOG(dialog), _("Use proposed PDF"), 1);
+        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Select another PDF"), 2);
+        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Remove PDF Background"), 3);
+        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Cancel"), 4);
```
bhennion, line 2162: Two things: first, put braces around every if/while/for scope, even one-liners (see [our coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#indentation)).

Secondly, maybe having an enum for the response values could help with readability:
```suggestion
        enum ResponseId : int { USE_PROPOSED_PDF, SELECT_ANOTHER, REMOVE_BACKGROUND, CANCEL };
        if (proposedPdfFileExits) {
            gtk_dialog_add_button(GTK_DIALOG(dialog), _("Use proposed PDF"), USE_PROPOSED_PDF);
        }
        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Select another PDF"), SELECT_ANOTHER);
        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Remove PDF Background"), REMOVE_BACKGROUND);
        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Cancel"), CANCEL);
```

src/core/control/Control.cpp
```diff
@@ -2132,30 +2132,43 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
         gtk_window_set_transient_for(GTK_WINDOW(dialog), GTK_WINDOW(this->getWindow()->getWindow()));
         int res = gtk_dialog_run(GTK_DIALOG(dialog));
         gtk_widget_destroy(dialog);
 
-        if (res == 2)  // remove PDF background
+        if (res == 3)  // remove PDF background
```
bhennion, line 2167: For readability, prefer a switch statement here:
```suggestion
        switch(res) {
        case REMOVE_BACKGROUND:
```

src/core/control/Control.cpp
```diff
@@ -2132,30 +2132,43 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
         {
             loadHandler.removePdfBackground();
             loadedDocument = loadHandler.loadDocument(filepath);
-        } else if (res == 1)  // select another PDF background
+        } else if (res == 2)  // select another PDF background
```
bhennion, line 2171: same here

src/core/control/Control.cpp
```diff
@@ -2164,6 +2177,12 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
                 loadHandler.setPdfReplacement(pdfFilename, attachToDocument);
                 loadedDocument = loadHandler.loadDocument(filepath);
             }
+        } else if (res == 1) // use proposed PDF
```
bhennion, line 2180: and here, and with a default in the end.

bhennion (changes requested):
First, thanks for the PR. What you implement seems very useful to me.

I only have a couple of subjective code readability nitpicks.

bhennion (changes requested):

src/core/control/Control.cpp
```diff
@@ -2130,59 +2130,82 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
         // give the user a second chance to select a new PDF filepath, or to discard the PDF
-
         const fs::path missingFilePath = fs::path(loadHandler.getMissingPdfFilename());
 
-        std::string msg;
+        string parentFolderPath;
+        string filename;
+
+        // since POSIX systems detect the whole Windows path as a filename, this checks whether missingFilePath contains
+        // a Windows path
```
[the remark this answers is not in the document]
Malex14: I think you are right. It`s only necessary on POSIX systems, since windows can handle paths separated by '/' and '\'. Now there is only a small problem: on windows 'C:' gets prepended to any POSIX path, but this is only a visual nitpick, it does not harm the functionality.

src/core/control/Control.cpp
```diff
@@ -2130,40 +2130,82 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+            msg = FS(_F("The attached background file \"{1}\" could not be found. It might have been moved, "
+                        "renamed or deleted.") %
+                     filename);
+        } else {
+            msg = FS(_F("The background file \"{1}\" could not be found. It might have been moved, renamed or "
+                        "deleted.") %
+                     filename);
+        }
+
+        msg += FS(_F("\nIt was last seen at: \"{1}\"") % parentFolderPath);
```
Malex14, line 2166: I would say that these two messages are very similar. One of the two should be enough, since the average user won't know the difference between a "attached" file and not attached file.
```suggestion
        auto msg = FS(_F("The background file \"{1}\" could not be found. It might have been moved, "
                         "renamed or deleted.\nIt was last seen at: \"{2}\"") %
                      filename % parentFolderPath);
```

bhennion: I agree the average user might not know, but it would help whenever the user submits a bug report. I'd keep it like that.

bhennion (approved):
Alright. This looks good to me!

bhennion:

PR #4185, opened 2022-08-05
add optional mode parameter for plugin callback functions

rolandlo:
This PR implements an optional `mode` parameter for `app.registerUi`. It allows to have multiple registered entries to share the same callback function executed in a different mode. If the `mode` parameter is specified the callback function should have the form
```lua
def functionname(mode)
  -- code
end
```
For example you may want to register a menu entry for each color in the color palette without defining a separate callback function for each color.


rolandlo:

PR #4198, opened 2022-08-14
fix bug for setsquare translation via arrow keys

rolandlo:
This is an obvious bug, since `xdir` and `ydir` get multiplied by `amount` twice.


rolandlo:
Merging right away. The fix is completely trivial.

PR #4211, opened 2022-08-25
Refactor setsquare to use MVC pattern

rolandlo:
~This is a continuation of #3962 and is based on #4137, #4158 and #4159 from @bhennion.~
This PR refactors the setsquare. ~Currently it contains adding a compass, but that will later be moved into another PR.~
Previously it contained a compass, but that was moved into PR #4443.
I tried to follow the MVC pattern (push variant) explained by @LittleHuba in https://github.com/xournalpp/xournalpp/pull/3962#discussion_r858021356. ~Only the last two commits are relevant.~
I will explain the most relevant commits: 

1) MVC pattern
In 6efc15adb79f885b221fa7e9b353914d7c2571c2 the MVC pattern is applied:

The class `Setsquare` is the model.
The class `SetsquareView` is the view.
The class `SetsquareController` is the controller.
The class `SetsquareInputHandler` is somewhere between a view and a controller.

- The model has a `viewPool` and a `handlerPool` for pushing updates to the `SetsquareView`(s) and `SetsquareInputHandler`(s).
It stores the data for the setsquare and the stroke (near the longest side of the setsquare or from a point close to a shorter side of the setsquare to the midpoint of the longest side) and has getters and setters.

- The view only uses the model for querying the `viewPool`. It doesn't update the model and contains only drawing methods except of the methods to update itself when pushed from the model. 

- The input handler just handles input and uses the controller to manipulate the model. It also updates itself when pushed from the model.

- The controller manipulates the model (moves, rotates and scales the setsquare, creates and updates strokes) and contains more methods, which didn't fit in one of the other classes.

For pushing updates to the view and input handler, I use the `xoj::util::Listener` class; for rendering the setsquare I use the `OverlayBase` and `ToolView` classes, all by @bhennion .

There ~are still two issues~ is still one issue with the rendering:
~1) When activating the setsquare, it is not rendered immediately. It is only rendered after it has been moved/scaled or rotated. I'm not sure why that happens. I tried to notify the setsquare after constructing instances of all classes (in `Control.cpp`), but that crashes.~

~2)~ Since the rendering uses an overlay of one of the page views and the pages are drawn one after the other, when moving the setsquare towards the next page, the fraction that lands on the next page appears to be cut, as if the setsquare was gliding under the next page. It's not like that when moving the setsquare to the page before.
SetsquareCutFromNextPage

2) Derive Setsquare classes from GeometryTool classes

In order to prepare for other geometry tools (like a compass or a ruler) to be added, in 714e479e1c04e618e1aed20f74701e07b206e8a8 the Setsquare classes are derived from more generic GeometryTool classes. Those classes basically contain the functionality of the Setsquare that all geometry tools will most likely share. This makes it easier to implement new geometry tools.

- `Setsquare` is derived from `GeometryTool`
- `SetsquareView` is derived from `GeometryToolView`
- `SetsquareController` is derived from `GeometryToolController`
- `SetsquareInputHandler` is derived from `GeometryToolHandler`

This strategy is demonstrated in ~commit bb512582867027fd5bb1aba64af264946e34b80f~ PR #4443  where the Compass is added. ~Although that commit won't be part of this PR, it is included so that the strategy can be checked to work.~

Any feedback is welcome. As I had no previous exposure to the MVC pattern, I find it quite tough to find the proper way for this refactoring.

~Only the two last commits are relevant:
ff9058228d7f4e27933c0b567bcb621d9360d19f is the combination of the commits in #3962 rebased on #4159. The setsquare is functionally equivalent to the version in master. 
The real work is in adfc317dcbc3e3e8f8218b77610ba504ab7229b4~

**Edit:** Updated description, described derivation from GeometryTool classes (Oct 15)
**Edit:** Updated description after moving the compass into another PR.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -345,7 +345,7 @@ void Control::initWindow(MainWindow* win) {
 
     fireActionSelected(GROUP_SNAPPING, settings->isSnapRotation() ? ACTION_ROTATION_SNAPPING : ACTION_NONE);
     fireActionSelected(GROUP_GRID_SNAPPING, settings->isSnapGrid() ? ACTION_GRID_SNAPPING : ACTION_NONE);
-    fireActionSelected(GROUP_SETSQUARE, ACTION_NONE);
+    fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_NONE);
 }
 
 auto Control::autosaveCallback(Control* control) -> bool {
@@ -647,12 +647,14 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
             }
             break;
         case ACTION_SETSQUARE:
-            if (auto xournal = this->win->getXournal(); !xournal->getSetsquareController()) {
-                xournal->resetSetsquare();
-                xournal->makeSetsquare();
+            if (auto xournal = this->win->getXournal();
+                !xournal->getGeometryToolController() ||
+                xournal->getGeometryToolController()->getType() != "setsquare") {
+                xournal->resetGeometryTool();
+                xournal->makeGeometryTool("setsquare");
                 xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             } else {
-                xournal->resetSetsquare();
+                xournal->resetGeometryTool();
                 xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             }
             break;
@@ -1357,11 +1359,11 @@ void Control::updateDeletePageButton() {
 void Control::deletePage() {
     clearSelectionEndText();
 
-    // if the current page contains the Setsquare, reset it
+    // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
-    auto setsquareController = win->getXournal()->getSetsquareController();
-    if (setsquareController && doc->indexOf(setsquareController->getPage()) == pNr) {
-        win->getXournal()->resetSetsquare();
+    auto geometryToolController = win->getXournal()->getGeometryToolController();
+    if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
+        win->getXournal()->resetGeometryTool();
     }
     // don't allow delete pages if we have less than 2 pages,
     // so we can be (more or less) sure there is at least one page.
@@ -2779,7 +2781,7 @@ auto Control::close(const bool allowDestroy, const bool allowCancel) -> bool {
     if (allowDestroy && discard) {
         this->closeDocument();
     }
-    win->getXournal()->resetSetsquare();
+    win->getXournal()->resetGeometryTool();
     return true;
 }
 
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -649,18 +649,15 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
                 selectTool(TOOL_HAND);
             }
             break;
-        case ACTION_SETSQUARE:
-            if (auto xournal = this->win->getXournal();
-                !this->geometryToolController ||
-                this->geometryToolController->getType() != GeometryToolType::SETSQUARE) {
-                resetGeometryTool();
+        case ACTION_SETSQUARE: {
+            bool needsNewSetsquare = !this->geometryToolController ||
+                                     this->geometryToolController->getType() != GeometryToolType::SETSQUARE;
+            resetGeometryTool();
+            if (needsNewSetsquare) {
                 makeGeometryTool(GeometryToolType::SETSQUARE);
-                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
-            } else {
-                resetGeometryTool();
-                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             }
             break;
+        }
         case ACTION_TOOL_FLOATING_TOOLBOX:
             if (enabled) {
                 selectTool(TOOL_FLOATING_TOOLBOX);
@@ -1066,25 +1063,40 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
 
 void Control::makeGeometryTool(GeometryToolType tool) {
     auto view = this->win->getXournal()->getViewFor(getCurrentPageNo());
-    if (tool == GeometryToolType::SETSQUARE) {
-        auto setsquare = new Setsquare();
-        view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoom));
-        this->geometryTool = std::unique_ptr<GeometryTool>(setsquare);
-        this->geometryToolController = std::make_unique<SetsquareController>(view, setsquare);
-        std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
-                std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
-        geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
-        auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
-        xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
-        fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
+    auto* xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
+    switch (tool) {
+        case SETSQUARE: {
+            auto setsquare = new Setsquare();
+            view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoom));
+            this->geometryTool = std::unique_ptr<GeometryTool>(setsquare);
+            this->geometryToolController = std::make_unique<SetsquareController>(view, setsquare);
+            std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
+                    std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
+            geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
+            xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
+            fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
+            geometryTool->notify();
+            break;
+        }
+        default:
+            g_warning("Unknown geometry tool type %d", tool);
     }
 }
 
 void Control::resetGeometryTool() {
+    Range rg;
+    bool hasGeometryTool = false;
+    if (this->geometryTool) {
+        hasGeometryTool = true;
+        rg = this->geometryTool->getToolRange(true);
+    }
     this->geometryToolController.reset();
     this->geometryTool.reset();
-    auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
+    auto* xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
     xournal->input->resetGeometryToolInputHandler();
+    if (win && hasGeometryTool) {
+        (win->getXournal()->getViewFor(getCurrentPageNo()))->rerenderRange(rg);
+    }
     fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_NONE);
 }
 
@@ -1388,11 +1400,14 @@ void Control::deletePage() {
 
     // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
-    doc->lock();
-    if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
-        resetGeometryTool();
+    if (geometryToolController) {
+        doc->lock();
+        auto page = doc->indexOf(geometryToolController->getPage());
+        doc->unlock();
+        if (page == pNr) {
+            resetGeometryTool();
+        }
     }
-    doc->unlock();
     // don't allow delete pages if we have less than 2 pages,
     // so we can be (more or less) sure there is at least one page.
     if (this->doc->getPageCount() < 2) {
@@ -2271,8 +2286,8 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         parentFolderPath = missingFilePath.parent_path().string();
         filename = missingFilePath.filename().string();
 #else
-        // since POSIX systems detect the whole Windows path as a filename, this checks whether missingFilePath contains
-        // a Windows path
+        // since POSIX systems detect the whole Windows path as a filename, this checks whether missingFilePath
+        // contains a Windows path
         std::regex regex(R"([A-Z]:\\(?:.*\\)*(.*))");
         std::cmatch matchInfo;
 
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -14,6 +14,7 @@
 #include "control/ClipboardHandler.h"                            // for Clip...
 #include "control/RecentManager.h"                               // for Rece...
 #include "control/ScrollHandler.h"                               // for Scro...
+#include "control/SetsquareController.h"                         // for Sets...
 #include "control/Tool.h"                                        // for Tool
 #include "control/ToolHandler.h"                                 // for Tool...
 #include "control/jobs/AutosaveJob.h"                            // for Auto...
@@ -50,7 +51,9 @@
 #include "gui/dialog/SettingsDialog.h"                           // for Sett...
 #include "gui/dialog/ToolbarManageDialog.h"                      // for Tool...
 #include "gui/dialog/toolbarCustomize/ToolbarDragDropHandler.h"  // for Tool...
+#include "gui/inputdevices/GeometryToolInputHandler.h"           // for Geom...
 #include "gui/inputdevices/HandRecognition.h"                    // for Hand...
+#include "gui/inputdevices/SetsquareInputHandler.h"              // for Sets...
 #include "gui/sidebar/Sidebar.h"                                 // for Sidebar
 #include "gui/toolbarMenubar/ToolMenuHandler.h"                  // for Tool...
 #include "gui/toolbarMenubar/model/ToolbarData.h"                // for Tool...
@@ -648,13 +651,13 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
             break;
         case ACTION_SETSQUARE:
             if (auto xournal = this->win->getXournal();
-                !xournal->getGeometryToolController() ||
-                xournal->getGeometryToolController()->getType() != GeometryToolType::SETSQUARE) {
-                xournal->resetGeometryTool();
-                xournal->makeGeometryTool(GeometryToolType::SETSQUARE);
+                !this->geometryToolController ||
+                this->geometryToolController->getType() != GeometryToolType::SETSQUARE) {
+                resetGeometryTool();
+                makeGeometryTool(GeometryToolType::SETSQUARE);
                 xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             } else {
-                xournal->resetGeometryTool();
+                resetGeometryTool();
                 xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             }
             break;
@@ -1061,6 +1064,26 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
     }
 }
 
+void Control::makeGeometryTool(GeometryToolType tool) {
+    auto view = this->win->getXournal()->getViewFor(getCurrentPageNo());
+    if (tool == GeometryToolType::SETSQUARE) {
+        auto setsquare = new Setsquare();
+        view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoom));
+        std::unique_ptr<GeometryTool> geometryTool = std::unique_ptr<GeometryTool>(setsquare);
+        this->geometryToolController = std::make_unique<SetsquareController>(view, setsquare);
+        std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
+                std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
+        geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
+        fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
+        this->win->getXournal()->setGeometryTool(std::move(geometryTool), std::move(geometryToolInputHandler));
+    }
+}
+
+void Control::resetGeometryTool() {
+    this->geometryToolController.reset();
+    this->win->getXournal()->resetGeometryTool();
+}
+
 auto Control::copy() -> bool {
     if (this->win && this->win->getXournal()->copy()) {
         return true;
@@ -1361,10 +1384,9 @@ void Control::deletePage() {
 
     // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
-    auto geometryToolController = win->getXournal()->getGeometryToolController();
     doc->lock();
     if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
-        win->getXournal()->resetGeometryTool();
+        resetGeometryTool();
     }
     doc->unlock();
     // don't allow delete pages if we have less than 2 pages,
@@ -2783,7 +2805,7 @@ auto Control::close(const bool allowDestroy, const bool allowCancel) -> bool {
     if (allowDestroy && discard) {
         this->closeDocument();
     }
-    win->getXournal()->resetGeometryTool();
+    resetGeometryTool();
     return true;
 }
 
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1069,23 +1069,23 @@ void Control::makeGeometryTool(GeometryToolType tool) {
     if (tool == GeometryToolType::SETSQUARE) {
         auto setsquare = new Setsquare();
         view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoom));
-        std::unique_ptr<GeometryTool> geometryTool = std::unique_ptr<GeometryTool>(setsquare);
+        this->geometryTool = std::unique_ptr<GeometryTool>(setsquare);
         this->geometryToolController = std::make_unique<SetsquareController>(view, setsquare);
         std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
                 std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
         geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
-        fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
-        this->win->getXournal()->setGeometryTool(std::move(geometryTool));
         auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
         xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
+        fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
     }
 }
 
 void Control::resetGeometryTool() {
     this->geometryToolController.reset();
-    this->win->getXournal()->resetGeometryTool();
+    this->geometryTool.reset();
     auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
     xournal->input->resetGeometryToolInputHandler();
+    fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_NONE);
 }
 
 auto Control::copy() -> bool {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -649,9 +649,9 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
         case ACTION_SETSQUARE:
             if (auto xournal = this->win->getXournal();
                 !xournal->getGeometryToolController() ||
-                xournal->getGeometryToolController()->getType() != "setsquare") {
+                xournal->getGeometryToolController()->getType() != GeometryToolType::SETSQUARE) {
                 xournal->resetGeometryTool();
-                xournal->makeGeometryTool("setsquare");
+                xournal->makeGeometryTool(GeometryToolType::SETSQUARE);
                 xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             } else {
                 xournal->resetGeometryTool();
@@ -1362,9 +1362,11 @@ void Control::deletePage() {
     // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
     auto geometryToolController = win->getXournal()->getGeometryToolController();
+    doc->lock();
     if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
         win->getXournal()->resetGeometryTool();
     }
+    doc->unlock();
     // don't allow delete pages if we have less than 2 pages,
     // so we can be (more or less) sure there is at least one page.
     if (this->doc->getPageCount() < 2) {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1075,13 +1075,17 @@ void Control::makeGeometryTool(GeometryToolType tool) {
                 std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
         geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
         fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
-        this->win->getXournal()->setGeometryTool(std::move(geometryTool), std::move(geometryToolInputHandler));
+        this->win->getXournal()->setGeometryTool(std::move(geometryTool));
+        auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
+        xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
     }
 }
 
 void Control::resetGeometryTool() {
     this->geometryToolController.reset();
     this->win->getXournal()->resetGeometryTool();
+    auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
+    xournal->input->resetGeometryToolInputHandler();
 }
 
 auto Control::copy() -> bool {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -87,6 +87,7 @@
 #include "util/serializing/InputStreamException.h"               // for Inpu...
 #include "util/serializing/ObjectInputStream.h"                  // for Obje...
 #include "view/SetsquareView.h"                                  // for Sets...
+#include "view/overlays/OverlayView.h"                           // for Over...
 
 #include "CrashHandler.h"                    // for emer...
 #include "FullscreenHandler.h"               // for Full...
@@ -646,21 +647,14 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
             }
             break;
         case ACTION_SETSQUARE:
-            if (!this->win->getXournal()->getSetsquareView()) {
-                // bring up setsquare in page center
-                auto setsquare = std::make_unique<Setsquare>();
-                auto view = win->getXournal()->getViewFor(getCurrentPageNo());
-                std::unique_ptr<SetsquareView> setsquareView = std::make_unique<SetsquareView>(view, setsquare);
-                if (!view) {
-                    setsquareView.reset(nullptr);
-                }
-                this->win->getXournal()->setSetsquareView(std::move(setsquareView));
-                fireActionSelected(GROUP_SETSQUARE, ACTION_SETSQUARE);
+            if (auto xournal = this->win->getXournal(); !xournal->getSetsquareController()) {
+                xournal->resetSetsquare();
+                xournal->makeSetsquare();
+                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             } else {
-                // hide setsquare
-                this->win->getXournal()->resetSetsquareView();
+                xournal->resetSetsquare();
+                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
             }
-            win->getXournal()->repaintSetsquare(true);
             break;
         case ACTION_TOOL_FLOATING_TOOLBOX:
             if (enabled) {
@@ -1365,9 +1359,9 @@ void Control::deletePage() {
 
     // if the current page contains the Setsquare, reset it
     size_t pNr = getCurrentPageNo();
-    auto setsquareView = win->getXournal()->getSetsquareView();
-    if (setsquareView && doc->indexOf(setsquareView->getPage()) == pNr) {
-        win->getXournal()->resetSetsquareView();
+    auto setsquareController = win->getXournal()->getSetsquareController();
+    if (setsquareController && doc->indexOf(setsquareController->getPage()) == pNr) {
+        win->getXournal()->resetSetsquare();
     }
     // don't allow delete pages if we have less than 2 pages,
     // so we can be (more or less) sure there is at least one page.
@@ -2785,7 +2779,7 @@ auto Control::close(const bool allowDestroy, const bool allowCancel) -> bool {
     if (allowDestroy && discard) {
         this->closeDocument();
     }
-    win->getXournal()->resetSetsquareView();
+    win->getXournal()->resetSetsquare();
     return true;
 }
 
--- a/src/core/control/GeometryToolController.cpp
+++ b/src/core/control/GeometryToolController.cpp
@@ -79,21 +79,20 @@ void GeometryToolController::addStrokeToLayer() {
     const auto layer = page->getSelectedLayer();
 
     const auto undo = control->getUndoRedoHandler();
-    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke));
-
-    layer->addElement(stroke);
+    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke.get()));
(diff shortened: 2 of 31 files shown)


bhennion:
Thanks for the great work!
I haven't had time to check the code in details yet, but I do have input on this point:

> 2\. Since the rendering uses an overlay of one of the page views and the pages are drawn one after the other, when moving the setsquare towards the next page, the fraction that lands on the next page appears to be cut, as if the setsquare was gliding under the next page. It's not like that when moving the setsquare to the page before.

To avoid issues like this, some overlays and tools will have to be owned, not by the PageController, but by the XournalController (currently a mix of XournalView and widget/XournalWidget). The SetSquare is one of those, as are selections for instance.
So in the end, a decent DocumentView class should have its own vector of OverlayView's.

As far as the SetSquare is concerned, I think none of its related classes should be owned by a PageSomething class, and the SetSquare classes should not refer to any specific page (there are bugs on the current master branch if you try to drag the setsquare to a different page and use it there).

rolandlo:
Thanks for the input.

> As far as the SetSquare is concerned, I think none of its related classes should be owned by a PageSomething class, and the SetSquare classes should not refer to any specific page (there are bugs on the current master branch if you try to drag the setsquare to a different page and use it there).

I think the setsquare must always refer to a specific page. What should happen otherwise, if you change the page layout (e.g. change from a 2-row setup to a 3-row setup), if the position of the setsquare is not relative to a specific page? Currently when you activate the setsquare, it is activated on a specific page and should only be used there. If you need it on a different page, you should deactivate the setsquare and activate it on the other page.

rolandlo:
I managed to fix the first rendering bug (where the setsquare was invisible until it was manipulated). So everything works again, just the setsquare is overpainted on the next page. Thinking about that issue a little more, it seems to be only a question about rendering priority, not about ownership. Maybe overlays could get a rendering priority assigned, so that overlays of lower priority would only get painted after all overlays (of all pages) of higher priority were painted.

bhennion:
Hey @rolandlo. Could you rebase this on latest master? I don't think this conflicts with #4158 or #4159, so it could be directly based on master. This would help the reviewing.
A couple of changes happened in #4137 before it got merged, so the rebasing on master may not be trivial.

rolandlo:
> Hey @rolandlo. Could you rebase this on latest master? I don't think this conflicts with #4158 or #4159, so it could be directly based on master. This would help the reviewing. A couple of changes happened in #4137 before it got merged, so the rebasing on master may not be trivial.

Sure, I will rebase on master. I have progressed more and already implemented the compass. Now I know better what the setsquare code and its integration in the code base should look like. It will take some time to split everything into sensible commits and PR's.

rolandlo:
> Hey @rolandlo. Could you rebase this on latest master? I don't think this conflicts with #4158 or #4159, so it could be directly based on master. This would help the reviewing. A couple of changes happened in #4137 before it got merged, so the rebasing on master may not be trivial.

The rebasing is finished. If you could take a look that would be great. I have updated the description on the top to help through the most important commits.

rolandlo:
Currently there are regular pointers
```
    GeometryTool* geometryTool = nullptr;
    GeometryToolController* geometryToolController = nullptr;
    GeometryToolHandler* geometryToolHandler = nullptr;
```

in `XournalView.h`. These should probably be `unique_ptr`. I tried to create `unique_ptr`s for the derived classes and `std::move` them to get `unique_ptr`s for the base classes (GeometryTool classes), but got a crash when executing. Maybe I forgot to change something elsewhere.

src/core/view/SetsquareView.cpp
```diff
@@ -2,216 +2,280 @@
...
-#include "model/Layer.h"                    // for Layer
-#include "model/Point.h"                    // for Point
-#include "model/Setsquare.h"                // for Setsquare
-#include "model/Stroke.h"                   // for Stroke
-#include "model/XojPage.h"                  // for XojPage
-#include "undo/InsertUndoAction.h"          // for InsertUndoAction
-#include "undo/UndoRedoHandler.h"           // for UndoRedoHandler
-#include "util/Rectangle.h"                 // for Rectangle
-#include "view/View.h"                      // for Context
+#include "gui/XournalView.h"          // for XournalView
```
bhennion, line 11:
```suggestion
#include "view/Repaintable.h"
```

src/core/view/SetsquareView.h
```diff
@@ -11,197 +11,104 @@
...
-#include "model/PageRef.h"    // for PageRef
 #include "model/Setsquare.h"  // for Setsquare
-#include "util/Point.h"       // for Point
+#include "util/DispatchPool.h"
+#include "view/overlays/BaseStrokeToolView.h"
+
+#include "GeometryToolView.h"
 
 class Stroke;
 class XojPageView;
```
bhennion, line 25:
```suggestion
```

bhennion (changes requested):
Below are a couple of details I noticed while skimming through the code. I have not look at every line of code.

I do have a more important structural remark: I feel like the very existence of gui/inputdevices/GeometryToolHandler (and derived classes) is problematic.

The content of gui/inputdevices/ should be an abstraction layer and no tool implementation should ever end up in this directory.
Of course, you created those classes because you need them: the input system is inadequate and does not handle multitouch as we'd need to avoid tool-specific code.
I'm not familiar enough with the input system to come up with a better structured solution at this point. I need to think.

(Also, we have to keep in mind that GTK4 sees a lot of changes in the input mechnisms, so porting to GTK4 will someday require extensive work on the input system.)

src/core/view/GeometryToolView.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+using namespace xoj::view;
+
+GeometryToolView::GeometryToolView(const GeometryTool* s, Repaintable* parent): ToolView(parent), s(s) {}
+
+GeometryToolView::~GeometryToolView() = default;
+
+void GeometryToolView::draw(cairo_t* cr) const {
+    xoj::util::CairoSaveGuard saveGuard(cr);
+    this->drawGeometryTool(cr);
```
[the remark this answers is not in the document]
rolandlo: The mask would also need resizing when the tool is rotated, right? So it would only improve performance when the tool or the page is moved. Do you think it is worth the effort?

rolandlo:
> I do have a more important structural remark: I feel like the very existence of gui/inputdevices/GeometryToolHandler (and derived classes) is problematic.
> 
> The content of gui/inputdevices/ should be an abstraction layer and no tool implementation should ever end up in this directory. Of course, you created those classes because you need them: the input system is inadequate and does not handle multitouch as we'd need to avoid tool-specific code. I'm not familiar enough with the input system to come up with a better structured solution at this point. I need to think.
> 

I agree that those files look misplaced. Maybe @LittleHuba has an opinion about how to deal with it.

> (Also, we have to keep in mind that GTK4 sees a lot of changes in the input mechnisms, so porting to GTK4 will someday require extensive work on the input system.)

We should probably get acquainted with the input mechanism in GTK4 as soon as possible and try to figure out what changes will be required when porting.

LittleHuba:
Historically the input system was only a fix to get rid of the old c-style input handling. In general I think it is simply impossible to get rid of all tool specific code in the input handling. Thus, for me the goal would be to at least separate the unspecific handling very strictly from the tool specific handling to get something like a layered system. How that would be possible, I honestly don't know yet. 
As far as I know, GTK4 goes more in line with other frameworks where each widget does it's own parsing of input and you have a bubbling down behavior. But I never looked deeper into that.

rolandlo:
Rebased and fixed merge conflict.

src/core/view/GeometryToolView.cpp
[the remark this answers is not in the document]
rolandlo: Sorry, it took a while, but I have figured it out now, see commit 45113f950c50791d6e759ebb6b91b3282ee96dfc.
The mask is drawn without (translation and) rotation. So the mask doesn't need redrawing when the geometry tool is rotated (or translated). It is translated and rotated right before `mask.paintTo`. @bhennion Can you take a look at it?

Commit c408a730e323b69e7dce0dafdc7629e2b923c6ff is independent of this PR. I just noticed that `zoomChanged` is fired even when the zoom has not changed at all. That meant that the mask needed resizing/redrawing when pinching to zoom with a scale increase/decrease that was below the threshold set in the preferences.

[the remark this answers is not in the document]
bhennion: Yes, that's what I had in mind. The code in 45113f9 looks good!

About https://github.com/xournalpp/xournalpp/commit/c408a730e323b69e7dce0dafdc7629e2b923c6ff, it may be worth splitting it to a separate PR, perhaps even destined to 1.1.3.

rolandlo:
Factored out everything related to the compass into PR #4443. Also rebased on master.
This feels ready now. @bhennion Can you take another look? Only the commits from "Apply reviewer suggestions" on have changed after your first review.

src/core/control/GeometryToolController.cpp
```diff
@@ -0,0 +1,118 @@
...
+    const auto undo = control->getUndoRedoHandler();
+    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, cross));
+
+    layer->addElement(cross);
+
+    const Rectangle<double> rect{cross->getX(), cross->getY(), cross->getElementWidth(), cross->getElementHeight()};
+    view->rerenderRect(rect.x, rect.y, rect.width, rect.height);
+}
+
+void GeometryToolController::addStrokeToLayer() {
```
bhennion, line 74: Is the document's mutex locked when this is called? If not, then it needs to be locked from `ensureLayerExists` to `layer->addElement`

src/core/control/GeometryToolController.h
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @brief the underlying geometry tool
+     */
+    GeometryTool* geometryTool;
+
+    /**
+     * @brief The stroke drawn
+     */
+    Stroke* stroke = nullptr;
```
bhennion, line 101: The Stroke is owned by the controller until it's added to the layer
```suggestion
    std::unique_ptr<Stroke> stroke;
```

src/core/control/SetsquareController.cpp
```diff
@@ -0,0 +1,119 @@
...
+    cairo_matrix_t matrix = geometryTool->getMatrix();
+    cairo_matrix_invert(&matrix);
+    cairo_matrix_transform_point(&matrix, &x, &y);
+    switch (leg) {
+        case HYPOTENUSE:
+            return utl::Point<double>(x, -y);
+        case LEFT_LEG:
+            return utl::Point<double>((y + x) / sqrt(2.), (y - x - geometryTool->getHeight()) / sqrt(2.));
+        case RIGHT_LEG:
+            return utl::Point<double>((y - x) / sqrt(2.), (y + x - geometryTool->getHeight()) / sqrt(2.));
```
bhennion, line 30: Nitpick
```suggestion
            return utl::Point<double>((y + x) / std::sqrt(2.), (y - x - geometryTool->getHeight()) / std::sqrt(2.));
        case RIGHT_LEG:
            return utl::Point<double>((y - x) / std::sqrt(2.), (y + x - geometryTool->getHeight()) / std::sqrt(2.));
```

src/core/control/SetsquareController.cpp
```diff
@@ -0,0 +1,119 @@
...
+        cairo_matrix_transform_point(&matrix, &qx, &qy);
+
+        stroke->addPoint(Point(qx, qy));
+    }
+    geometryTool->notify();
+}
+
+void SetsquareController::finalizeStroke() {
+    hypotenuseMax = NAN;
+    hypotenuseMin = NAN;
```
bhennion, line 108: Is this NAN value ever compared? Used in a `min` or `max`? Is so, then using `std::numeric_limits::max/lowest` seems more appropriate.

src/core/control/SetsquareController.cpp
```diff
@@ -0,0 +1,119 @@
...
+    addStrokeToLayer();
+}
+
+void SetsquareController::finalizeRadius() {
+    strokeAngle = NAN;
+    addStrokeToLayer();
+}
+
+auto SetsquareController::existsStroke() -> bool { return !std::isnan(hypotenuseMax) && !std::isnan(hypotenuseMin); }
```
bhennion, line 117: This is weird. Isn't `return stroke` more appropriate?

src/core/gui/XournalView.cpp
```diff
@@ -681,28 +685,33 @@ void XournalView::repaintSelection(bool evenWithoutSelection) {
...
-void XournalView::resetSetsquareView() {
-    GTK_XOURNAL(this->widget)->setsquareView.reset();
-    this->control->fireActionSelected(GROUP_SETSQUARE, ACTION_NONE);
-}
+auto XournalView::getGeometryToolController() -> GeometryToolController* { return geometryToolController; }
 
-auto XournalView::getSetsquareView() const -> SetsquareView* {
-    g_return_val_if_fail(this->widget != nullptr, nullptr);
-    g_return_val_if_fail(GTK_IS_XOURNAL(this->widget), nullptr);
+void XournalView::makeGeometryTool(std::string tool) {
```
bhennion, line 692: Wouldn't an enum type be better for the parameter here?

src/core/gui/XournalView.cpp
```diff
@@ -681,28 +685,33 @@ void XournalView::repaintSelection(bool evenWithoutSelection) {
...
+    auto view = getViewFor(control->getCurrentPageNo());
+    auto* zoomControl = this->getControl()->getZoomControl();
 
-    return GTK_XOURNAL(this->widget)->setsquareView.get();
+    if (tool == "setsquare") {
+        auto setsquare = new Setsquare();
+        view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoomControl));
+        this->geometryTool = setsquare;
+        this->geometryToolController = new SetsquareController(view, setsquare);
+        this->geometryToolHandler = new SetsquareInputHandler(this, this->geometryToolController);
```
bhennion, line 701: The controller should be owned by the Control instance. Not sure about the InputHandler. Also, the name `geometryToolHandler` is confusing. The distinction between controller and handler is not clear to me...

Overall, my strategy for determining "who owns what" is: ask yourself what will happen if and when we finally split PageView and XournalView into an MVC pattern. In this instance, it's not quite clear yet but it's a good thing to keep in mind.

Also, whoever owns them, they should be handled with a smart pointer.

src/core/gui/XournalView.h
```diff
@@ -183,6 +185,13 @@ class XournalView: public DocumentListener, public ZoomListener {
      */
     RepaintHandler* repaintHandler = nullptr;
 
+    /**
+     * Geometry tool, if active
+     */
+    GeometryTool* geometryTool = nullptr;
+    GeometryToolController* geometryToolController = nullptr;
+    GeometryToolHandler* geometryToolHandler = nullptr;
```
bhennion, line 193: Maybe moved to other classes (see comment above)

```suggestion
    std::unique_ptr<GeometryTool> geometryTool;
    std::unique_ptr<GeometryToolController> geometryToolController;
    std::unique_ptr<GeometryToolHandler> geometryToolHandler;
```

src/core/enums/ActionGroup.enum.h
```diff
@@ -56,7 +56,7 @@ enum ActionGroup {
 
     GROUP_HIGHLIGHT_POSITION,
 
-    GROUP_SETSQUARE,
+    GROUP_GEOMETRY_TOOL,
```
bhennion, line 59: Question: do we want to only ever have at most one geometric tool up?

src/core/control/Control.cpp
```diff
@@ -1352,11 +1348,11 @@ void Control::updateDeletePageButton() {
...
     clearSelectionEndText();
 
-    // if the current page contains the Setsquare, reset it
+    // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
-    auto setsquareView = win->getXournal()->getSetsquareView();
-    if (setsquareView && doc->indexOf(setsquareView->getPage()) == pNr) {
-        win->getXournal()->resetSetsquareView();
+    auto geometryToolController = win->getXournal()->getGeometryToolController();
+    if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
```
bhennion, line 1354: Not sure, but we may want the lock the document mutex for that

bhennion (commented):
Some partial review. I only reviewed 11/28 files for now.

One question: in the future, do we plan to have several geometric tools displayed at the same time?

src/core/control/SetsquareController.cpp
rolandlo: Maybe the naming is bad. You can have the `stroke` near the hypotenuse (`existsStroke`) or as a radius from a point to the middle point of the longest side (`existsRadius`)

src/core/gui/XournalView.cpp
rolandlo: Yes, I will change that.

src/core/gui/XournalView.cpp
rolandlo: `geometryToolHandler` could be renamed to `geometryToolInputHandler` if that's less confusing. The input handler is just for processing input. It asks the controller to actually control (move, scale, rotate ...) the model (geometry tool).
When migrating to Gtk4 we will just to modify the input handler, not the controller.

src/core/enums/ActionGroup.enum.h
rolandlo: Yes, at most one geometry tool. I see no benefit in having several geometry tools up at the same time. It's only distracting. On real paper you sometimes need two tools (ruler and set square) for parallel translation, but digitally that's the easiest thing to do (via one finger touch or arrow keys).

src/core/control/Control.cpp
rolandlo: Fine

src/core/gui/XournalView.h
rolandlo: I run into trouble with `unique_ptr`s (related to derived classes and casting), but I can try again.

src/core/control/SetsquareController.cpp
rolandlo: Yes, it is compared in `SetsquareController::updateStroke`. I can change that.

src/core/control/SetsquareController.cpp
rolandlo: Ok.

rolandlo:
> Some partial review. I only reviewed 11/28 files for now.
> 
> One question: in the future, do we plan to have several geometric tools displayed at the same time?

No, at least I don't. It's less distracting to switch (with one click).

src/core/control/GeometryToolController.cpp
rolandlo: Fine, I will lock it.

src/core/control/GeometryToolController.h
rolandlo: That looks unproblematic. I will change that.

src/core/control/GeometryToolController.cpp
rolandlo: Unfortunately the app freezes when I lock the document there.

src/core/control/Control.cpp
rolandlo: Done

src/core/gui/XournalView.cpp
rolandlo: Done

src/core/control/SetsquareController.cpp
rolandlo: Done

src/core/control/SetsquareController.cpp
rolandlo: Done

src/core/control/GeometryToolController.h
rolandlo: Done

src/core/gui/XournalView.cpp
bhennion: Yes, this would be less confusing I guess.

src/core/control/GeometryToolController.cpp
bhennion: So it means the lock is either already locked by the caller (I could not find where), or one of the functions called tries to lock the alread locked mutex. We need to figure this out.

src/core/control/SetsquareController.cpp
bhennion: I see. Then yes, the names could surely be improved. Something like existsEdgeStroke and existsRadialStroke maybe?

src/core/gui/XournalView.h
bhennion: Could you describe the problem you encountered? We might be able to help.

rolandlo: Strange. This time it worked at once. Can you please check?

src/core/gui/XournalView.cpp
rolandlo: Renaming is done.

src/core/control/SetsquareController.cpp
rolandlo: I like the suggestion. Renaming is done.

src/core/control/GeometryToolController.cpp
rolandlo: You're right, the already locked mutex gets locked. I will investigate who locked it.

rolandlo: The mutex when locked in the `GeometryToolController` gets locked again when calling
```
undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke.get()));
```
which calls `Control::updateWindowTitle()` via `UndoRedoHandler::fireUpdateUndoRedoButtons()` and `Control::undoRedoChanged()`.

If moving the `addUndoAction` past the mutex unlocking, there is no freeze anymore.

rolandlo: Done in https://github.com/xournalpp/xournalpp/pull/4211/commits/1ff09f070b02d5f44cddd3a0bffcf62e15055a04

bhennion: Wow! This is a good example of bad code structure. The window title has nothing to do with the undo-redo stack. There should absolutely not be any call to `updateWindowTitle()` in `undoRedoChanged()`...

rolandlo: When a document that has no changes compared to the saved version is modified, an asterix (*) is put into the window title (before the document name). When the modification is undone, the star is removed again. That's probably the reason, why the undo-redo stack was linked to the window title update.

src/core/gui/XournalView.cpp
rolandlo: > The controller should be owned by the Control instance. Not sure about the InputHandler. Also, the name `geometryToolHandler` is confusing. The distinction between controller and handler is not clear to me...
> 
> Overall, my strategy for determining "who owns what" is: ask yourself what will happen if and when we finally split PageView and XournalView into an MVC pattern. In this instance, it's not quite clear yet but it's a good thing to keep in mind.
> 
> Also, whoever owns them, they should be handled with a smart pointer.

The `geometryToolController` is now owned by the `Control` instance, the `geometryToolInputHandler` by the `InputContext` instance. That makes the most sense to me, as they are now where they are used.

rolandlo: The `geometryTool` is now owned by the `Control` instance. I see no reason to pass it forth and back.

rolandlo:
@bhennion I think I have applied all changes you requested. It cleared up quite some things. Thanks! I'll wait for the rest of your review now.

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+    const auto l = page->getSelectedLayer();
+    this->lines.clear();
+    for (const auto& e: l->getElements()) {
+        if (e->getType() == ELEMENT_STROKE) {
+            auto s = dynamic_cast<Stroke*>(e);
+            if (s->getPointCount() == 2) {
+                lines.push_back(s);
+            }
+        }
+    }
```
bhennion, line 234: I find this weird (even if the above comment about using `sequenceStart()` is resolved).

Filtering the entire list of elements everytime the user puts a finger on the screen is bad, performance-wise. I'm not sure how to handle the issue.

[ Also, for any filtering like that, the document's mutex needs to be locked: if an element is added to the layer by another thread, all the iterators could be invalidated, leading to a segfault ]

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+    const auto pos = Point(translationX + offset.x, translationY + offset.y);
+    double minDist = SNAPPING_DISTANCE_TOLERANCE;
+    const auto angleSetsquare = rotation;
+    double diffAngle{NAN};
+    for (const auto& l: lines) {
+        auto first = l->getPoint(0);
+        auto second = l->getPoint(1);
+        auto dist = Snapping::distanceLine(pos, first, second);
+        auto angleLine = std::atan2(second.y - first.y, second.x - first.x);
+        auto diff = std::remainder(angleLine - angleSetsquare, M_PI_2);
```
bhennion, line 260: I think to many `auto`'s make the code harder to understand.

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    const auto center = (this->priLastPageRel + this->secLastPageRel) / 2;
+    const auto shift = this->secLastPageRel - this->priLastPageRel;
+    const auto angle = atan2(shift.y, shift.x);
+
+    const auto offset = center - lastZoomScrollCenter;
+    controller->move(offset.x, offset.y);
+    const auto angleIncrease = angle - lastAngle;
+    const auto centerRel = (this->priLastPageRel + this->secLastPageRel) / 2.0;
```
bhennion, line 309: Same auto concern here

bhennion (changes requested):
Here are some more comments.
Mostly nitpicks, except my concerns about `GeometryToolInputHandler::sequenceStart()`. See the two related comments below.

src/core/control/GeometryToolController.h
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * @brief resizes the geometry tool by the factor f with respect to a given scaling center
+     * @param f the scaling factor
+     * @param cx the x-coordinate of the scaling center (in document coordinates)
+     * @param cy the y-coordinate of the scaling center (in document coordiantes)
+     */
+    void scale(double f, double cx, double cy);
+
+    /**
+     * @brief marks a point with a "x"
```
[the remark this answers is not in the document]
rolandlo: A picture says more than words:
grafik
Here the setsquare origin is marked. It can be useful for geometric constructions to mark the rotation center, even more so for the compass in the future.

src/core/control/GeometryToolController.h
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @brief the page view of the page with respect to which the geometry tool is initialized
+     */
+    XojPageView* getView() const;
+
+    /**
+     * @brief the page with respect to which the setsquare is initialized
+     */
+    PageRef getPage() const;
```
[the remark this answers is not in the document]
rolandlo: I tried to change that, but it resulted in a compiler warning and (when executing) in a crash.

```term
 warning: returning reference to temporary [-Wreturn-local-addr]
  119 | auto GeometryToolController::getPage() const -> const PageRef& { return view->getPage(); }
```

src/core/control/SetsquareController.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void SetsquareController::updateRadialStroke(double x, double y) {
+    stroke->deletePointsFrom(0);
+    const auto c = getPointForPos(0);
+    stroke->addPoint(Point(c.x, c.y));
+
+    const auto p = posRelToSide(HYPOTENUSE, x, y);
+    const auto rad = std::hypot(p.x, p.y);
+
+    if (rad >= geometryTool->getHeight() / std::sqrt(2.) - 1.15 || p.y > 0) {  // TODO
```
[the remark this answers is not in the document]
rolandlo: This is duplicating 

https://github.com/xournalpp/xournalpp/blob/fb63352f6d39bd1c5be712e31ff4529bbf241bf5/src/core/view/SetsquareView.cpp#L52-L52

There is probably a better way to do it, but I don't want to include `SetsquareView.h` in `SetsquareController.cpp`
Any idea?

src/core/control/GeometryToolController.h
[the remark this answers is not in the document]
bhennion: Ah right. PageView::getPage also returns a PageRef and not a reference to it. That's alright, just disregard my comment (or change PageView::getPage as well).

src/core/control/SetsquareController.cpp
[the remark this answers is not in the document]
bhennion: From what I could gather in a glance, you put all the lengths constants in SetSquareView. I think they should be part of the model (to which both the view and the controller have access). For instance DISTANCE_SEMICIRCLE_FROM_LEGS could be a member `static constexpr` in SetSquare. You could also imaging a static function radiusFromHeight() in SetSquare.

src/core/control/Control.cpp
```diff
@@ -646,21 +650,16 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
...
-                    setsquareView.reset(nullptr);
-                }
-                this->win->getXournal()->setSetsquareView(std::move(setsquareView));
-                fireActionSelected(GROUP_SETSQUARE, ACTION_SETSQUARE);
+            if (auto xournal = this->win->getXournal();
+                !this->geometryToolController ||
+                this->geometryToolController->getType() != GeometryToolType::SETSQUARE) {
+                resetGeometryTool();
+                makeGeometryTool(GeometryToolType::SETSQUARE);
+                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
```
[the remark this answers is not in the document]
rolandlo: I removed the page rendering and it still works fine. I had to pass the tool range of the setsquare as the last repaint range to the geometry tool constructor, since otherwise there were artifacts when translating the setsquare using the arrow keys.

src/core/control/Control.cpp
```diff
@@ -646,21 +650,16 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
...
             } else {
-                // hide setsquare
-                this->win->getXournal()->resetSetsquareView();
+                resetGeometryTool();
+                xournal->getViewFor(getCurrentPageNo())->rerenderPage();
```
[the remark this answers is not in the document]
rolandlo: Removed as well.

src/core/control/Control.cpp
```diff
@@ -1351,12 +1374,13 @@ void Control::updateDeletePageButton() {
...
+    // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
-    auto setsquareView = win->getXournal()->getSetsquareView();
-    if (setsquareView && doc->indexOf(setsquareView->getPage()) == pNr) {
-        win->getXournal()->resetSetsquareView();
+    doc->lock();
+    if (geometryToolController && doc->indexOf(geometryToolController->getPage()) == pNr) {
+        resetGeometryTool();
     }
+    doc->unlock();
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/control/Control.cpp
```diff
@@ -1053,6 +1052,30 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
...
+    auto view = this->win->getXournal()->getViewFor(getCurrentPageNo());
+    if (tool == GeometryToolType::SETSQUARE) {
+        auto setsquare = new Setsquare();
+        view->addOverlayView(std::make_unique<xoj::view::SetsquareView>(setsquare, view, zoom));
+        this->geometryTool = std::unique_ptr<GeometryTool>(setsquare);
+        this->geometryToolController = std::make_unique<SetsquareController>(view, setsquare);
+        std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
+                std::make_unique<SetsquareInputHandler>(this->win->getXournal(), geometryToolController.get());
+        geometryToolInputHandler->registerToPool(setsquare->getHandlerPool());
+        auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
```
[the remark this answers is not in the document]
rolandlo: Ok, done.

src/core/control/Control.cpp
```diff
@@ -1053,6 +1052,30 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
...
+        xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
+        fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_SETSQUARE);
+    }
+}
+
+void Control::resetGeometryTool() {
+    this->geometryToolController.reset();
+    this->geometryTool.reset();
+    auto xournal = GTK_XOURNAL(this->win->getXournal()->getWidget());
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/control/GeometryToolController.h
```diff
@@ -0,0 +1,102 @@
...
+    GeometryToolController(XojPageView* view, GeometryTool* geometryTool);
+    ~GeometryToolController();
+
+public:
+    /**
+     * @brief moves the geometry tool in x- and y-direction
+     * @param x the translation in x-direction (in document coordinates)
+     * @param y the translation in y-direction (in document coordinates)
+     */
+    void move(double x, double y);
```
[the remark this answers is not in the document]
rolandlo: Ok, I renamed it to `translate`. I used `move` before because it's the same in `EditSelection`.

src/core/control/GeometryToolController.h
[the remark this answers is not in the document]
rolandlo: Hopefully it's clear now.

src/core/control/GeometryToolController.h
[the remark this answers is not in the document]
rolandlo: I disregarded your comment. Didn't want to change stuff in other files, too.

src/core/control/SetsquareController.cpp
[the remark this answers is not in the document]
rolandlo: Done. I think the constants that are only used for rendering should remain in `SetsquareView` though.

src/core/control/SetsquareController.cpp
```diff
@@ -0,0 +1,119 @@
...
+    hypotenuseMin = std::numeric_limits<double>::max();
+    addStrokeToLayer();
+}
+
+void SetsquareController::finalizeRadialStroke() {
+    strokeAngle = NAN;
+    addStrokeToLayer();
+}
+
+auto SetsquareController::existsEdgeStroke() -> bool { return !(hypotenuseMax == std::numeric_limits<double>::min()); }
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/control/SetsquareController.h
```diff
@@ -0,0 +1,130 @@
...
+     */
+    bool existsRadialStroke();
+
+private:
+    /**
+     * @brief when a stroke aligned to the longest side (hypotenuse) of the setsquare is drawn, the minimal and maximal
+     * x-coordinates of the point to be drawn (with respect to an unrotated, and untranslated coordinate system) are
+     * saved in the variables hypotenuseMin and hypotenuseMax
+     */
+    double hypotenuseMax = std::numeric_limits<double>::min();
```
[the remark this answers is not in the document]
rolandlo: Thanks for the catch!

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+            // Set sequence data
+            sequenceStart(event);
+
+            // Even if zoom gestures are disabled,
+            // this is still the start of a sequence. Just
+            // don't start zooming.
+            if (zoomGesturesEnabled) {
+                zoomStart();
+                return true;
+            }
```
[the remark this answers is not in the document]
rolandlo: There was something wrong when zoom gestures were not enabled. I had not tried that out previously. As soon as you had two fingers on the screen the setsquare would not do anything. Now the setsquare will still rotate when zoom gestures are disabled, it just won't scale.

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+
+    if (event.type == MOTION_EVENT && this->primarySequence) {
+        if (this->primarySequence && this->secondarySequence && zoomGesturesEnabled) {
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+            zoomMotion(event);
+            return true;
+        } else if (event.sequence == this->primarySequence) {
+            scrollMotion(event);
+            return true;
+        } else if (this->primarySequence && this->secondarySequence) {
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+            sequenceStart(event);
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+            return true;
+        }
+    }
+
+    if (event.type == BUTTON_RELEASE_EVENT) {
+        if (event.sequence == this->primarySequence) {
+            // If secondarySequence is nullptr, this sets primarySequence
+            // to nullptr. If it isn't, then it is now the primary sequence!
+            this->primarySequence = this->secondarySequence;
+            this->secondarySequence = nullptr;
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+            this->priLastPageRel = this->secLastPageRel;
+        } else {
+            this->secondarySequence = nullptr;
+        }
+    }
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
```diff
@@ -0,0 +1,335 @@
...
+    return false;
+}
+
+auto GeometryToolInputHandler::handleKeyboard(InputEvent const& event) -> bool {
+    GdkEvent* gdkEvent = event.sourceEvent;
+    if (gdkEvent->type == GDK_KEY_PRESS) {
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
rolandlo: I added the document locking. 
Performance-wise I think there would have to be million or billions of elements on the layer to notice any difference. Under such circumstances other operations (like erasing) might become the bigger problem.
Whilst it would be possible to cache these lines and update the cache whenever needed, I think that would be more error prone and maybe not even be more performant.
Also note that users don't put a new finger on the screen every millisecond or so.

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
rolandlo: Ok, I replaced some `auto`s throughout this file.

src/core/gui/inputdevices/GeometryToolInputHandler.cpp
rolandlo: Ok

rolandlo:
Applied all reviewer suggestions and rebased on current master, while solving rebase conflicts. Also fixed an issue that occurred when zoom gestures were disabled.

src/core/control/SetsquareController.h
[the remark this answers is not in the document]
bhennion: Did you also change every other appearance of `std::numeric_limits<double>::min()` in the rest of the code (e.g. when comparing)?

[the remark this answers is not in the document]
rolandlo: Yes. The only occurence now is in `safe_casts.h` as a comment (which is completely unrelated to this PR).

src/core/control/GeometryToolController.cpp
```diff
@@ -0,0 +1,122 @@
...
+            stroke->setToolType(StrokeTool::PEN);
+            break;
+        case TOOL_HIGHLIGHTER:
+            stroke->setToolType(StrokeTool::HIGHLIGHTER);
+            break;
+        case TOOL_ERASER:
+            stroke->setToolType(StrokeTool::ERASER);
+            break;
+        default:  // Should not occur
+            break;
```
bhennion, line 116: Maybe print a warning here

src/core/model/GeometryTool.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+auto GeometryTool::getHandlerPool() const -> const std::shared_ptr<xoj::util::DispatchPool<GeometryToolInputHandler>>& {
+    return handlerPool;
+}
+
+auto GeometryTool::getStroke() const -> Stroke* { return this->stroke; }
+
+void GeometryTool::setStroke(Stroke* s) { this->stroke = s; }
+
+auto GeometryTool::computeRepaintRange(Range& rg) const -> Range {
```
bhennion, line 48: Are you purposefully modifying the parameter Range here?
```suggestion
auto GeometryTool::computeRepaintRange(Range rg) const -> Range {
```

bhennion (approved):
Two nitpicks. Otherwise, LGTM!

rolandlo:
> Two nitpicks. Otherwise, LGTM!

Thanks for the review once more! I have fixed the two issues. Now I will clean up the commit history a bit (without changing the total diff).

rolandlo:
Cleaned up the commit history a bit (with no functional changes) and rebased on current master. 

PR #4228, opened 2022-09-10
Fix attach mode issues

### LINKED ISSUE #3189: Fulll pdf path is saved regardless of mode.
**Affects versions :**
 - OS: Linux (doesn't matter)
 - X11 (doesn't matter)
 - Which version of libgtk do you use: 3.24.25 (doesn't matter)
 -  PPA (doesn't matter)
 
**Describe the bug**
The pdf background filename is saved in absolute mode, regardless of the mode it has been originally saved with previously. So if a document is created in attach mode, saved, reopened and saved again, the path to the pdf will now be in absolute mode.

**To Reproduce**
1. Open a pdf `test.pdf` with "Annotate pdf" and check the "attach to journal" checkbox on the bottom left of the file dialog
2. Edit the pdf and save the xopp-file `test.xopp`. Observe that a `test.xopp.bg.pdf` file is created in the same directory and the (ungzipped) xopp-file contains a tag 
```xml
<background type="pdf" domain="attach" filename="bg.pdf" pageno="1ll"/>
``` 
as expected
3. Open the xopp-file from the file manager or with "Open File"
4. Edit and save again
5. Observe that the (unzipped) xopp-file now contains a tag
```xml
<background type="pdf" domain="absolute" filename="full/path/to/pdf/test.xopp.bg.pdf" pageno="1ll"/>
``` 
instead of the attach mode

**Expected behavior**
A document that has been created in attach mode should remain in attach mode when reopend and saved again

**Additional context**
There is a related issue #2736, that relative pdf background filenames can be opened, but when they are saved, the relative path gets replaced by the full absolute path.

Joheyy:
Fixes #3189 and a new bug

**First commit (80e50c53f6801410cd0ad9cd98d6755407c4b629) is quite trivial. It fixes the following issue:**
The attach mode gets lost when loading and copying the Document. 

**Second commit (bf660d496ffec4b4d7212f817fac6a5f53984884) is more controversial:**
When opening a file with attach mode and you save it again then poppler fails to overwrite the pdf file and corrupts it. For this to happen you need the first commit to even be able to save a loaded xournal file in attach mode.

Strangely enough this doesn't happen when you create a new file and annotate a pdf in attach mode. Overwriting the pdf works just fine there. This suggests that something happens with the pdf when loading it.

The way I fixed this is pretty simple. I just don't overwrite the pdf file if it already exists because that's an unnecessary step anyway.

### DIFF
--- a/src/core/control/xojfile/LoadHandler.cpp
+++ b/src/core/control/xojfile/LoadHandler.cpp
@@ -465,6 +465,7 @@ void LoadHandler::parseBgPdf() {
                     pdfFilename = xournalFilepath.remove_filename() / pdfFilename;
                 }
             } else if (!strcmp("attach", domain)) {
+                attachToDocument = true;
                 // Handle old format separately
                 if (this->isGzFile) {
                     pdfFilename = (fs::path{xournalFilepath} += ".") += pdfFilename;
--- a/src/core/control/xojfile/SaveHandler.cpp
+++ b/src/core/control/xojfile/SaveHandler.cpp
@@ -2,6 +2,7 @@
 
 #include <cinttypes>  // for PRIx32, uint32_t
 #include <cstdio>     // for sprintf, size_t
+#include <filesystem> // for exists
 
 #include <cairo.h>                  // for cairo_surface_t
 #include <gdk-pixbuf/gdk-pixbuf.h>  // for gdk_pixbuf_save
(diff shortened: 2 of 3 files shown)


rolandlo (approved):
Thanks for the fix and the detailed analysis. It works fine and the fix makes a lot of sense to me.

rolandlo:
The PR is merged now. Thanks again @Joheyy. It's a really valuable fix.

PR #4253, opened 2022-09-25
fixed overriding attach mode when using "proposed pdf" in open dialog

Malex14:
This pr fixes a issue in  #4165, where in every file attach mode is enabled.

This issue started to appear after #4228 fixed the attach mode. Now the previous mode is kept as is.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2241,7 +2241,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         switch (res) {
             case USE_PROPOSED:
                 if (!proposedPdfFilepath.empty()) {
-                    loadHandler.setPdfReplacement(proposedPdfFilepath, true);
+                    loadHandler.setPdfReplacement(proposedPdfFilepath, loadHandler.isAttachedPdfMissing());
                     loadedDocument = loadHandler.loadDocument(filepath);
                 }
                 break;
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2206,21 +2206,19 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
 #endif
         std::string msg;
         if (loadHandler.isAttachedPdfMissing()) {
-            msg = FS(_F("The attached background file \"{1}\" could not be found. It might have been moved, "
-                        "renamed or deleted.") %
-                     filename);
+            msg = FS(_F("The attached background file could not be found. It might have been moved, "
+                        "renamed or deleted."));
         } else {
             msg = FS(_F("The background file \"{1}\" could not be found. It might have been moved, renamed or "
-                        "deleted.") %
-                     filename);
+                        "deleted.\nIt was last seen at: \"{2}\"") %
+                     filename % parentFolderPath);
         }
-
-        msg += FS(_F("\nIt was last seen at: \"{1}\"") % parentFolderPath);
 
         // try to find file in current directory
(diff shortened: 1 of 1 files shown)


rolandlo:
Can you please describe the issue more precisely?

Malex14:
Currently when you open a journal file, where the corresponding pdf is not found, but a replacement is proposed (added by #4165), the attach mode will always get enabled. With this pr the attach mode will stay as it is.

rolandlo (approved):
Thanks for the bug description and the fix. I have verified that your PR does what it promises. Nice fix! 
In case of an attached pdf background that cannot be found (because it is not in the same folder anymore as the .xopp-file), the situation is currently not ideal, since the dialog will say something like this:
grafik
The proposed PDF is not useful. But that is far less important (since for an attached pdf background the user should not forget to move the attached pdf background with the document anyway) could be fixed in a follow up PR. 

So unless there are any objections I suggest to merge this PR in 24 hours. It looks urgent to me.

Malex14:
The fix for the issue you mentioned should be very simple. There is a check in the code whether the proposed file exits, but since the filename is empty when attach mode is enabled, the proposed path points to a directory which is also a "file". The following diff shows the fix. Should I create a new pr for that or can I commit it to this pr?
```diff
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2219,8 +2219,9 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
 
         // try to find file in current directory
         auto proposedPdfFilepath = filepath.parent_path() / filename;
-        bool proposedPdfFileExits = fs::exists(proposedPdfFilepath);
-        if (proposedPdfFileExits) {
+        bool proposePdfFile = !loadHandler.isAttachedPdfMissing() && !fs::is_empty(proposedPdfFilepath) &&
+                              fs::exists(proposedPdfFilepath) && !fs::is_directory(proposedPdfFilepath);
+        if (proposePdfFile) {
             msg += FS(_F("\nProposed replacement file: \"{1}\"") % proposedPdfFilepath.string());
         }
         GtkWidget* dialog = gtk_message_dialog_new(getGtkWindow(), GTK_DIALOG_MODAL, GTK_MESSAGE_QUESTION,
@@ -2228,7 +2229,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
 
         enum dialogOptions { USE_PROPOSED, SELECT_OTHER, REMOVE, CANCEL };
 
-        if (proposedPdfFileExits) {
+        if (proposePdfFile) {
             gtk_dialog_add_button(GTK_DIALOG(dialog), _("Use proposed PDF"), USE_PROPOSED);
         }
         gtk_dialog_add_button(GTK_DIALOG(dialog), _("Select another PDF"), SELECT_OTHER);

```

rolandlo:
That looks like a good solution. Please add it to this PR in a second commit.

src/core/control/Control.cpp
```diff
@@ -2241,7 +2239,7 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
         switch (res) {
             case USE_PROPOSED:
                 if (!proposedPdfFilepath.empty()) {
-                    loadHandler.setPdfReplacement(proposedPdfFilepath, true);
+                    loadHandler.setPdfReplacement(proposedPdfFilepath, false);
```
Malex14, line 2242: I have set this to false, since a pdf is now only proposed when not in attach mode

rolandlo (approved):
Thanks for the quick fix. I will merge in 24 hours from now then, if no objections are raised.

PR #4255, opened 2022-09-26
Do nothing if the page background is 'Copy'

### LINKED ISSUE #4142: Invalid PageTypeFormat::Copy assigned to actual Page model
I (=@Technius) think I found why the copy format is being produced. Here's a simpler crash reproducer:

1. New document with 2 pages.
2. Use the page format dialog in the toolbar (not the menu) to set page type to copy.
3. Apply page format to current page.
4. Observe segfault.

The issue is in `PageBackgroundChangeController::commitPageTypeChange()`, which does not seem to have the correct handling for the copy page type.

The problem also affects `release-1.1`, so we will need to backport the corresponding fix.

_Originally posted by @Technius in https://github.com/xournalpp/xournalpp/issues/4127#issuecomment-1177231277_

WillNilges:
This is an attempt to fix #4142 by adding a check that straight-up avoids the crash by checking if the background on the current document is the Copy type, and refusing to set the background if that's the case.

### DIFF
--- a/src/core/control/PageBackgroundChangeController.cpp
+++ b/src/core/control/PageBackgroundChangeController.cpp
@@ -115,7 +115,11 @@ auto PageBackgroundChangeController::commitPageTypeChange(const size_t pageNum, 
     PageType origType = page->getBackgroundType();
 
     // Apply the new background
-    applyPageBackground(page, pageType);
+    if (pageType.format != PageTypeFormat::Copy)
+        applyPageBackground(page, pageType);
+    else
+        g_warning("Found 'Copy' page type. Doing nothing™.");
+
 
     control->firePageChanged(pageNr);
     control->updateBackgroundSizeButton();
--- a/src/core/control/PageBackgroundChangeController.cpp
+++ b/src/core/control/PageBackgroundChangeController.cpp
@@ -115,10 +115,11 @@ auto PageBackgroundChangeController::commitPageTypeChange(const size_t pageNum, 
     PageType origType = page->getBackgroundType();
 
     // Apply the new background
-    if (pageType.format != PageTypeFormat::Copy)
+    if (pageType.format != PageTypeFormat::Copy) {
         applyPageBackground(page, pageType);
-    else
+    } else {
         g_warning("Found 'Copy' page type. Doing nothing™.");
+    }
 
 
     control->firePageChanged(pageNr);
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -151,8 +151,9 @@ void PageTypeMenu::entrySelected(PageTypeInfo* t) {
     }
 
     // Disable "Apply to current page" if current format is "Copy."
-    if (this->menuEntryApply)
+    if (this->menuEntryApply) {
         gtk_widget_set_sensitive(this->menuEntryApply, t->page.format != PageTypeFormat::Copy);
+    }
 }
 
 void PageTypeMenu::setSelected(const PageType& selected) {
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -198,7 +198,7 @@ void PageTypeMenu::addApplyBackgroundButton(PageTypeApplyListener* pageTypeApply
                          this);
         // Do not initially activate this option if the "Copy" format is selected
         if (getSelected().format == PageTypeFormat::Copy) {
-            gtk_widget_set_sensitive(menuEntryApply, FALSE);
+            gtk_widget_set_sensitive(menuEntryApply, false);
         }
     }
 
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -130,11 +130,7 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
 
                              if (info.entry == togglebutton) {
                                  // Disable "Apply to current page" if current format is "Copy."
-                                 if (info.info->page.format == PageTypeFormat::Copy) {
-                                     gtk_widget_set_sensitive(self->menuEntryApply, FALSE);
-                                 } else {
-                                     gtk_widget_set_sensitive(self->menuEntryApply, TRUE);
-                                 }
+                                 gtk_widget_set_sensitive(self->menuEntryApply, info.info->page.format != PageTypeFormat::Copy);
                                  self->entrySelected(info.info);
                                  break;
                              }
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -127,7 +127,6 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
                          }
 
                          for (MenuCallbackInfo& info: self->menuInfos) {
-
                              if (info.entry == togglebutton) {
                                  self->entrySelected(info.info);
                                  break;
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -181,6 +181,11 @@ void PageTypeMenu::hideCopyPage() {
  */
 void PageTypeMenu::addApplyBackgroundButton(PageTypeApplyListener* pageTypeApplyListener, bool onlyAllMenu,
                                             ApplyPageTypeSource ptSource) {
+    if (this->menuEntryApply) {
+        g_warning("Button 'Apply to current page' already exists!");
+        return;
+    }
+
     this->pageTypeApplyListener = pageTypeApplyListener;
     this->pageTypeSource = ptSource;
 
@@ -191,7 +196,7 @@ void PageTypeMenu::addApplyBackgroundButton(PageTypeApplyListener* pageTypeApply
     menuY++;
 
     if (!onlyAllMenu) {
-        menuEntryApply = createApplyMenuItem(_("Apply to current page"));
+        this->menuEntryApply = createApplyMenuItem(_("Apply to current page"));
         gtk_menu_attach(GTK_MENU(menu), menuEntryApply, 0, PREVIEW_COLUMNS, menuY, menuY + 1);
         menuY++;
         g_signal_connect(menuEntryApply, "activate", G_CALLBACK(+[](GtkWidget* menu, PageTypeMenu* self) {
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -127,7 +127,14 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
                          }
 
                          for (MenuCallbackInfo& info: self->menuInfos) {
+
                              if (info.entry == togglebutton) {
+                                 // Disable "Apply to current page" if current format is "Copy."
+                                 if (info.info->page.format == PageTypeFormat::Copy) {
+                                     gtk_widget_set_sensitive(self->menuEntryApply, FALSE);
+                                 } else {
+                                     gtk_widget_set_sensitive(self->menuEntryApply, TRUE);
+                                 }
                                  self->entrySelected(info.info);
                                  break;
                              }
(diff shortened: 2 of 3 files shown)


WillNilges:
I'm not 100% sure if this is how we want to patch this issue, but I figure it's a start.

src/core/control/PageBackgroundChangeController.cpp
```diff
@@ -115,7 +115,11 @@ auto PageBackgroundChangeController::commitPageTypeChange(const size_t pageNum,
     PageType origType = page->getBackgroundType();
 
     // Apply the new background
-    applyPageBackground(page, pageType);
+    if (pageType.format != PageTypeFormat::Copy)
+        applyPageBackground(page, pageType);
+    else
+        g_warning("Found 'Copy' page type. Doing nothing™.");
```
bhennion, line 121:
```suggestion
    if (pageType.format != PageTypeFormat::Copy) {
        applyPageBackground(page, pageType);
    } else {
        g_warning("Found 'Copy' page type. Doing nothing™.");
    }
```
cf coding conventions

bhennion (commented):
Yeah, this is a workaround rather than a fix of the underlying issue (namely the very existence of `PageTypeFormat::Copy`).

I guess it's ok as a temporary fix, but ideally, we should refactor and, as you mentionned on gitter, disable the option "Apply to current page" if the "copy current page type" mode is selected.

WillNilges:
The former sounds like it'll take _quite_ a bit of time, but for the latter, lemme mess around with the UI and see if I can disable the menu option. Probably not a bad idea to keep this patch in there _just in case 🙃 _

bhennion:
> The former sounds like it'll take _quite_ a bit of time, but for the latter, lemme mess around with the UI and see if I can disable the menu option. Probably not a bad idea to keep this patch in there _just in case upside_down_face_

I think a quick fix (disabling the menu option + the safety test) is a good solution for the release-1.1 branch (so to appear in 1.1.2). So once you have figured out how to disable the menu entry, would you mind rebasing this onto release-1.1?
Note that on release-1.1, the bug does exists but does not crash the app. It does create invalid documents (and files) though.

A more consistent refactoring (removing the Copy value) would be nice, but as you say, will take longer. It would also be directed at the master branch (so for version 1.2 or so). Let's postpone this to a future PR.

WillNilges:
Just a quick update on this: I have been pawing through the GUI code trying to make sense of it, as I'm pretty unfamiliar with GTK. Still working on it.

WillNilges:
@bhennion I saw your recent work to refactor the main menubar. It looks stellar! I suppose that means that this PR is obsolete?

bhennion:
Well, it's not so clear whether #4508 will get merged before the next release is out, and it'd be good if said release had a fix for this bug.
We can still merge this, as a temporary workaround to be removed in #4508.

WillNilges:
@bhennion I think I figured out a solution and have now disabled the menu entry. Can you review this again and tell me if this approach is sane? If so, maybe we can get this rebased and merged for 1.2?

WillNilges:
CC @Technius

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -127,7 +127,14 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
                          }
 
                          for (MenuCallbackInfo& info: self->menuInfos) {
+
```
Technius, line 130: nit: unnecessary whitespace

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -127,7 +127,14 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
...
                              if (info.entry == togglebutton) {
+                                 // Disable "Apply to current page" if current format is "Copy."
+                                 if (info.info->page.format == PageTypeFormat::Copy) {
+                                     gtk_widget_set_sensitive(self->menuEntryApply, FALSE);
+                                 } else {
+                                     gtk_widget_set_sensitive(self->menuEntryApply, TRUE);
+                                 }
```
Technius, line 137: nit: this can be simplified to
```suggestion
                                 gtk_widget_set_sensitive(self->menuEntryApply, info.info->page.format != PageTypeFormat::Copy);
```

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -186,13 +193,17 @@ void PageTypeMenu::addApplyBackgroundButton(PageTypeApplyListener* pageTypeApply
     menuY++;
 
     if (!onlyAllMenu) {
-        GtkWidget* menuEntryApply = createApplyMenuItem(_("Apply to current page"));
+        menuEntryApply = createApplyMenuItem(_("Apply to current page"));
```
Technius, line 196: Why was this converted into a member field?

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -186,13 +193,17 @@ void PageTypeMenu::addApplyBackgroundButton(PageTypeApplyListener* pageTypeApply
...
         gtk_menu_attach(GTK_MENU(menu), menuEntryApply, 0, PREVIEW_COLUMNS, menuY, menuY + 1);
         menuY++;
         g_signal_connect(menuEntryApply, "activate", G_CALLBACK(+[](GtkWidget* menu, PageTypeMenu* self) {
                              self->pageTypeApplyListener->applySelectedPageBackground(false, self->pageTypeSource);
                          }),
                          this);
+        // Do not initially activate this option if the "Copy" format is selected
+        if (getSelected().format == PageTypeFormat::Copy) {
+            gtk_widget_set_sensitive(menuEntryApply, FALSE);
```
Technius, line 205: nit: use C++'s `false` keyword instead of C's FALSE macro:
```suggestion
            gtk_widget_set_sensitive(menuEntryApply, false);
```

Technius (changes requested):
Seems to work fine based on some light testing. I have a few nitpicks about the code.

src/core/control/pagetype/PageTypeMenu.cpp
Technius: Note that `addApplyBackgroundButton` can be called in multiple places, so this could potentially be overridden.

Technius (changes requested):
* Ok, I see that `menuEntryApply` was added as a member because we need to be able to toggle it when the page type changes. Please add a doc comment to the member that explains this.
* Also, the "Apply All" buttons in the toolbar dropdown is not disabled yet.
* One last thing: instead of toggling the sensitivity of menuEntryApply in the toggled event callback, we should be toggling the sensitivity in `entrySelected()`, to cover all cases (e.g., when `setSelected()` is called instead of a user interaction).

WillNilges:
Thanks for the review. I'm gonna have limited availability for the next week but all that sounds pretty quick to do.

WillNilges:
Ok, so I fixed the nitpicks, added a doc that explains why that variable exists, and moved the sensitivity toggle to `entrySelected()`.

Originally, I didn't do this for `applyAll()` because I don't recall the issue mentioning it, and I didn't observe a crash. Regardless, I tried to add it, basically in the same way as I did "Apply Current" but I keep getting crashes on startup that I'm having trouble debugging, and when it does launch correctly, the behavior does not seem to work. Do you have suggestions for that?

Also,
> Note that addApplyBackgroundButton can be called in multiple places, so this could potentially be overridden.
Looking at the code, I don't _think_ this can get overridden anymore? Not really sure how to deal with this issue :/

bhennion:
I don't think the "Apply to all" button should be greyed out when the copy option is selected. Clicking it should copy the current page's background to all other pages. (We may grey it out if copy is selected and the doc has only 1 page, but it's probably not worth the trouble)

About overriding the button with several calls to

bhennion:
I don't think the "Apply to all" button should be greyed out when the copy option is selected. Clicking it should copy the current page's background to all other pages. (We may grey it out if copy is selected and the doc has only 1 page, but it's probably not worth the trouble)

About overriding the button with several calls to addApplyBackgroundButton: we can just add a test in addApplyBackgroundButton detecting if a button already exists, and do nothing but issueing a warning in this case.

bhennion:
I don't think the "Apply to all" button should be greyed out when the copy option is selected. Clicking it should copy the current page's background to all other pages. (We may grey it out if copy is selected and the doc has only 1 page, but it's probably not worth the trouble)

About overriding the button with several calls to addApplyBackgroundButton: we can just add a test in addApplyBackgroundButton detecting if a button already exists, and do nothing but issueing a warning in this case.

bhennion:
I don't think the "Apply to all" button should be greyed out when the copy option is selected. Clicking it should copy the current page's background to all other pages. (We may grey it out if copy is selected and the doc has only 1 page, but it's probably not worth the trouble)

About overriding the button with several calls to addApplyBackgroundButton: we can just add a test in addApplyBackgroundButton detecting if a button already exists, and do nothing but issueing a warning in this case.

WillNilges:
@bhennion, Added a new commit. Is something like this what you're after?

bhennion:
> @bhennion, Added a new commit. Is something like this what you're after?

Yes, perfect! I'll give this a try later today (hopefully) and after that, it'll LGTM.

bhennion:
I pulled your branch, rebased it on master (no conflicts) and build it, but I get a SegFault on startup. Can you rebase and fix this?

Here is the log:
```
Date: Wed Jan  4 22:11:23 2023
Error: signal 11
[bt]: (0) ./install/bin/xournalpp(+0x3648ec) [0x5604b4b618ec]
[bt]: (1) /usr/lib/libc.so.6(+0x38a00) [0x7f886fb69a00]
[bt]: (2) /usr/lib/libgobject-2.0.so.0(g_type_check_instance_is_a+0x39) [0x7f8870472fd9]
[bt]: (3) /usr/lib/libgtk-3.so.0(gtk_widget_set_sensitive+0x47) [0x7f8870b3e307]
[bt]: (4) ./install/bin/xournalpp(_ZN12PageTypeMenu13entrySelectedEP12PageTypeInfo+0x145) [0x5604b4c94ae3]
[bt]: (5) ./install/bin/xournalpp(_ZN12PageTypeMenu11setSelectedERK8PageType+0x92) [0x5604b4c94b90]
[bt]: (6) ./install/bin/xournalpp(_ZN12PageTypeMenu15loadDefaultPageEv+0x82) [0x5604b4c9420a]
[bt]: (7) ./install/bin/xournalpp(_ZN12PageTypeMenuC2EP15PageTypeHandlerP8Settingsbb+0xd8) [0x5604b4c9414c]
[bt]: (8) ./install/bin/xournalpp(_ZSt11make_uniqueI12PageTypeMenuJRP15PageTypeHandlerRP8SettingsbbEENSt8__detail9_MakeUniqIT_E15__single_objectEDpOT0_+0x90) [0x5604b4c53542]
[bt]: (9) ./install/bin/xournalpp(_ZN7ControlC1EP13_GApplicationP15GladeSearchpath+0x530) [0x5604b4c412ac]
[bt]: (10) ./install/bin/xournalpp(_ZSt11make_uniqueI7ControlJRP13_GApplicationP15GladeSearchpathEENSt8__detail9_MakeUniqIT_E15__single_objectEDpOT0_+0x58) [0x5604b4b6b45a]
[bt]: (11) ./install/bin/xournalpp(+0x369519) [0x5604b4b66519]
[bt]: (12) /usr/lib/libgobject-2.0.so.0(g_closure_invoke+0x170) [0x7f887044e210]
[bt]: (13) /usr/lib/libgobject-2.0.so.0(+0x41ea8) [0x7f887047bea8]
[bt]: (14) /usr/lib/libgobject-2.0.so.0(g_signal_emit_valist+0x1045) [0x7f887046bf75]
[bt]: (15) /usr/lib/libgobject-2.0.so.0(g_signal_emit+0x94) [0x7f887046c204]
[bt]: (16) /usr/lib/libgio-2.0.so.0(g_application_register+0x248) [0x7f88705b3a18]
[bt]: (17) /usr/lib/libgio-2.0.so.0(+0xd3190) [0x7f88705b4190]
[bt]: (18) /usr/lib/libgio-2.0.so.0(g_application_run+0x14c) [0x7f88705bf6cc]
[bt]: (19) ./install/bin/xournalpp(_ZN11XournalMain3runEiPPc+0x781) [0x5604b4b68a09]
[bt]: (20) ./install/bin/xournalpp(main+0x25) [0x5604b4b6175e]
[bt]: (21) /usr/lib/libc.so.6(+0x23290) [0x7f886fb54290]
[bt]: (22) /usr/lib/libc.so.6(__libc_start_main+0x8a) [0x7f886fb5434a]
[bt]: (23) ./install/bin/xournalpp(_start+0x25) [0x5604b4b61665]
```

WillNilges:
Oh I see what happened. I'm not explicitly setting the button to `nullptr` in the class, and then rely on it later on in a check that's supposed to _prevent_ a crash. Did that, and the crash is gone.

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -127,6 +127,7 @@ void PageTypeMenu::addMenuEntry(PageTypeInfo* t) {
                          }
 
                          for (MenuCallbackInfo& info: self->menuInfos) {
+
```
bhennion, line 130:
```suggestion
```

src/core/control/pagetype/PageTypeMenu.cpp
```diff
@@ -149,6 +150,10 @@ void PageTypeMenu::entrySelected(PageTypeInfo* t) {
     if (listener != nullptr) {
         listener->changeCurrentPageBackground(t);
     }
+
+    // Disable "Apply to current page" if current format is "Copy."
+    if (this->menuEntryApply)
+        gtk_widget_set_sensitive(this->menuEntryApply, t->page.format != PageTypeFormat::Copy);
```
bhennion, line 156:
```suggestion
    if (this->menuEntryApply) {
        gtk_widget_set_sensitive(this->menuEntryApply, t->page.format != PageTypeFormat::Copy);
    }
```

bhennion (approved):
Just two nitpicks. Otherwise, LGTM!

bhennion:
LGTM! @Technius do you want to review this once more, or can we merge?

bhennion:

PR #4258, opened 2022-09-29
Plug leaking Stroke in StrokeHandler::strokeRecognizerDetected

bhennion:
This PR plugs a memory leak happening when the stroke recognizer replaces the stroke.
Esssentially, StrokeHandler::strokeRecognizerDetected clones the recognized stroke (i.e. the result of the recognition process) and adds the clone to the layer. The recognized stroke (not the clone) is then never deleted, and is no longer referenced anywhere once the StrokeRecognizerResult gets deleted.

This PR just removes the clone and directly adds the recognized stroke to the layer (which then owns it).


rolandlo (approved):
LGTM

bhennion:
Thanks @rolandlo for the review!

Issue #3412, opened 2021-09-24
Make Xournal++ revision control system (Git) friendly


xax:
## Suggestion
A plain text and line based sidecar file format would be very useful to facilitate PDF annotation suited for revision control systems like *Git*, as it would be quite well *diff*'able and *merge*able.

## Motivation: Scenarios and use cases
Consider annotating homework, students' assignments or scientific paper drafts with *Xournal++* and using a remote *Git* repository to wither do it collaboratively or from different devices (may be from the office and while travelling on your laptop).

The repository might contain the (mostly binary) PDF and the accompanying `.xopp` (*GZip*'ed XML, thus effectively binary, too). The files being opaque Blobs to Git has two downsides:
1. revision storage will be ineffective (entire rewrites will be stored instead of deltas),
2. changes to annotations cannot be merged, as *Git* can identify and merge – even conflict resolve – changes to lines of plain text, but `.xopp` as is is just a block of binary data in *Git*'s eyes.

Regarding the PDF there is nothing *Xournal++* can do – and the PDF won't be changed, anyways, as the annotations go into the sidecar `.xopp` file.

But if the annotations were in a plain text format, possibly each single annotation object in one line, it would be easy to merge annotations with standard merge tools (and 3-way merge strategies from revision control systems) – to a certain extent that is, i.e. conflicting changes to the same line/annotation object will still be difficult or impossible to automatically be resolved.

## Opportunities
While there seems to be a project to implement another file format for *Xournal++* described in #1515 that has seemingly rather opposite goals – i.e. a unified blob container (ODF like zip file) encapsulating meta data + annotations + PDF – that should provide the opportunity to decouple *Xournal++* from its storage implementation.

## Alternative solutions
The current `.xopp`, for what I understand, is basically an XML file *Gzip*'ed. So you might try to fiddle with *Git* filters to achieve a similar functionality as described (you find hints at that in other #issues or discussions). But that is error prone and not practically, if different collaborators have to set up each of their machines right to get that to work.

rolandlo:
Have you seen [the wiki page](https://github.com/xournalpp/xournalpp/wiki/Git-Diff-Filter-for-Xournal)?

xax:
Thanks, yes, I saw that. (→ ‘Alternative solutions“)

xax:
Note to myself and to whom it may concern:

A fast first progress on this issue could be established implementing the following two changes:

[...]

PR #4262, opened 2022-09-30
Ensure `filename` in xopp format is always saved as a relative path to the directory it resides in

### LINKED ISSUE #2638: Relative path to pdf (not the attach mode)
**My Situation**
I am using Xournal to annotate textbooks on Linux and Windows, on 3 different computers. I have these textbooks (as pdf format) and xopp files synced under Dropbox, but in different subfolders. Eg. `Dropbox/textbook/` for pdf, and `Dropbox/note/` for xopp files, because I don't want to mess them together.

The problem is, the xopp format saves the absolute path to pdf file, but the `Dropbox` locations vary on different platforms.
For me, it is `/home/xxx/Dropbox` on Linux, and `C:/User/xxx/Dropbox` on Windows, so my xopp file doesn't open on another computer.

But this problem would be solved if we can save the relative path. Because the `Dropbox` folder have a unique structure on all platforms, the pdf file can be referred by `../textbook/foo.pdf`. This will work everywhere.

**Additional context**
I don't need to make my files portable, so the attach mode is not the solution. Also it consumes additional space.

### LINKED ISSUE #2736: PDF background not saved in .xopp / not loaded from absolute path

**Affects versions :**
 - OS: Manjaro
     3.24.24
 - Version of Xournal++: 01107eb2
 - Installation method: AUR

**Describe the bug** 
After moving the .xopp file and the PDF to another folder, it cannot find the PDF background unless you put it back into the folder where it was when the .xopp was created.
 
**To Reproduce**
1. Open PDF with Xournal++
2. Save as .xopp
3. Move both files to another folder
4. Open .xopp again

**Expected behavior**
Background not found

**Desired behavioir**
Save PDF file in .xopp file or load from absolute path. I think the first is the better solution.

### LINKED ISSUE #2774: Customized load path for pdf annotations
It'd be nice if I can have one or multiple load paths for pdf annotations.

Allow user to add load path to xournalpp. Upon launching a pdf file, xournalpp will search `./` and `/load/path/1` `/load/path/2`.. in sequence.

### LINKED ISSUE #3273: Annotated pdf background not found if not residing in the same directory

**Affects versions :**
 - OS: Archlinux
 - (Linux only) i3 under X11
 - Which version of libgtk do you use:  3.24.30
 - Version of Xournal++: 1.1.0
 - Installation method: official Arch-repo package

**Describe the bug**

Annotating a pdf file does not save the full path of the pdf-background in the annotated xopp file.

**To Reproduce**
Goal: You wish to annotate ./abc.pdf and save the annotation in a subdir as xopp/abc.pdf.xopp.

1. xournal abc.pdf & 
2. # You annotate and save as:
    File->Save as->xopp/abc.pdf.xopp and close xournalpp
3. cd xopp/
4. xournalpp abc.pdf.xopp & 
5. # xournalpp throws an error: 
        "The background fiile abc.pdf could not be found, .... etc. It was last seen at: <empty>. Choose another PDF, etc.",
    # and creates a
       ~/.config/xournalpp/emergencysave.xopp

**Expected behavior**
   Annotated pdf-background should be saved with a full pathname in the annotating xopp-file, as in versions 1.0.*

**Additional context**
   Default filename suggestion for  an annotation of %.pdf should be %.pdf.xopp (and not %.xopp), as in versions 1.0.*

### LINKED ISSUE #3880: Bad way of storing path of background PDF file inside .xopp files?
Hi,

There is a problem in the way "background" PDF file paths are stored in `.xopp` files. When reopening a `.xopp` file from another directory than the one in which the `.xoop` was created results in the "background" PDF not being found and Xournalpp asking for another path. 

Steps to reproduce the problem:
- Get a PDF file, let's say `test.pdf`
  - `wget -U "Mozilla/35.0"https://arxiv.org/pdf/2112.13896.pdf -O test.pdf`
- Put it in some `subdir/` sub directory of `.`
  - `mkdir subdir && mv test.pdf subdir/`
- Stay in `.`
- Open `dir/test.pdf` in `xournalpp`
  - `xournalpp dir/test.pdf`
- Create some annotations
- Save the result in `test.xopp` file
- Move to `subdir`
  - `cd subdir/`
- Try to open `test.xopp` in `xournal` again
  - `xournal test.xopp`
 
**Actual Result**
xournalpp will complain that "background" PDF file cannot be found and prompt for a new location

**Expected result**
xournalpp opens the files (xopp & pdf) correctly

**Problem**
From my simple understanding (looking at the path `xournalpp` is trying to access), it seems that the path stored in the `.xopp` file is relative to the *current directory from which `xournalpp` is opened*, instead of **relative to where the background PDF is with respect to the path of the newly created `.xopp` file**.

hyperupcall:
There are several issues stemming from the fact that annotated PDF backgrounds break when:

1. Renaming / moving containing directories because `filename` is stored as an absolute path
2. Saving a `.xopp` file different than the `$PWD` when opening one from the command line (passing it as $1)
  - This bug doesn't surface in the common case because xournal is usually launched through a launcher, where the `$PWD` is equivalent to `$HOME`, which is nearly always the common parent of xopp and pdf files.
  
Specifically, this
- Closes #3880 (relative path is now relative with respect to the newly created `.xopp` file, rather than `$PWD`)
- Closes #3273 (this particular issue is a specialization of the previous issue)
- Closes #2736 (relative paths are guaranteed to be generated, since we call `std::filesystem::relative` with two absolute paths, which uses [`lexically_relative`](https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal) under the hood after canonicalization
- Closes #2638 (same as previous, the issue seems stale too)
- Closes #2774 (see [comment](https://github.com/xournalpp/xournalpp/pull/4262#issuecomment-1408067035))

One of the issues floated around [this changeset](https://github.com/emorice/xournalpp/commit/6bdc96d427a6d68acafb54c815d6e8abe9f83c83), which I unfortunately found only after I did my on work - which appears to do nearly the same thing except:
- I applied my changes to not just background PDFs, but for images too. (wherever I saw a `background->setAttrib("domain", "absolute")` near a `background->setAttrib("filename", ...)` in the save handler, I performed this change)

Validation method:
- manual testing

### DIFF
--- a/src/util/PathUtil.cpp
+++ b/src/util/PathUtil.cpp
@@ -30,16 +30,17 @@ constexpr auto CONFIG_FOLDER_NAME = "xournalpp"sv;
 #include <windows.h>
 
 auto Util::getLongPath(const fs::path& path) -> fs::path {
-    DWORD wLongPathSz = GetLongPathNameW(path.c_str(), nullptr, 0);
+    auto asWString = path.wstring();
+    DWORD wLongPathSz = GetLongPathNameW(asWString.c_str(), nullptr, 0);
 
     if (wLongPathSz == 0) {
         return path;
     }
 
     std::wstring wLongPath(wLongPathSz, L'\0');
-    GetLongPathNameW(path.c_str(), wLongPath.data(), static_cast<DWORD>(wLongPath.size()));
+    GetLongPathNameW(asWString.c_str(), wLongPath.data(), static_cast<DWORD>(wLongPath.size()));
     wLongPath.pop_back();
-    return fs::path(std::move(wLongPath));
+    return {std::move(wLongPath)};
 }
 #else
 auto Util::getLongPath(const fs::path& path) -> fs::path { return path; }
@@ -271,9 +272,7 @@ auto Util::ensureFolderExists(const fs::path& p) -> fs::path {
 }
 
 auto Util::normalizeAssetPath(const fs::path& asset_path, const fs::path& base) -> std::string {
-    // fs::relative() may return an empty or malformed path if the root_path()s are different
-    return (asset_path.root_path() == base.root_path() ?
-            fs::relative(asset_path, base) : fs::weakly_canonical(asset_path)).generic_u8string();
+    return fs::proximate(asset_path, base).generic_u8string();
 }
 
 auto Util::isChildOrEquivalent(fs::path const& path, fs::path const& base) -> bool {
--- a/test/unit_tests/util/PathTest.cpp
+++ b/test/unit_tests/util/PathTest.cpp
@@ -10,6 +10,8 @@
  */
 
 
+#include <filesystem>
+
 #include <gtest/gtest.h>
 
 #include "util/PathUtil.h"
@@ -121,19 +123,40 @@ TEST(UtilPath, testClearExtensions) {
 TEST(UtilPath, normalizeAssetPath) {
     auto p = fs::path();
 
-#if _WIN32
+#ifdef _WIN32
+    // Test \\ to / conversion + not relative
+
+    // Test, if the implementation of std::filesystem::proximate and std::filesystem::relative resolves "/" and "\\" to
+    // the same path
+    EXPECT_EQ(std::filesystem::proximate("D:/home/freddy", "C:/home"),
+              std::filesystem::proximate("D:\\home\\freddy", "C:\\home"));
+    EXPECT_EQ(std::filesystem::relative("D:/home/freddy", "C:/home"),
+              std::filesystem::relative("D:\\home\\freddy", "C:\\home"));
+
     p = Util::normalizeAssetPath("C:\\dir\\file.txt", "D:");
-    EXPECT_EQ(string("C:\\dir\\file.txt"), p.string());
-
+    EXPECT_EQ(string("C:/dir/file.txt"), p.string());
     p = Util::normalizeAssetPath("C:\\dir\\file.txt", "D:\\base");
-    EXPECT_EQ(string("C:\\dir\\file.txt"), p.string());
+    EXPECT_EQ(string("C:/dir/file.txt"), p.string());
 
-    // do not return empty if asset_path is relative
-    p = Util::normalizeAssetPath("../dir/file.txt", "D:");
-    EXPECT_TRUE(!p.empty());
+    p = Util::normalizeAssetPath("C:/dir/file.txt", "D:");
+    EXPECT_EQ(string("C:/dir/file.txt"), p.string());
+    p = Util::normalizeAssetPath("C:/dir/file.txt", "D:/base");
+    EXPECT_EQ(string("C:/dir/file.txt"), p.string());
 
-    p = Util::normalizeAssetPath("../dir/file.txt", "D:\\base");
-    EXPECT_TRUE(!p.empty());
+    // Test \\ to / conversion + relative
+    p = Util::normalizeAssetPath("C:\\dir\\file.txt", "C:\\dir");
+    EXPECT_EQ(string("file.txt"), p.string());
+    p = Util::normalizeAssetPath("C:\\dir\\file.txt", "C:\\base");
+    EXPECT_EQ(string("../dir/file.txt"), p.string());
+
+    auto current = fs::current_path();
+    auto gen = fs::weakly_canonical((current / "..\\dir\\file.txt")).generic_string();
+    // relative asset path
+    p = Util::normalizeAssetPath("..\\dir\\file.txt", "D:");
+    EXPECT_EQ(gen, p.string());
+
+    p = Util::normalizeAssetPath(".\\file.txt", fs::current_path());
+    EXPECT_EQ(string("file.txt"), p.string());
 #else
     p = Util::normalizeAssetPath("/dir/file.txt", "/");
     EXPECT_EQ(string("dir/file.txt"), p.string());


src/core/control/jobs/AutosaveJob.cpp
```diff
@@ -30,10 +30,7 @@ void AutosaveJob::run() {
 
     Document* doc = control->getDocument();
 
-    doc->lock();
-    handler.prepareSave(doc);
     auto filepath = doc->getFilepath();
```
bhennion, line 33: I think `doc->getFilepath();` still needs the mutex lock (imagine the user renames the file -- or saves as -- while the autosave fetchs the filepath.

src/core/control/jobs/SaveJob.cpp
```diff
@@ -98,13 +98,14 @@ auto SaveJob::save() -> bool {
     Document* doc = this->control->getDocument();
     SaveHandler h;
 
-    doc->lock();
-    h.prepareSave(doc);
     fs::path filepath = doc->getFilepath();
```
bhennion, line 101: Same remark here

bhennion (changes requested):
Only one mutex related comment. The rest LGTM.

Ideally, this should be tested on various platforms, on shared partitions (e.g. in dual boot), but also on external media (usb key, external drive), and on cloud sync'd folders (e.g. Dropbox) opened on different computers.

bhennion:
@rolandlo If I remember correctly, you had worked on the attached pdf mode back in the day. Would you mind giving this a glance in case there was a hidden reason for absolute paths I'm not aware of?

rolandlo:
> @rolandlo If I remember correctly, you had worked on the attached pdf mode back in the day. Would you mind giving this a glance in case there was a hidden reason for absolute paths I'm not aware of?

I will take a look at it. There absolutely might be a hidden reason for absolute paths, but I'm not aware of it right now and I will have to test in various situations.

Technius:
I think there's one use case for absolute paths, which is moving the `.xopp` file but leaving the `.pdf` file in the same place and only using these annotations on the same computer. Otherwise, I think relative paths are preferred in all other scenarios.

This whole relative/absolute path problem can be entirely avoided with the packaged file format, but I've had almost no time to work on it in the past few months.

hyperupcall:
Cool beans, got that locking fixed!

@Technius Yeah, that's the only use case I can think of for absolute paths too. I figured it's much less intuitive and is extremely fragile. For example, if a set of files is on a USB, it could break when switching between different Linux distros.

rolandlo:
Does that really work on Windows when the PDF (or image) and the `xopp`-file are on different drives?
Seems @Febbe has some experience, see [this question](https://stackoverflow.com/questions/63710574/stdfilesystemrelative-bug-in-mingw-or-msvc)
In general since the implementation of `std::filesystem` is operating system dependent, a lot of care and tests should be made before stepping ahead with this PR.

hyperupcall:
OMG!! MSVC returns an empty string? Yeah let's hold this - it looks like we have to give special treatment to msvc and cygwin or something...

Just to clarify, we are supporting cygwin/mingw or mysys2 as a platform? I usually forget Windows, so I'm not sure if to treat those as separate platforms or lump some of them together, or which ones are most common?

hyperupcall:
Yeah - I was thinking we can have an exception for Windows - where if the drive letter is different, use the absolute path. I suppose this would mean we would have to call a helper function, and do a little dirty work in there just for Windows - does that sound okay?

With some of my other PRs merged, I think this is now something I can consider implementing

Technius:
Given that the file name / path conversions have been a source of numerous bugs over the years, it would be nice if we could also refactor the path resolution logic into helper functions, and then write unit tests for them.

One potential solution for dealing with the separate drive problem is to only save as a relative path when their [root paths](https://en.cppreference.com/w/cpp/filesystem/path/root_path) are the same.

Febbe:
> So... I'm not sure what to do with this. I guess we should treat differently the case were the .xopp file and the .pdf file are not on the same partition. I don't know if this can be easily achieved using std::filesystem.

That should be achievable via std::filesystem
Build relative path, when that fails with empty path or exception use absolute path.

@Technius it might be better to directly try to calculate a relative path. When it does not exist or is unresolvable then use the canonical path.

hyperupcall:
The latest commit addresses feedback; namely:

- If the drive letters are different on Windows, then simply use the absolute path (see test)
- If the final calculated "relative path" is an empty string, then simply use the absolute path
  - This check was added so that if there are any other path bugs with the filesystem library in the future, it will use the safe path of using the absolute path. I didn't think this was a _full_ replacement for the first `.root_path()` check though. For example, if the filesystem library somehow fixes the bug and calculates the path correctly, I wanted to ensure that the behavior stays consistent by still using absolute paths

hyperupcall:
I just fixed up the PR again, renaming the method to `normalizeAssetPath`, and changing the following:

On Windows, the path is now normalized to use UNIX-style file separators. I checked that these paths are still serializable to `fs::path` on Windows. This was done to make the data in the file format more cross-platform and it additionally partially address #3412. By making the path separators the same, simply saving the file on a different platform won't needlessly clutter the diff.

Additional tests were also added to ensuer the saved path is non-empty (which is a possibility on windows when `.root_path()` is different, or when the first argument to `std::relative` is not an absolute path.

src/util/PathUtil.cpp
```diff
@@ -270,6 +270,30 @@ auto Util::ensureFolderExists(const fs::path& p) -> fs::path {
...
+    // always be correctly constructed
+    final_path = fs::relative(asset_path, base).string();
+#endif
+
+    // In some cases (if `asset_path` is a relative path), `final_path` may be
+    // empty, so ensure that a valid path is always returned from this function
+    if (final_path.empty())
+        final_path = fs::weakly_canonical(asset_path).string();
+
+    return final_path;
```
danemtsov, line 294: [path::generic_string() and friends](https://en.cppreference.com/w/cpp/filesystem/path/generic_string) does precisely what we want here. IMO we should always use this (and its counterpart, the constructor with `format::generic_format`) when dealing with file I/O.

```suggestion
    // fs::relative() may return an empty or malformed path if the root_path()s are different
    return (asset_path.root_path() == base.root_path() ?
            fs::relative(asset_path, base) : fs::weakly_canonical(asset_path)).generic_u8string();
```

test/unit_tests/util/PathTest.cpp
```diff
@@ -117,3 +117,35 @@ TEST(UtilPath, testClearExtensions) {
...
+
+    p = Util::normalizeAssetPath("C:\\dir\\file.txt", "D:\\base");
+    EXPECT_EQ(string("C:\\dir\\file.txt"), p.string());
+
+    // do not return empty if asset_path is relative
+    p = Util::normalizeAssetPath("../dir/file.txt", "D:");
+    EXPECT_TRUE(!p.empty());
+
+    p = Util::normalizeAssetPath("../dir/file.txt", "D:\\base");
+    EXPECT_TRUE(!p.empty());
```
danemtsov, line 136: We'll want the result to always have forward slashes.

src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -251,7 +251,8 @@ void SaveHandler::visitPage(XmlNode* root, PageRef p, Document* doc, int id) {
                 }
             } else {
                 background->setAttrib("domain", "absolute");
-                background->setAttrib("filename", doc->getPdfFilepath().string());
+                auto normalizedPath = Util::normalizeAssetPath(doc->getPdfFilepath().string(), target.parent_path());
```
danemtsov, line 254:
```suggestion
                auto normalizedPath = Util::normalizeAssetPath(doc->getPdfFilepath(), target.parent_path());
```

src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -276,7 +277,10 @@ void SaveHandler::visitPage(XmlNode* root, PageRef p, Document* doc, int id) {
             p->getBackgroundImage().setCloneId(id);
         } else {
             background->setAttrib("domain", "absolute");
-            background->setAttrib("filename", p->getBackgroundImage().getFilepath().string());
+            auto normalizedPath = Util::normalizeAssetPath(
+                    std::filesystem::current_path() / p->getBackgroundImage().getFilepath().string(),
+                    target.parent_path());
```
danemtsov, line 282:
```suggestion
            auto normalizedPath = Util::normalizeAssetPath(
                    p->getBackgroundImage().getFilepath(),
                    target.parent_path());
```

danemtsov (commented):
I think it's better to check the root path even for non-Windows platforms (see [cppreference](https://en.cppreference.com/w/cpp/filesystem/path) on path syntax) - better safe than sorry.
Plus I don't think it makes sense to store a relative path if the locations are that 'far apart' (i.e. not even same root path).

https://github.com/xournalpp/xournalpp/pull/4262/files#diff-132c5a6df123d556477b8e79c5e5154c83a5075ef45bb9feadae8b46cac01428R281

Probably better to ensure all paths are absolute?

BTW we must also ensure all the strings are converted as UTF-8 to avoid locale and issues.

Febbe:
@hyperupcall this PR is nearly done and merge-able, actually only the tests must be corrected. So my question is, if you still want to continue working on it and if you have time to do it?

If you are pretty busy, we would just finish it.

hyperupcall:
If I don't push any commits by April 13th, feel free to finish it up. I wish I was able to close this earlier

Febbe:
> Probably better to ensure all paths are absolute?

A heuristic, and the ability, to manage assets afterward would be good. Neither completely absolute nor relative paths are a good final solution. 
But we can implement the heuristic and the asset manager later.

test/unit_tests/util/PathTest.cpp
```diff
@@ -10,6 +10,8 @@
  */
 
 
+#include <filesystem>
+
```
Febbe, line 14: That should be reverted, or we should choose, to abandon compilers / STL implementations which do not provide `std::filesystem`

Febbe:
@xournalpp/core before we merge this, we should discuss, if we really wan't to override all old files.
 - I think in all loaded files, asset paths should be saved the same as they are loaded before.
 - when a user relocates a xournal file via "save as" he should be asked, if the relative paths should be updated, or if he want to keep them.
 - new assets should be added with an relative path

Technius:
Let's save this for 1.3.0. I don't think we should be making such a large change right before the next release.

KlausBlum:
+1 for relative paths to background PDFs!  
  
This would solve lots or problems. It would be cool to have an option for relative or absolute paths as default setting.

Are there any reasons this PR is still waiting?   
Putting it into a nightly build would enable us to do all kinds of tests. :-)

bhennion:
Closing in favour of #6540

Issue #4123, opened 2022-06-29
[feature request] Add a toolbar option / icon to change "fill opacity"


GithubUser699:
### Problem
Instead of going to _tools → pen / textmarker → fill-opacity_ it would be nice to have an icon that I can place in the toolbar, just the same as there is an icon to turn on _fill_.

![Bildschirmfoto von 2022-06-29 17-47-05](https://user-images.githubusercontent.com/74927253/176480041-1da51496-7b31-4360-971e-7f83dc044390.png)
Here, on the right column the last item, we have an icon to turn on _fill_. Beneath it, I'd like an icon to change _fill-opacity_.

### Usecase
I want to change fill opacity (especially of the marker) quite often (I draw pictures in my notes, they help me a lot), but currently it's quite a long way to change it.

### Version information from _help → info_
Version: 1.1.1
Git commit: nogitfound
Compiled: Feb. 13 2022, 19:40:23
with libgtk: 3.24.24

Jared234:
I am currently working on adding this feature. I will create a pull request in the next few days.

PR #4263, opened 2022-09-30
Added the "Fill Opacity" feature to the available toolbar items

Jared234:
Until now, if you wanted to change the **fill opacity**, you had to go to _tools -> pen/textmarker-> fill-opacity_. But this PR adds the functionality to reach the **fill opacity** option more easily by providing a new toolbar item, as requested in #4123. 

I also created a new icon. The color scheme was fitted to the already existing **fill** icon. If you don't like the icon or its colors and have any suggestions on how to improve it, just let me now and I will fix it.

### DIFF
--- a/src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
+++ b/src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
@@ -532,7 +532,7 @@ void ToolMenuHandler::initToolItems() {
     addToolItem(toolPageLayer);
 
     addCustomItemTgl("TOOL_FILL", ACTION_TOOL_FILL, GROUP_FILL, false, "fill", _("Fill"));
-
+    addCustomItem("PEN_FILL_OPACITY", ACTION_TOOL_PEN_FILL_OPACITY, "pen-fill-opacity", _("Fill Opacity"));
 
     /*
      * Non-menu items
--- a/ui/iconsColor-dark/hicolor/scalable/actions/xopp-pen-fill-opacity.svg
+++ b/ui/iconsColor-dark/hicolor/scalable/actions/xopp-pen-fill-opacity.svg
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg width="100%" height="100%" viewBox="0 0 12 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" xmlns:serif="http://www.serif.com/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:1.5;">
+    <g transform="matrix(-1.16326,-0.196413,-0.196399,-1.23673,22.6456,22.8712)">
+        <path d="M12,14L10.338,11.998" style="fill:none;stroke:rgb(177,102,230);stroke-width:0.51px;"/>
+    </g>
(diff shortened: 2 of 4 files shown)


bhennion:
Thanks for the PR! This sounds like a useful addition.
Could you maybe post a screenshot or two? (so that people can see the icons in context and tell if they look off to them, without having to pull and build the PR)

I assume (although I did not try your PR out) clicking on the icon makes the preexisting dialog appear (the dialog you get from the menu entry), right?
We could also imagine having a drop-down slider near the toolbar, instead of a pop-up window. That'd be more work though.

Jared234:
Sure, no problem. Attached are a few screenshots in all of the different icon themes. 

Yes, you are right. At the moment the icon only opens the already existing dialog. However, I think this is the desired solution. I already thought about a drop-down slider, but I don't think that would be the best approach. If you want to select opacity, it's great to have a real-time visualization of your adjustment. IMO, if we just added a dropdown slider, many people might be forced to go to _Tools -> Pen/Text Marker-> Fill Opacity_ to get the opacity just right. 

However, I might be missing something. What are your thoughts on this? 

CustomizeToolbars
OpacitySettings
ToolbarColorDark
ToolbarColorLight
ToolbarLucideDark
ToolbarLucideLight

rolandlo (approved):
I like the icons and the functionality. The only thing I'm not sure about, whether we should also add `PEN_FILL_OPACITY` to some of the toolbars. Currently only users that learned to customize the toolbar profit of the new feature.

bhennion:
> However, I might be missing something. What are your thoughts on this?

I guess the drop-down menu could contain both a slider and a small previous box, perhaps even, if there's enough room, an editable display of the value. The challenge is to keep the drop-down box relatively small.

Now that I see the icon in context, I'm not sure it can be understood intuitively. Wouldn't some form of gauge, filled with purple with a linear gradient of transparency, and cursor on it be more intuitive? Something like this 
(The shade of purple is off and there's obviously room for improvement):
Screenshot_20221001_123605

Jared234:
@rolandlo I could definitely add this to some of the default toolbars. But I'm also not sure how useful it would be. The question is how many people need to change the opacity so often that it would make sense to have it as the default icon.

@bhennion I agree that a slider in general could be an interesting idea. For example, you could also set the pen size with a slider instead of opening the font/pen size dialog. Correct me if I'm wrong, but I don't think this feature exists for icons yet. It might be worth opening another issue for it, as I think there may still be a need for discussion. Do you agree?
As for the icon: the icon I created was influenced by the default opacity icon design. Here's a list of icons that come up when you search for "opacity":  
image
But if you want me to look for other designs, I can give it another try.
I believe that your icon might be misinterpreted as a gradient tool. For example, in Photoshop this is the gradient icon:
b6c22-gradient-tool-selection

rolandlo:
In general I would say that putting too much effort into the GUI (creating a dropdown-box with some features for a single tool) is probably not worth it, since the GUI is something we will need to rethink more globally after we have moved to Gtk 4.
About the icons I think that both ideas (current PR icons and @bhennion's suggestion) are interesting. Not sure how good that gradient would work for the Lucide icons though.

bhennion:
Well, TBH, I don't quite like the icon, and i don't feel it reflects what clicking on it does. I'm happy to be outvoted though :)

Jared234:
Should I create a poll regarding the icon designs in the _Discussions_ section?
This is my first contribution to this project and I am not sure if this feature is actively used.

GithubUser699:
@Jared234
Thank you for implementing this feature!
Also, I really like the icons you've created. I think it fits the general look the icons in xournalpp have.

bhennion:
I guess this can be merged as is. No reason to block this for an icon looks disagreement.

GithubUser699:
> I guess this can be merged as is. No reason to block this for an icon looks disagreement. Merging in 24 hours unless someone objects.

Also, https://github.com/xournalpp/xournalpp/issues/4123 can be closed after merge.

PR #4264, opened 2022-09-30
Allow the default LaTeX text when inserting to be configurable

### LINKED ISSUE #4251: Change Latex default text x^2
Hi, is there a way to set the default Latex text different from `x^2`. I have a use case where I would like to insert current date easily and clicking on the Latex button with a default `$\today` would be helpful.

hyperupcall:
Closes #4251. Modeled after the "genCmd" setting key.

Verification:
	- Manual testing


rolandlo:
Thanks for the contribution. The `xournalpp.pot` file (which is auto-regenerated after the merge) should be removed.

hyperupcall:
Fixed by latest changes :+1:

rolandlo (approved):
Thanks again for this PR. Looks good codewise and works fine.

bhennion:
Thanks a lot for the great work!

PR #4281, opened 2022-10-04
Correct minimum CMake version

hyperupcall:
The documentation for building previously specified a minimum version of `3.10`; yet, the `CMakeLists.txt` and `src/CmakeLists.txt` files specified a minimum cmake versions of `3.12` and `3.13`. This fixes the documentation to match the build files


bhennion:
Thanks for the catch!

PR #4283, opened 2022-10-04
Minimum size for shape recognizer now customizable

### LINKED ISSUE #2531: Minimum stroke distance for shape recognizer
Block shape recognizer for strokes that don't exceed some minimum of length, so that handwriting does not get messed up.

It would probably be best to have a setting to change the minimum distance in Drawing Area. A sane default could be 10mm (or two boxes on graph paper), depending on your size of handwriting.

A workaround would be to disable grid snapping and shape recognizer with snapping for recognized shapes, but I kind of need that because I can't for the life of me draw straight line without them. I also increased the grid snapping distance to 0.50 because I could not get that to work for long strokes otherwise, but that makes the whole problem even worse.

**Additional context**
I regularly use Xournal++ for math and engineering homework where I need to write fractions or do simple drawings that greatly benefit from shape recognizer, but I just cant keep it turned on currently.

hyperupcall:
When "shape recognizer" mode is enabled, and when a "pen stroke happens", the new code does the following:
  - take the minimum and maximum values for both the X and Y axis
  - if the bigger X value minus the smaller X value is less than the "minimum size" (and if this holds true for the Y values too), then it does not draw the shape

Note that some of the changes in the diff are due to `clang-format`

Closes #2531

Validation:
- Manual testing

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -244,26 +244,25 @@ void ShapeRecognizer::optimizePolygonal(const Point* pt, int nsides, int* breaks
     }
 }
 
-auto ShapeRecognizer::isStrokeLargeEnough(Stroke* stroke, double strokeMinX, double strokeMinY) -> bool {
+/**
+ * Determine if a particular stroke is large enough as to make a shape out of it
+ */
+auto ShapeRecognizer::isStrokeLargeEnough(Stroke* stroke, double strokeMinSize) -> bool {
     if (stroke->getPointCount() < 3) {
         return false;
     }
 
     auto rect = stroke->getSnappedBounds();
-    if (rect.width < strokeMinX && rect.height < strokeMinY) {
-        return false;
-    }
-
-    return true;
+    return std::hypot(rect.width, rect.height) >= strokeMinSize;
 }
 
 /**
  * The main pattern recognition function
  */
-auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinX, double strokeMinY) -> Stroke* {
+auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinSize) -> Stroke* {
     this->stroke = stroke;
 
-    if (!isStrokeLargeEnough(stroke, strokeMinX, strokeMinY)) {
+    if (!isStrokeLargeEnough(stroke, strokeMinSize)) {
         return nullptr;
     }
 
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -87,7 +87,9 @@ auto ShapeRecognizer::tryRectangle() -> Stroke* {
     auto* s = new Stroke();
     s->applyStyleFrom(this->stroke);
 
-    for (int i = 0; i <= 3; i++) { rs[i].angle = avgAngle + i * M_PI / 2; }
+    for (int i = 0; i <= 3; i++) {
+        rs[i].angle = avgAngle + i * M_PI / 2;
+    }
 
     for (int i = 0; i <= 3; i++) {
         Point p = rs[i].calcEdgeIsect(&rs[(i + 1) % 4]);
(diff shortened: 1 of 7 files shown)


ui/settings.glade
```diff
@@ -1158,7 +1170,7 @@ If the file was not yet saved you can find it in ~/.cache/xournalpp/autosaves&lt
                                             <property name="visible">True</property>
                                             <property name="can-focus">False</property>
                                             <property name="tooltip-text" translatable="yes">Gaussian parameter: the higher, the more stabilization and the more lag between your cursor and the painting tip.</property>
-                                            <property name="label" translatable="no">σ</property>
```
bhennion, line 1161: Did you change this on purpose?

src/core/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -242,13 +244,43 @@ void ShapeRecognizer::optimizePolygonal(const Point* pt, int nsides, int* breaks
...
+        if (point.x > Xmax)
+            Xmax = point.x;
+        if (point.x < Xmin)
+            Xmin = point.x;
+
+        if (point.y > Ymax)
+            Ymax = point.y;
+        if (point.y < Ymin)
+            Ymin = point.y;
+    }
```
bhennion, line 268: You can retrieve this size by calling `stroke->getSnappedBounds()` The calculation is the same (thus removing code duplication), and the result is cached for later use (in Element::snappedBounds).

bhennion (commented):
Just a couple of tiny remarks here

ui/settings.glade
hyperupcall: Oops! I must have accidentally pressed something within the Glade UI - I'll check the diff next time so this doesn't sneak in again...

src/core/control/shaperecognizer/ShapeRecognizer.cpp
hyperupcall: :+1:

hyperupcall:
Alright, fixed those two issues! I'll also add that I changed the default value from `30` to `40`, since it's larger than the height of the page lines of the default background (rather than equal to).

bhennion:
The code LGTM. I'm a bit scared the change in behaviour might surprise regular users of the stroke recognizer.
@xournalpp/core what do you think?

hyperupcall:
Just an info on the recent activity - it rebases over `master` since there were conflicts

Technius:
> I'm a bit scared the change in behaviour might surprise regular users of the stroke recognizer.

If the default value is set to approximately the size of strokes ignored by the old behavior ("size" of three points), it should be fine. Not sure what "size 40" corresponds to.

src/core/control/shaperecognizer/ShapeRecognizer.h
```diff
@@ -37,6 +37,8 @@ class ShapeRecognizer {
 
     int findPolygonal(const Point* pt, int start, int end, int nsides, int* breaks, Inertia* ss);
 
+    bool isStrokeLargeEnough(Stroke* stroke, double strokeMinSize);
```
Technius, line 40: Needs documentation. Also, this can be a static function.

src/core/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -242,13 +244,26 @@ void ShapeRecognizer::optimizePolygonal(const Point* pt, int nsides, int* breaks
...
+    if (stroke->getPointCount() < 3) {
+        return false;
+    }
+
+    auto rect = stroke->getSnappedBounds();
+    if (rect.width < strokeMinSize && rect.height < strokeMinSize) {
+        return false;
+    }
+
+    return true;
```
Technius, line 257: nit: can be simplified to
```suggestion
    return rect.width >= strokeMinSize || rect.height >= strokeMinSize;
```

Technius: It looks like this only checks the axis-aligned bounding box. If someone sets a large minimum and draws diagonal lines, will there be a noticeable or awkward difference compared to drawing horizontal or vertical lines?

hyperupcall: Hm yeah there is no difference between drawing diagonal lines and completely vertical or horizontal lines because the condition only activates when both axis are smaller than the `strokeMinSize`. I don't think it'll be unnatural unless I'm not understanding your question correctly

hyperupcall: About the nit - I'm not sure if the simplification helps readability. It continues the pattern of the previous if-statement and the `<` makes it directly obvious that the width and lengths must both be within the bounding box - the suggested changes make it a bit less clear and there is a little more thinking involved. And it might be a bit of a stretch, but it might be more error prone when adding a conditional in the future, as the conditional would have to be DeMorganized again. But if you wish, shall I still make those changes?

rolandlo: > Hm yeah there is no difference between drawing diagonal lines and completely vertical or horizontal lines because the condition only activates when both axis are smaller than the `strokeMinSize`. I don't think it'll be unnatural unless I'm not understanding your question correctly

It might be more natural to use:
```return std::hypot(rect.width, rect.height) >= strokeMinSize```
(length of the diagonal of the rectangle)

hyperupcall:
I switched to `std::hypot` as suggested (to me there doesn't seem to be a huge difference, but the hypot might be more accurate when drawing lines, compared to circles and squares). i also cleaned up the `return` :+1:

Technius (approved):
LGTM. Last point of concern might be the min size -- a size of `0` corresponds to the old default (which people did not find helpful based on the issue linked to this PR) -- but a good default is something that we'll have to determine through trial and error.

Technius:

bhennion:
Thanks for the great work!

PR #4285, opened 2022-10-04
Fix many diagnostic warnings

hyperupcall:
Hopefully, I would like to get to a place where there are no warnings outputted to the console by default (if ya'll maintainers agree?) - either by fixes or by excluding the noise by removing the warning check. That way, any warnings or mistakes are obvious and never missed.

This PR includes changes to get rid of a few tens of warnings that are low hanging fruit, that have a very low probability of changing behavior.

To make it easier for review, I'll add comments in places where the application behavior has _potentially_ changed. In all other cases, things should be proven to have identical behavior, according to clang.

Also, I'll note that when passing in `-Wno-conversion -Wno-overloaded-virtual -Wno-delete-non-abstract-non-virtual-dtor -Wno-deprecated-declarations`, then there will be _no_ warnings printed to the console  (clang 14.0.6). So, those aforementioned errors are the only ones that will be dealing with in the future (from a clean build!)

### DIFF
--- a/src/core/audio/AudioRecorder.h
+++ b/src/core/audio/AudioRecorder.h
@@ -23,7 +23,8 @@ class PortAudioProducer;
 class Settings;
 class VorbisConsumer;
 
-struct AudioRecorder final {
+class AudioRecorder final {
+public:
     explicit AudioRecorder(Settings& settings);
     AudioRecorder(AudioRecorder const&) = delete;
     AudioRecorder(AudioRecorder&&) = delete;
--- a/src/core/audio/AudioRecorder.h
+++ b/src/core/audio/AudioRecorder.h
@@ -34,12 +34,10 @@ public:
 
     bool start(fs::path const& file);
     void stop();
-    bool isRecording() const;
+    [[nodiscard]] bool isRecording() const;
     std::vector<DeviceInfo> getInputDevices() const;
 
 private:
-    Settings& settings;
-
     std::unique_ptr<AudioQueue<float>> audioQueue;
     std::unique_ptr<PortAudioProducer> portAudioProducer;
     std::unique_ptr<VorbisConsumer> vorbisConsumer;
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -68,11 +68,6 @@ auto migrateSettings() -> MigrateResult;
 void checkForErrorlog();
 void checkForEmergencySave(Control* control);
 
-auto exportPdf(const char* input, const char* output, const char* range, ExportBackgroundType exportBackground,
-               bool progressiveMode) -> int;
-auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
-               ExportBackgroundType exportBackground) -> int;
-
 void initResourcePath(GladeSearchpath* gladePath, const gchar* relativePathAndFile, bool failIfNotFound = true);
 
 void initCAndCoutLocales() {
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -454,7 +454,9 @@ void on_startup(GApplication* application, XMPtr app_data) {
             }
         }
 
-        for (auto& p: iconLoadOrder) { gtk_icon_theme_prepend_search_path(gtk_icon_theme_get_default(), p.c_str()); }
+        for (auto& p: iconLoadOrder) {
+            gtk_icon_theme_prepend_search_path(gtk_icon_theme_get_default(), p.c_str());
+        }
     }
 
     auto& globalLatexTemplatePath = app_data->control->getSettings()->latexSettings.globalTemplatePath;
@@ -526,9 +528,9 @@ auto on_handle_local_options(GApplication*, GVariantDict*, XMPtr app_data) -> gi
     initCAndCoutLocales();
 
     auto print_version = [&] {
-        if(!std::string(GIT_COMMIT_ID).empty()){
+        if (!std::string(GIT_COMMIT_ID).empty()) {
             std::cout << PROJECT_NAME << " " << PROJECT_VERSION << " (" << GIT_COMMIT_ID << ")" << std::endl;
-        } else{
+        } else {
             std::cout << PROJECT_NAME << " " << PROJECT_VERSION << std::endl;
         }
         std::cout << "└──libgtk: " << gtk_get_major_version() << "."  //
--- a/src/core/gui/inputdevices/HandRecognition.h
+++ b/src/core/gui/inputdevices/HandRecognition.h
@@ -102,11 +102,6 @@ private:
     TouchDisableInterface* touchImpl = nullptr;
 
     /**
-     * XournalView
-     */
-    GtkWidget* widget;
-
-    /**
      * InputContext
      */
     InputContext* inputContext;
--- a/src/core/gui/toolbarMenubar/ToolPageLayer.h
+++ b/src/core/gui/toolbarMenubar/ToolPageLayer.h
@@ -31,7 +31,7 @@ class ActionHandler;
 
 class ToolPageLayer: public AbstractToolItem, public LayerCtrlListener {
 public:
-    ToolPageLayer(LayerController* lc, GladeGui* gui, ActionHandler* handler, std::string id, ActionType type,
+    ToolPageLayer(LayerController* lc, ActionHandler* handler, std::string id, ActionType type,
                   IconNameHelper iconNameHelper);
     ~ToolPageLayer() override;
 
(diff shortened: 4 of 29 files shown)


src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHelper.cpp
```diff
@@ -39,7 +39,7 @@ auto gdk_context_set_icon_from_image(GdkDragContext* ctx, GtkWidget* widget) ->
         }
         case GTK_IMAGE_SURFACE: {
             cairo_surface_t* surface{};
-            g_object_get(G_OBJECT(image), "surface", &surface);
+            g_object_get(G_OBJECT(image), "surface", &surface, nullptr);
```
hyperupcall, line 42: This behavior is potentially different because previously this sentinel value wasn't added to denote the end of the variadic arguments. So I guess the behavior was previously UB? According to the [docs](https://docs.gtk.org/gobject/method.Object.get.html) it specifies to add `NULL`, but Clang recommends passing in `nullptr` and [this](https://stackoverflow.com/questions/65279075/is-it-safe-to-pass-nullptr-to-c-function) answer seems to indicate that it is OK to pass `nullptr` (just covering all bases here)

hyperupcall:
Done commenting! There should be only two instances of potential behavior deviations

src/core/audio/AudioRecorder.h
```diff
@@ -33,11 +34,11 @@ struct AudioRecorder final {
 
     bool start(fs::path const& file);
     void stop();
-    bool isRecording() const;
+    [[nodiscard]] bool isRecording() const;
     std::vector<DeviceInfo> getInputDevices() const;
 
 private:
-    Settings& settings;
+    [[ maybe_unused ]] Settings& settings;
```
bhennion, line 41: I'm not sure marking those unused member refs/pointer makes a lot of sense. Maybe make two commits: the first with the other changes, and the second one actually removing the unused member refs/pointers.

src/core/control/XournalMain.cpp
```diff
@@ -68,9 +68,9 @@ auto migrateSettings() -> MigrateResult;
 void checkForErrorlog();
 void checkForEmergencySave(Control* control);
 
-auto exportPdf(const char* input, const char* output, const char* range, ExportBackgroundType exportBackground,
+[[maybe_unused]] auto exportPdf(const char* input, const char* output, const char* range, ExportBackgroundType exportBackground,
                bool progressiveMode) -> int;
-auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
+[[maybe_unused]] auto exportImg(const char* input, const char* output, const char* range, int pngDpi, int pngWidth, int pngHeight,
```
bhennion, line 73: Same as with member pointers: here only the forward declaration is unused (because the functions are used after their actual declarations). You can remove the forward declarations.

src/core/control/tools/BaseStrokeHandler.h
```diff
@@ -47,7 +47,6 @@ class BaseStrokeHandler: public InputHandler {
 private:
     virtual void drawShape(Point& currentPoint, const PositionInputData& pos) = 0;
     DIRSET_MODIFIERS drawModifierFixed = NONE;
-    int lastCursor = -1;  // avoid same setCursor
```
bhennion, line 50: This will disappear on its own when #4137 gets merged (presumably by the end of the week).

src/core/gui/inputdevices/HandRecognition.h
```diff
@@ -104,7 +104,7 @@ class HandRecognition {
     /**
      * XournalView
      */
-    GtkWidget* widget;
+    [[maybe_unused]] GtkWidget* widget;
```
bhennion, line 107: See above

src/core/gui/toolbarMenubar/ToolPageLayer.h
```diff
@@ -76,7 +76,7 @@ class ToolPageLayer: public AbstractToolItem, public LayerCtrlListener {
 
 private:
     LayerController* lc = nullptr;
-    GladeGui* gui = nullptr;
+    [[maybe_unused]] GladeGui* gui = nullptr;
```
bhennion, line 79: See above

src/core/gui/toolbarMenubar/ToolPageSpinner.h
```diff
@@ -48,7 +48,7 @@ class ToolPageSpinner: public AbstractToolItem {
     void updateLabels();
 
 private:
-    GladeGui* gui = nullptr;
+    [[maybe_unused]] GladeGui* gui = nullptr;
```
bhennion, line 51: See above

src/util/serializing/ObjectOutputStream.cpp
```diff
@@ -60,6 +60,7 @@ void ObjectOutputStream::writeData(const void* data, int len, int width) {
     }
 }
 
+[[ maybe_unused ]]
```
bhennion, line 63: This function can be remove (in a separate commit), I think. I can't imagine it being back in our good graces in the future.

src/core/model/Snapping.cpp
```diff
@@ -12,12 +12,12 @@ namespace Snapping {
 [[nodiscard]] inline double distance(Point const& a, Point const& b) { return std::hypot(b.x - a.x, b.y - a.y); }
 double snapVertically(double y, double gridSize, double tolerance) {
     double ySnapped = roundToMultiple(y, gridSize);
-    return abs(ySnapped - y) < tolerance * gridSize / 2.0 ? ySnapped : y;
+    return std::abs(ySnapped - y) < tolerance * gridSize / 2.0 ? ySnapped : y;
```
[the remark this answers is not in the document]
bhennion: That's fine.

src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHelper.cpp
bhennion: Yes, I was probably UB and segfault prone. Good catch!

bhennion (commented):
Overall, that's a good idea. There are a couple of changes that make little sense to me:
mainly marking unused class member refs/pointers as [[ maybe_unused ]]. This only shuts the warning up, but those class members should just be removed (in a separate commit, just in case).

src/core/audio/AudioRecorder.h
hyperupcall: Will do! I wasn't sure if I should have put `[[ maybe_unused ]]`, or instead, removed the private member, and removed it from constructor's member initializer list, and also removed its parameter from the constructor. I'll make a second commit with the removed unused member refs / pointers.

hyperupcall:
Made those changes :+1: 

I re-did the changes, and separated the commits as you mentioned, along with a clang-format. The new changes left out [one](https://github.com/xournalpp/xournalpp/pull/4285#discussion_r987027794) diagnostic error, which is extraneous since it is fixed in a different PR.

src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHelper.cpp
bhennion: Actually, now that I think about it, these specific changes should go into the release branch (this could fix some segfaults people might have experienced).
Would you mind taking them out of this PR and into a new PR targetted at release-1.1?

hyperupcall:
> Actually, now that I think about it, these specific changes should go into the release branch (this could fix some segfaults people might have experienced).
Would you mind taking them out of this PR and into a new PR targetted at release-1.1?

:+1:

bhennion:
Ah, one last thing (hopefully): when removing unused (member) variables, some includes/forward declarations become useless as well. Could you remove them as well? (in the same commit as the member variables removal)
I'm sorry for the slow iterative process, I should have thought of this point earlier.

hyperupcall:
Haha no worries :stuck_out_tongue_winking_eye:  - I looked over it again and removed the only instance of this

bhennion:
Alright, can we cleanup the commit tree a little bit, and then we merge :) I'd say we should:
- squash [38ab32b](https://github.com/xournalpp/xournalpp/pull/4285/commits/38ab32bdc9bcc692c518ae66e2c468f17253fcc8) into [8b5f46f](https://github.com/xournalpp/xournalpp/pull/4285/commits/8b5f46f6c3e2d29bc3d4c41bfbd0e2c8d8c07cad)
- squash [37939ed](https://github.com/xournalpp/xournalpp/pull/4285/commits/37939edad862ba611d099aeafa34615f5fe295a0) into [2cdbd94](https://github.com/xournalpp/xournalpp/pull/4285/commits/2cdbd94619372f7def31e10e547b564a33d86d6d)

That would make 4 fairly independent and self contained commits.

hyperupcall:
Sure thing!

bhennion:
Thanks! Perfect. Merging in 24 hours unless an objection is raised.

bhennion (approved):

PR #4292, opened 2022-10-06
Bind Ctrl+Shift+Z to redo

### LINKED ISSUE #1323: Add Redo shortcut: Ctrl+Shift+Z
The action "Redo" is mapped to Ctrl+Y, which is ok, but not standard, nor used by other apps.

I would like to map Redo to also Ctrl+Shift+Z (Ctrl+Y can be kept as well)

This won't be an issue if there would be a way to map actions to any key, chosen by the user. But I can imagine that change would require way more development.

**Additional context**
I am configuring the keys of the tablet, and it is painful to change the redo key for xournal, when most other apps use ctrl+shift+z.

Thanks a lot for xournal, it is great!

### LINKED ISSUE #1418: Draw boxes around selected objects
When multiple objects are selected, only one bounding box is drawn around the objects. This makes it difficult to determine which objects are selected if there are multiple overlapping objects around the selected region.

In addition to the larger bounding box, each selected object should have a bounding box drawn around it.

(contrived example)
image

The look of the boxes can be changed, e.g. dashed lines or transparency.

We can also fade out all other objects instead, but I'm not sure how useful that would be.

**Additional context**
N/A

hyperupcall:
Closes #1323

### DIFF
--- a/ui/main.glade
+++ b/ui/main.glade
@@ -367,6 +367,7 @@
                             <property name="use-stock">True</property>
                             <signal name="activate" handler="ACTION_REDO" swapped="no"/>
                             <accelerator key="y" signal="activate" modifiers="GDK_CONTROL_MASK"/>
+                            <accelerator key="z" signal="activate" modifiers="GDK_CONTROL_MASK|GDK_SHIFT_MASK"/>
                           </object>
                         </child>
                         <child>


bhennion (approved):
LGTM. Merging in 24 hours unless an objection is raised.

PR #4299, opened 2022-10-09
Fix PDF toolbox stroke actions when PDF page num is not current page

### LINKED ISSUE #3971: Can't use the new "Select PDF text" tool after I added a page
**Affects versions :**
 - OS: ArchLinux GNOME 42 Wayland, but also on Ubuntu 20.04 GNOME 3.36.8 Xorg, so I don't think it's a distro fault
 - Which version 
     - Arch: xournalpp 1.1.1+dev ibgtk: 3.24.33
     - Ubuntu: xournalpp 1.1.1+dev ibgtk: 3.24.20
 - Installation method: [e.g. PPA, Flatpak, official release, etc.]
     - Arch: AUR git version
     - Ubuntu: don't remember

**To Reproduce**
1. Open a PDF
2. Insert a page
3. Select text with the new "Select PDF text" tool, and try to click on one of the options. No option actually do anything. In the pages BEFORE the inserted page, it works. In the pages AFTER the inserted page, does not work.

**Expected behavior**
Should work as expected.

**Additional context**

Technius:
Fixes #3971.

Previously, `PdfFloatingToolbox::createStrokes` would incorrectly no-op if the page number (of the document) is not equal to the page number of the background PDF that the text selection is over.

This commit changes the fix to instead retrieve the PDF page number of the current document page, and compare that against the PDF page number of the text selection.

### DIFF
--- a/src/core/gui/PdfFloatingToolbox.cpp
+++ b/src/core/gui/PdfFloatingToolbox.cpp
@@ -13,6 +13,7 @@
 #include "control/ToolHandler.h"    // for ToolHandler
 #include "gui/PageView.h"           // for XojPageView
 #include "gui/XournalView.h"        // for XournalView
+#include "model/Document.h"         // for Document
 #include "model/Layer.h"            // for Layer
 #include "model/PageRef.h"          // for PageRef
 #include "model/Point.h"            // for Point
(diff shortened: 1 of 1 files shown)


bhennion (approved):

Technius:

PR #4304, opened 2022-10-10
Split tools/Selection into view/controller

bhennion:
This PR split the selection tools (rectangle selection or region selection) into view+controller.

It also:
- simplifies and optimizes some of the code in tools/Selection
- optimizes the size of the areas repainted (upon input events)

(Note that tools/Selection only handles the selection process, tools/EditSelection is the tool handling the selection moving, resizing, copying and so on)

This is ready for a review for whoever feels like it.

### DIFF
--- a/src/core/view/overlays/SelectionView.cpp
+++ b/src/core/view/overlays/SelectionView.cpp
@@ -1,0 +1,56 @@
+#include "SelectionView.h"
+
+#include <vector>
+
+#include "control/tools/Selection.h"
+#include "util/Color.h"
+#include "util/LoopUtil.h"
+#include "util/Range.h"
+#include "util/raii/CairoWrappers.h"
+#include "view/Repaintable.h"
+
+using namespace xoj::view;
+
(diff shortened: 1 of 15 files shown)


src/core/view/overlays/SelectionView.cpp
```diff
@@ -0,0 +1,55 @@
...
+        return;
+    }
+
+    xoj::util::CairoSaveGuard saveGuard(cr);  // cairo_save
+
+    // set the line always the same size on display
+    cairo_set_line_width(cr, BORDER_WIDTH_IN_PIXELS / this->parent->getZoom());
+
+    Util::cairo_set_source_rgbi(cr, selectionColor);
+
```
Technius, line 34: nit: maybe add a `cairo_new_path();` to make it clear that we're starting a new path.

Technius (approved):
Thanks. Didn't see any obvious problems.

bhennion:
Thanks a lot @Technius  for the review. I applied your recommendations.

PR #4305, opened 2022-10-10
Split control/SearchControl into view/controller

bhennion:
This PR splits control/SearchControl into a view and a controller.

Some small side-refactoring:
- searched strings are now passed by references (instead of by a succession of copies and moves)

Ready for a review.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1749,8 +1749,9 @@ auto Control::getCurrentPageNo() -> size_t {
     return 0;
 }
 
-auto Control::searchTextOnPage(string text, int p, int* occures, double* top) -> bool {
-    return getWindow()->getXournal()->searchTextOnPage(std::move(text), p, occures, top);
+auto Control::searchTextOnPage(const std::string& text, size_t pageNumber, size_t* occurrences,
+                               double* yOfUpperMostMatch) -> bool {
+    return getWindow()->getXournal()->searchTextOnPage(text, pageNumber, occurrences, yOfUpperMostMatch);
 }
 
 auto Control::getCurrentPage() -> PageRef {
--- a/src/core/control/Control.h
+++ b/src/core/control/Control.h
@@ -187,7 +187,14 @@ public:
 
     bool isFullscreen();
 
-    bool searchTextOnPage(std::string text, int p, int* occures, double* top);
+    /**
+     * @brief Search text on the given page. The matches (if any) are stored in the XojPageView::SearchControl instance.
+     * @param occurrences If not nullptr, the pointed variable will contain the number of matches on the page
+     * @param yOfUpperMostMatch If not nullptr, will contain the y coordinate of the first match on the page
+     *                          (Used for scrolling to the first match)
+     * @return true if at least one match was found
+     */
+    bool searchTextOnPage(const std::string& text, size_t pageNumber, size_t* occurrences, double* yOfUpperMostMatch);
 
     /**
      * Fire page selected, but first check if the page Number is valid
--- a/src/core/control/SearchControl.cpp
+++ b/src/core/control/SearchControl.cpp
@@ -9,38 +9,25 @@
 #include "model/Text.h"     // for Text
 #include "model/XojPage.h"  // for XojPage
 #include "view/TextView.h"  // for TextView
+#include "view/overlays/SearchResultView.h"
 
-using std::string;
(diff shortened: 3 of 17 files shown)


src/core/control/Control.cpp
```diff
@@ -1749,8 +1749,9 @@ auto Control::getCurrentPageNo() -> size_t {
     return 0;
 }
 
-auto Control::searchTextOnPage(string text, int p, int* occures, double* top) -> bool {
-    return getWindow()->getXournal()->searchTextOnPage(std::move(text), p, occures, top);
+auto Control::searchTextOnPage(const string& text, size_t pageNumber, size_t* occurences, double* yOfUpperMostMatch)
```
Technius, line 1752:
```suggestion
auto Control::searchTextOnPage(const std::string& text, size_t pageNumber, size_t* occurences, double* yOfUpperMostMatch)
```

src/core/control/Control.h
```diff
@@ -187,7 +187,7 @@ class Control:
 
     bool isFullscreen();
 
-    bool searchTextOnPage(std::string text, int p, int* occures, double* top);
+    bool searchTextOnPage(const std::string& text, size_t pageNumber, size_t* occurences, double* yOfUpperMostMatch);
```
Technius, line 190: typo?
```suggestion
    bool searchTextOnPage(const std::string& text, size_t pageNumber, size_t* occurrences, double* yOfUpperMostMatch);
```

Technius: It would also be helpful to document this method.

src/core/gui/SearchBar.cpp
```diff
@@ -99,88 +99,86 @@ void SearchBar::searchTextChangedCallback(GtkEntry* entry, SearchBar* searchBar)
...
+    size_t currentPage = control->getCurrentPageNo();
+    size_t count = control->getDocument()->getPageCount();
     if (count < 2) {
         // Nothing to do
         return;
     }
 
     MainWindow* win = control->getWindow();
-    int x = page - 1;
+    size_t searchedPage = currentPage == 0 ? count - 1 : currentPage - 1;
```
Technius, line 155: nit: this line should be moved closer to the loop.

src/core/gui/SearchBar.cpp
```diff
@@ -99,88 +99,86 @@ void SearchBar::searchTextChangedCallback(GtkEntry* entry, SearchBar* searchBar)
...
-        x = count - 1;
-    }
-
-    double top = 0;
-    int occures = 0;
+    double yOfUpperMostMatch = 0;
+    size_t occurences = 0;
 
-    while (x != page) {
+    while (searchedPage != currentPage) {
```
Technius, line 166:
```suggestion
    // Search backwards through the pages, wrapping around if needed.
    while (searchedPage != currentPage) {
```

Technius: Could also simplify into a for loop, but not sure if that would be more readable.

Technius (approved):
LGTM. Minor nitpicks below:

src/core/gui/SearchBar.cpp
bhennion: I refactored SearchPage::searchNext and searchPrevious to remove the code duplication, and I documented the functions.
I chose not the change the behaviour,: the current page is not searched.
Those functions are called when pressing the next/previous button on the search bar. This then finds the next/previous page with matches and scrolls to said page.
I think the behaviour is a little weird (it should first navigate through the matches of the current page), but changing this is out of scope.

src/core/control/SearchControl.cpp
```diff
@@ -9,39 +9,26 @@
...
-        cairo_stroke_preserve(cr);
-        auto applied = GdkRGBA{color.red, color.green, color.blue, 0.3};
-        gdk_cairo_set_source_rgba(cr, &applied);
-        cairo_fill(cr);
-    }
-}
-
-auto SearchControl::search(std::string text, int* occures, double* top) -> bool {
-    freeSearchResults();
+SearchControl::~SearchControl() { this->results.clear(); }
```
Technius, line 19: nit: won't `results` will be cleared automatically when it's destroyed?
```suggestion
SearchControl::~SearchControl() = default;
```

Technius (approved):

bhennion:
Thanks for the review @Technius!
I cleaned up the commit tree.

PR #4306, opened 2022-10-10
Split VerticalToolHandler into controller and view

bhennion:
This PR splits tools/VerticalToolHandler into controller and view.

- reduces the buffer size
- reduces the repainted/rerendered areas

This is not battle ready yet: for performance reasons (also necessary in order to potentially allow for infinite pages at some point), the size of the buffer had to be reduced (it used to be half of the entire page).
The new (and imperfect) buffer is the size of the visible part of the page, but it starts just below (or ends just above, if ctrl is pressed) the event starting the VerticalTool. This works perfectly as long as the user ends their vertical shift whthin the current page.
In particular, in the following two situations, the buffer is to small:

1. If the view is in between pages: starting the vertical tool on the bottom page and dragging it all above the page border will (can) show that the mask is to small at the bottom.
2. After starting the vertical tool whereever on a page, move your mouse (without releasing the button) to the top of the tool bar. You''ll see the mask is to short at the bottom as well.

The solution is probably to have the mask exactly the height of the window area (or something like that). I need to think on it. If anyone has an idea on the matter, please share!

### DIFF
--- a/src/core/control/tools/VerticalToolHandler.h
+++ b/src/core/control/tools/VerticalToolHandler.h
@@ -80,12 +80,12 @@ public:
         Below = 1,
     };
 
-    double getStartY() const { return startY; }
-    double getEndY() const { return endY; }
-    Side getSide() const { return spacingSide; }
+    inline double getStartY() const { return startY; }
+    inline double getEndY() const { return endY; }
+    inline Side getSide() const { return spacingSide; }
     double getPageWidth() const;
 
-    auto getViewPool() const -> std::shared_ptr<xoj::util::DispatchPool<xoj::view::VerticalToolView>> {
+    inline auto getViewPool() const -> std::shared_ptr<xoj::util::DispatchPool<xoj::view::VerticalToolView>> {
         return viewPool;
     }
 
@@ -107,6 +107,10 @@ private:
     PageRef page;
     Layer* layer;
     std::vector<Element*> elements;
+    /**
+     * @brief Stores the smallest box containing all the adopted elements. 
+     *     Used to only refresh the part of the screen that needs refreshing.
+     */
     Range ownedElementsOriginalBoundingBox;
 
     double startY;
--- a/src/core/control/tools/VerticalToolHandler.h
+++ b/src/core/control/tools/VerticalToolHandler.h
@@ -18,34 +18,40 @@
 #include <cairo.h>    // for cairo_surface_t, cairo_t
 #include <gdk/gdk.h>  // for GdkEventKey, GdkWindow
 
-#include "control/zoom/ZoomListener.h"  // for ZoomListener
-#include "model/ElementContainer.h"     // for ElementContainer
-#include "model/PageRef.h"              // for PageRef
+#include "model/ElementContainer.h"  // for ElementContainer
+#include "model/OverlayBase.h"
+#include "model/PageRef.h"  // for PageRef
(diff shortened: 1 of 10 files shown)


Technius:
@bhennion I can take a look at this in a few days if you rebase.

bhennion:
> @bhennion I can take a look at this in a few days if you rebase.

Done.

This buffer size issue is still pending. The ideal solution would be to know "how far can the input sequence take us" (e.g. the range of the input device -- for a mouse, the screen(s), for a tablet, the tablet, and so on). I don't know how to get that kind of info.

src/core/control/tools/VerticalToolHandler.h
```diff
@@ -64,16 +70,26 @@ class VerticalToolHandler: public ElementContainer, public ZoomListener {
...
     enum class Side {
         /** elements above the reference line */
         Above = -1,
         /** elements below the reference line */
         Below = 1,
     };
 
+    double getStartY() const { return startY; }
+    double getEndY() const { return endY; }
+    Side getSide() const { return spacingSide; }
```
Technius, line 85: nit:
```suggestion
    inline double getStartY() const { return startY; }
    inline double getEndY() const { return endY; }
    inline Side getSide() const { return spacingSide; }
```

src/core/control/tools/VerticalToolHandler.h
```diff
@@ -64,16 +70,26 @@ class VerticalToolHandler: public ElementContainer, public ZoomListener {
...
+    double getPageWidth() const;
+
+    auto getViewPool() const -> std::shared_ptr<xoj::util::DispatchPool<xoj::view::VerticalToolView>> {
+        return viewPool;
+    }
```
Technius, line 90: nit:
```suggestion
    inline auto getViewPool() const -> std::shared_ptr<xoj::util::DispatchPool<xoj::view::VerticalToolView>> {
        return viewPool;
    }
```

src/core/control/tools/VerticalToolHandler.h
```diff
@@ -82,33 +98,16 @@ class VerticalToolHandler: public ElementContainer, public ZoomListener {
...
     Layer* layer;
     std::vector<Element*> elements;
-
-    /**
-     * Image buffer containing a rendering of the elements being spaced. This
-     * buffer is rendered below the endY if direction is below, or above the
-     * endY if the direction is above.
-     */
-    cairo_surface_t* crBuffer = nullptr;
+    Range ownedElementsOriginalBoundingBox;
```
Technius, line 110: Needs a doc comment.

Technius (approved):
The overall changes look sensible to me. The zoom functionality works as it did before based on my local testing.

src/core/control/tools/VerticalToolHandler.h
bhennion:
```suggestion
    /**
     * @brief Stores the smallest box containing all the adopted elements. 
     *     Used to only refresh the part of the screen that needs refreshing.
     */
    Range ownedElementsOriginalBoundingBox;
```

bhennion:
Thanks for the review @Technius!
I applied your recommendations.

Should we be merging this then?

Technius:
Yeah. Build failure looks like a temporary hiccup.

bhennion:
(Squashing and) Merging in 24 hours unless an objection is raised.

Issue #4295, opened 2022-10-06
Technical Documentation Improvements?


hyperupcall:
**Is your feature request related to a problem? Please describe.**

I was recently attempting to build and run the tests, and that experience led me to a few thoughts that could improve our documentation:

1. There seems to be a lot of duplication between the `{Linux,Windows,Mac}Build.md` files, such as the clone / building. This makes it a bit harder to manage it, and less intuitive (for example, if building on Windows, one would have to look into `LinuxBuild.md` to figure out how to run tests)
    - A solution may be to have a `Build.md` file, and link there with common steps from each of the platform-specific files
    
2. I think shared titles could make navigation a bit easier. For example, instead of "Compiling" and "Build xournal++" steps, make them both "Compiling" or something similar.

3. Maybe move the "Branching Strategy" section from `Releases.md` to `CONTRIBUTING.md`? I didn't know abuot the branching procedures until looking at `Releases.md`, which most people probably don't read

4. Styling is also a bit inconsistent - maybe it would be good to add a markdown linter like `markdownlint`, and add that to CI so everything is more readable at the source level?

Thoughts on this? For me, documentation is one of the most important aspects of a project, and doing this can make it much easier for people to contribute (myself included :laughing:). If others are onboard I'd definitely be willing to create a PR to address these things :)

Technius:
I changed the title of this issue to clarify that it is about the technical documentation. The [documentation site](https://github.com/xournalpp/xournalpp.github.io) is another problem...

> There seems to be a lot of duplication between the `{Linux,Windows,Mac}Build.md` files, such as the clone / building

I agree that there is duplication that can be refactored out, but the focus is on setting up the environment and creating the distribution files, which is the hardest part. The Linux setup is _very_ easy compared to the others but depends on distro; Windows requires MSYS and a couple of hacky setup scripts; and Mac is so complicated to get working that I can't even reproduce the build on my own machine. Furthermore, the packaging steps are different for each platform/redistributable.

That being said, it would be helpful to pull out information like how to run CMake, what settings to provide to CMake, how to run tests, etc.; and put them into a file such as `Compile.md` or something. Not sure where we would put platform-specific settings like disabling X11, however.

> Maybe move the "Branching Strategy" section from `Releases.md` to `CONTRIBUTING.md`?

The maintenance team doesn't expect contributors to understand the branching strategy, because we will take of the details. But I agree in that we should at least incorporate a link to the `Releases.md` document.

[...]

PR #4307, opened 2022-10-10
Add EditorConfig file

hyperupcall:
I [mentioned](https://github.com/xournalpp/xournalpp/issues/4295#issuecomment-1272725392) in #4295 that an [EditorConfig](https://editorconfig.org) file would make it much easier to onboard new contributors because editor settings such as indentation, spacing, and trailing newline are automatically configured and just work. Additionally, it prevents bugs relating to ending files with a newline on POSIX since `insert_final_newline = true`. Thought it would be good to create a PR for this, as it's a pretty simple change

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+indent_size = 2
+trim_trailing_whitespace = false


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
```
Technius, line 5: Will this conflict with `.clang-format`?

hyperupcall: No because the `TabWidth=4` and `UseTab=Never` in `.clang-format`. Also EditorConfig only affects editor settings and configuration - it does not automatically reformat files. For example, in `ui/xournalpp.css`, it will not change the tabs to spaces - only if you manually de-tab and re-tab those individual lines.

Technius (approved):
LGTM. Merging in 24 hours unless there are objections.

Issue #3983, opened 2022-04-26
Text fields immediately close when created with middle mouse button


AstusRush:
**Affects versions :**
 - OS: Linux (_confirmed under_ Kubuntu 22.04 _and_ Tuxedo_OS (Ubuntu) 20.04 _(both using Xournal++ v1.1.1)_)
 - Desktop environment: KDE Plasma Desktop
 - Graphics Platform: X11
 - libgtk version: 3.24.29
 - Version of Xournal++: 1.1.1 (Git commit 4f113e94)
 - Installation method: Discover Software Centre (Source: ubuntu-jammy-universe (Ubuntu 22.04 LTS))

**Describe the bug**
When I select the text tool for the middle mouse button I can only type for as long as I hold down the button (except if I also select the text button for the left mouse button).

**To Reproduce**
1. Select the Pen tool for the left mouse button
2. Select the Text tool for the middle mouse button
3. Click inside the document with the middle mouse button

**Expected behavior**
The expected behaviour is that the text field stays open until I explicitly exit it e.g. by pressing Esc.

**Additional context**
This behaviour is caused by the tool always switching back to the one of the left mouse button after releasing the right or middle mouse button. The problem is, that this should not close the active text field. Xournal (not ++) handles this case correctly.

Jared234:
I am currently working on a fix for this bug. I will create a pull request in the next few days.

Technius:
From #4309:
> @Jared234: This is an attempt to fix the bug mentioned in #3983. After testing and debugging I realized that this bug not only appears if you use the middle mouse button, but also if you choose the right mouse button to open a text field.appears if you use the middle mouse button, but also if you choose the right mouse button to open a text field.

I can confirm that stylus side buttons also have the same issue. I believe this bug occurs whenever the text tool is assigned to any "non-primary" input button.

rolandlo:
Fixed via #4309.

PR #4309, opened 2022-10-11
Fixed bug that closed text fields immediately after creating them with the middle mouse button. 

Jared234:
This is an attempt to fix the bug mentioned in #3983. After testing and debugging I realized that this bug not only appears if you use the middle mouse button, but also if you choose the right mouse button to open a text field.
When this function is used for other tools, releasing the pressed key should return to the previous tool (e.g. the eraser should disappear after releasing the middle/right mouse button). This is not the case with the text tool, since you not only want to create a text box, but also write something down right away.
Therefore, I decided to actually (permanently) select the text tool if you press the middle mouse button. If you think there is a better fix, jut let me know and I can have another look at it.

### DIFF
--- a/src/gui/inputdevices/MouseInputHandler.cpp
+++ b/src/gui/inputdevices/MouseInputHandler.cpp
@@ -124,11 +124,11 @@ auto MouseInputHandler::changeTool(InputEvent const& event) -> bool {
 
     if (toolChanged) {
         ToolType toolType = toolHandler->getToolType();
-        if(toolType == TOOL_TEXT) 
+        if (toolType == TOOL_TEXT)
             toolHandler->selectTool(toolType);
         toolHandler->fireToolChanged();
     }
-    
+
     return true;
 }
 
--- a/src/gui/inputdevices/MouseInputHandler.cpp
+++ b/src/gui/inputdevices/MouseInputHandler.cpp
@@ -122,8 +122,13 @@ auto MouseInputHandler::changeTool(InputEvent const& event) -> bool {
     else
         toolChanged = toolHandler->pointActiveToolToToolbarTool();
 
-    if (toolChanged)
+    if (toolChanged) {
+        ToolType toolType = toolHandler->getToolType();
+        if(toolType == TOOL_TEXT) 
+            toolHandler->selectTool(toolType);
         toolHandler->fireToolChanged();
+    }
+    
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


Jared234:
@Technius Do you have any updates on whether this pull request can be merged into Master?

Technius (commented):
* Can you rebase onto `release-1.1` branch and change the merge target? This will help us get the bug fix into the next release, and we will also be able to merge the bugfix back into master eventually.
* I think a similar fix can be applied to the stylus tool as well (`StylusInputHandler.cpp`).

Jared234:
Thanks for the review. I've added a fix for the stylus and rebased to the release 1.1 branch. I'm relatively new to Github, so if my rebase is wrong, just let me know and I'll try to fix it.

Technius:
I tested this locally and it works well.

I don't think you rebased correctly. What you did was merge `release-1.1` back into `master`, but we want to push the commits on to branch `release-1.1`.
1. First, you should drop the commits that are not your original commits using `git rebase -i master`. [See this page for a graphical explanation of rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing). Or an easier alternative is to copy the changes you made, run `git reset --hard master`, paste your changes, and then commit your changes again.
2. Run `git rebase release-1.1` to rebase on to `release-1.1` branch. If you run `git log`, you should then see that one of the commits in the log will be marked as `release-1.1` -- this means the rebase is correct.
3. Click edit next to the PR title: 
image

4. Choose `release-1.1` from the dropdown: 
image

If you did this correctly, you should see only the commits you added for this PR -- and not a bunch of random commits from master.

Technius:
If you don't have time to rebase, I can handle the rebasing for you as well.

rolandlo:
> 2\. Run `git rebase release-1.1` to rebase on to `release-1.1` branch. If you run `git log`, you should then see that one of the commits in the log will be marked as `release-1.1` -- this means the rebase is correct.

Actually this works better than `git rebase release-1.1`:
```term
git rebase --onto release-1.1 master BRANCHNAME
```
where `BRANCHNAME` is the name of the local branch, `3983_bug_text_tool_closing` in this case.
With `git rebase release-1.1` you will have to fix a lot of merge conflicts in my experience.

Jared234:
I hope it works now. Thanks for the detailed instructions. 
I really appreciate your patience and your help.

Technius:
Thanks for the PR!

PR #4311, opened 2022-10-12
Automatically create empty last page if last page is drawn on

### LINKED ISSUE #1343: Please add an option for automatically appending empty pages after last page hase been reached
I frequently use Xournalpp not for annotating PDF files which have a fixed number of pages but for just taking notes. Then, when I reach the bottom of the page, I need to click a button to add another empty page to continue writing.

It would be great to have an option to append an empty page automatically when the end of the file (bottom of the last page) is reached. Maybe make it dependent from the context (off if a PDF file is loaded, on if not).

hyperupcall:
Closes #1343

I used `doc->isAttachPdf()` to determine if editing a "pdf" file (so empty last pages aren't created in PDF files), but I'm not sure if that is valid in all cases?

I also added the setting in the third section - is there a better place to put it or a better name?

image

### DIFF
--- a/src/core/control/PageBackgroundChangeController.h
+++ b/src/core/control/PageBackgroundChangeController.h
@@ -38,7 +38,7 @@ public:
     virtual void changeCurrentPageBackground(PageType& pageType);
     void changeCurrentPageBackground(PageTypeInfo* info) override;
     void changeAllPagesBackground(const PageType& pt);
-    void insertNewPage(size_t position);
+    void insertNewPage(size_t position, bool shouldScrollToPage = true);
     GtkWidget* getMenu();
 
     // DocumentListener
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -136,7 +136,7 @@ void StrokeHandler::paintTo(const Point& point) {
                 stroke->setLastPressure(std::max(endPoint.z, point.z) * stroke->getWidth());
             } else {
                 if (const double widthDelta = (point.z - endPoint.z) * stroke->getWidth();
-                    - widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
+                    -widthDelta > MAX_WIDTH_VARIATION || widthDelta > MAX_WIDTH_VARIATION) {
                     /**
                      * If the width variation is to big, decompose into shorter segments.
                      * Those segments can not be shorter than PIXEL_MOTION_THRESHOLD
@@ -272,9 +272,12 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
 
     if (settings->getEmptyLastPageAppend() == EmptyLastPageAppendType::OnDrawOfLastPage) {
         auto* doc = control->getDocument();
-        if (doc->getPdfPageCount() == 0) {
-            doc->lock();
+        doc->lock();
+        auto pdfPageCount = doc->getPdfPageCount();
+        doc->unlock();
+        if (pdfPageCount == 0) {
             auto currentPage = control->getCurrentPageNo();
+            doc->lock();
             auto lastPage = doc->getPageCount() - 1;
             doc->unlock();
             if (currentPage == lastPage) {
--- a/src/core/control/tools/StrokeHandler.cpp
+++ b/src/core/control/tools/StrokeHandler.cpp
@@ -15,6 +15,7 @@
 #include "control/ToolHandler.h"                      // for ToolHandler
 #include "control/layer/LayerController.h"            // for LayerController
 #include "control/settings/Settings.h"                // for Settings
+#include "control/settings/SettingsEnums.h"           // for EmptyLastPageAppendType
 #include "control/shaperecognizer/ShapeRecognizer.h"  // for ShapeRecognizer
 #include "control/tools/InputHandler.h"               // for InputHandler::P...
 #include "control/tools/SnapToGridInputHandler.h"     // for SnapToGridInput...
@@ -23,6 +24,7 @@
 #include "gui/PageView.h"                        // for XojPageView
 #include "gui/XournalView.h"                     // for XournalView
 #include "gui/inputdevices/PositionInputData.h"  // for PositionInputData
+#include "model/Document.h"                      // for Document
 #include "model/Layer.h"                         // for Layer
 #include "model/LineStyle.h"                     // for LineStyle
 #include "model/Stroke.h"                        // for Stroke, STROKE_...
@@ -266,11 +268,22 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
     Layer* layer = page->getSelectedLayer();
 
     UndoRedoHandler* undo = control->getUndoRedoHandler();
-
     undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke.get()));
 
-    ToolHandler* h = control->getToolHandler();
+    if (settings->getEmptyLastPageAppend() == EmptyLastPageAppendType::OnDrawOfLastPage) {
+        auto* doc = control->getDocument();
+        if (doc->getPdfPageCount() == 0) {
+            doc->lock();
+            auto currentPage = control->getCurrentPageNo();
+            auto lastPage = doc->getPageCount() - 1;
+            doc->unlock();
+            if (currentPage == lastPage) {
+                control->insertNewPage(currentPage + 1, false);
+            }
+        }
+    }
 
+    ToolHandler* h = control->getToolHandler();
     if (h->getDrawingType() == DRAWING_TYPE_STROKE_RECOGNIZER) {
         ShapeRecognizer reco;
 
--- a/ui/settings.glade
+++ b/ui/settings.glade
@@ -4450,18 +4450,18 @@ This setting can make it easier to draw with touch. </property>
                               </packing>
                             </child>
                             <child>
-                              <object class="GtkFrame">
+                              <object class="GtkFrame" id="sid198">
                                 <property name="visible">True</property>
                                 <property name="can-focus">False</property>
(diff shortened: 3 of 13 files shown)


src/core/control/tools/StrokeHandler.cpp
```diff
@@ -266,9 +267,20 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
     Layer* layer = page->getSelectedLayer();
 
     UndoRedoHandler* undo = control->getUndoRedoHandler();
-
     undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, stroke.get()));
 
+    // If stroke was on last page, ensure there is always an empty last page by creating one (if configured)
+    auto* doc = control->getDocument();
+    if (!doc->isAttachPdf() && settings->getAddEmptyLastPage()) {
```
rolandlo, line 274: `isAttachPdf()` only returns `true` if the PDF is attached to the document (which is usually not the case when annotating a PDF). It's better to check if there are pdf pages:
```suggestion
    if (doc->getPdfPageCount() == 0 && settings->getAddEmptyLastPage()) {
```

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -266,9 +267,20 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
...
+        doc->lock();
+        auto currentPage = control->getCurrentPageNo();
+        auto totalPages = doc->getPageCount() - 1;
+        doc->unlock();
+        if (currentPage == totalPages) {
```
rolandlo, line 279: The naming is a bit misleading. 
```suggestion
        auto lastPage = doc->getPageCount() - 1;
        doc->unlock();
        if (currentPage == lastPage) {
```

src/core/control/PageBackgroundChangeController.h
```diff
@@ -38,7 +38,7 @@ class PageBackgroundChangeController:
     virtual void changeCurrentPageBackground(PageType& pageType);
     void changeCurrentPageBackground(PageTypeInfo* info) override;
     void changeAllPagesBackground(const PageType& pt);
-    void insertNewPage(size_t position);
+    void insertNewPage(size_t position, bool shouldScrollToPage = false);
```
rolandlo, line 41: It looks inconsistent to set the different default values here and in `Control.h`.
```suggestion
    void insertNewPage(size_t position, bool shouldScrollToPage = true);
```

ui/settings.glade
```diff
@@ -4449,6 +4443,41 @@ This setting can make it easier to draw with touch. </property>
...
+                                <property name="can-focus">False</property>
+                                <property name="label-xalign">0</property>
+                                <child>
+                                  <object class="GtkAlignment">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="left-padding">12</property>
+                                    <child>
+                                      <object class="GtkCheckButton" id="cbAddEmptyLastPage">
+                                        <property name="label" translatable="yes">Automatically Add Empty Last Page</property>
```
rolandlo, line 4458:
```suggestion
                                        <property name="label" translatable="yes">Automatically add empty last page when not annotating a PDF</property>
```

ui/settings.glade
```diff
@@ -4449,6 +4443,41 @@ This setting can make it easier to draw with touch. </property>
...
+                                        <property name="draw-indicator">True</property>
+                                      </object>
+                                    </child>
+                                  </object>
+                                </child>
+                                <child type="label">
+                                  <object class="GtkLabel">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="label" translatable="yes">Automatically add empty last page</property>
```
rolandlo, line 4471:
```suggestion
                                    <property name="label" translatable="yes">Automatically Add Empty Last Page</property>
```

rolandlo (changes requested):
Thanks for your contribution. Mostly the feature works fine. 

The `doc->isAttachPdf()` is the wrong property though, as PDF's are usually not "attached" to the document, even if they get annotated. It's better to use the PDF page count.

I don't see a better place of where to put this setting in the preference.

mjg:
Could we make this so that the extra page is added when one tries to scroll beyond the last page, or that it is discarded on save if it is still empty?
So far, a user of this feature would always have to delete an extra empty page.

hyperupcall:
I think your suggestion opens up the possibility that users expect to do different things to create the last page. Maybe we can have mutually exclusive options so the user can pick their preferred method?

Atomatic Create Empty Last Page:
- Never
- When writing on last page, automatically create new one
- When writing on last page, automatically create new one (but remove it on save if it's empty)
- When scrolling to bottom of page, if it is the last page

After I apply the suggestions by @rolandlo, what are your thoughts on this @mjg? In this PR, I can add the first two options. Mayybe the first three options, but I don't think that mutating the current state of the document on save is a good idea in any case (even if it a user controlled-preferrence) - i would need to hear what other people think first. and if someone wants a new behavior in the future, they can just add an option).

hyperupcall:
Epic - so there are now three options, with `disabled` being the default

image

I implemented all the changes above in the new revision

ui/settings.glade
```diff
@@ -4449,6 +4449,104 @@ This setting can make it easier to draw with touch. </property>
...
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="label-xalign">0</property>
+                                <child>
+                                  <object class="GtkAlignment">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="left-padding">12</property>
+                                    <child>
+                                      <!-- n-columns=3 n-rows=3 -->
```
rolandlo, line 4463: Any reason why the grid has 3 columns opposed to just 1 column (in which case it could also be a GtkBox)?

ui/settings.glade
```diff
@@ -4449,6 +4449,104 @@ This setting can make it easier to draw with touch. </property>
...
+                                          <object class="GtkRadioButton" id="rdLastPageAppendOnDrawOfLastPage">
+                                            <property name="label" translatable="yes">when drawing on last page</property>
+                                            <property name="visible">True</property>
+                                            <property name="can-focus">True</property>
+                                            <property name="receives-default">False</property>
+                                            <property name="active">True</property>
+                                            <property name="draw-indicator">True</property>
+                                            <property name="group">rdLastPageAppendDisabled</property>
+                                            <signal name="group-changed" handler="seta" swapped="no"/>
+                                            <signal name="toggled" handler="setb" swapped="no"/>
```
rolandlo, line 4493: What are the `seta` and `setb` handlers? I guess these lines should be removed.

ui/settings.glade
```diff
@@ -4449,6 +4449,104 @@ This setting can make it easier to draw with touch. </property>
...
+                                <property name="can-focus">False</property>
+                                <property name="label-xalign">0</property>
+                                <child>
+                                  <object class="GtkAlignment">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="left-padding">12</property>
+                                    <child>
+                                      <!-- n-columns=3 n-rows=3 -->
+                                      <object class="GtkGrid">
```
rolandlo, line 4464: We use an id everywhere. Please add one.

ui/settings.glade
```diff
@@ -4449,6 +4449,104 @@ This setting can make it easier to draw with touch. </property>
                                 <property name="position">1</property>
                               </packing>
                             </child>
+                            <child>
+                              <object class="GtkFrame">
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="label-xalign">0</property>
+                                <child>
+                                  <object class="GtkAlignment">
```
rolandlo, line 4458: Please add an id.

rolandlo (changes requested):
This looks mostly good. I like the third mode you added. Still I have some minor comments.

ui/settings.glade
hyperupcall: Nope, I'll fix that

hyperupcall:
Sorry about some of the careless mistakes - I applied all your suggestions.

In the case of the locking, I modified the code to lock when accessing `doc->getPdfPageCount()` for now - and I'll change it if needed when we get more info about that :)

rolandlo:
Please check the document locking out. You also locked it when inserting a new page. The app might hang.

src/core/control/tools/StrokeHandler.cpp
```diff
@@ -266,11 +268,23 @@ void StrokeHandler::onButtonReleaseEvent(const PositionInputData& pos, double zo
...
+        doc->lock();
+        if (doc->getPdfPageCount() == 0) {
+            auto currentPage = control->getCurrentPageNo();
+            auto lastPage = doc->getPageCount() - 1;
+            if (currentPage == lastPage) {
+                doc->unlock();
+                control->insertNewPage(currentPage + 1, false);
+            }
+        }
+        doc->unlock();
```
rolandlo, line 284: You may unlock an already unlocked mutex here. I would suggest this:
```suggestion
        doc->lock();
        auto pdfPageCount = doc->getPdfPageCount();
        doc->unlock();
        if (pdfPageCount == 0) {
            doc->lock();
            auto currentPage = control->getCurrentPageNo();
            auto lastPage = doc->getPageCount() - 1;
            doc->unlock();
            if (currentPage == lastPage) {
                control->insertNewPage(currentPage + 1, false);
            }
        }
```
Similarly for the other occurence.

hyperupcall: Thank you for your suggestions - they are now applied :)

rolandlo (approved):
LGTM

rolandlo:
Thanks once more for the contribution @hyperupcall It looks really useful. I just merged it.

hyperupcall:
Happy to help!

PR #4312, opened 2022-10-12
Declare a default virtual destructor in XmlNode

hyperupcall:
This fixes the issue where Base classes in C++ need to have an explicitly declared virtual destructor (in this case, default destructor), so when instantitated derived classes (with a static type of the base class) are destructured, it's not UB. This shuts up a compiler warning diagnostic and [isn't present in release-1.1](https://github.com/xournalpp/xournalpp/blob/e2f7f0f562645f4c971b43b6e44dd2dc39c61763/src/control/xml/XmlNode.h#L23)

### DIFF
--- a/src/core/control/xml/XmlNode.h
+++ b/src/core/control/xml/XmlNode.h
@@ -24,6 +24,7 @@ class OutputStream;
 class XmlNode {
 public:
     XmlNode(const char* tag);
+    virtual ~XmlNode() = default;
 
 public:
     void setAttrib(const char* attrib, std::string value);


Technius (approved):
LGTM. I'll probably hold off on merging this until 1.1.2 is released this weekend, unless someone thinks otherwise.

Technius:
Can you rebase on release-1.1 and change the target to release-1.1?

hyperupcall:
@Technius I'm not sure how relevant this would be for `release-1.1`? [Here](https://github.com/xournalpp/xournalpp/blob/release-1.1/src/control/xml/XmlNode.h) it shows that the destructor has already been declared for release-1.1 and it has a user-provided implementation. So this issue doesn't show up there, right?

Technius:
Yes, you're right, in which case I think we can merge this immediately. Now I'm curious why `release-1.1` has it but `master` does not.

Technius:
Yes, you're right, in which case I think we can merge this immediately. Now I'm curious why `release-1.1` has it but `master` does not.

PR #4313, opened 2022-10-12
Add Duplicate Page option under Journal menu

### LINKED ISSUE #3831: Enhancement: add 'Duplicate' as option under 'Journal' menu
It is possible to duplicate a page by right clicking on the page in the left hand page menu bar.
Right clicking on the page isn't immediately intuitive.
As a new user I looked everywhere for this, expecting to find it under Journal (where new pages can be added).

I suggest simply adding the Duplicate option under the Journal menu item to make it easier to find.

(Also a newbie comment: Why is the Journal menu not simply called 'Page' since all the actions seem to relate to the page?)

hyperupcall:
closes #3831


src/core/control/Control.cpp
```diff
@@ -1375,6 +1378,13 @@ void Control::deletePage() {
     this->win->getXournal()->forceUpdatePagenumbers();
 }
 
+void Control::duplicatePage() {
+    auto page = getCurrentPage();
+    auto pageCopy = std::make_shared<XojPage>(*page);
```
rolandlo, line 1383:
```suggestion
    auto page = getCurrentPage();
    if (!page) {
        return;
    }
    auto pageCopy = std::make_shared<XojPage>(*page);
```

rolandlo (changes requested):
Thanks for the contribution. The case where `getCurrentPage()` returns `nullptr` should be taken acccount, as suggested in the code, I think. Otherwise this looks good to me.

rolandlo (commented):
One more change request: 
Please replace the code
https://github.com/hyperupcall/xournalpp/blob/531733db9d6211af56c9a2b447c768536024dc5f/src/core/gui/sidebar/previews/page/SidebarPreviewPages.cpp#L159-L179
by 
```cpp
control->duplicatePage();
break;
``` 
to avoid code duplication.

hyperupcall:
Sorry this took a while to get back to - made those changes :+1:

rolandlo (approved):
LGTM

rolandlo:
Thanks again for the contribution. Great to have the duplicate option in the `Journal` menu and less code duplication.

Issue #1169, opened 2019-05-22
Switching to a selection tool should autoselect any currently editing text object


Technius:
**Is your feature request related to a problem? Please describe.**
When I'm editing a text object and I want to reposition it, I have to switch to the select tool. But then I have to reselect the text object, which is annoying.

**Describe the solution you'd like**
Switching to a selection tool while a text object is being edited should select the text object automatically. 

**Describe alternatives you've considered**
Allow a text object to be repositioned by a drag (would require quite a few changes).

**How to implement**
Edit these lines: https://github.com/xournalpp/xournalpp/blob/dd30d205f753b83b611f7e9120ef2192a5f7b033/src/control/Control.cpp#L632-L649
In each case, check if a text object is being edited. If so, set the selection to that text object.

forgottosave:
I'll have a look at this issue in the next days

WillNilges:
@Technius  Did this get closed in #4315 ?

Technius:
Yes.

PR #4315, opened 2022-10-12
Auto-select edited text object when switching from text- to select-tool

forgottosave:
implements [#1169](https://github.com/xournalpp/xournalpp/issues/1169)

# Issue:
A requested enhancement was auto-selecting edited text objects after switching to a selection tool (TOOL_SELECT_RECT, TOOL_SELECT_REGION, TOOL_SELECT_OBJECT), so that you don't have to re-select a currently edited text object when wanting to move it.
GitHub Issue: [Switching to a selection tool should auto-select any currently editing text object #1169](https://github.com/xournalpp/xournalpp/issues/1169),
part of [Future Major Milestone](https://github.com/xournalpp/xournalpp/milestone/7)

# Implementation:
## general idea
When switching to a selection tool a new method in src/core/control/Control.cpp is being called: `selectAndEndText()`
This method:
- checks if the old tool was the text tool
- gets a reference to the Text object in the current TextEditor in PageView
- calls the `endText()` method to close the TextEditor and add the Text object to the layer (this method existed already)
- adds the Text object to the (if necessary new) Selection in Xournal
## changes needed in other files
1. `getActveTool()` in ToolHandler & `getToolType()` in Tool, to be able to check for the last used Tool
2. add respective headers
3. make removing the text editor in `endText()` in PageView optional (by default removing it, as it was before)
4. make deleting the text editor a separate method (avoid duplicate code)

This is my first contribution to this project, I hope it fits all the requirements. If anything's off I'd be happy about a short feedback so that I can fix it and prevent future complications, but I think everything should be alright :)


src/core/control/ToolHandler.cpp
```diff
@@ -199,6 +199,8 @@ void ToolHandler::addToolChangedListener(ToolChangedCallback listener) {
 
 auto ToolHandler::getTool(ToolType type) const -> Tool& { return *(this->tools[type - TOOL_PEN]); }
 
+auto ToolHandler::getActiveTool() const -> Tool& {return *(this->activeTool); }
```
Technius, line 202: This should return a pointer; if `activeTool` is null, this will segfault.

src/core/control/Control.cpp
```diff
@@ -1084,6 +1087,37 @@ void Control::selectFillAlpha(bool pen) {
     }
 }
 
+void Control::selectAndEndText() {
+    auto xournal = this->win->getXournal();
```
Technius, line 1091: This seems to duplicate existing functionality. I think the logic can be cleaned up as follows:

* First, check whether there is a text editor, and whether the text is currently non-empty. If not, then return. (There is nothing to be done here).
* Save the pointer to the text element.
* Call `clearSelectionEndText()` to close the text editor. Note that if the text element was actually added, it should be on some layer on some page (see `XojPageView::endText()` in `PageView.cpp`).
* If the pointer to the text element is not null, then set the current selection to the text element.

To further clean this up, we can add the type of the selection tool as a function parameter and then move the `selectTool()` logic in `actionPerformed()` into this method, so that all we need to do is call this method with the right tool type.

If you want to follow my suggestion here, I would also suggest renaming this method to something like `handleSelectToolAction`.

src/core/gui/PageView.cpp
```diff
@@ -199,6 +199,12 @@ void XojPageView::endText() {
         }
     }
 
+    if (removeTextEditor) {
+        deleteTextEditor();
+    }
```
Technius, line 204: To me, this doesn't seem to make sense in isolation. If `removeTextEditor` is false, the text editor will be left in an invalid state, since the text element will no longer be in the document anymore.

Technius (changes requested):
Thanks for the PR! I have a few major concerns.

Would also like to get a second opinion from @bhennion, who has refactored some of these files recently.

src/core/control/Control.cpp
forgottosave: I'm not sure if this would actually clean up code.

src/core/control/ToolHandler.cpp
forgottosave: True

src/core/control/Control.cpp
forgottosave: Thank you very much. I think this really cleans up a lot and works just as fine.
Although I'm not sure about moving the `selectTool()` logic in `actionPerformed()` into this method. It only requires another check if the selected tool is one of the selection tools every time it is called.
I still implemented it, but would recommend checking on this again.

forgottosave:
Looking at the error in [Build_Test_Stage Test Xournal++ on MacOS](https://dev.azure.com/xournalpp/xournalpp/_build/results?buildId=21713&view=logs&jobId=69082843-5ef9-5486-682b-24fa9b8c0790) I'm not sure where the issue is something I changed? Could it be an issue with the tests?

> Archive:  ninja-mac.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of ninja-mac.zip or
        ninja-mac.zip.zip, and cannot find ninja-mac.zip.ZIP, period.
##[error]Bash exited with code '9'.

src/core/control/Control.cpp
```diff
@@ -1779,7 +1779,28 @@ void Control::undoRedoPageChanged(PageRef page) {
     }
 }
 
-void Control::selectTool(ToolType type) {
+void Control::handleSelectToolAction(ToolType type) {
```
Technius, line 1782: Based on the way this is implemented, I don't think this method needs to be renamed.

src/core/control/Control.cpp
```diff
@@ -1779,7 +1779,28 @@ void Control::undoRedoPageChanged(PageRef page) {
...
+    // keep text-selection when switching from text to seletion tool
+    auto oldTool = getToolHandler()->getActiveTool();
+    if (oldTool && win
+                && (type == TOOL_SELECT_RECT || type == TOOL_SELECT_REGION || type == TOOL_SELECT_OBJECT)
```
Technius, line 1786: nit: instead of directly checking `type == TOOL_SELECT_RECT || ...`, it would be better to define this expression as a function `isSelectToolType(ToolType)` in `src/core/control/ToolEnums.h`, and then call the function here. This would allow us to potentially add additional selection tools in the future without needing to change this part of the code.

src/core/control/Control.cpp
```diff
@@ -1779,7 +1779,28 @@ void Control::undoRedoPageChanged(PageRef page) {
...
+                && oldTool->getToolType() == ToolType::TOOL_TEXT
+                && !(this->win->getXournal()->getTextEditor()->getText()->getText().empty())) {
+        auto xournal = this->win->getXournal();
+        Text* textobj = std::move(xournal->getTextEditor()->getText());
```
Technius, line 1790: nit: `xournal->getTextEditor()->getText()` returns a pointer, so a move is unnecessary.
```suggestion
        Text* textobj = xournal->getTextEditor()->getText();
```

src/core/control/Control.cpp
```diff
@@ -1779,7 +1779,28 @@ void Control::undoRedoPageChanged(PageRef page) {
...
+        clearSelectionEndText();
+
+        auto pageNr = getCurrentPageNo();
+        XojPageView* view = xournal->getViewFor(pageNr);
+        if (view == nullptr) {
+            return;
+        }
```
Technius, line 1797: Maybe I'm missing something, but under what scenario is `view` null? My understanding is that every page should have a view; if not, there is something wrong going on. Also, if `view` is indeed null, the `return` statement will prevent the tool switch from happening. 

I would replace this with an assertion:
```suggestion
        g_assert(view != nullptr);
```

src/core/control/Control.h
```diff
@@ -209,6 +209,7 @@ class Control:
     void updateDeletePageButton();
 
     // selection handling
+
```
Technius, line 212: nit: please remove this unnecessary line

Technius (commented):
Thanks for the changes. I have a few nitpicks, but other than that, it looks good to me so far.

> But looking at the now occurring test-error in [Build_Test_Stage Test Xournal++ on MacOS](https://dev.azure.com/xournalpp/xournalpp/_build/results?buildId=21713&view=logs&jobId=69082843-5ef9-5486-682b-24fa9b8c0790) I'm not sure whether the issue is something I changed?

I think this is a temporary problem; you can ignore it.

Technius (approved):
LGTM. Merging in 24 hours if there are no objections.

src/core/control/Control.cpp
```diff
@@ -1780,6 +1780,25 @@ void Control::undoRedoPageChanged(PageRef page) {
...
+                && !(this->win->getXournal()->getTextEditor()->getText()->getText().empty())) {
+        auto xournal = this->win->getXournal();
+        Text* textobj = xournal->getTextEditor()->getText();
+        clearSelectionEndText();
+
+        auto pageNr = getCurrentPageNo();
+        XojPageView* view = xournal->getViewFor(pageNr);
+        g_assert(view != nullptr);
+        PageRef page = this->doc->getPage(pageNr);
+        auto selection = new EditSelection(this->undoRedo, textobj, view, page);
```
bhennion, line 1797: Wait! Sorry I took a long time to check this out, but...
I fear there may be some "double ownership" cases here (textobj may be owned by both the layer and the selection). Let me check before merging.

bhennion: Ah, no, it seems to be ok. EditSelection(this->undoRedo, textobj, view, page); removes the element from the layer.

We need to lock the document's mutex though
```suggestion
        Text* textobj = xournal->getTextEditor()->getText();
        clearSelectionEndText();

        auto pageNr = getCurrentPageNo();
        XojPageView* view = xournal->getViewFor(pageNr);
        g_assert(view != nullptr);
        this->doc->lock()
        PageRef page = this->doc->getPage(pageNr);
        auto selection = new EditSelection(this->undoRedo, textobj, view, page);
        this->doc->unlock();
```

bhennion (approved):
LGTM. @Technius, you can merge at will :)

Issue #4316, opened 2022-10-13
Unwanted straight lines appear when drawing ellipses via pen input


tattsan:
**Affects versions :**
 - OS: Arch Linux / X11
 - libgtk: 3.24.34
 - Version of Xournal++: 31d7dad7ba8d566a17cc6480bd62df7b17870489 or after (master)
 - Installation method: build from source

**Describe the bug**
Unwanted straight lines appear when drawing ellipses.

**To Reproduce**
1. Start with Pen tool.
2. Set draw type 'ellipse'.
3. You can draw ellipses or circles as before, but weired segments can be seen.
4. If you only touch and  don't move the pen tool, only the weired segment is drawn.
5. Repat the step4, then you can see weired long lines.

**Expected behavior**
No straight line is displayed.

**Screenshots of Problem**

https://user-images.githubusercontent.com/6355486/195683632-73b1ce70-53f5-4ab6-9c4f-b882e520d510.mp4

**Additional context**
I feel that ellipses are displayed not so smooth as before the commit 31d7dad7ba8d566a17cc6480bd62df7b17870489.
Edit: I've created another issue #4318 on this point.

tattsan:
@bhennion  your recent energetic commits are really great! Could you please look at this report?

bhennion:
I opened a PR which may fix this. @tattsan would you mind trying it out and report in #4317?

bhennion:
This only happens when using the pen (see https://github.com/xournalpp/xournalpp/pull/4317#issuecomment-1278547276), and I can reproduce as well.

The root of the problem is that pen input sequences start with a ButtonPress event, followed by a MotionNotify event (without any motion), while mouse input sequence do not. This should be homogenized.

In any case, fixing the straight lines is really easy: just set BaseShapeHandler::currPoint upon the ButtonPress event. The following MotionNotify event will then be filtered out (as the motion is null, thus to small to satisfy validMotion == true). I'll add this to #4317.

PR #4317, opened 2022-10-13
Fix UB for BaseShapeHandler when clicking without moving

bhennion:
I let an UB slip in when refactoring the BaseShapeHandler: strokes without any points were created, leading to SegFaults upon calls to Stroke::calcsize. This PR fixes that.

It may as well solve #4316 (I can't know for sure, I can't reproduce the issue).
@tattsan would you mind trying this out and telling if this fixes your issue?

### DIFF
--- a/src/core/control/tools/BaseShapeHandler.cpp
+++ b/src/core/control/tools/BaseShapeHandler.cpp
@@ -160,6 +160,7 @@ void BaseShapeHandler::onButtonPressEvent(const PositionInputData& pos, double z
 
     this->startStrokeTime = pos.timestamp;
     this->startPoint = snappingHandler.snapToGrid(this->buttonDownPoint, pos.isAltDown());
+    this->currPoint = this->startPoint;
 
     this->stroke = createStroke(this->control);
 }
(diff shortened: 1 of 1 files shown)


tattsan:
Sorry, 282843b5e6b098d96a7a9289c007e74a6b4a926f doesn't solve #4316.

rolandlo:
I can reproduce the bug and confirm that https://github.com/xournalpp/xournalpp/commit/282843b5e6b098d96a7a9289c007e74a6b4a926f does not fix it. At least that commit fixes a crash I get otherwise using the Ellipse drawing type while drawing ellipses without moving the cursor.

rolandlo:
The bug from #4316 only appears when I use the stylus (not with the mouse). The lines point to the page origin (top left).

bhennion:
The last commit fixes the "tap" event line bugs, but it is not quite enough:

- Enable grid snapping
- Start drawing an ellipse
- Move the cursor back to the original point
- Release the button
- See the ugly line again.

Using `std::atan2` as @tattsan suggested (https://github.com/xournalpp/xournalpp/pull/4320#discussion_r995688691) would cleverly fix that. I'll wait for us to collegially take a decision about the formulas (see #4320) before doing that.

bhennion:
As this is somewhat independent of the ellipse resolution discussed in #4320, I'll be merging this trivial bugfix in 48 hours, unless someone objects.

Issue #4318, opened 2022-10-14
Ellipses don't look smooth


tattsan:
**Affects versions :**
 - Version of Xournal++: 31d7dad7ba8d566a17cc6480bd62df7b17870489 or after (master)

**Describe the bug**
Ellipses don't look smooth. 

**To Reproduce**
1. Zoom up the screen to some extent.
2. Draw an ellipse.

**Expected behavior**
Sure it's drawn as a polygon, but it should be drawn smooth enough that you won't notice the edges. 

**Screenshots of Problem**
![EllipseOrPolygon](https://user-images.githubusercontent.com/6355486/195795563-8ca49978-1845-48ad-adc4-c0970741ee22.png)

**Additional context**
The number of points to take on the ellipse seems somewhat small.
https://github.com/xournalpp/xournalpp/blob/6f4c6f9ab90c294269a50a5e45982e968199d05b/src/core/control/tools/EllipseHandler.cpp#L66-L67

tattsan:
The scale of the graph paper is 5mm, so if you print this, you may not notice the jaggies. However, the display on electronic devices can be enlarged, so I think it would be better to raise the resolution a little more.

PR #4320, opened 2022-10-14
Fix problems when drawing an ellipse

luca-venturini:
Fixed issues #4318 and #4316 relative to the ellipse drawing.

### DIFF
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -64,7 +64,7 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
      * Set resolution depending on the radius (heuristic)
      */
     auto nbPtsPerQuadrant =
-            static_cast<unsigned int>(std::ceil(5 + 0.5 * std::max(std::abs(radiusX), std::abs(radiusY))));
+            static_cast<unsigned int>(std::ceil(5 + 0.2 * std::max(std::abs(radiusX), std::abs(radiusY))));
     const double stepAngle = M_PI_2 / nbPtsPerQuadrant;
 
     std::pair<std::vector<Point>, Range> res;
@@ -79,7 +79,8 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
 
     shape.emplace_back(center_x + radiusX, center_y);
     for (unsigned int j = 1U; j < nbPtsPerQuadrant; j++) {
-        const double centerAngle = stepAngle * j;
+        const double tgtAngle = stepAngle * j;
+        const double centerAngle = 0.25 * (std::atan2(std::abs(radiusY) * std::sin(tgtAngle), std::abs(radiusX) * std::cos(tgtAngle))) + 0.75 * tgtAngle;
         double xp = center_x + radiusX * std::cos(centerAngle);
         double yp = center_y + radiusY * std::sin(centerAngle);
         shape.emplace_back(xp, yp);
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -64,7 +64,7 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
      * Set resolution depending on the radius (heuristic)
      */
     auto nbPtsPerQuadrant =
-            static_cast<unsigned int>(std::ceil(5 + 0.1 * std::max(std::abs(radiusX), std::abs(radiusY))));
+            static_cast<unsigned int>(std::ceil(5 + 0.5 * std::max(std::abs(radiusX), std::abs(radiusY))));
     const double stepAngle = M_PI_2 / nbPtsPerQuadrant;
 
     std::pair<std::vector<Point>, Range> res;
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -64,7 +64,7 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
      * Set resolution depending on the radius (heuristic)
      */
     auto nbPtsPerQuadrant =
-            static_cast<unsigned int>(std::ceil(5 + 0.2 * std::max(std::abs(radiusX), std::abs(radiusY))));
+            static_cast<unsigned int>(std::ceil(5 + 0.3 * std::max(std::abs(radiusX), std::abs(radiusY))));
     const double stepAngle = M_PI_2 / nbPtsPerQuadrant;
 
     std::pair<std::vector<Point>, Range> res;
--- a/src/core/control/tools/EllipseHandler.cpp
+++ b/src/core/control/tools/EllipseHandler.cpp
@@ -66,7 +66,6 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
     auto nbPtsPerQuadrant =
             static_cast<unsigned int>(std::ceil(5 + 0.1 * std::max(std::abs(radiusX), std::abs(radiusY))));
     const double stepAngle = M_PI_2 / nbPtsPerQuadrant;
-    const double excentricity = std::abs(radiusY / radiusX);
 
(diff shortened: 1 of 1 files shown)


tattsan:
Thank you for the fix! To avoid the weired line, you removed @bhennion 's code. But if you use `atan2` here, the bug is also resolved. I'll compare the two results (before applying the second patch).
**Result of @luca-venturini's code**
IgnoreBhennionCode
**Result of @bhennion`s code with atan2**
UsingAtan2

Which is better?

bhennion:
This does not really fix the issue #4316, it just hides it (the way it was hidden before). see https://github.com/xournalpp/xournalpp/issues/4316#issuecomment-1278887338

About the resolution: whatever we decide to do, keep in mind that there is a trade-off between the looks and the number of points (for performance reasons).
For a constant number of edges, the formulas currently in master are better for the "pointy" end of the ellipse, but a little worse on the sides.
In this screenshot, we have twice the same ellipses, drawn with the two formulas.
The left ellipse is computed with master's formula while the right ellipse is with this PR's formula.
Screenshot_20221014_135726

There is probably a better formula in between, simply by changing the coefficients in the linear combination
https://github.com/xournalpp/xournalpp/blob/6f4c6f9ab90c294269a50a5e45982e968199d05b/src/core/control/tools/EllipseHandler.cpp#L85

src/core/control/tools/EllipseHandler.cpp
```diff
@@ -80,9 +79,7 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
 
     shape.emplace_back(center_x + radiusX, center_y);
     for (unsigned int j = 1U; j < nbPtsPerQuadrant; j++) {
-        const double tgtAngle = stepAngle * j;
-        // This formula dispatches the points, so that points are denser where the ellipse has higher curvature.
-        const double centerAngle = 0.5 * (std::atan(excentricity * std::tan(tgtAngle)) + tgtAngle);
+        const double centerAngle = stepAngle * j;
```
tattsan, line 82: We can also resolve this bug by using `atan2` instead of removing @bhennion's code.
```suggestion
        const double tgtAngle = stepAngle * j;
        // This formula dispatches the points, so that points are denser where the ellipse has higher curvature.
        const double centerAngle = 0.5 * (std::atan2(std::abs(radiusY) * std::sin(tgtAngle), std::abs(radiusX) * std::cos(tgtAngle)) + tgtAngle);
```

tattsan:
Sorry, my post went back and forth while I was confused on how to navigate github.
> There is probably a better formula in between, simply by changing the coefficients in the linear combination

It sure looks that way!

tattsan:
Here is a comparison. Number of sample points is the same as master.
Mixtures

tattsan:
Balanced ratio is in betwen 2:8 and 3:7, so 1:3 would be the best. But at the same time, the resolution needs to be increased a bit more to make it same quality as before 31d7dad7ba8d566a17cc6480bd62df7b17870489.

bhennion:
@luca-venturini do you want to implement the changes @tattsan suggested, or would you rather have someone else do it?
Either is perfectly fine, just tell us so that we know whether to keep this PR open or to superseed it with a new one.

luca-venturini:
Ok I can try now

luca-venturini:
I have fixed the formula as @tattsan suggested and I increased a bit the resolution.

bhennion:
Great! Thanks!
@tattsan are you satisfied with the new resolution?

We'll wait until #4333 is fast forwarded before merging this. Once this is done and provide @tattsan agrees, I think we'll be good to go.

src/core/control/tools/EllipseHandler.cpp
```diff
@@ -81,8 +80,7 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
     shape.emplace_back(center_x + radiusX, center_y);
     for (unsigned int j = 1U; j < nbPtsPerQuadrant; j++) {
         const double tgtAngle = stepAngle * j;
-        // This formula dispatches the points, so that points are denser where the ellipse has higher curvature.
-        const double centerAngle = 0.5 * (std::atan(excentricity * std::tan(tgtAngle)) + tgtAngle);
+        const double centerAngle = 0.25 * (std::atan2(std::abs(radiusY) * std::sin(tgtAngle), std::abs(radiusX) * std::cos(tgtAngle))) + 0.75 * tgtAngle;
```
tattsan, line 83: If we use this formula, please leave @bhennion's comment.

tattsan:
Here is a comparison. It seems that the edges are more noticeable on a circle rather than a flattened ellipse.
ResolutionComparison

If the coefficient is less than 0.5, it would be better to have @bhennion 's dispatch, and if it is 0.5 or higher, the first code of @luca-venturini  seems to be sufficient. Personally, I think the first code of @luca-venturini  is better. If there are issues of performance, I would leave the decision to the core team.

bhennion:
The two formulas gives the same edges for a circle (a regular polygon, up to rounding errors). The question is the number of edges. I guess the `std::max(std::abs(radiusX), std::abs(radiusY))` formula for the number of edge is less appropriate for circles than for ellipses. We can try replacing it by `std::abs(radiusX) + std::abs(radiusY)`, thus doubling the number of edges for a circle (vs a flat ellipse). The coefficient in front could e.g. be set to 0.3 (as opposed to 0.2).

About performance: Xournal++ globally has some performance issues, but the ellipse tool is such a niche, I don't think it matters so much :) Also, in the long run, this discretisation should get replaced by splines, so this will eventually be removed.
So... I guess you guys can just make a choice, and I'll shut up about it :)

luca-venturini:
I have used two different formulas to compute the number of points on the basis of the eccentricity of the ellipses, I have also corrected the value of the coefficient to 0.3.

tattsan:
> I guess the std::max(std::abs(radiusX), std::abs(radiusY)) formula for the number of edge is less appropriate for circles than for ellipses. 

That's right. I's exactly why I felt the division was rough on the circle.

>We can try replacing it by std::abs(radiusX) + std::abs(radiusY), thus doubling the number of edges for a circle (vs a flat ellipse).

I agree.

> Also, in the long run, this discretisation should get replaced by splines, 

Exactly the answer I wanted to hear!

src/core/control/tools/EllipseHandler.cpp
```diff
@@ -63,8 +63,16 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
     /*
      * Set resolution depending on the radius (heuristic)
      */
-    auto nbPtsPerQuadrant =
-            static_cast<unsigned int>(std::ceil(5 + 0.2 * std::max(std::abs(radiusX), std::abs(radiusY))));
+    unsigned int nbPtsPerQuadrant;
+
+    const double excentricity = std::abs(radiusY / radiusX);
```
tattsan, line 68: I see no need to separate cases by `excentricity`, since rx + ry is an approximation to max(rx,ry) if either rx or ry is extremely smaller than the other. @bhennion's formula is good for all cases.

src/core/control/tools/EllipseHandler.cpp
```diff
@@ -63,8 +63,16 @@ auto EllipseHandler::createShape(bool isAltDown, bool isShiftDown, bool isContro
...
+
+
+    // Change the formula used to compute the points to construct the ellipse on the basis of the ellipse's excentricity 
+    if (excentricity < 0.4)
```
tattsan, line 72: If you see `0 < excentricity < 0.4` as  flattened (horizontally),  then `2.5 < excentricity` is also flattened (vertically). But as I wrote above, there is no need to separate cases.

tattsan:
LGTM. Thank you so much!

bhennion (approved):
Thanks to both of you for the work!

tattsan:
I then learned of Ramanujan's concise approximation of the perimeter of an ellipse:
$L\approx\pi\Big(3(a+b)-\sqrt{(3a+b)(a+3b)}\Bigr)$ where $a$ and $b$ are semi-axes of the ellipse.
The following site has more details.
http://www.ebyte.it/library/docs/math05a/EllipsePerimeterApprox05.html

PR #4351, opened 2022-10-22
Adds missing check to selectTool()

forgottosave:
A bug added in [this](https://github.com/xournalpp/xournalpp/pull/4315) pull request where xournalpp crashes when switching from the text tool to the selection tool without having a text selected. This occurred because of a missing check in the if statement.
This check is being added with this pull request


bhennion:
Thanks for the PR, but this is a duplicate of #4347...

forgottosave:
Oh, didn't realize that

Issue #2750, opened 2021-02-05
Add simple possibility to change layer of an object


Oli72:
**Is your feature request related to a problem? Please describe.**
"I'm always frustrated when" I want to change the layer of an object: I have to cut the object, change the layer, and paste it. But then the position changes.

**Describe the solution you'd like**
It would be nice to have a direct possibility to change the layer of an object, e.g. by a context menu entry.

**Describe alternatives you've considered**
./.

**Additional context**
./.

forgottosave:
I think this would be very useful. I'll take a look at this

PR #4352, opened 2022-10-22
Adds feature to change layer of an object

forgottosave:
## Issue
Feature Request:
Add simple possibility to change layer of an object #2750

## Feature description
Adds a feature to move all selected objects to another layer.
This feature can be found in `Journal > Move Selection Up / Down`, as can be seen in the following screenshot: 
image

#### main functionality
- Control.cpp:
  1) add cases `ACTION_MOVE_SELECTION_LAYER_UP` & `ACTION_MOVE_SELECTION_LAYER_DOWN`
  2) call new method `moveSelectionToLayer(layerNo)` in these cases
- Control::`moveSelectionToLayer(layerNo)`:
  1) check if new layer is legal
  2) move selected objects to new layer
  3) add UndoAction
  4) switch to new layer
#### further changes needed
- ActionType.enum.h:
  1) add `ACTION_MOVE_SELECTION_LAYER_UP` & `ACTION_MOVE_SELECTION_LAYER_DOWN`
- ActionType.generated.cpp
  1) add both actions to `ActionType_fromString()` 
  2) add both actions to `ActionType_toString()`
- main.glade
  1) insert both options in the Journal menu
- MainWindow.cpp
  1) `fireEnableAction()` in control with according limitations
- MoveSelectionToLayerUndoAction.h/cpp: *new files*
  1) add the according undo-action


forgottosave:
This has a small issue left with the layer switching, keeping it a draft until it's fixed

forgottosave:
@Technius, (or any other member,) is there any update on this PR? Should I change something?

src/core/control/Control.cpp
```diff
@@ -2979,6 +2986,31 @@ void Control::clipboardPasteXournal(ObjectInputStream& in) {
...
+    auto selectedElements = selection->getElements();
+    auto oldLayer = currentP->getSelectedLayer();
+    auto newLayer = currentP->getLayers()->at(layerNo);
+    auto moveSelUndo = std::make_unique<MoveSelectionToLayerUndoAction>(currentP, getLayerController(), oldLayer, currentP->getSelectedLayerId() - 1, layerNo);
+    clearSelectionEndText();
+    for (auto e : selectedElements) {
+        currentP->getSelectedLayer()->removeElement(e,false);
+        currentP->getLayers()->at(layerNo)->addElement(e);
+        moveSelUndo->addElement(newLayer, e, newLayer->indexOf(e));
+    }
```
bhennion, line 3008: So, this:

1. Takes selected elements and add them back onto the original layer (`clearSelectionEndText()`)
2. Removes the elements we just added from the original layer
3. Adds them to the target layer

This add-remove-add sequence is a waste.

bhennion (changes requested):
I disagree with the implement logic: I think that pressing the menu item should just modify a LayerIndex member of the selection itself, without effectively modifying the document.
The elements should still be selected after the operation (e.g. if you want to move by three layers by pressing the button three times).
They would only be added to the target layer once the selection is cleared.

[This means modifying EditSelection and/or EditSelectionContent to store the target layer, I guess]

src/core/control/Control.cpp
forgottosave: I agree, this is a weird sequence. Although I'd propose adding the option to not clear the selection when calling `getLayerController()->switchToLay(...)`, since now it basically is a layer switch without clearing the selection (draging the selection with you" while changing layer).
In that case it wouldn't be necessary to change EditSelection and/or EditSelectionContent.

bhennion: > Although I'd propose adding the option to not clear the selection when calling `getLayerController()->switchToLay(...)`

Yes, that sounds good!

forgottosave: Then I'll implement it that way and change the UndoAction accordingly

src/core/control/Control.cpp
```diff
@@ -537,6 +538,12 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GdkEvent* even
         case ACTION_PAPER_BACKGROUND_COLOR:
             changePageBackgroundColor();
             break;
+        case ACTION_MOVE_SELECTION_LAYER_UP:
+            moveSelectionToLayer(getCurrentPage()->getSelectedLayerId());
+            break;
+        case ACTION_MOVE_SELECTION_LAYER_DOWN:
+            moveSelectionToLayer(getCurrentPage()->getSelectedLayerId() - 2);
```
Technius, line 545: The layer id is unsigned, so the subtraction here can cause integer overflow. We should check that `getSelectedLayerId() >= 2` (e.g., ensure that the lowest possible layer number calculated here is 0).

For readability, I would also explicitly indicate that the parameter is a layer number by moving the calculation into another variable:
```suggestion
            Layer::Index currentLayerId = getCurrentPage()->getSelectedLayerId();
            if (currentLayerId >= 2) {
                size_t downLayerNo = currentLayerId - 2;
                moveSelectionToLayer(downLayerNo);
            }
```

src/core/control/Control.cpp
forgottosave: In the UndoAction I left the removing and adding to layers as when undoing / redoing the elements might not be selected anymore. It is more intuitive here to just move the elements from the fixed layers.

src/core/control/Control.cpp
```diff
@@ -2979,6 +2992,27 @@ void Control::clipboardPasteXournal(ObjectInputStream& in) {
...
+
+    auto* oldLayer = currentP->getSelectedLayer();
+    auto* newLayer = currentP->getLayers()->at(layerNo);
+    auto moveSelUndo = std::make_unique<MoveSelectionToLayerUndoAction>(currentP, getLayerController(), oldLayer, currentP->getSelectedLayerId() - 1, layerNo);
+    for (auto* e : selection->getElements()) {
+        moveSelUndo->addElement(newLayer, e, newLayer->indexOf(e));
+    }
+    undoRedo->addUndoAction(std::move(moveSelUndo));
+
+    getLayerController()->switchToLay(layerNo + 1, false, false);
```
Technius, line 3013: nit: the optional parameters should be commented:
```suggestion
    getLayerController()->switchToLay(layerNo + 1, /*hideShow=*/false, /*clearSelection=*/false);
```

src/core/undo/MoveSelectionToLayerUndoAction.cpp
```diff
@@ -0,0 +1,64 @@
...
+        this->undone = false;
+        return false;
+    }
+
+    for (const auto& elem: elements) {
+        elem.layer->removeElement(elem.element, false);
+        this->oldLayer->addElement(elem.element);
+    }
+
+    this->layerController->switchToLay(oldLayerNo + 1, false, false);
```
Technius, line 41: nit: should comment the optional parameters (similar to the other review comment)

Technius (commented):
Code looks good to me. However, I noticed that the selection menu items are in the `Journal` menu. This is unintuitive because the `Journal` menu contains options related to manipulating pages/layers/document, not individual elements. It would be better to put the new menu items under `Edit`.

bhennion (approved):
LGTM!

ui/main.glade
```diff
@@ -466,6 +466,32 @@
...
                           </object>
                         </child>
+                        <child>
+                          <object class="GtkSeparatorMenuItem" id="separatormenuitem19">
+                            <property name="visible">True</property>
+                            <property name="can-focus">False</property>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkMenuItem" id="menuJournalMoveSelectionLayerUp">
```
rolandlo, line 476: Nitpick: The `id` should have been renamed, since these menu items are under menu `Edit` now.
Also the PR description should have been updated accordingly (which is easy to do now).

ui/main.glade
```diff
@@ -466,6 +466,32 @@
...
+                            <property name="name">menuJournalMoveSelectionLayerUp</property>
+                            <property name="visible">True</property>
+                            <property name="can-focus">False</property>
+                            <property name="label" translatable="yes">Move Selection Layer Up</property>
+                            <property name="use-underline">True</property>
+                            <signal name="activate" handler="ACTION_MOVE_SELECTION_LAYER_UP" swapped="no"/>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkMenuItem" id="menuJournalMoveSelectionLayerDown">
```
rolandlo, line 486: Here as well

PR #4302, opened 2022-10-10
Fix #4301

bhennion:
Fix #4301

PR #4359, opened 2022-10-23
[Lua API] extend getToolInfo for selection and add getStrokes as inverse of addStrokes

atticus-sullivan:
Outsourcing of the luaApi stuff of PR https://github.com/xournalpp/xournalpp/pull/2176.

The aim is to have a function `getStrokes` which serves as inverse of the existing `addStrokes`.

TODO:
- [x] getWidthAndHeight -> `getToolInfo("selection")`
- [x] getStrokes
  - [x] documentation of in-/output
  - [x] style stuff (moving lines, outsourcing to helper function)
  - [ ] ~~testing~~

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -25,6 +25,7 @@
 #include "control/layer/LayerController.h"
 #include "control/pagetype/PageTypeHandler.h"
 #include "control/settings/Settings.h"
+#include "control/tools/EditSelection.h"
 #include "gui/Layout.h"
 #include "gui/MainWindow.h"
 #include "gui/XournalView.h"
@@ -42,7 +43,7 @@
 #include "undo/InsertUndoAction.h"
 #include "util/StringUtils.h"
 #include "util/XojMsgBox.h"
-#include "util/i18n.h"                  // for _
+#include "util/i18n.h"  // for _
 #include "util/safe_casts.h"
 
 extern "C" {
@@ -809,6 +810,141 @@ static int applib_addStrokes(lua_State* L) {
     lua_pop(L, 1);  // Stack is now the same as it was on entry to this function
 
     return 0;
+}
+
+/**
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Is inverse to app.addStrokes
+ *
+ * Required argument: type ("selection" or "layer")
+ *
+ * Example: local strokes = app.getStrokes("selection")
+ *
+ * possible return value:
+ * {
+ *         {   -- Inside a stroke are three tables of equivalent length that represent a series of points
+ *             ["x"]        = { [1] = 110.0, [2] = 120.0, [3] = 130.0, ... },
+ *             ["y"]        = { [1] = 200.0, [2] = 205.0, [3] = 210.0, ... },
+ *             -- pressure is only present if pressure is set -> pressure member might be nil
+ *             ["pressure"] = { [1] = 0.8,   [2] = 0.9,   [3] = 1.1, ... },
+ *             -- Each stroke has individually handled options
+ *             ["tool"] = "pen",
+ *             ["width"] = 3.8,
+ *             ["color"] = 0xa000f0,
+ *             ["fill"] = 0,
+ *             ["lineStyle"] = "plain",
+ *         },
+ *         {
+ *             ["x"]         = {207, 207.5, 315.2, 315.29, 207.5844},
+ *             ["y"]         = {108, 167.4, 167.4, 108.70, 108.7094},
+ *             ["tool"]      = "pen",
+ *             ["width"]     = 0.85,
+ *             ["color"]     = 16744448,
+ *             ["fill"]      = -1,
+ *             ["lineStyle"] = "plain",
+ *         },
+ *         {
+ *             ["x"]         = {387.60, 387.6042, 500.879, 500.87, 387.604},
+ *             ["y"]         = {153.14, 215.8661, 215.866, 153.14, 153.148},
+ *             ["tool"]      = "pen",
+ *             ["width"]     = 0.85,
+ *             ["color"]     = 16744448,
+ *             ["fill"]      = -1,
+ *             ["lineStyle"] = "plain",
+ *         },
+ * }
+ */
+static int applib_getStrokes(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    std::string type = luaL_checkstring(L, 1);
+    std::vector<Element*> elements = {};
+    Control* control = plugin->getControl();
+
+    if (type == "Layer") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            control->clearSelection();  // otherwise strokes in the selection won't be recognized
+        }
+        elements = control->getCurrentPage()->getSelectedLayer()->getElements();
+    } else if (type == "selection") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            elements = sel->getElements();
+        } else {
+            g_warning("There is no selection! ");
+            return 0;
+        }
+    } else {
+        g_warning("Unknown argument: %s", type.c_str());
+        return 0;
+    }
+
+    lua_newtable(L);  // create table of the elements
+    int currStrokeNo = 0;
+    int currPointNo = 0;
+
+    for (Element* e: elements) {
+        if (e->getType() == ELEMENT_STROKE) {
+            auto* s = static_cast<Stroke*>(e);
+            lua_pushnumber(L, ++currStrokeNo);  // index for later (settable)
+            lua_newtable(L);                    // create stroke table
+
+            lua_newtable(L);  // create table of x-coordinates
+            for (auto p: s->getPointVector()) {
+                lua_pushnumber(L, ++currPointNo);
+                lua_pushnumber(L, p.x);
+                lua_settable(L, -3);  // pops key and value from stack
+            }
+            lua_setfield(L, -2, "x");  // add x-coordinates to stroke
+            currPointNo = 0;
+
+            lua_newtable(L);  // create table for y-coordinates
+            for (auto p: s->getPointVector()) {
+                lua_pushnumber(L, ++currPointNo);
+                lua_pushnumber(L, p.y);
+                lua_settable(L, -3);
+            }
+            lua_setfield(L, -2, "y");  // add y-coordinates to stroke
+            currPointNo = 0;
+
+            if (s->hasPressure()) {
+                lua_newtable(L);  // create table for pressures
+                for (auto p: s->getPointVector()) {
+                    lua_pushnumber(L, ++currPointNo);
+                    lua_pushnumber(L, p.z);
+                    lua_settable(L, -3);
+                }
+                lua_setfield(L, -2, "pressure");  // add pressures to stroke
+                currPointNo = 0;
+            }
+
+            StrokeTool tool = s->getToolType();
+            if (tool == StrokeTool::PEN) {
+                lua_pushstring(L, "pen");
+            } else if (tool == StrokeTool::HIGHLIGHTER) {
+                lua_pushstring(L, "highlighter");
+            } else {
+                g_warning("Unknown STROKE_TOOL. ");
+                return 0;
+            }
+            lua_setfield(L, -2, "tool");  // add tool to stroke
+
+            lua_pushnumber(L, s->getWidth());
+            lua_setfield(L, -2, "width");  // add width to stroke
+
+            lua_pushinteger(L, int(uint32_t(s->getColor())));
+            lua_setfield(L, -2, "color");  // add color to stroke
+
+            lua_pushinteger(L, s->getFill());
+            lua_setfield(L, -2, "fill");  // add fill to stroke
+
+            lua_pushstring(L, StrokeStyle::formatStyle(s->getLineStyle()).c_str());
+            lua_setfield(L, -2, "lineStyle");  // add linestyle to stroke
+
+            lua_settable(L, -3);  // add stroke to elements
+        }
+    }
+    return 1;
 }
 
 /**
@@ -1762,6 +1898,7 @@ static const luaL_Reg applib[] = {{"msgbox", applib_msgbox},
                                   {"addSplines", applib_addSplines},
                                   {"getFilePath", applib_getFilePath},
                                   {"refreshPage", applib_refreshPage},
+                                  {"getStrokes", applib_getStrokes},
                                   // Placeholder
                                   //	{"MSG_BT_OK", nullptr},
 
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -1116,6 +1116,24 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
 }
 
 /*
+ * pushes an rectangle (width, height, x, y) to the lua stack as a new table
+ * in most cases you'll want to do a
+ * lua_newtable(L);  // create rectangle table
+ * before and a
+ * lua_setfield(L, -2, "rectangleTable");  // add rectangle table to other table
+ * after this function
+ */
+static void pushRectangleHelper(lua_State* L, xoj::util::Rectangle<double> rect) {
+    lua_pushnumber(L, rect.width);
+    lua_setfield(L, -2, "width");
+    lua_pushnumber(L, rect.height);
+    lua_setfield(L, -2, "height");
+    lua_pushnumber(L, rect.x);
+    lua_setfield(L, -2, "x");
+    lua_pushnumber(L, rect.y);
+    lua_setfield(L, -2, "y");
+}
+/*
  * Returns a table encoding all info on the chosen tool (active, pen, highlighter, eraser or text)
  * in a Lua table of one of the following shapes
  *
@@ -1175,6 +1193,33 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
  *
  * See /src/control/ToolEnums.cpp for possible values of "size".
  *
+ * for seiection:
+ * {
+ *   -- bounding box as drawn in the UI (includes padding on all sides)
+ *   "boundingBox" = {
+ *      "width"  = number
+ *      "height" = number
+ *      "x"      = number
+ *      "y"      = number
+ *   }
+ *   -- same as "boundingBox" but the state before any transformation was applied
+ *   "originalBounds" = {
+ *      "width"  = number
+ *      "height" = number
+ *      "x"      = number
+ *      "y"      = number
+ *   }
+ *   -- bounds used for snapping (doesn't include padding and doesn't account to line width)
+ *   -- for more information see https://github.com/xournalpp/xournalpp/pull/4359#issuecomment-1304395011
+ *   "snappedBounds" = {
+ *      "width"  = number
+ *      "height" = number
+ *      "x"      = number
+ *      "y"      = number
+ *   }
+ *   "rotation" = number
+ *   "isRotationSupported" = bool
+ * }
  *
  * Example 1: local penInfo = app.getToolInfo("pen")
  *            local size = penInfo["size"]
@@ -1202,6 +1247,11 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
  * Example 6: local highlighterInfo = app.getToolInfo("highlighter")
  *            local sizeName = highlighterInfo["size"]["name"]
  *            local opacity = highlighterInfo["fillOpacity"]
+ *
+ * Example 7: local selectionInfo = app.getToolInfo("selection")
+ *            local rotation = selectionInfo["rotation"]
+ *            local boundingX = selectionInfo["boundingBox"]["x"]
+ *            local snappedBoundsWidth = selectionInfo["snappedBounds"]["width"]
  */
 
 static int applib_getToolInfo(lua_State* L) {
@@ -1388,6 +1438,32 @@ static int applib_getToolInfo(lua_State* L) {
         lua_pushliteral(L, "color");
         lua_pushinteger(L, int(uint32_t(color)));
         lua_settable(L, -3);
+    } else if (strcmp(mode, "selection") == 0) {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (!sel) {
+            g_warning("There is no selection! ");
+            return 0;
+        }
+        auto rect = sel->getRect();
+
+        lua_newtable(L);  // create return table
+                          //
+        lua_pushnumber(L, sel->getRotation());
+        lua_setfield(L, -2, "rotation");
+        lua_pushboolean(L, sel->isRotationSupported());
+        lua_setfield(L, -2, "isRotationSupported");
+
+        lua_newtable(L);  // create originalBounds table
+        pushRectangleHelper(L, sel->getOriginalBounds());
+        lua_setfield(L, -2, "originalBounds");  // add originalBounds table to return table
+
+        lua_newtable(L);  // create snappedBounds table
+        pushRectangleHelper(L, sel->getSnappedBounds());
+        lua_setfield(L, -2, "snappedBounds");  // add snappedBounds table to return table
+
+        lua_newtable(L);  // create boundingBox table
+        pushRectangleHelper(L, rect);
+        lua_setfield(L, -2, "boundingBox");  // add boundingBox table to return table
     }
     return 1;
 }
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -272,6 +272,26 @@ auto EditSelection::getRect() const -> Rectangle<double> {
 }
 
 /**
+ * gets the minimal bounding box containing all elements of the selection used for e.g. grid snapping
+ */
+auto EditSelection::getSnappedBounds() const -> Rectangle<double> { return Rectangle<double>{this->snappedBounds}; }
+
+/**
+ * get the original bounding rectangle in document coordinates
+ */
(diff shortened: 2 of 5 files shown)


atticus-sullivan:
Hm compiles for me, `STROKE_TOOL_PEN` is available in `addStrokes` but error on usage in `getStrokes` :thinking:. Gonna check this the next days.

rolandlo:
> Hm compiles for me, `STROKE_TOOL_PEN` is available in `addStrokes` but error on usage in `getStrokes` thinking. Gonna check this the next days.

 It compiles on my machine, too. I'm also confused why it didn't compile; `STROKE_TOOL_PEN` is defined in `model/Stroke.h`, which is included (and used at other places in the same file, too).

src/core/plugin/luapi_application.h
```diff
@@ -811,6 +812,105 @@ static int applib_addStrokes(lua_State* L) {
...
+                for (auto p: s->getPointVector()) {
+                    lua_pushnumber(L, ++currPointNo);
+                    lua_pushnumber(L, p.z);
+                    lua_settable(L, -3);
+                }
+                lua_setfield(L, -2, "pressure");  // add pressures to stroke
+                currPointNo = 0;
+            }
+
+            // TODO split up like addStrokesHelper?
```
rolandlo, line 884: The idea of splitting `addStrokesHelper` off was to avoid code duplication (it is used in `addStrokes` and `addSplines`). There is currently no need to do so here.

src/core/plugin/luapi_application.h
```diff
@@ -811,6 +812,105 @@ static int applib_addStrokes(lua_State* L) {
...
+        return 0;
+    }
+
+    lua_newtable(L);  // create table of the elements
+    int currStrokeNo = 0;
+    int currPointNo = 0;
+
+    for (Element* e: elements) {
+        if (e->getType() == ELEMENT_STROKE) {
+            auto* s = dynamic_cast<Stroke*>(e);
```
rolandlo, line 851:
```suggestion
            auto* s = static_cast<Stroke*>(e);
```
looks preferable, since we have already checked that it is a `Stroke`

bhennion:
> > Hm compiles for me, `STROKE_TOOL_PEN` is available in `addStrokes` but error on usage in `getStrokes` thinking. Gonna check this the next days.
> 
> It compiles on my machine, too. I'm also confused why it didn't compile; `STROKE_TOOL_PEN` is defined in `model/Stroke.h`, which is included (and used at other places in the same file, too).

`STROKE_TOOL_PEN` and the like got renamed into `StrokeTool::PEN` when #4302 got merged.

rolandlo:
> > > Hm compiles for me, `STROKE_TOOL_PEN` is available in `addStrokes` but error on usage in `getStrokes` thinking. Gonna check this the next days.
> > 
> > 
> > It compiles on my machine, too. I'm also confused why it didn't compile; `STROKE_TOOL_PEN` is defined in `model/Stroke.h`, which is included (and used at other places in the same file, too).
> 
> `STROKE_TOOL_PEN` and the like got renamed into `StrokeTool::PEN` when #4302 got merged.

Ok @atticus-sullivan then please rebase on current master and replace `STROKE_TOOL_PEN` by `StrokeTool::PEN` and `STROKE_TOOL_HIGHLIGHTER` by `StrokeTool::HIGHLIGHTER`

src/core/plugin/luapi_application.h
atticus-sullivan: Yes, didn't wanted to think about this at the moment I wrote/adapted the function. Just thought maybe if we want to do something like `getTexts` there might be some code which can be shared, but at first glimpse only the color attribute is kinda shared (inherited from `Element`).
So yes I agree there is no need to extract this code (at the moment, and probably won't be without other major changes).

atticus-sullivan:
Alright :+1:  (still wondering why this was a problem since the state of this branch should have been consistent, so `STROKE_TOOL_PEN` should have been the right thing, but well let's just test it out).

rolandlo:
Suggested edit:
```diff
diff --git a/src/core/plugin/luapi_application.h b/src/core/plugin/luapi_application.h
index 1994c518e..89b29010b 100644
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -43,7 +43,7 @@
 #include "undo/InsertUndoAction.h"
 #include "util/StringUtils.h"
 #include "util/XojMsgBox.h"
-#include "util/i18n.h"                  // for _
+#include "util/i18n.h"  // for _
 #include "util/safe_casts.h"
 
 extern "C" {
@@ -852,7 +852,7 @@ static int applib_getStrokes(lua_State* L) {
             lua_pushnumber(L, ++currStrokeNo);  // index for later (settable) TODO move down to the settable call?
             lua_newtable(L);                    // create stroke table
 
-            lua_newtable(L);       // create table of x-coordinates
+            lua_newtable(L);  // create table of x-coordinates
             for (auto p: s->getPointVector()) {
                 lua_pushnumber(L, ++currPointNo);
                 lua_pushnumber(L, p.x);
@@ -861,7 +861,7 @@ static int applib_getStrokes(lua_State* L) {
             lua_setfield(L, -2, "x");  // add x-coordinates to stroke
             currPointNo = 0;
 
-            lua_newtable(L);       // create table for y-coordinates
+            lua_newtable(L);  // create table for y-coordinates
             for (auto p: s->getPointVector()) {
                 lua_pushnumber(L, ++currPointNo);
                 lua_pushnumber(L, p.y);
@@ -895,14 +895,14 @@ static int applib_getStrokes(lua_State* L) {
             lua_pushnumber(L, s->getWidth());
             lua_setfield(L, -2, "width");  // add width to stroke
 
-            lua_pushnumber(L, int(uint32_t(s->getColor())));
+            lua_pushinteger(L, int(uint32_t(s->getColor())));
             lua_setfield(L, -2, "color");  // add color to stroke
 
             lua_pushinteger(L, s->getFill());
             lua_setfield(L, -2, "fill");  // add fill to stroke
 
             lua_pushstring(L, StrokeStyle::formatStyle(s->getLineStyle()).c_str());
-            lua_setfield(L, -2, "linestyle");  // add linestyle to stroke
+            lua_setfield(L, -2, "lineStyle");  // add linestyle to stroke
 
             lua_settable(L, -3);  // add stroke to elements
         }

```

rolandlo:
The left hand was painted by myself, the (mirrored) right half was generated by this Test plugin: [Test.zip](https://github.com/xournalpp/xournalpp/files/9909332/Test.zip)


grafik

rolandlo:
One more remark: `app.getWidthAndHeight` is only partly needed, since we now have `app.getDocumentStructure`, which also contains the page width and height. For selection width and height you should extend `app.getToolInfo`, so that you can call `app.getToolInfo("selection")` to obtain the width, height (and more) of the selection.

atticus-sullivan:
> you should extend `app.getToolInfo`, so that you can call `app.getToolInfo("selection")` to obtain the width, height (and more) of the selection

Ah nice, I'm gonna have a look at this. Just for structuring, should this be in a new PR?

rolandlo:
> Ah nice, I'm gonna have a look at this. Just for structuring, should this be in a new PR?

Same PR is fine.

rolandlo:
> 
> Alright, tbh I hadn't tested it up to now and I didn't really found a difference between these ones (in lua this is a float either way as far as I know).

In Lua >=5.3 there is a difference between integer and number. From the [manual](https://www.lua.org/manual/5.3/manual.html):
> The type number uses two internal representations, or two subtypes, one called integer and the other called float. Lua has explicit rules about when each representation is used, but it also converts between them automatically as needed (see [§3.4.3](https://www.lua.org/manual/5.3/manual.html#3.4.3)). Therefore, the programmer may choose to mostly ignore the difference between integers and floats or to assume complete control over the representation of each number. Standard Lua uses 64-bit integers and double-precision (64-bit) floats, but you can also compile Lua so that it uses 32-bit integers and/or single-precision (32-bit) floats. The option with 32 bits for both integers and floats is particularly attractive for small machines and embedded systems. (See macro LUA_32BITS in file luaconf.h.)

You will see the difference in https://github.com/rolandlo/xournalpp/blob/7d484833d1850634ec094ef12a4413f19498247d/src/core/plugin/luapi_application.h#L478-L479

atticus-sullivan:
Ah, I see (didn't really noticed this when working with lua alone without C-Bindings). Thanks for explaining :+1:

atticus-sullivan:
I just played around with this a bit, and for me as some sort of a newcomer to lua in xournalpp it seems a bit weird to not have the `app.` library throwing lua exceptions (see https://www.lua.org/pil/24.3.2.html and https://www.lua.org/manual/5.4/manual.html#luaL_error). Is this something you already discussed and is intended?
Just asking because with the lua exception (aka error) we get much more precise information (file + line) where the error occurred (and with `pcall` the plugin author still can "catch" the lua-error). Otherwise e.g. simply `nil` is returned and the error is generated when using this value (if `nil` isn't catched)

Without `luaL_error`:
image

With `luaL_error` (+ as I didn't catch the error, the user got a popup displaying the message of the error (+ file and line which might be too much information for the end user)):
image

atticus-sullivan:
Oh and I wrote two functions which might be interesting in general when working with `getStrokes`:
```lua
local function stroke_print(v1, indent)
    indent = indent or 4
    io.write(string.rep(" ", indent),"{\n")
    for _,key in ipairs{"x", "y", "pressure"} do
        if v1[key] then
            io.write(string.rep(" ", indent*2), key, " = {")
            for i,v in ipairs(v1[key]) do
                io.write(v, ", ")
                if i > 8 then io.write("...") break end
            end
            io.write("},\n")
        end
    end
    for _,key in ipairs{"tool", "width", "color", "fill", "lineStyle"} do
        io.write(string.rep(" ", indent*2), key, " = ", v1[key], ",\n")
    end
    io.write(string.rep(" ", indent), "},\n")
end
local function strokes_print(t, indent)
    indent = indent or 4
    io.write("{\n")
    for _,stroke in ipairs(t) do
        stroke_print(stroke, indent)
    end
    io.write("}\n")
end
```
(`stroke..` vs `strokes...` isn't good naming though, I know).

Prints something like
image


Do you think these are interesting enough to somehow ship them with xournalpp (and if so, is there actually a way providing it that easily)?

atticus-sullivan:
Are there tests for the luaApi functions? (or moreover do I have to write some?)

rolandlo:
> I just played around with this a bit, and for me as some sort of a newcomer to lua in xournalpp it seems a bit weird to not have the `app.` library throwing lua exceptions (see https://www.lua.org/pil/24.3.2.html and https://www.lua.org/manual/5.4/manual.html#luaL_error). Is this something you already discussed and is intended? Just asking because with the lua exception (aka error) we get much more precise information (file + line) where the error occurred (and with `pcall` the plugin author still can "catch" the lua-error). Otherwise e.g. simply `nil` is returned and the error is generated when using this value (if `nil` isn't catched)
> 
> Without `luaL_error`: image
> 
> With `luaL_error` (+ as I didn't catch the error, the user got a popup displaying the message of the error (+ file and line which might be too much information for the end user)): image

No this hasn't been discussed. Seems nobody thought of it. Feel welcome to submit a PR about it. Knowing the line number where a plugin fails would certainly be useful.

rolandlo:
> Oh and I wrote two functions which might be interesting in general when working with `getStrokes`:
> 
> ```lua
> local function stroke_print(v1, indent)
>     indent = indent or 4
>     io.write(string.rep(" ", indent),"{\n")
>     for _,key in ipairs{"x", "y", "pressure"} do
>         if v1[key] then
>             io.write(string.rep(" ", indent*2), key, " = {")
>             for i,v in ipairs(v1[key]) do
>                 io.write(v, ", ")
>                 if i > 8 then io.write("...") break end
>             end
>             io.write("},\n")
>         end
>     end
>     for _,key in ipairs{"tool", "width", "color", "fill", "lineStyle"} do
>         io.write(string.rep(" ", indent*2), key, " = ", v1[key], ",\n")
>     end
>     io.write(string.rep(" ", indent), "},\n")
> end
> local function strokes_print(t, indent)
>     indent = indent or 4
>     io.write("{\n")
>     for _,stroke in ipairs(t) do
>         stroke_print(stroke, indent)
>     end
>     io.write("}\n")
> end
> ```
> 
> (`stroke..` vs `strokes...` isn't good naming though, I know).
> 
> Prints something like image
> 
> Do you think these are interesting enough to somehow ship them with xournalpp (and if so, is there actually a way providing it that easily)?

It's useful (better formatting than a generic dump of the table like with `var_dump` from the Example/Test plugin). Not sure how to ship it currently with Xournal++ though. Maybe opening a [discussion](https://github.com/xournalpp/xournalpp/discussions) with useful Lua snippets would be an acceptable solution for now.

rolandlo:
> Are there tests for the luaApi functions? (or moreover do I have to write some?)

Currently there are no tests for the luaApi functions, unfortunately. There should be, of course. I once tried, but run into some problem, and didn't find time to make it work since then. If you would like to work on that, you're more than welcome.

atticus-sullivan:
> Seems nobody thought of it. Feel welcome to submit a PR about it

Alright, then I'll go with the current `g_warning` stuff for now to be consistent and as you said `luaL_error` is for another PR.

> If you would like to work on that [tests for luaApi stuff], you're more than welcome.

For now I just wanted to know if tests have to be written to complete this PR, but I think I'll have a look at this (but this will take more time and is part of another PR).

atticus-sullivan:
> It's useful (better formatting than a generic dump of the table like with var_dump from the Example/Test plugin). Not sure how to ship it currently with Xournal++ though.

Thinking of it one might somehow do this via metamethod `__tostring` but maybe this already is somewhat overengineered (not completely sure though how to do this from the C side but it seems to be possible). Apart from this, doing this via the discussion page seems good (it's just relevant for plugin authors after all)

atticus-sullivan:
To now only `size_and_pos` are queried on `getToolInfo("selection")`. Not sure what information else is useful (maybe the type, but I'm a bit confused about the different selection tool types other than that I see no other information contained in a selection).

src/core/plugin/luapi_application.h
```diff
@@ -813,9 +813,46 @@ static int applib_addStrokes(lua_State* L) {
...
- * TODO document the returned table and the parameters
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Is inverse to app.addStrokes
+ *
+ * Required argument: type ("Selection" or "Layer")
+ *
+ * Example:
+ *
+ * app.getStrokes() = {
+ *         {   -- Inside a stroke are three tables of equivalent length that represent a series of points
```
rolandlo, line 824: This should be expressed slightly differently, since the examples thus far stand for code that you can execute. Here the right side is just a possible result of that code.

atticus-sullivan: You're right, this notation is a bit too short. Should be fixed now

rolandlo:
> To now only `size_and_pos` are queried on `getToolInfo("selection")`. Not sure what information else is useful (maybe the type, but I'm a bit confused about the different selection tool types other than that I see no other information contained in a selection).

In a plugin I might want to know the `snappedBounds` and the `rotation` angle, maybe also the `supportRotation` flag. Currently all of these are private to `EditSelection` class though.

I don't like the name `size_and_pos`. Better call it `bounds` or maybe `boundingBox`.

atticus-sullivan:
Oh these rotation properties sound interesting (eventhough there is no rotation functionality in the luaApi currently). Tbh I don't know what `snappedBounds` is, but after all it's coming down to writing getters for the members you named.

Any interest in the type of selection tool? (I think after the selection is done, the type of tool isn't that important anymore)

I like `boundingBox` (just only thought of `size` at the beginning and then thought, the position should be included in the name as well, that's where I was coming from).

rolandlo:
> Oh these rotation properties sound interesting (eventhough there is no rotation functionality in the luaApi currently). 

In any case a plugin that gets the points of a selection should know how much they are rotated (around the center of the `snappedBounds` rectangle). When you rotate a selection the strokes in the selection don't change, only the rotation angle changes. The selection just appears rotated when rendered to the screen.

> Tbh I don't know what `snappedBounds` is, but after all it's coming down to writing getters for the members you named.
> 
The `snappedBounds` is a minimal bounding box for all points of all strokes, without accounting for line width and without any padding. It is strictly contained in the `boundingBox` you see on the screen after creating a selection and used for grid snapping and rotation snapping. Say you create a rectangle with the rectangle tool and want to have it aligned exactly to the grid of the "graph" paper. Then you want the actual points (of one or more sides) to be on exactly one of the grid lines. 

grafik

In the selection above (containing only the red triangle) I have added (an approximate) `snappedBounds` in green.

> Any interest in the type of selection tool? (I think after the selection is done, the type of tool isn't that important anymore)
> 
I currently can't think of any practical importance, so I would not include this information. It can easily be added if someone needs it badly.

atticus-sullivan:
Alright, then this should be it so far, I think.

src/core/plugin/luapi_application.h
```diff
@@ -811,6 +812,139 @@ static int applib_addStrokes(lua_State* L) {
     return 0;
 }
 
+/**
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Is inverse to app.addStrokes
+ *
+ * Required argument: type ("Selection" or "Layer")
+ *
+ * possible return value:
```
rolandlo, line 821:
```suggestion
 * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
 * Is inverse to app.addStrokes
 *
 * Required argument: type ("Selection" or "Layer")
 *
 * Example: local strokes = app.getStrokes("Selection")
 *
 * possible return value:
```

src/core/plugin/luapi_application.h
```diff
@@ -1039,6 +1173,23 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
  *
  * See /src/control/ToolEnums.cpp for possible values of "size".
  *
+ * for seiection:
```
rolandlo, line 1176: Also add an `Example 7` for selections below.

src/core/plugin/luapi_application.h
```diff
@@ -811,6 +812,139 @@ static int applib_addStrokes(lua_State* L) {
     return 0;
 }
 
+/**
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Is inverse to app.addStrokes
+ *
+ * Required argument: type ("Selection" or "Layer")
```
rolandlo, line 819: We should be consistent with respect to lower case/upper case.

Please change "Selection" and "Layer" to lower case.

src/core/plugin/luapi_application.h
```diff
@@ -1762,6 +1950,8 @@ static const luaL_Reg applib[] = {{"msgbox", applib_msgbox},
                                   {"addSplines", applib_addSplines},
                                   {"getFilePath", applib_getFilePath},
                                   {"refreshPage", applib_refreshPage},
+                                  {"getStrokes", applib_getStrokes},
+                                  // {"getWidthAndHeight", applib_getWidthAndHeight},
```
rolandlo, line 1954: Remove this line
```suggestion
```

rolandlo (changes requested):
For selections that were moved/scaled we actually also need to know the `originalBounds` from `EditSelectionContents` (as you can see from `EditSelectionContents::finalizeSelection`). Can you add that info to `applib_getToolInfo("selection")`?

For the record:
```lua  
local selInfo = app.getToolInfo("selection")
  if selInfo then
    local rect = selInfo["snappedBounds"]
    local strokes = app.getStrokes("Selection")
    -- mirror each point of each stroke at the horizontally central axis of the snappedBounds
    for i=1,#strokes do
      for j=1, #strokes[i]["x"] do
    	  strokes[i]["x"][j] = 2*rect["x"] + rect["width"] - strokes[i]["x"][j]
      end
   end
   
   -- delete old selection and add mirrored strokes back to the canvas
   app.uiAction({action = "ACTION_DELETE"})
   app.addStrokes({["strokes"] = strokes})
   app.refreshPage()
  else
   print("No selection!")
 end
```
will mirror a selection (if it exists) horizontally such that the `snappedBounds` are preserved. It only works if the selection has not been moved/scaled/rotated after it has been created.

atticus-sullivan:
> will mirror a selection (if it exists) horizontally such that the `snappedBounds` are preserved. It only works if the selection has not been moved/scaled/rotated after it has been created.

Ah and you want to calculate the offset by diffing `originalBounds` and `boundingBox`, right?

atticus-sullivan:
In the comments I often see `(multiple with zoom)`  (e.g `src/core/control/tools/EditSelection.h` L:101) what does this mean? (I see no change when I zoom in/out)
[For me/now this is mostly relevant for the comment on `getOriginalBounds()`]

src/core/plugin/luapi_application.h
```diff
@@ -1217,6 +1243,11 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
  * Example 6: local highlighterInfo = app.getToolInfo("highlighter")
  *            local sizeName = highlighterInfo["size"]["name"]
  *            local opacity = highlighterInfo["fillOpacity"]
+ *
+ * Example 7: local selectionInfo = app.getToolInfo("selection")
+ *            local rotation = selectionInfo["rotation"]
+ *            local bounding_x = selectionInfo["boundingBox"]["x"]
+ *            local snappedBounds_width = selectionInfo["snappedBounds"]["width"]
```
rolandlo, line 1250: So far we used the naming conventions from the [wiki](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#naming-conventions) for Lua as well, so please stick with it.
```suggestion
 *            local boundingX = selectionInfo["boundingBox"]["x"]
 *            local snappedBoundsWidth = selectionInfo["snappedBounds"]["width"]
```

src/core/plugin/luapi_application.h
```diff
@@ -1113,6 +1115,24 @@ static int applib_changeBackgroundPdfPageNr(lua_State* L) {
     return 1;
 }
 
+/*
+ * pushes an rectangle (width, height, x, y) to the lua stack as a new table
+ * in most cases you'll want to do a
+ * lua_newtable(L);  // create rectangle table
+ * before and a
+ * lua_setfield(L, -2, "snappedBounds");  // add rectangle table to other table
```
rolandlo, line 1123: Instead of "snappedBounds" the name of the used rectangle should be used.

Maybe factor the name into the `pushRectangleHelper` method and call `lua_newtable(L)` at the beginning and `lua_setfield(L, -2, name)` at the end of it. That would simplify things, although it could be argued that someone might want to add more info to a rectangle in future, which would make it impossible to use this helper function. What do you think?

rolandlo (commented):
Some nitpicks left. Otherwise the PR looks fine to me. 
As a follow-up one might extend `addStrokes` and `addSplines` a bit, so that one could also add the strokes/splines to a newly created selection instead of the current page layer. That would make `addStrokes` more closely inverse to `getStrokes`. I think that is out of the scope of the present PR.

src/core/plugin/luapi_application.h
atticus-sullivan: > it could be argued that someone might want to add more info to a rectangle in future, which would make it impossible to use this helper function

I've been exactly at this point when writing that function and this was the reason why I decided to really just pushing the corresponding values while noting that in most cases these two lines will surround the function-call.
So for being able to just extend a table I'd keep it this way. (That's my take on this design decision)

atticus-sullivan:
> Some nitpicks left

That's what we need to get a nice code base, so already thank you for reviewing and all your work :+1: (sorry if some things should have been detected by `clang-format`, I definitely have to make some sort of pre-commit hook so that I don't forget formatting before committing)

src/core/plugin/luapi_application.h
rolandlo: I'm fine with this decision.

rolandlo:
> > updated the test plugin
> 
> Nice, gonna be the next PR (if it's ok if I (again) "steal" your code). My aim would be to build some gui like over this with `lgi` (options to select e.g. the axis, maybe embed a scaling option for exact scaling extending to a general transformation tool. But as already written this will be another feature -> other PR)

Sure, you may use the code to write a nice plugin.

rolandlo (approved):
The PR looks good to me now. Could you please do an interactive rebase to clean up the commit history? Let me know if you need help with that.

rolandlo:
Please also update the PR title.

atticus-sullivan:
Not quite sure how to group the commits. I'd do a single commit for `getStrokes` (I see no use in splitting up there) but I'm not sure if for `getToolInfo` it should be a single commit as well or not (maybe making one commit introducing the `selection` argument and then make commits gradually introducing the information included so that one can remove an information by reverting that commit, then 5-6 commits seem much for this and the benefit isn't that big).
Any advice?

rolandlo:
> Not quite sure how to group the commits. I'd do a single commit for `getStrokes` (I see no use in splitting up there) but I'm not sure if for `getToolInfo` it should be a single commit as well or not (maybe making one commit introducing the `selection` argument and then make commits gradually introducing the information included so that one can remove an information by reverting that commit, then 5-6 commits seem much for this and the benefit isn't that big). Any advice?

Looking at it, I think that one commit for `getStrokes` and one commit for the `selection` argument is enough.

atticus-sullivan:
Alright, almost done. When reading again, I figured it would be useful to just give a short description what's the difference of the different boundingBox information we return via `getToolInfo("selection")` (just like me, not every plugin author will be into the whole codebase).
Any comments on that? (if no I'll just squash this with the `extend getToolInfo` commit and this PR should be finished I think)

atticus-sullivan:
> Anything more you want to change?

No, that's it. If you or someone else doesn't come up with something else, I'm done here.

rolandlo:
Ok then, let's go ahead. Thanks a lot for your contribution to the plugin API. It will allow for many new interesting plugins that manipulate strokes in some way. 

PR #4360, opened 2022-10-23
Change LaTeX generator

VxDxK:
I think it's very strange that we can't generate such LaTeX:
image
So my commit fix this problem:
image

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -181,8 +181,9 @@ void LatexController::triggerImageUpdate(const string& texString) {
         this->dlg.setPreviewBackgroundColor(DARK_PREVIEW_BACKGROUND);
     }
 
+    const std::string formatted = LatexGenerator::deleteEmptyStrings(texString);
     this->lastPreviewedTex = texString;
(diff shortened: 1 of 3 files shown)


tattsan:
The latter sample will be send to LaTeX as:
```
\(\displaystyle
x^2

3\)
```
and this is invalid LaTeX code. But we may want to write like:
```
  \(%
  \displaystyle
  \begin{minipage}{0.3\linewidth}
    foo
    
    bar
  \end{minipage}
  {}\)
```
and this is valid code. I do not agree with your PR because it would prohibit such LaTeX code.  All you need is to write your LaTeX code correctly.

tattsan:
To ignore blank lines in math mode, see:
https://tex.stackexchange.com/questions/276072/ignore-blank-lines-in-math-mode

Technius:
As @tattsan mentions, the following latex code is not even valid, even in display mode:
```
\(
x^2

3
\)
```

I don't think a change to the code is necessary in this case. You can create a custom latex template that places the input into an environment that concatenates the input lines together so that they show on the same line.

If you disagree, let me know and I will reopen this PR.

PR #4362, opened 2022-10-24
Fix PDF toolbox showing at the wrong position on Wayland

### LINKED ISSUE #4298: PDF toolbox not displaying at correct location
**Affects versions :**
 - OS: Linux (probably all though)
 - Desktop environment: Gnome-Wayland
 - libgtk 3.24.34
 - Version of Xournal++: master branch, `dc2af923`
 - Installation method: compiled from source

**Describe the bug**
On my laptop with a 1440p display, the PDF toolbox is displayed at an incorrect location far away from the selected text.

I've tried the following, with no effect:
* Changing DPI scale to 1
* Changing the zoom levels

On another computer with a 1080p display, the PDF toolbox is displayed close to the selected text, as expected.

**To Reproduce**
Open a PDF file and use the Select PDF Text tools to select some text.

**Expected behavior**
The toolbox should be closer to the selected text.

**Screenshots of Problem**
image

This is how it looks on my laptop. On my other computer, the toolbox is located two lines higher and one word to the left.

**Additional context**
N/A

Technius:
Fixes #4298. See the commit message for details.

### DIFF
--- a/src/core/gui/PdfFloatingToolbox.cpp
+++ b/src/core/gui/PdfFloatingToolbox.cpp
@@ -7,6 +7,7 @@
 #include <vector>     // for vector
 
 #include <glib-object.h>  // for G_CALLBACK, g_signal_connect
+#include <gtk/gtk.h>
 
 #include "control/Control.h"        // for Control
 #include "control/ToolEnums.h"      // for ToolType, TOOL_SELECT_PDF_TEXT_LI...
@@ -27,7 +28,9 @@
 #include "MainWindow.h"  // for MainWindow
 
 PdfFloatingToolbox::PdfFloatingToolbox(MainWindow* theMainWindow, GtkOverlay* overlay):
-        theMainWindow(theMainWindow), position({0, 0}) {
+        theMainWindow(theMainWindow),
+        overlay(overlay, xoj::util::GObjectSPtr<GtkOverlay>::Ref{}),
+        position({0, 0}) {
     this->floatingToolbox = theMainWindow->get("pdfFloatingToolbox");
 
     gtk_overlay_add_overlay(overlay, this->floatingToolbox);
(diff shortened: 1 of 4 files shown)


src/core/gui/PdfFloatingToolbox.cpp
```diff
@@ -27,7 +28,7 @@
 #include "MainWindow.h"  // for MainWindow
 
 PdfFloatingToolbox::PdfFloatingToolbox(MainWindow* theMainWindow, GtkOverlay* overlay):
-        theMainWindow(theMainWindow), position({0, 0}) {
+        theMainWindow(theMainWindow), overlay(overlay), position({0, 0}) {
```
bhennion, line 31: The default GObjectSPtr constructor adopts the pointed element (so the ref count is not increased). This is targetted at GObjectSPtr p(g_new_object(...)).

In this instance, I'm assuming the GtkOverlay is not supposed to be owned by the PdfFloatingToolbox, right? So I'd change this 
```suggestion
        theMainWindow(theMainWindow), overlay(overlay, xoj::util::GObjectSPtr<GtkOverlay>::ref_sink), position({0, 0}) {
```
(seems safer than just `ref` in case the ref was still floating, but it really should not be floating at this point)

src/core/gui/PdfFloatingToolbox.h
```diff
@@ -9,6 +9,8 @@
 
 #include "control/tools/PdfElemSelection.h"  // for PdfElemSelection
 #include "pdf/base/XojPdfPage.h"             // for XojPdfPageSelectionStyle
+#include "util/raii/CLibrariesSPtr.h"
+#include "util/raii/GObjectSPtr.h"  // for GObjectSPtr
```
bhennion, line 13:
```suggestion
#include "util/raii/GObjectSPtr.h"  // for GObjectSPtr
```
You don't need to include the first one.

src/core/gui/PdfFloatingToolbox.cpp
```diff
@@ -71,28 +75,27 @@ void PdfFloatingToolbox::hide() {
...
+        // Get existing width and height
         GtkRequisition natural;
         gtk_widget_get_preferred_size(widget, nullptr, &natural);
         allocation->width = natural.width;
         allocation->height = natural.height;
 
         if (self->pdfElemSelection && self->pdfElemSelection->getPageView()) {
             // Make sure the "pdfFloatingToolbox" is fully displayed.
-            int gap = 5;
+            const int gap = 5;
```
bhennion, line 86: nit'
```suggestion
            constexpr int gap = 5;
```

bhennion (commented):
Apart from the ref_sink bit and a couple of nitpicks, LGTM!

src/core/gui/PdfFloatingToolbox.cpp
Technius: Seems unnecessary.

bhennion (approved):

PR #4354, opened 2022-10-23
Add LoadHandler::fixNullPressureValues

bhennion:
Fixes #4346

This adds a filter in LoadHandler to ensure that the pressure values of every point in every stroke is positive (or Point::NO_PRESSURE if the stroke is not pressure-sensitive).
All points with a non-positive pressure value are removed, and the corresponding stroke is cut. An invalid stroke can thus become several valid ones (or none at all).

A suitable test unit is also added.

Here is a comparison of what it looks like on corrupted strokes

https://user-images.githubusercontent.com/17818041/197397178-1feb75a1-02c5-456d-b7a4-087c0a7c6ed9.mp4

There is an invisible stroke with only non-positive pressure values in the test file. It gets removed.
The long stroke is riddled with non-positive pressure values. It gets cut into several strokes.
Valid strokes are not changed.

PR #4363, opened 2022-10-24
fix crash with zero width stroke segments

mjg:
553588a4c ("Split away helper functions from view/StrokeView", 2022-07-21) split away helper functions but also changed behaviour of the code. In particular, it introduced an assert for non-zero width of stroke segments of strokes with pressure which was not there before. As a consequence, xournalpp crashes on files on which it did not cross before.

Zero width stroke segments can happen for strokes with pressure due to fading as well as rounding. So, remove the assert which was introduced by the split.


rolandlo:
Thanks for the fix. I think we will go with #4354 though.

Issue #2316, opened 2020-10-14
Building Xournal++ and GTK+ to fix Windows 10 N-Trig Input Issue


Phabbits:
Hey All!

**Question**
My N-trig Windows 10 laptop cannot detect input devices due to a GTK+ bug which has an unmerged solution. I see there are documents on how to build Xournal++ with GTK+, but I am having a lot more trouble figuring out how to build GTK+ and I see no forums or ways to contact GTK+ for help. Could anyone help me build a copy of Xournal++ with the unmerged GTK+ changes so that I can use Xournal++ on my Windows 10 laptop?

**Background**
I am using Xournal++ 1.0.18 with Windows 10 on a Sony Vaio Flip. The Sony Vaio uses N-trig Active pen technology.

The issue I am having is that my pen is not recognized as an input device, and does not work properly. I can use the eraser button on my pen, and it will show the eraser icon. However, once I have erased something and unclicked the button, I can no longer erase anything more unless I restart the program. This is with the new input system. Without the new input system my eraser button is not recognized at all. In both cases the pen or touchscreen does not show up as an input device. I dual boot with Arch Linux, and the inputs are all detected and work through Arch.

Many examples of this same issue with Windows 10 and N-trig input devices can be seen in issue #1751.
Issue #1751 references an earlier issue #1346 which goes into more detail and links it to a GTK+ bug [#262](https://gitlab.gnome.org/GNOME/gtk/issues/262).

The conclusion from these two issues seems to be that to fix the Windows 10 N-trig input issue, we should wait for GTK+ to fix its bug. The GTK+ bug [#262](https://gitlab.gnome.org/GNOME/gtk/issues/262) has been inactive for a very long time, but recently Bacci has started working on it.

I have already attempted the process, but got confused on how to build GTK+. Recently I saw that another user, Trescott from GTK+, had successfully built the change with Xournal++ 1.0.18, and pointed out some issues that Baccai had fixed. This makes me think that if I were to build Xournal++ with it now, my input issues would be resolved.

Perhaps I should post this on the GTK+ merge request as well, but I need to make an account for it first.

Thanks for the help!

kubajaburek:
Hello everyone!

I compiled GTK and Xournal++ with the [WIP merge request to GTK that is supposed to add support for Windows Pointer APIs](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1563). Here’s what I got.

[...]

PR #4376, opened 2022-10-31
Fix assertion on selection existence when the stylus tip touches the screen

### LINKED ISSUE #4375: Xournal++ crash when "Select region" and tip touches screen.
When trying to select a region with the stylus while holding **Button 2 (Assigned "Select region")** and making the stylus tip touch the screen, Xournal++ crashes.

It does not crash if holding **Button 2** and selecting a region without touching the screen with the stylus.

I have Xournal++ 1.1.2+dev (Octoboer 29th 2022 release)

rolandlo:
Fixes #4375

Here is what happens in `Merging button events with stylus tip events` mode, when the region selection tool is assigned to a stylus button:

- While hovering, press the stylus button -> a button press event is received and a region selection is created (which is also visible on the screen)
- With the tip of the stylus touch the screen -> a button press event is received again and it is falsely asserted that there is no (region) selection => the app crashes.

So the assertion is wrong in `Merging button events with stylus tip events` mode, which explains the fix.

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -382,15 +382,25 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
                h->getToolType() == TOOL_PLAY_OBJECT || h->getToolType() == TOOL_SELECT_OBJECT ||
                h->getToolType() == TOOL_SELECT_PDF_TEXT_LINEAR || h->getToolType() == TOOL_SELECT_PDF_TEXT_RECT) {
         if (h->getToolType() == TOOL_SELECT_RECT) {
-            assert(!selection);
-            this->selection = std::make_unique<RectSelection>(x, y);
-            this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
-                    this->selection.get(), this, this->settings->getSelectionColor()));
+            if (!selection) {
+                this->selection = std::make_unique<RectSelection>(x, y);
+                this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
+                        this->selection.get(), this, this->settings->getSelectionColor()));
+            } else {
+                assert(settings->getInputSystemTPCButtonEnabled() &&
+                       "the selection has already been created by a stylus button press while the stylus was "
+                       "hovering!");
+            }
         } else if (h->getToolType() == TOOL_SELECT_REGION) {
-            assert(!selection);
-            this->selection = std::make_unique<RegionSelect>(x, y);
-            this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
-                    this->selection.get(), this, this->settings->getSelectionColor()));
+            if (!selection) {
+                this->selection = std::make_unique<RegionSelect>(x, y);
+                this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
+                        this->selection.get(), this, this->settings->getSelectionColor()));
+            } else {
+                assert(settings->getInputSystemTPCButtonEnabled() &&
+                       "the selection has already been created by a stylus button press while the stylus was "
+                       "hovering!");
+            }
         } else if (h->getToolType() == TOOL_SELECT_PDF_TEXT_LINEAR || h->getToolType() == TOOL_SELECT_PDF_TEXT_RECT) {
             // so if we selected something && the pdf selection toolbox is hidden && we hit within the selection
             // we could call the pdf floating toolbox again


bhennion:
It's a nice fix, but I think the real bug lies in the input system: when merging input events, we should not call PageView::OnButtonPressEvent twice (one for the button being pressed, one for the tip touching the tablet).
There may be other symptoms of this weird behaviour (typically a memory leak, but also possibly a segfault).

I'd simply add a `bool actionStarted` flag set (and unset) in `StylusInputHandler::handleImpl`, to avoid calling twice `actionStart()`.

rolandlo:
What to do with twice calling `actionEnd()` then? When `actionStart()` is called twice (and only executed the first time), would you wait for `actionEnd()` to be only executed the second time (when the stylus button is released). Currently the behavior is weird indeed. The action is started when the stylus button is pressed while hovering. It is ended when the stylus tip leaves the screen.

We must keep in mind that the `Merging button events with stylus tip events` mode was only designed for styluses that emit the "short sequences" as described in https://github.com/xournalpp/xournalpp/issues/2316#issuecomment-860083264 with regards to stylus button presses (i.e. when you press the button while hovering, no button press event is emitted. You only get the button press event when the stylus tip hits the screen; there is only one button release event when you leave the screen and none when releasing the stylus button).  The behavior is OS-dependent, on Windows all styluses behave like that, see https://github.com/xournalpp/xournalpp/issues/2316#issuecomment-867621727 
For those styluses there is currently no issue with regards to the assertion. There is only an issue when the stylus emits the "long sequences" and `Merging button events with stylus tip events`  is enabled.  We should consider all 2x2 combinations of input sequences (long/short) and `Merging button events with stylus tip events` (true/false), as we can't rely on users to set the checkbox according to their stylus.

bhennion:
I'd say we treat actionEnd similarly: if the flags actionStarted is false, don't trigger an actionEnd. The goal is that downstream, we are sure that each actionStart is matched with an actionEnd.

Also, this `Merging button events with stylus tip events` has been the subject of a lot of many bug reports. It'd be nice if we could figure out whether or not it's needed automatically. Something like WINDOWS || (X11 && TabletPCButton) (and I don't know for MacOs). We could always leave the setting accessible somewhere hidden (in an expert menu or so).

rolandlo:
Maybe we should still merge the present fix until we come up with a more substantial solution, that is tested properly.
The input system is really complex with regards to styluses and stylus buttons, a hot potato to touch. So many different setups have to be considered.

src/core/gui/PageView.cpp
```diff
@@ -382,15 +382,23 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
-            this->selection = std::make_unique<RectSelection>(x, y);
-            this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
-                    this->selection.get(), this, this->settings->getSelectionColor()));
+            if (!selection) {
+                this->selection = std::make_unique<RectSelection>(x, y);
+                this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
+                        this->selection.get(), this, this->settings->getSelectionColor()));
+            } else {
+                assert(settings->getInputSystemTPCButtonEnabled());
+                // the selection has already been created by a stylus button press while the stylus was hovering
```
Technius, line 391: nit: the comment should be folded into the assertion message to make debugging easier.
```suggestion
                assert(settings->getInputSystemTPCButtonEnabled() &&
                       "the selection has already been created by a stylus button press while the stylus was hovering!");
```

src/core/gui/PageView.cpp
```diff
@@ -382,15 +382,23 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
-            this->selection = std::make_unique<RegionSelect>(x, y);
-            this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
-                    this->selection.get(), this, this->settings->getSelectionColor()));
+            if (!selection) {
+                this->selection = std::make_unique<RegionSelect>(x, y);
+                this->overlayViews.emplace_back(std::make_unique<xoj::view::SelectionView>(
+                        this->selection.get(), this, this->settings->getSelectionColor()));
+            } else {
+                assert(settings->getInputSystemTPCButtonEnabled());
+                // the selection has already been created by a stylus button press while the stylus was hovering
```
Technius, line 400: nit: same here

src/core/gui/PageView.cpp
rolandlo: Thanks, I applied your suggestion.

bhennion (approved):
Go ahead. We should open an issue with this double events business so that we don't forget.

rolandlo:
Ok, I will merge and open an issue.

PR #4378, opened 2022-11-01
Refactor hardcoded colors

### LINKED ISSUE #3625: Refactor hardcoded Colors into one single include/source file.
The change looks good.

I'm just wondering if this should be taken from the default palette :thinking: in a followup PR.
Do you know for what these Colors are used in the application?

_Originally posted by @idotobi in https://github.com/xournalpp/xournalpp/pull/3623#discussion_r757884747_

RuleOfThrees:
This change introduces a namespace Colors, defined in util/Color.h, where Colors that are used in the code, and were previously constructed via unsigned literals, are constant variables. The names are either the appropriate name from the CSS4 named colors, or xopp_{name}, where name is the closest CSS4 named color (regarding distance of their (r, g, b) vectors. Places in the code where Color instances were constructed on-the-spot have been replaced with the respective constants.

Fixes #3625


rolandlo:
One question: How did you determine the closest CSS4 named color? Did you use a tool for that? Is the distance used for comparing two colors the usual Euclidean distance?

RuleOfThrees:
@rolandlo yes, I used the usual euclidean distance. I made a python script that uses [matplotlib.colors](https://matplotlib.org/stable/api/colors_api.html#module-matplotlib.colors) to get the css4 colors, and rank by distance to a color hexcode given as command line argument. I can clean up the code a bit and upload it here (or anywhere is best), if that's desirable!

rolandlo:
> @rolandlo yes, I used the usual euclidean distance. I made a python script that uses [matplotlib.colors](https://matplotlib.org/stable/api/colors_api.html#module-matplotlib.colors) to get the css4 colors, and rank by distance to a color hexcode given as command line argument. I can clean up the code a bit and upload it here (or anywhere is best), if that's desirable!

Yes, that is desirable. We may have to add new colors in the future. Also for checking the results it would be good to have the script.

RuleOfThrees:
Ok! I'll leave it here within the next days!

RuleOfThrees:
here's the script:
```python
#! /bin/python3

from functools import partial
from math import sqrt
from sys import argv

from matplotlib.colors import get_named_colors_mapping

no_colors_to_display = 20
usage = f"""{argv[0]} [ HEX_CODE_RGB ]
    shows the {no_colors_to_display} css4 colors that are closest
    to the argument by euclidean distance.
    """

def hex_to_rgb(hex_str):
    hexval = int(hex_str, 16)
    return (hexval >> 16, (hexval >> 8) & 0xff, hexval & 0xff)

css4_colors = map(lambda t: (t[0], hex_to_rgb(t[1][1:])),
                filter(lambda t: len(t[0]) > 1 and not t[0].startswith("xkcd"),
                get_named_colors_mapping().items()))

def eucl_dst(v1, v2):
    return sqrt(sum([(v1_i - v2_i) ** 2 for (v1_i, v2_i) in zip(v1, v2)]))



if __name__  == '__main__':
    if (len(argv) != 2 or argv[1] in ('-h', '--help')):
        print(usage)
        exit(0)

    point = None
    try:
        point = hex_to_rgb(argv[1])
    except Exception:
        print(f"Invalid hex code\n{usage}")
        exit(1)
    dist_from_point = partial(eucl_dst, point)
    print("Point: {}".format(point))
    for key, val in \
            sorted(css4_colors,
                key = lambda t: dist_from_point(t[1]))[:no_colors_to_display]:
        print("dist = {} color: {} code: {}".format(
            round(dist_from_point(val), 2),
            key,
            val))
```
The filtering from matplotlib's colors is because it also contains colors generated from a survey by xkcd (the webcomic) and some other values denoted by single-letter keys.

rolandlo:
The script looks good and works fine. However in the code there appear to be many more hardcoded colors. Just search for the string "0x" in all code files (there are some false positive, but most are hardcoded colors). Did you intentionally leave them out?

RuleOfThrees:
You are correct, the regular expression I was using was too restrictive and I missed a bunch of them. I added a commit that addresses this. There are some that I omitted intentionally, though:

- Colors initialized with opacity, believing it is a little too much to add a constant for each instance. There is currently only two of them in the code, though, so it may be better to change them, too.
- Colors that are used in unit tests. As unit tests also address consistency in the codebase, it may be preferable to have the test fail if a constant's definition has been changed unintentionally.

rolandlo:
> You are correct, the regular expression I was using was too restrictive and I missed a bunch of them. I added a commit that addresses this. There are some that I omitted intentionally, though:
> 
>     * Colors initialized with opacity, believing it is a little too much to add a constant for each instance. There is currently only two of them in the code, though, so it may be better to change them, too.
> 
>     * Colors that are used in unit tests. As unit tests also address consistency in the codebase, it may be preferable to have the test fail if a constant's definition has been changed unintentionally.

Looks like you caught all relevant colors now. 

We have a folder `scripts` with different (python- and bash-)scripts (only 2 so far) in the root folder of the repository. Would you mind committing your script there, with a license notice on the top of it (if you wish), that is also listed in `copyright.txt` in the root folder?

RuleOfThrees:
Done. Assumed that if I don't explicitly add a license, then it's distributed under GPLv2+ like the rest of the project? That is my intention, anyway!

rolandlo (approved):
LGTM. I agree that not putting a license info explicitly means the script is covered by the project license (GPLv2+)

rolandlo:
Thanks for your very welcome work @RuleOfThrees . I just merged it.

PR #4271, opened 2022-10-02
Split gui/TextEditor into view and controller

bhennion:
This PR splits gui/TextEditor into control/tools/TextEditor and view/overlays/TextEditionView.

It also:

1. Moves gui/TextEditor to control/tools
2. Moves (most of) the content of PageView::endText() and PageView::startText() to TextEditor's destructor and constructor

PR #4382, opened 2022-11-04
Fix raii/CLibrariesSPtr::reset(T*, Adopt) not calling adoption routine

bhennion:
All is in the title. I also added a test unit to make sure.

This commit is extracted from #4271. I feel it should not wait for this larger PR to get merged, as those RAII wrappers are used more and more in the code.


src/util/include/util/raii/CLibrariesSPtr.h
```diff
@@ -67,23 +68,21 @@ class CLibrariesSPtr {
     constexpr static struct RefSink {
     } refsink = RefSink();
 
-    CLibrariesSPtr(T* p, Adopt = adopt): p(H::adopt(p)) {}
-    CLibrariesSPtr(T* p, Ref): p(safeRef(p)) {}
-    CLibrariesSPtr(T* p, RefSink): p(safeSinkRef(p)) {}
+    explicit CLibrariesSPtr(T* p, Adopt = adopt): p(safeAdopt(p)) {}
```
Technius, line 71: We should make adopt explicit here. It's too easy to forget that the default is adopt instead of ref.

test/unit_tests/util/RAIIWrappersTest.cpp
```diff
@@ -286,3 +286,36 @@ TEST(UtilsRAII, testGObjectSPtr) {
...
+        t1.reset(t);  // Adopt a floating ref
+        EXPECT_EQ(t1.get(), t);
+        EXPECT_FALSE(g_object_is_floating(t1.get()));
+    }
+    EXPECT_EQ(TestObjectCount, 0);
+    {
+        TestObject* t = TEST_OBJECT(g_object_new(TEST_OBJECT_TYPE, nullptr));
+        EXPECT_EQ(TestObjectCount, 1);
+        TestPtr t1;
+        t1.reset(t, TestPtr::ref);
```
Technius, line 316: Should assert a count of two here.

bhennion: I don't know how to access a GObject's ref-count. Is there a way I missed?

bhennion:
Ok, I made adoption explicit for both reset() and contructor. I also moved the placeholders declaration to xoj::raii (e.g. xoj::raii::adopt instead of xoj::raii::CLibrariesSPtr<T>::adopt) to make things more readable.

Technius (approved):
LGTM. I reran the test and it passes now.

bhennion:
Thanks @Technius! Merging in 24 hours unless an objection is raised.

PR #4395, opened 2022-11-08
changes button id for MoveSelectionToLayer to fit menu

forgottosave:
Fixes the id of both buttons for the MoveSelectionToLayer to fit the menu "Edit" (moved there from "Journal"), which got introduced in #4352

### DIFF
--- a/ui/main.glade
+++ b/ui/main.glade
@@ -473,8 +473,8 @@
                           </object>
                         </child>
                         <child>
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
Thanks for the PR. Merging in 24 hours, if no objections are raised.

rolandlo:
It's no regression. The wrong id was introduced in the PR that added the menu item. I'm removing the "regression" label therefore.

PR #4409, opened 2022-11-13
Override std::exception::what rather than hiding virtual overload

alex1701c:
This fixes lots of warnings about this method declaration hiding a virtual overload.

```log
In file included from /home/user/projects/xournalpp/src/util/serializing/ObjectInputStream.cpp:9:
/home/user/projects/xournalpp/src/util/include/util/serializing/InputStreamException.h:24:25: warning: 'InputStreamException::what' hides overloaded virtual function [-Woverloaded-virtual]
    virtual const char* what();
                        ^
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/exception.h:76:5: note: hidden overloaded virtual function 'std::exception::what' declared here: different qualifiers ('const' vs unqualified)
    what() const _GLIBCXX_TXN_SAFE_DYN _GLIBCXX_NOTHROW;
```

### DIFF
--- a/src/util/include/util/serializing/InputStreamException.h
+++ b/src/util/include/util/serializing/InputStreamException.h
@@ -21,7 +21,7 @@ public:
     ~InputStreamException() override;
 
 public:
-    virtual const char* what();
+    const char* what() const noexcept override;
 
 private:
     std::string message;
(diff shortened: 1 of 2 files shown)


Technius (approved):
LGTM, thanks for the PR. Merging in 24 hours if there are no objections.

alex1701c:
@Technius Since no objections were raised, please merge this :)

Issue #4402, opened 2022-11-11
Selecting objects through multiple layers


forgottosave:
### Is your feature request related to a problem? Please describe.
Working with (multiple) layers still is a little bit of a hassle. A feature that would greatly improve this is to make the selection tool be able to select object through multiple layers.
A great working example would be GIMP, where all layers can be moved, by always choosing the top visible object. 

### Describe the solution you'd like
Similar to GIMP you would have an option to select objects only on the current layer, or over multiple layers.
 
![image](https://user-images.githubusercontent.com/88790311/201289922-e14581dd-4906-4e8d-84dd-fd6faafa2843.png)

When the `select on active layer` (in GIMP `Move the active layer`) is chosen the behavior stays the same as it is now.
Selecting objects over multiple layers (`select on all layers`) would then work the following way:
1. check all layers (top down) for objects and stop when one is found -> the top most objects will be selected
2. only allow further selections on that layer
3. *temporarily* switch to that layer until the selection is canceled
4. **now the user can move, delete, etc. this selection**
5. when the selection is cancelled (also when the tool changes) switch back to the previously selected layer

This feature wouldn't change the behavior for people not working with layers and therefor not bring any confusion, even if `select on all layers` is active.

### Describe alternatives you've considered
Alternatively I thought of making an extra layer called `any`, or `all layers`. When selecting this layer it works like the `select on all layers` described above.
But I don't think any other program does it this way and therefor it wouldn't be very intuitive.

forgottosave:
I could also immediately start working on this feature, if it's approved

Rmano:
+1 for this enhancement. The best option for me would be a new radio button in the layer menu that selects if the current layer is copied or not by a ctrl-v:

![image](https://user-images.githubusercontent.com/6414907/206701350-4d66268e-e2f3-4b45-b0b8-6e7b395b372c.png)

but even the choice about copying all or just the current one would be very welcome.

forgottosave:
@Rmano  I assume you mean with ctrl-c ;)
But don't you think it would be the most intuitive to select from either all or only the current layer?

If you want to it would be nice if you could test/review my version in PR #4413 and check if this would be a feasible implementation or give some comments there on what to change. Maybe the PR gets more attention then and will be commented on by the maintainers as well? :)

kassbohm:
Any update on this?

kassbohm:
ok, thanks.

forgottosave:
[...]

PR #4413, opened 2022-11-14
Selecting objects on non-selected layers

forgottosave:
## Issue
Feature Request:
Add feature to select objects on different layers than the active one #4402 

## Feature description
Adds two new tools, `Select Multi-Layer Rectangle` and `Select Multi-Layer Region`. Both tools are similar to the respective non-multi-layer tools, with the key-difference, that objects are selected from the top most visible layer, in which at least one object intersects the selection region/rectangle (independent of which layer is currently selected).

This feature can be found in `Tools > Select Multi-Layer Rectangle / Region`, as well as in the menu bar in the selection tool drop-down. Better name suggestions are welcome :)

#### main functionality
-  `RectSelection` and `RegionSelection` now both have an extra bool value stored, to switch between the "normal" mode and the "all layer" mode
- The `finalize()` method in both `RectSelection` and `RegionSelection` have been changed:
  1) they now check if all layers should be checked, or only the active one and set the selection accordingly
  2) their return value has changed to `size_t`, to be able to return the selected layer id
- Enum types were added for the two new actions:
  1) 2 tools: `TOOL_SELECT_MULTILAYER_RECT` & `TOOL_SELECT_MULTILAYER_REGION`
  2) and their actions: `ACTION_TOOL_SELECT_MULTILAYER_RECT` & `ACTION_TOOL_SELECT_MULTILAYER_REGION`
#### further changes
- Both tools got added to the menu `Tools`
- as well as to the tool bar
  1) 2 new entries in the selection drop down
  2) 2 new icons for those entries


forgottosave:
As described in the [issue](https://github.com/xournalpp/xournalpp/issues/4402) another useful implementation idea would be scraping the current selection tools completely and always select the top most object.
Some feedback would be appreciated on this feature implementation :)

forgottosave:
Last force-push brought the PR to a new state (without merge conflicts) again

rolandlo (commented):
Thanks for the valuable contribution. Sorry for the long waiting. Let's get started with reviewing now. I have tried out the PR (and skimmed through the code). The new selection modes look useful, even more useful than the existing ones (in most cases, I think). 
The description of the PR is misleading. Not just the top-most object is selected, but all the objects from the same visible layer, that is top-most among all visible layers intersecting the selection region/rectangle.
Before reading the PR description (or the code) I actually thought that Multi-layer selection means selecting all objects from visible layers that are in the selection. It's not like that, which is slightly confusing. The icons look nice, but may also contribute to that confusion.

I will give more feedback after studying the code. Meanwhile I hope someone (developer or user) who uses layers a lot gives some feedback as well.

forgottosave:
Thank you very much for the feedback and input

**description, feature name & icons**
I try to clarify this PR a little more in the PR-description, but am completely open for new names for this feature. Also the icons can of course be changed :)

**feature (selecting multiple objects)**
Also the feature could be changed so that one can select *multiple objects* from *multiple different layers*, but while testing I found it more useful to actually commit to one layer at a time when selecting objects.
If this should be changed it would require to also change the Selection, so that selections remember their original layer and get inserted there after the selection is cancelled. This would require some work, but I could probably implement it, if desired.

src/core/control/tools/Selection.cpp
```diff
@@ -17,29 +17,51 @@ Selection::~Selection() = default;
...
+            Layer* l = page->getLayers()->at(layerNo);
+            bool selectionOnLayer = false;
+            if (l->isVisible()) {
+                for (Element* e: l->getElements()) {
+                    if (e->isInSelection(this)) {
+                        this->selectedElements.push_back(e);
+                        selectionOnLayer = true;
+                    }
+                }
+            }
```
rolandlo, line 46: For readability I would prefer less nesting.
```suggestion
            if (!l->isVisible()) {
                continue;
            }
            bool selectionOnLayer = false;
            for (Element* e: l->getElements()) {
                if (e->isInSelection(this)) {
                    this->selectedElements.push_back(e);
                    selectionOnLayer = true;
                }
            }
```

src/core/control/tools/Selection.cpp
```diff
@@ -153,19 +178,40 @@ auto RegionSelect::contains(double x, double y) const -> bool {
...
+                layerId = page->getSelectedLayerId();
+            }
         }
     }
 
     this->viewPool->dispatchAndClear(xoj::view::SelectionView::DELETE_VIEWS_REQUEST, this->bbox);
 
-    return !this->selectedElements.empty();
+    return layerId;
 }
```
rolandlo, line 215: This whole method is a duplication of `RectSelection::finalize`. I suggest to make from these duplicated methods `Selection::finalize` and thus get rid of the duplication.

src/core/control/tools/Selection.cpp
```diff
@@ -17,29 +17,51 @@ Selection::~Selection() = default;
 
 //////////////////////////////////////////////////////////
 
-RectSelection::RectSelection(double x, double y) {
+RectSelection::RectSelection(double x, double y, bool multiLayer) {
```
rolandlo, line 20: Please use initializer lists in the constructors. Moreover `multiLayer` should be passed to the constructor of `Selection`, I think.

rolandlo (changes requested):
The feature looks good to me and the new icons express the purpose better than the old ones. There are some comments and corrections though, which I ask you to take into account.

Also please add `SELECT_MULTILAYER_RECTANGLE` and `SELECT_MULTILAYER_REGION` to the `Right hand Note Taking` toolbar in `toolbar.ini` and also list those tools in the comment section of that file.

I have also noticed that selecting an object via a single click only works for objects on the current layer even using the multilayer selection tools. Can that be changed?

src/core/control/ToolEnums.h
```diff
@@ -61,20 +61,22 @@ enum ToolType {
...
-    TOOL_DRAW_ELLIPSE = 13,
-    TOOL_DRAW_ARROW = 14,
-    TOOL_DRAW_DOUBLE_ARROW = 15,
-    TOOL_DRAW_COORDINATE_SYSTEM = 16,
-    TOOL_FLOATING_TOOLBOX = 17,
-    TOOL_DRAW_SPLINE = 18,
-    TOOL_SELECT_PDF_TEXT_LINEAR = 19,
-    TOOL_SELECT_PDF_TEXT_RECT = 20,
+    TOOL_SELECT_MULTILAYER_RECT = 7,
+    TOOL_SELECT_REGION = 8,
```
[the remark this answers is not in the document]
forgottosave: Sure thing. I should fix all these occasions in my next commit.

src/core/control/tools/Selection.cpp
forgottosave: Oh, yes. This is something I wanted to ask you about, but forgot to write into the PR-description. I also think this duplicate code should be merged into one method. I'll do that.

forgottosave:
I'll look at the remaining change requests (duplicate methods, toolbar.ini addition,...) in the next 1-2 days. The one-click-selection is something that should be possible and seems important for this feature, I'll also add that.

forgottosave:
Your change requests are now implemented. I also got the single-click selection to work and the additions to the `toolbar.ini` have been made.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -492,8 +492,9 @@ void ToolMenuHandler::initToolItems() {
     addCustomItemTgl("DRAW_SPLINE", ACTION_TOOL_DRAW_SPLINE, GROUP_RULER, false, "draw-spline", _("Draw Spline"));
 
     addCustomItemTgl("SELECT_REGION", ACTION_TOOL_SELECT_REGION, GROUP_TOOL, true, "select-lasso", _("Select Region"));
-    addCustomItemTgl("SELECT_RECTANGLE", ACTION_TOOL_SELECT_RECT, GROUP_TOOL, true, "select-rect",
-                     _("Select Rectangle"));
+    addCustomItemTgl("SELECT_RECTANGLE", ACTION_TOOL_SELECT_RECT, GROUP_TOOL, true, "select-rect", _("Select Rectangle"));
+    addCustomItemTgl("SELECT_MULITLAYER_REGION", ACTION_TOOL_SELECT_MULTILAYER_REGION, GROUP_TOOL, true, "select-lasso", _("Select Multi-Layer Region"));
+    addCustomItemTgl("SELECT_MULITLAYER_RECT", ACTION_TOOL_SELECT_MULTILAYER_RECT, GROUP_TOOL, true, "select-lasso", _("Select Multi-Layer Rect"));
```
rolandlo, line 497: There are still some mistakes:
```suggestion
    addCustomItemTgl("SELECT_RECTANGLE", ACTION_TOOL_SELECT_RECT, GROUP_TOOL, true, "select-rect", _("Select Rectangle"));
    addCustomItemTgl("SELECT_MULTILAYER_REGION", ACTION_TOOL_SELECT_MULTILAYER_REGION, GROUP_TOOL, true, "select-multilayer-lasso", _("Select Multi-Layer Region"));
    addCustomItemTgl("SELECT_MULTILAYER_RECT", ACTION_TOOL_SELECT_MULTILAYER_RECT, GROUP_TOOL, true, "select-multilayer-rect", _("Select Multi-Layer Rect"));
```

rolandlo (changes requested):
Only some small changes left. The rest (including one-click selection) looks good.

forgottosave:
Fixed the missing changes. Should I open a PR synchronizing the `TOOLBAR_INI_HEADER` with the `toolbar.ini` comment?

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
forgottosave: Should I add the `SELECT_MULTILAYER_RECTANGLE`/`REGION` in this PR? Or should I maybe open a PR synchronizing the `TOOLBAR_INI_HEADER` with the toolbar.ini comment all together  (including the `SELECT_OBJECT`, `PLAY_OBJECT`,...)?

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -492,8 +492,9 @@ void ToolMenuHandler::initToolItems() {
     addCustomItemTgl("DRAW_SPLINE", ACTION_TOOL_DRAW_SPLINE, GROUP_RULER, false, "draw-spline", _("Draw Spline"));
 
     addCustomItemTgl("SELECT_REGION", ACTION_TOOL_SELECT_REGION, GROUP_TOOL, true, "select-lasso", _("Select Region"));
-    addCustomItemTgl("SELECT_RECTANGLE", ACTION_TOOL_SELECT_RECT, GROUP_TOOL, true, "select-rect",
-                     _("Select Rectangle"));
+    addCustomItemTgl("SELECT_RECTANGLE", ACTION_TOOL_SELECT_RECT, GROUP_TOOL, true, "select-rect", _("Select Rectangle"));
+    addCustomItemTgl("SELECT_MULITLAYER_REGION", ACTION_TOOL_SELECT_MULTILAYER_REGION, GROUP_TOOL, true, "select-multilayer-lasso", _("Select Multi-Layer Region"));
```
rolandlo, line 496: Still wrong. MULIT -> MULTI
```suggestion
    addCustomItemTgl("SELECT_MULTILAYER_REGION", ACTION_TOOL_SELECT_MULTILAYER_REGION, GROUP_TOOL, true, "select-multilayer-lasso", _("Select Multi-Layer Region"));
```

rolandlo (changes requested):

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
forgottosave: Omg, I just now saw what's wrong in this line, sorry...

rolandlo:
> Fixed the missing changes. Should I open a PR synchronizing the `TOOLBAR_INI_HEADER` with the `toolbar.ini` comment?

Yes, that would be appreciated. Both files need to be checked carefully if something is missing.

rolandlo (commented):
Now everything works fine, it seems. Can you please clean up the commit history. There should be around 3-5 commits:
- at least one for the feature
- one for the single-click selection
- one for adding it to the Right hand note taking toolbar

forgottosave:
Managed to accidentally change stuff in the `main.glade` in the first force-push, fixed it in the second one.

rolandlo (approved):
LGTM. 

rolandlo:
Thanks @forgottosave for the PR once more. I just merged it. That will greatly help users who work with multiple layers.

forgottosave:
Thank you very much for taking the time for reviewing this. Amazing work by all of you maintaining this project :)

Issue #2007, opened 2020-05-23
Using numbers as hotkey for colors


Fatih20:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when wanting to quickly use another color I have to reach to the top panel, write the thing I need, and then switch back to black by reaching the top bar again.

**Describe the solution you'd like**
Use the number row as a hotkey for the color palette. Similar to how the number row is how you switch between weapon in most FPS games.

rolandlo:
You can use a plugin to do that. See the [plugin for cycling through a color list](#2251) and the website [section on plugins](https://xournalpp.github.io/guide/plugins/) and some more advice in issue #919. Note that you can register several shortcuts in the same plugin using multiple `registerUI` commands.

fpesari:
I agree that this functionality would be very useful to have without having to use an external plugin. 

Right now, when the highlighter tool is selected, numbers aren't bound to any function that I know of, they are a no-op. So adding support for selecting the colors in the toolbar by pressing the relative number key (and maybe binding ` to toggle between the last selected colors) would only be beneficial, IMHO.

PR #4416, opened 2022-11-14
Implemented color switching by number hotkeys

fpesari:
Implements #2007, that is, switching colors by using the 0-9 number keys.

I tested it while using various tools and it works. It has no conflicts with the text tool (so, if you are writing inside a textbox the number `1`, the first color won't be selected). 

(Note that I could have used metaprogramming to shorten the code _a lot_ but I decided not to in the name of clarity, so right now there are a lot of `if` statements, as done in other parts of the codebase)

### DIFF
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -26,6 +26,8 @@
 #include "gui/PdfFloatingToolbox.h"             // for PdfFloatingToolbox
 #include "gui/inputdevices/HandRecognition.h"   // for HandRecognition
 #include "gui/inputdevices/InputContext.h"      // for InputContext
+#include "gui/toolbarMenubar/ColorToolItem.h"   // for ColorToolItem
+#include "gui/toolbarMenubar/ToolMenuHandler.h" // for ToolMenuHandler
 #include "gui/widgets/XournalWidget.h"          // for gtk_xournal_get_layout
 #include "model/Document.h"                     // for Document
 #include "model/Element.h"                      // for Element, ELEMENT_STROKE
@@ -287,6 +289,20 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         return true;
     }
 
+    // Switch color on number key
+    auto &colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
+    if ((event->keyval >= GDK_KEY_0)
+        && (event->keyval < GDK_KEY_0 + std::min((std::size_t)10,
+                                                 colors.size()))) {
+            std::size_t index = std::min(colors.size() - 1,
+                                         (std::size_t)(9 + (event->keyval - GDK_KEY_0)) % 10);
+            auto colorToolItem = colors.at(index);
+            if (colorToolItem->isEnabled()) {
+                gtk_toggle_tool_button_set_active(
+                    GTK_TOGGLE_TOOL_BUTTON(colorToolItem->getItem()), true);
+            }
+            return true;
+    }
     return false;
 }
 
--- a/src/core/gui/toolbarMenubar/AbstractToolItem.h
+++ b/src/core/gui/toolbarMenubar/AbstractToolItem.h
@@ -67,6 +67,8 @@ public:
      */
     void enable(bool enabled) override;
 
+    GtkToolItem* getItem() const;
+
 protected:
     virtual GtkToolItem* newItem() = 0;
 
--- a/src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
+++ b/src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
@@ -610,6 +610,8 @@ auto ToolMenuHandler::isColorInUse(Color color) -> bool {
 
 auto ToolMenuHandler::getToolItems() -> std::vector<AbstractToolItem*>* { return &this->toolItems; }
 
+auto ToolMenuHandler::getColorToolItems() const -> const std::vector<ColorToolItem*>& { return this->toolbarColorItems; }
+
 void ToolMenuHandler::disableAudioPlaybackButtons() {
     setAudioPlaybackPaused(false);
 
--- a/src/core/gui/toolbarMenubar/AbstractToolItem.cpp
+++ b/src/core/gui/toolbarMenubar/AbstractToolItem.cpp
@@ -20,6 +20,10 @@ AbstractToolItem::~AbstractToolItem() {
     }
 }
 
+auto AbstractToolItem::getItem() const -> GtkToolItem* {
+    return this->item;
+}
+
 void AbstractToolItem::selected(ActionGroup group, ActionType action) {
     if (this->item == nullptr) {
         return;
--- a/src/core/gui/toolbarMenubar/ToolMenuHandler.h
+++ b/src/core/gui/toolbarMenubar/ToolMenuHandler.h
@@ -86,6 +86,7 @@ public:
     ToolbarModel* getModel();
 
     std::vector<AbstractToolItem*>* getToolItems();
+    const std::vector<ColorToolItem*>& getColorToolItems() const;
 
     Control* getControl();
 


rolandlo (changes requested):
Thanks for your contribution. I support the idea.

I would definitely prefer less code duplication here. Moreover take care that the `colorToolItem` exists. There is no guarantee that there are 10 colors in the toolbar.

fpesari:
> Thanks for your contribution. I support the idea.
> 
> I would definitely prefer less code duplication here. Moreover take care that the `colorToolItem` exists. There is no guarantee that there are 10 colors in the toolbar.

1. I will take care of checking if `colorToolItem` exists
2. I will take into consideration the fact that colors in the toolbar can be fewer than 10 
3. As for code duplication, I intentionally duplicated code to avoid metaprogramming. The shortest way to do it would have been a C preprocessor macro. If you authorize me to do it, I will do it. I did not do it straight away because I know macros are discouraged in some codebases.

rolandlo:
>     3. As for code duplication, I intentionally duplicated code to avoid metaprogramming. The shortest way to do it would have been a C preprocessor macro. If you authorize me to do it, I will do it. I did not do it straight away because I know macros are discouraged in some codebases.

I don't see why you would need a C preprocessor macro. Can't you use a simple `for`-loop?

fpesari:
> I don't see why you would need a C preprocessor macro. Can't you use a simple `for`-loop?

I think It's not possible because the `GDK_Key_xxx` (where `xxx` is the key) are identifiers, while the color indices are numbers.

Preprocessor macros would allow me to do it using the `##` preprocessor operator. Here is an example:

```C
#include <stdio.h>

#define ID_0 "First id"
#define ID_1 "Second id"

#define IDX(x) ID_##x

int main(int, char*) {
    printf("%s\n", IDX(0));
    printf("%s\n", IDX(1));
    return 0;
}
```

Using that inside a for loop, well, it could be possible with further preprocessor magic (or maybe using a function as well) but it would not be immediate (because C does not replace identifiers with their value even if they are defined as `const`).

In short, if you allow me to use macros, I can make that section with a lot of `if`s look like this:

```
BIND_NUMKEY_TO_COLOR(1);
BIND_NUMKEY_TO_COLOR(2);
BIND_NUMKEY_TO_COLOR(3);
BIND_NUMKEY_TO_COLOR(4);
BIND_NUMKEY_TO_COLOR(5);
BIND_NUMKEY_TO_COLOR(6);
BIND_NUMKEY_TO_COLOR(7);
BIND_NUMKEY_TO_COLOR(8);
BIND_NUMKEY_TO_COLOR(9);
BIND_NUMKEY_TO_COLOR(0); // This one would need an additional if clause in the macro, I think
``` 

But it would change very little in substance, since it would be expanded to the current code. I know `##` is not entirely immediate, that's why I did not use it.

rolandlo:
> > I don't see why you would need a C preprocessor macro. Can't you use a simple `for`-loop?
> 
> I think It's not possible because the `GDK_Key_xxx` (where `xxx` is the key) are identifiers, while the color indices are numbers.
> 
Note that e.g. `GDK_Key_7` is the same thing as `GDK_Key_0 + 7`.

See `/usr/include/gtk-3.0/gdk/gdkkeysyms.h`
``` 
#define GDK_KEY_0 0x030
#define GDK_KEY_1 0x031
#define GDK_KEY_2 0x032
#define GDK_KEY_3 0x033
#define GDK_KEY_4 0x034
#define GDK_KEY_5 0x035
#define GDK_KEY_6 0x036
#define GDK_KEY_7 0x037
#define GDK_KEY_8 0x038
#define GDK_KEY_9 0x039
```

fpesari:
I imagined something like that, it's the same pattern as ASCII, but perhaps I did not pursue it because it's a bit obscure (e.g. people won't be able to grep/ack/rg `GDK_KEY_6` and find where it's used in the codebase).

But this way, one can avoid macros, at least. If there are no objections, I will pursue this solution using a function. If it were my own program, I would still think about using a macro (because a function call is not actually needed, it's a simple substitution) but I'll avoid macros since I see they aren't used a lot in the codebase.

rolandlo:
> I imagined something like that, it's the same pattern as ASCII, but perhaps I did not pursue it because it's a bit obscure (e.g. people won't be able to grep/ack/rg `GDK_KEY_6` and find where it's used in the codebase).
> 
I don't know about other code editors, but in VS code I can just CTRL+click on one of these identifiers and get all the definitions. Even if I just hover over it, I can see the definition. No need to grep. 

> But this way, one can avoid macros, at least. If there are no objections, I will pursue this solution using a function. 

Not sure why you want to use a function either. Currently just a `for`-loop with the duplicated code inside (once) would suffice and be very readable. 

>If it were my own program, I would still think about using a macro (because a function call is not actually needed, it's a simple substitution) but I'll avoid macros since I see they aren't used a lot in the codebase.

The macros we have in Xournal++'s codebase are usually pretty old (except if it concerns compile flags). Macros have disadvantages. See e.g. https://stackoverflow.com/questions/17043090/why-should-i-avoid-macros-in-c

rolandlo:
> Another reason not to use the loop is that AFAIK `XournalView::onKeyPressEvent` is called very often and as such, it's a hot spot. True, it would increase code readability but it could also make the program a bit slower. A macro would not have that problem, of course. Do I go ahead with the `for` loop?

It's only called when a key is pressed and there is no selection. That's not very often. The potential slow-down is minimal. Please go ahead with the `for` loop.

fpesari:
Alright, done. I tested it and it works. The condition in the `for` loop should already ensure that bounds are always respected and I added an extra check for the `0` key, so there are no hard-coded assumptions about how many colors there are (so they could very well be fewer than 10).

The only quirk I encountered is that when a custom toolbar includes only some colors, the hotkey related to the missing color will still select it. IMHO, this is not something I should fix here but rather, the colors excluded from the custom toolbar should not be in the `std::vector` containing all colors. So perhaps I should open an issue for that, but I personally think this code should stay the same, and that small quirk happening in a very specific case (with no serious effect other than selecting a color, something which does not affect the document and can easily be reversed) does not outweigh the useful feature offered by this PR.

rolandlo (approved):
LGTM
Thanks again, nice feature!

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,23 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         return true;
     }
 
+    // Switch color on number key
+    auto colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
+    if (!colors) {
+        return false;
+    }
+    for (size_t i = 0; i < colors->size(); i++) {
```
Technius, line 297: The upper bound should be set to `std::min(colors->size(), 10)` since there are at most 10 number keys.

Also, it is possible to avoid the loop by checking whether `GDK_KEY_0 <= event->keyval < GDK_KEY_0 + std::min(10, colors->size())` and then computing the index into `colors` using `std::min(colors.size() - 1, (9 + (event->keyval - GDK_KEY_0)) % 10)`.  Note that the bounds check will force `colors.size() > 0`, so this should be safe even if there are no colors in the toolbar.

src/core/gui/toolbarMenubar/AbstractToolItem.h
```diff
@@ -67,6 +67,8 @@ class AbstractToolItem: public AbstractItem {
      */
     void enable(bool enabled) override;
 
+    GtkToolItem* getItem();
```
Technius, line 70: nit:
```suggestion
    GtkToolItem* getItem() const;
```

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -610,6 +610,8 @@ auto ToolMenuHandler::isColorInUse(Color color) -> bool {
 
 auto ToolMenuHandler::getToolItems() -> std::vector<AbstractToolItem*>* { return &this->toolItems; }
 
+auto ToolMenuHandler::getColorToolItems() -> std::vector<ColorToolItem*>* { return &this->toolbarColorItems; }
```
Technius, line 613: nit:
```suggestion
auto ToolMenuHandler::getColorToolItems() const -> const std::vector<ColorToolItem*>& { return this->toolbarColorItems; }
```

fpesari:
I implemented the changes suggested by @Technius , except the second one -- `const` in that case would not work, the compiler would complain about `[clang] Cannot initialize return object of type 'std::vector<ColorToolItem *> *' with an rvalue of type 'const std::vector<ColorToolItem *> *'`.

BTW, @Technius, excellent C-maths skills! I knew `%` could be a solution but I could not figure out how :smile_cat:

fpesari:
My bad - @Technius's second suggestion was good, too, it was I that missed a `const` identifier, which triggered the error. All of @Technius's suggestions have been implemented, then.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -610,6 +610,8 @@ auto ToolMenuHandler::isColorInUse(Color color) -> bool {
 
 auto ToolMenuHandler::getToolItems() -> std::vector<AbstractToolItem*>* { return &this->toolItems; }
 
+auto ToolMenuHandler::getColorToolItems() const -> const std::vector<ColorToolItem*>* { return &this->toolbarColorItems; }
```
bhennion, line 613: I think this should return a reference rather than a pointer:
```suggestion
auto ToolMenuHandler::getColorToolItems() const -> const std::vector<ColorToolItem*>& { return this->toolbarColorItems; }
```
(Of course, this also means adapting the calling function)

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,22 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         return true;
     }
 
+    // Switch color on number key
+    auto colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
+    if (!colors) {
+        return false;
+    }
+    if ((event->keyval >= GDK_KEY_0)
+        && (event->keyval < GDK_KEY_0 + std::min(10, (int)colors->size()))) {
```
Technius, line 298: nit: use an unsigned literal instead of casting, since everything here is already unsigned to begin with.
```suggestion
        && (event->keyval < GDK_KEY_0 + std::min(10U, colors->size()))) {
```

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,22 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
...
+            int index = std::min((int)colors->size() - 1,
+                                 (int)(9 + (event->keyval - GDK_KEY_0)) % 10);
```
Technius, line 300: nit: same here

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,22 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
...
+            auto colorToolItem = colors->at((size_t)index);
```
Technius, line 301: nit: index should be declared as `size_t` instead of requiring a cast

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,22 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         return true;
     }
 
+    // Switch color on number key
+    auto colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
```
Technius, line 293: nit: when using auto, it helps readability to be explicit about whether the type is a pointer or reference.
```suggestion
    auto *colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
```

Technius (approved):
The logic looks okay to me, but I have some nitpicks above the integer conversions and styles.

fpesari:
I implemented the changes suggested by @Technius and @bhennion .

fpesari:
So, I am on GNU/Linux and here and on MacOS the build finishes correctly, but it seems from CI that the build fails on Windows because of this:

```
 error: no matching function for call to 'min(long unsigned int, std::vector<ColorToolItem*>::size_type)'
```

It seems C++23 [will have size_t literals](https://www.sandordargo.com/blog/2022/05/25/literal_suffix_for_signed_size_t) but until Xournal++ switches to it, I think a cast is the easiest cross-platform solution. So I changed the code a bit from before (I think now the casts are cleaner) but I reverted to casting, which works on all platforms.

src/core/gui/XournalView.cpp
```diff
@@ -287,6 +289,20 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
...
     }
 
+    // Switch color on number key
+    auto &colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
+    if ((event->keyval >= GDK_KEY_0)
+        && (event->keyval < GDK_KEY_0 + std::min((std::size_t)10,
+                                                 colors.size()))) {
+            std::size_t index = std::min(colors.size() - 1,
+                                         (std::size_t)(9 + (event->keyval - GDK_KEY_0)) % 10);
+            auto colorToolItem = colors.at(index);
```
Technius, line 299: nit: for readability, please explicitly indicate that this is a pointer:
```suggestion
            auto *colorToolItem = colors.at(index);
```

Technius (approved):
LGTM, just one last minor readability nitpick. Otherwise I'll be merging in 24 hours unless someone objects.

PR #4417, opened 2022-11-15
First half of #4271

bhennion:
This is the first half of #4271, to ease the reviewing process.

### DIFF
--- a/src/core/gui/TextEditor.cpp
+++ b/src/core/gui/TextEditor.cpp
@@ -150,6 +150,10 @@ auto TextEditor::setColor(Color color) -> UndoAction* {
 
 void TextEditor::setFont(XojFont font) {
     this->text->setFont(font);
+    afterFontChange();
+}
+
+void TextEditor::afterFontChange() {
     xoj::view::TextView::updatePangoFont(this->layout, this->text);
     this->repaintEditor();
 }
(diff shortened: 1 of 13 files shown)


src/core/gui/TextEditor.cpp
```diff
@@ -353,55 +339,48 @@ void TextEditor::toggleOverwrite() {
...
  * Improve that later on...
  */
-void TextEditor::decSize() {
+void TextEditor::decreaseFontSize() {
     XojFont& font = text->getFont();
-    double fontSize = font.getSize();
-    fontSize--;
-    font.setSize(fontSize);
-    setFont(font);
+    font.setSize(font.getSize() - 1);
```
Technius, line 344: Does this need to check if the size > 1?

src/util/raii/CStringWrapper.cpp
```diff
@@ -0,0 +1,35 @@
+#include "util/raii/CStringWrapper.h"
+
+#include <utility>
+
+#include <glib.h>
+
+namespace xoj::util {
+inline namespace raii {
+CString::CString(char* s): data(s) {}
```
Technius, line 9: This feels error-prone to me because it takes ownership. I would make this a static method like `CString::steal()` or `CString::take()` or some other explicit name, or even suggest renaming the class to `UniqueString` or something.

src/core/gui/TextEditor.cpp
```diff
@@ -26,55 +28,67 @@
...
+            auto cursorBlinkingPeriod = static_cast<unsigned int>(tmp);
+            this->cursorBlinkingTimeOn = cursorBlinkingPeriod * CURSOR_ON_MULTIPLIER / CURSOR_DIVIDER;
+            this->cursorBlinkingTimeOff = cursorBlinkingPeriod - this->cursorBlinkingTimeOn;
+            /*
+             * The blink timeout is ignored. To remedy this, restore the following snippet and use cursorBlinkTimeout!
+             *
+             * g_object_get(settings, "gtk-cursor-blink-timeout", &tmp, nullptr);
+             * assert(tmp >= 0);
+             * this->cursorBlinkTimeout = static_cast<unsigned int>(tmp);
+             */
```
Technius, line 64: I would just delete this and the `cursorBlinkTimeout` variable. We never used it to begin with.

Technius (approved):
LGTM, some minor nitpicks below.

src/core/gui/TextEditor.cpp
bhennion: Nice catch!

src/util/raii/CStringWrapper.cpp
bhennion: Ok. I changed to the overly causious `static OwnedCString::assumeOwnership(char* s)`.

bhennion:
Implemented the required changed.

Issue #4367, opened 2022-10-26
Cannot delete selected items if the hand tool is active


K4LCIFER:
**Affects versions :**
 - OS: Arch Linux
 - (Linux only) Desktop environment: Gnome 42.5, Wayland
 - Which version of libgtk do you use 3.24.34
 - Version of Xournal++: 1.1.2
 - Installation method: flatpak

**Describe the bug**
if you select an item, you are unable to delete the selected item if the hand tool is active.

**To Reproduce**
1. Create an item (text etc)
2. select it (eg with the lasso select tool)
3. select the hand tool, and observe that you cannot delete it

**Expected behavior**
I'm not sure if this is intended behaviour, or not. I initially expected that I would be able to click the x, but I could understand an argument for why that would be undesirable

**Screenshots of Problem**
![test2](https://user-images.githubusercontent.com/7829231/197945028-952bb629-3723-4f79-8859-04d2c8e41593.gif)

RuleOfThrees:
Hello! I'd like to take a look at this one.

Jared234:
I agree with @forgottosave that the "delete" button shouldn't be shown when the hand tool is selected. I am interested in working on this issue and can create a PR request soon.

mjg:
Apparantly, the hand tool is the only tool which does not de-select an existing selection when it is activated. Changing that would be the more uniform solution than the one from #4419 . Otherwise, expect users to ask why the anchors disappear ...

Technius:
We "fixed" this issue by hiding the selection controls when using the hand tool (#4419).

PR #4419, opened 2022-11-15
Removed the "DeleteRect" if the hand tool is selected

Jared234:
This is an attempt to fix the bug mentioned in #4367.
The user who created this issue wanted to be able to use the hand tool to delete a selected item by clicking on the delete rectangle next to it. However, the hand tool should not be able to do this in the first place.
Still, I feel that the delete rectangle right next to the element can be a bit misleading. So I decided (as discussed in #4367) to simply not draw the "DeleteRect" when the hand tool is selected.

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -1037,8 +1037,10 @@ void EditSelection::paint(cairo_t* cr, double zoom) {
     // bottom right
     drawAnchorRect(cr, x + width, y + height, zoom);
 
-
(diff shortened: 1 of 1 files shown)


rolandlo (changes requested):
Thank you for your contribution. 
I think the anchors (for resizing and rotation) should also not be drawn when the hand tool is used, since they also can't be used with that tool.

Jared234:
Thanks for the review. I agree. 👍

rolandlo (approved):
LGTM, thanks!

bhennion:
Let's maybe wait until #4455 is merged before merging this.

Issue #1625, opened 2019-11-22
'First Page Offset' as document property


tattsan:
**Is your feature request related to a problem? Please describe.**
When we make pages paired, we set first page offset at 'Drawing Area' menu in 'Edit Preferences'. Prefered offset is different at each document, according to it has a cover page or not. We have to change this setting manually, but I want it to be done automatically.

**Describe the solution you'd like**
Create document property 'First Page Offset', and Xournalpp reads it to set the offset per document. This setting should be done manually only once. 'Drawing Area' menu would set its default value.

tattsan:
I have recently come to the conclusion that it would be better if the offset could be changed dynamically and quickly while working. I've made a PR #4424 to add an ACTION so that this can be done in a plugin.

PR #4424, opened 2022-11-18
Add ACTION_TOGGLE_PAIRS_PARITY

tattsan:
~These 8 actions  enables one to change the first page offset for paired pages quickly with plugins, without opening preferences menu.~ I've added `ACTION_TOGGLE_PAIRS_PARITY` as an alternative. See https://github.com/xournalpp/xournalpp/pull/4424#issuecomment-1324534545 below.

 This is an alternative solution for #1625.


src/core/control/Control.cpp
```diff
@@ -1602,6 +1634,38 @@ void Control::setViewPresentationMode(bool enabled) {
...
+        case 6:
+            action = ACTION_SET_PAIRS_OFFSET_6;
+            break;
+        case 7:
+            action = ACTION_SET_PAIRS_OFFSET_7;
+            break;
+        default:
+            action = ACTION_SET_PAIRS_OFFSET;
+    }
+
```
rolandlo, line 1668: What is this code segment used for? The variable `action` looks unused.

tattsan: I mistook it for an `action` in `fireActionSelected()` which is fired in `setViewColumns()`. I will remove it.

bhennion:
I'm a bit confused here. Is there really a difference between ACTION_SET_PAIRS_OFFSET_0 and ACTION_SET_PAIRS_OFFSET_2? (or between ACTION_SET_PAIRS_OFFSET_1 and ACTION_SET_PAIRS_OFFSET_3)

Technius:
Side note: I'm not really happy how we have to manually "inline" a function call into 8 different actions. To support functionality like this, we'll want to generalize our action system so that an action is a data structure and not just a constant.

bhennion:
> Side note: I'm not really happy how we have to manually "inline" a function call into 8 different actions. To support functionality like this, we'll want to generalize our action system so that an action is a data structure and not just an enum case.

This is already happening in #4372. To be discussed there.

rolandlo:
> I'm a bit confused here. Is there really a difference between ACTION_SET_PAIRS_OFFSET_0 and ACTION_SET_PAIRS_OFFSET_2? (or between ACTION_SET_PAIRS_OFFSET_1 and ACTION_SET_PAIRS_OFFSET_3)

There is actually. Take a document with lots of pages and use `Multiple Page Layout` combined with say a layout in 4 rows, then it looks like this with a 2 page offset:

grafik

offset 0, offset 1, offset 2 and offset 3 are all different in that case.

tattsan:
I use only ACTION_SET_PAIRS_OFFSET_0 and ACTION_SET_PAIRS_OFFSET_1 in my daily work. But in the preferences menu, the offset can be set up to :100:.

tattsan:
'Pairs offset' seems to function differently in its parity and its absolute value.
The former is related to a local issue of "which of the neighboring pages to pair with", and
the latter is related to global layout (especially when using multi-columns or multi-rows).
Perhaps the latter should be configurable independently of pairing pages.

Since the original motivation for this PR was the former, I've added `ACTION_TOGGLE_PAIRS_PARITY`
as an alternative. If there is support for this one,  I would like to remove the 8 ugly  ACTIONs.

tattsan:
I've removed 8 ugly actions, and request to add only 'ACTION_TOGGLE_PAIRS_PARITY'.

rolandlo (approved):
Looks good to me (and works fine) except that the local variable is called `enabled` which is already a function parameter. It would be cleaner to give it another name.

Also the commits will have to be squashed into one commit with a fitting commit description before (rebase and) merging

tattsan:
Done. Thanks for reviewing!

rolandlo (approved):
LGTM. Merging in 24 hours if no objections are raised.

rolandlo:
Thanks @tattsan for the PR. Just merged it.

PR #4426, opened 2022-11-18
Do not define in `XML_VERSION_STR` variable InputStreamException.cpp file

alex1701c:
We do not use this value in the InputStreamException.cpp and provide it in the Serializable.h using the `extern` declaration.
Considering that we can be sure the serializing utils are always up-to-date when rebuilding the project, there is no problem with having this value header-only for the few usages.

While at it, the value can be made const (as clang-tidy suggested).

### DIFF
--- a/src/util/include/util/serializing/Serializable.h
+++ b/src/util/include/util/serializing/Serializable.h
@@ -16,7 +16,7 @@
 class ObjectInputStream;
 class ObjectOutputStream;
 
-extern const char* XML_VERSION_STR;
+const static char* const XML_VERSION_STR = "XojStrm1:";
 
 class Serializable {
 public:
--- a/src/util/serializing/InputStreamException.cpp
+++ b/src/util/serializing/InputStreamException.cpp
@@ -1,7 +1,5 @@
 #include "util/serializing/InputStreamException.h"
 
-const char* XML_VERSION_STR = "XojStrm1:";
-
 InputStreamException::InputStreamException(const std::string& message, const std::string& filename, int line) {
     this->message = message + ", " + filename + ": " + std::to_string(line);
 }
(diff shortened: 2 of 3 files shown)


alex1701c:
@Technius @rolandlo Any thoughts?

rolandlo:
@alex1701c 
Thanks for the PR. I think `ObjectInputStreamTest.cpp` should also include `Serializable.h` in your commit. We want to "include what you use", see #4098.

alex1701c:
That is reasonable, I have adjusted the PR

rolandlo:
The PR looks good to me. But why does the Windows test fail?

alex1701c:
```log
test 1
    Start 1: test-units_NOT_BUILT
Could not find executable test-units_NOT_BUILT
Unable to find executable: test-units_NOT_BUILT
Looked in the following places:
test-units_NOT_BUILT
test-units_NOT_BUILT.exe
Release/test-units_NOT_BUILT
Release/test-units_NOT_BUILT.exe
Debug/test-units_NOT_BUILT
Debug/test-units_NOT_BUILT.exe
MinSizeRel/test-units_NOT_BUILT
MinSizeRel/test-units_NOT_BUILT.exe
RelWithDebInfo/test-units_NOT_BUILT
Errors while running CTest
Output from these tests are in: D:/a/1/s/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
##[error]Cmd.exe exited with code '8'.
```

That does not make sense to me, because I changed nothing in the buildsystem.

alex1701c:
Could one maybe re-run the pipeline?

rolandlo:
Rerunning the failed test. Let's see.

rolandlo (approved):
The second test run was successful.

PR #4428, opened 2022-11-20
Only fireZoomChanged when zoom has changed

rolandlo:
`fireZoomChanged` should only be called when the zoom has really changed. This is most notable when you pinch to zoom and have the "Start zoom after a distance ..." setting higher than `0\%`. The VertialSpace tool and the geometry tools from PR #4211 listen to zoom changes and currently need to redraw the mask when `fireZoomChanged` is called, regardless whether the zoom has really changed or not.


rolandlo:
I'm suggesting to merge this before the 1.1.3 release.

Technius (approved):

PR #4430, opened 2022-11-20
Fix unwanted position change after editing LaTex elements

### LINKED ISSUE #4415: Latex editing Maths equation
**Affects versions :**
 - OS: Windows 10 pRo
 - Version of Xournal++: 1.1.2
   
**Describe the bug**
While editing the latex used to insert maths symbol or equation. The position of the symbol or text gets changed. We have to drag it again to its original position, very much time consuming for maths inputs. Please fix this bug as soon as possible. With lot of thanks.

Chrisimx:
This PR fixes #4415. The reason for the changing position of the new LaTex element was that the coordinates of the selection were used as the new x and y position. The selection box has however a margin and is bigger than the actual LaTex element which caused an offset on completing editing. The commit in this PR updates the code to use the actual element coordinates of the old LaTex element fixing the bug.

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -114,8 +114,10 @@ void LatexController::findSelectedTexElement() {
     if (this->selectedElem) {
         // this will get the position of the Latex properly
         EditSelection* theSelection = control->getWindow()->getXournal()->getSelection();
-        this->posx = theSelection->getXOnView();
-        this->posy = theSelection->getYOnView();
+        // when latex is edited by double-clicking on it the selection is updated to the double clicked latex element
+        // therefore we can assume there's exactly one element in the selection
+        this->posx = theSelection->getElements()[0]->getX();
+        this->posy = theSelection->getElements()[0]->getY();
 
         if (auto* img = dynamic_cast<TexImage*>(this->selectedElem)) {
             this->initialTex = img->getText();
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -114,10 +114,9 @@ void LatexController::findSelectedTexElement() {
     if (this->selectedElem) {
         // this will get the position of the Latex properly
         EditSelection* theSelection = control->getWindow()->getXournal()->getSelection();
-        // when latex is edited by double-clicking on it the selection is updated to the double clicked latex element
-        // therefore we can assume there's exactly one element in the selection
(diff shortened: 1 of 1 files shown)


src/core/control/LatexController.cpp
```diff
@@ -114,8 +114,10 @@ void LatexController::findSelectedTexElement() {
     if (this->selectedElem) {
         // this will get the position of the Latex properly
         EditSelection* theSelection = control->getWindow()->getXournal()->getSelection();
-        this->posx = theSelection->getXOnView();
-        this->posy = theSelection->getYOnView();
+        // when latex is edited by double-clicking on it the selection is updated to the double clicked latex element
+        // therefore we can assume there's exactly one element in the selection
+        this->posx = theSelection->getElements()[0]->getX();
+        this->posy = theSelection->getElements()[0]->getY();
```
rolandlo, line 120: A slightly more elegant solution (in my opinion) to achieve the same thing would be:
```suggestion
        xoj::util::Rectangle<double> rect = theSelection->getSnappedBounds();
        this->posx = rect.x;
        this->posy = rect.y;
```

rolandlo (commented):
Thanks for the fix. What do you think of the suggested alternative?

Technius:
This fix seems simple enough to sneak into 1.1.3.

src/core/control/LatexController.cpp
Chrisimx: I like the alternative. Thanks a lot!

rolandlo:
> This fix seems simple enough to sneak into 1.1.3.

Ok, then we either have to use @Chrisimx's version, since `EditSelection::getSnappedBounds` is only available in master (since #4359), or backport some of #4359 (nothing non-trivial) to get that method into `release-1.1`.

rolandlo:
@Technius 
Which way would you prefer? Since @Chrisimx liked and applied my alternative and adding the declaration and definition of `EditSelection::getSnappedBounds` doesn't do any harm (when rebased on `release-1.1`), I would probably go that way.

Technius:
It seems more hassle than it's worth at the moment; it'll make merging the backported changes back more complicated. Probably easier to leave this PR as-is.

Chrisimx:
If it's okay, I'm going to squash the two commits then

Technius:

Don't worry about squashing, the person who does the merging can "Squash and Merge" from GitHub.

PR #4435, opened 2022-11-21
Remove dead copy selection code

rolandlo:
I haven't researched when this method became unused, but it's definitively dead code.


rolandlo:

Technius (approved):
LGTM. This must've become dead when we refactored `EditSelection::copySelection()` to preserve the insert order.

Issue #4032, opened 2022-05-16
Looks like `getSelectedLayerId` will insert a layer if there are no layers in the page. That's pretty undesirable imo, we should refactor it (perhaps in a followup) so that the page layers are always in a valid state.


Febbe:
Looks like `getSelectedLayerId` will insert a layer if there are no layers in the page. That's pretty undesirable imo, we should refactor it (perhaps in a followup) so that the page layers are always in a valid state.

_Originally posted by @Technius in https://github.com/xournalpp/xournalpp/pull/4025#discussion_r873280853_

forgottosave:
I also think it is very confusing, that all objects seem to be on the layer "Background" and as soon as you insert a new layer it becomes Layer 1 and a Layer 2 is added, resulting in "3 layers".

Did you mean it should be refactored so that one always works on "Layer 1", therefor resulting in an always valid layer?

Technius:
By that, what I mean is that the page should ensure that there is always a layer when the are created/deleted, instead of creating one in `getSelectedLayerId()` (e.g., it should not modify the layer vector).

forgottosave:
Okay, perfect. I worked on some layer related stuff recently and I already stumbled upon this problem.
I can create a PR soon.

PR #4439, opened 2022-11-23
Ensure layers are always in a valid state

forgottosave:
This PR solves #4032 

`getSelectedLayer()` created a layer if none was existent before.
This PR changes the XojPage, so that there is always minimum one valid layer.

I think this should be mostly finished, but I still have to check if there are any exception cases.

### DIFF
--- a/src/core/control/xojfile/LoadHandler.cpp
+++ b/src/core/control/xojfile/LoadHandler.cpp
@@ -318,7 +318,7 @@ void LoadHandler::parseContents() {
         double width = LoadHandlerHelper::getAttribDouble("width", this);
         double height = LoadHandlerHelper::getAttribDouble("height", this);
 
-        this->page = std::make_unique<XojPage>(width, height);
+        this->page = std::make_unique<XojPage>(width, height, /*suppressLayer*/true);
 
         pages.push_back(this->page);
     } else if (strcmp(elementName, "audio") == 0) {
@@ -903,8 +903,8 @@ void LoadHandler::parserEndElement(GMarkupParseContext* context, const gchar* el
         handler->pos = PASER_POS_FINISHED;
     } else if (handler->pos == PARSER_POS_IN_PAGE && strcmp(elementName, "page") == 0) {
         // handle unnecessary layer insertion in case of existing layers in file
-        if (handler->page->getLayerCount() > 1) {
-            handler->page->removeLayer(handler->page->getLayers()->at(0));
+        if (handler->page->getLayerCount() == 0) {
+            handler->page->addLayer(new Layer());
         }
         handler->pos = PARSER_POS_STARTED;
         handler->page = nullptr;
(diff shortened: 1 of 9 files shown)


bhennion:
This seems to cause all the save/load related test units to segfault. I assume something needs to be done about opening a file where some pages have no layers, but the issue might be deeper than that.

forgottosave:
> 

Yes, thank you :)
That's one of the issues I meant with exception cases. Opening old files there is currently always one layer too much. Because the layer is being inserted before the rest of the layers are loaded. Also I think there is an issue with pages where there are no layers.

I'll look into this and open it for review when it's fixed :)

forgottosave:
I think it should be ready now. While loading a page with at least one layer the first layer (inserted to keep at least one valid layer) will now be removed.

bhennion:
Sorry for the delay. I forgot about this. I'll try an review it next week. Meanwhile, could you rebase and resolve the conficts?

forgottosave:
No problem, I'll rebase & resolve the conflicts :)

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -902,6 +902,10 @@ void LoadHandler::parserEndElement(GMarkupParseContext* context, const gchar* el
     if (handler->pos == PARSER_POS_STARTED && strcmp(elementName, handler->endRootTag) == 0) {
         handler->pos = PASER_POS_FINISHED;
     } else if (handler->pos == PARSER_POS_IN_PAGE && strcmp(elementName, "page") == 0) {
+        // handle unnecessary layer insertion in case of existing layers in file
+        if (handler->page->getLayerCount() > 1) {
+            handler->page->removeLayer(handler->page->getLayers()->at(0));
+        }
```
bhennion, line 908: I find this code block a little weird. I think it'd be better if:

- The Page constructor had a flag "suppressLayerCreation" (defaulting to false), so that the loader would not create an empty layer and then remove it (in most cases).
- Then when a page is finished parsing, and if there is no layers in it, add an empty layer.

forgottosave: Sure, I can change that logic.

src/core/model/XojPage.h
```diff
@@ -26,7 +26,7 @@
 
 class XojPage: public PageHandler {
 public:
-    XojPage(double width, double height);
+    XojPage(double width, double height, bool suppressLayerCreations = false);
```
bhennion, line 29:
```suggestion
    XojPage(double width, double height, bool suppressLayerCreation = false);
```

bhennion (approved):
Just a typo left. Otherwise, LGTM!

src/core/control/xojfile/LoadHandler.cpp
bhennion: One way around this would be to only contruct the page instance once the parser is done parsing its content (the layers). This could be done in a later PR if someone thinks it matters enough.

src/core/model/XojPage.h
forgottosave: Thanks, fixed it

bhennion:
Thanks! Squashing and merging in 24 hours unless an objection is raised.

Issue #4410, opened 2022-11-13
Option to clear recent documents


priyavrat-misra:
Please consider adding a clear recent documents option in the next release.
It's not that big of a deal but it kind of feels odd since almost all apps have it and this doesn't. <del>It gives me OCD.</del>

forgottosave:
Could you specify what you mean by clear document?

priyavrat-misra:
@forgottosave Something like this, which will clear recently opened files.
![image](https://user-images.githubusercontent.com/73816335/202953648-ea19fd66-bd3e-4c96-add6-6e0db00a4bb5.png)


Below is VLC Media Player's version.
![image](https://user-images.githubusercontent.com/73816335/202953780-5494e4d1-3d2d-4d99-bdd5-50666de9d913.png)

PR #4440, opened 2022-11-23
Implemented clear function for recent files list

fpesari:
Implements #4410.

I only wrote the code, the `"Clear list"` string needs to be translated.

### DIFF
--- a/src/core/control/RecentManager.cpp
+++ b/src/core/control/RecentManager.cpp
@@ -70,6 +70,20 @@ void recentsMenuActivateCallback(GtkMenuItem* self, RecentManager* recentManager
     }
 }
 
+void recentsClearActivateCallback(GtkMenuItem* self, GtkRecentManager* recentManager) {
+    GList* items = gtk_recent_manager_get_items(recentManager);
+    auto item_view = GListView<GtkRecentInfo>(items);
+    for (auto& recent: item_view) {
+        if (filterRecent(recent, false) ||
+            filterRecent(recent, true)) {
+            gtk_recent_manager_remove_item(recentManager,
+                                           gtk_recent_info_get_uri(&recent),
+                                           nullptr);
+        }
+    }
+    g_list_free_full(items, GDestroyNotify(gtk_recent_info_unref));
+}
+
 }  // namespace
 
 RecentManagerListener::~RecentManagerListener() = default;
@@ -230,6 +244,18 @@ void RecentManager::updateMenu() {
             xoj_files.emplace_back(recent);
         }
     }
+
+    if (pdf_files.empty() && xoj_files.empty()) {
+        GtkWidget* noFilesItem = gtk_menu_item_new_with_mnemonic(
+            _("No recent files"));
+        gtk_menu_shell_append(GTK_MENU_SHELL(menu), noFilesItem);
+        gtk_widget_set_visible(GTK_WIDGET(noFilesItem), true);
+        gtk_widget_set_sensitive(GTK_WIDGET(noFilesItem), false);
+        this->menuItemList.push_back(noFilesItem);
+        g_list_free_full(items, GDestroyNotify(gtk_recent_info_unref));
+        return;
+    }
+
     std::sort(xoj_files.begin(), xoj_files.end(), std::greater<GtkRecentInfo>());
     std::sort(pdf_files.begin(), pdf_files.end(), std::greater<GtkRecentInfo>());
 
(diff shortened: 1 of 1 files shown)


src/core/control/RecentManager.cpp
```diff
@@ -245,12 +258,40 @@ void RecentManager::updateMenu() {
...
+
     g_list_free_full(items, GDestroyNotify(gtk_recent_info_unref));
 }
+
+auto RecentManager::listIsEmpty(GtkRecentManager* recentManager) -> bool {
+    GList* items = gtk_recent_manager_get_items(recentManager);
+    auto item_view = GListView<GtkRecentInfo>(items);
+    for (auto& recent: item_view) {
+        if (filterRecent(recent, false) || filterRecent(recent, true)) {
+            return false;
```
rolandlo, line 293:
```suggestion
            g_list_free(items);
            return false;
```

src/core/control/RecentManager.cpp
```diff
@@ -245,12 +258,40 @@ void RecentManager::updateMenu() {
...
+        }
+    }
+    return true;
```
rolandlo, line 296:
```suggestion
    g_list_free(items);
    return true;
```

src/core/control/RecentManager.cpp
```diff
@@ -70,6 +70,19 @@ void recentsMenuActivateCallback(GtkMenuItem* self, RecentManager* recentManager
...
+    auto item_view = GListView<GtkRecentInfo>(items);
+    for (auto& recent: item_view) {
+        if (filterRecent(recent, false) ||
+            filterRecent(recent, true)) {
+            gtk_recent_manager_remove_item(recentManager,
+                                           gtk_recent_info_get_uri(&recent),
+                                           nullptr);
+        }
+    }
+}
```
rolandlo, line 84:
```suggestion
    }
    g_list_free(items);
}
```

rolandlo (changes requested):
Thanks a lot for this valuable contribution! The features works fine and I'm looking forward to having it available. Codewise I have just one comment:
I think the `GList*`s need to be freed before the function return.

rolandlo:
Please don't merge `master` into your branch. Drop that commit again. Rather `rebase` on `master` (`git rebase upstream/master`)

fpesari:
> Please don't merge `master` into your branch. Drop that commit again. Rather `rebase` on `master` (`git rebase upstream/master`)

Yes, sorry, I will be fixing this soon.

fpesari:
Sorry, the commit was automatically closed while fixing this in git. Reopening it, everything should be OK now (and GList* should also be freed).

rolandlo (approved):
LGTM

Just mark the PR as "ready for Review". Translations will be handled by the [Crowdin](https://crowdin.com/project/xournalpp) project. Merging in 24 hours if no objections are raised and #4455 has been merged.

src/core/control/RecentManager.cpp
```diff
@@ -245,12 +259,43 @@ void RecentManager::updateMenu() {
...
+    if (!listIsEmpty(recentManager)) {
+        GtkWidget* separator = gtk_separator_menu_item_new();
+        gtk_menu_shell_append(GTK_MENU_SHELL(menu), separator);
+        gtk_widget_set_visible(GTK_WIDGET(separator), true);
+        this->menuItemList.push_back(separator);
+    }
 
     insert_items_job(pdf_files, base);
 
+    if (!listIsEmpty(recentManager)) {
```
bhennion, line 271: Those successive calls to listIsEmpty implies the list of recent files is iterated upon several times, and unnecessarily so. Note that `listIsEmpty(...) == false` if and only if `pdf_files.empty() && xopp_files.empty()`.
I guess a more optimal code would simply add
```
if (pdf_files.empty() && xopp_files.empty()) {
    g_list_free_full(items, GDestroyNotify(gtk_recent_info_unref));
    return;
}
```
right after the loop populating the vectors pdf_files and xopp_files.

With this, `listIsEmpty()` becomes unused and can be removed as well.

bhennion (changes requested):

fpesari:
Indeed @bhennion, nice catch. I fixed the code according to your suggestion.

bhennion:
Maybe one last thing: we could add a disabled menu entry saying "No recent files listed" or something of the sort, when the list is empty. I think it'd be better than having and empty submenu (which looks like a bug).

Also, the separator between the pdf and xopp sections could be removed when one of those sections is empty.

fpesari:
> Maybe one last thing: we could add a disabled menu entry saying "No recent files listed" or something of the sort, when the list is empty. I think it'd be better than having and empty submenu (which looks like a bug).
> 
> Also, the separator between the pdf and xopp sections could be removed when one of those sections is empty. It would look better, I think.

All good suggestions! Done.

src/core/control/RecentManager.cpp
```diff
@@ -230,6 +244,17 @@ void RecentManager::updateMenu() {
...
     }
+
+    if (pdf_files.empty() && xoj_files.empty()) {
+        GtkWidget* noFilesItem = gtk_menu_item_new_with_mnemonic(
+            _("No recent files"));
+        gtk_menu_shell_append(GTK_MENU_SHELL(menu), noFilesItem);
+        gtk_widget_set_visible(GTK_WIDGET(noFilesItem), true);
+        gtk_widget_set_sensitive(GTK_WIDGET(noFilesItem), false);
+        this->menuItemList.push_back(noFilesItem);
+        return;
```
bhennion, line 255: Is there a reason you removed the list clearing? Even if there are no recent files matching the types xopp or pdf, this GList might contain other types of files, I guess. Or did you get some warnings telling you that `items` was a nullptr?
```suggestion
        g_list_free_full(items, GDestroyNotify(gtk_recent_info_unref));
        return;
```

fpesari: > Is there a reason you removed the list clearing? Even if there are no recent files matching the types xopp or pdf, this GList might contain other types of files, I guess. Or did you get some warnings telling you that `items` was a nullptr?

I made some tests and it seemed to have no effect. Still, if you say that it could have, I may very well put it back in (and I just did, as a matter of fact!).

bhennion: The goal is to avoid a memory leak. The pointer `items` points to memory allocated via a gtk function. Said memory needs to be cleared before the destruction of the pointer itself.

bhennion (approved):
LGTM. Merging in 24 hours unless an objection is raised.

src/core/control/RecentManager.cpp
soumyatheman: Where is the details of recent documents actually written? I am trying to clear it manually

fpesari: > Where is the details of recent documents actually written? I am trying to clear it manually

`~/.local/share/recently-used.xbel`

PR #4442, opened 2022-11-23
TextEditor refactor prior to #4271

bhennion:
This contains more TextEditor refactoring, previously contained in #4271. It is split off from #4271 to ease the reviewing process.


bhennion:
A bunch of changes here require some testing. Most of it has already been thoroughly tested (at least before rebase), but some localization related things were not. In particular:

- In particular, if someone using (on a daily basis) an Input Method could give feedback on how this feels, both compared to master and to other applications.
- The feedback of a user whose localization involves right-to-left writing would also be great.

Technius:
What was broken about the undo/redo?

src/core/gui/TextEditor.cpp
```diff
@@ -630,31 +607,10 @@ void TextEditor::findPos(GtkTextIter* iter, double xPos, double yPos) const {
 
 void TextEditor::contentsChanged(bool forceCreateUndoAction) {
     std::string currentText = getText()->getText();
-
-    // I know it's a little bit bulky, but ABS on subtracted size_t is a little bit unsafe
-    if (forceCreateUndoAction ||
-        ((lastText.length() >= currentText.length()) ? (lastText.length() - currentText.length()) :
-                                                       (currentText.length() - lastText.length())) > 100) {
-        if (!lastText.empty() && !this->undoActions.empty() &&
-            this->undoActions.front().get().getUndoText() != currentText) {
```
bhennion, line 563: The point is that TextEditor::undoActions stayed forever empty because of this test, so there was no undo-redo while editing text.

Now, fixing this at this point seems a bit pointless, as GtkTextBuffer implements an undo-redo mechanism in GTK 4.

bhennion:
Rebase on master.

@tattsan If I remember correctly, you're using an IM in your daily workflow, right? Would you mind trying out this PR and report on whether the text editor behaves as you expect or not? (you can also try #4271 directly if you prefer)

tattsan:
Thank you for the notification @bhennion. Yes, I always use an IME for Japanese input.
 
I tried this PR, and found an issue. The IME adds some attributes  (underlines, highlights, and so on)  to pre-edit strings, which should be cleared after pre-editing, but some of the attributes are uncleared. When the next pre-editing starts, the garbage attributes are cleared.

If I use #4271 directly, the above issue does not occur. However, #4271 contains another issue which leads to crash.

tattsan:
So the attribute issue is really a splitting issue from #4271, and there's a more critical issue of crashing. I moved the report in #4271 here. To reproduce the crash:
* Start texteditor, and input an emoji with an IME.
* Insert a newline and input an emoji with an IME.
* Then the second emoji will be placed at the first line, so the two charactors are displayed overwraped.
* Exit the texteditor, but keep the state texttool.
* Click the overwraped characters, then the app will crash.

Please look in  the attached  errorlog file.
[errorlog.20230109-081341.log](https://github.com/xournalpp/xournalpp/files/10369743/errorlog.20230109-081341.log)

tattsan:
Early reports were inaccurate, please see the latest.

bhennion:
Thanks a lot @tattsan for testing this. I'll have a look and see if I can reproduce.

tattsan:
Emoji-overwrapping issue also occurs on master. So please ignore this one.

tattsan:
I could reproduce the crash with [IBus Typing Booster](https://mike-fabian.github.io/ibus-typing-booster/), an IME which seems to be useful for European people.

bhennion:
OK, I could reproduce the crash.

I could not reproduce the emoticon overwrapping though. Question: when typing kanji's, you're writing from right to left, correct? This may explain my inability to reproduce the overlapping.

tattsan:
> Question: when typing kanji's, you're writing from right to left, correct? This may explain my inability to reproduce the overlapping.

No, left to right. And the overwrapping issue occurs only for color emoji, not for kanji. I use `Noto Color Emoji` font, and it can be a font specific issue. Xournal++ version 1.1.0 already has this issue ( but not in 1.0.x ). I'll open a new issue for the emoji-overwrapping.

bhennion:
Actually, the crash has nothing to do with IME's. Just copy-paste this 😃😄 into a text box, and click in the middle of one of the emoji. You'll get the crash. I think I understand where this comes from.

bhennion:
Ok. The crash should be fixed now! @tattsan could you confirm that?

bhennion:
> No, left to right. And the overwrapping issue occurs only for color emoji, not for kanji. I use `Noto Color Emoji` font, and it can be a font specific issue.

Ok, I can reproduce with your font, both with master or with 1.1.3. I think it is a font specific issue. It seems the font reports a very small line height (see the cursor is also very short when using this font -- at least until you put in an actual character).

This weird line height also appears in purely GTK windows: here, the font chooser

https://user-images.githubusercontent.com/17818041/211343314-0e9aae1b-e1fb-4787-a4ef-2c00b3ed5db4.mp4

See how the blinking cursor gets tiny if no characters are entered. This only happens with this font and I'm assuming something is wrong somewhere in the font's encoding.

tattsan:
I've confirmed the crash resolved, thank you!
There remains the attribute issue I first reported. It doesn't occur with #4271.

bhennion:
> I've confirmed the crash resolved, thank you!

Thanks for testing!

> There remains the attribute issue I first reported. It doesn't occur with #4271.

I'm not quite sure what is the expected behaviour, and what goes wrong. Could it be possible to have a video of the expected behaviour, and a video of the wrong behaviour?

tattsan:
I'll show two videos, `Good-4271.mp4` and `Bad-4442.mp4`. While my IME is showing reverse triangle, showing underline or highlight is a normal behaviour. But when no reverse triangle is shown, underline or highlight should NOT  be shown.


https://user-images.githubusercontent.com/6355486/211691741-537ed923-afe7-4b28-8aab-d38e04377d1c.mp4


https://user-images.githubusercontent.com/6355486/211691806-b59d50fe-81d8-498b-8bc8-e69bcd950bde.mp4

bhennion:
Thanks a lot @tattsan! I finally figured out how to reproduce.https://github.com/xournalpp/xournalpp/pull/4442/commits/be798e1d6583b5f236a60d1f47c7eb7099d6bb22 should fix it!

tattsan:
I've confirmed the issue has been fixed. Thank you!

bhennion:
So now that the IM issues are fixed, this PR is open for review.

rolandlo (commented):
I have looked through the code and have only one nitpick (unused function). Otherwise the changes look good to me. The changes to the `TextEditor` class are really complex though and I can't say I grasped every detail.

Functionwise I found one thing that looks wrong to me (and which is already the same in master): When I write two or more emojis on top of each other (separated by a newline), they are much closer than they should be. This can be observed e.g. by pasting these 4 cats into Xournal++:
🙀
🙀
🙀
🙀
As I can see now, this emoj-overwrapping issue was already reported by @tattsan.

src/core/gui/TextEditor.cpp
```diff
@@ -28,20 +25,112 @@
...
+    GtkTextIter it = {nullptr};
+    gtk_text_buffer_get_start_iter(buf, &it);
+
+    // Fast forward to the beginning of the line containing our target destination
+    for (int linelength = gtk_text_iter_get_bytes_in_line(&it);
+         linelength <= byteIndex && gtk_text_iter_forward_line(&it);
+         byteIndex -= std::exchange(linelength, gtk_text_iter_get_bytes_in_line(&it))) {}
+
+    // If byteIndex is past the line end (so past the buffer end), this will not change it
+    gtk_text_iter_set_line_index(&it, byteIndex);
```
rolandlo, line 71: Actually this has some issues. When moving to an empty line I get the following warning (with level `Gtk-CRITICAL`):

``` term
(xournalpp:50699): Gtk-CRITICAL **: 15:05:32.081: gtk_text_iter_set_line_index: assertion 'byte_on_line <= bytes_in_line' failed
```
So it looks better to avoid this.

rolandlo: Ok, this is fixed now.

bhennion:
I fixed the warning you pointed out and removed the dead code. Thanks for catching this!

rolandlo (approved):
LGTM from my side. We can't fix the emoj-overwrapping bug from Pango and everything else looks fine now. So feel free to go ahead.

bhennion:
Rebased on master and cleaned up the commit tree.

PR #4443, opened 2022-11-24
Add compass

rolandlo:
This PR is split off from (and based on) #4211. Only the ~last commit 96354a9ca3bb3abadc59e04db2030b0beebfd67c is relevant~commits from "Add compass" are relevant.

Here is a little demo (mostly using the keyboard; `m` is for marking the center):

https://user-images.githubusercontent.com/40485433/204129189-8b9e029f-0a71-4553-ae34-b71b115d135a.mp4

**Edit:** Added demo (Nov 27)


rolandlo:
After #4211 has been reviewed and merged, this PR may be ready for review.

rolandlo:
Rebased on master after merging the setsquare refactoring. This is ready for review.

src/core/view/CompassView.h
```diff
@@ -0,0 +1,115 @@
...
+    void drawDisplays(cairo_t* cr) const override;
+
+    double height = Compass::INITIAL_HEIGHT;
+    double rotation = 0.;
+    cairo_matrix_t matrix{CM, 0., 0., CM, Compass::INITIAL_X, Compass::INITIAL_Y};
+
+    /**
+     * @brief the radius of the semi-circle for the angular marks
+     */
+    double radius = 4.5;
```
Technius, line 63: nit: These seem to be constants and should be marked as `const` or `constexpr`.

src/core/control/Control.cpp
```diff
@@ -1078,6 +1091,19 @@ void Control::makeGeometryTool(GeometryToolType tool) {
...
+            view->addOverlayView(std::make_unique<xoj::view::CompassView>(compass, view, zoom));
+            this->geometryTool = std::unique_ptr<GeometryTool>(compass);
+            this->geometryToolController = std::make_unique<CompassController>(view, compass);
+            std::unique_ptr<GeometryToolInputHandler> geometryToolInputHandler =
+                    std::make_unique<CompassInputHandler>(this->win->getXournal(), geometryToolController.get());
+            geometryToolInputHandler->registerToPool(compass->getHandlerPool());
+            xournal->input->setGeometryToolInputHandler(std::move(geometryToolInputHandler));
+            fireActionSelected(GROUP_GEOMETRY_TOOL, ACTION_COMPASS);
+            geometryTool->notify();
+            break;
```
Technius, line 1105: This shares a lot of code with the setsquare case. Could some of the common parts be extracted out?

Technius: Also, I noticed that on `master` branch (as of `d3cd981a`), the setsquare is not initially repainted when toggled. I imagine that this issue would also affect the compass. Here is one possible fix:
```suggestion
            Range geomRange = geometryTool->getToolRange(/*transformed=*/true);
            if (!geomRange.empty()) {
                view->repaintRect(geomRange.getX(), geomRange.getY(), geomRange.getWidth(), geomRange.getHeight());
            }
            break;
```

Technius (commented):
I skimmed through the code quickly, some comments below.

src/core/view/CompassView.h
rolandlo: They are not constant, but modified here:
https://github.com/rolandlo/xournalpp/blob/bfe1bd9d98e4612702f4e1d9efc3b17788713005/src/core/view/CompassView.cpp#L34-L47

src/core/control/Control.cpp
rolandlo: > This shares a lot of code with the setsquare case. Could some of the common parts be extracted out?

Probably it can be done using templates. Not sure we want that. I will make a suggestion.

rolandlo: > Also, I noticed that on `master` branch (as of `d3cd981a`), the setsquare is not initially repainted when toggled. I imagine that this issue would also affect the compass. 

That should happen via `geometryTool->notify()`, which dispatches a `FLAG_DIRTY_REGION` request that repaints the area in `XojPageView::flagDirtyRegion` through `GeometryToolView::on(FlagDirtyRegionRequest, const Range& rg)`. As far as I can observe in current master, the setsquare is repainted immediately when toggled. Is it different on your system?

rolandlo: > > This shares a lot of code with the setsquare case. Could some of the common parts be extracted out?
> 
> Probably it can be done using templates. Not sure we want that. I will make a suggestion.

Done!

Technius: > As far as I can observe in current master, the setsquare is repainted immediately when toggled. Is it different on your system?

Yeah. On my NixOS 21.11 machine running GNOME Wayland, it doesn't appear until the page is repainted (e.g., by switching tools).

Technius: > Probably it can be done using templates. Not sure we want that. I will make a suggestion.

The proposed function looks good to me as-is, but it does have the following drawback: it forces the view, controller, and input handler constructors to take a specific form. If we want to add another geometry tool in the future that requires more constructor parameters, this won't work anymore and we'll have to move the initialization out of the helper function.

rolandlo: > > As far as I can observe in current master, the setsquare is repainted immediately when toggled. Is it different on your system?
> 
> Yeah. On my NixOS 21.11 machine running GNOME Wayland, it doesn't appear until the page is repainted (e.g., by switching tools).

I found the bug(s). Those are fixed in #4532. Therefore I'll wait for #4532 to get merged before proceeding with the present PR.

src/core/view/CompassView.cpp
```diff
@@ -0,0 +1,202 @@
...
+CompassView::~CompassView() noexcept { this->unregisterFromPool(); };
+
+
+void CompassView::on(FlagDirtyRegionRequest, const Range& rg) { this->parent->flagDirtyRegion(rg); }
+
+void CompassView::on(UpdateValuesRequest, double h, double rot, cairo_matrix_t m) {
+    height = h;
+    rotation = rot;
+    matrix = m;
+    radius = height / std::sqrt(2.) - 1.15;
```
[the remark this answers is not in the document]
rolandlo: The radius was not needed and got removed.

src/core/view/CompassView.cpp
```diff
@@ -0,0 +1,202 @@
...
+    this->registerToPool(compass->getViewPool());
+}
+
+CompassView::~CompassView() noexcept { this->unregisterFromPool(); };
+
+
+void CompassView::on(FlagDirtyRegionRequest, const Range& rg) { this->parent->flagDirtyRegion(rg); }
+
+void CompassView::on(UpdateValuesRequest, double h, double rot, cairo_matrix_t m) {
+    height = h;
```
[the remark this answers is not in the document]
rolandlo: Done

src/core/control/Control.cpp
rolandlo: #4532 is merged, which solves the repainting bug(s).

rolandlo:

bhennion (approved):

Technius:
I played around with this briefly and noticed a few things:

* This uses the same controls as the setsquare, so that would be nice to note in https://github.com/xournalpp/xournalpp.github.io/pull/77
* When drawing a line from the center, the line can only extend in length. If the user overshoots, they have to undo the stroke and try again.
* For the center drawing feature, I think it would be better to have a feature where the user can drag from the center to the edge to fix an angle, and then move along the pending line to adjust the radius. This can be left to a future PR, however.
* The following bug affects both geometry tools and can be addressed in a follow up PR: only one arrow key direction can be active at a time. E.g., when pressing both up+right, the geometry tool only moves in the direction of the first arrow key pressed.

rolandlo:
Thanks for the feedback.

> I played around with this briefly and noticed a few things:
> 
>     * This uses the same controls as the setsquare, so that would be nice to note in [add setsquare documentation xournalpp.github.io#77](https://github.com/xournalpp/xournalpp.github.io/pull/77)
> 
I'm working on it.

>     * When drawing a line from the center, the line can only extend in length. If the user overshoots, they have to undo the stroke and try again.
> 
It's the same for drawing on the outline of the compass and for drawing on the longest side of the setsquare or drawing on physical paper. 

>     * For the center drawing feature, I think it would be better to have a feature where the user can drag from the center to the edge to fix an angle, and then move along the pending line to adjust the radius. This can be left to a future PR, however.
> 
Interesting idea. It wouldn't allow drawing circular ring (sectors) anymore though.

>     * The following bug affects both geometry tools and can be addressed in a follow up PR: only one arrow key direction can be active at a time. E.g., when pressing both up+right, the geometry tool only moves in the direction of the first arrow key pressed.
> 
Is that also a bug of moving the page around (without geometry tools then)? Haven't thought about having multiple arrow keys at the same time.

> 
> These are not major showstoppers, so they can be addressed in a follow up.

PR #4446, opened 2022-11-24
Code cleanup

forgottosave:
I don't know if PRs like this are needed/appreciated, but I've cleaned up some code in a few files. There are no changes in logic. I considered:
- missing spaces behind `if`, `for` and `while`
- append `; ...)` to assignment-if-statements
- nitpicky `// namespace` endings
- ~~avoid usage of `#include <iostream>` (prefer `sstream`)~~ in header files
- avoid `using namespace std`
- inconsistent `newline`-usages across files

If PRs like this are welcome I might check for some C vs. C++ code styling as well. If not I'll close this PR and leave it.

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -118,9 +118,9 @@ void LatexController::findSelectedTexElement() {
         this->posx = rect.x;
         this->posy = rect.y;
 
-        if (auto* img = dynamic_cast<TexImage*>(this->selectedElem)) {
+        if (auto* img = dynamic_cast<TexImage*>(this->selectedElem); img) {
             this->initialTex = img->getText();
-        } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem)) {
+        } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem); img) {
             this->initialTex = "\\text{" + txt->getText() + "}";
         }
         this->imgwidth = this->selectedElem->getElementWidth();
--- a/src/core/control/settings/Settings.cpp
+++ b/src/core/control/settings/Settings.cpp
@@ -1305,7 +1305,7 @@ void Settings::setCursorHighlightBorderColor(Color color) {
 auto Settings::getCursorHighlightBorderWidth() const -> double { return this->cursorHighlightBorderWidth; }
 
 void Settings::setCursorHighlightBorderWidth(double radius) {
-    if (this->cursorHighlightBorderWidth != radius) {
+    if (this->cursorHighlightBorderWidth == radius) {
         return;
     }
     this->cursorHighlightBorderWidth = radius;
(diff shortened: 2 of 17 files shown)


src/core/control/LatexController.cpp
```diff
@@ -118,9 +118,9 @@ void LatexController::findSelectedTexElement() {
         this->posx = rect.x;
         this->posy = rect.y;
 
-        if (auto* img = dynamic_cast<TexImage*>(this->selectedElem)) {
+        if (auto* img = dynamic_cast<TexImage*>(this->selectedElem); img) {
             this->initialTex = img->getText();
-        } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem)) {
+        } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem); img) {
```
rolandlo, line 123:
```suggestion
        } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem); txt) {
```

src/core/control/settings/Settings.cpp
```diff
@@ -1287,45 +1268,48 @@ void Settings::setHighlightPosition(bool highlight) {
...
+        return;
     }
+    this->cursorHighlightBorderColor = color;
+    save();
 }
 
 auto Settings::getCursorHighlightBorderWidth() const -> double { return this->cursorHighlightBorderWidth; }
 
 void Settings::setCursorHighlightBorderWidth(double radius) {
     if (this->cursorHighlightBorderWidth != radius) {
```
rolandlo, line 1308:
```suggestion
    if (this->cursorHighlightBorderWidth == radius) {
```

rolandlo (commented):
I found 2 mistakes. 
One question: Could you please explain 
- why `sstream` is preferred over `iostream`?
- why you suggest to append `; ...)` to assignment-if-statements

As for missing spaces this (and other things) could be handled globally like in PR #4149 (which unfortunately was never reviewed or merged). That would have the benefit that PR's would need less reformatting in future.

Other than that IMHO fixing compiler warnings should have priority. The benefits of small formatting changes (like consistency of empty lines) has to be weighted against the time it takes to review those PR's and the time it takes to solve merge conflicts arising from them.

src/core/control/LatexController.cpp
forgottosave: With certain compilers assignment if-statements produce a warning and it is best practice to ether wrap the assignment in extra brackets (`if ((...))`, source e.g. [llvm](https://llvm.org/docs/CodingStandards.html#treat-compiler-warnings-like-errors)) or (as it's done in other parts in the code) do it with the `;`.
I did the second option, to fit the rest of the code.

forgottosave:
Understandable, thank you for your reply. Makes sense that it takes too much time for basically no effect.

I will maybe look into some compiler warnings then :)

PR #4345, opened 2022-10-21
Fix crash when tex generator is not found

rolandlo:
Fixes #4334
Also fixes a warning when freeing a `g_error` that has not occured (is `nullptr`)

The PR (built from source) can be tested via `PATH="" ./src/xournalpp`, which makes the `PATH` environment variable not contain the latex generator path.

PR #4461, opened 2022-11-30
Fixup crash when TeX is not found

rolandlo:
This fixes an error introduced into master when merging the 
release-1.1 branch after releasing version 1.1.2.
Compare 5e439f58002058041fd8fa936b747b954a79b096

Test it by running 'PATH="" ./xournalpp' and clicking on the TeX
button

The crash was fixed for version 1.1.3 in #4345, but the fix 
didn't work when merged into master.

### DIFF
--- a/src/core/control/latex/LatexGenerator.cpp
+++ b/src/core/control/latex/LatexGenerator.cpp
@@ -10,11 +10,11 @@
 #include "control/settings/LatexSettings.h"  // for LatexSettings
 #include "util/PathUtil.h"                   // for getLongPath
 #include "util/PlaceholderString.h"          // for PlaceholderString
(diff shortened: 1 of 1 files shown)


rolandlo:
The crash is fixed with this PR. I don't like so much, that after the message
grafik
the LaTeX dialog shows up
grafik
with a new failure message like above after typing a new character

This behaviour was introduced before merging the 
release-1.1 branch with 5e439f58002058041fd8fa936b747b954a79b096 after releasing version 1.1.2. 

@Technius Should I only fix the crash with this PR or should I try to restore the old behavior when the LaTeX dialog didn't pop up when the LaTeX generator was not found?

rolandlo:

PR #4466, opened 2022-12-01
Fix copying images into clipboard

rolandlo:
This fixes copying in  `svg` format into the clipboard and copying images from selections that are translated

Check by pasting into a drawing app like Inkscape
or by using `wl-paste --type image/svg > output.svg` on Wayland

I have also tested with gdk scale factor 2 using `GDK_SCALE=2 GDK_BACKEND=x11 ./xournalpp`


rolandlo:
Since the fix is simple and well tested, I will merge it in 24 hours unless an objection is raised.

rolandlo:
I have also fixed the color for pasted text (which was White, when a tool without color like the selection tool was active). Now the color of the text tool is used, which seems quite logical.

Issue #4456, opened 2022-11-28
Better Presentation Mode


forgottosave:
## Description
A feature *Xournal++* is really lacking is a good **presentation mode**. The current mode does nothing but adjust the zoom.
This issue has already been addressed in PR #2408, but it is outdated by almost *1000* commits and hasn't had any activity for a long time.
This thread here is meant to discuss & keep track of suggestions & implementations regarding this problem.

## TODOs
A collection of ideas that have to be discussed & implemented for a good presentation mode:
- [ ] fix the page zoom and scroll to fit window
- [ ] stop rendering other pages than the selected one
- [ ] add some sort of laser pointer tool
- [ ] automatically hide the sidebar, menubar, etc.

#### Issues
- #2117
- #4340 
#### PRs
- #2408 
- #4434

forgottosave:
I'd love to work on some of these issues, as soon as they are approved to help build a better presentation mode :)

rolandlo:
I think the laser pointer tool (see #1778 and #1822), although important, is somewhat independent from the rest. The timer for a presentation mode would be nice to have, but not the first priority. The rest of the TODO's look all very desirable and important.

mjg:
Maybe it's good to agree on what purpose this mode serves, or how modes work in general. For example, shown/hidden elements can be adjusted in config:
```
<property name="presentationHideElements" value="mainMenubar,sidebarContents"/>
```
I have not checked whether leaving the mode resets those to the previous state or unhides them. I guess the real question is: Do we switch between modes (one being "default"), or can one use them together or even no mode at all?

E.g., "fullscreen" sounds as if it does just that (switch to fullscreen), when indeed it is a mode.

It might be worthwhile to think about modes as a preset of several things (fullscreen or not, fixed canvas or scrolling, zoom fixed at full page or some default, sidebars to be shown, maybe even the UI layout). This would allow everyone to have "their" presentation mode. It also means that missing presentation functionality should be made dependent on a setting, not the mode.

forgottosave:
In my current workin-progrees for this issue I implement it, so that the menu-hiding is actually dependent on the settings.
But I have to agree with you (@mjg), we could generalize it to every mode being a preset of those listed things. There could be default presets for the default- and presentation-mode (and other important ones) for anyone not wanting to configure anything. But it would also leave the option of setting additional custom presets as well.
What do others (especially maintainers) think of this idea?

forgottosave:
[...]

PR #4480, opened 2022-12-05
Leave Fullscreen / Presentation mode

forgottosave:
Solves one task from #4456
- [x] be able to leave presentation mode (currently it is not possible to end it properly, meaning leaving fullscreen mode, properly bringing back menu items)

The accelerator `F5` previously used to enter the presentation mode now also leaves it. Using `esc` to leave the presentation mode (if there is no selection to cancel) only seems confusing.

Also I removed some further code for leaving fullscreen in `XournalView::onKeyPressEvent`, which doesn't seem to do anything. Again using esc only seems confusing here (e.g. editing stuff in fullscreen and accidentally leaving it while cancelling a selection).

*sry for this PR looking kinda messy with all the force-pushing etc...*

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -410,6 +410,8 @@ void Control::updatePageNumbers(size_t page, size_t pdfPage) {
 }
 
 void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton* toolbutton, bool enabled) {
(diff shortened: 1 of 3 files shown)


forgottosave:
The commit history and branch will be cleaned up when finished

forgottosave:
Even though this PR looks very messy, it is only a simple change and `Ready to review`.

bhennion (approved):
LGTM! Merging in 24 hours unless an obstruction is raised

PR #4493, opened 2022-12-09
Remove rerender padding

hrdl-github:
As discussed with @bhennion (https://github.com/xournalpp/xournalpp/discussions/4071). Evolution of https://github.com/xournalpp/xournalpp/pull/4492.

Padding can be removed without introducing rendering artifacts. Removing padding completely causes artifacts when drawing ellipses.

Keeping the area to be redrawn small speeds up xournal++ when used with electrophoretic displays.

Tested with most tools on three zoom levels and three line widths. More extensive testing is welcome.

### DIFF
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -34,10 +34,18 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
      * For example, if rect.x = m + 0.9, rect.width = n + 0.2 and ratio = 1 and m and n are integers
      * We need a mask of width n+2 pixels for that...
      **/
-    const auto x = std::floor(rect.x * ratio);
-    const auto y = std::floor(rect.y * ratio);
-    const auto width = int(std::ceil(rect.width * ratio)) + 1;
-    const auto height = int(std::ceil(rect.height * ratio)) + 1;
+    constexpr int RENDER_PADDING_X_Y = 1;
+    constexpr int RENDER_PADDING_WIDTH_HEIGHT = 2;
+
+    const auto rx = rect.x - RENDER_PADDING_X_Y;
+    const auto ry = rect.y - RENDER_PADDING_X_Y;
+    const auto rwidth = rect.width + RENDER_PADDING_WIDTH_HEIGHT;
+    const auto rheight = rect.height + RENDER_PADDING_WIDTH_HEIGHT;
+
+    const auto x = std::floor(rx * ratio);
+    const auto y = std::floor(ry * ratio);
+    const auto width = int(std::ceil(rwidth * ratio)) + 1;
+    const auto height = int(std::ceil(rheight * ratio)) + 1;
 
     xoj::util::CairoSurfaceSPtr rectBuffer(cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height),
                                            xoj::util::adopt);
@@ -51,7 +59,7 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
 
     cairo_set_operator(crPageBuffer.get(), CAIRO_OPERATOR_SOURCE);
     cairo_set_source_surface(crPageBuffer.get(), rectBuffer.get(), 0, 0);
-    cairo_rectangle(crPageBuffer.get(), rect.x, rect.y, rect.width, rect.height);
+    cairo_rectangle(crPageBuffer.get(), rx, ry, rwidth, rheight);
     cairo_fill(crPageBuffer.get());
 }
 
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -30,22 +30,22 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
     const double ratio = view->xournal->getZoom() * this->view->xournal->getDpiScaleFactor();
 
     /**
-     * The +1 makes sure the mask is big enough
-     * For example, if rect.x = m + 0.9, rect.width = n + 0.2 and ratio = 1 and m and n are integers
-     * We need a mask of width n+2 pixels for that...
+     * Padding seems to be necessary to prevent artefacts of most strokes.
+     * These artefacts are most pronounced when using the stroke deletion
+     * tool on ellipses, but also occur occasionally when removing regular
+     * strokes.
      **/
-    constexpr int RENDER_PADDING_X_Y = 1;
-    constexpr int RENDER_PADDING_WIDTH_HEIGHT = 2;
+    constexpr int RENDER_PADDING = 1;
 
-    const auto rx = rect.x - RENDER_PADDING_X_Y;
-    const auto ry = rect.y - RENDER_PADDING_X_Y;
-    const auto rwidth = rect.width + RENDER_PADDING_WIDTH_HEIGHT;
-    const auto rheight = rect.height + RENDER_PADDING_WIDTH_HEIGHT;
+    const auto rx = rect.x - RENDER_PADDING;
+    const auto ry = rect.y - RENDER_PADDING;
+    const auto rwidth = rect.width + 2 * RENDER_PADDING;
+    const auto rheight = rect.height + 2 * RENDER_PADDING;
 
     const auto x = std::floor(rx * ratio);
     const auto y = std::floor(ry * ratio);
-    const auto width = int(std::ceil(rwidth * ratio)) + 1;
-    const auto height = int(std::ceil(rheight * ratio)) + 1;
+    const auto width = static_cast<int>(std::ceil((rx + rwidth) * ratio) - x);
+    const auto height = static_cast<int>(std::ceil((ry + rheight) * ratio) - y);
 
     xoj::util::CairoSurfaceSPtr rectBuffer(cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height),
                                            xoj::util::adopt);
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -836,15 +836,6 @@ double XojPageView::getWidth() const { return page->getWidth(); }
 double XojPageView::getHeight() const { return page->getHeight(); }
 
 void XojPageView::rerenderRect(double x, double y, double width, double height) {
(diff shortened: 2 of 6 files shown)


src/core/gui/PageView.cpp
```diff
@@ -755,8 +755,8 @@ void XojPageView::repaintPage() const { xournal->getRepaintHandler()->repaintPag
 
 void XojPageView::repaintArea(double x1, double y1, double x2, double y2) const {
     double zoom = xournal->getZoom();
-    xournal->getRepaintHandler()->repaintPageArea(this, std::lround(x1 * zoom) - 10, std::lround(y1 * zoom) - 10,
-                                                  std::lround(x2 * zoom) + 20, std::lround(y2 * zoom) + 20);
+    xournal->getRepaintHandler()->repaintPageArea(this, std::lround(x1 * zoom), std::lround(y1 * zoom),
+                                                  std::lround(x2 * zoom), std::lround(y2 * zoom));
```
bhennion, line 759: I think some artefacts are due to using `std::lround`. You should use `std::floor` on `x1` and `y1`, but `std::ceil` on `x2` and `y2` (to make sure the integral rectangle contains the double rectangle).

src/core/gui/PageView.cpp
```diff
@@ -793,10 +793,10 @@ double XojPageView::getWidth() const { return page->getWidth(); }
...
 
 void XojPageView::rerenderRect(double x, double y, double width, double height) {
-    int rx = std::lround(std::max(x - 10, 0.0));
-    int ry = std::lround(std::max(y - 10, 0.0));
-    int rwidth = std::lround(width + 20);
-    int rheight = std::lround(height + 20);
+    int rx = std::lround(std::max(x, 0.0));
+    int ry = std::lround(std::max(y, 0.0));
+    int rwidth = std::lround(width);
+    int rheight = std::lround(height);
```
bhennion, line 799: This whole rectangle preprocessing is useless, I think. I think `rerenderRect` should just pass its arguments to `addRenderRect` and not do anything else.
RenderJob::rerenderRect handles the double coordinates properly (calling std::floor and std::ceil as it should), so there is not need for any of this.
```suggestion
```

bhennion (changes requested):
I tried removing the padding using std::floor/std::ceil as explained in the comment below. I could not see any artefact anymore (using `EllipseHandler` from master, without your changes).
The artefacts you see are due to the use of std::lround instead of std::floor/std::ceil.

src/core/gui/PageView.cpp
bhennion:
```suggestion
    xournal->getRepaintHandler()->repaintPageArea(this, static_cast<int>(std::floor(x1 * zoom)), static_cast<int>(std::floor(y1 * zoom)),
                                                  static_cast<int>(std::ceil(x2 * zoom)), static_cast<int>(std::ceil(y2 * zoom)));
```

src/core/control/tools/EraseHandler.cpp
```diff
@@ -55,10 +55,9 @@ void EraseHandler::erase(double x, double y) {
     for (Element* e: l->getElements()) {
         if (e->getType() == ELEMENT_STROKE && e->intersectsArea(&eraserRect)) {
             eraseStroke(l, dynamic_cast<Stroke*>(e), x, y, range);
+            this->view->elementChanged(e);
```
bhennion, line 58: Actually, I think you augmented the rerendered range here. Before, only the required range was rerendered:

1. In "delete stroke mode", if a stroke was actually untouched by the eraser, it's bounding box was not added to the range.
2. In "normal" mode, only the square below the eraser was rerendered.

src/core/gui/PageView.cpp
```diff
@@ -793,12 +793,7 @@ double XojPageView::getWidth() const { return page->getWidth(); }
 double XojPageView::getHeight() const { return page->getHeight(); }
 
 void XojPageView::rerenderRect(double x, double y, double width, double height) {
-    int rx = std::lround(std::max(x - 10, 0.0));
-    int ry = std::lround(std::max(y - 10, 0.0));
-    int rwidth = std::lround(width + 20);
-    int rheight = std::lround(height + 20);
-
-    addRerenderRect(rx, ry, rwidth, rheight);
+    addRerenderRect(x, y, width, height);
```
bhennion, line 796: I guess here, we could actually remove XojPageView::addRerenderRect entirely (it's not called anywhere else) and move its content to XojPageView::rerenderRect directly.

bhennion (changes requested):

src/core/control/tools/EraseHandler.cpp
bhennion: Mmm, that's weird. Either the range is not large enough (I don't see why), or the repainted rectangle is too small when a Range is passed to view->rerenderRange().

bhennion: I'm currently leaning towards this theory: some amount of padding is necessary because of rounding errors and maybe anti-aliasing.

The rounding errors would come from the two different ways those coordinates are computed: using cairo internals when painting, and using our theoretical bounds otherwise. Two different computation methods would lead to (ever so slightly) different results. Not sure this could actually have a noticeable effect.

Anti-aliasing would make cairo paint stlightly outside of the theoretical bounds, thus our theoretical bound need to be padded when rerendering/repainting.

I'm not so sure about all this, so I need to think a bit more.

bhennion: Yes, I agree. The only thing: the padding should be located in a single function (I guess rerenderRect - so remove the padding in rerenderElement) and the size of the padding should be a constexpr.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -34,10 +34,18 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
...
-    const auto y = std::floor(rect.y * ratio);
-    const auto width = int(std::ceil(rect.width * ratio)) + 1;
-    const auto height = int(std::ceil(rect.height * ratio)) + 1;
+    constexpr int RENDER_PADDING_X_Y = 1;
+    constexpr int RENDER_PADDING_WIDTH_HEIGHT = 2;
+
+    const auto rx = rect.x - RENDER_PADDING_X_Y;
+    const auto ry = rect.y - RENDER_PADDING_X_Y;
+    const auto rwidth = rect.width + RENDER_PADDING_WIDTH_HEIGHT;
+    const auto rheight = rect.height + RENDER_PADDING_WIDTH_HEIGHT;
```
bhennion, line 43:
```suggestion
    const auto rx = rect.x - RENDER_PADDING;
    const auto ry = rect.y - RENDER_PADDING;
    const auto rwidth = rect.width + 2 * RENDER_PADDING;
    const auto rheight = rect.height + 2 * RENDER_PADDING;
```

src/core/control/jobs/RenderJob.cpp
```diff
@@ -34,10 +34,18 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
      * For example, if rect.x = m + 0.9, rect.width = n + 0.2 and ratio = 1 and m and n are integers
      * We need a mask of width n+2 pixels for that...
      **/
-    const auto x = std::floor(rect.x * ratio);
-    const auto y = std::floor(rect.y * ratio);
-    const auto width = int(std::ceil(rect.width * ratio)) + 1;
-    const auto height = int(std::ceil(rect.height * ratio)) + 1;
+    constexpr int RENDER_PADDING_X_Y = 1;
+    constexpr int RENDER_PADDING_WIDTH_HEIGHT = 2;
```
bhennion, line 38:
```suggestion
    constexpr int RENDER_PADDING = 1;
```

bhennion (commented):
I guess the following proposal looks less clumsy.

Also, could you add a comment with RENDER_PADDING to explain why such a padding is (seems) necessary?

rolandlo:
> Adding 2 works fine for me, even though this value can be reduced for all sides but the right one (though not to zero). Do you know what is appropriate here, @rolandlo?

It looks fine in my tests, so I really don't know what is appropriate when it doesn't work for you. The compass looks completely round on all sides. Can you take a screenshot when you see a rendering glitch? Is there a special way to trigger the rendering glitch or does it happen all the time?
I wouldn't mind if you added some more padding to it in #4493 . Probably the same has to be done for the setsquare in that case.

rolandlo:
> In the first picture the blue number is cut off. In the second picture both moving the compass leaves some marks on the right-hand side as well as on the top-hand side. The latter I also observe for the remaining two sides.  

Ok, I see. I only checked the (red) outline, not the (blue) numbers and marks. Adding some padding to the compass definitely makes sense then. Does it render fine now when the radius is 10? Adding a bit less to the setsquare is probably also a good idea.

src/core/model/Compass.cpp
```diff
@@ -24,7 +24,7 @@ auto Compass::getToolRange(bool transformed) const -> Range {
         rg.addPoint(-h, -h);
         rg.addPoint(h, h);
     }
-    rg.addPadding(.5 * xoj::view::CompassView::LINE_WIDTH_IN_CM * CM);  // account for line width
+    rg.addPadding(2 + .5 * xoj::view::CompassView::LINE_WIDTH_IN_CM * CM);  // account for line width
```
rolandlo, line 27: I would suggest to replace that number `2` by a constexpr as well and add a comment mentioning the digits and marks that should not be cut off.

src/core/model/Setsquare.cpp
```diff
@@ -30,7 +30,7 @@ auto Setsquare::getToolRange(bool transformed) const -> Range {
         rg.addPoint(-h, 0);
         rg.addPoint(0, h);
     }
-    rg.addPadding(.5 * xoj::view::SetsquareView::LINE_WIDTH_IN_CM * CM);  // account for line width
+    rg.addPadding(1 + .5 * xoj::view::SetsquareView::LINE_WIDTH_IN_CM * CM);  // account for line width
```
rolandlo, line 33: Here also a constexpr.

rolandlo (approved):
Looks good from my side. @bhennion Anything you still want to have changed?

bhennion (approved):
LGTM. (Squashing and) Merging in 24 hours unless an objection is raised.

rolandlo:
Thanks @hrdl-github for the performance improvement and fixes! I just merged it.

Issue #1863, opened 2020-04-12
Floating toolbox acts unexpectedly with "delete stroke" eraser option


jvwowd:
**Affects versions :**
 - OS: Ubuntu 16.04
 - (Linux only) Desktop environment: X11 with Unity
 - Which version of libgtk do you use: not sure
 - Version of Xournal++: 1.1.0+dev e4fdfef (clean working directory)
 - Installation method: build from source

**Describe the bug**
When using most tools or eraser in another mode, a soft quick click with my stylus on my tablet will bring up the floating toolbox, but this does not happen when the eraser tool is in "delete stroke" mode.

(passive pen on XP-PEN brand Deco 01 V2 usb tablet) 
(X++ "Action on Tool Tap" settings: 
- Enable Tap action: ON
- Try to select object first.: ON
- Show Floating Toolbox: ON
- Ignore Time (ms): 150
- Max Length (mm): 0.70
- Successive (ms): 500

**To Reproduce**
0. Select Floating Toolbar layout
1. Configure settings as above (I have not tried yet if this hold with other settings)
2. Select eraser tool with "delete stroke" mode
3. Soft tap stylus to bring up floating toolbox, but nothing happens.

**Expected behavior**
This should bring up the floating toolbar.

PR #4501, opened 2022-12-11
Unify tap filtering across tools

bhennion:
Various tools are filtering tap events (e.g. to show the floating toolbox).

This PR removes code duplication and makes tap events fully supported by PenInputHandler (which handles every pointing devices -- mouse, stylus, touch).

~~This PR is base on #4159.~~ merged since

### DIFF
--- a/src/core/control/ToolHandler.cpp
+++ b/src/core/control/ToolHandler.cpp
@@ -577,6 +577,18 @@ auto ToolHandler::isSinglePageTool() const -> bool {
            toolType == TOOL_SELECT_PDF_TEXT_RECT;
 }
 
+auto ToolHandler::supportsTapFilter() const -> bool {
+    ToolType toolType = this->getToolType();
+
+    return toolType == TOOL_PEN || toolType == TOOL_HIGHLIGHTER || toolType == TOOL_DRAW_RECT ||
+           toolType == TOOL_DRAW_ELLIPSE || toolType == TOOL_DRAW_COORDINATE_SYSTEM || toolType == TOOL_DRAW_ARROW ||
+           toolType == TOOL_DRAW_DOUBLE_ARROW || toolType == TOOL_DRAW_SPLINE;
+    // Add toolType == TOOL_ERASER
+    // Add? || toolType == TOOL_SELECT_REGION || toolType == TOOL_SELECT_RECT || toolType == TOOL_SELECT_OBJECT  ||
+    // toolType == TOOL_FLOATING_TOOLBOX || toolType == TOOL_SELECT_PDF_TEXT_LINEAR || toolType ==
+    // TOOL_SELECT_PDF_TEXT_RECT
+}
+
 auto ToolHandler::getSelectedTool(SelectedTool selectedTool) const -> Tool* {
     switch (selectedTool) {
         case SelectedTool::active:
--- a/src/core/control/ToolHandler.cpp
+++ b/src/core/control/ToolHandler.cpp
@@ -580,13 +580,9 @@ auto ToolHandler::isSinglePageTool() const -> bool {
 auto ToolHandler::supportsTapFilter() const -> bool {
     ToolType toolType = this->getToolType();
 
-    return toolType == TOOL_PEN || toolType == TOOL_HIGHLIGHTER || toolType == TOOL_DRAW_RECT ||
-           toolType == TOOL_DRAW_ELLIPSE || toolType == TOOL_DRAW_COORDINATE_SYSTEM || toolType == TOOL_DRAW_ARROW ||
-           toolType == TOOL_DRAW_DOUBLE_ARROW || toolType == TOOL_DRAW_SPLINE;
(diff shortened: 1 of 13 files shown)


rolandlo:
Please rebase on master and solve the merge conflicts.

bhennion:
> Please rebase on master and solve the merge conflicts.

Done

src/core/control/ToolHandler.cpp
```diff
@@ -577,6 +577,18 @@ auto ToolHandler::isSinglePageTool() const -> bool {
...
+    ToolType toolType = this->getToolType();
+
+    return toolType == TOOL_PEN || toolType == TOOL_HIGHLIGHTER || toolType == TOOL_DRAW_RECT ||
+           toolType == TOOL_DRAW_ELLIPSE || toolType == TOOL_DRAW_COORDINATE_SYSTEM || toolType == TOOL_DRAW_ARROW ||
+           toolType == TOOL_DRAW_DOUBLE_ARROW || toolType == TOOL_DRAW_SPLINE;
+    // Add toolType == TOOL_ERASER
+    // Add? || toolType == TOOL_SELECT_REGION || toolType == TOOL_SELECT_RECT || toolType == TOOL_SELECT_OBJECT  ||
+    // toolType == TOOL_FLOATING_TOOLBOX || toolType == TOOL_SELECT_PDF_TEXT_LINEAR || toolType ==
+    // TOOL_SELECT_PDF_TEXT_RECT
+}
```
bhennion, line 590: Here is something that should be discussed, I think.

Previously, the tap filter only existed for a handful of tools. If another tool was selected, tapping would not show the floating toolbox.
With this PR, it becomes easy to add more tools to the "tap detecting" list.
Do we want to? If so, which ones?

The question is: do we want to allow more tools?

rolandlo: I think we should. At very least we should add the eraser. I can't see any downside of it. There are no eraser tapping actions, so no overlap. See also #1863 where this has been requested.

In general I would let any tool support the tap filter that doesn't need tap actions by itself. For the selection tools (which also include multilayer rectangle and region tools by now) tap allows to select an object (and maybe on a different layer). For the pdf selection tools tap may allow to follow a link. So that has to be checked carefully.

One thing that is annoying with the floating toolbox is the fact, that once you switch (via the floating toolbox) to a tool that does not support the tap filter, you can't get the floating toolbox up that easily. You can't even put it onto a toolbar (even though that could be changed). That would also be an argument for letting more tools support the tap filter.

rolandlo (commented):
One thing that I have noticed:
When you select an object via tap and try to delete it by clicking the red "x" near the selection, you will not delete it, but bring up the floating toolbox (if configured that way in the settings).

src/core/control/ToolHandler.cpp
rolandlo: Experimenting with selection tools and PDF selection tools added to the tap filter, I would advise against adding them. For PDF selection tools it is annoying that you can't just click on a link any more to get the popup displayed. 
For selection tools it currently doesn't work at all (the selection action is still started and you get a crash).
Maybe it's wiser to just add the floating toolbox as a toolbar item via 
```c++    
addCustomItem("FLOATING_TOOLBOX", ACTION_TOOL_FLOATING_TOOLBOX, "floating-toolbox", _("Floating Toolobx"));
```
and by adding `FLOATING_TOOLBOX` to the `Floating toolbox (experimental)]` toolbar in `toolbar.ini` with according changes to `TOOLBAR_INI_HEADER` in `ToolbarModel.cpp`.

bhennion: I guess all tools should support the floating toolbox tap thingy, but only if the tool does not care about the tap: so, if the tap is not on a link, the PDF selection tools should open the floating toolbox; if there is nothing to select at the tap location, the section tools should open the floating toolbox.

It should not be too hard to do, it's just about prioritizing event responses by adding a couple of tool-specific lines in `PageView::onTapEvent()`. It probably goes a bit beyond this PR though.

bhennion:
I tried adding all the tools that do not handle clicks specifically to the list: eraser, vertical tool, hand.
It actually does not quite work out of the box for the eraser and vertical tool. Those tools need an implementation of `onSequenceCancelEvent()`.
For the other tools (selection tools), it needs more work and refactoring.

In the end, I only added the hand tool. I'd suggest we merge this as is.

> When you select an object via tap and try to delete it by clicking the red "x" near the selection, you will not delete it, but bring up the floating toolbox (if configured that way in the settings).

This, and the other problems you pointed out should be taken care of in a future PR (by whoever feels like it). We should make a clear list of the floating toolbox issues and have a dedicated ticket.

rolandlo (approved):
I'm fine with leaving the other floating toolbox issues to another PR. The logic improvement/simplification are what count in the present PR.
Among the issues I would add (just so that I don't forget): Don't pop up the floating toolbox on tap when it is empty. 

So please go ahead.

PR #4502, opened 2022-12-12
Repaint only areas instead of the entire widget

hrdl-github:
Previously the entire widget would be repainted, causing a noticeable delay on electrophoretic displays. Solution suggested by @bhennion here: https://github.com/xournalpp/xournalpp/discussions/4494#discussioncomment-4371802

### DIFF
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -90,11 +90,7 @@ void RenderJob::run() {
 }
 
 void RenderJob::repaintPage() {
-    int x1 = view->getX();
-    int y1 = view->getY();
-    int x2 = x1 + view->getDisplayWidth();
-    int y2 = y1 + view->getDisplayHeight();
-    repaintWidgetArea(view->xournal->getWidget(), x1, y1, x2, y2);
+    repaintWidgetArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
 }
 
 void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -96,7 +96,7 @@ static void repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2)
 }
 
 void RenderJob::repaintPage() const {
-    repaintWidgetArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
+    repaintPageArea(0, 0, view->getWidth(), view->getHeight());
 }
 
 void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) const {
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -98,9 +98,10 @@ void RenderJob::repaintPage() {
 }
 
 void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
+    double zoom = view->xournal->getZoom();
     int x = view->getX();
     int y = view->getY();
-    repaintWidgetArea(view->xournal->getWidget(), std::floor(x + x1), std::floor(y + y1), std::ceil(x + x2), std::ceil(y + y2));
+    repaintWidgetArea(view->xournal->getWidget(), std::floor(zoom * (x + x1)), std::floor(zoom * (y + y1)), std::ceil(zoom * (x + x2)), std::ceil(zoom * (y + y2)));
 }
 
 void RenderJob::repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -101,7 +101,7 @@ void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
     double zoom = view->xournal->getZoom();
     int x = view->getX();
     int y = view->getY();
-    repaintWidgetArea(view->xournal->getWidget(), std::floor(zoom * (x + x1)), std::floor(zoom * (y + y1)), std::ceil(zoom * (x + x2)), std::ceil(zoom * (y + y2)));
+    repaintWidgetArea(view->xournal->getWidget(), x + std::floor(zoom * x1), y + std::floor(zoom * y1), x + std::ceil(zoom * x2), y + std::ceil(zoom * y2));
 }
 
 void RenderJob::repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -91,19 +91,19 @@ void RenderJob::run() {
     }
 }
 
-void RenderJob::repaintPage() {
+static void repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
+    Util::execInUiThread([=]() { gtk_xournal_repaint_area(widget, x1, y1, x2, y2); });
+}
+
+void RenderJob::repaintPage() const {
     repaintWidgetArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
 }
 
-void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
+void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) const {
     double zoom = view->xournal->getZoom();
     int x = view->getX();
     int y = view->getY();
     repaintWidgetArea(view->xournal->getWidget(), x + std::floor(zoom * x1), y + std::floor(zoom * y1), x + std::ceil(zoom * x2), y + std::ceil(zoom * y2));
-}
-
-void RenderJob::repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
-    Util::execInUiThread([=]() { gtk_xournal_repaint_area(widget, x1, y1, x2, y2); });
 }
 
 void RenderJob::renderToBuffer(cairo_surface_t* buffer) const {
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -6,6 +6,7 @@
 #include <vector>   // for vector
 
 #include <cairo.h>  // for cairo_create, cairo_destroy, cairo_...
+#include "gui/widgets/XournalWidget.h"  // for gtk_xournal_repaint_area
 
 #include "control/Control.h"          // for Control
 #include "control/ToolEnums.h"        // for TOOL_PLAY_OBJECT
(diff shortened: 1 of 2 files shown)


src/core/control/jobs/RenderJob.cpp
```diff
@@ -26,6 +26,8 @@ RenderJob::RenderJob(XojPageView* view): view(view) {}
 auto RenderJob::getSource() -> void* { return this->view; }
 
 void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
+    this->view->repaintArea(rect.x, rect.y, rect.x + rect.width, rect.y + rect.height);
```
bhennion, line 29: I think this should go at the end of the function. The repaint should only happen once the buffer is updated.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -67,6 +69,7 @@ void RenderJob::run() {
     const int dpiScaleFactor = this->view->xournal->getDpiScaleFactor();
 
     if (rerenderComplete) {
+        this->view->repaintPage();
```
bhennion, line 72: Same here, at the end of the `if` block.

bhennion (changes requested):

src/core/control/jobs/RenderJob.cpp
bhennion: Yeah, you would usually not see any difference. The RenderJob is run asynchronously, so it may be that `rerenderRectangle` finishes its deed before the repainting actually happens (so you see no difference). But the converse could also happen, leading to the repainting happening too soon.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -52,6 +52,8 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
     cairo_set_source_surface(crPageBuffer.get(), rectBuffer.get(), 0, 0);
     cairo_rectangle(crPageBuffer.get(), rect.x, rect.y, rect.width, rect.height);
     cairo_fill(crPageBuffer.get());
+
+    this->view->repaintArea(rect.x, rect.y, rect.x + rect.width, rect.y + rect.height);
```
Technius, line 56: nit: I think the code is more readable if this line is moved into the loop where the function is being called, so that all of the `view->repaint`-type calls will be in the `run()` function.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +81,10 @@ void RenderJob::run() {
 
         std::lock_guard lock(this->view->drawingMutex);
         std::swap(this->view->crBuffer, newBuffer);
+        this->view->repaintPage();
```
Technius, line 84: I'm not 100% sure, but this may need to be wrapped in `Util::execInUiThread(...)`. The `RenderJob` runs in a separate thread from the GTK thread, and GTK is not thread safe.

Technius: How are you calling `Util::execInUiThread`?

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,32 @@ void RenderJob::run() {
...
+    int x2 = x1 + view->getDisplayWidth();
+    int y2 = y1 + view->getDisplayHeight();
+    repaintWidgetArea(view->xournal->getWidget(), x1, y1, x2, y2);
+}
+
+void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
+    double zoom = view->xournal->getZoom();
+    int x = view->getX();
+    int y = view->getY();
+    repaintWidgetArea(view->xournal->getWidget(), x + std::floor(zoom * x1), y + std::floor(zoom * y1), x + std::ceil(zoom * x2), y + std::ceil(zoom * y2));
```
Technius, line 104: The math doesn't look right to me here.
* `x + std::floor(zoom * x1)` here is the same as `view->getX() + std::floor(zoom * view->getX())` due to how the function is called. However, `view->getX()` is the top-left corner of the page in layout coordinates and does not need to be scaled by the zoom. Shouldn't the correct value be `view->getX()`?
* The x2 parameter is calculated as `x + std::ceil(zoom * x2)` which is equivalent to `view->getX() + std::ceil(zoom * view->getX() + view->getDisplayWidth())`. Shouldn't the correct value be `view->getX() + view->getDisplayWidth()`?

Technius: Oh, sorry, I got confused. I misread the call in `repaintWidgetArea` in `repaintPage` as `repaintPageArea`. This looks fine to me on a second look.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,32 @@ void RenderJob::run() {
...
+        }
     }
+}
+
+void RenderJob::repaintPage() {
+    int x1 = view->getX();
+    int y1 = view->getY();
+    int x2 = x1 + view->getDisplayWidth();
+    int y2 = y1 + view->getDisplayHeight();
+    repaintWidgetArea(view->xournal->getWidget(), x1, y1, x2, y2);
```
Technius, line 97: I think can be simplified to `repaintPageArea(0, 0, view->getWidth(), view->getHeight())`

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,28 @@ void RenderJob::run() {
...
+void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
+    double zoom = view->xournal->getZoom();
+    int x = view->getX();
+    int y = view->getY();
+    repaintWidgetArea(view->xournal->getWidget(), x + std::floor(zoom * x1), y + std::floor(zoom * y1), x + std::ceil(zoom * x2), y + std::ceil(zoom * y2));
+}
 
-    // Schedule a repaint of the widget
-    repaintWidget(this->view->getXournal()->getWidget());
+void RenderJob::repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
```
bhennion, line 103: This is just a helper function. No need to declare it in the header file. Just make it like so (maybe move it a little up to satisfy the compiler).

```suggestion
static void repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
```

src/core/control/jobs/RenderJob.h
```diff
@@ -37,10 +37,11 @@ class RenderJob: public Job {
     void run() override;
 
 private:
-    /**
-     * Repaint the widget in UI Thread
-     */
-    static void repaintWidget(GtkWidget* widget);
+    void repaintPage();
```
bhennion, line 40:
```suggestion
    void repaintPage() const;
```

src/core/control/jobs/RenderJob.h
```diff
@@ -37,10 +37,11 @@ class RenderJob: public Job {
...
+
+    void repaintPageArea(double x1, double y1, double x2, double y2);
```
bhennion, line 42:
```suggestion
    void repaintPageArea(double x1, double y1, double x2, double y2) const;
```

src/core/control/jobs/RenderJob.h
```diff
@@ -37,10 +37,11 @@ class RenderJob: public Job {
...
+
+    void repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2);
```
bhennion, line 44: Static functions that are only used in the source file should not be declared in the header. See comment above.

```suggestion
```

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,28 @@ void RenderJob::run() {
...
+            repaintPageArea(rect.x, rect.y, rect.x + rect.width, rect.y + rect.height);
+        }
     }
+}
+
+void RenderJob::repaintPage() {
+    repaintWidgetArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
+}
+
+void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) {
```
bhennion, line 96:
```suggestion
void RenderJob::repaintPageArea(double x1, double y1, double x2, double y2) const {
```

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,28 @@ void RenderJob::run() {
...
     } else {
-        for (Rectangle<double> const& rect: rerenderRects) { rerenderRectangle(rect); }
+        for (Rectangle<double> const& rect: rerenderRects) {
+            rerenderRectangle(rect);
+            repaintPageArea(rect.x, rect.y, rect.x + rect.width, rect.y + rect.height);
+        }
     }
+}
+
+void RenderJob::repaintPage() {
```
bhennion, line 92:
```suggestion
void RenderJob::repaintPage() const {
```

src/core/control/jobs/RenderJob.cpp
```diff
@@ -79,12 +80,28 @@ void RenderJob::run() {
 
         std::lock_guard lock(this->view->drawingMutex);
         std::swap(this->view->crBuffer, newBuffer);
+        repaintPage();
```
bhennion, line 83: I guess we no longer need the lock when calling repaintPage. Maybe regroup the lock_guard and the call to std::swap in a sub-scope, so the lock gets released ASAP.

bhennion (changes requested):
Some minor code nitpicks. I'll test it for a while to see if I get artefacts.

bhennion (commented):
LGTM. One question: is the SegFault issue gone?

rolandlo:
I tried out the PR. There is an issue with it. When I insert a page, sometimes the background (e.g. grid) is not painted immediately. The page appears blank. When I start drawing on it, only the grid lines near the cursor appear.

src/core/control/jobs/RenderJob.cpp
```diff
@@ -77,14 +78,32 @@ void RenderJob::run() {
...
+        }
     }
+}
+
+static void repaintWidgetArea(GtkWidget* widget, int x1, int y1, int x2, int y2) {
+    Util::execInUiThread([=]() { gtk_xournal_repaint_area(widget, x1, y1, x2, y2); });
+}
+
+void RenderJob::repaintPage() const {
+    repaintWidgetArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
```
Technius, line 99: Maybe I'm missing something, but won't this repaint the area around the top left corner of the canvas widget instead of the top left corner of the page? I think this should be:
```suggestion
    repaintPageArea(view->xournal->getWidget(), 0, 0, view->getWidth(), view->getHeight());
```

Technius (approved):
LGTM. I will test tomorrow (and I imagine @rolandlo may want to give it another try as well). Merging in 24 hours if there are no further objections.

> Should I squash my changes or do you prefer to keep the entire edit history?

I think this change will be cleaner as a single commit. Unless you want to sign your commit(s), we can handle the merging for you using GitHub's squash and merge.

Technius:
By the way, does this change sufficiently speed up the application for your use case?

rolandlo (approved):
It works all fine now (in combination with #4532)

bhennion (approved):

rolandlo:
Thanks @hrdl-github for your valuable contribution. Improving performance like that is a great step forward. I just squashed and merged it.

PR #4448, opened 2022-11-26
GMenu: Use GMenu in some dynamically generated submenus

bhennion:
This is part of #4372 and based on #4441.

This PR refactors the content of the dynamically generated submenus File->Recent Files, View->Toolbar and Plugin so that is uses GMenu and no longer explicitely relies on GtkMenuItem's.

- [x] Merge #4441 and rebase.

PR #4508, opened 2022-12-14
Refactor PageType and use GMenu's in PageType selection menus

### LINKED ISSUE #4142: Invalid PageTypeFormat::Copy assigned to actual Page model

bhennion:
This is part 3 of #4372 and is based on #4448.

With #4448 and this PR, all dynamically generated submenus of the menubar use GMenu's instead of relying on the deprecated GtkMenu interface.

This PR concerns PageType selection:
1. Refactor the PageType selection menus. Those menus appear in three places in the app:
    1. In the menubar Journal->Paper Background
    4. In the toolbar, when opening the dropdown menu associated to the "New Page" button
    5. In the PageTemplateDialog, to select the default background type
2. ~~Make PageTemplateDialog compatible with gtk4~~ moved to #5078

All three occurences of PageType selection menus are refactored. Sadly, not using the GtkMenu interface means those three different occurences share less code than before.
1. Create a GMenu, which is added to the menubar (gtk4 ready)
2. The dropdown menu is replaced by a GtkPopover. (gtk4 ready)
3. The menu in PageTemplateDialog: more complicated. The GtkMenu cannot be replaced by a GtkPopover yet (i.e. before gtk4), as GtkPopovers get cropped to the dialog's size. Some comments are left in the code for the port to gtk4 to share code with 2.

All the menus of course follow the content of pagetemplate.ini.

Here is a screen shot of the new popover (for 2.):
211191860-5ad025c9-6cee-4070-b53d-32c8f92a1cd6

**Edit:** now based on #5002 and #4763 and #5077

- [x] Merge #5002 and rebase
- [x] Merge #4763 and rebase
- [x] Merge #5077 and rebase


bhennion:
Screenshot_20230108_113953

New looks, using GtkToggleButton's instead of GtkRadioButton. This will also be simpler to actually port to GTK 4.

@Chrisimx I found a neat trick to forbid toggle buttons from unclicking themselves. It avoid having to store pointers to the buttons. See
https://github.com/xournalpp/xournalpp/blob/b803f44fc2025820db19fcf16448e982fdc55bd9/src/core/gui/menus/PageTypeSelectionMenuBase.cpp#L57-L62

rolandlo:
This PR has been ready for review for a while now. It's an important step toward moving to Gtk 4. I feel like it's too complex for me to review the code. Too many things I will miss. @xournalpp/core Could one of the more experienced devs review the code?

bhennion:
I'll try to split off the "Refactor PageType to remove the confusing PageType::Copy" part to ease the pain here. This is a bit of work, but this PR is clearly too complicated to review.

Febbe:
@rolandlo @bhennion we should also wait after the release 1.2 to merge this. For the nightly released, it is not so bad when we introduce some minor bugs. Those can and should be fixed quite fast then.

bhennion:
To ease the reviewing process, I split this into 3 PR's: #5077 with preliminary cleanup, this one and #5078 with the porting of the dialog.

src/core/gui/menus/popoverMenus/PageTypeSelectionPopoverGridOnly.cpp
```diff
@@ -0,0 +1,78 @@
...
+        gtk_container_add(GTK_CONTAINER(entry), preview);  // takes ownership of preview
+        if (gridX >= PAGE_TYPES_PER_ROW) {
+            gridX = 0;
+            gridY++;
+        }
+        gtk_menu_attach(gtkMenu, entry, gridX, gridX + 1, gridY, gridY + 1);
+        gridX++;
+    }
+    return GTK_WIDGET(gtkMenu);
+}
```
Febbe, line 61: move to anonymous namespace

src/core/gui/menus/PageTypeSelectionMenuBase.h
```diff
@@ -0,0 +1,56 @@
...
+
+class PageTypeHandler;
+class PageTypeInfo;
+class Settings;
+
+class PageTypeSelectionMenuBase {
+public:
+    PageTypeSelectionMenuBase(PageTypeHandler* typesHandler, const Settings* settings,
+                              const std::string_view& actionName);
+    virtual ~PageTypeSelectionMenuBase() = default;
```
Febbe, line 31: It looks like, this class should never be destructed directly, make the destructor protected and non virtual.

src/core/gui/toolbarMenubar/ToolButton.cpp
```diff
@@ -82,19 +88,42 @@ auto ToolButton::newItem() -> GtkToolItem* {
...
                     GTK_TOOL_BUTTON(it), gtk_image_new_from_icon_name(iconName.c_str(), GTK_ICON_SIZE_SMALL_TOOLBAR));
+            gtk_tool_button_set_label(GTK_TOOL_BUTTON(it), description.c_str());
         }
     } else {
         if (popupMenu) {
-            it = gtk_menu_tool_button_new(gtk_image_new_from_icon_name(iconName.c_str(), GTK_ICON_SIZE_SMALL_TOOLBAR),
-                                          description.c_str());
-            gtk_menu_tool_button_set_menu(GTK_MENU_TOOL_BUTTON(it), popupMenu.get());
+            if (GTK_IS_POPOVER(popupMenu.get())) {
+                // Why must I put GTK_ICON_SIZE_LARGE_TOOLBAR to get the same size as the other icons???
```
Febbe, line 96: Probably, because the box is not homogeneous.

src/util/include/util/raii/GVariantSPtr.h
```diff
@@ -0,0 +1,44 @@
...
+
+inline namespace raii {
+namespace specialization {
+
+class GVariantHandler {
+public:
+    constexpr static auto ref = g_variant_ref;
+    constexpr static auto unref = g_variant_unref;
+    constexpr static auto ref_sink = g_variant_ref_sink;
+    constexpr static auto adopt = g_variant_take_ref;
```
Febbe, line 32: This is c++20 code, but we use c++17. Either replace constexpr with inline and make auto a const, or wrap those functions into a lambda.

Febbe (approved):

bhennion:
Thanks a lot @Febbe  for the review!!

PR #4511, opened 2022-12-17
Stop rendering non-selected pages in presentation mode

forgottosave:
Solves the rendering TODO in #4456 

When in presentation mode the unnecessary non-selected pages won't be rendered anymore.

### DIFF
--- a/src/core/gui/widgets/XournalWidget.cpp
+++ b/src/core/gui/widgets/XournalWidget.cpp
@@ -263,13 +263,20 @@ static auto gtk_xournal_draw(GtkWidget* widget, cairo_t* cr) -> gboolean {
     // Add a padding for the shadow of the pages
     Rectangle clippingRect(x1 - 10, y1 - 10, x2 - x1 + 20, y2 - y1 + 20);
 
-    for (auto&& pv: xournal->view->getViewPages()) {
+    auto& pages = xournal->view->getViewPages();
+    for (size_t pageId = 0; pageId < pages.size(); pageId++) {
+        auto&& pv = pages.at(pageId);
         int px = pv->getX();
         int py = pv->getY();
         int pw = pv->getDisplayWidth();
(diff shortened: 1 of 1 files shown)


forgottosave:
This PR is ready for the current version, but should be merged **after** #4514 has been reviewed (and potentially merged), as there can be some changes making the mode detection better in combination with this other PR.

rolandlo (changes requested):
The code looks suboptimal to me performance-wise. In presentation mode there is no need to iterate through all pages (or page numbers) and check if it is the current page. Just use the current page immediately.

A more serious concern: in paired pages mode (combined with presentation mode), not only the current page should be rendered, but the page that is paired with it should be rendered as well.

forgottosave:
As for the performance loss in the loop I agree, but that is an easy fix. I didn't implement it for the paired-mode since the presentation mode was meant to work with the full-screen-zoom anyways.
But when the presentation mode gets added as a modifiable mode in #4514, this should probably be just another attribute for modifying modes. This would mean it has to work for all modes.

I'll change that and keep this PR as a draft until it is finished.

bhennion (commented):
I'm not quite sure what you gain from this. The pages outside of the visible area are not rendered anyway:
https://github.com/xournalpp/xournalpp/blob/3971c1a051293f6ef45f68a80ba49b96354062dd/src/core/gui/widgets/XournalWidget.cpp#L269-L271

forgottosave:
There are cases where the zoom-update doesn't work as intended (e.g. when hiding the toolbar) and therefor other pages start to clip through at the bottom. This is very undesirable for the presentation mode.
This PR would ensure, that in any case only the pages are shown which are selected, when using the presentation mode.

Another solution (at least for the zoom-update page clipping) would be to fix the zoom-update and zoom-size of the presented page, which I try to solve in #4434, but am stuck in that process right now.

Issue #2773, opened 2021-02-08
Presentation mode: Hide Menu bar, Hide SIdebar in Preferences is ignored


anastasiusz:
OS: Ubuntu 18.04.05 TLS 4.15.0-135-generic #139-Ubuntu SMP Mon Jan 18 17:38:24 UTC 2021 x86_64 x86_64 x86_64
    (Linux only) Desktop environment: Gnome
    libgtk-3-0: 3.22.30-1ubuntu4
    Xournal++ 1.0.20 ea457v9f Dec 17 2020 21:55:32, from github

**Describe the bug**
Presentation mode ignores Preferences>View: Presentation mode "Hide Sidebar", "HIde Menubar"

**To Reproduce**
In Preferences>View: Presentation mode: Chcek "Hide Sidebar", "HIde Menubar"
View> Presentation mode
Sidebar & Menu bar are still visible

**Expected behavior**
Sidebar & Menu bar should disappear

**Screenshots of Problem**
![Screenshot](https://user-images.githubusercontent.com/43650300/107213094-3536d280-6a08-11eb-9c93-b4845d719f15.png)

**Additional context**
Also the red frame around the PDF should disappear in presentation mode.

rolandlo:
These settings are currently only applied when entering presentation mode via the `F5` key. 
Also see #2408 where the presentation mode is improved (although currently nobody appears to be working on it).

dliessi:
On my system (Fedora 34, Xournal++ 1.1.0), View -> Presentation mode (or Presentation mode button) changes the zoom and disables scrolling and the hand tool; F5 does the same and activates the full screen mode using the full screen hide settings and not the presentation mode hide settings.
Moreover (but that may be a separate issue) only the setting for the menu bar is applied, both with F5 and F11, while the setting for the sidebars is ignored.

(A further separate issue is that F5 does not deactivate the presentation mode.)

forgottosave:
This issue got solved in #4514

The (further separate) issue @dliessi mentioned, with exiting the presentation mode with F5, is still open though.

Issue #1733, opened 2020-02-05
Inhibit screen when in presentation/fullscreen mode


dadosch:
**Is your feature request related to a problem? Please describe.**
It is possible that applications can register via dbus to inhibit the screen, that means, it isn't going to dim or suspend. For example, when Firefox plays a video, the screen isn't dimmed

**Describe the solution you'd like**
It would be nice to see xournalpp inhibit the screen in presentation mode or fullscreen mode. 

**Describe alternatives you've considered**

**Additional context**
https://www.freedesktop.org/wiki/Software/systemd/inhibit/

Technius:
This should be fairly easy to implement: just set the appropriate inhibit flags during application startup.

The problem is that we probably shouldn't make this the default yet, since there are currently some performance/power usage issues (#1451).

dadosch:
I think it would be better to have the inhibition on by default in presentation/fullscreen mode, but off by default in the standard edit mode.

When you have the presentation mode/fullscreen mode, you will likely not want the screen to dim.

Technius:
That's fair. Might be slightly more complicated but it'll still be easy to implement.

It will probably have to wait until after the viewport refactoring is complete.

PR #4514, opened 2022-12-18
Preset view modes

forgottosave:
## Issue
This PR solves the `automatically hide the sidebar, menubar, etc` TODO in #4456 (as well as #2773)

#### This PR
This PR introduces better view-modes. This means the default-, fullscreen- and presentation mode are now refactored to being `view modes` in the settings, with customizable attributes (hide/show widgets & go into fullscreen).

According attributes can be changed in the settings under `View` (where previously similar attributes were shown, but had no effect).

#### Future possibilities
This implementation not only fixes the  `automatically hide the sidebar, menubar, etc` TODO in #4456, but also allows for easy future changes with these modes. Also custom additional modes might be implemented in future.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2895,7 +2895,7 @@ void Control::showAbout() {
     dlg.show(GTK_WINDOW(this->win->getWindow()));
 }
 
-auto Control::loadViewMode(size_t mode) -> bool {
+auto Control::loadViewMode(ViewModeId mode) -> bool {
     if (!settings->loadViewMode(mode)) {
         return false;
     }
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -886,7 +886,7 @@ void Control::actionPerformed(ActionType type, ActionGroup group, GtkToolButton*
             break;
 
         case ACTION_FULLSCREEN:
-            setFullscreen(enabled);
+            setViewFullscreenMode(enabled);
             break;
 
         case ACTION_TOGGLE_PAIRS_PARITY: {
@@ -1614,6 +1614,14 @@ void Control::setViewPairedPages(bool enabled) {
     fireActionSelected(GROUP_PAIRED_PAGES, enabled ? ACTION_VIEW_PAIRED_PAGES : ACTION_NOT_SELECTED);
     win->getXournal()->layoutPages();
     scrollHandler->scrollToPage(getCurrentPageNo());
+}
+
+void Control::setViewFullscreenMode(bool enabled) {
+    if (enabled) {
+        this->loadViewMode(1);
+    } else {
+        this->loadViewMode(0);
+    }
 }
 
 void Control::setViewPresentationMode(bool enabled) {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2898,7 +2898,6 @@ auto Control::loadViewMode(size_t mode) -> bool {
     if (!settings->loadViewMode(mode)) {
         return false;
     }
-    std::cout << "Toolbar=" << settings->isToolbarVisible() << " & sidebar=" << settings->isSidebarVisible() << std::endl;
     if (settings->isMenubarVisible()) { // TODO move if...else into new method in MainWindow.cpp
         gtk_widget_hide(this->win->get("mainMenubar"));
     } else {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2898,6 +2898,7 @@ auto Control::loadViewMode(size_t mode) -> bool {
     if (!settings->loadViewMode(mode)) {
         return false;
     }
+    std::cout << "Toolbar=" << settings->isToolbarVisible() << " & sidebar=" << settings->isSidebarVisible() << std::endl;
     if (settings->isMenubarVisible()) { // TODO move if...else into new method in MainWindow.cpp
         gtk_widget_hide(this->win->get("mainMenubar"));
     } else {
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2882,6 +2882,17 @@ void Control::showAbout() {
     dlg.show(GTK_WINDOW(this->win->getWindow()));
 }
 
+auto Control::loadViewMode() -> bool {
+    if (settings->isMenubarVisible()) { // TODO move if...else into new method in MainWindow.cpp
+        gtk_widget_hide(this->win->get("mainMenubar"));
+    } else {
+        gtk_widget_show(this->win->get("mainMenubar"));
+    }
+    this->win->setToolbarVisible(settings->isToolbarVisible());
+    this->win->setSidebarVisible(settings->isSidebarVisible());
+    return false;
+}
+
 void Control::clipboardCutCopyEnabled(bool enabled) {
     fireEnableAction(ACTION_CUT, enabled);
     fireEnableAction(ACTION_COPY, enabled);
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2898,11 +2898,7 @@ auto Control::loadViewMode(size_t mode) -> bool {
     if (!settings->loadViewMode(mode)) {
         return false;
     }
-    if (settings->isMenubarVisible()) { // TODO move if...else into new method in MainWindow.cpp
-        gtk_widget_hide(this->win->get("mainMenubar"));
-    } else {
-        gtk_widget_show(this->win->get("mainMenubar"));
-    }
+    this->win->setMenubarVisible(settings->isMenubarVisible());
     this->win->setToolbarVisible(settings->isToolbarVisible());
     this->win->setSidebarVisible(settings->isSidebarVisible());
     setFullscreen(settings->isFullscreen());
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -33,6 +33,7 @@
 #include "control/settings/PageTemplateSettings.h"               // for Page...
 #include "control/settings/Settings.h"                           // for Sett...
 #include "control/settings/SettingsEnums.h"                      // for Button
+#include "control/settings/ViewModes.h"                          // for ViewM..
 #include "control/tools/EditSelection.h"                         // for Edit...
 #include "control/xojfile/LoadHandler.h"                         // for Load...
 #include "control/zoom/ZoomControl.h"                            // for Zoom...
@@ -1618,15 +1619,15 @@ void Control::setViewPairedPages(bool enabled) {
 
 void Control::setViewFullscreenMode(bool enabled) {
     if (enabled) {
-        this->loadViewMode(1);
+        this->loadViewMode(VIEW_MODE_FULLSCREEN);
     } else {
-        this->loadViewMode(0);
+        this->loadViewMode(VIEW_MODE_DEFAULT);
     }
 }
 
 void Control::setViewPresentationMode(bool enabled) {
     if (enabled) {
-        this->loadViewMode(2);
+        this->loadViewMode(VIEW_MODE_PRESENTATION);
 
         bool success = zoom->updateZoomPresentationValue();
         if (!success) {
@@ -1635,7 +1636,7 @@ void Control::setViewPresentationMode(bool enabled) {
             return;
         }
     } else {
-        this->loadViewMode(0);
+        this->loadViewMode(VIEW_MODE_DEFAULT);
 
         if (settings->isViewFixedRows()) {
             setViewRows(settings->getViewRows());
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1618,6 +1618,8 @@ void Control::setViewPairedPages(bool enabled) {
 
 void Control::setViewPresentationMode(bool enabled) {
     if (enabled) {
+        this->loadViewMode(2);
+
         bool success = zoom->updateZoomPresentationValue();
         if (!success) {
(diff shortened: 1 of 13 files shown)


src/core/control/Control.h
```diff
@@ -185,6 +186,11 @@ class Control:
     void startDragDropToolbar();
     bool isInDragAndDropToolbar();
 
+    /**
+     * Loads the view mode (hide/show menu-,tool-&sidebar)
+     */
+    bool loadViewMode(size_t mode);
```
jhilmer, line 192: Why not use `enum ViewMode` as argument here? Comment above is not need - function name says it all.

src/core/control/settings/Settings.cpp
```diff
@@ -163,8 +166,10 @@ void Settings::loadDefault() {
     this->buttonConfig[BUTTON_STYLUS_TWO] =
             new ButtonConfig(TOOL_NONE, Colors::black, TOOL_SIZE_NONE, DRAWING_TYPE_DEFAULT, ERASER_TYPE_NONE);
 
-    this->fullscreenHideElements = "mainMenubar";
-    this->presentationHideElements = "mainMenubar,sidebarContents";
+    // view modes
+    this->activeViewMode = ViewMode::VIEW_MODE_DEFAULT;
+    this->viewModes = std::vector<std::string>{"default","fullscreen","presentation"};
+    this->viewModeAttributes =std::vector<std::string>{"showMenubar,showToolbar,showSidebar","showToolbar,showSidebar,goFullscreen","showToolbar,goFullscreen"};
```
jhilmer, line 172: Formatting: Space after `=` and after `,` in argument list

src/core/control/settings/Settings.cpp
```diff
@@ -224,6 +229,42 @@ void Settings::loadDefault() {
     /**/
 }
 
+auto Settings::loadViewMode(size_t mode) -> bool {
```
jhilmer, line 232: Use `enum ViewMode` instaed of `size_t`

src/core/control/settings/Settings.h
```diff
@@ -858,11 +869,17 @@ class Settings {
     ButtonConfig* buttonConfig[BUTTON_COUNT]{};
 
     /**
-     * Which gui elements are hidden if you are in Fullscreen mode,
-     * separated by a colon (,)
+     * View-modes, 0=default, 1=fullscreen, 2=presentation
+     */
+    size_t activeViewMode;
```
jhilmer, line 874: Type should be `enum ViewMode`

src/core/control/settings/Settings.cpp
```diff
@@ -1812,18 +1866,8 @@ auto Settings::getButtonConfig(int id) -> ButtonConfig* {
...
-    this->fullscreenHideElements = std::move(elements);
-    save();
-}
-
-auto Settings::getPresentationHideElements() const -> string const& { return this->presentationHideElements; }
-
-void Settings::setPresentationHideElements(string elements) {
-    this->presentationHideElements = std::move(elements);
-    save();
+void Settings::setViewModeAttributes(size_t mode, std::string attributes) {
```
jhilmer, line 1869: Use `enum ViewMode` instaed of `size_t`

jhilmer (changes requested):
My first review in this project. The view mode is passed around as a `size_t mode` but at the same time you have declared `enum ViewMode` which sound like the natural type those places.

src/core/control/Control.cpp
```diff
@@ -1611,15 +1611,27 @@ void Control::setViewPairedPages(bool enabled) {
     scrollHandler->scrollToPage(getCurrentPageNo());
 }
 
+void Control::setViewFullscreenMode(bool enabled) {
+    if (enabled) {
+        this->loadViewMode(1);
```
bhennion, line 1616: Please use declared constexpr instead of plain numbers. Those really are view mode id's and should be immediately understandable to any coder.

src/core/control/settings/Settings.cpp
```diff
@@ -224,6 +229,42 @@ void Settings::loadDefault() {
...
+    }
+    this->activeViewMode = mode;
+    return true;
+}
+
+auto Settings::getActiveViewMode() const -> bool {
+    return activeViewMode;
+}
+
+auto Settings::getViewModeStrings() const -> std::vector<std::string> {
```
bhennion, line 260: Return a const ref to avoid unecessary copies

```suggestion
auto Settings::getViewModeStrings() const -> const std::vector<std::string>& {
```

src/core/control/settings/Settings.cpp
```diff
@@ -224,6 +229,42 @@ void Settings::loadDefault() {
...
+    return viewModes;
+}
+
+auto Settings::getViewModeAttributes() const -> std::vector<std::string> {
```
bhennion, line 264: same here
```suggestion
auto Settings::getViewModeAttributes() const -> const std::vector<std::string>& {
```

bhennion (changes requested):
I have one main concern: I think the view mode attribute list should be a fixed struct with predefined named flags (e.g. bool showMenubar = true, and so on).
Using strings is both a slight performance loss and error prone (typos in those strings).

The way this structure is converted into string for the benefits of config files should then be concentrated in a single file, where every string like "showMenubar" is defined and used as a constexpr (to avoid typos)

bhennion:
@forgottosave Could you confirm that this should fix #2773 as well?
Also, #1733 could be dealt with in here too (by adding and using relevant flags in the attributes list).

src/core/control/Control.cpp
forgottosave: Oh ,I thought I had replaced all those occurrences. I'll fix this one!

forgottosave:
Yes, this solves #2773 as well. (there was no line of code dedicated to using the config file for this until now, even though the config file entry existed...).

Also I'll take a look into adding #1733 and insert it into the attribute-list.

I also thought about putting the view mode attribute list into a separate struct or class, as it would lead to a better structure and better future changeability (e.g. adding attributes). I can refactor this easily.

forgottosave:
Do you think it could be useful to store the different view-modes in a `std::unordered_map`, instead of different variables? This would later allow to easily add additional custom modes.

src/core/control/Control.h
forgottosave: A future PR could add the possibility to be able to add custom modes. In that case the method should accept any mode-id (not only the default-modes defined in `enum ViewMode`).

But this gets refactored in my next commit anyway.

src/core/control/settings/Settings.cpp
forgottosave: Probably happened when copy-pasting. Will fix it, so the formatting fits again.

src/core/control/settings/Settings.cpp
forgottosave: Same as first [change-request](https://github.com/xournalpp/xournalpp/pull/4514#discussion_r1057764129)

src/core/control/settings/Settings.h
forgottosave: Same as first [change-request](https://github.com/xournalpp/xournalpp/pull/4514#discussion_r1057764129)

src/core/control/settings/Settings.cpp
forgottosave: Same as first [change-request](https://github.com/xournalpp/xournalpp/pull/4514#discussion_r1057764129)

forgottosave:
I now refactored the view modes to be a separate struct. When the settings.xml is being loaded the settings will be parsed into the according structs and vice versa for saving the settings. This leads to a better structure and a slightly improved runtime. In future I think the struct and settings could also hold the view mode name, so adding custom modes or renaming modes is possible.

Please tell me if you want to have some further changes :)

rolandlo:
@forgottosave Updating the authors list is taken care of in #4534. We update the list of authors before every major release (not in between). It's probably easier if you remove the commit adding your name from the PR, since it might create merge conflicts with #4534. You can check if your name/email address was added correctly there.

forgottosave:
Okay, thank you. Removed the commit from this branch :)

forgottosave:
Sorry for the `request feedback` flood. I was mildly confused by Github

bhennion (changes requested):
A couple of coding nitpicks. Otherwise, it look good!

src/core/gui/dialog/SettingsDialog.cpp
```diff
@@ -523,34 +523,23 @@ void SettingsDialog::load() {
...
-            hidePresentationSidebar = true;
-        }
-    }
-
-    loadCheckbox("cbHideFullscreenMenubar", hideFullscreenMenubar);
-    loadCheckbox("cbHideFullscreenSidebar", hideFullscreenSidebar);
-    loadCheckbox("cbHidePresentationMenubar", hidePresentationMenubar);
-    loadCheckbox("cbHidePresentationSidebar", hidePresentationSidebar);
+    auto activeViewMode = settings->getViewModes().at(VIEW_MODE_FULLSCREEN);
+    bool showFullscreenMenubar = activeViewMode.goFullscreen;
```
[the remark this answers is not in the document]
forgottosave: Oh, this variable name is indeed not fitting its purpose. I'll change it.

src/core/control/settings/ViewModes.h
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#include <string>           // for string
+#include <vector>           // for vector
+#include <unordered_map>    // for unordered_map
+
+// reserved default view mode ids
+constexpr auto VIEW_MODE_DEFAULT = 0;
+constexpr auto VIEW_MODE_FULLSCREEN = 1;
+constexpr auto VIEW_MODE_PRESENTATION = 2;
```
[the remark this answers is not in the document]
forgottosave: I tried to avoid a dedicated enum, because of the future possibility to add custom view-modes.
But I guess when implementing this in a follow up one can change that and having an enum in this PR is probably the prettier solution.

src/core/control/settings/Settings.cpp
```diff
@@ -163,8 +165,20 @@ void Settings::loadDefault() {
...
+    viewModeDefault.showMenubar = true;
+    viewModeDefault.showToolbar = true;
+    viewModeDefault.showSidebar = true;
+    ViewMode viewModeFullsceen;
+    viewModeFullsceen.goFullscreen = true;
+    viewModeFullsceen.showToolbar = true;
+    viewModeFullsceen.showSidebar = true;
+    ViewMode viewModePresentation;
+    viewModePresentation.goFullscreen = true;
+    viewModePresentation.showToolbar = true;
```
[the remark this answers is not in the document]
forgottosave: I can do that

src/core/gui/dialog/SettingsDialog.cpp
```diff
@@ -523,15 +523,15 @@ void SettingsDialog::load() {
...
-    bool showFullscreenMenubar = activeViewMode.goFullscreen;
-    bool showFullscreenToolbar = activeViewMode.showToolbar;
-    bool showFullscreenSidebar = activeViewMode.showSidebar;
-    activeViewMode = settings->getViewModes().at(VIEW_MODE_PRESENTATION);
-    bool showPresentationMenubar = activeViewMode.showMenubar;
-    bool showPresentationToolbar = activeViewMode.showToolbar;
-    bool showPresentationSidebar = activeViewMode.showSidebar;
-    bool goPresentationFullscreen = activeViewMode.goFullscreen;
+    auto viewMode = settings->getViewModes().at(ViewModeId::VIEW_MODE_FULLSCREEN);
+    bool showFullscreenMenubar = viewMode.goFullscreen;
```
bhennion, line 527: I'm still confused by this line. Shouldn't it be `viewMode.showMenubar`?

src/core/control/Control.cpp
```diff
@@ -2871,6 +2884,17 @@ void Control::showAbout() {
     dlg.show(GTK_WINDOW(this->win->getWindow()));
 }
 
+auto Control::loadViewMode(size_t mode) -> bool {
```
bhennion, line 2887:
```suggestion
auto Control::loadViewMode(ViewModeId mode) -> bool {
```
And similarly for all other occurences

bhennion (changes requested):

src/core/control/Control.cpp
forgottosave: Okay, I can do that

src/core/gui/dialog/SettingsDialog.cpp
forgottosave: Oh my, I only now saw what you meant xD Sorry about that. I'll fix this right away...

src/core/control/settings/Settings.h
```diff
@@ -858,11 +868,10 @@ class Settings {
     ButtonConfig* buttonConfig[BUTTON_COUNT]{};
 
     /**
-     * Which gui elements are hidden if you are in Fullscreen mode,
-     * separated by a colon (,)
+     * View-modes. Predefined: 0=default, 1=fullscreen, 2=presentation
      */
-    std::string fullscreenHideElements;
-    std::string presentationHideElements;
+    size_t activeViewMode;
```
bhennion, line 873:
```suggestion
    ViewModeId activeViewMode;
```

bhennion (approved):
With this last nitpick, it'll LGTM!

bhennion:
Also, could you rebase on current master and resolve the conflicts?

forgottosave:
Implemented last nitpick and rebased on current master.

bhennion:
Alright, thanks for the good work!

Issue #2843, opened 2021-02-22
Display a horizontal line when using Vertical Space tool


redtrumpet:
Hey, I think it would be useful if the Vertical Space tool would display a thin horizontal line, so that one sees if all written letters will be selected.

Sometimes when a line is long I miss some letters, and have to start over.

Technius:
This can be done as part of #3334.

rodrigo103:
The PR is merged. Should this issue be closed?

rolandlo:
I think this issue was not addressed in the PR. So the issue should stay open.

WillNilges:
@rolandlo What do you mean by that? Do you think that a horizontal line should be displayed even before a mouse click?

redtrumpet:
@WillNilges Yes, that was the intention of my original issue. To have a kind of "aiming guide".

WillNilges:
Gotcha. I'll poke at this and see what I can do.

WillNilges:
Is there another tool that draws a kind of guide in this way? My current approach, I think, is to simply make a 1px selection box appear at the cursor's position when the tool is active, but that doesn't feel quite right. I figured out how to change the mouse cursor as well.

WillNilges:
Or I might be able to specify a fancy css cursor as stated in [`XournalppCursor.cpp`](https://github.com/xournalpp/xournalpp/blob/0ed5ed1f2da7a17a7a6371c95e57acbffeffba05/src/core/gui/XournalppCursor.cpp)?

WillNilges:
I don't think a cursor is going to work. All I can get it to do so far is draw a really wide line over the window, which isn't bound to the document. With a little more work I got it to draw it starting at the mouse position going to the right side of the screen, but for some reason it won't let me draw anything to the left of the cursor.

WillNilges:
@rolandlo @Technius I think I got a rough draft of a solution. Basically, I made a new cursor. Take a look, if you have a chance?
https://github.com/xournalpp/xournalpp/pull/4525

PR #4525, opened 2022-12-24
Add new cursor for vertical space

WillNilges:
First take on vertical space cursor. Uses width of GTK window and cairo drawing to show where the vertical space will start.

### DIFF
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -525,6 +525,7 @@ auto XournalppCursor::createHorizontalLineCursor() -> GdkCursor* {
     GtkWidget* theWidget = control->getWindow()->getXournal()->getWidget();
     GdkWindow* theWindow = gtk_widget_get_window(theWidget);
 
+    // Get the position of the mouse cursor within the window
     gint x, y;
     GdkDevice *mouse_device;
 
@@ -543,22 +544,21 @@ auto XournalppCursor::createHorizontalLineCursor() -> GdkCursor* {
     if (page) {
         Range range = page->getVisiblePart();
         double zoom = control->getZoomControl()->getZoom();
-        const int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        const auto width = static_cast<int32_t>(std::round(range.getWidth() * zoom));
         const int height = 5; // Five is the default cursor size used elsewhere in the app.
-        gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
+        auto dx = (int32_t) (x - static_cast<int32_t>(range.getX() * zoom) - page->getX());
         // 64-bit field containing the current value of dx and width to check
         // if we need to re-draw the cursor
-        auto flavour = (gulong) (dx << (sizeof(gint) * 4) | width);
+        auto flavour = (int64_t) (dx << (sizeof(gint) * 4) | width);
         if (this->currentCursorFlavour == flavour) {
             return nullptr;
         }
 
+        // Re-draw the cursor if necessary 
         xoj::util::CairoSurfaceSPtr crCursor(
             cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height), xoj::util::adopt
         );
-
         xoj::util::CairoSPtr cr(cairo_create(crCursor.get()), xoj::util::adopt);
-
         cairo_set_line_width(cr.get(), 1.2);
         Util::cairo_set_source_rgbi(cr.get(), this->control->getSettings()->getSelectionColor());
         cairo_move_to(cr.get(), 0, 0);
@@ -570,7 +570,7 @@ auto XournalppCursor::createHorizontalLineCursor() -> GdkCursor* {
                 gtk_widget_get_display(theWidget), pixbuf.get(), dx, 0);
         this->currentCursor = CRSR_HORIZONTALLINE;
         // Shift the 'dx' and 'width' variables together to check if we need
-        // to re-draw the cursor.
+        // to re-draw the cursor next time this function is called.
         this->currentCursorFlavour = flavour;
         return gdkCursor; 
     }
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -524,12 +524,6 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
 auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindow* win) -> GdkCursor* {
     GdkWindow* theWindow = gtk_widget_get_window(xournal->getWidget());
     GdkWindow* mainWindow = gtk_widget_get_window(win->getWindow());
-    size_t page = xournal->getCurrentPage();
-    //auto pvRect = xournal->getVisibleRect(page);
-    XojPageView* pv = xournal->getViewFor(page);
-
-    int px = pv->getX();
-    int pw = pv->getDisplayWidth();
 
     gint x, y, winX, winY;
     GdkDevice *mouse_device;
@@ -547,13 +541,36 @@ auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindo
 
     // Five is the default cursor size used elsewhere in the app.
     int height = 5;
-    int pageWidth = gdk_window_get_width(theWindow);
-    int windowWidth = gdk_window_get_width(mainWindow);
 
-    cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, windowWidth, height);
-    cairo_t* cr = cairo_create(crCursor);
+    // Courtesy of Roland
+    Layout* const layout = this->control->getWindow()->getLayout();
+    XojPageView* page = layout->getPageViewAt(x, y);
+    if (page) {
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
 
-    if (px + pw < windowWidth) {
+        cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
+        cairo_t* cr = cairo_create(crCursor);
+
+        cairo_set_line_width(cr, 1.2);
+        cairo_set_source_rgb(cr, 100, 0, 0);
+        cairo_move_to(cr, 0, 0);
+        cairo_line_to(cr, width, 0);
+        cairo_close_path(cr);
+        cairo_stroke(cr);
+        cairo_destroy(cr);
+        GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, width, height);
+        cairo_surface_destroy(crCursor);
+        //x = std::clamp(x, 0, windowWidth);
+        GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
+                gtk_widget_get_display(control->getWindow()->getXournal()->getWidget()), pixbuf, dx, 0);
+        g_object_unref(pixbuf);
+        return gdkCursor; 
+    }
+
+    /*if (px + pw < windowWidth) {
         cairo_set_line_width(cr, 1.2);
         cairo_set_source_rgb(cr, 100, 0, 0);
         cairo_move_to(cr, px + (control->getSettings()->isSidebarVisible() ? control->getSettings()->getSidebarWidth() : 0), 0);
@@ -586,7 +603,7 @@ auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindo
                 gtk_widget_get_display(control->getWindow()->getXournal()->getWidget()), pixbuf, winX, 0);
         g_object_unref(pixbuf);
         return gdkCursor;
-    }
+    }*/
     return nullptr;
 }
 
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -545,6 +545,9 @@ auto XournalppCursor::createHorizontalLineCursor() -> GdkCursor* {
         int width = static_cast<int>(std::round(range.getWidth() * zoom));
         int height = 5; // Five is the default cursor size used elsewhere in the app.
         gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
+        if (this->currentCursorFlavour == dx) {
+            return nullptr;
+        }
 
         cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
         cairo_t* cr = cairo_create(crCursor);
@@ -562,6 +565,7 @@ auto XournalppCursor::createHorizontalLineCursor() -> GdkCursor* {
                 gtk_widget_get_display(theWidget), pixbuf, dx, 0);
         g_object_unref(pixbuf);
         this->currentCursor = CRSR_HORIZONTALLINE;
+        this->currentCursorFlavour = dx;
         return gdkCursor; 
     }
     
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -305,7 +305,7 @@ void XournalppCursor::updateCursor() {
             if (this->mouseDown) {
                 setCursor(CRSR_SB_V_DOUBLE_ARROW);
             } else if (this->insidePage) {
-                cursor = createHorizontalLineCursor(xournal);
+                cursor = createHorizontalLineCursor(xournal, win);
                 setCursor(CRSR_HORIZONTALLINE);
             }
         } else if (type == TOOL_SELECT_OBJECT) {
@@ -521,19 +521,16 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
 }
 
 
-auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal) -> GdkCursor* {
+auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindow* win) -> GdkCursor* {
 
     GdkWindow* theWindow = gtk_widget_get_window(xournal->getWidget());
+    GdkWindow* mainWindow = gtk_widget_get_window(win->getWindow());
     size_t page = xournal->getCurrentPage();
     //auto pvRect = xournal->getVisibleRect(page);
     XojPageView* pv = xournal->getViewFor(page);
-    int pvWidth = pv->getDisplayWidth();
-    int pageStart = pv->getX();
 
     int px = pv->getX();
-    int py = pv->getY();
     int pw = pv->getDisplayWidth();
-    int ph = pv->getDisplayHeight();
 
     gint x, y;
     GdkDevice *mouse_device;
@@ -546,29 +543,48 @@ auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal) -> GdkCur
     mouse_device = gdk_device_manager_get_client_pointer (devman);
     #endif
 
-    gdk_window_get_device_position (theWindow, mouse_device, &x, &y, NULL);
+    gdk_window_get_device_position (theWindow, mouse_device, &x, &y, nullptr);
 
     // Five is the default cursor size used elsewhere in the app.
     int height = 5;
-    int width = gdk_window_get_width(theWindow);
+    //int width = gdk_window_get_width(theWindow);
+    int windowWidth = gdk_window_get_width(mainWindow);
 
-    cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
+    cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, windowWidth, height);
     cairo_t* cr = cairo_create(crCursor);
 
-    cairo_set_line_width(cr, 1.2);
-    cairo_move_to(cr, px, 0);
-    cairo_line_to(cr, px + pw, 0);
-    cairo_close_path(cr);
-    cairo_stroke(cr);
-
-    cairo_destroy(cr);
-    GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, width, height);
-    cairo_surface_destroy(crCursor);
-    //x = std::clamp(x, 0, width);
-    GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
-            gtk_widget_get_display(control->getWindow()->getXournal()->getWidget()), pixbuf, x, 0);
-    g_object_unref(pixbuf);
-    return gdkCursor;
+    if (px + pw < windowWidth) {
+        cairo_set_line_width(cr, 1.2);
+        cairo_set_source_rgb(cr, 255, 0, 255);
+        cairo_move_to(cr, px, 0);
+        cairo_line_to(cr, px + pw, 0);
+        cairo_close_path(cr);
+        cairo_stroke(cr);
+        cairo_destroy(cr);
+        GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, windowWidth, height);
+        cairo_surface_destroy(crCursor);
+        x = std::clamp(x, 0, windowWidth);
+        GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
+                gtk_widget_get_display(control->getWindow()->getXournal()->getWidget()), pixbuf, x, 0);
+        g_object_unref(pixbuf);
+        return gdkCursor;
+    } else {
+        cairo_set_line_width(cr, 1.2);
+        cairo_set_source_rgb(cr, 0, 255, 0);
+        cairo_move_to(cr, px, 0);
+        cairo_line_to(cr, windowWidth, 0);
+        cairo_close_path(cr);
+        cairo_stroke(cr);
+        cairo_destroy(cr);
+        GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, windowWidth, height);
+        cairo_surface_destroy(crCursor);
+        x = std::clamp(x, 0, windowWidth);
+        GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
+                gtk_widget_get_display(control->getWindow()->getWindow()), pixbuf, x, 0);
+        g_object_unref(pixbuf);
+        return gdkCursor;
+    }
+    return nullptr;
 }
 
 
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -532,7 +532,7 @@ auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindo
     int px = pv->getX();
     int pw = pv->getDisplayWidth();
 
-    gint x, y;
+    gint x, y, winX, winY;
     GdkDevice *mouse_device;
 
     #if GTK_CHECK_VERSION (3,20,0)
@@ -544,10 +544,11 @@ auto XournalppCursor::createHorizontalLineCursor(XournalView* xournal, MainWindo
     #endif
 
     gdk_window_get_device_position (theWindow, mouse_device, &x, &y, nullptr);
+    gdk_window_get_device_position (mainWindow, mouse_device, &winX, &winY, nullptr);
 
     // Five is the default cursor size used elsewhere in the app.
     int height = 5;
-    //int width = gdk_window_get_width(theWindow);
+    int pageWidth = gdk_window_get_width(theWindow);
     int windowWidth = gdk_window_get_width(mainWindow);
 
     cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, windowWidth, height);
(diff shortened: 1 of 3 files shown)


WillNilges:
Closes #2843

WillNilges:
Just gonna set this as "Ready for Review" because well, I guess it already was :P

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,44 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+    GdkSeat *seat = gdk_display_get_default_seat (gdk_display_get_default ());
+    mouse_device = gdk_seat_get_pointer (seat);
+    #else
+    GdkDeviceManager *devman = gdk_display_get_device_manager (gdk_display_get_default ());
+    mouse_device = gdk_device_manager_get_client_pointer (devman);
+    #endif
+
+    gdk_window_get_device_position (theWindow, mouse_device, &x, &y, NULL);
+
+    int height = size;
```
rolandlo, line 537: Why don't you call the function parameter `height` to start with? Not sure it should be a function parameter rather than a constant anyway.

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,44 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+    int width = gdk_window_get_width(theWindow);
```
rolandlo, line 538: That only seems to work if all pages are fully displayed in the window. Otherwise the cursor gets too large. Just zoom in or add some pages (displayed in 1 row).

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,44 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+
+    int centerY = height / 2;
+    cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
+    cairo_t* cr = cairo_create(crCursor);
+
+    cairo_set_line_width(cr, 1.2);
+    cairo_move_to(cr, 0, centerY);
+    cairo_line_to(cr, width, centerY);
```
rolandlo, line 546: Why not just use y-coordinate 0? Then you don't have to translate it later on when creating the cursor from the pixbuf.

src/core/gui/XournalppCursor.cpp
```diff
@@ -301,6 +301,11 @@ void XournalppCursor::updateCursor() {
         } else if (type == TOOL_VERTICAL_SPACE) {
             if (this->mouseDown) {
                 setCursor(CRSR_SB_V_DOUBLE_ARROW);
+            } else {
+                setCursor(CRSR_TCROSS);
```
rolandlo, line 305: I don't see the point of this.
```suggestion
```

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,44 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+    cairo_t* cr = cairo_create(crCursor);
+
+    cairo_set_line_width(cr, 1.2);
+    cairo_move_to(cr, 0, centerY);
+    cairo_line_to(cr, width, centerY);
+    cairo_close_path(cr);
+    cairo_stroke(cr);
+
+    cairo_destroy(cr);
+    GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, width, height);
```
rolandlo, line 551: The same pixbuf is created over and over. Use some caching like for other cursors (using the `currentCursor` and `currentCursorFlavour` variables).

rolandlo (changes requested):
I like the idea, but there are several issues with the implementation.

WillNilges:
Finally back from the woods, thanks for the review :D Will address issues ASAP.

src/core/gui/XournalppCursor.cpp
WillNilges: I don't think I understand. I thought that by using centerY, the cursor would appear centered over where your mouse actually is. I suppose I don't see any reason _not_ to use 0, though.

src/core/gui/XournalppCursor.cpp
WillNilges: Looks like I'm gonna have to add to an enum somewhere? Working on it...

src/core/gui/XournalppCursor.cpp
WillNilges: Oof, yeah, you're right. I need to get the width of just that main view, or maybe the width of the visible pixels? I see that there's a way to get the visible document coordinates... can that be somehow converted to screen pixels?

WillNilges: I also tried getting the zoom factor, but I'm afraid those calculations may be too unreliable.

WillNilges: Yep, I think the best way to do this would be if I could somehow get the coordinates of the screen that the currently active page takes up and draw a line across _just that_ at the coords of the cursor. I'm like... 40% sure that's possible.

WillNilges: Getting there. I need to figure out if I can somehow clamp my cairo draw starting point at the edge of the window, or the edge of the sidebar.

WillNilges: I think I need to find the size of the sidebar as well....

WillNilges:
@rolandlo I'm gonna push something that I think isn't very good, but I wanted to show people. It's _getting there_, but I still need to figure out, essentially, how to only draw over the document. Ideally, I'd like to get the screen coordinates of the document view and just paint that, but I haven't been able to figure out how to do that, so I'm trying to take the long way around.

My faith in this approach is once again being tested.

WillNilges:
My other thought is to draw whatever overlay appears when you actually click and drag at a height of 1px or whatever, but I'm struggling to find where that actually happens so I can try to make a cursor out of it.

WillNilges:
Two major bugs remain (not to mention coding conventions and the like): The cursor doesn't switch back to default when `this->insidePage` is false, and it doesn't handle zooming in well.

WillNilges:
OK, that should solve the issue of drawing the cursor outside the bounds. Getting warnings, but I'll deal with that later. Now I need to make it not overdraw when zoomed in

WillNilges:
I have no idea what I'm doing lol. If anyone has a better idea of how to do this I would absolutely love to hear it because boy howdy this is _turbo_ not working.

WillNilges:
So like, pretty much everything works until you zoom in. The cursor seems to be drawing on the document view, which is causing the cursor to "spill" out of the window. I guess what I need is like, a visible offset of that view, which I _think_ exists?

src/core/gui/XournalppCursor.cpp
rolandlo: Have you considered using `getVisibleRect`?

WillNilges:
Ok, pretty close now. It works very well, except there's some weirdness when you zoom in real close in paired pages mode. I might be able to fix that with an offset controlled by a boolean, though, like I did the sidebar.

rolandlo:
I'm sorry to say, but a solution that depends on the layout of the main window doesn't look very maintainable. Consider that the layout might be changed at any moment, which would break your code.

WillNilges:
You're right. I don't really like this solution. If it were possible to get the screen coordinates of the document that was _actually_ visible, this would be trivial to do. I was hoping that's what could be achieved with `XojPageView::getX()` and `XojPageView::getDisplayWidth()`, but maybe that was the wrong move. Lemme try re-implementing this with something like `Layout::getVisibleRect()`. That might do what I want.

WillNilges:
Seems that getVisibleRect also doesn't respect the sidebar's existence. Not really sure what to do about that. It's as if the sidebar is like some kind of overlay that only half the program is aware of...

WillNilges:
I'm probably gonna give up on this approach. I wonder if it'd be possible to create a whole new document overlay controlled by the position of the mouse or something. Else maybe there's a way to render the cursor on the document directly? I might need to git blame this stuff and see if I can get in touch with the tool author. 

To be honest, I didn't think this would be so difficult to make this work.

rolandlo:
> Seems that getVisibleRect also doesn't really respect the presence of a sidebar, or, it sort of does? The width looks right, but x does not.

What basically works:

```c++
Layout* const layout = this->control->getWindow()->getLayout();
XojPageView* page = layout->getPageViewAt(x, y);
double zoom = control->getZoomControl()->getZoom();
int width = static_cast<int>(std::round(range.getWidth() * zoom));
gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
```
Then use `width` as before and `dx` in `gdk_cursor_new_from_pixbuf`. 

The case where there is no page under the cursor (i.e. if `!page`) must be treated separately. It does not suffice to return `nullptr` as then the same cursor is kept.

WillNilges:
You are a genius. That works shockingly well. Way less kludgy.

So then, with this basic concept, I still need to do the following
- [ ] Handle the `!page` case
- [ ] Ensure the cursor is caching properly
    - [ ] Solve the warning I'm getting (definitely related) `** (xournalpp:140): WARNING **: 17:59:53.186: CSS Cursor and backup not valid '', ''`
- [ ] Clean up the function and ensure that my OOP ducks are in a row

src/core/gui/XournalppCursor.cpp
WillNilges: Y'know I'm not sure if caching is really gonna do much good here. The cursor needs to be re-drawn pretty much whenever
- the x-position of the mouse changes
- the position of the page changes
- the zoom level changes

There's not a whole lot I can do to "skip" steps, unless we want to save the pixbuf or something, but that seems like too much work for not a whole lot of reward. What do you think?

WillNilges:
@rolandlo What do you think about this? I set the cursor back to the default if we're off page instead of using nullptr. I think I'm understanding how the caching works better, and I think it'd be unnecessarily complicated to try to cache this cursor, because we need to re-draw it in the correct offset every time the cursor moves anyway.

I suppose we could check the position/zoom/etc and return nullptr if it's just like.... _sitting there_. see my reply to your earlier review comment for more details.

Thanks for the help with this, by the way :)

src/core/gui/XournalppCursor.cpp
rolandlo: My idea was to save the pixbuf. I don't know how big the reward is, though.

rolandlo:
> @rolandlo What do you think about this? I set the cursor back to the default if we're off page instead of using nullptr. 

With default you mean `CRSR_ARROW`, right? That would make sense to me. 

I think I'm understanding how the caching works better, and I think it'd be unnecessarily complicated to try to cache this cursor, because we need to re-draw it in the correct offset every time the cursor moves anyway.
> 
> I suppose we could check the position/zoom/etc and return nullptr if it's just like.... _sitting there_. see my reply to your earlier review comment for more details.
> 
> Thanks for the help with this, by the way :)

You're welcome. :)

WillNilges:
Did a bit of testing, and I realized that (assuming I understand the caching correctly) we can probably use the dx variable as our "flavour" and only re-draw the cursor if that value differs. That can save us a pretty good deal of re-draws.

rolandlo:
> Did a bit of testing, and I realized that (assuming I understand the caching correctly) we can probably use the dx variable as our "flavour" and only re-draw the cursor if that value differs. That can save us a pretty good deal of re-draws.

As long as the horizontal line cursor keeps the same width and dx, we can keep it. The flavor should include the width.

WillNilges:
> As long as the horizontal line cursor keeps the same width and dx, we can keep it. The flavor should include the width.

Will
```C++
this->currentCursorFlavour = (gulong) (dx << (sizeof(gint) * 4) | width);
```
work?

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,68 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        int height = 5; // Five is the default cursor size used elsewhere in the app.
+        gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
+        if (this->currentCursorFlavour == (dx << (sizeof(gint) * 4) | width)) {
+            return nullptr;
+        }
+
+        cairo_surface_t* crCursor = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
```
Technius, line 552: nit: use the cairo smart pointers in `util/raii/CairoWrappers.h` to avoid the manual memory management.

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,68 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+        cairo_destroy(cr);
+        GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor, 0, 0, width, height);
+        cairo_surface_destroy(crCursor);
+        GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
+                gtk_widget_get_display(theWidget), pixbuf, dx, 0);
+        g_object_unref(pixbuf);
+        this->currentCursor = CRSR_HORIZONTALLINE;
+        // Shift the 'dx' and 'width' variables together to check if we need
+        // to re-draw the cursor.
+        this->currentCursorFlavour = (gulong) (dx << (sizeof(gint) * 4) | width);
```
Technius, line 570: This is hard to read and error prone. It would be better to define a union like this:
```c++
using HorizontalLineFlavour = union {
  struct {
    int dx : 32;
    int width : 32;
  } data;
  gulong raw;
};
static_assert(sizeof(HorizontalLineFlavour{}.data) == sizeof(HorizontalLineFlavour{}.raw),
              "Size of packed representation of HorizontalLineFlavour does not match size of integer representation");

// retrieve:
gulong myRawFlavour = this->currentCursorFlavour.raw;

// set:
this->currentCursorFlavour = HorizontalLineFlavour{data, width}.raw;
```

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,68 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+
+    Layout* const layout = this->control->getWindow()->getLayout();
+    XojPageView* page = layout->getPageViewAt(x, y);
+    if (page) {
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        int height = 5; // Five is the default cursor size used elsewhere in the app.
+        gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
+        if (this->currentCursorFlavour == (dx << (sizeof(gint) * 4) | width)) {
```
Technius, line 548: I'm not sure if including the dx will be very useful, since most people will not be moving the mouse/stylus in a perfectly vertical line downwards when they use the tool.

WillNilges: So, including dx in the caching decision is important because that number changes depending on where your mouse is horizontally. Because of the way that the cursor is set up, it needs to be redrawn so that the line is visibly over the page each time the cursor moves horizontally.

Honestly, I dunno how effective caching is going to be at all, so I threw together a quick way to tally how many times the cursor was cached vs redrawn in a variety of ways.

*With dx*
Randomly jiggling my Magic Trackpad up and down for a few seconds
```
Caches: 203 Redraws: 543
28% Caching Rate
```

Drawing a few random lines and moving them with my Magic Trackpad
```
Caches: 273 Redraws: 322
45% Caching Rate
```

Writing a few sentences and shapes and moving them with my Wacom Tablet
```
Caches: 686 Redraws: 704
49% Caching Rate
```

I tried to be as natural as possible, but this is _far_ from scientific. Coincidentally, my first day of class is tomorrow, so I could probably run this test during an actual lecture, though it's just syllabus week :P

src/core/gui/XournalppCursor.cpp
WillNilges: Interesting. Do I need to call `surface.release()` instead of `cairo_surface_destroy(surface)` when I'm done with it, or can I just let it go out of scope and let the stack deal with it?

WillNilges: Also, I guess this is something we should update with the rest of this file at some point, because a lot of these functions use surfaces and the like.

src/core/gui/XournalppCursor.cpp
WillNilges: Wow, that's pretty slick.

eatnumber1: FWIW per https://en.cppreference.com/w/cpp/language/union, "It is undefined behavior to read from the member of the union that wasn't most recently written. Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union."

The set example above of `HorizontalLineFlavour{data, width}.raw` is therefore undefined (the initializer writes to the `data` member, while the subsequent access reads from the `.raw` member (also I think you meant `HorizontalLineFlavour{dx, width}.raw`).

I see the `static_assert` making sure the sizes are the same, but that doesn't really address the UB.

FTR here's a compilable version of that example which, unfortunately, produces no warnings despite the UB. https://godbolt.org/z/j7dGcszqn

WillNilges: I didn't know that was UB... I think it's probably fine to do the bitshifting, as long as it's well documented in comments. That seems to be the common pattern in this file anyway (and some of these other ones look a lot more complicated with a whole lot less explanation.

WillNilges:
I got rid of the union, and am now using the same design pattern as the rest of the file whereby I create a `flavour` variable in order to centralize the existence of the hash. I also added a comment to hopefully de-mistify its existence and explain its function and its composition.

src/core/gui/XournalppCursor.h
```diff
@@ -16,6 +16,8 @@
 
 #include "control/tools/CursorSelectionType.h"  // for CursorSelectionType
 #include "gui/inputdevices/InputEvents.h"       // for InputDeviceClass, INP...
+#include "XournalView.h"
+#include "MainWindow.h"
```
bhennion, line 20: Why are those includes needed in the header file?

src/core/gui/XournalppCursor.cpp
bhennion: > Interesting. Do I need to call `surface.release()` instead of `cairo_surface_destroy(surface)` when I'm done with it, or can I just let it go out of scope and let the stack deal with it?

Just let the stack do its deed

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,73 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+
+    gdk_window_get_device_position (theWindow, mouse_device, &x, &y, nullptr);
+
+    Layout* const layout = this->control->getWindow()->getLayout();
+    XojPageView* page = layout->getPageViewAt(x, y);
+    if (page) {
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        int height = 5; // Five is the default cursor size used elsewhere in the app.
```
bhennion, line 546:
```suggestion
        const int width = static_cast<int>(std::round(range.getWidth() * zoom));
        const int height = 5; // Five is the default cursor size used elsewhere in the app.
```

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,73 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+        );
+        cairo_t* cr = cairo_create(crCursor.get());
+
+        cairo_set_line_width(cr, 1.2);
+        Util::cairo_set_source_rgbi(cr, this->control->getSettings()->getSelectionColor());
+        cairo_move_to(cr, 0, 0);
+        cairo_line_to(cr, width, 0);
+        cairo_close_path(cr);
+        cairo_stroke(cr);
+        cairo_destroy(cr);
```
bhennion, line 566: You can use a smart pointer for this as well. See `CairoWrappers.h`

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +520,73 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+        GdkPixbuf* pixbuf = xoj_pixbuf_get_from_surface(crCursor.get(), 0, 0, width, height);
+        crCursor.release();
+        GdkCursor* gdkCursor = gdk_cursor_new_from_pixbuf(
+                gtk_widget_get_display(theWidget), pixbuf, dx, 0);
+        g_object_unref(pixbuf);
```
bhennion, line 571: You can use a smart pointer here as well. See `GObjectSPtr.h`

bhennion (changes requested):
Some nitpicks in the code.

I do have a more concerning issue though. With this PR, when scrolling with the VerticalTool on, I get a total blinking of everything except Xournal++'s window. See at the end of this video (it's not an encoding error...)

https://user-images.githubusercontent.com/17818041/213872506-37665712-d045-4f53-8de1-4d36558cb9ed.mp4

I'm running KDE under X11 (archlinux, up to date). I have not tested other DE's.

WillNilges:
@bhennion Yikes, that's a weird bug. I can spin up some VMs and do some testing later.

bhennion:
> @bhennion Yikes, that's a weird bug. I can spin up some VMs and do some testing later.

I could reproduce on Gnome X11, but not on Gnome Wayland, nor on KDE Wayland. It seems to be an X11-related bug.

bhennion:
Hmm... It turns out, I also encounter this blinking with the eraser tool, on high zoom, so when the custom cursor is huge (still, X11 only).

WillNilges:
Do you think this is a bug with the particular implementation of this feature, or with something deeper within Xournal++?

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +521,71 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+    XojPageView* page = layout->getPageViewAt(x, y);
+    if (page) {
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        const int width = static_cast<int>(std::round(range.getWidth() * zoom));
+        const int height = 5; // Five is the default cursor size used elsewhere in the app.
+        gint dx = x - static_cast<int>(range.getX() * zoom) - page->getX();
+        // 64-bit field containing the current value of dx and width to check
+        // if we need to re-draw the cursor
+        auto flavour = (gulong) (dx << (sizeof(gint) * 4) | width);
```
Technius, line 551: This assumes that width is 32 bits, dx is 32 bits, and flavour is 64 bits. I would make this explicit by assigning `int32_t` to width and dx, and `int64_t` to flavour.

src/core/gui/XournalppCursor.cpp
Technius: This is a common and legal pattern in C, so this is disappointing to learn about.

src/core/gui/XournalppCursor.cpp
Technius: Also, it's not very clear to me whether dx and width are operated over as 64 bit integers in that last expression. Do they need to be cast to a 64 bit integer first?

WillNilges: The `auto` keyword should inherit from the cast (which makes the bit shift explicit), but I can re-assign width and dx.

WillNilges:
Hm. I just noticed that if you quit while unsaved, and you have the horizontal line tool active, the "Are you sure you want to quit?" dialog pops up and freezes the cursor, so it's just a horizontal line until you mouse out of the document view or over the dialog. Is there anything we can do about that?

src/core/gui/XournalppCursor.cpp
```diff
@@ -515,6 +521,71 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
...
+    XojPageView* page = layout->getPageViewAt(x, y);
+    if (page) {
+        Range range = page->getVisiblePart();
+        double zoom = control->getZoomControl()->getZoom();
+        const auto width = static_cast<int32_t>(std::round(range.getWidth() * zoom));
+        const int height = 5; // Five is the default cursor size used elsewhere in the app.
+        auto dx = (int32_t) (x - static_cast<int32_t>(range.getX() * zoom) - page->getX());
+        // 64-bit field containing the current value of dx and width to check
+        // if we need to re-draw the cursor
+        auto flavour = (int64_t) (dx << (sizeof(gint) * 4) | width);
```
Technius, line 552: I think this expression has a few mistakes:
* `dx` is int32_t, so `(dx << (sizeof(gint) * 4))` is of type `int32_t`. `dx` needs to be upcast to `int64_t` first to force the operations to be over 64 bit integers.
* We need `sizeof(decltype(width))` instead of `sizeof(gint)` to match the exact size of `width` in  bytes.
* Modern processors define 1 byte to be 8 bits, so this should be `* 8` and not `* 4`.
```suggestion
        auto flavour = static_cast<int64_t>(dx) << (sizeof(decltype(width)) * 8) | width;
```

Also:
* `this->currentCursorFlavour` is defined as `long`, which is only guaranteed a size of at least 32 bits. It may be possible for `long` to be defined as 32 bits, in which case `flavour` will be truncated. `this->currentCursorFlavour` should probably be refined as a long long.

Technius (changes requested):

src/core/gui/XournalppCursor.cpp
Technius: Sorry, I didn't explain my concern very clearly. See my new review comment.

src/core/gui/XournalppCursor.cpp
WillNilges: Ah, this was kind of why I wanted to avoid hashing it and just saving the value of DX.

I think we could probably re-evaluate the types of the involved variables. You're right, I also messed up the shift, and it's half what it should have been. I played around with a minimal code sample to confirm all this like I should have in the first place.

To the final point, I'd worry about re-defining `this->currentCursorFlavor`, since it's used so many other places with even more dubious bit hacking going on.

So, I think it's fine to explicity cast as an int64_t there, but if the precision of `this->currentCursorFlavor` is an issue, then we might need to re-think this hashing, to see if we can get it to fit in 32 bits while still having sufficiently unique values. All we need from it is, "have these two numbers changed?" A checksum, if you will :)

WillNilges: I did some debugging to see if I can check out what the hash looks like, and i came across this:
```
(gdb) p sizeof(gulong)
$1 = 8
```
I know it's probably not guaranteed, but on my computer, a modern, 64-bit Intel-based ThinkPad, `gulong` is 8 bytes, 64 bits... That's enough, isn't it? I stepped through the process of constructing that number, and it looks like that'd work fine.

My current understanding of the process is this

First, cast dx to a `gulong`, a 64-bit unsigned integer.
```C++
auto flavour = static_cast<gulong>(dx);
```

Next, shift it over by 32 bits
```C++
flavour <<= (sizeof(decltype(width)) * 8);
```

Then, bitwise `or` that with the width value
```C++
flavour |= width;
```

And you have a hash that looks like this.
```
0b00000000_00000000_00000001_01110110_00000000_00000000_00000010_01010011
```
That gets saved as a `gulong` in `this->currentCursorFlavor`, and bob's your uncle.

src/core/gui/XournalppCursor.cpp
Technius: 50% hit rate is a lot higher than I expected, let's leave `dx` in then.

Technius:
Flavor hashing logic LGTM. Before we can merge, I think the other issues that are mentioned above need to be addressed.

src/core/gui/XournalppCursor.cpp
WillNilges: I don't think saving the pixbuf is a worth it. I think any case we'd save that is a case where the new flavour value wouldn't change so we'd skip that step entirely.

WillNilges:
@bhennion I was unable to reproduce the blinking bug on Wayland as well, but I did see a crash when running Arch with LXQT related to the X window system. I also observed it on my other laptop running Rocky 9.1 with GNOME X11.

```
(xournalpp:3465): Gdk-WARNING **: 19:29:36.438: The program 'xournalpp' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadMatch (invalid parameter attributes)'.
  (Details: serial 149807 error_code 8 request_code 139 (RENDER) minor_code 27)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

WillNilges:
I think I fixed it

bhennion:
I can still trigger the overall blinking I described above (https://github.com/xournalpp/xournalpp/pull/4525#pullrequestreview-1264629533).

As I was saying above, I can also trigger this blinking with the eraser tool. There must be something wrong in the way we handle custom cursors. It needs fixing.

WillNilges:
I still can't reproduce it. Quite a strange issue. Could you give me a `neofetch` of your system or something? I can try to reproduce your system on my old laptop and work on it. Does this warrant its own issue/PR?

bhennion:
I can reproduce on a second laptop as well. Same as before. The laptops are both with an up-to-date Archlinux, running Plasma X11. I'm not quite sure what kind of information would be relevant to help reproduce here.

WillNilges:
Yikes. I guess we better open another issue and figure this out. Probably shouldn't merge this PR until we do, right? 😅

bhennion:
I played around with cursors in GtkDemo's examples. The issue comes from a fairly simple fact: cursors too big are not guaranteed to behave. There is a [gdk_display_get_maximal_cursor_size()](https://docs.gtk.org/gdk3/method.Display.get_maximal_cursor_size.html), which in my case returns 128x128 pixels as the max size.
I'm assuming this limit is set for performance reasons.

WillNilges:
That's quite unfortunate. Does this mean that this approach is dead in the water?

rolandlo:
> That's quite unfortunate. Does this mean that this approach is dead in the water?

I think so. I will close the PR for now.

PR #4526, opened 2022-12-25
Fix error handling in "Annotate PDF" menu option

### LINKED ISSUE #3368: Crash after opening non-PDF file with Annotate PDF menu option
**Affects versions :**
 - OS: Linux
 - Desktop environment: X11
 - libgtk version 3.24.30
 - Version of Xournal++: 1.1.0

**Describe the bug**
When a non-PDF file is chosen from the file picker dialog when using the "Annotate PDF" action, the drawing area goes blank and changing tools triggers a crash.

**To Reproduce**
1. Go to File -> Annotate PDF.
2. Change the file type in the file selection dialog from "PDF files" to "All files".
3. Select a non-PDF file (eg. a JPG image file) and open it.
4. Click OK when the error message is given. The drawing area will go blank.
5. Select a different tool (eg. the highlighter tool). The program will crash.

**Expected behavior**
The program would not crash and continue to be functional.

**Additional context**
This does not happen if File -> Open is used instead of File -> Open PDF.

Technius:
Fixes #3368. See the commit description for the details.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2523,7 +2523,6 @@ auto Control::annotatePdf(fs::path filepath, bool /*attachPdf*/, bool attachToDo
 
     const bool res = newDoc->readPdf(filepath, /*initPages=*/true, attachToDocument);
     getCursor()->setCursorBusy(false);
-    getCursor()->updateCursor();
 
     if (!res) {
         // Loading failed, so display the error to the user.
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2506,6 +2506,7 @@ auto Control::annotatePdf(fs::path filepath, bool /*attachPdf*/, bool attachToDo
         return false;
     }
 
+    // Prompt the user for a path if none is provided.
     if (filepath.empty()) {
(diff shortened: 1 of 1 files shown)


src/core/control/Control.cpp
```diff
@@ -2514,34 +2515,47 @@ auto Control::annotatePdf(fs::path filepath, bool /*attachPdf*/, bool attachToDo
...
+    // Update any global metadata...
+    RecentManager::addRecentFileFilename(filepath.c_str());
+
+    newDoc->lock();
+    auto metaFilename = newDoc->getEvMetadataFilename();
+    newDoc->unlock();
+    MetadataEntry md = MetadataManager::getForFile(metaFilename);
+    loadMetadata(md);
+
+    fireDocumentChanged(DOCUMENT_CHANGE_COMPLETE);
```
rolandlo, line 2558: The following suggestion is not only shorter, but it also updates the Window title and handles invalid metadata.
```suggestion
    fireDocumentChanged(DOCUMENT_CHANGE_COMPLETE);
    fileLoaded();
```

Technius:

rolandlo (approved):
LGTM

PR #4528, opened 2022-12-26
Add plugin to fit the page size to the content/selection

atticus-sullivan:
The idea behind this, is to be able to prototype small images for e.g. LaTeX documents (quick and dirty drawing before probably time consuming tikzing the drawing).

Caveat: Only works with strokes, no images and no text boxes are taken into account. Might change in the future, but for this we need get/add api-functions for these types.

Note: Created this as draft since I came up with this just right now. I want to avoid this being merged without thinking about this again. So for me most interesting right now is, what you think about this idea in general. And if you've got ideas regarding the decisions usin
1.  `getStrokes` instead of `toolInfo` and
2. how to handle the padding (reason behind this is we need something to get the whole stroke, as without padding thick strokes are being cut of at the edges)

### DIFF
--- a/plugins/FitToContent/main.lua
+++ b/plugins/FitToContent/main.lua
@@ -1,0 +1,60 @@
+-- Register all Toolbar actions and intialize all UI stuff
+function initUi()
+    app.registerUi{["menu"] = "Fit page to layer", ["callback"] = "fit_to_layer"};
+    app.registerUi{["menu"] = "Fit page to selection", ["callback"] = "fit_to_selection"};
+end
+
+local function fit_to_layer_or_selection(type)
+    if type ~= "selection" and type ~= "layer" then
+        app.msgbox("Invalid type for fitting", {[1]="Ok"})
+        return
+    end
+
+    local strokes = app.getStrokes(type)
+    -- padd the whole region just a little bit. Might be a thing to configure in the future
+    local padding = {x=0, y=0}
+    -- search for the outermost strokes which span the rectangle of the to be moved area
+    -- take the stroke width, or if set the pressure, into account hereby
+    local mima    = {minX=100000, maxX=0, minY=100000, maxY=0}
+    for _,stroke in ipairs(strokes) do
+        if stroke.pressure then
+            for i,x in ipairs(stroke.x) do
+                mima.maxX = math.max(mima.maxX, x + stroke.pressure[i]/2)
+                mima.minX = math.min(mima.minX, x - stroke.pressure[i]/2)
+            end
+            for i,y in ipairs(stroke.y) do
+                mima.maxY = math.max(mima.maxY, y + stroke.pressure[i]/2)
+                mima.minY = math.min(mima.minY, y - stroke.pressure[i]/2)
+            end
+        else
+            for _,x in ipairs(stroke.x) do
+                mima.maxX = math.max(mima.maxX, x + stroke.width/2)
+                mima.minX = math.min(mima.minX, x - stroke.width/2)
+            end
+            for _,y in ipairs(stroke.y) do
+                mima.maxY = math.max(mima.maxY, y + stroke.width/2)
+                mima.minY = math.min(mima.minY, y - stroke.width/2)
+            end
+        end
+    end
+    -- move strokes by minX and minY
+    for _,stroke in ipairs(strokes) do
+        for i,x in ipairs(stroke.x) do
+            stroke.x[i] = x - mima.minX + padding.x
+        end
+        for i,y in ipairs(stroke.y) do
+            stroke.y[i] = y - mima.minY + padding.y
+        end
+    end
+
+    -- make the current layer invisible and add the moved strokes to a new layer
+    app.setLayerVisibility(false)
+    app.layerAction("ACTION_NEW_LAYER")
+    app.setCurrentLayerName("moved for fitting")
+    app.addStrokes{strokes=strokes}
+
+    app.setPageSize(mima.maxX-mima.minX+2*padding.x, mima.maxY-mima.minY+2*padding.y)
+end
+
+function fit_to_layer() fit_to_layer_or_selection("layer") end
+function fit_to_selection() fit_to_layer_or_selection("selection") end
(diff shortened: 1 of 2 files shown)


atticus-sullivan:
I tested this for some images and for me it works fine.

If you've got any thoughts about the padding (fixed, if so which value vs relative to the width/height vs something completely different), please let me know.

rolandlo (commented):
I don't think padding for a fixed number of points is a good idea. If the stroke width is large that won't work. Instead the pressure values have to be taken into account as in `Stroke::calcSize()`. 
If you take the info about the bounding box from `toolInfo` you won't necessarily get the correct answer if the selection or layer includes elements other than strokes.

atticus-sullivan:
Alright, pressure and width were missing in the calculation that's right.
For me it was logical to check the pressure of the min/max setting point of the stroke and not calculating the max pressure of the stroke and apply it to all points (like in `Stroke::calcSize()`). Does this make sense to you?

And I still left a bit of padding (but reduced as the width/pressure is now taken into account) since this is what I'd expect normally.

plugins/FitToContent/main.lua
```diff
@@ -0,0 +1,49 @@
...
+    -- take the stroke width, or if set the pressure, into account hereby
+    local mima    = {minX=100000, maxX=0, minY=100000, maxY=0}
+    for _,stroke in ipairs(strokes) do
+        for i,x in ipairs(stroke.x) do
+            mima.maxX = math.max(mima.maxX, x + (stroke.pressure and stroke.pressure[i] or stroke.width/2))
+            mima.minX = math.min(mima.minX, x - (stroke.pressure and stroke.pressure[i] or stroke.width/2))
+        end
+        for i,y in ipairs(stroke.y) do
+            mima.maxY = math.max(mima.maxY, y + (stroke.pressure and stroke.pressure[i] or stroke.width/2))
+            mima.minY = math.min(mima.minY, y - (stroke.pressure and stroke.pressure[i] or stroke.width/2))
```
rolandlo, line 26: The pressure values must also be divided by 2. Moreover performance would be higher if `stroke.pressure` was only checked once for every stroke and not for every point of the stroke.

rolandlo:
> For me it was logical to check the pressure of the min/max setting point of the stroke and not calculating the max pressure of the stroke and apply it to all points (like in `Stroke::calcSize()`). Does this make sense to you?

Yes, that makes sense to me. The approach in Stroke::calcSize was used for a quick&dirty approximation.

plugins/FitToContent/main.lua
atticus-sullivan: Yes, performance would be higher this way (and it's right that performance in these loops is critical). But in fact we need to check for every point if the pressure value should be taken or the width.

The solution that comes to my mind (apart from duplicating the inner loops) is to work with functions like
```lua
local function padd_with_pressure(stroke, i)
    return stroke.pressure[i]
end
local function padd_without_pressure(stroke)
    return stroke.width
end
```
and then set a variable to this function before going into the inner loops to either of these functions (depending if pressure is present or not).
But I'm not sure, if this approach really increases the performance (and benchmarking is not that easy in this case, I think).

Any ideas on the performance/pressure checking part?

rolandlo: I would simply duplicate the inner loops.

jdujava:
I think there is a slight inconsistency with the implementation of `getStrokes`. 

https://github.com/xournalpp/xournalpp/blob/0d132dfd90d5d039fd923b6289314b306b2ad16e/src/core/plugin/luapi_application.h#L869-L875

For selection it is called with `"selection"` and for layer with `"Layer"`.

Moreover, in the description before the function they are both lowercase.
https://github.com/xournalpp/xournalpp/blob/0d132dfd90d5d039fd923b6289314b306b2ad16e/src/core/plugin/luapi_application.h#L825

atticus-sullivan:
Oh that's right. Looks like we (or mostly I) have missed that when implementing `getStrokes`. Thanks for noting this :+1: . As changes to the api are no subject to this PR, I'll make a new one.

atticus-sullivan:
Rebased on the current master. From my side this PR would be finished if there are no objections.

rolandlo (approved):
The plugin code looks good and works fine for layers/selections consisting only of strokes. 
The name in `plugin.ini` was the real name (not a pseudonym) for all plugins so far. If you don't mind, for consistency change it.
I should also point out that we plan to move all plugins (or at least those which are not absolutely relevant for the functioning of Xournal++) into a separate repository, which can be installed separately. Are you fine with that?

atticus-sullivan:
> The plugin code looks good and works fine for layers/selections consisting only of strokes. The name in `plugin.ini` was the real name (not a pseudonym) for all plugins so far. If you don't mind, for consistency change it.

I don't quite get what you mean by that. Is the name `plugin.ini` not fine? Looking at other plugins (e.g. [LayerActions](https://github.com/xournalpp/xournalpp/tree/master/plugins/LayerActions) or [MigrateFontSizes](https://github.com/xournalpp/xournalpp/tree/master/plugins/MigrateFontSizes)) it's the same naming.
(Don't get me wrong, I've got not problem with renaming the files. I just don't understand your request)

> I should also point out that we plan to move all plugins (or at least those which are not absolutely relevant for the functioning of Xournal++) into a separate repository, which can be installed separately. Are you fine with that?

Yes that's fine with me.

rolandlo:
> I don't quite get what you mean by that. Is the name `plugin.ini` not fine? 

I was talking about your (the author's) name atticus-sullivan is your Github name and I guess your real name is written differently.

atticus-sullivan:
Yes that's true ^^. Done.

So are we at the point where I squash the commits all to one (I think that's enough for this) and merge it afterwards, right?

rolandlo:
> Yes that's true ^^. Done.
> 
> So are we at the point where I squash the commits all to one (I think that's enough for this) and merge it afterwards, right?

Yes, please go ahead.

rolandlo:
As I wrote above, all plugins will be moved to a separate repository at a later stage.

rolandlo:
Thanks for the useful plugin @atticus-sullivan !

Issue #4490, opened 2022-12-08
(Nightly Version) occasional square shaped rendering artifacts while writing.


RichardScharf:
**Affects versions :**
 - OS: Linux Mint 20.3 x86_64
 - Xournal Version 1.1.2+dev, **Git commit:** 52e6a795 **Built on:** Nov 24 2022, 00:05:46
 - with libgtk 3.24.20

**Describe the bug**
Square rendering artifacts randomly appear when writing.
Full (re)rendering delayed by a couple of seconds.

**To Reproduce (seemingly related artifacts)**
Steps to reproduce related behavior **(seen in "Video of Problem")**:
1. Open a new graph page with border.
2. Zoom to 700% to the bottom right corner of the page.
3. Click frequently to draw dots.
4. After a while rendering artifacts start appearing (border is disappearing within square shaped chunks).

**Expected behavior**
Rendering artifacts should not hide the current strokes.

**Video of Problem**
[![Thumbnail](https://user-images.githubusercontent.com/72913455/206469519-c6468cb5-0c64-41d6-bcef-493b8570d949.png)](https://user-images.githubusercontent.com/72913455/206468486-6243c531-41ac-4af5-9338-f9edf79d8dde.mp4)

Technius:
Can't reproduce on the latest nightly build on NixOS 22.11, with gtk 3.24.34 running on Wayland.

Technius:
What are your settings in `Preferences > Input System > Input stabilization`?

RichardScharf:
disabled (none and none)

SaVoAMP:
I also experience this problem during notetaking in lectures, same OS, commit and build.

I am able to reproduce the border behavior as well in ~10 seconds, (on my Thinkpad and on my PC independently).
Drawing near the bottom border using the pen tool results in chunks disappearing, making the grey lines underneath visible for a few seconds at a time.

I use it because the set square is only available in the nightly version.

bhennion:
Just to confirm, none of you have encountered this on the release branch?
Would you be able/willing to bisect and find out what commit caused this?

RichardScharf:
Just compiled the latest nightly build inside a docker container, since there was no precompiled .deb package available under "assets".
First time I ever compiled anything from source! The instructions on how to compile were very helpful and GTKSourceView is also available for LaTeX now (it was disabled during build in the last precompiled version I downloaded). 

Issue seems fixed now, can't reproduce the bug anymore. Thanks!

[...]

PR #4530, opened 2022-12-26
Fix border artefacts when writing close to the edge of a Graph background

bhennion:
Fixes this:

https://user-images.githubusercontent.com/17818041/209571911-38219acd-8f84-4a77-9548-67cb8445d401.mp4


Technius:
Does this fix #4490?

bhennion:
I don't think so: I don't see any reason why this could fix something like #4490 (but I can't know for sure, I failed to reproduce #4490).

tattsan:
#4490  appears even after applying this PR.  It doesn't appear very often, so I can't do bisection.

Technius (approved):
Code LGTM, didn't try it though.

PR #4532, opened 2022-12-27
Fix setsquare activation and deactivation, fix toolRange

rolandlo:
Preparation for #4443 
The rendering issues when activating or deactivating the setsquare fixed in this PR are fully visible when combined with #4502

### DIFF
--- a/src/core/model/Setsquare.cpp
+++ b/src/core/model/Setsquare.cpp
@@ -8,10 +8,13 @@
 
 Setsquare::Setsquare(): Setsquare(INITIAL_HEIGHT, .0, INITIAL_X, INITIAL_Y) {}
 
-Setsquare::Setsquare(double height, double rotation, double x, double y):
-        GeometryTool(height, rotation, x, y, getToolRange(true)) {}
+Setsquare::Setsquare(double height, double rotation, double x, double y): GeometryTool(height, rotation, x, y) {
+    this->lastRepaintRange = getToolRange(true);
+}
 
-Setsquare::~Setsquare() { viewPool->dispatchAndClear(xoj::view::SetsquareView::FINALIZATION_REQUEST, Range()); }
+Setsquare::~Setsquare() {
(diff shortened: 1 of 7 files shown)


rolandlo:

src/core/view/SetsquareView.cpp
```diff
@@ -45,6 +46,9 @@ SetsquareView::~SetsquareView() noexcept { this->unregisterFromPool(); };
 void SetsquareView::on(FlagDirtyRegionRequest, const Range& rg) { this->parent->flagDirtyRegion(rg); }
 
 void SetsquareView::on(UpdateValuesRequest, double h, double rot, cairo_matrix_t m) {
+    if (h <= 0.) {
+        g_warning("Non-positive setsquare height: %f", h);
+    }
```
bhennion, line 51: Does this actually happen? Should this be asserted?

src/core/model/Setsquare.cpp
```diff
@@ -27,7 +30,7 @@ auto Setsquare::getToolRange(bool transformed) const -> Range {
         rg.addPoint(-h, 0);
         rg.addPoint(0, h);
     }
-    rg.addPadding(.5 * xoj::view::SetsquareView::LINE_WIDTH);  // account for line width
+    rg.addPadding(.5 * xoj::view::SetsquareView::LINE_WIDTH * CM);  // account for line width
```
bhennion, line 33: Could you make the unit of SetsquareView::LINE_WIDTH explicit? e.g. add a comment at the declaration point, and maybe rename it to LINE_WIDTH_IN_CM

bhennion (approved):
LGTM

src/core/model/Setsquare.cpp
rolandlo: I renamed it. A comment doesn't look necessary with the renamed constant.

src/core/view/SetsquareView.cpp
rolandlo: No, I don't think it ever happens. I changed it to an assertion.

PR #4533, opened 2022-12-28
Lock document's mutex upon adding stroke to layer in BaseShapeHandler

bhennion:
It's all in the title.

There are many places where the document's mutex is not locked when accessing/modifying the document. This PR simply plugs one of those instances that I forgot to deal with in #4137.


rolandlo (approved):
LGTM

PR #4534, opened 2022-12-28
Update authors list

rolandlo:
Updated the list of authors obtained in raw form via 
```
```
I have removed every member of the developer team (listed on the top) from the list of contributors (listed below) and removed duplicates and bots. 

Here is the diff between AUTHORS_raw and AUTHORS:
[diff.txt](https://github.com/xournalpp/xournalpp/files/10315780/diff.txt)


Technius:
I would also update the dates on the maintenance team, and then update the About dialog text to be consistent with the list of maintainers.

Technius (approved):
We'll probably need to update the author list again when we prepare the 1.2.0 release...

rolandlo:
I will squash the commits and rebase on master after merging #4970 (solving conflicts if necessary) in 24 hours if no objections are raised.

PR #4371, opened 2022-10-29
Update mac jhbuild instructions

Technius:
This PR improves the macOS jhbuild instructions and the mac build readme.
* Move homebrew section to top of the readme
* Add instructions on how to set up jhbuild correctly
* Create a custom jhbuild moduleset for installing 3rd party deps instead of using custom bash scripts
* Add lua package build (not built previous; now it's in the moduleset)

A completely fresh build works locally on my Mac M1 Mini running macOS 12.4 Monterey.

Fixes #3949. Fixes #4222.

PR #4536, opened 2022-12-29
tweak: Improve shell scripts

hyperupcall:
- In the same style as `setup-gdk-pixbuf.sh`, use `#!/usr/bin/env bash` (better practice)
- Improve quoting of variables
- Satisfies the linter


rolandlo (commented):
This PR clashes with  #4371. I think it would be better to request the changes there. Note that most of the scripts get removed there.

hyperupcall:
Sounds good :+1: I'll go ahead and do that

PR #4541, opened 2022-12-29
fix: Color diagnostics now properly show with Ninja

hyperupcall:
When using `Ninja` as the CMake Generator, diagnostics would fail to colorize. This fixes that. Significantly improves first time experience for new contributors.


rolandlo (approved):
LGTM

rolandlo:
Thanks for the improved contributor experience @hyperupcall ! I just merged it.

Issue #4519, opened 2022-12-22
Select Color Dialog Window


kacago22:
- OS: Windows10
 - Which version of libgtk do you use 3.24.35
 - Version of Xournal++:  1.1.3 dev (Dec 18)
 - Installation method: install.exe

It's not really a "bug".  I run xournal on the second screen - when I press the color square at the end of the color selector toolbar to open the Select Color Dialog Window it opens in the upper right corner of the first screen - where I cannot even reach with my stylus (on a Huion Graphical Screen) ...

rolandlo:
The Select Color Dialog opens up on the first screen on Windows for me as well when the main Window is displayed on the second screen.

jhilmer:
Think this is solved with my latest commit to #4543 - the parent attribute was not initilized

jhilmer:
Fix is in master now. Should be closed.

PR #4543, opened 2022-12-30
Hint placement of color select dialog

jhilmer:
I have been looking into issue #4519. I have not be able to reproduce. But by looking at how other dialogs are configured elsewhere, I guess 'gtk_window_set_trasient_for' and perhaps also 'gtk_window_set_position' was missing for the dialog.

As I understand it's up to the windows manager of a given system to place the window accordingly to these hints.

Hopefully somebody with a window build and a dual screen setup can try this out.

### DIFF
--- a/src/core/gui/toolbarMenubar/ColorToolItem.cpp
+++ b/src/core/gui/toolbarMenubar/ColorToolItem.cpp
@@ -23,6 +23,7 @@ ColorToolItem::ColorToolItem(ActionHandler* handler, ToolHandler* toolHandler, G
                              bool selektor):
         AbstractToolItem("", handler, selektor ? ACTION_SELECT_COLOR_CUSTOM : ACTION_SELECT_COLOR),
         namedColor{std::move(namedColor)},
+        parent(parent),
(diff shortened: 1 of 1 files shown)


rolandlo:
Unfortunately that didn't make any difference in my case (on Windows with dual screen).

rolandlo:
Interestingly the color chooser dialog for the background color (in menu `Edit > Paper color`) is displayed correctly. 
https://github.com/xournalpp/xournalpp/blob/f5b299471abded2400f595355742a9c150ce07b6/src/core/gui/dialog/SelectBackgroundColorDialog.cpp#L106-L110

The only difference I see is that it is activated from the menubar and not via the toolbar.

jhilmer:
Please retest now where the parent attribute is initilized

rolandlo (commented):
Yes, the initialization of `parent` fixes the issue. The first commit is not needed however. It doesn't make any difference. I suggest to drop it.

rolandlo (approved):
Thanks for the fix. Closes #4519 

PR #4546, opened 2022-12-30
Catch by const reference

hyperupcall:
This slightly improves the exception handling code by catching by const reference, as per [E.15](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e15-throw-by-value-catch-exceptions-from-a-hierarchy-by-reference).

Two other inconsistently named exceptions `except` and `err` were also renamed to `e`. Some east-consts were converted to west-consts, as this codebase seems to do.


bhennion (approved):
LGTM! Merging in 24 hours unless an objection is raised.

PR #4547, opened 2022-12-30
Removing passing StrokeTool as int to `g_warning`

hyperupcall:
Came across this compiler warning

```txt
.../xournalpp/src/core/control/xojfile/SaveHandler.cpp:110:19: warning: format ‘%i’ expects argument of type ‘int’, but argument 4 has type ‘StrokeTool’ [-Wformat=]
  110 |         g_warning("Unknown stroke tool type: %i", t);
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~
      |                                                   |
      |                                                   StrokeTool
```

This fixes it by removing the warning and opting for a more generic one instead. This also makes the error message more consistent across files.

Thought it would be good to mention that the [luapi_application.h](https://github.com/hyperupcall/xournalpp/blob/a575bfa63acea36b7eb76f306be18e5f6b770f99/src/core/plugin/luapi_application.h#L919) seems to forego a comparison with `StrokeTool::ERASER`, even though it was already defined at the time (git blame), and it prints an "unknown STROKE_TOOL" error. It seemed like fixing it would be a breaking change for Lua plugins so I didn't add it to this commit, but thought it would be good to mention anyways. Maybe this can be fixed in a different PR?


rolandlo:
> Thought it would be good to mention that the [luapi_application.h](https://github.com/hyperupcall/xournalpp/blob/a575bfa63acea36b7eb76f306be18e5f6b770f99/src/core/plugin/luapi_application.h#L919) seems to forego a comparison with `StrokeTool::ERASER`, even though it was already defined at the time (git blame), and it prints an "unknown STROKE_TOOL" error. It seemed like fixing it would be a breaking change for Lua plugins so I didn't add it to this commit, but thought it would be good to mention anyways. Maybe this can be fixed in a different PR?

Don't worry about breaking Lua plugins with adding a comparison with `StrokeTool::ERASER`. I don't think anyone has been interested in getting eraser strokes. But it would nonetheless make sense to add that comparison.

hyperupcall:
Thank you, sounds good - I added that comparison :+1:

src/core/plugin/luapi_application.h
```diff
@@ -919,10 +919,12 @@ static int applib_getStrokes(lua_State* L) {
             StrokeTool tool = s->getToolType();
             if (tool == StrokeTool::PEN) {
                 lua_pushstring(L, "pen");
+            } else if (tool == StrokeTool::ERASER) {
+                lua_pushstring(L, "eraser")
```
rolandlo, line 923:
```suggestion
                lua_pushstring(L, "eraser");
```

rolandlo (approved):
LGTM

rolandlo:
Thanks for the fix @hyperupcall !

PR #4548, opened 2022-12-31
Write program version to uninstall entry on Windows

SpecterShell:
*Only line 270 is related to this PR. It seems that there is a line ending issue with this file.* 🤔

This PR allows the Windows installer to write the program version to the uninstall registry entry, so [WinGet](https://github.com/microsoft/winget-cli) will be able to recognize the installed version of Xournal++.
image


rolandlo:
Are you sure the `DisplayVersion` has the right format?
The source https://nsis.sourceforge.io/A_simple_installer_with_start_menu_shortcut_and_uninstaller and
https://docs.google.com/document/pub?id=1aeNBd1vdLhLznaktvP_xF2nKaaUG3kNlb_0VSRJ8mCI (referenced by nsis)
recommend it in the format 
`%major%.%minor%.%build%[.%revision%]` (e.g. 1.1.3)

I agree with the line ending issue

rolandlo:
For reference: https://github.com/microsoft/winget-cli/issues/1255

SpecterShell:
Actually there is no strict rule for this property so it can be in any format. For example:
https://github.com/microsoft/winget-pkgs/tree/master/manifests/a/Alibaba/DingTalk
https://github.com/microsoft/winget-pkgs/tree/master/manifests/h/HumbleBundle/HumbleApp

And it follows the version of the main program (i.e. the `VERSION` file generated after building), so if the main program follows the semver format, so does the installer.

rolandlo (approved):
Ok, in that case I'm fine with the PR. Thanks for your contribution!

PR #4549, opened 2023-01-01
Fix compile under Visual Studio

hyperupcall:
By [default](https://stackoverflow.com/a/42099622), it seems `convert` is no longer installed with ImageMagick on Windows because it conflicts with the PowerShell "builtin" (at least on my computer). This fixes that.

It also adds `.vs/` to the `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ nbproject/
 .idea
 cmake-build-*
 
+# Visual Studio
+.vs/
+
 # API Documentation
 doc/html
 doc/latex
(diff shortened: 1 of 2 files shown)


Febbe:
Why is this PR still a draft? Or are you really not done with it?

hyperupcall:
I think I wanted to test this with ImageMagick versions that were installed with the native installer rather than with scoop

hyperupcall:
I just tested it with the native installer and it works fine, so I'll just un-draft it

PR #4550, opened 2023-01-01
Use `std::numeric_limits` instead of macros

hyperupcall:
This improves consistency with the rest of the codebase. I think this also improves compatability, since it's in #4092 as well

### DIFF
--- a/src/core/plugin/Plugin.cpp
+++ b/src/core/plugin/Plugin.cpp
@@ -240,7 +240,7 @@ auto Plugin::callFunction(const std::string& fnc, long mode) -> bool {
 
     int numArgs = 0;
 
-    if (mode != LONG_MAX) {
+    if (mode != std::numeric_limits<long>::max()) {
         lua_pushinteger(lua.get(), mode);
         numArgs = 1;
     }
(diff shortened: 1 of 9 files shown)


rolandlo (changes requested):
Can you please apply "Include what you use" for the `<liimts>` header in the sense of https://include-what-you-use.org/ ? For instance it is currently neither included in `EditSelectionContents.cpp` nor `EditSelectionContents.h`.

Otherwise LGTM.

hyperupcall:
Sorry about that - fixed now

src/core/plugin/Plugin.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <algorithm>  // for max
 #include <array>      // for array
+#include <limits>     // for numeric_limits
```
rolandlo, line 5: Already included in `Plugin.h`
```suggestion
```

rolandlo (approved):
I have one minor nitpick. Could you please squash the commits then into one and rebase on current master afterwards?

hyperupcall:
:+1:

rolandlo (approved):
LGTM. 

rolandlo:
Thanks for this PR as well! Always good to improve consistency.

PR #4551, opened 2023-01-01
chore: Improvements to `.gitignore`

hyperupcall:
- `.*DS_Store` can be `.DS_STORE` and isn't autotools specific
- `*~` isn't autotools specific
- `*/.vscode/` is redundant
- `.idea` can have a trailing slash
- `src/**/.DS_Store` entries are redundant

cc @Febbe wanted to get your OK since you authored e16faa7bcdbceb49d8a4d15e07df3fd963552c85 - in theory both VSCode entries should behave the same, but maybe there was a Git bug or troubles with cached untracked files? Maybe it's fixed by now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,11 +2,9 @@
 .deps
 .dirstamp
 .libs
-*.DS_Store
 *.o
 *.lo
 *.la
-*~
 Makefile
 Makefile.in
 libtool
@@ -38,18 +36,14 @@ nbproject/
 
 # Visual Studio Code
 .vscode/
-*/.vscode/
 
 # CLion
-.idea
+.idea/
 cmake-build-*
 
 # API Documentation
 doc/html
 doc/latex
-
-src/core/view/.DS_Store
-src/core/view/background/.DS_Store
 
 # Flatpak
 repo/
@@ -60,8 +54,12 @@ flatpak-build/
 # Lua
 lua-*
 
-# Clangd 
+# Clangd
 compile_commands.json
-.clangd
-.cache
+.clangd/
+.cache/
 Debug
+
+# Miscellaneous
+.DS_STORE
+*~


Febbe:
Not a git bug, we have some sort of subproject in this git repository. Opening them in another directory will add a settings folder there, therefore it is quite useful to include all possible vscode folders.

I think the .gitignore entries should be relatively general to reduce unintended untracked files.

Just for curiosity, why did you touched the list?

.gitignore
```diff
@@ -60,8 +54,12 @@ flatpak-build/
...
 
-# Clangd 
+# Clangd
 compile_commands.json
 .clangd
 .cache
 Debug
+
+# Miscellaneous
+.DS_STORE
```
Febbe, line 64: This will only match the top level directory, but you removed the others. Can you elaborate, why the others are not matched anymore?

hyperupcall:
@Febbe Yeah I agree the entries should be general as possible. I think it's important to ignore editor directories in subdirectories, and I think my changes reflect that. According to [Git](https://git-scm.com/docs/gitignore):

> If there is a separator [slash] at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

So `.DS_STORE` should match all files and `.vscode/` should still match all directories, in any subdirectory (since this particular `.gitignore` is at the root), if I am reading things correctly. I just doubled checked on my machine and things worked according to the definition.

If it doesn't work like that, at least on your machine, then I think we should edit the `.gitignore` file so that for every editor, we have entries like `**/.vscode/**`, so it works for `.vscode` directories that are more than one level deep. But I don't think that should be necessary unless there is some sort of bug.

> Just for curiosity, why did you touched the list?

I edited the Gitignore file for #4549, and I noticed some inconsistencies with it and thought it could be improved so it was less confusing for other people

.gitignore
hyperupcall: The only files that shouldn't be matched anymore are hypothetical ones that look like `somename.DS_STORE`. In github/gitignore, the [macOS](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore) gitignore file didn't have an asterick, so I opted to remove it from here for simplicity.

For your concern about only matching the top level directory, I tried to address that with the other comment I made, does that answer your question?

Febbe:
Ok thanks, I've learned something new👌. 

Since I am very busy with my RL , I am not maintaining actively. Therefore @xournalpp/core  should carry on here.

hyperupcall:
Sounds good, thank you for your input :+1:

hyperupcall:
Hey @rolandlo @Technius wondering if you can take a look? (I'm not sure how to ping @xournalpp/core)

.gitignore
Technius: macOS can create `.DS_STORE` files in any folder, so this should ideally be a wildcard like: `**/.DS_STORE`

.gitignore
```diff
@@ -60,8 +54,12 @@ flatpak-build/
 # Lua
 lua-*
 
-# Clangd 
+# Clangd
 compile_commands.json
 .clangd
 .cache
```
Technius, line 60: To be consistent with the changes above:
```suggestion
.clangd/
.cache/
```

hyperupcall: I changed the `.clangd` (which is no longer in use since clangd v11) and `.cache` to append a slash.

With the `.DS_STORE` thing, my changes still match the `.DS_STORE` in any directory - I made a comment about it [here](https://github.com/xournalpp/xournalpp/pull/4551#issuecomment-1368581406).

Technius (approved):

Technius:

PR #4552, opened 2023-01-01
Use ostringstream to generate id string

jhilmer:
IMHO better to use ostringstream to generate the `getId` string

### DIFF
--- a/src/core/gui/toolbarMenubar/ColorToolItem.cpp
+++ b/src/core/gui/toolbarMenubar/ColorToolItem.cpp
@@ -3,6 +3,7 @@
 #include <array>    // for array
 #include <cstdio>   // for snprintf, size_t
 #include <memory>   // for unique_ptr
+#include <sstream>  // for ostringstream
 #include <utility>  // for move
 
 #include <glib.h>  // for gchar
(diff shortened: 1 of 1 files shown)


src/core/gui/toolbarMenubar/ColorToolItem.cpp
```diff
@@ -70,12 +71,9 @@ auto ColorToolItem::getId() const -> std::string {
...
     }
 
-    // Todo (modernize, cpp20): use std::format or fmtlibs fmt::format
-    std::array<char, 64> buffer{'\0'};
-    auto size = snprintf(buffer.data(), buffer.size(), "COLOR(%zu)", this->namedColor.getIndex());
-    std::string id = {buffer.data(), static_cast<size_t>(size)};
-
-    return id;
+    std::ostringstream os;
+    os << "COLOR(" << this->namedColor.getIndex() << ")" << std::ends;
```
rolandlo, line 75: Is `std::ends` really needed here?

jhilmer: No it's not needed.

rolandlo (approved):
LGTM

rolandlo:
Thanks for your contributions, @jhilmer ! Very much appreciated.

PR #4554, opened 2023-01-02
Fix memory leaks

jhilmer:
I have run xournalpp with valgrind. Based on the output I have found and fix the following memory leaks.

### DIFF
--- a/src/core/control/pagetype/PageTypeMenu.cpp
+++ b/src/core/control/pagetype/PageTypeMenu.cpp
@@ -11,6 +11,7 @@
 #include "control/settings/Settings.h"              // for Settings
 #include "util/Color.h"                             // for Color
 #include "util/i18n.h"                              // for _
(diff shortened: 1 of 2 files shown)


bhennion (approved):
Nice catch!
You could use our RAII wrappers for those C structures to make sure those unref calls don't get removed accidentally in the future:
https://github.com/xournalpp/xournalpp/blob/be01b216fac9495d143bc857a9b1967a45360c73/src/util/include/util/raii/CairoWrappers.h#L48
https://github.com/xournalpp/xournalpp/blob/be01b216fac9495d143bc857a9b1967a45360c73/src/util/include/util/raii/PangoSPtr.h#L31

jhilmer:
@bhennion are this the correct use of the wrappers e.g. the use of `xoj::util::adapt`?

bhennion (approved):
LGTM! Merging in 24 hours unless an objection is raised.

PR #4558, opened 2023-01-03
Refactor SidebarPreviewBase to use std::unique_ptr

jhilmer:
SidebarPreviewBase attribute preiews (vector of SidebarPreviewBaseEntry pointers) has been refactored to a vector of unique_ptr of SidebarPreviewBaseEntry


bhennion (approved):
LGTM!

PR #4559, opened 2023-01-03
Fix SEGV if app closed while toolbar customization was open

jhilmer:
Keep Toolbar alive until ToolbarAdapter is destroyed

### DIFF
--- a/src/core/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp
+++ b/src/core/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp
@@ -86,8 +86,9 @@ void ToolbarAdapter::cleanToolItem(GtkToolItem* it) {
     }
     GdkWindow* window = gtk_widget_get_window(GTK_WIDGET(it));
 
-    if (window)
+    if (window) {
         gdk_window_set_cursor(window, nullptr);
+    }
 
     gtk_tool_item_set_use_drag_window(it, false);
     gtk_drag_source_unset(GTK_WIDGET(it));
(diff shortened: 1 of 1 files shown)


src/core/gui/dialog/toolbarCustomize/ToolbarAdapter.cpp
```diff
@@ -84,8 +84,10 @@ void ToolbarAdapter::cleanToolItem(GtkToolItem* it) {
     if (data) {
         gtk_widget_set_sensitive(GTK_WIDGET(it), ToolitemDragDrop::isToolItemEnabled(data));
     }
+    GdkWindow* window = gtk_widget_get_window(GTK_WIDGET(it));
 
-    gdk_window_set_cursor(gtk_widget_get_window(GTK_WIDGET(it)), nullptr);
+    if (window)
+        gdk_window_set_cursor(window, nullptr);
```
jhilmer, line 90: I receive `Gdk-CRITICAL **: 12:18:13.853: gdk_window_set_cursor: assertion 'GDK_IS_WINDOW (window)' failed` for the five GtkToolItem where have a submenu attached. Pen, Eraser, PdfTool, Draw, and Select. This hack remove this - but I don't like the solution. But for now the application can close without giving a SEGV.

rolandlo: So should we merge or are you still looking for a better solution? I can confirm that the segfault is fixed.

bhennion, line 90: Just one thing, to match our coding conventions:
```suggestion
    if (window) {
        gdk_window_set_cursor(window, nullptr);
    }
```

rolandlo (approved):

rolandlo:
Thanks for the fix @jhilmer Just merged it.

PR #3773, opened 2022-02-04
Fix Poor rendering of TeX objects when exported to PDF.

tattsan:
Fix #3692 partially.

I replaced `poppler_page_render` with `poppler_page_render_for_printing` in `TexImageView::draw()`, but we may need to use different ones for the screen display and for the exports respectively.

There is another issue concerning ColorEmoji renndering. ColorEmoji texts are rendered in low resolution in exported PDF.

Issue #3692, opened 2022-01-02
TeX object is not correctly rendered in exported PDF.


tattsan:
**Affects versions :**
 - OS: Arch Linux
 - Desktop environment: X11 with xmonad
 - gtk: 3.24.31
 - Version of Xournal++: both 1.1.0 and master.
 - Installation method: both Arch package and built from source.

**Describe the bug**
TikZ picture in TeX object is not correctly rendered in exported PDF.

**To Reproduce**
1. Prepare LaTeX template file which contains '\usepackage{tikz}' .
2. Create a TeX object like this:
```
\begin{tikzpicture} \foreach\i in{0,...,200}{\draw[ultra thin] ({\i/50},0) -- (\i/50,3);} \end{tikzpicture}
```

3. Save the .xopp file, and export as PDF.
4. Compare the exported pdf with the temporary /tmp/xournalpp-?????/tex/tex.pdf. You'll see the line width and the distance between lines are incorrect.

This is  correct picture.
[tex.pdf](https://github.com/xournalpp/xournalpp/files/7798814/tex.pdf)

This is the exported file.
[foo-exp.pdf](https://github.com/xournalpp/xournalpp/files/7798815/foo-exp.pdf)

rolandlo:
Not sure what is going on here, since `pdftocairo -pdf tex.pdf tex_exp.pdf` results in a correct export.

tattsan:
`xournalpp foo.xopp  --export-png-dpi=600  -i foo-exp.png` creates a correct png file.

tattsan:
When you reopen the .xopp file, you'll see the wrong picture first. After about 300 millisecond, the picture will be rendered correctly. So there seems to be  two rendering processes.

https://user-images.githubusercontent.com/6355486/147891837-d6750fdf-b973-4178-a71f-41bd1d68f16b.mp4

When exported to PDF, you'll get the first picture.
When exported to PNG, you'll get the second picture.

tattsan:
I found color emoji fonts are also rendered poorly. In the following sample, I input an emoji char U+1F604 with text input tool.
* This is exported PDF:
[foo-exp.pdf](https://github.com/xournalpp/xournalpp/files/7966168/foo-exp.pdf)
* This is exported PNG (600dpi):
![foo-exp](https://user-images.githubusercontent.com/6355486/151705802-0021fea9-b479-4384-8240-39354962ac2a.png)

tattsan:
[...]

PR #4566, opened 2023-01-07
Fix latex' thin lines

### LINKED ISSUE #4555: parts of LaTeX equations not resizing
**Affects versions :**
 - OS: Linux Mint 20.3 x86_64
 - Version of Xournal++: 1.1.3

**Describe the bug**
Equations containing square roots, fractions or boxes don't resize correctly

**To Reproduce**
1. Open the LaTeX tool
2. Type an equation containing square roots, fractions or boxes, e. g. `$\boxed{\sqrt{\frac{x^2}{4}}}$`
3. Press ok
4. Resize the equation to ~0.5 or less of its original size
5. Deselect the equation

**Expected behavior**
The equation should not change in appearance after deselecting

**Video of Problem**
[image](https://user-images.githubusercontent.com/95933072/210266869-59ba42c3-96bd-4228-8cb4-64d7fa8f5651.mp4)

**Additional context**
After exporting to .pdf resized equations look ugly.

bhennion:
Fixes #4555 

The issue came from the use of `cairo_surface_set_device_scale` on the rendering masks. Replacing it with calls to `cairo_scale` fixed poppler's behaviour. Let me try to explain why.

`cairo_surface_set_device_scale` is supposed to tell a cairo surface that it'll be used in a window with a certain GDK_SCALE value (so HiDPI kinda stuff). In practice, this device scale is very handy, as most operations can then be made without ever caring about the DPI scaling.
Now, I (unknowingly) perverted this function and used it for mask via `cairo_surface_set_device_scale(surf, zoom * dpiScaling, zoom * dpiScaling)`. This works well, except when Poppler is involved... The reason lies in Poppler's screen rendering optimizations (see [the doc](https://poppler.freedesktop.org/api/glib/poppler-Poppler-Page.html#poppler-page-render-for-printing)): PDF files may contain width 0 lines, which should be rendered with "the minimal width the device can do". When printing, this depends on the printer's resolution. When rendering for on-screen view, this depends on the screen resolution. The way Poppler finds this resolution out is by looking at the `cairo_surface_t`'s size (width/height) and device scale.

My messing up the device scale made Poppler believe the pixel size was huge, so width 0 lines appeared very thick.

This PR is a quick fix, but I think we should remove every occurence of `cairo_surface_set_device_scale`. Using it was a mistake of mine.

### DIFF
--- a/src/core/control/jobs/RenderJob.cpp
+++ b/src/core/control/jobs/RenderJob.cpp
@@ -49,10 +49,11 @@ void RenderJob::rerenderRectangle(Rectangle<double> const& rect) {
 
     xoj::util::CairoSurfaceSPtr rectBuffer(cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height),
                                            xoj::util::adopt);
-    cairo_surface_set_device_offset(rectBuffer.get(), -x, -y);
-    cairo_surface_set_device_scale(rectBuffer.get(), ratio, ratio);
 
-    renderToBuffer(rectBuffer.get());
+    renderToBuffer(rectBuffer.get(), ratio, x, y);
+
+    cairo_surface_set_device_scale(rectBuffer.get(), ratio, ratio);
+    cairo_surface_set_device_offset(rectBuffer.get(), -x, -y);
 
     std::lock_guard lock(this->view->drawingMutex);
     xoj::util::CairoSPtr crPageBuffer(cairo_create(view->crBuffer.get()), xoj::util::adopt);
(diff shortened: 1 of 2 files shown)


RichardScharf:
(bump) When is this PR going to get merged into the latest nightly build?

rolandlo (commented):
I can confirm that #4555 is fixed by this PR. But why is `repaintPageArea` replaced by the more costly `repaintPage()`? That looks like a likely performance regression.

bhennion:
> I can confirm that #4555 is fixed by this PR. But why is `repaintPageArea` replaced by the more costly `repaintPage()`? That looks like a likely performance regression.

Of course! This was a debugging thing. I rebased on master and restored the `repaintPageArea`

rolandlo (commented):
Hmm, when I export to PDF and open the exported PDF again (with Xournal++ or evince) the issue (too thick thin lines) is still there. It would be great to fix that as well.

bhennion:
> Hmm, when I export to PDF and open the exported PDF again (with Xournal++ or evince) the issue (too thick thin lines) is still there. It would be great to fix that as well.

That's actually a different issue (with a different source). See #3692 and #3773. The fix proposed in #3773 is not quite satisfactory though: should be `poppler_page_render_for_printing` when exporting to PDF/SVG/PS but `poppler_page_render` for on-screen display.

rolandlo (approved):
Ok, you're right. With `poppler_page_render_for_printing` like in #3773 the issue disappears. So feel free to merge then.

PR #4573, opened 2023-01-10
Make it possible to scroll an entire page with vim keybindings

QuantMint:
Until now there was no way to scroll an entire page using vim keybindings, SHIFT + DOWN_ARROW did work, but SHIFT + j didn't do anything. With this pr it is now possible to use SHIFT + j to scroll a page down and SHIFT + k to scroll a page up.


bhennion (approved):
LGTM!

Issue #4569, opened 2023-01-08
Quality of life changes for Color PIcker


Lolol4572:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when picking a new color using the hex tool. When I select a color and go to the custom color picker, it always selects red, rather than the hex of the color I was on.

**Describe the solution you'd like**
I would like it to go to the hex value of that color, like it does on Google Docs, which makes me able to figure out the hex value and make slight adjustments without screenshotting it and going to a color picker website.

**Describe alternatives you've considered**
Addition of a color picker tool, maybe?

**Additional context**

rolandlo:
@Lolol4572 If I understand correctly you click on the custom color button in the toolbar, choose one of the predefined colors and want to adjust it a little. Instead of clicking on the `+`-button you should right-click on the chosen predefined color and choose `customize`. Does that already solve your problem?
![grafik](https://user-images.githubusercontent.com/40485433/212478887-6b35ec6a-cef7-4ca8-a5f5-56a660013e0d.png)

Lolol4572:
Oopsie-daisies! I didn't know about that until now. I'm a fairly new user btw. I may as well close this issue now.

Technius:
As originally suggested, we can still improve the UX of using the hex color picker, so reopening. Also see #4575 which would address this problem.

PR #4575, opened 2023-01-11
Color tool opens with current color selected

jhilmer:
Fix 'Quality of life changes for Color PIcker #4569'

The dialog now opens up with the current selected color. By right clicking on the color button - you can choose 'customize' which open up the color palette at the hex values of the selected color. The dialogs used for this is the GTK default dialogs for picking colors - they don't behave excatly as google docs picker.


Technius (approved):
LGTM. The proposed change will encode that the "Select Color Dialog" will open with the current tool's color selected.

However, I noticed that there still seems to be one additional issue where clicking the "+" button to choose a color won't default to the current selected color. Not sure if we can do anything about it--GTK's color chooser widget API seems pretty limited.

Technius:

PR #4582, opened 2023-01-12
Fix non-virtual destructor of GeometryToolController and derived class

bhennion:
This PR makes the destructor of GeometryToolController virtual, thus fixing a plausible memory leak.


jhilmer (approved):

rolandlo (approved):
LGTM

PR #4592, opened 2023-01-16
feat: Search for `pagetemplates.ini` in user config dirs

### LINKED ISSUE #4473: Honnor pagetemplates.ini in ~/.config/xournalpp/
I followed #2137 and I can successfully create new page templates in `/usr/share/xournalpp/ui/pagetemplates.ini` (I'm on Arch Linux) and use it.

Then,  I also tried copying that file to `~/.config/xournalpp/ui/pagetemplates.ini` and changes there are not honored (after closing and reopening Xournalpp).

So, if this is currently not possible, I would like to suggest it as a feature (it would also help those who keep their dotfiles in a repo).

hyperupcall:
With this change, the application now first searches for `pagetemplates.ini` in user directories (`XDG_CONFIG_HOME`, `XDG_CONFIG_DIRS`). If none are found, then it searches through `gladeSearchPath`, as it only did before.

I created a `SearchPath` class because modifying `GladeSearchpath` just to fix this bug seemed like the wrong thing to do. Calling `gladePath->addSearchDirectory()` could mean Glade components could be loaded from those configuration directories.

To keep this behavior intuitive for users and consistent with other config, I propose using this `SearchPath` when searching for at least `settings.xml` and `toolbar.ini`. Some parts of it may need to be modified so the fallback behavior codepath isn't ran for things that can _only_ be in XDG config directories.

Closes #4473


Technius:
There is similar logic in `PluginController`. Do you think it would be possible to generalize them so that all of these places use the same search path logic?

hyperupcall:
I made those changes including `PluginController`

src/core/control/settings/Settings.cpp
```diff
@@ -44,7 +45,14 @@ constexpr auto DEFAULT_FONT_SIZE = 12;
...
 
-Settings::Settings(fs::path filepath): filepath(std::move(filepath)) { loadDefault(); }
+Settings::Settings(SearchPath* configSearchPath): configSearchPath{configSearchPath} {
+    auto file = configSearchPath->findFile(SETTINGS_XML_FILE);
+    if (file.empty())
+        file = Util::getConfigFile(SETTINGS_XML_FILE);
+
+    this->filepath = file;
+    loadDefault();
+}
```
Technius, line 55: I don't see any reason to search for `settings.xml` in the config path. The expectation is that the file will always be located in `$XDG_CONFIG_HOME/xournalpp/settings.xml` (or equivalent on non-Linux platforms). I also think this moves too much complexity into `Settings`, when finding the config file can be handled by the caller instead.

src/core/control/Control.cpp
```diff
@@ -1235,7 +1235,7 @@ void Control::manageToolbars() {
 
     this->win->updateToolbarMenu();
 
-    auto filepath = Util::getConfigFile(TOOLBAR_CONFIG);
+    auto filepath = configSearchPath->findFile(TOOLBAR_CONFIG_FILE);
     this->win->getToolbarModel()->save(filepath);
```
Technius, line 1239: If this finds a `toolbar.ini` in a non-writable system directory, it will attempt and fail to save. I think it may be better to load from the search path but save to the user's config directory.

src/core/control/settings/Settings.cpp
```diff
@@ -731,7 +739,7 @@ auto Settings::load() -> bool {
      *  - if error during parsing load default, but do not overwrite
      *    existing palette file (would be annoying for users)
      */
-    auto paletteFile = Util::getConfigFile(PALETTE_FILE);
+    auto paletteFile = this->configSearchPath->findFile(PALETTE_FILE);
     if (!fs::exists(paletteFile)) {
         Palette::create_default(paletteFile);
```
Technius, line 744: Same problem regarding the non-writable system directories.

Technius (changes requested):
There seem to be a few problems with the suggested approach:

* `SearchPath` duplicates the existing logic in `GladeSearchPath`. They should be combined, with the `GladeSearchPath::findFile()` replaced with the better `SearchPath::findFile()`. The former is only used in a few places, with the first argument always set to empty string.
  * It may be convenient to define `using GladeSearchPath = SearchPath` to avoid excessive changes.
* `TOOLBAR_CONFIG_FILE` is a better name than `TOOLBAR_CONFIG`, but changing it in this PR seems unnecessary.
* It's unclear what the desired behavior is when trying to save config files, and I think it is worth discussing this point in more detail. One reasonable solution might be to save config files in the user's config folder, but are there situations where we don't want to do that?

PR #4594, opened 2023-01-16
chore: Replace `sprintf` with `PlaceholderString` in `AboutDialog.cpp`

hyperupcall:
Removes another instance of using raw C arrays

### DIFF
--- a/src/core/gui/dialog/AboutDialog.cpp
+++ b/src/core/gui/dialog/AboutDialog.cpp
@@ -1,11 +1,11 @@
 #include "AboutDialog.h"
 
-#include <cstdio>  // for sprintf
 #include <memory>  // for allocator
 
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
LGTM

rolandlo:

rolandlo:
> Why not replace with libfmt alltogether? Or are we waiting for C++20 `std::format`?

I think waiting for C++20 `std::format` is preferable, since adding a dependency tends to create issues.

Technius:
I disagree with waiting for C++20; I think fmtlib is pretty popular and should be available on most distros. In contrast, it will take a while for all compilers to fully support C++20. It may make sense to add fmtlib for 1.3.0.

rolandlo:
Thanks @hyperupcall for the PR. 
We will discuss if we should add fmtlib for 1.3.0. First there is 1.2.0 to come.

Issue #2897, opened 2021-03-02
Please implement an option for a default PDF export file name including wildcards 


Photon89:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I open a PDF file with Xournal++, annotate it, then export the annotated file as PDF and the proposed file name matches the file name of the original PDF which cannot be overwritten. So I have to change from tablet to laptop mode each time I export the annotated file to use the keyboard and modify the file name.

**Describe the solution you'd like**
Provide a setting for the exported PDF default file name similar to the already existing setting for the default file name of the saved .xopp file. A possible wildcard could be the file name of the original PDF file (if existent) or .xopp file's name (if no background PDF existent) which then could be modified by some suffix, say %{original file name}_annotated.pdf

**Describe alternatives you've considered**
Another possible solution is already discussed in https://github.com/xournalpp/xournalpp/issues/2363.

nicolae-stroncea:
I think the benefit of a default _editable_ export name extension (like "%{original file name}_annotated") is twofold:
* Users don't need to change the name every time they export, as the current default is you get an error until you change it.
* Users can be allowed to forcefully export to the same name as the original pdf, by changing name to name of PDF. Because the default will be set to something like "_annotated.pdf", this would get rid of any errors where the user accidentally overrides the default. Moreover, a confirmation box can be displayed to confirm that the user will override the background if they choose to do so.

Technius:
Upgrading to high priority as this would solve all of the various complaints and differences in opinion about expected filename of exported files.

forgottosave:
I can take a look at this in the next week(s).

After implementing the logic for this, which shouldn't be too difficult (famous last words :P), I'd probably implement a new point in the settings in `Load / Save` named something like `Default pdf-export name` where a string can be entered using wildcards (similar to the suggestion in the issue) to change the default name.

forgottosave:
Should there be more wildcards, other than the original file name? I couldn't come up with any, other than maybe the current date and time.

And if the name is the only wildcard option, should I still implement it in a way, that future wildcards are easy to add? Because this does make a difference in the changed code.

Technius:
> I can also open a PR with my current changes to discuss implementation details, if this is preferred over the issue thread

That would be very helpful. It would also be great if you could include a list of wildcards that you plan to support.

bhennion:
Fixed via #4599

Photon89:
[...]

PR #4599, opened 2023-01-17
Default pdf-export name with wildcards

forgottosave:
## Issue
This PR implements the requested enhancement in #2897 

## Feature Description
This PR introduces the possibility to set a default name for exporting pdfs. Among others, this solves the issue of overwriting the background-pdf. Wildcards are used for allowing some file-related naming and use the `%{...}` syntax.

Wildcards | Description
---|---
name | original pdf name
date | current date
time | current time

The default pdf-name will be `%{name}_annotated.pdf`.

### DIFF
--- a/src/core/control/jobs/BaseExportJob.cpp
+++ b/src/core/control/jobs/BaseExportJob.cpp
@@ -69,7 +69,7 @@ auto BaseExportJob::showFilechooser() -> bool {
     Document* doc = control->getDocument();
     doc->lock();
     fs::path folder = doc->createSaveFolder(settings->getLastSavePath());
-    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), /*defaultPfdName*/settings->getDefaultPdfExportName());
+    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), settings->getDefaultPdfExportName());
     doc->unlock();
 
     gtk_file_chooser_set_local_only(GTK_FILE_CHOOSER(dialog), true);
--- a/src/core/control/jobs/BaseExportJob.cpp
+++ b/src/core/control/jobs/BaseExportJob.cpp
@@ -69,7 +69,7 @@ auto BaseExportJob::showFilechooser() -> bool {
     Document* doc = control->getDocument();
     doc->lock();
     fs::path folder = doc->createSaveFolder(settings->getLastSavePath());
-    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName());
+    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), /*defaultPfdName*/"%{name}_annotated"); // TODO get defaultPdfName from settings
     doc->unlock();
 
     gtk_file_chooser_set_local_only(GTK_FILE_CHOOSER(dialog), true);
--- a/src/core/control/jobs/BaseExportJob.cpp
+++ b/src/core/control/jobs/BaseExportJob.cpp
@@ -69,7 +69,7 @@ auto BaseExportJob::showFilechooser() -> bool {
     Document* doc = control->getDocument();
     doc->lock();
     fs::path folder = doc->createSaveFolder(settings->getLastSavePath());
-    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), /*defaultPfdName*/"%{name}_annotated"); // TODO get defaultPdfName from settings
+    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), /*defaultPfdName*/settings->getDefaultPdfExportName());
     doc->unlock();
 
     gtk_file_chooser_set_local_only(GTK_FILE_CHOOSER(dialog), true);
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -134,9 +134,9 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
 
 std::string Document::parseFilenameWildcard(const std::string& wildcard) {
     if (wildcard == "name") {
-        fs::path pdfPath = pdfFilepath.filename();
+        fs::path pdfPath = this->pdfFilepath.filename();
         Util::clearExtensions(pdfPath, ".pdf");
-        return pdfPath;
+        return pdfPath.u8string();
     }
     if (wildcard == "date" || wildcard == "time") {
         std::time_t time = std::chrono::system_clock::to_time_t(
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -170,8 +170,7 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
             }
 
             std::string wildcard = saveString.substr(pos + 2, endPos - pos - 2);
-            saveString.replace(pos, endPos + 1, parseFilenameWildcard(wildcard));
-
+            saveString.replace(pos, endPos + 1 - pos, parseFilenameWildcard(wildcard));
             pos = saveString.find(DEFAULT_PDF_WILDCARD_START, pos);
         }
 
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -17,6 +17,7 @@
 #include "pdf/base/XojPdfBookmarkIterator.h"  // for XojPdfBookmarkIterator
 #include "util/PathUtil.h"                    // for clearExtensions
 #include "util/PlaceholderString.h"           // for PlaceholderString
+#include "util/SaveNameUtils.h"               // for parseFilename
 #include "util/Util.h"                        // for npos
 #include "util/i18n.h"                        // for FS, _F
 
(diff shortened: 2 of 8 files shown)


jhilmer:
I suggest moving `parseFilenameWildcard` to a free function and create som unittest. What happen 'pdf-name' has incorrect syntax e.g. missing '}'?
''

forgottosave:
Should the wildcard not be implemented (e.g. `%{whatever}`) this will be replaced by an empty string. Although the expected behavior in this case is discussable :)

If the pdf name doesn't have a `}`, then it will not be recognized as a wildcard usage (as per definition it isn't) and it will be evaluated as a normal part of the name.

forgottosave:
Not sure if we should leave the wildcard parsing in this method `parseFilenameWildcard()`, or create a separate file to store all the different wildcards (maybe in `constexpr`s or an `enum`) and their parsing method.

src/core/model/Document.cpp
```diff
@@ -128,21 +132,52 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
...
-        if (this->attachPdf) {
-            Util::clearExtensions(p, ".pdf");
-        } else {
-            Util::clearExtensions(p);
+        // Build the pdf-filepath according to settings (replace wildcards)
+        std::string saveString = defaultPfdName;
+        size_t pos = saveString.find(DEFAULT_PDF_WILDCARD_START);
+
+        while (pos != std::string::npos) {
+            size_t endPos = saveString.find(DEFAULT_PDF_WILDCARD_END, pos + 2);
```
bhennion, line 167: The constant `2` is the length of `DEFAULT_PDF_WILDCARD_START`:
```suggestion
            size_t endPos = saveString.find(DEFAULT_PDF_WILDCARD_END, pos + std::strlen(DEFAULT_PDF_WILDCARD_START));
```

src/core/model/Document.cpp
```diff
@@ -128,21 +132,52 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
...
+            if (endPos == std::string::npos) {
+                break;
+            }
+
+            std::string wildcard = saveString.substr(pos + 2, endPos - pos - 2);
```
bhennion, line 172: same comment here

src/core/control/jobs/BaseExportJob.cpp
```diff
@@ -69,7 +69,7 @@ auto BaseExportJob::showFilechooser() -> bool {
     Document* doc = control->getDocument();
     doc->lock();
     fs::path folder = doc->createSaveFolder(settings->getLastSavePath());
-    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName());
+    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), /*defaultPfdName*/settings->getDefaultPdfExportName());
```
bhennion, line 72:
```suggestion
    fs::path name = doc->createSaveFilename(Document::PDF, settings->getDefaultSaveName(), settings->getDefaultPdfExportName());
```

bhennion (changes requested):
Thanks for the PR! That's a very nice feature you're adding here!
I have a couple of nitpicks (see other comments).

About that:

> Not sure if we should leave the wildcard parsing in this method parseFilenameWildcard(), or create a separate file to store all the different wildcards (maybe in constexprs or an enum) and their parsing method.

I guess a separate file, in src/util/ would make more sense. This has nothing to do with the Document model. I'd also argue that most of the code you added to `Document::createSaveFilename` would fit better in this new file as well.

src/core/model/Document.cpp
forgottosave: Oh yes, seems way better than this hard-coded stuff. Thank you :)

src/util/include/util/SaveNameUtils.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+#include <cstring>      // for size_t
```
jhilmer, line 14: Not needed in this file

jhilmer (approved):
I like it has been moved away from `Document` class. Now it will also be very easy to implement an unittest of `parseFilenameFromWildcardString` there can convince everybody that the code is working.

src/util/SaveNameUtils.cpp
```diff
@@ -0,0 +1,45 @@
+#include "util/SaveNameUtils.h"
+
+#include <sstream>            // for stringstream
+#include <chrono>             // for time
+#include <iomanip>            // for put_time
+
+#include "util/PathUtil.h"    // for clearExtensions
+
+
+auto SaveNameUtils::parseFilenameFromWildcardString(const std::string& wildcardString, fs::path defaultFilePath) -> std::string {
```
bhennion, line 10:
```suggestion
auto SaveNameUtils::parseFilenameFromWildcardString(const std::string& wildcardString, const fs::path& defaultFilePath) -> std::string {
```

src/util/SaveNameUtils.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return saveString;
+}
+
+auto SaveNameUtils::parseWildcard(const std::string& wildcard, fs::path defaultFilePath) -> std::string {
+    if (wildcard == WILDCARD_NAME) {
+        Util::clearExtensions(defaultFilePath, ".pdf");
+        return defaultFilePath.u8string();
+    }
```
bhennion, line 33: Only copy the path when actually needed.

```suggestion
auto SaveNameUtils::parseWildcard(const std::string& wildcard, const fs::path& defaultFilePath) -> std::string {
    if (wildcard == WILDCARD_NAME) {
        fs::path path = defaultFilePath;
        Util::clearExtensions(path, ".pdf");
        return path.u8string();
    }
```

src/util/include/util/SaveNameUtils.h
```diff
@@ -0,0 +1,33 @@
...
+constexpr auto WILDCARD_NAME = "name"; // default store name, e.g. original pdf name
+constexpr auto WILDCARD_DATE = "date"; // current date
+constexpr auto WILDCARD_TIME = "time"; // current time
+
+class SaveNameUtils {
+public:
+    static std::string parseFilenameFromWildcardString(const std::string& wildcardString, fs::path defaultFilePath);
+
+private:
+    static std::string parseWildcard(const std::string& wildcard, fs::path defaultFilePath);
```
bhennion, line 32:
```suggestion
    static std::string parseFilenameFromWildcardString(const std::string& wildcardString, const fs::path& defaultFilePath);

private:
    static std::string parseWildcard(const std::string& wildcard, const fs::path& defaultFilePath);
```

bhennion (approved):
LGTM, with the exception of the nitpick below (and @jhilmer's comment)

src/util/include/util/SaveNameUtils.h
jhilmer: I added a unitest - I found a problem if `defaultFilePath` eg. contains "xxxxxxx%{name}" (endless loop)

https://github.com/forgottosave/xournalpp/pull/4

jhilmer:
I added a unitest - I found a problem if `defaultFilePath` eg. contains "xxxxxxx%{name}" (endless loop)

https://github.com/forgottosave/xournalpp/pull/4

forgottosave:
Implemented the change requests and fixed the recursive search

src/util/include/util/SaveNameUtils.h
```diff
@@ -9,9 +9,6 @@
  * @license GNU GPLv2 or later
  */
 
-#pragma once
```
jhilmer: `#pragma once` should still be in the header file.

jhilmer (changes requested):

src/util/include/util/SaveNameUtils.h
forgottosave: Yes, was already confused about that one, sorry

forgottosave:
As soon as this PR is approved and ready to merge (no more changes requested) I can clean up the commit history and squash some stuff, as it really is a mess right now...

test/unit_tests/util/SaveNameUtilsTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+
+#include <gtest/gtest.h>
+
+#include "util/SaveNameUtils.h"
+
+using namespace std;
+
+TEST(SaveNameUtils, testWildcardExpandsion) {
```
bhennion, line 19: Typo:
```suggestion
TEST(SaveNameUtils, testWildcardExpansion) {
```

test/unit_tests/util/SaveNameUtilsTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @license GNU GPLv2 or later
+ */
+
+
+#include <gtest/gtest.h>
+
+#include "util/SaveNameUtils.h"
+
+using namespace std;
```
bhennion, line 17: Unneeded. Also, we're trying to avoid `using namespace`'s.
```suggestion
```

bhennion (approved):
Tiny nitpicks. Otherwise LGTM!
Thanks for the great job!

@forgottosave Don't worry about the commit tree, we'll squash everything before merging anyway.

bhennion:
> @forgottosave Don't worry about the commit tree, we'll squash everything before merging anyway.

Uh... Except there is a merge conflict to resolve. Could you rebase on latest master?

bhennion (approved):
LGTM!

**Squashing and merging** in 24 hours unless an objection is raised.

Issue #4580, opened 2023-01-12
Reduce rerendering of large areas


hrdl-github:
https://github.com/xournalpp/xournalpp/pull/4502 and https://github.com/xournalpp/xournalpp/pull/4493 reduce the size of repaint areas. In combination they decrease the lag significantly on the PINE64 PineNote, which features a electrophoretic display.

In addition I use https://github.com/xournalpp/xournalpp/commit/1639cc0cf26bbbf90203b4339227e46701fb27a2 and https://github.com/xournalpp/xournalpp/commit/05ffd0449a8aae53ed4f4ed65de0a997ccbb22c5 to prevent regular completed strokes from being repainted, which avoids large damage region to avoid occasional delays due to how the display driver processes damage regions.

I also skip rerendering and repainting pages upon change entirely (https://github.com/xournalpp/xournalpp/commit/e949b630004add56fab63e6317bb1c5afab762a9) for the same reason. At the moment this means that a changed background is not rerendered and repainted until e.g. the page is changed.

I have two questions:

1. Do you have an idea when rerendering and repainting is required or desired? Besides the delayed change of background I haven't noticed any difference in the rendered output. I've disabled rerendering only for the pen tool at the moment (https://github.com/xournalpp/xournalpp/commit/05ffd0449a8aae53ed4f4ed65de0a997ccbb22c5).
2. Would you be interested in merging these contributions once they are mature? They add a small amount of complexity to the rerendering logic and possibly don't yield any noticeable improvement for the majority of users / devices.

rolandlo:
About 2.: Performance is one of the major goals of Xournal++, so yes, even if it adds a small amount of complexity to the rendering logic and the majority of users/devices wouldn't feel any noticeable improvement, it would be great to have those contributions.

bhennion:
First, some comments about the patches you wrote, then I'll try and answer your question 1.

> In addition I use [1639cc0](https://github.com/xournalpp/xournalpp/commit/1639cc0cf26bbbf90203b4339227e46701fb27a2) and [05ffd04](https://github.com/xournalpp/xournalpp/commit/05ffd0449a8aae53ed4f4ed65de0a997ccbb22c5) to prevent regular completed strokes from being repainted, which avoids large damage region to avoid occasional delays due to how the display driver processes damage regions.

[...]

PR #4611, opened 2023-01-21
Remove or shrink unnecessary page rerender calls

hrdl-github:
This requires #4158 that requires fewer explicit redraws when layers are deleted or their visibility is modified. Addresses #4580 in combination with #4158. This reduces the number of page repaints required. Full repaints are still used upon
- finalisation of text edits
- layer changes
- initialisation

TODO:
- [x] More extensive testing
- [x] Remove debug messages

### DIFF
--- a/src/core/control/layer/LayerController.cpp
+++ b/src/core/control/layer/LayerController.cpp
@@ -328,11 +328,11 @@ void LayerController::switchToLay(Layer::Index layerId, bool hideShow, bool clea
 
     if (hideShow) {
         for (Layer::Index i = 1; i <= p->getLayerCount(); i++) { p->setLayerVisible(i, i <= layerId); }
-    }
 
-    // Repaint page
-    control->getWindow()->getXournal()->layerChanged(selectedPage);
-    fireLayerVisibilityChanged();
+        // Repaint page
+        control->getWindow()->getXournal()->layerChanged(selectedPage);
+        fireLayerVisibilityChanged();
+    }
 }
 
 /**
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -64,7 +64,7 @@ EditSelection::EditSelection(UndoRedoHandler* undo, const PageRef& page, XojPage
 
 EditSelection::EditSelection(UndoRedoHandler* undo, Selection* selection, XojPageView* view):
         snappingHandler(view->getXournal()->getControl()->getSettings()) {
-    calcSizeFromElements(selection->selectedElements);
+    Range r = calcSizeFromElements(selection->selectedElements);
 
     construct(undo, view, view->getPage());
 
@@ -80,7 +80,7 @@ EditSelection::EditSelection(UndoRedoHandler* undo, Selection* selection, XojPag
         this->sourceLayer->removeElement(e, false);
     }
 
-    view->rerenderPage();
+    view->rerenderRange(r);
 }
 
 EditSelection::EditSelection(UndoRedoHandler* undo, Element* e, XojPageView* view, const PageRef& page):
@@ -97,7 +97,7 @@ EditSelection::EditSelection(UndoRedoHandler* undo, Element* e, XojPageView* vie
 EditSelection::EditSelection(UndoRedoHandler* undo, const vector<Element*>& elements, XojPageView* view,
                              const PageRef& page):
         snappingHandler(view->getXournal()->getControl()->getSettings()) {
-    calcSizeFromElements(elements);
+    Range r = calcSizeFromElements(elements);
     construct(undo, view, page);
 
     for (Element* e: elements) {
@@ -105,13 +105,13 @@ EditSelection::EditSelection(UndoRedoHandler* undo, const vector<Element*>& elem
         this->sourceLayer->removeElement(e, false);
     }
 
-    view->rerenderPage();
+    view->rerenderRange(r);
 }
 
 EditSelection::EditSelection(UndoRedoHandler* undo, XojPageView* view, const PageRef& page, Layer* layer):
         snappingHandler(view->getXournal()->getControl()->getSettings()) {
     const auto& elements = layer->getElements();
-    calcSizeFromElements(elements);
+    Range r = calcSizeFromElements(elements);
     construct(undo, view, page);
 
     long i = 0L;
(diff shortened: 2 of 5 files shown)


src/core/gui/PageView.h
```diff
@@ -58,7 +58,7 @@ class XojPageView: public LegacyRedrawable, public PageListener, public xoj::vie
 
 public:
     void addOverlayView(std::unique_ptr<xoj::view::OverlayView>);
-    void rerenderPage() override;
+    void rerenderPage(bool repaint = false) override;
```
bhennion, line 61: This default value makes the reviewing process harder: the behaviour is only changed for the occurences of `rerenderPage` that **do not** appear in the diff.
Could you remove the default (or default to true), so that the places where the behaviour changed are easy to find?

bhennion (commented):
To be honest, I'm a bit conflicted about this PR.
One the one hand, it seems to work fine without the repainting you removed. On the other hand, what's the point of refreshing the page buffer if it's not to display it?

For instance, when an EditSelection is created (meaning some Elements just got selected), the Elements are:
- removed from the page, added to a selection, painted to a selection buffer, and all of that is then painted on screen **in the main loop**
- erased from the page buffer (via `rerenderPage()` which is an overkill) **in a separated thread**, followed by a full repaint.

With this PR, you removed the asynchronous full repaint, while I think the synchronous repaint is the superfluous one: we should wait for the page buffer to be updated before repainting.

That's just an example, and I think the other instances are similar. At this point, I can't imagine a situation where the page buffer needs updating without a follow up repaint being needed.

src/core/control/layer/LayerController.cpp
```diff
@@ -328,10 +328,10 @@ void LayerController::switchToLay(Layer::Index layerId, bool hideShow, bool clea
...
     if (hideShow) {
         for (Layer::Index i = 1; i <= p->getLayerCount(); i++) { p->setLayerVisible(i, i <= layerId); }
-    }
 
-    // Repaint page
-    control->getWindow()->getXournal()->layerChanged(selectedPage);
+        // Repaint page
+        control->getWindow()->getXournal()->layerChanged(selectedPage);
+    }
     fireLayerVisibilityChanged();
```
bhennion, line 335: This looks like it should be inside the `if (hideShow)` scope.

src/core/control/tools/EditSelection.cpp
```diff
@@ -75,12 +75,16 @@ EditSelection::EditSelection(UndoRedoHandler* undo, Selection* selection, XojPag
         order.insert(std::upper_bound(order.begin(), order.end(), i, EditSelectionContents::insertOrderCmp), i);
     }
 
+    Range range;
+
     for (const auto& [e, i]: order) {
+        range.addPoint(e->getX(), e->getY());
+        range.addPoint(e->getX() + e->getElementWidth(), e->getY() + e->getElementHeight());
```
bhennion, line 82: The call to `calcSizeFromElements` at the beginning of the constructor suggests the bounding box has already been computed. Can't we use that instead of recomputing it?

src/core/control/tools/EditSelection.cpp
```diff
@@ -100,12 +104,16 @@ EditSelection::EditSelection(UndoRedoHandler* undo, const vector<Element*>& elem
     calcSizeFromElements(elements);
     construct(undo, view, page);
 
+    Range range;
+
     for (Element* e: elements) {
+        range.addPoint(e->getX(), e->getY());
+        range.addPoint(e->getX() + e->getElementWidth(), e->getY() + e->getElementHeight());
```
bhennion, line 111: Same comment here

src/core/control/tools/EditSelection.cpp
```diff
@@ -121,8 +129,6 @@ EditSelection::EditSelection(UndoRedoHandler* undo, XojPageView* view, const Pag
     }
 
     layer->clearNoFree();
-
-    view->rerenderPage();
```
bhennion, line 125: Why wouldn't a rerendering be necessary here?

src/core/gui/PageView.cpp
```diff
@@ -1166,7 +1166,7 @@ void XojPageView::rectChanged(Rectangle<double>& rect) { rerenderRect(rect.x, re
 
 void XojPageView::rangeChanged(Range& range) { rerenderRange(range); }
 
-void XojPageView::pageChanged() { rerenderPage(); }
```
bhennion, line 1169: This instance should be kept, I think.

src/core/gui/XournalView.cpp
```diff
@@ -553,16 +553,9 @@ void XournalView::zoomChanged() {
 
 void XournalView::pageSizeChanged(size_t page) {
     layoutPages();
-    if (page != npos && page < this->viewPages.size()) {
-        this->viewPages[page]->rerenderPage();
-    }
```
bhennion, line 558: Same here. If the page size changed, then a rerender is certainly necessary!

src/core/gui/XournalView.cpp
```diff
@@ -553,16 +553,9 @@ void XournalView::zoomChanged() {
...
 }
 
-void XournalView::pageChanged(size_t page) {
-    if (page != npos && page < this->viewPages.size()) {
-        this->viewPages[page]->rerenderPage();
-    }
-}
```
bhennion, line 565: And here as well

bhennion (changes requested):

src/core/gui/XournalView.cpp
bhennion: `XournalView::pageChanged` is called via `DocumentHandler::firePageChange`, which is used when the page background is changed (either the color `Control::changePageBackgroundColor()` or the pattern `PageBackgroundChangeController::commitPageTypeChange`), and in the timer callback mentionned in https://github.com/xournalpp/xournalpp/issues/4580#issuecomment-1381807460.

So no, it is useful, even if we remove the timer story.
It is also what you expect of this function: if anyone fires a PageChanged event, then the receiving XournalView should update said page. Doing nothing in such a callback function, while any dev would assume otherwise, is very much prone to errors.

bhennion: I disagree. Quick and dirty fixes are not a good idea in the long run.
I think the best course of action here is to restore those callback functions as they were, merge the rest of your changes (I'll rereview before that), wait for #4158 to get merged and then see if the rerender timer can be cleaned up.
It ain't a lot of work, but a bit of patience, I guess.

src/core/control/tools/EditSelection.cpp
```diff
@@ -80,7 +80,7 @@ EditSelection::EditSelection(UndoRedoHandler* undo, Selection* selection, XojPag
         this->sourceLayer->removeElement(e, false);
     }
 
-    view->rerenderPage();
+    view->rerenderRect(snappedBounds.x, snappedBounds.y, snappedBounds.width, snappedBounds.height);
```
bhennion, line 83: That won't quite work: The snappedBounds of Stroke elements correspond to the bounding box of the stroke if it had width == 0 (it is used to snap object to the grid). So in this instance, it'll be too small a rerender.
The simplest thing would possibly be to have `calcSizeFromElements()` return the `Range range` it uses in its computations, and use this returned Range to rerender.

bhennion (approved):
LGTM!
Squashing and merging in 48 hours unless an objection is raised.

PR #4613, opened 2023-01-22
Fix of memory leaks

jhilmer:
By running bouth xournal and unittest with valgrind a have found a bunch of memory leaks problems.

### DIFF
--- a/src/core/control/xml/XmlNode.cpp
+++ b/src/core/control/xml/XmlNode.cpp
@@ -34,12 +34,8 @@ void XmlNode::setAttrib(const char* attrib, int value) { putAttrib(new IntAttrib
 
 void XmlNode::setAttrib(const char* attrib, size_t value) { putAttrib(new SizeTAttribute(attrib, value)); }
 
-/**
- * The double array is now owned by XmlNode and automatically deleted!
- */
 void XmlNode::setAttrib(const char* attrib, double* value, int count) {
     putAttrib(new DoubleArrayAttribute(attrib, std::vector<double>{value, value + count}));
-    g_free(value);
 }
 
 void XmlNode::writeOut(OutputStream* out, ProgressListener* listener) {
--- a/src/core/control/xojfile/LoadHandler.cpp
+++ b/src/core/control/xojfile/LoadHandler.cpp
@@ -43,6 +43,11 @@ using std::string;
         error = g_error_new(G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT, __VA_ARGS__); \
     }
 
+namespace {
+    constexpr size_t MAX_VERSION_LENGTH = 50;
+    constexpr size_t MAX_MIMETYPE_LENGTH = 25;
+}
+
 LoadHandler::LoadHandler():
         attachedPdfMissing(false),
         removePdfBackgroundFlag(false),
@@ -134,9 +139,9 @@ auto LoadHandler::openFile(fs::path const& filepath) -> bool {
                     FS(_F("The file is no valid .xopp file (Mimetype missing): \"{1}\"") % filepath.u8string());
             return false;
         }
-        char mimetype[25];
+        char mimetype[MAX_MIMETYPE_LENGTH + 1] = {};
         // read the mimetype and a few more bytes to make sure we do not only read a subset
-        zip_fread(mimetypeFp, mimetype, 25);
+        zip_fread(mimetypeFp, mimetype, MAX_MIMETYPE_LENGTH);
         if (!strcmp(mimetype, "application/xournal++")) {
             zip_fclose(mimetypeFp);
             this->lastError = FS(_F("The file is no valid .xopp file (Mimetype wrong): \"{1}\"") % filepath.u8string());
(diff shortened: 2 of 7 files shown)


src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -134,9 +139,9 @@ auto LoadHandler::openFile(fs::path const& filepath) -> bool {
                     FS(_F("The file is no valid .xopp file (Mimetype missing): \"{1}\"") % filepath.u8string());
             return false;
         }
-        char mimetype[25];
+        char mimetype[MAX_VERSION_LENGTH + 1] = {};
         // read the mimetype and a few more bytes to make sure we do not only read a subset
-        zip_fread(mimetypeFp, mimetype, 25);
+        zip_fread(mimetypeFp, mimetype, MAX_VERSION_LENGTH);
```
jhilmer, line 144: Ups should have been `MAX_MINETYPE_LENGTH`

bhennion (commented):
Thanks a lot for all your helpful code cleanup!
I only have a couple of nitpicks

src/core/model/Stroke.cpp
```diff
@@ -259,9 +259,9 @@ void Stroke::setLastPoint(const Point& p) {
 }
 
 void Stroke::addPoint(const Point& p) {
-    this->points.emplace_back(p);
     updateBounds(Element::x, Element::y, Element::width, Element::height, Element::snappedBounds, p,
                  hasPressure() ? p.z / 2.0 : this->width / 2.0);
+    this->points.emplace_back(p);
```
jhilmer, line 264: I am wrong here. As Point don't have move constructor `p` is copied when emplaced_back. The problem is that a Point can be constructed with the `z` attribute uninitialized. I will remove this commit - and ensure `Point(x,y)` initialize the z component to Point::NO_PRESSURE.

src/core/control/xml/XmlNode.cpp
```diff
@@ -34,12 +34,8 @@ void XmlNode::setAttrib(const char* attrib, int value) { putAttrib(new IntAttrib
 
 void XmlNode::setAttrib(const char* attrib, size_t value) { putAttrib(new SizeTAttribute(attrib, value)); }
 
-/**
- * The double array is now owned by XmlNode and automatically deleted!
- */
 void XmlNode::setAttrib(const char* attrib, double* value, int count) {
     putAttrib(new DoubleArrayAttribute(attrib, std::vector<double>{value, value + count}));
-    g_free(value);
```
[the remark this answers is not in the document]
jhilmer: I will try do do that

src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -125,8 +125,8 @@ void SaveHandler::visitStroke(XmlPointNode* stroke, Stroke* s) {
         for (int i = 0; i < pointCount; i++) {
             values[i + 1] = s->getPoint(i).z;
         }
-
         stroke->setAttrib("width", values, pointCount);
+        delete [] values;
```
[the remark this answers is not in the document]
jhilmer: Agree

test/unit_tests/util/ObjectIOStreamTest.cpp
```diff
@@ -19,50 +19,64 @@ std::string serializeData(const std::array<T, N>& data) {
     ObjectOutputStream outStream(new BinObjectEncoding);
     outStream.writeData(&data[0], N, sizeof(T));
     auto outStr = outStream.getStr();
-    return {outStr->str, outStr->len};
+    auto resStr = std::string{outStr->str, outStr->len};
+    g_string_free(outStr, true);
+    return resStr;
```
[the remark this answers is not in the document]
jhilmer: Agree about the long run here.

src/core/model/Stroke.cpp
jhilmer: No I am not wrong here!.. `Point` has an implicit move constructor `Point(const &&p)`. So `p` is moved after the call to `emplace_back` and therefore the `z` component of p is uninitialized - even `p` is a const reference to a Point.

bhennion: I don't quite understand what the point is. Could you dumb it down for me?

jhilmer: You are correct - the copy construct are called and not the move constructor, z is initialized

src/core/control/xml/XmlNode.cpp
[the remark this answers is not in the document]
Febbe: @bhennion `std::forward` is for forward types, but they only appear as `auto &&` or `template<class T> T&&`. 
Any parameter / type, which is not a template type && or `auto&&` can't be forwarding. 
So using `std::vector&&` does not result in a forward of it. So in this case, `std::move` is sufficient.
`Type&&` enforces, a binding from x-values, while `Type` binds to everything even if it results in a copy.

src/core/model/Stroke.cpp
Febbe: The main difference between `emplace` and `push_back` is, that `emplace` allows the type to be constructed in the container from any constructor in-place. It basically calls `new (address_in_my_buffer) Type(std::forward<Args>(args)...);` For push back, `emplace` is also called, but with the copy/move constructor. This may result in an additional move or copy, which is mostly optimized out. 

==

That a move "resets" the moved from object is a common misunderstanding. It is only guaranteed, that the value is in a valid and safely destructible state. For most cases, a move actually does nothing. And why should it, most of them are destructed anyway. For trivially copiable types, a move is the same as a copy. Generally speaking, the access of a member of an moved-from type is undefined for default and implicit move constructors, but implementation defined for explicitly defined move constructors.

==

The default initializer only initializes the member, when the called constructor does not do it in its initialization list.
On any move, there is no call of any constructor on the moved from object, therefore, no change happens to that value.
That value can only be changed over the moved-to move constructor, which does nothing when the move constructor is defaulted or implicit.

bhennion:
@jhilmer Would you want to finish up this PR (at some point)? Fixing those memleaks is a good idea :)
Othewise, someone can take over the PR and finish it for you.

bhennion:
Superseeded by #5164

PR #4618, opened 2023-01-24
Ensure to init z component of Point

jhilmer:


Technius:
Isn't `z` initialized here? https://github.com/xournalpp/xournalpp/blob/1038ec54def0d6c07590c114e8b238e243d7e692/src/core/model/Point.h#L99

But I do agree that we should make all of the initialization consistent: either we initialize all of the fields in the constructor, or we initialize them all where they're declared.

jhilmer:
Correct `z` was intialized at line 99.

Febbe:
Nothing to fix, therefore I am closing it.

PR #4625, opened 2023-01-26
Fixup 3971c1a051293f6ef45f68a80ba49b96354062dd

danemtsov:
.clangd is a config file 😅

Also I'm not sure what the `Debug` entry is for. It was added in #3326, to me it seems like it was just added _somewhere_, probably unrelated to clangd? @aeghn @Technius

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,7 +56,7 @@ lua-*
 
 # Clangd
 compile_commands.json
-.clangd/
+.clangd
 .cache/
 Debug
 


hyperupcall:
The original `.clangd` changes was a mistake by me from [this](https://github.com/xournalpp/xournalpp/pull/4551#discussion_r1071645492) comment. Although my comment was correct, I failed to mention that `.clang`, the configuration file, is still used by clangd.

bhennion:
> The original `.clangd` changes was a mistake by me from [this](https://github.com/xournalpp/xournalpp/pull/4551#discussion_r1071645492) comment. Although my comment was partially correct, I failed to mention that `.clang`, the configuration file (not the cache directory), is still used by clangd.

Are you saying that `.clang` should be added to the list as well?

hyperupcall:
> Are you saying that .clang should be added to the list as well?

I'm saying my changes from `.clang` to `.clangd/` were incorrect and this change fixes that mistake. It doesn't need to be added separately because `.clangd` matches both files and directories whereas `.clangd/` only matches directories.

Issue #3724, opened 2022-01-17
Be able to drag pages to where you want them in the side bar


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
It's somewhat of a pain to be able to organize pages in the side bar. Currently you have to right click on the page that you want to move, and select "move up" or "move down" from the right click menu which will then move the page up or down a single position.

**Describe the solution you'd like**
I would like to be able to just click and drag the page up or down in the side panel to where I want.

rolandlo:
The would improve the UX, I agree.

forgottosave:
I think this would be very useful, too, and will take a look at this in the upcoming days.

piegamesde:
Relatedly, please provide a way to organize pages that does not involve the side bar (some entry in the menus would be sufficient). I have the side bar disabled for space reasons, and did not even realize this feature existed. It is not discoverable.

forgottosave:
Yes, I also agree that in theory every important action should be reachable in the menu. Do you think a simple entry like `Journal > Move current page up / down` would be useful?

piegamesde:
Yes, absolutely. On the one hand, I'd probably look for it under "Journal" first, because that's where the "New page" stuff is. On the other hand I should mention that I've seen the "Edit -> Arrange" menu, and then had hoped this was it just to find out it's about the drawing order.

As a bonus, a "move page to page number" entry would be nice to have. It may be of rare you so does not need a button in the side bar, but in the cases where you have to move a page to the end or something clicking "move down" twenty times is not that great :D

forgottosave:
If the maintainers approve of this, I could quickly implement these menu entries

PR #4634, opened 2023-02-05
Add entry to move pages

forgottosave:
## Goals
This Pr aims to do 2 major things:
- [x] refactor the page movement logic from `actionPerformed()` in `SidebarPreviewPages` to a more general `movePage(from, to)` in `Control`. The old placement of this logic seems very wrong.
- [ ] add a menu entry `Journal > Move page to page number`, where the user is able to move pages more easily, entering the new page number in a pop-up window (discussed in #3724)

## Additional Comments
- If desired, I can split this into 2 PRs.
- Currently the page movement is very limited in Xournal++, only allowing to move pages through a right click in the Sidebar
-  `Journal > Move page up/down` entries could also be added, but are also covered by the move to page number entry

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -82,6 +82,7 @@
 #include "undo/AddUndoAction.h"                                  // for AddU...
 #include "undo/InsertDeletePageUndoAction.h"                     // for Inse...
 #include "undo/InsertUndoAction.h"                               // for Inse...
+#include "undo/MovePageUndoAction.h"                             // for Move...
 #include "undo/MoveSelectionToLayerUndoAction.h"                 // for Move...
 #include "undo/UndoAction.h"                                     // for Undo...
 #include "util/Color.h"                                          // for oper...
@@ -1449,14 +1450,13 @@ void Control::movePage(size_t oldPos, size_t newPos) {
     }
 
     doc->lock();
-    PageRef oldPage = doc->getPage(oldPos);
+    PageRef page = doc->getPage(oldPos);
     doc->deletePage(oldPos);
-    doc->insertPage(oldPage, newPos);
+    doc->insertPage(page, newPos);
     doc->unlock();
 
-    // TODO add UndoAction
-    //UndoRedoHandler* undo = getUndoRedoHandler();
-    //undo->addUndoAction(std::make_unique<SwapUndoAction>(oldPos - 1, true, oldPage, otherPage));
+    UndoRedoHandler* undo = getUndoRedoHandler();
+    undo->addUndoAction(std::make_unique<MovePageUndoAction>(page, oldPos, newPos));
 
     firePageDeleted(oldPos);
     firePageInserted(newPos);
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1442,6 +1442,28 @@ void Control::deletePage() {
     this->win->getXournal()->forceUpdatePagenumbers();
 }
 
+void Control::movePage(size_t oldPos, size_t newPos) {
+    Document* doc = getDocument();
+    if (oldPos == newPos || doc->getPageCount() <= 1 || oldPos > doc->getPageCount() || newPos > doc->getPageCount()) {
+        return;
+    }
+
+    doc->lock();
+    PageRef oldPage = doc->getPage(oldPos);
+    doc->deletePage(oldPos);
+    doc->insertPage(oldPage, newPos);
+    doc->unlock();
+
+    // TODO add UndoAction
+    //UndoRedoHandler* undo = getUndoRedoHandler();
+    //undo->addUndoAction(std::make_unique<SwapUndoAction>(oldPos - 1, true, oldPage, otherPage));
+
+    firePageDeleted(oldPos);
+    firePageInserted(newPos);
+    firePageSelected(newPos);
+    getScrollHandler()->scrollToPage(newPos);
+}
+
 void Control::duplicatePage() {
     auto page = getCurrentPage();
     if (!page) {
--- a/src/core/control/Control.h
+++ b/src/core/control/Control.h
@@ -217,6 +217,7 @@ public:
     void appendNewPdfPages();
     void insertPage(const PageRef& page, size_t position, bool shouldScrollToPage = true);
     void deletePage();
+    void movePage(size_t oldPos, size_t newPos);
 
     /**
      * Disable / enable delete page button
--- a/src/core/undo/MovePageUndoAction.cpp
+++ b/src/core/undo/MovePageUndoAction.cpp
@@ -1,0 +1,35 @@
+#include "MovePageUndoAction.h"
+
+#include "control/Control.h"        // for Control
+#include "control/ScrollHandler.h"  // for ScrollHandler
+#include "gui/XournalppCursor.h"    // for XournalppCursor
+#include "model/Document.h"         // for Document
+#include "model/PageRef.h"          // for PageRef
+#include "undo/UndoAction.h"        // for UndoAction
+#include "util/Util.h"              // for npos
+#include "util/i18n.h"              // for _
+
+MovePageUndoAction::MovePageUndoAction(const PageRef& page, size_t oldPos, size_t newPos):
+        UndoAction("MovePageUndoAction"),
+        oldPos(page, newPos, true),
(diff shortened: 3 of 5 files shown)


forgottosave:
The first 2 commits contain the refactoring (task 1), the last 3 commits contain the actual GUI implementation (task 2).

forgottosave:
This PR should be `Ready for review` now :)

src/core/gui/dialog/MovePageDialog.h
```diff
@@ -0,0 +1,35 @@
...
+public:
+    void show(GtkWindow* parent) override;
+
+    // returns the selected page or -1 if closed
+    int getSelectedPageFrom() const;
+    int getSelectedPageTo() const;
+
+private:
+    int selectedPageFrom = -1;
+    int selectedPageTo = -1;
```
jhilmer, line 34: Consider `std::optional<size_t>` instead of `int` - aren't we using `size_t` for page numbers elsewhere?

src/core/gui/dialog/MovePageDialog.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <gtk/gtk.h>  // for GtkWindow
+
+#include "gui/GladeGui.h"  // for GladeGui
+
+class GladeSearchpath;
+
+class MovePageDialog: public GladeGui {
+public:
+    MovePageDialog(GladeSearchpath* gladeSearchPath, int currentPage, int maxPage);
```
jhilmer, line 22: `size_t` for page numbers ?

src/core/control/Control.cpp
```diff
@@ -1152,7 +1153,15 @@ void Control::selectFillAlpha(bool pen) {
 }
 
 void Control::selectPageMoveTo() {
-    std::cout << "Page Move entry not implemented yet" << std::endl;
+    MovePageDialog dlg(gladeSearchPath, getCurrentPageNo(), int(this->doc->getPageCount()));
```
jhilmer, line 1158: Is the int cast needed? `getCurrentPageNo` returns `size_t` likewise `getPageCount`.

jhilmer (changes requested):

src/core/gui/dialog/MovePageDialog.h
forgottosave: Oh yes, `size_t` is what is usually used, I don't know why I used `int` and there is no good reason for it. I'll change that.

src/core/gui/dialog/MovePageDialog.h
forgottosave: Yes, I'll change that, there is no good reason for using `int`s here

src/core/control/Control.cpp
forgottosave: No, especially not if I change the page numbers in `MovePageDialog` to `size_t` anyways

jhilmer (approved):
First page has pageNo == 1 ? So a pageNo of 0 is an invalid page number?

forgottosave:
Yes, this should be the case. These values then get overwritten anyway when closing the MovePageDialog (with the new selected values).

src/core/control/Control.cpp
```diff
@@ -1147,6 +1154,18 @@ void Control::selectFillAlpha(bool pen) {
     }
 }
 
+void Control::selectPageMoveTo() {
```
bhennion, line 1157: Could we find a more explicit name for this function?

src/core/control/Control.cpp
```diff
@@ -1147,6 +1154,18 @@ void Control::selectFillAlpha(bool pen) {
...
+    MovePageDialog dlg(gladeSearchPath, getCurrentPageNo(), this->doc->getPageCount());
+
+    dlg.show(GTK_WINDOW(this->win->getWindow()));
+    auto oldPage = dlg.getSelectedPageFrom();
+    auto newPage = dlg.getSelectedPageTo();
+
+    if (oldPage > 0 && newPage > 0) {
+        this->movePage(size_t(oldPage - 1), size_t(newPage - 1));
+    }
```
bhennion, line 1166: This should be asserted, so that a putative bug would not go unnoticed.
```suggestion
    assert(oldPage > 0 && newPage > 0);
    this->movePage(size_t(oldPage - 1), size_t(newPage - 1));
```

src/core/control/Control.cpp
```diff
@@ -1442,6 +1461,27 @@ void Control::deletePage() {
     this->win->getXournal()->forceUpdatePagenumbers();
 }
 
+void Control::movePage(size_t oldPos, size_t newPos) {
+    Document* doc = getDocument();
+    if (oldPos == newPos || doc->getPageCount() <= 1 || oldPos > doc->getPageCount() || newPos > doc->getPageCount()) {
+        return;
+    }
```
bhennion, line 1468: Same here (assert)
Also, the doc should probably be locked for getPageCount() as well.

src/core/control/Control.h
```diff
@@ -217,6 +217,7 @@ class Control:
     void appendNewPdfPages();
     void insertPage(const PageRef& page, size_t position, bool shouldScrollToPage = true);
     void deletePage();
+    void movePage(size_t oldPos, size_t newPos);
```
bhennion, line 220: This would benefit a comment (in particular stating that the page positions start at 0).

src/core/gui/dialog/MovePageDialog.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+MovePageDialog::~MovePageDialog() = default;
+
+auto MovePageDialog::getSelectedPageFrom() const -> size_t { return this->selectedPageFrom; }
+auto MovePageDialog::getSelectedPageTo() const -> size_t { return this->selectedPageTo; }
+
+void MovePageDialog::show(GtkWindow* parent) {
+    gtk_window_set_transient_for(GTK_WINDOW(this->window), parent);
+    gtk_entry_set_activates_default(GTK_ENTRY(get("movePageFrom")), TRUE);
+    gtk_entry_set_activates_default(GTK_ENTRY(get("movePageTo")), TRUE);
```
bhennion, line 23: This could go in the Glade file

src/core/undo/MovePageUndoAction.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "util/i18n.h"              // for _
+
+MovePageUndoAction::MovePageUndoAction(const PageRef& page, size_t oldPos, size_t newPos):
+        UndoAction("MovePageUndoAction"),
+        oldPos(page, newPos, true),
+        newPos(page, oldPos, false) {
+    this->page = page;
+}
+
+MovePageUndoAction::~MovePageUndoAction() { this->page = nullptr; }
```
bhennion, line 12: Is there a hidden use to setting `page = nullptr` here?
```suggestion
MovePageUndoAction::~MovePageUndoAction() = default;
```

src/core/undo/MovePageUndoAction.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+auto MovePageUndoAction::undo(Control* control) -> bool {
+    bool oldRes = oldPos.undo(control);
+    bool newRes = newPos.undo(control);
```
bhennion, line 16: I'm a bit confused here. Is this reinserting the page at its old position first and then removing it from the new one? Should itt be the other way around?

Also, more explicit names for oldPos and newPos could help.

src/core/undo/MovePageUndoAction.cpp
```diff
@@ -0,0 +1,28 @@
...
+    return oldRes && newRes;
+}
+
+auto MovePageUndoAction::redo(Control* control) -> bool {
+    bool newRes = newPos.redo(control);
+    bool oldRes = oldPos.redo(control);
```
bhennion, line 22: Same question about the order here

bhennion (commented):
Here are a couple of comments and nitpicks.
I haven't had time to try it out yet. I'll try to do that this week. (Remind me if I forget)

forgottosave:
Hi, I've been away for quiet a while. Sadly my life was very busy and will continue to be so for the next 1-2 months. I'll still try to look at your change suggestions as soon as I can (and maybe do a big rebase, depending on what changed in the meantime)

PR #4637, opened 2023-02-06
Control::checkChangedDocument: notify page changes to subset of listeners

hrdl-github:
Continuation of  #4611. I'm splitting this one off as suggested by @bhennion to keep the discussion focused and get some feedback.

As before, I would like to avoid rerendering the entire page after most changes for performance reasons. Instead of using the general-purpose implementation `DocumentHandler` I introduce a similar mechanism in `Control`, which is only used for notifying a subset of `DocumentListener` instances (at the moment all instances of subclasses of `AbstractSidebarPage`).

This approach does not interfere with the existing observer pattern or the general rerendering logic.


rolandlo (approved):
LGTM. The name `changedDocumentListeners` sounds good enough to me. It's good though that you added a comment.

PR #4612, opened 2023-01-21
Fix rendering artefacts in standard eraser

bhennion:
This should fix #4610 .

PR #4659, opened 2023-02-18
Fix rendering artefacts in standard eraser - 2nd try

bhennion:
This PR replaces #4612, with a more precise computation of the rerendered boxes.

### DIFF
--- a/src/core/control/tools/EraseHandler.cpp
+++ b/src/core/control/tools/EraseHandler.cpp
@@ -119,7 +119,6 @@ void EraseHandler::eraseStroke(Layer* l, Stroke* s, double x, double y, Range& r
             doc->unlock();
             this->eraseUndoAction->addOriginal(l, s, pos);
             erasable->beginErasure(intersectionParameters, range);
-            paddedEraserBox.addToRange(range);
         }
     } else {
         /**
--- a/src/core/model/PathParameter.h
+++ b/src/core/model/PathParameter.h
@@ -23,7 +23,7 @@ struct PathParameter {
     bool operator<=(const PathParameter& p) const { return index < p.index || (index == p.index && t <= p.t); };
     bool operator>=(const PathParameter& p) const { return index > p.index || (index == p.index && t >= p.t); };
 
-    bool isValid() const { return t <= 1.0 && (t > 0.0 || (t == 0.0 && index == 0U)); }
+    bool isValid() const { return t <= 1.0 && t >= 0.0; }
 
     size_t index;
     double t;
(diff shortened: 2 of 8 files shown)


rolandlo:
There are still some rendering artifacts relating to segments shrinking in width while they are erased, which happens for decreasing pressure values along the stroke.
Try to erase the (artificially created) stroke in the file below (from bottom right to top left, i.e. from fat to thin).
[Test.zip](https://github.com/xournalpp/xournalpp/files/10774744/Test.zip)

bhennion:
Thanks for the catch!
I pushed a fix:

https://user-images.githubusercontent.com/17818041/219881631-ca39d36f-4fa2-4910-ba80-77a6b778c793.mp4

rolandlo:
Great, no more artefacts! However the following test FAILED:
	 55 - ErasableStroke.testGetStrokes (Failed)

rolandlo (approved):
LGTM. Great fix!

tattsan:
Thank you for the fix! I've also seen similar artefacts in 'delete stroke' eraser.

https://user-images.githubusercontent.com/6355486/219945049-25ff2ade-1040-4442-a666-cca450d7869a.mp4
 
These artefacts will disappear after a few seconds, and not harmful. I don't know if they are  due to the same cause.

rolandlo:
> Thank you for the fix! I've also seen similar artefacts in 'delete stroke' eraser.
> 
> These artefacts will disappear after a few seconds, and not harmful. I don't know if they are due to the same cause.

I don't get these artefacts with the (latest) version from this PR in 'delete stroke' mode. Did you observe that with the PR version or with some other version? In the first case please add a .xopp-file where this happens.

tattsan:
> I don't get these artefacts with the (latest) version from this PR in 'delete stroke' mode. Did you observe that with the PR version or with some other version? In the first case please add a .xopp-file where this happens.

I used the latest version of this PR. If I reopen the saved .xopp-file with strokes, no artefacts appear when I erase the strokes. The artefacts appears only when erasing the newly written strokes, so the saved file won't reproduce the phenomenon. The following settings will make these artefacts stand out:
* Use 'Input stabilization' and let 'Finalize the stroke' on.
* Increase the value of 'Minimum Pressure'.

bhennion:
I guess this is indeed somewhat independent of this PR (although it also affects the eraser). Let's handle it in another PR.

I rebased on master and cleaned up the commit tree. Merging in 48 hours unless an objection is raised.

tattsan:
Thank you for clarifiying the issue.  I'll test #4668 !

Issue #3669, opened 2021-12-16
Add toolbar icons and buttons for every possible action


bhennion:
This is a meta issue about missing toolbar icons.
There should be a toolbar icon and button for every possible action available in the dropdown menus.

See #3612 #3514 #2517 for requests about specific actions.

See also #2936 and #2431 for requests for custom buttons.

MichelReims:
I Found a freeware for add icons forevery possible action with shortcuts :
https://github.com/Martenfur/TabletFriend

rolandlo:
#4671 has been merged, which means it is possible to add icons that trigger plugin functions. This also substitutes the TabletFriend freeware mentioned by @MichelReims 

@bhennion Do you still want to add icons and buttons for every possible action?

MichelReims:
can you give several complete examples showing how this new possibility is used ?

rolandlo:
@MichelReims 
Here is one example: https://github.com/xournalpp/xournalpp/commit/b048f76e20012f7fce3200dddff7ab7c44621e8b

Here is a shorter one:
I will skip the `plugin.ini`. It will simply refer to `main.lua` as usual.

`main.lua`:
```lua
function initUi()
  app.registerUi({["menu"] = "fine black line", ["callback"] = "cb", ["toolbarId"] = "FINE_BLACK_LINE", ["iconName"] = "document-edit"})
end

function cb()
  app.uiAction({["action"]="ACTION_TOOL_PEN"})
  app.uiAction({["action"]="ACTION_RULER"})  
  app.uiAction({["action"]="ACTION_TOOL_PEN_SIZE_FINE"})
  app.changeToolColor({["color"] = 0x000000, ["selection"] = false})
end
```
This assumes that your icon theme has an icon called "document-edit". Otherwise use a different icon. You can even put a custom icon into the folder where Xournal++ stores its icons (just don't prepend it with `xopp`) or in the Adwaita icon theme folder. 

In `toolbar.ini` add
```
Plugin::FINE_BLACK_LINE
```
to the toolbar you use, e.g. like in the Xournal++ toolbar on the top left
```term
[Xournal++]
name = Xournal++
name[de] = Xournal++
name[it] = Xournal++
name[ru] = Xournal++
[...]

PR #4671, opened 2023-02-22
feature: add toolbar buttons via plugins

### LINKED ISSUE #2431: Customizable icons for selecting more pen properties at once (like in stylus buttons customization)
First of all thank you very much for your hard work, Xournal++ is an awesome app!
I've used Xournal++ on my tablet with a stylus for a few days now and I know that I can use the buttons on my pen to do a very similar thing but it would be great to have a way to select more pen properties at once, so that we can select **more** than N+1 templates real quick (where N is the number of buttons that our pen has + 1 for the tip).
For example if I want to write a title during my note-taking I'd use a thick red stroke, when I start writing normal text I'd use a fine black stroke, when I write questions I'd use a medium blue stroke, when I draw graphs I sometimes use a fine red dotted lines and so on.
If I only have 2 buttons and the tip I can only have 3 custom templates (and no place for the eraser), so for the others I have to enable every single property manually and this is sometimes time-consuming during lectures.

It would be fantastic to have more quick-select fully customizable icons that can be inserted in the floating toolbar or anywhere else to have access to all the templates.

### LINKED ISSUE #2936: Custom toolbar icon with custom event
It would be nice to have to possibility to create custom toolbar icon to change more than one thing.
With Lua i can create a shortcut event that change the pen size, the color and so on, but, if i use xournal with some tablet device like the Surface or some 2in1 i cannot use the keyboard and i have to select each option one by one. 

Add with Lua some custom icon that can performe the same command that a custom shortcut can (like uiAction, changeToolColor, etc...)

Add a set of icon (for example 5) that when clicked acts like a keyboard shortcut that a user can customize.
Then the user will create the Lua plugin with the shortcut.

### LINKED ISSUE #4669: Tool to trigger plugin
Hello guys,
I was trying to adapt the screenshot plugin for my workflow but in order to really make it work I would like to trigger the plugin from my toolbar. Thus I was wondering if you could add a tool for the toolbar to trigger a chosen plugin. You could even take this idea further and create a tool to run a custom script or something like that. I find it quite frustrating that the plugin implementation is so hidden and difficult to access. I think that this tool to trigger plugins would make it much more attractive to use plugins.
Besides that it would also be nice if users could create custom tools in a similar fashion to plugins. I don't know if you are already planing to implement this but the toolbar customizer kind of already suggests it with the heading "Default Tools". It would be wonderful if you could implement some method to add more tools to this list.

Cheers.

rolandlo:
Fixes #4669. Fixes #2936. Fixes #2431. Also addresses #3669.

This PR allows registering toolbar buttons (with any ID and any icon in the icon theme) that can be added to any toolbar.

These custom tools (executing plugin callback functions) can be added to the toolbar either by customizing the toolbar in the GUI or by editing `toolbar.ini`. 

The plugin adds the toolbar via e.g. 
```lua
app.registerUi({callback ="blueDashedPen", toolbarId="CUSTOM_PEN_1", iconName="bluePenIcon"})
```

It's also possible to combine adding a toolbar button with adding a menu entry:

```lua
app.registerUi({["menu"] = "Mirror strokes horizontally", ["callback"] = "cb", ["accelerator"] = "<Control>t", ["toolbarId"] = "MIRROR", ["iconName"] = "view-mirror-symbolic"})
```

Here `view-mirror-symbolic` is an icon that actually exists. I found via Gnome's symbol library (it's part of the Adwaita icon theme). 

grafik

A little demo (see the icon on the top left, right below the `File` menu):

https://user-images.githubusercontent.com/40485433/220704181-bbcaa122-db17-4366-b91a-fc6aec1cc06d.mp4

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -3014,6 +3014,10 @@ void Control::clipboardPaste(Element* e) {
     win->getXournal()->setSelection(selection);
 }
 
+void Control::registerPluginToolButtons(ToolMenuHandler* toolMenuHandler) {
+    pluginController->registerToolButtons(toolMenuHandler);
+}
+
 void Control::clipboardPasteXournal(ObjectInputStream& in) {
     auto pNr = getCurrentPageNo();
     if (pNr == npos && win != nullptr) {
--- a/src/core/control/Control.h
+++ b/src/core/control/Control.h
@@ -64,7 +64,7 @@ class Settings;
 class TextEditor;
 class XournalScheduler;
 class ZoomControl;
-
+class ToolMenuHandler;
 class Control:
         public ActionHandler,
         public ToolListener,
@@ -317,6 +317,9 @@ public:
     void deleteSelection() override;
 
     void clipboardPaste(Element* e);
+
+public:
+    void registerPluginToolButtons(ToolMenuHandler* toolMenuHandler);
 
 protected:
     /**
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -86,6 +86,9 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
     // "watch over" all events
     g_signal_connect(this->window, "key-press-event", G_CALLBACK(onKeyPressCallback), this);
 
+    // need to create tool buttons registered in plugins, so they can be added to toolbars
+    control->registerPluginToolButtons(this->toolbar.get());
+
     createToolbar();
 
     setToolbarVisible(control->getSettings()->isToolbarVisible());
--- a/src/core/gui/toolbarMenubar/PluginToolButon.cpp
+++ b/src/core/gui/toolbarMenubar/PluginToolButon.cpp
@@ -1,0 +1,25 @@
+#include <utility>  // for move
+
+#include "PluginToolButton.h"
+
+PluginToolButton::PluginToolButton(ActionHandler* handler, ToolbarButtonEntry* t):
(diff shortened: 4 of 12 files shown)


bhennion:
That's very nice!
I have only glanced at the code, but I have a question/concern: giving the user the ability to create new toolbarId's (the strings used in toolbar.ini), we could be opening the door to some form of spoofing: create a nasty plugin with toolbarId being, say, `SAVE` (and matching icon).

What happens then? Are we guarantied that when we click on the SAVE button, it saves and does not execute the plugin's callback?
I see two ways of ensuring that:
1. Making certain that stock buttons are prefered to plugin buttons in case of competing toolbarId's + issueing an error if this happens
2. Adding a PLUGIN:: namespace to all toolbarIds created via the plugin interface. (so basically add a `toolbarId = "PLUGIN::" + toolbarId;` somewhere).

I'd say 2. would be better, safer and likely to get unwillingly destroyed by our future selves.
The option 1. also has the drawbacks
- Future bug reports "I can't get my plugin to create a button" due to the author not realizing the toolbarId was already taken.
- if an update adds new stock toolbar buttons, and the added toolbarIds then break some user's plugin.

rolandlo:
I like your second solution and implemented it. Even from a user perspective it's cleaner looking at `toolbar.ini`.
Talking about security of the plugin system, we should also ensure that all plugins loaded from the `.config` directory are disabled by default.

jhilmer (approved):

rolandlo:
Rebased on current master and squashed the commits related to the actual feature.

Dropping the second commit (test plugin) and merging the first commit in 48 hours if no objections are raised.

PR #4672, opened 2023-02-22
Fix bounding box for pressure-sensitive strokes

### LINKED ISSUE #4668: Incorrect bounding box for strokes with varying pressure
**Affects versions :**
 - OS: GNU/Linux, probably others as well
 - Version of Xournal++: https://github.com/xournalpp/xournalpp/commit/517ce512a709d2951e84d7427a7d5399959270f2

**Describe the bug**
Rendering issue in connection to pressure-sensitive strokes. Affects eraser, undo, and redo and possibly other actions.

**To Reproduce**
1. Activate pressure sensitivity and set pressure multiplier to 4, as the effect is less visible at 1.
2. In my test: set the brush size to large and the eraser size to small. Set zoom to maximum.
3. Create a single "stroke" using the stylus, interleave it with button presses to activate the stroke-based eraser without lifting the stylus from the tablet. Stop until a stroke that causes artefacts was deleted.

**Expected behavior**
No artefacts when handling above stroke.

**Screenshots of Problem**
Erasure: https://user-images.githubusercontent.com/31923882/220463326-4ecf8ef7-f472-4a20-b3fa-716ac7b1e9f2.mp4

Redo / undo: https://user-images.githubusercontent.com/31923882/220464807-81d18891-8d19-4b92-982b-92410bac3a45.mp4

Stroke: [bug_stroke.zip](https://github.com/xournalpp/xournalpp/files/10798009/bug_stroke.zip)

**Additional context**
Originally mentioned here: https://github.com/xournalpp/xournalpp/pull/4659#issuecomment-1438169890

hrdl-github:
Fixes #4668.

### DIFF
--- a/src/core/model/Stroke.cpp
+++ b/src/core/model/Stroke.cpp
@@ -47,8 +47,7 @@ using xoj::util::Rectangle;
 #endif
 
 template <typename Float>
-constexpr void updateBounds(Float& x, Float& y, Float& width, Float& height, Rectangle<Float>& snap, Point const& p,
-                            double half_width) {
+constexpr void updateBoundingBox(Float& x, Float& y, Float& width, Float& height, Point const& p, double half_width) {
     {
         Float x2 = x + width;
         Float y2 = y + height;
@@ -60,6 +59,11 @@ constexpr void updateBounds(Float& x, Float& y, Float& width, Float& height, Rec
         width = x2 - x;
         height = y2 - y;
     }
+
+}
+
+template <typename Float>
+constexpr void updateSnappedBounds(Rectangle<Float>& snap, Point const& p) {
     {
         Float snapx2 = snap.x + snap.width;
         Float snapy2 = snap.y + snap.height;
@@ -249,8 +253,8 @@ void Stroke::addPoint(const Point& p) {
     if (hasPressure()) {
         updateBoundsLastTwoPressures();
     } else {
-        updateBounds(Element::x, Element::y, Element::width, Element::height, Element::snappedBounds, p,
-                     0.5 * this->width);
+        updateBoundingBox(Element::x, Element::y, Element::width, Element::height, p, 0.5 * this->width);
+        updateSnappedBounds(Element::snappedBounds, p);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/model/Stroke.cpp
```diff
@@ -392,14 +417,16 @@ void Stroke::setLastPressure(double pressure) {
         assert(pressure != Point::NO_PRESSURE);
         Point& back = this->points.back();
         back.z = pressure;
-        updateBounds(Element::x, Element::y, Element::width, Element::height, snappedBounds, back, 0.5 * pressure);
```
bhennion, line 395: Why is this necessary? The last point's pressure is not used in rendering, unless the stroke has exactly one point, in which case you end up calling `updateBounds` from `updateBoundsLastTwoPressures`

src/core/model/Stroke.cpp
```diff
@@ -379,6 +386,24 @@ auto Stroke::getAvgPressure() const -> double {
...
+
+    Point& p = this->points.back();
+    auto const pointCount = this->getPointCount();
+    if (pointCount >= 2) {
+        Point& p2 = this->points[pointCount - 2];
+        double pressure = p2.z;
+
+        updateBounds(Element::x, Element::y, Element::width, Element::height, snappedBounds, p, 0.5 * pressure);
+        updateBounds(Element::x, Element::y, Element::width, Element::height, snappedBounds, p2, 0.5 * pressure);
+    } else if (pointCount == 1) {
```
bhennion, line 402: This case should be totally excluded from `updateBoundsLastTwoPressures` (like with an assert).

src/core/model/Stroke.cpp
```diff
@@ -379,6 +386,24 @@ auto Stroke::getAvgPressure() const -> double {
...
+    }
+
+    Point& p = this->points.back();
+    auto const pointCount = this->getPointCount();
+    if (pointCount >= 2) {
+        Point& p2 = this->points[pointCount - 2];
+        double pressure = p2.z;
+
+        updateBounds(Element::x, Element::y, Element::width, Element::height, snappedBounds, p, 0.5 * pressure);
+        updateBounds(Element::x, Element::y, Element::width, Element::height, snappedBounds, p2, 0.5 * pressure);
```
bhennion, line 401: That's suboptimal: `snappedBounds` gets needlessly updated on the second call: p2 already lies in `snappedBounds`

bhennion (changes requested):
Thanks for the fix! You got to the core of the issue here.

I have a couple of nitpicks (see below)

bhennion (approved):
LGTM! Thanks for the great job!

Merging in 48 hours unless an objection is raised

PR #4674, opened 2023-02-23
Add spacer for drag&drop in toolbar customization

### LINKED ISSUE #4644: Need a SPACER for toolbar customizing

I found the default toolbar configurations "Portrait" and "Xournal++" do have some tools on the right hand of their bottom toolbars, which looks good. I checked the file "toolbar.ini" and found there might be a SPACER in the middle. But when I try to customize my toolbar, all I can do is easily remove the invisible SPACER but cannot add it again from the tool box, because there wasn't one. (As for the toolbox, there are tools and colors that not on the toolbar, and a SEPARATOR, but not a SPACER). Also, I cannot find a ini file to customize my tool bar with codes.(My system is windows10, Xournal++ v1.1.3)

Help me find a way to add a SPACER to my toolbar.

**Additional context**

rolandlo:
Fixes #4644 

This PR adds an icon for the spacer to the toolbar customization dialog (painted in `ToolbarSeparatorImage.cpp` as a variant of the separator icon) and allows for drag&drop for the spacer.
grafik

### DIFF
--- a/src/core/gui/toolbarMenubar/icon/ToolbarSeparatorImage.cpp
+++ b/src/core/gui/toolbarMenubar/icon/ToolbarSeparatorImage.cpp
@@ -3,25 +3,44 @@
 #include <cairo.h>        // for cairo_create, cairo_destroy, cairo_image_su...
 #include <gdk/gdk.h>      // for gdk_pixbuf_get_from_surface
 #include <glib-object.h>  // for g_object_unref
+#include <util/Color.h>   // for Colors::red, Colors::gray
+#include <util/Util.h>    // for cairo_set_source_rgbi
 
-auto ToolbarSeparatorImage::newImage() -> GtkWidget* {
-    GdkPixbuf* pixbuf = ToolbarSeparatorImage::getNewToolPixbuf();
+constexpr double LINE_WIDTH = 5.;
+constexpr double MARGIN = 3.;
+constexpr double TAIL_SIZE = 5.;
+
+auto ToolbarSeparatorImage::newImage(SeparatorType separator) -> GtkWidget* {
+    GdkPixbuf* pixbuf = ToolbarSeparatorImage::getNewToolPixbuf(separator);
     GtkWidget* w = gtk_image_new_from_pixbuf(pixbuf);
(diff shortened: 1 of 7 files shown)


rolandlo:
It works now. Previously there was an issue with data (SeparatorData) that was lost.

src/core/gui/toolbarMenubar/icon/ToolbarSeparatorImage.cpp
```diff
@@ -3,25 +3,44 @@
...
+
+auto ToolbarSeparatorImage::newImage(SeparatorType separator) -> GtkWidget* {
+    GdkPixbuf* pixbuf = ToolbarSeparatorImage::getNewToolPixbuf(separator);
     GtkWidget* w = gtk_image_new_from_pixbuf(pixbuf);
     g_object_unref(pixbuf);
     return w;
 }
 
-auto ToolbarSeparatorImage::getNewToolPixbuf() -> GdkPixbuf* {
+auto ToolbarSeparatorImage::getNewToolPixbuf(SeparatorType separator) -> GdkPixbuf* {
```
bhennion, line 20: I guess we'd better make an actual icon here, instead of drawing it with cairo.
The icons would then be theme-dependent: if the background color is the same as Colors::gray, we won't see anything.

bhennion (approved):
LGTM!

I have a remark about the icons (it holds also for the original separator): they should be icons and not be hardcoded. I can be delayed to a follow-up PR if you want.

src/core/gui/toolbarMenubar/icon/ToolbarSeparatorImage.cpp
rolandlo: Thanks for the review! 
I will open an issue about the icons and postpone it to another PR. Maybe someone with better designing skills can create them. The color of the widget background is one thing. The other one is the color of the page, when you drag&drop. So the icon should be clearly visible on any page color.

rolandlo: Merging in 24 hours if no objections are raised.

PR #4679, opened 2023-02-25
Ensure char arrays are null terminated

jhilmer:
Ensure char arrays are null terminated before they are used to build a std::string. Help fix one of the issues valgrind complained about.


bhennion (approved):
LGTM!

PR #4680, opened 2023-02-25
Fix GList memory leak

jhilmer:
gtk_container_get_children returns a GList which was newer deallocated In this context the return value is used as a boolean - which always is true and therefore redundant.


bhennion (approved):
LGTM. Thanks for the patch!

Merging in 24 hours.

PR #4681, opened 2023-02-25
Fix #4584 segfault when exporting from cli with missing bg

### LINKED ISSUE #4584: Segmentation Fault when exporting via CLI and background PDF doesn't exist
**Affects versions :**
 - OS: tested on Linux (nixos and Ubuntu)
 - (Linux only) Desktop environment: Using KDE Plasma / X11 but irrelevant for this bug because it's just CLI
 - Which version of libgtk do you use: 3.24.34 and 3.24.33
 - Version of Xournal++: 1.1.1
 - Installation method: nix and apt

**Describe the bug**
When exporting with `xournalpp -p`, if the background pdf is not found, the program segfaults (Signal 11)
```
** (xournalpp:4221): WARNING **: 12:21:48.727: [Crash Handler] Crashed with signal 11
** (xournalpp:4221): WARNING **: 12:21:48.728: [Crash Handler] Wrote crash log to: /home/kruemmelspalter/.cache/xournalpp/errorlogs/errorlog.20230113-122148.log
```

**To Reproduce**
1. Open a pdf with File > Annotate PDF or `xournalpp xyz.pdf`
2. Save the file (e. g. as `xyz.xopp`)
3. Delete the PDF
4. Try to export the xopp file to pdf (`xournalpp -p out.pdf xyz.xopp`)

**Expected behavior**
A clear error message instead of just a traceback

jhilmer:
Fix #4584

Now we give error message on the console stderr, and exit the process with -2.

Note if we return -2 instead of exiting - the gui startup and try to do the exporting ???

This fix ensure we dont segfault. But what should the correct behavior be?

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -237,6 +237,19 @@ void checkForEmergencySave(Control* control) {
     gtk_widget_destroy(dialog);
 }
 
+namespace {
+void exitOnMissingPdfFileName(const LoadHandler& loader) {
+    if (!loader.getMissingPdfFilename().empty()) {
+        auto msg =
+                FS(_F("The background file \"{1}\" could not be found. It might have been moved, renamed or deleted.") %
+                   loader.getMissingPdfFilename());
+        std::cerr << msg << std::endl;
+        exit(-2);
+    }
+}
+}  // namespace
+
+
 /**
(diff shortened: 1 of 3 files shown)


bhennion (approved):
LGTM! Merging in 24 hours.

bhennion (approved):
LGTM! Merging in 24 hours.

PR #4683, opened 2023-02-26
Use `optional` instead of -1 for exportImg parameters

jhilmer:
Many places '-1' is used as 'none'. Use 'std::optional' instead. Batch exporting to image is one of those places.

### DIFF
--- a/src/core/control/ExportHelper.cpp
+++ b/src/core/control/ExportHelper.cpp
@@ -34,8 +34,9 @@ namespace ExportHelper {
  *
  * @return 0 on success, -3 on export failure
  */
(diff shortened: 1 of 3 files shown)


rolandlo (changes requested):
According to my tests this breaks export from CLI.
I tested via
```
build/xournalpp --export-png-width=1234 -i test_img_1234.png test.xopp
``` 
vs
```
 xournalpp --export-png-width=1234 -i test_img_1234.png test.xopp
```
with some random `test.xopp` file.

So please check CLI export.

rolandlo:
Closing for now due to lack of activity.

PR #4685, opened 2023-02-26
Simplify custom export job

jhilmer:
My first idea was to remove the use of new/delete for ExportType object. Why do we use dynamic allocation for this? I search the git history for the PR where it went in, and stubbled over the following comment:

https://github.com/xournalpp/xournalpp/pull/594#discussion_r243860463

And therefore changed it to not use dynamic allocation for the ExportType.


bhennion (approved):
Good job! (And a thorough investigation is always great!)

LGTM, merging in 24 hours.

PR #4690, opened 2023-03-01
Fix the filename suggestions when annotating PDFs

MikeWalrus:
The current implementation of `createSaveFilename`(#4599) causes the suggested filename in the save dialog to be ".pdf" whenever a PDF is opened. This fixes the problem and also removes the pdf extension as the file will be saved as "*.xopp" instead of "*.pdf.xopp".

### DIFF
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -138,6 +138,11 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
         return p;
     }
     if (!pdfFilepath.empty()) {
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
Thanks for the fix of this regression! 

rolandlo:
@MikeWalrus Please drop the second commit. It doesn't have anything to do with the first one. Open a new PR instead.

MikeWalrus:
Accidentally pushed to the wrong branch. Sorry.

Issue #2652, opened 2021-01-11
Up/Down buttons on the Search widget don't navigate through search results


yurivict:
**Affects versions :**
 - OS: FreeBSD 12.2
 - (Linux only) Desktop environment: Xfce4
 - Which version of libgtk do you use: 3.24.24
 - Version of Xournal++: 1.0.20
 - Installation method: package

**Describe the bug**
Large single-page PDF is loaded. The PDF was produced by the ```dot``` utility from ```GraphViz```.
Ctrl-F brings focus to the search widget.
Typing some text shows the prompt: "Text 8 times found on this page" (please fix the grammar too!)
But pressing Up/Down buttons doesn't navigate between the text occurrences on the page.

**Expected behavior**
Navigation should be working from the Search widget.

rolandlo:
Not sure it is a bug. It could be intended behaviour. Just markup all occurences on one page. Pressing Up/Down will relocate you to the next/previous page, where the search string appears.

yurivict:
> Pressing Up/Down will relocate you to the next/previous page, where the search string appears.

If it can relocate to the page but not to the other location on the page - this is quite a confusing behavior.

Web browsers, for example,  have a similar "search" feature and it simply relocates the view to the location.

Technius:
I agree with @yurivict, I had to use the search feature yesterday and it is definitely not intuitive. It only counts text on the same page, and it jumps around strangely.

PR #4694, opened 2023-03-04
Improve text search

### LINKED ISSUE #3624: Search text goes to page but match is off screen
**Affects versions :**
 - OS: Ubuntu 21.10
 - (Linux only) Desktop environment: Gnome-Wayland
 - Which version of libgtk do you use: libgtk: 3.24.30
 - Version of Xournal++: 1.1.0
 - Installation method: PPA

**Describe the bug**
I use the search feature to find text. When I press <ENTER> or click the down arrow the page changes and is centered vertically on the screen.  The page is not scrolled to the place where the text is, so if I'm zoomed in then I don't see the matched text. I have to scroll up, and then down to find the highlighted box around the matched text. 

**To Reproduce**
1. Open a PDF
2. Maximize the window
3. Fit the width of the page to the window
4. Search for a word in a PDF that may appear on some pages only at the bottom or top of the page.

**Expected behavior**
I would expect to see the search term highlighted and visible.  Instead I have scroll, either up or down, not sure which, to find the match.

**Screenshots of Problem**
image

**Additional context**
I can understand that one may not want to center every match on the screen, as other PDF viewers do in Linux. What I would propose is to center the next match that is currently not visible.

MikeWalrus:
Fixes #3624.
I'll try to address #2652, which requires a bit more effort.


MikeWalrus:
Before:
before

After:
after

rolandlo (approved):
Thanks a lot for this great contribution! Except of the includes in `SearchBar.cpp` everything looks good, nice and clean. The feature works fine and it makes searching text on PDF's much more intuitive and useful.

rolandlo (approved):

rolandlo:
Thanks again @MikeWalrus! Your contribution improves search a lot. I just merged it.

MikeWalrus:
@rolandlo Thank you for reviewing and merging!

PR #4695, opened 2023-03-05
Use RAII MainWindow toolbarWidgets

jhilmer:
Also changed to RAII in ToolbarDragDropHandler as it was using the toolbarWidgets of MainWindow.

### DIFF
--- a/src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHandler.cpp
+++ b/src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHandler.cpp
@@ -16,37 +16,25 @@
 
 ToolbarDragDropHandler::ToolbarDragDropHandler(Control* control): control(control) {}
 
-ToolbarDragDropHandler::~ToolbarDragDropHandler() { clearToolbarsFromDragAndDrop(); }
+ToolbarDragDropHandler::~ToolbarDragDropHandler() = default;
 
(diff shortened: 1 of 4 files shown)


src/core/gui/dialog/toolbarCustomize/ToolbarDragDropHandler.cpp
```diff
@@ -16,37 +16,28 @@
...
-    this->toolbars[len] = nullptr;
+    this->toolbars.clear();
+    this->toolbars.reserve(len);
 
-    for (int i = 0; i < len; i++) {
-        GtkWidget* w = widgets[i];
-        this->toolbars[i] = new ToolbarAdapter(w, win->getToolbarName(GTK_TOOLBAR(w)),
-                                               control->getWindow()->getToolMenuHandler(), control->getWindow());
+    for (size_t i = 0; i < len; i++) {
+        GtkWidget* w = win->getToolbarWidget(i);
```
bhennion, line 29: Can we get a const reference for the `std::array` and iterate on it directly here?

jhilmer: Yes, and then we don't have a need for the above `getToolbarCount`

bhennion (approved):
LGTM! Merging in 24 hours

jhilmer:
Squashed the above, so it is ready to merge

PR #4698, opened 2023-03-06
Fix lifetime of references

### LINKED ISSUE #4697: g++ -O2 and -O1 break rerendering of the vertical tool
**Affects versions :**
 - OS Arch Linux (x86-64 and arm64)
 - GTK 3.24.37
 - Version of Xournal++: https://github.com/xournalpp/xournalpp/commit/b8733e9542f87835dbbd8fe035eb557bb5141f13
 - Installation method: AUR (xournalpp-git): g++ 12.2.1

**Describe the bug**
`makepkg` compiles packages using the compiler option `-O2` by default. When using the vertical tool the blue area is often not cleared as `minY`  and `maxY` in https://github.com/xournalpp/xournalpp/blob/b8733e9542f87835dbbd8fe035eb557bb5141f13/src/core/view/overlays/VerticalToolView.cpp#L80 seemingly are `0.0`. Recompiling with `-O0` fixes this.

**To Reproduce**
1. Compile with `-O1`, e.g. via `CXXFLAGS=-O1 cmake .. && cmake --build . -j $(nproc)`
2. Run xournalpp: `./xournalpp`
3. Use the vertical tool

**Expected behavior**
The blue area of the vertical tool vanishes after releasing the mouse button.

** Actual behavior**
The blue area of the vertical does not vanish after releasing the mouse button.

hrdl-github:
Fixes #4697.

### DIFF
--- a/src/core/view/overlays/VerticalToolView.cpp
+++ b/src/core/view/overlays/VerticalToolView.cpp
@@ -77,7 +77,9 @@ void VerticalToolView::on(VerticalToolView::SwitchDirectionRequest) {
 }
 
 void VerticalToolView::deleteOn(VerticalToolView::FinalizationRequest) {
-    auto [minY, maxY] = std::minmax(toolHandler->getStartY(), toolHandler->getEndY());
+    double startY = toolHandler->getStartY();
+    double endY = toolHandler->getEndY();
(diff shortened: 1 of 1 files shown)


bhennion (approved):
LGTM! Merging in 24 hours.
Thanks for the fix!

PR #4702, opened 2023-03-08
Unit test for model/StrokeStyle and refactor

jhilmer:
Tried to first put existing code under unit test and the refactor to use e.g. map, vector, sstring e.g. instead of mix of glib function and new/delete to allocate dynamic arrays.

### DIFF
--- a/src/core/model/StrokeStyle.cpp
+++ b/src/core/model/StrokeStyle.cpp
@@ -1,9 +1,11 @@
 #include "StrokeStyle.h"
 
 #include <cstring>  // for memcmp, strcmp, strncmp
+#include <iomanip>  // for setprecision
+#include <iterator> // for ostream_iterator
+#include <map>      // for map
+#include <sstream>  // for istringstream
 #include <vector>   // for vector
-
-#include <glib.h>  // for g_ascii_strtod, g_free, g_strdup_printf
 
 #include "model/LineStyle.h"  // for LineStyle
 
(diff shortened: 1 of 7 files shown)


src/core/model/StrokeStyle.cpp
```diff
@@ -1,91 +1,83 @@
...
+const std::map<std::string, std::vector<double>> predefinedPatterns = {
+    { "dash", {6, 3} }, 
+    { "dashdot", {6, 3, 0.5, 3} },
+    { "dot", {0.5, 3}}
+};
 
-    if (strncmp("cust: ", style, 6) != 0) {
-        return LineStyle();
-    }
+auto formatStyle(const double* dashes, int count) -> std::string {
```
bhennion, line 26: This function is not very optimal. I think you could go a little bit deeper into the refactoring and change LineStyle to replace the pair `dashes, count` by an `std::vector<double>`.
The way the LineStyle is used down the road is by passing dashes and count to cairo. This can be done by accessing the std::vector's data directly.

jhilmer: Yes,  I am working on LineStyle also. I will add it to this.

src/core/model/StrokeStyle.cpp
```diff
@@ -1,91 +1,82 @@
...
-        style.setDashes(def, sizeof(def) / sizeof((def)[0])); \
-        return style;                                         \
-    }
+namespace {
 
-auto StrokeStyle::parseStyle(const char* style) -> LineStyle {
-    PARSE_STYLE("dash", dashLinePattern);
-    PARSE_STYLE("dashdot", dashDotLinePattern);
-    PARSE_STYLE("dot", dotLinePattern);
+constexpr auto CUST_KEY = "cust: ";
```
bhennion, line 18:
```suggestion
constexpr auto CUSTOM_KEY = "cust: ";
```
This is more explicit

src/core/model/StrokeStyle.cpp
```diff
@@ -1,91 +1,82 @@
...
-    if (dash.empty()) {
-        return LineStyle();
     }
 
-    auto* dashesArr = new double[dash.size()];
-    for (int i = 0; i < static_cast<int>(dash.size()); i++) { dashesArr[i] = dash[i]; }
+    // Else generate custom dashes string
+    std::ostringstream custom;
+    custom << std::setprecision(2) << std::fixed;
+    custom << CUST_KEY;
```
bhennion, line 38:
```suggestion
    custom << CUSTOM_KEY;
```

src/core/model/StrokeStyle.cpp
```diff
@@ -1,91 +1,82 @@
...
+    std::vector<double> dashes;
 
-    for (int i = 0; i < count; i++) {
-        custom += " ";
-        char* str = g_strdup_printf("%0.2lf", dashes[i]);
-        custom += str;
-        g_free(str);
+    dashStream.seekg(strlen(CUST_KEY));
+    for (double value; dashStream >> value;)
+        dashes.push_back(value);
```
bhennion, line 65:
```suggestion
    for (double value; dashStream >> value;) {
        dashes.push_back(value);
    }
```
See Coding conventions

src/core/view/StrokeViewHelper.cpp
```diff
@@ -49,9 +45,9 @@ double xoj::view::StrokeViewHelper::drawWithPressure(cairo_t* cr, const std::vec
         cairo_stroke(cr);
     };
 
-    if (dashes) {
+    if (!dashes.empty()) {
         for (const auto& [p, q]: PairView(pts)) {
-            cairo_set_dash(cr, dashes, dashCount, dashOffset);
+            cairo_set_dash(cr, dashes.data(), static_cast<int>(dashes.size()), dashOffset);
```
bhennion, line 50: We could have a helper function, either LineStyle::set_cairo_dashes or Util::cairo_set_dash_from_vector. This would avoid the code duplication

bhennion (commented):
Overall, this looks like verygood refactoring you did there!

I'd keep it like that and let you do your deeper refactor in a follow up PR.

I made a couple of comments. I'd like for someone else to go over the code before merging though, in case I (we) missed something.

rolandlo (commented):
The code looks good. Nice refactoring!

One thing that could be improved: We want to document functions (and class members) in the header files, not in the source files. For the `LineStyle` class and the `StrokeViewHelper` class there is a mix between documenting in header and source file. 

Also I have noticed that we don't use `HexObjectEncoding` anywhere, not even in the tests. So I think we should get rid of it (and also the include in `ObjectIOStreamTest.cpp`)

jhilmer:
In the other PR build on top of this one - I have remove `HexObjectEncoding` - I will take a look on the documentation in the files.

bhennion (approved):
LGTM!

PR #3700, opened 2022-01-06
Feature autoexport

rolandlo:
This is a follow up on #3566 (export via plugin API) that introduces auto-export of changed pages.

PR #4703, opened 2023-03-08
WIP: Feature signal broadcasting

rolandlo:
This PR integrates a signaling system into the plugin system. It also adds the capability of calling Lua functions in regular time intervals in a non-blocking way (which is not possible in pure Lua).

Plugins can connect to signals via
```Lua
  app.connectToSignal(signalName, callbackName);
```
and add timeouts via
```Lua
  timeoutId = app.addTimeout(seconds, callback);
```

In contrast to signals from glib, the signals implemented here only allow to pass an integer to the callback function, e.g. the page number via the "pageChanged" signal. This is enough for many purposes and keeps the implementation simple (otherwise we would have to translate C++ data into Lua tables). The callback function should request other needed data by itself.

Timeouts could also be added in Lua using e.g. the `lgi` module, but that requires to have that module installed, which seems apparently nobody has managed to do on Windows. Since this is an important feature, it looks wise to add it to the plugin API like in this PR.

The changes to the `main.lua` file from the Export plugin serve for demonstration purpose. It basically works, but there are still some issues.

This PR improves upon PR #3700, which is much narrower in scope.
It's not hard to add more signals. Currently the signals are `pageChanged`, `pageSizeChanged`, `documentChanged`, `pageInserted`, `pageDeleted`, `undoRedoPageChanged` and `quit`. 

Any feedback on the approach taken here would be appreciated.


loopspace:
This looks like a great addition; the timer facility in particular would be useful for things like automating exporting/saving.  What would also be useful is some way to know if the user is interacting with xournal++ or not.  For example, exporting to a PDF takes resources and so it is best to do it when the user isn't actually trying to do something - like draw something - at that time.  I'm not sure if that would be best implemented by a _signal_, such as "user has been idle for 3 seconds", or just a flag that is set after 3 seconds of idling time (I'm picking 3 as an arbitrary number here!).

rolandlo:
@loopspace The "undoRedoPageChanged" signal is emitted whenever an element (stroke, text, image, TeX-Image) is added to a layer (after completion of that element). That already covers lots of relevant interactions. So doing that via signals may work fine. It depends a bit on how long the export/saving takes.

loopspace:
Thanks for the quick reply!

The `undoRedoPageChanged` is certainly part of the process, but it would also be necessary to detect when the user starts an interaction.  So it might not be a signal but rather a boolean that can be tested to see if the user is currently interacting (perhaps measured by some pointer being "down").  Then an `undoRedoPageChanged` signal can trigger a short timer, after which a plugin checks the `not interacting` boolean and if the user is still not interacting it triggers a PDF export.

(Many, many years ago I implemented something like this in the original `xournal`; not sure if I could figure out the corresponding implementation in `xournal++` but I'd be willing to take a look.)

AnonymerNiklasistanonym:
I just tried this PR (only the signals) on top of the current master in a fork and it worked without problems when using it to check if the document was changed (to e.g. get the document file path).
Are there any reasons this feature is currently not being considered to be ready (just asking since this draft is quite old)?

Would it be possible to maybe split this PR to only add the signals API since this could make plugins more useful and make it easy to add new signals in the future via the provided `PluginNotifier` which is just reusing existing listeners?

PR #4715, opened 2023-03-12
Fix #4675 No error on save if we don't have write permission

### LINKED ISSUE #4675: Xournalpp - no error message when saving file with no write permission
**Affects versions :**
 - OS: Ubuntu 22.04
 - Which version of libgtk do you use: 3.24.36
 - Version of Xournal++: 1.1.3

**Describe the bug and Steps To Reproduce**

I have two OS installed on my system: Ubuntu 22.04 and Windows 11. Sometimes I takes note from Windows 11 and I save them in the main NTFS partition.

Then, I might need to modify the files when I'm running Ubuntu. With Xournal++ I open the files from the NTFS partition and edit them. When I click the save button, no error message is shown, so I'm confident that the edits have been saved. Then I reopen the file just to discover that my edits were not saved. :|

So I opened the folder in the NTFS partition with my file manager in Ubuntu: turns out that I don't have write permission.

Then, I did another test: from Ubuntu, open Xournal++ and creates a new note. Save it under the same folder in the NTFS partition. This time I get an error message:

Screenshot_20230224_112552

I click OK and after a few seconds another error messages pops up:

Screenshot_20230224_112621

The last error message pops up rather frequently.

**Expected behavior**

Once I open an existing file, it would be really nice if Xournal++ would notify me right away that I don't have write permission. Knowing that, I wouldn't spend time editing the file...

Also, I found the error message rather cryptic: in particular, the _"Error operning file"_ part.... I'm trying to save the file. Maybe a _"Don't have write permission"_ would be better...

jhilmer:
During save we try to make a backup of the current file. This fails if there is no write permission to the file, and before this PR we returned without notifiy the user.

### DIFF
--- a/src/core/control/jobs/SaveJob.cpp
+++ b/src/core/control/jobs/SaveJob.cpp
@@ -114,6 +114,10 @@ auto SaveJob::save() -> bool {
             Util::safeRenameFile(target, fs::path{target} += "~");
         } catch (const fs::filesystem_error& fe) {
             g_warning("Could not create backup! Failed with %s", fe.what());
+            this->lastError = FS(_F("Save file error, can't backup: {1}") % std::string(fe.what()));
+            if (!control->getWindow()) {
+                g_error("%s", this->lastError.c_str());
+            }
             return false;
         }
     }


rolandlo (approved):
Thanks for the important fix! LGTM

PR #4716, opened 2023-03-12
Use RAII for resource allocation in MainWindow

jhilmer:
Now all objects allocated with `new` has been changed to use `std::unique_ptr` to ensure their resources are deallocated with MainWindow goes out of scope.

### DIFF
--- a/src/core/gui/MainWindow.h
+++ b/src/core/gui/MainWindow.h
@@ -181,7 +181,7 @@ private:
 private:
     Control* control;
 
-    XournalView* xournal = nullptr;
+    std::unique_ptr<XournalView> xournal;
     GtkWidget* winXournal = nullptr;
     ScrollHandling* scrollHandling = nullptr;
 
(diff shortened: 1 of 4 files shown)


src/core/gui/MainWindow.h
```diff
@@ -88,7 +86,8 @@ class MainWindow: public GladeGui, public LayerCtrlListener {
 
     Control* getControl() const;
 
-    PdfFloatingToolbox* getPdfToolbox() const;
+    const std::unique_ptr<PdfFloatingToolbox>& getPdfToolbox() const;
+    const std::unique_ptr<FloatingToolbox>& getFloatingToolbox() const;
```
Technius, line 90: I don't think there's any advantage to returning a const unique pointer over a raw pointer.

jhilmer: Yes sir :-)  I have dropped the last two commits.

Technius:

PR #4717, opened 2023-03-12
Clean up SettingDialog warnings and use of RAII

jhilmer:


Technius (approved):
Thanks, LGTM. Merging in 24 hours unless there are objections.

PR #4725, opened 2023-03-13
Deselect on undo/redo. Fix #4609

### LINKED ISSUE #4609: Clicking "Undo" does not deselect a selection
**Affects versions :**
 - OS: Linux
 - (Linux only) Desktop environment: Gnome 43.2, Wayland
 - Which version of libgtk do you use: 3.24.36
 - Version of Xournal++: 1.1.3 (2dbd16d2)

**Describe the bug**
I'm not entirely sure if this is intended behaviour, but if you make a selection and then click on undo, the previous action will be undone, but the selection will remain.

**To Reproduce**
A method to reproduce what I mean is as follows:
1. Draw 2 lines
2. Delete one of the lines
3. Select the remaining line
4. While the line is selected, click undo and observe that the previous line reappears while the selection remains.

**Expected behavior**
My initial reaction was that I thought that clicking undo would remove the selection, or perhpas bring back the previous stroke, AND remove the selection.

**Screenshots of Problem**
example

jhilmer:


bhennion (approved):
LGTM!

bhennion:
Fixes #4609

PR #4728, opened 2023-03-13
Ensure we can compile without plugins enabled. Fix #4726

### LINKED ISSUE #4726: Can't be compiled without enabling plugins
Master can't be compiled without enabling plugin. Guess it's due to https://github.com/xournalpp/xournalpp/commit/1dd2364d6a36446d841a51e89cabd913be5b2a15

Where 
```
PluginToolButton::PluginToolButton(ActionHandler* handler, ToolbarButtonEntry* t):
        ToolButton(handler, std::move(t->toolbarId), ACTION_NONE, std::move(t->iconName), std::move(t->description)),
        t(t) {}
```

`ToolbarButtonEntry` is only defined if plugins is enabled.

jhilmer:


rolandlo (approved):
LGTM. Thanks for catching this problem!

There is another regression with respect to compiling without plugins enabled, this time from #4448. The Plugins menu (with a "Plugin Manager" entry) shows up even if plugins are disabled. 
grafik

Previously we had code
```c++    
// If plugins are disabled - disable menu also
    GtkWidget* menuitemPlugin = control->getWindow()->get("menuitemPlugin");
    gtk_widget_hide(menuitemPlugin);
    return {};
```
that was hiding the menu. If you want to fix that as well in this PR (in another commit) please go ahead. Otherwise this can also be deferred to another PR.

rolandlo:
So then let's leave the hiding of the plugin manager menu to another PR. 

bhennion:
> So then let's leave the hiding of the plugin manager menu to another PR.

I can take a look at that, if noone else already is.

rolandlo:
Thanks again @jhilmer for the fix and catching the issue. I just merged the PR.

PR #4734, opened 2023-03-15
Fix dark background color of tex dialog

rolandlo:
Fixes a regression introduced in 727730512cd2ad85d93184816166f25201095da8 by setting the alpha channel of `Colors::black` (used as a background color for the TeX dialog when the text color is light) to `0xff`. This was incompatible with `rgb_to_hex_string`, that turned it into `#ff0000` instead of `#000000`. 

Bad: 
grafik

Good:
grafik


bhennion (approved):
LGTM!

rolandlo:
Thanks for reviewing @bhennion 

PR #4737, opened 2023-03-16
Add several clang-format versions in CI check

bhennion:


bhennion:
This allows for clang-format versions 15 (last release) and 16 (last pre-release).

Allowing for version 14 does not work, because the clang-format-14 available on https://apt.llvm.org/focal/dists/llvm-toolchain-focal-14/ does not return an error code if the code is not properly formatted.
Allowing for the clang-format's git version (17) does not seem very sensible to me.

bhennion:
The issue is: we'll need to bump by hand the clang-format versions once in a while. I don't really see a way around that for now.

rolandlo (approved):
LGTM

Updating the `CLANG_FORMAT_VERSIONS` once a while looks like a very moderate annoyance compared to the issues we encountered in the last weeks with respect to the Clang-format check.

PR #4739, opened 2023-03-17
Debug attempt: clang-format test ignores AllowShortBlocksOnASingleLine: Empty option

bhennion:
Testing for a fix...

### DIFF
--- a/azure-pipelines/clang-format-applied.yml
+++ b/azure-pipelines/clang-format-applied.yml
@@ -49,10 +49,12 @@ stages:
         # Exit on git merge-base failure
         [[ -z "$BASE_COMMIT" ]] && { echo "Error: BASE_COMMIT is empty" ; exit 1; }
         git fetch origin $BASE_COMMIT
-        
+
         for VERSION in $(CLANG_FORMAT_VERSIONS); do
-          echo "Testing clang-format-$VERSION"
-          git clang-format-$VERSION --diff $BASE_COMMIT -- && { echo "Success: changes are compatible with clang-format-$VERSION"; exit 0; }
+          echo "Testing $(clang-format-$VERSION --version)"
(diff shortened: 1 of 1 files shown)


bhennion:
Alright, I don't understand. It looks like the packages hosted at apt.llvm.org simply ignore the `AllowShortBlocksOnASingleLine: Empty` option.
This does not match what I have on Archlinux, with the same clang-format version number (15.0.7), where everything works as expected.

I'm not quite sure how to fix the CI check now...

bhennion:
I made the following test. I added stupid loops to test the formatting:
```
while (false) {}
while (false) {
}
    
while (false) { continue; }
while (false) {
    continue;
}
```
The result with `AllowShortBlocksOnASingleLine: Empty` should be 
```
while (false) {}
while (false) {}
    
while (false) {
    continue;
}
while (false) {
    continue;
}
```
but the CI asks for
```
while (false) {}
while (false) {}
    
while (false) { continue; }
while (false) { continue; }
```
I checked that the config file was indeed read: clang-format issues an error if Empty is replaced by ThisIsNotAValidOption.

Any idea?

bhennion:
> What's the output of clang-format --version git clang-format-## will most likely choose that instead of clang-format-xx. With your latest commit to the CI you removed the nightly version of clang-format which overrides any old version preinstalled on that system. So probably it will use clang-format-10

Indeed! There seems to be a bug in git-clang-format-##: the default clang-format binary is not `clang-format-##` but `clang-format`:
```
  p.add_argument('--binary',
                 default=config.get('clangformat.binary', 'clang-format'),
                 help='path to clang-format'),
```
Calling `git clang-format --binary=clang-format-$VERSION --diff $BASE_COMMIT --` does the trick!

Thanks a lot @Febbe!

bhennion:
I reported the issue upstream: https://github.com/llvm/llvm-project/issues/61494

PR #4743, opened 2023-03-19
Fix include in util/TinyVector

bhennion:


rolandlo:
Shouldn't `util/SmallVector.h` also include the `memory` header in that case? I know it includes `util/TinyVector.h`, but we want IWYU.

bhennion:
Probably yes. The thing is, when I applied IWYU to the code tree, it ignored every header that is not associated to a .cpp file, so all the templates in util/include/util/ were not IWYU'd. 

There is a way to get those files checked by IWYU. I'll try to do that later today.

PR #4744, opened 2023-03-19
Don't allocate dialogs on heap

jhilmer:
Some dialogs was allocated on heap other on stack. I have now align them to be allocated on stack. The benefit is that we then don't need to explict deallocate them.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1530,16 +1530,14 @@ void Control::insertPage(const PageRef& page, size_t position, bool shouldScroll
 }
 
 void Control::gotoPage() {
-    auto* dlg = new GotoDialog(this->gladeSearchPath, int(this->doc->getPageCount()));
+    auto dlg = GotoDialog(this->gladeSearchPath, int(this->doc->getPageCount()));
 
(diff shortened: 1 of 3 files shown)


bhennion (approved):
LGTM! Merging in 24 hours.

Febbe:
Actually, this is ok for now, but the whole Dialog::show implementation can't work like that. It utilizes gtk_dialog_run, which is not available in gtk4. In gtk4, dialogs are non-blocking and the function returning from it must return into the event loop.

Using C++ classes with local storage duration won't work then.
@bhennion, the question is, did you start to port dialogs for gtk4? Also, when you want to reuse my code of the gtk4 branch, those changes might cause conflicts.

bhennion:
> @bhennion, the question is, did you start to port dialogs for gtk4? Also, when you want to reuse my code of the gtk4 branch, those changes might cause conflicts.

No I didn't.
About the gtk4 branch, I think rebasing it is close to impossible (there are some version markers in there already). This changes won't make a dent.
My strategy for porting is to look at your work for inspiration, copy what I can by hand, redo the rest. Not optimal.

PR #4745, opened 2023-03-19
Page type radio button incorrect when changing to page type there fails.

jhilmer:
The radio button showing the current selected page type was
incorrect. E.g. create a new document, select pdf background
which fails due to no pdf file. The radio button still shows
pdf background even the page type was not changed.


bhennion:
The whole PageType menus and PageTypeHandler is refactored in #4508, so there'll be a lot of conflicts here.
We may still wanna adopt this more conservative approach until #4508 is merged (it could take time, I guess).

bhennion:
Closing, since #4508 have been merged and the issue has been fixed.

PR #4746, opened 2023-03-19
Hide plugin menu when compiling with -DENABLE_PLUGINS=off

bhennion:
also fixes the cmake flags so that configuring with -DENABLE_PLUGINS=off actually compiles without plugins (whether Lua is present or not)


rolandlo (approved):
LGTM!

PR #4749, opened 2023-03-19
Apply include-what-you-use to headers in util/ without associated .cpp file

bhennion:
Merging in 24 hours unless the pipeline fails or someone objects.


Febbe:
@bhennion I think you can merge such changes immediately after the pipeline succeeds.

PR #4756, opened 2023-03-23
Fixed some compilation issues with MSVC and clang

Febbe:


src/CMakeLists.txt
```diff
@@ -9,11 +9,12 @@ cmake_policy(VERSION 3.13)
...
 add_library(defaults INTERFACE)
-target_compile_options(defaults INTERFACE -Wall -Wreturn-type -Wuninitialized -Wunused-value -Wunused-variable -Wconversion)
 target_include_directories(defaults INTERFACE ${PROJECT_BINARY_DIR}/src)
 add_library(xoj::defaults ALIAS defaults)
-if (MSVC)
-  target_compile_definitions(defaults INTERFACE PUBLIC NOMINMAX WIN32_LEAN_AND_MEAN)
+if (MSVC) # clang-cl.exe, cl.exe
+target_compile_definitions(defaults INTERFACE NOMINMAX WIN32_LEAN_AND_MEAN)
+else(MSVC) # gnu clang with gnu driver, 
+target_compile_options(defaults INTERFACE -Wall -Wreturn-type -Wuninitialized -Wunused-value -Wunused-variable -Wconversion)
```
bhennion, line 17: Nit for readability
```suggestion
  target_compile_definitions(defaults INTERFACE NOMINMAX WIN32_LEAN_AND_MEAN)
else(MSVC) # gnu clang with gnu driver, 
  target_compile_options(defaults INTERFACE -Wall -Wreturn-type -Wuninitialized -Wunused-value -Wunused-variable -Wconversion)
```

src/util/XojMsgBox.cpp
```diff
@@ -9,7 +9,9 @@
 
 #ifdef _WIN32
 // Needed for help dialog workaround on Windows; see XojMsgBox::showHelp
-#include <shlwapi.h>
+#include <windows.h>
+//
```
bhennion, line 13:
```suggestion
```

bhennion (approved):
LGTM, beside two tiny nitpicks.
I can't test if this indeed solves compilation issues with MSVC, but I trust you it does :)

src/util/XojMsgBox.cpp
Febbe: windows.h must be included first.

bhennion: Then maybe add a comment saying just that instead of this empty comment line

Febbe:
> LGTM, beside two tiny nitpicks. I can't test if this indeed solves compilation issues with MSVC, but I trust you it does :)

Yes, it now compiles, but it does not work well. At least it does not crash :D.
But compiling with MSYS2/UCRT64 works well, we should swap from MINGW64 to that.

MSYS2/CLANG64 still has one issue with a constexpr static to a function ptr.

Febbe:
> @Febbe What do you mean by  "it does not work well"?
> 

Using the vcpkg packages, it just crashes. Using the msys/ucrt64 or clang64 packages scrolling and zooming does not work.
It may be, because vcpkg also just copied some of the msys packages. Instead of compiling them. Seems like, we must await gtk4, to be truly platform independent.

CMakeLists.txt
```diff
@@ -137,15 +137,18 @@ find_package(ZLIB REQUIRED)
...
+find_package(Lua) # Lua 5.4 is only supported with cmake >=3.18
+message(STATUS "Found Lua: ${Lua_FOUND}")
 if (Lua_FOUND AND ENABLE_PLUGINS)
     # currently not fully supported by cmake
-    message("Enable Xournal++ Plugins")
+    message(STATUS "Enable Xournal++ Plugins")
     add_library(lua INTERFACE)
     target_link_libraries(lua INTERFACE ${LUA_LIBRARIES})
     target_include_directories(lua INTERFACE ${LUA_INCLUDE_DIR})
+    target_compile_definitions(lua INTERFACE ENABLE_PLUGINS)
```
bhennion, line 148: @Febbe Do you remember why you added this line here? It causes some
```
./build/src/config-features.h:20: warning: "ENABLE_PLUGINS" redefined
   20 | #define ENABLE_PLUGINS
```
all over the place when compiling with GCC.

We can always remove the line from config-features.h (it works), but it sounds dirtier than to remove the above line.

Febbe: I've re-added it after someone removed it, because I got compilation errors.
It's probably some sort of religion/opinion: My opinion is, that those variables should be configured via CMake, so that they are available in all relevant source and header files, this reduces bugs, when a file forgets to include the config file.
It's probably the more modern way.

But of course it has drawbacks: It can happen, that this way might affect included third party headers, like the Windows API, when a name clash occurs. Including only the config file, where it is required, can circumvent this.
But that's a design flaw anyway, header files should always work the same regardless of the order in which they are included.
For example, the llvm include order enforces this, by including local headers first and system headers last. We use the google include order, which relaxes exactly this to retain some compatibility to old and badly designed C libraries. It is also more readable.

You can remove one of them, which one might not matter, your solution looks cleaner, because it's more consistent. But I think it would be cleaner, to configure all options in CMake. I also think `configure_file` has a big problem:

It only reconfigures, when the source.in file changed, not any of the variables:

> If the input file is modified the build system will re-run CMake to re-configure the file and generate the build system again. The generated file is modified and its timestamp updated on subsequent cmake runs only if its content is changed.
 - https://cmake.org/cmake/help/latest/command/configure_file.html

PR #4760, opened 2023-03-24
Remove Object*Stream write_data read_data  with raw ptr

jhilmer:


Febbe:
Do not use `std::string` for binary data, use `std::vector<std::byte>` for that. `std::string` is more than a managed `char` range, it represents actually a string and asserts, that all its content is somehow representable with the system encoding.

src/util/include/util/serializing/ObjectInputStream.h
```diff
@@ -52,4 +54,43 @@ class ObjectInputStream {
     size_t len = 0;
 };
 
-extern template void ObjectInputStream::readData(std::vector<double>& data);
+// This function requires that T is read from its binary representation to work (e.g. integer type)
+template <typename T>
+T readTypeFromSStream(std::istringstream& istream) {
+    if (istream.str().size() < sizeof(T)) {
```
Febbe, line 60: You are creating copies of the buffers, each time this point is reached, since this is called in a loop, this is very bad.

src/util/include/util/serializing/ObjectInputStream.h
```diff
@@ -52,4 +54,43 @@ class ObjectInputStream {
...
+        std::ostringstream oss;
+        oss << "End reached: trying to read " << sizeof(T) << " bytes while only " << istream.str().size()
```
Febbe, line 62: Same here, copy of string

src/util/include/util/serializing/ObjectInputStream.h
Febbe, line 60: It is hard to get the size of a stream, therefore it should not be done at all, let it crash or check the stream flags.

src/util/include/util/serializing/ObjectOutputStream.h
```diff
@@ -35,18 +33,24 @@ class ObjectOutputStream {
...
-    void writeData(const void* data, int len, int width);
-
     template <typename T>
     void writeData(const std::vector<T>& data);
 
     /// Writes the raw image data to the output stream.
     void writeImage(const std::string_view& imgData);
 
-    GString* getStr();
+    std::string getStr();
```
Febbe, line 42:
```suggestion
    std::string const& str();
```
If this is binary data:

 ```suggestion
    std::span<std::byte> data();
```
or, since we don't have c++20 yet:
```suggestion
    std::vector<std::byte> data();
```

src/util/serializing/BinObjectEncoding.cpp
```diff
@@ -1,11 +1,11 @@
...
-#include <glib.h>  // for g_string_append_len
+#include <string>
 
 BinObjectEncoding::BinObjectEncoding() = default;
 
 BinObjectEncoding::~BinObjectEncoding() = default;
 
 void BinObjectEncoding::addData(const void* data, int len) {
-    g_string_append_len(this->data, static_cast<const char*>(data), len);
+    this->data.append(std::string((const char*)data, (const char*)data + len));
```
Febbe, line 10: use std::stringview instead of creating/copiing a string, which is destructed immediatelly.

src/util/serializing/ObjectEncoding.cpp
```diff
@@ -1,13 +1,15 @@
...
+void ObjectEncoding::addStr(const char* str) { this->data.append(std::string(str)); }
 
-auto ObjectEncoding::getData() -> GString* {
-    GString* str = this->data;
-    this->data = nullptr;
-    return str;
+auto ObjectEncoding::getData() -> std::string {
+    std::string res = this->data;
+    this->data.resize(0);
+    return res;
```
Febbe, line 14: copy, and then reset is inefficient, move or exchange it.
```suggestion
    return std::move(this->data);
```

Febbe (changes requested):

src/util/serializing/ObjectEncoding.cpp
jhilmer: Done

src/util/serializing/BinObjectEncoding.cpp
jhilmer: `this->data` is now a `std::vector<std::byte>`, so this is done different now

src/util/include/util/serializing/ObjectOutputStream.h
jhilmer: Change to `std::vector<std::byte> const& getData();`

src/util/include/util/serializing/ObjectInputStream.h
jhilmer: I have removed all use of `istream.str()`

src/util/include/util/serializing/ObjectInputStream.h
jhilmer: I have removed all use of `istream.str()`

src/util/include/util/serializing/ObjectInputStream.h
jhilmer: I have removed all use of `istream.str()`

Febbe:
A hint, for a better working routine:  I've seen, that you are using clang-format manually on all edited files. This will format the hole file, no matter if you touched it or not. That has the drawback, that your PR will be cluttered with changes, that do not belong to your PR.

Therefore, we also only check, if the touched lines are formatted.
Instead, you can just call `git clang-format <first commit/revision hash>` example: `git clang-format HEAD~1` which will format the last 1 commits.

Also, I can recommend enabling automatic formatting *on save*, in your IDE, *based on the latest changes*.
With that, your commits are always correct.

src/util/serializing/ObjectInputStream.cpp
```diff
@@ -94,94 +76,46 @@ auto ObjectInputStream::readString() -> std::string {
...
 
-    if (len) {
-        data.resize(len);
-        istream.read((char*)data.data(), len * width);
-    }
+    return output;
 }
 
-auto ObjectInputStream::readImage() -> std::string {
+auto ObjectInputStream::readImage() -> const std::vector<std::byte> {
```
Febbe, line 91:
```suggestion
auto ObjectInputStream::readImage() -> std::vector<std::byte> {
```

Febbe: If the data must not be edited, you should return std::vector<std::byte const>, but that can result in undefined behav, when the original vector is also defined in that way and someone casts away the const qualifier:


```
1. std::vector<std::byte const> vec;
2. return vec;
3. reinterpret_cast<char*> (vec.data()); // undefined behav.
3. reinterpret_cast<char const*> (vec.data()); // valid
```
vs:

```
1. std::vector<std::byte> vec;
2. return vec; // as std::vector<std::byte const> 
3. reinterpret_cast<char*> (vec.data()); // valid
3. reinterpret_cast<char const*> (vec.data()); // valid
```

src/util/serializing/ObjectInputStream.cpp
```diff
@@ -94,94 +76,46 @@ auto ObjectInputStream::readString() -> std::string {
...
-    data.resize(len);
-    istream.read(data.data(), static_cast<int>(len));
 
     return data;
 }
 
 void ObjectInputStream::checkType(char type) {
-    if (istream.str().size() < 2) {
+    char t = 0, underscore = 0;
+    istream >> underscore >> t;
```
Febbe, line 112: **Question:** is the istream binary or formatted? Else, `>>` should not be used.

test/unit_tests/model/ImageTest.cpp
```diff
@@ -23,10 +23,15 @@ TEST(Image, testGetImageApplyOrientation) {
     // Image width is 500px and height 130px - but has exif data saying image should be
     // rotated 90 deg CW to have correct orientation.
     std::ifstream imageFile{GET_TESTFILE("images/r90.jpg"), std::ios::binary};
-    auto imageData = std::string(std::istreambuf_iterator<char>(imageFile), {});
+    imageFile.seekg(0, std::ios::end);
+    const std::streamsize size = imageFile.tellg();
```
Febbe, line 27: I think, tellg is not valid to determine the size of the stream. We should use the filesystem for that.

Febbe (commented):
@xournalpp/core I think this looks mostly good, but another one should take a look also.

test/unit_tests/model/ImageTest.cpp
jhilmer: Fixed. Now use `std::filesystem::file_size`

src/util/serializing/ObjectInputStream.cpp
jhilmer: The stream is binary. I have changed it to use `istream::read`

src/core/control/xml/XmlTexNode.h
```diff
@@ -29,5 +29,5 @@ class XmlTexNode: public XmlNode {
     /**
      * Binary .PNG or .PDF
      */
-    std::string binaryData;
+    const std::vector<std::byte> binaryData;
```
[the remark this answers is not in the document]
jhilmer: Hmm. I think I end up here to get to work. The `XmlTexNode` should brorrow the image from the Element. I in doubt about how I should handle this. But now the `SaveHandler` create a `XmlTexNode` which is constructed with a reference to the binary data owned by the Element.

src/core/control/xml/XmlTexNode.h
```diff
@@ -11,15 +11,15 @@
...
+#include <vector>  // for vector
 
 #include "XmlNode.h"  // for XmlNode
 
 class OutputStream;
 
 class XmlTexNode: public XmlNode {
 public:
-    XmlTexNode(const char* tag, std::string&& binaryData);
+    XmlTexNode(const char* tag, const std::vector<std::byte>&& binaryData);
```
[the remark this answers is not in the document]
jhilmer: See above

src/util/include/util/serializing/ObjectEncoding.h
```diff
@@ -11,20 +11,20 @@
...
     virtual ~ObjectEncoding();
 
 public:
-    void addStr(const char* str) const;
-    virtual void addData(const void* data, int len) = 0;
+    void addStr(const std::string_view& str);
+    virtual void addData(const void* data, size_t len) = 0;
 
-    GString* getData();
+    std::vector<std::byte> const& getData();
```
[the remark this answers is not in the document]
jhilmer: I have now implemented `stealData` - getData was only used once in the main program. The rest of the places was in unit tests. I have removed `getData` now.

jhilmer:
I think - I have address all issue in this PR

jhilmer:
Rebase on master

src/util/include/util/serializing/ObjectInputStream.h
```diff
@@ -39,17 +41,51 @@ class ObjectInputStream {
...
 
 private:
-    void checkType(char type);
+    void checkType(const std::byte type);
 
-    static std::string getType(char type);
+    static std::string getType(const std::byte type);
 
 private:
     std::istringstream istream;
```
bhennion, line 52: Wouldn't a std::basic_istream<std::byte> make more sense then?

src/util/include/util/serializing/ObjectEncoding.h
```diff
@@ -11,20 +11,20 @@
...
-    virtual void addData(const void* data, int len) = 0;
+    void addStr(std::string_view str);
+    virtual void addData(const void* data, size_t len) = 0;
 
-    GString* getData();
+    std::vector<std::byte> stealData();
 
 public:
-    GString* data;
+    std::vector<std::byte> data;
```
bhennion, line 29: Shouldn't data be protected here? (Otherwise I'm not sure what the accessor stealData() is for)

src/util/include/util/serializing/ObjectInputStream.h
Febbe: I also wondered if that would be a good alternative to char streams, but after reading https://open-std.org/JTC1/SC22/WG21/docs/papers/2020/p2146r2.pdf , I think this won't work, because std::byte is not a char type and therefore does not implement char_traits. I think, we should stay with the old streams, until a good alternative is accepted.

Issue #2722, opened 2021-01-28
GTK4 preparation


Technius:
GTK 4 was released last month and will probably become widely available over the next year or two. There's no rush to port to GTK 4 given the huge number of other important issues in our issue tracker, but it would be nice to [at least replace the deprecated GTK 3 parts that we can right now](https://developer.gnome.org/gtk4/4.0/gtk-migrating-3-to-4.html).

At least one low-hanging fruit I know of is the `gtk_pointer_grab` deprecation warning (or something along those lines) that's been around for over a year.

MikuChan03:
Big facts.

bhennion:
Do we have a roadmap for porting to GTK4?
#3306 took care of some preparations. What should the next steps be?

Technius:
We should probably start by adding a CMake option that lets us choose GTK4 and then see what compiler errors we get. I imagine that there will probably be a lot of places where we will need to make changes.

On another note, Glade does not seem to support GTK4. We will need to switch to another API like [`GtkBuilder`](https://docs.gtk.org/gtk4/class.Builder.html).

Technius:
Oh, it appears that we already use `GtkBuilder` internally; Glade is only used to edit the `.glade` files. So as far as "switching from Glade" is concerned, I think the only required change is to port the `.glade` files over to the GTK4 builder format.

bhennion:
I think a big source of troubles will be the input handling: https://docs.gtk.org/gtk4/migrating-3to4.html#stop-using-gtkwidget-event-signals. Most of the GTK4 replacements for deprecated GTK3 signals have been backported to GTK3, so we can start making changes without actually making the switch just yet.

bhennion:
@febbe You had worked a lot on this on a dedicated `gtk4` branch. It's not quit clear to me what part of this branch was merged with #3306 and what was not.
I could try and help on this front if you want

Febbe:
The complete gtk4 branch is not part of the current master. It must be rebased. And I would cherry pick small fixes which are compatible to gtk3. One of the problematic changes is the whole dialog systems. Dialogs do not have a run function anymore and are non blocking now.

Therefore it might be useful to replace those by cpp20 coroutines. This would reduce the effort to rewrite all Dialogs.

The next big problem is, that gtk4 dropped menu's so we have to switch from gtkmenu to gmenu which is only declarative. Every menu in xpp must be reworked.

[...]

PR #4763, opened 2023-03-25
Make some dialogs compatible with GTK4

bhennion:
This is an attempt to adapt our handling of dialogs, and make it compatible with Gtk4 (different from @Febbe's in the gtk4 branch).

As mentionned in #2722, the main problem is that dialogs in Gtk4 are not blocking. The philosophy is: noone should ever block the main loop, if you're waiting for something, make your response a callback.

I propose here a wrapper around `Dialog`s: it behaves like a smart pointer, but transfers the `Dialog`'s destruction to the associated window's closing signal once the dialog is shown.

I also change GtkDialog's into GtkWindow's, as GtkDialog's are deprecated in Gtk4.

For now, I implemented the changes to AboutDialog, FillOpacityDialog, ExportDialog, GotoDialog and FormatDialog.
Except for the first one, the commits are independent and can be reviewed one by one.

Those dialogs should compile fine with GTK4, but some properties set in the UI files will need to be replaced when the switch actually happens (e.g. a widget-based `<property name="has-default">true</property>` will need to be replaced by a window-wide `<property name="default-widget">widgetID</property>`).

I'll work on more dialogs in a follow up PR.

**Edit:** updated PR description


Febbe:
I must look into it deeper tomorrow. But a general question, why do we still need inheritance here. I would love to get rid of the GladeGui class, and the dialog class should also be overkill. Instead I would prefer a general dialog generator, which only depends on the handler and optionally the glade file.

Also, how do you want to handle functions, which return a value from the dialog, after blocking on it? We must basically rewrite all those functions to some sort of state machine, keeping track of the dialogs state.

With c++20 we could use coroutines instead. They wouldn't touch our control flow.

bhennion:
> I must look into it deeper tomorrow. But a general question, why do we still need inheritance here. I would love to get rid of the GladeGui class, and the dialog class should also be overkill. Instead I would prefer a general dialog generator, which only depends on the handler and optionally the glade file.

Yes, GladeGui seems redundant. The Dialog base class is used for its destructor:
```
Dialog::~Dialog() {
    std::cout << "deleting Dialog" << std::endl;
    gtk_window_destroy(GTK_WINDOW(this->window));
}
```
The actual window only gets destroyed if the Dialog gets destroyed, so any dev could spot if there were a data leak (there'd still be a window...).

> Also, how do you want to handle functions, which return a value from the dialog, after blocking on it? We must basically rewrite all those functions to some sort of state machine, keeping track of the dialogs state.

The callback should rely on two different set of variables:

1. the ones we use to get from the dialog itself (e.g. the state of a spin button and whatnot)
2. the ones we get from the original caller's local variables.

Dummy example
```
void caller() {
    int status = 34;
    SomeDialog dlg(...);
    dlg.show(...);
    double value = dlg.getValue();

    whatever_should(status);
    become_part_of_the_callback(value);
}
```
So here, `value` is of type 1. and `status` is of type 2.

Type 1. is no issue, we already have the getters for those, and `dlg` itself can easily be passed on to the callback (it outlives the callback).
Type 2. is more complicated. For now, I only implemented the case where the Type 2. variables were actually global app variables (e.g. a ref to Control or so). Then, no lifetime issues. In general, those Type 2. variables should be passed on to the callback in a state structure: using a coroutine's state for that seems a good solution.

> With c++20 we could use coroutines instead. They wouldn't touch our control flow.

I am not sure this changes the structure of Dialog and descendents. The point with using coroutines here is, I guess, that we can give a coroutine's resume() as a callback function for our dialogs.
Or did you imagine something else?

bhennion:
As a side question: are we ready to use coroutines in terms of compiler support?
GCC and MSVC seem [ok](https://en.cppreference.com/w/cpp/compiler_support/20), but Clang seems to have a problem on Windows (see [here](https://clang.llvm.org/cxx_status.html#cxx20))

Febbe:
> As a side question: are we ready to use coroutines in terms of compiler support?
> GCC and MSVC seem [ok](https://en.cppreference.com/w/cpp/compiler_support/20), but Clang seems to have a problem on Windows (see [here](https://clang.llvm.org/cxx_status.html#cxx20))

Regarding windows, I can tell that it works, we are using clang-cl as compiler and we use `asio::awaitable`. Seems like we did not hit those Abi missmatches yet. Also the question is, if our target systems support those new compilers.
Microsoft Win 11 is not a problem and Mac OS should also support it. But those many Linux derivates are problematic. Debian Buster for example does not support it natively. But we could install a own libc++ or libstdc++.

Febbe:
> > I must look into it deeper tomorrow. But a general question, why do we still need inheritance here. I would love to get rid of the GladeGui class, and the dialog class should also be overkill. Instead I would prefer a general dialog generator, which only depends on the handler and optionally the glade file.
> 
> Yes, GladeGui seems redundant. The Dialog base class is used for its destructor:
> ```
> Dialog::~Dialog() {
>     std::cout << "deleting Dialog" << std::endl;
>     gtk_window_destroy(GTK_WINDOW(this->window));
> }
> ```
> The actual window only gets destroyed if the Dialog gets destroyed, so any dev could spot if there were a data leak (there'd still be a window...).

Ok, but those handlers could also know, that it has to handle the superclass, therefore the base lass could be completely protected.

> 
> > Also, how do you want to handle functions, which return a value from the dialog, after blocking on it? We must basically rewrite all those functions to some sort of state machine, keeping track of the dialogs state.
> 
> The callback should rely on two different set of variables:
> 
> 1. the ones we use to get from the dialog itself (e.g. the state of a spin button and whatnot)
> 2. the ones we get from the original caller's local variables.
> 
> Dummy example
> ```
> void caller() {
>     int status = 34;
>     SomeDialog dlg(...);
>     dlg.show(...);
>     double value = dlg.getValue();
> 
>     whatever_should(status);
>     become_part_of_the_callback(value);
> }
> ```

Value is not initialized yet. The dlg may not even spawned at that point. And how do we want to handle the case, where caller wants to return a value determined by the Dialog?

> So here, `value` is of type 1. and `status` is of type 2.
> 
> Type 1. is no issue, we already have the getters for those, and `dlg` itself can easily be passed on to the callback (it outlives the callback).
> Type 2. is more complicated. For now, I only implemented the case where the Type 2. variables were actually global app variables (e.g. a ref to Control or so). Then, no lifetime issues. In general, those Type 2. variables should be passed on to the callback in a state structure: using a coroutine's state for that seems a good solution.

Without coroutines, those values must be moved into the dialogs own state or must live in the surrounding newly written state Maschine. Which itself should be managed via the lifetime of the dialog(to prevent a global state for the state machine@ ).


> > With c++20 we could use coroutines instead. They wouldn't touch our control flow.
> 
> I am not sure this changes the structure of Dialog and descendents. The point with using coroutines here is, I guess, that we can give a coroutine's resume() as a callback function for our dialogs.
> Or did you imagine something else?

The code will look like this for a caller with an value return:

```
xoj::task<double> caller() {
    int status = 34;
    SomeDialog dlg(...);
    double value = co_await dlg.async_show(...);
    whatever_should(status);
    co_return value;
}
```
xoj::task must implement the coroutine interface of cause.

Febbe:
> > > Dummy example
> > > ```
> > > void caller() {
> > >     int status = 34;
> > >     SomeDialog dlg(...);
> > >     dlg.show(...);
> > >     double value = dlg.getValue();
> > > 
> > >     whatever_should(status);
> > >     become_part_of_the_callback(value);
> > > }
> > > ```
> > 
> > 
> > Value is not initialized yet. The dlg may not even spawned at that point. And how do we want to handle the case, where caller wants to return a value determined by the Dialog?
> 
> I wasn't clear here: this is an example as in master: dlg.show() is blocking (so value is initialized). We want to adapt this.

Ah ok, this example is not problematic I think. Because we are here in the C domain and those gtk calls only support pointers, not small types, you must create either a data/handler class, which you pass as pointer, allocated on the heap, or you allocate the callback (a lambda with captures) directly in [dynamic storage](https://stackoverflow.com/questions/37924996/lambda-with-dynamic-storage-duration).

The more problematic cases are all those, where the result is used after the return of the control flow.
Something like that:

```

class A{
void get_foo(){
  FooDialog f{member_foo};
  f.show;
  member_foo = f.getFoo()
}

void do_task_with_double_member_foo(){
  getFoo();
  control.foouse(member_foo)
}
  double member_foo;
};
```

Or simple calls, which return the result of the Dialog
```
Foo getFoo(){
  FooDialog d;
  d.show();
  return d.getFoo();  
}
```

src/core/gui/dialog/Dialog.h
```diff
@@ -0,0 +1,56 @@
...
+    DialogSPtr(Dialog* dialog);
+    ~DialogSPtr();
+
+    void show(GtkWindow* parent);
+
+private:
+    static bool closeDialogCallback(Dialog* self);
+
+    Dialog* dialog = nullptr;
+};
```
Febbe, line 51: With the following, you can get rid of the `virtual void callCallback() const;` or just make all the functions + destructor in `Dialog` protected:

```suggestion
template <class DialogClass>
class DialogSPtr {
public:
    DialogSPtr(DialogClass* dialog);
    ~DialogSPtr();

    void show(GtkWindow* parent);

private:
    static bool closeDialogCallback(DialogClass* self);

    DialogClass* dialog = nullptr;
};
```

src/core/gui/dialog/Dialog.h
```diff
@@ -0,0 +1,56 @@
...
+
+class DialogSPtr {
+public:
+    DialogSPtr(Dialog* dialog);
+    ~DialogSPtr();
+
+    void show(GtkWindow* parent);
+
+private:
+    static bool closeDialogCallback(Dialog* self);
```
Febbe, line 48: Actually, when I think about this, it's the responsibility of the DialogImpl to register callbacks, isn't it?

bhennion:
I think I got to a structure that feels right and fits all cases I've encountered so far.
I have not encountered situations like the ones @Febbe described until now, but I suspect we will have issues in places where we rely on a `GtkFileChooserDialog`.

Febbe:
> I think I got to a structure that feels right and fits all cases I've encountered so far.
> I have not encountered situations like the ones @Febbe described until now, but I suspect we will have issues in places where we rely on a `GtkFileChooserDialog`.

Yes, your code looked also relatively good so far. Cases like the ones `GtkFileChooserDialog` can still be done later I think.

src/core/gui/Builder.h
```diff
@@ -0,0 +1,43 @@
...
+
+class GladeSearchpath;
+
+class Builder final {
+public:
+    Builder(GladeSearchpath* gladeSearchPath, const std::string& uiFile);
+    ~Builder();
+
+    template <class GObjectType>
+    GObjectType* get(const std::string& name) {
```
Febbe, line 30: May it be a good Idea to also return a guarded pointer like `GObjectSPtr<GObjectType>`?

src/core/gui/Builder.cpp
```diff
@@ -0,0 +1,10 @@
+#include "Builder.h"
+
+#include "GladeSearchpath.h"  // for GladeSearchpath
+#include "filesystem.h"       // for path
+
+Builder::Builder(GladeSearchpath* gladeSearchPath, const std::string& uiFile):
+        builder(gtk_builder_new_from_file(gladeSearchPath->findFile("", uiFile).u8string().c_str()), xoj::util::adopt) {
+}
+
+Builder::~Builder() = default;
```
Febbe, line 10: While defaulted destructors and (in some cases) constructors allow a forward declaration of incomplete field types, it also prevents several optimizations and forces the programmer to use the rule of 5.

Therefore, the rule of thumb is, to remove all defaulted special members, unless they must be declared + defined. Try to define them in the header.
```suggestion
```

Febbe:
@rolandlo do you have time to test this in production? I don't have much time because I'm writing my thesis right now. But I can still do code reviews.

rolandlo:
> @rolandlo do you have time to test this in production? I don't have much time because I'm writing my thesis right now. But I can still do code reviews.

Yes, I can test it thoroughly.

lb90:
Hello! Showing dialogs without nested loops is definitely the best way forward. Anyway, as a porting aid, gtk_dialog_run() can be reintroduced as in https://gitlab.com/inkscape/inkscape/-/merge_requests/5373

https://gitlab.com/inkscape/inkscape/-/blob/827e13285f/src/ui/dialog-run.cpp#L10

bhennion:
Alright, I think this is stable enough and up for review.

rolandlo:
Why are there minimize and maximize buttons (which do not work)?
grafik

rolandlo:
There are some rounding errors for the opacity value in the fill opacity dialog. When I close and reopen the dialog, I would expect to get the same value displayed again.

rolandlo (commented):
Other than the (superfluous and non-working) minimize and maximize buttons I didn't find a regression from this PR. The missing accuracy in the fill opacity dialog was already present before (although slightly differently it seems).

On a positive note: My stylus interacts with the dialogs (in particular with the buttons) just fine, which was mostly not the case with the old dialogs.

I have also noticed that the new dialogs can be moved and the main window can be resized while the dialogs are open. That looks advantageous to me, but I don't know if it was intended.

bhennion:
> Why are there minimize and maximize buttons (which do not work)? grafik

Strangly enough, I do not get the same thing: I have a minimize button, which minimizes the entire app (dialog+main window), but not maximize button
Screenshot_20230810_160003

I'm on KDE - X11, and this behaviour is exactly the same as what I have on master.
To be clear, those buttons did not appear for you with the master branch?

bhennion:
Did some more tests:

- KDE - Wayland: minimize button minimizes the dialog only, no maximize button -- same as master
- KDE - X11: minimize button minimizes the entire app, no maximize button -- same as master
- Gnome - Wayland: no minimize button, no maximize button -- same as master
- Gnome - X11: minimize and maximize buttons, with no effects -- same as master

rolandlo:
The relevant settings in the Gnome tweaks tool can also be obtained via
```term
gsettings get org.gnome.desktop.wm.preferences button-layout
```

bhennion:
Rebased onto master. Merging in 48 hours unless an objection is raised.
@rolandlo I don't think there is anything I can do about how the DE decides to implement modal windows.

rolandlo:
> Rebased onto master. Merging in 48 hours unless an objection is raised. @rolandlo I don't think there is anything I can do about how the DE decides to implement modal windows.

Ok, I'm fine with that.

Issue #4732, opened 2023-03-15
Add URL link tool


Zikoel:
**Is your feature request related to a problem? Please describe.**
 Very often a note can refers to some external resources. A clickable `URL link` would be very useful.
 
**Describe the solution you'd like**
Another tool similar to `Text` called `Link` with an UX  that today practically every text software has: A fiel for the URL and another for the displayed text inside the note.

**Describe alternatives you've considered**
Actually I embed the entire link as a text on the note. Very long URL in this case are problematic.

thisissami:
Any sense of whether or not this is being prioritized? This is the one thing missing in Xournal++ to make it perfect for my needs. 😺  Thanks either way!

rolandlo:
> Any sense of whether or not this is being prioritized? This is the one thing missing in Xournal++ to make it perfect for my needs. 😺 Thanks either way!

Yes, this is prioritized. See #6841

thisissami:
amazing! Thanks @rolandlo

PR #4673, opened 2023-02-23
Make pdf links popup without pressing Alt

rolandlo:
Fixes #4627.

Merging in 24 hours if no objections are raised.

PR #4772, opened 2023-03-29
Issue #4732 url link tool

CruzeBlade:
Hi, I added a new tool to embed links into the document as requested in issue #4732. 
This is just an early draft of that feature with basic functionality. 

I'm sharing the current status to get feedback and see if there is a chance to merge this in the future, as this will cause compatibility issues with older versions of Xournal++.

Here is a short demo:
ezgif com-video-to-gif

What still needs to be done:
- [x] Integrate with PDF export
- [x] Open URL by clicking in Browser


CruzeBlade:
After having work on this feature for the last couple of weeks, here is a small status update on what has changed since.

- Open the embedded link (normal click opens the editor dialog, and a ctrl + click will open the link in the default program) 
- Improved the link dialog editor
   - Automatically resizes the text field until a certain height
   - Changed text font and alignment options
   - link prefix combo box 
- "Hover" effect for the link element

Here are some images showing the current status:
| Improved dialog  | Hover effects |
|---|---|
| demo_improved_link_dialog  | demo_link_hover_effect |

Any feedback is greatly appreciated 😎!

netlimpopo:
>   * Open the embedded link (normal click opens the editor dialog, and a ctrl + click will open the link in the default program)
> 
>     * Improved the link dialog editor
Hello! 
Thanks for all the effort! This looks amazing hope it can get mefged. It would be a valueable contribution to he project!
I am no maintainer so I have bo say in it. 
In the past it has been noted in the forums though, that people with "touch devices only" need to be included as well (convertibles), so the programm gets more barrier free. They won't have a Ctrl-Key available when the transform their decive into a touch only one. 
Hence they won't be able to click the link, only edit the link. There have been similiar discussions on the implementation of the PDF Tools.  
#4673 back then "alt" was used, and later neglected. 
Regards N.

CruzeBlade:
Hi, thank you for the feedback. That is a very good point I haven't thought about. I will have to find a solution for that.

rolandlo (commented):
The tool looks great! I like it. Backwards compatibility is certainly something we need to consider carefully, but as it currently stands, files generated with the new version containing links can still be read by older versions, just the link elements are gone (and don't come back when saving). 

Another thing that has to be considered carefully is how the links can be activated and how they can be edited. I haven't thought it through carefully, but I tend to think that clicking on the link should activate it, not edit it. That would be consistent with PDF links with the PDF tool. Usually a link will need to be activated more often (and maybe by less versed people) than edited. Those who only use links may only have a touch device and stylus. Those who create links will usually have a keyboard, too, since otherwise how would they type the text. So CTRL+click (or similar) for editing (already created) links looks acceptable.

I hope this tool will eventually also allow to add internal links. It doesn't have to be done in this PR though.

I have briefly skimmed through the code and it looks nicely structured and readable. So thanks for putting all this effort into the new tool! It can bring Xournal++ a good step ahead.

CruzeBlade:
## Regarding how to open a link:

I'm personally not in favor of opening a link by simply clicking on it. Especially on touch devices, it is quite common to accidentally click somewhere (e.g. while scrolling). This would not only be annoying, but could potentially also be a security issue when somebody has embedded a malicious link. Therefore, I think it is very important that the user can see the full URL before opening it (hence the popover in the current version) and that there needs to be a mechanism to verify the user intentionally wants to open the link (hence the CTRL+click). But that's just my personal opinion on that topic.

My initial thought on how to improve the controls was:
* Double click for editing & creating
 -> In the link dialog, I would add a button to open the link
* Single click will highlight the link (see hover effects) 
 -> I would restyle the URL label as a link
 -> Clicking on the URL in the popover will open the link (as with pdf links)
* Single click + CTRL opens the link (as is)

That was just some first idea, and I'm open for discussion and other suggestions. 

## Regarding internal linking:

I would prefer to do this in a second PR, just to not further inflate the complexity of this PR. 

## Regarding backwards compatibility:

Does Xournal++ have a warning if users open a newer file with an older xournal++ version? If not something like this might be useful. (But out of scope for this PR)

CruzeBlade:
I have now implemented the new controls as mentioned in the previous post.

Here a small demo showing the enhanced popover (I added some control hints for the user):
Demo_new_controls

I currently don't own any touch device on which I can run Xournal++. Therefore, it would be great if somebody else could help me out and test if everything works as intended with touch devices.

CruzeBlade:
I have now finalized the implementation, and therefore I will make this a pull request.

rolandlo:
I have tried out the PR once more and I really like how you solved the way of choosing between editing and following the link. Could you please rebase on master? I want to look at the code in detail. Also please take into account  #4763 for making dialogs GTK4 compatible.

One minor thing I found that could possibly be improved: The default value for the text field for the URL is the same regardless of the URI scheme. It doesn't make sense though to write an email to `xournalpp.github.io` for example. So maybe a different default could be used for different schemes.

Mr-DblH:
Is there any news here yet? I find this almost indispensable for teaching purposes.

Exlonk:
Any news regarding this? It will be very useful. Thanks a lot

Photon89:
Hi guys,

are there any updates on this PR? Thanks!

rolandlo:
> Hi guys,
> 
> are there any updates on this PR? Thanks!

See #6841 for an update. [Binaries](https://github.com/xournalpp/xournalpp/pull/6841#issuecomment-3621804359) are available for testing.

rolandlo:
Closing this PR since #6841 was merged.

PR #4775, opened 2023-04-04
Drop Ubuntu 18.04, build AppImage on Ubuntu 20.04

rolandlo:
Needed in view of https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/azure-pipelines/jobs/release_ubuntu.yml
+++ b/azure-pipelines/jobs/release_ubuntu.yml
@@ -16,11 +16,6 @@ jobs:
     pool:
       vmImage: 'ubuntu-${{ parameters.ubuntu_version }}'
     displayName: 'Build for Ubuntu ${{ parameters.ubuntu_version }}'
-    variables:
-      # gtest version on Ubuntu 18.04 does not have a pkgconfig file,
-      # so download it instead.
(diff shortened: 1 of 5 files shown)


azure-pipelines/jobs/release_ubuntu.yml
```diff
@@ -16,11 +16,6 @@ jobs:
     pool:
       vmImage: 'ubuntu-${{ parameters.ubuntu_version }}'
     displayName: 'Build for Ubuntu ${{ parameters.ubuntu_version }}'
-    variables:
-      # gtest version on Ubuntu 18.04 does not have a pkgconfig file,
-      # so download it instead.
-      ${{ if eq(parameters['ubuntu_version'], '18.04') }}:
-        DOWNLOAD_GTEST_FLAG: '-DDOWNLOAD_GTEST=on'
```
bhennion, line 23: We can also remove the use of this variable (a couple of lines below)

rolandlo:
The Windows build (when creating installers) will be fixed separately. The AppImage (built on Ubuntu 20.04) works, but there is still an issue with the dark mode. That will be fixed separately as well. Merging therefore.

PR #4782, opened 2023-04-05
RAII in gui/GladeGui

bhennion:
This contains the first two (rather inoccuous) commits of #4763.


bhennion:
This is pretty trivial, so I'll be merging in 24 hours unless an objection is raised.

rolandlo (approved):
LGTM

Issue #4436, opened 2022-11-22
Copy to clipboard on Ctrl+C for PDF text selection tools


Technius:
copying on ctrl c would certainly be nice instead of pop up copy

_Originally posted by @Vozf in https://github.com/xournalpp/xournalpp/issues/1745#issuecomment-1322642485_

PR #4784, opened 2023-04-05
Enhancement #4436: Made Ctrl-C Shortcut to PDF Text Copy Functionality

TrevorShelton:
## Issue

#4436: PDF text could not be copied with a shortcut.

## Description

A Ctrl-C routine has been added to dive into the PDF Floating Tool Box and make it copy (if applicable) without needing to click the corresponding button.
(Function copyTextToClipboard() was changed from private to public in order to do this.)

### DIFF
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -464,6 +464,14 @@ auto MainWindow::onKeyPressCallback(GtkWidget* widget, GdkEventKey* event, MainW
         // editing text - give that control
         return false;
     }
+    if (event->keyval == GDK_KEY_c && event->state & GDK_CONTROL_MASK) {
+        // Shortcut to get selected PDF text.
+        PdfFloatingToolbox* tool = win->getPdfToolbox();
+        if (tool->hasSelection()) {
+            tool->copyTextToClipboard();
+            return true;
+        }
+    }
     if (event->keyval == GDK_KEY_Escape) {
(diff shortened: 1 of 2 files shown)


src/core/gui/MainWindow.cpp
```diff
@@ -464,6 +464,14 @@ auto MainWindow::onKeyPressCallback(GtkWidget* widget, GdkEventKey* event, MainW
         // editing text - give that control
         return false;
     }
+    if (event->keyval == GDK_KEY_c && event->state & GDK_CONTROL_MASK) {
+        // Shortcut to get selected PDF text.
+        PdfFloatingToolbox* tool = win->getPdfToolbox();
+        if (tool->hasSelection()) {
+            tool->copyTextToClipboard();
+        }
+        return true;
```
rolandlo, line 473: The event is only handled if there is a pdf text selection.
```suggestion
            return true;
        }
```

rolandlo (commented):
Thank you for the contribution. That's really a nice UX improvement. I have only a nitpick, otherwise LGTM.

rolandlo:
I looked at the whole key event handling in more details and improved it on the whole. I also included the PDF selection text copying via CTRL+c. 
We can still merge this PR and I will then rebase my PR on current master afterwards.

Could you please squash the two commits?

rolandlo:
Thanks again @TrevorShelton I merged your contribution.

Issue #3881, opened 2022-03-01
Strings not translated: toolbar names, color names, paper background names


bhennion:
The toolbar names, color names and paper background names are not translated through Crowdin.

This comes from the fact that those strings are specified in different configuration files

- Toolbar names: https://github.com/xournalpp/xournalpp/blob/master/ui/toolbar.ini
- Page template names: https://github.com/xournalpp/xournalpp/blob/master/ui/pagetemplates.ini
- Default palette color names: [here](https://github.com/xournalpp/xournalpp/blob/af1fdac91f74ecbd55a600f838c48b77f8f6dc97/src/core/gui/toolbarMenubar/model/ColorPalette.cpp#L103-130), then exported to a .gpl file in the user's config folder.

Ideally, all those strings should be fed to Crowdin, and Crowdin's tranlated strings .pot files should be used:

- Automatically added to the .ini files in the first two case
- A more complex system for the color names, since GIMP's .gpl format does not support string localization.

PR #4786, opened 2023-04-06
Add Turkish localization to some ini defines

sabriunal:
Some translatable ini defines translated into Turkish


bhennion:
See also #3881.

Merging this (as we don't have a long term more universal solution to this translation issue).

PR #4791, opened 2023-04-07
Cleanup latexcontroller

jhilmer:
I have removed some `double` to `int` conversion compiler warnigns. Use constexpr and anonymous namespace for some const parameters.

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -117,16 +117,16 @@ void LatexController::findSelectedTexElement() {
         // this will get the position of the Latex properly
         EditSelection* theSelection = control->getWindow()->getXournal()->getSelection();
         xoj::util::Rectangle<double> rect = theSelection->getSnappedBounds();
-        this->posx = static_cast<int>(rect.x);
-        this->posy = static_cast<int>(rect.y);
+        this->posx = rect.x;
+        this->posy = rect.y;
 
         if (auto* img = dynamic_cast<TexImage*>(this->selectedElem)) {
             this->initialTex = img->getText();
         } else if (auto* txt = dynamic_cast<Text*>(this->selectedElem)) {
             this->initialTex = "\\text{" + txt->getText() + "}";
         }
-        this->imgwidth = static_cast<int>(this->selectedElem->getElementWidth());
-        this->imgheight = static_cast<int>(this->selectedElem->getElementHeight());
+        this->imgwidth = this->selectedElem->getElementWidth();
+        this->imgheight = this->selectedElem->getElementHeight();
     } else {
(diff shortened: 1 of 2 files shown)


src/core/control/LatexController.cpp
```diff
@@ -115,16 +117,16 @@ void LatexController::findSelectedTexElement() {
         // this will get the position of the Latex properly
         EditSelection* theSelection = control->getWindow()->getXournal()->getSelection();
         xoj::util::Rectangle<double> rect = theSelection->getSnappedBounds();
-        this->posx = rect.x;
-        this->posy = rect.y;
+        this->posx = static_cast<int>(rect.x);
+        this->posy = static_cast<int>(rect.y);
```
bhennion, line 121: All those static_casts feel off to me. Bluntly casting should never happen: we probably want to std::round, std::floor or std::ceil those values before.
In this case, I don't know which one it should be. This needs understanding the class a bit better.

jhilmer: I this case the posx, posy, imgWidth, imgHeight should be double. They are used to in method call on TexImage there expect double.

jhilmer: By changing the attributes to double the static_cast is mostly avoided.

src/core/control/LatexController.cpp
```diff
@@ -109,8 +111,8 @@ void LatexController::findSelectedTexElement() {
     this->page = this->doc->getPage(pageNr);
     this->layer = page->getSelectedLayer();
 
-    this->selectedElem = view->getSelectedTex() != nullptr ? static_cast<Element*>(view->getSelectedTex()) :
-                                                             static_cast<Element*>(view->getSelectedText());
+    this->selectedElem = view->getSelectedTex() != nullptr ? dynamic_cast<Element*>(view->getSelectedTex()) :
+                                                             dynamic_cast<Element*>(view->getSelectedText());
```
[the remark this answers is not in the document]
jhilmer: Will revert

src/core/control/LatexController.cpp
```diff
@@ -42,8 +42,10 @@
 
 using std::string;
 
-const Color LIGHT_PREVIEW_BACKGROUND = Colors::white;
-const Color DARK_PREVIEW_BACKGROUND = Colors::black;
+namespace {
+constexpr Color LIGHT_PREVIEW_BACKGROUND = Colors::white;
+constexpr Color DARK_PREVIEW_BACKGROUND = Colors::black;
+}  // namespace
```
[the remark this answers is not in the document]
jhilmer: Will remove the namespace.  Wow I learn something new.

bhennion (approved):
LGTM!

PR #4792, opened 2023-04-07
PageView cleanup

jhilmer:


src/core/gui/PageView.cpp
```diff
@@ -644,10 +644,6 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
         if (this->getPage()->getPdfPageNr() != npos) {
             auto page = pdfDoc.getPage(this->getPage()->getPdfPageNr());
 
-            Layer* layer = this->page->getSelectedLayer();
-            const Layer::Index layerId = this->page->getSelectedLayerId();
-            const bool isBackgroundLayer = (layerId == 0);
```
jhilmer, line 661: Dead code ?? Are there any side effects of those lines? I think it safe to remove.

rolandlo: They can be removed indeed. It looks like I forgot to remove those lines in #4673.

src/core/gui/PageView.cpp
```diff
@@ -253,41 +243,40 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
...
-            this->inputHandler = nullptr;
+            eraseViewsOf(this->overlayViews, this->inputHandler.get());
+            this->inputHandler.reset();
         }
 
         Control* control = this->xournal->getControl();
         switch (h->getDrawingType()) {
             case DRAWING_TYPE_LINE:
-                this->inputHandler = new RulerHandler(control, getPage());
+                this->inputHandler = (std::make_unique<RulerHandler>(control, getPage()));
```
rolandlo, line 253: There is no need for extra brackets here and in the other cases.
```suggestion
                this->inputHandler = std::make_unique<RulerHandler>(control, getPage());
```

rolandlo (approved):
Just a minor nitpick. Otherwise this looks good to me.

rolandlo (approved):
Thanks @jhilmer !

PR #4795, opened 2023-04-10
Remove unused functions or marked as maybe_unused

rolandlo:


src/core/model/Stroke.h
```diff
@@ -129,7 +129,7 @@ class Stroke: public AudioElement {
     void setPointVectorInternal(const Range* const snappingBox);
 
 public:
-    void deletePoint(int index);
+    [[maybe_unused]] void deletePoint(int index);
```
bhennion, line 132: I think we can remove this function. It make little sense to me to be able to delete any point in a stroke

bhennion (approved):

src/core/model/Stroke.h
rolandlo: Thanks for the review. I removed that function as well in a second commit.

PR #4797, opened 2023-04-10
Refactor key event handling

### LINKED ISSUE #977: Ctrl+shift+(Left/right arrow) does not select text, it changes layers

rolandlo:
Fixes #977 (comments in the discussion). Supersedes #4784 

- Let the key-press-event and key-release-event for the main window propogate the event to the focus widget and up the focus container chain until a widget handles the event
- Make the keyboard input handler only direct the GdkKeyEvents to the XournalView's handler
- Make the XournalView's handler first let the current page handle the key event for tools owned by the page
 (textEditor, inputHandler, verticalSpace) and then
handle selection, scrolling, page operations and color switching on its own

The main improvement relates to the fact that now keyboard accelerators can be used (for plugins and menus) that would otherwise make the text editor unfunctional.

### DIFF
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -82,9 +82,9 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
 
     g_signal_connect(get("buttonCloseSidebar"), "clicked", G_CALLBACK(buttonCloseSidebarClicked), this);
 
-
-    // "watch over" all events
-    g_signal_connect(this->window, "key-press-event", G_CALLBACK(onKeyPressCallback), this);
+    // "watch over" all key events
+    g_signal_connect(this->window, "key-press-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
+    g_signal_connect(this->window, "key-release-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
 
     // need to create tool buttons registered in plugins, so they can be added to toolbars
     control->registerPluginToolButtons(this->toolbar.get());
@@ -450,33 +450,6 @@ void MainWindow::updateScrollbarSidebarPosition() {
 }
 
 void MainWindow::buttonCloseSidebarClicked(GtkButton* button, MainWindow* win) { win->setSidebarVisible(false); }
-
-auto MainWindow::onKeyPressCallback(GtkWidget* widget, GdkEventKey* event, MainWindow* win) -> bool {
-
-    if (win->getXournal()->getSelection()) {
-        // something is selected - give that control
-        return false;
-    }
-    if (win->getXournal()->getTextEditor()) {
-        // editing text - give that control
-        return false;
-    }
-    if (event->keyval == GDK_KEY_c && event->state & GDK_CONTROL_MASK) {
-        // Shortcut to get selected PDF text.
-        PdfFloatingToolbox* tool = win->getPdfToolbox();
-        if (tool->hasSelection()) {
-            tool->copyTextToClipboard();
-            return true;
-        }
-    }
-    if (event->keyval == GDK_KEY_Escape) {
-        win->getControl()->getSearchBar()->showSearchBar(false);
-        return true;
-    }
-
-
-    return false;
-}
 
 auto MainWindow::deleteEventCallback(GtkWidget* widget, GdkEvent* event, Control* control) -> bool {
     control->quit();
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -690,10 +690,6 @@ auto XojPageView::onKeyPressEvent(GdkEventKey* event) -> bool {
             endText();
             return true;
         }
-        if (xournal->getSelection()) {
-            xournal->clearSelection();
-            return true;
-        }
 
         return false;
     }
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -47,6 +47,10 @@
 
 using xoj::util::Rectangle;
 
+constexpr int REGULAR_MOVE_AMOUNT = 3;
+constexpr int SMALL_MOVE_AMOUNT = 1;
+constexpr int LARGE_MOVE_AMOUNT = 10;
+
 std::pair<size_t, size_t> XournalView::preloadPageBounds(size_t page, size_t maxPage) {
     const size_t preloadBefore = this->control->getSettings()->getPreloadPagesBefore();
     const size_t preloadAfter = this->control->getSettings()->getPreloadPagesAfter();
@@ -128,8 +132,12 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         }
     }
 
+    // Todo(gtk4) there is no GdkEventKey. Use a GdkEventControllerKey and connect to its key-pressed signal
+    guint keyval = event->keyval;
+    GdkModifierType keystate = static_cast<GdkModifierType>(event->state);
+
     if (auto* tool = getControl()->getWindow()->getPdfToolbox(); tool->hasSelection()) {
-        if (event->keyval == GDK_KEY_c && event->state & GDK_CONTROL_MASK) {
+        if (keyval == GDK_KEY_c && keystate & GDK_CONTROL_MASK) {
             // Shortcut to get selected PDF text.
             tool->copyTextToClipboard();
             return true;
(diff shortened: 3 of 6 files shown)


rolandlo:
A little demo on what this improves for key event handling. The plugin can now use a single-letter accelerator like "t" without messing with the text tool.

https://user-images.githubusercontent.com/40485433/231143064-6ba24da1-2337-40ab-8808-68e833e67a98.mp4

I have also tested the search bar, the LaTex tool, zooming in/out and the geometry tools. It works all fine.
This PR is ready to review. It's just a reorganization and does things like they are supposed to (the focus widget gets to handle the key event first).

src/core/gui/XournalView.cpp
```diff
@@ -139,6 +139,45 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
...
+        if (event->keyval == GDK_KEY_Escape) {
+            clearSelection();
+            return true;
+        }
+
+        int d = 3;
+        if (event->state & GDK_MOD1_MASK) {
+            d = 1;
+        } else if (event->state & GDK_SHIFT_MASK) {
+            d = 20;
```
bhennion, line 160: Could you make those magic constants constexpr somewhere?

src/core/gui/XournalView.cpp
```diff
@@ -139,6 +139,45 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
...
+        }
+
+        int xdir = 0;
+        int ydir = 0;
+        if (event->keyval == GDK_KEY_Left) {
```
bhennion, line 165: The [GTK4 migration guide](https://docs.gtk.org/gtk4/migrating-3to4.html#stop-using-direct-access-to-gdkevent-structs) suggests to avoid directly accessing member variables of GdkEvent.
They say gdk3 already has some accessors, but in this case I'm not sure they match the gdk4 ones.
Anyway, if you feel like it, you can try and adapt those accesses to limit the gtk4 porting effort down the road. I don't how much of that is actually possible, so feel free to ignore this comment.

src/core/gui/MainWindow.cpp
```diff
@@ -82,9 +82,9 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
 
     g_signal_connect(get("buttonCloseSidebar"), "clicked", G_CALLBACK(buttonCloseSidebarClicked), this);
 
-
-    // "watch over" all events
-    g_signal_connect(this->window, "key-press-event", G_CALLBACK(onKeyPressCallback), this);
+    // "watch over" all key events
+    g_signal_connect(this->window, "key-press-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
+    g_signal_connect(this->window, "key-release-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
```
bhennion, line 87: Are those not already the default callbacks? The `gtk_window_propagate_key_event` [doc](https://docs.gtk.org/gtk3/method.Window.propagate_key_event.html) says
> This is normally called by the default ::key_press_event and ::key_release_event handlers for toplevel windows

Could we maybe remove those two lines then?

bhennion (commented):
I have a couple of questions. See below

src/core/gui/MainWindow.cpp
rolandlo: I also thought that would mean we can remove these two lines. It's not true though, they make a difference. Plugin accelerators are called before the text editor can handle the input without those lines. So we must leave them.

src/core/gui/XournalView.cpp
rolandlo: Sure, I will.

src/core/gui/XournalView.cpp
rolandlo: Right, I keep forgetting about that. There is `gdk_event_get_keyval` (like already used e.g. in GeometryToolInputHandler)

src/core/gui/MainWindow.cpp
rolandlo: Here's the explanation (see https://docs.gtk.org/gtk3/input-handling.html):

> For key events, the top-level window’s default [GtkWidget::key-press-event](https://docs.gtk.org/gtk3/signal.Widget.key-press-event.html) and [GtkWidget::key-release-event](https://docs.gtk.org/gtk3/signal.Widget.key-release-event.html) signal handlers handle mnemonics and accelerators first. Other key presses are then passed to [gtk_window_propagate_key_event()](https://docs.gtk.org/gtk3/method.Window.propagate_key_event.html) which propagates the event upwards from the window’s current focus widget ([gtk_window_get_focus()](https://docs.gtk.org/gtk3/method.Window.get_focus.html)) to the top-level.

src/core/gui/XournalView.cpp
bhennion: Yes, but gdk4 only has gdk_key_event_get_keyval taking a GdkKeyEvent (but it still has gdk_event_get_modifier_state)

src/core/gui/PageView.cpp
```diff
@@ -304,8 +304,9 @@ auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
         auto* zoomControl = this->getXournal()->getControl()->getZoomControl();
         this->verticalSpace = std::make_unique<VerticalToolHandler>(this->page, this->settings, y, pos.isControlDown());
         this->overlayViews.emplace_back(this->verticalSpace->createView(this, zoomControl, this->settings));
-    } else if (h->getToolType() == TOOL_SELECT_RECT || h->getToolType() == TOOL_SELECT_REGION || h->getToolType() == TOOL_SELECT_MULTILAYER_RECT ||
-               h->getToolType() == TOOL_SELECT_MULTILAYER_REGION || h->getToolType() == TOOL_PLAY_OBJECT || h->getToolType() == TOOL_SELECT_OBJECT ||
+    } else if (h->getToolType() == TOOL_SELECT_RECT || h->getToolType() == TOOL_SELECT_REGION ||
+               h->getToolType() == TOOL_SELECT_MULTILAYER_RECT || h->getToolType() == TOOL_SELECT_MULTILAYER_REGION ||
+               h->getToolType() == TOOL_PLAY_OBJECT || h->getToolType() == TOOL_SELECT_OBJECT ||
```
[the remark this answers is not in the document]
bhennion: it looks true, but those changes would be out of scope of this PR, I think

MichelReims:
Indeed, it would be interesting if the shortcuts were not activated when writing text with Text Tool

src/core/gui/XournalView.cpp
```diff
@@ -139,36 +143,40 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
         }
     }
 
+    // TODO: With GTK 4 there is no GdkEventKey. Use a GdkEventControllerKey and connect to its key-pressed signal
```
bhennion, line 146: For consistency: I think most of the GTK4 indications are written this way
```suggestion
    // Todo(gtk4) there is no GdkEventKey. Use a GdkEventControllerKey and connect to its key-pressed signal
```

bhennion (approved):
LGTM!

src/core/gui/XournalView.cpp
rolandlo: Thanks for the review! I changed the comment like suggested.

mjg:
I don't yet know why, but apparantly this killed all shortcuts involving CTRL, in particular the tool modifiers such as line, rectangle etc.

rolandlo:
> I don't yet know why, but apparantly this killed all shortcuts involving CTRL, in particular the tool modifiers such as line, rectangle etc.

Looks like these shortcuts (CTRL+number) are absorbed by these lines:
https://github.com/xournalpp/xournalpp/blob/e6cf8a1b36ad58fa2b84ea1f0cff5e3646c0670b/src/core/gui/XournalView.cpp#L303-L313

CTRL+d still inserts a page, doesn't it?

mjg:
OK, not all :|
Still, the tool mods do not work, probably because they are absorbed. Note that they switch colours instead, i.e. the CTRL is absorbed or the colour shortcut applied regardless of CTRL.

rolandlo:
Yes, so we just need to make sure that the color switch only happens without modifiers. I will open a PR.

PR #4803, opened 2023-04-16
new feature: changing the pdf background

simone-romiti:
The `File` item in the toolbar now has the option `Change PDF Background`. 

## How it works

- If no PDF file has been opened, throws an error and doesn't modify the document.
- If a PDF has been opened, the background is changed.

## Implementation

- Added the item in the `File` menu of the toolbar-
- Modified the `ActionType` enum to include `ACTION_CHANGE_BACKGROUND`
- Added the case for the latter in `Control::actionPerformed`
- Defined the new method `Control::changePdfBackground()` that invokes `setPdfReplacement()` method of the `LoadHandler` in the scope of the function.


src/core/control/Control.h
```diff
@@ -88,6 +88,7 @@ class Control:
     // Menu File
     bool newFile(std::string pageTemplate = "", fs::path filepath = {});
     bool openFile(fs::path filepath = "", int scrollToPage = -1, bool forceOpen = false);
+    bool changePdfBackground(fs::path filepath = "", int scrollToPage = -1, bool forceOpen = false);
```
rolandlo, line 91: The parameters `filepath`, `scrollToPage` and `forceOpen` are not used anywhere. I think they should be removed.

src/core/control/Control.cpp
```diff
@@ -2394,6 +2397,39 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+    }
+
+    this->closeDocument();
+
+    this->doc->lock();
+    this->doc->clearDocument();
+    *this->doc = *loadedDocument;
+    this->doc->unlock();
+
+    fileLoaded(scrollToPage);
```
rolandlo, line 2429: `scrollToPage` is always `-1`, since `changePdfBackground` is only called with default parameters.

src/core/control/Control.cpp
```diff
@@ -2394,6 +2397,39 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
     return true;
 }
 
+auto Control::changePdfBackground(fs::path filepath, int scrollToPage, bool forceOpen) -> bool {
+
+    LoadHandler loadHandler;
+
+    const fs::path current_filepath = ((*this).doc)->getFilepath();
+    if (current_filepath.empty()) {
```
rolandlo, line 2405: There are some issues with this approach. Whether the PDF background can be changed only depends on whether the `.xopp`-file has been saved at some point or not. 
- There are `.xopp`-files without PDF background and it doesn't make sense to change the PDF background in that case.
- When the document has unsaved changes, these are discarded (without any warning)
- When you have a PDF background (annotate PDF), but it's not saved yet, you can't change the PDF background, although it would make sense.

rolandlo (changes requested):
Thanks for the PR. I think it can be useful. However there are cases, where the "Change PDF background" doesn't make sense or doesn't work as expected. So please think about these cases (mentioned in the code review).

src/core/control/Control.cpp
simone-romiti: Thank you for the comments :).

* Yes, in fact if you try to do that you get the error "Error in changing the background: please open some file first."
* Fixed with last commit
* Fixed with last commit

rolandlo:
Seems you merged master into your branch (or something like that). Instead you should have rebased your branch on master. I just want to have your commits based on master. Can you please fix this up?

simone-romiti:
I'm sorry, it's the first time I rebase. I did the following:
```
git remote add upstream https://github.com/xournalpp/xournalpp
git fetch upstream 
git rebase upstream/master
# here it asked me to pull before pushing again
git config pull.rebase true
git pull
```
is it correct?

rolandlo:
Those are your commits (except of the merge commit), right?
grafik

Usually rebasing (after adding the upstream remote as you did) works by
```term
git rebase upstream/master
git push --force
```
You should ignore the request to pull. However now things are messed up and it's more complicated. If you wish I can do it for you.

rolandlo:
Closing in favor of #5037

simone-romiti:
I see, I have to force the push... I simplified things creating a new PR here: https://github.com/xournalpp/xournalpp/pull/5037
I'd discard this one you agree.

Issue #3090, opened 2021-05-02
Cursor size doesn't update after using "ACTION_SIZE_..."


jdujava:
**Affects versions :**
 - OS: Linux (Arch)
 - Desktop environment: X11
 - Xournal++ 1.1.0+dev
 - libgtk: 3.24.29
 - Installation method: AUR - xournalpp-git

**Describe the bug**
I am using a plugin to change sizes of tools, but the size of cursor indicator
changes only after changing the color or after switching to another tool and
back.

**To Reproduce**
Create a simple plugin, bind following actions to any key:
`
app.uiAction({["action"] = "ACTION_SIZE_THICK"})
`
or
`
app.uiAction({["action"] = "ACTION_TOOL_PEN_SIZE_THICK"})
`
Pen size changes, but the cursor indicator stays the same.

**Expected behavior**
Cursor indicator changes automatically with change of the pen size.

**Additional context**
After quick inspection of `Control.cpp` code, it seems to me that calling
`
getCursor()->updateCursor();
`
or
`
toolSizeChanged();
`
at appropriate places could fix the problem (only a guess).

rolandlo:
This also happens when using the menubar or the toolbar to change the pen size.

rolandlo:
The bug seems to happen here:
https://github.com/xournalpp/xournalpp/blob/2d8201f4cebce3d5cc460185c185b6e5449aeb40/src/gui/XournalppCursor.cpp#L419-L424

Xournal++ thinks it doesn't need to change the cursor, because the `size` variable it uses is not the actual cursor Size (but either 3 or 5 depending on whether the pen or the highlighter is used)

https://github.com/xournalpp/xournalpp/blob/2d8201f4cebce3d5cc460185c185b6e5449aeb40/src/gui/XournalppCursor.cpp#L393-L393

https://github.com/xournalpp/xournalpp/blob/2d8201f4cebce3d5cc460185c185b6e5449aeb40/src/gui/XournalppCursor.cpp#L406-L406

VanniPrashar:
Can I fix this please

Issue #4718, opened 2023-03-12
Highlight Radius caps off too soon


slotThe:
**Affects versions :**

I reckon all of them, but just for completeness:

 - OS: GNU/Linux (Void Linux)
 - Desktop environment: X11
 - Which version of libgtk do you use: `3.24.37`
 - Version of Xournal++: `1.1.3`
 - Installation method: Void's package repos

**Describe the bug**

The "highlight cursor position" feature seems to cap both options (radius and border width) at 30 pixels:

![2023-03-12-171100_518x135_scrot](https://user-images.githubusercontent.com/50166980/224557436-95a5ebea-f417-4b79-97d5-97a5f7db65a3.png)

This seems like an arbitrary limit; could it perhaps be removed/increased significantly?  

The reason I am asking (and why this is a bug report, and not a feature request or similar) is because of accessibility.  A friend of mine—severely visually impaired—recently got a larger monitor, which sports a much higher resolution than the last one.  In fact, so much higher that there were several hindrances in scaling certain graphical applications to a level of "sort of acceptable".  In Xournal++, the 30px radius is much too small for him to see the cursor, so he's basically left guessing where it is, or writing something and searching the monitor for the text.  Not an ideal situation, to say the least!  I reckon more visually impaired users would also benefit from a larger radius.

bhennion:
Thanks for the report.
Can you not set up a HiDPI scaling for his screen? That'd increase the size of everything (Xournal++ and otherwise)

slotThe:
Thanks for the fast reply!  He played around with HiDPI stuff a bit, but things like `Xft.dpi` only affect text scaling, and `xrandr --scale` and friends make everything quite blurry (which apparently really hurts text recognition)

rolandlo:
Does running Xournal++ via 
```term
GDK_SCALE=2 xournalpp
``` 
(in the terminal) help?

slotThe:
Oh, indeed, this seems to work!  Can't believe we missed that, thank you.  Still, I think if there's no technical reason that the cap is at 30 pixels, it would nevertheless be better it that was removed (or increased)

bhennion:
There is a technical reason, actually: there is a bound to cursor size (you can get it by calling gdk_display_get_maximal_cursor_size()), most likely for performance reasons. On my setup, this returns 128x128 pixels. This means we could double the 30 pts radius limit, but nothing more.

slotThe:
[...]

PR #4807, opened 2023-04-18
Issue #3090 update cursor size

jen-kosek:
Hi, 

I changed the hash within XournalppCursor::createHighlighterOrPenCursor so that the cursor's size gets updated when the pen/highlighter's thickness changes.

Solves #3090.

### DIFF
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -441,12 +441,13 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
     auto cursorType = control->getSettings()->getStylusCursorType();
     auto cursor = (cursorType == STYLUS_CURSOR_NONE) ? CRSR_BLANK_CURSOR : CRSR_PENORHIGHLIGHTER;
     bool bright = control->getSettings()->isHighlightPosition();
-    int height = size;
-    int width = size;
+    double cursorSize = control->getToolHandler()->getThickness() * control->getZoomControl()->getZoom();
+    int width = 2 + static_cast<int>(cursorSize);
+    int height = 2 + static_cast<int>(cursorSize);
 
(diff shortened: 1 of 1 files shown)


src/core/gui/XournalppCursor.cpp
```diff
@@ -441,12 +441,13 @@ auto XournalppCursor::createHighlighterOrPenCursor(int size, double alpha) -> Gd
     auto cursorType = control->getSettings()->getStylusCursorType();
     auto cursor = (cursorType == STYLUS_CURSOR_NONE) ? CRSR_BLANK_CURSOR : CRSR_PENORHIGHLIGHTER;
     bool bright = control->getSettings()->isHighlightPosition();
-    int height = size;
-    int width = size;
+    double cursorSize = control->getToolHandler()->getThickness() * control->getZoomControl()->getZoom();
+    int width = 2 + static_cast<int>(cursorSize);
+    int height = 2 + static_cast<int>(cursorSize);
```
rolandlo, line 446: The cursor shouldn't become too big, see https://github.com/xournalpp/xournalpp/issues/4718#issuecomment-1468593480. Note that `height` and `width` are set to `90` when using the big stylus cursor or when highlighting the cursor. So what do you think about this suggestion:
```suggestion
    double cursorSize = std::min(90., control->getToolHandler()->getThickness() * control->getZoomControl()->getZoom());
    int width = static_cast<int>(std::ceil(cursorSize));
    int height = static_cast<int>(std::ceil(cursorSize));
```

rolandlo (changes requested):
Thanks for the PR.
Please apply clang format. Let me know if you need help with it.
Some changes are required, since the cursor must not become too big. Also there is an issue with cursor highlighting in case of a large cursor size, that could be solved like in the suggestion.

rolandlo:
Closing due to lack of inactivity. Please reopen if you decide to rebase on master and finish it.

PR #4796, opened 2023-04-10
More gtk4 dialogs

bhennion:
This follows #4763.
The structure set up in #4763 seems to fit all needs for now.

This PR can be split further down if need be (down to a per-dialog PR).
The diff is enormous, mainly because of Glade files changes (removing obsolete GtkAlignment makes huge changes mainly about indentation). Code changes are actually small (except SettingsDialog.cpp, but nothing complicated happened there either, just some renaming essentially).

PR #4808, opened 2023-04-19
Fix "File bug report" dialog and more

bhennion:
This PR fixes the "File bug report" dialog: it seems a dialog is supposed to pop-up on startup, when an errorlog is found.
This was broken (the errorlog files never matched the applied name filter, so the dialog never appeared).

This fixes that, as well as a couple of papercuts (encountered while working on #4796, but whose fix should be merged before that)

1. Failing assert when closing the Insert Pdf Page dialog via Escape, Alt+F4 or window-close icon.
2. Fix conversion warnings in Settings.cpp

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -10,6 +10,7 @@
 #include <locale>     // for locale
 #include <memory>     // for unique_ptr, allocator
 #include <optional>   // for optional, nullopt
+#include <sstream>    // for stringstream
 #include <stdexcept>  // for runtime_error
 #include <string>     // for string, basic_string
 #include <vector>     // for vector
@@ -138,6 +139,17 @@ auto migrateSettings() -> MigrateResult {
     return {MigrateStatus::NotNeeded, ""};
 }
 
+static void deleteFile(const fs::path& file) {
+    std::error_code error;
+    if (!fs::remove(file, error)) {
+        std::stringstream msg;
+        msg << FS(_F("Failed to delete file: {1}") % file.u8string()) << std::endl;
+        msg << error << std::endl << error.message() << std::endl;
+        msg << FS(_F("Please delete the file manually"));
+        XojMsgBox::showErrorToUser(nullptr, msg.str());
+    }
+}
+
 void checkForErrorlog() {
     const fs::path errorDir = Util::getCacheSubfolder(ERRORLOG_DIR);
     if (!fs::exists(errorDir)) {
@@ -146,7 +158,7 @@ void checkForErrorlog() {
 
     std::vector<fs::path> errorList;
     for (auto const& f: fs::directory_iterator(errorDir)) {
-        if (f.is_regular_file() && f.path().stem() == "errorlog") {
+        if (f.is_regular_file() && f.path().filename().string().substr(0, 8) == "errorlog") {
             errorList.emplace_back(f);
         }
     }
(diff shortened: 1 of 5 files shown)


src/core/control/XournalMain.cpp
```diff
@@ -158,47 +170,36 @@ void checkForErrorlog() {
     std::sort(errorList.begin(), errorList.end());
     std::string msg =
             errorList.size() == 1 ?
-                    _("There is an errorlogfile from Xournal++. Please send a Bugreport, so the bug may be fixed.") :
-                    _("There are errorlogfiles from Xournal++. Please send a Bugreport, so the bug may be fixed.");
+                    _("There is an errorlogfile from Xournal++. Please fill a Bugreport, so the bug may be fixed.") :
+                    _("There are errorlogfiles from Xournal++. Please fill a Bugreport, so the bug may be fixed.");
```
rolandlo, line 174: I guess you wanted to write it like this:
```suggestion
                    _("There is an errorlogfile from Xournal++. Please file a Bugreport, so the bug may be fixed.") :
                    _("There are errorlogfiles from Xournal++. Please file a Bugreport, so the bug may be fixed.");
```

src/core/gui/dialog/SettingsDialog.cpp
```diff
@@ -907,16 +908,16 @@ void SettingsDialog::save() {
...
             static_cast<double>(gtk_spin_button_get_value(GTK_SPIN_BUTTON(get("spStrokeRecognizerMinSize")))));
 
-    int selectedInputDeviceIndex = gtk_combo_box_get_active(GTK_COMBO_BOX(get("cbAudioInputDevice"))) - 1;
-    if (selectedInputDeviceIndex >= 0 && selectedInputDeviceIndex < static_cast<int>(this->audioInputDevices.size())) {
-        settings->setAudioInputDevice(static_cast<int>(this->audioInputDevices[selectedInputDeviceIndex].getIndex()));
+    size_t selectedInputDeviceIndex =
+            static_cast<size_t>(gtk_combo_box_get_active(GTK_COMBO_BOX(get("cbAudioInputDevice")))) - 1;
+    if (selectedInputDeviceIndex < this->audioInputDevices.size()) {
+        settings->setAudioInputDevice(this->audioInputDevices[selectedInputDeviceIndex].getIndex());
     }
```
rolandlo, line 915: This doesn't work (and didn't work) when switching from a non-default setting to the default setting. It should be possible to switch back to the system default.

src/core/gui/dialog/SettingsDialog.cpp
```diff
@@ -907,16 +908,16 @@ void SettingsDialog::save() {
...
-    int selectedOutputDeviceIndex = gtk_combo_box_get_active(GTK_COMBO_BOX(get("cbAudioOutputDevice"))) - 1;
-    if (selectedOutputDeviceIndex >= 0 &&
-        selectedOutputDeviceIndex < static_cast<int>(this->audioOutputDevices.size())) {
-        settings->setAudioOutputDevice(
-                static_cast<int>(this->audioOutputDevices[selectedOutputDeviceIndex].getIndex()));
+    size_t selectedOutputDeviceIndex =
+            static_cast<size_t>(gtk_combo_box_get_active(GTK_COMBO_BOX(get("cbAudioOutputDevice")))) - 1;
+    if (selectedOutputDeviceIndex < this->audioOutputDevices.size()) {
+        settings->setAudioOutputDevice(this->audioOutputDevices[selectedOutputDeviceIndex].getIndex());
     }
```
rolandlo, line 921: Same as above.

rolandlo (changes requested):
Interesting, I didn't know about the (previously non-working) bug report feature. Maybe we will get a lot of bug reports in the near future. Anyway good that you fixed it.

I have found one bug (regarding audio device settings) which existed before but could easily be fixed in this PR.

src/core/gui/dialog/SettingsDialog.cpp
bhennion: https://github.com/xournalpp/xournalpp/pull/4808/commits/3b3a1d58afbb3747e25e565d21fe07492c25192f should do the trick

rolandlo (approved):
LGTM

bhennion:
Thanks for the review!

PR #4811, opened 2023-04-20
Fix drawing type activation shortcuts

rolandlo:
Fixes an issue mentioned in https://github.com/xournalpp/xournalpp/pull/4797#issuecomment-1516222132

The number keys 1,2,...,0 should only change color if they are pressed without modifiers.
CTRL+1, CTRL+2, ..., CTRL+8 should activate the drawing types.


mjg:
Thanks, looking good here!
I guess the refactoring was fine and only exposed that one here.

rolandlo:
Ctrl+L (Create new layer), 
Shift+Page Up (Next Layer) 
Shift+Ctrl+Page Up (Previous annotated page) 
still don't work. The other shortcuts work fine.

jdujava:
Previous to #4797 I had a modified implementation of color switching (keys 1-9 were mapped to colors in the toolbar from bottom up, which I prefer) and everything worked fine. Do I understand correctly that as of now there is no way of remapping default shortcuts?

If a user could remap default shortcuts (with the text editor still functional), it would be really great!

rolandlo:
That's correct. Currently there is no way of remapping default shortcuts (except of those defined in the `.glade` files, which you could edit manually without any recompilation).

Technius:
The refactor is a good step towards implementing the custom keybindings feature, since we now intercept all key events in a single place :)

src/core/control/tools/TextEditor.cpp
```diff
@@ -319,14 +319,19 @@ auto TextEditor::onKeyPressEvent(GdkEventKey* event) -> bool {
...
             increaseFontSize();
             return true;
         }
-        if (event->keyval == GDK_KEY_minus) {
+        if (event->keyval == GDK_KEY_minus || event->keyval == GDK_KEY_KP_Subtract) {
             decreaseFontSize();
             return true;
         }
+    } else if ((event->state & modifiers) == (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) {
+        if (event->keyval == GDK_KEY_plus || event->keyval == GDK_KEY_KP_Add) {
```
bhennion, line 331: Could you add a comment as to why this case is special? Why `Ctrl`+`Shift`+`+` and not also `Ctrl`+`Shift`+`-`?
Is it related to some keyboard layouts having the + in a "Caps" position?

src/core/gui/XournalView.cpp
```diff
@@ -712,7 +710,8 @@ void XournalView::layoutPages() {
     Layout* layout = gtk_xournal_get_layout(this->widget);
     layout->recalculate();
 
-    // Todo (fabian): the following lines are conceptually wrong, the Layout::layoutPages function is meant to be called
+    // Todo (fabian): the following lines are conceptually wrong, the Layout::layoutPages function is meant to be
+    // called
     //                by an expose event, but removing it, will break "add page".
```
bhennion, line 715:
```suggestion
    // called by an expose event, but removing it, will break "add page".
```

bhennion (approved):
LGTM, appart from a comment question and a clang-format mishap

src/core/gui/XournalView.cpp
rolandlo: I found an issue with CTRL+<number> changing the color of the text in the text editor. I will have to look at it tomorrow.

src/core/control/tools/TextEditor.cpp
rolandlo: I found a keyboard layout independent solution using consumed modifiers.

rolandlo:
Squashing the last two commits into one and merging in 24 hours if no objections are raised.

PR #4824, opened 2023-04-24
Fix #4822: Rerouted Esc and Backspace in Spline Handler to More Complete Spline Finish Functions.

### LINKED ISSUE #4822: Crash when pressing `ESC` during spline drawing.
**Affects versions :**
 - OS: Linux (Arch)
 - Desktop environment: X11
 - Version of libgtk: 3.24.37
 - Version of Xournal++: 1.1.3+dev, built on April 17
 - Installation method: manual

**Describe the bug**
Pressing `ESC` during spline drawing results in a crash.

**To Reproduce**
1. Choose `Draw Spline` tool
2. Start to draw a spline
3. Press `ESC` to stop drawing the spline
4. Trying to draw another spline (or doing other actions) results in a crash.

**Expected behavior**
Pressing `ESC` should properly stop drawing the spline, so it is possible to draw another spline (or to do other actions) without crash.

**Additional context**
Following line of output when running `xournalpp` from the terminal is probably relevant:
```
xournalpp: .../xournalpp/src/core/control/tools/SplineHandler.cpp:211: virtual void SplineHandler::onButtonPressEvent(const PositionInputData&, double): Assertion `this->knots.empty() && this->tangents.empty()' failed.
```

TrevorShelton:
## Reason
Issue #4822

## Description
Fixed the Esc and Backspace keys' links to end the spline by relinking them to a more complete spline ending function in XournalView. For this, the MainWindow and XournalView headers needed to be added.

### DIFF
--- a/src/core/control/tools/SplineHandler.cpp
+++ b/src/core/control/tools/SplineHandler.cpp
@@ -60,13 +60,11 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
 
     switch (event->keyval) {
         case GDK_KEY_Escape: {
-            this->finalizeSpline();
             return true;
         }
         case GDK_KEY_BackSpace: {
             if (this->knots.size() == 1) {
-                this->finalizeSpline();
-                return true;
+                return false;
             }
             this->deleteLastKnotWithTangent();
             assert(!this->knots.empty() && this->knots.size() == this->tangents.size());
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -704,11 +704,8 @@ auto XojPageView::onKeyReleaseEvent(GdkEventKey* event) -> bool {
 
     if (this->inputHandler && this->inputHandler->onKeyEvent(event)) {
         DrawingType drawingType = this->xournal->getControl()->getToolHandler()->getDrawingType();
-        if (drawingType == DRAWING_TYPE_SPLINE) {  // Spline drawing has been finalized
-            if (this->inputHandler) {
-                assert(hasNoViewOf(overlayViews, inputHandler.get()));
-                this->inputHandler.reset();
-            }
+        if (drawingType == DRAWING_TYPE_SPLINE) {
+            endSpline();
         }
 
         return true;


bhennion:
Thanks for the PR! However, I don't think this is the right fix.
Things should no go back as high as XournalView. I think the issue should be fixed in XojPageView::onKeyPressEvent() and XojPageView::onKeyReleaseEvent().
We should catch the event after it has been passed to the inputHandler (specialized to a SplineHandler). Basically, if the handler is in finalised state after it has processed the event, delete it.

bhennion (approved):
LGTM.
Thank for the fix!

rolandlo (commented):
I like how it works, but I don't like as much how it is coded. In my opinion
`SplineHandler::onKeyEvent(GdkEventKey* event)` should return true whenever the event has been handled. Same with `XojPageView::onKeyReleaseEvent(GdkEventKey* event)`.
When the event is handled, it is not propagated further, in this case to the rest of `XojPageView::onKeyReleaseEvent(GdkEventKey* event)` and `XournalView::onKeyPressEvent(GdkEventKey* event)`. 

The current way runs into problems once these handlers handle some more keys.

bhennion:
> I like how it works, but I don't like as much how it is coded. In my opinion `SplineHandler::onKeyEvent(GdkEventKey* event)` should return true whenever the event has been handled. Same with `XojPageView::onKeyReleaseEvent(GdkEventKey* event)`. When the event is handled, it is not propagated further, in this case to the rest of `XojPageView::onKeyReleaseEvent(GdkEventKey* event)` and `XournalView::onKeyPressEvent(GdkEventKey* event)`.
> 
> The current way runs into problems once these handlers handle some more keys.

I agree with that, but:

1. The problem is already present without those changes (also for TextEditor and maybe other handlers as well)
2. I think changing this would require a deeper refactor, more suitable for after 1.2 is released.

rolandlo:
Suggested edit:
```diff
diff --git a/src/core/control/tools/SplineHandler.cpp b/src/core/control/tools/SplineHandler.cpp
index 04980ae4c..367f47d37 100644
--- a/src/core/control/tools/SplineHandler.cpp
+++ b/src/core/control/tools/SplineHandler.cpp
@@ -60,11 +60,12 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
 
     switch (event->keyval) {
         case GDK_KEY_Escape: {
+            this->endSpline = true;
             return true;
         }
         case GDK_KEY_BackSpace: {
             if (this->knots.size() == 1) {
-                return false;
+                return true;
             }
             this->deleteLastKnotWithTangent();
             assert(!this->knots.empty() && this->knots.size() == this->tangents.size());
@@ -398,3 +399,5 @@ auto SplineHandler::linearizeSpline(const SplineHandler::Data& data) -> std::vec
 
     return result;
 }
+
+auto SplineHandler::getEndSpline() const -> bool { return this->endSpline; }
diff --git a/src/core/control/tools/SplineHandler.h b/src/core/control/tools/SplineHandler.h
index 766ce05b5..90399d6be 100644
--- a/src/core/control/tools/SplineHandler.h
+++ b/src/core/control/tools/SplineHandler.h
@@ -94,6 +94,7 @@ public:
     Range computeTotalRepaintRange(const Data& data, double strokeWidth) const;
 
     Range computeLastSegmentRepaintRange() const;
+    bool getEndSpline() const;
 
 private:
     void addKnot(const Point& p);
@@ -120,6 +121,7 @@ private:
 
     bool isButtonPressed = false;
     bool inFirstKnotAttractionZone = false;
+    bool endSpline = false;
     SnapToGridInputHandler snappingHandler;
 
     std::shared_ptr<xoj::util::DispatchPool<xoj::view::SplineToolView>> viewPool;
diff --git a/src/core/gui/PageView.cpp b/src/core/gui/PageView.cpp
index f5b95c899..2d8e231a5 100644
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -704,7 +704,8 @@ auto XojPageView::onKeyReleaseEvent(GdkEventKey* event) -> bool {
 
     if (this->inputHandler && this->inputHandler->onKeyEvent(event)) {
         DrawingType drawingType = this->xournal->getControl()->getToolHandler()->getDrawingType();
-        if (drawingType == DRAWING_TYPE_SPLINE) {
+        auto* h = dynamic_cast<SplineHandler*>(this->inputHandler.get());
+        if (drawingType == DRAWING_TYPE_SPLINE && h && h->getEndSpline()) {
             endSpline();
         }
 

```

rolandlo:
> I agree with that, but:
> 
>     1. The problem is already present without those changes (also for TextEditor and maybe other handlers as well)
> 
>     2. I think changing this would require a deeper refactor, more suitable for after 1.2 is released.

I don't see right away how TextEditor has the same problem. Can we fix the issue as suggested above?

bhennion:
How about this simple patch? (diff from master)
```diff
diff --git a/src/core/control/tools/SplineHandler.cpp b/src/core/control/tools/SplineHandler.cpp
index f7e5ade13..9cfbad967 100644
--- a/src/core/control/tools/SplineHandler.cpp
+++ b/src/core/control/tools/SplineHandler.cpp
@@ -50,8 +50,8 @@ constexpr double MIN_TANGENT_LENGTH = 1.0;
 
 auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
     if (!stroke ||
-        (event->type != GDK_KEY_PRESS && event->keyval != GDK_KEY_Escape)) {  // except for escape key only act on key
-                                                                              // press event, not on key release event
+        ((event->type != GDK_KEY_PRESS) == (event->keyval != GDK_KEY_Escape))) {  // except for escape key only act on key
+                                                                                  // press event, not on key release event
         return false;
     }
 
@@ -65,7 +65,6 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
         }
         case GDK_KEY_BackSpace: {
             if (this->knots.size() == 1) {
-                this->finalizeSpline();
                 return true;
             }
             this->deleteLastKnotWithTangent();
```

rolandlo:
> How about this simple patch? (diff from master)
> 
> ```diff
> diff --git a/src/core/control/tools/SplineHandler.cpp b/src/core/control/tools/SplineHandler.cpp
> index f7e5ade13..9cfbad967 100644
> --- a/src/core/control/tools/SplineHandler.cpp
> +++ b/src/core/control/tools/SplineHandler.cpp
> @@ -50,8 +50,8 @@ constexpr double MIN_TANGENT_LENGTH = 1.0;
>  
>  auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
>      if (!stroke ||
> -        (event->type != GDK_KEY_PRESS && event->keyval != GDK_KEY_Escape)) {  // except for escape key only act on key
> -                                                                              // press event, not on key release event
> +        ((event->type != GDK_KEY_PRESS) == (event->keyval != GDK_KEY_Escape))) {  // except for escape key only act on key
> +                                                                                  // press event, not on key release event
>          return false;
>      }
>  
> @@ -65,7 +65,6 @@ auto SplineHandler::onKeyEvent(GdkEventKey* event) -> bool {
>          }
>          case GDK_KEY_BackSpace: {
>              if (this->knots.size() == 1) {
> -                this->finalizeSpline();
>                  return true;
>              }
>              this->deleteLastKnotWithTangent();
> ```

Yes, that one is better and simpler. Works fine!

TrevorShelton:
Seems good, then!

bhennion:
I made a PR #4847 (co-authoring you both) with this patch then. Closing this one.
Thanks for the fix!

PR #4832, opened 2023-04-25
fix the casing of the parameters of getStrokes 

atticus-sullivan:
the type parameter of `applib_getStrokes` was documented as `layer` or `selection` but the implementation used `*L*ayer` and `selection` (see [#4528 (Comment)](https://github.com/xournalpp/xournalpp/pull/4528#issuecomment-1519088312)).
This PR fixes this inconsistency.


rolandlo (approved):
Thanks for the fix. LGTM

PR #4837, opened 2023-04-25
Lua api refactor stack

atticus-sullivan:
As already written in https://github.com/xournalpp/xournalpp/pull/4652#discussion_r1133074071 I started refactoring the complete `luapi_application.h` regarding the use of the lua-stack. The main things I found were

1. function returns `!= 0` while effectively no value is supposed to be returned (at least I think so) (for example [`applib_scaleTextElements`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1896), [`applib_setBackgroundName`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1851), [`applib_uiActionSelected`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L369))

> To return values to Lua, a C [...] returns in C the number of results.

When we return 1 despite we do not deliberately return any lua value, this should simply return the top-most element on the lua-stack which in effect most of the time is the last value passed as an argument (idk what happens if there is no argument / no element on the stack). Thus this should do no harm (in most of the cases) but it's also not really nice.

2. pop directly before returning (for example [`applib_addSplines`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L665), [`applib_changeToolColor`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1125))

> Any other value in the stack below the results will be properly discarded by Lua.

So I think we can simply skip the `pop`s before returning.

3. Some other `pop`s might be removed which are not right before a `return` statement. Examples would be [L168](https://github.com/atticus-sullivan/xournalpp/blob/c1d9b383648d4224afcce8723a1dc01dc478b881/src/core/plugin/luapi_application.h#L168) or [L279](https://github.com/atticus-sullivan/xournalpp/blob/c1d9b383648d4224afcce8723a1dc01dc478b881/src/core/plugin/luapi_application.h#L279). They could be removed because everything then the stack works just like now its just a bit "higher" and when returning, the stack is cleaned up anyhow. Nevertheless I'm not sure since in general the goal should be to keep the stack as small as possible (reducing the memory footprint of the lua-stack). I think this is a matter of taste (while I'd consider 2. almost a must, as `pop` directly above `return` is just unnecessary).

4. Adding comments about the state of the stack
 
5. Oftentimes I saw the sequence of `lua_pushliteral(); lua_push{string,...}(); lua_settable();`. There exists the function `lua_setfield(L, index, "key");` with which we can omit pushing the literal key in advance (if it's a string). This should avoid one operation on the stack (not sure if it is also made internally in this case tbh) and I personally like this a bit more as it it more concise and better to read.

6. In `applib_getToolInfo` I found some additional lua tables get created. While this didn't intervene with the functionality right now, this might have been a candidate for bugs when changing things later (and it's not nice unnecessarily creating tables).

---

Source of the quotes are the [lua docs on `CFunction`s](https://www.lua.org/manual/5.4/manual.html#lua_CFunction) (I've linked the 5.4 docs but this is the same in 5.3 and 5.2)

---
1, 2, 4, 5, 6 are already implemented (mostly in different commits so that we can revert them if we don't like one change).
So let's discuss, what are your opinions? (1, 2, 4, 6 should be pretty clear I think, but the rest also contains decisions on style)

## TODOs
- Code formatting: I've forgotten this in the first place, and now it's difficult to do so while keeping the commits made so far reversible. So this has to be done in the end / after we decided whether to keep the commits made so far.

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -85,11 +85,13 @@ static int applib_glib_rename(lua_State* L) {
 
     g_file_move(from.get(), to.get(), G_FILE_COPY_OVERWRITE, nullptr, nullptr, nullptr, &err);
     if (err) {
+        // return nil, error message
         lua_pushnil(L);
         lua_pushfstring(L, "%s (error code: %d)", err->message, err->code);
         g_error_free(err);
         return 2;
     } else {
+        // return 1
         lua_pushinteger(L, 1);
         return 1;
     }
@@ -197,7 +199,7 @@ static int applib_msgbox(lua_State* L) {
     lua_settop(L, 2);
     luaL_checktype(L, 2, LUA_TTABLE);
 
-    lua_pushnil(L);
+    lua_pushnil(L);  // initial key for table traversal with `next`
 
     std::map<int, std::string> button;
 
@@ -338,10 +340,7 @@ static int applib_uiAction(lua_State* L) {
     Control* ctrl = plugin->getControl();
     ctrl->actionPerformed(action, group, toolbutton, enabled);
 
-    // Make sure to remove all vars which are put to the stack before!
-    lua_pop(L, 3);
-
-    return 1;
+    return 0;
 }
 
 /**
@@ -363,7 +362,7 @@ static int applib_uiActionSelected(lua_State* L) {
     Control* ctrl = plugin->getControl();
     ctrl->fireActionSelected(group, action);
 
-    return 1;
+    return 0;
 }
 
 /**
@@ -394,7 +393,7 @@ static int applib_sidebarAction(lua_State* L) {
     SidebarToolbar* toolbar = plugin->getControl()->getSidebar()->getToolbar();
     toolbar->runAction(pos->second);
 
-    return 1;
+    return 0;
 }
 
 /**
@@ -411,7 +410,7 @@ static int applib_layerAction(lua_State* L) {
     Control* ctrl = plugin->getControl();
     ctrl->getLayerController()->actionPerformed(action);
 
-    return 1;
+    return 0;
 }
 
 /**
@@ -468,6 +467,13 @@ static void addStrokeHelper(lua_State* L, Stroke* stroke) {
     lua_getfield(L, -3, "color");
     lua_getfield(L, -4, "fill");
     lua_getfield(L, -5, "lineStyle");
+    // stack now has following:
+    //   -6 = table
+    //   -5 = tool
+    //   -4 = width
+    //   -3 = color
+    //   -2 = fill
+    //   -1 = lineStyle
 
     tool = luaL_optstring(L, -5, "");  // We're gonna need the tool type.
 
@@ -531,6 +537,7 @@ static void addStrokeHelper(lua_State* L, Stroke* stroke) {
     else
         stroke->setLineStyle(StrokeStyle::parseStyle(lineStyle.data()));
 
+    // stack cleanup is needed as this is a helper function
     lua_pop(L, 5);  // Finally done with all that Lua data.
 
     // Add the stroke
@@ -596,26 +603,30 @@ static int applib_addSplines(lua_State* L) {
     if (!lua_istable(L, -1))
         return luaL_error(L, "Missing spline table!");
 
+    // stack now has following:
+    //  1 = table arg
+    // -1 = splines
+
     size_t numSplines = lua_rawlen(L, -1);
     for (size_t a = 1; a <= numSplines; a++) {
         std::vector<double> coordStream;
         Stroke* stroke = new Stroke();
         // Get coordinates
         lua_pushnumber(L, a);
-        lua_gettable(L, -2);
-        lua_getfield(L, -1, "coordinates");
+        lua_gettable(L, -2);                 // get current spline from splines table
+        lua_getfield(L, -1, "coordinates");  // get coordinates of the current spline
         if (!lua_istable(L, -1))
             return luaL_error(L, "Missing coordinate table!");
         size_t numCoords = lua_rawlen(L, -1);
         for (size_t b = 1; b <= numCoords; b++) {
             lua_pushnumber(L, b);
-            lua_gettable(L, -2);
+            lua_gettable(L, -2);  // get current coordinate from coordinates
             double point = lua_tonumber(L, -1);
             coordStream.push_back(point);  // Each segment is going to have multiples of 8 points.
-            lua_pop(L, 1);
+            lua_pop(L, 1);                 // cleanup fetched coordinate
         }
         // pop value + copy of key, leaving original key
-        lua_pop(L, 1);
+        lua_pop(L, 1);  // cleanup coordinates table
         // Handle those points
         // Check if the list is divisible by 8.
         if (coordStream.size() % 8 != 0)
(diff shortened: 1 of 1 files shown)


rolandlo:
@atticus-sullivan Can you please rebase on current master? I will take a look at it then.

rolandlo:
Sorry for the long wait. I really hoped to review this PR much sooner, but then other things got into the way and in the end we had to concentrate on releasing version 1.2
I will first answer in general terms and go through the changes afterwards.

>     1. function returns `!= 0` while effectively no value is supposed to be returned (at least I think so) (for example [`applib_scaleTextElements`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1896), [`applib_setBackgroundName`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1851), [`applib_uiActionSelected`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L369))
> 
> 
> > To return values to Lua, a C [...] returns in C the number of results.
> 
I agree that this should strictly be followed.
> 
>     2. pop directly before returning (for example [`applib_addSplines`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L665), [`applib_changeToolColor`](https://github.com/atticus-sullivan/xournalpp/blob/949edbe11a9f58e71f14e2735de5ca6a541e865d/src/core/plugin/luapi_application.h#L1125))
> 
> 
> > Any other value in the stack below the results will be properly discarded by Lua.
> 
> So I think we can simply skip the `pop`s before returning.
> 
Fine, let's skip the pop in that case.

>     3. Some other `pop`s might be removed which are not right before a `return` statement. Examples would be [L168](https://github.com/atticus-sullivan/xournalpp/blob/c1d9b383648d4224afcce8723a1dc01dc478b881/src/core/plugin/luapi_application.h#L168) or [L279](https://github.com/atticus-sullivan/xournalpp/blob/c1d9b383648d4224afcce8723a1dc01dc478b881/src/core/plugin/luapi_application.h#L279). They could be removed because everything then the stack works just like now its just a bit "higher" and when returning, the stack is cleaned up anyhow. Nevertheless I'm not sure since in general the goal should be to keep the stack as small as possible (reducing the memory footprint of the lua-stack). I think this is a matter of taste (while I'd consider 2. almost a must, as `pop` directly above `return` is just unnecessary).

Let's pop in that case, since it may lead to bugs otherwise when someone modifies the function.
> 
>     4. Adding comments about the state of the stack
> 
>     5. Oftentimes I saw the sequence of `lua_pushliteral(); lua_push{string,...}(); lua_settable();`. There exists the function `lua_setfield(L, index, "key");` with which we can omit pushing the literal key in advance (if it's a string). This should avoid one operation on the stack (not sure if it is also made internally in this case tbh) and I personally like this a bit more as it it more concise and better to read.
> 
I agree this is better (makes it more readable).

>     6. In `applib_getToolInfo` I found some additional lua tables get created. While this didn't intervene with the functionality right now, this might have been a candidate for bugs when changing things later (and it's not nice unnecessarily creating tables).
> 
Ok, let's don't artificially create tables.
> 
> Source of the quotes are the [lua docs on `CFunction`s](https://www.lua.org/manual/5.4/manual.html#lua_CFunction) (I've linked the 5.4 docs but this is the same in 5.3 and 5.2)
> 
> 1, 2, 4, 5, 6 are already implemented (mostly in different commits so that we can revert them if we don't like one change). So let's discuss, what are your opinions? (1, 2, 4, 6 should be pretty clear I think, but the rest also contains decisions on style)
> ## TODOs
> 
>     * Code formatting: I've forgotten this in the first place, and now it's difficult to do so while keeping the commits made so far reversible. So this has to be done in the end / after we decided whether to keep the commits made so far.

Ok. Also please don't merge master into the PR. Rebasing on master is the way to go.

src/core/plugin/luapi_application.h
```diff
@@ -1441,35 +1463,26 @@ static int applib_getToolInfo(lua_State* L) {
         Tool& tool = toolHandler->getTool(TOOL_TEXT);
         Color color = tool.getColor();
 
-        lua_newtable(L);
-
-        lua_pushliteral(L, "font");
-        lua_newtable(L);
```
rolandlo, line 1447: It looks to me that you removed one `lua_newtable(L)` too much (needed for the `font` table).

rolandlo (changes requested):
Looks mostly good to me, but some changes are required, see in the code.

src/core/plugin/luapi_application.h
atticus-sullivan: Oh that's pretty important. Thanks for noting. I must have been mixing up `{font={size=0, name="fontname"}, color=0x}` and `{size=0, name="fontname", color=0x}` (leaving in a state where neither the first nor the second one is correctly implemented). Fixed it to the first one (+ clarified in comments) as this is the structure being used before.

atticus-sullivan:
> Sorry for the long wait. I really hoped to review this PR much sooner, but then other things got into the way and in the end we had to concentrate on releasing version 1.2

No worries. At first I wondered as nothing happened after I rebased as requested, but then tbh I have forgotten about this PR for some time myself ^^.
(And there are some amazing new features in version 1.2, so that one is nice too)

> Also please don't merge master into the PR. Rebasing on master is the way to go.

Oh right sorry for messing this up this time. Correcting with the next push.

atticus-sullivan:
What do you think about squashing in the end? All to one commit? (I think this would be fine)
[Just asking because otherwise I'd already start with applying clang-format onto the single commits]

rolandlo:
> What do you think about squashing in the end? All to one commit? (I think this would be fine) [Just asking because otherwise I'd already start with applying clang-format onto the single commits]

Yes, please squash it all into one commit. The fixups looks fine.

rolandlo:
Thanks @atticus-sullivan for this cleanup. It's great to have these things cleared up.

src/core/plugin/luapi_application.h
```diff
@@ -883,7 +905,11 @@ static int applib_getStrokes(lua_State* L) {
     std::vector<Element*> elements = {};
     Control* control = plugin->getControl();
 
-    if (type == "layer") {
+    // Discard any extra arguments passed in
+    lua_settop(L, 1);
+    luaL_checktype(L, 1, LUA_TSTRING);
+
+    if (type == "Layer") {
```
atticus-sullivan, line 912: I introduced this spelling mistake again at this point (probably branched off before #4832 got merged).

atticus-sullivan: Any way to cleanly fix this? (merge the other PR a second time is not possible I think, and changing #4837 isn't possible anymore as well, is it?)

rolandlo: The cleanest way would be to fix this in a separate PR that I can merge right away.

PR #4843, opened 2023-04-26
Ghost page

### LINKED ISSUE #4820: Please improve the "Automatically Add Empty Last Page" feature!
I really appreciate that my proposal in #1343 has been tackled in PR #4311 and immediately tested it. However, I found that the existing options aren't really comfortable. 

- The "when scrolling to end of last page" option leads to unnecessary empty pages if I scroll too far by accident (which happens much more often than I expected). Often many new pages are created in one go.
- The "when drawing on last page" option would always lead to one empty page at the end of a document.

There are several possible approaches:
-  Add a "when scrolling past last non-empty page" option. This way, at least there will be no situation when many empty pages are created. I think this is the easiest thing to implement but it is also not completely solving the problem.
- Automatically remove trailing empty pages on save and export. This would solve the problem completely.
- What is described in https://github.com/xournalpp/xournalpp/issues/1343#issuecomment-510407923 by @Febbe. Basically, similar to the previous option, but with better optics. Probably the hardest to implement.

Currently my alternative is to disable this feature because it leads to PDF exports with empty pages...

**Additional context**
Nothing that comes to mind.

bhennion:
Fix #4820

This PR is a proposal for more UX friendly automated page addition when reaching the end of a non-pdf document:

https://user-images.githubusercontent.com/17818041/234678823-2c88cea2-df3b-4768-86c4-7870779b4d8b.mp4

As you can see in the video: when scrolling to the bottom of the document, a greyed out page (ghost page) is added. This page:

- is not saved/exported/printed nor counted in the total number of pages.
- is unghosted as soon as anything is added to it (stroke or any other element(s)), or when it's size is changed -- but not when we change all the pages backgrounds at once.

The implementation is (amazingly) no very complicated: literally add a page to the document, flag it as a ghost page and ignore ghost pages in most places.
There is a small scrolling issue visible at the end of the video, as well as other bugs.  I'll try to fix them once people have given their opinion on the proposal.

@Photon89 @Technius @rolandlo (and anyone else who might have an opinion): what do you think?

- [ ] Fix sidebar preview
- [ ] Fix scrolling issue
- [ ] Test some more and fix the bugs encountered :-)

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1444,8 +1444,8 @@ void Control::duplicatePage() {
     insertPage(pageCopy, getCurrentPageNo() + 1);
 }
 
-void Control::insertNewPage(size_t position, bool shouldScrollToPage) {
-    pageBackgroundChangeController->insertNewPage(position, shouldScrollToPage);
+void Control::insertNewPage(size_t position, bool ghostPage) {
+    pageBackgroundChangeController->insertNewPage(position, ghostPage);
 }
 
 void Control::appendNewPdfPages() {
@@ -1562,8 +1562,13 @@ void Control::paperFormat() {
 
     if (width > 0) {
         this->doc->lock();
-        Document::setPageSize(page, width, height);
+        page->setSize(width, height);
+        bool wasGhost = page->unghost();
         this->doc->unlock();
+
+        if (wasGhost) {
+            page->firePageUnghosted();
+        }
     }
 
     size_t pageNo = doc->indexOf(page);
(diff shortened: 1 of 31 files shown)


rolandlo:
At first glance I like it. It does look user friendly indeed. Usually I annotate PDF's though, so I'm not affected much by the issue.

Photon89:
This is awesome!! I'll test it asap!

Technius:
This is very coo! But my gut feeling is that we may run into subtle bugs here...

Photon89:
So far I couldn't see any other issues related to ghost pages than what is displayed in the video. One thing I saw is that the page shown on opening is not displayed properly (space right from the page is white, as if the page would extend to the right), but most probably it is unrelated to the ghost page code. It gets fixed automatically when scrolling.

anusser:
This feature looks amazing! :)

Not sure whether it's the right place (otherwise, please ignore): Would this PR easily enable an "A4 width page with infinite height"? Basically, a page where one can infinitely scroll vertically and, internally, pages get added automatically (and maybe having some horizontal lines that show the page breaks if one exports to A4 pdf) when writing.

I'm just bringing this up because it seems a quite powerful feature that might be relatively easy to build upon the feature of this pull request.

Thank you so much to everybody working on Xournal++!

PR #4848, opened 2023-04-29
Improve CrashLog

bhennion:
This PR adds useful information to the CrashLog, to help with bug reports.

The added pieces of information are:

1. Xournal++ and Gtk versions
2. Every log message issued via g_message, g_critical, g_debug and so on

Implementing 1. was pretty straghtforward. Implementing 2. meant using g_log_set_default_handler() to record every such message.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ add_subdirectory(po)
 
 # Development options
 option(DEV_CALL_LOG "Call log" OFF)
+add_compile_definitions(G_LOG_DOMAIN="xopp")
 
 # Debug options
 option(DEBUG_INPUT "Input debugging, e.g. eraser events etc" OFF)
--- a/src/core/control/CrashHandler.cpp
+++ b/src/core/control/CrashHandler.cpp
@@ -1,5 +1,6 @@
 #include "CrashHandler.h"
 
+#include <sstream>
 #include <string>  // for string
 
 #include <glib.h>  // for g_warning, g_error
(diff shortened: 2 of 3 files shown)


Technius:
At a first glance, I think it might be better to have the assert changes in a separate PR.

bhennion:
> At a first glance, I think it might be better to have the assert changes in a separate PR.

Done. The assert business is now in #4849. I also updated the description message.

bhennion:
I have a question about the CrashHandler: is there a reason why catching std::signals does not work on Windows? Or was it simply that we do not have a portable StackTrace for Windows?

rolandlo:
> I have a question about the CrashHandler: is there a reason why we do not catch the std::signals on Windows? Or was it simply that we do not have a portable StackTrace for Windows, so nothing to fill the crashlog with?

This distinction between Windows and Unix systems was introduced in #513. No reasons are provided there. I will check what goes wrong on Windows with the Unix implementation.

Febbe:
"we do not have a portable StackTrace for Windows"

The MSVC STL implements the c++23 stacktrace library. And gcc12 implements it partially. There are also plenty of custom implementations like `boost::stacktrace` out there.

We may want to switch to c++23 on windows, or use boost::stacktrace / another lib.

Febbe:
I've created a project for stacktraces, which uses the same API as the c++23 stacktrace library. It should run on all common systems.
https://github.com/Febbe/stacktrace

bhennion:
Should we merge this?

rolandlo:
> Should we merge this?

I think so. It looks good to me.

PR #4849, opened 2023-04-29
Custom assert for redirecting to crashlog

bhennion:
In order to redirect the assertion failing information to the crashlogs, we need cusom assert functions: neither the outputs from std's `assert()` nor glib's `g_assert()` can be redirected. I thus implemented a custom `xoj_assert()` and a custom `xoj_assert_message()` using only standard defined macros (`__FILE__`, `__LINE__` and `__func__`).

Replacing every assertion in the codebase by those custom macros touched a significant number of files.

This needs testing on various compilers and platforms (I only test gcc/linux so far).

~~This is based on #4848~~ Edit: merged since


src/util/include/util/Assert.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <string>
+
+#ifdef NDEBUG
+#define xoj_assert(expr) static_cast<void>(0)
+#define xoj_assert(expr, msg) static_cast<void>(0)
+#else
+namespace xoj::util {
+void assertFailure(const char* expr, const std::string& msg, const char* fileName, int line, const char* funcName);
```
Febbe, line 20: Either inline the function into the header, or mark it as `[[noreturn]]` since the compiler should know, that the program terminates here. (std::abort also has the attribute, but other compilation units can't know if it is not inlined / marked explicit)



```suggestion
[[noreturn]] void assertFailure(const char* expr, const std::string& msg, const char* fileName, int line, const char* funcName);
```

rolandlo:
It works great on MacOS compiled with AppleClang 12.0.0.

bhennion:
Should we merge this?

rolandlo (commented):
Codewise this looks good to me. Let me quickly test it on Linux and Windows. I have tested on MacOS already.

rolandlo (approved):
It works well on Linux, Windows and MacOS as far as I have tested. Please go ahead and merge.

PR #4854, opened 2023-05-03
Add GUI for setting default Pdf export name

### LINKED ISSUE #2897: Please implement an option for a default PDF export file name including wildcards

bhennion:
Fix #2897 

Screenshot_20230503_120109

### DIFF
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -131,29 +131,30 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
 }
 
 auto Document::createSaveFilename(DocumentType type, const std::string& defaultSaveName, const std::string& defaultPdfName) -> fs::path {
-    if (!filepath.empty()) {
-        // This can be any extension
-        fs::path p = filepath.filename();
-        Util::clearExtensions(p);
-        return p;
-    }
-    if (!pdfFilepath.empty()) {
-        if (type != Document::PDF) {
+    std::string wildcardString;
+    if (type != Document::PDF) {
+        if (!filepath.empty()) {
+            // This can be any extension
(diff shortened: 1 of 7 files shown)


src/core/model/Document.cpp
```diff
@@ -131,29 +131,33 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
...
         }
-        auto saveString = SaveNameUtils::parseFilenameFromWildcardString(defaultPdfName, this->pdfFilepath.filename());
-        if (!this->attachPdf) {
-            saveString += ".pdf";
+    } else {
+        if (!pdfFilepath.empty()) {
+            wildcardString =
+                    SaveNameUtils::parseFilenameFromWildcardString(defaultPdfName, this->pdfFilepath.filename());
+        } else if (!filepath.empty()) {
+            wildcardString = SaveNameUtils::parseFilenameFromWildcardString(defaultPdfName, this->filepath.filename());
```
Technius, line 152: nit: this can be flattened into the top level if

rolandlo (approved):
LGTM

bhennion:

Issue #4529, opened 2022-12-26
[lua API] work with pictures (add and possibly get them as well)


atticus-sullivan:
We already have `app.getStrokes` and `app.addStrokes`, in my opinion, these are the most important ones. But there are other use cases, where it would be nice to work with pictures (and text/tex -> different issue) as well.

My use case is to kinda diff beamer presentations. I've thought of
1. backupping the pdf
2. rerun latex
3. convert the pages of the backup to images with a specific transparency (e.g. with imagemagicks `convert`)
4. xournalpp -> add these transparent images on the corresponding page (new layer)

With the rerun of latex and backupping the pdf we might not be able to fully automate this (just maybe provide some defaults), but the essential aspect missing is the ability of adding images via the api.

Just to make it complete, it would be nice to have (like with `addStrokes`) also the corresponding inverse function (like with `getStrokes`).

Note: I'm willing to work on this, I've just not the time right now, so if someone else wants to... feel free (just write here, so that there's no duplicated work on this). One point to start with possibly is `src/core/gui/toolbarMenubar/ToolMenuHandler.cpp:476` where the image-tool is registered (starting to find the function how images usually are added manually with the image-tool)

rolandlo:
@atticus-sullivan In what form (datatype) do you want to pass the images around between C++ and Lua? As a string?

atticus-sullivan:
Hm, initially I thought of passing paths to files. The downside of this is that this won't work with clipboard stuff, and more bad is that this way the `getStrokes` is impossible.
I think I've seen passing images encoded in base64 if I remember correctly, so maybe this would be an option (-> yes this would be strings, I guess).
The cons on this solution is that we've got some overhead in handling the strings (due to two times the memory allocation on both (lua and cpp) sides and string interning on the lua-side). I'm just not quite sure how this turns out on the performance side.

So maybe it's worth to implement both (`addImages` with base64 encoded image (as reverse of `getImages`, and in general for images not being stored in the filesystem, e.g. for clipboard like stuff) and `addImages` with the path to a png/jpg/... file).

rolandlo:
[...]

PR #4858, opened 2023-05-04
Lua api add images from filepaths

atticus-sullivan:
Adds the Lua function `app.addImagesFromFilepath`. The function allows to add images to the document by specifying a path in the filesystem (required) and the desired position on the current page (optional).

This is the first step discussed in https://github.com/xournalpp/xournalpp/issues/4529

TODOs/discussion:
- [x] discuss if we should check if the given file exists (`g_file_query_exists`). The only difference if we do is that we could return a more specifc error(Msg), not just "Inserting the image failed.".
Of course the file might get deleted after our check but that should be no problem as we don't rely on this check, it's only to be able to return a more specific error.
- [x] is there a better way to get to the XojPageView than `control->getWindow()->getXournal()->getViewFor(control->getCurrentPageNo());`?
- [x] should we add any additional features (like scaling)?
The reason why I didn't add this in the first place is that with the current implementation re-using `imgHandler.insertImage`, I don't know how to get/identify the image after inserting it to be able to work with it.

### DIFF
--- a/src/core/control/tools/ImageHandler.cpp
+++ b/src/core/control/tools/ImageHandler.cpp
@@ -14,7 +14,6 @@
 #include "gui/MainWindow.h"                 // for MainWindow
 #include "gui/PageView.h"                   // for XojPageView
 #include "gui/XournalView.h"                // for XournalView
-#include "model/Image.h"                    // for Image, Image::NOSIZE
 #include "model/Layer.h"                    // for Layer
 #include "model/PageRef.h"                  // for PageRef
 #include "model/XojPage.h"                  // for XojPage
@@ -40,7 +39,7 @@ auto ImageHandler::insertImage(double x, double y) -> bool {
     return insertImage(file.get(), x, y);
 }
 
-auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
+auto ImageHandler::createImage(GFile* file, double x, double y) -> std::tuple<Image*, int, int> {
     Image* img = nullptr;
     {
         // Load the image data from disk
@@ -49,7 +48,7 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
         gsize length{};
         if (!g_file_load_contents(file, nullptr, &contents, &length, nullptr, &err)) {
             g_error_free(err);
-            return false;
+            return std::make_tuple(nullptr, 0, 0);
         }
 
         img = new Image();
@@ -69,9 +68,30 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
         delete img;
         XojMsgBox::showErrorToUser(this->control->getGtkWindow(),
                                    _("Failed to load image, could not determine image size!"));
-        return false;
+        return std::make_tuple(nullptr, 0, 0);
     }
 
+    return std::make_tuple(img, width, height);
+}
+
+auto ImageHandler::addImageToDocument(Image* img, bool addUndoAction) -> bool {
+    PageRef page = view->getPage();
+
+    page->getSelectedLayer()->addElement(img);
+
+    if (addUndoAction) {
+        control->getUndoRedoHandler()->addUndoAction(
+                std::make_unique<InsertUndoAction>(page, page->getSelectedLayer(), img));
+    }
+
+    view->rerenderElement(img);
+    auto* selection = new EditSelection(control->getUndoRedoHandler(), img, view, page);
+    control->getWindow()->getXournal()->setSelection(selection);
+
+    return true;
+}
+
+void ImageHandler::automaticScaling(Image* img, double x, double y, int width, int height) {
     double zoom = 1;
 
     PageRef page = view->getPage();
@@ -84,15 +104,13 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
 
     img->setWidth(width * zoom);
     img->setHeight(height * zoom);
+}
 
-    page->getSelectedLayer()->addElement(img);
-
-    control->getUndoRedoHandler()->addUndoAction(
-            std::make_unique<InsertUndoAction>(page, page->getSelectedLayer(), img));
-
-    view->rerenderElement(img);
-    auto* selection = new EditSelection(control->getUndoRedoHandler(), img, view, page);
-    control->getWindow()->getXournal()->setSelection(selection);
-
-    return true;
+auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
+    auto [img, width, height] = ImageHandler::createImage(file, x, y);
+    if (!img) {
+        return false;
+    }
+    automaticScaling(img, x, y, width, height);
+    return addImageToDocument(img, true);
 }
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include <climits>
+#include <cmath>  // for rounding
 #include <cstring>
 #include <limits>  // for numeric_limits
 #include <map>
@@ -27,6 +28,7 @@
 #include "control/pagetype/PageTypeHandler.h"
 #include "control/settings/Settings.h"
 #include "control/tools/EditSelection.h"
+#include "control/tools/ImageHandler.h"
 #include "gui/Layout.h"
 #include "gui/MainWindow.h"
 #include "gui/XournalView.h"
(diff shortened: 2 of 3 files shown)


atticus-sullivan:
3. alright. I think the the check of the `imgSize` should be contained in the first part. The the first part contains all image loading stuff and the second part/rest has all the rendering stuff. In the luaAPI we'll need both in the end (actually we need the exact same procedure we just want to be able to modify the image before rendering).
Should I do this refactoring in this PR or make a new one? (well the reason for it is the luaAPI but apart from that, the luaAPI has nothing to do with it)

> The name suggests that you can add multiple images at once. Not sure how important it is (compared to adding multiple strokes), but if you decide not to support multiple images, then please change the name accordingly.

I think typically one would add only few images and the benefit of having one call to insert multiple ones isn't that high here. If we notice this is necessary we are still able to implement this while still being backwards compatible by passing each image as `{path="", x=0, y=0}` varg (the interface doesn't change if one only passes a single one). Only the name would be misleading then.

But as I think of it, why not implement it like that right now. Tbh, the benefit over passing the images via an "array" is more of a aesthetic nature that `addImagesFromFiles{path="", x=0, y=0}` looks more nice than `addImagesFromFiles{{path="", x=0, y=0}}`.

rolandlo:
I suggest you do it in the same PR, but with commits clearly separated.

atticus-sullivan:
Please check if the way I split the `insertImage` is ok. The thought was to put everything which is responsible for creating and automatically scaling the image into the `createImage` function and only put the real adding to the layer/selection etc. into the trailing `addImageToDocument` function.
This way if we decide to scale the image after `createImage` the manual scaling won't get overwritten in `addImageToDocument`.

As a quick solution I decided to return the image, width and height in a `std::tuple` (not a new struct like you requested). But if wanted, we also can build a new struct around this.

Oh and I wasn't sure if there are any preferences on returning a pointer als `NULL` in case of an error, so I decided to return a `succ` boolean indicating an error (as I used a tuple anyhow). This too is a decision we can quickly change.

src/core/control/tools/ImageHandler.cpp
```diff
@@ -96,3 +100,10 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
 
     return true;
 }
+
+auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
+    auto [succ, img, width, height] = ImageHandler::createImage(file, x, y);
+    if (!succ)
+        return false;
```
rolandlo, line 107: Here and elsewhere: According to the [coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions) you should always use curly braces even if the block only has one line.

> Each if, for, while etc. are followed by curly braces, even if it only contains one line of code (because this is a common case for errors, e.g. search for "Apple goto fail bug")

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1977,135 @@ static int applib_openFile(lua_State* L) {
     return 1;
 }
 
+/*
+ // * Adds images from the provided paths on the current page on the current layer.
```
rolandlo, line 1981: The extra `//` look superfluous here.

src/core/control/tools/ImageHandler.cpp
```diff
@@ -85,6 +83,12 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
     img->setWidth(width * zoom);
     img->setHeight(height * zoom);
 
+    return std::make_tuple(true, img, width * zoom, height * zoom);
+}
+
+auto ImageHandler::addImageToDocument(Image* img) -> bool {
```
rolandlo, line 89: Add a boolean parameter `addInsertUndoAction` here, since that may be unwanted (say in the Lua API one may want to use only one such action for a whole bunch of images).

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1977,135 @@ static int applib_openFile(lua_State* L) {
...
+ // *    to preserve the aspect ratio? Is only set to false if the parameter is false, nil leads to the default value of
+ // *    true (default: true)
+ // *  - scale number: factor to apply to the both dimensions in the end after setting width/height (with/without
+ // *    keeping aspectratio) (default: 1)
+ // *
+ // * Returns nothing on success (lua will fill a variable set to the return value with nil) and an errorMsg on failure
+ // * (soft error as an error might be common for this function)
+ // *
+ // * Example: app.addImagesFromFilepath({path="/media/data/myImg.png", x=10, y=20, scale=2})
+ // * multiple images: app.addImagesFromFilepath({path="/media/data/myImg.png"}, {path="/media/data/myImg2.png"})
```
rolandlo, line 2001: Here you use a different format of passing multiple objects (compared with `addStrokes`, `addSplines`). It is convenient, but has the downside that you can't add any parameters (like `allowUndoRedoAction`) that concern all images at once. I would suggest that multiple objects are put into a table like in `addStrokes`.

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1977,135 @@ static int applib_openFile(lua_State* L) {
...
+            lua_pushstring(L, "Error: creating the image failed.");  // soft error
+            return 1;
+        }
+
+        if (widthParam != -1 && heightParam != -1) {
+            width = widthParam;
+            height = heightParam;
+        } else if (widthParam != -1 && heightParam == -1) {
+            if (aspectRatio) {
+                heightParam = static_cast<int>(static_cast<double>(height) / static_cast<double>(width) * widthParam);
```
rolandlo, line 2079: Here and elsewhere: It would make sense to round before casting to an integer.

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1977,135 @@ static int applib_openFile(lua_State* L) {
...
+        if (!g_file_query_exists(file.get(), NULL)) {
+            lua_pushstring(L, "Error: file does not exist.");  // soft error
+            return 1;
+        }
+
+        XojPageView* pv = control->getWindow()->getXournal()->getViewFor(control->getCurrentPageNo());
+        ImageHandler imgHandler(control, pv);
+        auto [succ, img, width, height] = imgHandler.createImage(file.get(), x, y);
+        if (!succ) {
+            lua_pushstring(L, "Error: creating the image failed.");  // soft error
```
rolandlo, line 2070: Here and elsewhere, the user should be informed about which image produced the failure.

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1977,135 @@ static int applib_openFile(lua_State* L) {
...
+        img->setWidth(width);
+        img->setHeight(height);
+
+        succ = imgHandler.addImageToDocument(img);
+        if (!succ) {
+            lua_pushstring(L, "Error: Inserting the image failed.");  // soft error
+            return 1;
+        }
+
+        lua_pop(L, 7);  // pop the params we fetched from the param-table from the stack
```
rolandlo, line 2104: That's not the right number, I think.

rolandlo (changes requested):

src/core/control/tools/ImageHandler.cpp
atticus-sullivan: Alright. I must have read this (and usually this is my code style as well), but I've read code without the curly braces in `luapi_application.h` so often that I thought this would be the convention. What I mean by this is that another refactoring of `luapi_application.h` is needed to conform to the code conventions.

src/core/plugin/luapi_application.h
atticus-sullivan: Alright. `pushstring` does not support a format string so we'd need something like `sprintf`. ~Generally I'd use more (modern cpp formatting with `format`](https://en.cppreference.com/w/cpp/utility/format)~ (only available as of Cpp20, we're using Cpp17).
So we should use [`cstdio`](https://en.cppreference.com/w/cpp/io/c/fprintf), to format a string like `Error: creating the imageNo {indexOfTheImage} failed.` right? (or have you thought of another way of identifying the image? Maybe we could also simply return a value per passed image encoding errors might get confusing as we shouldn't use `nil` for success as this makes iterating over the returned value difficult in lua)
As we're working with lua always the question arises should the index start with `1` like it's common in lua or with `0` common in most programming.

src/core/plugin/luapi_application.h
atticus-sullivan: I think `7` was right, because there are 7 params per image parameters and the way I was accessing the different image parameter tables did work more like a FIFO (working its way bottom up in the stack, not top down) so I was using the index to iterate over the tables.
But the number changed anyhow as the argument passing was changed.

src/core/control/tools/ImageHandler.cpp
```diff
@@ -85,6 +83,12 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
     img->setWidth(width * zoom);
     img->setHeight(height * zoom);
 
+    return std::make_tuple(true, img, width * zoom, height * zoom);
+}
```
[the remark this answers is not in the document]
atticus-sullivan: Ok I see we need to talk/think more about this. You're right doing the automatic scaling right when creating the image is wrong as if the user only sets `width` and `aspectratio=false`, `height` will have the wrong value.

Adding the automatic scaling to the second split function `addImageToDocument` is also wrong because then the image will/might get scaled after we did our manual scaling.

So I think we'd need an additional function which does the automatic scaling. Do you agree on this?

With that it would be logical to simply not use automatic scaling in the api function so that scaling is up to the api-user. But the question arises if this doesn't lead to many troubles where a user inserts an image without any scaling and the image is way too big.
A solution for this issue might be to check if any of `width`, `height`, `scale` was set. If so, disable automatic scaling, if not, enable automatic scaling. Still the user might want to insert the image completely unscaled (neither manually nor automatically) -> add another (bool) parameter for controlling automatic scaling (default to `true`)?

So I see the problem, but I'm not sure how to proceed in this issue.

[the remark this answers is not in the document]
rolandlo: The API user could get the width and image of the picture beforehand e.g. using  the ImageMagick `identify` tool (```identify -format "%wx%h" img.png```) and then decide based on the page size and insertion location. That may be slightly unpractical though.
What about introducing optional `maxWidth` and `maxHeight` parameters? I think no automatic (uncontrollable) scaling should ever be applied through the Lua API.

src/core/plugin/luapi_application.h
rolandlo: I actually thought about adding the path (not the index) to the string. There is `lua_pushfstring` for formatting.

Currently the "soft error" while returning early (before the other images have been processed) doesn't make a lot of sense. It would be better to accumulate these messages and return them at the end or otherwise use a "hard error".

Yes, in Lua the index should start with 1.

rolandlo:
> I guess the same behavior like in `addSplines` regarding `allowUndoRedoAction` should be used (with fixed strings for `grouped`, `individual` and `none`), right? And the question is if we should use the `addUndoAction` param of the `addImageToDocument` if `individual` was chosen or if we want to roll our own thing like in `addSplines`.

Yes, same behaviour. I think we should use `addUndoInsertAction=false` parameter for `addImageToDocument` and add the undo action ourselves in the Lua API in case of a `grouped` undo action.

src/core/control/tools/ImageHandler.cpp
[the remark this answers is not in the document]
atticus-sullivan: - manually get the width and height: I too think that's pretty impractical (especially for plugins which should be portable we'd need to always check if `identify` is available, and lots of `popen` calls are not nice anyhow)
- setting maximums sounds like a nice idea. So it's gonna be something like this
```
autoWidth, autoHeight = calc automatic scaling
ar = keep_aspect_ratio
if manualWidth set {
  width = min(manualWidth, autoWidth)
  if ar {
    set height based on width
  } else {
    height = autoHeight
  }
}
else if manualHeight set {
  height = min(manualHeight, autoHeight)
  if ar {
    set width based on height
  } else {
    width = autoWidth
  }
}
else if both are set {
  height = min(manualHeight, autoHeight)
  width = min(manualWidth, autoWidth)
}
else {
  width = autoWidth
  height = autoheight
}

```
 
The issue here is that if only one dimension is set and the aspect ratio should be kept the other dimension might be too large.

[Not a trivial task, trying to enforce certain dimensions while giving the user as much freedom as possible]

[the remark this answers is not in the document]
rolandlo: The maxWidth and maxHeight settings should take priority.
I would simply calculate like currently, but in the end if width > maxWidth set width = maxWidth (and scale height proportionally if aspect ratio should be preserved) and then if still height > maxHeight set height = maxHeight (and scale width proportionally if aspect ratio should be preserved).

src/core/plugin/luapi_application.h
atticus-sullivan: `lua_pushfstring` oh thanks :+1: 

I'd keep the "soft error" because a file not being present might be pretty common and is quite like the `ressource not available` when we were discussing about "soft/hard errors".

So I think we have two options:
1. return the errors like `return succ, succ, "Error: xyz", succ` (lua code)
2. return them like `return {succ, succ, "Error: xyz", succ}`. In this case we might even be able to eliminate the `true` elements (though we must note that `ipairs` will most probably not applicable anymore then).

I'm not decided yet what to favor, 1. looks a bit more lua idiomatic but 2. has its advantages as well. In the end, we must be clear that with `table.unpack` and `table.pack` these options are not that different.

And we need to decide what `succ` should be. The value common in lua would be `nil`. The drawback of this value is that it breaks `ipairs` in arraylike tables. The value nearest would be `false` but I'm not sure if this value is not a bit confusing at this point.

src/core/control/tools/ImageHandler.cpp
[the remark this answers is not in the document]
atticus-sullivan: Please check 0da200b9f and 9e1a8488b and tell me what you think?
(not sure if it fits in nicely so I keep this one open)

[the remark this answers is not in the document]
rolandlo: Well, 0da200b9f496f6f7eb481585e26c6e8eb7a745bc does some automatic scaling and completely ignores `maxWidth` and `maxHeight`. That can't be right.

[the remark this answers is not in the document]
atticus-sullivan: The trick is that the `width`/`height` passed into the `automaticScaling` function are the dimensions resulting from the manual scaling. Thus, `automaticScaling` will check if the dimensions are too large, if so it scales them down (keeping the current aspect ratio), if not, the dimensions are not being modified.
(At least that were my thoughts and that is what is supposed to happen)

The `maxWidth`/`maxHeight` parameters passed to the api function are already in the calculated `height`/`width` passed to `automaticScaling` (c.f. Lines 2100 - 2118).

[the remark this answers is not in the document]
rolandlo: In lines 2100-2118 I don't see `maxWidth`/`maxHeight` used. Have you pushed all your changes? Please check!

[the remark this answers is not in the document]
atticus-sullivan: Hm, maybe we're talking about different things. To clarify my intention, I've renamed the variables holding the `maxWidth`/`maxHeight` parameter in the code (see 793c7d5c5).
If this is not what you meant, please clarify.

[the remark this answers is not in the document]
rolandlo: Sorry, I should have read the code more carefully. The current implementation looks good. You should always test the API though to see if it really works as expected.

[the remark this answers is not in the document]
atticus-sullivan: No problem, glad that the current state is fine. Overlooking stuff happens to all of us. So far my tests look good (no images that are too big and no modification of images that fit on the page)

src/core/plugin/luapi_application.h
atticus-sullivan: Alright, I decided to go with 1. for now using `succ = nil`. If the user wants a table (which can't be iterated with `ipairs` because of the `nil`), this has to be created in the lua code with `table.pack` or by wrapping with `{app.addImagesFromFilepaths()}`.
See 12e7c8b29

atticus-sullivan:
So from my side this would be it. I hope I'm not forgetting something.

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1978,193 @@ static int applib_openFile(lua_State* L) {
...
+            continue;
+        }
+
+        lua_pop(L, 8);  // pop the params we fetched from the global param-table from the stack
+        lua_pushnil(L);
+    }
+
+    // Check how the user wants to handle undoing
+    lua_getfield(L, 1, "allowUndoRedoAction");
+    const char* allowUndoRedoAction = luaL_optstring(L, -1, "grouped");
```
rolandlo, line 2148: You need to pop this value again. Otherwise it will be on the stack when returning.

```suggestion
    lua_getfield(L, 1, "allowUndoRedoAction");
    const char* allowUndoRedoAction = luaL_optstring(L, -1, "grouped");
    lua_pop(L, 1);
```

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1978,193 @@ static int applib_openFile(lua_State* L) {
...
+ * Parameters per image:
+ *  - path string: filepath to the image (required)
+ *  - x number: x-Coordinate where to place the left upper corner of the image (default: 0)
+ *  - y number: y-Coordinate where to place the left upper corner of the image (default: 0)
+ *  scaling options:
+ *  - maxWidth integer: sets the width of the image in pixles. If that is too big for the page the image gets scaled
+ *    down keeping the aspect-ratio provided by maxWidth/masHeight (default: -1)
+ *  - maxHeight integer: sets the height of the image in pixles. If that is too big for the page the image gets scaled
+ *    down keeping the aspect-ratio provided by maxWidth/masHeight (default: -1)
+ *  - aspectRatio boolean: if only one of width and height is given should the image be scaled in the other dimension
```
rolandlo, line 1998: I think when both `maxWidth` and `maxHeight` are given and `aspectRatio` is true (or nil) then the aspect ratio should be preserved.

E.g. for `maxWidth = 500`, `maxHeight = 400`, `real width = 1000`, `realHeight = 900`, `aspectRatio = true` 
we should get a height of 400 and a width of 444 (not 500)

src/core/control/tools/ImageHandler.cpp
```diff
@@ -84,15 +104,13 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
...
-    control->getWindow()->getXournal()->setSelection(selection);
-
-    return true;
+auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
+    auto [img, width, height] = ImageHandler::createImage(file, x, y);
+    if (!img) {
+        return false;
+    }
+    ImageHandler::automaticScaling(img, x, y, width, height);
+    return ImageHandler::addImageToDocument(img, true);
```
rolandlo, line 115: I don't see the need for scoping here.
```suggestion
    automaticScaling(img, x, y, width, height);
    return addImageToDocument(img, true);
```

rolandlo (commented):
I have tried out your code, which works mostly fine, and have some finer comments. Mostly nitpicks, but some of them are essential.

src/core/plugin/luapi_application.h
atticus-sullivan: Yes you're right. Thanks for noting (cannot believe I didn't note this when testing the function)

src/core/plugin/luapi_application.h
```diff
@@ -1976,6 +1978,193 @@ static int applib_openFile(lua_State* L) {
...
+    } else if (strcmp("individual", allowUndoRedoAction) == 0) {
+        PageRef const& page = control->getCurrentPage();
+        Layer* layer = page->getSelectedLayer();
+        UndoRedoHandler* undo = control->getUndoRedoHandler();
+        for (Element* element: images) undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, element));
+    } else if (strcmp("none", allowUndoRedoAction) == 0)
+        g_warning("Not allowing undo/redo action.");
+    else {
+        return luaL_error(L, "Unrecognized undo/redo option: %s", allowUndoRedoAction);
+    }
```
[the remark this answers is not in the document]
atticus-sullivan: Yes this makes sense.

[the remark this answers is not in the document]
atticus-sullivan: See 7463e36d7 . I'm wondering how you think about introducing some sort of enum (on the c(pp) side) for the `allowUndoRedoAction` parameter so that we don't need to check in the helper function if this parameter is valid (so we don't need to throw the error in this function -> no `lua_State` parameter and `void` return type)

src/core/plugin/luapi_application.h
rolandlo: I disagree. For me it makes sense, since `maxWidth` only means the `width` should not be larger (and will be as large if nothing speaks against it).
Imagine you want to add multiple images in a grid, where the cells have fixed dimension (with some padding between the cells). I want to have the images as large as possible inside the cells but preserving the aspect ratio. Then I simply specify the `maxWidth` = cell width, `maxHeight` = cell height and `aspectRatio = true` and I don't have to worry, for which images the ratio between width and height is larger or smaller than the ratio between width and height of the cells. So let's say the cells have dimensions `500 x 400`. 
An `1000x900` image with would become `444x400` (it fits into the cell and is as large as possible),
an `600x400` image would become `500x333` (it also fits into the cell and is as large as possible).

Of course if all these three values are given the actual `width` and `height` must be <= `maxWidth` and `maxHeight`, respectively but maximal with the extra condition that the aspect ratio is preserved. You can either compute that by first taking care of `width<=maxWidth` and then `height<=maxHeight` or by comparing the aspect ratios `realWidth/realHeight` and `maxWidth/maxHeight`.

src/core/plugin/luapi_application.h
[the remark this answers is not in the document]
atticus-sullivan: Alright, I think for now/this PR we should stick to just pass the string.

src/core/plugin/luapi_application.h
atticus-sullivan: See 4750fb47a, should be fixed by now.

rolandlo (approved):
LGTM

Please squash your commits into reasonable pieces and rebase on current master.

src/core/control/tools/ImageHandler.cpp
```diff
@@ -69,9 +68,30 @@ auto ImageHandler::insertImage(GFile* file, double x, double y) -> bool {
...
+    }
+
+    view->rerenderElement(img);
+    auto* selection = new EditSelection(control->getUndoRedoHandler(), img, view, page);
+    control->getWindow()->getXournal()->setSelection(selection);
+
+    return true;
+}
+
+auto ImageHandler::automaticScaling(Image* img, double x, double y, int width, int height) -> void {
```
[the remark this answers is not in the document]
rolandlo: And please fix the clang format.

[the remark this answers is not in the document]
atticus-sullivan: Oh ok. But we only use non-trailing return types for returns of type `void`, don't we?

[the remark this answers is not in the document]
rolandlo: Correct.

atticus-sullivan:
Done (if the current split doesn't fit, I can undo it if you want).

rolandlo:
> Done (if the current split doesn't fit, I can undo it if you want).

The split looks good.

I suggest to merge this in 24 hours if no objections are raised.

rolandlo:
Thanks @atticus-sullivan for your dedication to work on the Lua API. Good job! I just merged the PR and I'm looking forward to using it.

PR #4861, opened 2023-05-04
Rework TouchInputHandler

### LINKED ISSUE #4850: Don't scroll on >2 fingers touch inputs
I use a handful of 3/4-finger gestures with lisgd to e.g. switch workspaces or manage windows in combination with a touch screen. When performing these xournalpp performs a scroll event, which I would like to avoid.

Ideally xournalpp should ignore inputs when more than two fingers are present until an appropriate handler is implemented for these.

Do you think this behaviour makes sense in general or should the behaviour be configurable perhaps?

hrdl-github:
Stops touch input processing when more than two fingers are active. Fixes #4850. See also #4852.

I've tested this on plasma-wayland and two versions of sway, one supporting touch cancel events (https://github.com/swaywm/sway/commit/3cd7c7178079bf61bb7967690e7767442b369782) and one that does not (https://github.com/swaywm/sway/commit/db7638a0e515635e3aa36fa8bef58f3e44b177d8). I've tested this using both the GDK wayland and the GDK x11 backend as well as with and without internal hand recognition. I've also tested GTK's inertia scroll mode, as this one produces inconsistent TOUCH_{BEGIN,END} events. The old version of sway can get stuck on zoom when touches are cancelled by libinput without this being communicated to the client application without internal hand recognition being active. This can be fixed by putting down two fingers and taking them up again, which is consistent with the previous implementation. This is not fixable: users of older versions of sway should enable internal hand recognition or update to sway > 1.8.1, once released.

This should probably be tested on other platforms, including native X11 platforms, macOS, and Windows.

### DIFF
--- a/src/core/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/core/gui/inputdevices/TouchInputHandler.cpp
@@ -23,10 +23,15 @@ auto TouchInputHandler::handleImpl(InputEvent const& event) -> bool {
     bool zoomGesturesEnabled = inputContext->getSettings()->isZoomGesturesEnabled();
 
     if (event.type == BUTTON_PRESS_EVENT) {
+        if (invalidActive.find(event.sequence) != invalidActive.end()) {
+            g_warning("Missed touch end/cancel event. Resetting touch input handler.");
+            invalidActive.clear();
+        }
         if (invalidActive.empty() && event.sequence != primarySequence && event.sequence != secondarySequence) {
             // All touches are previously valid and we did not miss an end/cancel event for the current touch
             if (primarySequence && secondarySequence) {
                 // All touches become invalid
+                g_debug("All touches become invalid");
                 invalidActive.insert({primarySequence, secondarySequence, event.sequence});
                 primarySequence = secondarySequence = nullptr;
             } else {
@@ -42,6 +47,7 @@ auto TouchInputHandler::handleImpl(InputEvent const& event) -> bool {
             }
         } else {
             invalidActive.insert(event.sequence);
+            g_debug("Add touch as invalid. %d touches are invalid now.", invalidActive.size());
         }
         return true;
     }
@@ -83,6 +89,7 @@ auto TouchInputHandler::handleImpl(InputEvent const& event) -> bool {
             secondarySequence = nullptr;
         } else {
             invalidActive.erase(event.sequence);
+            g_debug("Removing sequence from invalid list, %d inputs remain invalid.", invalidActive.size());
         }
         return true;
     }


rolandlo:
For reference: Here is what stands in the [Gnome HIG](https://developer.gnome.org/hig/guidelines/pointer-touch.html)

> In GNOME, three and four finger gestures are reserved for use by the system, so these should be avoided by apps. This includes both touchpad and touchscreen gestures. Drags from the top and bottom screen edges are also reserved.

src/core/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -22,64 +22,60 @@ TouchInputHandler::TouchInputHandler(InputContext* inputContext): AbstractInputH
...
-
-            // Set sequence data
-            sequenceStart(event);
-
-            // Even if zoom gestures are disabled,
-            // this is still the start of a sequence. Just
-            // don't start zooming.
-            if (zoomGesturesEnabled) {
-                zoomStart();
+        if (invalidActive.empty() && event.sequence != primarySequence && event.sequence != secondarySequence) {
```
bhennion, line 26: With `invalidActive` never being cleared (besides in `unblock()` which does not really matter for this), the touchscreen will become unresponsive if we somehow miss a `BUTTON_RELEASE_EVENT`.
I've seen such events being skipped for pen devices, and I assume some touch drivers might have this defect as well.
Essentially, we cannot assume that the events come in a sequence BUTTON_PRESS - MOTION - ... - BUTTON_RELEASE.
(As a side note: this kind of undue trust in the drivers is present all over our input system and we need to strengthen it)

I'm not quite sure what a good solution might be in this specific case though

src/core/gui/inputdevices/TouchInputHandler.cpp
```diff
@@ -22,64 +22,60 @@ TouchInputHandler::TouchInputHandler(InputContext* inputContext): AbstractInputH
...
-        if (this->primarySequence && this->secondarySequence && zoomGesturesEnabled) {
-            zoomMotion(event);
-        } else if (event.sequence == this->primarySequence) {
+    if (event.type == MOTION_EVENT) {
+        if (primarySequence == event.sequence && !secondarySequence) {
             scrollMotion(event);
-        } else if (this->primarySequence && this->secondarySequence) {
-            sequenceStart(event);
+            return true;
+        } else if (event.sequence && (primarySequence == event.sequence || secondarySequence == event.sequence)) {
```
bhennion, line 53: Maybe adding an assertion check here might be good, just in case...
```suggestion
        } else if (event.sequence && (primarySequence == event.sequence || secondarySequence == event.sequence)) {
        xoj_assert(primarySequence);
```

Also, if event.sequence could really be 0, then we need to check it's not in the first test above as well.

bhennion:
IMO, anything that strengthen our input system is more than welcome, and I would personnally prioritize this PR over gestures.
Without this PR, any touch event that does not come from the first finger is considered coming from the second. This leeds to weird spasms when 3 or more fingers are involved. Fixing that is important!

archisman-panigrahi:
Is there anything that is blocking the merge of this PR? Other PRs like https://github.com/xournalpp/xournalpp/pull/5133 are waiting for this to be merged.

Bvngee:
Apologize for the bump, just came here to say I would love to see this merged primarily to unblock #5133! :pray:

Issue #1083, opened 2019-04-19
Automatically open autosave after crash


markwmuller:
Can we have X++ behave like Vim -- upon opening a file, X++ immediately checks whether an autosave exists for this file. If it does, a dialog is shown to the user, telling the user whether the autosave is newer than the file opened, and then giving the option to:
1. Open the autosave under a new name
2. Open the autosave and replace the old file with it
3. Delete the autosave, keep working on old file. 

At the moment, the user must manually go find the autosave, copy it from the ~/.xournalpp folder into a new folder, etc. This is annoying, and error prone (I've lost work doing this). Ideally, the user never needs to read up where the autosaves are stored, and muck around with hidden files.

Alternatively, adopt the model of libreoffice -- upon opening X++, it detects if it was not closed properly recently, and then attempts to "recover" the document last open. My preference would be more vim-like.

markwmuller:
Interestingly, it seems the feature exists, but only for some crashes (I just saw it!). I'm not sure what the delineating feature is, but the issue I have is that the X++ interface will occasionally stop responding. In this case, I kill the program using `xkill`; this does not seem to give me the nice dialog. 

To me, the behaviour of Libreoffice is preferable, which does try to recover after I use `xkill` on it.

LittleHuba:
That might be a problem in itself. Using `xkill` sends a `SIGTERM` signal which might generate an exception in certain cases / or not. We currently only catch such exceptions and then call the appropriate handler.

If no such exception occurs then this never happens.
The way around this is to actually register a handler for the `SIGTERM` signal and handle it like it was an uncaught exception in the code.

PR #4867, opened 2023-05-08
Removed special linux treatment for crashes

Febbe:
- introduced crash handling on windows
 - add stacktraces on msw, if a pdb file exist.
 
FYI @bhennion @rolandlo 
The problem currently is, that it does not show any symbol on msys+mingw64+ucrt64


src/core/control/CrashHandler.cpp
```diff
@@ -1,27 +1,83 @@
...
+    }
+
+    emergencySave();
+
+    exit(1);
+}
+
+void installCrashHandlers() {
+    std::signal(SIGTERM, crashHandler);
+    std::signal(SIGINT, crashHandler);
```
eldipa, line 71: `SIGTERM` and `SIGINT` were not handled as crashes in the previous version, is this change ok?

Febbe: Actually for sigint no, sigint does not indicate, that the process is in an undetermined/ broken state. So we would require finally a signal handler, which only sets a flag. That flag must be checked to then open a close request.

For the terminate signal I would say yes but that's pretty irrelevant. Mostly any other signal is raised before a thread terminates.

src/core/control/CrashHandler.cpp
```diff
@@ -1,27 +1,83 @@
...
+
+    std::array<char, 128> stime;
+    size_t size = 0;
+    {
+        static constexpr std::array<char, 10> error_log_name{"errorlog."};
+        std::memcpy(stime.data(), error_log_name.data(), error_log_name.size() - 1);
+        assert(stime[8] == '.');
+        size = strftime(&stime[9], stime.size() - error_log_name.size(), "%Y%m%d-%H%M%S", localtime(&curtime));
+        std::memcpy(stime.data() + size + 9, ".log", 4);
+    }
```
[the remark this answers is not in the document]
Febbe: The reason is, that the maximum size is actually determinable, and that we can reduce heap allocations in the signal handler, also I don't know, if IO operations on a string stream are async signal safe. But beside that the handler itself can't be safe, since it writes the error log and does the emergency safe. So probably it doesn't matter. And a cleaner code might be more worth than signal safe and more efficient code. With c++23 we will get spanstream which is what we actually want here.

src/util/Stacktrace.cpp
```diff
@@ -4,13 +4,13 @@
 #include <array>      // for array
 #include <cstdio>     // for fgets, pclose, popen, snprintf, FILE
 #include <iostream>   // for operator<<, basic_ostream, basic_ostream::...
-#include <string>     // for string
+
+#include <fbbe/stacktrace.h>
```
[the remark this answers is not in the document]
Febbe: Yes, I only created the repo for this problem 😅. But will also be useful for other projects which want to switch to c++23 but can't do it currently.

src/core/control/CrashHandler.cpp
eldipa: I did a few experiments and `Xournal++ (1.1.1)` just closes on `sigint` and `sigterm`.

In those cases reporting a crash it may be weird but saving an emergency file it is definitively handy.

It may happen that someone presses `ctrl-c` in the incorrect terminal by mistake and loses all the modifications. Having an emergency doc it will mitigate the damage.

bhennion: See also https://github.com/xournalpp/xournalpp/issues/1083#issuecomment-485690664

Febbe:
No, I think this is a GCC bug. Or the code I took from the gnu STL c++23 has a bug. It might work with clang64 over msys2/mingw64-clang64... . Compiling with msvc might work, but the application can't scroll anymore and is buggy, so that's not an option.

rolandlo:
I must be missing something.

1) When installing `pacman -S clang` and running 
```term
cmake -GNinja -DCMAKE_CXX_COMPILER=clang++.exe -DCMAKE_C_COMPILER=clang.exe -DCMAKE_BUILD_TYPE=Debug ..
``` 
I get the error:
grafik

2) When installing `pacman -S clang` and running 
```term
cmake -GNinja -DCMAKE_CXX_COMPILER=/C/msys64/clang64/bin/clang++.exe -DCMAKE_C_COMPILER=/C/msys64/clang64/bin/clang.exe -DCMAKE_BUILD_TYPE=Debug ..
``` 
I get the error:

grafik

Febbe:
> I must be missing something.
> 
> 1) When installing `pacman -S clang` and running 
> ```term
> cmake -GNinja -DCMAKE_CXX_COMPILER=clang++.exe -DCMAKE_C_COMPILER=clang.exe -DCMAKE_BUILD_TYPE=Debug ..
> ``` 
> I get the error:
> grafik
> 
> 2) When installing `pacman -S clang` and running 
> ```term
> cmake -GNinja -DCMAKE_CXX_COMPILER=/C/msys64/clang64/bin/clang++.exe -DCMAKE_C_COMPILER=/C/msys64/clang64/bin/clang.exe -DCMAKE_BUILD_TYPE=Debug ..
> ``` 
> I get the error:
> 
> grafik
> 

You are using the msys2 packages. They won't work. You must start the clang64 console and then you have to install the mingw-w64-clang-x86_64-xxx packages both for build tools and dependencies.

The best would bey if you uninstall clang and cmake and install [mingw-w64-clang-x86_64-clang](https://packages.msys2.org/package/mingw-w64-clang-x86_64-clang?repo=clang64) and mingw-w64-clang-x86_64-cmake

rolandlo:
Thanks @Febbe for the explanations. I should have the right packages and the right console now. However now It fails when trying to compile `Stacktrace.cpp`:

grafik

Febbe:
> Thanks @Febbe for the explanations. I should have the right packages and the right console now. However now It fails when trying to compile `Stacktrace.cpp`:
> 
> grafik
> 

Oh, I have to take a look at it, seems like, the Toolchain defines `_MINGW64` or another define, which activates that GCC feature. but if you want to test it, you can edit the source in the build dir under _deps, so it does not use that feature.

Febbe:
Ok, I just tested it myself:

Neither the ucrt64 nor clang64 environments will print stack traces. I try to figure out why. We may have to use the MSWindows specific libraries, when it does not work with libbacktrace.

Febbe:
@rolandlo I fixed it in my stacktrace lib, all msys2 environments now use the msvc implementation. But to print meaningful information, we must provide the pdb file. When you now reconfigure the project, it should work also on your site.

rolandlo:
@Febbe Ok, it works now for me, too. See here:
grafik
[errorlog.20230607-173643.log](https://github.com/xournalpp/xournalpp/files/11679448/errorlog.20230607-173643.log)

src/core/control/CrashHandler.cpp
```diff
@@ -1,27 +1,83 @@
...
+    auto const& errorlogPath = Util::getCacheSubfolder(ERRORLOG_DIR) / std::string_view(stime.data(), size + 9 + 4);
+    std::ofstream fp(errorlogPath);
+    if (fp) {}
```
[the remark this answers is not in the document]
Febbe: fp can invalidate itself when an error occurs, so it is not possible, to move all into the first check.

Febbe:
@rolandlo what I have to do, to also add the pdb file into the binary package we provide?

Febbe:
fixes #1083

Febbe:
@rolandlo how can I test the installers?

windows-setup/package.sh
```diff
@@ -20,6 +20,7 @@ mkdir "$setup_dir"/lib
 
 echo "copy installed files"
 (cd ../build && cmake .. -DCMAKE_INSTALL_PREFIX= && DESTDIR=../windows-setup/"$setup_dir" cmake --build . --target install)
+copy xournalpp.pbd "$setup_dir"/bin/xournalpp.pbd
```
rolandlo, line 23: This should be `cp`, right?
```suggestion
cp xournalpp.pbd "$setup_dir"/bin/xournalpp.pbd
```

rolandlo:
> @rolandlo how can I test the installers?

If you write `/azp run create-installers` into the text box you get the installer. Then install and run it. Open the sidebar and switch to layer preview or layer stack preview mode and close the app right after that. Then it will crash. Then open the app again and you should be notified of a crashlog. You will have to open the file manually and see its output. Alternatively you can run the app from one of the MSYS2 consoles, so you see the backtrace immediately when it crashes.

LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NTkvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NTkvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NTkvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NTkvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 22.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NTkvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMi4wNC1wYWNrYWdlcw2/content?format=zip)

rolandlo:
Looks like there were some mistakes.

- The filename extension is `.pdb`, not `.pbd`. 
- It looks like the `.pdb` file was not generated. Maybe we need `-DCMAKE_BUILD_TYPE=Debug` to generate it or use a different solution from https://stackoverflow.com/questions/28178978/how-to-generate-pdb-files-for-release-build-with-cmake-flags.

Febbe:
MHH, that's interesting. we should not build with Debug but RelWidthDebug, because this does not enable asserts and other tests. It also introduces all optimisations while preserving the Debugsymbols.

LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NjUvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NjUvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NjUvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NjUvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 22.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjY2NjUvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMi4wNC1wYWNrYWdlcw2/content?format=zip)

rolandlo:
> How you found out, that the pdb file is not generated?

I thought it would have to be seen from the build logs. But apparently that's not the case when I build locally. You are right that the `.pdb` file is also built locally with the `RelWithDebInfo` build type.
Anyway the `xournalpp.pdb` file was not found in the pipeline current folder. Seems like we are in `windows-setup` folder instead of the `build` folder, since the brackets on the line above mean that the `cd ../build` doesn't have any effect on the next line.

windows-setup/package.sh
```diff
@@ -20,6 +20,7 @@ mkdir "$setup_dir"/lib
 
 echo "copy installed files"
 (cd ../build && cmake .. -DCMAKE_INSTALL_PREFIX= && DESTDIR=../windows-setup/"$setup_dir" cmake --build . --target install)
+cp xournalpp.pdb "$setup_dir"/bin/xournalpp.pdb
```
rolandlo, line 23:
```suggestion
cp ../build/xournalpp.pdb "$setup_dir"/bin/xournalpp.pdb
```

src/core/gui/MainWindow.cpp
```diff
@@ -76,6 +77,14 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
...
 
     // Window handler
+    static constexpr auto interruptCallback = +[](gpointer data) -> gboolean {
+        if (interrupted()) {
+            auto control = static_cast<Control*>(data);
+            control->quit();
+        }
+        return true;
+    };
+    g_idle_add(static_cast<GSourceFunc>(interruptCallback), static_cast<gpointer>(this->control));
```
bhennion, line 87: This does not feel like the right way to catch the SIGINT signal: for instance, it fails if a dialog is open when the signal is received.

bhennion (changes requested):
Very nice work!
I have encountered a couple of issues though (on Linux, with GCC 13)

1. The fbbe::stacktrace dependency gets recompiled (and stored in `build/_deps/fbbe_stacktrace-build/libbacktrace/lib`) every time I call `cmake --build build/ --target install`, even if all targets have already been built.
2. The interrupt signal is not caught properly when a dialog is opened while "interrupting"
3. The stacktrace get less information than before, typically with library functions:
    ```
    Date: Thu Jun  8 22:41:57 2023
    Error: signal 11
       0# crashHandler at /home/benjamin/xournalpp-fork/src/core/control/CrashHandler.cpp:70
       1#      at :0
       2#      at :0
       3# poll at :0
       4#      at :0
       5# g_main_loop_run at :0
       6# gtk_dialog_run at :0
       7# checkForEmergencySave at /home/benjamin/xournalpp-fork/src/core/control/XournalMain.cpp:221
       8# g_closure_invoke at :0
       9#      at :0
      10# g_signal_emit_valist at :0
      11# g_signal_emit at :0
      12# g_application_register at :0
      13#      at :0
      14# g_application_run at :0
      15# XournalMain::run(int, char**) at /home/benjamin/xournalpp-fork/src/core/control/XournalMain.cpp:690
      16#      at :0
      17# __libc_start_main at :0
      18# _start at :0
      19# 
    ```
    The lines `   at :0` are not very instructive. Maybe adding the binary name (e.g. `/usr/lib/libc.so.6`) would help. With our current master's backtrace mecanism, we get something like `[bt]: (1) /usr/lib/libc.so.6(+0x39ab0) [0x7f07f264fab0]`. I doubt we'd ever use the addresses, but the bin name could help the future dev wondering what those `at :0` mean.

src/core/control/CrashHandler.cpp
```diff
@@ -1,27 +1,101 @@
...
+static std::atomic<bool> sigint = false;
+
+static void forceClose(int sig) {
+    g_warning("Force close requested with signal %i", sig);
+    emergencySave();
+    exit(1);
+}
+
+static void gracefullyClose(int sig) {
+    std::signal(SIGINT, gracefullyClose);
```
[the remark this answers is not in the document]
Febbe: Signal handlers are removed, when they are called. In the case of an sigint we always want to handle it. Especially, when the user decides not to close the document.

src/core/gui/MainWindow.cpp
Febbe: What happens and what do you expect? There is also no other correct way I can think of. The problem with our dialogs is, that they block the event loop, which is a deprecated behavior from gtk2. And the close request must be made from the event loop.

bhennion: > What happens and what do you expect?

Nothing happens (besides the terminal outputing `Gracefully close requested with signal 2`).
I'd expect the app to close if the document has already been saved, or the "Save or Discard" dialog to pop up (on top or instead of the previously shown dialog). Or maybe to fallback on a disgraceful closure.

Febbe: Yes, that's kind of expected. When you press Ctrl+c a second time, xournalpp is closed disgracefully since the event loop could not handle the close request (in situations, where the event loop is blocked or the program hangs).

Note, that our possibilities are very limited in a signal handler. Every signal unsave function can introduce UB which is in any kind unwanted, when the user accidentally pressed ctrg+c.

Therefore I only set a flag which is then handled in the event loop.

Your wanted behavior should be introduced, when all dialogs get rid of the run function.

Febbe:
/azp run create-installers

> Very nice work! I have encountered a couple of issues though (on Linux, with GCC 13)
> 
>     1. The fbbe::stacktrace dependency gets recompiled (and stored in `build/_deps/fbbe_stacktrace-build/libbacktrace/lib`) every time I call `cmake --build build/ --target install`, even if all targets have already been built.

Unfortunately, this is not solvable with an old cmake version. At my WSL with a new version, it also does not happen. Seems like they fixed it. But we would have to bump the cmake version, which is undesirable. 

> 
>     2. The interrupt signal is not caught properly when a dialog is opened while "interrupting"

Solved, not an issue, see above.

>     3. The stacktrace get less information than before, typically with library functions:
>        ```
>        Date: Thu Jun  8 22:41:57 2023
>        Error: signal 11
>           0# crashHandler at /home/benjamin/xournalpp-fork/src/core/control/CrashHandler.cpp:70
>           1#      at :0
>           2#      at :0
>           3# poll at :0
>           4#      at :0
>           5# g_main_loop_run at :0
>           6# gtk_dialog_run at :0
>           7# checkForEmergencySave at /home/benjamin/xournalpp-fork/src/core/control/XournalMain.cpp:221
>           8# g_closure_invoke at :0
>           9#      at :0
>          10# g_signal_emit_valist at :0
>          11# g_signal_emit at :0
>          12# g_application_register at :0
>          13#      at :0
>          14# g_application_run at :0
>          15# XournalMain::run(int, char**) at /home/benjamin/xournalpp-fork/src/core/control/XournalMain.cpp:690
>          16#      at :0
>          17# __libc_start_main at :0
>          18# _start at :0
>          19# 
>        ```
>             
>        The lines `   at :0` are not very instructive. Maybe adding the binary name (e.g. `/usr/lib/libc.so.6`) would help. With our current master's backtrace mecanism, we get something like `[bt]: (1) /usr/lib/libc.so.6(+0x39ab0) [0x7f07f264fab0]`. I doubt we'd ever use the addresses, but the bin name could help the future dev wondering what those `at :0` mean.

You must compile xournal with RelWithDebug I think.

Febbe:
@bhennion I did a bit of research, gcc does not include this information in their implementation. I can change it in my fork, but then we must calculate with the fact, that it will disappear, as soon we change to c++23.

rolandlo:
Fine, the .pdb file is now copied and installed correctly (into the `bin` folder). It works.

````term 
Date: Fri Jun 09 19:39:53 2023
Error: signal 11
0> xournalpp!Z13emergencySavev+0xC03
1> xournalpp!gnu_exception_handler+0x162
2> ntdll!_C_specific_handler+0x96
3> ntdll!_chkstk+0x11F
4> ntdll!RtlRaiseException+0x434
5> ntdll!KiUserExceptionDispatcher+0x2E
6> libgtk_3_0!gtk_widget_destroy+0x1D
7> xournalpp!ZN24SidebarPreviewLayerEntryD0Ev+0x1B
8> xournalpp!ZN18SidebarPreviewBaseD2Ev+0x7E
9> xournalpp!ZN20SidebarPreviewLayersD0Ev+0x64
10> xournalpp!ZN7SidebarD2Ev+0x4F
11> xournalpp!ZN7SidebarD0Ev+0xD
12> xournalpp!ZN7ControlD1Ev+0x2B4
13> xournalpp!ZN7ControlD0Ev+0xD
14> xournalpp!ZN11XournalMain3runEiPPc+0x651
15> xournalpp!WinMainCRTStartup+0xFFFFFFFFFFFFFF1 
16> xournalpp!WinMainCRTStartup+0x16
17> KERNEL32!BaseThreadInitThunk+0x14
18> ntdll!RtlUserThreadStart+0x21
````

**Comment by rolandlo** [comment]
It compiles on MacOS, but doesn't work quite like on Linux. For testing I added a code block
```c++
    std::vector<int> a = {1,2,3,4,5};
    for (int i=0; i<100; i++) {
        auto x = a.at(i);
        g_message("x =  %d", x);
    }
```
to `Control::selectDefaultTool()` which results in a crash on both systems, but only gives a helpful errorlog on Linux:

On Linux:
[errorlog.20230810-150045.log](https://github.com/xournalpp/xournalpp/files/12313068/errorlog.20230810-150045.log)

On MacOS: (I will add it separately)


One more things I observed on MacOS: Whenever I do a partial rebuild (let's say after modifying one file and calling `ninja`) the fbbe:stacktrace gets reconfigured (`Performing configure step for 'libbacktrace'`), which takes some time and is annoying therefore.

A third thing I noticed (on Linux): Gracefully close doesn't close the window, when a dialog (say `help>about`) is opened.
```term
^C
** (xournalpp:35229): WARNING **: 15:15:51.993: Gracefully close requested with signal 2
^C
** (xournalpp:35229): WARNING **: 15:15:53.309: Gracefully close requested with signal 2
^C
** (xournalpp:35229): WARNING **: 15:15:55.159: Gracefully close requested with signal 2
```

Febbe:
> It compiles on MacOS, but doesn't work quite like on Linux. For testing I added a code block
> 
> ```c++
>     std::vector<int> a = {1,2,3,4,5};
>     for (int i=0; i<100; i++) {
>         auto x = a.at(i);
>         g_message("x =  %d", x);
>     }
> ```
> 
> to `Control::selectDefaultTool()` which results in a crash on both systems, but only gives a helpful errorlog on Linux:
> 
> On Linux: [errorlog.20230810-150045.log](https://github.com/xournalpp/xournalpp/files/12313068/errorlog.20230810-150045.log)
> 
> On MacOS: [errorlog.20230810-120849.log](https://github.com/xournalpp/xournalpp/files/12313146/errorlog.20230810-120849.log) The console gives slightly more info:

Looks like, that MacOS either does not report all PC's especially before the signal handler or raise. Or the signal handler is executed in a fresh thread. I don't know how to solve this, since I do not have a mac.

> One more things I observed on MacOS: Whenever I do a partial rebuild (let's say after modifying one file and calling `ninja`) the fbbe:stacktrace gets reconfigured (`Performing configure step for 'libbacktrace'`), which takes some time and is annoying therefore.

Without updating CMake, or adding CMake to libbacktrace, I can't do much against that. It's solvable tho, but I have no time to do this.

> A third thing I noticed (on Linux): Gracefully close doesn't close the window, when a dialog (say `help>about`) is opened.
> 
> ```
> ^C
> ** (xournalpp:35229): WARNING **: 15:15:51.993: Gracefully close requested with signal 2
> ^C
> ** (xournalpp:35229): WARNING **: 15:15:53.309: Gracefully close requested with signal 2
> ^C
> ** (xournalpp:35229): WARNING **: 15:15:55.159: Gracefully close requested with signal 2
> ```

This is kind of expected, the reason is, that our dialogs do not return to the event queue and that they block all current tasks.
That is fixed by @bhennion 's gtk4 dialog fixes. But it should force close after pressing it twice.

bhennion:
Closing in favour of #6563

Issue #2757, opened 2021-02-06
Undoing "insert page" moves user to another page if user is on non-new page


nicolae-stroncea:
**Affects versions :**

 - Ubuntu 20.04, Gnome X11
- Xournal++ 1.1.0+dev
- libgtk: 3.24.20
- PPA

**Description**:
As specified in title.

**To Reproduce**
1. Create a new page in the middle of 2 other pages.
2. Move to a page that is not the new page.
3. Click "Undo page".

Results:
* Xournal++ moves you to the next page in the document. This makes sense if you were on the new page, as it doesn't exist and you need to be moved somewhere(either previous or next page, both make sense). If you were on a different page, you should not be moved at all.

**Expected behavior**
User stays on the same page.

p0mm:
I have reproduced this & would like to work on this.
I agree that you should not be moved if you are no longer on the new page that is deleted & hope to implement this behavior!

Technius:
Fixed by #4875.

PR #4875, opened 2023-05-15
Fix for scroll bug when undoing insert page

p0mm:
## Fix for Bug #2757 

In the case that you insert a page and then scroll further down and click undo, you will no longer be scrolled down an additional page. 
This occurred because the index of the current page decreased, since a page was deleted before it.
### Steps to reproduce old buggy behavior:
- have more then one page
- insert a page somewhere (not at the end)
- scroll to a page below the inserted page, that is not the end
- undo insert page

-> you are scrolled one page down / the page below the current page is selected
#### commit 1: Fix for scroll bug when undoing insert Page (fix)
- whatever page you scrolled to before undoing the inserted page is re-selected / scrolled to
- this means you are pulled to the top of whatever page you where viewing, since it is re-selected
- which was the previously intended behavior
#### commit 2: No movement when undoing insert page (improvement?)
- when undoing insert page while on another page your view doesn't change - no automatic scrolling
- (not implemented for paired pages as they shift sideways, no movement seemed more confusing)
- pro: no movement may be less intrusive and more desirable - as the user was probably viewing something on that page
- con: it is good to have some kind of way to recognize that something happened, should you somehow click the undo button without noticing

### DIFF
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -579,12 +579,17 @@ void XournalView::pageChanged(size_t page) {
 }
 
 void XournalView::pageDeleted(size_t page) {
-    size_t currentPage = control->getCurrentPageNo();
+    const size_t currentPageNo = control->getCurrentPageNo();
 
-    viewPages.erase(begin(viewPages) + page);
+    viewPages.erase(begin(viewPages) + static_cast<long>(page));
 
     layoutPages();
-    control->getScrollHandler()->scrollToPage(currentPage);
+
+    if (currentPageNo > page) {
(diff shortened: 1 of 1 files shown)


rolandlo (changes requested):
I approve of the first commit. The second commit only works when the pages are all in one column. If you have all pages in one row, for example, it doesn't work. So let's keep it simple and drop the second commit.

p0mm:
The second commit is dropped!
Thank you for reviewing :)

rolandlo (approved):
Thanks for the fix and for quickly addressing my review!

PR #4887, opened 2023-05-22
Fix missing titlebar icon on wayland.

### LINKED ISSUE #4870: The Wayland logo is being displayed in the top left corner of Xournal++ windows instead of the Xournal++ icon.
**Affects versions :**
 - OS: Arch Linux
 - (Linux only) Desktop environment: KDE Plasma v5.27.5, Wayland
 - Which version of libgtk do you use v3.24.37
 - Version of Xournal++: 1.1.3 2dbd16d2

**Describe the bug**
The Wayland logo is being displayed in the top left corner of all Xournal++ windows

**To Reproduce**
N/A

**Expected behavior**
I expect that the Xournal++ logo would be displayed in the top left corner

**Screenshots of Problem**
image

mariakeating:
Fixes #4870

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -615,6 +615,7 @@ auto XournalMain::run(int argc, char** argv) -> int {
 
     XournalMainPrivate app_data;
     GtkApplication* app = gtk_application_new("com.github.xournalpp.xournalpp", APP_FLAGS);
+    g_set_prgname("com.github.xournalpp.xournalpp");
     g_signal_connect(app, "activate", G_CALLBACK(&on_activate), &app_data);
     g_signal_connect(app, "command-line", G_CALLBACK(&on_command_line), &app_data);
     g_signal_connect(app, "open", G_CALLBACK(&on_open_files), &app_data);
--- a/desktop/com.github.xournalpp.xournalpp.desktop
+++ b/desktop/com.github.xournalpp.xournalpp.desktop
@@ -15,7 +15,7 @@ Keywords=tablet;Wacom;pen input;PDF Annotation;markup
 Keywords[hu]=táblagép;Wacom;tollas bevitel;PDF-megjegyzés;korrektúra
 
 Exec=xournalpp %f
-StartupWMClass=xournalpp
+StartupWMClass=com.github.xournalpp.xournalpp
 Terminal=false
 StartupNotify=true
 MimeType=application/x-xoj;application/x-xojpp;application/x-xopp;application/x-xopt;application/pdf;


rolandlo:
Thanks for addressing this issue. Your PR is also changing the `WM_CLASS` from `xournalpp` to `com.github.xournalpp.xournalpp` as can be observed by hitting `ALT+F2`, executing `lg`  and looking at the `Windows` section.

grafik

So I guess we would also have to change the line `StartupWMClass=xournalpp` in the `.desktop` file accordingly, right?

mariakeating:
I'm not well versed on how .desktop files work on Linux but that seems like it would be correct.

rolandlo (approved):
LGTM. I'm not 100% sure though whether changing the program name like that could have unwanted side effects. So I'd be happy for a second opinion.

rolandlo:
@xournalpp/core Any second opinion on this PR?

Technius:
The changes in this PR seem to make the application satisfy [the requirements listed in the GTK docs](https://docs.gtk.org/gtk4/migrating-3to4.html#set-a-proper-application-id). Not sure how well this works in practice though, since there are always users that run into issues with the icons and the window association.

bhennion:
I think this is ok. Like @rolandlo and @Technius, I'm not 100% sure no problem will arise but that's easily reversable should anything go bad.
@mariakeating Sorry it took so long. Could you rebase on latest master and resolve the conflict so we can merge?

src/core/control/XournalMain.cpp
```diff
@@ -615,6 +615,7 @@ auto XournalMain::run(int argc, char** argv) -> int {
 
     XournalMainPrivate app_data;
     GtkApplication* app = gtk_application_new("com.github.xournalpp.xournalpp", APP_FLAGS);
+    g_set_prgname("com.github.xournalpp.xournalpp");
```
Febbe, line 618: I thought this is called by `gtk_application_new`. 
And since `GLib 2.72, this function can be called multiple times and is fully thread safe. Prior to GLib 2.72, this function could only be called once per process.` Therefore, I would remove it. Can you test if this makes a difference?

bhennion: This line is precsely what fixes the titlebar icon... Removing it makes this PR pointless.
You can test with `g_get_prgname()` that `gtk_application_new()` actually doesn't set the name.

bhennion: If we are a bit scared, we can always test against the GLib version and only fix the bug for up-to-date setups (i.e. glib >= 2.72)

Febbe: We can simply prove this, by either asking first for the name, or by searching the source code of grk3, if that function (`g_set_prgname`) is used somewhere. Also, the documentation for it mentions this: `If you are using GApplication the program name is set in g_application_run(). In case of GDK or GTK+ it is set in gdk_init(), which is called by gtk_init() and the GtkApplication::startup handler. The program name is found by taking the last component of argv[0].`.

So my question here is, why is this function not called, does it only apply to gtk4 or did we miss something else? Or is it set to xournalpp since it is the name of the executable?

bhennion: Indeed, it is set to xournalpp once `g_application_run()` is called (I tested it in `XournalMain::on_startup()`).
With the line `g_set_prgname("com.github.xournalpp.xournalpp");`, the name is set properly, and not reset by `g_application_run()`.

Febbe: I also found this in https://gitlab.gnome.org/GNOME/gtk/-/blob/main/docs/reference/gtk/migrating-3to4.md

> Set a proper application ID
In GTK 4 we want the application's GApplication 'application-id' (and
therefore the D-Bus name), the desktop file basename and Wayland's xdg-shell
app_id to match. In order to achieve this with GTK 3.x call
g_set_prgname() with the same application ID you passed to
[[class@Gtk.Application](mailto:class@Gtk.Application)].  Rename your desktop files to match the application
ID if needed.
The call to g_set_prgname() can be removed once you fully migrated to GTK 4.
You should be aware that changing the application ID makes your
application appear as a new, different app to application installers.
You should consult the appstream documentation for best practices
around renaming applications. 

So I think this PR is correct.

bhennion:
Thanks @mariakeating for the fix!

michalkvasnicka:
Is the problem with missing icon or logo fully solved in latest version 1.2.1, or not?
I have the same problem with flatpak install and same problem with PPA: ppa:apandada1/xournalpp-stable

PR #4903, opened 2023-05-31
Update Arch Linux package info

anh0516:
The Arch Linux [community] repository recently [merged](https://archlinux.org/news/git-migration-announcement/) to [extra]. Update README.md to reflect this change.


rolandlo:
Thanks for the info and the PR. We'll have to update the website as well.

Issue #4936, opened 2023-06-19
can't use the find function in chinese 


domonnss:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: [e.g. Linux (which distro), Mac, Windows] 
 - windows
 - (Linux only) Desktop environment: [e.g. Wayland, Gnome-Wayland, XWayland, X11]
 - Which version of libgtk do you use 
 - 3.24.38
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Version of Xournal++: [e.g. 1.0.16, specific commit, specific branch] 
 - xournalpp-1.1.3+dev-nightly.20230610-windows
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Installation method: [e.g. PPA, Flatpak, official release, etc.]
offical release
**Describe the bug**
if i want to use the find function in chinese,i have to copy the chinese words from other place,and then paste it in the input box.Can't just type it in the box.And in english ,it works good. 
**To Reproduce**
1. Step 1 ctrl + f
2. Step 2 type in chinese 

**Expected behavior**
the words I have tpyed appear in the box
**Screenshots of Problem**
(Optional) If applicable, add screenshots to help explain your problem.

**Additional context**
Add any other information or context about the problem here.

domonnss:
new intersting discovery.After pasting the words in chinese in the find-box,i could use the find function without any problem.I could enter text in chinese even i have deleted all of the pasted words .
![image](https://github.com/xournalpp/xournalpp/assets/60506557/7946e06c-af4f-4407-81dd-213dcc8cf9f4)

TobySkarting:
I can reproduce the same issue with Chinese IME (Zhuyin) on Windows 10.

PR #4937, opened 2023-06-19
Fix search box not able to receive user input when Windows IME is act…

TobySkarting:
Fix https://github.com/xournalpp/xournalpp/issues/4936

It turns out that making the focus before showing the widget will cause IME malfunctioning.


bhennion:
This sounds like a reasonible change. Thanks a lot! Merging right away.

Issue #4183, opened 2022-08-02
Load image with specific size


youngminpark2559:
**Is your feature request related to a problem? Please describe.**
When I load large size images, I should always resize them into small size

**Describe the solution you'd like**
- Memorizing the size of image which was last loaded, and when image is loaded newly, this image should be loaded with the memorized size
- Or, it would be great if I could specify certain area or region. After that I use ctrl+shift+i or drag-and-drop to load image
  Loaded image onto that region should have the specific designated size

dov:
I agree with this. I was just about to open a feature request about something very similar.

Here is my suggested change:

- When in image mode and clicking on the canvas you will **not** get a file select dialog immediately. 
- The mouse click should starts the dragging of an on canvas rectangle. 
- On mouse release When the file select dialog is opened. 
- After selecting the image, it will be rescaled to fit into the dragged rectangle previously chosen and then put on the canvas.

ysngshn:
At least it would be nice to have a **consistent** image insert size based on image resolution and pdf page size/DPI, etc. I am trying to insert the same signature saved as png and depending on where I click, it is inserted in a different size.

p0mm:
I really like the idea with specifying the area for the image before selecting the image -> I would like to implement this!
I also agree that image insert size should be consistent, i have played around with inserting images:

https://github.com/xournalpp/xournalpp/assets/125290879/01cb191c-d737-43ca-9ce5-f70619fdfa8b

While there is reasons for the way xournalpp currently does it (zoom level & how close to the edge one adds an image) it doesn't feel very intuitive and i hope to improve it:
- maybe account for zoom? most users probably aren't working with 100% zoom, so pictures added will always be larger / smaller when they are added with original size for 100% zoom -> this could be changed to original size at current zoom?  
- allow user to grab and move image directly after inserting it (currently you add the same image again? which feels buggy)
- don't make the image smaller just because it was added closer to the edge, let the user decide to move or resize the image

I hope to open a PR soon, please feel welcome to suggest alternative behavior if you have any ideas - I'd be willing to try and implement them as well, so we can choose the best behavior in the end :)

PR #4946, opened 2023-06-28
Enhanced Image Adding & Image Frames

p0mm:
this PR is a response to #4183!

## General Idea:
- define space image should take before choosing the image
- make adding multiple images easier to plan
- possibility to template?

## Plans:
- [ ] define space by dragging the mouse (click - drag - let go)
- [ ] have a nice visual to see what space you are choosing
- [ ] frame for image comes into existence
- [ ] its size can be adjusted
- [ ] it will remain if you save, close and open xournalpp again
- [ ] allow multiple image frames to coexist
- [ ] add image to the frame by clicking it
- [ ] give a visual so this mechanic is easily understandable (show pointer as tiny image icon while hovering over image frame)
- [ ] add image by dragging and dropping it into the frame
- [ ] give different options on whether the image is scaled up, down or stretches and fills the space
- [ ] have a default setting for these options available in the preferences 
- [ ] still keep a simple add image directly option easily available

### DIFF
--- a/src/core/control/tools/ImageHandler.cpp
+++ b/src/core/control/tools/ImageHandler.cpp
@@ -23,21 +23,65 @@
 #include "util/i18n.h"                      // for _
 #include "util/raii/GObjectSPtr.h"          // for GObjectSPtr.h
 
-ImageHandler::ImageHandler(Control* control, XojPageView* view) {
-    this->control = control;
-    this->view = view;
-}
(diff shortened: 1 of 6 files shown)


forgottosave:
Awesome feature, would be great to have this! To actually be able to scale the image before inserting it really saves you the troubles of having a giant image covering the whole page.

PR #4956, opened 2023-07-19
Optionally Display Current Page Number on the Titlebar

VivekThazhathattil:
This feature can be enabled from `Edit>Preferences>View>Show page number in title bar`. It allows the user to see the current page number in the titlebar. This comes in handy when the user has toggled off `View>Show Toolbars`, but wish to see the current page number they are in, without having to go through the trouble of toggling toolbars back on.

MuPDF has it by default, and I thought it would be a nice little feature to incorporate into Xournal++ as well.

### Toggle option:
image
### Example:
image


bhennion (approved):
LGTM! Thanks for the feature!

If everyone is ok with that, I'll be (squashing and) merging this in 24 hours (to the master branch)

PR #4962, opened 2023-07-22
Hurd PATH_MAX

barak:
Define PATH_MAX if undefined, to address Hurd build failure.

Gbp-Pq: Name 0001-Hurd-PATH_MAX.patch

### DIFF
--- a/src/util/Stacktrace.cpp
+++ b/src/util/Stacktrace.cpp
@@ -71,10 +71,10 @@ fs::path Stacktrace::getExePath() {
 #else
 auto Stacktrace::getExePath() -> fs::path {
 #ifndef PATH_MAX
-  // This is because PATH_MAX is (per posix) not defined if there is
-  // no limit, e.g., on GNU Hurd. The "right" workaround is to not use
-  // PATH_MAX, instead stat the link and use stat.st_size for
-  // allocating the buffer.
+    // This is because PATH_MAX is (per posix) not defined if there is
+    // no limit, e.g., on GNU Hurd. The "right" workaround is to not use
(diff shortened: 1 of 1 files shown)


bhennion:
LGTM, except the clang format complain. Could you run `git clang-format` on your commit?

src/util/Stacktrace.cpp
```diff
@@ -71,10 +71,10 @@ fs::path Stacktrace::getExePath() {
...
 auto Stacktrace::getExePath() -> fs::path {
 #ifndef PATH_MAX
-  // This is because PATH_MAX is (per posix) not defined if there is
-  // no limit, e.g., on GNU Hurd. The "right" workaround is to not use
-  // PATH_MAX, instead stat the link and use stat.st_size for
-  // allocating the buffer.
+    // This is because PATH_MAX is (per posix) not defined if there is
+    // no limit, e.g., on GNU Hurd. The "right" workaround is to not use
+    // PATH_MAX, instead stat the link and use stat.st_size for
+    // allocating the buffer.
```
Febbe, line 77: > 
```
    The "right" workaround is to not use
    // PATH_MAX, instead stat the link and use stat.st_size for
    // allocating the buffer.
```
    
That's wrong there is no better workaround, since stat.st_size always returns the size of a file, not its path. The way to go here, is to replace the whole code with https://en.cppreference.com/w/cpp/filesystem/read_symlink

barak: Oops. You're right.

Blech.

Febbe: No problem, I'll fix it with #4867.

PR #4976, opened 2023-07-31
Fix errorlog opening

rolandlo:
This PR fixes the following problems that occur when an errorlog is present:
- The errorlogs directory doesn't open for the flatpak and snap package (since `nautilus`, `dolphin` and `konqueror` cannot be opened directly; rather `xdg-open` should be used)
- The errorlogs directory and errorlogs don't open on Windows (since the `start` command must use a title when there are quotes around the filename)

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -197,7 +197,7 @@ void checkForErrorlog() {
     } else if (response == OPEN_FILE) {
         Util::openFileWithDefaultApplication(errorlogPath);
     } else if (response == OPEN_DIR) {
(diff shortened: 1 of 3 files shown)


rolandlo:
I think bugfixes like this should be merged into release-1.2 rather than into master. Am I correct?

For reference: When rebasing on `release-1.2` in order to avoid merge conflicts execute:
```term
git rebase --onto=release-1.2 master  my-bugfix-branch
```

rolandlo:

Technius:
> I think bugfixes like this should be merged into release-1.2 rather than into master. Am I correct?

Yes, major or breaking changes go into master. Minor changes should be merged into the release branch (which will be regularly merged back into master).

PR #4996, opened 2023-08-02
Install libgtksourceview-4-dev on Azure pipelines

rolandlo:
The release builds on Ubuntu and Debian were compiled without gtksourceview. This PR fixes this problem for future releases.


rolandlo:

PR #5022, opened 2023-08-08
Add age filter on crash logs reported at startup

bhennion:
Limits the crashlogs reported by the popup at startup to those issued during the last week.


rolandlo (approved):
LGTM and works like intended.

bhennion:
Thanks for the review! Merging in 24 hours.

Issue #5025, opened 2023-08-08
Failing assert in StrokeViewHelper::drawWithPressure: assert(p.z > 0)


its-kks:
Date: Tue Aug  8 16:55:20 2023
Error: signal 6
Xournal++ version 1.2.0
Git commit: 10ed72c9
Gtk version 3.24.38

[bt]: (0) xournalpp(+0xe69ba) [0x561eb91a29ba]
[bt]: (1) /usr/lib/x86_64-linux-gnu/libc.so.6(+0x3f0c0) [0x7f70dc23f0c0]
[bt]: (2) /usr/lib/x86_64-linux-gnu/libc.so.6(+0x91184) [0x7f70dc291184]
[bt]: (3) /usr/lib/x86_64-linux-gnu/libc.so.6(raise+0x1e) [0x7f70dc23f00e]
[bt]: (4) /usr/lib/x86_64-linux-gnu/libc.so.6(abort+0xd7) [0x7f70dc2287fc]
[bt]: (5) /usr/lib/x86_64-linux-gnu/libc.so.6(+0x2871b) [0x7f70dc22871b]
[bt]: (6) /usr/lib/x86_64-linux-gnu/libc.so.6(+0x37657) [0x7f70dc237657]
[bt]: (7) xournalpp(+0x238c2b) [0x561eb92f4c2b]
[bt]: (8) xournalpp(_ZNK3xoj4view14StrokeToolView4drawEP6_cairo+0x263) [0x561eb93335c3]
[bt]: (9) xournalpp(_ZN11XojPageView21drawAndDeleteToolViewEPN3xoj4view8ToolViewERK5Range+0x80) [0x561eb929e080]
[bt]: (10) xournalpp(_ZN13StrokeHandler20onButtonReleaseEventERK17PositionInputDatad+0x22b) [0x561eb930e1ab]
[bt]: (11) xournalpp(_ZN11XojPageView20onButtonReleaseEventERK17PositionInputData+0x53) [0x561eb92a1b83]
[bt]: (12) xournalpp(_ZN15PenInputHandler9actionEndERK10InputEvent+0x108) [0x561eb91f2538]
[bt]: (13) xournalpp(_ZN17MouseInputHandler10handleImplERK10InputEvent+0x85) [0x561eb91f1c25]
[bt]: (14) xournalpp(_ZN12InputContext6handleEP9_GdkEvent+0x373) [0x561eb91f0a03]
[...]

PR #5028, opened 2023-08-09
Fix negative pressures #5025

bhennion:


LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjc0MzMvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjc0MzMvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjc0MzMvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjc0MzMvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 22.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjc0MzMvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMi4wNC1wYWNrYWdlcw2/content?format=zip)

Febbe:
I've checkt many code paths by hand and calculated possible ranges. 
I've found one suspicious code path:
In `PenInputHandler::inferPressureIfEnabled`, when distance == 0 and the sqrt(time difference /100) is lesser equal than 0.1 the pressure could be negative or 0. That should be fixed, but I can't tell, if this was the cause. Then the crash could only happen when drawing dots.

Febbe:
```
double PenInputHandler::inferPressureIfEnabled(PositionInputData const& pos, XojPageView* page) {
    if (pos.pressure == Point::NO_PRESSURE && this->inputContext->getSettings()->isPressureGuessingEnabled()) {
        PositionInputData lastPos = getInputDataRelativeToCurrentPage(page, this->lastEvent);
        constexpr uint32_t TO_BE_GREATER_ZERO_OFFSET = 1;
        double dt = (pos.timestamp - lastPos.timestamp + TO_BE_GREATER_ZERO_OFFSET) / 10.0;
        double distance = utl::Point<double>(pos.x, pos.y).distance(utl::Point<double>(lastPos.x, lastPos.y));
        double newPressure = std::max(0.001, [&]() {
            if (distance == 0) {
                // Handle the single-point case.
                return std::sqrt(dt / 10.0) - 0.1;
            } else {
                double inverseSpeed = dt / (distance + 0.001);
                // This doesn't have to be exact. Arctan is used here for its sigmoid-like shape,
                // so that lim inverseSpeed->infinity (newPressure) is some finite value.
                double newPressure = 3.142 / 2.0 + std::atan(inverseSpeed * 3.14 - 1.3);

                // This weighted average both smooths abrupt changes in newPressure caused
                // by changes to inverseSpeed and causes an initial increase in pressure.
                return std::min(newPressure, 2.0) / 5.0 + this->lastPressure * 4.0 / 5.0;
            }
        }());
        this->lastPressure = newPressure;
        // Final pressure tweaks...
        return (newPressure * 1.1 + 0.8) / 2.0;
    }

    return pos.pressure;
}
```

bhennion:
> I've checkt many code paths by hand and calculated possible ranges. I've found one suspicious code path: In `PenInputHandler::inferPressureIfEnabled`, when distance == 0 and the sqrt(time difference /100) is lesser equal than 0.1 the pressure could be negative or 0. That should be fixed, but I can't tell, if this was the cause. Then the crash could only happen when drawing dots.

Except pressure inference was not enabled by @francispoulin, and `PenInputHandler::inferPressureIfEnabled` is only use in the following function, so the std::max should forbid any non-positive values (`settings->getMinimumPressure() == 0.05` by default, and for @francispoulin as well)
```
double PenInputHandler::filterPressure(PositionInputData const& pos, XojPageView* page) {
    double filteredPressure = inferPressureIfEnabled(pos, page);
    Settings* settings = this->inputContext->getSettings();

    if (filteredPressure != Point::NO_PRESSURE) {
        filteredPressure *= settings->getPressureMultiplier();
        filteredPressure = std::max(settings->getMinimumPressure(), filteredPressure);
    }

    return filteredPressure;
}
```

Febbe:
Another possibility is, an event, which somehow does not have the Pressure axis but the previous and the next do have them, but all are fed to StrokeHandler/PenInputHandler.

I also think, it is badly designed, that a Stroke with Pressure can be fed with NO_PRESSURE and vice versa. Points should be split from their Pressure, and the pressure should be optional.

Febbe:
In `InputEvents::translateEvent` we get the pressure from the axis: 
I would add this to test the device:
```
    // Copy the pressure data
    auto found = gdk_event_get_axis(sourceEvent, GDK_AXIS_PRESSURE, &targetEvent.pressure);
    xoj_assert_message(!found || targetEvent.pressure >= 0, "Pressure must be positive! Either your device or your driver may be broken!");
```
When we find a device, that returns negative pressure, we can't use -1 as no pressure anymore.
Also, it would be good to have a calibration mode, where we can get the pressure ranges from that device including (step, min, max)

bhennion:
> Another possibility is, an event, which somehow does not have the Pressure axis but the previous and the next do have them, but all are fed to StrokeHandler/PenInputHandler.

Yes, that's what I had in mind when I made those installers. The log should contain the sequence of pressures values of the affected stroke. We'll see if the sign changes.

Febbe:
> Ok, I can reproduce the issue on my device, after disabling `touchscreen > activate internal hand recognition`
>
> The cause is, that somehow a touchscreen event which does not have a pressure is fed to the PenInputHandler
>
> _Originally posted by @Febbe in https://github.com/xournalpp/xournalpp/issues/5025#issuecomment-1673949253_

I think we should filter every event, which does not belong to the same device, which started an event sequence.

bhennion:
> I think we should filter every event, which does not belong to the same device, which started an event sequence.

Ok. Do you want to implement it or would you rather I did it?

Febbe:
> > I think we should filter every event, which does not belong to the same device, which started an event sequence.
> 
> Ok. Do you want to implement it or would you rather I did it?

Would be better, you touched that code lately, and I couldn't find a good place currently.

My best bet was 
```
void StrokeHandler::paintTo(Point point) {
    if (this->hasPressure && point.z > 0.0) {
        point.z *= this->stroke->getWidth();
    }
```

But I don't know how to handle the case, when the first event has no pressure, but the following. We should keep in mind, that we can have a Pen device, which does not report a pressure in general, and we can have devices which report no pressures for some events, including the first.

bhennion:
Would this new commit do the trick?
It does not handle the case where the first event has no pressure value (it behaves as if there was no pressure values at all), but handles the missing values of follow-up events.

Febbe:
From what I read, it would prevent crashing. But I am not sure, if it is like desired, since the Event looses its pressure information, no matter which pressure it had before, assuming 0.01 might be not desirable. Either we want to also infer the pressure for that point, or copy the value from the previous point (which only works, when the Stroke is itself pressure sensitive).
But I'll have to test this on m,y device, if there are graphical artifacts visible from this.

src/core/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -163,6 +164,13 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
     // Forward event to page
     if (currentPage) {
         PositionInputData pos = this->getInputDataRelativeToCurrentPage(currentPage, event);
+        if (pos.pressure != Point::NO_PRESSURE) {
+            pressureMode = PressureMode::DEVICE_PRESSURE;
+        } else if (this->inputContext->getSettings()->isPressureGuessingEnabled()) {
```
Febbe, line 169: I think something like this could do the trick:
```suggestion
        } else if (this->inputContext->getSettings()->isPressureGuessingEnabled() || (stroke->size() > 0 && stroke->hasPressure())) {
```

src/core/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -171,46 +179,57 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
...
-        PositionInputData lastPos = getInputDataRelativeToCurrentPage(page, this->lastEvent);
-
-        double dt = (pos.timestamp - lastPos.timestamp) / 10.0;
-        double distance = utl::Point<double>(pos.x, pos.y).distance(utl::Point<double>(lastPos.x, lastPos.y));
-        double inverseSpeed = dt / (distance + 0.001);
-
-        // This doesn't have to be exact. Arctan is used here for its sigmoid-like shape,
-        // so that lim inverseSpeed->infinity (newPressure) is some finite value.
-        double newPressure = 3.142 / 2.0 + std::atan(inverseSpeed * 3.14 - 1.3);
+double PenInputHandler::inferPressureValue(PositionInputData const& pos, XojPageView* page) {
```
Febbe, line 182: Didn't we have a paper or something for this algorithm? It would be good, to have the reference in the code.

Febbe:
Does not crash and I don't have any artifacts.

bhennion:
> I don't think that you need to change anything.

Alright! Thanks for testing and reviewing!
Merging once the CI check passes (I rebased on release-1.2)

Febbe:
After this is done, we can release xournalpp v1.2.1

rolandlo:
The `xoj_assert` is only in the master branch. You should use `assert` instead (and fix the include).

bhennion:
> The `xoj_assert` is only in the master branch. You should use `assert` instead (and fix the include).

:) Indeed!
This should be fixed now. Merging once the CI checks are passed.

Issue #4394, opened 2022-11-06
[lua Api] sidebarActions not reliably working


atticus-sullivan:
**Affects versions :**
 - OS: Ach Linux
 - (Linux only) Desktop environment: X11 with bspwm
 - Which version of libgtk do you use: libgtk: 3.24.34
 - Version of Xournal++: 1.1.2
 - Installation method: arch linux repo

**Describe the bug**
I'm testing the `clone`-function of the provided `LayerActions` plugin. The trouble I've got is that sometimes simply the current page is deleted while the annotations go to the next page.
I already had a look at the code and I'm pretty sure the problem is that `app.sidebarAction("COPY")` isn't executed.

In my tests this problem didn't occur on documents without a table of contents (sections, ... in LaTeX). After some more testing I figured out that if I manually open the sidebar and select the `page preview` "tab" is still works.
My conclusion is that the correctness of `app.sidebarAction` depends on the "tab" which is selected in the sidebar (`content`s seems to be selected by default if available -> not working, if not available `page preview` is selected -> working).

*Not working example*
![Peek 2022-11-06 20-44](https://user-images.githubusercontent.com/60581958/200192348-e900def7-dffd-4629-b96b-882c3cd0e6ed.gif)
Here I just opened the pdf, draw some stuff and applied the plugin function. Note that in the first line of the slide there is `1` and `A` indicating that (sub)sections are being used.

*Working example*
![Peek 2022-11-06 20-45](https://user-images.githubusercontent.com/60581958/200192390-099284dc-010b-4c87-aa6a-155833b16eeb.gif)
Here I only opened the sidebar and selected `page preview` before applying the plugin function.

**possible fix**
I'm not sure how the sidebar works (not into gui stuff yet) and if it's possible to call sidebar actions from another sidebar tab, but maybe a (bit dirty, since we might interfere a bit with the user) workaround could be exposing the sidebar tab to the lua api so that the plugin can switch the right tab and then switch back just before calling the sidebar action.

rolandlo:
I think `Sidebar::setSelectedPage(size_t page)` can be used for that.
`page = 0`: sidebar index page
`page = 1`: pages previews
`page = 2` : layer previews (not stacked)
`page = 3 `: layer previews (stacked)
See `Sidebar::initPages`. 
[...]

PR #5030, opened 2023-08-09
Refactor sidebar to use vector+iterator instead of list+pointer

atticus-sullivan:
The advantage over the former approach of using a std::list is the random access and the fast calculation of indices. And using an iterator over a pointer is the more modern way of cpp for this case. We also can rapidly calculate the index of the current page via std::distance.

The downside of using a std::vector is that insertion might be slow (if the vector needs to be copied), but as there are only a few pages and we can preallocate space for the vector this should be no huge disadvantage.

The big downside of using an iterator over a pointer is iterator invalidation. But as pages isn't exposed to the outside, we can ensude in Sidebar.cpp that on each possible iterator invalidation, the iterator is refreshed.
Also when using a pointer we would need to refresh the pointer on e.g. insertion into the std::vector because probably the vector got copied to another location at this point.

See https://github.com/xournalpp/xournalpp/issues/4394 for the motivation behind this change (implement a `getSidebarPageNo`).

Please also consider the doubts in this [comment](https://github.com/xournalpp/xournalpp/issues/4394#issuecomment-1533196776) as well as my [response](https://github.com/xournalpp/xournalpp/issues/4394#issuecomment-1533575506) when reviewing this.

### DIFF
--- a/src/core/gui/sidebar/Sidebar.cpp
+++ b/src/core/gui/sidebar/Sidebar.cpp
@@ -1,8 +1,8 @@
 #include "Sidebar.h"
 
-#include <cstdint>    // for int64_t
-#include <memory>     // for make_shared
-#include <string>     // for string
+#include <cstdint>  // for int64_t
+#include <memory>   // for std::make_unique and std::make_shared
+#include <string>   // for string
 
 #include <config-features.h>
 #include <gdk/gdk.h>      // for gdk_display_get...
@@ -35,22 +35,22 @@ Sidebar::Sidebar(GladeGui* gui, Control* control): control(control), gui(gui), t
 }
 
 void Sidebar::initPages(GtkWidget* sidebarContents, GladeGui* gui) {
-    addPage(new SidebarIndexPage(this->control, &this->toolbar));
-    addPage(new SidebarPreviewPages(this->control, this->gui, &this->toolbar));
+    addPage(std::make_unique<SidebarIndexPage>(this->control, &this->toolbar));
+    addPage(std::make_unique<SidebarPreviewPages>(this->control, this->gui, &this->toolbar));
     auto layersContextMenu = std::make_shared<SidebarLayersContextMenu>(this->gui, &this->toolbar);
-    addPage(new SidebarPreviewLayers(this->control, this->gui, &this->toolbar, false, layersContextMenu));
-    addPage(new SidebarPreviewLayers(this->control, this->gui, &this->toolbar, true, layersContextMenu));
+    addPage(std::make_unique<SidebarPreviewLayers>(this->control, this->gui, &this->toolbar, false, layersContextMenu));
+    addPage(std::make_unique<SidebarPreviewLayers>(this->control, this->gui, &this->toolbar, true, layersContextMenu));
 
     // Init toolbar with icons
 
     int i = 0;
-    for (AbstractSidebarPage* p: this->pages) {
+    for (auto&& p: this->pages) {
         GtkToolItem* it = gtk_toggle_tool_button_new();
         p->tabButton = it;
 
         gtk_tool_button_set_icon_widget(GTK_TOOL_BUTTON(it), gtk_image_new_from_icon_name(p->getIconName().c_str(),
                                                                                           GTK_ICON_SIZE_SMALL_TOOLBAR));
-        g_signal_connect(it, "clicked", G_CALLBACK(&buttonClicked), new SidebarPageButton(this, i, p));
+        g_signal_connect(it, "clicked", G_CALLBACK(&buttonClicked), new SidebarPageButton(this, i, p.get()));
         gtk_tool_item_set_tooltip_text(it, p->getName().c_str());
         gtk_tool_button_set_label(GTK_TOOL_BUTTON(it), p->getName().c_str());
 
@@ -77,7 +77,7 @@ void Sidebar::buttonClicked(GtkToolButton* toolbutton, SidebarPageButton* button
     }
 }
 
-void Sidebar::addPage(AbstractSidebarPage* page) { this->pages.push_back(page); }
+void Sidebar::addPage(std::unique_ptr<AbstractSidebarPage> page) { this->pages.push_back(std::move(page)); }
 
 void Sidebar::askInsertPdfPage(size_t pdfPage) {
     GtkWidget* dialog = gtk_message_dialog_new(control->getGtkWindow(), GTK_DIALOG_MODAL, GTK_MESSAGE_QUESTION,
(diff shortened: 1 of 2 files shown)


src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -29,6 +29,8 @@ Sidebar::Sidebar(GladeGui* gui, Control* control): control(control), gui(gui), t
 
     this->sidebarContents = gui->get("sidebarContents");
 
+    // currentPage needs to be in a valid state as other functions like addPage rely on it
+    this->currentPage = this->pages.begin();
```
Febbe, line 33: Move field initialization into the declaration in the header.
Use `size_t` instead.
```suggestion
```

src/core/gui/sidebar/Sidebar.h
```diff
@@ -124,7 +124,7 @@ class Sidebar: public DocumentListener, public SidebarToolbarActionListener {
     /**
      * Current active page
      */
-    AbstractSidebarPage* currentPage = nullptr;
+    std::vector<AbstractSidebarPage*>::iterator currentPage;
```
Febbe, line 127:
```suggestion
    size_t currentPageIdx{0};
```

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -77,7 +79,11 @@ void Sidebar::buttonClicked(GtkToolButton* toolbutton, SidebarPageButton* button
     }
 }
 
-void Sidebar::addPage(AbstractSidebarPage* page) { this->pages.push_back(page); }
+void Sidebar::addPage(AbstractSidebarPage* page) {
+    const auto ind = std::distance(this->pages.begin(), this->currentPage);
+    this->pages.push_back(page);
+    this->currentPage = this->pages.begin() + ind;
```
Febbe, line 85: I would store the page index in that case, it can't be invalidated. Therefore, it is not required to recalculate it.

```suggestion
    this->pages.emplace_back(page);
```

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -152,20 +158,20 @@ void Sidebar::selectPageNr(size_t page, size_t pdfPage) {
...
-            p->enableSidebar();
+            (*p)->enableSidebar();
         } else {
-            p->disableSidebar();
-            gtk_widget_hide(p->getWidget());
-            gtk_toggle_tool_button_set_active(GTK_TOGGLE_TOOL_BUTTON(p->tabButton), false);
+            (*p)->disableSidebar();
+            gtk_widget_hide((*p)->getWidget());
+            gtk_toggle_tool_button_set_active(GTK_TOGGLE_TOOL_BUTTON((*p)->tabButton), false);
         }
```
Febbe, line 175: This is inefficient, when using a vector/deque, we have 3 phases: disable all pages previous to **and** after `page`, activate page.

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -128,20 +134,20 @@ Sidebar::~Sidebar() {
...
     this->sidebarContents = nullptr;
-    this->currentPage = nullptr;
 }
 
 /**
  * Called when an action is performed
  */
 void Sidebar::actionPerformed(SidebarActions action) {
-    if (!this->currentPage) {
+    if (this->currentPage == this->pages.end()) {
```
Febbe, line 146:
```suggestion
    if (this->currentPageIdx >= this->pages.size()) {
```

Febbe (changes requested):
Using a std::list<any_ptr<T>> is in general a bad design. `std::vector` and `std::deque` will outperform this in literally any case. Even insertions in the middle of that vector are negligible, There is also no reason, to discuss the iterator invalidation, since we work over pointers. The data will always be valid.

So the question is, whether we use `std::vector<std::unique_ptr<AbstractSidebarPage>>` `std::vector<AbstractSidebarPage>`, `std::deque<AbstractSidebarPage>`, `std::list<AbstractSidebarPage>`

The following comments assume we decide to use std::vector<std::unique_ptr<AbstractSidebarPage>> or a std::deque

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -128,20 +134,20 @@ Sidebar::~Sidebar() {
         delete p;
     }
     this->pages.clear();
+    this->currentPage = this->pages.end();  // mark as "invalid"
```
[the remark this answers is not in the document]
atticus-sullivan: But with the current implementation we need to free the memory allocated for the pages (no smart pointers used so far). But we might change this of course.

atticus-sullivan:
Alright, I see we're still not clear on how to store the data in the Sidebar-class. The Idea was not to use only the index but some sort of wrapped pointer for faster access (at least it's intuitive to have faster access if we already "know" which element to retrieve).

At the moment, I see no real advantage of using a list/deque instead of a vector.

I'm positive on using an `unique_ptr` instead of the raw pointer for the pages container (I see no reason not to and you're right the cleanup is just much nicer).

The question still is how to store the currently active page in the container. As far as I understand you, you're in favor of just using the index (because the threat of iterator invalidation), aren't you?

I just want us to be clear on these things before discussing changes in detail.

Febbe:
I think, that there is nearly no benefit of storing a pointer/iterator over an index, when using a random access container with O(1) access time.
It can be, that the code is running on an architecture, which has the same access time for  p+0 as for p+x, then your approach has zero benefit. It is also harder to maintain, because we must always be sure to synchronize that value.
This means we must consequently write a test which assures this.

Regarding the iterator invalidation: That is barely relevant, if it's now a problem, the whole code before was already not thread safe but accessed from a different thread and saving an iterator for later reuse was ever a code smell.

Since AbstractSidebarPage is a superclass we must use pointers anyway, so storing it as value is not possible.

src/core/gui/sidebar/Sidebar.cpp
atticus-sullivan: Could you explain further what you mean by that? Do you refer to something like branch prediction? Also I don't get how this step is inefficient with a vector but when using a list it isn't.

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -77,7 +77,7 @@ void Sidebar::buttonClicked(GtkToolButton* toolbutton, SidebarPageButton* button
     }
 }
 
-void Sidebar::addPage(AbstractSidebarPage* page) { this->pages.push_back(page); }
+void Sidebar::addPage(std::unique_ptr<AbstractSidebarPage>&& page) { this->pages.push_back(std::move(page)); }
```
Febbe, line 80: Nitpick: unique_ptr's are already move only types, better to pass it by value:
```suggestion
void Sidebar::addPage(std::unique_ptr<AbstractSidebarPage> page) { this->pages.push_back(std::move(page)); }
```

src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -1,8 +1,8 @@
 #include "Sidebar.h"
 
-#include <cstdint>    // for int64_t
-#include <memory>     // for make_shared
-#include <string>     // for string
+#include <cstdint>  // for int64_t
+#include <memory>   // for make_shared
```
Febbe, line 4:
```suggestion
#include <memory>   // for std::make_unique
```

Febbe (approved):
Ok, your PR looks good so far, but the commit description is wrong. Please sqash your changes and improve the commit description. I can help with that, when you don't know how to do this.

Generally prefer `auto&&` or `auto const&` over `auto &` in for loops (Unlike `Sometype&&` `auto&&` and `template <class T> T&&` are forward references, so they exactly infer the type returned)

src/core/gui/sidebar/Sidebar.cpp
Febbe: Yes, that's a branch prediction concern.
You are right, it's also inefficient with a std::list. 
But as we only have 4 pages in average, it does not matter (I just reviewed the loop, without checking the whole context).
Just leave it as it.

atticus-sullivan:
Alright, I added your changes and used `auto&&` for the loops (thanks for this comment, didn't quite thought of that).

Just let me know if the current commit message is what you've expected.

Febbe:
LGTM

PR #5032, opened 2023-08-10
Make bugreports use bug report template

rolandlo:
The bug reports we received didn't use the bug report template, so we had to ask a lot of follow up questions that would be answered when filling out the template. This PR fixes that problem.


rolandlo:

PR #5037, opened 2023-08-11
Changing the pdf background - preparing for merge to master

simone-romiti:
Fixed the rebase confusion in https://github.com/xournalpp/xournalpp/pull/4803


src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,58 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+
+    if (this->undoRedo->isChanged()) {
+        string msg = FS(_F("Error in changing the background: please save the recent changes first.")) +
+                     loadHandler.getLastError();
+        XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+        return true;
+    }
+
+    if (current_filepath.empty()) {
+        string msg = FS(_F("Error in changing the background: please open and save some file first.")) +
```
rolandlo, line 2434: That formulation can be improved. What about this?
```suggestion
        string msg = FS(_F("Error in changing the background: please save the document first.")) +
```

src/core/enums/ActionType.enum.h
```diff
@@ -29,6 +29,7 @@ enum ActionType {
     // Menu file
     ACTION_NEW = 100,
     ACTION_OPEN,
+    ACTION_CHANGE_BACKGROUND,
```
rolandlo, line 32: I suggest to rename this variable to `ACTION_CHANGE_PDF_BACKGROUND`.

ui/main.glade
```diff
@@ -188,6 +188,16 @@
                             <property name="use-underline">True</property>
                           </object>
                         </child>
+                        <child>
+                          <object class="GtkMenuItem" id="menuFileChangeBackground">
+                            <property name="name">menuFileChangeBackground</property>
+                            <property name="visible">True</property>
+                            <property name="can-focus">False</property>
+                            <property name="label" translatable="yes">_Change _PDF _Background</property>
```
rolandlo, line 196: Only one letter should be highlighted.
```suggestion
                            <property name="label" translatable="yes">_Change PDF Background</property>
```

rolandlo (changes requested):
It works fine now, including attach mode. I have some minor comments though to take into consideration.

src/core/control/Control.cpp
simone-romiti: I wrote it like this because the user may try to change the background before having saved. Maybe like this?
`"Error in changing the background: please open (and save) a document first."`
what do you think?

ui/main.glade
simone-romiti: done

src/core/enums/ActionType.enum.h
simone-romiti: done

src/core/control/Control.cpp
rolandlo: I don't agree. It's about saving the document (.xopp-file). You can open a PDF (via `File > Open`) and try to change the PDF background. Then it doesn't make sense if you see the error message "Please open (and save) a document first".

simone-romiti: You're right, I've changed it to the message you suggested https://github.com/xournalpp/xournalpp/pull/5037#discussion_r1292102884

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,58 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
     return true;
 }
 
+auto Control::changePdfBackground() -> bool {
+
+    LoadHandler loadHandler;
+
+    const fs::path current_filepath = ((*this).doc)->getFilepath();
+    const fs::path current_PDFfilepath = ((*this).doc)->getPdfFilepath();
```
rolandlo, line 2414: According to the [coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions#naming-conventions) you should use camelBack for these variables.
```suggestion
    const fs::path currentFilepath = ((*this).doc)->getFilepath();
    const fs::path currentPdfFilepath = ((*this).doc)->getPdfFilepath();
```

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,58 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+
+    auto pageNr = getCurrentPageNo();
+    PageRef page = this->doc->getPage(pageNr);
+    Layer* layer = page->getSelectedLayer();
+
+    if (current_PDFfilepath.empty()) {
+        string msg = FS(_F("Error in changing the background: please open a pdf first.")) + loadHandler.getLastError();
+        XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+        return true;
```
rolandlo, line 2423: The return value should mean `success`.
```suggestion
        return false;
```

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,58 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+    }
+
+    if (this->undoRedo->isChanged()) {
+        string msg = FS(_F("Error in changing the background: please save the recent changes first.")) +
+                     loadHandler.getLastError();
+        XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+        return true;
```
rolandlo, line 2430:
```suggestion
        return false;
```

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,58 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+    }
+
+    if (current_filepath.empty()) {
+        string msg = FS(_F("Error in changing the background: please save the document first.")) +
+                     loadHandler.getLastError();
+        XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+        return true;
```
rolandlo, line 2437:
```suggestion
        return false;
```

rolandlo (changes requested):
On closer inspection I have found a couple of more issues.

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,60 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+    XojOpenDlg dlg(getGtkWindow(), this->settings);
+    auto pdfFilename = dlg.showOpenDialog(true, attachToDocument);
+    if (!pdfFilename.empty()) {
+        loadHandler.setPdfReplacement(pdfFilename, attachToDocument);
+        loadedDocument = loadHandler.loadDocument(current_filepath);
+        if (!loadedDocument) {
+            string msg = FS(_F("Error opening file \"{1}\"") % currentFilepath.u8string()) + "\n" + loadHandler.getLastError();
+            XojMsgBox::showErrorToUser(getGtkWindow(), msg);
+            return false;
+    }
```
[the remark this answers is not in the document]
simone-romiti: yes, also some variable names were outdated.

simone-romiti:
There's one thing I don't know how to implement: if the user changes the pdf background and closes the document without saving, no warning message is thrown. How can I change that using the `"This document is not saved yet."` pop-up window?

Technius:
When the page background is changed, you'll have to call `UndoRedoHandler::addUndoAction()` to add an undo action that tracks the change. I think you should be able to reuse the existing `PageBackgroundChangedUndoAction` class to create an undo action.

Technius:
On second thought, `PageBackgroundChangedUndoAction` is meant to track the page type of a specific page. We might have to create another subclass of `UndoAction` that represents a background PDF change for the entire document.

simone-romiti:
@Technius this seems a separate issue, isn't it? The same type of behavior of https://github.com/xournalpp/xournalpp/pull/5037#issuecomment-1675923042 shows up also when opening a pdf and not saving, independently of whether we change the PDF background or not.

Can we merge this branch to master?

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,62 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+    LoadHandler loadHandler;
+
+    const fs::path currentFilepath = ((*this).doc)->getFilepath();
+    const fs::path currentPdfFilepath = ((*this).doc)->getPdfFilepath();
+
+    auto pageNr = getCurrentPageNo();
+    PageRef page = this->doc->getPage(pageNr);
+    // Layer* layer = page->getSelectedLayer();
+
+    if (currentPdfFilepath.empty()) {
```
Technius, line 2420: Is this restriction necessary? It might be useful to allow an empty current PDF in case the user wants to attach a PDF background to a document that does not already have one.

Technius (changes requested):
Thank you for this PR, I think there are a few concerns that need to be addressed before this can be merged (see review comments).

> this seems a separate issue, isn't it?

Sorry, I wasn't very clear. The application relies on using the undo/redo stack to track when a document is "modified" (see line checking `this->undoRedo->isChanged()`), and changing the PDF background does not add an UndoAction to the stack. To demonstrate this, try drawing on the document after changing the PDF background--the prompt will appear when closing the application.

To allow a PDF background change to "modify" the document, we'll need to implement an UndoAction for the background change. However, I currently do not think it would be easy to implement given the current approach in this PR (i.e., reloading the whole document).

> The same type of behavior of [#5037 (comment)](https://github.com/xournalpp/xournalpp/pull/5037#issuecomment-1675923042) shows up also when opening a pdf and not saving, independently of whether we change the PDF background or not.

I'm confused, could you clarify the steps you are using? If I use "File > Open" and select a PDF file, then I get a "please save the document first" message. The same thing happens when using "File > Annotate PDF".

src/core/control/Control.cpp
simone-romiti: In principle no, though the logic would be cumbersome I think. §
If you haven't opened a pdf, you can't change the background. For that you should use "Open" or "Annotate PDF". Do you agree?

simone-romiti:
> If I use "File > Open" and select a PDF file, then I get a "please save the document first" message. The same thing happens when using "File > Annotate PDF".

Sorry, I don't understand the question. I didn't change the "Open" and "Annotate PDF".

src/core/control/Control.cpp
Technius: No, I don't see why we would need a limitation like that. For example, suppose I create a blank document but then decide to add some PDF pages later. In this case, it would make sense to allow using the Change PDF Background menu item to append the extra PDF pages.

src/core/control/Control.cpp
```diff
@@ -2403,6 +2406,62 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
...
+            return false;
+        }
+    }
+
+    this->closeDocument();
+
+    this->doc->lock();
+    this->doc->clearDocument();
+    *this->doc = *loadedDocument;
+    this->doc->unlock();
```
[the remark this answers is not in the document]
Technius: Probably something like this:
<details>
<summary>Patch to the existing PR</summary>

```diff
diff --git a/src/core/control/Control.cpp b/src/core/control/Control.cpp
index 45306ce8e..281c2f07f 100644
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2435,30 +2435,24 @@ auto Control::changePdfBackground() -> bool {
         return false;
     }
 
-    Document* loadedDocument = loadHandler.loadDocument(currentFilepath);
-
     bool attachToDocument = false;
     XojOpenDlg dlg(getGtkWindow(), this->settings);
     auto pdfFilename = dlg.showOpenDialog(true, attachToDocument);
-    if (!pdfFilename.empty()) {
-        loadHandler.setPdfReplacement(pdfFilename, attachToDocument);
-        loadedDocument = loadHandler.loadDocument(currentFilepath);
-        if (!loadedDocument) {
-            string msg = FS(_F("Error opening file \"{1}\"") % currentFilepath.u8string()) + "\n" +
-                         loadHandler.getLastError();
-            XojMsgBox::showErrorToUser(getGtkWindow(), msg);
-            return false;
-        }
-    }
-
-    this->closeDocument();
 
-    this->doc->lock();
-    this->doc->clearDocument();
-    *this->doc = *loadedDocument;
-    this->doc->unlock();
+    // To prevent render jobs from running while we are changing the document,
+    // we need to lock the scheduler.
+    getScheduler()->lock();
+    bool success = this->doc->readPdf(pdfFilename, false, attachToDocument);
+    if (!success) {
+        getScheduler()->unlock();
+        g_warning("Failed to read new PDF document.");
+    } else {
+        fileLoaded(-1);
+        getScheduler()->unlock();
+        // Pretend the whole document has changed to force re-render.
+        this->win->getXournal()->documentChanged(DOCUMENT_CHANGE_COMPLETE);
+    }
 
-    fileLoaded(-1);
     return true;
 }

```
</details>

The last call to `his->win->getXournal()->documentChanged(DOCUMENT_CHANGE_COMPLETE)` is a hack to invalidate the PdfCache, which will hold a reference to the poppler PDF document. I think the old poppler PDF document will be destroyed after we load the new PDF, hence the need to reset the cache.

I haven't tested this though, so not sure if it will work. This is more complicated code, but it will also allow a UndoAction to be defined (if that is something you'd like to address in this PR).

src/core/control/Control.cpp
simone-romiti: I had a similar discussion (but somewhat in the other direction) in the parent PR https://github.com/xournalpp/xournalpp/pull/4803 with @rolandlo: see e.g. [this comment](https://github.com/xournalpp/xournalpp/pull/4803#pullrequestreview-1561614906).
Could you please suggest a patch on which [you](@Technius) and @rolandlo agree? Thank you in advance.

src/core/control/Control.cpp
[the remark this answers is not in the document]
simone-romiti: Thank you for the patch, the code compiles without errors.
I wasn't thinking about an UndoAction for this PR.

PR #5057, opened 2023-08-15
Fixed UB due to incompatible callbacks

Febbe:
Many callbacks had the return value `bool`, which is a c++ only type, but the C API from gtk/gdk/glib expected `gboolean`.
The reinterpret_cast to the corresponding functions is undefined behavior.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -91,6 +91,7 @@
 #include "util/Stacktrace.h"                                     // for Stac...
 #include "util/Util.h"                                           // for exec...
 #include "util/XojMsgBox.h"                                      // for XojM...
+#include "util/glib_casts.h"                                     // for wrap_v
 #include "util/i18n.h"                                           // for _, FS
 #include "util/serializing/InputStreamException.h"               // for Inpu...
 #include "util/serializing/ObjectInputStream.h"                  // for Obje...
@@ -155,7 +156,7 @@ Control::Control(GApplication* gtkApp, GladeSearchpath* gladeSearchPath): gtkApp
     /**
      * This is needed to update the previews
      */
-    this->changeTimout = g_timeout_add_seconds(5, reinterpret_cast<GSourceFunc>(checkChangedDocument), this);
+    this->changeTimout = g_timeout_add_seconds(5, xoj::util::wrap_v<checkChangedDocument>, this);
 
     this->pageBackgroundChangeController = new PageBackgroundChangeController(this);
 
@@ -377,7 +378,7 @@ void Control::enableAutosave(bool enable) {
 
     if (enable) {
         auto timeout = guint(settings->getAutosaveTimeout()) * 60U;
-        this->autosaveTimeout = g_timeout_add_seconds(timeout, reinterpret_cast<GSourceFunc>(autosaveCallback), this);
+        this->autosaveTimeout = g_timeout_add_seconds(timeout, xoj::util::wrap_v<autosaveCallback>, this);
     }
 }
 
@@ -2479,7 +2480,6 @@ public:
  */
 auto Control::loadMetadataCallback(MetadataCallbackData* data) -> bool {
     if (!data->md.valid) {
-        delete data;
         return false;
     }
     ZoomControl* zoom = data->ctrl->zoom;
@@ -2493,9 +2493,6 @@ auto Control::loadMetadataCallback(MetadataCallbackData* data) -> bool {
         zoom->setZoom(data->md.zoom * zoom->getZoom100Value());
     }
     data->ctrl->scrollHandler->scrollToPage(data->md.page);
-
-    delete data;
-
     // Do not call again!
     return false;
 }
@@ -2505,7 +2502,8 @@ void Control::loadMetadata(MetadataEntry md) {
     data->md = std::move(md);
     data->ctrl = this;
 
-    g_idle_add(reinterpret_cast<GSourceFunc>(loadMetadataCallback), data);
+    g_idle_add_full(G_PRIORITY_DEFAULT_IDLE, xoj::util::wrap_v<loadMetadataCallback>, data,
+                    &xoj::util::destroy_cb<MetadataCallbackData>);
 }
 
 auto Control::annotatePdf(fs::path filepath, bool /*attachPdf*/, bool attachToDocument) -> bool {
--- a/src/core/control/jobs/BlockingJob.cpp
+++ b/src/core/control/jobs/BlockingJob.cpp
@@ -1,6 +1,6 @@
 #include "BlockingJob.h"
 
-#include <glib.h>     // for g_idle_add, GSourceFunc
+#include <glib.h>     // for g_idle_add
 #include <gtk/gtk.h>  // for gtk_widget_grab_focus
 
 #include "control/Control.h"   // for Control
@@ -8,6 +8,7 @@
 #include "gui/MainWindow.h"    // for MainWindow
 #include "gui/XournalView.h"   // for XournalView
 #include "util/Util.h"         // for execInUiThread
+#include "util/glib_casts.h"   // for wrap_for_once_v
 
 BlockingJob::BlockingJob(Control* control, const std::string& name): control(control) { control->block(name); }
 
@@ -16,7 +17,7 @@ BlockingJob::~BlockingJob() { this->control = nullptr; }
 void BlockingJob::execute() {
     this->run();
 
-    g_idle_add(reinterpret_cast<GSourceFunc>(finished), this->control);
+    g_idle_add(xoj::util::wrap_for_once_v<finished>, this->control);
 }
 
 auto BlockingJob::finished(Control* control) -> bool {
--- a/src/core/control/jobs/BlockingJob.h
+++ b/src/core/control/jobs/BlockingJob.h
@@ -30,9 +30,7 @@ public:
     JobType getType() override;
 
 protected:
-    static bool finished(Control* control);
+    static auto finished(Control* control) -> bool;
 
-private:
-protected:
     Control* control;
 };
--- a/src/core/control/jobs/Job.cpp
+++ b/src/core/control/jobs/Job.cpp
@@ -1,7 +1,10 @@
 #include "Job.h"
 
 #include <gdk/gdk.h>  // for gdk_threads_add_idle
-#include <glib.h>     // for g_source_remove, GSourceFunc
+#include <glib.h>     // for g_source_remove
+
+#include "util/glib_casts.h"  // for wrap_for_once_v
+
 
 Job::Job(): refCount(1) {}
 
@@ -46,7 +49,7 @@ void Job::callAfterRun() {
     }
 
     this->ref();
-    this->afterRunId = gdk_threads_add_idle(reinterpret_cast<GSourceFunc>(Job::callAfterCallback), this);
+    this->afterRunId = gdk_threads_add_idle(xoj::util::wrap_for_once_v<Job::callAfterCallback>, this);
 }
 
 void Job::afterRun() {}
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -30,6 +30,7 @@
 #include "undo/UndoRedoHandler.h"                  // for UndoRedoHandler
 #include "util/Range.h"                            // for Range
 #include "util/Util.h"                             // for cairo_set_dash_from_vector
+#include "util/glib_casts.h"                       // for wrap_v
 #include "util/i18n.h"                             // for _
 #include "util/serializing/ObjectInputStream.h"    // for ObjectInputStream
 #include "util/serializing/ObjectOutputStream.h"   // for ObjectOutputStream
@@ -771,8 +772,7 @@ void EditSelection::moveSelection(double dx, double dy, bool addMoveUndo) {
 void EditSelection::setEdgePan(bool pan) {
     if (pan && !this->edgePanHandler) {
         this->edgePanHandler = g_timeout_source_new(1000 / PAN_TIMER_RATE);
-        g_source_set_callback(this->edgePanHandler, reinterpret_cast<GSourceFunc>(EditSelection::handleEdgePan), this,
-                              nullptr);
+        g_source_set_callback(this->edgePanHandler, xoj::util::wrap_v<EditSelection::handleEdgePan>, this, nullptr);
         g_source_attach(this->edgePanHandler, nullptr);
     } else if (!pan && this->edgePanHandler) {
         g_source_destroy(this->edgePanHandler);
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -33,6 +33,7 @@
 #include "util/PathUtil.h"                              // for getConfigFile
 #include "util/Util.h"                                  // for execInUiThread, npos
 #include "util/XojMsgBox.h"                             // for XojMsgBox
+#include "util/glib_casts.h"                            // for wrap_for_once_v
 #include "util/i18n.h"                                  // for FS, _F
 
 #include "GladeSearchpath.h"     // for GladeSearchpath
@@ -76,10 +77,13 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
     setSidebarVisible(control->getSettings()->isSidebarVisible());
 
     // Window handler
-    g_signal_connect(this->window, "delete-event", G_CALLBACK(deleteEventCallback), this->control);
-    g_signal_connect(this->window, "window_state_event", G_CALLBACK(windowStateEventCallback), this);
+    g_signal_connect(this->window, "delete-event", xoj::util::wrap_for_g_callback_v<deleteEventCallback>,
+                     this->control);
+    g_signal_connect(this->window, "window_state_event", xoj::util::wrap_for_g_callback_v<windowStateEventCallback>,
+                     this);
 
-    g_signal_connect(get("buttonCloseSidebar"), "clicked", G_CALLBACK(buttonCloseSidebarClicked), this);
+    g_signal_connect(get("buttonCloseSidebar"), "clicked", xoj::util::wrap_for_g_callback_v<buttonCloseSidebarClicked>,
+                     this);
 
     // "watch over" all key events
     g_signal_connect(this->window, "key-press-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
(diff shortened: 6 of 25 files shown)


Febbe:
@bhennion / @rolandlo, pls merge after aproval.

bhennion (commented):
I fear that the changes from ArgumentType* to gpointer's all over the place reduce readability quite a bit.
Other than that, you can use a g_idle_add_once or similar functions when the callback is always only called once, to remove the return type and imprve readability.

Febbe:
> I fear that the changes from ArgumentType* to gpointer's all over the place reduce readability quite a bit.
> Other than that, you can use a g_idle_add_once or similar functions when the callback is always only called once, to remove the return type and imprve readability.

I don't like it also, but gtk relies on the fact, that undefined behavior is implemented in all compilers the same (reinterpreting functions with the same argument size and count).
The only way to make it generally correct would be to use gpointer for data parameters.
It also reduces the usage of reinterpret_cast.

We could use some sort of additional callback_gtk function. Or rename the data parameters to be more meaningful.

Febbe:
> I have noticed that some callbacks/functions in `Scheduler.cpp`, `TextEditor.cpp`, `Control.cpp`, `ToolAdapter.cpp` and `Document.cpp` whose return type were changed to `gboolean` continue to return `true` and/or `false. I don't get the logic behind that.

Implicit casts are not UB but discouraged (readability), but reinterpreting a function which returns a byte into a function which returns a 4 byte value is UB. For example even when all return values have at least sizeof(intptr_t) the compiler can decide, that the byte returning the bool is the 2nd in the range of the sizeof(intptr_t), or that the rest is not initialized with zero(this actually happens with msvc),therefore every return can be randomly true, even false ist returned.

src/core/control/jobs/BlockingJob.h
```diff
@@ -30,9 +33,7 @@ class BlockingJob: public Job {
     JobType getType() override;
 
 protected:
-    static bool finished(Control* control);
+    static auto finished(gpointer data) -> gboolean;
```
Febbe, line 36:
```suggestion
    static void finished(gpointer data);
```

src/core/control/jobs/BlockingJob.cpp
```diff
@@ -29,7 +30,7 @@ auto BlockingJob::finished(Control* control) -> bool {
     });
 
     // do not call again
-    return false;
+    return G_SOURCE_REMOVE;
```
Febbe, line 33:
```suggestion
    return;
```

src/core/gui/MainWindow.cpp
```diff
@@ -327,7 +328,7 @@ void MainWindow::dragDataRecived(GtkWidget* widget, GdkDragContext* dragContext,
             const char* uri = uris[i];
 
             GCancellable* cancel = g_cancellable_new();
-            int cancelTimeout = g_timeout_add(3000, reinterpret_cast<GSourceFunc>(cancellable_cancel), cancel);
+            auto cancelTimeout = g_timeout_add(3000, cancellable_cancel, cancel);
```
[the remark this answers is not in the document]
bhennion: Too bad...

Febbe:
This should also not block 1.2.1.

Febbe:
@rolandlo replaced all `bool` returns with the corresponding macro return value.

rolandlo (approved):
From my side this looks fine.

bhennion:
> We could use some sort of additional callback_gtk function. Or rename the data parameters to be more meaningful.

Could we have a template like this:
```
template<class T, gboolean fun(T*)>
gboolean wrap(gpointer d) { return fun(reinterpret_cast<T*>(d)); }
```
to be used e.g.:
```
this->changeTimout = g_timeout_add_seconds(5, wrap<Control, checkChangedDocument>, this);
```
This'd work for GSourceFunc and we would need a similar template for vaious GCallback forms.

bhennion:
Actually, glib/gdk/gtk heavily rely on function pointer recasting UB, so fixing it on our end is mostly cosmetic. Look e.g. at GLib's
```
void
g_array_sort (GArray       *farray,
              GCompareFunc  compare_func)
{
  GRealArray *array = (GRealArray*) farray;

  g_return_if_fail (array != NULL);

  /* Don't use qsort as we want a guaranteed stable sort */
  if (array->len > 0)
    g_qsort_with_data (array->data,
                       array->len,
                       array->elt_size,
                       (GCompareDataFunc)compare_func,
                       NULL);
}
```
with `GCompareFunc = gint(gconstpointer, gconstpointer)*` but `GCompareDataFunc = gint(gconstpointer, gonstpointer, gpointer)*`.

See also this [discussion](https://stackoverflow.com/questions/559581/casting-a-function-pointer-to-another-type)

Febbe:
> > We could use some sort of additional callback_gtk function. Or rename the data parameters to be more meaningful.
> 
> Could we have a template like this:
> 
> ```
> template<class T, gboolean fun(T*)>
> gboolean wrap(gpointer d) { return fun(reinterpret_cast<T*>(d)); }
> ```
> 
> to be used e.g.:
> 
> ```
> this->changeTimout = g_timeout_add_seconds(5, wrap<Control, checkChangedDocument>, this);
> ```
> 
> This'd work for GSourceFunc and we would need a similar template for vaious GCallback forms. (There must be a way to get the first template parameter inferred, but I did not look)

Also thought about something like this, but it was too complicated, to infer all different template arguments.
The wrapper should also work without reinterpreting, note that casting a void* to a concrete type can be done with a `static_cast`.

Febbe:
> Actually, glib/gdk/gtk heavily rely on function pointer recasting UB, so fixing it on our end is mostly cosmetic.

Not mostly cosmetic, it prevents us also from using `-Werror` with`-Wall` and `-Wpedantic` / `-Wextra`

bhennion:
> Also thought about something like this, but it was too complicated, to infer all different template arguments.

It's a bit ugly, but that works. I could not find a decent way to handle any number of parameters so I made one for each. There may be signal callbacks with 4 or 5 parameters, we can add the missing cases if so.
The template parameter inference works as excpected: `wrap<anyCallbackFunction>` is a function like `anyCallbackFunction` except the last parameter is now a gpointer (as long as anyCallbackFunction has 1, 2 or 3 parameters).

```
template <typename T>
struct signature {};

template <typename R, typename... T>
struct signature<R (*)(T... ts)> {
    using return_type = R;
    using parameters = std::tuple<T...>;
    using parameter_number = std::tuple_size<parameters>;
    template <size_t n>
    using parameter_type = std::tuple_element_t<n, parameters>;
    using last_parameter_type = parameter_type<parameter_number::value - 1>;
};

template <auto fun, std::enable_if_t<signature<decltype(fun)>::parameter_number::value == 1, bool> = true>
typename signature<decltype(fun)>::return_type wrap(gpointer d) {
    return fun(static_cast<typename signature<decltype(fun)>::last_parameter_type>(d));
}

template <auto fun, std::enable_if_t<signature<decltype(fun)>::parameter_number::value == 2, bool> = true,
          typename A = typename signature<decltype(fun)>::template parameter_type<0>>
typename signature<decltype(fun)>::return_type wrap(A* a, gpointer d) {
    return fun(a, static_cast<typename signature<decltype(fun)>::last_parameter_type>(d));
}

template <auto fun, std::enable_if_t<signature<decltype(fun)>::parameter_number::value == 3, bool> = true,
          typename A = typename signature<decltype(fun)>::template parameter_type<0>,
          typename B = typename signature<decltype(fun)>::template parameter_type<1>>
typename signature<decltype(fun)>::return_type wrap(A a, B b, gpointer d) {
    fun(a, b, static_cast<typename signature<decltype(fun)>::last_parameter_type>(d));
}
```

bhennion:
> > Actually, glib/gdk/gtk heavily rely on function pointer recasting UB, so fixing it on our end is mostly cosmetic.
> 
> Not mostly cosmetic, it prevents us also from using `-Werror` with`-Wall` and `-Wpedantic` / `-Wextra`

Doesn't G_SOURCE_FUNC do that for us: https://docs.gtk.org/glib/func.SOURCE_FUNC.html

Febbe:
> > > Actually, glib/gdk/gtk heavily rely on function pointer recasting UB, so fixing it on our end is mostly cosmetic.
> > 
> > Not mostly cosmetic, it prevents us also from using `-Werror` with`-Wall` and `-Wpedantic` / `-Wextra`
> 
> Doesn't G_SOURCE_FUNC shut those errors off for us? https://docs.gtk.org/glib/func.SOURCE_FUNC.html

Yep, it's a reinterpret_cast and as I know not usable for every function (only signal connector like functions). I'll try to implement a general approach.

Febbe:
@bhennion I found a way with callback_once, but there might be more elegant ways to do it:

```
template <auto SrcFn, class ST, class... Args>
auto callback_once_impl(Args... args, void* data) -> gboolean {
    SrcFn(args..., static_cast<ST>(data));
    return G_SOURCE_REMOVE;
}

template <typename F>
struct FunctionTraits;

template <typename R, typename... Args>
struct FunctionTraits<R (&)(Args...)> {
    using Reference = R (&)(Args...);
    using RetType = R;
    using ArgTypes = std::tuple<Args...>;
    static constexpr std::size_t ArgCount = sizeof...(Args);
    template <std::size_t N>
    using NthArg = std::tuple_element_t<N, ArgTypes>;
    using FirstArg = NthArg<0>;
    using LastArg = NthArg<ArgCount - 1>;
};

template <typename R, typename... Args>
struct FunctionTraits<R (*)(Args...)> {
    using Pointer = R (*)(Args...);
    using RetType = R;
    using ArgTypes = std::tuple<Args...>;
    static constexpr std::size_t ArgCount = sizeof...(Args);
    template <std::size_t N>
    using NthArg = std::tuple_element_t<N, ArgTypes>;
    using FirstArg = NthArg<0>;
    using LastArg = NthArg<ArgCount - 1>;
};

template <typename F, size_t... Is>
auto indices_impl(F f, std::index_sequence<Is...>) {
    return f(std::integral_constant<size_t, Is>()...);
}

template <size_t N, typename F>
auto indices(F f) {
    return indices_impl(f, std::make_index_sequence<N>());
}

template <typename F, typename... Ts>
auto drop_last(F f, Ts... ts) {}

template <auto SrcFn>
auto callback_once() -> auto {
    static constexpr auto traits = FunctionTraits<decltype(SrcFn)>{};
    //using traits = FunctionTraits<decltype(SrcFn)>;
    return indices<FunctionTraits<decltype(SrcFn)>::ArgCount - 1>([&](auto... Is) {
        return callback_once_impl<SrcFn, typename FunctionTraits<decltype(SrcFn)>::LastArg, std::remove_reference_t<decltype(std::get<Is>(std::declval<typename FunctionTraits<decltype(SrcFn)>::ArgTypes>()))>...>;
    });
}

static void tester(int*, double*, char*) {}

static auto v1 = callback_once<tester>();
static auto v2 = wrap<tester>();
```

Febbe:
@bhennion I suggest to also replace the rest of the `GCALLBACK`s in another PR, which is merged into master, after 1.2.2 is merged (1.2.1, when this lands in 1.2.1). We could also enforce, that the `wrap_for_once` requires `void` as return type and refactor all of those functions in a follow-up PR.

src/core/model/Document.cpp
```diff
@@ -32,8 +33,7 @@ Document::~Document() {
 
 void Document::freeTreeContentModel() {
     if (this->contentsModel) {
-        gtk_tree_model_foreach(this->contentsModel.get(),
-                               reinterpret_cast<GtkTreeModelForeachFunc>(freeTreeContentEntry), this);
+        gtk_tree_model_foreach(this->contentsModel.get(), xoj::util::wrap_for_once<freeTreeContentEntry>(), this);
```
bhennion, line 36: This is weird. The fact that G_SOURCE_REMOVE coincides with "do not break the foreach" is a coincidence.
```suggestion
        gtk_tree_model_foreach(this->contentsModel.get(), xoj::util::wrap<freeTreeContentEntry>(), this);
```

src/core/model/Document.cpp
```diff
@@ -290,8 +290,7 @@ auto Document::fillPageLabels(GtkTreeModel* treeModel, GtkTreePath* path, GtkTre
 
 void Document::updateIndexPageNumbers() {
     if (this->contentsModel) {
-        gtk_tree_model_foreach(this->contentsModel.get(), reinterpret_cast<GtkTreeModelForeachFunc>(fillPageLabels),
-                               this);
+        gtk_tree_model_foreach(this->contentsModel.get(), xoj::util::wrap_for_once<fillPageLabels>(), this);
```
bhennion, line 293:
```suggestion
        gtk_tree_model_foreach(this->contentsModel.get(), xoj::util::wrap<fillPageLabels>(), this);
```

src/util/include/util/Util.h
```diff
@@ -46,7 +47,20 @@ bool isFlatpakInstallation();
...
+    } else {
+        constexpr auto fn = +[](gpointer functor) -> int {
+            auto fun = static_cast<Fun*>(functor);
+            (*fun)();
+            return G_SOURCE_REMOVE;
+        };
+        gdk_threads_add_idle_full(priority, fn, new auto(std::forward<Fun>(callback)),
+                                  [](gpointer functor) { delete static_cast<Fun*>(functor); });
+    }
+}
```
bhennion, line 63: This is very nice!

src/util/include/util/glib_casts.h
```diff
@@ -0,0 +1,120 @@
...
+constexpr auto wrap_impl(Args... args, void* data) -> std::conditional_t<std::is_same_v<bool, R>, gboolean, R> {
+    if constexpr (std::is_same_v<void, R>) {
+        SrcFn(args..., static_cast<ST>(data));
+        return;
+    } else if constexpr (std::is_same_v<bool, R>) {
+        return SrcFn(args..., static_cast<ST>(data)) ? G_SOURCE_CONTINUE : G_SOURCE_REMOVE;
+    } else if (!std::is_same_v<void, R>) {
+        return SrcFn(args..., static_cast<ST>(data));
+    } else
+        static_assert(!std::is_same_v<R, void>, "Invalid return type");
```
bhennion, line 79:
```suggestion
    } else {
        return SrcFn(args..., static_cast<ST>(data));
    }
```

src/util/include/util/glib_casts.h
```diff
@@ -0,0 +1,120 @@
...
+}  // namespace detail
+
+template <auto SrcFn>
+constexpr auto wrap_for_once() -> auto {
+    using namespace detail;
+    using FTI = FunctionTraits<decltype(SrcFn)>;
+    return indices<FTI::ArgCount - 1>([&](auto... Is) {
+        return callback_once_impl<
+                SrcFn, typename FTI::LastArg,
+                std::remove_reference_t<decltype(std::get<Is>(std::declval<typename FTI::ArgTypes>()))>...>;
```
bhennion, line 91:
```suggestion
                std::tuple_element_t<Is, typename FTI::ArgTypes>...>;
```
is shorter and more readable, IMO

src/util/include/util/glib_casts.h
```diff
@@ -0,0 +1,120 @@
...
+    });
+}
+
+template <auto SrcFn>
+constexpr auto wrap() -> auto {
+    using namespace detail;
+    using FTI = FunctionTraits<decltype(SrcFn)>;
+    return indices<FTI::ArgCount - 1>([&](auto... Is) {
+        return wrap_impl<SrcFn, typename FTI::RetType, typename FTI::LastArg,
+                         std::remove_reference_t<decltype(std::get<Is>(std::declval<typename FTI::ArgTypes>()))>...>;
```
bhennion, line 101:
```suggestion
                         std::tuple_element_t<Is, typename FTI::ArgTypes>...>;
```

bhennion (changes requested):
Very nice!
One thing: could we add a
```
template <auto SrcFn> constexpr auto wrap_v = wrap<SrcFn>();
```
I feel the syntax `g_idle_add(wrap<myFunction>())` makes it less obvious that the parameter is a function.
Something like `g_idle_add(wrap_v<myFunction>)`, or even `g_idle_add(wrap<myFunction>)` feels more intuitive to me (that's obviously pretty subjective)

src/util/include/util/glib_casts.h
Febbe: My indennt behind this was, that this function must be evaluated as constexpr.

src/core/control/Control.cpp
```diff
@@ -2505,7 +2502,9 @@ void Control::loadMetadata(MetadataEntry md) {
     data->md = std::move(md);
     data->ctrl = this;
 
-    g_idle_add(reinterpret_cast<GSourceFunc>(loadMetadataCallback), data);
+    g_idle_add_full(
+            G_PRIORITY_DEFAULT_IDLE, xoj::util::wrap<loadMetadataCallback>(), data,
+            +[](gpointer data) { delete static_cast<MetadataCallbackData*>(data); });
```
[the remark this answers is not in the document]
Febbe: Added it, but replacing other occurrences should be done in another PR.

src/core/model/Document.cpp
Febbe: good catch!

src/core/control/jobs/Job.cpp
```diff
@@ -46,7 +49,7 @@ void Job::callAfterRun() {
     }
 
     this->ref();
-    this->afterRunId = gdk_threads_add_idle(reinterpret_cast<GSourceFunc>(Job::callAfterCallback), this);
+    this->afterRunId = gdk_threads_add_idle(xoj::util::wrap_for_once<Job::callAfterCallback>(), this);
```
bhennion, line 52:
```suggestion
    this->afterRunId = gdk_threads_add_idle(xoj::util::wrap_for_once_v<Job::callAfterCallback>, this);
```

src/core/control/tools/EditSelection.cpp
```diff
@@ -771,8 +772,7 @@ void EditSelection::moveSelection(double dx, double dy, bool addMoveUndo) {
 void EditSelection::setEdgePan(bool pan) {
     if (pan && !this->edgePanHandler) {
         this->edgePanHandler = g_timeout_source_new(1000 / PAN_TIMER_RATE);
-        g_source_set_callback(this->edgePanHandler, reinterpret_cast<GSourceFunc>(EditSelection::handleEdgePan), this,
-                              nullptr);
+        g_source_set_callback(this->edgePanHandler, xoj::util::wrap<EditSelection::handleEdgePan>(), this, nullptr);
```
bhennion, line 775:
```suggestion
        g_source_set_callback(this->edgePanHandler, xoj::util::wrap_v<EditSelection::handleEdgePan>, this, nullptr);
```

bhennion (approved):
Two nitpicks, otherwise LGTM!

bhennion:
You shoud probably squash the last 3 commits together before merging.

src/util/include/util/glib_casts.h
```diff
@@ -0,0 +1,120 @@
...
+    });
+}
+
+template <auto SrcFn>
+constexpr auto wrap_for_g_callback() -> GCallback {
+    return G_CALLBACK(wrap<SrcFn>());
+}
```
eldipa: Would consider add some docs on these wrappers? Specially about when to use them.

Febbe: Added some documentation.

src/core/control/jobs/Job.cpp
Febbe: done all

Febbe:
Will merge in 24h by myself (or fast-forward this if desired).

PR #5059, opened 2023-08-15
Feature lua add getSidebarPageNo/setSidebarPageNo

### LINKED ISSUE #4394: [lua Api] sidebarActions not reliably working

atticus-sullivan:
Adds two lua api functions to set and get the index of the currently active page of the sidebar.  This now is possible due to #5030 and will fix #4394 since now we can
- obtain the currently active page
- set the sidebar into the state needed for the sidebar action
- reset the sidebar to the state it was before calling this function

Maybe it would be nice to be able to work with strings instead of just the indices, but I'm not sure about this (and it would make things much harder)

TODOs:
- [x] write api documentation in the comments

### DIFF
--- a/src/core/gui/sidebar/Sidebar.cpp
+++ b/src/core/gui/sidebar/Sidebar.cpp
@@ -140,6 +140,10 @@ void Sidebar::selectPageNr(size_t page, size_t pdfPage) {
     }
 }
 
+size_t Sidebar::getNumberOfPages() { return this->pages.size(); }
+
+size_t Sidebar::getSelectedPage() { return this->currentPageIdx; }
+
 void Sidebar::setSelectedPage(size_t page) {
     this->visiblePage = nullptr;
 
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -64,7 +64,7 @@ extern "C" {
  *         They may also `return nil, errorMessage`. This behavior is reserved for
  *         things that are *expected to happen* (e.g. ressource is not
  *         available).
-*/
+ */
 
 /**
  * Renames file 'from' to file 'to' in the file system.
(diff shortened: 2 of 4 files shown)


src/core/gui/sidebar/Sidebar.cpp
```diff
@@ -140,6 +140,10 @@ void Sidebar::selectPageNr(size_t page, size_t pdfPage) {
     }
 }
 
+size_t Sidebar::getAmountPages() { return this->pages.size(); }
```
rolandlo, line 143: The name sounds a bit awkward to me.
```suggestion
size_t Sidebar::getNumberOfPages() { return this->pages.size(); }
```

rolandlo (commented):
Could you fix the `LayerActions` as well in this PR?

atticus-sullivan:
Fixed the LayerActions plugin

src/core/plugin/luapi_application.h
```diff
@@ -396,6 +396,53 @@ static int applib_sidebarAction(lua_State* L) {
...
+ *
+ * Example: app.setSidebarPageNo(2) -- sets the sidebar-page to preview Layer
+ */
+static int applib_setSidebarPageNo(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Sidebar* sidebar = plugin->getControl()->getSidebar();
+
+    // Discard any extra arguments passed in
+    lua_settop(L, 1);
+    if (!lua_isnumber(L, 1)) {
```
rolandlo, line 426: It should be an integer
```suggestion
    if (!lua_isinteger(L, 1)) {
```

rolandlo:
One could argue that counting (page numbers) should start with 1 in Lua. What do you think about that?

atticus-sullivan:
> One could argue that counting (page numbers) should start with 1 in Lua. What do you think about that?

I think this is a valid point. It also seems to be more consistent with setting the background page or setting the currently active page in the document (at least I think so by skimming trough the functions). Also indexing the layer seems to start with 1 (but one could argue the background layer is layer 0).

Long story short, I also think sidebar-pages should start at 1.

rolandlo (approved):
LGTM!

Anything you want to change? Merging in 24 hours otherwise.

rolandlo:
I forgot to say: please squash your commits into 2-3 commits (one separate for the plugin).

atticus-sullivan:
Let me know if I shall squash further (sidebar-pages and luaAPI maybe?).

rolandlo:
> Let me know if I shall squash further (sidebar-pages and luaAPI maybe?).

No, I think it's fine. Merging in 24 hours if no objections are raised. In case merge conflicts with #5029 appear when merging that PR, I will ask you to rebase and fix the conflicts.

rolandlo:
Thanks for this contribution as well and for fixing the LayersAction plugin with it.

PR #5061, opened 2023-08-16
luaAPI refactoring: get elements from selection/layer helper function

atticus-sullivan:
Really rough draft for this so we have a place for discussion (like discussed [here](https://github.com/xournalpp/xournalpp/pull/5029#discussion_r1295293761).

Open questions:
- [x] how to handle the `luaL_error`s. In my opinion, a helper function should not directly error but return the error so that the calling function can decide how to handle it.

TODO:
- [ ] move helper function to another place (putting it right on top is just temporary)
- [x] change name?

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -57,6 +57,32 @@ extern "C" {
 #include <lualib.h>   // for luaL_openlibs
 }
 
+
+static std::tuple<std::optional<std::string>, std::vector<Element*>> getElementsFromHelper(Control* control,
+                                                                                           const std::string& type) {
+    std::vector<Element*> elements = {};
+    if (type == "layer") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            control->clearSelection();  // otherwise texts in the selection won't be recognized
+        }
+        elements = xoj::refElementContainer(control->getCurrentPage()->getSelectedLayer()->getElements());
+    } else if (type == "selection") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            elements = sel->getElements();
+        } else {
+            return std::make_tuple(std::make_optional("There is no selection"), elements);
+        }
+    } else {
+        std::stringstream err_msg;
+        err_msg << "Unknown argument (" << type << ") for getting selection";
+        return std::make_tuple(std::make_optional(err_msg.str()), elements);
+    }
+    return std::make_tuple(std::nullopt, elements);
+}
+
+
 /*
  * Code conventions:
  *     Error handling:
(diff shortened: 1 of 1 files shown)


atticus-sullivan:
We could return something tuple like `success:bool, elements:vector, error-string:str` or maybe you think doing a `luaL_error` in the helper function isn't so bad after all.

Just dump your thoughts here :)

rolandlo:
> We could return something tuple like `success:bool, elements:vector, error-string:str` or maybe you think doing a `luaL_error` in the helper function isn't so bad after all.
> 
> Just dump your thoughts here :)

I guess doing a `luaL_error` in the helper function has disadvantages, but we have to see, how that exactly behaves. If that is so, one could also return a pair `error-string : std::optional<string>, elements : std::vector`. If there is an error string, the returned vector is empty. If there is no error-string, the returned vector contains the elements.

atticus-sullivan:
> I guess doing a `luaL_error` in the helper function has disadvantages, but we have to see, how that exactly behaves.

It feels just a bit like a `exit(-1)` in (almost) any function different than the main-function. In the end, I **think** (not sure though) there should be no direct trouble with lua (cleans the stack anyhow). It's just not nice as it takes away control over the control-flow (e.g. imagine a function which calls our helper in a loop and the loop shall continue even if one cycle failed).
But tbh in `handleUndoRedoActionHelper` we also just did `luaL_error`. I think back then I didn't want to return a string and/or _wanted to use the formatting capabilities of `luaL_error`_.

---

Oh and `getSelectionLayerHelper` was just the simplest name I came up as I needed to name the function somehow (struggled a bit with the *or* in *selection or layer* and just `getElements` is a bit too brief for this helper function I think) . I'm not really happy with that name so suggestions are welcome.

rolandlo:
> It feels just a bit like a `exit(-1)` in (almost) any function different than the main-function. In the end, I **think** (not sure though) there should be no direct trouble with lua (cleans the stack anyhow). It's just not nice as it takes away control over the control-flow (e.g. imagine a function which calls our helper in a loop and the loop shall continue even if one cycle failed).

What is the difference in the error message that the user gets displayed?

rolandlo:
> Oh and `getSelectionLayerHelper` was just the simplest name I came up as I needed to name the function somehow (struggled a bit with the _or_ in _selection or layer_ and just `getElements` is a bit too brief for this helper function I think) . I'm not really happy with that name so suggestions are welcome.

Maybe `getElementsFromHelper`?

atticus-sullivan:
> What is the difference in the error message that the user gets displayed?

I think that's not what I mean. I thought of something like `app.apiFunction{{elements="selection", action="mirror"}, {elements="layer", action="move"}}`. So that the api function calls our new helper multiple times and even if the first action failed somehow it should continue with the next action.

I know the example is not really a good one and with the current behavior of clearing the selection when obtaining elements from the whole layer, we should never write a function that way -> maybe I'm a bit too pedantic in this case and `luaL_error` is fine in that place.

rolandlo:
I understand that you were talking about something else. It's a valid point. How `luaL_error`s are reported to the user (what line number and error message are displayed) is another point, but of importance, too.

atticus-sullivan:
Yes that's right, but the line number will be the same if the error is thrown in the helper or in the function calling the helper as it will be the same line number in the lua code.

atticus-sullivan:
I decided to go with your suggestion of returning `(optional<string>, vector<elements>)`. The only drawback is that the returned error message doesn't contain the "mode" that was (wrongly) passed to the function, but I think the (unparametrized) message is enough for this case.

So from my side this is ready for review.

rolandlo (approved):
LGTM

src/core/plugin/luapi_application.h
```diff
@@ -56,6 +56,30 @@ extern "C" {
...
+        elements = xoj::refElementContainer(control->getCurrentPage()->getSelectedLayer()->getElements());
+    } else if (type == "selection") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            elements = sel->getElements();
+        } else {
+            return std::make_tuple(std::make_optional("There is no selection"), elements);
+        }
+    } else {
+        return std::make_tuple(std::make_optional("Unknown argument for getting selection"), elements);
```
rolandlo, line 77: You could add the `type` to the string here.

atticus-sullivan: But how would you build the string? (we're just passing the string so we don't have the `lua_...` formatting capabilities at that point)

rolandlo: E.g. like this:
```c++
        std::stringstream msg;
        msg << "Unknown argument for getting selection: " << type;
        return std::make_tuple(std::make_optional(msg.str()), elements);
```

atticus-sullivan: :+1: Thanks.

(Somehow I tried to circumvent this dynamic string building at first, but I guess when dealing with user feedback we should do this.)

atticus-sullivan:
Ready from my side (already rebased and squashed)

rolandlo:
Ok, merging in 24 hours if no objections are raised.

rolandlo:
Thanks @atticus-sullivan for the refactoring. It looks way cleaner now.

PR #5078, opened 2023-08-21
Make PageTemplateDialog gtk4-compatible

bhennion:
This concludes the series of PR's about PageType-related widgetry. Based on #4508

- [x] Merge #4508 and rebase

### DIFF
--- a/src/core/gui/dialog/PageTemplateDialog.cpp
+++ b/src/core/gui/dialog/PageTemplateDialog.cpp
@@ -13,40 +13,63 @@
 #include "control/pagetype/PageTypeHandler.h"  // for PageTypeInfo, PageType...
 #include "control/settings/Settings.h"         // for Settings
 #include "control/stockdlg/XojOpenDlg.h"       // for XojOpenDlg
+#include "gui/Builder.h"                       // for Builder
 #include "gui/PopupWindowWrapper.h"            // for PopupWindowWrapper
 #include "gui/menus/popoverMenus/PageTypeSelectionPopoverGridOnly.h"
-#include "gui/widgets/PopupMenuButton.h"  // for PopupMenuButton
-#include "model/FormatDefinitions.h"      // for FormatUnits, XOJ_UNITS
-#include "model/PageType.h"               // for PageType
-#include "util/Color.h"                   // for GdkRGBA_to_argb, rgb_t...
-#include "util/PathUtil.h"                // for fromGFilename, readString
-#include "util/i18n.h"                    // for _
+#include "gui/toolbarMenubar/ToolMenuHandler.h"
+#include "model/FormatDefinitions.h"  // for FormatUnits, XOJ_UNITS
+#include "model/PageType.h"           // for PageType
+#include "util/Color.h"               // for GdkRGBA_to_argb, rgb_t...
+#include "util/PathUtil.h"            // for fromGFilename, readString
+#include "util/i18n.h"                // for _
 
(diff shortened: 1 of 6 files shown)


src/core/gui/dialog/PageTemplateDialog.cpp
```diff
@@ -13,40 +13,63 @@
...
-    g_signal_connect(get("btLoad"), "clicked",
-                     G_CALLBACK(+[](GtkToggleButton* togglebutton, PageTemplateDialog* self) { self->loadFromFile(); }),
-                     this);
+    g_signal_connect_swapped(builder.get("btSave"), "clicked",
+                             G_CALLBACK(+[](PageTemplateDialog* self) { self->saveToFile(); }), this);
 
-    g_signal_connect(get("btSave"), "clicked",
-                     G_CALLBACK(+[](GtkToggleButton* togglebutton, PageTemplateDialog* self) { self->saveToFile(); }),
-                     this);
+    g_signal_connect_swapped(builder.get("btCancel"), "clicked", G_CALLBACK(gtk_window_close), this->window.get());
```
rolandlo, line 65: It looks more consistent to use the `getWindow()` method.
```suggestion
    g_signal_connect_swapped(builder.get("btCancel"), "clicked", G_CALLBACK(gtk_window_close), this->getWindow());
```

src/core/gui/dialog/PageTemplateDialog.cpp
```diff
@@ -13,40 +13,63 @@
...
+    g_signal_connect_swapped(builder.get("btOk"), "clicked", G_CALLBACK(+[](PageTemplateDialog* self) {
+                                 self->saveToModel();
+                                 self->settings->setPageTemplate(self->model.toString());
+                                 self->toolMenuHandler->setDefaultNewPageType(self->model.getPageInsertType());
+                                 gtk_window_close(self->window.get());
```
rolandlo, line 70: Same here:
```suggestion
                                 gtk_window_close(self->getWindow());
```

ui/pageTemplate.glade
```diff
@@ -1,308 +1,301 @@
...
           </packing>
         </child>
       </object>
     </child>
-    <action-widgets>
-      <action-widget response="2">btCancel</action-widget>
-      <action-widget response="1">btOk</action-widget>
-    </action-widgets>
   </object>
 </interface>
```
rolandlo, line 301: The GtkBox containing the buttons and the GtkBox containing everything don't have an ID. Not sure if that could ever be an issue, but as I recall everything in our Glade files had an ID so far. So maybe add one here, too.

rolandlo (commented):
It works fine according to my tests. I have only some minor nitpicks about the code.

ui/pageTemplate.glade
bhennion: I'm not sure about that. I think that naming everything may be a mistake actually: I haven't looked under the hood, but I'd assume that GtkBuilder has a map id->GObject* of named objects. Uselessly giving widgets an id would just increase the map's size thus slowing down (even so slightly) the name-matching searches.
It's probably not worth removing the unused id, but why not avoid adding more?

rolandlo: Your arguments make sense to me. So let's avoid adding unused id's in future.

rolandlo (approved):

bhennion:
Thanks for the review! Merging in 24 hours.

PR #5079, opened 2023-08-21
Make plugin dialogs GTK4-compatible

bhennion:
The main commit is based on @rolandlo's https://github.com/xournalpp/xournalpp/pull/4796#issuecomment-1685296876
I kept the old mecanism, adding a deprecation warning so that user-end plugins have time to adapt.

The second commit is only for testing purposes and should be removed before merging

- [ ] Remove example


src/core/plugin/luapi_application.h
```diff
@@ -218,6 +220,48 @@ static int applib_msgbox(lua_State* L) {
...
+
+        buttons.emplace_back(buttonText, index);
+    }
+
+    std::string cb = luaL_optstring(L, 3, "");
+    g_message("applib_openDialog: Callback name: %s", cb.c_str());
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+
+    const std::string& pluginName = plugin->getName();
+    bool error = false;
```
rolandlo, line 253: We should probably add an optional parameter `error` to `openDialog`, which defaults to false.

src/core/plugin/luapi_application.h
```diff
@@ -218,6 +220,48 @@ static int applib_msgbox(lua_State* L) {
...
+    while (lua_next(L, 2) != 0) {
+        int index = static_cast<int>(lua_tointeger(L, -2));
+        const char* buttonText = luaL_checkstring(L, -1);
+        lua_pop(L, 1);
+
+        buttons.emplace_back(buttonText, index);
+    }
+
+    std::string cb = luaL_optstring(L, 3, "");
+    g_message("applib_openDialog: Callback name: %s", cb.c_str());
```
rolandlo, line 249: I would remove that message
```suggestion
```

src/util/XojMsgBox.cpp
```diff
@@ -131,17 +137,17 @@ void XojMsgBox::showPluginMessage(const std::string& pluginName, const std::stri
 auto XojMsgBox::askPluginQuestion(const std::string& pluginName, const std::string& msg,
                                   const std::vector<Button>& buttons, bool error) -> int {
     /*
-     * Todo(gtk4): This should be a simple wrapper around XojMsgBox::askQuestion(), but we need to figure out how to
-     * handle non-blocking dialogs in lua
+     * Todo(gtk4): Remove this function entirely.
      */
-    auto header = (error ? std::string("<b>Error in </b>") : "") + std::string("Xournal++ Plugin «") + pluginName + "»";
+    std::string header = "<i>Warning: The plugin interface function msgbox() is deprecated and will soon be removed. "
+                         "Please adapt you plugin to use the function openDialog() instead</i>\n\n";
```
rolandlo, line 143:
```suggestion
                         "Please adapt your plugin to use the function openDialog() instead</i>\n\n";
```

plugins/Example/main.lua
```diff
@@ -15,6 +16,16 @@ end
 
 -- Callback if the menu item is executed
 function exampleCallback()
+  app.openDialog("Test123", {[1] = "Yes", [2] = "No"}, "dialogCallback");
```
rolandlo, line 19: I would actually be better to use it that way (to avoid having the order of the buttons arbitrary)
```suggestion
  app.openDialog("Test123", {"Yes", "No"}, "dialogCallback");
```

rolandlo (changes requested):
The plugins using `app.msgbox` we ship  should be adapted to using `app.openDialog`
Let me know if I should do it.

plugins/LayerActions/main.lua
```diff
@@ -13,28 +13,40 @@ function clone()
...
-    app.msgbox("Next page has no pdf background. ", {[1] = "OK"})
+    app.openDialog("Next page has no pdf background. ", {"OK"}, "", true)
     return
   end
 
-  local nextPdfPage = docStructure["pages"][nextPage]["pdfBackgroundPageNo"]
-  print(nextPdfPage)
-
   local isAnnotated = docStructure["pages"][nextPage]["isAnnotated"]
   if isAnnotated then
```
bhennion, line 27: This test does not work (it does not work on master either). @rolandlo do you know how to fix that?

bhennion:
> The plugins using `app.msgbox` we ship should be adapted to using `app.openDialog` Let me know if I should do it.

Done

plugins/LayerActions/main.lua
rolandlo: Yes, I see the problem. There is a typo in 
https://github.com/xournalpp/xournalpp/blob/a6ba9797f2d0a71508b4efcf820f24ac8681a9be/src/core/plugin/luapi_application.h#L1933-L1934
and 
https://github.com/xournalpp/xournalpp/blob/a6ba9797f2d0a71508b4efcf820f24ac8681a9be/src/core/plugin/luapi_application.h#L1977-L1978

"Annoated" should be "Annotated"

rolandlo: It's a regression from https://github.com/xournalpp/xournalpp/commit/45062283113b7db70d5981d0223bf2d49c8beaac by the way.

rolandlo (approved):
LGTM, just a minor nitpick.

bhennion:
Thanks for the assist and review, @rolandlo!

PR #5139, opened 2023-09-12
Add marker opacity in select PDF Text options

### LINKED ISSUE #4691: Select Linear Text Hightlighter with more opacity
When I use the Hightlighter the color is stronger than the Select Linear Text Hightlighter in the pdf which is very weak.
Windows 11. xournalpp-1.1.3+dev-nightly.20230302-windows

a little more opacity

**Additional context**
Captura de tela 2023-03-02 191444

jerdoe:
This PR adds a menu to specify Marker Opacity for 'Select PDF Text' Tool instead of using the hardcoded value of 60, which could look too faint for some people.

Marker Opacity can be defined using Tools -> Select PDF Text Options -> Marker Opacity.
Fixes #4691


rolandlo (changes requested):
Thanks for the PR. I like the idea of a customizable opacity for the PDF text selection. However the implementation has some issues and needs to be improved:
- Instead of using the `ToolType` I would suggest to use a new enum `FillType` (with currently 3 values, say `FILL_PEN`, `FILL_HIGHLIGHTER`, `FILL_PDF_TEXT_MARKER`). First of all for most tool types opacity doesn't make sense and secondly the two different pdf text selection modes share the same alpha value.
- In `Control::actionPerformed` you should not pass the active tool, but the enum value that corresponds to the action (e.g. `FILL_PEN` for `ACTION_TOOL_PEN_FILL_OPACITY`)
- The boolean `pen` should be removed from the `FillOpacityDialog` and replaced by the `FillType`
- The callback from `xoj::popup::PopupWindowWrapper<xoj::popup::FillOpacityDialog>` should not use the boolean `pen`, but the `FillType`. The toolHandler should not be used for querying the ToolType or FillType.
- In `xoj::popup::FillOpacityDialog::setPreviewImage` the `pen` property for determining whether to paint an extra rectangle should be replaced by `fillType == FILL_PEN`. The pdf text marker is more like the highlighter than the pen.

jerdoe:
Thank you very much for having considered my PR and for the very relevant comments.
Feel free to request any other changes or improvements.

jerdoe:
> Thanks for the PR. I like the idea of a customizable opacity for the PDF text selection. However the implementation has some issues and needs to be improved:
> 
> * Instead of using the `ToolType` I would suggest to use a new enum `FillType` (with currently 3 values, say `FILL_PEN`, `FILL_HIGHLIGHTER`, `FILL_PDF_TEXT_MARKER`). First of all for most tool types opacity doesn't make sense and secondly the two different pdf text selection modes share the same alpha value.
> * In `Control::actionPerformed` you should not pass the active tool, but the enum value that corresponds to the action (e.g. `FILL_PEN` for `ACTION_TOOL_PEN_FILL_OPACITY`)
> * The boolean `pen` should be removed from the `FillOpacityDialog` and replaced by the `FillType`
> * The callback from `xoj::popup::PopupWindowWrapper<xoj::popup::FillOpacityDialog>` should not use the boolean `pen`, but the `FillType`. The toolHandler should not be used for querying the ToolType or FillType.
> * In `xoj::popup::FillOpacityDialog::setPreviewImage` the `pen` property for determining whether to paint an extra rectangle should be replaced by `fillType == FILL_PEN`. The pdf text marker is more like the highlighter than the pen.

Thank you very much for having considered my PR and for the relevant comments.
Feel free to request any other changes or improvements.

src/core/gui/PdfFloatingToolbox.cpp
```diff
@@ -113,7 +113,8 @@ void PdfFloatingToolbox::userCancelSelection() {
 }
 
 void PdfFloatingToolbox::highlightCb(GtkButton* button, PdfFloatingToolbox* pft) {
-    pft->createStrokes(PdfMarkerStyle::POS_TEXT_MIDDLE, PdfMarkerStyle::WIDTH_TEXT_HEIGHT, 60);
+    int markerOpacity = pft->theMainWindow->getXournal()->getControl()->getToolHandler()->getSelectPDFTextFill();
```
rolandlo, line 116: A bit more direct:
```suggestion
    int markerOpacity = pft->theMainWindow->getControl()->getToolHandler()->getSelectPDFTextFill();
```

rolandlo (approved):
LGTM! Thanks for cleaning this up so quickly. I have only a very minor nitpick.

bhennion (changes requested):
First, thanks for the PR!

I have some concerns about code readability though:

1. This mixes two notions that are very different (at least in the way they are implemented): the alpha value of a stroke (made via the Pdf text highlighter) with the alpha value of a stroke's filling.
2. The enum FillType does not represent various types of filling. I'd call a filling type something like "plain", "stripes" or whatever pattern a stroke could be filled with. Here, FillType really just represents the tool involved in producing the stroke.

jerdoe:
> First, thanks for the PR!
> 
> I have some concerns about code readability though:
> 
> 1. This mixes two notions that are very different (at least in the way they are implemented): the alpha value of a stroke (made via the Pdf text highlighter) with the alpha value of a stroke's filling.
> 2. The enum FillType does not represent various types of filling. I'd call a filling type something like "plain", "stripes" or whatever pattern a stroke could be filled with. Here, FillType really just represents the tool involved in producing the stroke.

It's all my pleasure!
I feel your concerns. Do you think we should create a specific property in ToolBase to store the pdf text marker opacity or should we extend the approach and consider something like a StrokeAlpha property ?

bhennion:
I guess using the Color's alpha channel is what'd make the most sense. Right now, this channel is not really used anywhere.

jerdoe:
> I guess using the Color's alpha channel is what'd make the most sense. Right now, this channel is not really used anywhere.

I've started reworking on it but a question has emerged:
In my new implementation (not yet pushed), pdf marker opacity can be set through the tool settings menu and is stored in the alpha property of Color of the tool 'Select PDF Text...'
Is it planned to add alpha definition through the palette and to store it for each individual color?

When supported, what is the expected behaviour if a user switches color while using pdf text selection. Should it pick up the individual alpha of the selected color or the one defined in the tool settings menu ?

bhennion:
I think the color chooser dialog and the color buttons should not alter the opacity. Using the already coded opacity slider for that makes more sense I think.

I realized that under the hood, the strokes created by the PdfFloatingToolbox are in fact "filled" strokes, but only as a workaround in order to control the opacity. (see `PdfFloatingToolbox::createStrokes()`). At some point, we should aake the highlighter strokes opacity variable (via the Color::alpha channel) instead of using the fixed StrokeView::HIGHLIGHTER_OPACITY (and remove this workaround).

bhennion:
Maybe it'd be best to keep this PR somewhat small (maybe just rename the FillType enum or so) and leave a deeper refactor for a future PR.

src/core/gui/dialog/SelectOpacityDialog.cpp
```diff
@@ -7,40 +7,48 @@
...
+        addBorderForPreviewImage(addBorderForPreviewImage), callbackOnNewAlphaConfirmed(callbackOnNewAlphaConfirmed) {
     Builder builder(gladeSearchPath, UI_FILE);
     this->window.reset(GTK_WINDOW(builder.get(UI_DIALOG_NAME)));
 
     previewImage = GTK_IMAGE(builder.get("imgPreview"));
     alphaRange = GTK_RANGE(builder.get("scaleAlpha"));
 
-    gtk_range_set_value(alphaRange, byteToPercent(alpha));
+    gtk_label_set_label(GTK_LABEL(builder.get("label1")),
+                        FS(_F("<b>{1} opacity settings</b>\nSelect opacity for {1}") % opacityForLabel).c_str());
```
jerdoe, line 35: Is it designed to be compatible and to support easy translation?

jerdoe:
Do you have any insights into why it doesn't pass the clang-format test? I've checked it on my local repository by using the following commands:
`git reset --soft HEAD~1
git-clang-format`
and the output states "clang-format did not modify any files."

jerdoe:
Ready for review. Feel free for any clarifications, changes or improvements.

src/core/control/ToolHandler.cpp
```diff
@@ -31,6 +32,24 @@ ToolHandler::ToolHandler(ToolListener* stateChangeListener, ActionHandler* actio
...
 }
 
+class ToolSelectPDFText: public Tool {
+public:
+    ToolSelectPDFText(std::string name, ToolType type, Color color):
+            Tool(name, type, color, TOOL_CAP_COLOR | TOOL_CAP_RULER, std::nullopt) {}
+
+    ~ToolSelectPDFText(){};
+
+    void setColor(Color color) {
```
bhennion, line 42: Could you mark those as `override`?

src/core/gui/dialog/SelectOpacityDialog.cpp
```diff
@@ -7,23 +7,58 @@
 #include <glib.h>         // for gdouble
 
 #include "gui/Builder.h"
+#include "util/Stacktrace.h"
+#include "util/i18n.h"
 #include "util/raii/CairoWrappers.h"
 #include "util/raii/GObjectSPtr.h"
+using std::string;
```
bhennion, line 14: We are trying to avoid `using` statements like those in newly written code
```suggestion
```

src/core/gui/dialog/SelectOpacityDialog.cpp
```diff
@@ -7,23 +7,58 @@
...
-xoj::popup::FillOpacityDialog::FillOpacityDialog(GladeSearchpath* gladeSearchPath, int alpha, bool pen,
-                                                 std::function<void(int, bool)> callback):
-        pen(pen), callback(callback) {
+static inline void buildLabel(Builder& builder, OpacityFeature opacityFeature) {
+    // Used to set the label of the dialog in the form of:
+    // <b>{toolOptionsDesc}</b>\n
+    // Select opacity for: {opacityFeatureDesc}
+    string toolOptionsDesc;
+    string selectOpacityFor = _("Select opacity for: ");
+    string opacityFeatureDesc;
```
bhennion, line 30:
```suggestion
    std::string toolOptionsDesc;
    std::string selectOpacityFor = _("Select opacity for: ");
    std::string opacityFeatureDesc;
```

bhennion (commented):
The code loogs fine to me (besides the following nitpicks).
I'll try to find time to test it soon

jerdoe:
> The code loogs fine to me (besides the following nitpicks). I'll try to find time to test it soon

Thank you for your prompt code review and for highlighting its flaws. Your feedback is greatly appreciated.
No rush at all, take your time to test it whenever you're comfortable.

PR #3587, opened 2021-11-18
Change autosave dir to ~/.cache/xournalpp/autosaves

fabian-thomas:
- updated path used in code
- updated translations

Closes #3348

In the original issue there were two directories for autosaves discussed. `~/.local/share/xournalpp/autosave` and `~/.cache/xournalpp/autosave`. I decided to go with the cache directory since this would be the first one I would search for autosaves.

I'm pretty sure that I did something wrong with the translation files so please explain me quickly what I've should done with those. Should I've just changed xournalpp.pot and then ran make?

PR #5140, opened 2023-09-13
Fix shouldFileOpen

rolandlo:
In #3587 the location for autosave files was moved from the .config folder to the .cache folder


rolandlo:

bhennion (approved):

Issue #1683, opened 2019-12-23
Be able to place duplicate buttons in config file


abondis:
**Is your feature request related to a problem? Please describe.**
I like the floating toolbox, but I also like to select colors on the side. Having the same color showed in the sidebar and the floating toolbox would be helpful.

Config file does not seem to permit having the same tool declared in different toolbars

**Describe the solution you'd like**
Make the config file accept/work with duplicated tools declarations

**Describe alternatives you've considered**
I don't see any...

**Additional context**
I tried to add the red color selector in the side toolbar and the floating toolbox. Doing so from the UI is not possible since it just moves the tool from toolbar to toolbar. But it seems doable since by default the view with the floating toolbox lets select red, and red is also on the sidebar.
Using the config file I had no luck either, but if it is already doable don't hesitate to point me towards the doc.

jvwowd:
I was able to do this with colors by directly editing the toolbar.ini file (plaintext file and on Linux it is ~/.xournalpp/toolbar.ini). If you go to View -> Toolbars -> Manage and copy the default Floating Toolbox, then there will be an entry for this copy in the toolbar.ini file which you can edit accordingly.

I was unable to use this approach to have duplicates of other toolbar buttons though.

felurx:
I would also really like to be able to have, for example, the pen, eraser and highlighter both in my toolbar and floating toolbox. I would just like to ask, why is implementing this considered hard? @Technius Just curious.

Technius:
It's not so much hard as it is a lot of work. Need to ensure that each tool widget has consistent states, that there aren't any deadlocks of any sort, etc.

bhennion:
The task will simplify a lot once #5154 is merged.

PR #5154, opened 2023-09-16
Use GAction in Toolbar items

bhennion:
This builds on #4372 and refactors every ToolItem to rely on the new ActionDatabase.
The old ActionHandler is removed and a backward compatibility layer is added to avoid breaking plugins.

Overall, the new infrastructure is much simpler and IMO cleaner: almost every interaction between the UI (drawing widget aside) and Control is done via GAction and the ActionDatabase.

#5114 is now based on this PR.


bhennion:
This is now ready for review.
Note also that implementing #1683 will be very easy now (if anyone wants to do that).

Febbe:
Also, the hotkey feature might be implementable now.

Febbe:
Didn't see, that you implemented `std::to_underlying` in the std namespace. Probably I am too pedantic here, since it works, but implementing anything in the std namespace is undefined behav.. 
Also, you don't use [version testing macros](https://en.cppreference.com/w/cpp/feature_test#Library_features), when I do this in my projects, I at least make sure, that the functionality is not implemented already.
It's better to use `xoj::` or `xoj::util::` here.

```
#if !defined(__cpp_lib_to_underlying) 
namespace std {
template <class Enum>
constexpr std::underlying_type_t<Enum> to_underlying(Enum e) noexcept {
    return static_cast<std::underlying_type_t<Enum>>(e);
}
}  // namespace std
#endif
```

test/unit_tests/model/ColorPaletteTest.cpp
```diff
@@ -64,11 +64,11 @@ TEST(ColorPalette, testRainbowLoad) {
     EXPECT_EQ((size_t)6, palette.size());
     EXPECT_EQ(palette.getColorAt(4).getName(), std::string{"Royal Blue"});
     EXPECT_EQ(palette.getColorAt(4).getIndex(), (size_t)4);
-    EXPECT_EQ(palette.getColorAt(4).getColorU16().alpha, 0U);
+    EXPECT_EQ(palette.getColorAt(4).getColorU16().alpha, 0xFFFFU);
     EXPECT_EQ(palette.getColorAt(4).getColorU16().red, 0U);
     EXPECT_EQ(palette.getColorAt(4).getColorU16().green, 0x4D4DU);
     EXPECT_EQ(palette.getColorAt(4).getColorU16().blue, 0xFFFFU);
-    EXPECT_EQ(palette.getColorAt(4).getColor(), Color{0x004DFF});
+    EXPECT_EQ(palette.getColorAt(4).getColor(), Color{0xFF004DFF});
```
Febbe, line 71:
```suggestion
    EXPECT_EQ(palette.getColorAt(4).getColor(), Color{0xFF004DFFU});
```

Febbe (changes requested):

Febbe:
Most of the comments are nits, but I marked nearly everything I stumbled on, others are only marked, because you refactored some parts of the code next to the problem, and it would be good to have them refactored too.

src/core/model/XojPage.cpp
```diff
@@ -15,6 +15,7 @@ XojPage::XojPage(double width, double height, bool suppressLayerCreation): width
     if (!suppressLayerCreation) {
         // ensure at least one valid layer exists
         this->addLayer(new Layer());
+        this->currentLayer = 1;
```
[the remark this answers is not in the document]
bhennion: ?
PageHandler only has a default constructor, no `PageHandler(PageRef)`

src/core/gui/toolbarMenubar/DrawingTypeComboToolButton.h
```diff
@@ -0,0 +1,76 @@
...
+        Entry() = default;
+        Entry(std::string name, std::string icon, const ActionDatabase& db, Action a);
+        std::string name;
+        std::string icon;
+        ActionRef gAction;
+        std::string fullActionName;
+    };
+    DrawingTypeComboToolButton(std::string id, IconNameHelper& icons, const ActionDatabase& db);
+
+    ~DrawingTypeComboToolButton() override;
```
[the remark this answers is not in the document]
bhennion: Not this one. The destructor disconnects the signals.

src/core/gui/toolbarMenubar/DrawingTypeComboToolButton.h
```diff
@@ -0,0 +1,76 @@
...
+    enum class Type : size_t {
+        RECTANGLE,
+        ELLIPSE,
+        ARROW,
+        DOUBLE_ARROW,
+        LINE,
+        COORDINATE_SYSTEM,
+        SPLINE,
+        SHAPE_RECOGNIZER,
+        _COUNT
```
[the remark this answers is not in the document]
bhennion: Hmm. Too bad.
Is there a naming convention that would make clear the enum value `COUNT` (or whatever) should not be interpreted as an actual value by the devs? Would `COUNT_` do?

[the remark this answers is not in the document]
Febbe: This is a common technique, so no experienced c/c++ dev should misinterpret this. But you can change it to something like ENUMERATOR_COUNT or just use COUNT.

src/core/model/XojPage.cpp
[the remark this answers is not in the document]
Febbe: Mhh, somehow clangd had the opinion the copy/move constructor is default generated. Does it make sense, to make PageHandler copy and moveable?

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -269,40 +244,32 @@ void ToolMenuHandler::load(ToolbarData* d, GtkWidget* toolbar, const char* toolb
...
 
     control->getActionDatabase()->enableAction(Action::SHOW_SIDEBAR, !disabled);
 }
 
-void ToolMenuHandler::addToolItem(AbstractToolItem* it) { this->toolItems.push_back(it); }
-
-void ToolMenuHandler::registerMenupoint(GtkWidget* widget, ActionType type, ActionGroup group) {
-    this->menuItems.push_back(new MenuItem(listener, widget, type, group));
-}
+void ToolMenuHandler::addToolItem(AbstractToolItem* it) { this->toolItems.emplace_back(it); }
```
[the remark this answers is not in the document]
bhennion: This is always used as `addToolItem(new Somthing())`. How about
```
template<class tool_item, class... Args>
tool_item& emplaceItem(Args&& ...args) {
    return static_cast<tool_item&>(*toolItems.emplace_back(std::make_unique<tool_item>(std::forward<Args>(args)...)));
}
```

src/core/model/XojPage.cpp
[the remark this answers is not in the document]
bhennion: Copiable, no (the listeners only have one handler). Moveable, certainly.

Febbe (approved):
Regarding the code, it looks good, but we should test it thoroughly.

rolandlo:
> Regarding the code, it looks good, but we should test it thoroughly.

I have done some testing on Linux and will do more testing on MacOS and Windows.

- One positive thing I have noticed: In a multi-monitor setup, where Xournal++ is fullscreen or takes most space on one monitor, there were issues with popups (like when creating a new layer from the dropdown) that would be displayed where nobody could see it (e.g. below the edge of the screen). This is no longer an issue.
- One thing I noticed, which may be intended or not: When clicking on a popover menu item (say "Create new layer" in the layer popover or "dashed" on the pen popover) the popover is not closed automatically any more. It closes when I click on the canvas. It's different for the drawing type combo and the PDF text selection combo, where the popover still closes automatically.
- One negative thing (which in this case is already on master, but it is related to the popovers):
When I have say the eraser selected and change the line style for the pen (without switching to the pen), I can only do it once. The second time I try to do it, the click is simply ignored.

I have also found a regression in the "Migrate Font sizes" plugin, but it's related to the callback not knowing about the local variable "factor" and not to this PR. There is also an issue with the "Middle Mouse" combo in the preferences and the "stroke style" which can only be applied to the Right Mouse button, but not for stylus or default tool (any more?!). That's also in master.

rolandlo:
On MacOS the global menu looks nice. There is even a nice search integrated into menu Help.
Bildschirmfoto 2023-10-05 um 18 33 57

However the global menu cannot be used on startup. Nothing happens when clicking on the individual menus. I first have to change focus to a different window (say terminal) and then change back focus to Xournal++ again.
I get some warnings, but they do not appear any more after deleting/renaming the settings.xml file.

Other things I have noticed (not only on MacOS, but also on Linux): 
- The output device for audio recordings is saved wrongly (it takes the next entry). The input device is saved correctly.
- When pressing a toolbar button like "Setsquare" (or "Vertical space") the focus remains on the toolbar widget. So I cannot use the arrow keys (etc) to move the setsquare around, without first clicking onto the canvas.

bhennion:
> I have done some testing on Linux and will do more testing on MacOS and Windows.

Thanks a lot!!

> * One thing I noticed, which may be intended or not: When clicking on a popover menu item (say "Create new layer" in the layer popover or "dashed" on the pen popover) the popover is not closed automatically any more. It closes when I click on the canvas. It's different for the drawing type combo and the PDF text selection combo, where the popover still closes automatically.

It wasn't intentional, but it's not a bad thing, IMO: it was a bit annoying that the popover went away when changing layer visibility. You may want to change the visibility of several layers.
I could make it so the popover closes when you change the active layer, but not when you change visibility. Would that be ok?

- [ ] Fix that

> * One negative thing (which in this case is already on master, but it is related to the popovers):
>   When I have say the eraser selected and change the line style for the pen (without switching to the pen), I can only do it once. The second time I try to do it, the click is simply ignored.

I don't get the same behaviour (probably DE related). In any case, I'm not sure what the best behaviour would be. I see two viable options:

1. The GtkMenuButton (the little arrow) is disabled if the (say) pen tool is not selected. 
2. The popover appears and selecting something in it switches the tool to Pen + this line style

(and of course similar behaviour for the eraser popover) Which would you prefer? Would you want another behaviour?

- [ ] Fix that

> I have also found a regression in the "Migrate Font sizes" plugin, but it's related to the callback not knowing about the local variable "factor" and not to this PR. There is also an issue with the "Middle Mouse" combo in the preferences and the "stroke style" which can only be applied to the Right Mouse button, but not for stylus or default tool (any more?!). That's also in master.

Those two are probably unrelated (the first one due to the new plugin dialogs, the second due to the SettingDialog refactor, I assume). They still need fixing though.

> However the global menu cannot be used on startup. Nothing happens when clicking on the individual menus. I first have to change focus to a different window (say terminal) and then change back focus to Xournal++ again. I get some warnings, but they do not appear any more after deleting/renaming the settings.xml file.

I'm confused here: are the warnings related to the menu bar or are those two independent issues?
Also, are the warnings new (compared to before #4372 was merged) or are they older?

> * The output device for audio recordings is saved wrongly (it takes the next entry). The input device is saved correctly.

Probably also related to the SettingDialog refactor. We should create tickets for those.

- [ ] Fix 
    > * When pressing a toolbar button like "Setsquare" (or "Vertical space") the focus remains on the toolbar widget. So I cannot use the arrow keys (etc) to move the setsquare around, without first clicking onto the canvas.

- [ ] There is also the leftover todo (see [this comment](https://github.com/xournalpp/xournalpp/pull/5154#discussion_r1345835849)) I have to fix: reinstate the "temporarily disable actions" when a blocking job is running

rolandlo:
On Windows I get a crash when changing the toolbar say from "Portrait" to "Xournal++" or reversed (it seems to depend on the toolbar)

```term
(com.github.xournalpp.xournalpp:19916): GLib-GObject-CRITICAL **: 19:51:05.475: ../glib-2.78.0/gobject/gsignal.c:2777: instance '000001b265941230' has no handler with id '8340'

(com.github.xournalpp.xournalpp:19916): GLib-GObject-CRITICAL **: 19:51:05.475: ../glib-2.78.0/gobject/gsignal.c:2777: instance '000001b26593ed60' has no handler with id '8365'
Segmentation fault
````

**Comment by bhennion** [comment]
- [ ] Fix
    > On Windows I get a crash when changing the toolbar say from "Portrait" to "Xournal++" or reversed (it seems to depend on the toolbar)
    
    We also get a bunch of warning on linux

**Comment by rolandlo** [comment]
Testing now 211526116. Something seems seriously wrong with eraser modes and pen styles. For example I have the eraser activated in standard mode and it draws white arrows. It depends on the sequence in which I activated the different tools and modes/line styles though.

**Comment by bhennion** [comment]
- [ ] Fix
    > Testing now [2115261](https://github.com/xournalpp/xournalpp/commit/2115261169535ca1df9b84f4b9b3961421e2cd5c). Something seems seriously wrong with eraser modes and pen styles. For example I have the eraser activated in standard mode and it draws white arrows. It depends on the sequence in which I activated the different tools and modes/line styles though.

^^ Fun...

I fixed the warnings when changing toolbars. Could you try and see if you still get the SegFault?

**Comment by rolandlo** [comment]
> I fixed the warnings when changing toolbars. Could you try and see if you still get the SegFault?

Yes, that works without segfault and warnings now. Great!

**Comment by rolandlo** [comment]
Customizing the toolbar (on Windows) segfaults though when creating a copy.

**Comment by rolandlo** [comment]
And also when trying to customize the toolbar without copying the toolbar it segfaults.

**Comment by bhennion** [comment]
- [ ] Fix
    > And also when trying to customize the toolbar without copying the toolbar it segfaults.

   it looks related to #5199. I can't reproduce on Linux though. Could you give me a backtrace?

**Comment by rolandlo** [comment]
>       it looks related to [You cannot customize any toolbar, because the program closes. #5199](https://github.com/xournalpp/xournalpp/issues/5199). I can't reproduce on Linux though. Could you give me a backtrace?

It's #5199, I agree. On Windows the backtrace may be of limited use. I run xournalpp.exe via gdb and got:

````
[New Thread 8856.0x40d8]
[New Thread 8856.0x1220]
warning: onecore\base\appmodel\runtime\winrt\lib\windows.applicationmodel.package.cpp(1008)\Windows.ApplicationModel.dll!0
0007FF8A3E6F63F: (caller: 00007FF8A3E5C5BC) ReturnHr(1) tid(1b0c) 80070002 Das System kann die angegebene Datei nicht find
en.
warning: onecore\base\appmodel\runtime\winrt\lib\windows.applicationmodel.package.cpp(512)\Windows.ApplicationModel.dll!00
007FF8A3E6F4B1: (caller: 00007FF8656C2CE2) ReturnHr(2) tid(1b0c) 80070002 Das System kann die angegebene Datei nicht finde
n.
[New Thread 8856.0xbb4]
[New Thread 8856.0x1370]
[Thread 8856.0xbb4 exited with code 0]
[Thread 8856.0x1370 exited with code 0]
[New Thread 8856.0x1b30]
[New Thread 8856.0x980]
[Thread 8856.0x1b30 exited with code 0]
[Thread 8856.0x980 exited with code 0]
[New Thread 8856.0x480c]
[Thread 8856.0x3540 exited with code 0]
[Thread 8856.0x668 exited with code 0]
[New Thread 8856.0x421c]
[New Thread 8856.0x4f58]
[Thread 8856.0x421c exited with code 0]
[Thread 8856.0x4f58 exited with code 0]
[New Thread 8856.0xd28]
[New Thread 8856.0x3aec]
[Thread 8856.0xd28 exited with code 0]
[Thread 8856.0x3aec exited with code 0]
[New Thread 8856.0x20f4]
[New Thread 8856.0x1680]
[Thread 8856.0x20f4 exited with code 0]
[Thread 8856.0x1680 exited with code 0]
[New Thread 8856.0x34dc]
[New Thread 8856.0x2f2c]
[Thread 8856.0x34dc exited with code 0]
[Thread 8856.0x2f2c exited with code 0]
[New Thread 8856.0x36e4]
[New Thread 8856.0x3dd8]
[Thread 8856.0x36e4 exited with code 0]
[Thread 8856.0x3dd8 exited with code 0]
[Thread 8856.0x40d8 exited with code 0]

Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff842ea023a in ?? () from C:\msys64\mingw64\bin\libgtk-3-0.dll
(gdb) bt
#0  0x00007ff842ea023a in ?? () from C:\msys64\mingw64\bin\libgtk-3-0.dll
#1  0x00007ff876623112 in ?? () from C:\msys64\mingw64\bin\libgdk-3-0.dll
#2  0x00007ff875b5eb1c in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#3  0x00007ff875b61943 in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#4  0x00007ff875b6206d in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#5  0x00007ff8656c8cb3 in ?? () from C:\msys64\mingw64\bin\libgio-2.0-0.dll
#6  0x00007ff770767465 in XournalMain::run(int, char**) ()
#7  0x00007ff770761499 in main ()
````

The warnings are from the startup. I can also see this after the crash:
grafik

**Comment by rolandlo** [comment]
I noticed that customizing the toolbar works fine if I start with a toolbar that is not predefined. It just segfaults with the predefined toolbars, which may be related to the fact that `customizeToolbars()` calls itself recursively.

**Comment by rolandlo** [comment]
> > * One thing I noticed, which may be intended or not: When clicking on a popover menu item (say "Create new layer" in the layer popover or "dashed" on the pen popover) the popover is not closed automatically any more. It closes when I click on the canvas. It's different for the drawing type combo and the PDF text selection combo, where the popover still closes automatically.
> 
> It wasn't intentional, but it's not a bad thing, IMO: it was a bit annoying that the popover went away when changing layer visibility. You may want to change the visibility of several layers. I could make it so the popover closes when you change the active layer, but not when you change visibility. Would that be ok?

I think that would be OK. When I change the linestyle in the pen combo I would also like the popover to close automatically though.
> 
>     * [ ]  Fix the above
> 
> 
> > * One negative thing (which in this case is already on master, but it is related to the popovers):
> >   When I have say the eraser selected and change the line style for the pen (without switching to the pen), I can only do it once. The second time I try to do it, the click is simply ignored.
> 
> I don't get the same behaviour (probably DE related). In any case, I'm not sure what the best behaviour would be. I see two viable options:
> 
>     1. The GtkMenuButton (the little arrow) is disabled if the (say) pen tool is not selected.
> 
>     2. The popover appears and selecting something in it switches the tool to Pen + this line style
> 
> 
> (and of course similar behaviour for the eraser popover) Which would you prefer? Would you want another behaviour?

I would prefer the second behaviour. When I change the line style or eraser mode it usually means I want to use the tool next. However on Windows currently, when the eraser is active, I can open the pen popover but I can't select anything in it (even not the first time).
> 
>     * [ ]  Fix the above
> 
> 
> > I have also found a regression in the "Migrate Font sizes" plugin, but it's related to the callback not knowing about the local variable "factor" and not to this PR. There is also an issue with the "Middle Mouse" combo in the preferences and the "stroke style" which can only be applied to the Right Mouse button, but not for stylus or default tool (any more?!). That's also in master.
> 
> Those two are probably unrelated (the first one due to the new plugin dialogs, the second due to the SettingDialog refactor, I assume). They still need fixing though.
> 
> > However the global menu cannot be used on startup. Nothing happens when clicking on the individual menus. I first have to change focus to a different window (say terminal) and then change back focus to Xournal++ again. I get some warnings, but they do not appear any more after deleting/renaming the settings.xml file.
> 
> I'm confused here: are the warnings related to the menu bar or are those two independent issues? Also, are the warnings new (compared to before #4372 was merged) or are they older?
> 
I can't reproduce the warnings any more. They don't seem to be related to the menu bar. Let's ignore them for now.

> > * The output device for audio recordings is saved wrongly (it takes the next entry). The input device is saved correctly.
> 
> Probably also related to the SettingDialog refactor. We should create tickets for those.
>

**Comment by rolandlo** [comment]
> However the global menu cannot be used on startup. Nothing happens when clicking on the individual menus. I first have to change focus to a different window (say terminal) and then change back focus to Xournal++ again.

This issue does not appear in the .app bundle, as I have come to realize. It's just an issue when running the installation from the $HOME/gtk folder via `jhbuild run xournalpp`. See https://stackoverflow.com/questions/67622779/cross-platform-gtk3-app-menu-bar-unresponsive-at-first where the same issue is described. Since this does not affect users, we do not need to find a fix for that.

**Comment by bhennion** [comment]
> On Windows the backtrace may be of limited use. I run xournalpp.exe via gdb and got:

Maybe using @Febbe's #4867 could help get a workable backtrace. I have no clue as to where to look.

**Comment by rolandlo** [comment]
> > On Windows the backtrace may be of limited use. I run xournalpp.exe via gdb and got:
> 
> Maybe using @Febbe's #4867 could help get a workable backtrace. I have no clue as to where to look.

After merging #4867 I get
```term
(com.github.xournalpp.xournalpp:12488): xopp-WARNING **: 11:59:39.252: [Crash Handler] Crashed with signal 11
0> xournalpp!Z20setEmergencyDocumentP8Document+0x321
1> xournalpp!gnu_exception_handler+0x162
2> ntdll!_C_specific_handler+0x96
3> ntdll!_chkstk+0x11F
4> ntdll!RtlRaiseException+0x484
5> ntdll!KiUserExceptionDispatcher+0x2E
6> libgtk_3_0!gtk_widget_path_has_parent+0x144A
7> libgdk_3_0!gdk__private__+0xE2
8> libglib_2_0_0!g_clear_list+0x11AC
9> libglib_2_0_0!g_get_monotonic_time+0xA73
10> libglib_2_0_0!g_main_context_iteration+0x2D
11> libgio_2_0_0!g_application_run+0x183
12> xournalpp!ZN11XournalMain3runEiPPc+0x7DA
13> xournalpp!main+0x49
14> xournalpp!WinMainCRTStartup+0xFFFFFFFFFFFFFF1
15> xournalpp!WinMainCRTStartup+0x16
16> KERNEL32!BaseThreadInitThunk+0x14
17> ntdll!RtlUserThreadStart+0x21

(com.github.xournalpp.xournalpp:12488): xopp-WARNING **: 11:59:39.622: [Crash Handler] Wrote crash log to: C:\Users\Roland\AppData\Local\Microsoft\Windows\INetCache\xournalpp\errorlogs\errorlog.20231008-115939.log

(com.github.xournalpp.xournalpp:12488): xopp-WARNING **: 11:59:39.622: Trying to emergency save the current open document?

(com.github.xournalpp.xournalpp:12488): xopp-WARNING **: 11:59:39.622: Successfully saved document to "C:\Users\Roland\AppData\Local\xournalpp\emergencysave.xopp"
````

**Comment by bhennion** [comment]
Thanks a lot for taking the time! The output is sadly not very helpful.
Question: does it happen when you select customize in the menu and/or when you open the Manage Toolbar dialog and click on customize?

**Comment by rolandlo** [comment]
> Thanks a lot for taking the time! The output is sadly not very helpful. Question: does it happen when you select customize in the menu and/or when you open the Manage Toolbar dialog and click on customize?

It happens when you customize in the menu and then either answer yes or no to the question whether you want to create a copy of the (predefined) toolbar. When you customize a not predefined toolbar, no segfault happens. I don't see where in the Manage Toolbar dialog you can click on customize, by the way. 

https://github.com/xournalpp/xournalpp/blob/dc297ae60520ee0b06ded8ab59305ab842cfd128/src/core/control/Control.cpp#L563-L582

**Comment by bhennion** [comment]
I rebased, cleaned up the commit history. Every thing is fixed except

1. > Testing now [2115261](https://github.com/xournalpp/xournalpp/commit/2115261169535ca1df9b84f4b9b3961421e2cd5c). Something seems seriously wrong with eraser modes and pen styles. For example I have the eraser activated in standard mode and it draws white arrows. It depends on the sequence in which I activated the different tools and modes/line styles though.

    I could not reproduce. It may have been unwillingly fixed. I don't know

2. and #5199. I don't know how to investigate without a windows setup to reproduce

**Comment by bhennion** [comment]
> 2\. and [SegFault when trying to customize the toolbar (Windows only) #5199](https://github.com/xournalpp/xournalpp/issues/5199). I don't know how to investigate without a windows setup to reproduce

If I am correct and this is in fact a Dialog issue (and not a GMenu/Toolbar issue), it should actually already be fixed (both on master and on this PR - which I rebased since). @rolandlo would you mind checking when you have time?

**Comment by Febbe** [comment]
> > 2\. and [SegFault when trying to customize the toolbar (Windows only) #5199](https://github.com/xournalpp/xournalpp/issues/5199). I don't know how to investigate without a windows setup to reproduce
> 
> If I am correct and this is in fact a Dialog issue (and not a GMenu/Toolbar issue), it should actually already be fixed (both on master and on this PR - which I rebased since). @rolandlo would you mind checking when you have time?

Couldn't even reproduce the original bug you fixed with that commit(compiled with MSVC STL and both clang-cl.exe and msvc's cl.exe) but I can try again.

**Comment by rolandlo** [comment]
> > 2. and [SegFault when trying to customize the toolbar (Windows only) #5199](https://github.com/xournalpp/xournalpp/issues/5199). I don't know how to investigate without a windows setup to reproduce
> 
> If I am correct and this is in fact a Dialog issue (and not a GMenu/Toolbar issue), it should actually already be fixed (both on master and on this PR - which I rebased since). @rolandlo would you mind checking when you have time?

You're right. It's fixed now after the rebase.

**Comment by rolandlo** [comment]
> I rebased, cleaned up the commit history. Every thing is fixed except
> 
>     1. > Testing now [2115261](https://github.com/xournalpp/xournalpp/commit/2115261169535ca1df9b84f4b9b3961421e2cd5c). Something seems seriously wrong with eraser modes and pen styles. For example I have the eraser activated in standard mode and it draws white arrows. It depends on the sequence in which I activated the different tools and modes/line styles though.
>        
>        
>        I could not reproduce. It may have been unwillingly fixed. I don't know
> 

After fixing one line in my `settings .xml` ("tools" section):
```term
    <data name="eraser">
      <attribute name="drawingType" type="string" value="arrow"/>
```
to
```term
    <data name="eraser">
      <attribute name="drawingType" type="string" value="default"/>
```

I can't reproduce any of this issue any longer. So it seems to be fixed. Great!

However I have found a different issue: When I select an existing stroke and try to change the line style, I can't do it using the pen toolbar combo, since the selection is lost once I press on the arrow next to the pen icon. I can still change the line style via the menu though.

src/core/gui/toolbarMenubar/ToolButton.cpp
```diff
@@ -52,9 +50,25 @@ auto ToolButton::createItem(bool horizontal) -> GtkToolItem* {
     GtkToolItem* it = gtk_tool_item_new();
     if (popover) {
         GtkMenuButton* menubutton = GTK_MENU_BUTTON(gtk_menu_button_new());
+        gtk_widget_set_can_focus(GTK_WIDGET(menubutton), false);  // todo(gtk4) not necessary anymore
         gtk_menu_button_set_popover(menubutton, popover.get());
         gtk_menu_button_set_direction(menubutton, GTK_ARROW_DOWN);
 
+        if (target && toggle) {
+            // Pressing the menubutton also selects the tool
```
rolandlo, line 58: Looks like this has a serious downside with this: When the selection tool is active and you press the menubutton for the pen to change the line style, the selection is lost and you can't change the line style of the selection.

bhennion: Yes. I've been looking for a good solution, but it's not clear yet.

bhennion: The toolbar can have individual buttons to select the linestyle
Screenshot_20231013_181541

What behaviour do we want of them?

1. Always sensitive. Change the linestyle if the current tool is PEN, change the selection's line style if something is selected, change the current tool to PEN otherwise.
2. Change the linestyle if the current tool is PEN, change the selection's line style if something is selected, and disabled otherwise.

As far as I can tell, there are 3 places where the line style can be selected: in the menu (behaviour 1. is better, I think), in the dropdown popover of the toolbar PEN button (behaviour 1. is better, I think) and those individual line style selectors (1. or 2., I don't know)

rolandlo: With "change the current tool to PEN otherwise" you probably mean "change the current tool to PEN and set the line style according to the button/menu".

In general I think having sensitive widgets that can do something useful is better than having insensitive widgets (that are still visible). So I would go with option 1 for all 3 places.

bhennion: I'm a bit scared we'll get a bunch of "Setting dashed line on highlighter does not work". I'd say that if we use 1, then the icon should make it clear this will activate the pen. It'd be like a "tool button with specific options" that has been asked somewhere in an issue.

rolandlo: Ok, fair point, the action should be intuitively understandable, which may not be the case for the individual line style selectors changing to the pen tool. To create "tool buttons with specific options" icons (like in https://github.com/xournalpp/xournalpp/issues/2431#issuecomment-820060571) looks out of scope for this PR though. So then I change my opinion and recommend to disable the individual line style selectors if the current tool is not the PEN and nothing is selected, at least if it doesn't make things too complicated.

bhennion: TBH it's simpler to create a bunch of icons with a pen on top of the linestyle and use those for the standalone linestyle buttons. It's cleaner than the solution I have in mind for implementing 2.

bhennion: Like that: 
Screenshot_20231013_205153

rolandlo: Ok, looks good.

bhennion: And the Lucide ones
Screenshot_20231013_210241

bhennion:
Done. Added the icons and made so that clicking on them, or in the menu, or in the pen tool  popover does as in 1.

rolandlo:
I'm confused now. It seems like highlighter strokes can have a (non-plan) line style, too. They shouldn't, I think. Right??!
Already in version 1.2.0 you can make highlighter strokes having a line style by putting them into a selection containing a pen stroke and then changing the line style.

I have also discovered an issue which is not present on master: When I change the toolbar, the popover for the pen does not appear anymore when clicking on the menu button.

bhennion:
> I have also discovered an issue which is not present on master: When I change the toolbar, the popover for the pen does not appear anymore when clicking on the menu button.

This should be fixed now.

> I'm confused now. It seems like highlighter strokes can have a (non-plan) line style, too. They shouldn't, I think. Right??! Already in version 1.2.0 you can make highlighter strokes having a line style by putting them into a selection containing a pen stroke and then changing the line style.

One of those "not a feature, but a bug" I think.

rolandlo (approved):
I gave it another look. The issue with toolbar switching is fixed indeed. Please go ahead.

Febbe:
Yes, go along and ship it

PR #5164, opened 2023-09-20
Fix some memory leaks

bhennion:
Fixes some memory leaks.
Some of the fixes are taken from #4613 (and flagged as coauthored with @jhilmer).

I also tried enabling `fsanitize=leak,address,undefined` to the tests' build env to implicitely add the test "no leaks in whatever the tests call". It detected some dataleaks in SaveHandler but I have not found the source of the leaks (yet).
I'll investigate and make a second PR once I found the issue.

### DIFF
--- a/src/core/control/xml/XmlPointNode.cpp
+++ b/src/core/control/xml/XmlPointNode.cpp
@@ -9,7 +9,7 @@
 
 XmlPointNode::XmlPointNode(const char* tag): XmlAudioNode(tag) {}
 
-void XmlPointNode::addPoint(Point point) { points.emplace_back(std::move(point)); }
+void XmlPointNode::setPoints(std::vector<Point> pts) { this->points = std::move(pts); }
 
 void XmlPointNode::writeOut(OutputStream* out) {
     /** Write stroke and its attributes */
--- a/src/core/control/xml/XmlPointNode.h
+++ b/src/core/control/xml/XmlPointNode.h
@@ -24,7 +24,7 @@ public:
     XmlPointNode(const char* tag);
 
 public:
-    void addPoint(Point point);
+    void setPoints(std::vector<Point> points);
     void writeOut(OutputStream* out) override;
 
 private:
--- a/src/core/control/xojfile/LoadHandler.cpp
+++ b/src/core/control/xojfile/LoadHandler.cpp
@@ -29,6 +29,7 @@
 #include "util/LoopUtil.h"
 #include "util/PlaceholderString.h"  // for PlaceholderString
 #include "util/i18n.h"               // for _F, FC, FS, _
+#include "util/raii/GObjectSPtr.h"
 
 #include "LoadHandlerHelper.h"  // for getAttrib, getAttribDo...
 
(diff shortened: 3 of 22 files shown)


bhennion:
I fixed all the memleaks I could find by running the test units through Valgrind, with the exception of some memleaks in Pango (that I reported [upstream](https://gitlab.gnome.org/GNOME/pango/-/issues/765)).

src/core/control/xml/XmlPointNode.cpp
```diff
@@ -9,7 +9,7 @@
 
 XmlPointNode::XmlPointNode(const char* tag): XmlAudioNode(tag) {}
 
-void XmlPointNode::addPoint(Point point) { points.emplace_back(std::move(point)); }
+void XmlPointNode::setPoints(const std::vector<Point>& pts) { this->points = pts; }
```
Febbe, line 12:
```suggestion
void XmlPointNode::setPoints(std::vector<Point> pts) { this->points = std::move(pts); }
```

src/core/control/xml/XmlPointNode.h
```diff
@@ -24,7 +24,7 @@ class XmlPointNode: public XmlAudioNode {
     XmlPointNode(const char* tag);
 
 public:
-    void addPoint(Point point);
+    void setPoints(const std::vector<Point>& points);
```
Febbe, line 27:
```suggestion
    void setPoints(std::vector<Point> pts);
``

File: src/core/control/xojfile/LoadHandler.cpp [hunk 3]
```diff
@@ -407,27 +408,35 @@ void LoadHandler::parseBgPixmap() {
...
-        GInputStream* inputStream = g_memory_input_stream_new_from_bytes(attachment);
+        /**
+         * To avoid an unecessary copy, the GBytes is initialized with noop free function. The data is still managed by
+         * the std::optional<std::string> instance.
+         * The GBytes instance forbids any modification of the data until it has been released, so do not remove the
+         * const qualifier on readResult or imgData
+         */
+        GBytes* data = g_bytes_new_with_free_func(
+                imgData.data(), imgData.size(), +[](gpointer) {}, nullptr);
+        xoj::util::GObjectSPtr<GInputStream> inputStream(g_memory_input_stream_new_from_bytes(data), xoj::util::adopt);
```
Febbe, line 425: I think `g_memory_input_stream_new_from_data ` is the better choice here.

test/unit_tests/util/ObjectIOStreamTest.cpp
```diff
@@ -14,78 +14,69 @@
...
-}
-
 template <typename T>
 std::string serializeDataVector(const std::vector<T>& data) {
     ObjectOutputStream outStream(new BinObjectEncoding);
     outStream.writeData(data);
     auto outStr = outStream.getStr();
-    return {outStr->str, outStr->len};
+    auto resStr = std::string{outStr->str, outStr->len};
+    g_string_free(outStr, true);
```
Febbe, line 23: Nit: we could also wrap the g_string into a RAII guard.

Febbe (changes requested):

test/unit_tests/util/ObjectIOStreamTest.cpp
bhennion: I wondered as well, but this is only in the test unit, and the real refactor would be to get rid of GString's in the ObjectOutputStream. Too big for the release branch though.

src/core/control/xojfile/LoadHandler.cpp
bhennion: Indeed it is!

Febbe (approved):
LGTM, go along and merge!

bhennion:
Just rebased to cleanup the commit tree. Merging once the CI passes

Issue #5087, opened 2023-08-24
Assert failure: negative pressure values


chandan-harry:
![Screenshot from 2023-08-24 20-54-46](https://github.com/xournalpp/xournalpp/assets/143109218/4a9684b1-985c-4f4d-be29-94563bf9078e)

rolandlo:
Please open the logfile and paste the content here.

eduardo23maf:
Date: Sun Aug 27 18:46:01 2023
Error: signal 6
Xournal++ version 1.2.1
Git commit b48932a2 from xournalpp/HEAD
Gtk version 3.24.38

[bt]: (0) /snap/xournalpp/86/usr/bin/xournalpp(+0x3e3555) [0x55c7b8f9e555]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x42520) [0x7f967eeed520]
[bt]: (2) /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x12c) [0x7f967ef41a7c]
[bt]: (3) /lib/x86_64-linux-gnu/libc.so.6(raise+0x16) [0x7f967eeed476]
[bt]: (4) /lib/x86_64-linux-gnu/libc.so.6(abort+0xd3) [0x7f967eed37f3]
[bt]: (5) /lib/x86_64-linux-gnu/libc.so.6(+0x2871b) [0x7f967eed371b]
[bt]: (6) /lib/x86_64-linux-gnu/libc.so.6(+0x39e96) [0x7f967eee4e96]
[bt]: (7) /snap/xournalpp/86/usr/bin/xournalpp(+0x6042d3) [0x55c7b91bf2d3]
[bt]: (8) /snap/xournalpp/86/usr/bin/xournalpp(_ZN3xoj4view16StrokeViewHelper16drawWithPressureEP6_cairoRKSt6vectorI5PointSaIS5_EERK9LineStyled+0x206) [0x55c7b91bf557]
[bt]: (9) /snap/xournalpp/86/usr/bin/xournalpp(_ZNK3xoj4view14StrokeToolView4drawEP6_cairo+0x276) [0x55c7b923f948]
[bt]: (10) /snap/xournalpp/86/usr/bin/xournalpp(_ZN11XojPageView9paintPageEP6_cairoP20_cairo_rectangle_int+0x1f5) [0x55c7b913c791]
[...]

PR #5170, opened 2023-09-22
Check device hash for input sequences

### LINKED ISSUE #5116: Crash when drawing pen and mouse triggered at the same time.
### Operating System

Linux

### (Linux only) Distribution

Arch

GNOME

X.Org

### Installation Method

Aur

### Xournal++ Version

xournalpp 1.2.1

### libgtk Version

xournalpp 1.2.1

### Bug Description

If I press down with a drawing tablet pen at the same time as pressing down with a mouse pen, it causes the software to crash.

### Expected Behaviour

To ignore the input of the mouse and not crash.

### Steps to Reproduce

1. Get your tablet pen and mouse
2. Click on the mouse at the same time as pressing down with the drawing pen

### Additional Context

DEBUG   : GLib-GIO :: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)
DEBUG   : GLib-GIO :: _g_io_module_get_default: Found default implementation gvfs (GDaemonVfs) for â€˜gio-vfsâ€™
DEBUG   : GLib :: unsetenv() is not thread-safe and should not be used after threads are created
MESSAGE : xopp :: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale
INFO    : xopp :: Loading plugins from: /usr/share/xournalpp/plugins
INFO    : xopp :: Loading plugins from: /home/samio/.config/xournalpp/plugins
MESSAGE : xopp :: Plugin "MigrateFontSizes" UI initialized
DEBUG   : GdkPixbuf :: gdk_pixbuf_from_pixdata() called on:
DEBUG   : GdkPixbuf :: 	Encoding raw
DEBUG   : GdkPixbuf :: 	Dimensions: 20 x 25
DEBUG   : GdkPixbuf :: 	Rowstride: 80, Length: 2024
DEBUG   : GdkPixbuf :: 	Copy pixels == false
DEBUG   : GdkPixbuf :: gdk_pixbuf_from_pixdata() called on:
DEBUG   : GdkPixbuf :: 	Encoding raw
DEBUG   : GdkPixbuf :: 	Dimensions: 20 x 25
DEBUG   : GdkPixbuf :: 	Rowstride: 80, Length: 2024
DEBUG   : GdkPixbuf :: 	Copy pixels == false
WARNING : xopp :: Found a stroke with some non-positive pressure values. Removing the affected points.

bhennion:
This PR adds a device hash and rejects input events that have not been issued by the pointing device that started a sequence.
Fixes #5116, possibly #5087 if this was the only source of negative pressure values.

This deserves some testing on various devices.

### DIFF
--- a/src/core/gui/inputdevices/DeviceId.h
+++ b/src/core/gui/inputdevices/DeviceId.h
@@ -1,0 +1,22 @@
+/*
+ * Xournal++
+ *
+ * Device identifier
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+struct DeviceId {
+    DeviceId() = default;
+    explicit DeviceId(const void* id): id(id) {}
+    void clear() { id = nullptr; }
+    operator bool() const { return id != nullptr; }
+
+private:
+    const void* id = nullptr;
+};
--- a/src/core/gui/inputdevices/DeviceId.h
+++ b/src/core/gui/inputdevices/DeviceId.h
@@ -12,10 +12,12 @@
 #pragma once
 
 struct DeviceId {
-    DeviceId() = default;
-    explicit DeviceId(const void* id): id(id) {}
-    void clear() { id = nullptr; }
-    operator bool() const { return id != nullptr; }
+    constexpr DeviceId() = default;
+    explicit constexpr DeviceId(const void* id): id(id) {}
+    void reset(const void* id = nullptr) { this->id = id; }
+    explicit constexpr operator bool() const { return id != nullptr; }
+    constexpr bool operator==(const DeviceId& o) const { return id == o.id; }
+    constexpr bool operator!=(const DeviceId& o) const { return !(*this == o); }
 
 private:
     const void* id = nullptr;
--- a/src/core/gui/inputdevices/DeviceId.h
+++ b/src/core/gui/inputdevices/DeviceId.h
@@ -14,7 +14,7 @@
 struct DeviceId {
     constexpr DeviceId() = default;
     explicit constexpr DeviceId(const void* id): id(id) {}
-    void reset(const void* id = nullptr) { this->id = id; }
+    constexpr void reset(const void* id = nullptr) { this->id = id; }
     explicit constexpr operator bool() const { return id != nullptr; }
     constexpr bool operator==(const DeviceId& o) const { return id == o.id; }
     constexpr bool operator!=(const DeviceId& o) const { return !(*this == o); }
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -209,11 +209,11 @@ void XojPageView::endSpline() {
 }
 
 auto XojPageView::onButtonPressEvent(const PositionInputData& pos) -> bool {
-    if (currentSequenceDeviceHash) {
+    if (currentSequenceDeviceId) {
         // An input sequence is already under way from another device
         return false;
     }
-    currentSequenceDeviceHash = pos.deviceHash;
+    currentSequenceDeviceId = pos.deviceId;
 
     Control* control = xournal->getControl();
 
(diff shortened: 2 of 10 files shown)


Febbe:
Pretty weird, I checked the device name and it's class on my laptop and they were the same for both events with and without pressure

bhennion:
So you still get a negative pressure assertion failure with this patch?

Febbe:
Not yet, but at the time I tried to debug it, all events passed to the PenInputhandler were from the same device.

bhennion:
There may be another source for #5087 then.
Another test may be: are the events processed by the same instance of PenInputHandler? (there are several of them)

Febbe:
Another thing is, that generating the hash for each event is most likely slower than just comparing the char pointer and the event type.
Probably, we can just store the pointer / increase ref count of the name without a copy (shouldn't be freed, since the device will exist, until it's plugged out).

bhennion:
Or, assuming that Gdk only allocates one GdkDevice per physical device, test the address. I can't find any documentation on that though

src/core/gui/inputdevices/DeviceId.h
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+struct DeviceId {
+    DeviceId() = default;
+    explicit DeviceId(const void* id): id(id) {}
+    void clear() { id = nullptr; }
```
Febbe, line 17: Dunno, if we might need this anytime:
```suggestion
    void reset(void const* id = nullptr) { this->id = id; }
```

src/core/gui/inputdevices/DeviceId.h
```diff
@@ -0,0 +1,22 @@
...
+    operator bool() const { return id != nullptr; }
```
Febbe, line 18: Conversion operators, should be marked "explicit", especially boolean conversions. 
Also, the equality and inequality operator is missing; the code only compiles, because `lhs != rhs` is implicitly converted to: `(lhs.id != nullptr) != (rhs.id != nullptr)` which will most likely be `false != false`

```suggestion
    constexpr explicit operator bool() const { return id != nullptr; }
    constexpr auto operator==(DeviceId rhs) -> bool { return id == rhs.id; }
    constexpr auto operator!=(DeviceId rhs) -> bool { return id == rhs.id; }
```

Febbe (changes requested):
Looks good, but it must be tested, if the device pointers are unique and persistent.
Also the implicit cast bug must be fixed.

Febbe:
> Looks good, but it must be tested, if the device pointers are unique and persistent.

I found a lot of descriptions which always state that:
A GdkDevice with pointer capabilities. This object is owned by GTK and must not be freed.

bhennion:
Thanks for the catch. Of course the version I had pushed did not work...
This new version seems to work on my devices.

src/core/gui/inputdevices/DeviceId.h
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @license GNU GPLv2 or later
+ */
+
+#pragma once
+
+struct DeviceId {
+    constexpr DeviceId() = default;
+    explicit constexpr DeviceId(const void* id): id(id) {}
+    void reset(const void* id = nullptr) { this->id = id; }
```
Febbe, line 17: Nit: can also be constexpr

Febbe (approved):
LGTM, just rebase and merge.

src/core/gui/inputdevices/DeviceId.h
bhennion:
```suggestion
    constexpr void reset(const void* id = nullptr) { this->id = id; }
```

MikuChan03:
This breaks my workflow at least. I have a thinkpad that has both a touchpad and one of those red trackpoints. Like this:

sef

The thing is, the buttons above and below the touchpad belong to the two different devices. The above two buttons are for the trackpoint, the below for the touchpad. I used to be able to press whichever, if I wanted to draw a line in xournal++. Now the above two don't work, for it's a different device. What do?
Many thanks

rolandlo:
I can confirm the issue on my Lenovo Thinkpad.

Issue #799, opened 2019-01-21
Organize notes into notebooks


dexter0195:
It would be great to organize notes into notebooks or folders from inside the app as it is done in OneNote.
It's true that you can navigate the filesystem from a file explorer and then select the file to edit but the feeling is not the same.
Furthermore you could make hyperlinks between notes with that.

amirji:
It would be great if there would be such that.

veecue:
I will be working on such a feature. I was thinking about another Document type called a xournalpp workspace that organizes .xopp files on two directory levels.
Those would be displayed nicely in the sidebar similar to OneNote

Erater:
Are there any Updates on this feature?

dibbel26:
I would really enjoy such a feature, as you said, it might be possible to stick to the current file format but with some sort of "One Note like" notebook organisation in which case the existing folder structure could be interpreted as such notebooks.

Fatih20:
> I would really enjoy such a feature, as you said, it might be possible to stick to the current file format but with some sort of "One Note like" notebook organisation in which case the existing folder structure could be interpreted as such notebooks.

Yes. User could just setup a special folder that Xournal then can use to make folders within it to organize the notebooks.

Fatih20:
@veecue Any update?

veecue:
Nothing so far. Just haven't had the time so far. Feel free to implement it ;). The easiest possibility would be to allow opening folders and just have a sidebar alternative that displays the opened folder in a tree view.

rolandlo:
@veecue Your ideas sound good to me. I think this would really make a big impact. Really hope you will find time to code it up.

rolandlo:
Having discovered [Joplin](https://joplinapp.org/) for managing notebooks and its ability to attach all kind of files (including xopp-files), I am not so sure anymore whether this is really something we should implement in Xournal++. Checkout [this blog post](https://linux-in-der-schule.de/2020/11/) and in particular the video near the word Xournal++. It looks great to me. Even syncing up to Nextcloud (and alike) is supported. I will organize my stuff in that way in future, I think. That's kind of the missing piece I needed to organize my workflow.

mizhozan:
[...]

PR #5189, opened 2023-10-01
Workspace sidebar

MatteoGalletta:
Added Workspace sidebar (by default it is hidden, open it from "View > Show Workspaces").

Tries to close issue #799 inspiring from Notepad++ workspace behaviour.

Features:
	-Add folders to the workspace from the File menu "Add folder to Workspace" or using the button on the sidebar
	-Remove all folders from the workspace from the File Menu "Clear and close workspace" or using the button on the sidebar (the button doesn't close the workspace tho, it just clears it)
	-Show/Hide the workspace from View > Show Workspace
	-Resize the workspace (the width gets stored in the settings)
	-The workspace is just one and it's saved in ~/.config/xournalpp/workspace.txt and is persistent between app executions.
	-Immediatly open the files contained in the folders added to the workspace by double-clicking on them. The files are filtered by extension and the ones visible are the ones that Xournal++ can open (which are PDF, xopp, xoj, xopt, moj).

Known Issues:
	-F8 shortcut doesn't work in most of the situations (e.g. if you start drawing)
	-There's the following warning (I tried to remove the property but then the GtkPaned behaviour changes): "GtkBox does not have a child property called shrink"

Next Steps (I'd like to follow):
	-Add icons (for folders and different extensions files) - currently files and empty folders look the same
	-Improve UI colors (currently the TreeView and the label background are slightly different)
	-Allow to remove a single folder
	-Implement other right click functionalities (such as rename/delete file, copy path, etc)

I also fixed a bug (maybe I should have made a separate pull request but it was just an if statement) in the sidebar, which could be replicated by doing such steps:
	1. Resize the sidebar (and remember the new width).
	2. Hide the sidebar (F12)
	3. Close the app, open the app and close it again.
	4. Open the app and show the sidebar (F12).
	5. You can notice that the sidebar width hasn't been saved as it should. Now it's fixed.

I tested the changes only on (Manjaro) Linux.

I'm new to contributing, I can see that the main.glade file changes list is huge. I tabbed everything to add a wrapper container but looks like git doesn't understand it?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,7 @@ mark_as_advanced(FORCE
         )
 
 # Advanced development config
+set(DEV_WORKSPACE_FILE "workspaces.txt" CACHE STRING "Workspace folders list file name")
 set(DEV_TOOLBAR_CONFIG "toolbar.ini" CACHE STRING "Toolbar config file name")
 set(DEV_SETTINGS_XML_FILE "settings.xml" CACHE STRING "Settings file name")
 set(DEV_PALETTE_FILE "palette.gpl" CACHE STRING "Color Palette")
@@ -253,7 +254,7 @@ endif ()
 
 mark_as_advanced(FORCE
         DEV_TOOLBAR_CONFIG DEV_SETTINGS_XML_FILE DEV_PRINT_CONFIG_FILE DEV_METADATA_FILE
-        DEV_ENABLE_GCOV DEV_CHECK_GTK3_COMPAT
+        DEV_ENABLE_GCOV DEV_CHECK_GTK3_COMPAT DEV_WORKSPACE_FILE
         )
 
 configure_file(
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -34,6 +34,7 @@ Here you can find complete list of Xournal++ CMake flags (sorted by categories).
 | `DEV_METADATA_MAX_ITEMS` *[A]* | 50               | Maximal amount of metadata elements
 | `DEV_PRINT_CONFIG_FILE` *[A]*  | print-config.ini | Print config file name
 | `DEV_SETTINGS_XML_FILE` *[A]*  | settings.xml     | Settings file name
+| `DEV_WORKSPACE_FILE` *[A]*     | workspace.txt    | Workspace folders list file name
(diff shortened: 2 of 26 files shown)


rolandlo:
Thanks for working on this feature. It already looks quite useful even though only basic functionality has been implemented so far. Some quick feedback:
- Currently compilation fails on MS Windows. I think `Workspace::fillTreeFromFolderPath` should accept a `fs::path`, not a `std::string` as `folderPath`. 
```c++
D:/a/1/s/src/core/gui/Workspace.cpp: In member function 'virtual void Workspace::addFolder(std::filesystem::__cxx11::path)':
D:/a/1/s/src/core/gui/Workspace.cpp:110:42: error: cannot convert 'std::filesystem::__cxx11::path' to 'std::string' {aka 'std::__cxx11::basic_string<char>'}
  110 |     fillTreeFromFolderPath(store, &iter, folderPath);
      |                                          ^~~~~~~~~~
      |                                          |
      |                                          std::filesystem::__cxx11::path
D:/a/1/s/src/core/gui/Workspace.cpp:52:94: note:   initializing argument 3 of 'void Workspace::fillTreeFromFolderPath(GtkTreeStore*, GtkTreeIter*, std::string)'
```
- Please apply clang format
- The icons you use (for adding a folder, closing all folders) are not in the Adwaita icon theme. So you should add icons to the `ui` folder.
- [Rnote](https://github.com/flxzt/rnote) also has a workspace browser (an already quite refined one), maybe also inspired from Notepad++. You might want to take some inspiration from there as well.

Febbe (changes requested):

rolandlo:
One question about the design: Instead of creating a separate sidebar for the workspace, wouldn't it make more sense to present it in a way that you can switch between preview and workspace (and something like "library" in future), but not see both at once? Here is a mockup from the [Gtk4/Libadwaita redesign](https://github.com/rolandlo/xournalpp-libadwaita-mockup/blob/main/main/main.blp)
Bildschirmfoto vom 2023-10-28 15-42-16

jonjampen:
This is exactly the feature I was looking for! Any updates on this?

JoshuaMariner:
This is also a feature I've been looking for in Xournal++, hope to see it as a feature implemented!

PR #5200, opened 2023-10-05
Fix double gtk_window_close in XojMsgBox::askQuestionWithMarkup

### LINKED ISSUE #5197: Segmentation fault on startup popup close
### Operating System

Linux

### (Linux only) Distribution

openSUSE Tumbleweed

KDE Plasma

Wayland

### Installation Method

Build from source

### Xournal++ Version

1.2.1+dev (a3686cf6)

### libgtk Version

3.24.38

### Bug Description

When an emergency save or an error log (or both) exist, a popup appears on startup to delete or restore the file, respectively to view/report the error log. Clicking on any of the buttons will execute the specified action but also cause a segmentation fault. 

### Expected Behaviour

Xournal++ should not throw a segmentation fault. 

### Steps to Reproduce

1. Create a file `~/.config/xournalpp/emergencysave.xopp`
2. Open Xournal++
3. Click on any of the buttons

### Additional Context

I have already tried to track down this issue a little: 
This function call:
https://github.com/xournalpp/xournalpp/blob/5fd2dd3703ef7b7cd07e63ef738fc0affcaa287f/src/util/gtk4_helper.cpp#L26
will throw the segmentation fault when called (undirectly) from the `XojMsgBox` destructor
https://github.com/xournalpp/xournalpp/blob/5fd2dd3703ef7b7cd07e63ef738fc0affcaa287f/src/util/include/util/XojMsgBox.h#L26

But I haven't found out what the underlying problem is. When in the debug console, `win` seems to be a valid gtk object. 
Has anyone else encountered this issue?

bhennion:
Fixes #5197
@tmoerschell could you try this PR out and check it fixes the original issue indeed?


tmoerschell:
Yes, this does work! 
Were you not able to reproduce the issue?
Thank you for the quick reaction!

bhennion:
> Yes, this does work!
> Were you not able to reproduce the issue?

No I wasn't :)

Febbe:
Can you also fix bool -> gboolean and the missing GClosure* parameter, since they have a high severity?

Febbe:
Looks good to me, is there a reason, you use reinterpret_cast, when a static cast would also work (`void*` -> `KnownType*`)? Also, we probably should still use the GTK_CASTS, even they are mostly just a reinterpret_cast

bhennion:
Mainly because of 

> Unlike static_cast, but like const_cast, the reinterpret_cast expression does not compile to any CPU instructions (except when converting between integers and pointers, or between pointers on obscure architectures where pointer representation depends on its type). It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type.

But they are probably equivalent in this case, I guess.

Febbe:
> Mainly because of 
> 
> > Unlike static_cast, but like const_cast, the reinterpret_cast expression does not compile to any CPU instructions (except when converting between integers and pointers, or between pointers on obscure architectures where pointer representation depends on its type). It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type.
> 
> But they are probably equivalent in this case, I guess.

Yes, between void* and Type* the static_cast does nearly the same. But it can't cast away const and volatile qualifiers. It also does not raise a warning due to the usage of reinterpret_cast. Last but not least, I think the fact that it does generate any CPU instruction is a trap. Sounds good and efficient, but is nearly every time a bug or UB.

PR #5209, opened 2023-10-07
Make Mac version use Meta instead of Ctrl again

rolandlo:
Fixes a regression from #4372 regarding accelerator modifiers, that are not defined in `main.glade` anymore, but in `mainmenubar.xml`.

### DIFF
--- a/mac-setup/build-app.sh
+++ b/mac-setup/build-app.sh
@@ -78,8 +78,8 @@ export GTKDIR="$1/inst"
 
 "$GTK_MAC_BUNDLER" xournalpp.bundle
 
-echo "Replace GDK_CONTROL_MASK by GDK_META_MASK in main.glade"
-sed -i -e 's/GDK_CONTROL_MASK/GDK_META_MASK/g' ./Xournal++.app/Contents/Resources/ui/main.glade
+echo "Replace Ctrl by Meta in mainmenubar.xml"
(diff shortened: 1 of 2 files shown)


bhennion:
There are also accelerators defined in control/actions/ActionProperties.h. Typically for undo/redo which are added dynamically to the menubar (to be able to change the description).

rolandlo:
There are also `GDK_KEY_Control_L` and `GDK_KEY_Control_R` modifiers hardcoded in `BaseShapeHandler.cpp` and `VerticalToolHandler.cpp`, and `GDK_CONTROL_MASK` in `BaseShapeHandler.cpp`, `TextEditor.cpp`, `TextEditorWidget.h`,  `ZoomControl.cpp`, `XournalView.cpp` and `PositionInputData.cpp` and `texdialog.glade`. Should this all be part of this PR?

Some further modifications may be needed, since there are more differences between MacOS and Linux/Windows regarding keyboard shortcuts. See https://gaphor.org/de/2022/12/10/gtk4-macos-keybindings/ and https://support.apple.com/en-us/HT201236

bhennion:
> Should this all be part of this PR?

Probably not.

Note that the rest of them could be fixed in 1.2.x as well

rolandlo:
Merging after the pipeline succeeds. It works when built locally on MacOS.

PR #5213, opened 2023-10-08
fix Debian build

rolandlo:
`g_warning` is not `constexpr` and the build fails on Debian

See also https://dev.azure.com/xournalpp/xournalpp/_build/results?buildId=28536&view=logs&j=30d9c38c-bfb1-5863-ddb1-d34ae18a0686&t=f29032be-d771-562d-9c95-a9db7f2cfcc4


bhennion:
I'd rather we replaced the g_warning than remove the constexpr

rolandlo:
That's fine for me, too. I have amended the change and will merge once the pipeline succeeds.

PR #5217, opened 2023-10-08
Release 1.2.2

bhennion:
Draft PR. Do not merge. This will be fast forwarded onto the release branch.
@xournalpp/core Please have a look

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+xournalpp (1.2.3~dev-1) UNRELEASED; urgency=medium
+
+  * 
+
+ -- Benjamin Hennion <benjamin.hennion@wanadoo.fr>  Sun, 08 Oct 2023 15:53:47 +0200
+
 xournalpp (1.2.2-1) unstable; urgency=medium
 
   * Fixed various memory leaks and segmentation faults
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,11 @@
-xournalpp (1.2.2~dev-1) UNRELEASED; urgency=medium
+xournalpp (1.2.2-1) unstable; urgency=medium
 
-  * 
+  * Fixed various memory leaks and segmentation faults
+  * Fixed some issues with input sequences mixing up devices and buttons
+  * Fixed LaTeX tool failing to open some PDF files
+  * Fixed audio output device selection
 
- -- rolandlo <roland_loetscher@hotmail.com>  Thu, 24 Aug 2023 17:22:55 +0200
+ -- Benjamin Hennion <benjamin.hennion@wanadoo.fr>  Sun, 08 Oct 2023 15:53:47 +0200
 
 xournalpp (1.2.1-1) unstable; urgency=medium
 
@@ -10,7 +13,7 @@ xournalpp (1.2.1-1) unstable; urgency=medium
   * Fixed some crashes and various issues with the crash log handling
   * Added options to the command line and preferences to disable the audio system used for audio recording
   * The official release builds are now all built with gtksourceview styling for the LaTeX tool and in "RelWithDeb" mode. Thus failing asserts do not result in a crash anymore.
- 
+
  -- rolandlo <roland_loetscher@hotmail.com>  Thu, 24 Aug 2023 17:22:54 +0200
 
 xournalpp (1.2.0-1) unstable; urgency=medium
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.2.2~dev (Unreleased)
+## 1.2.2
 
 Bugfixes from various community contributors.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## 1.2.2~dev (Unreleased)
 
+Bugfixes from various community contributors.
+
+* Fixed undefined behaviour due to signal callbacks wrong type (#5057)
+* Fixed reference handling of sidebar widgets, fixing crashes and warning (#5083,
+  #5074, #4026)
+* Default build type is now "RelWithDebInfo" (#5118)
+* Fixed error message when trying to open an autosave file (#5140)
+* Fixed empty toolbars being shown (#5160, #5051)
+* Added a device id check for input sequences, so devices don't interfere with one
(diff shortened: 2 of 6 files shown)


bhennion:
Actually, this is not in a releasable state, I think.
79f7c8cd06b causes this on linux (g++) as well
```
(com.github.xournalpp.xournalpp:3107): xopp-WARNING **: 17:54:26.521: XournalMain: System default locale could not be set.
 - Caused by: locale::facet::_S_create_c_locale name not valid
 - Note that it is not supported to set the locale using mingw-w64 on windows.
 - This could be solved by compiling xournalpp with msvc

(com.github.xournalpp.xournalpp:3107): xopp-WARNING **: 17:54:26.524: Failed to set locale : locale::facet::_S_create_c_locale name not valid
```
This is surely caused by the `std::locale::global(std::locale(".UTF-8"));` which relies on too many assumptions about the system. @Febbe could you have a look?

bhennion:
> This is surely caused by the `std::locale::global(std::locale(".UTF-8"));` which relies on too many assumptions about the system. @Febbe could you have a look?

I'm guessing the goal was to set the std::codecvt facet to something, but which?

Febbe:
`std::locale(".UTF-8")` should get the systems user locale, but as UTF-8 version. `std::locale("")` is the systems user locale.

bhennion:
> `std::locale(".UTF-8")` should get the systems user locale, but as UTF-8 version. `std::locale("")` is the systems user locale.

Sure, but we should be able to write it as std::locale("", new std::codecvt<...>) to replace the encoding facet.

rolandlo (commented):
Generally speaking this looks good to me (not regarding the localization issue, that may need a different fix). Some nitpicks: I noticed that there is a stylistic change in CHANGELOG.md and debian/changelog, using present tense instead of past (e.g. "Fix ..." instead of "Fixed ...."). For consistency I would have stuck with past tense. Also I would continue writing "LaTeX" and "PDF" and using the same capitalization for "crashlog" twice.

rolandlo:
> Actually, this is not in a releasable state, I think. [79f7c8c](https://github.com/xournalpp/xournalpp/commit/79f7c8cd06b7b65fa36cd11b72b9c292c81885d7) causes this on archlinux (g++) as well
> 
> ```
> (com.github.xournalpp.xournalpp:3107): xopp-WARNING **: 17:54:26.521: XournalMain: System default locale could not be set.
>  - Caused by: locale::facet::_S_create_c_locale name not valid
>  - Note that it is not supported to set the locale using mingw-w64 on windows.
>  - This could be solved by compiling xournalpp with msvc
> 
> (com.github.xournalpp.xournalpp:3107): xopp-WARNING **: 17:54:26.524: Failed to set locale : locale::facet::_S_create_c_locale name not valid
> ```
> 
> This is surely caused by the `std::locale::global(std::locale(".UTF-8"));` which relies on too many assumptions about the system. @Febbe could you have a look?

I also get this warning on Linux and Windows. On MacOS the warning reads slightly differently:
```term
(com.github.xournalpp.xournalpp:4378) xopp-WARNING **: 06:16:33.682: XournalMain: System default locate could not be set.
- Caused by: collate_byname<char>::collate_byname failed to construct for .UTF-8
- Note that it is not supported to set the locale using mingw-w64 on windows.
- This could be solved by compiling xournalpp with msvc
```

bhennion:
Maybe we should revert this change for now for the release, until we find a portable solution. No?

rolandlo:
> Maybe we should revert this change for now for the release, until we find a portable solution. No?

I agree. That change can wait for 1.2.3.

Febbe:
I think, reverting the whole commit is overkill. I'll create a PR which reverts the locale changes. Without c++20 / c++23 the utf8 support is bad in any way.

bhennion:
Rebased after #5220, and fixed the typesetting issues @rolandlo noted.
Could you give this one last lookover before I go a release?

debian/changelog
```diff
@@ -1,16 +1,25 @@
...
 
- -- rolandlo <roland_loetscher@hotmail.com>  Thu, 24 Aug 2023 17:22:55 +0200
+ -- Benjamin Hennion <benjamin.hennion@wanadoo.fr>  Sun, 08 Oct 2023 15:53:47 +0200
+
+xournalpp (1.2.2-1) unstable; urgency=medium
+
+  * Fix various memory leaks and segmentation faults.
+  * Fix some issues with input sequences mixing up devices and buttons
+  * Fix LaTex tool failing to open some PDF files
+  * Fix audio output device selection
```
rolandlo, line 12: Here you still use present tense.

bhennion: ^^ Sorry I missed that. Fixed

debian/changelog
```diff
@@ -1,16 +1,25 @@
...
   * 
 
- -- rolandlo <roland_loetscher@hotmail.com>  Thu, 24 Aug 2023 17:22:55 +0200
+ -- Benjamin Hennion <benjamin.hennion@wanadoo.fr>  Sun, 08 Oct 2023 15:53:47 +0200
+
+xournalpp (1.2.2-1) unstable; urgency=medium
+
+  * Fixed various memory leaks and segmentation faults.
+  * Fixed some issues with input sequences mixing up devices and buttons
+  * Fixed LaTex tool failing to open some PDF files
```
rolandlo, line 11: Sorry, one typo that I missed previously.
```suggestion
  * Fixed LaTeX tool failing to open some PDF files
```

debian/changelog
```diff
@@ -1,16 +1,25 @@
...
+xournalpp (1.2.3~dev-1) UNRELEASED; urgency=medium
 
   * 
 
- -- rolandlo <roland_loetscher@hotmail.com>  Thu, 24 Aug 2023 17:22:55 +0200
+ -- Benjamin Hennion <benjamin.hennion@wanadoo.fr>  Sun, 08 Oct 2023 15:53:47 +0200
+
+xournalpp (1.2.2-1) unstable; urgency=medium
+
+  * Fixed various memory leaks and segmentation faults.
```
rolandlo, line 9:
```suggestion
  * Fixed various memory leaks and segmentation faults
```

bhennion:
Fixed those as well. Thanks a lot for finding them!

rolandlo (approved):

PR #5220, opened 2023-10-09
Fix locale issue

Febbe:
Applies default locale "", will be overriden later with userlang.UTF_8.

Fallback order: userlang.UTF_8 -> "" -> C

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2883,12 +2883,6 @@ void Control::applyPreferredLanguage() {
 #else
     setenv("LANGUAGE", lang.c_str(), 1);
 #endif
-    try {
-        auto glob = std::locale{std::locale(".UTF-8"), lang, std::locale::all};
-        std::locale::global(glob);
-    } catch (std::runtime_error const& e) {
-        g_warning("Failed to set locale %s: %s", lang.c_str(), e.what());
-    }
 }
 
 void Control::initButtonTool() {
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -96,7 +96,7 @@ void initLocalisation() {
 
     // Not working on GNU g++(mingww) forWindows! Only working on Linux/macOS and with msvc
     try {
-        std::locale::global(std::locale(".UTF-8"));  // "" - system default locale
+        std::locale::global(std::locale(""));  // "" - system default locale
     } catch (const std::runtime_error& e) {
         g_warning("XournalMain: System default locale could not be set.\n - Caused by: %s\n - Note that it is not "
                   "supported to set the locale using mingw-w64 on windows.\n - This could be solved by compiling "


Febbe:
@bhennion, @rolandlo can you test this?

bhennion:
Seems to be working on my devices (all archlinux though)

rolandlo:
> @bhennion, @rolandlo can you test this?

I'm currently away from home without any testing devices. I will come back tomorrow though.

Febbe:
The interesting parts are actually, what happens with different inputs for default filenames especially when non ASCII but ANSI or even unicode characters are used.

The locale selection should be better now, since the systems default is chosen in any way. On Linux this will most likely be langid_STATEID.UTF-8 sometimes, it's just the C locale which is ASCII(subset of UTF-8). The fallback is always the C locale. On top the user setting regarding the locale and language is now taken into account if the operating system + stdlib supports the locale.

rolandlo:
On Linux, when I change system default to "de", I get the following warning on startup:
```term
roland@roland-ThinkPad-X1-Yoga-Gen-6:~/dev/Reviews/xournalpp/build$ ./xournalpp 
xopp-Message: 16:35:04.221: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale

(com.github.xournalpp.xournalpp:35990): xopp-WARNING **: 16:35:04.231: Failed to set user locale de: locale::facet::_S_create_c_locale name not valid
 - Falling back to default locale.
 ```
Also when I run `xournalpp` via `LC_ALL=de_DE xournalpp`
```term
(com.github.xournalpp.xournalpp:37099): Gtk-WARNING **: 16:40:26.482: Locale not supported by C library.
        Using the fallback 'C' locale.
xopp-Message: 16:40:26.522: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale

(com.github.xournalpp.xournalpp:37099): xopp-WARNING **: 16:40:26.522: XournalMain: System default locale could not be set.
 - Caused by: locale::facet::_S_create_c_locale name not valid
 - Note that it is not supported to set the locale using mingw-w64 on windows.
 - This could be solved by compiling xournalpp with msvc
 ```
That doesn't look OK to me.

**Comment by Febbe** [comment]
It seems like the locale support is just not installed on your system.

**Comment by rolandlo** [comment]
> It seems like the locale support is just not installed on your system.

How do I install it then? I'm on Ubuntu 23.10 and `locale -a` returns
```term
C
C.utf8
de_AT.utf8
de_BE.utf8
de_CH.utf8
de_DE.utf8
de_IT.utf8
de_LI.utf8
de_LU.utf8
en_AG
en_AG.utf8
en_AU.utf8
en_BW.utf8
en_CA.utf8
en_DK.utf8
en_GB.utf8
en_HK.utf8
en_IE.utf8
en_IL
en_IL.utf8
en_IN
en_IN.utf8
en_NG
en_NG.utf8
en_NZ.utf8
en_PH.utf8
en_SG.utf8
en_US.utf8
en_ZA.utf8
en_ZM
en_ZM.utf8
en_ZW.utf8
POSIX
```

Febbe:
Ok, it is installed, but you set the ANSI version. On Linux no one wants this. Try LC_ALL=de_DE.utf8

rolandlo:
> Ok, the UTF-8 version is installed, but you set the ANSI version. On Linux usually no one wants this, therefore no one installs the support for it. Try LC_ALL=de_DE.utf8 /LC_ALL=de_DE.UTF-8

Ok, that way I don't get any warnings. However that doesn't solve the issue when I set the language to "de" in the preferences.

Febbe:
> > Ok, the UTF-8 version is installed, but you set the ANSI version. On Linux usually no one wants this, therefore no one installs the support for it. Try LC_ALL=de_DE.utf8 /LC_ALL=de_DE.UTF-8
> 
> Ok, that way I don't get any warnings. However that doesn't solve the issue when I set the language to "de" in the preferences.

Of course not, thank you for the review, I created a UTF-8 based locale with `lang+"UTF-8"` but I do not use it. Can you check if it works with `langUtf8`?

rolandlo:
> > > Ok, the UTF-8 version is installed, but you set the ANSI version. On Linux usually no one wants this, therefore no one installs the support for it. Try LC_ALL=de_DE.utf8 /LC_ALL=de_DE.UTF-8
> > 
> > 
> > Ok, that way I don't get any warnings. However that doesn't solve the issue when I set the language to "de" in the preferences.
> 
> Of course not, thank you for the review, I created a UTF-8 based locale with `lang+"UTF-8"` but I do not use it. Can you check if it works with `langUtf8`?

Can you please make precise how I should check this?

src/core/control/Control.cpp
```diff
@@ -2878,16 +2878,17 @@ void Control::closeDocument() {
...
     auto const& lang = this->settings->getPreferredLocale();
+    auto const& langUtf8 = lang + ".UTF-8";  // Assuming lang has never an encoding suffix.
 #ifdef _WIN32
     _putenv_s("LANGUAGE", lang.c_str());
 #else
     setenv("LANGUAGE", lang.c_str(), 1);
 #endif
     try {
-        auto glob = std::locale{std::locale(".UTF-8"), lang, std::locale::all};
+        auto glob = std::locale{lang};
```
Febbe, line 2888:
```suggestion
        auto glob = std::locale{langUtf8};
```

Febbe (commented):
This here.

rolandlo:
> This here.

That doesn't work. It results in the same warning.

rolandlo:
It looks like we need to figure out the region code when using `std::locale`.

Febbe:
> It looks like we need to figure out the region code when using `std::locale`. 

That's hard without hard coding a map into xpp or adding a setting for this. Best we can do is coping the language and appending it as upper case.

rolandlo:
> > It looks like we need to figure out the region code when using `std::locale`.
> 
> That's hard without hard coding a map into xpp or adding a setting for this. Best we can do is coping the language and appending it as upper case.

Appending the language in upper case wouldn't work that often as it seems from the output of
```term
cat /usr/share/i18n/SUPPORTED | grep UTF-8
```

Febbe:
> > > It looks like we need to figure out the region code when using `std::locale`.
> > 
> > That's hard without hard coding a map into xpp or adding a setting for this. Best we can do is coping the language and appending it as upper case.
> 
> Appending the language in upper case wouldn't work that often as it seems from the output of
> ```term
> cat /usr/share/i18n/SUPPORTED | grep UTF-8
> ```

Yes, an austrian will not necessarily have installed de_DE he has de_AU same for swiss and other languages with multiple regions.

The question for me is now, shall I remove it completely or shall I add the same region code like the language?

For users this will have nearly zero impact. It just falls back to whatever the user sets in his user settings. Msys users won't have any benefit (supports only C and POSIX). 

For people which have the locale installed it will change it accordingly.

rolandlo:
> Yes, an austrian will not necessarily have installed de_DE he has de_AU same for swiss and other languages with multiple regions.
> 
Even worse, an Englishman or an American won't have `en_EN`, but `en_GB` and `en_US`, respectively. There are only 31 language/region combinations, where the region has the same letters like the language.

> The question for me is now, shall I remove it completely or shall I add the same region code like the language?
> 
> For users this will have nearly zero impact. It just falls back to whatever the user sets in his user settings. Msys users won't have any benefit (supports only C and POSIX).
> 
> For people which have the locale installed it will change it accordingly.

I would remove it completely. Seeing this error message (which is new for Linux and MacOS users) at startup is annoying.

src/core/control/Control.cpp
```diff
@@ -2878,16 +2878,17 @@ void Control::closeDocument() {
 
 void Control::applyPreferredLanguage() {
     auto const& lang = this->settings->getPreferredLocale();
+    auto const& langUtf8 = lang + ".UTF-8";  // Assuming lang has never an encoding suffix.
```
Febbe, line 2881:
```suggestion
```

src/core/control/Control.cpp
```diff
@@ -2878,16 +2878,17 @@ void Control::closeDocument() {
...
     setenv("LANGUAGE", lang.c_str(), 1);
 #endif
     try {
-        auto glob = std::locale{std::locale(".UTF-8"), lang, std::locale::all};
+        auto glob = std::locale{langUtf8};
         std::locale::global(glob);
     } catch (std::runtime_error const& e) {
-        g_warning("Failed to set locale %s: %s", lang.c_str(), e.what());
+        g_warning("Failed to set user locale %s: %s\n - Falling back to default locale.", lang.c_str(), e.what());
     }
```
Febbe, line 2892:
```suggestion
```

bhennion:
Shouldn't we still imbue `std::locale("")` ?

bhennion:
i.e. just reverting the changes made in the previous PR?

Febbe:
> Shouldn't we still imbue `std::locale("")` ?

We never did, and we shouldn't, since creating the locale `""` can throw an exception...
This PR just reverts the problematic changes previously made.

PR #5234, opened 2023-10-14
Fix updating of the cursor size

### LINKED ISSUE #3090: Cursor size doesn't update after using "ACTION_SIZE_..."

jdujava:
Just implemented the proposed solution in #4807, since there was no activity for a while.

Credit goes to @jen-kosek and @rolandlo.

Fixes #3090, #4807.

### DIFF
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -411,7 +411,7 @@ auto XournalppCursor::getHighlighterCursor() -> GdkCursor* {
         return createCustomDrawDirCursor(48, this->drawDirShift, this->drawDirCtrl);
     }
 
-    return createHighlighterOrPenCursor(5, 120 / 255.0);
+    return createHighlighterOrPenCursor(120 / 255.0);
 }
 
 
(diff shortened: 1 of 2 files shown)


rolandlo (changes requested):
Thanks for the PR. 
The feature works fine. Please remove the unused variable `size` from `createHighlighterOrPenCursor` though.

jdujava:
Implemented your suggestion and rebased on master.
The `if` statement whether to draw `STYLUS_CURSOR_BIG` with the current color is now based on the passed `alpha` value.

PR #5239, opened 2023-10-15
Fix ColorChooser alpha issue and selecting colors via unmber keys

### LINKED ISSUE #5236: Changing colors via keyboard bindings (numbers) no longer works.
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

dwm

X.Org

### Installation Method

source

### Xournal++ Version

1.2.2+dev (32208bb8)

### libgtk Version

3.24.38

### Bug Description

 Changing the colors via number shortcurts no longer works after merging of #5154. Just before commits corresponfing to #5154 (that is release-1.2) they work fine. 

### Expected Behaviour

Pressing `1`–`9` should appropriately change the tool color.

### Steps to Reproduce

1. Choose for example the Pen tool.
2. Try to press different numbers.
3. The color of the tool does not change.

### Additional Context

_No response_

bhennion:
Fix #5236

### DIFF
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -14,6 +14,7 @@
 #include "control/PdfCache.h"                    // for PdfCache
 #include "control/ScrollHandler.h"               // for ScrollHandler
 #include "control/ToolHandler.h"                 // for ToolHandler
+#include "control/actions/ActionDatabase.h"      // for ActionDatabase
 #include "control/jobs/XournalScheduler.h"       // for XournalScheduler
 #include "control/settings/MetadataManager.h"    // for MetadataManager
 #include "control/settings/Settings.h"           // for Settings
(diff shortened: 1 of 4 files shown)


src/core/gui/XournalView.cpp
```diff
@@ -300,11 +301,11 @@ auto XournalView::onKeyPressEvent(GdkEventKey* event) -> bool {
 
     // Switch color on number key
     auto& colors = control->getWindow()->getToolMenuHandler()->getColorToolItems();
-    if (!state && (keyval >= GDK_KEY_0) && (keyval < GDK_KEY_0 + std::min((std::size_t)10, colors.size()))) {
+    if ((keyval >= GDK_KEY_0) && (keyval < GDK_KEY_0 + std::min((std::size_t)10, colors.size()))) {
```
rolandlo, line 304: Remove this change. <Control>+number should switch between drawing types and colors should only be switched, when no modifier is pressed.

rolandlo (changes requested):
I wanted to suggest a different fix, using
```c++
 if (gtk_widget_is_sensitive(colorToolItem->getItem())) {
     control->getToolHandler()->setColor(colorToolItem->getColor(), false);
 }
```
But yours is more direct. However don't remove `!state`.

src/core/gui/XournalView.cpp
rolandlo: Maybe then we need to mask out consumed modifiers like in `TextEditor::onKeyPressEvent`
>
    /*
    According to https://docs.gtk.org/gdk3/method.Keymap.translate_keyboard_state.html
    consumed modifiers should be masked out. For instance, on a US keyboard, the plus symbol is shifted, so when
    comparing a key press to a <Control>plus accelerator <Shift> should be masked out.
    */

The Ctrl+NUMBER accelerator is set in `main.glade`, e.g.
```glade
                            <signal name="toggled" handler="ACTION_TOOL_DRAW_SPLINE" swapped="no"/>
                            <accelerator key="8" signal="activate" modifiers="GDK_CONTROL_MASK"/>
```

rolandlo (approved):
LGTM

PR #5240, opened 2023-10-15
Fix Spline live segment display

### LINKED ISSUE #5237: Spline indicator line is not working/moving
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

dwm

X.Org

### Installation Method

source

### Xournal++ Version

1.2.2+dev (32208bb8), but also isn't working on release-1.2.2

### libgtk Version

3.24.38

### Bug Description

When drawing with the `Draw Spline` tool, the indicator line (interpolating from the last placed node to the cursor) is not shown (more precisely it is hidden, see the screenshots below).

The indicator line is hidden behind the green slope indicator of the last (right) node.

This can be seen by moving the node using arrows (pressing `Left` couple times).

Even after deleting the node via `Backspace`, the indicator line points to the position of the last placed node.
image

### Expected Behaviour

The spline indicator line should be properly following the mouse cursor.

### Steps to Reproduce

1. Choose `Draw Spline` tool.
2. Draw couple of nodes.
3. Notice the lack of indicator line.
4. Reveal that it is stuck pointing to the place of the last node by moving the node via arrows.

### Additional Context

_No response_

bhennion:
Fix #5237

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -501,7 +501,7 @@ auto XojPageView::onButtonTriplePressEvent(const PositionInputData& pos) -> bool
 }
 
 auto XojPageView::onMotionNotifyEvent(const PositionInputData& pos) -> bool {
-    if (currentSequenceDeviceId != pos.deviceId) {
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
LGTM.

PR #5243, opened 2023-10-16
correct fom to from in toolbar customize

halfhiddencode:


rolandlo:
Thanks for making aware of the typo @halfhiddencode

PR #5248, opened 2023-10-16
Fix SegFault when copying the entire text in TexDialog

### LINKED ISSUE #5231: Xournal crashing when copying code from the latex dialog
### Operating System

Linux

### (Linux only) Distribution

Ubuntu

Unity

X.Org

### Installation Method

PPA

### Xournal++ Version

1.2.1+dev

### libgtk Version

3.24.20

### Bug Description

The program crashes when I try to copy a code from the latex dialog. It works fine the first few times I do it and then it crashes.

### Expected Behaviour

I should to be able to copy as many times as I want from the Latex dialog without Xournal to crash.

### Steps to Reproduce

1. Write a code at the Latex dialog.
2. Copy this code to paste in another Latex dialog. It will work the first few times.
3. Eventually, Xournal crashes when doing the ctrl+c.

### Additional Context

Below is the errorlog attached.

[errorlog.20231012-180141.log](https://github.com/xournalpp/xournalpp/files/12887234/errorlog.20231012-180141.log)

bhennion:
Fix #5231

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2322,14 +2322,10 @@ void Control::fontChanged(const XojFont& font) {
  */
 void Control::runLatex() {
     /*
-     * LatexController::run() will open a non-blocking dialog. This dialog uses the LatexController instance until its
-     * completion. The LatexController instance must thus outlive this scope. It'll get deleted when the dialog is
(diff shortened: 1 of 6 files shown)


rolandlo (approved):
I can confirm that this PR solves the issue.

Issue #4431, opened 2022-11-20
Mouse wheel not scrolling


failingapbio:
Hi,
I use Xournalpp 1.1.2 on MacOs Catalina
Mouse wheel doesn't scroll on the document. Mouse wheel only scrolls when mouse in on the scroll bar on the right of the window.
What I expect: scroll up and down when I use mouse wheel in the document. Bug or bad configuration?

Affects versions : 1.1.2
Mac

To Reproduce
Open Xournal++ and scroll on a PDF. It doesn't scroll unless the cursor is on the bar.
Expected behavior
Open Xournal++ and scroll on a PDF. It should scroll all the time when the cursor is on the page.

Technius:
I can confirm this on Monterey as well, not sure what causes this. There is only one input device reported ("Quartz device"), which was set to Pen by default for me. Changing the device type to mouse does not fix the issue.

TheMathGuyd:
Just here to confirm that this is still a bug in Ventura. Scrolling does nothing

MahdeenSky:
Has this been resolved? This is quite the problem.

hnykda:
Confirm the same on Ventura.

advaypakhale:
I can confirm the same issue on Monterey. I'm on the Mar 19 nightly build.

I'm using a Wacom Intuos S tablet, which shows up as "Quartz Pen" and "Quartz Eraser" in the input devices settings.

The trackpad doesn't work for scrolling, neither does attaching an external mouse with a scroll wheel help. I scoured the preferences to see if I was missing anything; I haven't seen any relevant settings, however. I could be entirely wrong.

The only way to scroll is either to use the arrow keys, or to manually click and drag the scroll bar. After doing the latter and attempting to write, the first few strokes are still registered as moving the scroll bar, which causes the page to jump erratically.

As someone who uses the trackpad extensively to zoom in and position my cursor while writing, this issue is rendering the program basically unusable. I would greatly appreciate it if this issue could be fixed or if a change to the default preferences is suggested. Thanks.

Technius:
Is everyone affected using Apple Silicon? The nightly build is compiled an on x86 machine and therefore is run using Rosetta 2. Scrolling does not work with the nightly build. However, when I manually compiled Xournal++ on my Mac M1 device, scrolling does work.

MahdeenSky:
> 

Yeah, I'm on an M1, could you provide the manually compiled Xournal? My cmake fails while doing so.

[...]

Issue #1269, opened 2019-06-18
[Question] Cannot open remote files


AlessandroW:
**Affects versions :**
 - OS: Arch Linux 
 - Xournal++ 1.0.12

**Describe the bug**
Cannot open remote files. 

**To Reproduce**
1. Open Gnome-Control-Center
2. Log in to a Google-Drive account
3. Open a remote .xopp file

**Expected behavior**
Usual workflow.

**Screenshots of Problem**
![image](https://user-images.githubusercontent.com/29565912/59678573-18171580-91cd-11e9-932e-a337546c10b1.png)

- Why is this not possible? 
- Is it on the roadmap?

AlessandroW:
Can someone add some information why it is not possible?
If it is not too difficult maybe I could implement it.

LittleHuba:
@andreasb242 Could you answer this?

andreasb242:
Xournal++ used first the standard ZLib API, not the GLib wrapper, and they don't support URLs, Gnome 2 returned the real URLs from Open Dialog, like smb://server/share/file.xoj, and this cannot be opened.

File reading / writing was completely refactored, so may this is no issue anymore. May also Gnome 3 handle this different.

Maybe only add the remote supported flag to the Open Dialog, and everything works, probably nobody has tried this since years ;-)

andreasb242:
XojOpenDlg.cpp

>gtk_file_chooser_set_local_only(GTK_FILE_CHOOSER(dialog), true);

Here is the question, what is returned now?
Theoretically there should always be a local path, e.g. if I mount an FTP, it ismounted local at:
/run/user/<my-user-id, e.g. 1000>/gvfs/ftp:host=192.168.1.2,port=123

and this is a local mounted path, so if we can get this Path from Filechooser, it should work without changes of the file loading / writing, because we don't use GTK file API.

LittleHuba:
The way to go is using GIO for this as mentioned in the documentation. Unfortunately the libraries required for the new fileformat are C libraries and therefore incapable of using GIO.

Only one solution comes to mind but this is really ugly:
We could use GIO to download remote files to a local directory and then open the file with the normal routines. For saving we could also use this abstraction.

andreasb242:
For a Java Application I did simple a hack, and resolve a mountpoint, then you have a local path, and it's working. This should also be possible for this case.

embeddedt:
[...]

PR #5256, opened 2023-10-19
Make MacOS Finder open xopp-files

### LINKED ISSUE #5229: Cannot open any file using `open with` on mac
### Operating System

Mac OS

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Official release

### Xournal++ Version

1.2.1

### libgtk Version

3.24.38

### Bug Description

On a mac (using Sonoma with a M2 processor), even after giving full disk access to xournal++, files cannot be opened using finder. They can only be opened when navigating to the file using the `open file` menu.

Opening any xopp file will result in this error 
image

### Expected Behaviour

The file is opened and no spurrious error of the type "incompatible file type" should appear.

### Steps to Reproduce

1. Have a valid `.xopp` file
2. Open the file with `xournal++` from finder
3. See an error message 

### Additional Context

_No response_

rolandlo:
Finder uses Launch Services to identify Xournal++ as the preferred app for xopp-files as declared in the Info.plist, to launch it and send it an "Open Documents" event. Gtk receives the event and uses g_application_open to make the handler for the "open" signal react to it. See gtkapplication-quartz.c.

Fixes #5229.

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -442,9 +442,10 @@ void on_open_files(GApplication* application, GFile** files, gint numFiles, gcha
             XojMsgBox::showErrorToUser(GTK_WINDOW(app_data->win->getWindow()), msg);
         }
     } catch (const fs::filesystem_error& e) {
-        const std::string msg = FS(_F("Sorry, Xournal++ cannot open remote files at the moment.\n"
-                                      "You have to copy the file to a local directory.") %
-                                   p.u8string() % e.what());
+        const std::string msg = FS(_F("Filesystem error: {1}\n"
+                                      "Sorry, Xournal++ cannot open the file: {2}\n"
+                                      "Consider copying the file to a local directory.") %
+                                   e.what() % p.u8string());
         XojMsgBox::showErrorToUser(GTK_WINDOW(app_data->win->getWindow()), msg);
     }
     gtk_window_present(GTK_WINDOW(app_data->win->getWindow()));
@@ -539,9 +540,10 @@ void on_startup(GApplication* application, XMPtr app_data) {
                 opened = app_data->control->newFile("", p);
             }
         } catch (const fs::filesystem_error& e) {
-            const std::string msg = FS(_F("Sorry, Xournal++ cannot open remote files at the moment.\n"
-                                          "You have to copy the file to a local directory.") %
-                                       p.u8string() % e.what());
+            const std::string msg = FS(_F("Filesystem error: {1}\n"
+                                          "Sorry, Xournal++ cannot open the file: {2}\n"
+                                          "Consider copying the file to a local directory.") %
+                                       e.what() % p.u8string());
             XojMsgBox::showErrorToUser(GTK_WINDOW(app_data->win->getWindow()), msg);
             opened = app_data->control->newFile("", p);
         }
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -50,7 +50,8 @@
 
 namespace {
 
-constexpr auto APP_FLAGS = GApplicationFlags(G_APPLICATION_SEND_ENVIRONMENT | G_APPLICATION_NON_UNIQUE);
+constexpr auto APP_FLAGS =
+        GApplicationFlags(G_APPLICATION_SEND_ENVIRONMENT | G_APPLICATION_NON_UNIQUE | G_APPLICATION_HANDLES_OPEN);
 
 /// Configuration migration status.
 enum class MigrateStatus {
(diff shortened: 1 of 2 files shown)


rolandlo:
@Doublonmousse Please test the MacOS app-bundle generated below. It should open xopp-files in Finder.

LittleHuba:

- [Windows](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjg2OTcvYXJ0aWZhY3ROYW1lL3dpbmRvd3M1/content?format=zip)
- [macOS](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjg2OTcvYXJ0aWZhY3ROYW1lL21hY09T0/content?format=zip)
- [Debian](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjg2OTcvYXJ0aWZhY3ROYW1lL2xpbnV4LWRlYmlhbi1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 20.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjg2OTcvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMC4wNC1wYWNrYWdlcw2/content?format=zip)
- [Ubuntu 22.04](https://artprodsu6weu.artifacts.visualstudio.com/A4bb312fe-95d7-456e-96fb-7669bb27c94b/ce85de88-4b71-44a0-8a57-2707c0b9a6e7/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3hvdXJuYWxwcC9wcm9qZWN0SWQvY2U4NWRlODgtNGI3MS00NGEwLThhNTctMjcwN2MwYjlhNmU3L2J1aWxkSWQvMjg2OTcvYXJ0aWZhY3ROYW1lL2xpbnV4LXVidW50dS0yMi4wNC1wYWNrYWdlcw2/content?format=zip)

Doublonmousse:
Yes, it fixes the issue.

Somehow it also fixes #4431 as well on my machine ...

src/core/control/XournalMain.cpp
```diff
@@ -424,10 +425,29 @@ void on_command_line(GApplication*, GApplicationCommandLine*, XMPtr) {
...
+            app_data->control->openFile(p);
+        } else {
+            const std::string msg = FS(_F("File {1} does not exist.") % p.u8string());
+            XojMsgBox::showErrorToUser(GTK_WINDOW(app_data->win->getWindow()), msg);
+        }
+    } catch (const fs::filesystem_error& e) {
+        const std::string msg = FS(_F("Sorry, Xournal++ cannot open remote files at the moment.\n"
+                                      "You have to copy the file to a local directory.") %
+                                   p.u8string() % e.what());
+        XojMsgBox::showErrorToUser(GTK_WINDOW(app_data->win->getWindow()), msg);
```
Febbe, line 448: I would use a more generic error message.

I think, `std::filesystem` should handle remote files over any NFS like SMB/CIFS. So when we can't open it, it should fail because of some GIO stuff. The second is, that there are many other filesystem errors.

Febbe (commented):
This should also change the behavior on other operating systems. For now I can only test this on windows. Linux should be tested also.

Febbe:
> Yes, it fixes the issue.
> 
> Somehow it also fixes #4431 as well on my machine ...

Should not be the reason. I think it could be the UB I fixed in a previous commit. Some eventhandlers returned a bool but were casted to a function returning a gboolean, which is UB.

Doublonmousse:
I did a quick test on pop os 22.04 and did not encounter any issue opening files (neither from the app menu nor from nautilus)

rolandlo:
> This should also change the behavior on other operating systems. For now I can only test this on windows. Linux should be tested also.

The only place in the Gtk sources, where `g_application_open` is called, is in `gtkapplication-quartz.c`. This only affects MacOS.
The only place in the Glib sources, where `g_application_open` is called, is in `g_application_real_local_command_line`, but only in the case when there are unparsed remaining command line options (otherwise `g_application_activate` is called). Since we already parse remaining command line options in
https://github.com/xournalpp/xournalpp/blob/7f5cd26c751e5e02006792121f80d23d3c26abb9/src/core/control/XournalMain.cpp#L636-L637
we never get `g_application_open` through `g_application_real_local_command_line` for Xournal++.

So I don't think that other OS'es are affected.

Opening files on Gnome or KDE from the file manager is governed by desktop files and so simply pass the file(s) to open as argument(s) to the executable listed in the `Exec` key. On Windows it's similar, just that desktop files are replaced by registry entries.

Just in case, I have tested it on Ubuntu 23.10 and it works all fine.

src/core/control/XournalMain.cpp
rolandlo: That was copy-pasted from `on_startup`. I will change both messages simultaneously.

About remote files, see #1269.

rolandlo: Done.

Febbe:
Thank you for the explanation, that is probably the reason, why I did not implement the function back then.

rolandlo:
Thanks for the review @Febbe. Merging in 24 hours if no objections are raised.

PR #5258, opened 2023-10-19
Improve arrow shape with respect to fill

rolandlo:
Filled arrows and double arrows currently look asymmetrical. This commit fixes that by adding an additional point. Instead of
`tail -> head -> side 1 of head -> head -> side 2 of head`
it's now
`tail -> head -> side 1 of head -> head -> side 2 of head -> head`

As a consequence filled and unfilled arrows looks pretty much the same (although they behave differently when something from the head is erased).


rolandlo:

rolandlo:
I rebased on release-1.2, since it's a bugfix.

Issue #3293, opened 2021-07-31
Latex inserts line breaks for some "long" expressions


NelkonParker:
**Affects versions :**
 - OS: Windows 10
 - Which version of libgtk do you use: 3.24.20
 - Version of Xournal++: 1.1.0 (with hotfix)
 - Installation method:  official release

**Describe the bug**

Latex expressions seem to have a hard-coded length limit, after which line breaking will occur.

When entering latex in the "Insert latex" dialog, "long" expressions are line-broken at some apparently pre-determined limit. I can't figure out what this limit is, however. 

The bug may depend on whether or not the emitted latex contains white space, so for example, the input:

"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

does not get a line break (but is truncated at some length) whereas

"x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x=x="

does get a line break. 

I can find no way to override this behaviour.

**To Reproduce**
See attached screen recording.

https://user-images.githubusercontent.com/88211140/127740325-7066fc09-884b-4096-b300-5278ca3bd8aa.mp4

**Expected behavior**
I would prefer no upper bound on the size of generated latex output.

**Screenshots of Problem**
See attached screen recording.

**Additional context**
This behaviour was not present in the versions of Xournal++ immedately prior to the 1.1.0 series.

tattsan:
In default_template.tex, standalone.cls is called with a option `varwidth=true` :

```
\documentclass[varwidth=true, crop, border=5pt]{standalone}
```
If you prepare your own tex-template with the option `varwidth=false` ,  you get a pdflatex output with no breaks.
You may feel the option causes the opposite effect, but the option `varwidth=true` of `standalone.cls` is intended to suppress a short line to be stretched to paper width `( \linewidth )` .

By the way, this default is inconvinient for my use, so I use my own tex-template with  the option `varwidth=false`.

NelkonParker:
Thanks. 

After changing the template file, I can confirm that varwidth=false makes Xournalpp revert to the previous behaviour. I guess someone changed this for the 1.1.0 release for some reason. 

[...]

PR #5264, opened 2023-10-21
Fix long text being truncated in LaTeX tool

tattsan:
This change was proposed in #3293.  Based on release-1.2.

### DIFF
--- a/resources/default_template.tex
+++ b/resources/default_template.tex
@@ -1,5 +1,11 @@
 % This template uses the scontents package, which is only available on relatively recent TeX distributions. In case it is not available on your system, use the legacy_template.tex
-\documentclass[varwidth=true, crop, border=5pt]{standalone}
+\documentclass[varwidth=0.999\maxdimen, crop, border=5pt]{standalone}
+% The upper limit value of 'varwidth' can be referenced by \hsize.
+\newcommand*{\setTextWidthReference}{%
+  \setlength{\textwidth}{345.0pt}% Same value when you use 'varwidth=true'.
+  \setlength{\linewidth}{\textwidth}%
+  \setlength{\columnwidth}{\textwidth}%
+}
 
 % Packages
 \usepackage{amsmath}
@@ -25,6 +31,7 @@
 \end{scontents}
 
 \begin{document}
+  \setTextWidthReference
   % Check if the formula is empty
   \settoheight{\pheight}{\getstored[1]{preview}}%
   \ifthenelse{\pheight=0}{\GenericError{}{xournalpp:blankformula}{}{}}
--- a/resources/legacy_template.tex
+++ b/resources/legacy_template.tex
@@ -1,5 +1,11 @@
 % This template avoids the scontents package, which is only available on relatively recent TeX distributions
-\documentclass[varwidth=true, crop, border=5pt]{standalone}
+\documentclass[varwidth=0.999\maxdimen, crop, border=5pt]{standalone}
+% The upper limit value of 'varwidth' can be referenced by \hsize.
+\newcommand*{\setTextWidthReference}{%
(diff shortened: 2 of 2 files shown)


rolandlo:
Can you please elaborate why you prefer `varwidth=false` to `varwidth=\maxdimen` as was suggested in https://github.com/xournalpp/xournalpp/issues/3293#issuecomment-1619194026?

tattsan:
Note that no `varwidth` option is equivalent to `varwidth=false`.
We can see the difference between  `varwidth=false` and `varwidth=\maxdimen` when a TeX code contains  paragraph skips. Consider the following 3 sources.

1. 
```
a\)\(b
```
2.
```
a\)
\(b
```
3.
```
a\)

\(b
```
First two contain no paragraph skip, and the result is same for both `varwidth` option.

1. 
AB_H_nonSP

2. 
AB_H_withSP

The third code contains a paragraph skip. If you use `varwidth=false`, the result is the same as 2. If you use `varwidth=\maxdimen`, the result is the following:

3. when `varwidth=\maxdimen`
AB_V

If you want to write a code which exit math mode like above,  `varwidth=\maxdimen` is preferable. However, if you want to create paragraphs, I think it's better to explicitly use the minipage environment.

tattsan:
The above result shows `varwidth=false` is incompatibile with the current version.
However `varwidth=\maxdimen` is also incompatible in that it changes the value of `\linewidth`.

* With `varwidth=true/false` the code `\the\linewidth` shows the value `345.0pt`,
* With `varwidth=\maxdimen` the code `\the\linewidth` shows the value `16383.99998pt`.

For those who have been using the ‘\linewidth`'s value, this is inconvenient.

tattsan:
If `varwidth=false`, the `varwidth` package will not be loaded and you will not be able to use the `varwidth` environment.
So we should add `\RequirePackage{varwidth}` in this case.

tattsan:
The `\linewidth`  is essentially meaningless, since we are trying to eliminate width limitations for TeX objects. 
On the other hand, however, some of you who have been inserting coherent texts in the minipage environment, 
for example, may have specified the size in multiples of `\linewidth`. This can be reconciled by changing the value of
 `\linewidth` in the middle. This does not affect the width limit of the TeX object.
```
\begin{document}
  \setlength{\linewidth}{345.0pt}%
  % Check if the formula is empty

  % Render the user input
  \textcolor{xpp_font_color}{\getstored[1]{preview}}
\end{document}
```

However, it is not a good idea to write a magic number of 345.0pt (the default value of `\linewidth` of the article.cls), 
so it is better to refer to some value (such as the setting of standalone.cfg or the size of the background paper in Xournal++).

tattsan:
I have tried to keep compatibility with the previous version and set varwidth near to maxmum.
If the value of varwidth is `\maxdimen`, an error occurs when the width of the content is close to `\maxdimen`, 
so I set it to `0.999\maxdimen`.  The `\textwidth`, `\linewidth`, and `\columnwidth`, which may be referred to as text width,
are the same values as before. The upper limit of `varwidth` matches  the value of`\hsize`.

tattsan:
> If the value of varwidth is \maxdimen, an error occurs when the width of the content is close to \maxdimen,

This error occurs because the width of the box becomes `\maxdimen + 2*border` .

> so I set it to 0.999\maxdimen

So this can be maximized by the following:
```
\makeatletter
  \setlength{\hsize}{\maxdimen}%
  \addtolength{\hsize}{-\sa@border@left}%
  \addtolength{\hsize}{-\sa@border@right}%
}
\makeatother
```
which should be executed  after `\begin{document}`.

rolandlo (approved):
LGTM

tattsan:
Thanks!

rolandlo:
Thanks @tattsan for this fix!

Issue #5269, opened 2023-10-21
Zoom center is not correct when it is not on the active page


tmoerschell:
### Operating System

Linux

### (Linux only) Distribution

openSUSE Tumbleweed

### (Linux only) Desktop Environment

KDE Plasma

### (Linux Only) Display Server

Wayland

### Installation Method

Build from source

### Xournal++ Version

1.2.1+dev (889aefd3)

### libgtk Version

3.24.38

### Bug Description

When zooming to a point that is not on the currently active page (the one with a red border), the center will slightly shift and not stay in place. 

### Expected Behaviour

The zoom center should stay at the same position. 

### Steps to Reproduce

1. Create multiple pages and zoom out so you see a few
2. Zoom in or out to a position not on the currently active page (easiest with `ctrl`+scroll)
3. Mouse pointer position will shift slightly

### Additional Context

I have already fixed the issue and will propose a PR shortly

PR #872, opened 2019-02-12
Multilayout - more columns, more arrangement of pages.

JJones780:
Kudos to the xournal++ team for an excellent program! 
I've added a few capabilities for my own use ( but  got carried away ) and cleaned it up for inclusion if it meets your standards.
I'm able to make changes for the next short while.. so please let me know.

Multilayout:
Added capability to have many columns (pairing works as expected - will force even # columns).
Added capability to order pages horizontally or vertically and order them: left-to-right/right-to-left, top-to-bottom/bottom-to-top combinations .

More potential arrangements easy to code in ( like 2 columns of 4 across, or randomized order !?!  etc). Please let me know.

Also coded for fixed rows instead but not in menu yet.

Added setting for page offset: 0 or 1 makes sense but larger numbers work depending on columns etc.

New Preferences/View setting:
      Pairs Offset: First Page Offset
New View menus: 
Columns: 1,2,3,4,5,6,7,8
Layout:
  Horizontal/Vertical
  Left to Right/Right to Left
  Top to Bottom/Bottom to Top

TODO:  A few more gui additions.
Note: simple linear search works well enough that I haven't gotten around to a proper quadtree or other to replace the binary search ( which was modified to do paired pages ).

PR #5270, opened 2023-10-21
Align pages horizontally and center them vertically

### LINKED ISSUE #5268: Page layout is unaligned and pages are not centered vertically
### Operating System

Linux

### (Linux only) Distribution

openSUSE Tumbleweed

KDE Plasma

Wayland

### Installation Method

Build from source

### Xournal++ Version

xournalpp 1.2.1+dev (889aefd3)

### libgtk Version

libgtk: 3.24.38

### Bug Description

When pages of different sizes are layout in multiple rows, pages are not aligned as they should (there is some padding missing on the more narrow pages). 
Additionally they are not centered vertically. 

### Expected Behaviour

All layout grid cells should be aligned, and pages should be centered horizontally and vertically within them. 

### Steps to Reproduce

1. Set layout to 3-4 columns
2. Open a document with different page sizes, eg. my beautiful physics notes
not_really_pretty_layout

### Additional Context

I have a fix ready. Looks like this:
better_layout
I know that horizontal alignement and vertical centering are kind of separate issues but I think it looks much nicer like this. 
PR incoming

tmoerschell:
Fixes #5268. 
I'm not too sure what branch to base it on. It's currently on master, but it's possible to rebase it on release-1.2. I took it on as a sub-issue of #5269, as it required knowing all about the padding around the pages. It would be nice to be able to merge this one and then use it as a base for the zoom center fix.


tmoerschell:
Oh I'm sorry, I didn't know this was actually the wanted behavior. The current implementation looks kinda odd: it does include some padding to compensate for different page widths, but not enough to align the pages. I have now also checked the PR that introduced this code (#872), and it doesn't seem to mention a reason for this. 

About the branch issue, I agree that it isn't a good idea to merge this into the release branch. But the PR I want to build upon afterwards was, and the changes aren't in the master branch. So I wonder: do those get merged into the master branch as well after some time? Why not right away?


The use case I have is inserting portrait A4 pages between 3:2 slides to take additional notes. 
Adding some extra space to PDF pages or rotating them is certainly also something that could happen in the future. 

I personally don't actually use the multi-column layout. I just came across it trying to fix another issue. I'm always writing something on a single page, so I rarely have multiple pages in my view anyways. I'd think what is happening between the pages is of little importance to most users. For me, it should just be as pretty as possible. 

Anyways, I think there's mainly two options you could go about the layout:
 * Keep a constant padding between pages, disregarding alignment issues. A horizontal layout may have rows of different lengths and a vertical layout may have columns of different heights. 
 * Put the pages into a simple grid, disregarding padding consistency within rows and columns. 

One issue with the grid would certainly be that if one has a page that is *a lot* bigger than the others, it will cause big gaps between other pages. But who would ever do that? 

I don't know what an infinite canvas would look like, or if there already is an idea about how to implement it. A simple version of that could be just pages with an arbitrarily large, automatically extending size. This could cause wildly different page sizes. I guess a solution to that could then be to crop/scale over-sized pages when zoomed out.

rolandlo:
> I agree it is kinda odd. To be clear, I don't think your changes are not a good idea, but I think we should investigate how it'll influence various use cases (if we can find someone actually using it).

I do use a 1-row or 2-row layout on my external monitor for displaying my solutions when correcting an exam. Usually the pages have all the same format though. Different page formats can appear when you have a document, where some pages are landscape and others are portrait. Another case may appear when the PDF was generated from (camera or scanned) images. There are also some books and journals where some of the first pages have a different format.

The solution proposed in this PR looks logical from the point of view that all layout settings are symmetrical with respect to vertical/horizontal. But the current layout method has its advantages, too.

tmoerschell:
Ok, finally made some progress with this: I implemented constant padding layout. 
It's the first time I touched something on the GUI / action database side. I hope I didn't miss too many things. 

There are some parts of the code that I'm not too fond of (especially the repeated loops in `Layout.cpp`). I'm sure it must be possible to this in a better way, but couldn't find it. 

Do also note that this is not fully functional yet: the visual part is good, but it is in some cases not possible to draw on the pages, for example, because most notably I haven't adapted `Layout::getPageViewAt(x, y)` yet.

tmoerschell:
Ok, everything is good now, except for `getPaddingAbovePage()` and `getPaddingLeftOfPage()`, which I plan to change in #5295 anyways, so I'd leave it for then. 

Can't guarantee that I caught everything, but I will put it up for review again.

src/core/gui/Layout.cpp
```diff
@@ -254,57 +285,163 @@ void Layout::layoutPages(int width, int height) {
...
+                }
+
+                // center page vertically
+                pages[index].paddingTop = (this->pc.heightRows[r] - pages[index].height) / 2.0;
+
+                pages[index].width = pc.widthCols[c];
+                pages[index].height = pc.heightRows[r];
+            }
+        }
+    } else if (this->layoutType == LAYOUT_TYPE_CONST_PADDING) {
```
tmoerschell, line 353: rather `else if (...)` or a simple `else` like on line 428? In any case it should be consistent.

tmoerschell:
I've improved some of the things I was unhappy with now, so I'd like to request a review from someone on this :-)

PR #5288, opened 2023-10-25
Identify Touchpad and Trackpoint as same device

### LINKED ISSUE #5287: Device hash check breaks touchpad+trackpad combo
This breaks my workflow at least. I have a thinkpad that has both a touchpad and one of those red trackpoints. Like this:

sef

The thing is, the buttons above and below the touchpad belong to the two different devices. The above two buttons are for the trackpoint, the below for the touchpad. I used to be able to press whichever, if I wanted to draw a line in xournal++. Now the above two don't work, for it's a different device. What do?
Many thanks @bhennion

_Originally posted by @MikuChan03 in https://github.com/xournalpp/xournalpp/issues/5170#issuecomment-1778681579_

bhennion:
Fix #5287 

@rolandlo could you check if this indeed works? I don't have such a device.

### DIFF
--- a/src/core/gui/inputdevices/DeviceId.cpp
+++ b/src/core/gui/inputdevices/DeviceId.cpp
@@ -1,0 +1,17 @@
+#include "DeviceId.h"
+
+#include <gdk/gdk.h>
+
+DeviceId::DeviceId(const GdkDevice* id): id(id) {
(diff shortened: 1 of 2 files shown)


MikuChan03:
Thank you so much!

PR #5108, opened 2023-08-31
Rebased: Fix zoom jumping and zoom steps (Partial Zoom Refactor)

michael-markl:
This PR rebases the stale PR #2837 onto the latest master in a hope to speed up fixing #4678.

When rebasing, I solved some emerging conflicts. I also applied the suggestions by @personalizedrefrigerator and @LittleHuba given in the original PR.

### Original PR description:

This is a splitup part of https://github.com/xournalpp/xournalpp/pull/2743 .
see https://github.com/xournalpp/xournalpp/pull/2743 for older comments.

This patch fixes the issues with zoom jumping to previous/different pages. It also changes zoom steps to be relative for a much better zooming experience.

Internally this is a change from local to global coordinates, because the local coordinates had weird position spikes in them. This is honestly mostly due to my missing understanding of where the local coordinates come from. I am sure that part of this patch can be done differently.

I have also simplified some of the logic to avoid addition/subtraction of large floating point values and simpler access of the zoom calls.

PR #5295, opened 2023-10-27
Fix zoom center

### LINKED ISSUE #5269: Zoom center is not correct when it is not on the active page

tmoerschell:
Fixes #5269

This is depending on two other things: 

* [x] Merging commits from #5108 from release-1.2 to master (there is just one `assert()` that has to be changed to `xoj_assert()` for this)
* [ ] We have to clear up what we want to do in PR #5270. If there are changes to that, this PR will have to be reworked as well. 

I have included the commits from the above points for now, so you can see / test the changes. I will rebase/do whatever, when we decide to move forward and have an appropriate base to merge upon. As such, the only really relevant commit here is the last one (017450c). 

Question to maintainers: when should I open a PR for merging the commits of #5108 onto master? Now, when the other issue is cleared up, or not at all (merging them through this PR)?

### DIFF
--- a/src/core/control/zoom/ZoomControl.cpp
+++ b/src/core/control/zoom/ZoomControl.cpp
@@ -146,23 +146,21 @@ void ZoomControl::startZoomSequence(utl::Point<double> zoomCenter) {
     this->zoomWidgetPos = zoomCenter;  // widget space coordinates of the zoomCenter!
     this->zoomSequenceStart = this->zoom;
 
-    // * set unscaledPixels padding value
-    size_t currentPageIdx = this->view->getCurrentPage();
-
     // To get the layout, we need to call view->getWidget(), which isn't const.
     // As such, we get the view and determine `unscaledPixels` here, rather than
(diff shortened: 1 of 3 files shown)


tmoerschell:
Rebased onto master now. Thanks @bhennion !

PR #5297, opened 2023-10-27
Fix UI stuff

rolandlo:
Most importantly this fixes a regression from #4763 in the export dialog, which made it impossible to set the quality parameter for png-export.


rolandlo:
Thanks for the review.

Issue #4917, opened 2023-06-08
IDK, i installed Xournal++ then the Nighty version and I am getting this bug report.


yadavmayank742:
Date: Fri Jun  2 22:26:51 2023
Error: signal 11
[bt]: (0) xournalpp(+0x1deb0b) [0x55ffa2dbdb0b]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x3bf90) [0x7fa52a45af90]
[bt]: (2) xournalpp(_ZNK7Element7getTypeEv+0) [0x55ffa2d92030]
[bt]: (3) xournalpp(_ZNK12DocumentView11drawElementEP6_cairoP7Element+0x15) [0x55ffa2db5055]
[bt]: (4) xournalpp(_ZN12DocumentView9drawLayerEP6_cairoP5Layer+0x84) [0x55ffa2db51e4]
[bt]: (5) xournalpp(_ZN12DocumentView8drawPageESt10shared_ptrI7XojPageEP6_cairobbbb+0x106) [0x55ffa2db5bf6]
[bt]: (6) xournalpp(_ZN10PreviewJob8drawPageEv+0x37f) [0x55ffa2cbf87f]
[bt]: (7) xournalpp(_ZN10PreviewJob3runEv+0x28) [0x55ffa2cbfbc8]
[bt]: (8) xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0xf8) [0x55ffa2cc2608]
[bt]: (9) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x865cd) [0x7fa52bbd35cd]
[bt]: (10) /lib/x86_64-linux-gnu/libc.so.6(+0x88fd4) [0x7fa52a4a7fd4]
[bt]: (11) /lib/x86_64-linux-gnu/libc.so.6(+0x1095bc) [0x7fa52a5285bc]

Try to get a better stracktrace...
[bt] #1 xournalpp(+0x1df101) [0x55ffa2dbe101]
??:0
[bt] #2 /lib/x86_64-linux-gnu/libc.so.6(+0x3bf90) [0x7fa52a45af90]
??:0
[bt] #3 xournalpp(_ZNK7Element7getTypeEv+0) [0x55ffa2d92030]
??:0
[bt] #4 xournalpp(_ZNK12DocumentView11drawElementEP6_cairoP7Element+0x15) [0x55ffa2db5055]
??:0
[...]

PR #5317, opened 2023-11-04
Refactor EditSelection contructors

bhennion:
This PR refactors the way selections are contructed, in order to remove some race conditions and optimize (quite a bit) the cost of selecting elements (in layers with a lot of elements). More specifically:
 - Properly lock the document mutex when needed (will surely help with #4917
 - Optimized out 2 out of 3 loops looking for elements in a layer
 - Avoid adding elements to the layer just before removing them for selection
 - Make SelectionFactory to have explicitly named "contructors"

I tested quite a bit, but there are many different ways of getting a selection: selection tools (+ shift to aggregate or multilayer), LaTex editor, Image insertion, switching from TextEditor to any selection tool, pasting from clipboard and others I may have forgotten. I possibly missed a couple of corner cases.

This is up for review.

### DIFF
--- a/src/core/control/UndoRedoController.cpp
+++ b/src/core/control/UndoRedoController.cpp
@@ -10,6 +10,7 @@
 #include "model/Document.h"               // for Document
 #include "model/Layer.h"                  // for Layer
 #include "model/PageRef.h"                // for PageRef
+#include "model/XojPage.h"                // for XojPage
 #include "undo/UndoRedoHandler.h"         // for UndoRedoHandler
 
 #include "Control.h"  // for Control
@@ -18,20 +19,16 @@ class XojPageView;
 
 UndoRedoController::UndoRedoController(Control* control): control(control) {}
 
-UndoRedoController::~UndoRedoController() {
-    this->control = nullptr;
-    this->layer = nullptr;
-    elements.clear();
-}
+UndoRedoController::~UndoRedoController() = default;
 
 void UndoRedoController::before() {
     EditSelection* selection = control->getWindow()->getXournal()->getSelection();
     if (selection != nullptr) {
         layer = selection->getSourceLayer();
-        std::copy(selection->getElements().begin(), selection->getElements().end(), std::back_inserter(elements));
+        elements = selection->getElements();
     }
 
-    control->clearSelection();
+    control->clearSelectionEndText();
 }
 
 void UndoRedoController::after() {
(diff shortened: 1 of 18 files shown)


src/core/model/Layer.cpp
```diff
@@ -97,7 +98,39 @@ auto Layer::removeElement(Element* e, bool free) -> Element::Index {
...
+                Stacktrace::printStracktrace();
+                continue;
+            }
+        }
+        if (free) {
+            delete e;
+        }
+        elements[static_cast<size_t>(pos)] = nullptr;
+    }
+    this->elements.erase(std::remove(this->elements.begin(), this->elements.end(), nullptr), this->elements.end());
```
Febbe, line 130: I am not sure, but deleting the pointer from a remove_if predicate should be legal, since you don't modify the pointer itself. This would allow removing the elements in O(n), instead of 2O(n)

Febbe: Yes, you are right, somehow I got confused and thought you are iterating both times through this->elements

src/core/control/UndoRedoController.cpp
```diff
@@ -45,35 +42,39 @@ void UndoRedoController::after() {
     Document* doc = control->getDocument();
 
     PageRef page = control->getCurrentPage();
+    doc->lock();
```
Febbe, line 45: Can you introduce a scoped or unique lock here?

src/core/control/tools/Selection.cpp
```diff
@@ -15,22 +16,25 @@ Selection::Selection(bool multiLayer): multiLayer(multiLayer), viewPool(std::mak
...
 Selection::~Selection() = default;
 
-auto Selection::finalize(PageRef page) -> size_t {
+auto Selection::finalize(PageRef page, bool disableMultilayer, Document* doc) -> size_t {
     this->page = page;
     size_t layerId = 0;
 
-    if (multiLayer) {
-        for (int layerNo = page->getLayers()->size() - 1; layerNo >= 0; layerNo--) {
+    doc->lock();
```
Febbe, line 23: use a smart lock here pls

src/core/gui/PageView.cpp
```diff
@@ -693,26 +696,27 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
...
+                // Aggregate selection
+                sel = SelectionFactory::addElementsFromActiveLayer(control, xournal->getSelection(),
+                                                                   selection->releaseElements());
+            } else {
+                // if selection->multiLayer == true, the selected objects might be on another layer
+                xournal->getControl()->getLayerController()->switchToLay(layerOfFinalizedSel);
+                sel = SelectionFactory::createFromElementsOnActiveLayer(control, page, this,
+                                                                        selection->releaseElements());
+            }
+            xournal->setSelection(sel.release());
```
Febbe, line 714:
```suggestion
            xournal->setSelection( [&](){
                if (aggregate) {
                    // Aggregate selection
                    return SelectionFactory::addElementsFromActiveLayer(control, xournal->getSelection(),
                                                                       selection->releaseElements());
                }
                // if selection->multiLayer == true, the selected objects might be on another layer
                xournal->getControl()->getLayerController()->switchToLay(layerOfFinalizedSel);
                return SelectionFactory::createFromElementsOnActiveLayer(control, page, this,
                                                                        selection->releaseElements());
            }().release());
```

Febbe:
Currently, I've only done a code review, when reviewing the behavior, after which changes should I look and which change in behavior is desired.

Febbe:
> Thanks for the review @Febbe! I made the required changes except the two I commented about.
> 
> About the visible expected changes in behaviour:
> 
>     * Fixed that copied and pasted selection were not rotated properly
> 
>     * Disabled multilayer selection if `Shift` is down: we cannot select elements from two different layers. The arch is not made for that (UndoRedo issues, elements positions).
> 
> 
> I think that's it. Among the things that deserve some attention in testing, is whether the elements get the right z-index (when selected, when aggregated into a preexisting selection, when pasting from clipboard, when dropped on a layer).

Ok, I tested that and it works. 
Am I assuming right, that you removed the "feature" to select over multiple layers, because the architecture does not support it?

Tried to do more selection tests and I didn't find a weird behav.
LGTM

PR #5322, opened 2023-11-06
Fix PROJECT_HOMEPAGE_URL

rolandlo:
Instead of
`<title>Xournal++ document - see https://xournalpp.github.io/</title>` we had
`<title>Xournal++ document - see </title>`
written into all xopp-files since 
#2570 (cmake scripts overhaul)


rolandlo:

PR #5348, opened 2023-11-18
Ensure repainted region is non-empty in PdfElemeSelection

### LINKED ISSUE #5343: Assertion failure when selecting text from PDF background
### Operating System

Linux

### (Linux only) Distribution

Ubuntu

GNOME

Wayland

### Installation Method

PPA / built from source

### Xournal++ Version

1.2.2+dev

### libgtk Version

3.24.38

### Bug Description

When selecting text with the PDF text element selection tool, Xournal++ crashes with a assertion error:
Assertion failure

```term
(com.github.xournalpp.xournalpp:92889): xopp-CRITICAL **: 16:13:56.199: Assertion failed: rv0 == static_cast<Float>(rv1)
    in function floor_cast
    at line 99 of /home/roland/dev/xournalpp/src/util/include/util/safe_casts.h
```
The issue is that 
- `PdfElemSelection::getRegionBbox()` may return an empty range, which has
```c++
    double minX = std::numeric_limits<double>::max();
    double minY = std::numeric_limits<double>::max();
    double maxX = std::numeric_limits<double>::lowest();
    double maxY = std::numeric_limits<double>::lowest();
```
- that range is passed to `XojPageView::flagDirtyRegion` and `XojPageView::repaintArea`
- the latter multiplies these number with `zoom`, so they become `inf` when `zoom > 1` , and then applies `floor_cast<int>` to the product
- In `floor_cast`:
```c++
    auto rv0 = std::floor(f);
    auto rv1 = Integral(rv0);
    xoj_assert(rv0 == static_cast<Float>(rv1));
```
`rv0` is `inf`, but `rv1` is `-2147483648` and the assumption fails, because the latter is not `inf` when casted back to a `double`

In other words `floor_cast` fails when `inf` is passed to it (which kind of makes sense; at least more sense than the result `-2147483648`). 

@Febbe @bhennion How should we fix that? It's very possible that the same issue appears in different places of the code, too.

### Expected Behaviour

No crash.

### Steps to Reproduce

Open a PDF and select text with the PDF text selection tool.

### Additional Context

Somehow the PPA is still compiled in debug mode...

bhennion:
Fix #5343 

This is a quick fix. Adding asserts in Range's member functions will come in a separate PR.

### DIFF
--- a/src/core/control/tools/PdfElemSelection.cpp
+++ b/src/core/control/tools/PdfElemSelection.cpp
@@ -47,7 +47,9 @@ auto PdfElemSelection::finalizeSelectionAndRepaint(XojPdfPageSelectionStyle styl
     Range rg = getRegionBbox();
     bool result = this->finalizeSelection(style);
     rg = rg.unite(getRegionBbox());
-    this->viewPool->dispatch(xoj::view::PdfElementSelectionView::FLAG_DIRTY_REGION_REQUEST, rg);
+    if (!rg.empty()) {
+        this->viewPool->dispatch(xoj::view::PdfElementSelectionView::FLAG_DIRTY_REGION_REQUEST, rg);
+    }
     return result;
 }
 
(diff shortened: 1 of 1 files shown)


rolandlo (approved):

rolandlo:
Consider rebasing this fix on release-1.2. Even though there are no (direct) crashes in release-1.2 related to this issue, it's still potentially problematic behaviour already existent there.

Issue #2249, opened 2020-09-17
Autosave not active when opening .pdf


mfauvain:
Not really a bug, but when opening a .pdf the autosave function does not do it's job. It would be good to 'autosave' the pdf to a '.pdffilename.autosave.xopp' as it does for a regular .xopp file so it could be recovered in case of crash.

jghauser:
Do I understand this correctly? When opening a pdf (that has an attached xopp file) in xournalpp, then it's currently expected that no autosave xopp files will ever be created? If so, then I think I've just come across this now. It would really be great if autosave worked also when opening pdfs. Just lost many many pages of edits because of this...

Thanks for the great software, I know this is a volunteer project! I'm just a bit cranky about having to read 30 pages of Heidegger again ;D

jghauser:
Woop! I thought that because I had already saved the xopp file, it should be in the same folder as that xopp file (as mentioned in xournalpp's preferences). I now went carefully over all the files in the autosave folder, found one with the right date, and indeed it was the autosave file! Thanks!

I guess what would be great is if xournalpp would check on opening whether there is an autosave file that is more recent than the xopp file associated with the pdf, and then ask whether to restore it.

Also, the instructions in the preferences are somewhat wrong (not only about the location of the file). It speaks of a xoj file (but it's xopp) and gives the location as .xournalpp (while this has now moved into .config/xournalpp).

In any case, thanks a lot! This saves me a bunch of time!

Issue #1153, opened 2019-05-15
Periodic autosave cleanup


Technius:
**Is your feature request related to a problem? Please describe.**
After working on a large number of documents over a long span of time, the autosave folder becomes cluttered with old, unneeded autosaves. Users will not be aware of this clutter until they check the folder manually.

**Describe the solution you'd like**
Autosaves should be periodically cleaned up automatically, perhaps at intervals set by the user. Perhaps it can have settings such as "Delete autosaves older than `<length of time>`".

This feature should be enabled by default so that users won't encounter an issue with clutter, but it should be configurable so that users can have the option of always having autosaves available to load.

**Describe alternatives you've considered**
* Manually run autosave cleanups (e.g. by providing a button in the settings). This should bring up a dialog that has a list of all of the autosaves, each with a checkbox indicating whether to keep or delete it.
* Periodically check the size of the autosave folder, and if it exceeds a user-configurable threshold, show a message to the user prompting them to clean up the autosave folder.

**Additional context**
Note that the alternatives are not mutually exclusive; it may be helpful to implement all of them.

See #1142 for how this feature request originated.

LittleHuba:
Timestamps is definitely the best solution. I don't see myself cleaning the autosave folder by hand when it is too big...

PR #5361, opened 2023-11-24
Add error checks for saving operations

### LINKED ISSUE #4934: Corrupted autosave files
image

I had a power outage and the autosave/backup file isn't opening.

Windows 10, version: 1.1.3 +dev, gtk version: 3.24.38,

bhennion:
Fix #4934

This is targetted at the release branch. It adds simply error code checks so that a failure in writing an autosave file will abort and not lead to a corrupted file replacing a previously valid file.

I also removed XmlStrokeNode.cpp/.h which was unused...


rolandlo (changes requested):
While the code improves the situation (in particular when there are filesystem errors or compression errors), I don't think that it helps when there is a power outage during autosaving.

Currently the system works as follows:
- Unsaved documents get autosaved in the cache folder under `<pid>.autosave.xopp`. From the second autosave on, it is renamed to `<pid>.old.autosave.xopp` and a new `<pid>.autosave.xopp` file is created immediately. When Xournal++ is quit (normally) the `<pid>.autosave.xopp` file is deleted and the `<pid>.old.autosave.xopp` file is kept.
- Saved documents get autosaved in the same folder as the document as `.<docname>.autosave.xopp`. From the second autosave on, it is renamed to `<pid>.old.autosave.xopp` in the cache folder a a new `.<docname>.autosave.xopp` is generated immediately in the folder of the document. When Xournal++ is quit normally, the autosave file in the folder of the document is deleted.

So say we have a power outage during autosaving (which means that likely no signal is sent to Xournal++, it's just killed unorderly; sending the `SIGTERM` signal to the app via `killall xournalpp` is simliar; both don't allow creating of an emergencysave file). Then the `<pid>.old.autosave.xopp` file in the cache directory will survive, but the new one (either in `<pid>.autosave.xopp` or `<docname.autosave.xopp`) may be corrupt. 

Since we didn't tell users to look for the `<pid>.old.autosave.xopp` file (or at least I didn't) they may have unnecessarily lost data.

It seems to me that there is a bug in 
https://github.com/xournalpp/xournalpp/blob/905218a928cee75e6fefc0a10aefbbccd28c942c/src/core/control/Control.cpp#L231-L238

Here `filename.string()` is an absolute path and will never start with '.'
So the `else` doesn't occur. Instead of being saved as something like `<pid>foo.xopp` it is saved as `<pid>.old.autosave.xopp` (both in the cache folder). This makes it harder to find it and should be fixed!

By the way the `XmlStrokeNode` class was also removed in the release-1.2 branch via 0121de5fa4b358f6f5c4f7029f39c2f6b5557485

rolandlo:
> Ok. So we could change that by:
> 
>     1. Always autosaving in a file with a "new" or a "~" appended
> 
>     2. If this succeeds, rename this file to the supposed autosave name, thus overriding any previous file.
> 
That sounds like a good idea.
> 
> About 1.: I'd rather do it in the same folder than in a cache folder. The cache folder may not be on the same fs, so the renaming operation may last longer than we think and a crash or error during it could still lead to corrupted data.

I'm fine with that.

Currently it's not quite clear to me when autosave files should be deleted. I would say it's enough to have the most recent (functional) autosave file around. But should it ever be deleted? Currently the most recent autosave file is deleted when Xournal++ is closed whereas an older version may still be in the .cache folder. That doesn't make lot of sense, does it?

There are some other issues with autosaving that will need to be addressed in future (but not necessarily in this PR):
- #1153
- [Ask to restore autosave newer than saved file on startup](https://github.com/xournalpp/xournalpp/issues/2249#issuecomment-896787364).

rolandlo (approved):
LGTM

PR #5368, opened 2023-11-27
Relax shape recognizer for rectangles

tmoerschell:
Triggering shape recognition for rectangles is quite difficult currently. In fact, I didn't know it existed until I looked at the code, despite trying. 

This PR fixes this by initially allowing less precise lines to be considered as polygon segments. If we have only a single line, these are then excluded again. 

I acknowledge the risk of triggering false positives. At a first glance, it seems to be doing alright (haven't seen any yet). Anyhow, I will keep this as a draft PR for a few days and test it out. If I do not find any false positive until Wednesday evening, I will convert it to a normal PR.

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -127,7 +127,7 @@ auto ShapeRecognizer::findPolygonal(const Point* pt, int start, int end, int nsi
         i1 = start + (k * (end - start)) / nsides;
         i2 = start + ((k + 1) * (end - start)) / nsides;
         s.calc(pt, i1, i2);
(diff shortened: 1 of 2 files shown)


tmoerschell:
I have seen no wrongly recognized rectangles, with a minimal stroke size as low as 10. It has sometimes happened that a badly drawn rectangle was recognized as a circle, and the recognizer won't recognize rectangles that are too stretched out. I'm not too sure how it'd be possible to solve this yet. But nevertheless, I think this PR is a step in the right direction. 
Will open it up for review now.

rolandlo (approved):
Thanks @tmoerschell for this contribution.
I have used the shape recognizer for rectangles before and had some mixed success. Your solution makes sense to me and seems to improve the recognition considerably.

rolandlo:

PR #5393, opened 2023-12-07
Ensure non-empty repaint ranges in SearchResultView and VerticalToolView

rolandlo:
When the search result is on a not currently visible page, we get an empty repaint range.
As in #5348 the repaint range must be non-empty, otherwise we get an assertion error (and hence a crash in Debug mode).

The repaint range in the vertical tool view was invalid when shifting below or above the visible part of the page. This could easily be reproduced (start with any document, switch to the vertical tool and move the mouse cursor below the visible part of the page). This lead to a `g-fatal-warning`, which is fixed now.


rolandlo:

Issue #5120, opened 2023-09-04
Space on the right of a page is white on launch but gets gray after any mouse event


Photon89:
### Operating System

Linux

### (Linux only) Distribution

Manjaro Linux

### (Linux only) Desktop Environment

Xfce with Compiz

### (Linux Only) Display Server

X.Org

### Installation Method

Official Release

### Xournal++ Version

1.2.1

### libgtk Version

3.24.38

### Bug Description

See screenshot:

![Workspace 1_004](https://github.com/xournalpp/xournalpp/assets/1191480/30e13b56-fdf0-4040-aa30-a485d6c370e6)

When launching Xournal++ by opening a .xopp file which contains annotations for a PDF file, the space right of the page is white. As soon as any mouse event is triggered (a click, a scroll etc) the space becomes gray like it is supposed to be.

### Expected Behaviour

It should be symmetrically gray left and right of the page.

### Steps to Reproduce

1. Annotate a PDF file and save in a .xopp file.
2. Open the .xopp file.
3. Observe the issue.
4. Click a mouse button and observe the issue disappear.

### Additional Context

I think, the issue is present since the 1.2.0 release.

tattsan:
Duplicate of #5102 .  This bug was introduced by  9d4f59c0d560fba8c69ab695b75695d74a1c32ed .

Photon89:
Thanks, I'll subscribe to updates there then!

Issue #5102, opened 2023-08-28
Unerased white background outside the pages


tattsan:
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

### (Linux only) Desktop Environment

No DE

### (Linux Only) Display Server

X.Org

### Installation Method

Built from source

### Xournal++ Version

Master  (since 9d4f59c0d560fba8c69ab695b75695d74a1c32ed )

### libgtk Version

3.24.38

### Bug Description

When I jump to the first page or the last page, white background is sometimes displayed outside pages. This happens since 9d4f59c0d560fba8c69ab695b75695d74a1c32ed .

https://github.com/xournalpp/xournalpp/assets/6355486/9fdea859-822d-449f-b827-b64e29a860a2

### Expected Behaviour

No white background seen.

### Steps to Reproduce

1. Open .xopp file which have more than 10 pages.
2. Use [End] or [Home] key to jump to the last page or the first page.

### Additional Context

If you jump to non-cached page using side-bar click, the bug also sometimes appears.

itshog:
Can confirm the bug, this happens to me as well

rolandlo:
Closing as fixed by #5399.

PR #5399, opened 2023-12-11
Fix wrongly sized loading pages

tmoerschell:
The default "Loading..." pages that are displayed when pages aren't ready (rendered) at the time of painting are not sized correctly. The zoom was compensated for twice. This also led to some background regions not being redrawn properly afterwards.

To see the issue, zoom in or out quickly, especially with "Preferences -> View -> Clear cached pages while scrolling" checked.

Question to maintainers: should I always open an issue first, or is this fine?

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -871,15 +871,9 @@ auto XojPageView::actionDelete() -> bool {
 void XojPageView::drawLoadingPage(cairo_t* cr) {
     static const string txtLoading = _("Loading...");
 
-    double zoom = xournal->getZoom();
-    int dispWidth = getDisplayWidth();
-    int dispHeight = getDisplayHeight();
-
     cairo_set_source_rgb(cr, 1, 1, 1);
-    cairo_rectangle(cr, 0, 0, dispWidth, dispHeight);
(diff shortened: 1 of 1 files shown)


rolandlo:
> The default "Loading..." pages that are displayed when pages aren't ready (rendered) at the time of painting are not sized correctly. The zoom was compensated for twice. This also led to some background regions not being redrawn properly afterwards.
> 
> To see the issue, zoom in or out quickly, especially with "Preferences -> View -> Clear cached pages while scrolling" checked.

Thanks for the fix. I will try it out soon. Is this related to #5120/#5102?

> 
> Question to maintainers: should I always open an issue first, or is this fine?

For small fixes like that we don't always open an issue. There must at least a clear description of the issue (as you did here) though.

tmoerschell:
Yes, it is at least related to #5102. Regions that weren't redrawn because of this didn't disappear with a single mouse event as described in #5120 though. They would only be redrawn after being moved out of the view and back again. 
The redrawing issue only affects zoom levels > 100%.

rolandlo:
Thanks for the valuable fix @tmoerschell ! I merged it right away, since the flaw with double-compensating the zoom existed as described (and as introduced in 9d4f59c0d560fba8c69ab695b75695d74a1c32ed) and the PR fixed those easily reproducible issues.

PR #5423, opened 2023-12-22
[luaAPI] add getPaletteColor function to query the current palette

atticus-sullivan:
Since https://github.com/xournalpp/xournalpp/pull/2379 has been merged, xournalpp uses a color-palette for the colors in the toolbar. It would be nice for functions like `changeToolColor` to use already configured colors from that palette.

This PR adds a function to get an RGB color-value which then can be used in functions like `changeToolColor`.

Since maybe we'd want to extend the api for this function, I chose to use a table, even if currently there is only one argument. Maybe in the future we'd also like to allow querying for names and then an api like `app.getPaletteColor(nil, "name")` (or interpreting the argument depending on the type) would be rather ugly.

Using negative indices result in a wrap around, but I'm not sure whether we should disallow values out of bounds of the palette (currently we rely on the palette implementation to handle such things which currently results in a wrap around and a (maybe misleading since the toolbar is referenced) warning on the console).

Any suggestions?

---

At first I thought about simply adding a flag `use_palette` to `changeToolColor` but then `color` would have to different meanings depending on that flag or we'd need a new flag for the index but then what if the index and color are being set.
And more important a separate `getPaletteColor` function is much more generic and can also be used for other usecases (e.g. when adding strokes).

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -33,6 +33,7 @@
 #include "gui/XournalView.h"
 #include "gui/sidebar/Sidebar.h"
 #include "gui/sidebar/previews/base/SidebarToolbar.h"
+#include "gui/toolbarMenubar/model/ColorPalette.h"  // for Palette
 #include "gui/widgets/XournalWidget.h"
 #include "model/Document.h"
 #include "model/Element.h"
@@ -1188,7 +1189,7 @@ static int applib_getTexts(lua_State* L) {
             lua_setfield(L, -2, "size");  // add size to text
             lua_setfield(L, -2, "font");  // insert font-table to text element
 
-            lua_pushinteger(L, int(uint32_t(t->getColor()) & 0xffffffU));
+            lua_pushinteger(L, as_signed(uint32_t(t->getColor()) & 0xffffffU));
             lua_setfield(L, -2, "color");  // add color to text
 
             lua_pushnumber(L, t->getX());
@@ -1348,7 +1349,7 @@ static int applib_getStrokes(lua_State* L) {
             lua_pushnumber(L, s->getWidth());
             lua_setfield(L, -2, "width");  // add width to stroke
 
-            lua_pushinteger(L, int(uint32_t(s->getColor()) & 0xffffffU));
+            lua_pushinteger(L, as_signed(uint32_t(s->getColor()) & 0xffffffU));
             lua_setfield(L, -2, "color");  // add color to stroke
 
             lua_pushinteger(L, s->getFill());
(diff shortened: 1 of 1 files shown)


rolandlo (changes requested):
Thanks for the contribution. I agree that the API should allow to query for palette colors.

I would prefer to have `app.getColorPalette()` (without any arguments) though that returns the whole color palette as a table. So the API user gets the whole palette at once and doesn't have to interact with the API repeatedly. Also when querying for a palette color it might be essential to know how many colors the palette contains, which would otherwise have to be queried by a function that is additional to `app.getPaletteColor`. If you have `app.getColorPalette` as a table, you can easily deduce its size. Names should be added as well to this table, so I suggest a return value of the form.

```lua
{ 
  [1] = { color = 0xffffff, name = "white" }, 
  [2] = { color = 0x000000, name = "black" },
  [3] = { color = 0xff0000, name = "red" },
  ...
}

**Comment by atticus-sullivan** [comment]
You've definitely got a point here. All error handling (invalid index) is now done in lua and having the names as well might open some opportunities. We just need to think custom palettes might have different/custom names.

File: src/core/plugin/luapi_application.h [hunk 1]
```diff
@@ -1402,6 +1403,43 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
...
+    lua_settop(L, 0);
+
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Settings* settings = plugin->getControl()->getSettings();
+    const Palette& palette = settings->getColorPalette();
+
+
+    lua_newtable(L);  // return table
+    for (size_t i{0}; i < palette.size(); i++) {
+        lua_pushinteger(L, static_cast<lua_Integer>(i) + 1);  // index in return table
```
rolandlo, line 1429: For consistency: 
```suggestion
    for (size_t i = 1; i <= palette.size(); i++) {
        lua_pushinteger(L, as_signed(i));  // index in return table
```

src/core/plugin/luapi_application.h
```diff
@@ -1402,6 +1403,43 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
...
+        lua_newtable(L);                                      // item table
+
+        const auto& col = palette.getColorAt(i);
+        lua_pushstring(L, col.getName().c_str());
+        lua_setfield(L, -2, "name");
+        lua_pushinteger(L, uint32_t(col.getColor()));
```
rolandlo, line 1435: We don't want RGBA, but only RGB
```suggestion
        lua_pushinteger(L, int(uint32_t(col.getColor()) & 0xffffffU));
```

src/core/plugin/luapi_application.h
```diff
@@ -1402,6 +1403,43 @@ static int applib_changeCurrentPageBackground(lua_State* L) {
     return 0;
 }
 
+/**
+ * Query all colors of the current palette.
+ *
+ * Example1: palette = app.getColorPalette()
```
rolandlo, line 1409:
```suggestion
 * Example: palette = app.getColorPalette()
```

rolandlo (changes requested):
Thanks for the update. I have some more requests.

src/core/plugin/luapi_application.h
atticus-sullivan: Oh didn't think about that one, thanks for pointing out.

Wouldn't be `as_signed` be more consistent?

src/core/plugin/luapi_application.h
atticus-sullivan: I think `i < palette.size()` was correct. At `i = palette.size()` the wrap around will already happen.

rolandlo: I missed that you would have to use `palette.getColorAt(i-1)` then.

src/core/plugin/luapi_application.h
rolandlo: Agree, `as_signed` is better. We have the cast to `int` for colors on multiple places in the code file though, so that would have to be changed accordingly.

rolandlo (approved):
Can you please apply clang format, squash the commits into one and rebase on current master?

atticus-sullivan:
:+1: So the `as_signed` for colors will be a different PR?

rolandlo:
> 👍 So the `as_signed` for colors will be a different PR?

How you wish. It's not urgent.

rolandlo:
Ok, merging in 24 hours unless an objection is raised. Thanks @atticus-sullivan for the nice contribution!

PR #5443, opened 2024-01-10
Add undo handler for missing PDF replacement

### LINKED ISSUE #5435: File doesn't ask to save when replacing PDF
### Operating System

Linux

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

deb

### Xournal++ Version

1.2.2

### libgtk Version

3.24.20

### Bug Description

When you move the PDF file or rename it, Xournal++ doesn't find it anymore and offers the "find PDF" dialog. Once you find the right PDF, you can work as normal.
If you only replace the PDF and you don't make any other changes and close the window. There is no dialog asking to save the changes. This results in Xournal++ not finding the PDF again the next time you open the xopp file. 

### Expected Behaviour

When a PDF has been replaced, Xournal++ should ask whether to save the file when you close it even if no other changes have been made to the document.

### Steps to Reproduce

1) Create a xopp from a PDF
2) Save and close
3) rename PDF
4) open xopp
5) follow "find PDF" dialog and use the renamed PDF
6) close the window without any other changes
7) open xopp again

### Additional Context

_No response_

tmoerschell:
Fixes #5435

~~Undoing simply reopens the whole file.~~
Now involves a slight refactoring of how the page background is replaced.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2278,86 +2278,6 @@ auto Control::openFile(fs::path filepath, int scrollToPage, bool forceOpen) -> b
 
     LoadHandler loadHandler;
     Document* loadedDocument = loadHandler.loadDocument(filepath);
-    if ((loadedDocument != nullptr && loadHandler.isAttachedPdfMissing()) ||
-        !loadHandler.getMissingPdfFilename().empty()) {
-        // give the user a second chance to select a new PDF filepath, or to discard the PDF
-        const fs::path missingFilePath = fs::path(loadHandler.getMissingPdfFilename());
-
-        std::string parentFolderPath;
-        std::string filename;
-#if defined(WIN32)
-        parentFolderPath = missingFilePath.parent_path().string();
-        filename = missingFilePath.filename().string();
-#else
-        // since POSIX systems detect the whole Windows path as a filename, this checks whether missingFilePath
-        // contains a Windows path
-        std::regex regex(R"([A-Z]:\\(?:.*\\)*(.*))");
-        std::cmatch matchInfo;
-
-        if (std::regex_match(missingFilePath.filename().string().c_str(), matchInfo, regex) && matchInfo[1].matched) {
-            parentFolderPath = missingFilePath.filename().string();
-            filename = matchInfo[1].str();
-        } else {
-            parentFolderPath = missingFilePath.parent_path().string();
-            filename = missingFilePath.filename().string();
-        }
-#endif
-        std::string msg;
-        if (loadHandler.isAttachedPdfMissing()) {
-            msg = FS(_F("The attached background file could not be found. It might have been moved, "
-                        "renamed or deleted."));
-        } else {
-            msg = FS(_F("The background file \"{1}\" could not be found. It might have been moved, renamed or "
-                        "deleted.\nIt was last seen at: \"{2}\"") %
-                     filename % parentFolderPath);
-        }
-
-        // try to find file in current directory
-        auto proposedPdfFilepath = filepath.parent_path() / filename;
-        bool proposePdfFile = !loadHandler.isAttachedPdfMissing() && !filename.empty() &&
-                              fs::exists(proposedPdfFilepath) && !fs::is_directory(proposedPdfFilepath);
-        if (proposePdfFile) {
-            msg += FS(_F("\nProposed replacement file: \"{1}\"") % proposedPdfFilepath.string());
-        }
-        GtkWidget* dialog = gtk_message_dialog_new(getGtkWindow(), GTK_DIALOG_MODAL, GTK_MESSAGE_QUESTION,
-                                                   GTK_BUTTONS_NONE, "%s", msg.c_str());
-
-        enum dialogOptions { USE_PROPOSED, SELECT_OTHER, REMOVE, CANCEL };
-
-        if (proposePdfFile) {
-            gtk_dialog_add_button(GTK_DIALOG(dialog), _("Use proposed PDF"), USE_PROPOSED);
-        }
-        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Select another PDF"), SELECT_OTHER);
-        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Remove PDF Background"), REMOVE);
-        gtk_dialog_add_button(GTK_DIALOG(dialog), _("Cancel"), CANCEL);
-        gtk_window_set_transient_for(GTK_WINDOW(dialog), GTK_WINDOW(this->getWindow()->getWindow()));
-        int res = gtk_dialog_run(GTK_DIALOG(dialog));
-        gtk_widget_destroy(dialog);
-
-        switch (res) {
-            case USE_PROPOSED:
-                if (!proposedPdfFilepath.empty()) {
-                    loadHandler.setPdfReplacement(proposedPdfFilepath, false);
-                    loadedDocument = loadHandler.loadDocument(filepath);
-                }
-                break;
-            case SELECT_OTHER: {
-                bool attachToDocument = false;
-                XojOpenDlg dlg(getGtkWindow(), this->settings);
-                auto pdfFilename = dlg.showOpenDialog(true, attachToDocument);
-                if (!pdfFilename.empty()) {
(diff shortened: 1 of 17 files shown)


Febbe:
Looks like there is a conceptual problem. When replacing the document, it should automatically clear the undo redo handling. But this is only done when the document is closed.

The attempt, to just reopen the document, might be very clunky / time-consuming.
The best way would be, to unload the pdf and to issue the pdf/asset find dialog. Dunno if that's a feasable solution.

You might want to merge this into release1.2.x instead of master.
Generally the code looks good to me. And as a fix it is acceptable. What do you think @xournalpp/core?

tmoerschell:
> Looks like there is a conceptual problem. When replacing the document, it should automatically clear the undo redo handling. But this is only done when the document is closed.

If you're talking about my manually clearing the undo/redo handler at the beginning of the `MissingPdfUndoAction::undo()`, this was only done to cover a weird corner case consistently: 

Since you could redo this action (without any effect), you could have two missing PDF undo action in the undo list. If you then undo again, you would get asked for validation to discard your existing changes (the second undo).

But I found another way. See next commit. 

> The attempt, to just reopen the document, might be very clunky / time-consuming. The best way would be, to unload the pdf and to issue the pdf/asset find dialog. Dunno if that's a feasable solution.

I did look into this and decided it was not worth it. We would essentially only win time on not having to close the document properly (as done with `Control::close()`), a few extra checks, and `Control::fileLoaded()`. All the rest of the `Control::fileOpen()` function more or less needs to be run anyways. 

> You might want to merge this into release1.2.x instead of master. 

Ok, will do.

Febbe:
> If you're talking about my manually clearing the undo/redo handler at the beginning of the MissingPdfUndoAction::undo()

Yes, and I understand why it is necessary. My critic regarding the concept should not address your code. It should address the current state. In my opinion your way should not be required when the undo/redo handling would be part of the document itself. So that it is cleared implicitly, when the document is exchanged.
Just a note, when someone wants to do some refactoring and has time to do it.

Regarding the build failure: probably a dependency is missing after Microsoft updated their image or something similar. This is not a blocker for your PR.

Febbe:
Ok, I rethought about this and I have questions / problems with this approach:
 - What happens, when 
     1. I do some work after resetting the pdf
     2. Pressing many times undo and accidentally pressing it too much, thus reverting changes I did not want to revert?
     I assume that the document is reloaded, and the redo list is cleared. 
     That will mean, I cannot redo them.
     If that's the case, It would be a blocker.
     I also don't like the way you chose, to fill an empty redo element into the queue, which violates the invariant, that all elements in the queues are not null.

tmoerschell:
You're right, I didn't think about this. This means we cannot trigger the dialog directly on undo. 

There might be a more elegant way to go about this, too: 

Add an undo action only if the user didn't click on "Cancel". 
Undo reverts changes to the background: result is as if the user had clicked on "Cancel". 
Redo will actually redo the changes. I see no way to reopen the dialog though, so we would need an option such as #5037 in the menu. 

Also we need to separate the dialog code from the `fileOpen()` code: open the file first, ask to replace the background afterwards. 

Does that sound better ?
I wil mark this as a Work In Progress for now.

Febbe:
> Add an undo action only if the user didn't click on "Cancel". Undo reverts changes to the background: result is as if the user had clicked on "Cancel". Redo will actually redo the changes. I see no way to reopen the dialog though, so we would need an option such as #5037 in the menu.

Yes, that would be the desired way. Maybe we want to move the loaded PDF into the UndoHandler on Undo and reset it on redo.
That will lead to a defered deletion, when a new action is pushed into the queue.

> 
> Also we need to separate the dialog code from the `fileOpen()` code: open the file first, ask to replace the background afterwards.

Yes.

tmoerschell:
Well this resulted in considerably more changes than expected, but it works pretty well. I will clean up all this tomorrow.

tmoerschell:
All done :-) I only didn't implement the storing of the PDF document in the undo handler. The PDF file is reread on redo. I hope that's ok

tmoerschell:
Just did some minor cleanup :-) Could I request a review on this ?

Febbe:
Maybe I have time at the weekend. Currently we are all very busy...

src/core/undo/MissingPdfUndoAction.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+MissingPdfUndoAction::~MissingPdfUndoAction() = default;
+
+auto MissingPdfUndoAction::undo(Control* control) -> bool {
+    Document* doc = control->getDocument();
+
+    fs::path redoFilepath = doc->getPdfFilepath();
+
+    doc->resetPdf();
+    doc->setPdfAttributes(filepath, doc->isAttachPdf());
```
Febbe, line 24: May it happen, that the value of `doc->isAttachPdf()` may change between those two states?

src/core/control/Control.cpp
```diff
@@ -2464,6 +2390,85 @@ void Control::fileLoaded(int scrollToPage) {
     updateDeletePageButton();
 }
 
+void Control::promptMissingPdf(LoadHandler& loadHandler, const fs::path& filepath) {
```
Febbe, line 2393: I have 2 issues here, which are fine for 1.2, as bugfix (nitpick). 
First, the function does more than just prompting for a missing pdf.
Second, the current way is not compatible with gtk4, which is relevant for master.

So, when you have time, you could refactor this after this PR is merged, to accept a handler function, which is called after the dialog has been closed.

It would look then like this:
```suggestion
void Control::promptMissingPdf(LoadHandler& loadHandler, const fs::path& filepath, std::function<void(std::filesystem::path)> handler) {
```

Febbe (approved):

src/core/undo/MissingPdfUndoAction.cpp
tmoerschell: You're right, the user may have changed it when he chose a new PDF background. I will amend the last commit to account for that.

src/core/control/Control.cpp
tmoerschell: The function name will have to change anyways, so I might as well include it here. I propose `attemptReplaceMissingPdf()`, except if you have something better in mind.

tmoerschell: Actually I'm not sure I understand how this should work with GTK4. We only create and show the dialog, and even if it is modal, GTK will immediately return and call the provided handler upon action from the user?
In this case, why should the handler function already be passed as a parameter to `promptMissingPdf()`? The latter should be able to access the handler function through the class anyways, right?

Would you prefer to omit the name change since it is fitting for GTK4?

Febbe: Yes, just leave the name. 
Gtk3 deprecated old modal dialogs. And Gtk4 removed them. Now you must register a handler function/closure to each dialog, which is called by the event loop, when the dialog closes. 

With c++20 we could just use coroutines and leave the original function structure as is.

tmoerschell: I see that GTK 4.10 deprecated `GtkMessageDialog` and introduced `GtkAlertDialog`. Can I use `GtkAlertDialog`, or do you want to not have a dependency on 4.10 (at first)?

Febbe: I think we are talking past each other. For now this is fine, since this PR is meant to be merged into 1.2.

On master this would still work, but since we want to make the transition to gtk4. We will have to refactor it. When you have the time and this PR is merged into 1.2, you could do this, since you already touched that code lately.
So for now, just squashing your work is fine here.

Regarding your future work; we first need to remove everything, which got deprecated before GTK 3.24 and which is removed in gtk4.0. This means, we are still on 3.24 and you can't use a dependency from 4.10. GtkAllertDialog won't be available.

tmoerschell: Ok, good I understand now. Thank you! 
You I squash everything or should I keep the refactoring and the undo handler addition separate?

Febbe: Squashing both together is fine for me.

tmoerschell:
Ok, it's all good from my side!

Febbe:
Thank you for your work. This bug fix will be part of the next sub release.

PR #5447, opened 2024-01-11
Fix windows build because of missing libpng

Febbe:
Somehow, mingw-w64-x86_64-libpng is not a required dependency of ImageMagick. But we need it.


Febbe:
Closing, since it is a fixed bug in MSYS2/MINGW-packages: https://github.com/msys2/MINGW-packages/pull/19707

PR #5469, opened 2024-01-21
SRS Document

vasrask:
This is a Software Requirements Specification document I attempted for v1.2.2. The document aims to outline the software requirements and specifications based on my understanding. I would appreciate any feedback, suggestions, or improvements from the community.


rolandlo:
That looks pretty good, pretty professional. :-) Thanks for putting all the work into it. I haven't read the whole document yet, but I certainly will once I find time. What I have read so far sounded pretty accurate.

vasrask:
Thank you very much! I would appreciate any comments or observations, especially concerning parts that I may not have understood correctly or points where I may have missed important information.

rolandlo:
Finally I found the time to read the whole document. Here is some feedback:

2.1: The SourceForge repo is just an unofficial mirror, not sure it should be mentioned. There are other sources like package repos of various Linux distros (see https://pkgs.org/search/?q=xournalpp). On Windows there is chocolatey https://community.chocolatey.org/packages/xournalplusplus and there are others who provide the binaries. 

2.5: It's GPL 2 or later

2.7: Xournal++ uses GTK, GLib, Poppler and Cairo (for rendering) as main dependencies.

3.1: 
- It's not clear to me what you are referring to with "forms" and "windows". Maybe you mean dialogs. Also "scrollbar" should be "popup" on various places, I think.
- You describe the Portrait toolbar (which is the default one), but at least it should be mentioned that there are other toolbars (see `Edit > Toolbars`) and custom toolbars can be created as well.
- You describe different sections of the same toolbar. I would write e.g. "Edit section" instead of "Edit toolbar"
- "New Xournal" -> "New Journal"
- When you describe the sidebar (note that it can also be on the right if you choose that preference) you should mention that for PDF's there is also a contents section listing all chapters and section of the document. 

4.1.2: "saved in memory" -> "saved on disk"

4.2.1: "Xouranl++" -> "Xournal++"

4.2.2: "at a new window" -> "in a new window"

4.4.1: I think you mean `%{name}.bg.pdf`

4.7.2 the range of pages can also be chosen for exporting to PDF. Moreover there are options of which background types should be exported.

4.8.2: I'm surprised to read that Ruled and Graph backgrounds do not get printed. They should be and at least on Linux they are in my experience (haven't tried on other OSes)

4.10.1: The Arrange options apply to the whole selection (which may contain multiple objects).

4.10.3: The objects must share the same page layer

4.14.1: "rendered" -> "created"

4.17.2: Snapping is not only applied while working with a selection, but also while drawing

4.18 
- For Autoloading the .xopp-file must share the same basename with the PDF so that it gets loaded
- Usually there are various devices (not just "System Aggregated Pointer") that can be remapped to a different device class. Mention that this should be used if the device class was not correctly recognized automatically.

4.27.2 When deleting a page the next page becomes the current one (unless the last page was deleted), I think

4.32.1: The Whiteout eraser actually paints a white stroke to cover the content.

4.36.1: There are more images formats supported (in particular `.webp` is missing), so I would write "include" rather than "are" (and add `.webp` to the list)

4.36.2 and 4.39.2: "copy" -> "duplicate"

4.42.3: Note that the vertical space tool moves elements above the cursor when "CTRL" is pressed

4.46.2 "in memory" -> "on disk"

4.49.1: The MigrateFontSizes plugin only uses the Lua lgi-module for displaying a dialog. It can be used to scale text even if the lgi-module is missing

4.49.3: All plugins require a Lua installations, but some plugins might require additional Lua modules installed

Appendix A (Glossary) for Layers: "within the journal" -> "within the same page"


If the document should be merged into the repository, the source of the document (not just the PDF) should be included, so that we don't have to start from scratch when some changes must be applied or a different version should be described.
It looks like the source is a MS Word-document. In that case an open text format would be preferable.

vasrask:
Thanks a lot for the feedback, I made most of the changes but I have some comments/questions:

2.7: Cairo went totally over my head. Where was I supposed to find it?

3.1: I replaced "scrollbar" with "dropdown menu"; I think that is the term I needed here.

4.7.2: I couldn't find this functionality, could you give me more details?

4.8.2: I was using Windows. I didn't manage to print those, but maybe it was something in my settings. I will check again.

4.17.2: I added "or draws" but couldn't really see how, in order to explain it.

4.18: Do you mean I shouldn't mention System Aggregated Pointer at all, and just state that these should be changed in case of incorrect device class recognition? I just now noticed that these may vary (on Linux I got Wayland Wheel Scrolling and Wayland Pointer options using the same device(s)).

rolandlo:
Thanks for the update. First let me answer your questions:

> Thanks a lot for the feedback, I made most of the changes but I have some comments/questions:
> 
> 2.7: Cairo went totally over my head. Where was I supposed to find it?

Looks like we should be more transparent about it. The GLib bindings of Poppler that we use require Cairo for rendering. If you look at the code base of Xournal++, Cairo is used in 156 files, Poppler in 7 files.

> 3.1: I replaced "scrollbar" with "dropdown menu"; I think that is the term I needed here.

Ok

> 
> 4.7.2: I couldn't find this functionality, could you give me more details?
There are different ways:
1) Use the command line. Here's the output of `xournalpp --help-export`:
grafik
2) Use menu `File > Export As...`
grafik

> 
grafik

> 4.8.2: I was using Windows. I didn't manage to print those, but maybe it was something in my settings. I will check again.
> 
> 4.17.2: I added "or draws" but couldn't really see how, in order to explain it.
> 
[SnappingDemo.webm](https://github.com/xournalpp/xournalpp/assets/40485433/2a97c02b-c76d-40ad-b48f-7f9c8c456a25)

> 4.18: Do you mean I shouldn't mention System Aggregated Pointer at all, and just state that these should be changed in case of incorrect device class recognition?

Yes, I think so.

> I just now noticed that these may vary (on Linux I got Wayland Wheel Scrolling and Wayland Pointer options using the same device(s)).

vasrask:
Hello, so sorry for the huge delay! I made the necessary changes to the document and included the ODT.

rolandlo:
Thanks for the update. The numbering was lost in the docx->odt conversion. Please fix it. Also some hyperlinks are displayed wrongly (like 4.26 in the table of contents). Please check carefully if there are other conversion errors.
In the end the docx should be removed and the PDF should be generated from the odt file.

rolandlo:
Actually when I save the doc file as odt in Libre Office the numbering and links look fine. In case you did it with Microsoft Word that may explain the result. 
I think you could install Libre Office on Windows. Otherwise let me know and I will add the correctly converted document.

vasrask:
Hello, I pushed a corrected .odt created the way you suggested.

PR #5481, opened 2024-02-01
Add undo handler for missing PDF replacement - master

tmoerschell:
Same as #5443, but solved conflicts and made the dialog ready for GTK4. 

One problem might be that this is that this would make the release and master branch diverge. I'm not sure how I could fast forward the commits from the release branch onto master. If I do a merge commit, the original commits are lost. Could someone help me with this?


tmoerschell:
Ok so I do the merge PR and then I put the GTK4-related changes into an additional commit here?

tmoerschell:
Didn't work out as I thought it would. Rebased it now, so it's all clean

tmoerschell:
It's done :-) It's also much cleaner with `std::bind()`. Thank you !

Issue #5474, opened 2024-01-27
With 1/22/2024 development update = ARMS Macs can't even run Xournal++ with Rosetta anymore!!!


lcalder:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

Official Release

### Xournal++ Version

1.2.2+dev (Jan 22 2024)

### libgtk Version

3.24.38

### Bug Description

Xournal++ now crashes immediately upon launch. I guess it now requires the lua dylib to be x86_64 architecture as well. All previous Mac versions of Xournal++ up until this latest would just work after installation.

I was able to get it to run after installing lua with x86_64 architecture on my M2 Mac, but even now, periodically a message comes up "Filesystem error: {2}" --- maybe it is now accessing the filesystem in a different way when performing autosaves??? (my guess only)

Since Sonoma 14.2 MacOS version, any UI dialog boxes that open up will hang, often indefinitely, (this includes filebrowser dialogs, transparency, font lists, as well as this new Filesystem error) and so you just have to force quit Xournal++... it means the software has become almost unusable on ARM Macs.  Are there any plans to issue an ARM Version? I've tried to compile several times, but each time I tried I gave up after a few days of trying because it was just one problem after another... Is there anyone willing to help me with this? I'd sure like to try again to see if this fixes all these problems with this irreplaceable program.

### Expected Behaviour

Should just install as it used to and all dialogs should respond to user input instead of giving error messages and hanging as it used to.

### Steps to Reproduce

Install on a Mac ARM with Sonoma 14.2 or higher

rolandlo:
> Xournal++ now crashes immediately upon launch. I guess it now requires the lua dylib to be x86_64 architecture as well. All previous Mac versions of Xournal++ up until this latest would just work after installation.

I have no idea what has changed with respect to the Lua dylib. What was the latest nightly that didn't have this problem?
 
> I was able to get it to run after installing lua with x86_64 architecture on my M2 Mac, but even now, periodically a message comes up "Filesystem error: {2}" --- maybe it is now accessing the filesystem in a different way when performing autosaves??? (my guess only)

[...]

PR #5485, opened 2024-02-02
Avoid using fs::equivalent on empty path

rolandlo:
Fixes two issues
- Filesystem error on MacOS when a document is autosaved the first time after startup
- Typo in error message displaying `{2}` instead of error message

The first issue was described in https://github.com/xournalpp/xournalpp/issues/5474#issue-2103370589


rolandlo:
Thanks for the quick review @bhennion

PR #5490, opened 2024-02-06
Fix *.pdf filter for ExportAsPdf dialog

bhennion:
Fix *.pdf filter for ExportAsPdf dialog.

Note: we could also replace the pattern *.pdf by the MIME type application/pdf (so that the filter would match a pdf file without extension for instance)


rolandlo (approved):
LGTM as is.

In case you prefer to replace the pattern *.pdf by the MIME type application/pdf this would have to be done in `CustomExportJob` constructor for all 4 MIME types:
- application/pdf
- application/png
- application/svg
- application/x-xoj

PR #5500, opened 2024-02-11
More towards gtk4

bhennion:
This PR replaces the uses of some gtk3 functions which are retired in gtk4.
It also makes the export dialogs non-blocking (again, for gtk4).

There are two noticable changes in behaviour in the export dialogs:

1. The file filters are now using the MIME type instead of an extension pattern. This needs testing on various platforms.
2. gtk4 has no equivalent of `gtk_file_chooser_set_local_only()`, which removes distant filesystems from the dialog. I removed the use of this function entirely but it may create issues on some platforms and some kinds of distant fs.

Both those changes require some amount of testing (and could be reversed if need be).

### DIFF
--- a/mac-setup/Info.plist
+++ b/mac-setup/Info.plist
@@ -73,6 +73,108 @@
         <key>CFBundleTypeRole</key>
         <string>Editor</string>
       </dict>
+      <dict>
+        <key>CFBundleTypeExtensions</key>
+        <array>
+          <string>xopt</string>
+        </array>
+        <key>CFBundleTypeIconFile</key>
+	        <string>xournalpp.icns</string>
+        <key>CFBundleTypeName</key>
+        <string>Xournal++ Template</string>
+        <key>CFBundleTypeMIMETypes</key>
+        <array>
+          <string>application/x-xopt</string>
+        </array>
+        <key>CFBundleTypeRole</key>
+        <string>Editor</string>
+      </dict>
+      <dict>
+        <key>CFBundleTypeExtensions</key>
+        <array>
+          <string>xoj</string>
+        </array>
+        <key>CFBundleTypeIconFile</key>
+	        <string>xournalpp.icns</string>
+        <key>CFBundleTypeName</key>
+        <string>Xournal Document</string>
+        <key>CFBundleTypeMIMETypes</key>
+        <array>
+          <string>application/x-xojpp</string>
+        </array>
+        <key>CFBundleTypeRole</key>
+        <string>Editor</string>
+      </dict>
+    </array>
+
+    <key>UTExportedTypeDeclarations</key>
+    <array>
+      <dict>
+        <key>UTTypeIdentifier</key>
+        <string>com.github.xournalpp.xopp</string>
+        <key>UTTypeConformsTo</key>
+        <array>
+          <string>public.data</string>
+          <string>org.gnu.gnu-zip-archive</string>
+        </array>
+        <key>UTTypeDescription</key>
+        <string>Xournal++ Document</string>
+        <key>UTTypeTagSpecification</key>
+        <dict>
+          <key>public.mime-type</key>
+          <array>
+            <string>application/x-xopp</string>
+          </array>
+          <key>public.filename-extension</key>
+          <array>
+            <string>xopp</string>
+          </array>
+        </dict>
+      </dict>
+      <dict>
+        <key>UTTypeIdentifier</key>
+        <string>com.github.xournalpp.xopt</string>
+        <key>UTTypeConformsTo</key>
+        <array>
+          <string>public.data</string>
+          <string>public.text</string>
+        </array>
+        <key>UTTypeDescription</key>
+        <string>Xournal++ Template</string>
+        <key>UTTypeTagSpecification</key>
+        <dict>
+          <key>public.mime-type</key>
+          <array>
+            <string>application/x-xopt</string>
+          </array>
+          <key>public.filename-extension</key>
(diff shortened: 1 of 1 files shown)


src/core/control/Control.cpp
```diff
@@ -1737,8 +1733,9 @@ auto Control::showSaveDialog() -> bool {
         }
     }
 
-    auto filename = Util::fromGFilename(gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(dialog)));
-    settings->setLastSavePath(filename.parent_path());
+    auto file = Util::fromGFile(
+            xoj::util::GObjectSPtr<GFile>(gtk_file_chooser_get_file(GTK_FILE_CHOOSER(dialog)), xoj::util::adopt).get());
```
rolandlo, line 1737: A few lines down it should be `this->doc->setFilepath(file);` I think.

rolandlo:
The png and svg MIME types don't work for me on Linux (Ubuntu 23.10) when using menu `File > Export as...`. Those files do not show up in the file dialog when I choose these MIME types.

src/core/control/jobs/CustomExportJob.cpp
```diff
@@ -25,29 +26,28 @@
 
 CustomExportJob::CustomExportJob(Control* control): BaseExportJob(control, _("Custom Export")) {
     // Supported filters
-    filters.insert({_("PDF files"), ExportType(".pdf")});
-    filters.insert({_("PNG graphics"), ExportType(".png")});
-    filters.insert({_("SVG graphics"), ExportType(".svg")});
-    filters.insert({_("Xournal (Compatibility)"), ExportType(".xoj")});
+    filters.insert({_("PDF files"), ExportType(".pdf", "application/pdf")});
+    filters.insert({_("PNG graphics"), ExportType(".png", "application/png")});
+    filters.insert({_("SVG graphics"), ExportType(".svg", "application/svg")});
```
rolandlo, line 31: This should be:
```suggestion
    filters.insert({_("PNG graphics"), ExportType(".png", "image/png")});
    filters.insert({_("SVG graphics"), ExportType(".svg", "image/svg+xml")});
```

Compare https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

rolandlo:
Yes, I have time to test this on MacOS and Windows as well.

For the record: I got two new warnings on MacOS (seemingly unrelated to this PR):

```term
[87/446] Building CXX object src/util/CMakeFiles/util.dir/GtkUtil.cpp.o
/Users/xournal-dev/xournalpp/src/util/GtkUtil.cpp:104:31: warning: unused variable 'tgt' [-Wunused-variable]
                    GVariant* tgt = gtk_actionable_get_action_target_value(GTK_ACTIONABLE(btn));
                              ^
```
and
```term
[350/446] Building CXX object src/core/CMakeFiles/xournalpp-core.dir/model/Stroke.cpp.o
/Users/xournal-dev/xournalpp/src/core/model/Stroke.cpp:724:26: warning: lambda capture 'outerBox' is not used [-Wunused-lambda-capture]
            [&innerBox, &outerBox](const Point& lastKnot, const Point& halfTangentControlPoint) -> bool {
                      ~~~^~~~~~~~
1 warning generated.
```

rolandlo:
On MacOS I can see the pdf, svg and png files. I can't see the xoj files though (unlike on Linux).

rolandlo:
On Windows I can see all the files (pdf, svg, png, xoj) in the export file chooser.

When accessing C:\ I get some weird warnings, but they are also there in the 2023, Nov 16 version of 1.2.2+dev, so they are unrelated to your PR.

````term
(com.github.xournalpp.xournalpp:16560): GLib-GIO-CRITICAL **: 10:16:35.680: GFileInfo created without standard::type

(com.github.xournalpp.xournalpp:16560): GLib-GIO-CRITICAL **: 10:16:35.680: file ../glib-2.78.1/gio/gfileinfo.c: line 1611 (g_file_info_get_file_type): should not be reached
````

**Comment by rolandlo** [comment]
The reason why on MacOS the xoj MIME filter is not recognized lies probably in `Info.plist`, where only `application/x-xopp` is registered. I will test a fix.

**Comment by rolandlo** [comment]
> The reason why on MacOS the xoj MIME filter is not recognized lies probably in `Info.plist`, where only `application/x-xopp` is registered. I will test a fix.

It turned out that this is not true. Even if .xoj files are registered in the Info.plist (and therefore are opened with Xournal++ when trying to open them from Finder or via the `open` command) they don't show up.

I can see the difference between Linux/Windows and MacOS using [Gio.conent_type_guess](https://docs.gtk.org/gio/func.content_type_guess.html)

The following Python code returns `

```python
from gi.repository import Gio
Gio.content_type_guess(filename="sample.xoj", data=None)
```
('application/x-xojpp', result_uncertain=True) on Linux and
('dyn-age81u55', result_uncertain=True) on MacOS

It's similar for xopp-files ('dyn-age81u55usa` on MacOS), whereas standard file formats like pdf, svg or png are recognized correctly.

I looked at the code in [`gosxcontenttype.m`](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gosxcontenttype.m?ref_type=heads#L538-559) and it looks like on MacOS even `.ui` and `.txt` files had to be hard-coded in order to be recognized.

So at least on MacOS we should probably not use the MIME type for `.xoj` files.

rolandlo (approved):
Code looks good to me as well.

PR #5426, opened 2023-12-29
Allow for multiple instances of tools in the toolbars

bhennion:
Allow for multiple instances of tools in the toolbars

It's not quite done yet. It overall works fine, but I still need to:

- [x] Fix the zoom slider item
- [x] Rearrange the toolbar customize menu (e.g. in sections) so that item are easy to find and browse.
- [x] Link related issues to this PR

PR #5510, opened 2024-02-17
Fix toolbar management Segfault and out of date menu - Release branch

bhennion:
See https://github.com/xournalpp/xournalpp/pull/5426#issuecomment-1950248551


rolandlo (approved):
That works fine! Great you fixed that.

PR #5512, opened 2024-02-19
Do not export Pdf outline when only exporting part of the document

### LINKED ISSUE #4897: Unable to export specific pages when editing PDF containing bookmarks

**Affects versions :**
 - OS: [Windows]
 - libgtk : 3.24.34
 - Version of Xournal++: 1.1.3
 - Installation method: official release github

**Describe the bug**
Unable to export specific pages when editing a PDF that contains bookmarks.

**To Reproduce**
1. Have a PDF with bookmark (in my case 3 blank pages, bookmark on page 2)
2. Open the PDF with Xournal++
3. draw some stuff on the pages
4. Export as (Ctrl+E)
5. Select Current page or Custom pages (NOT all pages, this one works)
6. The generated PDF should be corrupted
(Note: Xournal++ seems to never close the PDF file as it is flagged as used by another applications, until the app is closed)

**Expected behavior**
The PDF should open normally

bhennion:
Fix #4897
The issue was: the cairo_pdf_surface was in failed state when cairo_surface_destroy() was calling (internally) cairo_surface_finish().
Indeed, if the outline has links pointing out of the document's bounds (e.g. to a page past the end of the document), then the surface enters a failed state upon cairo_surface_finish() (and not before, as it does not yet know the total number of pages...)

This PR:

1. Avoid exporting the outline when only exporting part of the document (via export as)
2. Calls cairo_surface_finish() by hand and checks for error, so that anything of the sort won't go undetected anymore.

### DIFF
--- a/src/core/pdf/base/XojCairoPdfExport.cpp
+++ b/src/core/pdf/base/XojCairoPdfExport.cpp
@@ -43,22 +43,24 @@ void XojCairoPdfExport::setExportBackground(ExportBackgroundType exportBackgroun
     this->exportBackground = exportBackground;
 }
 
-auto XojCairoPdfExport::startPdf(const fs::path& file) -> bool {
+auto XojCairoPdfExport::startPdf(const fs::path& file, bool exportOutline) -> bool {
     this->surface = cairo_pdf_surface_create(file.u8string().c_str(), 0, 0);
     this->cr = cairo_create(surface);
 
 #if CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 16, 0)
     cairo_pdf_surface_set_metadata(surface, CAIRO_PDF_METADATA_TITLE, doc->getFilepath().filename().u8string().c_str());
     cairo_pdf_surface_set_metadata(surface, CAIRO_PDF_METADATA_CREATOR, PROJECT_STRING);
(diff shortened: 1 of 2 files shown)


src/core/pdf/base/XojCairoPdfExport.cpp
```diff
@@ -173,7 +183,11 @@ auto XojCairoPdfExport::createPdf(fs::path const& file, const PageRangeVector& r
         return false;
     }
 
-    if (!startPdf(file)) {
+    // Only export the outline if we are exporting the entire document. otherwise, links may point to page numbers
+    // greater than the total number of pages, leading to corrupt pdf files
+    bool outline = range.size() == 1 && range.front().first == 0 && range.front().last >= doc->getPageCount() - 1;
```
rolandlo, line 188: This is only a sufficient condition for having all pages exported. If you export say a 3 page PDF with 
```xournalpp input.xopp -p export.pdf --export-range=1,2,3```
you get a different output (one without outline) than when you leave away the export-range.

This looks like a minor issue and maybe it's even wanted, but I'm pointing it out in case it wasn't noticed.

rolandlo (commented):
I have one comment, which you may or may not want to address. Otherwise it looks all good to me.

rolandlo (approved):
Agree. Then please go ahead.

PR #5522, opened 2024-02-25
Update changelog for 1.2.3

rolandlo:
Here is the changelog for version 1.2.3. I suggest to follow up with the 1.2.3 release soon.


rolandlo:
Merging tonight unless objections are raised.

PR #5550, opened 2024-03-09
Make "Replace File?" dialog transient for its parent dialog

### LINKED ISSUE #3253: "do you want to replace" dialog do not stays on top
**Affects versions :**
 - OS: Linux
 - Desktop environment: Gnome-X11
 - Which version : 
```
% xournalpp --version
Xournal++ 1.1.0
└──libgtk: 3.24.20
```
 - from source, compiled to `.deb` and installed

**Describe the bug**
When exporting to PDF, if you have to rewrite a file, the window with "Are you sure to overwrite?" pops up. 

If for whatever reason this window goes under the main file window, for example by clicking on it (see the screenshot) it has gone forever. There is no other way to exit from this situation than pressing "ESCAPE", but this is not so evident and the whole app seems frozen.

**To Reproduce**
1. export to PDF with an existing name
2. click as shown in the screenshot
3. Try to exit the dialog if you do not know about pressing ESC

**Expected behavior**
The "overwrite" dialog should not be allowed to go under anything

**Screenshots of Problem**

image

**Additional context**
Not a show-stopper, but it puzzled me during a while (and I did `killall xournalpp` before thinking of ESC).

bhennion:
Fix #3253

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2677,7 +2677,7 @@ auto Control::showSaveDialog() -> bool {
         Util::clearExtensions(fileTmp);
         fileTmp += ".xopp";
         // Since we add the extension after the OK button, we have to check manually on existing files
-        if (askToReplace(fileTmp)) {
+        if (askToReplace(fileTmp, GTK_WINDOW(dialog))) {
             break;
         }
     }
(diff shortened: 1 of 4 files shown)


rolandlo (approved):
LGTM

Issue #5555, opened 2024-03-12
Add declaration file for the lua api


atticus-sullivan:
### Is Your Feature Request Related to a Problem?

Writing with the lua api you always need to have the docs / the header file open which isn't that nice to code with.

### Feature Description

Nowadays most ("general purpose") editors can use a language server. For lua there is [LuaLS](https://github.com/LuaLS/lua-language-server). Using this, the coding experience is much nicer. Still for proper suggestions, the language server needs to get to know the xournalpp lua api. For this purpose one can write some sort of declaration file.

What do you think about adding such file to the repo? The big drawback I see is that we need to keep that file in sync with the implementation of the luaAPI (which doesn't quite work with the website currently tbh)

### Additional Context

I've already created such file for me personally (docs on the functions are pretty brief though) which I could share / add to the repo.

rolandlo:
That sounds like an interesting idea. How does one write a declaration file, are there some docs about it? Also please share an example, so we can get an idea of how it looks like.

atticus-sullivan:
Implemented/Closed via https://github.com/xournalpp/xournalpp/pull/5572

PR #5563, opened 2024-03-14
[luaAPI] add a declaration file

atticus-sullivan:
The declaration file can be used with
[luaLS](https://github.com/LuaLS/lua-language-server)

See #5555 for previous discussion and short example of usage in neovim.

TODOs:
- [ ] do we really want to place the file side-by-side with the implementation of the API?
- [ ] note about this on the website -> will be a PR on the other repository (also note that this is not the proper/complete documentation, for this one needs to have a look at the (comments in the) API implementation)
- [ ] Check out how to configure this generally

### DIFF
--- a/src/core/plugin/luapi_application.def.lua
+++ b/src/core/plugin/luapi_application.def.lua
@@ -1,0 +1,181 @@
+---@meta
+
+--*** app ***
+app = {}
+
+---Simple dialog window -- DEPRECATED
+---@param message string
+---@param options table<integer, string> integer to button texts
+---@return integer integer-key of the selected button
+function app.msgbox(message, options) end
+
+---Simple dialog window
+---@param message string
+---@param options table<integer, string> integer to button texts
+---@param cb string name of the callback function(button:integer) to call on user interaction
+---@param error boolean is this message an error (optional)
+function app.openDialog(message, options, cb, error) end
+
+---Renames file 'from' to file 'to' in the file system.
+---@param from string 
+---@param to string
+---@return number|nil Returns 1 on success, and (nil, message) on failure.
+---@return string
+function app.glib_rename(from, to) end
+
+---"Save as" dialog
+---@param filename string suggestion for a filename, defaults to "untitlted"
+---@return string path of the selected location
+function app.saveAs(filename) end
+
+---Allow to register menupoints/toolbar icons, this needs to be called from initUi
+---@param opts {menu: string, callback: string, toolbarID: string, mode:integer, accelerator:string} options (`mode`, `toolbarID` and `accelerator` are optional)
+---@return {menuId:integer}
+function app.registerUi(opts) end
+
+---Execute an UI action (usually internally called from Toolbar / Menu)
+---@param opts {action: string, group: string, enabled:boolean} options (`enabled` is `true` by default, `group` is only for debugging)
+function app.uiAction(opts) end
+
+---Execute an action from the sidebar menu
+---@param action string the desired action
+function app.sidebarAction(action) end
+
+---Execute an action from layer controller
+---@param action string the desired action
+function app.layerAction(action) end
+
+---Change color of a specified tool or of the current tool 
+---@param opts {color: integer, tool: string, selection: boolean} options (`selction=true` -> new color applies to active selection as well, `tool=nil` (unset) -> currently active tool is being used)
+function app.changeToolColor(opts) end
+
+---Get all colors of the currently active color palette
+---@return {color: integer, name: string} options
+function app.getColorPalette() end
+
+---Change page background of current page
+---@param background string backgroundType (e.g. "graph")
+function app.changeCurrentPageBackground(background) end
+
+---Select Background Pdf Page for Current Page
+---@param pageNr integer page-nunmber
+---@param relative boolean should `pageNr` be interpreted relative?
+function app.changeBackgroundPdfPageNr(pageNr, relative) end
+
+---Returns a table encoding all info on the chosen tool
+---@param tool string tool to get the information for
+---@return {} info varies a lot depending on the tool
+function app.getToolInfo(tool) end
+
+---Get the index of the currently active sidebar-page
+---@return integer pageNr pageNr of the sidebar page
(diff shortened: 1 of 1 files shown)


atticus-sullivan:
About configuration:

I just checked placing
```json
{
  "workspace.library": ["../../src/core/plugin/luapi_application.def.lua"],
}
```
as `.luarc.json` in the directory of the plugin. Turns out this is working, at least for me (in neovim).

The thing I'm really unsure about is what the term `workspace` refers to.
1. In **neovim** a language server is associated with a `root directory` which I believe to be the same as the `workspace` referenced in the [luaLS documentation](https://github.com/LuaLS/lua-language-server/wiki/Settings#workspacelibrary) but I'm not completely sure.
I'm setting this via `root_dir = function(fname) return util.root_pattern('main.lua')(fname) or util.find_git_ancestor(fname) end` but others might set it differently (or don't use `main.lua`) which would result in different (relative) paths.
2. In **vscode** the workspace maybe is something like the path where you opened the project (I really have no experience with vscode)? This also would result in different paths, as someone just working on a plugin might just open the directory of the plugin while others open the directory of the whole project.

So maybe the best way is to explain how to set this config, not provide a default one in the example plugin and also not add the configs to git (-> `gitignore`).

Can someone test the paths with vscode? (or knows it since I think the term workspace comes from the vscode world here)

atticus-sullivan:
> Also, there must already be such a script out there...

I'd expect one as well, but on my first fast research I didn't found one (yet). Maybe people are only using such a script in other projects and don't publish it as separate script.

But I agree, such an extraction script would make keeping declaration file and source in sync much easier.

rolandlo:
> > * do we really want to place the file side-by-side with the implementation of the API?
> 
> If we add the comments to the luapi_application.h in the somewhat standard format we have in parts of the code, e.g.
> 
Would the examples we have for applying the Lua library functions go into the @brief then? Or would you throw them away?

atticus-sullivan:
Just tried naively running doxygen to parse the doc-comments and get the result in a machine readable format. The output for one function looks something like this:

<details><summary>doxygen xml</summary>

```xml
      <memberdef kind="function" id="luapi__application_8h_1a3fb533faf6fe9e9a8daf0c47ad5f0507" prot="public" static="yes" const="no" explicit="no" inline="no" virt="non-virtual">
        <type>int</type>
        <definition>static int applib_getImages</definition>
        <argsstring>(lua_State *L)</argsstring>
        <name>applib_getImages</name>
        <param>
          <type>lua_State *</type>
          <declname>L</declname>
        </param>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
<para>Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack. Is inverse to app.addImages</para>
<para>Required argument: type (&quot;selection&quot; or &quot;layer&quot;)</para>
<para>Example: local images = app.getImages(&quot;selection&quot;)</para>
<para>return value: { { [&quot;x&quot;] = number, [&quot;y&quot;] = number, [&quot;width&quot;] = number, (width when inserted into Xournal++) [&quot;height&quot;] = number, (height when inserted into Xournal++) [&quot;data&quot;] = string, [&quot;format&quot;] = string, [&quot;imageWidth&quot;] = integer, [&quot;imageHeight&quot;] = integer, }, { ... }, ... } </para>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="src/core/plugin/luapi_application.h" line="2633" column="12" bodyfile="src/core/plugin/luapi_application.h" bodystart="2633" bodyend="2691"/>
      </memberdef>
```

</details>


<details><summary>for the following doc-comment</summary>

```cpp
/**
 * Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
 * Is inverse to app.addImages
 *
 * Required argument: type ("selection" or "layer")
 *
 * Example: local images = app.getImages("selection")
 *
 * return value:
 * {
 *     {
 *         ["x"] = number,
 *         ["y"] = number,
 *         ["width"] = number,    (width when inserted into Xournal++)
 *         ["height"] = number,   (height when inserted into Xournal++)
 *         ["data"] = string,
 *         ["format"] = string,
 *         ["imageWidth"] = integer,
 *         ["imageHeight"] = integer,
 *     },
 *     {
 *         ...
 *     },
 *     ...
 * }
 * @param type string "selection" or "layer"
 * @return {path:string, data:string, x:number, y:number, maxHeight:number, maxWidth:number, aspectRatio:boolean}[] images
 */
```

</details>

Issues:
- `@return` and `@param` don't seem to be recognized.

Most probably doxygen tries to link the doc-comment to the code, but the arguments do not match.

----

So then maybe I'd go the way of manually writing some sort of parser which checks for the functions found by e.g.
```bash
sed -n '/static const luaL_Reg applib/,${/applib_.*/{s/^[[:space:]]*{"\(.*\)",[[:space:]]*\(applib_.*\)},/\1\t\2/p}}' src/core/plugin/luapi_application.h
```

Searching for the functions which shall be checked isn't too hard (assuming `static const luaL_Reg applib` is always almost at the bottom of the file) using shell utilities like `sed`.
The next step I think would be to obtain the lines/the comment preceeding the respective function declaration which we are able to simply copy to the declaration file (if we write this doc-comment the right way). A single pass over the lines checking for `/**`, collecting following lines until `*/` and checking if the next line declares a function which occurred in the result of above.

Would using python be ok for you at this point ((compatible, no zsh etc) shells are always lacking the datastructures needed for this task)? We could make a task which automatically runs the script when pushing a change to `luapi_application.h` (I only have experience with github actions though and xournalpp uses azure runners).


---

Note: I'm really not good at writing parsers, so no hard feelings if you decide to do it differently / on your own ^^

rolandlo:
> Would using python be ok for you at this point

Sure. We have some python-scripts in the `scripts` folder of the repository already.

atticus-sullivan:
Since writing a script to generate the declaration/definition file is different than just kinda hardcoding it, I'm opening another PR #5572 for this. The PR is still a draft, since there are a few things (mainly organizational stuff) to talk about before merging, but from the coding part it should be already complete.

PR #5564, opened 2024-03-14
Fix path separator for Windows

rolandlo:
Now that `lgi` is shown to work on Windows (see installation instructions [here](https://github.com/xournalpp/xournalpp/discussions/4522#discussioncomment-4490657)) plugins must care about OS specific stuff like path separators.


rolandlo:

PR #5572, opened 2024-03-18
Python script to generate LuaLS definition file from luapi header file

atticus-sullivan:
The resulting declaration file can be used with
[luaLS](https://github.com/LuaLS/lua-language-server)

Note: The term definition might not be completely correct as the file only declares the functions, it doesn't really define it. Nevertheless, LuaLS uses the term 'definition files' so I stick to it when referring to LuaLS.

Replaces #5563 (hardcode definition file), now automatically generate it to make keeping in sync easier.

TODOs:
- [x] azure runner?
- [x] where to place/how to name the definition file (currently same path as `.h` but with `.def.lua` extension)
- [x] add respective annotations to the doc-comments in `luapi_application.h`
- [ ] if someone knows his/her way around python docstrings, feel free to correct my usage (I'm working the first time with them and especially tuples, lists and dicts might be documented in a non standard way)

See #5563 for example how to configure LuaLS.
See #5555 for references on LuaLS.

<details><summary>Smaller example</summary>

<details><summary>smaller luapi_application.h:</summary>

```cpp
/**
 * Renames file 'from' to file 'to' in the file system.
 * Overwrites 'to' if it already exists.
 *
 * @param from string
 * @param to string
 * @return number|nil Returns 1 on success, and (nil, message) on failure.
 * @return string
 *
 * Example:
 *   assert(app.glib_rename("path/to/foo", "other/bar"))
 *
 * Preferred to os.rename() because it works across
 * partitions. Uses glib's rename function.
 *
 * Returns 1 on success, and (nil, message) on failure.
 */
static int applib_glib_rename(lua_State* L) {
    GError* err = nullptr;
    xoj::util::GObjectSPtr<GFile> to(g_file_new_for_path(lua_tostring(L, -1)), xoj::util::adopt);
    xoj::util::GObjectSPtr<GFile> from(g_file_new_for_path(lua_tostring(L, -2)), xoj::util::adopt);

    g_file_move(from.get(), to.get(), G_FILE_COPY_OVERWRITE, nullptr, nullptr, nullptr, &err);
    if (err) {
        // return nil, error message
        lua_pushnil(L);
        lua_pushfstring(L, "%s (error code: %d)", err->message, err->code);
        g_error_free(err);
        return 2;
    } else {
        // return 1
        lua_pushinteger(L, 1);
        return 1;
    }
}


/**
 * Allow to register menupoints and toolbar buttons. This needs to be called from initUi
 *
 * @param opts {menu: string, callback: string, toolbarID: string, mode:integer, accelerator:string} options (`mode`, `toolbarID` and `accelerator` are optional)
 * @return {menuId:integer}
 *
 * Example 1: app.registerUi({["menu"] = "HelloWorld", callback="printMessage", mode=1, accelerator="<Control>a"})
 * registers a menupoint with name "HelloWorld" executing a function named "printMessage", in mode 1,
 * which can be triggered via the "<Control>a" keyboard accelerator
 *
 * Example 2: app.registerUi({callback ="blueDashedPen", toolbarId="CUSTOM_PEN_1", iconName="bluePenIcon"})
 * registers a toolbar icon named "bluePenIcon" executing a function named "blueDashedPen", which can be added
 * to a toolbar via toolbar customization or by editing the toolbar.ini file using the name "Plugin::CUSTOM_PEN_1"
 * Note that in toolbar.ini the string "Plugin::" must always be prepended to the toolbarId specified in the plugin
 *
 * The mode and accelerator are optional. When specifying the mode, the callback function should have one parameter
   that receives the mode. This is useful for callback functions that are shared among multiple menu entries.
 */
static int applib_registerUi(lua_State* L) {
    if (!plugin->isInInitUi()) {
        return luaL_error(L, "registerUi needs to be called within initUi()");
    }

    // discard any extra arguments passed in
    lua_settop(L, 1);
    luaL_checktype(L, 1, LUA_TTABLE);

    // Now to get the data out of the table
    // 'unpack' the table by putting the values onto
    // the stack first. Then convert those stack values
    // into an appropriate C type.
    lua_getfield(L, 1, "accelerator");
    lua_getfield(L, 1, "menu");
    lua_getfield(L, 1, "callback");
    lua_getfield(L, 1, "mode");
    lua_getfield(L, 1, "toolbarId");
    lua_getfield(L, 1, "iconName");
    // In Example 1 stack now has following:
    //    1 = {"menu"="MenuName", callback="functionName", mode=1, accelerator="<Control>a"}
    //   -6 = "<Control>a"
    //   -5 = "MenuName"
    //   -4 = "functionName"
    //   -3 = mode
    //   -2 = nil
    //   -1 = nil

    const char* accelerator = luaL_optstring(L, -6, "");
    const char* menu = luaL_optstring(L, -5, "");
    const char* callback = luaL_optstring(L, -4, nullptr);
    const ptrdiff_t mode = luaL_optinteger(L, -3, std::numeric_limits<ptrdiff_t>::max());
    const char* toolbarId = luaL_optstring(L, -2, "");
    const char* iconName = luaL_optstring(L, -1, "");
    if (callback == nullptr) {
        return luaL_error(L, "Missing callback function!");
    }

    size_t menuId = plugin->registerMenu(menu, callback, mode, accelerator);
    plugin->registerToolButton(menu, toolbarId, iconName, callback, mode);

    // Make sure to remove all vars which are put to the stack before!
    lua_pop(L, 6);

    // Add return value to the Stack
    lua_createtable(L, 0, 2);

    lua_pushstring(L, "menuId");
    lua_pushinteger(L, as_signed(menuId));
    lua_settable(L, -3); /* 3rd element from the stack top */

    return 1;
}

/*
 * The full Lua Plugin API.
 * See above for example usage of each function.
 */
static const luaL_Reg applib[] = {{"glib_rename", applib_glib_rename},
                                  {"registerUi", applib_registerUi},
                                  {nullptr, nullptr}};

/**
 * Open application Library
 */
inline int luaopen_app(lua_State* L) {
    luaL_newlib(L, applib);
    //	lua_pushnumber(L, MSG_BT_OK);
    //	lua_setfield(L, -2, "MSG_BT_OK");
    return 1;
}
```

</details>
<details><summary>Generated LuaLS definition file:</summary>

```lua
--- Renames file 'from' to file 'to' in the file system.
--- Overwrites 'to' if it already exists.
--- 
--- @param from string
--- @param to string
--- @return number|nil Returns 1 on success, and (nil, message) on failure.
--- @return string
--- 
--- Example:
---   assert(app.glib_rename("path/to/foo", "other/bar"))
--- 
--- Preferred to os.rename() because it works across
--- partitions. Uses glib's rename function.
--- 
--- Returns 1 on success, and (nil, message) on failure.
function app.glib_rename(from, to) end

--- Allow to register menupoints and toolbar buttons. This needs to be called from initUi
--- 
--- @param opts {menu: string, callback: string, toolbarID: string, mode:integer, accelerator:string} options (`mode`, `toolbarID` and `accelerator` are optional)
--- @return {menuId:integer}
--- 
--- Example 1: app.registerUi({["menu"] = "HelloWorld", callback="printMessage", mode=1, accelerator="<Control>a"})
--- registers a menupoint with name "HelloWorld" executing a function named "printMessage", in mode 1,
--- which can be triggered via the "<Control>a" keyboard accelerator
--- 
--- Example 2: app.registerUi({callback ="blueDashedPen", toolbarId="CUSTOM_PEN_1", iconName="bluePenIcon"})
--- registers a toolbar icon named "bluePenIcon" executing a function named "blueDashedPen", which can be added
--- to a toolbar via toolbar customization or by editing the toolbar.ini file using the name "Plugin::CUSTOM_PEN_1"
--- Note that in toolbar.ini the string "Plugin::" must always be prepended to the toolbarId specified in the plugin
--- 
--- The mode and accelerator are optional. When specifying the mode, the callback function should have one parameter
---    that receives the mode. This is useful for callback functions that are shared among multiple menu entries.
function app.registerUi(opts) end

```

</details>
</details>

### DIFF
--- a/azure-pipelines/gen-lua-def-run.yml
+++ b/azure-pipelines/gen-lua-def-run.yml
@@ -1,0 +1,40 @@
+trigger: none
+pr:
+  branches:
+    include:
+    - '*'
+  paths:
+    include:
+      - 'src/core/plugin/luapi_application.h'
+
+stages:
+- stage: 'Build_Test_Stage'
+  jobs:
+  - job: 'Build_Test'
+    pool:
+      vmImage: 'ubuntu-20.04'
+    displayName: 'Test if lua definitions file is up-to-date'
+    steps:
+    - checkout: self
+    - bash: |
+        sudo apt-get update
+        sudo apt-get install python3
+      displayName: 'Install python3'
+    - bash: |
+        echo $(git status)
+        echo "generate the file fresh"
+        if ! python3 scripts/lua_def_file.py ; then
+            # Failure: Generating script failed
+            echo "Error generating the definitions file"
+            exit 1
+        fi
+
+        # check if it matches the one present in the repository
+        if ! git diff --exit-code plugins/luapi_application.def.lua ; then
+            # Failure: lua definitions file is not up-to-date
+            echo "Error: generated lua definitions file does not match the present one"
+            exit 1
+        fi
+        echo "Success: lua definitions file is up-to-date"
+        exit 0
+      displayName: 'Generate lua definitions file and check if the one in the repo is still up-to-date'
--- a/scripts/lua_def_file.py
+++ b/scripts/lua_def_file.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python
+
+import re
+import sys
+import os.path
+
+# def gather_functions(file_name:str):
+def gather_functions(file_name):
+    '''
+    Gathers the (lua) API functions defined in the given file
+
+        Parameters:
+            file_name (str): Path to the file to process
+
+        Yields:
+            (c_function_name (str), lua_function_name (str))
+
+        Assumption: lua API is defined via "static const luaL_Reg applib"
+    '''
+    # define the regex pattern and compile them only once for multiple use later
+    start_pattern = re.compile(r'static const luaL_Reg applib\[\] = {')
+    main_pattern = re.compile(r'\s*{"(.*)",\s*(applib_.*)},')
+
+    # already capturing lines?
+    capture = False
+
+    with open(file_name, 'r') as file:
+        for line in file:
+            # search for starting pattern (only if not already capturing anyhow)
+            if not capture and start_pattern.match(line):
+                capture = True
+                # make line fit for further detection of an API function
+                line = start_pattern.sub("", line, count=1)
+
+            if capture:
+                # extract information from lines if matching
+                match = main_pattern.match(line)
+                if match:
+                    groups = match.groups()
+                    yield (groups[1], groups[0])
+
+    if not capture:
+        raise Exception("luaL_Reg applib not found -> is this file really defining the lua-API?")
+
+
+# def docs_for_functions(file_name:str, functions: dict[str, str]):
+def docs_for_functions(file_name, functions):
+    '''
+    Search for doc-comments for given functions in given file
+
+        Parameters:
+            file_name (str): Path to the file to process
+            functions (dict[str,str]): mapping c_function_name -> lua_function_name
+
+        Yields:
+            (lua_function_name (str), comments (list[str]), params (list[str]))
+    '''
+    # define the regex pattern and compile them only once for multiple use later
+
+    # patterns to detect start/end of the comment
+    start_pattern = re.compile(r'\s*/\*\*')
+    end_pattern = re.compile(r'(.*\*/)')
+
+    # patterns to extract information
+    extract_func = re.compile(r'\s*static.*(applib_.*)\(')
+    extract_param = re.compile(r'\s*@param\s+([^\s]+)')
+
+    # pattern to replace the C comment prefix (' *') with the one of lua(LS) ('---')
+    repl_comment_prefix = re.compile(r'^\s\*\s?')
+
+    # works like a simple state maching -> define possible states here
+    STATE_SEARCHING     = 0 # searching for doc-comment
+    STATE_COLLECTING    = 1 # collect the doc-comment
+    # check if the line following the doc-comment references a function given
+    # in 'functions'
+    STATE_CHECKING_FUNC = 2
+
+    with open(file_name, 'r') as file:
+        # collect the lines with a doc-comment
+        comments = list()
+        # collect the parameters referenced in the doc-comment
+        params = list()
+        # start in searching state
+        state = STATE_SEARCHING
+
+        for line in file:
+            skip_empty = False
+            # strip only one trailing newline
+            if line[-1] == "\n": line = line[:-1]
+            # switch-case over the state
+            # ATTENTION: working with if, no elif => continue needed
+            if state == STATE_SEARCHING:
+                # search for a line starting the doc-comment
+                if start_pattern.match(line):
+                    state = STATE_COLLECTING
+                    # replace the comment prefix
+                    line = start_pattern.sub("", line, count=1)
+                    skip_empty = True
+                    # fallthrough-like to collect this line as usual
+                    # 1. line might contain parameter
+                    # 2. line might also end the comment as well
+            if state == STATE_COLLECTING:
+                # check if line ends the comment
+                match = end_pattern.match(line)
+                if match:
+                    skip_empty = True
+                    state = STATE_CHECKING_FUNC
+                    line = match.groups()[0]
+                    line = end_pattern.sub("", line, count=1)
+                # check if the line contains the declaration of a parameter
+                match = extract_param.search(line)
+                if match:
+                    params.append(match.groups()[0])
+                # collect the line after replacing the leading comment prefix
+                line = repl_comment_prefix.sub("", line, count=1)
+                if line != "" or not skip_empty:
+                    comments.append("--- " + line)
+                continue
+            if state == STATE_CHECKING_FUNC:
+                # check if function given in 'functions' is declared in the line
+                match = extract_func.match(line)
+                if match and match.groups()[0] in functions:
+                    yield (functions[match.groups()[0]], comments, params)
+                    # remove function to ensure the doc is only emitted once
+                    # and to mark it as processed
+                    del functions[match.groups()[0]]
+                # re-initialize state variables
+                comments = list()
+                params = list()
+                state = STATE_SEARCHING
+                continue
+
+        # 'functions' only contains the unprocessed functions -> check if there
+        # are any left
+        if len(functions) > 0:
+            raise Exception(f"doc strings for functions [{', '.join(functions.values())}] missing")
+
+# def fmt_luaLS_def(file, function_name:str, comments:list[str] = [], params:list[str] = []):
+def fmt_luaLS_def(file, function_name, comments = [], params = []):
+    '''
+    Emit code for a luaLS definition to given file for a function
+
+        Parameters:
+            file (file): file to which to emit the code
+            function_name (str): name of the function to declare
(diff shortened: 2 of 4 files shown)


atticus-sullivan:
This PR is still a draft, since there are a few things (mainly organizational stuff, see list above) left to talk about before merging.
But from the basic coding part it should be already complete.

atticus-sullivan:
Thanks first of all for the review. I also think generating this file only when necessary is the go-to way. I've once worked on a project where they generated some docs via a github action and this action added the "compiled" docs via a commit automatically (so the dev never needed to generate the doc manually).

Not sure if this is possible with the azure runner (or moreover if this is wanted by the maintainers, it's a bit of a weird workflow after all and you need to make sure to pull the commit of the action/runner after pushing). But still like you mentioned we could make it like with `clang-format`. and check if the runner generated file is the same like in the repository.
This way the dev needs to run this script manually, but only if working on the luaAPI and we see the failing pipeline in the PR so we can aid in generating the file.

Do you know who would need to write the azure-runner?

Also before adding the runner, we should write the doc comments for the functions currently in the API.

Note for me:
- [ ] We then need to make sure the output is "stable", so the order stays the same (working with dicts this not necessarily needs to hold especially if the script potentially is run with different python versions).

atticus-sullivan:
Hm doesn't seem to work (the azure pipeline). Are you it works this way? Or could you check if I maybe copied too much so maybe some identifier is clashing with the clang-format action?

rolandlo:
> Uh, bummer. We may have to parametrize something on the Azure pipeline interface, I don't know. Adding a Github Action to a PR does work and I assumed the Azure pipeline was similar. @rolandlo @Febbe Does one of you know how to add a CI check?

I think we just have to create a new pipeline and point to the .yaml when creating it.
grafik
Currently this PR is not a branch on xournalpp/xournalpp, so we can't use it for our repo. I can create a branch in rolandlo/xournalpp from this PR and test the .yaml out before we go forward.

atticus-sullivan:
Well the only thing still missing is the decision where the definitions file should be placed. So far it resides alongside the `.h` file (defining the API) with the extension `.def.lua` (so still in the `src/` folder).

rolandlo:
Here is the result of the first run:

grafik

atticus-sullivan:
Ok thanks for running it. The errors are a bit weird though, as to this point the only difference is that I don't define the `CLANG_FORMAT_VERSIONS` variable (and install `python3` instead of `clang-format-xx`).

On the other hand, I just blindly copied the `clang_format` pipeline to some extend. Actually these `System.` variables and finding the merge-commit is not necessary at all for this purpose.

Maybe @rolandlo could check if it runs now.

rolandlo:
It succeeded
grafik

atticus-sullivan:
Oh well but it shouldn't have due to the error while running the script :thinking:
EDIT: Ok the error handling in the pipeline is still flawed.

Could you try running [9c6fdc1](https://github.com/xournalpp/xournalpp/pull/5572/commits/9c6fdc189236ca23a7f703ede2b77b79c33e3bb6)? (it should fail)

Any also [1c5426e](https://github.com/xournalpp/xournalpp/pull/5572/commits/1c5426e6e1a41b1b11971e7eb8083662caee121b) which should succeed.

Then try editing the doc-comment of an API function and then rerun the pipeline which now should be failing again. Running the script, commiting the updated `.def.lua` should make the pipeline succeeding again.

---

I think this would be all we need to test in this regard. (Running this various tests is a little more effort than only running the pipeline on the current HEAD of this PR, so take your time)

rolandlo:
Not relevant any more, but as a self note: System.PullRequest.TargetBranch: see https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
grafik

Not sure how to set a branch policy if the repo is on Github and not on Azure Repos.

atticus-sullivan:
Well it seems to be working for the clang-format pipeline (this pipeline uses this variable, apparently successfully) somehow :shrug:

rolandlo:
[9c6fdc1](https://github.com/xournalpp/xournalpp/pull/5572/commits/9c6fdc189236ca23a7f703ede2b77b79c33e3bb6)
grafik

[1c5426e](https://github.com/xournalpp/xournalpp/pull/5572/commits/1c5426e6e1a41b1b11971e7eb8083662caee121b)
grafik

rolandlo:
After fixing the extraneous bracket:
grafik

After changing a doc comment:

grafik

After running the script and committing the changes:
grafik

So please fix the bracket and then it should all work.

azure-pipelines/gen-lua-def-run.yml
```diff
@@ -0,0 +1,39 @@
...
+            exit 1
+        fi
+
+        # check if it matches the one present in the repository
+        if ! git diff --exit-code src/core/plugin/luapi_application.def.lua ; then
+            # Failure: lua definitions file is not up-to-date
+            echo "Error: generated lua definitions file does not match the present one"
+            exit 1
+        fi
+        echo "Success: lua definitions file is up-to-date"; exit 0; }
```
rolandlo, line 38:
```suggestion
        echo "Success: lua definitions file is up-to-date"; exit 0
```

scripts/lua_def_file.py
```diff
@@ -0,0 +1,193 @@
+#!/bin/python3
```
rolandlo, line 1:
```suggestion
#!/usr/bin/python3
```
And why not make the script executable?

atticus-sullivan: No particular reason, I just looked at the other python script in that folder which isn't executable and I just followed.

atticus-sullivan:
Thanks again for testing. @rolandlo do you have a preference where to place the definitions file? (see thoughts above about `src/` and build-dir)

rolandlo:
> Thanks again for testing. @rolandlo do you have a preference where to place the definitions file? (see thoughts above about `src/` and build-dir)

I would probably put it into the `resources` folder (where the LaTeX templates are). Those files will get installed in `usr/share/xournalpp/resources` (`Xournal++.app/Contents/Resources` on MacOS, `C:\Programs\Xournal++\share\xournalpp\resources` on Windows).

An alternative would be to put it into the `plugins` folder.

atticus-sullivan:
I feel like `plugins/` is a bit more natural.

rolandlo (changes requested):
I have some comments, most of them are minor improvements in the docstrings. Very nice python script by the way!
Your work makes writing plugins a whole lot more comfortable. Great job!

atticus-sullivan:
Alright, should be fixed now. Thanks for reading through the whole lua-docstrings and fixing even the typos.

rolandlo (approved):
Please squash your commits into one or a few commits.

atticus-sullivan:
Done. One commit per step. Let me know if you'd like a different split? (or none at all)

rolandlo:
I'm fine with the split. Merging in 24 hours if no objections are raised.

atticus-sullivan:
:+1:
Also don't forget to add the new pipeline (and maybe check if the current state really still passes it to make sure not to confuse people in other PRs)

rolandlo:
> 👍 Also don't forget to add the new pipeline (and maybe check if the current state really still passes it to make sure not to confuse people in other PRs)

I have added the new pipeline. It still passes.

grafik

Thanks again for this very nice contribution.

PR #4813, opened 2023-04-20
Added more precise edge panning for large selections

c-ola:
This adds the functionality mentioned in #4476.  Now if a selection is larger than the visual view size, the mouse position is considered for panning when moving the selection.

Issue #4476, opened 2022-12-05
consider cursor position instead of selection when scrolling automatically


Ingegneus:
**Is your feature request related to a problem? Please describe.**
when you have a large selection (a screenshot or just a selection of strokes) that is similar in size or larger than your visible draw area and try to move/reposition your selection, the page automatically scolls up or down because the selection ist close to the edge of the draw area. this makes it next to impossible to move large selections around precisely, because you have to scroll out in order to not trigger the scrolling.
![image](https://user-images.githubusercontent.com/78076361/205609061-01f2f78f-a60d-4975-99f2-7fca6f37dea8.png)

**Describe the solution you'd like**
IMO it would be better to trigger the scrolling if the cursor is next to the edge and not the selection itself.

K4LCIFER:
I have to agree that this behaviour has been troublesome; however, it could be argued that if the image is larger than the viewed area, you need to scroll out to position it properly anyways.

Issue #4172, opened 2022-07-30
 Dragging an image upwards when screen wrap is enabled


vladQll:
Affected versions:** **
 - OPERATING SYSTEM: [Windows].
 - What version of libgtk do you use [3.24.34].
 - Xournal++ version: [1.1.1+dev]
 - Installation method: [.exe]

**Describe the bug**
 - when I select fit screen and then paste a 1280 x 960px image and try to drag the image to be able to place it on top of the paper, it does not allow me to do it in the intuitive way, even more when I try to drag up, the image automatically slides down the only solution I found is to zoom out with the Zoom tool so I can quickly drag up.

**Playback
1. adjust the screen 
2. copy and paste a large image of about 1280 x 960px or larger
3. try to drag the image to the top of the screen.

**Expected behavior
I would like to be able to easily drag the image like in a word or libreOffice document.

**Screenshots of the problem**.

https://user-images.githubusercontent.com/28522011/181905840-31c73048-c998-4426-9bd1-f9802a1661dd.mp4

jduf:
I face the same issue. Using the arrow keys helps me a bit.

PR #5575, opened 2024-03-20
Improved edge panning on large selections

c-ola:
This PR fixes issues mentioned in #4476 and #4172.

It can be difficult to move around selections larger than the view (in either dimension) due to the current edge pan handler only considering how far the selection is outside of the view.

 I fixed this by also considering (for only the large selections) how far the mouse position is from the center of the view, and then only panning if it's outside of a specified range from the middle (it feels nicer than if it were to just always move when you click on a selection).

I tried to fix this about a year ago in #4813 with a nearly identical solution but it was crashing for someone else in `PageView::getXournal()` (the backtrace can be seen in #4813). I have been unable to get it to crash in my testing and I'm unsure to why it was crashing with my previous additions at all.

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -857,24 +857,40 @@ bool EditSelection::handleEdgePan(EditSelection* self) {
 
     // Helper function to compute scroll amount for a single dimension, based on visible region and selection bbox
     const auto computeScrollAmt = [&](double visMin, double visLen, double bboxMin, double bboxLen,
-                                      double layoutSize) -> double {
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
This time I don't get any weird crashes. It works nicely, good job!

rolandlo:
Indeed, merging this on the release branch makes more sense.

c-ola:
> @c-ola would you be able to move this pr to the release branch, so that your changes would make it to v1.2.4? It essentially means creating a branch based on release-1.2, and cherry-picking your commits there. It may then be easier to open a new PR using that new branch.

Yeah I'm on it.

PR #5583, opened 2024-03-23
Fix various Dark/Light Theme inconsistencies

### LINKED ISSUE #4336: Apply dark / light mode without a restart
Applying a dark or light theme currently requires to go to the settings, to (un)check a checkbox and to restart Xournal++.
Doing this twice a day, in the mornings and in the evenings, can be cumbersome.

It'd be great if Xournal++ followed the system theme automatically. 

E.g. when the global theme on Plasma is changed from Breeze Twilight to Breeze Dark, other GTK apps (Firefox/Thunderbird/Gajim/etc.) automatically switch from Adwaita to Adwaita Dark and vice versa, without the need of a restart.

### LINKED ISSUE #5422: Dark Theme won't go away
### Operating System

Linux

### (Linux only) Distribution

Fedora 39

Gnome

Wayland

### Installation Method

dnf 

### Xournal++ Version

1.2.2

### libgtk Version

3.24.38

### Bug Description

As you can see in the following image the dark theme is disabled but it persist even after I reboot my operating system:

image

Please may be provide a command line option to disable or enable dark mode as several options such as Latex have very bad visibility in dark mode. Its hard to identify icons due to poor contrast (see below screenshot: rendered latex equations can't be seen)!

image

### Expected Behaviour

Ability to disbale dark theme

### Steps to Reproduce

See screenshot above

### Additional Context

_No response_

### Requested Solution

Please provide a cli option to force light mode `--light` or dark mode `--dark`

### LINKED ISSUE #5445: Dark theme does not work
### Operating System

Linux

### (Linux only) Distribution

Debian

xfce

X.Org

### Installation Method

Flatpak

### Xournal++ Version

1.2.2, 09ae3333-dirty

### libgtk Version

3.24.38

### Bug Description

When dark theme was enabled, the application does not properly use the dark theme

### Expected Behaviour

The dark theme to be properly used

### Steps to Reproduce

1. Enable dark theme in the preferences > view > colors
2. Restart

### Additional Context

Screenshot_2024-01-11_14-48-45

bhennion:
Fixes #4336 Fixes #5422 Fixes #5445

This PR changes the "Dark theme" setting to become a combo box with 3 options: USE_SYSTEM, FORCE_LIGHT or FORCE_DARK.
With this, the gtk theme is always consistent with the icon set.

There are themes out there for which this does not work: e.g. breeze-gtk does not seem to allow the app to control its variant (dark/light) via gtk_settings::gtk-application-prefer-dark-theme.
For those users, the bug should be reported to the theme's maintainer or packager maybe.

### DIFF
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -44,6 +44,13 @@
 #include "config-dev.h"          // for TOOLBAR_CONFIG
 #include "filesystem.h"          // for path, exists
 
+#ifdef __APPLE__
+// the following header file contains a definition of struct Point that conflicts with model/Point.h
+#define Point Point_CF
+#include <CoreFoundation/CoreFoundation.h>
+#undef Point
+#endif
+
 using std::string;
 
 
@@ -237,7 +244,22 @@ static ThemeProperties getThemeProperties(GtkWidget* w) {
         props.darkSuffix = sm[2].length() > 0;
     }
     gboolean dark = false;
+
+#ifdef __APPLE__
+    CFStringRef interfaceStyle =
+            (CFStringRef)CFPreferencesCopyAppValue(CFSTR("AppleInterfaceStyle"), kCFPreferencesCurrentApplication);
+    if (interfaceStyle) {
+        char buffer[128];
+        if (CFStringGetCString(interfaceStyle, buffer, sizeof(buffer), kCFStringEncodingUTF8)) {
+            std::string style = buffer;
+            if (auto pos = style.find("Dark"); pos != std::string::npos) {
+                dark = true;
+            }
+        }
+    }
+#else
     g_object_get(gtk_widget_get_settings(w), "gtk-application-prefer-dark-theme", &dark, nullptr);
+#endif
 
     g_message("Extracted theme info: Name = %s, rootname = %s, dark = %s", name.get(), props.rootname.c_str(),
               dark ? "true" : "false");


rolandlo:
This looks like a great attempt to get dark/light theme working. Unfortunately even on a widely popular distro like Ubuntu it doesn't work (with the theme variant). In my case I get:
grafik
(which is kind of self-contradicting) and I can't change that from the Ubuntu settings to Theme name: Ubuntu-Yaru-dark, Theme variant: dark. If I want to change the setting I have to edit `~/.config/gtk-3.0/settings.ini`

I'm wondering if we could use the `color-scheme` gsetting instead to determine the theme variant like the AppImage does, see https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/pull/58/files.

rolandlo:
Well yes, with Yaru-purple and `gtk-application-prefer-dark-theme=1` in `~/.config/gtk-3.0/settings.ini` (or the "Force dark" option in the Xournal++ settings) it works. The trouble is I can't follow the system theme and simply use the Dark style switch here: 
grafik

rolandlo:
> I think the Adwaita:dark can be fixed easily. The follow system theme issue, I have no idea. What's the terminal output is this case?

I always get Theme name: Adwaita, Theme Variant: light when following the system theme, regardless of whether I use `GTK_THEME=Adwaita:dark` or not. When using that option everything except the icons use dark theme.

rolandlo:
One question: Is it possible (in Linux) to react to theme (and dark-preference) changes, so that when a user changes the theme the appearence of Xournal++ changes automatically (without opening the preferences)?

rolandlo:
> > > One question: Is it possible (in Linux) to react to theme (and dark-preference) changes, so that when a user changes the theme the appearence of Xournal++ changes automatically (without opening the preferences)?
> > 
> > 
> > Probably, lemme try.
> 
> This is now done. 

That's fantastic. Our users will love it! It works great on Ubuntu 24.04.

> I also added a bit more terminal output to see if something can be done on MacOS.
I will test it when I'm back home (on Saturday or Sunday).

rolandlo:
From https://gitlab.gnome.org/GNOME/gtk/blob/90d84a2af8b367bd5a5312b3fa3b67563462c0ef/gtk/gtksettings.c#L1567-L1622
when the css provider loads a theme, it first checks the environment variable `GTK_THEME` for determining theme name and theme variant. When it is not set it uses the "gtk-theme-name" and "gtk-application-prefer-dark-theme" properties or fallsback to the default theme. These two properties are not updated by the `GTK_THEME` variable.

So when we want to handle the case where the user (or e.g. a script in the AppImage) sets `GTK_THEME` we should first read that variable as well and use it to decide theme name and theme variant.

rolandlo:
> > From https://gitlab.gnome.org/GNOME/gtk/blob/90d84a2af8b367bd5a5312b3fa3b67563462c0ef/gtk/gtksettings.c#L1567-L1622 when the css provider loads a theme, it first checks the environment variable `GTK_THEME` for determining theme name and theme variant. When it is not set it uses the "gtk-theme-name" and "gtk-application-prefer-dark-theme" properties or fallsback to the default theme. These two properties are not updated by the `GTK_THEME` variable.
> > So when we want to handle the case where the user (or e.g. a script in the AppImage) sets `GTK_THEME` we should first read that variable as well and use it to decide theme name and theme variant.
> 
> Nice find!! I just pushed a commit doing that

It works nicely on Linux. I will test on MacOS and Windows as well when I'm home.

rolandlo:
> > > One question: Is it possible (in Linux) to react to theme (and dark-preference) changes, so that when a user changes the theme the appearence of Xournal++ changes automatically (without opening the preferences)?
> > 
> > 
> > Probably, lemme try.
> 
> This is now done. I also added a bit more terminal output to see if something can be done on MacOS.

The terminal output doesn't seem to help much. In dark mode I get:
grafik

However the following code block seems to work to detect the system dark preference (tested on MacOS Catalina)
```cpp
    bool dark = false;
    CFStringRef interfaceStyle = (CFStringRef)CFPreferencesCopyAppValue(CFSTR("AppleInterfaceStyle"), kCFPreferencesCurrentApplication);
    if (interfaceStyle) {
      char buffer[128];
      if (CFStringGetCString(interfaceStyle, buffer, sizeof(buffer), kCFStringEncodingUTF8)) {
        std::string style = buffer;
        g_message("Interface stlye: %s", style.c_str());
        if (auto pos = style.find("Dark"); pos != std::string::npos) {
            dark = true;
        }
      }
    }
```

It uses the CoreFoundation C API, so include the header file via
```cpp
#include <CoreFoundation/CoreFoundation.h>
````

**Comment by rolandlo** [comment]
For Windows there are docs here how to determine the system dark mode preference: https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes#know-when-dark-mode-is-enabled

**Comment by rolandlo** [comment]
> I don't quite feel confident I can cook up something without being able to test it directly. Maybe you could write a commit for MacOS and one fo Windows and either append them to this PR or to a follow-up?

I have added a commit for MacOS. It works for me (on MacOS Catalina).

Turns out the Windows counterpart is more troublesome. The `winrt` header 
```c++
#include <winrt/Windows.UI.ViewManagement.h>
```
 must be included, which is not available by default. After installing [mingw-w64-x86_64-cppwinrt](https://packages.msys2.org/package/mingw-w64-x86_64-cppwinrt?repo=mingw64) the compiler complains that C++/WinRT requires coroutine support, which is currently missing and recommends to enable C++20 in the compiler. I don't think we want to do that at this moment. So Windows users will have to use "Force dark" or "Force light" in the settings and can't just follow the system theme.

rolandlo:
> Should we merge this then?

I think so, yes.

PR #5591, opened 2024-03-26
Remove dummy GtkWidget in TextEditor

bhennion:
Remove dummy GtkWidget in TextEditor
        * Key bindings are now handled by a std::unordered_map and do
          not involve GTK anymore -> thus now gtk4-compatible...
        * All Key bindings are now in a single file


src/core/control/tools/TextEditor.cpp
```diff
@@ -618,7 +563,7 @@ void TextEditor::markPos(double x, double y, bool extendSelection) {
     findPos(&newplace, x, y);
 
     // call moveCursor() even if the cursor hasn't moved, since it cancels the selection
```
rolandlo, line 565:
```suggestion
    // call moveCursorIterator() even if the cursor hasn't moved, since it cancels the selection
```

src/core/control/tools/TextEditor.cpp
```diff
@@ -579,7 +524,7 @@ void TextEditor::moveCursor(GtkMovementStep step, int count, bool extendSelectio
     }
 
     // call moveCursor() even if the cursor hasn't moved, since it cancels the selection
```
rolandlo, line 526:
```suggestion
    // call moveCursorIterator() even if the cursor hasn't moved, since it cancels the selection
```

rolandlo (approved):
This is a nice simplification. It works fine and I have only some nitpicks about comments.

PR #5596, opened 2024-04-01
Draft: Transformable elements proposal

### LINKED ISSUE #918: Image rotation support
https://drive.google.com/file/d/1l94pBaPIhii599Tk5xJOKxAgqzah5Z79/view?usp=sharing
I'm a math teacher. I'm trying to use the ruler and the protractor in Xournal++ adding two images. How do I rotate the image of the ruler around the top left corner and the image of the protractor around his center to measure the angle? Thanks!

### LINKED ISSUE #2040: Text rotation
No, it's just a new function

It would be useful to be able to rotate text object in a similar way to images.

Make a PNG of the text with other application and import it. I have to take angle measurements to get the exact orientation...

**Additional context**
To annotate vertical or oblique graphic objects, to add diagonally written "ads" that attract attention...
With an alpha option, it would make it possible to make watermark-style notations.

### LINKED ISSUE #4899: Rotate the added text
**Resuest a enhancement for rotate the text that can be added**

That can be saved and exported to PDF.

Febbe:
This PR tends to implement one of the most desired enhancements (#918), to rotate Images, Text and Latex.
For this, I replaced every explicit position, scale and rotation state in the Elements with affine transformations.
They are never applied to the data, instead the affine transformation is handled to cairo, which itself applies an affine transformation anyway. 
This has the advantage, 
 - that we get rotation support out of the box via cairo and 
 - that we can reduce computing power, since only one transformation is applied to each Shape
 - when cairo drawing uses hardware acceleration, strokes now also will use hardware acceleration for the transformation

Currently, this is not ready to merge, since several code changes added bugs, which are not fixed yet.
But testers and interested people can test it out and give feedback.
Also, it's not possible, to store the transformations.

Animation

fixes #918 
fixes #2040 
fixes #3622 
fixes #4899 

It will require changes to #3931 and will simplify the implementation.
Only a mirror matrix must be applied to the selection with a pivot point/dimension.
E.g. Matrix(pivot) * mirrorMatrix * Matrix(-pivot)


Febbe:
@xournalpp/core we need to discuss what we want to do with the file format.
My proposal is, to 
 - advance the version of it and 
 - to add the ability, to export to an older version.
 - old versions should still be loadable.

rolandlo:
> @xournalpp/core we need to discuss what we want to do with the file format. My proposal is, to
> 
>     * advance the version of it and
> 
>     * to add the ability, to export to an older version.
> 
>     * old versions should still be loadable.
> 
> 
> But this will be again a lot of work.

That sounds reasonable. How would you export a rotated images/text/TeXImages though to an older version?

rolandlo:
> * when cairo drawing uses hardware acceleration,

Does it though? I do not think so. For hardware acceleration one has to use the Snapshot/Render Nodes approach in Gtk 4. I have created an example of that in Workbench (the Snapshot demo).

Febbe:
> > @xournalpp/core we need to discuss what we want to do with the file format. My proposal is, to
> > 
> >     * advance the version of it and
> > 
> >     * to add the ability, to export to an older version.
> > 
> >     * old versions should still be loadable.
> > 
> > 
> > But this will be again a lot of work.
> 
> That sounds reasonable. How would you export a rotated images/text/TeXImages though to an older version?

My idea is, to render the rotated PDFs onto a PDF surface and put it as binary data into the xournalpp file.

rolandlo:
> I am totally not sure. I've read something about, that this is implementation defined. Could be, that it only works on Linux with OpenGL or that I misinterpreted the section I've read.

The OpenGL support was removed and has basically unmaintained for a long time before. See e.g. https://news.ycombinator.com/item?id=39442886

I just wanted to argue about this fact about Cairo hardware acceleration, not about the general approach (which makes sense). Yes, in the Snapshot/RenderNodes world using a `Gsk.TransformNode` or `gtk_snapshot_transform_matrix` is much simpler than doing extra calculation.

One thing that looked troublesome to me with the transformation matrix approach is scaling strokes, since the stroke segments get distorted when the transformation is not a similarity. This makes a difference when scaling just in one direction. How do you plan to deal with that?

Febbe:
> One thing that looked troublesome to me with the transformation matrix approach is scaling strokes, since the stroke segments get distorted when the transformation is not a similarity. This makes a difference when scaling just in one direction. How do you plan to deal with that?

Right, I have this on my list. @bhennion has the approach, to add an extra parameter in his transformation matrix (don't have his branch on the hand right now).
But I would like to split the stroke size from the basic transformation. 

My approach for strokes would be, to also add a transformation matrix for the stroke size, and to scale it like:

`| TM * (z,z) |`

Febbe:
> > But testers and interested people can test it out and give feedback.
> 
> 1. How can I test this? Do I open [this](https://github.com/Febbe/xournalpp/tree/transformable_elements) and build xournalpp according to the build instructions?
> 2. What would you need to be tested, @Febbe? Just if rotating stuff works? Or also if errors / warnings are printing if build with debug flags on? Are there certain things / steps a "test" should include?

Yes, you have to checkout my branch and build xournalpp yourself.

It's a proposal not a final implementation. So general Feedback about the look and feel is enough. E.g. a feature is missing. Rotating and scaling behaves unexpected.

Developers and maintainers should check build warnings and console/debug warnings themself before they merge. This is nothing a user should take care of.

eyalroz:
@bhennion : Does the change of milestone to "future major version" mean that this is delayed into the far future? :-(

JMak22:
Not a contributor,  long-time user of Xournal++, looking at this PR because rotation of elements is something I've seen many users requesting/commenting on/finding workarounds.

Would it make sense to consider a smaller intermediate step, like fixed-angle rotations only (ex 90° / 180°) for user-added elements on the page?

Many daily workflows, like annotating PDFs/lecture slides or rearranging existing docs for annotation only needs quarter-turn rotations. If this could be done with fewer structural changes, it might give functionality sooner while the full transformation system is being worked out.

Of course, if trying for fixed rotations will break things then this is moot, but I wanted to raise the idea in case it helps reduce scope for an incremental release.

Thanks for all you do already, it's great to have a tool like Xournal++!

Febbe:
> Not a contributor,  long-time user of Xournal++, looking at this PR because rotation of elements is something I've seen many users requesting/commenting on/finding workarounds.
> 
> Would it make sense to consider a smaller intermediate step, like fixed-angle rotations only (ex 90° / 180°) for user-added elements on the page?
> 
> Many daily workflows, like annotating PDFs/lecture slides or rearranging existing docs for annotation only needs quarter-turn rotations. If this could be done with fewer structural changes, it might give functionality sooner while the full transformation system is being worked out.
> 
> Of course, if trying for fixed rotations will break things then this is moot, but I wanted to raise the idea in case it helps reduce scope for an incremental release.
> 
> Thanks for all you do already, it's great to have a tool like Xournal++!

The rotation system itself works mostly.
Problem is the file format, it never supported transformations. Therefore the file format version has to be advanced. This will break old xournalpp versions. Also exporting to the old file format won't work, as soon transformations are used (or they will be lost on save).

The 90°*x step may work. But I wouldn't implement a different approach, just limit this approach to only support 90°x. And exclude every ressource which can't be embedded into the file format without changing it. 
With that, you can rotate the picture itself and store a rotated Ressource in the file.

bhennion:
Closing in favor of #7712

PR #5601, opened 2024-04-04
Add affine transformation matrix

Febbe:
This adds a basic affine transformation matrix similar to the cairo_matrix_t, but independent and in the xoj namespace.
 - fixed an issue of a missing implementation of the `ostream& operator<<` overload for Rectangles.


Febbe:
@bhennion what do you think about that simpler version of an affine transformation? I did not include a scaling parameter for strokes, since this is bound to strokes only. I also think, that the width of a stroke should not be a parameter of the element's virtual scale function.

Should I move all matrix functions into the header file? This might improve performance, but it will increase build time a bit.

bhennion (commented):
This looks reasonible to me. Of course, the question of speed an efficiency is crucial, so having a benchmark (e.g. as a disabled test unit) wouldn't hurt.

Febbe:
> > Should I move all matrix functions into the header file? This might improve performance, but it will increase build time a bit.
> 
> Benchmark it and if it gives a significant execution speed gain, then yes!

Interesting, 10 times faster, using constexpr

```
PS C:\Users\Febbe\workspace\xournalpp\test\bench\matrix\build> .\Release\bench_matrix.exe
2024-04-05T00:32:46+02:00
Running C:\Users\Febbe\workspace\xournalpp\test\bench\matrix\build\Release\bench_matrix.exe
Run on (12 X 4691 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 1024 KiB (x6)
  L3 Unified 32768 KiB (x1)
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
v1               9.31 ns         8.37 ns    112000000
v2              0.943 ns        0.753 ns    746666667
```

src/util/include/util/detail/MatrixMultiply.h
```diff
@@ -0,0 +1,41 @@
...
+    util::Point<double> p3{rect.x, rect.y + rect.height};
+    util::Point<double> p4{rect.x + rect.width, rect.y};
+    auto s1 = multiply(matrix, p1);
+    auto s2 = multiply(matrix, p2);
+    auto s3 = multiply(matrix, p3);
+    auto s4 = multiply(matrix, p4);
+    auto x = std::min({s1.x, s2.x, s3.x, s4.x});
+    auto y = std::min({s1.y, s2.y, s3.y, s4.y});
+    auto w = std::max({s1.x, s2.x, s3.x, s4.x}) - x;
+    auto h = std::max({s1.y, s2.y, s3.y, s4.y}) - y;
```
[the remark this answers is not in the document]
Febbe: good catch!

src/util/Matrix.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+auto constexpr inverse(Matrix const& self) -> Matrix {
+    auto det = self.a * self.d - self.b * self.c;  // 3
+    assert(det != 0);
+    return {self.d / det,
+            (-self.b) / det,
+            (-self.c) / det,
+            self.a / det,
+            (self.c * self.ty - self.d * self.tx) / det,
+            (self.b * self.tx - self.a * self.ty) / det};
```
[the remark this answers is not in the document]
Febbe: The benchmark is quite outdated. 
When I run the benchmark, the division approach is sometimes faster and sometimes slower, bot mostly they have the same runtime.

Seems like compilers and CPU's improved over time.

src/util/include/util/Matrix.h
```diff
@@ -0,0 +1,42 @@
...
+
+    [[nodiscard]] auto rotate(double angle) const -> Matrix;
+    [[nodiscard]] auto scale(double sx, double sy) const -> Matrix;
+    [[nodiscard]] auto translate(double dx, double dy) const -> Matrix;
+    [[nodiscard]] auto inverse() const -> Matrix;
+
+    [[nodiscard]] auto operator*(Matrix const& other) const -> Matrix;
+    [[nodiscard]] auto operator*(Point<double> const&) const -> Point<double>;
+    [[nodiscard]] auto operator+(Matrix const& other) const -> Matrix;
+
```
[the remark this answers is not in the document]
Febbe: The + operator actually has no value for affine transformations and is not even representable (m1_33 + m2_33 = 2) but is asserted to be 1, I will remove it therefore.

src/util/include/util/detail/MatrixMultiply.h
```diff
@@ -0,0 +1,41 @@
...
+inline namespace detail {
+
+struct Size {
+    double width;
+    double height;
+};
+
+constexpr auto multiply(const Matrix& matrix, const Size& size) -> Size {
+    return {matrix.a * size.width + matrix.b * size.height, matrix.c * size.width + matrix.d * size.height};
+}
```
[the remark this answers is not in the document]
Febbe: It was an idea of adding a Size Struct to make a difference between Points.

Febbe:
> There is one thing that is not so clear to me. You want every Element to store its transformation matrix, correct? From a drawing perspective, is sounds good, but there are other bits for which things look more complicated: e.g. determining if a Stroke is within a "Region selection" (or even a "Rectangle selection" for that matter, e.g. after a rotation).
> 
> Also, for rendering purposes, it matters that we can easily access the bounding box of a Stroke. If a rotation is applied, we can rotate the original bounding box and take the smallest rectangle containing it (so your Matrix::multiply(Rectangle) -> Rectangle), but this is typically way larger than the actual bounding box, thus leading to bigger areas being rerendered.
> 
> The same question applies to snapping and the computation of the "snappingBounds".

I also stumbled over that. And we should discuss this. However, for the other elements, where the inner representation does not change, it is better to use a transformation matrix (Tex, Text, Pictures, PDF) it is better to use the transformation matrix as I proposed it.

For strokes, we have the problem, that the line width is messed up, as soon the x and y scaling is distinct. Even it behaves exactly like it should from the drawing perspective. But the user might only want to transform the points, not the linewidth.

Regarding snapping, I would not return a rectangle anymore:
For points, the first and last, maybe center points are more interesting
For rectangular Shapes like pictures/PDF all the edges might be better to snap to.
Therefore, I would return a `SmallVector` of Points, rather a Rectangle.

Regarding the bounds to update the drawing, I don't think that the impact is that large, the maximum factor of the resulting Rectangle is exactly 2. Always looping over hundreds of points might be worse.
There is also a way to retrieve the area with the `cairo_*_extents` function, like `cairo_stroke_extents`.
I probably should benchmark this.

test/bench/matrix/MatrixConst.h
```diff
@@ -0,0 +1,34 @@
...
+auto constexpr inverse(Matrix const& self) -> Matrix {
+    auto det = self.a * self.d - self.b * self.c;  // 3
+    assert(det != 0);
+    return {self.d / det,
+            (-self.b) / det,
+            (-self.c) / det,
+            self.a / det,
+            (self.c * self.ty - self.d * self.tx) / det,
+            (self.b * self.tx - self.a * self.ty) / det};
+}
```
Febbe, line 34:
```suggestion
}

```

test/bench/matrix/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  OPTIONS "BENCHMARK_ENABLE_TESTING Off"
+)
+
+if(benchmark_ADDED)
+  # enable c++11 to avoid compilation errors
+  set_target_properties(benchmark PROPERTIES CXX_STANDARD 11)
+endif()
+
+# link dependencies
+target_link_libraries(bench_matrix benchmark::benchmark)
```
Febbe, line 31:
```suggestion
target_link_libraries(bench_matrix benchmark::benchmark)

```

src/util/include/util/matrix/PointMultiply.h
```diff
@@ -0,0 +1,12 @@
...
+#include "util/Matrix.h"
+#include "util/Point.h"
+
+namespace xoj::util {
+
+[[nodiscard]] constexpr auto operator*(Matrix const& lhs, Point<double> const& pt) -> Point<double> {
+    return {lhs.a * pt.x + lhs.b * pt.y + lhs.tx, lhs.c * pt.x + lhs.d * pt.y + lhs.ty};
+}
+
+}  // namespace xoj::util
```
Febbe, line 12:
```suggestion
}  // namespace xoj::util

```

test/bench/matrix/MatrixV1.cpp
```diff
@@ -0,0 +1,45 @@
...
+auto Matrix::operator+(Matrix const& other) const -> Matrix {
+    return {a + other.a, b + other.b, c + other.c, d + other.d, tx + other.tx, ty + other.ty};
+}
+
+static_assert(multiply(test_ident, test_ident) == test_ident);
+static_assert(multiply(test_ident, test_trans) == test_trans);
+static_assert(multiply(test_trans, test_ident) == test_trans);
+static_assert(multiply(inverse(test_rot), multiply(test_rot, test_all)) == test_all);
+
+}  // namespace xoj::util::v1
```
Febbe, line 45:
```suggestion
}  // namespace xoj::util::v1

```

test/bench/matrix/MatrixV1.h
```diff
@@ -0,0 +1,36 @@
...
+
+auto operator*(double scalar, const Matrix& matrix) -> Matrix;
+
+constexpr auto test_ident = Matrix{1, 0, 0, 1, 0, 0};
+constexpr auto test_trans = Matrix{1, 0, 0, 1, 10, 10};
+constexpr auto test_scale = Matrix{2, 0, 0, 2, 0, 0};
+constexpr auto test_rot = Matrix{0, -1, 1, 0, 0, 0};
+constexpr auto test_all = Matrix{2, -1, 1, 2, 10, 10};
+
+} // namespace xoj::util::v1
```
Febbe, line 36:
```suggestion
} // namespace xoj::util::v1

```

test/bench/matrix/MatrixV2.h
```diff
@@ -0,0 +1,61 @@
...
+constexpr auto test_scale = Matrix{2, 0, 0, 2, 0, 0};
+constexpr auto test_rot = Matrix{0, -1, 1, 0, 0, 0};
+constexpr auto test_all = Matrix{2, -1, 1, 2, 10, 10};
+
+static_assert(multiply(test_ident, test_ident) == test_ident);
+static_assert(multiply(test_ident, test_trans) == test_trans);
+static_assert(multiply(test_trans, test_ident) == test_trans);
+static_assert(multiply(inverse(test_rot), multiply(test_rot, test_all)) == test_all);
+
+}  // namespace xoj::util::v2
```
Febbe, line 61:
```suggestion
}  // namespace xoj::util::v2

```

test/bench/matrix/cmake/cpm.cmake
```diff
@@ -0,0 +1,24 @@
...
+
+# Expand relative path. This is important if the provided path contains a tilde (~)
+get_filename_component(CPM_DOWNLOAD_LOCATION ${CPM_DOWNLOAD_LOCATION} ABSOLUTE)
+
+file(DOWNLOAD
+     https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
+     ${CPM_DOWNLOAD_LOCATION} EXPECTED_HASH SHA256=${CPM_HASH_SUM}
+)
+
+include(${CPM_DOWNLOAD_LOCATION})
```
Febbe, line 24:
```suggestion
include(${CPM_DOWNLOAD_LOCATION})

```

test/bench/matrix/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+  for (auto _ : state) {
+    auto test = trans * trans;
+    auto inv = test.inverse();
+    auto res = inv * test;
+    benchmark::DoNotOptimize(res);
+  }
+}
+BENCHMARK(v2);
+
+BENCHMARK_MAIN();
```
Febbe, line 36:
```suggestion
BENCHMARK_MAIN();

```

bhennion:
Closing, replaced by #7712

PR #5602, opened 2024-04-04
Refactor OpenFile logic - make it gtk4-compatible

bhennion:
This PR builds on #5585 and refactors Control::openFile and similar functions so that the dialogs they call become non-blocking and compatible with gtk4.

The logic is pretty complicated here. I tried to make it as easy to read and maintain as possible.
There are a couple of notable things to look at when reviewing:

1. I tried to homogenize the various file opening paths: e.g. the file version is now check even when the user selects a pdf-file, has the isAutoloadPdfXoj() setting to true and the associated .xopp file has an unsupported file version. This may have introduced unwanted behaviour changes.
2. Testing every scenario (e.g. `Annotate PDF` while having an unsaved document, or opening a file with missing PDF background,, or a combination of the two, and so on). I tried to test them all, but there are many and I could have missed one.
3. Implementation of xoj::util::move_only_function (similar to cpp23's std::move_only_function), used so that the callbacks can own an instance of std::unique_ptr<Document>, corresponding to the document being open. (For instance, if the file version is more recent than supported by this Xournal++ version, the document is already parsed, but we need to ask the user if they really want this document opened or not)
4. A question: if a file has not been saved yet and the user clicks on `Open File`, which dialog should pop up first: the FileChooser or the Save or discard dialog? I changed this behaviour compared to master (master shows the FileChooser first, this implements the other way around). Any opinion is welcome.

I also fixed a couple of inocuous bugs (e.g. if a file is not saved and the user tries to open a PDF file, the "Save or Discard" dialog appears twice on master).

~~For reviewers, only look the last commit, the other ones are from #5585 and will be rebased away once #5585 is merged.~~ Edit: #5585 is merged and this has been rebased

There are still old gtk3-still dialogs in the code base. ~~The next step is to refactor Control::save() and Control::saveAs(), in a follow up PR.~~ Edit: this is now part of this PR

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -54,6 +54,7 @@
 #include "gui/dialog/SettingsDialog.h"                           // for Sett...
 #include "gui/dialog/ToolbarManageDialog.h"                      // for Tool...
 #include "gui/dialog/XojOpenDlg.h"                               // for XojO...
+#include "gui/dialog/XojSaveDlg.h"                               // for XojS...
 #include "gui/dialog/toolbarCustomize/ToolbarDragDropHandler.h"  // for Tool...
 #include "gui/inputdevices/CompassInputHandler.h"                // for Comp...
 #include "gui/inputdevices/GeometryToolInputHandler.h"           // for Geom...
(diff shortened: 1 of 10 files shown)


rolandlo (approved):
I did some extensive testing (mostly on Linux, but also on Windows and MacOS) and haven't found any regressions. Good job! Two things I noticed (which are also present on master):

1) `emergencysave.xopp` is added to the list of recently opened documents, which doesn't make much sense.
2) When a file in the "new" format with PDF background like `xournalpp/test/files/packaged_xopp/pdfBackground/new.xopp` is emergency-saved, the PDF background cannot be restored via the emergencysave.xopp file.

From the code side I have only some minor remarks. Maybe @Febbe can take a look at `move_only_function.h`, since that takes a lot of C++ knowledge.

src/util/include/util/move_only_function.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * Xournal++
+ *
+ * Xournal util functions
+ *
+ * @author Xournal++ Team
+ * https://github.com/xournalpp/xournalpp
```
eldipa, line 7: My 2 cents:

Is this implementation a ported version of the C++23 [move_only_function](https://en.cppreference.com/w/cpp/utility/functional/move_only_function) ?
If that is the case, are any differences between this and C++23 impl?

I would like to see a documentation of what it does and how to use it plus a link to C++23 [move_only_function](https://en.cppreference.com/w/cpp/utility/functional/move_only_function) (if applies).
I think it will increase the chances that a future developer use this kind of helpers.

eldipa: Thanks!

Febbe:
> Unless @Febbe objects I suggest to move forward with this PR and merge it.

I currently have no time to do a full review. I second that. We can still push a fix to correct unwanted behavior.

src/util/include/util/move_only_function.h
```diff
@@ -0,0 +1,69 @@
...
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace xoj::util {
+template <class T>
+class move_only_function final {};
```
Febbe, line 20: Should be only a class template prototype (remove `{}`). Since an invalid specialisation should result in a compile error directly.

Issue #5587, opened 2024-03-25
jack server is not running or cannot be started


dbabits:
### Operating System

Linux

### (Linux only) Distribution

WSL2/Ubuntu

### (Linux only) Desktop Environment

X

### (Linux Only) Display Server

X.Org

### Installation Method

./xournalpp-1.2.3-x86_64.AppImage

### Xournal++ Version

1.2.3

### libgtk Version

libgtk: 3.24.20    

### Bug Description

```
lsb_release -a
Distributor ID: Ubuntu                                                                                                                                                                                                      Description:    Ubuntu 20.04.6 LTS                                                                                                                                                                                          Release:        20.04                                                                                                                                                                                                       Codename:       focal       

xxx@LG-Gram:~/bin/Xournal++$ ./xournalpp-1.2.3-x86_64.AppImage
xopp-Message: 20:12:27.876: TEXTDOMAINDIR = /tmp/.mount_xourna4E1F4j/usr/share/locale:, Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /tmp/.mount_xourna4E1F4j/usr/share/locale

(com.github.xournalpp.xournalpp:22153): xopp-WARNING **: 20:12:27.890: Settings file /home/dima/.config/xournalpp/settings.xml does not exist. Regenerating.
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1246:(snd_func_refer) error evaluating name
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5220:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM sysdefault
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
[...]

PR #5603, opened 2024-04-05
Add uim-data to AppImage.

### LINKED ISSUE #5493: IME on Linux don't work
### Operating System

Windows

### (Linux only) Distribution

Manjaro

MATE

None

### Installation Method

Appimage

### Xournal++ Version

1.2.2

### libgtk Version

3.24.20

### Bug Description

IME like fcitx don't work.

### Expected Behaviour

It works.

### Steps to Reproduce

1. Press the shortcut of activation “Ctrl+Space” for fcitx.

### Additional Context

_No response_

tattsan:
May fix #5493 and partially #5587.
This will install many `scheme` libraries for `uim` in the AppImage, and I don't think it appropriate.

### DIFF
--- a/azure-pipelines/steps/install_deps_ubuntu.yml
+++ b/azure-pipelines/steps/install_deps_ubuntu.yml
@@ -10,7 +10,7 @@ steps:
       sudo apt-get update
       sudo apt-get install -y librsvg2-dev gcc-${{ parameters.gcc_version }} g++-${{ parameters.gcc_version }} cmake ninja-build \
         libgtk-3-dev libpoppler-glib-dev portaudio19-dev libsndfile-dev \
-        ibus-gtk3 fcitx-frontend-gtk3 fcitx5-frontend-gtk3 uim-gtk3-immodule gcin-gtk3-immodule \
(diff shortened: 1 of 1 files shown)


tattsan:
The ubuntu package `uim-data` contains '/usr/share/uim/lib/sigscheme-init.scm',  but 'xournalpp-1.2.4-x86_64.AppImage' doesn't contain the file. What's wrong?

rolandlo:
Well, looking at it, how should `linuxdeploy` know that it needs to package the `uim-data` files? It just takes the xournalpp tarball, its dependencies (that can be seen from `ldd /usr/bin/xournalpp`), desktop files and icons and the Gtk related stuff (since we use the  `linuxdeploy-plugin-gtk` Plugin).

tattsan:
I see, so it doesn't take in all the files installed with install_deps_ubuntu.yml. Originally it was not a good idea to include the uim scheme code in the Xournal++ appimage, it was just an experiment to see if it would prevent crashes.

As https://github.com/xournalpp/xournalpp/issues/5493#issuecomment-2038569512  states, using /usr/lib/gtk-3.0/3.0.0/immodules/im-uim.so in the user's environment instead of in appimage would solve this problem (it should be installed in the user's environment for uim users), but it is not possible to specify the path.

To prevent crashes, it would be better to start with GTK_IM_MODULE=ibus if the environment variable GTK_IM_MODULE is  empty.

PR #5605, opened 2024-04-05
Fix incorrect z-position upon undoing delete selection

### LINKED ISSUE #5592: Selecting something, deleting it then undoing does not restore correct Z-position
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Official release

### Xournal++ Version

1.2.3

### libgtk Version

3.24.41

### Bug Description

When selecting something, it goes on the front in the Z-plane. So when deleting it while selected, then if undo is done, the Z-position will not be correctly restored.
It will restore the z-position from when it was selected, so in the foreground.

### Expected Behaviour

Instead it should restore the original z-position of the element.

### Steps to Reproduce

Draw 2 perpendiculars lines
Select the one behind the other.
Delete then undo
It will not be in front of the other line.

### Additional Context

_No response_

tmoerschell:
Fixes #5592 
The hard work had already been done by @Febbe

### DIFF
--- a/src/core/control/tools/EditSelectionContents.cpp
+++ b/src/core/control/tools/EditSelectionContents.cpp
@@ -323,13 +323,8 @@ auto EditSelectionContents::setColor(Color color) -> UndoActionPtr {
 void EditSelectionContents::fillUndoItem(DeleteUndoAction* undo) {
     Layer* layer = this->sourceLayer;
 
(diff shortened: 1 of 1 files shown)


Febbe:
It was @bhennion 😅🫣. I just ensured memory safety.

PR #5610, opened 2024-04-08

bhennion:
The only conflicts were around the changes of #5583. I think I got it right but better safe than sorry.

@febbe @rolandlo @c-ola could you please check that your changes were merged as expected?

### DIFF
--- a/plugins/MigrateFontSizes/main.lua
+++ b/plugins/MigrateFontSizes/main.lua
@@ -3,7 +3,8 @@ function initUi()
   app.registerUi({["menu"] = "Migrate font sizes with factor displayDPI / 72", ["callback"] = "migrate"});
   app.registerUi({["menu"] = "Show font size migration dialog", ["callback"] = "showDialog"});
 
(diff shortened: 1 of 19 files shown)


Febbe:
My changes look good

c-ola:
My changes look good as well.

rolandlo (commented):
My changes have been merged as expected, but:

- Can the warnings https://github.com/xournalpp/xournalpp/pull/5599#issuecomment-2037624420 be fixed before merging back? They are really annoying.
- In #5583 we added quite some debug messages for console output. Do we really want to keep them?
- In #5583 I found one case for MacOS where there is an inconsistency between icons and background: When `GTK_THEME` is set to a light theme, but MacOS uses dark theme. Then the background is light and the icons are dark. The wrong icons are used. There is an easy fix, so would you allow me to push it before merging back? Sorry for not noticing earlier.

rolandlo:
> @rolandlo The warnings from #5599 are already on master (this PR was merged to master). This has nothing to do with merging the release branch.

Ok, I see. The pipeline is failing for some (unrelated) reason though.

PR #5612, opened 2024-04-11
Fix theme variant on macos when GTK_THEME is set and remove debug messages

rolandlo:
I suggest to merge this PR before merging release-1.2 back to master.

(The warnings introduced in #5599 should be fixed as well before merging back).


rolandlo:
> I'd rather we made the g_messages into g_debug, rather than removing them. There is essentially no performance cost in keeping them, and then we can ask a complaining user to run with debug output to get info easily.

Done. I did the same for the locale dir. Also I used `[[maybe unused]]` for `useEnv`, since that variable is only used on MacOS and should not create compile warnings on Linux and Windows.

Issue #139, opened 2015-08-12
Word Wrap for the Text Tool


rhnvrm:
There should be an option for Word Wrapping the text in the Text tool as it is often required when pasting from wikipedia etc.

wbrenna:
This should be possible, though I recall looking at it and it required a bit more time with some potential pitfalls. Pango word wrapping can be done with something like [this](https://gist.github.com/bert/262331), though it might be confusing since I suspected the gtk_text_buffer might not wrap text during editing - it would wrap only when text is not selected.

andreasb242:
Xournal did not support Text wrapping, so I didn’t implement it. But It’s a feature which makes really sense.

andreasb242:
We will brake compatibility, and the usage is not that high, I close this ticket (cleanup of old tickets...)

goekce:
https://github.com/xournalpp/xournalpp/issues/1756

Issue #137, opened 2015-05-30
SVG for Highlighter


wbrenna:
I'm getting an error ```Couldn't find pixmap file: tool_highlighter.png```. Probably the old PNG should be put back in unless someone wants to make an SVG for that icon.

MarPiRK:
I think, that we need icon redesign - every icon is in different style. Also there are some standard theme icons, such as document-new, user-bookmark, insert-text, gtk-fullscreen, image, emblem-default and so on, which can be used instead of self made.

wbrenna:
I think this is still present in the most recent pull request - did you find a highlighter in gtk standards?

MarPiRK:
No, it's just my opinion about our current icons. I've told with my sister (she's graphic design  student) and she agreed she may redesign tool icons.

andreasb242:
The icons are now from different sources, some are SVG and self made, some pixel graphics.
I would prefer to have all icons from the same source, and as SVG. But I cannot do it ;-)

MarPiRK:
For now my sister has too much of her own work, so it's not possible to her to make new icons.
I think that for now we should revert old icon.
(I'll do it in a moment)

MarPiRK:
Can be closed

Issue #59, opened 2014-01-13
Element.cpp Crash on Quit


wbrenna:
Small document with some written text and TeX, along with a pasted image. Reopening does not reproduce the bug.

```
** (xournalpp:18255): WARNING **: XojTypeCheck failed: expected Element but get something else on model/Element.cpp:26
```

and

```
Try to get a better stracktrace...
[bt] #1 /home/wilson/documents/Dropbox/misc/xournalpp/xournalpp_fedora/src/xournalpp() [0x826a900]
/home/wilson/documents/Dropbox/misc/xournalpp/xournalpp_fedora/src/util/CrashHandler.cpp:124
[bt] #2 [0xb778b400]
??:0
```

PR #5615, opened 2024-04-12
GTK4 compiling branch

### LINKED ISSUE #2722: GTK4 preparation

bhennion:
~~It doesn't run yet, but we're closing in, I think.~~

This branch compiles (and links) and starts ~~but the .glade files need to be adapted before testing anything.~~ edit: and works

There are a couple of features I deactivated/commented out to get it to compile, so those need fixing either in gtk4, or before that, in gtk3 (e.g. the sidebar could be pre-ported).

- [ ] Optimize cursor creation (cairo->pixbuf->texture->cursor)
- [x] Restore gtk-sourceview (5 is gtk4)
- [ ] Zoom
    - [x] Scrollzoom
    - [x] Touch pinch zoom
    - [ ] Touchpad pinch zoom
    - [ ] Long term: refactor -> new Navigation = ScrollHandler + ZoomControl + unify event handling
- [x] Toolbar Customization
    - [x] Create suitable GtkWidget for toolbars
    - [x] Port ToolbarCustomizeDialog
    - [x] Port drag'n'drop customization
- [x] Colorize svg for toolbar color icons (so they scale)
- [x] Sidebar
    - [x] Port Sidebar
    - [x] Restore page numbers
- [x] Clipboard
- [x] ImageDialog + add PdfPage dialog
- [x] Landscape/portrait icons in FormatDialog
- [x] Fix `Gdk-CRITICAL **: 21:20:45.390: gdk_key_event_get_keyval: assertion 'GDK_IS_EVENT_TYPE (event, GDK_KEY_PRESS) || GDK_IS_EVENT_TYPE (event, GDK_KEY_RELEASE)'` errors
- [x] Plugins
    - [x] Refactor API for dialogs. Add callbacks
    - [x] Port plugin dialog
    - [x] Expose new action system. Retire old one.
- [ ] Replace GdkPixbuf by GdkTexture?
- [ ] Fix/report to GDK issues with x11 backend
    - [ ] Mouse/Touchpad event report a pressure value (always 0) -> reported [upstream](https://gitlab.gnome.org/GNOME/gtk/-/issues/7359)
    - [ ] Touchpad tap do not bring up floating toolbox (but using the physical buttons works) https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2290932348 (same [upstream issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/7359))
    - [ ] Zoom gestures on touchscreen draw weird strokes instead of zooming in/out https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2290899864 -- probably related: touch input always draws, even if touch drawing is disabled (same [upstream issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/7359))
    - [ ] Freeze when copy-pasting across two instances https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2247648579
- [x] Double and triple click
- [x] Fetch event history for stylus events? (see gdk_event_get_history())
- [ ] Refactor device tool classification https://github.com/xournalpp/xournalpp/pull/5615#discussion_r1646561232
- [ ] Fix double/float conversion due to changes in GdkColor (4 x double became 4 x float)
- [ ] Fix "Use Stock Icon" option - see IconNameHelper.cpp
- [x] Port or remove gui/inputdevices/touchdisable/TouchDisableX11.cpp (cf gui/inputdevices/HandRecognition.cpp:L74)
- [x] Remove code duplication: gui/toolbarMenubar/ToolMenuHandler.cpp:createItem() createIcon() load()
- [x] Fix accelerators priority (e.g. `Ctrl + A` when editing text selects the elements on the page, not the content of the current text box)
- [ ] Fix MacOS copy/paste across instances pastes as image [see this post](https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2250277053)
    Edit: Probably fixed upstream via [this MR](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7850) -> TO BE TESTED
- [ ] Opening Help URL fails on MacOS (at least ARM) see [here](https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2250662801) and [here](https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2250705228)
- [ ] Fix cannot insert image on Windows [see this post](https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2252404305)
- [x] Fix Icon while dragging in toolbar customization. see https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2290966618
- [ ] Report and/or workaround GTK drag-n-drop bugs on MacOS/Windows (see https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2266703168 for some bug reports)
    [Report upstream](https://gitlab.gnome.org/GNOME/gtk/-/issues/6906) 
    Edit: Supposedly fixed [upstream](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7850) on MacOS -> TO BE TESTED
- [x] Port MacOS ARM cirrus CI

~~NB: this branch is built on a merge of various PR's. Not really a pull request...~~ Not anymore


rolandlo:
Oh, wow, that's awesome! Great job @bhennion ! I'm really excited! 

grafik

rolandlo:
Can you please replace `mingw-w64-x86_64-gtk3` by  `mingw-w64-x86_64-gtk4` in `build_windows.yml`? Then the Windows pipeline should succeed. At least  it works for me locally. Drawing and typing work as well.

I will try to prepare a Gtk4 blob on MacOS so that MacOS can be tested as well. It will take some time though.

rolandlo:
It compiles on Windows with `-DENABLE_PLUGINS=off`, I need to add.

rolandlo:
I got it [b9c6a68](https://github.com/xournalpp/xournalpp/pull/5615/commits/b9c6a689c8e0ab5abaeb4131835d77b3ce7ecca6) to compile on MacOS as well. I have built Gtk4 and the other dependencies with jhbuild and I'm ready to create a Gtk4 blob. Two questions.

- I haven't built GtkSourceView yet. Do we need it at this point? If so, which version?
- Where should I put the blob? Shall I create a gtk4 directory for it in https://github.com/xournalpp/xournalpp-pipeline-dependencies/tree/macos-10.15-jhbuild or should I create a different branch in the xournalpp-pipeline-dependencies repo?

By the way, the last two commits don't compile (as also seen from the Pipeline).

rolandlo:
> > * I haven't built GtkSourceView yet. Do we need it at this point? If so, which version?
> 
> From what I could gather, gtksourceview 4 runs with gtk 3, but gtksourceview 5 runs gtk 4... That's awkward but that what they say in their README: https://gitlab.gnome.org/GNOME/gtksourceview/-/tree/gtksourceview-4-8?ref_type=heads.
> 
> We don't exactly need it yet, but at some point, someone will have to port its integration in our LatexDialog.
> 
> > * Where should I put the blob? Shall I create a gtk4 directory for it in https://github.com/xournalpp/xournalpp-pipeline-dependencies/tree/macos-10.15-jhbuild or should I create a different branch in the xournalpp-pipeline-dependencies repo?
> 
> I'd say in a new branch.
> 
> > By the way, the last two commits don't compile (as also seen from the Pipeline).
> 
> Yes, I pushed a typo... I'll fix it tomorrow.

The Gtk4 blob is now in the branch https://github.com/xournalpp/xournalpp-pipeline-dependencies/tree/macos-10.15-gtk4
I hope it works. It includes gtksourceview-5 for future use.

Please replace `macos-10.15-jhbuild` by `macos-10.15-gtk4`
in 
https://github.com/xournalpp/xournalpp/blob/ef5419038c18e13c22ac1fd4aeb6062887f64feb/azure-pipelines/steps/build_mac.yml#L15-L16

rolandlo:
> > The Gtk4 blob is now in the branch https://github.com/xournalpp/xournalpp-pipeline-dependencies/tree/macos-10.15-gtk4 I hope it works. It includes gtksourceview-5 for future use.
> 
> Thanks for the blobs!
> 
> > Please replace `macos-10.15-jhbuild` by `macos-10.15-gtk4`
> 
> I did that, but something is still wrong (cf CI error messages). It seems something is not found, leading to a opts-NOTFOUND as a parameter.

I'll try to figure it out. I get it locally as well. Reconfiguring via `cmake -DCMAKE_INSTALL_PREFIX=$HOME/gtk/inst -DENABLE_PLUGINS=off -GNinja ..` and running `ninja` again solves the problem for me. I don't understand why.

rolandlo:
> > I did that, but something is still wrong (cf CI error messages). It seems something is not found, leading to a opts-NOTFOUND as a parameter.
> 
> I'll try to figure it out. I get it locally as well. Reconfiguring via `cmake -DCMAKE_INSTALL_PREFIX=$HOME/gtk/inst -DENABLE_PLUGINS=off -GNinja ..` and running `ninja` again solves the problem for me. I don't understand why.

Actually running the `cmake ..` step twice before building suffices.

rolandlo:
The first time configuringI find the following line in `build.ninja`:

```
  LINK_FLAGS = -L/Users/xournal-dev/gtk/inst/lib -L/Users/xournal-dev/gtk/inst/lib -arch x86_64 -L/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.15 -Wl,-headerpad_max_install_names -L/Users/xournal-dev/.new_local/share/pyenv/versions/3.11.7/lib    opts-NOTFOUND
```

The second time it is
```
  LINK_FLAGS = -L/Users/xournal-dev/gtk/inst/lib -L/Users/xournal-dev/gtk/inst/lib -arch x86_64 -L/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.15 -Wl,-headerpad_max_install_names -L/Users/xournal-dev/.new_local/share/pyenv/versions/3.11.7/lib    -framework CoreAudio -framework AudioToolbox -framework AudioUnit -framework CoreFoundation -framework CoreServices
```

rolandlo:
> The first time configuringI find the following line in `build.ninja`:
> 
> ```
>   LINK_FLAGS = -L/Users/xournal-dev/gtk/inst/lib -L/Users/xournal-dev/gtk/inst/lib -arch x86_64 -L/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.15 -Wl,-headerpad_max_install_names -L/Users/xournal-dev/.new_local/share/pyenv/versions/3.11.7/lib    opts-NOTFOUND
> ```
> 
> The second time it is
> 
> ```
>   LINK_FLAGS = -L/Users/xournal-dev/gtk/inst/lib -L/Users/xournal-dev/gtk/inst/lib -arch x86_64 -L/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.15 -Wl,-headerpad_max_install_names -L/Users/xournal-dev/.new_local/share/pyenv/versions/3.11.7/lib    -framework CoreAudio -framework AudioToolbox -framework AudioUnit -framework CoreFoundation -framework CoreServices
> ```

I found a solution to the linking problem on MacOS: Simply remove the lines 
https://github.com/xournalpp/xournalpp/blob/d5164fc5dc2153500182ae2bb3b0edee1b09c2f0/CMakeLists.txt#L124-L139

By the way, in
https://github.com/xournalpp/xournalpp/blob/d5164fc5dc2153500182ae2bb3b0edee1b09c2f0/src/core/control/XournalMain.cpp#L496-L496 
`about.glade` should be `about.ui` now.

rolandlo:
Great work @bhennion! It's a giant step forward! 
Sure, cleaning up the commit history and making a dedicated branch are good ideas. Probably I'd like to work on some of the remaining issues as well in the coming two weeks (e.g. the plugins), so I can then make a PR to the dedicated branch.

We should also review your previous PR's that went into this PR, so that they can be merged into master.

Moreover I think it's time for some testing of the features that are supposed to work. I will share my observations in this PR.

rolandlo:
Some observations (on Linux):
- Typing in the text tool I get warnings: `xopp-WARNING **: 21:06:03.497: InputEvents::translateEvent() but GdkEvent has no position`
- Maximizing the window does not work (neither by pressing the window maximize button nor by double clicking on the window
- Resizing the window also doesn't work properly. 
- On startup I get warnings:
```shell
Gtk-WARNING **: 21:10:16.209: Theme parser error: xournalpp.css:152:3-15: No property named "margin-start"

(com.github.xournalpp.xournalpp:90944): Gtk-WARNING **: 21:10:16.209: Theme parser error: xournalpp.css:153:3-13: No property named "margin-end"
```
- I got some critical warnings 
```shell
(com.github.xournalpp.xournalpp:93399): Gdk-CRITICAL **: 21:20:45.390: gdk_key_event_get_keyval: assertion 'GDK_IS_EVENT_TYPE (event, GDK_KEY_PRESS) || GDK_IS_EVENT_TYPE (event, GDK_KEY_RELEASE)' failed
```
and
```shell
(com.github.xournalpp.xournalpp:93812): GLib-GObject-CRITICAL **: 21:28:18.463: ../../../gobject/gsignal.c:2685: instance '0x6041d4b66f30' has no handler with id '375'
```
- When drawing the stroke is displaced from the cursor (except in fullscreen mode where it's fine).
- Resizing and rotating selections with the handles doesn't work as expected (because of the cursor displacement).
- Inserting an image crashes with signal 4 (and warning: implement ImageOpenDlg)
- The icons for landscape and portrait paper format seem to be missing
- In the sidebar layer previews are in opposite order
- Snapping doesn't seem to work (no effect)

rolandlo:
> That's interesting. Could you try putting a - instead of a + in util/GtkUtil.cpp xoj::util::gtk::gdkSurfaceToWidgetCoordinates() (see the comment therein?

Yeah, that fixes the displacement (and other things as well; more tests needed). When I have a toolbar (I used the Xournal++ toolbar), `q` has the coordinates: `q = (14.000000, 12.000000)`.

rolandlo:
One more issue I observed: When drawing with the mouse (and having pressure multiplier 2.1, minimal pressure 0.1), the stroke width looks tiny.

Snapping still doesn't work with correct coordinates.

rolandlo:
> > * In the sidebar layer previews are in opposite order
> 
> Are they? It should be upper layer first, background last. You have something else?

I take that back. The order is as you say and I mistakenly thought it was wrong.

rolandlo:
> One more issue I observed: When drawing with the mouse (and having pressure multiplier 2.1, minimal pressure 0.1), the stroke width looks tiny.
> 
> Snapping still doesn't work with correct coordinates.

The pressure issue observed above only happens with  `GDK_BACKEND=x11` (which is automatically set when I run it from VS Code). 

The resizing issue is gone.

rolandlo:
Please remove the snapping issue. It works fine now. Seems I didn't test with correct settings before.

rolandlo:
> > * The icons for landscape and portrait paper format seem to be missing
> 
> That's curious. Those are stock icons.
> 
> > *

They have been deprecated since Gtk 3.10:

https://developer-old.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-ORIENTATION-PORTRAIT:CAPS

I don't see good alternatives in Adwaita, but we could add these icons: 
grafik

which I found in Workbench and Gnome Builder.

rolandlo:
For building the app bundle on MacOS `gtk-mac-bundler` will have to add Gtk 4 support (and `xournalpp.modules` will have to be adapted slightly as well). I have raised an issue on the `gtk-mac-bundler` repo and would try to make it work myself if the developer is not interested in doing it himself.

rolandlo:
@LittleHuba From the logs it looks to me that the [Debian docker image](https://hub.docker.com/r/xournalpp/debian-latest-sudo/tags) you created two years ago is based on bullseye (which doesn't have Gtk 4) instead of bookworm (that has Gtk 4). Is that something you could update easily?

rolandlo:
@bhennion if you like you can cherrypick or use (some of) the most recent 4 commits from https://github.com/rolandlo/xournalpp/commits/gtksourceview-transition/
- GtkSourceView 5 reactivated (it works all fine on wayland; there is still an issue with `GDK_BACKEND=x11` when entering a `^`, which gets doubled)
- Build AppImage on Ubuntu 22.04 (building works and it runs, but unfortunately segfaults quickly; I'm investigating why)
- Fixup `xournalpp.bundle` for MacOS (that will fix building the app bundle in combination with the merge request I made to `gtk-mac-bundler`)

rolandlo:
The GtkSourceView 5 issue with GDK_BACKEND=x11 also appears in gnome-text-editor. It may only be reproducible with a German (Swiss) keyboard layout.

rolandlo:
The GtkSourceView 5 issue with GDK_BACKEND=x11 disappeared after uninstalling `ibus-gtk4` (version 1.5.29). It seems to have been fixed in version 1.5.30, see https://github.com/ibus/ibus/issues/2585

rolandlo:
For the record, there are 4 issues when converting the current .ui files to the .blp format (which would be desirable imho). They are all reported on the issue tracker https://gitlab.gnome.org/jwestman/blueprint-compiler/-/issues

about.ui: `unsupported feature` -> #59
main.ui: `unsupported feature` -> #59
mainmenubar.ui: `unsupported feature` -> #137
texdialog.ui: `unsupported feature` -> #59
settings.ui: `unsupported XML tag: <columns>, <items>` -> #150, #139
exportSettings.ui: `unsupported XML tag: <columns>` -> #150, #59
settingsButtonConfig.ui: `unsupported XML tag: <items>` -> #139
settingsDeviceClassConfig.ui: `unsupported XML tag: <items>` #139

LittleHuba:
> @LittleHuba From the logs it looks to me that the [Debian docker image](https://hub.docker.com/r/xournalpp/debian-latest-sudo/tags) you created two years ago is based on bullseye (which doesn't have Gtk 4) instead of bookworm (that has Gtk 4). Is that something you could update easily?

Sorry I don't have the exact steps in mind anymore. I can take a look over the weekend if that is enough for you?

rolandlo:
> > @LittleHuba From the logs it looks to me that the [Debian docker image](https://hub.docker.com/r/xournalpp/debian-latest-sudo/tags) you created two years ago is based on bullseye (which doesn't have Gtk 4) instead of bookworm (that has Gtk 4). Is that something you could update easily?
> 
> Sorry I don't have the exact steps in mind anymore. I can take a look over the weekend if that is enough for you?

Great, thanks! That's soon enough.

rolandlo:
FYI @bhennion: I'm working on the plugins resurrection in https://github.com/rolandlo/xournalpp/tree/fix-plugins
`app.getFilePath` already works. It takes the name of the callback function as a parameter and calls that function when the user has selected the filepath. The function has one parameter, which is the filepath (a string).

The sidebar plugin API had need to be fixed and the .glade-files were converted to Gtk4 compatible .ui files.
Now I hope I can get `app.saveAs` work in a similar fashion.

rolandlo:
> That's very good! One question: how much of it could be targetted for master?

If the sidebar changes get targetted for master, then I guess all of it except the glade to ui conversion and maybe the removal of app.msgbox can be targetted for master. I can look at it once things work with Gtk4. 

In general I think it's a good idea to keep the diff between master and the Gtk4 version small. Also giving plugin writers some time to port their plugins is a good idea.

LittleHuba:
> > > @LittleHuba From the logs it looks to me that the [Debian docker image](https://hub.docker.com/r/xournalpp/debian-latest-sudo/tags) you created two years ago is based on bullseye (which doesn't have Gtk 4) instead of bookworm (that has Gtk 4). Is that something you could update easily?
> > 
> > 
> > Sorry I don't have the exact steps in mind anymore. I can take a look over the weekend if that is enough for you?
> 
> Great, thanks! That's soon enough.

https://github.com/xournalpp/xournalpp/pull/5645
The new setup will keep the Debian image constantly updated by rebuilding it daily. The Dockerfile relies on `debian:latest` which would also update to a newer Debian version as soon as available. If you want to stick to a specific Debian version, I would recommend modification of the Dockerfile.

src/core/gui/FloatingToolbox.cpp
```diff
@@ -24,9 +24,13 @@ FloatingToolbox::FloatingToolbox(MainWindow* theMainWindow, GtkOverlay* overlay)
...
 
     gtk_overlay_add_overlay(overlay, this->floatingToolbox);
-    gtk_overlay_set_overlay_pass_through(overlay, this->floatingToolbox, true);
-    gtk_widget_add_events(this->floatingToolbox, GDK_LEAVE_NOTIFY_MASK);
-    g_signal_connect(this->floatingToolbox, "leave-notify-event", G_CALLBACK(handleLeaveFloatingToolbox), this);
+    // TODO needed?
+    // gtk_overlay_set_overlay_pass_through(overlay, this->floatingToolbox, true);
+
+    auto* ctrl = gtk_event_controller_motion_new();
+    g_signal_connect(ctrl, "leave", G_CALLBACK(handleLeaveFloatingToolbox), this);
```
Febbe, line 31: Use our check to avoid UB.

src/core/gui/FloatingToolbox.cpp
```diff
@@ -77,13 +81,15 @@ auto FloatingToolbox::floatingToolboxActivated() -> bool {
...
         const char* guiName = TOOLBAR_DEFINITIONS[index].guiName;
-        GtkToolbar* toolbar1 = GTK_TOOLBAR(this->mainWindow->get(guiName));
-        count += gtk_toolbar_get_n_items(toolbar1);
+        GtkWidget* tb = this->mainWindow->get(guiName);
+        if (gtk_widget_get_first_child(tb)) {
+            return true;
+        }
     }
 
     return count;
```
Febbe, line 95: Why was this change required?
Also, remove count and return false.

src/core/gui/FloatingToolbox.cpp
```diff
@@ -105,14 +112,15 @@ void FloatingToolbox::showForConfiguration() {
 
 
 void FloatingToolbox::show() {
+    // TODO really??
     gtk_widget_hide(this->floatingToolbox);  // force showing in new position
-    gtk_widget_show_all(this->floatingToolbox);
+    gtk_widget_show(this->floatingToolbox);
```
Febbe, line 117: Shouldn't be required anymore.

src/core/gui/FloatingToolbox.cpp
```diff
@@ -93,9 +99,10 @@ auto FloatingToolbox::countWidgets() -> int {
 void FloatingToolbox::showForConfiguration() {
     if (this->floatingToolboxActivated())  // Do not show if not being used - at least while experimental.
     {
-        GtkWidget* boxContents = this->mainWindow->get("boxContents");
-        gint wx = 0, wy = 0;
-        gtk_widget_translate_coordinates(boxContents, gtk_widget_get_toplevel(boxContents), 0, 0, &wx, &wy);
+        GtkWidget* w = this->mainWindow->get("panedMainContents");
+        double wx = 0, wy = 0;
+        gtk_widget_translate_coordinates(w, gtk_widget_get_ancestor(w, GTK_TYPE_WINDOW), 0, 0, &wx, &wy);
```
Febbe, line 104: Deprecated, use https://docs.gtk.org/gtk4/method.Widget.compute_point.html.

src/core/gui/FloatingToolbox.cpp
```diff
@@ -105,14 +112,15 @@ void FloatingToolbox::showForConfiguration() {
...
     gtk_widget_hide(this->floatingToolbox);  // force showing in new position
-    gtk_widget_show_all(this->floatingToolbox);
+    gtk_widget_show(this->floatingToolbox);
 
     if (this->floatingToolboxState != configuration) {
         gtk_widget_hide(this->mainWindow->get("labelFloatingToolbox"));
     }
 
-    if (this->floatingToolboxState == configuration || countWidgets() > 0) {
+    if (this->floatingToolboxState == configuration || hasWidgets() > 0) {
```
Febbe, line 123: All usages of hasWidgets should be reviewed (now a bool). So don't compare them with an int.

src/core/gui/MainWindow.cpp
```diff
@@ -115,14 +109,6 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
         control->setShowMenubar(control->getSettings()->isMenubarVisible());
     });
 
-    // Drag and Drop
```
Febbe, line 118: how is that handled now?

src/core/gui/SearchBar.cpp
```diff
@@ -23,33 +23,37 @@ SearchBar::SearchBar(Control* control): control(control) {
 
     GtkWidget* next = win->get("btSearchForward");
     GtkWidget* previous = win->get("btSearchBack");
-    g_signal_connect(next, "clicked", G_CALLBACK(+[](GtkButton* button, SearchBar* self) { self->searchNext(); }),
-                     this);
+    g_signal_connect(next, "clicked",
+                     G_CALLBACK(+[](GtkButton*, gpointer d) { static_cast<SearchBar*>(d)->searchNext(); }), this);
```
Febbe, line 27: Use our safe casts here, this is actually no plain callback. The same applies for the next ~20 lines.

src/core/gui/ToolbarDefinitions.h
```diff
@@ -12,33 +12,28 @@
...
+        {"tbRight1", "toolbarRight1"},   {"tbRight2", "toolbarRight2"},
+        {"tbBottom1", "toolbarBottom1"}, {"tbBottom2", "toolbarBottom2"},
+        {"tbFloat1", "toolbarFloat1"},  // define this index below as TBFloatFirst
+        {"tbFloat2", "toolbarFloat2"},   {"tbFloat3", "toolbarFloat3"},
+        {"tbFloat4", "toolbarFloat4"}  // define this index below as TBFloatLast
 };
 
 
-const static int TOOLBAR_DEFINITIONS_LEN = G_N_ELEMENTS(TOOLBAR_DEFINITIONS);
+constexpr size_t TOOLBAR_DEFINITIONS_LEN = G_N_ELEMENTS(TOOLBAR_DEFINITIONS);
```
Febbe, line 36: Maybe better to use a std::array here?

src/core/gui/XournalView.cpp
```diff
@@ -327,7 +332,8 @@ auto XournalView::onKeyReleaseEvent(const KeyEvent& event) -> bool {
 void XournalView::onRealized(GtkWidget* widget, XournalView* view) {
     // Disable event compression
     if (gtk_widget_get_realized(view->getWidget())) {
-        gdk_window_set_event_compression(gtk_widget_get_window(view->getWidget()), false);
+        g_warning("Find replacement for gdk_window_set_event_compression()");
+        // gdk_window_set_event_compression(gtk_widget_get_window(view->getWidget()), false);
```
Febbe, line 336: No need, is active by default and can't be turned off

> With GTK4 one needs to look at the event history between motion events to reconstruct the ones that where coalesced. In gtk4 event delivery rate is fixed to run by the frame clock (that is, 60hz). get_motion_history() (function name may have changed) will only return the motion coordinates that were accumulated during the handling of that frame.

Febbe (commented):
Reviewed 50% of the code so far.

src/core/gui/XournalView.cpp
Febbe: > Yes, but we'd want to turn it off... Or at least it is turned off on master.

Oh right, overlooked the false. But using gdk_event_get_history should do the trick.

rolandlo:
@bhennion 
Now I have time to do testing on Linux, MacOS and Windows as suggested in https://github.com/xournalpp/xournalpp/pull/5615#issuecomment-2184077354. Could you please rebase on master first though?

rolandlo:
The `SettingsDialogPaletteTabTest` fails because of
https://github.com/xournalpp/xournalpp/blob/3d3102c75f7946ed1319fc894580a3f4b76a8d57/test/gtk_tests/dialog/SettingsDialogPaletteTabTest.cpp#L23-L24

rolandlo:
When I type in the text editor, plugin keyboard accelerators take precedence (unlike in master), so we can't use single letter accelerators without modifiers any more as accelerators.

rolandlo:
On Linux: 
Copy&paste works fine in both directions for Inkscape, Krita and GIMP except from Xournal++ to GIMP.
Drag&drop works from nautilus. Inkscape, Krita and GIMP don't seem to support that.
Copy&paste text works, drag&drop text from gnome-text-editor works too.
Copy&paste between different instances of Xournal++ works too.

rolandlo:
Duplicating a selection via mouse right click does not work!

rolandlo:
I got a warning about the .css: 
```shell
(com.github.xournalpp.xournalpp:204957): Gtk-WARNING **: 12:26:24.255: Theme parser error: xournalpp.css:173:11-15: "dark" is not a valid color name.
```

rolandlo:
Copy&paste between two different instances makes one instance hang when run with the option `GDK_BACKEND=x11`

rolandlo:
While editing text the keybinding Ctrl+a selects the whole selection, not only the text. Ctrl+slash still works for me.

Ctrl+plus and Ctrl+minus don't just increase/decrease font size but the zoom.

Ctrl+Shift+Delete doesn't delete till the beginning of the line, but only only word. Same on master though.

rolandlo:
Touchpad zooming (when holding Ctrl) doesn't work.

rolandlo:
The PDF toolbox looks different (just one column instead of a larger rectangle), but it works fine (on Linux).

rolandlo:
As I remember @bhennion you told me how to drag&drop tools to and from the floating toolbox during toolbar customization, but I forgot. How do I permanently open the floating toolbox while moving tools?

rolandlo:
After trying to customize the tools from the floating toolbox I got some warnings:

grafik

rolandlo:
Even when I have audio deactivated I get a warning with empty audio folder:
```shell
(com.github.xournalpp.xournalpp:28458): xopp-WARNING **: 17:00:01.510: SettingsDialog::setAudioRecordingFolder(): provided path is not a directory:
```

rolandlo:
The plugin dialog still has to be migrated (I have it somewhere).

rolandlo:
There is still an issue with the palette test not finding the paletteSettings.ui:

grafik

rolandlo:
Compiling with clang (18.1.3) on Linux failed:
image

rolandlo:
On MacOS ARM I had to change the link to a dynamic library (libsndfile) to make the test units run:  

```shell
install_name_tool -change @rpath/libsndfile.1.dylib /Users/roland/gtk/inst/lib/libsndfile.1.dylib test/test-unit
```

Not sure if that's a problem in Xournal++ or in the Gtk4-blob I built using jhbuild.

rolandlo:
On MacOS ARM I get some critical warnings related to size allocation:
grafik

rolandlo:
> > Compiling with clang (18.1.3) on Linux failed:
> 
> I don't get this issue, with clang 18.1.8 (neither with master nor with this branch).

With `-DCMAKE_CXX_COMPILER=clang++` it works indeed. I had `-DCMAKE_CXX_COMPILER=clang` instead.

rolandlo:
On MacOS ARM when I copy a stroke from one Xournal++ instance to another it gets pasted as image. That doesn't happen on Linux and neither on MacOS ARM when copying/pasting within the same instance.

rolandlo:
On MacOS zooming in/out using the mouse scroll wheel works with Ctrl, not with Meta.

rolandlo:
Toolbar customization doesn't work properly on MacOS ARM. Tools from the toolbar can be drag&dropped, but tools from the dialog cannot be dropped anywhere. I get a warning `Drop::leave` but no `DragItem`.

rolandlo:
On MacOS ARM when activating menu `Help > Help` I don't get to see the help in the browser, but a dialog "Operation not supported for the current backend". Maybe that's a Gtk issue?!

rolandlo:
I'm using Gtk 4.14.4 by the way.

rolandlo:
It looks related to this issue: https://gitlab.gnome.org/GNOME/glib/-/issues/3403

rolandlo:
> It looks related to this issue: https://gitlab.gnome.org/GNOME/glib/-/issues/3403

Actually if I somehow randomly add a (32 digit) id as file `/etc/machine-id` I get this precise error message.

rolandlo:
One more observation on MacOS ARM: Drawing is pretty jittery compared to master (with Gtk 3).

rolandlo:
> > Toolbar customization doesn't work properly on MacOS ARM. Tools from the toolbar can be drag&dropped, but tools from the dialog cannot be dropped anywhere. I get a warning `Drop::leave` but no `DragItem`.
> 
> Mmm... And you don't get any warnings/messages from within Drop::enter? Could you try the latest commit, with G_MESSAGES_DEBUG=all and see if you get a "Drop::enter but not GdkDrag - the drag probably comes from another app"?

I get `Drop::enter` events both when dragging from a toolbar or the dialog. I don't get any messages when I try to drop the tool on a toolbar, when it comes from the dialog. When it comes from a toolbar I get the warning `Drop::leave but no DragItem`.

I also get some Gdk-Criticals `_gdk_macos_surface_update_fullscreen_state: assertion `GDK_IS_MACOS_SURFACE (self) failed` whenenver I start to drag from either toolbar or dialog.

rolandlo:
> > On MacOS ARM when activating menu `Help > Help` I don't get to see the help in the browser, but a dialog "Operation not supported for the current backend". Maybe that's a Gtk issue?!
> 
> This certainly is a GTK issue, but there may be a workaround. Is the GTK version for MacOS ARM at least 4.10? If so, could you try the latest commit out and tell me if this works? They changed how URI's are to be opened in GTK 4.10 with a new GtkUriLauncher class. Could you also test this commit on Windows (if GTK >= 4.10 there as well)?

On Windows (with Gtk 4.14.4) it works fine.

rolandlo:
On Windows the palette test fails:

````
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from RenderedPaletteTabTest
[ RUN      ] RenderedPaletteTabTest.renderedPaletteTabShouldNotChangeSelectedColorPalette

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.862: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.863: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.864: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.865: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.867: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.891: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.892: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.895: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.902: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.904: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes

(xournalpp_test:8776): xopp-WARNING **: 18:19:11.918: Unexpected type of palette path encountered C:
/msys64/home/Roland/xournalpp/test/files/palettes
C:/msys64/home/Roland/xournalpp/test/gtk_tests/dialog/SettingsDialogPaletteTabTest.cpp:63: Failure
Expected equality of these values:
  palettePath.u8string()
    Which is: "C:/msys64/home/Roland/xournalpp/test/files/palettes/xournalpp.gpl"
  paletteTab.getSelectedPalette().value().u8string()
    Which is: "C:/msys64/home/Roland/xournalpp/test/files/palettes\\xournalpp.gpl"
[  FAILED  ] RenderedPaletteTabTest.renderedPaletteTabShouldNotChangeSelectedColorPalette (309 ms)
[----------] 1 test from RenderedPaletteTabTest (309 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (309 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] RenderedPaletteTabTest.renderedPaletteTabShouldNotChangeSelectedColorPalette

 1 FAILED TEST
````

**Comment by Febbe** [comment]
> On Windows the palette test fails:

Since when? @idotobi if you want to fix that, they must be lexicographic normalized. We can also fix that if that's too much hustle for you right now.

**Comment by rolandlo** [comment]
On Windows I didn't find a text editor from which I could drag&drop text to Xournal++. When trying to do so e.g. from Wordpad I get a Gtk-Warning: "Failed to receive drop data: Kein kompatibles Übertragungsformat gefunden".

**Comment by rolandlo** [comment]
On Windows I can't open any images (it works on master). I get an error message dialog: "Failed to load image". In the console "GDK_IS_PIXBUF (pixbuf)" fails. 
Drag&drop from the file manager works though.

**Comment by rolandlo** [comment]
On Windows during toolbar customization drag&drop from the dialog to a toolbar works, but drag&drop from one toolbar to another one doesn't work. It's like the tool gets lost, it's not placed on the other toolbar.

**Comment by rolandlo** [comment]
I have finished testing on Linux, MacOS ARM and Windows and have reported all issues I found. The only thing I couldn't test (because I didn't know how) is how to move tools to/from the floating toolbox during toolbar customization. IME should be tested more carefully by people who use it. It seemed to work from my perspective. I could only had a touchscreen to test on Linux, so if anyone has a touchscreen on Windows or MacOS, please test it.

**Comment by rolandlo** [comment]
> > It takes unusually long to open some documents (PDFs with annotations), clearly a regression from master.
> 
> This should be fixed now.

There is still something wrong. With maximized window, when I first open some PDF and then switch to a different PDF I get the lag. The document shows up immediately once I press on the window, but if I don't it can take a long time. 

Also I get some warnings like this:
grafik

**Comment by idotobi** [comment]
> > On Windows the palette test fails:
> 
> Since when? @idotobi if you want to fix that, they must be lexicographic normalized. We can also fix that if that's too much hustle for you right now.

@Febbe : If you could fix that, this would be great :pray:. I'm going on Vacation  tomorrow for a week (without PC) and I'll won't manage to fix it before. If it's not as urgent I can also try to fix it once I'm back (05th August).

File: src/util/include/util/Interval.h [hunk 10]
```diff
@@ -83,8 +83,8 @@ class Interval {
      * @return true if *this and other intersect, false otherwise
      */
     [[maybe_unused]] std::optional<Interval<T>> intersect(const Interval& other) const {
-        double newMin = std::max(min, other.min);
-        double newMax = std::min(max, other.max);
+        T newMin = std::max(min, other.min);
```
Febbe, line 86: @bhennion what is the background, that this interval is left open and not right open? This is unintuitive in computer science, because ranges are right-open, but intervals are not.

Doublonmousse:
> One more observation on MacOS ARM: Drawing is pretty jittery compared to master (with Gtk 3).

Does that only happen on MacOS ARM ? Maybe it has to do with the new `ngl` renderer being slower than `gl`, and maybe disproportionately so on apple silicon. It's worth trying to launch it with `GSK_RENDERER=gl` to check this.

Doublonmousse:
Do you still have clipboard issues ? I've noticed a similar behavior with rnote. Copy/paste only works correctly inside of the same instance of the application but not between two different ones (as the actual clipboard content does not include anything other than a png image).

This only occurs with MacOS on ARM (and seemingly this issue is not totally specific to gtk4).
I do have another clipboard implementation library that I know works with no issue on the same OS, maybe I can try to diff the two to find why this occurs.

rolandlo:
> > One more observation on MacOS ARM: Drawing is pretty jittery compared to master (with Gtk 3).
> 
> Does that only happen on MacOS ARM ? Maybe it has to do with the new `ngl` renderer being slower than `gl`, and maybe disproportionately so on apple silicon. It's worth trying to launch it with `GSK_RENDERER=gl` to check this.

I just tested. It happens on MacOS Intel as well. And using `GSK_RENDERER=gl` doesn't appear to help on either architecture.

rolandlo:
> Do you still have clipboard issues ? I've noticed a similar behavior with rnote. Copy/paste only works correctly inside of the same instance of the application but not between two different ones (as the actual clipboard content does not include anything other than a png image).
> 
> This only occurs with MacOS on ARM (and seemingly this issue is not totally specific to gtk4). I do have another clipboard implementation library that I know works with no issue on the same OS, maybe I can try to diff the two to find why this occurs.

It happened to me with MacOS on Intel as well though.

rolandlo:
On MacOS Intel with Gtk 4.10.4 I don't have the preferences and the "Info about Xournal++" in any menu. Accessing the preferences via the keyboard accelerator command+, also doesn't work. It's a Gtk issue, I believe, which was fixed in more recent versions. I will need to create a Gtk blob with a more recent Gtk version therefore.

rolandlo:
> Thanks a lot @rolandlo for your thorough testing! About the floating toolbox, you need to enable it in Settings->Drawing Area: Copie d'écran_20240726_160230
> 
> Then, when editing the toolbars, you'll see the floating toolbox: Copie d'écran_20240726_160407
> 
> You should be able to drag'n'drop icon therein as in any other toolbar.

Thanks. That works indeed (with the limitations on MacOS and Windows that applies to other toolbars as well).

Doublonmousse:
> Limitations in Drag/Drop or Clipboard interactions on MasOS could have the same cause. They go through the same GdkContentProvider interface.
> 
> Could you try opening two instances of `gtk4-demo --run=dnd` on MacOS, and dragging a color from the bottom row of one instance to one of the "items" of the other instance, on MacOS? It should change to color of the receiving item. I suspect it won't work and that could make a clear ticket for GTK's bugtracker.

Tested with gtk 4.14.4 on mac os and the drag and drop registers but the color isn't applied. On windows 11 the drag and drop operation never happens as the focus never switches to the second window ...

Linux is the only OS where this works correctly.

src/util/include/util/Interval.h
Febbe: Yes, I mean, that an integral interval class should model [a,b). I don't like complete closed intervals, because they need special treatment for integers (length) and they can't model empty ranges.

rolandlo:
With the latest commit c4f6f477c91ad72fd67253de12c3868bc53e8dfc I often get spinning wheels when switching the sidebar tab (both on Linux and MacOS Intel, where I tested this).

Doublonmousse:
I opened issues for clipboard and drag and drop related issues on the gtk gitlab :
- https://gitlab.gnome.org/GNOME/gtk/-/issues/6906 
- https://gitlab.gnome.org/GNOME/gtk/-/issues/6905

rolandlo:
Looks like there is an error with the sidebar page layout when the screen has scaling different to 100%. The pages are too far apart. It only happens with the page tab, not with the layer tabs. 
grafik

You can test with `GDK_BACKEND=x11 GDK_SCALE=2`.

rolandlo:
> > With the latest commit [c4f6f47](https://github.com/xournalpp/xournalpp/commit/c4f6f477c91ad72fd67253de12c3868bc53e8dfc) I often get spinning wheels when switching the sidebar tab (both on Linux and MacOS Intel, where I tested this).
> 
> Should be fixed now

This was fixed indeed.

rolandlo:
With `GDK_BACKEND=x11` the zoom in/out gestures don't work. Instead I get drawing like this:
grafik

rolandlo:
On my Fydetab Duo tablet (with Arch derived BredOS) tapping on the Touchpad to bring up the floating toolbox doesn't work somehow. It works with the bluetooth mouse though.

rolandlo:
During toolbar customization tools dragged from one of the toolbars don't get correctly displayed while dragging. It's most notable with `GDK_BACKEND=x11` since you see a black rectangle. On Wayland it's just transparent.

rolandlo:
During `cmake --build . --target test-units` I get a `-Wnarrowing` error with c++ 14.1.1 (on my ARCH based Fydetab) from the line:
```c++
testReadDataType(std::vector<char>{0, 42, -42});
```
It's about the narrowing conversion of -42 from int to char.

Febbe:
> > During `cmake --build . --target test-units` I get a `-Wnarrowing` error with c++ 14.1.1 (on my ARCH based Fydetab) from the line:
> > 
> > ```c++
> > testReadDataType(std::vector<char>{0, 42, -42});
> > ```
> > 
> > It's about the narrowing conversion of -42 from int to char.
> 
> This has nothing to do with this PR, I think.

Actually the compilers should not warn here, it's a constexpr literal and it's provable, that the value does not change. Also warnings in tests can be fixed in a different PR directly to master.

Febbe:
> > Looks like there is an error with the sidebar page layout when the screen has scaling different to 100%. The pages are too far apart. It only happens with the page tab, not with the layer tabs.
> > You can test with `GDK_BACKEND=x11 GDK_SCALE=2`.
> 
> This is a sneaky thing... In order to lay the miniatures out in the sidebar, we call (as we should) gtk_widget_measure(). For a GtkPicture (which we use to display the miniatures), this returns the size of the image buffer **without taking hi-resolution scaling into account**. So with GDK_SCALE=2, this returns twice the actual display size, leading to the bug.
> 
> Now, GtkImage is not affected by this bug, but it has a weird padding I can't get rid of (see also [this post](https://stackoverflow.com/questions/74628787/how-to-get-rid-of-padding-around-gtk-image-in-gtk-box-gtk-grid-in-gtk4))
> Copie d'écran_20240818_105400
> 
> To be honest, I don't quite understand what is precisely happening here. It may be related to our using a GtkFixed to lay the miniatures out (this is frowned upon). It may be that things are kinda working by accident because of [this](https://discourse.gnome.org/t/gtkpicture-not-displayed-inside-gtkfixed/6136).
> 
> For now, I'll just disable hi-resolution miniatures (so in hi-res setups, the miniatures will be a bit blurry/pixelized) until I find a proper fix.

Have you tried to modify margin/padding via css? Maybe there is also a property for widgets you can modify.

rolandlo:
> The plugin dialog still has to be migrated (I have it somewhere).

@bhennion Can you cherry-pick the head commit 18a6a755f7cb6a0ae8ed98b0a0ec0cf392e6132f of https://github.com/rolandlo/xournalpp/tree/migrate-plugin-dialog

rolandlo:
> @rolandlo Would you be able to give the MacOS ARM CI a look? I'm sure we need to pack GTK4 in the bundle somewhere. (Also, do we need to rename the bundle so the Cirrus server keeps both GTK3 and 4 versions in cache?)

Sure, I have the needed changes and will push a branch with them. Let's wait until the 1.2.4 release is over, so we don't mess it up. Yes, the Cirrus server should keep both GTK3 and 4 versions in cache. Probably renaming the task is enough, but practice will show.

rolandlo:
> > @rolandlo Would you be able to give the MacOS ARM CI a look? I'm sure we need to pack GTK4 in the bundle somewhere. (Also, do we need to rename the bundle so the Cirrus server keeps both GTK3 and 4 versions in cache?)
> 
> Sure, I have the needed changes and will push a branch with them. Let's wait until the 1.2.4 release is over, so we don't mess it up. Yes, the Cirrus server should keep both GTK3 and 4 versions in cache. Probably renaming the task is enough, but practice will show.

Please cherrypick the last two commits (1293b39e42daa5b09633d1f637c34a151a5bbf05 and cc94ffa6623706770a977e3f16e87e831c38e427) from https://github.com/rolandlo/xournalpp/tree/update-cirrus-to-gtk4

I have already checked that the Gtk4 version built works and that the Gtk3 blob can still be used from other branches without recreating it.

MathisP75:
Does anything special needs to be done when compiling this branch? When I try to open the gtk4 demo, I get the following:

`(com.github.xournalpp.xournalpp:3839): xopp-WARNING **: 17:41:16.796: gtk3-demo was not found in path`

bhennion:
> Does anything special needs to be done when compiling this branch? When I try to open the gtk4 demo, I get the following:
> 
> `(com.github.xournalpp.xournalpp:3839): xopp-WARNING **: 17:41:16.796: gtk3-demo was not found in path`

That's a typo in the error message. The executable that is not found is `gtk4-demo`. In any case, if you want to run gtk4-demo, you need to install it... The way to do so depends on your OS and distribution.

MathisP75:
Perfect, thank you. It's the `gtk4-demos` package on Arch. Is this branch supposed to make Xournal use the Libadwaita theme, as I saw in some screenshots here? I successfully compiled it and it says it's using gtk 4.20.3, but I barely see any visual changes.

rolandlo:
> Perfect, thank you. It's the `gtk4-demos` package on Arch. Is this branch supposed to make Xournal use the Libadwaita theme, as I saw in some screenshots here? I successfully compiled it and it says it's using gtk 4.20.3, but I barely see any visual changes.

This branch is just to get Xournal++ compiling with Gtk 4 and mostly functioning and looking like before. It doesn't use anything related to Libadwaita.

youssefadly237:
just realized GTK has Android backend `gdk/android`, not sure how stable it really is yet, but it might be an opportunity for having an experimental android app with the same codebase as master

bhennion:
> just realized GTK has Android backend `gdk/android`, not sure how stable it really is yet, but it might be an opportunity for having an experimental android app with the same codebase as master

Yeah. I wanted to try that at some point. The main issue is the [packaging system](https://github.com/sp1ritCS/gtk-android-builder). Last time I checked, it required a meson script. Porting our CMake scripts to Meson would be a long and difficult task, I think.

PR #5630, opened 2024-04-20
Make BackgroundSelectDialog's GTK4-compatible

bhennion:
Trying to make small PR's prior to #5615, to ease the reviewing process.

This one simply applies the same strategy as always to make the background change dialogs non-blocking and gtk4-ready.


src/core/gui/dialog/backgroundSelect/BackgroundSelectDialogBase.cpp
```diff
@@ -102,12 +102,16 @@ void BackgroundSelectDialogBase::setSelected(size_t selected) {
...
-    if (selected < elements.size()) {
-        elements[selected]->setSelected(true);
+    if (selected < entries.size()) {
+        entries[selected]->setSelected(true);
         this->selected = selected;
     }
+
+    if (selected < this->entries.size()) {
+        gtk_widget_set_sensitive(okButton, true);
+    }
```
rolandlo, line 116: Can't this be simplified as follows?
```suggestion
    if (selected < entries.size()) {
        entries[selected]->setSelected(true);
        this->selected = selected;
        gtk_widget_set_sensitive(okButton, true);
    }
```

rolandlo (changes requested):
I found one issue:
When replacing an image background by a larger image or a larger pdf page (via "Apply to current page" from the toolbar popover) the layout of the pages looks screwed up, the pages overlap.

Another issue: The red border for the images in the preview looks displaced.
image

Otherwise everything seems to work fine (with Gtk 3). Please take into account the additional comments I made to the code.

rolandlo:
> > I found one issue: When replacing an image background by a larger image or a larger pdf page (via "Apply to current page" from the toolbar popover) the layout of the pages looks screwed up, the pages overlap.
> 
> I solved every thing except this. I couldn't reproduce. What settings are you using for the layout?

Looks like it only happens with `GDK_BACKEND=x11`. After zooming in or out everything is normal again.

rolandlo:
I start with 3 A4-pages (graph background) and replace the second one by a small image, then replace it by a bigger one.

rolandlo (approved):
It works on my end fine as well now.

PR #5638, opened 2024-04-25
Fix conversion problems with `fs::path` on Windows

panpaul:
(I'm testing on MS Windows and my system language is set to Chinese)

I have observed these problems:

1. Locales will not load when installed to a path with non-ASCII characters.
2. Saved annotated PDF files (*.xopp) could not be reopened when the PDF file-path contains non-ASCII characters.
3. If there are non-ASCII characters in the "default saved name" template, the name generated in the saving dialog will be garbled.

These problems are mainly caused by the conversion between UTF-16 and UTF-8, fixed by using `.u8string()` and `fs::u8path()`


Febbe:
#5640 must be applied first.

src/core/control/XournalMain.cpp
```diff
@@ -85,7 +85,7 @@ void initCAndCoutLocales() {
 void initLocalisation() {
 #ifdef ENABLE_NLS
     fs::path localeDir = Util::getGettextFilepath(Util::getLocalePath().u8string().c_str());
-    bindtextdomain(GETTEXT_PACKAGE, localeDir.u8string().c_str());
+    bindtextdomain(GETTEXT_PACKAGE, localeDir.string().c_str());
```
Febbe, line 88: How you know that `bindtextdomain` requires system narrow encoded strings instead of utf-8?

Febbe:
I have the bad feeling, that this patch will break old xopp files with paths containing non ASCII characters, since they stored paths as ANSI/system encoded. Now those paths are assumed to be encoded in utf-8.

Febbe:
> One question: There are other instances of fs::path::string() in the code (see a list below). Should some of those also be changed? Some are for error messages, so they should not be converted to utf8 (I think), but some are really path manipulations, and should probably be converted as well.
> List of fs::path::string() occurrences

src/util/PathUtil.cpp:
79: dangerous to use, but only one usage where a system encoded path is required ⇒ no bug here.
85: function returns a boolean and works over ASCII only characters. It would only be a problem if windows still uses code pages, which do not extend ASCII.
90: same as 85.
95: dangerous, no known fix, since it's ambiguous, whether `const std::string& ext` is UTF-8 or ANSI/System encoded. Safe when used with ASCII characters only.
297: g_warning does not know anything about encodings, it's forwarded to the console, which has most likely system encoding ⇒ string is the better choice here.

src/core/control/jobs/AutosaveJob.cpp:45 : the same of g_warning applies for g_message

src/core/control/latex/LatexGenerator.cpp:53 : The variable name already states, that the path must be encoded as OS encoded string ⇒ string is correct here, `u8string` would be a bug.

src/core/control/settings/MetadataManager.cpp
30: string is better choice
36: same
95: result is used as parameter to strtoll, I bet it only supports ASCII. Chinese numbers will never be accepted, whether it is utf8 or ASCII/ANSY/any other encoding.

src/core/control/settings/Settings.cpp: All are used in a g_xxxxx log function, so string is the better choice.

src/core/control/tools/ImageHandler.cpp:69 : Not part of 1.2.x, **but u8string would be better here**. Has only visual effects

src/core/control/xojfile/SaveHandler.cpp: **Can't change those without breaking old files, u8string is better here!**. But it has no negative impact on computers with the same system encoding

src/core/control/xojfile/LoadHandler.cpp: 
All uses of error(...) are prone to bugs and may not work as expected.
This should be refactored. Can't tell, whether error is used only in g_xxx log functions or if it is also used in an error message.
No negative impact for the correctness of xournalpp only look and feel.
Changing string to u8string for the rest will also break old files.

src/core/control/CrashHandler.cpp: Safe
src/core/control/LatexController.cpp: Safe

src/core/control/XournalMain.cpp:
124: Safe
125: Safe
167: Safe, since ASCII only characters are used
191: **u8string should be used for better look and feel**, no breaking bug
446: **u8string should be used for better look and feel**, no breaking bug
504: Safe

src/core/control/Control.cpp:
238:  **u8string should be used for better look and feel**, no breaking bug
1424: Safe
1608: u8string is the better choice for LAF
1609 neither string nor u8string are good here, string is required for correctness, u8string for LAF, best would be to keep filename as path.
1616: ???
1617: **u8string LAF**
1620: **u8string LAF**
1621: Same as 1609
1639: **u8string LAF**

src/core/gui/GladeSearchpath.cpp:37: Safe

src/core/plugin/luapi_application.h: I think lua works on the system path, but @rolandlo is the better one to ask.
src/core/plugin/Plugin.cpp:
src/core/plugin/PluginController.cpp: I think lua works on the system path, but @rolandlo is the better one to ask.

src/exe/osx/setup-env.cpp:
Completely secure, since osx and unix work on utf-8 anyway, there is no difference between u8string and string.

src/core/control/XournalMain.cpp
Febbe: https://github.com/microsoft/vcpkg/issues/9854

We must be careful here, vcpkg fixed this for gettext.

src/util/PathUtil.cpp
```diff
@@ -189,7 +189,7 @@ auto Util::getGettextFilepath(const char* localeDir) -> fs::path {
     }
     const char* dir = (gettextEnv) ? directories.c_str() : localeDir;
     g_debug("TEXTDOMAINDIR = %s, Platform-specific locale dir = %s, chosen directory = %s", gettextEnv, localeDir, dir);
-    return fs::path(dir);
+    return fs::u8path(dir);
```
Febbe, line 192: This is generally wrong, since gettextEnv has the system encoding. So we can't just expect utf-8.
See #5640 for a correct fix.

src/core/model/Document.cpp
```diff
@@ -168,7 +168,7 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
             c = '_';
         }
     }
-    fs::path p = filename;
```
Febbe, line 171: Filename is not necessarily UTF-8.

src/core/control/XournalMain.cpp
Febbe: > Maybe we should use `wbindtextdomain`, as suggested by the [documentation](https://www.gnu.org/software/gettext/manual/gettext.html#:~:text=On%20native%20Windows%20platforms%2C%20the%20main%20function%20may%20call%20wbindtextdomain%20instead%20of%20bindtextdomain.).

I think we are with that on the secure side. The patch does not seem to be applied upstream, only in vcpkg.

src/util/PathUtil.cpp
Febbe: After your test in https://github.com/xournalpp/xournalpp/pull/5640 this is correct, but the function and it's usage was odd in any way. It should accept a fs::path and not a string of unknown encoding.

Febbe:
> In the original `LoadHandler`: pdf filename is parsed as UTF-8 string [here](https://github.com/xournalpp/xournalpp/blob/efabdd83cd8e80b317fb58590d4b5e410d270fac/src/core/control/xojfile/LoadHandler.cpp#L448), so I modified `SaveHandler` to align with this behaviour.

Loading of the system file encoding would be better, since old files are already stored with it.

> Regarding the background image filename, yes, it may break compatibility with old `xopp` file. I'll revert those background image related changes to maintain compatibility.

Maybe, we can find a solution to use utf-8 strings in general, e.g. by advancing the file version. We would like to have them in utf8 anyway. That would improve portability of xopp files, and keep old files compatible.

Febbe:
@panpaul you can increase the DEV_FILE_FORMAT_VERSION in https://github.com/xournalpp/xournalpp/blob/8dbb0cf81a0aecf2e5c1c47537f2654f62b72e94/CMakeLists.txt#L251 from 4 to 5 and
 - patch the LoadHandler to use fs::path for version less than 5
 - patch the Load- and SaveHandler to use u8string for versions greater 4

This would be the best of both versions.

Febbe (approved):

Febbe:
@bhennion what do you think of that solution?

src/core/model/Document.cpp
Febbe: Since MSVC does weird things here, I would go for 2. 
Can you also file a bug report for MSVC, so they can fix that?
Can't believe, they did that intentionally.

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -1045,7 +1063,7 @@ void LoadHandler::parserText(GMarkupParseContext* context, const gchar* text, gs
                     handler->stroke->setPressure(handler->pressureBuffer);
                 }
             } else {
-                g_warning("%s", FC(_F("xoj-File: {1}") % handler->filepath.string().c_str()));
+                g_warning("%s", FC(_F("xoj-File: {1}") % handler->filepath.u8string()));
```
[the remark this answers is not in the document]
Febbe: Thank you, then we at least know that g_error and siblings assume utf-8 but it seems to be that this is still bugged on windows. One example is this: https://gitlab.gnome.org/GNOME/glib/-/issues/3341 but we should use utf-8 in the future, because it's not intended to not support utf-8.

Febbe:
>[bhennion](https://github.com/bhennion) commented [Apr 26, 2024](https://github.com/xournalpp/xournalpp/pull/5638#issuecomment-2079711965)
>
> It sounds good to me. I'd add a test unit for version 5 files, and conversion from version 4 to 5 though.

@panpaul are you willing to do this, or is that too much effort for you?

Febbe:
Regarding [1], when is g_markup_parse_context_parse called? 
Clang:`std::format::path(std::format::path::string)` should always create a valid path, even when the characters are interpreted as garbled with utf8.
Is that a bug in clang, or do we store the string formatted with any locale?

> If so, I think we may not have to upgrade the file format version to 5

I think, since we ever only supported mingw, we can really change `.string()` to `u8string()` and `path{...}` to `u8path(...)` without any version bumb or utf-8 detection.
This would be a fix for clang and msvc **custom** builds only.

Febbe:
> Regarding [1], when is g_markup_parse_context_parse called? 
> Clang:`std::format::path(std::format::path::string)` should always create a valid path, even when the characters are interpreted as garbled with utf8.
> Is that a bug in clang, or do we store the string formatted with any locale?
> 
> > If so, I think we may not have to upgrade the file format version to 5
> 
> I think, since we ever only supported mingw, we can really change `.string()` to `u8string()` and `path{...}` to `u8path(...)` without any version bumb or utf-8 detection.
> This would be a fix for clang and msvc **custom** builds only.

Test files, that test non western utf8 symbols would still be good.

Febbe:
> std::format::path ? I guess you want to say std::filesystem::path(std::string)

Yes

> To be honest, I don't quite understand the constructor part...

It's just an invariant I assume. Constructing a path from the narrow string representation of another path via the constructor for narrow strings should result in an equal path object.

> Sorry, I didn't catch you. So should I close this PR and wait for #5641 (and add tests for non western utf8 symbol in #5641)

No, just return to your first version without the version bump. Exporting paths with a clang/msvc compiled binary should also be correct.
So changing string to u8string is good
And changing the path construction from fs::path{...} to fs::u8path(...) is really mandatory.
People using our official installation won't have any negative impact.

Febbe:
Thank you @panpaul!

PR #5642, opened 2024-04-26
RGB to HSL

lancechoe:
For the given task, I implemented two new methods in the color.cpp file of Xournal++. The first method, rgb_to_hsl, converts an RGB color to the HSL (Hue, Saturation, Lightness) color space. The second method, rgb_to_cmyk, converts an RGB color to the CMYK (Cyan, Magenta, Yellow, Black) color space.

In the rgb_to_hsl function, I first normalized the RGB components to the range [0, 1]. Then, I calculated the maximum and minimum values among the RGB components to determine the lightness (L) and delta (Δ) values. Using these values, I computed the hue (H) and saturation (S) values according to the HSL color space formulas. Finally, I ensured that the hue value is normalized within the range [0, 360] degrees.

In the rgb_to_cmyk function, I also normalized the RGB components to the range [0, 1]. Then, I computed the black (K) component as the maximum of the normalized RGB components. Using the black component, I calculated the cyan (C), magenta (M), and yellow (Y) components according to the CMYK color space formulas. I normalized these components to ensure they fall within the range [0, 1].

Both functions utilize standard color space conversion formulas to accurately transform RGB colors to the desired color space. These functions contribute to enhancing the versatility of Xournal++ by enabling users to work with colors in different formats, facilitating a wider range of color manipulation and understanding within the application.

### DIFF
--- a/src/util/PathUtil.cpp
+++ b/src/util/PathUtil.cpp
@@ -334,3 +334,74 @@ auto Util::getLocalePath() -> fs::path {
     return getDataPath() / ".." / "locale";
 #endif
 }
+
+float Util::as_grayscale_color(Color color) {
+    GdkRGBA components = rgb_to_GdkRGBA(color);
+    float componentAvg = static_cast<float>(components.red + components.green + components.blue) / 3.0f;
+
+    return componentAvg;
+}
+
+float Util::get_color_contrast(Color color1, Color color2) {
+    float grayscale1 = as_grayscale_color(color1);
+    float grayscale2 = as_grayscale_color(color2);
+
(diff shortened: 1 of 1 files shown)


rolandlo:
@lancechoe Thanks for your PR. Please make sure that
A) The PR compiles (see Details in the failed checks)
B) The functions you introduce are used somewhere to the benefit of the user of the app
C) Explain why you think the user of the app benefits from your code changes

rolandlo:
Closing this PR for now. Feel free to reopen once A), B) and C) from above are met.

PR #5644, opened 2024-04-28
Refactor and simplify sidebar code

bhennion:
This is part of the simplification made for gtk4. It includes

- Moving the control part of the sidebar to Control/ActionDatabase -> @rolandlo, that's what you need for the plugins right?
- Simplifying the UI construction.

It won't compile as is (unless -DENABLE_PLUGINS=off), as changes to the plugin API are required and @rolandlo already did those on some branch somewhere :)

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1623,7 +1623,7 @@ void Control::promptMissingPdf(LoadHandler& loadHandler, const fs::path& filepat
             std::bind(&Control::missingPdfDialogResponseHandler, this, proposedPdfFilepath, std::placeholders::_1));
 }
 
-void Control::missingPdfDialogResponseHandler(const fs::path& proposedPdfFilepath, int responseId) {
+void Control::missingPdfDialogResponseHandler(fs::path proposedPdfFilepath, int responseId) {
     switch (static_cast<MissingPdfDialogOptions>(responseId)) {
         case MissingPdfDialogOptions::USE_PROPOSED:
             if (!proposedPdfFilepath.empty()) {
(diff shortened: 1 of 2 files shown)


rolandlo:
> TBH, I don't understand the sidebar related plugin API. It makes no sense to expose internal indices (e.g. the index of the current sidebar tab) to plugins. It makes adding a new tab a change potentially breaking some plugins. We should really be more careful to what we expose in this API, I think.

You're right about that from today's perspective. When the sidebar related plugin API was introduced (#2406) there was no other way to do things like duplicating a page or layer and it was realized in #5059 that the result of executing sidebar actions depend on the currently selected sidebar page. Now everything can be obtained via the new action system, so I think it's probably time to deprecate sidebar API and layer actions.

rolandlo:
You were looking for https://github.com/rolandlo/xournalpp/commits/fix-plugins/
I have added these sidebar related changes to this PR. You can work from there if you like.

rolandlo:
> I think this is ready for a review then.

One question: When do we deprecate sidebar related API and layer actions API (and adapt existing plugins). Should we do that now or when `uiAction` is deprecated as well and replaced by a function that does without the BackwardCompatibilityLayer?

> There is an issue that could be dealt with while we're at it:
> 
>     * The color of the background behind the miniatures: I think it should follow the theme's background. This is fine in page or "progressive layers" modes, but not in "individual layers" mode. Should we paint a background to the individual layers? If so, of what color? We could use the paper color if its defined, but how do we choose the color in the case of pdf/image backgrounds?

I have encountered this issue many times (with PDF background) and I don't know of a good solution. In my case (where the PDF background is mostly White), I would prefer that the miniatures have a White background as well, but other users may have a mostly black PDF, when the situation is different again.

rolandlo:
Ok, I'll work on it.

rolandlo:
> OK, I added a commit doing two things:
> 
>     1. Fixes the previews when the background is not visible: before, the previews showed the checkers board, while it should always show the background (so the user knows what they can show/hide).
> 
>     2. For the layer view, the background color is always painted before the layer. For pdf or image backgrounds, this means a white background, otherwise the selected background color is used. This seemed to me like the most reasonable behaviour (but that's debatable, of course).

This looks nice. One issue I have discovered while testing it out: When I have moved files and need to replace the background by the suggested background the previews don't use that background. They only do so once I (save and) reload the file.

rolandlo:
> Did you check the code as well?

I'm almost done with going through the code, but not quite yet.

rolandlo (approved):
LGTM (except of the include in Sidebar.cpp)

PR #5646, opened 2024-04-28
Unify Key handling accross tool handlers and use GtkEventControllerKey

bhennion:
Again, a preliminary to gtk4. It removes direct access to GdkEvent's members.


rolandlo (commented):
This looks mostly good. I have just some minor comments.

Issue #5663, opened 2024-05-06
I can't change the type of sheet, the program closes when I try to change it.


vladQll:
### Operating System

Windows

### (Linux Only) Display Server

None

### Installation Method

official release

### Xournal++ Version

v1.2.3+dev

### libgtk Version

3.24.41

### Bug Description

I try to change to grid sheet but the program closes when I do it.

### Expected Behaviour

would like it to work as usual

### Steps to Reproduce

1. open xournal++

2. go to sheet options

3. choose any type of different sheet style

rolandlo:
The fix (#5665) will be available in the next nightly.

PR #5665, opened 2024-05-07
Fix crash when changing page type

rolandlo:
This fixes a regression from #5630
The crash happens when using menu Journal > Page Background (e.g. changing from graph to lined). In that case `pageTypeForNewPages` may not hold a value.


rolandlo:
> This fixes #5663, right?

Yes, it does. I didn't notice there was already an issue raised. 

Merging right away, no need to wait for confirmation imho.

PR #5667, opened 2024-05-08
Load plugins only from subfolders of plugin folder

rolandlo:
Currently the file plugins/luapi_application.def.lua raises a warning.

The warning only appears on master (since #5572), so I think it's enough to target it to the master branch.


rolandlo:

PR #5681, opened 2024-05-12
Fix PdfExport of PDF background

### LINKED ISSUE #5680: PDF Background is not exported any more
### Operating System

Linux

### (Linux only) Distribution

Ubuntu 24.04

Gnome

None

### Installation Method

PPA

### Xournal++ Version

1.2.3+dev

### libgtk Version

3.24.41

### Bug Description

When exporting a xopp-file with PDF background to PDF, the background is not exported, just the annotations.

### Expected Behaviour

The background should be exported as it did previously.

### Steps to Reproduce

1. Open a PDF with Xournal++
2. Annotate
3. Export as PDF
4. Observe the missing background

### Additional Context

_No response_

bhennion:
Fixes #5680

Thank you for the catch @rolandlo!

### DIFF
--- a/src/core/control/PrintHandler.cpp
+++ b/src/core/control/PrintHandler.cpp
@@ -48,8 +48,11 @@ void drawPage(GtkPrintOperation* /*operation*/, GtkPrintContext* context, int pa
         }
     }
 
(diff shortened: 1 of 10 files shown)


rolandlo (approved):
It works fine. Thanks for the quick fix,.

PR #5688, opened 2024-05-14
Fix geometric tool keyboard accelerators

### LINKED ISSUE #5687: Geometry tools don't take into account the shift key any more
### Operating System

Linux

### (Linux only) Distribution

Ubuntu 24.04

Gnome

Wayland

### Installation Method

PPA

### Xournal++ Version

1.2.3+dev

### libgtk Version

3.24.41

### Bug Description

When scaling (with s/S) or rotating (r/R) a geometry tool (setsquare/compass) via the keyboard, the shift key is not taken into account any more. So the geometry tool can currently only be enlarged and rotated counter-clockwise via the keyboard, not shrinked or rotated clockwise.
The Alt key is still working.

### Expected Behaviour

The geometry tools keyboard navigation should work like it did previously.

### Steps to Reproduce

1. Switch on the setsquare or compass
2. Press s and Shift+s and notice there is no difference
3. Press r and Shift+r and notice there is no difference

### Additional Context

Probably a regression from #5646 (didn't check though).

bhennion:
Fix #5687

### DIFF
--- a/src/core/gui/inputdevices/GeometryToolInputHandler.cpp
+++ b/src/core/gui/inputdevices/GeometryToolInputHandler.cpp
@@ -150,25 +150,20 @@ auto GeometryToolInputHandler::keyPressed(KeyEvent const& event) -> bool {
             ydir = 1;
             break;
         case GDK_KEY_r:
-        case GDK_KEY_R: {  // r like "rotate"
+            angle = (event.state & GDK_MOD1_MASK) ? -ROTATE_AMOUNT_SMALL : -ROTATE_AMOUNT;
+            break;
+        case GDK_KEY_R:  // r like "rotate"
             angle = (event.state & GDK_MOD1_MASK) ? ROTATE_AMOUNT_SMALL : ROTATE_AMOUNT;
-            angle = (event.state & GDK_SHIFT_MASK) ? angle : -angle;
             break;
-        }
         case GDK_KEY_s:
-        case GDK_KEY_S: {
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
Thanks for the fix. The rotating and scaling works again.

I just noticed another issue: In the stable version when you use the text tool while a geometry tool is active, you can type normally. Now when you type keys like r and s are not absorbed by the text tool, but by the geometry tool. Should I raise an issue about that?

rolandlo:
> Hmm... Not sure what exactly caused that. It actually was already a problem with SplineHandler prior to the recent PRs.
> 
> TBH, I think there is something weird with the GeometricTools. No other tool has a "InputHandler". The input handlers really correspond to devices. A consequence of that is that we have limited control on what tool gets the events first. We can basically say (more or less on a per event type basis)
> 
>     * The geometric tools always gets events first (current behaviour)
> 
>     * The geometric tools always gets the events after all the other tools (via the PageView event handlers).
> 
> 
> A quick fix would be to have the geometric tool get keyboard events last, but other events first. That's fine. I'll add it to this PR, but I think this is symptomatic of a deeper issue in the code structure.

That doesn't work though. For example the arrow keys should move the geometry tool when there is one and not the page.

Why not just repeat the lines
```c++
    if (xournal->getSelection() || xournal->getControl()->getTextEditor()) {
        return false;
    }
```
from `GeometryToolInputHandler::handle` in `GeometryToolInputHandler::keyPressed`?

PR #5702, opened 2024-05-24
Prevent focus loss after save

### LINKED ISSUE #5132: Save lose the focus
### Operating System

Linux

### (Linux only) Distribution

Fedora

GNOME

Wayland

### Installation Method

Flatpak

### Xournal++ Version

1.2.1

### libgtk Version

whatever the flatpak use

### Bug Description

When I save the focus is lost, ie the text insertion point is gone, where ever I was editing the text.

### Expected Behaviour

The focus (ie where the text insertion point) is lost.

### Steps to Reproduce

1. Start writing some text
2. Save (press ctrl-S or select Save)

You can't type more.

### Additional Context

_No response_

MeiramSh:
Fixes #5132

This commit changes `Control::save`. From now on, `Control::save` remembers cursor position, saves edition by exiting, then starts new edition on the same position.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1794,7 +1794,7 @@ auto Control::save(bool synchron) -> bool {
 
     auto v = win->getXournal()->getViewFor(getCurrentPageNo());
     auto edit = v->getTextEditor();
-    if(edit) {
+    if (edit) {
         PositionInputData pos = {};
         pos.x = edit->getTextElement()->getX() + edit->getCursorBox().getX();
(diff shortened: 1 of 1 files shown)


rolandlo:
Thanks for your contribution. Unfortunately it doesn't work on my side. The cursor jumps to a different position, where a new text field is opened, when saving.

MeiramSh:
@rolandlo, yeah, sorry, I realized that I only tested the functionality at 100% zoom. You are correct; it doesn't work properly when the page is scaled. I have opened a new pull request #5742 which works with scaling.

PR #5733, opened 2024-06-13
Fix warning in image actions plugin

rolandlo:
The inner and outer loop used the same name for the loop variable, which raised a well justified warning.


rolandlo:

PR #5742, opened 2024-06-24
Prevent focus loss after save

### LINKED ISSUE #5132: Save lose the focus

MeiramSh:
Fixes #5132

This pull request updates #5702 by now accounting for zooming.

## Explanation

In a previous pull request, `pos.x` and `pos.y` already accounted for zooming, while `onButtonPressEvent` and `onButtonReleaseEvent` expected positions that did not account for zooming.

`onButtonPressEvent` and `onButtonReleaseEvent` were applying zooming by dividing coordinates by a zooming factor.

To place the cursor in the correct position, I had to multiply `pos.x` and `pos.y` by the zooming factor.

## Possible Improvements

This is not the only solution I considered. There were improvements I wanted to make, but due to my limited understanding of the contribution requirements, I was unsure whether they would be beneficial.

While writing this pull request, I tried to be as concise as possible, changing a minimal number of files and making the least number of additions. Yet, it is not the most elegant or *correct* solution.

I believe that the *correct* solution would be to add a new method to the `TextEditor` class named `sync`, which would synchronize the `TextEditor` contents with a text element, without closing the `TextEditor` (freeing the pointer). Then, we would refactor the `closeSelectionEndText` method into `closeSelectionSyncText`, which would call the `sync` method instead of freeing the pointer. However, this seemed as an overkill for such small enhancement, so I implemented a simple, *hacky* solution.

My current solution feels *hacky* because the `onButtonPressEvent` and `onButtonReleaseEvent` methods weren't intended to be used this way. The fact that coordinates were divided by the zooming factor, only to be multiplied by it again, seems a bit inelegant. Nonetheless, the solution works, and it is the only one I could implement without refactoring many classes and methods.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -59,6 +59,7 @@
 #include "gui/inputdevices/CompassInputHandler.h"                // for Comp...
 #include "gui/inputdevices/GeometryToolInputHandler.h"           // for Geom...
 #include "gui/inputdevices/HandRecognition.h"                    // for Hand...
+#include "gui/inputdevices/PositionInputData.h"
 #include "gui/inputdevices/SetsquareInputHandler.h"              // for Sets...
 #include "gui/menus/menubar/Menubar.h"                           // for Menubar
 #include "gui/sidebar/Sidebar.h"                                 // for Sidebar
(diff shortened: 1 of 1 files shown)


rolandlo:
Thanks for the improved PR and the explanations.
When testing it worked fine on Wayland, but not so on X11. Running it via
```
GDK_BACKEND=x11 path/to/xournalpp
```
I experienced jumping in the cursor position when saving while editing text.

MeiramSh:
Thanks for reviewing my PR!

The video demonstration is also made using x11, on Debian 12, Xfce spin, which is quite interesting.

I will try to reproduce the behavior later.

I could be wrong but my initial thoughts are that the problem may lie in floating point operation precision e.g. dividing and then multiplying by a zooming factor may not preserve initial coordinates.

MeiramSh:
I tried to replicate the wrong behavior, but had no luck:
Uploading 2024-08-30 19-16-26.mp4…

PR #5747, opened 2024-07-01
Master-merge

Febbe:
Please check if everything is alright.


Febbe:
@xournalpp/core will merge in 24h if no objections are raised.

Issue #5285, opened 2023-10-25
Add a empty page after every page in a pdf


hanna2826:
### Is Your Feature Request Related to a Problem?

I use Xournal++ for Uni, but when annotating books, or other pdfs, I often run out of space. 
My Uni does offer a printed version of a lot of the books where the left side is a page from the pdf and the right one is empty for annotations. 
Is it possible to implement this? And also change the background of the empty space for example to lines/ checkered?

### Feature Description

Option to add a clear page after every page in pdfs

rolandlo:
There is no built-in option for that in the GUI, but you could easily write a [plugin](https://xournalpp.github.io/guide/plugins/plugins/) doing that. You basically need to iterate through all the pages,
use
`app.uiAction({["action"] = "ACTION_NEW_PAGE_AFTER"})`  to insert a new page after the current one
and
`app.changeCurrentPageBackground("graph")`

for changing to "graph" page style (or "lined", "ruled", whatever you need).

rolandlo:
Closed in view of #5836.

PR #5836, opened 2024-09-01
Plugin: Add pages for notes automatically

atticus-sullivan:
Adds a plugin which adds empty pages after each page showing a page from the pdf if the page is the last page of the document or the following page also shows a page from the pdf.

This rather complicated conditional makes the function idempotent, so executing it more than once won't do more changes.

To provide full flexibility to the user, there is also a `force` mode which simply skips the check and unconditionally adds an empty page after every page in the document.

Implements #5285.

After applying this, one can e.g. set the view to two-column mode so that one has always the pdf page on the left and have space on the right for notes.

---
Note: Initially I thought about naming the plugin `BatchChanges` as it allows us to apply one operation on the complete document. Also in the beginning I had functionalities like *change the background color of all pages* or so in mind (which I dropped for now because those functionalities would require more complex user input (-> lgi) which I have not really experience working with.

Just wanted to add this note so that you know I'm totally fine if you later decide to move the functionality of this plugin into another plugin if it fits.

### DIFF
--- a/plugins/SpaceForNotes/main.lua
+++ b/plugins/SpaceForNotes/main.lua
@@ -1,0 +1,28 @@
+local m = {
+    NORMAL = 0,
+    FORCE = 1,
+}
+
+-- Register all Toolbar actions and intialize all UI stuff
+function initUi()
+  app.registerUi{menu="Add one empty pages after every pdf page once", callback="AddEmpty", mode=m.NORMAL}
+  app.registerUi{menu="Add empty pages after every page", callback="AddEmpty", mode=m.FORCE}
+end
+
+function AddEmpty(mode)
+    local struct = app.getDocumentStructure()
+    for p=#struct.pages,1,-1 do
+        local page = struct.pages[p]
+        local pred = true
+        if mode ~= m.FORCE then
+            local nextPage = struct.pages[p+1]
+            -- insert a new empty page if p is a pdf-page and succeeding page either does not exist or is also a pdf-page
+            pred = page.pdfBackgroundPageNo ~= 0 and (nextPage == nil or nextPage.pdfBackgroundPageNo ~= 0)
+        end
+        -- if mode is force => pred is set to true
+        if pred then
+            app.setCurrentPage(p)
+            app.uiAction{action="ACTION_NEW_PAGE_AFTER"}
+        end
+    end
+end
--- a/plugins/SpaceForNotes/plugin.ini
+++ b/plugins/SpaceForNotes/plugin.ini
@@ -1,0 +1,15 @@
+[about]
+## Author / Copyright notice
+author=Lukas Heindl
+
+description=Add empty pages to have space for notes
+
+## If the plugin is packed with Xournal++, use
+## <xournalpp> then it gets the same version number
+version=<xournalpp>
+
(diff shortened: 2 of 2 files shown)


plugins/SpaceForNotes/main.lua
```diff
@@ -0,0 +1,36 @@
+local m = {
+    NORMAL = 0,
+    FORCE = 1,
+}
+
+-- Register all Toolbar actions and intialize all UI stuff
+function initUi()
+  app.registerUi{menu="Add one empty pages after all pdf page once", callback="AddEmpty", mode=m.NORMAL}
```
rolandlo, line 8:
```suggestion
  app.registerUi{menu="Add one empty pages after every pdf page once", callback="AddEmpty", mode=m.NORMAL}
```

plugins/SpaceForNotes/main.lua
```diff
@@ -0,0 +1,36 @@
...
+                local n = struct.pages[p+1]
+                -- no p is the last page -> also insert an empty one
+                if n == nil then
+                    pred = true
+                else
+                    pred = n.pdfBackgroundPageNo ~= 0
+                end
+            end
+        end
+        if pred or mode == m.FORCE then
```
rolandlo, line 31: `pred` is true if `mode == m.FORCE`
```suggestion
        if pred then
```

plugins/SpaceForNotes/main.lua
```diff
@@ -0,0 +1,36 @@
...
+            if pred then
+                -- check if page p+1 is also a pdf page (makes the operation idempotent)
+                local n = struct.pages[p+1]
+                -- no p is the last page -> also insert an empty one
+                if n == nil then
+                    pred = true
+                else
+                    pred = n.pdfBackgroundPageNo ~= 0
+                end
+            end
```
rolandlo, line 29: Simpler:
```suggestion
            nextPage = struct.pages[p+1]
            pred = page.pdfBackgroundPageNo ~= 0 and (nextPage == nil or nextPage.pdfBackgroundPageNo ~= 0)
```

rolandlo (commented):
I have some simplifications. Otherwise this looks good to me.

plugins/SpaceForNotes/main.lua
atticus-sullivan: Oh true, haven't thought about it. Not quite sure if the shorter form is maybe too condensed to be easily understood. Maybe we should shorten, but also add a comment to explain this.

plugins/SpaceForNotes/main.lua
atticus-sullivan: Oh yes I see. Still maybe we should keep the comments (in some form) so that the intention can be reconstructed. (also `nextPage` should be `local`)

rolandlo (approved):
LGTM. Please squash your commits into one commit.

atticus-sullivan:
Done. I guess after merging, we can also close https://github.com/xournalpp/xournalpp/issues/5285

rolandlo:

rolandlo:
Thanks for the useful plugins you contributed @atticus-sullivan

Issue #1647, opened 2019-12-03
LaTeX Beamer and Xournalpp


gyps:
It would be great if one could compile a latex beamer presentation into a xournalpp-file. With overlay slides as overlays in xournalpp. Since I don't know anything about the xournalpp file format I don't know if that would be hard to implement or even possible at all.

LittleHuba:
We are currently working on a completely new fileformat. Follow our organisation on Github to get a notification once we release the repository for the new library.

What would most likely be the easiest way is to internally render the beamer presentation as PDF and then use it just as background. Is that what you would like to have or did I missunderstand you?

dbosk:
Beamer overlays allows certain content on a slide to be hidden and shown as progressing through a slide. Like animations. Say that we have a bullet list. First we show the first bullet, whereas the second is hidden. Then we show the second, and so on. Beamer solves it by rendering the slide several times: one PDF page of the slide with the first visible, the second hidden; the following PDF page is the same slide but with the first bullet hidden and second shown.

My interpretation of his request is to detect this and keep them as the same slide. Thus any annotations made while the first bullet is shown, would still be visible when the second is shown instead. However, this would probably require Beamer to generate xournalpp-files instead of PDFs. Or adding some magic in xournalpp to detect this, similarly to how [pdfpc](https://pdfpc.github.io/) detects it.

gyps:
> My interpretation of his request is to detect this and keep them as the same slide. Thus any annotations made while the first bullet is shown, would still be visible when the second is shown instead. 

That's exactly what I meant (thanks for clarifying)

> However, this would probably require Beamer to generate xournalpp-files instead of PDFs. Or adding some magic in xournalpp to detect this, similarly to how [pdfpc](https://pdfpc.github.io/) detects it.

Both options are possible but option 2 has probably the additional advantage that if xournalpp were aware of additional information provided within the PDF anotation could be linked to tags in the PDF so after recreating the PDF they would still be associated with the correct slides.

LittleHuba:
Making Beamer generate xournalpp-files is not really a good option. The new format is not as simple as the current one. Also updating the slides would mean you loose all annotations. The new format also has the paradigm of not modifying attachments and stroing attachments in a way that they can be exchanged later on without issue. That is not really possible with this option.

[...]

PR #5837, opened 2024-09-01
luaAPI: obtain page labels / plugin for beamer presentations

atticus-sullivan:
Adds a function `getPageLabel` to the luaAPI to obtain the page label of a page of the underlying pdf background.
For this to be possible, I needed to add a function to `XojPdfPage`. Please comment on if I did that correctly (note that I'm converting the `gchar*` to a `std::string`).

As the motivation for why one would like to have this functionality, I implemented a plugin which allows to step through a beamer presentation slide by slide (not page by page). See https://github.com/xournalpp/xournalpp/issues/1647#issuecomment-2323332765

Or should I split this into two PRs? (one adding to the luaAPI and one adding the plugin)


atticus-sullivan:
To further illustrate the idea behind the plugin:
Peek 2024-09-01 20-20

plugins/BeamerPresentation/main.lua
```diff
@@ -0,0 +1,46 @@
...
+end
+
+function Beamer(mode)
+    local struct = app.getDocumentStructure()
+    local diff
+    if mode == m.inc then
+        diff = 1
+    elseif mode == m.dec then
+        diff = -1
+    end
```
rolandlo, line 19: I would prefer to write this in one line, but it's a matter of style preference.
```suggestion
    local diff = (mode == m.inc) and 1 or -1
```

plugins/BeamerPresentation/main.lua
```diff
@@ -0,0 +1,46 @@
...
+
+
+    local pCurr = app.getPageLabel(struct.pages[struct.currentPage].pdfBackgroundPageNo)
+
+    local pNext = app.getPageLabel(struct.pages[struct.currentPage].pdfBackgroundPageNo+diff)
+
+    if pCurr and pNext and pCurr == pNext then
+        app.changeBackgroundPdfPageNr(diff, true)
+        app.refreshPage()
+        return
```
rolandlo, line 29: The function returns anyway after this.
```suggestion
```

plugins/BeamerPresentation/main.lua
```diff
@@ -0,0 +1,46 @@
...
+    else
+        app.scrollToPage(diff, true)
+        return
```
rolandlo, line 32:
```suggestion
```

src/core/plugin/luapi_application.h
```diff
@@ -2280,6 +2280,35 @@ static int applib_scrollToPos(lua_State* L) {
...
+    const size_t pagenr = static_cast<size_t>(luaL_checkinteger(L, 1)) - 1;
+    const size_t last = doc.getPageCount() - 1;
+
+    if (pagenr > last) {
+        lua_pushnil(L);
+        lua_pushfstring(L, "page nr %d is out of range", pagenr);
+        return 2;
+    }
+
+    std::string label = doc.getPage(pagenr)->getPageLabel();
```
rolandlo, line 2307: Strictly speaking we should lock the document during access here, right? There are other places in the Lua API code though, where we also don't do it.

plugins/BeamerPresentation/plugin.ini
```diff
@@ -0,0 +1,15 @@
+[about]
+## Author / Copyright notice
+author=Lukas Heindl
+
+description=Step through LaTeX beamer presentation
```
rolandlo, line 5: Maybe add a bit more info on how it works.

rolandlo (commented):
Thank you for the contribution. The plugin has its use case. Some more explanations are needed for those who only find the plugin without its intended use. I suggest to add them to the `plugin.ini`.

Are you still aware of and fine with the fact that we will eventually move most plugins into a separate repository?

plugins/BeamerPresentation/plugin.ini
atticus-sullivan: True. I always feel the way we're presenting this setting in the plugin manager (only one line) this should be short. Also to not clutter the overview page of the plugin manager (maybe adding a new setting (something like `usage`) which is only shown when clicking on the plugin in the plugin manager would solve this, but that's not for this PR).
I'll make the description a bit longer (probably explaining that this plugin uses one page per slide...)

src/core/plugin/luapi_application.h
atticus-sullivan: Yea we had this on [gitter once](https://matrix.to/#/!mIWOZbAzBNmtxiyIii:gitter.im/$-z71mhwupQ1YqkAnGLlTrNYmNvXegvq0m7Wq6rHu0OQ?via=gitter.im&via=matrix.org&via=beeper.com). Tbh I didn't check what happened afterwards, but my POV on this was we'd try to solve this more general than inserting locks in the `luapi_application.h`. But you might argue that this leads to unstable behavior and we should do our best in the luapi to do the locking (and only after we fixed this more generally we should remove the lock statements).
Coming from the *avoid crashes* side, I think I'd be in favor to add the locks in the luapi. Would be a new PR in my opinion though.

plugins/BeamerPresentation/main.lua
atticus-sullivan: True, the idea was to keep this portion of the code expandable in case a new mode is added (and with three or more cases the oneliner gets pretty hard to read in my opinion). Still, in the current state, the oneliner is nicer I agree (and we can always expand this if adding new modes).

atticus-sullivan:
> Are you still aware of and fine with the fact that we will eventually move most plugins into a separate repository?

Yes I know. I'm fine with this.

plugins/BeamerPresentation/plugin.ini
atticus-sullivan: Should be addressed.

src/core/plugin/luapi_application.h
atticus-sullivan: Tbh, I'm not sure which functions need locking. For now I just said every set referring to the document (not the control etc) should acquire a lock before executing the action. (Not sure if it's correct to spare the getters)
First suggestion at https://github.com/atticus-sullivan/xournalpp/tree/luapi-locking

Any comments? (I made a draft PR here https://github.com/xournalpp/xournalpp/pull/5888)

plugins/BeamerPresentation/plugin.ini
```diff
@@ -0,0 +1,15 @@
+[about]
+## Author / Copyright notice
+author=Lukas Heindl
+
+description=Step through LaTeX beamer presentation. Typical workflow: First use the function which removes all pdf-pages related to an animation. Then, step through the slides with the specialized functions instead of manually going forward/backwards. The plugin will decide whether to in-/decrese the background page (for the animation) or go to the next slide.
```
rolandlo, line 5:
```suggestion
description=Step through LaTeX beamer presentation. Typical workflow: First use the function which removes all pdf-pages related to an animation. Then, step through the slides with the specialized functions instead of manually going forward/backwards. The plugin will decide whether to in-/decrease the background page (for the animation) or go to the next slide.
```

src/core/plugin/luapi_application.h
```diff
@@ -2280,6 +2280,35 @@ static int applib_scrollToPos(lua_State* L) {
...
+    const Plugin* plugin = Plugin::getPluginFromLua(L);
+    const Control* control = plugin->getControl();
+    const XojPdfDocument& doc = control->getDocument()->getPdfDocument();
+
+    const size_t pagenr = static_cast<size_t>(luaL_checkinteger(L, 1)) - 1;
+    const size_t last = doc.getPageCount() - 1;
+
+    if (pagenr > last) {
+        lua_pushnil(L);
+        lua_pushfstring(L, "page nr %d is out of range", pagenr);
```
rolandlo, line 2303:
```suggestion
        lua_pushfstring(L, "page nr %zu is out of range", pagenr);
```

src/core/plugin/luapi_application.h
```diff
@@ -2280,6 +2280,35 @@ static int applib_scrollToPos(lua_State* L) {
...
+ **/
+static int applib_getPageLabel(lua_State* L) {
+    const Plugin* plugin = Plugin::getPluginFromLua(L);
+    const Control* control = plugin->getControl();
+    const XojPdfDocument& doc = control->getDocument()->getPdfDocument();
+
+    const size_t pagenr = static_cast<size_t>(luaL_checkinteger(L, 1)) - 1;
+    const size_t last = doc.getPageCount() - 1;
+
+    if (pagenr > last) {
```
rolandlo, line 2301: This looks safer (just for case there is no pdf page)
```suggestion

    if (pagenr >= doc.getPageCount()) {
```

rolandlo (changes requested):
I have some more comments. It's fine for me if we leave the document locking to the other PR.

rolandlo (approved):
Please squash the commits into one commit for the API addition and one commit for the plugin.

PR #5843, opened 2024-09-07
Fix selection order bring forward action for plugins

Rapha149:
In the file "ActionBackwardCompatibilityLayer.cpp" the action for bringing the selection to the front was duplicated instead of the second one being to bring the selection forward.


Rapha149:
No problem :)

PR #5845, opened 2024-09-07
Add methods app.getZoom and app.setZoom to plugin api.

Rapha149:
I noticed that while you could zoom in and out using the actions, you couldn't set the zoom to a specific level or get the current zoom level using the plugin api.
In this pull request I added the methods `app.getZoom` and `app.setZoom` to the `luapi_application.h` file.
The `app.getZoom` method calls `ZoomControl::getZoomReal` and returns it as a number, the `app.setZoom` calls the same methods to set the zoom level that are also used in [ActionProperties.h](https://github.com/xournalpp/xournalpp/blob/2e9c84a88271d18e64ca64e293a6521f13bb2c3c/src/core/control/actions/ActionProperties.h#L430) (this also means first multiplying the supplied zoom level by the `zoom100Value` which means that you set the real zoom value using the api, which corresponds to getting the real zoom value with `app.getZoom`)

I hope I did everything correctly concerning the code style etc., this is my first time with Lua and C++ :sweat_smile:

### DIFF
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -710,6 +710,21 @@ function app.scaleTextElements(factor) end
 --- Example: app.getDisplayDpi()
 function app.getDisplayDpi() end
 
+--- Gets the current zoom.
+--- 
+--- @return double current zoom level
+--- 
+--- Example: app.getZoom()
+function app.getZoom() end
+
+--- Sets the current zoom.
+--- 
+--- @param zoom number
+--- 
+--- Example: app.setZoom(2.5)
+--- Changes zoom level to 2.5
+function app.setZoom(zoom) end
+
 --- Exports the current document as a pdf or as a svg or png image
 --- 
 --- @param opts {outputFile:string, range:string, background:string, progressiveMode: boolean}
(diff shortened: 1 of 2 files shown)


Rapha149:
I don't know why that check failed, I'm not even sure what it checks for, could someone explain what I did wrong?

rolandlo:
You must run the python script in `scripts/lua_def_file.py`. It will make some changes to the docs of the Lua API in `plugins/luapi_application.def.lua` that you should add to your commit introducing new API.

Rapha149:
Thanks for the info! Now it should be fine.

rolandlo (approved):
Thanks for your contribution. The code looks all good to me.
Please squash the two commits into one.

Rapha149:
Thanks for approving! 
Do you mean that I should squash the commits before merging somehow or that the person who merges should squash them?

rolandlo:
> Thanks for approving! Do you mean that I should squash the commits before merging somehow or that the person who merges should squash them?

I would prefer that you squash the commits before I do the "Rebase and merge"ing.

The easiest way to do it is `git rebase -i HEAD~2` and mark the second commit as fixup.

Rapha149:
Ok, I will do that this evening, I'm not at home at the moment.

Rapha149:
Commits are squashed

rolandlo:

Robin-Sch:
This is not working (anymore) for me, I get the following error:
"Attempt to call a nil value (field 'setZoom')

PR #5849, opened 2024-09-09
Update xournalpp.spec

luyatshimbalanga:
This update also adds color palettes support in the Fedora spec file.


rolandlo (approved):
Thanks for the update. LGTM

PR #5864, opened 2024-09-14
Fix UB with GTK signals

bhennion:
- Some callbacks were returning void while they should have returned gboolean, leading to inconsistent behaviour -- see for instance https://github.com/xournalpp/xournalpp/discussions/5842, but the UB was more spread than just there.
- GtkRange::change-value is also replaced by GtkRange::value-changed (which is more suited for our uses)

As a side note, I've been working on a way to enforce at compile time that the signals callbacks have the right type, but that'll be for the master branch.

### DIFF
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -101,8 +101,11 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
                      this);
 
     // "watch over" all key events
-    g_signal_connect(this->window, "key-press-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
-    g_signal_connect(this->window, "key-release-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
+    auto keyPropagate = +[](GtkWidget* w, GdkEvent* e, gpointer) {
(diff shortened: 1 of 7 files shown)


src/core/gui/MainWindow.cpp
```diff
@@ -101,8 +101,11 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
                      this);
 
     // "watch over" all key events
-    g_signal_connect(this->window, "key-press-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
-    g_signal_connect(this->window, "key-release-event", G_CALLBACK(gtk_window_propagate_key_event), nullptr);
+    auto keyPropagate = +[](GtkWidget* w, GdkEvent* e, gpointer) {
```
rolandlo, line 104: I don't understand why you use `GdkEvent* e` instead of `GdkEventKey* e` here, when the signature in the docs is 
```c
gboolean
key_release_event (
  GtkWidget* self,
  GdkEventKey event,
  gpointer user_data
)
```

rolandlo (approved):
I have one question. The rest looks good to me.

src/core/gui/MainWindow.cpp
rolandlo: Ok, thanks for the explanations. Please go ahead and merge.

PR #5888, opened 2024-09-25
Draft: [luapi] first suggestion for locking the document when doing changes

atticus-sullivan:
See https://matrix.to/#/!mIWOZbAzBNmtxiyIii:gitter.im/$-z71mhwupQ1YqkAnGLlTrNYmNvXegvq0m7Wq6rHu0OQ?via=gitter.im&via=matrix.org&via=beeper.com and https://github.com/xournalpp/xournalpp/pull/5837#discussion_r1769972525

At the moment, we do not lock the document when we make changes to it in the lua api. This might lead to unexpected behavior and crashes when using the api functions often enough.

I'm not quite familiar with the entire codebase, so I'm not quite sure when we need to lock. I assumed that when calling full actions (like `ui action` or `sidebar action`) locking is already being done.

When manually adding elements to the document, the reasoning is to lock before all elements are inserted once and wait until the undo/redo action is inserted until releasing the lock. Not quite sure if this is needed/wanted.

Also for now I spared the getters which should be fine I guess, since they acquire the structures once and then convert them to lua. If the document changes in the meantime, the information of course will be outdated, but there's nothing to prevent this.

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -862,6 +862,7 @@ static int applib_addSplines(lua_State* L) {
     // -1 = splines
 
     size_t numSplines = lua_rawlen(L, -1);
+    ctrl->getDocument()->lock();
     for (size_t a = 1; a <= numSplines; a++) {
         std::vector<double> coordStream;
         auto stroke = std::make_unique<Stroke>();
@@ -870,6 +871,7 @@ static int applib_addSplines(lua_State* L) {
         lua_gettable(L, -2);                 // get current spline from splines table
         lua_getfield(L, -1, "coordinates");  // get coordinates of the current spline
         if (!lua_istable(L, -1)) {
+            ctrl->getDocument()->unlock();
             return luaL_error(L, "Missing coordinate table!");
         }
         size_t numCoords = lua_rawlen(L, -1);
@@ -885,6 +887,7 @@ static int applib_addSplines(lua_State* L) {
         // Handle those points
         // Check if the list is divisible by 8.
         if (coordStream.size() % 8 != 0) {
+            ctrl->getDocument()->unlock();
             return luaL_error(L, "Point table incomplete!");
         }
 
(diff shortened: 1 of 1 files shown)


src/core/plugin/luapi_application.h
```diff
@@ -862,6 +862,7 @@ static int applib_addSplines(lua_State* L) {
     // -1 = splines
 
     size_t numSplines = lua_rawlen(L, -1);
+    ctrl->getDocument()->lock();
```
eldipa, line 865: I cannot opinion about the need of a lock or of the placement of them but I definitely I would like to see a RAII object for this.

In the `Document` class we could have a method:

```cpp
std::unique_lock<std::mutex> takeLock() { return std::unique_lock<std::mutex>(this->documentLock); }
```

Then, in the places where you wrote `...->getDocument()->lock();` you should have `auto lck = ...->getDocument()->takeLock();`
And with this all the `unlock` calls could be removed.

This is not only simpler and cleaner, but also safer. It is much harder to forget an unlock when the RAII object does it for you.

atticus-sullivan: Yea, fixing the locking issue globally definitely should be the goal in the end (I think this also is what they meant on gitter/matrix some time ago, not sure what exactly the progress is).
To avoid crashes in the meantime, I'm trying to add some locking to the luapi in its current state.

rolandlo:
> Also for now I spared the getters which should be fine I guess, since they acquire the structures once and then convert them to lua. If the document changes in the meantime, the information of course will be outdated, but there's nothing to prevent this.

So what happens in this loop in `app.getDocumentStructure` https://github.com/xournalpp/xournalpp/blob/baa7e5c3c9192aa95ec90186afdaf1c806beed9b/src/core/plugin/luapi_application.h#L2139-L2143
when a page has been deleted from a different thread while the loop is running? Or is that not a possible scenario?

atticus-sullivan:
Oh yes that might be an issue. Looking at
https://github.com/xournalpp/xournalpp/blob/baa7e5c3c9192aa95ec90186afdaf1c806beed9b/src/core/model/Document.cpp#L422-L431
checking the returned `page` for `nullprt` (and probably `break` the loop) should be enough I think.

I'm also not entirely sure if this can happen, but why shouldn't it be a possible scenario? (and so we should harden the implementation against it) -- Good catch :+1:

Issue #5882, opened 2024-09-23
macOS Silicon (ARM64) support


aidan-gibson:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

homebrew

### Xournal++ Version

1.2.3

### libgtk Version

n/a

### Bug Description

macOS release for silicon

### Expected Behaviour

My apps to run natively

### Steps to Reproduce

`brew install --cask xournal++`
Open it and frown

### Additional Context

pls <3

rolandlo:
For next release (1.2.4) we plan to add an official MacOS ARM build. For now there is an ARM build linked in this comment https://github.com/xournalpp/xournalpp/issues/5643#issuecomment-2083480167

PR #5783, opened 2024-07-29
Add Ubuntu 24.04 to builds

rolandlo:
Closes #5782

PR #5897, opened 2024-09-29

bhennion:
Only very simple fixes were added since the last merge. Merging this once the CI clears.


rolandlo:
Can you wait one moment and then include #5783 which I will rebase quickly on release-1.2?

rolandlo:
> Yes of course. Maybe we wanna tackle #5882 as well?

Ah, right. Let's me check what needs to be included, so the cirrus pipeline works as it should.

PR #5898, opened 2024-09-29
Cirrus pipeline task

rolandlo:
This PR is based on work of @Technius and adds a Cirrus configuration file and build task in order to create a MacOS ARM build. The task is executed on https://cirrus-ci.com/ and on a MacOS Ventura runner.

The lock file (and possibly GITHUB token) still need to be updated to the latest versions and tested.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -63,6 +63,8 @@ build_xournalpp_task:
     cmake .. -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo
     cmake --build .
     cmake --install . --prefix /opt/devroot/gtk/inst
+  pkgconf_symlink_script: | # used for gtk-mac-bundler
+    ln -sf $(which pkgconf) /opt/devroot/gtk/inst/bin/pkg-config
   package_script: |
     (cd mac-setup && bash build-app.sh ~/gtk && bash build-dmg.sh Xournal++.app Xournal++.dmg)
   xournalpp_artifacts:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 macos_instance:
-  image: ghcr.io/cirruslabs/macos-ventura-base:latest
+  image: ghcr.io/cirruslabs/macos-runner:sonoma
 
 # Need to override the user HOME so we can cache the jhbuild stuff.
 env:
@@ -32,7 +32,7 @@ build_binary_blob_task:
   upload_caches:
     - binary_blob
   upload_binary_blob_artifacts:
-    path: blobdir
+    path: "blobdir/*"
 
 build_xournalpp_task:
   << : *UNINSTALL_HOMEBREW
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,69 @@
+macos_instance:
+  image: ghcr.io/cirruslabs/macos-ventura-base:latest
+
+# Need to override the user HOME so we can cache the jhbuild stuff.
+env:
+  HOME: /opt/devroot
+
+# Define a template that is reused for both binary blob task & build task.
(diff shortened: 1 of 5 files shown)


rolandlo:
The cirrus build works again now. I'm waiting for my MR https://gitlab.gnome.org/GNOME/gtk-osx/-/merge_requests/60 to be merged, so that I can basically drop the last commit.

rolandlo:
The cirrus-ci is now integrated into this repository. Pushing to it will trigger a build on https://cirrus-ci.com/github/xournalpp/xournalpp, similarly as the Azure Pipelines. It takes roughly 6 minutes, unless the binary blob has to be regenerated after modifying the file `jhbuild-version.lock`, in which case it takes about another 30 minutes.

rolandlo:
I have also created a fine grained PAT (personal access token) for the upload task with the required permissions. It was encrypted on the settings page of this repository on cirrus-ci.com. The encrypted PAT is exposed in the .cirrus.yml

rolandlo:
> The cirrus build works again now. I'm waiting for my MR https://gitlab.gnome.org/GNOME/gtk-osx/-/merge_requests/60 to be merged, so that I can basically drop the last commit.

The MR to gtk-osx has been merged and this PR is now ready. I'll wait for the pipelines to succeed. Any comments/objections?

rolandlo:
> I have a couple of questions/remarks:
> 
>     * Is the binary blob going to be rebuilt for every commit of every PR, or is it stored somewhere?

The binary blob is stored on cirrus-ci. It is only rebuilt if either we change 
- the `gtk-osx` commit we use
- the `xournalpp-pipeline-dependencies` commit (for the python script building the blob) we use
- something in `xournalpp.modules`

So that's not very frequently.

> 
>     * Could we give the CI tasks a more explicit name?
>       
>       * build_binary_blob -> Build blob for MacOS ARM
>       * build_xournalpp -> Build Xournal++ for MacOS ARM

Yes, we can give it more explicit names. Let me try.
> 
>     * The CI seems to only build/install Xournal++. It should also run the tests.
> 
> 
I can try to add a task for that.

> I also have a question about Cirrus itself. There "Free for OSS" plan cap the computing time
> 
> > Additionally there is an upper monthly limit on free usage equal to 50 [compute credits](https://cirrus-ci.org/pricing/) (which is equal to 10,000 CPU-minutes for Linux tasks or 500 minutes for macOS tasks which always use 4 CPUs).
> 
> If we only spend <5 minutes per build, that should be fine, but if the blob is rebuilt every time, that's not gonna work.

As mentioned above the blob is not rebuilt often. The `build_xournalpp` task took 3 minutes.

rolandlo:
Ok, so it was 48 seconds for the build_binary_blob task (which reused the previously built blob) and 3 minutes 48 seconds for the build_xournalpp task:

Bildschirmfoto vom 2024-10-18 10-31-28


✅ 00:08 clone
✅ 00:27 uninstall_homebrew
✅ 00:01 create_devroot
✅ 00:03 binary_blob
✅ 00:00 Upload 'binary_blob' cache
✅ 00:03 upload_binary_blob


✅ 00:11 clone
✅ 00:38 uninstall_homebrew
✅ 00:01 create_devroot
✅ 00:02 binary_blob
✅ 00:02 unpack
✅ 00:01 install_ninja
✅ 01:51 build_and_test
✅ 00:01 pkgconf_symlink
✅ 00:51 package
✅ 00:00 xournalpp
✅ 00:00 Upload 'binary_blob' cache

rolandlo:
The tests were successful by the way.
If it turns out that we run out of free compute credits we could try to make the packaging optional and avoid uninstalling homebrew (which messes with jhbuild) in the build_binary_blob task when the blob doesn't need to be regenerated. Actually I'm not sure it is needed in the build_xournalpp task. At least it shouldn't be needed without the packaging.

PR #5922, opened 2024-10-06
Fix crash whenentering presentation mode with tiny page size

### LINKED ISSUE #2943: Crash when entering presentation mode with a small page size
**Affects versions :**
 - OS: Arch linux repository package, and also latest appimage. (both on X11)
 - gtk 3.24.26
 - Xournal++: 1.0.20-hotfix and latest appimage

**Describe the bug**

When opening a pdf that has a small page size, if I enter full screen and presentation mode, xournal++ crashes with a SIGSEGV. In particular, I've been able to reproduce this with a pdf for which presentation mode would require more than the maximum 700% zoom.

**To Reproduce**
1. Open the attached pdf
2. Enter presentation mode and full screen
3. crash (if it turns out that the 700% zoom threshold is relevant, this step might depend on your monitor resolution)

[xournal-crash.pdf](https://github.com/xournalpp/xournalpp/files/6135054/xournal-crash.pdf)

bhennion:
Fix #2943 

The issue was: when entering presentation mode, if the required zoom value was bigger that the max (700%), we entered an infinite loop.

### DIFF
--- a/src/core/control/zoom/ZoomControl.cpp
+++ b/src/core/control/zoom/ZoomControl.cpp
@@ -264,10 +264,11 @@ auto ZoomControl::getZoom() const -> double { return this->zoom; }
 auto ZoomControl::getZoomReal() const -> double { return this->zoom / this->zoom100Value; }
 
 void ZoomControl::setZoom(double zoomI) {
+    zoomI = std::clamp(zoomI, this->zoomMin, this->zoomMax);
     if (this->zoom == zoomI) {
         return;
     }
-    this->zoom = std::clamp(zoomI, this->zoomMin, this->zoomMax);
+    this->zoom = zoomI;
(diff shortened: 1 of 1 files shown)


rolandlo (commented):
I can't reproduce the issue on latest master. When entering presentation mode with a tiny page size the zoom is set to 700% (which then doesn't fill the entire space, but is the best thing to do in my opinion). With the PR the zoom doesn't change when entering presentation mode (and may remain pretty small). So I'm not sure this is the way to go. Can you still reproduce the crash @bhennion?

rolandlo:
Right, I can reproduce the crash on the release branch as well. The behaviour on master seems better than the one in this PR though.

By the way "when entering" should be two words.

rolandlo (approved):
LGTM

Issue #5910, opened 2024-10-02
dragging selection to next page generates crash


lurix66:
### Operating System

Linux

### (Linux only) Distribution

Unbuntu

### (Linux only) Desktop Environment

Gnome

### (Linux Only) Display Server

X.Org

### Installation Method

Snap

### Xournal++ Version

xournalpp 1.2.3 (c6970f47)

### libgtk Version

libgtk: 3.24.38

### Bug Description

when using rectangular selection, copying the selected (copied, or cut) rectangle and hitting the page boundary, it frequently crashes.
Also when reviving the session from the autosaved document, it happens that also the selection is revived, it is in close contact with the page boundary, so it crashes again 

### Expected Behaviour

it should not crash. It might be good to beep, or to traverse boundaries and jump to the next page, or ask the user what he wants. Including asking the user to undo the last movement of the selection or the selection itself. 

### Steps to Reproduce

select a rectangular area, drag it down towards the page boundary or try to move it to the next page.

### Additional Context

maybe also other kind of selections (lazo?), or any kind of selection? would generate the malfunction

Issue #1232, opened 2019-06-11
Crash/error/freeze if undo is pressed when selection is on page boundary 


Technius:
**Affects versions :**
 - OS: Linux
 - Version of Xournal++: master as of 99b7e18, most likely in earlier versions too

**Describe the bug**
Pressing an undo button when a selection is between two pages will cause an error (not user visible) and do nothing else. It may also cause the entire application to freeze (which I have not been able to reproduce consistently yet). If there is no freeze, a crash will occur when changing documents. The resulting emergency save will have no pages when loaded (no emergency save is produced if the application freezes).

**To Reproduce**
1. Create a blank document and add a second page.
2. Create some strokes on the first page.
3. Select the stroke and move it on to the page boundary so that it is mostly on the second page but still on the first page.
4. Trigger an undo (e.g. by pressing <kbd>Ctrl</kbd><kbd>z</kbd>)
5. Observe the error in the console (or a freeze).
6. If there is no freeze, create a new document.
7. Observe a crash.

**Expected behavior**
The undo should occur successfully without any errors.

**Screenshots of Problem**
![xournal_crash](https://user-images.githubusercontent.com/1066652/59314159-dd5f2b80-8c68-11e9-9360-b344ca34fcce.gif)

**Additional context**
[crashlog.log](https://github.com/xournalpp/xournalpp/files/3278979/crashlog.log)

LittleHuba:
I can explain some background on this problem. It is mainly caused by the active page system.
Objects on pages are not allowed to be placed with negative coordinates for whatever reason. Thus moving the stroke selects the second page as active page and the stroke returns to the original position as its new coordinates would be negative.
If you try this with the mouse staying on page one everything should work.

So this has two fixes:
1) Allow negative coordinates (probably the best as this would be what one expects)
2) Fix that active page system to still assign the correct page

lehmanju:
@Technius Is this still reproducable? I can't seem to reproduce the bug. In your screenshot it also seems the page selection is not quite working, in the sidebar the first page is selected whereas the second page is selected in the main view.

Technius:
Yes, I still get the error in the console as well as a segfault in current master.

rolandlo:
I can't reproduce this bug any more. It must have be fixed.

Technius:
Assuming that there are rare cases where it is possible to place a selection "out" of a page, this bug can still occur (as evidenced by #2889, which isn't really a "rare" case but demonstrates the possibility).

[...]

PR #5936, opened 2024-10-13
Fix crash when undoing selection moves across pages

### LINKED ISSUE #5931: Crash after undoing several moves
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

Gnome

Wayland

### Installation Method

Arch Linux official repository

### Xournal++ Version

1.2.3 (c6970f47)

### libgtk Version

3.24.43

### Bug Description

I selected an object, and used the arrow keys to move it to the previous page. Then I held down Ctrl-Z to undo the moves. Xournal++ crashed while undoing.

### Expected Behaviour

Don't crash

### Steps to Reproduce

1. Open a new document
2. Create a new page
3. Draw something on the second page
4. Select the object
5. Use up arrow to move object to previous page
6. Hold Ctrl-Z. The object will disappear as soon as you press Ctrl-Z. The second page will disappear after holding a bit.
7. Draw something
8. Press Ctrl-Z. Xournalpp should crash as soon as you press it.

### Additional Context

[Screencast From 2024-10-10 16-59-07.webm](https://github.com/user-attachments/assets/1e4d942f-bab9-4021-a5ed-ea7cad461c50)

[errorlog.20241010-165924.log](https://github.com/user-attachments/files/17335468/errorlog.20241010-165924.log)

bhennion:
Fixes #5931 #5910 and #1232

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -217,6 +217,12 @@ void EditSelection::finalizeSelection() {
         this->snappedBounds.x = this->x + ox;
         this->snappedBounds.y = this->y + oy;
         v = this->contents->getSourceView();
+
+        PageRef page = v->getPage();
+        Layer* layer = page->getSelectedLayer();
+        // Create an Undo action to compensate - avoids Segfault/Freeze if the user presses undo after this happened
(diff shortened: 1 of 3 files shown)


rolandlo (commented):
The issue is fixed, I agree. I have only an objection about the debug messages (parameter n).

PR #5954, opened 2024-10-17
Merge the release branch in master

bhennion:
This was a complicated merge, so please check carefully.

1- Changes to .po/.pot files: I used `git checkout --ours po/*`. The idea being: those will be updated by the CI/the next "New Crowdin updates" PR targetted at master
2- Changes from #5950 are essentially reverted in the merge commit - this was a backport and the fix is not needed on master - @rolandlo is that right?

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -273,6 +273,12 @@ void EditSelection::finalizeSelection() {
         this->snappedBounds.x = this->x + ox;
         this->snappedBounds.y = this->y + oy;
         v = this->contents->getSourceView();
+
+        PageRef page = v->getPage();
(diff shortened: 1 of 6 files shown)


rolandlo:
> This was a complicated merge, so please check carefully.
> 
> 1- Changes to .po/.pot files: I used `git checkout --ours po/*`. The idea being: those will be updated by the CI/the next "New Crowdin updates" PR targetted at master 

That makes sense.

> 2- Changes from #5950 are essentially reverted in the merge commit - this was a backport and the fix is not needed on master - @rolandlo is that right?

Yes, that's correct.

The pipeline didn't succeed, so something must still be fixed.

rolandlo (approved):
LGTM
We might include #5898 though for the ARM Mac support if you agree with merging it.

PR #5958, opened 2024-10-18

rolandlo:
To be pushed from CLI.


rolandlo:
@bhennion Any objections against merging this (by pushing via CLI)? I have tested the MacOS Arm build and it works fine.

PR #5962, opened 2024-10-19
Fix some translatable string and improve context

bhennion:
This is based on mistakes/questions flagged by translators on Crowdin

This must NOT be merged before the 1.2.4 release: we need to give time to translators to do their deed.


rolandlo (approved):
LGTM. Thanks for taking care of those issues.

PR #5968, opened 2024-10-20
Add Github actions for Build/Test

bhennion:


rolandlo:
> I ported all the continuous integration tests to Github Actions. As you can see in the CI results, it turns out to run about twice as fast as the same scripts on the Azure Pipeline. Porting is also pretty easy: there is plenty of documentation and a lot of reusable actions for many tasks (e.g. setup-msys2 which caches the packages you ask it to install, things like that).
> 
> Depending on what @xournalpp/core thinks, I can move forward and migrate the remaining things.

Yes, please go ahead. That looks like a nice improvement.

rolandlo:
> Does anyone know why we rely on jhbuild for the MacOS blobs and not on Homebrew?

While it's easy to build using Homebrew, the binaries are not easily relocatable, as Homebrew is quite careless about where it gets its dependencies. See https://gitlab.gnome.org/GNOME/gtk-mac-bundler/-/issues/8

rolandlo:
> Thanks. Another question: is there a reason the scripts for MacOS and MacOS ARM are different other than the fact they ran on different platforms (thus requiring different syntax)? Could `mac-setup/cirrus-jhbuild.sh` be used to generate the blobs for the Intel chips?

`mac-setup/cirrus-jhbuild.sh` could also be used for MacOS Intel. On cirrus-ci there are just no MacOS Intel containers https://cirrus-ci.org/guide/macOS/

rolandlo:
> And how are https://github.com/xournalpp/xournalpp-pipeline-dependencies/raw/macos-10.15-jhbuild/gtk/gtk-bin.tar.gz.aa and https://github.com/xournalpp/xournalpp-pipeline-dependencies/raw/macos-10.15-jhbuild/gtk/gtk-bin.tar.gz.ab generated? I can't find which script built that. Is there something on the Azure Pipelines that generates that?

That was built on my old MacBook Pro (on MacOS Catalina) following the steps in the Wiki. It takes ages (several hours) and sometimes there are issues to be solved along the way.

rolandlo:
> > That was built on my old MacBook Pro (on MacOS Catalina) following the steps in the Wiki. It takes ages (several hours) and sometimes there are issues to be solved along the way.
> 
> I see. So it would make sense to use the same CI as for the ARM builds then. 

Yes, indeed.

> Should we make the MacOS gtk blobs available for devs to avoid building GTK themselves? Would that work?

Yes, I think that should work.

rolandlo:
Maybe the patch in https://bugs.launchpad.net/intltool/+bug/1197875 could solve the problem.

rolandlo:
> Actually, it was simpler than that. There is only one version of perl already (jhbuild doesn't add another one). Just adding
> 
> ```
>   - uses: perl-actions/install-with-cpm@v1
>     with:
>       install: XML::Parser
> ```
> 
> to the .yml file did the trick. It just installs the XML::Parser module for perl.
> 
> The build time for the MacOS-13 image isn't great though. Building the blob took 80 minutes. Building Xournal++ once the blob is cached takes 9 minutes.

I'm glad you got it to work. I didn't expect better building time on MacOS Intel. There's a lot of stuff to build and the surprising part is that MacOS ARM is so much faster.

rolandlo:
That's awesome! I think the AppImage should only be generated on the oldest Ubuntu version (20.04 currently) though.

rolandlo:
> I have a couple of questions:
> 
>     1. What's the point of the Windows "portable" version? Or rather, what's the point of the non-portable one which is only 5% smaller?
> 
The installer creates startmenu links and registry entries to properly integrate the application into the desktop. The portable installation doesn't do any of that, but has the advantage that it can be freely moved (also onto a USB stick that you can carry around). I think we should keep both versions.

>     2. For MacOS: should we publish a .zip or a .dmg? The Cirrus CI produced a .dmg, the Azure pipeline a .zip. 
I think it's better to publish a `.dmg`. It makes it easy to install the app in the Applications folder, where apps belong.

rolandlo:
In the portable Windows folder the root folder (e.g. xournalpp-1.2.4-windows-portable) is missing, so the bin, lib and share folders are directly at the top.

rolandlo:
> You may be able to access the generated artefacts here (scroll down) if you have time to test them:
> https://github.com/bhennion/xournalpp/actions/runs/11533575572

I have tested the Windows, MacOS and the Ubuntu 24.04 assets. They all work fine.

rolandlo (approved):
I looked through the workflow files and it looks fine to me.
Just one minor point (somewhat unrelated to this PR though):

The CMake flag `-DCMAKE_DEBUG_INCLUDES_LDFLAGS` is not used anywhere any more. It should have been removed in #2570 I think. Previously there was a section:

```cmake
option (CMAKE_DEBUG_INCLUDES_LDFLAGS "List include dirs and ldflags for xournalpp target" OFF)
mark_as_advanced (FORCE CMAKE_DEBUG_INCLUDES_LDFLAGS)
if (CMAKE_DEBUG_INCLUDES_LDFLAGS)
    message ("Include directories: ${xournalpp_INCLUDE_DIRS}")
    message ("LDFLAGS/LIBRARIES: ${xournalpp_LDFLAGS}")
endif (CMAKE_DEBUG_INCLUDES_LDFLAGS)
```
and now only the first two lines are left, which is pointless. So we should remove everything related to the flag.

rolandlo:
> Oh, there is another question: Should this be merged into the release branch? It's quite a lot of new code but having to maintain both pipelines would be a hassle (provided the Azure pipeline has magically been fixed, btw).

Yes, definitely I think that this should be merged into the release branch. While it's a lot of changes, they are all about the build process and won't affect how Xournal++ runs in case it builds.

rolandlo (approved):
LGTM

rolandlo:
> The command `\action create-installers bla bli blu` now:
> 
>     * no longer works with commas (Is that ok?)

Sure, that's fine. The warning in case of unrecognized inputs is very helpful.

PR #5988, opened 2024-10-27
Use MultiUser.nsi for choosing between all users or current user install

rolandlo:
Remark: The MultiUser.nsi is preinstalled and suits our prupose:
        - In the GUI a admin can choose between installation for all users or current user
        - Silent mode installation works
	It has some limitations though, in particular
        - Privilige elevation is asked even if a user with admin priviliges wants to install
          for current user
        - To work around this issues the NsisMultiUser plugin on
          https://github.com/Drizin/NsisMultiUser could be used.
          It is not preinstalled though and would complicate the build process.

Some remarks about the changes:
- The custom page for choosing the install mode was replaced by the page MULTIUSER_PAGE_INSTALLMODE from the MultiUser plugin
- SetShellVarContext is handled by the MultiUser plugin.
- The RefreshShellIcons macro is already defined in the `FileFunc.nsh`  that is included via `MultiUser.nsh`.

PR #5990, opened 2024-10-28
Use MultiUser.nsi for choosing between all users or current user install

rolandlo:
This is the same PR as #5988. Hopefully the Azure pipeline is triggered this time.

Remark: The MultiUser.nsi is preinstalled and suits our prupose:
- In the GUI a admin can choose between installation for all users or current user
- Silent mode installation works
It has some limitations though, in particular
- Privilige elevation is asked even if a user with admin priviliges wants to install
for current user
- To work around this issues the NsisMultiUser plugin on
https://github.com/Drizin/NsisMultiUser could be used.
It is not preinstalled though and would complicate the build process.

Some remarks about the changes:
- The custom page for choosing the install mode was replaced by the page MULTIUSER_PAGE_INSTALLMODE from the MultiUser plugin
-  SetShellVarContext is handled by the MultiUser plugin.
-  The RefreshShellIcons macro is already defined in the FileFunc.nsh that is included via MultiUser.nsh.

### DIFF
--- a/ui/plugin.glade
+++ b/ui/plugin.glade
@@ -118,6 +118,7 @@
                 <property name="can_focus">False</property>
                 <property name="margin_top">5</property>
                 <property name="margin_bottom">5</property>
+                <property name="wrap">true</property>
                 <property name="label" translatable="yes">Plugin changes are only applied after Xournal++ is restarted</property>
               </object>
               <packing>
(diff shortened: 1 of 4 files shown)


rolandlo:
No, it didn't help to create a new PR. Closing therefore.

PR #5992, opened 2024-10-28
Add wrapping for long labels in plugin manager

rolandlo:
The description of a plugin can be quite long. The whole dialog gets messed up when there is no wrapping. The line `Plugin changes are only applied after Xournal++ is restarted` is quite long as well.

The wrapping is already done on the GTK4 branch (.glade -> .ui conversion)


rolandlo:
Even here the Azure pipeline is not triggered.

PR #6023, opened 2024-11-05
Fix undo and redo when adding strokes from a plugin

### LINKED ISSUE #6015: Undo/redo doesn't work any more for strokes added via a plugin
### Operating System

Linux

### (Linux only) Distribution

Ubuntu 24.10

GNOME

Wayland

### Installation Method

PPA

### Xournal++ Version

1.2.4+dev

### libgtk Version

3.24.43

### Bug Description

When adding strokes via a plugin (`app.addStrokes`) undo doesn't undo the strokes. Git bisection showed that this is a regression from  #5350 (commit 399b10537b596998318f361bf0d610bfc67161fd)
@Febbe Can you see what is wrong?

### Expected Behaviour

Undo/redo works for elements added via a plugin.

### Steps to Reproduce

1. Unzip the contents of [StrokeTest.zip](https://github.com/user-attachments/files/17614175/StrokeTest.zip) into the plugins subfolder of the config folder.
2. Start xournalpp and use "Add a stroke" from the plugin menu. Observe the triangle and square get inserted.
3. Hit undo a couple of times. Observe that the two strokes are not undone (but they are on the stable version 1.2.4).

### Additional Context

_No response_

rolandlo:
Fixes #6015

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -1091,6 +1091,7 @@ static int applib_addStrokes(lua_State* L) {
         }
 
         // Finish building the Stroke and apply it to the layer.
+        strokes.push_back(stroke.get());
         addStrokeHelper(L, std::move(stroke));
         // Onto the next stroke
         lua_pop(L, 1);  // cleanup stroke table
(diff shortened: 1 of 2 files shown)


Febbe:
LGTM, it may be good, to directly reset the elementsOwn vector after the first loop/the document unlocking to prevent confusion.

Febbe (approved):

rolandlo:
Thanks for the review @Febbe. Merging once the CI clears.

Issue #5994, opened 2024-10-29
Export PDF crashes Xournal++


efaulhaber:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

Homebrew

### Xournal++ Version

1.2.4, 2a4ed3fd

### libgtk Version

3.24.38

### Bug Description

This one file causes Xournal++ to crash on PDF export. Other files worked file previously.

### Expected Behaviour

No crash.

### Steps to Reproduce

Log:
```
Date: Tue Oct 29 15:11:16 2024
Error: signal 6
Xournal++ version 1.2.4
Git commit 2a4ed3fd from xournalpp/HEAD
Gtk version 3.24.38

[bt]: (0) 0   xournalpp                           0x0000000102e4b29f _ZL12crashHandleri + 95
[bt]: (1) 1   libsystem_platform.dylib            0x00007ff8084fafdd _sigtramp + 29
[bt]: (2) 2   ???                                 0x0000000000000000 0x0 + 0
[bt]: (3) 3   libsystem_c.dylib                   0x00007ff8083f1a79 abort + 126
[bt]: (4) 4   libc++abi.dylib                     0x00007ff808484922 abort_message + 241
[bt]: (5) 5   libc++abi.dylib                     0x00007ff808477110 _ZL28demangling_terminate_handlerv + 240
[bt]: (6) 6   libobjc.A.dylib                     0x00007ff8081142bf _ZL15_objc_terminatev + 104
[bt]: (7) 7   libc++abi.dylib                     0x00007ff808483d7b _ZSt11__terminatePFvvE + 6
[bt]: (8) 8   libc++abi.dylib                     0x00007ff808486907 __cxa_get_exception_ptr + 0
[bt]: (9) 9   libc++abi.dylib                     0x00007ff8084868ce _ZN10__cxxabiv1L22exception_cleanup_funcE19_Unwind_Reason_CodeP17_Unwind_Exception + 0
[bt]: (10) 10  libc++.1.dylib                      0x00007ff8084321d4 _ZNSt3__121__throw_runtime_errorEPKc + 56
[bt]: (11) 11  xournalpp                           0x0000000102f4d090 _ZNSt3__14__fs10filesystem8_PathCVTIwE14__append_rangeERNS_12basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEEPKwSC_ + 304
[...]

PR #6038, opened 2024-11-09
Replace deprecated codecvt with mbstowcs

Febbe:
- Fixes export
 - Refactored the use of std::string const& with std::string_view
 
 @bhennion can you test this, I couldn't reproduce this


rolandlo:

rolandlo:
\action create-installers mac-arm mac-x64

Febbe:
> I couldn't reproduce either.

The thing is that we expect utf8 strings everywhere, and gtk/glib should produce utf8 strings. Mixing utf8 and ANSI is a bug in any way. If this doesn't fix it, we have a source that produces ANSI strings somewhere. However, getting an ANSI string from a plugin is not guarded. The thing which confuses me is, that it also fails for others, which do not use a plugin to create a path.

Febbe:
\action create-installers mac-arm mac-x64

Febbe:
> @Febbe Could you cherry pick (or copy) the commit https://github.com/bhennion/xournalpp/pull/18/commits/3df9437fa0d81ddec337ac439078429f08fd8f4e of this branch https://github.com/bhennion/xournalpp/tree/testUTF8
> 
> I added a really dumb test and it actually fails on Ubuntu 20, 22 and 24 (but succeeds on MacOS and Windows)
> ```
> [----------] 1 test from DocumentName
> [ RUN      ] DocumentName.testUTF8
> unknown file: Failure
> C++ exception with description "filesystem error: Cannot convert character sequence: Invalid or incomplete multibyte or wide character" thrown in the test body.
> [  FAILED  ] DocumentName.testUTF8 (0 ms)
> [----------] 1 test from DocumentName (0 ms total)
> ```
> (see https://github.com/bhennion/xournalpp/pull/18)

You need to prefix utf8 strings with u8, enforce compilation with utf8 or use code points. Otherwise it's not guaranteed, that the compiler compiles this as utf8.

Febbe:
> And it still fails with your commits.

Yes, with your test I can reproduce it now🥳. Seems like libstdc++ does not define some of the code points.

Febbe:
> If it is too much of a hassle, we can also use [g_utf8_validate()](https://docs.gtk.org/glib/func.utf8_validate.html), [g_utf8_to_ucs4()](https://docs.gtk.org/glib/func.utf8_to_ucs4.html) or the [like of them](https://docs.gtk.org/glib/index.html?q=utf8)

C++'s support for utf8 is so terrible, especially before C++20. We may get better alternatives in C++26 tho.
But using glib also has a disadvantage: We need to copy all strings to store them in std::strings. 

This current solution should work now. It relies on the installed C.utf8 locale. It can and should be improved, by setting all locales to some sort of *.utf8 at the program start or when changing locales in the settings.

Febbe:
> I think that @panpaul is right in [#5994 (comment)](https://github.com/xournalpp/xournalpp/issues/5994#issuecomment-2467675776) Simply reverting [4bee81f](https://github.com/xournalpp/xournalpp/commit/4bee81fb9d35d4e55ddce276ff15cd5e68158dbb) should work fine on every target. At least it passes the test (even on the Windows runner).

Yes, it will work for all platforms whose locale is utf8, on most (older) Windows OS's this is not the case. Especially, when compiled with MSVC or clang-cl.
The reason is, that we imbue `std::locale()` via `ss.imbue(std::locale());`, this takes the global locale, which we set to the system's locale. The best would be, to ensure, that std::locale() is utf8. Or, we must correctly convert any string to utf8, before mixing it with other utf-8 strings.

PR #6045, opened 2024-11-10
Add Github Action to build and cache MacOS blobs

bhennion:
This adds an Action (which we can trigger via the Actions tab or the API).
This action builds (if needed) and caches the MacOS blobs on to the master branch cache. This cache is (readonly) accessible from every other branch and every PR.
So having this cache means faster builds all over.


rolandlo (approved):
LGTM

PR #6048, opened 2024-11-10
Check permissions to run \action create-installers CI

bhennion:
We now check the user has (at least) write permissions to the repo when running \action create-installers


rolandlo (approved):
LGTM
Let's see if it works. :-)

PR #6055, opened 2024-11-15
Use same targets for CI tests as for other builds

bhennion:
This PR does two things:

1. It uses .github/available-build-targets.json as a source for build targets in CI test: this way, all targets are configured via this file
2. It adds the ability for the create-installers workflow to add a version string to the built artifacts. Artifacts built via \action create-installers get "-PR-####-########" (with the PR number + short commit sha) in their name. This will also be useful for nightly builds (to add nightly-YYYYMMDD).

### DIFF
--- a/.github/workflows/make-installer-linux.yml
+++ b/.github/workflows/make-installer-linux.yml
@@ -25,6 +25,9 @@ on:
       head_sha:         # SHA of the commit to build. Defaults to the head of the triggering event (cf github doc.)
         type: string
         default: ''
+      artifact_version_suffix:  # Additional version string appended to 1.2.3+dev (for instance) in artifacts names
+        type: string
+        default: ''
     outputs:
       link:
         description: URL of the produced installer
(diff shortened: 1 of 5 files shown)


.github/workflows/make-installer-linux.yml
```diff
@@ -75,6 +78,15 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=/usr
             -DCPACK_GENERATOR="TGZ;DEB"
           cmake_commands: '--target package'
+      - name: 'Rename assets'
+        if: ${{ inputs.artifact_version_suffix != '' }}
+        shell: bash
+        run: |
+          for f in build/packages/xournalpp-*; do
+            correct_name="$(echo "$f" | sed "s/xournalpp-\(\(\:\?\(\:\?[0-9]\+\.\)\+[0-9]\+\)\(\:\?[+~][^-]\+\)\?\)-/xournalpp-\1${{ inputs.artifact_version_suffix }}-/")"
```
rolandlo, line 86: Please add an example (about what $correct_name results from which $f and $inputs.artifact_version_suffix) as a comment for this complicated expression.

rolandlo: Yes, like that.

PR #6058, opened 2024-11-17
Github action for releases and nightly builds

bhennion:
~~This PR is based on #6055.~~ Edit: now merged
It adds the necessary Github workflow to handle releases -- both nightly builds and normal releases.

It triggers nightly builds every night, moves the tag in case of success and updates the nightly build release note. See [here](https://github.com/bhennion/xournalpp/releases/) for example releases.

It also triggers a build whenever a tag of the form vA.B.C (with A, B and C any numbers) is pushed to the repository and publishes the resulting assets in a draft release (devs still need to add a short changelog, review the artifacts and approve via the web interface).
It should be compatible with the script/release_helper.sh script.
See [here](https://github.com/bhennion/xournalpp/releases/) for examples/tests.

~~Nb: there is an outage in Gitlab making it impossible to build the blob or package the app on MacOS for now. So I removed the MacOS targets for testing and for the time being.~~ Edit: The outage has been fixed. I added the MacOS targets back.

With this PR, the Azure pipeline and Cirrus integrations become entirely redundant and we can remove them.

### DIFF
--- a/mac-setup/build-app.sh
+++ b/mac-setup/build-app.sh
@@ -34,7 +34,7 @@ if [ ! -f "$GTK_MAC_BUNDLER" ]; then
   python3 -m venv "$GTK_MAC_BUNDLER_VENV"
 
   if [ ! -d gtk-mac-bundler ]; then
-    git clone https://gitlab.gnome.org/GNOME/gtk-mac-bundler.git
+    git clone --depth=1 https://gitlab.gnome.org/GNOME/gtk-mac-bundler.git
   fi
   pushd gtk-mac-bundler
   # The install script assumes it will install to some global path


rolandlo:
Fine, I will review it today.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,135 @@
...
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+  schedule:
+    # Nightly release every night...
+    - cron: "12 1 * * *"  # Once a day, at 1:12 am UTC (arbitrary but avoids CI load picks on round hours)
+
+env:
+  build-targets-json: '.github/available-build-targets.json'
+  build-type: 'RelWithDebInfo'           # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  nightly-build-branch: 'pr/releaseaction'
```
rolandlo, line 16: This is just for testing, and must be replaced before merging, right?

rolandlo:
> There are also a bunch of things in mac-setup/ that are no longer used I think. I could remove mac-setup/complete-build.sh, 

I agree. The MacBuild.md should be updated though, since it refers to that file.

> mac-setup/upload-release.sh, 

Yes, definitely. It was only used for the cirrus build.

> mac-setup/setup-gdk-pixbuf.sh 

This one is still used in 
https://github.com/xournalpp/xournalpp/blob/e851091b979aec9dcdbf29db84929c5cd66e4715/test/CMakeLists.txt#L19-L29

> mac-setup/build-gtk3.sh 

I agree.

rolandlo:
> And mac-setup/portaudio-2.0.pc?

I think it is not needed any more.

rolandlo:
> > This one is still used in
> > https://github.com/xournalpp/xournalpp/blob/e851091b979aec9dcdbf29db84929c5cd66e4715/test/CMakeLists.txt#L19-L29
> 
> Is it though? This is only run if the variable MACOSX is set. Is it normally set on recent targets? (It certainly isn't used by the CI runners: I removed the file already) I could not find anything about this variable (neither online nor in any other cmake file on the repo).

You're right, the variable MACOSX was set manually in `./CMakeLists.txt` in https://github.com/xournalpp/xournalpp/pull/3476/files via
```cmake
# Check for OS
if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
    set(MACOSX TRUE)
```
That's no longer the case, so we should remove the code block starts with `if(MACOSX)`.

PR #6084, opened 2024-11-23
Improve testing of locale issues

bhennion:
This PR allows to run the test-units with the locale set as in `control/XournalMain.cpp`. By doing so

1. I found two occurrences of missing calls to `stringstream::imbue()` on the release branch (they are now fixed)
2. It allows to test that `std::put_time()` used in `Document::createSaveFilename()` works even with locales using multibyte characters for, e.g., date printing, month/day names, or even year numbering.

I also added the test used in #6038 and ~~reverted https://github.com/xournalpp/xournalpp/commit/4bee81fb9d35d4e55ddce276ff15cd5e68158dbb (at least until #6038 is finished).~~

If anyone knows how to properly test those localization problems on Windows (the issue is `std::locale("")` throws when compiled with g++), I'll gladly add that here. Otherwise, this is up for review.

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -602,10 +602,7 @@ void XournalMain::initLocalisation() {
 #endif
 
     textdomain(GETTEXT_PACKAGE);
-
-#ifdef _WIN32
     bind_textdomain_codeset(GETTEXT_PACKAGE, "UTF-8");
-#endif
 
 #endif  // ENABLE_NLS
 


Febbe:
> Ideally, we would have a test unit that tells us that https://github.com/xournalpp/xournalpp/commit/4bee81fb9d35d4e55ddce276ff15cd5e68158dbb was indeed necessary for some platform and both this and https://github.com/xournalpp/xournalpp/pull/6038 should then be merged.

We would need a test suite, preferably on Windows, where `std::locale("")` would generate a non-utf8 CJK locale.
But maybe installing and setting such a locale in a Linux runner is easier.

Febbe:
FYI: I still thinking about this. But I'll have very few moments to actually implement something.

Some remarks I have:
- using wchar_t is not required, but the original locale is not necessary utf-8 so we convert it to wstring and back to FS::path.
- we could convert it directly to utf-8.
- c++ has no way to convert locales to utf8 via C++ locales directly (beside the deprecated functions)
- C++20 explicitly adds a C23 function to do this, which unfortunately only works on C locales. Which are not thread safe.
- All strings from GTK are utf-8 beside paths, which are utf-8 on windows, and the c-locale on all other systems

src/core/model/Document.cpp
```diff
@@ -156,14 +156,36 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
         wildcardString = SaveNameUtils::parseFilenameFromWildcardString(defaultPdfName, this->filepath.filename());
     }
 
-    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+
+    std::cout << "Locale has facet: " << std::has_facet<std::codecvt<wchar_t, char, std::mbstate_t>>(std::locale()) << std::endl;
```
Febbe, line 160: Consider, that the larger problem is, that the default locale of std::string is std::locale(""), but the strings we have as input is utf-8 (hopefully, otherwise we mix different encodings at other places). So the conversion from utf-8 to wchar_t might fail, since the locale used expects another encoding.
That's why I imbued a locale with utf encoding in my version, but this locale must be installed.

Febbe: The conversion to `wstring`/`wchar_t` could also be omitted, by directly imbuing the utf-8 locale for the time conversion via `ostringstream`. Using wchar_t only has the benefit on Linux, that the code point is UCS4 and **I believe** that they are Unicode characters and that they are the same for all encodings.

src/core/model/StrokeStyle.cpp
```diff
@@ -27,6 +27,7 @@ auto formatStyle(const std::vector<double>& dashes) -> std::string {
 
     // Else generate custom dashes string
     std::ostringstream custom;
+    custom.imbue(std::locale::classic());
```
Febbe, line 30:
```suggestion
    auto custom = serdes_stream<std::ostringstream>();
```

src/core/model/StrokeStyle.cpp
```diff
@@ -51,6 +52,7 @@ auto StrokeStyle::parseStyle(const std::string& style) -> LineStyle {
     }
 
     std::stringstream dashStream(style);
+    dashStream.imbue(std::locale::classic());
```
Febbe, line 55:
```suggestion
    auto dashStream = serdes_stream<std::stringstream>(style);
```

Febbe (approved):

Febbe:
We may want to also fix wrong usages of g_filename all over the place. Currently we almost always pass them directly to fs::u8path, which is wrong on Unix for all non utf8 locales.Instead we should use Util::fromGFilename

Febbe:
> > Not a blocker: We may want to also fix wrong usages of g_filename all over the place. Currently we almost always pass them directly to fs::u8path, which is wrong on Unix for all non utf8 locales.Instead we should use Util::fromGFilename. I already started this on a local branch.
> 
> We definitely should. Maybe the best would be to open a ticket listing all the utf-8 we notice, so we don't forget.
> I just wanna merge this so that #5994 gets fixed. We should probably release 1.2.5 with this fix ASAP, seeing the number of affected users.

Yes, merge this ASAP.

Issue #6066, opened 2024-11-19
Openning SVG file crashes


gutow:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

Official Release

### Xournal++ Version

1.2.3

### libgtk Version

3.24.38

### Bug Description

Crash when trying to import an SVG file

### Expected Behaviour

Image would be loaded and displayed.

### Steps to Reproduce

1. Make a simple SVG file with some text in it.
2. Try to import the image.

### Additional Context

Crash log:
Date: Tue Nov 19 08:41:17 2024
Error: signal 6
Xournal++ version 1.2.3
Git commit c6970f47 from xournalpp/HEAD
Gtk version 3.24.38

[bt]: (0) 0   xournalpp                           0x000000010024cacf _ZL12crashHandleri + 95
[bt]: (1) 1   libsystem_platform.dylib            0x00007ff8187b2fdd _sigtramp + 29
[bt]: (2) 2   ???                                 0x0000000000000000 0x0 + 0
[bt]: (3) 3   libsystem_c.dylib                   0x00007ff8186a9a79 abort + 126
[bt]: (4) 4   libglib-2.0.0.dylib                 0x000000010da7bc12 g_assertion_message + 434
[bt]: (5) 5   libglib-2.0.0.dylib                 0x000000010da7bc7e g_assertion_message_expr + 94
[bt]: (6) 6   xournalpp                           0x000000010034f3a5 _ZNK5Image8getImageEv + 405
[bt]: (7) 7   xournalpp                           0x00000001002a46c4 _ZN12ImageHandler11createImageEP6_GFiledd + 292
[bt]: (8) 8   xournalpp                           0x00000001002a44be _ZN12ImageHandler11insertImageEdd + 110
[bt]: (9) 9   xournalpp                           0x00000001002e9438 _ZN11XojPageView18onButtonPressEventERK17PositionInputData + 3864
[bt]: (10) 10  xournalpp                           0x00000001003193eb _ZN15PenInputHandler11actionStartERK10InputEvent + 571
[bt]: (11) 11  xournalpp                           0x0000000100318c8b _ZN17MouseInputHandler10handleImplERK10InputEvent + 347
[bt]: (12) 12  xournalpp                           0x0000000100317f17 _ZN12InputContext6handleEP9_GdkEvent + 951
[...]

PR #6087, opened 2024-11-23
Fix some Image import errors

bhennion:
Fixes the image import of either huge images or svg images with huge width and height parameters.
For SVG files, GdkPixbufLoader creates by default a pixbuf of the size the SVG's viewport. As an SVG's width and height parameters can be enormous, this causes crashes as in #6066.
This PR fixes that by capping the number of pixels to 32M - more than enough for A4 at 72pp.
It also adds error messages in case of loading error.

### DIFF
--- a/src/core/control/tools/ImageHandler.cpp
+++ b/src/core/control/tools/ImageHandler.cpp
@@ -59,15 +59,19 @@ auto ImageHandler::createImage(GFile* file, double x, double y) -> std::tuple<Im
     }
 
     // Render the image.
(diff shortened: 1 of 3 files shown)


rolandlo (commented):
The code looks fine. Do you have an image to test this with? The svg provided in https://github.com/xournalpp/xournalpp/issues/6066#issuecomment-2495548687 renders empty in Xournal++ (but fine in Inkscape), as you have already pointed out.

rolandlo (approved):
Thanks for the svg for testing. It works fine indeed.

PR #6119, opened 2024-12-02
feat: Add custom thickness slider. Ports PR #3288 to current master with small changes

atticus-sullivan:
Like discussed in https://github.com/xournalpp/xournalpp/pull/3288#issuecomment-2509753637, this is the attempt to port as much as possible from #3288 to the current master.

My idea (not fully implemented) is to implement the custom thickness similar to the fill opacity. So in effect there will be two entries for the toolbar and in the menubar. One for choosing the thickness and which activates the custom thickness. This way, the user can choose a custom thickness once, and switch between `thin` to the custom thickness without the need to choose a new custom thickness over and over again.

So far, only the buttons for activating the custom thickness are present. Also for now I simply used the icons added in #3288. Now we need two icons, so the icons might change in the future.
The `TODOs` are just markers where analog to #3288 one might open up the dialog window. With the new idea of using a different button these markers maybe are not needed anymore, we will see.

Also I'm not quite sure how to handle the "problem" that the bounds of the thickness slider maybe should be different for the highlighter compared to the pen (or just use a wide range).
Also maybe the window opening up should not only contain a slider but also some numeric input (in the best case synced with the slider)


rolandlo:
For discussion some screenshots would help. Maybe also compare to [this design](https://github.com/xournalpp/xournalpp/discussions/4561#discussioncomment-7138492).

atticus-sullivan:
Oh I'd say I wasn't aware of the redesign plan, but as I think of it, I believe I've seen that discussion and simply forgot it again.

In this case do we have a plan (regarding timing) for the redesign? (asking whether it still makes sense to implement the thickness slider in the current UI)

The (new) current state looks like this:
Peek 2024-12-03 11-36


Note: The code is still quite messed up since I used that other PR as base to implement this. So I think there are some changes that are not needed anymore. Also the naming is not entirely consistent sometimes `thickness` (wording from the former PR) and sometimes `size` (my wording) is used.

rolandlo:
> In this case do we have a plan (regarding timing) for the redesign? (asking whether it still makes sense to implement the thickness slider in the current UI)

No, we don't have a plan and the discussion about the redesign is just a discussion, where different opinions can be shared in the light of some concrete ideas for a redesign.

atticus-sullivan:
Nice suggestions. For now I just wanted to make the old PR working (well didn't work out so far as we can see in the tests, I still need to look into that) and use that as a starting point for the rest (having the functions to set a custom size etc probably is useful)

1. I like the popover idea, I'll just need to find out in what aspects this is different to creating a new window (I'll look at *add pages* for an example). I'm not sure about having the slider directly in the toolbar like with the zooming as these sliders take up quite some space especially if we want to have additional buttons (I for example would also like some way of setting the size precisely so something like a number input.

So I imagine something like this (icon placed in the left toolbar with popover open):
image

2. + 3. Yes maybe the size icons in the toolbar could be similar to the (old) color icons (regarding how they are drawn and specifying in the toolbar.ini). So the `toolbar.ini` would contain something like `SIZE(10)` or `THICKNESS(10)`. The question here would be how do we handle it that the highlighter sizes should be larger than the pen size? Simply use a fixed (?) factor or let the user freely specify the individual sizes?

In case we go with 2. + 3. (not having the current mean of setting the thickness) at least from what I imagine, the changes to the codebase will be a bit larger and it probably makes sense to start from the current master, not from that old PR which only added an additional custom element to the enum and a slider to set the corresponding size. In this case I think we should move this discussion to an issue/the new PR to not have the whole discussion in a dead PR.

As I think about the pinning idea: If we'd allow for setting individual sizes for the different tools, how would we know which size goes with which other size? (reading from the `toolbar.ini` this is simple, but how can we represent this in the UI?)
Also how would we delete old pinned sizes? (also thinking UI wise)

Or maybe it makes sense to rethink the pinning approach. We could add `x` thickness slots and an icon to configure the different thicknesses. Then in the popover, the user would select via some sort of tab/menu which tool+slot to configure with the slider in the popover. Not sure if this approach is really intuitive though. We might skip the configure button and maybe we can listen for a right click on the individual thickness button which would open the popover for that specific button/slot.

Opinions? (or decision to move this to a new PR)

PR #6124, opened 2024-12-03
Apply patch to cairo on Windows

### LINKED ISSUE #6014: Export PDF generated by XeLaTeX fails
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Official Release

### Xournal++ Version

1.2.4

### libgtk Version

3.24.43

### Bug Description

Open a pdf generated by XeLaTeX via xouralpp, then export to pdf, an error will occur. it says "Cairo error: out of memory"

### Expected Behaviour

Export pdf successfully.

### Steps to Reproduce

1. Use XeLaTeX to compile the following tex source:

```
\documentclass{article}
\begin{document}
test
\end{document}
```
(save to test.tex, then excute command `xelatex.exe --interaction=errorstopmode --synctex=-1 "test.tex"`

2. Open the generated pdf
3. Click export pdf and save
4. Error occurs.

### Additional Context

sample pdf: [test.pdf](https://github.com/user-attachments/files/17614151/test.pdf)

bhennion:
Fix #6014

The goal is to use a version of cairo patched with https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/595 until this fix reaches msys's package.

The cairo package is rebuilt by the CI with the patch (and then shipped with any releases).

@rolandlo Does that work for you as a workaround for #6014/#6117?

### DIFF
--- a/.github/actions/install_deps_windows/action.yml
+++ b/.github/actions/install_deps_windows/action.yml
@@ -24,3 +24,24 @@ runs:
           mingw-w64-x86_64-lua
           mingw-w64-x86_64-gtksourceview4
           mingw-w64-x86_64-imagemagick
+    - shell: msys2 {0}
+      # Apply https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/595 to cairo-1.18.2-1 to fix #6014
+      # Can be removed if either msys' packages include that fix by hand or cairo-1.18.3 is released.
+      run: |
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
LGTM

By the way the build failure was fixed upstream slightly differently via https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/600. But your fix looks good enough as well.

PR #6039, opened 2024-11-09
Add Github Action for Crowdin updates

bhennion:
This PR creates two Github Action:

1. upload-translatable-strings-to-crowdin.yml which computes xournalpp.pot and uploads it to Crowdin whenever something is pushed to either master or release-1.2. It does not push anything to the repository.
2. download-translations-from-crowdin.yml which pulls translations from Crowdin and push them to a PR. It also pushes the up-to-date xournalpp.pot file to the same PR. It is run once a week or on demand (in the Github Actions tab, e.g. before a release)

PR #6128, opened 2024-12-04
Update changelogs and release v1.2.5

bhennion:
@xournalpp/core Could you have a look and check it all seems ok?

MUST NOT BE MERGED VIA GITHUB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+## 1.2.6~dev (Unreleased)
+
 ## 1.2.5
 
 Bugfixes from various community contributors.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,13 +11,13 @@ endif()
 ## Project related Variables
 ## Also update changelog in debian folder!
 project(xournalpp
-        VERSION 1.2.5
+        VERSION 1.2.6
         DESCRIPTION "Xournal++ - Open source hand note-taking program"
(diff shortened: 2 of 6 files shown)


rolandlo (approved):
LGTM

Thanks for taking care of the release. Only question is whether we should merge #6039 before releasing 1.2.5? I would have time to finish reviewing it tonight.

rolandlo:
> I'm not quite sure whether the Azure pipeline/Cirrus will kick in or not, but even if it does, it will fail and won't interfere.

I have suspended both Github Apps, so they can't interfere at all.
Bildschirmfoto vom 2024-12-04 17-55-43

rolandlo:
> I did no modifications of the release message yet. We can debate what's in it.
> @xournalpp/core If you could have a look, test some binaries, it'd be perfect.

I have tested the Windows portable and non-portable, the MacOS-X64 and the AppImage. It all works fine. I currently cannot login to my Mac M1 Mini (because the bluetooth keyboard doesn't want to connect at login), so this will be left untested this time.

About the release message: Basically we use to reuse the summary from the debian/changelog and add an introductory sentence like 
> This is a new minor version of Xournal++ with bug fixes and improvements from the community.

plus a link to the changelog at the end via: 

> See the [changelog](https://github.com/xournalpp/xournalpp/blob/release-1.2/CHANGELOG.md) for a full list of the changes..

See the 1.2.3 release as an example. There is nothing more to tell this time, I think.

PR #6132, opened 2024-12-05
Generate release messages automatically

bhennion:
As discussed in #6128 

I tested it. It gives:
Copie d'écran_20241205_112733

For nightly builds, I simply displayed the last 15 commits - similar to what happened with the Azure pipelines


rolandlo (approved):
LGTM

Issue #6115, opened 2024-12-02
Single key shortcuts ?


Vedrinic:
### Is Your Feature Request Related to a Problem?

It's a bit of first world problem but why can't we use single key shortcuts on Xournal without a plugin ? Like P for Pen, E for eraser, Space for hand tool like all similar apps does ? Those keys are not assigned to anything in the app. Can't even switch to a Pen tool without using 2 hands when it's like this. (ctrl+shift +p) 

(Thanks for your time for creating this amazing application. It completely changed how I take notes and I'm super happy with it.)

### Feature Description

Shortcuts that make sense

MiltonBalaOfficial:
I also asked same type of question. The answer was: https://github.com/xournalpp/xournalpp/issues/5918#issue-2567680132

Vedrinic:
It's not actually about my personal preference about a shortcut. It's a question of why do something the hard way when you can do it easy.  Single keys not bind to anything in the application, they are free to use. Maybe I am missing something because of my lack of knowledge on coding.

PR #6140, opened 2024-12-07
Unify key binding implementations

bhennion:
The long term goal (for a follow up PR) is to implement configurable shortcuts.

This PR simply uses the KeyBinding architecture used in TextEditor for key bindings various locations in the code.

- [x] Shortcuts in XournalView
- [x] Shortcuts in SplineHandler
- [ ] Shortcuts in GeometryToolInputHandler
- [ ] Shortctus in BaseShapeHandler
- [ ] Shortcuts in SearchBar

It does not mean all of those key bindings should be open to configuration, but it gives us a common base to work with.

### DIFF
--- a/src/core/control/CompassController.cpp
+++ b/src/core/control/CompassController.cpp
@@ -8,6 +8,9 @@
 #include "model/Stroke.h"
 #include "model/XojPage.h"
 
+constexpr double MIN_HEIGHT = 0.5;
+constexpr double MAX_HEIGHT = 10.0;
+
 CompassController::CompassController(XojPageView* view, Compass* compass): GeometryToolController(view, compass) {}
 
 CompassController::~CompassController() = default;
@@ -123,3 +126,6 @@ void CompassController::finalizeRadialStroke() {
 auto CompassController::existsOutlineStroke() -> bool { return angleMax != std::numeric_limits<double>::lowest(); }
 
 auto CompassController::existsRadialStroke() -> bool { return radiusMax != std::numeric_limits<double>::lowest(); }
+
+auto CompassController::getMinHeight() const -> double { return MIN_HEIGHT; }
+auto CompassController::getMaxHeight() const -> double { return MAX_HEIGHT; }
--- a/src/core/control/CompassController.h
+++ b/src/core/control/CompassController.h
@@ -116,6 +116,9 @@ public:
      */
     bool existsRadialStroke();
 
+    double getMinHeight() const override;
+    double getMaxHeight() const override;
+
 private:
     /**
      * @brief when a stroke near the radius with the measuring marks
--- a/src/core/control/GeometryToolController.cpp
+++ b/src/core/control/GeometryToolController.cpp
@@ -43,8 +43,13 @@ void GeometryToolController::scale(double f, const xoj::util::Point<double>& cen
     this->scale(f);
 }
 void GeometryToolController::scale(double f) {
-    geometryTool->setHeight(geometryTool->getHeight() * f);
-    geometryTool->notify(true);
+    const double h = geometryTool->getHeight() * f;
+    if (h <= getMaxHeight() && h >= getMinHeight()) {
+        geometryTool->setHeight(h);
+        geometryTool->notify(true);
+    } else {
+        geometryTool->notify(false);
+    }
 }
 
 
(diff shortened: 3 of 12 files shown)


rolandlo:
I haven't made up my mind yet, about which shortcuts/keypresses should be configurable yet. But about discoverability let me add that there is another way to help with that: the [GtkShortcutsWindow](https://docs.gtk.org/gtk4/class.ShortcutsWindow.html) (which is present both in Gtk3 and Gtk4). It's just used for presenting shortcuts/keypresses, not for modifying them.
In the Libadwaita redesign mockup I have added a GtkShortcutsWindow with all shortcuts I could find. For example here is the section about the geometry tools:
Bildschirmfoto vom 2024-12-10 05-35-02

rolandlo:
> > I haven't made up my mind yet, about which shortcuts/keypresses should be configurable yet. But about discoverability let me add that there is another way to help with that: the [GtkShortcutsWindow](https://docs.gtk.org/gtk4/class.ShortcutsWindow.html) (which is present both in Gtk3 and Gtk4).
> 
> That's a good point, but I am not quite sure how usable it is with configurable shortcuts (even just for displaying them). It seems that GtkShortcutsShortcut/GtkShortcutsSection/GtkShortcutsWindow have no public constructor. The only way to build them is via a GtkBuilder and a UI file. It makes it complicated to generate a GtkShortcutsWindow dynamically.

Apparently that changed in Gtk 4.14 though with new API
`gtk_shortcuts_window_add_section`
`gtk_shortcuts_section_add_group`
`gtk_shortcuts_group_add_shortcut `, see
https://docs.gtk.org/gtk4/class.ShortcutsSection.html?q=shortcuts

It's unclear to me however how to construct a `GtkShortcutsShortcut` though. In Python I can do `Gtk.ShortcutsShortcut()` to construct one, but I don't see how to set its properties.

rolandlo:
> My feeling is that:
> 
>     * Key bindings defined in mainmenubar.xml and in control/actions/ActionProperties.h should be configurable. The user should be able to add/modify shortcuts for most (if not all) of the actions in the ActionDatabase (with a choice of parameter if the GAction has a parameter)

I agree.

> 
>     * Key bindings in BaseShapeHandler could be configurable: improves discoverability

I agree.

> 
>     * Key bindings in SplineHandler should be configurable: allows to reassign "single key shortcuts" to other uses (see e.g. [Single key shortcuts ? #6115](https://github.com/xournalpp/xournalpp/issues/6115)) + improves discoverability

I agree.
> 
>     * Key bindings in VerticalToolHandler could be configurable (improves discoverability)

I agree.
> 
>     * Key bindings in EditSelection: I don't know. Maybe not configurable

Not configurable, I would say. These are pretty standard and intuitive.

> 
>     * Key bindings in TextEditor shoud **not** be configurable: the keyboard input is too important to let users meddle with it here (like introducing a single key shortcut for instance...).
> 
I agree.

>     * Key bindings in GeometryToolInputHandler should be configurable: allows to reassign "single key shortcuts" to other uses (see e.g. [Single key shortcuts ? #6115](https://github.com/xournalpp/xournalpp/issues/6115)) + improves discoverability
> 
I agree.

>     * Key bindings in SearchBar should not be configurable
> 
I agree. 

>     * Key bindings in XournalView: I would have said "don't allow for customization" but as I mentioned above, contributors have already customized them (by adding vim-like navigation). So maybe yes, or maybe partially, or maybe the navigation things should be triggered via the ActionDatabase and the color things should be configurable
> 
The color things should definitely be configurable. The arrow keys should probably not be configurable, but allowing to add other navigation keys (like vim-style navigation) should be allowed.

> 
> So, @xournalpp/core do you have any opinion, insights or comments about those questions?
> 
> List of current keybindings:
> Key bindings in BaseShapeHandler
> 
> Key bindings in SplineHandler
> 
> Key bindings in VerticalToolHandler
> 
> Key bindings in EditSelection
> 
> Key bindings in TextEditor
> 
> Key bindings in GeometryToolInputHandler
> 
> Key bindings in SearchBar
> 
> Key bindings in XournalView

bhennion:
And could you have a quick look at the code and tell me if it looks like progress or not? It feels to me it cleans up things quite a bit, in particular by getting rid of the `else if (key ==GDK_KEY_??? && state == GDK_MODIFIER_???) {` chains.

rolandlo:
> And could you have a quick look at the code and tell me if it looks like progress or not? It feels to me it cleans up things quite a bit, in particular by getting rid of the `else if (key ==GDK_KEY_??? && state == GDK_MODIFIER_???) {` chains.

Sure, it feels like a big step forward.

PR #6158, opened 2024-12-13
Package lua-lgi and dependencies on Windows

rolandlo:
Packaging `lua-lgi` and dependencies on Windows takes (only) roughly 4 MB when extracted.
- The .typelib files take 3.3 MB
- The libraries (libgirepository-1.0-1.dll, libgirepository-1.0.dll.a, corelgilua51.dll) take 0.5 MB
- The .lua files take 0.26 MB

It seems to work all fine (even without changing the lua package path or package cpath).


rolandlo:

rolandlo:
Merging in 24 hours if no objections are raised. I have tested both the installer and the portable version and they work both fine.

Issue #2305, opened 2020-10-09
Adding a start page to Xournalpp for finding files easily at startup


Narice:
**Is your feature request related to a problem? Please describe.**
I have a lot of xournalpp documents (as this app is amazing for remote learning x)) so finding back specific lecture notes, even when having a bunch of folders, is kind of a pain.

**Describe the solution you'd like**
Having a start page with thumbsnails of the notes and maybe with a search bar would be an absolute blast! here is a rough sketch done in xournalpp itself ;)
![image](https://user-images.githubusercontent.com/31109099/95630899-f48f7280-0a82-11eb-9e10-9814170173ab.png)
The last edited files could be listed there.
Or there could be a home directory for notes (personnaly I have all my notes under a single xournal directory in my home directory) where xournalpp could look for notes to list them there for completeness or even for organization purposes if folders themselves are displayed in the list (this would be really good, but I know the difficulty of possibly adding these types of features):
![image](https://user-images.githubusercontent.com/31109099/95631498-2228eb80-0a84-11eb-8ee2-ff73d9da8065.png)
I hope my sketches are clear! :smile:

**Describe alternatives you've considered**
There is no alternative to xournalpp (trust me I've searched thoroughly x)).

JMLTUnderCode:
Hi! Nice to meet you Xournalpp members @LittleHuba, @rolandlo and @Technius. We are a team involved in software development. More specifically, we are students of Computer Engineering at the Simon Bolivar University of Venezuela, in particular we are taking the course Software Engineering I and in it we have as project the contribution of a real open source project. The team will be formed by 
@A6575
@AndreaDiMartin
@lauraparilli
@cmhd2001
@lgarciausb
And myself @JMLTUnderCode, who will be the organizer of this team and I am taking the time to inform you that we want to work on the development of this Issue #2305 during the next three weeks. We were interested in developing and contributing to this application since several of us use Xournal++ for our academic environment and we find it great to be trying to implement this functionality. 

Now, I will make a fork of the repository to start working on it. However, we have a question related to the issue assignment, is the issue assigned to us or it's enough with leaving this comment to suggest that we are working on it?

Best regards (☞ﾟヮﾟ)☞

LittleHuba:
Hi @JMLTUnderCode,

Sounds cool! Have fun working on this issue. The comment should be enough.

PR #6159, opened 2024-12-14
StartPage Implement

JMLTUnderCode:
# Description

In this PR a preliminary version of issue #2305 is implemented. Here, it is intended to have a basic version of what is requested in the task in order to establish a guideline structure for the future implementation of this issue.

# Workflow

## Main Idea
Initialize two windows. Mainwindow without major change and the new StartPage window. In first instance only SHOW must be applied to StartPage, depending on the interaction occurred inside the StartPage this one is destroyed and shows the MainWindow, simulating this way a home page transition. Buttons were included dynamically in a grid of 4 columns according to the recent files and that have the characteristic of obtaining the name of these documents. On the search engine side, a simple keyboard input is used to search for text matches in the names of recent files and update the grid of buttons.

Translated with DeepL.com (free version)

## Created Files

* src/core/gui/HomeWindow.h
* src/core/gui/HomeWindow.cpp
* ui/homepage.glade

## Modified Files

* src/core/gui/GladeGui.cpp
* src/core/gui/MainWindow.h
* src/core/gui/MainWindow.cpp
* src/core/gui/menus/menubar/RecentDocumentsSubmenu.h
* src/core/gui/menus/menubar/RecentDocumentsSubmenu.cpp
* src/core/control/Control.h
* src/core/control/Control.cpp
* src/core/control/XournalMail.cpp

# Preview

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -40,6 +40,7 @@
 #include "control/xojfile/LoadHandler.h"                         // for Load...
 #include "control/zoom/ZoomControl.h"                            // for Zoom...
 #include "gui/MainWindow.h"                                      // for Main...
+#include "gui/HomeWindow.h"
 #include "gui/PageView.h"                                        // for XojP...
 #include "gui/PdfFloatingToolbox.h"                              // for PdfF...
 #include "gui/SearchBar.h"                                       // for Sear...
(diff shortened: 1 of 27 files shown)


rolandlo:
@JMLTUnderCode Is there still work being done on this feature? Or what is the current status?

PR #6161, opened 2024-12-16
Package lua-lgi for MacOS

rolandlo:
This PR most notably packages `lua-lgi` on MacOS in the same spirit as #6158. That means `lua-lgi` will come bundled on all platforms where we provide bundled installations (Windows, MacOS, snap, flatpak), which makes it much easier for plugin authors that want to create some graphical user interface elements.

~While we already built Lua on MacOS as specified in `xournalpp.modules` and thus enabled plugins on MacOS, we somehow didn't package Lua on MacOS. This is fixed with this PR.~

I have also updated the Lua version from 5.4.6 to 5.4.7 (released in June 2024). 
There was a typo in `libpoppler-glib*.dylib`, which suggests that these lines could be omitted as well. I have just fixed the typo.

The assets will still need to be tested.

**Edit:** `liblua.a` is a static library and baked into the xournalpp executable, so it doesn't need packaging of course.

### DIFF
--- a/mac-setup/Info.plist
+++ b/mac-setup/Info.plist
@@ -182,6 +182,12 @@
     <key>NSDownloadsUbiquitousContents</key>
     <true />
 
+    <key>LSEnvironment</key>
+    <dict>
+        <key>DYLD_FALLBACK_LIBRARY_PATH</key>
+        <string>/Applications/Xournal++.app/Contents/Resources/lib</string>
+    </dict>
+
     <key>NSHighResolutionCapable</key>
     <true />
     <key>NSRequiresAquaSystemAppearance</key>
--- a/mac-setup/jhbuild-version.lock
+++ b/mac-setup/jhbuild-version.lock
@@ -1,2 +1,2 @@
-gtk-osx=8c5052f4a57d5ec54d74bf13aaa1e1aeccad75f1
-xournalpp-pipeline-dependencies=5fb1da3de6ef84f129071d828ec99025ffa83894
+gtk-osx=6b8603cd4a40a54cbac9a1b8e722aadec1c3022f
+xournalpp-pipeline-dependencies=0ba50861b076ae472a69fb9335193aa8ce683c4d
--- a/mac-setup/xournalpp.bundle
+++ b/mac-setup/xournalpp.bundle
@@ -40,7 +40,7 @@
 
   <!-- Poppler -->
   <binary>
-    ${prefix}/lib/libpoppler-glib*.dynlib
+    ${prefix}/lib/libpoppler-glib*.dylib
   </binary>
 
   <!-- Translation filenames, one for each program or library that you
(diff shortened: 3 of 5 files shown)


rolandlo:
\actions create-installers macos

rolandlo:
\action create-installers macos

rolandlo:

rolandlo:

rolandlo:

rolandlo:

rolandlo:

rolandlo:
> Is it worth merging to the release branch rather?

I don't think so. There is a lot of new API in the master branch and interesting plugins using lgi will use the master branch anyway.

By the way it doesn't work yet.
- The `package.path` and `package.cpath` are not setup automatically yet. This can be achieved by setting `LUA_PATH` and `LUA_CPATH` in `setup-env.cpp`. 
- More importantly on MacOS there is an issue with core modules not being found https://github.com/lgi-devs/lgi/issues/151 which stems from https://github.com/lgi-devs/lgi/commit/dab172c9769e161037e59bc34f9a665d33b89736. Probably I will have to revert that commit (or replace it by something working), since it determines the GObject introspection library directory at compile time, which is pretty bad when lgi is compiled on a different machine than it is run.

rolandlo:

rolandlo:

rolandlo:
Ok, now up to creating a rather unfortunate symlink
```term
sudo mkdir -p /Users/runner/gtk/inst/lib
sudo ln -s /Applications/Xournal++.app/Contents/Resources/lib/libcairo.2.dylib /Users/runner/gtk/inst/lib/libcairo.2.dylib
```
`lua-lgi` works out of the box with this PR on MacOS Intel and MacOS ARM. Possibly more such symlinks are needed for things that are loaded manually with
```
lua
lgi = require "lgi"
core = require "lgi.core"
core.module(name, version)
```
In the `lgi` code it is only needed with name: "cairo" and version: 2.

In order to fix this `lgi` will need patching, maybe just reverting lgi-devs/lgi@dab172c. I will investigate. Also I will check what samples from the `lgi` repository work.

rolandlo:
> Also I will check what samples from the `lgi` repository work.

Pretty much everything works (except the Images demo, which crashes after a short while). I had to replace GtkSource.View by Gtk.TextView, since the correct GtkSourceView version is not preinstalled.

rolandlo:
> In order to fix this `lgi` will need patching, maybe just reverting [lgi-devs/lgi@dab172c](https://github.com/lgi-devs/lgi/commit/dab172c). I will investigate.

It looks like the issues addressed by that problematic commit were fixed in glib from version 2.76 on. We have version 2.82. in our MacOS build, so it looks like we can really revert that commit and see whether the gtk-demo examples still work (the same way as before)..

rolandlo:

rolandlo:

PR #6195, opened 2025-01-09
Fix recolor settings change not being applied

kchibisov:
The change was reverted back to normal since it was reset to default values, due to the light/dark being swapped.

--

To repro just change them to custom values and restart, will revert back on the next launch.


atticus-sullivan:
I don't quite get your note, but after looking at the change (and what I've written) I see the point. Yes somehow this slipped me.
Your change makes sense and I also already tested it -> works for me

One note: Actually for me the old version didn't completely changed both color values back to the defaults (only the dark color). But I think this also depended on the ordering of the entries in the `settings.xml` (well this reading the settings was just completely messed up and instable).


So in my words:
The issue is when the user customizes the recoloring colors (two of the parameters). Xournalpp messed around with these two parameters. Depending on some factors (like the order in which these two parameters are read) one or even both parameters were just reset to their default values.

(didn't catch this when testing for the original PR as I chose the colors of my liking as defaults and thus never changed them)

kchibisov:
CI failure is unrelated.

Issue #6165, opened 2024-12-18
Pen marker is not recoloured when the option is active


RobertoBochet:
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

### (Linux only) Desktop Environment

Gnome

### (Linux Only) Display Server

Wayland

### Installation Method

Compiling from source

### Xournal++ Version

1.2.5+dev (dcae23d0)

### libgtk Version

3.24.43

### Bug Description

I'm trying out the wonderful `Recolor drawing area` feature that @atticus-sullivan recently introduced with #6090. I notice that if I use a stylus with tablet, the pen cursor on the drawing area is not recolored.

Here I try to show you
![image](https://github.com/user-attachments/assets/4fe4c15c-a145-45ad-8d76-09f861d78b61)
The selected color is black, the stroke appears white (as it should be) but the pen marker is still black.

### Expected Behaviour

Also the pen marker should be recolored

### Steps to Reproduce

1. Use a drawing tablet with a stylus
2. Place the pen in the drawing area without drawing

rolandlo:
I have also noticed that. No idea if that was intentional. @atticus-sullivan

atticus-sullivan:
Oh it's nice to see someone actually using the feature. Well tbh I just didn't really bother since the pen is still visible. Also it kinda shows that the stroke in "reality" is drawn with a different color an the white color (in your case) is just the view-color, not the stroke color.

Do you see issues with that state? / Do you think we should recolor the pen marker as well?

RobertoBochet:
Are you kidding? You are the savior of my eyes, I love your feature, thank you very much.

I think there are two kind of issues (both can be considered not a real bugs).

1. Consistency
    ![Screenshot from 2024-12-19 01-33-41](https://github.com/user-attachments/assets/b4a79fcc-7319-4e26-b650-5364ad8bd4da)
    ![Screenshot from 2024-12-19 01-33-59](https://github.com/user-attachments/assets/67ee9f5a-291d-439e-b727-b3a2663310ce)
    When the recolor is disabled the cursor have the same color of the of the stroke (as you can see in the screenshot above).
    ![Screenshot from 2024-12-19 01-34-15](https://github.com/user-attachments/assets/0944324c-c8cb-49c5-a52a-eac42e37eadc)
    But, when the recolor is enabled this behavior is not preserved(as you can see in the original screenshot with the black and above with the red)

2. Accessibility
[...]

PR #6203, opened 2025-01-12
recolor: adjust pen coloring as well

atticus-sullivan:
Like noted in https://github.com/xournalpp/xournalpp/issues/6165 we also want to adjust the coloring of the pen when the recolor feature is active.

This is a follow-up from https://github.com/xournalpp/xournalpp/pull/6090.

See these two examples:
2025-01-12-111003_1920x1080_scrot
2025-01-12-111016_1920x1080_scrot

Looking at the code, I would have expected I also need to adjust https://github.com/xournalpp/xournalpp/blob/b81bdc5a7312b5736dd3ebbb59ae28011a6f2d0c/src/core/gui/XournalppCursor.cpp#L497
but for me this was not needed. Can someone confirm this is true?

Also do you think we should obtain `Settings` once in `createHighlighterOrPenCursor` (instead of requesting it (now) 4 times via `control->getSettings()`? (probably doesn't change much)

### DIFF
--- a/src/core/gui/XournalppCursor.cpp
+++ b/src/core/gui/XournalppCursor.cpp
@@ -436,6 +436,9 @@ auto XournalppCursor::getPenCursor() -> GdkCursor* {
 
 auto XournalppCursor::createHighlighterOrPenCursor(double alpha) -> GdkCursor* {
     auto irgb = control->getToolHandler()->getColor();
+    if(control->getSettings()->getRecolorParameters().recolorizeMainView) {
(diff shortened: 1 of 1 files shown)


atticus-sullivan:
formatting job fails because
```
The following packages have unmet dependencies:
 clang-format-18 : Depends: libllvm18 (= 1:18.1.8~++20240731025011+3b5b5c1ec4a3-1~exp1~20240731145104.143) but 1:18.1.3-1ubuntu1 is to be installed
```
seems to be a CI problem, nothing with the code itself.

PR #6204, opened 2025-01-12
Add bindings for XF86Copy/Paste/Cut

kchibisov:
Those bindings are very common and used by a lot of applications on X11/Wayland.


kchibisov:
I've updated, though, I'm not sure there's any visual indicator of that. I did got a label using the gtk method to ensure that I've specified the right thing.

Also, is it expected that `ctrl+c` is used on macOS? It's usually `Cmd+c`, however, I don't have macOS to test this on, just know that it's usually with cmd, thus just by looking on how copy was handled it seems like it's just ctrl+c everywhere.

kchibisov:
Updated.

Issue #6215, opened 2025-01-16
Accessibility / a11y: Select Linear Text  does not expose selection to xsel / clipboard / PRIMARY selection


tgrushka:
### Operating System

Linux

### (Linux only) Distribution

Mint 22

### (Linux only) Desktop Environment

Cinnamon

### (Linux Only) Display Server

X.Org

### Installation Method

from recent GitHub nightly

### Xournal++ Version

1.2.5+dev b81bdc5a

### libgtk Version

3.24.41

### Bug Description

I am legally blind and want to use my screen reader (which basically uses `xsel` to get the current selection and read it aloud) in Xournal++.

When I make a selection using the **Select Linear Text** tool, a popup appears with icons that are difficult to use and get in the way of using my screen reader. Other apps, such as Firefox, Terminal, and Document Viewer, allow me to make a selection unobtrusively like any other desktop unix application. Xournal++ has this popup that:
- has icons that are too small to read (although hovering does show tooltips)
- does not go away when I hit Esc / Escape key (keyboard a11y)
- clears the selection if I click outside, as I don't want any of the actions in the popup.

Thus, **the popup cannot be dismissed without choosing an action or losing the selection.**

More critically, the `PRIMARY` selection is not exposed to `xsel` as it should be, therefore I cannot access the text at all with my screen reader.

I am unable to find any options in the menu for disabling the popup on text selection.

Since this popup is an additional (and I posit, unexpected) side-effect, is it possible to allow us to disable it?

### Expected Behaviour

When I make a text selection using the **Select Linear Text** tool:
1. Commensurate with most other Linux/Unix desktop apps, selected text should be immediately available as the `PRIMARY` selection in `xsel` as soon as the text is selected and the mouse button released (the user is finished making the selection);
2. There should be no undesired side-effects such as popups when making a text selection, especially with icon-only selections, which can present major problems for both:
  a. users with legal blindness / visual impairment (myself),
  b. totally blind users, and possibly
  c. neurodivergent users.

### Steps to Reproduce

1. Open a PDF.
2. Select **Select Linear Text** in the **Tools** menu.
3. Make a text selection somewhere in the document.
4. Notice the popup of icons.
5. Open Terminal.
6. Type `xsel`.
7. Notice that the output does **not** contain the active selection in Xournal++.

tgrushka:
[...]

PR #6221, opened 2025-01-18
Inform windowing system of active text selections

bhennion:
Exposes selected text (both when in a Text element and in a background PDF) to the windowing system.
With this fix, utilities like `xsel` (X11) or `wl-paste --primary` (Wayland) will print out the selected text.

This fixes half of #6215.


rolandlo (approved):
LGTM

PR #6257, opened 2025-02-02
Update gtk-osx and remove obsolete fixes

rolandlo:
The two fixes have been merged into gtk-osx. So I'm removing them here.


rolandlo:
Oh, I forgot about that. Thanks for pointing it out. I will do it right away.

PR #6265, opened 2025-02-07
Fix compass outline drawing

### LINKED ISSUE #6256: Compass tool causing weird behaviour when drawing with it over and over again.
### Operating System

Linux

### (Linux only) Distribution

Fedora 41

GNOME

X.Org

### Installation Method

Official Release (rpm package)

### Xournal++ Version

1.2.5

### libgtk Version

3.24.43

### Bug Description

When selecting the Compass tool and rapidly drawing circles in the same direction in the same spot different patterns emerge.

### Expected Behaviour

Nothing was supposed to happen, just a normal circle even if drawn over again.

### Steps to Reproduce

1. Enable Compass tool
2. Select pen and draw circles over and over again
3. After a moment a pattern will emerge
4. Doing this for long enough would cause the pattern to reverse and go back to where you started.

https://github.com/user-attachments/assets/67c44d85-6b38-4a28-92db-d5d28802219e

### Additional Context

Video attached for further clarity.

rolandlo:
Fixes #6256

### DIFF
--- a/src/core/control/CompassController.cpp
+++ b/src/core/control/CompassController.cpp
@@ -87,7 +87,7 @@ void CompassController::updateOutlineStroke(double x) {
     }
     for (auto i = 0; i <= 100; i++) {
         const utl::Point<double> p =
-                getPointForAngle(angleMin + static_cast<double>(i) / 100.0 * (angleMax - angleMin));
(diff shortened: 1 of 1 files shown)


rolandlo:
This is tested and fixes the issue. Merging right away.

PR #6326, opened 2025-03-09
Fix some undefined behaviour (and possible SegFaults) with GTK signals

bhennion:
This fixes some UB around various signal handlers 

1. some had a return value and shouldn't have (probably inocuous)
2. some had no return value but should have had one
3. The FontButton callback was really messed up
    - some gpointer was cast to an unrelated type (GtkToolItem* vs GtkFontButton*)
    - The callback was set up twice (hence called twice)
4. some callbacks were dangerous, potentially causing SegFaults

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -410,13 +410,18 @@ void initResourcePath(GladeSearchpath* gladePath, const gchar* relativePathAndFi
 
 void on_activate(GApplication*, XMPtr) {}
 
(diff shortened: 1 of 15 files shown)


rolandlo (changes requested):
The changes unrelated to the font button look fine. The font button doesn't work properly. I can't change the font of any text element any more (and I can on 1.2.6).

rolandlo (approved):
It all works now.

PR #6338, opened 2025-03-16
Remove Cairo fix after Cairo 1.18.4 has hit MSYS2

rolandlo:
See https://packages.msys2.org/packages/mingw-w64-x86_64-cairo


rolandlo:
Github actions indeed uses version 1.18.4-1 now.

Bildschirmfoto vom 2025-03-16 07-35-42

Issue #4931, opened 2023-06-13
Anyone an idea


HTRiepe:
Date: Tue Jun 13 12:44:27 2023
Error: signal 11
[bt]: (0) /usr/bin/xournalpp(+0xbf169) [0x55deab34c169]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x3c4b0) [0x7fd61263c4b0]
[bt]: (2) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x340b5c) [0x7fd613740b5c]
[bt]: (3) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x3528e5) [0x7fd6137528e5]
[bt]: (4) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x170) [0x7fd6130e6010]
[bt]: (5) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x423d6) [0x7fd6131133d6]
[bt]: (6) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x102a) [0x7fd61310369a]
[bt]: (7) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7fd613103923]
[bt]: (8) /lib/x86_64-linux-gnu/libgtk-3.so.0(gtk_gesture_set_sequence_state+0xa5) [0x7fd6135a9d75]
[bt]: (9) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x991b0) [0x7fd6134991b0]
[bt]: (10) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x11cc) [0x7fd61310383c]
[bt]: (11) /lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7fd613103923]
[bt]: (12) /lib/x86_64-linux-gnu/libgtk-3.so.0(+0x1aa45d) [0x7fd6135aa45d]
[bt]: (13) /lib/x86_64-linux-gnu/libgdk-3.so.0(+0x335fb) [0x7fd6133375fb]
[bt]: (14) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x5ba42) [0x7fd612fe5a42]
[...]

PR #6356, opened 2025-03-24
Plug memory leaks

bhennion:
I ran with Valgrind trying desperately to reproduce #4931. I couldn't. But I got a good memory leak report:
 - plenty of memleaks in GTK/GDK/PortAudio/rsvg/pango. Nothing we can really do about them. Possibly some are false positives.
 - I found a couple of memleaks. The most notable one: strokes erased with "standard" eraser were never deleted.

This PR fixes those memleaks I could locate and fix.


rolandlo (approved):
LGTM

PR #6362, opened 2025-03-29
Const correctness of Element pointers

bhennion:
- Most `Element*` are changed into `const Element*`. We should only modify an element via its owner `std::unique_ptr<Element>`.
- Layers and other containers expose a view of `const Element*` to see their content.

This PR is based on #5263.

This PR highlights 3 places where we break the MVC pattern: Elements are modified by someone else than their owner. Those places are:
1. In EraseHandler: we call setErasable() on Stroke instances
2. In TextEditor: we call setInEditing() on Text instances
3. In the Lua interface `applib_scaleTextElements()`. It seems like a weird feature to expose. It was probably added for the font migration plugin. We should probably retire this plugin by now and remove this function from the Lua interface.

I see good reasons for the way things are done in 1. (to preserve the z-index while erasing) and 2. (so the edited text still appears in autosaves). I don't think we should do anything about it for now, but I wanted to point this fact out.

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -83,7 +83,7 @@ auto createFromFloatingElements(Control* ctrl, const PageRef& page, Layer* layer
             std::make_unique<EditSelection>(ctrl, std::move(elts), page, layer, view, bounds, snappingBounds), bounds);
 }
 
-auto createFromElementOnActiveLayer(Control* ctrl, const PageRef& page, XojPageView* view, Element* e,
+auto createFromElementOnActiveLayer(Control* ctrl, const PageRef& page, XojPageView* view, const Element* e,
                                     Element::Index pos) -> std::unique_ptr<EditSelection> {
     Document* doc = ctrl->getDocument();
     Layer* layer = nullptr;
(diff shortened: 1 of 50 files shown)


rolandlo:
Can you please rebase on master, so I can take a closer look?

src/core/control/tools/EditSelection.cpp
```diff
@@ -232,7 +232,7 @@ EditSelection::EditSelection(Control* ctrl, InsertionOrder elts, const PageRef&
     this->view->getXournal()->getCursor()->setRotationAngle(0);
     this->view->getXournal()->getCursor()->setMirror(false);
 
-    for (auto&& e: contents->getElements()) {
+    for (auto&& e: contents->getElementsView()) {
```
rolandlo: Shouldn't this loop use const auto&, since only const methods are called?
```suggestion
    for (const auto& e: contents->getElementsView()) {
```

src/core/control/tools/Selector.cpp
```diff
@@ -23,17 +23,17 @@ auto Selector::finalize(PageRef page, bool disableMultilayer, Document* doc) ->
...
         for (auto it = layers.rbegin(); it != layers.rend(); it++) {
-            Layer* l = *it;
+            const Layer* l = *it;
             if (!l->isVisible()) {
                 continue;
             }
             bool selectionOnLayer = false;
             Element::Index pos = 0;
-            for (auto&& e: l->getElements()) {
+            for (auto&& e: l->getElementsView()) {
```
rolandlo:
```suggestion
            for (const auto& e: l->getElementsView()) {
```

src/core/control/tools/Selector.cpp
```diff
@@ -45,11 +45,11 @@ auto Selector::finalize(PageRef page, bool disableMultilayer, Document* doc) ->
         }
     } else {
         std::lock_guard lock(*doc);
-        Layer* l = page->getSelectedLayer();
+        const Layer* l = page->getSelectedLayer();
         Element::Index pos = 0;
-        for (auto&& e: l->getElements()) {
+        for (auto&& e: l->getElementsView()) {
```
rolandlo:
```suggestion
        for (const auto& e: l->getElementsView()) {
```

rolandlo (approved):
LGTM. I have only some minor comments about for loops.

Issue #2486, opened 2020-12-05
PDF annotation interoperability with other apps


zeidoo:
**Is your feature request related to a problem? Please describe.**
Xournal++ doesn't keep annotations made in other editors, annotations are flattened into the PDF when the PDF is exported.

**Describe the solution you'd like**
I would like Xournal++ to treat annotations as first class citizens so that these annotations can be edited in other programs such as Okular.

**Describe alternatives you've considered**
Using Okular for PDF annotations and Xjournal++ for handwritten notes. Would be nice to do things in one place.

**Additional context**
Example using highlighting.
1. Open PDF in Okular and highlight words. Notice that Okular keeps track of the annotations in the annotations section where they can be deleted or edited.
![image](https://user-images.githubusercontent.com/1094134/101227452-ed8d7680-3665-11eb-8800-9ad7f70cfd7c.png)
2. Save the edited PDF file and exit Okular.
3. Open the same PDF file in Microsoft Edge. Edge can edit and delete annotations made in Okular. Also can add it's own highlight annotations.
![image](https://user-images.githubusercontent.com/1094134/101227706-cf744600-3666-11eb-9751-7788a719f73f.png)

4. Save the PDF and close Edge.
5. Open it back up in Okular. Okular can now edit and delete annotations made in Edge.
![image](https://user-images.githubusercontent.com/1094134/101227785-1d894980-3667-11eb-9c12-b992c09f8687.png)

Would be nice if Xournal++ can interact similarly with these apps.

IsaacWeiss:
I have a related, yet opposite, problem/request.

Image annotations are **not** fully flattened when exporting to PDF. For my use case, signing paperwork, this is a security concern.

I have a transparent PNG signature file, which I can position on the signature line of a form. I want to then flatten the image so that it cannot be extracted from the PDF (or, at least, so that it would be just as difficult to extract it as if it was a scan of a paper document I had signed with ink).

Whether I use `File > Export as PDF` or `File > Print > Print to File`, if I open the resulting document in LibreOffice Draw, the signature is recognized as an "image with transparency"; it can be selected, manipulated, and copied and pasted independent of the background. In other words, anyone receiving my signed form can get a beautiful transparent PNG of my signature to use as they wish.

I can open a separate issue if this is different enough from the original, just please let me know.

rolandlo:
@IsaacWeiss. I think this should be in a separate issue. You can still mention that it is related to this one.

[...]

Issue #3177, opened 2021-06-21
Don't use poppler for exporting to PDFs with background


piegamesde:
**Is your feature request related to a problem? Please describe.**

I've had to learn this the hard way too. Basically, the `page.render(context)` command is not a good one if context belongs to a `PdfSurface`. That command will translate the PDF into Cairo commands which on its turn translate it back into PDF commands. This is not faithful and inherently lossy. In best cases, only some re-formatting happens, making the PDF larger. In worst cases, the exported PDF is *wrong* and has glitches that were not there in the original.

**Describe the solution you'd like**
A more faithful approach would be to render the xopp file to an in-memory PDF, and then merge them with the background in a separate step that does not involve Poppler or Cairo anymore. I've hacked together a proof of concept [using pikepdf](https://github.com/pikepdf/pikepdf/issues/210), thus you should be able to get this done rather easily using QPDF.

**Describe alternatives you've considered**

- I haven't looked at other PDF rendering libraries, and how well they work. I think Poppler is capable of such things as well, but I don't know how much of this is exposed through public API.
- When sticking to Poppler+Cairo, their rasterization is fine. Of course, transforming the background to an image kind of defeats the purpose of using PDFs in the first place, but at least it would be faithful.

**Additional context**
(Coming soon)

mjg:
Note that you can export xopp without PDF background and then overlay everything with your tool of choice, which might be cumbersome unless you have the simple case of annotating all pages of a single PDF.

Are you suggesting to implement a different PDF export within xournalpp?

piegamesde:
> Are you suggesting to implement a different PDF export within xournalpp?

Kind of. The solution I see at the moment definitely involves a modification of the existing export code, yes. But there may also be other solutions to this that I'm not aware of.

> Note that you can export xopp without PDF background and then overlay everything with your tool of choice, which might be cumbersome unless you have the simple case of annotating all pages of a single PDF.

This is probably what I'm going to do as a workaround for now. But a proper solution would be preferable nevertheless.

tattsan:
Great work! I've tried to build the app with ENABLE_MUPDF flag. For testing, I've created a PDF generated by `pdflatex` with the following  LaTeX code.
```
\documentclass{article}
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
[...]

Issue #5736, opened 2024-06-15
CMakeLists unnecessarily break build on macOS by incorrect hardcoded settings


barracuda156:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

MacPorts

### Xournal++ Version

1.2.3

### libgtk Version

3.24.41

### Bug Description

CMakeLists currently use a few settings which break the build on macOS, unnecessarily so.
They can just be removed.

1. Using an archaic TargetArch module serves no purpose but breaks arm64 and ppc.
2. `rdynamic` should not be used on macOS.
3. Alleged ”fix” for pkg-config behavior breaks linking, since opt is undefined.
4. Deployment target is set to an arbitrarily recent macOS version. Hardcoding a deployment target is wrong.

### Expected Behaviour

The build should work without a need to hack CMakeLists.

### Steps to Reproduce

Run the build of v. 1.2.3 on Sonoma arm64 (will fail due to a broken TargetArch module).
Run the build on macOS earlier that hardcoded target.
Etc.

### Additional Context

Getting rid of breaking settings, the app can be built on much earlier macOS versions and support arm64 and ppc.
![xournalpp](https://github.com/xournalpp/xournalpp/assets/92015510/984a83ec-add0-4f65-b430-582de1f7b418)

rolandlo:
A MacOS ARM build has been released with version 1.2.4. Point 3 of your list has been addressed. Would you mind creating a pull request with the other changes you envision?

barracuda156:
@rolandlo Thank you. I can test the new version in a week (away from the hardware atm).

Is there any reason to hardcore deployment target? This obviously gonna break the app on any system prior to Catalina. If later OS need to fall back to 10.15 as the target, rather do it conditionally, perhaps?

rolandlo:
> @rolandlo Thank you. I can test the new version in a week (away from the hardware atm).
> 
> Is there any reason to hardcore deployment target? This obviously gonna break the app on any system prior to Catalina. If later OS need to fall back to 10.15 as the target, rather do it conditionally, perhaps?

For the current version of pcre2 on MacOS ARM that is required for building glib I got the error message "Must target Big Sur or newer" when setting deployment target lower than that. See also https://github.com/PCRE2Project/pcre2/issues/217 Not sure what is required for MacOS Intel.

[...]

PR #6364, opened 2025-03-31
Use alternative libraries for PDF exports

bhennion:
I played around a bit with alternative libraries for exporting PDF (see #3177)
You can try out two options (one via [PoDoFo](https://github.com/podofo/podofo/tree/master) (you'll need the latest dev version of this library) and one via [libmupdf](https://mupdf.com/))
They can be enabled via the cmake flags ENABLE_POFODO and ENABLE_MUPDF.

If enabled, you can export via PoDoFo or mupdf from the command line (only there for now) with the options --export-podofo and --export-mupdf (together with foo.xopp -p bar.pdf).

I'm happy to get any feedback on those attempts (especially if someone encounters a pdf/xopp file where the output is not as expected). The TOC, hyperlinks, Pdf-annotations and other data in the PDF should be preserved.

Forwarded message from #3177:

> Also, I (maybe wrongly) assumed that the new export backends would "pass through" the background PDF and put annotations on top. But, at least in this case, both cairo and mupdf increase the file size (though not terribly) with mupdf > cairo.

This is exactly what I did. The xopp-annotations are exported to PDF by cairo (without the pdf background) and added to the background pdf as an XObject (some pdf feature). I had made a mistake in the case the background PDF page already had such XObjects.
Now, the PDF file is opened, the annotations are added, and then the pdf is saved again. Under the hood, the library may decompress bit of data, do something on it, and resave it with a different compression level (see e.g. this striking https://github.com/podofo/podofo/issues/163). This may lead to size increases.

Another possible source of (ever so slight) size increase is with the fact that some internal structure is changed on purpose by my code: specifically, the content of a page can be stored in two ways. Either a single compressed stream or an array of streams. In order to add the overlayed xopp-annotations, I chose to always go with an array of streams, even if the original page had a single stream. This way, I can add by hand the stream necessary for displaying the xopp-annotation overlay without ever touching (decompressing) the original stream.
In a similar manner, I had to add some indirections between the page structures and some of their data to minimize size increase in the case where the user duplicated a PDF page (e.g. the xopp shows twice the first page of the PDF background).
All this may add a handful of bytes per page (maybe up to a kB/page, I don't really know).

One last thing is that pages are modified even is they have no annotations on them. This (and more) could/should be optimized out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -419,7 +419,7 @@ set(CPACK_DEBIAN_PACKAGE_RELEASE 2)
 set(CPACK_DEBIAN_PACKAGE_VERSION "${CPACK_PROJECT_VERSION}")
 set(CPACK_DEBIAN_PACKAGE_SECTION "graphics")
 set(CPACK_DEBIAN_PACKAGE_DEPENDS
-        "libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6")
+        "libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, qpdf (>= 10.6.0)")
 set(CPACK_DEBIAN_PACKAGE_SUGGESTS "texlive-base, texlive-latex-extra")  # Latex tool
 # Use debian's arch scheme; we only care about x86/amd64 for now but feel free to add more
 if (${PACKAGE_ARCH} STREQUAL "x86_64")
--- a/debian/control
+++ b/debian/control
@@ -3,12 +3,12 @@ Maintainer: Andreas Butti <andreasbutti@gmail.com>
 Section: misc
 Priority: optional
 Standards-Version: 4.5.1
-Build-Depends: debhelper (>= 11), cmake, help2man, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile1-dev | libsndfile-dev, liblua5.3-dev, libzip-dev, librsvg2-dev, libgtksourceview-4-dev
+Build-Depends: debhelper (>= 11), cmake, help2man, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile1-dev | libsndfile-dev, liblua5.3-dev, libzip-dev, librsvg2-dev, libgtksourceview-4-dev, libqpdf-dev (>= 10.6)
 Homepage: https://github.com/xournalpp/xournalpp/
 
 Package: xournalpp
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2-2 (>= 2.40), libgtksourceview-4-0
+Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.32), libgtk-3-0 (>= 3.18), libpoppler-glib8 (>= 0.41.0), libxml2 (>= 2.0.0), libportaudiocpp0 (>= 12), libsndfile1 (>= 1.0.25), liblua5.3-0, libzip4 (>= 1.0.1) | libzip5, zlib1g, libc6, librsvg2-2 (>= 2.40), libgtksourceview-4-0, qpdf (>= 10.6)
 Suggests: texlive-base, texlive-latex-extra
 Recommends: lua-lgi
 Description: Xournal++ - Open source hand note-taking program
--- a/.github/actions/install_deps_ubuntu/action.yml
+++ b/.github/actions/install_deps_ubuntu/action.yml
@@ -38,6 +38,7 @@ runs:
           help2man
           libgtest-dev
           libgtksourceview-4-dev
+          libqpdf-dev
       run: |
         sudo apt-get update
         sudo apt-get install -y ${{env.packages}}
--- a/.github/actions/install_deps_windows/action.yml
+++ b/.github/actions/install_deps_windows/action.yml
@@ -24,3 +24,4 @@ runs:
           mingw-w64-x86_64-gtksourceview4
           mingw-w64-x86_64-imagemagick
           mingw-w64-x86_64-lua-lgi
+          mingw-w64-x86_64-qpdf
--- a/mac-setup/CI_jhbuild.sh
+++ b/mac-setup/CI_jhbuild.sh
@@ -101,6 +101,14 @@ EOF
     sed -i '' -e 's/^\(module_cmakeargs\["freetype-no-harfbuzz"\]\) =/module_cmakeargs.setdefault("freetype-no-harfbuzz", ""); \1 +=/' ~/.config/jhbuildrc
 
     echo "interact = False" >> ~/.config/jhbuildrc
+    echo "exit_on_error = True" >> ~/.config/jhbuildrc
+    echo "shallow_clone = True" >> ~/.config/jhbuildrc
+    echo "use_local_modulesets = True" >> ~/.config/jhbuildrc
+    echo "disable_Werror = False" >> ~/.config/jhbuildrc
+
+    # MODULEFILE contains dependencies to other module sets - they need to be in the same directory
+    cp "$MODULEFILE" "$HOME/gtk-osx-custom/modulesets-stable/"
+    export MODULEFILE="$(basename $MODULEFILE)"
 }
 
 echo "::group::Setup jhbuild"
@@ -114,7 +122,7 @@ echo "::endgroup::"
 download() {
     jhbuild update $GTK_MODULES
     jhbuild -m "$MODULEFILE" update meta-xournalpp-deps
-    jhbuild -m ~/gtk-osx-custom/modulesets-stable/bootstrap.modules update meta-bootstrap
+    jhbuild -m bootstrap.modules update meta-bootstrap
     echo "Downloaded all jhbuild modules' sources"
 }
 echo "::group::Download modules' sources"
@@ -123,7 +131,7 @@ echo "::endgroup::"
 
 ### Step 3: bootstrap
 bootstrap_jhbuild() {
-    jhbuild -m ~/gtk-osx-custom/modulesets-stable/bootstrap.modules build --no-network meta-bootstrap
+    jhbuild -m bootstrap.modules build --no-network meta-bootstrap
 }
 echo "::group::Bootstrap jhbuild"
 bootstrap_jhbuild
--- a/mac-setup/xournalpp.bundle
+++ b/mac-setup/xournalpp.bundle
@@ -67,6 +67,11 @@
     ${prefix}/lib/girepository-1.0/*.typelib
   </data>
 
+  <!-- QPDF -->
+  <binary>
+    ${prefix}/lib/libqpdf*.dylib
+  </binary>
+
   <!-- Translation filenames, one for each program or library that you
        want to copy in to the bundle. The "dest" attribute is
        optional, as usual. Bundler will find all translations of that
--- a/mac-setup/xournalpp.modules
+++ b/mac-setup/xournalpp.modules
@@ -5,6 +5,8 @@
 
 <?xml-stylesheet type="text/xsl" href="moduleset.xsl"?>
 <moduleset>
+  <include href="gtk-osx.modules" />
+
   <repository type="tarball" name="github-tarball" href="https://github.com/"/>
   <repository type="tarball" name="poppler" href="https://poppler.freedesktop.org/"/>
   <repository type="tarball" name="libzip" href="https://libzip.org/download/"/>
@@ -94,6 +96,18 @@
     </dependencies>
   </cmake>
 
+  <cmake id="libqpdf" cmakeargs="-DBUILD_SHARED_LIBS=true -DCMAKE_BUILD_TYPE=RelWithDebInfo">
+    <branch repo="github-tarball"
+            version="12.0.0"
+            module="qpdf/qpdf/archive/refs/tags/v12.0.0.tar.gz"
+            checkoutdir="libqpdf-12.0.0"/>
+    <dependencies>
+      <dep package="zlib"/>
+      <dep package="libjpeg"/>
+      <dep package="openssl"/>
+    </dependencies>
+  </cmake>
+
   <autotools id="lua"
              skip-autogen="true"
              supports-non-srcdir-builds="no"
@@ -136,6 +150,7 @@
 
   <metamodule id="meta-xournalpp-deps">
     <dependencies>
+      <dep package="libqpdf"/>
       <dep package="poppler"/>
       <dep package="libzip"/>
       <dep package="portaudio"/>
--- a/readme/LinuxBuild.md
+++ b/readme/LinuxBuild.md
@@ -18,35 +18,35 @@ The minimum required CMake version is 3.13, but we recommend using >=3.15. Also,
 ```sh
 sudo pacman -S cmake gtk3 base-devel libxml2 portaudio libsndfile \
   poppler-glib texlive-bin texlive-pictures gettext libzip lua53 lua53-lgi \
-  gtksourceview4 help2man
+  gtksourceview4 help2man qpdf
 ```
 
 #### For Fedora:
 ```sh
 sudo dnf install gcc-c++ cmake gtk3-devel libxml2-devel portaudio-devel libsndfile-devel \
   poppler-glib-devel texlive-scheme-basic texlive-dvipng 'tex(standalone.cls)' gettext libzip-devel \
-  librsvg2-devel lua-devel lua-lgi gtksourceview4-devel help2man
+  librsvg2-devel lua-devel lua-lgi gtksourceview4-devel help2man qpdf-devel
 ```
 
 #### For CentOS/RHEL:
 ```sh
 sudo dnf install gcc-c++ cmake gtk3-devel libxml2-devel cppunit-devel portaudio-devel libsndfile-devel \
   poppler-glib-devel texlive-scheme-basic texlive-dvipng 'tex(standalone.cls)' gettext libzip-devel \
-  librsvg2-devel gtksourceview4-devel
+  librsvg2-devel gtksourceview4-devel qpdf-devel
 ```
 
 #### For Ubuntu/Debian and Raspberry Pi OS:
 ```sh
 sudo apt-get install cmake libgtk-3-dev libpoppler-glib-dev portaudio19-dev libsndfile-dev \
   dvipng texlive libxml2-dev liblua5.3-dev libzip-dev librsvg2-dev gettext lua-lgi \
-  libgtksourceview-4-dev help2man
+  libgtksourceview-4-dev help2man libqpdf-dev
 ```
 
 #### For openSUSE:
 ```sh
 sudo zypper install gcc-c++ cmake gtk3-devel portaudio-devel libsndfile-devel \
   texlive-dvipng texlive libxml2-devel libpoppler-glib-devel libzip-devel librsvg-devel lua-devel lua-lgi \
-  gtksourceview4-devel lsb-release help2man
+  gtksourceview4-devel lsb-release help2man qpdf-devel
 ```
 
 #### For Solus:
@@ -54,7 +54,7 @@ sudo zypper install gcc-c++ cmake gtk3-devel portaudio-devel libsndfile-devel \
 sudo eopkg it -c system.devel
 sudo eopkg it cmake libgtk-3-devel libxml2-devel poppler-devel libzip-devel \
   portaudio-devel libsndfile-devel alsa-lib-devel lua-devel \
-  librsvg-devel gettext libgtksourceview-devel
+  librsvg-devel gettext libgtksourceview-devel qpdf-devel
 ```
 
 ## Building and Testing
--- a/readme/WindowsBuild.md
+++ b/readme/WindowsBuild.md
@@ -46,7 +46,8 @@ pacman -S \
   mingw-w64-x86_64-libsndfile \
   mingw-w64-x86_64-libzip \
   mingw-w64-x86_64-lua \
-  mingw-w64-x86_64-portaudio
+  mingw-w64-x86_64-portaudio \
+  mingw-w64-x86_64-qpdf
 ```
 
 If prompted, confirm or use all default values.
--- a/windows-setup/package.sh
+++ b/windows-setup/package.sh
@@ -85,6 +85,10 @@ mkdir -p "$setup_dir"/share/lua/5.4
 cp "$prefix"/share/lua/5.4/lgi.lua "$setup_dir"/share/lua/5.4
 cp -r "$prefix"/share/lua/5.4/lgi/ "$setup_dir"/share/lua/5.4
 
+echo "copy qpdf"
+cp "$prefix"/bin/libqpdf*.dll "$setup_dir"/bin
+cp "$prefix"/lib/libqpdf* "$setup_dir"/lib
+
 echo "create installer"
 bash make_version_nsh.sh
 "/c/Program Files (x86)/NSIS/Bin/makensis.exe" xournalpp.nsi


mjg:
I tried building with podofo-1.0.0-beta1 - do you need even newer (git master)? Seems so, given that some structures/types are off.

mjg:
OK, that (beta2) works. Noticed:
- podofo inserts a "modified date", mupdf does not
- podofo spits out a bunch of `Fixing references in 26 0 R by 25` lines, cairo and mupdf do not

tattsan:
Thanks! Cmake passed cleanly.
There still remains non-exported xopp-annotation issue, in case 'duplicated PDF page with XObjects'.

src/core/pdf/base/MuPdfExport.cpp
```diff
@@ -153,20 +156,21 @@ bool MuPdfExport::overlayAndSave(const fs::path& saveDestination, std::stringstr
...
                     auto resources = page.pdf_page_resources();
-                    auto xobjects = resources.pdf_dict_gets("XObject");
+                    auto xobjects = resources.pdf_dict_gets(XOBJECT_KEY);
                     if (!xobjects) {
-                        xobjects = resources.pdf_dict_puts_dict("XObject", 1);
+                        xobjects = resources.pdf_dict_puts_dict(XOBJECT_KEY, 1);
                     } else if (xobjects.pdf_is_indirect()) {
                         // We need to copy the XObject dict so other copies of this page may have a different overlay
                         xobjects = xobjects.pdf_copy_dict();
-                        page.pdf_page_resources().pdf_dict_puts("Resources", xobjects);
```
mjg: This line is the actual fix, right?

tattsan:
Thanks! I've confirmed that the bug has been fixed.

rolandlo:
Great job @bhennion ! That looks like a great feature.

I have found one issue though: Exporting a 2-page xopp-file with PDF background I got a 3-page document using podofo and the 3rd page is slightly larger than the others. With another xopp-file (without PDF background), I didn't get an additional page.

rolandlo:
> > I have found one issue though: Exporting a 2-page xopp-file with PDF background I got a 3-page document using podofo and the 3rd page is slightly larger than the others. With another xopp-file (without PDF background), I didn't get an additional page.
> 
> Thank all of you for the testing!
> 
> This issue should be fixed now (can you confirm that you had not annotated any page when you encountered this?).

I can confirm that the issue is fixed and that I hadn't annotated any of those two pages.

mjg:
On Fedora, mupdf-cpp-libs+deps weighs in at 55MiB when installed. 6.2MiB are leptonica+tesseract, 46.7MiB mupdf-libs. I think the main reason is embedded fonts (which are trimmed down a bit in Fedora compared to the default) and the overall functionality.

From a packager and user perspective, the ideal solution would be export plugins which can be built as subpackages and installed separately. For a mupdf/zathura/qpdfview/... user, the xournalpp-mupdf-plugin does not pull in any extra dependencies which they haven't installed already. Others may prefer other export plugins. Subpackages would give everyone their choice.

I know this requires extra effort, but since you've abstracted the interface already ...  ;-)

rolandlo:
Another question is what do we do on platforms where we ship the whole thing (Windows, MacOS, Flatpak, Snap). Do we include any of those PDF libraries?

mjg:
```
[ 78%] Building CXX object src/core/CMakeFiles/xournalpp-core.dir/pdf/base/QPdfExport.cpp.o
In file included from /usr/include/qpdf/Buffer.hh:24,
                 from /usr/include/qpdf/QPDF.hh:36,
                 from /home/mjg/src/xournalpp/src/core/pdf/base/QPdfExport.cpp:10:
/usr/include/qpdf/PointerHolder.hh:32:3: warning: #warning "POINTERHOLDER_TRANSITION is not defined -- see qpdf/PointerHolder.hh" [-Wcpp]
   32 | # warning "POINTERHOLDER_TRANSITION is not defined -- see qpdf/PointerHolder.hh"
      |   ^~~~~~~
```
and then
```
/home/mjg/src/xournalpp/src/core/pdf/base/QPdfExport.cpp: In function ‘void reorderBackgrounds(QPDF&, const std::vector<HybridPdfExport::OutputPageInfo>&)’:
/home/mjg/src/xournalpp/src/core/pdf/base/QPdfExport.cpp:104:49: error: cannot convert ‘__gnu_cxx::__alloc_traits<std::allocator<QPDFPageObjectHelper>, QPDFPageObjectHelper>::value_type’ {aka ‘QPDFPageObjectHelper’} to ‘const QPDFObjectHandle&’
  104 |             if (!ps[nbValidPages].isSameObjectAs(backgroundPages[n])) {
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/qpdf/QPDF.hh:42:
/usr/include/qpdf/QPDFObjectHandle.hh:311:25: note: initializing argument 1 of ‘bool QPDFObjectHandle::isSameObjectAs(const QPDFObjectHandle&) const’
  311 |     bool isSameObjectAs(QPDFObjectHandle const&) const;
      |                         ^~~~~~~~~~~~~~~~~~~~~~~
```
This is with `qpdf-devel 11.10.1` on Fedora 42. Maybe differently placed `const` (value/ref?

mjg:
Regarding info: We could view this as two modes.
1. defaut/cairo: xournalpp is the creator, all dates and info refer to that (xournalpp document, potentially with other elements such as PDF)
2. preserve-pdf/qpdf: xournalpp merely adds annotations but preserves the PDF, including its info dict, and merely adds modification info (date+?)
Dunno whether keeping all backends (which would be nice) would be worth (your) maintenance effort.

rolandlo:
> This is ready for a review. I've tested locally and the AppImage. Testing the MacOS/Windows installers would be welcome (@rolandlo? Would you have time?).
> 
> > Dunno whether keeping all backends (which would be nice) would be worth (your) maintenance effort.
> 
> It remains this question. I think for now we can keep the code in. If people tinker enough to go and try the other backends (so PoDoFo and mupdf), they may help maintaining as well. If the maintaining becomes a burden, we can always drop the backends then. It may also be of help to keep the backends there in case QPDF turns out not to fit the bill.

Yes, I will find some time on Thursday or Friday.

tattsan:
When doing the cairo-export as before  without `--export-pdf-backend=cairo` option, Xpp would crash. But when I specified a non-existend backend like `--export-pdf-backend=foo` the export succeeded.

mjg:
Yep, that's fixed.

podofo is still spitting out `Fixing references in 45 0 R by 25` lines for my test xopp with PDF background (from LibreOffice).

I have one build with mupdf (on my host system), one build with all three extra exporters (in a distrobox container). "default" is picked up correctly. Export file sizes differ by 1% for the two when both use mupdf - interesting, but not critical. (The build environments are similar, but not identical.)

debian/control
```diff
@@ -3,12 +3,12 @@ Maintainer: Andreas Butti <andreasbutti@gmail.com>
 Section: misc
 Priority: optional
 Standards-Version: 4.5.1
-Build-Depends: debhelper (>= 11), cmake, help2man, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile1-dev | libsndfile-dev, liblua5.3-dev, libzip-dev, librsvg2-dev, libgtksourceview-4-dev
+Build-Depends: debhelper (>= 11), cmake, help2man, git, libgtk-3-dev, libpoppler-glib-dev, libxml2-dev, portaudio19-dev, libsndfile1-dev | libsndfile-dev, liblua5.3-dev, libzip-dev, librsvg2-dev, libgtksourceview-4-dev, libqpdf-devel (>= 10.6)
```
rolandlo, line 6: Isn't the package named `libqpdf-dev`?

rolandlo (commented):
The qpdf backend works fine on Windows, Linux and MacOS both from CLI and GUI. I have two suggestions though:
1) When you specify e.g. mupdf on the CLI and mupdf is not available (instead the default is used) there should be some warning like when you specify a non-existing backend name
2) The Lua API  `applib_export` should also allow to specify the backend

CMakeLists.txt
```diff
@@ -141,6 +141,33 @@ pkg_check_modules(ExternalModules REQUIRED IMPORTED_TARGET ${pkg-module-spec})
...
 
+option(ENABLE_PODOFO "Compile with PoDoFo export support (Requires PoDoFo >= 1.0.0 beta 2)" OFF)
+if (ENABLE_PODOFO)
+    find_package(PoDoFo REQUIRED)
+    message(STATUS "Found PoDoFo: ${PoDoFo_FOUND}")
+endif ()
+
+option(ENABLE_MUPDF "Compile with MuPDF export support" OFF)
+if (ENABLE_MUPDF)
+    find_library(MUPDF NAMES libmupdfcpp* REQUIRED)
```
tattsan, line 152: This doesn't pass in my environment (Arch Linux). Instead, the following works.
```suggestion
    find_library(MUPDF NAMES mupdfcpp REQUIRED)
```
Does `find_library` support wildcards?

rolandlo:
> > 3\. The Lua API  `applib_export` should also allow to specify the backend
> 
> This could indeed be done, but in another PR (and perhaps by someone else). There should also be a function to list the available backends.

Ok, you can leave it. I will add it once this PR is merged.

rolandlo:
> Are we happy with the code? I tried to comment enough, but I'm guessing bits of it may be hard to read without having looked at the PDF specifications.

I haven't had the time to study `QPdfExport.cpp`, `PoDoFoPdfExport.cpp` and `MuPdfExport.cpp` in any detail yet. If it's not too urgent I would like to spend some more time with it. The changes to the other files look fine.

rolandlo:
One question: What are the legal consequences when incorporating any of those PDF libraries?
- The QPDF library stands under the Apache 2.0 license, which is only compatible with GPL v3, not GPL v2. We have GPL v2 or later.
- The PoDoFO library is licensed under LGPL v2 or later, which appears to be compatible with GPL v2 or later.
- The MuPdf library is licensed under AGPL v3 or later, which is compatible with GPL v3, but not directly with GPL v2.

Here's a [discussion](https://opensource.stackexchange.com/questions/1357/can-i-link-a-apache-2-0-library-into-software-under-gplv2)

rolandlo:
> > One question: What are the legal consequences when incorporating any of those PDF libraries?
> > 
> > * The QPDF library stands under the Apache 2.0 license, which is only compatible with GPL v3, not GPL v2. We have GPL v2 or later.
> > * The PoDoFO library is licensed under LGPL v2 or later, which appears to be compatible with GPL v2 or later.
> > * The MuPdf library is licensed under AGPL v3 or later, which is compatible with GPL v3, but not directly with GPL v2.
> > 
> > Here's a [discussion](https://opensource.stackexchange.com/questions/1357/can-i-link-a-apache-2-0-library-into-software-under-gplv2)
> 
> Actually, the QPDF library is licensed under either Apache 2.0 or Artistic 2.0:
> 
> > Versions of qpdf prior to version 7 were released under the terms of version 2.0 of the Artistic License. At your option, you may continue to consider qpdf to be licensed under those terms. Please see the manual for additional information.
> 
> And Artistic 2.0 and GPL2 seem to be compatible: https://www.gnu.org/licenses/license-list.en.html#ArtisticLicense2

Indeed! I'm relieved.

rolandlo:
The code for the `QPdfExport` looks good. I have done some more tests after studying the code and those passed fine as well.
The `PoDoFo` backend has a problem with PDF background pages duplicated in the xopp-file that contain forms or annotations though. Only the first occurence shows the forms and annotations.

rolandlo:
> > The code for the `QPdfExport` looks good. I have done some more tests after studying the code and those passed fine as well. The `PoDoFo` backend has a problem with PDF background pages duplicated in the xopp-file that contain forms or annotations though. Only the first occurence shows the forms and annotations.
> 
> Thanks for the thorough testing! The PoDoFo backend now also duplicates annotations and form contents.

It's improved and forms seem to work fine. I still have some annotations that get lost in the Podofo export though. While trying to create a minimal example I run into this crash:

<details>

<summary>Crash log</summary>

```term
Date: Thu May  1 08:56:53 2025
Error: signal 11
Xournal++ version 1.2.6+dev
Git commit fe53bc63 from rolandlo/pr/bhennion/6364
Gtk version 3.24.49

[bt]: (0) inst/bin/xournalpp(+0x108e87) [0x5c590d515e87]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x45810) [0x75008ce45810]
[bt]: (2) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfName10GetRawDataEv+0x14) [0x75008e4a18f2]
[bt]: (3) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo17PdfNameInequalityclERKNS_7PdfNameERKSt17basic_string_viewIcSt11char_traitsIcEE+0x25) [0x75008e3cfc5b]
[bt]: (4) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt8_Rb_treeIN6PoDoFo7PdfNameESt4pairIKS1_NS0_9PdfObjectEESt10_Select1stIS5_ENS0_17PdfNameInequalityESaIS5_EE17_M_lower_bound_trISt17basic_string_viewIcSt11char_traitsIcEEvEESt23_Rb_tree_const_iteratorIS5_ERKT_+0x6b) [0x75008e3d2095]
[bt]: (5) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt8_Rb_treeIN6PoDoFo7PdfNameESt4pairIKS1_NS0_9PdfObjectEESt10_Select1stIS5_ENS0_17PdfNameInequalityESaIS5_EE10_M_find_trISt17basic_string_viewIcSt11char_traitsIcEEvEESt23_Rb_tree_const_iteratorIS5_ERKT_+0x37) [0x75008e3d12c5]
[bt]: (6) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt3mapIN6PoDoFo7PdfNameENS0_9PdfObjectENS0_17PdfNameInequalityESaISt4pairIKS1_S2_EEE4findISt17basic_string_viewIcSt11char_traitsIcEEEEDTcldtdtdefpT4_M_t10_M_find_trfp_EERKT_+0x27) [0x75008e3d07fd]
[bt]: (7) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary6getKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x3a) [0x75008e3ce666]
[bt]: (8) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary7findKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x37) [0x75008e3ce6f9]
[bt]: (9) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary7FindKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x27) [0x75008e3cef6b]
[bt]: (10) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfInfo21getStringFromInfoDictERKSt17basic_string_viewIcSt11char_traitsIcEE+0x41) [0x75008e4959eb]
[bt]: (11) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfInfo8GetTitleEv+0x48) [0x75008e4966f2]
[bt]: (12) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZN6PoDoFo11PdfMetadata17ensureInitializedEv+0x74) [0x75008e49ea9c]
[bt]: (13) /lib/x86_64-linux-gnu/libpodofo.so.3(_ZN6PoDoFo11PdfMetadata11SetProducerENS_8nullableIRKNS_9PdfStringEvEE+0x2f) [0x75008e49e0e1]
[bt]: (14) inst/bin/xournalpp(_ZN15PoDoFoPdfExport14overlayAndSaveERKNSt10filesystem7__cxx114pathERNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEERKSt6vectorIN15HybridPdfExport14OutputPageInfoESaISE_EE+0xb55) [0x5c590d6c77f5]
[bt]: (15) inst/bin/xournalpp(_ZN15HybridPdfExport9createPdfERKNSt10filesystem7__cxx114pathERKSt6vectorI17ElementRangeEntrySaIS6_EEb+0x58c) [0x5c590d6dfc0c]
[bt]: (16) inst/bin/xournalpp(_ZN15CustomExportJob3runEv+0xc6) [0x5c590d5f8cc6]
[bt]: (17) inst/bin/xournalpp(_ZN11BlockingJob7executeEv+0xe) [0x5c590d5f5fbe]
[bt]: (18) inst/bin/xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0xd0) [0x5c590d525a20]
[bt]: (19) /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0xa64cb) [0x75008ed8f4cb]
[bt]: (20) /lib/x86_64-linux-gnu/libc.so.6(+0xa2ef1) [0x75008cea2ef1]
[bt]: (21) /lib/x86_64-linux-gnu/libc.so.6(+0x13445c) [0x75008cf3445c]


Try to get a better stacktrace...
[bt] #0 inst/bin/xournalpp(+0x10954a) [0x5c590d51654a]
crashHandler(int) bei /home/roland/dev/Reviews/xournalpp/src/core/control/CrashHandlerUnix.h:157
[bt] #1 /lib/x86_64-linux-gnu/libc.so.6(+0x45810) [0x75008ce45810]
__restore_rt bei libc_sigaction.c:?
[bt] #2 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfName10GetRawDataEv+0x14) [0x75008e4a18f2]
PoDoFo::PdfName::GetRawData() const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfName.cpp:329
[bt] #3 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo17PdfNameInequalityclERKNS_7PdfNameERKSt17basic_string_viewIcSt11char_traitsIcEE+0x25) [0x75008e3cfc5b]
PoDoFo::PdfNameInequality::operator()(PoDoFo::PdfName const&, std::basic_string_view<char, std::char_traits<char> > const&) const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfName.h:185 (discriminator 1)
[bt] #4 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt8_Rb_treeIN6PoDoFo7PdfNameESt4pairIKS1_NS0_9PdfObjectEESt10_Select1stIS5_ENS0_17PdfNameInequalityESaIS5_EE17_M_lower_bound_trISt17basic_string_viewIcSt11char_traitsIcEEvEESt23_Rb_tree_const_iteratorIS5_ERKT_+0x6b) [0x75008e3d2095]
std::_Rb_tree_const_iterator<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> > std::_Rb_tree<PoDoFo::PdfName, std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject>, std::_Select1st<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> >, PoDoFo::PdfNameInequality, std::allocator<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> > >::_M_lower_bound_tr<std::basic_string_view<char, std::char_traits<char> >, void>(std::basic_string_view<char, std::char_traits<char> > const&) const bei /usr/include/c++/14/bits/stl_tree.h:1334 (discriminator 2)
[bt] #5 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt8_Rb_treeIN6PoDoFo7PdfNameESt4pairIKS1_NS0_9PdfObjectEESt10_Select1stIS5_ENS0_17PdfNameInequalityESaIS5_EE10_M_find_trISt17basic_string_viewIcSt11char_traitsIcEEvEESt23_Rb_tree_const_iteratorIS5_ERKT_+0x37) [0x75008e3d12c5]
std::_Rb_tree_const_iterator<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> > std::_Rb_tree<PoDoFo::PdfName, std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject>, std::_Select1st<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> >, PoDoFo::PdfNameInequality, std::allocator<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> > >::_M_find_tr<std::basic_string_view<char, std::char_traits<char> >, void>(std::basic_string_view<char, std::char_traits<char> > const&) const bei /usr/include/c++/14/bits/stl_tree.h:1302
[bt] #6 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNKSt3mapIN6PoDoFo7PdfNameENS0_9PdfObjectENS0_17PdfNameInequalityESaISt4pairIKS1_S2_EEE4findISt17basic_string_viewIcSt11char_traitsIcEEEEDTcldtdtdefpT4_M_t10_M_find_trfp_EERKT_+0x27) [0x75008e3d07fd]
decltype ((((*this)._M_t)._M_find_tr)({parm#1})) std::map<PoDoFo::PdfName, PoDoFo::PdfObject, PoDoFo::PdfNameInequality, std::allocator<std::pair<PoDoFo::PdfName const, PoDoFo::PdfObject> > >::find<std::basic_string_view<char, std::char_traits<char> > >(std::basic_string_view<char, std::char_traits<char> > const&) const bei /usr/include/c++/14/bits/stl_map.h:1250
[bt] #7 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary6getKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x3a) [0x75008e3ce666]
PoDoFo::PdfDictionary::getKey(std::basic_string_view<char, std::char_traits<char> > const&) const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfDictionary.cpp:169
[bt] #8 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary7findKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x37) [0x75008e3ce6f9]
PoDoFo::PdfDictionary::findKey(std::basic_string_view<char, std::char_traits<char> > const&) const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfDictionary.cpp:178 (discriminator 1)
[bt] #9 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo13PdfDictionary7FindKeyERKSt17basic_string_viewIcSt11char_traitsIcEE+0x27) [0x75008e3cef6b]
PoDoFo::PdfDictionary::FindKey(std::basic_string_view<char, std::char_traits<char> > const&) const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfDictionary.cpp:312
[bt] #10 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfInfo21getStringFromInfoDictERKSt17basic_string_viewIcSt11char_traitsIcEE+0x41) [0x75008e4959eb]
PoDoFo::PdfInfo::getStringFromInfoDict(std::basic_string_view<char, std::char_traits<char> > const&) const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfInfo.cpp:47 (discriminator 2)
[bt] #11 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZNK6PoDoFo7PdfInfo8GetTitleEv+0x48) [0x75008e4966f2]
PoDoFo::PdfInfo::GetTitle() const bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfInfo.cpp:141
[bt] #12 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZN6PoDoFo11PdfMetadata17ensureInitializedEv+0x74) [0x75008e49ea9c]
PoDoFo::PdfMetadata::ensureInitialized() bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfMetadata.cpp:320
[bt] #13 /lib/x86_64-linux-gnu/libpodofo.so.3(_ZN6PoDoFo11PdfMetadata11SetProducerENS_8nullableIRKNS_9PdfStringEvEE+0x2f) [0x75008e49e0e1]
PoDoFo::PdfMetadata::SetProducer(PoDoFo::nullable<PoDoFo::PdfString const&, void>) bei /home/roland/dev/podofo-1.0.0-beta2/src/podofo/main/PdfMetadata.cpp:146
[bt] #14 inst/bin/xournalpp(_ZN15PoDoFoPdfExport14overlayAndSaveERKNSt10filesystem7__cxx114pathERNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEERKSt6vectorIN15HybridPdfExport14OutputPageInfoESaISE_EE+0xb55) [0x5c590d6c77f5]
PoDoFoPdfExport::overlayAndSave(std::filesystem::__cxx11::path const&, std::__cxx11::basic_stringstream<char, std::char_traits<char>, std::allocator<char> >&, std::vector<HybridPdfExport::OutputPageInfo, std::allocator<HybridPdfExport::OutputPageInfo> > const&) bei /home/roland/dev/Reviews/xournalpp/src/core/pdf/base/PoDoFoPdfExport.cpp:216
[bt] #15 inst/bin/xournalpp(_ZN15HybridPdfExport9createPdfERKNSt10filesystem7__cxx114pathERKSt6vectorI17ElementRangeEntrySaIS6_EEb+0x58c) [0x5c590d6dfc0c]
HybridPdfExport::createPdf(std::filesystem::__cxx11::path const&, std::vector<ElementRangeEntry, std::allocator<ElementRangeEntry> > const&, bool) bei /home/roland/dev/Reviews/xournalpp/src/core/pdf/base/HybridPdfExport.cpp:96
[bt] #16 inst/bin/xournalpp(_ZN15CustomExportJob3runEv+0xc6) [0x5c590d5f8cc6]
CustomExportJob::run() bei /home/roland/dev/Reviews/xournalpp/src/core/control/jobs/CustomExportJob.cpp:145 (discriminator 1)
[bt] #17 inst/bin/xournalpp(_ZN11BlockingJob7executeEv+0xe) [0x5c590d5f5fbe]
BlockingJob::execute() bei /home/roland/dev/Reviews/xournalpp/src/core/control/jobs/BlockingJob.cpp:20
[bt] #18 inst/bin/xournalpp(_ZN9Scheduler17jobThreadCallbackEPS_+0xd0) [0x5c590d525a20]
Scheduler::jobThreadCallback(Scheduler*) bei /home/roland/dev/Reviews/xournalpp/src/core/control/jobs/Scheduler.cpp:248
[bt] #19 /lib/x86_64-linux-gnu/libglib-2.0.so.0(+0xa64cb) [0x75008ed8f4cb]
g_once_init_leave_pointer bei ??:?
[bt] #20 /lib/x86_64-linux-gnu/libc.so.6(+0xa2ef1) [0x75008cea2ef1]
start_thread bei ./nptl/./nptl/pthread_create.c:448
[bt] #21 /lib/x86_64-linux-gnu/libc.so.6(+0x13445c) [0x75008cf3445c]
__GI___clone3 bei ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:80


Execution log:

8146999198µs: DEBUG   : GLib-GIO :: _g_io_module_get_default: Found default implementation gvfs (GDaemonVfs) for ‘gio-vfs’
8147023138µs: DEBUG   : GLib :: unsetenv() is not thread-safe and should not be used after threads are created
8147025815µs: DEBUG   : xopp :: TEXTDOMAINDIR = (null), Platform-specific locale dir = /home/roland/dev/Reviews/xournalpp/build/inst/share/xournalpp/../locale, chosen directory = /home/roland/dev/Reviews/xournalpp/build/inst/share/xournalpp/../locale
8147026948µs: INFO    : xopp :: Loading plugins from: /home/roland/dev/Reviews/xournalpp/build/inst/share/xournalpp/plugins
8147027100µs: INFO    : xopp :: Loading plugins from: /home/roland/.config/xournalpp/plugins
8147027937µs: MESSAGE : xopp :: Plugin "FunctionPlotter" UI initialized
8147027977µs: MESSAGE : xopp :: Plugin "ImageActions" UI initialized
8147027987µs: MESSAGE : xopp :: Plugin "MyPlugin" UI initialized
8147079966µs: DEBUG   : xopp :: Extracted theme info: Name = Yaru-purple-dark, rootname = Yaru-purple, dark = false
8147101233µs: DEBUG   : xopp :: Theme name: Yaru-purple-dark
8147101240µs: DEBUG   : xopp :: Modified in GtkSettings: false
8147101241µs: DEBUG   : xopp :: Theme variant: dark
8147101242µs: DEBUG   : xopp :: Icon theme: iconsColor
8147115460µs: MESSAGE : xopp :: Add toolbar button with id: Plugin::FINELINER and icon: document-edit
8151342786µs: DEBUG   : GLib-GIO :: _g_io_module_get_default: Found default implementation dconf (DConfSettingsBackend) for ‘gsettings-backend’
8154510048µs: MESSAGE : xopp :: Datei: /home/roland/dev/code-analysis/qpdf/part.pdf
8180026796µs: WARNING : xopp :: [Crash Handler] Crashed with signal 11
8180036446µs: WARNING : xopp :: [Crash Handler] Wrote crash log to: /home/roland/.cache/xournalpp/errorlogs/errorlog.20250501-085653.log
```
</details>

Here is the PDF I used:

[part.pdf](https://github.com/user-attachments/files/19997320/part.pdf)

ceztko:
PoDoFo maintainer here. Few replies to your very interesting comments:

> See https://github.com/podofo/podofo/issues/253 for the report upstream

Should be fixed in master. See my comment [here](https://github.com/podofo/podofo/issues/253#issuecomment-2857851010).

> podofo inserts a "modified date", mupdf does not

You can disable metadata updating when saving with `PdfSaveOptions::NoMetadataUpdate`

> * podofo spits out a bunch of `Fixing references in 26 0 R by 25`

Oh, yes: this makes no sense at all. I demoted the log to Debug in https://github.com/podofo/podofo/commit/85c6e68111c8768cc68b0d683e3fdb4680e621f8 .

> and resave it with a different compression level (see e.g. this striking https://github.com/podofo/podofo/issues/163). This may lead to size increases.

This is sad but it will be definitely fixed at some point after 1.0 release. It just didn't escalate in my job so far, as I have other priorities.

> PoDoFo: C++ library. The interface is pretty intuitive, but it is young (still beta) and in active development

Actually it is not that young (first commit dates back to 2006). I'm the current maintainer after the original author retired. The stable release is 0.10.4 but 1.0 is coming soon (I hope 1-2 weeks away) and it will feature a robust (if not stable) API.

> It won't be readily available on ubuntu or similar LTS. 

We are in contact with Debian/Ubuntu devs so 1.0 will receive a fast track and 0.10.x releases skipped.

> we'll most likely encounter some bugs on weird PDF files

Some [broken PDFs](https://github.com/podofo/podofo/issues/242) (not standard compliant) are not supported and will require a recovery mode as implemented in other parsers. Unfortunately this is currently not my priority, so again it's something that will come after 1.0.

> The PoDoFO library is licensed under LGPL v2 or later, which appears to be compatible with GPL v2 or later.

PoDoFo will be soon relicensed to dual LGPL 2+/MPL2. Probably shortly after 1.0.

ceztko:
> I can confirm that podofo doesn't crash anymore on latest master when exporting the 1-page document provided. The annotations get lost though. Why is that?

Please open a new issue and provide a concise sample as the trivial code in https://github.com/podofo/podofo/issues/253 doesn't show up the issue (output.txt has a text annotation as in the original testfile.pdf).

ceztko:
I just noticed that I am not doing anything similar to [this](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-configuration-file), which it may be required to emit a functional `<package>Config.cmake`. As I said, I'm not using the feature myself, I just trusted existing code and merged patches. Again, I will try to evaluate this integration in time for 1.0.

rolandlo:
> > I can confirm that podofo doesn't crash anymore on latest master when exporting the 1-page document provided. The annotations get lost though. Why is that?
> 
> Please open a new issue and provide a concise sample as the trivial code in [podofo/podofo#253](https://github.com/podofo/podofo/issues/253) doesn't show up the issue (output.txt has a text annotation as in the original testfile.pdf).

Sorry, I made a mistake, using the wrong backend. The export works fine indeed with the latest podofo, not only with this simple document, but also with a more complex one.

ceztko:
I greatly [improved](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-version-file) creation of `podofo-config.cmake` by adding public header exports and added creation of `podofo-config-version.cmake`, which was missing. Such file convention is [supported](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-version-file) as well. I have no time to test them but at least everything should be in place now and PoDoFo should be findable as a package (provided the export config files are reachable) and be usable with just `target_link_libraries(target podofo::podofo)`. 1.0.0 is now very close to be released and there is only one last known critical bug waiting to be fixed.

ceztko:
> > What if I want the test the mechanism in windows so `find_package(PoDoFo 1.0.0 REQUIRED)` will work when installing the library in a arbitrary location?
> 
> I'm guessing this won't work.

Altering [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) does the trick! With latest changes in master now you can do:

```
cmake_minimum_required(VERSION 3.23)

project(PoDoFoSample)

set(CMAKE_CXX_STANDARD 17)

# In Windows, or if you installed the package to non
# standard paths
list(APPEND CMAKE_PREFIX_PATH "D:\\Test\\PoDoFoInstallDir")

# Alternative to altering CMAKE_PREFIX_PATH
#set(PoDoFo_DIR "D:\\Test\\PoDoFoInstallDir\\share\\cmake\\podofo")

find_package(PoDoFo 1.0.0 REQUIRED)
message(STATUS "Found PoDoFo: ${PoDoFo_VERSION}")

add_executable(helloworld main.cpp)
target_link_libraries(helloworld podofo::podofo)
```

It's the first time ever I try to locate libraries this way, I have to say it's an interesting mechanism for package managers. I am now tagging 1.0-rc1. I will inform you here when 1.0 final is released (few days).

ceztko:
PoDoFo 1.0.0 released ([announcement](https://www.mail-archive.com/podofo-users@lists.sourceforge.net/msg05005.html)).

Rehan-MALAK:
@bhennion 

Hi! I just wanted to thank you for your code. I just built your pr/podofo branch with ``-DENABLE_QPDF=yes``, and the exported PDF now includes the links ! I really missed that feature !

A typical use case for me is annotating math courses—mostly generated by LaTeX—with clickable table of contents, links to the bibliography, and links to previous definitions or theorems. This is a huge improvement!

Thanks again for your work!

bhennion:
@rolandlo What was the status of your review on this?

rolandlo:
> @rolandlo What was the status of your review on this?

Sorry, but I didn't find the time to study the PoDoFo implementation in detail yet. I plan to do so once the school leaving exams are over (end of next week). It worked fine last time I tested on all platforms.

Generally it seems we need to remove the mupdf backend due to the license issue. Moreover I would suggest to ship PoDoFo on Windows and MacOS (alongside QPdf) if that can be done without too much trouble. I could take of also shipping PoDoFo in the snap package and the flatpak once we release 1.3.0.

bhennion:
Alright, let met see about shipping podofo with all platforms. I'll remove the mupdf implementation as well (but keep it somewhere just in case, of course).

bhennion:
So, I managed to get it all to build with podofo, but there are issues:

1. Without packages out there (yet), we need to recompile podofo all the time, which doubles our CI time. MacOS already caches its dependencies, adding a cache for Linux build could be possible, but not really for Windows (see 2.)
2. On Windows, I had to build podofo as a static lib (otherwise, the test-units won't link). Maybe there is a workaround for that. An MSYS package would help here.
3. On MacOS, PoDoFo won't build without `-mmacosx-version-min=13.3` (or rather `-DCMAKE_OSX_DEPLOYMENT_TARGET=13.3`). Without this flag, you get
    ```PdfDeclarationsPrivate.cpp:1347:24: error: 'to_chars' is unavailable: introduced in macOS 13.3```
    Previously, the CMAKE_OSX_DEPLOYMENT_TARGET was set to 10.15 in our CMakeFiles, but upped to 11.7 in CI_jhbuild.sh. I don't really know how much users we'd be dropping by going from 11.7 to 13.3.

(not sure the installers generated by the runner will pack all that's needed yet, this needs to be looked at).

bhennion:
> 3\. I don't really know how much users we'd be dropping by going from 11.7 to 13.3.

See #5736 for a complain about CMAKE_OSX_DEPLOYMENT_TARGET being set to 10.15 already.

rolandlo:
> So, I managed to get it all to build with podofo, but there are issues:
> 
>     1. Without packages out there (yet), we need to recompile podofo all the time, which doubles our CI time. MacOS already caches its dependencies, adding a cache for Linux build could be possible, but not really for Windows (see 2.)
> 
>     2. On Windows, I had to build podofo as a static lib (otherwise, the test-units won't link). Maybe there is a workaround for that. An MSYS package would help here.
> 
>     3. On MacOS, PoDoFo won't build without `-mmacosx-version-min=13.3` (or rather `-DCMAKE_OSX_DEPLOYMENT_TARGET=13.3`). Without this flag, you get
>        `PdfDeclarationsPrivate.cpp:1347:24: error: 'to_chars' is unavailable: introduced in macOS 13.3`
>        Previously, the CMAKE_OSX_DEPLOYMENT_TARGET was set to 10.15 in our CMakeFiles, but upped to 11.7 in CI_jhbuild.sh. I don't really know how much users we'd be dropping by going from 11.7 to 13.3.
> 
> 
> (not sure the installers generated by the runner will pack all that's needed yet, this needs to be looked at).

Thanks for getting podofo to build. The issues are significant though and it may be better to postpone packaging podofo (split it off to a different PR that will be merged when the time is right). 
Raising the OSX deployment target from 11.7 to 13.3 would also mean that I can't test prebuilt bundles any more on my Macbook (I can only test it on the Mac Mini M2, i.e. on ARM), since Big Sur (MacOS 11) is the latest MacOS version running on it. MacOS 13 (Ventura) came out only 2 years and 8 months ago. Even though older versions are EOL there will still be plenty of users using older MacOS versions I would think, either because they have to (because of the hardware) or chose to not upgrade to a later version. Those old MacBooks are built pretty solidly and they work fine for a long time.

bhennion:
Here we go. I removed the podofo and mupdf backends from this branch. They are still available on [this branch](https://github.com/bhennion/xournalpp/tree/pr/podofo2).

rolandlo (approved):
LGTM! Sorry for the delay.

bhennion:
No worries! thanks for the review!

tockudex:
Does the PR fix #2486 ?

bhennion:
No. Xournal++'s data are stored on a separate PdfXObject that no apps would recognize as an annotation.

Issue #6313, opened 2025-03-05
how to avoid tex input from collapsing and losing line break when saved


bloboftext:
I have a pretty strange problem. I checked out a copy of xounal++ from feb 13 2024 on two different ubuntu machines, same hash c633d65375289d5ae12dbb3c1bd36cbdd76bd4b1

One is ubuntu 22.04, on this machine xournal++ will save all the tex equations without folding them, e.g.
```tex
v  = v_0 + at \\
v^2 = v_0^2 + \frac{1}{2} a t^2
```

On the ubuntu 24.04, when I open a saved .xopp file the tex formula all become concatenated into a single line, e.g.
```tex
v  = v_0 + at \\ v^2 = v_0^2 + \frac{1}{2} a t^2
```
Is there a code change I can make to force the tex source code to be saved and loaded preserving the line break? I looked at the source tree and it's pretty intimidating without a pointer to figure out where this could be. Thanks,

tattsan:
I also think the phenomenon that the line breaks are removed is inappropriate. Consider the following example:
```
\text{
\fbox{
a
}
and
\fbox{%
a%
}
are not equivalent.}
```
In TeX, inserting a line break without % results in a space being inserted, making these two cases non-equivalent. Therefore, when trying to explicitly indicate nested bracket structures using line breaks, inserting % is unavoidable. However, when saving the file and reloading it, this line break may sometimes be removed.
```
\text{\fbox{a}and\fbox{%a%}are not equivalent.}
```
When this happens, it's not just a matter of spaces being removed; everything to the right of % is treated as a comment, rendering the code invalid.

rolandlo:
> Any pointer to where the code saves the latex files and potentially strips away the carriage (0x0A)? Btw the tex code is not saved as clear text in the xopp files. I tried searching for the tex code in the xopp files, they cannot be found. They must have been scrambled somehow.

A TeXImage is saved in the form
```xml
<teximage text="(formula as a string)" left="(number)" top="(number)" right="(number)" bottom="(number)">
(base 64 decoded image string)
</teximage>
```

Saving a TeXImage in the SaveHandler happens like this:

https://github.com/xournalpp/xournalpp/blob/81d4064273825c621468c99f67a3f28379309966/src/core/control/xojfile/SaveHandler.cpp#L199-L210

Setting the formula happens through `TexImage:setText` via
[...]

Issue #4645, opened 2023-02-10
Newlines not saved properly for Tex


r2dev2:
**Affects versions :**
 - OS: Arch Linux with Kernel 6.1.3
 - (Linux only) Desktop environment: XFCE
 - Which version of libgtk do you use: 3.24.36
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Version of Xournal++: 1.1.3
     - can be obtained by running `xournalpp --version` or in the **About** window
 - Installation method: pacman

**Describe the bug**

Sometimes, when I use `\newline` and use shift + enter for a newline, the newline character is not saved in the file. This only seems to happen for files I have not opened in a while. Perhaps it is only for files I have exported, but I am not sure exactly when this happens.

Examining files which have the newline saved and files which do not have the newline saved, I find some interesting results.

For each of below screenshots, I have ran `cat file.xopp | gunzip | less` and searched for `<teximage>`

## Correct Newline

![correct newline](https://user-images.githubusercontent.com/50760816/218214661-85bfa0dd-a065-4761-89b7-d9387d7bd747.png)

In the file with correct newlines in tex, the sequence `&#13` is saved.

## Incorrect Newline

![incorrect newline](https://user-images.githubusercontent.com/50760816/218214836-e65298e8-1f87-478d-980b-b2bb6fa857f3.png)

In the file with incorrect newlines in tex, there is an unrecognized non-ascii character instead of the `&#13`.

**To Reproduce**

I keep trying to reproduce it in a controlled manner but have failed every time. The issue keeps popping up in all my documents. Each document is multi-page and has equations, pictures, and raw text. Below are my attempted steps to reproduce that maybe can be built upon to get the issue to appear.

I have [uploaded an xopp of this issue](https://github.com/r2dev2/bin/releases/download/v0.1.0/equationmonke.xopp) if that helps. In that notebook, the tex in red does not have correct shift+enter newlines but the tex in green has correct newlines.

![image](https://user-images.githubusercontent.com/50760816/218218213-09c7c962-dc01-430d-84c5-e4a3276d293d.png)

Attempted steps to reproduce the behavior:
1. create a new document with following latex
![image](https://user-images.githubusercontent.com/50760816/218215049-c16e83e9-3973-4534-ac23-ca78f377a2e6.png)
2. save the document
3. exit xournalpp
4. reopen xournalpp
5. export it to pdf
6. close xournalpp
7. open another document

**Expected behavior**
[...]

PR #6374, opened 2025-04-06
Fix linebreaks in LaTeX tool

### LINKED ISSUE #3060: LaTeX Dialog: Linebreaks Sometimes Disappear

**Affects versions :**
 - OS: Linux/Fedora 34, Linux/Manjaro
 - (Linux only) Desktop environment: GNOME 40, KDE Plasma
 - libgtk 3.24.28
 - Version of Xournal++: 1.1.0+dev built from #2809
 - Installation method: Built from source

**Describe the bug**
Despite #2849, linebreaks in LaTeX source still sometimes disappear. 

**To Reproduce**
1. Insert several large, multi-line LaTeX formulae, on several pages.
2. Save & close
3. Re-open
4. Continue editing and adding LaTeX
5. Edit one of the first LaTeX blocks.

**Expected behavior**
**Sometimes**, line breaks are missing from LaTeX blocks (but not always). It doesn't _seem_ to be caused by a load+save anymore...

**Additional context**
See #2849

tattsan:
Fix #3060, #4645, #6313.

Same PullRequest again. Sorry.


tattsan:
This file contains three TeX objects. They uses `\r`, `\n`, `\r\n` for linebreaks respectively.
[test_linebreaks.xopp.txt](https://github.com/user-attachments/files/19620296/test_linebreaks.xopp.txt)

tattsan:
Based on `TEST(ControlLoadHandler, imageSaveJpegBackwardCompat)` in 'LoadHandlerTest.cpp', I created `TEST(ControlLoadHandler, linebreaksLatex)`. For now, the test only targets the LaTeX element.

>    It might also be good to add a test case with this: save and reload a latex element (and a Text element, for good measure).

In the case of Text elements, line break characters are saved without being escaped. Unlike the line breaks inside LaTeX elements, this hasn't caused any issues so far. However, similar bugs might occur in the future.
To escape line break characters in Text elements, we would also need to modify 'src/core/control/xml/XmlTextNode.cpp' in a similar way to 'TextAttribute.cpp'. What do you think?

tattsan:
Done. Thanks for your kind instruction!

tattsan:
Squashed to one commit.

PR #6375, opened 2025-04-06
Fix regressions on master in export dialogs

bhennion:
While working on #6364, I noticed regressions on master in the export dialogs:
1. For Export As..., the file chooser did not disappear once the file was selected.
2. For all export dialogs: the "Do you want to overwrite?" test was not working when typing `foobar` in the file chooser to
export to `foobar.pdf`

This PR fixes those.


rolandlo (approved):
LGTM

Issue #4662, opened 2023-02-19
Improve stroke rendering performances


bhennion:
After some benchmarking (the source of which is available at https://github.com/bhennion/benchmarks/tree/main), I came to the conclusion that the way we display strokes is very very inefficient.

I made a benchmark using either
1. the current code from master on Xournal++: For each stroke segment, run
    ```
    cairo_set_line_wdth(cr, width);
    cairo_move_to(cr, ...);
    cairo_line_to(cr, ...);
    cairo_stroke(cr);
    ```
2. a new version, in which the boundary of the stroke is computed once (the computation is approximative in the benchmark), and then at each drawing operation, we feed the boundary to cairo and call
    ```
    cairo_fill(cr);
    ```

Running the above benchmark, I get the following unequivocal results on my laptop:
```
$ ./stroke-demo --xopp
Using backend "xlib"
1.26 fps
```
```
$ ./stroke-demo 
Using backend "xlib"
45.28 fps
```

This relates to https://github.com/xournalpp/xournalpp/issues/4112#issuecomment-1426854259 and the subsequent comment.

rolandlo:
Interesting. For me it was around 2.96 fps vs. 84.12 fps.

rolandlo:
With
```c++
for (auto it1 = stroke.path.begin(), it2 = std::next(it1); it2 != stroke.path.end() ; it1++, it2++) {
		cairo_set_line_width(cr, it1->z);
//		cairo_move_to(cr, it1->x, it1->y);
		cairo_line_to(cr, it2->x, it2->y);
}
cairo_stroke(cr);
```
I get 35.76 fps though.

bhennion:
> With
> 
> ```c++
> for (auto it1 = stroke.path.begin(), it2 = std::next(it1); it2 != stroke.path.end() ; it1++, it2++) {
> 		cairo_set_line_width(cr, it1->z);
> //		cairo_move_to(cr, it1->x, it1->y);
> 		cairo_line_to(cr, it2->x, it2->y);
> }
> cairo_stroke(cr);
> ```
> 
> I get 35.76 fps though.

It's not very surprising. I think that under the hood, every call to `cairo_stroke()` does two things: first compute the border of the stroke, depending on the parameters (join, cap, dashes, width) and then calls `cairo_fill()`.
The snippet you propose is faster because we only call `cairo_stroke()` once.
[...]

Issue #4112, opened 2022-06-23
Produces invalid svg


funnym0nk3y:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS: Win10 21H2
 - libgtk: 3.24.34
 - Version of Xournal++: 0d471f2e-dirty
 - Installation method: exe

**Describe the bug**
Exporting via export -> svg does not produce a valid svg file. The file cannot be opened with edge or inkscape (which even crashes). 

The file and the produced svg are attached below. 
[armijo rule.zip](https://github.com/xournalpp/xournalpp/files/8969567/armijo.rule.zip)

strajabot:
I am also having issues opening generated SVG files, but It may also be that the generated SVG code is too large for most programs to handle since my single page exports are around 35Mb

funnym0nk3y:
@strajabot That is not the case with my file. Yes, it is roughly 100x the size, but 4MB shouldn't be to bad.

Drakula44:
Hello, I would like to work on this. I understood the problem. SVG is not invalid the problem is that Cairo takes one stroke and turns it into a bunch of rectangles. SVG viewers just are not capable of rendering that many rectangles.
Instead, they are much more capable of rendering a stroke in SVG that takes many arguments at once.
`<path
       style="fill:#000000;stroke-width:0.264583"
       id="path183"
       d="M 183.6088,174.2278 C 183.6088,174.2278 183.6088,174.2278 183.6356,173.6477 \>`

As a solution, the approach should display a big path instead of a bunch of small rectangles. The first problem with that approach is that SVG does not support paths with varying widths.

That can be solved by calculating a stroke's outer edge and then drawing a closed path that will be filled inside. I have written an example script in python that transforms the xopp file into an SVG file.
The size of SVG is around 10x smaller than the xournalpp SVG export. And more importantly, it renders instantly.
When I took a look at the code I saw that the export job was delegated to Cairo. 

I will try to find a way if Cairo can do something similar to what I described. I saw that #4161 is refactoring export jobs but if I understood correctly it will not change this functionality.

bhennion:
You're more than welcome to give this a go. Even if this conflicted with #4161, that PR has been idle for a while.
Could PDF export also benefit from such a strategy? I assume the issue should be similar with any vector graphics export format.

In any case, it may a bit complicated to properly incorporate in the code base. You may need to modify StrokeView to have a vector graphics mode following the drawing pattern you're describing.
[...]

PR #6385, opened 2025-04-13
Improve rendering performances of strokes

bhennion:
This PR tries to improve rendering performances of strokes, following #4662.
As explained in #4662, the idea is to compute the contour of a stroke and call cairo_fill() only once per stroke (the current implementation calls cairo_stroke() - and hence cairo_fill() - once per point in the stroke).

The key is hence to compute the contour of a stroke. The contour means "some cairo_path on which to call cairo_fill() to obtain the desired stroke".
There are many different ways of getting a stroke from a sequence of points (x, y, pressure). I tried out two, so we have 3 rendering methods to discuss.

1. Current master.
2. Use every point's pressure value to make circles and add tangents:
Copie d'écran_20250413_100750
Make a contour out of that idea and call cairo_fill() only once.
3. Use a point's pressure to make an oblong shape going to the next point (the last point's pressure value is not used).
Copie d'écran_20250413_100438
Make a contour out of that idea and call cairo_fill() only once. This one gives the exact same output as current master's implementation.

You can see a bunch of unnatural test cases in this example file [example-v2.pdf](https://github.com/user-attachments/files/19721828/example.pdf) [example-v3.pdf](https://github.com/user-attachments/files/19721834/example-pp.pdf).
Version 2. is somehow more intuitive, but 3. (and 1.) is what we've been rendering so far. Of course, for a "normal" stroke, the difference is not really visible.

I benchmarked the three versions (you can try the [benchmarks](https://github.com/bhennion/benchmarks) on linux). I get (on a modest laptop with integrated GPU).

1. 1.4 fps for the current rendering method (from master) (`stroke-demo --xopp`)
2. 19.6 fps for implementation 2. (`stroke-demo`) (edit: latest version 24fps)
3. 17.6 fps for implementation 3. (`stroke-demo --pixel-precise`) (edit: latest version 19.3ps)

Importantly, the way we render strokes also has an effect on Exports. On an handwritten page, I get:

1. master: PDF 316 kB / SVG 39 MB  (see #4112 for the amazing 39 MB...)
2. implementation 2.: PDF 808kB / SVG 4.4 MB
3. implementation 3.: PDF 995kB / SVG 5.1 MB

To summarize, we have 3 rendering methods:
1. master version. Bad for screen rendering, bad for SVG. Produces smaller PDF exports (but more expensive the render)
2. Implementation 2. as above. Fastest so far, produces smallest SVG files but PDF files are significantly bigger than 1. (but cheaper to render - again, only one fill() operation per stroke)
3. Implementation 3. as above. Slightly slower than 2. (but still much much better than 1.). Produces larger SVG files and PDF files. The main argument for this one is: we do not change the output at all compared to master.

I tend to favor option 2. (for the efficiency and clean logic of the rendering) but a mix of 1. (for PDF exports) and 3. (for other renderings) could also be a good choice. @xournalpp/core I opened this draft PR with the following questions in mind:
- Would we care if the rendering output changed (in a way that would most likely never ever be noticed by a user)?
- Do we prefer smaller PDF outputs or bigger ones that are cheaper to render?

### DIFF
--- a/src/core/model/StrokeContour.cpp
+++ b/src/core/model/StrokeContour.cpp
@@ -1,0 +1,323 @@
+#include "StrokeContour.h"
+
+#include <cmath>
+
+#include "model/MathVect.h"
+#include "model/Point.h"
+#include "util/Assert.h"
+
+static_assert(std::numeric_limits<double>::is_iec559);  // Ensures atan2(0., 0.) does not error
+
+/// Takes a double between -2*pi and 2*pi and returns the corresponding angle between -pi and pi
+static constexpr double inMinusPiPiInterval(double t) {
+    xoj_assert(t >= -2 * M_PI && t <= 2 * M_PI);
+    return t > M_PI ? t - 2 * M_PI : t < -M_PI ? t + 2 * M_PI : t;
+}
+
+// Adds to cairo a segment corresponding to the cord of an arc obtained via cairo_arc() (with the same parameters)
+static void drawChord(cairo_t* cr, double cx, double cy, double radius, double angle1, double angle2) {
+    cairo_line_to(cr, cx + radius * std::cos(angle1), cy + radius * std::sin(angle1));
+    cairo_line_to(cr, cx + radius * std::cos(angle2), cy + radius * std::sin(angle2));
+}
+
+static void hookAfter(cairo_t* cr, double x, double y, double r, double a, double b) {
+    cairo_line_to(cr, x, y);
+    cairo_arc(cr, x, y, r, a, b);
+}
+static void hookBefore(cairo_t* cr, double x, double y, double r, double a, double b) {
+    cairo_arc(cr, x, y, r, a, b);
+    cairo_line_to(cr, x, y);
+}
+
+struct ReturnOp {
+    ReturnOp(void (*op)(cairo_t*, double, double, double, double, double), double x, double y, double r, double a,
+             double b):
+            op(op), x(x), y(y), r(r), a(a), b(b) {}
+    void (*op)(cairo_t*, double, double, double, double, double);
+    double x, y, r, a, b;
+
+    void operator()(cairo_t* cr) { op(cr, x, y, r, a, b); }
+};
+
+xoj::view::StrokeContour::StrokeContour(const std::vector<Point>& path): path(path) {}
+xoj::view::StrokeContour::~StrokeContour() = default;
+
+static inline void drawCoupling(cairo_t* cr, std::vector<ReturnOp>& ops, const Point& p2, double n1, double n3,
+                                double a1, double a3, double z1) {
+    if (z1 < p2.z) {
+        if (.5 * p2.z < n1  // Optimisation to avoid computing the std::hypot
+            || .5 * p2.z <= std::hypot(.5 * z1, n1)) [[likely]] {
+            double angleIn = std::asin(z1 / p2.z);
+            double a = inMinusPiPiInterval(a1 + angleIn);
+            double b = inMinusPiPiInterval(a3 - M_PI_2);
+            auto* arcFun =
+                    (inMinusPiPiInterval(a3 - a1) <= 0.0 || inMinusPiPiInterval(b - a) > 0.0) ? cairo_arc : drawChord;
+            arcFun(cr, p2.x, p2.y, .5 * p2.z, a, b);
+
+            double bp = inMinusPiPiInterval(a1 - angleIn);
+            double ap = inMinusPiPiInterval(a3 + M_PI_2);
+            auto* arcFunp =
+                    (inMinusPiPiInterval(a3 - a1) >= 0.0 || inMinusPiPiInterval(bp - ap) > 0.0) ? cairo_arc : drawChord;
+            ops.emplace_back(arcFunp, p2.x, p2.y, .5 * p2.z, ap, bp);
+        } else {
+            cairo_line_to(cr, p2.x, p2.y);
+            cairo_arc(cr, p2.x, p2.y, .5 * p2.z, a1, a3 - M_PI_2);
+            ops.emplace_back(hookBefore, p2.x, p2.y, .5 * p2.z, a3 + M_PI_2, a1);
+        }
+    } else {
+        if (.5 * z1 < n3  // Optimisation to avoid computing the std::hypot
+            || .5 * z1 <= std::hypot(.5 * p2.z, n3)) [[likely]] {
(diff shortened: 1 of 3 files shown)


rolandlo:
> * Would we care if the rendering output changed (in a way that would most likely never ever be noticed by a user)?

I don't think so. The rendering output would probably only look a bit more natural. 
The smallness of the output difference relies on the assumption that the pressure varies slowly, which may not be the case when adding strokes via a plugin, but even for plugins the rendering in 2 seems to make more sense in most situations.

> * Do we prefer smaller PDF outputs or bigger ones that are cheaper to render?

That's the harder question. I would probably go for the cheaper rendering. The impact looks more signficant and more in spirit of our goals: 
> Xournal++ is a hand note-taking software written in C++ with the target of
 flexibility, functionality and speed

So method 2 looks like the way to go for me. I will take a look how PDF sizes are affected for exports of the documents I write in class though.

I'm wondering how this will work out in Gtk 4 in case we eventually move to render node based rendering (using `gtk_snapshot_append_stroke`).

rolandlo:
> Alright. I went for the half/half solution involving 1. and 3. Even if 2. feels more natural, I think the export size of PDF files was just way too big. So I just kept using what master uses for PDF exports and used 3. for screen rendering. I think it matters that screen renderings and PDF exports give the same output (so I discarded the mixture of 1. and 2.).
> 
That sounds reasonable. There are no drawbacks compared to master and most of the rendering performance gain from 2 is there in 3 as well.

> On an implementation level, I kept debugging functions in. We may want to remove them.
> 
> And in terms of performances, this implementation is much faster than master's current one. In benchmarks, the FPS improvement seemed to depend on the ratio `average distance between consecutive points / average width`: it works better for elongated strokes than for fat ones. The bottleneck is pretty much the calls to std::hypot and std::asin. I do not see a way of getting rid of them (except the small optimization I already put in the code).
> 
> I think this is now ready for review and testing.

I'm particularly interested how this works out on MacOS, where rendering performance is still below par compared to Linux and Windows.

rolandlo:
\action create-installers macos windows

rolandlo:
\action create-installers mac-x64 mac-arm windows

chillibeanpaisley:
Look forward to latency improvements.
Suppose bottom line is that true real-time handwriting requires 20ms(equivalent to 50fps) at a minimum.
50fps may imply gpu compute code and pinned memory(no cpu calls).

Wonder what fps could be obtained by disabling pressure sensitivity and smoothing?
Would moving to vulkan provide much fps improvement with current code?

chillibeanpaisley:
@bhennion Would like to keep pressure sensitivity for complete stroke renders(once per whole stroke) but turn it off for real-time display of ink. Could this be arranged? Will this give 50fps?

When do you see gtk4 port coming out roughly?

src/core/model/StrokeContour.cpp
```diff
@@ -0,0 +1,323 @@
...
+            arcFun(cr, p2.x, p2.y, .5 * p2.z, a, b);
+
+            double bp = inMinusPiPiInterval(a1 - angleIn);
+            double ap = inMinusPiPiInterval(a3 + M_PI_2);
+            auto* arcFunp =
+                    (inMinusPiPiInterval(a3 - a1) >= 0.0 || inMinusPiPiInterval(bp - ap) > 0.0) ? cairo_arc : drawChord;
+            ops.emplace_back(arcFunp, p2.x, p2.y, .5 * p2.z, ap, bp);
+        } else {
+            cairo_line_to(cr, p2.x, p2.y);
+            cairo_arc(cr, p2.x, p2.y, .5 * p2.z, a1, a3 - M_PI_2);
```
rolandlo, line 64: Can you please explain this important case? I see where the path must head to, but why in this precise way?

rolandlo (approved):
Let's go forward with this. Although the precise performance gains (particularly on Windows and MacOS) are hard to quantify, the PR does improve the performance, the code looks good and gives us some flexibility in the future.

bhennion:
> Let's go forward with this. Although the precise performance gains (particularly on Windows and MacOS) are hard to quantify, the PR does improve the performance, the code looks good and gives us some flexibility in the future.

I made a new benchmarking tool: https://github.com/bhennion/xopp-bench (same repo as the old benchmark, but the benchmark is much simpler, easier to use and should really work on MacOS and Windows now).
You simply get some FPS. Interestingly, the results are extremely dependent on the backend:
|  | old | new | old-dashed | new-dashed |
|--------|--------|--------|--------|--------|
| Linux (Wayland) | 10 | 30| 17 | 17 |
| Linux (x11) | 3 | 37 | 5 | 17 |
| MacOS ARM (Quartz) | 11 | 29 | 20 | 20 |
| MacOS Intel (Quartz) |  6  | 20 |   7  | 12 |
| Windows (Win32)       |  11 | 39 | 15 | 25 |

One possible explanation is that the wayland backend does not use the GPU. With the old rendering method, on x11, we seem to be losing a lot of time in communication between CPU and GPU (so it is slower than CPU only).

@rolandlo: could you run it on MacOS/Windows to make sure we are not making things worse with this? It may also be worth benchmarking the ARM version - it could also behave differently.

[Note: there is also a gtk4 version of the benchmark - there is some improvement there but not much (at least until we port to Vulkan)]

**Edit:** Added other measurements

rolandlo:
> I made a new benchmarking tool: https://github.com/bhennion/xopp-bench (same repo as the old benchmark, but the benchmark is much simpler, easier to use and should really work on MacOS and Windows now). You simply get some FPS. 

That works indeed and is much more stable. On MacOS ARM (Quartz backend) I get:

> 	old 	new 	old-dashed 	new-dashed
> MacOS ARM (Quartz backend)	11 	29 	20 	20

 Wait a bit for Windows and MacOS Intel results.

rolandlo:
On Windows (Backend Win32) the improvements are quite significant:

old new old-dashed new-dashed
Windows 11 39 15 25

rolandlo:
On MacOS Intel (Quartz backend) I get:

old new old-dashed new-dashed
MacOS Intel 6 20 7 12

bhennion:
Thanks for all that @rolandlo!
I updated this [post](https://github.com/xournalpp/xournalpp/pull/6385#issuecomment-3135258750) with all the measurements. It means that indeed quite an improvement!

I'll merge it then

rolandlo:
On my Linux machine (using Wayland) I got

old new old-dashed new-dashed
Linux (Wayland) 22 32 31 31

and the conic contour was at amazing 60.

PR #6391, opened 2025-04-16
Add laser/temporary stroke tool for presentations

### LINKED ISSUE #1778: Laser

bhennion:
Fixes #1778 

This PR introduces a new "laser" tool that draws temporary strokes that fade out after a while.

[demo](https://github.com/user-attachments/assets/a53f03f0-fea2-40b2-89a7-c3aa2514ddf7)

You can find the tool in the toolbar customization dialog and add it to your toolbar. When the tool is selected, you can draw as with the pen (no shapes, no filling though), except the strokes are never added to the document and will fade out after a while.

~~For now, the fadeout time is hardcoded (and slow so that we can see it clearly). We may want to make it configurable.
Also, the icon used by the tool is for now the one of the stroke recognizer: a dedicated set of icons is needed.~~

- [x] Add icons
- [x] Make fadeout time configurable?


rolandlo:
I like the implementation of the laser tool except one aspect: The laser stroke is always full opacity until it fades out. When you want to highlight text from the PDF temporarily, that is inconvenient.

bhennion:
> I like the implementation of the laser tool except one aspect: The laser stroke is always full opacity until it fades out. When you want to highlight text from the PDF temporarily, that is inconvenient.

So what do you suggest? Another tool? Or a "mode" that we could use with the highligher as well (a bit like the filling stuff)?
I'm a bit worried about discoverability and making suitable icons here.

rolandlo:
> > I like the implementation of the laser tool except one aspect: The laser stroke is always full opacity until it fades out. When you want to highlight text from the PDF temporarily, that is inconvenient.
> 
> So what do you suggest? Another tool? Or a "mode" that we could use with the highligher as well (a bit like the filling stuff)? I'm a bit worried about discoverability and making suitable icons here.

A "mode" that could be used with the highlighter as well sounds interesting, if that is not too hard to implement. One icon for the laser should be enough.

bhennion:
Ok, we now have two versions of the laser, one pen, one highlighter. We can assign them to stylus buttons, choose colors and sizes independently.

rolandlo:
> Ok, we now have two versions of the laser, one pen, one highlighter. We can assign them to stylus buttons, choose colors and sizes independently.

That's a great solution, I love it! Now I will review the code.

src/core/gui/toolbarMenubar/ComboToolButton.h
```diff
@@ -35,14 +35,15 @@ class ComboToolButton: public AbstractToolItem {
...
+    using Entries = std::vector<Entry>;
     /**
      * @brief When all entries correspond to a single action (but different target values)
      * @param iconName Icon used in the toolbar customization dialog to represent this combo button.
      * @param description Description used in the toolbar customization dialog to explain this combo button.
      * @param entries Entries of the combo menu.
      */
-    ComboToolButton(std::string id, Category cat, std::string iconName, std::string description,
-                    std::vector<Entry> entries, ActionRef gAction);
+    ComboToolButton(std::string id, Category cat, std::string iconName, std::string description, Entries entries,
```
rolandlo, line 45: In the definition of this function it would make sense to also replace `std::vector<Entry>` by `Entries`.

rolandlo (commented):
The code looks good. I would add the Laser tool to some toolbars by default, at very least to the "All Tools" toolbar, but probably also to the other non-minimal/empty toolbars.

bhennion:
I added the laser to all default toolbars, right after the highlighter.

rolandlo (approved):

bhennion:
Thanks for the review! Merging once the CI clears

arynvi:
I couldn't find a keyboard shortcut for this tool: is there one?

Moreover, I can't see it in the **Tools**  menu

Other than this, I have been waiting for this tool for a long time: great job!!! 
Many thanks.

PR #6395, opened 2025-04-17
Fix bad rendering of TexImages in PDF exports

bhennion:
This is a rerun of #3773: fixing PDF exports of TexImages via using `poppler_page_render_for_printing()`, with the added care that screen rendering still uses `poppler_page_render()`


tattsan:
I compared the difference in behavior between this PR and the old one. As you can see from the  [video](https://github.com/xournalpp/xournalpp/issues/3692#issuecomment-1003789787) in #3692,  xournalpp first displays a rough image, then a detailed image. I could see that the old PR rendered this first display at a wastefully high DPI, and the new PR rendered it at a DPI commensurate with the roughness.

tattsan:
Thank you so much!

Issue #5636, opened 2024-04-22
Unable to Annotate PDFs from UNC Paths


neurolag:
### Operating System

Windows

### (Linux Only) Display Server

None

### Installation Method

Manual Build

### Xournal++ Version

efabdd83cd8e80b317fb58590d4b5e410d270fac

### libgtk Version

3.24.41

### Bug Description

When trying to open a `.pdf` file from a UNC path (such as `\\%COMPUTERNAME%\C$\Temp\test.pdf`) results in an error:

```
Error annotate PDF file "\\my-pc\c$\Temp\test.pdf"
Document not loaded! (\\my-pc\c$\Temp\test.pdf), Unknown error
```

When trying to set the PDF background to a UNC path (using the "Missing PDF background file" dialogue), Xournal++, too, reports an error saying:

```
Error reading PDF: Document not loaded! (\\my-pc\C$\Temp\test.pdf), Unknown error
```

Opening `.xopp` files from a UNC path works.

### Expected Behaviour

Much like `.xopp` files, `.pdf` files (and thus PDF backgrounds) should be able to be loaded from UNC paths.

### Steps to Reproduce

1. Download a PDF to the `C:` drive
2. Open the file location in your file explorer
3. In the address bar, replace `C:` with `\\%COMPUTERNAME%\C$`
4. Try to open the `.pdf` file in Xournal++
5. Take note that the described error occurs

Febbe:
That's a known issue. Currently, it is not supported, since we mix many different file handle drivers. Some does not support remote paths.

manuth:
Thanks for the rapid answer!
I was kind of wondering bc I couldn't find a related issue.

Feel free to close it in case it's a duplicate.

Nerixyz:
It seems like this is hitting a bug in libstdc++, specifically [Bug 99333](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99333) where `path.is_absolute()` returns `false` for UNC paths. When compiling with clang and `-stdlib=libc++`, opening and saving works fine.

A hacky workaround could be to check if the root path is `\\` [here](https://github.com/xournalpp/xournalpp/blob/b3cd88bce1d11daf8bfbff4cc8020d33d00ae780/src/util/PathUtil.cpp#L127) (on Windows, with libstdc++).

An MSVC build such as in https://github.com/xournalpp/xournalpp/pull/6214 should solve this as well.

gaboz:
[...]

PR #6409, opened 2025-05-02
Fix conversion of UNC paths to URIs with libstdc++

Nerixyz:
From https://github.com/xournalpp/xournalpp/issues/5636#issuecomment-2845381754 - libstdc++ doesn't handle UNC paths on Windows correctly. A path like `\\wsl.local\Foo\bar` isn't considered absolute. This PR fixes this issue by introducing a wrapper that checks for UNC paths.

With this PR, I'm able to open PDFs from WSL, which uses UNC paths. There are probably other issues with UNC paths, so I wouldn't say the issue is fixed.

Note: https://github.com/xournalpp/xournalpp/pull/6137 will remove the check for absolute paths and uses `fs::absolute` unconditionally there. This would break UNC paths with libstdc++ (as that's what currently happens).


Nerixyz:
> Out of curiosity, what does fs::absolute() return on a UNC path?

It appends the drive letter of the current working directory:

- input: `\\wsl.localhost\Ubuntu\[...]\doc.pdf`
- output: `F:\wsl.localhost\Ubuntu\[...]\doc.pdf`

> Could you try on your machine that the pipeline generated version does work as well?

That version works for me. When exporting to PDF, it still shows the warning dialog that the background overwrite check failed, but one can click "Replace" so this isn't a huge issue.

PR #6424, opened 2025-05-11
Disable crowdin action and release action in fork

tattsan:
Make Crowdin workflows only run in the main repo and are skipped entirely in forks.

Edit: Also Release Action.


tattsan:
Looks like the previous fix wasn’t enough, so I’ll add some more handling to fix it properly.

tattsan:
I’m still getting “Run failed” emails on my fork, so I guess this approach doesn’t really help.
Looks like this is something fork owners should handle themselves.
Sorry for the noise, and thanks for your understanding!

bhennion:
I merged a commit doing this on the release branch. It'll get to the master branch on its own.

tattsan:
Thanks!

PR #6427, opened 2025-05-12
Split LAYER_NEW into LAYER_NEW_BELOW_CURRENT and LAYER_NEW_ABOVE_CURRENT

### LINKED ISSUE #6403: Inserting a layer between background and first layer is no longer possible on master
Assume we have a page with background and a layer called MyLayer. It is no longer possible to add a layer between them to get a layer stack Background \ Layer 1 \ MyLayer.

This only concerns the master branch. In release 1.2.6, this could be achieved by selecting the background as the active layer and then clicking on `Add Layer` (see #2261). Since on master, the background can no longer be selected as the active layer, this possibility is no longer available.

Now, selecting the background as active layer doesn't really make sense, so re-enabling that possibility does not feel like the way to go.
The solution is probably to split `Add Layer` into `Insert layer above the Active Layer` and `Insert a layer beneath the Active Layer`. That will make it more explicit and allow to close #2261 as well.

bhennion:
Fix #6403


src/core/plugin/ActionBackwardCompatibilityLayer.cpp
```diff
@@ -65,7 +65,7 @@ static const std::map<std::string_view, Action> STATELESS_ACTION_MAP = {
         {"ACTION_PLUGIN_MANAGER"sv, Action::PLUGIN_MANAGER},
         {"ACTION_HELP"sv, Action::HELP},
         {"ACTION_ABOUT"sv, Action::ABOUT},
-        {"ACTION_NEW_LAYER"sv, Action::LAYER_NEW},
+        {"ACTION_NEW_LAYER"sv, Action::LAYER_NEW_ABOVE_CURRENT},
```
rolandlo, line 68: Maybe also add 
```c++
{"ACTION_NEW_LAYER_BELOW_CURRENT"sv, Action::LAYER_NEW_BELOW_CURRENT},
```
so that this can be achieved via plugins.

rolandlo (approved):
It works fine and the code looks fine as well. I have just one suggestion about the plugin API (see in the code).

PR #6451, opened 2025-05-24
Fix typing when GeometryTool is on

### LINKED ISSUE #5689: TextEditor typing and Selection/SplineHandler accelerators don't work when a geometric tool is shown
In the stable version when you use the text tool while a geometry tool is active, you can type normally. Now when you type keys like r and s are not absorbed by the text tool, but by the geometry tool.

_Originally posted by @rolandlo in https://github.com/xournalpp/xournalpp/pull/5688#pullrequestreview-2056800147_

See #5688 for a short discussion on finding a proper fix.

bhennion:
* Remove data duplication between GeometryTool / GeometryToolinputHandler
   * Feed keyboard events after other tools but before navigation

Fix #5689


rolandlo (approved):
LGTM. Nice simplification. I had noticed the typing issue when the geometry tool is on as well recently. Great that you fixed it.

PR #6476, opened 2025-06-11
Support editing LaTeX in an external editor

### LINKED ISSUE #6466: Option to edit LaTeX code in external editor

I am a vim user and don't enjoy writing LaTeX code in the builtin gtksourceview editor that opens with the LaTeX tool.

### Feature Description

I propose that we add a setting under the LaTeX section that allows the user to make xournalpp invoke some external editor command in order to edit LaTeX code instead of the standard editor window. xournalpp will pass a temporary file to the editor which the user can edit and read the content of the file once the editor exits. Additionally, we should also make the file extension of this temporary file configurable in case the user uses a LaTeX alternative like Typst.

### Alternative Solutions

Manually pasting the code into the LaTeX editor

### Additional Context

_No response_

LordMZTE:
This adds support for editing LaTeX in a user-configurable external editor instead of the built-in one (disabled by default).

This also includes a small refactor where LatexController is kept as  `shared_ptr` instead of a `unique_ptr` because without the dialog (necessarily) around, we would otherwise run into a scenario where the controller is destroyed before we run some callbacks.

Closes #6466


bhennion:
Thanks for the PR!
First, the code looks very clean. That's great! I'll give it a thorough read once the following remarks have been addressed:

1. The CI is failing on Windows
2. I tried setting the external editor to either "gedit -w" or "kwrite". It both cases, if I edit the file and save, I get the Invalid Tex message. The problem might be that most editors add a trailling \n at the file. I am not sure though.
3. Item 2. raises another problem: we no longer get access to compilation errors, nor can we preview the Tex element before adding it.
4. You can get a segfault by doing:
    1. Open a new document. Add a page.
    2. On the second page, open the Tex editor. Leave it open.
    3. Go to Xournal++'s main window, delete the second page and write something on the first page (to clear the undo stack)
    4. Close the external editor window.
    I'm guessing theLayerController tries to add the Tex element to a destroyed page, leading to a SegFault.

I think both 3. and 4. would be neatly solved by keeping a LatexDialog that would be modal (like the default one) and hence forbid any interaction with the main window.
It would open the external editor, handle the compilation and preview, and show the compilation output (so basically, the current LatexDialog without the `Tex Source` tab and with an open editor button in case the user is unhappy with the preview).
If you don't want to click the extra "ok" to add the element, you could have a "Close dialog on compilation success" option.

LordMZTE:
Thanks for the feedback! I've addressed your requests. I've split the `LatexDialog` class up into three classes:
- `AbstractLatexDialog` which implements the common funcitonality shared by the other two classes
- `LatexDialog` was renamed to `IntEdLatexDialog` which still does the same thing but now inherits from `AbstractLatexDialog`
- `ExtEdLatexDialog` is a lightweight version of the normal modal without the editor and an additional "Continue Editing" button as requested. It also manages the external editor. (This also shows error output now)

I've also written some code that removes a trailing newline (which I have confirmed kwrite does add), if any. I haven't been able to get LaTeX running myself as I use Typst instead which doesn't complain about a trailing newline, it would be appreciated if you could give it a test on your LaTeX setup.

An option to automatically confirm like you suggested has also been added.

src/core/control/LatexController.cpp
```diff
@@ -22,25 +22,26 @@
...
-#include "util/Color.h"                      // for Color, get_color_contrast
-#include "util/PathUtil.h"                   // for ensureFolderExists, getT...
-#include "util/PlaceholderString.h"          // for PlaceholderString
-#include "util/PopupWindowWrapper.h"         // for PopupWindowWrapper
-#include "util/Rectangle.h"                  // for Rectangle
-#include "util/Util.h"                       // for npos
-#include "util/XojMsgBox.h"                  // for XojMsgBox
-#include "util/i18n.h"                       // for FS, _, _F, N_
-#include "util/safe_casts.h"                 // for round_cast
+#include "gui/dialog/ExtEdLatexDialog.h"
```
```suggestion
#include "gui/dialog/ExtEdLatexDialog.h"  // for ExtEdLatexDialog
```
(to avoid the unnecessary changes due to clang formatting)

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+        gtk_window_close(getWindow());
+        return;
+    }
+
+    AbstractLatexDialog::setCompilationStatus(isTexValid, isCompilationDone, compilationOutput);
+};
+
+void ExtEdLatexDialog::openEditor() {
+    if (editorCancellable)
+        return;
```
bhennion: Code convention
```suggestion
    if (editorCancellable) {
        return;
    }
```

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    xoj::util::GErrorGuard err{};
+    xoj::util::GStrvGuard argv{};
+
+    std::string editorCmd = texCtrl->settings.externalEditorCmd;
+    if (editorCmd.empty()) {
+        auto maybeEditor = g_getenv("EDITOR");
+        if (!maybeEditor) {
+            XojMsgBox::showErrorToUser(
+                    texCtrl->control->getGtkWindow(),
```
bhennion: Should it be the Dialog's window here?

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    auto content = self->getBufferContents();
+
+    LatexController::triggerImageUpdate(self->texCtrl, content);
+}
+
+std::string ExtEdLatexDialog::getBufferContents() {
+    auto content = Util::readString(tempf);
+    if (!content)
+        return std::string();
```
bhennion:
```suggestion
    if (!content) {
        return std::string();
    }
```

src/core/control/LatexController.h
```diff
@@ -82,26 +83,27 @@ class LatexController final {
      * the given LaTeX string. If the preview is already being updated, then
      * this method will be a no-op.
      */
-    void triggerImageUpdate(const std::string& texString);
+    static void triggerImageUpdate(std::shared_ptr<LatexController> self, const std::string& texString);
```
bhennion: Is the shared_ptr business still required?

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+        XojMsgBox::showErrorToUser(texCtrl->control->getGtkWindow(),
+                                   FS(_F("Failed to parse external editor command: {1}") % err->message));
+        return;
+    }
+
+    xoj::util::GObjectSPtr<GSubprocess> process(
+            g_subprocess_newv(argv.get(), G_SUBPROCESS_FLAGS_NONE, xoj::util::out_ptr(err)), xoj::util::adopt);
+
+    if (err) {
+        XojMsgBox::showErrorToUser(texCtrl->control->getGtkWindow(),
```
bhennion: Should it be the Dialog's window here?

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+                     this);
+    g_signal_connect(btContEdit, "clicked", G_CALLBACK(+[](GtkButton*, gpointer d) {
+                         auto* self = static_cast<ExtEdLatexDialog*>(d);
+                         self->openEditor();
+                     }),
+                     this);
+}
+
+ExtEdLatexDialog::~ExtEdLatexDialog() {
+    fs::remove(tempf);
```
bhennion: Exceptions/IO errors should probably be caught here.

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+            XojMsgBox::showErrorToUser(
+                    texCtrl->control->getGtkWindow(),
+                    "No external editor configured and no $EDITOR environment variable set. Please configure an "
+                    "external editor command or disable the external editor functionality to use a built-in one.");
+            return;
+        }
+        editorCmd = maybeEditor;
+    }
+    editorCmd += " '";
+    editorCmd += tempf.string();
```
bhennion: I think it should be
```suggestion
    editorCmd += tempf.u8string();
```

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+    editorCmd += "'";
+
+    if (!g_shell_parse_argv(editorCmd.c_str(), nullptr, xoj::util::out_ptr(argv), xoj::util::out_ptr(err))) {
+        XojMsgBox::showErrorToUser(texCtrl->control->getGtkWindow(),
```
bhennion: Should it be the Dialog's window here?

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,183 @@
...
+                                   FS(_F("Could not wait for editor: {1}") % err->message));
+        return;
+    }
+
+    g_clear_object(&self->editorCancellable);
+    gtk_widget_set_sensitive(GTK_WIDGET(self->btContEdit), true);
+
+    auto status = g_subprocess_get_status(process);
+    if (status != 0) {
+        XojMsgBox::showErrorToUser(self->texCtrl->control->getGtkWindow(),
```
bhennion: Should it be the Dialog's window here?

LordMZTE:
Thanks! I unfortunately don't have a windows machine to test on.

rolandlo:
I have tested it on Windows using [gedit](https://packages.msys2.org/packages/mingw-w64-x86_64-gedit). There's some issue with line ending, which makes pdflatex unable to compile the `tex.tex` file.
See here for the tex-folder [tex.zip](https://github.com/user-attachments/files/20745324/tex.zip)

Relevant line in the log:
````txt
! Text line contains an invalid character.
l.30 ^^@
````

**Comment by LordMZTE** [comment]
That's interesting, it seems like `tex.tex` contains a null byte after your input. I'm unsure where that came from, but I suspect something must be up with the trailing CRLF removing logic in `ExtEdLatexDialog::getBufferContents`. I'll investigate.

**Comment by LordMZTE** [comment]
Looks like the code used to incorrectly look for `\n\r` instead of `\r\n`, which caused your `edit.tex` to contain `z^2\r`, which is definitely invalid, but I still have no clue how the null managed to get there. Could you please test again on the latest revision?

**Comment by rolandlo** [comment]
> Looks like the code used to incorrectly look for `\n\r` instead of `\r\n`, which caused your `edit.tex` to contain `z^2\r`, which is definitely invalid, but I still have no clue how the null managed to get there. Could you please test again on the latest revision?

We open the file in (text) reading mode. From https://en.cppreference.com/w/cpp/io/c/FILE.html#Binary_and_text_modes
> (in particular, C streams on Windows OS convert '\n' to '\r\n' on output, and convert '\r\n' to '\n' on input).
We should account for that  in some way.

**Comment by rolandlo** [comment]
> > Looks like the code used to incorrectly look for `\n\r` instead of `\r\n`, which caused your `edit.tex` to contain `z^2\r`, which is definitely invalid, but I still have no clue how the null managed to get there. Could you please test again on the latest revision?
> 
> We open the file in (text) reading mode. From https://en.cppreference.com/w/cpp/io/c/FILE.html#Binary_and_text_modes
> 
> > (in particular, C streams on Windows OS convert '\n' to '\r\n' on output, and convert '\r\n' to '\n' on input).
> > We should account for that  in some way.

It all works after applying this patch:
```patch
--- a/src/util/PathUtil.cpp
+++ b/src/util/PathUtil.cpp
@@ -63,11 +63,10 @@ auto Util::getLongPath(const fs::path& path) -> fs::path { return path; }
 auto Util::readString(fs::path const& path, bool showErrorToUser, std::ios_base::openmode openmode)
         -> std::optional<std::string> {
     try {
-        std::string s;
+        std::stringstream buffer;
         std::ifstream ifs{path, openmode};
-        s.resize(fs::file_size(path));
-        ifs.read(s.data(), as_signed(s.size()));
-        return {std::move(s)};
+        buffer << ifs.rdbuf();
+        return buffer.str();
     } catch (const fs::filesystem_error& e) {
         if (showErrorToUser) {
             XojMsgBox::showErrorToUser(nullptr, e.what());
```

LordMZTE:
Thank you! I've applied your patch!

src/core/gui/dialog/ExtEdLatexDialog.cpp
```diff
@@ -0,0 +1,187 @@
...
+    auto content = Util::readString(tempf);
+    if (!content) {
+        return std::string();
+    }
+
+    // Remove a possible trailing newline
+    if (content->length() >= 2 && content->compare(content->length() - 2, 2, "\r\n") == 0) {
+        content->pop_back();
+        content->pop_back();
+    } else if (!content->empty() && content->back() == '\n') {
```
rolandlo: Since `readString` normalizes line endings (replaces `\r\n` on Windows by `\n`) this can be simplified.
```suggestion
    if (!content->empty() && content->back() == '\n') {
```

src/util/PathUtil.cpp
```diff
@@ -63,11 +63,10 @@ auto Util::getLongPath(const fs::path& path) -> fs::path { return path; }
 auto Util::readString(fs::path const& path, bool showErrorToUser, std::ios_base::openmode openmode)
         -> std::optional<std::string> {
     try {
-        std::string s;
+        std::stringstream buffer;
```
rolandlo, line 67: On MacOS ARM it only compiles when including the sstream header:
```
#include <sstream>
```

rolandlo:
I tested on MacOS ARM. 
- With editor command `nano` nothing showed up (and there was no error message either).
- With editor command `open /Applications/TeX/TeXShop.app` the TeXShop app showed up and I could modify the formula. However after saving and quitting, the preview wasn't updated. Only after I hit `Continue editing` the preview updated. Not sure how this should be handled.

bhennion:
> * With editor command `open /Applications/TeX/TeXShop.app` the TeXShop app showed up and I could modify the formula. However after saving and quitting, the preview wasn't updated. Only after I hit `Continue editing` the preview updated. Not sure how this should be handled.

I'm guessing that TeXShop returns straight away (and its main process has been forked). So, on our end, the GSubprocess returns immediately, so the preview gets updated right after the editor is opened instead of after its termination.

I don't see a way to handle this kind of cases (other than running texshop with some kind of no-fork flag, if any is implemented).

rolandlo:
> > * With editor command `open /Applications/TeX/TeXShop.app` the TeXShop app showed up and I could modify the formula. However after saving and quitting, the preview wasn't updated. Only after I hit `Continue editing` the preview updated. Not sure how this should be handled.
> 
> I'm guessing that TeXShop returns straight away (and its main process has been forked). So, on our end, the GSubprocess returns immediately, so the preview gets updated right after the editor is opened instead of after its termination.
> 
> I don't see a way to handle this kind of cases (other than running texshop with some kind of no-fork flag, if any is implemented).

Could it be the `open` command that returns straight away after launching TeXShop? Maybe I need to use the `-W` option to wait for the application to quit. See https://www.jviotti.com/2022/11/28/launching-macos-applications-from-the-command-line.html. I will test once I return home.

rolandlo:
> > > * With editor command `open /Applications/TeX/TeXShop.app` the TeXShop app showed up and I could modify the formula. However after saving and quitting, the preview wasn't updated. Only after I hit `Continue editing` the preview updated. Not sure how this should be handled.
> > 
> > 
> > I'm guessing that TeXShop returns straight away (and its main process has been forked). So, on our end, the GSubprocess returns immediately, so the preview gets updated right after the editor is opened instead of after its termination.
> > I don't see a way to handle this kind of cases (other than running texshop with some kind of no-fork flag, if any is implemented).
> 
> Could it be the `open` command that returns straight away after launching TeXShop? Maybe I need to use the `-W` option to wait for the application to quit. See https://www.jviotti.com/2022/11/28/launching-macos-applications-from-the-command-line.html. I will test once I return home.

Unfortunately using the `-W` option with the `open` command didn't solve the problem. It waits indefinitely after closing the editor, even if I run it from the terminal. That's not how it's supposed to work, I think, but it does so on my system.

LordMZTE:
> Unfortunately using the `-W` option with the `open` command didn't solve the problem. It waits indefinitely after closing the editor, even if I run it from the terminal. That's not how it's supposed to work, I think, but it does so on my system.

Perhaps TeXShop forks off *additionally* to `open`? Could you check if it has a command line option regarding this?

(note that I have absolutely no experience with MacOS)

rolandlo:
> > Unfortunately using the `-W` option with the `open` command didn't solve the problem. It waits indefinitely after closing the editor, even if I run it from the terminal. That's not how it's supposed to work, I think, but it does so on my system.
> 
> Perhaps TeXShop forks off _additionally_ to `open`? Could you check if it has a command line option regarding this?
> 
> (note that I have absolutely no experience with MacOS)

There are no command line options for TeXShop on MacOS. See 

https://github.com/xournalpp/xournalpp/pull/6476#issuecomment-2985141360

It also cannot be started via `/path/to/bundle.app/Contents/MacOS/name-of-binary` like other app-bundles (e.g. Xournal++.app) for some reason. Same applies to the default text editor (TextEdit.app).

rolandlo:
I have also tried Texifier. This one can also be run directly through the binary: `/Applications/Texifier.app/Contents/MacOS/Texifier`. Neither that nor opening the app bundle via `open` or `open -W` worked. The preview didn't update after I closed the Texifier app. I haven't found any command line arguments. Seems like that's more of a Linux/open source cross platform thing than something integrated into common MacOS GUI apps.

bhennion:
One possible solution could be to switch from "I wait until the app terminates" to "I wait until the file is saved" using a [GFileMonitor](https://docs.gtk.org/gio/class.FileMonitor.html).

LordMZTE:
> One possible solution could be to switch from "I wait until the app terminates" to "I wait until the file is saved" using a [GFileMonitor](https://docs.gtk.org/gio/class.FileMonitor.html).

This poses quite a few questions:
- Some editors save files by deleting the old file, then creating a new one (neovim). If we use `g_file_monitor_file` that may cause us to miss events, if we use `g_file_monitor_directory` that function isn't supported on all platforms:
  > This may fail if directory monitoring is not supported.

  (that remark isn't present for `g_file_monitor_file`)
- How should this interact with the auto-confirm option? We now have no way of knowing when the user is actually finished editing, as there might just be intermediate saves (or even autosaves).
- How should this interact with the "Continue Editing" button? We now don't know when to disable it and when it should be available.
- This may cause some strange behavior if the user:
  1. Adds a LaTeX element, pressing Ok on the dialog while the editor is still open (which would normally kill the editor but we now can't do that since it's forked off)
  2. Adds another LaTeX element. Now we have two editors open editing the same file, which may cause confusion.
- What if the user closes the dialog with the editor open like above, but then keeps editing the file while expecting the LaTeX element to keep updating? If we want this to work, we'd have to do some major refactoring.
  - If we did want to implement this, how do we know when to stop watching the file (i.e. when the user is actually done)?

bhennion:
> This poses quite a few questions:

Indeed it does. I'm not sure this would be the best solution either...

LordMZTE:
How should we proceed here? Should I attempt to detect file changes like we discussed above, or should we keep the current mechanism of waiting for the editor to exit? Are there any entirely different suggestions?

bhennion:
> Neither that nor opening the app bundle via `open` or `open -W` worked. The preview didn't update after I closed the Texifier app

So what happens when running `/Applications/Texifier.app/Contents/MacOS/Texifier` from a terminal? Does the terminal return control immediatly? Same question when using `open` or `open -W`.


> How should we proceed here? Should I attempt to detect file changes like we discussed above, or should we keep the current mechanism of waiting for the editor to exit? Are there any entirely different suggestions?

I don't think it is worth it. I see several (not exclusive) possibilities

1. Disable the feature on MacOS if we can't figure out how to make it work.
2. Add a list of preconfigured external editors (platform dependent, of course - and the list would disable the configurations when the app is not present) once someone manages to get something working on MacOS as well
3. Add a "(Expert user only)" label to the setting, to limit the number of bug reports we'll get.

Personally, I would go for 3. and maybe 1. as well. As I already expressed earlier, I think this is a very specific feature that 99% of users will never use, so we shouldn't put too much efforts into it.

tattsan:
Thank you for this interesting PR! I tried it out briefly and would like to share some thoughts as a user.

When using the LaTeX tool in Xournal++, I think the experience differs depending on whether you're writing short LaTeX snippets or long code, such as complex TikZ diagrams. For short code, the internal editor is faster and more efficient than launching an external one.

For longer code, I prefer using an external editor with preview and log features—like Emacs with AUC-TeX in my case. I’ve customized default_template.tex to load edit.tex via \input{edit.tex}, and I compile it directly from Emacs. In such cases, Xournal++’s internal preview can get in the way, so I actually preferred the earlier version of this PR.

Previously, I used Ctrl+Alt+X to open Emacs, edit the LaTeX code externally, and then paste the result into Xournal++. This PR makes that workflow much smoother, which I really appreciate. That said, I wouldn't want to launch Emacs for quick edits. Ideally, I’d like to switch between the internal and external editor depending on the situation—perhaps calling the external one via a plugin. I think this kind of advanced feature would appeal to plugin authors and power users.

bhennion:
> In such cases, Xournal++’s internal preview can get in the way

How so? Are you quitting the editor every time you compile?

> Ideally, I’d like to switch between the internal and external editor depending on the situation

There could a single dialog, with both the internal editor and an "Open external editor" button - and possibly a check box "always open the external editor". The internal editor would be grayed out if the user opened an external editor.

tattsan:
>    How so? Are you quitting the editor every time you compile?

Not exactly. In my environment, the preview window and the editor in Xournal++ overlap, so I have to move one of them every time. Also, I have another previewer that is launched from Emacs, and it automatically updates after each compilation. So I prefer to use that preview instead. There's no point in keeping two previews side by side, especially when one is outdated.
That said, you're right that the output from an external compilation might not exactly match what Xournal++ would generate, so completely disabling the Xournal++ preview window could be inconvenient. I suppose this is something I should resolve on my end, given my custom setup.

>    There could be a single dialog, with both the internal editor and an "Open external editor" button...

I like that idea! Having a framework that allows launching an external editor via a plugin would also be nice, but I don't think this PR needs to implement that directly.

LordMZTE:
I've added an `(Expert user only)` label as requested. Should I implement any of the other proposals, or are we still uncertain about those?

tattsan:
Thanks again for working on this — it's a really cool feature!

About my earlier idea of switching between internal and external editors depending on how long the LaTeX code is: I realize that’s probably beyond the scope of this PR, and I definitely don’t want to slow things down. It’s more of a niche use case anyway, so I think it’s totally fine to leave it for a future PR if someone feels like picking it up.

Regarding the macOS issue: I’ve seen people mention that some editors on macOS launch a subprocess and exit right away, which makes it tricky for Xournal++ to tell when the editor is actually done. I can’t test this myself since I don’t have a Mac, but I feel like just adding a small note in the settings or docs — something like “If the editor doesn’t behave as expected, try using a wrapper script” — would be enough. Disabling the feature on macOS seems a bit too much, especially since this is aimed at expert users anyway.

In any case, I’m really happy to see this coming together — looking forward to the final result!

PR #6501, opened 2025-06-27
Fix plugin function addImages when maxWidth/maxHeight are present

### LINKED ISSUE #6500: Api app.addImages does not heed maxHeight and maxWidth
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

KDE Plasma

X.Org

### Installation Method

AUR xournalpp-git

### Xournal++ Version

1.2.7+dev

### libgtk Version

3.24.49

### Bug Description

When running a function containing
 `local refs = app.addImages{images={{path = "image.png", x=100, y=200, maxHeight=50, maxWidth=50}}}
`
for some sample image as a app.registerUi call, the image is inserted extending fully to the right side of the current page while it should only be a maximum of 50 pixels wide. 

### Expected Behaviour

Width and height should be limited.

### Steps to Reproduce

Create any image (in my case I took a partial screenshot) and run the given line in a plugin.

### Additional Context

Scale also does not work I think. 

Is it right that there is `addImagesFromFilepath` instead of `addImages` in [here](https://github.com/xournalpp/xournalpp/blob/release-1.2/src/core/plugin/luapi_application.h#L2217)?

bhennion:
Fix #6500

### DIFF
--- a/src/core/control/tools/ImageHandler.cpp
+++ b/src/core/control/tools/ImageHandler.cpp
@@ -103,11 +103,10 @@ bool ImageHandler::addImageToDocument(std::unique_ptr<Image> img, PageRef page, 
     return true;
 }
 
-void ImageHandler::automaticScaling(Image& img, PageRef page) {
+void ImageHandler::automaticScaling(Image& img, PageRef page, int width, int height) {
     double zoom = 1;
     double x = img.getX();
     double y = img.getY();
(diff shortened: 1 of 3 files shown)


bhennion:
The regression only affects the master branch. It was introduced in bb8b7af339

Merging once the CI clears.

Issue #2404, opened 2020-11-15
Plugins with non ASCII characters can't be load, even if they are part of ANSI


Febbe:
> Special folder names do not seem to work on Windows. I renamed the ColorCycle plugin to `àÄ@fix~d$$äö!è`, after which it was still correctly displayed in the Plugin Manager, but could not be loaded upon activating it and restarting Xournal++, see
> ![grafik](https://user-images.githubusercontent.com/40485433/99077216-ad822900-25bc-11eb-9909-1d4897cc7fe4.png)
> 
> **Edit** even with a simple folder name like `René` or `Ähnlichkeit` the plugin cannot be loaded under Windows.

It seems like, this is a dependency issue caused by mingw-64. If the path exist, fopen should be able to open path::string::c_str. Also the bug was there long before (glib paths are incompatible to c-paths on windows).

_Originally posted by @Febbe in https://github.com/xournalpp/xournalpp/pull/2391#issuecomment-727590656_

Febbe:
```main.cpp
#include <fstream>
#include <filesystem>
#include <clocale>
#include <cassert>
#include <memory>

int main() {
    fs::path ord1{u8"Ä"};
    fs::path ord2{u8"☺"};
    fs::path p1{ord1/"test.txt"};
    fs::path p2{ord2/"test.txt"};
    fs::create_directory(ord1);
    fs::create_directory(ord2);
    create_file(p1);
    create_file(p2);
    assert(fs::exists(p1));
    assert(fs::exists(p2));
    assert(fopen(p1.string().c_str(), "r"));
    assert(fopen(p2.string().c_str(), "r"));
}
```

```CMakeLists.txt
cmake_minimum_required(VERSION 3.17)
project(test_fs)

set(CMAKE_CXX_STANDARD 20)
if(MSVC)
    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
else()
    set(CMAKE_EXE_LINKER_FLAGS "-static-libgcc -static-libstdc++ -static")
endif()

set(CMAKE_)
add_executable(test_fs main.cpp)
```

Febbe:
A solution could be to load the file via Glib or `std::ifstream` into a buffer, and to pass the buffer of the file to lua.

Technius:
Did you try `fs::u8path`?

Febbe:
[...]

PR #6521, opened 2025-07-07
Issue #2404 resolved

Hicortab:
I've been looking for a long time to find the reason why it doesn't interpret the path correctly but I think it's a problem with the library you use to load the lua files that isn't ready to handle something different from ASCII and I decided to reimplement it, I hope it can be useful. It's a great project

### DIFF
--- a/src/core/plugin/Plugin.cpp
+++ b/src/core/plugin/Plugin.cpp
@@ -219,6 +219,63 @@ void Plugin::addPluginToLuaPath() {
     lua_pop(lua.get(), 1);
 }
 
+
+int lua_loadfile_via_gtk4(lua_State *L, const fs::path& pluginPath) {
+    std::string filename_utf8_glib = pluginPath.u8string();
+    const gchar* filename_gchar = filename_utf8_glib.c_str();
+
+    GFile *gfile = g_file_new_for_path(filename_gchar);
+    if (!gfile) {
+        g_warning("Unable to create GFile for path: \"%s\"", filename_utf8_glib);
+        return LUA_ERRFILE;
+    }
+
+    GError *error = NULL;
+    GFileInputStream *input_stream = g_file_read(gfile, NULL, &error);
+
+    if (!input_stream) {
+        g_warning("Unable to create GFile for path: \"%s\" ( $s )", filename_utf8_glib, error->message);
+        g_error_free(error);
+        g_object_unref(gfile);
+        return LUA_ERRFILE;
+    }
+
+    gsize bytes_read = 0;
+    std::vector<char> buffer;
+    gboolean success = TRUE;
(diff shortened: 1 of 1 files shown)


rolandlo:
Fixes #2404

bhennion:
Could you please also add a test unit that checks it works? The unit must fail without this patch, of course.

I fear we'll hit other encoding issues on other platforms, if the locale is not utf-8. From https://docs.gtk.org/gio/type_func.File.new_for_path.html
> The value is a platform-native string, using the preferred OS encoding on Unix and UTF-8 on Windows.

Hicortab:
The "unit test" (which I never wrote but tried manually) actually envisaged using the characters from the flag and then additional UTF-8 ones like the following:

àÄ@fix~d$$äö!è
ЖжШшДдΩωΣσπµ√™©®€£$¥¢°±×÷≠≈≡≤≥¿¡‱′″‴‟†‡•…‰′″‴‹›«»“”‘’
something in Japanese/Korean/Chinese/Hebrew etc...

If I have to create a "real" one (or at least stop doing it manually), please give me clearer guidelines (e.g. make sure it works with characters of a certain alphabet type or with symbols of a certain type that might be common in certain environments), so far it seems to work with all the "weird" characters.

For the fear that we might still have encoding problems I'd like to touch on this for a moment.

GLib has a preferred input encoding for path strings (often UTF-8) and std::filesystem::path::u8string() guarantees that the filename_utf8_glib string is always UTF-8, regardless of the operating system, so I don't see why you would care about the conversion part

And then could this problem really affect ordinary people who don't even know what an encoding is? Every operating system defaults to UTF-8 (except extremely old operating systems but at this point Xournal++ wouldn't even run due to the relatively old hardware).

bhennion:
My point is precisely that g_file_new_for_path() does not necessarily accept utf-8 strings, if the locale is encoded otherwise.
It could look like "this never happens anymore", but things like https://en.wikipedia.org/wiki/GB_18030 exist: this encoding is REQUIRED to use for software resellers in China.
So some of our users will have a non-utf8 locale.

We already run the CI tests in a variety of locales (on ubuntu), so if you add a test unit, we'll know for sure if there's a problem.
About the test unit itself, there is no need to test a hundred utf-8 characters: a handful should be enough to detect potential issues. See e.g. test/unit_tests/model/DocumentNameTest.cpp

bhennion:
I made a test unit you can find [here](https://github.com/bhennion/xournalpp/pull/34)
On the master branch, it passes on Linux and fails on Windows (it probably passes on MacOS but the CI is not done at the time of writing).

You can pick it up (either cherry-pick the commit or just apply the patch) here so we'll see if it passes with your patch.

Issue #6536, opened 2025-07-10
invalid path to background-pdf


friedFat:
### Operating System

Linux

### (Linux only) Distribution

VoidLinux

### (Linux only) Desktop Environment

none (dwm window manager)

### (Linux Only) Display Server

X.Org

### Installation Method

system package manager(xbps)

### Xournal++ Version

1.2.6

### libgtk Version

3.24.49

### Bug Description

When creating a .xopp file based on a .pdf background specified by a relative path(in the cli), it saves an invalid filename(see additional context), causing the background file path to be parsed wrongly.

### Expected Behaviour

The pdf background should be read correctly, so that the path does not contain the parent directory twice.
```sh
$ pwd
/home/k/Documents/scnotes
$ ls physics/ -A
.kernphysik.pdf  .rasierklingenversuch.pdf  kernphysik.xopp  rasierklingenversuch.xopp
$ xournalpp --create-img=a.png physics/rasierklingenversuch.xopp
The background file "physics/physics/.rasierklingenversuch.pdf" could not be found. It might have been moved, renamed or deleted.
```

### Steps to Reproduce

1. mkdir foo
2. get some pdf, save it to foo/bar.pdf
3. xournalpp foo/bar.pdf, save it to foo/bar.xopp
4. xournalpp --create-img=img.png foo/bar.xopp

### Additional Context

in src/core/control/xojfile/LoadHandler.cpp:452, I tried printing the `filename` attribute value:
```cpp
        const char* domain = LoadHandlerHelper::getAttrib("domain", false, this);
        {
            const char* sFilename = LoadHandlerHelper::getAttrib("filename", false, this);
            if (sFilename == nullptr) {
                error("PDF Filename missing!");
                return;
            }
            pdfFilename = fs::u8path(sFilename);
            error("%s", FC(_F("sFilename: {1}") % sFilename))  <--- line 452
        }
```
=> sFilename: physics/.rasierklingenversuch.pdf

filename is the full relative path the .xopp was created from(xournalpp physics/.rasierklingenversuch.pdf).

```cp

        if (!strcmp("absolute", domain))  // Absolute OR relative path
        {
            if (pdfFilename.is_relative()) {
                pdfFilename = xournalFilepath.remove_filename() / pdfFilename;
            }
        } else if (!strcmp("attach", domain)) {
```
In the code here(which directly follows the code where i added the debug message) we can see how we get the parent of the .xopp-file and append the pdfFilename, thus adding the directory name twice.

[...]

PR #6540, opened 2025-07-13
Save asset names as relative paths

### LINKED ISSUE #2736: PDF background not saved in .xopp / not loaded from absolute path

bhennion:
This PR is a rebase of #4262, with the added feature that assets in preexisting files are not made relative (see https://github.com/xournalpp/xournalpp/pull/4262#issuecomment-1509358321).

Fix #2736 (and dups)

### DIFF
--- a/test/unit_tests/model/ColorPaletteTest.cpp
+++ b/test/unit_tests/model/ColorPaletteTest.cpp
@@ -19,6 +19,7 @@
 
 #include "gui/toolbarMenubar/model/ColorPalette.h"
 #include "util/PathUtil.h"
+#include "util/i18n.h"
 
 
 std::string helper_read_file(fs::path path) {
@@ -101,7 +102,7 @@ TEST(ColorPalette, testAttributesLoad) {
 
     Palette palette2 = Palette{GET_TESTFILE("palettes/not_broken_attribute.gpl")};
     palette2.load();
-    EXPECT_EQ(std::string{"Gray"}, palette2.getColorAt(0).getName());
+    EXPECT_EQ(std::string(fetch_translation_context("Color", "Gray")), palette2.getColorAt(0).getName());
 }
 
 TEST(ColorPalette, testWrongColor) {


tattsan:
Thank you for the great work! I've tested the PR on my typical work (on Linux). After saving xopp file: 
* `<background type="pdf" domain="absolute" filename="./foo.pdf" pageno="1"/>` is converted to its absolute path.
* `<background type="pdf" domain="absolute" filename="../foo.pdf" pageno="1"/>` is not converted.

I feel this is inconsistent. It would be preferable if paths are not converted here.

bhennion:
Indeed, there was a mistake :) Should be better now.

tattsan:
Thanks! But there still remains this kind of issue. Due to this long-running bug like #6536, I have many files like:
```
<background type="pdf" domain="absolute" filename="././././././foo.pdf" pageno="1"/>
```
This is because I have often opened the file as ./foo.xopp from my terminal. These files are still converted to absolute path.

bhennion:
> Thanks! But there still remains this kind of issue. Due to this long-running bug like #6536, I have many files like:
> 
> ```
> <background type="pdf" domain="absolute" filename="././././././foo.pdf" pageno="1"/>
> ```
> 
> This is because I have often opened the file as ./foo.xopp from my terminal. These files are still converted to absolute path.

Do they? I can't reproduce this. I added this case to the CI test.

tattsan:
If I open a file with `xournalpp ./foo.xopp`, relative path is kept relative. However,
If I open a file with `xournalpp foo.xopp`, relative path is converted absolute.

bhennion:
I still cannot reproduce that. Do I get it right: you have a file with 
```
<background type="pdf" domain="absolute" filename="././foobar.pdf" pageno="1"/>
```
that you open with `./install/bin/xournalpp foobar.xopp` (`./install` is my CMAKE_INSTALL_PREFIX)?

For me, the path stays relative (the `././` get removed, but that's it).

tattsan:
Yes, that's my operation. I noticed that the bad behavior appears for all relative paths.
1. If I use `xournalpp foo.xopp`, relative paths ( `foo.pdf`,  `./foo.pdf`,  `././foo.pdf` )  are converted absolute (`/tmp/hoge/foo.pdf`). 
2. If I use `xournalpp ./foo.xopp`, above relative paths become `foo.pdf`.

I wonder what the difference is between your test environment and mine.

tattsan:
Suppose  `/tmp/hoge/foo.pdf` and `/tmp/hoge/bar/foo.xopp` exist, and in `foo.xopp`
```
<background type="pdf" domain="absolute" filename="../foo.pdf" pageno="1"/>
```
3. Do `cd /tmp/hoge/bar` and `xournalpp foo.xopp`. Then it is converted absolute. (NG)
4. Do `cd /tmp/hoge` and `xournalpp bar/foo.xopp`. Then the path is unchanged. (OK)

bhennion:
Are you certain you are running the right version? Your location may change which `xournalpp` is run.

I'll try to do some more tests tomorrow.

tattsan:
I added a bit of debug output to Util::normalizeAssetPath() to see what’s happening with paths like ././foo.pdf. When I opened a .xopp file that contains this kind of path and saved it like this:
```
xournalpp foo.xopp
```
I got this output:
```
xopp-Message: 23:51:46.582: normalizeAssetPath:LineA: fs::proximate(././foo.pdf, ) -> /tmp/hoge/foo.pdf
```
Looks like when base is empty, fs::proximate() just gives back the absolute path. So ././foo.pdf ends up turning into /tmp/hoge/foo.pdf, which probably isn’t what we want.

bhennion:
Ths issue is the base should never be empty. Could you add a backtrace to figure out where this comes from?

tattsan:
The backtrace:
```
#0  0x0000555555cd92ad in Util::normalizeAssetPath(std::filesystem::__cxx11::path const&, std::filesystem::__cxx11::path const&, Util::PathStorageMode) ()
#1  0x0000555555a227fd in SaveHandler::visitPage(XmlNode*, std::shared_ptr<XojPage const>, Document const*, int, std::filesystem::__cxx11::path const&) ()
#2  0x0000555555a20c7d in SaveHandler::prepareSave(Document const*, std::filesystem::__cxx11::path const&) ()
#3  0x0000555555b4e020 in SaveJob::save() ()
#4  0x0000555555b4d9b6 in SaveJob::run() ()
#5  0x0000555555c467b2 in BlockingJob::execute() ()
#6  0x00005555559c5295 in Scheduler::jobThreadCallback(Scheduler*) ()
#7  0x00007ffff675583c in ??? () at /usr/lib/libglib-2.0.so.0
#8  0x00007ffff5ca57eb in ??? () at /usr/lib/libc.so.6
#9  0x00007ffff5d2918c in ??? () at /usr/lib/libc.so.6
```
`SaveHandler::visitPage`'s  5th argument `target`'s value is `foo.xopp`, and this function passes `target.parent_path()` to `Util::normalizeAssetPath`  as 2nd argument `base`. And `target.parent_path()` 's value is empty.

bhennion:
Thanks! I'm sorry it took me so many iterations. I now understand what I was doing wrong, and I could reproduce, and this should be fixed now.

tattsan:
Great! I've tested various paths, and the bug no longer appears!

Next, I tried using symbolic links.
Suppose `foo.pdf` is a symbolic link to `foo_orig.pdf`, and `foo.xopp` refers to `foo.pdf` as its background PDF.
If you open `foo.xopp` and then save it, the stored PDF filename changes to `foo_orig.pdf`.
This isn’t a relativity issue, but it might be annoying for people who want to use symbolic links intentionally.

bhennion:
It should be ok now. Symlinks are not longer resolved and the CI test checks it as well.

tattsan:
Thanks a lot — everything works perfectly now, including symbolic links! Really appreciate your efforts.

One small suggestion for the future: it would be nice if users could choose whether to store the background PDF path as relative or absolute when saving a new .xopp.

Currently it's always relative (unless the ~.xopp is opened~ PDF is already stored with an absolute path), which is fine — but for those who prefer absolute paths, it might be a bit inconvenient.

Maybe in the future there could be an option like:

1. Follow how the PDF was opened (as before),
2. Let the user choose when saving a new .xopp,
3. Or provide a way to manually edit the stored path.

And ideally, a preference setting like:
"Always relative" / "Always absolute" / "Follow original path".

Totally outside the scope of this PR of course — just wanted to share the idea for later. Thanks again for the great work!

tattsan:
I found a wired behavior. Suppose xopp name is `foo.pdf.xopp` and the stored PDF is `./foo.pdf`. After saving, the stored PDF filename isn't normalized, remains `./foo.pdf`. This is harmless, but strange.

bhennion:
> I found a wired behavior. Suppose xopp name is `foo.pdf.xopp` and the stored PDF is `./foo.pdf`. After saving, the stored PDF filename isn't normalized, remains `./foo.pdf`. This is harmless, but strange.

Actually, it's another issue: the file is saved to `foo.xopp` (with the normalized asset path in it). But the save path should certainly not be changed. That's another issue for another PR.

bhennion:
> > I found a wired behavior. Suppose xopp name is `foo.pdf.xopp` and the stored PDF is `./foo.pdf`. After saving, the stored PDF filename isn't normalized, remains `./foo.pdf`. This is harmless, but strange.
> 
> Actually, it's another issue: the file is saved to `foo.xopp` (with the normalized asset path in it). But the save path should certainly not be changed. That's another issue for another PR.

This is now fixed on both release-1.2 and master (and hence on this branch as well).
@tattsan No more issues?

tattsan:
I can no longer find any problems. Thanks again!

bhennion:
Thanks for testing!

tattsan:
Sorry,
```
xournalpp --save=../OnParent.xopp foo.pdf
```
creates a xopp file with absolute path.

tattsan:
The issue I just pointed out seems minor enough, so I’d suggest going ahead with merging this PR. We can tackle the rest in a separate one!

bhennion:
> The issue I just pointed out seems minor enough, so I’d suggest going ahead with merging this PR. We can tackle the rest in a separate one!

I fixed it anyway :) (and added a test for that situation)

tattsan:
Thanks! I confirmed the bug is away!!

PR #6553, opened 2025-07-19
Refactor Windows package script

bhennion:
The Windows package script was a bit weird in various ways:

- It always created the assets in `$GIT_ROOT/windows-setup/`, hence polluting the git tree (there was a .git-ignore, ofc)
- It relied on a generated `xournalpp_version.nsh` to pass the VERSION variable to nsis
- The build directory (from where cmake --build . --target install is run) had to be $GIT_ROOT/build

Those things are fixed here. The package.sh script now takes a build directory as a parameter (defaults to `.`) and the assets are created in that directory. The relevant variables are passed to nsis using the -D flag (and no longer via a generated file).

This will create some changes for people who have been using the script for a while. In particular, they need to clean out their `$GIT_ROOT/windows-setup/`, e.g. by running `git clean -f -- windows-setup/`.

This PR also removes the unless win32/console.cpp (unneeded as the app is now compiled with `-mwindows`).


bhennion:
Because the CI packaging scripts are fetch from the master branch, \action create-installers would not work in this case. I ran it on my branch to check it all worked: https://github.com/bhennion/xournalpp/pull/30#issuecomment-3091989134

The installers generated in that post do work as expected.

bhennion:
@rolandlo I tested this, but if you could give this a quick read, it'd be great!

rolandlo (approved):
LGTM. Nice simplification.

bhennion:
Thanks for the review!
I'll merge it on the release branch (I'm not sure it'll work if we don't)

bhennion:
> I'll merge it on the release branch (I'm not sure it'll work if we don't)

Arf... I fumbled here. Those commits are only on the master branch. Hopefully it'll work like that. Otherwise, I'll cherry-pick https://github.com/xournalpp/xournalpp/pull/6553/commits/dbe3257fed2e93ee622f5e0ada75f4e1c1efae49 onto the release branch.
Sorry about that.

PR #6556, opened 2025-07-20
Automatically detect DPI - manual setting is still available

### LINKED ISSUE #6548: Set DPI Automatically

Setting the DPI manually is confusing. The number you input is not your actual screen's DPI, since the scale factor of the screen comes into play. On my system, KDE Plasma on Linux, the correct setting is DPI ÷ scale factor. I assume other platforms would be similar.

### Feature Description

Since both scale factor and DPI for each display should be available to xournalpp, DPI could be set automatically. 

I know the screen size info for some of my displays (both on the spec sheet and through display metadata) is slightly wrong, so an override or ± offset option would be nice, especially if it could be set per display.

### Alternative Solutions

- The ability for xournalpp to automatically take either DPI or scale factor into account if it can't reliably get both values automatically, and ask the user the other value
- The ability to input DPI and scale factor in preferences, so it is less confusing entering an incorrect dpi
- A note in settings instructing the user to divide their DPI by the scale factor

### Additional Context

May vary from platform to platform.

bhennion:
Fix #6548

The automatic DPI detection is only used for new users or if the user enables it in the settings. If the setting is off, choosing the DPI value by hand (using a slider) is possible.

I guess that it won't work for some corner cases, in particular if dragging the window across monitors (depending on the OS or DE) or if the monitor does not properly report its size.

### DIFF
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -96,12 +96,6 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
                      this);
 #endif
 
-    g_signal_connect(this->window, "realize",
-                     G_CALLBACK(+[](GtkWidget*, gpointer self) { static_cast<MainWindow*>(self)->setDPI(); }), this);
-    g_signal_connect(
-            this->window, "screen-changed",
-            G_CALLBACK(+[](GtkWidget*, GdkScreen*, gpointer self) { static_cast<MainWindow*>(self)->setDPI(); }), this);
-
     // "watch over" all key events
     auto keyPropagate = +[](GtkWidget* w, GdkEvent* e, gpointer) {
         return gtk_window_propagate_key_event(GTK_WINDOW(w), (GdkEventKey*)(e));
--- a/src/core/control/zoom/ZoomControl.cpp
+++ b/src/core/control/zoom/ZoomControl.cpp
@@ -10,6 +10,7 @@
 #include "control/settings/Settings.h"  // for Settings
 #include "control/zoom/ZoomListener.h"  // for ZoomListener
 #include "gui/Layout.h"                 // for Layout
+#include "gui/MainWindow.h"
 #include "gui/PageView.h"               // for XojPageView
 #include "gui/XournalView.h"            // for XournalView
 #include "gui/widgets/XournalWidget.h"  // for gtk_xournal_get_layout
@@ -88,6 +89,16 @@ auto onWindowSizeChangedEvent(GtkWidget* widget, GdkEvent* event, ZoomControl* z
         zoom->updateZoomFitValue();
         layout->recalculate();
     });
+
+    GdkWindow* gdkWindow = gtk_widget_get_window(widget);
+    GdkDisplay* display = gdkWindow ? gdk_window_get_display(gdkWindow) : nullptr;
+    GdkMonitor* monitor = display ? gdk_display_get_monitor_at_window(display, gdkWindow) : nullptr;
+    if (monitor && monitor != zoom->lastMonitor) {
+        zoom->lastMonitor = monitor;
+        const char* monitorName = gdk_monitor_get_model(monitor);
+        g_debug("Window moved to monitor \"%s\"", monitorName);
+        zoom->control->getWindow()->setDPI();
+    }
     return false;
 }
 
--- a/src/core/control/zoom/ZoomControl.h
+++ b/src/core/control/zoom/ZoomControl.h
@@ -259,4 +259,7 @@ private:
     size_t current_page = static_cast<size_t>(-1);
     size_t last_page = static_cast<size_t>(-1);
     bool isZoomFittingNow = false;
+
+    /// The last monitor the window has been moved to -- used for setting dpi
+    GdkMonitor* lastMonitor = nullptr;
 };


rolandlo (commented):
This is great, but it doesn't play nicely in a multi monitor setup yet. I have dpi 160 on my main screen and dpi 108 on my external screen. When I open Xournal++ on my main screen with automatic dpi I get the dpi from the external screen. Only after opening the preferences and deactivating and reactivating the automatic dpi setting I get the correct dpi.
Another issue: even when the dpi is fixed like this I need to press the 100% once more to get 100% zoom.

rolandlo:
See https://discourse.gnome.org/t/incorrect-monitor-size-in-realize-signal/22192 for someone else experiencing the same issue.

bhennion:
I think `screen-changed` signal I connected to does not really work: the problem is [this](https://docs.gtk.org/gdk3/class.Screen.html):

> X originally identified screens with physical screens, but nowadays it is more common to have a single GdkScreen which combines several physical monitors

GDK4 has a signal `GdkSurface::enter-monitor` which could be used. I couldn't find anything of the sort for GTK/GDK 3.

All in all, this feature is neat but I don't think it worth the hassle. At least for now.

src/core/gui/MainWindow.cpp
```diff
@@ -95,6 +96,12 @@ MainWindow::MainWindow(GladeSearchpath* gladeSearchPath, Control* control, GtkAp
                      this);
 #endif
 
+    g_signal_connect(this->window, "realize",
+                     G_CALLBACK(+[](GtkWidget*, gpointer self) { static_cast<MainWindow*>(self)->setDPI(); }), this);
```
rolandlo: Setting the dpi delayed in the Ui Thread fixes the issue (with wrong automatic dpi at startup in a multi monitor setup)
```suggestion
    g_signal_connect(this->window, "realize",
                     G_CALLBACK(+[](GtkWidget*, gpointer self) { Util::execInUiThread([self]() { static_cast<MainWindow*>(self)->setDPI(); }); }), this);
```

bhennion: That's certainly good enough for this case. But I guess we'd still need something for when you drag the window from one monitor to the other.
Currently, I think this should work on X11, but not on Wayland (and I don't know for Windows or MacOS).

rolandlo: Window dragging is caught in the configure-event. Adding these lines to `onWindowSizeChangedEvent`:

```c++
    GdkWindow *gdkWindow = gtk_widget_get_window(widget);
    GdkDisplay *display = gdk_window_get_display(gdkWindow);
    GdkMonitor *monitor = gdk_display_get_monitor_at_window(display, gdkWindow);
    if (monitor) {
        const char* monitorName = gdk_monitor_get_model(monitor);
        if (zoom->lastMonitorName != std::string(monitorName)) {
            g_message("Window moved to %s", monitorName);
            zoom->lastMonitorName = monitorName;
            zoom->control->getWindow()->setDPI();
        }
    }
```
and a member variable `lastMonitorName` in `ZoomControl.h` makes it work.

```term
xopp-Message: 21:44:46.296: Window moved to eDP-1
xopp-Message: 21:44:46.297: Automatic DPI conf: monitor height 190 mm - DPI = 160.421053
xopp-Message: 21:44:47.012: Window moved to HDMI-1
xopp-Message: 21:44:47.013: Automatic DPI conf: monitor height 340 mm - DPI = 107.576471
xopp-Message: 21:44:47.538: Window moved to eDP-1
xopp-Message: 21:44:47.538: Automatic DPI conf: monitor height 190 mm - DPI = 160.421053
```
Connecting to the realize signal is not needed any more.

bhennion:
Thanks!!

There you go. I simply changed the monitor identification to use the pointer rather than the name: I'm not certain every GDK backend will give different names to different monitors.

NB: you can show the debug messages by running `G_MESSAGES_DEBUG=xopp ./install/bin/xournalpp`

rolandlo:
This works fine enough now. I will test on Windows and MacOS to see if it works the same as on Linux there.

rolandlo:
\action create-installers macos-arm windows

rolandlo:
\action create-installers mac-arm windows

rolandlo:
\action create-installers mac-x64

bhennion:

bhennion:
(I rebased on latest master. This should have fixed the Windows package)

rolandlo:
On Windows the PR works as on Linux. On MacOS automatic dpi also works, but only with single monitor. ~With double monitor it can't distinguish between different monitors and~ the monitor names are "(null)" always. I think this is a GTK issue.

rolandlo (approved):
I think this can be merged.

rolandlo:
> On Windows the PR works as on Linux. On MacOS automatic dpi also works, but only with single monitor. With double monitor it can't distinguish between different monitors and the monitor names are "(null)" always. I think this is a GTK issue.

Actually on MacOS different monitors can be distinguished (like on Windows and MacOS), even if they don't get a model string. In a dual monitor setup everything works fine if the monitors use their maximal resolution. If the primary monitor uses some scaling it breaks down. And it only does so if a second monitor is attached. Strange, but nothing we can do about much.
On Linux I tried running with `GDK_BACKEND=x11 GDK_SCALE=2` or applying some fractional scaling and everything still works fine.

bhennion:
Thanks a lot for this @rolandlo!

PR #6567, opened 2025-07-24
Properly lock the document every time Layer::addElement/removeElement() or Page::addLayer/removeLayer() is called

bhennion:
This PR targets the release branch. Every occurrence of Layer::addElement(), Layer::removeElement(), Page::addLayer() or Page::removeLayer() is now within a doc->lock()/doc->unlock() pair.
The proposed changes are minimal to ensure the lock as I wanted to avoid a deeper refactor (which would go to master).

This should help with crashes like #4917.

I tried as many execution paths as I could think of without hitting a deadlock, so hopefully there is no "double locks". More testing would be very appreciated.

NB: in some cases (EditSelection constructors or undo/redo actions), the lock()/unlock() happens in the caller. This isn't great but this has already been refactored in the master branch.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -3006,10 +3006,10 @@ void Control::clipboardPaste(Element* e) {
     e->setY(y);
     layer->addElement(e);
 
+    auto* selection = new EditSelection(this->undoRedo, e, view, page);
     this->doc->unlock();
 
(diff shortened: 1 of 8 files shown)


rolandlo (approved):
LGTM

bhennion:
Thanks for the review!

PR #6573, opened 2025-07-29

bhennion:
Quite a few conflicts with #6567. Most of the changes from #6567 around the EditSelection were not necessary because of the refactor the master branch benefitted from.

It also bumps the MacOS dependencies, so there may be issues related to this.


bhennion:
\action create-installers all

rolandlo (approved):
LGTM

One thing I noticed though while testing: In the master branch (not only in this PR) I can't move the screen with two fingers when touch drawing is enabled and zoom gestures are disabled. This worked on the release branch.

PR #6577, opened 2025-07-31
Fix the SegFault that sometimes happen when quitting without saving

bhennion:
The SegFault is due to gtk_xournal_get_preferred_height() being called after gtk_xournal_dispose(). I don't understand what circumstances make this happen (I lean towards a bug in GTK).

This PR avoids the SegFault by adding simple sanity checks. The weird event gets reported in the terminal/logs.


bhennion:
(the MacOS CI failed because of some poppler server outage)

Issue #6112, opened 2024-12-01
xournalpp crashes sometimes when working with plugins


atticus-sullivan:
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

### (Linux only) Desktop Environment

(bspwm)

### (Linux Only) Display Server

X.Org

### Installation Method

build master with clang

### Xournal++ Version

xournalpp 1.2.4+dev (392f559e)

### libgtk Version

libgtk: 3.24.43

### Bug Description

From time to time xournalpp crashes when I close it.

I'm already trying to find out what action exactly causes this. So far the only thing that always was common is the interaction with some plugins. I suspected maybe https://github.com/xournalpp/xournalpp/pull/5888 is part of this issue bu the plugin I used to generate this crash does not make any changes to the document.

The plugin I used:
```lua
-- Register all Toolbar actions and intialize all UI stuff
function initUi()
	app.registerUi({menu = "mode1", callback= "main", mode=1, toolbarId="mode1", iconName="starred-symbolic"})
end

function main(mode)
	if mode == 1 then
		app.uiAction({["action"]="ACTION_TOOL_PEN"})
		app.uiAction({["action"]="ACTION_TOOL_PEN_SIZE_MEDIUM"})
		app.changeToolColor({["color"] = 0x000000, ["selection"] = false})
	end
end
```

(only with `app.uiAction({["action"]="ACTION_TOOL_PEN"})` there was no issue, but then maybe this only was pure chance)

Usually xournalpp does not crash everytime but only sometimes (when testing right now I managed to crash xournalpp in my 8th try but probably this also depends on how many strokes I draw or how often I invoke the plugin)

<details><summary>Details</summary>

```
Date: Sun Dec  1 12:58:05 2024
Error: signal 11
Xournal++ version 1.2.4+dev
Git commit 392f559e from atticus-sullivan/master
Gtk version 3.24.43

[bt]: (0) ./build/xournalpp(+0xd7fdc) [0x59a2fc4affdc]
[bt]: (1) /usr/lib/libc.so.6(+0x3d1d0) [0x7594b3a4c1d0]
[bt]: (2) /usr/lib/libc.so.6(pthread_mutex_lock+0x4) [0x7594b3aa69e4]
[bt]: (3) ./build/xournalpp(_ZNK6Layout16getMinimalHeightEv+0x13) [0x59a2fc5ae293]
[bt]: (4) ./build/xournalpp(+0x156093) [0x59a2fc52e093]
[...]

PR #6581, opened 2025-08-01
Properly destroy custom widgets when gtk_widget_destroy() is called on them

bhennion:
Our custom widgets with their own dispose/destroy functions did not call their parent class' dispose/destroy. The lead to #6112: calling gtk_widget_destroy() didn't actually remove the widget from its hierarchy.

This PR:
1. ~~Removes custom dispose functions (GObject) in favour of destroy functions (GtkWidget).~~
2. Properly calls the parent class' (GtkWidgetClass) ~~destroy~~ dispose function.

This is a better fix to #6112 than #6577 was.

### DIFF
--- a/src/core/gui/widgets/XournalWidget.cpp
+++ b/src/core/gui/widgets/XournalWidget.cpp
@@ -24,45 +24,22 @@
 
 using xoj::util::Rectangle;
 
-static void gtk_xournal_class_init(GtkXournalClass* klass);
-static void gtk_xournal_init(GtkXournal* xournal);
+/*
+ * Declares:
+ *      static void gtk_xournal_class_init(GtkXournalClass*);
+ *      static void gtk_xournal_init(GtkXournal*);
+ * Defines
+ *      gtk_xournal_parent_class (pointer to GtkWidgetClass instance)
+ *      GType gtk_xournal_get_type();
+ */
+G_DEFINE_TYPE(GtkXournal, gtk_xournal, GTK_TYPE_WIDGET)
+
 static void gtk_xournal_get_preferred_width(GtkWidget* widget, gint* minimal_width, gint* natural_width);
 static void gtk_xournal_get_preferred_height(GtkWidget* widget, gint* minimal_height, gint* natural_height);
 static void gtk_xournal_size_allocate(GtkWidget* widget, GtkAllocation* allocation);
 static void gtk_xournal_realize(GtkWidget* widget);
 static auto gtk_xournal_draw(GtkWidget* widget, cairo_t* cr) -> gboolean;
 static void gtk_xournal_dispose(GObject* object);
-
-auto gtk_xournal_get_type(void) -> GType {
-    static GType gtk_xournal_type = 0;
-
-    if (!gtk_xournal_type) {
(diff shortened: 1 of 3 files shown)


src/core/gui/widgets/XournalWidget.cpp
```diff
@@ -310,4 +287,6 @@ static void gtk_xournal_dispose(GObject* object) {
 
     delete xournal->input;
     xournal->input = nullptr;
+
+    GTK_WIDGET_CLASS(gtk_xournal_parent_class)->destroy(object);
```
rolandlo: Why not stick with dispose functions and calling the parent class' dispose function like this?

```term
 G_OBJECT_CLASS(gtk_xournal_parent_class)->dispose(object);
```

bhennion: It probably doesn't change anything, but GTK people seems to prefer it. See [GTK wiki](https://wiki.gnome.org/HowDoI/CustomWidgets)
> For widgets, the destroy vfunc is equivalent and normally used instead.

Using the destroy vfunc also matches what they do in GTK itself.

rolandlo: I read the opposite. See e.g. https://docs.zrythm.org/md_doc_2dev_2gtk__tips.html or https://discourse.gnome.org/t/memory-leak-dispose-and-finalize-what-are-the-good-practices/11415

Looking at the Gtk code they seem to prefer overriding the dispose function indeed.
See e.g. [gtk/gdk/drawingcontext.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/gdkdrawingcontext.c) or [examples/application10/exampleappprefs.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/examples/application10/exampleappprefs.c) or [examples/application10/exampleappwin.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/examples/application10/exampleappwin.c)

bhennion: The GdkDrawingContext is a bad example: it does not inherit from GtkWidget. But if you look e.g. at [GtkScrolledWindow](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkscrolledwindow.c#L538) or [GtkContainer](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkcontainer.c#L1686), they use destroy.
GtkButton, however, override the dispose vfunc.

So it really doesn't seem to matter. I'll revert to dispose then as you seem to favour it.

bhennion: [This](https://docs.gtk.org/gtk4/migrating-3to4.html#stop-using-the-gtkwidgetdestroy-vfunc) is an argument for using dispose

bhennion: This is done. New using dispose.

rolandlo (approved):

bhennion:
Thanks for the review and catching the destroy/dispose thingy!

PR #6586, opened 2025-08-07
Make audio support optional at compile time

### LINKED ISSUE #1468: Optionalise Portaudio,sndfile&ALSA (and add Pulseaudio support?)

bhennion:
Fix #1468

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,6 @@ endif (APPLE)
 list(APPEND pkg-module-spec "gthread-2.0 >= 2.4.0")
 list(APPEND pkg-module-spec "libxml-2.0 >= 2.0.0")
 list(APPEND pkg-module-spec "libzip >= 1.0.1")
-list(APPEND pkg-module-spec "sndfile >= 1.0.25")
 list(APPEND pkg-module-spec "librsvg-2.0 >= 2.40")
 
 # Profiling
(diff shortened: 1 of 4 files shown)


rolandlo (changes requested):
It doesn't compile without Portaudio (with `-DENABLE_AUDIO=off`). For example there is a line
https://github.com/xournalpp/xournalpp/blob/d2b556bf91f00cff2b8414b4f38108bd9b6316cd/src/core/control/settings/Settings.h#L26-L27
in `Settings.h` which is not guarded by the `ENABLE_AUDIO` flag. But there are more places in the code where Portaudio stuff is used even when `ENABLE_AUDIO` is turned off.

bhennion:
It should be alright now

rolandlo:
> It should be alright now

No, it still doesn't compile. There are many audio related places (e.g. in the Control class, ActionProperties, InputHandler, TextEditor, MainWindow, PageView, PageViewFindObjectHelper, SettingsDialog, Menubar, ToolMenuHandler) that may need to be fixed.

rolandlo (approved):
It works now.

bhennion:
Thanks for the testing and reviewing!

PR #6588, opened 2025-08-07
Expose new action system in Lua API and deprecate old one

rolandlo:
The basic functionality works already. E.g. one can write
```lua
app.activateAction("setsquare")
```
to toggle the setsquare or
```lua
app.changeActionState("set-layout-vertical", false)
```
to change to horizontal layout.

The action states are not always self-explanatory and this still needs to be addressed (brainstorming below). One can write e.g.
```lua
app.changeActionState("select-tool", 4)
```
but how would you know that 4 corresponds to the Text tool?

One idea would be to automatically generate a .lua file that defines all these enum values. The .lua file would have to put somewhere where it can be found.

Another idea would be to add runtime info (something like `app.getAllActionNames()` and `app.getActionStates(action))`
but that would leave the readability problem to the plugin developer and have issues when the state values change between versions for a certain action, e.g. because a new tool was introduced.

@bhennion @atticus-sullivan Any ideas/suggestions?

### DIFF
--- a/src/core/control/ToolEnums.cpp
+++ b/src/core/control/ToolEnums.cpp
@@ -1,99 +1,23 @@
 #include "ToolEnums.h"
 
+#include <algorithm>  // for find, remove_if
+
 #include "model/StrokeStyle.h"
 
-auto toolSizeToString(ToolSize size) -> std::string {
-    switch (size) {
-        case TOOL_SIZE_NONE:
-            return "none";
-        case TOOL_SIZE_VERY_FINE:
-            return "veryThin";
-        case TOOL_SIZE_FINE:
-            return "thin";
-        case TOOL_SIZE_MEDIUM:
-            return "medium";
-        case TOOL_SIZE_THICK:
-            return "thick";
-        case TOOL_SIZE_VERY_THICK:
-            return "veryThick";
-        default:
-            return "";
-    }
-}
 
 auto toolSizeFromString(const std::string& size) -> ToolSize {
-    if (size == "veryThin") {
-        return TOOL_SIZE_VERY_FINE;
-    }
-    if (size == "thin") {
-        return TOOL_SIZE_FINE;
-    }
-    if (size == "medium") {
-        return TOOL_SIZE_MEDIUM;
-    }
-    if (size == "thick") {
-        return TOOL_SIZE_THICK;
-    }
-    if (size == "veryThick") {
-        return TOOL_SIZE_VERY_THICK;
+    auto it = std::find(toolSizeNames.begin(), toolSizeNames.end(), size);
+    if (it != toolSizeNames.end()) {
+        return static_cast<ToolSize>(std::distance(toolSizeNames.begin(), it));
     }
     return TOOL_SIZE_NONE;
 }
 
-auto drawingTypeToString(DrawingType type) -> std::string {
-    switch (type) {
-        case DRAWING_TYPE_DONT_CHANGE:
-            return "dontChange";
-        case DRAWING_TYPE_DEFAULT:
-            return "default";
-        case DRAWING_TYPE_LINE:
-            return "line";
-        case DRAWING_TYPE_RECTANGLE:
-            return "rectangle";
-        case DRAWING_TYPE_ELLIPSE:
-            return "ellipse";
-        case DRAWING_TYPE_ARROW:
-            return "arrow";
-        case DRAWING_TYPE_DOUBLE_ARROW:
-            return "doubleArrow";
-        case DRAWING_TYPE_SHAPE_RECOGNIZER:
-            return "strokeRecognizer";
-        case DRAWING_TYPE_COORDINATE_SYSTEM:
-            return "drawCoordinateSystem";
-        case DRAWING_TYPE_SPLINE:
-            return "spline";
-        default:
-            return "";
-    }
-}
 
 auto drawingTypeFromString(const std::string& type) -> DrawingType {
-    if (type == "dontChange") {
-        return DRAWING_TYPE_DONT_CHANGE;
-    }
-    if (type == "line") {
-        return DRAWING_TYPE_LINE;
-    }
-    if (type == "rectangle") {
-        return DRAWING_TYPE_RECTANGLE;
-    }
-    if (type == "ellipse") {
-        return DRAWING_TYPE_ELLIPSE;
-    }
-    if (type == "arrow") {
-        return DRAWING_TYPE_ARROW;
-    }
-    if (type == "doubleArrow") {
-        return DRAWING_TYPE_DOUBLE_ARROW;
-    }
-    if (type == "strokeRecognizer") {
-        return DRAWING_TYPE_SHAPE_RECOGNIZER;
-    }
-    if (type == "drawCoordinateSystem") {
-        return DRAWING_TYPE_COORDINATE_SYSTEM;
-    }
-    if (type == "spline") {
-        return DRAWING_TYPE_SPLINE;
+    auto it = std::find(drawingTypeNames.begin(), drawingTypeNames.end(), type);
+    if (it != drawingTypeNames.end()) {
+        return static_cast<DrawingType>(std::distance(drawingTypeNames.begin(), it));
     }
     return DRAWING_TYPE_DEFAULT;
 }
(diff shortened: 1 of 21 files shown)


atticus-sullivan:
I think I'd add a `C` (for constants) table to the `app` library which holds all the constants. Similar what kinda was already attempted in the code as I've just seen
https://github.com/xournalpp/xournalpp/blob/01734d85fce88eee8d45bad6f9d4e95d65b45b36/src/core/plugin/luapi_application.h#L3185-L3186 (but without the indirection of a dedicated table for the constants).

If we write the code in a fixed format/style, it should be possible to adapt [`lua_def_file.py`](https://github.com/xournalpp/xournalpp/blob/master/scripts/lua_def_file.py) to also generate type hints (and thereby also suggestions via the IDE/editor) for the available constants. But I'd avoid defining the values in an (external) lua file which needs to be available to run the plugin (and needs to be kept in sync with xournalpp).

We could of course further differentiate between different "kinds" of enums using deeper nested tables (e.g. `app.C.tools.PEN`), but I think this would just complicate the creation of the library (and the type-hints) without giving much benefit over just using `app.C.TOOLS_PEN`.

This way the enums are native in lua, no strings are involved and also everything is kinda static (no dynamic resolution during runtime).

src/core/plugin/luapi_application.h
```diff
@@ -503,7 +525,133 @@ static int applib_registerUi(lua_State* L) {
...
+ * Set the action's state, without triggering callbacks.
+ * Normally you would want to use changeActionState instead.
+ * @param action string
+ * @param state any
+ *
+ * Example 1: app.setActionState("select-tool", 4)  -- but what does 4 mean?
+ * Example 2: app.setActionState("set-layout-vertical", false)
+ * Example 3: app.setActionState("set-columns-or-rows", -3)
+ */
+static int applib_setActionState(lua_State* L) {
```
bhennion: I do not think we should expose this. Do you have an application in mind?
I'm afraid it will leed to internal states that are somewhat implicitly assumed to be impossible.

For instance, the user could set the TOOL state to something else. That wouldn't actually select the tool, but it would make sensitive the toolbar items this tool is compatible with (e.g. dash buttons, filling and so on). I'm not certain what would happens then if the user clicks on one of those illegitimately sensitive buttons.

bhennion:
About the enums: I think the most important thing will be, as you said, that if we add an element to the enum (not necessarily at the end), it doesn't break all the plugins using this enum. In particular, letting the plugin dev use "4" for TEXT is not good.

This is more or less why the Actions themselves are referred to by their string id (and not by the enum value), so whatever solution is used could also be used for the Actions.

For information, how many different enums need to be exposed/translated?

src/core/plugin/luapi_application.h
rolandlo: No, I don''t have an application in mind. So I removed it. Thanks for pointing it out.

rolandlo:
> For information, how many different enums need to be exposed/translated?

Most of the enums from src/core/plugin/ActionBackwardCompatibilityLayer.cpp that have an integer state:
`ToolType`, `ToolSize`, `EraserType` and `EditSelection::OrderChange`, 

Some translation may also be needed for colors.

rolandlo:
> > For information, how many different enums need to be exposed/translated?
> 
> Most of the enums from src/core/plugin/ActionBackwardCompatibilityLayer.cpp that have an integer state: `ToolType`, `ToolSize`, `EraserType` and `EditSelection::OrderChange`,
> 
> Some translation may also be needed for colors.

So the user should write:
```lua
app.changeActionState("select-tool", "eraser")
```
or (with suggestion from the language server)
```lua
app.changeActionState("select-tool", app.C.SELECT_TOOL_ERASER)
```
with `app.C.TOOL_ERASER` defined as the string "eraser".
and `applib_changeActionState` would translate from "eraser" to 4 (or whatever value it has in that version)

Similarly with
```lua
app.changeToolState("eraser-size", "fine")
app.changeToolState("highlighter-size", "very-fine")
app.changeToolState("tool-size", "thick")
app.changeToolState("pen-size", "medium")

app.changeToolState("eraser-type", "whiteout")
app.changeToolState("arrange-selection-order", "bring-to-front")
```

atticus-sullivan:
In case we use the `app.C` option with help of the language server:

Can't we directly translate `app.C.TOOL_ERASER` to `4`?

If `app.C.TOOL_ERASER` is provided directly by xournalpp (`app.C` is constructed when the `app` library is created for lua), the values should be directly coupled to the integers xournalpp uses.

But then, using the string as intermediate value might open the door for more comprehensive debug output when writing a plugin (as the value is human readable).

Also

> letting the plugin dev use "4" for TEXT is not good.

would be acceptable for plugins then (if `app.C.TOOL_ERASER = 4)`. As long as plugin authors use the help of `app.C` I see no issue here (when the enum value changes, the value in `app.C` changes as well so the plugin is still compatible). But then there is the risk plugin authors ignore `app.C`, directly using the value which would lead to incompatibilities if the enum value changes :thinking:

Maybe part of the question here is if we should protect the plugin authors from such misuse by simply not offering the interface to the raw enum values (I'd say no and plugin authors *should* simply use `app.C` instead of manually setting the values, but that's of course opinionated)

bhennion:
> Maybe part of the question here is if we should protect the plugin authors from such misuse by simply not offering the interface to the raw enum values (I'd say no and plugin authors _should_ simply use `app.C` instead of manually setting the values, but that's of course opinionated)

I'm fine with that. It also seems simpler to expose the enum rather than to generate an "enum_to_string" and "enum_from_string" for each of those (although some already exist).

rolandlo:
It all works now (here in VS Code).

Bildschirmfoto vom 2025-08-10 19-34-28

rolandlo:
It would also help to get suggestions for the action names in `app.activateAction` and `app.changeActionState`. This shouldn't be hard to implement via an alias, so I will try.

rolandlo:
> It would also help to get suggestions for the action names in `app.activateAction` and `app.changeActionState`. This shouldn't be hard to implement via an alias, so I will try.

Done!
Bildschirmfoto vom 2025-08-10 20-39-33

rolandlo:
I removed the enum values for DrawingType, StrokeType and OpacityFeature again, since they are not used in any actions.

rolandlo:
@bhennion @atticus-sullivan This PR is up for review.

src/core/control/ToolEnums.h
```diff
@@ -15,75 +15,67 @@
...
-};
+#define FOR_TOOLSIZE(DO)                  \
+    DO(TOOL_SIZE_VERY_FINE, "veryThin")   \
+    DO(TOOL_SIZE_FINE, "thin")            \
+    DO(TOOL_SIZE_MEDIUM, "medium")        \
+    DO(TOOL_SIZE_THICK, "thick")          \
+    DO(TOOL_SIZE_VERY_THICK, "veryThick") \
+    DO(TOOL_SIZE_NONE, "none")
+
+#define DEFINE_ENUMERATION(name, str, ...) name,
```
bhennion: I find those macros impeed readability and are possibly error prone, but I do not have another solution yet. Let me reflect on this for a bit.

rolandlo: Maybe define the enum like before and use a std::map from the enum to std::string to associate the enum values with strings?! It will likely be less performant though.

bhennion: If the enum is contiguous, it could be a static array instead of a map, like for our Action enum.

atticus-sullivan:
> @atticus-sullivan This PR is up for review.

Sorry I won't get to this the next weeks. But from what I've seen the Lua side looks good (no detailed check though)

src/core/control/ToolEnums.h
rolandlo: Do you like it better now (with the static arrays instead of the macros)?

src/core/control/ToolEnums.h
```diff
@@ -24,6 +25,9 @@ enum ToolSize {
     // None has to be at the end, because this enum is used as memory offset
     TOOL_SIZE_NONE
 };
+
+static std::array<std::string, 6> toolSizeNames{"veryThin", "thin", "medium", "thick", "veryThick", "none"};
```
bhennion: I think you can make them constexpr if you use std::string_view:
```suggestion
static constexpr std::array<std::string_view, 6> toolSizeNames{"veryThin", "thin", "medium", "thick", "veryThick", "none"};
```
Then you should have all the operators you need (e.g. `==` actually compares the strings) and the compiler can optimize as constexpr.

rolandlo:
I have added `app.getActionState` (and a helper function `lua_push_gvariant`) Moreover I have removed the expectedTypes array, since the type can be obtained from the gAction as well. @bhennion could you take a look at it?

src/core/control/ToolEnums.cpp
```diff
@@ -322,32 +97,12 @@ auto strokeTypeToLineStyle(StrokeType type) -> LineStyle {
...
-            return "dash";
-        case STROKE_TYPE_DASHDOTTED:
-            return "dashdot";
-        case STROKE_TYPE_DOTTED:
-            return "dot";
-        default:
-            return "";
-    }
-}
+auto strokeTypeToString(StrokeType type) -> std::string { return strokeTypeNames.at(static_cast<size_t>(type)).data(); }
```
bhennion: If we want to be picky: all those could return a std::string_view to avoid any copy. They could then even be declared constexpr.

src/core/plugin/luapi_application.h
```diff
@@ -505,6 +506,189 @@ static int applib_registerUi(lua_State* L) {
 }
 
 /**
+ * Helper function to convert Lua stack items to GVariant*
+ * Currently only boolean, int32 and uint64 are supported.
+ */
+GVariant* lua_to_gvariant(lua_State* L, int idx, const GVariantType* typeHint) {
```
bhennion, line 534: The returned GVariant is floating, right? Could you add a comment saying that?

src/core/plugin/luapi_application.h
```diff
@@ -505,6 +506,189 @@ static int applib_registerUi(lua_State* L) {
...
+    }
+    Action action = Action_fromString(actionStr);
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* control = plugin->getControl();
+    auto* actionDB = control->getActionDatabase();
+    GAction* gAction = G_ACTION(actionDB->getAction(action).get());
+    auto* type = g_action_get_state_type(gAction);
+    GVariant* state = lua_to_gvariant(L, 2, type);
+    if (state) {
+        g_message(g_variant_print_string(state, nullptr, true)->str);
```
bhennion: This is for debug, right? Better not to keep it in the final commit (or at least condition it to some -DDEBUG_smoething variable)

src/core/control/ToolEnums.cpp
rolandlo: Thanks for the review. I changed the ToString methods in a separate commit. Some files had to be touched (including the ToolEnumsTest) using those methods, that are otherwise unrelated to this PR.

atticus-sullivan (approved):
First of all: Sorry it took me so long looking into this.

---

Looks good to me.

Just one question/observation: Is it correct none of the plugins bundled with xournalpp uses/needs to use `app.C` currently? :thinking:

Maybe to a user it looks a bit inconsistent not using `app.C` for the `Action` as well. But looking at the code it totally makes sense. Also the current state using strings and listing all values in the lua-def leading to the suggestions by the LSP (important in my opinion) is also fine.

rolandlo:
Thanks for the review @atticus-sullivan 
> Just one question/observation: Is it correct none of the plugins bundled with xournalpp uses/needs to use `app.C` currently? 🤔
> 
Yes, that's correct.

I have cleaned up the commit history and rebased on current master. Also I had to drop one commit that I introduced after @bhennion's review for simplifying getting the expected type of the action. It didn't work properly since there can be a state type without a parameter type for an action. 

Merging in 24 hours if no objections are raised, assuming the pipeline succeeds.

PR #6589, opened 2025-08-08
Fix CI build on Debian

bhennion:
See https://github.com/xournalpp/xournalpp/actions/runs/16819939609/job/47644731062


bhennion:
\action create-installers debian

PR #6595, opened 2025-08-11
Fixed compilation error

lluisalemanypuig:
Using the method `.getObjectHandle()`.


bhennion:
Thanks for the fix. For information, with which version of QPDF were you having this issue?

lluisalemanypuig:
Hello! I've just checked. The version that was installed via
```
sudo apt install libqpdf-dev
```
in Ubuntu 24.04 is `11.9.0`. (I had to look at the values of the macros `QPDF_*_VERSION` where * is `MAJOR`, `MINOR`, `PATCH`).

PR #6601, opened 2025-08-12
feat: Add configurable text placeholder tool with Lua API and UI

### LINKED ISSUE #6017: Allow plugins to write to status/tool bars

Currently plugins have no way to communicate information through the permanent xournalpp interface, it seems they can only open transient dialogs. It would be nice to be able to do that, for instance in the context of https://github.com/raw-bacon/vi-xournalpp/issues/2.

### Feature Description

Add to the Lua API a function allowing to add a GTK widget which is not necessarily a button to any xournalpp toolbar and return the widget for potential later updates. Add as a special case a function allowing to easily add a GTK Label object (and to update it of course). Even better would be to have a lua function simply returning the whole UI for use with lgi, but I guess this could create issues.

### Alternative Solutions

_No response_

### Additional Context

_No response_

youssefadly237:
## Description

- Implement a configurable text placeholder tool with config file support and toolbar integration
- Add Lua API for live updates of placeholder values
- Improve handling of multi-line and long values (escaping, truncation, ellipsis)

## Notes

This is my first contribution in C++. My background is mainly in Python, so I may have overlooked some C++ best practices. I'm open to feedback and happy to adjust the implementation based on reviewer suggestions.

fixes #6017


atticus-sullivan:
Hi, thanks for the contribution. First of all, sadly I won't have time to have a more detailed look here for the next time (but also I'm mostly familiar with the lua-api and I'm not too deep in the internals/the design of the rest of the code).

With that said, some comments from me:

tbh I didn't think too much about the way to configure all this (how it looks in the sidebar.ini and the placeholder.ini

The `#include "plugin/luapi_application.h"` in `src/core/control/XournalMain.cpp` looks a bit odd. But I guess this is related to a more fundamental structural issue. The object managing the UI element probably shouldn't be implemented in `src/core/plugin` (which is really just the place to define the interface to lua). Instead I think (not 100% sure here) this would be part of the *controller* (refer to the [*model-view-controller*](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) architecture/pattern (MVC) xournalpp uses).

Also these global variables (`g_textPlaceholderConfig` and `g_control`) is something you'd normally try to avoid. Instead the right place again would be probably the controller (`Control.h`). In the `luapi_application.h` you would then use `plugin->getControl()->getPlaceholderConfig()` or so.
I personally would implement all the lua specifics in the `luapi_application.h` (everything including the `lua_state`, specifically I refer to `set_placeholder_value` in `src/core/plugin/luapi_textplaceholder.cpp` here)

---

But then keep in mind, I'm no expert in applying MVC and not too deep in the way xournalpp implements it.

youssefadly237:
Got it, thanks for the feedback! I’ll recheck the structure with MVC in mind, move the Lua-specific parts accordingly, and update the PR soon.

youssefadly237:
Got it, thanks for the feedback! I'll recheck the structure with MVC in mind, move the Lua-specific parts accordingly, and update the PR as soon as I can.

bhennion:
If I may, it seems your implementation is a bit more complicated than required.
The toolbars are configurable, so all you have to do to add a new item in there (e.g. a GtkLabel with some text) is to mimick what is done in `src/core/gui/toolbarMenubar/PluginToolButton.cpp`.
You can in some new PluginToolInfo class (or whatever name) make createItem() create a label and register the PluginToolInfo instance to the Plugin instance.
Then, the Plugin instance can interact with the toolbar item any way it wants.

So I would say, there is no need at all for the whole placeholder.ini. The user will just have to add anywhere their toolbar the toolbar item corresponding.

youssefadly237:
Thanks for the feedback! I tried to avoid messing too much with the C++ code because I'm not extremely confident with it, which is why I implemented it this way.

I get your point about simplifying and using the toolbar system, but I still have a couple of questions

- how to handle multiple placeholders? (Each would need a unique name for the toolbar, so I would have to adjust the logic that parses `toolbar.ini` to support this.)
- where to store the values for each placeholder? (I am not sure if this is a must, but my logic could handle that, if there is a better approach without it I think it is ok)


an example of how my approach works now

I can create `placeholders.ini`

```ini
[vi-mode]
value=tool
```

and expose it to `vi-xournal`

```lua
function setMyPlaceholder(placeholder, val)
	app.setPlaceholderValue(placeholder, val)
end
```

so I can do something like this

```lua
	-- Mode Selection
	tool = {
		description = "Tool mode",
		buttons = { "t" },
		modes = ALL_MODES,
		call = function()
			currentMode = "tool"
			sticky = false
			setMyPlaceholder("vi-mode", currentMode)
		end,
	},
	color = {
		description = "Color mode",
		buttons = { "c" },
		modes = { "tool" },
		call = function()
			currentMode = "color"
			setMyPlaceholder("vi-mode", currentMode)
		end,
	},
```

bhennion:
The way it works for buttons in the toolbar is explained in the luapi documentation:

> Example 2: app.registerUi({callback ="blueDashedPen", toolbarId="CUSTOM_PEN_1", iconName="bluePenIcon"})
> registers a toolbar icon named "bluePenIcon" executing a function named "blueDashedPen", which can be added
> to a toolbar via toolbar customization or by editing the toolbar.ini file using the name "Plugin::CUSTOM_PEN_1"
> Note that in toolbar.ini the string "Plugin::" must always be prepended to the toolbarId specified in the plugin

So the point is `app.registerUi()` could be adapted (or another function created) to register a new "placeholder".
The list of placeholder (and the value of each one) can be stored by the Plugin instance like Plugin::toolbarButtonEntries does for the customizable plugin buttons. And the Plugin instance would also remember of the GtkLabels it must update when app.setPlaceholderValue(the-number-app.registerUi-returned, "my string") is called.

> How to handle multiple placeholders? (Each would need a unique name for the toolbar, so I would have to adjust the logic that parses `toolbar.ini` to support this.)

Well, no really no. You'd have to mimick `Plugin::registerToolButton()`, essentially.

youssefadly237:
Thanks for the clarification!
I dont think it's necessarily simpler, but I can admit it is cleaner. (it didn’t fully click for me at first because I'm not yet familiar with the full architectural pattern, but I'm learning and looking forward to contributing more to Xournal over time).
I’ll draft the PR for now and start working on it as soon as possible.

youssefadly237:
I've refactored the implementation as suggested, moving the placeholder logic into the plugin system and simplifying the design (the force push was for cleaning debris). Please let me know if further adjustments are needed. Thanks for guiding me through this!

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,72 @@
...
+    std::replace(labelText.begin(), labelText.end(), '\n', ' ');
+    std::replace(labelText.begin(), labelText.end(), '\r', ' ');
+    // Truncate and add ... if too long (e.g., > 32 chars)
+    constexpr size_t maxLen = 32;
+    if (labelText.length() > maxLen) {
+        size_t keep = maxLen - 3;
+        labelText = labelText.substr(0, keep) + "...";
+    }
+    xoj::util::WidgetSPtr item(gtk_label_new(labelText.c_str()), xoj::util::adopt);
+    this->labelWidget = item.get(); // keep raw pointer for updates
```
bhennion: `createItem()` could be called several times (because toolbar items can be put several times in the toolbars), so you would only update the last created label.

Also, the GtkLabel could be destroyed (e.g. if the user edits their toolbar to remove the placeholder), in which case you will get a SegFault when unreferencing `this->labelWidget` (e.g. if the plugin wants to update its value).

I see two solutions (to both problems)
1. Store an std::vector of xoj::util::WidgetSPtr's where you append the created items AND connect to each GtkLabel's [`destroy`](https://docs.gtk.org/gtk3/signal.Widget.destroy.html) signal to remove them from the vector when the widgets are removed from the toolbars.
2. (A bit of a hack) Let Glib/Gtk worry about the memory management: create a GtkLabel that you keep (but never return/display) and use [`g_object_bind_property()`](https://docs.gtk.org/gobject/method.Object.bind_property.html) to sync its value to the values of the created labels createItem() returns. When you update the placeholder value, you then simply update the value of this hidden GtkLabel and the others will sync automatically.

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+auto PluginPlaceholderLabel::getToolDisplayName() const -> std::string {
+    return this->t->description;
+}
+
+// Placeholders don’t really have icons
+auto PluginPlaceholderLabel::getNewToolIcon() const -> GtkWidget* {
+    // Return an empty label for the icon in customization dialog
+    return gtk_image_new();
+}
```
bhennion: If I'm not mistaken, this icon is displayed in the toolbar customize dialog. It may be worth having an icon that represents a placeholder (I don't know, maybe just three dots or something like that).
Not mandatory though.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -205,9 +206,15 @@ tool_item& ToolMenuHandler::emplaceItem(Args&&... args) {
...
+    emplaceItem<PluginPlaceholderLabel>(entry);
+}
 #endif /* ENABLE_PLUGINS */
 
+
 void ToolMenuHandler::initToolItems() {
+    using Cat = AbstractToolItem::Category;
+
     using Cat = AbstractToolItem::Category;
     /**
```
bhennion: No need to add this line twice.

src/core/plugin/luapi_application.h
```diff
@@ -3124,10 +3125,47 @@ static int applib_addToSelection(lua_State* L) {
...
+ * @param description string: The label description shown in the toolbar
+ *
+ * Example:
+ *   app.registerPlaceholder("vi-mode", "Current mode")
+ *
+ * The placeholder can then be updated with app.setPlaceholderValue(id, value).
+ */
+static int applib_registerPlaceholder(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    if (!plugin) return 0;
```
bhennion: Coding conventions:
```suggestion
    if (!plugin) {
        return 0;
    }
```
(You may also remove this test as other functions in this file all assume this is never nullptr)

src/core/plugin/luapi_application.h
```diff
@@ -3124,10 +3125,47 @@ static int applib_addToSelection(lua_State* L) {
...
+ *
+ * @param id string: The placeholder ID
+ * @param value string: The value to display
+ *
+ * Example: app.setPlaceholderValue("myPlaceholder", "Hello World")
+ * Updates the toolbar placeholder with the given value.
  */
+static int applib_setPlaceholderValue(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    if (!plugin) return 0;
```
bhennion: Same here

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
youssefadly237: I don't know how did this make it there

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -1,72 +1,86 @@
...
 auto PluginPlaceholderLabel::getToolDisplayName() const -> std::string {
     return this->t->description;
 }
 
-// Placeholders don’t really have icons
+/// Returns a theme-supported icon for the placeholder tool
 auto PluginPlaceholderLabel::getNewToolIcon() const -> GtkWidget* {
-    // Return an empty label for the icon in customization dialog
-    return gtk_image_new();
+    return gtk_image_new_from_icon_name("insert-text", GTK_ICON_SIZE_LARGE_TOOLBAR);
```
youssefadly237: this is the best icon I could think of, I am not sure it really fits

youssefadly237:
Hi @bhennion , just checking in, is there anything I should adjust or clarify in this PR to help move it forward?

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,86 @@
...
+    constexpr size_t maxLen = 32;
+    if (labelText.length() > maxLen) {
+        size_t keep = maxLen - 3;
+        labelText = labelText.substr(0, keep) + "...";
+    }
+    xoj::util::WidgetSPtr item(gtk_label_new(labelText.c_str()), xoj::util::adopt);
+    GtkWidget* rawLabel = item.get();
+    labelWidgets.push_back(rawLabel);
+    gtk_widget_set_can_focus(rawLabel, false);
+    gtk_widget_set_size_request(rawLabel, 200, -1);
```
bhennion: I don't think the size should be in terms of bytes for several reasons

1. UTF-8 characters may take several bytes and truncating like that could lead to invalid UTF-8 strings, causing an error. For example "€€€€€€€€€€€" will be 33 bytes long and your truncation will remove the last € and the last byte of another one. The resulting string is no longer valid.
2. Even single byte characters can have very different width (depending on the font of course) so you may end up with labels too large anyway.

A possible solution would be to rely on `gtk_label_set_ellipsize()` which is supposed to do just what we need.

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/// Updates all label widgets with new text
+void PluginPlaceholderLabel::setText(const std::string& text) {
+    std::string displayText = text;
+    if (displayText.empty()) {
+        displayText = t->toolbarId;
+    }
+    std::replace(displayText.begin(), displayText.end(), '\n', ' '); // sanitize newlines
+    std::replace(displayText.begin(), displayText.end(), '\r', ' '); // sanitize carriage returns
```
bhennion: For better maintainability, better to have the string sanitation in a single place and to remove the code duplication (with createItem above). Maybe `t->value` should already be sanitized.

Also, `setText()` should update `t->value`, otherwise items created after a `setText()` will not have the right text (tiny corner case, but still).

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,86 @@
...
+        auto self = static_cast<PluginPlaceholderLabel*>(user_data);
+        auto& widgets = self->labelWidgets;
+        widgets.erase(std::remove(widgets.begin(), widgets.end(), widget), widgets.end());
+    }), this);
+    return item;
+}
+
+/// Updates all label widgets with new text
+void PluginPlaceholderLabel::setText(const std::string& text) {
+    std::string displayText = text;
```
bhennion: If you are going to copy the string anyway, then better to pass the argument by value rather than reference. This way, the caller could do `setText(std::move(str))` to avoid the extra copy.

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/// Returns the display name for the placeholder tool
+auto PluginPlaceholderLabel::getToolDisplayName() const -> std::string {
+    return this->t->description;
+}
+
+/// Returns a theme-supported icon for the placeholder tool
+auto PluginPlaceholderLabel::getNewToolIcon() const -> GtkWidget* {
+    return gtk_image_new_from_icon_name("insert-text", GTK_ICON_SIZE_LARGE_TOOLBAR);
```
bhennion: Maybe "dialog-information"? It is a [standard icon](https://specifications.freedesktop.org/icon-naming-spec/latest/), typically a circled i like this: 🛈
It's not that important, take whatever you prefer.

youssefadly237:
Thanks for your review! I've addressed all the requested changes

bhennion:
The formatting CI tests are failing. Can you please run `git clang-format master` and `scripts/lua_def_file.py`?

@rolandlo @atticus-sullivan Does the API look good on the lua side? Does it work as you'd expect?

rolandlo:
> The formatting CI tests are failing. Can you please run `git clang-format master` and `scripts/lua_def_file.py`?
> 
> @rolandlo @atticus-sullivan Does the API look good on the lua side? Does it work as you'd expect?

It works as expected and the API looks reasonable. I'm just wondering if we should support adding widgets to the toolbar created and managed on the Lua side (via LuaGObject) instead of adding one kind of widget after the other (buttons, labels, what's next?) to the toolbar. Something like this might be possible (after adding API for `addToolbarWidget`) then:

```lua
local lgi = require "LuaGObject"
local Gtk = lgi.Gtk
lbl = Gtk.Label({ label = "Current mode" })   -- userdata
lbl:set_tooltip_text("This text described the label")

app.addToolbarWidget({ id = "vi-mode", description = "Current mode", widget = lbl, icon_name = "dialog-information" })

-- when updating the label
lbl:set_label("New mode")
```

Of course this would only work for users having LuaGObject installed, so providing API for some basic stuff like buttons and labels may be may be required anyway.

youssefadly237:
Hi @rolandlo,

LuaGObject looks like a great way to enable more complex plugin functionality. I suggest keeping the current button and label API as well, since it's simpler for users who use plugins but don’t have development experience. I see the potential of a more general system, but there hasn't been a request for it yet, and the Xournal plugin community is still small, so it might be overkill for now. I can start working on the more general `addToolbarWidget()` implementation for advanced use cases whenever convenient.

rolandlo:
> Alright, LGTM then. @rolandlo do you want to give the code a read?

I'd like to take a look tomorrow.

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.h
```diff
@@ -0,0 +1,43 @@
...
+    ~PluginPlaceholderLabel() override;
+
+    std::string getToolDisplayName() const override;
+
+    // Extra: allow updating the label text later
+    void setText(std::string text);
+
+protected:
+    xoj::util::WidgetSPtr createItem(bool horizontal) override;
+    GtkWidget* getNewToolIcon() const override;  // unused but required by base
```
rolandlo: This is used for the toolbar customization dialog, right?
```suggestion
    GtkWidget* getNewToolIcon() const override;
```

src/core/gui/toolbarMenubar/PluginPlaceholderLabel.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "plugin/Plugin.h"  // for ToolbarPlaceholderEntry
+#include "util/glib_casts.h"
+#include "util/gtk4_helper.h"  // for gtk_popover_new
+
+#include "config-features.h"
+
+#ifdef ENABLE_PLUGINS
+
+/// Constructor: links placeholder entry to this label
```
rolandlo: The method descriptions should be put into the header.

src/core/plugin/Plugin.cpp
```diff
@@ -326,4 +327,38 @@ auto Plugin::isDefaultEnabled() const -> bool { return defaultEnabled; }
...
+        return;
+    }
+    for (auto& pair: toolbarPlaceholderEntries) {
+        g_message("Plugin: Adding placeholder to ToolMenuHandler: %s", pair.first.c_str());
+        toolMenuHandler->addPluginPlaceholderItem(pair.second.get());
+    }
+}
+
+void Plugin::registerPlaceholder(const std::string& toolbarId, const std::string& description) {
+    if (toolbarPlaceholderEntries.find(toolbarId) == toolbarPlaceholderEntries.end()) {
```
rolandlo: I think there should be some prefix like for plugin toolbar buttons.

src/core/plugin/Plugin.h
```diff
@@ -154,6 +167,8 @@ class Plugin final {
 private:
     /// Load ini file
     void loadIni();
+    // Storage for toolbar placeholder entries
+    std::unordered_map<std::string, std::unique_ptr<ToolbarPlaceholderEntry>> toolbarPlaceholderEntries;
```
rolandlo: This line would better fit next to the one for `toolbarButtonEntries`.

rolandlo:

rolandlo:
Thanks @youssefadly237 for your contribution! I merged it.

Issue #6450, opened 2025-05-24
Please add an option to make layers global (for all pages)


Photon89:
### Is Your Feature Request Related to a Problem?

Recently I conducted oral exams and intended to use Xournal++ for making notes. I had a background PDF with lecture script and wanted to tick off and/or add comments for each student's presentations. To do this, I created n layers, one layer for each students. This way, I could switch visibility and only show the notes for a particular student.

During the exams to my surprise I noticed that the layers weren't there. At some point I figured out what happened: They were there, but only on the first page... Since I had no time to create n layers in each page, I resorted to using different colors. It would be useful to have each layer to span over all pages in a document though. Maybe there are use cases where this is undesirable, so probably it should be optional.

### Feature Description

An option to make layers span over all pages in a document.

lluisalemanypuig:
I'm working on making the possibility of any layer to span multiple pages (and thus can span over the entire document). A user creates a layer on a page (or selects an existing one) and can make it span to the page previous or next to the current one. For now, this has to be repeated as many times as pages the user wants the layer to span over. When drawing on the layer that spans multiple pages, the view of the layer is updated on all pages.

I still have to work on how to save/load the document with such layers, and work out other details, such as
- duplicating a layer that spans multiple pages: is it duplicated on only one layer, or over the entire range?
- deleting a layer that spans multiple pages: is it deleted on only one layer, or over the entire range?

Does this have any chance of actually being accepted? If not, I'll stop working on this, since it is not really worth it for just me.

rolandlo:
Saving/loading will likely create issues with backward compatibility. I share @bhennion's concerns that this is a niche feature and would better be realized via a plugin.

lluisalemanypuig:
Hello! I would like to continue this conversation in the pull request.

lluisalemanypuig:
I kind of agree with the backward compatibility with the loading/writing of documents. However, I haven't been to look into it and maybe it can actually be implemented with not too much difficulty. I mean, one can always write optional fields in the file to indicate the layer span. If they're not there (maybe an old file?) the layer belongs to a single page. If they're there, the layer is a span

lluisalemanypuig:
[...]

PR #6609, opened 2025-08-14
New feature: Layer spans

lluisalemanypuig:
I would like to start a proposal of implementation of the feature request in #6450. I call this layer spans, where a layer can be displayed in more than one page. What I've implemented only supports continuous interval of pages. Watch this 2-minute [video](https://youtu.be/U-_ZBGcTQ1k) showing a demo of what I have done so far.

Implemented features in this PR:
- In a page, a user can select a layer and make it be visible in one of the contiguous page (the next or previous). These two operations are called "Span up" and "Span down"
- To make a layer span a page over more than two pages, a user has to either "Span up" or "Span down" multiple times: go to page N, span down a layer, go to page N+1, then span down the same layer, and so on.
- A layer of a layer span can only be deleted from a page at its endpoints.
- A layer span can be duplicated at any page, and this does not affect the layer span in any way.

Not implemented yet:
- Undo actions for
    - deleting a layer span from a single page (
    - "Span up" or "Span down"
- Redrawing all pages when a span layer has been updated in a specific page. To be more efficient, a page could be redrawn only when a scrolled to that specific page, and only when there are things to be redrawn.
- Saving layer span into the xournal file: this can be done by saving the layer span only at the first page of the span, along with the indices of the page span.
- (Obviously) Reading a layer span from the xournal file: read the pages and layers as if each layer belonged to a single page. Then, update the document so that layer spans actually span over the pages indicated by their span.

Nice to have:
- Set the span of a page by indicating the page numbers, not just by spamming the "Span up" or the "Span down".
- Delete a chunk of layers of a layer span, either up or down starting at an end of the span, indicating the number of this the operation has to be repeated.

Please, do think about whether or not you want this feature. I am willing to commit to finishing it if you like it.

### DIFF
--- a/src/core/control/actions/ActionProperties.h
+++ b/src/core/control/actions/ActionProperties.h
@@ -994,6 +994,16 @@ struct ActionProperties<Action::LAYER_MOVE_DOWN> {
 };
 
 template <>
+struct ActionProperties<Action::LAYER_SPAN_UP> {
+    static void callback(GSimpleAction*, GVariant*, Control* ctrl) { ctrl->getLayerController()->spanUp(); }
+};
+
+template <>
+struct ActionProperties<Action::LAYER_SPAN_DOWN> {
+    static void callback(GSimpleAction*, GVariant*, Control* ctrl) { ctrl->getLayerController()->spanDown(); }
+};
+
+template <>
 struct ActionProperties<Action::LAYER_DELETE> {
     static void callback(GSimpleAction*, GVariant*, Control* ctrl) { ctrl->getLayerController()->deleteCurrentLayer(); }
 };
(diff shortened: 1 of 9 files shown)


lluisalemanypuig:
(Sorry, I messed up the branch and had to fix it. Ignore the push)

lluisalemanypuig:
I have updated this branch with a more complete functionality of this feature (read the commit message)

lluisalemanypuig:
I have updated the feature to be fully implemented:
- When drawing on a layer that spans over multiple pages, drawing in one page updates the element in all the other pages so that when scrolling the element is drawn without the need for zooming in or out.
- The document is saved so that layers that span over multiple pages are only fully written once; the fact that these layers span over multiple pages is ensured by writing a "reference" to the "original" layer.
- The document can be loaded from the new format. This **breaks forward compatibility** in that files created with newer versions of xournal++ can be loaded by older versions but they will create spurious layers (empty layers with no content).

PR #6614, opened 2025-08-16
Fixed a few Todo items

lluisalemanypuig:
I was browsing the source code and I found several "Todo" items, and I thought I could get them out of the way for you. Some of the changes are due to clang-format.

1. TODO: struct for return type in getAudioAttributes.
    
    I added a struct as a return type for a method in the class `AudioQueue`.

2. TODO: fictitious namespace 'std' for easier replacement of library functions.

    So I have replaced the `xoj_utils` namespace in some files with `std`. This is a little unorthodox, but this way the rest of the code uses the `to_underlying`, `to_address`, and `identity` functions in defined in xournalpp as if they were from the standard library. Once the project is upgraded to C++20, using these functions from the STL will be a little easier since you will only have to remove the code and replace the corresponding `#include`.

Other changes that are not "Todo" items

1. I found a lot of `Color` parameters (and return values) that were not declared as constant references. This implies copying, even if it just 4 `uint32_t`. So I changed the function signatures to `const Color&` wherever possible.

2. I removed an unused variable. I looked at the code and that was added on 2021.

### DIFF
--- a/src/util/include/util/PointerContainerView.h
+++ b/src/util/include/util/PointerContainerView.h
@@ -17,9 +17,9 @@
 
 #include "util/safe_casts.h"
 
-namespace xoj::util {
-
 // Todo(cpp20) use std::to_address instead
+namespace std {
+
 template <typename T>
 T* to_address(T*& t) {
     return t;
@@ -38,6 +38,10 @@ template <typename T>
 const T* to_address(const std::unique_ptr<T>& t) {
     return t.get();
 }
+
+}  // namespace std
+
+namespace xoj::util {
 
 template <typename container_type>
 class PointerContainerView {
@@ -60,8 +64,8 @@ public:
         iterator_impl(base_it it): base_it(it) {}
 
         // We must return by value
-        value_type operator*() const { return to_address(base_it::operator*()); }
-        value_type operator[](difference_type n) const { return to_address(base_it::operator[](n)); }
+        value_type operator*() const { return std::to_address(base_it::operator*()); }
+        value_type operator[](difference_type n) const { return std::to_address(base_it::operator[](n)); }
 
     private:
         using base_it::operator->;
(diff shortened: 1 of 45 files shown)


bhennion:
> 1. I found a lot of `Color` parameters (and return values) that were not declared as constant references. This implies copying, even if it just 4 `uint32_t`. So I changed the function signatures to `const Color&` wherever possible.

No, Color has size 4 (bytes), so passing it by reference is less efficient.

https://github.com/xournalpp/xournalpp/blob/70bb88aa7eda2d948c7342217e9bfcf1cc3c26c1/src/util/include/util/Color.h#L70

Please revert this.

lluisalemanypuig:
Oh, my. You're right. I could've sworn those were `uint32_t`. I'm very sorry. I'll revert it now.

src/util/include/util/PointerContainerView.h
```diff
@@ -17,9 +17,9 @@
 
 #include "util/safe_casts.h"
 
-namespace xoj::util {
-
 // Todo(cpp20) use std::to_address instead
+namespace std {
+
 template <typename T>
```
bhennion, line 23: Those todos are only there to remind the devs of those needed changes once we have switched to c++20.
But we haven't changed the c++ standards to 20 yet.
We certainly do not want to define functions within the `std` namespace (this is UB, with a handful of exceptions).

There are similarly some Todo(gtk4) that should not be fixed yet.

lluisalemanypuig:
Ok, I didn't know that redefining `std` was UB.

I was browsing the code when I saw those todos and I thought maybe I could fix those while I learn a bit more about the codebase. And when I finished I thought maybe they would be useful. But I don't think they are very useful, honestly. Sorry, I think I might as well close this since I feel I'm wasting your time, and mine as well. The only two changes that are remotely useful is that struct and the removal of an unused variable.

Issue #5637, opened 2024-04-24
Stroke Recognizer makes straight lines different than expected 


c-ola:
### Operating System

Linux

### (Linux only) Distribution

Fedora

### (Linux only) Desktop Environment

i3

### (Linux Only) Display Server

X.Org

### Installation Method

dnf package manager

### Xournal++ Version

1.2.3

### libgtk Version

3.24.41

### Bug Description

When using the stroke recognizer to draw straight diagonal lines, the line will sometimes get shortened if you go back towards the direction of the starting point (or even in the opposite direction if you go over the starting point).

### Expected Behaviour

I would expect the line to be straight along where it is drawn (go from min to max).

### Steps to Reproduce

1. Start drawing line diagonally.
2. Go back in direction of start point
3. Then release the drawing tool that you are using (occurs with both pen and highlighter

### Additional Context

 It also does not happen when the line is parallel to either the x or y axes.

 Drawing a line in the following way (before releasing the drawing tool):
![image](https://github.com/xournalpp/xournalpp/assets/59809814/c4ee643f-309f-4bb1-b438-06604cb23cf2)
Leads to this image:
![image](https://github.com/xournalpp/xournalpp/assets/59809814/8a3223a6-43df-4dbc-81d8-cd436797a505)

This is an example of it not happening parallel to the x axis.
![image](https://github.com/xournalpp/xournalpp/assets/59809814/3f438c98-0379-4ebc-9d0a-efd610ac8462)
![image](https://github.com/xournalpp/xournalpp/assets/59809814/989a473d-d5fe-4075-84a0-84a383c5ca57)

Febbe:
@kenballus you implemented the preserve the endpoint preservation in #3285, can you tell me, if you introduced the bug there?

kenballus:
Yes; I introduced this behavior. This is how I prefer the stroke recognizer to behave, but that is ultimately just an opinion. Keep in mind that my usual use case for the stroke recognizer is drawing straight lines between preexisting points on a page, so it is in my interested to precisely control the start and end points of a recognized line segment.

c-ola:
In that sense the behaviour makes sense, however, shouldn't it then behave the same way along the x and y axes as well?

kenballus:
Oh, sorry. I misunderstood the issue here.

I am not sure whether I introduced this behavior. If I did, it was definitely unintentional, and should be fixed.

lluisalemanypuig:
[...]

PR #6622, opened 2025-08-19
Try to avoid shortening line strokes unnecessarily

lluisalemanypuig:
Lines drawn from point P to point Q may be recognized by the stroke recognizer and drawn as a straight line. However, if there is a third point R as a result of an accidental touch of the device (stylus against the pad, or late release of the mouse's button) the stroke recognizer may shorten the line in an unexpected way (#5637).

This PR proposes a fix to avoid shortening a line when R is collinear with P and Q. The test used does not check if P, Q and R line on the same line, but rather the (squared) distance from R to the line through P and Q: when close enough to 0, R is not considered to make the line, when large enough R is used as an endpoint, in order to respect the current behaviour when this is justifiable. I found this method more robust than using an orientation test, that is, the side of the oriented line PQ on which the point R lies (when close to 0, the points are collinear).

Behaviour _before_ the changes are shown in [this video](https://youtu.be/0Qj5ZXaz4MI).
Behaviour _after_ the changes are shown in [this video](https://youtu.be/h2QIhu2Tlcg).

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -333,9 +333,26 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinSize) ->
                 s->addPoint(Point(rs->x1, rs->y1));
                 s->addPoint(Point(rs->x2, rs->y2));
             } else {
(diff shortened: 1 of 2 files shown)


lluisalemanypuig:
Happy to help! Xournal is a great project.

I'm happy with this PR being merged to master.

PR #6623, opened 2025-08-19
Refactor parsing on Settings.cpp

tiziodcaio:
Wrap up castings in template functions to make the file more readable

### DIFF
--- a/src/core/control/settings/Settings.cpp
+++ b/src/core/control/settings/Settings.cpp
@@ -463,6 +463,7 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     PARSE(strokeFilterEnabled)
     PARSE(doActionOnStrokeFiltered)
     PARSE(trySelectOnStrokeFiltered)
+    PARSE(numIgnoredStylusEvents)
 
     PARSE(latexSettings.autoCheckDependencies)
     PARSE(latexSettings.defaultText)
@@ -511,9 +512,10 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         this->minimumPressure = std::max(0.01, parse<double>(value));
     } else if (name == "sidebarNumberingStyle") {
         auto num = parse<int>(value);
-        if (num < static_cast<int>(SidebarNumberingStyle::MIN) || static_cast<int>(SidebarNumberingStyle::MAX) < num) {
-            num = static_cast<int>(SidebarNumberingStyle::DEFAULT);
+        if (num < SidebarNumberingStyle::MIN || SidebarNumberingStyle::MAX < num) {
+            this->sidebarNumberingStyle = SidebarNumberingStyle::DEFAULT;
             g_warning("Settings::Invalid sidebarNumberingStyle value. Reset to default.");
+            return;
         }
         this->sidebarNumberingStyle = static_cast<SidebarNumberingStyle>(num);
     } else if (name == "sidebarWidth") {
@@ -527,7 +529,7 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     } else if (name == "themeVariant") {
         this->themeVariant = themeVariantFromString(static_cast<string>(value));
     } else if (name == "pageTemplate") {
-        this->pageTemplateSettings.parse(value);
+        this->pageTemplateSettings.parse(parse<string>(value));
     } else if (name == "defaultViewModeAttributes") {
         this->viewModes.at(VIEW_MODE_DEFAULT) = settingsStringToViewMode(static_cast<string>(value));
     } else if (name == "fullscreenViewModeAttributes") {
@@ -550,17 +552,7 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
         setParsed(this->addHorizontalSpaceAmountLeft, value);
         setParsed(this->addHorizontalSpaceAmountRight, value);
     } else if (name == "scrollbarHideType") {
-        if (value == "both") {
-            this->scrollbarHideType = SCROLLBAR_HIDE_BOTH;
-        } else if (value == "horizontal") {
-            this->scrollbarHideType = SCROLLBAR_HIDE_HORIZONTAL;
-        } else if (value == "vertical") {
-            this->scrollbarHideType = SCROLLBAR_HIDE_VERTICAL;
-        } else {
-            this->scrollbarHideType = SCROLLBAR_HIDE_NONE;
-        }
-    } else if (name == "numIgnoredStylusEvents") {
-        this->numIgnoredStylusEvents = static_cast<int>(parse<unsigned int>(value));
+        this->scrollbarHideType = stringToScrollbarHideType(value);
     } else if (name == "emptyLastPageAppend") {
         this->emptyLastPageAppend = emptyLastPageAppendFromString(static_cast<string>(value));
     }
@@ -730,14 +722,11 @@ xmlNodePtr Settings::saveProperty(const std::string& key, T value, xmlNodePtr pa
     std::string str{};
 
     if constexpr (std::is_same_v<T, int>) {
-        const auto result = std::to_chars(buffer, buffer + sizeof(buffer), value);
-        str = {buffer, result.ptr};
+        str = std::to_string(value);
     } else if constexpr (std::is_same_v<T, uint32_t>) {
-        const auto result = std::to_chars(buffer, buffer + sizeof(buffer), value);
-        str = {buffer, result.ptr};
+        str = std::to_string(value);
     } else if constexpr (std::is_same_v<T, double>) {
-        const auto result = std::to_chars(buffer, buffer + sizeof(buffer), value);
-        str = {buffer, result.ptr};
+        str = g_ascii_dtostr(buffer, 20, value);
     } else if constexpr (std::is_same_v<T, bool>) {
         str = value ? "true" : "false";
     } else if constexpr (std::is_same_v<T, fs::path>) {
@@ -2179,7 +2168,7 @@ void Settings::setIgnoredStylusEvents(int numEvents) {
     if (this->numIgnoredStylusEvents == numEvents) {
         return;
     }
-    this->numIgnoredStylusEvents = std::max<int>(numEvents, 0);
+    this->numIgnoredStylusEvents = static_cast<unsigned int>(std::max<int>(numEvents, 0));
     save();
 }
 
--- a/src/core/control/settings/Settings.h
+++ b/src/core/control/settings/Settings.h
@@ -1120,9 +1120,9 @@ private:
 
     /**
      * How many stylus events since hitting the screen should be ignored before actually starting the action. If set to
-     * 0, no event will be ignored. Should not be negative.
+     * 0, no event will be ignored. It cannot be negative.
      */
-    int numIgnoredStylusEvents{};
+    unsigned int numIgnoredStylusEvents{};
 
     /**
      * Whether Wacom parameter TabletPCButton is enabled
--- a/src/core/control/settings/xmlutils.h
+++ b/src/core/control/settings/xmlutils.h
@@ -11,7 +11,6 @@
 #pragma once
 
 #include <charconv>
-#include <format>
 #include <string>
 #include <string_view>
 #include <type_traits>
@@ -93,16 +92,7 @@ T parse(const std::string_view strView, T defaultValue) {
     } else if constexpr (std::is_same_v<T, int>) {
         val = std::stoi(str);
     } else if constexpr (std::is_same_v<T, double>) {
-        double d;
-
-        const auto [ptr, ec] = std::from_chars(str.c_str(), str.c_str() + sizeof(str), d);
-
-        if (ec == std::errc::invalid_argument) {
-            g_error("Invalid argument");
-        }
-        if (ec == std::errc()) {
-            val = d;
-        }
+        val = g_ascii_strtod(str.c_str(), nullptr);
     } else if constexpr (std::is_same_v<T, bool>) {
         val = str == "true";
     } else if constexpr (std::is_same_v<T, Color> || std::is_same_v<T, ColorU8>) {
@@ -116,7 +106,8 @@ T parse(const std::string_view strView, T defaultValue) {
     } else if constexpr (std::is_same_v<T, XojFont>) {
         val = str;
     } else {
-        const std::string err = std::format("Xournalpp does not support the required type: {}", typeid(T).name());
+        char err[50];
+        std::snprintf(err, 50, "Xournalpp does not support the required type: %s", typeid(T).name());
         throw(std::runtime_error{err});
     }
 
--- a/src/core/control/ToolHandler.cpp
+++ b/src/core/control/ToolHandler.cpp
@@ -446,8 +446,8 @@ void ToolHandler::saveSettings() const {
         }
 
         static constexpr unsigned int SHAPE_CAPS = TOOL_CAP_RULER | TOOL_CAP_RECTANGLE | TOOL_CAP_ELLIPSE |
-                                           TOOL_CAP_ARROW | TOOL_CAP_DOUBLE_ARROW | TOOL_CAP_RECOGNIZER |
-                                           TOOL_CAP_SPLINE;
+                                                   TOOL_CAP_ARROW | TOOL_CAP_DOUBLE_ARROW | TOOL_CAP_RECOGNIZER |
+                                                   TOOL_CAP_SPLINE;
         if (tool->capabilities & SHAPE_CAPS) {
             st.set("drawingType", drawingTypeToString(tool->getDrawingType()).data());
         }
--- a/src/core/control/settings/SettingsEnums.h
+++ b/src/core/control/settings/SettingsEnums.h
@@ -107,8 +107,11 @@ enum class SidebarNumberingStyle {
     /* MIN, MAX and Default values. Need to be updated when new style is added. */
     MIN = SidebarNumberingStyle::NONE,
     MAX = SidebarNumberingStyle::NUMBER_WITH_SQUARE_BACKGROUND,
-    DEFAULT = SidebarNumberingStyle::NUMBER_BELOW_PREVIEW
+    DEFAULT = SidebarNumberingStyle::NUMBER_BELOW_PREVIEW,
 };
+
+inline bool operator<(SidebarNumberingStyle e, const int num) { return num < static_cast<int>(e); }
+inline bool operator<(const int num, SidebarNumberingStyle e) { return static_cast<int>(e) < num; }
 
 constexpr auto buttonToString(Button button) -> const char* {
     switch (button) {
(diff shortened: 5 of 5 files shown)


src/core/control/settings/xmlutils.cpp
```diff
@@ -0,0 +1,67 @@
...
+template <typename T>
+T parse(const xmlChar* xml, T defaultValue) {
+    if (xml == nullptr)
+        return defaultValue;
+
+    return cast(reinterpret_cast<const char*>(xml), defaultValue);
+}
+
+template <typename T>
+T cast(const string& str, T defaultValue) {
```
bhennion: To avoid the extra copy
```suggestion
T cast(std::string_view str, T defaultValue) {
```

src/core/control/settings/xmlutils.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if (str.empty())
+        return defaultValue;
+
+    if (str.empty())
+        return defaultValue;
```
bhennion: By convention, all your conditional (should) have brackets
```suggestion
    if (str.empty()) {
        return defaultValue;
    }
```
There are many other occurrences of this in your PR

src/core/control/settings/xmlutils.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <libxml/tree.h>
+#include <libxml/xmlmemory.h>
+#include <libxml/xmlstring.h>
+
+#include "util/Color.h"
+
+#include "filesystem.h"
+
+using namespace std;
```
bhennion: We avoid `using namespace foo` or `using nmspce::bar` in most files, and forbid them in any file that is included (this is error prone).

src/core/control/settings/Settings.cpp
```diff
@@ -28,7 +28,7 @@
 #include "config-dev.h"    // for PALETTE_FILE
 #include "config-dev.h"
 #include "filesystem.h"  // for path, u8path, exists
-
+#include "xmlutils.cpp"  // for parse, cast, getXml
```
bhennion: Make the unit header-only (i.e. remove xmlutils.h and rename xmlutils.cpp into xmlutils.h").

src/core/control/settings/Settings.cpp
```diff
@@ -276,462 +276,390 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
     for (xmlNodePtr x = cur->children; x != nullptr; x = x->next) {
-        if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("data"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            parseData(x, elem.child(reinterpret_cast<const char*>(name)));
-            xmlFree(name);
-        } else if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("attribute"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            xmlChar* value = xmlGetProp(x, reinterpret_cast<const xmlChar*>("value"));
-            xmlChar* type = xmlGetProp(x, reinterpret_cast<const xmlChar*>("type"));
+        const auto type = parse<string>(x->name);
```
bhennion: You should return an std::string_view to avoid the extra copy

src/core/control/settings/xmlutils.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+template <typename T>
+xmlChar* str2xmlChar(const T& str) {
+    if constexpr (is_same_v<T, string>)
+        return reinterpret_cast<xmlChar*>(str.c_str());
+    else if constexpr (std::is_convertible_v<T, xmlChar*>)
+        return reinterpret_cast<xmlChar*>(str);
+    else
+        return nullptr;
```
bhennion: This is dangerous: we would not detect an issue with a wrong type. You can add a static_assert(false) or something like that.

src/core/control/settings/xmlutils.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <libxml/xmlstring.h>
+
+#include "util/Color.h"
+
+#include "filesystem.h"
+
+using namespace std;
+
+template <typename T>
+xmlChar* str2xmlChar(const T& str) {
```
bhennion: I'm not sure how much this one is useful.
It is mostly applied to string litteral for which a string litteral operator makes more sense (e.g. "name"xml to get "name" as a const xmlChar*, defined as `constexpr const xmlChar* operator ""xml(const char*) { return reinterpret_cast...; }`.

src/core/control/settings/Settings.h
```diff
@@ -135,7 +131,7 @@ class Settings {
     ViewModeId getActiveViewMode() const;
 
     bool isPressureSensitivity() const;
-    void setPressureSensitivity(gboolean presureSensitivity);
+    void setPressureSensitivity(int pressureSensitivity);
```
bhennion: This should be a boolean
```suggestion
    void setPressureSensitivity(bool pressureSensitivity);
```

src/core/control/settings/Settings.cpp
```diff
@@ -763,48 +691,48 @@ void Settings::loadButtonConfig() {
             ToolType type = toolTypeFromString(sType);
             cfg->action = type;
 
-            if (type == TOOL_PEN) {
-                string strokeType;
-                cfg->strokeType =
-                        e.getString("strokeType", strokeType) ? strokeTypeFromString(strokeType) : STROKE_TYPE_NONE;
-            }
+            switch (type) {
```
bhennion: Probably a matter of opinion, but I find this switch without break way less readable than the previous ifs.

src/core/control/settings/Settings.cpp
```diff
@@ -276,462 +276,390 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
+    } else if (name == "showToolbar") {
+        this->showToolbar = cast<bool>(value);
+    } else if (name == "filepathShownInTitlebar") {
+        this->filepathShownInTitlebar = cast<bool>(value);
+    } else if (name == "pageNumberShownInTitlebar") {
+        this->pageNumberShownInTitlebar = cast<bool>(value);
+    } else if (name == "showSidebar") {
+        this->showSidebar = cast<bool>(value);
+    } else if (name == "sidebarNumberingStyle") {
+        auto num = cast<int>(value);
```
bhennion: So `cast()` actually parses, right? Maybe is should be renamed.

tiziodcaio:
Thank you for your review! It's my first PR with CPP code and I need surely more experience!
ASAP I'll read your comments

src/core/control/settings/Settings.h
tiziodcaio: Image

( [source](https://docs.gtk.org/glib/types.html#gboolean) )

Are we sure?

src/core/control/settings/Settings.cpp
tiziodcaio: reverted.

src/core/control/settings/Settings.cpp
tiziodcaio: changed to parse ;-)

bhennion:
The code looks good to me.
I'd be reassured if we had a test unit for the settings load/save mechanism (we don't) to make sure this doesn't break anything, but that'd be gravy.

I'll try to test it locally later (maybe today).

src/core/control/settings/Settings.cpp
```diff
@@ -276,462 +275,389 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
         }
     }
 }
 
 void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     // Parse data map
-    if (!xmlStrcmp(cur->name, reinterpret_cast<const xmlChar*>("data"))) {
-        xmlChar* name = xmlGetProp(cur, reinterpret_cast<const xmlChar*>("name"));
-        if (name == nullptr) {
+    if (cur->name == "data"_xml) {
```
bhennion: Here, you are comparing the pointer values instead of the strings...
```suggestion
    if (std::basic_string_view(cur->name) == "data"_xml) {
```
There are other instances of this.

bhennion: (you could also have some literal "data"_xml_sv that creates a basic_string_view<xmlChar>. that would work as well)

tiziodcaio: solved!

tiziodcaio:
I used u8string and u8char_t naming convention (C++20); if you dislike it, we can change back using xmlChar*

src/core/control/settings/xmlutils.h
```diff
@@ -0,0 +1,99 @@
...
+
+#include "util/Color.h"
+
+#include "xmlutils.h"
+
+// C++20 stubs
+using char8_t = unsigned char;
+
+using u8string = std::basic_string<char8_t>;
+using u8string_view = std::basic_string_view<char8_t>;
```
bhennion: I don't think spoofing c++2* type is a good idea. It'll cause trouble when we move to c++20. It also relies on a coincidence (that xmlChar = unsigned char and char8_t as well - at least after c++23).

src/core/control/settings/Settings.cpp
```diff
@@ -276,462 +275,385 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
     for (xmlNodePtr x = cur->children; x != nullptr; x = x->next) {
-        if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("data"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            parseData(x, elem.child(reinterpret_cast<const char*>(name)));
-            xmlFree(name);
-        } else if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("attribute"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            xmlChar* value = xmlGetProp(x, reinterpret_cast<const xmlChar*>("value"));
-            xmlChar* type = xmlGetProp(x, reinterpret_cast<const xmlChar*>("type"));
+        const string type = reinterpret_cast<const char*>(x->name);
```
bhennion: You went back to unnecessarily copying the data (here and below). This probably is not a big deal as we don't load the setttings so often, but it certainly is bad practice.

The point I'm making here (and I've been making in previous comments as well) is: the constructor `std::string(const char*)` copies the data. So better to use `std::string_view` which behaves like a `std::string` except the data is not owned (and thus never copied).

It is the same for std::basic_string<xmlChar> and std::basic_string_view<xmlChar>.

src/core/control/settings/xmlutils.h
```diff
@@ -0,0 +1,99 @@
...
+inline const char8_t* operator""_xml(const char* ch, size_t) { return reinterpret_cast<const char8_t*>(ch); }
+
+inline u8string operator""_xmlstr(const char* ch, size_t) { return reinterpret_cast<const char8_t*>(ch); }
+
+template <typename T>
+T xmlGet(const xmlNode* node, const std::string& property, T defaultValue) {
+    xmlChar* str = xmlGetProp(node, reinterpret_cast<const char8_t*>(property.c_str()));
+    const std::string ret{reinterpret_cast<const char*>(str)};
+
+    xmlFree(str);
```
bhennion: Here as well, the data is copied, so the code readability has a performance cost. This is not great. Especially as you could just as easily store the output of `parse()`, call `xmlFree()` and return.

src/core/control/settings/Settings.cpp
```diff
@@ -276,462 +275,385 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
         }
     }
 }
 
 void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     // Parse data map
-    if (!xmlStrcmp(cur->name, reinterpret_cast<const xmlChar*>("data"))) {
-        xmlChar* name = xmlGetProp(cur, reinterpret_cast<const xmlChar*>("name"));
-        if (name == nullptr) {
+    if (u8string(cur->name) == "data"_xmlstr) {
```
bhennion: For instance, if you analyze this line, both sides are copied and then compared. We can hope the compiler optimizes out the RHS copy but the LHS wil remain.

src/core/control/settings/xmlutils.h
tiziodcaio: something like this?
```diff
-    const std::string ret{reinterpret_cast<const char*>(str)};
+    const std::string_view ret = reinterpret_cast<const char*>(str);
```

bhennion:
Code-wise, I'm happy! It doesn't build though, but I guess it is simply a missing include:
```
/Users/runner/work/xournalpp/xournalpp/src/core/control/settings/Settings.cpp:749:66: error: use of undeclared identifier 'XML_PARSE_UNZIP'
```

tiziodcaio:
Fixed!

bhennion:
Thanks!
Still an issue on windows:
```
D:/a/xournalpp/xournalpp/src/core/control/settings/Settings.cpp:1159:40: error: cannot convert 'const std::filesystem::__cxx11::path::value_type*' {aka 'const wchar_t*'} to 'const char*'
 1159 |     xmlSaveFormatFileEnc(filepath.c_str(), doc, "UTF-8", 1);
```
There are also a lot of warnings due to the same thing: fs::path on Windows does not store the path as a utf-8 string but uses wide chars.
Every call to filepath.c_str() should be replaced by `filepath.u8string()` (`.c_str()` if required).

tiziodcaio:
I can't reproduce ARM64 bug at the moment... Any suggestions?

bhennion:
> I can't reproduce ARM64 bug at the moment... Any suggestions?

It has nothing to do with ARM. It is a Windows issue (the x64 Windows would fail as well, if it wasn't cancelled before). The solution is what I said in the [message above](https://github.com/xournalpp/xournalpp/pull/6623#issuecomment-3259050803):

> Every call to filepath.c_str() should be replaced by `filepath.u8string()` (`.c_str()` if required).

tiziodcaio:
I rebased again and I'll test on windows asap

tiziodcaio:
Tested on msys2/mingw64 VM environment. Tests are passing!

tiziodcaio:
Geez I'll try again ASAP

bhennion:
Mmm, that's my fault. Sorry. The standards do not require std::char_traits<unsigned char> to be defined, so compilation with Clang++ fails (this is the compiler that is used for the Windows ARM builds).

I guess we should go back to casting the `const xmlChar*` to `const char*` (or rather `std::string_view` to allow for `opereator==`). This possibly means the ""_xml operator won't be needed after all either...

tiziodcaio:
Rebased!

tiziodcaio:
> Mmm, that's my fault. Sorry. The standards do not require std::char_traits to be defined, so compilation with Clang++ fails (this is the compiler that is used for the Windows ARM builds).

Maybe it was a C++17 problem?
If ARM64 build is still broken, I'll revert some changes :)

tiziodcaio:
colorPaletteSetting was parsed from xml to Settings object as a raw string while it should be a fs::path. In windows paths are stored as wchar_t: that's why it was broken and (maybe) non-deterministic :))

src/core/control/settings/xmlutils.h
```diff
@@ -0,0 +1,105 @@
...
+#include <string_view>
+#include <type_traits>
+
+#include <libxml/tree.h>
+#include <libxml/xmlstring.h>
+
+#include "util/Color.h"
+
+using std::u8string;
+using std::u8string_view;
```
bhennion: As a coding convention, we try not to use such `using` statements in headers.

src/core/control/settings/Settings.cpp
```diff
@@ -1,11 +1,12 @@
...
-#include <cstdlib>      // for atoi
-#include <cstring>      // for strcmp
-#include <exception>    // for exception
+#include <algorithm>  // for max
+#include <cstdint>    // for uint32_t, int32_t
+#include <cstdio>     // for sscanf, size_t
+#include <cstdlib>    // for atoi
+#include <cstring>    // for strcmp
+#include <exception>  // for exception
+#include <list>
```
bhennion: Is this really used?

src/core/control/settings/Settings.cpp
```diff
@@ -277,461 +276,388 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
     for (xmlNodePtr x = cur->children; x != nullptr; x = x->next) {
-        if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("data"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            parseData(x, elem.child(reinterpret_cast<const char*>(name)));
-            xmlFree(name);
-        } else if (!xmlStrcmp(x->name, reinterpret_cast<const xmlChar*>("attribute"))) {
-            xmlChar* name = xmlGetProp(x, reinterpret_cast<const xmlChar*>("name"));
-            xmlChar* value = xmlGetProp(x, reinterpret_cast<const xmlChar*>("value"));
-            xmlChar* type = xmlGetProp(x, reinterpret_cast<const xmlChar*>("type"));
+        const std::u8string_view type = reinterpret_cast<const char8_t*>(x->name);
```
bhennion: Such casts are undefined behaviour in c++20. This is the reason for xoj:util::utf8.
But then, the operator==() won't be defined, so something must be done here (without adding an extra copy). Maybe using `std::ranges::equal`?

bhennion: Actually, as those tags are always on ASCII characters, it may be better to cast to const char* (that's not UB) and compare honest std::string_view's. I don't really know.

src/core/control/settings/xmlutils.h
tiziodcaio: Sadly I need to use `using namespace std::string_view_literals` because of [this](https://en.cppreference.com/w/cpp/string/basic_string_view/operator%2522%2522sv.html)... If you do not like it, I can remove it.

tiziodcaio:
@bhennion can tou enable CI pretty please?

bhennion:
Actually I cannot trigger the CI. Maybe you need to rebase on latest master first.

tiziodcaio:
Done!

tiziodcaio:
Any updates?

tiziodcaio:
All reviews are done... Can we merge?

src/core/control/settings/Settings.cpp
```diff
@@ -277,461 +273,388 @@ auto Settings::getViewModes() const -> const std::vector<ViewMode>& { return thi
...
-            this->colorPaletteSetting = paletteConfig;
-        }
+    } else if (name == "stabilizerAveragingMethod") {
+        this->stabilizerAveragingMethod = static_cast<StrokeStabilizer::AveragingMethod>(parse<int>(value));
+    } else if (name == "stabilizerPreprocessor") {
+        this->stabilizerPreprocessor = static_cast<StrokeStabilizer::Preprocessor>(parse<int>(value));
+    } else if (name == "stabilizerBuffersize") {
+        this->stabilizerBuffersize = parse<unsigned int>(value);
+    } else if (name == "stabilizerSigma") {
+        this->stabilizerSigma = parse<int>(value);
```
bhennion: I found this mistake by trying to open and save my own settings and looking at the diff.
```suggestion
        this->stabilizerSigma = parse<double>(value);
```

I am a bit scared there may be other such mistakes, that would lead to unusable parameters. I feel like I found such an issue every time I tried this PR.

We need to either have a complete test coverage (but that seems complicated), or to enforce that every value is parsed as the same type as it is written, or to find a way to ensure this PR does not introduce a hard-to-find issue.

tiziodcaio:
I understand and I'm sorry because I was not so helpful for testing since I'm new to coding.
Since is not a feature request, I'm not so expert I'm learning and it's frustrating for both touching that part of the codebase I might consider the idea to stale/close the PR at the moment. What do you think?

src/core/control/settings/Settings.cpp
tiziodcaio: I chose to follow your suggestion and partially enforce type-checking. However some edge-cases need to be tuned to the new structure.

Issue #4326, opened 2022-10-16
Empty filenames suggested on Windows


rolandlo:
**Affects versions :**
 - OS: Windows (only)
 - Which version of libgtk do you use: 3.24.34
 - Version of Xournal++: 1.1.1, master (does not matter)
 - Installation method: installation executable, self compiled (does not matter)

**Describe the bug**
When a new file (without PDF background) is saved the suggested filename is empty (under Windows). Users that don't realize that cannot even save the file. It happens with standard "default filename" (namely `%F-Note-%H-%M`) in the "Load/Save" tab in menu `Edit > Preferences`

**To Reproduce**
0. Set the "default filename" to be `%F-Note-%H-%M` (or something else containing `%F`, `%T` or `%V`)
1. Create a new file in Xournal++
2. Press the "Save" button from the toolbar (or use the menu for the same purpose)
3. Observe that the suggested filename is empty.

**Expected behavior**
The rules described near the "default filename" setting should be followed, so e.g. `%F` should be the same as `%Y-%m-%d`, and a valid filename should be suggested.

**Screenshots of Problem**
(Optional) If applicable, add screenshots to help explain your problem.

**Additional context**
The issue that 
> MinGW-w64 does not (yet) support strftime formatting codes like %F, %T or %V (ISO 8601)

 is known. [Upstream bug report](https://sourceforge.net/p/mingw-w64/bugs/793/)

At very least we could replace `%F-Note-%H-%M` by `%Y-%m-%d-Note-%H-%M` under Windows.

PR #6634, opened 2025-08-24
Fix default filename preprocessing for Windows

fusion1110:
Problem:
When creating a new file in Xournal++ on Windows (without a PDF background), the suggested filename is empty if the default filename contains %F, %T, or %V (e.g., %F-Note-%H-%M). This prevents users from saving the file unless they manually enter a name.

The root cause is that MinGW-w64 does not fully support strftime formatting codes like %F, %T, or %V (ISO 8601), which results in an empty string when generating the suggested filename.

Solution:
Replaced unsupported strftime codes with equivalents supported on Windows:

%F → %Y-%m-%d
%T → %H-%M-%S
%V → %U

Added a preprocessing function that converts the default filename format to a Windows-compatible version before generating the suggested filename.
Ensures forbidden characters are replaced and the filename is never empty.

Testing:
Tested on Windows 10 using a freshly compiled Xournal++ build from master.
Verified that new files now have a suggested filename according to the default pattern, e.g., 2025-08-25-Note-15-30.

Additional context:
Beginner contribution: This is my first pull request, and it’s not my last!
I really appreciate the maintainers and all contributors for this amazing app that I use daily.

### DIFF
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -136,6 +136,26 @@ auto Document::createSaveFoldername(const fs::path& lastSavePath) const -> fs::p
     return lastSavePath;
 }
 
+
+static std::string preprocessFormatString(const std::string& formatStr) {
+    std::string processed = formatStr;
+    // Replace %F with %Y-%m-%d
+    size_t pos;
+    while ((pos = processed.find("%F")) != std::string::npos) {
+        processed.replace(pos, 2, "%Y-%m-%d");
+    }
+    // Replace %T with %H-%M-%S
+    while ((pos = processed.find("%T")) != std::string::npos) {
+        processed.replace(pos, 2, "%H-%M-%S");
+    }
+    // Replace %V with %U
+    while ((pos = processed.find("%V")) != std::string::npos) {
+        processed.replace(pos, 2, "%U");
+    }
+    return processed;
+}
+
+
 auto Document::createSaveFilename(DocumentType type, const std::string& defaultSaveName,
                                   const std::string& defaultPdfName) const -> fs::path {
     constexpr static std::wstring_view forbiddenChars = {L"\\/:*?\"<>|"};
@@ -162,6 +182,7 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
 
     auto format_str = wildcardString.empty() ? defaultSaveName : wildcardString;
 
+    format_str = preprocessFormatString(format_str);
     std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
     auto format = converter.from_bytes(format_str);
 


rolandlo:
Thanks for the contribution. The issue you pointed out (#4326) is certainly something that needs to be fixed. It already exists in the release branch. Could you please rebase on the `release-1.2` branch (as follows)?

```term
git rebase --onto=upstream/release-1.2 HEAD~ HEAD
```
Then after force-pushing you can change the target branch of the pull request via the Edit button on the top right of this page in the Github interface.

src/core/model/Document.cpp
```diff
@@ -136,6 +136,26 @@ auto Document::createSaveFoldername(const fs::path& lastSavePath) const -> fs::p
     return lastSavePath;
 }
 
+
+static std::string preprocessFormatString(const std::string& formatStr) {
+    std::string processed = formatStr;
```
rolandlo: Since the string is copied anyway it would make more sense to pass it by value (and continue to work with `formatStr`)

src/core/model/Document.cpp
```diff
@@ -136,6 +136,26 @@ auto Document::createSaveFoldername(const fs::path& lastSavePath) const -> fs::p
...
+        processed.replace(pos, 2, "%Y-%m-%d");
+    }
+    // Replace %T with %H-%M-%S
+    while ((pos = processed.find("%T")) != std::string::npos) {
+        processed.replace(pos, 2, "%H-%M-%S");
+    }
+    // Replace %V with %U
+    while ((pos = processed.find("%V")) != std::string::npos) {
+        processed.replace(pos, 2, "%U");
+    }
```
rolandlo: It seems preferable to keep the variable scope tight:
```suggestion
    while ((size_t pos = processed.find("%F")) != std::string::npos) {
        processed.replace(pos, 2, "%Y-%m-%d");
    }
    // Replace %T with %H-%M-%S
    while ((size_t pos = processed.find("%T")) != std::string::npos) {
        processed.replace(pos, 2, "%H-%M-%S");
    }
    // Replace %V with %U
    while ((size_t pos = processed.find("%V")) != std::string::npos) {
        processed.replace(pos, 2, "%U");
    }
```

fusion1110:
> Thanks for the contribution. The issue you pointed out (#4326) is certainly something that needs to be fixed. It already exists in the release branch. Could you please rebase on the `release-1.2` branch (as follows)?
> 
> ```
> git rebase --onto=upstream/release-1.2 HEAD~ HEAD
> ```
> 
> Then after force-pushing you can change the target branch of the pull request via the Edit button on the top right of this page in the Github interface.

Rebased onto release-1.2, force-pushed, and updated the PR target branch, made the suggested changes.
 Ready for review!

src/core/model/Document.cpp
bhennion: This actually doesn't compile.
It is also inefficient because the string is ran through several times for each loop. Take as an example the string "this is my title - it is a bit long %F". The entire string is screened twice in the first loop because the second call to `std::string::find()` goes back to the beginning of the string. The calls to `std::string::find()` should make use of the last found position so that the beginning of the string is not screened a second time.

For instance, you could do:
```
    auto replace = [&](std::string_view pattern, std::string_view replacement) {
        for (size_t pos = processed.find(pattern); pos != std::string::npos; pos = processed.find(pattern, pos + replacement.length() - pattern.length())) {
            processed.replace(pos, pattern.length(), replacement);
        };
    };
    replace("%F", "%Y-%m-%d");
    replace("%T", "%H-%M-%S");
    replace("%V", "%U");
```

fusion1110: I sincerely apologise for the oversight. I’ve implemented your suggested improvements and tested the updated code, including verifying the build on Windows to ensure everything is functioning as expected

src/core/model/Document.cpp
```diff
@@ -158,6 +176,7 @@ auto Document::createSaveFilename(DocumentType type, const std::string& defaultS
 
     auto format_str = wildcardString.empty() ? defaultSaveName : wildcardString;
 
+    format_str = preprocessFormatString(format_str);
```
bhennion, line 189: One last thing: could you add a comment here, explaining why this is necessary? It's so that future devs don't unknowingly revert this.

fusion1110: Yes will do

bhennion:
Thanks!
Sorry to be a stickler here: the clang formatting CI test is not passing. Could you either run `git clang-format release-1.2` or simply apply 
```
diff --git a/src/core/model/Document.cpp b/src/core/model/Document.cpp
index 8143cf39..322b6b5c 100644
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -151,7 +151,7 @@ static std::string preprocessFormatString(std::string formatStr) {
     auto replace = [&formatStr](std::string_view pattern, std::string_view replacement) {
         for (size_t pos = formatStr.find(pattern); pos != std::string::npos;
              pos = formatStr.find(pattern, pos + replacement.length())) {
-                        formatStr.replace(pos, pattern.length(), replacement);
+            formatStr.replace(pos, pattern.length(), replacement);
         }
     };
```
to apply the code formatting to your patch.

fusion1110:
Thanks for catching that! I’m still pretty new to the project, so I missed the formatting step. I’ll run git clang-format release-1.2 and update the patch.
Lemme know if any further changes are required.

bhennion:
The clang format CI check is still complaining:
```
diff --git a/src/core/model/Document.cpp b/src/core/model/Document.cpp
index 6824b5e3..322b6b5c 100644
--- a/src/core/model/Document.cpp
+++ b/src/core/model/Document.cpp
@@ -150,7 +150,7 @@ auto Document::createSaveFolder(fs::path lastSavePath) -> fs::path {
 static std::string preprocessFormatString(std::string formatStr) {
     auto replace = [&formatStr](std::string_view pattern, std::string_view replacement) {
         for (size_t pos = formatStr.find(pattern); pos != std::string::npos;
-            pos = formatStr.find(pattern, pos + replacement.length())) {
+             pos = formatStr.find(pattern, pos + replacement.length())) {
             formatStr.replace(pos, pattern.length(), replacement);
         }
     };
```

fusion1110:
Formatting corrected per the CI diff. My bad for the back-and-forth, getting better at spotting these.

Issue #3665, opened 2021-12-16
Add more shape recognition options to the shape recognizer.


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
Currently it seems that the shape recognizer only recognizes rectangles, squares, and perfect circles.

**Describe the solution you'd like**
It would be more beneficial if the shape recognizer were able to recognize *all* basic shapes, i.e.
- triangles (right, isoceles, etc)
- eliipses

**Additional Context**
If possible, it would be nice to have a more general shape recognition tool that takes the form of predictive smoothing. For exampe, If I draw some function (parabaloid, exponential, sinusoid, etc), the shape recognizer should be able to smooth out my lines so that the functions follow the shape that they would mathematically follow.

I don't think there is any need for polynomial function recognition beyond a cubic function, but it would be good to support all of the basic functions:
- polynomial up to 3rd order
- sinusoid
- exponential/logarithmic
- etc

bhennion:
See also #2275 and the discussion therein.

lluisalemanypuig:
I agree: it would be nice if triangles (and other shapes) were recognized by the shape recognizer. So I thought maybe I could start with the triangles, and eventually did something that works. Here's a [video](https://youtu.be/7UuRd2_55MQ) that shows the result.

However, I'm not sure I understood the task properly. The rectangles, for example, are created with perfect 90º-angles. Triangles can be idealized (with, say, 60º angles, or detect when a triangle has a specific configuration of angles) but then the recognizer would recognize only so many triangles. The fix I have (for which I do not currently have a PR) creates triangles by ensuring that the three strokes make up a triangle, and uses the segments to create a single stroke, whichever resulting triangle we get. Are there any triangles that should _not_ be recognized?

Also, I think I can generalize what I've done to polygonal strokes. But I'm not sure this would be interesting since the only thing that the recognizer would do is straighten up the strokes.

Issue #2275, opened 2020-09-24
Drawing arrows and triangles with Stroke recognizer 


hyiltiz:
**Is your feature request related to a problem? Please describe.**
In technical illustrations, being able to quickly draw arrows and other shapes is very helpful. Stroke recognizer can help drawing rectangles, lines and circles really well. I am aware that there is a dedicated Draw Arrow tool, but just like the Draw Circle and Draw Rectangle tool, it is a bit inconvenient to use, esp. since they are inside a drop-down menu.

**Describe the solution you'd like**
Drawing from tail to the head and going back sideways in an acute angle while holding Stroke Recognizer should draw straight line from tail to the head with a proper arrowhead. For example, drawing by hand the following line from the left to right then towards top-left a bit:
```
               \
-----------------
```
would result in a leftward arrow →.

**Describe alternatives you've considered**
Draw Arrow tool. Notability app for iPad has a recognizer function that is loved by others around me, but I would like to support Xournal++.

**Additional context**
None.

rolandlo:
The code of the shape recognizer is pretty old and could definitively be improved. However, you may consider using a shortcut for the arrow tool via a plugin, see the [section on plugins](https://xournalpp.github.io/guide/plugins/) on the work-in-progress website. Then you also won't have to use the drop-down menu.

hyiltiz:
Hmm, any plans to refactor/revamp the shape recognizer?

rolandlo:
I personally don't have any plans to work on the shape recognizer in the near future. There are too much other things to work on. If anyone wants to give it a try, just do it. The code for the shape recognizer is isolated enough, so you won't have to know/read much Xournal++ code.

In your title, you ask about triangles getting recognized, too. That is actually possible with the shape recognizer code, but it was disabled, since it apparently didn't quite meet the expectations. Let me point out that you can also draw triangles with the spline tool.

hyiltiz:
Hmm, I checked out the shape recognizer code, and seems like recognizing arrows is also built-in.
https://github.com/xournalpp/xournalpp/blob/307d67f02d00c7c9305cc609736d49e2f54961d3/src/control/shaperecognizer/ShapeRecognizer.cpp#L97

It may also have been disabled, maybe due to poor performance (not computational performance)?

rolandlo:
Yes, I think this was a performance issue.

hyiltiz:
[...]

PR #6641, opened 2025-08-27
Recognize triangles

lluisalemanypuig:
I was wondering why triangles were not recognized and I found issue #3665. I have implemented the recognition, but rather than recognizing a triangle, the code simply straightens the segments when the shape is closed (all endpoints of contiguous segments are close enough). Video of the result [here](https://www.youtube.com/watch?v=7UuRd2_55MQ).

Details of the changes:
- I copied the rectangle detection algorithm for triangles. The only difference is the detection of orientation of the segments: I could not make that work using an orientation test, and used a distance test.
- When a triangle is detected, the strokes are simply straightened up and no specific class of triangle is drawn. That is, the recognizer does not discard, say, non-equilateral triangles, or triangles of other classes. To be clearer: all 3-sided closed shapes are recognized as triangles.
- The algorithm does not change the value of the angles as given by the RecoSegment class (and other functions in ShapeRecognizer class).

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -32,6 +32,77 @@ void ShapeRecognizer::resetRecognizer() {
     this->queueLength = 0;
 }
 
+inline double dist2(const Point& P, const Point& Q) {
+    const double dx = P.x - Q.x;
+    const double dy = P.y - Q.y;
+    return dx * dx + dy * dy;
+}
+
+auto ShapeRecognizer::tryTriangle() -> std::unique_ptr<Stroke> {
+    // first, we need whole strokes to combine to 3 segments...
+    if (this->queueLength < 3) {
+        return nullptr;
+    }
+
+    RecoSegment* rs = &this->queue[as_unsigned(this->queueLength - 3)];
+    if (rs->startpt != 0) {
+        return nullptr;
+    }
+
+    /*
+    Make segments be oriented so that, for every pair of neighbouring segments,
+    the first segment points towards the second. This should make the polygon
+    have all of its segments oriented either clockwise or counter-clockwise.
+
+    The direction of any segment R is defined from P to Q where
+    if R is not reversed then
+        P is (x1,y1)
+        Q is (x2,y2)
+    else
+        P is (x2,y2)
+        Q is (x1,y1)
+    */
+    for (int i = 0; i <= 2; i++) {
+        RecoSegment& r1 = rs[i];
+        const RecoSegment& r2 = rs[(i + 1) % 3];
+
+        const Point P(r1.x1, r1.y1);
+        const Point Q(r1.x2, r1.y2);
+        const Point R(r2.x1, r2.y1);
+        const Point S(r2.x2, r2.y2);
+        const double min_PR_PS = std::min(dist2(P, R), dist2(P, S));
+        const double min_QR_QS = std::min(dist2(Q, R), dist2(Q, S));
+        r1.reversed = min_PR_PS < min_QR_QS;
+    }
+
+    for (int i = 0; i <= 2; i++) {
+        const RecoSegment& r1 = rs[i];
+        const RecoSegment& r2 = rs[(i + 1) % 3];
+
+        const double x1 = r1.reversed ? r1.x1 : r1.x2;
+        const double y1 = r1.reversed ? r1.y1 : r1.y2;
+        const double x2 = r2.reversed ? r2.x2 : r2.x1;
+        const double y2 = r2.reversed ? r2.y2 : r2.y1;
+
+        const double dist = hypot(x1 - x2, y1 - y2);
+        if (dist > TRIANGLE_LINEAR_TOLERANCE * (r1.radius + r2.radius)) {
+            return nullptr;
+        }
+    }
+
+    auto s = std::make_unique<Stroke>();
+    s->applyStyleFrom(this->stroke);
+
+    for (int i = 0; i <= 2; i++) {
+        Point p = rs[i].calcEdgeIsect(&rs[(i + 1) % 3]);
+        s->addPoint(p);
+    }
+
+    s->addPoint(s->getPoint(0));
+
+    return s;
+}
+
 /**
  *  Test if segments form standard shapes
  */
@@ -305,6 +376,10 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinSize) ->
             rs[i].calcSegmentGeometry(stroke->getPoints(), brk[i], brk[i + 1], ss + i);
         }
 
+        if (auto result = tryTriangle(); result != nullptr) {
+            RDEBUG("return triangle");
+            return result;
+        }
         if (auto result = tryRectangle(); result != nullptr) {
             RDEBUG("return rectangle");
             return result;
--- a/src/core/control/shaperecognizer/ShapeRecognizer.h
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.h
@@ -31,7 +31,9 @@ public:
     void resetRecognizer();
 
 private:
+    auto tryTriangle() -> std::unique_ptr<Stroke>;
     auto tryRectangle() -> std::unique_ptr<Stroke>;
+
     // function Stroke* tryArrow(); removed after commit a3f7a251282dcfea8b4de695f28ce52bf2035da2
 
     static void optimizePolygonal(const Point* pt, int nsides, int* breaks, Inertia* ss);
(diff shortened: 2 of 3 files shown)


tmoerschell:
Cool and useful in my opinion. Did you dig in the commit history? There once was a `tryClosedPolygon` method, a long time ago, called for different amount of segments. See 8ff8ce418792695f9d2fc90697bc6a4942294421, bd64a63c4f0ccb2f3497ee1bb999cf2ae78bef77...
Can't really figure why it had been disabled. Would it make sense to accept general quadrilaterals as well?

See also #6169

lluisalemanypuig:
> Did you dig in the commit history? There once was a `tryClosedPolygon` method, a long time ago, called for different amount of segments. See [8ff8ce4](https://github.com/xournalpp/xournalpp/commit/8ff8ce418792695f9d2fc90697bc6a4942294421), [bd64a63](https://github.com/xournalpp/xournalpp/commit/bd64a63c4f0ccb2f3497ee1bb999cf2ae78bef77)...

No, I didn't dig into the commit history but I did read the discussion in the several issues that were opened (see #2275 and discussion therein). As far as I could see, the triangle recognizer was dropped because it was part of a bigger algorithm (I think it was the `tryClosedPolygon`).

> Would it make sense to accept general quadrilaterals as well?

Quadrilaterals recognition is a feature already present in the latest release of xournal++. The difference between the triangle recognizer in this PR and the already-existing quadrilateral recognizer is
- the triangle recognizer only straightens the segments, and leaves the angles unchanged
- the quadrilateral recognizer straightens the segments, and makes the angles be exactly equal to 90º.

lluisalemanypuig:
I thought about implementing a recognizer for _regular_ polygons, but I would like to focus on another feature I have in mind that would be interesting for xournal.

src/core/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -335,20 +410,18 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinSize) ->
...
                 const double num = dy * last.x - dx * last.y + Q.x * P.y - Q.y * P.x;
                 const double num2 = num * num;
                 const double den2 = dy * dy + dx * dx;
                 const double dist2 = num2 / den2;
-
                 if (dist2 < LINE_POINT_DIST2_THRESHOLD) {
                     s->addPoint(P);
                     s->addPoint(Q);
+
                 } else {
```
bhennion: Those spacing changes seems unwarranted. Can you revert them? (this one and the removed new lines a couple of lines above)

lluisalemanypuig:
I wasn't 100% sure what to do to "revert" the changes. So I did a `git revert` and re-applied the changes without any of those changes you mentioned. (Though I think I have accidentally introduced a new line in a header file.)

bhennion:
> I wasn't 100% sure what to do to "revert" the changes. So I did a `git revert` and re-applied the changes without any of those changes you mentioned. (Though I think I have accidentally introduced a new line in a header file.)

I didn't mean anything particular. In the end, your commits will be squashed upon merging anyway. That's all good now!

dexeonify:
Thanks for implementing this! I noticed that the tool seems to recognize triangles more intuitively than circles or rectangles. It's more _forgiving_ with the drawn shape, and I don't have to fully close the path for it to register.

tmoerschell:
I think the recognizer is almost a little _too_ sensitive. Especially in combination with #6622, I regularly have normal text being recognized as triangles. Especially the letter "D" triggers the recognizer quite often.
D_triangle
I always have the stroke recognizer active in order to be able to quickly underline things or make arrows, sketches... Do you think it would make sense not to use the logic of #6622 for triangles (or shapes in general)?

lluisalemanypuig:
I'm sorry that hand-written text is being recognized as triangles. I did not test that, since I do not own a stylus. And yes, seeing your gif I agree that the recognizer is a too sensitive. Unfortunately, I have tried myself to reproduce your gif, but I failed miserable (I blame the mouse).

The logic behind recognizing triangles is:
1. Divide the stroke into segments (I did not write this algorithm)
2. Check if the appropriate pairs of endpoints of the segments are close enough (note, point-to-point distance). If so, then the shape is a triangle. 

I did not know how to do better than this. The sensitiveness of this algorithm depends on the definition of "close enough" for the point-to-point distance check (variable `TRIANGLE_LINEAR_TOLERANCE` in `ShapeRecognizerConfig.h`). Maybe this should be increased to the point where we reach a compromise between not recognizing true positives, and avoiding recognizing false positives.

And yes, a different check would make sense. In rectangle recognition, the check is based on angles (all angles have to be 90º). This idea cannot be applied (I think) to triangles; if done, only a specific class of triangles would be recognized (triangles of degrees 60-60-60, or other). But I do not know of another way of solving triangle recognition.

tmoerschell:
I didn't mean that we'd need to constrain the angles, but maybe the fix from the other PR I mentionned could be disabled for shapes other than simple lines? I think most handwritten letters will have a double stroke somewhere, as does my D. 

I know there's a setting for min stroke length for the recognizer to kick in (which, admittedly, I might have lowered from its default value). Maybe it could be enough to substract one length of the double stroke and re-check for min length?

On another note, I've also found a stroke (originally a "t") which was recognized as a triangle with -nan coordinates. That is obviously something that should not happen.

Diff of stroke representation in unzipped file:
```diff
< <stroke tool="pen" color="#000000ff" width="1.41 0.39551431 0.44422477 0.46369604 0.48211307 0.50610254 0.53876265 0.59573495 0.65896818 0.71594049 0.76385489 0.78713436 0.80346441 0.81736324 0.83091783 0.84410666 0.85383154 0.85662852 0.85731701 0.85488579 0.85002335 0.84376242 0.82883085 0.78922133 0.72461112" capStyle="round">545.60001 178.93587 545.17845 179.25054 544.78512 179.72455 544.25866 180.53944 543.73423 181.82633 543.28644 183.58723 542.94555 185.82012 542.68334 188.42012 542.52399 191.1008 542.47154 193.67659 542.52399 195.85704 542.65711 197.48481 542.81444 198.5357 543.02422 198.82616 543.15533 198.48325 543.23399 197.48481 543.39133 195.98814 543.5769 194.12236 543.89156 192.07302 544.41801 189.97124 545.09978 188.13168 546.01956 186.7399 547.04625 185.95123 548.09714 185.87256 549.09559 186.34456</stroke>
---
> <stroke tool="pen" color="#000000ff" width="0.71198685" capStyle="round">543.9154 182.75837 -nan -nan -nan -nan 543.9154 182.75837</stroke>
```

lluisalemanypuig:
> I didn't mean that we'd need to constrain the angles, but maybe the fix from the other PR I mentionned could be disabled for shapes other than simple lines? I think most handwritten letters will have a double stroke somewhere, as does my D.

The code from #6622 does not play any role in shape recognition other than lines. That is only applied when the number of segments of the stroke is [exactly 1](https://github.com/xournalpp/xournalpp/blob/9ad7b5741ffb2001b626246055c9bdb2a69ed904/src/core/control/shaperecognizer/ShapeRecognizer.cpp#L390). The code for triangles is triggered when there are [exactly 3 segments](https://github.com/xournalpp/xournalpp/blob/9ad7b5741ffb2001b626246055c9bdb2a69ed904/src/core/control/shaperecognizer/ShapeRecognizer.cpp#L379).

> I know there's a setting for min stroke length for the recognizer to kick in (which, admittedly, I might have lowered from its default value). Maybe it could be enough to substract one length of the double stroke and re-check for min length?
>
> On another note, I've also found a stroke (originally a "t") which was recognized as a triangle with -nan coordinates. That is obviously something that should not happen.

Well, this seems to be a bug that I have introduced. But this PR cannot produce NaN since there are no divisions here. This was most likely introduced in #6622 and NaNs are probably produced in [line 422](https://github.com/xournalpp/xournalpp/blob/9ad7b5741ffb2001b626246055c9bdb2a69ed904/src/core/control/shaperecognizer/ShapeRecognizer.cpp#L422) of the shape recognizer for a straight line. This I believe should have an easy fix if I manage to reproduce it. Would you mind opening a new issue so that I can reference it in the new PR?

> Diff of stroke representation in unzipped file:
> 
> ```diff
> < <stroke tool="pen" color="#000000ff" width="1.41 0.39551431 0.44422477 0.46369604 0.48211307 0.50610254 0.53876265 0.59573495 0.65896818 0.71594049 0.76385489 0.78713436 0.80346441 0.81736324 0.83091783 0.84410666 0.85383154 0.85662852 0.85731701 0.85488579 0.85002335 0.84376242 0.82883085 0.78922133 0.72461112" capStyle="round">545.60001 178.93587 545.17845 179.25054 544.78512 179.72455 544.25866 180.53944 543.73423 181.82633 543.28644 183.58723 542.94555 185.82012 542.68334 188.42012 542.52399 191.1008 542.47154 193.67659 542.52399 195.85704 542.65711 197.48481 542.81444 198.5357 543.02422 198.82616 543.15533 198.48325 543.23399 197.48481 543.39133 195.98814 543.5769 194.12236 543.89156 192.07302 544.41801 189.97124 545.09978 188.13168 546.01956 186.7399 547.04625 185.95123 548.09714 185.87256 549.09559 186.34456</stroke>
> ---
> > <stroke tool="pen" color="#000000ff" width="0.71198685" capStyle="round">543.9154 182.75837 -nan -nan -nan -nan 543.9154 182.75837</stroke>
> ```

Issue #3452, opened 2021-10-11
State that xournalpp --create-pdf $filename replaces the source file


AlecsFerra:
Please state explicitly that this operation replaces the source file, and if possible make the default behavior create a new file.

AlecsFerra:
It's a bug not a feature request.

Technius:
I can confirm that is the case. If I try to write to the background pdf path, the background PDF overwritten with the exported file (without the background).
```
$ xournalpp /tmp/test.xopp -p /tmp/test.pdf

(xournalpp:3970): Poppler-CRITICAL **: 11:33:43.326: PopplerPage* poppler_document_get_page(PopplerDocument*, int): assertion '0 <= index && index < poppler_document_get_n_pages(document)' failed

(xournalpp:3970): GLib-GObject-CRITICAL **: 11:33:43.326: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

(xournalpp:3970): Poppler-CRITICAL **: 11:33:43.326: void poppler_page_render_for_printing(PopplerPage*, cairo_t*): assertion 'POPPLER_IS_PAGE(page)' failed
** Message: 11:33:43.330: PDF file successfully created
```

The fix should be to display an error message and halt if the file to write to is the same as the background PDF file.

fusion1110:
# Attempting to export over the same background PDF
./build/xournalpp test.xopp -p test.pdf

# Output:
# (com.github.xournalpp.xournalpp:344658): xopp-ERROR **: Do not overwrite the background PDF! This will cause errors!
# zsh: trace trap (core dumped)

# Exporting to a different file works fine
./build/xournalpp test.xopp -p output.pdf

# Output:
# xopp-Message: PDF file successfully created
# Exit code: 0

is this the expected behaviour?

fusion1110:
Attempting to export over the same background PDF
./build/xournalpp test.xopp -p test.pdf

Output:
(com.github.xournalpp.xournalpp:344658): xopp-ERROR **: Do not overwrite the background PDF! This will cause errors!
zsh: trace trap (core dumped)

Exporting to a different file works fine
./build/xournalpp test.xopp -p output.pdf
Output:
xopp-Message: PDF file successfully created
Exit code: 0

this is the expected behaviour right?

PR #6668, opened 2025-09-07
Fix: Prevent export from overwriting background PDF [Issue #3452]

fusion1110:
Added a check before export:

If the export path is the same as the background PDF, the export is aborted with an error message. 
This prevents overwriting the background PDF and ensures the export file contains all content.

Example:
./build/xournalpp test.xopp -p test.pdf           # _triggers error_
./build/xournalpp test.xopp -p output.pdf    #  _works fine_


src/core/control/ExportHelper.cpp
```diff
@@ -100,6 +100,20 @@ auto exportPdf(Document* doc, const char* output, const char* range, const char*
...
+    try {
+        if (!backgroundPDF.empty() && fs::exists(backgroundPDF)) {
+            if (fs::weakly_canonical(path) == fs::weakly_canonical(backgroundPDF)) {
+                g_error("Do not overwrite the background PDF! This will cause errors!");
+                return -3;  // Return error code for export failure
+            }
+        }
+    } catch (const fs::filesystem_error& fe) {
+        g_warning("The check for overwriting the background failed with: %s", fe.what());
+        // Continue with export since this is just a filesystem error in the check
```
bhennion: I think a filesystem error should stop the export process. I can't see any situation where the export could actually proceed and be successful. The most likely scenarii are
1. the user doesn't have the read access to the background pdf or
2. there was some hardward IO issue.

Either way, we should stop right now.

fusion1110:
Thanks for the review! That makes sense — f the filesystem check fails, proceeding with the export could indeed lead to unpredictable results. I'll update the code so that in case of a filesystem_error, we abort the export instead of continuing. Would you prefer I return a distinct error code for that case, or reuse the same one as for the overwrite check?

bhennion:
> Would you prefer I return a distinct error code for that case, or reuse the same one as for the overwrite check?

As you like!

fusion1110:
Sorry I messed it up.
Creating a new PR.

PR #6670, opened 2025-09-09
Refactor LoadHandler: take 2

tmoerschell:
This is a partial rewrite of the load handler, as had been suggested some long time ago should be necessary in #937.

This would close #5183. There are also many smaller issues with memory safety that are addressed with this. Ideally, it should be resilient against any ill-formed files.

This pull request replaces #5554, which I am leaving for comparison purposes instead of doing a force push. The most important change is that this version sticks with the currently used `GMarkupParser`, which allows streaming files while maintaining performance.

### Commits

The first three commits are preparatory changes. The main refactoring is in the fourth commit, and the last commit is a move away from `g_ascii_strtod` for the faster `std::from_chars`. I am not sure how good support for that feature is yet, but I could eventually squash it.

### Things I would like input on

* `std::from_chars` sounds awesome and offers an impressive speedup on double parsing speed (x5 vs `g_ascii_strtod`). I've seen there was limited support for it in libc++ until recently with the 20.1.0 version. The last commit should compile even if support for floating point `std::from_chars` is missing. Do you think this is the right way forward?
* If you look at `xml_tags` and "named enum" classes like `StrokeCapStyle` or `BackgroundType`, the structure is actually somewhat similar. Should this be uniformized, and do you have a preference for one or the other style? Those "named enums" are really nice for parsing, but I haven't found a good way to deduplicate some of the code (implicit conversion to underlying value and constructor). We could use `EnumIndexArray` instead of plain `std::array` for those as well, but that would again add complexity elsewhere (need to ignore/assert the `ENUMERATOR_COUNT` case in several places).
* As mentionned [here](https://github.com/xournalpp/xournalpp/pull/5554#issuecomment-3250972318), `GMarkupParser` could in theory split up contiguous text into several parts, even though the library's current implementation will not do that. The old `LoadHandler` takes advantage of that, and I have followed the same approach for now. Should I try to implement a "safe path" nevertheless? I believe it would add quite some complexity (as in convoluted code) for parsing stroke text (the coordinates).
* What do you think of the custom non-fatal error reporting? It currently only covers things that are reported from the `LoadHandler` class.
* I tried to accept unknown tags in different places for forward-compatibility, but is it worth the trouble?

### Benchmarking

[...]

PR #6676, opened 2025-09-10
C++20, again

bhennion:
This is a rerun of @Febbe's #5620.
@Febbe, I reused one of your commits, I hope you're ok with that?

So this PR moves from c++17 to c++20.

There are a couple of differences with #5620:

1. `fs::u8path()` is no longer used, replaced with the equivalent `fs::path(const char8_t*)`. This PR does not introduce any new warning.
2. I added in `util/PathUtil.cpp` a class `GFilename` which holds paths encoded in the [glib filename encoding](https://docs.gtk.org/glib/character-set.html#file-name-encodings) while minimizing the number of copies of the string (in particular if the glib filename encoding is uft8 -- ie most of the time).
    This class is used in a couple of places, but there are still a lot of places where it should be used. This is out of scope of this PR. (see also #6137)

### DIFF
--- a/src/core/control/settings/Settings.cpp
+++ b/src/core/control/settings/Settings.cpp
@@ -410,11 +410,11 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("selectedToolbar")) == 0) {
         this->selectedToolbar = reinterpret_cast<const char*>(value);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastSavePath")) == 0) {
-        this->lastSavePath = fs::path(xoj::util::utf8(reinterpret_cast<const char*>(value)));
+        this->lastSavePath = fs::path(xoj::util::utf8(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastOpenPath")) == 0) {
-        this->lastOpenPath = fs::path(xoj::util::utf8(reinterpret_cast<const char*>(value)));
+        this->lastOpenPath = fs::path(xoj::util::utf8(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastImagePath")) == 0) {
-        this->lastImagePath = fs::path(xoj::util::utf8(reinterpret_cast<const char*>(value)));
+        this->lastImagePath = fs::path(xoj::util::utf8(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("edgePanSpeed")) == 0) {
         this->edgePanSpeed = tempg_ascii_strtod(reinterpret_cast<const char*>(value), nullptr);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("edgePanMaxMult")) == 0) {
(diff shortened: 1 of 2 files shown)


src/core/control/settings/Settings.cpp
```diff
@@ -409,11 +409,11 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
...
         this->selectedToolbar = reinterpret_cast<const char*>(value);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastSavePath")) == 0) {
-        this->lastSavePath = fs::u8path(reinterpret_cast<const char*>(value));
+        this->lastSavePath = fs::path(reinterpret_cast<const char8_t*>(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastOpenPath")) == 0) {
-        this->lastOpenPath = fs::u8path(reinterpret_cast<const char*>(value));
+        this->lastOpenPath = fs::path(reinterpret_cast<const char8_t*>(value));
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("lastImagePath")) == 0) {
-        this->lastImagePath = fs::u8path(reinterpret_cast<const char*>(value));
+        this->lastImagePath = fs::path(reinterpret_cast<const char8_t*>(value));
```
Febbe: Undefined behavior, use utf8_view instead.

src/core/control/settings/Settings.cpp
```diff
@@ -496,9 +496,9 @@ void Settings::parseItem(xmlDocPtr doc, xmlNodePtr cur) {
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("useStockIcons")) == 0) {
         this->useStockIcons = xmlStrcmp(value, reinterpret_cast<const xmlChar*>("true")) == 0;
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("defaultSaveName")) == 0) {
-        this->defaultSaveName = reinterpret_cast<const char*>(value);
+        this->defaultSaveName = reinterpret_cast<const char8_t*>(value);
     } else if (xmlStrcmp(name, reinterpret_cast<const xmlChar*>("defaultPdfExportName")) == 0) {
-        this->defaultPdfExportName = reinterpret_cast<const char*>(value);
+        this->defaultPdfExportName = reinterpret_cast<const char8_t*>(value);
```
Febbe: Undefined behavior, use utf8_view instead.
Won't repeat it after this review commit (it happens very often).

It's not UB, if the cast is done on link time, e.g. a C function compiled with an RT of unsigned char* but included with char8_t *, since there, aliasing rules don't apply.

src/core/model/BackgroundImage.h
```diff
@@ -20,16 +20,15 @@
 #include "filesystem.h"  // for path
 
 struct BackgroundImage {
-    BackgroundImage();
-    BackgroundImage(const BackgroundImage& img);
-    BackgroundImage(BackgroundImage&& img) noexcept;
-    ~BackgroundImage();
-
+    BackgroundImage() = default;
```
Febbe: All are defaulted and should be removable.

src/util/PathUtil.cpp
```diff
@@ -179,29 +181,80 @@ auto Util::clearExtensions(fs::path& path, const std::string& ext) -> void {
...
+                g_filename_from_utf8(char_cast(p.u8string().c_str()), -1, nullptr, nullptr, &err));
+        if (err) {
+            g_warning("Failed to convert g_filename from utf8 with error code: %d\n%s", err->code, err->message);
+            g_error_free(err);
+            value = u8"";
+        }
+    } else {
+        value = p.u8string();
+    }
+}
```
Febbe: You can use an #ifdef for windows. gfilename is defined to be utf8 there, no runtime checks required.

Febbe:
Looks good so far, I saw, that I had a commit in my old approach to remove some UB have you also done the change that caused this. If yes, the commit to remove the UB should also be included.

Another thing I noted is, that you are reinterpreting a uint8_t* to a char* just to fill it in `util::utf8` May it be possible to allow any one byte char type including std::byte in utf8_view? If not/too complicated, leave it.

Febbe (approved):

bhennion:
> Looks good so far, I saw, that I had a commit in my old approach to remove some UB have you also done the change that caused this. If yes, the commit to remove the UB should also be included.

Yes, that commit has been squashed somewhere in the process

> Another thing I noted is, that you are reinterpreting a uint8_t* to a char* just to fill it in `util::utf8` May it be possible to allow any one byte char type including std::byte in utf8_view? If not/too complicated, leave it.

I implemented a simple template for single byte types that are convertible to char8_t. This removes the casts from xmlChar* (a.k.a. unsigned char*) to char*.
It does not work for std::byte though because std::byte is not convertible to char8_t. One would need to use std::to_integer then. I'd prefer not to go there as we don't need it for now.

bhennion:
@rolandlo Would merging this be ok?
It will most likely cause mild conflicts with #6623 (@tiziodcaio), but some of the c++20 features will certainly be of use (typically the range and format libraries) -- see a couple of comments I made in #6670.

tiziodcaio:
Fine :)

rolandlo:
> @rolandlo Would merging this be ok? 

I think so. It was reviewed by @Febbe, so if it compiles on all supported platforms, I'm fine with it.

bhennion:
Done! Thanks @Febbe for the original work and the review!

PR #6685, opened 2025-09-13
Fix drawing of dashed strokes while writing

### LINKED ISSUE #6680: Dashed lines are drawn incorrectly during stylus drawing
### Operating System

Linux

### (Linux only) Distribution

openSUSE Tumbleweed

KDE Plasma

Wayland

### Installation Method

Build from source

### Xournal++ Version

master, since 96b49d2b89734a3cf0c51844fed1b2d280feb0e7

### libgtk Version

3.24.50

### Bug Description

When drawing a dashed (or dash-dotted, dotted) line using the stylus, dash distance depends on stylus speed. Dashes are not visible at slow speeds, and can only get seen when moving the stylus fast. When releasing the stoke, the dashes get drawn correctly.

### Expected Behaviour

Dashes are drawn directly in place, as it happens with the mouse, or on release builds.

### Steps to Reproduce

https://github.com/user-attachments/assets/361a7ffb-9882-4614-9476-9ab9a2f604b4

### Additional Context

_No response_

bhennion:
Fix #6680 (only affects maser)

### DIFF
--- a/src/core/model/StrokeContour.cpp
+++ b/src/core/model/StrokeContour.cpp
@@ -1,10 +1,12 @@
 #include "StrokeContour.h"
 
 #include <cmath>
+#include <numeric>
 
 #include "model/MathVect.h"
 #include "model/Point.h"
 #include "util/Assert.h"
+#include "util/safe_casts.h"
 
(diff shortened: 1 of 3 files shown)


bhennion:
Thanks for the review!

Issue #3695, opened 2022-01-04
Request: Add drag and drop audio file to existing object 


Hunanbean:
**Is your feature request related to a problem? Please describe.**
 I'm always frustrated when i press recoed to make a voice note and it gets lost to the ether because i was not 'making' an object at the exact time.

**Describe the solution you'd like**
Please add drag and drop audio file attachment to existing object (If already present, it is not working on Manjaro)

Thank you very much, I really appreciate the software!

rolandlo:
Do I understand correctly: You want to drag an audio file from the file manager of your system onto an existing stroke in Xournal++ in order to attach it to that stroke? That sounds like an interesting suggestion to me and it shouldn't be too hard to implement it (using the [`drag-data-received` signal](https://docs.gtk.org/gtk3/signal.Widget.drag-data-received.html) of `Gtk`).

Hunanbean:
That is correct.
Thank you

PR #6690, opened 2025-09-18
Added functionality to drop .mp3, .wav, and .ogg files onto a selected object

CarelessNobody:
Tried to implement #3695

### DIFF
--- a/src/core/control/tools/EditSelection.cpp
+++ b/src/core/control/tools/EditSelection.cpp
@@ -39,6 +39,8 @@
 
 #include "EditSelectionContents.h"  // for EditSelectionConte...
 
+#include "filesystem.h"  // for path
+
 class XojFont;
 
 using std::vector;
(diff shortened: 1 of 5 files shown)


src/core/gui/MainWindow.cpp
```diff
@@ -369,6 +370,20 @@ void MainWindow::dragDataRecived(GtkWidget* widget, GdkDragContext* dragContext,
         for (int i = 0; uris[i] != nullptr && i < 3; i++) {
             const char* uri = uris[i];
 
+            //If the URI is a mp3 file and there is a valid, selected object, connect them
```
rolandlo:
```suggestion
            // If the URI is an audio file and there is a valid, selected object, connect them
```

src/core/control/tools/EditSelection.cpp
```diff
@@ -417,6 +419,10 @@ auto EditSelection::setFill(int alphaPen, int alphaHighligther) -> UndoActionPtr
     return this->contents->setFill(alphaPen, alphaHighligther);
 }
 
+void EditSelection::setAudioFilename(const fs::path& audioFile, unsigned int timestamp) {
+    this->contents->setAudioFilename(audioFile, timestamp);
+}
```
rolandlo, line 424: I'm not so happy with the naming, as setting the audio filename is not all this method does. Also it is only applied for timestamp 0, so I'm not sure the timestamp should be an argument of this method. If it ever needed it could be refactored later.

rolandlo (changes requested):
Thanks for your contribution. The feature works fine. Timestamps for audio elements getting assigned audio like this will always be zero (so the audio starts playing at the start for all elements in the selection), but that's to be expected.

Please apply clang format.

PR #6719, opened 2025-10-01
Improved save speed

Hicortab:
The document was being written entirely, regardless of whether pages were written or not. On a complex 150-page document (with lots of images, LaTeX code, etc.), it became very, very slow. This new solution leaves the unchanged pages and writes only the modified ones. Please test this feature and tell me what I could improve. :)

### DIFF
--- a/src/core/control/jobs/SaveJob.cpp
+++ b/src/core/control/jobs/SaveJob.cpp
@@ -114,15 +114,74 @@ std::string extractXmlFromXopp(const fs::path& filepath, fs::path tempDir) {
     return xml;
 }
 
+fs::path createTempDir() {
+    std::random_device rd;
+    std::mt19937 gen(rd());
+    std::uniform_int_distribution<int> dist(0, 9999);
+    int code = dist(gen);
+
+    fs::path tempDir = fs::temp_directory_path() / ("Xournal-" + std::to_string(code));
+    
+    return tempDir;
+}
+
+void saveFinalFile(const std::string& modifiedXMLStr, const std::string& originalXMLStr, 
+                   const fs::path& target, auto& lastError)
+{
+    pugi::xml_document doc1, doc2;
+
+    if (!doc1.load_string(modifiedXMLStr.c_str())) {
+        lastError = FS(_F("Error saving file, could not parse modified XML"));
+        return;
+    }
+
+    if (!doc2.load_string(originalXMLStr.c_str())) {
+        lastError = FS(_F("Error saving file, could not parse original XML"));
+        return;
+    }
+    
+    pugi::xml_document resultDoc;
+    pugi::xml_node root = resultDoc.append_copy(doc2.document_element());
+
+    int pageIndex = 0;
+    
+    std::vector<std::pair<pugi::xml_node, pugi::xml_node>> replacements;
+    
+    for (pugi::xml_node page : root.children("page")) {
+        if (std::find(UndoRedoHandler::pagesChanged.begin(), 
+                     UndoRedoHandler::pagesChanged.end(), 
+                     pageIndex) != UndoRedoHandler::pagesChanged.end()) {
+            
+            pugi::xpath_node pageDoc1 = doc1.select_node(
+                ("/xournal/page[" + std::to_string(pageIndex + 1) + "]").c_str());
+            
+            if (pageDoc1) {
+                replacements.push_back({page, pageDoc1.node()});
+            }
+        }
+        pageIndex++;
+    }
+    
+    for (auto& [oldPage, newPage] : replacements) {
+        root.insert_copy_before(newPage, oldPage);
+        root.remove_child(oldPage);
+    }
+
+    std::stringstream ss;
+    resultDoc.save(ss, "  ");
+    std::string mergedXml = ss.str();
+    
+    GzOutputStream out(target);
+    out.write(mergedXml.c_str(), mergedXml.length());
+    out.close();
+}
+
 auto SaveJob::save() -> bool {
     updatePreview(control);
     Document* doc = this->control->getDocument();
     SaveHandler h;
 
     long totalTime = 0;
-
-    auto startTotal = std::chrono::high_resolution_clock::now();
-    auto start = std::chrono::high_resolution_clock::now();
 
     doc->lock();
     fs::path target = doc->getFilepath();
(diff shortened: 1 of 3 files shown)


bhennion:
Thanks for the PR! Making those long files faster to save would certainly be a great thing to have!
There are however some things to be careful about:

1. How files are saved is a very critical operation, so those changes need to be discussed. In particular relying on an xml library here deserves discussing with @tmoerschell -- see #5554 and #6670 -- about file loading. The key point here being that we have no direct control over what the library outputs in the xml file, so we need to make sure it is always compatible with the subset of xml our parser can parse.
2. Because it is a very critical operation, test cases need to be added to our test suite: open a document, modify one or several pages, save the document to the same location/to a different location, and check that the resulting file is as expected (by reading it and deep-comparing the two Document instances).
3. Benchmarks could also be added: I suspect your solution has a cost for small to medium documents (e.g. a very dense single page document). It would be good to measure that so a informed decision can be made.
4. Changes should certainly not be in control/job/SaveJob but rather in control/xojfile/SaveHandler. For instance, your patch is not used when files are "autosaved".

About 1.: from what I remember, the issue with relying on a xml library for loading the files was performance loss. I don't recall if `pugixml` having specifically been benchmarked though.

Finally, from what I could gather skimming through your code, the file is rewritten entirely anyway. However, the conversion of a page `binary` -> `xml` is only done when the page has been modified. We could possibly fetch those `xml` pages from the previously saved docment by hand to get the same kind of optimization without relying on the extra library. It would probably be more work though.

@tmoerschell do you have an opinion/input on all that? Had you tried `pugixml` back when you benchmarked xml libs?

tmoerschell:
I have read about `pugixml`, but did not try it out because it requires building the whole DOM tree at once, and I wanted to keep a streaming parser. However, the approach used here surprises me. A big part of the cost paid for loading a file is the actual XML parsing. With `GMarkup`, it's about 22% from one file I've tried. Even if `pugixml` is fast, parsing the existing file doesn't come for free. Gzip compression/decompression should not be forgotten either. For the same file, decompression was 16% of the loading time. The XML library generating valid/identical output is the least of my worries here, though it should of course be tested.

The cost of saving a file is probably mostly formatting doubles and base64 encoding if the file contains a lot of images/LaTeX. The rest is mostly mapping the internal document representation to constant strings, which I do not think takes much longer than what `pugixml` needs to save the final document. In fact, the code in `SaveHandler.cpp` and `control/xml` is quite "simple" (as in "of low complexity"). Speed improvements could be gained from:
1. Using `std::to_chars` for formatting numbers
2. Keeping images as base64 instead of binary representation, though this would imply a larger document lifetime memory footprint.
3. Checking the current implementation for any bottlenecks by using tools like callgrind. It might perform some unnecessary copies but I'm unsure if those really matter.
4. The save handler currently creates the whole DOM tree before saving. This allows a satisfying call to `root->writeOut(out, listener);`, but writing the file while traversing the document might be more cache- and memory-friendly. Beware that this could be a rather big undertaking for a rather small gain (maybe 5%, at most the time it takes from the moment you trigger the save until the progress bar starts). Writing this, I realize that we might not need to lock the document during the call to `saveTo()`, like we do for autosave. https://github.com/xournalpp/xournalpp/blob/9ad7b5741ffb2001b626246055c9bdb2a69ed904/src/core/control/jobs/SaveJob.cpp#L132-L135

With the proposed solution, we may remove much of the encoding step, but we instead:
* Store the modified pages to the disk (including compression)
* Decompress and parse the original + the modified pages
* Rewrite and recompress the entire document

I think it may be possible to obtain gains using this approach, but for it to make sense, you should combine the unmodified pages with the freshly encoded modified pages, without doing a save/load cycle of the modifications to the disk. Further, though I do not know how much that would help, you could parse only down to the page tag and store (or ignore for modified ones) the whole page XML string (even without unescaping).

Most importantly however, I've tried out the proposed changes and:
* Saving a new file or "save as" does not work.
* With no modifications, a long file seems to take about 1.5 times longer to save (rough estimation, I have not set up a benchmark).
* When performing a modification, the save might be slightly faster, but **the modifications are not always saved**. `doc1.select_node("/xournal/page[<pageIndex + 1>]")` will not work as the indices of the modified pages are different.
* The application crashes when opening a new document, inserting two strokes and performing undo twice:
```
std::deque<_Tp, _Alloc>::reference std::deque<_Tp, _Alloc>::back() [with _Tp = int; _Alloc = std::allocator<int>; reference = int&]: Assertion '!this->empty()' failed.

#7  0x0000000000978793 in std::deque<int, std::allocator<int> >::back() at /home/Thomas/dev/xournalpp/build/xournalpp
#8  0x0000000000977207 in UndoRedoHandler::undo() at /home/Thomas/dev/xournalpp/build/xournalpp
```

Also, from a quick glance:
* Temporary directories are created without checking for collision, and are not removed afterwards.
* I do not think `control->getCurrentPageNo()` is sufficient for determining the page(s) a change actually affects. Consider for example replacing the background on all pages.
* There is no consideration for what should happen if we save the file multiple times.
* How do we ensure that the file that is on the disk is really the same that we loaded initially? If say another instance of Xournal++ is writing to it concurrently, some assumptions break. That could be solved for example by computing a hash of the originally loaded file and comparing it before saving.

tmoerschell:
I've analyzed two files more deeply and I've found the following:

Analysis lecture notes (contains 1-2 images / page):
- `cairo_surface_write_to_png_stream`: 63%, of which
  - `pngWriteFunction`: 5% (which is itself 89% gzip compression and 7% base64 encoding)
  - actual PNG image encoding: the other 95%
- `XmlPointNode::writeOut(OutputStream*)`: 37%, of which
  - 48% `OutputStream::write(char const*)`
  - 46% `g_ascii_formatd`
 
Many strokes:
- `OutputStream::write(char const*)`: 53%
- `g_ascii_formatd`: 43%

This confirms that writing images takes a lot of time. Base64 conversion not so much. The images have a data attribute, which should contain the original image data (there's some things about backwards compatibility for only saving PNGs though I believe). Using that instead of reencoding the images would be a significant point of optimization. 

Using `std::to_chars` could also provide up to a ~30% speedup for mostly-strokes files (I've benched it to be 3.6 times faster than `g_ascii_fromatd`).

And I found a random string copy here that amounts to a mere 0.2% of the total save time*: https://github.com/xournalpp/xournalpp/blob/9ad7b5741ffb2001b626246055c9bdb2a69ed904/src/core/control/xml/TextAttribute.cpp#L15

\* Numbers were calculated from callgrind instruction fetches, which isn't directly processing time, but a good indicator.

rolandlo:
Closing in favor of #6750

PR #6728, opened 2025-10-04
Properly document Chrome OS installation options

archisman-panigrahi:
Add information about Debian package as well as Android app

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,14 @@ The latest stable release is available in the main repository:
 sudo eopkg it xournalpp
 ```
 
+
+### Chrome OS
+
+You can [enable the linux container](https://support.google.com/chromebook/answer/9145439?hl=en), and install the [Debian package](https://github.com/xournalpp/xournalpp?tab=readme-ov-file#debian) from the official Debian repositories.
+```sh
+sudo apt install xournalpp
+```
+
 ### Flatpak
 
 The Xournal++ team officially supports a [FlatHub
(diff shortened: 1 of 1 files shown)


bhennion:
Thanks for the PR!

Two things: 

1. I'm not certain how much we want to document running on Chrome OS. We do not support it directly and we do not have any means of testing the releases. Maybe adding a "This is not officially supported by the development team" or something of the sort would be good.
2. I think we should remove entirely the mention to the android app: the app is no longer maintained or distributed.

@rolandlo an opinion on the matter?

bhennion:
Also, out of curiosity: it works without any problems on Chrome Os' linux container? Having the devices to work properly wasn't a hasle?

archisman-panigrahi:
I have a tablet running ChromeOS (this - https://www.gsmarena.com/acer_chromebook_tab_10-9139.php). xournal++ Debian version works perfectly with the stylus. I have not tried the android app in ChromeOS

archisman-panigrahi:
> Also, out of curiosity: it works without any problems on Chrome Os' linux container? 

I have two tablets (with stylus) running ChromeOS (https://www.gsmarena.com/acer_chromebook_tab_10-9139.php and https://www.asus.com/us/mobile-handhelds/tablets/chromebook-tablet/asus-chromebook-tablet-ct100/). Xournal++ Debian version works perfectly with the stylus. I have not tried the android app in ChromeOS

> Having the devices to work properly wasn't a hasle?

1. Enable linux container in the settings (once and for all).
2. Open a linux terminal and do `sudo apt install xournalpp` (alternatively download the .deb file and install -- chromeOS will offer you to install and integrate into the system). 
3. Xournalpp gets loaded in the menu and behaves as any other chromeOS app.

I would not call that hassle (there is no need to play with any drivers).

In the system settings, one can share the home folder of chromeOS with the linux container, and afterward xournal++ can directly open and save/export files with chromeOS home folder/downloads folder.

rolandlo:
> Thanks for the PR!
> 
> Two things:
> 
>     1. I'm not certain how much we want to document running on Chrome OS. We do not support it directly and we do not have any means of testing the releases. Maybe adding a "This is not officially supported by the development team" or something of the sort would be good.
> 
>     2. I think we should remove entirely the mention to the android app: the app is no longer maintained or distributed.
> 
> 
> @rolandlo an opinion on the matter?

I agree about not mentioning the Android app any more for the reasons mentioned.

About ChromeOS: Actually I have a FydeTab Duo on which ChromeOS based FydeOS runs by default and I can confirm that Xournal++ behaves fine in the Linux container and is nicely integrated into the desktop interface. So I could actually test new releases there.

archisman-panigrahi:
I removed the reference to the Android app for ChromeOS, and also moved it to the end of other distros (away from Android).

Ready from my side

rolandlo (approved):
LGTM

rolandlo:
Thanks @archisman-panigrahi for the update.

Issue #6717, opened 2025-09-30
Stroke recognizer produces NaN values


tmoerschell:
As mentionned in https://github.com/xournalpp/xournalpp/issues/6641#issuecomment-3352885851_ , it can happen that the stroke recognizer produces NaN coordinates.

Here's a file containing only the stroke that caused the issue. It should be easier to reproduce it from there.

```xml
<?xml version="1.0" standalone="no"?>
<xournal creator="tmoerschell" fileversion="4">
<title>Xournal++ document - see https://xournalpp.github.io/</title>
<page width="595.27559" height="841.88976">
<background type="solid" color="#ffffffff" style="plain"/>
<layer>
<stroke tool="pen" color="#000000ff" width="1.41 0.39551431 0.44422477 0.46369604 0.48211307 0.50610254 0.53876265 0.59573495 0.65896818 0.71594049 0.76385489 0.78713436 0.80346441 0.81736324 0.83091783 0.84410666 0.85383154 0.85662852 0.85731701 0.85488579 0.85002335 0.84376242 0.82883085 0.78922133 0.72461112" capStyle="round">545.60001 178.93587 545.17845 179.25054 544.78512 179.72455 544.25866 180.53944 543.73423 181.82633 543.28644 183.58723 542.94555 185.82012 542.68334 188.42012 542.52399 191.1008 542.47154 193.67659 542.52399 195.85704 542.65711 197.48481 542.81444 198.5357 543.02422 198.82616 543.15533 198.48325 543.23399 197.48481 543.39133 195.98814 543.5769 194.12236 543.89156 192.07302 544.41801 189.97124 545.09978 188.13168 546.01956 186.7399 547.04625 185.95123 548.09714 185.87256 549.09559 186.34456</stroke>
</layer>
</page>
</xournal>
```

lluisalemanypuig:
Are there any extra steps that have to be taken after loading the file? I've just updated the master branch to the latest commit (at the time of writing) and it does not crash for me.

tmoerschell:
You need to let the stroke recognizer run on that stroke to see it happening. There should be no crash, only abnormal results from the recognizer

lluisalemanypuig:
I cannot get the stroke recognizer to run on the stroke you gave. Could you provide some guidance on how to do that?

tmoerschell:
It's not designed to, but I imagine it's possible to hack it together for debugging. I'll see if I can help this evening (CET) if you get stuck.

[...]

PR #6737, opened 2025-10-09
Do not process segments that contain NaN values (#6717)

lluisalemanypuig:
This PR is to fix a bug discussed in issue #6717, which was first put forward in #6641.

The bug lies in the shape recognizer algorithm, which first breaks up a stroke into segments and then does its best to classify those segments. But sometimes the segments contain NaN values, and these were not handled properly. This fix detects when segments contain NaN values and avoids running the triangle, rectangle and line recognizers if they have them.

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -370,68 +370,74 @@ auto ShapeRecognizer::recognizePatterns(Stroke* stroke, double strokeMinSize) ->
         RecoSegment* rs = &this->queue[as_unsigned(this->queueLength)];
         this->queueLength += n;
 
+        bool is_bad = false;
         for (int i = 0; i < n; i++) {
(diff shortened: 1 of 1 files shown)


tmoerschell:
I have found another way to fix the issue further up. Your approach is of course still valid. Let's let the maintainers decide what they prefer.

bhennion:
> I have found another way to fix the issue further up. Your approach is of course still valid. Let's let the maintainers decide what they prefer.

I personnally tend to favour #6739, in that NaN's should NOT be generated at all. But I think there may be a small issue in #6739 (see https://github.com/xournalpp/xournalpp/pull/6739#discussion_r2426928562).

lluisalemanypuig:
Agree, the algorithm should not generate NaN values, but I didn't want to mess around with that.

lluisalemanypuig:
Given the solution in #6739, I do not think this merge request needs to still be open. You may close this when you see fit.

PR #6739, opened 2025-10-09
Fix zero inertia

### LINKED ISSUE #6717: Stroke recognizer produces NaN values

tmoerschell:
Fixes #6717

This is a different approach than #6737. I believe this fixes the source of the issue, rather than catching it later.

`Inertia::calc()` received two indices from `findPolygon()` that were both inclusive. This isn't the same as we usually know from iterators, and not the way it was intended to initially in `Inertia::calc()` as well as its usage in `CircleRecognizer::recognize()`. There's two ways of fixing that:
* Align `Inertia::calc()` with the usage of `findPolygon()` and `RecoSegment::calcSegmentGeometry()` (first commit)
* Use `std::span` as argument to `Inertia::calc()` (squash of both commits)

I think modifying the used pattern in `findPolygon()` is more risky than it should be. I find that `std::span` fits well here, as it is more expressive than `first, last` pointers, which was the cause of the issue at hand.

There is _technically_ a change in how the strokes are recognized, but since it only includes one more point in the initial detection, it shouldn't really change much.

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -3,6 +3,7 @@
 #include <cmath>     // for fabs, M_PI
 #include <iterator>  // for begin, next
 #include <memory>    // for allocator...
+#include <span>      // for span
 #include <utility>   // for move
 #include <vector>    // for vector
 
@@ -173,14 +174,16 @@ auto ShapeRecognizer::tryRectangle() -> std::unique_ptr<Stroke> {
     return s;
 }
 
-/*
- * check if something is a polygonal line with at most nsides sides
+/**
+ * Check if something is a polygonal line with at most nsides sides.
+ * [start, finish] (inclusive) specifies the range to operate on in the pt array.
  */
-auto ShapeRecognizer::findPolygonal(const Point* pt, int start, int end, int nsides, int* breaks, Inertia* ss) -> int {
+auto ShapeRecognizer::findPolygonal(const Point* pt, int start, int finish, int nsides, int* breaks, Inertia* ss)
+        -> int {
     Inertia s;
     int i1 = 0, i2 = 0, n1 = 0, n2 = 0;
 
-    if (end == start) {
+    if (finish == start) {
         return 0;  // no way
     }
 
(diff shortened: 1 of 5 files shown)


src/core/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -197,7 +198,7 @@ auto ShapeRecognizer::findPolygonal(const Point* pt, int start, int end, int nsi
     for (; k < nsides; k++) {
         i1 = start + (k * (end - start)) / nsides;
         i2 = start + ((k + 1) * (end - start)) / nsides;
-        s.calc(pt, i1, i2);
+        s.calc(std::span<const Point>(pt + i1, pt + i2 + 1));
```
bhennion, line 203: Is it guaranteed that pt + it2 + 1 is not past-the-end?

tmoerschell: `pt + i2 + 1` is precisely past-the-end, which is what the constructor of `std::span` requires. We could use the pointer + size constructor as well, but I thought a size of `i2 - i1 + 1` would be even more confusing.

`findPolygonal` is called the first time with `start` and `end` as inclusive indices:
https://github.com/xournalpp/xournalpp/blob/e504ec3ac2faf27fb45fa19f36506427f6ee73fc/src/core/control/shaperecognizer/ShapeRecognizer.cpp#L345-L346

`i2` is at most `start + (nsides * (end - start) / nsides)`, with `k == nsides - 1`, i.e. `i2 = end`.

bhennion: I see, then maybe `end` should be renamed (e.g. to `last`) for better readability.

tmoerschell: `std::algorithm` uses `first` and `last` iterators. I'll rename it to `finish`

tmoerschell:
Added comment, renamed variable `end` -> `finish`, and squashed commits

Issue #5535, opened 2024-03-03
Save too slowly when there are too many pages


zyjdmmm:
### Is Your Feature Request Related to a Problem?

Save too slowly when there are too many pages,more than 200 pages often take more than 10 seconds, which is difficult to accept
![image](https://github.com/xournalpp/xournalpp/assets/75295100/918a20d6-053e-4526-bf58-e3924656ec39)

### Feature Description

I want to save only the changed pages, so it's fast

vursc:
Since .xopp files are basically gzipped XML, implementing incremental saves is impossible without a radical change of the file format. However, the most time-consuming part of the save operation seems to be compression, which can be sped up by ~4x with a ~20% file size increase. Adding an option to save with faster compression would probably be possible.

SJK9906:
i have this prole with my book of math in .pdf format while editing it

eldipa:
> Since .xopp files are basically gzipped XML, implementing incremental saves is impossible without a radical change of the file format. 

Working on that! For the curious, we are tracking the progress in #937 

> However, the most time-consuming part of the save operation seems to be compression, which can be sped up by ~4x with a ~20% file size increase. Adding an option to save with faster compression would probably be possible.

It is true that the compression adds an overhead but depending of the content of the `.xopp` file it may not be the biggest problem. If the file has several Latex or images, those are getting copied over on save and that is a quite slow process.
I don't have the exact numbers but I remember a save operation with compression taking ~10 secs and without compression ~7 secs.

I hope #937 will change this for better.

eldipa:
By the way, @zyjdmmm would you like to share your `.xopp` file ? (assuming that has no personal information).
I'm collecting `.xopp` files for evaluate the performance of the current file format and the next generation in #4862

zyjdmmm:
@eldipa sure,here is!
[我的高数草稿本2.zip](https://github.com/xournalpp/xournalpp/files/14783586/2.zip)

eldipa:
> @eldipa sure,here is! [我的高数草稿本2.zip](https://github.com/xournalpp/xournalpp/files/14783586/2.zip)

Thanks!

eldipa:
> Only 10 seconds is good With multiple images (and my slow work pc it takes forever. But good to know it's being worked on. Thanks :D I could send an example file too if it helps

Yeah, I started to work on the new format exactly for that same reason!

Issue #5521, opened 2024-02-25
Scale option for images to improve performance


alessioC42:
### Is Your Feature Request Related to a Problem?

When inserting multiple images (or just one large image) into xournalpp, the application seems a bit slow. Exporting and saving .xopp files basically freezes the application. This problem does not exist when inserting the image as a background, so I guess the problem is the image data stored in the .xopp file.

### Feature Description

An option to scale image resulution (mb only on import or in general) should boost performance by a lot. Currently I am scaling the images with an external tool before importing in xournalpp to get a smoother experience.

rolandlo:
If you're on Linux you can actually already scale images (that you have selected) using the `ImageActions` plugin. It requires installing custom versions of `luaffifb` and `lua-vips` though.

See https://github.com/rolandlo/lua-vips/tree/fix-lua-5.3?tab=readme-ov-file#installation

alessioC42:
Well, that works great. But most people I know using Xournalpp are using it on Windows due to   data safety issues with onenote and other cloud based closed source services. The one thing is that they're on Windows and the other is, that they do not know how to follow build instructions...

At least I have a kind of what I wanted solution for my problem now. Thanks.

iFabse:
I have a fresh install of Fedora with GNOME 46 and installed xournal++ with flatpak (v. 1.2.3). I copied this plugin to the plugins folder of xournal++. I also installed lua (v. 5.4) luarocks and I also runned the installation of lua-vips and lua-luaffifb from the link above. 

But even with that, there is an error, that says: "/pathto/main.lua:50 attempt to call a nil value (field 'openDialog')"

In main.lua at line 50: 
local function checkVips()
  local hasVips, vips = pcall(require, "vips")
  if not hasVips then
    app.openDialog(
      "You need to have lua-vips installed and included in your Lua package path in order to use this plugin. \n" ..
      "Use luarocks install lua-vips and see https://github.com/libvips/lua-vips for further information.\n\n",
      { "OK" },
      "", true)
    return nil
  end
  return vips
end

It says that lua-vips has to be installed. But I've already installed it. 

How should I fix this problem since it worked on my old installation?

rolandlo:
`app.openDialog` uses the development version (nightly build). It's not available in the stable release 1.2.3. 
You could replace it by `app.msgBox` (with slightly different syntax), but then you would have to switch later, because `app.msgBox` is not Gtk 4 compatible. See #5079

iFabse:
[...]

Issue #4405, opened 2022-11-12
Split file into multiple


Dolland:
Since i always produce 200+ pages files, saving takes a lot of time (10+ seconds).
So, besides the ability to turn saving a multithread job (i know there is some work under it) it can be nice to be able to split a large file into multiple files, just in case someone like me forget how saving became a long task as pages grown.
Thanks!

eldipa:
@Dolland hi, the idea is to work on a new file format with a better performance than the current one (check #937 ). I will be publishing the design and progress but it is a long process however. 

If you can and want, you could submit you `.xopp` files for gathering statistics. Check #4862 to see what we learnt so far.

Dolland:
> @Dolland hi, the idea is to work on a new file format with a better performance than the current one (check #937 ). I will be publishing the design and progress but it is a long process however.
> 
> If you can and want, you could submit you `.xopp` files for gathering statistics. Check #4862 to see what we learnt so far.

Hi!
Here are some files i made on xournal++
Hope they can help you!

https://drive.google.com/file/d/1rzP2csfM0Zoc_u1y72qdor3BRClzmAWT/view?usp=sharing

eldipa:
Thanks! I took a look at your files and I understand why you wanted to split them into multiple ones and save them in background, I have a files of ~40MB too :)

Dolland:
> Thanks! I took a look at your files and I understand why you wanted to split them into multiple ones and save them in background, I have a files of ~40MB too :)

well the digital electronics one is around 100 MB, it causes a massive delay when saving, so splitting could be a nice workaround while rewriting the saving system imho :)

eldipa:
Yes, actually, it was my first idea as a workaround but I reviewed the code and I think it cannot be implemented directly.

Under the hood Xournal++ (mostly) does not know about the `.xopp` file: it has different objects in memory that represent all the elements of a document.

When a `.xopp` is read, it is read only once and all its content is loaded into memory. When you save the document, Xournal++ scans all the objects in memory and writes a single `.xopp` file.

Reading multiple `.xopp` files (like *parts* of the same document) it may be easy to implement (and it could be backward compatible). Writing may not.

Xournal++ does not know which objects (in memory) had changed or not so it always does a full scan and writes everything to disk. If we split the file into *parts* it will be have the same cost.

I know that implementing a new file format will *not* solve this: it should make the files smaller, yes, but Xournal++ will still write the file from scratch.

[...]

Issue #2684, opened 2021-01-21
Significantly Long Saving Times as Documents Become Longer


Izerpizer:
(Please complete the following information, and then delete this line)

**Affects versions :**
 - OS:  GNU/Linux (Ubuntu 20.10)
 - (Linux only) Desktop environment: Gnome X11
 - Which version of libgtk do you use: 3.24.23
 - Version of Xournal++: 1.1.0+dev
 - Installation method: apt

**Describe the bug**
As a file accumulates more and more content (I have only been writing hand written text) the saving wait times also increase. I would expect a slight increase to the save times for large files, but it seems that for normal sized documents, the saving times can become quite significant (on the order of quite a number of seconds (in my case for a 25 page file I was waiting about 3 seconds each save)). This isn't a big deal at all, I just see that there is some room for optimization in how the files are saved.

**To Reproduce**
1. Fill >10 pages with normal sized handwritten text
2. Save the document
3. observe the wait time for the cursor processing indicator.

**Expected behavior**
I expect that the saving behaviour would consistently (reasonably I would say) stay under 1 second, since the saving mechanic holds up the document for processing.

**Additional context**
Computer specs: 
CPU: Intel core i5-4690k
RAM: 8GB DDR3 1600MHz
SSD: Samsung 860 EVO
GPU: nvidia GTX 1080 and nvidia GTX 760

Izerpizer:
I would like to note that I want to change the label from "bug" to something like "optimization", but I can't figure out how.

tattsan:
gzip is used for compression, but it does not support multi-threading. If we could use multi-threading for compression, it would be much faster.

Technius:
> I would like to note that I want to change the label from "bug" to something like "optimization", but I can't figure out how.

This is a bug IMO, there's no excuse for even a small 10 page document to take more than a second to save.

> gzip is used for compression, but it does not support multi-threading. If we could use multi-threading for compression, it would be much faster.

Do you have profiling data to support this claim? I don't think it would be useful if we optimize a part of the program that turns out to only take up a small portion of the saving time.

Technius:
Related: #1869.

Technius:
See https://github.com/xournalpp/xournalpp/issues/1869#issuecomment-765176577 for how we can get an easy speedup.

tattsan:
What discussed in #1869 is commpression ratio, but not multi-threading.
I tested pigz, a multi-threading version of gzip, on a machine which has 8 cores.
[...]

PR #6750, opened 2025-10-16
Improved save speed

Hicortab:
When documents contain a lot of things (such as images), saving becomes really slow. I made sure to create a small "variant" of the xopp format that maintains the page uids that must be unique. These UIDs are used to identify pages when saving.

In particular, I create a new xml file which is resultDoc and I scroll through all the pages of the document with getPages(), I check if the UID of the page is marked in the changes and I rewrite it, otherwise I "copy-paste" it from the original file.

I used pugiXML because it seems to have interesting features and a good benchmark compared to other libraries. There are libraries that are much faster but which would make code management more complicated. With this solution of mine, the file I was testing took about 17 seconds on my laptop and now it takes about 6 seconds.

I tested it a bit and it seems to save the files correctly maintaining the right order even if they are moved, deleted, created other pages etc...

I made sure that it is also backwards compatible: If the user saves the file for the first time then it will write the file with the old system by inserting the UIDs in the XML format, if it is not the first time and the file has already been saved with this method then he will execute the new system which will be faster.

Could someone give me some tips on how to create testcases and review the code a bit? Maybe giving me some edge-cases I didn't think of.

### DIFF
--- a/.github/actions/install_deps_mac/action.yml
+++ b/.github/actions/install_deps_mac/action.yml
@@ -23,7 +23,9 @@ runs:
     if: steps.restore-blob.outputs.cache-hit != 'true'
     shell: bash
     run: |
-      brew install cargo-c
+      brew install cargo-c 
+      brew install openssl
+      brew install pugixml
       mkdir "${{github.workspace}}/blob"
       bash "${{github.workspace}}/mac-setup/CI_jhbuild.sh"
       mv xournalpp-binary-blob.tar.gz "${{github.workspace}}/blob/"
--- a/.github/actions/install_deps_ubuntu/action.yml
+++ b/.github/actions/install_deps_ubuntu/action.yml
@@ -39,6 +39,8 @@ runs:
           libgtest-dev
           libgtksourceview-4-dev
           libqpdf-dev
+          libssl-dev
+          libpugixml-dev
       run: |
         sudo apt-get update
         sudo apt-get install -y ${{env.packages}}
--- a/.github/actions/install_deps_windows/action.yml
+++ b/.github/actions/install_deps_windows/action.yml
@@ -35,6 +35,8 @@ runs:
           mingw-w64-${{inputs.msys_package_env}}-lua-luarocks
           mingw-w64-${{inputs.msys_package_env}}-qpdf
           mingw-w64-${{inputs.msys_package_env}}-zstd
+          mingw-w64-${{inputs.msys_package_env}}-openssl
+          mingw-w64-${{inputs.msys_package_env}}-pugixml
     - shell: msys2 {0}
       run: |
         # install LuaGObject via luarocks as long as there is no MinGW package


Hicortab:
@tmoerschell do you have any suggestions for improving this change?

tmoerschell:
Thank you for your continued work on this!

Skimming through the changes:
* I'm not a big fan of adding UIDs into the file, because it's not really information about the content. I understand that it makes your proposal easier to implement, and they might be widely used in other formats. If I had to implement this I'd rather track original page positions, say using a `std::optional<std::size_t> originalFilePagePos` member for each page, which is set when a page is loaded from a file, and reset when a page is modified or otherwise can no longer be fetched from the original file. This keeps the logic entirely in memory and avoids including implementation details in the file format. Other people might have different opinions though.
* I see that you made the change to `std::to_chars`. That's good! Each use is pretty similar and could be abstracted into a helper which returns a buffer or a string. In my understanding, if the encoding buffer is large enough, it should never fail, so we could assert that there was no failure, and unconditionally NUL-terminate the buffer.
* Instead of creating an `isLegacy` attribute, you would increment the file version and make decisions on that basis instead. Such a state should also not be stored in `StringUtils`.
* The save-load-save cycle for intermediate files still exists
* In general, please avoid commenting out code. If you change code, you should always replace it. Git allows you to check older versions much better. Also please avoid leaving any comments or error messages that are not in English.
* I believe the following points have not been addressed:
> * There is no consideration for what should happen if we save the file multiple times.
> * How do we ensure that the file that is on the disk is really the same that we loaded initially? If say another instance of Xournal++ is writing to it concurrently, some assumptions break. That could be solved for example by computing a hash of the originally loaded file and comparing it before saving.

I haven't tested correctness or performance. I think smaller, well-defined improvements, like moving to `std::to_chars`, would be easier to review and could get merged more quickly if submitted separately. I'm no maintainer though, so I decide nothing.

Hicortab:
Thank you so much for the valuable suggestions, really. I might wait for @bhennion to understand what direction to give to the PR.

bhennion:
Here's my point of view.
Overall, this is great. I did not suspect we could gain so much without modifying the file format (and implement random access therein).

I agree with @tmoerschell on his remarks though:

- Storing the uuid in the file does not seem necessary (and should hence be avoided if indeed possible)
- Switching to std::to_chars is great, but could/should be split off to a separate PR: it will make it much easier to review and test the changes independently.
- We should not introduce global variables (suh as StringUtil::isLegacy or UndoRedoHandler::pagesChanged/pagesChangedUndo). This is in general a bad habit. If it is indeed needed, then isLegacy should be a member of Document. The deques in UndoRedoHandler should not be static.
- The concerns about "consecutive saving ops" and e"xternal modifications" need to be addressed as well.
- The CI scripts need to be adapted to add your extra dependency in the CI tests.
- There are also some coding convention things, but those will come in a later review, once the core points have been addressed.

Hicortab:
@tmoerschell please excuse my silly questions, but I haven't quite understood how to handle the saving process in memory. Unless I write the file as a string, in which case I would have to rewrite all the writeOut functions, perhaps you had something in mind that I haven't thought of? Please give me some ideas.

Also, I haven't quite understood the "edge cases" regarding repeated saving and external modifications. For repeated saving, there doesn't seem to be any problem, while for external modifications, I think I read a Pull Request that proposes avoiding opening an already open file; why don't we consider implementing that?

tmoerschell:
The idea is indeed to save modified pages to a string. It is probably most easy to create a new child class of `OutputStream` that writes to a string instead of a gzip or zip file. You would then directly call `SaveHandler::saveTo(OutputStream* out, ...)` and it should work without major changes. Now we'd still have the write-read-write cycle, even though the first steps operate on an uncompressed string, which is already much faster.

How would we get rid of that last cycle? You currently merge two pugixml document trees, for which pugixml needs to parse both sides (even if it is fast). Instead, we could directly write out all pages through the `SaveHandler` and include all preexisting pages as a new node type, say `XmlPreserializedNode`, whose `writeOut()` function simply writes the original string. This means you'd do the original page lookup in `SaveHandler::visitPage()`. If the page was modified, create the `XmlNode` as is currently being done. If it was not modified, insert a `XmlPreserializedNode` that contains a string of the serialized original page. While this might require you to move many things around, I believe it has not only the benefit of being the most efficient, but also ensures that the saving logic is every time the same. We also retain more control over the XML output (header etc.) since pugixml does not write the whole file for us.

If you choose to go this route, here are some pointers for how _I would_ do the implementation (but feel free not to do it, or do it in another way):
* The page filtering logic can be removed.
* Saving the file to a string (first paragraph) is not required.
* Most of the preparations should be done in `SaveHandler::prepareSave()`. There you have access to the whole document and the target path, which is everything you need. First check if the target path exists and is the same as the original (see below). If it is, load it and forward a reference to it to the `visitPage()` function.
* In the `visitPage()` function, check if you can copy the page over. If yes, write it (via `pugi::xml_node::print()` I think) to a string and insert it as a `XmlPreserializedNode`. Otherwise, use the normal way.
* Right now, all `Xml*Node` class derivatives inherit from `XmlNode`. This doesn't work very well for our `XmlPreserializedNode` because the data representation is different, and most member functions don't apply (all `setAttrib()` overloads and `addChild()`). Either we rename the current `XmlNode` to something like `XmlContentNode`, and make the new `XmlNode` a base class with basically only the `writeOut()` functions, or we leave `XmlNode` as is and create a new class like `AbstractXmlNode`, which becomes the new base class, which both `XmlNode` and `XmlPreserializedNode` can inherit. I'd tend to prefer the second method, though one might be able to find better names. The current `XmlNode::addChild()` needs to accept the base class as parameter.

The comment about repeated saving was that when I save the document, all pages should afterwards be marked as unmodified (or "original"), such that when we re-save the file after having done further modifications, the pages that were modified the first time do not need to be rewritten again, but can be copied from the original file as if they had always been there. It's not critical, as it will not affect correctness of the document, but we might as well go all the way to use the new method as much as possible.

About external modifications, imagine the following: we open a 6-page file using two different Xournal++ instances, and perform different modifications in both. In instance 1, we remove page 2, then modify the two last pages, now 4 and 5, while in instance 2, we modify page 1. We save the file from instance 1, and it will contain the same content as is open in instance 1. If we then save the file from instance 2, it would currently read the modified file and consider that as the "original", which is wrong. Page 1 will be newly written, but pages 2 doesn't exist anymore, 3-6 are shifted, and if we manage to track which page is which, we would end up copying the two modified pages 5 and 6 instead of what we currently see in the application. For one, this is complicated to implement, and additionally the last point means the file we just wrote doesn't contain the same as what's in the application. I think we should detect if the file has changed since it was opened, and if it did, rewrite the entire file, as we currently do. In this way we do not need to worry about such issues. It is indeed caused by somewhat weird user behavior, but we should still guarantee consistency. The idea I had floated was that we could compute a hash of the file (maybe incrementally, while we are reading it, in `LoadHandler::parseXml()`), and when saving, recompute the hash to check if the file is still the same. There might be better methods though.

Preventing multiple instances from opening the same file would require a central knowledge of open documents, which we currently don’t have, or some kind of lock files. Even then, it wouldn't protect you against manual modifications, or more likely to occur, modifications from another app capable of working with `.xopp` files.

No need to be sorry for a good question. Keep up the good work! I'd like to mention again that I am _not_ a maintainer, so these are only thoughts that you may decide to follow or discard. I wouldn't want to come over as assuming too much authority.

Hicortab:
Okay, I think I understand much better what you meant, especially regarding the edge cases. Just to be safe, since there seem to be many changes, let's see what a maintainer thinks. @bhennion any suggestions?

bhennion:
I think what @tmoerschell proposes is reasonable. The extra benefit of storing the serialized pages in memory is that the "external modification" issue goes away: we don't need to read the on-disk file upon saving anymore.
There is the question of memory usage, but I think the cost is relatively small so it's ok.

Implementation-wise, it is actually pretty simple to keep track of modified pages now: when a page is modified, delete the PreserializedNode (if any). When saving, if there is no PreserializedNode in the Page instance, create it. Otherwise, use the preexisting one.

Hicortab:
Hey guys, could you take a look at my commit? Is there anything that could be improved? @bhennion. @tmoerschell I apologize for tagging you, but you were invaluable to me in this commit. Is there anything wrong? Could you also close https://github.com/xournalpp/xournalpp/pull/6719? I'm already taking care of it in this PR.

Hicortab:
Could we work on collecting and closing old PRs related to xournal's performance? For example, I believe that https://github.com/xournalpp/xournalpp/issues/4405 and https://github.com/xournalpp/xournalpp/issues/5521 https://github.com/xournalpp/xournalpp/issues/5535 https://github.com/xournalpp/xournalpp/issues/1869 https://github.com/xournalpp/xournalpp/issues/2684 are already addressed by this PR. 

I would like to see clear updated references in the README (given that some references to certain press releases date back to 2020, or even if they aren't that old, nobody has written to update them for years) to very complex pull requests such as:

* Changing the file format
* Supporting the embedding of PDF files in XOPP format (that would be really interesting)
* ...

Furthermore, I also read all of https://github.com/xournalpp/xournalpp/issues/937, but is it really necessary to change the format? Someone is still working on it? With the current modifications, is it possible to have xopp files on which to perform benchmarks? Maybe create a public collection of very large notes that reflect real user needs and not just 10,000 pages or 1000; I don't think anyone would write that many pages.

Hicortab:
@rolandlo @bhennion have you had time to see this PR?

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -242,6 +246,8 @@ auto LoadHandler::parseXml() -> bool {
         }
     } while (len >= 0 && valid && !error);
 
+    this->doc->setRawPageXmlDocument(xmlStream.str());
```
bhennion, line 249: So both the whole document and the individual pages are kept as xml in memory. That means two copies when one should be enough.
I don't think keeping the whole document in xml in memory is actually useful, is it?

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -256,9 +262,23 @@ auto LoadHandler::parseXml() -> bool {
...
     g_markup_parse_context_free(context);
 
-    // Add all parsed pages to the document
     this->doc->addPages(pages.begin(), pages.end());
 
+    std::vector<PageRef> pagesVector = doc->getPages();
+
+    size_t j = 0;
+
+    for (pugi::xml_node pageNode: this->doc->getRawPageXmlDocument().child("xournal").children("page")) {
```
bhennion, line 271: So the document xml is now actually parsed a second time? It doesn't look necessary.
@tmoerschell could the g_markup business give us the pages as xml?

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -1102,6 +1123,10 @@ auto LoadHandler::loadDocument(fs::path const& filepath) -> std::unique_ptr<Docu
     initAttributes();
     this->doc = std::make_unique<Document>(&dHanlder);
 
+    this->doc->setFileHash(StringUtils::calculateFileSHA256(filepath.string()));
```
bhennion, line 1126: So what is the hash used for now? We don't need to read the file upon saving anymore, so this doesn't look necessary for this PR.
(It may be worth worrying about overwriting external modifications, but it is independent of this PR and is another discussion/issue)

src/core/model/XojPage.h
```diff
@@ -85,12 +85,23 @@ class XojPage: public PageHandler {
...
+
 private:
     /**
      * The Background image if any
      */
     BackgroundImage backgroundImage;
 
+    std::string rawOriginalXml;
+
+    std::string uid;
```
bhennion, line 103: To identify a page in memory, no need for UIDs. Simply use the address.

src/core/undo/UndoRedoHandler.cpp
```diff
@@ -134,6 +145,14 @@ void UndoRedoHandler::redo() {
     this->undoList.emplace_back(std::move(this->redoList.back()));
     this->redoList.pop_back();
 
+    if (this->pagesChangedUndo.empty()) {
+        g_warning("UndoRedoHandler::redo(): pagesChangedUndo is empty while redoList is not. State is inconsistent!");
+    } else {
+        this->pagesChanged.emplace_back(this->pagesChangedUndo.back());
+        this->pagesChangedUndo.pop_back();
```
bhennion, line 152: I see several issues with those queues (both of them are already mentioned in old posts)

1. When the document is saved, a modified page is converted into xml. The page should then no longer be considered as modified (until another modification happens). This does not allow that.
2. As some UndoActions touch several pages, you single pop-push does not work. This can lead to data corruption in saved files.

I think the best solution (again, already mentioned) is to delete the "Preserialized" data of a page when it is modified. It would force the SaveHandler to redo this page's serialization upon saving. SaveHandler would then store this page's serialization for next time.

src/util/include/util/StringUtils.h
```diff
@@ -29,6 +29,13 @@ class StringUtils {
     static std::string trim(std::string str);
     static bool iequals(const std::string& a, const std::string& b);
     static bool isNumber(const std::string& input);
+    static std::string generateUniqueAlphanumericString();
+    static bool isLegacy;
```
bhennion, line 33: Please do not introduce global variables.

src/util/OutputStream.cpp
bhennion, line 1: This all file is modified, but I cannot see anywhere in the code the use of (Gz)InputStream. Please clean it up (unless I missed it).

src/util/Util.cpp
bhennion, line 1: There are a bunch of such files, where you just delete the trailing `\n`. Please clean that up.

src/core/control/xojfile/LoadHandler.cpp
tmoerschell: It is unfortunately not possible to directly retrieve elements as XML from GMarkup, nor obtain the element's locations in the reading buffer. It would however be possible to save the serialized pages while reading them for the normal file parsing, avoiding the additional decompression cycle. Only the logic would need to be implemented manually (searching for `<page ` and `</page>`, copying everything in between).

PR #6760, opened 2025-10-19
fix: handle unicode filenames in zip operations on windows #6757

### LINKED ISSUE #6757: Saved file cannot be opened again
### Operating System/Distribution

Windows 11 23H2

_No response_

None

### Installation Method

I downloaded it from github

### Xournal++ Version

xournalpp 1.2.8+dev (b41fb171 from /)

### libgtk Version

3.24.43

### Steps to Reproduce

1. Create a file and safe it.
2. Close the Xournal Programm
3. Open the saved file --> error "Your document has been saved in: FAILED" (the document can be opend via the "recently used files" section in the Xournal programm.

### Crash log

Image

### Additional Context

_No response_

youssefadly237:
tested on windows and linux works fine

closes #6757

### DIFF
--- a/src/exe/Xournalpp-wrapper.cpp
+++ b/src/exe/Xournalpp-wrapper.cpp
@@ -17,6 +17,13 @@
 #include <gio/gio.h>
 #include <gtk/gtk.h>
 
+#ifdef _WIN32
+// clang-format off
+#include <windows.h>
+#include <shellapi.h>
+// clang-format on
+#endif
+
 #include "util/PathUtil.h"
 #include "util/VersionInfo.h"
 #include "util/XojMsgBox.h"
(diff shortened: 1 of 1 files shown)


youssefadly237:
looks like I have to use windows

youssefadly237:
solved it but I need to clean up debugging and some other stuff will be amending it tomorrow I think

rolandlo (commented):
Thanks for the fix! I haven't tested the patch yet, but it seems to me that the issue also exists in the release branch and should be fixed there as well. So if that is the case could you rebase onto the release branch?

src/util/ZipUtil.cpp
```diff
@@ -0,0 +1,31 @@
...
+#ifdef _WIN32
+    // Convert wide string path to UTF-8 for libzip
+    const std::wstring widePath = path.wstring();
+    const int utf8Size = WideCharToMultiByte(CP_UTF8, 0, widePath.c_str(), -1, nullptr, 0, nullptr, nullptr);
+    std::string utf8Path(utf8Size - 1, '\0');
+    WideCharToMultiByte(CP_UTF8, 0, widePath.c_str(), -1, utf8Path.data(), utf8Size, nullptr, nullptr);
+    return zip_open(utf8Path.c_str(), flags, error);
+#else
+    return zip_open(char_cast(path.u8string().c_str()), flags, error);
+#endif
```
bhennion: Is this disjunction really necessary? path.u8string() should do exactly the same thing as your Windows workaround.

youssefadly237:
Sure

src/util/ZipUtil.cpp
youssefadly237: I have made this change before I spotted the real problem, so it didn't really affect this issue, but it's an extra safeguard `u8string()` isn't guaranteed to give proper UTF-8 on all windows toolchains, while the Win32 conversion always does. So I thought it is a good idea to keep it

https://stackoverflow.com/questions/71959452/stdfilesystempathu8string-might-not-return-valid-utf-8

src/exe/Xournalpp.cpp
```diff
@@ -19,10 +19,46 @@
...
+            }
+        }
+        utf8ArgPointers.push_back(nullptr);
+
+        LocalFree(wideArgs);
+
+        argc = wideArgCount;
+        argv = utf8ArgPointers.data();
+    }
+#endif
```
rolandlo: This is duplicated now.

youssefadly237: oh sorry

rolandlo:
According to my testing the stable version already works fine. It's just the wrapper (on master) that has a problem (both when opening from the file explorer and opening from the command line). So why is anything needed in auto main(int argc, char* argv[]) in Xournalpp.cpp? Also there may be a simpler fix for the wrapper then. I'm confused.

youssefadly237:
Sorry I didn't test it properly, I was a little busy with my finals, I just changed both to unify the behavior, I will review it again tomorrow and mame sure it is the most minimal problem to the solution

youssefadly237:
Hey, thanks for the feedback

I rebased onto `release-1.2` following the suggestion that the issue might exist there too. After testing both branches, I found that the stable version works fine, the issue only exists in master

The problem is in the wrapper executable introduced in commit 01734d85f. On Windows, argv contains strings in the system ANSI codepage (Windows-1252), not UTF-8. When filenames with special characters (like German umlauts) are passed to the subprocess launcher, which expects UTF-8, the encoding mismatch causes a crash

The fix uses `GetCommandLineW()` to get arguments in UTF-16, then converts them properly to UTF-8 before passing them to the subprocess

I'll move the PR back to target master since that's where the actual bug is

PR #6781, opened 2025-10-31
Fix plugins

rolandlo:
See commit message for further explanations of the changes to the non-Lua files.

### DIFF
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -409,7 +409,7 @@ function app.addTexts(opts) end
 ---             size = 8.0,
 ---            },
 ---     color = 0x0,
----     x = 150.0,,
+---     x = 150.0,
 ---     y = 70.0,
 ---     width = 55.0,
 ---     height = 23.0,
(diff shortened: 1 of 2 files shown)


mjg:
I can confirm that the fix works when I compile locally (or cherry-pick on top of an older master).
Distribution builds (Fedora copr) crash with an STL error
```
/usr/include/c++/15/bits/stl_algo.h:3638: constexpr const _Tp& std::clamp(const _Tp&, const _Tp&, const _Tp&) [with _Tp = int]: Assertion '!(__hi < __lo)' failed.
```
which I expect to be unrelated. I'll recompile locally with the more stringent flags and try to bisect.

src/core/plugin/luapi_application.h
youssefadly237: can you fix this typo `,,`?

plugins/luapi_application.def.lua
youssefadly237: seiection should be selection

rolandlo:
Thanks @mjg for testing. Merging in 24 hours if no objections are raised.

src/core/plugin/luapi_application.h
youssefadly237: this was not fixed in b54287c71 btw

rolandlo: I thought you are talking about the "seiection" typo. It's fixed now.

youssefadly237: Yeah I meant both, thanks for the fix 🙏🏼

PR #6682, opened 2025-09-11
Expose Floating Toolbox to Lua API; Floating Toolbox clamping logic;

youssefadly237:
## Changes

- Added `app.showFloatingToolbox()` function to show the floating toolbox ~at the current cursor position (using `app.getCursorPosition()`) or~ at specific coordinates (x, y) to the lua API
- Implemented the methods in `Control.cpp` for handling ~cursor-based and~ manual positioning of the floating toolbox
- Updated documentation to include usage examples for ~both cursor and~ coordinate-based triggering.
- Added clamping logic to `gui/FloatingToolbox.cpp`

Closes #6624

## Testing

Tested on:
- OS: Kubuntu 25.10 (Questing Quokka) x86_64
- DE: KDE Plasma 6.4.4
- WM: KWin (Wayland)

~**Note**
Tested with positive and negative fractional scaling (~works fine~) 
Not tested on X11 or Windows~ tests was about get `getCursorPosition` which is dropped now

~This is a **showcase** implementation, I think it still needs some refining (e.g, displaying the floating tool box outside of the window itself), I just need maintainers' feedback before polishing further~

PR #6788, opened 2025-11-02
Fix FloatingToolbox clamp assertion failure

youssefadly237:
Fixes a crash introduced in 6ed3d6b5d where std::clamp() was called with inverted bounds (min > max)

Reported by @mjg
issue comment: https://github.com/xournalpp/xournalpp/pull/6682#issuecomment-3478112111


mjg:
Thanks a lot, that indeed fixes things for me. Wonderful turn-around time :-)

rolandlo (approved):
LGTM. Thanks for the fix!

PR #6790, opened 2025-11-03
Fix grid snapping for backgrounds with margins/offsets

youssefadly237:
## Problem
Grid snapping assumes all grids start at (0,0), causing misalignment on backgrounds with margins (graph paper, isometric, ruled)

## Solution
- Add `setPageRef()` to `SnapToGridInputHandler` for background context
- Calculate offsets from `BackgroundConfig` based on page type
- Cache offsets per page
- Add `snapToIsometricGrid()` for triangular lattice geometry
- Extend `Snapping::` functions to accept x/y offsets

## Handles
- Graph: margins + centered alignment
- Isometric: centered triangular grid (~7 vertex candidates)
- Ruled: header spacing (vertical offset)
- Dotted/Plain: origin-based (no change)

**Before**: Snaps to phantom grid positions
**After**: Snaps to visible background grid

## Related discussion
#6744

### DIFF
--- a/src/core/model/Snapping.cpp
+++ b/src/core/model/Snapping.cpp
@@ -22,19 +22,15 @@ double snapHorizontally(double x, double gridSize, double tolerance, double xOff
     return std::abs(xSnapped - x) < tolerance * gridSize / 2.0 ? xSnapped : x;
 }
 
-Point snapToGrid(Point const& pos, double gridSize, double tolerance, double xSpacing, double ySpacing, double xOffset,
+Point snapToGrid(Point const& pos, double columnSpacing, double rowSpacing, double tolerance, double xOffset,
                  double yOffset) {
-    // Calculate actual column and row spacing
-    const double columnSpacing = xSpacing * gridSize;
-    const double rowSpacing = ySpacing * gridSize;
-
     // Round to nearest grid point
     Point nearestVertex{roundToMultiple(pos.x - xOffset, columnSpacing) + xOffset,
                         roundToMultiple(pos.y - yOffset, rowSpacing) + yOffset, pos.z};
 
-    // Only snap if within tolerance threshold
+    // Only snap if within tolerance threshold (fraction of the half-diagonal)
     const double dist = distance(pos, nearestVertex);
-    const double snapThreshold = gridSize * tolerance;
+    const double snapThreshold = 0.5 * std::hypot(columnSpacing, rowSpacing) * tolerance;
     return (dist < snapThreshold) ? nearestVertex : pos;
 }
 
--- a/src/core/model/Snapping.h
+++ b/src/core/model/Snapping.h
@@ -35,17 +35,16 @@ namespace Snapping {
 [[nodiscard]] double snapHorizontally(double x, double gridSize, double tolerance, double xOffset = 0.0);
 
 /**
- * @brief If a points distance to the nearest grid point is under a certain tolerance, it returns the nearest
- * grid point. Otherwise the original Point itself. Supports arbitrary grid spacing ratios.
+ * @brief If a point's distance to the nearest grid point is under a certain tolerance, it returns the nearest
+ * grid point. Otherwise the original Point itself. Supports arbitrary column/row spacings.
  * @param pos the position
- * @param gridSize the base grid size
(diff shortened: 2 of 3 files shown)


src/core/model/Snapping.h
```diff
@@ -23,6 +23,16 @@ namespace Snapping {
...
+/**
+ * @brief If a value is near enough to the y-coordinate of a grid point (with offset), it returns the nearest
+ * y-coordinate of the grid point. Otherwise the original value itself.
+ * @param y the value
+ * @param gridSize the distance to each snapping point
+ * @param tolerance the tolerance as a fraction of a half grid diagonal (assumed to be between 0 and 1)
+ * @param yOffset the vertical offset of the grid (e.g., from page margins)
+ */
+[[nodiscard]] double snapVertically(double y, double gridSize, double tolerance, double yOffset);
+
```
rolandlo, line 26: Please use default argument `yOffset = 0.` here instead of having two functions. Similarly for `snapHorizontally` and `snapToGrid`

src/core/model/Snapping.h
```diff
@@ -74,4 +105,15 @@ namespace Snapping {
...
+/**
+ * @brief Snaps to the nearest vertex of an isometric (triangular) grid
+ * @param pos the position
+ * @param triangleSize the size of the equilateral triangles in the grid
+ * @param tolerance the tolerance as a fraction of the triangle size (assumed to be between 0 and 1)
+ * @param xOffset the horizontal offset of the grid
+ * @param yOffset the vertical offset of the grid
+ */
+[[nodiscard]] Point snapToIsometricGrid(Point const& pos, double triangleSize, double tolerance, double xOffset,
+                                        double yOffset);
```
rolandlo: Excuse my ignorance, but the way it currently works, isn't snapping to isometric grid the same as snapping to a rectangular grid like this:

Bildschirmfoto vom 2025-11-04 04-42-09

youssefadly237: Yeah good point, but merging them would slow down the common rectangular case since it'd check 9 candidates instead of just rounding. Probably better to keep them separate even if there's some duplication, or what do you think ?

rolandlo: The question is if `snapToIsometricGrid` is really necessary, when instead `snapToGrid` (that only does some rounding) could be used (adjusted to different grid sizes in x and y direction).

youssefadly237: can you check the last commit and make sure I got what you mean correctly?

rolandlo: Yes, that's a way I meant it.

src/core/model/Snapping.cpp
```diff
@@ -10,20 +11,31 @@ namespace Snapping {
...
+    const double columnSpacing = xSpacing * gridSize;
+    const double rowSpacing = ySpacing * gridSize;
+
+    // Round to nearest grid point
+    Point nearestVertex{roundToMultiple(pos.x - xOffset, columnSpacing) + xOffset,
+                        roundToMultiple(pos.y - yOffset, rowSpacing) + yOffset, pos.z};
+
+    // Only snap if within tolerance threshold
+    const double dist = distance(pos, nearestVertex);
+    const double snapThreshold = gridSize * tolerance;
```
rolandlo: `snapThreshold` should be the largest distance among all points in the plane to a grid point multiplied by the tolerance, i.e. half the diagonal of a (rectangular) grid cell multiplied by the tolerance. 
The idea is that tolerance 1 means that every point gets snapped.

```suggestion
    const double snapThreshold = .5 * std::hypot(columnSpacing, rowSpacing) * tolerance;
```

Moreover I think that `columnSpacing` and `rowSpacing` should be passed directly instead of passing `gridSize`, `xSpacing` and `yspacing`. It's one less parameter.

youssefadly237: thanks for the feedback 🫡

rolandlo (approved):
LGTM

rolandlo:
I'll squash all commits into one while merging (using squash and merge).

rolandlo:
Thanks again @youssefadly237 for your effort. It's great that the snapping works for more backgrounds.

PR #6792, opened 2025-11-03
WIP/RFD: Laser switch back

mjg:
I'm trying out the new laser tools, and I'm liking them!
For a presentation style where you mostly draw with intermittent "lasering", you want to switch back to drawing quickly. One option is putting the laser pen or laser hightlighter on a mouse or pen button - they switch back to the previously active tool automatically. But there are only few of those ..

Another option is to have the laser switch back to the previously active tool once the fade has faded off. This PR is a "POC", or rather a way to try whether this mode makes sense. I *do not* suggest to make this the default. It could be a preference if the general idea is liked. There certainly are workflows where you mainly present and want the laser to be *the* active tool.
The implementation is so that "previously active tool" could be used for other purposes than just the laser.

I noticed that the lasers are not hooked up in the menu nor via shortcuts yet. I don't quite grok the new actionActivate LUA API yet - would I need to add some more code before that can activate any tool (such as the lasers)?
So, for now it's one tap on the tool, and let it switch back after the fade. With a shortcut it cut be be one (tablet) key press.

### DIFF
--- a/src/core/gui/PageView.cpp
+++ b/src/core/gui/PageView.cpp
@@ -213,8 +213,11 @@ void XojPageView::endSpline() {
 }
 
 void XojPageView::deleteLaserPointerHandler() {
+    Control* control = xournal->getControl();
+    ToolHandler* h = control->getToolHandler();
     xoj_assert(hasNoViewOf(overlayViews, laserPointer.get()));
     laserPointer.reset();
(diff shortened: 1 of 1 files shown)


mjg:
Switching back does not work 100% reliably, btw - in particular, the toolbar icon switching.

PR #6823, opened 2025-11-17
make the latex-button a proper tool

### LINKED ISSUE #6164: Can not change Latex Color
### Operating System

Linux

### (Linux only) Distribution

nixos

sway

Wayland

### Installation Method

package nixos

### Xournal++ Version

1.2.5

### libgtk Version

3.24.43

### Bug Description

When doing latex it stays on random colors and I can not change it. This
is a temp fix here: https://unix.stackexchange.com/questions/718317/xournal-does-not-change-the-color-of-latex-equation

Here I am on white and it's blue trying to tap on any other color has no effect.

image

### Expected Behaviour

For latex to change colors when you select a color from the toolbar.

### Steps to Reproduce

1. Enter a latex formula
2. Try changing the color

The color does not change when you select a color from the tool bar.

### Additional Context

You can find a temp solution here, which describes the same problem I am facing: https://unix.stackexchange.com/questions/718317/xournal-does-not-change-the-color-of-latex-equation

immi-h:
The behaviour of the insert latex button is currently somewhat unintuitive; see issue "Can not change Latex Color #6164
". I turned into a proper tool (like "insert image"); in particular:
* It can be chosen like the other tools. (It does not open the latex edit box right away.)
* To actually create a latex element, one has to click into the document. The element is created at the click location (before: in the middle of the window)
* When the latex tool is active, one can choose the color to be used, like for other tools.
* One can edit a latex element by clicking it while the latex tool is active. (Before: select the latex element and then click on the latex button).
* Editing a latex element by double clicking it with the selection tool works as before.

(One could now make some follow-up improvements, which I'll rather implement separately.)

While everything seems to be working, note that I'm completely new to xournalpp development (and also pretty new to github in general), so please review my code carefully. Also, I may have violated some coding conventions. Please let me know if I should change things. Some concrete things:
* I'm unsure unsure whether the "LatexController" should become a "LatexHandler" (given that it does something rather similar to the "ImageHandler")
* I first tried to open the latex dialog already on the buttonpress event, but this made xournal crash. Now it's opened only on buttonrelease and it works fine, but I'm using a somewhat ugly looking workaround to store the button press location. I'm not sure whether that's the way it should be done.


rolandlo:
The changes sound great! I'm happy you did it! Please rebase on upstream/master and force push, so only your commits are part of this PR.

```term
git fetch --all
git rebase upstream/master
git push --force
```

immi-h:
Sorry about the commit messages being in German; I wasn't aware that they would become public. Should I change them?

rolandlo:
> Sorry about the commit messages being in German; I wasn't aware that they would become public. Should I change them?

Yes, please change the commit messages to English.

src/core/control/actions/ActionProperties.h
```diff
@@ -886,11 +886,6 @@ struct ActionProperties<Action::AUDIO_STOP_PLAYBACK> {
     }
 };
 
-template <>
-struct ActionProperties<Action::TEX> {
-    static void callback(GSimpleAction*, GVariant*, Control* ctrl) { ctrl->runLatex(); }
-};
-
```
rolandlo, line 906: You removed the TEX action, but it's still in the `mainmenubar.xml`
https://github.com/xournalpp/xournalpp/blob/5157f45515947c2d370fe7bd9d6e550192d6c50c/ui/mainmenubar.xml#L847-L852
This makes the test
`ActionDatabaseTest.testActionTargetMatch`
fail. Please correct that.

immi-h: I changed the commit messages and fixed the menu bar.
(Note: I did not update the POT/PO files.)
Btw: In enum ToolType, I just added a new number at the very end for TOOL_LATEX. Should I change this to a better order?
Also: Would you prefer if I'd run the tests myself? (I don't know yet how that works and did not yet even try, but I could certainly find out.)

rolandlo: > Also: Would you prefer if I'd run the tests myself? (I don't know yet how that works and did not yet even try, but I could certainly find out.)

Yes, it would be a good idea to run the tests yourself. It still fails.
```term
 [ RUN      ] ActionDatabaseTest.testActionTargetMatch
/Users/runner/work/xournalpp/xournalpp/test/unit_tests/control/ActionDatabaseTest.cpp:76: Failure
Value of: expectedTypes[a] == nullptr
  Actual: false
Expected: true

/Users/runner/work/xournalpp/xournalpp/test/unit_tests/control/ActionDatabaseTest.cpp:76: Failure
Value of: expectedTypes[a] == nullptr
  Actual: false
Expected: true
```

How to run the tests is explained in [Compile.md](https://github.com/xournalpp/xournalpp/blob/master/readme/Compile.md)

rolandlo:
Also please apply clang-format. Assuming you have installed `clang-format-18`:

```term
 git-clang-format-18 --binary=clang-format-18 HEAD~<n>
```
where `<n>` must be replaced by the number of commits to format.

immi-h:
Done.

rolandlo:
I applied some fixes. The tests run successfully now. 

When changing the color of a selection I would expect that every TexImage contained would be rerendered with the new color. Can you arrange for that?

immi-h:
I was planning to do that, but I thought it should be a separate pull request. (It's an issue that was already there before independently.) Also, I don't know yet how easy this is to implement. But I'll try to implement it as soon as I find time. (Thanks for the fixes btw.)

src/core/control/LatexController.cpp
```diff
@@ -339,8 +269,27 @@ void LatexController::insertTexImage() {
...
+    if (this->selectedElem) {
+        auto sel = SelectionFactory::createFromElementOnActiveLayer(control, page, view, selectedElem);
+        view->getXournal()->deleteSelection(sel.release());
+        this->selectedElem = nullptr;
+    }
+
+    if (view->getPage() != page) {
+        g_warning("Active page changed while you edited the tex code. Aborting.");
+        return;
+    }
```
bhennion: If we are to abort, then we should not delete the old TexImage: those two `if`-blocks should be swapped.

src/core/control/LatexController.cpp
```diff
@@ -339,8 +269,27 @@ void LatexController::insertTexImage() {
     xoj_assert(this->isValidTex);
     xoj_assert(this->temporaryRender != nullptr);
 
+    /* Das folgende wird erst am Ende gemacht... */
```
bhennion: German comment

src/core/control/LatexController.cpp
```diff
@@ -356,14 +305,56 @@ void LatexController::cancelEditing() {
...
     }
 
-    self->findSelectedTexElement();
+    self->page = page;
+
+
+    // Is there already a teximage?
+    self->selectedElem = nullptr;
+
+    // Should we reverse this loop to select the most recent teximage rather than the oldest?
```
bhennion: Yes we should

src/core/control/LatexController.cpp
```diff
@@ -356,14 +305,56 @@ void LatexController::cancelEditing() {
...
+    for (auto&& e: page->getSelectedLayer()->getElements()) {
+        if (e->getType() == ELEMENT_TEXIMAGE || e->getType() == ELEMENT_TEXT) {
+            GdkRectangle matchRect = {gint(x), gint(y), 1, 1};
+            if (e->intersectsArea(&matchRect)) {
+                self->selectedElem = e.get();
+                // ti = dynamic_cast<const TexImage*>(e.get());
```
```suggestion
```

src/core/control/LatexController.cpp
bhennion: There should also be a comment here, explaining what we are doing and why. The name of the function you "inlined" gave this piece of info before.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -401,7 +401,7 @@ void ToolMenuHandler::initToolItems() {
                                 _("Highlighter"));
 
     emplaceCustomItemWithTarget("TEXT", Cat::TOOLS, Action::SELECT_TOOL, TOOL_TEXT, "tool-text", _("Text"));
-    emplaceCustomItem("MATH_TEX", Cat::TOOLS, Action::TEX, "tool-math-tex", _("Add/Edit TeX"));
+    emplaceCustomItemWithTarget("MATH_TEX", Cat::TOOLS, Action::SELECT_TOOL, TOOL_LATEX, "tool-math-tex", _("Latex"));
```
bhennion: I would say the original description "Add/Edit TeX" was more self-explanatory than just "LaTeX".

ui/mainmenubar.xml
```diff
@@ -542,6 +542,12 @@
      <attribute name="target" type="t">4</attribute> <!-- TOOL_TEXT -->
      <attribute name="accel">&lt;Ctrl&gt;&lt;Shift&gt;t</attribute>
     </item>
+    <item>
+     <attribute name="label" translatable="yes">LaTe_X</attribute>
```
bhennion, line 560:
```suggestion
     <attribute name="label" translatable="no">LaTe_X</attribute>
```

src/core/control/actions/ActionProperties.h
bhennion: > Should I change this to a better order?

I don't think so. It could introduce some bugs with little benefits.


> Would you prefer if I'd run the tests myself?

It would be easier indeed. To do so, you'll need to do that in your build directory:
```
cmake .. -DENABLE_GTEST=on   (-DDOWNLOAD_GTEST=on if you don't have a local install of gtest)
cmake --build . --target test-units
./test/test-units
```

src/core/control/Control.cpp
```diff
@@ -2495,16 +2495,6 @@ void Control::fontChanged(const XojFont& font) {
...
 }
 
-/**
- * The core handler for inserting latex
- */
-void Control::runLatex() {
-    /*
-     * LatexController::run() will open a non-blocking dialog.
-     */
-    LatexController::run(this);
```
bhennion, line 2625: Can we then remove the LatexController.h include then?

bhennion (commented):
Here are a bunch of comments on the code. I haven't test it (yet).

src/core/control/Control.cpp
immi-h: Yes. Done.

src/core/control/LatexController.cpp
immi-h: Done.

src/core/control/LatexController.cpp
immi-h: ... and moreover, it didn't make much sense. I just removed it.

src/core/control/LatexController.cpp
immi-h: Good point; done. However, I can't really test this; I don't know how I could make the active page change while the latex dialog is open. (Can this happen at all?)

src/core/control/LatexController.cpp
immi-h: Both things done.
Note hat the code for the loop in the wrong direction was stolen from the text tool (control/tools/TextEditor.cpp). So it would make sense to also change it there. (I can do that myself at some point, but I don't know yet when I'd find time. Or should I submit it as a bug?)

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
immi-h: I had tried to make it more consistent with the descriptions of the other tools. But you're right, the old description was better. I changed it back.

src/core/control/actions/ActionProperties.h
immi-h: Thanks for that. (All tests passed. :-) )

archisman-panigrahi:
Apart from rebasing and resolving the merge conflict, does this PR need anything else?

rolandlo:
> Apart from rebasing and resolving the merge conflict, does this PR need anything else?

It will take proper testing regarding tool activation/abortion. The hack with saving the position on button press event and activating it on button release event doesn't look very convincing.

Also there is at least one comment which hasn't been taken into account.

immi-h:
I must have overlooked that comment that I did not take care of. Can you remind me what that was?

And concerning the hack with saving the location of the button press event: I also don't really like it and I'm happy to change it; the question is what we want:
- One could insert the latex object at the button release position. (Easy to implement.)
- One could open the latex dialog already when the button press happens. Maybe this would be most natural, and that's what I tried to implement first, but it didn't work. It seems that something was left in some wrong internal state; I don't understand the xournal code well enough to get that working. (If somebody can help, I'd be happy to implement that.)
- One could make the latex tool behave like the text tool: It also inserts the text only at button release, but there, it makes more sense, since one can create a rectangle for the text to be inserted by clicking and dragging. I felt that one might not want this with the latex tool, since in contrast to text elements, the latex content will be scaled. However, I'm happy to implement that if you think that makes sense.

rolandlo:
> I must have overlooked that comment that I did not take care of. Can you remind me what that was?
>
Looks like I have already fixed that comment myself. Sorry about the noise.
 
> And concerning the hack with saving the location of the button press event: I also don't really like it and I'm happy to change it; the question is what we want:
> 
>     * One could insert the latex object at the button release position. (Easy to implement.)
> 
>     * One could open the latex dialog already when the button press happens. Maybe this would be most natural, and that's what I tried to implement first, but it didn't work. It seems that something was left in some wrong internal state; I don't understand the xournal code well enough to get that working. (If somebody can help, I'd be happy to implement that.)
> 
>     * One could make the latex tool behave like the text tool: It also inserts the text only at button release, but there, it makes more sense, since one can create a rectangle for the text to be inserted by clicking and dragging. I felt that one might not want this with the latex tool, since in contrast to text elements, the latex content will be scaled. However, I'm happy to implement that if you think that makes sense.

We have the same situation with the URL link tool #6841 (and also a bit similar with the Image tool): Upon clicking into the document a dialog is opened which gets the focus. So as I learned in the process the dialog must be opened on button release. Otherwise an action is started (`actionStart`) that will never be ended (because the Xournal widget doesn't have focus anymore) leading to various problems. However the object should be inserted at the button release position, not the button press position. This is easy to fix, as you already pointed out.

Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release (following button double press) again, not directly on button double press, since the same problems with a never ending action like mentioned above will result. So this must be changed as well.

For all of this compare with the [changes](https://github.com/xournalpp/xournalpp/pull/6841/changes#diff-49482fa62c7b8598389514a9a4d3a302ecad170c18ee8c27526865fc1ddf377f) to the PageView class in #6841.

We could wait for #6841 to get merged. The decisions taken there with respect to the page view class will have to be taken here as well.

archisman-panigrahi:
> Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release

Please make sure that it would work with touch screens. Happy to test and give feedback

immi-h:
> We have the same situation with the URL link tool #6841 (and also a bit similar with the Image tool): Upon clicking into the document a dialog is opened which gets the focus. So as I learned in the process the dialog must be opened on button release. Otherwise an action is started (`actionStart`) that will never be ended (because the Xournal widget doesn't have focus anymore) leading to various problems. However the object should be inserted at the button release position, not the button press position. This is easy to fix, as you already pointed out.

Ok, I'll do that as soon as I find time.
> 
> Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release (following button double press) again, not directly on button double press, since the same problems with a never ending action like mentioned above will result. So this must be changed as well.

Double-click editing is already implemented and seems to be working, but I'll double-check whether it's implemented correctly (i.e., the dialog being opened only on button release).

> 
> For all of this compare with the [changes](https://github.com/xournalpp/xournalpp/pull/6841/changes#diff-49482fa62c7b8598389514a9a4d3a302ecad170c18ee8c27526865fc1ddf377f) to the PageView class in #6841.
> 
> We could wait for #6841 to get merged. The decisions taken there with respect to the page view class will have to be taken here as well.

That could make sense.

immi-h:
> > Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release
> 
> Please make sure that it would work with touch screens. Happy to test and give feedback

I just tried it; it seems that currently, editing by double-tapping with the finger does *not* work. (I don't know why. Is this intentional?) It does not work with the text tool either.

immi-h:
> We have the same situation with the URL link tool #6841 (and also a bit similar with the Image tool): Upon clicking into the document a dialog is opened which gets the focus. So as I learned in the process the dialog must be opened on button release. Otherwise an action is started (`actionStart`) that will never be ended (because the Xournal widget doesn't have focus anymore) leading to various problems. However the object should be inserted at the button release position, not the button press position. This is easy to fix, as you already pointed out.

This is fixed...

> 
> Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release (following button double press) again, not directly on button double press, since the same problems with a never ending action like mentioned above will result. So this must be changed as well.

... but I still need to fix this.

immi-h:
> Moreover double clicking on a LaTeX formula shall allow to edit the formula, opening the dialog again. This must happen on button release (following button double press) again, not directly on button double press, since the same problems with a never ending action like mentioned above will result. So this must be changed as well.

This is now also fixed. This, too, feels a bit like an ugly work-around, but so far, it seems to be working.

What I'm a bit worried is if that for some reason, onButtonRelease does not get called, then the variable inLatexDoubleClick, which will result in strange behaviour. On the other hand, the same issue would exist with the variable inEraser, so maybe one can rely on onButtonRelease being called?

A side effect of opening the dialog only on button release is that one can double-click, keep the mouse pressed, drag the latex object around, and when one then releases the mouse, the latex dialog opens. That might feel a bit strange; on the other hand, I think it's ok, and it doesn't seem to cause any problems.

immi-h:
I just found another bug (or at least: rare strange behaviour):
- Create a latex object A
- Create a second latex object B somewhere else.
- Switch to the selection tool.
- Double-click on A, but instead of releasing the mouse right away after the second click, drag object A over object B. (Note that A is below B, but since A is selected, it is currently drawn above B. I guess this is intended.)
- Now release the mouse button. It is maybe a bit strange that the double-click is recalled and that the latex edit dialog opens. But what's really strange is that it's B which gets edited and not A.

I'll fix that.

immi-h:
> I just found another bug (or at least: rare strange behaviour):
> 
>     * Create a latex object A
> 
(...)
> 
>     * Now release the mouse button. It is maybe a bit strange that the double-click is recalled and that the latex edit dialog opens. But what's really strange is that it's B which gets edited and not A.
> 
> 
> I'll fix that.

Concerning the above issue: I just checked how this is done for text objects, and it has a similar issue, so I'd now have the tendency to leave it as it is for the moment, and once the new latex tool is part of the main branch, I can fix the problem for both tools at once.

Here is an explanation how the strange behavior arises: on a doubleclick event, xournalpp checks whether the click location is on a text or on a latex element, and if yes, it triggers opening the editor. But instead of telling the editor which object to edit (namely the one which was double-clicked), the selection is cleared and the editor is left to find out on its own which object to edit. In some occasions, this can be a different object than the one which triggered opening the editor.

bhennion:
> Concerning the above issue: I just checked how this is done for text objects, and it has a similar issue, so I'd now have the tendency to leave it as it is for the moment, and once the new latex tool is part of the main branch, I can fix the problem for both tools at once.

I think this is fine. This is a rare enough occurrence. You can open an issue ticket so that we don't forget about it.

I'll give your PR a read this weekend (hopefully). Could you rebase on latest master and resolve the conflicts first?

immi-h:
> I'll give your PR a read this weekend (hopefully). Could you rebase on latest master and resolve the conflicts first?

I'm not sure whether I'll find time for that today, but probably tomorrow (Sunday).

immi-h:
When rebasing, I had to resolve some conflicts, and there's one where I am not sure whether I might have introduced a new bug: In the master branch, LatexController::findSelectedTexElement() seems to lock and unlock the document. In my branch, LatexController::findSelectedTexElement() doesn't exist anymore, but I might have to lock and unlock the document in other places instead. Do you know what this locking is about? (If not, I'll try to find out and whether I have to do it.)

Edit: I checked what this locking is doing; I wasn't aware that several threads might be running in parallel; so far, I paid no attention at all to making my code multithread-safe. I'll fix this.

src/core/control/LatexController.cpp
```diff
@@ -337,8 +268,25 @@ void LatexController::insertTexImage() {
...
 
+    Layer* layer = page->getSelectedLayer();
+    XournalView* xournal = this->control->getWindow()->getXournal();
+    auto pageNr = xournal->getCurrentPage();
+    auto* view = xournal->getViewFor(pageNr);
+
+    if (view->getPage() != page) {
+        g_warning("Active page changed while you edited the tex code. Aborting.");
+        return;
+    }
```
bhennion, line 280: This accesses the document structure, only for reading -> we should get a shared_lock of the mutex for that:

```suggestion
    auto lock = std::shared_lock(*this->control->getDocument());
    Layer* layer = page->getSelectedLayer();
    XournalView* xournal = this->control->getWindow()->getXournal();
    auto pageNr = xournal->getCurrentPage();
    auto* view = xournal->getViewFor(pageNr);

    if (view->getPage() != page) {
        g_warning("Active page changed while you edited the tex code. Aborting.");
        return;
    }
    lock.unlock().
```

src/core/control/LatexController.cpp
```diff
@@ -354,14 +302,53 @@ void LatexController::cancelEditing() {
...
     if (!depStatus.success) {
         XojMsgBox::showErrorToUser(ctrl->getGtkWindow(), depStatus.errorMsg);
         return;
     }
 
-    self->findSelectedTexElement();
+    self->page = page;
+
+    // Is there already a teximage at the click location? If yes, find the most recent one.
+    self->selectedElem = nullptr;
```
bhennion, line 319: Same here: you need to hold a shared_lock for as long as the document or its descendents (e.g. Layer/Element) are used.

```suggestion
    self->selectedElem = nullptr;
    auto lock = std::shared_lock(*this->control->getDocument());
```

src/core/control/LatexController.cpp
```diff
@@ -354,14 +302,53 @@ void LatexController::cancelEditing() {
...
+        }
+
+        self->imgwidth = self->selectedElem->getElementWidth();
+        self->imgheight = self->selectedElem->getElementHeight();
+
+    } else {
+        self->posx = x;
+        self->posy = y;
+    }
+
```
bhennion: without forgetting to release the lock once we're done
```suggestion
   lock.unlock();
```

src/core/control/LatexController.cpp
```diff
@@ -93,66 +92,6 @@ auto LatexController::findTexDependencies() -> LatexController::FindDependencySt
...
-    this->view = this->control->getWindow()->getXournal()->getViewFor(pageNr);
-    if (view == nullptr) {
-        return;
-    }
-
-    // we get the selection
-    this->page = this->doc->getPage(pageNr);
-    this->layer = page->getSelectedLayer();
-
-    auto* tex = view->getSelectedTex();
```
bhennion, line 114: This doesn't seem to appear anymore, probably meaning that you removed (unknowingly?) a feature: before this PR, if you have selected a Tex or Text object and switch the tool to LaTex, the selected object gets edited.
If you removed that on purpose (or if you think it should be removed), could you please explain why? Otherwise, could you restore the behaviour?

bhennion (changes requested):
I put in some comments about locking the document mutex where it needs to be locked (I think).
See also the comment about pre-selected Tex/Text objects.

src/core/control/LatexController.cpp
immi-h: Oops, I indeed removed this feature unintentionally. I'll put it back in.

immi-h: Hmm, I just noticed that the text tool doesn't have this feature, i.e., if you select a text object and switch the tool to text, the selected object does not get edited.
So I'm now wondering: Should we drop that feature also for the latex tool, to keep things consistent?

Some more arguments why I'd now tend to drop that feature:
- If several latex objects are selected, then one of them gets edited; but it's intransparent to the user which one gets edited. (This could be fixed if editing only happens if a single latex/text object is selected and nothing else.)
- I think as a used I'd expect that switching the tool has no side effect. (Right now, I don't recall of any other drawing program where switching a tool would have such a side effect.)

For those reasons, I'd now tend to drop that feature. But that's not a strong opinion; I'm certainly also happy to put it back in, if that's desired (maybe in the above variant where it only happens if a single object is selected).

archisman-panigrahi: > Hmm, I just noticed that the text tool doesn't have this feature, i.e., if you select a text object and switch the tool to text, the selected object does not get edited.
So I'm now wondering: Should we drop that feature also for the latex tool, to keep things consistent?


Latex involves a lot of long expressions, and being able to edit it easily saves a lot of writing.

Please don't drop the feature.

Being able to ediy the text is also very helpful. Otherwise, to edit, you need to erase it and rewrite.

This would be a regression.

bhennion: > So I'm now wondering: Should we drop that feature also for the latex tool, to keep things consistent?

I think we can drop it indeed.

> Being able to ediy the text is also very helpful. Otherwise, to edit, you need to erase it and rewrite.

The point is that with this PR, Tex formula can be edited by selecting the Tex tool and clicking on a preexisting TexImage. That weird feature was, before this PR, the only way to edit a Tex formula, but not anymore.

immi-h:
I inserted the lock/unlocks you suggested and had another look over the code. I didn't see any other place either where things would need to be locked.

I also rebased once more to the latest version.

src/core/control/LatexController.cpp
bhennion: What you implemented here is not quite right. There was a reason why my suggestion used an `std::shared_lock`: if `view->getPage() != page`, the mutex never gets unlocked, leading to a deadlock down the road.

src/core/control/LatexController.cpp
```diff
@@ -354,14 +304,54 @@ void LatexController::cancelEditing() {
...
         XojMsgBox::showErrorToUser(ctrl->getGtkWindow(), depStatus.errorMsg);
         return;
     }
 
-    self->findSelectedTexElement();
+    self->page = page;
+
+    // Is there already a teximage at the click location? If yes, find the most recent one.
+    self->selectedElem = nullptr;
+    self->control->getDocument()->lock_shared();
```
bhennion: Same here: prefer using RAII wrapper such as `std::shared_lock` (we tend to only keep direct calls to (un)lock_shared() if the unlock is within a couple of lines from the lock)

bhennion (changes requested):
On top of the above two comments, you'll also need to apply clang-format (e.g. by running `git-clang-format master`)

src/core/control/LatexController.cpp
immi-h: Oops; I'll fix that as soon as possible (which might be the next weekend).

rolandlo:
Now that the URL link tool is merged, this PR also needs to be rebased on latest master. Let us know if you need help with that.

src/core/control/LatexController.cpp
immi-h: This should be fixed now. And I also re-run git-clang-format.

immi-h:
> Now that the URL link tool is merged, this PR also needs to be rebased on latest master. Let us know if you need help with that.

Thanks for the offer. I can first give it a try to do that myself - maybe this weekend. (Of course, if you or someone else find time to do that already before, I'm also fine with that. ;-) )

immi-h:
I did the rebase. I'm always a bit worried of making mistakes when rebasing, but things seem to be working.

One remark: Some class members of PageView are now inconsistent between the latex tool and the link tool: In both tools, on a mousedown event, a variable is set to true so that on the next mouseup event, editing is started. For the latex tool, I called that variable "inLatex" (since its meaning is somewhat similar to "inEraser"). For the link tool, the variable is called "startEditingOnButtonRelease", which maybe captures better what the variable means, but which may be confusing since that variable explicitly starts editing a link and nothing else. Maybe one should make this consistent at some point; on the other hand, maybe rather do that separately later; it seems strange to change a variable related to the link tool in a commit related to the latex tool.

Also, for the latex tool, I used a separate variable to start editing after a double click, whereas the link tool uses the same variable for both. It might make sense to make that consistent at some point; but again: maybe later?

immi-h:
I got a message telling me that the lua definition file is not up to date. I figured out that I had to run scripts/lua_def_file.py, but I don't really know what I was doing, i.e., if anything else needs to be adapted in the lua code, I didn't do that yet. (In particular, if lua needs to be aware of more stuff about the latex tool, that has not yet been implemented.)
Note also that in the master branch version lua_def_file.py, Tool_link was missing and that it was generated when I run the above script. If it was intentional that it was not there, I broke that.

rolandlo:
> I got a message telling me that the lua definition file is not up to date. I figured out that I had to run scripts/lua_def_file.py, but I don't really know what I was doing, i.e., if anything else needs to be adapted in the lua code, I didn't do that yet. (In particular, if lua needs to be aware of more stuff about the latex tool, that has not yet been implemented.) Note also that in the master branch version lua_def_file.py, Tool_link was missing and that it was generated when I run the above script. If it was intentional that it was not there, I broke that.

You're right that `Tool_link` is missing on the master branch. It wasn't detected because the Lua definition file check only runs when either      'src/core/plugin/luapi_application.h' or 'src/core/enums/generated/Action.NameMap.generated.h' has been changed, which was not the case.

bhennion:
> Maybe one should make this consistent at some point; on the other hand, maybe rather do that separately later; it seems strange to change a variable related to the link tool in a commit related to the latex tool.

It can be made uniform later. It is also not clear to me what the best version of this would be, so we can table this for now.

bhennion (commented):
The code looks good to me

bhennion (approved):

archisman-panigrahi:
Anything else we need here except rebasing to the latest master code?

immi-h:
I just did the rebase and also did one more round of trying whether everything works. While doing so, I noticed that one behaviour is a bit strange (and not coherent with the other tools): When you use the tex-tool to create a new tex object, then afterwards, that tex-object is selected even though the selected tool is still the tex-tool. Apart from being a bit strange, it's also a bit annoying, since I tend to click elsewhere in the document to unselect the object, but this then opens the tex editor again. So I'll fix that. (Or let me know if you think this should also be done later.)

bhennion:
> I just did the rebase and also did one more round of trying whether everything works. While doing so, I noticed that one behaviour is a bit strange (and not coherent with the other tools): When you use the tex-tool to create a new tex object, then afterwards, that tex-object is selected even though the selected tool is still the tex-tool. Apart from being a bit strange, it's also a bit annoying, since I tend to click elsewhere in the document to unselect the object, but this then opens the tex editor again. So I'll fix that. (Or let me know if you think this should also be done later.)

Here are my thoughts on this:

- It makes sense to me to allow the user to finely position the Tex object once it has been rendered. Having it selected is an easy way to allow that.
- You do want an easy way to unselect it though, once you're done moving it.

Maybe a solution would be to prevent the Tex tool from opening the editor if something is selected?

immi-h:
> > I just did the rebase and also did one more round of trying whether everything works. While doing so, I noticed that one behaviour is a bit strange (and not coherent with the other tools): When you use the tex-tool to create a new tex object, then afterwards, that tex-object is selected even though the selected tool is still the tex-tool. Apart from being a bit strange, it's also a bit annoying, since I tend to click elsewhere in the document to unselect the object, but this then opens the tex editor again. So I'll fix that. (Or let me know if you think this should also be done later.)
> 
> Here are my thoughts on this:
> 
>     * It makes sense to me to allow the user to finely position the Tex object once it has been rendered. Having it selected is an easy way to allow that.
> 
>     * You do want an easy way to unselect it though, once you're done moving it.
> 
> 
> Maybe a solution would be to prevent the Tex tool from opening the editor if something is selected?

Hmm, you're right, that seems even better. At first, I was afraid that this might cause problems because I thought this was the only situation where something is selected while the active tool is not the selection tool, but I now noticed that there are other situations when this happens, so I guess this should be fine.

Another question is then: Should it be reflected by the mouse cursor that clicking while something is selected doesn't create a new tex object?

Also: Given that this seems to need a bit of discussion, would it make sense to deal with that separately?

src/core/control/LatexController.cpp
```diff
@@ -337,31 +268,92 @@ void LatexController::insertTexImage() {
...
-    this->deleteOldImage();
+    if (this->selectedElem) {
+        auto sel = SelectionFactory::createFromElementOnActiveLayer(control, page, view, selectedElem);
+        view->getXournal()->deleteSelection(sel.release());
+        this->selectedElem = nullptr;
+    }
 
     control->getUndoRedoHandler()->addUndoAction(
             std::make_unique<InsertUndoAction>(page, layer, this->temporaryRender.get()));
+// this->temporaryRender.release() ???
```
```suggestion
```

src/core/control/LatexController.cpp
```diff
@@ -337,31 +268,92 @@ void LatexController::insertTexImage() {
...
 
     // Select element
     auto selection =
             SelectionFactory::createFromFloatingElement(control, page, layer, view, std::move(this->temporaryRender));
     view->getXournal()->setSelection(selection.release());
+
```
```suggestion
```

bhennion (approved):
As you wish. We can table that for now.
Only a couple of nitpicks and then we can merge this, I think.

bhennion:
Ah, and it seems you need to apply clang format: `git-clang-format master` should do it.

src/core/control/LatexController.cpp
immi-h: Oops, these things got in by mistake. Fixed.

bhennion:
Thanks a lot! Squashing and merging in 48 hours unless an objection is raised (or the CI fails...)

immi-h:
> Thanks a lot! Squashing and merging in 48 hours unless an objection is raised (or the CI fails...)

Yey, then I can start working on the next tex-tool related issues. :-)

Concerning
>     * It makes sense to me to allow the user to finely position the Tex object once it has been rendered. Having it selected is an easy way to allow that.
> 
>     * You do want an easy way to unselect it though, once you're done moving it.
> 
> Maybe a solution would be to prevent the Tex tool from opening the editor if something is selected?

I still like this idea, but there's also a situation where this might be disadvantageous, namely if somebody wants to quickly create lots of latex objects. Then, after each object created, they'll have to click twice to create the next one. So the question is, what's better: The current version or the version you suggest?

Btw: I guess here is not the best place to discuss this. Should I open a thread elsewhere? What would be the best place?

bhennion:
> Should I open a thread elsewhere? What would be the best place?

I think you can open an issue ticket, listing the problems you find are left. We can discuss them there.

bhennion:
> The current version or the version you suggest?

I still think that unselecting should not open the dialog. Clicking twice for chain-creating tex objects seems a minor drawback to me.

PR #6831, opened 2025-11-23
QPDF exports: Remove "Isolated" flag to page's overlay XObject.

### LINKED ISSUE #6829: PDF export hides underlying text when highlighting text selected via "Select Linear PDF Text"
### Operating System

Linux

### (Linux only) Distribution

EndeavourOS (Arch-based)

KDE Plasma

Wayland

### Installation Method

Manually installed nightly build placed in /usr/local/bin

### Xournal++ Version

1.2.8+dev

### libgtk Version

3.24.51

### Bug Description

When highlighting text using **Select Linear PDF Text**, the highlight renders correctly inside Xournal++ (semi-transparent, with the underlying text visible).  
However, after exporting the document to PDF, the highlight becomes opaque and completely covers the text underneath.

This makes the exported PDF difficult or impossible to read when using text-highlighting annotations.

### Expected Behaviour

Exported PDF should preserve transparency so the text underneath remains visible.

### Steps to Reproduce

1. Open a PDF in Xournal++.
2. Use **Tools → Select Linear PDF Text**.
3. Select a portion of text and apply a highlight.
4. Export the document to PDF.
5. Open the exported PDF in Chrome, Firefox, Edge, Okular, etc.

### Additional Context

_No response_

bhennion:
Fixes wrong render of highlighter strokes in resulting pdf. Fixes #6829

### DIFF
--- a/src/core/pdf/base/QPdfExport.cpp
+++ b/src/core/pdf/base/QPdfExport.cpp
@@ -99,7 +99,8 @@ bool QPdfExport::overlayAndSave(const fs::path& saveDestination, std::stringstre
         auto overlayPages = QPDFPageDocumentHelper(overlay).getAllPages();
         overlaysAsXObjects.reserve(overlayPages.size());
         for (auto&& p: overlayPages) {
-            overlaysAsXObjects.emplace_back(p.getFormXObjectForPage());
+            auto&& xobj = overlaysAsXObjects.emplace_back(p.getFormXObjectForPage());
+            xobj.getDict().getKey("/Group").removeKey("/I");
         }
 
         auto outputPages = QPDFPageDocumentHelper(background).getAllPages();


bhennion:
To understand why this is the right thing to do, read the description of the I flag, at the bottom of p.350 in the [PDF specs](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf).

rolandlo (changes requested):
That doesn't seem to fix the issue. On the document I tested
[standalone_test.pdf](https://github.com/user-attachments/files/23698939/standalone_test.pdf)
after highlighting some of the letters the exported document had those letters covered same as on master. `hexdiff`ing the exported PDFs I checked that the `/I true` flag is only set on master. It just doesn't help.

bhennion:
Really?? It does work for me. With what app are you opening the output PDF?

On my laptop, opening 
[standalone_test_out.pdf](https://github.com/user-attachments/files/23712314/standalone_test_out.pdf)

with Okular gives 
Copie d&#39;écran_20251124_103626

rolandlo:
Strange. Your exported file is displayed correctly, mine isn't (in either of Evince, Papers, Xournal++, Firefox, Inkscape, GIMP, also Xournal++).
[standalone_test_fixed.pdf](https://github.com/user-attachments/files/23713551/standalone_test_fixed.pdf)

Bildschirmfoto vom 2025-11-24 11-34-22

Here is how the xopp-file looks with the original background:
Bildschirmfoto vom 2025-11-24 11-35-54

My annotations:
```xml
<stroke tool="highlighter" color="#f6d32d7f" width="12.739426" fill="255" capStyle="butt">7.0238995 8.60645 21.459046 8.60645</stroke>
<stroke tool="highlighter" color="#f6d32d7f" width="12.739426" fill="255" capStyle="butt">71.996 8.60645 79.295347 8.60645</stroke>
```

bhennion:
Mmm... Interesting: on my laptop, your PDF file is displayed as it should on Okular, but not on Evince (on Evince, I get what you get). In fact, the same happens with my PDF here.
Copie d&#39;écran_20251124_115004

Both rely on poppler to display the PDF file, but evince/papers/inkscape on poppler-glib and okular on poppler-qt6. The difference may lie in those binding libraries (and from what I understand of the PDF specs, it may be a bug of poppler-glib. Not sure though).

rolandlo:
Yeah, I get the same results on my system. I just didn't notice the difference with your PDF at first sight.

bhennion:
> it may be a bug of poppler-glib

Actually, it is this bug:
https://gitlab.freedesktop.org/poppler/poppler/-/issues/992

rolandlo (approved):
Well in that case, it's the best we can do, I guess. So there is no reason to hold this fix back.

bhennion:
Thanks a lot for the testing and review!
For information: there is a draft MR to fix this poppler-glib bug: https://gitlab.freedesktop.org/poppler/poppler/-/merge_requests/1513

PR #6841, opened 2025-12-01
Rebase of #4772 (url link tool by CruzeBlade)

rolandlo:
This is a rebase of @CruzeBlade's url link tool from #4772 on current master. I have squashed all commits into one so to simplify the rebase process.
I will continue to make it ready for merging if that's what @CruzeBlade wants me to do.
It would be a pitty if this great feature was bit-rotting for longer.

@CruzeBlade Please tell me if you are fine with that. As I understood you are lacking time to finish work on this feature.


rolandlo:
I made the dialogs GTK4-compatible, which was the major reason the PR could not be merged previously.

rolandlo:
Now mostly everything works and is ready for the Gtk 4 switch. Touch-only devices are not supported yet. This should be handled in a separate PR together with the PDF selection tool.

Here are the things that I believe work (so far only tested on Linux; uncheck if something is broken):
- [x] Activate Link tool in the toolbar (any toolbar except minimal ones) and switch back to a different tool
- [x] Activate Link tool via menubar (keyboard shortcut CTRL+SHIFT+u)
- [x] Pressing via mouse or stylus into the document to open the link dialog for creating a link
- [x] Editing the link in the link dialog
- [x] Insert the link to the document via the OK button
- [x] Cancel the link edition via the Cancel button or via the x window button
- [x] Highlight a link by hovering over it (via mouse or stylus)
- [x] Select a link by clicking on it (via mouse or stylus)
- [x] Open a link by CTRL+clicking on it
- [x] Open a link by selecting it and clicking on the Link in the popover (via mouse or stylus)
- [x] Obtain an error message if the link is not valid
- [x] Copy/paste a selection containing links via clipboard
- [x] Clone a selection containing links by right clicking on it
- [x] Save/load xopp-file containing links
- [x] Export to PDF with clickable links
- [x] Undo/redo inserting a link
- [x] Undo/redo deleting a link
- [x] Undo/redo modifying a link
- [x] Move and scale a selection containing links
- [x] Change the font and font size for a link
- [x] Change the orientation (left, center, right) of a link
- [x] Change the zoom of the document while a link is selected
- [x] Cursor change when using the Link tool

rolandlo:
\action create-installers all

Photon89:
I just tested the Linux AppImage, works great for me!

searumblinguminari:
I've only tested the Windows package X64 x86_64.
It seems to work as intended.
However, I don't understand the correct behavior of "Change the orientation (left, center, right) of a link." It seems that no matter which I select, there is no change.

rolandlo:
> I've only tested the Windows package X64 x86_64. It seems to work as intended. However, I don't understand the correct behavior of "Change the orientation (left, center, right) of a link." It seems that no matter which I select, there is no change.

Thanks for testing. For the change of orientation: You will not see the change in the editor, but only in the link that is inserted into the document. Use a link text with multiple lines.

searumblinguminari:
> Thanks for testing. For the change of orientation: You will not see the change in the editor, but only in the link that is inserted into the document. Use a link text with multiple lines.

I was able to confirm that it works fine when I use multiple lines. Thank you.

Photon89:
In the long run it would be a nice to have radio buttons 

[  ] Use same color as text color
[  ] Use different color: (color chooser here)

But the new tool works very well as is already, so the idea shouldn't delay merging the PR!

rolandlo:
I have bumped the file format version (from 4 to 5), since documents containing links will be saved without links when saved in an older Xournal++ version.

A link element looks like this in a .xopp-file
```xml
<link align="0" font="Sans" size="12" x="117" y="116" color="#ff00ffff" url="https%3A%2F%2Fsbb.ch" text="Fahrplan"/>
```

rolandlo:
> In the long run it would be a nice to have radio buttons
> 
> [ ] Use same color as text color [ ] Use different color: (color chooser here)
> 
> But the new tool works very well as is already, so the idea shouldn't delay merging the PR!

Thanks for the suggestion. I agree. This will have to wait for another PR. The color info is already contained in the file format, so we don't run into troubles when allowing setting the color later.

rolandlo:
The code is now cleaned up and functional enough that I feel it is ready to be reviewed/merged after some squashing of commits.

There is one bug with popover arrows that I could not solve. Sometimes when switching between selection of two links I get popover arrows like this: 

linkpopover

It might very well be a Gtk bug. There are some bug reports on the Gtk issue tracker related to popover arrows, but I'm not sure which one of these might be relevant. I think it's not a dealbraker in any case.

rolandlo:
Rebased on master. Used git-blame-someone-else to make @CruzeBlade the author of 9dcc3ae5acc80b97fe7378efb2512b3e156da5b4. Made the placeholders consistent.

@bhennion This is up for review.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,183 @@
...
+                this->highlightPopover->show_all();
+                this->highlightPopover->popup();
+            }
+        } else if (e->getType() == ELEMENT_LINK && this->highlightPopover->getLink() == e.get()) {
+            Link* link = dynamic_cast<Link*>(e.get());
+            link->setHighlighted(false);
+            page->fireElementChanged(link);
+            GdkWindow* window = gtk_widget_get_window(view->getWidget());
+            GdkCursor* cursor = gdk_cursor_new_from_name(gdk_window_get_display(window), "hand2");
+            gdk_window_set_cursor(window, cursor);
```
bhennion: This should probably be done via our XournalppCursor

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,61 @@
...
+                          "<span size=\"smaller\"><i> Double click to edit. CTRL + click to open. </i></span>";
+        gtk_label_set_markup(this->label, str.c_str());
+    } else {
+        gtk_label_set_text(this->label, this->link->getUrl().c_str());
+    }
+}
+
+void LinkPopover::positionPopover() {
+    if (this->link) {
+        XojPageView* pageView = view->getViewFor(view->getCurrentPage());
```
bhennion: This looks like it could be dangerous, as the Document mutex is not locked.
But if we lock the mutex here, we'd go back to the issue that #7199 fixed.

One solution would be to simply hide the popover upon zoom change. I have no idea if that would impact a lot of use cases though.

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -732,6 +733,37 @@ void LoadHandler::parseTexImage() {
...
+    const char* sFont = LoadHandlerHelper::getAttrib("font", false, this);
+    double fontSize = LoadHandlerHelper::getAttribDouble("size", this);
+    double x = LoadHandlerHelper::getAttribDouble("x", this);
+    double y = LoadHandlerHelper::getAttribDouble("y", this);
+    const char* url = LoadHandlerHelper::getAttrib("url", false, this);
+    const char* text = LoadHandlerHelper::getAttrib("text", false, this);
+
+    this->link->setAlignment(align);
+
+    this->link->setText(std::string(g_uri_unescape_string(text, NULL)));
```
bhennion: The text could have non-ASCII characters and is UTF-8-encoded, so this should be made into a std::u8string

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -732,6 +733,37 @@ void LoadHandler::parseTexImage() {
...
+    this->link->setUrl(std::string(g_uri_unescape_string(url, NULL)));
```
bhennion: Either we make sure somewhere that the URL never contains non-ASCII characters, or this should also be a u8string

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -732,6 +733,37 @@ void LoadHandler::parseTexImage() {
     this->teximage->setText(string(imText, imTextLen));
 }
 
+void LoadHandler::parseLink() {
+    auto linkOwn = std::make_unique<Link>();
+    this->link = linkOwn.get();
```
bhennion: It doesn't look like LoadHandler::link is used outside of this function. It should probably be removed form the class members.

src/core/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -519,6 +519,8 @@ void PenInputHandler::actionLeaveWindow(InputEvent const& event) {
         if (!this->inputContext->getSettings()->getInputSystemDrawOutsideWindowEnabled()) {
             this->actionEnd(this->lastHitEvent);
         }
+    } else if (toolHandler->getToolType() == TOOL_LINK) {
```
bhennion: I'm a bit confused as to why this tool needs special handling here. What's different between this tool and other "singlePageTool"s?

src/core/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -86,6 +86,11 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
         cursor->updateCursor();
     }
 
+    if ((event.type == ENTER_EVENT || event.type == LEAVE_EVENT) &&
+        this->inputContext->getToolHandler()->getToolType() == TOOL_LINK) {
```
bhennion: Same question here.
Overall, I think the gui/inputdevices/* should be exempt of tool specific behaviours.

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,61 @@
...
+    gtk_popover_set_constrain_to(this->popover, GTK_POPOVER_CONSTRAINT_WINDOW);
+    GtkWidget* vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
+    this->label = GTK_LABEL(gtk_label_new(""));
+    gtk_box_pack_start(GTK_BOX(vbox), GTK_WIDGET(this->label), true, true, POPOVER_PADDING);
+    gtk_container_add(GTK_CONTAINER(this->popover), vbox);
+}
+
+LinkPopover::~LinkPopover() {
+    gtk_widget_destroy(GTK_WIDGET(this->label));
+    gtk_widget_destroy(GTK_WIDGET(this->popover));
```
bhennion: This is not GTK4-compatible.

Also, the label is owned by the box which is owned by the popover, so destroying the popover will destroy the label already. No need to do both.

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+void LinkPopover::show() { gtk_widget_show(GTK_WIDGET(this->popover)); }
+
+void LinkPopover::show_all() { gtk_widget_show_all(GTK_WIDGET(this->popover)); }
```
bhennion: I don't think this should be exposed. Calling GTK3's `gtk_widget_show_all()` should only be done once. I would just call it on the box inside the popover upon widget creation.

src/core/model/Link.cpp
```diff
@@ -0,0 +1,148 @@
...
+    link->url = this->url;
+    link->setColor(this->getColor());
+    link->x = this->x;
+    link->y = this->y;
+    link->width = this->width;
+    link->height = this->height;
+    link->snappedBounds = this->snappedBounds;
+    link->sizeCalculated = this->sizeCalculated;
+    link->highlighted = this->highlighted;
+    return std::move(link);
```
bhennion:
```suggestion
    return link;
```

src/core/gui/dialog/LinkDialog.cpp
```diff
@@ -0,0 +1,184 @@
...
+constexpr auto UI_FILE = "linkDialog.glade";
+constexpr auto UI_DIALOG_NAME = "linkDialog";
+
+void okButtonClicked(GtkButton* btn, LinkDialog* dialog) { dialog->okButtonPressed(btn); }
+void cancelButtonClicked(GtkButton* btn, LinkDialog* dialog) { dialog->cancelButtonPressed(btn); }
+void textChangedClb(GtkTextBuffer* buffer, LinkDialog* dialog) { dialog->textChanged(buffer); }
+void layoutToogledLeft(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::LEFT); };
+void layoutToogledCenter(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::CENTER); };
+void layoutToogledRight(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::RIGHT); };
+void urlPrefixChangedClb(GtkComboBoxText* source, LinkDialog* dialog) { dialog->urlPrefixChanged(source); };
```
bhennion:
```suggestion
static void okButtonClicked(GtkButton* btn, LinkDialog* dialog) { dialog->okButtonPressed(btn); }
static void cancelButtonClicked(GtkButton* btn, LinkDialog* dialog) { dialog->cancelButtonPressed(btn); }
static void textChangedClb(GtkTextBuffer* buffer, LinkDialog* dialog) { dialog->textChanged(buffer); }
static void layoutToogledLeft(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::LEFT); };
static void layoutToogledCenter(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::CENTER); };
static void layoutToogledRight(GtkToggleButton* source, LinkDialog* dialog) { dialog->layoutToggled(LinkAlignment::RIGHT); };
static void urlPrefixChangedClb(GtkComboBoxText* source, LinkDialog* dialog) { dialog->urlPrefixChanged(source); };
```

src/core/gui/dialog/LinkDialog.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+LinkDialog::LinkDialog(Control* control, std::function<void(LinkDialog*)> callbackOK,
+                       std::function<void()> callbackCancel):
+        callbackOK(std::move(callbackOK)), callbackCancel(std::move(callbackCancel)) {
+
+    Builder builder(control->getGladeSearchPath(), UI_FILE);
+
+    this->linkDialog = static_cast<xoj::util::raii::GtkWindowUPtr>(GTK_WINDOW(builder.get(UI_DIALOG_NAME)));
```
bhennion: using static_cast doesn't seem right here.
```suggestion
    this->linkDialog.reset(GTK_WINDOW(builder.get(UI_DIALOG_NAME)));
```

src/core/gui/dialog/LinkDialog.h
```diff
@@ -0,0 +1,92 @@
...
+
+class LinkDialog {
+public:
+    LinkDialog(Control* control, std::function<void(LinkDialog*)> callbackOK, std::function<void()> callbackCancel);
+    ~LinkDialog();
+
+public:
+    void preset(XojFont font, std::string text, std::string url, LinkAlignment layout = LinkAlignment::LEFT);
+    std::string getText();
+    std::string getURL();
```
bhennion, line 38: Same concern here about UTF-8. Those should be u8string

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void LinkPopover::popdown() { gtk_popover_popdown(this->popover); }
+
+bool LinkPopover::hasLink() { return (this->link != nullptr); }
+
+void LinkPopover::updateLabel(bool markup) {
+    if (markup) {
+        std::string url = this->link->getUrl();
+        std::string str = "<a href=\"" + url + "\"> " + url + "</a> \n" +
+                          "<span size=\"smaller\"><i> Double click to edit. CTRL + click to open. </i></span>";
```
bhennion: The explanation (without the markup tags) should be translatable.

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+void LinkPopover::positionPopover() {
+    if (this->link) {
+        XojPageView* pageView = view->getViewFor(view->getCurrentPage());
+        auto pos = pageView->getPixelPosition();
+        GdkRectangle rect{pos.x + int(this->link->getX() * pageView->getZoom()),
+                          pos.y + int(this->link->getY() * pageView->getZoom()),
+                          int(this->link->getElementWidth() * pageView->getZoom()),
+                          int(this->link->getElementHeight() * pageView->getZoom())};
```
bhennion: The int conversions should be controlled with a round_cast<int>(): it makes the semantics clearer.

src/core/gui/PageView.cpp
```diff
@@ -479,6 +492,13 @@ auto XojPageView::onButtonDoublePressEvent(const PositionInputData& pos) -> bool
                     this->xournal->setSelection(sel.release());
                 }
                 control->runLatex();
+            } else if (elemType == ELEMENT_LINK) {
+                this->xournal->clearSelection();
+                toolHandler->selectTool(TOOL_LINK);
+                toolHandler->fireToolChanged();
+                // Simulate a button double press; there's too many things that we
+                // could forget to do if we manually call startEditing
+                this->onButtonDoublePressEvent(pos);
```
bhennion: This looks like an infinite loop. Don't you mean to use `onButtonPressEvent`?

bhennion (changes requested):
Here is a partial review. I did not test it and I only read parts of the files.

rolandlo:
Thanks a lot for the careful review. I will address your suggestions and remarks and let you know when I'm done with it.

src/core/gui/inputdevices/PenInputHandler.cpp
rolandlo: The problem is that an action is started (via `actionStart(event)`) on a button press event but we don't get a button release event (calling `actionEnd(event)`), since the focus goes to the newly created LinkDialog first. This doesn't happen with other single page tools, except the latex tool once it is made into a proper tool
If we switch to a different tool, the action is still running. If we want to use the new tool we get warnings like this one (when using the mouse)
```term
(com.github.xournalpp.xournalpp:1010071): xopp-WARNING **: 20:50:11.829: MouseInputHandler: Double BUTTON_PRESS_EVENT on button 1
```
from
```C++
    // Trigger start of action when pen/mouse is pressed
    if (event.type == BUTTON_PRESS_EVENT) {
        if (pressedButton == BUTTON_NONE || !inputRunning) {
            pressedButton = event.button;
            this->actionStart(event);
        } else if (event.button == pressedButton) {
            g_warning("MouseInputHandler: Double BUTTON_PRESS_EVENT on button %u", event.button);
        }
        return true;
    }
```
in the MouseInputHandler.

I realized now that my hack to solve this doesn't quite work, in particular on Windows, where the stylus doesn't emit a leave event. If I switch from the link tool to the pen tool there I start drawing right away.

Do you know of a way how to handle that properly?

src/core/control/tools/LinkHandler.cpp
rolandlo: I agree. This is done.

src/core/gui/LinkPopover.cpp
rolandlo: I have switched to hiding the popover upon zoom change. Do I still need to lock the document mutex here? I thought we only lock the document mutex, when the document is used directly. Here we only use the page views.

src/core/control/xojfile/LoadHandler.cpp
rolandlo: Excuse my ignorance. Before I try to make text and url into std::u8strings. What can actually go wrong here? I have tested with non ASCII characters on Windows and haven't run into a problem, even when saving and loading. We don't iterate over the characters or do stuff like that with those strings. Do you have a way how to get a failure?

src/core/control/xojfile/LoadHandler.cpp
rolandlo: Done.

src/core/gui/dialog/LinkDialog.cpp
rolandlo: Done!

src/core/gui/LinkPopover.cpp
rolandlo: Done!

src/core/gui/LinkPopover.cpp
rolandlo: Done!

src/core/gui/PageView.cpp
rolandlo: Done! It wasn't an infinite loop because in the second run there was no selection (but the link tool was activated). There seems to be no harm in calling startEditing directly, so that's what is done now.

src/core/gui/inputdevices/PenInputHandler.cpp
rolandlo: Also without the special handling after closing the link editor it's not enough to press into the document again to create a new link. One has to click twice.

src/core/control/xojfile/LoadHandler.cpp
bhennion: You are right: nothing can go wrong as is. This is about semantics: if a string has type u8string, we (the devs) know for sure it is encoded in UTF-8.

The actual issues we can have happen when a string is converted. Most of the time, this is when we create a fs::path from a string, or when we send a string to a stream (typically std::cout/cerr - see #7211 for a bug report).

src/core/gui/inputdevices/PenInputHandler.cpp
bhennion: A possible solution would be to open the popover on the actionEnd event.
It would be like the buttons in most UI: the action is only triggered on mouse-up if mouse-down was also in the button (try wth the GTK buttons in our toolbar for instance).

src/core/gui/LinkPopover.cpp
bhennion: I think you don't need to lock the mutex. This is executed by the main thread, and the PageView vector is only modified in the main thread (I think, I hope).

It is also probably true of the Document itself, actually (it is only modified in the main thread), so we may only lock the mutex when we modify it in the main thread and when we access it in other threads (render/export/save). I'm not certain of that though, and this is a bit off topic.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                    doc->unlock();
+                    page->firePageChanged();
+                    const auto undo = control->getUndoRedoHandler();
+                    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, link));
+                },
+                []() {});
+        dialog.show(control->getGtkWindow());
+    } else {
+        this->linkElement->setHighlighted(true);
+        page->firePageChanged();
```
bhennion: Using firePageChanged here is an overkill. Only the area around the link should be rerendered.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+        auto dialog = xoj::popup::PopupWindowWrapper<LinkDialog>(
+                this->control,
+                [this, page = page](LinkDialog* dlg) {
+                    this->linkElement->setText(dlg->getText());
+                    this->linkElement->setUrl(dlg->getURL());
+                    this->linkElement->setAlignment(static_cast<PangoAlignment>(dlg->getLayout()));
+                    this->linkElement->setFont(dlg->getFont());
+                    this->linkElement->sizeChanged();
```
bhennion: I think LinkElement::setText/setFont/setAlignment should reset the flag directly, rather than exposing a function like sizeChanged. This flag should stay internal to Element.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                    this->linkElement->setHighlighted(false);
+                    page->firePageChanged();
```
bhennion: Using firePageChanged here is an overkill. Only the area around the link should be rerendered.

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,65 @@
...
+LinkPopover::LinkPopover(XournalView* view): view(view) {
+    this->popover.reset(GTK_POPOVER(gtk_popover_new(view->getWidget())), xoj::util::ref);
+    gtk_popover_set_modal(this->popover.get(), false);
+    gtk_popover_set_constrain_to(this->popover.get(), GTK_POPOVER_CONSTRAINT_WINDOW);
+    GtkWidget* vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
+    this->label = GTK_LABEL(gtk_label_new(""));
+    gtk_box_pack_start(GTK_BOX(vbox), GTK_WIDGET(this->label), true, true, POPOVER_PADDING);
+    gtk_container_add(GTK_CONTAINER(this->popover.get()), vbox);
+    gtk_widget_show_all(GTK_WIDGET(vbox));
+    g_signal_connect(this->label, "activate-link", G_CALLBACK(+[](GtkLabel*, gchar* uri, gpointer) {
```
bhennion: This is not GTK4-compatible

src/core/gui/LinkPopover.h
```diff
@@ -0,0 +1,47 @@
...
+class LinkPopover {
+public:
+    LinkPopover(XournalView* view);
+    void hide();
+    void show();
+    void popup();
+    void popdown();
+    void updateLabel(bool markup);
+    void positionPopover();
+    bool hasLink();
```
bhennion:
```suggestion
    bool hasLink() const;
```

src/core/gui/inputdevices/PenInputHandler.cpp
rolandlo: Thanks for the advice! It solves the Windows-issue and removes the need for the hack in gui/inputdevices*. Now what do I with the double press (with or without selection) for reopening the link editor/editing the link? Will I need to move that to the second button release event?

rolandlo: This seems to work fine now. I have deferred reopening the link editor to the button release event following the double button press event.

rolandlo:
There are still some rendering artifacts with highlights of links not going away. I will investigate and try to fix it.

rolandlo:
> There are still some rendering artifacts with highlights of links not going away. I will investigate and try to fix it.

This is fixed now. Those weren't rendering artifacts but missing link unhighlighting. It seems to work all very smoothly now.

rolandlo:
Rebased on master.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                    link->setUrl(dlg->getURL());
+                    link->setAlignment(static_cast<PangoAlignment>(dlg->getLayout()));
+                    link->setFont(dlg->getFont());
+                    link->setTextPos(x, y);
+                    Document* doc = control->getDocument();
+                    doc->lock();
+                    const auto layer = page->getSelectedLayer();
+                    layer->addElement(std::move(linkOwn));
+                    doc->unlock();
+                    page->firePageChanged();
```
bhennion: firePageChanged is overkill here

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                XojMsgBox::openURL(nullptr, link->getUrl().c_str());
+                return;
+            }
+
+            if (!link->isSelected()) {
+                if (link->isHighlighted()) {
+                    this->highlightPopover->hide();
+                }
+                link->setSelected(true);
+                // FixMe !!
```
bhennion: What is to be fixed here?

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                },
+                []() {});
+        dialog.show(control->getGtkWindow());
+    } else {
+        this->linkElement->setHighlighted(true);
+        page->fireElementChanged(this->linkElement);
+        auto dialog = xoj::popup::PopupWindowWrapper<LinkDialog>(
+                this->control,
+                [this, page = page](LinkDialog* dlg) {
+                    this->linkElement->setText(dlg->getText());
```
bhennion: This callback should create a suitable undo action as well. Presumably, this means creating a new LinkElement from scratch, replacing the old one by the new and creating the corresponding GroupUndoAction.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,171 @@
...
+                    layer->addElement(std::move(linkOwn));
+                    doc->unlock();
+                    page->firePageChanged();
+                    const auto undo = control->getUndoRedoHandler();
+                    undo->addUndoAction(std::make_unique<InsertUndoAction>(page, layer, link));
+                },
+                []() {});
+        dialog.show(control->getGtkWindow());
+    } else {
+        this->linkElement->setHighlighted(true);
```
bhennion: I am a bit uneasy with this `setHighlighted()` function. We are storing in the Document something that does not really concern the Document, but only how it is rendered.
In particular, the Document instance is modified every time while the mutex is not locked. This operation being simply setting a boolean, it is most likely safe (as long as it is only called by the main thread). It still seems to be breaking the Model-View-Controller paradigm.

Also, it means rerendering part of the document's buffer without actual modifications. This has a cost.

src/core/gui/inputdevices/StylusInputHandler.cpp
```diff
@@ -86,7 +86,6 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
         cursor->updateCursor();
     }
 
-
```
bhennion, line 89: Could you revert this isolated change?

src/core/gui/LinkPopover.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "util/safe_casts.h"  // for round_cast
+
+LinkPopover::LinkPopover(XournalView* view): view(view) {
+    this->popover = GTK_POPOVER(gtk_popover_new(view->getWidget()));
+    gtk_popover_set_modal(this->popover, false);
+    gtk_popover_set_constrain_to(this->popover, GTK_POPOVER_CONSTRAINT_WINDOW);
+    GtkWidget* vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0);
+    this->label = GTK_LABEL(gtk_label_new(""));
+    gtk_box_pack_start(GTK_BOX(vbox), GTK_WIDGET(this->label), true, true, POPOVER_PADDING);
+    gtk_container_add(GTK_CONTAINER(this->popover), vbox);
```
bhennion: Could you use `gtk_box_append()`?

src/core/control/tools/LinkHandler.cpp
bhennion: Same with `setSelected()`, btw.
As those flags only add an underlining, I think it could be done via an OverlayView.

src/core/gui/LinkPopover.cpp
rolandlo: I added a code path for Gtk 4.

src/core/model/Link.cpp
rolandlo: Done

src/core/control/tools/LinkHandler.cpp
rolandlo: Done!

src/core/control/tools/LinkHandler.cpp
rolandlo: Now as the whole link is replaced we don't need to reset the flag any more. Or should we just for case someone uses the method in a situation when no new Link element is created?

src/core/control/tools/LinkHandler.cpp
rolandlo: Done

src/core/gui/LinkPopover.cpp
rolandlo: I made it GTK4-compatible, adding `gtk_popover_set_autohide` to the gtk4 helper..

src/core/gui/LinkPopover.h
rolandlo: Done

src/core/control/tools/LinkHandler.cpp
rolandlo: Done

src/core/control/tools/LinkHandler.cpp
rolandlo: I was referring to [this issue](https://github.com/xournalpp/xournalpp/pull/6841#issuecomment-3694575525) which I believe lies in Gtk. The comment is removed now.

src/core/control/tools/LinkHandler.cpp
rolandlo: Done

src/core/control/tools/LinkHandler.cpp
rolandlo: The highlighting is now achieved via an OverlayView.

src/core/gui/inputdevices/StylusInputHandler.cpp
rolandlo: Done

src/core/gui/LinkPopover.cpp
rolandlo: I did.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+                [this, linkElement, page = page](LinkDialog* dlg) {
+                    auto linkOwn = std::make_unique<Link>();
+                    Link* link = linkOwn.get();
+                    link->setText(dlg->getText());
+                    link->setUrl(dlg->getURL());
+                    link->setAlignment(static_cast<PangoAlignment>(dlg->getLayout()));
+                    link->setFont(dlg->getFont());
+                    link->setX(linkElement->getX());
+                    link->setY(linkElement->getY());
+                    page->fireElementChanged(link);
```
bhennion: This call looks off. The element has not yet been added/removed at this point

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+                    auto groupUndoAction = std::make_unique<GroupUndoAction>();
+                    auto deleteUndoAction = std::make_unique<DeleteUndoAction>(page, false);
+
+                    Document* doc = control->getDocument();
+                    doc->lock();
+                    const auto layer = page->getSelectedLayer();
+                    layer->addElement(std::move(linkOwn));
+                    auto [orig, elementIndex] = layer->removeElement(linkElement);
+                    doc->unlock();
+                    page->fireElementChanged(linkElement);  // rerender region around previous element
```
bhennion: You are calling `page->fireElementChanged(link)` below as well. It's probably faster/better to call `page->fireRangeChanged()` on the union of the two link's boxes.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,221 @@
...
+                this->highlightPopover->popup();
+            }
+        } else if (e->getType() == ELEMENT_LINK && isHighlighted(dynamic_cast<Link*>(e.get()))) {
+            view->getControl()->getCursor()->setIsLinkHighlighted(false);
+            if (!isSelected(this->highlightPopover->getLink())) {
+                this->highlightPopover->hide();
+            }
+            this->highlightPopover->linkTo(nullptr);
+        }
+    }
```
bhennion: I think the intent of this code section is made unclear by the implementation. Am I correct in understanding this as

1. We want to find out if the pointer is over a link element
2. We want to update the popover's `linkTo(...)` target.

If my understanding is correct, then only 1. requires a loop, so 2. should be moved out of the loop, and the loop should stop once a Link element has been found.

src/core/gui/PageView.cpp
```diff
@@ -648,6 +666,25 @@ void XojPageView::deleteView(xoj::view::OverlayView* view) {
...
 auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
+    double zoom = xournal->getZoom();
+    auto x = round_cast<int>(pos.x / zoom);
+    auto y = round_cast<int>(pos.y / zoom);
+
+    if (this->startEditingOnButtonRelease) {
+        this->startEditingOnButtonRelease = false;
+        if (x < 0 || y < 0) {
+            return false;
+        }
```
bhennion: Is there a specific reason for that test? It looks pretty arbitrary to test this here

src/core/gui/PageView.cpp
```diff
@@ -669,6 +706,12 @@ auto XojPageView::onButtonReleaseEvent(const PositionInputData& pos) -> bool {
     } else if (auto tt = control->getToolHandler()->getToolType();
                this->laserPointer && (tt == TOOL_LASER_POINTER_PEN || tt == TOOL_LASER_POINTER_HIGHLIGHTER)) {
         this->laserPointer->onButtonReleaseEvent(pos, xournal->getZoom());
+    } else if (control->getToolHandler()->getToolType() == TOOL_LINK) {
+        if (x < 0 || y < 0) {
+            return false;
+        }
```
bhennion: Is there a specific reason for that test? It looks pretty arbitrary to test this here

src/core/model/Element.h
```diff
@@ -72,6 +72,7 @@ class Element: public Serializable {
     virtual bool intersectsArea(double x, double y, double width, double height) const;
     /// Returns the distance between the element "as drawn" and the point (x,y)
     virtual double distanceTo(double x, double y) const;
+    virtual bool containsPoint(double x, double y) const;
```
bhennion: The naming could be more explicit (so that dev would understand that Stroke::containsPoint() does NOT tell you that the stroke contains the point). Perhaps some explicit mention of the bounding box.

Also, there in the code other places where this function should be used to replace weird snippets like 
```
            GdkRectangle matchRect = {gint(x), gint(y), 1, 1};
            if (e->intersectsArea(&matchRect)) {
```
May be done in a later PR though.

src/core/model/Element.cpp
```diff
@@ -111,6 +111,12 @@ auto Element::distanceTo(double x, double y) const -> double {
     return std::hypot(x - projX, y - projY);
 }
 
+auto Element::containsPoint(double x, double y) const -> bool {
+    bool horizontal = x >= getX() && x <= getX() + getElementWidth();
+    bool vertical = y >= getY() && y <= getY() + getElementHeight();
+    return horizontal && vertical;
```
bhennion: Because of lazy boolean evaluation, the (aknowledgely less readable) one-liner would be faster here:

```suggestion
    return x >= getX() && x <= getX() + getElementWidth() && y >= getY() && y <= getY() + getElementHeight();
```
Even better would be to ensure that sizeCalculated == true only once.

bhennion:
You'll also need to rebase on latest once #6670 has been merged

rolandlo:
> You'll also need to rebase on latest once #6670 has been merged

Done! I was a unsure about what to do with the link alignment enum that needed to be made compatible with the LoadHandler. I put it into Link.h, since it is used in the Link class and LinkDialog class mostly.

src/core/model/Element.h
rolandlo: I renamed it to `hasBoundingBoxContaining`, haven't used it in other places though and would like to defer that to another PR.

src/core/gui/PageView.cpp
rolandlo: I didn't find a clear reason, so I removed it.

src/core/gui/PageView.cpp
rolandlo: I didn't find a clear reason, so I removed it.

src/core/control/tools/LinkHandler.cpp
rolandlo: I have simplified all of `startEditing`, `highlight` and `select`. It should be more readable now and still show the same behavior.

src/core/control/tools/LinkHandler.cpp
rolandlo: Done!

src/core/control/tools/LinkHandler.cpp
rolandlo: Done! See below.

rolandlo:
One thing I'm unsure about is the way that links are saved in the underlying xml of the xopp-file. Currently it's:

```xml
<text font="Sans" size="18" x="123" y="123" color="#000000ff">Hello Text</text>
```
for text versus
```xml
<link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io" text="Hello%20Link"/>
```
for link.

So we have a text attribute for link element but not so for a text element.

Would it be better to store this as
```xml
<link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io">Hello Link</link>
```

rolandlo:
Closes #4732

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+    for (auto&& e: page->getSelectedLayer()->getElements()) {
+        if (e->getType() == ELEMENT_LINK && e->hasBoundingBoxContaining(x, y)) {
+            return dynamic_cast<Link*>(e.get());
+        }
+    }
+    return nullptr;
+}
+
+void LinkHandler::startEditing(const PageRef& page, const int x, const int y) {
+    Link* linkElement = findLinkAtPos(page, x, y);
```
bhennion, line 71: We should lock the document mutex for this (as a shared_lock)

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+                this->control,
+                [this, linkElement, page = page](LinkDialog* dlg) {
+                    auto linkOwn = std::make_unique<Link>();
+                    Link* link = linkOwn.get();
+                    link->setText(dlg->getText());
+                    link->setUrl(dlg->getURL());
+                    link->setAlignment(dlg->getLayout());
+                    link->setFont(dlg->getFont());
+                    link->setX(linkElement->getX());
+                    link->setY(linkElement->getY());
```
bhennion: I'm not sure about this, but better safe than sorry I guess:
When this callback is called, the linkElement may have been destroyed. I think we should record the original position before the dialog pops up (and while the doc mutex is locked) and never dereference linkElement in the callback.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+                    const auto undo = control->getUndoRedoHandler();
+                    auto groupUndoAction = std::make_unique<GroupUndoAction>();
+                    auto deleteUndoAction = std::make_unique<DeleteUndoAction>(page, false);
+
+                    Document* doc = control->getDocument();
+                    doc->lock();
+                    const auto layer = page->getSelectedLayer();
+                    layer->addElement(std::move(linkOwn));
+                    auto [orig, elementIndex] = layer->removeElement(linkElement);
```
bhennion, line 122: Note that this use of linkElement is fine: the pointer acts as an identifier and is never dereferenced.

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+                },
+                [this]() { this->highlightPopover->linkTo(nullptr); });
+        dialog.getPopup()->preset(linkElement->getFont(), linkElement->getText(), linkElement->getUrl(),
+                                  linkElement->getAlignment());
+        dialog.show(control->getGtkWindow());
+    }
+}
+
+void LinkHandler::select(const PageRef& page, const int x, const int y, const bool controlDown, XojPageView* pageView) {
+    Link* link = findLinkAtPos(page, x, y);
```
bhennion, line 151: Same here: lock the mutex

src/core/control/tools/LinkHandler.cpp
```diff
@@ -0,0 +1,203 @@
...
+        if (isHighlighted(link)) {
+            this->highlightPopover->hide();
+        }
+        this->selectPopover->linkTo(link);
+        this->selectPopover->popup();
+    }
+}
+
+void LinkHandler::highlight(const PageRef& page, const int x, const int y, XojPageView* pageView) {
+    Link* link = findLinkAtPos(page, x, y);
```
bhennion, line 178: and here

src/core/gui/dialog/LinkDialog.h
```diff
@@ -0,0 +1,94 @@
...
+    inline GtkWindow* getWindow() const { return linkDialog.get(); }
+
+public:
+    static constexpr int DEFAULT_HEIGHT = 180;
+    static constexpr float MAX_HEIGHT_RATIO = 0.75;
+    static constexpr int INITIAL_NUMBER_OF_LINES = 2;
+
+    static constexpr char PLACEHOLDER_HTTPS[] = "xournalpp.github.io";
+    static constexpr char PLACEHOLDER_MAIL[] = "email-address@provider.domain";
+    static constexpr char PLACEHOLDER_FILE[] = "/absolute/path/to/file";
```
bhennion, line 95: Should this be a typical Windows path on Windows?

bhennion (commented):
Here are a couple of comments.

bhennion:
> One thing I'm unsure about is the way that links are saved in the underlying xml of the xopp-file. Currently it's:
> 
> ```
> <text font="Sans" size="18" x="123" y="123" color="#000000ff">Hello Text</text>
> ```
> 
> for text versus
> 
> ```
> <link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io" text="Hello%20Link"/>
> ```
> 
> for link.
> 
> So we have a text attribute for link element but not so for a text element.
> 
> Would it be better to store this as
> 
> ```
> <link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io">Hello Link</link>
> ```

Do both support the same situations? Typically, are `\n`'s supported in the current link version?

It would be good to add those to the test cases. Including cases with utf8 characters in the displayed name and corner cases like that.

src/core/gui/dialog/LinkDialog.h
rolandlo: Even on Windows it must start with a slash. Afterwards it doesn't matter if slashes or backslashes are used, but something like 
```term
/C:\Users\Joe\Documents\pic.jpg
``` 
looks strange, even though it works just like 
```term
/C:/Users/Joe/Documents/pic.jpg
```

Not sure how to communicate that to Windows users.

bhennion: > Even on Windows it must start with a slash

Why is that?

rolandlo: The general format for file URI is `file://host/path`, 
> where host is the fully qualified domain name of the system on which the path is accessible, and path is a hierarchical directory path of the form directory/directory/.../name. If host is omitted, it is taken to be "localhost", the machine from which the URL is being interpreted.

See https://en.wikipedia.org/wiki/File_URI_scheme.

bhennion: I see. Then I guess something like `/C:\foo\bar` would be better already.
We may also prepend the `/` ourselves (like adding it to the prefix, on Windows).

rolandlo: I thought about prepending `/` on Windows, too, but that excludes network locations like
```term
file://server/folder/data.xml
```
for those who are aware of it. I guess
```term
/C:/foo/bar
```
is then better, since it's just a typical example and doesn't look as strange as mixing slashes and backslashes.

rolandlo:
> > One thing I'm unsure about is the way that links are saved in the underlying xml of the xopp-file. Currently it's:
> > ```
> > <text font="Sans" size="18" x="123" y="123" color="#000000ff">Hello Text</text>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for text versus
> > ```
> > <link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io" text="Hello%20Link"/>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for link.
> > So we have a text attribute for link element but not so for a text element.
> > Would it be better to store this as
> > ```
> > <link align="0" font="Sans" size="12" x="123" y="123" color="#ff00ffff" url="https%3A%2F%2Fxournalpp.github.io">Hello Link</link>
> > ```
> 
> Do both support the same situations? Typically, are `\n`'s supported in the current link version?
> 
> It would be good to add those to the test cases. Including cases with utf8 characters in the displayed name and corner cases like that.

I think that both would support the same situations. Newlines are definitely supported in the current link versions (there would be no point in aligning links to the left, center or right otherwise), but they are stored as:
```xml
<link ... text="Multiline%0ALink"/>
```
so the newline character is escaped, which wouldn't be the case with the other form.

bhennion:
> I think that both would support the same situations.

Then both are good, I guess.

bhennion:
It all looks good to me, with the following reserve: it would be better to extend the load/save and ser/des test units to also test for this new element type.

rolandlo:
> It all looks good to me, with the following reserve: it would be better to extend the load/save and ser/des test units to also test for this new element type.

Done! I also fixed that the link alignment wasn't serialized/deserialized.

rolandlo:
> > I think that both would support the same situations.
> 
> Then both are good, I guess.

I have switched to using the link text as content rather than an attribute. The main reason is consistency: for all 4 element types we had there was a content in the xml and the content was what was really displayed (text type: the text, image type: the image, teximage type: the image of the tex formula, stroke: the stroke points). Another (although less important) reason is that for attributes there is more escaping needed, which makes it less readable.

I have also removed the use of `g_uri_escape_string` and `g_uri_unescape_string` for saving/loading. There is no reason to escape text and url just for saving them in the file.

rolandlo:
@bhennion Please check the last 4 commits. I'm now settled with the way links are represented in the xml. This is crucial, since it would be hard to change later on.

test/unit_tests/control/LoadHandlerTest.cpp
```diff
@@ -786,3 +787,32 @@ TEST(ControlLoadHandler, testRelativePath) {
...
+
+    auto elements = layer->getElementsView();
+    ASSERT_EQ((size_t)4, layer->getElementsView().size());
+
+    auto check_link = [&](size_t i, const auto* text, const auto* url) {
+        EXPECT_EQ(ELEMENT_LINK, elements[i]->getType());
+        auto* link = dynamic_cast<const Link*>(elements[i]);
+        ASSERT_NE(link, nullptr);
+        EXPECT_STREQ(link->getText().c_str(), char_cast(text));
+        EXPECT_STREQ(link->getUrl().c_str(), char_cast(url));
```
bhennion, line 809: Shouldn't alignment be check here as well?

bhennion (commented):
One small thing. Other than that, LGTM!
I agree that having the link as content feels more appropriate (without any hard argument though).

bhennion (approved):

test/unit_tests/control/LoadHandlerTest.cpp
rolandlo: Yes, indeed. I have added that.

rolandlo:
Merging in 48 hours if no objections are raised (and the checks are successful).
Thanks a lot for all the review @bhennion

rolandlo:
The URL link tool is finally merged. Thanks @CruzeBlade for all the initial work on the feature.

PR #6856, opened 2025-12-07
Update README with packaging status badge from repology

archisman-panigrahi:
Add repology packaging status badge for Xournal++ so that packages in official repos can be found at a glance


archisman-panigrahi:
Here is how it looks like after the addition

image

PR #6869, opened 2025-12-11
Update copyright file

rolandlo:
Updated the `copyright.txt` The compary-licenses script now succeeds.

I wasn't sure how to attribute the LatexDialog related files, since it was a mess. 
AbstractLatexDialog from #6476 is mostly a part of the former LatexDialog, so it might make more sense to attribute it (the .h and .cpp file) to Wilson Brenna and put corresponding notes into both files.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,7 +10,7 @@ Developer Team:
 2019-2019, Justin Jones <bjustjones@netscape.net>
 2019-2023, Bryan Tan <Technius@users.noreply.github.com>
 2019-2019, Justus Rossmeier <veecue@users.noreply.github.com>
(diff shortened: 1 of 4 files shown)


bhennion:
> I wasn't sure how to attribute the LatexDialog related files, since it was a mess.

I think it is ok. The file explicitly mentions the github commits as the source of truth, and from there, people can sort it out if required.

rolandlo:
> > I wasn't sure how to attribute the LatexDialog related files, since it was a mess.
> 
> I think it is ok. The file explicitly mentions the github commits as the source of truth, and from there, people can sort it out if required.

Ok, I merged it into `release-1.3`. Thanks for taking a look.

Issue #6879, opened 2025-12-12
Clipboard pasting broken with automatic DPI detection


rolandlo:
### Operating System

Windows

### (Linux only) Distribution

Ubuntu

### (Linux only) Desktop Environment

GNOME

### (Linux Only) Display Server

Wayland

### Installation Method

PPA

### Xournal++ Version

1.3.0

### libgtk Version

3.24.41

### Bug Description

With automatic DPI detection `displayDpi` (the return value of `getDisplayDpi`) is `-1`, which is used in inappropriate places like
https://github.com/xournalpp/xournalpp/blob/dde1c66be57634c5555a9990cfb5345ad8226ca5/src/core/control/Control.cpp#L2230-L2234
in `Control::clipboardPasteImage` (and other places).

### Expected Behaviour

Pasting from clipboard works even when automatic Dpi detection is enabled.

### Steps to Reproduce

1. Turn on automatic display Dpi detection
2. Make a screenshot of some part of the desktop and paste it into the document
3. Observe it doesn't work

bhennion:
I can work on a fix if you want.

PR #6887, opened 2025-12-13
Fix various occurrences of display DPI when using automated detection

bhennion:
In several places, we used settings->getDisplayDpi() to get the DPI. When automated detection is on, this returns -1 instead of the right value.
This causes several bugs (#6879, wrongly sized selection anchors and broken tap detection).

This PR fixes that, by replacing `Settings::getDisplayDpi()` by `ZoomControl::getZoom100Value() * Util::DPI_NORMALIZATION_FACTOR` almost everywhere.

I apologize for missing those predictable issues when automated detection was introduced.


bhennion:
Thanks for the review!

PR #6888, opened 2025-12-13
Fix GFilename conversions

bhennion:
Fixes the GFilename conversion issues that @febbe located in #6137.


bhennion:

bhennion:

PR #6889, opened 2025-12-13
fix(luapi): Preserve tool alpha when changing color via Lua API #6883

### LINKED ISSUE #6883: Setting Marker opacity in v1.3.0 is not permanent
### Operating System

Linux

### (Linux only) Distribution

Debian (Sid)

Gnome 48.5

Wayland

### Installation Method

Apt

### Xournal++ Version

xournalpp 1.2.10

### libgtk Version

libgtk: 3.24.51

### Bug Description

Hi !

I am using QPDF backend export to keep hyperlinks, so I think I have to build xournalpp myself.

```bash
mkdir build
cd build
cmake .. -DCMAKE_INSTALL_PREFIX=/home/jean/local/xournalpp -G Ninja -DENABLE_QPDF=yes -DENABLE_PLUGINS=yes
ninja
ninja install
```

When I use colors, for example with "Select linear PDF text", colors selected by the vi-xournalpp plugin are darker with self-built xournalpp than Debian xournalpp. Why ?

I am sorry, I don't know if this is a xournalpp problem or a vi-xournalpp plugin problem.

### Expected Behaviour

I expect color to be the same between Debian package and self-built xournalpp.

### Steps to Reproduce

```bash
sudo apt install xournalpp
cd /usr/share/xournalpp/plugins
sudo git clone https://github.com/raw-bacon/vi-xournalpp
```

then start ``/usr/bin/xournalpp``, activate the vi-xournalpp plugin and restart ``/usr/bin/xournalpp``

'c' 'b' to get the blue => the blue is ok to highlight text

Image

but done with the self-built version with the plugins installed in ~/local/xournalpp/share/xournalpp/plugins, the blue is too dark 

Image

### Additional Context

_No response_

youssefadly237:
~~Allow app.changeToolColor to accept both 0xRRGGBB and 0xAARRGGBB.
Alpha has to be explicitly provided to overwrite current.~~

`app.changeToolColor` fetches current alpha and apply it to the color to preserve alpha

closes #6883
closes lryffel/vi-xournalpp#48

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -1939,7 +1939,7 @@ static int applib_changeToolColor(lua_State* L) {
         color = static_cast<uint32_t>(as_unsigned(lua_tointeger(L, -1)));
         if (color > 0xffffff) {
             std::stringstream msg;
-            msg << "Color 0x" << std::hex << color << " is no valid RGB color.";
+            msg << "Color 0x" << std::hex << color << " is too large. Use 0xRRGGBB format.";
             return luaL_error(L, msg.str().c_str());  // luaL_error does not support %x for hex numbers
         }
     } else if (!lua_isnil(L, -1)) {
@@ -1951,7 +1951,9 @@ static int applib_changeToolColor(lua_State* L) {
     Tool& tool = toolHandler->getTool(toolType);
 
     if (tool.hasCapability(TOOL_CAP_COLOR)) {
-        tool.setColor(Color(color | 0xff000000U));
+        uint8_t currentAlpha = tool.getColor().alpha;
+        Color newColor = Color(color | (static_cast<uint32_t>(currentAlpha) << 24));
+        tool.setColor(newColor);
         ctrl->toolColorChanged();
         if (selection) {
             ctrl->changeColorOfSelection();


bhennion:
Most tools do not support transparency (it may accidentally work). Allowing plugins to set an alpha value for any tool will most likely bring in complications. We should be careful here.

youssefadly237:
I didn't test it I assumed all tools support alpha, I will check that

youssefadly237:
I think the implementation is safe because **rendering** code enforces opacity for tools that don't support transparency

since we are not using `gdk_cairo_set_source_rgba` directly instead a we are using a wrapper

https://github.com/xournalpp/xournalpp/blob/81ef756d26444cf526ab61498087aceffd9be0df/src/util/Util.cpp#L24-L32

Pen strokes uses default `alpha=1.0`, ignoring stored value

https://github.com/xournalpp/xournalpp/blob/81ef756d26444cf526ab61498087aceffd9be0df/src/core/view/StrokeView.cpp#L123-L129

Text tool, uses same pattern, default `alpha=1.0`

https://github.com/xournalpp/xournalpp/blob/81ef756d26444cf526ab61498087aceffd9be0df/src/core/view/TextView.cpp#L42-L44

Explicitly set

https://github.com/xournalpp/xournalpp/blob/81ef756d26444cf526ab61498087aceffd9be0df/src/core/view/overlays/BaseStrokeToolView.cpp#L24-L32

rolandlo:
Can you please rebase on `release-1.3` and switch the target branch on Github accordingly. This is a bugfix, not really a new feature.

youssefadly237:
Should I apply this fix only to `release-1.3`, or also to `master`

I mean nightly needs same fix too

rolandlo:
Only to `release-1.3`. It will get merged into master after some time. We can do so at any point of time we want, not only after releasing the next patch version.

bhennion:
> I think the implementation is safe because **rendering** code enforces opacity for tools that don't support transparency

That just means it is invisible. But if the color alpha gets saved in the files, then if one day we want to use that channel, we'll get users complaining.
So no, we should not allow it.

youssefadly237:
This is why I had rendering in bold 🫠

youssefadly237:
@bhennion how would you recommend I fix that ?

```cpp
            if (!xoj::tool::isPdfSelectionTool(toolType)) {
                return luaL_error(L, "tool \"%s\" does not support alpha channel. Use 0xRRGGBB format instead",
                                  toolTypeToString(toolType).data());
            }
```

this would work ?

or maybe drop the alpha channel from the api ?

bhennion:
> this would work ?

This looks good, yes!

youssefadly237:
this works but feels wrong, why it can not set the alpha for the highlighter for example

also, the current api methods for opacity open a small pop-up instead of letting you set a numeric value directly, which makes scripting or automated workflows awkward

should I implement `app.setToolAlpha` ?

also did any one asked for custom highlighter opacity before ?

bhennion:
> also did any one asked for custom highlighter opacity before ?

Yes it has been. There is some open ticket on this (possibly several). But this becomes a new feature and no longer a bug fix.

IMO, the thing to do is:
1. on release-1.3: simply fix the issue by fetching the tool's current color's alpha value and applying it to the given color (before feeding it to the tool). No new API function there.
2. on master: look carefully at what the features would be:
   1. Allow to configure the highlighter's alpha?
   2. Configure other elements' alpha? (probably more dangerous - I wouldn't go there just yet)
   3. Add Lua API to control this from a plugin?

Rehan-MALAK:
Hi @bhennion and @youssefadly237 

Thanks a lot for your time and your investigations.
I hope this solution
> fetching the tool's current color's alpha value and applying it to the given color

is simple enough and won't be too hard to implement.

In the meantime, I am working with this :

```diff
diff --git a/debian/rules b/debian/rules
index b82a24f6c..8cb074e1a 100755
--- a/debian/rules
+++ b/debian/rules
@@ -12,4 +12,4 @@ override_dh_dwz:
        dh_dwz --no-dwz-multifile
 
 override_dh_auto_configure:
-       dh_auto_configure -- -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_CPPTRACE=OFF
+       dh_auto_configure -- -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_CPPTRACE=OFF -DENABLE_QPDF=On
```

on top of @youssefadly237 branch [fix/plugin-preserve-alpha-on-changeToolColor](https://github.com/xournalpp/xournalpp/compare/master...youssefadly237:xournalpp:fix/plugin-preserve-alpha-on-changeToolColor)
 
I am using xournalpp everyday (and I really really don't want to lose url hyperlink during pdf export and a nice modal vi-style keybindings) so don't hesitate if you want me to test something.

Best,

youssefadly237:
@rolandlo do not forget to squash on merge

I will be opening another pr on master to add this fix alongside `app.setToolAlpha()`, `app.getToolAlpha`, and support for highlighter opacity

src/core/plugin/luapi_application.h
```diff
@@ -1960,9 +1962,11 @@ static int applib_changeToolColor(lua_State* L) {
         return luaL_error(L, "tool \"%s\" has no color capability", toolTypeToString(toolType).data());
     }
 
+    lua_pop(L, 3);
```
rolandlo: We agreed not to pop directly before return statement, since it is unnecessary. See https://github.com/xournalpp/xournalpp/pull/4837#issuecomment-1657824620

rolandlo:

Issue #4755, opened 2023-03-22
Better Typst support (it already works!)


rolandlo:
**Is your feature request related to a problem? Please describe.**
*LaTeX* is a great tool for typesetting math, but it has some shortcomings like a relative cumbersome and not very intuitive syntax syntax (compared to Markdown for example) and relatively long compile times. Also incrementally compiling LaTeX files is not possible.
[Typst](https://typst.app/) is a pretty recent effort to offer a better writing experience while being as powerful as LaTeX. The developers provide the following reasons why to use Typst instead of LaTeX:
> Typst...
    - previews your changes instantly.
    - provides clear, understandable error messages.
    - has a consistent styling system for configuring everything from fonts and margins to the look of headings and lists.
    - uses familiar programming constructs instead of hard-to-understand macros.

The LaTeX tool in Xournal++ can already been used with typst, even with a very simple template:
```
#set page(
width: auto,
height: auto,
margin: (x: .1cm, y: .1cm),
)

#set text(
fill: rgb("#%%XPP_TEXT_COLOR%%")
)

$ %%XPP_TOOL_INPUT%% $
```

To use it:
1) Download typst from https://github.com/typst/typst/releases and extract it where you want to find it
2) Store the above template in a file `typst.tex` (ending `.tex` so it gets displayed in the file chooser)
3) Set the template in the `LaTeX` tab of menu `Edit > Preferences`, and use LaTeX generator command: 
```
/path/to/typst/typst '{}'
``` 
4) Learn and use the typst syntax, if you haven't. It appears to be quite easy and intuitive. Two example (one for math, one for text):

![grafik](https://user-images.githubusercontent.com/40485433/226934167-a384198e-8934-4d12-a200-386783f1548a.png)

![grafik](https://user-images.githubusercontent.com/40485433/226938687-2c3ecb7e-9a75-4080-a837-fe2cb902eda4.png)
**Describe the solution you'd like**
The LaTeX tool should be generalized. Typst should get a separate button (Markdown, too) and it should be possible to set a template for each one separately. The dialogs should be rewritten and depend on the tool (LaTeX/Typst/Markdown...) in use.

Typst supports incremental updates and this should be supported out of the box in Xournal++ as well. It's pretty fast even without that, though.

**Describe alternatives you've considered**
Use typst as desribed above and don't mix typst and LaTeX in the same document.

trace1729:
[...]

PR #6905, opened 2025-12-17
Add UX typst initial support

tiziodcaio:
Simplify the LaTeX UI into something more simple to address simplification on #4755 UX.
image

It should be pretty stable and it's defaulting to the old implementation.

### DIFF
--- a/src/core/control/latex/LatexGenerator.cpp
+++ b/src/core/control/latex/LatexGenerator.cpp
@@ -95,18 +95,18 @@ auto LatexGenerator::templateSub(const std::string& input, const std::string& te
 }
 
 auto LatexGenerator::asyncRun(const fs::path& texDir, const std::string& texFileContents) -> Result {
-    // std::string cmd = this->settings.genCmd;
     GErrorGuard err{};
-    std::string texFilePathOSEncoding =
-            Util::GFilename(Util::getLongPath(texDir) / ("tex." + settings.genTmpFileExt)).c_str();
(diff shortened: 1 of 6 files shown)


tiziodcaio:
Right now you have to change manually the template. So it might not be so useful right now, but I hope to address that issue too soon.

rolandlo:
Thank you for working on this. We need a strategy for how to deal with different kind of formulas (say some are LaTeX formulas, other are typst formulas, yet other use [other generators](https://github.com/xournalpp/xournalpp/discussions/6909)) by the same user, maybe even in the same document. Have you thought about that already?

tiziodcaio:
Disclaimer: I will recognise it would be confusing talking about LaTeX settings when it would not be regarding LaTeX always, so we should call them "formula settings".

> Thank you for working on this. We need a strategy for how to deal with different kind of formulas (say some are LaTeX formulas, other are typst formulas, yet other use https://github.com/xournalpp/xournalpp/discussions/6909) by the same user, maybe even in the same document.

Oh I did not designed it thinking about it. Right now xournal saves the raw image format AND the original formula hoping that the user would be "smart enough" to not use different setups or different computers to generate all the formulas. That should not happen.
Maybe I can try setting these "formula settings" locally linked to the open file and remove them from the global settings page, but maybe it would give too much power to the author of the file . I have some UX ideas in my mind but I still need to polish a bit the rest of the PR.

> Have you thought about that already?

Right now I have some ideas. We can bundle in the xml's file some properties/metadata linked to the formula so you can tell how it was generated without losing everything

PR #6924, opened 2025-12-24
Merge back release 1.3.1

rolandlo:
To be pushed from CLI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,22 @@
 # Changelog
 
-## 1.3.0+dev (Unreleased)
+## 1.3.1+dev (Unreleased)
+
+## 1.3.1
+
+* Restored the rendering of graph backgrounds used before the 1.3.0 release (#6894)
+* Fixed image loading on Windows (#6893)
+* Fixed various issues when using automatic display DPI detection, like clipboard pasting not working (#6887)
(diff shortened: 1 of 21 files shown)


po/af_ZA.po
bhennion, line 1: The content of the po/ folder should not be merged (I guess using something like `git checkout --ours po/*` to resolve conflicts). Otherwise, the translation PRs will get conflicts.

bhennion (commented):
Just one thing about the translation files (see below). Other than that, LGTM!
Thanks for handling the 1.3.1 release!

po/af_ZA.po
rolandlo: Thanks for the review and for pointing out the issue with the po files, about which I was really in doubt. Now I helped myself with `git checkout HEAD~ -- po/*.po*` to restore the .po and .pot file(s).

The Windows pipeline failed because of the Gtk downgrade. It's not related to this PR, so I will merge.

bhennion: > The Windows pipeline failed because of the Gtk downgrade.

Do we still need this downgrade? (it was about a crash when changing keyboard layout/languages)

rolandlo: I have no idea. This needs to be tested.

PR #6925, opened 2025-12-24
fix: don't reconvert absolute paths to absolute ones

Nerixyz:
This fixes Windows users not being able to open documents from UNC paths (https://github.com/xournalpp/xournalpp/issues/5636#issuecomment-3690156180).

Like the original "fix" for #5636, this is rather hacky. With libstdc++ on Windows, UNC paths are not considered absolute. So when calling `fs::absolute`, a UNC path will get corrupted (a drive letter will be prepended, making it a regular path). If we check for the path already being absolute, we can avoid corrupting the path.

The proper solution would be to fix this in libstdc++, but my patch didn't receive any review since May. An alternative is to use libc++ instead of libstdc++. It implements the path operations on UNC paths correctly, but switching the STL opens another can of worms.

### DIFF
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -11,3 +11,19 @@ target_compile_features(util PUBLIC ${PROJECT_CXX_FEATURES})
 target_compile_features(util PUBLIC cxx_std_20)
 target_include_directories(util PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
 add_library(xoj::util ALIAS util)
+
+# Check if std::filesystem::absolute needs to be wrapped.
+# With libstdc++, there's a bug where it corrupts UNC paths.
+if (WIN32)
+    include(CheckCXXSymbolExists)
+    check_cxx_symbol_exists(__GLIBCXX__ version GLIBCXX)
+    if(GLIBCXX)
+        target_link_options(util PUBLIC
+            # std::filesystem::absolute(std::filesystem::__cxx11::path const&)
+            "-Wl,--wrap=_ZNSt10filesystem8absoluteERKNS_7__cxx114pathE"
+            # std::filesystem::weakly_canonical(std::filesystem::__cxx11::path const&)
+            "-Wl,--wrap=_ZNSt10filesystem16weakly_canonicalERKNS_7__cxx114pathE"
+        )
+        target_compile_definitions(util PUBLIC XOURNALPP_WRAP_STD_FS_ABSOLUTE)
+    endif()
+endif()
--- a/src/util/PathUtil.cpp
+++ b/src/util/PathUtil.cpp
@@ -149,25 +149,6 @@ auto Util::hasPngFileExt(const fs::path& path) -> bool {
     return StringUtils::toLowerCase(path.extension().string()) == ".png";
 }
 
-auto Util::isAbsolute(const fs::path& path) -> bool {
-#if defined(__GLIBCXX__) && defined(_WIN32)
-    if (path.is_absolute()) {
-        return true;
-    }
-
-    // Check for UNC paths like \\server\foo\bar.
-    // This works around a libstdc++ bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99333
-    const auto& str = path.native();
-    if (str.size() < 3) {
-        return false;
-    }
-    // support both forward and backward slashes
-    return (str[0] == '\\' || str[0] == '/') && (str[1] == '\\' || str[1] == '/');
-#else
-    return path.is_absolute();
-#endif
-}
-
 auto Util::clearExtensions(fs::path& path, const std::string& ext) -> void {
     auto rm_ext = [&path](const std::string ext) {
         if (StringUtils::toLowerCase(path.extension().string()) == StringUtils::toLowerCase(ext)) {
@@ -258,7 +239,7 @@ auto Util::fromUri(const std::string& uri) -> std::optional<fs::path> {
 auto Util::toUri(const fs::path& path) -> std::optional<std::string> {
     GError* error{};
     char* uri = [&] {
-        if (isAbsolute(path)) {
+        if (path.is_absolute()) {
             return g_filename_to_uri(GFilename(path).c_str(), nullptr, &error);
         }
         return g_filename_to_uri(GFilename(fs::absolute(path)).c_str(), nullptr, &error);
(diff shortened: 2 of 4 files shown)


bhennion:
There are other uses of fs::absolute or fs::isAbsolute in the code that may need wrapping.
It could be more reliable to use ld/lld's --wrap option (only when compiling with libstdc++ on Windows): the simple wrapper in Util will very quickly be forgotten in newly written code.

Also, FYI, our Windows ARM releases use libc++.

Nerixyz:
> There are other uses of fs::absolute or fs::isAbsolute in the code that may need wrapping.
> It could be more reliable to use ld/lld's --wrap option (only when compiling with libstdc++ on Windows): the simple wrapper in Util will very quickly be forgotten in newly written code.

You're right, that's a better approach. It looks like `std::filesystem::path::isAbsolute` doesn't need wrapping if `std::filesystem::absolute` is wrapped.

> Also, FYI, our Windows ARM releases use libc++.

Ah, so the `clang-*` targets will use libc++.

bhennion:
> Ah, so the `clang-*` targets will use libc++.

Yes. See https://www.msys2.org/docs/environments/

bhennion:
> You're right, that's a better approach.

Would you be willing to implementing it? I can't really test things on Windows on my end.

Nerixyz:
> Would you be willing to implementing it? I can't really test things on Windows on my end.

That's what I did in the current change. I only had to replace `std::filesystem::absolute` for loading/saving to work.

src/util/PathUtil.cpp
```diff
@@ -149,25 +149,6 @@ auto Util::hasPngFileExt(const fs::path& path) -> bool {
     return StringUtils::toLowerCase(path.extension().string()) == ".png";
 }
 
-auto Util::isAbsolute(const fs::path& path) -> bool {
```
bhennion, line 158: The declaration of isAbsolute in the header should also be removed

src/util/CMakeLists.txt
```diff
@@ -11,3 +11,30 @@ target_compile_features(util PUBLIC ${PROJECT_CXX_FEATURES})
...
+      uses_libstdcpp "${CMAKE_CURRENT_BINARY_DIR}"
+      SOURCES "${test_path}"
+      CXX_STANDARD 17
+      CXX_STANDARD_REQUIRED ON)
+
+    file(REMOVE "${test_path}")
+
+    if (uses_libstdcpp)
+        # std::filesystem::absolute(std::filesystem::__cxx11::path const&)
+        target_link_options(util PUBLIC "-Wl,--wrap=_ZNSt10filesystem8absoluteERKNS_7__cxx114pathE")
```
bhennion: It would be safer if the mangled name was computed/detected automatically rather than hardcoded.
I don't think it is guaranteed to stay unchanged across versions (of the compiler or of the library).

bhennion (commented):
Ah sorry. I didn't see you had already implemented that.

Here are some comments

src/util/CMakeLists.txt
```diff
@@ -11,3 +11,30 @@ target_compile_features(util PUBLIC ${PROJECT_CXX_FEATURES})
...
+    ]=])
+
+    set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+    try_compile(
+      uses_libstdcpp "${CMAKE_CURRENT_BINARY_DIR}"
+      SOURCES "${test_path}"
+      CXX_STANDARD 17
+      CXX_STANDARD_REQUIRED ON)
+
+    file(REMOVE "${test_path}")
```
bhennion:
```suggestion
check_cxx_symbol_exists(__GLIBCXX__ version GLIBCXX)
if(GLIBCXX)
  # Logic for libstdc++
endif()
```
This should be enough. No need to do that by hand. (you may need to include `include(CheckCXXSymbolExists)` but it is probably already included at this point in the CMake script).

src/util/CMakeLists.txt
Nerixyz: Not sure how this could be automatically detected here. Do we have to compile some test program and see what it links to?

I think the mangled name itself should be stable, because the library tries to be backwards compatible (libstdc++ also keeps track of the mangled names it exposes - see https://github.com/gcc-mirror/gcc/blob/4626921de245970b4bd6ab0a837841185cf7de2d/libstdc%2B%2B-v3/config/abi/pre/gnu.ver). However, the library could introduce a new inline namespace like `__cxx29` if there will be breaking changes.

bhennion (commented):
Just to make sure, we should also add a test unit (only on Windows, I guess), trying `fs::absolute()`, `fs::weakly_canonical()` (`fs canonical()` maybe?) and things like `fs::absolute().is_absolute()` or so.

test/unit_tests/util/PathTest.cpp
```diff
@@ -222,3 +224,34 @@ TEST(UtilPath, pathGFilenameConvertion) {
...
+    EXPECT_EQ(fs::absolute("C:\\Windows\\System32"), "C:\\Windows\\System32");
+    // UNC paths should not be modified
+    EXPECT_EQ(fs::absolute("\\\\server\\some\\path"), "\\\\server\\some\\path");
+    EXPECT_EQ(fs::absolute("\\\\server\\some\\path\\..\\other"), "\\\\server\\some\\path\\..\\other");
+}
+
+TEST(UtilPath, stdFsIsAbsolute) {
+    EXPECT_TRUE(fs::path("C:\\Windows\\System32").is_absolute());
+    // UNC paths are not considered absolute in libstdc++
+    EXPECT_FALSE(fs::path("\\\\server\\some\\path").is_absolute());
```
bhennion: Isn't the goal to test the wrap, which should return true here?
I think the assertion `fs::absolute(path).is_absolute() == true` should always hold. Here it doesn't hold if `path` is a UNC path.

test/unit_tests/util/PathTest.cpp
```diff
@@ -222,3 +224,34 @@ TEST(UtilPath, pathGFilenameConvertion) {
     test(u8"foo/µbar/Ñ/ë€ds测试q.pdf");
     test(u8"/foo/µbar/Ñ/ë€ds测试q.pdf");
 }
+
+#ifdef XOURNALPP_WRAP_STD_FS_ABSOLUTE
```
bhennion: I think we should always run this test on Windows (even if libc++ is used). We want our releases to have the same behaviour with respect to UNC paths

bhennion (commented):
Thanks for the changes!
I still have a couple of remarks

test/unit_tests/util/PathTest.cpp
Nerixyz: Enabled the tests on Windows. With some cases being specific to the wrap.

test/unit_tests/util/PathTest.cpp
Nerixyz: Sorry for the long delay, I marked this as read but forgot about it.

> Isn't the goal to test the wrap, which should return true here?

We can't wrap `std::filesystem::path::is_absolute` and `::is_relative`, because they're `inline` functions. `is_absolute` requires the path to have a root name (i.e. the drive letter). This isn't present in UNC paths. My fix for libstdc++ was to check for a UNC path there, but we can't do this due to the method being `inline`.

bhennion: But then it looks like some features might still no be working.
There are some occurrences of fs::is_relative in control/xojfile/LoadHandler.cpp for instance. I think for instance that if you open a file and add a background image which is on a UNC path, then the .xopp file won't be loaded correctly.

I'm sorry I hadn't realized that is_absolute/is_relative could not be wrapped. It may be that your original solution was better in the end (having custom isAbsolute or toAbsolute functions). At least it makes explicit the issue. I'm not sure.

Nerixyz: Another alternative to using libc++ could be to use Boost.Filesystem on Windows if needed. That was removed in 9260ea984823c26fe131c22821d499ed63613b70, but from what I can tell, it shouldn't be too much effort of bringing it back.

Rots:
The Windows x86_64 CI passed, but I can’t find a downloadable artifact to test the UNC-path fix. Would it be possible to upload a Windows test build, or point me to one? I’m happy to test it against Open with… behaviour with UNC paths as I have the issue with v1.3.7.

PR #6929, opened 2025-12-25
Fix pixelized Color icons

### LINKED ISSUE #6892: Pallet is blurry in the version 1.3.0
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Official Release

### Xournal++ Version

1.3.0

### libgtk Version

3.24.43

### Bug Description

In version 1.3.0, the pallet is not clear as other tool icons:
Image

### Expected Behaviour

like version 1.2.10: 
Image

### Steps to Reproduce

Set the system scaling to 200%. Open Xournal++

### Additional Context

_No response_

bhennion:
Fix #6892 

The color icons in the toolbar are not really scalable the way they are implemented in 1.3.
The reason is: we create a GtkImage from a GdkPixbuf and there is no way of telling GTK that this Pixbuf should be interpreted as a HiDPI buffer.
Two solutions

1. Revert back to how it worked in 1.2: connect to the signal Widget::draw and handle the drawing by hand. Not a solution that could work with GTK4 though.
2. Feed the icon as named icons for GTK and use gtk_image_new_from_icon_name(). This then relies in GTK's proper handling of icons: HiDPI works, changes in the CSS style are taken into account, and probably other things.

This PR proposes to implement 2. To achieve that each color icon is created as an SVG GResource and added to the global GResource context.
I could not find any documentation on how to create the GResources, so I looked at how they are made in `glib-compile-resource`. It relies on the "source-only" library libgvdb (GVariant DataBase) which is pulled in by the CMake.
An alternative would be to create icon files in a tmp directory, but it seems less clean to me and possibly less safe (in case the tmp files are modified, I'm not sure).

This solution works with GTK4 as well.

Note: I targetted this to release-1.3 but it may be a bit too big of a fix. To be debated.

### DIFF
--- a/src/core/gui/toolbarMenubar/icon/ColorIcon.cpp
+++ b/src/core/gui/toolbarMenubar/icon/ColorIcon.cpp
@@ -1,17 +1,142 @@
 #include "ColorIcon.h"
 
-#include <cmath>  // for M_PI
+#include <iomanip>
+#include <memory>
+#include <sstream>
 
-#include "util/raii/CairoWrappers.h"
+extern "C" {
+#include <gvdb/gvdb-builder.h>
+}
+
+#include "gui/toolbarMenubar/model/ColorPalette.h"
+#include "util/Recolor.h"
+#include "util/glib_casts.h"
 #include "util/raii/GObjectSPtr.h"
+#include "util/serdesstream.h"
 
 namespace ColorIcon {
+
+static constexpr auto ICON_THEME_RESOURCE_PATH = "/org/xournalpp/colors/icons/";
+static constexpr auto FULL_RESOURCE_PATH = "/org/xournalpp/colors/icons/scalable/actions/";
+
+static constexpr auto CIRCLE_ICON_CORE_NAME = "circle";
+static constexpr auto SQUARE_ICON_CORE_NAME = "square";
+static constexpr auto ICON_EXTENSION = ".svg";
+
+static std::string colorToHex(Color c) {
+    auto s = serdes_stream<std::stringstream>();
+    s << std::hex << std::setw(6) << std::setfill('0') << (uint32_t(c) & 0x00ffffff);
+    return s.str();
+};
+
+static GvdbItem* createParent(GHashTable* table) {
+    std::string path = FULL_RESOURCE_PATH;
+    auto* parent = gvdb_hash_table_insert(table, path.c_str());
+    // We need to add the whole hierarchy
+    auto* ancestor = parent;
+    for (auto it = std::next(path.rbegin()); it != path.rend(); it++) {
+        if (*it == '/') {
+            it[-1] = '\0';  // The c-string ends here now
+            auto* item = gvdb_hash_table_insert(table, path.c_str());
+            gvdb_item_set_parent(ancestor, item);
+            ancestor = item;
+        }
+    }
+    return parent;
+}
+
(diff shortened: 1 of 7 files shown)


src/core/gui/toolbarMenubar/icon/ColorIcon.cpp
```diff
@@ -1,17 +1,142 @@
...
+        gvdb_item_set_parent(item, parent);
+        // And the square icon
+        key = FULL_RESOURCE_PATH + colorAsHex + SQUARE_ICON_CORE_NAME + ICON_EXTENSION;
+        item = gvdb_hash_table_insert(table, key.c_str());
+        gvdb_item_set_value(item, makeGVariant(makeSVG(colorAsHex, "", false)));
+        gvdb_item_set_parent(item, parent);
+    }
+    if (recolor) {
+        // Also add icons with the converted color in a corner
+        // We still need to normal icons in various places (e.g. Toolbar customization or Palette settings)
```
rolandlo:
```suggestion
        // We still need the normal icons in various places (e.g. Toolbar customization or Palette settings)
```

rolandlo (commented):
> Note: I targetted this to release-1.3 but it may be a bit too big of a fix. To be debated.

In my opinion pulling in a new dependency is really too much for a bugfix.

For the PPA (where we can't fetch sources during build) we have already disabled (the optional) `cpptrace`. How is this going to work with the (non-optional) `libgvdb`? I guess we would have to create a new recipe for `libgvdb` that is then nested into the original recipe. 
https://documentation.ubuntu.com/launchpad/user/explanation/packaging/packaging/#nesting
This is non-trivial and requires communication with Andreas Butti, who manages the PPA. It may also be non-trivial for other packagers.

bhennion:
> In my opinion pulling in a new dependency is really too much for a bugfix.

Well, it is only a single .c/.h pair we are pulling/using, and we are linking them statically. But anyway, I'm fine with targetting this at master.

> For the PPA (where we can't fetch sources during build) we have already disabled (the optional) `cpptrace`. How is this going to work with the (non-optional) `libgvdb`?

libgvdb is not conceived to be distributed as an independent library. It is thought as a "pull-my-sources-in-you-project' dep. I think we can do what libglib2.0-dev does. I'm not sure how they do it yet but it seems to be possible.

Again, we could also opt for another solution. Either

1. Pulling the necessary gvdb sources directly in our repo
2. Avoiding using GResources (and thus gvdb) and creating icon files in a tmp directory. This would need more testing on various platforms. Among the questions I have: how protected are tmp files? could they be flushed during the app's lifetime?

tmoerschell:
Is it not possible to query the widget scale before creating the pixbuf, or does no widget exist at this point? In GTK4, textures/paintables may be preferable anyways. How I understand it, resources seem to be rather targeted at fully static elements.

bhennion:
> Is it not possible to query the widget scale before creating the pixbuf

We certainly can get the scale value when creating the pixbuf. What we cannot do is assign a pixbuf to a GtkImage and have that pixbuf being used as a HiDPI buffer: The icon will still appear blurry, altough twice as big. It won't have a more dense resolution. We could then try to set the image width/height by hand, but `gtk_widget_set_size_request()` only allow to enlarge widget, not shrink them. No luck there.

archisman-panigrahi:
Any updates on this PR?

PR #6935, opened 2025-12-26
feat: improve windows installer directory handling

chengxilo:
This PR enhances the Windows installer script to prevent accidental installation into mixed or incorrect directories.

## The Problem I Want to Address
Currently, when we choose a custom location in the installer, it won't check the directory is empty or not, and will not create a folder for the software automatically.

For example, I choose `C:\Users\cheng\Downloads\test` as my target to download. I actually hope xournal to be installed at `C:\Users\cheng\Downloads\test\Xournal++`. However, it will decompress the files into `C:\Users\cheng\Downloads\test`  directly. 🥲 And the `test1.txt` and `test2.txt` are mixed with the Xournal files. 
image

### When browsing for a target location, the installer now automatically appends `\Xournal++` to the selected path.

I chose C:\Users\cheng\Downloads here, it automatically append \Xournal++ at the end.
image

### When the folder we choose is not empty, we will get a warning.
image

### DIFF
--- a/windows-setup/xournalpp.nsi
+++ b/windows-setup/xournalpp.nsi
@@ -13,7 +13,6 @@ Unicode true
 !include x64.nsh
 !include "${SCRIPT_DIR}\FileAssociation.nsh"
 !include nsDialogs.nsh
-!include "WordFunc.nsh"
 
 ; Options for MultiUser plugin
 !define MULTIUSER_INSTALLMODE_INSTDIR "Xournal++"


bhennion:

bhennion:
Thanks for the PR! The changes here look good to me. I don't have access to a Windows machine at this point to test it though.
@rolandlo Could you give this a very quick test (whenever you have time - there is absolutely no hurry)?

archisman-panigrahi:
I have a windows machine. I'll test it in the weekend.

Issue #6915, opened 2025-12-20
New recolor feature ignores dark and light color definitions


nimaaskarian:
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

### (Linux only) Desktop Environment

dwm

### (Linux Only) Display Server

X.Org

### Installation Method

Official Release

### Xournal++ Version

xournalpp 1.3.0 (9b9778d9 from srcdest/makepkg)

### libgtk Version

libgtk: 3.24.51

### Bug Description

The new recolor feature completely ignores the light and dark color definitions and chooses the darker color (probably the one that has a lower luminance as i observed) between the two as background, making user unable to choose a light color as their background

### Expected Behaviour

for white (lighter colors in general) to be replaced with `recolor.light` and black (darker colors in general) to be replaced with `recolor.dark`.

### Steps to Reproduce

1. set `recolor.light` to a dark color, `recolor.dark` to a light color (not necessarily via the config file, can also be reproduced with the GUI settings) (here i used color0 and color7 of gruvbox-soft as an example)
```xml
  <property name="recolor.dark" value="4294108604"/>
  <property name="recolor.light" value="4283451717"/>
```
2. everything works fine
3. set `recolor.light` to a light color, `recolor.dark` to a dark color
```xml
  <property name="recolor.dark" value="4283451717"/>
  <property name="recolor.light" value="4294108604"/>
```
4. same output as the situation in step 2, but it actually should be reversed.

bhennion:
@atticus-sullivan Was this on purpose? Could you have a look?

atticus-sullivan:
Doesn't sound like something I did on purpose.
I'll look into it (maybe Monday).

nimaaskarian:
i'll try to review the code and find the problem even though i'm a bit busy

guess i'll switch to my gunzip color switch script for now ;)

atticus-sullivan:
https://github.com/xournalpp/xournalpp/blob/master/src/util/Recolor.cpp#L34

that's the core function. Ok now I see the idea of this implementing some kind of dark theme, leads to the initial color inversion. And that leads to the "failure" with non dark theme parameters.

We might check the "order" (colorwise) of the two parameter colors and apply the inversion based on this condition. But of course this is more code (one additional check) in frequently executed code (thinking of performance, but I don't know how large the impact really is).

bhennion:
[...]

PR #6939, opened 2025-12-28
feat: add settings reload on SIGUSR1

nimaaskarian:
reloads settings from file on SIGUSR1 (for dynamic themeing setups, like darkman, pywal etc)

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -124,10 +124,20 @@ using std::string;
 Control* Control::instance = nullptr;
 
 void Control::SigUsr1Handler(int sig) {
-    if (Control::instance) {
+    if (instance) {
         SettingsBefore settingsBefore = instance->getSettingsBefore();
-        Control::instance->LoadSettings();
+        instance->LoadSettings();
         instance->reloadSettings(settingsBefore);
+        auto pageNr = instance->getCurrentPageNo();
+        if (pageNr == npos) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


nimaaskarian:
the latest commit converts settings dialog callback lambda to a separate method called `Control::reloadSettings`, which then its used in both the SIGUSR1 handler and settings dialog's callback. also converts the anonymous struct type of `settingsBeforeDialog` into its own struct, called `SettingsBefore`. there's also a function that extracts this info from a control instance, called `Control::getSettingsBefore` which allows for defining these settings and reload behavior once, and used in both SIGUSR1 handler and settings dialog's callback.

nimaaskarian:
when #6915 gets discussed over and merged, either one of these should be used as `Recolor::operator==`, to avoid false positives in equality of recolors (e.g background #fff, foreground #000 and vice versa produce same diff and offset, but these are actually different settings when #6915 closes):

```cpp
    constexpr friend bool operator==(Recolor const& lhs, Recolor const& rhs) {
        return lhs.light == rhs.light && lhs.dark == rhs.dark;
    }

```
```cpp
    constexpr friend bool operator==(Recolor const& lhs, Recolor const& rhs) {
        return lhs.difference == rhs.difference && lhs.offset == rhs.offset && lhs.ref == rhs.ref;
    }
```

nimaaskarian:
@bhennion please check this PR as well, goes really well with #6948

PR #6940, opened 2025-12-28
Fix Gtk4 syntax errors

rolandlo:


src/core/gui/toolbarMenubar/ToolPageLayer.cpp
```diff
@@ -101,7 +101,7 @@ static std::tuple<GtkWidget*, std::vector<ShowLayerEntry>, std::vector<std::pair
                                                                  Layer::Index id) -> std::pair<GtkWidget*, gulong> {
         GtkWidget* btn = gtk_check_button_new_with_label(layerName.c_str());
         // Is grouping necessary here? The GTK4 doc is unclear
-        gtk_check_button_set_group(GTK_CHECK_BUTTON(btn), std::exchange(group, GTK_CHECK_BUTTON(btn));
```
bhennion, line 104: Actually, grouping is not necessary. Those two lines (and the `group` variable itself) can be removed entirely.

src/core/gui/dialog/DeviceTestingArea.cpp
```diff
@@ -211,10 +211,10 @@ DeviceTestingArea::DeviceTestingArea(GladeSearchpath* gladeSearchPath, GtkBox* p
                                             gtk_toggle_button_get_active(btn);
                                 }),
 #else
-                                (GtkCheckButton* btn,, gpointer d) {
```
bhennion, line 214: Did the clang formatting remove those spaces? It looks like they should stay.

rolandlo: Yes, it's the clang format. They can't stay when the Clang format test shall succeed.

src/core/gui/toolbarMenubar/ToolPageLayer.cpp
rolandlo: Does this apply to Gtk 3 as well?

bhennion: No, I don't think so.

src/core/gui/toolbarMenubar/StylePopoverFactory.cpp
```diff
@@ -34,7 +34,7 @@ GtkWidget* StylePopoverFactory::createPopover() const {
     auto appendLineStyleItem = [&](const char* label, const char* target, const char* icon) {
         GtkWidget* btn = gtk_check_button_new_with_label(layerName.c_str());
         // Is grouping necessary here? The GTK4 doc is unclear
-        gtk_check_button_set_group(GTK_CHECK_BUTTON(btn), std::exchange(group, GTK_CHECK_BUTTON(btn));
+        gtk_check_button_set_group(GTK_CHECK_BUTTON(btn), std::exchange(group, GTK_CHECK_BUTTON(btn)));
```
rolandlo: But here it's not necessary either, right?

bhennion: Indeed!

rolandlo:
Thanks for the review!

PR #6943, opened 2025-12-29
Fix image orientation for drag & drop

### LINKED ISSUE #6911: image import does not respect exif rotation (initial image load only)
### Operating System

Linux

### (Linux only) Distribution

ubuntu22.04

gnome, i3, x11

X.Org

### Installation Method

flatpak

### Xournal++ Version

1.3.0 from flathub

### libgtk Version

3.24.51

### Bug Description

submitting a bug as this seems distinct from https://github.com/xournalpp/xournalpp/issues/918, which is about the changes needed to UI & underlying 'element' in the (dom?) to support rotating it.  

This bug is about pasting the original image as the user (or, sometimes device like camera) has configured it, the `orientation` exif flag specifically.  This bug neither asks for nor implies any UI changes, only to the process of loading an image.   

Further (and with the huge, respectful caveat that I'm familiar with neither GTK nor xournal code specifically :) - reading through the image code it appears at least some attempt to handle this is being made already in `Image::setImage` and `Image::renderBuffer` of `core/model/Image.cpp`.  If built-in gtk/pixbuf functions like `rotate_simple` aren't applicable, manual `cairo_translate` and similar are possible.

relevant resources:

- code already seem to apply this via `gdk_pixbuf_apply_embedded_orientation` 
    - https://docs.gtk.org/gdk-pixbuf/method.Pixbuf.apply_embedded_orientation.html
- exif `orientation` enumeration definitions
    - tag id 0x0112 in https://exiftool.org/TagNames/EXIF.html
- if for some reason it's necessary to parse directly, `orientation` should already be present (if applicable) in the existing `xoj::util::GObjectSPtr<GdkPixbufLoader>`
    - https://docs.gtk.org/gdk-pixbuf/method.Pixbuf.get_option.html
    > Since 2.12, the TIFF and JPEG loaders return an “orientation” option string that corresponds to the embedded TIFF/Exif orientation tag (if present).
- and `cairo_rotate` / `_translate` / `_scale` accomplish the transform
    - https://www.cairographics.org/manual/cairo-Transformations.html#cairo_translate

### Expected Behaviour

the image appears in my journal in the same orientation that it appears in thumbnailer, image viewers, drag-drop into `libreoffice`, etc

### Steps to Reproduce

1. take an image, strip it of exif eg `exiftool -all:all=  image_file`
2. make a copy and use editor of your choice to affect it's `orientation` tag to a rotation.  A simple image viewer will do.  verify with `exiftool`
3. load / drag-drop both copies in to xournal

Both images are identical, but one should be rotated relative to the other.

### Additional Context

Thank you for `xournal++`, it has been a huge boon to enable me to take notes at uni on my preferred stack (linux) rather than having to use a windows device & onenote!!!

rolandlo:
Fixes #6911

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -2225,12 +2225,14 @@ void Control::clipboardPasteText(string text) {
 
 void Control::clipboardPasteImage(GdkPixbuf* img) {
     auto image = std::make_unique<Image>();
-    image->setImage(img);
+    xoj::util::GObjectSPtr<GdkPixbuf> pixbuf(gdk_pixbuf_apply_embedded_orientation(img), xoj::util::adopt);
+
+    image->setImage(pixbuf.get());
 
     auto zoom100 = this->getZoomControl()->getZoom100Value();
 
-    auto width = static_cast<double>(gdk_pixbuf_get_width(img)) / zoom100;
-    auto height = static_cast<double>(gdk_pixbuf_get_height(img)) / zoom100;
+    auto width = static_cast<double>(gdk_pixbuf_get_width(pixbuf.get())) / zoom100;
+    auto height = static_cast<double>(gdk_pixbuf_get_height(pixbuf.get())) / zoom100;
 
     auto pageNr = getCurrentPageNo();
     if (pageNr == npos) {


rolandlo:
For testing: Take say [this picture](https://github.com/recurser/exif-orientation-examples/blob/master/Landscape_7.jpg) and drag&drop it from the file explorer into Xournal++ (with and without this patch) and see the difference.

rolandlo:
~@bhennion Can you explain me the difference between the first commit (using `xoj::util::adopt`) and the combined two commits (manual `unref` after use)? The first one crashes (during `im:get_typeof("orientation")`) when using the image in a plugin:~

```lua
function orientation()
    local vips = require "vips"
    local images = app.getImages("selection")
    for k, img in pairs(images) do
       im = vips.Image.new_from_buffer(img["data"])
       -- im = im:copy()       avoids the crash
       print("Got image data")
       if im:get_typeof("orientation") == 0 then
           print(k, "No orientation")
       else
           print(k, im:get("orientation"))
       end
       im:write_to_file("/home/roland/Bild" .. tostring(k) .. ".jpg")
    end
end
```

Turns out I had `im = im:copy()` in the script when testing the second version. I managed to reproduce the bug in lua-vips independently of Xournal++.

rolandlo:
> LGTM!
> 
> One could argue that this procedure should happen in Image::setImage(), to avoid code duplication and reduce maintainance cost. It could be delayed to a later PR if you want though.

If this happens in `Image::setImage()` one would have to be careful with `width` and `height` of the `img` Pixbuf, which are used later. I will merge as is. Thanks for the review.

PR #6948, opened 2026-01-02
fix: fix light color as recolor.light (#6915)

nimaaskarian:
- add a `ref` property to Recolor to decide on different channels
- fix `Recolor::operator==` to resolve false negatives in recolor refresh

### DIFF
--- a/src/util/Recolor.cpp
+++ b/src/util/Recolor.cpp
@@ -25,18 +25,15 @@ void Recolor::recalcDiffAndOff() {
 ColorU8 Recolor::convertColor(const ColorU8& other) const {
     return ColorU8{
             static_cast<uint8_t>(
-                    std::min(255, std::min(255, (abs(static_cast<int>(ref.red) - static_cast<int>(other.red)) -
-                                                 static_cast<int>(other.red)) *
+                    std::min(255, std::min(255, std::abs(static_cast<int>(ref.red) - static_cast<int>(other.red)) *
                                                         static_cast<int>(difference.red) / 255) +
                                           offset.red)),
             static_cast<uint8_t>(
-                    std::min(255, std::min(255, (abs(static_cast<int>(ref.green) - static_cast<int>(other.green)) -
-                                                 static_cast<int>(other.green)) *
+                    std::min(255, std::min(255, std::abs(static_cast<int>(ref.green) - static_cast<int>(other.green)) *
                                                         static_cast<int>(difference.green) / 255) +
                                           offset.green)),
             static_cast<uint8_t>(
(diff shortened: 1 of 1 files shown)


atticus-sullivan (approved):
Implements https://github.com/xournalpp/xournalpp/issues/6915

Looks good to me. I also quickly tested (with that rainbow picture)
- red/green
- green/red
- blue/yellow
- my personal configuration

all except the last one are the edge cases I identified so far. Also swapping light/dark colors does work for me.

(Note: I'm just a "small" contributor, so not sure how much weight my review has. Also I didn't check for formatting/code-style in more detail)

bhennion (commented):
The code looks good to me. Two things though:

1. As this is a bug fix, it'll go into v1.3.2. This means the PR needs to be targetted at the release-1.3 branch. You can change the base branch by doing
    1.  Run `git rebase --onto release-1.3 master` and then `git push -f`
    2. Click on the `Edit` button on top of the PR page and change the base branch.
2. Clang-formatting needs to be applied on your patch. You can do that by running  `git clang-format release-1.3` (once 1. has been done).

nimaaskarian:
i don't think that's what you wanted to happen. lol

nimaaskarian:
in this force push i have
1. applied the previously discussed changes to `release-1.3`
2. did `git clang-format release-1.3` after committing

src/util/Recolor.cpp
```diff
@@ -15,26 +15,37 @@ void Recolor::recalcDiffAndOff() {
...
-                                                                     static_cast<int>(difference.green) / 255) +
-                                                       offset.green)),
-            static_cast<uint8_t>(std::min(
-                    255, std::min(255, (255 - static_cast<int>(other.blue)) * static_cast<int>(difference.blue) / 255) +
-                                 offset.blue)),
+            static_cast<uint8_t>(
+                    std::min(255, std::min(255, (abs(static_cast<int>(ref.red) - static_cast<int>(other.red)) -
+                                                 static_cast<int>(other.red)) *
+                                                        static_cast<int>(difference.red) / 255) +
+                                          offset.red)),
```
bhennion: Actually, this doesn't look like this formula is that same as what happens in cairo. Shouldn't it be the following?
```suggestion
                    std::min(255, std::min(255, (std::abs(static_cast<int>(ref.red) - static_cast<int>(other.red)) *
                                                        static_cast<int>(difference.red) / 255) +
                                          offset.red)),
```

nimaaskarian: yeah i had a migraine when committing those changes, looks like i kinda messed it up.

bhennion:
Thanks for the fix!

Issue #6950, opened 2026-01-03
保存失败


Abiding-lxh:
### Operating System/Distribution

Windows 11 24H2

### (Linux Only) Display Server

None

### Installation Method

官方发布

### Xournal++ Version

xournalpp 1.3.0 (9b9778d9 from /)

### libgtk Version

3.24.43

### Steps to Reproduce

1使用obsidian的Xournal++插件
2生成xopp文件
3保存文件
4报错保存失败

### Crash log

\BaiduSyncdisk\Obsidian\test/testIntro 1.xopp"
Xournal++ started with PID: 122456
Exited with error status: 1
    Date: Sun Jan 04 01:23:47 2026

xournalpp 1.3.0 (9b9778d9 from /)
├──libgtk: 3.24.43
├──glib: 2.86.3
├──cairo:  1.18.4
└──OS info: Windows 11 24H2
  (The GDK backend is probably printed out below)

*** Output: ***

*** Cerr: ***

(com.github.xournalpp.xournalpp:122456): xopp-ERROR **: 01:23:46.437: Do not overwrite the background PDF! This will cause errors!

*************************************************************************

[Crash Handler] Crashed 1 time(s) with signal 22
xournalpp 1.3.0 (9b9778d9 from /)
├──libgtk: 3.24.43
├──glib: 2.86.3
├──cairo:  1.18.4
└──OS info: Windows 11 24H2

Try to get a stacktrace...
Stack trace (most recent call first):
#0  0x00007ff62409b655 in Stacktrace::printStacktrace(std::ostream&) at D:/a/xournalpp/xournalpp/src/util/Stacktrace.cpp:15:29
#1  0x00007ff623ed2109 in crashHandler at D:/a/xournalpp/xournalpp/src/core/control/CrashHandler.cpp:58:36
#2  0x00007ffb16cbec00 in raise at C:\WINDOWS\System32\msvcrt.dll
#3  0x00007ffb16cc305a in abort at C:\WINDOWS\System32\msvcrt.dll
#4  0x00007ffaad726eb4 in g_io_channel_new_file_utf8 at C:\Users\Abiding\AppData\Local\Programs\Xournal++\bin\libglib-2.0-0.dll
#5  0x00007ffaad6ae1d2 in g_aligned_free_sized at C:\Users\Abiding\AppData\Local\Programs\Xournal++\bin\libglib-2.0-0.dll
[...]

PR #6956, opened 2026-01-06
Avoid `g_error` for command line parameter errors

tmoerschell:
Xournal++ currently call `g_error` for errors that can originate from the command line parameters:
* Opening non-existent file
* Saving / exporting to unwritable location
* Exporting to same file as background PDF (see #6950)

This triggers the crash handler (application is terminated), which is probably not what we want. The documentation for `g_error` also explicitly states:
 > This is not intended for end user error reporting.

Three questions:
 * The function comments specify negative failure codes. However `handle_local_options` will stop processing options only if the return value is positive or 0. If we keep a simple `return -3;`, the GUI window will open on a failed command line call. Better to use `std::exit(-3)` or change the return code to a positive value?
 * Language: Some strings are translated, others aren't. What should we use on the command line?
 * Re-throwing the exception in `ExportHelper::exportPdf()`: nested exceptions are complicated. Wrapping the exception message should give enough context. Good as is?

This is a little bit on the edge for bugfix vs master. The only user difference should be text output and more correct return codes. I added the -2 / -3 pattern to the `--save` option. Maybe not all commits should go to the same branch.

### DIFF
--- a/src/core/control/ExportHelper.cpp
+++ b/src/core/control/ExportHelper.cpp
@@ -73,6 +73,7 @@ auto exportImg(Document* doc, const char* output, const char* range, const char*
     std::string errorMsg = imgExport.getLastErrorMsg();
     if (!errorMsg.empty()) {
         g_message("Error exporting image: %s\n", errorMsg.c_str());
+        return -3;
     }
 
     g_message("%s", _("Image file successfully created"));
@@ -105,7 +106,7 @@ auto exportPdf(Document* doc, const char* output, const char* range, const char*
     try {
         if (!backgroundPDF.empty() && fs::exists(backgroundPDF)) {
             if (fs::weakly_canonical(path) == fs::weakly_canonical(backgroundPDF)) {
-                g_error("Do not overwrite the background PDF! This will cause errors!");
+                g_message("Do not overwrite the background PDF! This will cause errors!");
                 return -3;  // Return error code for export failure
             }
         }
@@ -128,7 +129,8 @@ auto exportPdf(Document* doc, const char* output, const char* range, const char*
     }
 
     if (!exportSuccess) {
-        g_error("%s", pdfe->getLastError().c_str());
+        g_message("%s", pdfe->getLastError().c_str());
+        return -3;
     }
 
     g_message("%s", _("PDF file successfully created"));
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -179,19 +179,27 @@ void exitOnMissingPdfFileName(const LoadHandler& loader) {
  *
  *  The priority is: pngDpi overwrites pngWidth overwrites pngHeight
  *
- * @return 0 on success, -2 on failure opening the input file, -3 on export failure
+ * @return 0 on success
+ *
+ * Calls std::exit(-2) on failure opening the input file and std::exit(-3) on export failure
  */
 auto exportImg(const char* input, const char* output, const char* range, const char* layerRange, int pngDpi,
                int pngWidth, int pngHeight, ExportBackgroundType exportBackground) -> int {
     LoadHandler loader;
     auto doc = loader.loadDocument(input);
     if (doc == nullptr) {
-        g_error("%s", loader.getLastError().c_str());
+        g_message("%s", loader.getLastError().c_str());
+        std::exit(-2);
     }
 
     exitOnMissingPdfFileName(loader);
 
-    return ExportHelper::exportImg(doc.get(), output, range, layerRange, pngDpi, pngWidth, pngHeight, exportBackground);
+    const auto res = ExportHelper::exportImg(doc.get(), output, range, layerRange, pngDpi, pngWidth, pngHeight,
+                                             exportBackground);
+    if (res != 0) {
+        std::exit(res);
+    }
+    return 0;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


bhennion:
I am not quite sure I understand what issue is addressed by this PR.

> This triggers the crash handler (application is terminated), which is probably not what we want.

If the user is running the app via the CLI, then the process should certainly terminate immediately in case of a failure. Using `std::exit()` vs `g_error()` does not change the core behaviour: the crash handler pops up either way. If the user (e.g. obsidian-xournalpp) does not want the crash handler, they should call `xournalpp` and not `xournalpp-wrapper`.
The only benefit of using `std::exit()` is a finer control on the return code. This becomes a feature (which has not been requested yet, AFAIK).

I'm also uncertain as to why using exceptions is better than using the already set-up return codes (for the export functions). In any case, this looks like a refactor rather than a bugfix.

tmoerschell:
The main executable does currently print a stack trace to stderr:
```sh
~ $ ~/dev/xournalpp/build/xournalpp sample_file.xopp -p /etc/shadow

(com.github.xournalpp.xournalpp:178707): xopp-ERROR **: 14:22:02.224: Error with overlay or final export:
open /etc/shadow: Permission denied



*************************************************************************

[Crash Handler] Crashed 1 time(s) with signal 5
xournalpp 1.3.1+dev (f590496d from xournalpp/master)
├──libgtk: 3.24.51
├──glib: 2.86.3
├──cairo:  1.18.4
└──OS info: openSUSE Tumbleweed 20260103


Try to get a stacktrace...
Stack trace (most recent call first):
#0  0x000055bafe932f66 in Stacktrace::printStacktrace(std::ostream&) at /home/Thomas/dev/xournalpp/build/xournalpp
#1  0x000055bafe788438 in crashHandler at /home/Thomas/dev/xournalpp/build/xournalpp
#2  0x00007f3e1c04290f at /lib64/libc.so.6
#3  0x00007f3e1d12e5ab in g_log_structured_array at /lib64/libglib-2.0.so.0
#4  0x00007f3e1d12ea1f in g_log_default_handler at /lib64/libglib-2.0.so.0
#5  0x00007f3e1d12ec88 in g_logv at /lib64/libglib-2.0.so.0
#6  0x00007f3e1d12efee in g_log at /lib64/libglib-2.0.so.0
#7  0x000055bafe839a4a in ExportHelper::exportPdf(Document*, char const*, char const*, char const*, ExportBackgroundType, bool, ExportBackend) at /home/Thomas/dev/xournalpp/build/xournalpp
#8  0x000055bafe78b95e at /home/Thomas/dev/xournalpp/build/xournalpp
#9  0x00007f3e1d527886 at /lib64/libgio-2.0.so.0
#10 0x00007f3e1df4398a in g_closure_invoke at /lib64/libgobject-2.0.so.0
#11 0x00007f3e1df586b9 at /lib64/libgobject-2.0.so.0
#12 0x00007f3e1df599be at /lib64/libgobject-2.0.so.0
#13 0x00007f3e1df5fad3 in g_signal_emit_valist at /lib64/libgobject-2.0.so.0
#14 0x00007f3e1df5fb8e in g_signal_emit at /lib64/libgobject-2.0.so.0
#15 0x00007f3e1d58858a at /lib64/libgio-2.0.so.0
#16 0x00007f3e1d588904 in g_application_run at /lib64/libgio-2.0.so.0
#17 0x000055bafe78924f in XournalMain::run(int, char**) at /home/Thomas/dev/xournalpp/build/xournalpp
#18 0x00007f3e1c02b2fa at /lib64/libc.so.6
#19 0x00007f3e1c02b3ca in __libc_start_main at /lib64/libc.so.6
#20 0x000055bafe787d44 in _start at /home/Thomas/dev/xournalpp/build/xournalpp

Trace/breakpoint trap      (core dumped) ~/dev/xournalpp/build/xournalpp sample_file.xopp -p /etc/shadow

~ $ echo $?
133  # 0x88 - 3
```

With `std::exit`, it becomes:
```sh
~ $ ~/dev/xournalpp/build/xournalpp sample_file.xopp -p /etc/shadow
Error exporting PDF: Error with overlay or final export:
open /etc/shadow: Permission denied

~ $ echo $?
253  # 0x100 - 3
```

Using the function return values would be possible, but requires us to do an explicit check for it instead of returning it directly, as GTK will launch the application upon receiving a negative return value. Or we could change to positive return codes. I thought exceptions were the nicest way to obtain the behavior drawn out as intended by the current code.

bhennion:
Fine then. But as a rule of thumb (for later PRs), it is better to keep the diff as small as possible when writing a bugfix: less chances of introducing another bug in the process. In this instance, I feel like switching to exceptions added some unecessary changes. Don't worry about it for this one though. It's ok.


> * Language: Some strings are translated, others aren't. What should we use on the command line?

If the messages are destined to the users, they should be translated, same in CLI and GUI. We should never assume the user (even a CLI user -- they may be using the CLI interface indirectly, like the Obsidian plugin users) understands english. On the other hand, if we get a report with translated content, we can go on Crowdin and fetch the original string.

For the strings you added, please make them translatable (e.g. wrap them in `_(...)`). If there are other strings that are not translatable, changing that can go into another PR.

tmoerschell:
No worries, this wasn't huge work. Here's a minimal fix. I'll target the further steps at master so we can discuss it separately

Issue #1971, opened 2020-05-12
Multi-tab feature


toludwig:
It is a little unexpected that on opening/creation of a new file, the one currently worked on is closed.
To be able to edit two files, I need two have instances of Xournal++ open.
This is a quite frequent scenario as I annotate several lectures.

It would be convenient to be able to have multiple tabs open just like in a PDF viewer or in a browser.

The "Recent Documents" list is a less convenient way of switching, but in my case, this does not update at all (see #1970).

Tabs don't necessarily have to appear at the top but could also be included in the sidebar.

LittleHuba:
You should be able to just have two instances of Xournal++ running simultaneously. We are currently unable to provide a tab feature but will do so in the rewrite. But that will take some time until it is up to productive use.

idotobi:
@LittleHuba : I think what @toludwig refers to is that when you use `File -> Open` to open  `2.xopp` from your already running instance with `1.xopp` open, then you just replace `1.xopp` with `2.xopp` in your already running instance.
Thus, you need to start a new xournalpp instance from outside of xournalpp if you want to have both open simultaneously, which actually can be a more elaborate process when having your screen already flipped.
_At least that's what happened to me sometimes in the past. 😅_

Possible Solutions I'd see:
 1. When calling `File -> Open` or `File -> Recent` show a dialog (similar to `This document is not saved yet`) with the Options: a) Open in new Window b) open in current Window.
 2. Change the default to always open in new window
 3. Leave the default

LittleHuba:
I'd see solution 1 as the sole solution here. (2) is infuriating if you have just opened a window of Xournal++ and want to open a document. (3) doesn't fix it if this is what he actually wanted to say (which I'm not sure it is).

Febbe:
Yes a tab feature would be too much work. I think we should implement version one. This is also exactly what clion/jetbrain IDEs do for example.

idotobi:
Agreed.

(2) could be implemented less infuriating by checking first if no changes were made yet to the default page, but that would be probably more confusing than helpful.
(3) Yes that this would mean keeping the status quo. (I just wanted to add it to make it explicit, as of course it's almost always an option).

idotobi:
@toludwig : Would (1) solve your problem, or at least be a good workaround for the time being?

toludwig:
> @toludwig : Would (1) solve your problem, or at least be a good workaround for the time being?

[...]

PR #6957, opened 2026-01-06
Feature/multiple windows

mingchai:
Add new window option for File → Open & File → New operations
Allows users to open files or create new documents in a separate window instead of
replacing the current window content.

This attempts to address issue #1971

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1549,18 +1549,20 @@ void Control::askToOpenFile() {
 
 void Control::openFileInNewWindow(fs::path filepath) {
     std::string binary = "xournalpp";
-#ifdef __APPLE__
-    // On macOS, try to find the binary in the app bundle first
-    if (!xoj::util::OwnedCString::assumeOwnership(g_find_program_in_path(binary.c_str()))) {
-        auto path = Util::getExePath() / binary;
-        binary = Util::toGFilename(path).c_str();
+    gchar* prog = nullptr;
+
+    // First, try to use the executable from the same directory as the current one
+    auto localPath = Util::getExePath() / "xournalpp";
+    if (fs::exists(localPath) && fs::is_regular_file(localPath)) {
+        prog = g_strdup(Util::toGFilename(localPath).c_str());
+    } else {
+        // Fallback to searching in PATH
+        prog = g_find_program_in_path(binary.c_str());
     }
-#endif
-    gchar* prog = g_find_program_in_path(binary.c_str());
+
     if (!prog) {
-        // Fallback to current executable path
-        auto path = Util::getExePath() / "xournalpp";
-        prog = g_strdup(Util::toGFilename(path).c_str());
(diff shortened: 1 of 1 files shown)


zandivx:
Hey @mingchai, I pulled your changes and tested them, works great! One thing I would do differently is opening the current executable for the new window (via `getExePath`) and searching in path only as a fallback. Right now, when I test your feature with my self-compiled binary, it opens the system binary as a new window, which to me was unexpected.

mingchai:
> Hey @mingchai, I pulled your changes and tested them, works great! One thing I would do differently is opening the current executable for the new window (via `getExePath`) and searching in path only as a fallback. Right now, when I test your feature with my self-compiled binary, it opens the system binary as a new window, which to me was unexpected.

Hi @zandivx , thanks for taking the time to review and test - I've updated the PR to address this, but do let me know if I'm off the mark. Appreciate any feedback or guidance.

zandivx:
> > Hey @mingchai, I pulled your changes and tested them, works great! One thing I would do differently is opening the current executable for the new window (via `getExePath`) and searching in path only as a fallback. Right now, when I test your feature with my self-compiled binary, it opens the system binary as a new window, which to me was unexpected.
> 
> Hi @zandivx , thanks for taking the time to review and test - I've updated the PR to address this, but do let me know if I'm off the mark. Appreciate any feedback or guidance.

lgtm! Now it has the effect I expected.

bhennion:
I do not quite like this solution. I do not think the Open/New buttons behaviour should be altered: by doing so, you are adding a click to one usepath ("yes, open in that same window") to remove a click from another usepath (just open another instance).

If your desktop does not allow you to run several instances, you can complain there, or we could maybe add a "Open new Window" entry in the menu.

@rolandlo, Do you have an opinion on the matter?

rolandlo:
> I do not quite like this solution. I do not think the Open/New buttons behaviour should be altered: by doing so, you are adding a click to one usepath ("yes, open in that same window") to remove a click from another usepath (just open another instance).
> 
> If your desktop does not allow you to run several instances, you can complain there, or we could maybe add a "Open new Window" entry in the menu.
> 
> @rolandlo, Do you have an opinion on the matter?

I agree. It would be annoying to need one more click for opening another file in the same window. A menu entry "New Window" like other applications provide would make sense.

zandivx:
> > I do not quite like this solution. I do not think the Open/New buttons behaviour should be altered: by doing so, you are adding a click to one usepath ("yes, open in that same window") to remove a click from another usepath (just open another instance).
> > If your desktop does not allow you to run several instances, you can complain there, or we could maybe add a "Open new Window" entry in the menu.
> > @rolandlo, Do you have an opinion on the matter?
> 
> I agree. It would be annoying to need one more click for opening another file in the same window. A menu entry "New Window" like other applications provide would make sense.

I would argue that opening a specific file in a new window is a more common need than just opening a new window and then deciding what to do with it (which is almost certainly open an existing document). We could also add a new drop-down entry 'Open in new window' right below File -> Open. @bhennion @rolandlo what do you think of this idea instead?

PR #6961, opened 2026-01-06
fix: Argument forceOpen in Control::openFile has no effect

Aurelius7309:
Apparently this argument was only ever tested with no unsaved changes present - the save prompt is skipped in that case regardless. All this does is forward the force argument into Control::close, where all the necessary logic is already present. Another potential thing to do here is to allow a callback to be specified into openFile from the lua api.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -1697,12 +1697,14 @@ void Control::openFile(fs::path filepath, std::function<void(bool)> callback, in
         return;
     }
 
-    this->close([ctrl = this, filepath = std::move(filepath), cb = std::move(callback),
-                 scrollToPage](bool closed) mutable {
-        if (closed) {
-            ctrl->openFileWithoutSavingTheCurrentDocument(std::move(filepath), false, scrollToPage, std::move(cb));
-        }
-    }, false, true, forceOpen);
+    this->close(
+            [ctrl = this, filepath = std::move(filepath), cb = std::move(callback), scrollToPage](bool closed) mutable {
+                if (closed) {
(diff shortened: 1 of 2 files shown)


src/core/control/Control.cpp
```diff
@@ -1702,7 +1702,7 @@ void Control::openFile(fs::path filepath, std::function<void(bool)> callback, in
         if (closed) {
             ctrl->openFileWithoutSavingTheCurrentDocument(std::move(filepath), false, scrollToPage, std::move(cb));
         }
-    });
+    }, false, true, std::move(forceOpen));
```
rolandlo: There is no reason to `std::move` a boolean like `forceOpen`.
```suggestion
    }, false, true, forceOpen);
```

rolandlo (commented):
Thanks for the fix! 
Basically this looks good to me, although I will still need to test it.

rolandlo:
I think this bugfix should go into the `release-1.3` branch, so it will be available in a future version 1.3.2.
Just do
```term
git rebase --onto=upstream/release-1.2 master
git push --force
```
and change the target branch to `release-1.3` via the Edit button on the top right of this page.

Aurelius7309:
All done for the rebase. Your command didn't work, what I ended up needing to do was `git rebase --onto upstream/release-1.3 upstream/master master`

rolandlo (commented):
It works fine. Please apply clang-format though, e.g. 
```term
git-clang-format-18 --binary=clang-format-18 HEAD~2
```
I don't understand why the checks are not automatically running.

Aurelius7309:
Done. It says on my end that you still need to approve the checks.

PR #6993, opened 2026-01-17
gitignore ./install directory

zandivx:
The official compile instructions (`./readme/Compile.md`) instruct to set the directory `install` as install prefix. This directory should be excluded from git as a user following these instructions automatically clutters their staging area.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ Debug
 # Miscellaneous
 .DS_STORE
 *~
+/install/


bhennion:
You misread the instructions. The install directory you get by applying them is `./build/install`, which is already ignored as `build` is.

zandivx:
Apologies, you're right, thx for the clarification

PR #3230, opened 2021-07-15
Downsize all action icons via svgo

rolandlo:
The action icons were downsized using the command
svgo -f ./actions ./actions_new

Fixes #3224 and related performance issues

@Fatih20 As noticed in #3224 the large file sizes of the action icons resulted in performance problems on some systems. Using the tool [svgo](https://github.com/svg/svgo#why) reduced the file sizes drastically. For instance the 64 scalable action icons in the dark theme folder combined went down in file size from 13 MB to 375 KB, which is less than 3% of the original size. As you see from the screenshots below, there is no noticeable difference/quality loss from reducing the file size. Moreover the original icons will always remain in the commit history, if they should be needed again. So please tell me, are you fine with the downsizing?

Original:
dark_original

Downsized:
dark_svgo

Original:
light_original

Downsized:
light_svgo

Original:
page-spinner_original

Downsized:
page-spinner_svgo

PR #6994, opened 2026-01-17
Implement a jump-list navigation history for explicit jumps

### LINKED ISSUE #5354: Pdf hyperlink link backjumping

_No response_

### Feature Description

Being able to follow hyperlinks in the pdf background as implemented in https://github.com/xournalpp/xournalpp/pull/4412 is a real gamechanger. It would be a formidable QOL improvement if there were a way to jump back to wherever you started from, implemented for example as a jumplist as suggested by https://github.com/xournalpp/xournalpp/issues/1027#issuecomment-1453105412. 

This is especially useful for reading scientific publications, where you might want to jump around between equations or references. Since usually references are listed at the bottom of the document, scrolling back to where you clicked that hyperlink can be tedious.

### Alternative Solutions

_No response_

### Additional Context

_No response_

zandivx:
As suggested in issue #5354, a jump list could be implemented for explicit jumps e.g. following hyperlinks. With the help of Codex I implemented a jump-list navigation history for explicit jumps (PDF links, clicking on TOC, clicking on sidebar preview, go to page, go to first/last page, Home/End), plus back/forward actions with Alt+Left/Alt+Right shortcuts. The Jump Back/Jump Forward actions are exposed in the Navigation menu and the Navigation components for toolbars. I took inspiration in how navigation is handled in [SumatraPDF](https://github.com/sumatrapdfreader/sumatrapdf), as I really missed this feature since switching from it.

## Behavior
- Records the current view before explicit jumps
- Back/Forward navigates to the exact saved view (page + visible rect).
- History is bounded (max 50 entries, hard-coded), forward history is discarded on new jumps, and duplicate positions are de‑duplicated.
- History is pruned when pages are deleted or document changes; it resets on document close.
- Actions are enabled/disabled based on availability.

## Implementation Notes
- dedicated class `NavigationHistory`
- Navigation state in `Control`
- Recording wired into `ScrollHandler::jumpToPage`
- Action enablement updates tied into page updates and history changes.

Closes #5354.

### DIFF
--- a/src/core/control/ScrollHandler.cpp
+++ b/src/core/control/ScrollHandler.cpp
@@ -125,7 +125,7 @@ void ScrollHandler::scrollToAnnotatedPage(bool next) {
 
     for (size_t i = this->control->getCurrentPageNo() + step; i < doc->getPageCount() && i != size_t(-1); i += step) {
         if (doc->getPage(i)->isAnnotated()) {
-            scrollToPage(i);
+            jumpToPage(i);
             break;
         }
     }


src/core/control/Control.cpp
bhennion, line 1: For future maintenance, it would be better to have a dedicated .cpp/.h pair instead of adding to Control.cpp. So a dedicated NavigationHistory class (or whatever name you want).

src/core/control/Control.h
```diff
@@ -555,6 +581,19 @@ class Control:
     std::unique_ptr<GeometryTool> geometryTool;
     std::unique_ptr<GeometryToolController> geometryToolController;
 
+    struct NavState {
+        PageRef page;
+        double x1 = -1;
+        double y1 = -1;
+        double x2 = -1;
+        double y2 = -1;
+        bool hasRect = false;
```
bhennion: For semantics, using std::optional<Range> or std::optional<xoj::util::Rectangle<double>> would be better, I think

src/core/gui/XournalView.cpp
```diff
@@ -302,11 +302,17 @@ auto XournalView::onKeyPressEvent(const KeyEvent& event) -> bool {
     }
 
     if (keyval == GDK_KEY_End || keyval == GDK_KEY_KP_End) {
+        if (control->getCurrentPageNo() + 1 < control->getDocument()->getPageCount()) {
+            control->recordNavPoint();
+        }
```
bhennion: I think all the `recordNavPoint()` calls should rather go into the ScrollHandler functions you call (e.g. ScrollHandler::goToLastPage scrollToPage goToFirstPage and so on)

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
```diff
@@ -345,7 +345,9 @@ void ToolMenuHandler::initToolItems() {
 
     emplaceStockItem("GOTO_FIRST", Cat::NAVIGATION, Action::GOTO_FIRST, "go-first", _("Go to first page"));
     emplaceStockItem("GOTO_BACK", Cat::NAVIGATION, Action::GOTO_PREVIOUS, "go-previous", _("Back"));
+    emplaceStockItem("NAVIGATE_BACK", Cat::NAVIGATION, Action::NAVIGATE_BACK, "go-previous", _("Jump back"));
```
bhennion: You should probably not be using the same icons as the GOTO_{BACK,NEXT}: The user won't be able to tell NAVIGATE_BACK appart from GOTO_BACK if both are in the toolbar.
I do not have a XDG standard icon name to propose though. There is `go-jump` which sounds like the right thing, but only in one direction.
To be debated

src/core/control/Control.cpp
zandivx: Done

src/core/control/Control.h
zandivx: Great advice, thx! Done.

src/core/gui/XournalView.cpp
zandivx: Also good idea. I also added another method `jumpToPage` that records a navigation point and then calls `scrollToPage`.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
zandivx: I changed the icons now to `edit-undo` and `edit-redo`, which is of course semantically wrong, but the icons looked reasonable to me. This does not clash with the already existing undo and redo icons. @bhennion what do you think about these icons instead?

bhennion: Well, what they look like is dependent on the theme. Also, it does clash with the existing icons if one uses the stock icons (there is an option somewhere to do that).
I don't really have anything better to propose though, short of making custom icons.

src/core/control/Control.cpp
```diff
@@ -922,9 +934,9 @@ void Control::insertPage(const PageRef& page, size_t position, bool shouldScroll
 void Control::gotoPage() {
     auto popup = xoj::popup::PopupWindowWrapper<xoj::popup::GotoDialog>(
             this->gladeSearchPath, this->getCurrentPageNo(), this->doc->getPageCount(),
-            [scroll = this->scrollHandler](size_t pageNumber) {
+            [ctrl = this](size_t pageNumber) {
```
bhennion: You can restore the capture to only have the ScrollHandler (as it was before)

src/core/control/Control.h
```diff
@@ -393,6 +399,7 @@ class Control:
     void saveImpl(bool saveAs, std::function<void(bool)> callback);
 
 private:
+
```
bhennion: Please avoid "empty" changes like this.
```suggestion
```

src/core/control/NavigationHistory.cpp
```diff
@@ -0,0 +1,218 @@
...
+    if (!actionDB) {
+        return;
+    }
+
+    actionDB->enableAction(Action::NAVIGATE_BACK, canNavigate(-1));
+    actionDB->enableAction(Action::NAVIGATE_FORWARD, canNavigate(1));
+}
+
+void NavigationHistory::recordNavPoint() {
+    if (restoring) {
```
bhennion: I've been trying to track down where this `restoring` flag is actually needed. Can you pinpoint the execution path taht makes it useful?

bhennion:
It seems you need to apply clang format and update the Lua definition file. You can do that by running `git-clang-format HEAD~4` and `./scripts/lua_def_files.py`.

src/core/control/Control.cpp
zandivx: Done.

src/core/control/Control.h
zandivx: Sure, reverted it.

src/core/control/NavigationHistory.cpp
zandivx: It isn't anymore, I removed it.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
zandivx: Do you want to leave it this way? Or should we discuss creating a custom icon?

zandivx:
> It seems you need to apply clang format and update the Lua definition file. You can do that by running `git-clang-format HEAD~4` and `./scripts/lua_def_files.py`.

Thx! Done.

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
bhennion: I think it would be best to have a dedicated custom icon. I think it should look like a navigation icon (rather than a undo/redo), whatever this can mean. Do your best and that'll be fine, I think.

zandivx: I couldn't find any icons for that functionality in other PDF viewers for reference, so I came up with the following: An arrow left/right similar to the default XDG arrow next to a chain indicating jumping to and from a hyperlink. What do you think of this?

ui/iconsColor-light/hicolor/scalable/actions/xopp-navigate-back.svg
bhennion, line 1: Is this visible on a light theme?

src/core/gui/toolbarMenubar/ToolMenuHandler.cpp
bhennion: Its fine I think

bhennion (commented):
I think your icons are fine.
One thing though: could you apply `svgo` to you icons (see https://github.com/xournalpp/xournalpp/pull/3230)?

ui/iconsColor-light/hicolor/scalable/actions/xopp-navigate-back.svg
zandivx: I used the wrong color, it is fixed now.

zandivx:
> I think your icons are fine. One thing though: could you apply `svgo` to you icons (see #3230)?

Done.

src/core/control/Control.h
```diff
@@ -35,6 +35,7 @@
 #include "undo/UndoRedoHandler.h"                   // for UndoRedoHandler (ptr only)
 
 #include "ClipboardHandler.h"  // for ClipboardListener
+#include "NavigationHistory.h"  // for NavigationHistory
```
bhennion: Forward declaration would be better here (so removing this line and adding a `class NavigationHistory;` below): the least chain includes there are, the faster the build goes.

src/core/control/Control.cpp
```diff
@@ -97,6 +99,7 @@
 #include "util/PathUtil.h"                                       // for clea...
 #include "util/PlaceholderString.h"                              // for Plac...
 #include "util/PopupWindowWrapper.h"                             // for PopupWindowWrapper
+#include "util/Rectangle.h"                                      // for Rectangle
```
bhennion: Is this include still used?

src/core/control/Control.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Control.h"
 
 #include <algorithm>  // for max
+#include <cstddef>    // for ptrdiff_t
```
bhennion: Is this include still used?

src/core/control/Control.h
zandivx: Done.

src/core/control/Control.cpp
zandivx: Not used anymore. Removed.

src/core/control/Control.cpp
zandivx: Not used anymore. Removed.

bhennion:
The code looks all good to me. I noticed that some explicit jumps were not pushing a NavPoint:
1. clicking on the sidebar to jump to a specific page
2. Using Goto next/previous
3. Using Goto first/last page
4. Setting a page number in the "page counter" toolbar item
5. Using the "Goto next annotated page" feature

I'm guessing this is not voluntary (or if it is, what's the rationale?).

zandivx:
> The code looks all good to me. I noticed that some explicit jumps were not pushing a NavPoint:
> 
>     1. clicking on the sidebar to jump to a specific page
> 
>     2. Using Goto next/previous
> 
>     3. Using Goto first/last page
> 
>     4. Setting a page number in the "page counter" toolbar item
> 
>     5. Using the "Goto next annotated page" feature
> 
> 
> I'm guessing this is not voluntary (or if it is, what's the rationale?).

1. I tested the TOC sidebar (which works), I will record a navigation point when using the preview sidebar as well.
2. Scrolling by one page is no "jump" worth recording in my opinion. And if it was considered a jump, then the back and next button would already handle that use case.
3. I did adapt `ScrollHandler::goToFirstPage` and `ScrollHandler::goToLastPage` but `Action::GOTO_FIRST` and `Action::GOTO_LAST` do not use it. I will fix this as I don't see why the existing methods are re-implemented in `ActionProperties.h`.
4. I will add that.
5. I will add that.

bhennion:
There is another case I haven't tested and may have been forgotten: when you search for a keyword in a PDF (try Ctrl+F) and hit enter, you'll be moved to the first occurrence of that text. This should probably record a NavPoint. 

Ideally, only the original position should be recorded, I guess, as the intermediate steps we pass through (if we hit enter several times) are not really relevent. This could be more subtle to implement and can be left to a later PR, I think.

zandivx:
> There is another case I haven't tested and may have been forgotten: when you search for a keyword in a PDF (try Ctrl+F) and hit enter, you'll be moved to the first occurrence of that text. This should probably record a NavPoint.
> 
> Ideally, only the original position should be recorded, I guess, as the intermediate steps we pass through (if we hit enter several times) are not really relevent. This could be more subtle to implement and can be left to a later PR, I think.

I can do that in a follow-up PR.

Issue #6979, opened 2026-01-12
--export-no-background with/without pdf bg provided


MCXCC303:
When I get a xopp file without its original pdf bg, I want to use `xournalpp -i file.png --export-no-background file.xopp` to export plain handwriting pages, but it gives me:
```
The background file "path/to/pdf" could not be found. It might have been moved, renamed or deleted.
```
Also I tried to use a xopp file with a pdf bg, using the same command, and it succeeded exported, but it also generate plenty of empty pages.

I'm not sure if this is a good strategy on exporting files with cli, in my opinion it is supposed to skip any bg and replace them to transparent with --export-no-background whether the pdf bg is given or not, and empty pages should be skipped to export (but with correct, ordered page no. on name suffix).

Looking for a kind reply.

AhmedSaid3617:
I modified export to allow this in #6997 , but i didn't do the empty pages checks, it just exports the empty pages.
Maybe the maintainers could give us some insight into that.

MCXCC303:
> I modified export to allow this in [#6997](https://github.com/xournalpp/xournalpp/pull/6997) , but i didn't do the empty pages checks, it just exports the empty pages. Maybe the maintainers could give us some insight into that.

Seems good. The empty check could be a new feature but might not be a general function, I'm considering to compile another fork with these change and use this on my own way. Appreciate anyway.

PR #6997, opened 2026-01-18
Allow export-no-background even if pdf file is missing.

AhmedSaid3617:
This addresses issue #6979 by @MCXCC303
I modified exportPdf and exportImg to allow for exporting without background even if the xopp file references a pdf file that doesn't exist. This way users can save their notes even if the pdf has been deleted.

The issue also suggests skipping empty pages, let me know if this behavior would be desirable or if it's better to export with the empty pages.

This is my first PR here and I'm not sure if I should PR to master or some other branch, please advise.

### DIFF
--- a/src/core/control/XournalMain.cpp
+++ b/src/core/control/XournalMain.cpp
@@ -188,7 +188,7 @@ auto exportImg(const char* input, const char* output, const char* range, const c
     if (doc == nullptr) {
         g_error("%s", loader.getLastError().c_str());
     }
-    
+
     if (exportBackground != EXPORT_BACKGROUND_NONE) {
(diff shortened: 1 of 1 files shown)


bhennion:
Thanks for the PR!

The code look fine to me, but the clang-format test is failing. Could you run `git clang-format master` to apply it to your patch?

About the feature itself: I'm not so sure about the necessity of this, but as the code changes are really small, it is fine.
I do not think empty pages should be skipped though. If the user wants to skip pages, they can use the `--export-range=...` option.

AhmedSaid3617:
It should pass clang-format now.
Happy to help.

Issue #7007, opened 2026-01-20
Lua api: get all texts in document


tobiasBora:
### Is Your Feature Request Related to a Problem?

I'm writing a plugin to easily grade exams in xournalpp question per question. For this I need a method to list all texts in a document and jump to a specific text. The problem is that `app.getTexts` only prints the texts on the current page, and `app.getDocumentStructure()` does not seem to show the texts (or I missed it).

### Feature Description

API to list all texts

### Alternative Solutions

None found

rolandlo:
You can iterate through all pages and layers as explained in the [wiki](https://xournalpp.github.io/guide/plugins/plugins/) (bottom of that page).

tobiasBora:
Thanks for your answer, but I read it 30 times and I still can't see which API helps, how do I go from page number + layer number to a list of texts? Maybe a quick example could help.

rolandlo:
Try this (if only need the texts; otherwise modify accordingly):
```lua
function getAllTexts()
    local texts = {}
    local index = 1
    local docStructure = app.getDocumentStructure()
    local numPages = #docStructure.pages
    for i=1, numPages do
        app.setCurrentPage(i)
        local numLayers = #docStructure.pages[page].layers
        for j=1, numLayers do
            app.setCurrentLayer(j)
            local textsOnLayer = app.getTexts("layer")
            for k=1, #textsOnLayer do
                texts[index] = textsOnLayer[k].text
                index = index + 1
            end
        end
    end
    return texts
end
```

tobiasBora:
Thanks for the help. Oh I thought that `setCurrentPage` and `setCurrentLayer` were only supposed to be used when we want to actually move the UI as it would be too inefficient in practice to physically move to a page just to iterate over a simple table… but I did try it… It does work (up to minor typo fix), but without too much surprise on my side it is **super slow**: on a 180 pages document (and my final document will be even larger) with just 2 annotations, it took 15 seconds just to get all texts…

So I guess what I'd prefer is that `getDocumentStructure` should output basically the same info as the one stored in the XML when saving, this way I can iterate super quickly and access to all data. A reverse operation `setDocumentStructure` to also super quickly modify a document without iterating over all pages would also be very welcome.


Code without typo:
```
function getAllTexts()
   local texts = {}
   local index = 1
   local docStructure = app.getDocumentStructure()
   local numPages = #docStructure.pages
   for i=1, numPages do
      app.setCurrentPage(i)
[...]

PR #7035, opened 2026-01-26
Allow get{Text,Stroke,Image} to retrieve all elements in the document

tobiasBora:
Fix https://github.com/xournalpp/xournalpp/issues/7007

The diff looks ugly, but it's mostly due to re-indentation since the test was moved in getElementsFromHelper following what @rolandlo did in https://github.com/xournalpp/xournalpp/issues/7007 (not sure why, I can revert to the old behavior if needed).

### DIFF
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -376,16 +376,20 @@ function app.addStrokes(opts) end
 --- }
 function app.addTexts(opts) end
 
---- Returns a list of lua table of the texts (from current selection / current layer).
---- Is mostly inverse to app.addTexts (except getTexts will also retrieve the width/height of the textbox)
+--- Returns a list of lua table of the texts (from current selection / current layer / current page / all pages).
+--- When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for the
+--- layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
+--- index together with its layer (all of them being indexed from 1).
 --- 
---- @param type string "selection" or "layer"
+--- Is mostly inverse to app.addTexts (except getTexts may also retrieve the width/height/page/layer of the textbox)
+--- 
+--- @param type string "selection" or "layer" or "page" or "all"
 --- @return {text:string, font:{name:string, size:number}, color:integer, x:number, y:number, width:number,
---- height:number, ref:lightuserdata}[] texts
+--- height:number, ref:lightuserdata, page:number|nil, layer:number|nil}[] texts
 --- 
---- Required argument: type ("selection" or "layer")
+--- Required argument: type ("selection" or "layer" or "page" or "all")
 --- 
---- Example: local texts = app.getTexts("layer")
+--- Example: local texts = app.getTexts("all")
 --- 
 --- possible return value:
 --- {
@@ -401,6 +405,8 @@ function app.addTexts(opts) end
 ---     width = 55.0,
 ---     height = 23.0,
 ---     ref = userdata: 0x5f644c0700d0
+---     page = 1, -- Only present when called with the "all" argument
+---     layer = 1, -- Only present when called with the "all" or "page" argument
 ---   },
 ---   {
 ---     text = "Testing",
@@ -414,21 +420,27 @@ function app.addTexts(opts) end
 ---     width = 55.0,
 ---     height = 23.0,
 ---     ref = userdata: 0x5f644c0701e8
+---     page = 2,
+---     layer = 1,
 ---   },
 --- }
 --- 
 function app.getTexts(type) end
 
---- Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+--- Puts a Lua Table of the Strokes (from the selection tool / selected layer / selected page / all document) onto the
+--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
+--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
+--- index together with its layer (all of them being indexed from 1).
+--- 
 --- Is inverse to app.addStrokes
 --- 
---- @param type string "selection" or "layer"
+--- @param type string "selection" or "layer" or "page" or "all"
 --- @return {x:number[], y:number[], pressure:number[], tool:string, width:number, color:integer, fill:number,
---- linestyle:string, ref:lightuserdata}[] strokes
+--- linestyle:string, ref:lightuserdata, page:number|nil, layer:number|nil}[] strokes
 --- 
---- Required argument: type ("selection" or "layer")
+--- Required argument: type ("selection" or "layer" or "page" or "all")
 --- 
---- Example: local strokes = app.getStrokes("selection")
+--- Example: local strokes = app.getStrokes("all")
 --- 
 --- possible return value:
 --- {
@@ -443,7 +455,9 @@ function app.getTexts(type) end
 ---             ["color"] = 0xa000f0,
 ---             ["fill"] = 0,
 ---             ["lineStyle"] = "plain",
----             ["ref"] = userdata: 0x5f644c02c538
+---             ["ref"] = userdata: 0x5f644c02c538,
+---             ["page"] = 1, -- Only present when called with "all"
+---             ["layer"] = 1, -- Only present when called with "all" or "page"
 ---         },
 ---         {
 ---             ["x"]         = {207, 207.5, 315.2, 315.29, 207.5844},
@@ -454,6 +468,8 @@ function app.getTexts(type) end
 ---             ["fill"]      = -1,
 ---             ["lineStyle"] = "plain",
 ---             ["ref"] = userdata: 0x5f644c02d440
+---             ["page"] = 2,
+---             ["layer"] = 1,
 ---         },
 ---         {
 ---             ["x"]         = {387.60, 387.6042, 500.879, 500.87, 387.604},
@@ -464,6 +480,8 @@ function app.getTexts(type) end
 ---             ["fill"]      = -1,
 ---             ["lineStyle"] = "plain",
 ---             ["ref"] = userdata: 0x5f644c0700d0
+---             ["page"] = 2,
+---             ["layer"] = 2,
 ---         },
 --- }
 function app.getStrokes(type) end
@@ -880,14 +898,18 @@ function app.openFile(path, pageNr, oldDocument) end
 --- maxWidth=400}}}
 function app.addImages(opts) end
 
---- Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
+--- Puts a Lua Table of the Images (from the selection tool / selected layer / selected page / all document) onto the
+--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
+--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
+--- index together with its layer (all of them being indexed from 1).
+--- 
 --- Is inverse to app.addImages
 --- 
---- @param type string "selection" or "layer"
+--- @param type string "selection" or "layer" or "page" or "all"
 --- @return {x:number, y:number, width:number, height:number, data:string, format:string, imageWidth:number,
---- imageHeight:number, ref:lightuserdata}[] images
+--- imageHeight:number, ref:lightuserdata, page:number|nil, layer:number|nil}[] images
 --- 
---- Required argument: type ("selection" or "layer")
+--- Required argument: type ("selection" or "layer" or "page" or "all")
 --- 
 --- Example: local images = app.getImages("selection")
 --- 
@@ -903,6 +925,8 @@ function app.addImages(opts) end
 ---         ["imageWidth"] = integer,
 ---         ["imageHeight"] = integer,
 ---         ["ref"] = userdata: 0x5f644c0700d0
+---         ["page"] = 1, -- Only present when called with "all"
+---         ["layer"] = 1, -- Only present when called with "all" or "page"
 ---     },
 ---     {
 ---         ...
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -75,20 +75,81 @@ extern "C" {
 #include "undo/PageSizeChangeUndoAction.h"
 }
 
-
-static std::tuple<std::optional<std::string>, std::vector<const Element*>> getElementsFromHelper(
-        Control* control, const std::string& type) {
-    std::vector<const Element*> elements = {};
-    if (type == "layer") {
+/*
+ * Helper function used in the rest of this code to obtain elements from the document:
+ * - "type" is a string ("layer", "page", "selection" or "all") specifying if we want to retrieve, respectively,
+ * elements from the current layer, page, selection, or from all the document.
+ * - "elType" is the kind of elements we want to extract (ELEMENT_TEXT, ELEMENT_IMAGE, ELEMENT_STROKE...)
+ * It returns a tuple containing:
+ * - an (optional) string to check if an error occurred
+ * - a list of tuples containing for each element the element itself, and optionally (when type == all) its page and
+ * optionally (when type == all or type == page) its layer number, indexed from 1 for interoperability with the lua API.
+ */
+static std::tuple<std::optional<std::string>,       // Error
+                  std::vector<std::tuple<           // For each element
+                          const Element*,           // return the element
+                          std::optional<size_t>,    // its page when type == all
+                          std::optional<size_t>>>>  // its layer when type == all
+        getElementsFromHelper(Control* control, const std::string& type, ElementType elType) {
+    std::vector<std::tuple<const Element*, std::optional<size_t>, std::optional<size_t>>> elements = {};
+    if (type == "all") {
         auto sel = control->getWindow()->getXournal()->getSelection();
         if (sel) {
             control->clearSelection();  // otherwise texts in the selection won't be recognized
         }
-        elements = control->getCurrentPage()->getSelectedLayer()->getElementsView().clone();
+        auto* doc = control->getDocument();
+        for (size_t i = 0; i < doc->getPageCount(); i++) {
+            PageRef p = doc->getPage(i);
+            size_t layerID = 1;  // Layer 0 is for background and is not listed here
+            for (const Layer* l: p->getLayersView()) {
+                auto v = l->getElementsView();
+                for (auto e = v.begin(); e != v.end(); ++e) {
+                    if ((*e)->getType() == elType) {
+                        // Non-background layers and pages start at index 1
+                        elements.push_back(std::make_tuple(*e, std::make_optional(i + 1), std::make_optional(layerID)));
+                    }
+                }
+                layerID++;
+            }
+        }
+    } else if (type == "page") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            control->clearSelection();  // otherwise texts in the selection won't be recognized
+        }
+        PageRef p = control->getCurrentPage();
+        size_t layerID = 1;  // Layer 0 is for background and is not listed here
+        for (const Layer* l: p->getLayersView()) {
+            auto v = l->getElementsView();
+            for (auto e = v.begin(); e != v.end(); ++e) {
+                if ((*e)->getType() == elType) {
+                    // We index pages from 1 since this is what the Lua API assumes,
+                    // and non-background layers also start at index 1
+                    elements.push_back(std::make_tuple(*e, std::nullopt, std::make_optional(layerID)));
+                }
+            }
+            layerID++;
+        }
+    } else if (type == "layer") {
+        auto sel = control->getWindow()->getXournal()->getSelection();
+        if (sel) {
+            control->clearSelection();  // otherwise texts in the selection won't be recognized
+        }
+        auto v = control->getCurrentPage()->getSelectedLayer()->getElementsView();
+        for (auto e = v.begin(); e != v.end(); ++e) {
+            if ((*e)->getType() == elType) {
+                elements.push_back(std::make_tuple(*e, std::nullopt, std::nullopt));
+            }
+        }
     } else if (type == "selection") {
         auto sel = control->getWindow()->getXournal()->getSelection();
         if (sel) {
-            elements = sel->getElementsView().clone();
+            auto v = sel->getElementsView();
+            for (auto e = v.begin(); e != v.end(); ++e) {
+                if ((*e)->getType() == elType) {
+                    elements.push_back(std::make_tuple(*e, std::nullopt, std::nullopt));
+                }
+            }
         } else {
             return std::make_tuple(std::make_optional("There is no selection"), elements);
         }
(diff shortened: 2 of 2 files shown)


atticus-sullivan:
Indentation is ensured with `clang-format` automatically, so there aren't really multiple ways of how to format the source code.
If you want to clean up, you could split your commit into multiple ones (one for the `ElementType` change leading to the "re-indentation" (which is more of a refactor) and one for the actual new feature). This might make checking on the changes easier.

But be aware that in the end, just before merging, you'll most probably be asked to squash it all into one commit again anyhow. (so I'm not sure how much you'll really gain by splitting the commit up)

tobiasBora:
> (so I'm not sure how much you'll really gain by splitting the commit up)

Ahah I agree, I think it might be a bit overkill since the first commit would still be ugly anyway. But if someone needs it, let me know.

But the changes are rather simple:
1. I (or rather @rolandlo) added a `if (type == "all") {` case in `getElementsFromHelper` based on his comment in #7007. This function should anyway full be reviewed since I modified stuff in all sections because of the change of the test.
2. I modified slighly the documentation to add this new argument
3. I changed all 3 occurrences of `for (const Element* e: elements) {` into `for (const auto [e, page_nb, layer]: elements) {` and removed the condition (already in `getElementsFromHelper`) + re-indented the code
4. I added at the end of each loop the following code:
 ```
         if (layer) {
            lua_pushnumber(L, (double)(layer.value()));
            lua_setfield(L, -2, "layer");  // add layer to text
        }

        if (page_nb) {
            lua_pushnumber(L, (double)(page_nb.value()));
            lua_setfield(L, -2, "page");  // add page to text
        }`
```

File: src/core/plugin/luapi_application.h [hunk 1]
```diff
@@ -1542,16 +1573,17 @@ static int applib_addTexts(lua_State* L) {
...
 /**
- * Returns a list of lua table of the texts (from current selection / current layer).
- * Is mostly inverse to app.addTexts (except getTexts will also retrieve the width/height of the textbox)
+ * Returns a list of lua table of the texts (from current selection / current layer / all pages).
+ * When retrieving all texts, a page and layer field are added containing the corresponding page and layer (indexed from 1).
+ * Is mostly inverse to app.addTexts (except getTexts will also retrieve the width/height/page/layer of the textbox)
  *
- * @param type string "selection" or "layer"
+ * @param type string "selection" or "layer" or "all"
  * @return {text:string, font:{name:string, size:number}, color:integer, x:number, y:number, width:number,
```
atticus-sullivan, line 1614: here

src/core/plugin/luapi_application.h
atticus-sullivan: here

src/core/plugin/luapi_application.h
```diff
@@ -1605,61 +1641,70 @@ static int applib_getTexts(lua_State* L) {
...
 
 /**
- * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer / all document) onto the stack.
+ * When retrieving all strokes, a page and layer field are added containing the corresponding page and layer (indexed from 1).
  * Is inverse to app.addStrokes
  *
- * @param type string "selection" or "layer"
+ * @param type string "selection" or "layer" or "all"
  * @return {x:number[], y:number[], pressure:number[], tool:string, width:number, color:integer, fill:number,
```
atticus-sullivan, line 1740: here

src/core/plugin/luapi_application.h
```diff
@@ -3156,13 +3209,14 @@ static int applib_addImages(lua_State* L) {
 
 /**
  * Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
+ * When retrieving all images, a page and layer field are added containing the corresponding page and layer (indexed from 1).
  * Is inverse to app.addImages
  *
- * @param type string "selection" or "layer"
+ * @param type string "selection" or "layer" or "all"
  * @return {x:number, y:number, width:number, height:number, data:string, format:string, imageWidth:number,
```
atticus-sullivan, line 3261: here

src/core/plugin/luapi_application.h
atticus-sullivan, line 3279: here

tobiasBora:
Thanks for the review.

> I just quickly skimmed over it. I'd adjust the documentation more to fully cover the new feature / fields. So far you've only fully adjusted the docs for getTexts.

What do you mean? Except for the examples that I've not modified, I do document `"all"` in the `@param` and `Required argument`, as well as `page:number|nil, layer:number|nil` in `@return` for all 3 `get*` functions.

But I also changed the example if it's what you mean.

> Maybe in the example return values even add comments (with --) in the "lua-code" to explain the new fields are only present when the function gets called in a specific way.

Done, thanks.

> The generation of the file will have to be done manually though 

Oh good to know, I just did it!

> I'd put the union of all possible keys in the "type-hints". Checking if the key exists (e.g. if it equals nil) is then the job of the plugin writer in my opinion.

I'm really confused, why? I always prefer to have "exact" type-hints, if a field may not exist, it is good to document it and add it in the type-hint, otherwise one may think that it is always ok to use it when it is not.

src/core/plugin/luapi_application.h
tobiasBora: I'm confused, what is wrong here? (see also my comment)

atticus-sullivan: Oh sorry, my bad. I just missed somehow that the "doc-string" continues on the next line :see_no_evil:. Nevermind (and just mark the comments as resolved)

atticus-sullivan:
> I'm really confused, why? I always prefer to have "exact" type-hints, if a field may not exist, it is good to document it and add it in the type-hint, otherwise one may think that it is always ok to use it when it is not.

Yes, `number|nil` is just fine (I just overlooked it somehow). My take was just on whether to include the key at all as it may not be present. But noting it as optional is the way to go here :+1:

src/core/plugin/luapi_application.h
```diff
@@ -1605,63 +1641,72 @@ static int applib_getTexts(lua_State* L) {
...
+        lua_setfield(L, -2, "height");  // add height to text
 
-            lua_pushnumber(L, t->getElementHeight());
-            lua_setfield(L, -2, "height");  // add height to text
+        lua_pushlightuserdata(L, const_cast<void*>(static_cast<const void*>(t)));
+        lua_setfield(L, -2, "ref");
 
-            lua_pushlightuserdata(L, const_cast<void*>(static_cast<const void*>(t)));
-            lua_setfield(L, -2, "ref");
+        if (layer) {
```
atticus-sullivan: Nitpick:
Not sure how consistent we are in this regard. But above we have
```cpp
if (err.has_value()) {
```
. Which makes the "unpacking" of the optional a bit inconsistent.
(same of course with the other functions and also regarding the `page` number)

src/core/plugin/luapi_application.h
```diff
@@ -3156,13 +3216,14 @@ static int applib_addImages(lua_State* L) {
 
 /**
  * Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
```
atticus-sullivan:
```suggestion
 * Puts a Lua Table of the Images (from the selection tool / selected layer / all document) onto the stack.
```

src/core/plugin/luapi_application.h
```diff
@@ -75,20 +75,51 @@ extern "C" {
 #include "undo/PageSizeChangeUndoAction.h"
 }
 
-
-static std::tuple<std::optional<std::string>, std::vector<const Element*>> getElementsFromHelper(
-        Control* control, const std::string& type) {
-    std::vector<const Element*> elements = {};
-    if (type == "layer") {
+static std::tuple<std::optional<std::string>, std::vector<std::tuple<const Element*, std::optional<size_t>, std::optional<size_t>>>> getElementsFromHelper(
```
atticus-sullivan: Maybe we should add some comment to this function describing what that return type is. It is not apparent what the `optional<size_t>` values are (at least for me, the whole nesting with the brackets and wrapped in optionals didn't help reading). Maybe this could also be solved by formatting the line differently (but comments are always good). Something like
```cpp
/*
 * Obtains elements from the document (filtered by elType and type)  and returns the elements each optionally associated with its page and layer.
 * For lua interoperability, errors are returned via the optional string in the outmost tuple.
 *
 * Before making use of the returned value, the error must be checked.
*/
```

Also:
Are you sure you ran `clang-format`? Just because this line looks a bit long

tobiasBora: Oh my bad, I thought this was done automatically by a bot after merging. See my latest push.

src/core/plugin/luapi_application.h
tobiasBora: Thanks, I applied this.

src/core/plugin/luapi_application.h
```diff
@@ -1722,82 +1774,90 @@ static int applib_getStrokes(lua_State* L) {
...
 
-            lua_pushstring(L, StrokeStyle::formatStyle(s->getLineStyle()).c_str());
-            lua_setfield(L, -2, "lineStyle");  // add linestyle to stroke
+        lua_pushlightuserdata(L, const_cast<void*>(static_cast<const void*>(s)));
+        lua_setfield(L, -2, "ref");
 
-            lua_pushlightuserdata(L, const_cast<void*>(static_cast<const void*>(s)));
-            lua_setfield(L, -2, "ref");
+        if (layer) {
+            lua_pushnumber(L, (double)(layer.value()));
```
rolandlo: Why are you pushing the layer (and page) as numbers and not as integers? That doesn't seem to make sense.

tobiasBora: Oh, i forgot that recent lua has a notion of integer, I was stuck back when lua made no difference between integers and floats. Changing that right away.

tobiasBora:
> Thinking of the different "filters" that are implemented now (selection, layer and all), it seems to me like "page" is the only missing one. Any thoughts on this? (if we opt implementing it, this might also be a separate PR)

It was trivial enough that I included this in my latest push. I also applied all your other comments.

@rolandlo the latest push also contains the integer fix, it should be good to review now.

tobiasBora:
Wait, seems like this is not the latest version

tobiasBora:
Ah, after a few iterations (forgot a few documentation & to run the update script), here we go ^^ Sorry for the numerous pushes.

plugins/luapi_application.def.lua
```diff
@@ -443,7 +455,9 @@ function app.getTexts(type) end
 ---             ["color"] = 0xa000f0,
 ---             ["fill"] = 0,
 ---             ["lineStyle"] = "plain",
----             ["ref"] = userdata: 0x5f644c02c538
+---             ["ref"] = userdata: 0x5f644c02c538,
+---             ["page"] = 1.0, -- Only present when called with "all"
+---             ["layer"] = 1.0, -- Only present when called with "all" or "page"
```
rolandlo: Now that you push them as an integer.
```suggestion
---             ["page"] = 1, -- Only present when called with "all"
---             ["layer"] = 1, -- Only present when called with "all" or "page"
```

plugins/luapi_application.def.lua
```diff
@@ -454,6 +468,8 @@ function app.getTexts(type) end
 ---             ["fill"]      = -1,
 ---             ["lineStyle"] = "plain",
 ---             ["ref"] = userdata: 0x5f644c02d440
+---             ["page"] = 2.0,
+---             ["layer"] = 1.0,
```
rolandlo:
```suggestion
---             ["page"] = 2,
---             ["layer"] = 1,
```

plugins/luapi_application.def.lua
```diff
@@ -464,6 +480,8 @@ function app.getTexts(type) end
 ---             ["fill"]      = -1,
 ---             ["lineStyle"] = "plain",
 ---             ["ref"] = userdata: 0x5f644c0700d0
+---             ["page"] = 2.0,
+---             ["layer"] = 2.0,
```
rolandlo:
```suggestion
---             ["page"] = 2,
---             ["layer"] = 2,
```

plugins/luapi_application.def.lua
```diff
@@ -880,14 +898,18 @@ function app.openFile(path, pageNr, oldDocument) end
 --- maxWidth=400}}}
 function app.addImages(opts) end
 
---- Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
+--- Puts a Lua Table of the Images (from the selection tool / selected layer / selected page / all document) onto the
+--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" containing
+--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```
rolandlo:
```suggestion
--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```

plugins/luapi_application.def.lua
```diff
@@ -903,6 +925,8 @@ function app.addImages(opts) end
 ---         ["imageWidth"] = integer,
 ---         ["imageHeight"] = integer,
 ---         ["ref"] = userdata: 0x5f644c0700d0
+---         ["page"] = 1.0, -- Only present when called with "all"
+---         ["layer"] = 1.0, -- Only present when called with "all" or "page"
```
rolandlo:
```suggestion
---         ["page"] = 1, -- Only present when called with "all"
---         ["layer"] = 1, -- Only present when called with "all" or "page"
```

plugins/luapi_application.def.lua
```diff
@@ -414,21 +420,27 @@ function app.addTexts(opts) end
...
+---     layer = 1.0,
 ---   },
 --- }
 --- 
 function app.getTexts(type) end
 
---- Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+--- Puts a Lua Table of the Strokes (from the selection tool / selected layer / selected page / all document) onto the
+--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" containing
+--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```
rolandlo:
```suggestion
--- stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
--- the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```

plugins/luapi_application.def.lua
```diff
@@ -401,6 +405,8 @@ function app.addTexts(opts) end
 ---     width = 55.0,
 ---     height = 23.0,
 ---     ref = userdata: 0x5f644c0700d0
+---     page = 1.0, -- Only present when called with the "all" argument
+---     layer = 1.0, -- Only present when called with the "all" or "page" argument
```
rolandlo:
```suggestion
---     page = 1, -- Only present when called with the "all" argument
---     layer = 1, -- Only present when called with the "all" or "page" argument
```

plugins/luapi_application.def.lua
```diff
@@ -414,21 +420,27 @@ function app.addTexts(opts) end
 ---     width = 55.0,
 ---     height = 23.0,
 ---     ref = userdata: 0x5f644c0701e8
+---     page = 2.0,
+---     layer = 1.0,
```
rolandlo:
```suggestion
---     page = 2,
---     layer = 1,
```

src/core/plugin/luapi_application.h
```diff
@@ -1542,16 +1604,20 @@ static int applib_addTexts(lua_State* L) {
 }
 
 /**
- * Returns a list of lua table of the texts (from current selection / current layer).
- * Is mostly inverse to app.addTexts (except getTexts will also retrieve the width/height of the textbox)
+ * Returns a list of lua table of the texts (from current selection / current layer / current page / all pages).
+ * When called with "page" to retrieve all elements on the current page, it also adds a field "layer" containing the
+ * layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```
rolandlo:
```suggestion
 * When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for the
 * layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```

src/core/plugin/luapi_application.h
```diff
@@ -1605,63 +1675,75 @@ static int applib_getTexts(lua_State* L) {
...
+        lua_settable(L, -3);  // add text to elements
     }
     return 1;
 }
 
 /**
- * Puts a Lua Table of the Strokes (from the selection tool / selected layer) onto the stack.
+ * Puts a Lua Table of the Strokes (from the selection tool / selected layer / selected page / all document) onto the
+ * stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" containing
+ * the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```
rolandlo:
```suggestion
 * stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
 * the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```

src/core/plugin/luapi_application.h
```diff
@@ -1676,7 +1758,9 @@ static int applib_getTexts(lua_State* L) {
  *             ["color"] = 0xa000f0,
  *             ["fill"] = 0,
  *             ["lineStyle"] = "plain",
- *             ["ref"] = userdata: 0x5f644c02c538
+ *             ["ref"] = userdata: 0x5f644c02c538,
+ *             ["page"] = 1.0, -- Only present when called with "all"
+ *             ["layer"] = 1.0, -- Only present when called with "all" or "page"
```
rolandlo:
```suggestion
 *             ["page"] = 1, -- Only present when called with "all"
 *             ["layer"] = 1, -- Only present when called with "all" or "page"
```

src/core/plugin/luapi_application.h
```diff
@@ -1687,6 +1771,8 @@ static int applib_getTexts(lua_State* L) {
  *             ["fill"]      = -1,
  *             ["lineStyle"] = "plain",
  *             ["ref"] = userdata: 0x5f644c02d440
+ *             ["page"] = 2.0,
+ *             ["layer"] = 1.0,
```
rolandlo:
```suggestion
 *             ["page"] = 2,
 *             ["layer"] = 1,
```

src/core/plugin/luapi_application.h
```diff
@@ -1697,6 +1783,8 @@ static int applib_getTexts(lua_State* L) {
  *             ["fill"]      = -1,
  *             ["lineStyle"] = "plain",
  *             ["ref"] = userdata: 0x5f644c0700d0
+ *             ["page"] = 2.0,
+ *             ["layer"] = 2.0,
```
rolandlo:
```suggestion
 *             ["page"] = 2,
 *             ["layer"] = 2,
```

src/core/plugin/luapi_application.h
```diff
@@ -3155,14 +3251,18 @@ static int applib_addImages(lua_State* L) {
 }
 
 /**
- * Puts a Lua Table of the Images (from the selection tool / selected layer) onto the stack.
+ * Puts a Lua Table of the Images (from the selection tool / selected layer / selected page / all document) onto the
+ * stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" containing
+ * the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```
rolandlo:
```suggestion
 * stack. When called with "page" to retrieve all elements on the current page, it also adds a field "layer" for
 * the layer containing the element, and when called with "all" it additionally adds a field "page" containing its page
```

src/core/plugin/luapi_application.h
```diff
@@ -3178,6 +3278,8 @@ static int applib_addImages(lua_State* L) {
  *         ["imageWidth"] = integer,
  *         ["imageHeight"] = integer,
  *         ["ref"] = userdata: 0x5f644c0700d0
+ *         ["page"] = 1.0, -- Only present when called with "all"
+ *         ["layer"] = 1.0, -- Only present when called with "all" or "page"
```
rolandlo:
```suggestion
 *         ["page"] = 1, -- Only present when called with "all"
 *         ["layer"] = 1, -- Only present when called with "all" or "page"
```

src/core/plugin/luapi_application.h
```diff
@@ -1580,6 +1648,8 @@ static int applib_addTexts(lua_State* L) {
  *     width = 55.0,
  *     height = 23.0,
  *     ref = userdata: 0x5f644c0701e8
+ *     page = 2.0,
+ *     layer = 1.0,
```
rolandlo:
```suggestion
 *     page = 2,
 *     layer = 1,
```

src/core/plugin/luapi_application.h
```diff
@@ -1567,6 +1633,8 @@ static int applib_addTexts(lua_State* L) {
  *     width = 55.0,
  *     height = 23.0,
  *     ref = userdata: 0x5f644c0700d0
+ *     page = 1.0, -- Only present when called with the "all" argument
+ *     layer = 1.0, -- Only present when called with the "all" or "page" argument
```
rolandlo:
```suggestion
 *     page = 1, -- Only present when called with the "all" argument
 *     layer = 1, -- Only present when called with the "all" or "page" argument
```

src/core/plugin/luapi_application.h
```diff
@@ -75,20 +75,82 @@ extern "C" {
...
-    std::vector<const Element*> elements = {};
-    if (type == "layer") {
+/*
+ * Helper function used in the rest of this code to obtain elements from the document:
+ * - "type" is a string ("layer", "page", "selection" or "all") specifying if we want to retrieve, respectively,
+ * elements from the current layer, page, selection, or from all the document.
+ * - "elType" is the kind of elements we want to extract (ELEMENT_TEXT, ELEMENT_IMAGE, ELEMENT_STROKE...)
+ * It returns a tuple containing:
+ * - an (optional) string to check if an error occurred
+ * - a list of tuple containing for each element the element it self, and optionally (when type == all) its page and
```
rolandlo:
```suggestion
 * - a list of tuples containing for each element the element itself, and optionally (when type == all) its page and
```

src/core/plugin/luapi_application.h
```diff
@@ -75,20 +75,82 @@ extern "C" {
...
+            control->clearSelection();  // otherwise texts in the selection won't be recognized
+        }
+        PageRef p = control->getCurrentPage();
+        size_t layerID = 1;  // Layer 0 is for background and is not listed here
+        for (const Layer* l: p->getLayersView()) {
+            auto v = l->getElementsView();
+            for (auto e = v.begin(); e != v.end(); ++e) {
+                if ((*e)->getType() == elType) {
+                    // We index pages from 1 since this is what the Lua API assumes,
+                    // and non-background layers also start at index 1
```
rolandlo, line 127:
```suggestion
                    // Non-background layers start at index 1
```

rolandlo (commented):
Except of some issues with the comments (highlighted in the code) this looks good to me.

tobiasBora:
Thanks, see latest version with the final squashed commit.

tobiasBora:
Arg, I checked like 3 times without seeing this, really sorry… This time it should be good enough I hope!

rolandlo:
Thanks @tobiasBora for this addition to the Lua API. Thanks @atticus-sullivan for helping with the review.

tobiasBora:
@rolandlo Great, thanks a lot for the work!

PR #7057, opened 2026-01-31
Store parsed PageTemplateSettings in Settings

tmoerschell:
- Use parsed template everywhere and remove repeated parsing
- Correctly initialize pageTypeForNewPages

This goes a little bit further than #7055, but no functionality is modified.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -666,10 +666,8 @@ void Control::setShowMenubar(bool enabled) {
 
 void Control::disableSidebarTmp(bool disabled) { this->sidebar->setTmpDisabled(disabled); }
 
-void Control::addDefaultPage(const std::optional<std::string>& pageTemplate, Document* doc) {
-    const std::string& templ = pageTemplate.value_or(settings->getPageTemplate());
-    PageTemplateSettings model;
-    model.parse(templ);
+void Control::addDefaultPage(const std::optional<PageTemplateSettings>& pageTemplate, Document* doc) {
+    const auto& model = pageTemplate.value_or(this->settings->getPageTemplateSettings());
 
     auto page = std::make_shared<XojPage>(model.getPageWidth(), model.getPageHeight());
     page->setBackgroundColor(model.getBackgroundColor());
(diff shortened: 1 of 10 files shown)


src/core/control/Control.cpp
```diff
@@ -666,10 +666,8 @@ void Control::setShowMenubar(bool enabled) {
 
 void Control::disableSidebarTmp(bool disabled) { this->sidebar->setTmpDisabled(disabled); }
 
-void Control::addDefaultPage(const std::optional<std::string>& pageTemplate, Document* doc) {
-    const std::string& templ = pageTemplate.value_or(settings->getPageTemplate());
-    PageTemplateSettings model;
-    model.parse(templ);
+void Control::addDefaultPage(const std::optional<PageTemplateSettings>& pageTemplate, Document* doc) {
+    const auto& model = pageTemplate ? *pageTemplate : this->settings->getPageTemplateSettings();
```
bhennion: You can use std::optional<>::value_or() for this

src/core/control/Control.cpp
```diff
@@ -1648,7 +1646,9 @@ bool Control::openXoptFile(fs::path filepath) {
         // Unable to read the template from the file
         return false;
     }
-    this->replaceDocument(createNewDocument(this, std::move(filepath), pageTemplate), -1);
+    PageTemplateSettings model;
+    model.parse(*pageTemplate);
```
bhennion: Can this fail? It would be good to have a safety here.

tmoerschell: Yes, if the first line is not `xoj/template`. Previously the page type was left unset, so the default background was opened. Better to display an error there.

I've discovered 2 related things, though they should go into another PR:
 * `Util::readString` is supposed to display an error message on failure, but no more `filesystem_error` is thrown since 00c79116884d6dfe64bc53534de1f9f3732a6190
 * It is possible to trigger an uncaught exception (from `fs::exists`) by opening a file in a directory we do not have permissions for, e.g. `xournalpp /root/whatever`.

bhennion (approved):
LGTM!
Thanks a lot!

PR #7060, opened 2026-02-01
Fix sidebar content not taking whole width when changing tabs 

### LINKED ISSUE #6653: Sidebar layout does not use all the room when switching tabs
1. Open a PDF with lots of pages
2. Open the sidebar with page previews and make the sidebar small in width
3. Switch to the layer previews and make the sidebar large
4. Switch back to page previews and observe that most space is unused

_Originally posted by @rolandlo in https://github.com/xournalpp/xournalpp/pull/6642#pullrequestreview-3171773171_
 

Affects both the release and master branches.

RayyFort:
#### Problem:
Switching away from a  sidebar tab, resizing the sidebar, and returning causes columns in the sidebar tab to display incorrectly, keeping their original width

#### Solution:
Defer layout() using g_idle_add() to run after GTK finishes size allocation.

### DIFF
--- a/src/core/gui/sidebar/Sidebar.cpp
+++ b/src/core/gui/sidebar/Sidebar.cpp
@@ -103,6 +103,15 @@ void Sidebar::setSelectedTab(size_t tab) {
 
         i++;
     }
+    if (this->visibleTab) {
+        g_idle_add(
+                [](gpointer data) -> gboolean {
+                    Sidebar* sidebar = static_cast<Sidebar*>(data);
+                    sidebar->layout();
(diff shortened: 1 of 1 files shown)


bhennion:
Thanks for the fix!

The code look good to me, but would you mind rebasing your PR to target the release-1.3 branch (so the fix will appear in v1.3.2)? You can do that by running `git rebase --onto release-1.3 master` and then change the PR target after clicking on the Edit button on top of the page.

bhennion (approved):

PR #7073, opened 2026-02-04
Up librsvg minimal version to 2.46

### LINKED ISSUE #7070: Build is broken with C version of librsvg due to xournalpp-thumbnailer.cpp
### Operating System

Mac OS

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

MacPorts

### Xournal++ Version

1.3.2

### libgtk Version

3.24.51

### Bug Description

A commit intending to get rid of some warning broke compatibility with non-rusty `librsvg`: https://github.com/xournalpp/xournalpp/commit/63e72cf85bc2e47cec8140b0fcc4ea10f6b8bc01
The build now fails due to `RsvgRectangle` and `viewport` not being defined.

### Expected Behaviour

Fixing warnings is not a worth reason to break compatibility, IMO. At this point non-rusty `librsvg` can likely be supported with no issues.
If that becomes unfeasible, at least configure should fail. Otherwise users waste a lot of time on compilation – only to see a failure in the end.

### Steps to Reproduce

Run the build with C version of `librsvg`.

### Additional Context

_No response_

bhennion:
Fix #7070

We are currently using rsvg_handle_render_document() which was introduced in v2.46

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ endif (APPLE)
 list(APPEND pkg-module-spec "gthread-2.0 >= 2.4.0")
 list(APPEND pkg-module-spec "libxml-2.0 >= 2.0.0")
 list(APPEND pkg-module-spec "libzip >= 1.0.1")
-list(APPEND pkg-module-spec "librsvg-2.0 >= 2.40")
+list(APPEND pkg-module-spec "librsvg-2.0 >= 2.46")
 
 # Profiling
 option (ENABLE_PROFILING "Link with gperftools" OFF)


barracuda156:
It will be a far better fix to add a fallback. We are talking about 7 lines of code in a single file.

barracuda156:
Well, maybe I make I PR with an actual fix then? It’s not hard to restore compatible code, there are version macros for that: https://www.manpagez.com/html/rsvg-2.0/rsvg-2.0-2.52.1/rsvg-Version-checks.php

bhennion:
> Well, maybe I make I PR with an actual fix then? It’s not hard to restore compatible code, there are version macros for that: https://www.manpagez.com/html/rsvg-2.0/rsvg-2.0-2.52.1/rsvg-Version-checks.php

You are more than welcome to make such a PR. I think we could merge it.
But to be honest, this or something else may get broken again in the future (we could bump another dep which indirectly relies on a more recent version of librsvg, for instance). Purposefully maintaining compatibilty with a decade-old unmaintained library is not something we should or would spend time on.

PR #7081, opened 2026-02-05
Improve Layout performances

bhennion:
With this PR:
    * The Layout precalculated data now only needs to be updated when a page is added/removed or has its size changed, or when the layout (nb of columns/rows) is changed. It is no longer updated when scrolling or zooming in or out.
    * There is no longer any loop on all the pages (except for the precalculated data) in Layout or gtk_xournal_draw.

The performance improvements are obvious (especially when using a file with a lot of pages).
On this video, the left side is the current master, the right side is this PR:

[Enregistrement d'écran_20260205_223503.webm](https://github.com/user-attachments/assets/c32bc1f5-13a2-488d-b8a6-952d71b166da)

For this test, I used the PDF specification document, 756 pages.

### DIFF
--- a/src/core/gui/Layout.cpp
+++ b/src/core/gui/Layout.cpp
@@ -525,7 +525,9 @@ static xoj::util::Point<int> getPixelCoords(const Layout::PreCalculated& pc, con
                             zoom) +
                     strict_cast<int>(pos.col) * XOURNAL_PADDING_BETWEEN + pc.paddingLeft +
                     pc.horizontalCenteringPadding,
-            floor_cast<int>((pos.row == 0 ? 0. : pc.stretchableVerticalPixelsAfterRow[pos.row - 1]) * zoom) +
+            floor_cast<int>(((pos.row == 0 ? 0. : pc.stretchableVerticalPixelsAfterRow[pos.row - 1]) +
+                             .5 * (pc.heightRows[pos.row] - pv.getHeight())) *
+                            zoom) +
                     strict_cast<int>(pos.row) * XOURNAL_PADDING_BETWEEN + pc.paddingTop + pc.verticalCenteringPadding};
 }
 


jclsn:
Sounds good :)

bhennion:
> The layout has changed though for documents using different page formats. Is that wanted?

It wasn't changed on purpose (and I hadn't noticed). I thought naively the layout was making a grid (as the code semantics suggests) and adding padding to center every page in its grid entry. This is what I implemented in the new version. Thanks for testing this (admitedly awkward) case and pointing the difference out!

Now that I've played a bit more with weird situations like this, I am all the more puzzled. I cannot see any decernable pattern in what the master version is doing... Sometimes a padding is added, sometimes none, sometimes only a fraction of what I'd expect.
The first column (or row) also seems to behave differently from the others.
All in all, it does not look like an intentional behaviour to me, but rather like the result of decades of spaghetti-code.

So I think I'm ready to argue that the new predictable behaviour is better than the current master. Of course, if you have either an explanation for the current behaviour, or a better behaviour to suggest, I'll happily try to adapt the PR to those.

rolandlo:
So this PR fixes #5268 raised by @tmoerschell. There was even a PR by him #5270, which we somehow forgot about.

tmoerschell:
The current behavior can be changed to act as a grid by simply removing a `/ 2.0`, so I'm pretty sure this was a typo. See https://github.com/xournalpp/xournalpp/commit/637d04b3417e52b9d426ca18f20462aa2897fd9c (first commit of my PR).

The constant padding option I had proposed back then is quite complex, so if you're also on board with the grid layout, I think that's the best solution.

bhennion:
> So this PR fixes #5268 raised by @tmoerschell. There was even a PR by him #5270, which we somehow forgot about.

Wow, I completely forgot about this. Sorry @tmoerschell! Now that I read back #5270's thread, I think I had not understood what you meant by constant padding... Really sorry. If I understand it correctly now, I think a version of it could be implemented while preserving the performance gains of this PR (out of scope for this PR though).

> The current behavior can be changed to act as a grid by simply removing a `/ 2.0`, so I'm pretty sure this was a typo. See [637d04b](https://github.com/xournalpp/xournalpp/commit/637d04b3417e52b9d426ca18f20462aa2897fd9c) (first commit of my PR).

Indeed! This is very good. So the only difference between this and "corrected" master is vertical centering of the pages in the grid entries. Easily fixable if we want it fixed. Do we?

bhennion:
> See [637d04b](https://github.com/xournalpp/xournalpp/commit/637d04b3417e52b9d426ca18f20462aa2897fd9c) (first commit of my PR).

We may also want to add this commit to the release branch. @tmoerschell do you want to make a PR for that?

tmoerschell:
> We may also want to add this commit to the release branch. @tmoerschell do you want to make a PR for that?

Sure, will do!

> Easily fixable if we want it fixed. Do we?

I personally think that centering the pages vertically looks better and will not break any use case

rolandlo:
> > We may also want to add this commit to the release branch. @tmoerschell do you want to make a PR for that?
> 
> Sure, will do!
> 
> > Easily fixable if we want it fixed. Do we?
> 
> I personally think that centering the pages vertically looks better and will not break any use case

Fine then let's make that fix on the release branch and merge the release branch into master before merging this PR.

bhennion:
Rebased on master
The failed clang-format test is a fluke

rolandlo:
@bhennion Can you add the vertical centering to the commit description? It seems important to document this change in layout. Other than that I'm fine with this PR.

bhennion:
I split the commit in two: the first commit does not alter the behaviour, the second commit centers the pages vertically.

Thanks for the review, testing and all that! Merging in 48 hours unless an objection is raised.

rolandlo:
> I split the commit in two: the first commit does not alter the behaviour, the second commit centers the pages vertically.
> 
Ok, that's even better.

rolandlo (approved):

bhennion:
Edit: this was to make sure the create-installers action worked. Has nothing to do with this PR in particular.
\action create-installers all

Issue #6748, opened 2025-10-15
PDF files filter in Annotate PDF - Open File not working properly, only shows PDFs when "All" files is selected


IkeEisenhauer:
### Operating System

Windows

### (Linux Only) Display Server

None

### Installation Method

Official Release

### Xournal++ Version

1.2.8+dev

### libgtk Version

3.24.43

### Bug Description

PDF files do not appear in the chooser window when using Annotate PDF with the PDF filter set.
They do show up if you set filter to ALL

### Expected Behaviour

1. Click Annotate PDF
2. PDF files show up in chooser window when PDF filter is on

### Steps to Reproduce

1. Click Annotate PDF
2. Change directory to where known PDFs are with PDF fitler on
3. No PDF show
4. Change filter to ALL
5. PDFs appear

### Additional Context

This might be due to changing the default PDF reader app from Adobe Acrobat to Foxit Editor. Did not notice the problem until I did that, worked properly however when default reader was Foxit Reader

rolandlo:
In my case I see all files in some folders (even .exe files), whereas the filters work fine in different folders.
Moreover I get warnings like in this yet unfixed GLib issue: https://gitlab.gnome.org/GNOME/glib/-/issues/3279

Issue #5033, opened 2023-08-10
Xournal++ troubles


ramaswamysriram:
When I open a pdf file in Xournal++ I get error messages and a prompt to send a bug report.

rolandlo:
And what are the error messages and errorlog?

ramaswamysriram:
Thanks for responding so fast.

Error message on the screen when I open the pdf file with Xournal++ reads:

There are errorlogfiles from Xournal++. Please find a Bugreport, so the bug may be fixed.
You're using "xournalapp/HEAD" branch. The most recent log file name: /home/sriram/snap/xournalapp/common/errorlogs/errorlog.20220715-194331.log

Error logfile:

Date: Fri Jul 15 19:43:31 2022
Error: signal 11
[bt]: (0) /snap/xournalpp/69/usr/bin/xournalpp(+0x435b8a) [0x5611cea47b8a]
[bt]: (1) /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7f94e1da8090]
[bt]: (2) /snap/xournalpp/69/usr/bin/xournalpp(_ZNK7Element7getTypeEv+0x10) [0x5611cea06a52]
[bt]: (3) /snap/xournalpp/69/usr/bin/xournalpp(_ZNK12DocumentView11drawElementEP6_cairoP7Element+0x24) [0x5611cea3a3d2]
[bt]: (4) /snap/xournalpp/69/usr/bin/xournalpp(_ZN12DocumentView9drawLayerEP6_cairoP5Layer+0x13a) [0x5611cea3a67c]
[bt]: (5) /snap/xournalpp/69/usr/bin/xournalpp(_ZN12DocumentView8drawPageESt10shared_ptrI7XojPageEP6_cairobbbb+0x16e) [0x5611cea3aeea]
[bt]: (6) /snap/xournalpp/69/usr/bin/xournalpp(_ZN10PreviewJob8drawPageEv+0x1fe) [0x5611ce8c2666]
[bt]: (7) /snap/xournalpp/69/usr/bin/xournalpp(_ZN10PreviewJob3runEv+0x4d) [0x5611ce8c2a25]
[bt]: (8) /snap/xournalpp/69/usr/bin/xournalpp(_ZN3Job7executeEv+0x27) [0x5611ce8c19f5]
[...]

Issue #29, opened 2013-09-17
Catch Addpoppler Branch up to Master


wbrenna:
Addpoppler has deviated from Master because of the difference in automake files. Once these are properly copied from Master, the branches can again be merged to keep addpoppler up to date with changes.

Issue #28, opened 2013-09-14
Add Colour to Insert Mathtex


wbrenna:
In util/LatexAction.cpp a colour option could be added to choose the Mathtex colour.

andreasb242:
Already possible, e.g. write \color{red}red

Therefore i close this, because of age

alexeymuranov:
Latex tool wishlist #1115:
> Rendered font color should be the color selected in the toolbar

PR #7088, opened 2026-02-06
Implement GtkScrollable for GtkXournal widget

### LINKED ISSUE #5333: XournalWidget blacks out with many pages on certain zoom levels
### Operating System

Linux

### (Linux only) Distribution

Ubuntu

GNOME

Wayland

### Installation Method

PPA

### Xournal++ Version

all

### libgtk Version

3.24.38

### Bug Description

When using a xopp-document with several hundreds or thousands of pages (with or without PDF background) the XournalWidget goes black on certain zoom levels. One would expect that this occurs only on high zoom levels. It's different though: When I increase the zoom in steps from minimum to maximum, on some zoom levels everything displays fine, whereas on other zoom levels, it goes black. For example in `graph.xopp` contained here [graph.zip](https://github.com/xournalpp/xournalpp/files/13325110/graph.zip) I have 7586 empty pages with graph background, all with the same size:
```xml
<page width="612" height="792">
<background type="solid" color="#ffffffff" style="graph"/>
<layer/>
```
- Using the (default) 1-column layout (and zooming via CTRL+plus) I get for instance (the values are not always the same)
  - correct display on zoom levels: 30%-58%, 125-183%, 268%-295%, 393%-432%, 523%, 633%
  - black widget on zoom levels: 64%-113%, 201%-244%, 325%-357%, 475%, 575%, 696-700%
  
[BlackingOut.webm](https://github.com/xournalpp/xournalpp/assets/40485433/73e79ea0-ea46-4acd-b019-bcec3511b0c4)

The same happens when I open a PDF with the same number of pages and page sizes (the current [stacks project reference book](https://stacks.math.columbia.edu/download/book.pdf) in this example).

The might be some bug in the layout system. I have also observed that when the widget is rendered black, the `gtk_xournal_draw` is not executed, even though the layout is recalculated. 

### Expected Behaviour

No blacking out.

### Steps to Reproduce

1. Open a xopp-document (like `graph.xopp` above) with thouands of pages
2. zoom in or out and observe the blacking out like in the screen recording

### Additional Context

_No response_

bhennion:
Fix #5333 

This fixes the issue as described in https://github.com/xournalpp/xournalpp/issues/5333#issuecomment-3239250724

~~This PR is based on #7081 and will wait until #7081 is merged.~~

### DIFF
--- a/src/core/control/zoom/ZoomControl.cpp
+++ b/src/core/control/zoom/ZoomControl.cpp
@@ -35,8 +35,9 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
                 (event->direction == GDK_SCROLL_UP || (event->direction == GDK_SCROLL_SMOOTH && event->delta_y < 0)) ?
                         ZOOM_IN :
                         ZOOM_OUT;
-        // translate absolute window coordinates to the widget-local coordinates and start zooming
-        zoom->zoomScroll(direction, Util::toWidgetCoords(widget, xoj::util::Point{event->x_root, event->y_root}));
+        xoj::util::Point<double> center;
+        gdk_event_get_coords((GdkEvent*)event, &center.x, &center.y);
+        zoom->zoomScroll(direction, center);
         return true;
     }
 
@@ -47,13 +48,15 @@ auto onScrolledwindowMainScrollEvent(GtkWidget* widget, GdkEventScroll* event, Z
 auto onTouchpadPinchEvent(GtkWidget* widget, GdkEventTouchpadPinch* event, ZoomControl* zoom) -> bool {
     if (event->type == GDK_TOUCHPAD_PINCH && event->n_fingers == 2) {
         switch (event->phase) {
-            case GDK_TOUCHPAD_GESTURE_PHASE_BEGIN:
+            case GDK_TOUCHPAD_GESTURE_PHASE_BEGIN: {
                 if (zoom->isZoomFitMode()) {
                     zoom->setZoomFitMode(false);
                 }
-                // translate absolute window coordinates to the widget-local coordinates and start zooming
-                zoom->startZoomSequence(Util::toWidgetCoords(widget, xoj::util::Point{event->x_root, event->y_root}));
+                xoj::util::Point<double> center;
+                gdk_event_get_coords((GdkEvent*)event, &center.x, &center.y);
+                zoom->startZoomSequence(center);
                 break;
+            }
             case GDK_TOUCHPAD_GESTURE_PHASE_UPDATE:
                 zoom->zoomSequenceChange(event->scale, true);
                 break;
--- a/src/core/gui/MainWindow.cpp
+++ b/src/core/gui/MainWindow.cpp
@@ -338,10 +338,6 @@ void MainWindow::setGtkTouchscreenScrollingEnabled(bool enabled) {
 
 auto MainWindow::getLayout() const -> Layout* { return this->xournal->getLayout(); }
 
-auto MainWindow::getNegativeXournalWidgetPos() const -> xoj::util::Point<double> {
-    return Util::toWidgetCoords(this->winXournal, xoj::util::Point{0.0, 0.0});
-}
-
 auto cancellable_cancel(GCancellable* cancel) -> bool {
     g_cancellable_cancel(cancel);
 
--- a/src/core/gui/MainWindow.h
+++ b/src/core/gui/MainWindow.h
@@ -113,14 +113,6 @@ public:
     [[maybe_unused]] Menubar* getMenubar() const;
 
     /**
-     * Get the position of the top left corner of screen (X11) or the window (Wayland)
-     * relative to the Xournal Widget top left corner
-     *
-     * @see Util::toWidgetCoords()
-     */
-    xoj::util::Point<double> getNegativeXournalWidgetPos() const;
-
-    /**
      * Disable kinetic scrolling if there is a touchscreen device that was manually mapped to another enabled input
      * device class. This is required so the GtkScrolledWindow does not swallow all the events.
      */
--- a/src/core/gui/inputdevices/TouchInputHandler.cpp
+++ b/src/core/gui/inputdevices/TouchInputHandler.cpp
@@ -156,11 +156,6 @@ void TouchInputHandler::zoomStart() {
     auto center = (this->priLastAbs + this->secLastAbs) / 2.0;
     this->lastZoomScrollCenter = center;
 
-    // translate absolute window coordinates to the widget-local coordinates
-    const auto* mainWindow = inputContext->getView()->getControl()->getWindow();
-    const auto translation = mainWindow->getNegativeXournalWidgetPos();
-    center += translation;
-
     zoomControl->startZoomSequence(center);
 
     this->startZoomReady = false;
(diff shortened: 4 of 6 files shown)


bhennion:
This PR deserves an explanation, as I think what's happening here is deeper than what the diff might suggest.

Before this PR, the widget structure was: GtkScrolledWindow -> GtkViewPort -> GtkXournal
With this PR, it becomes: GtkScrolledWindow -> GtkXournal

Removing the intermediate GtkViewPort is only possible because GtkXournal now implements the GtkScrollable interface.
It has several effects:
1. #5333 is fixed: not sure where the bug came from. Somewhere in the GtkViewPort code I guess
2. The GtkViewPort had a buffer (GtkPixelCache) caching a rendered part of the GtkXournal. So this removes this second buffer (we already cache the rendered pages), probably improving performances a little bit.
3. More subtly, concerning the GtkXournal instance
    1. Before this PR:
        1. It was (virtually) allocated a potentially huge area, corresponding to the (zoom dependent) total size of the Layout.
        2. Its widget coordinates were the same as the Layout pixel coordinates.
        3. Its gtk_xournal_draw() function was called at the discretion of the GtkViewPort implementation: when a piece of it needed to be refreshed, or when we scrolled enough so that the GtkViewPort buffer wasn't sufficient anymore.
    2. After the PR:
        1. It has a small allocation, corresponding to the visible area. This allocation never changes when scrolling/zooming.
        2. Its widget-coordinates are now the same as the GtkScrolledWindow's coordinates.
        3. Its gtk_xournal_draw() function is called exactly when we ask it to be (via a `gtk_widget_queue_draw()`). This means we have to call it by hand when we scroll.

Overall, those are, I think, positive effects. We have less Gtk callbacks due to reallocation. Also, the size of the Layout is no longer relevent from the GtkWidget point of view: we can allow it to be as big as necessary (and maybe down the line, infinite).
Still, the change of behaviour of the widget coordinates is worth highlighting.
It may also be that some comments in the code are now referring to the wrong coordinates. I'll try to search for those in the code.

bhennion:
> There is an issue with PDF hyperlinks. They don't work in this PR when clicking on them while having the PDF selection tool activated.

Indeed it was! The (Pdf)FloatingToolbox were also misplaced. I hope I got them all. I also cleaned up some comments, to make clearer what coordinate system was used. Here is a drawing of the different coordinate systems we often encounter in the code (after this PR):
Coordinates

Note that before this PR:
- widget coordinate meant Layout pixel coordinates
- what is now widget coordinates was sometimes called "absolute coordinate" (input system)

rolandlo:
> I got a crash (reproducibly) after creating a new document after annotating a large one (discarding the chances):
> 
> ```
> Try to get a stacktrace...
> Stack trace (most recent call first):
> #0  0x0000634d08a51558 in Stacktrace::printStacktrace(std::ostream&) at inst/bin/xournalpp
> #1  0x0000634d086d3e43 in crashHandler at inst/bin/xournalpp
> #2  0x00007bdeab845f5f at /lib/x86_64-linux-gnu/libc.so.6
> #3  0x00007bdeab8a5fed in pthread_kill at /lib/x86_64-linux-gnu/libc.so.6
> #4  0x00007bdeab845e2d in gsignal at /lib/x86_64-linux-gnu/libc.so.6
> #5  0x00007bdeab828887 in abort at /lib/x86_64-linux-gnu/libc.so.6
> #6  0x00007bdeabcad3d7 in std::__glibcxx_assert_fail(char const*, int, char const*, char const*) at /lib/x86_64-linux-gnu/libstdc++.so.6
> #7  0x0000634d0876a285 in std::vector<std::unique_ptr<XojPageView, std::default_delete<XojPageView> >, std::allocator<std::unique_ptr<XojPageView, std::default_delete<XojPageView> > > >::operator[](unsigned long) const at inst/bin/xournalpp
> #8  0x0000634d088c8274 in Layout::updateVisibility() at inst/bin/xournalpp
> #9  0x0000634d088c7120 in Layout::verticalScrollChanged(_GtkAdjustment*, Layout*) at inst/bin/xournalpp
> #10 0x00007bdeac92d5db in g_closure_invoke at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
> #11 0x00007bdeac94be08 at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
> #12 0x00007bdeac94dda5 at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
> #13 0x00007bdeac94e007 in g_signal_emit_valist at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
> #14 0x00007bdeac94e0c2 in g_signal_emit at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
> #15 0x00007bdeaccba0f8 in gtk_adjustment_clamp_page at /lib/x86_64-linux-gnu/libgtk-3.so.0
> #16 0x0000634d088c91e6 in Layout::ensureRectIsVisible(int, int, int, int) at inst/bin/xournalpp
> #17 0x0000634d08768435 in XournalView::scrollTo(unsigned long, XojPdfRectangle) at inst/bin/xournalpp
> #18 0x0000634d08769b89 in XournalView::documentChanged(DocumentChangeType) at inst/bin/xournalpp
> #19 0x0000634d087daa34 in DocumentHandler::fireDocumentChanged(DocumentChangeType) at inst/bin/xournalpp
> #20 0x0000634d087d0ebe in Document::operator=(Document const&) at inst/bin/xournalpp
> #21 0x0000634d088234dc in Control::replaceDocument(std::unique_ptr<Document, std::default_delete<Document> >, int) at inst/bin/xournalpp
> #22 0x0000634d08822c9e at inst/bin/xournalpp
> #23 0x0000634d08832a57 at inst/bin/xournalpp
> #24 0x0000634d0883072e at inst/bin/xournalpp
> #25 0x0000634d0882eaa3 at inst/bin/xournalpp
> #26 0x0000634d08839a15 in std::function<void (bool)>::operator()(bool) const at inst/bin/xournalpp
> #27 0x0000634d08829b4e in Control::close(std::function<void (bool)>, bool, bool, bool) at inst/bin/xournalpp
> #28 0x0000634d08822df7 in Control::newFile(std::filesystem::__cxx11::path) at inst/bin/xournalpp
> #29 0x0000634d0886c62c in ActionProperties<(Action)0>::callback(_GSimpleAction*, _GVariant*, Control*) at inst/bin/xournalpp
> ```

The easiest way to reproduce:
1) Start Xournal++ from this PR
2) Add a 2nd page (Ctrl+D)
3) Create a new document (Ctrl+N), discard changes

bhennion:
Actually, it is a problem with current master already... Surely due to the Layout refactor

Edit: See #7182 for a fix

rolandlo (approved):
LGTM. I have tested on Windows and MacOS ARM as well to be sure not to get any nasty surprises and haven't found issues with this PR.
There is an issue with file filters on Windows, but unrelated to this PR.

Great that #5033 gets fixed.

bhennion:
> There is an issue with file filters on Windows, but unrelated to this PR.

Would that be #6748?



> LGTM. I have tested on Windows and MacOS ARM as well to be sure not to get any nasty surprises and haven't found issues with this PR.

Thanks a lot for the thorough testing!

rolandlo:
> > There is an issue with file filters on Windows, but unrelated to this PR.
> 
> Would that be #6748?

Not exactly. All files show up in some folders (even .exe and the like). And there are some Gtk-Criticals connected to that.

Issue #7108, opened 2026-02-09
fs::path apparently it fails on Windows ARM (encoding issues with UTF8 conversion)


tiziodcaio:
> Well, apparently it fails on Windows ARM. It looks like some encoding issue (ie UTF8 conversion)
```
C:/a/xournalpp/xournalpp/test/unit_tests/control/SettingsTest.cpp:63: Failure
Expected equality of these values:
  settings.getColorPaletteSetting()
    Which is: "foo/bar€_palette"
  loaded.getColorPaletteSetting()
    Which is: "foo/barâ‚¬_palette"
```

_Originally posted by @bhennion in https://github.com/xournalpp/xournalpp/issues/7100#issuecomment-3873365350_

tiziodcaio:
Context: https://github.com/xournalpp/xournalpp/actions/runs/21802044167/job/63004298647?pr=7100

PR #7100, opened 2026-02-08
Add crude Settings save/load test

tiziodcaio:
Add crude testing to Xournal++ (ref #6623)

PR #7113, opened 2026-02-11
Add utf8-related test cases and fixe the newly found issues

bhennion:


bhennion:
I used a modified version of libstdc++'s filesystem header to remove any possibility of converting a std::string(_view) into a fs::path. This allowed me to track down remaining encoding related issues.
This PR adds test units catching those issues, and then fixes the issues. It also includes #7100 and fixes #7108.

My investigation showed another remaining encoding issue. I'll open a ticket for it later.

bhennion:
Thanks for the review!
I rebased. Merging once the CI clears

PR #7116, opened 2026-02-13
Fix xopp-WARNING "Fails to extract theme root name from: ..."

hundertzwei:
It is totaly legal to have numbers, underscore and even a dot in the name of a gtk-theme. F.E. Windows-8.1 or Unity-8 from the B00MERANG Project


bhennion:
Merging once the CI test passes

PR #7122, opened 2026-02-14
Update xournalpp.spec

luyatshimbalanga:
This commit disable cpptrace support as Fedora building system disallows external online access while enabling qpdf support.
The spec file receive some cleanup removing unneeded wrapper and manual configuration of .desktop file.

### DIFF
--- a/rpm/fedora/xournalpp.spec
+++ b/rpm/fedora/xournalpp.spec
@@ -1,6 +1,9 @@
 # Force out of source build
 %global __cmake_in_source_build 0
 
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
Thanks for the update. LGTM

mjg:
This appears to be taken from my fork or copr, but incompletely. It will not work without a proper buildrequire on libqpdf.

luyatshimbalanga:
Good catch. I prepare a new PR to address the issue.

PR #7125, opened 2026-02-14
Add custom variable support to LaTeX generator via `%xpp:key=value`

### LINKED ISSUE #7157: Support user-defined variables in LaTeX template via %xpp: directives

The LaTeX template system only supports two hardcoded substitution variables (`%%XPP_TOOL_INPUT%%`, and `%%XPP_TEXT_COLOR%%`), and passing anything to the templates is kind of awkward, we have no way to do so without duplicating content or editing the templates, and I have to keep selecting different templates with just one or 2 line differences instead of using `\ifthenelse` for example

### Feature Description

Extend the LaTeX template substitution to support user-defined variables declared inline within the LaTeX input (comments so it could not affect LaTeX itself), using a directive syntax at the top of the input field

```latex
%xpp:my_var=some value
```

Lines beginning with `%xpp:<name>=<value>` are parsed out before processing, available as `%%XPP_<NAME>%%` placeholders in the template

Variable names are case-insensitive and normalized to UPPERCASE. The reserved names `TOOL_INPUT` and `TEXT_COLOR` cannot be overridden

### Alternative Solutions

_No response_

### Additional Context

I have opened a pr to add this feature #7125

I did not add a docs, because I don't where to add it

youssefadly237:
Adds a way to define custom variables in LaTeX input that can be used in templates

Example:

```latex
\documentclass[varwidth=0.999\maxdimen, crop, border=5pt]{standalone}
\newcommand*{\setTextWidthReference}{%
  \setlength{\textwidth}{345.0pt}%
  \setlength{\linewidth}{\textwidth}%
  \setlength{\columnwidth}{\textwidth}%
}
% Packages
\usepackage{amsmath}
\usepackage{amssymb}
\usepackage{scontents}
\usepackage{ifthen}
\newlength{\pheight}
% Color support
\usepackage{xcolor}
\definecolor{xpp_font_color}{HTML}{%%XPP_TEXT_COLOR%%}

% Custom variable example:
% Add %xpp:mode=inline at the top of your input
% to make the formula render in inline mode
% it is display mode by default
\newcommand{\xppmode}{%%XPP_MODE%%}

% User input
\begin{scontents}[store-env=preview]
  \ifthenelse{\equal{\xppmode}{inline}}{%
    \( %%XPP_TOOL_INPUT%% \)
  }{%
    \( \displaystyle %%XPP_TOOL_INPUT%% \)
  }
\end{scontents}
\begin{document}
  \setTextWidthReference
  \textcolor{xpp_font_color}{\getstored[1]{preview}}
\end{document}
```

```latex
%xpp:mode=inline
\left[ x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a} \right]
```

`%%XPP_MODE%%` will be substituted in the template with `inline`

just like `%%XPP_TOOL_INPUT%%` and `%%XPP_TEXT_COLOR%%` was getting substituted

<p align="center">
  
  
</p>

btw writing `%xpp:mode=display` has no effect at all, the example is just (if inline, else) not else if, but you can custom it however you want

---

other useful tip

you can also add it in the beginning of your default prompt so you do not have to type it every time (Edit -> preferences -> LaTeX -> Default LaTeX text)

image

closes #7157

### DIFF
--- a/src/core/control/latex/LatexGenerator.cpp
+++ b/src/core/control/latex/LatexGenerator.cpp
@@ -1,5 +1,7 @@
 #include "LatexGenerator.h"
 
+#include <algorithm>    // for transform
+#include <map>          // for map
 #include <regex>        // for smatch, sregex_iterator
 #include <sstream>      // for ostringstream
 #include <string_view>  // for string_view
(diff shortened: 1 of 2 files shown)


rolandlo:
Please remove the changes to xournalpp.pot. The translation template update happens automatically after the PR is merged.

rolandlo:
Thanks again @youssefadly237 for the interesting feature.

PR #7139, opened 2026-02-16
Fix horizontal page alignement of unequal page sizes

### LINKED ISSUE #5268: Page layout is unaligned and pages are not centered vertically

tmoerschell:
Correct the right-side padding for pages that narrower than their layout column so that all pages align consistently in a grid.

Fixes #5268

### DIFF
--- a/src/core/gui/Layout.cpp
+++ b/src/core/gui/Layout.cpp
@@ -289,7 +289,7 @@ void Layout::layoutPages(int width, int height) {
                         }
                     } else {  // not paired page mode - center
                         paddingLeft = XOURNAL_PADDING_BETWEEN / 2.0 + columnPadding / 2.0;  // center justify
-                        paddingRight = XOURNAL_PADDING_BETWEEN - paddingLeft + columnPadding / 2.0;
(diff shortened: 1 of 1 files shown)


bhennion:
Thanks a lot!

PR #7163, opened 2026-02-19
[skip ci] add link to xournal and its author

archisman-panigrahi:
I thought it would be a good idea to add the link to the author of Xournal, who apparently created it single-handedly.
I locally tested that the links work.

### DIFF
--- a/ui/about.glade
+++ b/ui/about.glade
@@ -183,8 +183,8 @@ Luya Tshimbalanga</property>
                     <property name="name">lbXournal</property>
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
-                    <property name="label" translatable="yes">Partially based on Xournal
-by Denis Auroux</property>
+                    <property name="label" translatable="yes">Partially based on &lt;a href="https://xournal.sourceforge.net/"&gt;Xournal&lt;/a&gt;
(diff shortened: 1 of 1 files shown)


archisman-panigrahi:
For some reason CI started despite [skip ci]

bhennion:
I do not think it our place to put someone else's information (albeit public) on our repository. If the original author explicitely asked to, then maybe.

bhennion:
> For some reason CI started despite [skip ci]

The [skip ci] flag is something to add to a commit, not to a PR. It should however be used with parcimony as it can has unwanted consequences. For instance, if the day's last commit has [skip ci], then we would not get a nightly build. Another example: translatable strings add in a [skip ci] commit will not be uploaded to Crowdin until another commit is added.

All in all, this flag should really not be advertised in a PR template as it is. It promotes the use of something that has subtle consequences. At last, 99% of PRs will touch components that will require testing. This PR template adds noise to the common use case without real benefits.

archisman-panigrahi:
> I do not think it our place to put someone else's information (albeit public) on our repository. If the original author explicitely asked to, then maybe.

Sure. Closing, then.

> All in all, this flag should really not be advertised in a PR template as it is. It promotes the use of something that has subtle consequences. At last, 99% of PRs will touch components that will require testing. This PR template adds noise to the common use case without real benefits.

I created #7166 to stop advertising it in PR template.

PR #7167, opened 2026-02-19
add option to save last window position (fix #4488)

### LINKED ISSUE #4488: Is it possible for Xournal++ to remember the last window position coordinates?
It is frustrating to have to manually reposition a program window over and over each time an application is launched. Most modern applications remember where the user positioned them the last time and this would be very helpful.

archisman-panigrahi:
This will fix #4488 

image

I added a remember window position toggle switch that will use GTK3's inbuilt features to get the window position just before closing and save it in the config file. Upon loading, it will look for the last window position and window size in the config file (only if "remember window position" is checked) and apply them.

By default the feature is not turned on, but I can turn it on if you guys think it's helpful.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -279,6 +279,14 @@ void Control::saveSettings() {
 
     if (!this->win->isMaximized()) {
         this->settings->setMainWndSize(width, height);
+
+        // Save window position if remember option is enabled
+        if (this->settings->isRememberWindowPosition()) {
+            gint x = 0;
+            gint y = 0;
+            gtk_window_get_position(getGtkWindow(), &x, &y);
+            this->settings->setMainWndPosition(x, y);
+        }
     }
     this->settings->setMainWndMaximized(this->win->isMaximized());
 
--- a/src/core/control/settings/Settings.cpp
+++ b/src/core/control/settings/Settings.cpp
@@ -89,6 +89,9 @@ void Settings::loadDefault() {
 
     this->mainWndWidth = 800;
     this->mainWndHeight = 600;
+    this->mainWndX = -1;  // -1 means use default/system positioning
+    this->mainWndY = -1;
+    this->rememberWindowPosition = false;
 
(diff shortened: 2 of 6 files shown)


archisman-panigrahi:
I compiled and tested it in EndeavourOS+KDE

archisman-panigrahi:
> By default the feature is not turned on, but I can turn it on if you guys think it's helpful.

In my opinion it would not harm in anyway if this option is enabled by default.

Otherwise, ready from my side.

bhennion:
I'm not sure about this feature, because the API it must rely on has been removed in GTK4.
Also, most GDK backends don't actually support this: on my setup, it works for x11 but not for Wayland.

I think the rationale behind this API being removed is that the DE should be the one deciding what goes where: only the DE knows about multiple monitors, their relative positionning, and so on.
For KDE users (like yourself), there are tweaks out there to force new windows to be placed in the same location (e.g. https://store.kde.org/p/2324743/).

archisman-panigrahi:
> I'm not sure about this feature, because the API it must rely on has been removed in GTK4.

Have you all decided that you are definitely going to switch to GTK4? 
From my experience, GTK4 removed a lot of useful APIs (also useful features like app menus) and anyway GTK3 is not going to go away if powerful apps like GIMP or xournalpp keep using it.

> Also, most GDK backends don't actually support this: on my setup, it works for x11 but not for Wayland.

I only tested in X11. I'll see if this can be done in some way using Wayland.

archisman-panigrahi:
I can rename the setting to "Remember window position (for X11)". 

Anyway, Wayland removed the feature and we cannot do much (as it is upto DEs/compositors to implement it). But a large number of users are going to stick to X11 for various reasons for years to come and they want this feature.
Having this feature is going to help them.

Please let me know what you think.

rolandlo:
Yes, we'll definitely switch to Gtk 4 at some point, even though it's not the most urgent thing. I'm not happy with adding an option just for X11. So I'm closing this PR for now.

PR #7176, opened 2026-02-20
Add "Tablet mode" toolbar configuration

archisman-panigrahi:
There is a complaint among xournalpp user that the user interface looks dated, mostly because there are too many menus that consume too much space.
Nowadays a lot of people take handwritten notes in Android/iOS tablet apps where the writing pad consumes maxium area, and toolbars consume the minimum area (for example see pictures in https://techwiser.com/best-handwriting-apps-for-android/).

Here I tried to maximize the writing area, while minimizing the number of toolbars items, yet retaining nearly all necessary toolbar items. This is based on right hand note taking, but with a lot of changes to make it compatible with tablet usage.

I removed a lot of redundancies (e.g. only "save" among "save" and "save as" is kept, used multiple select option instead of all the different kinds of select options). Also, features such as "print" are rarely used while taking notes in a tablet -- I removed them. To save space, I also removed "Open" as files can be opened from file manager, and also, one can always open files from the main menu.

The next/previous page buttons are not shown, because one can go to the next/previous page using the "page number" tool. One can show/hide the sidebar with the toggle button and the toggle fullscreen button is retained as well.

This is fully compatible with **both portrait and landscape** screen orientations.
The "page number" and "toggle fullscreen" and "toggle sidebar" buttons are duplicated in the two panels so that they always show up in both portrait and landscape modes. In portrait modes, they get cut off from the horizontal panel but show up in the vertical panel. In horizontal mode they show up in vertical panel but get cut off from horizontal panel.

The toggle touch drawing button is available so that one can use it in both tablets with and without a dedicated stylus.

Tested in both 1366x768 and 1536x2048 screen resolutions.

----

Screenshots:

photo_2026-02-20_10-14-30
(Landscape mode)

photo_2026-02-20_10-14-48
(Portrait mode)

photo_2026-02-20_10-14-42

Notice how the page number and toggle fullscreen and toggle sidebar buttons are available in both the portrait and landscape modes.


rolandlo (approved):
Thanks for this clever toolbar. LGTM

hwilliams-NE:
Just a query, as a left hander I'd be interested in using this with the toolbar on the right hand side of the screen. Is that possible?

archisman-panigrahi:
I can send you a config file, loading which will give you this new configuration. I am slightly against the idea of adding another toolbar to the set of available toolbars, as there are already too many.

hwilliams-NE:
I hadn't realised it was just updating a configuration file, sorry. Can it be loaded at runtime?

archisman-panigrahi:
Unzip this file, and save `toolbar.ini` in `$HOME/.config/xournalpp/` and restart Xournal++. Then choose "Tablet mode (Left Hand)" configuration in `Edit -> Toolbars`.
[toolbar.ini.zip](https://github.com/user-attachments/files/25771254/toolbar.ini.zip)

By the way, I am curious to know: If you are right-handed, does it actually help in any way if the toolbar is on the left, and if you are left-handed, then the toolbar is on the right?

hwilliams-NE:
> By the way, I am curious to know: If you are right-handed, does it actually help in any way if the toolbar is on the left, and if you are left-handed, then the toolbar is on the right?

I guess I'll find out by trying both!

archisman-panigrahi:
@hwilliams-NE Please let us know about your experience, and we would consider adding a separate toolbar for Tablet mode for left handed users based on your feedback

archisman-panigrahi:
@rolandlo  Can we bring this to v1.3.5, because the change is quite trivial and it is not going to break anything?

PR #7186, opened 2026-02-24
xoj-preview/extractor/desktop-install.sh: minor shebang improvements

fkobi:


bhennion:
Thanks for the cleanup!

Issue #1783, opened 2020-03-19
Pinch to zoom is laggy


coldspark29:
**Affects versions :**
 - OS: Manjaro
 - (Linux only) Desktop environment: X11
 - Which version of libgtk do you use
    gtk3 1:3.24.14+9+g429a6287e0-1 
    gtk2 2.24.32-2
 - Version of Xournal++: 1.1.0+dev 46a4dc82
 - Installation method: AUR

**Describe the bug**
Zooming is generally laggy

Technius:
This is likely caused caused by the canvas widget having to resize on document changes. I'll see I can fix it in the upcoming viewport refactoring.

coldspark29:
Will you maybe also be able to fix the weird shifting to the side when zooming in?

Technius:
I'll look into it; the goal is to overhaul the current zoom/layout spaghetti code so that it's easier to fix these kinds of issues.

chron-isch:
This should be fixed by #2435.

Technius:
Has this lagginess been fixed?

coldspark29:
Yes, it has become smoother. There is still room for improvement though. I think it is less of a lag, but more a sluggish rescaling

coldspark29:
So #2718 is fixed and I thinkt he zooming is still laggy.

ohthehugemanatee:
Still laggy when pinch-zooming out quickly on 1.1.0+dev, commit 8f83a043, libgtk 3.24.26 .

To reproduce:
1) create a document
2) pinch to zoom in
3) unpinch to zoom out, quickly

- Doesn't seem affected by drawigs on the page
- It is much worse on a PDF. 
- It ends at the right place
- You can continue the gesture after the lag
- Nothing is output in stdout/stderr
- it is as if it just dropped a lot of video frames in the zoom.

Guessing that this is because we re-render at a scale-appropriate resolution at each step of the zoom. If that's the case, we could do it like OSX Preview does it, instead: resize a the document at the current resolution, and re-render as an asynchronous step.

maribox:
Still have the same problem as @ohthehugemanatee in 1.2.2+dev, commit [c5ddca9](https://github.com/xournalpp/xournalpp/commit/c5ddca935d23883288251a970ee1a16c27e02a5f)

bhennion:
See #7081 for progress in this direction

michelangelo4:
[...]

PR #7199, opened 2026-02-27
Remove metadata updates on every zoom/scroll event

bhennion:
This PR addresses the very basic cause of #1783 without touching the rendering methods.
The problem is that at every or scroll operation, the document mutex was locked, thus blocking the UI if a RenderJob was rendering in a separate thread.

This PR:

1. Postpones the update of the metadata to right before writing the metadata down to the disk (instead of updating it all the time), thus removing one source of the problem entirely
2. Only calls DocumentHandler::firePageSelected if the page has actually changed. This saves another doc->lock() in PageBackgroundChangeController::pageSelected() (calling Control::getCurrentPage() which locks the mutex).

The scroll/zoom feels way more fluid. 
You can try it out on this demanding file for instance
[render-benchmark-strokes.xopp.zip](https://github.com/user-attachments/files/25614687/render-benchmark-strokes.xopp.zip)
(rename the .zip away).

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -559,7 +559,11 @@ auto Control::firePageSelected(const PageRef& page) -> size_t {
     return pageId;
 }
 
-void Control::firePageSelected(size_t page) { DocumentHandler::firePageSelected(page); }
+void Control::firePageSelected(size_t page) {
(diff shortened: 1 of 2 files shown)


bhennion:
Thanks for the review and testing!

PR #7221, opened 2026-03-03
Fix palette directories not showing up in file explorer on Windows

rolandlo:
Two issues are fixed:

1) On Windows a valid file URI is not `file://<path>`, but `file:///<path>`. It doesn't matter if backslashes or slashes are being used in the path.

2) Link activation didn't work on Windows even with correct URI. That's why `XojMsgBox::openURL` is used.

### DIFF
--- a/cmake/include/GitRepo.cmake
+++ b/cmake/include/GitRepo.cmake
@@ -3,12 +3,10 @@
 #
 # Variables:
 #   GIT_ORIGIN_URL    origin URL of current git repo
-#   GIT_ORIGIN_OWNER  origin repo owner 
+#   GIT_ORIGIN_OWNER  origin repo owner
 #                     (for local clones it is the name of the directory from which xournalpp was cloned)
 #   GIT_ORIGIN_REPO   origin repo name
 #   GIT_BRANCH        current git branch
-#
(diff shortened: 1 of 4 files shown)


cmake/include/GitRepo.cmake
```diff
@@ -48,7 +46,3 @@ if (EXISTS "${PROJECT_SOURCE_DIR}/.git" AND PATH_GIT)
   endif ()
 endif ()
 
-# Hardcode bugreport address instead of using ${GIT_ORIGIN_OWNER} which is just a directory name for local clones
-set (PROJECT_BUGREPORT "https://github.com/xournalpp/xournalpp/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml")
-set (PROJECT_CRASHREPORT "https://github.com/xournalpp/xournalpp/issues/new?template=crash_report.yml")
```
bhennion, line 53: PROJECT_CRASHREPORT is still used in the wrapper.

rolandlo: Oh, I thought I had put it back. Missed this line. Thanks for the review!

rolandlo:
Merging once the pipeline succeeds.

PR #7225, opened 2026-03-05
Group plugin menu items into named submenus

youssefadly237:
This change updates the plugin menu system to create one sub menu per enabled plugin instead of appending all plugins actions into plugin menu directly

why? because a plugin like vi-xournalpp populates the plugin menu and makes it hard to find other plugins actions (it was made to be accessed via keyboard not the ui, which makes it more awkward)


bhennion:
The coed looks fine to me. @rolandlo are you ok with the change in behaviour?

rolandlo:
> The coed looks fine to me. @rolandlo are you ok with the change in behaviour?

Generally speaking this looks like a good idea. The alternative would be to allow plugin menu entires outside of a plugin-specific submenu (as currently) and those in a plugin-specific submenu simultaneously, in order to save some clicks. But I feel that the suggested solution is cleaner. Plugin actions are often activated by shortcuts (or toolbar buttons) anyway.

PR #7230, opened 2026-03-05
Use a shared_mutex to protect the document instance.

### LINKED ISSUE #1783: Pinch to zoom is laggy

### LINKED ISSUE #2890: Performance issues with scanned PDFs
**Affects versions :**
 - OS: Manjaro Linux
 - (Linux only) Desktop environment: X11 (Xfce)
 - Which version of libgtk do you use: 3.24.24
 - Version of Xournal++: 1.0.20-hotfix and 1.1.0+dev (slightly different behaviour in the two versions, see below)
 - Installation method: Repository package and AUR package (built from latest git)

**Describe the bug**
When opening some scanned PDF files with Xournal++, there are severe performance issues. It doesn't happen to all files but if it happens to a file, then it happens reproducibly. Strangely, some files have a very similar file size, page size and number of pages but one of them shows the issue and another one doesn't. The problematic files I am currently dealing with are around 3,5MB file size, 3 pages with page size 792 x 1056 pts.

My hardware specs are rather low which might be one of the roots of the problem: Thinkpad X200T (Intel Core2 Duo L9400, 8GB of RAM; SSD).

Version 1.0.20:

Xournal++ takes some seconds to initially load the file (not significantly longer than any PDF viewer). The first page is loaded, the other two pages not yet. When I try to annotate the file, after 1-2 strokes the CPU goes to 100%, Xournal++ becomes unresponsive and new strokes aren't shown. They are shown later if I give Xournal++ some time to "cool down". After a "cool down" I can again write a few strokes till the CPU goes to 100% and further strokes aren't shown.

However, if I scroll down such that all three pages are loaded, the performance issue doesn't appear, I can annotate the file (almost) without performance problems.

Version 1.1.0: 

Same problem as in version 1.0.20 (subjectively, a bit more severe, that is, longer loading times) but now it does *not* help to scroll down and load all pages. Even if all pages are loaded, the CPU goes to 100% after every 1-2 strokes.

**To Reproduce**
1. Open a scanned PDF file of around 3,5MB file size
2. Use the pen tool to annotate it

**Expected behavior**
There should be no freezes.

### LINKED ISSUE #7224: Very slow while navigating in PDF
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Official Release

### Xournal++ Version

1.3.3

### libgtk Version

3.24.51

### Bug Description

It's really frustrating that I can't navigate the PDF normally without the program frame-lags, making notes on the PDF is torture

### Expected Behaviour

When I scroll through pages it shouldn't freeze or frame-lag

### Steps to Reproduce

1. Open this slides in Xournal++ and try to scroll between pages or jum to specific page

[Slides_integrali-1-100.pdf](https://github.com/user-attachments/files/25737140/Slides_integrali-1-100.pdf)

### Additional Context

These slides are an excerpt from a 550 page slide show, I can't upload them but if you're interested I can find an alternative way to upload them

bhennion:
This PR improves the app's responsiveness by allowing several concurrent read-only accesses to the Document.

In practice, we use a shared_mutex and a shared lock for every read-only operations. Write operations still use the standard lock of the mutex.
It makes navigation way more fluid when opening heavy documents (see e.g. the document of #7224).

Note that for such documents, the rendering may still lag behind, but the user interface doesn't block anymore.

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -251,7 +251,7 @@ void Control::deleteLastAutosaveFile() {
 }
 
 auto Control::checkChangedDocument(Control* control) -> bool {
-    if (!control->doc->tryLock()) {
+    if (!control->doc->try_lock_shared()) {
         // call again later
         return true;
     }
@@ -264,7 +264,7 @@ auto Control::checkChangedDocument(Control* control) -> bool {
         }
     }
     control->changedPages.clear();
-    control->doc->unlock();
+    control->doc->unlock_shared();
 
     // Call again
     return true;
@@ -548,9 +548,9 @@ void Control::reorderSelection(EditSelection::OrderChange change) {
  * @return the page ID or size_t_npos if the page is not found
  */
 auto Control::firePageSelected(const PageRef& page) -> size_t {
-    this->doc->lock();
+    this->doc->lock_shared();
     size_t pageId = this->doc->indexOf(page);
-    this->doc->unlock();
+    this->doc->unlock_shared();
     if (pageId == npos) {
         return npos;
     }
@@ -703,9 +703,9 @@ void Control::deletePage() {
     // if the current page contains the geometry tool, reset it
     size_t pNr = getCurrentPageNo();
     if (geometryToolController) {
-        doc->lock();
+        doc->lock_shared();
         auto page = doc->indexOf(geometryToolController->getPage());
-        doc->unlock();
+        doc->unlock_shared();
         if (page == pNr) {
             resetGeometryTool();
         }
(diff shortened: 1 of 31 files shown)


rolandlo (commented):
Codewise LGTM. Also the scrolling improvements with heavy documents is staggering. I will test it out a bit more to make sure there's no issue with it.

rolandlo (approved):
It all seems to work fine.

rolandlo:
Just one question to add: Are you sure to target this at the master branch and not the release-1.3 branch?

bhennion:
> Just one question to add: Are you sure to target this at the master branch and not the release-1.3 branch?

I was hesitant. I rebased onto the release branch.
Some care will be needed when merging the release branch onto the master branch: the NavigationHistory business will need to use shared locks as well.

Merging once the CI clears

PR #7242, opened 2026-03-07
Fix shebang line for python script

GeoMinesGold:
As title says, current shebang line will not work as there is a space. Also use env instead of calling /bin/python3 directly to (1) maintain compatibility with non-FHS compliant operating systems (e.g., NixOS) and (2) use the python binary in the user's path.


rolandlo:
Thanks for the fix. Can you please rebase onto the release-1.3 branch?

PR #7248, opened 2026-03-09
Show PDF page labels in toolbar page spinner

youssefadly237:
## Problem

when using PDFs with non-sequential page labels (roman numerals, named front matter, ...) page spinner show pages indexed starting from 1, in order to navigate to a page you had to know the page shift (the diff between actual page order and its index order), then add it to the page number

## Changes

PDFs with non-sequential page labels now show those labels in the toolbar page spinner instead of the raw page index

At document load, `buildAndPushPageLabels` walks the xournal pages, reads their PDF page labels, and filters out any label that matches its numeric index. If any non-standard labels are found the full label vector is pushed to `SpinPageAdapter` via a `setPageLabels` path

The spinner entry then displays the PDF label and accepts label input for navigation, with a fallback to numeric page index for unlabelled pages or plain number input

As a cleanup, the `pdfPage` parameter is dropped from the `setPageInfo` call stack it was only used to show "PDF Page N" in the label

<sub>sorry for strange commit names I will clean them up on squash</sub>

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -741,6 +741,7 @@ void Control::deletePage() {
     this->doc->unlock();
 
     this->undoRedo->addUndoAction(std::make_unique<InsertDeletePageUndoAction>(page, pNr, false));
+    const size_t deletedPos = pNr;
 
     if (pNr >= this->doc->getPageCount()) {
         pNr = this->doc->getPageCount() - 1;
@@ -748,6 +749,7 @@ void Control::deletePage() {
 
     scrollHandler->scrollToPage(pNr);
     this->win->getXournal()->forceUpdatePagenumbers();
+    notifyPageDeleted(deletedPos);
 }
 
 void Control::duplicatePage() {
@@ -792,6 +794,7 @@ void Control::movePageTowardsBeginning() {
     this->firePageSelected(currentPageNo - 1);
 
     this->getScrollHandler()->scrollToPage(currentPageNo - 1);
+    notifyPagesSwapped(currentPageNo - 1, currentPageNo);
 }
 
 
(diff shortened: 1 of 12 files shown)


bhennion:
I'm not so sure about this. The feature in itself could be useful, but I feel the cost (at least of this implementation) is too expensive for long PDF documents (because of the bidirectional map label<->page number).
It also forces to preload all the PDF pages (for now, the .xopp files are NOT loaded lazily, but it is something we want to tend towards).
It also doesn't work in the following situations:

1. If one adds/deletes a page
2. If one adds a page and duplicates the PDF background

Those cases don't work in the current implementation, but they also raise deeper problems: labels are not necessarily unique, or a label can be a number and this number could also correspond to a label-less page.

Implementing an almost cost-free "display the page label" feature is easily doable, I think. However, the "label to page number" conversion comes with a significant cost, and encounters the above concerns.

youssefadly237:
@bhennion 

That's why I didn't squash, the first commit ac0280b61 is the lightweight display-only approach you're describing I think, it only fetches the label for the current page on each navigation (no preloading, no bidirectional map), and simply appends it to the `of N` label in the spinner

The second commit is where the full label-to-page-number navigation was added, introducing the `buildAndPushPageLabels()`

can you review the first commit only?

youssefadly237:
btw appending or deleting pages, makes navigation to specific page based on index hard, since the shift is not constant value across the whole pdf

I tried to

- remove the label to page map approach
- keep label input navigation, but made duplicate labels deterministic with next match and wraparound
- replaced full label rebuilds on page insert, delete, move, undo, and redo with incremental updates

there is still one label build pass at startup

can we add to the settings in navigation make it index based or label based? with a little note about the performance diff?

> Implementing an almost cost-free "display the page label" feature is easily doable

I don't think displaying page label is useful, since most pages already has page number in footer or header, so showing it in the spinner adds nothing I think, my point was to assist navigation

PR #7257, opened 2026-03-11
Fixes for shebang lines

GeoMinesGold:
As title says, these are changes to shebang lines to be more portable by using env instead of calling the binaries (e.g., python3) directly to (1) maintain compatibility with non-FHS compliant operating systems (e.g., NixOS) and (2) use the preferred binaries in the user's path. 

This is a new, more exhaustive PR (and rebased on release-1.3) instead of the earlier one I made.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,4 +1,4 @@
-#!/usr/bin/make -f
+#!/usr/bin/env -S make -f
 
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all
 
--- a/scripts/lua_def_file.py
+++ b/scripts/lua_def_file.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import re
 import sys
--- a/scripts/color-names
+++ b/scripts/color-names
@@ -1,4 +1,4 @@
-#! /bin/python3
+#!/usr/bin/env python3
 
 from functools import partial
 from math import sqrt
--- a/src/xoj-preview-extractor/desktop-install.sh
+++ b/src/xoj-preview-extractor/desktop-install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 export GCONF_CONFIG_SOURCE=`gconftool-2 --get-default-source`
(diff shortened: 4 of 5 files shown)


scripts/lua_def_file.py
GeoMinesGold: Explicitly state python3 instead of just python as python3 is not backwards-compatible with python2

debian/rules
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/make -f
+#!/usr/bin/env -S make -f
```
rolandlo, line 1: Are you sure of this change? In the [docs](https://www.debian.org/doc/debian-policy/ch-source.html#main-building-script-debian-rules) it's explicitly written that 
>debian/rules must start with the line #!/usr/bin/make -f, 

From the reasoning it seems both ways should be fine though (on a Debian system).
> so that it can be invoked by saying its name rather than invoking make explicitly. That is, invoking either of make -f debian/rules args... or ./debian/rules args... must result in identical behavior.

rolandlo (commented):
Thanks for the fixes. While you are at it I have some more suggestions:

- Also add a shebang to compare_license.py 
- Rename `color-names` to `color-names.py`
- Make sure those scripts are all executable

debian/rules
hyperupcall: Sorry for the drive-by comment, but it's also important to note that the `-S` flag of `env` is not portable.

GeoMinesGold: Debian indeed explicitly states a direct shebang, but I tried it with `env` (with `-S`, otherwise it will fail) in a Debian 13 VM, Fedora 43 VM, NixOS 25.11 VM, and Arch bare metal and it worked flawlessly.
However, @hyperupcall is correct about the `-S` flag not being portable (as in, it being a [somewhat new feature](https://github.com/coreutils/coreutils/commit/668306ed86c8c79b0af0db8b9c882654ebb66db2)), so I'm quite unsure on what is the optimal way to do this. An unclean solution may be a shell wrapper, or we could just revert it to what it was.

rolandlo: In case of doubt revert to what it was.

PR #7265, opened 2026-03-14
Fix undefined behaviour and dangling pointer

bhennion:
Each commit fixes a separate issue. See the commit descriptions.


src/core/plugin/luapi_application.h
bhennion, line 1: I removed the whole `if` statement: it was unsafe, dereferencing a dangling pointer (`filename` points to a std::string data, but the std::string has been deleted by the time `filename` is dereferenced).
Also, I do not see what use it could have had. It seems to have be supposed to replace `foo.xopp` by `./foo.xopp`. Those give equivalent paths.

rolandlo: Are you sure the replacement of `foo.bar` by `./foo.bar` is unnecessary? The former gives an empty `parent_path`, which is then used in `gtk_file_chooser_set_current_folder`.

bhennion: You're right! I corrected that.

bhennion:
Thanks for the catch and review!

PR #6594, opened 2025-08-11
Use tiles for rendering pages

bhennion:
This PR implements a tiling system, so that pages' buffers are no longer necessarily a single cairo_surface: they can now be a bunch of smaller surfaces (tiles).

The goal is to manage our RAM usage (in particular when zoomed in).

There are parameters to be tuned and about which I'd like to have other devs opinions:

1. The typical size of the tiles, in pixels or in userspace pixels (the difference being the DPI scaling ratio)
2. How to decide which tiles should be rendered and which should be discarded?

For now, the algorithm does this (for debug and proof-of-concept purposes, I put many tiny tiles to make this video - in real life, tiles must be much bigger and the number of tiles should be kept much smaller):

[Enregistrement d'écran_20250811_121710.webm](https://github.com/user-attachments/assets/7e7211b4-d12e-4cc1-9b75-1a55c2614fb5)

Each colored square is a different tile.

I made some tests with tiles 2000x2000 pixels on a 10 pages long document (pages were 21x19 cm for some reason...), with zoom 700%, I get 800MB (vs 2.2GB for master) in RSS ram usage (and about 140MB for both with zoom 80%). so it is a big gain.

With this PR, you can also create a page with size 23mx45m (or any huge values) and work on it as a virtually infinite canvas. However, because the Page elements are not sorted by tiles, it'll eventually start lagging if you write too much on it. Still, it is a first step.

You can activate debug (coloring as in the video + size info in the terminal) with the CMake flag `-DDEBUG_TILING=on`.

~~Note: there seems to be a memleak as the memusage is constantly increasing over time. I'll plug it of course.~~  Edit: done

PR #7266, opened 2026-03-14
Use a quad-tree as buffer for PageView

bhennion:
This is an alternative to #6594 with the same objectives:

1. Improve rendering performances and cost
2. Contain the RAM usage even with huge page size and high zoom level

In #6594, the buffer consists in tiles, all rendered with the same zoom level.
In this PR, the buffer is a quad-tree (each node has up to 4 children):

- The root corresponds to the whole page.
- For each node, each of the 4 possible children correspond to a quarter of the area covered by the node.
- Each node may of may not contain a buffer.
- Every buffer has the same size (=16MB in RAM)

This means that the root holds a rendering of the entire page, its children each may have a buffer of a quarter of the page, but with double the resolution, as so on.

The major benefit of that is: when we zoom-in/out, we do not call for a rerendering all the time: we just use the buffer with "at least enough pixels" to show the current zoom level.
The buffers are made on demand: if we need to show this part of the page with that precision, we look for a suitable buffer (or suitable buffers). If none exist, we trigger their creation.

Now, to avoid the memory usage from exploding, we regularly discard old (as in not needed recently) buffers.

Notes:
- There was an old PR #2982 doing a similar thing with the PdfCache (but not the PageView buffer).
- As #6594, this PR does not touch PdfCache, so the benefits will not be fully felt when using a PDF background.

Todos:
- [x] Also use quad tree in PdfCache (almost ready)
- ~~[ ] Allow for higher zoom limit~~

### DIFF
--- a/src/core/view/Mask.cpp
+++ b/src/core/view/Mask.cpp
@@ -114,6 +114,7 @@ void Mask::paintTo(cairo_t* targetCr) const {
     xoj::util::CairoSaveGuard guard(targetCr);
     cairo_scale(targetCr, 1. / zoom, 1. / zoom);
     cairo_set_source_surface(targetCr, cairo_get_target(const_cast<cairo_t*>(cr.get())), xOffset, yOffset);
+    cairo_pattern_set_filter(cairo_get_source(targetCr), CAIRO_FILTER_FAST);
     cairo_paint(targetCr);
 }
 
@@ -122,6 +123,7 @@ void Mask::paintToWithAlpha(cairo_t* targetCr, uint8_t alpha) const {
     xoj::util::CairoSaveGuard guard(targetCr);
     cairo_scale(targetCr, 1. / zoom, 1. / zoom);
     cairo_set_source_surface(targetCr, cairo_get_target(const_cast<cairo_t*>(cr.get())), xOffset, yOffset);
+    cairo_pattern_set_filter(cairo_get_source(targetCr), CAIRO_FILTER_FAST);
     cairo_paint_with_alpha(targetCr, alpha / 255.0);
 }
 


bhennion:
I adapted the quad-tree to be also usable in the main PdfCache, so we now can zoom-in in PDF documents as well. I think this is ready for a round of testing.
A couple of comments:

1. By default the cache is limited to 500MB (x2 = 1GB if a Pdf is involved). This is configurable in the Settings. Note that the memory usage can spike above this limit when scrolling.
2. You can see the tiles using -DDEBUG_QUAD_TREE=on
3. I put a little bit of "preloading": when a page is displayed, an area around the visible part is prerendered just in case the user scrolls.

To anyone willing to try this out, I'd love to get feedback on how it feels.

rolandlo:
I'm sorry to report, but on my device (Lenovo ThinkPad X1 Yoga Gen 6 running Ubuntu 25.10) it feels very laggy (with default settings).

bhennion:
> I'm sorry to report, but on my device (Lenovo ThinkPad X1 Yoga Gen 6 running Ubuntu 25.10) it feels very laggy (with default settings).

Mm. that's bad. Could you be a bit more specific?

1. Is navigation laggy (scrolling/zooming in or out)? Or is it only when you write?
2. Are you using Gnome+Wayland? Something else?
3. Is there some fractional scaling in your display configuration?

rolandlo:
> > I'm sorry to report, but on my device (Lenovo ThinkPad X1 Yoga Gen 6 running Ubuntu 25.10) it feels very laggy (with default settings).
> 
> Mm. that's bad. Could you be a bit more specific?
> 
>     1. Is navigation laggy (scrolling/zooming in or out)? Or is it only when you write?
> 
Already loading the very first page takes longer than usual. Navigation (scrolling by pages via PageDown) is lagging badly. Haven't even started to zoom or write.

>     2. Are you using Gnome+Wayland? Something else?
> 
Wayland
>     3. Is there some fractional scaling in your display configuration?

No. Just 100% scaling.

tattsan:
I'm testing this PR on an environment consisting of ‘X11 / Arch Linux / Wacom Cintiq Pro 32 / ThinkPad X1 Extreme Gen3 (NVIDIA GeForce GTX 1650 Ti with Max-Q Design (4GB GDDR6))’.
Scrolling and zooming feel slightly lagger than before. When I set Xournal++'s DPI to the maximum of 400 and zoom in to 700% to use the pen, there is a delay in rendering. With long strokes, the delay only occurs at the beginning.
Specifically,
➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰➰ (actually a continuous curve)
When moving the pen like this, the first ➰➰ or so isn’t drawn until it’s finished, and everything is drawn at the end. After that, the drawing follows the pen’s movements in real time. Before applying the patch, there was no drawing lag even with this level of magnification.

rolandlo:
By the way the app crashes when opening a document with missing PDF background:
```term
#0  0x0000562c6689a138 in Stacktrace::printStacktrace(std::ostream&) at inst/bin/xournalpp
#1  0x0000562c6650af03 in crashHandler at inst/bin/xournalpp
#2  0x00007a611f845f5f at /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007a611f8a74e4 in pthread_mutex_lock at /lib/x86_64-linux-gnu/libc.so.6
#4  0x0000562c6651d527 at inst/bin/xournalpp
#5  0x0000562c6651d7e3 at inst/bin/xournalpp
#6  0x0000562c6665a1a6 in xoj::view::QuadPdfCache::paintSingleTile(_cairo*, unsigned long, xoj::view::QuadCache::TileInfo const&, double, double) at inst/bin/xournalpp
#7  0x0000562c667ed592 in RenderJob::run() at inst/bin/xournalpp
#8  0x0000562c666cc880 in Job::execute() at inst/bin/xournalpp
```

bhennion:
Alright. So I could reproduce you laggy feeling (more gittery than laggy I would say) on a different computer. I think the input device from my usual dev setup doesn't produce fast enough events to notice the bottleneck issue.
I'll figure out where it comes from.

bhennion:
It should be much better now. I had forgotten to tell cairo to go fast :)

rolandlo:
> It should be much better now. I had forgotten to tell cairo to go fast :)

It's definitely faster with this fix, but the rendering quality has suffered a lot. I can see it especially on fonts from PDF backgrounds. Example (left is master, right is this PR):

Bildschirmfoto vom 2026-03-28 21-55-16

bhennion:
Yes you are right. There is no inbetween though... It may kill this idea all together. I'll try to think of something closer to #6594 but maybe a bit more conservative of resources.

LittleHuba:
I would not throw this idea away too fast. Why not use this approach for fast zooming and then when the zooming is finished, render in the correct resolution?

bhennion:
Yes, I've been pondering this idea as well, but in the end, I don't quite see a benefit in using a quad-tree for that.
Whenever we start a zoom sequence, we could use whatever rendered zoom level we have in cache and only rerender once the zoom-sequence is finished.

Another possibility would be to have a binary tree instead of a quad tree: we only split in two (e.g. vertically first, then horizontally on the next level, and so on). This will add intermediate buffers, possibly mitigating the loss in quality.

Also, keeping a quad/binary tree for the PdfCache anyway, while using the slow rescaling in Cairo, looks somewhat reasonable: this is done asynchronously anyway.

bhennion:
> Another possibility would be to have a binary tree instead of a quad tree: we only split in two (e.g. vertically first, then horizontally on the next level, and so on). This will add intermediate buffers, possibly mitigating the loss in quality.

No, that didn't work.

Issue #1883, opened 2020-04-21
Rotate pages


atticus-sullivan:
Hi,
it would be nice, if you were able to rotate pages (90° (counter)clockwise) while annotating a pdf (sometimes people having trouble to insert the pages in the right orientation into the pdf, especially when the pdf is made by scanning pages)

All3xJ:
it should definietely be a feature

ioneliabuzatu:
Also rotating them would give more space in general where to write.

Rmano:
Just a suggestion until this feature is in: I use extensively PDFarranger: https://github.com/pdfarranger/pdfarranger *before* annotating with `xournalpp`.  It is a life-saver ;-). 

What would be great is also having a rotate-in-current page option: for example, when annotating a portrait A4 document (say an exam), rotate and scale the page so that it occupies one side of a landscape A4 page. You can probably concoct something like this with `pdfjam`: https://github.com/rrthomas/pdfjam but well, it's a bit non-straightforward.

raffaem:
> Just a suggestion until this feature is in: I use extensively PDFarranger: https://github.com/pdfarranger/pdfarranger _before_ annotating with `xournalpp`. It is a life-saver ;-).
> 
> What would be great is also having a rotate-in-current page option: for example, when annotating a portrait A4 document (say an exam), rotate and scale the page so that it occupies one side of a landscape A4 page. You can probably concoct something like this with `pdfjam`: https://github.com/rrthomas/pdfjam but well, it's a bit non-straightforward.

<advertisement>
My [PDFsak](https://github.com/raffaem/pdftools) is also able to rotate PDF pages.
But you will lose annotations: so do it before annotating the PDF.
</advertisement>

GMTSE:
Just being able to (temporarily) rotate the view would already be an improvement.

leochras:
This should have higher priority, because this feature would greatly improve xournal's PDF reading capabilities..

Krzmbrzl:
I think this would also be a great feature when taking manual notes, if for some reason (e.g. space or layout) one can't simply write from left to right (horizontal) but would want to e.g. write along vertical line.

leochras:
I can confirm that (Arch Linux, xournalpp 1.1.1+dev)

As this is an open issue, I assume that this feature is not implemented yet.

atticus-sullivan:
`journal` -> `paper format` works for me in
```bash
$ xournalpp --version
Xournal++ 1.1.1
[...]

PR #7268, opened 2026-03-14
Feature: Rotating PDF pages

The-An0nym:
Seeks to implement https://github.com/xournalpp/xournalpp/issues/1883

## Implemented

- [x] UI for rotating individual PDF pages (may have to be adjusted)*
- [x] Rotation of individual PDF pages (via new page attribute) 
- [x] Adjusting page layout (toggle landscape/portrait)
- [x] File handling (save, load, print, export)
- [x] Add undo/redo

*I've added two new buttons on the right click menu of the document overview ("Rotate clockwise" and "Rotate counterclockwise").

## Feature(s) for a later date...

- [ ] Rotate layer elements with the page (images, text, strokes, etc.)

See https://github.com/xournalpp/xournalpp/pull/5596
A comment has been left in `Control.cpp` as an indicator for anyone wishing to implement the feature(s)

- [ ] Adjusting text selection

The implementation of this exceeds my current capabilities.

## Disclaimer

* Code may not be optimized (speed, memory, integration)
* Opening the new file format with older versions of Xournalpp can lead to weird page layouts

## Testing

I do not see how this feature could be tested automatically, thus I had to do so manually. I've used various testing PDFs, some of which already containing rotated pages, annotated them and exported or printed them (as file). This includes PostScript, as well as both exporting back ends (Cairo and QPDF). I hope to have caught all edge cases, but given the tediousness of testing, I may very well have missed one, so further testing is encouraged.

Tested on: Linux NixOS

## Notes

I've tried my best to keep the changes minimal, clean and appropriate, and have extensively tested with various PDFs. A new attribute `pageorient` would be added to the .xopp file. Old files can be read perfectly fine with my changes, however the new file version can cause a mismatch between PDF orientation and page orientation (landscape/portrait) when opened with older versions of Xournalpp.

Given that this feature has been requested over 5 years ago, I think it would be nice if this (i.e. an adjusted version of this) could already be merged with the main branch until other features are implemented at a later date - which would require rotation support for images, text, etc. which do not seem likely to be added in the near future and exceed my current capabilities. Personally, I think the necessity of such a feature is debatable, since I would rarely have a need for it (I would rotate the (PDF) pages _before_ annotating, not after), but I've included it nonetheless, since it would be the expected behavior of the fully implemented.


src/core/control/Control.cpp
```diff
@@ -828,6 +829,27 @@ void Control::movePageTowardsEnd() {
...
+    this->doc->lock_shared();
+    auto const& p = this->doc->getPage(pNr);
+    this->doc->unlock_shared();
+
+    if (!p) {
+        return;
+    }
+
+    int orient = p->getPdfPageOrientation();
+    int newOrient = (orient + n + 4) % 4;
```
rolandlo: Isn't that the same thing as
```suggestion
    int newOrient = (orient + n) % 4;
```

src/core/control/PrintHandler.cpp
```diff
@@ -41,11 +41,26 @@ void drawPage(GtkPrintOperation* /*operation*/, GtkPrintContext* context, int pa
...
+
     // For better quality printing, we use a dedicated pdf-renderer in this case
     if (page->getBackgroundType().isPdfPage()) {
         XojPdfPageSPtr popplerPage = doc->getPdfPage(page->getPdfPageNr());
         if (popplerPage) {
+            if (pageOrient != 0) {
+                cairo_save(cr);
+                cairo_translate(cr, width / 2, height / 2);
+                cairo_rotate(cr, pageOrient * M_PI_2);
+                if (pageOrient != 2)
```
rolandlo: Using an enum for the page orientation would make the code more readable.

rolandlo (commented):
Thanks for the PR. I have tested it out quickly, skimmed through the code and  would like to share some feedback.

- I can't say yet if the approach is good or not, but it does make some sense, since it's more convenient to rotate pages like that than modifying the PDF background with an external tool and reloading it in Xournal++. A user would expect the content to be rotated along with the background, so at least we should think about how the transition would look like if that would be implemented one day as well
- Please follow [coding conventions](https://github.com/xournalpp/xournalpp/wiki/Coding-conventions), in particular with regards to the rule:
> Each if, for, while etc. are followed by curly braces, even if it only contains one line of code 
- There were some warnings that should be fixed
```term
xournalpp/src/core/pdf/base/QPdfExport.cpp:112:53: warning: conversion from ‘long long int’ to ‘int’ may change value [-Wconversion]
  112 |                 int currRot = rotateObj.isInteger() ? rotateObj.getIntValue() : 0;
      |                               ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
and
```term
xournalpp/src/core/view/background/PdfBackgroundView.h:39:15: warning: ‘xoj::view::PdfBackgroundView::pdfCache’ will be initialized after [-Wreorder]
   39 |     PdfCache* pdfCache = nullptr;
      |               ^~~~~~~~
xournalpp/src/core/view/background/PdfBackgroundView.h:38:9: warning:   ‘int xoj::view::PdfBackgroundView::pageOrient’ [-Wreorder]
   38 |     int pageOrient;
      |         ^~~~~~~~~~
xournalpp/src/core/view/background/PdfBackgroundView.cpp:13:1: warning:   when initialized here [-Wreorder]
   13 | PdfBackgroundView::PdfBackgroundView(double pageWidth, double pageHeight, size_t pageNo, PdfCache* pdfCache,
      | ^~~~~~~~~~~~~~~~~
```
- The `DEV_FILE_FORMAT_VERSION` should be bumped in `CMakeLists.txt`, so users would be informed about opening a file created with the new version using an older version of Xournal++

src/core/control/Control.cpp
The-An0nym: Would be nice if it was, but the expression `orient + n` can take on negative values to represent counter clockwise rotation. Rotating a page counter clockwise for the first time would be `(0 - 1) % 4` which would equal `-1`
Technically my approach isn't entirely safe either, as it assumes that `n >= -4`, but it should suffice. 
If `0 <= newOrient <= 3` should be ensured for *any* input n, then `(orient + (n % 4) + 4) % 4` would do the trick.

src/core/control/PrintHandler.cpp
The-An0nym: Agreed, although finding intuitive names is a bit tricky - I've settled on `UP` `RIGHT` `DOWN` `LEFT` for now.

The-An0nym:
Thanks for the feedback. Most suggestions should be implemented now and the code should be updated to work with the newly created PageOrientation enum class.

> A user would expect the content to be rotated along with the background, so at least we should think about how the transition would look like if that would be implemented one day as well

I'm unsure what can already be thought of right now without knowing how images, text, etc. rotation would be implemented and the features that it would bring, but I imagine they would likely be rotated individually and using their center as pivot, meaning there would need to be some loop iterating over all page elements, rotating them individually and updating their position (elements that could be grouped together would of course be rotated together, e.g. annotations).

The-An0nym:
It seems some conflicts had emerged. They should be resolved now.

PR #7269, opened 2026-03-15
Solved TODO tasks: Constexpr string concatenations

lluisalemanypuig:
In PageTypeSubmenu.cpp I found some TODO tasks that asked to replace runtime string concatenations with compile-time string concatenations using C++20. I solved the tasks, though not with C++20, since it cannot be done (neither with C++20, or C++23) due to the restricted use cases of operator new during compilation time (see [this](https://stackoverflow.com/a/79736655/12075306) answer in a stackoverflow post).

I took inspiration from [this video](https://youtu.be/8rmnpuAh568?si=xXnqYvU0aiSaelP_) to do these tasks, the code in the second commit is explained [here](https://lluisalemanypuig.github.io/blog/2026/03/14).

Note: I want to make it extremely clear that the code I'm submitting here is quite similar to the original code from [this repo](https://github.com/think-cell/think-cell-library/). That code is licenced with a different licence than xournalpp's and I'm not sure how to proceed in this case.


lluisalemanypuig:
I just scanned the project for more chances of compile-time string concatenations and I couldn't see any more cases where it could be applied. I found some path concatenations here and there that could be done at compile-time, but this would require assuming the specific path separator for each platform and this may not be desirable.

Issue #7190, opened 2026-02-25
Plugin API to get current position


michelangelo4:
### Is Your Feature Request Related to a Problem?

I want to create a plugin that adds marks functionality like in neovim and want to jump to the exact spot I saved.

### Feature Description

As far as I can see in the [api docs](https://github.com/xournalpp/xournalpp/blob/master/plugins/luapi_application.def.lua) there is a function `app.scrollToPos` that scrolls to a specific position on a page but not something like `app.getCurrentPos` to get the current scroll position of the user on the page. It would be nice to have such a feature.

youssefadly237:
unfortunately this does not seem to be possible currently, check #6682 for ref

rolandlo:
> unfortunately this does not seem to be possible currently, check [#6682](https://github.com/xournalpp/xournalpp/pull/6682) for ref

The scroll position is not the same as the cursor position. The absolute scroll position could be easily obtained through
```C++
auto rect = control->getWindow()->getLayout()->getVisibleRect();
```
The absolute scroll position is then (rect.x, rect.y)

youssefadly237:
oh sorry, I misread the issue, thanks @rolandlo for correcting me

michelangelo4:
If it is not that hard, could this be implemented? I would really appreciate it.

youssefadly237:
If you want to do it yourself 

You will have to add another function to `src/core/plugin/luapi_application.h`

Call it `applib_getPos` for example, and add good docs

Add the function to the `luaL_Reg`

Test it

Use `scripts/lua_def_file.py`to update the docs **do not update it manually**

I can help with that if you don't want

michelangelo4:
So I added 
```
/**
 * Obtains the current scroll position.
 *
 * @return {x:number, y:number}
 *
 * Example: local scrollPos = app.getScrollPos()
 *
 * return value:
 * {
 *     ["x"] = number,
 *     ["y"] = number
 * }
 **/

static int applib_getScrollPos(lua_State* L) {
    Plugin* plugin = Plugin::getPluginFromLua(L);
    Control* control = plugin->getControl();
    auto rect = control->getWindow()->getLayout()->getVisibleRect();

    lua_newtable(L);                    // create table for current image

    // "x": number
    lua_pushnumber(L, rect.x);
    lua_setfield(L, -2, "x");

    // "y": number
    lua_pushnumber(L, rect.y);
    lua_setfield(L, -2, "y");

    return 1;
}
```
[...]

Issue #7120, opened 2026-02-13
`app.scrollToPos` changes page in absolute mode


tobiasBora:
### Operating System

Linux

### (Linux only) Distribution

NixOs

### (Linux only) Desktop Environment

Plasma

### (Linux Only) Display Server

Wayland

### Installation Method

Nix

### Xournal++ Version

master branch

### libgtk Version

3.24.51

### Bug Description

In a lua script I do something like:
```
      app.setCurrentPage(228)
      app.scrollToPage(228)
      app.refreshPage() -- Hope it helps, but no
      app.scrollToPos(30.38, 735.35, false) -- Absolute mode.
```
to center the page on a text (that's how I obtain the x/y). Sadly, this moves me to the first/second/3rd page depending on my zoom level… (the more I unzoom the further it goes in the document).

### Expected Behaviour

I should stay on the page 228, and the text at the specified position should be centered on the page irrespective of it's position in the page and irrespective of the zoom level.

### Steps to Reproduce

1. Create a lua plugin with a single function like:
```
function initUi()
  app.registerUi({["menu"] = "GradeExam/Advanced: debug", ["callback"] = "debug"});
end
function debug()
   app.setCurrentPage(30)
   app.scrollToPage(30)
   -- I want to center the view on this new text:
   app.addTexts({texts={{
                       text="Hello World",font={name="Noto Sans Mono Medium", size=8.0},color=0x1259b9,x = 30.38,y = 735.35
   }}})
   app.refreshPage() -- Hope it helps, but no. At least text is displayed.
   app.scrollToPos(30.38, 735.35, false) -- Absolute mode.
end
```
2. Create a new empty document with around 35 pages (starting `xournalpp` & clicking on `+` 35 times).
3. Run the Plugin > Debug function. See that you endup on page 1 instead of page 30. Try to unzoom until you see 3 pages (vertical mode) in the same window and click again. You end-up on page 3 now.

**EDIT 2**

I found how to implement what I wanted (except that it scrolls to put the text on top of the window instead of center, it's good enough for me but I'm still curious if you know if you know how to center it irrespective of the window size) by playing with the zoom & relative scrolling as follows:
```
function scrollTo(page, x, y)
   local zoom = app.getZoom()
   app.setCurrentPage(page)
   app.scrollToPage(page)
   app.scrollToPos(x*zoom, y*zoom)
end
```
[...]

PR #7278, opened 2026-03-16
Plugin API to get current Scroll position

michelangelo4:
This would be the pull reqeust to #7190. 

This adds the functionality to the api to get the current scroll position.

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -2647,7 +2647,8 @@ static int applib_scrollToPos(lua_State* L) {
 }
 
 /**
- * Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently visible window, measured in pixels.
+ * Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently
+ * visible window, measured in pixels.
  *
  * @return {x:number, y:number, width:number, height:number}
  *
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -730,7 +730,8 @@ function app.scrollToPage(page, relative) end
 --- scrolls to absolute pixel coordinates (200, 50) from top left corner of the layout (absolute mode)
 function app.scrollToPos(x, y, relative) end
 
---- Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently visible window, measured in pixels.
+--- Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently
+--- visible window, measured in pixels.
 --- 
 --- @return {x:number, y:number, width:number, height:number}
 --- 


rolandlo (commented):
Please apply clang-format. Also I'm wondering if it wouldn't be better to push the width and height of the visible rectangle as well. Otherwise we'll have to add some new API for that later as well.

michelangelo4:
I ran `clang-format -i luapi_application.h `, I hope this is what you mean.

How would be get width and height? I can do it if you tell me how to do it

rolandlo:
> I ran `clang-format -i luapi_application.h `, I hope this is what you mean.
> 
We usually one apply clang-format to the commits of the PR, but in that case it didn't make a difference.

> How would be get width and height? I can do it if you tell me how to do it

Well, `rect` is a `Rectangle<double>`, so you can just use `rect.width` and `rect.height`.

michelangelo4:
I added width and height, however when I test it those numbers don't change when I zoom or scroll or change page. Is that normal?

rolandlo:
> I added width and height, however when I test it those numbers don't change when I zoom or scroll or change page. Is that normal?

`getVisibleRect()` returns the rectangle with respect to layout pixel coordinates. So the width and height are in pixels that you see on your screen and don't change with zoom.
See https://github.com/xournalpp/xournalpp/pull/7088#issuecomment-3941115110 for a graphic explanation.
Width and height do change when you make the window larger or smaller.

It would be good if you clarified that layout pixel coordinates are used. `app.scrollToPos` also uses layout pixel coordinates (and are not relative to a page). The documentation is wrong about that. Could you fix that as well while you are at it, thereby fixing #7120?

src/core/plugin/luapi_application.h
```diff
@@ -2613,17 +2613,17 @@ static int applib_scrollToPage(lua_State* L) {
...
  * @param y number
  * @param relative boolean
  *
  * Example 1: app.scrollToPos(20,10)
- * scrolls 20pt right and 10pt down (relative mode)
+ * scrolls 20pt right and 10pt down from current position (relative mode)
  *
  * Example 2: app.scrollToPos(200, 50, false)
- * scrolls to page position 200pt right and 50pt down from the left page corner  (absolute mode)
+ * scrolls to absolute coordinates (200, 50) from top left corner (document origin) (absolute mode)
```
rolandlo:
```suggestion
 * scrolls to absolute pixel coordinates (200, 50) from top left corner of the layout (absolute mode)
```

src/core/plugin/luapi_application.h
```diff
@@ -2613,17 +2613,17 @@ static int applib_scrollToPage(lua_State* L) {
...
- * Scrolls to the position on the selected page specified relatively (by default) or absolutely
+ * Scrolls to the position relatively (by default) or absolutely (whole layout)
  *
  * @param x number
  * @param y number
  * @param relative boolean
  *
  * Example 1: app.scrollToPos(20,10)
- * scrolls 20pt right and 10pt down (relative mode)
+ * scrolls 20pt right and 10pt down from current position (relative mode)
```
rolandlo:
```suggestion
 * scrolls 20 pixel right and 10 pixel down from current position (relative mode)
```

src/core/plugin/luapi_application.h
```diff
@@ -2646,6 +2646,47 @@ static int applib_scrollToPos(lua_State* L) {
     return 0;
 }
 
+/**
+ * Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently visible window.
```
rolandlo:
```suggestion
 * Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently visible window, measured in pixels.
```

michelangelo4:
When I wrote my marks plugin I noticed that these scroll functions work fine when the zooming level stays the same, but yield weird behaviour if I include the zoom by dividing and multiplying with the zoom level. This is because the spaces between the pages stay the same but are rescaled by the division and multiplication.
Maybe a function that does what the description said in the first place wouldn't be so bad? so I give it a page and coordinates and it scrolls to that position and vice versa the get function gives me the current page and position.

In lua I used the following logic:
```
local function getPageRelativePos(page)
    local rawPos = app.getScrollPos()
    local zoom   = app.getZoom()
    app.scrollToPage(page, false)
    local pageTopRaw = app.getScrollPos()
    app.scrollToPos(rawPos.x, rawPos.y, false)
    return {x = (rawPos.x - pageTopRaw.x) / zoom, y = (rawPos.y - pageTopRaw.y) / zoom}
end

local function scrollToPageRelativePos(page, x, y)
    local zoom = app.getZoom()
    app.scrollToPage(page, false)
    local pageTopRaw = app.getScrollPos()
    app.scrollToPos(pageTopRaw.x + x * zoom, pageTopRaw.y + y * zoom, false)
end
```

Maybe there is a more direct way using built-in functions?

michelangelo4:
I have to run `clang-format -i luapi_application.h` again, right? How would you do it, so what command would I have to run to apply clang-format to the commits of the PR?

rolandlo:
> When I wrote my marks plugin I noticed that these scroll functions work fine when the zooming level stays the same, but yield weird behaviour if I include the zoom by dividing and multiplying with the zoom level. This is because the spaces between the pages stay the same but are rescaled by the division and multiplication. Maybe a function that does what the description said in the first place wouldn't be so bad? so I give it a page and coordinates and it scrolls to that position and vice versa the get function gives me the current page and position.
> 
> In lua I used the following logic:
> 
> ```
> local function getPageRelativePos(page)
>     local rawPos = app.getScrollPos()
>     local zoom   = app.getZoom()
>     app.scrollToPage(page, false)
>     local pageTopRaw = app.getScrollPos()
>     app.scrollToPos(rawPos.x, rawPos.y, false)
>     return {x = (rawPos.x - pageTopRaw.x) / zoom, y = (rawPos.y - pageTopRaw.y) / zoom}
> end
> 
> local function scrollToPageRelativePos(page, x, y)
>     local zoom = app.getZoom()
>     app.scrollToPage(page, false)
>     local pageTopRaw = app.getScrollPos()
>     app.scrollToPos(pageTopRaw.x + x * zoom, pageTopRaw.y + y * zoom, false)
> end
> ```
> 
> Maybe there is a more direct way using built-in functions?

I wouldn't know of a more direct way and see your point in asking for adding two more functions to the API for that.

michelangelo4:
What do you think of this function to get the layout pixel coordinates of the (current) page?
```
static int applib_getPageRect(lua_State* L) {
    Control* control = plugin->getControl();
    
    // Get page number from Lua
    const size_t pageNr = static_cast<size_t>(luaL_checkinteger(L, 1)) - 1;
    
    // Check if page number is valid
    const size_t pageCount = control->getDocument()->getPageCount();
    if (pageNr >= pageCount) {
        lua_pushnil(L);
        lua_pushfstring(L, "page nr %zu is out of range (document has %zu pages)", pageNr + 1, pageCount);
        return 2;
    }
    
    // Get the XournalView and the PageView for the specified page
    XournalView* xournal = control->getWindow()->getXournal();
    XojPageView* pageView = xournal->getViewFor(pageNr);
    
    if (!pageView) {
        lua_pushnil(L);
        lua_pushfstring(L, "could not get view for page nr %zu", pageNr + 1);
        return 2;
    }
    
    // Get the page's position and dimensions
    auto pos = pageView->getPixelPosition();
    int width = pageView->getDisplayWidth();
    int height = pageView->getDisplayHeight();
    
    // Create table for return value
    lua_newtable(L);
    
    // "x": number
    lua_pushnumber(L, pos.x);
    lua_setfield(L, -2, "x");
    
    // "y": number
    lua_pushnumber(L, pos.y);
    lua_setfield(L, -2, "y");
    
    // "width": number
    lua_pushnumber(L, width);
    lua_setfield(L, -2, "width");
    
    // "height": number
    lua_pushnumber(L, height);
    lua_setfield(L, -2, "height");
    
    return 1;
}
```
This yields the same coordinates as the scrolling to the page and back method. With that one could easily calculate the relative coordinates on the page without all the scrolling.

rolandlo:
> What do you think of this function to get the layout pixel coordinates of the (current) page?
> 
> ```
> static int applib_getPageRect(lua_State* L) {
>     Plugin* plugin = Plugin::getPluginFromLua(L);
>     Control* control = plugin->getControl();
>     
>     // Get page number from Lua
>     const size_t pageNr = static_cast<size_t>(luaL_checkinteger(L, 1)) - 1;
>     
>     // Check if page number is valid
>     const size_t pageCount = control->getDocument()->getPageCount();
>     if (pageNr >= pageCount) {
>         lua_pushnil(L);
>         lua_pushfstring(L, "page nr %zu is out of range (document has %zu pages)", pageNr + 1, pageCount);
>         return 2;
>     }
>     
>     // Get the XournalView and the PageView for the specified page
>     XournalView* xournal = control->getWindow()->getXournal();
>     XojPageView* pageView = xournal->getViewFor(pageNr);
>     
>     if (!pageView) {
>         lua_pushnil(L);
>         lua_pushfstring(L, "could not get view for page nr %zu", pageNr + 1);
>         return 2;
>     }
>     
>     // Get the page's position and dimensions
>     auto pos = pageView->getPixelPosition();
>     int width = pageView->getDisplayWidth();
>     int height = pageView->getDisplayHeight();
>     
>     // Create table for return value
>     lua_newtable(L);
>     
>     // "x": number
>     lua_pushnumber(L, pos.x);
>     lua_setfield(L, -2, "x");
>     
>     // "y": number
>     lua_pushnumber(L, pos.y);
>     lua_setfield(L, -2, "y");
>     
>     // "width": number
>     lua_pushnumber(L, width);
>     lua_setfield(L, -2, "width");
>     
>     // "height": number
>     lua_pushnumber(L, height);
>     lua_setfield(L, -2, "height");
>     
>     return 1;
> }
> ```
> 
> This yields the same coordinates as the scrolling to the page and back method. With that one could easily calculate the relative coordinates on the page without all the scrolling.

It makes sense to me. Could you open a new PR with it? Thanks for the present PR, which I merged.

tobiasBora:
Great thanks a lot! Out of curiosity, can you use this (or something else) to get the width/height of the user window? I'd need it to add text in the middle of the visible screen. Also, please tag me on the other PR to get page coordinates.

michelangelo4:
If I understand your question correctly, this should be the width and height parameter of the `getScrollPos()`-function added in this PR.

PR #7287, opened 2026-03-18
Fix objects vanishing when selected near page edges and clicking on another page

### LINKED ISSUE #4432: Moving objects partially outside the page results in different behaviors
**Affects versions :**
 - OS: Linux (Ubuntu)
 - Desktop environment: X11
 - Version of libgtk: 2.24.29
 - Version of Xournal++: 1.1.2+dev
 - Installation method: self built master branch

**Describe the bug**
Drag & dropping objects "over the edge" of the page results in a weird behavior:
1. When the mouse is beyond the edge of the page when "dropping" the object & the objects are deselected afterwards the selection jumps back to the original position
2. When the mouse is on the page when dropping the selection, the selection will stay in this half-outside position.

It seems like a very weird and not intended behavior.

**To Reproduce**
1. Open any Xournal++ document (e.g. a new empty document)
2. Create any object (e.g. draw something with the pen)
4. Select the object with any selection tool
5. Drag and drop the object, so that part of it is outside the page...
  ... once with the mouse being outside the page when dropping
  ... once with it being still on the page
  & deselect the object (by e.g. clicking on an empty place in the page)
-> now the object will, or will not jump back, depending on where the mouse was

**Expected behavior**
I don't really know which, but it should only be one of the 2 behaviors that occur.
Ether you should, or you shouldn't be be able to drag items partially outside the canvas.

**Screenshots of Problem**
Demonstrating both options.
https://user-images.githubusercontent.com/88790311/203000654-8db425e1-4d2a-4eac-aa97-1d01a9f4ec49.mp4

zandivx:
When you select an object near a page edge and click elsewhere to deselect, the object can silently disappear. This happened because `finalizeSelection` used `getPageViewUnderCursor` to decide which page to re-insert elements into but this function relies on stale mouse position data from the last `mouseDown`, not the current cursor position. When the selection sits near a page boundary, the stale coordinates can resolve to an adjacent page, and elements get inserted there without any coordinate translation. They're gone from page 1 and invisible on page 2.

This replaces the `getPageViewUnderCursor` logic with a simpler approach: always finalize on the page the selection already belongs to (`this->view`), and clamp the position so at least 20pt stays on-page. Page changes during drag are already handled correctly by `moveSelection`/`translateToView`, so `finalizeSelection` never needs to second-guess the page.

Closes #4432.

### DIFF
--- a/src/core/control/tools/EditSelectionContents.cpp
+++ b/src/core/control/tools/EditSelectionContents.cpp
@@ -390,19 +390,6 @@ auto EditSelectionContents::computePageClampOffset(const xoj::util::Rectangle<do
         return {0.0, 0.0};
     }
 
-    // Fast path: if the overall transformed content box is already fully inside the page
-    // with the required overlap, every element is necessarily inside as well.
-    const double keepOnPageX = std::min(minOverlap, std::abs(snappedBounds.width));
-    const double keepOnPageY = std::min(minOverlap, std::abs(snappedBounds.height));
-    const double minX = std::min(snappedBounds.x, snappedBounds.x + snappedBounds.width);
-    const double maxX = std::max(snappedBounds.x, snappedBounds.x + snappedBounds.width);
-    const double minY = std::min(snappedBounds.y, snappedBounds.y + snappedBounds.height);
-    const double maxY = std::max(snappedBounds.y, snappedBounds.y + snappedBounds.height);
-    if (minX >= keepOnPageX && maxX <= pageWidth - keepOnPageX && minY >= keepOnPageY &&
-        maxY <= pageHeight - keepOnPageY) {
-        return {0.0, 0.0};
-    }
-
     double fx = bounds.width / this->originalBounds.width;
     double fy = bounds.height / this->originalBounds.height;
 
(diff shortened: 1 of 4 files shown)


zandivx:
@bhennion any idea regarding the build failure? This seems unrelated to my proposed changes.

rolandlo:
> @bhennion any idea regarding the build failure? This seems unrelated to my proposed changes.

This image conversion issue is already solved. When you push again, the checks should succeed.

zandivx:
> > @bhennion any idea regarding the build failure? This seems unrelated to my proposed changes.
> 
> This image conversion issue is already solved. When you push again, the checks should succeed.

Thank you!

bhennion:
This only seems to solve the issue when only one object is selected. If you select two objects, one of them could still disappear:

[Enregistrement d'écran_20260331_190400.webm](https://github.com/user-attachments/assets/35665401-1d1a-45af-9179-d473019e8af2)

I don't know what the perfect solution would be though.

zandivx:
> This only seems to solve the issue when only one object is selected. If you select two objects, one of them could still disappear:
> Enregistrement.d.ecran_20260331_190400.webm
> 
> I don't know what the perfect solution would be though.

Good catch, I didn't test this case! I'll try to come up with something

zandivx:
@bhennion I introduced a loop over all the elements in the selection to find constraints for clamping each element. So when moving multiple elements at once every element should stay on the page with at least 20pt (hard-coded threshold). This is done after `makeMoveEffective` so I don't have to deal with scaling.
The example video you provided works now as intended.

bhennion (changes requested):
The implementaiton of finalizeSelection() seems way more complicated than necessary. In the end, what you are doing amounts to ensuring that the rectangle `{this->x, this->y, this->width, this->height}` is (mostly) in the page.
There is no need to run through the elements for that as this information is readily available.

bhennion:
> The implementaiton of finalizeSelection() seems way more complicated than necessary. In the end, what you are doing amounts to ensuring that the rectangle `{this->x, this->y, this->width, this->height}` is (mostly) in the page. There is no need to run through the elements for that as this information is readily available.

Now that I think of it, what you are doing is probably not the same as this. If so, then you may disregard this comment.

bhennion:
On the other hand, in most cases, the selection will still be entirely inside a page. It's probably worth testing for it and optimizing this case.

zandivx:
Sorry for taking so long. I found several issues that need addressing, but the ones I found are now fixed:
- The clamp is now in `mouseUp`, so it happens instantly after you move and not only after you deselect. This was necessary to record a correct state for undo/redo.
- 20pt min overlap was too much now imo, so I changed it to 5pt
- A fast path was added, like you suggested, that skips the heavier calculation when the selection is clearly already on page. Rotations are an edge case that is hopefully handled correctly now.

src/core/control/tools/EditSelectionContents.cpp
```diff
@@ -382,6 +382,129 @@ InsertionOrder EditSelectionContents::makeMoveEffective(const xoj::util::Rectang
...
+        if (move) {
+            transformed->move(mx, my);
+        }
+        if (scale) {
+            transformed->scale(bounds.x, bounds.y, fx, fy, 0, this->restoreLineWidth);
+        }
+        if (rotate) {
+            transformed->rotate(snappedBounds.x + this->lastSnappedBounds.width / 2,
+                                snappedBounds.y + this->lastSnappedBounds.height / 2, this->rotation);
+        }
```
bhennion, line 474: This loop looks pretty expensive: cloning an object could be costful and applying the transformation in 3 different passes is also a waste (for strokes). In fact, the scale and move transformation can be applied on the resulting SnappingBox directly and does not required iterating on the Points of a stroke at all.

A better solution (to avoid the cloning business) would be to add a virtual function `Element::computeSnappedBounds(double rotation)` (or similar), which (for strokes), iterates on the Points, appling the (precalculated) rotation matrix to each and doing the usual min/max business.

PR #7295, opened 2026-03-20
Prebuild Windows xournalpp.ico file

bhennion:
Tries to fix the failing CI builds on Windows, due to a cryptic ImageMagick error (see also https://github.com/ImageMagick/ImageMagick/discussions/2655 for a similar error).

I have no idea how to fix ImageMagick, so I propose we ship the pre-packaged icon xournalpp.ico instead of repackaging every time.


bhennion:

bhennion:
> I agree with the approach. How was xournalpp.ico that is added in this PR generated? When I run the command in te README_HOW_TO_MAKE_ICO I get an icon which is much smaller (31501 bytes instead of 285478 bytes).

It was generated with this command. It removed it and did it once more: same result. I don't know much about how .ico files work. Could yours be compressed (e.g. holding PNG assets) and mine not (holding BMP)?
The raw size of the bitmaps should be 274688 bytes, so this theory holds.

Do you get a working icon on Windows with your .ico file?

knipknap:
Just a pointer: I [encountered the same issue](https://github.com/msys2/MINGW-packages/issues/28550) in another app with no code changes. I switched from Imagemagick to PIL to generate the .ico to resolve it though it seems that whatever causes the issue also affects Gtk4 (SVGs are no longer displayed in the app). So I assume this is an error in one of the dependencies of Imagemagick that is also used by Gtk4.

rolandlo:
> > I agree with the approach. How was xournalpp.ico that is added in this PR generated? When I run the command in te README_HOW_TO_MAKE_ICO I get an icon which is much smaller (31501 bytes instead of 285478 bytes).
> 
> It was generated with this command. It removed it and did it once more: same result. I don't know much about how .ico files work. Could yours be compressed (e.g. holding PNG assets) and mine not (holding BMP)? The raw size of the bitmaps should be 274688 bytes, so this theory holds.
> 
> Do you get a working icon on Windows with your .ico file?

Yes, it appears that in the [xournalpp.ico.zip](https://github.com/user-attachments/files/26156205/xournalpp.ico.zip) generated on my computer the icon holds a PNG asset (at least for the highest resolution) while yours holds BMP assets: 
```term
identify xournalpp.ico 
xournalpp.ico[0] PNG 256x256 256x256+0+0 8-bit sRGB 16399B 0.000u 0:00.000
xournalpp.ico[1] ICO 48x48 48x48+0+0 8-bit sRGB 0.000u 0:00.000
xournalpp.ico[2] ICO 32x32 32x32+0+0 8-bit sRGB 0.000u 0:00.000
xournalpp.ico[3] ICO 16x16 16x16+0+0 8-bit sRGB 31501B 0.000u 0:00.000
```

```
identify ~/Downloads/xournalpp.ico 
/home/roland/Downloads/xournalpp.ico[0] ICO 256x256 256x256+0+0 8-bit sRGB 0.000u 0:00.001
/home/roland/Downloads/xournalpp.ico[1] ICO 48x48 48x48+0+0 8-bit sRGB 0.000u 0:00.000
/home/roland/Downloads/xournalpp.ico[2] ICO 32x32 32x32+0+0 8-bit sRGB 0.000u 0:00.000
/home/roland/Downloads/xournalpp.ico[3] ICO 16x16 16x16+0+0 8-bit sRGB 285478B 0.000u 0:00.000
```

From this [StackOverflow thread](https://stackoverflow.com/questions/11423711/recipe-for-creating-windows-ico-files-with-imagemagick)
I gathered that using auto-resize may have drawbacks:
> Some answers also suggest using ImageMagick's icon:auto-resize command line argument to avoid creating separate PNG files for every resolution. I don't recommend using it because to get the best quality it is better to avoid resizing as it is less accurate than rendering SVG file into each resolution separately.

Also we may need to add more resolutions:
> Necessary sizes in Windows are 16, 24, 32, 48, 64, 128, 256. Most scripts forgot those. I am unsure if 96 is really needed, but it doesn't hurt.

rolandlo:
With these commands:
```term
for size in {16,24,32,48,64,96, 128,256}; do rsvg-convert ../ui/pixmaps/com.github.xournalpp.xournalpp.svg --width=$size --height=$size -o xournalpp-icon-$size.png; done
magick convert xournalpp-icon-*.png xournalpp.ico
```
I get an icon that gets identified as
```term
magick identify xournalpp.ico
xournalpp.ico[0] ICO 128x128 128x128+0+0 8-bit sRGB 0.000u 0:00.003
xournalpp.ico[1] ICO 16x16 16x16+0+0 8-bit sRGB 0.000u 0:00.002
xournalpp.ico[2] ICO 24x24 24x24+0+0 8-bit sRGB 0.000u 0:00.002
xournalpp.ico[3] PNG 256x256 256x256+0+0 8-bit sRGB 11540B 0.000u 0:00.001
xournalpp.ico[4] ICO 32x32 32x32+0+0 8-bit sRGB 0.000u 0:00.001
xournalpp.ico[5] ICO 48x48 48x48+0+0 8-bit sRGB 0.000u 0:00.001
xournalpp.ico[6] ICO 64x64 64x64+0+0 8-bit sRGB 0.000u 0:00.001
xournalpp.ico[7] ICO 96x96 96x96+0+0 8-bit sRGB 151762B 0.000u 0:00.000
```
and has a file size of `151762` bytes.

Doing the same thing via
```term
magick -density 300 -define icon:auto-resize=256,128,96,64,48,32,24,16 -background none ../ui/pixmaps/com.github.xournalpp.xournalpp.svg xournalpp.ico
```
identifies as
```term
magick identify xournalpp.ico
xournalpp.ico[0] PNG 256x256 256x256+0+0 8-bit sRGB 16399B 0.000u 0:00.002
xournalpp.ico[1] ICO 128x128 128x128+0+0 8-bit sRGB 0.000u 0:00.002
xournalpp.ico[2] ICO 96x96 96x96+0+0 8-bit sRGB 0.000u 0:00.001
xournalpp.ico[3] ICO 64x64 64x64+0+0 8-bit sRGB 0.000u 0:00.001
xournalpp.ico[4] ICO 48x48 48x48+0+0 8-bit sRGB 0.000u 0:00.000
xournalpp.ico[5] ICO 32x32 32x32+0+0 8-bit sRGB 0.000u 0:00.000
xournalpp.ico[6] ICO 24x24 24x24+0+0 8-bit sRGB 0.000u 0:00.000
xournalpp.ico[7] ICO 16x16 16x16+0+0 8-bit sRGB 156621B 0.000u 0:00.000
```
and has a file size of `156621` bytes (which is slightly more).

bhennion:
> With these commands:

I'm fine with both. Note that the first one leaves some lingering .png files in the source tree though.

> Just a pointer: I [encountered the same issue](https://github.com/msys2/MINGW-packages/issues/28550) in another app with no code changes. I switched from Imagemagick to PIL to generate the .ico to resolve it though it seems that whatever causes the issue also affects Gtk4 (SVGs are no longer displayed in the app). So I assume this is an error in one of the dependencies of Imagemagick that is also used by Gtk4.

@knipknap Thanks for the pointer!
@rolandlo Do you encounter some similar issues with the binaries generated in [this post](https://github.com/xournalpp/xournalpp/pull/7295#issuecomment-4096241529)? It may also affect GTK3. In that case, we'll need a further fix.

bhennion:
Also, @rolandlo, feel free to take over this PR. You are obviously more qualified than me here.

bhennion:
Apparently, the issue has been fixed in the CI images. Last night's build was successful (https://github.com/xournalpp/xournalpp/actions/runs/23393332903).

PR #7298, opened 2026-03-21
Fix: dashed/dotted lines not scaling to stroke width

### LINKED ISSUE #6955: Line style (dashed, dotted) should adjust to line thickness
### Operating System

Linux

### (Linux only) Distribution

Manjaro

Xfce

X.Org

### Installation Method

xournalpp-git AUR package (builds latest git)

### Xournal++ Version

xournalpp 1.3.1+dev (f590496d from xournalpp-git/master)

### libgtk Version

3.24.51

### Bug Description

When selecting a non-standard line style for the pen tool (dotted or dashed), the distance between adjacent dots/dashes is a fixed value. When selecting "very thick" as pen thickness, the dots/dashes are glued together, because their radius/half-length is larger than this distance.

### Expected Behaviour

The distance between adjacent dots/dashes should be scaled with the line thickness.

### Steps to Reproduce

1. Select very thick line thickness.
2. Select dotted or dashed line style.
3. Draw a line.
4. Observe that the line looks almost like default style because the dots/dashes are almost non-distinguishable.

### Additional Context

Image

For context: I need the very thick pen when annotating a high res scan (pupil solution).

andre-ortiz-pagaime:
# The Bug
The dash pattern is defined as an alternating array of dash and gap lengths and it was used directly regardless of stroke thickness. The thicker the stroke, the smaller the gap between dashes. For very thick strokes the length of the dash was too short relative to the width, so consecutive dashes overlapped.

# The Fix
I added getDashesScaledToStrokeWidth, which multiplies each dash length by strokeWidth. Thicker strokes get proportionally longer dashes and gaps, thinner strokes get proportionally shorter ones.

For pressure-insensitive strokes, it simply scales the dash pattern using getDashesScaledToStrokeWidth. For pressure-sensitive strokes, where stroke width varies while drawing, the dash offset is tracked in normalized (width-independent) units and converted back to physical units when needed.

Before:
dashed_lines_before

After:
dashed_lines_after

Closes #6955


youssefadly237:
this makes much more sense tbh

but one concern, could this change break backward compatibility?

on older files that were created assuming the previous (unscaled) dash behavior, they will render differently now

bhennion:
Overall, it sounds like a good idea!

> but one concern, could this change break backward compatibility?

This is certainly an issue. When opening an old file, the appearance changes, and it should really not.

For "constant width" strokes, there is an easy solution to that: we can store the dash pattern multiplied by the width as a custom dash pattern in the file (e.g. `style="cust: 2.3 5.4"` in the .xopp file).
This is already supported, so even older Xournal++ versions would open and render the new files as they should be.
It also means adapting the PR so the dash-pattern scaling happens only once (and not on every rendering calls).

When the width varies, I see no other solution than bumping the file format version number. Then we can decide either:

1.  to add new line style key words like "wdot" or whatever, saying that a stroke with that is dotted, with dash-pattern scaled by width.
2. or that "if the version is more than N, then dash patterns should be interpreted as width-sensitive".

andre-ortiz-pagaime:
I changed the pressure-insensitive logic to scale the dash pattern on stroke creation instead of on rendering, as suggested. Now dashed lines from old files still render unscaled and new files can be opened in older versions with the scaled dash pattern preserved.

getDashesScaledToStrokeWidth is still used in the pressure-sensitive PDF path, so I kept it.

For the pressure-sensitive problem, I couldn't find an alternative beyond the two options mentioned, but I noticed an issue with the version-based approach: if a user loads a file written in an older version, with unscaled strokes, and saves it in the newer version, the file version gets updated. The next time the file is opened, the old unscaled strokes would be rendered with the new scaling logic. Avoiding this would require storing the version number per stroke.

For that reason, I think the "wdot" approach may be the correct solution.

andre-ortiz-pagaime:
I added a scaled variant for each dashed stroke style. New strokes now use the scaled styles. The unscaled styles are still recognized and rendered unscaled, so older documents don't change, but it's not possible to draw new strokes with them. I bumped the file format version.

andre-ortiz-pagaime:
Hi @bhennion, when you have a moment, I’d appreciate your feedback on this PR.
Does this approach look good, or would you like any changes? Thanks!

Photon89:
Thanks for working on a fix! I didn't notice that there is a PR tackling the issue! :smiley:

src/core/model/LineStyle.cpp
youssefadly237, line 35:
```suggestion
void LineStyle::serialize(ObjectOutputStream& out) const {
    out.writeObject("LineStyle");

    out.writeData(this->dashes);
    out.writeInt(this->scaleDashesToWidth ? 1 : 0);

    out.endObject();
}

void LineStyle::readSerialized(ObjectInputStream& in) {
    in.readObject("LineStyle");

    in.readData(dashes);
    this->scaleDashesToWidth = in.readInt() != 0;

    in.endObject();
}
```

binary serialization loses `scaleDashesToWidth` flag, breaks on copy and paste

Image


I did not review the whole pr btw

youssefadly237 (changes requested):

src/core/model/StrokeStyle.cpp
```diff
@@ -15,13 +15,15 @@ namespace {
...
+        {"dash", {6, 3}}, {"dashdot", {6, 3, 0.5, 3}}, {"dot", {0.5, 3}},
+        {"scaled_dash", {6, 3}}, {"scaled_dashdot", {6, 3, 0.5, 3}}, {"scaled_dot", {0.5, 3}}};
 
-auto formatStyle(const std::vector<double>& dashes) -> std::string {
+auto formatStyle(const std::vector<double>& dashes, bool scaleDashes) -> std::string {
 
     // Check if dashes match named predefined dashes.
     for (auto& pair: predefinedPatterns) {
         if (pair.second == dashes) {
+            if (scaleDashes && !pair.first.starts_with("scaled")) continue;
```
bhennion: First, for readability:
```suggestion
if (pair.second == dashes && (!scaleDashes || pair.first.starts_with("scaled"))) {
```
but then, are you sure this is the test you want? Shouldn't it be
```suggestion
if (pair.second == dashes && (scaleDashes == pair.first.starts_with("scaled"))) {
```

bhennion (commented):
The code looks fine to me (with the exception of the comment below).

I have a bunch of requests though, before this could be merged:

1. Could you add test cases to our test suite (in the test directory) so that bugs like the one @youssefadly237 found would be covered. That includes our ser/des test, and the load/save tests.
2. Did you check that the PDF and screen outputs were exaclty the same? It is a bit hard to know if the formula you entered are the right ones. Examples like the ones from [this file](https://github.com/user-attachments/files/20500350/example-pp.pdf) could help detect issues (I'll try to find the original .xopp for those).
3. Do you know how performances are affected by the changes? Ideally, a benchmark like the one mentionned in [this post](https://github.com/xournalpp/xournalpp/pull/6385#issuecomment-3135258750) would be perfect.

andre-ortiz-pagaime:
Thanks for the feedback. I’ll address this in the next couple of days.

andre-ortiz-pagaime:
I found two more bugs while working on this and already fixed them:
- dash patterns weren’t being scaled correctly when scaling strokes
- changing the line style would result in unscaled dashes

I’ve added more tests, including for ser/des and load/save so cases like the one @youssefadly237 found should now be caught.

I also compared the exported PDF output against the source files, and the dashes seem to match exactly from what I could tell. (Example below)

I spend some time trying to adapt the benchmark from the post you linked, but I couldn’t get it working properly, so I don’t have meaningful performance numbers to share.
Screenshot_20260525_004313

andre-ortiz-pagaime:
Hi @bhennion, whenever you have some free time, I'd appreciate your review. Let me know if everything looks good or if any further changes are necessary.

bhennion:
Sorry, yes, I'll try to have a look this week.

In the meantime, would you mind rebasing on latest master and solving the conflits?

andre-ortiz-pagaime:
The upstream updates conflicted with my changes, breaking the loading of pressure-sensitive strokes. I've already fixed it. I also applied clang-format.

bhennion (approved):
Looks good to me.
@rolandlo I'd like to have your opinion on the feature before merging this (in case I missed a potential issue arrising from it)

src/core/model/StrokeStyle.cpp
```diff
@@ -65,13 +73,18 @@ auto StrokeStyle::parseStyle(const std::string& style) -> LineStyle {
 
     LineStyle ls;
     ls.setDashes(std::move(dashes));
+
+    if (scale_dashes) {
+        ls.setScaleDashes();
+    }
```
rolandlo: I'm confused about this. We only land here if the string `style` starts with `cust: `, which excludes the possibility of it starting with `scaled`. So `scale_dashes` can't be true here.

Shouldn't we have custom styles with scaled dashes? Something like `scaled_cust: ...`?

test/files/load/scaled-dashes.xopp
rolandlo, line 1: The underlying xml-file is missing.

test/files/load/unscaled-dashes.xopp
rolandlo, line 1: Here the xml-file is missing as well.

src/core/model/LineStyle.h
```diff
@@ -50,6 +50,27 @@ class LineStyle: public Serializable {
...
+    std::vector<double> getDashesScaledToStrokeWidth(const double strokeWidth) const;
+
+    /**
+     * Scale dash vector to stroke width
+     *
+     * @param strokeWidth
+     */
+    void scaleDashesToStrokeWidth(const double strokeWidth);
+
+    void setScaleDashes() { scaleDashesToWidth = true; }
```
rolandlo, line 69: It feels a bit strange to only allow setting `scaleDashesToWidth` to `true`. Wouldn't it make more sense to allow for setting it to `false` as well?

test/unit_tests/model/StrokeStyleTest.cpp
```diff
@@ -48,3 +48,25 @@ TEST(StrokeStyle, testFormatStyle) {
...
+
+    LineStyle dotLine;
+    dotLine.setScaleDashes();
+    dotLine.setDashes(std::vector<double>(dotLinePattern, dotLinePattern + 2));
+    EXPECT_EQ(StrokeStyle::formatStyle(dotLine), "scaled_dot");
+
+    LineStyle custLine;
+    custLine.setScaleDashes();
+    custLine.setDashes(std::vector<double>(custPattern, custPattern + 3));
+    EXPECT_EQ(StrokeStyle::formatStyle(custLine), "cust: 0.01 -1.00 0.01");
```
rolandlo: How would a user create such strokes (custom with scaled dashes)?

rolandlo (commented):
@bhennion What's our policy for bumping the file format version? We only did it when adding a new element (url links), but I guess we should do it here, too.
When you create a document with scaled dashes in the new version, load it in the old version and safe it there, not only the scaling of the dashes is lost, but the line style is lost completely (no dashes anymore).

bhennion:
> What's our policy for bumping the file format version? We only did it when adding a new element (url links), but I guess we should do it here, too.

I would argue that we only need to bump the file format version once between two releases, but that's debatable of course

test/unit_tests/model/StrokeStyleTest.cpp
bhennion: Select the pen with e.g. Rectangle shape on. Make a dashed stroke.
When saved, you'll get a `cust: ...`. The idea behind this is that those strokes can be read by previous versions of Xournal++ (the file format already supports this).
Now, it may be a false good idea and it may be better to keep them as `scaled_dashed` (or so), I don't know.

rolandlo: I get the point now. Still I think that strokes with custom dash pattern that scale to width are missing. Currently one can set a custom dash pattern via plugin. 

From the Lua definition file: 
```term
--- Example 6: app.changeActionState("tool-pen-line-style", "cust: 1 5 3 5")
```

This makes perfect sense for creating a new line style similar to those that already exist.
Now of course a user may want to have it scale to width, but it's not possible, since the file format wouldn't allow it. We would be stuck with the very arbitrary choice of predefined line styles (scaled_dash, scaled_dot, scaled_dashdot) that scale to width. We better introduce a `scaled_cust: ...` right away.

andre-ortiz-pagaime:
I added scaled_cust, so pressure-sensitive strokes with custom patterns also scale to stroke width. I also added the missing scaled-dashes.xml and unscaled-dashes.xml files.

test/files/load/scaled-dashes.xml
bhennion, line 1: I can't seem to find a `scaled_cust` test case in here. Would you mind adding one?

bhennion (approved):
LGTM (except the missing test case)
@rolandlo

test/files/load/scaled-dashes.xml
andre-ortiz-pagaime: There is testScaledFormatStyle, but that's a good idea to add a scaled_cust stroke to scaled-dashes.xopp, so it is tested like the other strokes.

andre-ortiz-pagaime:
I added a scaled_cust stroke to scaled-dashes.xopp and a pressure-sensitive cust to unscaled-dashes.xopp so they also get tested in the load handler tests.

plugins/luapi_application.def.lua
```diff
@@ -134,6 +134,7 @@ function app.registerUi(opts) end
 --- Example 4: app.changeActionState("tool-color", 0xff0000)         # red color
 --- Example 5: app.changeActionState("zoom", 2.25)
 --- Example 6: app.changeActionState("tool-pen-line-style", "cust: 1 5 3 5")
+--- Example 7: app.changeActionState("tool-pen-line-style", "scaled_cust: 1 5 3 5")  # dashes scale with stroke width
```
rolandlo, line 137: These two don't seem to make any difference. I get scaled dashes in both cases, which manifests itself in the saved xopp-file.

rolandlo (changes requested):
Also please rebase on master and fix the merge conflict.

plugins/luapi_application.def.lua
andre-ortiz-pagaime: The "scaled_" prefix only affects pressure-sensitive strokes, it's not used in pressure-insensitive strokes. New strokes, with or without pressure, always scale, by design, only old strokes are kept unscaled.

Earlier in this PR, only the predefined patterns had "scaled_" variants, so custom patterns with pressure couldn't scale. I added scaled_cust just to close that gap.

PR #7306, opened 2026-03-24
Update Example plugin to Lua 5.5

### LINKED ISSUE #7305: The Example plugin raises an error with Lua 5.5
### Operating System

Linux

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Arch pacman

### Xournal++ Version

1.3.3

### libgtk Version

irrelevent

### Bug Description

Image

### Expected Behaviour

no error

### Steps to Reproduce

1. Enable the Example plugin
2. Restart the app
3. Get the error popup

### Additional Context

_No response_

rolandlo:
Fixes #7305 

I'll check if there are other plugins we ship which are not yet compatible with Lua 5.5. At first glance (checking all for loops) I couldn't find any.

### DIFF
--- a/plugins/Example/main.lua
+++ b/plugins/Example/main.lua
@@ -1,12 +1,12 @@
 -- This is an example Xournal++ Plugin - copy this to get started
 
-var_dump = require "var_dump"
+local var_dump = require "var_dump"
 
 -- Register all Toolbar actions and initialize all UI stuff
 function initUi()
   print("Hello from Example: Plugin initUi called\n");
 
-  ref = app.registerUi({["menu"] = "Test123", ["callback"] = "exampleCallback", ["accelerator"] = "<Control>t"});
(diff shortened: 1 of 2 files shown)


rolandlo:
I have tested all plugins included in Xournal++ with Lua 5.5 and haven't discovered any more problems.

In order to build Xournal++ with Lua 5.5 I had to download the current [FindLua.cmake](https://raw.githubusercontent.com/Kitware/CMake/refs/heads/master/Modules/FindLua.cmake), since `cmake < 4.2.3` doesn't support Lua 5.5 (instead Lua <=5.4 is found). On Ubuntu 26.04 I have cmake 4.2.3, which is not recent enough.

As a side note, for installing modules for Lua 5.5 via Luarocks, the latest luarocks version (3.13.0 released in Dec 2025) is needed. I installed it with the older luarocks version (using `sudo luarocks install luarocks` and copying the config files `/etc/luarocks/config-5.*.lua` to `/usr/local/etc/luarocks/`).

Thus merging in 24 hours if no objections are raised.

PR #7318, opened 2026-03-27
Fix #7112: Replace findPolygonal with RDP for improved shape recognition

### LINKED ISSUE #7112: Shape recognizer turns thin rectangles into long triangles
### Operating System

Linux

### (Linux only) Distribution

Arch Linux

KDE Plasma

Wayland

### Installation Method

AUR

### Xournal++ Version

1.3.2

### libgtk Version

3.24.51

### Bug Description

When drawing a thin rectangle (not very tall and long), shape recognizer turns it into an absurdly long triangle that sometimes goes beyond the drawing area. 

### Expected Behaviour

It should draw a rectangle.

### Steps to Reproduce

1. Turn on shape recognizer
2. Draw a thin rectangle

### Additional Context

https://github.com/user-attachments/assets/59fa49fe-e32e-4b2f-94e8-35110153ea86

SalvadorPereira374:
This patch significantly improves the ShapeRecognizer component by addressing its sensitivity to hand-drawn tremor, which previously caused false corner detection and distorted shapes like thin rectangles. To fix this, the legacy approach was entirely replaced with the Ramer-Douglas-Peucker (RDP) algorithm. The RDP algorithm evaluates the stroke by finding the point with the maximum perpendicular distance from the line segment. By applying an empirically tested error threshold, it effectively filters out natural hand jitter while accurately identifying true, intentional corners.

Even with RDP, user strokes can occasionally produce more vertices than needed. To guarantee accurate rectangle and triangle recognition, a dedicated failsafe function named reduceVerticesToFour was implemented. This mechanism iteratively scans the identified corners and removes the vertex with the lowest perpendicular distance error until only the four most structurally significant internal corners remain. 
Closes #7112

### DIFF
--- a/src/core/control/shaperecognizer/ShapeRecognizer.cpp
+++ b/src/core/control/shaperecognizer/ShapeRecognizer.cpp
@@ -6,6 +6,7 @@
 #include <span>      // for span
 #include <utility>   // for move
 #include <vector>    // for vector
+#include <numeric>   
 
 #include <glib.h>  // for g_message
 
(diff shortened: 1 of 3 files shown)


src/core/control/shaperecognizer/ShapeRecognizer.cpp
```diff
@@ -39,6 +39,65 @@ inline double dist2(const Point& P, const Point& Q) {
...
 }
 
+void ShapeRecognizer::reduceVerticesToFour(std::vector<Corner>& cornerList) {
+    while (cornerList.size() > 5) {
+        size_t remove = 0;
+        double min = std::numeric_limits<double>::max();
+        // We iterate starting from index 1 and stop before the last index
+        // The first and last corners are the endpoints of the stroke and
+        // must be preserved to maintain the shape's boundaries
+        for (size_t x = 1; x < cornerList.size() - 1; x++) {
```
bhennion, line 49: This double loop is quite inefficient. The complexity here should be linear in N = cornerList.size(). Here, you have something in O(N^3) (counting the erase() below).

I'm guessing that we can get many corners in a single (very long) stroke, so it could matter.

I haven't read the rest of the PR, but could you please check and optimize all around: this would be called every time the user writes something, so efficiency is very important (if only for consumption concerns).

SalvadorPereira374: Hi Benjamin, thanks for the review and for catching that. I've updated the code and brought down the complexity to O(n). All tests are passing on my end.Let me know if there’s anything else I should adjust.

TheKingOfLight:
I tested it and found that it currently has problems detecting straight horizontal or vertical lines where Δx or Δy is too small and the 1.0 fallback is insufficient. Increasing the minimum offset to 5 works much better for me.
Here is an example of a horizontal line that was not identified:
image

SalvadorPereira374:
Sorry for the late reply @TheKingOfLight. I've applied the requested fix. Let me know if everything looks good now.

PR #7319, opened 2026-03-27
Plugin API to get page position

michelangelo4:
This pull request adds an API function to get the layout position in pixel coordinates of the current (or any other) page as discussed in #7278. I also added a comment on `getScrollPos()`, because I hadn't fully understood what was meant with visible window, so I figured others may have the same issue, too.

### DIFF
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -730,8 +730,9 @@ function app.scrollToPage(page, relative) end
 --- scrolls to absolute pixel coordinates (200, 50) from top left corner of the layout (absolute mode)
 function app.scrollToPos(x, y, relative) end
 
---- Obtains the current absolute scroll position (position on the whole layout) and width and height of the currently
---- visible window, measured in pixels.
+--- Obtains the current absolute scroll position (position of the upper left corner of the currently visible window on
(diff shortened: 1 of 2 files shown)


michelangelo4:
I wonder if it would be better to name `getScrollPos` something like `getVisibleWindowPosition` and setup separate functions for scrolling that are independent of zoom, because, intuitively, when I'd use `getScrollPos` and `scrollToPos` I would expect to be scrolled to where I actually got the scroll position from and not anywhere depending on what zoom level I have right now. In that regard, maybe renaming `scrollToPos` to `scrollToPixelCoordinate` would be an option. 
For the functions, getting scroll position would then look like: Calculate the center of screen-coordinates using data from the `getVisibleRect()` function (zooming should revolve around the screen center) and then calculate the relative coordinates of that center with respect to the current page's upper left corner with `getPixelCoordinates()`. Make them zoom independent by dividing by the current zoom level and save those effective coordinates along with the current page number.
To scroll to that position, first get the (due to zooming possibly new) pixel coordinates of the saved page, add the saved effective coordinates multiplied by the current zoom level to obtain the new center, then calculate the new upper left corner depending on the zoom and then scroll to that position using `scrollToPixelCoordinate()`. At some point one would have to add a `x = max(0, x)` for the case that we are zoomed out too much.
What do you think of that?

Issue #4110, opened 2022-06-21
Precise image resize


Cris70:
**Is your feature request related to a problem? Please describe.**
Xournal++ makes it easy to add an image to a PDF and then resize it with the mouse. However, if you want to resize it to an exact size, or you just want to know the resulting size after you resized with the mouse, you can't.

**Describe the solution you'd like**
There should be a GUI element that tells you the size of the image as you resize it (in the status bar?). There should also be the possibility to manually define the final size you want (something like right-click ==> properties).

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #7321, opened 2026-03-28
Add a properties dialog for selected images to allow for precise resizing

Cris70:
## Summary

This adds a context menu for a single selected image with two actions:

- `Properties`
- `Clone`

The new `Properties` action opens a dialog that shows the current image width and height, allows entering exact dimensions, and supports optional proportion locking. 
This PR is developed as a solution for issue #4110 

## Details

- right-clicking a single selected image now opens an image-specific popover instead of immediately cloning it
- the popover still exposes `Clone`, so the previous action remains available
- `Properties` opens a dialog for exact width and height editing
- the dialog uses the current selection bounds, so values stay in sync with later mouse resizes
- when `Keep proportions` is enabled, leaving one dimension field updates the other accordingly
- applying the dialog resizes the image through the existing undoable selection update path

## Notes

- this only applies to selections containing exactly one image element
- non-image selections and other selection paths keep their existing behavior

## Disclosure

This has been developed with the help of OpenAI Codex, mainly to speed up the understanding of the codebase and to accelerate development. I have tried to review all the code produced, and I have extensively tested the new functionality.

This was developed on linux. I have not compiled this on Windows or on MacOS.

### DIFF
--- a/src/core/gui/inputdevices/PenInputHandler.h
+++ b/src/core/gui/inputdevices/PenInputHandler.h
@@ -25,6 +25,12 @@ enum class PressureMode { NO_PRESSURE, DEVICE_PRESSURE, INFERRED_PRESSURE };
 class PenInputHandler: public AbstractInputHandler {
 protected:
     /**
+     * Mouse-specific hint: opening a context popover may swallow the matching
+     * release event, so subclasses can clear any latched pressed-button state.
+     */
+    bool cancelButtonPressLatch = false;
+
+    /**
      * Whether the current device class has a device with button 1 in pressed state
      */
     bool deviceClassPressed = false;
--- a/src/core/gui/inputdevices/MouseInputHandler.cpp
+++ b/src/core/gui/inputdevices/MouseInputHandler.cpp
@@ -37,6 +37,10 @@ auto MouseInputHandler::handleImpl(InputEvent const& event) -> bool {
         if (pressedButton == BUTTON_NONE || !inputRunning) {
             pressedButton = event.button;
             this->actionStart(event);
+            if (this->cancelButtonPressLatch) {
+                pressedButton = BUTTON_NONE;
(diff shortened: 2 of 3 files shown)


src/core/gui/inputdevices/PenInputHandler.h
```diff
@@ -24,6 +24,12 @@ enum class PressureMode { NO_PRESSURE, DEVICE_PRESSURE, INFERRED_PRESSURE };
 
 class PenInputHandler: public AbstractInputHandler {
 protected:
+    /**
+     * Mouse-specific hint: opening a context popover may swallow the matching
+     * release event, so subclasses can clear any latched pressed-button state.
+     */
+    bool cancelButtonPressLatch = false;
```
bhennion, line 31: The way that those kind of issues are typically handled is by only opening the dialog/popover when the button is released.

bhennion:
> * right-clicking a single selected image now opens an image-specific popover instead of immediately cloning it

I'm not certain about this. We typically do not hard-code what device does what. The right button of the mouse could be configured to pick a different tool, for instance.
Also, how would/could the feature be made accessible to touch-only devices?

src/core/gui/inputdevices/PenInputHandler.h
Cris70: Thank you @bhennion, I'll try to address this issue by following your suggestion.

Cris70:
> > * right-clicking a single selected image now opens an image-specific popover instead of immediately cloning it
> 
> I'm not certain about this. We typically do not hard-code what device does what. The right button of the mouse could be configured to pick a different tool, for instance. Also, how would/could the feature be made accessible to touch-only devices?

Thank you for your input.
So how do you think this should be implemented?
How would you approach the need expressed in issue #4110 ?

PR #7323, opened 2026-03-28
hierarchical plugin menus via `parentPath`

### LINKED ISSUE #7317: Submenu for lua plugins

I wish to organize the menu created by my lua plugin as for now they all appear in a huge, hard-to-search, list.

### Feature Description

Allow lua plugins to register submenu, for instance by turning `/` into a submenu.

### Alternative Solutions

_No response_

### Additional Context

See https://github.com/xournalpp/xournalpp/discussions/7119

youssefadly237:
plugins can now place menu items into nested submenus

```lua
app.registerUi({menu="PDF", callback="exportPDF", parentPath="Tools/Export"})
```

I made an earlier attempt encoded the hierarchy directly into the `menu` field (`menu="Tools/Export/PDF"`), but this broke existing plugins that have slashes in their label strings and conflated two distinct conceptsc into one field, `parentPath` keeps label and location separate.

No existing plugins are affected.

closes #7317

### DIFF
--- a/plugins/luapi_application.def.lua
+++ b/plugins/luapi_application.def.lua
@@ -96,8 +96,8 @@ function app.openDialog(message, options, cb, error) end
 
 --- Allow to register menupoints and toolbar buttons. This needs to be called from initUi
 --- 
---- @param opts {menu: string, callback: string, toolbarID: string, mode:integer, accelerator:string} options (`mode`,
----  `toolbarID` and `accelerator` are optional)
+--- @param opts {menu: string, callback: string, toolbarID: string, mode:integer, accelerator:string, parentPath:string}
+---   options (`mode`, `toolbarID`, `accelerator` and `parentPath` are optional)
 --- @return {menuId:integer}
 --- 
 --- Example 1: app.registerUi({["menu"] = "HelloWorld", callback="printMessage", mode=1, accelerator="<Control>a"})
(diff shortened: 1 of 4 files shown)


tobiasBora:
Thanks for this. Just, is there a way to check if we run xournal with this enable or not? (I don't want people running old versions of xournal to be unable to use my plugin just because I use sub-menu)

youssefadly237:
It is fine older versions of xournalpp will just ignore the field as if did not exist and register normally

tobiasBora:
Ok it's great. I was also thinking that it may still be helpful to know if this feature is available to show different things in the menu based on the version. E.g. for old versions I would show
```
app.registerUi({menu="My plugin / Move to next student", callback="…"})
```
while with the new interface I would remove the `My plugin` (since a new submenu is automatically created, so no need to write it twice) and maybe remove the `Move to` that is also redundent to keep only:
```
app.registerUi({menu="Next student", callback="…", parentPath = "Move to"})
```
Maybe a way to allow this without adding too many overhead would be to allow an additional optional field, say `menuIfParentPath` field that would overwrite `menu` on new versions of xournalpp when available, so that I could write the above as:
```
app.registerUi({menu="My plugin / Move to next student", menuIfParentPath="Next Student", callback="…", parentPath = "Move to"})
```

Anyway it's nitpicking, the new behavior is already a great improvement.

bhennion:
I am a bit confused: does this revert the changes in behaviour of #7225 or not? (on top of adding the possibility of longer menu paths)

youssefadly237:
they are more of complementary

#7225 -> Plugin-level submenus
#7323 -> parent path support within a single plugin

bhennion (commented):
The feature is good, but I think the implementation of `getOrCreateParent()` is pretty inefficient. It is not that important as this function is really not critical, but if you want to clean it up, it would be better I think.

My concern is the following: when creating the menu, you go up (from root to leaf), checking first that `/foo` has been created, then `/foo/bar` and then `/foo/bar/subdir`.
Now, if you add several entries in `/foo/bar/subdir`, you will check again `/foo`, `/foo/bar` and then `/foo/bar/subdir`.
Now, because you have your string-to-menu map, it would be faster to go the other way around: you start from a string `/foo/bar/subdir`.
If the submenu `/foo/bar/subdir` already exists, add your new entry there directly. Otherwise, look for `/foo/bar` and if you found it, add a submenu `subdir` and your entry in that submenu, and so on. This is typically done recursively.

bhennion (approved):
LGTM!
@rolandlo What do you think of this?

rolandlo (approved):
LGTM

bhennion:
Alright then! Meging in 24 hours unless an objection is raised.

PR #7333, opened 2026-03-29
Allow selecting objects that extend beyond page boundaries

### LINKED ISSUE #1292: Can't select objects out of boundary
- OS: probably any
 - Version: probably any with the new input system

If you past an object (like a picture) into the journal and it juts out of the boundary, you can't select or delete it anymore. The only way is to overwrite it. 

The new input system did solve the issue of drawing over the boundaries and it would be nice if the same would work for selecting.

zandivx:
As mentioned in #1292 objects extending beyond the page boundary cannot be selected via rectangular or lasso selection tool. This PR introduces the extension of these selections to infinity if the selection touches the page boundary. As this selection still happens on the same page (just with a huge box) this shouldn't affect any neighboring pages and does not open the can of multi-page selection. I did the same for the lasso selection tool with a polygon extending to infinity whenever the lasso touches the boundary. The lasso extension is precomputed.

This should actually be a full fix and close #1292.

### DIFF
--- a/src/core/control/tools/Selector.cpp
+++ b/src/core/control/tools/Selector.cpp
@@ -91,14 +91,18 @@ void RectangularSelector::extendAtPageEdges() {
     const double pageHeight = page->getHeight();
 
     if (pageWidth > 0 && pageHeight > 0) {
-        if (bbox.minX <= THRESHOLD)
+        if (bbox.minX <= THRESHOLD) {
             bbox.minX = -INF;
-        if (bbox.minY <= THRESHOLD)
+        }
+        if (bbox.minY <= THRESHOLD) {
             bbox.minY = -INF;
-        if (bbox.maxX >= pageWidth - THRESHOLD)
+        }
+        if (bbox.maxX >= pageWidth - THRESHOLD) {
             bbox.maxX = INF;
-        if (bbox.maxY >= pageHeight - THRESHOLD)
+        }
+        if (bbox.maxY >= pageHeight - THRESHOLD) {
             bbox.maxY = INF;
+        }
     }
 }
 
@@ -158,18 +162,18 @@ void LassoSelector::extendAtPageEdges() {
         return p.x <= THRESHOLD || p.x >= pageWidth - THRESHOLD || p.y <= THRESHOLD || p.y >= pageHeight - THRESHOLD;
     };
 
+    auto const extendCoordinate = [&](double value, double pageExtent) -> double {
+        if (value <= THRESHOLD) {
+            return -INF;
+        }
+        if (value >= pageExtent - THRESHOLD) {
+            return INF;
+        }
+        return value;
+    };
+
     auto const project = [&](BoundaryPoint const& p) -> BoundaryPoint {
-        double px = p.x;
-        double py = p.y;
-        if (px <= THRESHOLD)
-            px = -INF;
-        if (px >= pageWidth - THRESHOLD)
-            px = INF;
-        if (py <= THRESHOLD)
-            py = -INF;
-        if (py >= pageHeight - THRESHOLD)
-            py = INF;
-        return {px, py};
+        return {extendCoordinate(p.x, pageWidth), extendCoordinate(p.y, pageHeight)};
     };
 
     extendedBoundaryPoints.clear();
--- a/src/core/control/tools/Selector.cpp
+++ b/src/core/control/tools/Selector.cpp
@@ -21,20 +21,23 @@ auto Selector::finalize(PageRef page, bool disableMultilayer, Document* doc) -> 
     this->page = page;
     this->pageWidth = page->getWidth();
     this->pageHeight = page->getHeight();
-    this->selectionDomain = Range();
 
-    auto addToSelectionDomain = [this](const Layer* layer) {
+    const bool useMultiLayer = multiLayer && !disableMultilayer;
+    Range selectionDomain;
+
+    // First pass: compute selectionDomain from all candidate elements
+    auto addToSelectionDomain = [&selectionDomain](const Layer* layer) {
         if (!layer->isVisible()) {
             return;
         }
         for (const auto& e: layer->getElementsView()) {
-            this->selectionDomain = this->selectionDomain.unite(Range(e->boundingRect()));
+            selectionDomain = selectionDomain.unite(Range(e->boundingRect()));
         }
     };
 
     {
         std::shared_lock lock(*doc);
-        if (multiLayer && !disableMultilayer) {
+        if (useMultiLayer) {
             for (const Layer* layer: page->getLayersView()) {
                 addToSelectionDomain(layer);
             }
@@ -43,15 +46,16 @@ auto Selector::finalize(PageRef page, bool disableMultilayer, Document* doc) -> 
         }
     }
 
-    if (this->selectionDomain.empty()) {
-        this->selectionDomain = this->bbox;
+    if (selectionDomain.empty()) {
+        selectionDomain = this->bbox;
     }
 
-    this->tailorToSelectionDomain();
+    this->tailorToSelectionDomain(selectionDomain);
 
+    // Second pass: find selected elements
     size_t layerId = 0;
 
-    if (multiLayer && !disableMultilayer) {
+    if (useMultiLayer) {
         std::shared_lock lock(*doc);
         const auto layers = page->getLayersView();
         for (auto it = layers.rbegin(); it != layers.rend(); it++) {
@@ -106,7 +110,7 @@ RectangularSelector::~RectangularSelector() = default;
 
 auto RectangularSelector::contains(double x, double y) const -> bool { return extendedBbox.contains(x, y); }
 
-void RectangularSelector::tailorToSelectionDomain() {
+void RectangularSelector::tailorToSelectionDomain(const Range& selectionDomain) {
     constexpr double THRESHOLD = 1.0;  // pt
 
     extendedBbox = bbox;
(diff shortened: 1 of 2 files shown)


src/core/control/tools/Selector.cpp
```diff
@@ -19,6 +19,14 @@ Selector::~Selector() = default;
 
 auto Selector::finalize(PageRef page, bool disableMultilayer, Document* doc) -> size_t {
     this->page = page;
+    this->pageWidth = page->getWidth();
+    this->pageHeight = page->getHeight();
+
+    // Let subclasses prepare extended geometry now that page dimensions are known
+    if (auto* lasso = dynamic_cast<LassoSelector*>(this)) {
+        lasso->buildExtendedBoundary();
+    }
```
bhennion: This kinda defeats the purpose of polymorphism. Why not have a virtual function `tailorToPage()` (or whatever) which directly adapts the selector's status to extend to infinity if its points are on the page's edge?

src/core/control/tools/Selector.cpp
```diff
@@ -74,7 +82,30 @@ RectangularSelector::RectangularSelector(double x, double y, bool multiLayer):
...
+    constexpr double INF = 1e9;
+
+    double minX = bbox.minX;
+    double minY = bbox.minY;
+    double maxX = bbox.maxX;
+    double maxY = bbox.maxY;
+
+    if (pageWidth > 0 && pageHeight > 0) {
+        if (minX <= eps)
+            minX = -INF;
```
bhennion: Better to use [`std::numeric_limits<double>::lowest()`](https://en.cppreference.com/w/cpp/types/numeric_limits/lowest.html) and `std::numeric_limits<double>::max()` for those values.

src/core/control/tools/Selector.cpp
```diff
@@ -74,7 +82,30 @@ RectangularSelector::RectangularSelector(double x, double y, bool multiLayer):
...
+auto RectangularSelector::contains(double x, double y) const -> bool {
+    // If the selection rectangle touches a page edge, extend it to infinity
+    // in that direction. This allows selecting objects that extend beyond the page.
+    constexpr double eps = 1.0;
+    constexpr double INF = 1e9;
+
+    double minX = bbox.minX;
+    double minY = bbox.minY;
+    double maxX = bbox.maxX;
+    double maxY = bbox.maxY;
```
bhennion: Why not modify bbox once and for all (e.g. in the implementation of the proposed tailorToPage())?

bhennion (changes requested):
Thnaks for the PR. I think the end-goal is laudable, but the implementation is pretty weird, I think. See below

zandivx:
@bhennion I've reworked the PR based on your feedback:
- Instead of the `dynamic_cast` I added a virtual method `tailorToSelectionDomain`.
- Replacing the constant `1e9` with floating point inf could lead to bugs later on, so instead I now compute a finite `selectionDomain` from the bounding boxes of all elements. The selector geometry is extended only to this domain, not to infinity.
- `RectangularSelector::contains` now just checks against the precomputed `extendedBbox`, matching what `LassoSelector` already did. From what I saw directly extending the `bbox` can lead to issues later on and unnecessary redraw, so I sticked with the separate `extendedBbox` approach.

Let me know what you think of this!

bhennion:
> @bhennion I've reworked the PR based on your feedback:

Thanks!

> * Replacing the constant `1e9` with floating point inf could lead to bugs later on, so instead I now compute a finite `selectionDomain` from the bounding boxes of all elements. The selector geometry is extended only to this domain, not to infinity.

Really? What kind of bugs?
The solution you implemented is quite inefficient. We want to iterate over all the elements as little as we can.

> * `RectangularSelector::contains` now just checks against the precomputed `extendedBbox`, matching what `LassoSelector` already did. From what I saw directly extending the `bbox` can lead to issues later on and unnecessary redraw, so I sticked with the separate `extendedBbox` approach.

Sure. That's fine. Another solution is to move the redraw up in the finalize() function, to before bbox is modified.

zandivx:
> Really? What kind of bugs?

I checked again: you're right, should be no problem, those are just comparisons with `inf`, no arithmetic.

> The solution you implemented is quite inefficient. We want to iterate over all the elements as little as we can.

I didn't feel any delay, but then again I only tested with a minimal example, I'm sure, you know whether this could be problematic, so I reverted it back to the first idea, just with `inf` instead of `1e9`.

> Another solution is to move the redraw up in the finalize() function, to before bbox is modified.

Did that too. Now, the redraw happens at the start of `Selector::finalize` and after that, the original `bbox` is extended, removing the need for the `extendedBbox`.

src/core/control/tools/Selector.cpp
```diff
@@ -76,6 +82,24 @@ RectangularSelector::~RectangularSelector() = default;
 
 auto RectangularSelector::contains(double x, double y) const -> bool { return bbox.contains(x, y); }
 
+void RectangularSelector::extendAtPageEdges() {
+    constexpr double THRESHOLD = 1.0;  // pt
+    constexpr double INF = std::numeric_limits<double>::infinity();
```
bhennion, line 87: If you want to use `std::numeric_limits<double>::infinity()` and `-std::numeric_limits<double>::infinity()` (as opposed to max/lowest), you need to add a static assertion `static_assert(std::numeric_limits<double>::is_iec559());` as this won't work on systems that don't use iec559.

src/core/control/tools/Selector.cpp
```diff
@@ -76,6 +82,24 @@ RectangularSelector::~RectangularSelector() = default;
...
+    const double pageWidth = page->getWidth();
+    const double pageHeight = page->getHeight();
+
+    if (pageWidth > 0 && pageHeight > 0) {
+        if (bbox.minX <= THRESHOLD)
+            bbox.minX = -INF;
```
bhennion, line 93: Per our coding conventions, every if/for/while/... statement must have enclosing brackets:

```suggestion
        if (bbox.minX <= THRESHOLD) {
            bbox.minX = -INF;
        }
```
Also, you may optimize those 4 tests a tiny bit, bit grouping them two by two with an else statement.

src/core/control/tools/Selector.cpp
```diff
@@ -118,22 +142,88 @@ void LassoSelector::currentPos(double x, double y) {
     }
 }
 
+void LassoSelector::extendAtPageEdges() {
```
bhennion, line 145: In most cases, this function will do nothing. It may be worth adding a check here: if bbox is entirely contained in the page (-THRESHOLD), no need to go through all the points.

src/core/control/tools/Selector.cpp
```diff
@@ -118,22 +142,88 @@ void LassoSelector::currentPos(double x, double y) {
...
+        double px = p.x;
+        double py = p.y;
+        if (px <= THRESHOLD)
+            px = -INF;
+        if (px >= pageWidth - THRESHOLD)
+            px = INF;
+        if (py <= THRESHOLD)
+            py = -INF;
+        if (py >= pageHeight - THRESHOLD)
+            py = INF;
```
bhennion, line 169: Same comment as above.
You may also want to remove the code duplication.

src/core/control/tools/Selector.cpp
```diff
@@ -118,22 +142,88 @@ void LassoSelector::currentPos(double x, double y) {
...
+            extendedBoundaryPoints.push_back(current);
+        } else {
+            // Interior of an edge run: emit only projected
+            extendedBoundaryPoints.push_back(project(current));
+        }
+    }
+
+    // Extend bounding box to infinity where needed
+    for (auto const& p: extendedBoundaryPoints) {
+        bbox.addPoint(p.x, p.y);
```
bhennion, line 206: This should be done in the main loop, to avoid an extra run through the vector.

src/core/control/tools/Selector.h
```diff
@@ -95,4 +102,9 @@ class LassoSelector: public Selector {
     bool contains(double x, double y) const override;
     bool userTapped(double zoom) const override;
     const std::vector<BoundaryPoint>& getBoundary() const override;
+
+    void extendAtPageEdges() override;
+
+private:
+    std::vector<BoundaryPoint> extendedBoundaryPoints;
```
bhennion, line 109: Is this one still necessary? Can't you just replace the content of boundaryPoints?

src/core/control/tools/Selector.cpp
zandivx: Thx! Done

src/core/control/tools/Selector.cpp
zandivx: Done.

src/core/control/tools/Selector.cpp
zandivx: Short-circuit path introduced.

src/core/control/tools/Selector.cpp
zandivx: Extracted into a lambda to be more concise with these checks.

src/core/control/tools/Selector.h
zandivx: It is indeed not, I tested it. A temporary buffer is needed to avoid inserting points into a buffer we are currently looping over. As long as no one assumes these are the original points drawn by the user later on in some code path, we should be fine (I did not find that case).

src/core/control/tools/Selector.cpp
zandivx: Fixed.

zandivx:
I addressed your remarks, they were very helpful, thx a lot!

src/core/control/tools/Selector.cpp
```diff
@@ -12,15 +13,24 @@
 #include "model/XojPage.h"         // for XojPage
 #include "util/safe_casts.h"       // for as_unsigned
 
+static_assert(std::numeric_limits<double>::is_iec559);
```
bhennion, line 17:
```suggestion
// Ensures that `-std::numeric_limits<double>::infinity()` behaves as minus infinity
static_assert(std::numeric_limits<double>::is_iec559);
```

bhennion (commented):
Just a nitpick. Other than that, LGTM.

@rolandlo Do you have a take on the feature?

bhennion (approved):

bhennion (approved):
LGTM!

PR #7336, opened 2026-03-30
Toggle all touchscreens in X11 touch disabler

### LINKED ISSUE #7300: Hand recognition not working (dual touchscreen setup)
### Operating System

Linux

### (Linux only) Distribution

Debian 13

KDE

X.Org

### Installation Method

.deb package from github

### Xournal++ Version

1.3.3

### libgtk Version

3.24.49

### Bug Description

The feature "Internal Hand Recognition" (X11) does not work correctly with my dual touchscreen setup. Touch input is only ignored for the screen that was initially used for stylus input. If I move the Xournal++ window to my second touchscreen display, the deactivated touchscreen device is not updated dynamically.
Only when I start writing with the stylus on my second screen from the beginning does hand recognition work correctly there.

### Expected Behaviour

Internal hand recognition should detect the correct touchscreen device currently used for stylus input, even after moving the application window to another touchscreen.

### Steps to Reproduce

1. Enable internal hand recognition
2. Write with stylus on touchscreen 1
3. Move the window to the second touchscreen
4. Try to write with the stylus -> internal hand recognition does not work on the second screen.

### Additional Context

_No response_

tmoerschell:
Fixes #7300

@Y4NN1K1308, can you test if this fixes the issue for you?

I will try to launch a pipeline build, but I don't know if I can. If that doesn't work, would you be comfortable building from source ?

### DIFF
--- a/src/core/gui/inputdevices/touchdisable/TouchDisableX11.cpp
+++ b/src/core/gui/inputdevices/touchdisable/TouchDisableX11.cpp
@@ -2,6 +2,8 @@
 
 #ifdef X11_ENABLED
 
+#include <vector>  // for vector
+
 #include <X11/Xatom.h>          // for XA_INTEGER
 #include <X11/extensions/XI.h>  // for XI_TOUCHSCREEN
 #include <gdk/gdk.h>            // for gdk_display_get_default
@@ -11,10 +13,10 @@
 TouchDisableX11::TouchDisableX11() = default;
 
 TouchDisableX11::~TouchDisableX11() {
-    if (touchdev) {
-        XCloseDevice(display, touchdev);
-        touchdev = nullptr;
+    for (auto dev: touchdevs) {
+        XCloseDevice(display, dev);
     }
+    touchdevs.clear();
 
     display = nullptr;
 }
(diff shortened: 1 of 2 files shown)


bhennion:
\action create-installers debian

Y4NN1K1308:
I'll give an update as soon as I've bought new batteries for my stylus and can test it. After two years, it just so happens to be dead right now. (But that definitely wasn't responsible for this problem.)

Y4NN1K1308:
Hi, your fix works perfectly for my setup. I didn't encounter any issues during my testing. Switching the touchscreen multiple times no longer affects the integrity of the hand recognition under X11.

src/core/gui/inputdevices/touchdisable/TouchDisableX11.cpp
```diff
@@ -11,10 +13,12 @@
 TouchDisableX11::TouchDisableX11() = default;
 
 TouchDisableX11::~TouchDisableX11() {
-    if (touchdev) {
-        XCloseDevice(display, touchdev);
-        touchdev = nullptr;
+    if (!touchdevs.empty()) {
```
bhennion: This test is not necessary: if it is empty, the loop is empty

bhennion:
Thanks for the fix!

PR #7339, opened 2026-03-31
add atomic library

### LINKED ISSUE #7214: Build failure on powerpc; missing -latomic link flag
### Operating System

Linux

### (Linux only) Distribution

Debian GNU/Linux

Gnome

Wayland

### Installation Method

wajig install

### Xournal++ Version

1.3.3

### libgtk Version

3.24.51

### Bug Description

Build failure on Debian autobuilders for arch powerpc,
https://buildd.debian.org/status/fetch.php?pkg=xournalpp&arch=powerpc&ver=1.3.3-1&stamp=1772461552&raw=0

The error is:
```
/usr/bin/powerpc-linux-gnu-ld.bfd: core/libxournalpp-core.a(Scheduler.cpp.o): undefined reference to symbol '__atomic_load_8@@LIBATOMIC_1.0'
/usr/bin/powerpc-linux-gnu-ld.bfd: /usr/lib/powerpc-linux-gnu/libatomic.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
This is characteristic of a missing `-latomic`. Most but not all platforms will compile okay without the option it as of today, but the right™ thing is to use it whenever atomic operations are in use.

### Expected Behaviour

compiles without error on all platforms

### Steps to Reproduce

compile on powerpc running debian unstable

### Additional Context

powerpc is not a "release architecture" so that won't hold anything up. But there's no harm adding `-latomic` and it might avoid more serious problems in the future.

archisman-panigrahi:
This will likely fix #7214.
I don't have powerpc hardware to test. @barak Any thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,7 +441,7 @@ set(CPACK_DEBIAN_PACKAGE_RELEASE 2)
 set(CPACK_DEBIAN_PACKAGE_VERSION "${CPACK_PROJECT_VERSION}")
 set(CPACK_DEBIAN_PACKAGE_SECTION "graphics")
 set(CPACK_DEBIAN_PACKAGE_DEPENDS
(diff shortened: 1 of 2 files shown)


archisman-panigrahi:
\action create-installers debian (turns out, someone with admin privilege needs to run this)

PR #7358, opened 2026-04-04
Fix macos environment setup

rolandlo:
- Set environment variables on MacOS only for app bundles
- Prepend to XDG_DATA_DIRS and XDG_CONFIG_DIRS instead of overwriting
- Prepend to LUA_PATH and LUA_CPATH (when built using jhbuild) instead of overwriting
- Use g_getenv and g_setenv consistently
- Output most relevant environment variables when run through G_MESSAGES_DEBUG=xopp

The following methods of running Xournal++ on MacOS have been tested to work
1) Opening the Xournal++.app bundle either from Finder or using the open command in the terminal
2) Running the binary in Xournal++.app/Contents/MacOS directly from the terminal
3) Compiling Gtk using jhbuild, building and installing Xournal++ in $JHBUILD_PREFIX and running it
   using jhbuild run xournalpp (in order to set up the environment)
4) Compiling Xournal++ using Homebrew, installing it anywhere and running it from that folder
5) Compiling Xournal++ using MacPorts, installing it anywhere and running it from that folder

Method 4) only worked previously when installing into $(brew --prefix), which was inconvenient

The DYLD_LIBRARY_PATH is only set for method 1) through Info.plist. For the other methods it may need
to be set for some use cases, e.g. for plugins using LuaGObject.
The LUA_PATH and LUA_CPATH must be set when Lua (or Lua modules) are installed in non-standard locations

### DIFF
--- a/src/exe/osx/setup-env.cpp
+++ b/src/exe/osx/setup-env.cpp
@@ -10,60 +10,95 @@
 
 #include "filesystem.h"
 
+namespace {
+void prependPathToEnvVar(const char* name, fs::path const& path) {
+    const char* currentVal = g_getenv(name);
+    std::string newVar;
+    if (currentVal == nullptr || strlen(currentVal) == 0) {
+        newVar = path.string();
+    } else {
+        newVar = path.string() + ":" + currentVal;
+    }
+    g_setenv(name, newVar.c_str(), 1);
+}
+
+void prependPathToLua(const char* name, const std::string& path) {
+    const char* currentVal = g_getenv(name);
+    std::string newVar;
+    if (currentVal == nullptr || strlen(currentVal) == 0) {
+        newVar = path + ";;";
+    } else {
+        newVar = path + ";" + currentVal;
+    }
+    g_setenv(name, newVar.c_str(), true);
+}
+}  // end namespace
+
 void setupEnvironment() {
     /**
      * see https://gitlab.gnome.org/GNOME/gtk-mac-bundler/-/blob/master/examples/gtk3-launcher.sh
      * and https://gitlab.gnome.org/GNOME/gtk-mac-bundler/-/issues/12
      */
-    auto base = Util::getExePath().parent_path();  // Xournal++.app/Contents or $HOME/gtk/inst
+    auto base = Util::getExePath().parent_path();  // Xournal++.app/Contents or $HOME/gtk/inst or e.g. build/inst (via
+                                                   // Homebrew or MacPorts)
 
-    if (fs::exists(base / "Resources")) {  // app-bundle
-        base = base / "Resources";
-        // Set poppler data directory for (relocatable) app bundles. Otherwise a compile time value is used.
-        auto popplerDataDir = base / "share" / "poppler";
-        poppler::set_data_dir(popplerDataDir.string().c_str());
-    }  // Now base is Xournal++.app/Contents/Resources or $HOME/gtk/inst
+    bool isAppBundle = fs::exists(base / "Resources");
+    bool underJHBuild = g_strcmp0(g_getenv("UNDER_JHBUILD"), "true") == 0;
+
+    if (isAppBundle) {
+        base = base / "Resources";  // Now base is Xournal++.app/Contents/Resources
+    }
 
     auto libPath = base / "lib";
     auto dataPath = base / "share";
-    auto xdgPath = base / "etc" / "xdg";
 
-    auto imModuleFile = libPath / "gtk-3.0" / "3.0.0" / "immodules.cache";
-    auto pixbufModuleFile = libPath / "gdk-pixbuf-2.0" / "2.10.0" / "loaders.cache";
+    if (isAppBundle) {
+        // Set poppler data directory for (relocatable) app bundles. Otherwise a compile time value is used.
+        auto popplerDataDir = dataPath / "poppler";
+        poppler::set_data_dir(popplerDataDir.string().c_str());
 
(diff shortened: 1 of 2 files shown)


rolandlo:

rolandlo:
I have checked with MacPorts and Nix as well. In the case of MacPorts it worked fine out of the box (just need to set the `DYLD_LIBRARY_PATH` in the terimnal when using plugins using LuaGObject). With Nix the official package seems broken. There were multiple issues to address:
- Added `/nix/store/<hash>-adwaita-icon-theme-49.0/share` to `XDG_DATA_DIRS` so that `image-missing.svg` was found
- There are two `loaders.cache` files, one for `gdk-pixbuf` and one for `librsvg` and neither of them contained an entry for the `svg` loader (apparantly becuase it is a `.dylib`, not a `.so` like the other loaders). I copied one of the `loaders.cache` files to the build folder and added an entry for the svg loader. I then let `GDK_PIXBUF_MODULE_FILE` refer to the `loaders.cache` file in the build folder.
- Now the svg loader still couldn't be loaded, because it is referred to by `"@rpath/librsvg-2.2.dylib"`. I added the path `/nix/store/<hash>-librsvg-2.61.4/lib` to `DYLD_FALLBACK_LIBRARY_PATH` to make it work.

It seems we can't do better with regards to Nix. The nix package (and librsvg maybe as well) should be fixed instead.

For future reference: I built the nix package by adding a `shell.nix` file to the root of the xournalpp repository with the following content:

```nix
{ pkgs ? import <nixpkgs> {} }:

pkgs.xournalpp.overrideAttrs (old: {
  nativeBuildInputs = old.nativeBuildInputs ++
  [
    pkgs.git
    pkgs.ninja
  ];
})
```
and then running `nix-shell` to get into a build environment for Xournal++. I had to use `-DENABLE_CPPTRACE=off` in `cmake`, since the FetchContent thing is not what Nix allows.

bhennion:
> since the FetchContent thing is not what Nix allows.

I've always been confused about this. How do those platforms (Nix, Launchpad, others?) handle fetching secondary sources? Many projects out there use external sources, including GLib for instance (see https://gitlab.gnome.org/GNOME/glib/-/tree/main/subprojects?ref_type=heads).

rolandlo:
> > since the FetchContent thing is not what Nix allows.
> 
> I've always been confused about this. How do those platforms (Nix, Launchpad, others?) handle fetching secondary sources? Many projects out there use external sources, including GLib for instance (see https://gitlab.gnome.org/GNOME/glib/-/tree/main/subprojects?ref_type=heads).

GLib uses external sources as Git submodules. That means the sources are already there if the repository gets cloned (with the `--recurse-submodules` option). We could do that as well with the `cpptrace` library I think. What we currently do is using `FetchContent`. So the `cpptrace` sources are only fetched when we configure by running `cmake ..` This means that the sources are not available before the build process starts and one needs network access during build. For the Flatpak this is not permitted and it seems it is the same thing with Launchpad recipes and Nix.

For the flatpak we fixed that by specifying the sources (and destination folder) in the manifest 
```term 
- type: git
        url: https://github.com/jeremy-rifkin/cpptrace.git
        commit: 3db8da80111171c219ab5839905771386bee06b3
        tag: v1.0.4
        dest: cpptrace-src
```
and pointing FetchContent to that folder using the cmake option 
```term
-DFETCHCONTENT_SOURCE_DIR_CPPTRACE=/run/build/xournalpp/cpptrace-src
```       
Using that trick the sources are all available before the build process starts and FetchContent doesn't need network access.

src/exe/osx/setup-env.cpp
```diff
@@ -10,60 +10,95 @@
...
+namespace {
+void prependPathToEnvVar(const char* name, fs::path const& path) {
+    const char* currentVal = g_getenv(name);
+    std::string newVar;
+    if (currentVal == nullptr || strlen(currentVal) == 0) {
+        newVar = path.string();
+    } else {
+        newVar = path.string() + ":" + currentVal;
+    }
+    g_setenv(name, newVar.c_str(), 1);
```
bhennion, line 22: From the g_setenv documentation, the encoding seems to be the same as for GFilename. I would use Util::toGFilename(path) instead of just path.string().
There are other instances of this below.

src/exe/osx/setup-env.cpp
```diff
@@ -10,60 +10,95 @@
...
+    auto base = Util::getExePath().parent_path();  // Xournal++.app/Contents or $HOME/gtk/inst or e.g. build/inst (via
+                                                   // Homebrew or MacPorts)
 
-    if (fs::exists(base / "Resources")) {  // app-bundle
-        base = base / "Resources";
-        // Set poppler data directory for (relocatable) app bundles. Otherwise a compile time value is used.
-        auto popplerDataDir = base / "share" / "poppler";
-        poppler::set_data_dir(popplerDataDir.string().c_str());
-    }  // Now base is Xournal++.app/Contents/Resources or $HOME/gtk/inst
+    bool isAppBundle = fs::exists(base / "Resources");
```
bhennion, line 45: Probably better to always wrap filesystem io's in a try/catch (or use the variant with std::error_code). Those are more likely to throw than most functions.

bhennion (commented):
There is the encoding matter. I'm not sure what MacOS uses for encoding these days (especially in China), so it may or may not be an actual issue.
Other than that, LGTM. (but I have no real knowledge of the MacOS env)

bhennion (approved):
LGTM!

rolandlo:
Thanks for the review @bhennion

Issue #7332, opened 2026-03-29
Xournal++ 1.3.4 AppImage can't start on ZorinOS 18 Core


jozsefk9:
### Operating System/Distribution

Zorin OS 18

### (Linux only) Desktop Environment

zorin wayland

### (Linux Only) Display Server

Wayland

### Installation Method

AppImage

### Xournal++ Version

xournalpp 1.3.4 (008db584 from xournalpp/HEAD)

### libgtk Version

3.24.33

### Steps to Reproduce

1. Download the AppImage from Github
2. Make it executable
3. Run it in Terminal

### Crash log

<img width="1089" height="387" alt="Image" src="https://github.com/user-attachments/assets/5d3090f6-c767-48d4-ad78-7e0838306a70" />

Executing "/tmp/.mount_xournalaPEmj/usr/bin/xournalpp --disable-audio"
Xournal++ started with PID: 77229
Exited with error status: 127
    Date: Sun Mar 29 16:31:57 2026

xournalpp 1.3.4 (008db584 from xournalpp/HEAD)
├──libgtk: 3.24.33
├──glib: 2.72.4
├──cairo:  1.16.0
└──OS info: Zorin OS 18
  (The GDK backend is probably printed out below)

*** Output: ***

/tmp/.mount_xournalaPEmj/usr/bin/xournalpp: error while loading shared libraries: libjack.so.0: cannot open shared object file: No such file or directory

### Additional Context

It seems it's looking for libjack.so.0 even if I try to start it with `--disable-audio` option.

bhennion:
Not really a bug. Just install jack (I do not know the package name for Zorin).

archisman-panigrahi:
Zorin is based on ubuntu. Try `sudo apt install pipewire-jack libjack0`

jozsefk9:
You are right. I tried it on Void Linux with libjack pre installed and Xournal works fine. Thank you.

Issue #7271, opened 2026-03-15
crash on startup due to pulseaudio issue


ganeshtv5557-hue:
### Operating System/Distribution

GNOME 49 (Flatpak runtime)

### (Linux only) Desktop Environment

zorin-os-lite x11

### (Linux Only) Display Server

None

### Installation Method

both official and the software store that came with zorin

### Xournal++ Version

xournalpp 1.3.3 (6b7c1db2 from xournalpp/HEAD)

### libgtk Version

3.24.51

### Steps to Reproduce

1. download from the official release
2. Installation completes but while launching below issue occurs

### Crash log

Executing "/app/bin/xournalpp"
Xournal++ started with PID: 3
Crashed with signal: 11
    Date: Sun Mar 15 13:39:03 2026

xournalpp 1.3.3 (6b7c1db2 from xournalpp/HEAD)
├──libgtk: 3.24.51
├──glib: 2.86.3
├──cairo:  1.18.4
└──OS info: GNOME 49 (Flatpak runtime)
  (The GDK backend is probably printed out below)

*** Output: ***

(com.github.xournalpp.xournalpp:3): xopp-DEBUG: 13:38:59.064: TEXTDOMAINDIR = (null), Platform-specific locale dir = /app/share/locale, chosen directory = /app/share/locale

(com.github.xournalpp.xournalpp:3): xopp-WARNING **: 13:38:59.118: Settings file /home/ganesh/.var/app/com.github.xournalpp.xournalpp/config/xournalpp/settings.xml does not exist. Regenerating. 
ALSA lib ../../../src/pcm/pcm_direct.c:2048:(snd1_pcm_direct_parse_open_conf) The field ipc_gid must be a valid group (create group audio)
ALSA lib ../../../src/pcm/pcm_direct.c:2048:(snd1_pcm_direct_parse_open_conf) The field ipc_gid must be a valid group (create group audio)
ALSA lib ../../../src/pcm/pcm.c:2722:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib ../../../src/pcm/pcm.c:2722:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib ../../../src/pcm/pcm.c:2722:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
ALSA lib ../../../src/pcm/pcm_route.c:878:(find_matching_chmap) Found no matching channel map
[...]

Issue #5314, opened 2023-11-03
Appimage uses XWayland instead of native Wayland


ondras12345:
### Operating System

Linux

### (Linux only) Distribution

Ubuntu 22.04.3 LTS

### (Linux only) Desktop Environment

GNOME

### (Linux Only) Display Server

Wayland

### Installation Method

Appimage

### Xournal++ Version

1.2.2 (09ae3333)

### libgtk Version

3.24.20

### Bug Description

When I run `xournalpp-1.2.2-x86_64.AppImage`, `xeyes` follow my cursor when over the xournalpp window, indicating that it is using XWayland instead of running natively in Wayland.
When I run an older version installed using `apt`, `xeyes` do not move.
```
$ /usr/bin/xournalpp --version
Xournal++ 1.1.1
└──libgtk: 3.24.33
```
Version 1.1.1 appimage uses XWayland (`xeyes` move):
```
$ ~/bin/xournalpp-1.1.1-x86_64.AppImage --version
Xournal++ 1.1.1
└──libgtk: 3.22.30
```

### Expected Behaviour

Appimage version should run natively in Wayland.

### Steps to Reproduce

1. Download appimage release from github.
2. `chmod +x` and run it in a Wayland session.
3. Run `xeyes`.
4. See `xeyes` follow your cursor when over the Xournal++ window. This does not happen with native Wayland windows.

### Additional Context

System package:
```
$ /usr/bin/xournalpp
** Message: 13:51:56.829: TEXTDOMAINDIR = (null), Platform-specific locale dir = /usr/share/xournalpp/../locale, chosen directory = /usr/share/xournalpp/../locale
ALSA lib pcm.c:2664:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2664:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2664:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
ALSA lib pcm_route.c:877:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:877:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:877:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_route.c:877:(find_matching_chmap) Found no matching channel map
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
[...]

PR #7361, opened 2026-04-05
Implement runtime-dependencies: audio

bhennion:
I wanted to make a proposal to make some of our less central features and dependencies checked and loaded at runtime.

In this PR, I implemented that for the audio support. The audio libraries (portaudio, libsndfile) as well as all the `audio/...` code is dynamically loaded on startup. If a dependency is missing (like in #7332), the app still runs (issueing a warning) without audio support.
It will mostly be useful for flatpak/snap/appimage packages.

In the long run, I think other features could be made "runtime-optional" like that, so that package managers downstream can put in optional deps (I'm thinking of gtksourceview, lua and qpdf).
Of course, for Windows and MacOS, as we ship all the dependencies with the app, it should be seldom useful.

In practice, a shared library called libxournalpp-audio is created and loaded on the fly. If it is loaded successfully (meaning that its deps portaudio and libsndfile are there), then audio is on. Otherwise, it is off.

Note that when compiling, the dependencies need to be present or the configuring will fail (unless ENABLE_AUDIO is set to off).

I tested it on Linux (removing by hand the portaudio library before running the app...). I did my best to ensure the lib libxournalpp-audio is shipped with the installers on Windows and MacOS, but this needs to be tested.

@rolandlo What do you think of this idea?


bhennion:
\action create-installers all

rolandlo:
> @rolandlo What do you think of this idea?

I like the idea, haven't thought through the packaging details yet. Why do you say it will be particularly useful for the flatpak (amidst snap and Appimage)? Will runtime dependencies be delivered as extensions (like the texlive extension) that users are free to install or not (and by default are not installed)?

bhennion:
> Why do you say it will be particularly useful for the flatpak (amidst snap and Appimage)? Will runtime dependencies be delivered as extensions (like the texlive extension) that users are free to install or not (and by default are not installed)?

No no, I didn't think of it that thoroughly. It's just that we do not know on what system they'll be running. In the case of #7332, the OS does not seem to be relying on portaudio. We do not ship all of portaudio's deps in the AppImage apparently (and I don't know if it would have worked anyway).

rolandlo:
> > Why do you say it will be particularly useful for the flatpak (amidst snap and Appimage)? Will runtime dependencies be delivered as extensions (like the texlive extension) that users are free to install or not (and by default are not installed)?
> 
> No no, I didn't think of it that thoroughly. It's just that we do not know on what system they'll be running. In the case of #7332, the OS does not seem to be relying on portaudio. We do not ship all of portaudio's deps in the AppImage apparently (and I don't know if it would have worked anyway).

The AppImage is very different than flatpak and snap in that regard. For the latter two we ship everything that is not already in the runtime environment.

By the way I'm not optimistic about the future of AppImage. It seems to be stuck in the past, not running with native wayland. See #5314, which is still relevant.

bhennion:
> The AppImage is very different than flatpak and snap in that regard. For the latter two we ship everything that is not already in the runtime environment.

I didn't know there was a big difference here. Thanks!
Still, if the underlying OS does not rely on jack for its audio support, the flatpak also fails, but with a different error: see #7271/#6885.

rolandlo:
> > The AppImage is very different than flatpak and snap in that regard. For the latter two we ship everything that is not already in the runtime environment.
> 
> I didn't know there was a big difference here. Thanks! Still, if the underlying OS does not rely on jack for its audio support, the flatpak also fails, but with a different error: see #7271.

Ok, in case of audio we have in the flatpak manifest the finish-arg 
```
--socket=pulseaudio - Access to PulseAudio. It includes sound input (mic), sound output/playback, MIDI and ALSA sound devices in /dev/snd. This permission can be sensitive in certain situations.
```

PR #7384, opened 2026-04-14
Reduce PDF cache retention for off-screen pages

ChasingZenith:
## Summary

This change reduces memory growth when scrolling through PDF-backed pages, especially at very high zoom levels.

When writing in narrow PDF margins at around 400% zoom, scrolling through pages could cause memory usage to grow very aggressively, sometimes beyond **10 GB**. The main issue was that cleaning up page view buffers did not aggressively evict the corresponding PDF render cache entries, so large high-zoom PDF surfaces could accumulate in memory.

## What changed

- Reworked `XournalView::cleanupBufferCache()` to:
  - first collect the set of PDF pages that must be retained because they are visible or inside the preload range
  - then delete off-screen page view buffers
  - evict PDF cache entries for off-screen pages whose PDF backgrounds are no longer retained
- Added ~~`PdfCache::evict(size_t pdfPageNo)` to remove cached renderings for a specific PDF page~~ `PdfCache::evictAllExcept(const std::unordered_set<size_t>& retainedPdfPages)` to prune cached renderings in one pass based on the retained PDF pages
- Changed `PdfCache::cache()` to always replace existing entries for the same PDF page instead of allowing duplicates
- ~~Guarded cache setting updates with the existing render mutex~~

This may also fix two off-by-one or indexing issue in the preload range logic.

- Handled the `maxSize == 0` case explicitly by rendering directly without storing a cache entry
- Fixed `preloadPageBounds()` usage, since the cleanup path previously mixed inclusive checks with `i + 1`, while other callers use half-open zero-based bounds

## Results

After this change, only visible or preloaded PDF pages keep their cached renderings, so memory remains much more tightly bounded during scrolling.

Manual results from testing this scenario:

- Before: memory usage could exceed **10 GB** while scrolling at around 400% zoom
- After: under the same workflow, memory usage rarely exceeds **1.8 GB**, and I do not notice any lag despite the reduced caching.

### DIFF
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -11,6 +11,7 @@
 
 #include "control/settings/Settings.h"  // for Settings
 #include "pdf/base/XojPdfDocument.h"    // for XojPdfDocument
+#include "util/Assert.h"                // for xoj_assert
 #include "util/Range.h"                 // for Range
 #include "util/i18n.h"                  // for _
 #include "util/safe_casts.h"            // for as_unsigned
@@ -72,9 +73,8 @@ void PdfCache::evictAllExcept(const std::unordered_set<size_t>& retainedPdfPages
     std::lock_guard<std::mutex> lock(this->renderMutex);
 
     for (auto& entry: this->data) {
-        if (!entry || !entry->popplerPage) {
-            continue;
-        }
+        xoj_assert(entry);
+        xoj_assert(entry->popplerPage);
 
         const size_t pdfPageNo = static_cast<size_t>(entry->popplerPage->getPageId());
         if (retainedPdfPages.find(pdfPageNo) == retainedPdfPages.end()) {
@@ -96,13 +96,15 @@ auto PdfCache::lookup(size_t pdfPageNo) const -> const PdfCacheEntry* {
 }
 
 auto PdfCache::cache(XojPdfPageSPtr popplerPage, xoj::view::Mask&& buffer) -> const PdfCacheEntry* {
-    const auto pageId = popplerPage ? popplerPage->getPageId() : -1;
+    xoj_assert(popplerPage);
+    const auto pageId = popplerPage->getPageId();
 
-    this->data.erase(std::remove_if(this->data.begin(), this->data.end(),
-                                    [pageId](const auto& entry) {
-                                        return entry && entry->popplerPage && entry->popplerPage->getPageId() == pageId;
-                                    }),
-                     this->data.end());
+    auto existingIt = std::find_if(this->data.begin(), this->data.end(), [pageId](const auto& entry) {
+        return entry->popplerPage->getPageId() == pageId;
+    });
+    if (existingIt != this->data.end()) {
+        this->data.erase(existingIt);
+    }
 
     if (this->maxSize == 0) {
         this->data.clear();
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -94,6 +94,7 @@ auto PdfCache::lookup(size_t pdfPageNo) const -> const PdfCacheEntry* {
 }
 
 auto PdfCache::cache(XojPdfPageSPtr popplerPage, xoj::view::Mask&& buffer) -> const PdfCacheEntry* {
+    xoj_assert(this->maxSize > 0);
     xoj_assert(popplerPage);
     const auto pageId = popplerPage->getPageId();
 
@@ -101,11 +102,6 @@ auto PdfCache::cache(XojPdfPageSPtr popplerPage, xoj::view::Mask&& buffer) -> co
                                    [pageId](const auto& entry) { return entry->popplerPage->getPageId() == pageId; });
     if (existingIt != this->data.end()) {
         this->data.erase(existingIt);
-    }
-
-    if (this->maxSize == 0) {
-        this->data.clear();
-        return nullptr;
     }
 
     if (this->data.size() >= this->maxSize) {
@@ -149,9 +145,9 @@ void PdfCache::render(cairo_t* cr, size_t pdfPageNo, double zoom, double pageWid
         if (this->maxSize == 0) {
             buffer.paintTo(cr);
             return;
-        } else {
-            cacheResult = cache(popplerPage, std::move(buffer));
         }
+
+        cacheResult = cache(popplerPage, std::move(buffer));
     }
 
     cacheResult->buffer.paintTo(cr);
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -162,4 +162,3 @@ void PdfCache::renderMissingPdfPage(cairo_t* cr, double pageWidth, double pageHe
     cairo_move_to(cr, pageWidth / 2 - extents.width / 2, pageHeight / 2 - extents.height / 2);
     cairo_text_path(cr, strMissing.c_str());
 }
-
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -38,12 +38,10 @@ public:
     xoj::view::Mask buffer;
 };
 
-namespace {
-double getPercentZoomChange(double oldZoom, double newZoom) {
+static double getPercentZoomChange(double oldZoom, double newZoom) {
     double averagedZoom = (oldZoom + newZoom) / 2.0;
     return std::abs(oldZoom - newZoom) * 100.0 / averagedZoom;
 }
-}  // namespace
 
 PdfCache::PdfCache(const XojPdfDocument& doc, Settings* settings): pdfDocument(doc) { updateSettings(settings); }
 
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -162,3 +162,4 @@ void PdfCache::renderMissingPdfPage(cairo_t* cr, double pageWidth, double pageHe
     cairo_move_to(cr, pageWidth / 2 - extents.width / 2, pageHeight / 2 - extents.height / 2);
     cairo_text_path(cr, strMissing.c_str());
 }
+
--- a/src/core/control/PdfCache.cpp
+++ b/src/core/control/PdfCache.cpp
@@ -47,13 +47,9 @@ PdfCache::PdfCache(const XojPdfDocument& doc, Settings* settings): pdfDocument(d
 
 PdfCache::~PdfCache() = default;
 
-void PdfCache::setRefreshThreshold(double threshold) {
-    std::lock_guard<std::mutex> lock(this->renderMutex);
-    this->zoomRefreshThreshold = threshold;
-}
+void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
 
 void PdfCache::setMaxSize(size_t newSize) {
-    std::lock_guard<std::mutex> lock(this->renderMutex);
     this->maxSize = newSize;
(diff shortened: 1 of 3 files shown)


src/core/gui/XournalView.cpp
```diff
@@ -121,13 +126,43 @@ auto XournalView::cleanupBufferCache() -> void {
...
+        const size_t pdfPageNo = page->getPage()->getPdfPageNr();
+
+        if (shouldRetain) {
+            if (pdfPageNo != npos) {
+                retainedPdfPages.insert(pdfPageNo);
+            }
+        }
+    }
+
+    for (size_t i = 0; i < this->viewPages.size(); i++) {
```
bhennion: Looping a second time on all the pages is no very efficient here. The number of pages could be quite large.
It can be avoided by 

1. Calling page->deleteViewBuffer() in the first loop
2. Passing the set of "retained pages" directly to the PdfCache
3. In PdfCache, instead of repeatedly calling `std::erase()/std::remove()` (via calls to `evict()`), run through the cache once to destroy the cache entry (set the std::unique_ptr to nullptr) if the corresponding page is not in the set of retained pages
4. Call std::erase/remove only once on the PdfCache deque to remove the nullptr's.

bhennion (changes requested):
Thanks for the PR! The issue it fixes was bad.

I find there are a several unecessary loops that could/should be removed, so this PR doesn't otherwise impact performances. See comment/suggestion below.

bhennion:
Also, if the PR stays small, I think it can be targetted to our release-1.3 branch, so the changes appear in 1.3.5 (instead of waiting for 1.4.0).

src/core/gui/XournalView.cpp
ChasingZenith: Thanks for the review, this was a good suggestion.
I applied this suggestion. `cleanupBufferCache()` now deletes page view buffers during the first (now the only) `for` loop. In `PdfCache`, cleanup is now done in one scan by resetting non-retained entries to `nullptr`, followed by a single `erase/remove` on the deque. I also dropped the old per-page `evict()` path since it is no longer needed.

ChasingZenith:
Hi, @bhennion . I hope the follow-up changes meet your expectations. I’d be excited to see this merged!

ChasingZenith:
Now, this PR is targeted to release-1.3 branch.

src/core/control/PdfCache.cpp
```diff
@@ -57,6 +68,23 @@ void PdfCache::updateSettings(Settings* settings) {
     }
 }
 
+void PdfCache::evictAllExcept(const std::unordered_set<size_t>& retainedPdfPages) {
+    std::lock_guard<std::mutex> lock(this->renderMutex);
+
+    for (auto& entry: this->data) {
+        if (!entry || !entry->popplerPage) {
+            continue;
+        }
```
bhennion: Is this ever supposed to happen? If not, better to assert those (using `xoj_assert()`) rather than testing.

src/core/gui/XournalView.cpp
```diff
@@ -120,14 +125,29 @@ auto XournalView::cleanupBufferCache() -> void {
...
     for (size_t i = 0; i < this->viewPages.size(); i++) {
         auto&& page = this->viewPages[i];
-        const size_t pageNum = i + 1;
-        const bool isPreload = pagesLower <= pageNum && pageNum <= pagesUpper;
-        if (!isPreload && !page->isVisible() && page->hasBuffer()) {
+        const bool isPreload = pagesLower <= i && i < pagesUpper;
+        const bool shouldRetain = isPreload || page->isVisible();
+        const size_t pdfPageNo = page->getPage()->getPdfPageNr();
+
+        if (shouldRetain) {
```
bhennion: Not needed otherwise
```suggestion

        if (shouldRetain) {
            const size_t pdfPageNo = page->getPage()->getPdfPageNr();
```

src/core/control/PdfCache.cpp
```diff
@@ -68,8 +96,21 @@ auto PdfCache::lookup(size_t pdfPageNo) const -> const PdfCacheEntry* {
 }
 
 auto PdfCache::cache(XojPdfPageSPtr popplerPage, xoj::view::Mask&& buffer) -> const PdfCacheEntry* {
-    if (this->data.size() > this->maxSize) {
-        this->data.resize(this->maxSize);
+    const auto pageId = popplerPage ? popplerPage->getPageId() : -1;
```
bhennion: Here also: `popplerPage` should never be null. It should be asserted rather than tested.

src/core/control/PdfCache.cpp
```diff
@@ -68,8 +96,21 @@ auto PdfCache::lookup(size_t pdfPageNo) const -> const PdfCacheEntry* {
...
+
+    this->data.erase(std::remove_if(this->data.begin(), this->data.end(),
+                                    [pageId](const auto& entry) {
+                                        return entry && entry->popplerPage && entry->popplerPage->getPageId() == pageId;
+                                    }),
+                     this->data.end());
```
bhennion: This enforces that only one Cache entry can be kept per page. It can thus be optimized using std::find_if instead of std::remove_if: we can stop once the first occurrence is found.

bhennion (commented):
Here is another batch of tiny comments. Mostly nitpicks

ChasingZenith:
Thank you for taking the time to review the code. I’ve made the changes. I’ve used assertions, paid attention to scope and replaced `remove_if` with `find_if`.

src/core/gui/XournalView.cpp
```diff
@@ -120,14 +125,29 @@ auto XournalView::cleanupBufferCache() -> void {
...
+            const size_t pdfPageNo = page->getPage()->getPdfPageNr();
+            if (pdfPageNo != npos) {
+                retainedPdfPages.insert(pdfPageNo);
+            }
+            continue;
+        }
+
+        if (page->hasBuffer()) {
```
bhennion: Better readability
```suggestion
        } else if (page->hasBuffer()) {
```

src/core/control/PdfCache.cpp
```diff
@@ -37,13 +38,24 @@ class PdfCacheEntry {
...
 PdfCache::~PdfCache() = default;
 
-void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
+void PdfCache::setRefreshThreshold(double threshold) {
+    std::lock_guard<std::mutex> lock(this->renderMutex);
+    this->zoomRefreshThreshold = threshold;
+}
 
 void PdfCache::setMaxSize(size_t newSize) {
+    std::lock_guard<std::mutex> lock(this->renderMutex);
```
bhennion: Unnecessary: this function is only called upon construction of the PdfCache instance
```suggestion
```

src/core/control/PdfCache.cpp
```diff
@@ -37,13 +38,24 @@ class PdfCacheEntry {
...
+
 PdfCache::PdfCache(const XojPdfDocument& doc, Settings* settings): pdfDocument(doc) { updateSettings(settings); }
 
 PdfCache::~PdfCache() = default;
 
-void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
+void PdfCache::setRefreshThreshold(double threshold) {
+    std::lock_guard<std::mutex> lock(this->renderMutex);
+    this->zoomRefreshThreshold = threshold;
+}
```
bhennion: Same here
```suggestion
void PdfCache::setRefreshThreshold(double threshold) { this->zoomRefreshThreshold = threshold; }
```

src/core/control/PdfCache.cpp
```diff
@@ -37,13 +38,24 @@ class PdfCacheEntry {
     xoj::view::Mask buffer;
 };
 
+namespace {
+double getPercentZoomChange(double oldZoom, double newZoom) {
+    double averagedZoom = (oldZoom + newZoom) / 2.0;
+    return std::abs(oldZoom - newZoom) * 100.0 / averagedZoom;
+}
+}  // namespace
```
bhennion: We tend to favor `static` over nameless namespaces
```suggestion
static double getPercentZoomChange(double oldZoom, double newZoom) {
    double averagedZoom = (oldZoom + newZoom) / 2.0;
    return std::abs(oldZoom - newZoom) * 100.0 / averagedZoom;
}
```

src/core/gui/XournalView.cpp
```diff
@@ -55,10 +56,14 @@ constexpr int SMALL_MOVE_AMOUNT = 1;
 constexpr int LARGE_MOVE_AMOUNT = 10;
 
 std::pair<size_t, size_t> XournalView::preloadPageBounds(size_t page, size_t maxPage) {
+    if (page == npos || page >= maxPage) {
+        return {0, 0};
+    }
+
```
bhennion: Why is that change necessary?

bhennion (commented):
Another batch of comments

bhennion:
Also, the clang-format test is failing. Can you run `git-clang-format` to you changes?

src/core/control/PdfCache.cpp
ChasingZenith: Removed

src/core/gui/XournalView.cpp
ChasingZenith: I'll just removed this changes to keep this PR minimal. It may not be necessary, but `page` is not guaranteed to be a valid page index here. Without this guard, the subsequent calculation becomes problematic:

```cpp
const size_t lower = page > preloadBefore ? page - preloadBefore : 0;
```

Since `size_t` is unsigned, if `page == npos`, arithmetic yields a huge integer instead of a meaningful "invalid" result. Similar guards appear elsewhere in `XournalView.cpp`—see [line 411](https://github.com/ChasingZenith/xournalpp/blob/aac97b0e0834358a08e10b782cf6b98cc34576ca/src/core/gui/XournalView.cpp#L411) for an example.

ChasingZenith:
Hi @bhennion . Thank for your time in the new round review. 

Here is the code. I change the code according to your suggestion. I try to keep this PR only make necessary changes.

bhennion:
Thanks! The code all looks good to me. I tested and it seems to work fine.
There is still a small clang-format issue (just a missing new-line at the end of PdfCache.pdf).
Once this is fixed, I'll squash and merge.

ChasingZenith:
Thank you! I fixed it. learned a lot during the way. Looking forward for the merge.

src/core/control/PdfCache.cpp
```diff
@@ -127,3 +162,4 @@ void PdfCache::renderMissingPdfPage(cairo_t* cr, double pageWidth, double pageHe
     cairo_move_to(cr, pageWidth / 2 - extents.width / 2, pageHeight / 2 - extents.height / 2);
     cairo_text_path(cr, strMissing.c_str());
 }
+
```
bhennion: You seem to have added an extra line instead of a single trailing new-line (ie '\n')
```suggestion
```
It may be your code editor doing that. I don't know.

ChasingZenith: In fact, I am still quite confusing about it.
I will commit your suggestion directly and try.

ChasingZenith: Sorry for bothering. I checked with `cat -A`. I think it should work this time.

bhennion:
It worked. For next time: ideally, you should have run `git-clang-format THE_COMMIT_YOUR_PR_IS_BASED_ON`, so something like `git-clang-format release-1.3`. That would have done the right thing in a single go.

bhennion (approved):

bhennion:

PR #7392, opened 2026-04-15
Add Space modifier to move in-progress shape without resizing

### LINKED ISSUE #7381: Allow moving shapes while creating them (drag reposition during draw)

During the drawing process, if the initial click position is slightly off, the only option is to cancel and redraw the shape. This interrupts workflow and slows down the design process for note-taking.

### Feature Description

Currently, when creating shapes (rectangle, circle, etc.), the user can:
- Hold Shift to preserve aspect ratio
- Hold Ctrl to draw from center

However, it is not possible to reposition the shape while it is being created.

### Proposed Solution
Allow users to move/reposition the shape *while still holding the mouse button*, before releasing it.

Suggested interaction:
- While dragging to create a shape:
  - Press and hold an additional key (e.g., Space like in design tools such as Photoshop/Figma)
  - Move the mouse to reposition the shape
  - Release the key to continue resizing

### Expected Behavior
- Smooth repositioning during shape creation
- Works alongside existing modifiers (Shift, Ctrl)

### Why This Matters
As a frequent user (especially for design and structured notes), this would:
- Improve precision
- Reduce redraw actions
- Align with UX patterns in professional design tools

### Alternative Solutions

_No response_

### Additional Context

This behavior is standard in tools like Adobe Photoshop and Figma, making it intuitive for designers.

I'd be happy to help implement this feature if the maintainers approve the approach.

abdosorour7:
## Summary
- Add a new shape-drawing interaction: hold `Space` while drawing to move the in-progress shape without resizing it.
- Preserve existing modifier behavior: `Ctrl` still resizes from center and `Shift` still constrains aspect ratio.
- Keep geometry frozen during `Space`-move, then resume normal resizing from the new position after releasing `Space`.

## Why
- Improves precision and workflow when placing shapes.
- Avoids cancel/restart when the initial anchor point is slightly off.
- Aligns shape creation behavior with common UX patterns in drawing/design tools.

## Implementation
- Implemented in `BaseShapeHandler` so all shape tools using this base get consistent behavior.
- Added a dedicated translation path for in-progress shapes while `Space` is held.
- Prevented resize recomputation during move mode to ensure movement-only behavior.

## Test Plan
- Start drawing a shape and verify normal resize behavior.
- Hold `Shift` and verify aspect ratio is constrained.
- Hold `Ctrl` and verify resizing is center-based.
- While drawing, hold `Space` and move pointer; verify shape translates without resizing.
- Release `Space`; verify shape keeps moved geometry and can be resized again.
- Verify no regressions across rectangle, ellipse, arrow, ruler, and coordinate system tools.

Closes #7381

### DIFF
--- a/src/core/control/tools/BaseShapeHandler.cpp
+++ b/src/core/control/tools/BaseShapeHandler.cpp
@@ -43,7 +43,25 @@ void BaseShapeHandler::updateShape(bool isAltDown, bool isShiftDown, bool isCont
     viewPool->dispatch(xoj::view::ShapeToolView::FLAG_DIRTY_REGION, repaintRange);
 }
 
+void BaseShapeHandler::translateShape(double dx, double dy) {
+    if (this->shape.empty() || (dx == 0.0 && dy == 0.0)) {
+        return;
+    }
+
+    for (Point& point: this->shape) {
+        point.x += dx;
+        point.y += dy;
+    }
+
+    Range repaintRange = this->lastSnappingRange;
+    this->lastSnappingRange.translate(dx, dy);
+    repaintRange = repaintRange.unite(this->lastSnappingRange);
+    repaintRange.addPadding(0.5 * this->stroke->getWidth());
+    this->viewPool->dispatch(xoj::view::ShapeToolView::FLAG_DIRTY_REGION, repaintRange);
+}
+
 void BaseShapeHandler::cancelStroke() {
+    this->modSpace = false;
     this->shape.clear();
     Range repaintRange = this->lastSnappingRange;
     repaintRange.addPadding(0.5 * this->stroke->getWidth());
@@ -62,9 +80,19 @@ auto BaseShapeHandler::onKeyEvent(const KeyEvent& event, bool pressed) -> bool {
         isControlDown = pressed;
     } else if (event.keyval == GDK_KEY_Alt_L || event.keyval == GDK_KEY_Alt_R) {
         isAltDown = pressed;
+    } else if (event.keyval == GDK_KEY_space) {
+        this->modSpace = pressed;
+        return true;
     } else {
         return false;
     }
+
+    // While moving with Space, keep the current geometry frozen.
+    // Recompute only after Space is released and pointer moves again.
+    if (this->modSpace) {
+        return true;
+    }
+
     this->updateShape(isAltDown, isShiftDown, isControlDown);
 
     return true;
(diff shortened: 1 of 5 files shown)


abdosorour7:
@bhennion could u please review my PR?

bhennion:
The code looks ok to me. The feature also.
The only question is our policy towards AI generated PRs (which hasn't quite been decided yet).
@rolandlo What's your opinion on this PR?

abdosorour7:
Are you referring to Copilot being displayed as one of the collaborators -or the reviewers I think- once the PR is accepted? If so, I just wanted to clarify that the PR itself was not AI-generated.

Copilot was enabled for this PR only for the review, either automatically or possibly because I enabled it accidentally can't remember. But I had actually disabled this feature a while ago, so it shouldn't happen again.

If that's the concern, I can simply delete the current PR and create it again without involving Copilot at all. And if there's anything in the code itself that you'd like me to change or revise, I'm more than happy to take care of that as well.

And thanks for taking the time to review it!

bhennion:
> If that's the concern, I can simply delete the current PR and create it again without involving Copilot at all.

Yes, that may be the easiest. Our AI policy is not settled yet and excluding copilot might be in there at some point.

> And if there's anything in the code itself that you'd like me to change or revise, I'm more than happy to take care of that as well.

No, I think the code is good. I'll just test it when I have time

> And thanks for taking the time to review it!

No problem!

bhennion:
It seems to be working alright. 
There are sometimes interactions between this feature and the `Space` shortcut (which scrolls a bit, see `XournalView::onKeyPressEvent()`). I do not understand where this comes from and what execution path leads to those weird interactions.

abdosorour7:
Thanks for pointing that out! I investigated the execution paths causing the unwanted scrolling and pushed a fix:

### What caused the conflict:
1. **Space key auto-repeat after mouse release:**
   When finishing a shape and releasing the mouse button first, `PageView::onButtonReleaseEvent` destroyed `inputHandler`. If Space was still held down for a fraction of a second, the OS sent auto-repeated `key-press-event`s. Since `inputHandler` was null, the event fell through to `XournalView::onKeyPressEvent` and triggered `layout->scrollRelative()`.
2. **Multi-page continuous view mismatch:**
   Key events were only forwarded to `viewPages[getCurrentPage()]`. If a shape was drawn on another visible page, that page's handler never received the Space key event, triggering the scroll fallback.

### What was fixed:
1. **Auto-Repeat Guard (`spaceConsumedByTool`):** Added tracking in `XournalView` so any repeated Space key presses after a shape is finalized are consumed without scrolling until the Space key is physically released.
2. **Active Input Guard:** `XournalView::onKeyPressEvent` now checks `hasActiveInput()` across pages to ensure Space never scrolls while any drawing operation is in progress.
3. **Cross-Page Event Routing:** Forwarded key press/release events to all visible `viewPages` so handlers on any page receive modifier keys.
4. **Pre-drag Sync:** Initialized `modSpace` in `BaseShapeHandler::onButtonPressEvent` from the current key state in case Space is pressed just before mouse down.

Please test it out whenever you get a chance and let me know your feedback! Also, if you'd prefer me to open a clean/fresh PR instead of this one to keep the review history tidy (without copilot), I'd be happy to do that once we confirm the fix works well on your end.

PR #7411, opened 2026-04-23
Feat: Auto reload PDF when it is modified

### LINKED ISSUE #6255: Auto-reload xopp when pdf is updated

My usecase is a teaching lesson with a LaTeX-pdf with two actions:

(1) Add some notes with xournalpp
(2) Update the LaTeX-pdf during the lesson, e.g. change the parameter of a plot requested by the audience

In case of (2) I need to do: 

    File -> Recent Documents -> 1.

which reloads the most recent xopp file.

### Feature Description

It is possible to add an "auto-reload" switch to automatically reload the current xopp file when the pdf-file is updated?

### Alternative Solutions

_No response_

### Additional Context

_No response_

ChasingZenith:
## Summary

Add automatic detection, reloading and flicker-free re-rendering of modified external PDF backgrounds, replacing the previous manual workflow where users had to reload manually. This feature is very helpful when opening a LaTeX generated PDF. Fix #6255 . 

## Key Changes

- **Add `PdfAutoReloadController`** — Periodically polls the external PDF file's last-modified time. Uses a debounce mechanism to wait until the file size is stable before triggering a reload, avoiding partial-write issues. As a sub-controller, it can be started and stopped by `Controller`.
- **Reuse previous code logic to reload and re-render PDF** Add `DOCUMENT_CHANGE_PDF_CONTENT` event type, allowing listeners to handle PDF content refreshes by reusing logic in `XournalView::documentChanged`, `SidebarPreviewBase::documentChanged`, etc.
- **Add three new settings in `Load / Save` Page -- `Auto-Reload PDF` (new)** (`pdfAutoReloadEnabled`, `pdfAutoReloadIntervalMs`, `pdfAutoReloadDebounceMs`)
-**Handling Page resizing on PDF update edge case** — resizes pages without annotation to match updated PDF dimensions. Pages with annotations are left unchanged to avoid content displacement, and the user is warned about skipped pages.

## For Reviewer

To look at the code, a possible way is to start from changes in `src/core/control/Control.cpp` (**~20 lines**) and then the new controller `src/core/control/PdfAutoReloadController.cpp` (**~150 lines**). There are **~20 lines** for reusing the previous code to handle rerendering. 
The other code are  boilerplate for settings UI and show warning dialog for PDF page size changing.

## Screencast

https://github.com/user-attachments/assets/9d739524-9d20-4000-9930-1a596c1e4454

### DIFF
--- a/src/core/control/Control.cpp
+++ b/src/core/control/Control.cpp
@@ -114,6 +114,7 @@
 #include "CrashHandler.h"                    // for emer...
 #include "LatexController.h"                 // for Late...
 #include "PageBackgroundChangeController.h"  // for Page...
+#include "PdfAutoReloadController.h"
 #include "PrintHandler.h"                    // for print
 #include "UndoRedoController.h"              // for Undo...
 #include "config-dev.h"                      // for SETT...
@@ -162,12 +163,14 @@ Control::Control(GApplication* gtkApp, GladeSearchpath* gladeSearchPath, bool di
     this->toolHandler->loadSettings();
     this->initButtonTool();
 
+    this->pageBackgroundChangeController = std::make_unique<PageBackgroundChangeController>(this);
+    this->pdfAutoReloadController = std::make_unique<PdfAutoReloadController>(this);
+
     /**
      * This is needed to update the previews
      */
     this->changeTimout = g_timeout_add_seconds(5, xoj::util::wrap_v<checkChangedDocument>, this);
-
-    this->pageBackgroundChangeController = std::make_unique<PageBackgroundChangeController>(this);
+    this->pdfAutoReloadController->startMonitoring();
 
     this->layerController = new LayerController(this);
     this->layerController->registerListener(this);
@@ -270,6 +273,13 @@ auto Control::checkChangedDocument(Control* control) -> bool {
     // Call again
     return true;
 }
+
+void Control::firePdfContentChanged() {
+    xoj_assert(this->win);
+    this->fireDocumentChanged(DOCUMENT_CHANGE_PDF_CONTENT);
+}
+
+void Control::refreshAfterPdfChange() { this->pdfAutoReloadController->onPdfChanged(); }
 
 void Control::saveSettings() {
     this->toolHandler->saveSettings();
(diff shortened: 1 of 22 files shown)


ChasingZenith:
I've been running this version for several days and everything works as expected.

Would someone be able to take a look at this PR at your convenience?

youssefadly237:
how does it handle page insertion and deletion?

ChasingZenith:
It does not handle page insertion and deletion.

ChasingZenith:
This PR does not intend to modify the behavior when the number of PDF pages increases or decreases. When PDF pages are added, the original behavior remains unchanged: the user should click `Journal -> Append New PDF Pages` to add new pages; when PDF pages are removed, the missing pages are shown as before.

ChasingZenith:
I have just added a commit https://github.com/xournalpp/xournalpp/pull/7411/changes/598b9abe0fddcf407903496e34242071b6051d6d to fix a missing nullptr bug. Although this bug is not directly related to the newly added feature, it has been triggered frequently after the feature was added. Details are stated in the commit message.

bhennion:
I must admit that I am a bit sceptical about the feature. It introduces the user expectation that modifying the background is actually supported, while it really isn't (for instance, the highlighting of a part of the PDF text won't follow if the text is moved). @rolandlo do you have an opinion on the matter?

As an aside: it may be worth merging the bugfix commit https://github.com/xournalpp/xournalpp/commit/598b9abe0fddcf407903496e34242071b6051d6d to the release branch (so it'll be contained in v1.3.5 and won't have to wait until v1.4.0 is released).

ChasingZenith:
The main use case for this feature is exactly when I want to edit a LaTeX document but also need to do some handwriting at the same time – being able to quickly take notes and make drafts. I find it very convenient.

Regarding the concern about highlights not following text changes: I don't know how text highlighting is implemented. My impression is that it works like a normal Highlighter on a selected text region (Maybe I am wrong). It doesn't move when the underlying PDF changes. For my quick note-taking and drafting workflow, I don't really care about the highlighted parts staying aligned.

So from my perspective, this auto-reload feature is still very useful even with that limitation.

I would later raise a separate PR to take care of the bugfix commit.

PR #7413, opened 2026-04-23
fix: adjustments for MSVC STL

Nerixyz:
This fixes compilation when the MSVC STL is used. While compilation with MSVC itself isn't possible due to [C1061](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1061), clang-cl will work fine there.

The main change is that iterators for `std::string_view` are no longer expected to be pointers. In MSVC's STL, the string view iterator is wrapped to be able to use checked iterators. Note that this will also be a problem if a libc++ with `_LIBCPP_ABI_BOUNDED_ITERATORS` is used.

Other than that, two includes for `<string>` were missing.

---

While this allows compilation without MinGW, it isn't enough to run the app. There are two main issues: 
- The gtk3 vcpkg package doesn't include the assets/icons that would otherwise come with [adwaita-icon-theme](https://packages.msys2.org/base/mingw-w64-adwaita-icon-theme) (haven't tested with conan, but I assume it's similar there).
- Both vcpkg and conan don't support compilation of Rust code right now, so they're missing an up-to-date version of librsvg (https://github.com/microsoft/vcpkg/issues/20619, https://github.com/conan-io/conan-center-index/issues/11392#issuecomment-1222129410).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,8 +215,8 @@ check_cxx_source_compiles([[
     int main() {
         const std::string_view sv = "7.381450";
         double val{};
-        auto [ptr, ec] = std::from_chars(sv.begin(), sv.end(), val);
-        return ptr != sv.end() || ec != std::errc{};
+        auto [ptr, ec] = std::from_chars(sv.data(), sv.data() + sv.size(), val);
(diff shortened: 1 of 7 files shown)


bhennion (approved):
Thanks for the PR!

Issue #6861, opened 2025-12-09
Zoom "Fit to screen" resets scroll position


federico-calandra:
### Operating System

Linux

### (Linux only) Distribution

Arch

### (Linux only) Desktop Environment

KDE Plasma

### (Linux Only) Display Server

X.Org

### Installation Method

pacman

### Xournal++ Version

1.2.7  Git commit 01cd11fa

### libgtk Version

3.24.50

### Bug Description

when clicking View > Best fit the zoom is adjusted so the page occupies max screen width, and the page is scrolled up to the beginning of the active page.

### Expected Behaviour

when clicking View > Best fit the zoom is adjusted so the page occupies max screen width, and the page is not scrolled up to the beginning of the active page but it stays in the same vertical position

### Steps to Reproduce

1. Open document
2. Scroll down the page
3. Click View menu
4. Click Best fit

gitalm:
The same happens, if i paste a image from the clipboard under Ubuntu. If i deactivate **Automtic DPI** it works.
The image isn't inserted and it scrolls to the first page.

bhennion:
> The same happens, if i paste a image from the clipboard under Ubuntu. If i deactivate **Automtic DPI** it works. The image isn't inserted and it scrolls to the first page.

You are referring to a different issue #6879 which has been fixed (to appear in v1.3.1)

PR #7427, opened 2026-04-29
Keep the scroll position when selecting 'Fit to Screen'.

TheKingOfLight:
Currently, if "Fit to Screen" or "Presentation Mode" is selected, the scroll position is forced back to the top of the current page. This was reported in #6861.

This behavior makes sense for Presentation Mode (where the page is locked to the display), but creates a annoying jump in Fit to Screen. In "Fit to Screen" mode it scales based on the page width, but still allows vertical scrolling.

I don't see the reason to force the view to the top of the page when the width was adjusted to fit the screen. I believe the same "reset to top" logic may have been applied to both modes by mistake.

This change ensures that "Fit to Screen" keeps the current vertical position instead of jumping back to the page start when the mode is toggled, like a normal zoom.

Please excuse me if this is the intended behaviour. If so, I would like to hear the reasoning behind it.

### DIFF
--- a/src/core/gui/XournalView.cpp
+++ b/src/core/gui/XournalView.cpp
@@ -571,7 +571,7 @@ void XournalView::zoomChanged() {
 
     layoutPages();
 
-    if (zoom->isZoomPresentationMode() || zoom->isZoomFitMode()) {
+    if (zoom->isZoomPresentationMode()) {
         scrollTo(currentPage);
     } else if (zoom->isZoomSequenceActive()) {
         auto pos = zoom->getScrollPositionAfterZoom();


bhennion:
I agree with you, and I don't recall any reason for this.

Could you maybe retarget your PR to the release-1.3 branch? I think this counts as a bug fix and should appear in v1.3.5 (as opposed to waiting for v1.4.0 to come out).
You can do this by doing both:
1. Clicking on Edit at the top of the PR page, and changing the target branch there
2. Running `git rebase --onto release-1.3 master` and then `git push -f`

bhennion:
(The windows CI failing has nothing to do with your changes, don't worry about it).

TheKingOfLight:
Thanks for the help and instructions. I have updated the target branch to release-1.3 and rebased the PR.

bhennion (approved):
Thanks a lot! Merging once the CI clears.

bhennion:
> Thanks a lot! Merging once the CI clears.

Arf, it doesn't clear. Could you apply clang format to your changes: `git clang-format HEAD~1`, and then `git commit -a --amend` and `git push -f`

TheKingOfLight:
Sorry about the formatting! I have updated the PR using clang-format. Thank you for your support and guidance.

bhennion:
Thanks for the fix!

PR #7437, opened 2026-05-02
Fix Windows CI 

bhennion:


.github/actions/install_deps_windows/action.yml
```diff
@@ -38,5 +38,5 @@ runs:
     - shell: msys2 {0}
       run: |
         # install LuaGObject via luarocks as long as there is no MinGW package
-        luarocks install --lua-version=5.4 LuaGObject
+        luarocks install LuaGObject
```
bhennion, line 41: This fixes the failing CI, and looking at the logs, it seems that lua 5.5 is indeed used by default.
@rolandlo Do you remember why the lua version was pinned in that command?

rolandlo: I guess it was just a habit of pinning the Lua version when trying out different Lua versions locally. If it works without it's all fine.

bhennion: Alright then! Merging. Thanks!

PR #7441, opened 2026-05-04
Update Lua version in Windows packaging

rolandlo:
Fixes the issue mentioned in https://github.com/xournalpp/xournalpp/discussions/6008#discussioncomment-16793341

### DIFF
--- a/windows-setup/package.sh
+++ b/windows-setup/package.sh
@@ -83,14 +83,14 @@ cp "$prefix"/bin/gtk3-demo.exe "$setup_dir"/bin
 
 echo "copy lua-gobject and dependencies"
 cp "$prefix"/bin/libgirepository-2.0-0.dll "$setup_dir"/bin
-mkdir -p "$setup_dir"/lib/lua/5.4/LuaGObject
-cp "$prefix"/lib/lua/5.4/LuaGObject/lua_gobject_core.dll "$setup_dir"/lib/lua/5.4/LuaGObject
+mkdir -p "$setup_dir"/lib/lua/5.5/LuaGObject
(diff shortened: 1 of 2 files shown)


rolandlo:

windows-setup/package.sh
```diff
@@ -83,14 +83,14 @@ cp "$prefix"/bin/gtk3-demo.exe "$setup_dir"/bin
 
 echo "copy lua-gobject and dependencies"
 cp "$prefix"/bin/libgirepository-2.0-0.dll "$setup_dir"/bin
-mkdir -p "$setup_dir"/lib/lua/5.4/LuaGObject
-cp "$prefix"/lib/lua/5.4/LuaGObject/lua_gobject_core.dll "$setup_dir"/lib/lua/5.4/LuaGObject
+mkdir -p "$setup_dir"/lib/lua/5.5/LuaGObject
+cp "$prefix"/lib/lua/5.5/LuaGObject/lua_gobject_core.dll "$setup_dir"/lib/lua/5.5/LuaGObject
```
bhennion, line 87: Can we make it "version agnostic"? e.g. by installing all available versions? That would avoid the issue reappearing on next update.
Otherwise, can we make sure the script fails if the lib is not found?

rolandlo: It doesn't make much sense to me to install versions that can't be used from the Lua dll used for plugins. It took 5.5 years from Lua 5.4 to Lua 5.5, so it won't happen again quickly, but yes, I can make the script fail when the lib is not found.

bhennion: We could just restore the --lua-version I removed this weekend and add a comment next to it. That way we detect changes even without creating the installers.

PR #4161, opened 2022-07-21
[WIP] Refactor export jobs

forgedOctopus:
Closes #4150.

Work in progress.

PR #7463, opened 2026-05-11
use GtkFileChooserNative on non-Linux platforms (#3822)

### LINKED ISSUE #3822: Use system file manager on Windows

ellentane:
Closes #3822.

This switches the open, save, and export dialogs from `gtk_file_chooser_dialog_new` to `gtk_file_chooser_native_new`, so on Windows you get the real Explorer file picker, on macOS the Cocoa panel, and on Linux the portal file chooser (same as before on most desktops).

## What's in the PR

The series starts with some groundwork and then flips the three remaining call sites:

- A small helper, `NativeFileChooserHelper`, owns the native dialog, handles the async response signal, and destroys the dialog once the caller's response handler returns. All the lifetime fiddliness lives in one place so the callers stay readable.
- `XojOpenDlg` drops its internal `FileDlg` class. Every open flow (regular file, annotate PDF, template, image, multi-format) goes through the helper now.
- `XojSaveDlg` replaces the old `SaveExportDialog` class with two free functions, `showSaveDialog` and `showXoppSaveDialog`. Callers configure the dialog with a small lambda instead of a three-step setup.
- `BaseExportJob::showFileChooser`, `Control::saveImpl`, `DeviceTestingArea::saveLog`, `PageTemplateDialog::saveToFile`, and `applib_fileDialogSave` all use the new API.
- `XojMsgBox::replaceFileQuestion` gains an optional `onCancel` callback. The save flow needs it because the native dialog is already gone by the time the overwrite prompt shows, so we can't just re-open it on cancel.

## A note on the follow-up commit

I tested the first version on Windows and the picker still looked like the GTK one. It turns out calling `gtk_file_chooser_native_new` isn't enough by itself: GTK's Win32 backend only knows how to translate plain extension-pattern filters into the IFileDialog filter format. If the dialog has MIME filters, a custom match callback, or `gtk_file_filter_add_pixbuf_formats`, GTK silently falls back to the internal `GtkFileChooserDialog`. All of our filter helpers were MIME-based, so we were hitting that fallback on every save/open.

The last commit fixes this by converting the filter helpers to use extension patterns (`*.xopp`, `*.pdf`, etc.). A few side effects:

- `ExportType` no longer carries a `mimeType` field. Just the extension is enough now.
- Filters register both lowercase and uppercase globs (`*.pdf` and `*.PDF`) because glib's pattern matching is case-sensitive on Linux. Windows/macOS shell matching is case-insensitive, so the extra pattern is a harmless no-op there.
- Filtering now matches by filename extension rather than by MIME database. For everything Xournal++ actually opens and saves, that's equivalent.

While I was in there I also turned on `gtk_file_chooser_set_do_overwrite_confirmation` on the native dialog and skipped `add_shortcut_folder` on macOS (the native panel doesn't support shortcut folders). The `replaceFileQuestion` prompt still runs afterwards because `pathValidation` can append an extension, which may turn the name into an existing file that the OS dialog couldn't have flagged.

Because `gtk_file_chooser_native_new` is GTK 3.20, I bumped the minimum GTK from 3.18.9 to 3.20 in `CMakeLists.txt` and `debian/control`. All supported runners already ship 3.24+ (Ubuntu 22.04, jhbuild on macOS, MSYS2 mingw64), so no CI targets are affected.

## Commits

- `util: add onCancel callback to replaceFileQuestion` — backward-compatible extension
- `gui: add NativeFileChooserHelper for async native dialogs` — new helper, no callers yet
- `gui: convert XojOpenDlg to GtkFileChooserNative`
- `gui: convert save/export dialogs to GtkFileChooserNative`
- `gui: use extension-pattern filters so native file chooser stays native`

## Testing

Manually verified on Windows (MSYS2 mingw64) that these all show the real Windows file dialog now:

- Save As
- Open
- Export as PDF / custom export (PDF/PNG/SVG/XOJ)
- Annotate PDF
- Insert image
- Preferences → Input → Device testing area → Export logs
- Page template save

No existing unit tests cover these dialogs, and I didn't add new ones. I don't have a macOS machine, so a macOS-equipped reviewer would be very welcome.

## Tradeoffs and edge cases

- The "attach file to the journal" checkbox in the Annotate PDF dialog goes through `gtk_file_chooser_add_choice`. On Windows this maps to `IFileDialogCustomize::AddCheckButton` so it should render natively. On macOS the native panel can't display it, so the option silently falls back to its default ("false"). Happy to add a small follow-up dialog for macOS if reviewers would rather keep the option visible there.
- Narrow case: if `pathValidation` appends an extension that turns the chosen name into an existing file, the user may see two overwrite prompts — one from the OS dialog, one from `replaceFileQuestion`. Fixing this properly needs a refactor of the extension-munging, so I've left it out of scope.
- `PageTemplateDialog`: cancelling the overwrite prompt now closes the flow instead of leaving the file chooser open. The user re-opens save if they want a different name. Minor but worth flagging.

## Note on dormant PR #4161

#4161 (export-job refactor) has been dormant since April 2023. This PR only touches `BaseExportJob::showFileChooser`'s dialog wiring and drops the `mimeType` field on the filter struct, so conflicts should be small.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,7 +141,8 @@ endif()
 
 set(pkg-module-spec "")
 list(APPEND pkg-module-spec "glib-2.0 >= 2.32.0")
-list(APPEND pkg-module-spec "gtk+-3.0 >= 3.18.9")
+# GtkFileChooserNative (used for native Open/Save pickers on Win32/macOS) is GTK 3.20+.
+list(APPEND pkg-module-spec "gtk+-3.0 >= 3.20")
 list(APPEND pkg-module-spec "poppler-glib >= 0.41.0")
 if (APPLE)
   list(APPEND pkg-module-spec "poppler-cpp >= 0.41.0")
@@ -276,7 +277,7 @@ target_link_libraries(external_modules INTERFACE
 
 target_compile_definitions(external_modules INTERFACE
         GLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_40
-        GDK_VERSION_MIN_REQUIRED=GDK_VERSION_3_18
+        GDK_VERSION_MIN_REQUIRED=GDK_VERSION_3_20
         $<$<TARGET_EXISTS:X11::X11>:X11_ENABLED>
         )
 target_link_options(external_modules INTERFACE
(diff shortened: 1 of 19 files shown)


ellentane:
Closing to open a clean PR from the same branch; the description got mangled during a paste and the edit history is not something I can clean up in-place. The replacement PR will point at the same commits.

PR #7465, opened 2026-05-11
use GtkFileChooserNative on non-Linux platforms (#3822)

### LINKED ISSUE #3822: Use system file manager on Windows

ellentane:
Closes #3822.

This keeps the native chooser change narrow and GTK4-friendly, following the concern from #6049.

## Scope

- Keep the existing GTK file chooser path as the default, including Linux behavior.
- Use `GtkFileChooserNative` only on Windows/macOS, and only when building against GTK >= 3.20.
- Do not bump the project GTK minimum and do not touch packaging files.
- Avoid `gtk_native_dialog_run()` in the new code. The native path uses `gtk_native_dialog_show()` and the `response` signal.
- Keep chooser-specific UI out of native dialogs. For attach-file flows, native selection is followed by a small attach/don't attach question.

## Implementation notes

Native file choosers on Windows need plain extension-pattern filters; MIME filters or pixbuf-backed filters can make GTK fall back to its built-in file chooser. The native paths therefore use extension-pattern filters, while the normal GTK paths keep the existing MIME/pixbuf filters.

For custom export, the format is selected before opening the native save dialog. That avoids relying on the selected native filter after the dialog returns. The save dialog then receives only the chosen format's extension filter, so extension handling is deterministic.

Save/export flows share the existing `SaveExportDialog` path where possible. The native branch is a small wrapper around the native dialog lifecycle, not a replacement of the GTK dialog classes.

## Covered native paths

- Open file
- Open template
- Open image, with attach choice after file selection
- Annotate PDF, with attach choice after file selection
- Multi-format open dialog
- Save As `.xopp`
- PDF export
- Custom export
- Page template save
- Device testing log export
- Lua `fileDialogSave`

## Testing

Built locally with:

- `git diff --check`
- `git diff HEAD~1..HEAD --check`
- `ninja -C build xournalpp`


bhennion:
Thanks for the PR. There was already an attempt in #6049, but the changes need to be more subtle than what you propose. See this [comment](https://github.com/xournalpp/xournalpp/pull/6049#issuecomment-2486715596) and the following ones to get a description of the issue.

ellentane:
Thanks for the pointer. I pushed a follow-up commit that keeps this closer to the approach discussed in #6049.

The new native path no longer adds any `gtk_native_dialog_run()` calls. It uses a small helper around `gtk_native_dialog_show()` and the `response` signal, with native choosers enabled only on Windows/macOS when GTK >= 3.20 is available. Linux stays on the existing GTK chooser path.

I also moved the Custom Export format choice before the native save panel, so we don't depend on reading the selected filter after the native dialog returns. Native filters are extension-pattern based to avoid the Windows fallback to GTK's chooser. For the attach-file cases, the native file selection is followed by a small attach/don't attach question instead of trying to put that control inside the native chooser.

Local checks passed:

- `git diff --check`
- `git diff HEAD~1..HEAD --check`
- `ninja -C build xournalpp`

ellentane:
Happy to adjust anything if you have any notes left. @bhennion

bhennion:
One general comment about the PR (I'll dive into the code once this is settled):
I don't think we should treat Linux differently from Windows or MacOS here. Also, GTK 3.20 has been out for over 10 years now and I think it is safe to bump our requirements to 3.20.

The point is: having two significant parallel pieces of code for the same fonctionality is quite bad in terms of maintenance cost. We should only keep one (either the old one, or the one using native dialogs).

PR #7482, opened 2026-05-22
Added snap packaging

kenvandine:


rolandlo:
@kenvandine  We generally try to keep the packaging separate from the code repository. I would suggest that we create another repository (like xournalpp/xournalpp-snap) for the manifest and publishing workflow (in which we could use a chron job). Or is there a specific reason that you suggest adding those two files into the main xournalpp repository?

kenvandine:
Oh, a separate repository would be easy! You could push my packaging-only repository there. I can request the transfer now.

rolandlo:
> Oh, a separate repository would be easy! You could push my packaging-only repository there. I can request the transfer now.

Ok, I did so: https://github.com/xournalpp/xournalpp-snap

Issue #6505, opened 2025-06-28
Stylus input continues drawing after pressure is released


SantiagoBDEV:
### Operating System

Mac OS

### (Linux Only) Display Server

None

### Installation Method

Official release

### Xournal++ Version

1.2.7+dev

### libgtk Version

3.24.42

### Bug Description

When using any tool with the stylus, if pressure has been applied and then released, a trace follows the cursor (only when moved with the stylus) (e. g when drawing pressure works fine, but when stylus is up a trace keeps following the cursor and if the stylus is put away and usage is continued using the trackpad the trace stops until input from the tablet is given again). To make this cease, sudden and strong stylus input is required which seems to stop the tracing effect sometimes.

### Expected Behaviour

I expected that releasing pressure and separating the stylus from the tablet would stop drawing.

### Steps to Reproduce

Select pen, try to draw using Stylus, after that it keeps drawing even if no pressure at all is applied (stylus not touching tablet).

 Any tool remains active.

### Additional Context

I've installed 1.2.6, 1.2.7 and as a final effort 1.2.7+dev but no version seems to work for me. I've also updated my tablet's drivers. 

MacOS Sequoia 15.5

Following the steps to provide input logs gives
[ 97%] Linking CXX executable ../xournalpp
clang++: error: no such file or directory: 'AudioToolbox'
clang++: error: no such file or directory: 'AudioUnit'
clang++: error: no such file or directory: 'CoreFoundation'
clang++: error: no such file or directory: 'CoreServices'

(I've made sure to run xcode-select --install before, which confirms "Command line tools are already installed". I reinstalled them too.)

https://github.com/user-attachments/assets/65169d74-ccfb-470f-bb5e-c11f56fbf014

Thanks.

bhennion:
This is weird. It seems we are not receiving the "end of contact" event.

If you want to help us by providing some debug info we would need, you can

1. install the version from [this post](https://github.com/xournalpp/xournalpp/pull/6491#issuecomment-3016336289)
2. Go to Edit->Preferences then Input System
3. With your pen on the testing area: hover for 1 second, write for 1 second then hover for 1 second
4. Without other interactions with the testing area, just press the `Export logs` button and post here the resulting .zip.

Santiago-Beltran:
[resulting-logs.zip](https://github.com/user-attachments/files/20970586/resulting-logs.zip)

I'm using a Veikk A15 pen tablet.

Thanks once again

bhennion:
[...]

PR #7483, opened 2026-05-23
Sync modifier states on MOTION_EVENTS

### LINKED ISSUE #6833: Pen bindings get stuck when releasing button before lifting pen
### Operating System

Windows

### (Linux only) Distribution

Arch

Hyprland

Wayland

### Installation Method

Repos

### Xournal++ Version

1.2.7

### libgtk Version

3.24.51

### Bug Description

If you bind actions to pen buttons and don't lift the pen from the drawing surface before releasing a pressed pen button, its action gets stuck until pressed again.

I'm mildly certain that this is a regression from one of the recent releases.

### Expected Behaviour

not that

### Steps to Reproduce

1. press pen action button
2. lower pen onto surface, "activating" it
3. release button
4. lift pen
5. stuck action

### Additional Context

_No response_

MathisP75:
Fixes #6833. On some Wayland compositors (like Hyprland), the release event for stylus buttons is not registered on BUTTON_RELEASE_EVENTs while the pen tip is pressed down. This is probably because of [Wayland's implicit grab behavior](https://wayland.freedesktop.org/docs/book/Protocol.html#input), so the state of the buttons have to be synchronized on MOTION_EVENTs. Please let me know if there is a better way to implement this fix.

### DIFF
--- a/.github/available-build-targets.json
+++ b/.github/available-build-targets.json
@@ -35,6 +35,16 @@
       "displayed_name": "Ubuntu 24.04 ARM",
       "runner": "ubuntu-24.04-arm",
       "run_ci": "false"
+    },
+    "ubuntu-26": {
+      "displayed_name": "Ubuntu 26.04",
+      "runner": "ubuntu-26.04",
+      "run_ci": "false"
+    },
+    "ubuntu-26-arm": {
+      "displayed_name": "Ubuntu 26.04 ARM",
+      "runner": "ubuntu-26.04-arm",
+      "run_ci": "false"
     }
   },
   "macos_targets": {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -470,8 +470,15 @@ set(CPACK_STRIP_FILES ON) # Debian packages must be stripped from debug messages
 set(CPACK_DEBIAN_PACKAGE_RELEASE 2)
 set(CPACK_DEBIAN_PACKAGE_VERSION "${CPACK_PROJECT_VERSION}")
 set(CPACK_DEBIAN_PACKAGE_SECTION "graphics")
+set(_LIBXML2_DEP "libxml2 (>= 2.0.0)")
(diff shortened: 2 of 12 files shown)


bhennion:
Hey there. Thanks for the PR. I think however that the issue looks like an issue in hyprland, as if hyprland's mecanism to emulate mouse buttons with stylus swallowed the event. It should definitely be reported there as other apps may be affected.
It could be a little more subtle than just "swallowing": you could get similar behaviour if the event was not emitted by the same device (see eg https://github.com/xournalpp/xournalpp/issues/6505 for a similar thing).

In any case, it would be good to add input device debug logs in #6833 (using the device testing area in Settings->Input), so we can really know what hyprland is feeding us and argue why they do it wrong.

MathisP75:
Upon further testing, this indeed seems to be a Hyprland issue. It never issues a `BUTTON_RELEASE` event for the stylus modifier button, while both KDE and Gnome do. I will open an issue on Hyprland's GitHub page. I was unsure if the issue was on Hyprland's side, because Krita does not have this issue, but I think that's because Krita just sets the modifier buttons to false whenever the stylus leaves the screen. This would not add a check in the motion loop, and would be as simple as this :

```cpp
    if (event.type == LEAVE_EVENT) {
        this->inputContext->getHandRecognition()->unblock();
        this->actionLeaveWindow(event);
        this->modifier2 = false;
        this->modifier3 = false;
    }
```

I believe this would be harmless to add, and it would be an extra guardrail, since the modifier states should be `false` anyway when the pen is not touching/hovering the screen. I'll add some of the logs I collected in #6833. If this seems like a good idea, I could open another PR.

bhennion:
> If this seems like a good idea

Our input system definitely needs to be hardened against faulty drivers and upstream libraries. There are other issues of this sort (#6505) that would deserve a workaround similar to this. It is somewhat hard to make sure the workaround will not affect other devices though, so a wide range of testing is paramount.

PR #7526, opened 2026-06-10
Fix Laser pointer configuration assigned to button

### LINKED ISSUE #7524: Button color settings are not preserved
### Operating System

Linux

### (Linux only) Distribution

Fedora

Sway

Wayland

### Installation Method

git

### Xournal++ Version

c68f9841f ("Fix mishandled button configurations", 2026-04-06)

### libgtk Version

3.24.52

### Bug Description

The report is about laserpointer (and possibly other tools) assigned to a mouse or styles button, and the issue is bisected down to `c68f9841f``.
Before that commit, laserpointer would use the color as configured for the tool (in my case: red).
With that commit, settings would offer to set the color (and thickness) while the button is pressed and color is switched. Default is black, and when I change this to red it works for the running instance, but this setting is nor preserved. It does not get stored in the settings file and reverts back to black for the next start.

### Expected Behaviour

Preserve setting

### Steps to Reproduce

1. Assign laserpointer to button in settings
2. Change color in settings, OK
3. restart xournalpp

### Additional Context

_No response_

bhennion:
Fixes #7524 

This PR makes sure that color and size are stored in the settings file depending on the tool's capabilities, instead of the hard-coded "pen or highlighter (or text)". Doing this required a little bit more work than I hoped though (see changes in ToolEnums and ToolHandler).

I tested a dozen cases, so hopefully I did not introduce any new bug. Note that the transfer of information (the tool capabilities) from ToolHandler to ToolEnums was automated (copied + regex) so there should not be any new issues because of that.

@mjg Would you be able to test this and confirm is works?

### DIFF
--- a/src/core/gui/dialog/ButtonConfigGui.cpp
+++ b/src/core/gui/dialog/ButtonConfigGui.cpp
@@ -276,9 +276,9 @@ void ButtonConfigGui::enableDisableTools() {
     gtk_tree_model_get_value(model, &iter, 2, &value);
     auto action = static_cast<ToolType>(g_value_get_int(&value));
 
-    gtk_widget_set_visible(cbThickness, xoj::tool::hasCapability(action, TOOL_CAP_SIZE));
-    gtk_widget_set_visible(colorButton, xoj::tool::hasCapability(action, TOOL_CAP_COLOR));
-    gtk_widget_set_visible(cbDrawingType, xoj::tool::hasCapability(action, TOOL_CAP_RECTANGLE));
+    gtk_widget_set_visible(cbThickness, action != TOOL_NONE && xoj::tool::hasCapability(action, TOOL_CAP_SIZE));
+    gtk_widget_set_visible(colorButton, action != TOOL_NONE && xoj::tool::hasCapability(action, TOOL_CAP_COLOR));
+    gtk_widget_set_visible(cbDrawingType, action != TOOL_NONE && xoj::tool::hasCapability(action, TOOL_CAP_RECTANGLE));
     gtk_widget_set_visible(cbEraserType, action == TOOL_ERASER);
     gtk_widget_set_visible(cbStrokeType, action == TOOL_PEN);
 }
--- a/src/core/gui/dialog/ButtonConfigGui.cpp
+++ b/src/core/gui/dialog/ButtonConfigGui.cpp
@@ -276,67 +276,9 @@ void ButtonConfigGui::enableDisableTools() {
     gtk_tree_model_get_value(model, &iter, 2, &value);
     auto action = static_cast<ToolType>(g_value_get_int(&value));
 
-    switch (action) {
-        case TOOL_PEN:
-            gtk_widget_set_visible(cbThickness, true);
-            gtk_widget_set_visible(colorButton, true);
-            gtk_widget_set_visible(cbDrawingType, true);
-            gtk_widget_set_visible(cbEraserType, false);
-            gtk_widget_set_visible(cbStrokeType, true);
-            break;
-        case TOOL_HIGHLIGHTER:
-            gtk_widget_set_visible(cbThickness, true);
-            gtk_widget_set_visible(colorButton, true);
-            gtk_widget_set_visible(cbDrawingType, true);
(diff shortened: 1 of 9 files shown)


mjg:
Thanks, the PR compiles and works: settings are preserved.
Note that while it merges cleanly to master, master does not compile with this - I guess the new link tool needs to adjust to the new function signature from this PR.

rolandlo (changes requested):
Something is wrong with the PDF text selection tools. Why can we choose the drawing type, thickness when assigning a stylus button to that tool (not just the color) and why is the color ignored when we then use the stylus button for marking text?

Also please rebase on master and note that you need to remove the capability argument when creating the link tool in ToolHandler.cpp (as pointed out by @mjg).

bhennion:
> Something is wrong with the PDF text selection tools. Why can we choose the drawing type, thickness when assigning a stylus button to that tool (not just the color) and why is the color ignored when we then use the stylus button for marking text?

This is not new to this PR. I may as well try and fix it here though. I'll have a look.

> Also please rebase on master and note that you need to remove the capability argument when creating the link tool in ToolHandler.cpp (as pointed out by @mjg).

I don't think that's what @mjg was saying. This fix should go into the release branch. It is simply that changes on the master branch will force us to be careful when merging the release branch onto the master branch.

rolandlo:
> > Something is wrong with the PDF text selection tools. Why can we choose the drawing type, thickness when assigning a stylus button to that tool (not just the color) and why is the color ignored when we then use the stylus button for marking text?
> 
> This is not new to this PR. I may as well try and fix it here though. I'll have a look.
> 
> > Also please rebase on master and note that you need to remove the capability argument when creating the link tool in ToolHandler.cpp (as pointed out by @mjg).
> 
> I don't think that's what @mjg was saying. This fix should go into the release branch. It is simply that changes on the master branch will force us to be careful when merging the release branch onto the master branch.

Sorry, I misunderstood. Usually we have regressions in the master branch, not in the release branch.

bhennion:
> Sorry, I misunderstood. Usually we have regressions in the master branch, not in the release branch.

np. Technically, it is not a regression. It is just that a previous fix made this other bug apparent/matter.

mjg:
Yes, I assumed the fix to go on the release branch and noticed that it merges cleanly to master but does not compile. You would have noticed anyways, but just to save you the trouble I mentioned it.
Meanwhile I figured that it's just a one-line change anyways (https://github.com/mjg/xournalpp/tree/fixtools-master)

bhennion:
> > Something is wrong with the PDF text selection tools. Why can we choose the drawing type, thickness when assigning a stylus button to that tool (not just the color) and why is the color ignored when we then use the stylus button for marking text?
> 
> This is not new to this PR. I may as well try and fix it here though. I'll have a look.

This should be fixed now. Any other messed up tool like that?

bhennion:
I made it so the combo boxes displayed in the settings always match the tools capabilities. This is much cleaner an easier to maintain.

src/core/gui/dialog/ButtonConfigGui.cpp
```diff
@@ -276,67 +276,9 @@ void ButtonConfigGui::enableDisableTools() {
...
-            break;
-        default:
-            g_warning("Unhandled tool in ButtonConfigGui");
-            break;
-    }
+    gtk_widget_set_visible(cbThickness, xoj::tool::hasCapability(action, TOOL_CAP_SIZE));
+    gtk_widget_set_visible(colorButton, xoj::tool::hasCapability(action, TOOL_CAP_COLOR));
+    gtk_widget_set_visible(cbDrawingType, xoj::tool::hasCapability(action, TOOL_CAP_RECTANGLE));
+    gtk_widget_set_visible(cbEraserType, action == TOOL_ERASER);
+    gtk_widget_set_visible(cbStrokeType, action == TOOL_PEN);
```
rolandlo: I think a check is missing that `action` is not `TOOL_NONE`. Otherwise we run into an assertion failure in `typeToCapabilities()`

bhennion: Indeed. I added a check

rolandlo (approved):
LGTM

PR #7539, opened 2026-06-17
Load settings from system config dirs as fallback

### LINKED ISSUE #4657: Fully implement the XDG Base Directory Specification
I would like to create configuration files at /etc/xdg/xournalpp and have xournal++ read options from there if these options are not set locally in `$XDG_CONFIG_DIR/xournalpp`. These options are independent of the current user.

Instead of just using `get_user_config_dir` to get the configuration files location in `$XDG_CONFIG_HOME`, i would propose to use `get_system_config_dirs` to look for configuration files in `$XDG_CONFIG_DIRS`, if none are found in `$XDG_CONFIG_HOME`.
This is part of the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):
>  $XDG_CONFIG_DIRS defines the preference-ordered set of base directories to search for configuration files in addition to the $XDG_CONFIG_HOME base directory. [...]
> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used. 
>  [...]
> The base directory defined by $XDG_CONFIG_HOME is considered more important than any of the base directories defined by $XDG_CONFIG_DIRS.

None

**Additional context**
I use NixOS, which allows me to declaratively define a reproducible system. While tools like home-manager allow me to place configuration files in $XDG_CONFIG_HOME, pure NixOS only allows system-wide configuration files (e.g. in `/etc/xdg`).
This works for most applications, but Xournal++ is one of the few exceptions.

kazam0180:
When `settings.xml` file doesn't exist in `XDG_CONFIG_HOME`, try to load from paths in `XDG_CONFIG_DIRS`.
This closes #4657 I guess.

I am not very good in C/C++. After bonking my head for two days, I asked an LLM for help (not vibe coded). If there is some issue in my changes, I will be here to fix it.

### DIFF
--- a/src/core/control/settings/Settings.h
+++ b/src/core/control/settings/Settings.h
@@ -104,6 +104,7 @@ private:
 class Settings {
 public:
     /*[[implicit]]*/ Settings(fs::path filepath);
+    /*[[implicit]]*/ Settings(fs::path filepath, fs::path loadPath);
     Settings(const Settings& settings) = delete;
     void operator=(const Settings& settings) = delete;
     virtual ~Settings();
@@ -615,13 +616,19 @@ public:
 
     LatexSettings latexSettings{};
 
-    inline const fs::path& getSettingsFile() const { return filepath; }
+    inline const fs::path& getSettingsFile() const { return savePath; }
 
 private:
     /**
-     *  The config filepath
+     *  The config file path used for saving
      */
-    fs::path filepath;
+    fs::path savePath;
+
+    /**
+     *  The config file path used for loading, may differ from savePath
+     *  when falling back to system directories
+     */
+    fs::path loadPath;
 
 private:
     /**
--- a/src/core/control/settings/Settings.h
+++ b/src/core/control/settings/Settings.h
@@ -103,8 +103,8 @@ private:
 
 class Settings {
 public:
-    /*[[implicit]]*/ Settings(fs::path savePath);
-    /*[[implicit]]*/ Settings(fs::path savePath, fs::path loadPath);
+    explicit Settings(fs::path savePath);
+    Settings(fs::path savePath, fs::path loadPath);
     Settings(const Settings& settings) = delete;
     void operator=(const Settings& settings) = delete;
     virtual ~Settings();
--- a/src/core/control/settings/Settings.h
+++ b/src/core/control/settings/Settings.h
@@ -103,8 +103,8 @@ private:
 
 class Settings {
 public:
-    /*[[implicit]]*/ Settings(fs::path filepath);
-    /*[[implicit]]*/ Settings(fs::path filepath, fs::path loadPath);
+    /*[[implicit]]*/ Settings(fs::path savePath);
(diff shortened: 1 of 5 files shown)


src/util/PathUtil.cpp
```diff
@@ -398,6 +398,33 @@ auto Util::getConfigFile(const fs::path& relativeFileName) -> fs::path {
...
+    // Look in system config directories
+    g_message("Looking for config file in system config directories");
+    const gchar* const* systemConfigDirs = g_get_system_config_dirs();
+    if (systemConfigDirs != nullptr) {
+        for (int i = 0; systemConfigDirs[i] != nullptr; i++) {
+            fs::path systemConfigPath = GFilename(systemConfigDirs[i]).toPath().value_or(fs::path());
+            systemConfigPath /= CONFIG_FOLDER_NAME;
+            systemConfigPath /= relativeFileName;
+
+            if (fs::exists(systemConfigPath)) {
```
bhennion: Functions that interact with the filesystem should be shielded against IO errors. In this instance, this probably means a try/catch scope around all this, to catch the exceptions raised in case of an IO issue.

src/core/control/settings/Settings.h
```diff
@@ -103,7 +103,8 @@ class SElement final {
 
 class Settings {
 public:
-    /*[[implicit]]*/ Settings(fs::path filepath);
+    /*[[implicit]]*/ Settings(fs::path savePath);
+    /*[[implicit]]*/ Settings(fs::path savePath, fs::path loadPath);
```
bhennion:
```suggestion
    explicit Settings(fs::path savePath);
    Settings(fs::path savePath, fs::path loadPath);
```

bhennion (changes requested):
The PR looks mostly ok to me (see two comments below). It must be tested on different platforms before being merged though (Windows/MacOS).

src/util/PathUtil.cpp
kazam0180: ah ok, i will update

kazam0180: @bhennion should i use `g_warning` to print exception?

bhennion: That's fine I think. The better alternative would be to show a popup informing the user of the error (via `XojMsgBox::showErrorToUser()` in util/XojMsgBox.h), but it's not mandatory I guess.

bhennion:
There is a small clang-format issue (you can run `git-clang-format master` from your branch to fix it).
Other than that, the code look good.
Needs to be thoroughly tested on Windows/MacOS though.

kazam0180:
> you can run `git-clang-format master` from your branch to fix it

I had a feeling that clang-format check would fail due to version difference. I didn't know about `git-clang-format`. I will do it when I get on desktop

bhennion (approved):
The changes look good to me. @rolandlo would you have time to test this on MacOS/Windows at some point (no hurry, of course)

PR #7553, opened 2026-06-24
Add wrapping to Text edition

### LINKED ISSUE #1756: Adding automatic linebreaks when using a text box
I am using xournal++ version 1.0.16, commit 1429879c, built on Jan 24 2020, 21:22:59.

Currently if you use the texbox and you keep typing in it, it goes beyond the extend of the page. This annoys me because I have to add my own line break by pressing enter. However this screws everything up if I want to remove some words from a textbox since later words don't move up.

I want the program to automatically add a linebreak when the textbox goes beyond the extend of the page. 

Also if I remove some words from the textbox, I want the linebreaks to automatically readjust.

A good example if how textboxes currently work in libreoffic write. Version 6.3.4.2.

Screenshot from 2020-03-02 09-40-19

Text going beyond the page

Screenshot from 2020-03-02 09-40-44

What the linebreaks should look like

Screenshot from 2020-03-02 09-40-57

Removing some words from the textbox, but it doesn't automatically resize.

Screenshot from 2020-03-02 09-41-19

How it should look after resizing.

### LINKED ISSUE #6693: 📝 Feature Request: Resize and Move the Text Tool Like in Design Software

Yes. Currently, in **Xournal++**, adjusting text size and style is inconvenient. Every time I want to change the text size or font, I have to go through the **Tools** section.  
This slows down the workflow, especially when making quick notes.

## **Problem**  
Currently, in **Xournal++**, adjusting text size and style is inconvenient. Every time I want to change the text size or font, I have to go through the **Tools** section.  
This slows down the workflow, especially when making quick notes or annotations.  

## **Proposed Solution**  
Make the **Text Tool** behave more like in design/photo-editing applications (e.g., **Canva, Photoshop, PowerPoint**):  

- **Resize by dragging corners** → After typing text, users should be able to resize it freely by dragging the corners of the text box.  
- **Move by dragging** → Text should be movable by clicking and dragging anywhere inside the text box.  
- **Quick formatting options** → Options like **font size, font family, bold, italic, underline, and text color** should be available directly in the **toolbar/home screen**, instead of hidden under the Tools menu.  

## **Benefits**  
- Makes working with text **faster and more intuitive**.  
- Reduces the number of clicks needed for formatting.  
- Improves **usability for annotations, notes, and diagram design**.  

## **Summary**  
This feature would make the text tool in **Xournal++** behave more like modern design software, creating a smoother and more user-friendly experience.

## **Alternatives**  
- Keep the current Tools menu workflow.  
- Add a **floating formatting toolbar** that appears whenever a text box is selected.  

### Additional Context

_No response_

bhennion:
This PR adds wrapping when editing texts:

[Enregistrement d'écran_20260624_220428.webm](https://github.com/user-attachments/assets/570383bb-aca8-4ff8-80eb-0641ab90e585)

Fix #1756 

Implementation details:

 - The wrapping itself is handled by Pango. We just give it a fixed width.
 - The icons are GtkButtons and are thus themed correctly

Things to discuss:

1. What should be done with preexisting Text objects?
2. Should creating non-wrapped text still be possible?
3. What should be the default wrapping width (if any)?

There are a couple of things left to do

- [x] Find/Make suitable icons
- [x] Optimize icons for size before merging
- [x] Move icons when scrolling/zooming
- [x] Add load/save test units
- [x] Implement the changes related to the above questions
- [x] Test and adapt for RTL languages? (I don't know how)


bhennion:
I made half decent icons

Capture d&#39;écran_20260626_101053

 For the rest, I'll need input and answers to the questions in the OP.

youssefadly237:
> What should be done with preexisting Text objects?

I don't think it should change at all on the user level, if you used inkscape or illustrator before, you get both behaviors, if you click to add a textbox it just has infinite width as we had before, but if you click hold and drag you get a textbox with specified max width

> Should creating non-wrapped text still be possible?

in my opinion I think yes

> What should be the default wrapping width (if any)?

the width of the drag

> Test and adapt for RTL languages? (I don't know how)

I am a native Arabic speaker, I could help with testing RTL

rolandlo:
> > What should be done with preexisting Text objects?
> 
> I don't think it should change at all on the user level, if you used inkscape or illustrator before, you get both behaviors, if you click to add a textbox it just has infinite width as we had before, but if you click hold and drag you get a textbox with specified max width
> 
> > Should creating non-wrapped text still be possible?
> 
> in my opinion I think yes
> 
> > What should be the default wrapping width (if any)?
> 
> the width of the drag

I agree. The Inkscape behaviour is pretty reasonable.

bhennion:
I implemented the following behaviour:

- You can only create a Text object by clicking
- When you create a Text object, it has no wrapping
- If you start dragging the "wrap" handle on the text box, wrap turns on
- If you edit a preexisting text object, the wrap mode and width are preserved from the last time this object was edited.

I am a bit reticent to the creation of text object by dragging and selecting an area. I've encountered it in several apps but it always left me dubious for the following reason: I feel like it creates a blurry expectation regarding the height of the box. In most implementations I've seen, the height is in fact never really used (and I don't think there is a reasonable use for it either). This, of course, can be alleviated by a better display than a simple rectangle (e.g. if the height of the rectangle does not vary).

As another weak reason against that idea: it means more work and a deeper restructuration of the Text tool (using the RectangleSelector class as a base would help, I guess).

bhennion:
> > Test and adapt for RTL languages? (I don't know how)
> 
> I am a native Arabic speaker, I could help with testing RTL

I've been experimenting a bit by pasting arabic sentences in the Text object. Independently if this PR, it seems to behave pretty weirdly. We may want to address this in another PR (possibly even targetted at the release branch).

youssefadly237:
> I am a bit reticent to the creation of text object by dragging and selecting an area. I've encountered it in several apps but it always left me dubious for the following reason: I feel like it creates a blurry expectation regarding the height of the box. In most implementations I've seen, the height is in fact never really used (and I don't think there is a reasonable use for it either). This, of course, can be alleviated by a better display than a simple rectangle (e.g. if the height of the rectangle does not vary).

in InDesign if I am not mistaken the height was used to clip the text inside the box, and on text overflow it gives you a little red mark on the bottom right to mark the text box, but I feel that this is more of publication thingy, I don't see myself picking that over just an infinite height text box in xournalpp

youssefadly237:
bidi works fine btw

but I have noticed that navigating the text with up and down arrows became a little awkward, shouldn't it be column-based rn?

also can we have a setting for hyphenation (off by default)?

bhennion:
> but I have noticed that navigating the text with up and down arrows became a little awkward, shouldn't be column-based rn?

Thanks for the catch! Should be fixed now.

> also can we have a setting for hyphenation (off by default)?

Pango offers hyphenation-based wrapping, but it doesn't take into account any language-based rules. It results in hyphens at really weird places
Capture d&#39;écran_20260627_192340

See for instance the split word "by" in this example.

I find it a bit too weird to be useful.

Later down the road, we could add an option to justify the text though. I think that'd help more that hyphenation.

bhennion:
> I implemented the following behaviour:
> 
> * You can only create a Text object by clicking
> * When you create a Text object, it has no wrapping
> * If you start dragging the "wrap" handle on the text box, wrap turns on

I found an issue with this: if you click to create a text box and paste a large enough amount of text, the handle to enable wrapping disappears out of the page and you get stuck this way.
I see two possible solutions

1. Automatically enable wrapping if the edge of the page is reached
2. Keep the wrapping handle in the page when the text goes beyond the edge.

IMO, 1. makes more sense as invisible text out of the page is not something we really like. Any opinion?

youssefadly237:
I don't think this is a new issue since it also affect the delete button (except that we can delete via keyboard shortcut)

I like the style in #4938 since we may add text options later

rolandlo:
It feels strange to me that the wrapping width is not scaled when scaling a selection with text elements. Did you decide so on purpose? It's still possible to only change the font size by clicking on the font button.

rolandlo:
> > I implemented the following behaviour:
> > 
> > * You can only create a Text object by clicking
> > * When you create a Text object, it has no wrapping
> > * If you start dragging the "wrap" handle on the text box, wrap turns on
> 
> I found an issue with this: if you click to create a text box and paste a large enough amount of text, the handle to enable wrapping disappears out of the page and you get stuck this way. I see two possible solutions
> 
>     1. Automatically enable wrapping if the edge of the page is reached
> 
>     2. Keep the wrapping handle in the page when the text goes beyond the edge.
> 
> 
> IMO, 1. makes more sense as invisible text out of the page is not something we really like. Any opinion?

To me 1. seems preferable. I wonder though if wrapping can be disabled again for a text field where the user has enabled it by either dragging the handle or by pasting a large chunk of text.

bhennion:
> I wonder though if wrapping can be disabled again for a text field where the user has enabled it by either dragging the handle or by pasting a large chunk of text.

For now, I just can't really imagine any use case where the user would really want to disable any form of wrapping. I'd suggest implementing it in case there is some demand for it in the future.

> It feels strange to me that the wrapping width is not scaled when scaling a selection with text elements. Did you decide so on purpose? It's still possible to only change the font size by clicking on the font button.

No, I just didn't think of it.

bhennion:
> > It feels strange to me that the wrapping width is not scaled when scaling a selection with text elements. Did you decide so on purpose? It's still possible to only change the font size by clicking on the font button.
> 
> No, I just didn't think of it.

It is fixed now

bhennion:
> To me 1. seems preferable.

Done.

rolandlo (approved):
It all looks fine to me now. This is a great! Wrapped text was really missing badly.

bhennion:
Thanks a lot! Merging in 24 hours unless an objection is raised.

rawfiul:
this is it. Time for onenote to pack it up.

tobiasBora:
Great thanks a lot! Any plan to integrate it with the lua API?

youssefadly237:
> Great thanks a lot! Any plan to integrate it with the lua API?


I don't know but the api can not trigger while you are typing, so you mean on selected text (toggle and resize)?

rolandlo:
> Great thanks a lot! Any plan to integrate it with the lua API?

Sure. I'm working on adding missing API related to getting and adding elements, including a "wrap" field for text elements.

tobiasBora:
> Sure. I'm working on adding missing API related to getting and adding elements, including a "wrap" field for text elements.

It's what I meant, thanks! It would also be so great to be able to interact in lua with the text while typing as I'd use it a lot… (e.g. to change the content of the current text, as I would use it to change the grade given to a student based on the typed shortcut etc) So if you know how to do that it would be even more awesome.

PR #7574, opened 2026-07-02
Clear modifier states when stylus leaves the screen

### LINKED ISSUE #6833: Pen bindings get stuck when releasing button before lifting pen

MathisP75:
Fixes #6833. See #7483 for more context. This should not cause any issues since the modifiers should be off when the stylus leaves the screen. For my stylus, it's modifiers 2 and 3, but I do not know how it would affect other tablets, so testing would be necessary.

### DIFF
--- a/src/core/gui/inputdevices/StylusInputHandler.cpp
+++ b/src/core/gui/inputdevices/StylusInputHandler.cpp
@@ -103,6 +103,8 @@ auto StylusInputHandler::handleImpl(InputEvent const& event) -> bool {
     if (event.type == LEAVE_EVENT) {
         this->inputContext->getHandRecognition()->unblock();
         this->actionLeaveWindow(event);
+        this->modifier2 = false;
+        this->modifier3 = false;
     }
 
     // Trigger end of action if pen tip leaves screen or mouse button is released


bhennion:
I think we can merge this. I tested as much as I could (so, no much) but I don't think it can do real harm.

Merging in 72 hours unless an objection is raised

youssefadly237:
I noticed a change in behavior I don't think it harmful but

1. activating eraser with side button
2. going outside the window while the pressing the tip
3. back to the window while the window is still pressed

the eraser used to keep working normally, now it switches back to the previous tool

PR #7579, opened 2026-07-05
Clean up some of the toolbar items

### LINKED ISSUE #4712: Create icons for separator and spacer instead replacing hardcoded drawing
Currently the icons for the separator and spacer that are used for customizing the toolbar are hardcoded (drawn with cairo in `ToolSeparatorImage.cpp`). This means that the icons will look the same in all themes. It would be better to have real, theme-dependent icons.
Currently if the theme uses a gray background for dialogs, the two icons cannot be distinguished. It would be even worse with red background.
The color of the widget background is one thing. The other one is the color of the page, during drag&drop. So the icons should be clearly visible on any page color.

Icons for separator and spacer.

Continue hardcoded drawing, but improved and making it dependent on the choice of theme.

**Additional context**

Currently (in #4674 and after that PR is merged) it looks like this (without the text):
grafik

In the code:
https://github.com/rolandlo/xournalpp/blob/2b14ed30eef54710d57254c5024c817852164d64/src/core/gui/toolbarMenubar/icon/ToolbarSeparatorImage.cpp#L20-L49

bhennion:
Two things (and a half)

1. Ensure that a FontButton in a vertical toolbar does not stretch the toolbar's width. This is done by turning off the display of the font's name/size in this case.
2. Remove the code that is specific to SPACER and SEPARATOR. They are now treated lke any other tool item, simplifying the code.
This added the need for proper icons (thus implementing #4712) and made me rename the vertical tool icon from "spacer" to "vertical-tool", so the name spacer becomes available for, well, the spacer.


rolandlo (commented):
The changes look good to me. However I have a problem and currently don't understand it. The spacer is displayed like the vertical space tool in the customize dialog.

Image

rolandlo (approved):
I found the cause for my problem. I hadn't cleaned up my inst folder, so there was an old xopp-spacer icon left in the iconsColor-dark folder. Hopefully that doesn't happen to other users when installing Xournal++. Otherwise we could put a copy of the iconsLucide-dark variant into the iconsColor-dark folder.

PR #7582, opened 2026-07-05
Adaptative toolkit

bhennion:
The PR implements an "adaptative toolkit". It is an item the user can add to their toolbars. The content of the toolbar then depends on the active tool, only showing the buttons that are relevant when using this tool.
Here is an example of a possible configuration:

[Enregistrement d'écran_20260705_142000.webm](https://github.com/user-attachments/assets/c678c8a7-9aa7-4d4e-9b95-d45af4d33972)

The map `ToolType -> Content of the toolkit` is hardcoded. Making it configurable could be follow up PR. Not sure it is worth the effort though.

The implementation is surprisingly simple. There are however two small issues:

1. If there isn't enough room, ~~the whole toolkit disappears~~ **Edit:** Now fixed
2. The color shortcuts (Ctrl + 0-9) do not take the color items in the toolkit into account.

Problem 2. is surely fixable with a small refactor of those color shortcuts (I'll try to prepare another PR for that)
~~Problem 1. is more complicated. I do not see any easy solution. I still think the feature is valuable with this defect though.~~

This PR is based on #7579.


MichelReims:
Hello,
I don't know how I can use adaptative toolbar.
Can you give me an example ?

bhennion:
> Hello, I don't know how I can use adaptative toolbar. Can you give me an example ?

I don't quite understand your question here. This is just a way of always having the relevent buttons accessible, without having an enormous toolbar. When using Text, we have the alignment + font buttons, when using Pen we have the size +filling buttons, and so on.

MichelReims:
I don’t know how to implement and configure adaptative toolbar.

youssefadly237:
> I don’t know how to implement and configure adaptative toolbar.

this feature did not land in upstream xournalpp yet, you would have to clone and build this pr yourself or you can wait till this feature gets merged then you can use it in the nightly releases, and later in the stable ones

ad2004teaster-cmyk:
This addition might seem useful to some, and I support including it in Xournal++, provided it is optional, meaning it can be enabled or disabled.

Personally, I prefer a static toolbar layout where icons for tools unavailable to the active mode appear dimmed and unclickable.

**This contrasts with the "Adaptive toolkit" option, which constantly hides and reveals icons as you switch tools; that approach causes significant distraction, especially with a customized toolbar like mine, where I place all the tools I need in a single top bar, leaving the rest of the space and all other sides dedicated entirely to the document.**

Issue #7583, opened 2026-07-05
Latex tool: to do list


immi-h:
Now that the latex button is a proper tool of its own, here is a list of (potential) follow-up improvements. (I'm planning to work on some of those, but some might need discussion first.)

- [ ] 1. When one edits a latex object, undoing that edit requires doing undo twice.
- [ ] 2. After creating a tex-object, that object is selected. It often happens to me that I click elsewhere to unselect it, which then opens the tex-dialog again. Suggestion: If a first click only unselects the object; only a second click opens the tex-dialog again. 
- [ ] 3. When one edits a latex object, the scaling factor should remain the same. (Currently, the height remains the same, so the scaling factor may change, e.g. if one replaces "X" by "\hat{X}".)
- [ ] 4. Selecting a latex object and clicking on a color should change the color of that object.

More suggestions can be found in #1115.

immi-h:
Some more comments:

Concerning 1: This is just a small bug which I should be able to fix quickly.

Concerning 2: The above suggestion would be inconsistent with how other tools behave. Nevertheless, this might be the best solution; see the discussion at the bottom of #6823.

Concerning 3: Currently, the scaling factor is not stored in a TexImage object. It can be recovered by comparing its width or height with the size of pdf object, but that feels a bit unsatisfactory. (Theoretically, rounding errors might add up.) Should one nevertheless use this method to preserve the scaling factor when editing, or should one make the TexImage object store the scaling factor? (This would involve changing the xopp-file format.) I'd suggest doing it the easy way first (recovering the scaling factor) and keeping in mind that one might one day improve this (maybe when the file format needs to be changed anyway, for some other reason).

Concerning 4: This is maybe the most important issue, but I don't know yet how this could be implemented. The problem is that changing the color would (currently) mean that the tex object has to be re-rendered. Here are some options:

- Re-render the selected tex-objects synchronously.
Problem: xournalpp might freeze for a long time (or even forever if there are latex problems.)
- Re-render the selected tex-objects asynchronously.
Problem/questions: What to do if the file is saved when not yet everything has been re-rendered? Mark some of the tex-objects as non-yet rendered and continue rendering when the file is opened again? (Would the file then be marked as modified, even if the user didn't modify anything?)
- Change the way in which color is applied to latex objects, namely: instead of letting latex apply the color, let latex create a "color-neutral" pdf and change the color in the pdf afterwards.
[...]

PR #7584, opened 2026-07-05
Tex undo fix

immi-h:
Currently, editing a tex object creates two undo actions (one removing the old object and one creating the new one); see #7583. This is a fix for this.

However, I am not sure whether I locked the document as necessary in LatexController::insertTexImage (probably not). Concretely:
- Does one need to acquire a unique lock when calling layer->removeElement?
- Does one need to acquire a unique lock when creating a selected element using SelectionFactory::createFromFloatingElement and setSelection?
- Does one need to acquire a shared lock while using a variable layer (obtained from page->getSelectedLayer())?
(If this is the case, then there's probably also a similar bug in LinkHandler::startEditing, where a layer variable is created within a lock, but still used after the lock.)

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -300,8 +300,8 @@ void LatexController::insertTexImage() {
         auto insertUndoAction = std::make_unique<InsertUndoAction>(page, layer, this->temporaryRender.get());
         groupUndoAction->addAction(std::move(insertUndoAction));
         undo->addUndoAction(std::move(groupUndoAction));
-        Range oldRange(selectedElem->getSnappedBounds());
-        Range newRange(temporaryRender->getSnappedBounds());
+        Range oldRange(selectedElem->boundingRect());
+        Range newRange(temporaryRender->boundingRect());
         Range repaintRange = oldRange.unite(newRange);
         page->fireRangeChanged(repaintRange);
     } else {
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -30,6 +30,8 @@
 #include "model/TexImage.h"                  // for TexImage
 #include "model/Text.h"                      // for Text
 #include "model/XojPage.h"                   // for XojPage
+#include "undo/DeleteUndoAction.h"           // for DeleteUndoAction
+#include "undo/GroupUndoAction.h"            // for GroupUndoAction
 #include "undo/InsertUndoAction.h"           // for InsertUndoAction
 #include "undo/UndoRedoHandler.h"            // for UndoRedoHandler
 #include "util/Assert.h"                     // for xoj_assert
@@ -268,7 +270,9 @@ void LatexController::insertTexImage() {
     xoj_assert(this->isValidTex);
     xoj_assert(this->temporaryRender != nullptr);
 
-    auto lock = std::shared_lock(*this->control->getDocument());
+    Document* doc = this->control->getDocument();
+
+    auto lock = std::shared_lock(*doc);
     Layer* layer = page->getSelectedLayer();
     XournalView* xournal = this->control->getWindow()->getXournal();
     auto pageNr = xournal->getCurrentPage();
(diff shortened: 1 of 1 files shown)


bhennion:
Your branch is off. I think it is based on your old PR's branch instead of on latest master. Just rebase the last commit onto master and it should be fixed.

The document mutex must be locked:

- with a unique_lock whenever some part of the document is modified
- with a shared_lock whenever the document is accessed but **not modified**

To answer your points individually:

> * Does one need to acquire a unique lock when calling layer->removeElement?

Yes. The document is modified (Removing an element means some iterators will get invalidated and a concurrent thread reading the data may end up causing a SegFault)

> * Does one need to acquire a unique lock when creating a selected element using SelectionFactory::createFromFloatingElement and setSelection?

No. `SelectionFactory::createFromFloatingElement()` never needs a lock: the selection is not part of the document (the selected objects are not in the document structure tree) and neither is the floating element.
Note that the functions in SelectionFactory already lock the mutex when needed.

> * Does one need to acquire a shared lock while using a variable layer (obtained from page->getSelectedLayer())?
>   (If this is the case, then there's probably also a similar bug in LinkHandler::startEditing, where a layer variable is created within a lock, but still used after the lock.)

It depends what "using" means. If you are dereferencing the pointer, then yes, you need some form of lock. If you are just using the pointer as a unique identifier, then no, you don't need the lock.
I'm guessing you are referring to things like `std::make_unique<InsertUndoAction>(page, layer, link)` in `LinkHandler::startEditing()`, right?
There, the pointer is not dereferenced, only stored for later use (and when used later, the mutex is locked).

immi-h:
Thanks for the detailed answers. This helps me a lot understanding what I'm doing (also for things I might do in the future).
I'll fix things accordingly and also rebase the branch as soon as I find time.

immi-h:
> Your branch is off. I think it is based on your old PR's branch instead of on latest master. Just rebase the last commit onto master and it should be fixed.

Done.

> (...)
> 
> > * Does one need to acquire a unique lock when calling layer->removeElement?
> 
> Yes. The document is modified (Removing an element means some iterators will get invalidated and a concurrent thread reading the data may end up causing a SegFault)

Done.

> > * Does one need to acquire a shared lock while using a variable layer (obtained from page->getSelectedLayer())?
> >   (If this is the case, then there's probably also a similar bug in LinkHandler::startEditing, where a layer variable is created within a lock, but still used after the lock.)
> 
> It depends what "using" means. If you are dereferencing the pointer, then yes, you need some form of lock. If you are just using the pointer as a unique identifier, then no, you don't need the lock. I'm guessing you are referring to things like `std::make_unique<InsertUndoAction>(page, layer, link)` in `LinkHandler::startEditing()`, right? There, the pointer is not dereferenced, only stored for later use (and when used later, the mutex is locked).

Then I'm wondering about the following: at the beginning of LatexController::insertTexImage, a shared lock is acquired when checking whether the active page has changed. Is this lock really required? (I'm a bit unsure which pointers are part of the document and which are not.)

src/core/control/LatexController.cpp
```diff
@@ -280,18 +284,31 @@ void LatexController::insertTexImage() {
...
+        if (elementIndex != Element::InvalidIndex) [[likely]] {
+            deleteUndoAction->addElement(layer, std::move(orig), elementIndex);
+        }
+        groupUndoAction->addAction(std::move(deleteUndoAction));
+
+        auto insertUndoAction = std::make_unique<InsertUndoAction>(page, layer, this->temporaryRender.get());
+        groupUndoAction->addAction(std::move(insertUndoAction));
+        undo->addUndoAction(std::move(groupUndoAction));
+        Range oldRange(selectedElem->getSnappedBounds());
+        Range newRange(temporaryRender->getSnappedBounds());
```
```suggestion
        Range oldRange(selectedElem->boundingRect());
        Range newRange(temporaryRender->boundingRect());
```
For now, there is no difference for Tex objects, but

1. there might be, someday
2. the intent is really to get what needs to be repainted

bhennion:
> Then I'm wondering about the following: at the beginning of LatexController::insertTexImage, a shared lock is acquired when checking whether the active page has changed. Is this lock really required? (I'm a bit unsure which pointers are part of the document and which are not.)

It is required. Dereferencing `page` is safe because it is a shared_ptr, but when calling `getSelectedLayer()`, we are fetching something in the page's `std::vector<Layer*>`. If a concurrent thread were to add/remove a layer at the same time, we may be having a SegFault (or, if we are lucky, just get the wrong Layer*).

immi-h:
> > Then I'm wondering about the following: at the beginning of LatexController::insertTexImage, a shared lock is acquired when checking whether the active page has changed. Is this lock really required? (I'm a bit unsure which pointers are part of the document and which are not.)
> 
> It is required. Dereferencing `page` is safe because it is a shared_ptr, but when calling `getSelectedLayer()`, we are fetching something in the page's `std::vector<Layer*>`. If a concurrent thread were to add/remove a layer at the same time, we may be having a SegFault (or, if we are lucky, just get the wrong Layer*).

Ok. Then do I see correctly that getSelectedLayer() is the only thing where the lock is required (so that in principle, one could release it right afterwards)?  (I'm asking this only so that I get a better understanding of this.)

immi-h:
And concerning your fix of Range rectangles: Note that I stole this code from the LinkHandler, so I guess that there, it should be fixed, too.

bhennion:
> Ok. Then do I see correctly that getSelectedLayer() is the only thing where the lock is required (so that in principle, one could release it right afterwards)? (I'm asking this only so that I get a better understanding of this.)

Well, in theory, if another thread were to remove a page, it would also delete the corresponding PageView, so dereferencing `view` would not be safe without the lock. However, to my knowledge, all those dangerous document modifications are all done in the main thread, and this code is also only executed by the main thread, so there should really be no risk at all. Better safe than sorry, I would say.

bhennion (approved):
(Squashing and) Merging in 24 hours unless an objection is raised.
Thanks for the PR!

PR #7593, opened 2026-07-07
Make LinkDialog compatible with GTK-4

bhennion:
I didn't notice while reviewing #6841, but some things are incompatible with GTK4. This PR fixes that.
There is one thing I removed for simplicity: the window was getting resized as the user was typing in the GtkTextView. There may be a way to port it but I'm not sure we should.
@rolandlo did you put that in on purpose or was it a leftover of the original PR you didn't really care about?

### DIFF
--- a/src/core/gui/dialog/LinkDialog.cpp
+++ b/src/core/gui/dialog/LinkDialog.cpp
@@ -3,6 +3,7 @@
 #include "control/Control.h"           // for Control
 #include "gui/Builder.h"               // for Builder
 #include "gui/GladeSearchpath.h"       // for GladeSearchPath
+#include "util/glib_casts.h"           // for wrap_for_g_callback_v
 #include "util/raii/CStringWrapper.h"  // for OwnedCString
 
 constexpr auto UI_FILE = "linkDialog.glade";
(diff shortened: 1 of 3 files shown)


src/core/gui/dialog/LinkDialog.cpp
```diff
@@ -45,15 +39,18 @@ LinkDialog::LinkDialog(Control* control, std::function<void(LinkDialog*)> callba
...
 
+#if GTK_MAJOR_VERSION == 3
     g_signal_connect(G_OBJECT(layoutLeft), "released", G_CALLBACK(layoutToogledLeft), this);
     g_signal_connect(G_OBJECT(layoutCenter), "released", G_CALLBACK(layoutToogledCenter), this);
     g_signal_connect(G_OBJECT(layoutRight), "released", G_CALLBACK(layoutToogledRight), this);
+#else
+    g_signal_connect(G_OBJECT(layoutLeft), "clicked", G_CALLBACK(layoutToogledLeft), this);
+    g_signal_connect(G_OBJECT(layoutCenter), "clicked", G_CALLBACK(layoutToogledCenter), this);
+    g_signal_connect(G_OBJECT(layoutRight), "clicked", G_CALLBACK(layoutToogledRight), this);
+#endif
```
rolandlo, line 52: I'm confused about this now. Why does the GTK 4 way (connecting to "clicked" signal) not work with GTK 3? I get a segfault when trying to use the "clicked" (or "toggled") signal with Gtk 3 as soon as I click on one of these toogle buttons. Would it also segfault with GTK 4 or not?
I've also noticed a little typo (that I had missed previously): Toogled should be Toggled.

rolandlo:
> I didn't notice while reviewing #6841, but some things are incompatible with GTK4. This PR fixes that. 

Thanks for cleaning that up. I didn't notice either.

There is one thing I removed for simplicity: the window was getting resized as the user was typing in the GtkTextView. There may be a way to port it but I'm not sure we should. @rolandlo did you put that in on purpose or was it a leftover of the original PR you didn't really care about?

It was a left over I didn't care about. I'm fine with having the resizing removed.

src/core/gui/dialog/LinkDialog.cpp
bhennion: Because of the difference between
https://docs.gtk.org/gtk3/method.ToggleButton.set_active.html
and 
https://docs.gtk.org/gtk4/method.ToggleButton.set_active.html

In Gtk3, `gtk_toggle_button_set_active()` triggers the GtkbButton::clicked signal, leading to an infinite loop (well, a stack overflow rather). Not in Gtk4.

rolandlo (approved):
LGTM then. Thanks for the explanation.

bhennion:
Thanks for the review! Squashing and merging in 24 hours unless an objection is raised.

PR #7594, opened 2026-07-08
Add alignment and justify to Text elements

bhennion:
This PR builds on #7553 to add text alignment and justification.

For now, the alignment buttons are only available as toolbar buttons (I added them to the "All In" toolbar for now, if you want to test), but we probably should discuss UI a bit further on this.
Possible UI choices include:

1. A popover like in #4938, above the Text element
2. Having those buttons in the toolbar, like for every other tool. Things like #7582 would make this solution better, as it avoids cluttering the toolbars.

I am in favour of 2., mainly because for large Text elements, placing the popover is not a trivial thing: it could end up outside the view, or we'd have to move it and it could hide part of the text.
[Note that I would still be in favour of a solution like 1. for more localized changes which only affect the *selected* text (like in #4938)]


bhennion:
I notice an issue with this, in case the wrapping is too small for one word to fit and we set align to the right:
Capture d&#39;écran_20260711_104401
I'll try to fix it (but I'm not sure what the expected behaviour should be)

bhennion:
> I notice an issue with this, in case the wrapping is too small for one word to fit and we set align to the right. I'll try to fix it (but I'm not sure what the expected behaviour should be)

I think I got to the right behaviour, but it required more changes than I had hoped:

[Enregistrement d'écran_20260713_102132.webm](https://github.com/user-attachments/assets/96687937-8e5c-45a9-8160-cf4365e699cf)

The problem is that, in situations like this, the natural origin of the Text element is not the upper-left corner of the bounding box, so is not given by `{Element::getX(), Element::getY()}`:
Capture d&#39;écran_20260713_102844

To get things to work, I had to remove the Element::setX/setY functions to a `virtual void setOrigin(double x, double y);`. The getters Element::getX/getY have two different semantics in the code: either we want the element's position (replaced by getOrigin()) or we want the bounding box (left as is, but should be replaced by Element::boundingRect() all over in the long run, I think).

For review, I'll try to split the semantics `getX()/getY()` becoms `getOrigin()` or `boundingRect()` in a prequel PR.

bhennion:
> For review, I'll try to split the semantics `getX()/getY()` becoms `getOrigin()` or `boundingRect()` in a prequel PR.

Done in #7613.

src/core/control/tools/TextEditor.cpp
```diff
@@ -330,6 +332,18 @@ void TextEditor::setFont(XojFont font) {
...
+void TextEditor::setAlignment(TextAlignment al) {
+    this->textElement->setAlignment(al);
+    this->layoutStatus = LayoutStatus::NEEDS_PARAMETERS_UPDATE;
+    repaintEditor(true);  // The size may change if the text overflows
+}
+
+void TextEditor::setJustify(bool justify) {
+    this->textElement->setJustify(justify);
+    this->layoutStatus = LayoutStatus::NEEDS_PARAMETERS_UPDATE;
+    repaintEditor(false);
```
rolandlo: It seems to me that the size can change slightly when justifying/unjustifying text. 
```suggestion
    repaintEditor(true);
```

src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -198,6 +198,12 @@ void SaveHandler::visitLayer(XmlNode* page, const Layer* l) {
             if (auto w = t->getWrap(); w != Text::NO_WRAP) {
                 text->setAttrib(xoj::xml_attrs::WRAP_STR, w);
             }
+            if (auto al = t->getAlign(); al != TextAlignment::LEFT) {
+                text->setAttrib(xoj::xml_attrs::ALIGN_STR, al);
```
rolandlo: As in the link case this attribute should be saved as a string. See also `parseEnum` in `XmlParserHelper.h`.

src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -198,6 +198,12 @@ void SaveHandler::visitLayer(XmlNode* page, const Layer* l) {
...
+            }
+            if (t->getJustify()) {
+                text->setAttrib(xoj::xml_attrs::JUSTIFY_STR, 1);
```
rolandlo: Maybe save it as "true".

src/core/control/xojfile/LoadHandler.cpp
```diff
@@ -361,6 +361,8 @@ void LoadHandler::addText(std::string font, double size, double x, double y, Col
     this->text->setOrigin(x, y);
     this->text->setColor(color);
     this->text->setWrap(wrap.value_or(Text::NO_WRAP));
+    this->text->setAlignment(align.value_or(TextAlignment::LEFT));
+    this->text->setJustify(justify);
```
rolandlo, line 365: The LoadHandlerTest should account for these new attributes as well.

src/core/model/Text.cpp
```diff
@@ -68,15 +70,53 @@ void Text::setWrap(double wrap) {
...
     auto layout = createPangoLayout();
     pango_layout_set_text(layout.get(), this->text.c_str(), static_cast<int>(this->text.length()));
-    int w = 0;
-    int h = 0;
-    pango_layout_get_size(layout.get(), &w, &h);
-    this->boundingBox.width = (static_cast<double>(w)) / PANGO_SCALE;
-    this->boundingBox.height = (static_cast<double>(h)) / PANGO_SCALE;
-    this->updateSnapping();
+
+    auto boxes = computeBoxesForLayout(layout.get(), {this->snappedBounds.x, this->snappedBounds.y}, this->wrapWidth);
```
rolandlo:
```suggestion
    auto boxes = computeBoxesForLayout(layout.get(), this->getOrigin(), this->wrapWidth);
```

rolandlo (changes requested):
I'm fine with having the buttons for alignment and justify in the toolbar. That's something we can change later in case we have a better idea about the ideal UI. You will probably want to add those buttons to other toolbars as well.

src/core/control/tools/TextEditor.cpp
bhennion: I don't see how. Did you experience a case like that?
In any case, we can apply a "better-safe-than-sorry" policy here, as this operation is a rare one anyway.

src/core/control/xojfile/LoadHandler.cpp
bhennion: Done

src/core/control/tools/TextEditor.cpp
rolandlo: Yes, I experienced cases like that with wrapped text.

bhennion: I put a true there

bhennion:
> You will probably want to add those buttons to other toolbars as well.

I added them to every toolbar with a SELECT_FONT button (right before that button).

rolandlo (approved):
LGTM

bhennion:
Thanks for the review!
Merging

PR #7606, opened 2026-07-13
LinkAlignment moved to model/TextAlignment.h

bhennion:
Prequel to #7594. Just a renaming PR, so merging once the CI clears

### DIFF
--- a/src/core/control/xojfile/SaveHandler.cpp
+++ b/src/core/control/xojfile/SaveHandler.cpp
@@ -227,7 +227,7 @@ void SaveHandler::visitLayer(XmlNode* page, const Layer* l) {
 
             const XojFont& f = l->getFont();
 
-            link->setAttrib(xoj::xml_attrs::ALIGN_STR, LinkAlignment::NAMES[l->getAlignment()]);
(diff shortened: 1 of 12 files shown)


src/core/control/xojfile/SaveHandler.cpp
```diff
@@ -227,7 +227,7 @@ void SaveHandler::visitLayer(XmlNode* page, const Layer* l) {
 
             const XojFont& f = l->getFont();
 
-            link->setAttrib(xoj::xml_attrs::ALIGN_STR, LinkAlignment::NAMES[l->getAlignment()]);
+            link->setAttrib(xoj::xml_attrs::ALIGN_STR, l->getAlignment());
```
rolandlo, line 230: There was a change in how the alignment attribute got saved (like `<link align="left" ` became `<link align="0"`. I think it should be reverted, since it's not compatible with loading and with `test/files/load/links.xml`.

Also note that capStyle is also saved as a string, not as the enum value.

bhennion: You're right. Sorry about that. I thought that `setAttrib()` was relying on `XmlParserHelper::has_names<>()` to do this automatically but it doesn't. I think it should, but that's for another day. I'll revert this.

PR #7613, opened 2026-07-16
Refactor Element position

bhennion:
This PR cleans up the position related functions in Element:

1. Element::getX/getY/setX/setY/getElementWidth/getElementHeight are removed
2. In most places, the getters are replaced by the more efficient Element::getBoundingBox() (renamed from boundingRect())
3. In the few places where the intent is to get the position of the element (rather than its "ink" bounding box), getX()/getY() are replaced by a getOrigin()
4. The setters setX/setY are replaced by a setOrigin()

The motivation for this lies in #7594 (see this [comment](https://github.com/xournalpp/xournalpp/pull/7594#issuecomment-4956079770)): there is no reason a priori for an element's positioning origin to be the same as its ink bounding box's upper-left corner. In fact the setters setX/setY do not really make sense (and certainly don't work as expected) for Stroke already.

Another example where this distinction will become useful: if and when we implement object rotation, then the origin of (say) an Image should be the upper-left corner (or maybe the center) of the unrotated image, while the ink bounding box is much larger.

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -250,11 +250,11 @@ auto LatexController::loadRendered(string renderedTex) -> std::unique_ptr<TexIma
         return nullptr;
     }
 
-    img->setX(posx);
-    img->setY(posy);
+    img->setOrigin(posx, posy);
     img->setText(std::move(renderedTex));
     if (std::abs(imgheight) > 1024 * std::numeric_limits<double>::epsilon()) {
-        double ratio = img->getElementWidth() / img->getElementHeight();
+        const auto& box = img->getBoundingBox();
+        double ratio = box.width / box.height;
         if (ratio == 0) {
             img->setWidth(imgwidth == 0 ? 10 : imgwidth);
         } else {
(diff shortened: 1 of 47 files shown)


src/core/control/LatexController.cpp
```diff
@@ -250,11 +250,10 @@ auto LatexController::loadRendered(string renderedTex) -> std::unique_ptr<TexIma
         return nullptr;
     }
 
-    img->setX(posx);
-    img->setY(posy);
+    img->setOrigin(posx, posy);
     img->setText(std::move(renderedTex));
     if (std::abs(imgheight) > 1024 * std::numeric_limits<double>::epsilon()) {
-        double ratio = img->getElementWidth() / img->getElementHeight();
+        double ratio = img->getBoundingBox().width / img->getBoundingBox().height;
```
rolandlo: I would prefer it this way, but it's of course a minor detail.
```suggestion
        const auto& box = img->getBoundingBox(); 
        double ratio = box.width / box.height;
```

src/core/view/TexImageView.cpp
```diff
@@ -35,11 +37,12 @@ void TexImageView::draw(const Context& ctx) const {
         double pageHeight = 0;
         poppler_page_get_size(page, &pageWidth, &pageHeight);
 
-        double xFactor = texImage->getElementWidth() / pageWidth;
-        double yFactor = texImage->getElementHeight() / pageHeight;
+        const auto& box = texImage->getBoundingBox();
```
rolandlo: It would look more consistent to move this out of the two conditional branches next to `const auto& origin = texImage->getOrigin();`

rolandlo (approved):
LGTM

bhennion:
Thanks for the review! Merging once the CI clears.

PR #7628, opened 2026-07-23
Add lua api for url links

rolandlo:
Added "wrap" field to app.addTexts and app.getTexts
Added Lua PAI app.getLinks and app.addLinks

~This PR will need rebasing after #7613 is merged.~ Done!

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -50,6 +50,7 @@
 #include "model/Element.h"
 #include "model/Font.h"
 #include "model/Image.h"
+#include "model/Link.h"
 #include "model/SplineSegment.h"
 #include "model/Stroke.h"
 #include "model/StrokeStyle.h"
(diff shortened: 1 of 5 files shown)


src/core/plugin/luapi_application.h
```diff
@@ -1654,21 +1661,23 @@ static int applib_addTexts(lua_State* L) {
  *     color = 0x1259b9,
  *     x = 127.0,
  *     y = 70.0,
+ *     wrap = -1.0, -- No wrapping
```
bhennion: Is there a way to handle optional return values more explicitly? Could the plugin code test whether the `wrap` field is present? If so, I think we should test internally if `wrap == Text::NO_WRAP` rather than export the internal value.

src/core/plugin/luapi_application.h
```diff
@@ -1755,6 +1767,326 @@ static int applib_getTexts(lua_State* L) {
...
+ *     x = 150.0,
+ *     y = 50.0,
+ *   },
+ * }
+ */
+static int applib_addLinks(lua_State* L) {
+    Plugin* plugin = Plugin::getPluginFromLua(L);
+    Control* control = plugin->getControl();
+    PageRef const& page = control->getCurrentPage();
+    Layer* layer = page->getSelectedLayer();
```
bhennion: It would be safer if those were only done when the mutex is locked. Ideally, I guess all the links should be added together, in a single mutex lock (where those infos are fetched).

src/core/plugin/luapi_application.h
```diff
@@ -1755,6 +1767,326 @@ static int applib_getTexts(lua_State* L) {
...
+            return luaL_error(L, "'color' must be an integer/hex-code or unset");
+        }
+
+        if (!lua_isnumber(L, -2)) {  // Check if x was provided
+            return luaL_error(L, "Missing X-Coordinate!/must be a number");
+        }
+        if (!lua_isnumber(L, -1)) {  // Check if y was provided
+            return luaL_error(L, "Missing Y-Coordinate!/must be a number");
+        }
+        link->setOrigin(lua_tonumber(L, -2), lua_tonumber(L, -1));
```
bhennion, line 1925: Just to be sure: do the calls to `lua_tonumber()` independent of their calling order? I.e. are the code snippets
```
auto a = lua_tonumber(L, -2);
auto b = lua_tonumber(L, -1);
```
and
```
auto b = lua_tonumber(L, -1);
auto a = lua_tonumber(L, -2);
```
equivalent?

src/core/plugin/luapi_application.h
```diff
@@ -1755,6 +1767,326 @@ static int applib_getTexts(lua_State* L) {
...
+            lua_setfield(L, -2, "layer");  // add layer to link
+        }
+
+        if (page_nb.has_value()) {
+            lua_pushinteger(L, as_signed(page_nb.value()));
+            lua_setfield(L, -2, "page");  // add page to link
+        }
+
+        lua_settable(L, -3);  // add link to elements
+    }
```
bhennion, line 2089: We should probably strengthen the Lua API to ensure the mutex is locked (shared lock when reading like this). I know it is not done in the other getters, but it may be a good idea.

src/core/plugin/luapi_application.h
rolandlo: Yes, `lua_tonumber` and similar functions don't change the stack, they just read the values at the given index and try to convert them to the specified type.

src/core/plugin/luapi_application.h
rolandlo: I have added a `shared_lock`.

src/core/plugin/luapi_application.h
rolandlo: `control->getCurrentPage()` already has a shared lock in it. I moved `page->getSelectedLayer` to a place where there is a lock. Is that fine for you?

rolandlo:
Previously I had forgotten to add the Alignment enum to the app constants. This is done now.

bhennion (approved):
LGTM!

src/core/plugin/luapi_application.h
rolandlo: I forgot about this comment. Now the wrap field is only exported if the text is wrapped. It's easy in Lua to test for a value being `nil` or not.

rolandlo:
Merging once the CI clears. Thanks for the review @bhennion!

PR #7631, opened 2026-07-23
Fix applib_changeToolColor for tools whose names are not all lowercase.

### LINKED ISSUE #7623: Plugins can't access certain tools
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

_No response_

### Installation Method

Official Release

### Xournal++ Version

1.3.6

### libgtk Version

3.24.52

### Bug Description

In lua-plugins, the function app.changeToolColor(color, toolname) converts the string for toolname to lowercase.
But many tools are in camelCase like laserPointerPen.
This prevents plugins from selecting those tools.
Either remove the tolowercase (could break some existing plugins),
or remove the case-sensitivity check when searching for the appropriate tool in the tool list.

### Expected Behaviour

The line:
 app.changeToolColor({["color"] = 0xFF0000, ["tool"] = "laserPointerPen"})
should appropriately change the color of laser-pointer to red and NOT throw an error that tool "none" does not exist.

### Steps to Reproduce

write lua-plugin with this line:
app.changeToolColor({["color"] = 0xFF0000, ["tool"] = "laserPointerPen"})
it will throw an error because tool "none" is not found.
on the other hand the line:
app.changeToolColor({["color"] = 0x000000, ["tool"] = "pen"})
WILL work because there's no camelCase in the tool name "pen".

### Additional Context

The solution shouldn't break existing plugins if possible.
For example just deleting the tolowercase could break plugins that depend on it.
The line:
app.changeToolColor({["color"] = 0x000000, ["tool"] = "PEN"})
works right now, but would stop working when deleting the tolowercase.

rolandlo:
Fixes #7623

### DIFF
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -1940,7 +1940,10 @@ static int applib_changeToolColor(lua_State* L) {
     ToolType toolType = toolHandler->getToolType();
     const char* toolStr = luaL_optstring(L, -2, nullptr);
     if (toolStr != nullptr) {
-        toolType = toolTypeFromString(StringUtils::toLowerCase(toolStr));
+        toolType = toolTypeFromString(toolStr);
+        if (toolType == TOOL_NONE) {  // for backwards compatibility
+            toolType = toolTypeFromString(StringUtils::toLowerCase(toolStr));
+        }
     }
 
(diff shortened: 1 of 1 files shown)


bhennion (approved):
LGTM

rolandlo:
Thanks for the review @bhennion!

Issue #2900, opened 2021-03-03
Discrepency between "highlighter" and "fill" on dark backgrounds


392781:
A picture is a thousand words, so let's start with that:
![image](https://user-images.githubusercontent.com/25296775/109775973-fa632d00-7bb6-11eb-8c88-a1d950313788.png)

As you can see, I have a dark background set up for my Xournal page.  When I use the highlighting function on this dark background, it is effectively useless highlighting.  However, the fill option produces the result that highlighting should... and this words regardless of light or dark background.

**My proposal:** Make highlighting work the way fill does (drawing over the background instead of blending into it like it currently does).

Let me know if you need additional information.

personalizedrefrigerator:
See also #2471.

personalizedrefrigerator:
It looks like the issue is that we're using `CAIRO_OPERATOR_MULTIPLY` for highlighters and `CAIRO_OPERATOR_OVER` for filling semi-transparent strokes.

https://github.com/xournalpp/xournalpp/blob/5d2683662f560b5254d52973b9f4281036e43e24/src/view/StrokeView.cpp#L56-L59

https://github.com/xournalpp/xournalpp/blob/5d2683662f560b5254d52973b9f4281036e43e24/src/control/tools/StrokeHandler.cpp#L42-L44

A list of operators can be found here:
https://www.cairographics.org/operators/#clear

While we could change this to `OPERATOR_OVER` (which would simplify logic -- we wouldn't need a special case for the highlighter), it **might** make highlighted text less readable on a light background. 

As such, I propose we use `CAIRO_OPERATOR_OVER` on dark backgrounds and `CAIRO_OPERATOR_MULTIPLY` on light backgrounds.

392781:
I mean Xournal++ also allows to lower opacity of highlights anyway.  You could just use the `OVER` option everywhere and lower default opacity by a notch.

mjg:
If you compare xournal's (legacy) rendering of the highlighter to xournalpp's you can see the improvement. (I even re-rendered legacy xojs using xournalpp to improve legibility.) It directly comes from the different oprator. In fact, up until now I thought that xournalpp cleverly z-orders the highlighter to draw directly on top of the background (below other strokes) instead of on top of the stroke.

[...]

PR #7637, opened 2026-07-28
Fix(model): Reset backgroundColor to white when changing page type to image or PDF (i.e. special)

### LINKED ISSUE #7635: Background Color of the page not reset when changing background type (e.g. from plain to image or to PDF)
### Operating System

Linux

### (Linux only) Distribution

Void Linux

Sway wm

Wayland

### Installation Method

Official Relaes, and locally built

### Xournal++ Version

1.3.4 for the official relaes, 1.3.6 for the built one

### libgtk Version

3.24.52

### Bug Description

When changing a page's background type using XojPage::setBackgroundType(), the stored backgroundColor attribute is not reset or synchronized with the new background state. If a page had a dark background color and then the user changed it into an Image or PDF background XojPage::getBackgroundColor() continues to return the old color.

## Why this matters?
This makes handling dynamic rendering inconsistent (e.g. making highlighter work on dark and light pages #2900), I am currently working on a fix for the highlighter issue that dynamicaly sets the cairo operator used to multiply or screen based on the page luminance according to the standard ITU-R BT.709 luminance formula

### Expected Behaviour

When changing the background type to an image or PDF, background color should either be
-  reset to the default (Colors::white)
- or changed to a a light color (e.g. white) if the image is mostly light or a dark color (e.g. black) if the image is mostly black
  - However, I don't know how the second option can be implemented and I will go with the first. (Open for suggestions though)

### Steps to Reproduce

1. Open a page with a non-default background color (e.g. dark or colored pages)
2. Change the background type to an Image (or PDF).
3. Inspect XojPage::getBackgroundColor() programmatically 

or 
1. Build xounralpp from my fork with the commit https://github.com/Andrew-Elia/xournalpp/commit/d602cb8f264c2f576a14d408dfb44d0af03209d9 in branch [fix/2900-fixing-highlighter-on-dark-backgrounds](https://github.com/Andrew-Elia/xournalpp/tree/fix/2900-fixing-highlighter-on-dark-backgrounds)
2. open a page with a dark background
3. make some highlights (highlighter now uses screen operator and dark highlights works)
4. Change background type to an image
5. The highlights don't appear as they should on an image (using operator multiply), as the backgroundColor wasn't updated.

### Additional Context

- **Status:** I am currently working on a fix for this issue.
- **Impact:** Fixing this ensures accurate background state reporting, which is critical for dynamic rendering behaviors (e.g., highlighter operator blending over light vs. dark backgrounds (related to issue #2900 )).
- I will submit a Pull Request shortly with the fix.

**Worth Noting:**
- the background color for journals with background type as image aren't saved in the xopp file (when reloaded the background color is assumed white)
- when annotating PDF the background color is assumed white for PDF pages, for manually added pages it is set as the default color and when changing background to "with PDF background" the same phenomenon as with image happens (background color not reset) but is also reset on reloading the xopp file

Andrew-Elia:
# Summary
This PR fixes an issue where switching a page's background type to an Image or Pdf via `XojPage::setBackgroundType()` did not update or reset the internal backgroundColor attribute.

If a page previously had a custom or dark background color, XojPage::getBackgroundColor() would continue returning the old color in memory even though .xopp file reloads assumes white paper background for Image and PDF pages.

## Why This Is Needed
Dynamic rendering systems that inspect `XojPage::getBackgroundColor()` to adjust dynamic draw parameters (e.g., dynamic highlighter cairo operator selection based on page luminance) receive stale color data when users switch background types dynamically during an active session.

## Changes Made:
- Updated `XojPage::setBackgroundType()` to reset `backgroundColor` to `Colors::white` whenever the target `bgType` is special (i.e., Image or Pdf).
- Added a unit test to verify backgroundColor state transitions in XojPage.

## **How Has This Been Tested?**

1. **Unit Tests:** Added GoogleTest coverage verifying that `XojPage::setBackgroundType()` resets `backgroundColor` to `Colors::white` when transitioning from a dark background to an Image or PDF.

2. **Manual Testing:**
* Created a page with a dark background.
* Switched background type to an Image.
* Verified visually changing the page type to plain creates a page with white background.
* Verified that saving and reopening `.xopp` files remains consistent.

## **Worth Noting**:
- Whenever page type is changed to a special type the background color is changed to white, thus if the page type was later changed to a non-special type (e.g.) plain, it is a plain page with *white* background not the user specified default background color.
- The unit test code was generated using Gemini Flash 3.6 Extended thinking.
- **Related Issues**: Closes #7635 and prevents a corner case failure in the upcoming dynamic highlighter operator for light and dark backgrounds fix (#2900), specifically when transitioning a dark background page to a special background type.

### DIFF
--- a/src/core/model/XojPage.cpp
+++ b/src/core/model/XojPage.cpp
@@ -150,6 +150,9 @@ void XojPage::setBackgroundType(const PageType& bgType) {
     if (!bgType.isImagePage()) {
         this->backgroundImage.free();
     }
+    if (bgType.isSpecial()) {
+        this->backgroundColor = Colors::white;
+    }
 }
 
 auto XojPage::getBackgroundType() const -> PageType { return this->bgType; }
--- a/test/unit_tests/model/XojPageTest.cpp
+++ b/test/unit_tests/model/XojPageTest.cpp
@@ -1,0 +1,28 @@
+#include <gtest/gtest.h>
+
+#include "model/PageType.h"
+#include "model/XojPage.h"
+#include "util/Color.h"
+
+TEST(XojPageTest, SetBackgroundTypeResetsColorForSpecialBackgrounds) {
+    XojPage page(595.0, 842.0);  // Standard page dimensions
+
+    // 1. Set a dark background color initially
+    page.setBackgroundColor(Color(0x1E1E1E));
+    EXPECT_EQ(page.getBackgroundColor(), Color(0x1E1E1E));
+
+    // 2. Transition background type to Image
+    PageType imageBg(PageTypeFormat::Image);
+    page.setBackgroundType(imageBg);
+
+    // 3. Verify color is reset to default white
+    EXPECT_EQ(page.getBackgroundColor(), Colors::white);
+
+    // 4. Test PDF background transition as well
+    page.setBackgroundColor(Color(0x000000));
+    PageType pdfBg(PageTypeFormat::Pdf);
+    page.setBackgroundType(pdfBg);
+
+    // 5. Verify color is reset to white again
+    EXPECT_EQ(page.getBackgroundColor(), Colors::white);
+}


bhennion:
I don't think I agree with the diagnosis here. There is no expectation for (say) an image background to even return something on `getBackgroundColor()`: it just doesn't really make sense.
Note that in the code, `getBackgroundColor()` is actually only called with normal backgrounds where it makes sense.

I agree that the way it is currently coded is not great. Ideally, `getBackgroundColor()` should return an optional, only filled when the background is not `isSpecial()`.

Andrew-Elia:
@bhennion I agree with you that `getBackgroundColor()` shouldn't even be called in the first place for special page types.

However I do think that — at least for now, until a better implementation is proposed — the page color should be reset to white whenever the page type is changed to a special one. (which is the proposed change)

Regarding #2900 the proposed solution in #7649 works well. However in the case of changing the page type from a (plain, ruled, dotted, etc) to special body type, the net solution (without resetting the background color) determines the selector based on the backgroundColor of the page before changing its type until the file is reloaded or xournalpp is restarted.
That is how I determined the issue and came to the idea of resetting the page color to white whenever the page type is changed to a special one.

I am open to suggestions regarding if there are better ways to implement this solution or fix the problem. Or if decided that it isn't an important case and should be ignored.

Andrew-Elia:
Hello @bhennion, I thought of a way to prevent that issue from happening (highlighter operator chooses SCREEN if changing from a dark background to special page type during runtime) without doing this PR. Which is modifying the operator selection code in #7649 to include a conditional for special page types and always setting them to MULTIPLY. However, I believe this solution includes much more runtime complexity and time consumption as the isSpecial test will run each and every time a new highlighter stroke is drawn compared to resetting page color to white when and only when the page type changes to special. I personally prefer the latter. 
If the latter implementation is impossible I would also just ignore this case and encourage reloading the file than increasing the complexity and operations needed for each stroke.

I would love to know your opinion.

bhennion:
> However, I believe this solution includes much more runtime complexity and time consumption as the isSpecial test will run each and every time a new highlighter stroke is drawn compared to resetting page color to white when and only when the page type changes to special

I don't think that would actually be more expensive. `isSpecial()` is a simple test, while when setting the color to white, you are still computing the color's luminance (of white...) which does have a cost as well.

PR #7638, opened 2026-07-28
feature: preserve hues for dark mode colour inversion

owl10124:
As mentioned in several comments of https://github.com/xournalpp/xournalpp/pull/6090, the current method for applying custom light/dark colours inverts hues as well as brightness, which causes readability / specific colour use to be lost.

This tiny PR adds an optional setting to (approximately) **preserve the hues**, by naïvely inverting them (using Cairo's HSL_HUE operator on an inverted copy) before the usual processing.
It's a bit janky + laggy -- there's probably a better way to do this -- but it does the job.

image
image

Hopefully I didn't break anything :P

(Note -- this doesn't quite work / **inverts the colours instead** on light backgrounds, but the default setting should work fine for those.)

### DIFF
--- a/src/core/control/settings/Settings.cpp
+++ b/src/core/control/settings/Settings.cpp
@@ -203,7 +203,7 @@ void Settings::loadDefault() {
     this->selectionMarkerColor = Colors::xopp_cornflowerblue;
     this->activeSelectionColor = Colors::lawngreen;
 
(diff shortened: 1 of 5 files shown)


bhennion (changes requested):
I think the feature is good (we could also drop the existing recoloring mode if this one worked well), but we cannot merge something that doesn't really work and lags.
If you find a way to achieve this recoloring without creating an extra cairo_surface_t (don't know if this is possible), you could get similar performances and this could get merged.

Issue #7258, opened 2026-03-11
Scroll the page using a stylus


zhoujl2005:
NEED：Hope to add the ability to use a stylus to swipe outside the page area to scroll the page up and down, achieving the function of a mouse wheel.

<img width="1280" height="717" alt="Image" src="https://github.com/user-attachments/assets/464bdd77-a8e4-4e1b-b4c2-7bd138603fae" />

youssefadly237:
I want to confirm that I understand what you mean

using stylus outside page area should turn tool into hand and scroll?

XsavciX:
"I agree with the OP. Xournal++ should definitely support auto-scrolling while drawing with a stylus. I use a Wacom Slate (A4 size), but the A4 page setting in Xournal++ doesn't fit my screen vertically. When I reach the bottom of the physical tablet, the software stops registering the stroke because the remaining part of the A4 page is off-screen. If an auto-scroll feature were implemented, the digital canvas would stay in sync with my tablet's physical dimensions.

Sorry for any mistakes in my English, I hope my point is clear!"

PR #7656, opened 2026-08-03
Make pen strokes starting outside page area act like hand tool

wishawa:
Implements what is requested in #7258.

I take lectures with Xournal++ and they get to dozens of pages long, at which point using the scrollbar is not precise enough. Being able to scroll with the stylus is extremely useful!

Tested:

[Screencast.webm](https://github.com/user-attachments/assets/dd154164-8e57-4988-a9b1-3285694a957c)

Also, I haven't set up the compiler environment properly (I used the flatpak build pipeline to build and test this), so I haven't run the formatter, linter, etc. I hope this is not a problem since the changes are very simple.

### DIFF
--- a/src/core/control/ToolHandler.cpp
+++ b/src/core/control/ToolHandler.cpp
@@ -571,6 +571,13 @@ bool ToolHandler::pointActiveToolToButtonTool(Button button) {
     return true;
 }
 
+bool ToolHandler::pointActiveToolToToolType(ToolType type) {
+    if (this->activeTool->type == type)
+        return false;
+    this->activeTool = &getTool(type);
+    return true;
+}
+
 bool ToolHandler::pointActiveToolToToolbarTool() {
     if (this->activeTool == this->toolbarSelectedTool)
         return false;
--- a/src/core/gui/inputdevices/PenInputHandler.cpp
+++ b/src/core/gui/inputdevices/PenInputHandler.cpp
@@ -94,6 +94,8 @@ auto PenInputHandler::actionStart(InputEvent const& event) -> bool {
     if (!changeTool(event))
         return false;
 
+    this->inputStartedOutsidePageArea = currentPage == nullptr;
+
     // Used for pressure inference
     this->lastPressure = 0.0;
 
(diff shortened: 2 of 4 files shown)


wishawa:
Note that I have only tested this with mouse, touchpad, and drawing tablet. I believe it shouldn't do anything with touchscreen inputs, but I don't really understand how touchscreen input handling works in Xournal++, and I don't have a touchscreen device to test.

If you have a touchscreen device and can test this, please help test it!

ad2004teaster-cmyk:
AMAZING!

bhennion:
That's a nice feature and I think we can add it. The implementation however looks a bit weird and I suspect it could have unexpected consequences.

I would be more confident if the implementation actually changed the active tool (so that other parts of the app would not be "surprised" when fetching the active tool while an action is happening). I think it could be done with a simple
```
        toolHandler->selectTool(TOOL_HAND);
        toolHandler->fireToolChanged();
```
in PenInputHandler::actionStart() (but I did not test this).

wishawa:
I changed it to actually switch to the hand tool as you suggested. The implementation is a bit more subtle though.

I tried the simple approach (backup the current tool, `selectTool(TOOL_HAND)`, and restore the backed-up tool at the end of the action). The problem is if the user drag outside the page with their button tool, then we'll mess up their selected toolbar tool.
And AFAICT there is no public way to determine if the input is made with button or toolbar tool.

The approach I just pushed require adding a new method to the ToolHandler to allow changing the active tool without changing the selected toolbar tool. I use this to set the hand tool. Please let me know what you think!

src/core/control/ToolHandler.cpp
```diff
@@ -598,6 +571,13 @@ bool ToolHandler::pointActiveToolToButtonTool(Button button) {
     return true;
 }
 
+bool ToolHandler::pointActiveToolToToolType(ToolType type) {
+    if (this->activeTool->type == type)
+        return false;
```
bhennion, line 576:
```suggestion
    if (this->activeTool->type == type) {
        return false;
    }
```

src/core/gui/inputdevices/PenInputHandler.cpp
```diff
@@ -290,7 +296,10 @@ auto PenInputHandler::actionMotion(InputEvent const& event) -> bool {
     GtkXournal* xournal = this->inputContext->getXournal();
     ToolHandler* toolHandler = this->inputContext->getToolHandler();
 
-    this->changeTool(event);
+    // If this is a drag starting outside the page area, we'd have set active tool to HAND.
+    // In this case we don't want changeTool to change it back to the toolbar / button tool.
+    if (!this->inputStartedOutsidePageArea)
+        this->changeTool(event);
```
bhennion, line 302:
```suggestion
    if (!this->inputStartedOutsidePageArea) {
        this->changeTool(event);
    }
```

bhennion (commented):
Just a couple of nitpick, to fit our coding conventions

bhennion:
Also, could you apply clang format to you patch (e.g. running `git clang-format master` from your repo)

wishawa:
I ran clang format and also fixed the convention as you suggested!

bhennion (approved):
LGTM! Thanks a lot. Merging in 24 hours unless an objection is raised.

PR #7673, opened 2026-08-08
Fix horizontal arrow scrolling with hidden scrollbar

### LINKED ISSUE #7647: Horizontal page scrolling using the keyboard
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Exe

### Xournal++ Version

1.3.5+dev

### libgtk Version

3.24.52

### Bug Description

Horizontal page scrolling using the keyboard arrow keys is not possible if "Hide horizontal scrollbar" is enabled in "Preferences".

### Expected Behaviour

Horizontal page scrolling using the keyboard arrow keys may be possible even if "Hide horizontal scrollbar" is enabled in "Preferences".

### Steps to Reproduce

"Hide horizontal scrollbar" is enabled in "Preferences".
Scroll left or right with keyboard

### Additional Context

_No response_

bhennion:
Fix #7647

The issue was: to optimize calls to `Layout::updateVisibility()` when moving or zooming, the call was blocked in the callback signaling horizontal scrolling. When using the arrow to scroll horizontally, the vertical scroll value does not change so the callback for vertical scrolling is not called, and `Layout::updateVisibility()` (as well as the attending `gtk_widget_queue_draw()`) was never called.
For some reason, when the horizontal scrollbar is visible, GTK secretely calls `gtk_widget_queue_draw()`, so we did not see the problem. Still, even in that case, we need to actually call `Layout::updateVisibility()` so adding a `gtk_widget_queue_draw()` would not be enough.

### DIFF
--- a/src/core/gui/Layout.cpp
+++ b/src/core/gui/Layout.cpp
@@ -38,27 +38,32 @@ constexpr auto const XOURNAL_PADDING_BETWEEN = 15;
 
 
 Layout::Layout(XournalView* view, ScrollHandling* scrollHandling): view(view), scrollHandling(scrollHandling) {
-    g_signal_connect(scrollHandling->getHorizontal(), "value-changed", G_CALLBACK(horizontalScrollChanged), this);
-    g_signal_connect(scrollHandling->getVertical(), "value-changed", G_CALLBACK(verticalScrollChanged), this);
-
-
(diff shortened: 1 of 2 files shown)


rolandlo (approved):
LGTM

PR #7701, opened 2026-08-19
Use PDF's CropBox instead of MediaBox for QPDF overlays in export

### LINKED ISSUE #7699: Annotations misaligned when exporting to PDF in Archlinux
### Operating System

Linux

### (Linux only) Distribution

Archlinux

XFCE4

X.Org

### Installation Method

Official repo

### Xournal++ Version

xournalpp 1.3.7-1

### libgtk Version

libgtk: 3.24.52

### Bug Description

When trying to export to PDF, all annotations on the new PDF are misaligned, here I provide the original PDF, the xopp modification and the exported PDF

Original + xopp: 

[pdf + xopp.zip](https://github.com/user-attachments/files/31205511/pdf.%2B.xopp.zip)

Output pdf: 

[Output.pdf](https://github.com/user-attachments/files/31205519/Output.pdf)

### Expected Behaviour

correctly aligned text

### Steps to Reproduce

1. open provided xopp
2. select export to pdf
3. open exported pdf

### Additional Context

It looks like it works correctly on windows, at least on the store version.

bhennion:
Fix #7699
The PDF page's CropBox is supposed to be "what is displayed on screen", so I'm assuming that is what Poppler does. I could not find any documentation on this though.

### DIFF
--- a/src/core/pdf/base/QPdfExport.cpp
+++ b/src/core/pdf/base/QPdfExport.cpp
@@ -121,7 +121,7 @@ bool QPdfExport::overlayAndSave(const fs::path& saveDestination, std::stringstre
                     // Generate content to place the form XObject centered within destination page's trim box.
                     QPDFMatrix m;
                     std::string content =
-                            page.placeFormXObject(localXObj, name, page.getMediaBox().getArrayAsRectangle(), m);
+                            page.placeFormXObject(localXObj, name, page.getCropBox().getArrayAsRectangle(), m);
                     if (!content.empty()) {
                         // Append the content to the page's content. Surround the original content with q...Q to
                         // the new content from the page's original content.


bhennion:
Ideally, we'd have a PDF document where all the boxes are different, to make sure. I don't know how to produce such a document though.

bhennion:
[There seems to be an issue with ftpmirror.gnu.org, so the MacOS CI fails to download dependency sources]

rolandlo:
> Ideally, we'd have a PDF document where all the boxes are different, to make sure. I don't know how to produce such a document though.

Turns out this can easily be produced using pdflatex's `pdfpageattr` command, e.g.
```tex
\pdfpageattr{
	/MediaBox [0 0 595 841]     % 1. Physical page (595 x 841)
	/BleedBox [30 30 595 841]   % 2. 30pt inward (535 x 781)
	/CropBox  [60 60 595 841]   % 3. 60pt inward (475 x 721) -> PDF viewers clip here!
	/TrimBox  [90 90 595 841]   % 4. 90pt inward (415 x 661)
	/ArtBox   [120 120 595 841] % 5. 120pt inward (355 x 601)
}
```

To check that it works use `pdfinfo -box filename.pdf`

Here's a sample .tex file, .pdf output, .xopp file with some annotations and exports using Cairo, QPDF old as in master, QPDF new as in this PR:

[boxes_asym.zip](https://github.com/user-attachments/files/31553898/boxes_asym.zip)

With this PR the export works fine. Interestingly the problem didn't occur when the boxes were symmetrically placed on the page. So in the example the boxes are asymmetrically placed.

rolandlo (approved):
Works according to my tests (see previous post).

bhennion:
> Turns out this can easily be produced using pdflatex's `pdfpageattr` command, e.g.

Nice! Thanks!

Merging then. Thanks for the review and testing

PR #7440, opened 2026-05-03
Feat add latex api

mknap:
Provides an `addLatex` API for lua. 

Thank you for xournalpp.  I would love to learn about your review process!

PR #7711, opened 2026-08-26
Add Lua API app.addTexImages

rolandlo:
Follow up on #7440 

Apart from addressing reviewer comments there, the TeX rendering is now happening asynchronously (and simultaneously) like it should (otherwise the whole app would be blocked during rendering, which may take seconds).

There are still some things to be addressed:
- [x] Proper error handling
- [x] Lifetime of the LatexController
- [x] options for undo/redo
- [x] replace function name string by function
- [x] order of TexImages may be different from order provided by the user

Question: The user can specify a callback function in Lua. Currently it runs every time a TexImage has been rendered. Would it be better that the callback only runs after all TexImages have been rendered?

### DIFF
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -206,6 +206,8 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
     if (self->callback) {  // tex formula from plugin
         if (self->isValidTex) {
             self->callback(self->loadRendered(self->initialTex));
+        } else {
+            self->callback(self->texProcessOutput);
         }
         g_clear_object(&proc);
         return;
@@ -382,7 +384,7 @@ void LatexController::insertLatex(PageRef page, Control* ctrl, double x, double 
 static int jobNo = 0;
 
 void LatexController::renderTexImage(Control* ctrl, std::string latex, Color color,
-                                     const std::function<void(std::unique_ptr<TexImage>)>& callback, std::string* errorMessage) {
+                                     const std::function<void(CallbackArg)>& callback, std::string* errorMessage) {
     auto self = std::make_unique<LatexController>(ctrl, jobNo++);
     self->callback = callback;
     self->initialTex = latex;
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -154,10 +154,8 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
     bool procExited = false;
     GSubprocess* proc = G_SUBPROCESS(procObj);
 
-    std::unique_ptr<LatexController> guard(self);
-    if (!self->callback) {
-        guard.release();  // don't delete the LatexController on exit
-    }
+    // When called from a plugin, this function owns *self and will destroy it upon return
+    std::unique_ptr<LatexController> guard(self->callback ? self : nullptr);
 
     // Extract the process' output and store it.
     {
@@ -408,11 +406,6 @@ void LatexController::renderTexImage(Control* ctrl, std::string latex, Color col
 
     auto* proc = std::get<GSubprocess*>(result);
 
-    // Render the TeX and capture the process' output.
-    char* stdinBuff = nullptr;  // No stdin
-
-    LatexController* self_ptr = self.release();
-
-    g_subprocess_communicate_utf8_async(proc, stdinBuff, nullptr,
-                                        reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), self_ptr);
+    // Render the TeX and capture the process' output. The callback takes ownship of *self
+    g_subprocess_communicate_utf8_async(proc, nullptr, nullptr, xoj::util::wrap_v<onPdfRenderComplete>, self.release());
 }
--- a/src/core/control/LatexController.cpp
+++ b/src/core/control/LatexController.cpp
@@ -52,13 +52,11 @@ using std::string;
 constexpr Color LIGHT_PREVIEW_BACKGROUND = Colors::white;
 constexpr Color DARK_PREVIEW_BACKGROUND = Colors::black;
 
-LatexController::LatexController(Control* control):
+LatexController::LatexController(Control* control, int jobNo):
         control(control),
         settings(control->getSettings()->latexSettings),
-        texTmpDir(Util::getTmpDirSubfolder("tex")),
-        generator(settings) {
-    Util::ensureFolderExists(this->texTmpDir);
-}
+        texTmpDir(Util::getTmpDirSubfolder("tex", jobNo)),
+        generator(settings) {}
 
 LatexController::~LatexController() {
     if (updating_cancellable) {
@@ -156,6 +154,11 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
     bool procExited = false;
     GSubprocess* proc = G_SUBPROCESS(procObj);
 
+    std::unique_ptr<LatexController> guard(self);
+    if (!self->callback) {
+        guard.release();  // don't delete the LatexController on exit
+    }
+
     // Extract the process' output and store it.
     {
         char* procStdout_ptr = nullptr;
@@ -198,6 +201,14 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
     if (!self->isValidTex) {
         fs::path pdfPath = self->texTmpDir / "tex.pdf";
         fs::remove(pdfPath);
+    }
+
+    if (self->callback) {  // tex formula from plugin
+        if (self->isValidTex) {
+            self->callback(self->loadRendered(self->initialTex));
+        }
+        g_clear_object(&proc);
+        return;
     }
 
     const string currentTex = self->dlg->getBufferContents();
@@ -366,4 +377,40 @@ void LatexController::insertLatex(PageRef page, Control* ctrl, double x, double 
     lock.unlock();
 
     showTexEditDialog(std::move(self));
+}
+
+static int jobNo = 0;
+
+void LatexController::renderTexImage(Control* ctrl, std::string latex, Color color,
+                                     const std::function<void(std::unique_ptr<TexImage>)>& callback, std::string* errorMessage) {
+    auto self = std::make_unique<LatexController>(ctrl, jobNo++);
+    self->callback = callback;
+    self->initialTex = latex;
+
+    auto depStatus = self->findTexDependencies();
+    if (!depStatus.success) {
+        if (errorMessage) {
+            *errorMessage = depStatus.errorMsg;
+        }
+        return;
+    }
+
+    const std::string texContents = LatexGenerator::templateSub(latex, self->latexTemplate, color);
+    auto result = self->generator.asyncRun(self->texTmpDir, texContents);
+    if (auto* err = std::get_if<LatexGenerator::GenError>(&result)) {
+        if (errorMessage) {
+            *errorMessage = err->message;
+        }
+        return;
+    }
+
+    auto* proc = std::get<GSubprocess*>(result);
+
+    // Render the TeX and capture the process' output.
+    char* stdinBuff = nullptr;  // No stdin
+
+    LatexController* self_ptr = self.release();
+
+    g_subprocess_communicate_utf8_async(proc, stdinBuff, nullptr,
+                                        reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), self_ptr);
 }
--- a/src/core/plugin/luapi_application.h
+++ b/src/core/plugin/luapi_application.h
@@ -1641,11 +1641,10 @@ static int applib_addTexts(lua_State* L) {
  *
  * Global parameters:
  *   - texItems table: array of latex-parameter-tables
- *   - cb string: Name of the callback function run after a TexImage has been added
+ *   - cb function: Callback function run after a TexImage has been added
  *
  * @param opts {textItems:{formula:string, color:integer, x:number, y:number, width:number|nil, height:number|nil}[],
-cb:string}
-}
+ * cb:function}
  *
  * Parameters per texImage:
  *   - formula string: the tex formula (required)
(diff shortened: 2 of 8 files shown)


src/core/control/LatexController.cpp
```diff
@@ -371,3 +395,45 @@ void LatexController::insertLatex(PageRef page, Control* ctrl, double x, double
...
+
+    auto* proc = std::get<GSubprocess*>(result);
+
+    // Render the TeX and capture the process' output.
+    char* stdinBuff = nullptr;  // No stdin
+
+    LatexController* self_ptr = self.release();
+
+    g_subprocess_communicate_utf8_async(proc, stdinBuff, nullptr,
+                                        reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), self_ptr);
```
bhennion: Avoid the UB convertion between function pointers
```suggestion
    g_subprocess_communicate_utf8_async(proc, stdinBuff, nullptr,
                                        xoj::util::wrap_v<onPdfRenderComplete>, self_ptr);
```

src/core/control/LatexController.cpp
```diff
@@ -156,6 +154,11 @@ void LatexController::onPdfRenderComplete(GObject* procObj, GAsyncResult* res, L
     bool procExited = false;
     GSubprocess* proc = G_SUBPROCESS(procObj);
 
+    std::unique_ptr<LatexController> guard(self);
+    if (!self->callback) {
+        guard.release();  // don't delete the LatexController on exit
+    }
```
bhennion: This is a bit weird. A comment would help (and this version might be more readable)
```suggestion
    // When called from a plugin, this function owns *self and will destroy it upon return
    std::unique_ptr<LatexController> guard(self->callback ? self : nullptr);
```

bhennion: Also: what happens if a plugin does not set up a callback?

src/core/control/LatexController.cpp
bhennion: For readability:
```suggestion
    // The callback takes ownership of *self
    g_subprocess_communicate_utf8_async(proc, stdinBuff, nullptr,
                                        reinterpret_cast<GAsyncReadyCallback>(onPdfRenderComplete), self.release());
```

bhennion: For readability and to avoid the UB of function pointer conversions
```suggestion
    // Render the TeX and capture the process' output. The callback takes ownship of *self
    g_subprocess_communicate_utf8_async(proc, nullptr, nullptr, xoj::util::wrap_v<onPdfRenderComplete>, self.release());
```

src/core/control/LatexController.cpp
```diff
@@ -52,13 +52,11 @@ using std::string;
 constexpr Color LIGHT_PREVIEW_BACKGROUND = Colors::white;
 constexpr Color DARK_PREVIEW_BACKGROUND = Colors::black;
 
-LatexController::LatexController(Control* control):
+LatexController::LatexController(Control* control, int jobNo):
         control(control),
         settings(control->getSettings()->latexSettings),
-        texTmpDir(Util::getTmpDirSubfolder("tex")),
-        generator(settings) {
-    Util::ensureFolderExists(this->texTmpDir);
```
bhennion, line 60: Is this no longer needed?

src/core/control/LatexController.cpp
```diff
@@ -252,15 +265,26 @@ auto LatexController::loadRendered(string renderedTex) -> std::unique_ptr<TexIma
...
+        if (box.width == 0) {
             img->setWidth(imgwidth == 0 ? 10 : imgwidth);
         } else {
             img->setWidth(imgheight * ratio);
         }
         img->setHeight(imgheight);
+    } else if (hasWidth) {
+        const double ratio = box.height / box.width;
+        img->setWidth(imgwidth);
+        img->setHeight(box.height == 0 ? 10 : imgwidth * ratio);
```
bhennion: This will conflict with changes I proposed in #7722. I find it odd that the positionning of the TexImage happens here. For the Latex Editor, it means the element is positionned every time it is rendered in the dialog, while it should really only be positionned when the user clicks Ok and the element is added.

Of course, for Plugin's insertion, that's a bit different. I think the awkwardness comes from the weird role of LatexController. Maybe both LatexController::loadRendered and LatexController::onPdfRenderComplete should be split off to a dedicated class (owning a callback), and LatexController would remain for the Latex Editor interaction only, passing to this new class the suitable callback the same way your Plugin API does.
Does this make sense?

If you don't want/have time to do this, it can also stay this way and wait until someone else digs into this code down the road.

src/core/plugin/luapi_application.h
```diff
@@ -1634,6 +1636,206 @@ static int applib_addTexts(lua_State* L) {
...
+                        if (allowUndoRedoAction == "grouped" && index == numItems) {
+                            undo->addUndoAction(std::make_unique<InsertsUndoAction>(page, layer, *texItems.get()));
+                        }
+                        plugin->callFunction(callbackRef, msg, index);
+                        return;
+                    }
+                    auto texImage = std::get<std::unique_ptr<TexImage>>(std::move(arg));
+                    auto* texImagePtr = texImage.get();
+                    texItems->push_back(texImagePtr);
+                    layer->addElement(std::move(texImage));
```
bhennion: The document's mutex should definitively be locked for this operation.

bhennion: To answer your question in the OP, it may be simpler to only call the plugin callback once all the elements are rendered. I don't think it would hurt the plugin API's capabilities, and it would simplifiy this section of code (and allow to only lock the mutex once).

bhennion (changes requested):

src/core/plugin/luapi_application.h
rolandlo: These are all valid arguments. The reason why I'm still not sure about which solution to choose: Rendering a LaTeX formula can take some time. It may be better for performance reasons to allow operations to run immediately after a TexImage has been added instead of waiting for all rendering jobs to have finished. It may also be desired that TexImages show up immediately one after the other (by calling app.refreshPage every time) rather than all of them after a longer time period.

src/core/control/LatexController.cpp
rolandlo: It's now immediately returning a luaL_error complaining about the missing callback/wrong type. Before it would complain when trying to execute the function ('Attempt to call a nil value').

src/core/control/LatexController.cpp
rolandlo: In `Util::getTmpDirSubfolder` it's already ensured that the folder exists, so it wasn't needed.

src/core/control/LatexController.cpp
rolandlo: That would make sense to me. I'll think about it when rebasing after #7722 has been merged.

src/core/plugin/luapi_application.h
rolandlo: Done!

rolandlo:
Thanks for the review @bhennion! I have addressed all comments except the one about running the callback only once and about possibly splitting the LatexController class. Do you think the advantages you mentioned of running the callback only once outweigh the performance gain and flexibility offered by the current solution?

src/core/plugin/luapi_application.h
bhennion: But then we can't ensure the elements are added in the same order they came in, can we? If the second element in the list takes shorter than the first to compile? (I'm assuming compilation is parallel)


> It may also be desired that TexImages show up immediately one after the other (by calling app.refreshPage every time) rather than all of them after a longer time period.

I think if the plugin coder really wants this, they could call the method several times (once per element for instance). Couldn't they?

I guess the real question is: what use cases will the function have? If a plugin wants to add several heavy tex objects, we can assume a fast response.

One thing that may help with the user experience is adding a popup "Compiling Tex elements" dialog (maybe even with a progress bar). I'm thinking of a situation where a Tex compilation takes a while, the user gets frustrated and thinks it didn't work and clicks on the button once more, triggering a second run of the plugin function.

rolandlo: > But then we can't ensure the elements are added in the same order they came in, can we? If the second element in the list takes shorter than the first to compile? (I'm assuming compilation is parallel)

No we can't ensure that, but the callback function gets to know which formula it corresponds to via the second argument (`no`) of the callback function `cb=function(ref_or_msg, no)`. 

> I think if the plugin coder really wants this, they could call the method several times (once per element for instance). Couldn't they?

Yes, they could, but at least it wouldn't be possible to undo all TexImage insertions at once (hitting Ctrl+z for instance).

> One thing that may help with the user experience is adding a popup "Compiling Tex elements" dialog (maybe even with a progress bar). I'm thinking of a situation where a Tex compilation takes a while, the user gets frustrated and thinks it didn't work and clicks on the button once more, triggering a second run of the plugin function.

That is an interesting idea indeed.

bhennion: > Yes, they could, but at least it wouldn't be possible to undo all TexImage insertions at once (hitting Ctrl+z for instance).

That's true, but it makes me think of another issue: with the current implementation, the undo stack is in an invalid state if a first element has been added and until the last one has finished compiling. Hitting Ctrl+Z during this time could lead to UBs and Segfaults, I think. For instance in the following (theoretical, I didn't test it) sequence:

1. Add a page
2. Add two tex elements via plugin, one of which is very long to compile
3. When the first one is added to the new page but before the second one finishes, hit Ctrl+Z.

The page should get removed and noone knows what's going to happen next. I guess a call to the corresponding PageView will SegFault.

I think I'm still for a "wait for everyone" solution. It looks safer...

rolandlo: I have now implemented the "wait for everyone" solution. The TexImages will get inserted on the page/layer that is active at the time they all have been rendered.

src/core/plugin/luapi_application.h
```diff
@@ -1634,6 +1637,239 @@ static int applib_addTexts(lua_State* L) {
...
+    Tool& tool = toolHandler->getTool(TOOL_LATEX);
+    Color default_color = tool.getColor();
+
+    auto texItems = std::make_shared<std::vector<Element const*>>();
+
+    lua_settop(L, 1);
+    luaL_checktype(L, 1, LUA_TTABLE);
+
+    lua_getfield(L, 1, "cb");
+    if (!lua_isfunction(L, -1)) {
```
atticus-sullivan, line 1715: so `nil` (omitting it from the args) is not an acceptable value? We should document this then or change the behavior.

I know you can of course specify a no-op (lua) function if needed here but if this is a common case it might be more convenient to just specify `nil`. I'm not entirely sure on how the usecases will be here.

src/core/plugin/luapi_application.h
```diff
@@ -1634,6 +1637,239 @@ static int applib_addTexts(lua_State* L) {
...
+        //   -7 = current tex-params table
+        //   -6 = formula
+        //   -5 = x
+        //   -4 = y
+        //   -3 = width
+        //   -2 = height
+        //   -1 = color
+
+        if (!lua_isstring(L, -6)) {
+            return luaL_error(L, "Missing formula!/'formula' must be a string");
```
atticus-sullivan, line 1765: The index at which the error occurs would be a valuable information fir debugging. But then adding it to the string would also make it hard to match it (thinking of a `try-catch` similar logic).

But the current behavior is consistent with the other functions in the lua-api, so a change should probably be consistent and be made in a new PR (if we decide on this change)

src/core/plugin/luapi_application.h
```diff
@@ -1634,6 +1637,239 @@ static int applib_addTexts(lua_State* L) {
...
+                std::stringstream msg;
+                msg << "Color 0x" << std::hex << color << " is no valid RGB color.";
+                return luaL_error(L, msg.str().c_str());  // luaL_error does not support %x for hex numbers
+            }
+            col = Color(color | 0xff000000U);
+        } else if (lua_isnil(L, -3)) {
+            col = default_color;
+        } else {
+            return luaL_error(L, "'color' must be an integer/hex-code or unset");
+        }
```
atticus-sullivan, line 1802: this is a reoccurring pattern (now occurs 3 times), so maybe this should be a helper function.

(also the check with `<= 0xffffff` and `|= 0xff000000U` is not present for the `addStrokeHelper`)

But this might be something for a new PR.

src/core/plugin/luapi_application.h
```diff
@@ -1634,6 +1637,239 @@ static int applib_addTexts(lua_State* L) {
...
+                        if (msg.empty()) {
+                            msg = "TexItem could not be added";
+                        }
+                        args->at(index - 1) = msg;
+                    } else {
+                        auto texImage = std::get<std::unique_ptr<TexImage>>(std::move(arg));
+                        texImage->setOrigin(x, y);
+
+                        const auto& box = texImage->getBoundingBox();
+                        if (width != 0 && height != 0) {
```
atticus-sullivan, line 1822: This means setting `width=0` and `height=0` is equivalent to specifying neither (or passing `nil` of course) if I understand it correctly. We should document this behavior.

Also what happens if both are set? Is it harmful if no `texImage->setHeight` nor `texImage->setWidth` gets called? If so, we should forbid this constellation or catch that case.
If we cannot scale disproportionally, we could take the smaller or larger scale (but while this might be useful sometimes, without selecting what should be chose it might be a bit unintuitive). Another option is of course simply forbidding such a combination of arguments.

atticus-sullivan (commented):
tbh I don't know too much about how LaTeX is handled in xournalpp, but I read in the implementation for the lua-api.

src/core/plugin/luapi_application.h
rolandlo: Yes, `nil` is not acceptable. It's hard to think of a case where you don't want to at least refresh the page once all TexImages have been added.

src/core/plugin/luapi_application.h
rolandlo: If both are set you usually get something that doesn't preserve aspect ratio, which may be wanted in rare cases. I don't think we should forbid it. If none is set you get the width and height that the TeX generator produces.

src/core/plugin/luapi_application.h
rolandlo: > > One thing that may help with the user experience is adding a popup "Compiling Tex elements" dialog (maybe even with a progress bar). I'm thinking of a situation where a Tex compilation takes a while, the user gets frustrated and thinks it didn't work and clicks on the button once more, triggering a second run of the plugin function.
> 
> That is an interesting idea indeed.

I'm not so sure anymore we should add that. The plugin author is free to show a popup himself or to use a placeholder for showing the current status.

src/core/plugin/luapi_application.h
bhennion: So, shouldn't we rather always refresh the page (and allow nil)?

src/core/plugin/luapi_application.h
bhennion: I've giventhis some thoughts. There is still something bothering me: while the Tex elements are being processed, the user can freely interact with the document. Are we sure the API is safe there? Can the user not inadvertantly trigger a bug/segfault by modifying the document (e.g. removing a page) during compilation?

PR #7712, opened 2026-08-27
Add rotation (and any affine transformation) to all elements

### LINKED ISSUE #918: Image rotation support

### LINKED ISSUE #1566: Add way to rotate text. Text box tool?
There is no way to rotate text, as far as I can tell.

Maybe add a text box tool, which has a little rotate icon in a corner, that you can move to rotate the text.

**Additional context**
A text box tool could also add more possibilities, like having a fixed size box, that does not increase in size when you write more text. This would make it possible to write text, without having to manually set newline characters.

### LINKED ISSUE #4899: Rotate the added text

bhennion:
The PR picks up the work of #5601 to allow elements to be rotated.
Note that by allowing rotations (along with the already supported scaling), any affine transformation is now available:
Capture d&#39;écran_20260827_112613
Capture d&#39;écran_20260827_111908

Not quite done yet, but it is functional and can be tested.

- [x] Fix plugin API automatic image scaling
- [x] Add test units
- [x] Split off smaller refact PRs to ease reviewing
- [ ] Later PR: adapt the plugin API


tattsan:
Thank you for working on adding rotation support! This is a great feature, but I noticed a backward compatibility issue with files created in older versions of Xournal++.

**Problem Description**

When opening .xopp files created prior to this PR, scaled TeX objects appear misplaced or completely disappear in this build.
While forward incompatibility (opening new format files in older releases) is expected, backward compatibility seems broken for TeX elements.

**Observations**

* This regression specifically affects TeX objects that were resized/scaled in legacy versions.
* Standard text boxes and image elements are not affected and continue to render correctly.

**Steps to Reproduce**

1. Open a pre-PR build of Xournal++.
2. Insert a TeX object and scale/resize it.
3. Save the file.
4. Open the saved file in a build with this PR applied.
5. Notice that the TeX object is missing or shifted to an incorrect position.

bhennion:
> When opening .xopp files created prior to this PR, scaled TeX objects appear misplaced or completely disappear in this build. While forward incompatibility (opening new format files in older releases) is expected, backward compatibility seems broken for TeX elements.

Indeed, the coordinates were multiplied by the scaling ratios. This is fixed now. Thanks a lot for testing!

tattsan:
Works fine. Thanks!

bhennion:
I expanded the Load/Save test units quite a bit. I also split most test files into a "file version 4" and a "file version 5" versions, to ensure backward compatibility of the LoadHandler is maintained.

The added test files makes the diff quite bigger, but the "must-be-reviewed" part did not really change.

I think this PR is ready for review, once #7722 and #7723 (on which it is now based) are merged.

src/core/model/Element.h
```diff
@@ -17,7 +17,8 @@
 
 #include <gdk/gdk.h>  // for GdkRectangle
 
-#include "util/Color.h"                     // for Color
+#include "util/Color.h"  // for Color
+#include "util/Matrix.h"
```
rolandlo: The Matrix header is not really used here. It is used in `XmlParser.cpp` though, so better include it there than here. Also add the `// for Matrix` comment to make the diff not larger than needed.

src/core/model/RectangularElement.cpp
```diff
@@ -1,67 +1,85 @@
 #include "RectangularElement.h"
 
-#include <algorithm>  // for clamp
+#include <algorithm>  // for max, min
+#include <cmath>      // for ceil, floor, NAN
+#include <cstdint>    // for uint32_t
```
rolandlo:
```suggestion
#include <algorithm>  // for clamp
```

src/core/model/RectangularElement.cpp
```diff
@@ -1,67 +1,85 @@
...
 
-#include "util/Assert.h"
 #include "util/Point.h"
+#include "util/safe_casts.h"                      // for as_unsigned
```
rolandlo:
```suggestion
```

src/core/model/TexImage.cpp
```diff
@@ -40,14 +41,25 @@ auto TexImage::cloneTexImage() const -> std::unique_ptr<TexImage> {
...
+
+        xoj::util::Size<double> res{0, 0};
+        poppler_page_get_size(page, &res.width, &res.height);
+        return res;
+    }
+    if (image) {
+        return xoj::util::Size<double>(cairo_image_surface_get_width(image.get()),
+                                       cairo_image_surface_get_height(image.get()));
+    }
+    g_warning("TexImage without PDF nor Image...");
```
rolandlo:
```suggestion
    g_warning("TexImage without a PDF or an Image...");
```

src/core/model/TexImage.cpp
```diff
@@ -6,13 +6,14 @@
 #include <poppler-document.h>  // for poppler_document_ge...
 #include <poppler-page.h>      // for poppler_page_get_size
 
-#include "model/Element.h"                        // for Element, ELEMENT_TE...
-#include "util/Rectangle.h"                       // for Rectangle
+#include "model/Element.h"   // for Element, ELEMENT_TE...
+#include "util/Rectangle.h"  // for Rectangle
+#include "util/matrix/RectangleMultiply.h"
 #include "util/raii/GObjectSPtr.h"                // for GObjectSPtr
```
rolandlo:
```suggestion
#include "model/Element.h"                        // for Element, ELEMENT_TE...
#include "util/Rectangle.h"                       // for Rectangle
#include "util/matrix/RectangleMultiply.h"        // for operator*
#include "util/raii/GObjectSPtr.h"                // for GObjectSPtr
```

src/core/control/Control.cpp
```diff
@@ -2373,21 +2373,17 @@ void Control::clipboardPasteImage(GdkPixbuf* img) {
...
-        scaledWidth = pageWidth;
-        scaledHeight = (scaledWidth * height) / width;
+        scaling = pageWidth / width;
     }
-
-    if (scaledHeight > pageHeight) {
-        scaledHeight = pageHeight;
-        scaledWidth = (scaledHeight * width) / height;
+    if (height > pageHeight) {
+        scaling = std::max(scaling, pageHeight / height);
```
rolandlo: This looks wrong.
```suggestion
        scaling = std::min(scaling, pageHeight / height);
```

src/core/control/Control.cpp
rolandlo, line 2358: Inserting an image to a small page doesn't work properly.

src/core/view/overlays/TextEditionView.cpp
```diff
@@ -11,76 +12,100 @@ using namespace xoj::view;
...
     xoj::util::CairoSaveGuard saveGuard(cr);
 
     const double zoom = parent->getZoom();
     cairo_set_line_width(cr, BORDER_WIDTH_IN_PIXELS / zoom);
     Util::cairo_set_source_argb(cr, this->textEditor->getSelectionColor());
-    Range box = textEditor->getContentBoundingBox();
+    const auto& boxes = textEditor->getBoxes();
 
+
+    // We will apply the matrix be hand to the points when drawing the box(es), so that the line width is not deformed
```
rolandlo: Looks like there is some typo here.
```suggestion
    // We will apply the matrix by hand to the points when drawing the box(es), so that the line width is not deformed
```

rolandlo (commented):
This is great work! Thanks for filling that long-standing gap! Except of nitpicks I have only found three things:
- Clipboard pasting of images on small pages doesn't work as expected
- Exporting in .xoj compatibility mode stores the matrices and can't be read by Xournal
- After changing the font size the selection should be repainted in the new size

In the first commit (from Fabian) there are a couple of mistakes in matrix computations (like inverse matrix and translation), which you fixed in the second commit (without mentioning). It would be better to split the fixes into a separate commit, I feel, since otherwise one would expect that the Matrix class introduced there works as expected.

src/core/control/Control.cpp
bhennion: Does this fix the issue your described as "Clipboard pasting of images on small pages doesn't work as expected"?

rolandlo: No, it's not enough to fix the issue. There must be something more to it. The pasted image can be too wide and too high.

bhennion:
Thanks for the review!

> * Clipboard pasting of images on small pages doesn't work as expected

I'll have a look **Edit:** Fixed

> * Exporting in .xoj compatibility mode stores the matrices and can't be read by Xournal

What is our policy regarding this? How should the export behave? I see three possibilities

1. Do some form of "best we can", exporting the element's snapping bounds even if there is a transformation matrix (warn the user?)
2. Only export elements with transformation matrix being a scaling + translation (warn the user about the other elements?)
3. Refuse to export a document if it contains elements that are not compatible with the .xoj format

I'm not sure what I'd be in favour of here. Should we really maintain this export feature to a long-unmaintained app?

We may also want to implement something for .xopp documents, saving as before this PR the elements whose matrix is only a scaling + translation. The idea being that a document made with a new Xournal++ could then still be opened by an old one as soon as new features are not used.

> * After changing the font size the selection should be repainted in the new size

Actually, this has nothing to do with this PR. It is already present in v1.3.7. I'll have a look and open a PR for the relese branch.

> In the first commit (from Fabian) there are a couple of mistakes in matrix computations (like inverse matrix and translation), which you fixed in the second commit (without mentioning). It would be better to split the fixes into a separate commit, I feel, since otherwise one would expect that the Matrix class introduced there works as expected.

Done.

src/core/control/Control.cpp
bhennion: Fixed

bhennion:
> > * After changing the font size the selection should be repainted in the new size
> 
> Actually, this has nothing to do with this PR. It is already present in v1.3.7. I'll have a look and open a PR for the relese branch.

I opened #7743 about this so we don't forget. I could not figure out an easy fix

rolandlo (approved):
The clipboard pasting works fine now. The xoj-export is already broken in master and we can address that in a different PR. I'm not sure yet if we should fix it or remove it and rather add a fileversion 4 compatibility export. We could ask the Debian maintainer of Xournal about it.

bhennion:
> rather add a fileversion 4 compatibility export

I'm not sure I would add a specific feature. I think I would just ensure (adding a test) that if a file does not make effective use of the features introduced in v5, then it is readable as v4.
In practice, with the current changes in v5, it means that if there is no link and if the RectangularElements are not rotated and if the Texts are not streched unevenly, then the saved .xopp is readable by Xournal++ 1.3. The user would still get a warning when opening the file with v1.3.

We could even do one better: ensure the file is always saved in the oldest format that supports its content (well, v4 and up, I don't wan't to dig through the history too much either)

bhennion:
> In practice, with the current changes in v5, it means that if there is no link and if the RectangularElements are not rotated and if the Texts are not streched unevenly, then the saved .xopp is readable by Xournal++ 1.3. The user would still get a warning when opening the file with v1.3.

@rolandlo I implemented this idea in the last commit I just pushed on this branch. Could you have a look at the commit and tell me if you think this idea is worth the trouble?

rolandlo:
> > In practice, with the current changes in v5, it means that if there is no link and if the RectangularElements are not rotated and if the Texts are not streched unevenly, then the saved .xopp is readable by Xournal++ 1.3. The user would still get a warning when opening the file with v1.3.
> 
> @rolandlo I implemented this idea in the last commit I just pushed on this branch. Could you have a look at the commit and tell me if you think this idea is worth the trouble?

Yes, I think it is worth it. The implementation looks good to me. We can drop legacy saving in a couple of years when most distros ship the new version.

bhennion:
I rebase and cleaned up the commit tree. @rolandlo are you ok with me merging this?

rolandlo:
> I rebase and cleaned up the commit tree. @rolandlo are you ok with me merging this?

Great! I have no objections.

PR #7714, opened 2026-08-27
Fix Lasso selector

### LINKED ISSUE #7708: Select Region: selection malformed when cursor moves outside canvas near page edge
### Operating System

Windows

### (Linux only) Distribution

_No response_

_No response_

None

### Installation Method

Nightly build

### Xournal++ Version

1.3.5+dev

### libgtk Version

3.24.52

### Bug Description

The selection ends up malformed/incorrect.

A short video demonstrating the bug is attached (unlisted on YouTube): https://youtu.be/uU4Vne3XyCE

### Expected Behaviour

The selection should be correctly clamped to the page edge.

### Steps to Reproduce

1. Select the "Select Region" tool.
2. Start drawing a selection region normally on the page.
3. Move the cursor outside the drawable canvas area, so the intended selection touches the edge of the page on that side.
4. Release/complete the selection.Select Region: selection malformed when cursor moves outside canvas near page edge

### Additional Context

_No response_

bhennion:
- Fixed a bug #7708 introduced with 1e73a07
- Added signs to the selection algorithm so that WYSIWYG: when circling twice around an object, the object is selected

### DIFF
--- a/src/core/control/tools/Selector.cpp
+++ b/src/core/control/tools/Selector.cpp
@@ -1,20 +1,15 @@
 #include "Selector.h"
 
 #include <algorithm>  // for max, min
-#include <cmath>      // for abs, NAN
+#include <cmath>      // for abs
 #include <limits>     // for numeric_limits
 #include <memory>     // for __shared_ptr_access
-
-#include <gdk/gdk.h>  // for GdkRGBA, gdk_cairo_set_source_rgba
 
 #include "gui/LegacyRedrawable.h"  // for Redrawable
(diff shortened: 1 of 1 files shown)


rolandlo (approved):
LGTM, but is there a reason we don't simply use `cairo_in_fill` for that?

bhennion:
Well, we don't have a cairo context at this point. I also think we'd rather avoid relying on the display library to do this, if only to preserve the MVC pattern.

bhennion:
Thanks for the review!
Merging then

PR #7720, opened 2026-08-29
Clean up Audio content and Element inheritance

bhennion:
This PR is a prequel to #7712 

It cleans up the Audio business a tiny bit, to make room for another inheritance tree for Elements. In practice, this PR does:

1. Decorelates the audio content from the Element class. AudioElement becomes AudioContent and no longer inherits from Element. Text and Stoke now inherit from both Element and AudioContent.
2. Removes unused XmlAudioNode

Point 1. is useful in #7712 so we can have a `class RectangularElement: Element` without using virtual inheritance. It also makes more sense to decorelate Audio and Element as those do not really interact.


src/core/model/Text.cpp
```diff
@@ -32,8 +32,8 @@ auto Text::cloneText() const -> std::unique_ptr<Text> {
     text->font = this->font;
     text->text = this->text;
     text->setColor(this->getColor());
+    *static_cast<AudioContent*>(text.get()) = *this;
```
rolandlo: Maybe this could be made more readable:
```suggestion
    static_cast<AudioContent&>(*text) = static_cast<const AudioContent&>(*this);
```

src/core/model/Stroke.cpp
```diff
@@ -95,7 +95,7 @@ void Stroke::applyStyleFrom(const Stroke* other) {
     setStrokeCapStyle(other->getStrokeCapStyle());
     setLineStyle(other->getLineStyle());
 
-    cloneAudioData(other);
+    *static_cast<AudioContent*>(this) = *other;
```
rolandlo: Same comment here:
```suggestion
    static_cast<AudioContent&>(*this) = static_cast<const AudioContent&>(*other);
```

rolandlo (approved):
LGTM

bhennion:
Thanks for the review!
Merging once the CI clears

PR #7721, opened 2026-08-29
Small clean up of model/Image.cpp

bhennion:
This PR is a prequel to #7712.

It contains code cleanup in model/Image.cpp:

- Use RAII for the buffer
- Remove code duplication
- Introduce xoj::util::Size for {width, height}


rolandlo (approved):
LGTM

PR #7722, opened 2026-08-29
Fix potential SegFault in LatexController if selectedElem gets destroyed

bhennion:
This PR is a prequel to #7712

It cleans up a couple of things in LatexController, including a potential SegFault.


rolandlo (approved):
LGTM

PR #7724, opened 2026-08-29
Update MacOS deps

bhennion:
Because of instability of ftpmirror.gnu.org, we have been having issues with the MacOS dependency build CI.

This PR fixes that and bumps the MacOS deps versions as well.

Note: I did not bump lua from 5.4 to 5.5, but we could. @rolandlo would that be safe?

### DIFF
--- a/.github/actions/install_deps_mac/action.yml
+++ b/.github/actions/install_deps_mac/action.yml
@@ -10,13 +10,13 @@ runs:
   steps:
   - name: 'Attempt to pull blob from cache'
     id: restore-blob
-    uses: actions/cache/restore@v5
+    uses: actions/cache/restore@v6
     with:
       path: "${{github.workspace}}/blob"
       key: ${{ runner.os }}-${{ runner.arch }}-blob-${{ hashFiles('mac-setup/CI_jhbuild.sh', 'mac-setup/jhbuild-version.lock', 'mac-setup/xournalpp.modules', 'mac-setup/gtk-osx.patch') }}
   - name: 'Install Perl XML::Parser' # Needed for compilation of Intltool in the blob
     if: steps.restore-blob.outputs.cache-hit != 'true'
-    uses: perl-actions/install-with-cpm@v1
+    uses: perl-actions/install-with-cpm@v2
     with:
       install: XML::Parser
   - name: 'Compile blob'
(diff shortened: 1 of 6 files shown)


bhennion:
> Because of instability of ftpmirror.gnu.org, we have been having issues with the MacOS dependency build CI.

So it turns out going back to `ftp.gnu.org` does not quite work either.
We struck a bit of luck though, with one [CI run](https://github.com/xournalpp/xournalpp/actions/runs/33256197225) managing to build the deps (at least for PR targetted at the master branch).

Still, bumping the deps versions is probably a good idea.

bhennion:
\action create-installers mac mac-arm

bhennion:

rolandlo:
> Note: I did not bump lua from 5.4 to 5.5, but we could. @rolandlo would that be safe?

As long as only plugins that we ship are run bumping Lua from 5.4 to 5.5 should be perfectly safe. There could be other plugins in use that are not (yet made) compatible with Lua 5.5. There could also be Lua modules in use that are not compatible with Lua 5.5 yet. However I would think this won't affect many users (if any) and we'll have to make this version bump sooner or later anyway. We could do it now.

bhennion:
Actually, our Windows builds already ship Lua 5.5. I don't know why we didn't bump the MacOS version then.

bhennion:

bhennion:

bhennion:
\action create-installers mac-arm

bhennion:
\action create-installers mac-x64

bhennion:

bhennion:
> CI successful! Available artifacts:
> 
> * [MacOS package macOS-ARM64](https://github.com/xournalpp/xournalpp/actions/runs/35234992800/artifacts/10505891422)
> * [MacOS package macOS-X64](https://github.com/xournalpp/xournalpp/actions/runs/35234992800/artifacts/10509642869)

@rolandlo Would you have time to do a quick test of those?
After that, I'll prepare a v1.3.8 release, I think

rolandlo:
> > CI successful! Available artifacts:
> > 
> > * [MacOS package macOS-ARM64](https://github.com/xournalpp/xournalpp/actions/runs/35234992800/artifacts/10505891422)
> > * [MacOS package macOS-X64](https://github.com/xournalpp/xournalpp/actions/runs/35234992800/artifacts/10509642869)
> 
> @rolandlo Would you have time to do a quick test of those? After that, I'll prepare a v1.3.8 release, I think

In setup-env.cpp there are some 5.4 Lua paths. Please fix them so LuaGObject works. Otherwise everything runs smoothly on MacOS ARM. I will also briefly test MacOS Intel.

bhennion:

rolandlo (approved):
It's working now.

PR #7733, opened 2026-09-03
Meson scripts

bhennion:
This PR adds meson script for building the app. Hopefully, with this, we can later try GTK's Android backend.

To build/install, you can for instance run, from the repo root
```
meson setup --prefix=$(pwd)/install build
meson compile -C build
meson install -C build
```
You'll then have `./build/install/bin/xournalpp`

Building, installing and running should work, but some work is still required (see below). It is overall much easier to read than the cmake scripts.

- [x] Make test units work
- [x] Reinstate cpptrace
- [ ] Replace CPACK packaging tool (with what?)
- [ ] Reinstate gcov?
- [x] Reinstate Windows icon generation (or add the generated icon to the sources)
- [ ] Reinstate iOS-specific code
   - [ ]  in src/meson.build
   - [x] in meson.build (deployment version)
- [x] Reinstate Windows .rc file compilation
- [ ] Remove PACKAGE_TIMESTAMP in generated VERSION file?
- [x] ~~Preset~~ option INSTALL_DESKTOP_FILES dependong on target system **Edit:** ignored on Windows/iOS
- [ ] Reinstate ONLY_CONFIGURE_FOR_POT_TARGET
- [x] Reinstate configuration of GLIB_VERSION_MIN_REQUIRED, GDK_VERSION_MIN_REQUIRED, X11_ENABLED and G_LOG_DOMAIN
- [ ] Cleanup commented cmake code
- [ ] Adapt all the github actions
- [x] Remove glob generation of source lists?
- [ ] Test and test and test


rolandlo:
The Windows build works for me locally with this patch:

````patch
diff --git a/src/exe/win32/meson.build b/src/exe/win32/meson.build
index ab4c290e9..a451cc712 100644
--- a/src/exe/win32/meson.build
+++ b/src/exe/win32/meson.build
@@ -8,20 +8,27 @@ endif

 image_magick = find_program('magick')

-xournalpp_extra_deps += custom_target('ico',
+icon_target = custom_target('ico',
     command: [image_magick, '-density', '300', '-define', 'icon:auto-resize=256,128,96,64,48,32,24,16', '-background', 'none', meson.project_source_root() / 'ui/pixmaps/com.github.xournalpp.xournalpp.svg', '@OUTPUT@'],
     output : 'xournalpp.ico'
 )

-add_languages('rc')
 rc_config = configuration_data()
 rc_config.set('PROJECT_VERSION', meson.project_version())
-version_components = meson.project_version().split('.')
+version_without_suffix = meson.project_version().split('+')[0].split('-')[0].split('~')[0]
+version_components = version_without_suffix.split('.')
 rc_config.set('PROJECT_VERSION_MAJOR', version_components[0])
 rc_config.set('PROJECT_VERSION_MINOR', version_components[1])
 rc_config.set('PROJECT_VERSION_PATCH', version_components[2])

 xournalpp_rc_file = configure_file(input : 'xpp.rc.in', output : 'xpp.rc', configuration: rc_config, format : 'cmake')

-xournalpp_main_sources += xournalpp_rc_file
-xournalpp_wrapper_sources += xournalpp_rc_file
+win = import('windows')
+compiled_rc_file = win.compile_resources(
+    xournalpp_rc_file,
+    depends: icon_target,
+    include_directories : meson.current_build_dir()
+)
+
+xournalpp_main_sources += compiled_rc_file
+xournalpp_wrapper_sources += compiled_rc_file
```

bhennion:
\action create-installers windows mac-x64 mac-arm windows-arm

bhennion:
> \action create-installers windows mac-x64 mac-arm windows-arm

So I can't really test that here because the scripts are pulled from the master branch. I'll test on my fork and post here the results (once I get it to work)

bhennion:
Alright, so I got the CI to build and package something on Windows and Mac. See [this post](https://github.com/bhennion/xournalpp/pull/42#issuecomment-5667521005) for the latest installers I generated (on my fork, for technical reasons).
There are CMake flags (CMAKE_WIN32_EXECUTABLE and CMAKE_MACOS_BUNDLE) which don't have an equivalent in Meson. Since I'm not quite certain what they really do, it may or may not work well. If someone could test some of the provided executables and tell me.

Now, for Ubuntu/Debian packaging, we cannot use CPACK anymore, but I can make it work on our CI easily enough.
We have another problem though (and it extends to other distributions).
The translation files must be installed in `/usr/share/locale/${lang}/LC_MESSAGES/xournalpp.mo` where `${lang}` is replaced by the language code (e.g `de` or `en_UK`). In order to do that in Meson, I used `custom_target()`'s `build_subdir` kwarg, introduced in Meson 1.10.
The problem is that most distributions are shipping older versions of meson (the oldest mainstream one is ubuntu 22.04LTS, with meson 0.61.2)

The problem is: we cannot have several targets outputting the same file name (`xournalpp.mo`), and we cannot rename a file `${lang}.mo` into `${lang}/LC_MESSAGES/xournalpp.mo` upon installation either (before v1.10). I found a hack around the problem (see last commit), using the built-in `i18n.gettext()` but it has the following caveats:
1. The toolbar.ini.in and pagetemplates.ini.in must be renamed toolbar.ini.desktop.in/pagetemplates.ini.desktop.in to fool xgettext into thinking those are desktop files
2. The `name` attribute in toolbar.ini.in and pagetemplates.ini.in must be renamed `Name`, so the files behave like desktop files
3. `ui/mainmenubar.xml` must be renamed `ui/mainmenubar.ui`

Now, 3. is harmless, 2. could be handled with extra care, 1. could induce some confusion amongst devs (those are really not .desktop files...).
So the questions are: are those caveats acceptable? or is there another solution? how much do we care about porting to meson?

PR #7735, opened 2026-09-04
docs: remove unnecessary dvipng dependency

youssefadly237:
Remove the unused `dvipng` / `texlive-dvipng` dependency

relates to [Debian BTS #1126180](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1126180)


rolandlo (approved):
Thanks for cleaning this up!

PR #7741, opened 2026-09-08
Fix false "impossible event" discards when panning with hand tool

youssefadly237:
rn when holding hand tool with the tip of the stylus going out and then in the window it fires false positive

```
xopp-Message: xx:xx:xx.xxx: Discard impossible event - this is a sign of bugged hardware or drivers
```

which is not true


bhennion (approved):
LGTM!

Issue #1975, opened 2020-05-13
When I use Highlighter tool with a button on stylus it only allows the usage of a specific color (missing option "Color - don`t change")


kostyfisik:
**Affects versions :**
 - OS: Ubuntu Linux 19.10
 - (Linux only) Desktop environment: Gnome
 - Which version of libgtk do you use 3.24
 - Version of Xournal++: 1.1.0-dev
 - Installation method: PPA

**Describe the bug**
When I use Highlighter tool with a button on stylus is only allow usage of black color

**To Reproduce**
1. Set stylus button to become a Highlighter
2. Select Highlighter color to be non-black
3. Try to highlight when it is selected on toolbar
4. Try it via button on stylus.

**Expected behavior**
Use the current color of the Highlighter tool 

*Additional context**

steps 3 and 4 from description above:

![our-final-product](https://user-images.githubusercontent.com/5239997/81811477-59bf5600-952d-11ea-868e-420a350bff3e.gif)

rolandlo:
I can't reproduce this behaviour. Are you sure you set the highlighter color to something else than black in the pen settings? Could you post a picture of your pen settings, please?

kostyfisik:
See new screencast with switch from HL to pen and back using a mouse (everything is OK) and stylus (the HL color is reset to black as soon as I use pen button)
![our-final-product](https://user-images.githubusercontent.com/5239997/81830060-44a1f180-9544-11ea-92bc-20a288b8234b.gif)

rolandlo:
Still this doesn't answer what you have in the settings. Look at my settings then:
![image](https://user-images.githubusercontent.com/40485433/81831373-8da27780-953d-11ea-8760-c3eb4d1e34a5.png)
The pen buttons are mapped to the highliter with yellow and violet color (see on the right). If you have black color selected there, you will get a black highliter, whenever you use the pen button. It doesn't matter what you have selected in the toolbar when you use the highliter via pen button. All that counts is what is in your settings.

kostyfisik:
Ok, I see. So, it is a missing default option for "Color - don`t change". 

It`s a bit strange (and non-consistent) to allow one parameter (thickness) of highlight to be unchanged (by default) and to force to set another to a specific value. This can be treated as UX bug.

A side note that I needed a second look to find a black color choice box using dark theme :)

See my options.

[...]

PR #7752, opened 2026-09-15
feat: add "Color - don't change" for button mapping

youssefadly237:
add the ability to skip choosing a fixed color on mouse/stylus buttons mapping, by adding a "don't change" mirroring thickness/type

refs: #1975, #7748


bhennion (commented):
Thanks for the PR!
The code looks good to me. I'll try to find some time to test it later this week

bhennion (approved):
LGTM, with the following nitpick: could we maybe grey-out the colorpicker rather than hiding it when "Don't change" is selected? It would avoid the window resizing when changing the combobox.

youssefadly237:
Yeah sure

youssefadly237:
sorry for being late I was not home the last 2 days

bhennion:
> sorry for being late I was not home the last 2 days

No worries of course! Thanks!

Squashing and merging in 48 hours unless an objection is raised.

PR #7756, opened 2026-09-16
Fix UBs on release branch

bhennion:
This PR comes from warnings I noticed on various compilers when working on #7733.
It fixes instances of undefined behaviour which could lead to memleaks or bugs:

- virtual destructors for SaveHandler and XojPdfPage: otherwise, when destroying a instance of a derived class via a pointer to the base, we get a memleak
- use locale-independent functions to parse metadata files (same as when writing them)


rolandlo (approved):
LGTM

bhennion:
Thanks for the review!

